PR #4006, opened 2020-09-29
fix(electron): do not use --require to throttle startup

### LINKED ISSUE #3997: [BUG] Cannot launch compiled electron application
**Context:**
- Playwright Version: playwright-electron@0.4.0
- Operating System: Mac
- Node.js version: v12.18.3
- Browser: Electron

**Code Snippet**

Throws error: `Exception has occurred: TypeError: Cannot read property 'evaluateHandle' of undefined`

```javascript
const { electron } = require("playwright-electron");
(async () => {
  const app = await electron.launch(
    "/Applications/Slack.app/Contents/MacOS/Slack"
  );
  await app.close();
})();
```

Works with spectron

```javascript
const { Application } = require("spectron");

(async () => {
  const app = new Application({
    path: "/Applications/Slack.app/Contents/MacOS/Slack",
  });
  await app.start();
  console.log("title", await app.client.getTitle());
  await app.stop();
})();
```

**Describe the bug**

It seems the issue is that [`_nodeElectronHandle`](https://github.com/microsoft/playwright/blob/109688a066a3168177b0244534b0d35dcd6161a9/src/server/electron/electron.ts#L133) never resolves causing the [`_onPage`](https://github.com/microsoft/playwright/blob/109688a066a3168177b0244534b0d35dcd6161a9/src/server/electron/electron.ts#L81) to throw an error.

pavelfeldman:
Fixes #3997


jperl:
Thank you 🙏, I was able to run the script with this fix.

Issue #3996, opened 2020-09-28
Review the new video and trace APIs


pavelfeldman:

arjun27:
Once the API is finalized, we should also add videos to the [verification](https://playwright.dev/#version=v1.4.2&path=docs%2Fverification.md&q=) doc.

PR #4015, opened 2020-09-30
api: update videos api, hide tracing

dgozman:
In preparation for 1.5, this patch:
- Moves `artifactsPath` to `newContext() / launchPersistentContext()`, renames it to `videosPath`.
- Hides `_traceResourcesPath` and `_tracePath` in `newContext()`.
- Throws for `videospath` in remote connected browser since we do not stream it to the client yet.
- Adds a test for videos in persistent context.

References #3996.


src/client/types.ts
```diff
@@ -82,7 +82,6 @@ export type LaunchServerOptions = {
     password?: string
   },
   downloadsPath?: string,
```
yury-s, line 84: Do we still need a separate path for downloads or we could just merge it with artifactsPath?

yury-s (approved):

src/client/types.ts
dgozman: We'll probably move `downloadsPath` to `newContext` as well. Let's see whether videos work out as we want them to.

pavelfeldman (approved):

PR #4018, opened 2020-09-30
fix(close): fix a race during context.close and page.close

dgozman:
There is a race between "close" event coming from the server and
"close" command issued from the client.

This is similar to calling close after disconnect, so added tests.


src/client/page.ts
```diff
@@ -451,11 +451,19 @@ export class Page extends ChannelOwner<channels.PageChannel, channels.PageInitia
...
+    try {
+      await this._wrapApiCall('page.close', async () => {
+        await this._channel.close(options);
+        if (this._ownedContext)
+          await this._ownedContext.close();
+      });
+    } catch (e) {
+      if (e.message === 'page.close: Browser has been closed')
+        return;
+      if (e.message === 'page.close: Target browser or context has been closed')
```
yury-s, line 463: This is rather brittle, should we just use a special class for such errors/mark them with a symbol?

yury-s (approved):

src/client/page.ts
dgozman: I can do a shared constant for this.

PR #4039, opened 2020-10-02
browser(firefox): remove multisession logic

aslushnikov:
This patch:
1. Changes `SimpleChannel` to buffer messages to the namespace that
   hasn't been registered yet. This allows us to create `SimpleChannel`
   per target on the browser side right away.
2. Removes multisession support. Now there's only one `PageAgent` in the
   content process, which talks to a single `PageHandler` on the browser
   side. Both ends can be created as-soon-as-needed; thanks to
   `SimpleChannel` bufferring, no messages will be lost and all messages
   will be delivered in proper order. (This is currently the reason why
   build 1178 flakes on windows).
3. Straightens up the target reporting. Targets are reported as soon
   as they appear on the browser side.

**NOTE:** this doesn't yet remove sessions from protocol.

References #3995

### DIFF
--- a/browser_patches/firefox/juggler/SimpleChannel.js
+++ b/browser_patches/firefox/juggler/SimpleChannel.js
@@ -30,6 +30,7 @@ class SimpleChannel {
     this._connectorId = 0;
     this._pendingMessages = new Map();
     this._handlers = new Map();
+    this._bufferedRequests = [];
     this.transport = {
       sendMessage: null,
       dispose: null,
@@ -70,6 +71,14 @@ class SimpleChannel {
     if (this._handlers.has(namespace))
       throw new Error('ERROR: double-register for namespace ' + namespace);
     this._handlers.set(namespace, handler);
+    // Try to re-deliver all pending messages.
+    Promise.resolve().then(() => {
+      const bufferedRequests = this._bufferedRequests;
+      this._bufferedRequests = [];
+      for (const data of bufferedRequests) {
+        this._onMessage(data);
+      }
+    });
     return () => this.unregister(namespace);
   }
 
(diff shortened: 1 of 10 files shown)


browser_patches/firefox/juggler/SimpleChannel.js
```diff
@@ -70,6 +71,14 @@ class SimpleChannel {
     if (this._handlers.has(namespace))
       throw new Error('ERROR: double-register for namespace ' + namespace);
     this._handlers.set(namespace, handler);
+    // Try to re-deliver all pending messages.
+    Promise.resolve().then(() => {
```
aslushnikov, line 75: do an explicit method instead: `ready`!

browser_patches/firefox/juggler/protocol/BrowserHandler.js
```diff
@@ -128,8 +127,10 @@ class BrowserHandler {
   }
 
   async newPage({browserContextId}) {
-    const targetId = await this._targetRegistry.newPage({browserContextId});
-    return {targetId};
+    const target = await this._targetRegistry.newPage({browserContextId});
```
aslushnikov, line 130: revert this back

browser_patches/firefox/juggler/TargetRegistry.js
```diff
@@ -423,13 +402,10 @@ class PageTarget {
...
   }
 
-  connectSession(session) {
-    this._channel.connect('').send('attach', { sessionId: session.sessionId() });
-  }
-
-  disconnectSession(session) {
-    if (!this._disposed)
-      this._channel.connect('').emit('detach', { sessionId: session.sessionId() });
+  async disconnectSession(session) {
```
aslushnikov, line 405: kill disconnect

dgozman (approved):

PR #4045, opened 2020-10-02
browser(firefox): start screencast in existing pages upon setScreencastOptions

dgozman:
This makes it work in persistent context.

To achieve this, we have to move screencast logic into PageTarget and
make PageHandler listen to an event.

References #3996.

### DIFF
--- a/browser_patches/firefox/BUILD_NUMBER
+++ b/browser_patches/firefox/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1179
-Changed: lushnikov@chromium.org Fri Oct  2 03:14:15 PDT 2020
+1180
+Changed: dgozman@gmail.com Fri Oct  2 09:36:06 PDT 2020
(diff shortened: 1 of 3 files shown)


aslushnikov:
Why didn't it work today?

yury-s (approved):
I'd rather change our default context logic to manually create first page, it would allow to get rid of this workaround as well as hairy logic in _loadDefaultContext. This is a bigger cross-browser change though so we can probably live with the status quo for longer.

yury-s:
> Why didn't it work today?

With persistent context browser creates first page which may happen before setScreencastOptions take effect. I believe this is a general problem that first page in default context may not inherit some of the context properties.

dgozman:
> I'd rather change our default context logic to manually create first page, it would allow to get rid of this workaround as well as hairy logic in _loadDefaultContext. This is a bigger cross-browser change though so we can probably live with the status quo for longer.

Yeah, we can put this into our backlog 😄

PR #4048, opened 2020-10-02
fix(api.json): do not copy documentation from base class members

yury-s:
* Add `extends` field
* Do not copy fields from super class to subclass definition


pavelfeldman:
Are you trying to save on the footprint? (We'll need to update Python generators)

pavelfeldman (approved):

yury-s:
No, this polutes generated java interfaces with extra methods instead of adding only the new ones.

PR #4065, opened 2020-10-06
browser(firefox): enable document channel

aslushnikov:
In the current tip-of-tree Firefox, document channel is enabled by
default, so we have to enable it in order to roll further.

This patch:
1. Removes content disposition sniffing from content process since it
   crashes renderer with document channel.
2. Merges all page-related handlers in a single `PageHandler` and
   serializes network events wrt the `Page.frameAttached` event.

The serialization mentioned in (2) is necessary: frame attachment is
reported from the content process, and network events are reported from
the browsers process. This is an inherent race, that becomes exposed by
the document channel.

On a side note, (2) makes it possible to synchronously report all
buffered events in `SimpleChannel` (cc offline discussion with @dgozman
that highlighted an unsighty approach that we currently employ there: reporting
events in a subsequent microtask.)

References #3995


aslushnikov:
TBR: @dgozman

PR #4070, opened 2020-10-06
fix: Page.close() should wait for beforeunload to finish

yury-s:


yury-s:
Will keep current behavior with updated docs: #4086

PR #4076, opened 2020-10-07
fix(windows): fix dependencies check on windows

kblok:
When running the playwright driver on ISS Express, LoadLibraryEx is not loading `dlls` found in the same directory as the browser.
To solve that I'm adding the current directory as a dll path `SetDllDirectoryA(".");` and adding the flags ` | LOAD_LIBRARY_SEARCH_USER_DIRS | LOAD_LIBRARY_SEARCH_SYSTEM32`

Related to https://github.com/microsoft/playwright-sharp/issues/911


browser_patches/winldd/PrintDeps.cpp
```diff
@@ -95,8 +95,9 @@ const DepsMap getDependencies(const HMODULE hMod)
 }
 
 int printDependencies(const char* library)
-{
-    HMODULE hMod = LoadLibraryEx(library, NULL, DONT_RESOLVE_DLL_REFERENCES);
+{
+    SetDllDirectoryA(".");
+    HMODULE hMod = LoadLibraryEx(library, NULL, DONT_RESOLVE_DLL_REFERENCES | LOAD_LIBRARY_SEARCH_USER_DIRS | LOAD_LIBRARY_SEARCH_SYSTEM32);
```
yury-s, line 100: Shouldn't we call `SetDefaultDllDirectories(LOAD_LIBRARY_SEARCH_USER_DIRS | LOAD_LIBRARY_SEARCH_SYSTEM32)` or even `SetDefaultDllDirectories(LOAD_LIBRARY_SEARCH_DEFAULT_DIRS)` ? I suspect with the latter you may not need `SetDllDirectoryA(".");`

yury-s (approved):
nit: we can keep bin/PrintDeps.exe and let the next build override it

browser_patches/winldd/PrintDeps.cpp
kblok: Will try that!

kblok: It didn't work @yury-s. 
I think that `LOAD_LIBRARY_SEARCH_APPLICATION_DIR` won't work because the app is the PrintDeps.
`LOAD_LIBRARY_SEARCH_SYSTEM32` is not enough because we have dlls in the chromium folder.
The [docs](https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-setdefaultdlldirectories?WT.mc_id=DT-MVP-5003814) say that to make `LOAD_LIBRARY_SEARCH_USER_DIRS` you need to call `AddDllDirectory` or `SetDllDirectory`.

What do you think?

kblok:
@yury-s I rollbacked the exe.

lvmajor:
Is there any chance that this might get merged soon? This is blocking a project I'm working on as a downstream project (Playwright-Sharp) is awaiting this to be merged upstream before updating its own package. 

Thanks in advance!

browser_patches/winldd/PrintDeps.cpp
yury-s: Let's leave it as is then. The app dir for the deps checker is different indeed I missed that fact.

yury-s (approved):

PR #4080, opened 2020-10-07
browser(firefox): a different way to emit 'load' event

aslushnikov:
Using WebProgressListener events works in all cases. Currently
used `pageshow` event will stop being emitted in future when loading
was stopped with `window.stop()` api.

References #3995

### DIFF
--- a/browser_patches/firefox/BUILD_NUMBER
+++ b/browser_patches/firefox/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1184
-Changed: lushnikov@chromium.org Tue Oct  6 12:04:41 PDT 2020
+1185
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1949,6 +1949,16 @@
       "dev": true,
       "optional": true
     },
+    "bindings": {
```
yury-s, line 1952: Is it intended to be a part of this PR?

yury-s (approved):

dgozman (approved):

package-lock.json
aslushnikov: reverted

PR #4081, opened 2020-10-07
browser(firefox): wait for search and addon manager initialization

aslushnikov:
It looks like terminating browser when search service or addon manager is
not fully initialized results in a broken shutdown sequence. As of
today, this results in multiple errors in the browser STDERR. In future,
this might also result in browser stalling instead of terminating.

This starts awaiting search and addon manager termination.

References #3995


dgozman (approved):
I wonder whether this slows down Firefox launch even more? 😄

aslushnikov:
> I wonder whether this slows down Firefox launch even more? 😄

We'll see how much! 🤷‍♂️

PR #4088, opened 2020-10-07
devops: fix firefox build

aslushnikov:
New firefox build requires a pre-created python virtual environment.
We should detect it and create if necessary.

References #3995


aslushnikov:
TBR @yury-s

Issue #4084, opened 2020-10-07
[BUG] more user-friendliness when using docker containers


aslushnikov:
Based on the feedback we received, here are two action items to improve user experience of playwright:
- [x] Add `pwuser` back to the playwright docker image, but keep running containers under `root`. **Motivation:** This would be similar to the `node` user in the node docker images, and `nginx` user in nginx images. 
- [x] Switch `chromiumSandbox` to `false` by default. **Motivation:** Certain environments like Azure Pipelines override default container user with a custom one and fail to launch container with a proper seccomp profile. Auto-detecting chromium sandboxing capabilities is hard - the best way would be to launch a browser and wait for renderer crash. This is a fragile approach; instead, default to no sandbox similarly to the webkit.

aslushnikov:
cc'ing @mxschmitt who had a patch for `pwuser`.

PR #4090, opened 2020-10-07
fix: disable chromium sandbox by default

aslushnikov:
Certain environments, e.g. Azure Pipelines, override default user
inside container with a custom one, whereas fail to pass proper
seccomp profile for the docker image.

As a result, chromium sandboxing fails.

To ease life of devops deploying tests in various CI's, this patch
disables Chromium sandbox by default.

References #4084


pavelfeldman (approved):
Please mark this as a breaking change in the release notes.

mxschmitt (commented):
can we remove the logic which enabled automatically the sandbox if the user is root?

PR #4095, opened 2020-10-08
chore: fix driver compatibility with PKG

mxschmitt:


pavelfeldman (approved):

src/driver.ts
```diff
@@ -31,7 +32,8 @@ const readFileAsync = util.promisify(fs.readFile);
 const writeFileAsync = util.promisify(fs.writeFile);
 
 export async function copyPrintDeps(destination: string) {
-  const content = await readFileAsync(require.resolve('../bin/PrintDeps.exe'));
+  // Needs to be path.join() otherwise PKG wants to minify it
+  const content = await readFileAsync(path.join(__dirname, '../bin/PrintDeps.exe'));
```
pavelfeldman, line 36: Why would we want to bake those into the driver?

pavelfeldman:
I'm thinking of making playwright-cli do this all. So this file is likely to go away.

mxschmitt:
its now done in the cli

PR #4097, opened 2020-10-09
feat: tap

JoelEinbinder:
This is currently just the api and documentation. I'll add tests an impls as I finish them, but I think its easier to review this part independently.

Adds
`page.touchscreen.tap`
`page.tap`
`frame.tap`
`elementHandle.tap`

`hasTouch` must be enabled on the browser context level to use tap, otherwise it throws.

Blocked on #4099, #4101 and #4102

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -18,6 +18,7 @@
 - [class: FileChooser](#class-filechooser)
 - [class: Keyboard](#class-keyboard)
 - [class: Mouse](#class-mouse)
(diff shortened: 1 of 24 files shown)


dgozman (commented):
Do we have tests for tap? I assume we need good coverage, checking touch/pointer events, touchstart preventDefault and triggered mouse/click events.

dgozman (approved):

PR #4100, opened 2020-10-09
feat: introduce `mouse.currentState()` api

megaserg:
This adds a new `mouse.currentState()` method that will return
current mouse state. For now, this has only the `x` and `y`
coordinates.


pavelfeldman:
Closing this in favor of `PWDEBUG=1` that has already landed.

PR #4101, opened 2020-10-09
browser(firefox): Page.dispatchTapEvent

JoelEinbinder:
Adds a new method to send a `touchup`, `touchend`, and the associated fake mouse events. The c++ code plumbs through a boolean to make sure the mouse events to create duplicate pointer events.

https://github.com/JoelEinbinder/juggler/commit/20acd5c472fcb2c8516372577c97e445df2161a1

### DIFF
--- a/browser_patches/firefox/BUILD_NUMBER
+++ b/browser_patches/firefox/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1189
-Changed: lushnikov@chromium.org Mon Oct 12 22:21:30 PDT 2020
+1190
+Changed: joel.einbinder@gmail.com Tue 13 Oct 2020 09:08:57 AM PDT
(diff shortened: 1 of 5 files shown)


dgozman (commented):
I see the manual dispatch of mousemove, mousedown and mouseup. Does this also produce "click"? I wonder why some of the logic is hardcoded, and some is automatic.

JoelEinbinder:
This produces clicks. For real user input, the tap gesture is recognized by the operating system/platform specific code and sent after the touches. Firefox then sends a ton of messages between processes to decide if/when/how the fake click will happen.

Doing everything in JavaScript and duplicating some logic is a lot easier, and matches what we (and marionette) do with mouse and keyboard input.

The playwright side of this patch has a lot of tests. It generates click/pointer events/hovers correctly.

aslushnikov:
So why do we want to disable pointer events?

JoelEinbinder:
When you click on the screen, you use get one set of pointer events and one set of mouse events. When you tap on the screen you get one set of pointer events, one set of touch events, and one set of mouse events.

When we are emulating the tap, we send touch events which generate pointer events. The generated mouse events are usually sent by the browser process, so this patch sends them directly like we sent other mouse events. We have to disable the pointer events though otherwise we'd get two sets of pointer events.

dgozman (approved):

PR #4127, opened 2020-10-13
Removing "Codex" as it isn't using Playwright

danawoodman:
I looked at that repo and could not find anywhere where Playwright was being used. I think it would be good to remove to prevent confusion.


arjunattam (approved):

arjunattam:
I believe the Playwright tests have been moved to a different repo: https://github.com/codex-src/codex-playwright

PR #4132, opened 2020-10-13
api(video): restore the missing video path accessor

### LINKED ISSUE #4131: [Question] How to name video files
Hi team. I love the new video API in 1.5.0. Very clean. I don't see any information about how to name video files in the examples though. Seems like each file is just assigned a random name with no way to change that. When the API was in beta, we could get a handle to the file in order to rename it. Am I missing something in the docs?
https://playwright.dev/#path=docs%2Fverification.md&q=videos

Thanks!

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/4131

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -14,6 +14,7 @@
 - [class: ConsoleMessage](#class-consolemessage)
 - [class: Dialog](#class-dialog)
 - [class: Download](#class-download)
+- [class: Video](#class-video)
 - [class: FileChooser](#class-filechooser)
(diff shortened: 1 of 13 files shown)


yury-s (approved):
I don't think we need this event. We'll very soon need to add things like saveAs, waitForFinished etc. All the CI use cases should work with the existing API where one can specify video folder per context and each test is supposed to use separate context. So I'd wait for more feedback before adding the event back.

dgozman (approved):

PR #4133, opened 2020-10-13
feat(firefox): roll to r1190

aslushnikov:


JoelEinbinder:
This test is unhappy: `workers.spec.ts:97:1 › should clear upon cross-process navigation`

JoelEinbinder (approved):

PR #4136, opened 2020-10-14
Firefox user preferences only applied to persistent

mjgp2:
These preferences are currently only applied for persistent browsers, but they should be applied to both.

### DIFF
--- a/src/server/firefox/firefox.ts
+++ b/src/server/firefox/firefox.ts
@@ -63,10 +63,9 @@ export class Firefox extends BrowserType {
       throw new Error('Pass userDataDir parameter instead of specifying -profile argument');
     if (args.find(arg => arg.startsWith('-juggler')))
       throw new Error('Use the port parameter instead of -juggler argument');
-    const firefoxUserPrefs = isPersistent ? undefined : options.firefoxUserPrefs;
-    if (firefoxUserPrefs) {
+    if (options.firefoxUserPrefs) {
       const lines: string[] = [];
-      for (const [name, value] of Object.entries(firefoxUserPrefs))
+      for (const [name, value] of Object.entries(options.firefoxUserPrefs))
         lines.push(`user_pref(${JSON.stringify(name)}, ${JSON.stringify(value)});`);
       fs.writeFileSync(path.join(userDataDir, 'user.js'), lines.join('\n'));
     }


src/server/firefox/firefox.ts
```diff
@@ -63,10 +63,9 @@ export class Firefox extends BrowserType {
...
     if (args.find(arg => arg.startsWith('-juggler')))
       throw new Error('Use the port parameter instead of -juggler argument');
-    const firefoxUserPrefs = isPersistent ? undefined : options.firefoxUserPrefs;
-    if (firefoxUserPrefs) {
+    if (options.firefoxUserPrefs) {
       const lines: string[] = [];
-      for (const [name, value] of Object.entries(firefoxUserPrefs))
+      for (const [name, value] of Object.entries(options.firefoxUserPrefs))
         lines.push(`user_pref(${JSON.stringify(name)}, ${JSON.stringify(value)});`);
       fs.writeFileSync(path.join(userDataDir, 'user.js'), lines.join('\n'));
```
pavelfeldman, line 70: We end up creating a file in the persistent profile this way, which you should be able to do directly, without using Playwright API. You have profile, you have profile options, so you can ensure user.js is there!

mjgp2: Hi! Thanks for your comment :) My use case here is that I do not want a persistent profile, but I need to set user preferences for the transient profile via the `browser.launch()` api. Looking at the code as I understand it this directory is created whether or not the profile is persistent, with the main difference being for transient browser launches is under `os.tmpdir()` and deleted (along with any other temp dirs) when the browser closes. As such I thought this change correctly fixes the `browser.launch()` API to have user prefs passed into it applied to a transient browser, which currently does not happen as per the API documentation?

dgozman: Sorry for the late response. If I read the code correctly, `options.firefoxUserPrefs` are actually ignored (`undefined`) in the persistent profile. However, if you use `browser.launch()`, these options are not ignored, because `isPersistent` is false. Do I read it wrong?

mjgp2: No, seemingly you don't read it wrong, I think I read it wrong at the time, which is very confusing because I am sure they were not getting applied before I made this (seemingly incorrect) patch. I will go back and redo the test next week and double check.

PR #4138, opened 2020-10-14
fix(keyboard): send alternate keys on mac and webkit

JoelEinbinder:
The shiftKey test was failing on Firefox, but actually it was the other browsers that were wrong. On windows and linux, holding alt while pressing keys prevents text from being typed. But on Mac, holding alt causes alternate text to be typed. I updated the USKeyboardLayout with the alternate text, and it is now properly typed when the browser is running on a mac. I also turn it on unconditionally for webkit, because we sort of always are emulating a mac for webkit.


src/server/firefox/ffBrowser.ts
```diff
@@ -91,6 +91,7 @@ export class FFBrowser extends Browser {
 
   async _initVersion() {
     const result = await this._connection.send('Browser.getInfo');
+    this._isMac = result.userAgent.includes('Macintosh');
```
mxschmitt, line 94: If the user would overwrite the user agent when the host is e.g. Mac and the new User-Agent is Linux, this would not work or am I wrong?

JoelEinbinder: This userAgent is the real one that comes from the browser, not the one overwritten by the user. So it depends what you mean by not work".

JoelEinbinder: We don't emulate `navigator.platform`, but if we started it might make sense to wrap this behavior in there.

dgozman (approved):

JoelEinbinder:
Not to myself because I keep forgetting. This patch doesn't quite work as is because webkit for windows/linux refuses to send characters when alt is held down. That makes perfect sense, because windows and linux don't do that. But it messes with our idea that "webkit is mac".

JoelEinbinder:
Firefox for mac has some issues with this patch. On the bots, mac shortcuts are broken. Locally on my mac they work, however `åÅ` is instead typed as `aA`. I'll reopen this patch again later when its more ready.

PR #4144, opened 2020-10-14
fix(screencast): correctly process videos with 1 frame

yury-s:


src/server/chromium/videoRecorder.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import { ChildProcess } from 'child_process';
+import { time } from 'console';
```
pavelfeldman: stray import

src/server/chromium/videoRecorder.ts
```diff
@@ -124,8 +126,8 @@ export class VideoRecorder {
 
     if (this._lastWriteTimestamp) {
       const durationSec = (Date.now() - this._lastWriteTimestamp) / 1000;
-      if (durationSec > 1 / fps)
-        this.writeFrame(this._lastFrameBuffer!, this._lastFrameTimestamp + durationSec);
+      if (!this._lastWritePromise || durationSec > 1 / fps)
+        this._flushLastFrame(durationSec).catch(e => this._progress.log('Error while writing frame: ' + e));
```
pavelfeldman, line 129: Shouldn't you await it?

yury-s: No, we chain them and wait for the tail in stop().

src/server/chromium/videoRecorder.ts
yury-s: Fixed.

src/server/chromium/videoRecorder.ts
pavelfeldman: Can we chain them explicitly?

pavelfeldman (approved):

PR #4146, opened 2020-10-14
fix(screencast): bump chromium video quality

### LINKED ISSUE #4142: [BUG] Video Recording Chromium compression is too high
**Context:**
- Playwright Version: 1.5
- Operating System: Linux
- Node.js version: 12
- Browser: Chromium

Example:

````js
// @ts-check

/**
 * Helper function which will compare val1 with val2.
 * If they dont equal itself it will throw an error.
 */
const expect = (val1, val2) => {
  if (val1 !== val2) {
    throw new Error(`${val1} does not match ${val2}`)
  }
}

const TODO_NAME = "Bake a cake";

(async () => {
  const browser = await playwright.chromium.launch({
    slowMo: 100
  });
    videosPath: 'videos/'
  });

  await page.goto("http://todomvc.com/examples/react/");

  // Helper function to get the amount of todos on the page
  const getCountOfTodos = () => page.$$eval("ul.todo-list > li", el => el.length)

  // Initially there should be 0 entries
  expect(await getCountOfTodos(), 0)

  // Adding a todo entry (click in the input, enter the todo title and press the Enter key)
  await page.click("input.new-todo");
  await page.type("input.new-todo", TODO_NAME);
  await page.press("input.new-todo", "Enter");

  // After adding 1 there should be 1 entry in the list
  expect(await getCountOfTodos(), 1)

  // Here we get the text in the first todo item to see if it's the same which the user has entered
  const textContentOfFirstTodoEntry = await page.$eval("ul.todo-list > li:nth-child(1) label", el => el.textContent)
  expect(textContentOfFirstTodoEntry, TODO_NAME)

  // The todo list should be persistent. Here we reload the page and see if the entry is still there
  await page.reload({
    waitUntil: "networkidle"
  })
  expect(await getCountOfTodos(), 1)

  // Set the entry to completed
  await page.click('input.toggle');

  // Filter for active entries. There should be 0, because we have completed the entry already
  await page.click('"Active"');
  expect(await getCountOfTodos(), 0)

  // If we filter now for completed entries, there should be 1
  await page.click('"Completed"');
  expect(await getCountOfTodos(), 1)

  // Clear the list of completed entries, then it should be again 0
  await page.click('"Clear completed"');
  expect(await getCountOfTodos(), 0)

  await browser.close();
})();

````

Interactive: https://try.playwright.tech/?e=todo-mvc

Expected: centered small text is readable
Actual: It's blurry and not readable. FF and WK is fine. Only persistent in CR.

Example recordings: [wk-cr-recordings.zip](https://github.com/microsoft/playwright/files/5380099/wk-cr-recordings.zip)

### PR DESCRIPTION
**Pull request by pavelfeldman**
Fixes https://github.com/microsoft/playwright/issues/4142

### DIFF
--- a/src/server/chromium/videoRecorder.ts
+++ b/src/server/chromium/videoRecorder.ts
@@ -53,7 +53,7 @@ export class VideoRecorder {
     assert(!this._isRunning());
     const w = options.width;
     const h = options.height;
-    const args = `-loglevel error -f image2pipe -c:v mjpeg -i - -y -an -r ${fps} -c:v vp8 -qmin 0 -qmax 50 -crf 8 -vf pad=${w}:${h}:0:0:gray,crop=${w}:${h}:0:0`.split(' ');
+    const args = `-loglevel error -f image2pipe -c:v mjpeg -i - -y -an -r ${fps} -c:v vp8 -qmin 0 -qmax 50 -crf 8 -b:v 1M -vf pad=${w}:${h}:0:0:gray,crop=${w}:${h}:0:0`.split(' ');
     args.push(options.outputFile);
     const progress = this._progress;
 

### CODE REVIEW

File: src/server/chromium/videoRecorder.ts [hunk 1]
```diff
@@ -53,7 +53,7 @@ export class VideoRecorder {
     assert(!this._isRunning());
     const w = options.width;
     const h = options.height;
-    const args = `-loglevel error -f image2pipe -c:v mjpeg -i - -y -an -r ${fps} -c:v vp8 -qmin 0 -qmax 50 -crf 8 -vf pad=${w}:${h}:0:0:gray,crop=${w}:${h}:0:0`.split(' ');
+    const args = `-loglevel error -f image2pipe -c:v mjpeg -i - -y -an -r ${fps} -c:v vp8 -minrate 800K -maxrate 1M -b:v 1M -vf pad=${w}:${h}:0:0:gray,crop=${w}:${h}:0:0`.split(' ');
```
yury-s, line 56: Won't `-minrate 800K` result in a much bigger file with big resolution video?

yury-s (approved):

src/server/chromium/videoRecorder.ts
pavelfeldman: dunno!

PR #4150, opened 2020-10-15
browser(firefox): always send focus events

JoelEinbinder:
I was going to do it the right way with a protocol method, but we already are unconditionally overriding the focus behavior in other places. So this one line hack I think is ok.

This fixes the test `frame.focus should work multiple times`


aslushnikov (approved):
Oops sorry, I side-stepped you on the r1192!

PR #4152, opened 2020-10-15
test: disable failing test "should be able to remove headers"

JoelEinbinder:
I'm not sure if the test or the behavior is wrong, but this is failing consistently on all of the bots/locally. @yury-s you touched it last, ptal.

### DIFF
--- a/test/page-route.spec.ts
+++ b/test/page-route.spec.ts
@@ -96,7 +96,10 @@ it('should work when header manipulation headers with redirect', async ({page, s
   await page.goto(server.PREFIX + '/rrredirect');
 });
 // @see https://github.com/GoogleChrome/puppeteer/issues/4743
-it('should be able to remove headers', async ({page, server}) => {
+it('should be able to remove headers', (test, {browserName}) => {
+  // On WebKit the accept header gets set to "*/*"
+  test.fail(browserName === 'webkit');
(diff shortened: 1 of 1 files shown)


JoelEinbinder:
Or maybe this is fine, and the test should be rewritten like:
```ts
// @see https://github.com/GoogleChrome/puppeteer/issues/4743
it('should be able to remove headers', async ({page, server, browserName}) => {
  await page.route('**/*', route => {
    const headers = Object.assign({}, route.request().headers(), {
      foo: 'bar',
      accept: undefined, // remove "accept" header
    });
    route.continue({ headers });
  });

  const [serverRequest] = await Promise.all([
    server.waitForRequest('/empty.html'),
    page.goto(server.PREFIX + '/empty.html')
  ]);

  expect(serverRequest.headers.accept).toBe(browserName === 'webkit' ? '*/*' : undefined);
});
```

It looks like `*/*` should be equivalent to an empty accept header, but I'm not familiar with how this all works.

mxschmitt:
Maybe its better to use e.g. Fetch there and use a custom request header instead of accept which has edge cases in WebKit.

yury-s (approved):
Seems like a bug in WebKit. As for the test I'll rewrite it using fetch as Max suggests.

PR #4165, opened 2020-10-16
fix(screencast): do not stop screencast on page close

pavelfeldman:


src/server/chromium/crPage.ts
```diff
@@ -197,6 +197,7 @@ export class CRPage implements PageDelegate {
   }
 
   async closePage(runBeforeUnload: boolean): Promise<void> {
+    await this._mainFrameSession._stopScreencast();
```
yury-s, line 200: Call this from _didClose as it doesn't send any messages to the browser?

yury-s (approved):

src/server/chromium/crPage.ts
pavelfeldman: It is already so

PR #4171, opened 2020-10-17
docs: Use green check marks in browser compatibility matrix

ElijahLynn:
Because green check marks look so much better than white check marks, and is better for accessibility!


pavelfeldman:
I believe that's the opposite on non-Mac. We picked it so that it looks good on all platforms (and browsers)!

PR #4202, opened 2020-10-21
doc: update playwright-sharp link

kblok:
I don't know who is that hardkoded. But that's wrong.


mxschmitt (approved):
Can you update it in `docs/languages.md` too?

PR #4207, opened 2020-10-21
fix(headers): report raw request headers on Chromium

pavelfeldman:


src/protocol/protocol.yml
```diff
@@ -1798,6 +1798,13 @@ Response:
     url: string
     status: number
     statusText: string
+    requestHeaders:
```
aslushnikov, line 1801: Not: `rawheaders` might be a more descriptive name

aslushnikov (approved):

src/protocol/protocol.yml
pavelfeldman: These are request headers on response object!

Issue #4225, opened 2020-10-23
[BUG] Cannot override `.npmrc` env var `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` from true to false


majapw:
**Context:**
- Playwright Version: 1.5.1
- Operating System: Linux for CI, Mac locally
- Node.js version: 12.16.2
- Browser: All

**Describe the bug**

The scenario we have is that we want to skip installing browsers by default in our ecosystem, but manually install browsers if a person is running playwright code locally or as part of the CI job that runs playwright code. As a result we set the following in our `.npmrc`:
```
playwright_skip_browser_download=1
```

Unfortunately, `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD="" yarn` does not override this value (or any variation of this). I *think* this is because the concept of overriding env vars in the playwright repo is predicated on the idea of overriding falsey values with truthy values and not the other way around.

If we look at [the code to skip the browser install](https://github.com/microsoft/playwright/blob/master/src/install/installer.ts#L36-L39):
```js
  if (getFromENV('PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD')) {
    browserFetcher.logPolitely('Skipping browsers download because `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` env variable is set');
    return false;
  }
``
We need `getFromENV('PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD')` to return a falsey value to get past this block. When we look at the code from [`getFromEnv`](https://github.com/microsoft/playwright/blob/master/src/utils/utils.ts#L98-L103):
```
export function getFromENV(name: string) {
  let value = process.env[name];
  value = value || process.env[`npm_config_${name.toLowerCase()}`];
  value = value || process.env[`npm_package_config_${name.toLowerCase()}`];
  return value;
}
```
we see that it's impossible to override the truthy value of `process.env.npm_config_playwright_skip_browser_download` with a falsey value no matter what we do. I think changes the `||` operators to `??` might fix the issue.

Additionally, it might be worth being explicit about possible values for `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD`, cast them from a string to a number (`"0"` is true after all, while `0` is not) and do a !! in the check. 

I'll put up a corresponding PR for your review.

majapw:
An alternative would be to expose a script that allows for the end-user to manually install the browser binaries whenever they want. I believe `Cypress` has this option.

[...]

PR #4226, opened 2020-10-23
Update `getFromEnv` logic to validate that value is undefined (instead of falsey) before redefining it

majapw:
Additionally, cast the `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` check to a number to allow for values of 0 or 1

Fix for https://github.com/microsoft/playwright/issues/4225

### DIFF
--- a/src/install/installer.ts
+++ b/src/install/installer.ts
@@ -33,7 +33,8 @@ const fsWriteFileAsync = util.promisify(fs.writeFile.bind(fs));
 const removeFolderAsync = util.promisify(removeFolder);
 
 export async function installBrowsersWithProgressBar(packagePath: string) {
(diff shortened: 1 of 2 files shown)


pavelfeldman:
Ouch, I did not see your change as I crafted mine. Let me actually cancel mine!

majapw:
No worries! Are the test failures expected flake? Also happy to make this a bit more robust. I was expecting only `0`/`1` values for the env variable

pavelfeldman (approved):

pavelfeldman:
> Are the test failures expected flake?

We have a broken screencast test on the tip of the tree, should cycle green soon-ish.

> I was expecting only 0/1 values for the env variable

That's Ok for now. I might follow up with respecting `=true` / `=false`

PR #4227, opened 2020-10-23
fix(env): allow overriding env with falsy value

### LINKED ISSUE #4225: [BUG] Cannot override `.npmrc` env var `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` from true to false

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/4225

### DIFF
--- a/src/install/installer.ts
+++ b/src/install/installer.ts
@@ -22,7 +22,7 @@ import * as removeFolder from 'rimraf';
 import * as lockfile from 'proper-lockfile';
 import * as browserPaths from '../utils/browserPaths';
 import * as browserFetcher from './browserFetcher';
-import { getFromENV } from '../utils/utils';
+import { getAsBooleanFromENV } from '../utils/utils';
 
 const fsMkdirAsync = util.promisify(fs.mkdir.bind(fs));
 const fsReaddirAsync = util.promisify(fs.readdir.bind(fs));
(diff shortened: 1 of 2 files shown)


JoelEinbinder (approved):

pavelfeldman:
Closing in favor of https://github.com/microsoft/playwright/pull/4226

PR #4228, opened 2020-10-23
fix(env): respect =true/false as env values for boolean flags

pavelfeldman:


pavelfeldman:
@majapw: merging bits of mine on tip of the tree!

JoelEinbinder (approved):

PR #4230, opened 2020-10-24
chore: add newest node version

rwoll:


mxschmitt:
Looks like we have to call npm prepare manually now: https://github.com/npm/cli/issues/1905

PR #4245, opened 2020-10-26
browser(webkit): fix mac build after last roll

yury-s:
https://github.com/yury-s/webkit/commit/1d04b36fe4905516ed52861cf8cf29fffd535b94

* Those .idls were recently removed from DerivedSources.make (https://trac.webkit.org/changeset/268849/webkit)

### DIFF
--- a/browser_patches/webkit/BUILD_NUMBER
+++ b/browser_patches/webkit/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1366
-Changed: pavel.feldman@gmail.com Mon Oct 26 16:07:00 PDT 2020
+1367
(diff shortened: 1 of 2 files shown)


yury-s:
@JoelEinbinder upstream WebKit is removing references to the touch related .idls as apparently they are used only in the builds that depend on apple internal sdk. Do you have a clue on how we could make it less brittle with every roll?

pavelfeldman (approved):

JoelEinbinder:
I'm not sure how to make it less brittle. We are kind of living in a bad place using the touch code here. It's mostly influenced by the internal apple code.

JoelEinbinder (approved):

PR #3642, opened 2020-08-26
feat: upgrade node to 14.x in Docker image

cbxp:
Node 14 supports many new language features, like optional chaining, that are really convenient to use in tests

PR #4262, opened 2020-10-27
chore: use Node.js 14 (new LTS) in Docker image

### LINKED ISSUE #4254: [Question] Will node js be updated to 14 lts in docker image?
Since node js 14 lts is active now can we have 14 version in docker image? I use some features of 14 node and would like to use official image of docker playwright in CI with 14 version of node js. Can you consider updating it?

mxschmitt:
Discussion / background: https://github.com/microsoft/playwright/pull/3642#issuecomment-686640755

Closes #4254

### DIFF
--- a/docs/docker/Dockerfile.bionic
+++ b/docs/docker/Dockerfile.bionic
@@ -1,8 +1,8 @@
 FROM ubuntu:bionic
 
-# 1. Install node12
+# 1. Install node14
 RUN apt-get update && apt-get install -y curl && \
-    curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
+    curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
     apt-get install -y nodejs
 
 # 2. Install WebKit dependencies
--- a/docs/docker/Dockerfile.focal
+++ b/docs/docker/Dockerfile.focal
@@ -1,8 +1,8 @@
 FROM ubuntu:focal
 
-# 1. Install node12
+# 1. Install node14
 RUN apt-get update && apt-get install -y curl && \
-    curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
+    curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
     apt-get install -y nodejs
 
 # 2. Install WebKit dependencies


pavelfeldman:
What about those that use Node 12? Maybe we could have a side-by-side image for early Node 14 LTS adopters?

mxschmitt:
> What about those that use Node 12? Maybe we could have a side-by-side image for early Node 14 LTS adopters?

They can use our Docker image as a base image and install their custom version on top of it (we can document that in the docs maybe). We should always use the LTS version in my mind, I think lushikov had the same opinion their back then in our sync.

niyarlatotep:
Any update on this?

arjunattam:
Thanks for the follow-up, @niyarlatotep. On looking at [Node.js release calendar](https://nodejs.org/en/about/releases/), we plan to replace Node.js 12 with 14 in our base image when active LTS for 12 ends: November 30, 2020. This will align with the timeline for v1.7. For now, I would recommend installing Node.js 14 in a user-land Dockerfile.

```
FROM mcr.microsoft.com/playwright:v1.5.2

RUN apt-get update && apt-get install -y curl && \
    curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
    apt-get install -y nodejs
```

aslushnikov (approved):

mxschmitt:
> > They can use our Docker image as a base image and install their custom version on top of it (we can document that in the docs maybe).
> 
> Yes please. I'm in need of using a specific version of node but haven't had success in upgrading to it via GitLab CI:
> 
> ```
> {
> 	"image": "playwright/chromium:latest",
> 	"before_script": [
> 		"apt-get update",
> 		"apt-get install -y curl",
> 		"curl -fsSL https://deb.nodesource.com/setup_18.x | bash -",
> 		"apt-get install -y nodejs",
> 		"npm install"
> 	],
> ```
> 
> I get what looks like a permissions error on `apt-get update` because of the privileges of the user the image is configured to run as.
> 
> If there's a workaround until Node 18 is proper LTS, I'd be open to that too.

These are not our official images, see here for their support: https://aerokube.com/images/latest/

PR #4290, opened 2020-10-30
browser(firefox): don't send messages to content too quickly

JoelEinbinder:
Sending messages too quickly to the content process could have them reach there before the channel is set up. Then they would be silently discarded.


aslushnikov:
Closing in favor of https://github.com/microsoft/playwright/pull/4311

PR #4291, opened 2020-10-30
fix(firefox): wait for the utility world to be created

JoelEinbinder:
This patch actually reveals a different race than the one in the comment. Sometimes `Page.addScriptToEvaluateOnNewDocument` never resolves. I have a browser patch to fix it at #4290.

### DIFF
--- a/src/server/firefox/ffPage.ts
+++ b/src/server/firefox/ffPage.ts
@@ -44,7 +44,7 @@ export class FFPage implements PageDelegate {
   readonly _networkManager: FFNetworkManager;
   readonly _browserContext: FFBrowserContext;
   private _pagePromise: Promise<Page | Error>;
-  _pageCallback: (pageOrError: Page | Error) => void = () => {};
+  _pageCallback!: (pageOrError: Page|Error) => void;
   _initializedPage: Page | null = null;
   readonly _opener: FFPage | null;
   private readonly _contextIdToContext: Map<string, dom.FrameExecutionContext>;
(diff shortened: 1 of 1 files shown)


dgozman (commented):
I do not understand what's happening here :(

JoelEinbinder:
There was something strange with types.d.ts. Fixed that.

In our other browsers we have large async `initialize` methods in Page. In firefox we only have one line of initialization, `Page.addScriptToEvaluateOnNewDocument`. This patch makes sure that call is resolved before we consider the page initialized. The `pageOrError` construct makes it a bit awkward, but I decided against refactoring that in this patch for fear of breaking something.

Consider what happens if `Page.addScriptToEvaluateOnNewDocument` is slow, and also throws an error. Before this patch, the page would be returned on `Page.ready`, and the error of `Page.addScriptToEvaluateOnNewDocument` would be ignored. Then when someone tried to use the utility context, it would exhibit undefined behavior.

With this patch, `browserContext.newPage()` will throw the error from `Page.addScriptToEvaluateOnNewDocument`

src/server/firefox/ffPage.ts
```diff
@@ -90,15 +90,17 @@ export class FFPage implements PageDelegate {
...
+    this._pagePromise = Promise.all([
+      new Promise<Page|Error>(f => this._pageCallback = f),
+      // Ideally, we somehow ensure that utility world is created before Page.ready arrives, but currently it is racy.
+      // Therefore, we can end up with an initialized page without utility world, although very unlikely.
+      this._session.send('Page.addScriptToEvaluateOnNewDocument', { script: '', worldName: UTILITY_WORLD_NAME }),
+    ]).then(x => x[0]).catch(error => error);
     session.once(FFSessionEvents.Disconnected, () => this._page._didDisconnect());
     this._session.once('Page.ready', () => {
       this._pageCallback(this._page);
       this._initializedPage = this._page;
```
dgozman, line 102: I _think_ that `this._initializedPage` should only be present when `_pagePromise` is resolved.

src/server/firefox/ffPage.ts
```diff
@@ -90,15 +90,17 @@ export class FFPage implements PageDelegate {
       helper.addEventListener(this._session, 'Page.webSocketFrameReceived', this._onWebSocketFrameReceived.bind(this)),
       helper.addEventListener(this._session, 'Page.webSocketFrameSent', this._onWebSocketFrameSent.bind(this)),
     ];
-    this._pagePromise = new Promise(f => this._pageCallback = f);
+    this._pagePromise = Promise.all([
```
dgozman, line 93: This seems to introduce a race, where we will potentially miss some events (e.g. main network request in a popup), because we do not report the page until `Page.addScriptToEvaluateOnNewDocument` resolves, and we also do not pause the page.

Note that in other browsers this works because we manually resume the page after initialization.

dgozman (commented):
Thanks for the explanation!

I wonder whether the fix should be on the backend side? For example, perform `addScriptToEvaluateOnNewDocument` on the browser context, so that it is initialized in the browser before `Page.ready`.

src/server/firefox/ffPage.ts
JoelEinbinder: I checked this scenario when I was writing the patch, but things are hairy here so I might have missed something. Playwright internally operates on an uninitialized page. But I think you are right that the user might miss some events.

JoelEinbinder:
I don't think this patch fixes any pressing bugs. It just makes some error messages nicer and is sort of a test of #4290. It seems like it might cause more races, so I'll close it.

PR #4292, opened 2020-10-30
fix(firefox): don't crash when messages come after close

JoelEinbinder:
In [this test run](https://github.com/microsoft/playwright/runs/1323672300), we get an unexpected protocol response. This can only happen if the session receives a message after it has been closed. There is a comment in ffConnection.ts that implies as much:
```typescript
      const callback = this._callbacks.get(message.id);
      // Callbacks could be all rejected if someone has called `.dispose()`.
      if (callback) {
```
I haven't been able to repro it locally, but I am confident that this patch is neutral or beneficial.

### DIFF
--- a/src/server/firefox/ffConnection.ts
+++ b/src/server/firefox/ffConnection.ts
@@ -188,6 +188,8 @@ export class FFSession extends EventEmitter {
   }
 
   dispatchMessage(object: ProtocolResponse) {
+    if (this._disposed)
+      return;
     if (object.id && this._callbacks.has(object.id)) {
       const callback = this._callbacks.get(object.id)!;
       this._callbacks.delete(object.id);


JoelEinbinder:
Our other connections might have the same race, but I'll wait until we see a crash before mucking about there.

dgozman (commented):
By inspecting the code I see that:
- We never dispose the session from `_onDetachedFromTarget`, this is an unrelated bug 😄 
- We only dispose the session from `FFConnection._onClose`, so we should not get any messages after that. If we do, let's fix the transport instead.
- There is also a usage of `FFSession` for workers. Perhaps we might be getting messages from the worker after `workerDestroyed`? Seems like something to be fixed in Juggler. This should also not be relevant to the linked bot failure, because it does not have workers in the test.

JoelEinbinder:
> By inspecting the code I see that:
> 
> * We never dispose the session from `_onDetachedFromTarget`, this is an unrelated bug :smile:

`_onDetachedFromTarget` calls `ffPage.didClose()` which calls `this._session.dispose();`

> * We only dispose the session from `FFConnection._onClose`, so we should not get any messages after that. If we do, let's fix the transport instead.

We dispose the session from `_onDetachedFromTarget`, so we can still receive messages.

`Browser.detachedFromTarget` is sent from the browser process in juggler. It is triggered when the browser is notified that a tab closes or crashes. I believe this happens on a different channel than the one that sends our protocol messages between the browser and renderer. So we can still receive some delayed messages, or more likely errors, from a target that looks to be closed on the browser side. We can fix this in juggler by doing the same tracking of sessions that appear to be closed and swallowing messages from them. But I'd rather fix it this way considering we already have this logic for the connection level.

dgozman:
> > By inspecting the code I see that:
> > 
> > * We never dispose the session from `_onDetachedFromTarget`, this is an unrelated bug 😄
> 
> `_onDetachedFromTarget` calls `ffPage.didClose()` which calls `this._session.dispose();`

I missed that!

> 
> > * We only dispose the session from `FFConnection._onClose`, so we should not get any messages after that. If we do, let's fix the transport instead.
> 
> We dispose the session from `_onDetachedFromTarget`, so we can still receive messages.

I still don't understand how this fix helps. In `FFSession.dispose` we remove the session from `FFConnection._sessions` map. Before calling `FFSession.dispatchMessage`, we first retrieve it from the map. So, we should not be able to dispatch a message on the disposed session.

JoelEinbinder:
I forgot what this does

PR #4300, opened 2020-10-31
feat(scopes): make page a scope

pavelfeldman:
Although we end up with a structure where Frame is on the same level as the page, it is very convenient to attribute console messages, dialogs, web sockets to the page rather than the browser context. We get a lot of `page()` methods for free on those.

### DIFF
--- a/src/dispatchers/dispatcher.ts
+++ b/src/dispatchers/dispatcher.ts
@@ -73,7 +73,7 @@ export class Dispatcher<Type, Initializer> extends EventEmitter implements chann
 
     (object as any)[dispatcherSymbol] = this;
     if (this._parent)
-      this._connection.sendMessageToClient(this._parent._guid, '__create__', { type, initializer, guid }, !!isScope);
+      this._connection.sendMessageToClient(this._parent._guid, '__create__', { type, initializer, guid });
   }
 
   _dispatchEvent(method: string, params: Dispatcher<any, any> | any = {}) {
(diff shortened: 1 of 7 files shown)


src/dispatchers/dispatcher.ts
```diff
@@ -178,26 +177,23 @@ export class DispatcherConnection {
...
     }
   }
 
-  private _replaceDispatchersWithGuids(payload: any, allowDispatchers: boolean): any {
+  private _replaceDispatchersWithGuids(payload: any): any {
     if (!payload)
       return payload;
-    if (payload instanceof Dispatcher) {
-      if (!allowDispatchers)
-        throw new Error(`Channels are not allowed in the scope's initialzier`);
```
dgozman, line 192: This particular check was a result of the following problem:

BrowserContext had `pages` in the initializer, and since we send channels from the initializer before the object, these pages were created in the parent scope (the Browser), rather than BrowserContext's scope. This lead to leaks where pages were not cleaned up upon BrowserContext's dispose. I am afraid this could happen again.

dgozman (commented):
FWIW, it seems like the PR is actually about a convenient `parent` for each object, not about the scopes. What if we actually send whatever parent we want, but only dispose the BrowserContext and Browser explicitly?

I see three issues with this particular implementation:
- the "no channels in initializers" check (see the separate comment);
- all frames will outlive the page and won't be disposed - that means we don't actually clean up any resources on page close;
- methods like `Download.path` will return a wrong error `Target browser or context has been closed` when in fact it was the page that was closed, while the context is still alive.

pavelfeldman:
> that means we don't actually clean up any resources on page close

My understanding is that in this case we clean up:
- console messages
- dialogs
- injected bindings
- requests with headers and cached content
- responses, same
- web sockets
- downloads (probably need a parent scope as per contract)
- videos (same?)

src/dispatchers/dispatcher.ts
pavelfeldman: This looks like an edge case, let's add a test instead?

pavelfeldman:
> all frames will outlive the page and won't be disposed

That's what happens today as well, so it isn't changing. Stale pointers already have undefined behavior anyways.

dgozman (approved):

PR #4332, opened 2020-11-04
browser(webkit): do not hang on close when there is a dialog

yury-s:


browser_patches/prepare_checkout.sh
```diff
@@ -124,11 +124,11 @@ git apply --index --whitespace=nowarn $PATCHES_PATH/*
 
 if [[ ! -z "${WEBKIT_EXTRA_FOLDER_PATH}" ]]; then
   echo "-- adding WebKit embedders"
-  cp -r "${WEBKIT_EXTRA_FOLDER_PATH}" ./Tools/Playwright
+  cp -r "${WEBKIT_EXTRA_FOLDER_PATH}/" ./Tools/Playwright
```
aslushnikov, line 127: Let's not land this ATM - trailing slash in the "SRC" is an anti-pattern. I'd like to understand first why would it be needed.

aslushnikov (approved):

PR #4336, opened 2020-11-04
api(websocket): require captureWebSocketFrames for framesent/framereceived

dgozman:
These events incur significant overhead, so we make them opt-in.


yury-s (commented):
Can we instead send the events from server to client only when client has listeners for websockets? We anyway always receive the events from the browser so there is already non-trivial overhead or you planning to disable them in the browsers too?

dgozman:
> Can we instead send the events from server to client only when client has listeners for websockets? We anyway always receive the events from the browser so there is already non-trivial overhead or you planning to disable them in the browsers too?

I plan to address this in browsers once it's a problem. "only when client has listeners for websockets" has a race between adding a listener and websocket already receiving messages in the page.

yury-s:
> I plan to address this in browsers once it's a problem.

I wonder why it is a problem in client-server link but not in browser-server today?

> "only when client has listeners for websockets" has a race between adding a listener and websocket already receiving messages in the page.

If the listener is added to the page/context before navigation there should be no race, otherwise it is no worse than what we have now (no guarantees when the listener is added after navigation). What am I missing?

PR #4342, opened 2020-11-04
feat(fill): allow filling based on the label selector

### LINKED ISSUE #3466: [Feature] Fill input field based on label text
When an input field is properly connected to a label it would be nice to be able to fill the input field based on the label text using `await page.fill('text=Name', 'John Doe');`.

It's often possible to fill the input field based on a CSS selector like `id` or `name`, but making it possible to fill by label text would make it more straight-forward to both read and write the test. Especially if you write tests for screens you did not make yourself.

Examples of cases where it should work:
```
<label for="input1">Name</label>
<input id="input1" value="">
```

```
<label>Name <input value=""></label>

```

dgozman:
This enables filling the input based on the connected label:

```html
<label for=target>Name</label><input id=target>
```

```js
await page.fill('text=Name', 'Alice');
```

Fixes #3466.

### DIFF
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -261,10 +261,10 @@ export class InjectedScript {
     return this.pollRaf((progress, continuePolling) => {
       if (node.nodeType !== Node.ELEMENT_NODE)
         return 'error:notelement';
-      const element = node as Element;
-      if (!element.isConnected)
+      const element = this.findLabelTarget(node as Element);
+      if (element && !element.isConnected)
         return 'error:notconnected';
-      if (!this.isVisible(element)) {
+      if (!element || !this.isVisible(element)) {
         progress.logRepeating('    element is not visible - waiting...');
         return continuePolling;
       }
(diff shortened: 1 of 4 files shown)


src/server/injected/injectedScript.ts
```diff
@@ -438,27 +438,28 @@ export class InjectedScript {
     return 'done';
   }
 
+  findLabelTarget(element: Element): Element | undefined {
+    if (element.nodeName === 'LABEL') {
+      const forId = element.getAttribute('for');
+      if (forId && element.ownerDocument)
+        return element.ownerDocument.querySelector(`[id="${forId}"]`) || undefined;
```
JoelEinbinder, line 445:
```suggestion
        return element.ownerDocument.querySelector(`[id="${CSS.escape(forId)}"]`) || undefined;
```

src/server/injected/injectedScript.ts
```diff
@@ -438,27 +438,28 @@ export class InjectedScript {
...
+      return element.querySelector('input,select,textarea,meter,progress') || undefined;
+    }
+    return element;
```
JoelEinbinder, line 448:
```suggestion
    return element.nodeName === 'LABEL' ? (element as HTMLLabelElement).control || undefined : element;
```

src/server/injected/injectedScript.ts
```diff
@@ -261,10 +261,10 @@ export class InjectedScript {
     return this.pollRaf((progress, continuePolling) => {
       if (node.nodeType !== Node.ELEMENT_NODE)
         return 'error:notelement';
-      const element = node as Element;
-      if (!element.isConnected)
+      const element = this.findLabelTarget(node as Element);
+      if (element && !element.isConnected)
         return 'error:notconnected';
-      if (!this.isVisible(element)) {
+      if (!element || !this.isVisible(element)) {
```
JoelEinbinder, line 267: This means I can't click a label if it is missing a `for`? But in that case shouldn't we just click the label itself.

src/server/injected/injectedScript.ts
dgozman: Wonderful! Done.

JoelEinbinder (approved):

mxschmitt:
Lets maybe add it to the docs if it was not possible before to fill by label text?

dgozman:
> Lets maybe add it to the docs if it was not possible before to fill by label text?

Good idea, done.

PR #4344, opened 2020-11-04
feat: add support for using existing cdp websocket in launchServer

### LINKED ISSUE #4054: [REGRESSION]: chromium.connect does not work with vanilla CDP servers anymore
Hi there,
apologies if this has been asked before but I didn't find anything.

Up until `v1.3.0` it was possible to `.connect` to a vanilla browser websocket URL (either created through launching Chrome with `--remote-debugging`) or using Puppeteer.

`v1.4.0` introduced the new "New Client / Server Wire protocol", since then it's not possible anymore to use `.connect` with other original CDP websocket servers anymore (it will hang [here)](https://github.com/microsoft/playwright/blob/e214f795e014028a6059709c01a697c8d49d7689/src/client/browserType.ts#L153). I also noticed through logging that the new wire protocol looks much different from the regular CDP chatter.

My question would be if there's still a way to use `chromium.connect` with the traditional CDP websocket URLs or if there are pointers of how to best write a translation layer to accomplish that.

My use-case is to have a single websocket url which exposes a fleet of chromium browsers by proxying the websocket connections (similar to [browserless.io](https://browserless.io)) and both puppeteer & playwright being able to `connect` to it and control the browsers.

The following worked up until `v1.4.0`

```js
const puppeteer = require("puppeteer")

puppeteer
  .launch({
    headless: true,
    defaultViewport: null,
    args: [
      "--remote-debugging-port=9222",
      "--remote-debugging-address=0.0.0.0",
    ],
  })
  .then(async (browser) => {
    console.log(browser._connection.url())
  })
// => ws://0.0.0.0:9222/devtools/browser/490a2b32-ce62-4b6b-a530-d4931fdeb046
```
```js
const pw = require("playwright")

;(async () => {
  const browser = await pw.chromium.connect({
    wsEndpoint: "ws://0.0.0.0:9222/devtools/browser/490a2b32-ce62-4b6b-a530-d4931fdeb046",
  })

  const page = await context.newPage()

  await page.goto("https://www.example.com/")
  await page.screenshot({ path: "example.png" })

  await browser.close()
})()
```

berstend:
This PR:
- adds a `cdpWebsocketEndpoint` option to `.launchServer`
- adds a Playwright user-agent header to websocket connections (`.connect` & `.launchServer`)
- fixes #4054

### CDP

As a side effect of switching to the new wire protocol we lost the ability to connect to existing CDP websockets. This is an issue for services that wish to make existing browsers (not created by Playwright) available to Playwright clients.

### User-Agent

Adding a user-agent makes it much easier for services that wish to expose a single websocket to both Playwright and Puppeteer clients. In combination with `cdpWebsocketEndpoint` Playwright clients can then be piped through a wire protocol websocket created by `launchServer` that in turn will communicate with an existing CDP websocket endpoint.

```
                   +------------+                             +---------------+
pptr client  +-->  |            |  +----------------------->  |               |
                   |  custom    |                             |  Chromium CDP |
                   |  websocket |        playwright           |  websocket    |
  pw client  +-->  |            |  +-->  wire protocol  +-->  |               |
                   +------------+        websocket            +---------------+

```

I tried following the available naming pattern & style guide, kindly let me know if there are more preferred options. :-)


berstend:
Urgh, seems like optional chaining is not supported in CI

test/browsertype-launch-server.spec.ts
```diff
@@ -17,7 +17,7 @@
 
 import { it, expect, describe } from './fixtures';
 
-describe('lauch server', (suite, { wire }) => {
+describe('launch server', (suite, { wire }) => {
```
JoelEinbinder, line 20: :smile:

berstend: Thought I fix that while I'm there 😄

test/browsertype-launch-server.spec.ts
```diff
@@ -62,4 +62,17 @@ describe('lauch server', (suite, { wire }) => {
     expect(result['exitCode']).toBe(0);
     expect(result['signal']).toBe(null);
   });
+
+  it('should allow using an existing cdp endpoint', async ({ browserType, server}) => {
```
JoelEinbinder, line 66: This test don't really prove that the cdp websocket works. @dgozman should we add another bot that runs all the tests via cdp?

Or we can just have a smoke test that.
1. launches a browser server with `--remote-debugging-port`
2. launches a second browser server that connects to the first via cdpWebSocketEndpoint
3. connect to that second browser server and create a page and eval 1+1=2 in there.

berstend:
I think mapping `upgrade` to `onRequest` in the testserver has side-effects, I made a dedicated `onUpgrade` handler

test/browsertype-launch-server.spec.ts
berstend: Yeah, doing that was my first instinct :-) I had the impression the other tests were a bit shallow so I opted to just test the very basics. Happy to change this to a more full-fledged test.

berstend: Just wondering about the best way to get the `webSocketDebuggerUrl` from `/json/version` (there doesn't seem to be an http client around)

src/client/browserType.ts
```diff
@@ -32,13 +32,14 @@ import { assert, makeWaitForNextTask, mkdirIfNeeded } from '../utils/utils';
...
 import { Stream } from './stream';
+const packageVersion = require('../../package.json').version;
 
 export interface BrowserServerLauncher {
   launchServer(options?: LaunchServerOptions): Promise<BrowserServer>;
 }
 
 export interface BrowserServer {
-  process(): ChildProcess;
+  process(): ChildProcess | undefined;
```
dgozman, line 42: Please do not change this signature, it's not backwards-compatible. Instead, throw an error from `process()` when running in "connect over cdp" mode.

src/browserServerImpl.ts
```diff
@@ -93,11 +96,13 @@ export class BrowserServerImpl extends EventEmitter implements BrowserServer {
   }
 
   async close(): Promise<void> {
-    await this._browser._options.browserProcess.close();
+    if (this._browser._options.browserProcess)
```
dgozman, line 99: I think the right way is to connect right here, in the BrowserServerImpl, instead of going all the way to _innerLaunch and back.

Now that I think about this more: when connecting over existing cdp, all other launch parameters are actually ignored. This hints at using a different method, something like `chromium.connectServer` that only takes the websocket url. WDYT?

test/browsertype-launch-server.spec.ts
dgozman: You can do `await page.evaluate(() => fetch(url))` and use the browser 😄

berstend: went with vanilla `http` 😄 

image

src/browserServerImpl.ts
berstend: I agree, when connecting to an existing cdp websocket virtually all of the launch code isn't needed. Also the flag feels a bit shoehorned into there, so it might be cleaner to keep it separate. If promoting this feature to it's own top-level API is fine then I'm happy to work on the changes 😄

berstend:
Thanks for all the valuable input and suggestions, will get back at it tomorrow. :-)

test/browsertype-launch-server.spec.ts
berstend: @JoelEinbinder added an additional smoke test to ensure the cdp websocket works :-)

berstend:
@dgozman I added a dedicated `browserType.connectServer` API as well, so we can see if we like it better:
image

Personally I think it's the better approach than cramping this into the existing `.launchServer` 😄 

Note: I placed the meaty stuff in `server/browserType.ts` as I'd otherwise would need to re-import a bunch of things in `src/browserServerImpl.ts` :-)

src/browserServerImpl.ts
berstend: Added a dedicated API implementation for comparison

dgozman (commented):
I definitely like the separate method more! Please give us some time to discuss the shape of this new API, and I'll get back to you with detailed PR review 😄

mxschmitt:
@berstend seems like it was implemented in #5207

pavelfeldman:
This is now obsolete and we have connectOverCDP, so closing!

PR #4352, opened 2020-11-05
devops: speedup initial browser checkout

aslushnikov:
Instead of checking out the whole repository, we now do a shallow
clone.

We then gradually "unshallow" the clone, looking for the `BASE_REVISION`.

This should fix experimental mac-11 builder.


aslushnikov:
FYI @yury-s

PR #4353, opened 2020-11-05
devops: checkout, build, and upload webkit-mac-11

aslushnikov:


.github/workflows/build_webkit.yml
```diff
@@ -1,17 +1,20 @@
-name: tests
+name: browser builds
 
 on:
   push:
```
mxschmitt, line 4: pull_request would also be helpful, so you don't need to merge always for testing. And upload it to the CDN only if its master.

PR #4354, opened 2020-11-05
devops: actually export secrets into webkit building workflow

aslushnikov:


.github/workflows/build_webkit.yml
```diff
@@ -5,6 +5,11 @@ on:
     branches:
       - master
 
+env:
```
mxschmitt, line 8: It's better to move them on the jobs or even better step level. So if a GitHub Action gets compromised, its only accessible in the specific Step.

Issue #1032, opened 2020-02-15
[BUG] Fetching webgl2 canvas context does not work in Firefox headless mode


Tyriar:
**Context:**
- Playwright Version: 0.10.0
- Operating System: Mac

**Code Snippet**

Running within a page:

```javascript
// Should be non-null on Firefox
canvas.getContext('webgl2');
```

**Describe the bug**

Fetching the `webgl2` context of a canvas returns null when Firefox is running in headless mode but not when running with the window displayed. This prevents me from running the webgl renderer tests of [xterm.js](https://github.com/xtermjs/xterm.js) on firefox.

Note that it happens always on webkit [but that's expected](https://caniuse.com/#search=webgl2).

pavelfeldman:
I'll close this as a part of the bug triaging process. We have hundreds of bugs and feature requests with dozens and even hundreds of upvotes, while this one only has a few thumbs up. Please feel free to open a new bug and link this one if you'd like to see it addressed.

PR #4358, opened 2020-11-05
browser(firefox): enable webgl on headless

JoelEinbinder:
This patch just comments out the code that disables webgl on headless. It seems to work?! The [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1375585) referenced says that it might cause a crash when running without a display. I haven't been able to repro that. There is also a [patch in review](https://phabricator.services.mozilla.com/D88945) that we maybe should just wait for.

#1032


dgozman:
Let's wait a bit for the upstream patch? 😄

JoelEinbinder:
> Let's wait a bit for the upstream patch? 😄

Definitely. Just leaving this as a draft so that we remember one way or another.

PR #4367, opened 2020-11-06
fix(launcher): print default list of required packages on failure

yury-s:


src/server/validateDependencies.ts
```diff
@@ -432,3 +436,89 @@ const MANUAL_LIBRARY_TO_PACKAGE_NAME_UBUNTU: { [s: string]: string} = {
   // gstreamer1.0-libav -> libavcodec57 -> libx264-152
   'libx264.so': 'gstreamer1.0-libav',
 };
+
+
+// This map should be in sync with the Dockerfiles.
+const defaultRequirements: { [s: string]: { [s: string]: string[]} } = {
```
aslushnikov, line 442: I believe this should not be needed since these all are inherted by the launch doctor.

yury-s: Reverted.

PR #4378, opened 2020-11-07
[wip] feat(selectors): support {index} syntax

### LINKED ISSUE #3697: [Feature] allow selecting the second (k-th) selector match.
I'd like to select second 'text="John"' in the list.

dgozman:
Supports selectors like `button {2}` to match the second button.

Note that `{}` syntax is chosen to not conflict with xpath's own `[]` syntax, so that we can still avoid escaping and continue to auto-detect.

Fixes #3697.

### DIFF
--- a/docs/selectors.md
+++ b/docs/selectors.md
@@ -10,7 +10,7 @@ Selectors query elements on the web page for interactions, like [page.click](api
 <!-- GEN:stop -->
 
 ## Syntax
-Selectors are defined by selector engine name and selector body, `engine=body`. 
+Selectors are defined by selector engine name and selector body, `engine=body`.
 
 * `engine` refers to one of the [supported engines](#selector-engines)
   * Built-in selector engines: [css], [text], [xpath] and [id selectors][id]
@@ -46,6 +46,11 @@ document
 ```
 
 If a selector needs to include `>>` in the body, it should be escaped inside a string to not be confused with chaining separator, e.g. `text="some >> text"`.
+
+### Indexed matches
+By default, selectors resolve to the first matching element, according to the DOM traversal order. A selector can be followed by `{n}` to match the n-th element instead.
+
+For example, `text=list item {3}` matches the third element with the text `list item`. Similarly, `article {2} >> .show-button` matches the `.show-button` element in the second `article`.
 
 ### Intermediate matches
 By default, chained selectors resolve to an element queried by the last selector. A selector can be prefixed with `*` to capture elements that are queried by an intermediate selector.
--- a/docs/core-concepts.md
+++ b/docs/core-concepts.md
@@ -161,6 +161,11 @@ await page.click('text=Hello w');
 ```
 
 ```js
+// Click the third button
+await page.click('button {3}');
+```
+
+```js
 // Explicit CSS and XPath notation
 await page.click('css=div');
 await page.click('xpath=//html/body/div');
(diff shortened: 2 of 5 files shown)


mxschmitt:
I remember the case when you have multiple matches and it resolved to multiple elements and the first one did no fulfil the Actionability criteria. (Since by default it uses the first match). Would it make sense to display some information about the other matched elements in the TimeoutError error if multiple elements did match?

docs/selectors.md
```diff
@@ -47,6 +47,11 @@ document
 
 If a selector needs to include `>>` in the body, it should be escaped inside a string to not be confused with chaining separator, e.g. `text="some >> text"`.
 
+### Indexed matches
```
JoelEinbinder, line 50: Does the index start at 0 or at 1?

Does `page.click('button {2}')` match the second button to fulfill the actionability check? Or always the second button, which is then checked for actionability?

Is `page.click('button {1}')` equivalent to `page.cick('button')`?

JoelEinbinder:
While only implemented in safari, the spec says nth-child accepts an optional "of" which I think covers all your use cases? https://www.w3.org/TR/selectors-4/#nth-child-pseudo. Maybe we shouldn't invent new syntax here?

mxschmitt:
> While only implemented in safari, the spec says nth-child accepts an optional "of" which I think covers all your use cases? https://www.w3.org/TR/selectors-4/#nth-child-pseudo. Maybe we shouldn't invent new syntax here?

Once I was creating end-to-end tests I used this syntax (nth-child) for that use-case.

dgozman:
> While only implemented in safari, the spec says nth-child accepts an optional "of" which I think covers all your use cases? https://www.w3.org/TR/selectors-4/#nth-child-pseudo. Maybe we shouldn't invent new syntax here?

How does this work for text selector?

JoelEinbinder:
> > While only implemented in safari, the spec says nth-child accepts an optional "of" which I think covers all your use cases? https://www.w3.org/TR/selectors-4/#nth-child-pseudo. Maybe we shouldn't invent new syntax here?
> 
> How does this work for text selector?

How does yours work with the text selector? You have an example  `article {2} >> .show-button` which implies it is inside the css parser? Can we do `text=foo >> .show-button' today?

dgozman:
> > > While only implemented in safari, the spec says nth-child accepts an optional "of" which I think covers all your use cases? https://www.w3.org/TR/selectors-4/#nth-child-pseudo. Maybe we shouldn't invent new syntax here?
> > 
> > 
> > How does this work for text selector?
> 
> How does yours work with the text selector? You have an example `article {2} >> .show-button` which implies it is inside the css parser? Can we do `text=foo >> .show-button` today?

Yes we can 😄  Mine works as `text=foo {2}`.

JoelEinbinder:
Fancy. So can I do `.foo:not(text=bar)`? Then I'd expect it to be `:nth-child(2 of text=bar)`

docs/selectors.md
```diff
@@ -47,6 +47,11 @@ document
...
+By default, selectors resolve to the first matching element, according to the DOM traversal order. A selector can be followed by `{n}` to match the n-th element instead.
```
JoelEinbinder, line 51:
```suggestion
By default, selectors resolve to the first matching element, according to the DOM traversal order. A Playwright selector can be followed by `{n}` to match the n-th element instead.
```

No meaning change here, but adding the word playwright helps people jumping to this part of the docs without reading the above sections.

PR #4383, opened 2020-11-09
test: increase timeout for video bots

dgozman:


dgozman:
This did not help - so failing tests are probably races, not timeouts.

Issue #4242, opened 2020-10-26
[BUG] Helpful error message for lockfile


arjunattam:
From a [user report on Slack](https://playwright.slack.com/archives/CSUHZPVLM/p1603522725233600):

```
error /Users/xxx/node_modules/playwright: Command failed.
Exit code: 1
Command: node install.js
Arguments: 
Directory: /Users/xxx/node_modules/playwright
Output:
/Users/xxx/node_modules/proper-lockfile/lib/lockfile.js:68
                return callback(Object.assign(new Error('Lock file is already being held'), { code: 'ELOCKED', file }));
                                              ^
Error: Lock file is already being held
    at /Users/xxx/node_modules/proper-lockfile/lib/lockfile.js:68:47
    at callback (/Users/xxx/node_modules/graceful-fs/polyfills.js:295:20)
    at FSReqCallback.oncomplete (node:fs:184:5) {
```

Manually cleaning up the browsers directory and reinstalling fixed it. Not sure how the lockfile was initialized (hopefully not a yarn quirk), but we could show a more helpful error message suggesting towards a fix.

Update: This response might explain the cause

> Okay great. Not sure if it will help your debugging but I had a poor internet connection that time so the installation was taking too long to go through. I had to terminate the installation Ctrl + C in the middle and then ran yarn add … again.

buggedcom:
If anyone else arrives here due to searching about lock file errors. I had the same problem, but came to realise that since I had included playwright and all the browser packages as dependencies that the lockfile was causing issues because the browser packages were simultaneously being built. 

To fix, just change

```
		"playwright": "^1.9.1",
		"playwright-chromium": "^1.9.1",
		"playwright-firefox": "^1.9.1",
		"playwright-webkit": "^1.9.1",
```
to
```
		"playwright": "^1.9.1",
```

in your `package.json`

PR #4396, opened 2020-11-10
fix(lock): nicer lockfile error

JoelEinbinder:
Show the path in the lockfile error so users have a chance of debugging it. #4242

### DIFF
--- a/src/install/installer.ts
+++ b/src/install/installer.ts
@@ -41,6 +41,7 @@ export async function installBrowsersWithProgressBar(packagePath: string) {
 
   const browsersPath = browserPaths.browsersPath(packagePath);
   await fsMkdirAsync(browsersPath, { recursive: true });
+  const lockfilePath = path.join(browsersPath, '__dirlock');
   const releaseLock = await lockfile.lock(browsersPath, {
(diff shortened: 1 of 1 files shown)


src/install/installer.ts
```diff
@@ -41,6 +41,7 @@ export async function installBrowsersWithProgressBar(packagePath: string) {
 
   const browsersPath = browserPaths.browsersPath(packagePath);
   await fsMkdirAsync(browsersPath, { recursive: true });
+  const lockfilePath = path.join(browsersPath, '__dirlock');
   const releaseLock = await lockfile.lock(browsersPath, {
```
dgozman, line 45: From what I see on 'proper-lockfile' usage, `.lock()` may also throw, as well as `releaseLock`. Should we capture them and rethrow with a nicer error as well?

JoelEinbinder: My understanding is that `onCompromised` is called on somewhat expected errors. Like when another process won't let go of our lock. Unexpected errors, like we tried to release the same lock twice, just get thrown as usual.

dgozman (approved):

Issue #4379, opened 2020-11-08
[Question] Is it possible to extend launchServer?


Alexeykhr:
Hi, thanks for the wonderful library. Can I extend the logic for working with a websocket connection?

[lauchServer](https://github.com/microsoft/playwright/blob/master/src/browserServerImpl.ts#L43)

For example, monitor connections, catch messages, etc.

And the `logger`, should it display the commands being executed?

![image](https://user-images.githubusercontent.com/14747569/98465428-abd9ef00-21d1-11eb-9cf3-0f6297502e8d.png)

pavelfeldman:
You should be getting log messages when client attaches to the socket, but they will be low-level ones. Commands being executed are available to the logger on the client side. What do you mean by catch messages?

Alexeykhr:
I'm sorry, I probably didn't understand the logic of creating a server :)

I originally thought this. If a couple of clients connect to the server, then how can I track/limit the number of connections, maybe filter commands, etc.?

And also the question, if I cannot track the commands and everything that comes from the client to the server, then perhaps through the logs on the server I can track the commands being executed, since the property is present, why not :)

So .. Apparently I didn't understand the original purpose of the method `lauchServer`

pavelfeldman:
Hang on, it looks like logger passed into launchServer is broken. If you are interested in sniffing to the protocol, you can run Playwright with `DEBUG=pw:channel*` env variable, no need to use `launchServer` for that, things work over the protocol internally anyways.

yury-s:
The protocol messages are now passed to the logger. Feel free to reopen if you have further questions.

PR #4403, opened 2020-11-11
fix(launchServer): stream protocol logs into options.logger

dgozman:
This has regressed in v1.4.

References #4379.

### DIFF
--- a/src/browserServerImpl.ts
+++ b/src/browserServerImpl.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { LaunchServerOptions } from './client/types';
+import { LaunchServerOptions, Logger } from './client/types';
 import { BrowserType } from './server/browserType';
 import * as ws from 'ws';
 import * as fs from 'fs';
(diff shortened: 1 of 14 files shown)


src/browserServerImpl.ts
```diff
@@ -46,7 +47,7 @@ export class BrowserServerLauncherImpl implements BrowserServerLauncher {
       ignoreDefaultArgs: Array.isArray(options.ignoreDefaultArgs) ? options.ignoreDefaultArgs : undefined,
       ignoreAllDefaultArgs: !!options.ignoreDefaultArgs && !Array.isArray(options.ignoreDefaultArgs),
       env: options.env ? envObjectToArray(options.env) : undefined,
-    });
+    }, toProtocolLogger(options.logger));
     return new BrowserServerImpl(browser, options.port);
```
yury-s, line 51: I believe it would be helpful to also see logs of incoming connections and calls of the methods on BrowserServer (there was a question about the connection logging in the bug).

yury-s (approved):

pavelfeldman (approved):

Issue #4393, opened 2020-11-10
[BUG] TypeError: Cannot read property 'childFrames' of null with series of redirects


DJ-Glock:
**Context:**
- Playwright Version: 1.5.2
- Operating System: Windows
- Node.js version: 12.18.3
- Browser: Chromium

**Code Snippet**
Please be aware that this code snippet is applicable for my website and it's internal. I tried to find some public website with similar scenario to reproduce the issue, but was no able to do it.

```javascript
const playwright = require('playwright');

(async () => {
  const browser = await playwright['chromium'].launch( { headless: true });
  const context = await browser.newContext();
  const page = await context.newPage();
  
  await page.goto(`https://our-internal-website/`, { waitUntil: 'load'});
  // Browser will be redirected to login page
  await page.type(`[id="login-field"]`,'login');
  await page.type(`[id="password-field"]`,'pass');
  await page.click(`[name="button"][value="login"]`);
  // Once logged in, browser will be redirected to website url
  await page.waitForNavigation({ url: 'https://our-internal-website/'});

  // Main issue is here:
  await page.click(`[data-test-id="toolbar-user"]`);
  await page.click(`[data-test-id="logout-menu-item"]`);
  // Once clicked, browser will be redirected to logout page
  // And here it fails in headful mode (in headless it works fine)

  await page.waitForNavigation({ url: /\/logged-out/, waitUntil: 'networkidle' });

  await browser.close();
})();
```

**Describe the bug**
There should be several redirects to logout page. But they failed in **headful** mode with the following error:
```
(node:4780) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'childFrames' of null
    at FrameManager.removeChildFramesRecursively (D:\Apps\..\QA_Automation_Example\qa-automation\node_modules\playwright\lib\server\frames.js:229:35)
    at FrameSession._onAttachedToTarget (D:\Apps\..\QA_Automation_Example\qa-automation\node_modules\playwright\lib\server\chromium\crPage.js:488:38)
    at CRSession.<anonymous> (D:\Apps\..\QA_Automation_Example\qa-automation\node_modules\playwright\lib\server\chromium\crPage.js:302:101)
    at CRSession.emit (events.js:315:20)

```
In headless mode everything works fine. Please check logs with good and bad examples.

**Good example (headless mode):**
```
  pw:api => browserType.launch started +0ms
  pw:api <= browserType.launch succeeded +623ms
  pw:api => browser.newContext started +1ms
[...]

PR #4410, opened 2020-11-11
fix(oopif): store child frame id between frameDetached and attachedToTarget

yury-s:
#4393

### DIFF
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -556,9 +556,11 @@ class FrameSession {
 
   private _frameMaybeSwappingOut(frameId: string) {
     const frame = this._page._frameManager.frame(frameId);
-    if (!frame) return;
+    if (!frame)
+      return;
     const parent = frame.parentFrame() as any;
-    if (!parent) return;
+    if (!parent)
+      return;
     if (!parent[swappingOutChildFrames])
       parent[swappingOutChildFrames] = new Set<string>();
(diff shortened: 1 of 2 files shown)


src/server/chromium/crPage.ts
```diff
@@ -584,7 +607,15 @@ class FrameSession {
       // Frame id equals target id.
       const targetId = event.targetInfo.targetId;
       const frame = this._page._frameManager.frame(targetId)!;
```
dgozman, line 609: nit: remove "!"

src/server/chromium/crPage.ts
```diff
@@ -542,12 +544,33 @@ class FrameSession {
...
+        continue;
+      if ((frame as any)[swappingOutChildFrames].delete(targetId))
+        return frame._id;
+    }
+    return null;
+  }
+
+  private _frameMaybeSwappingOut(frameId: string) {
+    const frame = this._page._frameManager.frame(frameId);
+    if (!frame) return;
```
dgozman, line 559: style: return on the next line. I am surprised that eslint does not complain...

dgozman (approved):

src/server/chromium/crPage.ts
yury-s: yeah, I relied on it

src/server/chromium/crPage.ts
yury-s: done

Issue #4269, opened 2020-10-28
[Feature] Export credential


nicoandmee:
[Original Issue on puppeteer](https://github.com/puppeteer/puppeteer/issues/4766)

- Currently, there is no reliable way to clone a browser (all session data/cookies/user data dir structure) without persisting a user data directory on disk. This makes running tests that require persistent authentication (like MFA prompts) somewhat tricky. 

- Proposing adding functions to API to serve this purpose: `exportCredential()`, `importCredential(str)`. Export simply returns a base64-encoded string. You then provide this string when instantiating the browser, which unpacks the session data and sets the cookies. Sort of like freezing the browser in time, and then reactivating it.

- I am using node-tar to gzip the userDataDir, the implementation looks something like this:

```
const stream = await tar.x({
      cwd: browser.userDataDir,
      gzip: true,
    });
    stream.end(Buffer.isBuffer(browser.parsedCredential.chromeUserData) ? browser.parsedCredential.chromeUserData : Buffer.from(browser.parsedCredential.chromeUserData, 'base64'));

```

Import function looks like:
```
async importCredential(zbuf) {
    await new Promise((resolve, reject) => {
      zlib.gunzip(Buffer.from(zbuf, 'base64'), (err, buf) => {
        if (err) return reject(err);
        browser.parsedCredential = JSON.parse(buf.toString());
        return resolve();
      });
    });

    await this.loadCredentials();
  }
```
parsedCredential is composed of:
- all cookies (including third party): from `._client.send('Network.getAllCookies')`
- cookies: from `page.cookies()`
- chromeUserData: from:
```

async function streamToBuffer(stream) {
  return new Promise((resolve, reject) => {
    const buffers = [];
    stream.on('error', reject);
    stream.on('data', data => buffers.push(data));
    stream.on('end', () => resolve(Buffer.concat(buffers)));
  });
}

 const buf = await streamToBuffer(await tar.c({
        gzip: true,
        cwd: browser.userDataDir,
        filter(path) {
          if (/BrowserMetrics-active\.pma$/i.test(path)) return false;
          if (/\/Favicons(-journal)?$/.test(path)) return false;
          if (/\/History(-journal)?$/.test(path)) return false;
          if (/\/Top Sites(-journal)?$/.test(path)) return false;
          if (/\/Last Session$/.test(path)) return false;
          if (/\/Last Tabs$/.test(path)) return false;
          if (/\/Visited Links$/.test(path)) return false;
[...]

PR #4412, opened 2020-11-12
feat(state): allow getting / setting context state

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/4269

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -240,6 +240,22 @@ Indicates that the browser is connected.
     - `size` <[Object]> Optional dimensions of the recorded videos. If not specified the size will be equal to `viewport`. If `viewport` is not configured explicitly the video size defaults to 1280x720. Actual picture of each page will be scaled down if necessary to fit the specified size.
       - `width` <[number]> Video frame width.
       - `height` <[number]> Video frame height.
+  - `storageState` <[Object]> Populates context with given storage state. This method can be used to initialize context with logged-in information obtained via [browserContext.storageState()](#browsercontextstoragestate).
+    - `cookies` <[Array]<[Object]>> Optional cookies to set for context
+      - `name` <[string]> **required**
+      - `value` <[string]> **required**
+      - `url` <[string]> Optional either url or domain / path are required
+      - `domain` <[string]> Optional either url or domain / path are required
+      - `path` <[string]> Optional either url or domain / path are required
+      - `expires` <[number]> Optional Unix time in seconds.
+      - `httpOnly` <[boolean]> Optional httpOnly flag
+      - `secure` <[boolean]> Optional secure flag
+      - `sameSite` <"Strict"|"Lax"|"None"> Optional sameSite flag
+    - `origins` <[Array]<[Object]>> Optional localStorage to set for context
+      - `origin` <[string]>
+      - `localStorage` <[Array]<[Object]>>
+        - `name` <[string]>
+        - `value` <[string]>
 - returns: <[Promise]<[BrowserContext]>>
 
 Creates a new browser context. It won't share cookies/cache with other browser contexts.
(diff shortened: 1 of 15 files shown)


docs/api.md
```diff
@@ -680,6 +682,46 @@ Provide credentials for [HTTP authentication](https://developer.mozilla.org/en-U
...
+    - `localStorage` <[Array]<[Object]>>
+      - `name` <[string]>
+      - `value` <[string]>
+- returns: <[Promise]>
+
+Populates context with given storage Storagestate. This method can be used to initialize context with logged in information obrained via [browserContext.storageState()](#browsercontextstoragestate).
+
+#### browserContext.storageState()
+- returns: <[Promise]<[Object]>>
+  - `cookies` <[Object]>
```
dgozman, line 708: Array of Objects

docs/api.md
```diff
@@ -680,6 +682,46 @@ Provide credentials for [HTTP authentication](https://developer.mozilla.org/en-U
 - `offline` <[boolean]> Whether to emulate network being offline for the browser context.
 - returns: <[Promise]>
 
+#### browserContext.setStorageState(options)
+- `options` <[Object]> Optional
```
dgozman, line 686: "Optional" should be on the next line?

docs/api.md
```diff
@@ -680,6 +682,46 @@ Provide credentials for [HTTP authentication](https://developer.mozilla.org/en-U
...
+    - `secure` <[boolean]> Optional secure flag
+    - `sameSite` <"Strict"|"Lax"|"None"> Optional sameSite flag
+  - `origins` <[Array]<[Object]>> Optional
+    - `origin` <[string]>
+    - `localStorage` <[Array]<[Object]>>
+      - `name` <[string]>
+      - `value` <[string]>
+- returns: <[Promise]>
+
+Populates context with given storage Storagestate. This method can be used to initialize context with logged in information obrained via [browserContext.storageState()](#browsercontextstoragestate).
```
dgozman, line 704: "storage Storagestate"

src/client/browserContext.ts
```diff
@@ -219,6 +219,18 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel,
     return result;
   }
 
+  async setStorageState(options: SetStorageState) {
+    await this._wrapApiCall('browserContext.setStorageState', async () => {
+      await this._channel.setStorageState(options);
+    });
+  }
+
+  async storageState(): Promise<StorageState> {
```
dgozman, line 228: Since this is async function, I'd prefer a verb in the name, e.g. `saveStorageState` or `captureStorageSnapshot` or something.

src/server/page.ts
```diff
@@ -428,6 +438,14 @@ export class Page extends EventEmitter {
     this._video = video;
     this.emit(Page.Events.VideoStarted, video);
   }
+
+  frameNavigated(frame: frames.Frame) {
+    this.emit(Page.Events.FrameNavigated, frame);
+    const url = frame.url();
+    if (url === 'about:blank' || url.startsWith('chrome:'))
```
dgozman, line 445: Do we want origins for data: urls, file: urls, other about: urls? Perhaps `!url.startsWith('http')` ?

src/server/browserContext.ts
```diff
@@ -310,6 +311,56 @@ export abstract class BrowserContext extends EventEmitter {
...
+        originStorage.localStorage = storage.localStorage;
+      }
+      await page.close();
+    }
+    return result;
+  }
+
+  async setStorageState(state: types.SetStorageState) {
+    if (state.cookies)
+      await this.addCookies(state.cookies);
```
dgozman, line 346: Should we clear cookies before that?

test/browsercontext-storage-state.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  const context1 = await browser.newContext();
+  const page1 = await context1.newPage();
+  await page1.route('**/*', route => {
+    route.fulfill({ body: '<html></html>' }).catch(() => {});
+  });
+  await page1.goto('https://www.example.com');
+  await page1.evaluate(() => {
+    localStorage['name1'] = 'value1';
+  });
+  const { origins } = await context1.storageState();
```
dgozman, line 30: Let's navigate to another origin, to check that we handle multiple origins correctly.

dgozman (approved):

src/server/browserContext.ts
pavelfeldman: Good question, we would need to clear localstorage as well. And that way we are no longer composable...

src/client/browserContext.ts
pavelfeldman: You mean like `page.screenshot()` or `page.pdf()`?

src/server/browserContext.ts
pavelfeldman: Removed the method!

src/client/browserContext.ts
IanVS: I do have to say, this confused me in the [docs](https://playwright.dev/docs/auth#reuse-authentication-state):

```
// Save storage state into the file.
await context.storageState({ path: 'state.json' });
```

Especially since the description is `The following code snippet retrieves state from an authenticated context and creates a new context with that state.`.  I couldn't figure out why the result of `context.storageState`, which seemed to be getting storage state, wasn't returning anything that could be saved in the next step.  I'm a brand new playwright user, trying to figure out how I'd handle auth that includes MFA, which is how I ended up looking at this PR.

PR #4428, opened 2020-11-13
browser(firefox): force a layout before dispatching a tap

JoelEinbinder:
Sometimes tap tests would fail in firefox. The touch events would miss, but the generated mouse events would still hit their target. This fixes it.


browser_patches/firefox/BUILD_NUMBER
```diff
@@ -1,2 +1,2 @@
-1206
-Changed: yurys@chromium.org Thu 12 Nov 2020 10:23:24 AM PST
+1207
```
dgozman, line 1: You'll need to rebase.

dgozman (approved):

Issue #1489, opened 2020-03-23
[Question] Unable to redirect url using Page.route


sarut0bi:
hi,
i tried to make a redirection on XHR on another url unsucessfully
here is what i tried unsuccessfully:
```
await page.route('https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikipedia-logo-v2@1.5x.png', (request) => {
  request.continue({url: 'https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikinews-logo_sister@2x.png'});
}
await page.goto('https://www.wikipedia.org/');
```
and 

```
await page.route('https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikipedia-logo-v2@1.5x.png', (request) => {
  request._url = 'https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikinews-logo_sister@2x.png';
  request.continue();
}
await page.goto('https://www.wikipedia.org/');
```

Could you provide me a good way to do it?

arjun27:
Hi @sarut0bi, it is not possible to override the url in the `continue` method ([overrides doc](https://github.com/microsoft/playwright/blob/master/docs/api.md#routecontinueoverrides)). However, you could `fulfill` the request with the desired image ([docs](https://github.com/microsoft/playwright/blob/master/docs/api.md#routefulfillresponse))

sarut0bi:
i ve fulfilled it with a 301 :)
```
        request.fulfill(    {status: 301,
          headers: {Location: 'https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikinews-logo_sister@2x.png' });
```

arjun27:
That's a better solution! :)

AnnaIvakhnik:
Hi! Thank you for the solution! It works great for me while running chromium and firefox, but does not work for WebKit :( Do you know what could be the problem? 
ps: it not the cors problem, but i tried adding header 'Access-control-allow-origin': '*',
Playwright 1.5.2
WebKit 14.0
Ubuntu 18.04

yury-s:
There is a bug in WebKit implementation of fulfill, I'll have a look.

yury-s:
Request interception on mac is handled within network stack so it doesn't kick in when request if "fulfilled" which leads to wrong behavior there. Instead of trying to work around that (which could end up in another implementation of redirect logic in the browser) we decided to allow to override `url` in `route.continue()`.

[...]

PR #4436, opened 2020-11-13
browser(firefox): allow to override request url

yury-s:
#1489


dgozman (approved):
Let's check somewhere that we are rewriting into http/https, otherwise netwerk might get confused.

yury-s:
Let's check on frontend as discussed.

PR #4449, opened 2020-11-16
fix(route): throw on attempt to fulfill with redirect in WebKit

yury-s:
#1489

### DIFF
--- a/src/server/webkit/wkInterceptableRequest.ts
+++ b/src/server/webkit/wkInterceptableRequest.ts
@@ -74,6 +74,9 @@ export class WKInterceptableRequest implements network.RouteDelegate {
   }
 
   async fulfill(response: types.NormalizedFulfillResponse) {
+    if (300 <= response.status && response.status < 400)
+      throw new Error('Cannot fulfill with redirect status: ' + response.status);
+
     await this._interceptedPromise;
 
     // In certain cases, protocol will return error if the request was already canceled
--- a/test/page-route.spec.ts
+++ b/test/page-route.spec.ts
@@ -395,7 +395,7 @@ it('should intercept main resource during cross-process navigation', async ({pag
 });
 
 it('should fulfill with redirect status', (test, { browserName, headful}) => {
-  test.skip(browserName === 'webkit');
+  test.fixme(browserName === 'webkit', 'in WebKit the redirects are handled by the network stack and we intercept before');
 }, async ({page, server}) => {
   await page.goto(server.PREFIX + '/title.html');
   await page.route('**/*', async (route, request) => {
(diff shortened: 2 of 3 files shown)


src/server/network.ts
```diff
@@ -212,6 +212,8 @@ export class Route {
   }
 
   async fulfill(response: { status?: number, headers?: types.HeadersArray, body?: string, isBase64?: boolean }) {
+    if (response.status && 300 <= response.status && response.status < 400)
+      throw new Error('Cannot fulfill with redirect status');
```
dgozman, line 216: nit: let's include the bad status in the error message

dgozman (approved):

src/server/network.ts
yury-s: Done.

JoelEinbinder:
Redirects are functionally different to fulfilling the redirected content. Consider intercepting a module script. If we fetch the script ourselves instead of providing a redirect, all of the relative imports in that script will point to the wrong place. Can we leave the current behavior for browsers in which it works?

yury-s:
@JoelEinbinder In the example with a module script intercepting all requests with a given prefix should solve the issue or am I missing something? I certainly can leave existing behavior for FF and CR (not sure to what extent it works though) but it'd be nice if the method behaved consistently across browsers.

JoelEinbinder:
if I have `www.example.com/realdata.jsm`, and I want to redirect to replace just this file with a new module located at at `www.example.com/for-testing/mockdata.jsm`. `mockdata.jsm` imports `./foo.jsm` and `./bar.jsm`. If I have a redirect, they will access `for-testing/foo.jsm` and `for-testing/bar.jsm`. If I fulfill the content of `mockdata.jsm`, they will attempt to access the toplevel `/foo.jsm` and `bar.jsm`. I can't use a prefix to know which files were from `mockdata.jsm` and should have different paths.

That's just one example. It also effects `import.meta`, relative urls in css, and I imagine a lot of other subtle scenarios.

dgozman:
> if I have `www.example.com/realdata.jsm`, and I want to redirect to replace just this file with a new module located at at `www.example.com/for-testing/mockdata.jsm`. `mockdata.jsm` imports `./foo.jsm` and `./bar.jsm`. If I have a redirect, they will access `for-testing/foo.jsm` and `for-testing/bar.jsm`. If I fulfill the content of `mockdata.jsm`, they will attempt to access the toplevel `/foo.jsm` and `bar.jsm`. I can't use a prefix to know which files were from `mockdata.jsm` and should have different paths.
> 
> That's just one example. It also effects `import.meta`, relative urls in css, and I imagine a lot of other subtle scenarios.

Sounds like a pretty convoluted usecase. You basically end up with two different `foo.jsm`, one from the real code, and one from the mock data. Could be reasonable, but perhaps we could leave this to the user to sort out?

I think we should strive for cross-browser compatibility as much as we can.

JoelEinbinder:
I've done the module redirecting case. The linked issue has multiple users that want url redirecting, one of which is currently successfully using a 301 redirect. This patch is a regression and will break people.

yury-s:
Ok, let's limit this to webkit.

yury-s:
Updated the code to throw only in webkit if the client tries to fulfill with a redirect. Updated tests to reflect current behavior across browsers.

dgozman (approved):

PR #4482, opened 2020-11-18
test(route): test that intercepted XHR is actually paused

yury-s:


test/page-route.spec.ts
```diff
@@ -146,16 +146,32 @@ it('should show custom HTTP headers', async ({page, server}) => {
...
     request.open('GET', '/logo.png', false);  // `false` makes the request synchronous
     request.send(null);
     return request.status;
+  }).then(r => {
+    xhrFinished = true;
+    return r;
   });
+  const route = await routePromise;
+  // Check that intercepted request is actually paused.
+  await new Promise(r => setTimeout(r, 100));
```
pavelfeldman, line 169: this doesn't sound right, 100 means it'll flake?

test/page-route.spec.ts
```diff
@@ -159,6 +159,58 @@ it('should work with redirect inside sync XHR', async ({page, server}) => {
...
+  const statusPromise = page.evaluate(async () => {
+    const response = await fetch('/global-var.html');
+    return response.status;
+  }).then(r => {
+    fetchFinished = true;
+    return r;
+  });
+  const route = await routePromise;
+  // Check that intercepted request is actually paused.
+  await new Promise(r => setTimeout(r, 100));
```
pavelfeldman, line 205: This needs to be at least 500ms if you want to test the race.

yury-s: Done.

PR #4496, opened 2020-11-19
feat(launcher): stop handling signals, add a watchdog

dgozman:
Changes:
- We used to handle signals in the node process to carefully shutdown launched browsers, for example on SIGINT. This has been removed.
- We used to launch the browser process directly. We now use a watchdog wrapper that ensures correct termination if the browser misbehaves.

Pros:
- No `handleSIGINT` and other signal options in our API. Users do not have to think about it. One example is our own driver that had to keep track of all browsers to close them gracefully, and can just exit the process now.
- Killing the node process with `SIGKILL` does not leave the browser running.

Cons:
- One more wrapper process.
- Protocol piping could be slower, although it was not measured.

The only scenario that leaves the zombie browser is when a) user sends a `SIGKILL` to the watchdog process and b) the browser misbehaves and does not close on pipe disconnect.

Follow up:
Once we don't need compatibility with stock Chrome <= 88, we can stop sending `Browser.close` command and instead just close the pipe, since our browsers recognize that as a signal to close.

Alternatives:
- Leave everything as is.
- Remove the watchdog and hope that browser does not misbehave.
- Just replace `Browser.close` with closing pipe in the future.


docs/api.md
```diff
@@ -4494,9 +4494,9 @@ This methods attaches Playwright to an existing browser instance.
   - `downloadsPath` <[string]> If specified, accepted downloads are downloaded into this directory. Otherwise, temporary directory is created and is deleted when browser is closed.
   - `chromiumSandbox` <[boolean]> Enable Chromium sandboxing. Defaults to `false`.
   - `firefoxUserPrefs` <[Object]<[string], [string]|[number]|[boolean]>> Firefox user preferences. Learn more about the Firefox user preferences at [`about:config`](https://support.mozilla.org/en-US/kb/about-config-editor-firefox).
-  - `handleSIGINT` <[boolean]> Close the browser process on Ctrl-C. Defaults to `true`.
-  - `handleSIGTERM` <[boolean]> Close the browser process on SIGTERM. Defaults to `true`.
-  - `handleSIGHUP` <[boolean]> Close the browser process on SIGHUP. Defaults to `true`.
+  - `handleSIGINT` <[boolean]> Deprecated. The value is ignored.
```
JoelEinbinder, line 4497: I think we can safely delete these from the api docs, and add them into typescript overrides marked with `@deprecated`.

JoelEinbinder: I'll do it in a follow up to this patch.

src/server/watchdog.ts
```diff
@@ -0,0 +1,137 @@
...
+  // and sending the signal to our own process, so that default handler does the magic.
+  process.removeListener('SIGINT' as any, doNothing);
+  process.removeListener('SIGTERM' as any, doNothing);
+  process.kill(process.pid, signal!);
+});
+
+const inputStream = closeMethod === 'none' ? process.stdin : fs.createReadStream('', {fd: 3});
+const outputStream = closeMethod === 'none' ? spawnedProcess.stdin : (spawnedProcess.stdio as any)[3] as NodeJS.WritableStream;
+// Note: we cannot use 'pipe' here since it prohibits our own 'write' to close the browser.
+inputStream.on('data', data => outputStream.write(data));
```
JoelEinbinder, line 106: I think [`stream.pipe`](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options) is warranted here. I don't fully understand what it does, but I think it's smarter than "on data write data".

src/server/watchdog.ts
```diff
@@ -0,0 +1,137 @@
...
+  // Here we mimic the signal of the browser process: removing our signal handlers
+  // and sending the signal to our own process, so that default handler does the magic.
+  process.removeListener('SIGINT' as any, doNothing);
+  process.removeListener('SIGTERM' as any, doNothing);
+  process.kill(process.pid, signal!);
+});
+
+const inputStream = closeMethod === 'none' ? process.stdin : fs.createReadStream('', {fd: 3});
+const outputStream = closeMethod === 'none' ? spawnedProcess.stdin : (spawnedProcess.stdio as any)[3] as NodeJS.WritableStream;
+// Note: we cannot use 'pipe' here since it prohibits our own 'write' to close the browser.
```
JoelEinbinder, line 105: We can't `unpipe` before closing the browser?

PR #4505, opened 2020-11-20
browser(firefox): ensure detachedFromTarget is always sent

dgozman:
LinkedBrowser can throw when removing listeners in PageTarget.dispose,
and that prevents BrowserHandler from sending Browser.detachedFromTarget.

Using a try-catch seems good enough.

### DIFF
--- a/browser_patches/firefox/juggler/TargetRegistry.js
+++ b/browser_patches/firefox/juggler/TargetRegistry.js
@@ -521,7 +521,12 @@ class PageTarget {
     this._browserContext.pages.delete(this);
     this._registry._browserToTarget.delete(this._linkedBrowser);
     this._registry._browserBrowsingContextToTarget.delete(this._linkedBrowser.browsingContext);
-    helper.removeListeners(this._eventListeners);
+    try {
+      helper.removeListeners(this._eventListeners);
+    } catch (e) {
+      // In some cases, removing listeners from this._linkedBrowser fails
+      // because it is already half-destroyed.
+    }
     this._registry.emit(TargetRegistry.Events.TargetDestroyed, this);
   }
 }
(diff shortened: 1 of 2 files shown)


browser_patches/firefox/juggler/TargetRegistry.js
```diff
@@ -521,7 +521,12 @@ class PageTarget {
     this._browserContext.pages.delete(this);
     this._registry._browserToTarget.delete(this._linkedBrowser);
     this._registry._browserBrowsingContextToTarget.delete(this._linkedBrowser.browsingContext);
-    helper.removeListeners(this._eventListeners);
+    try {
+      helper.removeListeners(this._eventListeners);
```
JoelEinbinder, line 525: this will fail to remove event listeners from anything after this._linkedBrowser. Does that matter?

JoelEinbinder:
The bug here is that page.close sometimes times out? Are we seeing that on the bots?

JoelEinbinder (approved):

browser_patches/firefox/juggler/TargetRegistry.js
dgozman: Not today, since linkedBrowser is the last one. Unfortunately, I don't see a better alternative. Doing a generic try/catch around each event listener in `Helper.removeListeners` could mask other problems in the future, so it seems like limiting the issues to this particular place is better. WDYT?

JoelEinbinder: Let's dump the error and not worry about it.

dgozman:
> The bug here is that page.close sometimes times out? Are we seeing that on the bots?

This is a fix for `autowaiting-no-hang.spec.ts:60:1 › opening a popup` on video bots. Not related to `page.close`, but rather to popup detach.

browser_patches/firefox/juggler/TargetRegistry.js
```diff
@@ -521,7 +521,12 @@ class PageTarget {
...
+    } catch (e) {
+      // In some cases, removing listeners from this._linkedBrowser fails
+      // because it is already half-destroyed.
```
JoelEinbinder, line 528:
```suggestion
      // because it is already half-destroyed.
      if (e)
        dump(e.message + '\n' + e.stack + '\n');
```

PR #4508, opened 2020-11-20
test(drag): more tests for drag and drop

JoelEinbinder:
Adds a bunch more drag and drop tests. All set to fixme right now, but #4506 will turn on four for firefox. Chromium and Webkit patches are larger, but coming soon. The golden event orders here were tested against mostly working versions of patches in all browsers.

### DIFF
--- a/test/drag.spec.ts
+++ b/test/drag.spec.ts
@@ -1,0 +1,161 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


test/drag.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+describe('Drag and drop', test => {
+  test.fixme();
+}, () => {
+  it('should work', async ({server, page, context}) => {
+    await page.goto(server.PREFIX + '/drag-n-drop.html');
+    await page.hover('#source');
+    await page.mouse.down();
+    await page.hover('#target');
+    await page.mouse.up();
+    expect(await page.$eval('#target', target => target.contains(document.querySelector('#source')))).toBe(true); // could not find source in target
```
dgozman, line 29: What does "could not find source in target" mean here?

dgozman (approved):

test/drag.spec.ts
JoelEinbinder: We use to have error messages after the error as a string, which was helpful when the test failed. When we moved to the expect library, these got removed. But the error reporter shows the line that failed now, so I just put a message in a comment after the line.

Issue #1094, opened 2020-02-25
[Feature] drag and drop


JoelEinbinder:
Drag and drop should work with playwright. Using the existing mouse apis should generate working `dragstart`, `drag`, and `drop` events. There should also be new APIs to simulate dropping in external content, and to inspect content dragged from the web page.

JoelEinbinder:
# Chromium
I had a patch that tries to do this years ago. Maybe I can revive it? It ended up touching a lot of layers.
https://chromium-review.googlesource.com/c/chromium/src/+/1340803

# WebKit
Blocked on WPE not having drag and drop enabled. I'll have to enable it first before deciding how to mock it with playwright.

# Firefox
I haven't investigated firefox yet.

isenese:
Is it possible to drag and drop at this stage?

skum-hub:
Is there a way to drag and drop an element to a particular location on the page? I want to be able to drag and drop either horizontally or vertically by a certain pixels. The current dragAnddrop() method brings the dragger back to its original position as I'm not able to drag it enough to drop it in placeholder.

MedBechirSaid:
Hi ! i wanna know if there is a solution to test drag and drop angular with playwright.
I tried to use this code: 
await page.dispatchEvent('#target', 'a', { draggableElem });
but it didn't work ! 
any help plz !
i'm using angular 9 / playwright 1.3.0

lironzizva:
Any update on this? when drag & drop is planned to be implemented?

lind-eduard:
Want to bring this topic up....Are there any progress on this?

hptabster:
Is there a workaround in Playwright 1.8?

cscheffauer:
+1 for this

janpfajfr:
+1 for this

chenshuichao:
+1 for this

lind-eduard:
For all who struggling with this issue, I found a workaround for it:
```
async dragElementToOtherElement(originSelector, destinationSelector) {
		const origin = await page.waitForSelector(originSelector);
		const destination = await page.waitForSelector(destinationSelector);
		const originBox = await origin.boundingBox();
		const destinationBox = await destination.boundingBox();
		const lastPositionCoordenate = (box) => ({
			x: box.x + box.width / 2,
			y: box.y + box.height
		});
		const getPayload = (box) => ({
			bubbles: true,
			cancelable: true,
			screenX: lastPositionCoordenate(box).x,
[...]

PR #4519, opened 2020-11-24
feat(chromium): basic drag and drop support

JoelEinbinder:
This does drag and drop for chromium in a very hacky way all in the client. It covers the fast majority of use cases. We can remove the majority of the code by adding a CDP method to attack drag data to the mouse and another to intercept the dragstart. The chromium patches involve a lot of tricky IPC races/security concerns, and so this hacky patch lets us take our time over there.

#1094

closes #4508

### DIFF
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -39,6 +39,7 @@ import * as sourceMap from '../../utils/sourceMap';
 import { rewriteErrorMessage } from '../../utils/stackTrace';
 import { assert, headersArrayToObject, createGuid } from '../../utils/utils';
 import { VideoRecorder } from './videoRecorder';
+import { DragManager } from './crDragDrop';
 
 
 const UTILITY_WORLD_NAME = '__playwright_utility_world__';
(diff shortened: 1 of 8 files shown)


src/server/chromium/crPage.ts
```diff
@@ -127,10 +130,17 @@ export class CRPage implements PageDelegate {
   }
 
   async exposeBinding(binding: PageBinding) {
-    await this._forAllFrameSessions(frame => frame._initBinding(binding));
+    await this._forAllFrameSessions(frame => frame._initBinding(binding, 'main'));
     await Promise.all(this._page.frames().map(frame => frame._evaluateExpression(binding.source, false, {}).catch(e => {})));
   }
 
+  async exposeInternalBinding(name: string, callback: frames.FunctionWithSource) {
```
dgozman, line 137: I think we should move this to the common Page class.

src/server/injected/dragScript.ts
```diff
@@ -0,0 +1,115 @@
+/**
```
dgozman, line 1: Let's make this a part of InjectedScript, but install the `dragstart` handler lazily as you do now.

test/drag.spec.ts
```diff
@@ -127,8 +127,8 @@ describe('Drag and drop', test => {
       'mousedown',
       isFirefox ? 'dragstart' : 'mousemove',
       isFirefox ? 'mousemove' : 'dragstart',
-      'dragenter',
-      'dragover',
+      // 'dragenter',
```
dgozman, line 130: What's up with these?

dgozman (commented):
Could you please split this into 3 PRs: internal binding, new tests and implementation?

PR #4522, opened 2020-11-25
feat(selectors): introduce css parser

dgozman:
It is not used for anything just yet.

### DIFF
--- a/src/server/common/cssParser.ts
+++ b/src/server/common/cssParser.ts
@@ -1,0 +1,238 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as css from './cssTokenizer';
+
+// TODO: Consider giving more information, e.g. whether the argument was a quoted string or not.
+type ParsedSelectorLiteral = string;
+type ClauseCombinator = '' | '>' | '+' | '~';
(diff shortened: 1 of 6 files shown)


src/server/common/cssParser.ts
```diff
@@ -0,0 +1,205 @@
...
+  function isComma(p = pos) {
+    return tokens[p] instanceof css.CommaToken;
+  }
+
+  function isCloseParen(p = pos) {
+    return tokens[p] instanceof css.CloseParenToken;
+  }
+
+  function isStar(p = pos) {
+    return (tokens[p] instanceof css.DelimToken) && tokens[p].value === '*';
```
aslushnikov, line 80: this needs to check for `p < tokens.length`

src/server/common/cssParser.ts
```diff
@@ -0,0 +1,205 @@
...
+    let rawCSSString = '';
+    const funcs: { name: string, args: ParsedSelectorList }[] = [];
+
+    while (!isSelectorClauseEnd()) {
+      if (isIdent() || isStar()) {
+        rawCSSString += tokens[pos++].toSource();
+      } else if (tokens[pos] instanceof css.HashToken) {
+        rawCSSString += tokens[pos++].toSource();
+      } else if ((tokens[pos] instanceof css.DelimToken) && tokens[pos].value === '.') {
+        pos++;
```
aslushnikov, line 139: can we have a helper function `advance()` that'll check for out-of-bounds for `pos`? I think most of the code here will work since running `instanceof` against `undefined` is fine, but it feels shady.

src/server/common/cssParser.ts
```diff
@@ -0,0 +1,205 @@
...
+          if (!isCloseParen())
+            throw unexpected();
+          pos++;
+        } else {
+          throw unexpected();
+        }
+      } else if (tokens[pos] instanceof css.OpenSquareToken) {
+        rawCSSString += '[';
+        pos++;
+        while (!(tokens[pos] instanceof css.CloseSquareToken))
```
aslushnikov, line 168: this while loop should check for out-of-bounds

test/css-parser.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+  expectError('div)');
+  expectError('()');
+  expectError(':not(##)');
+  expectError(':not()');
+  expectError(':not(.)');
+  expectError('div,');
+  expectError(',div');
+  expectError('div,,span');
+  expectError('div > > span');
+  expectError('div > > > > span');
```
aslushnikov, line 96: expectError(`div >`);

aslushnikov (approved):

PR #4529, opened 2020-11-25
Handle headers == undefined

zmitry:
here is stack trace where where I've got this issue.
      at Object.headersArrayToObject (node_modules/playwright/lib/utils/utils.js:128:35)
      at new Response (node_modules/playwright/lib/client/network.js:186:42)
      at Connection._createRemoteObject (node_modules/playwright/lib/client/connection.js:208:26)
      at Connection.dispatch (node_modules/playwright/lib/client/connection.js:105:18)
      at Immediate.<anonymous> (node_modules/playwright/lib/client/browserType.js:112:50)

Maybe it's caused by another issue, then just close the pr

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -125,6 +125,10 @@ export function headersObjectToArray(headers: HeadersObject): HeadersArray {
 }
 
 export function headersArrayToObject(headers: HeadersArray, lowerCase: boolean): HeadersObject {
+  // In some cases headers are undefined.
+  if(!Array.isArray(headers)) {
+     return {}
+  }
   const result: HeadersObject = {};
   for (const { name, value } of headers)
     result[lowerCase ? name.toLowerCase() : name] = value;


JoelEinbinder:
Hi! Thanks for the report. I don't really want to land this until we can figure out whats going on. Can you provide some more info? What browser/operating are you using? Do you have some steps I can run to reproduce the issue, or does it happen sporadically? Thanks!

zmitry:
Hi, sure I guess the problem somewhere else. So the setup I had simple cra setup + jest-playwright integration. I was using your docker image. I've opened docker server and started playwright server there. Then I've connected to playwright using playwright.connect api. I was using chrome only with default device. Within the test I tried to do simple screenshot test and that's it. Just open html and do screenshot. I bet the problem because of the connection. As for server I just used example from the doc.

zmitry:
If you could not reproduce it, close this pr. I've switched to puppeteer so I no longer have that code.

PR #4536, opened 2020-11-28
feat: async support for waitForResponse predicate

### LINKED ISSUE #4535: [Feature] page.waitForResponse support for async predicate function
The predicate with which I want to capture a response sometimes involves looking at the response body which returns a promise. It would be nice if the predicate function had promise support.

`const response = await page.waitForResponse(async (response) => { return await response.text().includes('<html />') });`

mickykebe:
Fixes #4535

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2142,6 +2142,7 @@ await page.waitForRequest(request => request.url().searchParams.get('foo') === '
 ```js
 const firstResponse = await page.waitForResponse('https://example.com/resource');
 const finalResponse = await page.waitForResponse(response => response.url() === 'https://example.com' && response.status() === 200);
+const finalResponse = await page.waitForResponse(async (response) => { await response.text().includes('<html>') });
 return finalResponse.ok();
 ```
 
--- a/test/page-wait-for-response.spec.ts
+++ b/test/page-wait-for-response.spec.ts
@@ -59,7 +59,7 @@ it('should work with predicate', async ({page, server}) => {
 it('should work with async predicate', async ({page, server}) => {
   await page.goto(server.EMPTY_PAGE);
   const [response] = await Promise.all([
-    page.waitForEvent('request', async response => response.url() === server.PREFIX + '/digits/2.png'),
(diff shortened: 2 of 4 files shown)


JoelEinbinder:
Thanks for the PR! We may want to audit our other predicates to make sure they behave the same way. Our docs say that `waitForEvent` can take an async predicate, but looking at the code I don't think it does.

docs/api.md
```diff
@@ -2142,6 +2142,7 @@ await page.waitForRequest(request => request.url().searchParams.get('foo') === '
 ```js
 const firstResponse = await page.waitForResponse('https://example.com/resource');
 const finalResponse = await page.waitForResponse(response => response.url() === 'https://example.com' && response.status() === 200);
+const finalResponse = await page.waitForResponse(async (response) => { await response.text().includes('<html>') });
```
dgozman, line 2145: Can we come up with a more realistic example?
Also, you probably want to name the variable something else but `finalResponse` and most likely meant to do `(await response.text()).includes(...)` instead of calling `includes` on the `Promise` object?

test/page-wait-for-response.spec.ts
```diff
@@ -56,6 +56,19 @@ it('should work with predicate', async ({page, server}) => {
   expect(response.url()).toBe(server.PREFIX + '/digits/2.png');
 });
 
+it('should work with async predicate', async ({page, server}) => {
+  await page.goto(server.EMPTY_PAGE);
+  const [response] = await Promise.all([
+    page.waitForResponse(async response => response.url() === server.PREFIX + '/digits/2.png'),
```
dgozman, line 62: Let's do a more realistic one? For example, load `one-style.html` and wait for response with `background-color` in the body.

dgozman (commented):
Thank you for the PR! I left some more comments in addition to what Joel says, but great work overall. Also, could you please rename the PR to `api: ...` because it touches the public api?

pavelfeldman:
Looks stale, sorry about not being able to merge it.

PR #4540, opened 2020-11-30
devops(flakiness): azure function to store dashboard in a new format

aslushnikov:
This refactors azure function and adds a new dashboard format implementation that is 15x smaller while tracking specs with non-empty annotations.

Old dashboard is still updated.

### DIFF
--- a/utils/flakiness-dashboard/processing/dashboard_v2.js
+++ b/utils/flakiness-dashboard/processing/dashboard_v2.js
@@ -1,0 +1,155 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const {SimpleBlob, flattenSpecs} = require('./utils.js');
+
+const DASHBOARD_VERSION = 1;
+
+class Dashboard {
+  constructor() {
+    this._specs = new Map();
+    this._commits = new Map();
+  }
+
+  initialize(jsonData) {
+    if (jsonData.version !== DASHBOARD_VERSION) {
+      // Run migrations here!
+    }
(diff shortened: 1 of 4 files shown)


utils/flakiness-dashboard/processing/utils.js
```diff
@@ -0,0 +1,80 @@
...
+      blobHTTPHeaders: {
+        blobContentEncoding: 'gzip',
+        blobContentType: 'application/json; charset=UTF-8',
+      }
+    });
+  }
+}
+
+async function deleteBlob(container, blobName) {
+  // First we do - delete the blob.
```
JoelEinbinder, line 75:
```suggestion
```
I think first we delete the blob, last we delete the blob. In a function called delete blob. I think this comment is not needed :).

utils/flakiness-dashboard/processing/index.js
```diff
@@ -14,160 +14,22 @@
...
-      URL: ${report.metadata.runURL}
-      timestamp: ${report.metadata.commitTimestamp}
-      added specs: ${addedSpecs}
-    ===== complete in ${Date.now() - timestamp}ms =====
-    `);
-  } catch (e) {
-    context.log(e);
-    return;
-  }
+  // Uplaod report to both dashboards.
```
JoelEinbinder, line 29:
```suggestion
  // Upload report to both dashboards.
```

JoelEinbinder:
When do we stop updating the v1 dashboard?

utils/flakiness-dashboard/processing/dashboard_v2.js
```diff
@@ -0,0 +1,155 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const {SimpleBlob, flattenSpecs} = require('./utils.js');
+
+const DASHBOARD_VERSION = 1;
```
JoelEinbinder, line 19: should this not be 2?
```suggestion
const DASHBOARD_VERSION = 2;
```

JoelEinbinder (approved):

utils/flakiness-dashboard/processing/dashboard_v2.js
aslushnikov: v2 dashboard is side-by-side with the v1 dashboard in a separate blob - so no changes are necessary

aslushnikov:
> When do we stop updating the v1 dashboard?

Once front-end is migrated to the new format and it proves to be working just fine :-)

PR #4547, opened 2020-11-30
fix(lint): update check_deps for src/remote

JoelEinbinder:
check_deps was throwing an error, but nobody was catching it and it still returned an exit code 0. I fixed that, and also fixed the error by adding deps for src/remote.

### DIFF
--- a/utils/check_deps.js
+++ b/utils/check_deps.js
@@ -124,9 +124,8 @@ DEPS['src/trace/'] = ['src/utils/', 'src/client/**', 'src/server/**'];
 DEPS['src/debug/'] = ['src/utils/', 'src/generated/', 'src/server/**', 'src/debug/**'];
 
 // The service is a cross-cutting feature, and so it depends on a bunch of things.
-DEPS['src/remote/playwrightClient.ts'] = ['src/client/'];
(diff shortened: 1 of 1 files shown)


utils/check_deps.js
```diff
@@ -123,4 +123,12 @@ DEPS['src/trace/'] = ['src/utils/', 'src/client/**', 'src/server/**'];
 // Debug is a server plugin, nothing should depend on it.
 DEPS['src/debug/'] = ['src/utils/', 'src/generated/', 'src/server/**', 'src/debug/**'];
 
-checkDeps();
+// The service is a cross-cutting feature, and so it depends on a bunch of things.
+DEPS['src/remote/playwrightClient.ts'] = ['src/client/'];
```
dgozman, line 127: Might as well do a single DEPS entry for `src/remote/`.

dgozman (approved):

utils/check_deps.js
JoelEinbinder: The client and server aren't separate? Or we don't care about the distinction?

dgozman: I think once we do care about the distinction, we'll put them into separate folders.

PR #4551, opened 2020-11-30
fix(rpc): collect network requests and responses on frame navigation

dgozman:
To avoid unbounded growth when reusing a single Page instance.

### DIFF
--- a/test/channels.spec.ts
+++ b/test/channels.spec.ts
@@ -17,7 +17,7 @@
 
 import domain from 'domain';
 import { folio } from './fixtures';
-import type { ChromiumBrowser } from '..';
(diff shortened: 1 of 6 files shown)


test/channels.spec.ts
```diff
@@ -218,3 +254,17 @@ function trimGuids(object) {
     return object ? object.match(/[^@]+/)[0] : '';
   return object;
 }
+
+function countType(object, type) {
+  if (Array.isArray(object))
```
yury-s, line 259: Why not simply look at `_guid` and `.objects` array, can the scope ever have different structure?

yury-s (approved):

dgozman:
This solution introduces too much complexity, we'll be looking for something better.

PR #4572, opened 2020-12-02
fix(doclint): exit 1 when doclint throws an error

JoelEinbinder:


utils/doclint/cli.js
```diff
@@ -31,7 +31,10 @@ const RED_COLOR = '\x1b[31m';
 const YELLOW_COLOR = '\x1b[33m';
 const RESET_COLOR = '\x1b[0m';
 
-run();
+run().catch(e => {
+  console.error(e && e.stack ? e.stack : e);
```
mxschmitt, line 35: doesn't `console.error(new Error())` already print the stack automatically to the console?

JoelEinbinder: yep, I've spent too much time with jest's serializable errors.

yury-s (approved):

PR #4573, opened 2020-12-02
feat(selectors): introduce css evaluator

dgozman:
Not used for production yet.

### DIFF
--- a/src/debug/injected/debugScript.webpack.config.js
+++ b/src/debug/injected/debugScript.webpack.config.js
@@ -23,7 +23,7 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.tsx?$/,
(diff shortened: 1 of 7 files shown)


src/debug/injected/debugScript.webpack.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.tsx?$/,
+        test: /\.(j|t)sx?$/,
```
JoelEinbinder, line 26: where is the js file?

dgozman: cssTokenizer.js

aslushnikov (approved):

PR #4577, opened 2020-12-03
devops(win): fix paths to vswhere.exe in 64-bit shells

aslushnikov:
The `$PROGRAMFILES` variable is defined per the bitness of the
application: 32-bit shells get `C:\Program Files (x86)`, whereas
64-bit shells get `C:\Program Files`.

Visual Studio, however, has only 32-bit build and is always located in
the `C:\Program Files (x86)` folder.

This patch fixes `//browser_patches/firefox/build.sh` on 64-bit windows
shell.


yury-s (approved):
we could also try %programfiles(x86)%, I wonder why it's not working on some systems though

PR #4566, opened 2020-12-01
chore: add internal method for utility context bindings

JoelEinbinder:
This only works in chromium, throws for other browsers. Split from #4519.

PR #4598, opened 2020-12-04
chore: fix internal binding

JoelEinbinder:
follow up to #4566 . The binding was being evaluated in the wrong context, and the api was missing on the page.


dgozman (approved):
Perhaps it makes sense to add a test? The easiest way is to probably add non-wire test against implementation.

PR #4603, opened 2020-12-04
devops: force LF line endings for all text files

aslushnikov:


yury-s (approved):

.gitattributes
```diff
@@ -1,3 +1,5 @@
-# text files must be lf for golden file tests to work
-*.txt eol=lf
-*.json eol=lf
+# text files must be lf for golden file tests to work and for browser
+# patches to apply.
+#
+# In general, it seems to be safe to default to lf for all text files.
+* text=auto eol=lf
```
yury-s, line 5: I assume it won't affect binary files.

aslushnikov: it claims to keep them intact; I'm not yet sure this would be needed though.

Thank you for the review!

aslushnikov:
I think we can wait with this.

PR #4609, opened 2020-12-05
fix: #4610

### LINKED ISSUE #4610: [Question] Is there a way to access the `playwright` object in dev tools in browsers launched within `launch()`?
Hi team,

I'm launching a browser in this way:
```python3
with sync_playwright() as playwright:
    browser = playwright.chromium.launch(headless=False)
    ...
```
In this browser, is there a way to access the `playwright` object in the Developer Tools Console, similarly to what you can do when launching a browser using `open`, e.g.,
```bash
python -m playwright open http://localhost:3000
```
and as described in [you documentation](https://github.com/microsoft/playwright-cli#inspect-selectors)? I think it could be useful to debug issues, for instance when a selector cannot find the given element because of a recent change and the flow gets stuck, without the need to manually reproduce the page state generated by the Python code after `launch`.

Alternatively, what could be a good workaround? For instance, I'm wondering if it's possible in the Developer Tools Console to do something similar to
```javascript
playwright.selector($0)
```
when debugging in a browser started with `launch`. 

I tried launching the browser with
```python
playwright.chromium.launch(headless=False, env={"PWDEBUG": True})
```
but even assuming this is the proper way, I get an error—I can give you more details on this, if you want.

Thanks!

kumaraditya303:
Fix #4610


pavelfeldman:
I assume you intend to use this code from playwright-cli, but that requires additional changes.

kumaraditya303:
> I assume you intend to use this code from playwright-cli, but that requires additional changes.

@pavelfeldman It would be helpful if you could specify what additional changes are required.

pavelfeldman:
I can't tell on top of my head, but if you try reusing it from the playwright-cli, you would know!

kumaraditya303:
> I can't tell on top of my head, but if you try reusing it from the playwright-cli, you would know!

@pavelfeldman I didn't noticed any additional changes required yet.

PR #4614, opened 2020-12-07
feat(selectors): proximity selectors

dgozman:
This includes 'left-of', 'right-of', 'above', 'below', 'near' and 'within'.

### DIFF
--- a/src/server/injected/selectorEvaluator.ts
+++ b/src/server/injected/selectorEvaluator.ts
@@ -52,6 +52,12 @@ export class SelectorEvaluatorImpl implements SelectorEvaluator {
     this._engines.set('xpath', xpathEngine);
     for (const attr of ['id', 'data-testid', 'data-test-id', 'data-test'])
       this._engines.set(attr, createAttributeEngine(attr));
+    this._engines.set('right-of', createProximityEngine('right-of', boxRightOf));
+    this._engines.set('left-of', createProximityEngine('left-of', boxLeftOf));
+    this._engines.set('above', createProximityEngine('above', boxAbove));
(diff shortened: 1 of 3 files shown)


src/server/injected/selectorEvaluator.ts
```diff
@@ -442,6 +448,45 @@ function createAttributeEngine(attr: string): SelectorEngine {
   };
 }
 
+function areCloseRanges(from1: number, to1: number, from2: number, to2: number) {
+  return to1 >= from2 - 50 && to2 >= from1 - 50;
```
aslushnikov, line 452: I'm not sure I follow this condition.

test/selectors-misc.spec.ts
```diff
@@ -75,3 +75,96 @@ it('should work with :visible', async ({page}) => {
...
+      div.style.height = box[3] + 'px';
+      container.appendChild(div);
+    }
+  }, boxes);
+
+  expect(await page.$eval('div:within(#id0)', e => e.id)).toBe('id6');
+  expect(await page.$eval('div:within(div)', e => e.id)).toBe('id6');
+  expect(await page.$('div:within(#id6)')).toBe(null);
+
+  expect(await page.$eval('div:right-of(#id6)', e => e.id)).toBe('id3');
```
aslushnikov, line 141: can we have `$$eval` here as well?

Will it result in all rects that are to the right?

aslushnikov (approved):

test/selectors-misc.spec.ts
dgozman: Done.

PR #4632, opened 2020-12-08
devops: Separate base image from NodeJS image

kumaraditya303:
Details -
> - Discussion - https://github.com/microsoft/playwright-java/pull/91#discussion_r537801716 

Separating base image from NodeJS has following benefits-
- Central dependency management for browsers dependencies.
- Reduces code duplication
- Every related playwright project can build its image on top of it 
- Easier for users to build custom images .
@mxschmitt @pavelfeldman ^^

### DIFF
--- a/.github/workflows/publish_canary_docker.yml
+++ b/.github/workflows/publish_canary_docker.yml
@@ -33,8 +33,10 @@ jobs:
     - run: ./docs/docker/build.sh focal playwright:localbuild-focal
     - name: tag & publish
       run: |
-        ./docs/docker/tag_and_push.sh playwright:localbuild-bionic playwright.azurecr.io/public/playwright:next
         ./docs/docker/tag_and_push.sh playwright:localbuild-bionic playwright.azurecr.io/public/playwright:next-bionic
         ./docs/docker/tag_and_push.sh playwright:localbuild-bionic playwright.azurecr.io/public/playwright:sha-${{ github.sha }}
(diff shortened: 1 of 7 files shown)


kumaraditya303:
With this change we build two base images `playwright:focal-base` and `playwright:bionic-base` as base images with only the browser dependencies and the rest are built on top of them!

aslushnikov:
Hey @kumaraditya303,

Both you and Yury sent very [similar PRs](https://github.com/microsoft/playwright/pull/4634) :-) We discussed this approach offline with @yury-s.

It looks like using the suggested `playwright-bionic-base` from-inside playwright-java repository will not be convenient. Since downstream has no idea when upstream base image is published and updated, you'll have to:
- tag the base image with SHA
- inherit downstream from this SHA
- and trigger the downstream docker build manually whenever the base is published to the container registry.

While this process might easy some pain in managing the list of browser dependencies, I don't think it's a worthy trade-off since browser dependencies don't actually change that much.

So it looks like there are two approaches here: 
1. split docker file into layers like you did it here, and co-locate all docker files for all languages in the same repository. This way you can reliably build & publish all the images.
2. copy docker files to downstream repos

Since (1) is somewhat involved, let's stick with (2) for now since it seems to be a more light-weight approach. We can also add some tests to make sure docker image is functional - there are already some at https://devops.aslushnikov.com.

With this I'll close both this and Yury's PRs.

PR #4634, opened 2020-12-08
devops: extract base docker image

yury-s:
This was base image with browser dependencies could be reused in Dockerfiles for different language bindings without repeating the deps installation steps in each of them.


yury-s:
Discussed this with @aslushnikov offline. Looking at the complexity of the separate base image and associated labels business we are not sure it is justified at this point. It's easier to simply maintain a copy of the deps in each repo for now, they don't change that often.  If it becomes a problem we will revisit the decision.

PR #4637, opened 2020-12-08
chore(docker): put browser deps instructions first

yury-s:
The dependencies part should be common between language bindings and this way it should be easier to compare.

### DIFF
--- a/docs/docker/Dockerfile.bionic
+++ b/docs/docker/Dockerfile.bionic
@@ -1,11 +1,8 @@
 FROM ubuntu:bionic
 
-# 1. Install node14
-RUN apt-get update && apt-get install -y curl && \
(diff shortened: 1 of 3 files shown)


aslushnikov (approved):

mxschmitt (commented):
Let's maybe also put the `# 3. Symlink downloaded browsers for root user` up or would it break since the folders does not exist yet?

yury-s:
> Let's maybe also put the `# 3. Symlink downloaded browsers for root user` up or would it break since the folders does not exist yet?

Yeah, it currently placed right after the instruction that creates the folder. I'm not fond of creating such link if the target directory won't be created. In any case we can change that in a separate PR, my main goal is to ensure that the Dockerfiles in language bindings have same prefix as currently the first RUN instruction will differ between typescript, python and java.

Issue #1122, opened 2020-02-26
[Question] Is it possible to run Playwright tests on real mobile device browsers?


Cosmin-Gramada:
Is there any way to connect to a device, start a browser (e.g. chromium instance on Android) and run the tests?

pavelfeldman:
There is no way to attach to Chromium on Android or MobileSafari on iOS. Playwright aims at running on your desktop and in your CI/CD, targeting device farms is outside its scope.

We will emulate device metrics, geolocation, orientation, etc on Chromium on Android and MobileSafari though. Could you share more about your use case?

Cosmin-Gramada:
We took into account the possibility of moving away from Selenium WebDriver based tools and switching to DevTools API based tools.

A really big portion of our tests are running on mobile web browsers, where we tests our own player that makes use of browser related technologies (e.g. what types of medias is able to render, what's the appropriate format, etc) and sends specific tracking events.

Our concern is that there will be differences between a real device and an emulated one, as we're able to emulate any device resolution, but only the viewport size is guaranteed to be accurate.

pavelfeldman:
> Our concern is that there will be differences between a real device and an emulated one

Yes, there will be

> as we're able to emulate any device resolution, but only the viewport size is guaranteed to be accurate

More than the viewport size, but yes, if you need mobile-specific media to be accurate, you need to attach to the device.

This is technically feasible, but is outside of the scope for v1.

tanaypratap:
This could be useful for us too. We are a fully fledged PWA email app. Testing on actual devices would catch bugs. We have also faced this issue with non-existent media queries on some phones which @Cosmin-Gramada mentioned. Especially older phones.

Filipoliko:
Since Playwright is trying to be more testing-friendly, it would be really great to have real mobile device support. More and more users are coming to our websites from mobile devices, it is nearly 50:50 in my case. Most of our bugs on mobile devices are not discoverable via simple mobile emulation. It is really important for us to be able to test on real mobile devices, or at least on emulated Android/iOS systems.

arjun27:
@Filipoliko - thank you for sharing. Can you elaborate on which bugs are not discoverable via emulation? Examples will really help!

Filipoliko:
[...]

PR #4657, opened 2020-12-09
feat(adb): support webviews

pavelfeldman:
#1122

### DIFF
--- a/android-types-internal.d.ts
+++ b/android-types-internal.d.ts
@@ -14,6 +14,52 @@
  * limitations under the License.
  */
 
+import { EventEmitter } from 'events';
+
+export interface AndroidDevice<BrowserContextOptions, BrowserContext, Page> extends EventEmitter {
+  input: AndroidInput;
+
+  setDefaultTimeout(timeout: number): void;
+  on(event: 'webview', handler: (webView: AndroidWebView<Page>) => void): this;
+  waitForEvent(event: string, predicate?: (data: any) => boolean): Promise<any>;
+
+  serial(): string;
+  model(): string;
+  webViews(): AndroidWebView<Page>[];
+  shell(command: string): Promise<string>;
+  launchBrowser(options?: BrowserContextOptions & { packageName?: string  }): Promise<BrowserContext>;
+  close(): Promise<void>;
+
+  wait(selector: AndroidSelector, options?: { state?: 'gone' } & { timeout?: number }): Promise<void>;
+  fill(selector: AndroidSelector, text: string, options?: { timeout?: number }): Promise<void>;
+  press(selector: AndroidSelector, key: AndroidKey, options?: { duration?: number } & { timeout?: number }): Promise<void>;
+  tap(selector: AndroidSelector, options?: { duration?: number } & { timeout?: number }): Promise<void>;
+  drag(selector: AndroidSelector, dest: { x: number, y: number }, options?: { speed?: number } & { timeout?: number }): Promise<void>;
+  fling(selector: AndroidSelector, direction: 'down' | 'up' | 'left' | 'right', options?: { speed?: number } & { timeout?: number }): Promise<void>;
+  longTap(selector: AndroidSelector, options?: { timeout?: number }): Promise<void>;
+  pinchClose(selector: AndroidSelector, percent: number, options?: { speed?: number } & { timeout?: number }): Promise<void>;
+  pinchOpen(selector: AndroidSelector, percent: number, options?: { speed?: number } & { timeout?: number }): Promise<void>;
+  scroll(selector: AndroidSelector,  direction: 'down' | 'up' | 'left' | 'right', percent: number, options?: { speed?: number } & { timeout?: number }): Promise<void>;
+  swipe(selector: AndroidSelector, direction: 'down' | 'up' | 'left' | 'right', percent: number, options?: { speed?: number } & { timeout?: number }): Promise<void>;
+
+  info(selector: AndroidSelector): Promise<AndroidElementInfo>;
+}
+
+export interface AndroidInput {
+  type(text: string): Promise<void>;
+  press(key: AndroidKey): Promise<void>;
+  tap(point: { x: number, y: number }): Promise<void>;
+  swipe(from: { x: number, y: number }, segments: { x: number, y: number }[], steps: number): Promise<void>;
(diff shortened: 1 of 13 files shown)


yury-s (approved):

android-types-internal.d.ts
```diff
@@ -14,6 +14,51 @@
...
  */
 
+import { EventEmitter } from 'events';
+
+export interface AndroidDevice<BrowserContextOptions, BrowserContext, Page> extends EventEmitter {
+  input: AndroidInput;
+
+  setDefaultTimeout(timeout: number): void;
+  on(event: 'webviewadded', handler: (webView: AndroidWebView<Page>) => void): this;
+  on(event: 'webviewremoved', handler: (webView: AndroidWebView<Page>) => void): this;
```
dgozman, line 24: Implementation seems to have 'close' event on WebView instead.

src/client/android.ts
```diff
@@ -72,6 +115,11 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
     });
   }
 
+  async press(selector: apiInternal.AndroidSelector, key: apiInternal.AndroidKey, options?: types.TimeoutOptions) {
+    await this.tap(selector, options);
+    await this.input.press(key);
```
dgozman, line 120: This ignores the timeout.

src/client/android.ts
```diff
@@ -146,6 +195,17 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
       return BrowserContext.from(context);
     });
   }
+
+  async waitForEvent(event: string, optionsOrPredicate: types.WaitForEventOptions = {}): Promise<any> {
+    const timeout = this._timeoutSettings.timeout(typeof optionsOrPredicate === 'function' ? {} : optionsOrPredicate);
+    const predicate = typeof optionsOrPredicate === 'function' ? optionsOrPredicate : optionsOrPredicate.predicate;
+    const waiter = new Waiter();
+    waiter.rejectOnTimeout(timeout, `Timeout while waiting for event "${event}"`);
+    waiter.rejectOnEvent(this, Events.AndroidDevice.Close, new Error('Device closed'));
```
dgozman, line 204: if (event !== Events.AndroidDevice.Close)...

src/client/android.ts
```diff
@@ -129,6 +177,7 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
   async close() {
     return this._wrapApiCall('androidDevice.close', async () => {
       await this._channel.close();
+      this.emit(Events.AndroidDevice.Close);
```
dgozman, line 180: I wonder how this works if I disconnect the cable?

src/client/events.ts
```diff
@@ -16,6 +16,15 @@
  */
 
 export const Events = {
+  AndroidDevice: {
+    WebView: 'webview',
```
dgozman, line 20: This is named 'webviewadded' in types.

src/dispatchers/androidDispatcher.ts
```diff
@@ -30,14 +30,21 @@ export class AndroidDispatcher extends Dispatcher<Android, channels.AndroidIniti
...
+  }
 }
 
 export class AndroidDeviceDispatcher extends Dispatcher<AndroidDevice, channels.AndroidDeviceInitializer> implements channels.AndroidDeviceChannel {
   constructor(scope: DispatcherScope, device: AndroidDevice) {
     super(scope, device, 'AndroidDevice', {
       model: device.model,
-      serial: device.serial
+      serial: device.serial,
+      webViews: device.webViews()
```
dgozman, line 44: If you do this, webviews will not be cleaned up together with AndrodiDevice, because they end up in the parent's scope. I'd suggest to emit 'webViewAdded' for all initial webviews.

src/server/android/android.ts
```diff
@@ -150,20 +183,26 @@ export class AndroidDevice {
   }
 
   async close() {
-    const driver = await this._driver();
-    driver.close();
+    if (this._pollingWebViews)
```
dgozman, line 186: We should cleanup stuff like this on disconnect.

src/server/android/android.ts
```diff
@@ -51,35 +53,64 @@ export interface SocketBackend extends EventEmitter {
...
     const model = await backend.runCommand('shell:getprop ro.product.model');
-    return new AndroidDevice(backend, model);
+    const device = new AndroidDevice(android, backend, model.trim());
+    await device._init();
+    return device;
+  }
+
+  async _init() {
+    await this._refreshWebViews();
+    this._pollingWebViews = setInterval(async () => this._refreshWebViews().catch(() => {}), 500);
```
dgozman, line 109: I think it's better to employ the Throttler pattern, just in case refreshing web views takes a long time.

src/server/android/android.ts
```diff
@@ -173,8 +212,20 @@ export class AndroidDevice {
...
     }
     debug('pw:android')('Got the socket, connecting');
-    const androidBrowser = new AndroidBrowser(this, packageName, socketName);
+    return await this._connectToBrowser(socketName, options);
+  }
+
+  connectToWebView(pid: number): Promise<BrowserContext> {
+    const webView = this._webViews.get(pid);
+    if (!webView)
+      throw new Error('WebView with given pid does not exist');
```
dgozman, line 221: Let's throw "WebView has been closed", because that's what will happen in real scenarios.

src/server/android/android.ts
```diff
@@ -195,6 +246,49 @@ export class AndroidDevice {
...
+      if (!match)
+        continue;
+      const pid = +match[1];
+      newPids.add(pid);
+    }
+    for (const pid of newPids) {
+      if (this._webViews.has(pid))
+        continue;
+
+      const procs = await (await this._backend.runCommand(`shell:ps -A | grep ${pid}`)).split('\n');
```
dgozman, line 269: Double await for fun!

dgozman (approved):

src/server/android/android.ts
pavelfeldman: VS Code!

src/server/android/android.ts
pavelfeldman: Yeah, will follow up.

src/client/android.ts
pavelfeldman: we'll figure it with time...

PR #4668, opened 2020-12-10
test: fix test server on Node 15

aslushnikov:
This patch fixes a bug in our test server that manifests itself in
Node 15.

Context: Node 14 does not support Apple Silicon (and probably will not),
so we currently have to run tests on Node 15 on new macs.

### DIFF
--- a/utils/testserver/index.js
+++ b/utils/testserver/index.js
@@ -272,8 +272,6 @@ class TestServer {
       response.end(`File not found: ${filePath}`);
       return;
     }
-    // after async hop, the request might've been already aborted, so attempting to change response
-    // will throw.
     const extension = filePath.substring(filePath.lastIndexOf('.') + 1);
     const mimeType = extensionToMime[extension] || 'application/octet-stream';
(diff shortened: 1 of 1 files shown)


utils/testserver/index.js
```diff
@@ -258,27 +263,31 @@ class TestServer {
...
+    const {err, data} = await readFileAsync(filePath).then(data => ({data})).catch(err => ({err}));
+    // The HTTP transaction might be already terminated after async hop here - do nothing in this case.
+    if (response.writableEnded)
+      return;
+    if (err) {
+      response.statusCode = 404;
+      response.end(`File not found: ${filePath}`);
+      return;
+    }
+    // after async hop, the request might've been already aborted, so attempting to change response
```
yury-s, line 275: Should this comment be placed next to the `writableEnded` check above?

yury-s (approved):

utils/testserver/index.js
aslushnikov: thanks! stray comment

PR #4680, opened 2020-12-11
chore: refactor matchesParents and add some comments

dgozman:


dgozman:
The old version was better.

PR #4689, opened 2020-12-11
feat(android): expose installAPK(path) and ADB socket

mxschmitt:
https://github.com/microsoft/playwright/issues/1122


pavelfeldman:
Few things:
- installApk, accept buffer and filename
- we don't read from filesystem on the server side, should be client
- you pass parameters unconditionally, but it is typical to pass different args while installing
- users might benefit from the actual open, so i wonder if it makes sense plumbing the connection object itself.

src/client/android.ts
```diff
@@ -189,14 +193,31 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
...
+    });
+  }
+
+  installApk(file: string | Buffer): Promise<void> {
+    return this._wrapApiCall('androidDevice.installApk', async () => {
+      const normalisedApk = await convertApkFile(file);
+      const adbSocket = await this.open(`shell:cmd package install -r -t -S ${normalisedApk.length}`);
+      debug('pw:android')('Writing driver bytes: ' + normalisedApk.length);
+      await adbSocket.write(normalisedApk);
+      const success = await new Promise(f => adbSocket.on('data', f));
```
pavelfeldman, line 214: This response can be "Failed"

src/client/android.ts
```diff
@@ -217,6 +238,30 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
...
+    return (androidDevice as any)._object;
+  }
+
+  constructor(parent: ChannelOwner, type: string, guid: string, initializer: channels.AndroidAdbSocketInitializer) {
+    super(parent, type, guid, initializer);
+    this._channel.on('data', ({ data }) => this.emit('data', data)); // TODO(mxschmitt): use constant
+  }
+
+  async write(data: Buffer): Promise<void> {
+    return this._wrapApiCall('androidDevice.inputType', async () => {
```
pavelfeldman, line 253: androidDevice.write

src/client/android.ts
```diff
@@ -217,6 +238,30 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
...
+      await this._channel.write({ data: data.toString('base64') });
+    });
+  }
+}
+
+async function convertApkFile(file: string | Buffer): Promise<Buffer> {
```
pavelfeldman, line 259: nit: readApkFile

src/client/android.ts
```diff
@@ -217,6 +238,30 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
...
 
+export class AndroidAdbSocket extends ChannelOwner<channels.AndroidAdbSocketChannel, channels.AndroidAdbSocketInitializer> {
+
+  static from(androidDevice: channels.AndroidAdbSocketChannel): AndroidAdbSocket {
+    return (androidDevice as any)._object;
+  }
+
+  constructor(parent: ChannelOwner, type: string, guid: string, initializer: channels.AndroidAdbSocketInitializer) {
+    super(parent, type, guid, initializer);
+    this._channel.on('data', ({ data }) => this.emit('data', data)); // TODO(mxschmitt): use constant
```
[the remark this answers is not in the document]
pavelfeldman: You are emitting a string, not a buffer.

src/client/connection.ts
```diff
@@ -154,6 +154,9 @@ export class Connection {
       case 'AndroidDevice':
         result = new AndroidDevice(parent, type, guid, initializer);
         break;
+      case 'AndroidAdbSocket':
```
pavelfeldman, line 157: That's not how sorting works :)

src/dispatchers/androidDispatcher.ts
```diff
@@ -147,6 +156,15 @@ export class AndroidDeviceDispatcher extends Dispatcher<AndroidDevice, channels.
   }
 }
 
+export class AndroidAdbSocketDispatcher extends Dispatcher<AndroidAdbSocket, channels.AndroidAdbSocketInitializer> implements channels.AndroidAdbSocketChannel {
+  constructor(scope: DispatcherScope, socket: AndroidAdbSocket) {
+    super(scope, socket, 'AndroidAdbSocket', {}, true);
```
pavelfeldman, line 161: I would think that you listen to the socket.

src/protocol/protocol.yml
```diff
@@ -2083,6 +2083,19 @@ Android:
...
         timeout: number
 
+AndroidAdbSocket:
+  type: interface
+
+  commands:
+
+    write:
+      parameters:
+        data: string
```
pavelfeldman, line 2093: data: binary

src/protocol/protocol.yml
```diff
@@ -2083,6 +2083,19 @@ Android:
...
+
+  events:
+    data:
+      parameters:
+        data: string
```
pavelfeldman, line 2098: data: binary

src/protocol/protocol.yml
```diff
@@ -2275,12 +2288,22 @@ AndroidDevice:
...
+      returns:
+        socket: AndroidAdbSocket
+
     shell:
       parameters:
         command: string
       returns:
         result: string
 
+    installApk:
```
pavelfeldman, line 2303: Looks like you don't need it.

src/server/android/android.ts
```diff
@@ -297,6 +295,25 @@ export class AndroidDevice extends EventEmitter {
   }
 }
 
+export class AndroidAdbSocket extends EventEmitter {
```
pavelfeldman, line 298: I don't think you need this class, just use backend socket.

src/server/android/android.ts
```diff
@@ -338,7 +355,7 @@ Sec-WebSocket-Version: 13\r
     await new Promise(f => this._socket!.once('data', f));
 
     // Start sending web frame to receiver.
-    this._socket.on('data', data => this._receiver._write(data, 'binary', () => {}));
+    this._socket.on('data', data => this._receiver._write(data, 'binary', () => { }));
```
pavelfeldman, line 358: Revert

src/server/android/android.ts
mxschmitt: Do you mean to move the logic into the BufferedSocketWrapper class since it implements the SocketBackend interface?

pavelfeldman: What logic?

android-types-internal.d.ts
```diff
@@ -46,6 +48,10 @@ export interface AndroidDevice<BrowserContextOptions, BrowserContext, Page> exte
   info(selector: AndroidSelector): Promise<AndroidElementInfo>;
 }
 
+export interface AndroidSocket extends EventEmitter {
+  write(data: Buffer): Promise<void>
```
pavelfeldman, line 52: declare on('data') here as well (see other examples)

src/dispatchers/androidDispatcher.ts
```diff
@@ -147,6 +157,17 @@ export class AndroidDeviceDispatcher extends Dispatcher<AndroidDevice, channels.
...
+export class AndroidSocketDispatcher extends Dispatcher<SocketBackend, channels.AndroidSocketInitializer> implements channels.AndroidSocketChannel {
+  constructor(scope: DispatcherScope, socket: SocketBackend) {
+    super(scope, socket, 'AndroidSocket', {}, true);
+    socket.on(Events.AndroidSocket.Data, (data: string) => this._dispatchEvent('data', { data }));
+  }
+
+  async write(params: channels.AndroidSocketWriteParams, metadata?: channels.Metadata): Promise<void> {
+    await this._object.write(Buffer.from(params.data, 'base64'));
+  }
+}
```
pavelfeldman, line 169: We are missing `close()`, every socket should have one.

src/server/android/backendAdb.ts
```diff
@@ -111,7 +111,7 @@ class BufferedSocketWrapper extends EventEmitter implements SocketBackend {
     this._socket.on('data', data => {
       debug('pw:adb:data')(data.toString());
       if (this._isSocket) {
-        this.emit('data', data);
+        this.emit('data', data.toString('base64'));
```
pavelfeldman, line 114: This doesn't sound right, my socket is a buffer socket.

src/dispatchers/androidDispatcher.ts
```diff
@@ -147,6 +157,17 @@ export class AndroidDeviceDispatcher extends Dispatcher<AndroidDevice, channels.
   }
 }
 
+export class AndroidSocketDispatcher extends Dispatcher<SocketBackend, channels.AndroidSocketInitializer> implements channels.AndroidSocketChannel {
+  constructor(scope: DispatcherScope, socket: SocketBackend) {
+    super(scope, socket, 'AndroidSocket', {}, true);
+    socket.on(Events.AndroidSocket.Data, (data: string) => this._dispatchEvent('data', { data }));
```
pavelfeldman, line 163: `{ data: data.toString('base64') }`

src/server/android/backendAdb.ts
mxschmitt: good catch, fixed.

src/dispatchers/androidDispatcher.ts
mxschmitt: good catch.

pavelfeldman (approved):

pavelfeldman:
There are tiny nits that we can follow up on.

PR #4697, opened 2020-12-13
feat(adb): implement push

pavelfeldman:
https://github.com/microsoft/playwright/issues/1122


mxschmitt (approved):
lets follow up with a test for it once the the android bots are merged 👍

Issue #4683, opened 2020-12-11
[Question] Ubuntu 20.04 instructions for browser dependencies


thernstig:
According to https://github.com/microsoft/playwright/blob/master/docs/troubleshooting.md I need to install these dependencies:

<details>
  <summary>Debian (e.g. Ubuntu) Dependencies</summary>

```
gconf-service
libasound2
libatk1.0-0
libatk-bridge2.0-0
libc6
libcairo2
libcups2
libdbus-1-3
libexpat1
libfontconfig1
libgcc1
libgconf-2-4
libgdk-pixbuf2.0-0
libglib2.0-0
libgtk-3-0
libnspr4
libpango-1.0-0
libpangocairo-1.0-0
libstdc++6
libx11-6
libx11-xcb1
libxcb1
libxcomposite1
libxcursor1
libxdamage1
libxext6
libxfixes3
libxi6
libxrandr2
libxrender1
libxss1
libxtst6
ca-certificates
fonts-liberation
libappindicator1
libnss3
lsb-release
xdg-utils
wget
libgbm1
```

</details>

But the text also says:

>  For dependencies on Ubuntu, please refer to [Dockerfile ](https://github.com/microsoft/playwright/blob/master/docs/docker/Dockerfile.bionic)which is used to run our tests

**Questions**

1. It is slightly confusing that it first says to read the Dockerfile and then also still lists the dependencies for Ubuntu. Maybe only one of them should exist, or the text rephrased to say something like that you can _also_ see the dependencies in the Dockerfile (or similar)?

2. The link to the Dockerfile goes to Ubuntu 18.04 (bionic) whilst there is also one for Ubuntu 20.04 (focal) https://github.com/microsoft/playwright/blob/master/docs/docker/Dockerfile.focal - that the link only points to one confuses users if they are using Ubuntu 20.04. Should a link to both exist on that page, or just a general link to https://github.com/microsoft/playwright/tree/master/docs/docker ?

3. https://github.com/microsoft/playwright/tree/master/docs/docker and the first paragraph starts with:
> Dockerfile.bionic can be used to run Playwright scripts in Docker environments.

I assume it would be good to also mention `focal` here?

4. When I run Playwright in a custom Dockerfile built on Ubuntu 20.04 where I forgot to install many of the dependencies I get:
```
Error: Host system is missing dependencies!

  Install missing packages with:
      sudo apt-get install libnss3\
          libnspr4\
          libatk1.0-0\
[...]

PR #4706, opened 2020-12-14
docs: remove outdated troubleshooting

aslushnikov:
Most of the troubleshooting.md doc is not relevant any more:
- dependencies are maintained and helped with by launch doctor
- sandbox is disabled in Chromium by default

References #4683


arjunattam (approved):

mxschmitt (approved):

mxschmitt:
Maybe for the future, if not already possible there should be an example/guide that it works under WSL.

arjunattam:
@mxschmitt - I think the launch doctor would show missing dependencies/required config. Is there a case we're missing?

mxschmitt:
> @mxschmitt - I think the launch doctor would show missing dependencies/required config. Is there a case we're missing?

I unfortunately don't have a Windows machine but remember users who had issues, not sure. Let's wait for incoming issues. 👍

PR #4711, opened 2020-12-14
chore(tests): add Android as a testing target

mxschmitt:
Now we can run existing tests on Android, example:

`PWMODE=android npx folio test/accessibility.spec.ts  -p browserName=chromium --workers=1 --timeout=60000 --global-timeout=5400000 --retries=0 --reporter=list`


test/android/android.fixtures.ts
```diff
@@ -32,4 +34,11 @@ fixtures.device.init(async ({ playwright }, runTest) => {
   await device.close();
 });
 
+fixtures.page.override(async ({ device }, runTest) => {
+  await device.shell('am start org.chromium.webview_shell/.WebViewBrowserActivity');
```
pavelfeldman, line 38: That's a good idea, but we would also want to test browser. Should we run this app only once? We can later add our own testing apk that gives new webview for each test...

test/http.fixtures.ts
```diff
@@ -35,11 +35,11 @@ fixtures.httpService.init(async ({ testWorkerIndex }, test) => {
   const assetsPath = path.join(__dirname, 'assets');
   const cachedPath = path.join(__dirname, 'assets', 'cached');
 
-  const port = 8907 + testWorkerIndex * 2;
+  const port = process.env.PWMODE ? 8180 : (8907 + testWorkerIndex * 2);
```
pavelfeldman, line 38: We can extract mode to be a root fixture this depends on.

test/playwright.fixtures.ts
```diff
@@ -114,6 +115,10 @@ fixtures.browserType.init(async ({ playwright, browserName }, run) => {
 }, { scope: 'worker' });
 
 fixtures.browser.init(async ({ browserType, browserOptions }, run) => {
+  if (process.env.PWMODE === 'android') {
```
pavelfeldman, line 118: you should be able to use `mode` here, also you can probably throw?

test/playwright.fixtures.ts
```diff
@@ -153,7 +158,16 @@ fixtures.contextOptions.init(async ({ video, testInfo }, run) => {
...
+fixtures.androidContext.init(async ({ playwright }, run) => {
+  if (process.env.PWMODE !== 'android')
+    return;
+  const [device] = await playwright._android.devices();
+  const context = await device.launchBrowser();
+  await run(context);
+  await context.close();
+}, { scope: 'worker' });
+
+fixtures.contextFactory.init(async ({ browser, contextOptions, testInfo, screenshotOnFailure, androidContext }, run) => {
```
pavelfeldman, line 170: androidContext unused?

test/android/android.fixtures.ts
mxschmitt: yup reverted for now. Lets focus in this PR only that we can run existing tests no Android.

test/playwright.fixtures.ts
mxschmitt: The Playwright fixtures should be quite in-depended since they get used in playwright-test, so I think its better to do it like that with env vars.

PR #4730, opened 2020-12-15
fix(adb): leaking socket connections

mxschmitt:


src/server/android/backendAdb.ts
```diff
@@ -68,11 +68,15 @@ async function runCommand(command: string, serial?: string): Promise<Buffer> {
...
+  let commandOutput: Buffer;
   if (!command.startsWith('shell:')) {
     const remainingLength = parseInt((await socket.read(4)).toString(), 16);
-    return (await socket.read(remainingLength));
+    commandOutput = await socket.read(remainingLength);
+  } else {
+    commandOutput = await socket.readAll();
   }
-  return (await socket.readAll());
+  await socket.close();
```
pavelfeldman, line 78: I'm surprised those are not closed by the remote party, is that not happening? Would be nice to dispose AndroidSocketDispatcher upon socket close. I remembered that adb was closing the sockets on their end, so I did not care to do that. I might be remembering wrong. If you dispose the dispatcher, you would be able to write a test for it

pavelfeldman (approved):

PR #4736, opened 2020-12-16
chore(tests): restructure and optimise test files

mxschmitt:
Extracted for #4711. The goal is to run the Android tests on the `test/page*.ts` files since these are files which don't require a browser, browserType instance or call launch or newContext.


pavelfeldman (approved):

dgozman:
@mxschmitt This PR made two tests fail consistently on Chromium video bot, both changes in the PR. Could you please take a look?

PR #4747, opened 2020-12-16
test: fixed failing test on video bot

mxschmitt:
Caused by #4736 reverted my change.

@dgozman  (https://github.com/microsoft/playwright/pull/4736#issuecomment-747105121)

### DIFF
--- a/test/page-event-popup.spec.ts
+++ b/test/page-event-popup.spec.ts
@@ -35,7 +35,9 @@ it('should work with window features', async ({page, server}) => {
   expect(await popup.evaluate(() => !!window.opener)).toBe(true);
 });
 
-it('should emit for immediately closed popups', async ({page}) => {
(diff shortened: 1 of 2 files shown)


dgozman (approved):
I think I also saw `should be able to capture alert` failing.

mxschmitt:
> I think I also saw `should be able to capture alert` failing.

do you have an idea why this change needs to be reverted? Weird behaviour for me, even that this is passing with the normal tests.

dgozman:
> do you have an idea why this change needs to be reverted? Weird behaviour for me, even that this is passing with the normal tests.

No, not really. Probably needs investigation.

dgozman:
Should we land this?

mxschmitt:
> Should we land this?

Would say yes, reverted the relevant changes which I did.

PR #4751, opened 2020-12-17
feat(a11y): Integrate Axe into Playwright

avodovnik:
The goal of this PR is to get [Axe Core](https://github.com/dequelabs/axe-core) integrated with Playwright, to enable easy accessibility auditing for users.


JoelEinbinder:
Thanks for the PR! We are probably not going to get to reviewing this until after the holidays. @avodovnik

pavelfeldman:
Does not look like something we plan to land

PR #4753, opened 2020-12-17
build: Cache Dependencies & Browsers for faster CI

kumaraditya303:


.github/workflows/infra.yml
```diff
@@ -1,25 +1,33 @@
...
-    steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-node@v1
-      with:
-        node-version: 10.15
-    - uses: microsoft/playwright-github-action@v1
-    - run: npm ci
-    - run: npm run build
-    - run: npm run lint
+name: "infra"
```
mxschmitt, line 1: Did you change the line endings? Since it does show a diff for the whole file.

.github/workflows/tests.yml
```diff
@@ -29,6 +29,14 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: 10
+    - uses: actions/cache@v2
+      with:
+        path: |
+          ~/.npm
+          ~/.cache/ms-playwright
```
mxschmitt, line 36: If browsers should be cached, the browsers.json needs to be added as a cache key, but not sure if they should be cached.

kumaraditya303: Even if the browsers are cached if the browsers mismatch then `node install-from-github.js` will clear it out .

yury-s: +1 to what Max said, if you want to cache browsers the cache key should consist of the minimal required set of files which is browsers.json in our case regardless of whether install step invalidates the cache or not. It should be obvious to a person reading the config why it is ok to cache given directory with a particular key.

yury-s (commented):
The cache doesn't seem to save much while adding potential source of problems: look at e.g. `macos-11.0 (chromium)` on these two runs:
- https://github.com/microsoft/playwright/pull/4747/checks?check_run_id=1567369907
- https://github.com/microsoft/playwright/pull/4753/checks?check_run_id=1570874678

The latter restored everything from cache as you configured and the `npm ci` step took 57s  compared to 1m10s without the cache (also cache step took 2s). I would say 11s saved out of >7m (and this is one of the fastest runs) doesn't justify extra complexity. If you look at `windows(chromiun)` the run with cache (31s + 32s) took even longer than without cache (48s), some of that may be due to difference between the vms but overall it's unclear to me if this cache is a positive change overall.

kumaraditya303:
> The cache doesn't seem to save much while adding potential source of problems: look at e.g. `macos-11.0 (chromium)` on these two runs:
> 
> * https://github.com/microsoft/playwright/pull/4747/checks?check_run_id=1567369907
> * https://github.com/microsoft/playwright/pull/4753/checks?check_run_id=1570874678
> 
> The latter restored everything from cache as you configured and the `npm ci` step took 57s compared to 1m10s without the cache (also cache step took 2s). I would say 11s saved out of >7m (and this is one of the fastest runs) doesn't justify extra complexity. If you look at `windows(chromiun)` the run with cache (31s + 32s) took even longer than without cache (48s), some of that may be due to difference between the vms but overall it's unclear to me if this cache is a positive change overall.

@yury-s `windows(chromium)` is slow because the zstd compression program isn't used on windows hence it takes longer, for that we can just remove the caching for windows. Also there are differences between vms as you said which also causes latency but we should not be counting that as it isn't caused by cache but by the actual vms. This caching will greatly speed up android test as the emulator is cached by this PR also as @mxschmitt pointed the required `browsers.json` is added to the `hashFiles` list.

kumaraditya303:
Only linux, macos and android are now cached skipped windows as it does not support zstd.

PR #4754, opened 2020-12-17
fix(selector): bring back v1 query logic

### LINKED ISSUE #4742: [REGRESSION]: css/xpath selector engines don't work when chained with >>
**Context:**
- GOOD Playwright Version: 1.6.2
- BAD Playwright Version: 1.7.0
- Extra: -

**Code Snippet**

```html
  <div>
    <label>Test</label>
    <input>
  </div>
```
```javascript

(async () => {
  const input = await page.$('label >> xpath=.. >> input');
  await input.fill('Hello');
})();
```

**Describe the bug**

Hello,

Since the 1.7.0, it seems it's not possible to chain css/xpath selector engines. The code above ends with the following error:
```
TypeError: Cannot read property 'fill' of null
```
It worked well in the 1.6.2 version.

It also works well if done with several lines, this way:
```javascript
  const label = await page.$('label');
  const parent = await label.$('xpath=..');
  const input = await parent.$('input');
```

Thanks !

dgozman:
It turned out that v1 query logic is not shimmable by v2 logic.
This change brings back v1 query logic for `>>` combinator.

Fixes #4742.

### DIFF
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -15,13 +15,13 @@
  */
 
 import { createAttributeEngine } from './attributeSelectorEngine';
-import { createCSSEngine } from './cssSelectorEngine';
 import { SelectorEngine, SelectorRoot } from './selectorEngine';
 import { createTextSelector } from './textSelectorEngine';
 import { XPathEngine } from './xpathSelectorEngine';
-import { ParsedSelector, ParsedSelectorV1, parseSelector, selectorsV2Enabled, selectorsV2EngineNames } from '../common/selectorParser';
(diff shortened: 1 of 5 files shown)


src/server/injected/injectedScript.ts
```diff
@@ -129,8 +117,20 @@ export class InjectedScript {
     const candidates = Array.from(set) as Element[];
     if (!partsToCheckOne.length)
       return candidates;
-    const partial = { parts: partsToCheckOne };
-    return candidates.filter(e => !!this._querySelectorRecursivelyV1(e, partial, 0));
+    const partial = { parts: partsToCheckOne, names: [] };
```
aslushnikov, line 120: i don't understand names!!

aslushnikov (approved):

PR #4766, opened 2020-12-18
feat(firefox): roll Firefox to r1223

aslushnikov:


yury-s:
We are already at 1224

PR #4769, opened 2020-12-18
browser(webkit): postpone creation of the first page

dgozman:
When we create the first page in the default context in headless mode on mac,
it gets `NSWindow` that is "not visible". Although we call `[window setIsVisible:YES]`,
later on `window.isVisible` still returns `NO`.

We create our offscreen "headless" `NSWindow` directly from `applicationDidFinishLaunching:`.
Experiments show that delaying this by 100ms makes everything work. As a symptom,
we get `applicationDidUnhide:` notification that does not happen when we create the window
immediately.

Perhaps, we create the window too early, and there is some essential initialization
that happens after `applicationDidFinishLaunching:`. However, if we call
`[NSApp activateIgnoringOtherApps:YES]` like we do in headful mode, everything works.

The only solution that worked so far is creating the first page after a timeout.

### DIFF
--- a/browser_patches/webkit/embedder/Playwright/mac/AppDelegate.m
+++ b/browser_patches/webkit/embedder/Playwright/mac/AppDelegate.m
@@ -250,8 +250,18 @@ const NSActivityOptions ActivityOptions =
     if (_noStartupWindow)
         return;
 
-    [self createNewPage:0];
-    _initialURL = nil;
+    // Force creation of the default browser context.
+    [self defaultConfiguration];
+    // Creating the first NSWindow immediately makes it invisible in headless mode,
+    // so we postpone it for 50ms. Experiments show that 10ms is not enough, and 20ms is enough.
+    // We give it 50ms just in case.
+    [NSTimer scheduledTimerWithTimeInterval: 0.05
+                                    repeats: NO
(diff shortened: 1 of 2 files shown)


browser_patches/webkit/embedder/Playwright/mac/AppDelegate.m
```diff
@@ -250,8 +250,18 @@ - (void)applicationDidFinishLaunching:(NSNotification *)aNotification
...
+    // Force creation of the default browser context.
+    [self defaultConfiguration];
+    // Creating the first NSWindow immediately makes it invisible in headless mode,
+    // so we postpone it for 50ms. Experiments show that 10ms is not enough, and 20ms is enough.
+    // We give it 50ms just in case.
+    [NSTimer scheduledTimerWithTimeInterval: 0.05
+                                    repeats: NO
+                                      block:(void *)^(NSTimer* timer)
+    {
+        [self createNewPage:0 withURL:_initialURL ? _initialURL : @"about:blank"];
```
yury-s, line 262: Would it make sense to simply create the page on the client side (and always start in _noStartupWindow mode) ?

yury-s (approved):

PR #4773, opened 2020-12-19
feat(cli): first few cli commands

dgozman:
This sets up cli capabilities and brings in a few commands from playwright-cli.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,10 +16,4 @@
 
 const { setUnderTest } = require('./lib/utils/utils');
 setUnderTest(); // Note: we must call setUnderTest before initializing.
-
-const { Playwright } = require('./lib/server/playwright');
-const { setupInProcess } = require('./lib/inprocess');
-const path = require('path');
-
-const playwright = new Playwright(__dirname, require(path.join(__dirname, 'browsers.json'))['browsers']);
-module.exports = setupInProcess(playwright);
+module.exports = require('./lib/inprocess');
--- a/package.json
+++ b/package.json
@@ -38,7 +38,11 @@
     "name": "Microsoft Corporation"
   },
   "license": "Apache-2.0",
+  "bin": {
+    "playwright": "./lib/cli/cli.js"
+  },
   "dependencies": {
+    "commander": "^6.1.0",
     "debug": "^4.1.1",
     "extract-zip": "^2.0.1",
     "https-proxy-agent": "^5.0.0",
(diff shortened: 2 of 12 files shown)


index.js
```diff
@@ -16,10 +18,6 @@
...
 setUnderTest(); // Note: we must call setUnderTest before initializing.
-
-const { Playwright } = require('./lib/server/playwright');
-const { setupInProcess } = require('./lib/inprocess');
-const path = require('path');
-
-const playwright = new Playwright(__dirname, require(path.join(__dirname, 'browsers.json'))['browsers']);
-module.exports = setupInProcess(playwright);
+module.exports = require('./lib/inprocess');
+if (require.main === module)
```
JoelEinbinder, line 22: this is a strange python pattern. The standard way ive seen is to set the "bin" field in `package.json` with a separate file. See https://github.com/microsoft/folio/blob/master/package.json#L4

packages/installation-tests/installation-tests.sh
```diff
@@ -40,18 +40,20 @@ function copy_test_scripts {
...
-  test_playwright_global_installation_subsequent_installs
-  test_playwright_should_work
-  test_playwright_chromium_should_work
-  test_playwright_webkit_should_work
-  test_playwright_firefox_should_work
-  test_playwright_global_installation
-  test_playwright_global_installation_cross_package
-  test_playwright_electron_should_work
-  test_electron_types
+  # test_screencast
```
JoelEinbinder, line 43: just for the draft?

packages/installation-tests/sanity.js
```diff
@@ -42,7 +42,7 @@ const installer = require(requireName + '/lib/install/installer');
       await browser.close();
     } catch (e) {
       console.error(`Should be able to launch ${browserType} from ${requireName}`);
-      console.error(err);
+      console.error(e);
```
JoelEinbinder, line 45: :+1:

packages/installation-tests/installation-tests.sh
dgozman: Thanks for catching this!

index.js
dgozman: Done.

dgozman:
@JoelEinbinder Please take another look.

package.json
```diff
@@ -38,7 +38,11 @@
     "name": "Microsoft Corporation"
   },
   "license": "Apache-2.0",
+  "bin": {
+    "playwright": "./lib/cli/cli.js"
```
yury-s, line 42: Do we want to merge playwright-cli npm into playwright? I thought we wanted to keep them separate.

yury-s (approved):

package.json
dgozman: I think that's the point of merging!

PR #4783, opened 2020-12-21
fix(webkit): properly detect arm64 on apple silicon

JoelEinbinder:
`os.arch()` returns `"x64"` on all platforms. `uname -u` returns `i386` when called from node. However `sysctl` tells us the real architecture, even if we are running under Rosetta.

Depending on how we merge this into a release, we should also bump the browser number.

### DIFF
--- a/src/utils/browserPaths.ts
+++ b/src/utils/browserPaths.ts
@@ -35,7 +35,13 @@ export const hostPlatform = ((): BrowserPlatform => {
     const macVersion = execSync('sw_vers -productVersion', {
       stdio: ['ignore', 'pipe', 'ignore']
     }).toString('utf8').trim().split('.').slice(0, 2).join('.');
-    const archSuffix = os.arch() === 'arm64' ? '-arm64' : '';
+    let arm64 = false;
+    if (!macVersion.startsWith('10.')) {
+      arm64 = execSync('sysctl -in hw.optional.arm64', {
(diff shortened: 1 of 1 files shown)


yury-s:
Is this specific to Node 14? In Node 15 `os.arch()` returns 'arm64'.

src/utils/browserPaths.ts
```diff
@@ -35,7 +35,10 @@ export const hostPlatform = ((): BrowserPlatform => {
     const macVersion = execSync('sw_vers -productVersion', {
       stdio: ['ignore', 'pipe', 'ignore']
     }).toString('utf8').trim().split('.').slice(0, 2).join('.');
-    const archSuffix = os.arch() === 'arm64' ? '-arm64' : '';
+    const arm64 = execSync('sysctl -in hw.optional.arm64', {
```
yury-s, line 38: Let's restrain this method to macOS 11+ and Node <15 ?

yury-s (approved):

src/utils/browserPaths.ts
JoelEinbinder: Node 15 reports 'x64' unless you compile it yourself.

JoelEinbinder:
as per discussion with dgozman offline, added a check that doesn't run this code unless you are on macos 11 or higher.

dgozman (approved):

PR #4801, opened 2020-12-22
chore: fix check-deps

dgozman:


yury-s (approved):

src/dispatchers/androidDispatcher.ts
```diff
@@ -176,8 +175,8 @@ export class AndroidDeviceDispatcher extends Dispatcher<AndroidDevice, channels.
 export class AndroidSocketDispatcher extends Dispatcher<SocketBackend, channels.AndroidSocketInitializer> implements channels.AndroidSocketChannel {
   constructor(scope: DispatcherScope, socket: SocketBackend) {
     super(scope, socket, 'AndroidSocket', {}, true);
-    socket.on(Events.AndroidSocket.Data, (data: Buffer) => this._dispatchEvent('data', { data: data.toString('base64') }));
-    socket.on(Events.AndroidSocket.Close, () => {
+    socket.on('data', (data: Buffer) => this._dispatchEvent('data', { data: data.toString('base64') }));
```
mxschmitt, line 178: Out of curiosity, what was the reason that this was changed?

I thought its fine since its similar like here (imported Events too) https://github.com/microsoft/playwright/blob/d20e56e197dd5e64d8d4503db8c83e76cdb78930/src%2Fdispatchers%2FbrowserContextDispatcher.ts#L32-L38

dgozman: `BrowserContextDispatcher` uses private events from `server/browserContext.ts` and that's how it is supposed to be.

`AndroidDispatcher` was importing client-side events that have no relation to what happens on the server side. Instead, we use standard socket events there. We could go ahead and introduce server-side events for these if needed to make code more readable. However, using client events would be a mistake, because the emitting side does not use them.

PR #4807, opened 2020-12-22
feat(cli): share console api between cli and debug mode

dgozman:


src/cli/cli.ts
```diff
@@ -284,6 +285,7 @@ async function openPage(context: BrowserContext, url: string | undefined): Promi
 
 async function open(options: Options, url: string | undefined) {
   const { context } = await launchContext(options, false);
+  (context as any)._extendInjectedScript(consoleApiSource.source);
```
JoelEinbinder, line 288: These types should just work right? Because this is the internal context not the types we generate.

JoelEinbinder (approved):

src/cli/cli.ts
dgozman: Good point!

PR #4825, opened 2020-12-26
chore: generate types, api.json off md rather than html

pavelfeldman:


spkml:
Forgive me if this seems awkward as I'm a little new to this whole collaboration through github protocol for interaction...  My name is Christopher Johnson and I'm working on developing a team for a project on github.com/spektrum-creations/LPA-webApp￼ - Where I will also be starting at least two other projects almost immediately before the new year we're on the new year￼ -
I like the work you're doing on this project by requiring the ￼ The inclusion of this API for ￼the purpose of unification￼ of code is brilliant.  I see this as putting the past behind all of us by declaring an end to the browser wars in every aspect and calling for a unified CodeBase for the web leading to seamless interoperability which I would expect nothing less as ￼we move forward into the future with these kinds of ￼projects.
with that said, ￼I would like to see if I can work towards achieving some of the same ends as a unified set of standards did encourage web standards and related files to be much more standardized and someways more strict￼ than any of the more loose parsing of generalized html pages with more of an XHTML Requirement￼￼ supporting very similar nested set of concepts using a more natural language approach￼ I can get a rather aggressive to create a language called SPKML (Semantic Programming & Knowledge Machine-translated￼ Language) which my hope is will be potentially utilized in the same way PHP nesting extended ￼HTML tag set (Utilizing and incorporating the current HTML Living Standard tagset what's perhaps slightly modified DTD  definitions ￼and perhaps even re-implementing certain deprecated tags like bold and italics only for quick and easy css free coded text styling and also including some tags considered for inclusion presently in the current html specification￼￼) incorporating The more fluid natural language capability as well as￼ concepts of the upcoming standards in the works such as languages like￼ ￼the current work being done to the core EMCA-JavaScript language while also Encouraging plug-and-play-style mini-APIs generally loaded as js files like jquery which is a great example to represent the suggested change.  they will be treated as dependencies on top of the core JavaScript included in every browser as the EMCA engine￼￼(Which seems that if you include ￼Console as well {additionally, this￼ console Could be easily integrated by including projects Such as Lib-term or other cli consoles that could be dropped in ths way to enable￼ processing of￼ python/bash/js, etc ￼￼￼}.  This core codebase of the compiled core Javascript engune could be enabled to allow for compression/minifying ￼through compilation of every HTML-based document with the inclusion of Another layer of compression which could utilize Unicode character sets As the basis for the encoding of such a document Setting￼ each Repetitious portion such as the body tag, the head tag￼￼, anchor tag, etc. to be represented I single Characters Within Unicode character search like you TF8 or UTF 16 or even UT have 32 or 64 will extend the character set so much that you could encapsulate and minify into and from Machine translated code both read by machines without translation and processed by interpretation for human readability.￼ Enabling the creation of￼￼ a machine readable only Version of the compiled code containing all elements directly linked and contained or required as dependencies by the parent document while also ￼perhaps requiring again that there be a Console-based ￼decompiler scripted program suite allowing Translation post to and from not just English put any and all languages￼ ￼ defined as usable in a ￼Human readable code format supprted as an international standars￼￼￼￼). This essentially Must be provided initially at least on installation￼ provides a common core code processing engine that can be used￼ ￼in and  by any browser.  my vision here for my take rather would be to put any additional library's over and above the core javascript required to run functions and define define variables etc. in JavaScript will still allow for a custom code that will make it extraordinarily easy to utilize code library's that are loaded ￼￼￼from CDN￼ repositories... With this change it's mainly￼ Seems to require that W a definition of no less than three levels of code from not just the system, network ￼and user spaces but also Security for each level from a basis where permissions are required no matter what system you're on because it sandbox within the browser was proper permissions included disallowing in most level level of protection that disallows execution at the system level from a browser and a code from an ￼￼￼extended branch Script more of any Script￼ injection to the core￼ JavaScript ￼or Standard EMCA engine￼ as to set and meet goals for usability, interoperability and security￼.

Again my apologies if this seems a little Heavy handed in its declaration￼ or even unrelated￼...  To the best of my knowledge this is how it can work. If there's any￼ Technical flaws that I've overlooked or that I'm unaware of￼ Relating to my understanding of code ability as well as function or operation please do not feel afraid to enlighten me and inform me properly. I personally welcome the ability to gain correct knowledge and have no real issue with incorporating new information if necessary especially when it comes to technical details.￼

I will be posting a link to this repositories' URL on a new page for a repo that ￼I'm starting, that I have created in the SPEKTRUM-Creations￼/SPKML-Language space on github.com and I will also be sharing some of this vision related in this email message on that page.  I look forward to receiving a response. I appreciate any and all help you can offer and if I am able to help at all moving forward with my rather limited coding ability actually which is pretty mediocre ultimately in my opinion I do offer my assistance where I can and where I understand the overall subjects related...   Thanks in advance￼￼￼.

Christopher Johnson
New Media Artisan￼￼
SPEKTRUM Creations 
SPKML.com

utils/generate_types/index.js
```diff
@@ -396,8 +397,6 @@ function memberJSDOC(member, indent) {
   if (member.comment)
     lines.push(...member.comment.split('\n'));
   lines.push(...member.argsArray.map(arg => `@param ${arg.name.replace(/\./g, '')} ${arg.comment.replace('\n', ' ')}`));
-  if (member.returnComment)
```
dgozman, line 399: I'm curious where did this go?

dgozman (approved):

dgozman (approved):

utils/generate_types/index.js
pavelfeldman: It is now always a part of the method body.

PR #4831, opened 2020-12-27
chore: remove JS types checker, rely on typescript

dgozman:
We keep checking that all methods are documented, and no extra methods
are documented, but rely on typescript for everything else.


docs-src/api-body.md
```diff
@@ -3811,9 +3811,6 @@ When all steps combined have not finished during the specified [`option: timeout
 
 Returns the `node.textContent`.
 
-## method: ElementHandle.toString
```
JoelEinbinder, line 3814: 👍

utils/doclint/check_public_api/missingDocs.js
```diff
@@ -0,0 +1,175 @@
...
+  });
+  const checker = program.getTypeChecker();
+  const apiClassNames = new Set();
+  const apiMethods = new Map();
+  const apiSource = program.getSourceFiles().find(f => f.fileName === apiFileName);
+
+  /**
+   * @param {ts.Type} type
+   */
+  function signatureForType(type) {
```
JoelEinbinder, line 106: JSBuilder lives!

JoelEinbinder (approved):

utils/doclint/check_public_api/missingDocs.js
dgozman: I was debating whether to add this check for missing parameters, and decided it only costs us this one JSBuilder function 😄

JoelEinbinder: The typescript compiler api isn't quite stable, so this function has actually broken once or twice. But I think we can afford it.

PR #4835, opened 2020-12-28
test: remove hacky requires, use imports instead

dgozman:
Since our client implements our types, we can now import
implementation in tests without type conflicts.

### DIFF
--- a/test/css-parser.spec.ts
+++ b/test/css-parser.spec.ts
@@ -15,10 +15,7 @@
  */
 
 import { it, expect } from './fixtures';
-import * as path from 'path';
-
-const { parseCSS, serializeSelector: serialize } =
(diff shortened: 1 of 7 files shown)


JoelEinbinder:
I'm not sure how the patches were related. We should've been able to do this once there was a client/server split right?

JoelEinbinder (approved):

dgozman:
> I'm not sure how the patches were related. We should've been able to do this once there was a client/server split right?

Perhaps. I just remember that passing `Page` imported from somewhere to public methods was an error, but everything works now, so that does not matter 😄

JoelEinbinder:
I don't think we do that anymore, again since the server split. All the juicy hidden internal methods are server side now.

PR #4843, opened 2020-12-29
doc: generator code health (2)

pavelfeldman:


utils/doclint/generateApiJson.js
```diff
@@ -62,6 +62,7 @@ function serializeMember(member) {
   const result = { ...member };
   sanitize(result);
   result.args = {};
+  delete member.clazz;
```
dgozman, line 65: Put this to sanitize()?

utils/generate_types/index.js
```diff
@@ -239,6 +260,12 @@ function parentClass(classDesc) {
 
 function writeComment(comment, indent = '') {
   const parts = [];
+  
+  comment = comment.replace(/\[`([^`]+)`\]\(#([^\)]+)\)/g, '[`$1`](https://github.com/microsoft/playwright/blob/master/docs/api.md#$2)');
```
dgozman, line 264: Should we have a shared place for all these magic md regexes scattered around the files?

dgozman (approved):

utils/doclint/generateApiJson.js
pavelfeldman: Sure.

utils/generate_types/index.js
pavelfeldman: This code is volatile, I don't think sharing these makes sense.

utils/doclint/generateApiJson.js
pavelfeldman: (follow up)

PR #4844, opened 2020-12-29
feat: don't scare users with browser downloads

JoelEinbinder:
We get a lot of requests to install playwright and only download one of the browsers. But its unclear whether users actually need to save disk space/bandwidth or just get annoyed seeing the three loading bars. I combine them into one bar and make it a little bit less intense when we download browsers.

This patch also downloads the browsers in parallel, which might speed things up significantly.

### DIFF
--- a/src/install/browserFetcher.ts
+++ b/src/install/browserFetcher.ts
@@ -154,29 +154,46 @@ function revisionURL(browser: BrowserDescriptor, platform = browserPaths.hostPla
   return util.format(urlTemplate, serverHost, browser.revision);
 }
 
-export async function downloadBrowserWithProgressBar(browsersPath: string, browser: BrowserDescriptor): Promise<boolean> {
-  const browserPath = browserPaths.browserDirectory(browsersPath, browser);
-  const progressBarName = `${browser.name} v${browser.revision}`;
-  if (await existsAsync(browserPath)) {
-    // Already downloaded.
-    return false;
-  }
-
+export async function downloadBrowsersWithProgressBar(browsersPath: string, ...browsers: BrowserDescriptor[]): Promise<boolean[]> {
   let progressBar: ProgressBar;
   let lastDownloadedBytes = 0;
+  const totals = browsers.map(() => 0);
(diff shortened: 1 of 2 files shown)


src/install/installer.ts
```diff
@@ -100,12 +100,11 @@ async function validateCache(packagePath: string, browsersPath: string, linksDir
...
   const myBrowsersToDownload = await readBrowsersToDownload(packagePath);
-  for (const browser of myBrowsersToDownload) {
-    await browserFetcher.downloadBrowserWithProgressBar(browsersPath, browser).catch(e => {
-      throw new Error(`Failed to download ${browser.name}, caused by\n${e.stack}`);
-    });
+  await browserFetcher.downloadBrowsersWithProgressBar(browsersPath, ...myBrowsersToDownload).catch(e => {
+    throw new Error(`Failed to download browser, caused by\n${e.stack}`);
+  });
+  for (const browser of myBrowsersToDownload)
     await fsWriteFileAsync(browserPaths.markerFilePath(browsersPath, browser), '');
```
aslushnikov, line 107: this should be moved inside the the download browser part - otherwise they're marked as "donwloaded" only when all of them complete download.

src/install/browserFetcher.ts
```diff
@@ -154,30 +154,47 @@ function revisionURL(browser: BrowserDescriptor, platform = browserPaths.hostPla
...
-
+export async function downloadBrowsersWithProgressBar(browsersPath: string, ...browsers: BrowserDescriptor[]): Promise<boolean[]> {
   let progressBar: ProgressBar;
   let lastDownloadedBytes = 0;
+  const totals = browsers.map(() => 0);
+  const currents = browsers.map(() => 0);
 
-  function progress(downloadedBytes: number, totalBytes: number) {
+  function progress(index: number, current: number, total: number) {
+    if (index === 2)
```
aslushnikov, line 164: debug?

pavelfeldman:
Looks stale now

PR #4864, opened 2021-01-01
doc: split classes into files

pavelfeldman:


docs/cli.md
```diff
@@ -1,26 +1,17 @@
-<!-- THIS FILE IS NOW GENERATED -->
+---
+id: ci
```
dgozman, line 2: Wrong id and title here.

docs-src/actionability.md
```diff
@@ -1,4 +1,7 @@
-# Actionability
+---
+id: actionability
```
dgozman, line 2: It seems like id is always the same as the file name. Do we need two? Can they differ?

dgozman (approved):

docs-src/actionability.md
pavelfeldman: Dunno, that's what Arjun generates!

pavelfeldman: Dunno, that's what Arjun geberates!

PR #4877, opened 2021-01-04
feat(webkit): bump to 1410

dgozman:


yury-s (commented):
there is no windows build for 1410 due to the ANGLE issue.

PR #4896, opened 2021-01-05
docs: mark some paths as such

pavelfeldman:


docs/src/api-params.md
```diff
@@ -67,7 +67,7 @@ modifiers back. If not specified, currently pressed modifiers are used.
 Defaults to `left`.
 
 ## input-files
-- `files` <[string]|[Array]<[string]>|[Object]|[Array]<[Object]>>
```
yury-s, line 70: Is there such type as `path`?

pavelfeldman: Yes

PR #4907, opened 2021-01-06
chore: bundle small build of highlight.js

dgozman:
This follows the "build from source" configuration (https://highlightjs.org/usage/#source)
and reduces the footprint from 2mb to 100kb.


aslushnikov (approved):
alternative: headless mode for codemirror?

Issue #4910, opened 2021-01-06
[BUG] setOffline does not emulate offline for web sockets


kahoot-karl:
**Context:**
- Playwright Version: 1.7.1
- Operating System: Mac
- Node.js version: 14.9.0
- Browser: Firefox, Chrome

System:
    OS: macOS 11.1
    Memory: 2.13 GB / 16.00 GB
  Binaries:
    Node: 14.9.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.8 - /usr/local/bin/npm
  Languages:
    Bash: 3.2.57 - /bin/bash

**Describe the bug**

Sorry, I don't have a small repro of this problem, but when using `context.setOffline(true)`, I can confirm that network requests fail, but all open web socket connections continue to be able to send and receive data.

JoelEinbinder:
Thanks for the report. I confirmed it with a test.

pavelfeldman:
#### Why was this issue closed?

We are prioritizing the features based on the upvotes, recency and activity in the issue thread. It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it. We are closing issues like this one to keep our bug database maintainable. Please feel free to open a new issue and link this one to it if you think this is a mistake.

PR #4912, opened 2021-01-06
test: failing test for websockets + offline context

JoelEinbinder:
It turns out our offline mode doesn't affect websockets.

#4910

### DIFF
--- a/test/http.fixtures.ts
+++ b/test/http.fixtures.ts
@@ -17,6 +17,7 @@
 import { folio as base } from 'folio';
 import path from 'path';
 import socks from 'socksv5';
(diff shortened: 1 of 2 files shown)


test/http.fixtures.ts
```diff
@@ -65,6 +67,14 @@ fixtures.httpsServer.init(async ({ httpService }, test) => {
   await test(httpService.httpsServer);
 });
 
+fixtures.webSocketServer.init(async ({ testWorkerIndex }, run) => {
```
dgozman, line 70: I wonder why all other ws tests do not use this. Can we unify our ws harness?

PR #4913, opened 2021-01-06
fix(electron): return a ChromiumBrowserContext for electron

### LINKED ISSUE #4905: [BUG] Electron: neither ElectronApplication.context() nor ElectronPage.context() are ChromiumBrowserContexts
**Context:**

Playwright-electron version: 0.5.0

  System:
    OS: macOS 10.15.6
    Memory: 60.15 GB / 128.00 GB
  Binaries:
    Node: 14.15.0 - ~/.nodenv/versions/14.15.0/bin/node
    Yarn: 1.17.3 - ~/.yarn/bin/yarn
    npm: 6.14.8 - ~/.nodenv/versions/14.15.0/bin/npm
  Languages:
    Bash: 5.0.7 - /usr/local/bin/bash

**Code Snippet**

```js
// index.js
const { app, BrowserWindow } = require('electron');
 
function createWindow () {
  let win = new BrowserWindow({
    width: 800,
    height: 600,
  });
 
  win.loadFile('index.html');
}
 
app.whenReady().then(createWindow);
```

```
<!-- index.html -->
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>Hello World!</title>
    <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-inline';" />
  </head>
  <body>
    <h1>Hello World!</h1>
    <button onclick="console.log('click')">Click me</button>
  </body>
</html>

```

```javascript
// pw.js
const { electron } = require('playwright-electron');
const assert = require('assert');
const electronPath = require('electron');
const path = require('path')

async function test() {
    const app = await electron.launch(electronPath, {
	path: electronPath,
	args: [__dirname],
    });

    const appContext = app.context();
    console.log(appContext.constructor.name); // BrowserContext; should be ChromiumBrowserContext per TypeScript types
    console.log(appContext.newCDPSession); // undefined; should be a function

    const page = await app.firstWindow();
    const pageContext = await page.context();
    console.log(pageContext.constructor.name); // BrowserContext; matches the TypeScript type, but should it be a ChromiumBrowserContext?
    console.log(pageContext.newCDPSession); // undefined

    await app.close();
}

test();
```

Then run `node pw.js`

**Describe the bug**

I am trying to use Playwright/Playwright-electron to collect performance testing statistics of an Electron app (hence the desire to use the Chrome DevTools Protocol).

The TypeScript type of `ElectronApplication.context()` is `ChromiumBrowserContext`, but it is merely a `BrowserContext`. 

Similarly, while the type of `ElectronPage.context()` is `BrowserContext`, I wonder if it could instead be `ChromiumBrowserContext` to allow, among other things, creation of a CDP session.

JoelEinbinder:
fixes #4905

### DIFF
--- a/src/dispatchers/browserDispatcher.ts
+++ b/src/dispatchers/browserDispatcher.ts
@@ -45,21 +45,21 @@ export class BrowserDispatcher extends Dispatcher<Browser, channels.BrowserIniti
   }
 
   async crNewBrowserCDPSession(): Promise<channels.BrowserCrNewBrowserCDPSessionResult> {
-    if (this._object._options.name !== 'chromium' && this._object._options.name !== 'electron')
+    if (!this._object._options.isChromium)
       throw new Error(`CDP session is only available in Chromium`);
     const crBrowser = this._object as CRBrowser;
     return { session: new CDPSessionDispatcher(this._scope, await crBrowser.newBrowserCDPSession()) };
   }
 
   async crStartTracing(params: channels.BrowserCrStartTracingParams): Promise<void> {
-    if (this._object._options.name !== 'chromium' && this._object._options.name !== 'electron')
+    if (!this._object._options.isChromium)
       throw new Error(`Tracing is only available in Chromium`);
     const crBrowser = this._object as CRBrowser;
     await crBrowser.startTracing(params.page ? (params.page as PageDispatcher)._object : undefined, params);
   }
 
   async crStopTracing(): Promise<channels.BrowserCrStopTracingResult> {
-    if (this._object._options.name !== 'chromium' && this._object._options.name !== 'electron')
+    if (!this._object._options.isChromium)
       throw new Error(`Tracing is only available in Chromium`);
     const crBrowser = this._object as CRBrowser;
     const buffer = await crBrowser.stopTracing();
--- a/src/dispatchers/browserDispatcher.ts
+++ b/src/dispatchers/browserDispatcher.ts
@@ -45,21 +45,21 @@ export class BrowserDispatcher extends Dispatcher<Browser, channels.BrowserIniti
   }
 
   async crNewBrowserCDPSession(): Promise<channels.BrowserCrNewBrowserCDPSessionResult> {
-    if (this._object._options.name !== 'chromium')
+    if (this._object._options.name !== 'chromium' && this._object._options.name !== 'electron')
       throw new Error(`CDP session is only available in Chromium`);
     const crBrowser = this._object as CRBrowser;
     return { session: new CDPSessionDispatcher(this._scope, await crBrowser.newBrowserCDPSession()) };
   }
 
   async crStartTracing(params: channels.BrowserCrStartTracingParams): Promise<void> {
-    if (this._object._options.name !== 'chromium')
+    if (this._object._options.name !== 'chromium' && this._object._options.name !== 'electron')
       throw new Error(`Tracing is only available in Chromium`);
     const crBrowser = this._object as CRBrowser;
     await crBrowser.startTracing(params.page ? (params.page as PageDispatcher)._object : undefined, params);
   }
 
   async crStopTracing(): Promise<channels.BrowserCrStopTracingResult> {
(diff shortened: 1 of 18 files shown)


src/dispatchers/browserDispatcher.ts
```diff
@@ -45,21 +45,21 @@ export class BrowserDispatcher extends Dispatcher<Browser, channels.BrowserIniti
   }
 
   async crNewBrowserCDPSession(): Promise<channels.BrowserCrNewBrowserCDPSessionResult> {
-    if (this._object._options.name !== 'chromium')
+    if (this._object._options.name !== 'chromium' && this._object._options.name !== 'electron')
```
mxschmitt, line 48: maybe add a helper function like isChromiumBasedBrowser() since its duplicated logic which is "hard" to maintain.

test/electron/electron-window.spec.ts
```diff
@@ -48,4 +48,5 @@ describe('electron window', (suite, { browserName }) => {
     await window.keyboard.type(text);
     expect(await window.evaluate(() => document.querySelector('textarea').value)).toBe(text);
   });
+
```
mxschmitt, line 51:
```suggestion
```

test/electron/electron-app.spec.ts
```diff
@@ -124,4 +124,12 @@ describe('electron app', (suite, { browserName }) => {
     const clipboardContentRead = await application.evaluate(async ({clipboard}) => clipboard.readText());
     await expect(clipboardContentRead).toEqual(clipboardContentToWrite);
   });
+
+  it.only('should be able to send CDP messages', async ({application, window}) => {
```
mxschmitt, line 128:
```suggestion
  it('should be able to send CDP messages', async ({application, window}) => {
```

src/dispatchers/browserDispatcher.ts
dgozman: Yeah, something like isChromiumBased(). I wonder what happens with android? Can we do newCDPSession there? We probably should.

JoelEinbinder: Browsers now register themselves with `isChromium`. I changed the protocol so that `isChromium` is sent rather than `browserName`. And android now reports a ChromiumBrowserContext.

src/protocol/protocol.yml
```diff
@@ -507,7 +507,7 @@ BrowserContext:
   type: interface
 
   initializer:
-    browserName: string
+    isChromium: boolean
```
dgozman, line 510: Why don't we send both? I find browserName pretty useful.

dgozman (approved):

src/protocol/protocol.yml
JoelEinbinder: `browserName` in the client was only used to distinguish chromium and non-chromium based browsers. Removing it ensures that nobody does `browserName === 'chromum'` when they mean `isChromium`.

If will need it back, it will probably be for some feature that requires larger protocol changes. So let's not have it until we need it.

dgozman: Are you sure about what other language clients do? I am somewhat hesitant breaking them without a good reason.

JoelEinbinder: I thought none of them distinguish browsers. I'll check to confirm. I imagine if they are using browserName, they will want to want to fix the same bug and consider electron to be chromium.

JoelEinbinder: python reads the browser name, but should switch to `isChromium` after this patch. https://github.com/microsoft/playwright-python/blob/30c5b52cc460f3cce8ba1910682c9d74e0f6f5bf/playwright/_impl/_object_factory.py#L57

JoelEinbinder: Same in C#: https://github.com/microsoft/playwright-sharp/blob/afb06a5a3c69f48f524260ad382aab59efd13e77/src/PlaywrightSharp/Page.cs#L793

Java, Go, and Crystal do not appear to distinguish between browsers by name.

PR #4920, opened 2021-01-06
feat(cli): bring in trace viewer

dgozman:


src/cli/injected/recorder.webpack.config.js
```diff
@@ -20,6 +20,7 @@ const InlineSource = require('../../server/injected/webpack-inline-source-plugin
 module.exports = {
   entry: path.join(__dirname, 'recorder.ts'),
   devtool: 'source-map',
+  mode: 'development',
```
JoelEinbinder, line 23: Dev mode react is pretty different from prod mode. I'll look into us getting a production webpack build step

src/cli/cli.ts
```diff
@@ -136,6 +137,22 @@ program
       });
     });
 
+if (process.env.PWTRACE) {
```
JoelEinbinder, line 140: This is our way of marking this as experimental?

src/cli/cli.ts
```diff
@@ -400,3 +417,9 @@ function validateOptions(options: Options) {
     process.exit(0);
   }
 }
+
+function resolveHome(filepath: string | undefined) {
+  if (filepath && filepath[0] === '~')
```
JoelEinbinder, line 422: I'm very confused by this. Shouldn't bash do this for us?

JoelEinbinder (approved):

src/cli/injected/recorder.webpack.config.js
mxschmitt: maybe we can make it based on the `NODE_ENV` environment variable like
```js
process.env.NODE_ENV === 'production' ? 'production' : 'development'
```

I think thats common and I saw this already a few times.

dgozman: So, we used to have everything in 'development' mode. Now, I want to minify the trace viewer, so we use 'production' mode in build.js. However, turning everything into 'production' does not work for our injected scripts, so I manually set 'development' in most webpack configs. Does that make sense?

src/cli/cli.ts
dgozman: Yep.

src/cli/cli.ts
dgozman: Don't know, I just copied it. Let me check.

dgozman: Removed.

src/cli/injected/recorder.webpack.config.js
mxschmitt: Curious why it does not work for the injected scripts?

dgozman: Because we expect something like `.default` being exported from the result of webpack bundle. I am sure this can be solved, but nobody looked into it, because benefits are not clear.

PR #4930, opened 2021-01-07
docs: introduce deprecation annotation and any type

pavelfeldman:


types/types.d.ts
```diff
@@ -6461,10 +6458,8 @@ export interface BrowserType<Browser> {
     userAgent?: string;
 
     /**
-     * **NOTE** Use `recordVideo` instead, it takes precedence over `videoSize`. Specifies dimensions of the automatically
-     * recorded video. Can only be used if `videosPath` is set. If not specified the size will be equal to `viewport`. If
-     * `viewport` is not configured explicitly the video size defaults to 1280x720. Actual picture of the page will be scaled
-     * down if necessary to fit specified size.
+     * **DEPRECATED** Use `recordVideo` instead.
```
yury-s, line 6461: nit: drop **DEPRECATED** as we already have @deprecated annotation?

yury-s (approved):

types/types.d.ts
pavelfeldman: VS Code shows a nicer snippet for Markdown than it does for deprecated attribute...

PR #4951, opened 2021-01-08
fix: use int instead of float for location/size parameters

yury-s:


pavelfeldman (changes requested):
This was deliberate.

yury-s:
> This was deliberate.

Why so? We don't expect people to specify fraction of millisecond as the timeout,do we?

yury-s:
> This was deliberate.

Reverted timeout part for now.

PR #4962, opened 2021-01-11
fix(trace viewer): updating default traceStorageDir value

domderen:
When `npx playwright show-trace <tracePath>` command is executed, without providing the `resources` optional parameter, the function expects the `traceStorageDir` default value to be the same directory in which the `tracePath` resides. This change updates it to the `dirname(tracePath)/trace-resources` if it exists. Such directory hierarchy is created by Tracer by default.

### DIFF
--- a/src/cli/traceViewer/traceViewer.ts
+++ b/src/cli/traceViewer/traceViewer.ts
@@ -142,6 +142,10 @@ export async function showTraceViewer(traceStorageDir: string | undefined, trace
     files = [tracePath];
     if (!traceStorageDir)
       traceStorageDir = path.dirname(tracePath);
+
+      if(fs.existsSync(traceStorageDir + '/trace-resources')) {
+        traceStorageDir = traceStorageDir + '/trace-resources';
+      }
   } else {
     files = collectFiles(tracePath);
     if (!traceStorageDir)
--- a/src/cli/traceViewer/traceViewer.ts
+++ b/src/cli/traceViewer/traceViewer.ts
@@ -140,12 +140,12 @@ export async function showTraceViewer(traceStorageDir: string | undefined, trace
   let files: string[];
   if (fs.statSync(tracePath).isFile()) {
     files = [tracePath];
-    if (!traceStorageDir)
+    if (!traceStorageDir) {
       traceStorageDir = path.dirname(tracePath);
 
-      if(fs.existsSync(traceStorageDir + '/trace-resources')) {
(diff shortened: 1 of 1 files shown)


dgozman:
@domderen Thank you for the PR! We usually put a "namespace" or "component" in the change description. I will go ahead and rename this PR according to the [guidelines](https://github.com/microsoft/playwright/blob/master/CONTRIBUTING.md#commit-messages).

dgozman (approved):

src/cli/traceViewer/traceViewer.ts
```diff
@@ -140,8 +140,12 @@ export async function showTraceViewer(traceStorageDir: string | undefined, trace
   let files: string[];
   if (fs.statSync(tracePath).isFile()) {
     files = [tracePath];
-    if (!traceStorageDir)
+    if (!traceStorageDir) {
       traceStorageDir = path.dirname(tracePath);
+
+      if (fs.existsSync(traceStorageDir + '/trace-resources'))
```
dgozman, line 146: Let's move this right before the line 155, to cover the "tracePath is a directory" case as well?

domderen: Wouldn't that potentially break for the case when user runs `npx playwright show-trace --resource /project/trace-resources /project/trace.ndjson` ? I guess that would only break if `/project/trace-resources/trace-resources` would also exist...

dgozman: Good point. Let's copy this heuristic to line 152 then?

src/cli/traceViewer/traceViewer.ts
```diff
@@ -137,19 +137,13 @@ export async function showTraceViewer(traceStorageDir: string | undefined, trace
...
-    }
-  } else {
-    files = collectFiles(tracePath);
-    if (!traceStorageDir)
-      traceStorageDir = tracePath;
+  if (!traceStorageDir) {
+    traceStorageDir = fs.statSync(tracePath).isFile() ? path.dirname(tracePath) : tracePath;
+
+    if (fs.existsSync(traceStorageDir + '/trace-resources'))
+      traceStorageDir = traceStorageDir + '/trace-resources';
```
domderen, line 146: @dgozman What do you think about this? I'm not sure what's your stance on ternary operators.

dgozman: Looks good, thanks!

dgozman (approved):

PR #4968, opened 2021-01-11
chore: bump electron version to 11

aslushnikov:
Electron 11+ supports apple silicon.


JoelEinbinder:
This just changes what our tests run against right? Should we be careful to make sure our electron support continues to work with v9? @pavelfeldman

pavelfeldman (approved):

PR #4974, opened 2021-01-11
fix(driver): use ncc to reduce the number of files

dgozman:
Instead of zipping the whole node_modules directory, use `ncc`
(uses `webpack` internally) to produce a single bundle that is
put to `lib/cli/cli.js`.

For `__dirname`-based file resolution to work:
- introduce `packageRoot()` utility function;
- disallow `__dirname` usage across the `src`.

### DIFF
--- a/utils/build/build-playwright-driver.sh
+++ b/utils/build/build-playwright-driver.sh
@@ -47,7 +47,27 @@ function build {
   cp ./output/${NODE_DIR}/LICENSE ./output/playwright-${SUFFIX}/
   cd ./output/playwright-${SUFFIX}/package
   PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 node "../../${NODE_DIR}/${NPM_PATH}" install --production
+
+  # ----------- Minifying magic ----------
+  # We mess up with lib files location, and therefore use _PW_PACKAGE_ROOT
+  # environment variable to guide our code (see packageRoot() in src).
+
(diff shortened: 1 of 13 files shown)


mxschmitt (approved):
great use case for ncc. I would maybe even put ncc into the devDependencies.

yury-s (approved):

utils/build/build-playwright-driver.sh
```diff
@@ -47,8 +47,28 @@ function build {
   cp ./output/${NODE_DIR}/LICENSE ./output/playwright-${SUFFIX}/
   cd ./output/playwright-${SUFFIX}/package
   PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 node "../../${NODE_DIR}/${NPM_PATH}" install --production
+
+  # ----------- Minifying magic ----------
+  # We mess up with lib files location, and therefore use _PW_PACKAGE_ROOT
+  # environment variable to guide our code (see packageRoot() in src).
+
+  # The goal here is to have "lib2/cli/cli.js" and "lib2/cli/traceViewer/web/*".
+  npm install -D @vercel/ncc@0.26.2
```
aslushnikov, line 56: This would change root `package.json` unless there's a `package.json` in this folder.

How about:
```
npm init -y && npm install -D @vercel/ncc@0.26.2
```

aslushnikov (approved):

PR #4978, opened 2021-01-11
fix(connect): provide an error message when ws endpoint is incorrect

### LINKED ISSUE #4966: [BUG] No exception is thrown if invalid token is provided when connecting to a remote instance of BrowserServer
**Context:**
- Playwright Version: 1.7
- Operating System: Windows
- Node.js version: 10.12
- Browser: All
 
**Code Snippet**

Script to fire up server:

```javascript

(async () => {
    const browserServer = await chromium.launchServer({
        port: 9090
    })
})();
```

Client script that connects to the remote instance of BrowserServer:

```javascript

(async () => {
    // IP address is valid, however, token c0a52db5d206c93707156e5b85b4c02 is not.
    var wsEndpoint = "ws://127.0.0.1:9090/c0a52db5d206c93707156e5b85b4c02"
    try {
        const browser = await chromium.connect({
            wsEndpoint: wsEndpoint });
        //....
    } catch (error) {
        // Invalid token exception is never thrown
        console.log(error.message)
    }
})();
```

**Describe the bug**

- Issue: When trying to connect to a remote instance of BrowserServer with an invalid Token Id, my node.js application exits with no error code/message.
- Expected Behavior: If an invalid Token Id is provided to the wsEndpoint parameter of chromium.connect(), it should throw a catchable exception, so we can properly handle application workflow.
- Use Case: Since I have a virtual machine on Azure running the remote browser server instance, my IP and Token will vary whenever a new internet connection is renewed. Once my node.js app is up, it sends the current IP and Token to the database so all clients can connect to with valid token access.

BTW: I love the extra layer of security by adding tokenization to the connection. Last time I checked Puppeteer, it was not able to perform it. Keep doing the great job, Playwright team!

dgozman:
Fixes #4966.

### DIFF
--- a/src/browserServerImpl.ts
+++ b/src/browserServerImpl.ts
@@ -76,17 +76,13 @@ export class BrowserServerImpl extends EventEmitter implements BrowserServer {
     this._ready = new Promise<void>(f => readyCallback = f);
 
     const token = createGuid();
-    this._server = new ws.Server({ port }, () => {
+    this._server = new ws.Server({ port, path: '/' + token }, () => {
       const address = this._server.address();
       this._wsEndpoint = typeof address === 'string' ? `${address}/${token}` : `ws://127.0.0.1:${address.port}/${token}`;
       readyCallback();
     });
 
     this._server.on('connection', (socket: ws, req) => {
-      if (req.url !== '/' + token) {
-        socket.close();
-        return;
-      }
       this._clientAttached(socket);
     });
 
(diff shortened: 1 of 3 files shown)


aslushnikov (approved):

fabiomoggi:
How can I download this fix from npm?

dgozman:
> How can I download this fix from npm?

Use `npm install playwright@next` to install the newest version (updated daily). This is not a stable version though, use at your own risk.

PR #4986, opened 2021-01-12
docs: python api review

pavelfeldman:


docs/src/api/class-browsercontext.md
```diff
@@ -219,6 +225,7 @@ Whether to pass the argument as a handle, instead of passing by value. When pass
 supported. When passing by value, multiple arguments are supported.
 
 ## async method: BrowserContext.exposeFunction
+**DEPRECATED** Use [`method: BrowserContext.exposeBinding`] instead.
```
mxschmitt, line 228: lets remove it maybe from Python directly?

types/types.d.ts
```diff
@@ -415,9 +419,12 @@ export interface Page {
    * Emitted when attachment download started. User can access basic file operations on downloaded content via the passed
    * [Download] instance.
    * 
-   * > **NOTE** Browser context **must** be created with the `acceptDownloads` set to `true` when user needs access to the
-   * downloaded content. If `acceptDownloads` is not set or set to `false`, download events are emitted, but the actual
-   * download is not performed and user has no access to the downloaded files.
+   * :::note
```
mxschmitt, line 422: this `:::note` isn't official markdown syntax, so shall we maybe transform it for the types and api.json so editors like VSC can display it correctly?

docs/src/api/class-browsercontext.md
pavelfeldman: We actually considered it and decided to leave it as is. I'll remove the deprecation note.

types/types.d.ts
pavelfeldman: Yeah, we should do it at a codegen stage, was planning on doing it later, but might as well do it right away.

PR #4991, opened 2021-01-13
devops(flakiness): persist all test reports, aggregated per commit sha

aslushnikov:
We can either load it raw as-is and process later on the front-end side,
or post-process while experimenting with different storage solutions.


aslushnikov:
TBR @yurys

yury-s:
lgtm

PR #5002, opened 2021-01-13
fix(cli): do not extend injected script on same-document navigations

### LINKED ISSUE #4993: Two recorders are created in one document under CLI
```sh
node lib/cli/cli.js codegen github.com/microsoft
```
type 'vscode'
click vscode

dgozman:
Otherwise, the injected script has to be ready for reentrancy.

Fixes #4993.

### DIFF
--- a/src/cli/injected/recorder.ts
+++ b/src/cli/injected/recorder.ts
@@ -372,7 +372,7 @@ export class Recorder {
     if (event.key === '@' && event.code === 'KeyL')
       return false;
     // Allow and ignore common used shortcut for pasting.
-    if (process.platform === 'darwin') {
+    if (navigator.platform.includes('Mac')) {
       if (event.key === 'v' && event.metaKey)
         return false;
     } else {
--- a/src/server/frames.ts
+++ b/src/server/frames.ts
@@ -199,7 +199,7 @@ export class FrameManager {
     frame.emit(Frame.Events.Navigation, navigationEvent);
     if (!initial) {
       debugLogger.log('api', `  navigated to "${url}"`);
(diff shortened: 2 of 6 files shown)


src/server/frames.ts
```diff
@@ -213,7 +213,6 @@ export class FrameManager {
     const navigationEvent: NavigationEvent = { url, name: frame._name };
     frame.emit(Frame.Events.Navigation, navigationEvent);
     debugLogger.log('api', `  navigated to "${url}"`);
-    this._page.frameNavigated(frame);
```
pavelfeldman, line 216: Are we no longer reporting frame navigated on the Page object for history API?

src/cli/injected/recorder.ts
```diff
@@ -372,7 +374,7 @@ export class Recorder {
     if (event.key === '@' && event.code === 'KeyL')
       return false;
     // Allow and ignore common used shortcut for pasting.
-    if (process.platform === 'darwin') {
+    if (this._platform === 'darwin') {
```
JoelEinbinder, line 377: this is a drive-by? I assume the process.platform was being compiled by webpack into the platform of whoever compiled the bundle. Why not use navigator.platform?

pavelfeldman: Yeah, i would also use navigator for this to feel more webby...

pavelfeldman:
Can we get a test for it as well?

src/cli/injected/recorder.ts
dgozman: Done.

pavelfeldman (approved):

PR #5009, opened 2021-01-14
feat(trace viewer): Extending existing NetworkTab view.

domderen:
Currently the network tab contains a limited amount of information on the resources that were loaded in the browser. This change proposes extending the details displayed for each resource, to include:

- HTTP method,
- Status
- Full url,
- Easily visible response content type,
- Request headers,
- Request & response bodies.

Such level of information could help quickly understand what happened in the application, when it was communicating with backend services. This can help debug tests quicker to figure out why they are failing.

What do you think about such extension?

Zrzut ekranu 2021-01-14 o 15 31 33
Zrzut ekranu 2021-01-14 o 19 58 58
Zrzut ekranu 2021-01-14 o 02 53 39
Zrzut ekranu 2021-01-14 o 15 31 59


src/cli/traceViewer/web/ui/networkResourceDetails.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import './networkResourceDetails.css';
+import * as React from 'react';
+import { Expandable } from './helpers';
+import { NetworkResourceTraceEvent } from '../../../../trace/traceTypes';
+
+
+export const NetworkResourceDetails: React.FunctionComponent<{
+  resource: NetworkResourceTraceEvent,
+  selected: string,
```
dgozman, line 25: Let's make this boolean, and move the selected identity management to the parent.

src/cli/traceViewer/web/ui/networkTab.tsx
```diff
@@ -17,23 +17,18 @@
...
 import './networkTab.css';
 import * as React from 'react';
-import { Expandable } from './helpers';
+import { NetworkResourceDetails } from './networkResourceDetails';
 
 export const NetworkTab: React.FunctionComponent<{
   actionEntry: ActionEntry | undefined,
 }> = ({ actionEntry }) => {
-  const [selected, setSelected] = React.useState(0);
+  const [selected, setSelected] = React.useState('');
```
dgozman, line 25: Why don't we make this a number and use index? There could be multiple resources with the same url.

src/cli/traceViewer/web/ui/networkResourceDetails.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <h4>URL</h4>
+        <div className='network-request-details-url'>{resource.url}</div>
+        <h4>Request Headers</h4>
+        <div className='network-request-headers'>{resource.requestHeaders.map(pair => `${pair.name}: ${pair.value}`).join('\n')}</div>
+        <h4>Response Headers</h4>
+        <div className='network-request-headers'>{resource.responseHeaders.map(pair => `${pair.name}: ${pair.value}`).join('\n')}</div>
+        {resource.requestSha1 !== 'none' ? <h3>Request Body</h3> : ''}
+        {resource.requestSha1 !== 'none' ? <div className='network-request-body'>{formatBody(requestBody, requestContentType)}</div> : ''}
+        <h4>Response Body</h4>
+        {resource.sha1 !== 'none' && responseBody !== null && resource.contentType.includes('image') ? <img src={`data:${resource.contentType};base64,${responseBody}`} /> : ''}
```
dgozman, line 101: nit: is `resource.sha1` check necessary here?

src/cli/traceViewer/web/ui/networkResourceDetails.css
```diff
@@ -0,0 +1,112 @@
...
+  padding: 0px 5px;
+  margin-right: 5px;
+}
+
+.status-success {
+  background-color: rgb(124, 187, 62);
+}
+
+.status-failure {
+  background-color: rgb(211, 35, 35);
```
dgozman, line 61: Could you please use some colors from common.css? If you need another one, let's introduce it.

dgozman (commented):
What a wonderful PR! Sorry for being slow, somehow I missed it when going over PRs. I've added a few comments, but overall looks great.

src/trace/snapshotter.ts
```diff
@@ -34,6 +34,10 @@ export type SnapshotterResource = {
   url: string,
   contentType: string,
   responseHeaders: { name: string, value: string }[],
+  requestHeaders: { name: string, value: string }[],
+  method: string,
+  status: number,
+  requestSha1: string,
   sha1: string,
```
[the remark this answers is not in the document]
domderen: This one is renamed now to `responseSha1`.

src/cli/traceViewer/web/ui/networkResourceDetails.css
```diff
@@ -0,0 +1,112 @@
...
+.network-request-title-status {
+  font-weight: bold;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  padding: 0px 5px;
+  margin-right: 5px;
+}
+
+.status-success {
```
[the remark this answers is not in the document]
domderen: Ok, made those css selectors more specific, so they will now only work in the network tab title.

src/cli/traceViewer/web/ui/networkResourceDetails.tsx
domderen: I guess it is not, since I'm already checking if the responseBody is not null. Got rid of it.

src/cli/traceViewer/web/ui/networkResourceDetails.css
domderen: I'm assuming you meant the variables in `styles.tsx` file? If so, I added two variables there, and used them in the changes I added to the network tab.

src/cli/traceViewer/web/ui/networkResourceDetails.tsx
domderen: Done

src/cli/traceViewer/web/ui/networkTab.tsx
domderen: I switched it back to being a boolean, and using the mapping index. At some point I wanted to make sure that `expanded` & `selected` values will be reset when switching between actions, as I think it looks a bit weird when you open some network resources, then switch actions, and network resources with the same indices are expanded and one of them is selected.

Anyways, I added an effect to `NetworkResourceDetails` component, to reset those values. It should look better now.

domderen:
Hey @dgozman thanks for taking a look at this PR, I think all the changes you pointed out should be fixed now :)

domderen:
One more thing, I was looking at the failing checks, and most of them seems unrelated, but one of them peeked my interest, as it is related to the test I added here: https://github.com/microsoft/playwright/pull/5009/checks?check_run_id=1757111083

It looks like in this particular case, the trace file does not contain the `NetworkResourceTraceEvent` for `/file.json` request. I saw that issue before when I was adding those changes, but they went away. I thought it was related to the way the test was set up, but now I'm starting to wonder if there is potentially a race condition somewhere, that doesn't always wait for network events to resolve, before finishing an action like "click". Do you think this makes sense? Should I create a bug for it, and try to look for it?

test/trace.spec.ts
```diff
@@ -45,6 +45,59 @@ it('should record trace', test => test.fixme(), async ({browser, testInfo, serve
...
 });
+
+it('should record trace with POST', async ({browser, testInfo, server}) => {
+  const traceDir = testInfo.outputPath('trace');
+  const context = await browser.newContext({ _traceDir: traceDir } as any);
+  const page = await context.newPage();
+  const url = server.PREFIX + '/trace-resources.html';
+  await page.goto(url);
+  await page.click('text=Download');
+  await context.close();
```
dgozman, line 69: I think this will solve the race condition:
```js
await page.waitForSelector(`#response-status:text("404")`);
```

src/cli/traceViewer/web/ui/networkResourceDetails.tsx
```diff
@@ -0,0 +1,112 @@
...
+      </div>
+    } body={
+      <div className='network-request-details'>
+        <h4>URL</h4>
+        <div className='network-request-details-url'>{resource.url}</div>
+        <h4>Request Headers</h4>
+        <div className='network-request-headers'>{resource.requestHeaders.map(pair => `${pair.name}: ${pair.value}`).join('\n')}</div>
+        <h4>Response Headers</h4>
+        <div className='network-request-headers'>{resource.responseHeaders.map(pair => `${pair.name}: ${pair.value}`).join('\n')}</div>
+        {resource.requestSha1 !== 'none' ? <h3>Request Body</h3> : ''}
```
dgozman, line 104: Everything is h4, but this one is h3 - looks strange.

src/cli/traceViewer/web/ui/networkResourceDetails.tsx
```diff
@@ -0,0 +1,112 @@
...
+        {resource.requestSha1 !== 'none' ? <div className='network-request-body'>{formatBody(requestBody, requestContentType)}</div> : ''}
+        <h4>Response Body</h4>
+        {responseBody !== null && resource.contentType.includes('image') ? <img src={`data:${resource.contentType};base64,${responseBody}`} /> : ''}
+        {responseBody !== null && !resource.contentType.includes('image') ? <div className='network-request-response-body'>{formatBody(responseBody, resource.contentType)}</div> : ''}
```
dgozman, line 108: I do see a lot of response bodies being empty (for xhrs). Let's do `<empty>` when it is empty, and `<unavailable>` when it is null?

dgozman (commented):
Looks great, just a few minor comments and this is good to go.

test/trace.spec.ts
domderen: Hey, I added this fix, but in general, isn't Playwright supposed to wait for network requests to resolve, before finishing actions such as click? This fix will solve the test, but I assume there is a chance that the Network Resource will be assigned to `waitForSelector` action in some cases, right? Isn't that a potential bug somewhere deeper in Playwright?

src/cli/traceViewer/web/ui/networkResourceDetails.tsx
domderen: Doh, should be fixed now.

src/cli/traceViewer/web/ui/networkResourceDetails.tsx
domderen: I went for something more descriptive for the `<unavailable>` case, what do you think?

test/trace.spec.ts
dgozman: There is a balance to strike here. We do not wait for every possible xhr to finish, otherwise we'll wait forever. Usually, the next action will wait for the logical result of the previous one (for example it will try to click the element that appears as a result of the previous action). However, we do recommend manual waiting in the less common cases.

domderen: Ok makes sense.

dgozman (approved):
Looks great, thank you!

PR #5016, opened 2021-01-14
devops: support frequent minor releases of MacOS BigSur.

### LINKED ISSUE #4775: [Feature] Playwright does not support chromium on mac11.1
(node:1278) UnhandledPromiseRejectionWarning: Error: ERROR: Playwright does not support chromium on mac11.1

aslushnikov:
Pre-BigSur, MacOS updates were labeled as "minor" releases, so we had
to bake separate builds for different 10.X releases.

In BigSur era, it doesn't seem to be the case, so for now we can re-use
our BigSur builds across all BigSur versions (11.0, 11.1 and 11.2).

If we ever need to have a custom build for some bigsur minor version,
e.g. `11.6`, we'll have a new browser platform along with generic
`mac11` platform.

Fixes #4775.


src/install/browserFetcher.ts
```diff
@@ -107,10 +103,8 @@ function getDownloadUrl(browserName: BrowserName, revision: number, platform: Br
         ['mac10.13', '%s/builds/firefox/%s/firefox-mac-10.14.zip'],
         ['mac10.14', '%s/builds/firefox/%s/firefox-mac-10.14.zip'],
         ['mac10.15', '%s/builds/firefox/%s/firefox-mac-10.14.zip'],
-        ['mac11.0', '%s/builds/firefox/%s/firefox-mac-10.14.zip'],
-        ['mac11.0-arm64', '%s/builds/firefox/%s/firefox-mac-10.14.zip'],
-        ['mac11.1', '%s/builds/firefox/%s/firefox-mac-10.14.zip'],
-        ['mac11.1-arm64', '%s/builds/firefox/%s/firefox-mac-10.14.zip'],
+        ['mac11', '%s/builds/firefox/%s/firefox-mac-10.14.zip'],
```
dgozman, line 106: This all looks like a mess...

dgozman (approved):

PR #5025, opened 2021-01-15
feat(trace): show dialogs, navigations and misc events

dgozman:


src/trace/traceTypes.ts
```diff
@@ -77,14 +84,53 @@ export type ActionTraceEvent = {
...
+export type NavigationEvent = {
+  timestamp: number,
+  type: 'navigation',
+  contextId: string,
+  pageId: string,
+  url: string,
+  sameDocument: boolean,
+};
+
+export type MiscEvent = {
```
JoelEinbinder, line 113: I'd rather these be split into standalone load and error events.

dgozman: Done.

JoelEinbinder (approved):

PR #5031, opened 2021-01-15
docs: brush up some python docs (2)

pavelfeldman:


docs/src/api/python.md
```diff
@@ -1,22 +1,3 @@
-## async method: Playwright.start
```
yury-s, line 1: Why did it go away?

yury-s (approved):

docs/src/api/python.md
pavelfeldman: It does not exist!

dgozman (approved):

PR #5036, opened 2021-01-16
fix(dom): make selectOption wait for options

### LINKED ISSUE #4921: [BUG] selectOption doesn't auto-wait for the options being selected
Playwright docs talk a lot about reliable execution by auto-waiting for elements to be ready. Unfortunately `selectOption` doesn't live up to that. It does auto-wait for the given selector, but not for the values to be found inside that selector. E.g:

```javascript
  await page.selectOption('#my-select', {label: "badlabel"});
```

**Desired behavior**: `selectOption` waits until `badlabel` can be found, eventually throwing a TimeoutError.

**Actual behavior**: If `#my-select` is found, but `badlabel` is not, `selectOption` immediately clears the selection in `#my-select` and returns without throwing an error.

The current behavior leads to flaky executions in pages where options are dynamically added to `select` elements.

Even worse, it can lead to confusing and dangerous bugs by causing the wrong element to be selected. After clearing the selection on a select element (like `selectOption` does when there's no match), the next element added to the dropdown will automatically become the selection.

So you can end up with an arbitrary option in the dropdown being mistakenly selected. Imagine that  causing the wrong record in a database to be updated, or even deleted. Something similar to that just happened to me.

### Compatibility

Not sure the best way to handle backwards compatibility. Maybe make a new `selectOptionWait` function and deprecate the old one, or at least strongly recommend using the new one?

yury-s:
Ideally we'd use `InjectedScriptPollHandler` for polling options but that may result in an exception if selecting an option triggers navigation.

Fixes #4921

### DIFF
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -31,7 +31,7 @@ export type InjectedScriptProgress = {
 };
 
 export type InjectedScriptPoll<T> = {
-  result: Promise<T>,
+  run: () => Promise<T>,
   // Takes more logs, waiting until at least one message is available.
   takeNextLogs: () => Promise<string[]>,
   // Takes all current logs without waiting.
@@ -242,19 +242,23 @@ export class InjectedScript {
       },
     };
 
-    const result = task(progress);
+    const run = () => {
+      const result = task(progress);
 
-    // After the task has finished, there should be no more logs.
-    // Release any pending `takeNextLogs` call, and do not block any future ones.
-    // This prevents non-finished protocol evaluation calls and memory leaks.
-    result.finally(() => {
-      taskFinished = true;
-      logReady();
-    });
+      // After the task has finished, there should be no more logs.
+      // Release any pending `takeNextLogs` call, and do not block any future ones.
+      // This prevents non-finished protocol evaluation calls and memory leaks.
+      result.finally(() => {
+        taskFinished = true;
+        logReady();
+      });
+
+      return result;
+    };
 
     return {
       takeNextLogs,
-      result,
+      run,
       cancel: () => { progress.aborted = true; },
       takeLastLogs: () => unsentLogs,
     };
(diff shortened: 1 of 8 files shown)


src/server/injected/injectedScript.ts
```diff
@@ -267,35 +271,52 @@ export class InjectedScript {
...
+            matches = matches && optionToSelect.index === index;
+          return matches;
+        };
+        if (!remainingOptionsToSelect.some(filter))
+          continue;
+        selectedOptions.push(option);
+        if (select.multiple) {
+          remainingOptionsToSelect = remainingOptionsToSelect.filter(o => !filter(o));
+        } else {
+          remainingOptionsToSelect = [];
```
dgozman, line 305: This logic seems strange. I guess we should throw when multiple options are passed for non-multiple select? Otherwise there is no way we select them all.

src/server/injected/injectedScript.ts
```diff
@@ -267,35 +271,52 @@ export class InjectedScript {
...
+          break;
+        }
+      }
+      if (remainingOptionsToSelect.length) {
+        progress.logRepeating('    did not find some options - waiting... ');
+        return continuePolling;
+      }
+      select.value = undefined as any;
+      selectedOptions.forEach(option => option.selected = true);
+      progress.log('    selected soecified option(s)');
```
dgozman, line 315: typo: soecified

dgozman (approved):

src/server/injected/injectedScript.ts
yury-s: done.

src/server/injected/injectedScript.ts
yury-s: From our docs: " If the <select> has the multiple attribute, all matching options are selected, otherwise only the first option matching one of the passed options is selected."

PR #5038, opened 2021-01-16
feat(trace viewer): improve source tab

dgozman:
- Show stack trace and allow to select one.
- Fix an issue when the page is closed before action end and we lack an id.
- Fix timeline time labels.

### DIFF
--- a/src/cli/traceViewer/web/ui/sourceTab.tsx
+++ b/src/cli/traceViewer/web/ui/sourceTab.tsx
@@ -21,27 +21,82 @@ import './sourceTab.css';
 import '../../../../third_party/highlightjs/highlightjs/tomorrow.css';
 import * as highlightjs from '../../../../third_party/highlightjs/highlightjs';
 
+type StackInfo = string | {
+  frames: {
+    filePath: string,
+    fileName: string,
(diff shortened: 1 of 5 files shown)


pavelfeldman (approved):

src/cli/traceViewer/web/ui/sourceTab.tsx
```diff
@@ -21,27 +21,82 @@ import './sourceTab.css';
...
+  fileContent: Map<string, string>;
+};
+
 export const SourceTab: React.FunctionComponent<{
   actionEntry: ActionEntry | undefined,
 }> = ({ actionEntry }) => {
-  const location = React.useMemo<{ fileName?: string, lineNumber?: number, value?: string }>(() => {
+  const [lastAction, setLastAction] = React.useState<ActionEntry | undefined>();
+  const [selectedFrame, setSelectedFrame] = React.useState<number>(0);
+  const [needReveal, setNeedReveal] = React.useState<boolean>(false);
```
mxschmitt, line 39: note/nit: React internally infers the types automatically if you pass them into useState. For cases to allow e.g. null you need to use their generic.

PR #5050, opened 2021-01-19
feat(pause): page.pause to wait for user to click resume

JoelEinbinder:
This is to replace the pattern `await new Promise(() => {})` that we see when authoring tests.

 - Is `pause` confusing because it implies that the javascript should be paused?
 - I left the UI simple. Its just a big ▶️ emoji. I assume the UI experts here will have opinions.
 - Should the UI be made optional? I can't imagine a situation where you want to pause but can't afford to have anything covering a small part of the screen. But maybe it would be a good explainer option.
 - Folio should disable timeouts when the user pauses the page, which implies that there should be an event emitted when a page is paused/resumed. Would other clients make use of this, or should we make it internal?
 - Maybe it should be `context.pause`? The ergonomics are kind of annoying for that, but since we are waiting in the node process we are likely pausing every page anyway.
 - I want to start codegen from within my pause ui. But I'm not sure where to put the output because we don't control the terminal.

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -1509,6 +1509,15 @@ The page's main frame. Page is guaranteed to have a main frame which persists du
 
 Returns the opener for popup pages and `null` for others. If the opener has been closed already the returns `null`.
 
+## async method: Page.pause
+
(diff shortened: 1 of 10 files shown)


pavelfeldman (approved):
Could you name it _pause for now?

PR #5054, opened 2021-01-19
docs: add file chooser example, remove links to js samples

pavelfeldman:


docs/src/input.md
```diff
@@ -485,14 +487,33 @@ page.set_input_files(
 )
 ```
 
-You can select input files for upload using the [`method: Page.setInputFiles`] method. It expects first argument to point to an [input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) with the type `"file"`. Multiple files can be passed in the array. If some of the file paths are relative, they are resolved relative to the [current working directory](https://nodejs.org/api/process.html#process_process_cwd). Empty array clears the selected files.
+You don't have input element in hand (it is created dynamically), you can handle the [`event: Page.filechooser`] event
```
dgozman, line 490: If you don't ...

docs/src/input.md
```diff
@@ -429,6 +429,8 @@ Note that you still need to specify the capital `A` in `Shift-A` to produce the
 
 ## Upload files
 
+You can select input files for upload using the [`method: Page.setInputFiles`] method. It expects first argument to point to an [input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) with the type `"file"`. Multiple files can be passed in the array. If some of the file paths are relative, they are resolved relative to the [current working directory](https://nodejs.org/api/process.html#process_process_cwd). Empty array clears the selected files.
```
dgozman, line 432: Remove the node.js link to cwd.

dgozman (approved):

PR #5081, opened 2021-01-21
feat(selectors): nth-match selector

### LINKED ISSUE #3697: [Feature] allow selecting the second (k-th) selector match.

dgozman:
Introduces `:nth-match(ul > li, 3)` css extension, with one-based index.

Fixes #3697.

### DIFF
--- a/docs/src/selectors.md
+++ b/docs/src/selectors.md
@@ -284,6 +284,59 @@ converts `'//html/body'` to `'xpath=//html/body'`.
 
 Attribute engines are selecting based on the corresponding attribute value. For example: `data-test-id=foo` is equivalent to `css=[data-test-id="foo"]`, and `id:light=foo` is equivalent to `css:light=[id="foo"]`.
 
+## Pick n-th match from the query result
+
+Sometimes page contains a number of similar elements, and it is hard to select a particular one. For example:
+
+```html
+<section> <button>Buy</button> </section>
+<article><div> <button>Buy</button> </div></article>
+<div><div> <button>Buy</button> </div></div>
+```
+
+In this case, `:nth-match(:text("Buy"), 3)` will select the third button from the snippet above. Note that index is one-based.
+
+```js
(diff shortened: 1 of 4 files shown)


docs/src/selectors.md
```diff
@@ -284,6 +284,61 @@ converts `'//html/body'` to `'xpath=//html/body'`.
 
 Attribute engines are selecting based on the corresponding attribute value. For example: `data-test-id=foo` is equivalent to `css=[data-test-id="foo"]`, and `id:light=foo` is equivalent to `css:light=[id="foo"]`.
 
+## Selecting element from the list
```
pavelfeldman, line 287: I find this title misleading - selecting element from the list can be done with nth-child. You should name it "Pich n-th match from the query result" or something like this.

docs/src/selectors.md
```diff
@@ -284,6 +284,61 @@ converts `'//html/body'` to `'xpath=//html/body'`.
...
+
+Sometimes page contains a number of indistinguishable elements. For example:
+
+```html
+<ul>
+  <li>Item<button>Buy</button></li>
```
pavelfeldman, line 293: Also not a great example.

docs/src/selectors.md
```diff
@@ -284,6 +284,61 @@ converts `'//html/body'` to `'xpath=//html/body'`.
...
+  <li>Item<button>Buy</button></li>
+  <li>Item<button>Buy</button></li>
+</ul>
+```
+
+Playwright supports `:nth-match(inner selector, index)` to target a particular item in this case. It will query the inner selector and return an element with the specified `index` from the query result. Note that index is one-based.
```
pavelfeldman, line 299: It is unclear what `inner selector` is in this context, let's steer this from being a spec to being a guide and start with the example that you have below.

pavelfeldman (approved):

PR #5082, opened 2021-01-21
chore: add storybook dep

pavelfeldman:


src/third_party/highlightjs/highlightjs/index.d.ts
```diff
@@ -235,14 +235,6 @@ declare module 'highlight.js' {
     export = hljs;
 }
 
-declare module 'highlight.js/lib/core' {
```
JoelEinbinder, line 238: it looks like our webpack is configured in a strange way, which I think is why these are causing issues. I'll look into changing it after you land this to avoid rebasing.

That being said, these module declarations are incorrect since we don't have highlightjs in node_modules. So you should delete the `declare module` statements above and below as well.

dgozman (approved):

PR #5089, opened 2021-01-21
chore(dotnet): Dotnet Generation from Markdown

avodovnik:
This PR introduces a generator using the markdown specs of our API, to generate the API surface for the .NET Language Port. You can see the surface API suggestion in the corresponding port PR (https://github.com/microsoft/playwright-sharp/pull/1113). 

The generator isn't entirely complete. This is a list of things I'll still need to get around to:
- generating the value for parameter documentation
- re-parsing `<paramref>` links to arguments that might have been renamed in the generator code


utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  let assumedName = name.charAt(0).toUpperCase() + name.substring(1);
+  switch (memberKind) {
+    case "interface":
+      return `I${assumedName}`;
+    case "method":
+      return `${assumedName}Async`;
+    case "event":
+      return `On${assumedName}`;
```
kblok, line 277: This will be quite breaking 🤓

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,333 @@
...
+  const out = [];
+
+  /**** METHODS  ***/
+  member.methodsArray.forEach(method => {
+    let name = translateMemberName(method.kind, method.name, method);
+    let returnType = "Task";
+
+    if (method.type.name !== 'void') {
+      returnType = `Task<${translateType(method.type)}>`;
+    }
```
kblok, line 314: I think we have an `async` flag.

utils/doclint/generateDotnetApi.js
avodovnik: I agree. I'll get rid of it and keep using the _actual_ name from the docs.

utils/doclint/generateDotnetApi.js
avodovnik: What do you mean?

avodovnik:
For reviewers, ignore `xmlDocumentation.js` as that's a part of a separate PR (https://github.com/microsoft/playwright/pull/5164) and I'll drop the commit from here when I merge that one. I expect to have at least some remarks here though :-)

docs/src/api/class-browsercontext.md
```diff
@@ -216,6 +216,8 @@ The default browser context cannot be closed.
 :::
 
 ## async method: BrowserContext.cookies
+* langs:
+  - alias-csharp: GetCookiesAsync
```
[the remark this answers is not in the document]
avodovnik: I do, for most use cases (https://github.com/microsoft/playwright/blob/b7125097693c587598741f79a7e2e703b492f6c7/utils/doclint/generateDotnetApi.js#L399) but this one is special, because it also accepts an argument. 

I'd much rather have this alias'd than hard-code an edge case in the generator.

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+ */
+function generateNameDefault(member, name, t, parent) {
+  if (!t.properties
+    && !t.templates
+    && !t.union
+    && t.expression === '[Object]')
+    return 'object';
+
+  // we'd get this call for enums, primarily
+  let enumName = generateEnumNameIfApplicable(member, name, t, parent);
```
yury-s, line 298: How about this:

```js
if (enumName)
  return enumName;

if (!member)
  return t.name;
```

to simplify remaining part of the method?

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+
+    if (innerType.expression === '[Object]<[string], [string]>') {
+      // do nothing, because this is handled down the road
+    } else if (!isArray && !innerType.properties) {
+      type = `dynamic`;
+    } else {
+      type = classNameMap.get(innerType.name);
+      if (!type) {
+        type = typeResolve(innerType);
+        // additionalTypes.set(type, innerType);
```
yury-s, line 379: remove commented code

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+  }
+
+  // HACK: special case for generics handling!
+  if (type === 'T') {
+    name = `${name}<T>`;
+  }
+
+  // adjust the return type for async methods
+  if (member.async)
+    if (type === 'void')
```
yury-s, line 409: nit: {} for multi-line if body

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+
+  let parseArg = (/** @type {Documentation.Member} */ arg) => {
+    if (arg.name === "options") {
+      arg.type.properties.forEach(prop => {
+        parseArg(prop);
+      });
+      return;
+    }
+
+    if (arg.type.expression === '[string]|[path]') {
```
yury-s, line 438: Can we define this as overloaded params in the documentation? I don't quite understand what should happen with this logic in this case of setInputFiles files parameter ( <[path]|[Array]<[path]>|[Object]|[Array]<[Object]>>).

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+        pushArg(newArg, newArgName, arg);
+        addParamsDoc(newArgName, argDocumentation);
+      }
+      return;
+    }
+
+    addParamsDoc(argName, XmlDoc.renderTextOnly(arg.spec, maxDocumentationColumnWidth));
+
+    if (argName === 'timeout' && argType === 'decimal') {
+      args.push(`int timeout = 0`); // a special argument, we ignore our convention
```
yury-s, line 485: timeout is double everywhere else, should be decimal in c# too I believe.

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+ */
+
+/**
+ *  @param {Documentation.Type} type 
+ *  @param {Documentation.Class|Documentation.Type} parent
+ *  @param {generateNameCallback} generateNameCallback
+*/
+function translateType(type, parent, generateNameCallback = null) {
+
+  if (!generateNameCallback) {
```
yury-s, line 521: just do it in the params list:  `..., generateNameCallback = t => t.name) {` ?

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+    generateNameCallback = (t) => {
+      return t.name;
+    }
+  }
+
+  // a few special cases we can fix automatically
+  if (type.expression === '[null]|[Error]')
+    return 'void';
+  else if (type.expression === '[boolean]|"mixed"')
+    return 'MixedState';
```
yury-s, line 531: let's change accessibility snapshot type to string and be done with this "mixed" state nonsense? that's what we ended up doing in java and what Joel suggested

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+          || innerTypeName === 'int') {
+          return `${innerTypeName}?`;
+        }
+
+        // if it's not a value type, it'll be nullable by default, so we can ignore it
+        return `${innerTypeName}`;
+      }
+    }
+
+    if (type.union.filter(u => u.name.startsWith(`"`)).length == type.union.length
```
yury-s, line 556: You can detect both nullable and non-nullable union by looking at `type.unionName` which is always not null for the head element or for the second element for nullable unions (see https://github.com/microsoft/playwright/blob/96e099aca32f87f2d88c1cdb8acbddfae645b7c7/utils/doclint/documentation.js#L432)

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+      if (isNullableEnum)
+        return `${enumName}?`;
+      return enumName;
+    }
+
+    if (type.expression === '[string]|[Buffer]')
+      return `byte[]`; // TODO: make sure we implement extension methods for this!
+    else if (type.expression === '[string]|[float]'
+      || type.expression === '[string]|[float]|[boolean]'
+      || (type.union && type.union[0].name === 'string' && type.union[1].name === 'float' && type.union[2].name === 'boolean'))
```
yury-s, line 586: is there a difference between last two conditions?

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+      return 'string'; // because users can send either 100 or 100px for most of these arguments
```
yury-s, line 587: if this is for Page.pdf.margin all of these have type string for c# and java already

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+    else if (type.union.length == 2 && type.union[1].name === 'Array' && type.union[1].templates[0].name === type.union[0].name)
+      return `IEnumerable<${type.union[0].name}>`; // an example of this is [string]|[Array]<[string]>
```
yury-s, line 589: does this automatically accept single element?

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+      return enumName;
+    }
+
+    if (type.expression === '[string]|[Buffer]')
+      return `byte[]`; // TODO: make sure we implement extension methods for this!
+    else if (type.expression === '[string]|[float]'
+      || type.expression === '[string]|[float]|[boolean]'
+      || (type.union && type.union[0].name === 'string' && type.union[1].name === 'float' && type.union[2].name === 'boolean'))
+      return 'string'; // because users can send either 100 or 100px for most of these arguments
+    else if (type.union.length == 2 && type.union[1].name === 'Array' && type.union[1].templates[0].name === type.union[0].name)
```
yury-s, line 588: it's somewhat brittle (think about setInputFiles and selectOption)

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+      return `IEnumerable<${type.union[0].name}>`; // an example of this is [string]|[Array]<[string]>
+    else if (type.union[0].name === 'path')
+      // we don't support path, but we know it's usually an object on the other end, and we expect
```
yury-s, line 591: again, out of curiosity, why .net doesn't support path parameters?

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+    }
+    // this is an additional type that we need to generate
+    let objectName = generateNameCallback(type);
+    if (objectName === 'Object') {
+      throw 'Object unexpected';
+    } else if (type.name === 'Object') {
+      registerAdditionalType(objectName, type);
+    }
+
+    if (['object'].includes(objectName))
```
yury-s, line 631: Are both of these cases possible (objectName === 'Object' ` above and `objectName === 'object' ` here) ?

yury-s (approved):

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+  // map some types that we know of
+  classNameMap.set('Error', 'Exception');
+  classNameMap.set('TimeoutError', 'TimeoutException');
+  classNameMap.set('EvaluationArgument', 'object');
+  classNameMap.set('boolean', 'bool');
+  classNameMap.set('Serializable', 'T');
+  classNameMap.set('any', 'object');
+  classNameMap.set('Buffer', 'byte[]');
+  classNameMap.set('path', 'string');
+  classNameMap.set('URL', 'Uri');
```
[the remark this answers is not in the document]
avodovnik: This is an interesting debate - in theory, I agree with the recommendation from the SDK team, that we should use idiomatic concepts from languages as much as possible. So I wanted to do the same, and we can introduce separate overloads later, that are *not* auto-generated, that simplify the usage.

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+    if (member.kind === 'event') {
+      return `${name}Payload`;
+    }
+  }
+
+  return enumName || t.name;
+}
+
+function generateEnumNameIfApplicable(member, name, type, parent) {
+  if (!type.union)
```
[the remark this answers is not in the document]
avodovnik: Ah, awesome! Must have missed that change - modified now.

utils/doclint/generateDotnetApi.js
avodovnik: Yeah, I've added the handling when `name` exists. I _sort of_ want to keep the rest, just in case, but also, probably best to just `throw` if an enum without a name happens...

utils/doclint/generateDotnetApi.js
```diff
@@ -0,0 +1,695 @@
...
+  classNameMap.set('TimeoutError', 'TimeoutException');
+  classNameMap.set('EvaluationArgument', 'object');
+  classNameMap.set('boolean', 'bool');
+  classNameMap.set('Serializable', 'T');
+  classNameMap.set('any', 'object');
+  classNameMap.set('Buffer', 'byte[]');
+  classNameMap.set('path', 'string');
+  classNameMap.set('URL', 'Uri');
+  classNameMap.set('RegExp', 'Regex');
+  classNameMap.set('float', 'decimal');
```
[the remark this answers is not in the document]
kblok: That would be an accuracy vs precision question.
I wouldn't use double if there is no math involved. And if we would move to double, even a `Single` would be a better fit for our numbers.

utils/doclint/generateDotnetApi.js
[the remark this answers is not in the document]
kblok: I think this compares to the Path/FileInfo debate.
Using Uri and FileInfo will make the user write more code (`new Uri(myUrlString)` and `new FileInfo(myFilePath)`), and maybe requiring the user to add a new using. On the other side we would do a ToString and pass it to the diver. 
The user losses we lose.

utils/doclint/generateDotnetApi.js
kblok: why is a double, can we pass milliseconds with decimals?

utils/doclint/generateDotnetApi.js
kblok: Related to the Uri comment, what Java users win with a Path?

utils/doclint/generateDotnetApi.js
avodovnik: Yeah, this was changed as I was working on the generator :-). The question now is, do I leave that in, or assume this will be adhered to as we add stuff... I could just throw an exception and make sure we fix it.

utils/doclint/generateDotnetApi.js
avodovnik: No, but (1) doing `new [] { element }` isn't uncommon for C# devs, and (2) we can introduce overloads later manually which will be cheaper to maintain than doing this in the generator.

utils/doclint/generateDotnetApi.js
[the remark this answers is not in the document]
avodovnik: At the moment, PlaywrightSharp uses `decimal` and I felt like it's worth keeping that. That said, maybe it's worth migrating to `Single`. That is technically the closest to a `float` (khm, it is a float, as they're both equally precise and have the same size).

utils/doclint/generateDotnetApi.js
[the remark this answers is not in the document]
avodovnik: @kblok well, the PR I had open to change `Path` into `FileInfo` was closed :) so, right now, that resolves to `string`.  Despite the advice, it might be beneficial for us to use `string`...  As @yury-s suggested, it results in easer to use API.

utils/doclint/generateDotnetApi.js
avodovnik: This looks like a remnant of a copy-paste/refactoring effort :D

PR #5090, opened 2021-01-21
fix: don't parse potentially invalid urls in event handlers

### LINKED ISSUE #5065: [BUG] Malformed URL leads to UnhandledPromiseRejection
**Context:**
- Playwright Version: 1.7.1
- Operating System: both on Mac running locally, and when ran via a Google CloudFunction on a linux based docker
- Node.js version: 13.14.0 locally, 12.20.0 on Google CloudFunction
- Browser: chromium
- Extra: We couldn't get the CloudFunction to run with just Playwright, but installing Chromium via Puppeteer and using that executable worked 😆 . Locally I am using just Playwright.

**Code Snippet**

```javascript

(async () => {
  const page = await browser.newPage({ ignoreHTTPSErrors: true });
  await page.goto('https://www.youtube Panel Title.com/'); // Notice the spaces in the URL
})();
```

**Describe the bug**
When passing a malformed URL to the page.goto method, an UnhandledPromiseRejection is thrown, even if I wrap my code in a try-catch block. I got around this issue by trying to create a URL object from the string I get before trying to open it:

```javascript
try {
  new URL(url);
} catch (e) {
  console.warn('Malformed URL', url);
  return;
}
```

The same error occurs occasionally, when the page I load has a well formed URL, but one of the assets it tries to load doesn't. This crashes the running function, even though all of my code is wrapped in a try-catch block, or uses await-to-js to catch rejected promises.

Add any other details about the problem here.

This is the error log I get when trying to open a malformed URL.
```
(node:23419) UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_URL]: Invalid URL: https://www.youtube%20panel%20title.com/
    at onParseError (internal/url.js:257:9)
    at new URL (internal/url.js:333:5)
    at stripFragmentFromUrl (/services/browser/node_modules/playwright-chromium/lib/server/network.js:63:20)
    at new Request (/services/browser/node_modules/playwright-chromium/lib/server/network.js:83:21)
    at new InterceptableRequest (/services/browser/node_modules/playwright-chromium/lib/server/chromium/crNetworkManager.js:364:24)
    at CRNetworkManager._onRequest (/services/browser/node_modules/playwright-chromium/lib/server/chromium/crNetworkManager.js:235:25)
    at CRNetworkManager._onRequestWillBeSent (/services/browser/node_modules/playwright-chromium/lib/server/chromium/crNetworkManager.js:114:18)
    at CRSession.emit (events.js:315:20)
    at CRSession.EventEmitter.emit (domain.js:485:12)
    at /services/browser/node_modules/playwright-chromium/lib/server/chromium/crConnection.js:155:22
```

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/5065

### DIFF
--- a/src/server/network.ts
+++ b/src/server/network.ts
@@ -61,12 +61,18 @@ export function rewriteCookies(cookies: types.SetNetworkCookieParam[]): types.Se
   });
 }
 
-function stripFragmentFromUrl(url: string): string {
-  if (!url.indexOf('#'))
+export function parsedURL(url: string): URL | null {
+  try {
+    return new URL(url);
+  } catch (e) {
+    return null;
+  }
+}
+
+export function stripFragmentFromUrl(url: string): string {
+  if (!url.includes('#'))
     return url;
-  const parsed = new URL(url);
-  parsed.hash = '';
-  return parsed.href;
+  return url.substring(0, url.indexOf('#'));
 }
 
 export class Request {
--- a/src/cli/traceViewer/snapshotRouter.ts
+++ b/src/cli/traceViewer/snapshotRouter.ts
@@ -18,6 +18,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as util from 'util';
 import type { Route } from '../../..';
+import { parsedURL } from '../../client/clientHelper';
 import type { FrameSnapshot, NetworkResourceTraceEvent, PageSnapshot } from '../../trace/traceTypes';
 import { ContextEntry } from './traceModel';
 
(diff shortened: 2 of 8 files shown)


src/server/network.ts
```diff
@@ -61,12 +61,18 @@ export function rewriteCookies(cookies: types.SetNetworkCookieParam[]): types.Se
   });
 }
 
-function stripFragmentFromUrl(url: string): string {
+export function parsedURL(url: string): URL | null {
```
dgozman, line 64: Let's move all these methods to utils/ to share between client, server, cli and everyone.

dgozman (approved):

PR #5139, opened 2021-01-25
chore: add Python to docker images

aslushnikov:
This adds +100MB to network transfer size but enables our
users to use playwright-python from-inside docker container.

### DIFF
--- a/utils/docker/Dockerfile.bionic
+++ b/utils/docker/Dockerfile.bionic
@@ -68,6 +68,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends git ssh && \
 # Create the pwuser (we internally create a symlink for the pwuser and the root user)
 RUN adduser pwuser
 
+# Install Python 3.8
+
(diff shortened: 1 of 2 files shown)


yury-s (approved):
What about bionic image?

mxschmitt:
Is the package-lock.json change intended?

Also not sure about the handling there so how Java, C# would handle that, should that also be added as a dependency to the image? (feel free to follow up offline)

aslushnikov:
> What about bionic image?

@yurys I originally thought that it's hard to get Python 3.8 on bionic, but turns out it's not. Added it as well.

> Is the package-lock.json change intended?

These are reverted!

> Also not sure about the handling there so how Java, C# would handle that, should that also be added as a dependency to the image?

There are no plans to add these to the image atm, and there's not much interest for that. This is a shortcut for us to get python running without too much investment.

Also, there are some discussions on monorepoing everything. If this happens, it'll simplify management and we'll be able to have multiple images next to each other.

PR #5144, opened 2021-01-25
docs(api): use string enum for Dialog.type

yury-s:


pavelfeldman (approved):

yury-s:
Discussed offline. Will make it a string in java instead.

PR #5164, opened 2021-01-27
chore(docs): add ability to generate xmldocs

avodovnik:
This introduces a naive [XmlDocs](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/) renderer for Markdown nodes. It needs to be called separately for now, i.e. the `generateSourceCodeComment` in `Documentation` doesn't integrate with it yet, but I'll add that somewhere down the line.

### DIFF
--- a/utils/doclint/xmlDocumentation.js
+++ b/utils/doclint/xmlDocumentation.js
@@ -1,0 +1,197 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @ts-check
+
+const Documentation = require('./documentation');
+const { visitAll } = require('../markdown');
+/**
+ * @param {Documentation.MarkdownNode[]} nodes
+ * @param {number} maxColumns
+ */
+function renderXmlDoc(nodes, maxColumns) {
+    const summary = [];
+    const examples = [];
+    /** @type {Documentation.MarkdownNode} */
+    let lastNode;
+
+    summary.push('<summary>');
+
+    visitAll(nodes, (node, depth) => {
+        lastNode = innerRenderXmlNode(node, lastNode, summary, examples, maxColumns);
+    });
+
+    if (summary.length == 1) { // just the <summary> node 
+        return [];
+    }
+
+    // we might have a stray list, if the <li> is the last element
+    if (lastNode && lastNode.type === 'li') {
+        summary.push('</list>');
+    }
+    summary.push('</summary>');
+
+    // add examples
+    summary.push(...examples);
+    return summary.map(n => `/// ${n}`);
+}
+
+/**
+ * @param {Documentation.MarkdownNode} node
+ * @param {Documentation.MarkdownNode} lastNode
+ * @param {number=} maxColumns
+ * @param {string[]} summary
+ * @param {string[]} examples
+ */
+function innerRenderXmlNode(node, lastNode, summary, examples, maxColumns) {
+    /** @param {string[]} a */
+    const newLine = (a) => {
+        if (a[a.length - 1] !== '')
+            a.push('');
+    };
+
+    let escapedText = node.text;
+    // resolve links (within [])
+
+    if (node.type === 'text') {
+        // clear up the list, if there was one
+        if (lastNode && lastNode.type === 'li') {
+            summary.push('</list>');
+        }
+
+        summary.push(...wrapText(escapedText, maxColumns));
+
+        return node;
+    }
+
+    if (node.type === 'li') {
+        if (escapedText.startsWith('extends: ')) {
+            summary.push(...wrapText(`<seealso cref="${escapedText.substring(9)}"/>`, maxColumns));
+            return undefined;
+        }
+
+        // if the previous node was not li, start list
+        if (lastNode && lastNode.type !== 'li') {
+            summary.push(`<list>`);
+        }
+
+        summary.push(...wrapText(`<item><description>${escapedText}</description></item>`, maxColumns));
+    }
+
+    // TODO: this should really move to the examples array
+    if (node.type == 'code' && node.codeLang == "csharp") {
+        summary.push('<![CDATA[');
+        summary.push(...node.lines);
+        summary.push(']]>');
+    }
+
+    return node;
+}
+
+/**
+ * @param {string} text
+ * @param {number=} maxColumns
+ * @param {string=} prefix
+ */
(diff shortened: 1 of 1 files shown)


utils/doclint/xmlDocumentation.js
```diff
@@ -0,0 +1,197 @@
...
+    /** @type {Documentation.MarkdownNode} */
+    let lastNode;
+
+    summary.push('<summary>');
+
+    visitAll(nodes, (node, depth) => {
+        lastNode = innerRenderXmlNode(node, lastNode, summary, examples, maxColumns);
+    });
+
+    if (summary.length == 1) { // just the <summary> node
```
pavelfeldman, line 37: Why do we need to handle this edge case?

utils/doclint/xmlDocumentation.js
```diff
@@ -0,0 +1,197 @@
...
+    visitAll(nodes, (node, depth) => {
+        lastNode = innerRenderXmlNode(node, lastNode, summary, examples, maxColumns);
+    });
+
+    if (summary.length == 1) { // just the <summary> node 
+        return [];
+    }
+
+    // we might have a stray list, if the <li> is the last element
+    if (lastNode && lastNode.type === 'li') {
```
pavelfeldman, line 42: It looks like md's list items differ from the xmldoc ones in a way that it wants a wrapping container. I would render lists as md, i.e. via prepending `- `. That'll erase this `lastNode` handling.

utils/doclint/xmlDocumentation.js
```diff
@@ -0,0 +1,197 @@
...
+// @ts-check
+
+const Documentation = require('./documentation');
+const { visitAll } = require('../markdown');
+/**
+ * @param {Documentation.MarkdownNode[]} nodes
+ * @param {number} maxColumns
+ */
+function renderXmlDoc(nodes, maxColumns) {
+    const summary = [];
```
pavelfeldman, line 26: Give us some 2 space indents!

utils/doclint/xmlDocumentation.js
```diff
@@ -0,0 +1,197 @@
...
+        if (a[a.length - 1] !== '')
+            a.push('');
+    };
+
+    let escapedText = node.text;
+    // resolve links (within [])
+
+    if (node.type === 'text') {
+        // clear up the list, if there was one
+        if (lastNode && lastNode.type === 'li') {
```
pavelfeldman, line 71: nit: we don't wrap single-liners with {}

utils/doclint/xmlDocumentation.js
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    if (node.type === 'li') {
+        if (escapedText.startsWith('extends: ')) {
+            summary.push(...wrapText(`<seealso cref="${escapedText.substring(9)}"/>`, maxColumns));
+            return undefined;
+        }
+
+        // if the previous node was not li, start list
+        if (lastNode && lastNode.type !== 'li') {
```
pavelfeldman, line 87: Yeah, definitely don't use <list>, your summary would actually read better without it.

utils/doclint/xmlDocumentation.js
```diff
@@ -0,0 +1,197 @@
...
+            summary.push(`<list>`);
+        }
+
+        summary.push(...wrapText(`<item><description>${escapedText}</description></item>`, maxColumns));
+    }
+
+    // TODO: this should really move to the examples array
+    if (node.type == 'code' && node.codeLang == "csharp") {
+        summary.push('<![CDATA[');
```
pavelfeldman, line 96: Sneaky! i like it!

utils/doclint/xmlDocumentation.js
```diff
@@ -0,0 +1,197 @@
...
+ * @param {string[]} examples
+ */
+function innerRenderXmlNode(node, lastNode, summary, examples, maxColumns) {
+    /** @param {string[]} a */
+    const newLine = (a) => {
+        if (a[a.length - 1] !== '')
+            a.push('');
+    };
+
+    let escapedText = node.text;
```
pavelfeldman, line 66: probably const escapedText = `escape(node.text)` ?

utils/doclint/xmlDocumentation.js
```diff
@@ -0,0 +1,197 @@
...
+
+    return node;
+}
+
+/**
+ * @param {string} text
+ * @param {number=} maxColumns
+ * @param {string=} prefix
+ */
+function wrapText(text, maxColumns = 0, prefix = '') {
```
pavelfeldman, line 109: It looks like you are inserting the xml tags and then fighting with it. I'd suggest that you only wrap user content instead. I.e. instead of
```
wrapText(`<item><description>${escapedText}</description></item>`, maxColumns)
```

do
```
`<item><description>\n${wrapText(escapedText, maxColumns)}\n</description></item>`)
```

That way your wrapText becomes trivial.

utils/doclint/xmlDocumentation.js
avodovnik: So, VS isn't _smart_ enough to actually render that in a meaningful way. So I'll always need to wrap it in a `<list>` if I want it to display at least somewhat nicely to the end user.

This:

```csharp
    /// <summary>
    /// Remember, that sometimes, you need to do the following:
    /// - Test 1
    /// - Test 2
    /// - Test 3
    /// - Test 4
    /// 
    /// But there are times, when you want to do
    /// <list type="bullet">
    /// <item><description>Something smart</description></item>
    /// <item><description>Something smarter</description></item>
    /// <item><description>The really smart thing</description></item>
    /// <item><description>Just a list</description></item>
    /// </list>
    /// Tis a thing <seealso 
    /// cref="Browser"/> test.
    /// <a href="https://www.google.com">This is a link.</a>
    /// </summary>
```
turns into this:

image

utils/doclint/xmlDocumentation.js
avodovnik: It's supposed to be a _cheap_ way of getting rid of:

```csharp
    /// <summary>
    /// </summary>
```

but instead of ignoring it, I might be better off outputting a warning as that means it's not documented.

utils/doclint/xmlDocumentation.js
avodovnik: I haven't ran into text I needed to escape, so I've omitted the escaping part for now.

JoelEinbinder (commented):
Code looks fine. I'd have some style nits, but I think they are all because this file is not covered by eslint. We should probably restore eslint for .js files or convert doclint to typescript.

utils/doclint/xmlDocumentation.js
avodovnik: Fixed now... :-) Looks like my settings were wrong. Fixed that as well.

utils/doclint/xmlDocumentation.js
avodovnik: I've mostly avoided doing that now, but then `eslint` seems to have thrown a fit. Either way, point taken, thanks!

utils/doclint/xmlDocumentation.js
avodovnik: I've reworked that code, so it's a bit more digestable, but I decided to keep using the `li` part for the reasons outlined above.

avodovnik:
@JoelEinbinder I reformatted using 2 space tabs, instead of 4, and I ran `eslint` hopefully fixing the majority of the nits. Hopefully this is approvable now, and I can improve it as I go along :-)

pavelfeldman (approved):

PR #5169, opened 2021-01-27
devops: fetch chromium checkout if it has not been before

aslushnikov:
Checkouts should reside outside of gihtub action working directory
so will be reused between builds.


browser_patches/chromium/build.sh
```diff
@@ -67,6 +67,11 @@ compile_chromium() {
   # Update Chromium checkout. One might think that this step should go to `prepare_checkout.sh`
   # script, but the `prepare_checkout.sh` is in fact designed to prepare a fork checkout, whereas
   # we don't fork Chromium.
+  if [[ ! -d "${CR_CHECKOUT_PATH}" ]]; then
+    mkdir "${CR_CHECKOUT_PATH}"
+    cd "${CR_CHECKOUT_PATH}"
+    fetch chromium
```
yury-s, line 73: maybe --no-hooks?

yury-s (approved):

Issue #4366, opened 2020-11-06
[BUG] Webkit in Docker only scroll once


diegofernandezmrf:
**Context:**
- Playwright Version: 1.5.2
- Operating System: Linux
- Node.js version: 12
- Browser: Chromium, WebKit
- Extra: It's the Docker image provided by you
```
  System:
    OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
    Memory: 2.25 GB / 3.84 GB
    Container: Yes
  Binaries:
    Node: 12.19.0 - /usr/bin/node
    Yarn: 1.22.10 - /usr/bin/yarn
    npm: 6.14.8 - /usr/bin/npm
  Languages:
    Bash: 5.0.17 - /usr/bin/bash
```
**Code Snippet**

I created a repository that could be used to easily reproduce the error: https://github.com/Marfeel/use-cases/tree/master/playwright-docker-webkit

but here is the gist of it

```javascript
test('the page should scroll down twice', async () => {
  await page.goto('http://localhost:8084')

  // the scroll is at the very top
  expect(await page.evaluate(() => window.scrollY)).toBe(0);
  // there is enough scroll heigth available
  expect(await page.evaluate(() => document.documentElement.scrollHeight)).toBeGreaterThan(400);

  await page.waitForFunction(() => window.scroll(0, 100) || true);
  // just to check the scroll worked
  expect(await page.evaluate(() => window.scrollY)).toBe(100);

  await page.waitForFunction(() => window.scroll(0, 200) || true);
  // this expectation should pass on host and fail on docker
  expect(await page.evaluate(() => window.scrollY)).toBe(200);
});

```

**Describe the bug**

My team was using playwright in our computers (all macs) but we wanted to make it part of our CI tests.
I changed the docker image we used to yours and tests that rely on scrolling to trigger some behaviour started failing in webkit (all tests run fine on chromium).
The problem seems to be that only the first call to `window.scroll(...)` is taking effect.

Please let me know if you need more information, and available to help with it.

JoelEinbinder:
I can repro most of this locally. Docker doesn't change the results, but it looks like there is something very strange with how scrolls are handled in our headless webkit with emulation. If I scroll 0 to 100 to 200, it ends up at 100. If I scroll 0 to 100 to 110, it ends at 10(?!).

Thanks for the report! I'm still debugging.

diegofernandezmrf:
[...]

PR #5173, opened 2021-01-27
browser(webkit): fix scrolling a second time on linux

JoelEinbinder:
This code doesn't appear to be needed. All tests pass without it, and it breaks scrolling for #4366.


yury-s:
I'm pretty sure GTK didn't work without that code.

JoelEinbinder:
All tests pass when I delete it. If theres another way to repro let me know please.

yury-s:
I've tried locally with cnn.com and github.com and iphone 11 emulation and didn't find any issues with video recording. The code would only run on the first window.scrollBy call though.

yury-s (approved):
Let's give it a try!

PR #5200, opened 2021-01-28
docs: support argument overrides

yury-s:
* Supported parameter overrides - default definition can can replaced by a language specific one. Override is matched with the default definition by either name or alias. This ensures that overriding parameter always goes in the same order as the default one,.
* Inverted `pageFunction` parameter definition: js is now an override, python version became default
* Use separate if  for `force_expression` in python to avoid confusion

Corresponding change to the docs site: https://github.com/microsoft/playwright.dev/pull/93


docs/src/api/javascript.md
```diff
@@ -0,0 +1,16 @@
+### param: ElementHandle.$eval.pageFunction = %%-js-evalonselector-pagefunction-%%
```
pavelfeldman, line 1: We should not allow overriding by alias, this is confusing. Let's always use original name instead.

docs/src/api/params.md
```diff
@@ -154,6 +154,38 @@ Toggles bypassing page's Content-Security-Policy.
 
 Sets a consistent viewport for each page. Defaults to an 1280x720 viewport. `null` disables the default viewport.
 
+## evaluate-expression
+* langs:
+  - alias-js: pageFunction
```
pavelfeldman, line 159: Follow up: we won't need alias here since we have enough info in the override itself.

utils/doclint/api_parser.js
```diff
@@ -122,7 +122,18 @@ class ApiParser {
     if (!method)
       throw new Error('Invalid method ' + match[2] + '.' + match[3]);
     if (match[1] === 'param') {
-      method.argsArray.push(this.parseProperty(spec));
+      const arg = this.parseProperty(spec);
+      const existingArg = method.argsArray.find(m => m.name === arg.name || m.langs.aliases && Object.values(m.langs.aliases).indexOf(arg.name) !== -1);
```
pavelfeldman, line 126: drop the aliases match here.

utils/doclint/api_parser.js
```diff
@@ -122,7 +122,18 @@ class ApiParser {
...
+      if (existingArg) {
+        for (const lang of arg.langs.only) {
+          if (existingArg.name !== arg.name && existingArg.langs.aliases[lang] !== arg.name)
```
pavelfeldman, line 129: Remove this check, it is to override in anextensible manner without the core knowing.

utils/doclint/api_parser.js
```diff
@@ -302,7 +313,8 @@ function extractLangs(spec) {
     return {
       only: only ? only.split(',') : undefined,
       aliases,
-      types: {}
+      types: {},
```
pavelfeldman, line 316: Do you want to migrate existingMemeber to `overrides` as well?

utils/doclint/generateApiJson.js
```diff
@@ -35,7 +35,7 @@ const PROJECT_DIR = path.join(__dirname, '..', '..');
   });
   documentation.generateSourceCodeComments();
   const result = serialize(documentation);
-  fs.writeFileSync(path.join(PROJECT_DIR, 'api.json'), JSON.stringify(result));
+  fs.writeFileSync(path.join(PROJECT_DIR, 'api.json'), JSON.stringify(result, null, 2));
```
pavelfeldman, line 38: revert this

utils/doclint/documentation.js
```diff
@@ -322,11 +323,12 @@ Documentation.Member = class {
         continue;
       if (arg.langs.aliases && arg.langs.aliases[lang])
         arg.alias = arg.langs.aliases[lang];
-      arg.filterForLanguage(lang);
-      arg.type.filterForLanguage(lang);
-      if (arg.name === 'options' && !arg.type.properties.length)
+      const overriddenArg = (arg.langs.overrides && arg.langs.overrides[lang]) || arg;
+      overriddenArg.filterForLanguage(lang);
+      overriddenArg.type.filterForLanguage(lang);
```
pavelfeldman, line 328: Drive-by: looks like we are already doing this in line 319, so drop this line.

pavelfeldman (approved):

docs/src/api/javascript.md
yury-s: Done.

docs/src/api/params.md
yury-s: Done.

utils/doclint/api_parser.js
yury-s: Done.

utils/doclint/api_parser.js
yury-s: Done.

utils/doclint/documentation.js
yury-s: Done.

utils/doclint/generateApiJson.js
yury-s: Done.

PR #5205, opened 2021-01-28
feat(trace viewer): Adds _debugName BrowserContextOption to let users define a name for their contexts

### LINKED ISSUE #5157: [Feature] Ability to provide user-defined name for Tracers
In the trace viewer, I'd like to be able to see traces for all my executed tests, with the ability to quickly identify which trace relates to which test. At the moment, each trace for a single test is defined in a separate file, and its name is a random guid. I would like to specify a name for that trace. The name doesn't have to be reflected in the filename, but when I open trace-viewer, I would like to be able to select traces by name in the dropdown highlighted on the screenshot below.

Zrzut ekranu 2021-01-26 o 17 54 22

I'm happy to contribute code that would provide such functionality, but before going too far, I wanted to see if such feature makes sense for others as well.

domderen:
This change is adding a new property on the BrowserContextOptions class called `_debugName`. This property allows defining a user-friendly name for the browser context, and currently it is being used in one place, the Trace Viewer. When user provides the new value in the following way:

```typescript

(async () => {
  const context = await browser.newContext({ _traceDir: __dirname, _debugName: 'My custom testcase name' });
  await context.close();
  await browser.close();
})();
```

The `_debugName` will be saved in the `*.trace` file for this browser context, on the `context-created` event, under the key `debugName`.

Later, when such a trace is displayed using Trace Viewer, the `debugName` will be displayed in the dropdown in the top right part of the app instead of the actual trace filename.

Fixes #5157.

Zrzut ekranu 2021-01-28 o 13 52 59

### DIFF
--- a/src/protocol/protocol.yml
+++ b/src/protocol/protocol.yml
@@ -373,6 +373,7 @@ BrowserType:
           - no-preference
         acceptDownloads: boolean?
         _traceDir: string?
+        _debugName: string?
         recordVideo:
           type: object?
           properties:
@@ -445,6 +446,7 @@ Browser:
           - no-preference
         acceptDownloads: boolean?
         _traceDir: string?
+        _debugName: string?
         recordVideo:
           type: object?
           properties:
@@ -2352,6 +2354,7 @@ AndroidDevice:
           - no-preference
         acceptDownloads: boolean?
         _traceDir: string?
+        _debugName: string?
         recordVideo:
           type: object?
           properties:
--- a/test/trace.spec.ts
+++ b/test/trace.spec.ts
@@ -35,6 +35,7 @@ it('should record trace', (test, { browserName, platform }) => {
 
   const contextEvent = traceEvents.find(event => event.type === 'context-created') as trace.ContextCreatedTraceEvent;
   expect(contextEvent).toBeTruthy();
+  expect(contextEvent.debugName).toBeUndefined();
   const contextId = contextEvent.contextId;
 
   const pageEvent = traceEvents.find(event => event.type === 'page-created') as trace.PageCreatedTraceEvent;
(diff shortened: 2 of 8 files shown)


domderen:
Hey, two things to clarify here:

1) I went with the `_debugName` rather than `debugName` to follow the pattern set by `_traceDir`. I'm assuming it means that this option is not yet public and documented. Happy to change it, if this one is not that important at the moment.
2) In the tests file I saw this:

```typescript
it('should record trace', (test, { browserName, platform }) => {
  test.fixme();
}, async ({browser, testInfo, server}) => {
  ...
});
```

This looks like those tests are disabled on purpose at the moment. I added a new test checking the `_debugName` option, but annotated it with the same `fixme` call for now. Should those be enabled now?

domderen:
Oh I also didn't add the docs yet, again not sure if those get added once the functionality is ready for deployment, or if it should be added in this PR.

test/trace.spec.ts
```diff
@@ -111,3 +113,19 @@ it('should record trace with POST', (test, { browserName, platform }) => {
   expect(fs.existsSync(path.join(traceDir, 'resources', resourceEvent.requestSha1))).toBe(true);
   expect(fs.existsSync(path.join(traceDir, 'resources', resourceEvent.responseSha1))).toBe(true);
 });
+
+it('should record trace with a debugName', (test, { browserName, platform }) => {
+  test.fixme();
```
dgozman, line 118: Tracing tests are flaky on the bots, I still need to figure out why and reenable them.

src/protocol/protocol.yml
```diff
@@ -373,6 +373,7 @@ BrowserType:
           - no-preference
         acceptDownloads: boolean?
         _traceDir: string?
+        _debugName: string?
```
dgozman, line 376: Underscore is good. We'll make it public once we are sure about the api. No need for docs yet.

dgozman (approved):
This looks great!

test/trace.spec.ts
domderen: What do you mean by bots?

dgozman: I mean our GitHub actions.

domderen: Ah ok thanks

PR #5209, opened 2021-01-28
feat(element state): support aria-disabled attribute

TrySound:
Click must be blocked if aria-disabled="true" is specified.
isEnabled() and isDisabled() should consider aria-disabled
when computing value.

This will enable better support for custom accessible buttons.

### DIFF
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -662,7 +662,7 @@ export class InjectedScript {
 
   isElementDisabled(element: Element): boolean {
     const elementOrButton = element.closest('button, [role=button]') || element;
-    return ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName) && elementOrButton.hasAttribute('disabled');
+    return ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName) && elementOrButton.hasAttribute('disabled') || elementOrButton.getAttribute('aria-disabled') === 'true';
   }
 
   isElementReadOnly(element: Element): boolean {
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -662,7 +662,9 @@ export class InjectedScript {
 
   isElementDisabled(element: Element): boolean {
     const elementOrButton = element.closest('button, [role=button]') || element;
-    return ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName) && elementOrButton.hasAttribute('disabled') || elementOrButton.getAttribute('aria-disabled') === 'true';
+    const ariaDisabled = elementOrButton.getAttribute('aria-disabled')
+    return ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName) && elementOrButton.hasAttribute('disabled') ||
+      ariaDisabled != null && ariaDisabled.toLowerCase() !== 'false';
   }
 
   isElementReadOnly(element: Element): boolean {
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -662,9 +662,12 @@ export class InjectedScript {
 
   isElementDisabled(element: Element): boolean {
     const elementOrButton = element.closest('button, [role=button]') || element;
-    const ariaDisabled = elementOrButton.getAttribute('aria-disabled')
-    return ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName) && elementOrButton.hasAttribute('disabled') ||
-      ariaDisabled != null && ariaDisabled.toLowerCase() !== 'false';
+    if (['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName)) {
+      return elementOrButton.hasAttribute('disabled');
+    } else {
+      const ariaDisabled = elementOrButton.getAttribute('aria-disabled');
+      return ariaDisabled !== null && ariaDisabled.toLowerCase() !== 'false';
+    }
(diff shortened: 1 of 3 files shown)


src/server/injected/injectedScript.ts
```diff
@@ -662,7 +662,7 @@ export class InjectedScript {
 
   isElementDisabled(element: Element): boolean {
     const elementOrButton = element.closest('button, [role=button]') || element;
-    return ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName) && elementOrButton.hasAttribute('disabled');
+    return ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName) && elementOrButton.hasAttribute('disabled') || elementOrButton.getAttribute('aria-disabled') === 'true';
```
dgozman, line 665: - Let's not check `aria-disabled` for buttons, inputs, selects and textareas. AFAIU, they have no effect there.
- From my testing on Chromium, everything but "false", "FALSE" and other case variations is considered "disabled". Let's change the check accordingly.

test/elementhandle-convenience.spec.ts
```diff
@@ -176,6 +176,8 @@ it('isEnabled and isDisabled should work', async ({ page }) => {
     <button disabled>button1</button>
     <button>button2</button>
     <div>div</div>
+    <div role="button" aria-disabled="true">role=button1</div>
+    <div role="button" aria-disabled="false">role=button2</div>
```
dgozman, line 180: Let's add a testcase with `aira-disabled="foo"`, and change false case to `aria-disabled="faLSE"`

dgozman (commented):
Thank you for the PR! I went ahead and changed the title according to our [guidelines](https://github.com/microsoft/playwright/blob/master/CONTRIBUTING.md#commit-messages).

I have a few comments about edge cases. Otherwise, this should be good to go.

src/server/injected/injectedScript.ts
TrySound: I don't check aria-disabled for buttons. There is `||`. Or what do you mean?

dgozman: If you have `<button aria-disabled='true'>`, this will return `true` (because second branch is true). I believe that browsers will ignore aria-disabled in this case though, and the button will be considered enabled.

JoelEinbinder: aria never effects the behavior of the page, only the behavior of screen readers reading the page. So a button with aria-disabled=true will be clickable, but the screen reader will report it as not clickable.
image

src/server/injected/injectedScript.ts
```diff
@@ -662,7 +662,12 @@ export class InjectedScript {
 
   isElementDisabled(element: Element): boolean {
     const elementOrButton = element.closest('button, [role=button]') || element;
-    return ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName) && elementOrButton.hasAttribute('disabled');
+    if (['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(elementOrButton.nodeName)) {
+      return elementOrButton.hasAttribute('disabled');
+    } else {
+      const ariaDisabled = elementOrButton.getAttribute('aria-disabled');
```
JoelEinbinder, line 668: The logic for aria-disabled is actually more complicated than this. If the element is focusable, the disabled state can be inherited from parent elements.
See https://www.w3.org/TR/wai-aria-1.1/#aria-disabled and https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/accessibility/ax_object.cc;l=3171;drc=6cdb24a4ce9d4151035c1f133833137d2e2881d1

TrySound: Can `elementOrButton.closest('[aria-disabled]')` lookup be sufficient?

JoelEinbinder (approved):

src/server/injected/injectedScript.ts
JoelEinbinder: sorry for the late reply. The logic here would actually be more complicated, because you have to first check if the element takes focus, and only then do `element.closest`. I think its fine to do what you did and differ from the aria spec here. We can fix edge cases if people complain.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/5855

pavelfeldman:
@einbinder: could you rebase it in place and land?

gcvs:
I would replace line 421 of /src/server/injected/injectedScript.ts with

```
    const inputNode = ['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(element.nodeName);
    const disabled = (inputNode || element.hasAttribute('role')) && (element.getAttribute('aria-disabled') === 'true' ||
      inputNode && element.hasAttribute('disabled'));
```

From https://www.w3.org/TR/html-aria/

>Unless otherwise stated, authors MAY use aria-* attributes in place of their HTML equivalents on HTML elements where the aria-* semantics would be expected. For example, authors MAY use aria-disabled=true on a button element, rather than the disabled attribute.

>Authors MAY use the aria-disabled attribute on any element that is allowed the disabled attribute in HTML, or any element with a WAI-ARIA role which allows the aria-disabled attribute.

Essentially all roles currently defined allow the [aria-disabled attribute](https://www.w3.org/TR/wai-aria-1.1/#aria-disabled).

This would be disabled:
`<button aria-disabled="true">Click</button>
`
so would this:
`<div role="button" aria-disabled="true">Ok</div>
`
but this would not be disabled:
`<div aria-disabled="true">Enabled</button>
`

TrySound:
@mxschmitt Wasn't it reviewed and approved? Are all external contributions responded like this?
image

mxschmitt:
hey @TrySound are you planning to work on this? https://github.com/microsoft/playwright/pull/9927 landed a few weeks ago which seemed related.

If not, no worries. We'd like to close it then so that we don't have many open PRs lying around.

PR #5232, opened 2021-01-31
fix(trace viewer): Bringing back the ability to display images in Network Tab in Trace Viewer.

domderen:
This fix restores the ability to display images in the Network Tab inside Trace Viewer. Previously data returned from the server was coming back as base64 encoded string, but now it is returned as a Buffer object. This required convertion to either base64 encoded string or utf8 encoded string on the frontend.


dgozman (approved):
Oh, I broke it. Thank you for the fix!

PR #5239, opened 2021-02-01
fix(installer): retry installer when hitting ETIMEDOUT as well

### LINKED ISSUE #5233: [BUG] Failed to download Chromium on the bots
https://github.com/microsoft/playwright/pull/5229/checks?check_run_id=1803708222

Downloading browsers...
Done. Use "npm run watch" to compile.
Failed to install browsers, caused by
Error: Failed to download chromium, caused by
Error: connect ETIMEDOUT 13.107.246.13:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1107:14)
    at browserFetcher.downloadBrowserWithProgressBar.catch.e (/Users/runner/work/playwright/playwright/lib/install/installer.js:106:19)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! playwright-internal@1.9.0-next prepare: `node install-from-github.js`
npm ERR! Exit status 1

aslushnikov:
Fixes #5233

### DIFF
--- a/src/install/browserFetcher.ts
+++ b/src/install/browserFetcher.ts
@@ -137,7 +137,8 @@ export async function downloadBrowserWithProgressBar(browsersPath: string, brows
       const {error} = await downloadFile(url, zipPath, progress);
       if (!error)
         break;
-      if (attempt < N && error && typeof error === 'object' && typeof error.message === 'string' && error.message.includes('ECONNRESET')) {
+      const errorMessage = typeof error === 'object' && typeof error.message === 'string' ? error.message : '';
+      if (attempt < N && (errorMessage.includes('ECONNRESET') || errorMessage.includes('ETIMEDOUT'))) {
         // Maximum delay is 3rd retry: 1337.5ms
         const millis = (Math.random() * 200) + (250 * Math.pow(1.5, attempt));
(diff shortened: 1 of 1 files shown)


src/install/browserFetcher.ts
```diff
@@ -137,7 +137,8 @@ export async function downloadBrowserWithProgressBar(browsersPath: string, brows
       const {error} = await downloadFile(url, zipPath, progress);
       if (!error)
         break;
-      if (attempt < N && error && typeof error === 'object' && typeof error.message === 'string' && error.message.includes('ECONNRESET')) {
+      const errorMessage = typeof error === 'object' && typeof error.message === 'string' ? error.message : '';
```
mxschmitt, line 140:
```suggestion
      const errorMessage = String(error);
```
I think thats a valid use-case for the `String` constructor, or what do you think? Since it's include in the end, the prefix `Error: ` should also have no side effects.

aslushnikov: We only want to do this for well-known node errors and want to disregard everything else. So we'd rather be on a strict side

dgozman (approved):

PR #5256, opened 2021-02-02
feat(fill): allow filling elements when pointing to their wrapper

dgozman:
This addresses a pattern where fillable element is wrapped with some
layout parent, e.g. `<span class=add-margins><input></span>`.
In this case, target selector may point to the wrapper span, for example
when using layout-based selectors or codegen.

### DIFF
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -319,21 +319,42 @@ export class InjectedScript {
     });
   }
 
+  private _isFillableElement(element: Element) {
+    return element.nodeName.toLowerCase() === 'input' ||
+      element.nodeName.toLowerCase() === 'textarea' ||
+      (element as HTMLElement).isContentEditable;
+  }
+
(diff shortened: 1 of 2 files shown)


yury-s (approved):

src/server/injected/injectedScript.ts
```diff
@@ -319,21 +319,42 @@ export class InjectedScript {
...
+      if (!this._isFillableElement(element)) {
         // Go up to the label that might be connected to the input/textarea.
-        node = (node as Element).closest('label') || node;
+        element = element.closest('label') || element;
+      }
+      element = this.findLabelTarget(element);
+      if (element && !this._isFillableElement(element)) {
+        let child = element;
+        // Go down a single child route to find the input/textarea.
+        while (child.childElementCount === 1)
```
JoelEinbinder, line 344:
```suggestion
        while (child.childElementCount === 1 && !this._isFillableElement(child))
```
Otherwise this will exhibit strange behavior of contentEditable elements with one element child. For example `<div contentEditable=true><br></div>`

dgozman:
Closing this one, it is too fuzzy and solves too little.

PR #5276, opened 2021-02-03
feat(snapshot): page.snapshot prototype with tests

JoelEinbinder:
I started adding `page.snapshot` and `page.loadSnapshot` to enable some tests of our snapshotting engine.

Some thoughts:
- `loadSnapshot` should probably be on the browser instead of the page. Because the snapshots will eventually contain isMobile. Or we could put `compareToSnapshot` on the page.
- It takes a lot of refactoring to display a snapshot outside of the trace viewer. I rolled back my changes to that made resources to work so that this patch is more approachable.
- The important edge cases that are needed for pixel perfect screenshots are not easy to add into the current snapshot format. I started to add focused node into the document, but doing so requires identifying nodes inside the serialized html. It's certainly doable, but its a lot of code so I left it out of this patch. We also need selection, `:hover`, `:active`, etc.

### DIFF
--- a/src/cli/traceViewer/snapshotServer.ts
+++ b/src/cli/traceViewer/snapshotServer.ts
@@ -19,8 +19,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import type { TraceModel, trace, ContextEntry } from './traceModel';
 import { TraceServer } from './traceServer';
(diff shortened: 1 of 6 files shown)


pavelfeldman:
Closing for now

PR #5284, opened 2021-02-03
feat: auto-detect expression/function in js server

pavelfeldman:


src/server/javascript.ts
```diff
@@ -174,28 +173,25 @@ export async function evaluate(context: ExecutionContext, returnByValue: boolean
...
-      new Function('(' + functionText  + ')');
-    } catch (e2) {
-      // We tried hard to serialize, but there's a weird beast here.
-      throw new Error('Passed function is not well-serializable!');
+      new Function('(' + expression + ')');
+    } catch (e1) {
+      // This means we might have a function shorthand. Try another
+      // time prefixing 'function '.
+      if (expression.startsWith('async '))
+        expression = 'async function ' + expression.substring('async '.length);
```
yury-s, line 186: I'd move this to the js client where we now if it's a function or not and just need to update the serialization logic to replace the prefix if needed. Could be done separately.

yury-s (approved):
Looks good as long as Dmitry is cool with removing the source maps support.

src/server/webkit/protocol.ts
```diff
@@ -7428,7 +7428,7 @@ the top of the viewport and Y increases as it proceeds towards the bottom of the
       /**
        * Indicates if this type description has been type Function.
        */
-      isFunction: boolean;
+      isFunction: boolean | undefined;
```
dgozman, line 7431: How come our protocol.ts is changed without browsers.json? It should be rolled together with the browser.

dgozman (approved):

src/server/webkit/protocol.ts
pavelfeldman: oops, this is just a mass replace

src/server/javascript.ts
pavelfeldman: I need to patch 10 call sites for that :/

PR #5293, opened 2021-02-03
feat(text selector): match text in child nodes

dgozman:
This changes `text=` and `:text()` selectors to match the element when:
- it's combined text content matches the text;
- combined text content of any immediate child does not match the text.

This allows the following markup to match "Some bold and italics text":
`<div>Some <b>bold</b> and <i>italics</i> text</div>`.

For the reference, "combined text content" is almost equal to `element.textContent`,
but with some changes like using value of `<input type=button>` or ignoring `<head>`.

This also includes some caching optimizations, meaningful in complex matches
that involve multiple calls to the text engine.

Performance changes (measured on large page with ~25000 elements):
- `:has-text()` - 14% faster.
- `text=` - 50% faster.
- `:text()` - 0-35% slower.
- `:text-matches()` - 28% slower.


src/server/injected/injectedScript.ts
```diff
@@ -137,10 +140,32 @@ export class InjectedScript {
...
+        return this._evaluator.query({ scope: root as Document | Element, pierceShadow: shadow }, toCSS(selector));
+      }
+    };
+  }
+
+  private _createTextEngine(shadow: boolean): SelectorEngine {
+    const elements = (root: SelectorRoot): Element[] => {
+      const result = this._evaluator._queryCSS({ scope: root as Document | Element, pierceShadow: shadow }, '*');;
+      if (root.nodeType === Node.ELEMENT_NODE)
+        result.unshift(root as Element);
```
aslushnikov, line 155: we can inline this to line 162 to remove extra O(N) pass

src/server/injected/injectedScript.ts
```diff
@@ -115,18 +116,20 @@ export class InjectedScript {
     if (!partsToCheckOne.length)
       return candidates;
```
aslushnikov, line 117: this early return doesn't clear evaluator caches

aslushnikov (approved):

TrySound:
Is there a use case for `:has-text` if `:text` provides almost equal functionality?

dgozman:
> Is there a use case for `:has-text` if `:text` provides almost equal functionality?

Consider the following example:
```html
<my-input>
  <span>Your address</span>
  <span class=wrapper><input></span>
</my-input>
```

One way to target the input is `page.fill('my-input:has-text("Your address") input', 'my address')`.

PR #5309, opened 2021-02-04
fix(docs): ignore case when validating order of events and methods

avodovnik:
This popped up when I replaced `domcontentloaded` with `DOMContentLoaded`. It feels like ignoring casing is sensible in this case.

### DIFF
--- a/utils/doclint/documentation.js
+++ b/utils/doclint/documentation.js
@@ -229,7 +229,7 @@ Documentation.Class = class {
       const member2 = this.membersArray[i + 1];
       if (member1.kind !== 'event' || member2.kind !== 'event')
         continue;
-      if (member1.name > member2.name)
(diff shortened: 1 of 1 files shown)


yury-s (approved):
Please also update corresponding file in playwright.dev

utils/doclint/documentation.js
```diff
@@ -229,7 +229,7 @@ Documentation.Class = class {
       const member2 = this.membersArray[i + 1];
       if (member1.kind !== 'event' || member2.kind !== 'event')
         continue;
-      if (member1.name > member2.name)
+      if (member1.name.toLowerCase() > member2.name.toLowerCase())
```
pavelfeldman, line 232: nit: all js folks will tell you to use a.localeCompare(b)

avodovnik:
The playwright.dev one is https://github.com/microsoft/playwright.dev/pull/106

PR #5320, opened 2021-02-05
fix(oopifs): account for various races between processes

### LINKED ISSUE #4624: Endless frameattached events [BUG]
**Context:**
- Playwright Version: 1.6.2
- Operating System: Mac
- Node.js version: 14.15.1
- Browser: All

**Code Snippet**

```javascript

let browser;

(async () => {
  browser = await playwright.chromium.launch();

  let frameCount = 0;
  let currentFrameCount = 0;
  page.on("frameattached", () => {
    frameCount += 1;
    currentFrameCount += 1;
    console.log(`frame ${frameCount} attached`);
  });

  page.on("framedetached", () => {
    currentFrameCount -= 1;
    console.log(`frame detached. ${currentFrameCount} remain`);
  });

  await page.goto("https://www.w3schools.com/js/js_variables.asp");
})();

process.on("SIGINT", async () => {
  if (browser) await browser.close();
  process.exit();
});
```

(Replace "chromium" above with any browser. Happens the same for all.)

**Describe the bug**

Paste the example code in a file, install latest Playwright package, and run it with Node. You will see something like this:

```sh
frame 363 attached
frame 364 attached
frame 365 attached
frame 366 attached
frame 367 attached
frame 368 attached
frame 369 attached
frame 370 attached
frame 371 attached
frame 372 attached
frame 373 attached
frame 374 attached
frame 375 attached
frame 376 attached
frame 377 attached
frame 378 attached
frame 379 attached
frame 380 attached
frame 381 attached
frame 382 attached
frame 383 attached
frame 384 attached
frame 385 attached
frame 386 attached
frame 387 attached
frame 388 attached
frame 389 attached
frame 390 attached
frame 391 attached
frame 392 attached
frame 393 attached
frame 394 attached
frame 395 attached
frame 396 attached
frame 397 attached
frame 398 attached
frame 399 attached
frame 400 attached
frame 401 attached
frame 402 attached
frame 403 attached
frame 404 attached
frame 405 attached
frame 406 attached
frame 407 attached
frame 408 attached
frame 409 attached
frame 410 attached
frame 411 attached
frame 412 attached
frame detached. 314 remain
frame detached. 313 remain
frame detached. 312 remain
frame detached. 311 remain
frame detached. 310 remain
frame detached. 309 remain
frame 413 attached
frame 414 attached
frame 415 attached
frame 416 attached
frame 417 attached
frame 418 attached
frame 419 attached
frame 420 attached
frame 421 attached
frame 422 attached
frame detached. 318 remain
frame detached. 317 remain
frame detached. 316 remain
frame detached. 315 remain
frame detached. 314 remain
frame detached. 313 remain
frame detached. 312 remain
frame detached. 311 remain
frame detached. 310 remain
frame detached. 309 remain
frame detached. 308 remain
frame detached. 307 remain
frame detached. 306 remain
frame detached. 305 remain
# ... and so on forever until it crashes or CTRL+C
```

The w3schools page is just one example. It happens on every w3schools page that has ads, as well as other websites. One other example is html.com.

Through debugging, I believe this is due to iframes created by Google code, either tag manager or ads or analytics or some combination. I can try to filter out some iframes based on whether they are visible, etc. but ultimately it is pretty impossible to get any use out of the "frameattached" event when on a page such as these.

I'm considering this a bug for a few reasons:

1. With a "frameattached" listener, the page (in non-headless) is much slower and sometimes unresponsive. Sometimes it eventually crashes. Gets my CPU humming, too. Without a "frameattached" listener, things are fine. So it seems at the very least that Playwright is not doing a good job of managing frame references on the browser side or is somehow mucking things up.
2. You will also see an error in the Node console logs now and then as this runs (I think usually only in non-headless runs):

   ```
   UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'childFrames' of undefined
    at FrameManager.removeChildFramesRecursively (/Users/eric/DevProjects/qawolf/spirit/runner/node_modules/playwright/lib/server/frames.js:232:35)
    at FrameManager.frameCommittedNewDocumentNavigation (/Users/eric/DevProjects/qawolf/spirit/runner/node_modules/playwright/lib/server/frames.js:119:14)
    at FrameSession._onFrameNavigated (/Users/eric/DevProjects/qawolf/spirit/runner/node_modules/playwright/lib/server/chromium/crPage.js:452:34)
    at CRSession.<anonymous> (/Users/eric/DevProjects/qawolf/spirit/runner/node_modules/playwright/lib/server/chromium/crPage.js:288:97)
    at CRSession.emit (events.js:315:20)
    at /Users/eric/DevProjects/qawolf/spirit/runner/node_modules/playwright/lib/server/chromium/crConnection.js:151:22
    at runNextTicks (internal/process/task_queues.js:58:5)
    at processImmediate (internal/timers.js:434:9)
   (node:72938) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 10)
   ```

3. Memory issues aside, this makes "frameattached" kind of useless when you really just want "real" iframes. It might be more a feature than a bug fix, but is there some way Playwright can help here? Maybe pass more frame information like `isVisible` or `isEphemeral` (no idea how to detect that), or even just an option to ignore frames generated by ad libraries since it's a very common situation.

dgozman:
Consider the following situation (one among many possible).
- FrameA has an oopif child FrameB;
- FrameA navigates to same-process origin (e.g. about:blank);
- at the same time, FrameC is attached to the FrameB in the FrameB's process.

In this case, we get `frameNavigated` event for FrameA, immediately
followed by `frameAttached` event for FrameC. Since we detach all
FrameA's child frames on navigation, including the oopif FrameB,
there is no parent frame for FrameC to attach to.

In general, multiple processes (main and oopifs) may send their
events in wildly different order, and their view about the frame
tree may not always correspond to the "up to date" frame tree as
seen from the main frame's process. We try to keep our frame tree
aligned with what main process thinks, and ignore events that
reference frames absent in this tree.

Drive-by:  fix async processing of file chooser that may throw.

Fixes #4624.

### DIFF
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -532,14 +532,25 @@ class FrameSession {
     if (frameSession && frameId !== this._targetId) {
       // This is a remote -> local frame transition.
       frameSession._swappedIn = true;
-      const frame = this._page._frameManager.frame(frameId)!;
-      this._page._frameManager.removeChildFramesRecursively(frame);
+      const frame = this._page._frameManager.frame(frameId);
+      // Frame or even a whole subtree may be already gone, because some ancestor did navigate.
+      if (frame)
+        this._page._frameManager.removeChildFramesRecursively(frame);
+      return;
+    }
+    if (parentFrameId && !this._page._frameManager.frame(parentFrameId)) {
+      // Parent frame may be gone already because some ancestor frame navigated and
+      // destroyed the whole subtree of some oopif, while oopif's process is still sending us events.
+      // Be careful to not confuse this with "main frame navigated cross-process" scenario
+      // where parentFrameId is null.
       return;
     }
     this._page._frameManager.frameAttached(frameId, parentFrameId);
   }
 
   _onFrameNavigated(framePayload: Protocol.Page.Frame, initial: boolean) {
+    if (!this._page._frameManager.frame(framePayload.id))
+      return; // Subtree may be already gone because some ancestor navigation destroyed the oopif.
     this._page._frameManager.frameCommittedNewDocumentNavigation(framePayload.id, framePayload.url + (framePayload.urlFragment || ''), framePayload.name || '', framePayload.loaderId, initial);
     if (!initial)
       this._firstNonInitialNavigationCommittedFulfill();
@@ -609,7 +620,9 @@ class FrameSession {
     if (event.targetInfo.type === 'iframe') {
       // Frame id equals target id.
       const targetId = event.targetInfo.targetId;
-      const frame = this._page._frameManager.frame(targetId)!;
+      const frame = this._page._frameManager.frame(targetId);
+      if (!frame)
+        return; // Subtree may be already gone due to renderer/browser race.
       this._page._frameManager.removeChildFramesRecursively(frame);
       const frameSession = new FrameSession(this._crPage, session, targetId, this);
       this._crPage._sessions.set(targetId, frameSession);
@@ -715,6 +728,8 @@ class FrameSession {
   }
 
   _onDialog(event: Protocol.Page.javascriptDialogOpeningPayload) {
+    if (!this._page._frameManager.frame(this._targetId))
+      return; // Our frame/subtree may be gone already.
     this._page.emit(Page.Events.Dialog, new dialog.Dialog(
         this._page,
         event.type,
@@ -749,10 +764,18 @@ class FrameSession {
   }
 
   async _onFileChooserOpened(event: Protocol.Page.fileChooserOpenedPayload) {
-    const frame = this._page._frameManager.frame(event.frameId)!;
-    const utilityContext = await frame._utilityContext();
-    const handle = await this._adoptBackendNodeId(event.backendNodeId, utilityContext);
-    this._page._onFileChooserOpened(handle);
+    const frame = this._page._frameManager.frame(event.frameId);
+    if (!frame)
+      return;
+    let handle;
+    try {
+      const utilityContext = await frame._utilityContext();
+      handle = await this._adoptBackendNodeId(event.backendNodeId, utilityContext);
+    } catch (e) {
+      // During async processing, frame/context may go away. We should not throw.
+      return;
+    }
+    await this._page._onFileChooserOpened(handle);
   }
 
   _onDownloadWillBegin(payload: Protocol.Page.downloadWillBeginPayload) {
--- a/src/server/webkit/wkPage.ts
+++ b/src/server/webkit/wkPage.ts
@@ -559,9 +559,15 @@ export class WKPage implements PageDelegate {
   }
 
   private async _onFileChooserOpened(event: {frameId: Protocol.Network.FrameId, element: Protocol.Runtime.RemoteObject}) {
-    const context = await this._page._frameManager.frame(event.frameId)!._mainContext();
-    const handle = context.createHandle(event.element).asElement()!;
-    this._page._onFileChooserOpened(handle);
+    let handle;
+    try {
+      const context = await this._page._frameManager.frame(event.frameId)!._mainContext();
+      handle = context.createHandle(event.element).asElement()!;
+    } catch (e) {
+      // During async processing, frame/context may go away. We should not throw.
+      return;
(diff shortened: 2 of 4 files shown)


src/server/chromium/crPage.ts
```diff
@@ -694,7 +707,9 @@ class FrameSession {
       // @see https://github.com/GoogleChrome/puppeteer/issues/3865
       return;
     }
-    const context = this._contextIdToContext.get(event.executionContextId)!;
+    const context = this._contextIdToContext.get(event.executionContextId);
```
aslushnikov, line 710: this should not be necessary

src/server/chromium/crPage.ts
```diff
@@ -708,7 +723,9 @@ class FrameSession {
   }
 
   async _onBindingCalled(event: Protocol.Runtime.bindingCalledPayload) {
-    const context = this._contextIdToContext.get(event.executionContextId)!;
+    const context = this._contextIdToContext.get(event.executionContextId);
```
aslushnikov, line 726: this should not be necessary as well

src/server/chromium/crPage.ts
```diff
@@ -708,7 +723,9 @@ class FrameSession {
...
+    if (!context)
+      return;
     const pageOrError = await this._crPage.pageOrError();
     if (!(pageOrError instanceof Error))
       await this._page._onBindingCalled(event.payload, context);
```
aslushnikov, line 731: onDialog: let's check that frameManager has *us* in a frame tree, and in this case report it.

src/server/chromium/crPage.ts
```diff
@@ -749,10 +766,16 @@ class FrameSession {
...
-    const utilityContext = await frame._utilityContext();
-    const handle = await this._adoptBackendNodeId(event.backendNodeId, utilityContext);
-    this._page._onFileChooserOpened(handle);
+    const frame = this._page._frameManager.frame(event.frameId);
+    if (!frame)
+      return;
+    try {
+      const utilityContext = await frame._utilityContext();
+      const handle = await this._adoptBackendNodeId(event.backendNodeId, utilityContext);
+      this._page._onFileChooserOpened(handle);
```
aslushnikov, line 775:
```js
await this._page._onFileChooserOpened(handle);
```

src/server/webkit/wkPage.ts
```diff
@@ -559,9 +559,13 @@ export class WKPage implements PageDelegate {
   }
 
   private async _onFileChooserOpened(event: {frameId: Protocol.Network.FrameId, element: Protocol.Runtime.RemoteObject}) {
-    const context = await this._page._frameManager.frame(event.frameId)!._mainContext();
-    const handle = context.createHandle(event.element).asElement()!;
-    this._page._onFileChooserOpened(handle);
+    try {
+      const context = await this._page._frameManager.frame(event.frameId)!._mainContext();
+      const handle = context.createHandle(event.element).asElement()!;
+      this._page._onFileChooserOpened(handle);
```
aslushnikov, line 565: await is missing

src/server/chromium/crPage.ts
```diff
@@ -749,10 +766,16 @@ class FrameSession {
...
+    } catch (e) {
+      // During async processing, frame/context may go away. We should not throw.
+    }
   }
 
   _onDownloadWillBegin(payload: Protocol.Page.downloadWillBeginPayload) {
```
aslushnikov, line 781: before reporting download to users, let's make sure that `this._frameId` is present inside `this._frameManager` (similarly to dialog event)

aslushnikov (approved):

src/server/chromium/crPage.ts
dgozman: This is a browser-side event, so we should be good. Also, detaching the frame does not cancel the download, so we should report it anyway.

PR #5330, opened 2021-02-05
docs(csharp): updating broken references

avodovnik:


docs/src/api/class-chromiumbrowser.md
```diff
@@ -39,6 +39,7 @@ const { chromium } = require('playwright');
 ```
 
 ## async method: ChromiumBrowser.newBrowserCDPSession
+* langs: js, python
```
yury-s, line 42: The entire class is marked as js-only:
```
# class: ChromiumBrowser
* langs: js	* langs: js
```
this is usually sufficient to filter out all child entities too.

docs/src/api/class-browsertype.md
```diff
@@ -153,7 +153,7 @@ Firefox user preferences. Learn more about the Firefox user preferences at
 
 ### option: BrowserType.launch.firefoxUserPrefs
 * langs: csharp, java
-- `firefoxUserPrefs` <[Object]<[string], [any]>>
```
yury-s, line 156: I'd keep it `any` to highlight that it can also be a primitive type, we also use `any` in other places in the docs so you need support for that anyway. Does it cause any problems?

avodovnik: Sure, I can flip the handling around. The current logic there was, that I'd translate that to a generic `<T>`, so this caused problems. But even if I do actually define that return as `object` and add an extension method down the line that is generic, it'll be fine.

docs/src/api/class-chromiumbrowser.md
avodovnik: You're right. Not sure why I had it anymore. Removed it. Also, I changed the way I fixed the problem by removing the reference in the paragraph for java and csharp which is probably the _better_ way of handling it...

docs/src/api/class-browsercontext.md
```diff
@@ -525,11 +525,18 @@ The [origin] to grant permissions to, e.g. "https://example.com".
...
 - returns: <[Array]<[Page]>>
 
 Returns all open pages in the context. Non visible pages, such as `"background_page"`, will not be listed here. You can
 find them using [`method: ChromiumBrowserContext.backgroundPages`].
 
+## method: BrowserContext.pages
+* langs: csharp, java
+- returns: <[Array]<[Page]>>
+
+Returns all open pages in the context. Non visible pages, such as `"background_page"`, will not be listed here.
```
pavelfeldman, line 538: Just trim docs to `Returns all open pages in the context.`

pavelfeldman (approved):

PR #5343, opened 2021-02-06
devops: start publishing canary at midnight every day

aslushnikov:
This patch:
- starts publishing canary NPM package at 00:10AM UTC
- canary version is published from default (`master`) branch and is 
  named with a date. E.g. for a version published on Feb 5, 2021, the
  version would be `1.8.0-alpha-feb-5-2021`
- versions from release branches are still published on every commit and have the 
  regular commit timestamp suffix

### DIFF
--- a/utils/build/update_canary_version.js
+++ b/utils/build/update_canary_version.js
@@ -19,10 +19,20 @@ const fs = require('fs');
 const path = require('path');
 const { execSync } = require('child_process');
 
-const timestamp = execSync('git show -s --format=%ct HEAD', {
-  stdio: ['ignore', 'pipe', 'ignore']
-}).toString('utf8').trim() + '000';
 const packageJSON = require('../../package.json');
-packageJSON.version = packageJSON.version + '-' + timestamp;
+if (process.argv[2] === '--today-date') {
+  const date = new Date();
+  const month = ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep', 'oct', 'nov', 'dec'][date.getMonth()];
+  const day = date.getDate();
+  const year = date.getFullYear();
+  packageJSON.version = `${packageJSON.version}-alpha-${month}-${day}-${year}`;
+} else if (prcess.argv[2] === '--commit-timestamp') {
+  const timestamp = execSync('git show -s --format=%ct HEAD', {
+    stdio: ['ignore', 'pipe', 'ignore']
+  }).toString('utf8').trim();
+  packageJSON.version = `${packageJSON.version}-${timestamp}000`;
+} else {
+  throw new Error('This script must be run with either --timestamp or --current-date parameter');
+}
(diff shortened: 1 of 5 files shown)


pavelfeldman:
We still need driver per commit, does it only slow down publishing npms?

utils/publish_all_packages.sh
```diff
@@ -48,9 +48,6 @@ fi
 
 NPM_PUBLISH_TAG="next"
 
-if [[ $1 == "--tip-of-tree" ]]; then
```
dgozman, line 51: `git status` check above fails if you update version beforehand.

utils/build/update_canary_version.js
```diff
@@ -19,10 +19,20 @@ const fs = require('fs');
...
+  const day = date.getDate();
+  const year = date.getFullYear();
+  packageJSON.version = `${packageJSON.version}-alpha-${month}-${day}-${year}`;
+} else if (prcess.argv[2] === '--commit-timestamp') {
+  const timestamp = execSync('git show -s --format=%ct HEAD', {
+    stdio: ['ignore', 'pipe', 'ignore']
+  }).toString('utf8').trim();
+  packageJSON.version = `${packageJSON.version}-${timestamp}000`;
+} else {
+  throw new Error('This script must be run with either --timestamp or --current-date parameter');
```
dgozman, line 35: nit: wrong arguments in the message

utils/build/update_canary_version.js
```diff
@@ -19,10 +19,20 @@ const fs = require('fs');
...
-  stdio: ['ignore', 'pipe', 'ignore']
-}).toString('utf8').trim() + '000';
 const packageJSON = require('../../package.json');
-packageJSON.version = packageJSON.version + '-' + timestamp;
+if (process.argv[2] === '--today-date') {
+  const date = new Date();
+  const month = ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep', 'oct', 'nov', 'dec'][date.getMonth()];
+  const day = date.getDate();
+  const year = date.getFullYear();
+  packageJSON.version = `${packageJSON.version}-alpha-${month}-${day}-${year}`;
```
dgozman, line 28: Just curious, why do we need "alpha"?

dgozman:
> We still need driver per commit, does it only slow down publishing npms?

We can try driver-per-day, as long as we roll languages in advance 😄

utils/publish_all_packages.sh
aslushnikov: fixed

utils/build/update_canary_version.js
aslushnikov: fixed

utils/build/update_canary_version.js
aslushnikov: just to ensure that all there versions go before all others (there's alphabetical version sorting)

dgozman (approved):

Issue #5355, opened 2021-02-08
[BUG] Firefox: ASRouterNewTabHook.jsm, line 61: TypeError: this._router is null


mxschmitt:
Reported:

- https://github.com/microsoft/playwright-sharp/issues/1101
- https://github.com/microsoft/playwright-python/issues/477#issuecomment-774386406
- https://github.com/microsoft/playwright-python/issues/509 - with the normal JS variant

~Upstream issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1690231~

The issue itself causes a full defect of Firefox for some users.

aslushnikov:
We get this error on every browser launch, yet only some clients report that Firefox fails to start... it doesn't look like this error is at fault here.

aslushnikov:
I checked and the mentioned upstream patch doesn't change things for me.

We also *think* this is a dupe of https://github.com/microsoft/playwright/issues/4660, so speculatively merging there.

PR #5382, opened 2021-02-09
browser(firefox): cherry pick router fix from 86bbfadfe291

JoelEinbinder:
While waiting for https://hg.mozilla.org/integration/autoland/rev/86bbfadfe291 to land in beta, we should cherry pick it because it seems to fix a major issue.

#5355


aslushnikov:
I looked into this and actually don't think this has anything to do with the bug reported. The JS error should not crash browser.

PR #5385, opened 2021-02-09
chore: introduce instrumentation api

pavelfeldman:


src/server/progress.ts
```diff
@@ -59,21 +43,22 @@ export class ProgressController {
   private _deadline: number = 0;
   private _timeout: number = 0;
   private _logRecording: string[] = [];
-  private _listener?: ProgressListener;
-
-  constructor() {
+  readonly metadata: CallMetadata;
+  readonly instrumentation: Instrumentation;
+  readonly sdkObject: SdkObject;
```
dgozman, line 48: Why is this public? Anyone uses it?

src/trace/tracer.ts
```diff
@@ -147,24 +154,24 @@ class ContextTracer implements SnapshotterDelegate, ActionListener {
     return (page as any)[pageIdSymbol];
   }
 
-  async onActionCheckpoint(name: string, metadata: ActionMetadata): Promise<void> {
+  async onActionCheckpoint(name: string, sdkObject: SdkObject, metadata: CallMetadata): Promise<void> {
     const snapshotId = createGuid();
     snapshotsForMetadata(metadata).push({ name, snapshotId });
-    await this._snapshotter.forceSnapshot(metadata.page, snapshotId);
+    await this._snapshotter.forceSnapshot(sdkObject.attribution.page!, snapshotId);
```
dgozman, line 160: I _think_ we are coming here for all actions now, so `page` could easily be null, e.g. `context.newPage()`.

src/trace/tracer.ts
```diff
@@ -147,24 +154,24 @@ class ContextTracer implements SnapshotterDelegate, ActionListener {
...
+  async onAfterAction(result: ActionResult, sdkObject: SdkObject, metadata: CallMetadata): Promise<void> {
     const event: trace.ActionTraceEvent = {
       timestamp: monotonicTime(),
       type: 'action',
       contextId: this._contextId,
-      pageId: this.pageId(metadata.page),
-      action: metadata.type,
-      selector: typeof metadata.target === 'string' ? metadata.target : undefined,
-      value: metadata.value,
+      pageId: this.pageId(sdkObject.attribution.page!),
```
dgozman, line 168: Same here.

src/trace/tracer.ts
```diff
@@ -147,24 +154,24 @@ class ContextTracer implements SnapshotterDelegate, ActionListener {
...
+      objectType: metadata.type,
+      method: metadata.method,
+      params: metadata.params,
```
dgozman, line 171: Writing all parameters into the json could be expensive (think all evaluation scripts), and also easily non-serializable. We should follow up.

dgozman (approved):

src/server/progress.ts
pavelfeldman: why not?

PR #5414, opened 2021-02-11
docs: use custom type for exposed binding and function in java

yury-s:


pavelfeldman (approved):

yury-s:
Discussed offline and decided to define a custom type manually in java with the idea that usage examples for the method should be sufficient.

Issue #5273, opened 2021-02-03
[Question] Chromium: browserContext.close hanging after failed download


15alja:
In one of my tests, I am expecting a download and asserting on the suggested file name. Recently, the downloads executed within Playwright/Chromium are resulting in a "Bad Certificate" error. I can live with this since I am asserting on the suggested file name, not the path it is saved to, and that is available regardless of the error. The download works fine with Google Chrome outside of Playwright. 

However, whenever this "Bad Certificate" error appears, it seems that `browserContext.close` hangs indefinitely. I could just use `browser.close` which does not hang, but does save the HAR file.

Two questions:
- Any ideas on what could cause the "Bad Certificate" error? For reference, I am using Playwright 1.8.0, but I know it has worked on this version before. If anything, how can I further debug what causes this issue?
- Why does the download failure cause a hang in `browserContext.close`?

Thanks!

Update: I tried running the test with Firefox and there are no issues.
Update 2: As a workaround, I am using promise race to timeout in case it hangs.

dgozman:
> Any ideas on what could cause the "Bad Certificate" error? For reference, I am using Playwright 1.8.0, but I know it has worked on this version before. If anything, how can I further debug what causes this issue?

We need more information to help here. Could you please provide a repro script, or a page url? For you own debugging, try running with `headless: false` option and look at Developer Tools Network Panel.

> Why does the download failure cause a hang in browserContext.close?

That's a bug, should be fixed in the next version with #5424.

PR #5424, opened 2021-02-11
fix(download): do not stall BrowserContext.close waiting for downloads

dgozman:
We might not ever get the "download finished" event when closing the context:
- in Chromium, for any ongoing download;
- in all browsers, for failed downloads.

This should not prevent closing the context. Instead of waiting for the
download and then deleting it, we force delete it immediately and reject
any promises waiting for the download completion.

References #5273.

### DIFF
--- a/src/server/browserContext.ts
+++ b/src/server/browserContext.ts
@@ -252,15 +252,6 @@ export abstract class BrowserContext extends SdkObject {
 
       await this.instrumentation.onContextWillDestroy(this);
 
-      // Collect videos/downloads that we will await.
-      const promises: Promise<any>[] = [];
-      for (const download of this._downloads)
-        promises.push(download.delete());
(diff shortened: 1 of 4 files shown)


src/server/browserContext.ts
```diff
@@ -258,7 +258,7 @@ export abstract class BrowserContext extends SdkObject {
       // Collect videos/downloads that we will await.
       const promises: Promise<any>[] = [];
       for (const download of this._downloads)
-        promises.push(download.delete());
+        promises.push(download.forceDelete());
```
yury-s, line 261: I believe we should call rm only after the context has been closed as otherwise it may file because the file is still held by the browser or am I missing something?

yury-s (approved):

src/server/browserContext.ts
dgozman: Good catch!

PR #5425, opened 2021-02-11
chore(recorder): move recording output into the gui app

pavelfeldman:


test/cli/cli-codegen-javascript.spec.ts
```diff
@@ -115,20 +115,22 @@ it('should print load/save storageState', async ({ runCLI, testInfo }) => {
...
 
 (async () => {
   const browser = await chromium.launch({
     headless: false
   });
   const context = await browser.newContext({
     storageState: '${loadFileName}'
-  });
+  });`;
+  await cli.waitFor(expectedResult1);
```
dgozman, line 127: Perhaps we should accumulate results in the cli fixture, so that we can wait for a large chunk as before?

dgozman (approved):

PR #5426, opened 2021-02-12
test: add basic end-to-end driver test

dgozman:
- Introduce internal "out of process" start()/stop() mode.
- This mode is used both in regular tests and installation tests.
- Emulate basic driver installation, browser download and running.

This should prevent future regressions like https://github.com/microsoft/playwright-python/issues/494.


aslushnikov (approved):
nit: i'd consider moving `outofprocess.ts` out of src to avoid shipping it to users, but it's tiny so it's optional

PR #5432, opened 2021-02-12
docs: introduce param overload(), use it for route/unroute

yury-s:


aslushnikov (approved):

yury-s:
We decided to generate overloads automatically for all parameters that have type union type.

PR #5457, opened 2021-02-13
feat(inspector): instrument wait for event

pavelfeldman:


src/client/waiter.ts
```diff
@@ -15,15 +15,37 @@
...
 import { EventEmitter } from 'events';
-import { rewriteErrorMessage } from '../utils/stackTrace';
+import { captureStackTrace, rewriteErrorMessage } from '../utils/stackTrace';
 import { TimeoutError } from '../utils/errors';
-import { debugLogger } from '../utils/debugLogger';
+import { createGuid } from '../utils/utils';
+import { Channel, WaitForEventInfo } from '../protocol/channels';
+import { ChannelOwner } from './channelOwner';
+
+export interface ChannelWithWaitForEventCall extends Channel {
```
dgozman, line 24: Seems unused.

src/protocol/protocol.yml
```diff
@@ -14,18 +14,39 @@
 
 
 # This object can be send with any rpc call in the "metadata" field.
```
dgozman, line 16: This comment is now out of place.

src/client/waiter.ts
```diff
@@ -64,7 +87,7 @@ export class Waiter {
 
   log(s: string) {
     this._logs.push(s);
-    debugLogger.log('api', s);
+    this._channelOwner._waitForEventInfoLog(this._waitId, s);
```
dgozman, line 90: This makes me think whether we can now remove api call logging from the client and emit it from the server? We should have all the information already.

dgozman (approved):

src/client/waiter.ts
pavelfeldman: Yes, that's the goal. We need to add api call name to metadata for that.

PR #5465, opened 2021-02-15
fix(codegen): generate proper Python method calls for selectOption

fkalinski:
Currently camelCase method name are output from codegen with Python target, what obviously doesn't work.


mxschmitt (approved):
LGTM

pavelfeldman:
Thanks for the heads up and the PR. I have a large-ish PR in the works that also fixes context manager value await and setInputFiles. I'll upload it in a day or so...

pavelfeldman:
https://github.com/microsoft/playwright/pull/5466 - contains a fix now, found a handful of similar problems and added tests for them in the same patch.

fkalinski:
Your commit is much more complete! Good to have it fixed.

PR #5471, opened 2021-02-16
docs(csharp): separate script and path

avodovnik:


pavelfeldman:
Why is this happening? Can we keep C# and Java aligned?

avodovnik:
> Why is this happening? Can we keep C# and Java aligned?

Java has a `Path` object, C# does not. Paths are usually represented as `string`, which in my case mean that the generator trips up, as it would produce invalid code (`string script, script script`). 

I considered using the `FileInfo` object, or more specifically, the `FileSystemInfo` because we also need to pass directories sometimes, but that goes against the common usage pattern of C# libraries and it makes the library less usable IMHO.

The cleanest way to think about this, then, is to differentiate the arguments. Which meant diverging C# away for these three methods, instead of making ~40 others less usable. 

The other option here could be to make this an edge case in the generator itself and handle it there.

avodovnik:
After more consideration and a chat, I'll be taking care of this in the generator so as to not diverge the docs more.

PR #5485, opened 2021-02-17
feat(waitForResponse): print regex pattern when waiting for request/response

### LINKED ISSUE #5441: [Feature] Add url/method to the error message for waitForResponse method
Hi, team! I think it's a bit hard now to understand what's wrong with `waitForResponse` method, because we have no information about method or/and url in the error message.

Current error message:
```
Timeout while waiting for event "response"
Note: use DEBUG=pw:api environment variable and rerun to capture Playwright logs.
```

So it's needed to dig into the code, because we have no idea what we've waited.

Proposal, add information about response url and method:
```
Timeout while waiting for event "response", requested POST https://test.com/api/v1/login
Note: use DEBUG=pw:api environment variable and rerun to capture Playwright logs.
```

Real case: I've deleted '/' before api path because `axios` don't need it. But I used the same paths in UI tests and just concatenated them. With URL logged bug will be obvious from the message.

Thanks!

dgozman:
Fixes #5441.

### DIFF
--- a/src/client/page.ts
+++ b/src/client/page.ts
@@ -368,7 +368,7 @@ export class Page extends ChannelOwner<channels.PageChannel, channels.PageInitia
         return urlOrPredicate(request);
       };
       const trimmedUrl = trimUrl(urlOrPredicate);
-      const logLine = trimmedUrl ? `waiting for request "${trimmedUrl}"` : undefined;
+      const logLine = trimmedUrl ? `waiting for request ${trimmedUrl}` : undefined;
       return this._waitForEvent(Events.Page.Request, { predicate, timeout: options.timeout }, logLine);
     });
   }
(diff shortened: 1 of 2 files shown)


aslushnikov (approved):

src/client/page.ts
```diff
@@ -699,10 +699,15 @@ export class BindingCall extends ChannelOwner<channels.BindingCallChannel, chann
   }
 }
 
+function trimEnd(s: string): string {
+  if (s.length > 50)
+    s = s.substring(0, 50) + '\u2026';
```
mxschmitt, line 704:
```suggestion
    s = s.substring(0, 50) + '…';
```
I think using the direct symbol makes it directly clear which symbol it is.

PR #5491, opened 2021-02-18
docs(dotnet): use framework objects instead of arbitrary strings

avodovnik:
This was one of the suggestions from the Azure SDK team - to switch to a more .NET friendly object for describing the locale. While this is a fairly cosmetic change, it does improve the look & feel of the API.


kblok (approved):

pavelfeldman:
Let's not!

PR #5492, opened 2021-02-18
docs(dotnet): removing common abbreviations

avodovnik:
_Another set of suggestions from the SDK team._


pavelfeldman:
Let's not!

PR #5496, opened 2021-02-18
test(webkit): add new scrolling tests to ensure correct webkit behavior

JoelEinbinder:
scrollable2 is just there so I can navigate between two unique but same-origin pages that have scrollbars. Our server doesnt support query params.

### DIFF
--- a/test/browsercontext-device.spec.ts
+++ b/test/browsercontext-device.spec.ts
@@ -67,10 +67,6 @@ describe('device', (suite, { browserName }) => {
     const context = await contextFactory({
       ...contextOptions,
       ...device,
-      viewport: {
-        width: 400,
-        height: 400,
-      }
     });
     const page = await context.newPage();
     await page.setContent(`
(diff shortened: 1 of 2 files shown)


test/browsercontext-device.spec.ts
```diff
@@ -58,4 +59,76 @@ describe('device', (suite, { browserName }) => {
...
   });
+
+  it('should scroll twice when emulated',(test, {browserName}) => {
+    test.skip(browserName === 'webkit');
+  }, async ({server, contextFactory, playwright, contextOptions}) => {
+    const device = playwright.devices['iPhone 6'];
+    const context = await contextFactory({
+      ...contextOptions,
+      ...device,
+      viewport: {
```
dgozman, line 70: Why both device and viewport? Here and everywhere.

test/browsercontext-device.spec.ts
```diff
@@ -58,4 +59,76 @@ describe('device', (suite, { browserName }) => {
...
+          <button>hi</button>
+  `);
+    await new Promise(x => setTimeout(x, 500));
```
dgozman, line 81: Why this one?

test/browsercontext-device.spec.ts
```diff
@@ -58,4 +59,76 @@ describe('device', (suite, { browserName }) => {
...
+    const page = await context.newPage();
+    await page.goto(server.PREFIX + '/input/scrollable.html');
+    await page.evaluate(() => window.scroll(0, 100));
+
+    await page.goto(server.PREFIX + '/input/scrollable2.html');
+    expect(await page.evaluate(() => window.scrollY)).toBe(0);
+    await context.close();
+  });
+
+  it('should scroll to a precise position with mobile scale', (test, {browserName}) => {
```
dgozman, line 114: This test seems very similar to the previous one. Let's combine them?

dgozman (approved):

test/browsercontext-device.spec.ts
JoelEinbinder: Video needed even viewport demensions for when I was debugging. Removed.

test/browsercontext-device.spec.ts
JoelEinbinder: Forgot about it. removed.

test/browsercontext-device.spec.ts
JoelEinbinder: They test distinct issues which are not going to be solved with the same patch, so I'd prefer they remain separate.

PR #5507, opened 2021-02-19
fix: return non-secure cookies with HTTPS URLs

### LINKED ISSUE #5504: Unsecure cookies are not returned from `BrowserContext.cookies(...)` for HTTPS urls
Reading [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies), it sounds like browsers send cookies according to the following table

| Secure attribute value | Request URL scheme | Allow cookie to be sent |
| -- | -- | -- |
| true                     | HTTPS              | Yes |
| true                     | HTTP                | No |
| false (or unset)    | HTTPS              | Yes |
| false (or unset)    | HTTP                | Yes |

It seems that the logic for retrieving cookies in playwright trys to mirror this behaviour. It's the third row in that table I'm interested in.

If a `BrowserContext` has a cookie that has `{ secure: false }`, then this line prevents returning the cookie for urls that use `HTTPS`: 

https://github.com/microsoft/playwright/blob/970e230fcba1c62aad9ffb8ddb858e18caa478a0/src/server/network.ts#L36

If this is not intended behaviour, I'd be happy to make a PR, since the change (barring any testing) to align with the table above amounts to:

```diff
- if ((parsedURL.protocol === 'https:') !== c.secure)
+ if ((parsedURL.protocol === 'https:') || !c.secure)
```

zevisert:
Cookies have a "Secure" attribute which tells the browsers that a given cookie should only be sent via HTTPS. In it's absense "Secure" is falsy and these cookies should be sent with both HTTP and HTTPS requests. Playwright now returns only the "Non-Secure" cookies for HTTP URLs, and both "Secure" and "Non-Secure" cookies for HTTPS URLs.

Fixes #5504

### DIFF
--- a/src/server/network.ts
+++ b/src/server/network.ts
@@ -36,7 +36,7 @@ export function filterCookies(cookies: types.NetworkCookie[], urls: string[]): t
         continue;
       if (!parsedURL.pathname.startsWith(c.path))
         continue;
-      if ((parsedURL.protocol !== 'https:') && c.secure)
+      if (parsedURL.protocol !== 'https:' && c.secure)
         continue;
       return true;
     }
--- a/src/server/network.ts
+++ b/src/server/network.ts
@@ -36,7 +36,7 @@ export function filterCookies(cookies: types.NetworkCookie[], urls: string[]): t
         continue;
       if (!parsedURL.pathname.startsWith(c.path))
         continue;
-      if ((parsedURL.protocol === 'https:') !== c.secure)
+      if ((parsedURL.protocol !== 'https:') && c.secure)
         continue;
       return true;
     }
--- a/test/browsercontext-cookies.spec.ts
+++ b/test/browsercontext-cookies.spec.ts
@@ -210,3 +210,46 @@ it('should not return cookies with empty value', async ({context, page, server})
   expect(cookies.length).toBe(0);
 });
 
+it('should return secure cookies based on HTTP(S) protocol', async ({context}) => {
+  await context.addCookies([{
+    url: 'https://foo.com',
+    name: 'doggo',
+    value: 'woofs',
+    secure: true
+  }, {
+    url: 'http://foo.com',
+    name: 'catto',
+    value: 'purrs',
+    secure: false
+  }]);
+  const cookies = await context.cookies('https://foo.com');
+  expect(cookies.sort((a, b) => a.name.localeCompare(b.name))).toEqual([{
+    name: 'catto',
+    value: 'purrs',
+    domain: 'foo.com',
+    path: '/',
+    expires: -1,
+    httpOnly: false,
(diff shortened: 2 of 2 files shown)


zevisert:
I didn't end up changing the API docs for this fix; under `BrowserContext.cookies` we currently have

> If no URLs are specified, this method returns all cookies. If URLs are specified, only cookies that affect those URLs are returned.

And that seems appropriate still. I thought about adding another sentence there, along the lines of...
> If no URLs are specified, this method returns all cookies. If URLs are specified, only cookies that affect those URLs are returned. Secure cookies are only returned for HTTPS URLs, while non-secure cookies are returned for both HTTP and HTTPS URLs.

But that seems like part of "only cookies that affect those URLs". Let me know what you think!

test/browsercontext-cookies.spec.ts
```diff
@@ -210,3 +210,46 @@ it('should not return cookies with empty value', async ({context, page, server})
...
+    value: 'woofs',
+    secure: true
+  }, {
+    url: 'http://foo.com',
+    name: 'catto',
+    value: 'purrs',
+    secure: false
+  }]);
+  const cookies = await context.cookies('https://foo.com');
+  expect(cookies.sort((a, b) => a.name.localeCompare(b.name))).toEqual([{
```
mxschmitt, line 226:
```suggestion
  expect(new Set([cookies])).toEqual(new Set([{
```
If I get you right, you want to have a sort independent toEqual, then you can use a Set for that.

zevisert: Ah yeah, good idea. Two of the other tests in the file also use a `.sort`, should I change those here too? I'd make this change  locally anyway, since your suggestion doesn't have a new closing paren for the second `Set` and `cookies` is already an iterable.

zevisert: I changed them all.

zevisert:
Oh darn, that last commit should have read `chore: use sets for group equality in cookie tests`, not `in context text` 🤦. If it's a big deal, let me know and I can rebase

src/server/network.ts
```diff
@@ -36,7 +36,7 @@ export function filterCookies(cookies: types.NetworkCookie[], urls: string[]): t
         continue;
       if (!parsedURL.pathname.startsWith(c.path))
         continue;
-      if ((parsedURL.protocol === 'https:') !== c.secure)
+      if ((parsedURL.protocol !== 'https:') && c.secure)
```
aslushnikov, line 39: nit: drop parenthesis 

```suggestion
      if (parsedURL.protocol !== 'https:' && c.secure)
```

aslushnikov (approved):

PR #5518, opened 2021-02-19
feat(inspector): send api names along with metainfo (step 1)

pavelfeldman:


src/client/channelOwner.ts
```diff
@@ -84,11 +84,11 @@ export abstract class ChannelOwner<T extends channels.Channel = channels.Channel
     };
   }
 
-  async _wrapApiCall<T>(apiName: string, func: () => Promise<T>, logger?: Logger): Promise<T> {
+  async _wrapApiCall<T, K extends channels.Channel>(apiName: string, func: (channel: K) => Promise<T>, logger?: Logger): Promise<T> {
```
dgozman, line 87: ```async _wrapApiCall<R>(apiName: string, func: (channel: T) => Promise<R>, logger?: Logger): Promise<R> {```

I believe this should infer the channel type from `ChannelOwner<T>` and you would not have to write `channel: channels.FooChannel` everywhere.

src/client/channelOwner.ts
```diff
@@ -84,11 +84,11 @@ export abstract class ChannelOwner<T extends channels.Channel = channels.Channel
...
     logger = logger || this._logger;
     try {
       logApiCall(logger, `=> ${apiName} started`);
-      const result = await func();
+      const result = await func(this._channel as any);
```
dgozman, line 91: Should we make `this._channel` private so that we always wrap with some api name?

dgozman (approved):

PR #5526, opened 2021-02-19
devops: fix nits in browser compilation infrastructure

aslushnikov:
This patch:
- makes sure that failed runs actually show as failed in GHA
- attempts to use MINGW's `du` and `awk` to get upload size


yury-s (commented):
lgtm

PR #5535, opened 2021-02-20
chore: remove //browser_patches/buildbots folder

aslushnikov:
Since we migrated all our builders/testers to the GitHub self-hosted
runners, we no longer need scripts to operate our own bots.


yury-s (approved):
Would be nice to have these build scripts publicly available somewhere so that we could point at them if people ask about build steps.

aslushnikov:
Build steps are actually described in https://github.com/microsoft/playwright/blob/master/browser_patches/README.md

Also, all the scripts are publicly available in the `//browser_patches`. This patch only removes our home-grown buildbot infra since we now use Github Actions.

PR #5597, opened 2021-02-24
chore: move trace viewer into server

pavelfeldman:


src/server/trace/viewer/screenshotGenerator.ts
```diff
@@ -16,7 +16,7 @@
 
 import fs from 'fs';
 import path from 'path';
-import * as playwright from '../../..';
+import * as playwright from '../../../..';
```
dgozman, line 19: Should we get rid of trace viewer using playwright client?

dgozman (approved):

PR #5603, opened 2021-02-24
fix(text selector): make quoted selector match by text nodes

dgozman:
This change turns quoted match to be case-sensitive (as before),
but not strictly full-string for the whole element's text.

This is a fix for a case where element contains text nodes and child elements:
```html
<div>text1<span>child node</span>text2</div>
```
We now match this div by `text="text1"` and `text="text2"`.

### DIFF
--- a/docs/src/selectors.md
+++ b/docs/src/selectors.md
@@ -148,7 +148,7 @@ Text selector has a few variations:
   page.click("text=Log in")
   ```
 
-- `text="Log in"` - text body can be escaped with single or double quotes for full-string case-sensitive match. For example `text="Log"` does not match `<button>Log in</button>` but instead matches `<span>Log</span>`.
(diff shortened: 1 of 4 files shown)


pavelfeldman (approved):

niranjan-borawake:
> This change turns quoted match to be case-sensitive (as before),
> but not strictly full-string for the whole element's text.
> 
> This is a fix for a case where element contains text nodes and child elements:
> 
> ```
> <div>text1<span>child node</span>text2</div>
> ```
> 
> We now match this div by `text="text1"` and `text="text2"`.

Hello @dgozman ,

What's the recommended way to do strict full string search? 

If I have - 

`<div>Hello World..!</div> <div>Hello</div>`

What's the way to select second div?

ychaudhari:
I had element like this:
```
<h2>
	<span>Title:</span>
	&nbsp;
	<span>text</span>
</h2>
```
My selector await `expect(page).toHaveSelector('"text"');` was working before 1.8.1 but breaking in 1.9.0 and 1.9.1

PR #5610, opened 2021-02-25
docs: add instructions for executing first script

kevin-brotcke:
Make it a little clearer how to execute your first script when going
through the introduction documentation.

### DIFF
--- a/docs/src/intro-js.md
+++ b/docs/src/intro-js.md
@@ -55,6 +55,12 @@ const { webkit } = require('playwright');
 })();
 ```
 
+Save the script above in a filename of your choosing, such as `first_script.js`. Then execute the script from the command-line using Node.js.
+
+```
+npx node first_script.js
+```
+
 By default, Playwright runs the browsers in headless mode. To see the browser UI, pass the `headless: false` flag while launching the browser. You can also use `slowMo` to slow down execution. Learn more in the debugging tools [section](./debug.md).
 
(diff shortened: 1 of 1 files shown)


kevin-brotcke:
I've used other test frameworks where you execute the tests using the binary (in this case that would be `playwright first_script.js`) so it wasn't immediately clear to me you just execute a regular `node` command. Hope this helps others onboard.

docs/src/intro-js.md
```diff
@@ -55,6 +55,12 @@ const { webkit } = require('playwright');
 })();
 ```
 
+Save the script above in a filename of your choosing, such as `first_script.js`. Then execute the script from the command-line using Node.js.
+
+```
+npx node first_script.js
```
pavelfeldman, line 61: Looks like you don't need npx here, `node first_script.js` is sufficient.

pavelfeldman:
Looks like this was dropped.

docs/src/intro-js.md
kevin-brotcke: I wanted to show good practices but it's not necessary for the example. I'll remove it.

kevin-brotcke:
Fixed. Doesn't look like I have access to open this pull request so I created a new one https://github.com/microsoft/playwright/pull/6208

Also turned on GitHub notifications so I don't miss comments. :)

PR #5620, opened 2021-02-25
chore: remove ProgressController.abort

dgozman:
It is used in a few places, but we can do a Promise.race. This change will allow us to pass existing progress to multi-step operations.

### DIFF
--- a/src/server/electron/electron.ts
+++ b/src/server/electron/electron.ts
@@ -100,17 +100,11 @@ export class ElectronApplication extends SdkObject {
   }
 
   async close() {
-    const closed = this._waitForEvent(ElectronApplication.Events.Close);
+    const progressController = new ProgressController(internalCallMetadata(), this);
+    const closed = progressController.run(progress => helper.waitForEvent(progress, this, ElectronApplication.Events.Close).promise, this._timeoutSettings.timeout({}));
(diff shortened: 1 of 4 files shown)


src/server/electron/electron.ts
```diff
@@ -100,19 +100,13 @@ export class ElectronApplication extends SdkObject {
...
+    const closed = progressController.run(progress => helper.waitForEvent(progress, this, ElectronApplication.Events.Close).promise, this._timeoutSettings.timeout({}));
     await this._nodeElectronHandle!.evaluate(({ app }) => app.quit());
     this._nodeConnection.close();
     await closed;
   }
 
-  private async _waitForEvent(event: string, predicate?: Function): Promise<any> {
-    const progressController = new ProgressController(internalCallMetadata(), this);
-    if (event !== ElectronApplication.Events.Close)
-      this._browserContext._closePromise.then(error => progressController.abort(error));
```
pavelfeldman, line 112: Why do we no longer need it?

pavelfeldman (approved):

src/server/electron/electron.ts
dgozman: We only called this method with 'close' event.

PR #5640, opened 2021-02-27
docs: rename Page.console to consoleMessage in java

yury-s:
We already have `waitForConsoleMessage` and this change aligns `onConsoleMessage` name with it.


pavelfeldman:
I would leave it as is, as in Python.

pavelfeldman:
Wong button.

pavelfeldman (approved):

PR #5646, opened 2021-02-27
chore(snapshot): brush up, start adding tests

pavelfeldman:


src/server/snapshot/snapshotter.ts
```diff
@@ -88,28 +92,42 @@ export class Snapshotter {
...
-            return undefined;
-          }, { kSnapshotStreamer, snapshotId });
-        } catch (e) {
-        }
-      })
-    ]);
+  async captureSnapshot(page: Page, snapshotId: string) {
+    const snapshotFrame = async (frame: Frame) => {
+      const context = await frame._mainContext();
+      await context.evaluateInternal(({ kSnapshotStreamer, snapshotId }) => {
```
dgozman, line 110: When I was playing with snapshots on alexa websites, evaluating in some frames was problematic (some of them were in the perpetual navigation). That made a promise that awaits all frames to never resolve, and blocked the action from proceeding.

src/server/snapshot/snapshotterInjected.ts
```diff
@@ -27,6 +27,8 @@ export type SnapshotData = {
   viewport: { width: number, height: number },
   url: string,
   snapshotId: string,
+  timestamp: number,
```
dgozman, line 30: This timestamp is from the page, while all  timestamps in tracing are from the node. Let's name them differently.

Also, how do we use this one?

src/server/snapshot/snapshotter.ts
```diff
@@ -88,28 +92,42 @@ export class Snapshotter {
...
+        (window as any)[kSnapshotStreamer].captureSnapshot(snapshotId);
+        return undefined;
+      }, { kSnapshotStreamer, snapshotId });
+    };
+    await Promise.all(page.frames().map(frame => snapshotFrame(frame).catch(() => {})));
+  }
+
+  async setAutoSnapshotInterval(interval: number): Promise<void> {
```
dgozman, line 118: How do we pass this interval to any new frames attached after setAutoSnapshotInterval() call?

test/snapshotter.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { folio as baseFolio } from './fixtures';
+import { InMemorySnapshotter } from '../lib/server/snapshot/inMemorySnapshotter';
+
+type TestFixtures = {
+  snapshotter: any;
```
dgozman, line 21: `snapshotter: InMemorySnapshotter` ?

test/snapshotter.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+fixtures.snapshotter.init(async ({ context, toImpl }, runTest) => {
+  const snapshotter = new InMemorySnapshotter(toImpl(context));
+  await snapshotter.initialize();
+  await runTest(snapshotter);
+  await snapshotter.dispose();
+});
+
+const { it, beforeEach, describe, expect } = fixtures.build();
+
+describe('snapshots', (suite, { mode }) => {
```
dgozman, line 34: Great tests!

dgozman (approved):

src/server/snapshot/snapshotter.ts
```diff
@@ -86,33 +91,46 @@ export class Snapshotter {
...
-        } catch (e) {
-        }
-      })
-    ]);
+  async captureSnapshot(page: Page, snapshotId: string) {
+    const snapshotFrame = async (frame: Frame) => {
+      const context = frame._existingMainContext();
+      await context?.evaluateInternal(({ kSnapshotStreamer, snapshotId }) => {
+        (window as any)[kSnapshotStreamer].captureSnapshot(snapshotId);
+        return undefined;
```
mxschmitt, line 105: nit: do we need the `return undefined` statements? (a few in the PR).

mxschmitt (approved):

src/server/snapshot/snapshotter.ts
pavelfeldman: Not in this case, but typically we make sure that if we don't need the return value, return undefined prevents it from being serialized.

PR #5678, opened 2021-03-02
chore(dotnet): distinguish generated files by name, change root namespace

avodovnik:


utils/doclint/generateDotnetApi.js
```diff
@@ -94,7 +94,7 @@ let classNameMap;
 
   let writeFile = (name, out, folder) => {
     let content = template.replace('[CONTENT]', out.join(`${EOL}\t`));
-    fs.writeFileSync(`${path.join(folder, name)}.cs`, content);
```
yury-s, line 97: I assume they are renamed at some point later and it's not what end user will see, right?

yury-s (approved):

utils/doclint/generateDotnetApi.js
avodovnik: Yeah, this is not something that's ever visible to the end-user. It's just something that helps _us_ distinguish the generated files from everything else. :-)

PR #5679, opened 2021-03-02
feat(debug): introduce npx playwright debug

pavelfeldman:


src/cli/cli.ts
```diff
@@ -81,22 +143,6 @@ for (const {alias, name, type} of browsers) {
...
-      console.log('');
-      console.log('Examples:');
-      console.log('');
-      console.log('  $ codegen');
-      console.log('  $ codegen --target=python');
-      console.log('  $ -b webkit codegen https://example.com');
-    });
-
 program
     .command('screenshot <url> <filename>')
```
dgozman, line 147: Don't we need all other options for screenshot as well? And for pdf? And cr/ff/wk aliases?

dgozman (approved):

PR #5702, opened 2021-03-03
chore(dotnet): add csharp examples to accessibility docs

avodovnik:


docs/src/api/class-accessibility.md
```diff
@@ -91,6 +96,28 @@ function findFocusedNode(node) {
 }
 ```
 
+```csharp
+Func<AccessibilitySnapshotResult, AccessibilitySnapshotResult> findFocusedNode = root =>
```
yury-s, line 100: Can you add a call to `await Page.Accessibility.SnapshotAsync()` to make the snippet complete similar to other languages? Also I thought we were going to change type of ax snapshot to string in all languages going forward (Joel pointed out that the format may well change and we don't want to commit to it).

yury-s (approved):

PR #5707, opened 2021-03-03
chore: split client and server channels

dgozman:
- Leave common type declarations in protocol/types
- Server side channels now accept the right CallMetadata parameter
- Extracted RecorderParameters to avoid deps violation


yury-s (approved):

pavelfeldman:
Looks stale

PR #5713, opened 2021-03-04
feat(cli): launch-server command

yury-s:
Simple command to run browserType.launchServer

### DIFF
--- a/src/cli/driver.ts
+++ b/src/cli/driver.ts
@@ -18,6 +18,8 @@
 
 import fs from 'fs';
 import path from 'path';
+import * as playwright from '../..';
+import { BrowserType } from '../client/browserType';
 import { DispatcherConnection } from '../dispatchers/dispatcher';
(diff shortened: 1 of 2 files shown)


src/cli/driver.ts
```diff
@@ -53,6 +55,12 @@ export function runServer() {
   new PlaywrightDispatcher(dispatcherConnection.rootDispatcher(), playwright);
 }
 
+export async function launchBrowserServer(browserName: string) {
+  const browserType = (playwright as any)[browserName] as BrowserType;
+  const server = await browserType.launchServer();
```
pavelfeldman, line 60: This looks good, but i would fs.readFileSync(config) and pass it here. That way user can opt into accepting downloads, respecting proxy, etc.

pavelfeldman (approved):

src/cli/driver.ts
yury-s: Done.

PR #5718, opened 2021-03-04
docs(dotnet): csharp example for worker

avodovnik:


docs/src/api/class-worker.md
```diff
@@ -37,6 +37,20 @@ for worker in page.workers:
     print("    " + worker.url)
 ```
 
+```csharp
+Page.Worker += (_, worker) =>
```
yury-s, line 41: I wonder what is the first argument of the listener and if it's necessary to have in C#?

yury-s (approved):

docs/src/api/class-worker.md
avodovnik: Yeah, events always have this same syntax of `sender, argument`, but in our example, we don't need the sender, so we use [discards](https://docs.microsoft.com/en-us/dotnet/csharp/discards#:~:text=Starting%20with%20C%23%207.0%2C%20C%23,don%27t%20have%20a%20value.).

PR #5720, opened 2021-03-04
docs: add a pom.xml example for java intro

pavelfeldman:


docs/src/intro-java.md
```diff
@@ -39,6 +38,65 @@ public class Example {
...
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+  <repositories>
+    <repository>
+      <id>snapshots-repo</id>
```
yury-s, line 80: We don't need this section for release versions only for snapshots (aka @next)

docs/src/intro-java.md
```diff
@@ -39,6 +38,65 @@ public class Example {
...
+      <snapshots><enabled>true</enabled></snapshots>
+    </repository>
+  </repositories>
+</project>
+```
+
+</TabItem>
+</Tabs>
+
+With the Example.java and pom.xml above, compile and execute your new program as follows:
```
yury-s, line 92: There is a catch, the user will have to know where to put Example.java in the maven project

yury-s (approved):

PR #5730, opened 2021-03-05
docs(dotnet): mark waitForClose as async

avodovnik:
@yury-s this method looks like it's only useful when it's `async`. Does this mess with Java at all?


yury-s (approved):

yury-s:
There is required callback parameter which is expected to close the page so it looks like this in java:
```java
page.waitForClose(() -> {
  // do something
  page.evaluate("window.close()");
});
```
Java api generator symply ignores all async prefixes so it doesn't hurt to have it.

PR #5734, opened 2021-03-05
docs: document java exceptions

yury-s:


docs/src/api/class-playwrightexception.md
```diff
@@ -0,0 +1,7 @@
+# class: PlaywrightException
+* langs: java
```
pavelfeldman, line 2: Let's move all langs: java to java.md

pavelfeldman:
I vaguely remember that the naming convention for unrecoverable runtime exceptions in Java was FooError. So it looks like your change is the opposite.

yury-s:
Decided to keep it simple and use TimeoutError in java.

PR #5746, opened 2021-03-06
docs(docker): use focal by default

### LINKED ISSUE #5591: [Docker] Follow up in docs
Some findings which would improve the docs regarding Docker where I saw issues coming up:

- [ ] Use focal by default
  - [ ] Promote focal by default in the documentation
  - [ ] Publish maybe even focal on the `:latest` tag
- [ ] Promote that users will use tags when they are using the image -> Otherwise correct browsers are not installed

Any other ideas?

```
Glossar:
- Bionic = Ubuntu 18.04
- Focal = Ubuntu 20.04
```

mxschmitt:
Closes #5591

@aslushnikov not sure how we can explain it clearer that they should use e.g. the v1.9.0 tag - seems like how I did it is a bit confusing.

### DIFF
--- a/.github/workflows/publish_canary_docker.yml
+++ b/.github/workflows/publish_canary_docker.yml
@@ -33,8 +33,8 @@ jobs:
     - run: ./utils/docker/build.sh focal playwright:localbuild-focal
     - name: tag & publish
       run: |
-        ./utils/docker/tag_and_push.sh playwright:localbuild-bionic playwright.azurecr.io/public/playwright:next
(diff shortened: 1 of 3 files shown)


pavelfeldman:
@mxschmitt Could you rebaseline it?

pavelfeldman (approved):

pavelfeldman (approved):

mxschmitt:
> @mxschmitt Could you rebaseline it?

We (Andrey + me) try to land #5951 first, which contains non-breaking changes and then this one.

aslushnikov (approved):

PR #5756, opened 2021-03-08
chore: refactor trace viewer to reuse snapshot storage

pavelfeldman:


src/server/snapshot/snapshotterInjected.ts
```diff
@@ -382,7 +382,9 @@ export function frameSnapshotStreamer() {
             if (name === 'src' && (nodeName === 'IFRAME' || nodeName === 'FRAME')) {
               // TODO: handle srcdoc?
               const frameId = (element as any)[kSnapshotFrameId];
-              value = frameId || 'data:text/html,<body>Snapshot is not available</body>';
+              if (!frameId)
+                continue;
```
dgozman, line 386: Make it `value = ''` instead, to avoid random src.

dgozman (approved):

PR #5757, opened 2021-03-08
fix: drop support for webkit on macos 10.14

yury-s:
Mojave support was dropped from WebKit ToT (https://bugs.webkit.org/show_bug.cgi?id=221211) and latest webkit roll doesn't compile on macos 10.14. So we are going to remove support for 10.14 starting with v1.10


pavelfeldman:
Hang on, I thought we agreed to keep it, just not update it?

PR #5763, opened 2021-03-09
docs(dotnet): adding missing methods from dotnet port

avodovnik:
This adds some methods from the dotnet port that should be a part of the documented API.

### DIFF
--- a/docs/src/api/csharp.md
+++ b/docs/src/api/csharp.md
@@ -1,0 +1,16 @@
+## method: Request.getPayloadAsJson
+* langs: csharp
+- returns: <[JsonDocument]>
+
+Returns a [JsonDocument] representation of [`method: Request.postDataBuffer`].
+
+### option: Request.getPayloadAsJson.serializerOptions
+- `documentOptions` <[null]|[JsonDocumentOptions]>
+
+The options that control custom behaviour when parsing the JSON.
+
+## method: Response.statusCode
+* langs: csharp
+- returns: <[System.Net.HttpStatusCode]>
+
+Gets the [System.Net.HttpStatusCode] code of the response.


docs/src/api/class-request.md
```diff
@@ -221,3 +221,14 @@ print(request.timing)
 - returns: <[string]>
 
 URL of the request.
+
+## method: Request.getPayloadAsJson
```
yury-s, line 225: let's move c#-specific stuff to csharp.md similar to python and java.

docs/src/api/class-request.md
```diff
@@ -221,3 +221,14 @@ print(request.timing)
...
+* langs: csharp
+- returns: <[JsonDocument]>
+
+Returns the <see cref="JsonDocument"/> representation of <see cref="PostDataBuffer"/>.
```
yury-s, line 229: can we stick with md format in the docs and use [JsonDocument] here instead of a formatted tag? our docs generator already has support for link substitution and you probably have it in the xml doc gen too

yury-s (approved):

docs/src/api/class-request.md
avodovnik: Since `JsonDocument` is a .NET type, I think this will break the documentation stuff, because the type won't be resolved... My thinking was that maybe I can define the types separately for C# only, but that felt like an obscure way of doing this, considering this is .NET only.. I'm torn :-).

yury-s: We have language-specific link maps(see e.g. [this one](https://github.com/microsoft/playwright.dev/blob/80bf24fcfe11d2773e9802b5817cf8fbc690b49a/common/links-java.md#L15) which is java-specific), so it should not break anything. I'm not saying that this is necessarily better than what you are doing in this PR but I believe it's easier to stick to one and the same approach across all languages.

avodovnik: I did not know that! Perfect, I'll use that!

PR #5772, opened 2021-03-09
fix(route): catch abort/continue/fulfill

dgozman:
So that it can be called at any time, even after page/context/connection closure.


yury-s (approved):
What if it fails due to e.g. page navigation, is it always ok to swallow such errors?

dgozman:
This fails a reasonable test, so closing.

PR #5785, opened 2021-03-10
docs(dotnet): reference the available constants for csharp

avodovnik:
Since the `resourceType` returned is strings, C# introduces some constants for this, to make comparison easier. This change makes them discoverable for the user.

### DIFF
--- a/docs/src/api/class-request.md
+++ b/docs/src/api/class-request.md
@@ -157,6 +157,10 @@ Contains the request's resource type as it was perceived by the rendering engine
 following: `document`, `stylesheet`, `image`, `media`, `font`, `script`, `texttrack`, `xhr`, `fetch`, `eventsource`,
 `websocket`, `manifest`, `other`.
 
+:::note csharp
+The resource types are available as constants in [ResourceTypes].
+:::
+
 ## async method: Request.response
 - returns: <[null]|[Response]>
 


docs/src/api/class-request.md
```diff
@@ -157,6 +157,13 @@ Contains the request's resource type as it was perceived by the rendering engine
 following: `document`, `stylesheet`, `image`, `media`, `font`, `script`, `texttrack`, `xhr`, `fetch`, `eventsource`,
 `websocket`, `manifest`, `other`.
 
+## method: Request.resourceType
+* langs: csharp
```
yury-s, line 161: This will only override return type but not the comment of the method. To override both you need to either explicitly list other langs in the original definition or update our docs parser.

Returned value is a string everywhere else (including typescript where we could easily narrow it down to a list of constants) and nobody complained so far. I'd rather we kept C# API consistent with it.

avodovnik: Ugh, looks like I discarded one change too many before pushing this one.

avodovnik: Also, I did only intend to override the _comment_, not the actual return type. We still return `string` (to keep things consistent), but the strings are available in the C# port through the constants file. What did you do for these cases in Java to help discoverability?

yury-s: > Also, I did only intend to override the comment, not the actual return type. 

I don't think it's supported by current parser, it's either return type override or full method (if the lang filters are mutually exclusive).

> What did you do for these cases in Java to help discoverability?

Nothing for now (we don't provide such assistance in any other language), just the documentation. It was tempting to create constants or even enum in java too but we decided to keep it simple.

avodovnik: So, the override _worked_, but I thought maybe just having a small, language-specific, note would help, so here's a PR that would render this PR useless: #5810

avodovnik: Assuming https://github.com/microsoft/playwright/pull/5810 merges, I've made this change "smaller" and using that feature :-).

avodovnik: Note: when I merge #5810, I'll rebase this, and the build error here should go away

yury-s (approved):

PR #5800, opened 2021-03-11
docs(dotnet): generate arguments in a consistent order

kblok:
I would like to get the `timeout`, `force`, and `noWaitAfter` in the same order across the board.
Although we have named arguments, C# users are not that used to that code style.

This could be done at this level, it would be also moved where the args are parsed, or we could also change the order in the actual documentation.


utils/doclint/generateDotnetApi.js
```diff
@@ -461,9 +461,18 @@ function renderMethod(member, parent, output, name) {
...
-      });
+      arg.type.properties
+        .sort((a, b) => {
+          if (a.name === 'timeout' && b.name !== 'force' && b.name !== 'noWaitAfter')
+            return 1;
+          if (a.name === 'force' && b.name !== 'noWaitAfter')
+            return 1;
+          if (a.name === 'noWaitAfter')
+            return 1;
+
```
avodovnik, line 472: small nit, but the empty line is probably redundant

avodovnik (approved):

utils/doclint/generateDotnetApi.js
```diff
@@ -461,9 +461,17 @@ function renderMethod(member, parent, output, name) {
 
   let parseArg = (/** @type {Documentation.Member} */ arg) => {
     if (arg.name === "options") {
-      arg.type.properties.forEach(prop => {
-        parseArg(prop);
-      });
+      arg.type.properties
+        .sort((a, b) => {
```
pavelfeldman, line 465: I don't think your comparator is transitive. Try it on ['a', 'force', 'b', 'timeout', 'noWaitAfter']. I always use weights when I encounter a situation like this.

kblok: How about now?

```
const fixedArguments = ['timeout', 'force', 'noWaitAfter'];

const sort = (a, b) => {
    return fixedArguments.indexOf(a) - fixedArguments.indexOf(b);
};

console.log(['a', 'force', 'b', 'timeout', 'noWaitAfter'].sort(sort));
console.log(['b', 'force', 'a', 'timeout', 'noWaitAfter'].sort(sort));
console.log(['b', 'force', 'a', 'timeout', 'c', 'noWaitAfter'].sort(sort));
console.log(['a', 'force', 'c', 'noWaitAfter', 'b', 'timeout'].sort(sort));
```

kblok:
@pavelfeldman What's the reason for [this sort](https://github.com/microsoft/playwright/blob/master/utils/doclint/documentation.js#L317)?

I bet that's the root cause of this extra order we need. It also brings `omitBackground` before `path` on  `ElementHandle.screenshot`, where the most common use would be `el.ScreenshotAsync(somePath)`.

kblok:
@pavelfeldman @avodovnik I changed the approach here. I think it's better for csharp users to just simply honor the order of the arguments in the doc. In this way, we can express in the docs the order of importance of the arguments.

utils/doclint/documentation.js
```diff
@@ -314,8 +316,11 @@ Documentation.Member = class {
       this.args.set(arg.name, arg);
       arg.enclosingMethod = this;
       if (arg.name === 'options') {
+        arg.type.properties.forEach((p,i) => {
+          p.enclosingMethod = this,
+          p.order = p.order || i + 1;
```
yury-s, line 321: How about keeping properties in the original order and sorting them where needed (just have `sortedProperties()` method on `type`) would that be equivalent? If it doesn't work let's at least rename order to something more specific like `definitionOrder`.

utils/doclint/documentation.js
```diff
@@ -278,8 +278,9 @@ Documentation.Member = class {
    * @param {MarkdownNode[]=} spec
    * @param {boolean=} required
    * @param {string[]=} templates
+   * @param {number} order
    */
-  constructor(kind, langs, name, type, argsArray, spec = undefined, required = true, templates = []) {
+  constructor(kind, langs, name, type, argsArray, spec = undefined, required = true, templates = [], order = null) {
```
yury-s, line 283: Is `order` ever set to anything but null? if not, just drop it from the argument list

yury-s (approved):

utils/doclint/documentation.js
kblok: I think we don't need a `sortedProperties` on our side. But this would affect the generators on Java and Python, right?

Do you want me to remove this and create issues on both repos to check the impact?

yury-s: Java and Python generate from api.json at the moment so we just need to make sure that the order is preserved in the output of generateApiJson.js

kblok: will do!

kblok:
How about now @yury-s? I can confirm that the api.json output is the same.

utils/doclint/generateApiJson.js
```diff
@@ -92,18 +92,19 @@ function sanitize(result) {
...
-function serializeType(type) {
+function serializeType(type, sortProperties = false) {
   /** @type {any} */
   const result = { ...type };
   if (type.properties)
-    result.properties = type.properties.map(serializeProperty);
+    result.properties = (sortProperties ? type.sortedProperties() : type.properties).map(serializeProperty);
   if (type.union)
-    result.union = type.union.map(serializeType);
+    result.union = type.union.map(type => serializeType(type));
```
yury-s, line 103: does `.map(serializeType)` not work with default argument?

utils/doclint/generateApiJson.js
```diff
@@ -92,18 +92,19 @@ function sanitize(result) {
...
  * @param {Documentation.Type} type
+ * @param {boolean} sortProperties
  */
-function serializeType(type) {
+function serializeType(type, sortProperties = false) {
   /** @type {any} */
   const result = { ...type };
   if (type.properties)
-    result.properties = type.properties.map(serializeProperty);
+    result.properties = (sortProperties ? type.sortedProperties() : type.properties).map(serializeProperty);
```
yury-s, line 101: You also need to update [type generator](https://github.com/microsoft/playwright/blob/master/utils/generate_types/index.js), see the linter failures. Also let's roll these files to https://github.com/microsoft/playwright.dev/tree/master/src and make sure it produces correct results (we may need to update generator there as well)

utils/doclint/generateApiJson.js
kblok: No, because the second argument is the `index` of the map function.

utils/doclint/generateApiJson.js
kblok: I was just taking a look at it. Thanks for the hint.

utils/doclint/generateDotnetApi.js
```diff
@@ -458,18 +458,16 @@ function renderMethod(member, parent, output, name) {
    * @param {Documentation.Member} argument 
    */
   const pushArg = (innerArgType, innerArgName, argument) => {
-    let isEnum = enumTypes.has(innerArgType);
     let isNullable = nullableTypes.includes(innerArgType);
+
```
avodovnik, line 462: nit: empty line? :)

utils/doclint/documentation.js
```diff
@@ -314,8 +314,7 @@ Documentation.Member = class {
       this.args.set(arg.name, arg);
       arg.enclosingMethod = this;
       if (arg.name === 'options') {
-        arg.type.properties.sort((p1, p2) => p1.name.localeCompare(p2.name));
-        arg.type.properties.forEach(p => p.enclosingMethod = this);
+        arg.type.properties.forEach((p) => p.enclosingMethod = this );
```
avodovnik, line 317: nit: is that parenthesis needed?

utils/doclint/generateDotnetApi.js
```diff
@@ -458,18 +458,16 @@ function renderMethod(member, parent, output, name) {
...
     const requiredSuffix = argument.required ? "" : " = default";
     args.push(`${innerArgType}${requiredPrefix} ${innerArgName}${requiredSuffix}`);
   };
 
   let parseArg = (/** @type {Documentation.Member} */ arg) => {
     if (arg.name === "options") {
-      arg.type.properties.forEach(prop => {
-        parseArg(prop);
-      });
+      arg.type.properties.forEach(parseArg);
```
avodovnik, line 470: should this not be `sortedProperties()`? I might be missing something though :)

kblok: All the changes I'm making in this PR is so we have the properties in the original order in dotnet (use `properties` instead of `sortedProperies`)

utils/doclint/documentation.js
kblok: done

utils/doclint/generateDotnetApi.js
kblok: done

utils/doclint/generateApiJson.js
kblok: How about now @yury-s?

I ran `SRC_DIR=~/code/playwright npm run roll` on playwright.dev but I got no changes. Maybe you can give me a hand when this one is merged.

kblok:
@yury-s @avodovnik do you think this is ready to go?

yury-s (approved):

utils/doclint/generateApiJson.js
yury-s: Absolutely

PR #5806, opened 2021-03-11
chore: allow skipping host requirements validation

olebedev:
### Background

The `playwright` library uses patched browsers' binaries that are not working with [nix] because the host requirements might not be satisfied. 

A way to make them work with nix is to create a nix [derivation] that wraps provided binaries within their dependencies and provides them already available by a path to a folder, so it's easy to point to them by using the `PLAYWRIGHT_BROWSERS_PATH` env variable. That means nix cares about the browsers' dependencies on its own. Hence, it doesn't make sense to validate host requirements on the `playwright` library level in this case. 

However, the library [validates host's requirements](https://github.com/microsoft/playwright/blob/master/src/server/browserType.ts#L180-L181) in such cases, which leads to a failure even when the binaries are fully operational (since they wrapped with nix) but the host requirements defined in the library are not satisfied.

### In this PR

This PR adds a condition to allow skipping host requirements validation by providing boolean `PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS` env variable. 

### Use case

In `shell.nix`

```nix
{ nixpkgs }:

with nixpkgs;

let
  playwright-browsers = callPackage ./playwright-browsers/default.nix { };
in stdenv.mkDerivation {
  name = "codeceptjs-testing-env";
  # ...
  PLAYWRIGHT_BROWSERS_PATH = playwright-browsers;
  PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS = true;
}
```

and then

```
$ nix-shell --pure ./shell.nix
[nix-shell:/workdir/web]# $PLAYWRIGHT_BROWSERS_PATH/chromium-844399/chrome-linux/chrome --version
Chromium 90.0.4392.0
[nix-shell:/workdir/web]# node
Welcome to Node.js v12.16.1.
Type ".help" for more information.
> const playwright = require('playwright');
undefined
> const browser = playwright.chromium.launch();
undefined
> 
Skipping host requirements validation logic because `PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS` env variable is set.
```

[nix]: https://nixos.org/manual/nix/unstable/introduction.html
[derivation]: https://nixos.org/manual/nix/unstable/expressions/derivations.html

### DIFF
--- a/src/server/validateDependencies.ts
+++ b/src/server/validateDependencies.ts
@@ -20,6 +20,7 @@ import * as os from 'os';
 import { spawn } from 'child_process';
 import { getUbuntuVersion } from '../utils/ubuntuVersion';
 import * as registry from '../utils/registry';
+import * as utils from '../utils/utils';
 import { printDepsWindowsExecutable } from '../utils/binaryPaths';
 
 const accessAsync = util.promisify(fs.access.bind(fs));
@@ -28,6 +29,10 @@ const statAsync = util.promisify(fs.stat.bind(fs));
 const readdirAsync = util.promisify(fs.readdir.bind(fs));
 
 export async function validateHostRequirements(registry: registry.Registry, browserName: registry.BrowserName) {
+  if (utils.getAsBooleanFromENV('PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS')) {
+    process.stdout.write('Skipping host requirements validation logic because `PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS` env variable is set.\n');
+    return;
+  }
   const ubuntuVersion = await getUbuntuVersion();
   if (browserName === 'firefox' && ubuntuVersion === '16.04')
     throw new Error(`Cannot launch firefox on Ubuntu 16.04! Minimum required Ubuntu version for Firefox browser is 18.04`);


pavelfeldman (approved):

olebedev:
Hi @pavelfeldman, thank you for such a prompt merge

I am trying to find a document regarding the release cycle but no luck so far. Can you please point me in the right direction or maybe you can give me an idea when to expect this git commit to be published to the npm? 

Many thanks!

mxschmitt:
The change should be available in 15-16 hours when you install playwright via `npm install playwright@next`.

voidus:
I know it's been over a year, but this is the best lead I have, @olebedev do you still have that `playwright-browsers/default.nix` or could give me some hints on how to reconstruct that?

I guess it could just be the browsers' binaries, but I'm trying to figure out if something can be done to avoid version mismatch problems.

olebedev:
@voidus, this is an intellectual property of the company I work for so I can't publish it, but yeah, I still have it. A recommendation I can give is to understand which version of pre-built chromium you use from the playwright project, find a commit of the closest next version within the nixpkgs repo and use that nixpkgs commit to pick up dependencies, also you would want to wrap binary with appropriate PATH/LD_LIBRARY_PATH and patch ELF headers. Hope that helps.

voidus:
In case anyone comes here later, I've put up a PoC that seems to work for me: https://github.com/voidus/nix-playwright-poc

PR #5810, opened 2021-03-12
chore(docs): add support for language specific notes

avodovnik:
This PR lets us define notes in markdown, that are language specific, for example:

```markdown
Contains the request's resource type as it was perceived by the rendering engine. ResourceType will be one of the
following: `document`, `stylesheet`, `image`, `media`, `font`, `script`, `texttrack`, `xhr`, `fetch`, `eventsource`,
`websocket`, `manifest`, `other`.

:::note csharp
The resource types are available as constants in [ResourceTypes].
:::
```

which would mean we can include small information nuggets like [here](https://github.com/microsoft/playwright/pull/5785) more easily without having to override the entire comment.

### DIFF
--- a/utils/markdown.js
+++ b/utils/markdown.js
@@ -141,10 +141,12 @@ function buildTree(lines) {
     }
 
     if (content.startsWith(':::')) {
+      let noteType = content.substring(3).split(' ');
       /** @type {MarkdownNode} */
       const node = {
         type: 'note',
-        noteType: content.substring(3)
+        noteType: noteType[0],
+        codeLang: noteType[1]
       };
       line = lines[++i];
       const tokens = [];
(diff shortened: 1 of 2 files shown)


utils/markdown.js
```diff
@@ -141,10 +141,12 @@ function buildTree(lines) {
     }
 
     if (content.startsWith(':::')) {
+      let noteType = content.substring(3).split(' ');
```
yury-s, line 144: Can we keep the same syntax for language filter as in other places? i.e.

```
:::note
* lang: csharp
<...>
:::
```

avodovnik: I mimicked the syntax for code samples:

```markdown
```csharp
```
```

because I believe notes like this won't be cross-used between languages & this keeps it simpler. Do you think there's a genuine need for supporting multiple languages at the same time?

yury-s (approved):

PR #5813, opened 2021-03-12
chore: add missing await to floating promises

yury-s:
We have quite a few places where returned promises have no catch block. Some of them are valid and never expected to throw (this are now marked with `void` operator) others were missing `await` or `catch` block. Fixing some of them to give an idea of what it would look like. If we fix remaining occurrences we can enable [`no-floating-promises`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md) eslint check to prevent this problem in the future.

### DIFF
--- a/src/client/android.ts
+++ b/src/client/android.ts
@@ -46,7 +46,7 @@ export class Android extends ChannelOwner<channels.AndroidChannel, channels.Andr
 
   setDefaultTimeout(timeout: number) {
     this._timeoutSettings.setDefaultTimeout(timeout);
-    this._channel.setDefaultTimeoutNoReply({ timeout }).catch(e => {});
+    this._channel.setDefaultTimeoutNoReply({ timeout });
   }
 
   async devices(): Promise<AndroidDevice[]> {
@@ -90,7 +90,7 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
 
   setDefaultTimeout(timeout: number) {
     this._timeoutSettings.setDefaultTimeout(timeout);
-    this._channel.setDefaultTimeoutNoReply({ timeout }).catch(e => {});
+    this._channel.setDefaultTimeoutNoReply({ timeout });
   }
 
   serial(): string {
--- a/src/client/android.ts
+++ b/src/client/android.ts
@@ -46,7 +46,7 @@ export class Android extends ChannelOwner<channels.AndroidChannel, channels.Andr
 
   setDefaultTimeout(timeout: number) {
     this._timeoutSettings.setDefaultTimeout(timeout);
-    this._channel.setDefaultTimeoutNoReply({ timeout });
+    this._channel.setDefaultTimeoutNoReply({ timeout }).catch(e => {});
   }
 
   async devices(): Promise<AndroidDevice[]> {
@@ -90,7 +90,7 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel, c
 
   setDefaultTimeout(timeout: number) {
     this._timeoutSettings.setDefaultTimeout(timeout);
-    this._channel.setDefaultTimeoutNoReply({ timeout });
+    this._channel.setDefaultTimeoutNoReply({ timeout }).catch(e => {});
   }
 
   serial(): string {
--- a/src/server/chromium/crConnection.ts
+++ b/src/server/chromium/crConnection.ts
@@ -100,7 +100,7 @@ export class CRConnection extends EventEmitter {
     for (const session of this._sessions.values())
       session._onClosed(browserDisconnectedLogs);
     this._sessions.clear();
-    Promise.resolve().then(() => this.emit(ConnectionEvents.Disconnected));
+    void Promise.resolve().then(() => this.emit(ConnectionEvents.Disconnected));
   }
 
   close() {
@@ -173,8 +173,8 @@ export class CRSession extends EventEmitter {
     });
   }
 
-  _sendMayFail<T extends keyof Protocol.CommandParameters>(method: T, params?: Protocol.CommandParameters[T]): Promise<Protocol.CommandReturnValues[T] | void> {
-    return this.send(method, params).catch((error: Error) => debugLogger.log('error', error));
+  _sendNoReply<T extends keyof Protocol.CommandParameters>(method: T, params?: Protocol.CommandParameters[T]): void {
+    this.send(method, params).catch((error: Error) => debugLogger.log('error', error));
   }
 
   _onMessage(object: ProtocolResponse) {
@@ -187,7 +187,7 @@ export class CRSession extends EventEmitter {
         callback.resolve(object.result);
     } else {
       assert(!object.id);
-      Promise.resolve().then(() => {
+      void Promise.resolve().then(() => {
         if (this._eventListener)
           this._eventListener(object.method!, object.params);
         this.emit(object.method!, object.params);
@@ -211,7 +211,7 @@ export class CRSession extends EventEmitter {
       callback.reject(rewriteErrorMessage(callback.error, `Protocol error (${callback.method}): ` + errorMessage));
     this._callbacks.clear();
     this._connection = null;
-    Promise.resolve().then(() => this.emit(CRSessionEvents.Disconnected));
+    void Promise.resolve().then(() => this.emit(CRSessionEvents.Disconnected));
   }
 }
 
--- a/src/server/chromium/crConnection.ts
+++ b/src/server/chromium/crConnection.ts
@@ -100,7 +100,7 @@ export class CRConnection extends EventEmitter {
     for (const session of this._sessions.values())
       session._onClosed(browserDisconnectedLogs);
     this._sessions.clear();
-    void Promise.resolve().then(() => this.emit(ConnectionEvents.Disconnected));
+    Promise.resolve().then(() => this.emit(ConnectionEvents.Disconnected));
   }
 
   close() {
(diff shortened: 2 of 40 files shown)


src/client/android.ts
```diff
@@ -46,7 +46,7 @@ export class Android extends ChannelOwner<channels.AndroidChannel, channels.Andr
 
   setDefaultTimeout(timeout: number) {
     this._timeoutSettings.setDefaultTimeout(timeout);
-    this._channel.setDefaultTimeoutNoReply({ timeout });
+    void this._channel.setDefaultTimeoutNoReply({ timeout });
```
dgozman, line 49: Let's do a catch for all NoReply methods?

src/server/chromium/crConnection.ts
```diff
@@ -100,7 +100,7 @@ export class CRConnection extends EventEmitter {
     for (const session of this._sessions.values())
       session._onClosed(browserDisconnectedLogs);
     this._sessions.clear();
-    Promise.resolve().then(() => this.emit(ConnectionEvents.Disconnected));
+    void Promise.resolve().then(() => this.emit(ConnectionEvents.Disconnected));
```
dgozman, line 103: I don't quite like these voids for no real reason 😄

src/server/chromium/crNetworkManager.ts
```diff
@@ -152,23 +152,23 @@ export class CRNetworkManager {
...
+      void this._client._sendMayFail('Fetch.continueRequest', {
         requestId: event.requestId
       });
     }
     if (!event.networkId) {
       // Fetch without networkId means that request was not recongnized by inspector, and
       // it will never receive Network.requestWillBeSent. Most likely, this is an internal request
       // that we can safely fail.
-      this._client._sendMayFail('Fetch.failRequest', {
+      void this._client._sendMayFail('Fetch.failRequest', {
```
dgozman, line 171: Perhaps we should make `_sendMayFail` not return a promise?

src/server/chromium/crPage.ts
```diff
@@ -646,7 +646,7 @@ class FrameSession {
     session.once('Runtime.executionContextCreated', async event => {
       worker._createExecutionContext(new CRExecutionContext(session, event.context));
     });
-    Promise.all([
+    void Promise.all([
```
dgozman, line 649: Can just drop Promise.all.

src/server/chromium/crPage.ts
```diff
@@ -969,7 +969,7 @@ class FrameSession {
     });
     const frameId = nodeInfo && typeof nodeInfo.node.frameId === 'string' ?
       nodeInfo.node.frameId : null;
-    documentElement.dispose();
+    void documentElement.dispose();
```
dgozman, line 972: Same here, we can make handle.dispose() not return a Promise.

dgozman (commented):
Overall, I think this adds more noise than solves problems. Some are good catches though. I'd advise against enabling the plugin.

src/server/chromium/crNetworkManager.ts
yury-s: There are methods like `InterceptableRequest.continue` that wait on the returned promise.

src/server/chromium/crPage.ts
yury-s: Yes, but it would be even more verbose as it'd require `void` prefix for each statement.

src/server/chromium/crPage.ts
yury-s: > Overall, I think this adds more noise than solves problems. Some are good catches though. I'd advise against enabling the plugin.

Yeah, I'm not fond of this `void` prefixes all over the code. Without them there is a lot of noise in the linter output masking real issues but we can live with that for now.

src/client/android.ts
yury-s: Done.

src/server/chromium/crConnection.ts
yury-s: For these `void Promise.resolve().then(` calls we can either introduce a helper function that will do `.catch` internally or can add `.catch()` inline which is silly. Which way you like better?

src/server/chromium/crNetworkManager.ts
yury-s: Ok, updated the code.

src/server/chromium/crPage.ts
yury-s: Removed Promise.all and made replaced the methods with synchronous version.

src/server/chromium/crPage.ts
yury-s: Changed `dispose()` to not return promise.

src/client/android.ts
yury-s: Oops, only did it for the new ones, updating existing as well.

yury-s: Since these are protocol calls, added .catch blocks.

src/server/chromium/crConnection.ts
dgozman: I like neither 😄

src/server/firefox/ffExecutionContext.ts
```diff
@@ -90,7 +90,7 @@ export class FFExecutionContext implements js.ExecutionContextDelegate {
     await this._session.send('Runtime.disposeObject', {
       executionContextId: this._executionContextId,
       objectId: handle._objectId,
-    }).catch(error => {});
+    });
```
dgozman, line 93: Why this change?

src/server/input.ts
```diff
@@ -210,7 +210,7 @@ export class Mouse {
   async click(x: number, y: number, options: { delay?: number, button?: types.MouseButton, clickCount?: number } = {}) {
     const { delay = null, clickCount = 1 } = options;
     if (delay) {
-      this.move(x, y);
+      await this.move(x, y);
```
dgozman, line 213: This was not awaiting on purpose.

src/server/snapshot/inMemorySnapshotter.ts
```diff
@@ -53,11 +52,11 @@ export class InMemorySnapshotter extends BaseSnapshotStorage implements Snapshot
     await this._server.stop();
   }
 
-  async captureSnapshot(page: Page, snapshotName: string, element?: ElementHandle): Promise<SnapshotRenderer> {
+  async captureSnapshot(page: Page, snapshotName: string): Promise<SnapshotRenderer> {
     if (this._frameSnapshots.has(snapshotName))
       throw new Error('Duplicate snapshot name: ' + snapshotName);
 
-    this._snapshotter.captureSnapshot(page, snapshotName, element);
+    this._snapshotter.captureSnapshot(page, snapshotName);
```
dgozman, line 59: Bad merge?

src/server/snapshot/snapshotter.ts
```diff
@@ -183,26 +183,31 @@ export class Snapshotter {
...
-      (window as any)[kSnapshotStreamer].markIframe(frameElement, frameId);
-    }, { kSnapshotStreamer, frameElement, frameId: frame.uniqueId });
-    frameElement.dispose();
-  } catch (e) {
-  }
+function annotateFrameHierarchy(frame: Frame) {
+  _annotateFrameHierarchyAsync(frame).catch(e => {});
+}
+
+async function _annotateFrameHierarchyAsync(frame: Frame) {
```
dgozman, line 201: Reminds me of C# 😄

src/server/supplements/inspectorController.ts
```diff
@@ -45,15 +45,15 @@ export class InspectorController implements InstrumentationListener {
         case 'log':
           const originalMetadata = this._waitOperations.get(info.waitId)!;
           originalMetadata.log.push(info.message);
-          this.onCallLog('api', info.message, sdkObject, originalMetadata);
+          await this.onCallLog('api', info.message, sdkObject, originalMetadata);
```
dgozman, line 48: I think these were without await on purpose.

src/server/supplements/recorderSupplement.ts
```diff
@@ -402,18 +402,18 @@ export class RecorderSupplement {
...
-      this._snapshotter.captureSnapshot(sdkObject.attribution.page, snapshotName);
+      await this._snapshotter.captureSnapshot(sdkObject.attribution.page, snapshotName);
     }
   }
 
   async onBeforeCall(sdkObject: SdkObject, metadata: CallMetadata): Promise<void> {
     if (this._mode === 'recording')
       return;
-    this._captureSnapshot(sdkObject, metadata, 'before');
+    await this._captureSnapshot(sdkObject, metadata, 'before');
```
dgozman, line 416: It is important to no await here.

src/server/webkit/wkPage.ts
```diff
@@ -113,7 +113,7 @@ export class WKPage implements PageDelegate {
     promises.push(this.updateHttpCredentials());
     if (this._browserContext._permissions.size) {
       for (const [key, value] of this._browserContext._permissions)
-        this._grantPermissions(key, value);
+        promises.push(this._grantPermissions(key, value));
```
dgozman, line 116: Good catch!

src/server/input.ts
yury-s: Sounds sketchy

src/server/snapshot/inMemorySnapshotter.ts
yury-s: No, that parameter is unused.

src/server/supplements/inspectorController.ts
yury-s: should be .catch then?

src/server/supplements/recorderSupplement.ts
yury-s: hm, this should be commented then, maybe .catch too?

src/server/firefox/ffExecutionContext.ts
yury-s: It's caller's responsibility to catch exceptions if any and it does so now in .dispose().

yury-s:
Removed controversial parts, left only simple ones.

dgozman (approved):
Looks great!

PR #5814, opened 2021-03-13
chore: update error message when using userDataDir arg

psybuzz:
New Playwright users may unaware of the proper way to set the browser's
`userDataDir`. Currently, calling `launch({args: ['--userDataDir=...']})`
throws shows users an error message, leading some to try
`launch({userDataDir})`, which also throws and shows a different error
message, leading users to the proper invocation,
`launchPersistentContext(userDataDir, ...)`.

This change updates the first error message's text to hopefully hint users to
the proper invocation more quickly.


yury-s (approved):

psybuzz:
Thanks for merging this :)

PR #5817, opened 2021-03-13
feat: allow picking chromium channel

pavelfeldman:


src/server/chromium/findChromiumChannel.ts
```diff
@@ -0,0 +1,92 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import fs from 'fs';
+import path from 'path';
+
+function darwin(channel: string): string | undefined {
+  switch (channel) {
+    case 'chrome': return '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
```
dgozman, line 22: What about beta and dev?

dgozman (approved):

src/server/chromium/findChromiumChannel.ts
pavelfeldman: They are using the same folder...

PR #5827, opened 2021-03-15
chore: do not delete unused browsers with PLAYWRIGHT_SKIP_BROWSER_GC

### LINKED ISSUE #5797: [Feature] Option to not remove unused browsers
I have a global cache for all browsers that will run in different docker images and with different versions. Specifically, I have a /.cache directory that stores all the browsers I care about and I plan on pruning that as necessary.

Is it possible to introduce an ENV variable to opt out of the browser autocleanup?

TL;DR: ENV var to opt out of "Removing unused browser" install step

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/5797

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -285,3 +285,11 @@ Playwright downloads Chromium, Firefox and WebKit browsers by default. To instal
 $ pip install playwright
 $ playwright install firefox
 ```
+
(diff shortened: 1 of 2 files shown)


yury-s (approved):
Let's add a test for this.

PR #5829, opened 2021-03-15
fix(inspector): await inspector init to avoid races

pavelfeldman:


yury-s (approved):
All these would be caught by the check from https://github.com/microsoft/playwright/pull/5813

Issue #5833, opened 2021-03-16
[REGRESSION]: support mac 10.14 on webkit


aslushnikov:
Official webkit no longer supports Mac 10.14. However, we still want to download a latest good version of
WebKit for Mac 10.14 that supports all the good Playwright features.

Things to do:
- [x] fork our webkit fork - with the last known WebKit version that compiles on Mac 10.14
- [x] download a compatible webkit version on Mac 10.14
- [x] setup bots on Mac 10.14 to make sure things work

aslushnikov:
Mac bots are now setup on private repo. Results are available on https://flaky.aslushnikov.com

PR #5837, opened 2021-03-16
devops: start downloading webkit fork on Mac 10.14

aslushnikov:
References #5833

### DIFF
--- a/src/utils/registry.ts
+++ b/src/utils/registry.ts
@@ -31,6 +31,7 @@ type BrowserDescriptor = {
   name: BrowserName,
   revision: string,
   download: boolean,
+  browserDirectory: string,
 };
 
 const EXECUTABLE_PATHS = {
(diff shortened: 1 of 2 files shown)


src/utils/registry.ts
```diff
@@ -203,7 +203,11 @@ export class Registry {
 
   constructor(packagePath: string) {
     const browsersJSON = JSON.parse(fs.readFileSync(path.join(packagePath, 'browsers.json'), 'utf8'));
-    this._descriptors = browsersJSON['browsers'];
+    this._descriptors = browsersJSON['browsers'].map((obj: any) => ({
+      name: obj.name,
```
yury-s, line 207: This way browser path stays the same and there is no way for us to tell if webkit-1444 in the cache is legacy mac14 build or the one that works for all newer OSes (if e.g. somebody puts it in a bug report). I can also imagine scenario where people upgrade their mac from 14 to something newer and use legacy build because it has matching version in the registry. Considering all that, I'd rather we put them in separate directories in the client side cache as well. WDYT?

aslushnikov: >  I can also imagine scenario where people upgrade their mac from 14 to something newer and use legacy build because it has matching version in the registry.

@yury-s Good point!

src/utils/registry.ts
```diff
@@ -203,13 +204,24 @@ export class Registry {
 
   constructor(packagePath: string) {
     const browsersJSON = JSON.parse(fs.readFileSync(path.join(packagePath, 'browsers.json'), 'utf8'));
-    this._descriptors = browsersJSON['browsers'];
+    this._descriptors = browsersJSON['browsers'].map((obj: any) => {
+      const name = obj.name;
+      const revisionOverride = (obj.revisionOverrides || {})[hostPlatform];
+      const revision = revisionOverride || obj.revision;
+      const browserDirectory = revisionOverride ? `${name}-${hostPlatform}-special-${revision}` : `${name}-${revision}`;
```
yury-s, line 211: do we really need 'special' in the name, maybe drop it?

yury-s (approved):

src/utils/registry.ts
aslushnikov: I want to make sure it stands out. Otherwise, the name `webkit-mac10.14-1234` is not eye-catching in any way, given that I'm on mac10.14

PR #5847, opened 2021-03-16
chore(evaluate): respect signals when evaluating on handle

pavelfeldman:


yury-s (approved):

test/page-autowaiting-basic.spec.ts
```diff
@@ -120,63 +160,30 @@ it('should await navigation when assigning location twice', async ({page, server
...
-    page.waitForEvent('framenavigated').then(() => messages.push('navigated')),
-  ]);
-  expect(frame.url()).toBe(server.EMPTY_PAGE);
-  expect(messages.join('|')).toBe('route|navigated|click');
-});
-
 it('should work with noWaitAfter: true', async ({page, server}) => {
   server.setRoute('/empty.html', async () => {});
-  await page.setContent(`<a href="${server.EMPTY_PAGE}">empty.html</a>`);
+  await page.setContent(`<a id="anchor" href="${server.EMPTY_PAGE}">empty.html</a>`);
```
yury-s, line 174: Revert this and the one below?

test/page-autowaiting-basic.spec.ts
```diff
@@ -120,63 +160,30 @@ it('should await navigation when assigning location twice', async ({page, server
...
-  });
-
-  await page.setContent(`
-    <a href="${server.EMPTY_PAGE}" target=target>empty.html</a>
-    <iframe name=target></iframe>
-  `);
-  const frame = page.frame({ name: 'target' });
-  await Promise.all([
-    page.click('a').then(() => messages.push('click')),
-    page.waitForEvent('framenavigated').then(() => messages.push('navigated')),
```
yury-s, line 153: can we have a test that checks that we don't wait for the frame navigation ?

yury-s (approved):

test/page-autowaiting-basic.spec.ts
pavelfeldman: It might be tricky - we would need to simulate slow iframe navigation commit :/

PR #5856, opened 2021-03-17
docs(dotnet): move options arguments last

kblok:
Language-specific arguments are being parsed after the options argument. We need options last.

### DIFF
--- a/utils/doclint/generateDotnetApi.js
+++ b/utils/doclint/generateDotnetApi.js
@@ -521,9 +521,8 @@ function renderMethod(member, parent, output, name) {
     pushArg(argType, argName, arg);
   };
 
-  Array.from(member.args)
-    .map(i => i[1])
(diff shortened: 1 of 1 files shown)


utils/doclint/generateDotnetApi.js
```diff
@@ -521,7 +521,10 @@ function renderMethod(member, parent, output, name) {
     pushArg(argType, argName, arg);
   };
 
-  member.args.forEach(parseArg);
+  Array.from(member.args)
+    .map(i => i[1])
+    .sort((a, b) =>  b.alias ==='options' ? -1 : 0) //move options to the back to the arguments list
+    .forEach(parseArg);
```
mxschmitt, line 527:
```suggestion
  member.argsArray
    .sort((a, b) =>  b.alias === 'options' ? -1 : 0) //move options to the back to the arguments list
    .forEach(parseArg);
```
argsArray contains the members as an array already. Should also work.

yury-s (approved):

PR #5860, opened 2021-03-17
chore(dotnet): improve name generation for objects

avodovnik:


utils/doclint/generateDotnetApi.js
```diff
@@ -316,6 +326,8 @@ function generateNameDefault(member, name, t, parent) {
     if (member.kind === 'method' || member.kind === 'property') {
       // this should be easy to name... let's call it the same as the argument (eternal optimist)
       let probableName = `${parent.name}${translateMemberName(``, name, null)}`;
+      if (specialNameResolutionMap.has(probableName))
```
yury-s, line 329: Like I mentioned in the chat it'd be nice if the list had only exceptions and most of the names would be generated in accord with some common rule (like ParamName or ClassParamName etc). While `SelectOption` looks like an exception Position could probably be derived by a common rule. We'll need this mapping at least in one more place - playwright.dev generator

yury-s (approved):

utils/doclint/generateDotnetApi.js
avodovnik: Yeah, that's a valid point. 

I've gone and improved this so the names now being generated follow a _logical_ rule. The only "special case" I had to add was for names that are _too generic_, like `value`, which gets translated to `SelectOptionValue`. But still, I'm much happier with this iteration.

utils/doclint/generateDotnetApi.js
```diff
@@ -314,19 +314,37 @@ function generateNameDefault(member, name, t, parent) {
...
+      let attemptedName = names.pop();
+      let compareType = function (left, right) {
+        if (!left || !right) return false;
+        if (left.expression && right.expression)
+          return left.expression !== right.expression;
+        return JSON.stringify(right.properties) !== JSON.stringify(left.properties);
+      }
+      while (true) {
+        // crude attempt at removing plurality
+        if (attemptedName.endsWith('s'))
```
yury-s, line 333: This is brittle, will fail on names like `properties`. Perhaps whitelist known names?

utils/doclint/generateDotnetApi.js
```diff
@@ -314,19 +314,37 @@ function generateNameDefault(member, name, t, parent) {
...
-      // this should be easy to name... let's call it the same as the argument (eternal optimist)
-      let probableName = `${parent.name}${translateMemberName(``, name, null)}`;
-      let probableType = additionalTypes.get(probableName);
-      if (probableType) {
-        // compare it with what?
-        if (probableType.expression != t.expression) {
-          throw new Error(`Non-matching types with the same name. Panic.`);
+      let names = [
+        parent.name,
+        translateMemberName(``, member.name, null),
```
yury-s, line 319: I believe it should be member.alias (and parent.alias above)

utils/doclint/generateDotnetApi.js
```diff
@@ -314,19 +314,37 @@ function generateNameDefault(member, name, t, parent) {
...
+        translateMemberName(``, name, null),
+      ];
+      if (names[2] === names[1])
+        names.pop(); // get rid of duplicates, cheaply
+      let attemptedName = names.pop();
+      let compareType = function (left, right) {
```
yury-s, line 325: Can you change it into isSameType or something, otherwise it seems that the method returns true if the types are different  but I don't quite understand how `if (!left || !right) return false;` statement fits into that logic.

utils/doclint/generateDotnetApi.js
avodovnik: It is, to some extent... But also, we only have a few of these that are actually plural, so I can work around with something like 

```
if (attemptedName.endsWith('s') && !attemptedName.endsWith('ies`))
```

:-) 

When we add something that'll break this simple logic, I can always revisit this... It's the same if I have to add a map entry, or a check, right?

utils/doclint/generateDotnetApi.js
```diff
@@ -320,19 +320,37 @@ function generateNameDefault(member, name, t, parent) {
...
     if (member.kind === 'method' || member.kind === 'property') {
-      // this should be easy to name... let's call it the same as the argument (eternal optimist)
-      let probableName = `${parent.name}${translateMemberName(``, name, null)}`;
-      let probableType = additionalTypes.get(probableName);
-      if (probableType) {
-        // compare it with what?
-        if (probableType.expression != t.expression) {
-          throw new Error(`Non-matching types with the same name. Panic.`);
+      let names = [
+        parent.alias || parent.name,
```
yury-s, line 324: `alias` is always initialized with `name` by default for members so you can omit `|| parent.name` here and below

yury-s (approved):

utils/doclint/generateDotnetApi.js
avodovnik: This might be a problem with something I did around the code, but there *are* instances where this was `undefined` for me, hence using the `||`...

PR #5864, opened 2021-03-18
fix(chromium): fix race with Chromium lifecycle events

### LINKED ISSUE #5228: [BUG] launchPersistentContext hanging on "about:blank" [REOPENED]
**Context:**
- Playwright Version: 1.7.1
- Operating System: Mac
- Node.js version: v14.15.1
- Browser: Chromium

**Code Snippet**
```js

(async () => {
  const options = {
                  headless : false, 
                  userAgent: 'Mozilla/5.0 (iPhone; CPU iPhone OS 14_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/87.0.4280.77 Mobile/15E148 Safari/604.1'
                }

                const browser = await chromium.launchPersistentContext(`udd1`, options)
                var [page] = await browser.pages();
                await page.goto("https://www.google.com/");
})();
```

**Describe the bug**

Sometimes the browser hangs on "about:blank". I've tested and it seems that if I even console.log(browser), then it doesn't even log it when this occurs... The only way to get past this is to refresh the page by using CMD + R on my keyboard. I'm unsure why this is happening as it doesn't occur every time I run my code. It seems as if the browser isn't even launching as it doesn't log.

https://github.com/microsoft/playwright/issues/4981

aslushnikov:
We should subscribe to lifecycle events not later than we send the
`Page.setLifecycleEventsEnabled` event.

Currently, we subscribe to the lifecycle events only after the frame
tree is returned. However, this way we might miss some of the lifecycle
events, which results in Chromium hanging.

Fixes #5228

### DIFF
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -401,7 +401,6 @@ class FrameSession {
       const { windowId } = await this._client.send('Browser.getWindowForTarget');
       this._windowId = windowId;
     }
-    let lifecycleEventsEnabled: Promise<any>;
     if (!this._isMainFrame())
       this._addRendererListeners();
     this._addBrowserListeners();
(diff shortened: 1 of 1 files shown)


yury-s (approved):

aslushnikov:
Turns out the problem is in a different place! The code is ALL GOOD.

PR #5868, opened 2021-03-18
fix(pageOrError): throw in launchPersistentContext if context page errors

### LINKED ISSUE #5866: [Internal] pageOrError masks exceptions and makes API inconsistent
If there is an exception while initializing the page (for example remove ffmpeg and run 'should capture static page in persistent context' test):
- pageOrError is instanceof Error
- yet BrowserContext.Events.Page is emitted
- with context.pages() not having that page since it is not initialized
- exception is completely hidden

I vaguely remember that this code fixed some bug, but I think that it creates bigger problems above, so should be removed.

```
    if (pageOrError instanceof Error) {
      // Initialization error could have happened because of
      // context/browser closure. Just ignore the page.
      if (this._browserContext.isClosingOrClosed())
        return;
      this._setIsError();
    }
    this._browserContext.emit(BrowserContext.Events.Page, this);
```

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/5866

### DIFF
--- a/src/server/browserContext.ts
+++ b/src/server/browserContext.ts
@@ -195,9 +195,13 @@ export abstract class BrowserContext extends SdkObject {
     if (!this.pages().length) {
       const waitForEvent = helper.waitForEvent(progress, this, BrowserContext.Events.Page);
       progress.cleanupWhenAborted(() => waitForEvent.dispose);
-      await waitForEvent.promise;
+      const page = (await waitForEvent.promise) as Page;
+      if (page._pageIsError)
+        throw page._pageIsError;
     }
     const pages = this.pages();
+    if (pages[0]._pageIsError)
+      throw pages[0]._pageIsError;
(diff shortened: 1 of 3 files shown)


src/server/browserContext.ts
```diff
@@ -195,9 +195,13 @@ export abstract class BrowserContext extends SdkObject {
     if (!this.pages().length) {
       const waitForEvent = helper.waitForEvent(progress, this, BrowserContext.Events.Page);
       progress.cleanupWhenAborted(() => waitForEvent.dispose);
-      await waitForEvent.promise;
+      const page = (await waitForEvent.promise) as Page;
+      if (page._pageIsError)
+        throw page._pageIsError;
```
yury-s, line 200: consider creating a new error and chaining it with page init error for better stack trace

yury-s (approved):

PR #5876, opened 2021-03-18
docs: add the browsers section

pavelfeldman:


docs/src/browsers.md
```diff
@@ -0,0 +1,95 @@
...
+
+Having said that, testing policies often require regression testing to be performed against
+the current publicly available browsers. In this case, you can opt into one of the stable channels,
+`"chrome"` or `"msedge"`.
+
+**Media codecs**
+
+Another reason for testing using official binaries is to test functionality related to media codecs.
+Chromium does not have all the codecs that Google Chrome or Microsoft Edge are bundling due to
+various licensing considerations and agreements. If your site relies this kind of codecs (which is
```
mxschmitt, line 87:
```suggestion
various licensing considerations and agreements. If your site relies on this kind of codecs (which is
```

docs/src/browsers.md
```diff
@@ -0,0 +1,95 @@
...
+browser = playwright.chromium.launch(channel="chrome")
+```
+
+### When to use Google Chrome & Microsoft Edge and when not to?
+
+**Defaults**
+
+Using default Playwright configuration with the latest Chromium is a good idea most of the time.
+Since Playwright is ahead of Stable channels for the browsers, it gives piece of mind that the
+upcoming Google Chrome or Microsoft Edge releases won't not break your site. You catch breakage
```
mxschmitt, line 74:
```suggestion
upcoming Google Chrome or Microsoft Edge releases won't break your site. You catch breakage
```

docs/src/browsers.md
```diff
@@ -0,0 +1,95 @@
...
+Each version of Playwright needs specific versions of browser binaries to operate. By default Playwright downloads Chromium, WebKit and Firefox browser versions that it supports. With every release, Playwright updates the versions
+of the browsers it supports, so that the latest Playwright would support the latest browsers at any moment.
+
+<!-- TOC -->
+
+## Chromium
+
+Playwright's Chromium version matches the latest
+[Chrome Dev Channel](https://support.google.com/chrome/a/answer/9027636?hl=en) build at the time of the release.
+So when the world is on Google Chrome 89, Playwright already supports Chrome and Edge all the way to version 91
```
mxschmitt, line 15:
```suggestion
So when the world is on Google Chrome 89, Playwright already supports Google Chrome and Microsoft Edge all the way to version 91
```

mxschmitt (approved):
just nits, nice work!

It might make sense to add there also that WK and FF is custom and why. But not a lot of users ask that, so should be fine. Saw a few users occasionally who tried to use stock FF which failed.

docs/src/browsers.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+Each version of Playwright needs specific versions of browser binaries to operate. By default Playwright downloads Chromium, WebKit and Firefox browser versions that it supports. With every release, Playwright updates the versions
+of the browsers it supports, so that the latest Playwright would support the latest browsers at any moment.
+
+<!-- TOC -->
+
+## Chromium
+
+Playwright's Chromium version matches the latest
```
yury-s, line 13: This section sounds scary for someone who is not familiar with the browser channels (and differences between them) and just wants to use the latest one (aka playwright today) so I'd rephrase it to say that by default it will test against the latest dev version of chrome and put the rest of the explanation elsewhere.

docs/src/browsers.md
```diff
@@ -0,0 +1,95 @@
...
+# Can be "msedge", "chrome-beta", "msedge-beta", "msedge-dev", etc.
+browser = playwright.chromium.launch(channel="chrome")
+```
+
+### When to use Google Chrome & Microsoft Edge and when not to?
+
+**Defaults**
+
+Using default Playwright configuration with the latest Chromium is a good idea most of the time.
+Since Playwright is ahead of Stable channels for the browsers, it gives piece of mind that the
```
yury-s, line 73: piece of mind -> peace of mind

docs/src/browsers.md
pavelfeldman: rewrote it.

PR #5899, opened 2021-03-22
chore(dotnet): alias for dblclick in C#

kblok:
DblClick is better in pascal case


avodovnik (approved):
@yury-s I wonder if we should do the same thing here that we did with similar changes, make the default _correct_ case, and alias for Python and JavaScript... this seems like a change that should be in Java too, no?

avodovnik (approved):

PR #5904, opened 2021-03-22
fix(installer): retain browsers installed via Playwrigth CLI

### LINKED ISSUE #5902: [BUG] browsers are eventually removed when installed with Playwright CLI
Repro:

In the first package:

1. Install `npm install playwright-chromium@v1.9.0`
2. Install firefox additionally with `npx playwright install firefox`

In the second package:

1. Install `npm install playwright@v1.8.0`

Notice that Firefox is getting removed.

This bug is critical and manifests itself when working with downloadable channels.

aslushnikov:
Browser registry is responsible for 3 things:
1. Remove downloaded browsers if there are no packages that refer to them
2. Install default browsers needed for the current package
3. Install browsers on-demand when used through Playwright CLI

Currently, registry relies on a single "download" field in `browsers.json`
to carry both (1) and (2). However, browsers in (3) are marked as
`download: false` so that they aren't installed automatically in (2), so
auto-remove procedure in (1) removes them on subsequent installation.

One possible approach to fix this would be modifying package's `browsers.json` to
change `download: false` to `true` when browsers are installed with
Playwright CLI. This approach was explored here:
https://github.com/microsoft/playwright/commit/bc04a51800d6d6322e43b7d147fc0ec42181e084

We decided against this since we have a history of issues related to
package modifications after NPM installation. This breaks all
sorts of yarn/npm caching mechanisms.

Instead, this patch is a two-step refactor:
- remove the "download" field in `browsers.json`. Now, all registries
(including old ones from previously-released versions) will retain any
browsers that are mentioned in the `browsers.json`.
- add a new flag "installByDefault", that is **only used** for default
installation.

With this change, the registry tasks are done like this:
- (1) auto-removal: if browser has a back reference, it is retained,
otherwise it is removed from registry
- (2) default installation: use only `installByDefault` to carry default installations
- (3) CLI installation: simply installs browsers. Since we retain
everythings that's referenced in (1), browsers aren't removed.

Fixes #5902

### DIFF
--- a/browsers.json
+++ b/browsers.json
@@ -4,17 +4,17 @@
     {
       "name": "chromium",
       "revision": "857950",
-      "download": true
+      "installByDefault": true
     },
     {
       "name": "firefox",
       "revision": "1238",
-      "download": true
+      "installByDefault": true
     },
     {
       "name": "webkit",
       "revision": "1446",
-      "download": true,
+      "installByDefault": true,
       "revisionOverrides": {
         "mac10.14": "1443"
       }
@@ -22,7 +22,7 @@
     {
       "name": "ffmpeg",
       "revision": "1005",
-      "download": true
+      "installByDefault": true
     }
   ]
 }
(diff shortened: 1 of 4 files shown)


yury-s (approved):
Nice! I also like the description!

PR #5908, opened 2021-03-22
fix: installer compilation

aslushnikov:
For some reason typescript can't find electron types when using
nested tsconfig - workaround the bug.

Drive-by: surface installer compilation problems.

### DIFF
--- a/install-from-github.js
+++ b/install-from-github.js
@@ -21,8 +21,11 @@ const {execSync} = require('child_process');
 
 console.log(`Rebuilding installer...`);
 try {
-  execSync('npm run tsc-installer');
+  execSync('npm run tsc-installer', {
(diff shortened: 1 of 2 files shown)


dgozman (approved):

yury-s (approved):

install-from-github.js
```diff
@@ -21,8 +21,11 @@ const {execSync} = require('child_process');
 
 console.log(`Rebuilding installer...`);
 try {
-  execSync('npm run tsc-installer');
+  const {stdout, stderr} = execSync('npm run tsc-installer', {
```
yury-s, line 24: drop `const {...} = `  ?

aslushnikov: gracias!

Issue #5929, opened 2021-03-23
[BUG] Control+Shift+ArrowLeft fails on contenteditables with FF + Linux


trueadm:
**Context:**
- Playwright Version: 1.9.2
- Operating System: Linux
- Node.js version: 12
- Browser: Firefox

**Code Snippet**

```javascript
const {chromium, webkit, firefox} = require('playwright');

(async () => {
  const browser = await chromium.launch();
  // ...
  // locate <div contenteditable="true" id="contenteditable"></div>
  await page.focus('#contenteditable');
  await page.keyboard.type('Hello world');
  await page.keyboard.down('Control');
  await page.keyboard.down('Shift');
  await page.keyboard.press('ArrowLeft');
  await page.keyboard.up('Shift');
  await page.keyboard.up('Control');
  // Expectation is that the word "world" should be selected.
})();
```

**Describe the bug**

This was a strange issue for us. We encountered it on our CI that runs Linux and after playing around, we were able to determine that this only occurs for Firefox on Linux. Specifically, it also only seems to affect `contenteditable="true"` nodes. It works fine for `<input>` elements.

It seems that this does nothing in FF for Linux. The expected behavior is to select the word "world", as the keyboard shortcut Control + Shift + ArrowLeft is intended to do (works fine natively using Linux and contenteditables with FF).

Note: when we run this test, we are also running the same test on other browsers too. Could it be that there's a foreground issue with selection and FF?

JoelEinbinder:
Added a test that passes for me on firefox/linux. Maybe there is something mores specific happening with your contenteditable div?

trueadm:
@JoelEinbinder Interesting. I'll dig more into it today. I suspect it's a foreground problem maybe on the CI machine.

trueadm:
I'm fairly certain something is going wrong here. I've stripped back the test case so we aren't running any JS and the page is just an empty document with a contenteditable div. This is the HTML:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
  </head>
  <body>
    <div id="contenteditable" contenteditable="true"></div>
  </body>
</html>

```

This is the test case:

```js
it(`Can select and delete a word`, async () => {
  const {page} = e2e;

  await page.focus('#contenteditable');
  await page.keyboard.type('Hello world');
  await page.keyboard.down('Control');
  await page.keyboard.down('Shift');
  await page.keyboard.press('ArrowLeft');
[...]

PR #5938, opened 2021-03-24
test(keyboard): add test with contenteditable and selection

JoelEinbinder:
Add a passing test relevant to #5929


yury-s (approved):
It doesn't quite work with Firefox on Mac according to the bots.

PR #5939, opened 2021-03-24
chore(dotnet): Improve enum values

kblok:
`domcontentloaded` won't look pretty if we grab it as it is. The same with `NetworkIdle`.
I think a few more will come.

### DIFF
--- a/utils/doclint/generateDotnetApi.js
+++ b/utils/doclint/generateDotnetApi.js
@@ -39,11 +39,6 @@ const nullableTypes = ['int', 'bool', 'decimal', 'float'];
 let documentation;
 /** @type {Map<string, string>} */
 let classNameMap;
-/** @type {Map<string, string>} */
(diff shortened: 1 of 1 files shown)


utils/doclint/generateDotnetApi.js
```diff
@@ -40,6 +39,11 @@ const nullableTypes = ['int', 'bool', 'decimal', 'float'];
 let documentation;
 /** @type {Map<string, string>} */
 let classNameMap;
+/** @type {Map<string, string>} */
+let knownNames = new Map([
```
avodovnik, line 43: There are already a few "maps" around the code, can we call this `knownEnumValues` or something _meaningful_?

kblok: how about now?

avodovnik (approved):
Yep, all good, thanks!

PR #5941, opened 2021-03-24
fix(video): do not stall video in popups

dgozman:
There are two problems, exposed by existing tests:

- We do not send Page.startScreeencast before Runtime.runIfWaitingForDebugger
  because we launch video recorder in between. This stalls when the page is busy
  immediately after resuming, e.g. with alert().
  Fixed by starting video recorder in advance.

- We wait for the first frame that may not come - for example, context-wide interception
  is blocking essential resource and first frame.
  Fixed by only waiting for the first frame before reporting the video, not the page.

### DIFF
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -401,6 +401,22 @@ class FrameSession {
       const { windowId } = await this._client.send('Browser.getWindowForTarget');
       this._windowId = windowId;
     }
+
+    let screencastOptions: types.PageScreencastOptions | undefined;
+    if (this._isMainFrame() && this._crPage._browserContext._options.recordVideo && hasUIWindow) {
+      const size = this._crPage._browserContext._options.recordVideo.size || this._crPage._browserContext._options.viewport || { width: 1280, height: 720 };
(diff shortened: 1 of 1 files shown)


src/server/chromium/crPage.ts
```diff
@@ -401,6 +401,22 @@ class FrameSession {
       const { windowId } = await this._client.send('Browser.getWindowForTarget');
       this._windowId = windowId;
     }
+
+    let screencastOptions: types.PageScreencastOptions | undefined;
+    if (this._isMainFrame() && this._crPage._browserContext._options.recordVideo && hasUIWindow) {
+      const size = this._crPage._browserContext._options.recordVideo.size || this._crPage._browserContext._options.viewport || { width: 1280, height: 720 };
```
yury-s, line 407: recordVideo.size should already be validated in `validateBrowserContextOptions`

yury-s (approved):

src/server/chromium/crPage.ts
dgozman: I'll address this in a follow up.

PR #5946, opened 2021-03-25
feat: support Moto G4 device in emulated devices for performance testing

fili:
Adding Moto G4 device to emulated devices for performance testing, based on:

https://bugs.chromium.org/p/chromium/issues/detail?id=924693
https://source.chromium.org/chromium/chromium/src/+/master:out/Debug/gen/devtools/emulation/EmulatedDevices.js;l=1226

I changed the UA to the one currently used in Lighthouse 7.1.0 (see image below).

Lighthouse-Report-Viewer


dgozman:
@fili We missed this PR for some reason, and are very sorry about that! If you'd still like to contribute this, I'd suggest to run `npm run build` locally to regenerate `types.d.ts` instead of modifying it  by hand. This way the bots that check the generated version will pass and we can land this PR.

fili:
> @fili We missed this PR for some reason, and are very sorry about that! If you'd still like to contribute this, I'd suggest to run `npm run build` locally to regenerate `types.d.ts` instead of modifying it by hand. This way the bots that check the generated version will pass and we can land this PR.

@dgozman Thanks, I actually did run that and the ```npm test``` command.

dgozman (approved):
@fili Looks great!

fili:
@dgozman Thanks, also for approving ;)

PR #5951, opened 2021-03-25
devops: use ubuntu focal on bots and docs

mxschmitt:
- drive-by: updated NPM Action
- replaced ubuntu-latest by ubuntu-20 so its pinned

Extracted from #5746

### DIFF
--- a/.github/workflows/infra.yml
+++ b/.github/workflows/infra.yml
@@ -13,10 +13,10 @@ on:
 jobs:
   doc-and-lint:
     name: "docs & lint"
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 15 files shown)


yury-s (approved):

mxschmitt:
Firefox on ubuntu 20 does not work with WebGL, not sure if it's flaky or failing. Will check later.

mxschmitt:
@aslushnikov
I tested if WebGL in Firefox (the failing tests) would work on Ubuntu 20 headed on the last 10 PW versions, on all the versions it's not working. Should I simply adjust the `test.fixme` so its `test.fixme(browserName === 'firefox');` instead of `test.fixme(browserName === 'firefox' && !headful);`?

Once this is clarified, we can land it.

mxschmitt:
Update: ready to land

aslushnikov (approved):

PR #5957, opened 2021-03-26
browser(webkit): refresh embedder UI on macOS

pavelfeldman:


yury-s (approved):
Why not inherit from WK2BrowserWindowController and avoid all changes like this in the future?

PR #5960, opened 2021-03-26
chore(dotnet): treat file as a reserved word

kblok:
If we generate a File class it will create a collision with System.IO.File and customers will hate us. The end result of this change is `SiteInputFilesFile` which it's not the nicest thing but it's friendlier.

### DIFF
--- a/utils/doclint/generateDotnetApi.js
+++ b/utils/doclint/generateDotnetApi.js
@@ -40,6 +40,13 @@ let documentation;
 /** @type {Map<string, string>} */
 let classNameMap;
 
+/** @type {Map<string, string>} */
+const customTypeNames = new Map([
+  ['domcontentloaded', 'DOMContentLoaded'],
+  ['networkidle', 'NetworkIdle'],
+  ['File', 'FilePayload'],
+]);
+
 {
   const typesDir = process.argv[2] || '../generate_types/csharp/';
   let checkAndMakeDir = (path) => {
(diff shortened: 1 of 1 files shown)


utils/doclint/generateDotnetApi.js
```diff
@@ -343,7 +343,7 @@ function generateNameDefault(member, name, t, parent) {
           attemptedName = attemptedName.substring(0, attemptedName.length - 1);
         let probableType = additionalTypes.get(attemptedName);
         if ((probableType && typesDiffer(t, probableType))
-          || (["Value"].includes(attemptedName))) {
+          || (["Value", "File"].includes(attemptedName))) {
```
yury-s, line 346: In java we ended up naming this type `FilePayload` (special cased in the java generator too), perhaps we could use the same name in C#

yury-s (approved):

utils/doclint/generateDotnetApi.js
kblok: I have that name actually. I bet that's the name in Typescript.
@avodovnik can we declare that in the md file or we need to add some code here?

kblok: Do you buy a Map like this @avodovnik https://github.com/microsoft/playwright-java/blob/7ca9b33f46422a7333bae22d74eacd5a0d209967/tools/api-generator/src/main/java/com/microsoft/playwright/tools/ApiGenerator.java#L256 ?

kblok:
@avodovnik @yury-s I refactored this a little but. I unified the map to the custom names for enum and custom names for classes into a map closer to the Java generator. I also change the name so we use the same as in Java.

avodovnik (approved):

PR #5967, opened 2021-03-26
feat(size): emulate window.screen size

pavelfeldman:


yury-s (approved):

yury-s (approved):

JoelEinbinder:
To be strictly correct we need an offset for the top/left of the viewport in the screen, as well as [safe area insets](https://developer.mozilla.org/en-US/docs/Web/CSS/env()).

PR #5970, opened 2021-03-26
fix(page): add name property to pageerror event

mxschmitt:
Before that change the `name` property of an Error was not correctly passed over to the `pageerror` event and it was always `Error`.

```js
const error = new Error('foo bar message');
error.name = 'FooBarError';
```

this PR fixes it.

### DIFF
--- a/src/protocol/serializers.ts
+++ b/src/protocol/serializers.ts
@@ -36,6 +36,7 @@ export function parseError(error: SerializedError): Error {
   }
   const e = new Error(error.error.message);
   e.stack = error.error.stack || '';
+  e.name = error.error.name;
   return e;
 }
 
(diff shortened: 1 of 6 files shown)


dgozman (commented):
Looks good, but you'll need to rebase because test harness has changed :(

mxschmitt:
@dgozman rebased, thanks for the review!

dgozman (approved):

yury-s (approved):
Not directly related to this PR but can we have a string representation of the error that could be used in other languages?

mxschmitt:
> Not directly related to this PR but can we have a string representation of the error that could be used in other languages?

What do you mean with string representation? It's already sent over the protocol so the languages should be able to adopt it. Or do you mean adding another class?

yury-s:
I mean that we give to the clients as a string anyway and each language has to assemble the string from all these name, message, stack etc fields. This only makes sense for js, while for all other languages could be formatted on the server end.

mxschmitt:
I think having this split logic in each language in place adds more complexity in the end, if a language wants a different format, they can use the parts instead to assemble it together instead of vice versa. Seems also a rare-case IMO.

yury-s:
My point is that we don't need to split in the other languages, we can simply return the string as is and have it consistent across bindings.

PR #5973, opened 2021-03-27
chore(dotnet): add Get prefixes in Page and Frame

kblok:
The generator is able to add the `Get` prefix in `IElementHandle` because those functions have no arguments there.
But it doesn't add the `Get` in Page and Frame. In order to be consistent, I'm adding these custom aliases.

### DIFF
--- a/docs/src/api/class-frame.md
+++ b/docs/src/api/class-frame.md
@@ -744,6 +744,8 @@ When all steps combined have not finished during the specified [`option: timeout
 ### option: Frame.hover.timeout = %%-input-timeout-%%
 
 ## async method: Frame.innerHTML
+* langs:
+  - alias-csharp: GetInnerHTMLAsync
 - returns: <[string]>
 
 Returns `element.innerHTML`.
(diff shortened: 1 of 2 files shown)


pavelfeldman (commented):
But these also don't have arguments. Also, we violated get prefixes in Python and in Java where they are encouraged, how evil is it to drop them from C#?

docs/src/api/class-frame.md
```diff
@@ -744,6 +744,8 @@ When all steps combined have not finished during the specified [`option: timeout
 ### option: Frame.hover.timeout = %%-input-timeout-%%
 
 ## async method: Frame.innerHTML
+* langs:
+  - alias-csharp: GetInnerHTMLAsync
```
avodovnik, line 748: I don't like this. It means we need to add this across the entire docs and make sure to keep consistent in the future. If we can already say _it has to be this way_, we can force the `Get` prefix for async methods without parameters in the generator.

avodovnik (commented):
I would prefer to do this in the generator, but I wouldn't _drop_ the prefix - to avoid confusing with properties.

kblok:
closing in favor of  #6074

avodovnik:
@kblok I think, based on our conversation, we can close this PR, right?

Issue #5977, opened 2021-03-28
[BUG] Firefox crash with SVG favicon on Win10


faulpeltz:
**Context:**
- Playwright Version: v1.10, v1.9.x
- Operating System: Windows 10 64bit
- Node.js version: 14.16 (not relevant for this bug)
- Browser: Firefox
- Extra: Works on Ubuntu 20.04

**Describe the bug**

The bundled firefox version (firefox-1238, and earlier) crashes when using a SVG favicon:

```html
<!doctype html>
<html>
  <head>
    <meta charset="utf-8">
    <link rel="icon" type="image/svg+xml" href="/favicon.svg">
    <title>SVG ICON TEST</title>
  </head>
<body>
  NO TEXT
</body>
</html>
```

```xml
<svg width="32" height="32" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg">
  <style>
  @media (prefers-color-scheme: dark) {
    .dummy { fill: #f22 }
  }
  </style>
  <circle fill="red" cx="16" cy="16" r="12" />
</svg>
```

**Windows event log**
```
Faulting application name: firefox.exe, version: 87.0.0.7746, time stamp: 0x6051c457
Faulting module name: xul.dll, version: 87.0.0.7746, time stamp: 0x6051c464
Exception code: 0xc0000005
Fault offset: 0x00000000034e8512
```

**Notes**
- This happens when running the bundled FF instance manually or from PW
- Does not happen with various "stock" FF nightly/beta/release versions
- Reproducible on multiple Win10 machines with different hardware
- Safe mode does not help
- Headless also does not work
- When running manually the cached svg favicon causes a crash when opening history

**edit**: only crashes with the "prefers-color-scheme" media query, this might be uncommon in favicons - but it *does* cause a hard crash

rwoll:
I've added a test case here and that as a repros: https://github.com/microsoft/playwright/pull/5979. macOS Playwright FF crashes in both headful and headless mode when you run the test while the stock FF does not crash.

rwoll:
I don't yet have a Playwright dev FF setup, but I want to see if we're executing https://github.com/microsoft/playwright/blob/0943af28069774b09dd2481059568072c88cd3dd/browser_patches/firefox/patches/bootstrap.diff#L641 when the SVG loads, and if there's a null pointer or some other crash there.

faulpeltz:
[...]

PR #5979, opened 2021-03-29
fix: favicon with color pref crashes firefox (#5977)

rwoll:


rwoll:
Add a WIP test. Letting it run on all the bots to get a better cross-platform baseline. I'll remove the `only` directive, and add `fails` as necessary, along with fixing some typos in the comments once the bots finish. 🤖

browser_patches/firefox/patches/bootstrap.diff
```diff
@@ -638,7 +638,7 @@ index 48af2c30f3604a94e6bc1da7848aed3c8ac11c1e..f9c1df0b36b3c9342b81b4c2c7bc7aec
      IgnoreRFP aIgnoreRFP) const {
 +  auto* docShell = static_cast<nsDocShell*>(GetDocShell());
 +  nsIDocShell::ColorSchemeOverride colorScheme;
-+  if (docShell->GetColorSchemeOverride(&colorScheme) == NS_OK &&
++  if (docShell && docShell->GetColorSchemeOverride(&colorScheme) == NS_OK &&
```
rwoll, line 641: NB: This change avoids crashing Firefox when there is a
`prefers-color-scheme` query in an SVG favicon, and assumes
it's OK if Playwright's `colorScheme` emulation does not
apply to the favicon context. I think the favicon is outside the scope of Playwright and most testing; however, if not, is there a way to invoke `GetColorSchemeOverride` on some other object that would give us what we want in this case?

test/favicon.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import path from 'path';
+import { it } from './fixtures';
+
+it('should load svg favicon with prefer-color-scheme', (test, {browserName, browserChannel, headful}) => {
+  test.fixme(browserName === 'firefox', 'firefox hard crashes in both headless and headful');
```
rwoll, line 22: Once the included patch lands we can remove this.

pavelfeldman (approved):

PR #5983, opened 2021-03-29
chore(dotnet): improve return type classes

kblok:
Use the same code we have to infer the types in arguments but in return types
The impact of this is : ElementHandleBoundingBoxResult is not Bounding box and PageViewportSizeResult is ViewportSize

### DIFF
--- a/utils/doclint/generateDotnetApi.js
+++ b/utils/doclint/generateDotnetApi.js
@@ -394,11 +394,7 @@ function generateEnumNameIfApplicable(member, name, type, parent) {
  * @param {Function} output
  */
 function renderMethod(member, parent, output, name) {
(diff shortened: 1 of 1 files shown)


avodovnik:
I remember doing this for a specific reason. The returned objects are technically not the same and there was conversation about doing it this way. What's the benefit here?

kblok:
@avodovnik the reason is to make #5985 we are inferring the types from arguments and from return types in different ways.
We need the ViewportSize we pass to the NewContextAsync, to be the same as the one the Page would return.

I didn't notice any side-effect here.
I also recalled that you created those `pop` tricks for the arguments after solving the return types. I don't see a reason to infer arguments and return types in a different way.

PR #5984, opened 2021-03-29
chore(dotnet): improve self-referenced events

kblok:
Not all the events referencing the parent type are `EventHandler`.  The `Popup` even references the same type but that `IPage` is not `this` but the popup page.


utils/doclint/generateDotnetApi.js
```diff
@@ -280,7 +280,7 @@ function renderMember(member, parent, out) {
         throw new Error(`No Event Type for ${name} in ${parent.name}`);
       if (member.spec)
         output(XmlDoc.renderXmlDoc(member.spec, maxDocumentationColumnWidth));
-      if (parent && (classNameMap.get(parent.name) === type))
+      if (parent && (classNameMap.get(parent.name) === type) && !['Popup'].includes(name))
```
pavelfeldman, line 283: We managed to not hardcode things like this in Python, any way to work around it?

kblok:
closing in favor of #6076

PR #5985, opened 2021-03-29
chore(generators): use the same definition for viewPortSize

kblok:
The page viewportSize should use the same definition as the viewport used to create the context. They should reference the same type in code

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -2531,10 +2531,7 @@ Shortcut for main frame's [`method: Frame.url`].
 
 Video object associated with this page.
 
-## method: Page.viewportSize
-- returns: <[null]|[Object]>
-  - `width` <[int]> page width in pixels.
-  - `height` <[int]> page height in pixels.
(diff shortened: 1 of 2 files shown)


docs/src/api/class-page.md
```diff
@@ -2531,10 +2531,7 @@ Shortcut for main frame's [`method: Frame.url`].
 
 Video object associated with this page.
 
-## method: Page.viewportSize
-- returns: <[null]|[Object]>
-  - `width` <[int]> page width in pixels.
-  - `height` <[int]> page height in pixels.
+## method: Page.viewportSize = %%-option-viewport-%%
```
avodovnik, line 2534: This doesn't look right.

avodovnik (changes requested):
This breaks the lint/tip of tree test.

docs/src/api/class-page.md
kblok: Yeah, this won't work. It's taking the description of the new context here.

PR #6005, opened 2021-03-30
api(video): implement video.saveAs and video.delete

### LINKED ISSUE #5853: [Feature] Redirect video stream to new path
Some of us run Playwright as test cases, e.g. via Jest using `it` statements and test files (suites). When using [the awesome videos functionality](https://playwright.dev/docs/videos) it saves one huge video in case the browser context never closes between tests.

It would be cool to have hooks to `page.video()` e.g. `page.video().path('some-path')` to redirect the buffer to a new file. It would also be cool if I can chose the file name of the video in those scenarios, as I might want to name them as the test case. (I can add a UUID or similar myself if wanted).

dgozman:
These methods are safe to call while the page is still open, or when it is
already closed. Works in remotely connected browser as well.

Also makes `video.path()` to throw for remotely connected browser.

Under the hood migrated Download and Video to use the common Artifact object.

Fixes #5853.


src/client/video.ts
```diff
@@ -14,25 +14,62 @@
...
-  _setRelativePath(relativePath: string) {
-    this._pathCallback!(path.join(this._page.context()._options.recordVideo!.dir, relativePath));
+  _setImpl(impl: VideoImpl) {
+    this._impl = impl;
+    this._implCallback();
   }
 
-  path(): Promise<string> {
-    return this._pathPromise;
+  async path(): Promise<string> {
```
pavelfeldman, line 41: Can we inherit / aggregate Download and Video from Artifact?

src/dispatchers/pageDispatcher.ts
```diff
@@ -32,16 +32,19 @@ import { FileChooser } from '../server/fileChooser';
...
-  constructor(scope: DispatcherScope, page: Page) {
+  // Note: Video must outlive Page and BrowserContext, so that client can saveAs it
+  // after closing the context.
+  constructor(scope: DispatcherScope, page: Page, scopeForVideo: DispatcherScope) {
     // TODO: theoretically, there could be more than one frame already.
     // If we split pageCreated and pageReady, there should be no main frame during pageCreated.
     super(scope, page, 'Page', {
       mainFrame: FrameDispatcher.from(scope, page.mainFrame()),
-      videoRelativePath: page._video ? page._video._relativePath : undefined,
+      video: page._video ? new VideoDispatcher(scopeForVideo, page._video) : undefined,
```
pavelfeldman, line 47: If scope for video is on the parent level, the you should handle it on that level? Like Page has input, BrowserContext has video?

src/dispatchers/videoDispatcher.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Video } from '../server/browserContext';
+
+export class VideoDispatcher extends Dispatcher<Video, channels.VideoInitializer> implements channels.VideoChannel {
+  constructor(scope: DispatcherScope, video: Video) {
+    super(scope, video, 'Video', {
+      absolutePath: video._path,
+    });
+  }
+
+  async saveAs(params: channels.VideoSaveAsParams): Promise<channels.VideoSaveAsResult> {
```
pavelfeldman, line 32: Same here, let's not copy-paste entire classes!

src/server/artifact.ts
```diff
@@ -0,0 +1,117 @@
...
+      return this._unaccessibleErrorMessage;
+    await this._finishedPromise;
+    return this._failureError;
+  }
+
+  async delete(): Promise<void> {
+    if (this._unaccessibleErrorMessage)
+      return;
+    const fileName = await this.localPathAfterFinished();
+    if (this._deleted)
```
pavelfeldman, line 82: move this check up?

src/server/browser.ts
```diff
@@ -60,7 +61,7 @@ export abstract class Browser extends SdkObject {
   private _downloads = new Map<string, Download>();
   _defaultContext: BrowserContext | null = null;
   private _startedClosing = false;
-  readonly _idToVideo = new Map<string, Video>();
+  readonly _idToVideo = new Map<string, { context: BrowserContext, artifact: Artifact }>();
```
pavelfeldman, line 64: I'm a little worried you just leaked all the contexts.

pavelfeldman (approved):

src/server/artifact.ts
dgozman: If we move this up, there could be a race where another delete() happens during `await localPathAfterFinished()`.

src/server/browser.ts
dgozman: Sure, let's clear these videos in `BrowserContext._didCloseInternal` just in case.

Issue #6009, opened 2021-03-30
[BUG] WebKit on Mac crashes when RTCPeerConnection is used.


mxschmitt:
**Context:**
- Playwright Version: 1.10
- Operating System: Mac 11.2.2 and 10.15.7
- Node.js version: v14.8.0
- Browser: WebKit
- Original issue: https://github.com/playwright-community/jest-playwright/issues/659

**To Reproduce**

Execute this script:

```js
// @ts-check
const playwright = require('playwright');

(async () => {
  const browser = await playwright.webkit.launch();
  const context = await browser.newContext();
  const page = await context.newPage();
  await page.goto("https://www.whatismybrowser.com/")
  await page.screenshot({ path: `example.png` });
  await browser.close();
})();
```

or from the original bug report:

1. `git clone https://github.com/jsjoeio/jest-playwright-repro.git`
2. `yarn`
3. `yarn test`

**Describe the bug**

Actual: WebKit page crashes with `Navigation failed because page crashed`
Expected: Test passes

It works for Chromium and Firefox.

~~It seems like its because `https://whatismybrowser.com/` is a redirect -> 301 redirect.~~

mxschmitt:
Seems like related to their JavaScript:

```
it.only("foo", async ({ page, server }) => {
  await page.route("https://cdn.whatismybrowser.com/prod-website/static/main/js/site.min.js**", route => {
    return route.abort()
  })
  await page.goto('https://www.whatismybrowser.com/')
  expect(page.url()).toBe('https://www.whatismybrowser.com/')
});
```

this passes.

jsjoeio:
When will this fix land?

yury-s:
It's fixed in @next and will be released in v1.11.0

PR #6013, opened 2021-03-30
test: add test for new RTCPeerConnection()

mxschmitt:
Test for #6009

### DIFF
--- a/test/page-goto.spec.ts
+++ b/test/page-goto.spec.ts
@@ -499,3 +499,28 @@ it('should not throw unhandled rejections on invalid url', async ({page, server}
   const e = await page.goto('https://www.youtube Panel Title.com/').catch(e => e);
   expect(e.toString()).toContain('Panel Title');
 });
+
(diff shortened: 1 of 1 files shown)


test/page-goto.spec.ts
```diff
@@ -499,3 +499,28 @@ it('should not throw unhandled rejections on invalid url', async ({page, server}
   const e = await page.goto('https://www.youtube Panel Title.com/').catch(e => e);
   expect(e.toString()).toContain('Panel Title');
 });
+
+it('should not crash when RTCPeerConnection is used',(test, { browserName }) => {
+  test.fail(browserName === 'webkit');
```
yury-s, line 504: Does it fail on linux/windows too?

yury-s (approved):

test/page-goto.spec.ts
pavelfeldman: It passes on Linux for me. But we'll fix it shortly, so no big deal

PR #6017, opened 2021-03-31
docs: replace headful with headed

mxschmitt:
There are far more occurrences of headful than for headed, so let's stick with headful and be consistent?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -166,7 +166,7 @@ it.skip('should work', async ({server, page}) => {
 });
 ```
 
-- To run tests in non-headless (headful) mode:
+- To run tests in non-headless (headed) mode:
 
 ```bash
 HEADFUL=1 npm run ctest
--- a/docs/src/test-runners-python.md
+++ b/docs/src/test-runners-python.md
@@ -29,8 +29,8 @@ To run your tests, use pytest CLI.
 # Run tests (Chromium and headless by default)
 $ pytest
 
-# Run tests in headful mode
-$ pytest --headful
+# Run tests in headed mode
+$ pytest --headed
 
 # Run tests in a different browser (chromium, firefox, webkit)
 $ pytest --browser firefox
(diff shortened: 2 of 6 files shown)


pavelfeldman:
We are slowly going the other way around since headful has particular meaning not matching our needs.

mxschmitt:
> We are slowly going the other way around since headful has particular meaning not matching our needs.

shall I do it the other way around or just close it? Cypress e.g. is only using the headed wording everywhere.

pavelfeldman:
We can probably go headed everywhere. I hope it isn't too massive of a change.

docs/src/test-runners-python.md
```diff
@@ -29,8 +29,8 @@ To run your tests, use pytest CLI.
 # Run tests (Chromium and headless by default)
 $ pytest
 
-# Run tests in headful mode
-$ pytest --headful
+# Run tests in headed mode
+$ pytest --headed
```
mxschmitt, line 33: https://github.com/microsoft/playwright-pytest/pull/48

.github/workflows/tests.yml
```diff
@@ -129,8 +129,8 @@ jobs:
     - run: node lib/cli/cli install-deps
     - run: bash packages/installation-tests/installation-tests.sh
 
-  headful_linux:
```
pavelfeldman, line 132: Let's keep the original values for internal systems - easier to keep required bots assigned

CONTRIBUTING.md
```diff
@@ -166,10 +166,10 @@ it.skip('should work', async ({server, page}) => {
 });
 ```
 
-- To run tests in non-headless (headful) mode:
+- To run tests in non-headless (headed) mode:
 
 ```bash
-HEADFUL=1 npm run ctest
+HEADED=1 npm run ctest
```
pavelfeldman, line 172: ditto

src/server/browser.ts
```diff
@@ -41,7 +41,7 @@ export type BrowserOptions = PlaywrightOptions & {
   isChromium: boolean,
   channel?: types.BrowserChannel,
   downloadsPath?: string,
-  headful?: boolean,
+  headed?: boolean,
```
pavelfeldman, line 44: Also, let's keep the code as is for simplicity.

pavelfeldman (commented):
Let's only update the user-visible docs for now. Otherwise, there are too many moving parts!

pavelfeldman (approved):

PR #6040, opened 2021-04-01
api: remove Chromium* classes

pavelfeldman:


src/client/browserContext.ts
```diff
@@ -59,11 +65,23 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel,
     super(parent, type, guid, initializer);
     if (parent instanceof Browser)
       this._browser = parent;
+    this._isChromium = this._browser?._name === 'chromium';
```
dgozman, line 68: I think we should have `initializer.isChromium` instead, which is also set for Electron and Android.

types/types.d.ts
```diff
@@ -1484,10 +1484,11 @@ export interface Page {
   context(): BrowserContext;
 
   /**
-   * Browser-specific Coverage implementation, only available for Chromium atm. See
-   * [ChromiumCoverage](#class-chromiumcoverage) for more details.
+   * > NOTE: Only available for Chromium atm.
```
dgozman, line 1487: Can we rephrase to not mention "Automated Teller Machine"? 😄

dgozman (approved):

src/client/browserContext.ts
pavelfeldman: i'll follow up

types/types.d.ts
pavelfeldman: I'd rather not focus on it atm :P

PR #6044, opened 2021-04-02
fix: dispatch popup event on the client end

yury-s:


src/dispatchers/pageDispatcher.ts
```diff
@@ -38,13 +38,21 @@ import { Download } from '../server/download';
 export class PageDispatcher extends Dispatcher<Page, channels.PageInitializer> implements channels.PageChannel {
   private _page: Page;
 
+  private static from(scope: DispatcherScope, page: Page | undefined): PageDispatcher | undefined {
```
pavelfeldman, line 41: We called it createNullable in ElementHandle

src/server/chromium/crPage.ts
```diff
@@ -93,7 +93,13 @@ export class CRPage implements PageDelegate {
     }
     // Note: it is important to call |reportAsNew| before resolving pageOrError promise,
     // so that anyone who awaits pageOrError got a ready and reported page.
-    this._pagePromise = this._mainFrameSession._initialize(hasUIWindow).then(() => {
+    this._pagePromise = this._mainFrameSession._initialize(hasUIWindow).then(async r => {
+      await this._page.initOpener(this._opener);
```
pavelfeldman, line 97: This should be in page.ts because it is the same across ports?

pavelfeldman (commented):
Looks great, let's also call initOpener from page.ts

src/dispatchers/pageDispatcher.ts
```diff
@@ -38,13 +38,21 @@ import { Download } from '../server/download';
...
+    if (!page)
+      return undefined;
+    const result = existingDispatcher<PageDispatcher>(page);
```
dgozman, line 44: There must be a dispatcher for opener already, right? Perhaps we don't need to create a new one here, just lookup?

src/protocol/protocol.yml
```diff
@@ -644,6 +644,7 @@ Page:
         width: number
         height: number
     isClosed: boolean
+    opener: Page?
```
dgozman, line 647: Can also remove Page.opener method from the protocol.

src/server/firefox/ffPage.ts
```diff
@@ -46,6 +46,7 @@ export class FFPage implements PageDelegate {
   private _pagePromise: Promise<Page | Error>;
   private _pageCallback: (pageOrError: Page | Error) => void = () => {};
   _initializedPage: Page | null = null;
+  private _initialiazationFailed = false;
```
dgozman, line 49: There is a typo in this field.

src/dispatchers/pageDispatcher.ts
```diff
@@ -38,13 +38,21 @@ import { Download } from '../server/download';
...
+
   constructor(scope: DispatcherScope, page: Page) {
     // TODO: theoretically, there could be more than one frame already.
     // If we split pageCreated and pageReady, there should be no main frame during pageCreated.
     super(scope, page, 'Page', {
       mainFrame: FrameDispatcher.from(scope, page.mainFrame()),
       viewportSize: page.viewportSize() || undefined,
-      isClosed: page.isClosed()
+      isClosed: page.isClosed(),
+      opener: PageDispatcher.from(scope, page.opener())
```
dgozman, line 55: Even though we check that opener is not closed at the time we assign `page._opener`, it may be closed after that. Let's instead check it here, or somehow reset it to undefined upon opener closure.

test/page-basic.spec.ts
```diff
@@ -87,14 +87,14 @@ it('should provide access to the opener page', async ({page}) => {
   expect(opener).toBe(page);
 });
 
-it('should return null if parent page has been closed', async ({page}) => {
+it('should return parent page even if it has been closed', async ({page}) => {
```
dgozman, line 90: Why do we want this changed?

src/dispatchers/pageDispatcher.ts
yury-s: done

src/server/chromium/crPage.ts
yury-s: I considered placing it into `Page.reportAsNew` but it would require making that method async which I'd like to avoid and keep it sending sync events only. This code needs more cleanup anyway, I can fix that separately.

src/dispatchers/pageDispatcher.ts
yury-s: I suspect so, but I couldn't prove it strictly so I'd rather stay on the safe side and keep current logic.

src/server/firefox/ffPage.ts
yury-s: fixed

test/page-basic.spec.ts
yury-s: Reverted.

src/protocol/protocol.yml
yury-s: Done.

src/dispatchers/pageDispatcher.ts
yury-s: Moved this logic on the client end.

src/server/page.ts
```diff
@@ -191,13 +196,8 @@ export class Page extends SdkObject {
...
     }
     this._browserContext.emit(BrowserContext.Events.Page, this);
-    const openerDelegate = this._delegate.openerDelegate();
-    if (openerDelegate) {
-      openerDelegate.pageOrError().then(openerPage => {
-        if (openerPage instanceof Page && !openerPage.isClosed())
-          openerPage.emit(Page.Events.Popup, this);
-      });
-    }
+    if (this.isClosed())
```
dgozman, line 199: Let's add a comment explaining this one.

dgozman (approved):

src/server/page.ts
yury-s: Done.

PR #6048, opened 2021-04-02
test: migrate cli tests to new folio

dgozman:


tests/config/cliEnv.ts
```diff
@@ -0,0 +1,74 @@
...
+    const result = await super.beforeEach(testInfo);
+    const { page, context, toImpl } = result;
+    return {
+      ...result,
+      httpServer: {
+        setHandler: newHandler => this._handler = newHandler,
+        PREFIX: `http://127.0.0.1:${this._port}`,
+      },
+      runCLI: this._runCLI.bind(this),
+      openRecorder: async () => {
```
JoelEinbinder, line 51: Why does this change from `recorder` to `openRecorder`. It's because the new system doesn't magically resolve the fixtures I use?

dgozman: Exactly. We use cliTest for both recorder tests and runCLI tests, and I didn't want to open recorder for the latter.

JoelEinbinder: Is there still magic for the built in fixtures? Or do we launch a browser per test now?

Issue #5952, opened 2021-03-25
[BUG] Request Interception Breaks Worker importScripts with Remote Resource


rwoll:
**Context:**
- Playwright Version: 1.11.0-next (but effects at least the last couple of versions)
- Operating System: macOS
- Node.js version: v14.16.0
- Browser: Chromium
- Extra: -

**Code Snippet**

```javascript
it('should not break remote worker importScripts',  (test, { browserName }) => {
  test.fail(browserName === "chromium");
}, async ({ page, server, context }) => {
  context.route('**', async request => {
    await request.continue();
  });
  await page.goto(server.PREFIX + '/worker/worker-http-import.html');
  await page.waitForSelector("#status:has-text('finished')");
});
```

`/worker/worker-http-import.html`:

```html
<!DOCTYPE html>
<html>
<head>
    <title>Remote importScripts Test</title>
</head>
<html>
    <div id="status">-</div>
    <script>
    new Worker('worker-http-import.js').addEventListener("message", ({ data }) => {
        document.getElementById("status").innerText = data;
    });
    </script>
</html>
</html>
```

`/worker/worker-http-import.js`:

```javascript
console.log("hello from worker-http-import.js");
importScripts("./import-me.js")
console.log("successfully imported");
self.postMessage("finished");
```

`/worker/import-me.js`:

```javascript
console.log("hello from import-me.js");
```

**Describe the bug**

Request interception causes `importScripts("./import-me.js")` to hang. If you
disable request interception (i.e. comment out the `context.route(…)` bits),
the importScripts succeeds, `hello from import-me.js` is logged, and the `waitForSelector`
succeeds and matches the `finished` message that's bubbled up.

---

**Additional Debugging Notes**

* With `DEBUG=pw:protocol` we see a `Fetch.requestPaused` for `http://localhost:4747/import-me.js`,
  but no protocol messages about that domain after when we would eventually expect
  a `requestWillBeSent` and a resume.
* For `http://localhost:4747/import-me.js`, `requestWillBeSent` is `undefined` here in `crNetworkManager.js`: https://github.com/microsoft/playwright/blob/495085cbb2092bac71e36120e718e5cd92ad7374/src/server/chromium/crNetworkManager.ts#L182
* The test passes for Firefox and Webkit.
* Not all network requests originating from the worker hang. For example, in
  the below demo (`fetch-worker.js`) the `fetch` request succeeds.
[...]

PR #6055, opened 2021-04-02
browser(chromium): roll to 868876 (#5952)

rwoll:


yury-s:
We switched to rolling to Chromium Dev revisions instead of random ones. According to https://omahaproxy.appspot.com/ current dev is based on 867878.

rwoll:
👍 Ah, thanks! I'll close for now and then re-open once a dev revision includes the needed change if it hasn't already been rolled before then.

PR #6076, opened 2021-04-05
chore(dotnet): generate generic EventHandlers

kblok:
We are going to generate the generic `EventHandler` even if the generic type references the sender.

With this change, we are going to have, for instance, `event EventHandler<IPage> Close;` where the event args will be the same as `object sender` and `event EventHandler<IPage> Popup;`, where the event args will reference the popup page.


yury-s (approved):

avodovnik:
Just as a note, @kblok, this will likely require a bit more changes after we finish merging most of the new interfaces - i.e. we'll have to go back and make some additional adaptations.

avodovnik (approved):

PR #6118, opened 2021-04-07
chore(dotnet): generate console location object

kblok:


kblok:
Why do we have `Location` as a string in Java?

yury-s:
We found this is how people use it just printing to the console. We'd do the same for js and python but we missed the boat. Let's make it string in C# too as it hasn't been released yet.

PR #6119, opened 2021-04-07
chore(dotnet): generate coverage in dotnet

kblok:
We already have the implementation for this.


yury-s:
We try to minimize exposure of Chromium-only APIs in the new languages and decided to add them only when there is explicit justified request from the users. I'd stick to the same rule with C# even if the implementation is there.

PR #6138, opened 2021-04-08
docs: move waitUntil doc before timeout

kblok:
This will help positional argument developers (a.k.a. dotnet devs) to get the waitUntil argument, which I consider it will be more used, before the timeout argument.

### DIFF
--- a/docs/src/api/class-frame.md
+++ b/docs/src/api/class-frame.md
@@ -711,9 +711,9 @@ Headless mode doesn't support navigation to a PDF document. See the
 
 URL to navigate frame to. The url should include scheme, e.g. `https://`.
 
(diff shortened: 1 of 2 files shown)


dgozman (commented):
I am afraid this will break at any moment. I have just inserted some options in the middle, to keep them lexicographically ordered, and nothing stopped me. We either need a solution that warns us this will break C#, or let C# codegen handle it by itself (always move timeout to be last?)

kblok:
@dgozman agreed on that and that's beyond the timeout argument. **Any** argument reorder or argument renames should be considered a breaking change.

yury-s (approved):

PR #6146, opened 2021-04-08
fix(screencast): properly stop screencast on context closure

yury-s:
This fixes ['should capture static page in persistent context'](https://devops.aslushnikov.com/flakiness2.html#filter_spec=screencast) in Chromium Headful.

### DIFF
--- a/src/server/browserContext.ts
+++ b/src/server/browserContext.ts
@@ -93,10 +93,6 @@ export abstract class BrowserContext extends SdkObject {
     }
     this._closedStatus = 'closed';
     this._downloads.clear();
-    for (const [id, video] of this._browser._idToVideo) {
-      if (video.context === this)
-        this._browser._idToVideo.delete(id);
-    }
     this._closePromiseFulfill!(new Error('Context closed'));
     this.emit(BrowserContext.Events.Close);
   }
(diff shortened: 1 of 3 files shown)


src/server/chromium/crPage.ts
```diff
@@ -876,6 +879,7 @@ class FrameSession {
   async _startVideoRecording(options: types.PageScreencastOptions) {
     const screencastId = this._screencastId;
     assert(screencastId);
+    this._page.once(Page.Events.Close, () => this._stopVideoRecording());
```
dgozman, line 882: Do you want to catch this?

src/server/browserContext.ts
```diff
@@ -93,10 +93,6 @@ export abstract class BrowserContext extends SdkObject {
     }
     this._closedStatus = 'closed';
     this._downloads.clear();
-    for (const [id, video] of this._browser._idToVideo) {
```
dgozman, line 96: Why removing this? It does not hurt to cleanup, just in case something went wrong and we leaked stuff.

src/server/chromium/crPage.ts
yury-s: it's actually expected to not throw

src/server/browserContext.ts
yury-s: Because this is the reason the test is failing. The video gets removed from the map and after that reportFinished fails.

dgozman (approved):

PR #6151, opened 2021-04-08
fix: finish all artifacts when browser exits

yury-s:


src/server/browser.ts
```diff
@@ -118,6 +119,10 @@ export abstract class Browser extends SdkObject {
     if (this._defaultContext)
       this._defaultContext._browserClosed();
     this.emit(Browser.Events.Disconnected);
+
+    for (const video of this._idToVideo.values())
```
dgozman, line 123: Perhaps, emit disconnected event after this?

dgozman (approved):

src/server/browser.ts
yury-s: don't see a difference, swapped

Issue #6057, opened 2021-04-02
[BUG] page.check() unable to check specific checkbox because it always clicks in the middle


klhex:
**Context:**
- Playwright Version: 1.10.0
- Operating System: MacOS 11.2.3
- Node.js version: 15.11.0
- Browser: Chromium, Firefox

**Code Snippet**

```javascript
await page.check('text=Ich habe die');
```

**Describe the bug**

![image](https://user-images.githubusercontent.com/71178731/113423632-20a30500-93cf-11eb-9584-6e7666a9256e.png)

On a page with a sticky header (see above screenshot), PW is unable to check the first of three checkboxes, apparently due to that sticky header. Instead of checking the checkbox, PW actually clicks the second link included in the label of the checkbox and throws the following error, both in Chromium and Firefox, headless or headful (if I replicate the html code for the 3 checkboxes locally (w/o the sticky header which I cannot reproduce locally), all of them can be checked w/o any issue):

<details><summary>PW Logs</summary>

```bash
  pw:api => page.check started +1ms
  pw:api waiting for selector "text=Ich habe die" +1ms
  pw:api   selector resolved to visible <span class="c-checkbox__label">…</span> +3ms
  pw:api attempting click action +2ms
  pw:api   waiting for element to be visible, enabled and stable +0ms
  pw:api     element is not stable - waiting... +29ms
  pw:api   element is visible, enabled and stable +17ms
  pw:api   scrolling into view if needed +0ms
  pw:api   done scrolling +0ms
  pw:api   checking that element receives pointer events at (592.48,600.23) +1ms
  pw:api   element does receive pointer events +2ms
  pw:api   performing click action +0ms
  pw:api   click action done +134ms
  pw:api   waiting for scheduled navigations to finish +0ms
  pw:api   navigations have finished +173ms
  pw:api <= page.check failed +13ms

page.check: Clicking the checkbox did not change its state
=========================== logs ===========================
waiting for selector "text=Ich habe die"
  selector resolved to visible <span class="c-checkbox__label">…</span>
attempting click action
  waiting for element to be visible, enabled and stable
    element is not stable - waiting...
  element is visible, enabled and stable
  scrolling into view if needed
  done scrolling
  checking that element receives pointer events at (592.48,600.23)
  element does receive pointer events
  performing click action
  click action done
  waiting for scheduled navigations to finish
  navigations have finished
============================================================
```
</details>

When I try instead to make PW check that checkbox via its CSS ID it like so
```javascript
[...]

PR #6153, opened 2021-04-08
api: add option `position` to check/uncheck

dgozman:
Since check/uncheck does click under the hood, sometimes it might need to click at a different position.

One example would be a long label that contains links inside, and clicking in the center happens to hit the link instead of the label itself. Passing `position` is a possible solution.

References #6057.

### DIFF
--- a/docs/src/api/class-elementhandle.md
+++ b/docs/src/api/class-elementhandle.md
@@ -134,6 +134,8 @@ When all steps combined have not finished during the specified [`option: timeout
 
 ### option: ElementHandle.check.noWaitAfter = %%-input-no-wait-after-%%
 
+### option: ElementHandle.check.position = %%-input-position-%%
+
 ### option: ElementHandle.check.timeout = %%-input-timeout-%%
 
 ## async method: ElementHandle.click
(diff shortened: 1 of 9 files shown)


yury-s (commented):
What's the added value of this option compared to getting the elements bounding box and doing `page.mouse.click(x, y)` at the desired point?

dgozman:
> What's the added value of this option compared to getting the elements bounding box and doing `page.mouse.click(x, y)` at the desired point?

Simplicity! I expect this to be used as `page.check('selector', { position: { x: 10, y: 10 } })`.

docs/src/api/class-elementhandle.md
```diff
@@ -134,6 +134,8 @@ When all steps combined have not finished during the specified [`option: timeout
 
 ### option: ElementHandle.check.noWaitAfter = %%-input-no-wait-after-%%
 
+### option: ElementHandle.check.position = %%-input-position-%%
```
yury-s, line 137: how about offset to avoid confusion with position within viewport ?

yury-s (approved):

Issue #5993, opened 2021-03-30
[Feature] Firefox release-branch channel


pavelfeldman:
It'd be nice to be able to test against Firefox built off the non-beta release branch.

steps:

- [x] build firefox-stable + juggler
- [x] add ability to install firefox-stable via cli
- [x] run tests for firefox-stable
- [ ] add types + documentation

PR #6177, opened 2021-04-10
feat: start downloading firefox-stable channel

aslushnikov:
References #5993

### DIFF
--- a/src/server/firefox/firefox.ts
+++ b/src/server/firefox/firefox.ts
@@ -18,6 +18,7 @@
 import * as os from 'os';
 import fs from 'fs';
 import path from 'path';
+import { assert } from '../../utils/utils';
 import { FFBrowser } from './ffBrowser';
 import { kBrowserCloseMessageId } from './ffConnection';
 import { BrowserType } from '../browserType';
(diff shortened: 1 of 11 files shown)


yury-s (approved):

src/server/firefox/firefox.ts
```diff
@@ -31,6 +32,18 @@ export class Firefox extends BrowserType {
     super('firefox', playwrightOptions);
   }
 
+  executablePath(options?: types.LaunchOptions): string {
+    if (options?.channel) {
+      let executablePath = undefined;
+      if ((options.channel as any) === 'stable')
```
mxschmitt, line 38: Instead of calling it stable, let's call it maybe `firefox-stable` and also add it to the `BrowserChannel` type?

mxschmitt (approved):

src/server/firefox/firefox.ts
aslushnikov: Agreed on `firefox-stable`. Will probably postpone adding to `BrowserChannel` type for now to avoid documenting.

PR #6181, opened 2021-04-11
chore(dotnet): add request argument to route handlers

kblok:


avodovnik (approved):
I assume the generator renders the output you expect right? LGTM in that case.

kblok:
@avodovnik it does :)

yury-s (commented):
We dropped request param in java because it is redundant with `Route.request()`. Can we actually do the same in C# for the sake of simpler API?

kblok:
Great point @yury-s

PR #6184, opened 2021-04-12
chore(dotnet): waitForCloseAsync

kblok:
[Don't Block on Async Code](https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html)

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -2544,22 +2544,13 @@ Video object associated with this page.
   - `width` <[int]> page width in pixels.
   - `height` <[int]> page height in pixels.
 
-## method: Page.waitForClose
-* langs: java
+## async method: Page.waitForClose
+* langs: csharp, java
 - returns: <[Page]>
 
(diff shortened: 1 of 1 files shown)


avodovnik (changes requested):
This doesn't look like a change that warrants its own method declaration. You can just add the parameter and filter it to `csharp`, to the original method. IIRC, that should work.

docs/src/api/class-page.md
```diff
@@ -2545,13 +2545,22 @@ Video object associated with this page.
   - `height` <[int]> page height in pixels.
 
 ## method: Page.waitForClose
```
avodovnik, line 2547: @yury-s I think you ignore the `async` method in Java generation right, so @kblok, I think you can just mark this as `async`.

avodovnik (approved):
Looks good to me, but I just want to get confirmation from @yury-s before I merge it :)

yury-s (approved):

PR #5318, opened 2021-02-04
refactor: remove browserPaths in favor of Registry class

aslushnikov:
This patch introduces a new Registry class that incapsulates
all logic regarding browsers and their paths.

Fixes #5278

PR #6186, opened 2021-04-12
chore: read browsers.json with require

### LINKED ISSUE #5862: [BUG] Error: ENOENT: no such file or directory, open '/var/task/node_modules/playwright-core/browsers.json' on Vercel with Next.js
After upgrade Playwright to 1.9.2 this error is happening on Vercel functions.

```
2021-03-16T19:57:07.806Z	undefined	ERROR	Error: ENOENT: no such file or directory, open '/var/task/node_modules/playwright-core/browsers.json'
    at Object.openSync (fs.js:462:3)
    at Object.readFileSync (fs.js:364:35)
    at new Registry (/var/task/node_modules/playwright-core/lib/utils/registry.js:213:54)
    at new Playwright (/var/task/node_modules/playwright-core/lib/server/playwright.js:46:23)
    at Object.createPlaywright (/var/task/node_modules/playwright-core/lib/server/playwright.js:59:12)
    at setupInProcess (/var/task/node_modules/playwright-core/lib/inprocess.js:24:37)
    at Object.<anonymous> (/var/task/node_modules/playwright-core/lib/inprocess.js:42:18)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/var/task/node_modules/playwright-core/browsers.json'
}
RequestId: fafc5e6c-b338-4ba2-90b0-a019fd78923b Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

mxschmitt:
This fixes the compatibility on Vercel with Next.js when it's used in
a serverless function.
Next.js uses https://github.com/vercel/nft to trace down the
dependencies which a serverless function is using which
is currently not capable of detecting the browsers.json in our current
setup. Previously we used `require` to load the browers.json which was
replaced by `readFileSync` in #5318. Since then it was broken.

`require` has not much down-sides except its caching which in our-case
should not have any side-effects.

Without that it caused a `Error: ENOENT: no such file or directory, open '/var/task/node_modules/playwright-core/browsers.json'`.

Fixes #5862

### DIFF
--- a/src/utils/registry.ts
+++ b/src/utils/registry.ts
@@ -16,7 +16,6 @@
  */
 
 import { execSync } from 'child_process';
-import fs from 'fs';
 import * as os from 'os';
 import path from 'path';
 import * as util from 'util';
@@ -235,7 +234,9 @@ export class Registry {
   }
 
   constructor(packagePath: string) {
-    const browsersJSON = JSON.parse(fs.readFileSync(path.join(packagePath, 'browsers.json'), 'utf8'));
+    // require() needs to be used there otherwise it breaks on Vercel serverless
+    // functions. See https://github.com/microsoft/playwright/pull/6186
+    const browsersJSON = require(path.join(packagePath, 'browsers.json'));
     this._descriptors = browsersJSON['browsers'].map((obj: any) => {
       const name = obj.name;
       const revisionOverride = (obj.revisionOverrides || {})[hostPlatform];


src/utils/registry.ts
```diff
@@ -235,7 +234,9 @@ export class Registry {
   }
 
   constructor(packagePath: string) {
-    const browsersJSON = JSON.parse(fs.readFileSync(path.join(packagePath, 'browsers.json'), 'utf8'));
+    // require() needs to be used there otherwise it breaks on Vercel serverless
+    // functions. See https://github.com/microsoft/playwright/pull/6186
+    const browsersJSON = require(path.join(packagePath, 'browsers.json'));
```
pavelfeldman, line 239: I would call require on the call site where it is relative

aslushnikov (approved):

juice49:
Unfortunately, this fix no longer works. It seems like bigger changes would need to be made in order for the `browsers.json` file to be included in the files traced by NFT.

One of the NFT maintainers has provided more info about this at [vercel/nft#211](https://github.com/vercel/nft/issues/211#issuecomment-846407965).

mxschmitt:
Interesting, maybe adding a special case would work for us? https://github.com/vercel/nft/blob/main/src/utils/special-cases.ts

yizhiyan:
Try to set the interface for playwright with electron, but the prompt after introduction fails to find browsers.json

PR #6187, opened 2021-04-12
fix(codegen): move injected recorder scripts to utility world

yury-s:
This way we avoid clashes with overridden intrinsic methods. E.g. overridden String.trim() on cvs.com leads to >1s spent in elementMatchesText on every mouse move.


dgozman (approved):

pavelfeldman:
Looks broken and abandoned.

PR #6191, opened 2021-04-13
docs: change the position argument location in check functions

kblok:
I think `position` is more relevant to the user than the "noWaitAfter, force, timeout" pack


avodovnik:
@yury-s does this affect Java at all?

avodovnik (approved):
I'm fine with this change, as long as it doesn't break other languages - i.e. at least @yury- s needs to approve as well, before we merge it.

pavelfeldman (approved):

PR #6198, opened 2021-04-14
docs(dotnet): introduce separate csharp viewport option

avodovnik:
Due to our handling of `null`, we want to be explicit in the docs, that the user needs to pass in a special value for `undefined`.

### DIFF
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -209,14 +209,22 @@ Whether to ignore HTTPS errors during navigation. Defaults to `false`.
 Toggles bypassing page's Content-Security-Policy.
 
 ## context-option-viewport
-* langs: js, java, csharp
+* langs: js, java
   - alias-java: viewportSize
-  - alias-csharp: viewportSize
 - `viewport` <[null]|[Object]>
   - `width` <[int]> page width in pixels.
   - `height` <[int]> page height in pixels.
 
 Emulates consistent viewport for each page. Defaults to an 1280x720 viewport. `null` disables the default viewport.
+
+## csharp-context-option-viewport
+* langs: csharp
+  - alias-csharp: viewportSize
+- `viewport` <[null]|[Object]>
+  - `width` <[int]> page width in pixels.
+  - `height` <[int]> page height in pixels.
+
+Emulates consistent viewport for each page. Defaults to an 1280x720 viewport. Use `ViewportSize.NoViewport` to disable the default viewport.
 
 ## context-option-screen
 * langs:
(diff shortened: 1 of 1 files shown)


avodovnik:
The other option here, would be to introduce code examples for all of the languages to demonstrate how to disable the viewport, but that feels like a higher overhead than the proposed solution.

Now, if only we had language-specific notes >:-]

pavelfeldman:
I'd either follow js or python, no need for unique approach in c#

pavelfeldman:
https://playwright.dev/python/docs/api/class-browser/#browsernew_contextkwargs - check out `no_viewport`

pavelfeldman:
Looks abandoned.

avodovnik:
@pfeldman In C#, the approach of using a constant is more familiar than using a separate parameter. We also have a `ViewportSize.Default` for that same reason. The _better_ option would be to have the default value be set by the generator, but that goes against the other options where we say if it's `null`, the value is default from the driver.

For that reason, I'd like to keep it this way.

kblok:
@pavelfeldman @avodovnik can we land this PR? :)

docs/src/api/params.md
```diff
@@ -218,6 +218,15 @@ Toggles bypassing page's Content-Security-Policy.
 
 Emulates consistent viewport for each page. Defaults to an 1280x720 viewport. `null` disables the default viewport.
 
+## context-option-viewport-csharp
```
yury-s, line 221: nit: rename to `csharp-context-option-viewport` to match other keys where the language goes first, e.g. `python-context-option-viewport`. It'd arguably be better to make language a suffix so that all overload definitions are grouped together but let's be consistent with the current approach, we can rename all of them at once if need be.

yury-s (approved):

yury-s (approved):

docs/src/api/params.md
```diff
@@ -209,15 +209,23 @@ Whether to ignore HTTPS errors during navigation. Defaults to `false`.
...
 Emulates consistent viewport for each page. Defaults to an 1280x720 viewport. `null` disables the default viewport.
 
+## csharp-context-option-viewport
+* langs: csharp
+  - alias-csharp: viewportSize
+- `viewport` <[null]|[Object]>
+  - `width` <[int]> page width in pixels.
+  - `height` <[int]> page height in pixels.
+
+Emulates consistent viewport for each page. Defaults to an 1280x720 viewport. Use `ViewportSize.NoViewport` to disable the default viewport.
```
yury-s, line 227: An afterthought: maybe rename it to NoDefaultViewport for clarity?

kblok: Not strong to say no, but I don't know if it will bring more clarity. If you start typing `ViewportSize.` as a static call, you kinda know where you are going to... you are going to `No(something)`.

PR #6201, opened 2021-04-14
feat(waitForEvent): allow async predicate

### LINKED ISSUE #6199: [Question] How to wait for several responses that differs only with body?
After clicking button multiple requests to GraphQL are made. URL is the same, only body is different. I tried the following:

```
async uploadFile() {
    const [fileChooser] = await Promise.all([
        this.page.waitForEvent('filechooser'),
        this.page.click(this.elements.csvInput)
    ]);
    await fileChooser.setFiles('./resources/upload_findings.csv');
    await this.page.click(this.elements.uploadButton);
}

async function waitForFindingUpload(page, findingName) {
    return await page.waitForResponse(async (response) => {
        let body = await response.text();
        return body.includes(findingName)
    }, {
        timeout: 10000
    });
}
const [response1, response2] = await Promise.all([
    waitForFindingUpload(page, "Name of Finding 1"),
    waitForFindingUpload(page, "Name of Finding 2"),
    uploadFindingsPage.uploadFile()
])
```
These response1, response2 are both either "Name of Finding 1" or "Name of Finding 2". I can not get both responses. 
How can I do this ?

dgozman:
We await the predicate now.

For synchronous predicates nothing should change because the await will finish before the next protocol message arrives (thanks to `makeWaitForNextTask`).

Fixes #6199.

### DIFF
--- a/src/client/waiter.ts
+++ b/src/client/waiter.ts
@@ -42,12 +42,12 @@ export class Waiter {
     return new Waiter(channelOwner, `${target}.waitForEvent(${event})`);
   }
 
-  async waitForEvent<T = void>(emitter: EventEmitter, event: string, predicate?: (arg: T) => boolean): Promise<T> {
+  async waitForEvent<T = void>(emitter: EventEmitter, event: string, predicate?: (arg: T) => boolean | Promise<boolean>): Promise<T> {
     const { promise, dispose } = waitForEvent(emitter, event, predicate);
     return this.waitForPromise(promise, dispose);
   }
 
-  rejectOnEvent<T = void>(emitter: EventEmitter, event: string, error: Error, predicate?: (arg: T) => boolean) {
+  rejectOnEvent<T = void>(emitter: EventEmitter, event: string, error: Error, predicate?: (arg: T) => boolean | Promise<boolean>) {
     const { promise, dispose } = waitForEvent(emitter, event, predicate);
     this._rejectOn(promise.then(() => { throw error; }), dispose);
   }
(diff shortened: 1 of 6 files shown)


src/client/waiter.ts
```diff
@@ -92,12 +92,12 @@ export class Waiter {
   }
 }
 
-function waitForEvent<T = void>(emitter: EventEmitter, event: string, predicate?: (arg: T) => boolean): { promise: Promise<T>, dispose: () => void } {
+function waitForEvent<T = void>(emitter: EventEmitter, event: string, predicate?: (arg: T) => any): { promise: Promise<T>, dispose: () => void } {
```
yury-s, line 95: Why not `=> boolean | Promise<boolean>` ?

tests/page-wait-for-response.spec.ts
```diff
@@ -63,6 +63,44 @@ it('should work with predicate', async ({page, server}) => {
...
+    page.evaluate(() => {
+      fetch('/simple.json').then(r => r.json());
+      fetch('/file-to-upload.txt').then(r => r.text());
+    })
+  ]);
+  expect(response1.url()).toBe(server.PREFIX + '/file-to-upload.txt');
+  expect(response2.url()).toBe(server.PREFIX + '/simple.json');
+});
+
+it('sync predicate should be only called once', async ({page, server}) => {
```
yury-s, line 86: why is it specific to sync ?

yury-s (approved):

tests/page-wait-for-response.spec.ts
dgozman: Because async predicate takes time, and we might concurrently call another one if the next event comes quickly enough. For example, responses are coming, and we are fetching the body to find the right response.

yury-s: Makes sense!

PR #6207, opened 2021-04-16
fix(chromium): drag and drop works in chromium

### LINKED ISSUE #1094: [Feature] drag and drop

JoelEinbinder:
Waiting for #6203 to percolate to the cdn. But this all works locally.

Fixes #1094

### DIFF
--- a/src/server/chromium/crDragDrop.ts
+++ b/src/server/chromium/crDragDrop.ts
@@ -1,0 +1,137 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+import { CRPage } from './crPage';
+import * as types from '../types';
+import { assert } from '../../utils/utils';
+import { Protocol } from './protocol';
+import { toModifiersMask } from './crProtocolHelper';
+
+declare global {
+  interface Window {
+    __cleanupDrag?: () => Promise<boolean>;
+  }
+}
+
+export class DragManager {
+  private _crPage: CRPage;
+  private _dragState: Protocol.Input.DragData | null = null;
+  private _lastPosition = {x: 0, y: 0};
+  constructor(page: CRPage) {
+    this._crPage = page;
+  }
+
+  async cancelDrag() {
+    if (!this._dragState)
+      return false;
+    await this._crPage._mainFrameSession._client.send('Input.dispatchDragEvent', {
+      type: 'dragCancel',
+      x: this._lastPosition.x,
+      y: this._lastPosition.y,
+      data: {
+        items: [],
+        dragOperationsMask: 0xFFFF,
+      }
+    });
+    this._dragState = null;
+    return true;
+  }
+
+  async interceptDragCausedByMove(x: number, y: number, button: types.MouseButton | 'none', buttons: Set<types.MouseButton>, modifiers: Set<types.KeyboardModifier>, moveCallback: () => Promise<void>): Promise<void> {
+    this._lastPosition = {x, y};
+    if (this._dragState) {
+      await this._crPage._mainFrameSession._client.send('Input.dispatchDragEvent', {
+        type: 'dragOver',
+        x,
+        y,
+        data: this._dragState,
+        modifiers: toModifiersMask(modifiers),
+      });
+      return;
+    }
+    if (button !== 'left')
+      return moveCallback();
+
+    const client = this._crPage._mainFrameSession._client;
+    let onDragIntercepted: (payload: Protocol.Input.dragInterceptedPayload) => void;
+    const dragInterceptedPromise = new Promise<Protocol.Input.dragInterceptedPayload>(x => onDragIntercepted = x);
+
+    await Promise.all(this._crPage._page.frames().map(async frame => {
+      await frame.nonStallingEvaluateInExistingContext((function() {
+        let didStartDrag = Promise.resolve(false);
+        let dragEvent: Event|null = null;
+        const dragListener = (event: Event) => dragEvent = event;
+        const mouseListener = () => {
+          didStartDrag = new Promise<boolean>(callback => {
+            window.addEventListener('dragstart', dragListener, {once: true, capture: true});
+            setTimeout(() => callback(dragEvent ? !dragEvent.defaultPrevented : false), 0);
+          });
+        };
+        window.addEventListener('mousemove', mouseListener, {once: true, capture: true});
+        window.__cleanupDrag = async () => {
+          const val = await didStartDrag;
+          window.removeEventListener('mousemove', mouseListener, {capture: true});
+          window.removeEventListener('dragstart', dragListener, {capture: true});
+          return val;
+        };
+      }).toString(), true, 'utility').catch(() => {});
+    }));
+
+    client.on('Input.dragIntercepted', onDragIntercepted!);
+    try {
+      await client.send('Input.setInterceptDrags', {enabled: true});
+    } catch {
+      // If Input.setInterceptDrags is not supported, just do a regular move.
+      // This can be removed once we stop supporting old Electron.
+      client.off('Input.dragIntercepted', onDragIntercepted!);
+      return moveCallback();
+    }
+    await moveCallback();
+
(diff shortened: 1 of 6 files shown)


src/server/chromium/crDragDrop.ts
```diff
@@ -0,0 +1,112 @@
...
+    const dragSignalPromises: Promise<boolean>[] = [];
+    for (const frame of this._page._page.frames()) {
+      const promise = frame.evaluateExpression((function(){
+        return new Promise<boolean>(callback => {
+          document.addEventListener('mousemove', () => setTimeout(() => callback(false), 0), {once: true});
+          document.addEventListener('dragstart', () => callback(true), {once: true});
+        });
+      }).toString(), true, undefined, 'utility');
+      // don't resolve on errors. This could be replaced by Promise.any after Node v15
+      dragSignalPromises.push(promise.catch(e => new Promise(() => {})));
```
dgozman, line 72: Why not resolve on errors? What if we have just a single frame and is has just navigated under our feet? We'll stall forever.

src/server/chromium/crDragDrop.ts
```diff
@@ -0,0 +1,112 @@
...
+    const client = this._page._mainFrameSession._client;
+    let onDragIntercepted: (payload: Protocol.Input.dragInterceptedPayload) => void;
+    const dragInterceptedPromise = new Promise<Protocol.Input.dragInterceptedPayload>(x => onDragIntercepted = x);
+
+    const dragSignalPromises: Promise<boolean>[] = [];
+    for (const frame of this._page._page.frames()) {
+      const promise = frame.evaluateExpression((function(){
+        return new Promise<boolean>(callback => {
+          document.addEventListener('mousemove', () => setTimeout(() => callback(false), 0), {once: true});
+          document.addEventListener('dragstart', () => callback(true), {once: true});
```
dgozman, line 68: Check defaultPrevented?

src/server/chromium/crDragDrop.ts
```diff
@@ -0,0 +1,112 @@
...
+  constructor(page: CRPage) {
+    this._page = page;
+  }
+
+  async cancelDrag() {
+    if (!this._dragState)
+      return false;
+    await this._page._mainFrameSession._client.send('Input.dispatchDragEvent', {
+      type: 'dragCancel',
+      x: 0,
```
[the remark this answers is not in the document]
JoelEinbinder: Reasonable! I was going based off of Firefox, which has (0,0) for the position on all `dragend` events. But Chromium does report the position of `dragend`.

src/server/chromium/crDragDrop.ts
JoelEinbinder: It doesn't stall because these are used with `Promise.race`

JoelEinbinder:
Ready for review!

tests/page/page-drag.spec.ts
```diff
@@ -19,8 +19,8 @@ import { test as it, expect } from './pageTest';
 import { attachFrame } from '../config/utils';
 
 it.describe('Drag and drop', () => {
-  it.skip(({ isAndroid }) => isAndroid);
-  it.fixme(({ browserName }) => browserName === 'chromium');
+  it.skip(({isAndroid}) => isAndroid);
+  it.skip(({isChromium, browserMajorVersion}) => isChromium && browserMajorVersion < 91);
```
dgozman, line 23: I was under impression that we are waiting for the stable to pick up the protocol changes.

tests/page/page-drag.spec.ts
```diff
@@ -174,6 +173,40 @@ it.describe('Drag and drop', () => {
...
+    await page.hover('#target');
+    await page.mouse.up();
+    expect(await page.$eval('#target', target => target.contains(document.querySelector('#source')))).toBe(true);
+  });
+
+  it('should be able to drag the mouse in a frame', async ({page, server}) => {
+    await page.goto(server.PREFIX + '/frames/one-frame.html');
+    await page.mouse.move(30, 30);
+    await page.mouse.down();
+    await page.mouse.move(60, 60);
```
dgozman, line 208: Let's do a mouseup just in case.

src/server/chromium/crDragDrop.ts
```diff
@@ -0,0 +1,134 @@
...
+            delete (window as any).dontWaitForMouseMove;
+            setTimeout(() => callback(dragEvent ? !dragEvent.defaultPrevented : false), 0);
+          }, {once: true, capture: true});
+          window.addEventListener('dragstart', event => dragEvent = event, {once: true, capture: true});
+        });
+      }).toString(), true, undefined, 'utility');
+      dragSignalPromises.push(promise.catch(() => false));
+    }
+    // make sure that all previous evaluate expressions hit the renderer
+    await this._crPage._page.mainFrame().evaluateExpression('1', false, undefined, 'utility').catch(() => {});
```
dgozman, line 81: You have to do this for all oopifs, I guess.

src/server/chromium/crDragDrop.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (button !== 'left')
+      return moveCallback();
+
+    const client = this._crPage._mainFrameSession._client;
+    let onDragIntercepted: (payload: Protocol.Input.dragInterceptedPayload) => void;
+    const dragInterceptedPromise = new Promise<Protocol.Input.dragInterceptedPayload>(x => onDragIntercepted = x);
+
+    const dragSignalPromises: Promise<boolean>[] = [];
+    for (const frame of this._crPage._page.frames()) {
+      const promise = frame.evaluateExpression((function() {
```
dgozman, line 67: I think you need something like `nonStallingRawEvaluateInExistingMainContext` her (there is such a function) to avoid stalling mouse move because some frame is perpetually navigating.

tests/page/page-drag.spec.ts
JoelEinbinder: There is a try-catch in the code.

tests/page/page-drag.spec.ts
JoelEinbinder: Done. Added an event tracker as well.

src/server/chromium/crDragDrop.ts
```diff
@@ -0,0 +1,129 @@
...
+      client.off('Input.dragIntercepted', onDragIntercepted!);
+      return moveCallback();
+    }
+    await moveCallback();
+
+    const expectingDrag = (await Promise.all(this._crPage._page.frames().map(async frame => {
+      return frame.evaluateExpression((() => window.__didStartDrag).toString(), true, undefined, 'utility').catch(() => false);
+    }))).some(x => x);
+    this._dragState = expectingDrag ? (await dragInterceptedPromise).data : null;
+    client.off('Input.dragIntercepted', onDragIntercepted!);
```
yury-s, line 99: Consider removing 'mousemove' listeners too

src/server/chromium/crDragDrop.ts
```diff
@@ -0,0 +1,129 @@
...
+        }, {once: true, capture: true});
+      }).toString(), true, undefined, 'utility').catch(() => {});
+    }));
+
+    client.on('Input.dragIntercepted', onDragIntercepted!);
+    try {
+      await client.send('Input.setInterceptDrags', {enabled: true});
+    } catch {
+      // If Input.setInterceptDrags is not supported, just do a regular move.
+      // This can be removed once Input.setInterceptDrags lands into Chromium Stable.
```
dgozman, line 89: This will be needed for quite a while for Electron, right? Let's update the comment.

src/server/chromium/crDragDrop.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    if (button !== 'left')
+      return moveCallback();
+
+    const client = this._crPage._mainFrameSession._client;
+    let onDragIntercepted: (payload: Protocol.Input.dragInterceptedPayload) => void;
+    const dragInterceptedPromise = new Promise<Protocol.Input.dragInterceptedPayload>(x => onDragIntercepted = x);
+
+    await Promise.all(this._crPage._page.frames().map(async frame => {
+      await frame.evaluateExpression((function() {
```
dgozman, line 72: I still think that awaiting all frames when some oopif could be perpetually navigating is going to stall. Did we figure out the solution?

JoelEinbinder: I'll go add a test

src/server/chromium/crDragDrop.ts
JoelEinbinder: They have `{once: true}` and have no side effects, so I am not worried about them.

bsteffensTempus:
What is the correct way to implement this functionality from within a playwright test?

Dospios:
How use that from tests?

JoelEinbinder:
Today you can just do `page.mouse.down`, `page.mouse.move`, and `page.mouse.up`. Follow #7562 for the high level api.

PR #6208, opened 2021-04-16
docs: add instructions for executing first script

kevin-brotcke:
Make it a little clearer how to execute your first script when going
through the introduction documentation.

### DIFF
--- a/docs/src/intro-js.md
+++ b/docs/src/intro-js.md
@@ -55,6 +55,12 @@ const { webkit } = require('playwright');
 })();
 ```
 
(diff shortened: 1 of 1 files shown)


pavelfeldman:
We require certain level of familiarity with Node.js, which includes ability to run scripts. But let's say we'd like to welcome Java folks that migrate from other tools. Your snippet helps making one of the steps, but there is also an `npm init` and `npm install` steps. If you are up for adding `npm init` and `npm i`, `node file.js` would also fit.

kevin-brotcke:
This pull request became moot after playwright test runner release.

PR #6214, opened 2021-04-16
fix(inspector): fall back to custom executable path for UI

### LINKED ISSUE #6158: [Question] How to run chrome in debug mode from executable path?
In my organization i can't download executable files from internet. Chrome version managed by oraganizationsecurity department.
Playwright works fine on chrome which i passed to executablePath: but in debug mode playwright tries to find executable in path where it should be downloaded by playwright. How can i change executablePath for debug mode? (PWDEBUG=1)

yury-s:
Fixes #6158

### DIFF
--- a/packages/installation-tests/installation-tests.sh
+++ b/packages/installation-tests/installation-tests.sh
@@ -39,6 +39,7 @@ mkdir -p "${TEST_ROOT}"
 NODE_VERSION="$(node --version)"
 
 function copy_test_scripts {
+  cp "${SCRIPTS_PATH}/inspector-custom-executable.js" .
   cp "${SCRIPTS_PATH}/sanity.js" .
   cp "${SCRIPTS_PATH}/screencast.js" .
   cp "${SCRIPTS_PATH}/esm.mjs" .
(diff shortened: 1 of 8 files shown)


packages/installation-tests/installation-tests.sh
```diff
@@ -194,6 +196,32 @@ function test_skip_browser_download {
...
+  fi
+
+  if [[ "$(uname)" != "Linux" ]]; then
+    echo
+    echo "Skipping test on non-Linux platform"
+    echo
+    return
+  fi
+
+  OUTPUT=$(PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 PWDEBUG=1 node inspector-custom-executable.js)
```
dgozman, line 216: Just do double-check: `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1` is not required here, right?

dgozman (approved):

packages/installation-tests/installation-tests.sh
yury-s: Yes, you are right

yury-s: Removed.

PR #6216, opened 2021-04-17
feat(chromium): bump to r878941

yury-s:
Blocked by https://bugs.chromium.org/p/chromium/issues/detail?id=1200964

~Since roll to r870763 failed, we are rolling to r871980 which contains relevant upstream fix.~
~Roll to r871980 failed due to missing vk_switshader.dll, we are rolling to a newer r874500.~
~Roll to r874500 failed due to a crash (http://crbug.com/1203505), we are rolling to a newer r876873 which includes a fix~
Roll to r878941


dgozman (approved):

tests/download.spec.ts
```diff
@@ -378,7 +378,7 @@ it.describe('download event', () => {
       page.context().close(),
     ]);
     expect(downloadPath).toBe(null);
-    expect(saveError.message).toContain('File deleted upon browser context closure.');
+    expect(saveError.message).toContain('File not found on disk. Check download.failure() for details.');
```
JoelEinbinder, line 381: I don't think this is the correct error message, but if it is it needs to be conditional on the browser version here.

yury-s: The message has changed because we now map  download state `INTERRUPTED` to `cancelled` on [chrome side](https://chromium-review.googlesource.com/c/chromium/src/+/2804358/8/content/browser/devtools/protocol/page_handler.cc#592) (previously that state was incorrectly mapped to `InProgress`). Due to this change we know receive download cancelled  event during context closure and that is the first error dispatched to the download artifact in playwright. It looks legit to me, I'll add a condition based on browser version as you suggest.

JoelEinbinder (approved):

pavelfeldman (approved):

yury-s:
I'll revert the flag that forces swiftshader in https://github.com/microsoft/playwright/pull/6300, it will fix windows tests.

aslushnikov (approved):

aslushnikov (approved):

PR #6224, opened 2021-04-19
feat(tracing): keyboard navigate lists

pavelfeldman:


src/web/traceViewer/ui/actionList.tsx
```diff
@@ -33,7 +33,27 @@ export const ActionList: React.FC<ActionListProps> = ({
   onSelected = () => {},
   onHighlighted = () => {},
 }) => {
-  return <div className='action-list'>{actions.map(actionEntry => {
+  return <div
+    className='action-list'
+    tabIndex={0}
```
JoelEinbinder, line 38: The more correct way to do this would be to put tabIndex=0 on the currently selected action, and tabIndex=-1 on all others. But you do you :)

dgozman (approved):

PR #5820, opened 2021-03-14
infra(channel): wire release channel to all tests

pavelfeldman:

PR #6227, opened 2021-04-19
fix: browser dependency validation

### LINKED ISSUE #6226: [Bug] [Regression] Host dependencies don't get validated anymore
Hi,

I can't make playwright to work with webkit on Ubuntu 20.04.2. 

I tested :
```python
import time

with sync_playwright() as p:
    browser = p.webkit.launch(headless=False)
    page = browser.new_page()
    page.goto('http://whatsmyuseragent.org/')
    time.sleep(5)
    browser.close()
```
and got the following error :
```python
File "run.py", line 18, in <module>
    page = browser.new_page()
  File "/home/***/venv/lib/python3.8/site-packages/playwright/sync_api/_generated.py", line 8751, in new_page
    self._sync(
  File "/home/***/venv/lib/python3.8/site-packages/playwright/_impl/_sync_base.py", line 103, in _sync
    return task.result()
  File "/home/***/venv/lib/python3.8/site-packages/playwright/_impl/_browser.py", line 137, in new_page
    page = await context.new_page()
  File "/home/***/venv/lib/python3.8/site-packages/playwright/_impl/_browser_context.py", line 122, in new_page
    return from_channel(await self._channel.send("newPage"))
  File "/home/***/venv/lib/python3.8/site-packages/playwright/_impl/_connection.py", line 36, in send
    return await self.inner_send(method, params, False)
  File "/home/***/venv/lib/python3.8/site-packages/playwright/_impl/_connection.py", line 47, in inner_send
    result = await callback.future
playwright._impl._api_types.Error: Protocol error (Runtime.enable): The page has been closed.
```

If I replace webkit by chromium, it work.
Do I need to install some extra dependencies or something ?

mxschmitt:
Refactoring the logic which was broken since #5820

This means:
- channels? -> the channels which are in the browsers.json should be validated. The other browsers not.
- executablePath given? no validation, the user knows what he is doing
- no executable path given? validation, default behavior

Fixes #6226

### DIFF
--- a/src/utils/registry.ts
+++ b/src/utils/registry.ts
@@ -292,21 +292,25 @@ export class Registry {
 
   linuxLddDirectories(browserName: BrowserName): string[] {
     const browserDirectory = this.browserDirectory(browserName);
-    if (browserName === 'chromium')
-      return [path.join(browserDirectory, 'chrome-linux')];
-    if (browserName === 'firefox')
-      return [path.join(browserDirectory, 'firefox')];
-    if (browserName === 'webkit') {
-      return [
-        path.join(browserDirectory, 'minibrowser-gtk'),
-        path.join(browserDirectory, 'minibrowser-gtk', 'bin'),
-        path.join(browserDirectory, 'minibrowser-gtk', 'lib'),
-        path.join(browserDirectory, 'minibrowser-wpe'),
-        path.join(browserDirectory, 'minibrowser-wpe', 'bin'),
-        path.join(browserDirectory, 'minibrowser-wpe', 'lib'),
-      ];
+    switch (browserName) {
+      case 'chromium':
+        return [path.join(browserDirectory, 'chrome-linux')];
+      case 'webkit':
+      case 'webkit-technology-preview':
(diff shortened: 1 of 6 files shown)


aslushnikov:
> channels? -> no validation, e.g. apt/yum should install all dependencies

Channels come from apt **or** from the `npx playwright install` command, e.g. `npx playwright install webkit-technology-preview`. So some of the channels do make sense to validate!

pavelfeldman (approved):

src/utils/registry.ts
```diff
@@ -303,6 +303,10 @@ export class Registry {
     return tokens ? path.join(browserDirectory, ...tokens) : undefined;
   }
 
+  exists(browserName: BrowserName): boolean {
```
aslushnikov, line 306: this is equal to the `shouldRetain` method below - let's just use it.

As a side note, `shouldRetain` is a pretty vague name - could you please rename it to something like `isSupportedBrowser(browserName)`?

aslushnikov (approved):

PR #6242, opened 2021-04-20
fix(webkit): swallow requests from detached frames

### LINKED ISSUE #6144: [BUG] UnhandledPromiseRejectionWarning in CI logs
**Context:**
- Playwright Version: v1.10.0
- Operating System: Mac
- Node.js version: v15.4.0
- Browser: Webkit
- Extra: Docker image - v1.10.0-focal

**Describe the bug**
Noticed the following logs in Jenkins console output. These logs are only present when a specific test fails and only on webkit browser. 
```
(node:20) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 22)
(node:20) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'attribution' of null
    at new SdkObject (/app/node_modules/playwright/lib/server/instrumentation.js:24:40)
    at new Request (/app/node_modules/playwright/lib/server/network.js:83:9)
    at new WKInterceptableRequest (/app/node_modules/playwright/lib/server/webkit/wkInterceptableRequest.js:69:24)
    at WKPage._onRequestWillBeSent (/app/node_modules/playwright/lib/server/webkit/wkPage.js:823:25)
    at WKSession.<anonymous> (/app/node_modules/playwright/lib/server/webkit/wkPage.js:343:100)
    at WKSession.emit (events.js:315:20)
    at /app/node_modules/playwright/lib/server/webkit/wkConnection.js:139:47
    at runNextTicks (internal/process/task_queues.js:58:5)
    at processImmediate (internal/timers.js:434:9)
```
The playwright logs for test failure are:
Build 1 - page.click: Evaluation failed: Protocol error: Execution context was destroyed, most likely because of a navigation.
Build 2 - page.click: waitForFunction failed: frame got detached.

JoelEinbinder:
I tried to make a test to repro this by sending hundreds of requests from a frame with redirects, and while they were inflight removing the frame, but I couldn't get it to happen. However the user stack trace is pretty strong evidence that sometimes network requests come in for frames that have been detached. If its a redirect, I get the frame from the first request. Otherwise the request is silently swallowed instead of crashing with unhandled promise rejection.

Fixes #6144.

### DIFF
--- a/src/server/webkit/wkPage.ts
+++ b/src/server/webkit/wkPage.ts
@@ -924,8 +924,10 @@ export class WKPage implements PageDelegate {
 
   _onRequestIntercepted(event: Protocol.Network.requestInterceptedPayload) {
     const request = this._requestIdToRequest.get(event.requestId);
-    if (!request)
+    if (!request) {
+      this._session.sendMayFail('Network.interceptRequestWithError', {errorType: 'Cancellation', requestId: event.requestId});
       return;
+    }
     if (!request._allowInterception) {
       // Intercepted, although we do not intend to allow interception.
       // Just continue.
--- a/src/server/webkit/wkPage.ts
+++ b/src/server/webkit/wkPage.ts
@@ -898,7 +898,12 @@ export class WKPage implements PageDelegate {
         redirectedFrom = request.request;
       }
     }
-    const frame = this._page._frameManager.frame(event.frameId)!;
+    const frame = redirectedFrom ? redirectedFrom.frame() : this._page._frameManager.frame(event.frameId);
+    // sometimes we get stray network events for detached frames
+    // TODO(einbinder) why?
+    if (!frame)
+      return;
+
     // TODO(einbinder) this will fail if we are an XHR document request
     const isNavigationRequest = event.type === 'Document';
     const documentId = isNavigationRequest ? event.loaderId : undefined;


src/server/webkit/wkPage.ts
```diff
@@ -898,7 +898,12 @@ export class WKPage implements PageDelegate {
         redirectedFrom = request.request;
       }
     }
-    const frame = this._page._frameManager.frame(event.frameId)!;
+    const frame = redirectedFrom ? redirectedFrom.frame() : this._page._frameManager.frame(event.frameId);
+    // sometimes we get stray network events for detached frames
+    // TODO(einbinder) why?
+    if (!frame)
+      return;
```
dgozman, line 905: Let's also update line 928 to continue all requests that we did not find in the map, to avoid stalling requests that did not have frameId for some reason.

JoelEinbinder: I canceled them instead of continuing, to avoid accidentally purchasing items via requests that were supposed to be intercepted.

dgozman (approved):

PR #6250, opened 2021-04-20
chore: push dispatcher guid into object, reuse it in trace

pavelfeldman:


src/server/console.ts
```diff
@@ -14,16 +14,18 @@
...
-export class ConsoleMessage {
+export class ConsoleMessage extends SdkObject {
   private _type: string;
   private _text?: string;
   private _args: js.JSHandle[];
   private _location: ConsoleMessageLocation;
 
-  constructor(type: string, text: string | undefined, args: js.JSHandle[], location?: ConsoleMessageLocation) {
+  constructor(parent: SdkObject, type: string, text: string | undefined, args: js.JSHandle[], location?: ConsoleMessageLocation) {
+    super(parent, 'console-message');
```
dgozman, line 28: IIRC, we used `TitleCase@guid` before.

src/server/snapshot/snapshotTypes.ts
```diff
@@ -60,6 +60,7 @@ export type FrameSnapshot = {
   html: NodeSnapshot,
   resourceOverrides: ResourceOverride[],
   viewport: { width: number, height: number },
+  isTop: boolean,
```
dgozman, line 63: Why not `isMainFrame`? We have `page.mainFrame()` in all our apis.

dgozman (approved):

PR #6259, opened 2021-04-21
chore: remove stack from WaitForEventInfo

yury-s:
It was only passed in phase 'before' and was done synchronously so the stack is identical to the one in request metadata.


pavelfeldman:
Don't forget the call site!

pavelfeldman (approved):

PR #6268, opened 2021-04-22
chore(dotnet): enable arg argument for addInitScript

avodovnik:


docs/src/api/class-browsercontext.md
```diff
@@ -227,7 +227,7 @@ Script to be evaluated in all pages in the browser context.
 Script to be evaluated in all pages in the browser context.
 
 ### param: BrowserContext.addInitScript.arg
-* langs: js
```
yury-s, line 230: We dropped the argument in python and java, it only exists in javascript. Let's ditch it in C# too? It is always possible to workaround like this: `"(" + script + ")(" + toJson(arg) + ")"`

pavelfeldman:
As per above, closing it.

Issue #6267, opened 2021-04-22
[BUG] keyboard.type doesn't raise all events


hshristov:
**Context:**
- Playwright Version: 1.9.2
- Operating System: Windows
- Node.js version: 12.20.1.0
- Browser: Chromium [probably All]
- Extra: I'm using PlaywrightSharp 0.192.0 with the embedded Chromium and Node.js but my assuption is that this issue comes from Playwright

System:
    OS: Windows 10 10.0.19042
    Memory: 15.84 GB / 31.93 GB
Binaries:
    Node: 14.15.4 - C:\Program Files\nodejs\node.EXE
    npm: 6.14.11 - C:\Program Files\nodejs\npm.CMD

**Describe the bug**

When calling `await page.keyboard.type('한');` only `input` event is fired. According to documentation `keydown, keypress/input, and keyup` should be fired as well.

JoelEinbinder:
For characters that are not type-able on a US keyboard, we only send `input` events.

Is this just confusion with the docs, or do you need to test a user typing `한` on a Korean keyboard?

hshristov:
Got it. I don't have issues with specific site. It is the framework that I'm working on that I want to be able to raise all keyboard events. Any tips on how to implement it on my own are really appreciated. Thanks.

PR #6273, opened 2021-04-22
docs(keyboard): clarify how page.type works for non-US characters

JoelEinbinder:
#6267


mxschmitt (commented):
why only US keyboard? On a German keyboard there is e.g. öäüß which won't result in a keydown/up, a user might not expect that.

JoelEinbinder:
We have the US keyboard mapped in `usKeyboardLayout.ts`. Mapping other keyboards and having a `keyboard.selectLayout` method is doable, there just hasn't been many people asking for it.

JoelEinbinder:
If you are interested in adding a bunch more keyboard layouts, you can look at x11.  `/usr/share/X11/xkb/symbols/` on Linux.

mxschmitt (approved):

PR #6276, opened 2021-04-22
fix(connect): handle disconnect in various situations

dgozman:
There are a few ways for `connect()` to finish:
- `Browser.close()` from the client side.
- Browser on the server side did exit (e.g. crashed).
- Connection was dropped by either of the sides.

We reduce all the cases to the last one by dropping the
connection when client wants calls `Browser.close()` or
server-side browser exits.

In all these cases we should properly cleanup on the server side,
and ensure that all promises reject on the client side.

### DIFF
--- a/src/client/browser.ts
+++ b/src/client/browser.ts
@@ -28,7 +28,7 @@ export class Browser extends ChannelOwner<channels.BrowserChannel, channels.Brow
   readonly _contexts = new Set<BrowserContext>();
   private _isConnected = true;
   private _closedPromise: Promise<void>;
-  _isRemote = false;
+  _remoteType: 'owns-connection' | 'uses-connection' | null = null;
   readonly _name: string;
 
   static from(browser: channels.BrowserChannel): Browser {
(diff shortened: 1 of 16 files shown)


src/client/browser.ts
```diff
@@ -100,7 +101,10 @@ export class Browser extends ChannelOwner<channels.BrowserChannel, channels.Brow
   async close(): Promise<void> {
     try {
       await this._wrapApiCall('browser.close', async (channel: channels.BrowserChannel) => {
-        await channel.close();
+        if (this._remoteClose)
+          this._remoteClose();
```
pavelfeldman, line 105: This should probably be channel._connection.close();

src/client/browserType.ts
```diff
@@ -160,6 +156,7 @@ export class BrowserType extends ChannelOwner<channels.BrowserTypeChannel, chann
           const browser = Browser.from(playwright._initializer.preLaunchedBrowser!);
           browser._logger = logger;
           browser._isRemote = true;
+          browser._remoteClose = () => ws.close();
```
pavelfeldman, line 159: Let's not skip the layers and do this on the transport (via connection.close())

pavelfeldman (approved):

mxschmitt (approved):

PR #6289, opened 2021-04-23
docs: inline parsing should honor template location

kblok:
Having this doc:

```
### option: Browser.newContext.-inline- = %%-shared-context-params-list-%%

### option: Browser.newContext.proxy = %%-context-option-proxy-%%

### option: Browser.newContext.storageState = %%-js-python-context-option-storage-state-%%

### option: Browser.newContext.storageState = %%-csharp-java-context-option-storage-state-%%

### option: Browser.newContext.storageStatePath = %%-csharp-java-context-option-storage-state-path-%%
```

The `inline` elements were being pushed back in the list instead of honoring the position of the template in the main doc.

### DIFF
--- a/utils/doclint/api_parser.js
+++ b/utils/doclint/api_parser.js
@@ -228,6 +228,7 @@ function applyTemplates(body, params) {
     if (node.text && node.text.includes('-inline- = %%')) {
       const [name, key] = node.text.split('-inline- = ');
       const list = paramsMap.get(key);
+      const newChildren = [];
       if (!list)
         throw new Error('Bad template: ' + key);
       for (const prop of list.children) {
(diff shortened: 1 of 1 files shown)


avodovnik (commented):
Do we expect many changes from this? Have you checked against the current baseline which files get modified?

kblok:
Not many changes, @avodovnik. Only in `NewContextAsync` and `NewPageAsync`. You can take a look at it here https://github.com/microsoft/playwright-sharp/pull/1316/files#diff-830df37f7e83748f967f8fcbce73a543c3525c745e1f80f998db27992c322a92R184 

@yury-s, do you think this could break Java?

pavelfeldman:
As long as our -inline- are only for the options, it should not break Java/Python!

pavelfeldman (approved):

Issue #6286, opened 2021-04-23
[BUG]ZIP file containing binaries will not be downloaded complete with WebKit


RFC2109:
**Context:**
- Playwright Version: playwright-internal@1.11.0-next
- Operating System: Windows
- Node.js version: v14.15.1
- Browser: Chromium, Firefox, WebKit
- Extra: [any specific details about your environment]

**Code Snippet**

```javascript
 it.only('should download binary.zip complete', async ({browser, server, browserName}) => {
    const page = await browser.newPage({ acceptDownloads: true });
    await page.goto(server.PREFIX + '/empty.html');
    await page.setContent(`<a href="${server.PREFIX}/binary.zip" download="binary.zip">download</a>`);
    const [ download ] = await Promise.all([
      page.waitForEvent('download'),
      page.click('a')
    ]);
    const downloadPath = await download.path();
    const downloadFileSize = fs.statSync(downloadPath).size;
    const originalFileSize = fs.statSync(path.join(__dirname, 'assets','binary.zip')).size;
    expect(downloadFileSize).toEqual(originalFileSize);
    await page.close();
  });
```
You can generate the zip file with the following command
```sh
head -c 1m </dev/urandom >binary.dat
zip -q  binary.zip binary.dat
```

**Describe the bug**

the result
```
npm run test

> playwright-internal@1.11.0-next test E:\projects\playwright
> folio --config=tests/config/default.config.ts


Running 3 tests using 2 workers
  1) download.spec.ts:65:6 › [webkit] download event should download binary.zip complete ===========

    Error: expect(received).toEqual(expected) // deep equality

    Expected: 1048906
    Received: 524

      74 |     const downloadFileSize = fs.statSync(downloadPath).size;
      75 |     const originalFileSize = fs.statSync(path.join(__dirname, 'assets','binary.zip')).size;
    > 76 |     expect(downloadFileSize).toEqual(originalFileSize);
         |                              ^
      77 |     await page.close();
      78 |   });
      79 | 

        at Spec.fn (E:\projects\playwright\tests\download.spec.ts:76:30)
        at WorkerRunner._runTestWithBeforeHooks (E:\projects\playwright\node_modules\folio\out\workerRunner.js:299:13)

  2 passed (33s)
  1 failed      
    download.spec.ts:65:6 › [webkit] download event should download binary.zip complete ============

  Slow test: download.spec.ts (19s)
```

mxschmitt:
[...]

PR #6312, opened 2021-04-24
test: added test for downloading zip file

mxschmitt:
test for #6286 to see if it passes on the bots.


mxschmitt:
Verified that #6345 fixes it.

PR #6315, opened 2021-04-25
docs(python): add Error base class

mxschmitt:


docs/src/api/class-error.md
```diff
@@ -0,0 +1,26 @@
+# class: Error
```
yury-s, line 1: shouldn't it be more specific like PlaywrightError so that it's easier to tell what kind of error is being caught?

mxschmitt: it should be that would result in a breaking change on Playwright Python side.

yury-s (approved):

docs/src/api/class-error.md
kumaraditya303: @yury-s https://github.com/microsoft/playwright-python/issues/653

PR #6337, opened 2021-04-27
chore(tracing): fix some of the start/stop scenarios

pavelfeldman:


src/server/snapshot/snapshotterInjected.ts
```diff
@@ -432,10 +432,12 @@ export function frameSnapshotStreamer(snapshotStreamer: string, snapshotBinding:
       };
 
       let allOverridesAreRefs = true;
-      for (const sheet of this._allStyleSheetsWithUrlOverride) {
+      for (const sheet of this._staleStyleSheets) {
```
dgozman, line 435: I think we need to send all stylesheets that ever had an override every time, even when they are not stale (in this case it will be a number ref). Otherwise, the renderer will use the non-modified resource for this snapshot, instead of the last modified one.

src/server/snapshot/snapshotterInjected.ts
```diff
@@ -162,29 +162,29 @@ export function frameSnapshotStreamer(snapshotStreamer: string, snapshotBinding:
...
-      this._streamSnapshot(snapshotName);
-    }
+    reset() {
+      this._staleStyleSheets.clear();
 
-    setSnapshotInterval(interval: number) {
-      this._interval = interval;
-      if (interval)
-        this._streamSnapshot();
+      const visitNode = (node: Node | ShadowRoot) => {
```
dgozman, line 168: In theory, we can replace the cache with WeakMap and just clear it here. Up to you.

src/server/frames.ts
```diff
@@ -197,10 +198,11 @@ export class FrameManager {
 
     frame._onClearLifecycle();
     const navigationEvent: NavigationEvent = { url, name, newDocument: frame._currentDocument };
-    frame.emit(Frame.Events.Navigation, navigationEvent);
+    this._page.frameNavigatedToNewDocument(frame);
```
dgozman, line 201: Let's undo this line 😄

src/server/frames.ts
```diff
@@ -197,10 +198,11 @@ export class FrameManager {
...
+    this._responses.length = 0;
     if (!initial) {
       debugLogger.log('api', `  navigated to "${url}"`);
-      this._page.frameNavigatedToNewDocument(frame);
+      frame.emit(Frame.Events.Navigation, navigationEvent);
```
dgozman, line 205: And this one 😄

dgozman (approved):

src/server/snapshot/snapshotterInjected.ts
pavelfeldman: I won't claim I understand this part, but I don't change behavior here. Maybe you can follow up?

PR #6341, opened 2021-04-27
Changes to add websocket tunnel support in the playwright server and the client code

Sunil1821:
This pr contains changes to provide websocket tunnel support for the playwright server and the client code. Also tests have been added to test different protocol like websocket, https, http long polling.

### DIFF
--- a/src/remote/playwrightClient.ts
+++ b/src/remote/playwrightClient.ts
@@ -14,72 +14,276 @@
  * limitations under the License.
  */
 
-import WebSocket from 'ws';
-import { Connection } from '../client/connection';
-import { Playwright } from '../client/playwright';
-// const http = require('http');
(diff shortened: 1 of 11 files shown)


src/remote/playwrightClient.ts
```diff
@@ -14,42 +14,276 @@
...
-      closePromise
-    ]);
-    return new PlaywrightClient(playwright as Playwright, ws);
+  constructor(wsEndpoint: string, ports?: [number]) {
+    this._portsOpened = false;
+    this._wsEndpoint = wsEndpoint;
+    this._closePromise = new Promise((f) => this.ws.on("close", f));
+    this._tcpConnections = {};
+    this.addHandlers();
+    this._ports = ports;
```
puagarwa, line 57: we need to probably add some validation to disallow some restricted ports.

PR #1166, opened 2020-02-29
types: better types

JoelEinbinder:
#6

A place for discussion

PR #6344, opened 2021-04-28
fix(types): fix waitForSelector typing to not union null when appropriate

jharwig:
I found an issue with the `types.d.ts` file regarding `waitForSelector` when passing some options, but no state. When no `state` key is provided, the return type would include null. This is despite the default state value is `visible` and shouldn't allow `null`.

I discovered when doing something like below. The example shows passing no options gives the correct return type, but not when options are passed.
Screen Shot 2021-04-27 at 7 08 06 PM
<details>
  <summary>Source Code</summary>

```ts
  const el = await page.waitForSelector(tid('document-viewer'), {
    timeout: docLoadTimeout, // no state passed, should default to visible
  })
  await el.click() // Object is possible 'null'

  const elNoOptions = await page.waitForSelector(tid('document-viewer'))
  await elNoOptions.click()
```
</details>

I found the solution, added some tests, but I also fixed the `canBeNull` assertion because my new test case wouldn't fail the tests as they were (even though the types clearly included null, but the assertion was missing it.) Looking at some [code review comments this might have been the case for a while](https://github.com/microsoft/playwright/pull/1166#discussion_r391294546). If anyone has some insight, I'd love to hear it.

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -21,10 +21,10 @@ import { Readable } from 'stream';
 import { Serializable, EvaluationArgument, PageFunction, PageFunctionOn, SmartHandle, ElementHandleForTag, BindingSource } from './structs';
 
 type PageWaitForSelectorOptionsNotHidden = PageWaitForSelectorOptions & {
-  state: 'visible'|'attached';
+  state?: 'visible'|'attached';
 };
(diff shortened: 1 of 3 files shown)


JoelEinbinder (approved):
Great patch. Thanks!

PR #6345, opened 2021-04-28
feat(webkit): bump to 1468

yury-s:
#6286

### DIFF
--- a/browsers.json
+++ b/browsers.json
@@ -18,7 +18,7 @@
     },
     {
       "name": "webkit",
-      "revision": "1467",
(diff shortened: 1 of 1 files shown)


mxschmitt:
Verified that #6312 passes now. Closed its PR since it was not a good test, shall we add one so it won't regress or is it unlikely?

yury-s:
> Verified that #6312 passes now. Closed its PR since it was not a good test, shall we add one so it won't regress or is it unlikely?

We should certainly add a test. I thought you were going to land your PR.

PR #6346, opened 2021-04-28
test(screencast): add a test with multiple colors

JoelEinbinder:
Adds a test that uses a wonderful company logo to make sure the screencast has the right pixels in the right places.

This test fails for me with firefox on at least linux and windows.

### DIFF
--- a/tests/screencast.spec.ts
+++ b/tests/screencast.spec.ts
@@ -93,6 +93,27 @@ function almostRed(r, g, b, alpha) {
   expect(alpha).toBe(255);
 }
 
+function almostBlue(r, g, b, alpha) {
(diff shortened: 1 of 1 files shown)


JoelEinbinder:
I bisected down to https://github.com/microsoft/playwright/pull/3720 as the breaking change, at least on linux.

JoelEinbinder:
The patch in question scales the video if the video width is less than the content window width. Firefox desktop has a min-width of 450px, which the test doesn't expect. So the 320 viewport it tries to set is actually 450.

JoelEinbinder:
This test is currently failing headful chromium linux.

JoelEinbinder:
@pavelfeldman do you still want this test?

PR #6348, opened 2021-04-28
fix: deal with delayed _nodeElectronHandle

msyfls123:
Since `process.mainModule.require('electron')` maybe be slow, we could possibly get undefined `_nodeElectronHandle`, so just treat it as a Promise.
image

### DIFF
--- a/src/server/electron/electron.ts
+++ b/src/server/electron/electron.ts
@@ -58,7 +58,8 @@ export class ElectronApplication extends SdkObject {
   private _nodeConnection: CRConnection;
   private _nodeSession: CRSession;
   private _nodeExecutionContext: js.ExecutionContext | undefined;
-  _nodeElectronHandle: js.JSHandle<any> | undefined;
+  _nodeElectronHandlePromised: Promise<js.JSHandle<any>>;
+  private _resolveNodeElectronHandle!: (handle: js.JSHandle<any>) => void;
   private _windows = new Set<ElectronPage>();
   private _lastWindowId = 0;
   readonly _timeoutSettings = new TimeoutSettings();
@@ -73,6 +74,7 @@ export class ElectronApplication extends SdkObject {
     this._browserContext.on(BrowserContext.Events.Page, event => this._onPage(event));
     this._nodeConnection = nodeConnection;
     this._nodeSession = nodeConnection.rootSession;
+    this._nodeElectronHandlePromised = new Promise(resolve => this._resolveNodeElectronHandle = resolve);
   }
 
   private async _onPage(page: ElectronPage) {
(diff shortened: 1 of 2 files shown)


src/server/electron/electron.ts
```diff
@@ -87,7 +87,7 @@ export class ElectronApplication extends SdkObject {
     this._windows.add(page);
 
     // Below is async.
-    const handle = await this._nodeElectronHandle!.evaluateHandle(({ BrowserWindow }, windowId) => BrowserWindow.fromId(windowId), windowId).catch(e => {});
+    const handle = await (await this._nodeElectronHandle)!.evaluateHandle(({ BrowserWindow }, windowId) => BrowserWindow.fromId(windowId), windowId).catch(e => {});
```
pavelfeldman, line 90: In case of mentioned race, this._nodeElectronHandle can still be undefined here.

src/server/electron/electron.ts
```diff
@@ -58,7 +58,7 @@ export class ElectronApplication extends SdkObject {
   private _nodeConnection: CRConnection;
   private _nodeSession: CRSession;
   private _nodeExecutionContext: js.ExecutionContext | undefined;
-  _nodeElectronHandle: js.JSHandle<any> | undefined;
+  _nodeElectronHandle: Promise<js.JSHandle<any> | undefined>;
```
pavelfeldman, line 61: `Promise<js.JSHandle<any>> | undefined;`

pavelfeldman (changes requested):
Thanks for the PR! I can repro the issue, so I'll be able to add a test for it.

src/server/electron/electron.ts
msyfls123: I'm not very sure whether '_nodeElectronHandle' will be reached before '_init' has finished or not, I just find '_onPage' and 'close' in server part. 

It looks that '_init' will always finish before 'onPage' when I have tested this PR in my application. 

If we wanna ensure '_nodeElectronHandle' to be a Promise, maybe we could extract its initialization to a new method like 'initNodeHandle', and await launch process at top of it?

pavelfeldman: Electron support uses two independent connections, one going to Node and another going to the browser. These connections race naturally and if you throttle the Node one, you can see how browser talks to us before node handle is initialized. You can add `await new Promise(f => setTimeout(f, 1000));` before `await this._nodeSession.send('Runtime.enable', {}).catch(e => {});` in the init to trigger it. You last push should resolve it! Let's see if it fixes my local test and if it does, we merge it. I'll follow up with the test and a small refactoring.

msyfls123: @pavelfeldman Hope it works! I found some eslint issues in my commit, just fix and amend it to the only commit.

pavelfeldman (approved):

PR #6352, opened 2021-04-28
feat(actions): ensure we perform the action on the expected element

dgozman:
This changes previous `checkHitTarget` heurisitc that took place before the action, to `setupHitTargetInterceptor` that works during the action:
- Before the action we set up capturing listeners on the window.
- During the action we ensure that event target is the element we expect to interact with.
- After the action we clear the listeners.

This should catch the "layout shift" issues where things move between action point calculation and the actual action.

Possible issues:
- **Risk:** All events will be visible to the capturing listener on the `Window` object that was added before the click. The immediate mitigation with `addInitScript` and adding a listener once does not work because `document.open()` resets listeners on `window` object.
- **Risk:** `{ trial: true }` might dispatch move events like `mousemove` or `pointerout`, because we do actually move the mouse but prevent all other events.
- **Timing**: The timing of "hit target check" has moved, so this may affect different web pages in different ways, for example expose more races. In this case, we should retry the click as before.
- **No risk**: There is still a possibility of mistargeting with iframes shifting around, because we only intercept in the target frame. This behavior does not change.

### DIFF
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -41,11 +41,16 @@ export type InjectedScriptPoll<T> = {
 export type ElementStateWithoutStable = 'visible' | 'hidden' | 'enabled' | 'disabled' | 'editable' | 'checked';
 export type ElementState = ElementStateWithoutStable | 'stable';
 
+export type HitTargetInterceptionResult = 'error:notconnected' | {
+  stop: () => 'done' | { hitTargetDescription: string };
+};
+
 export class InjectedScript {
   private _enginesV1: Map<string, SelectorEngine>;
   _evaluator: SelectorEvaluatorImpl;
   private _stableRafCount: number;
   private _replaceRafWithTimeout: boolean;
+  private _hitTargetListeners: { listener: (event: Event) => void, event: string }[] = [];
 
   constructor(stableRafCount: number, replaceRafWithTimeout: boolean, customEngines: { name: string, engine: SelectorEngine}[]) {
     this._enginesV1 = new Map();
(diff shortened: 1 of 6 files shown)


tests/tap.spec.ts
```diff
@@ -58,7 +58,7 @@ it('trial run should not tap', async ({}) => {
   await page.tap('#a');
   const eventsHandle = await trackEvents(await page.$('#b'));
   await page.tap('#b', { trial: true });
-  expect(await eventsHandle.jsonValue()).toEqual([]);
+  expect(await eventsHandle.jsonValue()).toEqual(['pointerover', 'pointerenter', 'pointerout', 'pointerleave']);
```
JoelEinbinder, line 61: I don't like the idea that the trail run has user visible events, but at the very least this test needs to be rewritten so that it actually tests something. In its current form it would pass with `{ trial: false }`.

tests/page-mouse.spec.ts
```diff
@@ -111,7 +111,7 @@ it('should trigger hover state', async ({page, server}) => {
 it('should trigger hover state on disabled button', async ({page, server}) => {
   await page.goto(server.PREFIX + '/input/scrollable.html');
   await page.$eval('#button-6', (button: HTMLButtonElement) => button.disabled = true);
-  await page.hover('#button-6', { timeout: 5000 });
+  await page.hover('#button-6', { timeout: 2000 });
```
JoelEinbinder, line 114: this is just a drive by to make this test faster right? It doesn't somehow fail with 5000ms now?

tests/page-click.spec.ts
```diff
@@ -780,3 +780,29 @@ it('should click zero-sized input by label', async ({page}) => {
...
+    blocker.style.width = '400px';
+    blocker.style.height = '400px';
+    blocker.style.left = '0';
+    blocker.style.top = '0';
+    document.body.appendChild(blocker);
+
+    const allEvents = [];
+    (window as any).allEvents = allEvents;
+    for (const name of ['mousedown', 'mouseup', 'click', 'dblclick', 'auxclick', 'contextmenu', 'pointerdown', 'pointerup'])
+      button.addEventListener(name, e => allEvents.push(e.type));
```
JoelEinbinder, line 798: Shouldn't this test verify that no events are fired on the blocker/window. How would events ever be fired on the button?

tests/page-click-timeout-4.spec.ts
```diff
@@ -29,3 +29,19 @@ it('should timeout waiting for stable position', async ({page, server}) => {
   expect(error.message).toContain('waiting for element to be visible, enabled and stable');
   expect(error.message).toContain('element is not stable - waiting');
 });
+
+it('should click for the second time after first timeout', async ({page, server, mode}) => {
+  it.skip(mode !== 'default');
+
+  await page.goto(server.PREFIX + '/input/button.html');
+  const __testHookBeforePointerAction = () => new Promise(f => setTimeout(f, 2000));
```
JoelEinbinder, line 37:
```suggestion
  const __testHookBeforePointerAction = () => new Promise(f => setTimeout(f, 1500));
```

This is supposed to test that a timing out click does not actually click right? If it was broken, it would be racing against the page.click on line 45, because they both happen at 2 seconds.

tests/tap.spec.ts
dgozman: Not really, with `{trial:false}` we'll get a lot more events - that's what the previous test does.

tests/page-mouse.spec.ts
dgozman: Yep.

tests/page-click-timeout-4.spec.ts
dgozman: Great catch!

tests/page-click.spec.ts
dgozman: Indeed. Changed as you suggest.

src/server/injected/injectedScript.ts
```diff
@@ -594,36 +599,79 @@ export class InjectedScript {
...
+          if (rootHitTargetDescription)
+            result = { hitTargetDescription: `${hitTargetDescription} from ${rootHitTargetDescription} subtree` };
+          else
+            result = { hitTargetDescription };
+        }
       }
-      element = parentElementOrShadowHost(element);
+      if (blockAllEvents || result !== 'done') {
+        event.preventDefault();
+        event.stopPropagation();
```
JoelEinbinder, line 659:
```suggestion
        event.stopImmediatePropagation();
```

src/server/injected/injectedScript.ts
```diff
@@ -594,36 +599,79 @@ export class InjectedScript {
...
+    };
+
+    const stop = () => {
+      this._removeHitTargetListeners();
+      return result!;
+    };
+
+    this._removeHitTargetListeners();
+    for (const eventName of events) {
+      window.addEventListener(eventName, listener, { capture: true });
```
JoelEinbinder, line 670: Event listeners are fired in the order they were added so I believe these should be added in evaluateOnNewDocument and then just enabled as needed here. Otherwise an existing window listener will get the bad events. Please cover with a test.

PR #6361, opened 2021-04-29
browser(firefox): don't record video outside the viewport

JoelEinbinder:
The browser window has a min width of 450px. If the viewport was smaller than that, the video would try to capture outside the viewport. Passes the test in #6346 on linux. I have not tested on other operating systems yet.


JoelEinbinder:
tbr @aslushnikov

PR #6370, opened 2021-04-30
feat(network): expose network events via browser context

### LINKED ISSUE #6340: [Feature] Capture network requests on BrowserContext
Currently when page issues requests, we can use the `request` event on the page to capture the network request and get the timing information. But if the page opens a new popup or window, the network request information for those requests gets lost 

```html

<p><a href="popup2.html" target="_blank" id="linkWithPopupNoNewWindow">Click</a>here (id=linkWithPopupNoNewWindow) to open - Test Page - new tab</p>
```

```ts

// will not log the requests to the popup page and also the resources in page1. 
page.on("request", req => console.log("Request", request.url());

await page.goto("index.html"); // goes to index page

const [page1] = await Promise.all([
      context.waitForEvent('page')
      page.click('#linkWithPopupNoNewWindow'),
]);

await page1.close();
```

From the above code, we would miss out the logging for both the request to the `popup2.html` page and also the resources inside that page. However, the resources inside `page1` can be logged by again listening to `page1.on('request')`, but still there is no way to get the request activity to the popup in new tab. 

Is there a way to get all the network activity logged in the context of the browser level CDP session instead of creating CDP session for every page object.

vigneshshanmugam:
+ fix #6340 
+ Exposes all the network related events (request, response, requestfailed, requestfinished) through the browser context to allow for managing network activity even if the is any navigations through popups or to new tabs which could result in creation of multiple page objects.


vigneshshanmugam:
@dgozman WDYT of the above approach? Can i go ahead and expose all `network` events (`requestfailed`, `requestfinishes`, `response`) if the approach feels good enough?

src/protocol/channels.ts
```diff
@@ -584,12 +584,13 @@ export type BrowserContextInitializer = {
...
 export interface BrowserContextChannel extends Channel {
   on(event: 'bindingCall', callback: (params: BrowserContextBindingCallEvent) => void): this;
-  on(event: 'close', callback: (params: BrowserContextCloseEvent) => void): this;
+  on(event: 'close', scallback: (params: BrowserContextCloseEvent) => void): this;
   on(event: 'page', callback: (params: BrowserContextPageEvent) => void): this;
   on(event: 'route', callback: (params: BrowserContextRouteEvent) => void): this;
   on(event: 'video', callback: (params: BrowserContextVideoEvent) => void): this;
   on(event: 'backgroundPage', callback: (params: BrowserContextBackgroundPageEvent) => void): this;
   on(event: 'serviceWorker', callback: (params: BrowserContextServiceWorkerEvent) => void): this;
+  on(event: 'request', callback: (params: PageRequestEvent) => void): this;
```
dgozman, line 593: Instead of manually changing this file, update the `protocol.yml` with the new event. If you run `npm run watch` to compile code in the background, it should regenerate this file for you.

tests/browsercontext-network-event.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { browserTest as it, expect } from './config/browserTest';
+
+it('BrowserContext.Events.Request', async ({browser, server}) => {
```
dgozman, line 20: Let's add a test for the intended usecase with the popup and its main html request.

dgozman (commented):
> Can i go ahead and expose all network events (requestfailed, requestfinishes, response) if the approach feels good enough?

Good idea, let's do that for all events altogether.

Note that we'll have to change the documentation as well. Take a look at `event: Page.request` in `docs/src/api/class-page.md` - we'll need to copy documentation to the `class-browsercontext.md`.

tests/browsercontext-network-event.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  const context = await browser.newContext();
+  const page = await context.newPage();
+  const requests = [];
+  context.on('request', request => requests.push(request));
+  await page.goto(server.EMPTY_PAGE);
+  await page.setContent('<a target=_blank rel=noopener href="/one-style.html">yo</a>');
+  const [popup] = await Promise.all([
+    page.waitForEvent('popup'),
```
vigneshshanmugam, line 28: @dgozman For some reason in chromium alone, I am not able to get the event for popup navigation request alone `one-style.html`. Firefox and webkit reports it correctly. Any clue?

dgozman: Not really. Try running with `DEBUG=pw:protocol` and see whether it comes over CDP.

src/server/browserContext.ts
```diff
@@ -299,6 +303,26 @@ export abstract class BrowserContext extends SdkObject {
     await this._closePromise;
   }
 
+  requestStarted(request: network.Request) {
```
dgozman, line 306: Sorry for missing it earlier, but instead of these methods, let's emit browser context  events from the FrameManager, in addition to page events. That should be easier.

src/client/browserContext.ts
```diff
@@ -96,6 +100,19 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel,
       page._opener.emit(Events.Page.Popup, page);
   }
 
+  private _onRequestFailed(request: network.Request, responseEndTiming: number, failureText: string | undefined) {
```
dgozman, line 103: Now looking at these methods, it seems like we should do the logic in them just once (not once in Page and once in BrowserContext).

That means we should remove the corresponding events from the Page interface (in protocol.yml), add the `page: Page?` parameter to them, and emit events here both on the context and the page if any.

To pass the right `Page` in `BrowserContextDispatcher`, we can reach for it as `request.frame().page()`. However, that page might not be initialized yet, so we should only send it once it is initialized, and `reportAsNew` method could mark the page as "initialized".

Sorry that this turns out to be much more complicated than anticipated. If you don't feel like working too much on this, I can pick it up 😄

docs/src/api/class-browsercontext.md
```diff
@@ -155,6 +155,35 @@ await browser_context.add_cookies([cookie_object1, cookie_object2])
 browser_context.add_cookies([cookie_object1, cookie_object2])
 ```
 
+## event: BrowserContext.request
+- argument: <[Request]>
+
+Emitted when a request is issued from any pages created through this context.
+The [request] object is read-only. In order to intercept and mutate requests,
+see [`method: BrowserContext.route`] or [`method: page.route`].
```
dgozman, line 163: typo: `page.route` -> `Page.route`.

docs/src/api/class-browsercontext.md
```diff
@@ -155,6 +155,35 @@ await browser_context.add_cookies([cookie_object1, cookie_object2])
...
+
+## event: BrowserContext.requestFailed
+- argument: <[Request]>
+
+Emitted when a request fails, for example by timing out.
```
dgozman, line 168: Let's also reference corresponding page events from these ones, and vice versa. Something like `to only listen for requests from a particular page, use [...page.requestFailed reference...]`.

src/server/browserContext.ts
vigneshshanmugam: Done

src/client/browserContext.ts
vigneshshanmugam: Ah I see, i have cleaned up the code a bit, I hope I did the right thing as per your suggestion, would like you to take a second pass again.

src/client/browserContext.ts
```diff
@@ -96,6 +100,47 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel,
       page._opener.emit(Events.Page.Popup, page);
   }
 
+  private _onRequest(request: network.Request, pageChannel: channels.PageChannel | undefined) {
+    this.emit(Events.BrowserContext.Request, request);
+    if (pageChannel) {
+      const page = Page.from(pageChannel);
+      if (page._opener && !page._opener.isClosed())
```
dgozman, line 107: Drop this condition, it is relevant for popups, not for requests.

src/client/browserContext.ts
```diff
@@ -86,6 +86,10 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel,
       this._serviceWorkers.add(serviceWorker);
       this.emit(Events.BrowserContext.ServiceWorker, serviceWorker);
     });
+    this._channel.on('request', ({ request, page }) => this._onRequest(network.Request.from(request), page));
```
dgozman, line 89: I'd do `Page.fromNullable(page)` right here and pass `Page | null` to the function.

src/server/frames.ts
```diff
@@ -263,20 +263,21 @@ export class FrameManager {
...
+    this._page._browserContext.emit(BrowserContext.Events.Request, request);
     this._page._requestStarted(request);
   }
 
   requestReceivedResponse(response: network.Response) {
     if (response.request()._isFavicon)
       return;
     this._responses.push(response);
-    this._page.emit(Page.Events.Response, response);
+    this._page._browserContext.emit(BrowserContext.Events.Response, response);
```
dgozman, line 274: I think the fact that you've dropped the `Page.Events.Response` and similar ones might break some internal listeners, e.g. `HarTracer`. I hope that tests should point that 😄

src/dispatchers/browserContextDispatcher.ts
```diff
@@ -63,6 +64,25 @@ export class BrowserContextDispatcher extends Dispatcher<BrowserContext, channel
         this._dispatchEvent('serviceWorker', { worker: new WorkerDispatcher(this._scope, serviceWorker)});
       context.on(CRBrowserContext.CREvents.ServiceWorker, serviceWorker => this._dispatchEvent('serviceWorker', { worker: new WorkerDispatcher(this._scope, serviceWorker) }));
     }
+    context.on(BrowserContext.Events.Request, (request: Request) => this._dispatchEvent('request', {
+      request: RequestDispatcher.from(this._scope, request),
+      page: request.frame()._page
```
dgozman, line 69: This is the place where `page` might not be initialized yet (as in not finished the setup), so sending it over protocol would be an issue. That's why I suggest to add a method like `page.initializedOrNull()` that will return the page itself after the `reportAsNew` was called.

dgozman (commented):
> Ah I see, i have cleaned up the code a bit, I hope I did the right thing as per your suggestion, would like you to take a second pass again.

Almost there, thank you!

src/server/frames.ts
vigneshshanmugam: Do we need to emit events on page as well? Since we already do emit it for the consuming part on `client/browserclient.ts`?

src/dispatchers/browserContextDispatcher.ts
vigneshshanmugam: Even after calling reportAsNew, it fails to initialize the page. Can you help here?

src/server/frames.ts
vigneshshanmugam: Ahh i get it, its for the server side stuff? I brought them back, but we can move to BrowserContext.Request in another PR to make sure we dont miss our requests with HAR when popups are involved.

src/dispatchers/browserContextDispatcher.ts
```diff
@@ -63,6 +64,27 @@ export class BrowserContextDispatcher extends Dispatcher<BrowserContext, channel
         this._dispatchEvent('serviceWorker', { worker: new WorkerDispatcher(this._scope, serviceWorker)});
       context.on(CRBrowserContext.CREvents.ServiceWorker, serviceWorker => this._dispatchEvent('serviceWorker', { worker: new WorkerDispatcher(this._scope, serviceWorker) }));
     }
+    context.on(BrowserContext.Events.Request, (request: Request) =>  {
+      return this._dispatchEvent('request', {
+        request: RequestDispatcher.from(this._scope, request),
+        page: request.frame()._page.initializedOrNull()
```
vigneshshanmugam, line 70: @dgozman I tried with `PageDispatcher.fromNullable(scope, request.frame()._page.initializedOrNull())`, dont think its worthing as expected. 

```
   Error:
        at Object.assert (/Users/vigneshh/opensource/playwright/src/utils/utils.ts:67:11)
        at new Dispatcher (/Users/vigneshh/opensource/playwright/src/dispatchers/dispatcher.ts:67:5)
        at new PageDispatcher (/Users/vigneshh/opensource/playwright/src/dispatchers/pageDispatcher.ts:51:5)
        at CRBrowserContext.<anonymous> (/Users/vigneshh/opensource/playwright/src/dispatchers/browserContextDispatcher.ts:53:88)
        at CRBrowserContext.emit (events.js:315:20)
        at Page.reportAsNew (/Users/vigneshh/opensource/playwright/src/server/page.ts:198:26)
        at Page.initializedOrNull (/Users/vigneshh/opensource/playwright/src/server/page.ts:207:10)
        at CRBrowserContext.<anonymous> (/Users/vigneshh/opensource/playwright/src/dispatchers/browserContextDispatcher.ts:70:72)
        at CRBrowserContext.emit (events.js:315:20)
        at FrameManager.requestStarted (/Users/vigneshh/opensource/playwright/src/server/frames.ts:266:32)
```

docs/src/api/class-browsercontext.md
```diff
@@ -155,6 +155,40 @@ await browser_context.add_cookies([cookie_object1, cookie_object2])
 browser_context.add_cookies([cookie_object1, cookie_object2])
 ```
 
+## event: BrowserContext.request
+- argument: <[Request]>
+
+Emitted when a request is issued from any pages created through this context.
+The [request] object is read-only. To only listen for requests from a particular
+page, use [`method: Page.request`].
```
dgozman, line 163: typo: `method: Page.request` -> `event: Page.request`. Similar typos below.

src/server/frames.ts
dgozman: > ... but we can move to BrowserContext.Request in another PR to make sure we dont miss our requests with HAR when popups are involved.

Nice catch!

src/server/page.ts
```diff
@@ -203,6 +203,11 @@ export class Page extends SdkObject {
       this.emit(Page.Events.Close);
   }
 
+  initializedOrNull() {
```
dgozman, line 206: I meant that we should _only_ report the page after it is initialized. To do that:
- Introduce `private _initialized = false;` in the `Page` class.
- At the line 198, right before emitting `BrowserContext.Events.Page`, assign `_initialized` to `true`.
- `return this._initialized ? this : undefined` here.
- I also just realized that we should send `undefined` instead of `null` in this case, so let's call the method `initializedOrUndefined`.

src/dispatchers/browserContextDispatcher.ts
dgozman: `PageDispatcher.fromNullable(scope, request.frame()._page.initializedOrNull())` looks very good, just use `this._scope` instead of `scope`. The error is due to the `reportAsNew` vs `initializedOrNull` - see my comment there.

dgozman (commented):
This is looking fantastic! I think with a small change around `initializedOrNull` this is going to work flawlessly. And your comment about server-side follow up is exactly on point, we'll do that.

src/dispatchers/browserContextDispatcher.ts
vigneshshanmugam: Still the same error, Is there a race condition here? Trying to dispatch to the page but connection isn't established still? https://github.com/microsoft/playwright/blob/dddfbaaeb257ca80e07e4aed3bb6d3f9b90923fd/src/dispatchers/dispatcher.ts#L67

dgozman: I believe the problem is that we call `reportAsNew` twice. Just drop `this.reportAsNew()` from the `initializedOrUndefined` method - that should fix it.

vigneshshanmugam: Ahh nice, totally missed that. Thanks that fixes the issue.

src/server/chromium/crNetworkManager.ts
```diff
@@ -207,10 +207,10 @@ export class CRNetworkManager {
       frame = this._page._frameManager.frame(requestPausedEvent.frameId);
 
     // Check if it's main resource request interception (targetId === main frame id).
-    if (!frame && requestPausedEvent && requestWillBeSentEvent.frameId === (this._page._delegate as CRPage)._targetId) {
```
vigneshshanmugam, line 210: @dgozman For the new popup page created as delegate from main page, the event was not getting fired without this change in chromium. 

Not sure if this is the right way, LMK WDYT.

dgozman: @vigneshshanmugam Good catch. I just traced the protocol myself, and in some cases we are getting the network request from the browser before the frame tree arrives from the renderer. I think this is a good fix!

vigneshshanmugam: Nice, Thanks for double checking 👍

tests/browsercontext-network-event.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    page.click('a'),
+  ]);
+  await popup.close();
+  const urls = requests.map(r => r.url());
+  expect(requests.length).toBe(3);
```
dgozman, line 33: I'd suggest to remove this line. I see this test failing on some bots, and without this line it is easy to see which request is missing, because of the next `expect`.

dgozman (commented):
This looks good, thank you! There are two things missing:
- The new test is failing on some bots (e.g. Chrome Stable Linux).
- Run `npm run lint` to fix lint/documentation errors - this is why the line bot is failing.
Other than that, the implementation looks great!

vigneshshanmugam:
Force pushed to trigger the previously hanging CI builds.

dgozman (approved):
I think this is now great, thank you! I only have two minor nits that will make tests non-flaky.

This PR changes `protocol.yml` so it will require some work next time we roll into other languages (e.g. Playwright for python). Don't worry - this does not block the PR. However, since the next release should happen very soon, let's hold on with landing until v1.11 is cut, to avoid last-minute changes across all languages. I will come back here and land it for you. Thank you again for the great PR!

tests/browsercontext-network-event.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  const page = await context.newPage();
+  const requests = [];
+  context.on('request', request => requests.push(request));
+  await page.goto(server.EMPTY_PAGE);
+  await page.setContent('<a target=_blank rel=noopener href="/one-style.html">yo</a>');
+  const [page1] = await Promise.all([
+    context.waitForEvent('page'),
+    page.click('a'),
+  ]);
+  await page1.waitForEvent('load');
```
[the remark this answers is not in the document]
vigneshshanmugam: done

vigneshshanmugam:
@dgozman Great and thank you for the help throughout the review.

vigneshshanmugam:
@dgozman Is it a good time to get this in? Unclear on why the tests timeouts now.

dgozman:
> @dgozman Is it a good time to get this in? Unclear on why the tests timeouts now.

@vigneshshanmugam Yep, it is good time now. If you could rebase on top of the latest, tests will kick in and pass 😄

dgozman:
@vigneshshanmugam This PR went in. Thank you!

PR #6376, opened 2021-04-30
chore(dotnet): run dotnet format after generation

avodovnik:
The generated files are not properly formatted. This should, at the very least, fix the basic formatting mistakes and problems.

### DIFF
--- a/utils/doclint/generateDotnetApi.js
+++ b/utils/doclint/generateDotnetApi.js
@@ -201,7 +201,7 @@ const customTypeNames = new Map([
       });
     }, enumsDir));
   
-  if (process.argv[3] !== "--skip") {
+  if (process.argv[3] !== "--skip-format") {
     // run the formatting tool for .net, to ensure the files are prepped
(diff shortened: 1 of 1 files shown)


utils/doclint/generateDotnetApi.js
```diff
@@ -199,6 +200,11 @@ const customTypeNames = new Map([
         out.push(`\t${escapedName},`);
       });
     }, enumsDir));
+  
+  if (process.argv[3] !== "--skip") {
```
avodovnik, line 204: I'm sure there's a better way to do this, but this solves the problem _for now_ and lets us skip the formatting on machines that may not have `dotnet` installed.

kblok (approved):

utils/doclint/generateDotnetApi.js
yury-s: should it be `--skip-format` ?

yury-s (approved):

PR #6379, opened 2021-04-30
Update class-page example code

xg-wang:
`await` needs `async` keyword

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -141,7 +141,7 @@ page.on('console', async msg => {
   for (let i = 0; i < msg.args().length; ++i)
     console.log(`${i}: ${await msg.args()[i].jsonValue()}`);
 });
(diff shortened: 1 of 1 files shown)


docs/src/api/class-page.md
```diff
@@ -137,7 +137,7 @@ The arguments passed into `console.log` appear as arguments on the event handler
 An example of handling `console` event:
 
 ```js
-page.on('console', msg => {
+page.on('console', async msg => {
   for (let i = 0; i < msg.args().length; ++i)
     console.log(`${i}: ${await msg.args()[i].jsonValue()}`);
 });
```
pavelfeldman, line 143: Looks like page.evaluate() below also needs await before it...

xg-wang: Oops didn't copy that :p

PR #6381, opened 2021-05-01
chore(electron): fix node/browser race conditions, expose browser win…

pavelfeldman:
…dow asynchronously

### DIFF
--- a/src/client/electron.ts
+++ b/src/client/electron.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import type { BrowserWindow } from 'electron';
 import * as structs from '../../types/structs';
 import * as api from '../../types/types';
 import * as channels from '../protocol/channels';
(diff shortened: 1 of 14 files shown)


src/server/electron/electron.ts
```diff
@@ -105,19 +83,25 @@ export class ElectronApplication extends SdkObject {
...
     this._nodeConnection.close();
     await closed;
   }
 
+  async browserWindow(page: Page): Promise<js.JSHandle<BrowserWindow>> {
+    const electronHandle = await this._nodeElectronHandlePromise;
+    return await electronHandle.evaluateHandle(({ BrowserWindow }, windowId) => BrowserWindow.fromId(windowId), (page as any)._browserWindowId);
+  }
+
   async _init()  {
```
dgozman, line 96: nit: doesn't have to be async now.

dgozman (approved):

src/client/electron.ts
```diff
@@ -109,6 +108,13 @@ export class ElectronApplication extends ChannelOwner<channels.ElectronApplicati
     return result;
   }
 
+  async browserWindow(page: Page): Promise<JSHandle<BrowserWindow>> {
+    return this._wrapApiCall('electronApplication.browserWindow', async (channel: channels.ElectronApplicationChannel) => {
+      const result = await channel.browserWindow({ page: page._channel });
```
msyfls123, line 113: I‘m interested that what will we get if `page` is actually an Electron `BrowserView`?

PR #6402, opened 2021-05-04
fix: support relative downloadsPath directory for downloads

### LINKED ISSUE #6401: [BUG] relative downloadsPath does not work on Windows
**Context:**
- Playwright Version: 1.10.0
- Operating System: Windows
- Node.js version: v14.15.4
- Browser: Chromium

**Code Snippet**

I copied this file into my project and ran.
https://github.com/nmanikiran/playwright-demos/blob/master/download.js

**Describe the bug**

If I run it with ```const browser = await chromium.launch({ headless: false });```
Then I see it say at the bottom left of the page "Failed - download error".

mxschmitt:
Before this change there were two issues:
- On Windows it was not working at all when a relative download path was passed
- Artifact.path() -> Download.path() returned the relative path + uuid, it returns now it's absolute variant to be consistent with the behavior when an absolute path was given

Fixes #6401

### DIFF
--- a/src/server/artifact.ts
+++ b/src/server/artifact.ts
@@ -15,7 +15,6 @@
  */
 
 import fs from 'fs';
-import path from 'path';
 import * as util from 'util';
 import { SdkObject } from './instrumentation';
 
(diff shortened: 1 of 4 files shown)


dgozman (commented):
I'd rather resolve the path in `validateLaunchOptions` and assume absolute path throughout the rest of the code. This way we won't need any changes in Artifact or launch.

dgozman (approved):

PR #6408, opened 2021-05-04
docs(page): opener is not async anymore

kblok:
I know you might consider this a breaking change. But `opener` is not `async` anymore.
I don't know if we want to make this change now, or leave it as it is and, create async wrappers in the .NET binding.

### DIFF
--- a/src/client/page.ts
+++ b/src/client/page.ts
@@ -220,7 +220,7 @@ export class Page extends ChannelOwner<channels.PageChannel, channels.PageInitia
     return this._browserContext;
   }
 
-  async opener(): Promise<Page | null> {
+  opener(): Page | null {
     if (!this._opener || this._opener.isClosed())
       return null;
     return this._opener;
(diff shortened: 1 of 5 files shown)


src/client/page.ts
```diff
@@ -220,7 +220,7 @@ export class Page extends ChannelOwner<channels.PageChannel, channels.PageInitia
     return this._browserContext;
   }
 
-  async opener(): Promise<Page | null> {
+  opener(): Page | null {
```
yury-s, line 223: This is not fully backwards compatible as it will break code like this:
```js
  page.opener().then(p => console.log(p.url()));
```
I'm also not quite sure about implications of removing `async` from a function in python. Considering that I'd leave current signature as is, the implementation can still be synchronous.

kblok: Yeah. This is more to open the conversation and how we want to proceed with a method that we do know it's async.
Maybe starting with a custom sync csharp method?

yury-s: I'd keep it async for consistency with other languages, the implementation can be sync or async the client should not care. Does it cause any issues with the implementation?

kblok: Nope, we can return a resolved task.

PR #6411, opened 2021-05-05
devops: do a full browser checkout by default on Dev machines

aslushnikov:
The `git fetch --unshallow` is a non-trivial command to run, so we
should default to a full checkout unless we're on CI.


yury-s (approved):
Will this work nicely with existing shallow checkouts?

aslushnikov:
> Will this work nicely with existing shallow checkouts?

Yep, it won't change these at all.

PR #6417, opened 2021-05-05
fix(types): add better descriptions for event emitter methods

### LINKED ISSUE #3174: [BUG] 1.2.1 Docs: Page event listeners description in documentation
Buenos dias amigos!
I noticed that PW documentation does not contain any information about `page.on(), page.off(), page.once()` event listeners. Can you please add some description? Difference between on() and once() or usage of off() can be not obvious for those who did not work with event listeners before. And hints are not so helpful (at least in VS Code):

image
image

JoelEinbinder:
We currently use the same description for `on`, `off`, `once`, `addListener`, and `removeListener` methods. But describing the event itself is only important for `on` and `addListener`.

Fixes #3174

### DIFF
--- a/utils/generate_types/index.js
+++ b/utils/generate_types/index.js
@@ -220,10 +220,17 @@ function createEventDescriptions(classDesc) {
 function classBody(classDesc) {
   const parts = [];
   const eventDescriptions = createEventDescriptions(classDesc);
+  const commentForMethod = {
+    off: 'Removes and event listener added by `on` or `addListener`',
+    removeListener: 'Removes and event listener added by `on` or `addListener`',
+    once: 'Adds an event listener that will be automatically removed after it is triggered once. See `addListener` for more information about this event.'
+  }
   for (const method of ['on', 'once', 'addListener', 'removeListener', 'off']) {
     for (const {eventName, params, comment} of eventDescriptions) {
-        if (comment)
+        if ((method === 'on' || method === 'addListener') && comment)
           parts.push(writeComment(comment, '  '));
+        else
+          parts.push(writeComment(commentForMethod[method], '  '));
         parts.push(`  ${method}(event: '${eventName}', listener: (${params}) => void): this;\n`);
     }
   }
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -573,178 +573,97 @@ export interface Page {
   on(event: 'worker', listener: (worker: Worker) => void): this;
 
   /**
-   * Emitted when the page closes.
+   * Adds an event listener that will be automatically removed after it is triggered once. See `addListener` for more information about this event.
    */
   once(event: 'close', listener: (page: Page) => void): this;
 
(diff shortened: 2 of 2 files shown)


utils/generate_types/index.js
```diff
@@ -425,4 +434,4 @@ function renderJSSignature(args) {
   if (hasOptional)
     tokens.push(']');
   return tokens.join('');
-}
```
mxschmitt, line 428: nit: EOF line was removed, would be good to keep it how it was.

utils/generate_types/index.js
```diff
@@ -211,10 +211,19 @@ function createEventDescriptions(classDesc) {
...
+    on: 'Adds an event listener.',
+    addListener: 'Adds an event listener.',
+    off: 'Removes an event listener.',
+    removeListener: 'Removes an event listener.',
+    once: 'Adds an event listener that will be automatically removed after it is triggered once.'
+  }
   for (const method of ['on', 'once', 'addListener', 'removeListener', 'off']) {
     for (const {eventName, params, comment} of eventDescriptions) {
-        if (comment)
+        if ((method === 'on' || method === 'addListener') && comment)
```
mxschmitt, line 223: `once` might make sense there too, or do you think its more rarely used?

mxschmitt (approved):

utils/generate_types/index.js
JoelEinbinder: I think users are more like to know what the event they are listening to does than know what once does.

utils/generate_types/index.js
```diff
@@ -211,10 +211,19 @@ function createEventDescriptions(classDesc) {
 function classBody(classDesc) {
   const parts = [];
   const eventDescriptions = createEventDescriptions(classDesc);
+  const commentForMethod = {
+    on: 'Adds an event listener.',
```
dgozman, line 215: on and addListener items here are unused.

utils/generate_types/index.js
```diff
@@ -211,10 +211,19 @@ function createEventDescriptions(classDesc) {
...
+    addListener: 'Adds an event listener.',
+    off: 'Removes an event listener.',
```
dgozman, line 217: `Removes and event listener added by "on" or "addListener".`

utils/generate_types/index.js
```diff
@@ -211,10 +211,19 @@ function createEventDescriptions(classDesc) {
...
+    removeListener: 'Removes an event listener.',
```
dgozman, line 218: Same here.

utils/generate_types/index.js
```diff
@@ -211,10 +211,19 @@ function createEventDescriptions(classDesc) {
...
+    once: 'Adds an event listener that will be automatically removed after it is triggered once.'
```
dgozman, line 219: `... See "addListener" for more information about this event.`

PR #6436, opened 2021-05-06
test: fixed flaky connectOverCDP tests

mxschmitt:
Currently in some places we are using `BrowserType.launch` to launch a Chromium browser with the `--remote-debugging-port=` argument. Chromium internally takes a bit to expose the WebSocket server. This resulted on Windows to flaky tests.

~q: I thought about when `remote-debugging-port` is passed as an argument, should we use actually this WebSocket connection for Playwright or stick with the pipe transport?~
~q: We could also probably fix it in the tests by passing in the few places `useWebSocket: true` but I think its better to auto-wait internally.~

See: https://devops.aslushnikov.com/flakiness2.html#filter_spec=should+connect+to&commits=50&browser=chromium&platform=Windows

### DIFF
--- a/src/server/browserType.ts
+++ b/src/server/browserType.ts
@@ -186,7 +186,8 @@ export abstract class BrowserType extends SdkObject {
       await validateHostRequirements(this._registry, options.channel as registry.BrowserName);
 
     let wsEndpointCallback: ((wsEndpoint: string) => void) | undefined;
-    const wsEndpoint = options.useWebSocket ? new Promise<string>(f => wsEndpointCallback = f) : undefined;
+    const shouldWaitForWSListening = options.useWebSocket || options.args?.some(a => a.startsWith('--remote-debugging-port'));
+    const waitForWSEndpoint = shouldWaitForWSListening ? new Promise<string>(f => wsEndpointCallback = f) : undefined;
     // Note: it is important to define these variables before launchProcess, so that we don't get
     // "Cannot access 'browserServer' before initialization" if something went wrong.
     let transport: ConnectionTransport | undefined = undefined;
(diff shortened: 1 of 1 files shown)


src/server/browserType.ts
```diff
@@ -186,7 +186,8 @@ export abstract class BrowserType extends SdkObject {
       await validateHostRequirements(this._registry, options.channel as registry.BrowserName);
 
     let wsEndpointCallback: ((wsEndpoint: string) => void) | undefined;
-    const wsEndpoint = options.useWebSocket ? new Promise<string>(f => wsEndpointCallback = f) : undefined;
+    const connectOverWebSocket = options.useWebSocket || options.args?.some(a => a.startsWith('--remote-debugging-port='));
```
dgozman, line 189: I don't think we actually want to use the websocket connection in this case. Even the current "useWebsocket" is a test-only hack.

mxschmitt: Discussed offline: If the argument is passed, we will wait that it's listening but not use it.

src/server/browserType.ts
```diff
@@ -186,7 +186,8 @@ export abstract class BrowserType extends SdkObject {
       await validateHostRequirements(this._registry, options.channel as registry.BrowserName);
 
     let wsEndpointCallback: ((wsEndpoint: string) => void) | undefined;
-    const wsEndpoint = options.useWebSocket ? new Promise<string>(f => wsEndpointCallback = f) : undefined;
+    const shouldWaitForWSListening = options.useWebSocket || options.args?.some(a => a.startsWith('--remote-debugging-port='));
```
dgozman, line 189: nit: It is possible to pass two arguments: `--remote-debugging-port`, `1234`. So, let's drop the `=` sign from `startsWith`.

dgozman (approved):

PR #6438, opened 2021-05-06
devops: use chromium's FILES.cfg to compute archive files

aslushnikov:
Since the include list of files to pack with Chromium changes
eventually, we should use their list instead of hardcoded one
on our side.

### DIFF
--- a/browser_patches/chromium/compute_files_to_archive.py
+++ b/browser_patches/chromium/compute_files_to_archive.py
@@ -1,0 +1,57 @@
+#!/usr/bin/python
+
+import sys
+import json
+
+if len(sys.argv) < 2:
+    print "ERROR: expected arch: 32bit or 64bit"
+    sys.exit(1)
+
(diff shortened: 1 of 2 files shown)


browser_patches/chromium/compute_files_to_archive.py
```diff
@@ -0,0 +1,57 @@
...
+if sys.version_info > (3, 0):
+    exec(open(file_name).read())
+    descriptors = FILES
+else:
+    variables = {}
+    execfile(file_name, variables)
+    descriptors = variables['FILES']
+
+def filter_descriptors(entry):
+    if 'archive' in entry:
```
yury-s, line 43: shouldn't those be included?

yury-s (approved):

browser_patches/chromium/compute_files_to_archive.py
aslushnikov: nah, this means that descriptor belongs to non-default archive. We care about chromium's archive only.

Issue #6439, opened 2021-05-06
[BUG] screencast tests fail on Mac10.14


aslushnikov:
Offending patch: 791443d71f0c014f41895e2fceecbb762d3f38d7
Failing tests: https://devops.aslushnikov.com/flakiness2.html#branch=master&browser=webkit&platform=Darwin+10.14&timestamp=1620322239924

aslushnikov:
This is fixed now. We discussed #6445 with Yury and decided to keep the frontend approach since its tiny.

PR #6445, opened 2021-05-06
browser(deprecated-webkit-mac-10.14): align Screencast protocol calls

aslushnikov:
This patch aligns protocol calls for deprecated webkit for mac10.14 with
the tip-of-tree protocol calls.

References #6439


yury-s (commented):
I'd rather leave it as is given that #6440 fixes the problem on the client side.

PR #6467, opened 2021-05-08
feat: add common mac editing  commands.

msyfls123:
Editing commands works fine on Chrome now, while still fails on non-win32 Webkit.
|Browser(macOS)|Before|After|
|---|---|---|
|Chrome|❌|✅|
|Firefox|✅|✅|
|Webkit|❌|❌|


JoelEinbinder:
Thanks for the PR. For awhile these commands were done differently in Chrome. I'll look into why they broke.

yury-s:
@JoelEinbinder did you find anything or should we land this?

mxschmitt:
Closing since it seems stale.

PR #6498, opened 2021-05-11
devops(docker): upgrade to node 16

chambo-e:
Hello 👋 

Now that node@16 is out 

I haven't seen a node@14 requirement, so I've overwritten the current image, but I could decline the image in flavors if you prefer (`bionic-node14`, `bionic-node16`, ...)

### DIFF
--- a/utils/docker/Dockerfile.bionic
+++ b/utils/docker/Dockerfile.bionic
@@ -2,9 +2,9 @@ FROM ubuntu:bionic
 
 # === INSTALL Node.js ===
 
-# Install node14
+# Install node16
 RUN apt-get update && apt-get install -y curl && \
-    curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
+    curl -sL https://deb.nodesource.com/setup_16.x | bash - && \
     apt-get install -y nodejs
 
 # Feature-parity with node.js base images.
@@ -40,4 +40,3 @@ RUN mkdir /ms-playwright && \
     DEBIAN_FRONTEND=noninteractive npx playwright install-deps && \
     rm -rf /tmp/pw && rm /tmp/playwright.tar.gz && \
     chmod -R 777 /ms-playwright
-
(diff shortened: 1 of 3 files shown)


matthprost (approved):

dmzski (approved):

utils/docker/Dockerfile.bionic
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:bionic
 
 # Install node14
```
DorianMaliszewski, line 5: Need to change this in every files ? 🤔

chambo-e: Absolutely !

pavelfeldman (approved):

pavelfeldman:
@aslushnikov : this looks good to me, any reason not to do this?

mxschmitt:
> @aslushnikov : this looks good to me, any reason not to do this?

(so far we are only testing on 12 and 14, so it might not work)

philibea (approved):

aslushnikov (approved):

mxschmitt:
> @aslushnikov @pavelfeldman In the future, can you please continue to publish the last LTS up until it is actually deprecated? Due to this change my organization is unable to move past this SHA because we weren't ready for Node 16 yet (meaning we are now three months behind on playwright upgrades). It would be more useful to have the docker images tagged for multiple versions of Node instead of just the ubuntu image it is on. I would expect Playwright to support Node 14 until at least the end of the maintenance LTS (which happens to have been yesterday according to https://nodejs.org/en/about/releases/). What changes need to be made so that in the future there can be images created that reflect the Node LTS matrix? I'm happy to open an issue on this to discuss more too.

You can install a custom Node.js version inside the docker image by uninstalling the previous Node.js version and installing your preferred one see here: https://github.com/nodesource/distributions/blob/master/README.md#debinstall

Would that work for you?

PR #6501, opened 2021-05-11
feat(download): add Page in Download

regseb:
The `Page` class provides a [method](https://playwright.dev/docs/api/class-page/#pagecontext) to retrieve its `BrowserContext`. The `BrowserContext` has a [method](https://playwright.dev/docs/api/class-browsercontext#browsercontextbrowser) for its `Browser`. This pull request adds a method to retrieve the `Page` of a `Download`.

### DIFF
--- a/docs/src/api/class-download.md
+++ b/docs/src/api/class-download.md
@@ -69,6 +69,11 @@ Deletes the downloaded file. Will wait for the download to finish if necessary.
 
 Returns download error if any. Will wait for the download to finish if necessary.
 
+## method: Download.page
+- returns: <[Page]>
+
+Get the page that the download belongs to.
+
 ## async method: Download.path
 - returns: <[null]|[path]>
 
--- a/src/client/download.ts
+++ b/src/client/download.ts
@@ -17,16 +17,23 @@
 import { Readable } from 'stream';
 import * as api from '../../types/types';
 import { Artifact } from './artifact';
+import { Page } from './page';
 
 export class Download implements api.Download {
+  private _page: Page;
   private _url: string;
   private _suggestedFilename: string;
   private _artifact: Artifact;
 
-  constructor(url: string, suggestedFilename: string, artifact: Artifact) {
+  constructor(page: Page, url: string, suggestedFilename: string, artifact: Artifact) {
+    this._page = page;
     this._url = url;
     this._suggestedFilename = suggestedFilename;
     this._artifact = artifact;
+  }
+
+  page(): Page {
+    return this._page;
   }
 
   url(): string {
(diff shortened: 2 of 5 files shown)


pavelfeldman:
> I have a method that opens a page, starts a download and returns the instance of Download. Then the parent code waits for the download to finish and closes the page.

It looks like you could modify it so that the code could encapsulate the download process: open a page, wait for download, close page, return download? Trying to figure out if we want that convenience page accessor there.

The issue is that in many cases, new page (window.open) would be converted into the download process and the page instance inside Download will be invalid from the get go. Downloads belong to the contexts, not pages, so I'm hesitant making this connection strong.

regseb:
I want the pages to be processed sequentially and the downloads to be done in parallel. Here is a pseudo-code:

```JavaScript
for (const i of ["foo", "bar"]) {
    const download = await functionThatStartsDownload(i);
    download.path().then((path) => {
        download.page().close();
        // do something.
    });
}
```

If the downloads belong to the contexts, we can pass the context to the downloads constructor: `this.emit(Events.Page.Download, new Download(this._browserContext, url, suggestedFilename, artifactObject));`. And add a method returning it.
I could close the context which will close its pages.

pavelfeldman:
> I want the pages to be processed sequentially and the downloads to be done in parallel. Here is a pseudo-code

There is a slight problem with this code in that you will initiate all the downloads sequentially, the for loop will end and you will close your browser, interrupting all the downloads. So you absolutely need to store these download.path() promises and then Promise.all them. And since you anyways need to do that, you could as well modify your code to be:

```js
const downloadPromises = [];
for (...) {
 downloadPromises.push(await functionThatStartsDownload(i));
}
await Promise.all(downloadPromises);
```

where returned value is a promise that awaits path and closes the page...

pavelfeldman:
Having said that, let me discuss with the team how risky it is to expose page on the Download object.

pavelfeldman (approved):

regseb:
For the problem of closing the browser, I have created a _pool_ that automatically closes the browser when there is no more page open:

```JavaScript
const BrowserPool = class {

    _browser = null;
    _options = {};

    constructor(options = {}) {
        this._options = options;
    }

    async newPage(options) {
        if (null === this._browser) {
            this._browser = await chromium.launch(this._options);
        }

        const page = await this._browser.newPage(options);
        page.on("close", () => {
            page.context().close();
            // Si c'était le dernier contexte : fermer le navigateur.
            if (1 === this._browser.contexts().length) {
                this._browser.close();
                this._browser = null;
            }
        });
        return page;
    }
};
```

PR #6503, opened 2021-05-11
docs(dotnet): BrowserContext and BrowserType

avodovnik:
- BrowserContext
- BrowserType

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -49,6 +49,18 @@ page.goto("https://example.com")
 context.close()
 ```
 
+```csharp
+using var playwright = await Playwright.CreateAsync();
+var browser = await playwright.Firefox.LaunchAsync(headless: false);
+// Create a new incognito browser context
+var context = await browser.NewContextAsync();
+// Create a new page inside context.
+var page = await context.NewPageAsync();
+await page.GoToAsync("https://bing.com");
+// Dispose context once it"s no longer needed.
+await context.CloseAsync();
+```
+
 ## event: BrowserContext.backgroundPage
 * langs: js, python
 - argument: <[Page]>
@@ -119,6 +131,13 @@ page = page_info.value
 print(page.evaluate("location.href"))
 ```
 
+```csharp
+var popupTask = context.WaitForPageAsync();
+await Task.WhenAll(popupTask, page.ClickAsync("a"));
+
+Console.WriteLine(popupTask.Result.EvaluateAsync<string>("location.href"));
+```
+
 :::note
 Use [`method: Page.waitForLoadState`] to wait until the page gets to a particular state (you should not need it in most
 cases).
@@ -153,6 +172,10 @@ await browser_context.add_cookies([cookie_object1, cookie_object2])
 
 ```python sync
 browser_context.add_cookies([cookie_object1, cookie_object2])
+```
+
+```csharp
+await context.AddCookiesAsync(new[] { cookie1, cookie2 });
 ```
 
 ### param: BrowserContext.addCookies.cookies
@@ -204,6 +227,10 @@ await browser_context.add_init_script(path="preload.js")
 ```python sync
 # in your playwright script, assuming the preload.js file is in same directory.
 browser_context.add_init_script(path="preload.js")
+```
+
+```csharp
+await context.AddInitScriptAsync(scriptPath: "preload.js");
 ```
 
 :::note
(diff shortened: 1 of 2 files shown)


docs/src/api/class-browsercontext.md
```diff
@@ -49,6 +49,18 @@ page.goto("https://example.com")
 context.close()
 ```
 
+```csharp
+using var playwright = await Playwright.CreateAsync();
```
yury-s, line 53: shouldn't it be just `var playwright...` ?

avodovnik: No, this is C#'s shorthand for doing a `using(var playwright...) { }`

docs/src/api/class-browsercontext.md
```diff
@@ -119,6 +131,15 @@ page = page_info.value
 print(page.evaluate("location.href"))
 ```
 
+```csharp
+var popupTask = context.WaitForPageAsync();
+await Task.WhenAll(popupTask, page.ClickAsync("a"))
+    .WithTimeout(1000) // add using Microsoft.Playwright.Helpers; to get this helper
```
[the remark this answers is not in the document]
avodovnik: It's intentionally not the same. As you point out, the idea here is to have a way to make sure the example does not dead-lock if something goes wrong _downstream_ from us. And this also applies to both tasks and the potential opportunity for a deadlock. Granted, in this example, that possibility is low, but as a general pattern, waiting for both tasks to complete might introduce a dead-lock and I think the right thing to do, is have a good pattern for it everywhere we use it.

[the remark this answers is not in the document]
pavelfeldman: What kind of deadlock are you talking about? In the present system, we only manage timeouts on the server side, so I'm with Yury here and I actually left a similar comment in another example. We should only use timeouts inside ClickAsync and WaitForPageAsync.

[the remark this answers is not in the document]
avodovnik: I think you've got a point - the "example" code comes from our (.net) tests so it's a bit more defensive (i.e. we don't want it hanging in case we made a coding mistake and mis-handled something from downstream) which is where that timeout comes into play. 

So, from a simplicity perspective, I'm happy to loose it. We can add a section on best practices later/in places more appropriate. Considering this isn't _test_ or _library_ code per se, we can get rid of the `ConfigureAwait` as well.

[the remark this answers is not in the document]
kblok: Yeah, let's remove the `WithTimeout`. That's internal. Let's also remove the `ConfigureAwait` bellow.

docs/src/api/class-browsercontext.md
```diff
@@ -155,6 +176,10 @@ await browser_context.add_cookies([cookie_object1, cookie_object2])
 browser_context.add_cookies([cookie_object1, cookie_object2])
 ```
 
+```csharp
+context.AddCookiesAsync(new[] { cookie1, cookie2 });
```
kblok, line 180: Add the `await`?

docs/src/api/class-browsercontext.md
```diff
@@ -478,6 +542,26 @@ page.set_content("""
 """)
 ```
 
+```csharp
+var result = new TaskCompletionSource<string>();
+var page = await Context.NewPageAsync();
+await Context.ExposeBindingAsync("clicked", async (BindingSource _, IJSHandle t) =>
```
kblok, line 548: @dgozman We already have a `Func<BindingSource, T, TResult>` overload :)
So you can run `async` code, but that `Task` won't be awaited inside our handler.

docs/src/api/class-browsercontext.md
```diff
@@ -774,6 +903,14 @@ page.goto("https://example.com")
 browser.close()
 ```
 
+```csharp
+var context = await browser.NewContextAsync();
+var page = await context.NewPageAsync();
+await context.RouteAsync(new System.Text.RegularExpressions.Regex("(\\.png$)|(\\.jpg$)"), r => r.AbortAsync());
```
kblok, line 909: How about?

```suggestion
await context.RouteAsync(new Regex("(\\.png$)|(\\.jpg$)"), r => r.AbortAsync());
```

docs/src/api/class-browsercontext.md
```diff
@@ -1029,6 +1184,13 @@ with context.expect_event("page") as event_info:
 page = event_info.value
 ```
 
+```csharp
+var t = page.WaitForEventAsync("page");
+await Task.WhenAll(t, page.ClickAsync("button"))
+    .WithTimeout(1000) // make sure to add Microsoft.Playwright.Helpers
+    .ConfigureAwait(false);
```
kblok, line 1191:
```suggestion
await Task.WhenAll(t, page.ClickAsync("button"));
```

docs/src/api/class-browsercontext.md
avodovnik: Sure, I kept it compilable without `using` :) Works for me though.

docs/src/api/class-browsercontext.md
dgozman: Nice!

PR #6509, opened 2021-05-11
fix: propagate custom executable path to codegen

yury-s:


tests/inspector/inspectorTest.ts
```diff
@@ -186,7 +189,7 @@ class CLIMock {
     });
     this.exited = new Promise((f, r) => {
       this.process.stderr.on('data', data => {
-        r(new Error(data));
+        console.error(data.toString());
       });
       this.process.on('exit', f);
```
dgozman, line 194: Let's maybe reject if the exit code is not zero?

dgozman (approved):

tests/inspector/inspectorTest.ts
yury-s: done.

PR #6511, opened 2021-05-11
fix(chromium): wait for existing pages when connecting

### LINKED ISSUE #679: can't get browser's context pages after connect_over_cdp
**Test on Chrome Version 90.0.4430.212 (Official Build) (64-bit) (Windows 10 20H2 OS Build 19042.928)
chrome lunched with "--remote-debugging-port=9222"**

### code:
```
##########
playwright = sync_playwright().start()
chrome = playwright.chromium.connect_over_cdp("http://localhost:9222")
c1 = chrome.contexts[0] # default context
print(c1.pages) # []
print(len(c1.pages)) # 0
##########
```

### but i am able to get all the opened pages after i open a new_page:
```
##########
playwright = sync_playwright().start()
chrome = playwright.chromium.connect_over_cdp("http://localhost:9222")
c1 = chrome.contexts[0]
c1.new_page()
print(c1.pages)
print(len(c1.pages)) # 6 (5 opened pages + the new page)
##########
```

JoelEinbinder:
fixes https://github.com/microsoft/playwright-python/issues/679
﻿

### DIFF
--- a/src/server/chromium/crBrowser.ts
+++ b/src/server/chromium/crBrowser.ts
@@ -61,12 +61,16 @@ export class CRBrowser extends Browser {
       return browser;
     }
     browser._defaultContext = new CRBrowserContext(browser, undefined, options.persistent);
-
     await Promise.all([
-      session.send('Target.setAutoAttach', { autoAttach: true, waitForDebuggerOnStart: true, flatten: true }),
+      session.send('Target.setAutoAttach', { autoAttach: true, waitForDebuggerOnStart: true, flatten: true }).then(async () => {
+        // Target.setAutoAttach has a bug where it does not wait for new Targets being attached.
+        // However making a dummy call afterwards fixes this.
+        // This can be removed after https://chromium-review.googlesource.com/c/chromium/src/+/2885888 lands in stable.
+        await session.send('Target.getTargetInfo');
+      }),
       (browser._defaultContext as CRBrowserContext)._initialize(),
     ]);
-
+    await browser._waitForAllPagesToBeInitialized();
     return browser;
   }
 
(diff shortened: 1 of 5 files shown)


src/server/chromium/crBrowser.ts
```diff
@@ -61,12 +61,16 @@ export class CRBrowser extends Browser {
...
     }
     browser._defaultContext = new CRBrowserContext(browser, undefined, options.persistent);
-
     await Promise.all([
-      session.send('Target.setAutoAttach', { autoAttach: true, waitForDebuggerOnStart: true, flatten: true }),
+      session.send('Target.setAutoAttach', { autoAttach: true, waitForDebuggerOnStart: true, flatten: true }).then(async () => {
+        // Target.setAutoAttach has a bug where it does not wait for new Targets being attached.
+        // However making a dummy call afterwards fixes this.
+        // This can be removed after https://chromium-review.googlesource.com/c/chromium/src/+/2885888 lands in stable.
+        await session.send('Target.getTargetInfo');
```
yury-s, line 69: We used to have code with `setDiscoverTargets` which did this, maybe it's easier to just revert that?

yury-s (approved):

src/server/chromium/crBrowser.ts
JoelEinbinder: The code with setDiscoverTargets was way more complicated had bugs in it related to connecting twice to service workers.

JoelEinbinder:
Looks like this is causing an electron test to fail. Looking into it

JoelEinbinder:
Electron had a race where it was not expecting pages to appear before the browser was finished launching. I fixed it, and a test that was relying on that behavior. This could affect user code. I don't think we have a problem, because we used `firstWindow` everywhere in the docs. @pavelfeldman what do you think?

PR #6512, opened 2021-05-11
docs(java): add  multithreading section

yury-s:


mxschmitt (commented):
Maybe we can abstract it and apply it for Python, Java, C# too? Just for each language with different snippets, since this applies for all languages in general.

yury-s:
I don't think multithreding is that common in Python, especially considering our async API so I would not merge this. For Java people are more tempted to create new threads. For C# I don't yet know what the recommended way there is but given that we have async/await there too I expect the wording to be different. Considering all this I'd start with a separate description for java as it's the only language where we've seen several questions from the clients about the matter. Should need be we can extend it to other languages too.

mxschmitt:
Fine, I mentioned it because on Python side around ~5 issues were already posted who had questions about thread-safety or ran into issues because of it.

dgozman (approved):

PR #6514, opened 2021-05-11
docs: add ref to waitForSelector from querySelector

### LINKED ISSUE #680: [BUG] Cannot find an unordered list element when crawling linkedin
**Context:**
- Playwright Version: [1.11.0]
- Operating System: [Debian 10 Linux]
- Python version: [3.9.2]
- Node version (in the python site-packages): [v12.20.1]
- Browser: [test on: Chromium. Haven't tried webkit]

**Code Snippet**

On [this](https://www.linkedin.com/search/results/people/?keywords=Quantum%20Computing&origin=SWITCH_SEARCH_VERTICAL) search results page on linked in (needs login), I query the following, with result in comments

```python 
result = await page.query_selector("//ul[@class='reusable-search__entity-results-list list-style-none']")
# <class 'NoneType'>

result = await page.query_selector("[class='reusable-search__entity-results-list list-style-none']")
#  <class 'NoneType'>

results = await page.query_selector_all("ul")
# has elements, but does not include class of desired element: 'reusable-search__entity-results-list list-style-none'

results = await page.query_selector_all("li.entity-result__item")
# has no elements
```

**Describe the bug**
The comment line following each result query shows whats contained in the `result` variable

I should be getting an element handle, but I am only getting `None` values. The first `query_selector_all("ul")` results in a list of 4 handles, but does not include the one I am trying to access, which is the search results directly

Not sure if this is a bug or if I am missing something

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/680

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -2046,7 +2046,7 @@ Time to wait between `keydown` and `keyup` in milliseconds. Defaults to 0.
 - returns: <[null]|[ElementHandle]>
 
 The method finds an element matching the specified selector within the page. If no elements match the selector, the
-return value resolves to `null`.
+return value resolves to `null`. To wait for an element on the page, use [`method: Page.waitForSelector`].
 
 Shortcut for main frame's [`method: Frame.querySelector`].
 
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -167,7 +167,8 @@ export interface Page {
 
   /**
    * The method finds an element matching the specified selector within the page. If no elements match the selector, the
-   * return value resolves to `null`.
+   * return value resolves to `null`. To wait for an element on the page, use
+   * [page.waitForSelector(selector[, options])](https://playwright.dev/docs/api/class-page#pagewaitforselectorselector-options).
    *
    * Shortcut for main frame's [frame.$(selector)](https://playwright.dev/docs/api/class-frame#frameselector).
    * @param selector A selector to query for. See [working with selectors](https://playwright.dev/docs/selectors) for more details.


docs/src/api/class-page.md
```diff
@@ -2059,7 +2059,7 @@ Shortcut for main frame's [`method: Frame.querySelector`].
 - returns: <[Array]<[ElementHandle]>>
 
 The method finds all elements matching the specified selector within the page. If no elements match the selector, the
-return value resolves to `[]`.
+return value resolves to `[]`. To wait for at least an element on the page, use [`method: Page.waitForSelector`].
```
dgozman, line 2062: I don't think this helps, because there is no way to wait for all elements matching the selector. We can refer to the `querySelector` for a single-element version though. WDYT?

dgozman (approved):

docs/src/api/class-page.md
mxschmitt: sounds good, updated the PR.

dgozman (approved):

mxschmitt:
> Question: Why have a `wait_for_selector` and `query_selector` instead of just `query_selector` with a `wait=[True|False]` option?
> 
> I guess thats kiinda unrelated though

interesting idea, for `query_selector_all` it's not possible, since it's never known how many items are on the page. But for `query_selector` I could imagine to allow a second parameter which represents the `wait_until` state, see here: https://playwright.dev/docs/api/class-page#pagewaitforselectorselector-options

Feel free to create a feature request as an issue. :+1:

PR #6518, opened 2021-05-11
devops: run less tests on each PR

aslushnikov:
With this patch, we have the following behaviors:
- For PR's, we will run only Linux tests for Chromium, Firefox and
  Webkit
- For commits to master and release-*, we'll run all the tests

To run all tests on PR, assign a "CQ1" label to the PR.


.github/workflows/tests_secondary.yml
```diff
@@ -1,11 +1,12 @@
-name: tests
+name: secondary tests
 
 on:
   push:
     branches:
       - master
       - release-*
   pull_request:
+    types: [ labeled ]
```
yury-s, line 9: Is it any label or `cq1` specifically?

yury-s (approved):
Primary and secondary is somewhat misleading as they make "secondary" tests appear less important. How about 'minimal test suite' and 'all tests' or something like that?

yury-s (approved):

PR #6526, opened 2021-05-12
docs(dotnet): examples for dialog, download and filechooser

avodovnik:


pavelfeldman (approved):

docs/src/api/class-dialog.md
```diff
@@ -80,6 +80,30 @@ with sync_playwright() as playwright:
...
+
+class DialogExample
+{
+    public static async Task Run()
+    {
+        using var playwright = await Playwright.CreateAsync();
+        var browser = await playwright.Chromium.LaunchAsync();
+        var page = await browser.NewPageAsync();
+
+        page.Dialog += async (_, dialog) =>
```
pavelfeldman, line 95: Does C# support async event handlers?

docs/src/api/class-download.md
```diff
@@ -48,6 +48,15 @@ download = download_info.value
 path = download.path()
 ```
 
+```csharp
+var downloadTask = page.WaitForDownloadAsync();
+await Task.WhenAll(downloadTask, page.ClickAsync("#downloadButton"))
+    .WithTimeout(1000)
```
pavelfeldman, line 54: - There is a natural timeout on both waitForDownload and click, why do we add another one?
- I don't think we need ConfigureAwait(false) in any of our examples and code.

docs/src/api/class-dialog.md
avodovnik: Yup - like in the example. It makes calling `async` methods further easier :-).

docs/src/api/class-download.md
```diff
@@ -48,6 +48,12 @@ download = download_info.value
 path = download.path()
 ```
 
+```csharp
+var downloadTask = page.WaitForDownloadAsync();
+await Task.WhenAll(downloadTask, page.ClickAsync("#downloadButton"));
+System.Console.WriteLine(await downloadTask.Result.PathAsync());
```
kblok, line 54:
```suggestion
Console.WriteLine(await downloadTask.Result.PathAsync());
```

kblok (approved):

kblok (approved):

docs/src/api/class-dialog.md
pavelfeldman: We should follow up to fix this here (and maybe in other places).

PR #6528, opened 2021-05-12
docs(dotnet): fix link regex on xmldocs

kblok:
That regex was generating invalid XML when it found text in brackets with no links. For instance: `[contenteditable]`  in ElementHandle fill.

Output code: https://github.com/microsoft/playwright-sharp/pull/1363


avodovnik (approved):
Just to keep the cases here, @kblok prepared the comparison of before & after:

Before: https://regex101.com/r/4ib21z/1
After: https://regex101.com/r/Bj6PLw/1

pavelfeldman (approved):

PR #6529, opened 2021-05-12
chore(dotnet): fix goto casing

kblok:
The dotnet team is using this casing https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/GotoExpression.cs#L12

@pavelfeldman keep in mind that this will slightly change the code being shown in the upcoming demo.

Output code: https://github.com/microsoft/playwright-sharp/pull/1363

### DIFF
--- a/docs/src/api/class-frame.md
+++ b/docs/src/api/class-frame.md
@@ -686,7 +686,6 @@ Attribute name to get the value for.
 ## async method: Frame.goto
 * langs:
   - alias-java: navigate
-  - alias-csharp: GoToAsync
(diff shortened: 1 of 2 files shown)


avodovnik (approved):
This breaks most of the current docs and the PRs in flight. We have an issue opened for it, and I'm happy to merge, but we *must* fix the references across the existing PRs.

pavelfeldman (approved):

PR #6533, opened 2021-05-12
docs(dotnet): rename getPayloadAsJson to PostDataJsonAsync

kblok:
We should give it the same name as in upstream.

Related to https://github.com/microsoft/playwright-sharp/issues/1361

### DIFF
--- a/docs/src/api/csharp.md
+++ b/docs/src/api/csharp.md
@@ -1,10 +1,13 @@
-## method: Request.getPayloadAsJson
+## method: Request.PostDataJsonAsync
 * langs: csharp
 - returns: <[JsonDocument]>
 
-Returns a [JsonDocument] representation of [`method: Request.postDataBuffer`].
+Returns parsed request's body for `form-urlencoded` and JSON as a fallback if any.
 
-### option: Request.getPayloadAsJson.serializerOptions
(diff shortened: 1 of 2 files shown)


docs/src/api/csharp.md
```diff
@@ -1,10 +1,13 @@
-## method: Request.getPayloadAsJson
+## method: Request.PostDataJsonAsync
```
pavelfeldman, line 1: All we need is a return type override, right?

Call it as the origianal, "## method: Request.postDataJSON" so that we could convert it to the override in a follow-up.

pavelfeldman (approved):

utils/doclint/api_parser.js
```diff
@@ -118,20 +118,27 @@ class ApiParser {
    * @param {MarkdownNode} spec
    */
   parseArgument(spec) {
+    const langMatches =function (memberLangs, argumentLangs) {
```
kblok, line 121: @pavelfeldman maybe I'm doing something dumb here. But I had to add this piece of code because my `serializerOptions` param was getting added to the `js, python` version of `postDataJSON` of the method. I need it in the csharp method version.

That reminded me that I use a different casing to fool that part.

thoughts?

pavelfeldman: It works for me w/o your modifications :/

PR #6535, opened 2021-05-12
chore(dotnet): generate generic WaitForEvents

kblok:
This will generate the `WaitForEventAsync` we had before (and were adding through overloads).


avodovnik:
I have strong feelings *against* merging this PR. While I (think I) understand the desire to keep only the generics, it makes it _more aligned_ with the upstream API. Sometimes, Do you have a particular reason why you don't want the `string` overload around?

kblok:
@avodovnik my main concern is not the argument but the `Task<object>` result.

pavelfeldman:
We don't have a generic WaitForEvent in Java, maybe we don't need one in C# as well? What is not covered if we just remove it?

dgozman:
@kblok With the latest API, I think we should close this PR?

kblok:
yes!

PR #6541, opened 2021-05-12
fix(CR): handle backgroundPages() onClose

mxschmitt:
Ref https://github.com/microsoft/playwright-python/issues/681


src/client/page.ts
```diff
@@ -208,7 +208,10 @@ export class Page extends ChannelOwner<channels.PageChannel, channels.PageInitia
 
   _onClose() {
     this._closed = true;
-    this._browserContext._pages.delete(this);
+    if (this._browserContext._pages.has(this))
```
pavelfeldman, line 211: nit: in JS, it is easier to just delete, no need to check.

pavelfeldman (approved):

PR #6545, opened 2021-05-12
docs: follow up on adding trace dir, unify launch options

pavelfeldman:


docs/src/api/class-tracing.md
```diff
@@ -1,7 +1,50 @@
...
+const context = await browser.newContext();
+await context.tracing.start({ name: 'trace', screenshots: true, snapshots: true });
+const page = await context.newPage();
+await page.goto('https://playwright.dev');
+await context.tracing.stop();
+await context.tracing.export('trace.zip');
+```
+
+```java
+Browser browser = chromium.launch(new BrowserType.LaunchOptions().setTraceDir("trace"
```
dgozman, line 19: This line is truncated.

types/types.d.ts
```diff
@@ -10882,7 +10906,7 @@ export interface LaunchOptions {
   channel?: "chrome"|"chrome-beta"|"chrome-dev"|"chrome-canary"|"msedge"|"msedge-beta"|"msedge-dev"|"msedge-canary"|"firefox-stable";
 
   /**
-   * Enable Chromium sandboxing. Defaults to `false`.
+   * Enable Chromium sandboxing. Defaults to `true`.
```
dgozman, line 10909: Why did this change?

dgozman (approved):

types/types.d.ts
pavelfeldman: There was a bug I think...

kblok:
@pavelfeldman, @dgozman  My two cents here. Let's remember that a change in the arguments order could be considered a breaking change.

PR #6555, opened 2021-05-13
docs(dotnet): Frame examples

avodovnik:


docs/src/api/class-frame.md
```diff
@@ -97,6 +97,33 @@ with sync_playwright() as playwright:
     run(playwright)
 ```
 
+```csharp
+using Microsoft.Playwright;
+using System;
+using System.Threading.Tasks;
+
```
kblok, line 104: Do you think we need to add the using in examples? just wondering.

docs/src/api/class-frame.md
```diff
@@ -97,6 +97,33 @@ with sync_playwright() as playwright:
...
+{
+    public static async Task Main()
+    {
+        using var playwright = await Playwright.CreateAsync();
+        var browser = await playwright.Firefox.LaunchAsync();
+        var page = await browser.NewPageAsync();
+
+        await page.GoToAsync("https://www.bing.com");
+        DumpFrameTree(page.MainFrame, string.Empty);
+        await browser.CloseAsync();
```
kblok, line 115: Why not doing `await using`? we would be teaching people new stuff :)

docs/src/api/class-frame.md
kblok: This applies to the rest of the open PRs

docs/src/api/class-frame.md
kblok: This applies to the rest of the open PRs

docs/src/api/class-frame.md
avodovnik: I mimic the other languages. Some examples are "full", and it feels natural for us to do the same. *Ideally* down the road, we'll have a "Run this example" somewhere, but... until then... I'm up for consistency.

docs/src/api/class-frame.md
avodovnik: Yup, I've changed all three occurrences.

docs/src/api/class-frame.md
```diff
@@ -421,6 +463,10 @@ divs_counts = await frame.eval_on_selector_all("div", "(divs, min) => divs.lengt
 divs_counts = frame.eval_on_selector_all("div", "(divs, min) => divs.length >= min", 10)
 ```
 
+```csharp
+var divsCount = await page.EvalOnSelectorAllAsync<bool>("div", "(divs, min) => divs.length >= min", 10);
```
avodovnik, line 467:
```suggestion
var divsCount = await frame.EvalOnSelectorAllAsync<bool>("div", "(divs, min) => divs.length >= min", 10);
```

PR #6577, opened 2021-05-14
docs(dotnet): remove async event handlers in dialog

avodovnik:
This is the start (and place for discussion) for cleaning up `async` event handlers. 

cc: @kblok and @pfeldman

In places like this, this "fire & forget" pattern is good enough. That said, I think we should be careful about promoting potential bad practices, but maybe that's a topic for the guides, rather than API docs.


pavelfeldman (approved):

mxschmitt:
@avodovnik still relevant?

PR #6583, opened 2021-05-14
docs(dotnet): First part/pass for guides

avodovnik:


docs/src/auth.md
```diff
@@ -71,6 +71,18 @@ page.click('text=Submit')
 # Verify app is logged in
 ```
 
+```csharp
+var page = await browser.NewPageAsync();
+await page.GoToAsync("https://github.com/login");
```
pavelfeldman, line 76: GoToAsync is now Goto

docs/src/auth.md
```diff
@@ -290,6 +329,20 @@ with sync_playwright() as p:
...
+```csharp
+using Microsoft.Playwright;
+
+class Guides
+{
+  public async void Main()
+  {
+      using var playwright = await Playwright.CreateAsync();
+      var chromium = playwright.Chromium;
+      var context = chromium.LaunchPersistentContextAsync(@"C:\path\to\director\", headless: false);
```
pavelfeldman, line 341: director -> directory

docs/src/browsers.md
```diff
@@ -72,6 +72,19 @@ Using `firefox-stable` is a 2-steps process:
     browser = playwright.firefox.launch(channel="firefox-stable")
     ```
 
+    ```csharp
```
pavelfeldman, line 75: I'll nuke this section

docs/src/auth.md
avodovnik: Yeah, I was going to go and retrospectively fix them all in one swoop down the line. The fact it's still here is a result of me checking/compiling the samples as I write them, and I have an _old_-ish version of the API.

PR #6590, opened 2021-05-14
fix(chromium): don't cancel downloads with interception

### LINKED ISSUE #6573: [BUG] Downloads are canceled on Chromium when requests are intercepted
**Context:**
- Playwright Version: 1.8.1 and 1.11.0
- Operating System: Linux
- Node.js version: 14.15
- Browser: Chromium only, works fine on firefox and webkit
- Extra: -

**Code Snippet**

Downloads are canceled if using interception of requests

Sample of code : 
```javascript
const util = require('util');
const { chromium, firefox, webkit, devices } = require('playwright');
const options = {
    devtools: false,
    dumpio: true,
    ignoreHTTPSErrors: true,
    args: [
        '--disable-web-security',
        '--allow-running-insecure-content',
        '--disable-dev-shm-usage',
        '--no-sandbox',
        '--disable-setuid-sandbox',
        '--disable-gpu',
        '',
        '--lang=fr',

    ],
};

chromium.launch(options).then(async (browser) => {
        acceptDownloads: true
    })
    await context.route(/.*/, (route) => {
         route.continue();
    });

    page.on('download', async (download) => {
        console.log(util.inspect(download.suggestedFilename()));
        const failureMsg = await download.failure();
        const path = await download.path();

        console.log(util.inspect({
            failureMsg: failureMsg,
            path: path
        }));
    });
    await page.goto('https://www.trionsnosdechets-dijon.fr/fr/Particulier/Reduire-mes-dechets/Jardinage-et-compostage');
    const element = await page.$('[download="Guide du jardinage au naturel et du compostage.pdf"]:visible');
    await element.click();
});
```

**Describe the bug**

Using the interception of requests makes download are always canceled, even if using route.continue() as showed in my example.

Disabled it will make download works as expected : 
```
 // await context.route(/.*/, (route) => {
 //        route.continue();
 //   });
```

JoelEinbinder:
Fixes #6573

Links with a download attribute are caught by Fetch but don't send Network.requestWillBeSent events.
We were canceling these, which canceled the downloads.﻿

### DIFF
--- a/src/server/chromium/crNetworkManager.ts
+++ b/src/server/chromium/crNetworkManager.ts
@@ -165,6 +165,22 @@ export class CRNetworkManager {
       });
     }
     if (!event.networkId) {
+      // Some downloads come in as document requests without a network id.
+      // Forge a requestWillBeSent event for them so that users can handle them.
+      if (event.resourceType === 'Document') {
+        this._onRequest(workerFrame, {
+          documentURL: event.request.url,
+          initiator: {type: 'other'},
+          loaderId: '',
+          timestamp: 0,
+          wallTime: 0,
+          request: event.request,
+          requestId: event.requestId,
+          frameId: event.frameId,
+          type: event.resourceType
+        }, event);
+        return;
+      }
       // Fetch without networkId means that request was not recongnized by inspector, and
       // it will never receive Network.requestWillBeSent. Most likely, this is an internal request
       // that we can safely fail.
--- a/tests/download.spec.ts
+++ b/tests/download.spec.ts
@@ -484,4 +484,18 @@ it.describe('download event', () => {
     await page.close();
   });
 
+  it('should report downloads with interception and a download attribute', async ({browser, server}) => {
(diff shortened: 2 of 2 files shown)


src/server/chromium/crNetworkManager.ts
```diff
@@ -165,6 +165,22 @@ export class CRNetworkManager {
       });
     }
     if (!event.networkId) {
+      // Some downloads come in as document requests without a network id.
+      // Forge a requestWillBeSent event for them so that users can handle them.
+      if (event.resourceType === 'Document') {
```
dgozman, line 170: I looked at this issue. This direction seems fine, but we have to go further. For example, if we allow intercepting this request, we must ensure that it gets the corresponding request/response/requestfinished/requestfailed events with all the data inside. Basically, let's forge all the events when `networkId` is missing. Note that in the case of interception, we have to forge events after the interceptor runs.

JoelEinbinder:
Closing to rethink

PR #6593, opened 2021-05-15
docs(dotnet): examples for navigation.md, network.md, selectors.md

pavelfeldman:


docs/src/navigations.md
```diff
@@ -294,6 +341,15 @@ with page.expect_navigation():
     page.click("a")
 ```
 
+```csharp
+// Using waitForNavigation with a callback prevents a race condition
```
dgozman, line 345: This comment is very java-ish, better copy from js :)

docs/src/navigations.md
```diff
@@ -334,6 +390,15 @@ with page.expect_navigation(url="**/login"):
     page.click("a")
 ```
 
+```csharp
+// Running action in the callback of waitForNavigation prevents a race
```
dgozman, line 394: Same here.

PR #6604, opened 2021-05-16
fix(inspector): do not pause while recording

pavelfeldman:


yury-s (approved):

yury-s (commented):
Perhaps debugger can be removed from the instrumentation/recorder instead to avoid growing number of state flags?

src/server/supplements/debugger.ts
```diff
@@ -47,21 +48,33 @@ export class Debugger extends EventEmitter implements InstrumentationListener {
...
 
   async onBeforeInputAction(sdkObject: SdkObject, metadata: CallMetadata): Promise<void> {
+    if (this._muted)
+      return;
     if (this._enabled && this._pauseOnNextStatement)
       await this.pause(sdkObject, metadata);
   }
 
   async onCallLog(logName: string, message: string, sdkObject: SdkObject, metadata: CallMetadata): Promise<void> {
+    if (this._muted)
```
dgozman, line 70: I think this one should be removed - if I started with DEBUG=pw:api, let me see the logs no matter what :)

dgozman (approved):

PR #6627, opened 2021-05-18
browser(webkit): add ip address to Network.responseReceived

JoelEinbinder:
For #6624﻿

PR #6631, opened 2021-05-18
feat(har): record remote IP:PORT and SSL details

### LINKED ISSUE #6624: [Feature] HAR: IP Address and SSL Info
It would be helpful to have the remote IP address for connections as well as SSL details in the Playwright generated HAR.

Currently, we run a separate proxy to get this info for our tests.

I'd like to drop the proxy and only use Playwright's HAR. 

In regional e2e smoketests I want to assert the browser is picking certain CDN server blocks.

Additionally, I'd like to monitor connections made be embedded SDKs and ensure they only reach out to certain IP blocks (based on geoip) to ensure they are maintaining contract commitments.

Let me know priority, and I'd be happy to contribute the chrome portion of the work if someone else could assist with WK and FF. (Maybe I could do those two but I'm much less familiar.)

rwoll:
Updates in the HAR capture Log Entries:

* `serverIPAddress`: (string) This now gets populated.
* `_serverPort`: (number) This now gets populated. There isn't an
  official field according to this [w3 doc][w3] (and I didn't see
  a suitable field in a HAR download from Chromium DevTools).

  I used the field name from [WK WebInspector HARBuilder.js][wk].
* `_securityDetails`: non-standard info that is helpful in making
  assertions about cert expiries.

This also introduces new APIs on `network.Response` exposing the
new info.

NB: At first these APIs were synchronous and populated with info that is
readily available at the beginning of the Response creation, but WK does
not emit IP address info until _loadingFinished, so the API's were all
turned into Promise-based approaches.

Closes #6624.

[w3]: https://www.w3.org/community/bigdata-tools/files/2017/10/HAR_Spec_TO_HAR_Vocabulary.pdf
[wk]: https://opensource.apple.com/source/WebInspectorUI/WebInspectorUI-7608.4.9.1.3/UserInterface/Controllers/HARBuilder.js.auto.html

### DIFF
--- a/src/server/chromium/crNetworkManager.ts
+++ b/src/server/chromium/crNetworkManager.ts
@@ -296,7 +296,22 @@ export class CRNetworkManager {
         responseStart: -1,
       };
     }
-    return new network.Response(request.request, responsePayload.status, responsePayload.statusText, headersObjectToArray(responsePayload.headers), timing, getResponseBody);
+    const response = new network.Response(request.request, responsePayload.status, responsePayload.statusText, headersObjectToArray(responsePayload.headers), timing, getResponseBody);
+    if (responsePayload?.remoteIPAddress && typeof responsePayload?.remotePort === 'number')
+      response._serverAddrFinished({
+        ipAddress: responsePayload.remoteIPAddress,
+        port: responsePayload.remotePort,
+      });
+    else
+      response._serverAddrFinished();
+    response._securityDetailsFinished({
+      protocol: responsePayload?.securityDetails?.protocol,
+      subjectName: responsePayload?.securityDetails?.subjectName,
+      issuer: responsePayload?.securityDetails?.issuer,
+      validFrom: responsePayload?.securityDetails?.validFrom,
+      validTo: responsePayload?.securityDetails?.validTo,
+    });
+    return response;
   }
 
   _handleRequestRedirect(request: InterceptableRequest, responsePayload: Protocol.Network.Response, timestamp: number) {
--- a/src/server/network.ts
+++ b/src/server/network.ts
@@ -263,6 +263,19 @@ export type ResourceTiming = {
   responseStart: number;
 };
 
+export type RemoteAddr = {
+  ipAddress: string;
+  port: number;
+}
+
+export type SecurityDetails = {
+    protocol?: string;
+    subjectName?: string;
+    issuer?: string;
+    validFrom?: number;
+    validTo?: number;
+};
+
 export class Response extends SdkObject {
   private _request: Request;
   private _contentPromise: Promise<Buffer> | null = null;
@@ -275,6 +288,10 @@ export class Response extends SdkObject {
   private _headersMap = new Map<string, string>();
   private _getResponseBodyCallback: GetResponseBodyCallback;
   private _timing: ResourceTiming;
+  private _serverAddrPromise: Promise<RemoteAddr|undefined>;
+  private _serverAddrPromiseCallback: (arg?: RemoteAddr) => void = () => {};
+  private _securityDetailsPromise: Promise<SecurityDetails|undefined>;
+  private _securityDetailsPromiseCallback: (arg?: SecurityDetails) => void = () => {};
 
   constructor(request: Request, status: number, statusText: string, headers: types.HeadersArray, timing: ResourceTiming, getResponseBodyCallback: GetResponseBodyCallback) {
     super(request.frame(), 'response');
@@ -287,10 +304,24 @@ export class Response extends SdkObject {
     for (const { name, value } of this._headers)
       this._headersMap.set(name.toLowerCase(), value);
     this._getResponseBodyCallback = getResponseBodyCallback;
+    this._serverAddrPromise = new Promise(f => {
+      this._serverAddrPromiseCallback = f;
+    });
+    this._securityDetailsPromise = new Promise(f => {
+      this._securityDetailsPromiseCallback = f;
+    });
     this._finishedPromise = new Promise(f => {
       this._finishedPromiseCallback = f;
     });
     this._request._setResponse(this);
+  }
+
+  _serverAddrFinished(addr?: RemoteAddr) {
+    this._serverAddrPromiseCallback(addr);
+  }
+
+  _securityDetailsFinished(securityDetails?: SecurityDetails) {
+    this._securityDetailsPromiseCallback(securityDetails);
   }
 
   _requestFinished(responseEndTiming: number, error?: string) {
@@ -324,6 +355,14 @@ export class Response extends SdkObject {
 
   timing(): ResourceTiming {
     return this._timing;
+  }
+
+  async serverAddr(): Promise<RemoteAddr|null> {
+    return await this._serverAddrPromise || null;
+  }
+
+  async securityDetails(): Promise<SecurityDetails|null> {
+    return await this._securityDetailsPromise || null;
   }
 
   body(): Promise<Buffer> {
--- a/src/server/supplements/har/harTracer.ts
+++ b/src/server/supplements/har/harTracer.ts
@@ -209,6 +209,18 @@ export class HarTracer {
       receive,
     };
     harEntry.time = [dns, connect, ssl, wait, receive].reduce((pre, cur) => cur > 0 ? cur + pre : pre, 0);
+
+    this._addBarrier(page, response.serverAddr().then(server => {
+      if (server?.ipAddress)
+        harEntry.serverIPAddress = server.ipAddress;
+      if (server?.port)
+        harEntry._serverPort = server.port;
+    }));
+    this._addBarrier(page, response.securityDetails().then(details => {
+      if (details)
+        harEntry._securityDetails = details;
+    }));
+
     if (!this._options.omitContent && response.status() === 200) {
       const promise = response.body().then(buffer => {
         harEntry.response.content.text = buffer.toString('base64');
--- a/tests/har.spec.ts
+++ b/tests/har.spec.ts
@@ -322,7 +322,7 @@ it('should have security details', async ({ contextFactory, httpsServer, browser
   expect(serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
   expect(port).toBe(httpsServer.PORT);
   if (browserName === 'webkit' && platform === 'win32')
-    expect(securityDetails).toEqual({subjectName: 'puppeteer-tests', validFrom: 1550084863});
+    expect(securityDetails).toEqual({subjectName: 'puppeteer-tests', validFrom: 1550084863, validTo: -1});
   else if (browserName === 'webkit')
     expect(securityDetails).toEqual({protocol: 'TLS 1.3', subjectName: 'puppeteer-tests', validFrom: 1550084863, validTo: 33086084863});
   else
--- a/tests/har.spec.ts
+++ b/tests/har.spec.ts
@@ -301,3 +301,65 @@ it('should not contain internal pages', async ({ browserName, contextFactory, se
   const log = await getLog();
   expect(log.pages.length).toBe(1);
 });
+
+it('should have connection details', async ({ contextFactory, server, browserName, platform }, testInfo) => {
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  await page.goto(server.EMPTY_PAGE);
+  const log = await getLog();
+  const { serverIPAddress, _serverPort: port, _securityDetails: securityDetails } = log.entries[0];
+  expect(serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
+  expect(port).toBe(server.PORT);
+  expect(securityDetails).toEqual({});
+});
+
+it('should have security details', async ({ contextFactory, httpsServer, browserName, platform }, testInfo) => {
+  it.fail(browserName === 'webkit' && platform === 'linux', 'https://github.com/microsoft/playwright/issues/6759');
+
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  await page.goto(httpsServer.EMPTY_PAGE);
+  const log = await getLog();
+  const { serverIPAddress, _serverPort: port, _securityDetails: securityDetails } = log.entries[0];
+  expect(serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
+  expect(port).toBe(httpsServer.PORT);
+  if (browserName === 'webkit' && platform === 'win32')
+    expect(securityDetails).toEqual({subjectName: 'puppeteer-tests', validFrom: 1550084863});
+  else if (browserName === 'webkit')
+    expect(securityDetails).toEqual({protocol: 'TLS 1.3', subjectName: 'puppeteer-tests', validFrom: 1550084863, validTo: 33086084863});
+  else
+    expect(securityDetails).toEqual({issuer: 'puppeteer-tests', protocol: 'TLS 1.3', subjectName: 'puppeteer-tests', validFrom: 1550084863, validTo: 33086084863});
+});
+
+it('should have connection details for redirects', async ({ contextFactory, server, browserName }, testInfo) => {
+  server.setRedirect('/foo.html', '/empty.html');
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  await page.goto(server.PREFIX + '/foo.html');
+  const log = await getLog();
+  expect(log.entries.length).toBe(2);
+
+  const detailsFoo = log.entries[0];
+
+  if (browserName === 'webkit') {
+    expect(detailsFoo.serverIPAddress).toBeUndefined();
+    expect(detailsFoo._serverPort).toBeUndefined();
+  } else {
+    expect(detailsFoo.serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
+    expect(detailsFoo._serverPort).toBe(server.PORT);
+  }
+
+  const detailsEmpty = log.entries[1];
+  expect(detailsEmpty.serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
+  expect(detailsEmpty._serverPort).toBe(server.PORT);
+});
+
+it('should have connection details for failed requests', async ({ contextFactory, server, browserName, platform }, testInfo) => {
+  server.setRoute('/one-style.css', (_, res) => {
+    res.setHeader('Content-Type', 'text/css');
+    res.connection.destroy();
+  });
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  await page.goto(server.PREFIX + '/one-style.html');
+  const log = await getLog();
+  const { serverIPAddress, _serverPort: port } = log.entries[0];
+  expect(serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
+  expect(port).toBe(server.PORT);
+});
--- a/src/server/firefox/ffNetworkManager.ts
+++ b/src/server/firefox/ffNetworkManager.ts
@@ -89,6 +89,20 @@ export class FFNetworkManager {
       responseStart: this._relativeTiming(event.timing.responseStart),
     };
     const response = new network.Response(request.request, event.status, event.statusText, event.headers, timing, getResponseBody);
+    if (event?.remoteIPAddress && typeof event?.remotePort === 'number')
+      response._serverAddrFinished({
+        ipAddress: event.remoteIPAddress,
+        port: event.remotePort,
+      });
+    else
+      response._serverAddrFinished()
+    response._securityDetailsFinished({
+      protocol: event?.securityDetails?.protocol,
+      subjectName: event?.securityDetails?.subjectName,
+      issuer: event?.securityDetails?.issuer,
+      validFrom: event?.securityDetails?.validFrom,
+      validTo: event?.securityDetails?.validTo,
+    });
     this._page._frameManager.requestReceivedResponse(response);
   }
 
--- a/src/server/supplements/har/har.ts
+++ b/src/server/supplements/har/har.ts
@@ -55,6 +55,8 @@ export type Entry = {
   timings: Timings;
   serverIPAddress?: string;
   connection?: string;
+  _serverPort?: number;
+  _securityDetails?: SecurityDetails;
 };
 
 export type Request = {
@@ -143,4 +145,12 @@ export type Timings = {
   wait: number;
   receive: number;
   ssl?: number;
+};
+
+export type SecurityDetails = {
+  protocol?: string;
+  subjectName?: string;
+  issuer?: string;
+  validFrom?: number;
+  validTo?: number;
 };
--- a/src/server/webkit/wkPage.ts
+++ b/src/server/webkit/wkPage.ts
@@ -69,6 +69,7 @@ export class WKPage implements PageDelegate {
   _firstNonInitialNavigationCommittedReject = (e: Error) => {};
   private _lastConsoleMessage: { derivedType: string, text: string, handles: JSHandle[]; count: number, location: types.ConsoleMessageLocation; } | null = null;
 
+  private readonly _requestIdToResponseReceivedPayloadEvent = new Map<string, Protocol.Network.responseReceivedPayload>();
   // Holds window features for the next popup being opened via window.open,
   // until the popup page proxy arrives.
   private _nextWindowOpenPopupFeatures?: string[];
@@ -953,6 +954,8 @@ export class WKPage implements PageDelegate {
 
   private _handleRequestRedirect(request: WKInterceptableRequest, responsePayload: Protocol.Network.Response, timestamp: number) {
     const response = request.createResponse(responsePayload);
+    response._securityDetailsFinished();
+    response._serverAddrFinished();
     response._requestFinished(responsePayload.timing ? helper.secondsToRoundishMillis(timestamp - request._timestamp) : -1, 'Response body is unavailable for redirect responses');
     this._requestIdToRequest.delete(request._requestId);
     this._page._frameManager.requestReceivedResponse(response);
@@ -979,6 +982,7 @@ export class WKPage implements PageDelegate {
     // FileUpload sends a response without a matching request.
     if (!request)
       return;
+    this._requestIdToResponseReceivedPayloadEvent.set(request._requestId, event);
     const response = request.createResponse(event.response);
     if (event.response.requestHeaders && Object.keys(event.response.requestHeaders).length)
       request.request.updateWithRawHeaders(headersObjectToArray(event.response.requestHeaders));
@@ -1003,8 +1007,19 @@ export class WKPage implements PageDelegate {
     // Under certain conditions we never get the Network.responseReceived
     // event from protocol. @see https://crbug.com/883475
     const response = request.request._existingResponse();
-    if (response)
+    if (response) {
+      const responseReceivedPayload = this._requestIdToResponseReceivedPayloadEvent.get(request._requestId);
+      response._serverAddrFinished(parseRemoteAddress(event?.metrics?.remoteAddress));
+      response._securityDetailsFinished({
+        protocol: isLoadedSecurely(response.url(), response.timing()) ? event.metrics?.securityConnection?.protocol : undefined,
+        subjectName: responseReceivedPayload?.response.security?.certificate?.subject,
+        validFrom: responseReceivedPayload?.response.security?.certificate?.validFrom,
+        validTo: responseReceivedPayload?.response.security?.certificate?.validUntil,
+      });
       response._requestFinished(helper.secondsToRoundishMillis(event.timestamp - request._timestamp));
+    }
+
+    this._requestIdToResponseReceivedPayloadEvent.delete(request._requestId);
     this._requestIdToRequest.delete(request._requestId);
     this._page._frameManager.requestFinished(request.request);
   }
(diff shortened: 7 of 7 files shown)


src/server/webkit/wkInterceptableRequest.ts
```diff
@@ -126,7 +126,24 @@ export class WKInterceptableRequest implements network.RouteDelegate {
...
+      ipAddress: '',
+      port: -1,
+    };
+
+    let securityDetails: network.SecurityDetails | undefined;
+    if (responsePayload.security) {
+      securityDetails = {
+        protocol: responsePayload.security.connection?.protocol || '',
+        subjectName: responsePayload.security.certificate?.subject || '',
+        issuer: '',
```
rwoll, line 140: WK is the only browser to not have this. It also appears protocol is always blank for WK. I'm happy to drop everything except `validFrom`/`validTo`. Alternatively, I'm happy to pass everything through and `any` type sec details allowing it to be browser specific. (Though that did not feel in the spirit of PW.)

src/server/webkit/wkInterceptableRequest.ts
```diff
@@ -126,7 +126,24 @@ export class WKInterceptableRequest implements network.RouteDelegate {
       requestStart: timingPayload ? wkMillisToRoundishMillis(timingPayload.requestStart) : -1,
       responseStart: timingPayload ? wkMillisToRoundishMillis(timingPayload.responseStart) : -1,
     };
-    return new network.Response(this.request, responsePayload.status, responsePayload.statusText, headersObjectToArray(responsePayload.headers), timing, getResponseBody);
+
+    const serverIPAddressAndPort: network.IPAddressAndPort = {
+      ipAddress: '',
+      port: -1,
```
rwoll, line 132: TODO: This will be actual values once https://github.com/microsoft/playwright/pull/6627 lands.

rwoll (commented):
/cc @JoelEinbinder 🎉

src/server/helper.ts
```diff
@@ -116,6 +116,29 @@ class Helper {
...
+        };
+      } else { // IPv6
+        const [address, port] = value.split('.');
+        return {
+          ipAddress: `[${address}]`,
+          port: +port,
+        };
+      }
+    } catch (_) {}
+  }
```
rwoll, line 141: Is this too naïve? FF and CR give us the info split up with the proper typing, WK we have to parse it out ourselves.

src/server/supplements/har/harTracer.ts
```diff
@@ -212,6 +212,15 @@ export class HarTracer {
...
     harEntry.time = [dns, connect, ssl, wait, receive].reduce((pre, cur) => cur > 0 ? cur + pre : pre, 0);
+
+    const promise = response.finished().then(() => {
+      const serverIPAddressAndPort = response.serverIPAddressAndPort();
+      harEntry.serverIPAddress = serverIPAddressAndPort.ipAddress === '' ? undefined : serverIPAddressAndPort.ipAddress;
+      harEntry.connection = serverIPAddressAndPort.port === -1 ? undefined : serverIPAddressAndPort.port.toString();
+      harEntry.securityDetails = response.securityDetails();
+    }).catch(() => {});
+    this._addBarrier(page, promise);
+
```
rwoll, line 223: WK only gives us the IP info later in the request lifecycle. Should we change `response.serverIPAddressAndPort()` and `response.securityDetails()` to return promises that internally wait on finished? Without that, if you call those methods for WK without waiting for finished(), you'll get the placeholder values.

src/server/helper.ts
```diff
@@ -116,6 +116,29 @@ class Helper {
       return '';
     return '\n' + '='.repeat(20) + ' Browser output: ' + '='.repeat(20) + '\n' + logs.join('\n');
   }
+
+  static parseRemoteAddress(value?: string) {
```
JoelEinbinder, line 120: This method doesn't work with [::1]:80. I think parsing ipv6 addresses is somewhat complicated. I'd do `new URL('http://' + address)`. Note that if the port is empty is actually 80.

src/server/network.ts
```diff
@@ -326,6 +370,14 @@ export class Response extends SdkObject {
     return this._timing;
   }
 
+  serverIPAddressAndPort(): IPAddressAndPort {
+    return this._serverIPAddressAndPort;
+  }
+
+  securityDetails(): SecurityDetails | undefined {
+    return this._securityDetails;
+  }
```
rwoll, line 379: These might need to be change to returning Promises to accom WK's request lifecycle. Thoughts?

src/server/network.ts
```diff
@@ -275,15 +288,19 @@ export class Response extends SdkObject {
   private _headersMap = new Map<string, string>();
   private _getResponseBodyCallback: GetResponseBodyCallback;
   private _timing: ResourceTiming;
+  private _serverIPAddressAndPort: IPAddressAndPort;
+  private _securityDetails?: SecurityDetails;
```
rwoll, line 292: @JoelEinbinder Should we combine all this info? Or leave it as separate fields? Also, what's the consensus on using  making all members of SecurityDetails/IPAddressAndPort optional instead of using the `-1/''` special values.

src/server/helper.ts
rwoll: WK returns addresses like `::1.8910` which explode explode with `new URL('http://' + address)`

tests/har.spec.ts
```diff
@@ -281,3 +281,29 @@ it('should have popup requests', async ({ contextFactory, server }, testInfo) =>
   expect(entries[1].request.url).toBe(server.PREFIX + '/one-style.css');
   expect(entries[1].response.status).toBe(200);
 });
+
+it.only('should have connection details', async ({ contextFactory, server, browserName, browserMajorVersion }, testInfo) => {
```
rwoll, line 285: I'll remove this only with the next pass :)

src/server/helper.ts
JoelEinbinder: oh no :(. ok if its not a valid ip then i guess its fine

src/server/network.ts
JoelEinbinder: id prefer optional instead of special values

src/server/network.ts
JoelEinbinder: They shouldn't ever need to be promises. We would instead just delay emitting the response event.

rwoll: Is this the line that is currently triggering the emission of the response event:

https://github.com/microsoft/playwright/blob/475a417dfc29caeb644eb721cbcabd5bb8980636/src/server/webkit/wkPage.ts#L980

I don't think moving that into 

https://github.com/microsoft/playwright/blob/475a417dfc29caeb644eb721cbcabd5bb8980636/src/server/webkit/wkPage.ts#L991

works out but I also am likely misunderstanding the flow of events.

src/server/network.ts
rwoll: Done! I got to delete a lot of awkward code :)

rwoll:
@JoelEinbinder I cleaned up the types (including deleting some of the awkward code dealing with the non-standard zero values.) I also added a comment to the WebKit address parsing code with some examples from my testing.

I haven't yet figured out the appropriate changes to emit the Response at the correct time after `_onLoadingFinished` where the IP address details are available for WebKit so I left the signal barrier in place in the HAR code. I'll remove that once we sort out dealing Response emission timing (directly or indirectly through guarding the newly added fields behind a Promise api). If we can avoid adding more external Promises to Response, I agree that would be much nicer.

src/server/supplements/har/harTracer.ts
```diff
@@ -212,6 +212,15 @@ export class HarTracer {
...
     harEntry.time = [dns, connect, ssl, wait, receive].reduce((pre, cur) => cur > 0 ? cur + pre : pre, 0);
+
+    const promise = response.finished().then(() => {
+      const serverIPAddressAndPort = response.serverIPAddressAndPort();
+      harEntry.serverIPAddress = serverIPAddressAndPort?.ipAddress;
+      harEntry.connection = serverIPAddressAndPort?.port?.toString();
+      harEntry.securityDetails = response.securityDetails();
+    }).catch(() => {});
+    this._addBarrier(page, promise);
+
```
rwoll, line 223: This code should go away once we sequence the Response emission for WK differently (or use a Promised based API for the new additions.)

This iteration actually caused CR to regress due to what I think is unexpected behavior with CR response.finish(). Repro/Failing Test Case: https://github.com/microsoft/playwright/pull/6653

src/server/network.ts
rwoll: Updated per Slack comments. Everything's a Promise!

rwoll:
@JoelEinbinder Here's the latest pass! Promise-based APIs because of WK. See https://github.com/microsoft/playwright/pull/6631#issue-646501474 with updated description.

src/server/network.ts
```diff
@@ -476,3 +540,15 @@ export function mergeHeaders(headers: (types.HeadersArray | undefined | null)[])
     result.push({ name: lowerCaseToOriginalCase.get(lower)!, value });
   return result;
 }
+
+function filterEmpties(obj: { [k: string]: string|number|undefined } | undefined) {
+  if (obj)
+    return Object.entries(obj).filter(([_, v]) => v !== null && v !== undefined && (typeof v === 'number' ? v >= 0 : v !== '')).reduce((o, [prop, val]) => { const merged = (o || {}); merged[prop] = val; return merged; }, undefined as { [k: string]: string|number|undefined }|undefined);
```
rwoll, line 546: This grew organically and should be re-written as a simple for-loop for readability. Will do w/ follow up change.

rwoll:
@JoelEinbinder: This should now pass on all the browser+platform variants. (Thanks platform-dependent WK networking! 😉 )

Let me know your feedback and I will make any necessary adjustments.

It's also worth skimming through https://github.com/microsoft/playwright/pull/6695 to what additional information will be coming in next.

src/server/network.ts
```diff
@@ -263,6 +264,24 @@ export type ResourceTiming = {
   responseStart: number;
 };
 
+export type IPAddressAndPort = {
```
mxschmitt, line 267: I think we can call it RemoteAddr, at least in the Go world, this commonly get's used for ip+port.

src/server/helper.ts
```diff
@@ -116,6 +116,73 @@ class Helper {
...
+   * 2606:2800:220:1:248:1893:25c8:1946.443
+   * 127.0.0.1:8000
+   *
+   * ubuntu:
+   * ::1:8907
+   * 127.0.0.1:8000
+   *
+   * NB: They look IPv4 and IPv6's with ports but use an alternative notation.
+   */
+  static parseRemoteAddress(value?: string) {
```
mxschmitt, line 134: this get's only used in WebKit, so I would move it to the `wkPage.ts`.

src/server/network.ts
```diff
@@ -326,6 +381,16 @@ export class Response extends SdkObject {
     return this._timing;
   }
 
+  async serverIPAddressAndPort(): Promise<IPAddressAndPort|null> {
```
mxschmitt, line 384: I would call it remoteAddr there too. But let's wait for e.g. Joels opinion there.

tests/har.spec.ts
```diff
@@ -281,3 +281,33 @@ it('should have popup requests', async ({ contextFactory, server }, testInfo) =>
...
+
+it('should have security details', async ({ contextFactory, httpsServer, browserName, platform }, testInfo) => {
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  await page.goto(httpsServer.EMPTY_PAGE);
+  const log = await getLog();
+  const { serverIPAddress, _serverPort: port, _securityDetails: securityDetails } = log.entries[0];
+  expect(serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
+  expect(port).toBe(httpsServer.PORT);
+  if (browserName === 'webkit' && platform === 'linux')
+    expect(securityDetails).toBeUndefined();
```
mxschmitt, line 306: Is there a bug in the WebKit linux network stack?

src/server/network.ts
rwoll: 👍 I'll address in next changeset incorporating any comments from Joel.

tests/har.spec.ts
rwoll: TBD—I haven't yet debugged to figure out if it's intended (and simply a shortcoming of Linux WK/libsoup) or if it's just broken.

rwoll: I'm going to try to fix this in WK implementation. At first glance, it looks like it's available via libsoup, just not fully plumbed. FMI [see this Slack thread](https://playwright.slack.com/archives/C0141EZ6ZNH/p1621909162023100?thread_ts=1621906702.021200&cid=C0141EZ6ZNH).

rwoll: See https://github.com/microsoft/playwright/issues/6759.

We'll address this as a follow up, but I don't think it should block this MR.

tests/har.spec.ts
```diff
@@ -281,3 +281,33 @@ it('should have popup requests', async ({ contextFactory, server }, testInfo) =>
...
+
+it('should have connection details', async ({ contextFactory, server, browserName, platform }, testInfo) => {
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  await page.goto(server.EMPTY_PAGE);
+  const log = await getLog();
+  const { serverIPAddress, _serverPort: port, _securityDetails: securityDetails } = log.entries[0];
+  expect(serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
+  expect(port).toBe(server.PORT);
+  if (browserName === 'webkit' && platform === 'darwin')
+    expect(securityDetails).toEqual({protocol: 'SSL 2.0'});
```
JoelEinbinder, line 293: WebKit on mac thinks that 127.0.0.1 is SSL? This doesn't match what I see in Safari.

image

tests/har.spec.ts
JoelEinbinder: Instead of baking in the bad behavior here, let's mark this as failing on linux.

```ts
it.fail(browserName === 'webkit' && platform === 'linux');
```

src/server/network.ts
```diff
@@ -275,15 +294,33 @@ export class Response extends SdkObject {
...
+    if (serverIPAddressAndPort)
+      filteredConnectionDetails.serverIPAddressAndPort = serverIPAddressAndPort;
+    const securityDetails = helper.filterEmpties(connectionDetails.securityDetails);
+    if (securityDetails)
+      filteredConnectionDetails.securityDetails = securityDetails;
+    this._connectionDetails = filteredConnectionDetails;
+    this._connectionDetailsPromise = new Promise(f => {
+      this._connectionDetailsFinishedCallback = f;
+    });
+    if (connectionDetails.complete ?? true)
```
JoelEinbinder, line 321: Too fancy! This is a sign that something has gone wrong with the design.

src/server/network.ts
```diff
@@ -275,15 +294,33 @@ export class Response extends SdkObject {
   private _headersMap = new Map<string, string>();
   private _getResponseBodyCallback: GetResponseBodyCallback;
   private _timing: ResourceTiming;
+  private _connectionDetails: ConnectionDetails;
+  private _connectionDetailsFinishedCallback: (details: ConnectionDetails) => void = _ => {};
+  private _connectionDetailsPromise: Promise<ConnectionDetails>;
 
-  constructor(request: Request, status: number, statusText: string, headers: types.HeadersArray, timing: ResourceTiming, getResponseBodyCallback: GetResponseBodyCallback) {
+  constructor(request: Request, status: number, statusText: string, headers: types.HeadersArray, timing: ResourceTiming, getResponseBodyCallback: GetResponseBodyCallback, connectionDetails: ConnectionDetails & {complete?: boolean}) {
```
JoelEinbinder, line 301: Let's not pass in the connection details here, and only pass them in fully formed to _requestFinished.

src/server/helper.ts
```diff
@@ -116,6 +116,73 @@ class Helper {
...
+        continue;
+
+      filtered[k] = v;
+    }
+
+    if (Object.keys(filtered).length > 0)
+      return filtered;
+  }
+
+  static mergeUpdates(prev?: { [k: string]: string|number|undefined }, updates?: { [k: string]: string|number|undefined }) {
```
JoelEinbinder, line 180: I'm hoping that this method can be removed in the final version of this patch.

src/server/network.ts
```diff
@@ -263,6 +264,24 @@ export type ResourceTiming = {
...
+
+export type SecurityDetails = {
+    protocol?: string;
+    subjectName?: string;
+    issuer?: string;
+    validFrom?: number;
+    validTo?: number;
+};
+
+export type ConnectionDetails = {
```
JoelEinbinder, line 280: serverIPAddressAndPort and securityDetails seem to be largely independent from one another, and being merged together in this struct appears to complicate the code in this file. Maybe they are two separate top level items?

src/server/network.ts
```diff
@@ -263,6 +264,24 @@ export type ResourceTiming = {
   responseStart: number;
 };
 
+export type IPAddressAndPort = {
+  ipAddress?: string;
+  port?: number;
+}
+
+export type SecurityDetails = {
+    protocol?: string;
```
JoelEinbinder, line 273: Are these all actually optional in the final HAR? I'd expect to always have a protocol.

tests/har.spec.ts
rwoll: Oops…this should probably be marked failing for now. It looks like we are getting placeholder values (`SSL 2.0` and `TLS_NULL_WITH_NULL_NULL `) over the wire from WebKit protocol:

(scroll over all the way over to `securityConnection` section)

```json
{
  "method": "Target.dispatchMessageFromTarget",
  "params": {
    "targetId": "page-8",
    "message": "{\"method\":\"Network.loadingFinished\",\"params\":{\"requestId\":\"9.1\",\"timestamp\":0.027748466171033215,\"metrics\":{\"protocol\":\"http/1.1\",\"priority\":\"high\",\"remoteAddress\":\"127.0.0.1:8000\",\"connectionIdentifier\":\"093BB7CD-C601-4A3A-B68F-0654B44123B1\",\"requestHeaders\":{\"Accept\":\"text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8\",\"Upgrade-Insecure-Requests\":\"1\",\"Host\":\"127.0.0.1:8000\",\"User-Agent\":\"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.2 Safari/605.1.15\",\"Accept-Language\":\"en-us\",\"Accept-Encoding\":\"gzip, deflate\",\"Connection\":\"keep-alive\"},\"requestHeaderBytesSent\":354,\"requestBodyBytesSent\":0,\"responseHeaderBytesReceived\":184,\"responseBodyBytesReceived\":99,\"responseBodyDecodedSize\":99,\"securityConnection\":{\"protocol\":\"SSL 2.0\",\"cipher\":\"TLS_NULL_WITH_NULL_NULL\"}}}}"
  },
  "browserContextId": "8000000000000002",
  "pageProxyId": "7"
}
```

JoelEinbinder: This code from WebInspector might be relevant:

```js
    function loadedSecurely()
    {
        if (this.urlComponents.scheme !== "https" && this.urlComponents.scheme !== "wss" && this.urlComponents.scheme !== "sftp")
            return false;
        if (isNaN(this._timingData.secureConnectionStart) && !isNaN(this._timingData.connectionStart))
            return false;
        return true;
    }
```

src/server/network.ts
rwoll: Hehe! Totally agree, but might need a few pointers here. For some background:

1. Started simple: just passed in Security Details and Connection Details in create response…but WebKit didn't have this information until _loadingFinished
2. …so tried to move everything there…[but Chromium sometimes doesn't always fire _loadingFinished](https://github.com/microsoft/playwright/pull/6653#discussion_r636250024)…
3. …so arrived at this very awkward design that needs some help 😄 ❤️

JoelEinbinder: When chromium doesn't fire loadingFinished, we never dispatch the response event. So I don't think this patch needs to worry about that.

rwoll: Sweet—yeah, I lead myself astray (and to unideal code) by focusing on that test case! Will update.

rwoll:
@JoelEinbinder I took a quick pass to refactor to defer populating the information across all browser impls in `_onRequestFinished`. Please let me know if the general approach looks good. (i.e. save off intermediate request data in a map until we need it later in request finished.) If all looks kosher, I will clean up (i.e. ensure we're deleting out of the map not to leak/bloat mem; additionally, we likely need to do something in some of the other terminal request handlers (i.e. loading failed)—I'll add some more test cases, too.)

Unfortunately, this brings me back to why I ended up implementing this and then needing to change to the hybrid approach from the last review: _should include postData_ and _should include binary postData_ now hang because we never fire request finished. Do you have tips to ensure we don't fail on these?

src/server/webkit/wkInterceptableRequest.ts
```diff
@@ -126,6 +126,7 @@ export class WKInterceptableRequest implements network.RouteDelegate {
       requestStart: timingPayload ? wkMillisToRoundishMillis(timingPayload.requestStart) : -1,
       responseStart: timingPayload ? wkMillisToRoundishMillis(timingPayload.responseStart) : -1,
     };
+
```
rwoll, line 129: There are a number of errant extra newlines that I'll clean up. They are an artifact of refactoring within this MR.

src/server/network.ts
rwoll: Yes, everything in the Security domain in WK is marked as optional: https://github.com/WebKit/WebKit/blob/8afe31a018b11741abdf9b4d5bb973d7c1d9ff05/Source/JavaScriptCore/inspector/protocol/Security.json

And this holds in practice. See branching in the tests which I arrived at watching bot fails.

rwoll:
> @JoelEinbinder I took a quick pass to refactor to defer populating the information across all browser impls in `_onRequestFinished`. Please let me know if the general approach looks good. (i.e. save off intermediate request data in a map until we need it later in request finished.) If all looks kosher, I will clean up (i.e. ensure we're deleting out of the map not to leak/bloat mem; additionally, we likely need to do something in some of the other terminal request handlers (i.e. loading failed)—I'll add some more test cases, too.)
> 
> Unfortunately, this brings me back to why I ended up implementing this and then needing to change to the hybrid approach you saw in the last review: _should include postData_ and _should include binary postData_ now hang because we never fire request finished. Do you have tips to ensure we don't fail on these? (relates: [#6631 (comment)](https://github.com/microsoft/playwright/pull/6631#discussion_r644123722))

I suppose one approach would be to add `response._connectionDetailsFinished(serverIPAddressAndPort, securityDetails)` that FF and CR can call immediately after `createResponse`, and WK can call this in `_requestFinished(…)`.

rwoll:
> > Unfortunately, this brings me back to why I ended up implementing this and then needing to change to the hybrid approach you saw in the last review: _should include postData_ and _should include binary postData_ now hang because we never fire request finished. Do you have tips to ensure we don't fail on these? (relates: [#6631 (comment)](https://github.com/microsoft/playwright/pull/6631#discussion_r644123722))
> 
> I suppose one approach would be to add `response._connectionDetailsFinished(serverIPAddressAndPort, securityDetails)` that FF and CR can call immediately after `createResponse`, and WK can call this [before] `_requestFinished(…)`.

Here's what that might look like: https://github.com/rwoll/playwright/pull/2. @JoelEinbinder Let me know which approach you think is cleaner. I like doing everything in request finished for simplicity/unity, but I'm not sure how to compromise between Chrome and WebKit. I don't love https://github.com/rwoll/playwright/pull/2 since—while it's simple for Chrome and FF—you can accidentally hang the HAR capture in WK if the finish callbacks aren't invoked in all terminal response paths.

rwoll:
Updated with option 2 since fixing Chromium not firing loading finished is out of scope.

src/server/network.ts
```diff
@@ -287,12 +305,26 @@ export class Response extends SdkObject {
...
+      this._securityDetailsPromiseCallback = f;
+    });
     this._finishedPromise = new Promise(f => {
       this._finishedPromiseCallback = f;
     });
     this._request._setResponse(this);
   }
 
+  _serverIPAddressAndPortFinished(serverIPAddress?: IPAddressAndPort) {
+    this._serverIPAddressAndPortPromiseCallback(helper.filterEmpties(serverIPAddress));
```
JoelEinbinder, line 321: I don't think you need filterEmpties here and below. Everything is eventually going to be JSON.stringified into the har.

src/server/network.ts
```diff
@@ -263,6 +264,19 @@ export type ResourceTiming = {
   responseStart: number;
 };
 
+export type IPAddressAndPort = {
+  ipAddress?: string;
```
JoelEinbinder, line 268: We can have an ipAddress without a port? Or a port without an ip? Or maybe these can be required and the whole object becomes optional?

Not important to me either way, I just want to make sure I understand. ip without a port hurts my head.

src/server/supplements/har/harTracer.ts
```diff
@@ -212,6 +212,22 @@ export class HarTracer {
...
+        if (server?.port)
+          harEntry._serverPort = server.port;
+      }),
+      response.securityDetails().then(details => {
+        if (details)
+          harEntry._securityDetails = details;
+      }),
+    ]).then(() => {});
+    this._addBarrier(page, promise);
```
JoelEinbinder, line 229: I refactored it a bit while reading. Feel free to keep your code if you like it. However if it's functionally different to mine let me know because it means I misunderstood something. Thanks!

```suggestion
    this._addBarrier(page, response.serverIPAddressAndPort().then(server => {
      if (server?.ipAddress)
        harEntry.serverIPAddress = server.ipAddress;
      if (server?.port)
        harEntry._serverPort = server.port;
    }));
    this._addBarrier(page, response.securityDetails().then(details => {
      if (details)
        harEntry._securityDetails = details;
    }));
```

src/server/helper.ts
```diff
@@ -116,6 +116,22 @@ class Helper {
       return '';
     return '\n' + '='.repeat(20) + ' Browser output: ' + '='.repeat(20) + '\n' + logs.join('\n');
   }
+
+  static filterEmpties(obj: { [k: string]: string|number|undefined } | undefined) {
```
JoelEinbinder, line 120: This can go.

JoelEinbinder (approved):

src/server/network.ts
rwoll: Good point. I don't think I've seen a case where we only get one, so will make them both non-optional.

src/server/network.ts
rwoll: I've updated this, but we will get some empty objects for securityDetails in the HAR. (Tests updated accordingly.)

src/server/chromium/crNetworkManager.ts
```diff
@@ -332,6 +347,7 @@ export class CRNetworkManager {
     const response = request.request._existingResponse();
     if (response)
       response._requestFinished(helper.secondsToRoundishMillis(event.timestamp - request._timestamp));
+
```
mxschmitt, line 350: nit: revert this line-change.

tests/har.spec.ts
```diff
@@ -281,3 +281,65 @@ it('should have popup requests', async ({ contextFactory, server }, testInfo) =>
...
 });
+
+it('should have connection details', async ({ contextFactory, server, browserName, platform }, testInfo) => {
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  await page.goto(server.EMPTY_PAGE);
+  const log = await getLog();
+  const { serverIPAddress, _serverPort: port, _securityDetails: securityDetails } = log.entries[0];
+  expect(serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
+  expect(port).toBe(server.PORT);
+  expect(securityDetails).toEqual({});
```
mxschmitt, line 292: should we maybe make it that it returns null if there are no security details / its no secure connection?

mxschmitt (approved):
Awesome work! (don't forget to rebase)

tests/har.spec.ts
rwoll: That's what `filterEmpties` was doing before (well, `undefined` instead of `null`). I'll bring that behavior back without adding back `filterEmpties`.

rwoll:
All feedback addressed! Thanks for the reviews @JoelEinbinder and @mxschmitt!

tests/har.spec.ts
mxschmitt: Hmm, sorry my bad. I was confused since I thought its a method which users will consume like PPTR [has](https://pptr.dev/#?product=Puppeteer&version=v10.0.0&show=api-httpresponsesecuritydetails). Feel free to do any now what is compliant with the spec.

rwoll: This section does not have a spec (or much precedent) to my knowledge, so it comes down to what we prefer. I've put it back to the empty object `{}` since that simplifies our code and saves you from doing an extra null/undefined check when consuming the HAR. (This is a non-issue in JS/TS with the details?.foo notation, but for Python it's nice to assume the object is there and just check for a property. ie. just `details.get("foo")` and not `(deatils || {}).get("foo")`)

This does, however mean, there will be a bunch of empty objects in the har for non-secure connections.

/cc @JoelEinbinder

tests/har.spec.ts
```diff
@@ -322,7 +322,7 @@ it('should have security details', async ({ contextFactory, httpsServer, browser
   expect(serverIPAddress).toMatch(/^127\.0\.0\.1|\[::1\]/);
   expect(port).toBe(httpsServer.PORT);
   if (browserName === 'webkit' && platform === 'win32')
-    expect(securityDetails).toEqual({subjectName: 'puppeteer-tests', validFrom: 1550084863});
+    expect(securityDetails).toEqual({subjectName: 'puppeteer-tests', validFrom: 1550084863, validTo: -1});
```
rwoll, line 325: Not sure what the best approach is here…should I just mark this test as fully failing on Win32? `filterEmpties` used to filter out the -1's we were getting from WK.

PR #6646, opened 2021-05-18
feat: added reduced motion media query emulation

### LINKED ISSUE #3320: [Feature] Emulate reduced motion media query
Puppeteer allows emulating the reduced motion media query: https://github.com/puppeteer/puppeteer/blob/v5.2.1/docs/api.md#pageemulatemediafeaturesfeatures

It would be great if we can support this in playwright as well. It's widely supported in browsers: https://caniuse.com/#feat=prefers-reduced-motion

mxschmitt:
Fixes #3320

### DIFF
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -1006,7 +1006,7 @@ class FrameSession {
   async _updateEmulateMedia(initial: boolean): Promise<void> {
     if (this._crPage._browserContext._browser.isClank())
       return;
-    const colorScheme = this._page._state.colorScheme || this._crPage._browserContext._options.colorScheme || 'light';
+    const colorScheme = this._page._state.colorScheme === null ? '' : this._page._state.colorScheme;
     const reducedMotion = this._page._state.reducedMotion === null ? '' : this._page._state.reducedMotion;
     const features = [
       { name: 'prefers-color-scheme', value: colorScheme },
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -1007,11 +1007,12 @@ class FrameSession {
     if (this._crPage._browserContext._browser.isClank())
       return;
     const colorScheme = this._page._state.colorScheme || this._crPage._browserContext._options.colorScheme || 'light';
-    const reducedMotion = this._page._state.reducedMotion || this._crPage._browserContext._options.reducedMotion || 'no-preference';
+    const reducedMotion = this._page._state.reducedMotion === null ? '' : this._page._state.reducedMotion;
     const features = [
       { name: 'prefers-color-scheme', value: colorScheme },
       { name: 'prefers-reduced-motion', value: reducedMotion },
     ];
+    // Empty string disables the override.
     await this._client.send('Emulation.setEmulatedMedia', { media: this._page._state.mediaType || '', features });
   }
 
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -1007,7 +1007,11 @@ class FrameSession {
     if (this._crPage._browserContext._browser.isClank())
       return;
     const colorScheme = this._page._state.colorScheme || this._crPage._browserContext._options.colorScheme || 'light';
-    const features = colorScheme ? [{ name: 'prefers-color-scheme', value: colorScheme }] : [];
+    const reducedMotion = this._page._state.reducedMotion || this._crPage._browserContext._options.reducedMotion || 'no-preference';
+    const features = [
+      { name: 'prefers-color-scheme', value: colorScheme },
+      { name: 'prefers-reduced-motion', value: reducedMotion },
+    ];
     await this._client.send('Emulation.setEmulatedMedia', { media: this._page._state.mediaType || '', features });
   }
 
--- a/src/server/firefox/ffBrowser.ts
+++ b/src/server/firefox/ffBrowser.ts
@@ -198,10 +198,14 @@ export class FFBrowserContext extends BrowserContext {
       promises.push(this.setGeolocation(this._options.geolocation));
     if (this._options.offline)
       promises.push(this.setOffline(this._options.offline));
-    if (this._options.colorScheme)
-      promises.push(this._browser._connection.send('Browser.setColorScheme', { browserContextId, colorScheme: this._options.colorScheme }));
-    if (this._options.reducedMotion)
-      promises.push(this._browser._connection.send('Browser.setReducedMotion', { browserContextId, reducedMotion: this._options.reducedMotion }));
+    promises.push(this._browser._connection.send('Browser.setColorScheme', {
+      browserContextId,
+      colorScheme: this._options.colorScheme !== undefined  ? this._options.colorScheme : 'light',
+    }));
+    promises.push(this._browser._connection.send('Browser.setReducedMotion', {
+      browserContextId,
+      reducedMotion: this._options.reducedMotion !== undefined  ? this._options.reducedMotion : 'no-preference',
+    }));
     if (this._options.recordVideo) {
       promises.push(this._ensureVideosPath().then(() => {
         return this._browser._connection.send('Browser.setVideoRecordingOptions', {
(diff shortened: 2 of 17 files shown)


src/server/chromium/crPage.ts
```diff
@@ -1007,7 +1007,11 @@ class FrameSession {
     if (this._crPage._browserContext._browser.isClank())
       return;
     const colorScheme = this._page._state.colorScheme || this._crPage._browserContext._options.colorScheme || 'light';
-    const features = colorScheme ? [{ name: 'prefers-color-scheme', value: colorScheme }] : [];
+    const reducedMotion = this._page._state.reducedMotion || this._crPage._browserContext._options.reducedMotion || 'no-preference';
+    const features = [
+      { name: 'prefers-color-scheme', value: colorScheme },
+      { name: 'prefers-reduced-motion', value: reducedMotion },
```
dgozman, line 1013: How does this reset to system default? Or do we always want 'no-preference' by default?

src/server/firefox/ffBrowser.ts
```diff
@@ -200,6 +200,8 @@ export class FFBrowserContext extends BrowserContext {
       promises.push(this.setOffline(this._options.offline));
     if (this._options.colorScheme)
       promises.push(this._browser._connection.send('Browser.setColorScheme', { browserContextId, colorScheme: this._options.colorScheme }));
+    if (this._options.reducedMotion)
```
dgozman, line 203: For some reason, logic here is different - we don't always send the value and rely on system default.

src/server/page.ts
```diff
@@ -359,15 +361,19 @@ export class Page extends SdkObject {
     }), this._timeoutSettings.navigationTimeout(options));
   }
 
-  async emulateMedia(options: { media?: types.MediaType | null, colorScheme?: types.ColorScheme | null }) {
+  async emulateMedia(options: { media?: types.MediaType | null, colorScheme?: types.ColorScheme | null, reducedMotion?: types.ReducedMotion | null }) {
     if (options.media !== undefined)
```
dgozman, line 365: It is safe to drop these checks, client should take care of it. And if it does not, protocol validator will yell anyway.

src/server/webkit/wkPage.ts
```diff
@@ -181,7 +181,7 @@ export class WKPage implements PageDelegate {
     if (contextOptions.userAgent)
       promises.push(session.send('Page.overrideUserAgent', { value: contextOptions.userAgent }));
     if (this._page._state.mediaType || this._page._state.colorScheme)
```
dgozman, line 183: Missing `|| this._page._state.reducedMotion`

src/server/webkit/wkPage.ts
```diff
@@ -591,6 +591,14 @@ export class WKPage implements PageDelegate {
       }
       promises.push(session.send('Page.setForcedAppearance', { appearance }));
     }
+    if (reducedMotion !== null) {
```
dgozman, line 594: Interestingly, we never reset to "system default" in this method for both reduced motion and color scheme.

tests/page/page-emulate-media.spec.ts
```diff
@@ -114,3 +114,15 @@ it('should change the actual colors in css', async ({page}) => {
   await page.emulateMedia({ colorScheme: 'light' });
   expect(await getBackgroundColor()).toBe('rgb(255, 255, 255)');
 });
+
+it('should emulate reduced motion', async ({page}) => {
```
dgozman, line 118: Let's have a test for default behavior, if we decide to not fallback to system default.

dgozman (commented):
I think we should figure out whether we want default behavior be system-dependent, or hardcoded. Depending on that, quite a few places will have to change.

.github/workflows/tests_secondary.yml
```diff
@@ -94,6 +94,8 @@ jobs:
         DEBUG: pw:install
     - run: npm run build
     - run: node lib/cli/cli install-deps
+    - name: Setup tmate session
```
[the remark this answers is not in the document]
mxschmitt: Was for debugging the failing win test, will remove it.

src/server/chromium/crPage.ts
mxschmitt: Passing null should reset it to system defaults. Hard to test since browsers implement the detection differently, some do it only for windows and some not.
I will test it and keep you updated.

src/server/firefox/ffBrowser.ts
mxschmitt: It's related to the question if we want no-preference by default, I think we should to have reproducible tests across operating systems. Otherwise it will end up like on the color scheme side to follow up with defaults.

src/server/page.ts
mxschmitt: So which types instead?a

src/server/webkit/wkPage.ts
mxschmitt: Good catch.

src/server/webkit/wkPage.ts
mxschmitt: Good catch, probably hard to add there a system-test since GitHub Action environments are all quite different.

tests/page/page-emulate-media.spec.ts
mxschmitt: how shall we test it?

src/server/page.ts
```diff
@@ -160,6 +161,7 @@ export class Page extends SdkObject {
       emulatedSize: browserContext._options.viewport ? { viewport: browserContext._options.viewport, screen: browserContext._options.screen || browserContext._options.viewport } : null,
       mediaType: null,
       colorScheme: null,
```
dgozman, line 163: Let's do the same here - default to 'light'.

src/server/page.ts
dgozman: Not sure what you mean with the last question.My point is that we can remove lines 365-370 in the new code, and all tests will still pass.

src/server/chromium/crPage.ts
```diff
@@ -1007,7 +1007,12 @@ class FrameSession {
     if (this._crPage._browserContext._browser.isClank())
       return;
     const colorScheme = this._page._state.colorScheme || this._crPage._browserContext._options.colorScheme || 'light';
```
dgozman, line 1009: Let's do the same here, use `page._state` only, and pass empty string for null.

src/server/firefox/ffPage.ts
```diff
@@ -353,10 +353,12 @@ export class FFPage implements PageDelegate {
 
   async updateEmulateMedia(): Promise<void> {
     const colorScheme = this._page._state.colorScheme || this._browserContext._options.colorScheme || 'light';
```
dgozman, line 355: Same here.

src/server/webkit/wkPage.ts
```diff
@@ -591,6 +591,13 @@ export class WKPage implements PageDelegate {
       }
       promises.push(session.send('Page.setForcedAppearance', { appearance }));
```
dgozman, line 592: Let's fix the bug here as well - we have to send this command even when colorScheme changes from something to null.

src/server/webkit/wkPage.ts
```diff
@@ -591,6 +591,13 @@ export class WKPage implements PageDelegate {
...
     }
+    let reducedMotionWk: any = '';
```
dgozman, line 594: I'd just default to `undefined` here and drop the last case.

src/server/webkit/wkPage.ts
```diff
@@ -610,7 +617,8 @@ export class WKPage implements PageDelegate {
 
   async updateEmulateMedia(): Promise<void> {
     const colorScheme = this._page._state.colorScheme || this._browserContext._options.colorScheme || 'light';
```
dgozman, line 619: Do the same here.

tests/page/page-emulate-media.spec.ts
dgozman: Just match `no-preference` before the first `emulateMedia` call.

dgozman (approved):
This looks good, let's see whether the tests pass :)

PR #6653, opened 2021-05-19
test: chromium fails to resolve response.finished() on 404 page

rwoll:


rwoll:
It looks like we hit…

```
_onResponseReceived - /empty.html
_onLoadingFinished - /empty.html
_onResponseReceived - /post
…missing loading finished for /post?
```

rwoll:
Here it is: https://github.com/microsoft/playwright/issues/6666

tests/page/page-wait-for-response.spec.ts
```diff
@@ -113,3 +113,13 @@ it('should work with no timeout', async ({page, server}) => {
   ]);
   expect(response.url()).toBe(server.PREFIX + '/digits/2.png');
 });
+
+it('should resolve response.finished() 404 page', async ({ page, server, browserName }) => {
+  it.fixme(browserName === 'chromium');
+  await page.goto(server.EMPTY_PAGE);
+  const [response] = await Promise.all([
+    page.waitForResponse('**/404'),
+    page.evaluate(() => fetch('./404'))
```
dgozman, line 122: This should be `fetch('./404').then(r => r.text())`. Unless you fetch the body, the response never finishes because the browser just does not read from the network.

With that change, this test passes for me in all browsers.

rwoll: Thanks, @dgozman! That makes a lot of sense. I was incorrectly assuming the three browsers would behave the same with this test, but it looks like they have divergent behavior (but not a PW bug).

PR #6657, opened 2021-05-19
chore: install all FF system dependencies with --full on build

mxschmitt:
Before that it actually didn't install all needed dependencies and failed after that when starting the compiler suite. WebKit also installs dependencies with apt, so it should be fine to install them on OS level.


yury-s (approved):
Hopefully there are no conflicts with webkit deps :grin:

mxschmitt:
> Hopefully there are no conflicts with webkit deps

Note: tested on my environment (new Azure 20.04 VM) and both compiled successfully.

aslushnikov (approved):

PR #6663, opened 2021-05-19
devops: restructure test workflows

aslushnikov:
This patch introduces multiple workflows with unique names. This way
we might be able to use environment protection feature to schedule
workflow runs manually, but more importantly we can re-start a subset of jobs
when they fail, thus saving on the concurrency. 

Note: all workflows will still be triggered with the `CQ1` label.


yury-s (approved):
Not sure if the extra duplication of the configs is worth it.

mxschmitt:
Seems not relevant anymore.

PR #6693, opened 2021-05-21
docs: add TypeScript $eval type-hint notes

rwoll:
Some of the examples do not compile in TS as is, so this adds a hint
about what to do in TS.

This is based on feedback from a [Slack discussion][slack].

[slack]: https://playwright.slack.com/archives/CSUHZPVLM/p1621489888069900?thread_ts=1621486497.064200&cid=CSUHZPVLM

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -995,6 +995,8 @@ Examples:
 const searchValue = await page.$eval('#search', el => el.value);
 const preloadHref = await page.$eval('link[rel=preload]', el => el.href);
 const html = await page.$eval('.main-container', (e, suffix) => e.outerHTML + suffix, 'hello');
+// In TypeScript, this example requires an explicit type annotation (HTMLLinkElement) on el:
+const preloadHrefTS = await page.$eval('link[rel=preload]', (el: HTMLLinkElement) => el.href);
 ```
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):

docs/src/api/class-page.md
```diff
@@ -989,6 +989,9 @@ Examples:
 const searchValue = await page.$eval('#search', el => el.value);
 const preloadHref = await page.$eval('link[rel=preload]', el => el.href);
 const html = await page.$eval('.main-container', (e, suffix) => e.outerHTML + suffix, 'hello');
+// In TypeScript, it is necessary to add an explicit type hint if you are not using a simple tag selector
```
dgozman, line 992: I'd prefer these examples to use the same selector as above. Like, `select#choose-options` or `link[rel=preload]` as above. I'd also skip the "using a simple tag selector" part since it just works. WDYT?

rwoll: Updated! I removed the simple tag example. The intent was to to show you don't always need it; I don't want people to see the type-hint docs and assume they always need it (and then potentially also give wrong hints). However, I suppose including the simple case, doesn't help communicate that unless you're already thinking about it.

PR #6695, opened 2021-05-21
feat(har): record sizes, compression and httpVersion

tnolet:
This PR attempts to update a set of missing properties in the HAR file, where earlier a default -1 was reported and also report the correct HTTP Version, which was hard coded to `HTTP/1.1`

The test in `tests/har.spec.ts:244` should show what is being done.
- `bodySize`
- `headersSize`
- `httpVersion`
- `_transferSize`: note this is not an official property, but used by Chrome.
- `content.size`
- `content.compression`

The biggest change is that the fields at the end of the request/response cycle are now done on `Page.Events.onRequestFinished` instead of `Page.Events.onResponse`.

## Note 1
This does not work yet for Firefox as no `transferSize` like property is available. However, we can probably make that happen with some work on the NetworkObserver.js patches.

## Note 2
I went down the rabbit hole of solving this problem by populating the various `size` using the `performance.getEntriesByName()` API from the browser. However, CORS settings makes this unreliable.

### DIFF
--- a/src/server/supplements/har/harTracer.ts
+++ b/src/server/supplements/har/harTracer.ts
@@ -21,6 +21,7 @@ import { helper } from '../../helper';
 import * as network from '../../network';
 import { Page } from '../../page';
 import * as har from './har';
+import * as types from '../../types';
 
 const fsWriteFileAsync = util.promisify(fs.writeFile.bind(fs));
 
@@ -68,7 +69,7 @@ export class HarTracer {
     this._pageEntries.set(page, pageEntry);
     this._log.pages.push(pageEntry);
     page.on(Page.Events.Request, (request: network.Request) => this._onRequest(page, request));
-    page.on(Page.Events.Response, (response: network.Response) => this._onResponse(page, response));
+    page.on(Page.Events.RequestFinished, (request: network.Request) => this._onRequestFinished(page, request));
 
     page.on(Page.Events.DOMContentLoaded, () => {
       const promise = page.mainFrame().evaluateExpression(String(() => {
@@ -120,18 +121,18 @@ export class HarTracer {
       request: {
         method: request.method(),
         url: request.url(),
-        httpVersion: 'HTTP/1.1',
+        httpVersion: 'http/1.1',
         cookies: [],
         headers: [],
         queryString: [...url.searchParams].map(e => ({ name: e[0], value: e[1] })),
-        postData: undefined,
+        postData: postDataForHar(request) || undefined,
         headersSize: -1,
         bodySize: -1,
       },
       response: {
         status: -1,
         statusText: '',
-        httpVersion: 'HTTP/1.1',
+        httpVersion: 'http/1.1',
         cookies: [],
         headers: [],
         content: {
@@ -140,7 +141,8 @@ export class HarTracer {
         },
         headersSize: -1,
         bodySize: -1,
-        redirectURL: ''
+        redirectURL: '',
+        _transferSize: -1
       },
       cache: {
         beforeRequest: null,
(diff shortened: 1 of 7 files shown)


src/server/supplements/har/harTracer.ts
```diff
@@ -68,7 +70,7 @@ export class HarTracer {
     this._pageEntries.set(page, pageEntry);
     this._log.pages.push(pageEntry);
     page.on(Page.Events.Request, (request: network.Request) => this._onRequest(page, request));
-    page.on(Page.Events.Response, (response: network.Response) => this._onResponse(page, response));
+    page.on(Page.Events.RequestFinished, (request: network.Request) => this._onRequestFinished(page, request));
```
dgozman, line 73: Not all requests finish, some of them stall, and others fail. I think we should update half of the fields on Response event, and another half on RequestFinished/RequestFailed. WDYT?

src/server/supplements/har/harTracer.ts
```diff
@@ -152,38 +155,50 @@ export class HarTracer {
...
+    const harEntry = this._entries.get(request)!;
     // Rewrite provisional headers with actual
-    const request = response.request();
     harEntry.request.headers = request.headers().map(header => ({ name: header.name, value: header.value }));
     harEntry.request.cookies = cookiesForHar(request.headerValue('cookie'), ';');
-    harEntry.request.postData = postDataForHar(request) || undefined;
+
+    const response = await request.response();
+
+    if (!response) return;
```
dgozman, line 172: style: we usually return on a separate line.

src/server/supplements/har/harTracer.ts
```diff
@@ -152,38 +155,50 @@ export class HarTracer {
...
+
+    const httpVersion = response.protocol() ? response.protocol() : FALLBACK_HTTP_VERSION;
```
dgozman, line 174: There is already a fallback in the `protocol()` getter. Let's have it in one place (I prefer this one).

src/server/webkit/wkPage.ts
```diff
@@ -966,8 +966,11 @@ export class WKPage implements PageDelegate {
     // Under certain conditions we never get the Network.responseReceived
     // event from protocol. @see https://crbug.com/883475
     const response = request.request._existingResponse();
-    if (response)
+    if (response) {
```
dgozman, line 969: Do we have `protocol` available in WebKit?

src/server/supplements/har/harTracer.ts
```diff
@@ -152,38 +155,50 @@ export class HarTracer {
...
+    const transferSize = response.encodedDataLength();
+    const headersSize = calcResponseHeadersSize(httpVersion, response.status(), response.statusText(), response.headers());
```
dgozman, line 176: Note that in Chromium we have `headersText` arriving in `Network.responseReceivedExtraInfo`, so that you don't have to calculate it. Let's add a TODO to use that.

src/server/supplements/har/harTracer.ts
```diff
@@ -120,27 +122,28 @@ export class HarTracer {
...
         url: request.url(),
-        httpVersion: 'HTTP/1.1',
+        httpVersion: FALLBACK_HTTP_VERSION,
         cookies: [],
         headers: [],
         queryString: [...url.searchParams].map(e => ({ name: e[0], value: e[1] })),
-        postData: undefined,
+        postData: postDataForHar(request) || undefined,
         headersSize: -1,
         bodySize: -1,
```
dgozman, line 131: Do you want to set `bodySize` here as well? See how DevTools does it here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/models/har/Log.ts;l=378-388?q=f:front_end%20-f:out%2F%20responseHeadersText.

dgozman (commented):
Looks good overall, left a few comments.

src/server/supplements/har/harTracer.ts
tnolet: yes, this is a correct observation. Somehow it is feels messy though. Reading the code it's not super clear what gets done when. I do think we have no other way, this is how HTTP / Browser networking is.

Testing this is tricky. We would need a local server for tests that does not respond or stalls some requests.

src/server/supplements/har/harTracer.ts
tnolet: gotcha

src/server/supplements/har/harTracer.ts
tnolet: @dgozman do you mean: "switch to my `FALLBACK_HTTP_VERSION`" or do you mean "keep current fallback"?

src/server/supplements/har/harTracer.ts
dgozman: I think you can just add a test with a failed request and see that some information is there. I do not think that extra testing complexity will be worth it in this particular case.

src/server/supplements/har/harTracer.ts
dgozman: I mean "make `response.protocol()` nullable" and keep `FALLBACK_HTTP_VERSION` here.

src/server/webkit/wkPage.ts
tnolet: Good catch, we need to add that at `src/server/webkit/wkInterceptableRequest.ts:113`

tnolet: The issue is it only becomes available later in the cycle. The `metrics` property (which holds the `protocol` gets populated in the `loadingFinishedPayload` at `src/server/webkit/protocol.ts:5475`

src/server/supplements/har/harTracer.ts
tnolet: aha, e.g.

```
 // Response class
  protocol(): string | undefined {
    return this._protocol;
  }
```

tnolet: ok, this was an easy fix. Pushed it already

src/server/webkit/wkPage.ts
tnolet: @dgozman this was actually not that hard to fix. I added a patch for it in the last commit. We are now using the http protocol reported by webkit.

src/server/supplements/har/harTracer.ts
tnolet: @dgozman I think a TODO would be fine, right now my solution is useful as it is uniform for all supported browsers.

aslushnikov:
Hey @tnolet! What's the status of this? 

For the firefox, it'll be very easy to export transfer size: https://ffsearch.azurewebsites.net/#path=%2Fhome%2Fjoe%2Fplaywright%2Fbrowser_patches%2Ffirefox%2Fcheckout%2Fnetwerk%2Fprotocol%2Fhttp%2FnsIHttpChannel.idl&line=90

I can help with this if needed!

tnolet:
> Hey @tnolet! What's the status of this?
> 
> For the firefox, it'll be very easy to export transfer size: https://ffsearch.azurewebsites.net/#path=%2Fhome%2Fjoe%2Fplaywright%2Fbrowser_patches%2Ffirefox%2Fcheckout%2Fnetwerk%2Fprotocol%2Fhttp%2FnsIHttpChannel.idl&line=90
> 
> I can help with this if needed!

@aslushnikov I think we are pretty damn far with this. It would be great if you can help patch FF so we can keep this PR simple and not make any exemptions for FF. How should we coordinate this? I see the following steps:

- have FF export transferSizes
- integrate the FF tranferSizes so they are available to the harTracer.
- fix any merge conflicts
- profit?

mxschmitt:
Closing since https://github.com/microsoft/playwright/pull/7470 is merged.

PR #6714, opened 2021-05-24
test: skip from toImpl fixture

dgozman:
This way we don't need `skip(mode !== 'default')` everywhere.


tests/inspector/cli-codegen-2.spec.ts
```diff
@@ -18,8 +18,6 @@ import { test, expect } from './inspectorTest';
 import * as url from 'url';
 
 test.describe('cli codegen', () => {
-  test.skip(({ mode }) => mode !== 'default');
-
   test('should contain open page', async ({ openRecorder }) => {
```
aslushnikov, line 21: where will this test be skipped?

aslushnikov (approved):

PR #6717, opened 2021-05-24
docs(dotnet): rename webkit to WebKit

avodovnik:
Proposal to change this to keep the name "correct" as `WebKit`.


avodovnik:
Following @mxschmitt's proposal to do this... I thought about renaming it to `webKit`, but that would require aliases for *all* the other languages so I felt the C# one makes more sense. Happy to flip it around. @pavelfeldman what do you think?

mxschmitt:
Seems stale now and we can't change it anymore since its v1.0.

PR #6724, opened 2021-05-24
docs(class-page.md): Add additional clarification on requestFailed event

gboer:
The current note makes it look like only special responses are considered 'successful' requests, but any request that completes - no matter how - will never show up in the requestFailed event.

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -422,7 +422,8 @@ Emitted when a request fails, for example by timing out.
 
 :::note
 HTTP Error responses, such as 404 or 503, are still successful responses from HTTP standpoint, so request will complete
-with [`event: Page.requestFinished`] event and not with [`event: Page.requestFailed`].
+with [`event: Page.requestFinished`] event and not with [`event: Page.requestFailed`]. A request will only be considered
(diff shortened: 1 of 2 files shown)


docs/src/api/class-page.md
```diff
@@ -422,7 +422,8 @@ Emitted when a request fails, for example by timing out.
 
 :::note
 HTTP Error responses, such as 404 or 503, are still successful responses from HTTP standpoint, so request will complete
-with [`event: Page.requestFinished`] event and not with [`event: Page.requestFailed`].
+with [`event: Page.requestFinished`] event and not with [`event: Page.requestFailed`]. A request will only be considered
+failed when the client cannot get a response from the server like net::ERR_FAILED.
```
yury-s, line 426: How about rephrasing it to "...cannot get an HTTP response from the server, e.g. due to network error net::ERR_FAILED." ?

gboer: sounds good, I'll update it :)

gboer: done!

yury-s (approved):

yury-s:
Please run `npm run lint` to update types/types.d.ts, it should fix the lint errors.

gboer:
Thanks, should be good now.

PR #6729, opened 2021-05-24
docs(installation): use RFC5735 IPs for examples

rwoll:
Updates examples to use IPs from [TEST-NET-* blocks][rfc] designated for
example documentation instead of likely live (albeit innocuous) router IPs.

[rfc]: https://datatracker.ietf.org/doc/html/rfc5735

### DIFF
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -154,43 +154,43 @@ resources. In this case, Playwright can be configured to download browsers via a
 
 ```sh js
 # Linux/macOS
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):

mxschmitt:
I unfortunately changed this part a few days ago, would be awesome if you could rebase, then we can land it.

rwoll:
@mxschmitt rebased! 🎊

PR #6735, opened 2021-05-25
feat: do not rely on chocolatey to install Google Chrome Beta

aslushnikov:


utils/install-chrome-beta/reinstall_win.ps1
```diff
@@ -0,0 +1,22 @@
+$url = 'https://dl.google.com/tag/s/dl/chrome/install/beta/googlechromebetastandaloneenterprise.msi';
```
yury-s, line 1: Drop this or remove `else` branch from the if below

yury-s (approved):

PR #6745, opened 2021-05-25
fix(port-forwarding): on WebKit Win

mxschmitt:
Should fix the red bots on master, will follow-up with Mac fixes, needs to be in WK.

### DIFF
--- a/tests/portForwardingServer.spec.ts
+++ b/tests/portForwardingServer.spec.ts
@@ -20,7 +20,7 @@ import { contextTest as it, expect } from './config/browserTest';
 import type { LaunchOptions, ConnectOptions } from '../index';
 
 it.skip(({ mode }) => mode !== 'default');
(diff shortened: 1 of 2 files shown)


tests/portForwardingServer.spec.ts
```diff
@@ -20,6 +20,7 @@ import { contextTest as it, expect } from './config/browserTest';
 import type { LaunchOptions, ConnectOptions } from '../index';
 
 it.skip(({ mode }) => mode !== 'default');
+it.fixme(({platform}) => platform === 'darwin');
```
dgozman, line 23: Why not `&& browserName === 'webkit'`?

dgozman (approved):

tests/portForwardingServer.spec.ts
mxschmitt: forgot that, good catch :+1:

PR #6753, opened 2021-05-26
fix(port-forwarding): close socket on unexpected payloads

mxschmitt:
Before: it terminated the Node.js script, now it disconnects the socket and writes the stack-trace to the debug-logger.

### DIFF
--- a/src/server/socksServer.ts
+++ b/src/server/socksServer.ts
@@ -15,7 +15,7 @@
  */
 import net from 'net';
 
-import { assert } from '../utils/utils';
+import { debugLogger } from '../utils/debugLogger';
 import { SdkObject } from './instrumentation';
 
 export type SocksConnectionInfo = {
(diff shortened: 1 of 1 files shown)


src/server/socksServer.ts
```diff
@@ -96,10 +96,16 @@ class SocksV5ServerParser {
...
+      if (!condition) {
+        socket.end();
+        debugLogger.log('proxy', `Closed socket caused by false assertion: ${new Error().stack}`);
+      }
+    }
     while (i < chunk.length && this._phase !== ConnectionPhases.DONE) {
       switch (this._phase) {
         case ConnectionPhases.VERSION:
-          assert(readByte() === SOCKS_VERSION);
+          assertAndClose(readByte() === SOCKS_VERSION);
```
dgozman, line 108: I think you should also return from this function, or we'll get a lot of errors following the assertion, or perhaps even an infinite cycle.

src/server/socksServer.ts
```diff
@@ -96,10 +96,15 @@ class SocksV5ServerParser {
     const socket = this._socket;
     let i = 0;
     const readByte = () => chunk[i++];
+    const closeSocketOnError = () => {
+      socket.end();
+      debugLogger.log('proxy', `Closed Socket due error: ${new Error().stack}`);
```
dgozman, line 101: I wonder whether we should also resolve or reject `this._ready`? It depends on how we call it.

dgozman (approved):

PR #6761, opened 2021-05-26
api(dotnet): specialize waitForEvent

pavelfeldman:


docs/src/api/class-page.md
```diff
@@ -2993,7 +2994,7 @@ frame = event_info.value
 ```
 
 ```csharp
-var frame = await page.RunAndWaitForEventAsync(PageEvent.FrameNavigated, async () =>
+var frame = await page.RunAndWaitForFrameNavigatedAsync(async () =>
```
dgozman, line 2997: This is not a thing, right?

dgozman (approved):

docs/src/api/class-page.md
pavelfeldman: true

PR #6765, opened 2021-05-26
chore(types): Add missing return types to outofprocess.ts

thomaschaplin:
Added missing return types to the functions inside outofprocess.ts


thomaschaplin:
@mxschmitt I will close this PR too

PR #6769, opened 2021-05-27
chore(eval): merge internal evaluate functions

JoelEinbinder:
Our internal evaluations were getting pretty ugly with so many methods and arguments. I tried to consolidate
most of them into a method `eval`﻿that takes an options object with default values.

I think there is room to do more with some waitForSignals methods, but those have a strange relationship
with slowmo that I want to figure out in a separate patch.

### DIFF
--- a/src/server/frames.ts
+++ b/src/server/frames.ts
@@ -507,7 +507,7 @@ export class Frame extends SdkObject {
     this._nonStallingEvaluations.add(callback);
     try {
       return await Promise.race([
-        context.evaluateExpression(expression, isFunction),
+        context.eval(expression, { isFunction }),
         frameInvalidated
       ]);
     } finally {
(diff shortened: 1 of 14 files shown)


src/server/frames.ts
```diff
@@ -636,25 +636,10 @@ export class Frame extends SdkObject {
     return this._context('utility');
   }
 
-  async evaluateExpressionHandleAndWaitForSignals(expression: string, isFunction: boolean | undefined, arg: any, world: types.World = 'main'): Promise<any> {
+  async eval(pageFunction: string|Function, options: Parameters<js.ExecutionContext['eval']>[1] & {world?: types.World} = {}) {
```
dgozman, line 639: I'd prefer an explicit type for EvalOptions.

src/server/frames.ts
```diff
@@ -636,25 +636,10 @@ export class Frame extends SdkObject {
...
-      await this._page._doSlowMo();
-    return value;
-  }
-
-  async evaluateExpressionAndWaitForSignals(expression: string, isFunction: boolean | undefined, arg: any, world: types.World = 'main'): Promise<any> {
-    const context = await this._context(world);
-    const value = await context.evaluateExpressionAndWaitForSignals(expression, isFunction, arg);
+    const value = await context.eval(pageFunction, options);
     if (world === 'main')
       await this._page._doSlowMo();
```
dgozman, line 644: I am afraid about slowMo for internal evaluations.

JoelEinbinder: Yeah the tests should cover it, and they all pass. But it was a strange place for slowmo to be in the first place. I'll do the slowmo refactoring patch before we land this one to make sure it's ok.

PR #6772, opened 2021-05-27
fix(codegen): still open a page after clearing the codegen

JoelEinbinder:
Clicking the 'Clear' button after running `npx playwright codegen` would remove the
line that opened a new page. This caused the generated code to fail.

I am not sure if `this._params.startRecording` or `this._options.generateHeaders` is
the right signal to know if we need to generate the `const page = context.newPage()` code.

I am also not sure where tests for this should go. Do we have a test that presses
the clear button somehow?


dgozman:
> Do we have a test that presses the clear button somehow?

Here is an example: https://github.com/microsoft/playwright/pull/7165/files#diff-5551a9d827a079e8148f366fa36cc7369d84aa4675689241b7352a1fde46af45

mxschmitt:
Seems stale and not needed.

PR #6786, opened 2021-05-28
feat(port-forwarding): add playwrightclient support

mxschmitt:
Changes:

- moved port-forwarding from `BrowserServerImpl` to Playwright level so that we can use it in both (pwserver + launchServer) places
- accepting an `options` object now on pw client/server side
- added `timeout` support on pw-client connect


src/dispatchers/playwrightDispatcher.ts
```diff
@@ -40,11 +40,12 @@ export class PlaywrightDispatcher extends Dispatcher<Playwright, channels.Playwr
...
       preLaunchedBrowser,
     }, false);
-    this._portForwardingCallback = portForwardingCallback;
   }
 
   async enablePortForwarding(params: channels.PlaywrightEnablePortForwardingParams): Promise<void> {
-    assert(this._portForwardingCallback, 'Port forwarding is only supported when using connect()');
-    this._portForwardingCallback(params.ports);
+    this._object._setForwardedPorts(params.ports);
+    this._object._portForwardingServer.on('incomingSocksSocket', (socket: SocksInterceptedSocketHandler) => {
```
mxschmitt, line 47: When using the remote server this get's initialized two times, a) by the inprocess and b) by the server. If we have the on(dispatch in the constructor, this results in incorrect behavior. Maybe there is a better way to handle that.

src/remote/playwrightServer.ts
```diff
@@ -66,12 +77,12 @@ export class PlaywrightServer {
...
 
     const path = this._delegate.path;
-    const wsEndpoint = await new Promise<string>(resolve => {
+    const wsEndpoint = await new Promise<string>((resolve, reject) => {
       server.listen(port, () => {
         const address = server.address();
         const wsEndpoint = typeof address === 'string' ? `${address}${path}` : `ws://127.0.0.1:${address.port}${path}`;
         resolve(wsEndpoint);
-      });
+      }).on('error', reject);
```
mxschmitt, line 85: Fixes an issue when the port which was provided by the user was already used that it silently failed.

tests/portForwardingServer.spec.ts
```diff
@@ -161,3 +165,26 @@ it('should should not allow to connect when the server does not allow port-forwa
 
   await browserServer.close();
 });
+
+it('should proxy local requests', async ({ getPlaywright, mode, browserName, server, browserOptions }, workerInfo) => {
+  it.skip(mode !== 'service');
```
mxschmitt, line 170: Maybe an own service.spec.ts makes sense.

src/remote/playwrightClient.ts
```diff
@@ -18,24 +18,41 @@ import WebSocket from 'ws';
...
+    const playwrightPromise = connection.waitForObjectWithKnownName<Playwright>('Playwright');
+    let timer: NodeJS.Timeout;
+    try {
+      await Promise.race([
+        playwrightPromise,
+        errorPromise,
+        closePromise,
+        new Promise((_, reject) => timer = setTimeout(reject, timeout))
+      ]);
+      const playwright = await playwrightPromise;
```
dgozman, line 49: Put lines 49-52 into `playwrightPromise` to respect the timeout.

src/server/browser.ts
```diff
@@ -35,6 +36,7 @@ export interface BrowserProcess {
 export type PlaywrightOptions = {
   registry: registry.Registry,
   rootSdkObject: SdkObject,
+  portForwardingServer: BrowserServerPortForwardingServer,
```
dgozman, line 39: I don't like port forwarding server here. What do we need from it, just a port?

src/server/playwright.ts
```diff
@@ -26,6 +26,7 @@ import { WebKit } from './webkit/webkit';
 import { Registry } from '../utils/registry';
 import { CallMetadata, createInstrumentation, SdkObject } from './instrumentation';
 import { debugLogger } from '../utils/debugLogger';
+import { BrowserServerPortForwardingServer } from './socksSocket';
```
dgozman, line 29: It is time to rename the class to just `PortForwardingServer`.

src/server/playwright.ts
```diff
@@ -43,9 +45,11 @@ export class Playwright extends SdkObject {
         debugLogger.log(logName as any, message);
       }
     });
+    this._portForwardingServer = new BrowserServerPortForwardingServer(this);
```
dgozman, line 48: I'd rather not create this until we need it.

src/server/playwright.ts
```diff
@@ -54,6 +58,18 @@ export class Playwright extends SdkObject {
     this.android = new Android(new AdbBackend(), this.options);
     this.selectors = serverSelectors;
   }
+
+  _enablePortForwarding() {
+    this._portForwardingServer.start();
```
dgozman, line 63: Shouldn't we await this?

src/server/playwright.ts
```diff
@@ -54,6 +58,18 @@ export class Playwright extends SdkObject {
...
+  }
+
+  _disablePortForwarding() {
+    this._portForwardingServer.stop();
+  }
+
+  _setForwardedPorts(ports: number[]) {
```
dgozman, line 70: Who calls this? Should this be combined with "enable"?

tests/config/baseTest.ts
```diff
@@ -57,33 +62,48 @@ class DriverMode {
...
   private _serviceProcess: childProcess.ChildProcess;
+  private _tmpConfigFile?: string;
 
-  async setup(workerIndex: number) {
-    const port = 10507 + workerIndex;
-    this._serviceProcess = childProcess.fork(path.join(__dirname, '..', '..', 'lib', 'cli', 'cli.js'), ['run-server', String(port)], {
-      stdio: 'pipe'
+  async setup(workerIndex: number, options?: PlaywrightModeSetupOptions) {
+    if (options?.acceptForwardedPorts) {
+      this._tmpConfigFile = path.join(os.tmpdir(), `pw-server-config-${workerIndex}.json`);
```
dgozman, line 72: I'd prefer explicit CLI arguments for now, e.g. "first arg is a port number", "other args are forwarded ports" or something like that. Or maybe a JSON string argument for the ports. But not the temp file.

tests/config/baseTest.ts
```diff
@@ -36,6 +40,7 @@ type BaseOptions = {
 };
 type BaseFixtures = {
   platform: 'win32' | 'darwin' | 'linux';
+  getPlaywright: (options?: PlaywrightModeSetupOptions) => Promise<typeof import('../../index')>;
```
dgozman, line 43: Just add an option to BaseOptions called `service`, and use it in `playwright` fixture. Later, `test.use({ service: whatever })` in the test file.

This will make a nice prototype for the actual use.

tests/config/baseTest.ts
```diff
@@ -57,33 +62,48 @@ class DriverMode {
...
 
-  async setup(workerIndex: number) {
-    const port = 10507 + workerIndex;
-    this._serviceProcess = childProcess.fork(path.join(__dirname, '..', '..', 'lib', 'cli', 'cli.js'), ['run-server', String(port)], {
-      stdio: 'pipe'
+  async setup(workerIndex: number, options?: PlaywrightModeSetupOptions) {
+    if (options?.acceptForwardedPorts) {
+      this._tmpConfigFile = path.join(os.tmpdir(), `pw-server-config-${workerIndex}.json`);
+      const config = {acceptForwardedPorts: options.acceptForwardedPorts};
+      await fs.promises.writeFile(this._tmpConfigFile, JSON.stringify(config));
```
dgozman, line 74: Why not pass a specific port as before?

tests/portForwardingServer.spec.ts
dgozman: I wonder whether the tests should be agnostic to service/connect, and we could run them in both modes? I'd suggest a helper function right in this file that would either launch the browser in the service mode, or launchServer+connect in default mode. WDYT?

src/dispatchers/playwrightDispatcher.ts
mxschmitt: Yes currently its like that because the server side Playwright instance get's reused and passed a second time into the PlaywrightDispatcher, see here where it get's reused: https://github.com/microsoft/playwright/blob/0b0512a62492aacccaea7cce475c7000c1772da9/src/browserServerImpl.ts#L48

This results in that we have two times the `incomingSocksSocket` event listener which results in two events getting passed over the protocol. We should probably create a new Playwright event for new Servers created via launchServer.

src/client/connection.ts
```diff
@@ -63,9 +63,9 @@ export class Connection extends EventEmitter {
     this._onClose = onClose;
   }
 
-  async waitForObjectWithKnownName(guid: string): Promise<any> {
+  async waitForObjectWithKnownName<R=any>(guid: string): Promise<R> {
```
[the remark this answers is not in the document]
mxschmitt: will revert it for now.

src/server/browser.ts
mxschmitt: just the forwarded Ports and to get the proxy server. Can also reduce it so that we have just these two values:
- forwardPorts
- proxyOverride

tests/config/baseTest.ts
mxschmitt: I wanted to have them auto-generated via listen(0) on the server side, thought it might be better.

tests/config/baseTest.ts
mxschmitt: I just wanted to align it with the implementation of our launch-server mode.

src/server/playwright.ts
mxschmitt: IMO we need it separately. Since for `launchServer` we first start the socks server, then wait for incoming connections and then setting the forwardedPorts from the client side.

src/remote/playwrightServer.ts
```diff
@@ -111,6 +121,7 @@ export class PlaywrightServer {
         debugLog('Client error ' + error);
         disconnected();
       });
+      const onDisconnect = await this._delegate.onConnect(scope, forceDisconnect);
```
dgozman, line 124: Note that line 114 may throw now, please be careful.

src/server/playwright.ts
```diff
@@ -35,6 +36,7 @@ export class Playwright extends SdkObject {
   readonly firefox: Firefox;
   readonly webkit: WebKit;
   readonly options: PlaywrightOptions;
+  _portForwardingServer: PortForwardingServer | undefined;
```
dgozman, line 39: We should either make this private, or leave it public and remove all the methods below. Otherwise, it is not clear who owns this property.

tests/config/baseTest.ts
dgozman: What about this one? I still think we either do the `service` option or revert any changes in this file.

tests/config/baseTest.ts
dgozman: I still think that hardcoded port was better - gives less room for errors.

dgozman (approved):

PR #6805, opened 2021-05-31
docs: use sha256 for exposeFunction everywhere

### LINKED ISSUE #727: [documentation] s/md5/sha1/
image

This is really trivial comment. Documentation shows how to use 'BrowserContext#expose_function' with `sha1` where the sentence indicates `md5`

https://playwright.dev/python/docs/next/api/class-browsercontext#browser_contextexpose_functionname-callback

---

It seems that JS has a example code with `md5` but other languages have examples with `sha1`

https://github.com/microsoft/playwright/blob/master/docs/src/api/class-browsercontext.md

It would be better to describe all examples with `sha1` (or `sha256` if preferred)

mxschmitt:
Before we sometimes used sha-1 and md5. Now we are using md5 everywhere.

Fixes https://github.com/microsoft/playwright-python/issues/727

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -624,7 +624,7 @@ If the [`param: callback`] returns a [Promise], it will be awaited.
 
 See [`method: Page.exposeFunction`] for page-only version.
 
-An example of adding an `md5` function to all pages in the context:
+An example of adding a `sha256` function to all pages in the context:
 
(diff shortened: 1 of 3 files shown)


pavelfeldman:
Can we do sha1 instead? Since md5 has been broken, it should not be used pretty much anywhere.

pavelfeldman (approved):

PR #6813, opened 2021-06-01
feat(user-agent): Adding User-Agent in headers while making connection to browser

### LINKED ISSUE #6736: [Question] Determine playwright Version
How do I determine the playwright version running on the machine when user makes websocket connection?

karanshah-browserstack:
feat(user-agent): Adding User-Agent in headers while making a connection to the browser.

Adding User-Agent in connection headers while connecting to the browser.
Example: `'Playwright/1.12.0-next (x64/darwin/20.4.0)'`


src/client/browserType.ts
```diff
@@ -109,12 +109,14 @@ export class BrowserType extends ChannelOwner<channels.BrowserTypeChannel, chann
 
   async connect(params: ConnectOptions): Promise<Browser> {
     const logger = params.logger;
+    const paramsHeaders = params.headers ? params.headers : {};
+    paramsHeaders['User-Agent'] = getUserAgent();
```
yury-s, line 113: This is way it cannot be overridden by the client, let's change it to something like `Object.assign({'User-Agent': getUserAgent()}, params.headers)`

tests/browsertype-connect.spec.ts
```diff
@@ -76,13 +76,13 @@ test('should send extra headers with connect request', async ({browserType, star
     browserType.connect({
       wsEndpoint: `ws://localhost:${server.PORT}/ws`,
       headers: {
-        'User-Agent': 'Playwright',
+        'framework': 'Playwright',
```
yury-s, line 79: users should be able to override it

yury-s (changes requested):
Please add tests for the new header (see browsertype-connect.spec.ts)

tests/browsertype-connect.spec.ts
karanshah-browserstack: Done

src/client/browserType.ts
karanshah-browserstack: Resolved. Committed the change

yury-s (approved):
I feel like we may need to add client library language as well. We can do it in the future though as need be.

PR #6815, opened 2021-06-01
chore(evaluate): remove private _evaluateInUtility methods

JoelEinbinder:
These methods aren't used anywhere and it's been more than a few months since they were added.
There is some confusing logic around them and slowMo which I am trying to straighten out. Removing them is of
course easier.


dgozman (approved):
Is this unused? Nice!

PR #6816, opened 2021-06-01
feat(test): introduce @playwright/test

dgozman:
- `@playwright/test` is like `playwright-core`, but also includes test runner.
- You are supposed to `npx playwright install` to install browsers when using it.
- Exposes `npx playwright test` to run the tests.


aslushnikov (approved):

packages/build_package.js
```diff
@@ -129,9 +138,6 @@ if (!args.some(arg => arg === '--no-cleanup')) {
     bin: {
       playwright: './lib/cli/cli.js',
     },
-    engines: {
```
mxschmitt, line 132: What's the reason for removing that?

src/cli/testRunner.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (match)
+    return new RegExp(match[1], match[2]);
+  return new RegExp(pattern, 'g');
+}
+
+function overridesFromOptions(options: { [key: string]: any }): Config {
+  const shardPair = options.shard ? options.shard.split('/').map((t: string) => parseInt(t, 10)) : undefined;
+  return {
+    forbidOnly: options.forbidOnly ? true : undefined,
+    globalTimeout: options.globalTimeout ? parseInt(options.globalTimeout, 10) : undefined,
```
mxschmitt, line 162: Seems like it was not latest folio ToT since https://github.com/microsoft/folio/commit/4f81d520422bcb97968c6a30c14259ba0c134458 is missing.

packages/build_package.js
dgozman: It is duplicated above.

src/cli/fixtures.ts
```diff
@@ -0,0 +1,151 @@
...
+      recordVideo: recordVideo ? { dir: testInfo.outputPath('') } : undefined,
+      ...contextOptions,
+    };
+    if (acceptDownloads !== undefined)
+      options.acceptDownloads = acceptDownloads;
+    if (bypassCSP !== undefined)
+      options.bypassCSP = bypassCSP;
+    if (colorScheme !== undefined)
+      options.colorScheme = colorScheme;
+    if (deviceScaleFactor !== undefined)
```
mxschmitt, line 87: `reducedMotion` is missing, would be nice if we could have it more automated or at least leading to a TypeScript linting error when we forget a property for context/launch options.

PR #6831, opened 2021-06-02
feat: install chrome-beta via cli

aslushnikov:
This patch starts introduces `npx playwright install chrome-beta`
command and switches our Chrome Beta tests to it.


src/utils/utils.ts
```diff
@@ -19,9 +19,25 @@ import fs from 'fs';
...
 
 const mkdirAsync = util.promisify(fs.mkdir.bind(fs));
 
+export function spawnAsync(cmd: string, args: string[], options: any): Promise<{stdout: string, stderr: string, code: number, error?: Error}> {
+  const process = spawn(cmd, args, options);
+
+  return new Promise(resolve => {
+    let stdout = '';
+    let stderr = '';
+    if (stdout)
```
yury-s, line 32: This is always `false` as you have a local var `stdout` that evaluates to false

yury-s (approved):

PR #6836, opened 2021-06-02
feat: install media pack on windows with `npx playwright install-deps`

aslushnikov:
Chromium on Windows requires Media Pack to be installed. This patch
moves media pack installation under the `npx playwright install-deps`
umbrella.


bin/install_media_pack.ps1
```diff
@@ -0,0 +1 @@
+Install-WindowsFeature Server-Media-Foundation
```
yury-s, line 1: Does it work on regular win 10 (non server)?

src/install/installDeps.ts
```diff
@@ -16,15 +16,27 @@
...
 export async function installDeps(browserTypes: string[]) {
-  if (os.platform() !== 'linux')
-    return;
   if (!browserTypes.length)
     browserTypes = ['chromium', 'firefox', 'webkit'];
+  if (os.platform() === 'win32') {
+    if (browserTypes.includes('chromium')) {
+      const {code} = await utils.spawnAsync('powershell.exe', [path.join(SCRIPTS_DIRECTORY, 'install_media_pack.ps1')], { cwd: SCRIPTS_DIRECTORY, stdio: 'inherit' });
+      if (code !== 0)
+        throw new Error('Failed to install windows dependencies!');
```
yury-s, line 34: In the context of the comment above, maybe convert this in a warning?

yury-s (approved):

bin/install_media_pack.ps1
aslushnikov: Should work! At least according to stack overflow

src/install/installDeps.ts
aslushnikov: I'd keep it failing - it'll be more visible in the CI logs.

PR #6851, opened 2021-06-02
fix(waitForSelector): retry when context is gone during node adoption

### LINKED ISSUE #6800: WaitForSelector throws protocol error: Cannot find context with specified id
**Context:**
- Playwright Version: `1.11.1`
- Operating System: `Ubuntu 20.04.2`
- Node.js version: `14.17`
- Browser: Chromium
- Extra: Error was observed in GitHub Actions CI environment, and I can't reproduce it locally.

`envinfo` output:
```
 System:
    OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
    Memory: 2.06 GB / 6.79 GB
    Container: Yes
  Binaries:
    Node: 14.17.0 - /tmp/yarn--1622414329483-0.3784126230023115/node
    Yarn: 1.22.10 - /tmp/yarn--1622414329483-0.3784126230023115/yarn
    npm: 6.14.13 - /usr/local/bin/npm
  Languages:
    Bash: 5.0.17 - /usr/bin/bash
  npmPackages:
    playwright: ^1.11.1 => 1.11.1 
    playwright-chromium: ^1.11.1 => 1.11.1 
```

**Code Snippet**
```ts
await jestPlaywright.resetContext();
context.setDefaultTimeout(localGlobal.defaultTimeout);
await page.goto(localGlobal.url);

const $excerpt = await page.waitForSelector('data-testid=PostPreview-Excerpt'); // < exception happens here
```

**Exception observed:**

```
page.waitForSelector: Protocol error (DOM.describeNode): Cannot find context with specified id
    waiting for selector "data-testid=PostPreview-Excerpt" to be visible
      selector resolved to visible <p data-testid="PostPreview-Excerpt">Everyone knows AngularJs is done and dusted! Why …</p>
    Note: use DEBUG=pw:api environment variable to capture Playwright logs.
      45 |
      46 |   it('has list of posts', async () => {
    > 47 |     const $excerpt = await page.waitForSelector('data-testid=PostPreview-Excerpt');
```

Setting `DEBUG: pw:api` env variable helped to figure out the root cause.

In my case app router (JS) was performing "soft" redirect and part of the page being tested is re-rendered. Hence there is some kind of race condition - if node is found fast enough (before soft redirect and re-render) - no error occurs, if it is a little slower - playwright/chrome looses the context.

Error happened in github action execution environment and I could not reproduce it locally. But I don't believe there is anything specific to GH actions in it besides increased probability to hit the race condition.

Discussed this on slack and @JoelEinbinder suggested to open an issue.

dgozman:
There is a small window after finishing the "rerunnable task" where
we adopt the node to the main world and navigation could destroy the context.

Fixes #6800.

### DIFF
--- a/src/server/frames.ts
+++ b/src/server/frames.ts
@@ -677,13 +677,26 @@ export class Frame extends SdkObject {
     const task = dom.waitForSelectorTask(info, state);
     return controller.run(async progress => {
       progress.log(`waiting for selector "${selector}"${state === 'attached' ? '' : ' to be ' + state}`);
-      const result = await this._scheduleRerunnableHandleTask(progress, info.world, task);
-      if (!result.asElement()) {
-        result.dispose();
-        return null;
+      while (progress.isRunning()) {
+        const result = await this._scheduleRerunnableHandleTask(progress, info.world, task);
+        if (!result.asElement()) {
+          result.dispose();
+          return null;
+        }
+        if ((options as any).__testHookBeforeAdoptNode)
+          await (options as any).__testHookBeforeAdoptNode();
+        try {
+          const handle = result.asElement() as dom.ElementHandle<Element>;
+          const adopted = await handle._adoptTo(await this._mainContext());
+          return adopted;
+        } catch (e) {
+          // Navigated while trying to adopt the node.
+          if (!js.isContextDestroyedError(e) && !e.message.includes(dom.kUnableToAdoptErrorMessage))
+            throw e;
+          result.dispose();
+        }
       }
-      const handle = result.asElement() as dom.ElementHandle<Element>;
-      return handle._adoptTo(await this._mainContext());
+      return null;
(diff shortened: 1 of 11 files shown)


src/server/frames.ts
```diff
@@ -677,13 +677,23 @@ export class Frame extends SdkObject {
...
+        if (!result.asElement()) {
+          result.dispose();
+          return null;
+        }
+        try {
+          const handle = result.asElement() as dom.ElementHandle<Element>;
+          return handle._adoptTo(await this._mainContext());
+        } catch (e) {
+          // Navigated while trying to adopt the node.
+          if (!e.message.includes('Cannot find context with specified id'))
```
pavelfeldman, line 691: relying upon this message requires a test.

pavelfeldman (approved):

PR #6860, opened 2021-06-02
test: add tests for port-forwarding via playwrightclient

mxschmitt:
drive-by: allow getting the server instance when started via `startDefault()` to be able to close it from the server side.

### DIFF
--- a/tests/portForwardingServer.spec.ts
+++ b/tests/portForwardingServer.spec.ts
@@ -15,111 +15,135 @@
  */
 
 import http from 'http';
+import net from 'net';
 
-import { contextTest as it, expect } from './config/browserTest';
+import { PlaywrightClient } from '../lib/remote/playwrightClient';
+import { PlaywrightServer } from '../lib/remote/playwrightServer';
+
+import { contextTest, expect } from './config/browserTest';
 import type { LaunchOptions, ConnectOptions } from '../index';
+import type { Page, BrowserServer } from '..';
 
-it.skip(({ mode }) => mode !== 'default');
-it.fixme(({platform, browserName}) => platform === 'darwin' && browserName === 'webkit');
+type PageFactoryOptions = {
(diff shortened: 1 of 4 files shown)


tests/portForwardingServer.spec.ts
```diff
@@ -15,111 +15,134 @@
...
+import type { Page, BrowserServer } from '..';
+
+type PageFactoryOptions = {
+  acceptForwardedPorts: boolean
+  forwardPorts: number[]
+};
+
+type LaunchMode = 'playwrightclient' | 'launchServer';
+
+const it = contextTest.extend<{ pageFactory: (options?: PageFactoryOptions) => Promise<Page> }, {launchMode: LaunchMode}>({
```
dgozman, line 34: It is fine like it is now, but `launchMode` might as well be test-scoped - you are not doing anything worker-scoped here.

tests/portForwardingServer.spec.ts
```diff
@@ -15,111 +15,134 @@
...
-});
-
-it('should proxy local requests', async ({ browserType, browserOptions, server }, workerInfo) => {
-  process.env.PW_TEST_PROXY_TARGET = port.toString();
-  let reachedOriginalTarget = false;
-  server.setRoute('/foo.html', async (req, res) => {
-    reachedOriginalTarget = true;
-    res.end('<html><body></body></html>');
+function startTestServer() {
+  const server = http.createServer((req: http.IncomingMessage, res: http.ServerResponse) => {
```
dgozman, line 81: Don't you have to await for `listening` callback?

dgozman (approved):

tests/portForwardingServer.spec.ts
mxschmitt: > This function is asynchronous
from their documentation. So better to do so.

PR #3008, opened 2020-07-17
feat: support atomic browser installation - attempt 2

aslushnikov:
Currently, Ctrl-C while extracting browser might yield users in
a bad place.

This patch adds a marker file inside browser directory to make
sure that browser extraction completed.

Note: this was already attempted in #2489, but was eventually
reverted in #2534.

References #2660

PR #6871, opened 2021-06-03
chore: use fs.promises API instead of promisify

mxschmitt:
Ran the tests also on Node.js 12 which are passing for me.

### DIFF
--- a/src/install/installer.ts
+++ b/src/install/installer.ts
@@ -16,18 +16,12 @@
 
 import fs from 'fs';
 import path from 'path';
-import util from 'util';
 import lockfile from 'proper-lockfile';
 import {Registry, allBrowserNames, isBrowserDirectory, BrowserName, registryDirectory} from '../utils/registry';
 import * as browserFetcher from './browserFetcher';
 import { getAsBooleanFromENV, calculateSha1, removeFolders } from '../utils/utils';
 
-const fsMkdirAsync = util.promisify(fs.mkdir.bind(fs));
-const fsReaddirAsync = util.promisify(fs.readdir.bind(fs));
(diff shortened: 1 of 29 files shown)


src/install/installer.ts
```diff
@@ -16,18 +16,12 @@
...
 import * as browserFetcher from './browserFetcher';
 import { getAsBooleanFromENV, calculateSha1, removeFolders } from '../utils/utils';
 
-const fsMkdirAsync = util.promisify(fs.mkdir.bind(fs));
-const fsReaddirAsync = util.promisify(fs.readdir.bind(fs));
-const fsReadFileAsync = util.promisify(fs.readFile.bind(fs));
-const fsExistsAsync = (filePath: string) => fsReadFileAsync(filePath).then(() => true).catch(e => false);
-const fsUnlinkAsync = util.promisify(fs.unlink.bind(fs));
-const fsWriteFileAsync = util.promisify(fs.writeFile.bind(fs));
+const fsExistsAsync = (filePath: string) => fs.promises.readFile(filePath).then(() => true).catch(e => false);
```
JoelEinbinder, line 24: Not your code, but this is a really funny way to do fs.exists.

JoelEinbinder (approved):

src/install/installer.ts
mxschmitt: cc @aslushnikov maybe you are interested in a nit pick (rename or change it to fs.stat) since it was introduced by one of your patches: https://github.com/microsoft/playwright/pull/3008/files#diff-0b52ef30133aabfbf9ddde0aaa90370d351c36157422fbb08cd38c12be7b8f1dR29

Issue #3865, opened 2020-09-13
[BUG] $$ selector seems to be using invalid context


zackr:
**Context:**
- Playwright Version: 1.4.0
- Operating System: Mac
- Node.js version: 14.9.0
- Browser: All
- Extra: Nothing specific about it.

Here's a sample script showing that $$ is incorrectly returning an empty array when it should return an array of 14 elements:

const { chromium } = require('playwright');

(async () => {
    const browser = await chromium.launch();

    const page = await browser.newPage();
    await page.goto('https://www.roguefitness.com/rogue-hi-temp-bumper-plates');

    let groupedItemsHandle = await page.$(`#super-product-table > div.grouped-item`);
    if (groupedItemsHandle) {
        console.log('Regular selector worked and returned : %s', await groupedItemsHandle.innerText());
    }

    const numItems = await page.evaluate(() => document.querySelectorAll('#super-product-table > div.grouped-item').length);
    groupedItemsHandle = await page.$$(`#super-product-table > div.grouped-item`);
    assert.deepEqual(numItems, groupedItemsHandle.length);

    await browser.close();
})();

```

This seems to be caused by the fact that src/server/selectors.ts:_queryAll isn't setting up the context correctly. Changing line 90 from:
const context = await frame._context(allowUtilityContext ? info.world : 'main');
to:
const context = await frame._context(info.world);
would make it work, but I haven't had the time to figure out why _query, _queryArray and _queryAll all have different ways of setting up the context so it's probably a side-effect of something else.

dgozman:
Your analysis is correct - we pass `allowUtilityContext` as an optimization to avoid adopting elements one by one - this might be slow for large results. I think correctness is more important though. We might employ some kind of a heuristic, or maybe an option to say "do it faster, in the main context".

For the record, this particular page overrides `Array.from` with some partial implementation, so that `Array.from(new Set(['foo']))` does not work.

zackr:
Thanks for looking at this! FWIW $$eval(...) is likely broken too, so with the above mentioned code:
    const numEvalItems = await page.$$eval(`#super-product-table > div.grouped-item`, (es) => es.length);
    console.log(numEvalItems);
is going to print 0, because _queryArray is unconditionally using _mainContext. 

Either it would have to get the same adoptToMain code you've added to queryAll or start using _context(info.world) plus explicit adoption like _query does. In general unifying that logic between all those three would seem like a good idea. 
[...]

PR #6876, opened 2021-06-03
chore: do not use Array.from in injected script

### LINKED ISSUE #6869: [BUG] $$eval does not evaluate the expected amounts of elements
**Context:**

- Playwright Version: tried both 1.9.2 and 1.11.1-1621490832000
- Operating System: Ubuntu 20.10
- Python Version: 3.9

**Code Snippet**

```python
import time

with sync_playwright() as p:
    browser = p.chromium.launch()
    page = browser.new_page()

    page.goto("https://www.littlegreene.com/wallpaper/collection/")

    # Neither of these are necessary, it's just in case.
    time.sleep(15)
    for elem in page.query_selector_all("div.categories-landing h3 a"):
        elem.scroll_into_view_if_needed()

    # print(len(page.eval_on_selector_all("div.categories-landing h3 a", "(l) => l.map(a => a.href)")))
    print(len(page.eval_on_selector_all("div.categories-landing h3 a", "(l) => l")))
    print(len(page.query_selector_all("div.categories-landing h3 a")))
```

**Describe the bug**

The output is 

```
0
7
```

when I would expect the first snippet to also return results. I can run the selector with the `map` in the Chrome devtools and running it through `page.evaluate`

```python
page.evaluate("$$('div.categories-landing h3 a').map(a => a.href)")
```
works as well. This is the first time I've encountered this issue.

Edit: it seems like selectors don't work through `eval_on_selector` or `eval_on_selector_all`, 

```python
    print(page.eval_on_selector("a", "l => l"))
    print(page.eval_on_selector("div", "l => l"))
```

return `None` and `{}` respectively.

dgozman:
This method is most often overridden by some bad polyfill that does not work correctly and breaks `$eval` and `$$eval` methods. See #3865 and #6869 as some examples.

As a best-effort fix, use `[...iterable]` throughout the code.

Fixes #6869.

### DIFF
--- a/src/server/injected/selectorEvaluator.ts
+++ b/src/server/injected/selectorEvaluator.ts
@@ -68,9 +68,9 @@ export class SelectorEvaluatorImpl implements SelectorEvaluator {
     this._engines.set('near', createPositionEngine('near', boxNear));
     this._engines.set('nth-match', nthMatchEngine);
 
-    const allNames = Array.from(this._engines.keys());
+    const allNames = [...this._engines.keys()];
     allNames.sort();
-    const parserNames = Array.from(customCSSNames).slice();
+    const parserNames = [...customCSSNames];
     parserNames.sort();
     if (allNames.join('|') !== parserNames.join('|'))
       throw new Error(`Please keep customCSSNames in sync with evaluator engines: ${allNames.join('|')} vs ${parserNames.join('|')}`);
--- a/src/server/injected/injectedScript.ts
+++ b/src/server/injected/injectedScript.ts
@@ -124,7 +124,7 @@ export class InjectedScript {
         }
         set = newSet;
       }
-      let result = Array.from(set) as Element[];
+      let result = [...set] as Element[];
       if (partsToCheckOne.length) {
         const partial = { parts: partsToCheckOne };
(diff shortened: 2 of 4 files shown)


src/server/injected/selectorEvaluator.ts
```diff
@@ -701,3 +701,19 @@ function sortInDOMOrder(elements: Element[]): Element[] {
...
+// For some reason, Array.from is often replaced with some bad polyfill.
+// Use our own implementation that works.
+export function arrayFrom<T>(collection: Iterable<T> | ArrayLike<T>): T[] {
+  const result: T[] = [];
+  if (Symbol.iterator in collection) {
+    const iterable = collection as Iterable<T>;
+    for (const item of iterable)
+      result.push(item);
+  } else {
+    const arrayLike = collection as ArrayLike<T>;
```
JoelEinbinder, line 714: when do we have a non-iterable array-like?

JoelEinbinder: Oh thats just in the definition of `Array.from` TIL

yury-s (approved):

JoelEinbinder (approved):

mxschmitt (approved):

PR #6884, opened 2021-06-03
chore: do not use a subshell hack when using XVFB

aslushnikov:
This was originally introduced in 9caa61aed181552473cfd794299ae66dd5197a2a.
However, we no longer save process STDERR.

### DIFF
--- a/.github/workflows/tests_fyi.yml
+++ b/.github/workflows/tests_fyi.yml
@@ -29,9 +29,7 @@ jobs:
         DEBUG: pw:install
     - run: npm run build
     - run: node lib/cli/cli install-deps ${{ matrix.browser }} chromium
-    # XVFB-RUN merges both STDOUT and STDERR, whereas we need only STDERR
-    # Wrap `npm run` in a subshell to redirect STDERR to file.
-    - run: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- bash -c "npm run test -- --project=${{ matrix.browser }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/tests_fyi.yml
```diff
@@ -29,9 +29,7 @@ jobs:
         DEBUG: pw:install
     - run: npm run build
     - run: node lib/cli/cli install-deps ${{ matrix.browser }} chromium
-    # XVFB-RUN merges both STDOUT and STDERR, whereas we need only STDERR
-    # Wrap `npm run` in a subshell to redirect STDERR to file.
-    - run: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- bash -c "npm run test -- --project=${{ matrix.browser }}"
+    - run: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- npm run test -- --project=${{ matrix.browser }}
```
mxschmitt, line 32: I locally run the tests without `--auto-servernum --server-args="-screen 0 1280x960x24` also in the Python repo we don't have it. Can we remove it?

mxschmitt (approved):

.github/workflows/tests_fyi.yml
aslushnikov: The `--auto-servernum` is a general good practice; this sets up isolated x displays for different commands. In our case we never have a second consumer for xvfb, but I'd still keep it.

The screen configuration was added at f0fcdc8f51dcbb30cc80c10c9d2fddd52501392b and I don't think anything has changed with that regard. I'd keep it as-is too.

PR #6888, opened 2021-06-03
fix(tracing): error handling

dgozman:
- Reject when ZipFile signals an error.
- Make sure snapshotter does not save trace events after `stop()`.
- Await pending blob writes on `stop()`.

### DIFF
--- a/src/server/trace/recorder/tracing.ts
+++ b/src/server/trace/recorder/tracing.ts
@@ -17,6 +17,7 @@
 import fs from 'fs';
 import path from 'path';
 import yazl from 'yazl';
+import { EventEmitter } from 'events';
 import { calculateSha1, createGuid, mkdirIfNeeded, monotonicTime } from '../../../utils/utils';
 import { Artifact } from '../../artifact';
 import { BrowserContext } from '../../browserContext';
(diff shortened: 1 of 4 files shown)


src/server/trace/recorder/tracing.ts
```diff
@@ -127,7 +133,10 @@ export class Tracing implements InstrumentationListener {
       return;
     const snapshotName = `${name}@${metadata.id}`;
     metadata.snapshots.push({ title: name, snapshotName });
-    await this._snapshotter!.captureSnapshot(sdkObject.attribution.page, snapshotName, element);
+    const promise = this._snapshotter!.captureSnapshot(sdkObject.attribution.page, snapshotName, element);
+    this._pendingSnapshots.add(promise);
```
pavelfeldman, line 137: Think _pendingCalls already ensures it

src/server/trace/recorder/tracing.ts
```diff
@@ -103,21 +105,25 @@ export class Tracing implements InstrumentationListener {
   }
 
   async export(): Promise<Artifact> {
-    if (!this._traceFile)
-      throw new Error('Tracing directory is not specified when launching the browser');
+    if (!this._traceFile || this._started)
+      throw new Error('Must start and stop tracing before exporting');
     const zipFile = new yazl.ZipFile();
-    zipFile.addFile(this._traceFile, 'trace.trace');
```
pavelfeldman, line 109: Can't we just try/catch it all together?

mxschmitt:
Seems like it fixes the following flaky test too: https://github.com/microsoft/playwright-python/runs/2748148009#step:12:881

pavelfeldman (approved):

PR #6901, opened 2021-06-04
feat(cli): allow to list-only the dependencies without installing them

mxschmitt:
I want to leverage this in the Heroku buildpack. There I don't need the to install it via APT, I need to pipe the list of dependencies into a special file which contains the dependencies. This can also get re-used by other tools probably.

Maybe `--dry` is a better name?

For that: https://github.com/mxschmitt/heroku-playwright-buildpack

### DIFF
--- a/src/cli/cli.ts
+++ b/src/cli/cli.ts
@@ -152,10 +152,11 @@ async function installChromeChannel(channel: string) {
 
 program
     .command('install-deps [browserType...]')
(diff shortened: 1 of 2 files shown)


yury-s (approved):
I like `--dry-run` better and would prefer it to print out actual commands that it is going to run but if it's more convenient for your use case to have just a list of libraries I'm ok with that too.

PR #6916, opened 2021-06-04
fix: various test-related fixes

dgozman:
- Closing inspector window resumes the script.
- Replace FOLIO_WORKER_INDEX with TEST_WORKER_INDEX.
- Account for `@playwright/test` stack traces.

### DIFF
--- a/tests/config/default.config.ts
+++ b/tests/config/default.config.ts
@@ -63,7 +63,7 @@ const config: folio.Config<CommonOptions & PlaywrightOptions> = {
 const browserNames = ['chromium', 'webkit', 'firefox'] as BrowserName[];
 for (const browserName of browserNames) {
   const executablePath = getExecutablePath(browserName);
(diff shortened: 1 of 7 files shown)


tests/config/default.config.ts
```diff
@@ -63,7 +63,7 @@ const config: folio.Config<CommonOptions & PlaywrightOptions> = {
 const browserNames = ['chromium', 'webkit', 'firefox'] as BrowserName[];
 for (const browserName of browserNames) {
   const executablePath = getExecutablePath(browserName);
-  if (executablePath && !process.env.FOLIO_WORKER_INDEX)
+  if (executablePath && !process.env.TEST_WORKER_INDEX)
```
JoelEinbinder, line 66: Where is the other side of this? Or it already landed in folio?

dgozman: Yep.

PR #6919, opened 2021-06-05
docs(dotnet): add test runner docs

pavelfeldman:


docs/src/test-runners-csharp.md
```diff
@@ -0,0 +1,112 @@
+---
+id: test-runners
+title: "Test Runners"
+---
+
+With a few lines of code, you can hook up Playwright to your favorite Java test runner.
```
dgozman, line 6: Java -> .NET

docs/src/intro-csharp.md
```diff
@@ -73,12 +72,56 @@ By default, Playwright runs the browsers in headless mode. To see the browser UI
 await playwright.Firefox.LaunchAsync(new BrowserTypeLaunchOptions { Headless = false, SlowMo = 50 });
 ```
 
+## First test
+
+You can choose to use NUnit test fixtures that come bundled with Playwright. These fixtures support running tests on multiple browsers out of the box. They also allow running tests in parallel. Learn more about [Playwright with NUnit](./test-runners.md).
```
dgozman, line 77: "multiple browser" is talking about running in parallel, right?  Let's make it a bit more clear.

dgozman (approved):

PR #6926, opened 2021-06-06
feat: switch to the Firefox Stable equivalent by default

### LINKED ISSUE #6817: [BUG] roll Firefox to 89 and switch to it
- [x] roll firefox to new stable (firefox 89 is out!)
- [ ] switch to stable by default
- [ ] rename `firefox` browser to `firefox-beta` everywhere to avoid confusion (folders / bots / CDN)

aslushnikov:
This patch:
- starts downloading Firefox Stable equivalent by default
- starts running Firefox-Stable on our smoke tests (tests-1)
- starts running Firefox-Beta on our CQ1 tests (tests-2)

Note: there's a little confusion right now with browser names:
- `firefox-stable` - firefox-stable equivalent
- `firefox`- firefox-beta equivalent

I'll rename `firefox` to `firefox-beta` in a follow-up.

Fixes #6817

### DIFF
--- a/.github/workflows/infra.yml
+++ b/.github/workflows/infra.yml
@@ -19,9 +19,9 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: 12
-    - uses: microsoft/playwright-github-action@v1
     - run: npm ci
     - run: npm run build
+    - run: node lib/cli/cli install-deps chromium firefox webkit
     - run: npm run lint
     - name: Verify clean tree
       run: |
@@ -39,8 +39,8 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: 12
-    - uses: microsoft/playwright-github-action@v1
     - run: npm ci
     - run: npm run build
+    - run: node lib/cli/cli install-deps chromium
     - run: node utils/build/update_canary_version.js --today-date
     - run: utils/build/build-playwright-driver.sh
--- a/src/server/browserType.ts
+++ b/src/server/browserType.ts
@@ -35,20 +35,22 @@ import { CallMetadata, SdkObject } from './instrumentation';
 const ARTIFACTS_FOLDER = path.join(os.tmpdir(), 'playwright-artifacts-');
 
 export abstract class BrowserType extends SdkObject {
-  private _name: registry.BrowserName;
+  private _name: 'chromium'|'firefox'|'webkit';
+  private _binaryName: registry.BrowserName;
   readonly _registry: registry.Registry;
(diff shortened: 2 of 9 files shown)


.github/workflows/infra.yml
```diff
@@ -19,9 +19,9 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: 12
-    - uses: microsoft/playwright-github-action@v1
```
mxschmitt, line 22: Not related to this PR: We use `microsoft/playwright-github-action` in quite a few other places in this repository. Can we fully remove it from the repo?

src/server/browserType.ts
```diff
@@ -35,20 +35,22 @@ import { CallMetadata, SdkObject } from './instrumentation';
...
   readonly _playwrightOptions: PlaywrightOptions;
 
-  constructor(browserName: registry.BrowserName, playwrightOptions: PlaywrightOptions) {
+  constructor(name: 'chromium'|'firefox'|'webkit', binaryName: registry.BrowserName, playwrightOptions: PlaywrightOptions) {
     super(playwrightOptions.rootSdkObject, 'browser-type');
     this.attribution.browserType = this;
     this._playwrightOptions = playwrightOptions;
-    this._name = browserName;
+    this._name = name;
+    this._binaryName = binaryName;
```
mxschmitt, line 48: nit: I'm not a huge fan of `_binaryName`. Maybe:
- `_name` -> `_baseBrowserName`
- `_binaryName` -> `_browserName`

mxschmitt (approved):

src/server/browserType.ts
aslushnikov: The word "browser" is very overloaded, so I'd prefer to have distinct names for distinct entities. 

To elaborate: both `//src/utils/registry.ts` and playwright servers `//src/server/browserType.ts` use word "browser" here and there, yet these are two very different things:
- in playwright servers, "browser" means the browser interface we use to communicate. So we have "chromium" for CDP, "webkit" for WebInspector Remote Debugging Protocol, and "firefox" for juggler. These all are "browsers". 
- in registry, "browser" means an entity that we download from our CDN. Originally we were only downloading browsers, but we now have FFMPEG and we are considering moving WinLDD there as well. 

If we were to start over, we'd establish some other terminology for the registry that doesn't include name "browser" - "binary" seems to be fine since it's distinct enough and is not overused around the codebase.

.github/workflows/infra.yml
aslushnikov: yess good point

PR #6930, opened 2021-06-07
fix(test runner): remove folio/jest namespaces in expect matchers

dgozman:
We'll be able to bring jest namespace back if/when needed.

### DIFF
--- a/docs/src/test-advanced.md
+++ b/docs/src/test-advanced.md
@@ -467,16 +467,9 @@ For TypeScript, also add the following to `global.d.ts`. You don't need it for J
 
 ```js
 // global.d.ts
-declare namespace folio {
+declare namespace PlaywrightTest {
   interface Matchers<R> {
     toBeWithinRange(a: number, b: number): R;
   }
 }
-```
-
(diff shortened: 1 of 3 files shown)


docs/src/test-advanced.md
```diff
@@ -467,16 +467,9 @@ For TypeScript, also add the following to `global.d.ts`. You don't need it for J
...
 // global.d.ts
-declare namespace folio {
+declare namespace PlaywrightTest {
   interface Matchers<R> {
     toBeWithinRange(a: number, b: number): R;
   }
 }
 ```
-
-To import expect matching libraries like [jest-extended](https://github.com/jest-community/jest-extended#installation) you can import it from your `global.d.ts`:
```
mxschmitt, line 477: Wouldn't something like that work for users then? (move extending from jest from us to the use-space). Then the native expect-extend imports should work out of the box.

```ts
declare namespace PlaywrightTest {
  interface Matchers<R> extends jest.Matchers<R> {
  }
}
```

mxschmitt (approved):

PR #6941, opened 2021-06-07
docs: add missing -c tests to the reporter examples

yury-s:


pavelfeldman:
We just removed that!

PR #6952, opened 2021-06-08
feat(test runner): add tests for playwright-specific fixtures

dgozman:


tests/playwright-test/playwright.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+      });
+    `,
+  });
+
+  expect(result.exitCode).toBe(1);
+  expect(result.passed).toBe(1);
+  expect(result.failed).toBe(1);
+
+  const dirPass = testInfo.outputPath('test-results', 'a-pass-chromium');
+  const videoPass = fs.existsSync(dirPass) ? fs.readdirSync(dirPass).find(file => file.endsWith('webm')) : undefined;
```
JoelEinbinder, line 157: why are we unsure if there will be a directory here?

dgozman: Flakes on the bots. I think that video was not written, because we did not wait enough time. Let me do a single worker which might help.

aslushnikov (approved):

PR #6959, opened 2021-06-08
feat(pwt): webServer config option

JoelEinbinder:
Adds a new fixture `baseURL` that can be used in tests. `baseURL` is either set by the `PW_BASE_URL` environment variable
or by a `webServer` config option. `webServer` accepts a shell command to run and a port on which to wait for a server to appear.

### DIFF
--- a/src/test/index.ts
+++ b/src/test/index.ts
@@ -46,6 +46,12 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
     await browser.close();
   }, { scope: 'worker' } ],
 
+  baseURL: [async ({ }, use) => {
+    if (!process.env.PW_BASE_URL)
+      throw new Error('A web server must be specified to use baseURL. See `webServer` in the config options, or set the PW_BASE_URL environment variable.');
+    await use(process.env.PW_BASE_URL);
+  }, {scope: 'worker'}],
+
   screenshot: 'off',
   video: 'off',
   trace: 'off',
--- a/src/test/runner.ts
+++ b/src/test/runner.ts
@@ -34,6 +34,8 @@ import EmptyReporter from './reporters/empty';
 import { ProjectImpl } from './project';
 import { Minimatch } from 'minimatch';
 import { Config } from './types';
+import { launchWebServer } from './server';
+import type { ChildProcess } from 'child_process';
 
 const removeFolderAsync = promisify(rimraf);
 const readDirAsync = promisify(fs.readdir);
@@ -151,6 +153,7 @@ export class Runner {
     let globalSetupResult: any;
     if (config.globalSetup)
       globalSetupResult = await this._loader.loadGlobalHook(config.globalSetup, 'globalSetup')(this._loader.fullConfig());
+    const webServer: ChildProcess|null = config.webServer && await launchWebServer(config.webServer);
     try {
       for (const file of allTestFiles)
         this._loader.loadTestFile(file);
@@ -232,6 +235,10 @@ export class Runner {
         await globalSetupResult(this._loader.fullConfig());
       if (config.globalTeardown)
         await this._loader.loadGlobalHook(config.globalTeardown, 'globalTeardown')(this._loader.fullConfig());
+      if (webServer) {
+        webServer.kill();
+        delete process.env.PW_BASE_URL;
+      }
     }
   }
 }
--- a/src/test/server.ts
+++ b/src/test/server.ts
@@ -1,0 +1,37 @@
+import { exec } from 'child_process';
+import net from 'net';
+import { monotonicTime, raceAgainstDeadline } from './util';
+import { WebServerConfig } from '../../types/test';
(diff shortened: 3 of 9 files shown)


src/test/server.ts
```diff
@@ -0,0 +1,36 @@
...
+        throw new Error(`failed to start web server on port ${port} via "${command}"`);
+    }
+    process.env.PW_BASE_URL = `http://localhost:${port}`;
+    return webServer;
+}
+async function waitForSocket(port: number, delay: number, cancellationToken: {canceled: boolean}) {    
+    while (!cancellationToken.canceled) {
+        const connected = await new Promise((resolve) => {
+        const conn = net
+            .connect(port)
```
mxschmitt, line 21: It would be cool if we could verify that at least the server responds. Its often that a server is started on the TCP level, then express e.g. will use it and don't respond until the actual dev server is started. Thats takes sometimes 30-120 seconds to boot up then working server.

Besides that we could think about checking the status codes.

JoelEinbinder: Theres a grace period in the browser when you `goto` an unresponsive browser. For Chromium it seems to be about 500ms if nothing is listening on the port, and then it's more like 30 seconds if something is listening. So I think that waiting for something to be listening is the most important part.

If the server takes 120 seconds to start up that is still going to break. `jest-dev-server` has a default timeout of 5 seconds. What's going on that causes a server to take 2 minutes?

src/test/index.ts
```diff
@@ -148,5 +148,11 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
   page: async ({ context }, use) => {
     await use(await context.newPage());
   },
+
+  baseURL: [async ({ }, use) => {
+    if (!process.env.PW_BASE_URL)
+      throw new Error('A web server must be specified to use baseURL. See `webServer` in the config options, or set the PW_BASE_URL environment variable.');
+    await use(process.env.PW_BASE_URL);
```
dgozman, line 155: Let's rename PW -> PLAYWRIGHT

src/test/server.ts
```diff
@@ -0,0 +1,36 @@
+import { exec } from 'child_process';
```
dgozman, line 1: Missing copyright.

src/test/runner.ts
```diff
@@ -232,6 +235,10 @@ export class Runner {
         await globalSetupResult(this._loader.fullConfig());
       if (config.globalTeardown)
         await this._loader.loadGlobalHook(config.globalTeardown, 'globalTeardown')(this._loader.fullConfig());
+      if (webServer) {
+        webServer.kill();
+        delete process.env.PW_BASE_URL;
```
dgozman, line 240: Why do delete the env here, but set it in some other place?

src/test/index.ts
```diff
@@ -148,5 +148,11 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
   page: async ({ context }, use) => {
     await use(await context.newPage());
   },
+
+  baseURL: [async ({ }, use) => {
+    if (!process.env.PW_BASE_URL)
+      throw new Error('A web server must be specified to use baseURL. See `webServer` in the config options, or set the PW_BASE_URL environment variable.');
```
dgozman, line 154: Note that you can just do `use: { baseURL: 'foo' }` in the config, so perhaps we do not need the environment variable?

tests/playwright-test/server.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+        webServer: {
+          command: 'node ${path.join(__dirname, 'assets', 'simple-server.js')} ${port}',
+          port: ${port},
+          timeout: 100,
+        }
+      };
+    `,
+  });
+  expect(result.exitCode).toBe(1);
+  expect(result.output).toContain(`failed to start web server on port 10500 via "node`);
```
dgozman, line 67: This 10500 substring is definitely flaky :)

types/test.d.ts
```diff
@@ -1105,6 +1126,11 @@ export type PlaywrightWorkerArgs = {
    * Browser instance, shared between multiple tests.
    */
   browser: Browser;
+
+  /**
+   * URL of the web server. For example, http://localhost:1234
+   */
+  baseURL: String;
```
dgozman, line 1133: - I do think this should go to the PlaywrightWorkerOptions.
- I would also expect my `page` to be pre-navigated to this url? Perhaps we should not rush with this one, and expose something like `webServer: string` fixture instead?

src/test/server.ts
```diff
@@ -0,0 +1,36 @@
...
+import net from 'net';
+import { monotonicTime, raceAgainstDeadline } from './util';
+
+export async function launchWebServer({command, port, timeout = 10000}: {command: string, port: number, timeout?: number}) {
+    const webServer = exec(command);
```
dgozman, line 6: Indents are off in this file.

mxschmitt:
Closing in favour of #7368.

PR #6991, opened 2021-06-09
fix(test-runner): support passing slowMo option

### LINKED ISSUE #6984: slowMo does not exist in type 'Fixtures<{}, {}, PlaywrightTestOptions, PlaywrightWorkerOptions>'
Hi!
I'm trying to set up config (TypeScript) for test runner (1.12.0) as described here - https://playwright.dev/docs/test-configuration
But got the error for `slowMo` property - 
`Object literal may only specify known properties, and 'slowMo' does not exist in type 'Fixtures<{}, {}, PlaywrightTestOptions, PlaywrightWorkerOptions>'`

Screenshot from 2021-06-09 12-38-08

mxschmitt:
Fixes #6984
Revers #6967

### DIFF
--- a/src/test/index.ts
+++ b/src/test/index.ts
@@ -28,9 +28,10 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
   playwright: [ require('../inprocess'), { scope: 'worker' } ],
   headless: [ undefined, { scope: 'worker' } ],
   channel: [ undefined, { scope: 'worker' } ],
+  slowMo: [ undefined, { scope: 'worker' } ],
   launchOptions: [ {}, { scope: 'worker' } ],
 
-  browser: [ async ({ playwright, browserName, headless, channel, launchOptions }, use) => {
+  browser: [ async ({ playwright, browserName, headless, channel, slowMo, launchOptions }, use) => {
     if (!['chromium', 'firefox', 'webkit'].includes(browserName))
       throw new Error(`Unexpected browserName "${browserName}", must be one of "chromium", "firefox" or "webkit"`);
     const options: LaunchOptions = {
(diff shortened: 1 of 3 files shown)


src/test/cli.ts
```diff
@@ -146,7 +146,7 @@ function overridesFromOptions(options: { [key: string]: any }): Config {
     forbidOnly: options.forbidOnly ? true : undefined,
     globalTimeout: isDebuggerAttached ? 0 : (options.globalTimeout ? parseInt(options.globalTimeout, 10) : undefined),
     grep: options.grep ? forceRegExp(options.grep) : undefined,
-    maxFailures: options.x ? 1 : (options.maxFailures ? parseInt(options.maxFailures, 10) : undefined),
+    maxFailures: options.maxFailures ? 1 : (options.maxFailures ? parseInt(options.maxFailures, 10) : undefined),
```
pavelfeldman, line 149: What is happening here?

src/test/index.ts
```diff
@@ -87,6 +91,8 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
       options.bypassCSP = bypassCSP;
     if (colorScheme !== undefined)
       options.colorScheme = colorScheme;
+    if (reducedMotion !== undefined)
```
pavelfeldman, line 94: What does this all code do? :)

types/test.d.ts
```diff
@@ -997,6 +1011,12 @@ export type PlaywrightTestOptions = {
    */
   colorScheme: ColorScheme | undefined;
 
+  /**
```
pavelfeldman, line 1014: I thought those would go through as is, now I can see that we wrap each value :/

src/test/index.ts
mxschmitt: Pass it from the worker options to the context options.

types/test.d.ts
mxschmitt: Yeah :/ lets rework it in a follow-up.

src/test/index.ts
aslushnikov: is this a drive-by?

mxschmitt: It is 👍

aslushnikov (approved):

PR #6992, opened 2021-06-09
fix(@playwright/test): plumbing

JoelEinbinder:
`npx playwright test` should find and forward to the @playwright/test package.

### DIFF
--- a/src/cli/cli.ts
+++ b/src/cli/cli.ts
@@ -250,8 +250,20 @@ program
     });
 
 if (!process.env.PW_CLI_TARGET_LANG) {
-  if (packageJSON.name === '@playwright/test' || process.env.PWTEST_CLI_ALLOW_TEST_COMMAND) {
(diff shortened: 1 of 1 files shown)


src/cli/cli.ts
```diff
@@ -250,8 +250,20 @@ program
...
+      playwrightTestPackagePath = '../test/cli';
+    } else {
+      playwrightTestPackagePath = require.resolve('@playwright/test/lib/test/cli', {
+        paths: [__dirname, process.cwd()]
+      });
+    }
+  } catch {}
+
+  if (playwrightTestPackagePath) {
+    require(playwrightTestPackagePath).addTestCommand(program);
```
pavelfeldman, line 266: I think this is what will throw, not the code above.

JoelEinbinder: No, require.resolve throws if the package can't be found.

pavelfeldman (approved):

PR #7005, opened 2021-06-10
docs: fix the global setup example

pavelfeldman:


docs/src/test-advanced.md
```diff
@@ -248,7 +248,7 @@ Now add `globalSetup` option to the configuration file.
 ```js js-flavor=js
 // playwright.config.js
 module.export = {
-  globalSetup: 'global-setup.js',
+  globalSetup: require.resolve('./global-setup'),
```
pjcalvo, line 251: I was using 
```javascript
 globalSetup: `${process.cwd()}/src/global-setup.ts`,
```
instead...Either case works..

PR #7010, opened 2021-06-10
docs: added extensions page

### LINKED ISSUE #5586: [Question]Can I use playwright to test chrome extension?
I would like to test a chrome extension using playwright. 
I would like to make the simple actions on the extension html - like click, navigates etc'

Is there an option to do that using playwright?

mxschmitt:
Fixes #5586

When we moved from `ChromiumBrowser` to Browser this [got deleted ](https://github.com/microsoft/playwright/blob/cac119f3bf1b6e86462fa3556760b585d91d0ffd/docs/src/api/class-chromiumbrowser.md#class-chromiumbrowser)and users keep asking for it. My idea would be to add it like that, since we also don't refer to the mobile page.

image

### DIFF
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -1,6 +1,6 @@
 ---
 id: extensions
-title: "Testing extensions (experimental)"
+title: "Testing extensions"
 ---
 
 :::note
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -1,0 +1,29 @@
+---
+id: extensions
+title: "Testing extensions (experimental)"
+---
+
+:::note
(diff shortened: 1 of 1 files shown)


pavelfeldman:
You can add it into the guides, there is nothing experimental, but it needs to have a note saying it is Chrome-only.

pavelfeldman (approved):

docs/src/extensions.md
```diff
@@ -0,0 +1,29 @@
+---
+id: extensions
+title: "Testing extensions (experimental)"
```
mxschmitt, line 3:
```suggestion
title: "Testing extensions"
```

PR #7014, opened 2021-06-10
fix(test-runner): do only match JS/TS files when collecting

### LINKED ISSUE #338: [Bug] Folio wants to parse snapshots as JS/TS
After the latest snapshot change Folio wants to parse my snapshots with the TS compiler. This leads to a `SyntaxError: Error while reading ` since the snapshots have now the test file name as a file prefix.

Having a strict file extension filter should fix that to only allow JS and TS.

config:

```js
  {
    name: 'api',
    testMatch: /api/,
  },
```

this fixes it:

```
  {
    name: 'api',
    testMatch: /api.*ts$/,
  },
```

mxschmitt:
Fixes https://github.com/microsoft/folio/issues/338

### DIFF
--- a/src/test/runner.ts
+++ b/src/test/runner.ts
@@ -143,7 +143,8 @@ export class Runner {
       const allFiles = await collectFiles(project.config.testDir);
       const testMatch = createMatcher(project.config.testMatch);
       const testIgnore = createMatcher(project.config.testIgnore);
-      const testFiles = allFiles.filter(file => !testIgnore(file) && testMatch(file) && testFileFilter(file));
+      const testFileExtension = (file: string) => ['.js', '.ts'].includes(path.extname(file));
+      const testFiles = allFiles.filter(file => !testIgnore(file) && testMatch(file) && testFileFilter(file) && testFileExtension(file));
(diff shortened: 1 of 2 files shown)


src/test/runner.ts
```diff
@@ -143,7 +143,8 @@ export class Runner {
       const allFiles = await collectFiles(project.config.testDir);
       const testMatch = createMatcher(project.config.testMatch);
       const testIgnore = createMatcher(project.config.testIgnore);
-      const testFiles = allFiles.filter(file => !testIgnore(file) && testMatch(file) && testFileFilter(file));
+      const testFileExtension = (file: string) => ['.js', '.ts'].includes(path.extname(file));
+      const testFiles = allFiles.filter(file => !testIgnore(file) && testMatch(file) && testFileFilter(file) && testFileExtension(file));
```
pavelfeldman, line 147: Do we support .mjs yet?

pavelfeldman (approved):

src/test/runner.ts
mxschmitt: Not yet.

PR #7018, opened 2021-06-10
browser(firefox): fixed Array.prototype.toJSON workaround

### LINKED ISSUE #7015: [BUG] Firefox: strange undefined toJSON property on JS objects
**Context:**
- Playwright Version: 1.12.1
- Operating System: Windows
- Node.js version: 14.6.1
- Browser: Firefox

**Describe the bug**
After evaluating some custom code in the context of the **page**, some of JS objects receive strange `toJSON` property, which is `undefined`. Started to happen after upgrade to 1.12. And this happens on Firefox ONLY.

mxschmitt:
References #7015


yury-s (commented):
Let's revert original fix for now if the issue is severe enough and just use serializer that doesn't modify user objects.

PR #7019, opened 2021-06-10
Updating README for reference to Playwright .NET

timheuer:
Adding a reference in this main README to Playwright for .NET as there is not good redirection from here right now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@
 
 Playwright is a Node.js library to automate [Chromium](https://www.chromium.org/Home), [Firefox](https://www.mozilla.org/en-US/firefox/new/) and [WebKit](https://webkit.org/) with a single API. Playwright is built to enable cross-browser web automation that is **ever-green**, **capable**, **reliable** and **fast**.
 
+> Looking for **Playwright for .NET?** Head over to https://github.com/microsoft/playwright-dotnet for the latest info and release!
+
 |          | Linux | macOS | Windows |
 |   :---   | :---: | :---: | :---:   |
 | Chromium <!-- GEN:chromium-version -->93.0.4530.0<!-- GEN:stop --> | :white_check_mark: | :white_check_mark: | :white_check_mark: |


README.md
```diff
@@ -6,6 +6,8 @@
 
 Playwright is a Node.js library to automate [Chromium](https://www.chromium.org/Home), [Firefox](https://www.mozilla.org/en-US/firefox/new/) and [WebKit](https://webkit.org/) with a single API. Playwright is built to enable cross-browser web automation that is **ever-green**, **capable**, **reliable** and **fast**.
 
+> Looking for **Playwright for .NET?** Head over to https://github.com/microsoft/playwright-dotnet for the latest info and release!
```
pavelfeldman, line 9: We don't actually want to send customers to the github readme, most of the useful content is here: https://playwright.dev/dotnet/docs/intro/. And we link to it from https://playwright.dev, in every page toolbar. I wonder if that is sufficient!

timheuer: The reality is the readme exists, has a LOT of info on it, and would be helpful to the .NET developers as a pointer to the right repo where their Playright source/info is at as well.  I understand the marketing site playwright.dev but as long as the README's exist on this main repo (which playwright.dev links to as well), it seems a nominal addition to add clarity to the visitor of this repo.

avodovnik, line 9:
```suggestion
> More comfortable in another programming language? Playwright is also available for [Java](https://github.com/microsoft/playwright-java), [.NET](https://github.com/microsoft/playwright-dotnet) and [Python](https://github.com/microsoft/playwright-python).
```

There are more flavours available to Playwright, and (despite my clear bias) I'd prefer not to single out .NET :-)

avodovnik: I agree with Tim - our page already links to all the languages, as it should. Our readme _should_ also at least mention the languages we support - but, as per my suggestion here, I'd want to mention all the languages. If/when this PR merges, I'll also open PRs on other languages to cross-link.

timheuer: See I wasn't even aware of the other ones ;-) -- as the primary repo, this is great just to provide a small, but helpful addition to the readme that I don't think takes away at all from the official site, etc.  Thanks!

pavelfeldman: The problem is that we don't want to send anyone to https://github.com/microsoft/playwright-dotnet. We don't really maintain it, don't update it with latest and greatest snippets, etc. etc.

> The reality is the readme exists, has a LOT of info on it

So far it is two snippets and a paragraph of text, so I'm trying to make sure it does not get a lot of information.

> I understand the marketing site playwright.dev

That's exactly the misconception I'm trying to mitigate - playwright.dev contains 20,000 lines of markdown documentation for Playwright, per language, with guides, examples, etc. Repo does not have any of that information, only playwright.dev does. If 140 is a LOT I wonder what you say about 20K! I'd like to make sure people don't think README.md is a viable replacement for playwright.dev.

pavelfeldman: Is there a way to rephrase it as something like "playwright.dev is where you learn about Playwright"? From there, they can pick a language?

timheuer: I think you are underestimating the confusion currently caused and how developers look for information and especially in open source environments when repos exist.  This felt simple but not a hill I'm willing o die on to be honest.  I logged the request because I fell into the trap. I'm a .NET developer and all the documentation and messaging didn't point to me. There are links to the GH repo and from there a dead-end to me for a .NET dev.  If we don't want people to go there then maybe we should take off the prominent link to the repo that's all.

pavelfeldman: Sorry about the hill feeling, not my intention here.

> If we don't want people to go there then maybe we should take off the prominent link to the repo that's all.

You mean hide https://github.com/microsoft/playwright-dotnet altogether? We were considering turning Playwright into monorepo, so that main repo would consume all the others. But we were not sure our branching model would handle it before we released all the ports. We might revisit it. But yes, we definitely want to make it feel 'internal'. Current [Getting Started](https://github.com/microsoft/playwright-dotnet#getting-started) section there is plain harmful - it completely hides the actual [Getting Started](https://playwright.dev/dotnet/docs/intro) guide.

pavelfeldman: Let me fix it: https://github.com/microsoft/playwright-dotnet/pull/1551

PR #7049, opened 2021-06-10
feat(codegen): allow generating tests

### LINKED ISSUE #7020: [Feature] Codegen should be able to emit @playwright/test syntax

pavelfeldman:
#7020

### DIFF
--- a/src/server/supplements/recorder/codeGenerator.ts
+++ b/src/server/supplements/recorder/codeGenerator.ts
@@ -168,8 +168,11 @@ export class CodeGenerator extends EventEmitter {
     const text = [];
     if (this._options.generateHeaders)
       text.push(languageGenerator.generateHeader(this._options));
-    for (const action of this._actions)
-      text.push(languageGenerator.generateAction(action));
+    for (const action of this._actions) {
+      const actionText = languageGenerator.generateAction(action);
+      if (actionText)
+        text.push(actionText);
+    }
     if (this._options.generateHeaders)
       text.push(languageGenerator.generateFooter(this._options.saveStorage));
     return text.join('\n');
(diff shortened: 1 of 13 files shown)


src/server/supplements/recorder/codeGenerator.ts
```diff
@@ -168,8 +168,11 @@ export class CodeGenerator extends EventEmitter {
     const text = [];
     if (this._options.generateHeaders)
       text.push(languageGenerator.generateHeader(this._options));
-    for (const action of this._actions)
-      text.push(languageGenerator.generateAction(action));
+    for (const action of this._actions) {
+      const actionText = languageGenerator.generateAction(action);
```
JoelEinbinder, line 172: Did we ever have new lines to split blocks of code? I expected an explicit null check instead of empty string but if they are equivalent then I don't care.

src/server/supplements/recorder/javascript.ts
```diff
@@ -84,8 +93,12 @@ export class JavaScriptLanguageGenerator implements LanguageGenerator {
 
     if (emitPromiseAll)
       formatter.add(`]);`);
-    else if (signals.assertNavigation)
-      formatter.add(`  // assert.equal(${pageAlias}.url(), ${quote(signals.assertNavigation.url)});`);
+    else if (signals.assertNavigation) {
+      if (this._isTest)
+        formatter.add(`  expect(${pageAlias}.url()).toBe(${quote(signals.assertNavigation.url)});`);
+      else
+        formatter.add(`  // assert.equal(${pageAlias}.url(), ${quote(signals.assertNavigation.url)});`);
```
JoelEinbinder, line 100: Maybe remove the asserts? We have tests with expect for people now.

JoelEinbinder (approved):

src/server/supplements/recorder/codeGenerator.ts
pavelfeldman: Same thing, yes

src/server/supplements/recorder/javascript.ts
pavelfeldman: Yes, I did it and then test started failing and i don't want to fix it for merge. Will follow up.

PR #7056, opened 2021-06-11
docs: support typescript snippets in intro

### LINKED ISSUE #6971: Add typescript example to getting started docs
I chose to use Playwright because it is written in TS and so I assume has good TS support. However I struggled to get a simple hello world working in typescript because I was importing chromium like `import chromium from "playwright"` not `import { chromium } from "playwright"`.

Please add a minimal TS example, currently the only TS code in the docs is `let page: import('playwright').Page;`. I suggest adding a TS version of your getting started script:
```js
const { webkit } = require('playwright');

(async () => {
  const browser = await webkit.launch();
  await page.goto('http://whatsmyuseragent.org/');
  await page.screenshot({ path: `example.png` });
  await browser.close();
})();
```

aslushnikov:
Fixes #6971

### DIFF
--- a/docs/src/intro-js.md
+++ b/docs/src/intro-js.md
@@ -43,7 +43,7 @@ Playwright APIs are asynchronous and return Promise objects. Our code examples u
 
 In our first script, we will navigate to `whatsmyuseragent.org` and take a screenshot in WebKit.
 
-```js
+```js js-flavor=js
 const { webkit } = require('playwright');
 
 (async () => {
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):

mxschmitt:
I also tried it the only thing which I didn't like about it was that TS is default. Then users when they copy/paste it will get the import error since it uses import.

pavelfeldman:
Let's leave it as is then?

aslushnikov:
closing then

PR #7059, opened 2021-06-11
fix(tracing): do not stall on dialogs

### LINKED ISSUE #7048: [BUG] Dialogs cannot be dismissed if tracing is on in Chromium or Webkit
**Context:**
- Playwright Version: 1.12.1
- Operating System: macOS 11.2.3
- Node.js version: 12.16.3
- Browser: Chromium, Webkit
- Extra: Using the @playwright/test runner

**Code Snippet**

Setup a new playwright project with the following test file and config file:

```typescript
// test.spec.ts

test("basic test", async ({ page }) => {
  await page.goto("https://playwright.dev/");

  page.on("dialog", async (dialog) => {
    console.log('dismissing')

    await dialog.accept(); // hangs here

    console.log('dismissed')
  });

  await page.evaluate(() => {
    confirm("are you sure");
  });

  const name = await page.innerText(".navbar__title");
  expect(name).toBe("Playwright");
});
```

```typescript
// playwright.config.ts
export default {
  projects: [
    {
      name: "Chrome",
      use: {
        browserName: "chromium",
        channel: "chrome",
        trace: "on", // works when trace is off
      },
    },

    {
      name: "Firefox",
      use: { browserName: "firefox", trace: "on" }, // works regardless
    },

    {
      name: "WebKit",
      use: { browserName: "webkit", trace: "on" }, // works when trace is off
    },
  ],
};

```
**Describe the bug**

When tracing is enabled via playwright.config.ts, dialogs cannot be dismissed. Neither the default dialog dismissing or manual `page.on('dialog', dialog => { ... })` methods work. When dismissing manually, it seems to hang at the `dialog.accept()` method.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/7048

### DIFF
--- a/tests/tracing.spec.ts
+++ b/tests/tracing.spec.ts
@@ -103,6 +103,21 @@ test('should collect two traces', async ({ context, page, server }, testInfo) =>
   }
 });
 
+test('should not stall on dialogs', async ({ page, context, server }) => {
+  await context.tracing.start({ screenshots: true, snapshots: true });
+  await page.goto(server.EMPTY_PAGE);
+
+  page.on("dialog", async (dialog) => {
+    await dialog.accept();
+  });
+
+  await page.evaluate(() => {
+    confirm("are you sure");
+  });
+  await context.tracing.stop();
+});
+
+
 for (const params of [
   {
     id: 'fit',
--- a/tests/tracing.spec.ts
+++ b/tests/tracing.spec.ts
@@ -112,7 +112,7 @@ test('should not stall on dialogs', async ({ page, context, server }) => {
   });
 
   await page.evaluate(() => {
-    confirm("are you sure");
+    confirm('are you sure');
   });
   await context.tracing.stop();
 });
--- a/tests/tracing.spec.ts
+++ b/tests/tracing.spec.ts
@@ -107,7 +107,7 @@ test('should not stall on dialogs', async ({ page, context, server }) => {
   await context.tracing.start({ screenshots: true, snapshots: true });
   await page.goto(server.EMPTY_PAGE);
 
-  page.on("dialog", async (dialog) => {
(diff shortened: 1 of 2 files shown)


aslushnikov (approved):

tests/tracing.spec.ts
```diff
@@ -103,6 +103,21 @@ test('should collect two traces', async ({ context, page, server }, testInfo) =>
...
+test('should not stall on dialogs', async ({ page, context, server }) => {
+  await context.tracing.start({ screenshots: true, snapshots: true });
+  await page.goto(server.EMPTY_PAGE);
+
+  page.on('dialog', async dialog => {
+    await dialog.accept();
+  });
+
+  await page.evaluate(() => {
+    confirm("are you sure");
```
JoelEinbinder, line 115:
```suggestion
    confirm('are you sure');
```

tests/tracing.spec.ts
```diff
@@ -103,6 +103,21 @@ test('should collect two traces', async ({ context, page, server }, testInfo) =>
   }
 });
 
+test('should not stall on dialogs', async ({ page, context, server }) => {
+  await context.tracing.start({ screenshots: true, snapshots: true });
+  await page.goto(server.EMPTY_PAGE);
+
+  page.on('dialog', async dialog => {
+    await dialog.accept();
```
JoelEinbinder, line 111: Do we have to catch this promise and await it? I am worried that `dialog.accept` will race with `context.close` and throw an unhandled promise rejection.

JoelEinbinder (approved):

tests/tracing.spec.ts
pavelfeldman: I believe we are ok since the following call to confirm is blocking and is serialized with the accept ack.

PR #7061, opened 2021-06-11
fix(test-runner): resolve global hooks relative to the config dir

### LINKED ISSUE #7004: [test runner] Error: Error while reading global-setup.ts: Cannot find module 'global-setup.ts'
This does not work:

```ts
  globalSetup: 'global-setup.ts',
};
```

This works
```ts
  globalSetup: path.join(__dirname, 'global-setup.ts'),
};
```

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/7004

### DIFF
--- a/src/test/loader.ts
+++ b/src/test/loader.ts
@@ -57,9 +57,9 @@ export class Loader {
       if (config && typeof config === 'object' && ('default' in config))
         config = config['default'];
       this._config = config;
+      this._configFile = file;
       const rawConfig = { ...config };
       this._processConfigObject(path.dirname(file));
-      this._configFile = file;
       return rawConfig;
     } catch (e) {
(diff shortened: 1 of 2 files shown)


aslushnikov (approved):

src/test/loader.ts
```diff
@@ -77,6 +77,12 @@ export class Loader {
   private _processConfigObject(rootDir: string) {
     validateConfig(this._config);
 
+    // Resolve script hooks relative to the root dir.
+    if (this._config.globalSetup)
+      this._config.globalSetup = path.resolve(rootDir, this._config.globalSetup);
```
JoelEinbinder, line 82: I believe this should be resolved to the config file dir not to the root dir.

pavelfeldman: Root dir is the file dir (in this method :))

PR #7080, opened 2021-06-11
docs(test-runner): fixed wrong function calls

mxschmitt:


pavelfeldman (approved):

shiftkey:
:wave: not sure where to file this, but the live docs on `playwright.dev` do not contain these fixes:



Is there a better place to report this fix not actually being published?

mxschmitt:
The fixes get released in the next version 1.13: or today via the next tag: https://playwright.dev/docs/next/test-fixtures

PR #7084, opened 2021-06-12
fix(snapshotter): support constructed CSSStyleSheet

### LINKED ISSUE #7085: [BUG] Traceviewer screens are not recorded well when using constructable stylesheets
**Context:**
- Playwright Version: 12.1.1
- Operating System: mac
- Node.js version: 16
- Browser: Chromium

when running trace viewer on a site built with stencil-js, the html snapshots are displayed incorrectly. 
This is due to the use of constructable stylesheet.

JoelEinbinder:
Fixes #7085

### DIFF
--- a/src/server/snapshot/snapshotterInjected.ts
+++ b/src/server/snapshot/snapshotterInjected.ts
@@ -39,6 +39,7 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
   const kShadowAttribute = '__playwright_shadow_root_';
   const kScrollTopAttribute = '__playwright_scroll_top_';
   const kScrollLeftAttribute = '__playwright_scroll_left_';
+  const kStyleSheetAttribute = '__playwright_style_sheet_';
 
   // Symbols for our own info on Nodes/StyleSheets.
   const kSnapshotFrameId = Symbol('__playwright_snapshot_frameid_');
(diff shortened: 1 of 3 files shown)


pavelfeldman:
I am reading about constructable ss [here](https://developers.google.com/web/updates/2019/02/constructable-stylesheets) and it feels like you should dedupe them. It will actually make the code simpler to follow.

The doc is saying that you are supposed to create a bunch of constructable stylesheets and then add them into the adopted collections left and right. I would mimic this spirit in the serialization process:
- dedupe constructables by reference, serialize them once
- use your array of stylesheets only on the snapshot itself, not on shadow attrs
- put the index in that table in the template attributes for hydration

That way you are as scaleable as the web itself.

pavelfeldman (approved):

src/server/snapshot/snapshotterInjected.ts
```diff
@@ -345,6 +355,15 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
           }
           for (let child = node.firstChild; child; child = child.nextSibling)
             visitChild(child);
+          let ownedDocument = null;
```
dgozman, line 358: Let's call this `documentOrShadowRoot`

src/server/snapshot/snapshotterInjected.ts
```diff
@@ -397,6 +416,23 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
         return checkAndReturn(result);
       };
 
+      const visitStyleSheet = (sheet: CSSStyleSheet) => {
+        const data = ensureCachedData(sheet);
+        const oldCSSText = data.cssText;
+        const cssText = this._updateStyleElementStyleSheetTextIfNeeded(sheet) || '';
+        const equals = cssText === oldCSSText;
+        const checkAndReturn = (n: NodeSnapshot): { equals: boolean, n: NodeSnapshot } => {
```
dgozman, line 424: Let's inline this method.

dgozman (approved):

PR #7087, opened 2021-06-13
feat(test-runner): introduce debug mode

pavelfeldman:


src/test/index.ts
```diff
@@ -71,103 +76,135 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
...
+    locale: testOptions.locale,
+    offline: testOptions.offline,
+    permissions: testOptions.permissions,
+    proxy: testOptions.proxy,
+    storageState: testOptions.storageState,
+    timezoneId: testOptions.timezoneId,
+    userAgent: testOptions.userAgent,
+    viewport: testOptions.viewport,
+  };
+  return { ...testOptions.contextOptions, ...topLevelContextOptions };
```
dgozman, line 207: I think that if I do `test.use({ viewport: undefined })` to revert to the default, it will actually be undefined instead of the default. There is a helper against that: https://github.com/microsoft/playwright/blob/master/src/test/util.ts#L171-L180.

PR #7100, opened 2021-06-14
fix(test-runner): do not require esModuleInterop=true

### LINKED ISSUE #7099: [test-runner] Missing  types for the expect library
Hi !

In vscode, the expect library imported from '@playwright/test' proposes only a few choices:

image

And using the available assertions displays an error. For example:

```ts
import { expect } from '@playwright/test';
expect(...).toBe(...)
// Property 'toBe' does not exist on type 'Matchers<string>'.ts(2339)
```

mxschmitt:
They have a `export = foo` export [in place,](https://unpkg.com/browse/expect@27.0.2/build/index.d.ts) which would require `esModuleInterop` or `allowSyntheticDefaultImports`. This PR does workaround that, since its not a TypeScript default which confuses users.

Fixes #7099

### DIFF
--- a/tests/playwright-test/expect.spec.ts
+++ b/tests/playwright-test/expect.spec.ts
@@ -73,6 +73,31 @@ test('should work with default expect matchers', async ({runTSC}) => {
   expect(result.exitCode).toBe(0);
 });
 
+test('should work with default expect matchers and esModuleInterop=false', async ({runTSC}) => {
+  const result = await runTSC({
+    'a.spec.ts': `
+      const { test } = pwt;
+      test.expect(42).toBe(42);
+    `,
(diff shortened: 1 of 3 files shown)


pavelfeldman (commented):
It is a little hard for me to assess the implications of this. It is a breaking change to our users?

mxschmitt:
> It is a little hard for me to assess the implications of this. It is a breaking change to our users?

Should not be in my opinion. By this change it works in both modes, without the default export support and with. I just moved a few lines from the 3rd-party package expect into our codebase.

dgozman:
I'd like Joel to review this instead of me 😄

JoelEinbinder (approved):

PR #7113, opened 2021-06-14
chore: bump Node.js TypeScript typings to v14

mxschmitt:


src/remote/playwrightServer.ts
```diff
@@ -78,7 +79,7 @@ export class PlaywrightServer {
     const wsEndpoint = await new Promise<string>((resolve, reject) => {
       server.listen(port, () => {
         const address = server.address();
-        const wsEndpoint = typeof address === 'string' ? `${address}${path}` : `ws://127.0.0.1:${address.port}${path}`;
+        const wsEndpoint = `ws://127.0.0.1:${(address as AddressInfo).port}${path}`;
```
JoelEinbinder, line 82: why do we lose the check if address is a string?

PR #7125, opened 2021-06-15
fix(test-runner): video: 'retain-on-failure' leaves empty folders behind

### LINKED ISSUE #7093: Folder for a test-case is getting generated in test-results even if Test Case Passes when properties are given on Failure
Hi Guys,

     I am using Playwright and also have recommended this tool to my friends ITS AWESOME. 
I have a doubt, For latest Playwright Runner in "playwright.config.js" I have given properties as 
     "screenshot: 'only-on-failure',
        video: 'retain-on-failure',
        trace: 'retain-on-failure',"

Still a Blank Folder with Test Case Name is Created when all my test Cases are passing. Is it expected? 
Because I think Folder must be generated only if Test Case Fails.
 
Thanks in Advance

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/7093

### DIFF
--- a/src/test/index.ts
+++ b/src/test/index.ts
@@ -162,7 +162,7 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
     await context.close();
 
     if (video === 'retain-on-failure' && testFailed) {
-      await Promise.all(allPages.map(async (page, i) => {
+      await Promise.all(allPages.map(async page => {
         const video = page.video();
         if (!video)
           return;
--- a/src/test/index.ts
+++ b/src/test/index.ts
@@ -16,12 +16,16 @@
 
 import * as fs from 'fs';
 import * as path from 'path';
+import * as os from 'os';
 import type { LaunchOptions, BrowserContextOptions, Page } from '../../types/types';
 import type { TestType, PlaywrightTestArgs, PlaywrightTestOptions, PlaywrightWorkerArgs, PlaywrightWorkerOptions } from '../../types/test';
 import { rootTestType } from './testType';
-
+import { createGuid, removeFolders } from '../utils/utils';
 export { expect } from './expect';
 export const _baseTest: TestType<{}, {}> = rootTestType.test;
+
+const artifactsFolder = path.join(os.tmpdir(), 'pwt-' + createGuid());
+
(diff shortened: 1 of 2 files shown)


utils/check_deps.js
```diff
@@ -164,7 +164,7 @@ DEPS['src/server/trace/recorder/'] = ['src/server/trace/common/', ...DEPS['src/s
 DEPS['src/server/trace/viewer/'] = ['src/server/trace/common/', ...DEPS['src/server/trace/common/']];
 
 // No dependencies for test runner.
```
JoelEinbinder, line 166:
```suggestion
```

src/test/index.ts
```diff
@@ -16,12 +16,16 @@
...
+import * as os from 'os';
 import type { LaunchOptions, BrowserContextOptions, Page } from '../../types/types';
 import type { TestType, PlaywrightTestArgs, PlaywrightTestOptions, PlaywrightWorkerArgs, PlaywrightWorkerOptions } from '../../types/test';
 import { rootTestType } from './testType';
-
+import { removeFolders } from '../utils/utils';
 export { expect } from './expect';
 export const _baseTest: TestType<{}, {}> = rootTestType.test;
+
+const artifactsFolder = path.join(os.tmpdir(), 'playwright-test-artifacts-');
```
JoelEinbinder, line 27: this looks like it should use mkdrtmp to generate some random characters after 'playwright-test-artifacts-'. Right now it will race if two tests run in parallel.

src/test/index.ts
```diff
@@ -150,17 +161,17 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
     }
     await context.close();
 
-    const deleteVideos = video === 'retain-on-failure' && !testFailed;
-    if (deleteVideos) {
-      await Promise.all(allPages.map(async page => {
+    if (video === 'retain-on-failure' && testFailed) {
+      await Promise.all(allPages.map(async (page, i) => {
```
JoelEinbinder, line 165: nit: unused `i`
```suggestion
      await Promise.all(allPages.map(async page => {
```

JoelEinbinder (approved):

PR #7150, opened 2021-06-15
feat(response): expose connection details in API

### LINKED ISSUE #7147: [Feature] Add Response.securityDetails
Expose `securityDetails` to the client side and make it a getter since it can be quite big. Since HAR is a persistent format we want to make available with out public API too.

rwoll:
This allows users to access the connection information outside of the
HAR file in realtime.

Closes #7147.
Relates #6631.

### DIFF
--- a/docs/src/api/class-response.md
+++ b/docs/src/api/class-response.md
@@ -48,6 +48,27 @@ Contains a boolean stating whether the response was successful (status in the ra
 
 Returns the matching [Request] object.
 
+## async method: Response.securityDetails
+- returns: <[null]|[Object]>
+  - `issuer` <[string]> Common Name component of the Issuer field.
+    from the certificate. This should only be used for informational purposes. Optional.
+  - `protocol` <[string]> The specific TLS protocol used. (e.g. `TLS 1.3`). Optional.
+  - `subjectName` <[string]> Common Name component of the Subject
+    field from the certificate. This should only be used for informational purposes. Optional.
+  - `validFrom` <[int]> Unix timestamp (in seconds) specifying
+    when this cert becomes valid. Optional.
+  - `validTo` <[int]> Unix timestamp (in seconds) specifying
+    when this cert becomes invalid. Optional.
+
+Returns SSL and other security information.
+
+## async method: Response.serverAddr
+- returns: <[null]|[Object]>
+  - `ipAddress` <[string]> IPv4 or IPV6 address of the server.
+  - `port` <[int]>
+
+Returns the IP address and port of the server.
+
 ## method: Response.status
 - returns: <[int]>
 
--- a/src/dispatchers/networkDispatchers.ts
+++ b/src/dispatchers/networkDispatchers.ts
@@ -80,6 +80,14 @@ export class ResponseDispatcher extends Dispatcher<Response, channels.ResponseIn
   async body(): Promise<channels.ResponseBodyResult> {
     return { binary: (await this._object.body()).toString('base64') };
   }
+
+  async securityDetails(): Promise<channels.ResponseSecurityDetailsResult> {
+    return { value: await this._object.securityDetails() || undefined };
+  }
+
+  async serverAddr(): Promise<channels.ResponseServerAddrResult> {
+    return { value: await this._object.serverAddr() || undefined };
+  }
 }
 
 export class RouteDispatcher extends Dispatcher<Route, channels.RouteInitializer> implements channels.RouteChannel {
--- a/src/protocol/channels.ts
+++ b/src/protocol/channels.ts
@@ -2372,6 +2372,8 @@ export type ResponseInitializer = {
 export interface ResponseChannel extends Channel {
   body(params?: ResponseBodyParams, metadata?: Metadata): Promise<ResponseBodyResult>;
   finished(params?: ResponseFinishedParams, metadata?: Metadata): Promise<ResponseFinishedResult>;
+  securityDetails(params?: ResponseSecurityDetailsParams, metadata?: Metadata): Promise<ResponseSecurityDetailsResult>;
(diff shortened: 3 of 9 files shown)


utils/generate_types/overrides.d.ts
```diff
@@ -222,6 +222,19 @@ type AccessibilityNode = {
...
+  subjectName?: string,
+  validFrom?: number,
+  validTo?: number,
+};
+
+export type RemoteAddr = {
+  ipAddress: string,
+  port: number,
+};
+
```
rwoll, line 237: These types kept not ending up in `types.d.ts`, so I added them here. Is that expected or should they have automatically showed up in `types.d.ts`. I was running…

```
$ node utils/generate_types/
$ node utils/generate_channels.js
$ npm run build
$ npm run lint 
```

docs/src/api/class-response.md
```diff
@@ -48,6 +48,16 @@ Contains a boolean stating whether the response was successful (status in the ra
 
 Returns the matching [Request] object.
 
+## async method: Response.securityDetails
+- returns: <[null]|[SecurityDetails]>
```
pavelfeldman, line 52: Checkout `method: Request.timing` and expand it here, that way you get a chance to document fields and also you won't need overrides.

pavelfeldman (commented):
Looks sharp!

src/dispatchers/networkDispatchers.ts
```diff
@@ -80,6 +80,14 @@ export class ResponseDispatcher extends Dispatcher<Response, channels.ResponseIn
   async body(): Promise<channels.ResponseBodyResult> {
     return { binary: (await this._object.body()).toString('base64') };
   }
+
+  async securityDetails(): Promise<channels.ResponseSecurityDetailsResult> {
+    return { value: await this._object.securityDetails() || undefined };
```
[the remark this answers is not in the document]
rwoll: `await this._object.securityDetails()` returns `T|null` which TS was complaining wasn't compatible with `value: T|undefined`

docs/src/api/class-response.md
rwoll: Thanks! Updated to use that strategy.

However, it looks like inline optionals are not supported so I used an explicit union.

This works fine (notice the explicit `[null]|T`)
```
- returns: <[null]|[Object]>
  - `issuer?` <[string]> Common Name component of the Issuer field
```
but (no union, `?` on `Object` type instead):
```
- returns: <[Object?]>
  - `issuer?` <[string]> Common Name component of the Issuer field
```
errors with:
```
$ node utils/generate_types/
Error: Nested properties given, but there are no objects in type expression: [Object?]
    at Function.parse (/Users/rwoll/code/playwright/utils/doclint/documentation.js:421:13)
    at ApiParser.parseType (/Users/rwoll/code/playwright/utils/doclint/api_parser.js:183:31)
    at ApiParser.parseMember (/Users/rwoll/code/playwright/utils/doclint/api_parser.js:90:27)
    at /Users/rwoll/code/playwright/utils/doclint/api_parser.js:50:14
    at visit (/Users/rwoll/code/playwright/utils/markdown.js:366:3)
    at visit (/Users/rwoll/code/playwright/utils/markdown.js:368:5)
    at Object.visitAll (/Users/rwoll/code/playwright/utils/markdown.js:358:5)
    at new ApiParser (/Users/rwoll/code/playwright/utils/doclint/api_parser.js:48:8)
    at parseApi (/Users/rwoll/code/playwright/utils/doclint/api_parser.js:308:10)
    at /Users/rwoll/code/playwright/utils/generate_types/index.js:40:19
```

I was modeling the `?` syntax off:

```
## async method: Response.json
* langs: csharp
- returns: <[JsonElement?]>
```

docs/src/api/class-response.md
```diff
@@ -48,6 +48,27 @@ Contains a boolean stating whether the response was successful (status in the ra
 
 Returns the matching [Request] object.
 
+## async method: Response.securityDetails
+- returns: <[null]|[Object]>
+  - `issuer?` <[string]> Common Name component of the Issuer field
```
pavelfeldman, line 53: `issuer` <[string]> Common Name component of the Issuer field. Optional.

pavelfeldman: This 'Optional' in the end will make it optional. I'm really sorry about this one.

src/dispatchers/networkDispatchers.ts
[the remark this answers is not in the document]
pavelfeldman: Oh

src/protocol/channels.ts
```diff
@@ -2383,6 +2385,29 @@ export type ResponseFinishedOptions = {};
 export type ResponseFinishedResult = {
   error?: string,
 };
+export type ResponseSecurityDetailsParams = {};
+export type ResponseSecurityDetailsOptions = {};
+export type ResponseSecurityDetailsResult = {
+  value?: SecurityDetails,
```
pavelfeldman, line 2391: I think our type system thinks about it as a required field named `value?` :)

rwoll: I think TS is resolving the protocol.yml as intended:

```
securityDetails:
  returns:
    value: SecurityDetails?
```

a field named `value` that's `undefined | SecurityDetails`.

I suppose we could skip the wrapper `value` (for SecurityDetails since all fields are optional, but not for ServerAddr), but I wanted to maintain the same API that we have internally and keep SecurityDetails and ServerAddr consistent with eachother.

https://github.com/microsoft/playwright/blob/10a82f862c0c9b37ad8a35df55a1da42e3ccd3d0/src/server/network.ts#L364

pavelfeldman (approved):

PR #7151, opened 2021-06-15
fix(test-runner): case insensitive test filtering

### LINKED ISSUE #7039: [Feature] Ignore capitalization when matching file names
Since file-systems didn't take the capitalization in consideration in the past and it's easy to miss-spell it, we should maybe also ignore it when matching against the file-names.

Jest also doesn't take it in consideration.

JoelEinbinder:
fixes #7039

This makes regular test filters case insensitive, while it leaves regex behavior unchanged.

I think this is fine, but we could also try to match the current file system case sensitivity.


mxschmitt (approved):
nice

PR #7159, opened 2021-06-15
devops: fix reporting for chromium-with-symbols

aslushnikov:
Chromium-with-symbols is the first time we produce a second set of
builds for the same browser.

This patch adds support for this usecase to report build progress
appropriately.


browser_patches/checkout_build_archive_upload.sh
```diff
@@ -397,18 +413,26 @@ BUILD_ALIAS="$BUILD_FLAVOR r$BUILD_NUMBER"
...
+  (
+    for i in $(cat "${BROWSER_NAME}/${BUILDS_LIST}"); do
+      URL="${HOST}/${BROWSER_NAME}/${BUILD_NUMBER}/$i"
+      if ! [[ $(curl -s -L -I $URL | head -1 | cut -f2 -d' ') == 200 ]]; then
+        # Exit subshell
+        echo "Missing build at ${URL}"
+        exit
+      fi
+    done;
+    LAST_COMMIT_MESSAGE=$(git log --format=%s -n 1 HEAD -- "./${BROWSER_NAME}/BUILD_NUMBER")
```
yury-s, line 433: Unrelated to this PR but this message oftentimes is unrelated to the last browser patch, maybe we can fix that too?

yury-s (approved):

browser_patches/checkout_build_archive_upload.sh
aslushnikov: > Unrelated to this PR but this message oftentimes is unrelated to the last browser patch, maybe we can fix that too?

Somehow git on mingw ignores arguments after the `--` - I'll look into that separately (requires booting win)

PR #7165, opened 2021-06-16
feat(inspector): pause on accessor calls like `page.textContent('div')`

dgozman:
- Also capture snapshot with target element highlighted in tracing.
- Also fix a bug where `timeout` option is not reset to zero under `PWDEBUG`.
- Note this removes "atomicity" of accessor methods, but so far we did not encounter issues with that in other methods like click.


tests/page/elementhandle-convenience.spec.ts
```diff
@@ -69,94 +69,6 @@ it('textContent should work', async ({ page, server }) => {
   expect(await page.textContent('#inner')).toBe('Text,\nmore text');
 });
 
-it('textContent should be atomic', async ({ playwright, page }) => {
```
pavelfeldman, line 72: what's up with these?

PR #7171, opened 2021-06-16
feat(cli codegen): added authentication proxy function

szktmyk38f:
The current `codegen` specification has failed to break through proxies that require authentication.
Therefore, by adding the `--proxy-username` and `--proxy-password` options, codegen can be used even for proxies that require authentication.

example:
```bash
npx playwright codegen --proxy-server "https://proxy.xxx.yyy.zz:8080" --proxy-username "xxxxx" --proxy-password "yyyyy"
```

### DIFF
--- a/src/cli/cli.ts
+++ b/src/cli/cli.ts
@@ -296,6 +296,8 @@ type Options = {
   lang?: string;
   loadStorage?: string;
   proxyServer?: string;
+  proxyUsername?: string;
+  proxyPassword?: string;
   saveStorage?: string;
   timeout: string;
   timezone?: string;
@@ -341,7 +343,9 @@ async function launchContext(options: Options, headless: boolean, executablePath
 
   if (options.proxyServer) {
     launchOptions.proxy = {
-      server: options.proxyServer
+      server: options.proxyServer,
+      username: options.proxyUsername,
+      password: options.proxyPassword
     };
   }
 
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):

src/cli/cli.ts
```diff
@@ -341,7 +343,9 @@ async function launchContext(options: Options, headless: boolean, executablePath
 
   if (options.proxyServer) {
     launchOptions.proxy = {
-      server: options.proxyServer
+      server: options.proxyServer,
+      username: options.proxyUsername,
+      password: options.proxyPassword
```
mxschmitt, line 348:
```suggestion
     ...(options.proxyUsername && {username: options.proxyUsername}),
     ...(options.proxyPassword && {password: options.proxyPassword}),
```
Something like this should make the bots happy. They are trying to assert the generated code to the code in the tests and have probably currently undefined or null values from it.

szktmyk38f: Thank you. As you pointed out, the added parameter was giving an error in the test.

pavelfeldman:
`http://user:password@domain.com/` is a semi-standard way of passing auth for proxies. Let's use it?

szktmyk38f:
With the current codegen specification, even if you write in that notation, you cannot break through the proxy. 
Have you tried it? I couldn't do it in my environment.
Therefore, you need to put the parameters in username and password.

pavelfeldman:
What I meant was that you accept value in the form of "http://user:password@domain.com/" and then parse it using `new URL()` to retrieve username and password.

szktmyk38f:
So that's it. understood. That is simpler and easier to use.

mxschmitt:
Closed since the PR seems stale, please create a new one with the fixes if you are still interested in contributing to it.

PR #7172, opened 2021-06-16
chore: rebase documentation and markdown classes from docs

mxschmitt:
Before we had a copy for markdown/documentation.js classes in our docs. They keep changing on both sides and its hard to maintain and rebase manually. With this change I updated/merged it with the classes from playwright.dev and added on the docs site a script which will sync the classes on each roll. See here: https://github.com/microsoft/playwright.dev/pull/249

This might break Java since I reverted: https://github.com/microsoft/playwright/pull/5629/files#diff-a98c2c034c3d7cc33f659fb7c6d1615f4b6377d9d3cb5a1a0a9443e8b24b8d32L230

### DIFF
--- a/utils/markdown.js
+++ b/utils/markdown.js
@@ -227,7 +227,7 @@ function render(nodes, maxColumns) {
  */
 function innerRenderMdNode(indent, node, lastNode, result, maxColumns) {
   const newLine = () => {
-    if (result.length && result[result.length - 1] !== '')
+    if (result[result.length - 1] !== '')
       result.push('');
   };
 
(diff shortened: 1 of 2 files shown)


utils/markdown.js
```diff
@@ -227,7 +227,7 @@ function render(nodes, maxColumns) {
  */
 function innerRenderMdNode(indent, node, lastNode, result, maxColumns) {
   const newLine = () => {
-    if (result.length && result[result.length - 1] !== '')
```
mxschmitt, line 230: Referring to this change @yury-s.

pavelfeldman (approved):

utils/markdown.js
yury-s: I ran java api generator with api.json produced after this patch has been applied and it produced the same result as before (no diffs) so it is good to go.

yury-s (approved):

PR #7192, opened 2021-06-17
feat(test-runner): tags

### LINKED ISSUE #7169: [Feature]: Add tag support for test-runner
Hi Folks,

I have a question, I have written a single test case inside a single .ts file, So if I want to run 2 test cases I must run two .ts files. So my question is there a way to tag only these 2 .ts files and run? Is there any other way?

Like for example in a folder I have 10 .ts files but I want to run only two of them.

Thanks in Advance.

JoelEinbinder:
I am not sure what we want the api to look like for describes so I left it out of this patch. The implementation is all there though.﻿

### DIFF
--- a/src/test/cli.ts
+++ b/src/test/cli.ts
@@ -55,6 +55,7 @@ export function addTestCommand(program: commander.CommanderStatic) {
   command.option('--retries <retries>', `Maximum retry count for flaky tests, zero for no retries (default: no retries)`);
   command.option('--shard <shard>', `Shard tests and execute only the selected shard, specify in the form "current/all", 1-based, for example "3/5"`);
   command.option('--project <project-name>', `Only run tests from the specified project (default: run all projects)`);
+  command.option('-t, --tag <tag...>', `Only run tests with the specified tag`);
   command.option('--timeout <timeout>', `Specify test timeout threshold in milliseconds, zero for unlimited (default: ${defaultTimeout})`);
   command.option('-u, --update-snapshots', `Update snapshots with actual results (default: only create missing snapshots)`);
   command.option('-x', `Stop after the first failure`);
@@ -126,8 +127,7 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
     // If not, scan the world.
     runner.loadEmptyConfig(process.cwd());
   }
-
-  const result = await runner.run(!!opts.list, args.map(forceRegExp), opts.project || undefined);
+  const result = await runner.run(!!opts.list, args.map(forceRegExp), opts.tag || [], opts.project || undefined);
   if (result === 'sigint')
     process.exit(130);
   process.exit(result === 'passed' ? 0 : 1);
--- a/src/test/runner.ts
+++ b/src/test/runner.ts
@@ -21,7 +21,7 @@ import * as path from 'path';
 import { promisify } from 'util';
 import { Dispatcher } from './dispatcher';
 import { createMatcher, monotonicTime, raceAgainstDeadline } from './util';
-import { Suite } from './test';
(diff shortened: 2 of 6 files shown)


src/test/cli.ts
```diff
@@ -55,6 +55,7 @@ export function addTestCommand(program: commander.CommanderStatic) {
   command.option('--retries <retries>', `Maximum retry count for flaky tests, zero for no retries (default: no retries)`);
   command.option('--shard <shard>', `Shard tests and execute only the selected shard, specify in the form "current/all", 1-based, for example "3/5"`);
   command.option('--project <project-name>', `Only run tests from the specified project (default: run all projects)`);
+  command.option('-t, --tag <tag...>', `Only run tests with the specified tag`);
```
pavelfeldman, line 58: Is that || or &&? I think both make sense and will be requested.

src/test/runner.ts
```diff
@@ -250,6 +253,23 @@ function filterOnly(suite: Suite) {
   return false;
 }
 
+function filterTags(suite: Suite, tags: string[]): boolean {
+  function tagsMatch(a: string[], b: Spec|Suite) {
```
pavelfeldman, line 257: So that is &&. Call it `hasAllTags(spec, string[])`?

src/test/test.ts
```diff
@@ -26,9 +26,15 @@ class Base {
   parent?: Suite;
 
   _only = false;
+  _tags: Set<string>;
 
   constructor(title: string) {
-    this.title = title;
+    this.title = title.split(' ').filter(x => !x.startsWith('#')).join(' ');
```
pavelfeldman, line 32: Let's be more strict? `'#smoke #perf #fast Test something'.match(/((?:(?:#[^\s]+)[\s])+)(.*)*/)`

src/test/test.ts
```diff
@@ -26,9 +26,15 @@ class Base {
...
+    this._tags = tagsFromTitle(title);
+  }
+
+  hasTag(tag: string): boolean {
+    return this._tags.has(tag) || (!!this.parent && this.parent.hasTag(tag));
```
pavelfeldman, line 37: `this.parent?.hasTag(tag)`

tests/playwright-test/tags.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+import { test, expect } from './playwright-test-fixtures';
+
+test('should work', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.test.js': `
+      const { test } = pwt;
+      test('test 1', async ({}, testInfo) => {
+        throw new Error('do not run this');
+      });
+      test('test 2 #smoke', async ({}, testInfo) => {
```
pavelfeldman, line 26: In reality, names are much longer than tags, so I would lead w/ tags

tests/playwright-test/tags.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  expect(result.exitCode).toBe(0);
+  expect(result.output).toContain('I am great');
+  expect(result.output).not.toContain('I am #not great');
+});
+
+test('tags should be able to come from extend', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.test.js': `
+      const { test } = pwt;
+      const tagged = test.extend({}, {tag: 'smoke'})
```
pavelfeldman, line 75: I know I told you to rock'n'roll, but this really feels uncalled for. Can we wait for a request? If we already had options it would be a super-easy sell.

akshayp7: thanks for looking into this feature !!!!!

PR #7194, opened 2021-06-17
devops: support `npx playwright install --with-deps` command

aslushnikov:
The common pattern today is to have two commands to configure CI:

```sh
npx playwright install-deps chromium
npx playwright install chromium
```

With this patch, this becomes:

```sh
npx playwright install --with-deps chromium
```

Note: `--with-deps` might call `sudo`.


yury-s (approved):
Consider installing deps by default and overriding that behavior with --no-deps

PR #7208, opened 2021-06-17
feat(test-runner): allow to focus a test in a location

mxschmitt:
~~Pavel's idea to only have it matched by the line instead of line + column. Any thoughts?~~

Needed for the Visual Studio Code extension.

### DIFF
--- a/src/test/cli.ts
+++ b/src/test/cli.ts
@@ -22,6 +22,7 @@ import * as path from 'path';
 import type { Config } from './types';
 import { Runner } from './runner';
 import { stopProfiling, startProfiling } from './profiler';
+import type { FilePatternFilter } from './util';
 
 const defaultTimeout = 30000;
 const defaultReporter = process.env.CI ? 'dot' : 'list';
@@ -131,7 +132,15 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
     runner.loadEmptyConfig(process.cwd());
   }
 
-  const result = await runner.run(!!opts.list, args.map(forceRegExp), opts.project || undefined);
+  const filePatternFilters: FilePatternFilter[] = args.map(arg => {
+    const splitted = (typeof arg === 'string' && arg.includes(':')) ? arg.split(':') : null;
+    const containsLineNumber = splitted?.length === 2;
+    return {
+      re: forceRegExp(containsLineNumber ? splitted![0] : arg),
+      line: containsLineNumber ? parseInt(splitted![1], 10) : null,
+    };
+  });
+  const result = await runner.run(!!opts.list, filePatternFilters, opts.project || undefined);
   await stopProfiling(undefined);
 
   if (result === 'sigint')
--- a/src/test/cli.ts
+++ b/src/test/cli.ts
@@ -133,11 +133,10 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
   }
 
   const filePatternFilters: FilePatternFilter[] = args.map(arg => {
-    const splitted = (typeof arg === 'string' && arg.includes(':')) ? arg.split(':') : null;
-    const containsLineNumber = splitted?.length === 2;
+    const match = /^(.*):(\d+)$/.exec(arg);
     return {
-      re: forceRegExp(containsLineNumber ? splitted![0] : arg),
-      line: containsLineNumber ? parseInt(splitted![1], 10) : null,
+      re: forceRegExp(match ? match[1] : arg),
+      line: match ? parseInt(match[2], 10) : null,
     };
   });
   const result = await runner.run(!!opts.list, filePatternFilters, opts.project || undefined);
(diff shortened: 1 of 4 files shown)


src/test/cli.ts
```diff
@@ -44,6 +44,7 @@ export function addTestCommand(program: commander.CommanderStatic) {
   command.option('-c, --config <file>', `Configuration file, or a test directory with optional "${tsConfig}"/"${jsConfig}"`);
   command.option('--forbid-only', `Fail if test.only is called (default: false)`);
   command.option('-g, --grep <grep>', `Only run tests matching this regular expression (default: ".*")`);
+  command.option('-gTL, --grepTestLocations <grep>', `Only run tests by the file location. Comma separated values in the form of line:column.`);
```
pavelfeldman, line 47: I would not do that, i'd rather accept plain argument in the form of page.spec.ts:45

src/test/runner.ts
```diff
@@ -254,6 +255,23 @@ function filterOnly(suite: Suite) {
   return false;
 }
 
+function filterTestFileByLine(suite: Suite, focusedTestFileLines: FilePatternFilter[]) {
+  const testFileLineMatches = (specFileName: string, specLine: number) => focusedTestFileLines.some(({re, line}) => {
+    re.lastIndex = 0;
+    return re.test(specFileName) && line === specLine
```
dgozman, line 261: What happened to the "missing semicolons" rule? :)

src/test/cli.ts
```diff
@@ -131,7 +132,15 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
     runner.loadEmptyConfig(process.cwd());
   }
 
-  const result = await runner.run(!!opts.list, args.map(forceRegExp), opts.project || undefined);
+  const filePatternFilters: FilePatternFilter[] = args.map((arg) => {
+    const splitted = (typeof arg === 'string' && arg.includes(':')) ? arg.split(':') : null;
```
dgozman, line 136: Since we support regex here, I wonder how something like `/[ab:]/` is going to work? Perhaps, we should treat the string as a regexp, but match `file:line:column` instead of matching just `file`?

src/test/runner.ts
```diff
@@ -254,6 +255,23 @@ function filterOnly(suite: Suite) {
   return false;
 }
 
+function filterTestFileByLine(suite: Suite, focusedTestFileLines: FilePatternFilter[]) {
```
dgozman, line 258: I see a lot of similarity to `filterOnly`. Let's maybe generalize to `filterSuite(suite, Spec => boolean)`?

src/test/util.ts
```diff
@@ -139,17 +139,26 @@ export function isRegExp(e: any): e is RegExp {
...
   const filePatterns: string[] = [];
   for (const pattern of Array.isArray(patterns) ? patterns : [patterns]) {
     if (isRegExp(pattern)) {
       reList.push(pattern);
     } else {
-      if (!pattern.startsWith('**/') && !pattern.startsWith('**/'))
-        filePatterns.push('**/' + pattern);
-      else
-        filePatterns.push(pattern);
+      filePatterns.push(cleanFilePattern(pattern));
```
dgozman, line 161: Note that this also affects testIgnore and testMatch. Do we want the same logic there?

src/test/runner.ts
mxschmitt: Was busted on ToT, got fixed in #7268.

src/test/cli.ts
mxschmitt: for now I didn't add support for handling regular expressions. I think thats not needed there, since regexp would mean run it on multiple files and files usually are differently structured.

src/test/runner.ts
mxschmitt: This is a great idea.

src/test/util.ts
mxschmitt: I would say no, so reverted it.

src/test/runner.ts
```diff
@@ -243,8 +244,24 @@ export class Runner {
...
-  const onlyTests = suite.specs.filter(spec => spec._only);
+  const suiteFilter = (suite: Suite) => suite._only;
+  const specFilter = (spec: Spec) => spec._only;
+  return filterSuite(suite, suiteFilter, specFilter);
+}
+
+function filterByFocusedLine(suite: Suite, focusedTestFileLines: FilePatternFilter[]) {
+  const testFileLineMatches = (specFileName: string, specLine: number) => focusedTestFileLines.some(({re, line}) => {
+    re.lastIndex = 0;
+    return re.test(specFileName) && line === specLine;
```
dgozman, line 255: I wonder how this works with null line? For example, `npx playwright test a.spec.ts:80 b.spec.ts` - will this run one test + all tests?

dgozman (approved):

src/test/runner.ts
mxschmitt: The extension probably won't need it but to be feature-complete I added support and a test for that too.

Issue #7030, opened 2021-04-18
[Feature] Track Code Coverage via istanbul


mxschmitt:
### Update: It can be easily integrated in the user-land now.

For that follow the guide [here](https://github.com/mxschmitt/playwright-test-coverage).

For any suggestions please comment inside this issue.

barakbs1:
We currently use jest-playwrights' code coverage feature and we depend on it. We would love to migrate to new playwright test-runner but we have to have the code coverage feature, and would really hope it would be integrated. Keep up the amazing work guys!

mpuertao:
Thanks, i need coverage with istanbul. stay tuned

mxschmitt:
Small update, we created [an example guide](https://gist.github.com/mxschmitt/3d5c27b8410ce27695df0ef2b9a575a3) how you can use it. It is currently there to get an understanding which additional features are necessary to provide a full story around code coverage collections. For any suggestions / feedback it would be highly appreciated to get input especially:

- If you are using the babel-plugin-istanbul?
- If single browser code coverage like Chromium would be enough for you?
- If you would prefer an approach with minified JS + source maps instead of using the babel plugin? (this would require to have the source maps either locally or available on the web-server)
- What are you doing with the coverage data? text-summary, uploading it to a provider, serving the html, only for manual checking etc.

mpuertao:
For me:

- If you are using the babel-plugin-istanbul? R:// Yes.
- If single browser code coverage like Chromium would be enough for you? R:// Yes, the coverage is a quantitative aspect, and with a single browser I can determine what components I am covering from the E2E tests.
- What are you doing with the coverage data? text-summary, uploading it to a provider, serving the html, only for manual checking etc. R:// Generate lcov.info o json o html, we are send this data to elasticsearch and after grafana

anfruiz:
I implemented the baseFixtures.ts in the test folder and I create my test like this: 
`import { test } from "./baseFixtures";
test(`P`, async ({ page }) => {
  await page.goto("http://localhost:3000");
  const commandLine = await page.innerText("p");
  test.expect(commandLine).toBe("Edit src/App.js and save to reload.");
});
`

mxschmitt:
How did you add the babel-plugin? This requires a transcompilation step which is not out-of-the box available. This looks like create-react-app and needs by that an ejected installation.

anfruiz:
> This requires a transcompilation step which is not out-of-the box available

[...]

PR #7209, opened 2021-06-17
feat(test-runner): added Istanbul coverage collection

mxschmitt:
An experiment how it i works in general, implementation etc. is TBD. There are two way so far how to add it:
a) Chromium coverage with [v8-to-istanbul](https://www.npmjs.com/package/v8-to-istanbul) package and save the individual reports and then let the user merge it via [nyc](https://www.npmjs.com/package/nyc)
b) Use [babel-plugin-instanbul](https://github.com/istanbuljs/babel-plugin-istanbul) and save the individual reports and then let the user merge it via [nyc](https://www.npmjs.com/package/nyc)

We don't want the `nyc` dependency since it introduces too many dependencies.

### DIFF
--- a/src/test/index.ts
+++ b/src/test/index.ts
@@ -25,6 +25,7 @@ export { expect } from './expect';
 export const _baseTest: TestType<{}, {}> = rootTestType.test;
 
 const artifactsFolder = path.join(os.tmpdir(), 'pwt-' + createGuid());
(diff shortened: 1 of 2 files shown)


mxschmitt:
Closed, done in user-land for now: https://github.com/microsoft/playwright/issues/7030#issuecomment-867174653

PR #7212, opened 2021-06-17
fix(docker): use node 14 not node 16

JoelEinbinder:
The current LTS is Node 14. Node 16 is a bit unstable to ship to users.

### DIFF
--- a/utils/docker/Dockerfile.bionic
+++ b/utils/docker/Dockerfile.bionic
@@ -2,9 +2,9 @@ FROM ubuntu:bionic
 
 # === INSTALL Node.js ===
 
-# Install node16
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

sreuss:
When will this change find it's way into the published image(s)?

The 1.12.3 release does not appear to incorporate this change:

```
docker run -it --rm mcr.microsoft.com/playwright:v1.12.3-focal
root@2e19c6dca57b:/# node --version
v16.4.0
```

mxschmitt:
You can use `mcr.microsoft.com/playwright:next-focal`. It gets shipped in v1.13.

PR #7202, opened 2021-06-17
browser(webkit): roll to 06/17/21

dpino:
This pull-request rebases `webkit/bootstrap.diff` to webkit ToT (a60349c98390c46d865cbaa55431b074c2144639).

It was necessary to disable Cog compilation as it was producing a compilation error in WPE.

PR #7216, opened 2021-06-18
Follow-up #7202. Use WTFMove to avoid extra copy.

dpino:
Applies change suggested in https://github.com/microsoft/playwright/pull/7202#discussion_r653723896.


mxschmitt:
When making changes to browsers, the export script needs to be executed, so that a new build number gets created automatically. A build number is our internal representation of a browser build in the tree which gets stored on the CDN. See [here](https://github.com/microsoft/playwright/tree/master/browser_patches#3-exporting-your-change-to-playwright-repo) about more information regarding the `export.sh` script. In your case `./browser_patches/export.sh webkit` should do it.

Also we would use the following commit style / PR title for that: `browser(webkit): use WTFMove toavoid extra copy` and put the previous PR number where you are referencing to into the commit body.

mxschmitt:
Closed since it was done in #7278.

PR #7225, opened 2021-06-18
fix(test-runner): allow viewport null values

### LINKED ISSUE #7223: [BUG] test-runner viewport can't be null
```ts
    {
      name: 'Chromium',
      use: {
        browserName: 'chromium',
        // Make Chromium maximized
        // @ts-ignore
        viewport: null,
        launchOptions: {
          headless: false,
          args: ["--start-maximized"]
        },
      },
    },
  ],
```

See the ts-ignore.

If we go with the #7221 approach, then we don't have to fix that.

mxschmitt:
Fixes #7223

### DIFF
--- a/types/test.d.ts
+++ b/types/test.d.ts
@@ -1139,7 +1139,7 @@ export type PlaywrightTestOptions = {
    * Viewport used for all pages in the test. Takes priority over `contextOptions`.
    * @see BrowserContextOptions
    */
-  viewport: ViewportSize | undefined;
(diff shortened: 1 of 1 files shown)


JoelEinbinder (approved):
This works, only typescript complains?

mxschmitt:
> This works, only typescript complains?

Yes we have a strict `undefined` check in place:

https://github.com/microsoft/playwright/blob/388c6f51963d96f9e43a284db9676290679fe53c/src/test/index.ts#L139-L140

PR #7234, opened 2021-06-19
docs(python): enhance intro

mxschmitt:
Since a few useful information pieces got removed in https://github.com/microsoft/playwright-python/pull/761 and moved over to the docs, I followed up with adding the missing parts. Especially the REPL usage is important for the Jupyter / REPL users.

### DIFF
--- a/docs/src/intro-python.md
+++ b/docs/src/intro-python.md
@@ -12,12 +12,16 @@ See [system requirements](#system-requirements).
 
 ### Pip
 
+[![PyPI version](https://badge.fury.io/py/playwright.svg)](https://pypi.python.org/pypi/playwright/)
+
 ```bash
 pip install playwright
 playwright install
 ```
 
 ### Conda
+
+[![Anaconda version](https://img.shields.io/conda/v/microsoft/playwright)](https://anaconda.org/Microsoft/playwright)
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/src/intro-python.md
```diff
@@ -89,6 +93,54 @@ Command Line Interface [CLI](./cli.md) can be used to record user interactions a
...
+>>> await page.screenshot(path="example.png")
+>>> await browser.close()
+>>> await playwright.stop()
+```
+
+## Known issues
+
+### `time.sleep()` leads to outdated state
+
+You need to use `page.wait_for_timeout(5000)` instead of `time.sleep()`. It is better to not wait for a timeout at all, but sometimes it is useful for debugging. In these cases, use our wait method instead of the system one. This is because we internally rely on asynchronous operations and when using `time.sleep(5)` they can't get processed correctly.
```
kumaraditya303, line 142:
```suggestion
You should use `page.wait_for_timeout(5000)` instead of `time.sleep(5)` and it is better to not wait for a timeout at all, but sometimes it is useful for debugging. In these cases, use our wait method instead of the `time` module. This is because we internally rely on asynchronous operations and when using `time.sleep(5)` they can't get processed correctly.
```

yury-s (approved):

PR #7243, opened 2021-06-21
fix(test runner): properly keep track of requireFile to support helpers/wrappers

dgozman:
This fixes an issue where we incorrectly labeled and assigned ids for tests that declared tests in require'd files or used test wrappers.
See new tests for examples.

### DIFF
--- a/src/test/testType.ts
+++ b/src/test/testType.ts
@@ -63,10 +63,11 @@ export class TestTypeImpl {
       throw errorWithCallLocation(`test() can only be called in a test file`);
     const location = callLocation(suite.file);
 
-    const ordinalInFile = countByFile.get(suite.file) || 0;
-    countByFile.set(location.file, ordinalInFile + 1);
+    const ordinalInFile = countByFile.get(suite._requireFile) || 0;
+    countByFile.set(suite._requireFile, ordinalInFile + 1);
 
     const spec = new Spec(title, fn, ordinalInFile, this);
+    spec._requireFile = suite._requireFile;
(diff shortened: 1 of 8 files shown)


tests/playwright-test/basic.spec.ts
```diff
@@ -251,3 +251,103 @@ test('should work with default export', async ({ runInlineTest }) => {
   expect(result.passed).toBe(1);
   expect(result.failed).toBe(0);
 });
+
+test('should work with test wrapper', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
```
JoelEinbinder, line 256: How did this fail before?

dgozman: "2 passed, 2 skipped" instead of "4 passed"

src/test/testType.ts
```diff
@@ -63,10 +63,11 @@ export class TestTypeImpl {
       throw errorWithCallLocation(`test() can only be called in a test file`);
     const location = callLocation(suite.file);
 
-    const ordinalInFile = countByFile.get(suite.file) || 0;
-    countByFile.set(location.file, ordinalInFile + 1);
```
JoelEinbinder, line 67: this mismatch with `suite.file` and `location.file` was just broken before right?

JoelEinbinder (approved):

src/test/testType.ts
dgozman: Yeah...

PR #7251, opened 2021-06-21
fix: Jest isError handling to work with TypeError

### LINKED ISSUE #7238: [BUG] When serializing internal errors with Jest and Playwright it leads to an exception
Repro: https://github.com/mxschmitt/jest-microtask-issue

Expected: throw a `browserType.connectOverCDP: Invalid URL: undefined` exception
Actual: Getting the attached exception, because the message attribute is undefined.

```log
    TypeError: Cannot read property 'length' of undefined
        at Object.rewriteErrorMessage (/Users/max/development/jest-microtask-issue/node_modules/playwright-chromium/lib/utils/stackTrace.js:35:102)
        at BrowserType._wrapApiCall (/Users/max/development/jest-microtask-issue/node_modules/playwright-chromium/lib/client/channelOwner.js:83:26)
        at Object.<anonymous> (/Users/max/development/jest-microtask-issue/issue.test.js:9:5)
```

Version: latest next release.

Without Jest it works as expected, maybe it has to do something with our microtask scheduling.

Related to #7094

mxschmitt:
Fixes #7238

### DIFF
--- a/src/protocol/serializers.ts
+++ b/src/protocol/serializers.ts
@@ -151,5 +151,5 @@ function isDate(obj: any): obj is Date {
 }
 
 function isError(obj: any): obj is Error {
-  return obj instanceof Error || (obj && obj.__proto__ && obj.__proto__.name === 'Error');
+  return obj instanceof Error || obj?.__proto__?.name === 'Error' || (obj?.__proto__ && isError(obj.__proto__));
 }


src/protocol/serializers.ts
```diff
@@ -151,5 +151,5 @@ function isDate(obj: any): obj is Date {
 }
 
 function isError(obj: any): obj is Error {
-  return obj instanceof Error || (obj && obj.__proto__ && obj.__proto__.name === 'Error');
+  return obj instanceof Error || obj && obj.__proto__.name === 'Error' || obj && isError(obj.__proto__);
```
JoelEinbinder, line 154: Was missing a check for `obj.__proto__` truthiness. People can throw an object with `{ __proto__: null }`.

```suggestion
  return (obj instanceof Error) || (obj && obj.__proto__ && obj.__proto__.name === 'Error') || (obj && isError(obj.__proto__));
```

JoelEinbinder (approved):

src/protocol/serializers.ts
mxschmitt: clarified offline: optional chaining solution is fine.

PR #7122, opened 2021-06-14
feat: response interception

yury-s:
Initial API and implementation for Chromium and WebKit.

#1774

PR #7255, opened 2021-06-22
test: skip some failing request interception tests for current Chrome stable

mxschmitt:
Some tests are currently failing Chrome/Edge stable and Electron:

- page/page-request-intercept.spec.ts:60:1 › [chromium] should support fulfill after intercept
- page/page-request-intercept.spec.ts:73:1 › [chromium] should support request overrides

Was introduced in #7122. This change should give us again full green bots.


aslushnikov (approved):

dgozman:
Any ideas why does this fail? We don't use any new protocol commands here, do we?

PR #7265, opened 2021-06-23
fix(test runner): expose real stack traces and speed up locations

dgozman:
Stop wrapping/prepending error messages so that we do not loose the stack trace. For this, update a few manually thrown errors with better messages (usually including a file path).

Speed up locations by doing manual `sourceMapSupport.wrapCallSite()` for a single call site. Performance gain in the runner process with 100 files x 100 tests each:
- 25% on the fresh run without babel cache;
- 80% on the cached run where babel is almost instant.

Also some obvious cleanups around stack traces (removing unused code).


tests/playwright-test/playwright-test-fixtures.ts
```diff
@@ -225,7 +225,7 @@ export const test = base.extend<Fixtures>({
       runResult = await runPlaywrightTest(baseDir, params, env);
       return runResult;
     });
-    if (testInfo.status !== testInfo.expectedStatus && runResult)
+    if (testInfo.status !== testInfo.expectedStatus && runResult && !process.env.PW_RUNNER_DEBUG)
```
JoelEinbinder, line 228: :*( I found this always useful that it printed the output.

tests/playwright-test/fixture-errors.spec.ts
```diff
@@ -367,7 +367,7 @@ test('should print nice error message for problematic fixtures', async ({ runInl
   });
   expect(result.exitCode).toBe(1);
   expect(result.output).toContain('oh my!');
-  expect(result.output).toContain('x.spec.ts:5:29');
+  expect(result.output).toContain('x.spec.ts:6:49');
```
JoelEinbinder, line 370: the new column looks right but why did this change? Its hard for me to verify the line number.

tests/playwright-test/playwright-test-fixtures.ts
dgozman: This prevents double test output when using `PW_RUNNER_DEBUG` env.

tests/playwright-test/fixture-errors.spec.ts
dgozman: It was using location of `.extend()` instead of location of `new Error()`.

tests/playwright-test/loader.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        foo: 'bar';
+      };
+    `
+  }, {}, {
+    FORCE_COLOR: '0'
+  });
+  expect(result.exitCode).toBe(1);
+  expect(result.passed).toBe(0);
+  expect(result.failed).toBe(0);
+  expect(result.output).toContain('error.spec.ts');
```
JoelEinbinder, line 70: This should have the location not just the file!

dgozman: It does, but babel error format is different, something like `error.spec.ts: Some text at (6; 5)`

src/test/transform.ts
```diff
@@ -21,11 +21,15 @@ import * as fs from 'fs';
 import * as pirates from 'pirates';
 import * as babel from '@babel/core';
 import * as sourceMapSupport from 'source-map-support';
+import type { Location } from './types';
 
 const version = 4;
 const cacheDir = process.env.PWTEST_CACHE_DIR || path.join(os.tmpdir(), 'playwright-transform-cache');
 const sourceMaps: Map<string, string> = new Map();
 
+const kStackTraceLimit = 15;
```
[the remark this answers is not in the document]
dgozman: This is just a move from `testType.ts`. Let me address that separately, just in case.

JoelEinbinder (approved):

PR #7274, opened 2021-06-23
update(showcase): add playwright-expect library and playwright test boilerplate

elaichenkov:
**update(showcase):** 
 - add playwright-expect library 
 - add playwright test boilerplate


mxschmitt:
Thank you for your initiative! Since we are internally adding expect matchers in the next release for Playwright Test and there is already a community driven library called [expect-playwright](https://github.com/playwright-community/expect-playwright) I'm closing it.

PR #7278, opened 2021-06-23
browser(webkit): roll to 06/23/21

q66:
my branch with merge conflicts resolved: https://github.com/q66/WebKit/tree/playwright

There is a new potential problem - libwebrtc in webkit recently updated, which brings a new dependency on libvpx at least 1.10. Linux distros generally do not ship this version (it hasn't changed soname, but as far as I'm aware there was an ABI breakage). I've done my build with a custom libvpx build, but this is probably going to be problematic.

Related to this, I found that trying to disable webrtc results in undefined references to libvpx functions in unrelated Microsoft code. That probably shouldn't happen, so ideally the build system should get fixed.

Also, running the tests results in some failures: https://gist.github.com/q66/e5e89ade2e49df55a4799c0aad520a6a


aslushnikov:
> There is a new potential problem - libwebrtc in webkit recently updated, which brings a new dependency on libvpx at least 1.10. Linux distros generally do not ship this version (it hasn't changed soname, but as far as I'm aware there was an ABI breakage). I've done my build with a custom libvpx build, but this is probably going to be problematic.

@q66 Can the new libvpx be added upstream to the JHBuild?

> Also, running the tests results in some failures: https://gist.github.com/q66/e5e89ade2e49df55a4799c0aad520a6a

- The `route.intercept is not a function` errors happened probably because you didn't re-compile the playwright, yet pulled a new one in. This function has been landed recently (this monday) - https://github.com/microsoft/playwright/commit/47341dc1a9716230b6a43405951e50fcbc18321e
- The viewport changes in `browsercontext-basic.spec.ts:89:1 › [webkit] should propagate default viewport to the page` shouldn't happen. Looks like this roll changed something and requires investigation.
- Failed screenshots are probably related to viewport changes

q66:
> > There is a new potential problem - libwebrtc in webkit recently updated, which brings a new dependency on libvpx at least 1.10. Linux distros generally do not ship this version (it hasn't changed soname, but as far as I'm aware there was an ABI breakage). I've done my build with a custom libvpx build, but this is probably going to be problematic.
> 
> @q66 Can the new libvpx be added upstream to the JHBuild?

jhbuild doesn't seem to have a module for libvpx, but webkit, or rather libwebrtc, ships an entire copy (the bundled copy is used on macos already); i wonder if that could be used (would require modifications to the cmake buildsystem...)

the general stance appears to be "you're not supposed to use libwebrtc with webkit" (there are reasons for that, like this one and that it causes licensing concerns as boringssl is not GPL-compatible) and it's actually not explicitly enabled by playwright, but rather is pulled in with other experimental features since it's a developer build - do you think we could get away with disabling it entirely, or it is actually needed for something? (we'd need to add buildsystem bits for the microsoft code though - it could probably just keep using an older version like until now)

> 
> > Also, running the tests results in some failures: https://gist.github.com/q66/e5e89ade2e49df55a4799c0aad520a6a
> 
>     * The `route.intercept is not a function` errors happened probably because you didn't re-compile the playwright, yet pulled a new one in. This function has been landed recently (this monday) - [47341dc](https://github.com/microsoft/playwright/commit/47341dc1a9716230b6a43405951e50fcbc18321e)

oh, yeah, that makes sense - indeed i did not recompile playwright after having tested it last friday :) i was just put on the rebasing job, so please do bear with me for a while

> 
>     * The viewport changes in `browsercontext-basic.spec.ts:89:1 › [webkit] should propagate default viewport to the page` shouldn't happen. Looks like this roll changed something and requires investigation.
> 
>     * Failed screenshots are probably related to viewport changes

aslushnikov:
>  i was just put on the rebasing job, so please do bear with me for a while

@q66 sure! Is this ready for review then? Or should I wait for the tests to pass?

q66:
feel free to review/test this yourself, though you will need to deal with the libvpx problem somehow (i'm not 100% sure what the best thing to do would be here - well, the best thing would probably be to disable libwebrtc since it's experimental/unsupported-ish, but i first needed to know if you are actually relying on it)

q66:
indeed, WPE tests come out clean after rebuilding playwright, the GTK tests still fail. I compared some of the screenshots and the differences are fairly slight

q66:
the latest update adds libvpx into jhbuild, so things should now build even on old ubuntu and so on; I also pulled in https://github.com/microsoft/playwright/pull/7216 as instructed by @dpino, and my branch showing how i did the changes has been updated as well.

browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -21534,8 +21534,43 @@ index 46388b47a0f433fbdae3874a958fb2d207916c45..d42933e66178ff007a5008dc807483c4
...
  
+diff --git a/Tools/jhbuild/jhbuild-minimal.modules b/Tools/jhbuild/jhbuild-minimal.modules
+index 545b04743043..dc7ee36b969f 100644
+--- a/Tools/jhbuild/jhbuild-minimal.modules
++++ b/Tools/jhbuild/jhbuild-minimal.modules
+@@ -8,6 +8,7 @@
+       <dep package="wpebackend-fdo"/>
+       <dep package="libsoup"/>
+       <dep package="manette"/>
++      <dep package="libvpx"/>
```
pavelfeldman, line 21545: Should we get this upstream?

q66: i fully intend to, but for now i wanted to get it here since you need it the most right now

aslushnikov (approved):

aslushnikov:
Can you please revert package-lock.json changes?

q66:
oh yeah, sorry, that got in unintentionally

q66:
done

PR #7291, opened 2021-06-24
chore: roll playwright-test to 24-06

mxschmitt:


dgozman (commented):
Why did we pick up so many dependencies?

mxschmitt:
> Why did we pick up so many dependencies?

I don't know, seems like something is messed up. Will check after the meeting.

PR #7298, opened 2021-06-24
fix(test-runner): make it compatible when running as a sub-process

mxschmitt:
Works around the following bug: https://github.com/nodejs/node/issues/12921

TL;DR process.exit() does kill the process right away and does not wait until the parent process has received the data.

I only received the JSON partly which was by that not parsable.

### DIFF
--- a/src/test/runner.ts
+++ b/src/test/runner.ts
@@ -109,14 +109,9 @@ export class Runner {
   async _flushOutput() {
     // Calling process.exit() might truncate large stdout/stderr output.
     // See https://github.com/nodejs/node/issues/6456.
-    //
-    // We can use writableNeedDrain to workaround this, but it is only available
-    // since node v15.2.0.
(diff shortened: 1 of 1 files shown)


dgozman (commented):
- We have `_flushOutput` method that should help in this case.
- We cannot afford to not call `process.exit()` because there might be any random unfinished operations, e.g. setTimeout() called from describe or something like that.

src/test/runner.ts
```diff
@@ -109,14 +109,9 @@ export class Runner {
   async _flushOutput() {
     // Calling process.exit() might truncate large stdout/stderr output.
     // See https://github.com/nodejs/node/issues/6456.
-    //
-    // We can use writableNeedDrain to workaround this, but it is only available
-    // since node v15.2.0.
```
mxschmitt, line 114: Node.js 15+ is not relevant for us since we need to support Node.js 12+. The new fix also works in Node.js 12+.

dgozman (approved):
Nice!

PR #7344, opened 2021-06-27
feat: added more logging info for `$`

### LINKED ISSUE #6162: [Feature] Add more info to pw:api log
I suggest to add more debug info to pw:api output.

As i see this log displays enough info for navigation and wait for selector functions
```
pw:api navigated to "http://localhost:" +86ms 
pw:api => page.waitForSelector started +1ms 
pw:api waiting for selector ".awesome-page" to be visible +2ms 
pw:api selector resolved to visible <div class="awesome-page">-</div> +52ms 
pw:api <= page.waitForSelector succeeded +6ms 
```
Can you add same logging info for `$` and `evaluate ` functions ?

anirudh-modi:
This patch is to close #6162

## Investigation
These are my initial investigation

## Logging `selector` value in `$` invocation
Function to be modified for `$` invocation will be in [[this code](https://github.com/microsoft/playwright/blob/409aeaa4438585f44484bfc9d43e7d51f9b5c22a/src/server/frames.ts#L683)]

## Logging function params and function in `evaluate` invocation
Function to be modified for `evaluate` invocation will be in [[this code](https://github.com/microsoft/playwright/blob/409aeaa4438585f44484bfc9d43e7d51f9b5c22a/src/server/dom.ts#L54)]

## How to add a new log?
From what I could understand is that we will have to use the `debugLogger.log` function for logging, as shown in [[this code](https://github.com/microsoft/playwright/blob/409aeaa4438585f44484bfc9d43e7d51f9b5c22a/src/server/frames.ts#L206)]

@mxschmitt @JoelEinbinder  @dgozman These are my finding based on the investigation. There were other files with similar function name, but based on the `client`, `dispatcher` and `server`, I felt these are the place where I need to change. 

The issue also asked for `fill` however, when I turned on the `pw:api` mode it was already sending appropriate output, and hence I have not added any log for the same.

I believe there needs to be some sort of test case for the same?

Please, let me know, if I have missed anything for the same or not.  So that I can fix them and raise a PR with appropriate changes


anirudh-modi:
Hi @pavelfeldman is there anything else required on this PR in your opinion? I have also added my reasoning behind how I added a new log.

I could not find any existing relevant tests which I could refer to.. so if the tests are required it will be be helpful if you can point me to the same file.

dgozman:
@anirudh-modi Thank you for the PR! Some high-level comments:
- I am not sure we'd like to see the full `expression` when evaluating. These expressions may be quite big, and it will be impossible to read the log. Unfortunately, I don't have a nice solution for this. I'd suggest to trim the expression, but then it won't be useful at all.
- Logs for `$` look good!
- You'll have to sign the CLA to be able to contribute.
- Here is a test that checks a similar log, you can do something similar: https://github.com/microsoft/playwright/blob/master/tests/page/page-wait-for-selector-1.spec.ts#L168-L169

anirudh-modi:
Hi @dgozman Thanks alot for the feedback!! 

1. Let me have a look at how can I meaningfully log the expression.. I guess we are already truncating when some dom element is being fetched. I will try to keep it consistent
2. I have already contacted my employer and I am awaiting clearance from my employer regarding the CLA, so it might take a day or 2 for me.
3. Thanks, for sharing the tests.. I did earlier look at the same file and tried doing the same implementation, but I guess I may have been wrongly implementing, I will give it another try and if I am stuck again I will reach to you guys.

mxschmitt:
Hi, also please revert the changes in the package-lock.json. They were made because you are using npm 6 and not 7. `npm i -g npm@7` should fix it for the future.

anirudh-modi:
Hi @dgozman I tried implementing the test cases the way you mentioned to me.. here is the sample code for the same..
```
it.only('should log message as evaluating', async ({ page,server }) => {
  const frame = page.mainFrame();
  const watchdog = frame.evaluate(() => {
    const a = 1;
    return a + 1;
  });
  await giveItTimeToLog(frame);
  const error = await watchdog.catch(e => e);
  console.log(error);
});
```

However, when I was logging, the error was only returning the actual evaluation result itself, and never returned the logs
Screenshot 2021-06-29 at 11 14 42 PM


I believe this is due to implementation which is being used for logging in `waitForSelector` [progress.log [this code](https://github.com/microsoft/playwright/blob/a270fc52069ca83bc58bf1afbca89191da64b644/src/server/frames.ts#L696-L720)] as compared to the `evaluate/$` where I am using the `debugLogger` module

anirudh-modi:
@mxschmitt package-lock reverted..

dgozman:
I'd do a test like this:

```js
it.only('should log message for page.$', async ({ page }) => {
  await page.setContent(`<div>not a span!</div>`);
  const error = await page.$('span').catch(e => e);
  expect(error.message).toContain('whatever log you are looking for');
});
```

anirudh-modi:
I tried the same.. but I am afraid in this case I am not getting the desired result, reason being in case of `$` and `ElementHandle` is generally being returned.

This could be due to be the implementation difference between `$` as compared to `waitForSelector`.

1. `$` is simply returning `this._page.selectors._query(this, selector);` which is the one I suppose is returning `ElementHandle`
2. whereas the `waitForSelector` is returning a new instance of `ProgressController`..

dgozman:
@anirudh-modi Good point! Sorry I missed that. Let's go without a test for now.

Let's also keep the log for evaluate for another PR. It is controversial, and I am not sure we want it in.

anirudh-modi:
@dgozman Sure let me then remove the evaluate from this PR.

anirudh-modi:
I have removed the evaluate log as mentioned..

I am still awaiting approval from my employer hopefully by tomorrow I should have some answer for the same..

anirudh-modi:
Hi @dgozman CLA has been signed and ready.. let me know if I have to do anything more for the PR to be merged..

dgozman (approved):

dgozman:
@anirudh-modi Looks great! Merging in.

anirudh-modi:
Thanks alot @dgozman!!!

PR #7345, opened 2021-06-28
fix(test-runner): handle negated toMatchSnapshot result

DZakh:
Fixed incorrect logic for `.not.toMatchSnapshot`:  

- Add detailed error instead of the "No message was specified for this matcher." from jest.
- Exit with the zero code if test runs with the "update-snapshots" flag
- Exit with the 1 code if missing an initial snapshot
- Doesn't create actual, diff, and expected snapshots as on fail artifacts, because there's no difference between them in this case.

### DIFF
--- a/src/test/golden.ts
+++ b/src/test/golden.ts
@@ -92,18 +92,24 @@ export function compare(
   const snapshotFile = snapshotPath(name);
 
   if (!fs.existsSync(snapshotFile)) {
-    const writingActual = updateSnapshots === 'all' || updateSnapshots === 'missing';
-    if (writingActual) {
+    const isWriteMissingMode = updateSnapshots === 'all' || updateSnapshots === 'missing';
+    const commonMissingSnapshotMessage = `${snapshotFile} is missing in snapshots`;
+    if (withNegateComparison) {
+      const message = `${commonMissingSnapshotMessage}${isWriteMissingMode ? ', matchers using ".not" won\'t write them automatically.' : '.'}`;
+      return { pass: true , message };
+    }
+    if (isWriteMissingMode) {
       fs.mkdirSync(path.dirname(snapshotFile), { recursive: true });
       fs.writeFileSync(snapshotFile, actual);
     }
-    const message = snapshotFile + ' is missing in snapshots' + (writingActual ? ', writing actual.' : '.');
+    const message = `${commonMissingSnapshotMessage}${isWriteMissingMode ? ', writing actual.' : '.'}`;
     if (updateSnapshots === 'all') {
       console.log(message);
-      return { pass: withNegateComparison ? false : true, message };
+      return { pass: true, message };
     }
-    return { pass: withNegateComparison ? true : false, message };
+    return { pass: false, message };
   }
+
   const expected = fs.readFileSync(snapshotFile);
   const extension = path.extname(snapshotFile).substring(1);
   const mimeType = extensionToMimeType[extension] || 'application/octet-string';
@@ -117,33 +123,28 @@ export function compare(
 
   const result = comparator(actual, expected, mimeType, options);
   if (!result) {
-    if (!withNegateComparison)
-      return { pass: true };
+    if (withNegateComparison) {
+      const message = [
+        colors.red('Snapshot comparison failed:'),
+        '',
+        indent('Expected result should be different from the actual one.', '  '),
+      ].join('\n');
+      return {
+        pass: true,
+        message,
+      };
+    }
 
-    const message = [
-      colors.red('Snapshot comparison failed:'),
-      '',
-      indent('Expected result should be different from the actual one.', '  '),
-    ].join('\n');
-    return {
-      pass: true,
-      message,
-    };
+    return { pass: true };
   }
 
-  if (updateSnapshots === 'all') {
+  if (updateSnapshots === 'all' && !withNegateComparison) {
     fs.mkdirSync(path.dirname(snapshotFile), { recursive: true });
     fs.writeFileSync(snapshotFile, actual);
     console.log(snapshotFile + ' does not match, writing actual.');
     return {
-      pass: withNegateComparison ? false : true,
+      pass: true,
       message: snapshotFile + ' running with --update-snapshots, writing actual.'
-    };
-  }
-
-  if (withNegateComparison) {
-    return {
-      pass: false,
     };
   }
 
@@ -155,6 +156,12 @@ export function compare(
   fs.writeFileSync(actualPath, actual);
   if (result.diff)
     fs.writeFileSync(diffPath, result.diff);
+
+  if (withNegateComparison) {
+    return {
+      pass: false,
+    };
+  }
 
   const output = [
     colors.red(`Snapshot comparison failed:`),
--- a/src/test/golden.ts
+++ b/src/test/golden.ts
@@ -138,7 +138,13 @@ export function compare(
     return { pass: true };
   }
 
-  if (updateSnapshots === 'all' && !withNegateComparison) {
+  if (withNegateComparison) {
+    return {
+      pass: false,
+    };
+  }
+
+  if (updateSnapshots === 'all') {
     fs.mkdirSync(path.dirname(snapshotFile), { recursive: true });
     fs.writeFileSync(snapshotFile, actual);
     console.log(snapshotFile + ' does not match, writing actual.');
@@ -156,12 +162,6 @@ export function compare(
   fs.writeFileSync(actualPath, actual);
   if (result.diff)
     fs.writeFileSync(diffPath, result.diff);
-
-  if (withNegateComparison) {
-    return {
-      pass: false,
-    };
-  }
 
   const output = [
     colors.red(`Snapshot comparison failed:`),
--- a/src/test/golden.ts
+++ b/src/test/golden.ts
@@ -61,7 +61,7 @@ function compareImages(actualBuffer: Buffer | string, expectedBuffer: Buffer, mi
       errorMessage: `Sizes differ; expected image ${expected.width}px X ${expected.height}px, but got ${actual.width}px X ${actual.height}px. `
     };
   }
-  const diff = new PNG({width: expected.width, height: expected.height});
+  const diff = new PNG({ width: expected.width, height: expected.height });
   const count = pixelmatch(expected.data, actual.data, diff.data, expected.width, expected.height, { threshold: 0.2, ...options });
   return count > 0 ? { diff: PNG.sync.write(diff) } : null;
 }
@@ -80,8 +80,17 @@ function compareText(actual: Buffer | string, expectedBuffer: Buffer): { diff?: 
   };
 }
 
-export function compare(actual: Buffer | string, name: string, snapshotPath: (name: string) => string, outputPath: (name: string) => string, updateSnapshots: UpdateSnapshots, options?: { threshold?: number }): { pass: boolean; message?: string; } {
+export function compare(
+  actual: Buffer | string,
+  name: string,
+  snapshotPath: (name: string) => string,
+  outputPath: (name: string) => string,
+  updateSnapshots: UpdateSnapshots,
+  withNegateComparison: boolean,
+  options?: { threshold?: number }
+): { pass: boolean; message?: string; } {
   const snapshotFile = snapshotPath(name);
+
   if (!fs.existsSync(snapshotFile)) {
     const writingActual = updateSnapshots === 'all' || updateSnapshots === 'missing';
     if (writingActual) {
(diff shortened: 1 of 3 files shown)


tests/playwright-test/golden.spec.ts
```diff
@@ -65,6 +65,70 @@ Line7`,
...
+      test('is a test', ({}) => {
+        expect('Hello world updated').toMatchSnapshot('snapshot.txt');
+      });
+    `
+  });
+  expect(result.exitCode).toBe(1);
+  expect(result.output).toContain('Snapshot comparison failed:');
+  expect(result.output).toMatch(/ {4}Expected: .+?snapshot-expected\.txt/);
+  expect(result.output).toMatch(/ {4}Received: .+?snapshot-actual\.txt/);
+  expect(result.output).not.toMatch(/ {8}Diff: .+?snapshot-diff\.txt/);
```
dgozman, line 82: Why no diff in this case?

tests/playwright-test/golden.spec.ts
```diff
@@ -80,6 +144,22 @@ test('should write missing expectations locally', async ({runInlineTest}, testIn
   expect(data.toString()).toBe('Hello world');
 });
 
+test('should write missing expectations locally with negated matcher', async ({runInlineTest}, testInfo) => {
+  const result = await runInlineTest({
+    'a.spec.js': `
+      const { test } = pwt;
+      test('is a test', ({}) => {
+        expect('Hello world').not.toMatchSnapshot('snapshot.txt');
```
dgozman, line 152: I don't think this should write the expectation file, because next time you run the test it will match and the test will fail.

tests/playwright-test/golden.spec.ts
```diff
@@ -96,6 +176,52 @@ test('should update expectations', async ({runInlineTest}, testInfo) => {
   expect(data.toString()).toBe('Hello world updated');
 });
 
+test('should update expectations with negated matcher', async ({runInlineTest}, testInfo) => {
+  const result = await runInlineTest({
+    'a.spec.js-snapshots/snapshot.txt': `Hello world`,
+    'a.spec.js': `
+      const { test } = pwt;
+      test('is a test', ({}) => {
+        expect('Hello world updated').not.toMatchSnapshot('snapshot.txt');
```
dgozman, line 185: Same here.

tests/playwright-test/golden.spec.ts
```diff
@@ -96,6 +176,52 @@ test('should update expectations', async ({runInlineTest}, testInfo) => {
...
+      });
+    `
+  }, { 'update-snapshots': true });
+  expect(result.exitCode).toBe(0);
+  expect(result.output).toContain('snapshot.txt does not match, writing actual.');
+  const data = fs.readFileSync(testInfo.outputPath('a.spec.js-snapshots/snapshot.txt'));
+  expect(data.toString()).toBe('Hello world updated');
+});
+
+test('should silently write missing expectations locally with the update-snapshots flag', async ({runInlineTest}, testInfo) => {
```
dgozman, line 195: Don't we already have a test for this?

tests/playwright-test/golden.spec.ts
```diff
@@ -96,6 +176,52 @@ test('should update expectations', async ({runInlineTest}, testInfo) => {
...
+  const data = fs.readFileSync(testInfo.outputPath('a.spec.js-snapshots/snapshot.txt'));
+  expect(data.toString()).toBe('Hello world');
+});
+
+test('should silently write missing expectations locally with the update-snapshots flag with negate matcher', async ({runInlineTest}, testInfo) => {
+  const result = await runInlineTest({
+    'a.spec.js': `
+      const { test } = pwt;
+      test('is a test', ({}) => {
+        expect('Hello world').not.toMatchSnapshot('snapshot.txt');
```
dgozman, line 215: Same here, I am not sure this is the right behavior.

dgozman (commented):
Thank you for the great PR!

I have a few comments about the desired behavior for updating snapshots. Generally, I think that it would be ideal for the test to pass after `--update-snapshots`. However, with `not.toMatchSnapshot` we cannot provide any meaningful expectation, so perhaps we should error in this case?

src/test/golden.ts
```diff
@@ -102,23 +111,42 @@ export function compare(actual: Buffer | string, name: string, snapshotPath: (na
...
+      '',
+      indent('Expected result should be different from the actual one.', '  '),
+    ].join('\n');
+    return {
+      pass: true,
+      message,
+    };
+  }
 
   if (updateSnapshots === 'all') {
```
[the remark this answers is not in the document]
DZakh: I agree with you. In my case I use `not.toMatchSnapshot` after a successfull `toMatchSnapshot` and it's going to be upsetting to have my snapshot overwritten. I haven't seen the whole picture when I wrote the code.  
But should we show an error or warning in this case?

tests/playwright-test/golden.spec.ts
DZakh: That's rational. Then the only way to use `not.toMatchSnapshot` will be by reusing a snapshot created by something else.

tests/playwright-test/golden.spec.ts
DZakh: Ok

tests/playwright-test/golden.spec.ts
DZakh: It's a different one. The test you're talking about assumed that a snapshot already exists and we need to update it. This one assumes that snapshot doesn't exist and we need to create a new one.

DZakh: Although I'm not really sure about the `locally` part where we pass an empty `CI` env variable. I copied it from another test and probably it shouldn't be here. If you have some knowledge about it, I'd be glad to hear it out.

tests/playwright-test/golden.spec.ts
DZakh: It's a new one, I've unswerd in the [thread](https://github.com/microsoft/playwright/pull/7345/files/fe01d1df92b0eabea96b8ce3baf38432f29231b5#r661722006). But I'm going to change the logic, it'll fail if snapshot is missing.

tests/playwright-test/golden.spec.ts
DZakh: There's just none. I've checked it out and it's because only `compareImages` returns the `diff` field, `compareText` and `compareBuffers` don't  have one.  
Should I create a separate test with the `test.fixme` annotation?

src/test/golden.ts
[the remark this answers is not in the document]
dgozman: I think we can throw if there is no expectation and we run in updateSnapshots mode.

DZakh (commented):
You're right, the update snapshots logic needs to be changed.

dgozman:
> You're right, the update snapshots logic needs to be changed.

Looking forward to the changes, let me know when I should take another look.

DZakh:
Also, I'd like to test artifacts in the test-result folder, but I've found the only way how to get the path without hardcoding, and it's not very beatiful. It'll look something like this: `result.projects[0].output`, what do you thing about it?

dgozman:
> Also, I'd like to test artifacts in the test-result folder, but I've found the only way how to get the path without hardcoding, and it's not very beatiful. It'll look something like this: `result.projects[0].output`, what do you thing about it?

Are you talking about test runner tests? Here is an example: https://github.com/microsoft/playwright/blob/master/tests/playwright-test/playwright.spec.ts#L170

src/test/golden.ts
```diff
@@ -128,6 +157,12 @@ export function compare(actual: Buffer | string, name: string, snapshotPath: (na
   if (result.diff)
     fs.writeFileSync(diffPath, result.diff);
 
+  if (withNegateComparison) {
```
dgozman, line 160: I think we should skip the block above (lines 151-158) that writes actual/expected/diff in this case, because everything worked out as expected (snapshot did not match), so we don't need extra artifacts.

tests/playwright-test/golden.spec.ts
dgozman: Oh, I see! I think that diff for text snapshot might be useful, so file an issue if you feel like it. No need for a fixme test just yet.

tests/playwright-test/golden.spec.ts
```diff
@@ -74,26 +122,101 @@ test('should write missing expectations locally', async ({runInlineTest}, testIn
...
+    'a.spec.js': `
+      const { test } = pwt;
+      test('is a test', ({}) => {
+        expect('${ACTUAL_SNAPSHOT}').not.toMatchSnapshot('snapshot.txt');
+      });
+    `
+  }, { 'update-snapshots': true });
+
+  expect(result.exitCode).toBe(0);
+  const snapshotOutputPath = testInfo.outputPath('a.spec.js-snapshots/snapshot.txt');
```
dgozman, line 183: I wouldn't expect these files here. Everything passed just fine!

dgozman (commented):
I think this is very close! Just a single issue with writing files when `not.toMatchSnapshot` actually does not match.

tests/playwright-test/golden.spec.ts
DZakh: I don't really understand what you mean here :) 
(Sorry my Engilsh isn't perfect)

src/test/golden.ts
DZakh: I've decided to return in because it may be usefull to understand what's going on. And if needed take an expected image from the `test-results` and update snapshot manually.  
But **I agree with you** because I personally find artifacts for passing tests quite annoying.  
Maybe we should keep them with another suffix?

tests/playwright-test/golden.spec.ts
DZakh: If  you want to say that `expect(data.toString()).toBe(EXPECTED_SNAPSHOT);` is unneeded, I'd argue with you because it checks that the expected snapshot wasn't updated.

DZakh (commented):
Have something to discuss

src/test/golden.ts
dgozman: I'd suggest to not write these files for now, and then gather feedback from the users first. 

Do they miss details in this case? From what I see we'll never print the path to the actual/expected/diff in this case, so 99% won't know where to look.

tests/playwright-test/golden.spec.ts
dgozman: My bad, I was reading it wrong. This test is great!

DZakh:
It should be the final version if there are no more comments.

dgozman (approved):
Looks great!

tests/playwright-test/golden.spec.ts
```diff
@@ -65,6 +65,77 @@ Line7`,
...
+        expect('Hello world updated').toMatchSnapshot('snapshot.txt');
+      });
+    `
+  });
+
+  expect(result.exitCode).toBe(1);
+  expect(result.output).toContain('Snapshot comparison failed:');
+  const expectedSnapshotArtifactPath = testInfo.outputPath('test-results', 'a-is-a-test', 'snapshot-expected.txt');
+  const actualSnapshotArtifactPath = testInfo.outputPath('test-results', 'a-is-a-test', 'snapshot-actual.txt');
+  expect(result.output).toMatch(new RegExp(` {4}Expected: .+?${expectedSnapshotArtifactPath}`));
```
dgozman, line 83: This one fails on Windows, probably because '\\' in RegExp has a special meaning. I'd suggest to use:
```
expect(result.output).toContain(`Expected: ${expectedSnapshotArtifactPath}`)
```

tests/playwright-test/golden.spec.ts
```diff
@@ -217,9 +363,18 @@ test('should compare different PNG images', async ({runInlineTest}) => {
...
   });
+
   expect(result.exitCode).toBe(1);
-  expect(result.output).toContain('Snapshot comparison failed');
-  expect(result.output).toContain('snapshot-diff.png');
+  expect(result.output).toContain('Snapshot comparison failed:');
+  const expectedSnapshotArtifactPath = testInfo.outputPath('test-results', 'a-is-a-test', 'snapshot-expected.png');
+  const actualSnapshotArtifactPath = testInfo.outputPath('test-results', 'a-is-a-test', 'snapshot-actual.png');
+  const diffSnapshotArtifactPath = testInfo.outputPath('test-results', 'a-is-a-test', 'snapshot-diff.png');
+  expect(result.output).toMatch(new RegExp(` {4}Expected: .+?${expectedSnapshotArtifactPath}`));
```
dgozman, line 372: The same issue here.

dgozman (commented):
I just noticed that Windows bot has failed, so I am not able to merge this in. Could you please update the tests?

tests/playwright-test/golden.spec.ts
DZakh: Because the path is colored it's wrapped in some invisible chars and toContain doesn't work the way you've written. I have 4 options what to do:  
- Test that path exists without describtion `expect(result.output).toContain(expectedSnapshotArtifactPath)`
- Escape path in RegExp using `.replace(/[#$()*+,.?[\\\]^{|}-]/g, '\\$&')`
- Import `colors` and wrap the path with `colors.yellow` in test
- Not to test  the output result  

What do you think?

dgozman: I think you can use `stripAscii` function we use in other places: https://github.com/microsoft/playwright/blob/742cce3a1d578dd736b3f949d22efa4794fc064e/tests/playwright-test/base-reporter.spec.ts#L126. If that does not work, just remove this `expect` call, it's not worth too much trouble.

dgozman (approved):

dgozman:

DZakh:
It was a pleasure to work with you, thanks for your guidance.

Issue #7348, opened 2021-06-28
[BUG] HAR: Body Content Race for Short-Lived Pages (Chromium-only) 


rwoll:
**Context:**
- Playwright Version: 1.12.1, 1.13.0-next-alpha-jun-28-2021
- Operating System: macOS, Linux
- Node.js version: v15.14.0
- Browser: Chromium

Add the following test to `har.spec.ts` and run the test 20 times via

```
npm run ctest -- --repeat-each 20
```

```ts
it.only('should include content for JavaScript redirects', async ({ contextFactory, server }, testInfo) => {
  server.setRoute('/js-redirect', (req, res) => {
    res.setHeader('Content-Type', 'text/html');
    res.end(`
    <!DOCTYPE html>
    <html>
      <head>
        <title>Redirect to Empty</title>
      </head>
      <body>
        redirect content
        <script>
          window.location.href = "${server.EMPTY_PAGE}"
        </script>
      </body>
    </html>
  `);
  });

  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
  await page.goto(server.PREFIX + '/js-redirect');

  const log = await getLog();

  const content = log.entries[0].response.content;
  expect(content.encoding).toBe('base64');
  expect(content.mimeType).toBe('text/html');
  expect(Buffer.from(content.text, 'base64').toString()).toContain('redirect content');

  expect(log.entries[1].request.url).toBe(server.EMPTY_PAGE);
});
```
**Describe the bug**

The response body of `/js-redirect` is not consistently captured in the HAR. In one set of 20 runs, 12 failed.

Given the following modification to harTracer.ts

```diff
diff --git a/src/server/supplements/har/harTracer.ts b/src/server/supplements/har/harTracer.ts
index b34c3fdb..295bb547 100644
--- a/src/server/supplements/har/harTracer.ts
+++ b/src/server/supplements/har/harTracer.ts
@@ -225,7 +225,7 @@ export class HarTracer {
       const promise = response.body().then(buffer => {
         harEntry.response.content.text = buffer.toString('base64');
         harEntry.response.content.encoding = 'base64';
-      }).catch(() => {});
+      }).catch((err) => { console.error(response.url(), err); });
       this._addBarrier(page, promise);
     }
   }
```

whenever the test fails, we get the following error when trying to get the body:

```
http://localhost:8907/js-redirect Protocol error (Network.getResponseBody): No resource with given identifier foundError
[...]

PR #7349, opened 2021-06-28
test(chromium): HAR missing content for JS redirect

rwoll:
The two new tests are identical except for the `; charset=UTF-8` bit.
You'll notice that the tests pass, except for Chromium when the charset
is specified.

NB: If you reverse the order of these tests (assuming a serial run),
all the tests pass for both browsers.

Relates https://github.com/microsoft/playwright/issues/7348

### DIFF
--- a/tests/har.spec.ts
+++ b/tests/har.spec.ts
@@ -253,6 +253,71 @@ it('should include content', async ({ contextFactory, server }, testInfo) => {
   expect(Buffer.from(content2.text, 'base64').toString()).toContain('pink');
 });
 
+it('should include content for JavaScript redirects with charset', async ({ contextFactory, server, browserName }, testInfo) => {
+  it.fail(browserName === 'chromium');
+  server.setRoute('/js-redirect', (req, res) => {
+    res.setHeader('Content-Type', 'text/html; charset=UTF-8');
+    res.end(`
+    <!DOCTYPE html>
+    <html>
+      <head>
+        <title>Redirect to Empty</title>
+      </head>
+      <body>
+        redirect content
+        <script>
+          window.location.href = "${server.EMPTY_PAGE}"
+        </script>
+      </body>
+    </html>
+  `);
+  });
+
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  await page.goto(server.PREFIX + '/js-redirect');
+
+  const log = await getLog();
+
+  const content = log.entries[0].response.content;
+  expect(content.encoding).toBe('base64');
+  expect(content.mimeType).toBe('text/html; charset=UTF-8');
+  expect(Buffer.from(content.text, 'base64').toString()).toContain('redirect content');
+
+  expect(log.entries[1].request.url).toBe(server.EMPTY_PAGE);
+});
+
+it('should include content for JavaScript redirects', async ({ contextFactory, server }, testInfo) => {
+  server.setRoute('/js-redirect', (req, res) => {
+    res.setHeader('Content-Type', 'text/html');
+    res.end(`
+    <!DOCTYPE html>
+    <html>
+      <head>
+        <title>Redirect to Empty</title>
+      </head>
+      <body>
+        redirect content
+        <script>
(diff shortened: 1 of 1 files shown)


rwoll:
It looks like these passed on the bots. However, if you run locally via `npm run test` they consistently fail (at least on macOS).

rwoll:
@dgozman It looks like it's still passing on the bots, but here's what I get when I run locally after adding `.only` to the two tests and using `npm run test`:

```
Running 6 tests using 3 workers
  1) har.spec.ts:289:4 › [chromium] should include content for JavaScript redirects ================

    Expected: "base64"
    Received: undefined

      312 |
      313 |   const content = log.entries[0].response.content;
    > 314 |   expect(content.encoding).toBe('base64');
          |                            ^
      315 |   expect(content.mimeType).toBe('text/html');
      316 |   expect(Buffer.from(content.text, 'base64').toString()).toContain('redirect content');
      317 |

        at …/playwright/tests/har.spec.ts:314:28
        at WorkerRunner._runTestWithBeforeHooks (/Users/rwoll/code/playwright/tests/config/test-runner/node_modules/@playwright/test/lib/test/workerRunner.js:333:7)


  1 failed
    har.spec.ts:289:4 › [chromium] should include content for JavaScript redirects =================
  5 passed (4s)
```

Do you get the same failure?

rwoll:
It looks like both tests are flaky to varying degrees (and I'm either doing something wrong or there is a race in the PW or Chromium code). If you add `await page.waitForLoadState();` immediately after the goto, both tests consistently pass. Without this, we get to the end destination, so I don't think it should be necessary to have this wait in order to see the intermediate content in the HAR. Perhaps an event is coming too "late"?

rwoll:
(looking through the harTracer code now to see if there's some assumptions of event ordering and/or timing we aren't meeting)

rwoll:
> It looks like both tests are flaky to varying degrees (and I'm either doing something wrong or there is a race in the PW or Chromium code). If you add `await page.waitForLoadState();` immediately after the goto, both tests consistently pass. Without this, we still get to the end destination, so I don't think it should be necessary to have this wait in order to see the intermediate content in the HAR. Perhaps an event is coming too "late"?

I take that back partially—it looks like waitForLoadState doesn't totally help in the charset case.

rwoll:
@dgozman given the following modification to harTracer.ts

```diff
diff --git a/src/server/supplements/har/harTracer.ts b/src/server/supplements/har/harTracer.ts
index b34c3fdb..295bb547 100644
--- a/src/server/supplements/har/harTracer.ts
+++ b/src/server/supplements/har/harTracer.ts
@@ -225,7 +225,7 @@ export class HarTracer {
       const promise = response.body().then(buffer => {
         harEntry.response.content.text = buffer.toString('base64');
         harEntry.response.content.encoding = 'base64';
-      }).catch(() => {});
+      }).catch((err) => { console.error(response.url(), err); });
       this._addBarrier(page, promise);
     }
   }
```

whenever the test fails, we get the following error when trying to get the body:

```
http://localhost:8907/js-redirect Protocol error (Network.getResponseBody): No resource with given identifier foundError
    at /Users/rwoll/code/playwright/src/server/chromium/crConnection.ts:174:56
    at new Promise (<anonymous>)
    at CRSession.send (/Users/rwoll/code/playwright/src/server/chromium/crConnection.ts:173:12)
    at Response.getResponseBody [as _getResponseBodyCallback] (/Users/rwoll/code/playwright/src/server/chromium/crNetworkManager.ts:284:43)
    at /Users/rwoll/code/playwright/src/server/network.ts:392:21
```

(It looks like doing 10 repeats of charset test locally is enough to find a mix of passes and failures.)

pavelfeldman (approved):

pavelfeldman:
I approved before reading the thread and seeing the bots. Might be easier to fix and test rather that trigger the error?

rwoll:
Agreed—will close this for now. There's an issue, but it's turning out to be flaky so not sure we can have a great test for it on the bots.

Issue #7262, opened 2021-06-22
[BUG] BeforeEach/AfterEach hooks run ignoring the skip, fixme, fail annotation 


yana-kirichenko:
**Context:**
- Playwright Version: 1.12.2
- Operating System: Windows
- Node.js version: 10.0

**Code Snippet**

BeforeEach/AfterEach hooks run regardless of the use of any annotation the skip, fixme, fail. It brings 2 problems:

1. The test might be failed in hook block due to some condition, e.g. Browser type.
2. BeforeEach/AfterEach hooks waste build time as an ignored test will not be run anyway.

The same issue happens if we put an annotation on describe level (pls, see the second picture in the attachment).

```typescript
test.describe('Describe',
    () => {
        test.beforeEach(async () => {
            console.log('this is before each hook');
        });

        test('Test itself', async () => {
            test.fixme();
            console.log('this is test');
        });

        test.afterEach(async () => {
            console.log('this is after each hook');
        });
    }
);
```

**Evidence that BeforeEach/AfterEach hooks were run**

**Example on test level**
![image](https://user-images.githubusercontent.com/41748631/122956235-7dd4a400-d389-11eb-97ae-7408751a5fe2.png)

**Example on describe level**
![image](https://user-images.githubusercontent.com/41748631/122956041-567dd700-d389-11eb-8018-1ae9734b4c9d.png)

aslushnikov:
in other words, this is a feature-request for a static-time test annotations.

> The test might be failed in hook block due to some condition, e.g. Browser type.

@yana-kirichenko This is very valid! Out of curiosity, what kind of stuff do you run in the beforeeach/aftereach hooks that fail?

yana-kirichenko:
@aslushnikov
Thanks!
These are UI actions on the page that are preconditions and postconditions for some tests. And we need to fully ignore these tests including these hooks for a particular browser.

dgozman:
@yana-kirichenko You can skip from inside the hook - this should work in this particular case.

```js
test.beforeEach(async ({ page, browserName }, testInfo) => {
  testInfo.skip(browserName === 'chromium');
  // Some setup using page...
});
```

Does this work for you?

dgozman:
With #7436, you can use a separate `test.skip()` statement in the block that will avoid running hooks if possible:

```js
// Unconditional skip
test.skip();

// Alternatively, skip based on a browser name
[...]

PR #7361, opened 2021-06-28
fix(test runner): skip without a callback inside describe should prevent hooks

dgozman:
References #7262.


pavelfeldman:
I don't think this helps - reported issue implies there is a condition around browser type.

PR #7365, opened 2021-06-28
feat(install): docs and help

dgozman:
We recommend to use `npx playwright install` on the intro page, but never document it.

Drive-by: remove test-intro and references to it.


docs/src/release-notes-js.md
```diff
@@ -9,7 +9,7 @@ title: "Release notes"
 
 #### ⚡️ Introducing Playwright TestRunner
 
-[Playwright TestRunner](./test-intro.md) is a **new test runner** built from scratch by Playwright team specifically to accommodate end-to-end testing needs:
+[Playwright TestRunner](./intro.md) is a **new test runner** built from scratch by Playwright team specifically to accommodate end-to-end testing needs:
```
mxschmitt, line 12: not related to this PR: sometimes we use test-runner, sometimes TestRunnner and mostly Playwright Test afaik. Can we stick with Playwright or clarify for a standardised variant?

pavelfeldman (approved):

mxschmitt (commented):
We should be careful with removing the test-intro and add a redirect on playwright.dev side because we (or at least I) referenced to it in a lot of places, GH issues etc.

PR #7368, opened 2021-06-28
feat(test-runner): add webServer

### LINKED ISSUE #7398: [Feature] Add web server support
Hi Guys,
My current test solution uses jest as test runner. I would like to migrate to the @playwright/test. But I don`t find the following functionality in the doc. Jest-playwright allows to launch the app before test running.
module.exports = {
    serverOptions: {
        command: 'yarn start',
        port: 3000,
        launchTimeout: 300000,
      },
  }
Is there same functionality in @playwright/test?

mxschmitt:
Tested with React, Vue.js and Angular which works. The big difference to jest-dev-server is that we don't wait for 2xx, we only wait for accepted TCP connections.

Supersedes #6959
Fixes #7398

### DIFF
--- a/docs/src/test-advanced.md
+++ b/docs/src/test-advanced.md
@@ -42,6 +42,7 @@ These options would be typically different between local development and CI oper
 - `reportSlowTests: { max: number, threshold: number } | null` - Whether to report slow tests. When `null`, slow tests are not reported. Otherwise, tests that took more than `threshold` milliseconds are reported as slow, but no more than `max` number of them. Passing zero as `max` reports all slow tests that exceed the threshold.
 - `shard: { total: number, current: number } | null` - [Shard](./test-parallel.md#shards) information.
 - `updateSnapshots: boolean` - Whether to update expected snapshots with the actual results produced by the test run.
+- `webServer: { command: string, port?: number, cwd?: string, timeout?: number, env?: object }` - Launch a web server before the tests will start. It will automaticially detect the port when it got printed to the stdout which works for React.js, Vue.js, Angular etc.
 - `workers: number` - The maximum number of concurrent worker processes to use for parallelizing tests.
 
 Note that each [test project](#projects) can provide its own test suite options, for example two projects can run different tests by providing different `testDir`s. However, test run options are shared between all projects.
@@ -199,6 +200,57 @@ export const test = base.extend<{ saveLogs: void }>({
   }, { auto: true } ]
 });
 ```
+
+## Launching a development web server during the tests
+
+To launch a web server during the tests, use the `webServer` option in the [configuration file](#configuration-object).
+
+This will automatically launch and wait for the web server to be ready before the tests will start:
+
+```js js-flavor=ts
+// playwright.config.ts
+import { PlaywrightTestConfig } from '@playwright/test';
+
+const config: PlaywrightTestConfig = {
+  webServer: {
+    command: 'npm run start',
+    timeout: 120 * 1000,
+  },
+  projects: [
+    {
+      name: 'Chrome',
+      use: {
+        browserName: 'chromium',
+      },
+    },
+  ],
+};
+
+export default config;
+```
+
+```js js-flavor=js
+// @ts-check
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+const config = {
+  webServer: {
+    command: 'npm run start',
+    timeout: 120 * 1000,
+  },
+  projects: [
+    {
+      name: 'Chrome',
+      use: {
+        browserName: 'chromium',
+      },
+    },
+  ],
+};
+
+mode.exports = config;
+```
+
+If your dev web-server does not print the listening URL to the console, you can also manually specify a `port` or additional environment variables, see [here](#configuration-object).
 
 ## Global setup and teardown
 
--- a/docs/src/test-advanced.md
+++ b/docs/src/test-advanced.md
@@ -207,7 +207,7 @@ To launch a web server during the tests, use the `webServer` option in the [conf
 
 Playwright Test does automatically detect if a localhost URL like `http://localhost:3000` gets printed to the stdout.
 The port from the printed URL gets then used to check when its accepting requests and passed over to Playwright as a
-`baseURL`. You can also manually specify a `port` or additional environment variables, see [here](#configuration-object).
+[`param: baseURL`] when creating the context [`method: Browser.newContext`]. You can also manually specify a `port` or additional environment variables, see [here](#configuration-object).
 
 ```js js-flavor=ts
 // playwright.config.ts
@@ -223,6 +223,16 @@ const config: PlaywrightTestConfig = {
 export default config;
 ```
 
+```js js-flavor=ts
+// test.spec.ts
+import { test } = from '@playwright/test';
+
+test('test', async ({ page }) => {
+  // This will result in e.g. http://localhost:3000/foo when your dev-server prints a http://localhost:3000 address
+  await page.goto('/foo');
+});
+```
+
 ```js js-flavor=js
 // @ts-check
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
@@ -234,6 +244,16 @@ const config = {
 };
 
 mode.exports = config;
+```
+
+```js js-flavor=js
+// test.spec.js
+const { test } = require('@playwright/test');
+
+test('test', async ({ page }) => {
+  // This will result in e.g. http://localhost:3000/foo when your dev-server prints a http://localhost:3000 address
+  await page.goto('/foo');
+});
 ```
 
 ## Global setup and teardown
--- a/docs/src/test-advanced.md
+++ b/docs/src/test-advanced.md
@@ -42,7 +42,7 @@ These options would be typically different between local development and CI oper
 - `reportSlowTests: { max: number, threshold: number } | null` - Whether to report slow tests. When `null`, slow tests are not reported. Otherwise, tests that took more than `threshold` milliseconds are reported as slow, but no more than `max` number of them. Passing zero as `max` reports all slow tests that exceed the threshold.
 - `shard: { total: number, current: number } | null` - [Shard](./test-parallel.md#shards) information.
 - `updateSnapshots: boolean` - Whether to update expected snapshots with the actual results produced by the test run.
-- `webServer: { command: string, port?: number, cwd?: string, timeout?: number, env?: object }` - Launch a web server before the tests will start. It will automaticially detect the port when it got printed to the stdout which works for React.js, Vue.js, Angular etc.
+- `webServer: { command: string, port?: number, cwd?: string, timeout?: number, env?: object }` - Launch a web server before the tests will start. It will automaticially detect the port when it got printed to the stdout.
 - `workers: number` - The maximum number of concurrent worker processes to use for parallelizing tests.
 
 Note that each [test project](#projects) can provide its own test suite options, for example two projects can run different tests by providing different `testDir`s. However, test run options are shared between all projects.
@@ -205,7 +205,9 @@ export const test = base.extend<{ saveLogs: void }>({
 
 To launch a web server during the tests, use the `webServer` option in the [configuration file](#configuration-object).
 
-This will automatically launch and wait for the web server to be ready before the tests will start:
+Playwright Test does automatically detect if a localhost URL like `http://localhost:3000` gets printed to the stdout.
+The port from the printed URL gets then used to check when its accepting requests and passed over to Playwright as a
+`baseURL`. You can also manually specify a `port` or additional environment variables, see [here](#configuration-object).
 
 ```js js-flavor=ts
 // playwright.config.ts
@@ -216,14 +218,6 @@ const config: PlaywrightTestConfig = {
     command: 'npm run start',
     timeout: 120 * 1000,
   },
-  projects: [
-    {
-      name: 'Chrome',
-      use: {
-        browserName: 'chromium',
-      },
-    },
-  ],
 };
 
 export default config;
@@ -237,20 +231,10 @@ const config = {
     command: 'npm run start',
     timeout: 120 * 1000,
   },
-  projects: [
-    {
-      name: 'Chrome',
-      use: {
-        browserName: 'chromium',
-      },
-    },
-  ],
 };
 
 mode.exports = config;
 ```
-
-If your dev web-server does not print the listening URL to the console, you can also manually specify a `port` or additional environment variables, see [here](#configuration-object).
 
 ## Global setup and teardown
 
--- a/docs/src/test-configuration.md
+++ b/docs/src/test-configuration.md
@@ -161,6 +161,7 @@ In addition to configuring [Browser] or [BrowserContext], videos or screenshots,
 - `testIgnore`: Glob patterns or regular expressions that should be ignored when looking for the test files. For example, `'**/test-assets'`.
 - `testMatch`: Glob patterns or regular expressions that match test files. For example, `'**/todo-tests/*.spec.ts'`. By default, Playwright Test runs `.*(test|spec)\.(js|ts|mjs)` files.
 - `timeout`: Time in milliseconds given to each test.
+- `webServer: { command: string, port?: number, cwd?: string, timeout?: number, env?: object }` - Launch a web server before the tests will start. It will automaticially detect the port when it got printed to the stdout which works for React.js, Vue.js, Angular etc.
 - `workers`: The maximum number of concurrent worker processes to use for parallelizing tests.
 
 You can specify these options in the configuration file.
--- a/docs/src/test-configuration.md
+++ b/docs/src/test-configuration.md
@@ -161,7 +161,7 @@ In addition to configuring [Browser] or [BrowserContext], videos or screenshots,
 - `testIgnore`: Glob patterns or regular expressions that should be ignored when looking for the test files. For example, `'**/test-assets'`.
 - `testMatch`: Glob patterns or regular expressions that match test files. For example, `'**/todo-tests/*.spec.ts'`. By default, Playwright Test runs `.*(test|spec)\.(js|ts|mjs)` files.
 - `timeout`: Time in milliseconds given to each test.
-- `webServer: { command: string, port?: number, cwd?: string, timeout?: number, env?: object }` - Launch a web server before the tests will start. It will automaticially detect the port when it got printed to the stdout which works for React.js, Vue.js, Angular etc.
+- `webServer: { command: string, port?: number, cwd?: string, timeout?: number, env?: object }` - Launch a web server before the tests will start. It will automaticially detect the port when it got printed to the stdout.
 - `workers`: The maximum number of concurrent worker processes to use for parallelizing tests.
 
 You can specify these options in the configuration file.
--- a/src/server/processLauncher.ts
+++ b/src/server/processLauncher.ts
@@ -19,7 +19,7 @@ import * as childProcess from 'child_process';
 import * as readline from 'readline';
 import { helper } from './helper';
 import * as types from './types';
-import { isUnderTest, removeFolders } from '../utils/utils';
+import { isUnderTest, killProcessGroup, removeFolders } from '../utils/utils';
 
 export type Env = {[key: string]: string | number | boolean | undefined};
 
--- a/src/server/processLauncher.ts
+++ b/src/server/processLauncher.ts
@@ -19,14 +19,15 @@ import * as childProcess from 'child_process';
 import * as readline from 'readline';
 import { helper } from './helper';
 import * as types from './types';
-import { isUnderTest, killProcessGroup, removeFolders } from '../utils/utils';
+import { isUnderTest, removeFolders } from '../utils/utils';
 
 export type Env = {[key: string]: string | number | boolean | undefined};
 
 export type LaunchProcessOptions = {
-  executablePath: string,
-  args: string[],
+  command: string,
+  args?: string[],
   env?: Env,
+  shell?: boolean,
 
   handleSIGINT?: boolean,
   handleSIGTERM?: boolean,
@@ -62,20 +63,18 @@ if (maxListeners !== 0)
 
 export async function launchProcess(options: LaunchProcessOptions): Promise<LaunchResult> {
   const stdio: ('ignore' | 'pipe')[] = options.stdio === 'pipe' ? ['ignore', 'pipe', 'pipe', 'pipe', 'pipe'] : ['pipe', 'pipe', 'pipe'];
-  options.log(`<launching> ${options.executablePath} ${options.args.join(' ')}`);
-  const spawnedProcess = childProcess.spawn(
-      options.executablePath,
-      options.args,
-      {
-        // On non-windows platforms, `detached: true` makes child process a leader of a new
-        // process group, making it possible to kill child process tree with `.kill(-pid)` command.
-        // @see https://nodejs.org/api/child_process.html#child_process_options_detached
-        detached: process.platform !== 'win32',
-        env: (options.env as {[key: string]: string}),
-        cwd: options.cwd,
-        stdio,
-      }
-  );
+  options.log(`<launching> ${options.command} ${options.args ? options.args.join(' ') : ''}`);
+  const spawnOptions: childProcess.SpawnOptions = {
+    // On non-windows platforms, `detached: true` makes child process a leader of a new
+    // process group, making it possible to kill child process tree with `.kill(-pid)` command.
+    // @see https://nodejs.org/api/child_process.html#child_process_options_detached
+    detached: process.platform !== 'win32',
+    env: (options.env as {[key: string]: string}),
+    cwd: options.cwd,
+    shell: options.shell,
+    stdio,
+  };
+  const spawnedProcess = childProcess.spawn(options.command, options.args, spawnOptions);
 
   const cleanup = async () => {
     options.log(`[pid=${spawnedProcess.pid || 'N/A'}] starting temporary directories cleanup`);
--- a/src/test/index.ts
+++ b/src/test/index.ts
@@ -74,9 +74,12 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
   timezoneId: undefined,
   userAgent: undefined,
   viewport: undefined,
+  baseURL: async ({ }, use) => {
+    await use(process.env.PLAYWRIGHT_TEST_BASE_URL);
+  },
   contextOptions: {},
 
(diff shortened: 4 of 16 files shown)


src/test/webServer.ts
```diff
@@ -0,0 +1,136 @@
...
+    ])) as { timedOut: number };
+    cancellationToken.canceled = true;
+    if (timedOut) {
+      this.kill();
+      throw new Error(`failed to start web server on port ${this._port} via "${this._config.command}"`);
+    }
+  }
+  public async kill() {
+    assert(this._process);
+    await new Promise(resolve => kill(this._process!.pid, 'SIGINT', resolve));
```
dgozman, line 115: Why SIGINT?

src/test/runner.ts
```diff
@@ -166,9 +167,17 @@ export class Runner {
       testFiles.forEach(file => allTestFiles.add(file));
     }
 
-    let globalSetupResult: any;
+    let globalSetupResultTeardown: any;
```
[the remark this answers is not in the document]
mxschmitt: I already reverted it, seems like your review version was outdated. hmm.

tests/playwright-test/assets/simple-server-with-stdout.js
```diff
@@ -0,0 +1,10 @@
+const {TestServer} = require('../../../utils/testserver/');
+// delay creating the server to test waiting for it
+setTimeout(() => {
+    TestServer.create(__dirname, process.argv[2] || 3000).then(server => {
```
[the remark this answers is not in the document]
mxschmitt: (was 4 before, changed to 2.)

src/test/index.ts
```diff
@@ -52,6 +52,12 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
     await removeFolders([artifactsFolder]);
   }, { scope: 'worker' } ],
 
+  baseURL: [async ({ }, use) => {
```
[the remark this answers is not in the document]
mxschmitt: I agree there, linking here [this](https://pypi.org/project/pytest-base-url/) to it.
btw. why webServerUrl instead and not webServerURL?

src/test/index.ts
```diff
@@ -52,6 +52,12 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
...
+    if (!process.env.PLAYWRIGHT_BASE_URL)
+      throw new Error('A web server must be specified to use baseURL. See `webServer` in the config options, or set the PLAYWRIGHT_BASE_URL environment variable.');
```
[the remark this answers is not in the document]
mxschmitt: Depends which story we want to go. Customers keep asking of "how to pass a base URL externally into the test-runner". But fine with removing it for now.

src/test/webServer.ts
```diff
@@ -0,0 +1,136 @@
...
+    this.push(chunk.toString().split(os.EOL).map((line: string): string => line ? `[Playwright WebServer] ${line}` : line).join(os.EOL));
+    callback();
+  },
+});
+
+export class WebServer {
+  _process?: ChildProcess;
+  _port?: number;
+  private _processExitedWithNonZeroStatusCode!: Promise<unknown>;
+  private _config: Omit<WebServerConfig, 'port'>;
```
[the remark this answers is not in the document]
mxschmitt: We internally move port out of it and the Omit ensures that we not use it by accident.

src/test/webServer.ts
mxschmitt: SIGTERM should be better. SIGINT came out of the frustration that child_process.ChildProcess.terminate() didn't send the signal to the sub-processes.

src/test/webServer.ts
```diff
@@ -0,0 +1,136 @@
...
+  transform(this: stream.Transform, chunk: any, encoding: string, callback: stream.TransformCallback) {
+    this.push(chunk.toString().split(os.EOL).map((line: string): string => line ? `[Playwright WebServer] ${line}` : line).join(os.EOL));
+    callback();
+  },
+});
+
+export class WebServer {
+  _process?: ChildProcess;
+  _port?: number;
+  private _processExitedWithNonZeroStatusCode!: Promise<unknown>;
```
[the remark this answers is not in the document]
mxschmitt: I changed it to Promise<any>, otherwise we would need to cast the type in the Promise.race when we use it to get rid of TS errors.

src/test/runner.ts
```diff
@@ -169,6 +170,14 @@ export class Runner {
     let globalSetupResult: any;
     if (config.globalSetup)
       globalSetupResult = await this._loader.loadGlobalHook(config.globalSetup, 'globalSetup')(this._loader.fullConfig());
+    let webServer: WebServer|null = null;
+    try {
+      if (config.webServer)
+        webServer = await WebServer.create(config.webServer);
+    } catch (error) {
+      console.log(error instanceof Error ? error.message : error);
```
dgozman, line 178: Let's throw instead? We throw in many places here and get default reporting of the error.

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+import { WebServerConfig } from '../../types/test';
+import { assert } from '../utils/utils';
+
+const DEFAULT_ENVIRONMENT_VARIABLES = {
+  'BROWSER': 'none', // Disable that create-react-app will open the page in the browser
+};
+
+const newProcessLogPrefixer = () => new stream.Transform({
+  transform(this: stream.Transform, chunk: any, encoding: string, callback: stream.TransformCallback) {
+    this.push(chunk.toString().split(os.EOL).map((line: string): string => line ? `[Playwright WebServer] ${line}` : line).join(os.EOL));
```
dgozman, line 32: I'd drop "Playwright" because it's not our own web server :)

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+import { monotonicTime, raceAgainstDeadline } from './util';
+import { WebServerConfig } from '../../types/test';
+import { assert } from '../utils/utils';
+
+const DEFAULT_ENVIRONMENT_VARIABLES = {
+  'BROWSER': 'none', // Disable that create-react-app will open the page in the browser
+};
+
+const newProcessLogPrefixer = () => new stream.Transform({
+  transform(this: stream.Transform, chunk: any, encoding: string, callback: stream.TransformCallback) {
```
dgozman, line 31: Do you have to use `encoding` when doing `toString`?

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+    this._process.stdout.pipe(newProcessLogPrefixer()).pipe(process.stdout);
+    this._process.stderr.pipe(newProcessLogPrefixer()).pipe(process.stderr);
+    if (!this._port)
+      this._process.stdout.on('data', collectPort);
+    let processExitedWithNonZeroStatusCodeCallback = (error: Error) => { };
+    this._processExitedWithNonZeroStatusCode = new Promise((_, reject) => processExitedWithNonZeroStatusCodeCallback = reject);
+    this._process.on('exit', code => processExitedWithNonZeroStatusCodeCallback(new Error(`WebServer was not able to start. Exit code: ${code}`)));
+    if (!this._port) {
+      this._port = await collectPortPromise;
+      console.log(`[Playwright Test] Port ${this._port} by process '${this._config.command}' was automatically detected. Starting the tests...`);
```
dgozman, line 99: Why the prefix? We never prefix anything else in our messages.

I'd also drop "starting the tests..." because you never know what's next in the pipeline :) Instead, you can print "waiting for the port to respond" if it takes more than 500(?) ms.

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+  private async _verifyFreePort() {
+    assert(this._port);
+    const cancellationToken = { canceled: false };
+    const portIsUsed = await Promise.race([
+      new Promise(resolve => setTimeout(() => resolve(false), 100)),
+      waitForSocket(this._port, 100, cancellationToken),
+    ]);
+    cancellationToken.canceled = true;
+    if (portIsUsed)
+      throw new Error(`Port ${this._port} is used, make sure that nothing is running on the port`);
```
dgozman, line 71: Is this just a nicety? I think starting the server will just fail if we do not verify the port. It is also racy...

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+    ]));
+    cancellationToken.canceled = true;
+    if (timedOut) {
+      this.kill();
+      throw new Error(`failed to start web server on port ${this._port} via "${this._config.command}"`);
+    }
+  }
+  public async kill() {
+    assert(this._process);
+    await new Promise(resolve => kill(this._process!.pid, 'SIGTERM', resolve));
```
dgozman, line 119: Btw, the process might have exited at this point, due to any reason. Will this throw? I think `kill` would better not throw.

src/test/webServer.ts
mxschmitt: > If the chunk is a string, then this is the encoding type. If chunk is a buffer, then this is the special value 'buffer'. Ignore it in that case.

So I think no, since its guaranteed to be a Buffer and not a string.

src/test/webServer.ts
mxschmitt: Since most of the users are using web frameworks it mostly takes long, thats why I'm printing now the "detecting the port" message always.

src/test/webServer.ts
mxschmitt: It was for the case when the user specifies a command and manually a port. We would need to ensure then that the process is running while the port is accepting connections which was otherwise more racy than that in my mind.

docs/src/test-advanced.md
```diff
@@ -42,6 +42,7 @@ These options would be typically different between local development and CI oper
 - `reportSlowTests: { max: number, threshold: number } | null` - Whether to report slow tests. When `null`, slow tests are not reported. Otherwise, tests that took more than `threshold` milliseconds are reported as slow, but no more than `max` number of them. Passing zero as `max` reports all slow tests that exceed the threshold.
 - `shard: { total: number, current: number } | null` - [Shard](./test-parallel.md#shards) information.
 - `updateSnapshots: boolean` - Whether to update expected snapshots with the actual results produced by the test run.
+- `webServer: { command: string, port?: number, timeout?: number, env?: object }` - Launch a web server before the tests will start.
```
JoelEinbinder, line 45:
```suggestion
- `webServer: { command: string, port?: number, timeout?: number, env?: object }` - Launch a web server before the tests start.
```

JoelEinbinder: Something about the regex/baseurl here?

docs/src/test-advanced.md
```diff
@@ -200,6 +201,57 @@ export const test = base.extend<{ saveLogs: void }>({
 });
 ```
 
+## Launching a development web server during the tests
```
dgozman, line 204: I think this warrants a separate doc. Maybe combined with "how to build and run tests", or "how to integrate with my dev workflow". Perhaps in a follow up PR.

docs/src/test-advanced.md
```diff
@@ -200,6 +201,57 @@ export const test = base.extend<{ saveLogs: void }>({
...
+```js js-flavor=ts
+// playwright.config.ts
+import { PlaywrightTestConfig } from '@playwright/test';
+
+const config: PlaywrightTestConfig = {
+  webServer: {
+    command: 'npm run start',
+    timeout: 120 * 1000,
+  },
+  projects: [
```
dgozman, line 219: You can just omit projects here, it will still work.

docs/src/test-advanced.md
```diff
@@ -200,6 +201,57 @@ export const test = base.extend<{ saveLogs: void }>({
...
+        browserName: 'chromium',
+      },
+    },
+  ],
+};
+
+mode.exports = config;
+```
+
+If your dev web-server does not print the listening URL to the console, you can also manually specify a `port` or additional environment variables, see [here](#configuration-object).
```
dgozman, line 253: Let's be more specific about "listening URL to the console" - give an example that we'll detect.

docs/src/test-advanced.md
```diff
@@ -42,6 +42,7 @@ These options would be typically different between local development and CI oper
 - `reportSlowTests: { max: number, threshold: number } | null` - Whether to report slow tests. When `null`, slow tests are not reported. Otherwise, tests that took more than `threshold` milliseconds are reported as slow, but no more than `max` number of them. Passing zero as `max` reports all slow tests that exceed the threshold.
 - `shard: { total: number, current: number } | null` - [Shard](./test-parallel.md#shards) information.
 - `updateSnapshots: boolean` - Whether to update expected snapshots with the actual results produced by the test run.
+- `webServer: { command: string, port?: number, cwd?: string, timeout?: number, env?: object }` - Launch a web server before the tests will start. It will automaticially detect the port when it got printed to the stdout which works for React.js, Vue.js, Angular etc.
```
dgozman, line 45: I'd drop the "which works..." part, seems out of place.

docs/src/test-configuration.md
```diff
@@ -161,6 +161,7 @@ In addition to configuring [Browser] or [BrowserContext], videos or screenshots,
 - `testIgnore`: Glob patterns or regular expressions that should be ignored when looking for the test files. For example, `'**/test-assets'`.
 - `testMatch`: Glob patterns or regular expressions that match test files. For example, `'**/todo-tests/*.spec.ts'`. By default, Playwright Test runs `.*(test|spec)\.(js|ts|mjs)` files.
 - `timeout`: Time in milliseconds given to each test.
+- `webServer: { command: string, port?: number, cwd?: string, timeout?: number, env?: object }` - Launch a web server before the tests will start. It will automaticially detect the port when it got printed to the stdout which works for React.js, Vue.js, Angular etc.
```
dgozman, line 164: This is a good place to link to the more detailed description.

src/server/processLauncher.ts
```diff
@@ -172,12 +172,9 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
...
       // Force kill the browser.
       try {
-        if (process.platform === 'win32') {
-          const stdout = childProcess.execSync(`taskkill /pid ${spawnedProcess.pid} /T /F`);
-          options.log(`[pid=${spawnedProcess.pid}] taskkill output: ${stdout.toString()}`);
-        } else {
-          process.kill(-spawnedProcess.pid, 'SIGKILL');
-        }
+        killProcessGroup(spawnedProcess.pid, {
+          onTaskkillExit: (stdout: Buffer): void => options.log(`[pid=${spawnedProcess.pid}] taskkill output: ${stdout.toString()}`)
```
dgozman, line 176: I'd just return the output from `killProcessGroup` instead of passing the callback.

tests/playwright-test/web-server.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+      };
+    `,
+  });
+  expect(result.exitCode).toBe(0);
+  expect(result.passed).toBe(1);
+  expect(result.report.suites[0].specs[0].tests[0].results[0].status).toContain('passed');
+  server.close();
+});
+
+test('should be able to specify the baseURL via the environment variable', async ({ runInlineTest }, { workerIndex }) => {
```
dgozman, line 160: I'd remove this test, since we are not sure about specific usecases just yet. Perhaps also make the env variable more obscure? :)

types/test.d.ts
```diff
@@ -206,6 +230,11 @@ interface ConfigBase {
    */
   updateSnapshots?: UpdateSnapshots;
 
+  /**
+   * Launch a web server before the tests start.
```
dgozman, line 234: "before running tests"

types/test.d.ts
```diff
@@ -118,6 +118,30 @@ export interface Project<TestArgs = {}, WorkerArgs = {}> extends ProjectBase {
...
+   * The port that your server is expected to appear on. If not specified, it does get automatically collected via the 
+   * command output when a localhost URL gets printed.
+   */
+  port?: number,
+  /**
+   * Environment variables which get set during the execution. Default it will use the environment variables from the current process.
+   */
+  env?: Record<string, string>
+  /**
+   * Current working directory of the spawned process. Default its process.cwd().
```
dgozman, line 136: its -> is

types/test.d.ts
```diff
@@ -118,6 +118,30 @@ export interface Project<TestArgs = {}, WorkerArgs = {}> extends ProjectBase {
...
+  command: string,
+  /**
+   * The port that your server is expected to appear on. If not specified, it does get automatically collected via the 
+   * command output when a localhost URL gets printed.
+   */
+  port?: number,
+  /**
+   * Environment variables which get set during the execution. Default it will use the environment variables from the current process.
+   */
+  env?: Record<string, string>
```
dgozman, line 134: - Missing comma
- I think env also takes booleans and numbers? Check Node's types

types/test.d.ts
```diff
@@ -118,6 +118,30 @@ export interface Project<TestArgs = {}, WorkerArgs = {}> extends ProjectBase {
...
+   * Shell command to start the webserver. For example `npm run start`.
+   */
+  command: string,
+  /**
+   * The port that your server is expected to appear on. If not specified, it does get automatically collected via the 
+   * command output when a localhost URL gets printed.
+   */
+  port?: number,
+  /**
+   * Environment variables which get set during the execution. Default it will use the environment variables from the current process.
```
dgozman, line 132: "WebServer environment variables, process.env by default"

docs/src/test-advanced.md
```diff
@@ -200,6 +201,57 @@ export const test = base.extend<{ saveLogs: void }>({
 });
 ```
 
+## Launching a development web server during the tests
+
+To launch a web server during the tests, use the `webServer` option in the [configuration file](#configuration-object).
+
+This will automatically launch and wait for the web server to be ready before the tests will start:
```
dgozman, line 208: Please explain the baseURL connection here, and give an example test that uses it.

tests/playwright-test/web-server.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+import { test, expect } from './playwright-test-fixtures';
+
+test('should create a server', async ({ runInlineTest }, { workerIndex }) => {
+  const port = workerIndex + 10500;
+  const result = await runInlineTest({
+    'test.spec.ts': `
+      const { test } = pwt;
+      test('connect to the server via the full URL navigation', async ({baseURL, page}) => {
+        expect(baseURL).toBe('http://localhost:${port}');
+        await page.goto(baseURL + '/hello');
```
dgozman, line 28: Let's make this just `/hello` to check the baseURL fixture.

src/test/webServer.ts
```diff
@@ -0,0 +1,146 @@
...
+    cancellationToken.canceled = true;
+    if (timedOut)
+      throw new Error(`failed to start web server on port ${port} via "${this.config.command}"`);
+  }
+  public async kill() {
+    assert(this._process);
+    if (this._process.exitCode !== null || this._process.killed)
+      return;
+    const waitForExit = new Promise(resolve => this._process?.on('exit',resolve));
+    killProcessGroup(this._process.pid);
```
dgozman, line 124: I think you'll need all the facilities of processLauncher, e.g. sync kill on process.exit and other such stuff. Let's follow up.

src/server/processLauncher.ts
mxschmitt: since its Windows specific, I would keep it separate for now.

tests/playwright-test/web-server.spec.ts
mxschmitt: Indeed, this is an implementational detail which happens internally, so we can remove it from the tests.

types/test.d.ts
mxschmitt: They are using `NodeJS.ProcessEnv` which is `[key: string]: string | undefined` so I think we are good with it, `undefined` values are ignored referring to their docs.

src/test/webServer.ts
mxschmitt: sounds good, I adjusted it so it uses launchProcess now.

src/server/processLauncher.ts
```diff
@@ -62,29 +63,31 @@ if (maxListeners !== 0)
 
 export async function launchProcess(options: LaunchProcessOptions): Promise<LaunchResult> {
```
dgozman, line 64: Instead of all the changes in this function, why not pass dummy log, empty args and tempDirectories, etc?

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    const { launchedProcess, kill } = await launchProcess({
+      command: this.config.command,
+      env: {
+        ...DEFAULT_ENVIRONMENT_VARIABLES,
+        ...process.env,
+        ...this.config.env,
+      },
+      cwd: this.config.cwd,
+      stdio: 'pipe',
```
dgozman, line 84: I think `stdio: 'stdin'` is better in this case (look at videoRecorder).

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+    });
+    this._killProcess = kill;
+
+    launchedProcess.stdout.pipe(newProcessLogPrefixer()).pipe(process.stdout);
+    launchedProcess.stderr.pipe(newProcessLogPrefixer()).pipe(process.stderr);
+    if (!this.config.port)
+      launchedProcess.stdout.on('data', collectPort);
+    let processExitedWithNonZeroStatusCodeCallback = (error: Error) => { };
+    this._processExitedWithNonZeroStatusCode = new Promise((_, reject) => processExitedWithNonZeroStatusCodeCallback = reject);
+    launchedProcess.on('exit', code => processExitedWithNonZeroStatusCodeCallback(new Error(`WebServer was not able to start. Exit code: ${code}`)));
```
dgozman, line 95: Instead of this, you can now use options.onExit from the launchProcess.

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (this.config.port)
+      return this.config.port;
+    console.log(`Starting WebServer port detection.`);
```
dgozman, line 98: I'd move this to the start of the function and change it to 'Starting web server with ${command}...'

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+    const detectedPort = await Promise.race([
+      this._processExitedWithNonZeroStatusCode,
+      collectPortPromise,
+    ]);
+    console.log(`Port ${detectedPort} by process '${this.config.command}' was automatically detected.`);
```
dgozman, line 103: Instead of this, I'd print 'Using web server at ${baseURL}' at the line 48.

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+  private async _waitForAvailability(port: number) {
+    const launchTimeout = this.config.timeout || 60 * 1000;
+    const cancellationToken = { canceled: false };
+    const { timedOut } = (await Promise.race([
+      raceAgainstDeadline(waitForSocket(port, 100, cancellationToken), launchTimeout + monotonicTime()),
+      this._processExitedWithNonZeroStatusCode,
+    ]));
+    cancellationToken.canceled = true;
+    if (timedOut)
+      throw new Error(`failed to start web server on port ${port} via "${this.config.command}"`);
```
dgozman, line 116: 'Timed out waiting ${timeout}ms for web server'

src/test/webServer.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+  public async kill() {
+    await this._killProcess?.();
```
dgozman, line 119: This is not async anymore, is it?

utils/check_deps.js
```diff
@@ -162,7 +162,7 @@ DEPS['src/utils/'] = ['src/common/', 'src/protocol/'];
 DEPS['src/server/trace/common/'] = ['src/server/snapshot/', ...DEPS['src/server/']];
 DEPS['src/server/trace/recorder/'] = ['src/server/trace/common/', ...DEPS['src/server/trace/common/']];
 DEPS['src/server/trace/viewer/'] = ['src/server/trace/common/', 'src/server/trace/recorder/', 'src/server/chromium/', ...DEPS['src/server/trace/common/']];
-DEPS['src/test/'] = ['src/test/**', 'src/utils/utils.ts'];
+DEPS['src/test/'] = ['src/test/**', 'src/utils/utils.ts', 'src/server/processLauncher.ts'];
```
dgozman, line 165: Hmm... Let's follow up with moving processLauncher to utils?

src/test/webServer.ts
mxschmitt: So you would not write something like "Detecting port.."?

src/test/webServer.ts
mxschmitt: Good catch, types were wrong of the processLauncher, fixed it.

src/server/processLauncher.ts
mxschmitt: Was also thinking about that but then doing it the other way around.

docs/src/test-advanced.md
```diff
@@ -200,6 +201,41 @@ export const test = base.extend<{ saveLogs: void }>({
 });
 ```
 
+## Launching a development web server during the tests
+
+To launch a web server during the tests, use the `webServer` option in the [configuration file](#configuration-object).
+
+Playwright Test does automatically detect if a localhost URL like `http://localhost:3000` gets printed to the stdout.
+The port from the printed URL gets then used to check when its accepting requests and passed over to Playwright as a
+`baseURL`. You can also manually specify a `port` or additional environment variables, see [here](#configuration-object).
```
dgozman, line 210: Can we linkify baseURL to api docs?

src/test/webServer.ts
```diff
@@ -0,0 +1,144 @@
...
+      stdio: 'stdin',
+      shell: true,
+      attemptToGracefullyClose: async () => {},
+      log: () => {},
+      onExit: code => processExitedReject(new Error(`WebServer was not able to start. Exit code: ${code}`)),
+      tempDirectories: [],
+    });
+    this._killProcess = kill;
+
+    launchedProcess.stderr.pipe(newProcessLogPrefixer()).pipe(process.stderr);
```
dgozman, line 99: I think we need to drain `launchedProcess.stdout` just in case, otherwise it may block on Windows because stdio buffer will be full.

docs/src/test-advanced.md
```diff
@@ -200,6 +201,41 @@ export const test = base.extend<{ saveLogs: void }>({
...
+const config = {
+  webServer: {
+    command: 'npm run start',
+    timeout: 120 * 1000,
+  },
+};
+
+mode.exports = config;
+```
+
```
dgozman, line 238: Please add a test example here that navigates to a relative path making use of baseURL.

dgozman (approved):

PR #7370, opened 2021-06-28
feat(test runner): improve reporters api

### LINKED ISSUE #7343: [Feature] Support for Async event hooks for Folio Reporter
Currently folio reporter expects event hooks like onEnd(), onTestEnd() to be synchronous, which creates a limitation when there is a requirement to make async calls in those functions, for instance making a network call when test ends. 

Similar test reporters like jest support event hooks to be async, below is the jest ReporterDispatcher [https://github.com/facebook/jest/blob/5f4dd187d89070d07617444186684c20d9213031/packages/jest-core/src/ReporterDispatcher.ts](url)

Having a similar change would enable anyone to await async calls in events like onEnd() . This would essentially require the following changes:

- **multiplexer.ts** – Update event hooks like onEnd, onTestEnd to be async and awaiting the reporter event hook calls. 
- **runner.ts** – Update event hook calls to be awaited. this._reporter.onEnd(); => await this._reporter.onEnd();

Here is a[ sample PR ](https://github.com/microsoft/folio/pull/332) which has sample changes to make onEnd() async.
This is just a basic PR suggesting the changes and may not be in a good state to be merged.

dgozman:
- onEnd may return a Promise
- onEnd now takes a result for the full run
- onTimeout is replaced with onEnd(result)

Fixes #7343.

### DIFF
--- a/src/test/reporter.ts
+++ b/src/test/reporter.ts
@@ -60,13 +60,15 @@ export interface TestResult {
   stdout: (string | Buffer)[];
   stderr: (string | Buffer)[];
 }
+export type FullResult = {
+  status: 'passed' | 'failed' | 'timedout' | 'interrupted';
+};
 export interface Reporter {
   onBegin(config: FullConfig, suite: Suite): void;
   onTestBegin(test: Test): void;
   onStdOut(chunk: string | Buffer, test?: Test): void;
   onStdErr(chunk: string | Buffer, test?: Test): void;
   onTestEnd(test: Test, result: TestResult): void;
-  onTimeout(timeout: number): void;
   onError(error: TestError): void;
-  onEnd(): void;
+  onEnd(result: FullResult): void | Promise<void>;
 }
--- a/tests/playwright-test/config.spec.ts
+++ b/tests/playwright-test/config.spec.ts
@@ -350,7 +350,8 @@ test('should work with custom reporter', async ({ runInlineTest }) => {
         onError() {
           console.log('\\n%%reporter-error%%');
         }
-        onEnd() {
+        async onEnd() {
+          await new Promise(f => setTimeout(f, 500));
           console.log('\\n%%reporter-end%%' + this.options.end);
         }
       }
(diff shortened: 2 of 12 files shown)


JoelEinbinder:
We do breaking changes to the reporters?

tests/playwright-test/config.spec.ts
```diff
@@ -350,7 +350,8 @@ test('should work with custom reporter', async ({ runInlineTest }) => {
         onError() {
           console.log('\\n%%reporter-error%%');
         }
-        onEnd() {
+        async onEnd() {
+          await new Promise(f => setTimeout(f, 500));
```
JoelEinbinder, line 354:
```suggestion
          await new Promise(f => setTimeout(f, 50));
```

src/test/reporter.ts
```diff
@@ -60,13 +60,15 @@ export interface TestResult {
   stdout: (string | Buffer)[];
   stderr: (string | Buffer)[];
 }
+export type FullResult = {
+  status: 'passed' | 'failed' | 'timedout' | 'interrupted';
```
JoelEinbinder, line 64: What happens if you call test.skip in the test body?

mxschmitt:
> We do breaking changes to the reporters?

since no one is using them and its an undocumented feature its fine IMO.

src/test/reporter.ts
```diff
@@ -60,13 +60,15 @@ export interface TestResult {
...
+};
 export interface Reporter {
   onBegin(config: FullConfig, suite: Suite): void;
   onTestBegin(test: Test): void;
   onStdOut(chunk: string | Buffer, test?: Test): void;
   onStdErr(chunk: string | Buffer, test?: Test): void;
   onTestEnd(test: Test, result: TestResult): void;
```
pavelfeldman, line 71: While you are here: https://github.com/microsoft/playwright/issues/7343

src/test/reporter.ts
dgozman: It is skipped. How's that related?

JoelEinbinder: Oh I misread it. This timeout is for the entire test run, got it.

PR #7373, opened 2021-06-29
fix(test-runner): work with .mjs files

JoelEinbinder:


packages/installation-tests/esm-playwright-test.mjs
```diff
@@ -14,11 +14,21 @@
  * limitations under the License.
  */
 
-import { chromium, firefox, webkit, selectors, devices, errors, test } from '@playwright/test';
+import { chromium, firefox, webkit, selectors, devices, errors, test, expect } from '@playwright/test';
```
JoelEinbinder, line 17: This test needed to be stronger. Now i checks that `expect` is also exported (it wasn't) and that both `test` and `expect` are functions.

packages/playwright-test/index.mjs
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import * as playwright from './index.js';
+import playwright from './index.js';
```
JoelEinbinder, line 17: When node imports a commonjs module, it sets the entire module as the default. There is no way to destructure values from a commonjs module, and if you do `* as foo` you get an object like `{default: 'foo'}`.

src/test/loader.ts
```diff
@@ -216,7 +226,7 @@ function toReporters(reporters: 'dot' | 'line' | 'list' | 'junit' | 'json' | 'nu
 }
 
 function errorWithFile(file: string, message: string) {
-  return new Error(`${file}: ${message}`);
+  return new Error(`${path.relative(process.cwd(), file)}: ${message}`);
```
JoelEinbinder, line 229: drive-by: all these error messages had ugly absolute paths. Now they should be a lot smaller and nicer.

src/test/transform.ts
```diff
@@ -107,8 +108,10 @@ export function wrapFunctionWithLocation<A extends any[], R>(func: (location: Lo
     const oldPrepareStackTrace = Error.prepareStackTrace;
     Error.prepareStackTrace = (error, stackFrames) => {
       const frame: NodeJS.CallSite = sourceMapSupport.wrapCallSite(stackFrames[1]);
+      const fileName = frame.getFileName();
+      const file = (fileName && fileName.startsWith('file://')) ? url.fileURLToPath(fileName) : fileName;
```
JoelEinbinder, line 112: node error stacks for modules use `file://` urls instead of paths.

packages/installation-tests/esm-playwright-test.mjs
mxschmitt: We could also add `expect(1).toBe(1)` to be on the safe side. Since this call is lightweight it makes probably sense to add it, wdyt?

src/test/loader.ts
```diff
@@ -191,7 +202,7 @@ export class Loader {
       name: takeFirst(this._configOverrides.name, projectConfig.name, this._config.name, ''),
       testDir,
       testIgnore: takeFirst(this._configOverrides.testIgnore, projectConfig.testIgnore, this._config.testIgnore, []),
-      testMatch: takeFirst(this._configOverrides.testMatch, projectConfig.testMatch, this._config.testMatch, '**/?(*.)+(spec|test).[jt]s'),
+      testMatch: takeFirst(this._configOverrides.testMatch, projectConfig.testMatch, this._config.testMatch, '**/?(*.)+(spec|test).+(ts|js|mjs)'),
```
mxschmitt, line 205: Is the + there intended? This would mean:

`foo.spec.tststs` would also be a match.

src/test/loader.ts
```diff
@@ -216,7 +227,7 @@ function toReporters(reporters: 'dot' | 'line' | 'list' | 'junit' | 'json' | 'nu
 }
 
 function errorWithFile(file: string, message: string) {
-  return new Error(`${file}: ${message}`);
+  return new Error(`${path.relative(process.cwd(), file)}: ${message}`);
```
mxschmitt, line 230: I think we should use there the project rootDir which is stored inside the config. At least thats what I did with other related error reporting.

src/test/transform.ts
mxschmitt: (Would that be a reason to add it as a comment? - not sure about the comment rule here yet)

tests/playwright-test/loader.spec.ts
```diff
@@ -104,3 +104,35 @@ test('should validate configuration object', async ({ runInlineTest }) => {
...
+        });
+      `
+  });
+  expect(passed).toBe(1);
+  expect(exitCode).toBe(0);
+});
+
+test('should throw a nice error if a js file uses import', async ({ runInlineTest }) => {
+  // modules don't work with node < 12
+  test.skip(parseInt(process.versions.node, 10) < 12);
```
mxschmitt, line 125: We never run tests with Node.js < 12 and we only support 12+. So we can remove that.

tests/playwright-test/loader.spec.ts
```diff
@@ -104,3 +104,35 @@ test('should validate configuration object', async ({ runInlineTest }) => {
   expect(result.failed).toBe(0);
   expect(result.output).toContain('playwright.config.ts: config.timeout must be a non-negative number');
 });
+
+test('should load an mjs file', async ({ runInlineTest }) => {
+  // modules don't work with node < 12
+  test.skip(parseInt(process.versions.node, 10) < 12);
```
mxschmitt, line 110: ditto.

tests/playwright-test/playwright-test-fixtures.ts
```diff
@@ -69,9 +75,10 @@ async function writeFiles(testInfo: TestInfo, files: Files) {
   await Promise.all(Object.keys(files).map(async name => {
     const fullName = path.join(baseDir, name);
     await fs.promises.mkdir(path.dirname(fullName), { recursive: true });
-    const isTypeScriptSourceFile = name.endsWith('ts') && !name.endsWith('d.ts');
-    const header = isTypeScriptSourceFile ? headerTS : headerJS;
-    if (/(spec|test)\.(js|ts)$/.test(name)) {
+    const isTypeScriptSourceFile = name.endsWith('.ts') && !name.endsWith('.d.ts');
+    const isJSModule = name.endsWith('.mjs');
+    const header = isTypeScriptSourceFile ? headerTS : (isJSModule ? headerMJS : headerJS);
```
mxschmitt, line 80: sounds like we should move it inside an own function `getFileHeader(fileName: string): string`

mxschmitt (approved):

src/test/loader.ts
JoelEinbinder: I'll fix the +(spec|test) in a follow up.

tests/playwright-test/playwright-test-fixtures.ts
JoelEinbinder: I prefer that code only called once stays inline.

src/test/loader.ts
JoelEinbinder: Undid this because its controversial.

src/test/workerRunner.ts
```diff
@@ -184,7 +184,7 @@ export class WorkerRunner extends EventEmitter {
     const testId = test._id;
 
     const baseOutputDir = (() => {
-      const relativeTestFilePath = path.relative(this._project.config.testDir, spec._requireFile.replace(/\.(spec|test)\.(js|ts)/, ''));
+      const relativeTestFilePath = path.relative(this._project.config.testDir, spec._requireFile.replace(/\.(spec|test)\.(js|ts|mjs)/, ''));
```
aslushnikov, line 187: this is missing a '$' in the end

aslushnikov (approved):

PR #7374, opened 2021-06-29
fix(registry): make hostPlatform lazy

dgozman:
This avoids unnecessary execSync at the time of require('playwright').

### DIFF
--- a/src/utils/registry.ts
+++ b/src/utils/registry.ts
@@ -173,7 +173,7 @@ const DOWNLOAD_URLS = {
   },
 };
 
-export const hostPlatform = ((): BrowserPlatform => {
+function lazyHostPlatform(): BrowserPlatform {
   const platform = os.platform();
   if (platform === 'darwin') {
     const [major, minor] = execSync('sw_vers -productVersion', {
@@ -208,7 +208,7 @@ export const hostPlatform = ((): BrowserPlatform => {
   if (platform === 'win32')
     return os.arch() === 'x64' ? 'win64' : 'win32';
   return platform as BrowserPlatform;
-})();
+}
 
 export const registryDirectory = (() => {
   let result: string;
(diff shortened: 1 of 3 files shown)


JoelEinbinder (approved):

src/utils/registry.ts
```diff
@@ -173,7 +173,7 @@ const DOWNLOAD_URLS = {
   },
 };
 
-export const hostPlatform = ((): BrowserPlatform => {
+function lazyHostPlatform(): BrowserPlatform {
```
mxschmitt, line 176:
```suggestion
function hostPlatform(): BrowserPlatform {
```
I would not call it lazy, since it doesn't contain any logic to load it e.g. asynchronously or store its value somewhere. Normal `hostPlatform` should be fine. Ah saw you are caching it later on, but then this file has nothing to do with cache.

src/utils/registry.ts
```diff
@@ -263,36 +265,52 @@ export class Registry {
...
-        // To avoid older registries erroneously removing 'webkit-technology-preview', we have to
-        // ensure that browser folders to never include dashes inside.
-        browserDirectory: browserDirectoryPrefix.replace(/-/g, '_') + '-' + revision,
-      };
-    });
+    this._packagePath = packagePath;
+  }
+
+  private _descriptors(): BrowserDescriptor[] {
+    if (!this._loadedDescriptors) {
```
mxschmitt, line 272: lets invert it to make it more readable and reduce nesting?
```js
if (this._loadedDescriptors)
  return this._loadedDescriptors;
... // old code
```

mxschmitt (approved):

pavelfeldman:
```
function hostVersion() {
  const ver = require('os').release().split('.').map((a: string) => parseInt(a));
  if (ver[0] === 18)
    return { major: 10, minor: 14 };
  if (ver === 19)
    return { major: 10, minor: 15 };
  return { major: ver[0] - 9, minor: ver[1] - 1 };
}
```

PR #7376, opened 2021-06-29
fix(tracing): explicitly list actions to snapshot

### LINKED ISSUE #7371: [Feature] Group duplicate actions in trace viewer
In our tests we setup a `page.route` listener that conditionally blocks a route based on its contents:

```js
page.route('/some/route', (route: Route) => {
  const body = route.request().postDataJSON()

  if (body.something) {
    return route.abort()
  }

  return route.continue()
})
```

this is on every page in our [fixture](https://playwright.dev/docs/test-fixtures). Unfortunately, the request happens frequently enough that our traces look like this in between page actions:

Screen Shot 2021-06-28 at 7 08 24 PM

it'd be helpful if duplicate actions were grouped together, perhaps accordion-style and collapsed by default.

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/7371


src/server/trace/recorder/tracing.ts
```diff
@@ -230,4 +225,70 @@ export class Tracing implements InstrumentationListener {
...
+  'Page.keyboardDown',
+  'Page.keyboardUp',
+  'Page.keyboardInsertText',
+  'Page.keyboardType',
+  'Page.keyboardPress',
+  'Page.mouseMove',
+  'Page.mouseDown',
+  'Page.mouseUp',
+  'Page.mouseClick',
+  'Page.touchscreenTap',
```
mxschmitt, line 293: I'm not a huge fan of this list, we keep adding new APIs and then we might forget one to add there. :/

pavelfeldman: I am with you, but I have not figured out anything better.

mattjennings: Would it be better to have a blacklist that filters out instead of a whitelist that filters in?  Maybe even allow that list to be overridden by playwright.config?

dgozman (approved):

mxschmitt (approved):

PR #7382, opened 2021-06-29
fix(test-runner): better test match glob

JoelEinbinder:
We copied the test match from jest, but it incorrectly matches `specspecspec.ts`.
https://github.com/micromatch/micromatch#extglobs

### DIFF
--- a/tests/playwright-test/loader.spec.ts
+++ b/tests/playwright-test/loader.spec.ts
@@ -105,7 +105,7 @@ test('should validate configuration object', async ({ runInlineTest }) => {
   expect(result.output).toContain('playwright.config.ts: config.timeout must be a non-negative number');
 });
 
-test.only('should match tests well', async ({ runInlineTest }) => {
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
@dgozman this could be considered breaking, but I think it is a bug fix?

tests/playwright-test/loader.spec.ts
```diff
@@ -104,3 +104,47 @@ test('should validate configuration object', async ({ runInlineTest }) => {
   expect(result.failed).toBe(0);
   expect(result.output).toContain('playwright.config.ts: config.timeout must be a non-negative number');
 });
+
+test.only('should match tests well', async ({ runInlineTest }) => {
```
JoelEinbinder, line 108:
```suggestion
test('should match tests well', async ({ runInlineTest }) => {
```

dgozman (approved):

PR #7384, opened 2021-06-29
feat(test-runner): smaller error message paths

JoelEinbinder:
We were dumping absolute paths which is a bit impolite.

This uses process.cwd() instead of rootDir. My gut says cwd is better
and my brain says its because not all of these files are necessarily
in the rootDir. The config, for example, might be somewhere else.﻿

### DIFF
--- a/src/test/fixtures.ts
+++ b/src/test/fixtures.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { wrapInPromise } from './util';
+import { formatLocation, wrapInPromise } from './util';
 import * as crypto from 'crypto';
 import { FixturesWithLocation, Location, WorkerInfo, TestInfo } from './types';
 
@@ -352,8 +352,4 @@ function errorWithLocations(message: string, ...defined: { location: Location, n
     message += `\n  ${prefix}defined at ${formatLocation(location)}`;
   }
   return new Error(message);
-}
-
-function formatLocation(location: Location) {
-  return location.file + ':' + location.line + ':' + location.column;
 }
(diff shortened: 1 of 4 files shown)


mxschmitt:
(if we change it, would be also good to change it [here](https://github.com/microsoft/playwright/blob/a270fc52069ca83bc58bf1afbca89191da64b644/src/test/runner.ts#L390))

dgozman:
Please also change this line: https://github.com/microsoft/playwright/blob/a270fc52069ca83bc58bf1afbca89191da64b644/src/test/fixtures.ts#L347

There may be more errors like this...

mxschmitt:
IMO its better to have it based on rootDir. [Jest is also using it](https://sourcegraph.com/github.com/facebook/jest/-/blob/packages/jest-message-util/src/index.ts?L249) and it feels better from my point of view and terminal experience.

JoelEinbinder:
> IMO its better to have it based on rootDir. [Jest is also using it](https://sourcegraph.com/github.com/facebook/jest/-/blob/packages/jest-message-util/src/index.ts?L249) and it feels better from my point of view and terminal experience.

That makes sense for test files, which are necessarily in the rootDir. But config files/reporters/etc might not be in the root dir, so the cwd is the best context we have for them.

mxschmitt:
Okay agreed 👍   Seems more convenient, when e.g. you open just a subfolder of your project, then the terminal gets spawned there, you execute the tests and they get reported based on the cwd.

Jest seems sometimes doing it:

- the file name at the top is cwd relative
- the file name in the middle in the stacktrace is relative to your rootDir

image

dgozman (approved):

PR #7387, opened 2021-06-29
feat(test runner): support test.setTimeout for the whole block

### LINKED ISSUE #7266: [Question] test.setTimeout() in test hooks
I have an expensive calculation in `beforeAll`.
So, I need to increase the timeout value for this specific test. (playwright default `30000`)
When I used Jest, I just put `jest.setTimeout()` top of the file.

Now, I'm using Playwright Test. And I found `test.setTimeout()` which looks similar to `jest.setTimeout()`.
But, `test.setTimeout()` can only be called inside 1.test or 2.fixture.
I tried 1. but failed because heavy calculation was in `beforeAll` hook.

Do I have to use fixture? How to solve this problem?

dgozman:
This enables `test.setTimeout()` outside of the test body, that affects all tests in the block (either file or describe).

Drive-by: fix list reporter regression.

Fixes #7266.

### DIFF
--- a/src/test/testType.ts
+++ b/src/test/testType.ts
@@ -125,9 +125,15 @@ export class TestTypeImpl {
   }
 
   private _setTimeout(timeout: number) {
+    const suite = currentlyLoadingFileSuite();
+    if (suite) {
+      suite._timeout = timeout;
+      return;
+    }
+
     const testInfo = currentTestInfo();
     if (!testInfo)
-      throw new Error(`test.setTimeout() can only be called inside test or fixture`);
+      throw new Error(`test.setTimeout() can only be called from a test file`);
     testInfo.setTimeout(timeout);
   }
 
--- a/tests/playwright-test/timeout.spec.ts
+++ b/tests/playwright-test/timeout.spec.ts
@@ -68,6 +68,35 @@ test('should respect test.setTimeout', async ({ runInlineTest }) => {
   expect(result.output).toContain('Timeout of 1000ms exceeded');
 });
 
+test('should respect test.setTimeout outside of the test', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.spec.ts': `
+      const { test } = pwt;
+
+      test.setTimeout(500);
+      test('fails', async ({}) => {
+        await new Promise(f => setTimeout(f, 1000));
+      });
+      test('passes', async ({}) => {
+        await new Promise(f => setTimeout(f, 100));
+      });
+
+      test.describe('suite', () => {
+        test.setTimeout(50);
+        test('fails', async ({}) => {
+          await new Promise(f => setTimeout(f, 100));
+        });
+        test('passes', async ({}) => {
+        });
+      });
+    `
+  });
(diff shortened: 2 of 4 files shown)


src/test/testType.ts
```diff
@@ -125,9 +125,15 @@ export class TestTypeImpl {
   }
 
   private _setTimeout(timeout: number) {
+    const suite = currentlyLoadingFileSuite();
```
JoelEinbinder, line 128: currentlyLoadingFileSuite returns the active describe suite? Strange

src/test/testType.ts
```diff
@@ -125,9 +125,15 @@ export class TestTypeImpl {
...
+    if (suite) {
+      suite._timeout = timeout;
+      return;
+    }
+
     const testInfo = currentTestInfo();
     if (!testInfo)
-      throw new Error(`test.setTimeout() can only be called inside test or fixture`);
+      throw new Error(`test.setTimeout() can only be called inside test, describe block or fixture`);
```
JoelEinbinder, line 136: How can this error happen? If we are in a config? Maybe it should say something like `test.setTimeout() can only be called from a test file`?

tests/playwright-test/timeout.spec.ts
```diff
@@ -68,6 +68,35 @@ test('should respect test.setTimeout', async ({ runInlineTest }) => {
...
+      });
+
+      test.describe('suite', () => {
+        test.setTimeout(50);
+        test('fails', async ({}) => {
+          await new Promise(f => setTimeout(f, 100));
+        });
+        test('passes', async ({}) => {
+        });
+      });
```
JoelEinbinder, line 91: Whats the intended/actual behavior of this code
```ts
test.setTimeout(50);
test.describe(() => {
        test('unknown', async ({}) => {
          await new Promise(f => setTimeout(f, 100));
        });
});
test.setTimeout(500);
```

src/test/reporters/list.ts
```diff
@@ -79,7 +79,7 @@ class ListReporter extends BaseReporter {
     if (result.status === 'skipped') {
       text = colors.green('  - ') + colors.cyan(title);
     } else {
-      const statusMark = ('  ' + (result.status === 'passed' ? POSITIVE_STATUS_MARK : NEGATIVE_STATUS_MARK)).padEnd(5);
+      const statusMark = ('  ' + (result.status === 'passed' ? POSITIVE_STATUS_MARK : NEGATIVE_STATUS_MARK)).padEnd(4);
```
JoelEinbinder, line 82: ?

JoelEinbinder (approved):

src/test/testType.ts
dgozman: Yeah, we'll refactor it.

src/test/testType.ts
dgozman: Sure.

tests/playwright-test/timeout.spec.ts
dgozman: Actual behavior is 500 for all tests. What would you expect?

PR #7393, opened 2021-06-30
test(trace-viewer): start adding tests

pavelfeldman:


dgozman (approved):

mxschmitt (approved):
Can we maybe register page.on(pageError => process.emit(unhandledException)) to verify that the trace viewer does not crash / throw exceptions?

pavelfeldman:
> Can we maybe register page.on(pageError => process.emit(unhandledException)) to verify that the trace viewer does not crash / throw exceptions?

Sounds like a good idea!

PR #7407, opened 2021-06-30
fix(chromium): disable use-zoom-for-dsf

### LINKED ISSUE #7362: [Internal] page-drag tests fail on macOS 11.0

JoelEinbinder:
On Mac, use-zoom-for-dsf﻿was enabled by default which caused `Input.dispatchDragEvent` to miss
its target. Disabling the flag while we wait for an upstream fix.

Fixes #7362

### DIFF
--- a/src/server/chromium/chromium.ts
+++ b/src/server/chromium/chromium.ts
@@ -150,6 +150,11 @@ export class Chromium extends BrowserType {
       throw new Error('Arguments can not specify page to be opened');
     const chromeArguments = [...DEFAULT_ARGS];
     chromeArguments.push(`--user-data-dir=${userDataDir}`);
+
+    // See https://github.com/microsoft/playwright/issues/7362
+    if (os.platform() === 'darwin')
+      chromeArguments.push('--enable-use-zoom-for-dsf=false');
+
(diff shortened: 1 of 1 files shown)


src/server/chromium/chromium.ts
```diff
@@ -221,6 +221,7 @@ const DEFAULT_ARGS = [
   '--use-mock-keychain',
   // See https://chromium-review.googlesource.com/c/chromium/src/+/2436773
   '--no-service-autorun',
+  '--enable-use-zoom-for-dsf=false',
```
dgozman, line 224: Let's only do this on Mac? I am afraid that screenshots on Windows will change if you disable this feature.

JoelEinbinder: All tests pass? I'm afraid that dragging will break on windows if its enabled!

JoelEinbinder: I don't have access to a high dpi windows computer, but I wasn't able to repro this issue on linux. Added the flag for macos only for now.

dgozman (approved):

mxschmitt (approved):

PR #7409, opened 2021-06-30
feat: add browser.newContext({ baseUrl })

### LINKED ISSUE #7408: [Feature] browser.newContext({ baseURL })
```js
await browser.newContext({
  baseURL: "http://localhost:3000",
})
await page.goto("/foobar") -> "http://localhost:3000/foobar"

await page.goto("http://microsoft.com") -> http://microsoft.com
```

mxschmitt:
Relevant related implementation for customers who are migrating see [here](https://github.com/cypress-io/cypress/blob/b457010bd742a64da64daf7816ce8832f63990f7/packages/driver/src/cypress/location.js#L199).

There is `pytest-base-url`, but they don't construct a new URL, see [here](https://github.com/pytest-dev/pytest-base-url).

Fixes #7408

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -852,6 +852,7 @@ Returns all open pages in the context.
 
 Routing provides the capability to modify network requests that are made by any page in the browser context. Once route
 is enabled, every request matching the url pattern will stall unless it's continued, fulfilled or aborted.
+It will consider the base URL if you set it via [`option: baseURL`] when creating the context.
 
 An example of a naive handler that aborts all image requests:
 
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -852,7 +852,6 @@ Returns all open pages in the context.
 
 Routing provides the capability to modify network requests that are made by any page in the browser context. Once route
 is enabled, every request matching the url pattern will stall unless it's continued, fulfilled or aborted.
-It will consider the base URL if you set it via [`option: baseURL`] when creating the context.
 
 An example of a naive handler that aborts all image requests:
 
@@ -1002,6 +1001,8 @@ Enabling routing disables http cache.
 - `url` <[string]|[RegExp]|[function]\([URL]\):[boolean]>
 
 A glob pattern, regex pattern or predicate receiving [URL] to match while routing.
+When a [`option: baseURL`] via the context options was provided and the passed URL is a path,
+it gets merged via the [`new URL()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) constructor.
 
 ### param: BrowserContext.route.handler
 * langs: js, python
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -216,10 +216,9 @@ Toggles bypassing page's Content-Security-Policy.
 ## context-option-baseURL
 - `baseURL` <[string]>
 
-Prefixes the URL on navigations when a path (no origin) is passed to [`method: Page.goto`]. You can specify also a
-base URL with a path. Examples:
+When using [`method: Page.goto`], [`method: Page.route`], [`method: Page.waitForURL`], [`method: Page.waitForRequest`], or [`method: Page.waitForResponse`] it takes the base URL in consideration by using the [`URL()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) constructor for building the corresponding URL. Examples:
 * baseURL: `http://localhost:3000` and navigating to `/bar.html` results in `http://localhost:3000/bar.html`
-* baseURL: `http://localhost:3000/foo/` and navigating to `/bar.html` results in `http://localhost:3000/foo/bar.html`
+* baseURL: `http://localhost:3000/foo/` and navigating to `./bar.html` results in `http://localhost:3000/foo/bar.html`
 
 ## context-option-viewport
 * langs: js, java
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -213,6 +213,14 @@ Whether to ignore HTTPS errors during navigation. Defaults to `false`.
 
 Toggles bypassing page's Content-Security-Policy.
 
+## context-option-baseURL
+- `baseURL` <[string]>
+
+Prefixes the URL on navigations when a path (no origin) is passed to [`method: Page.goto`]. You can specify also a
+base URL with a path. Examples:
+* baseURL: `http://localhost:3000` and navigating to `/bar.html` results in `http://localhost:3000/bar.html`
+* baseURL: `http://localhost:3000/foo/` and navigating to `/bar.html` results in `http://localhost:3000/foo/bar.html`
+
 ## context-option-viewport
 * langs: js, java
   - alias-java: viewportSize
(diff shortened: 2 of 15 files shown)


docs/src/api/params.md
```diff
@@ -213,6 +213,11 @@ Whether to ignore HTTPS errors during navigation. Defaults to `false`.
 
 Toggles bypassing page's Content-Security-Policy.
 
+## context-option-baseUrl
+- `baseUrl` <[string]>
```
dgozman, line 217: I now we have `waitForURL` with all caps. Do we have a `Url` precedent in our api?

docs/src/api/params.md
```diff
@@ -213,6 +213,11 @@ Whether to ignore HTTPS errors during navigation. Defaults to `false`.
...
+
+Prefixes the URL on navigations when a path (no URL) is passed to [`method: Page.goto`].
```
dgozman, line 219: Please add an example here with relative and absolute page.goto(). Also, strictly speaking, path+query+hash also works, so perhaps say "no origin" instead of "path"?

src/server/frames.ts
```diff
@@ -556,8 +557,9 @@ export class Frame extends SdkObject {
   }
 
   async goto(metadata: CallMetadata, url: string, options: types.GotoOptions = {}): Promise<network.Response | null> {
+    const constructedNavigationUrl = constructNavigationUrl(this._page._browserContext._options.baseUrl, url);
```
dgozman, line 560: I am thinking whether any other urls we have should follow this pattern? `waitForURL`, `waitForRequest`, `route`, something else?

src/server/frames.ts
```diff
@@ -1371,3 +1373,15 @@ function verifyLifecycle(name: string, waitUntil: types.LifecycleEvent): types.L
...
 }
+
+function constructNavigationUrl(baseUrl: string | undefined, givenUrl: string): string {
+  if (!baseUrl)
+    return givenUrl;
+  try {
+    new URL(givenUrl);
+    return givenUrl;
+  } catch (error) {
+    const optionalSlash = (givenUrl && baseUrl[baseUrl.length] !== '/' && givenUrl[0] !== '/') ? '/' : '';
```
dgozman, line 1384: - `baseUrl[baseUrl.length]` is always undefined
- Let's make `http://localhost:3000/build/` and `/foo.html` play together nicely?

tests/browsercontext-base-url.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+it('should construct a new URL when a relative path in browser.newPage is passed to page.goto', async function({browser, server}) {
+  const page = await browser.newPage({
+    baseUrl: server.PREFIX + '/url-construction',
+  });
+  const response = await page.goto('mypage.html');
+  expect(response.url()).toBe(server.PREFIX + '/url-construction/mypage.html');
+  await page.close();
+});
+
+it('should not navigate to the base URL if / or empty string is passed', async function({browser, server}) {
```
dgozman, line 57: Title says "should not", but expect says "should"?

tests/browsercontext-base-url.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  const response = await page.goto('/empty.html');
+  expect(response.url()).toBe(server.EMPTY_PAGE);
+  await page.close();
+});
+
+it('should not construct a new URL when the new given URL is valid', async function({browser, server}) {
+  const page = await browser.newPage({
+    baseUrl: 'http://microsoft.com',
+  });
+  const response = await page.goto(server.EMPTY_PAGE);
```
dgozman, line 43: Let's also try 'about:blank' and 'data:text/html,Hello world'

src/server/frames.ts
mxschmitt: On navigation is in my mind the most crucial part. For route I don't want to have more magic, we already have glob/regexp and predicates there IMO.

mxschmitt:
I reverted the URL building via `new URL()` because I though it adds too much complexity. I was only able to use it with these conditions inside the catch.

```js
    if (givenURL[0] === '/' && baseURL[baseURL.length - 1] !== '/') {
      givenURL = givenURL.substr(1);
      baseURL = baseURL + '/';
    }
    if (baseURL[baseURL.length - 1] !== '/' && givenURL[0] !== '/')
      baseURL = baseURL + '/';

    if (givenURL[0] === '/')
      givenURL = '.' + givenURL;
    else if (givenURL[0]  !== '.')
      givenURL = './' + givenURL;

    return (new URL.URL(givenURL, baseURL)).toString();
```

docs/src/api/class-browsercontext.md
```diff
@@ -852,6 +852,7 @@ Returns all open pages in the context.
 
 Routing provides the capability to modify network requests that are made by any page in the browser context. Once route
 is enabled, every request matching the url pattern will stall unless it's continued, fulfilled or aborted.
+It will consider the base URL if you set it via [`option: baseURL`] when creating the context.
```
dgozman, line 855: I don't think this tells me much. Can we move this to the description of the first "url" parameter, and give an example of how baseURL affects it? Same for other places.

docs/src/api/params.md
```diff
@@ -213,6 +213,14 @@ Whether to ignore HTTPS errors during navigation. Defaults to `false`.
 
 Toggles bypassing page's Content-Security-Policy.
 
+## context-option-baseURL
+- `baseURL` <[string]>
+
+Prefixes the URL on navigations when a path (no origin) is passed to [`method: Page.goto`]. You can specify also a
+base URL with a path. Examples:
+* baseURL: `http://localhost:3000` and navigating to `/bar.html` results in `http://localhost:3000/bar.html`
+* baseURL: `http://localhost:3000/foo/` and navigating to `/bar.html` results in `http://localhost:3000/foo/bar.html`
```
dgozman, line 222: Is this example correct now? Shouldn't it say `./bar.html`?

src/client/clientHelper.ts
```diff
@@ -65,9 +65,11 @@ export function parsedURL(url: string): URL | null {
   }
 }
 
-export function urlMatches(urlString: string, match: types.URLMatch | undefined): boolean {
+export function urlMatches(baseURL: string | undefined, urlString: string, match: types.URLMatch | undefined): boolean {
   if (match === undefined || match === '')
     return true;
+  if (baseURL && isString(match) && !match.startsWith('*') && urlMatches(undefined, urlString, constructURLBasedOnBaseURL(baseURL, match)))
```
dgozman, line 71:
```js
if (isString(match) && !match.startsWith('*'))
  match = constructURLBasedOnBaseURL(baseURL, match);
```

src/utils/utils.ts
```diff
@@ -313,3 +313,16 @@ export function getUserAgent() {
   const packageJson = require('./../../package.json');
   return `Playwright/${packageJson.version} (${os.arch()}/${os.platform()}/${os.release()})`;
 }
+
+export function constructURLBasedOnBaseURL(baseURL: string | undefined, givenURL: string): string {
```
dgozman, line 317: Hmm...  I'd make it safer this way:

```js
try {
  return (new URL.URL(givenURL, baseURL)).toString();
} catch (e) {
  return givenURL;
}
```

Why do we need to special-case '/' or ''? Same for `!baseURL`.

docs/src/api/params.md
mxschmitt: with the new implementation you are absolutely right! 👍

docs/src/api/class-browsercontext.md
mxschmitt: good idea with moving it, but not sure if it's better how its now with the wording.

dgozman (approved):

PR #7410, opened 2021-06-30
fix(launch): replace 'npm install playwright' with 'npx playwright install'

### LINKED ISSUE #7342: [BUG] Better exception message when executable is not there
Expected: hey use npx playwright install chrome to probably fix it
Actual: Executable path file does not exist

Relevant code https://github.com/microsoft/playwright/blob/103f8ddd06d81beebca13ced8a1589d5679e5457/src/server/browserType.ts#L168-L169

Problem: Users who are using Playwright Test get recommended to install playwright additionally. We should print use npx playwright install.

dgozman:
... in the error message

Fixes #7342.

### DIFF
--- a/src/server/browserType.ts
+++ b/src/server/browserType.ts
@@ -166,7 +166,7 @@ export abstract class BrowserType extends SdkObject {
       const errorMessageLines = [`Failed to launch ${this._name} because executable doesn't exist at ${executable}`];
       // If we tried using stock downloaded browser, suggest re-installing playwright.
       if (!executablePath)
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):
Just thinking out loud:
- This works when they use the test-runner, since we have a `playwright` script which then calls `@playwight/test`.
- This works when they only use Playwright, since playwright ends up in the Playwright NPM package.

Can we maybe make it more prominent e.g. with using colors - reasoning behind it is that its one of our most viewed GitHub issues?

aslushnikov (approved):

PR #7412, opened 2021-06-30
test(drag): add high dpi drag test

JoelEinbinder:
Adds a test that verifies drag and drop works when the device scale factor is 2.

### DIFF
--- a/tests/browsercontext-viewport.spec.ts
+++ b/tests/browsercontext-viewport.spec.ts
@@ -111,3 +111,14 @@ browserTest('should report null viewportSize when given null viewport', async ({
   expect(page.viewportSize()).toBe(null);
   await context.close();
 });
+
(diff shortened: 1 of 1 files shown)


tests/page/pageTest.ts
```diff
@@ -27,4 +27,4 @@ export type PageTestFixtures = {
   isElectron: boolean;
 };
 
-export const test = baseTest.declare<PageTestFixtures>();
+export const test = baseTest.declare<PageTestFixtures, { browser: Browser }>();
```
dgozman, line 30: Page tests do not provide a browser instance, for example when running with Android.

mxschmitt: (need to be moved outside of the page folder)

dgozman (approved):

PR #7417, opened 2021-07-01
test(pageError): add a failing FF test for page error

pavelfeldman:


mxschmitt (approved):
Maybe a fixme instead?

PR #7418, opened 2021-07-01
feat(trace-viewer): render console messages

pavelfeldman:


tests/trace-viewer/trace-viewer.spec.ts
```diff
@@ -66,12 +81,18 @@ const test = playwrightTest.extend<{ showTraceViewer: (trace: string) => Promise
...
+test.beforeAll(async ({ browser, browserName }, workerInfo) => {
   const context = await browser.newContext();
   await context.tracing.start({ name: 'test', screenshots: true, snapshots: true });
   const page = await context.newPage();
   await page.goto('data:text/html,<html>Hello world</html>');
   await page.setContent('<button>Click</button>');
+  await page.evaluate(() => {
+    console.log('Info');
+    console.warn('Warning');
+    console.error('Error');
```
mxschmitt, line 93:
```
console.log({ foo: { is: true, window: document.window }})
console.log([1, undefined, null, "foo", new Error("barfoo")])
console.log(() => {})
```

lets add a test for these too? So that they get correctly serialised and then deserialised/shown to the user.

src/web/traceViewer/ui/consoleTab.tsx
```diff
@@ -0,0 +1,87 @@
...
+        if (event.metadata.method === 'pageError')
+          entries.push({ error: event.metadata.params.error });
+      }
+    }
+    return entries;
+  }, [context, action]);
+
+  return <div className='console-tab'>{
+    entries.map((entry, index) => {
+      const { message, error } = entry;
```
mxschmitt, line 51: image

Not sure how much value it brings to have collapsable arrays/objects there - more like nice-to-have.

mxschmitt (approved):

PR #7426, opened 2021-07-01
browser(webkit): roll to 07/01/21

q66:
running tests doesn't show any regressions compared to last roll's status

had to fix build after rebasing because of upstream API changes; my branch showing how I fixed conflicts as well as the build fix here: https://github.com/q66/webkit/tree/playwright

three way diff: https://github.com/q66/WebKit/commit/455e6c0c4f7ae17d9d50e47c332ce840984ed394


browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -1,8 +1,8 @@
 diff --git a/Source/JavaScriptCore/CMakeLists.txt b/Source/JavaScriptCore/CMakeLists.txt
-index 3566ca4859e09aaddaa7f97175a4b12f902780fc..ebab5e27a3dd4760291acaf74a5db117539fa8d8 100644
+index ad391d29fbf3..7361fa249d56 100644
```
pavelfeldman, line 2: I think your git settings are off - we end up with all hashes changing their format.

pavelfeldman:
We only need a single commit like https://github.com/q66/WebKit/commit/455e6c0c4f7ae17d9d50e47c332ce840984ed394 with the three-way merge for rolls, we put it in the PR description!

pavelfeldman (approved):

browser_patches/webkit/patches/bootstrap.diff
q66: ah, i was about to update the patch just as you merged it :) well, i'll use the full format next week

q66:
alright, noted, thanks

browser_patches/webkit/patches/bootstrap.diff
mxschmitt: @q66 which Git version are you using and do you have the option `core.abbrev` in the config set? Curious to find out why it's only using at your machine the short ones.

q66: 2.30.2 (on ubuntu), and it's unset

mxschmitt: are you using the export.sh script? This should automatically give you full-indexes since we set the override there.

(if you are using it, a config dump via `git config --list` would be helpful to debug it further.)

Issue #7283, opened 2021-06-23
[Feature] Add Download.cancel in Firefox and WebKit


mxschmitt:
Relates #7279
Relates #6236

mxschmitt:
@starrify are you interested in collaborating to Playwright for adding Firefox and WebKit Download.cancel support?

starrify:
Hi @mxschmitt, thanks for asking and the answer is surely a yes! :)

Initially I was waiting for the Chromium-only PR (#6236) to be merged for that contained some generic changes.
Since that happened, I have started looking into Firefox a few days ago (was rather unfamiliar with its implement before) and also the way Playwright applies its browser patches. I'm expecting to spend more time on that (major learning, though) however I'm unable to propose a promising timeline.

PR #7432, opened 2021-07-01
browser(webkit): add support to cancel a download

mxschmitt:
~~Seems like the upstream repo (WebKit/WebKit) is too big for some reason and does not allow me to push my branches anymore - trying to figure there something out.~~ -> use WebKit/WebKit-http instead fixes it

Changes: https://github.com/mxschmitt/WebKit-http/commit/93e624d34eaa2098a628a496b34381b730041432

(I'm in contact with Igalia why their hash-length differs on their environment - probably they are not using the export.sh script and doing it manually).

I tested this change on Linux and Mac. Also quite confident that it works on Windows.

Relates #7283

### DIFF
--- a/browser_patches/webkit/patches/bootstrap.diff
+++ b/browser_patches/webkit/patches/bootstrap.diff
@@ -1,5 +1,5 @@
 diff --git a/Source/JavaScriptCore/CMakeLists.txt b/Source/JavaScriptCore/CMakeLists.txt
-index ad391d29fbf3..7361fa249d56 100644
+index ad391d29fbf3dc05e2092b436e2fa412199c3bea..7361fa249d564eac2032a3da771ec147012373ec 100644
 --- a/Source/JavaScriptCore/CMakeLists.txt
 +++ b/Source/JavaScriptCore/CMakeLists.txt
 @@ -1240,22 +1240,27 @@ set(JavaScriptCore_INSPECTOR_DOMAINS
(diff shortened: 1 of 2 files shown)


browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -12362,20 +12369,25 @@ index c61bb3fd2ee0..0fee368e9364 100644
...
  }
  
  DownloadProxy::~DownloadProxy()
-@@ -79,6 +84,8 @@ void DownloadProxy::cancel(CompletionHandler<void(API::Data*)>&& completionHandl
+@@ -75,9 +80,12 @@ static RefPtr<API::Data> createData(const IPC::DataReference& data)
+ void DownloadProxy::cancel(CompletionHandler<void(API::Data*)>&& completionHandler)
+ {
+     if (m_dataStore) {
+-        m_dataStore->networkProcess().sendWithAsyncReply(Messages::NetworkProcess::CancelDownload(m_downloadID), [this, protectedThis = makeRef(*this), completionHandler = WTFMove(completionHandler)] (const IPC::DataReference& resumeData) mutable {
++        auto* instrumentation = m_dataStore->downloadInstrumentation();
```
yury-s, line 12381: Is there a reason the instrumentation cannot be accessed inside the callback as before?

browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -15507,6 +15520,19 @@ index 000000000000..1484aa4a192e
...
 +    m_frontendDispatcher->downloadFinished(uuid, error);
++    m_downloads.remove(uuid);
++}
++
++Inspector::Protocol::ErrorStringOr<void> InspectorPlaywrightAgent::cancelDownload(const String& uuid)
++{
++    if (!m_isEnabled)
++        return { };
++    auto download = m_downloads.get(uuid);
++    if (!download) {
```
yury-s, line 15531: nit: no { } around one liner

yury-s (approved):

browser_patches/webkit/patches/bootstrap.diff
mxschmitt: Inside downloadMapProxy.downloadFinished the m_dataStore reference gets cleared which leads to a accessing a null pointer otherwise.

PR #7436, opened 2021-07-02
feat(test runner): run modifier functions once if they do not depend on test fixtures

dgozman:


pavelfeldman (approved):

tests/playwright-test/test-modifiers.spec.ts
```diff
@@ -214,3 +214,74 @@ test('modifier with a function should throw in the test', async ({ runInlineTest
   expect(result.exitCode).toBe(1);
   expect(result.output).toContain('test.skip() with a function can only be called inside describe block');
 });
+
+test('test.skip with worker fixtures only should skip before hooks and tests', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.test.ts': `
+      const test = pwt.test.extend({
+        bar: 'bar',
```
JoelEinbinder, line 222: bar is unused?

tests/playwright-test/test-modifiers.spec.ts
```diff
@@ -214,3 +214,74 @@ test('modifier with a function should throw in the test', async ({ runInlineTest
...
+      });
+      const logs = [];
+      test.beforeEach(() => {
+        console.log('\\n%%beforeEach');
+      });
+      test('passed', () => {
+        console.log('\\n%%passed');
+      });
+      test.describe('suite1', () => {
+        test.skip(({ foo }) => foo === 'foo', 'reason');
```
JoelEinbinder, line 233: I'd expect a log here. How are we testing that this is run only once? I assume that the old behavior had the beforeEach running for each skip? I didn't realize that beforeEach ran before skips in the first place.

tests/playwright-test/test-modifiers.spec.ts
```diff
@@ -214,3 +214,74 @@ test('modifier with a function should throw in the test', async ({ runInlineTest
...
+        test.beforeAll(() => {
```
JoelEinbinder, line 234: is beforeAll supposed to run before or after the modifier. Couldn't this beforeAll modify `foo` to change how its behavior in `skip`.

tests/playwright-test/test-modifiers.spec.ts
dgozman: Yep!

tests/playwright-test/test-modifiers.spec.ts
dgozman: The old behavior was to run skip for each test, now we only run it once when possible, before any beforeAll.

tests/playwright-test/test-modifiers.spec.ts
dgozman: In theory it could, but we run the skip before beforeAll. This way we can avoid hooks entirely when `skip` says so.

PR #7457, opened 2021-07-04
fix(recorder): Fix recorder out of view

### LINKED ISSUE #5696: [BUG] Playwright Inspector window is opened outside of visible screen
**Context:**
- Playwright Version: 1.9.0
- Operating System: Windows10 
- Node.js version: v12.18.2
- Browser: Chromium 
- Extra: laptop with 13,3" screen, 1920 x 1080 (FullHD) resolution, scaled to 150%, setup with external monitor placed above laptop screen

Example script 
```javascript
(async () => {
})().then(() => console.log('finished'));
```

When trying to debug script (set PWDEBUG=1) Playwright Inspector window is opened by Playwright outside of visible screen space - it is opened past the right border of the screen.

in 
src\server\supplements\recorder\recorderApp.ts RecorderApp.open
there is a definitions of arguments for opening the window
const args = [
      '--app=data:text/html,',
      '--window-size=600,600',
      '--window-position=1280,10',
    ];
when 1280 is changed to say 700 then Playwright Inspector windows is visible after it is first opened by Playwright.

kumaraditya303:
Fixes https://github.com/microsoft/playwright/issues/5696

### DIFF
--- a/src/server/supplements/recorder/recorderApp.ts
+++ b/src/server/supplements/recorder/recorderApp.ts
@@ -89,7 +89,6 @@ export class RecorderApp extends EventEmitter {
     const args = [
       '--app=data:text/html,',
       '--window-size=600,600',
-      '--window-position=1280,10',
     ];
     if (process.env.PWTEST_RECORDER_PORT)
       args.push(`--remote-debugging-port=${process.env.PWTEST_RECORDER_PORT}`);


pavelfeldman:
This is hardly a fix, you fix those with logical screens < 1280, but regress experience for everyone.

kumaraditya303:
Even in trace viewer the window position is not set and it works fine  https://github.com/microsoft/playwright/blob/f40611306e950bf18071c5c4d810510d90fd5937/src/server/trace/viewer/traceViewer.ts#L132

pavelfeldman:
It'll be showing on top of the target window, which would be a major annoyance...

kumaraditya303:
> It'll be showing on top of the target window, which would be a major annoyance...

Currently on windows it is just not visible, with this patch it shows, do you have any other fix for this else I think something is better than nothing

mxschmitt:
Seems stale, closing as part of the triage process.

PR #7469, opened 2021-07-05
feat: Connect support for csharp

Archish27:


docs/src/api/class-browsertype.md
```diff
@@ -85,7 +85,7 @@ class BrowserTypeExamples
 ```
 
 ## async method: BrowserType.connect
-* langs: js, python, java
+* langs: js, python, java, csharp
```
mxschmitt, line 88: Instead of doing this, you should be able to drop this line completely (and below) which results in having support for connect in all the languages.

mxschmitt (approved):

docs/src/api/class-browsertype.md
Archish27: Boom 🔥 will make a change

Archish27: done

mxschmitt: needs to be changed also in the other places in your PR.

Archish27: I think not required since it doesn't include all languages only java, python, csharp

mxschmitt (changes requested):
Please add also the following:

- option: BrowserType.connect.slowMo
- option: BrowserType.connect.timeout

Archish27:
> Please add also the following:
> 
> * option: BrowserType.connect.slowMo
> * option: BrowserType.connect.timeout

Currently, this is not implemented. I don't think we need to expose as of now since this won't work. Gradually when implementing we can enable these properties.

docs/src/api/class-browsertype.md
```diff
@@ -102,13 +101,13 @@ This methods attaches Playwright to an existing browser instance.
...
 ### param: BrowserType.connect.wsEndpoint
-* langs: java, python
+* langs: java, python, csharp
 - `wsEndpoint` <[string]>
 
 A browser websocket endpoint to connect to.
 
 ### option: BrowserType.connect.headers
-* langs: java, python
+* langs: java, python, csharp
```
pavelfeldman, line 110: What's up with BrowserType.connect.slowMo & BrowserType.connect.timeout?

Archish27: This is not implemented as of now so was suggesting we can add this later when we make changes to playwright dotnet. Will implement this feature in follow-up PR.  So,  was thinking when we start developing only then we expose. Right now I don’t see a need of doing it since this is not being used.

Archish27: What are your thoughts? Should I change it?

mxschmitt: Yes, we only ship full APIs, so having support for slowMo and connection timeout needs to be implemented before shipping it.

Archish27: Updated

mxschmitt:
This got implemented in 99c95648108e547d7b8c708dc43c4a6e9d4977c6 and 7b861dcee63e14f8283a19dbbd3e64ad16db429d. Thank you for your contribution to the project!

PR #7471, opened 2021-07-06
docs: promote cli command to install chrome stable

yury-s:
https://github.com/microsoft/playwright-java/issues/507

### DIFF
--- a/docs/src/browsers.md
+++ b/docs/src/browsers.md
@@ -82,10 +82,7 @@ class Program
 }
 ```
 
-:::note
-Playwright bundles a recent Chromium build, but not Google Chrome or Microsoft Edge browsers - these should be installed manually before use.
-
-You can use Playwright command line interface to install the browsers:
+Note that Playwright bundles a recent Chromium build, but not Google Chrome or Microsoft Edge browsers - these should be installed manually before use. You can use Playwright command line interface to install the browsers:
 
 ```bash js
 # Install Chrome stable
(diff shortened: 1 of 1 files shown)


docs/src/browsers.md
```diff
@@ -82,9 +82,27 @@ class Program
 }
 ```
 
-:::note
-Playwright bundles a recent Chromium build, but not Google Chrome or Microsoft Edge browsers - these should be installed manually before use.
-:::
+Note that Playwright bundles a recent Chromium build, but not Google Chrome or Microsoft Edge browsers - these should be installed manually before use. You can use Playwright command line interface to install the browsers:
```
mxschmitt, line 85: Would be good to reference there to the dedicated CLI page like we do in the other places. (Search for command line)

docs/src/browsers.md
```diff
@@ -82,9 +82,27 @@ class Program
...
+
+```bash js
+# Install Chrome stable
```
mxschmitt, line 88: Looks similar to https://github.com/microsoft/playwright/blob/master/docs/src/cli.md#install-browsers which got recently added. Might make sense to dedup/reference?

yury-s: Dunno, having it inline is convenient but let's keep a link for now.

docs/src/browsers.md
yury-s: done

mxschmitt (approved):

dgozman (approved):

PR #7480, opened 2021-07-06
chore: replace downloadBrowser with downloadAndExtractZip utility

dgozman:


src/utils/registry.ts
```diff
@@ -248,38 +245,46 @@ export function isBrowserDirectory(browserDirectory: string): boolean {
...
-let currentPackageRegistry: Registry | undefined = undefined;
+type BrowserDescriptor = {
+  name: BrowserName,
+  revision: string,
+  installByDefault: boolean,
+  browserDirectory: string,
+};
+
+function readDescriptors(packagePath: string) {
+  // require() needs to be used thhereere otherwise it breaks on Vercel serverless
```
mxschmitt, line 256: nit: this comment was wrong, we can remove it.

Issue #5777, opened 2021-03-10
[Feature] IME support


JoelEinbinder:
Right now our [IME](https://en.wikipedia.org/wiki/Input_method) support is limited. We insert text, but do not allow for testing complicated scenarios with composition events.

The Web api: https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent
The Mac api: https://developer.apple.com/documentation/appkit/nstextinputclient?language=objc
The Chromium mojo api: https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/public/mojom/input/input_handler.mojom

This request comes from @trueadm on twitter: https://twitter.com/trueadm/status/1369332999242792967

My initial api proposal:

```js
await page.keyboard.startComposition(start, end); // optional start and end
await page.keyboard.updateComposition('foo');
await page.keyboard.updateComposition('bar');
await page.keyboard.press('A'); // what should this do? Send keyboard events without generating text?
await page.keyboard.endComposition('baz'); // optional ending text
```

The only time I've seen composition events used in the wild is in CodeMirror to draw a synthetic composition highlight. It would be helpful to have some better use cases here to make sure that the api handles them without growing too large.

- Does the composition state need to be inspected, or just controlled? That is, do we need `keyboard.compositionRange()`?
- Do we instead need a high level composition api. Like `keyboard.typeWithComposition("è")` types AltGr+\` followed by e?
- Do we need be able to trigger composition via keys. Right now our keyboard api largely ignores dead keys that would start composition.

trueadm:
> It doesn't. It also doesn't fire composition start/end, but rather does a text replacement for the character before the cursor. The web page sees the initial keydown/input, a bunch of repeated key downs without text as the e key is held, a keydown for the digit 1 to select the variation, and then an input event with the new text.
Is the complexity here important? Right now we would just fire the final input event with è. This is similar to what you'd get on an iOS keyboard.

> Can you give an example of where a component breaks with IME? You have to be doing something a little bit strange for composition events to be relevant, so I want to make sure I understand what we are testing for.

If I hold down `e` the IME appears, if I press RightArrow, it enters composition. This is an important bug, because this is broken in DraftJS right now (see FB.com or Twitter). We've fixed it in a new editor we're working on, but bugs like this are why we need good IME regression tests.

[...]

PR #7487, opened 2021-07-06
(feat): Composition support with keyboard.imeSetComposition, keyboard.imeCancelComposition and keyboard.imeCommitComposition

t-javierc:
Implemented composition support.

Related feature thread: https://github.com/microsoft/playwright/issues/5777

Chromium changes are currently in review: https://chromium-review.googlesource.com/c/chromium/src/+/3009937


JoelEinbinder:
Thanks for the PR!

The api here is going to take some work to get right. Aside from nits like camel casing and api names, we need to have a better sense of how user tests should look. I want to reduce the size of this api to make it easier for us to manage and for users to use correctly. For example I don't think we need `delay`.

The point of `triggering_key` seems to be to send a keypress without the associated text. Maybe we can leave this out of the api for now if its not essential? Or if it is essential, can we help the users with what key they should actually put here?

Can you send some examples of real world tests that use this api? If we can't write a test that uses selection/delay etc, lets leave that out of the api for now.

@dgozman

JoelEinbinder:
Let's remove the `triggering_key` part of this patch in favor of [this patch](https://github.com/microsoft/playwright/pull/7659) which does options for `keyboard.down`. In the OS apis for IME the keyboard events and IME events are not associated, so we don't need to associate them here. Anyone using this api is doing something pretty advanced, so we can prioritize correctness over convenience.

dgozman:
Here is a proposed API draft. Assumes #7659.

```js
// Compose
await page.keyboard.down('s', { eventKey: 's', text: '' });  // Optional
await page.keyboard.up('s', { eventKey: 's' });  // Optional
await page.keyboard.imeSetComposition({ text: 'ｓ', selectionStart: 1, selectionEnd: 1 });
await page.keyboard.down('u', { eventKey: 'す', text: '' });  // Optional
await page.keyboard.up('u', { eventKey: 'す' });  // Optional
await page.keyboard.imeSetComposition({ text: 'す', selectionStart: 1, selectionEnd: 1, replacementStart: 0, replacementEnd: 1 });

// Commit
await page.keyboard.insertText('す');

// Cancel
await page.keyboard.down('Escape');  // Optional
await page.keyboard.up('Escape');  // Optional
await page.keyboard.insertText('');
```

Overall:
- Drop `triggering_key` in favor of `keybpard.down` options.
- Remove `delay` since we don't have usecases for it.
- Use `keyboard.insertText()` to commit/cancel.

trueadm:
@JoelEinbinder has there been any more traction on this? I was wondering if this feature was still under development.

trueadm:
I followed up with a patched version of this PR, rebased with changes from main. https://github.com/microsoft/playwright/pull/9535

trueadm:
@alexkeng Thanks for letting me know. It would be great if we could maybe chat and talk about the design and also this as a general feature. We're building a new text editor at Facebook (we hope to open source it soon) that already powers many parts of Facebook. IME automated testing is going to play a very important role and it would be awesome if we could sync up. My email is domgan [at] fb.com if you wanted reach out.

aslushnikov:
Closing this in favor of https://github.com/microsoft/playwright/pull/9535

Issue #1774, opened 2020-04-14
[Question] Intercepting responses


Insensitivity:
Hey, I was wondering whether it is possible to intercept the response to a network request. I did notice the possibility of using routes to intercept the requests, but is it possible to get the response without re-sending the request manually? If so how? and would it work for HTTP data streams?

I've skimmed through the API docs but couldn't really find any specifics regarding intercepting the response.

pavelfeldman:
Response interception is not currently available. This is something we can add in the future - no fundamental problems with having it. Could you share more about your use case?

hollligan:
it would be useful for mock.
puppeteer example:

```
// Intercept API response and pass mock data for Puppeteer
await page.setRequestInterception(true);
page.on('request', request => {
    if (request.url() === constants.API) {
        request.respond({
            content: 'application/json',
            headers: {"Access-Control-Allow-Origin": "*"},
            body: JSON.stringify(constants.biddersMock)
        });
    }
    else {
        request.continue();
    }
});
```

Nisthar:
@hollligan does that code work in puppeteer?

hollligan:
@Nisthar I didn't check it, I took it from the puppeteer documentation. for a playwright, here's the working code -
[https://playwright.dev/#version=v1.4.2&path=docs%2Fnetwork.md&q=handle-requests](https://playwright.dev/#version=v1.4.2&path=docs%2Fnetwork.md&q=handle-requests)

lironzizva:
It will be useful to troubleshoot scenario that pages are not fully loaded while running multiple tests in parallel in headless mode.

emil45:
Is it still not implemented ?

onno-amsterdam:
Can't you just use: page.waitForResponse? I'm able to intercept responses for XHR requests (and parse it to JSON) like so:
const response = await page.waitForResponse('**/my-endpoint');
JSON.parse(response.body().toString());

ua2028:
@onno-amsterdam correct me if I'm wrong but  response = await page.waitForResponse will return the finalized response, while the question refers to modifying the response prior to it's consumption by the web app.

biiishal:
This would definitely be helpful to use fixtures for repeating network requests. ⬆️ upvote from me.

IluhaSk:
@pavelfeldman 
is this feature planned in next release?

ThompsonJonM:
Playwright allows for route interception and mocking/stubbing.

```
const mock = { books: [] }

[...]

PR #7509, opened 2021-07-08
browser(firefox): support response interception

yury-s:
Corresponding client side changes: #7510

#1774

### DIFF
--- a/browser_patches/firefox/juggler/NetworkObserver.js
+++ b/browser_patches/firefox/juggler/NetworkObserver.js
@@ -73,8 +73,15 @@ class PageNetwork {
     this._interceptedRequests.clear();
   }
 
-  resumeInterceptedRequest(requestId, url, method, headers, postData) {
+  async resumeInterceptedRequest(requestId, url, method, headers, postData, interceptResponse) {
+    if (interceptResponse) {
+      const intercepted = this._interceptedRequests.get(requestId);
+      if (!intercepted)
+        throw new Error(`Cannot find request "${requestId}"`);
+      return { response: await intercepted.interceptResponse(url, method, headers, postData) };
+    }
     this._takeIntercepted(requestId).resume(url, method, headers, postData);
+    return {};
   }
 
   fulfillInterceptedRequest(requestId, status, statusText, headers, base64body) {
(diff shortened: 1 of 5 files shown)


browser_patches/firefox/juggler/NetworkObserver.js
```diff
@@ -180,6 +187,77 @@ class NetworkRequest {
...
+          stream.close();
+        }
+      });
+    });
+
+    const pageNetwork = this._activePageNetwork();
+    if (pageNetwork)
+      pageNetwork._responseStorage.addResponseBody(this, newChannel, body);
+
+    const responseHeaders = [];
```
dgozman, line 243: Do you want a helper method for lines 243-256? I think we have a copy of that :)

browser_patches/firefox/juggler/NetworkObserver.js
```diff
@@ -684,6 +752,8 @@ class NetworkObserver {
   _onRequest(channel, topic) {
     if (!(channel instanceof Ci.nsIHttpChannel))
       return;
+    if (this._responseInterceptionChannels.has(channel))
```
dgozman, line 755: I think you should also ignore this in more places, e.g. `_onRedirect`or `_onResponse`? Perhaps we need a test for intercepted response after a redirect?

dgozman (approved):

browser_patches/firefox/juggler/NetworkObserver.js
yury-s: `_onResponse` will ignore the channel since it is not in `this._channelToRequest` map (skipped in `_onRequest`, the redirects chain maybe more tricky though, I'll add a test.

browser_patches/firefox/juggler/NetworkObserver.js
yury-s: Yeah, added.

browser_patches/firefox/juggler/NetworkObserver.js
yury-s: Response interception is broken one way or the other in all 3 browsers in case of redirects so I'll address it separately.

Issue #7496, opened 2021-07-07
[Bug]: A new worker gets created when skip is used.


mxschmitt:
### Playwright version

1.12.0

### Operating system

Windows, MacOS, Linux

### What happened? / Describe the bug

> I was playing around with playwright test runner and test.fixme and test.skip annotation. I’ve noticed that using skip or fixme result in creating a new worker to run remaining tests after the skipped one. Is there any technical reason for doing that ? Here’s a screen of a sample test run

![image](https://user-images.githubusercontent.com/17984549/124747471-3e2dc080-df22-11eb-902a-b4f9088debfe.png)

From Slack: https://playwright.slack.com/archives/C01AG7QGW2F/p1625649237089500

pavelfeldman:
We restart workers on exceptions, fixme is internally implemented as exception, but we surely don't want to restart workers on them.

mxschmitt:
Closed since fixed in https://github.com/microsoft/playwright/pull/7511.

PR #7511, opened 2021-07-08
fix(runner): do not restart worker after skipping tests

yury-s:
#7496

### DIFF
--- a/src/test/workerRunner.ts
+++ b/src/test/workerRunner.ts
@@ -336,7 +336,7 @@ export class WorkerRunner extends EventEmitter {
     if (!preserveOutput)
       await removeFolderAsync(testInfo.outputDir).catch(e => {});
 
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):
bonus points: add a test for test.fail() and maybe a `throw new Error()` test case which asserts a new worker ID since it seems we don't have one yet.

dgozman (approved):

yury-s:
> bonus points: add a test for test.fail() and maybe a `throw new Error()` test case which asserts a new worker ID since it seems we don't have one yet.

Will add them in another pr.

dgozman:
> bonus points: add a test for test.fail() and maybe a `throw new Error()` test case which asserts a new worker ID since it seems we don't have one yet.

I am pretty sure we have multiple such tests :)

PR #7517, opened 2021-07-08
chore(eval): merge internal evaluate functions

JoelEinbinder:
Our internal evaluations were getting pretty ugly with so many methods and arguments. I tried to consolidate
most of them into a method eval﻿that takes an options object with default values.

I think there is room to do more with some waitForSignals methods, but those have a strange relationship
with slowmo that I want to figure out in a separate patch.

Re-opening of #6769﻿

### DIFF
--- a/src/server/dom.ts
+++ b/src/server/dom.ts
@@ -50,36 +50,6 @@ export class FrameExecutionContext extends js.ExecutionContext {
     return null;
   }
 
-  async evaluate<Arg, R>(pageFunction: js.Func1<Arg, R>, arg?: Arg): Promise<R> {
-    return js.evaluate(this, true /* returnByValue */, pageFunction, arg);
-  }
-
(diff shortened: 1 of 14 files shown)


src/server/dom.ts
```diff
@@ -135,7 +105,7 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
 
   private async _evaluateInMainAndWaitForSignals<R, Arg>(pageFunction: js.Func1<[js.JSHandle<InjectedScript>, ElementHandle<T>, Arg], R>, arg: Arg): Promise<R> {
     const main = await this._context.frame._mainContext();
-    return main.evaluateAndWaitForSignals(pageFunction, [await main.injectedScript(), this, arg]);
+    return main.eval(pageFunction, {arg: [await main.injectedScript(), this, arg], waitForSignals: true});
```
yury-s, line 108: This `arg: ` field containing an array of arguments looks odd now that we have `args:` option, consider using that one instead.

yury-s (approved):

PR #7528, opened 2021-07-09
docs(test-advanced.md): Overriding fixtures

### LINKED ISSUE #7512: [Feature]: Document "fixture enhancement"
### Feature request

As I've began to utilize the Playwright test runner, I discovered that it supports overriding/enhancing existing fixtures.  After more digging, I found other examples in issues including this example shared by @mxschmitt.

https://github.com/microsoft/playwright/issues/7051#issuecomment-859916019

```js
import { test as baseTest } from '@playwright/test';

const test = baseTest.extend({
  page: async ({ page }, use) => {
    await page.goto('http://localhost:3579');
    await use(page);
  },
});

export const it = test;
export const expect = test.expect;
```

This is a very powerful pattern as you can add functionality in layers by adding more extends on base fixtures based on various use cases a test suite might have.  Could we add documentation for this type of pattern somewhere in the docs?

I'd be more than happy to write the docs if you are all good with adding it.  Just point me in the direction of where to go!

mskelton:
Fixes #7512 

Adds a section to the advanced fixtures documentation explaining fixture augmentation.  I waffled quite a bit on what to call this, whether it should be fixture enhancement, fixture augmentation, or something else altogether.

### DIFF
--- a/docs/src/test-fixtures.md
+++ b/docs/src/test-fixtures.md
@@ -344,7 +344,6 @@ In addition to creating your own fixtures, you can also override existing fixtur
 const base = require('@playwright/test');
 
 exports.test = base.test.extend({
-  baseURL: 'http://localhost:8080',
   page: async ({ baseURL, page }, use) => {
     await page.goto(baseURL);
     await use(page);
@@ -356,7 +355,6 @@ exports.test = base.test.extend({
 import { test as base } from '@playwright/test';
 
 export const test = base.extend({
-  baseURL: 'http://localhost:8080',
   page: async ({ baseURL, page }, use) => {
     await page.goto(baseURL);
     await use(page);
@@ -364,7 +362,7 @@ export const test = base.extend({
 });
 ```
 
-Notice that in this example we added another fixture named `baseURL` which the `page` fixture uses. This allows us to override the `baseURL` used by the `page` fixture in our tests using `test.use`.
+Notice that in this example, the `page` fixture is able to depend on other built in fixtures such as `baseURL`. This allows us to override the `baseURL` used by the `page` fixture in our tests using `test.use`.
 
 ```js js-flavor=js
 test.use({ baseURL: 'https://playwright.dev' })
--- a/docs/src/test-fixtures.md
+++ b/docs/src/test-fixtures.md
@@ -336,40 +336,54 @@ const test = base.extend<{}, ExpressWorkerFixtures>({
 export default test;
 ```
 
-## Augmenting fixtures
+## Overriding fixtures
 
-In addition to creating your own fixtures, you can also augment existing fixtures to fit your needs. Consider the following example which augments the `page` fixture by navigating to a specified URL:
+In addition to creating your own fixtures, you can also override existing fixtures to fit your needs. Consider the following example which overrides the `page` fixture by navigating to a specified URL:
 
 ```js js-flavor=js
-// button.spec.js
 const base = require('@playwright/test');
 
 const test = base.test.extend({
-  page: async ({ page }, use) => {
-    await page.goto('http://localhost:8080');
+  baseURL: 'http://localhost:8080',
+  page: async ({ baseURL, page }, use) => {
+    await page.goto(baseURL);
     await use(page);
   },
-});
-
-test('should click the button', async ({ page }) => {
-  await page.click('button')
-  // ...
 });
 ```
 
 ```js js-flavor=ts
-// button.spec.ts
 import { test as base } from '@playwright/test';
 
 const test = base.extend({
-  page: async ({ page }, use) => {
-    await page.goto('http://localhost:8080');
+  baseURL: 'http://localhost:8080',
+  page: async ({ baseURL, page }, use) => {
+    await page.goto(baseURL);
     await use(page);
   },
 });
+```
 
-test('should click the button', async ({ page }) => {
-  await page.click('button')
-  // ...
+Notice that in this example we added another fixture named `baseURL` which the `page` fixture uses. This allows us to override the `baseURL` used by the `page` fixture in our tests using `test.use`.
+
+```js
+test.use({ baseURL: 'https://playwright.dev' })
+```
+
+Fixtures can also be overridden where the base fixture is completely replaced with something different. For example, we could override the `storageState` fixture to provide our own data.
+
+```js js-flavor=js
+const base = require('@playwright/test');
+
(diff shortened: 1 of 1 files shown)


docs/src/test-fixtures.md
```diff
@@ -335,3 +335,41 @@ const test = base.extend<{}, ExpressWorkerFixtures>({
 
 export default test;
 ```
+
+## Augmenting fixtures
```
mxschmitt, line 339: I'm not a native english speaker, and think we should use probably a different word for that. I would propose one of the following ones:

- overwrite
- redefine

Maybe you have a better idea.

mxschmitt (approved):

docs/src/test-fixtures.md
mskelton: Overwrite was one of the words I considered, the only part I didn't like was the fact that it gives the idea of replacing the existing fixture rather than layering logic on top of the existing fixture.  That might be overcomplicating it though and overwrite might be the best term 🙂

mxschmitt: Inheritance might be also a good candidate.

kumaraditya303: How about extending fixtures ?

mskelton: I kind of like extending, but wonder if that is confusing given `test.extend` 😕 

Inheritance might work

dgozman: Let's use the "override" word, because you are essentially overriding the fixture with your own. Let's also mention (and maybe give an example):
- you can depend on other fixtures (good example would be `page` fixture that depends on `baseURL`);
- you can depend on the same fixture (as you do already in this example) to "extend" it;
- you can ignore the base fixture and return something entirely different (good example would be overriding the `storageState` fixture and provide your own data).

mskelton: Awesome, thanks for the info.  I'll go ahead and update this PR tomorrow to use override as well as add the other examples you mentioned.

mskelton: I updated the docs to what I think covers the points you made.  Please take a look and see if it makes sense.

docs/src/test-fixtures.md
```diff
@@ -335,3 +335,55 @@ const test = base.extend<{}, ExpressWorkerFixtures>({
...
+test.use({ baseURL: 'https://playwright.dev' })
+```
+
+Fixtures can also be overridden where the base fixture is completely replaced with something different. For example, we could override the `storageState` fixture to provide our own data.
+
+```js js-flavor=js
+const base = require('@playwright/test');
+
+const test = base.test.extend({
+  storageState: { path: 'state.json' },
```
mskelton, line 379: @dgozman @mxschmitt I'm not really familiar with `storageState` or what @dgozman was suggesting in his comment, so if this is a poor example, let me know how it should be improved.

docs/src/test-fixtures.md
```diff
@@ -335,3 +335,55 @@ const test = base.extend<{}, ExpressWorkerFixtures>({
...
+import { test as base } from '@playwright/test';
+
+const test = base.extend({
+  baseURL: 'http://localhost:8080',
+  page: async ({ baseURL, page }, use) => {
+    await page.goto(baseURL);
+    await use(page);
+  },
+});
+```
```
dgozman, line 365: `export default test;`

docs/src/test-fixtures.md
```diff
@@ -335,3 +335,55 @@ const test = base.extend<{}, ExpressWorkerFixtures>({
...
 ```
+
+## Overriding fixtures
+
+In addition to creating your own fixtures, you can also override existing fixtures to fit your needs. Consider the following example which overrides the `page` fixture by navigating to a specified URL:
+
+```js js-flavor=js
+const base = require('@playwright/test');
+
+const test = base.test.extend({
```
dgozman, line 346: `exports.test = ...`

docs/src/test-fixtures.md
```diff
@@ -335,3 +335,55 @@ const test = base.extend<{}, ExpressWorkerFixtures>({
...
+const test = base.extend({
+  baseURL: 'http://localhost:8080',
+  page: async ({ baseURL, page }, use) => {
+    await page.goto(baseURL);
+    await use(page);
+  },
+});
+```
+
+Notice that in this example we added another fixture named `baseURL` which the `page` fixture uses. This allows us to override the `baseURL` used by the `page` fixture in our tests using `test.use`.
```
dgozman, line 367: This is very nice,  but we have just added the built-in baseURL fixture, so you don't have to define it in the `extend()` 😄

docs/src/test-fixtures.md
dgozman: I meant something like this:

```js
const test = base.test.extend({
  storageState: async ({}, use) => {
    const cookie = await getAuthCookie();
    await use({ cookies: [cookie] });
  }
});
```

docs/src/test-fixtures.md
```diff
@@ -335,3 +335,55 @@ const test = base.extend<{}, ExpressWorkerFixtures>({
...
+```js
+test.use({ baseURL: 'https://playwright.dev' })
+```
+
+Fixtures can also be overridden where the base fixture is completely replaced with something different. For example, we could override the `storageState` fixture to provide our own data.
+
+```js js-flavor=js
+const base = require('@playwright/test');
+
+const test = base.test.extend({
```
dgozman, line 378: Missing export here and below.

docs/src/test-fixtures.md
```diff
@@ -335,3 +335,55 @@ const test = base.extend<{}, ExpressWorkerFixtures>({
...
+  page: async ({ baseURL, page }, use) => {
+    await page.goto(baseURL);
+    await use(page);
+  },
+});
+```
+
+Notice that in this example we added another fixture named `baseURL` which the `page` fixture uses. This allows us to override the `baseURL` used by the `page` fixture in our tests using `test.use`.
+
+```js
```
dgozman, line 369: Please copy this snippet for js and ts - that makes nicer docs.

dgozman (commented):
Very nice! Just a few comments.

mskelton:
@dgozman Updated per your comments.  Thanks for all the great feedback, I'm learning a lot just by writing these docs!

dgozman (approved):
Looks great, merging in.

PR #7531, opened 2021-07-09
feat: add High DPI and Low DPI Desktop device descriptors

### LINKED ISSUE #7406: [Feature] Support MacBook Pro, Surface etc. device type

mxschmitt:
Fixes #7406

### DIFF
--- a/src/server/deviceDescriptorsSource.json
+++ b/src/server/deviceDescriptorsSource.json
@@ -1157,7 +1157,7 @@
     "defaultBrowserType": "chromium"
   },
   "Desktop Edge HiDPI": {
-    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/93.0.4543.0 Safari/537.36 Edg/91.0.864.37",
+    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/93.0.4543.0 Safari/537.36 Edg/93.0.4543.0",
     "screen": {
       "width": 1792,
       "height": 1120
@@ -1172,7 +1172,7 @@
     "defaultBrowserType": "chromium"
   },
   "Desktop Firefox HiDPI": {
-    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:89.0) Gecko/20100101 Firefox/89.0",
+    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:90.0 Gecko/20100101 Firefox/89.0",
     "screen": {
       "width": 1792,
       "height": 1120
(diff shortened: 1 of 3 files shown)


yury-s (approved):

src/server/deviceDescriptorsSource.json
```diff
@@ -1140,5 +1140,95 @@
...
+    "viewport": {
+      "width": 1280,
+      "height": 720
+    },
+    "deviceScaleFactor": 2,
+    "isMobile": false,
+    "hasTouch": false,
+    "defaultBrowserType": "webkit"
+  },
+  "MacBook Pro 2019 16-inch": {
```
dgozman, line 1219: I think this does not solve the problem of running tests with "MacBook Chrome", "Surface Chrome", "MacBook Safari", "Surface Firefox", "Android Chrome" and "iPhone Safari". Any ideas?

mxschmitt: When using the test-runner with a device, we automatically use the correct one via the `defaultBrowserType` deviceDescriptor property and pass it over to the browserName. Also in theory we should not prevent users from testing the devices in other browsers, or do I miss something?

pavelfeldman:
I think Dima means that some macOS users (especially web developers) prefer Chrome.

Given that the viewport is the same for all these, we are forcing the user to make a choice that won't actually affect anything. Looking at the https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide, our default viewport still makes the most sense, so all we need to do is to pick Hi DPI vs regular notebook.

Another issue that you are introducing here is stale user agents. I'd suggest 1) making user agents live 2) only offering "Desktop Chrome", "HiDPI Desktop Chrome" for Chrome, Safari, Edge and Firefox.

src/server/deviceDescriptorsSource.json
```diff
@@ -1140,5 +1140,125 @@
     "isMobile": true,
     "hasTouch": true,
     "defaultBrowserType": "chromium"
+  },
+  "High DPI Notebook Google Chrome": {
```
pavelfeldman, line 1144: There is a professional slang for mobile browsers vs the desktop browsers (https://gs.statcounter.com/platform-market-share/desktop-mobile-tablet).

`Desktop Safari` - always high dpi
`Desktop Chrome`
`Desktop Edge`
`Desktop Firefox`
`Desktop Chrome HiDPI`
`Desktop Edge HiDPI`
`Desktop Firefox HiDPI`

mxschmitt: implemented it like that. Choose Windows as a OS, since it has the biggest market share. For Safari I kept MacOS.

src/server/deviceDescriptorsSource.json
```diff
@@ -1140,5 +1140,110 @@
...
+      "width": 1280,
+      "height": 720
+    },
+    "deviceScaleFactor": 2,
+    "isMobile": false,
+    "hasTouch": false,
+    "defaultBrowserType": "chromium"
+  },
+  "Desktop Edge HiDPI": {
+    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/93.0.4543.0 Safari/537.36 Edg/91.0.864.37",
```
pavelfeldman, line 1160: `Chrome/93.0.4543.0 Safari/537.36 Edg/91.0.864.37` - is that expected?

src/server/deviceDescriptorsSource.json
```diff
@@ -1140,5 +1140,110 @@
...
+      "width": 1280,
+      "height": 720
+    },
+    "deviceScaleFactor": 1,
+    "isMobile": false,
+    "hasTouch": false,
+    "defaultBrowserType": "chromium"
+  },
+  "Dekstop Edge": {
+    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/93.0.4543.0 Safari/537.36 Edg/91.0.864.37",
```
pavelfeldman, line 1220: ditto

pavelfeldman (approved):

PR #7535, opened 2021-07-09
Revert "chore(eval): merge internal evaluate functions (#7517)"

mxschmitt:
This reverts commit 7a5ef0d157d57b9ebb04f8387956f1ef365badc2.

This broke the following which was working in Python:

```py
import asyncio
from playwright.async_api import async_playwright

async def main():
    async with async_playwright() as p:
        browser = await p.chromium.launch()
        page = await browser.new_page()
        await page.goto("https://example.com")

        await page.route("**/", lambda route, _: route.fulfill(body="OK"))

        response = await page.evaluate("""async () => (await fetch("/")).text()""")
        assert response == "OK"
        print(response)

        await browser.close()

asyncio.run(main())
```

protocol logs:

```
SEND> {
  "id": 6,
  "guid": "frame@a28ad22eb2d5bcc31acd72118c816774",
  "method": "evaluateExpression",
  "params": {
    "expression": "async () => (await fetch(\"/\")).text()",
    "arg": {
      "value": {
        "v": "null"
      },
      "handles": []
    }
  },
---

RECV> {
  "id": 6,
  "result": {
    "value": {
      "v": "undefined"
    }
  }
}
```

### DIFF
--- a/src/dispatchers/frameDispatcher.ts
+++ b/src/dispatchers/frameDispatcher.ts
@@ -67,22 +67,11 @@ export class FrameDispatcher extends Dispatcher<Frame, channels.FrameInitializer
   }
 
   async evaluateExpression(params: channels.FrameEvaluateExpressionParams, metadata: CallMetadata): Promise<channels.FrameEvaluateExpressionResult> {
-    return { value: serializeResult(await this._frame.eval(params.expression, {
-      arg: parseArgument(params.arg),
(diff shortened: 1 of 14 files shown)


mxschmitt:
LGTM from Yury.

yury-s (commented):
lgtm!

Issue #7452, opened 2021-07-03
[BUG] ERR_REQUIRE_ESM


bennypowers:
**Context:**
- Playwright Version: `"@playwright/test": "1.12.3"`
- Operating System: [Mac]
- Node.js version: [16.3.0]
- Browser: [n/a]

```
## System:
 - OS: macOS Mojave 10.14.6
 - Memory: 5.00 GB / 32.00 GB
## Binaries:
 - Node: 16.3.0 - ~/.nvm/versions/node/v16.3.0/bin/node
 - Yarn: 1.22.10 - ~/.nvm/versions/node/v15.8.0/bin/yarn
 - npm: 7.15.1 - ~/.nvm/versions/node/v16.3.0/bin/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash
 ```

**Code Snippet**

```
$ yarn playwright test presets/rocket-preset-code-tabs/test/code-tabs.test.js    830ms
yarn run v1.22.10
$ /a/b/c/rocket-presets/node_modules/.bin/playwright test presets/rocket-preset-code-tabs/test/code-tabs.test.js
Error [ERR_REQUIRE_ESM]: Error while reading /a/b/c/rocket-presets/presets/rocket-preset-code-tabs/test/code-tabs.test.js:
Must use import to load ES Module: /a/b/c/rocket-presets/presets/rocket-preset-code-tabs/test/code-tabs.test.js
require() of ES modules is not supported.
require() of /a/b/c/rocket-presets/presets/rocket-preset-code-tabs/test/code-tabs.test.js from /a/c/b/rocket-presets/node_modules/@playwright/test/lib/test/loader.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename code-tabs.test.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /a/b/c/rocket-presets/presets/rocket-preset-code-tabs/package.json.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**Describe the bug**

Thanks for publishing this package, it's a joy to use.

I've got a yarn monorepo with `"type": "module"` in the root and all packages. My test file is in a package dir. I would expect playwright-test to load using the appropriate method depending on module type.

JoelEinbinder:
[...]

PR #7542, opened 2021-07-09
feat(test-runner): support esm modules in more places

JoelEinbinder:
This extends our .mjs support everywhere to include config files, globalSetup/teardown, and reporters.﻿

If a user has `"type": "module"` in thier package.json, today we will fail with an `ERR_REQUIRE_ESM` error.
This patch makes it so that we catch this error and instead load the file as a module. Performance impact
of the try-catch is minimal.

### DIFF
--- a/src/test/cli.ts
+++ b/src/test/cli.ts
@@ -29,6 +29,7 @@ const defaultReporter = process.env.CI ? 'dot' : 'list';
 const builtinReporters = ['list', 'line', 'dot', 'json', 'junit', 'null'];
 const tsConfig = 'playwright.config.ts';
 const jsConfig = 'playwright.config.js';
+const mjsConfig = 'playwright.config.mjs';
 const defaultConfig: Config = {
   preserveOutput: 'always',
   reporter: [ [defaultReporter] ],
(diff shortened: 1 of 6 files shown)


src/test/workerRunner.ts
```diff
@@ -89,11 +89,11 @@ export class WorkerRunner extends EventEmitter {
     return this._project.config.timeout ? monotonicTime() + this._project.config.timeout : undefined;
   }
 
-  private _loadIfNeeded() {
+  private async _loadIfNeeded() {
```
dgozman, line 92: I think this method is used in two places, perhaps you need a rebase?

src/test/cli.ts
```diff
@@ -29,6 +29,7 @@ const defaultReporter = process.env.CI ? 'dot' : 'list';
 const builtinReporters = ['list', 'line', 'dot', 'json', 'junit', 'null'];
 const tsConfig = 'playwright.config.ts';
 const jsConfig = 'playwright.config.js';
+const mjsConfig = 'playwright.config.mjs';
```
dgozman, line 32: Please also update documentation, for example https://playwright.dev/docs/next/intro#create-a-configuration-file.

dgozman (approved):

src/test/workerRunner.ts
JoelEinbinder: Yep there was one in cleanup. done.

src/test/cli.ts
JoelEinbinder: I think we don't need to call out `.mjs` in the docs. It works for consistency but it's really just noise in the docs.

mxschmitt:
Fixed #7452

Issue #6946, opened 2021-06-07
[BUG] ffmpeg can't get used when before downloaded with different architecture


mxschmitt:
**Context:**
- Playwright Version: ToT
- Operating System: Windows
- Node.js version: 14

**Steps to reproduce**

1. Use Playwright for Python on x64 Windows which will use the x64 driver and download ffmpeg x64
2. Use Node.js x32. Run `npx playrwight install` and use the recordVideo functionality in Chromium. This requires ffmpeg 
3. Then Node.js reports: "Hey please install ffmpeg"
4. I run `npx playwright install ffmpeg` - nothing happens

Expected: It does download the x32 executable which is in this scenario required for Node.js
Actual: It does not download anything because the folder is there.

aslushnikov:
We're dropping win32 altogether, so this will become irrelevant!

PR #7544, opened 2021-07-09
chore: replace Registry api with Executable list

dgozman:


src/cli/cli.ts
```diff
@@ -122,24 +122,24 @@ program
 program
     .command('install [browserType...]')
     .description('ensure browsers necessary for this version of Playwright are installed')
-    .action(async function(args) {
+    .action(async function(args: any[]) {
```
mxschmitt, line 125:
```suggestion
    .action(async function(args: string[]) {
```

src/cli/cli.ts
```diff
@@ -122,24 +122,24 @@ program
...
       try {
         // Install default browsers when invoked without arguments.
         if (!args.length) {
           await registry.installBinaries();
           return;
         }
-        const browserNames: Set<BrowserName> = new Set(args.filter((browser: any) => allBrowserNames.has(browser)));
-        const browserChannels: Set<BrowserChannel> = new Set(args.filter((browser: any) => allBrowserChannels.has(browser)));
-        const faultyArguments: string[] = args.filter((browser: any) => !browserNames.has(browser) && !browserChannels.has(browser));
+        const binaries = args.map(arg => registry.findBinary(arg)).filter(b => !!b) as RegistryBinary[];
```
mxschmitt, line 132:
```suggestion
        const binaries = args.map(arg => registry.findBinary(arg)).filter(Boolean) as RegistryBinary[];
```
personal preference nit, does afaik the same.

src/cli/cli.ts
```diff
@@ -122,24 +122,24 @@ program
...
+        const faultyArguments: string[] = args.filter((browser: any) => !binaries.find(b => b.name === browser) && !browserChannels.has(browser));
         if (faultyArguments.length) {
           console.log(`Invalid installation targets: ${faultyArguments.map(name => `'${name}'`).join(', ')}. Expecting one of: ${suggestedBrowsersToInstall}`);
           process.exit(1);
         }
-        if (browserNames.has('chromium') || browserChannels.has('chrome-beta') || browserChannels.has('chrome') || browserChannels.has('msedge') || browserChannels.has('msedge-beta'))
-          browserNames.add('ffmpeg');
-        if (browserNames.size)
-          await registry.installBinaries([...browserNames]);
+        if (browserChannels.has('chrome-beta') || browserChannels.has('chrome') || browserChannels.has('msedge') || browserChannels.has('msedge-beta'))
```
mxschmitt, line 139: We internally have somewhere already a reference that chrome-beta is chromium or that msedge is chromium afaik. So the best would be if we could use it there too to just check: if chromium -> ffmpeg.

src/server/trace/viewer/traceViewer.ts
```diff
@@ -140,7 +140,7 @@ export class TraceViewer {
     // Null means no installation and no channels found.
     let channel = null;
     if (traceViewerBrowser === 'chromium') {
-      if (canAccessFile(registry.executablePath('chromium')!)) {
```
mxschmitt, line 143: something not fully related to the PR. canAccessFile is also true if the input argument is a folder path which leads to wrong error messages when a user passes a folder as an executable path.

src/utils/registry.ts
```diff
@@ -232,123 +207,184 @@ function readDescriptors(packagePath: string) {
...
-        windowsExeAndDllDirectories.push(path.join(browserDirectory, 'chrome-win'));
-      if (browserName === 'firefox' || browserName === 'firefox-beta')
-        windowsExeAndDllDirectories.push(path.join(browserDirectory, 'firefox'));
-      if (browserName === 'webkit')
-        windowsExeAndDllDirectories.push(browserDirectory);
-      return await validateDependenciesWindows(windowsExeAndDllDirectories);
-    }
+    if (os.platform() === 'linux')
+      return await validateDependenciesLinux(linuxLddDirectories.map(d => path.join(browserDirectory, d)), dlOpenLibraries);
+    if (os.platform() === 'win32' && os.arch() === 'x64')
```
mxschmitt, line 376: Also not fully related to the PR, but this makes me feel bad bad when I see it - tried to rework it locally but was not happy with the approach.

Problem: When a user has installed the 32 bit version of Node.js on a 64 bit Windows system which quite a few user by miss-installation do, os.arch() returns `ia32` which has some major disadvantages.
- Wrong statistical metrics -> we get too much downloads for ia32 and think these are actually real ia32 systems (main goal: drop unused architectures but we can't due wrong downloads get generated)
- Performance downsides -> 32bit browsers get used under 64 bit windows
- When the user has installed a different language binding, that finds out the system arch correctly like Python to download chromium -> it downloads 64bit ffmpeg, then uses wrongly installed Node.js (32 bit) -> it tries to download 32 bit ffmpeg - (os arch based executable path) - but sees the folder which is there and thinks it there. Still internally when launching ffmpeg it fails, see #6946

Action item at some point: We should change how we determine if its a 32 or 64 bit system on Windows, maybe through the `PROCESSOR_ARCHITEW6432 ` env var, but unsure. It needs to be the same on all the language bindings.

mxschmitt (approved):

src/utils/registry.ts
```diff
@@ -232,123 +207,184 @@ function readDescriptors(packagePath: string) {
...
     };
     return descriptor;
   });
 }
 
+export type BrowserName = 'chromium' | 'firefox' | 'webkit';
+type InternalTool = 'ffmpeg' | 'firefox-beta' | 'chromium-with-symbols';
+const allDownloadable = ['chromium', 'firefox', 'webkit', 'ffmpeg', 'firefox-beta', 'chromium-with-symbols'];
+
+export interface RegistryBinary {
```
aslushnikov, line 220: Executable?

src/server/browserType.ts
```diff
@@ -36,16 +36,18 @@ const ARTIFACTS_FOLDER = path.join(os.tmpdir(), 'playwright-artifacts-');
...
   private _name: BrowserName;
   readonly _playwrightOptions: PlaywrightOptions;
+  readonly _binary: RegistryBinary;
 
   constructor(browserName: BrowserName, playwrightOptions: PlaywrightOptions) {
     super(playwrightOptions.rootSdkObject, 'browser-type');
     this.attribution.browserType = this;
     this._playwrightOptions = playwrightOptions;
     this._name = browserName;
+    this._binary = registry.findBinary(browserName);
```
aslushnikov, line 46: do not save binary on browser type since it depends on options in launch

src/cli/cli.ts
dgozman: Yeah, this code will change one more time and will be simplified.

src/server/browserType.ts
```diff
@@ -45,7 +45,7 @@ export abstract class BrowserType extends SdkObject {
   }
 
   executablePath(channel?: string): string {
-    return registry.executablePath(this._name) || '';
+    return registry.findExecutable(this._name).executablePathIfExists() || '';
```
aslushnikov, line 48: This is functional change.

Today, when installing `@playwright/test` without running `npm playwright install`, the `chromium.executablePath()` will return a value. With this change, it will return empty string until the browsers are installed.

Do you think it's worth keeping the current behavior just to be on the safe side?

src/server/firefox/firefox.ts
```diff
@@ -37,7 +37,7 @@ export class Firefox extends BrowserType {
     if (channel) {
       let executablePath = undefined;
       if ((channel as any) === 'firefox-beta')
-        executablePath = registry.executablePath('firefox-beta');
+        executablePath = registry.findExecutable('firefox-beta')!.executablePathIfExists();
       assert(executablePath, `unsupported firefox channel "${channel}"`);
       assert(fs.existsSync(executablePath), `"${channel}" channel is not installed. Try running 'npx playwright install ${channel}'`);
```
aslushnikov, line 42: nit: this assert never fails.

src/server/chromium/crPage.ts
```diff
@@ -845,7 +845,7 @@ class FrameSession {
 
   async _createVideoRecorder(screencastId: string, options: types.PageScreencastOptions): Promise<void> {
     assert(!this._screencastId);
-    const ffmpegPath = registry.executablePath('ffmpeg');
+    const ffmpegPath = registry.findExecutable('ffmpeg')!.executablePathIfExists();
     if (!ffmpegPath)
```
aslushnikov, line 849: IIUC this check was never hit before, and becomes an actual check with this patch. Cool!

src/utils/registry.ts
```diff
@@ -385,22 +412,34 @@ export class Registry {
...
+          await executable._download();
+        else
+          throw new Error(`ERROR: Playwright does not support installing ${executable.name}`);
       }
     } finally {
       await releaseLock();
     }
   }
 
+  private async _downloadExecutable(descriptor: BrowsersJSONDescriptor, executablePath: string | undefined, downloadURLTemplate: string | undefined, downloadHostEnv: string) {
```
aslushnikov, line 427: nit: this doesn't rely on registry and can be a function instead of a private registry method

src/utils/registry.ts
```diff
@@ -385,22 +412,34 @@ export class Registry {
...
+    const downloadURL = util.format(downloadURLTemplate, downloadHost, descriptor.revision);
+    const title = `${descriptor.name} v${descriptor.revision}`;
+    const downloadFileName = `playwright-download-${descriptor.name}-${hostPlatform}-${descriptor.revision}.zip`;
+    await downloadBrowserWithProgressBar(title, descriptor.dir, executablePath, downloadURL, downloadFileName).catch(e => {
+      throw new Error(`Failed to download ${title}, caused by\n${e.stack}`);
+    });
+    await fs.promises.writeFile(markerFilePath(descriptor.dir), '');
+  }
+
   private async _validateInstallationCache(linksDir: string) {
```
aslushnikov, line 443: nits:
- this doesn't rely on registry instance and can be made static. It'll make it more clear
- `linksDir` can be made static too the same way `registryDirectory` is static.

src/utils/registry.ts
```diff
@@ -385,22 +412,34 @@ export class Registry {
...
-        await downloadBrowserWithProgressBar(title, browserDirectory, this.executablePath(browserName)!, this._downloadURL(browserName), downloadFileName).catch(e => {
-          throw new Error(`Failed to download ${title}, caused by\n${e.stack}`);
-        });
-        await fs.promises.writeFile(markerFilePath(browserDirectory), '');
+      // Install browsers for this package.
+      for (const executable of executables) {
+        if (executable._download)
+          await executable._download();
+        else
+          throw new Error(`ERROR: Playwright does not support installing ${executable.name}`);
```
aslushnikov, line 420: IIUC this will change later when channels move here too. Let me know if I misunderstand.

src/utils/registry.ts
```diff
@@ -232,123 +207,176 @@ function readDescriptors(packagePath: string) {
...
-    return this._descriptors.some(browser => browser.name === browserName);
+  findExecutable(name: BrowserName): Executable;
+  findExecutable(name: string): Executable | undefined;
+  findExecutable(name: string): Executable | undefined {
+    return this._executables.find(b => b.name === name);
   }
 
-  private _installByDefault(): BrowserName[] {
-    return this._descriptors.filter(browser => browser.installByDefault).map(browser => browser.name);
+  private _addRequirementsAndDedupe(executables: Executable[] | undefined): ExecutableImpl[] {
```
aslushnikov, line 347: nit: make function instead of a method? I'd pop the no-executables usecase to the call sites. It'll make the `install` and `installDeps` API more clear since the default value will be readily available.

src/utils/registry.ts
```diff
@@ -232,123 +207,176 @@ function readDescriptors(packagePath: string) {
...
+    for (const executable of executables as ExecutableImpl[]) {
+      set.add(executable);
+      for (const req of executable._requires)
+        set.add(this._executables.find(b => b.name === req)!);
+    }
+    return Array.from(set);
   }
 
-  async validateHostRequirements(browserName: BrowserName) {
+  private async _validateHostRequirements(browserName: BrowserName, browserDirectory: string, linuxLddDirectories: string[], dlOpenLibraries: string[], windowsExeAndDllDirectories: string[]) {
```
aslushnikov, line 359: nit: move out of registry class?

aslushnikov (approved):

src/utils/registry.ts
dgozman: In a follow-up!

src/utils/registry.ts
dgozman: Exactly.

src/utils/registry.ts
dgozman: Yep, doing that in a follow up.

src/utils/registry.ts
dgozman: I'll see about that.

src/server/browserType.ts
dgozman: Good catch! My next PR fixes this, so I'll port the changes here.

PR #7569, opened 2021-07-12
test: unskip/unfixme tests which are in Chrome stable now

mxschmitt:
drive-by some minor cleanups and a new TC.

Download.cancel is by that ready.


mxschmitt:
Closed since Electron is still running on an old Electron.

PR #7573, opened 2021-07-12
docs(test-runner): section for multiple reporters

dgozman:


docs/src/test-reporters.md
```diff
@@ -38,21 +38,50 @@ const config: PlaywrightTestConfig = {
 export default config;
 ```
 
-You can use different reporters locally and on CI.
+### Multiple reporters
+
+You can use multiple reporters at the same time, for example `'list'` for nice terminal output and `'json'` for comprehensive json file with test results.
```
JoelEinbinder, line 43:
```suggestion
You can use multiple reporters at the same time. For example  you can use`'list'` for nice terminal output and `'json'` to get a comprehensive json file with the test results.
```

PR #7580, opened 2021-07-13
browser(firefox): introduce auto-open-devtools-for-tabs CLI flag

mxschmitt:
This adds support for the `auto-open-devtools-for-tabs` CLI flag like its implemented in Chromium. We then can use this to provide support for the `devtools: true` flag when launching Firefox.

Works fine in general except when you are creating a page and closing it's context in a short time interval (when it was not fully initialised) it leads to a broken open devtools instance (white window). I tried to fix it but was not able to, probably we can live with it.

https://github.com/mxschmitt/gecko-dev/commit/478dab6f4992660d9e93ef9ff6ab86ff5ac29a47

### DIFF
--- a/browser_patches/firefox-beta/patches/bootstrap.diff
+++ b/browser_patches/firefox-beta/patches/bootstrap.diff
@@ -171,6 +171,67 @@ index 040c7b124dec6bb254563bbe74fe50012cb077a3..b4e6b8132786af70e8ad0dce88b67c28
  
    const transportProvider = {
      setListener(upgradeListener) {
+diff --git a/devtools/startup/DevToolsStartup.jsm b/devtools/startup/DevToolsStartup.jsm
+index 2c1f767f89f596fb5732572ad4e0cad3cafcc7c3..99401071ad93a2f2a3ba37e7643920f2e05e8b25 100644
+--- a/devtools/startup/DevToolsStartup.jsm
++++ b/devtools/startup/DevToolsStartup.jsm
+@@ -375,7 +375,7 @@ DevToolsStartup.prototype = {
+ 
+       // Store devtoolsFlag to check it later in onWindowReady.
+       this.devtoolsFlag = flags.devtools;
+-
++      this.autoOpenDevtoolsForTabs = flags.autoOpenDevtoolsForTabs;
+       /* eslint-disable mozilla/balanced-observers */
+       // We are not expecting to remove those listeners until Firefox closes.
+ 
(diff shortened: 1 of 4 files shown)


browser_patches/firefox-beta/patches/bootstrap.diff
```diff
@@ -124,6 +124,199 @@ index d0449bcae84ef1fad06f26a8168eaef7599a0d7a..22314a34e8d12679643214fc416a1ba8
...
  @RESPATH@/components/TestInterfaceJS.manifest
+diff --git a/browser/locales/l10n-changesets.json b/browser/locales/l10n-changesets.json
+index 4bd1e88953b54ac24cce6824cd306b7edacd9152..33655204b740a4827c8c4922ddeecda3c11bf74d 100644
+--- a/browser/locales/l10n-changesets.json
++++ b/browser/locales/l10n-changesets.json
+@@ -87,7 +87,7 @@
+             "win64-aarch64-devedition",
+             "win64-devedition"
+         ],
+-        "revision": "90f850248aab9bda6eb64281ecdcf9bf58b87b0e"
```
mxschmitt, line 135: Will try to revert that.

yury-s (commented):
Can you upload pretty diff too?

browser_patches/firefox-beta/patches/bootstrap.diff
```diff
@@ -171,6 +364,67 @@ index 040c7b124dec6bb254563bbe74fe50012cb077a3..b4e6b8132786af70e8ad0dce88b67c28
...
+     JsonView.initialize();
+   },
+ 
++  handleDevToolsOpenForEveryTab(window) {
++    const require = this.initDevTools("CommandLine");
++    const { gDevTools } = require("devtools/client/framework/devtools");
++    window.gBrowser.tabContainer.addEventListener('TabOpen', async (event) => {
++      await gDevTools.showToolboxForTab(event.target);
++    });
++    gDevTools.showToolboxForTab(window.gBrowser.selectedTab).catch(() => {});
```
yury-s, line 414: I assume DevTools have their own browser and this code will not run recursively.

yury-s (approved):

browser_patches/firefox-beta/patches/bootstrap.diff
mxschmitt: Currently this runs in the Playwright world per Page, since we create for every Page a new Window.

At least from my testing it didn't.

Issue #7576, opened 2021-07-13
[Bug]: Mousemove events have inconsistent button states on webkit


yongjik:
### Playwright version

1.12.3

### Operating system

Linux

### What browsers are you seeing the problem on?

WebKit

### Other information

Linux: Ubuntu 20.04, Node.js version: 15.8.0. (Saw the same bug with Python version - Python 3.9.1, Playwright v1.12.1)

### What happened? / Describe the bug

When I call `page.mouse.down()` and then `page.mouse.move()`, the fired event's `buttons` property should indicate that the mouse button is currently down, but it's not properly set on Webkit.

If fixing it is too much work, it might be nice if there's at least an option to set the button state manually with `mouse.move()`.

### Code snippet to reproduce your bug

```shell
const { webkit } = require('playwright');

(async () => {
    const browser = await webkit.launch();
    const page = await browser.newPage();
    await page.goto('file:///' + __dirname + '/pwtest.html');

    await page.mouse.move(50, 50);
    await page.mouse.down();
    await page.mouse.move(300, 150, {steps: 3});
    await page.mouse.up();
    await page.mouse.move(500, 120, {steps: 3});

    const dbglog = await page.waitForSelector('#dbglog');
    console.log(await dbglog.textContent());

    await browser.close();
})();

// Add the following in the same directory as "pwtest.html"
<html>
<head>
<style>
#test { height: 300px; width: 80%; border: 1px solid blue; }
#dbglog { white-space: pre; }
</style>

<script>
function onload() {
    'use strict';

    let target = document.querySelector('#test')
    for (let evtype of ['mousedown', 'mousemove', 'mouseup']) {
        target.addEventListener(evtype, (ev) => {
            let msg = `${evtype}: x=${ev.clientX} y=${ev.clientY} btn=${ev.buttons}\n`;
            document.getElementById('dbglog').appendChild(
                document.createTextNode(msg));
        });
    }
}
</script>
</head>

<body onload="onload();">
<div id="test"></div>
<div id="dbglog"></div>
</body>

</html>
```

### Relevant log output

```shell
pw:api => browserType.launch started +0ms
  pw:api <= browserType.launch succeeded +126ms
  pw:api => browser.newContext started +0ms
  pw:api <= browser.newContext succeeded +6ms
  pw:api => browserContext.newPage started +1ms
  pw:api <= browserContext.newPage succeeded +128ms
  pw:api => page.goto started +0ms
[...]

PR #7584, opened 2021-07-13
test(webkit): add failing test for event.buttons

JoelEinbinder:
#7576﻿


yury-s:
Is it different from https://github.com/microsoft/playwright/pull/7577 ?

JoelEinbinder:
> Is it different from #7577 ?

I didn't see yours but I do like mine better :)

PR #7591, opened 2021-07-14
devops: enforce hard xcode version match for browser builds

aslushnikov:
Browsers require particular xcode versions to be installed on the host
machine to produce successful build.

This patch starts using explicit xcode versions for all mac browser
builds.

### DIFF
--- a/browser_patches/webkit/pw_run.sh
+++ b/browser_patches/webkit/pw_run.sh
@@ -2,7 +2,9 @@
 
 function runOSX() {
   # if script is run as-is
-  if [[ -d $SCRIPT_PATH/checkout/WebKitBuild/Release/Playwright.app ]]; then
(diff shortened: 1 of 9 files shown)


browser_patches/webkit/pw_run.sh
```diff
@@ -2,7 +2,9 @@
 
 function runOSX() {
   # if script is run as-is
-  if [[ -d $SCRIPT_PATH/checkout/WebKitBuild/Release/Playwright.app ]]; then
+  if [[ -f "${SCRIPT_PATH}/EXPECTED_BUILDS" && -n "$WK_CHECKOUT_PATH" -d "$WK_CHECKOUT_PATH/WebKitBuild/Release/Playwright.app" ]]; then
```
aslushnikov, line 5: drive-by: support running webkit from custom checkout path.

PR #7594, opened 2021-07-14
feat(test-runner): implement test.step

dgozman:
```js
test('example test', async ({ page }) => {
  await test.step('Sign in', async () => {
    await page.fill('#username', 'Username');
    await page.fill('#password', 'Password');
    await page.click('text=Sign in');
  });
  ...
});
```

Test step title is printed next to the test title when available:

```sh
  ✓ example.spec.ts:9:6 › [chromium] example test
    example.spec.ts:9:6 › [firefox] example test › Go to cart › substep
  ✘ 1) example.spec.ts:9:6 › [webkit] example test › Sign in (419ms)
```

### DIFF
--- a/src/test/reporter.ts
+++ b/src/test/reporter.ts
@@ -57,18 +57,26 @@ export interface TestResult {
   duration: number;
   status?: TestStatus;
   error?: TestError;
+  errorStep?: TestStep;
   data: { [key: string]: any };
   stdout: (string | Buffer)[];
   stderr: (string | Buffer)[];
+  steps: TestStep[];
 }
-export type FullResult = {
+export interface TestStep {
+  title: string;
+  steps: TestStep[];
+  titlePath(): string[];
+}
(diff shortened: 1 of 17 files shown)


src/test/reporter.ts
```diff
@@ -57,18 +57,26 @@ export interface TestResult {
...
+export interface FullResult {
   status: 'passed' | 'failed' | 'timedout' | 'interrupted';
-};
+}
 export interface Reporter {
   onBegin(config: FullConfig, suite: Suite): void;
   onTestBegin(test: Test): void;
   onStdOut(chunk: string | Buffer, test?: Test): void;
   onStdErr(chunk: string | Buffer, test?: Test): void;
+  onTestStep(test: Test, step?: TestStep): void;
```
pavelfeldman, line 79: It feels like onStepBegin / onStepEnd would work better due to things like timing steps.

PR #7599, opened 2021-07-14
browser(webkit): support event.buttons for MacOS

JoelEinbinder:
#7576

### DIFF
--- a/browser_patches/webkit/patches/bootstrap.diff
+++ b/browser_patches/webkit/patches/bootstrap.diff
@@ -10280,6 +10280,18 @@ index 7a5893eb68ad24cf92c832070485489b7cfafa0c..a2d68a5eb59d8b7155c1e17b75a12e55
  #endif
  
      const String& text() const { return m_text; }
+diff --git a/Source/WebKit/Shared/WebMouseEvent.h b/Source/WebKit/Shared/WebMouseEvent.h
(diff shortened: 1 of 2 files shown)


browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -17123,11 +17135,8 @@ index 0000000000000000000000000000000000000000..94e9f24353337169992724e2fcdf7086
 +
 +private:
 +    void platformDispatchKeyEvent(WebKeyboardEvent::Type type, const String& text, const String& unmodifiedText, const String& key, const String& code, const String& keyIdentifier, int windowsVirtualKeyCode, int nativeVirtualKeyCode, bool isAutoRepeat, bool isKeypad, bool isSystemKey, OptionSet<WebEvent::Modifier> modifiers, Vector<String>& commands, WallTime timestamp);
-+#if PLATFORM(WPE)
-+    void platformDispatchMouseEvent(WebMouseEvent::Type type, int x, int y, WebMouseEvent::Button button, OptionSet<WebEvent::Modifier> modifiers);
```
pavelfeldman, line 17127: Somehow this split was important.

JoelEinbinder: Maybe it was, but this method is not implemented anywhere and never called.

PR #7604, opened 2021-07-14
fix(test-runner): start webServer before globalSetup/teardown

### LINKED ISSUE #7597: [Bug]: Webserver should run before global-setup
### Playwright version

@next

### Operating system

_No response_

_No response_

### Other information

_No response_

related to #7368

When using @next with the webserver config global-setup runs first. (I want to use e2e tests in an angular project)

where global-setup is being used to create an initial state as per https://playwright.dev/docs/auth#reuse-authentication-in-playwright-test this will fail as the webserver has not started.

I have tested a fix by moving the webserver line up before global-setup and it "works", but I have not built any tests
https://github.com/microsoft/playwright/blob/master/src/test/runner.ts#L170-L173

```shell
    webServer: {
        command: 'npx ng serve',
        timeout: 120 * 1000,
    },
    use: {
        // Browser options
        slowMo: 50,

        // Context options
        //viewport: { width: 2048, height: 1280 },
        ignoreHTTPSErrors: true,
        //permissions:['clipboard-read'],

        // Artifacts
        video: 'retry-with-video',
    },
};
export default config;


```typescript
import { firefox } from '@playwright/test';

async function globalSetup() {
  //const browser = await firefox.launch();
  const browser = await firefox.launch({});

    ignoreHTTPSErrors: true
});
  //const page = await browser.newPage();
  await page.goto(`${process.env.PLAYWRIGHT_TEST_BASE_URL}/login`);
  await page.fill('input[type="text"][name="username"]', process.env.USERNAME);
  await page.fill('input[type="password"][name="password"]', process.env.PASSWORD);
  await page.click('input[type="submit"]');
  await page.waitForTimeout(2000)
  await page.context().storageState({ path: 'state.json' });
  await browser.close();
}

export default globalSetup;
```

### Relevant log output

_No response_

mxschmitt:
Fixes #7597

### DIFF
--- a/tests/playwright-test/web-server.spec.ts
+++ b/tests/playwright-test/web-server.spec.ts
@@ -35,13 +35,38 @@ test('should create a server', async ({ runInlineTest }, { workerIndex }) => {
         webServer: {
           command: 'node ${JSON.stringify(path.join(__dirname, 'assets', 'simple-server.js'))} ${port}',
           port: ${port},
-        }
+        },
+        globalSetup: 'globalSetup.ts',
+        globalTeardown: 'globalTeardown.ts',
+      };
+    `,
+    'globalSetup.ts': `
+      module.exports = async () => {
+        console.log('globalSetup')
+        return () => console.log('globalSetup teardown');
+      };
+    `,
+    'globalTeardown.ts': `
+      module.exports = async () => {
+        const http = require("http");
+        const response = await new Promise(resolve => {
+          const request = http.request("http://localhost:${port}/hello", resolve);
+          request.end();
+        })
+        console.log('globalTeardown-status-'+response.statusCode)
       };
     `,
   });
   expect(result.exitCode).toBe(0);
   expect(result.passed).toBe(1);
   expect(result.report.suites[0].specs[0].tests[0].results[0].status).toContain('passed');
+
+  const expectedLogMessages = ['Starting WebServer', 'globalSetup', 'globalSetup teardown', 'globalTeardown-status-200'];
+  const actualLogMessages = expectedLogMessages.map(log => ({
+    log,
+    index: result.output.indexOf(log),
+  })).sort((a, b) => a.index - b.index).filter(l => l.index !== -1).map(l => l.log);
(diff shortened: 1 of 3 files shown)


plessbd:
If I knew it wasnt going to need any tests, I would have done this myself ;)

mxschmitt:
@plessbd god point! I've added a test-case.

plessbd:
Looks good, still pretty simple, and I should have done it myself. So I feel bad, but I will blame it on it was midnight when I opened the ticket, yeah...
Thank you, and if I find anything more I will try to do more than just a bug report

mxschmitt:
@plessbd we release in ~1 day, thats why I picked it up. For the future, I recommend to write something like "I'm willing to contribute a fix" to the issue report. We are more than happy to introduce new contributors to the project. :)

tests/playwright-test/web-server.spec.ts
```diff
@@ -35,13 +35,38 @@ test('should create a server', async ({ runInlineTest }, { workerIndex }) => {
...
+        })
+        console.log('globalTeardown-status-'+response.statusCode)
       };
     `,
   });
   expect(result.exitCode).toBe(0);
   expect(result.passed).toBe(1);
   expect(result.report.suites[0].specs[0].tests[0].results[0].status).toContain('passed');
+
+  const expectedLogMessages = ['Starting WebServer', 'globalSetup', 'globalSetup teardown', 'globalTeardown-status-200'];
```
dgozman, line 64: We usually use `%%` for this kind of stuff: https://github.com/microsoft/playwright/blob/master/tests/playwright-test/fixtures.spec.ts#L619-L625

dgozman (approved):

tests/playwright-test/web-server.spec.ts
mxschmitt: Tried, but since Starting WebServer is quite dynamic (followed by the command which is escaped) I would keep it like that for now.

FezVrasta:
It looks like the global setup is running before the web server becomes available.

zomars:
Same here @FezVrasta using latest version as of today.

PR #7608, opened 2021-07-14
docs: add note about ServiceWorkers and request interception

### LINKED ISSUE #7588: [Question]: Is route interception on a page supposed to work in every case (Chromium)?
### Your question

Is Page.route supposed to work on page requests alongside a service worker in every case?

I found a reproducible case where this doesn't work. Not sure if this is a bug, just not documented, or I missed it.

After troubleshooting and observation (and finding a workaround), I went back and created a simple, minimal example where Page.route (for page requests), alongside a service worker, will hang a test indefinitely.

Here are the conditions required for trouble:
* Chromium, as installed by Playwright
* MacOs, Catalina
1. Page.route called before any navigation for complete capture
2. Page.click called on [type='submit'] to trigger form, handled with client fetch as 'application/json'
3. Expressjs (4.17.1 bare-bones log and reply) on http://localhost (`localhost` important, perhaps protocol too but **I didn't check proto** - https remote host works)
4. A service worker (workbox 6.1.5) installed, `NetworkOnly` (no cache) strategy on the route POST

The observed failure:
In the failure condition, the request is still made (as logged by Expressjs) to localhost, but any `Page.on('response', handler)` or `Page.route(() => true, handler)` or matching `Page.waitForResponse` will never execute.

The workaround for me is to avoid Page.route or stop the service worker from installing.

I debugged with `pw:api` and all combos of Page.click options. This led me to code the minimal reproducible example, and ask the question.

In the minimal reproduction:
When Page.route, Page.click, and localhost server are used, adding or removing the service worker toggles the failure condition.

aslushnikov:
Fixes #7588

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -2383,6 +2383,12 @@ Once routing is enabled, every request matching the url pattern will stall unles
 The handler will only be called for the first url if the response is a redirect.
 :::
 
+
+:::note
+`page.route` will not affect requests intercepted by ServiceWorker. See https://github.com/microsoft/playwright/issues/1090
+We recommend disabling ServiceWorkers when using request interception.
+:::
+
 An example of a naive handler that aborts all image requests:
 
 ```js


mxschmitt (approved):
I would maybe include the following snippet:

```
await context.addInitScript(() => delete window.navigator.serviceWorker);
```

and use a markdown link instead of a raw URL.

PR #7610, opened 2021-07-14
feat(dotnet): add revision for generated API

avodovnik:


mxschmitt (commented):
Don't we generate it when we update the driver? The driver includes a file called "commitinfo" inside the package dir afaik.

avodovnik:
> Don't we generate it when we update the driver? The driver includes a file called "commitinfo" inside the package dir afaik.

This is not related to the driver, but the docs (markdown files) we generate the API from. Those two can (and do) drift.

pavelfeldman:
I need a hint understanding what this helps with.

avodovnik:
> I need a hint understanding what this helps with.

I've started working on improving the .NET engineering side. There's an open [PR](https://github.com/microsoft/playwright-dotnet/pull/1618) which brings this repo as a submodule. Doing this will let me keep better track of changes as I move forward with rolling stuff. It'll also help me keep a pointer to the commit from which we generated the API for .NET. 

This marker file will help me identify drift (i.e. the submodule getting updated, but nothing would get regenerated) etc.

mxschmitt:
Closing since it does not seem relevant anymore with the new submodule logic. (Driver version -> upstream commit sha + verification bots)

PR #7347, opened 2021-06-28
chore: remove hard-coded wrapApi names

pavelfeldman:

PR #7622, opened 2021-07-14
chore: disable source-map reference in JS for production builds

### LINKED ISSUE #7325: [BUG] Output files are with non-existing source-maps
Either remove source-map annotation or add source-maps.

mxschmitt:
Fixes #7325

### DIFF
--- a/src/utils/stackTrace.ts
+++ b/src/utils/stackTrace.ts
@@ -39,8 +39,13 @@ const PW_LIB_DIRS = [
   path.join('@playwright', 'test'),
 ].map(packageName => path.sep + packageName);
 
-const runnerLib = path.join('@playwright', 'test', 'lib', 'test');
+const runnerNpmPkgLib = path.join('@playwright', 'test', 'lib', 'test');
+const runnerLib = path.join('lib', 'test');
 const runnerSrc = path.join('src', 'test');
+
+function includesFileInPlaywrightSubDir(subDir: string, fileName: string) {
+  return PW_LIB_DIRS.map(p => path.join(p, subDir)).some(libDir => fileName.includes(libDir));
+}
 
 export type ParsedStackTrace = {
   frames: StackFrame[];
@@ -59,7 +64,7 @@ export function captureStackTrace(): ParsedStackTrace {
   const lines = stack.split('\n').reverse();
   let apiName = '';
 
-  const isTesting = process.env.PWTEST_CLI_ALLOW_TEST_COMMAND || isUnderTest();
+  const isTesting = !!process.env.PWTEST_CLI_ALLOW_TEST_COMMAND || isUnderTest();
 
   for (const line of lines) {
     const frame = stackUtils.parseLine(line);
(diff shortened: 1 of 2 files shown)


dgozman (approved):

src/utils/stackTrace.ts
```diff
@@ -87,6 +90,13 @@ export function captureStackTrace(): ParsedStackTrace {
   return { frames, frameTexts, apiName };
 }
 
+function shouldExtractAPIName(isTesting: boolean, fileName: string): boolean {
```
JoelEinbinder, line 93: nit:
```suggestion
function isFilePartOfPlaywrightDriver(isTesting: boolean, fileName: string): boolean {
```

JoelEinbinder (approved):

dgozman (commented):
Is there any description of what is going on? I am worried that we grow the logic here, but have no tests for it...

mxschmitt:
> Is there any description of what is going on? I am worried that we grow the logic here, but have no tests for it...

The logic in general didn't change much, the reasoning behind this refactor was that it was not very readable. What has changed is that it accepts now the lib/ part too instead of src/ only. Since before we had source-maps locally, which resolved the JS stack-traces, so we get while testing src/* based file names. We still have src/* when we use npm run watch, since for that we have source-maps to allow-us to use e.g. vsc to debug or to have full stack-traces when errors got thrown.

src/utils/stackTrace.ts
```diff
@@ -39,9 +39,12 @@ const PW_LIB_DIRS = [
   path.join('@playwright', 'test'),
 ].map(packageName => path.sep + packageName);
 
-const runnerLib = path.join('@playwright', 'test', 'lib', 'test');
+const runnerNpmPkgLib = path.join('@playwright', 'test', 'lib', 'test');
+const runnerLib = path.join('lib', 'test');
 const runnerSrc = path.join('src', 'test');
 
+const includesFileInPlaywrightSubDir = (subDir: string, fileName: string) => PW_LIB_DIRS.map(p => path.join(p, subDir)).some(libDir => fileName.includes(libDir));
```
dgozman, line 46: Please make this a regular function.

src/utils/stackTrace.ts
```diff
@@ -87,6 +90,13 @@ export function captureStackTrace(): ParsedStackTrace {
   return { frames, frameTexts, apiName };
 }
 
+function isFilePartOfPlaywright(isTesting: boolean, fileName: string): boolean {
+  const isPlaywrightTest = fileName.includes(runnerNpmPkgLib);
```
dgozman, line 94: Why this check? What are we trying to achieve? Perhaps all this logic should be replaced by "fileName is not in <playwright>/lib/client"?

dgozman (commented):
Is there a way to add tests for this?

src/utils/stackTrace.ts
mxschmitt: This complexity was introduced in #7347, so let's ask Pavel.

pavelfeldman: We can try simplifying it. The goal here is to identify the point where we hit Playwright Library API, treating @lpaywright/test as user land. As far as I remember, the complexity is around behaving differently while under test and while in production.

pavelfeldman:
I don't see changes that relate to the PR title anymore. Would it make sense to only generate source maps for lib/client and not anything else?

JoelEinbinder:
> I don't see changes that relate to the PR title anymore.

This patch is mainly the one line in build.js. It is still there

> Would it make sense to only generate source maps for lib/client and not anything else?

We can revisit whether we need source maps in a separate patch, but it looks like we don't. Today we generate source maps, but we don't ship them via .npmignore. This causes errors for Max in vscode because it warns when a sourceMappingURL directive points to a file that isn't on disk.

So not generating the sourceMappingURLs is a good thing, and we do it by just not generating source maps for our production run. This causes the bots to fail. Our bots were running with source maps and the source-map-support, which caused them to have different stack traces when compared to production. The changes to stackTrace.ts allow us to run our tests both with and without sourcemaps.

PR #7623, opened 2021-07-14
fix: add log entry if selector resolves to multiple elements on click

### LINKED ISSUE #3264: [Feature] Page.click() add log entry if > 1 elements will be found in actionability checks
We faced into the issue that we have element A and B with the same buttons and the same text inside. Element A does not pass the actionability checks. But element B does.

For me as a user the best would be if element B will be used instead then, since @pavelfeldman mentioned in a chat we should not change the API, a log message would make sense there if more than 1 element will be found so we can give the user a hint in the Error which will be thrown.

Minimal example: https://try.playwright.tech/?s=47h4i
Removing the first div with the hidden class will lead to the popup 2. If we don't modify it it won't find any element since its waiting for the first div.

mxschmitt:
Should bring value in the trace viewer and when using debug logs.

~~Probably its better to get away from querySelector + querySelectorAll and use just querySelectorAll at the top?~~

Fixes #3264

### DIFF
--- a/src/server/dom.ts
+++ b/src/server/dom.ts
@@ -938,15 +938,19 @@ export function waitForSelectorTask(selector: SelectorInfo, state: 'attached' | 
     let lastElement: Element | undefined;
 
     return injected.pollRaf((progress, continuePolling) => {
-      const element = injected.querySelector(parsed, root || document);
+      const elements = injected.querySelectorAll(parsed, root || document);
+      const element = elements[0];
       const visible = element ? injected.isVisible(element) : false;
 
       if (lastElement !== element) {
         lastElement = element;
-        if (!element)
+        if (!element) {
           progress.log(`  selector did not resolve to any element`);
-        else
+        } else {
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

src/server/dom.ts
```diff
@@ -943,10 +943,14 @@ export function waitForSelectorTask(selector: SelectorInfo, state: 'attached' |
 
       if (lastElement !== element) {
         lastElement = element;
-        if (!element)
+        if (!element) {
           progress.log(`  selector did not resolve to any element`);
-        else
+        } else {
+          const elementCount = injected.querySelectorAll(parsed, root || document).length;
```
dgozman, line 949: Please don't call querySelectorAll immediately after querySelector with the same selector - this is twice as bad. Even more, I am not sure this warning is worth the performance regression of querying a single element vs querying all of them.

dgozman (approved):

PR #7637, opened 2021-07-15
feat(firefox): add support for the devtools option in browser.launch()

mxschmitt:


pavelfeldman:
How about we go backwards and deprecate it?

mxschmitt:
Closing and reverting in #7717.

PR #7651, opened 2021-07-15
docs(intro-js): make js and ts example consistent

Nav-2d:
ts example was missing an assertion step when compared to js example. Added it to make it consistent.


dgozman (approved):
Thank you!

PR #7657, opened 2021-07-15
Update why-playwright.md

chauey:


pavelfeldman:
We would not typically make a single character changes as a standalone PR. Each PR runs hours and hours of machine time on CI...

PR #7659, opened 2021-07-15
feat(keyboard): low level override for text produced by keyboard.down

JoelEinbinder:
This adds options to `keyboard.down` to override the `event.key` property of the event (for non-US layouts) and the `text` generated (for IME when certain keys will produce no text).

This is a draft to talk about the api. The finished version would need `key` in `keyboard.up` and `key`/`text` in `keyboard.press`. There is also a juggler patch needed to allow overriding the `text` to be blank.

### DIFF
--- a/docs/src/api/class-keyboard.md
+++ b/docs/src/api/class-keyboard.md
@@ -172,6 +172,16 @@ Modifier keys DO influence `keyboard.down`. Holding down `Shift` will type the t
 
 Name of the key to press or a character to generate, such as `ArrowLeft` or `a`.
 
+### option: Keyboard.down.key
+- `key` <[string]>
+
+Override the event.key property of this event.
+
+### option: Keyboard.down.text
+- `text` <[string]>
+
+Override the text produced by this event.
+
 ## async method: Keyboard.insertText
 
 Dispatches only `input` event, does not emit the `keydown`, `keyup` or `keypress` events.
--- a/src/client/input.ts
+++ b/src/client/input.ts
@@ -25,8 +25,8 @@ export class Keyboard implements api.Keyboard {
     this._channel = channel;
   }
 
-  async down(key: string) {
-    await this._channel.keyboardDown({ key });
+  async down(key: string, overrides: channels.PageKeyboardDownOptions['overrides']) {
+    await this._channel.keyboardDown({ key, overrides });
   }
 
   async up(key: string) {
--- a/src/dispatchers/pageDispatcher.ts
+++ b/src/dispatchers/pageDispatcher.ts
@@ -156,7 +156,7 @@ export class PageDispatcher extends Dispatcher<Page, channels.PageInitializer> i
   }
 
   async keyboardDown(params: channels.PageKeyboardDownParams, metadata: CallMetadata): Promise<void> {
-    await this._page.keyboard.down(params.key);
+    await this._page.keyboard.down(params.key, params.overrides);
   }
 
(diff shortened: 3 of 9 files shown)


src/client/input.ts
```diff
@@ -25,8 +25,8 @@ export class Keyboard implements api.Keyboard {
     this._channel = channel;
   }
 
-  async down(key: string) {
-    await this._channel.keyboardDown({ key });
+  async down(key: string, overrides: channels.PageKeyboardDownOptions['overrides']) {
```
dgozman, line 28: `options: channels.PageKeyboardDownOptions`? Why doesn't the standard way work here?

docs/src/api/class-keyboard.md
```diff
@@ -172,6 +172,16 @@ Modifier keys DO influence `keyboard.down`. Holding down `Shift` will type the t
 
 Name of the key to press or a character to generate, such as `ArrowLeft` or `a`.
 
+### option: Keyboard.down.key
```
dgozman, line 175: Call this `eventKey` instead, to avoid confusion between key and key?

docs/src/api/class-keyboard.md
```diff
@@ -172,6 +172,16 @@ Modifier keys DO influence `keyboard.down`. Holding down `Shift` will type the t
...
+- `key` <[string]>
+
+Override the event.key property of this event.
```
dgozman, line 178: Let's link to MDN?

docs/src/api/class-keyboard.md
```diff
@@ -172,6 +172,16 @@ Modifier keys DO influence `keyboard.down`. Holding down `Shift` will type the t
...
+
+### option: Keyboard.down.text
+- `text` <[string]>
+
+Override the text produced by this event.
```
dgozman, line 183: Unfortunately, I do not understand what this means :) I think both properties need examples here.

src/client/input.ts
JoelEinbinder: Because key is duplicated as the parameter name and as an override. I one of them should probably get renamed.

docs/src/api/class-keyboard.md
JoelEinbinder: Sure

docs/src/api/class-keyboard.md
JoelEinbinder: When typing with a US keyboard layout, pressing `s` produces `{code: 'KeyS', key: 's'}` and will synchronously generate `s` in an input field assuming the key event was not canceled.

When typing with a Japanese keyboard layout, pressing `s` produces `{code: 'KeyS', key: 'と'}` and will asynchronously generate the text `と` through ime regardless of whether or not the event was canceled.

We could try to duplicate this behavior through more keyboard layout files like we do for usKeyboardLayout, but the amount of layouts is too big for us to reasonably bake into playwright. Most layouts are also significantly more complicated than the US layout. So instead this api provides a low level override.

trueadm:
I think the better approach here is not have any `overrides` and keep the keyboard APIs as they are. Instead, we can tell from `keyboard.imeSetComposition` when we're in composition, and for any key events that occur during this phase, keydown, keyup events ets have `isComposing: true` and `keyCode: 229`, plus they don't actually insert text. This would match what actually happens from researching this in all the major browsers.

pavelfeldman:
Closing as obsolete

PR #7674, opened 2021-07-16
test: mark color-scheme reset test as failing on MacOS <= 11.2

mxschmitt:
Our build bots are still on MacOS 11.0 which leads to a failing test there. Instead of upgrading them, we mark it now as failing on <= 11.2.

### DIFF
--- a/tests/page/page-emulate-media.spec.ts
+++ b/tests/page/page-emulate-media.spec.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import os from 'os';
 import { test as it, expect } from './pageTest';
 
 it.skip(({ isAndroid }) => isAndroid);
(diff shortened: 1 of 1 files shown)


pavelfeldman (commented):
It means it will regress to all our customers on GHA, do we know what broke?

mxschmitt:
> It means it will regress to all our customers on GHA, do we know what broke?

It only fails when the colorScheme get's reset to the system defaults by passing null, see the following test with the comment where it's failing. So the scope of it is very small and should not cause any issues since the main functionality is working as expected.

```ts
it('should default to light', async ({page}) => {
  expect(await page.evaluate(() => matchMedia('(prefers-color-scheme: light)').matches)).toBe(true);
  expect(await page.evaluate(() => matchMedia('(prefers-color-scheme: dark)').matches)).toBe(false);

  await page.emulateMedia({ colorScheme: 'dark' });
  expect(await page.evaluate(() => matchMedia('(prefers-color-scheme: dark)').matches)).toBe(true);
  expect(await page.evaluate(() => matchMedia('(prefers-color-scheme: light)').matches)).toBe(false);

  await page.emulateMedia({ colorScheme: null });
  expect(await page.evaluate(() => matchMedia('(prefers-color-scheme: dark)').matches)).toBe(false); // <-- it fails here
  expect(await page.evaluate(() => matchMedia('(prefers-color-scheme: light)').matches)).toBe(true);
});
```

pavelfeldman (approved):

mxschmitt:
Closing since the reason for that was that the buildbot was configured in dark-mode...

PR #7675, opened 2021-07-16
test: fix Chromium proxy tests on Windows

mxschmitt:
They were previously failing on Windows with the following message:

```
browser.newContext: Browser needs to be launched with the global proxy. If all contexts override the proxy, global proxy will be never used and can be any string, for example "launch({ proxy: { server: 'http://per-context' } })"
```

This should make them green on the flakiness dashboard.

### DIFF
--- a/tests/browsercontext-proxy.spec.ts
+++ b/tests/browsercontext-proxy.spec.ts
@@ -21,11 +21,17 @@ it.use({ proxy: { server: 'per-context' } });
 it('should throw for missing global proxy on Chromium Windows', async ({ browserName, platform, browserType, browserOptions, server }) => {
   it.skip(browserName !== 'chromium' || platform !== 'win32');
 
-  delete browserOptions.proxy;
-  const browser = await browserType.launch(browserOptions);
-  const error = await browser.newContext({ proxy: { server: `localhost:${server.PORT}` } }).catch(e => e);
-  expect(error.toString()).toContain('Browser needs to be launched with the global proxy');
-  await browser.close();
+  let browser;
+  try {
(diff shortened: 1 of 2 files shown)


pavelfeldman:
This needs a lot of try/finally or a new factory for browsers.

tests/browsercontext-proxy.spec.ts
```diff
@@ -14,9 +14,24 @@
...
-import { browserTest as it, expect } from './config/browserTest';
-
-it.use({ proxy: { server: 'per-context' } });
+import os from 'os';
+import { Browser } from '..';
+import { browserTest, expect } from './config/browserTest';
+
+const LAUNCH_OPTION_OVERRIDES = os.platform() === 'win32' ? { proxy: { server: `per-context` } } : {};
+
+const it = browserTest.extend<{ browserTestFactory: () => Promise<Browser> }>({
```
pavelfeldman, line 23: Looks like you added it twice. Also, can we start a single browser for these tests to save time? A worker-level browserWithProxy for windows that is...

While you are at it, line #40 can be fixed.

tests/browsercontext-proxy.spec.ts
```diff
@@ -46,7 +61,6 @@ it('should work when passing the proxy only on the context level', async ({brows
   const page = await context.newPage();
   await page.goto('http://non-existent.com/target.html');
   expect(await page.title()).toBe('Served by the proxy');
-  await browser.close();
```
pavelfeldman, line 49: You just leaked a browser.

dgozman (approved):

PR #7686, opened 2021-07-16
feat: expose issueTime method in network.Request

### LINKED ISSUE #7662: [Feature]: Adding blocked and queued time to the ResourceTiming
### Feature request

To understand how much time the browser spent on a single request, Playwright already exposes the timing information to some extend that would help in figuring out the time spent on `dns`, `ssl`, `connect`, and `download` events. 

To understand the timing data before the request was even fired, It would be great to have the ability to extract `blocked`, `queued` time which is not possible to obtain at the moment. Proposing to have more fields like `requestSentTime/requestIssueTime`, `responseReceived` and `responseEndTime`

https://github.com/microsoft/playwright/blob/master/src/server/network.ts#L274-L283

Usecase: Currently we directly hook on to the CDP protocol to understand the full timing information on all events including `blocked`, `queued`, `proxy` etc, Having these available in PW would make it cross browser and also helps remove the need for hooking in to CDP. 

Would it be of interest from the PW team, I am happy to work on a PR. Thoughts?

vigneshshanmugam:
+ fix https://github.com/microsoft/playwright/issues/7662
+ Added `issueTime` which is the time that the connection is issued by the browser before the request is being queued.

### DIFF
--- a/browser_patches/firefox-beta/juggler/NetworkObserver.js
+++ b/browser_patches/firefox-beta/juggler/NetworkObserver.js
@@ -662,14 +662,15 @@ class NetworkRequest {
     httpChannel.QueryInterface(Ci.nsIHttpChannelInternal);
     httpChannel.QueryInterface(Ci.nsITimedChannel);
     const timing = {
-      startTime: httpChannel.channelCreationTime,
-      domainLookupStart: httpChannel.domainLookupStartTime,
-      domainLookupEnd: httpChannel.domainLookupEndTime,
-      connectStart: httpChannel.connectStartTime,
-      secureConnectionStart: httpChannel.secureConnectionStartTime,
-      connectEnd: httpChannel.connectEndTime,
-      requestStart: httpChannel.requestStartTime,
-      responseStart: httpChannel.responseStartTime,
+      issueTime: this.httpChannel.channelCreationTime,
+      startTime: this.httpChannel.asyncOpenTime,
+      domainLookupStart: this.httpChannel.domainLookupStartTime,
+      domainLookupEnd: this.httpChannel.domainLookupEndTime,
+      connectStart: this.httpChannel.connectStartTime,
+      secureConnectionStart: this.httpChannel.secureConnectionStartTime,
+      connectEnd: this.httpChannel.connectEndTime,
+      requestStart: this.httpChannel.requestStartTime,
+      responseStart: this.httpChannel.responseStartTime,
     };
 
     const { status, statusText, headers } = responseHead(httpChannel, opt_statusCode, opt_statusText);
--- a/browser_patches/firefox-beta/juggler/NetworkObserver.js
+++ b/browser_patches/firefox-beta/juggler/NetworkObserver.js
@@ -662,15 +662,14 @@ class NetworkRequest {
     httpChannel.QueryInterface(Ci.nsIHttpChannelInternal);
     httpChannel.QueryInterface(Ci.nsITimedChannel);
     const timing = {
-      issueTime: this.httpChannel.channelCreationTime,
-      startTime: this.httpChannel.asyncOpenTime,
-      domainLookupStart: this.httpChannel.domainLookupStartTime,
-      domainLookupEnd: this.httpChannel.domainLookupEndTime,
-      connectStart: this.httpChannel.connectStartTime,
-      secureConnectionStart: this.httpChannel.secureConnectionStartTime,
-      connectEnd: this.httpChannel.connectEndTime,
-      requestStart: this.httpChannel.requestStartTime,
-      responseStart: this.httpChannel.responseStartTime,
+      startTime: httpChannel.channelCreationTime,
+      domainLookupStart: httpChannel.domainLookupStartTime,
+      domainLookupEnd: httpChannel.domainLookupEndTime,
+      connectStart: httpChannel.connectStartTime,
+      secureConnectionStart: httpChannel.secureConnectionStartTime,
+      connectEnd: httpChannel.connectEndTime,
+      requestStart: httpChannel.requestStartTime,
+      responseStart: httpChannel.responseStartTime,
     };
 
     const { status, statusText, headers } = responseHead(httpChannel, opt_statusCode, opt_statusText);
--- a/docs/src/api/class-request.md
+++ b/docs/src/api/class-request.md
@@ -183,11 +183,14 @@ following: `document`, `stylesheet`, `image`, `media`, `font`, `script`, `texttr
 
 Returns the matching [Response] object, or `null` if the response was not received due to error.
 
+## method: Request.issueTime
+- returns: <[float]>
+
+Denotes the monotonically increasing time in seconds when the request is issued by the browser.
+
 ## method: Request.timing
 - returns: <[Object]>
-  - `issueTime` <[float]> Request issued time in milliseconds elapsed since January 1, 1970 00:00:00 UTC
-  - `startTime` <[float]> Request start time in milliseconds elapsed since
-    January 1, 1970 00:00:00 UTC. Includes time spent in queue waiting for a network connection.
+  - `startTime` <[float]> Request start time in milliseconds elapsed since January 1, 1970 00:00:00 UTC
   - `domainLookupStart` <[float]> Time immediately before the browser starts the domain name lookup for the
     resource. The value is given in milliseconds relative to `startTime`, -1 if not available.
   - `domainLookupEnd` <[float]> Time immediately after the browser starts the domain name lookup for the resource.
--- a/docs/src/api/class-request.md
+++ b/docs/src/api/class-request.md
@@ -185,7 +185,9 @@ Returns the matching [Response] object, or `null` if the response was not receiv
 
 ## method: Request.timing
 - returns: <[Object]>
-  - `startTime` <[float]> Request start time in milliseconds elapsed since January 1, 1970 00:00:00 UTC
+  - `issueTime` <[float]> Request issued time in milliseconds elapsed since January 1, 1970 00:00:00 UTC
+  - `startTime` <[float]> Request start time in milliseconds elapsed since
+    January 1, 1970 00:00:00 UTC. Includes time spent in queue waiting for a network connection.
   - `domainLookupStart` <[float]> Time immediately before the browser starts the domain name lookup for the
     resource. The value is given in milliseconds relative to `startTime`, -1 if not available.
   - `domainLookupEnd` <[float]> Time immediately after the browser starts the domain name lookup for the resource.
--- a/src/server/chromium/crNetworkManager.ts
+++ b/src/server/chromium/crNetworkManager.ts
@@ -289,8 +289,10 @@ export class CRNetworkManager {
     const timingPayload = responsePayload.timing!;
     let timing: network.ResourceTiming;
     if (timingPayload) {
+      const requestSentTime = Math.min(request._timestamp, timingPayload.requestTime);
       timing = {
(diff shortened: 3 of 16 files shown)


browser_patches/firefox-beta/juggler/NetworkObserver.js
```diff
@@ -519,7 +519,8 @@ class NetworkRequest {
     this.httpChannel.QueryInterface(Ci.nsIHttpChannelInternal);
     this.httpChannel.QueryInterface(Ci.nsITimedChannel);
     const timing = {
-      startTime: this.httpChannel.channelCreationTime,
+      issueTime: this.httpChannel.channelCreationTime,
+      startTime: this.httpChannel.asyncOpenTime,
```
vigneshshanmugam, line 523: @pavelfeldman I am not sure if this is correct in FF - https://searchfox.org/mozilla-central/source/netwerk/base/nsITimedChannel.idl#96

docs/src/api/class-request.md
```diff
@@ -185,7 +185,9 @@ Returns the matching [Response] object, or `null` if the response was not receiv
 
 ## method: Request.timing
 - returns: <[Object]>
-  - `startTime` <[float]> Request start time in milliseconds elapsed since January 1, 1970 00:00:00 UTC
+  - `issueTime` <[float]> Request issued time in milliseconds elapsed since January 1, 1970 00:00:00 UTC
```
pavelfeldman, line 188: This structure corresponds to https://developer.mozilla.org/en-US/docs/Web/API/Resource_Timing_API/Using_the_Resource_Timing_API, we can leave it as is and 'fix' the startTime on Chromium / Firefox.

vigneshshanmugam:
If the `issueTime` doesn't feel like the right place, will move it to `request.requestTime` as suggested in the issue.

src/server/chromium/crNetworkManager.ts
```diff
@@ -288,8 +288,10 @@ export class CRNetworkManager {
     const timingPayload = responsePayload.timing!;
     let timing: network.ResourceTiming;
     if (timingPayload) {
+      const requestSentTime = Math.min(request._timestamp, timingPayload.requestTime);
       timing = {
-        startTime: (timingPayload.requestTime - request._timestamp + request._wallTime) * 1000,
```
pavelfeldman, line 292: `startTime` will now be `request._wallTime * 1000)`, but we need to recompute all the timings relative to it.

src/server/chromium/crNetworkManager.ts
```diff
@@ -300,6 +302,7 @@ export class CRNetworkManager {
       };
     } else {
       timing = {
+        issueTime: request._timestamp * 1000,
```
pavelfeldman, line 305: ditto

pavelfeldman (commented):
Note that you should do juggler changes in a separate PR and roll to the new Firefox build prior to landing the client changes. You can limit this first PR to Chromium / Webkit. I think WebKit will stay the same in fact.

docs/src/api/class-request.md
vigneshshanmugam: If we are not going to add `issueTime` in here, How would we derive the `blocked` time? Do you want to expose the `requestSentTime` from `request` object?

vigneshshanmugam:
> Note that you should do juggler changes in a separate PR and roll to the new Firefox build prior to landing the client changes.

Gotcha 👍🏽 Should i follow this - https://github.com/microsoft/playwright/blob/master/browser_patches/README.md#2-developing-a-new-change?

docs/src/api/class-request.md
vigneshshanmugam: Would like some inputs here - was thinking like 
```
const blocked = timings.StartTime - timings.IssueTime
```

src/server/chromium/crNetworkManager.ts
```diff
@@ -288,15 +288,19 @@ export class CRNetworkManager {
...
-        startTime: (timingPayload.requestTime - request._timestamp + request._wallTime) * 1000,
-        domainLookupStart: timingPayload.dnsStart,
-        domainLookupEnd: timingPayload.dnsEnd,
-        connectStart: timingPayload.connectStart,
-        secureConnectionStart: timingPayload.sslStart,
-        connectEnd: timingPayload.connectEnd,
-        requestStart: timingPayload.sendStart,
-        responseStart: timingPayload.receiveHeadersEnd,
+        startTime: timingPayload.requestTime,
+        domainLookupStart: calculateOffset(timingPayload.dnsStart),
```
vigneshshanmugam, line 297: Exposed timings need to be `timestamp` that denotes the event in RT API - https://w3c.github.io/resource-timing/#sec-performanceresourcetiming

I guess this is the closest we can go as far to keeping in Sync. Previous code was in millisecond which was not matching with RT spec.

src/server/supplements/har/harTracer.ts
```diff
@@ -235,20 +235,22 @@ export class HarTracer {
     const timing = response.timing();
     if (pageEntry.startedDateTime.valueOf() > timing.startTime)
       pageEntry.startedDateTime = new Date(timing.startTime);
+    const blocked = timing.startTime !== -1 ? helper.millisToRoundishMillis(timing.startTime - request.issueTime()) : -1;
```
vigneshshanmugam, line 238: Need to change this to `secondsToMillis` once we agree on the solution here - https://github.com/microsoft/playwright/pull/7686/files#r671585936

docs/src/api/class-request.md
pavelfeldman: Blocking time has a lax definition. We typically take the earliest probe: renderer request creation for subresource or browser request creation for navigation request. That's when the page decided it wants the request. After that it is queued and then some non-zero signal occurs in the network stack (present in resource timing). The delta between the first network stack number we get and the start is our queueing time.

Connection handshake (dns, ssl, etc) makes things harder because those can happen out of order per browser liking. So in the end, you will most likely want the delta between sending request and start.

Issue time does not really have any physical sense, I don't think it matters or even is defined. If there is a delta between request being sent and issue time, that's still blocking/queueing.

vigneshshanmugam: Thanks a lot for the details on how the stack works. Keeping this in mind and also the RT API, I went through the dev-tools protocol and this is version I came up with. 

`request.IssueTime()` or `request.requestSentTime()` - Denotes the timestamp when the request was queued. 

Then we get the signal from the network stack via Resource Timing API - `timing.requestTime`(CDP), so to calculate blocking/queued we could do - https://github.com/microsoft/playwright/pull/7686/files#r671585936
```ts
const blocking = request.timing().startTime - request.issueTime() 
```

Also I have adjusted the timings to be compatible with RT spec - https://github.com/microsoft/playwright/pull/7686/commits/42a9fc997d78e1098d27c3591c6d97273fa2d48e#r671585936

pavelfeldman:
Sorry for being slow on this, got distracted. Is there a test page that you are using? I'm still figuring out whether we want issue time that is different from start time or we want to fix the startTime in Chromium to include renderer-side queueing.

vigneshshanmugam:
@pavelfeldman Don't have any specific URL for testing, But trying out mostly this with PW test  `await page.goto("https://www.elastic.co/") and then trying to find the queuing time. 

Piece of code from chrome dev tools on where I was going for - https://github.com/ChromeDevTools/devtools-frontend/blob/7f5478d8ceb7586f23f4073ab5c2085dac1ec26a/front_end/network/RequestTimingView.js#L160-L162

pavelfeldman:
> Piece of code from chrome dev tools on where I was going for - https://github.com/ChromeDevTools/devtools-frontend/blob/7f5478d8ceb7586f23f4073ab5c2085dac1ec26a/front_end/network/RequestTimingView.js#L160-L162

It is quite likely that I authored that code many years ago, does not mean it is what we want here. I would create a test page to see what actually happens these days.

vigneshshanmugam:
> It is quite likely that I authored that code many years ago, does not mean it is what we want here. I would create a test page to see what actually happens these days.

Sure, Thanks much. Let me know if I can help in any way.

vigneshshanmugam:
> I would create a test page to see what actually happens these days.

Apologies for the delay, What would be the minimal test that you would be helpful to see this PR moving? Do we want to see the blocked behaviour for the network request?

vigneshshanmugam:
@pavelfeldman Are you looking for any other specific tests? Would like to get this PR in a ready state once we agree on a solution. Please do let me know what you think.

tests/resource-timing.spec.ts
```diff
@@ -26,10 +26,74 @@ it('should work', async ({ contextFactory, server }) => {
...
   verifyConnectionTimingConsistency(timing);
+  expect(timing.startTime).toBeGreaterThanOrEqual(request.issueTime());
   expect(timing.requestStart).toBeGreaterThanOrEqual(timing.connectEnd);
   expect(timing.responseStart).toBeGreaterThanOrEqual(timing.requestStart);
   expect(timing.responseEnd).toBeGreaterThanOrEqual(timing.responseStart);
-  expect(timing.responseEnd).toBeLessThan(10000);
+  await context.close();
+});
+
+it('should account for blocked and queued time', async ({ contextFactory, server }) => {
```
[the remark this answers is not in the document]
pavelfeldman: So I just ran your test w/o your modifications and I got this for connectStart:

```
0.405
-1
0.703
0.923
0.464
0.526
0.693
206.547
206.497
206.517
206.517
206.651
206.766
414.953
```

- So it looks like as far as blocking time is concerned, we are covered with the existing code, at least on Chromium.
- There is no clear distinction between blocking and queueing time, so we should decide what you want to see there. If you are interested in deferring requests by the renderer based on their priority (scripts go before images), we would need a test page for that. Otherwise all your numbers are basically zeros, a cost of an mojo plumbing calls.

pavelfeldman:
Just posted a reply above!

tests/resource-timing.spec.ts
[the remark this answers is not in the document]
vigneshshanmugam: Thanks for testing it and the details. 

> So it looks like as far as blocking time is concerned, we are covered with the existing code, at least on Chromium.

What would that look like with the existing code? duration between startTime and dns/connect start? I would like to at-least patch that in the hartracer. 


>  If you are interested in deferring requests by the renderer based on their priority (scripts go before images), we would need a test page for that.

Yes when i was referring to that for the queue time but thought of doing that later on in separate PR.

To conclude if we can get the blocking time now without exposing the issueTime, I am okay with closing down this PR.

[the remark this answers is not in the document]
pavelfeldman: > What would that look like with the existing code? duration between startTime and dns/connect start? I would like to at-least patch that in the hartracer.

Yes, I ran your test and printed `connectStart`. If you have other tests, I can run those as well. But this particular use case seems to be covered with the existing code.

> Yes when i was referring to that for the queue time but thought of doing that later on in separate PR.

Sure, we would need a different test for that.

> To conclude if we can get the blocking time now without exposing the issueTime, I am okay with closing down this PR.

That's up to you. There still are inconsistencies in other browsers, for example WebKit on a mac was unblocking not when 6 connections were closed, but when they were responded to. But I think this reflects the reality.

Since your changes were somewhat Chromium centric, I don't think they are adding to the blocking time, so yeah feel free to close.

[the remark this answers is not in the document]
vigneshshanmugam: Thanks for the details, Will close this PR and will try to incorporate changes to harTracer for blocked time.

PR #7693, opened 2021-07-17
test: rework port forwarding server tests to launch server out of process

mxschmitt:
Previously the PlaywrightServer instances were started in the same process. This lead to the problem when they closed to call the following method:
 
https://github.com/microsoft/playwright/blob/08da9d207effe7d888668a945d5a9c55e9ab21b0/src/remote/playwrightServer.ts#L44-L46

By calling that, other processes -> browsers got closed and future tests were failing which lead to

> browser.newContext: Target page, context or browser has been closed

When creating new contexts for future tests. This patch changes it to out of process instead and fixes random failing tests, including [this](https://devops.aslushnikov.com/flakiness2.html#filter_spec=resource-timing.spec.ts&commits=50&timestamp=1626519718412) one.

### DIFF
--- a/src/cli/cli.ts
+++ b/src/cli/cli.ts
@@ -243,7 +243,7 @@ if (!process.env.PW_CLI_TARGET_LANG) {
 if (process.argv[2] === 'run-driver')
   runDriver();
 else if (process.argv[2] === 'run-server')
-  runServer(process.argv[3] ? +process.argv[3] : undefined, process.argv?.[4]).catch(logErrorAndExit);
+  runServer(process.argv[3] ? +process.argv[3] : undefined, process.argv[4]).catch(logErrorAndExit);
 else if (process.argv[2] === 'print-api-json')
   printApiJson();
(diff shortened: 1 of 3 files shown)


src/cli/cli.ts
```diff
@@ -243,7 +243,7 @@ if (!process.env.PW_CLI_TARGET_LANG) {
 if (process.argv[2] === 'run-driver')
   runDriver();
 else if (process.argv[2] === 'run-server')
-  runServer(process.argv[3] ? +process.argv[3] : undefined);
+  runServer(process.argv[3] ? +process.argv[3] : undefined, process.argv?.[4]).catch(logErrorAndExit);
```
dgozman, line 246: nit: why `argv?` - could it be undefined?

dgozman (approved):

src/cli/cli.ts
mxschmitt: Good point, fixed.

PR #7696, opened 2021-07-17
fix: make generated example run with jbang

maxandersen:
I really like playwright and realised that if the java codegen would just add the following lines on top the generated examples just work if you have jbang.dev installed:

```
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS com.microsoft.playwright:playwright:RELEASE
```

jbang is a small wrapper around javac and java that can take instructions like //DEPS <gav> and compile and run source code directly.

Can even run playwright cli without setting up a maven project first:

```
jbang -m com.microsoft.playwright.CLI com.microsoft.playwright:playwright:RELEASE codegen -o Example.java
```

would love to see this in - let me know what you think.

thanks.

### DIFF
--- a/src/server/supplements/recorder/java.ts
+++ b/src/server/supplements/recorder/java.ts
@@ -128,6 +128,8 @@ export class JavaLanguageGenerator implements LanguageGenerator {
   generateHeader(options: LanguageGeneratorOptions): string {
     const formatter = new JavaScriptFormatter();
     formatter.add(`
+    ///usr/bin/env jbang "$0" "$@" ; exit $?
+    //DEPS com.microsoft.playwright:playwright:RELEASE
     import com.microsoft.playwright.*;
     import com.microsoft.playwright.options.*;
     import java.util.*;


maxandersen:
p.s. I haven't been able to run npm build/tests - they failed even before doing my changes.

aslushnikov:
@maxandersen Yury is on vacation right now, so there might be some delay reviewing this

maxandersen:
Jbang is new (2020) but trying to make it more known. The good thing is that it does not conflict with existing tooling as it's literally just encoded in java comments.

yury-s:
I'd wait for jbang to get more traction before adding it to each generated example. My main concern is that it will raise a lot of questions from the people that are not using it. At the same time for those that do use it adding the header manually should not be a problem.

yury-s:
Closing for now per the commend above, we may reconsider this later if jbang gets more widely adopted.

maxandersen:
> Closing for now per the commend above, we may reconsider this later if jbang gets more widely adopted.

Having playwright show up as something useful for the kind of workflows would help it getting more widely adopted :)  playwright 

But I understand and respect your hesitance for adding it into the default Java setup. 

Would you be open for a PR that adds a "Java w/jbang" option like you have "Python" and "Python async" variances ?

Issue #7639, opened 2021-07-15
[Bug]: Problem in runner with report: junit and shard=x/y


Belisckner:
### Playwright version

1.12.0

### Operating system

MacOS, Linux

### What browsers are you seeing the problem on?

Chromium, Firefox, WebKit

### Other information

test-runner

### What happened? / Describe the bug

When running tests by the command `playwright test --reporter=junit --shard=1/2 `, in the junit report the skipped tests are marked "passed" with time '0'.

At the same time, in gitlab ci in Test, after testing in two different threads,  the number of tests in the gitlab is more than it actually is.
![image](https://user-images.githubusercontent.com/37740578/125773766-824542be-fb36-4646-a812-90e7a7e8532e.png)

I suggest not to include tests with time '0' in the JUnit report or don't include not running tests

mxschmitt:
We recently fixed a similar bug #7568 sounds like a duplicate or related.

Belisckner:
It's related, yes.
But not the same.

in #7568 suggest marked "skip" test. But in gitlab-ci there are both variant of test. Skipped and passed (with time).
In first thread test would be "skipped"  
In second thread test would be "passed" with example time="1.234"

For gitlab, it would be different tests, but this is actually one test

dgozman:
Fixed by #7708, should be available in v1.14.

PR #7708, opened 2021-07-18
fix(test-runner): sharding tests does not show a lot of skips

dgozman:
References #7639.

### DIFF
--- a/src/test/runner.ts
+++ b/src/test/runner.ts
@@ -20,7 +20,7 @@ import rimraf from 'rimraf';
 import * as fs from 'fs';
 import * as path from 'path';
 import { promisify } from 'util';
-import { Dispatcher } from './dispatcher';
+import { Dispatcher, TestGroup } from './dispatcher';
 import { createMatcher, FilePatternFilter, monotonicTime, raceAgainstDeadline } from './util';
 import { TestCase, Suite } from './test';
 import { Loader } from './loader';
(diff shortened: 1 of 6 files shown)


src/test/runner.ts
```diff
@@ -230,6 +230,43 @@ export class Runner {
 
       await Promise.all(Array.from(outputDirs).map(outputDir => removeFolderAsync(outputDir).catch(e => {})));
 
+      let testGroups = createTestGroups(rootSuite);
+
+      const shard = config.shard;
+      if (shard) {
```
mxschmitt, line 236: maybe an own function like we have for filterByFocusedLine and filterOnly since the logic from 237 to 260 is pretty specific for shards?

```suggestion
      if (shard)
        testGroups = filterGroupByShard(shard, rootSuite)
```

src/test/runner.ts
```diff
@@ -230,6 +230,43 @@ export class Runner {
...
+              shardTests.add(test);
+          }
+          current += group.tests.length;
+        }
+
+        testGroups = shardGroups;
+        filterSuite(rootSuite, () => false, test => shardTests.has(test));
+      }
+
+      if (process.stdout.isTTY) {
```
mxschmitt, line 263: I saw this logic was also there before, but would it not be even more relevant on the bots to see if shards are actively used and which one is used exactly?

mxschmitt (approved):

src/test/runner.ts
pavelfeldman: @mxschmitt: we can only touch console in reporters, library can't touch it. In this case, we make an exception, but ensure live user.

pavelfeldman (approved):

Issue #7682, opened 2021-07-16
release-1.13: testrunner fixes


aslushnikov:
### Your question

- [x] list reporter to-be-fixed
- [x] reporter API
- [x] a big friendly message for our users about running `npx playwright install` 
    - [x] `install-deps` for the webkit
    - [x] github-playwright-action

PR #7715, opened 2021-07-19
chore: nicer message to install dependencies

aslushnikov:
If we know how to install all dependencies, then we should recommend
a Playwright CLI one-liner to install dependencies.

image

References #7682

### DIFF
--- a/src/utils/dependencies.ts
+++ b/src/utils/dependencies.ts
@@ -175,6 +175,20 @@ export async function validateDependenciesLinux(linuxLddDirectories: string[], d
     }
   }
 
+  // Happy path: known dependencies are missing for browsers.
+  // Suggest installation with a Playwright CLI.
+  if (missingPackages.size && !missingDeps.size) {
+    throw new Error('\n' + utils.wrapInASCIIBox([
+      `Host system is missing a few dependencies to run browsers.`,
+      `Please install them with the following command:`,
+      ``,
+      `    sudo npx playwright install-deps`,
+      ``,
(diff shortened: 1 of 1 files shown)


src/utils/dependencies.ts
```diff
@@ -175,6 +175,20 @@ export async function validateDependenciesLinux(linuxLddDirectories: string[], d
...
   }
 
+  // Happy path: known dependencies are missing for browsers.
+  // Suggest installation with a Playwright CLI.
+  if (missingPackages.size && !missingDeps.size) {
+    throw new Error('\n' + utils.wrapInASCIIBox([
+      `Host system is missing a few dependencies to run browsers.`,
+      `Please install them with the following command:`,
+      ``,
+      `    sudo npx playwright install-deps`,
```
mxschmitt, line 185: show sudo conditional? 

`const needsSudo = process.platform !== 'win32 && process.getuid() !== 0`

src/utils/dependencies.ts
```diff
@@ -175,6 +175,20 @@ export async function validateDependenciesLinux(linuxLddDirectories: string[], d
...
+      ``,
+      `<3 Playwright Team`,
```
mxschmitt, line 187:
```suggestion
      `<3 PlayWright Team`,
```
:D

mxschmitt (approved):

src/utils/dependencies.ts
aslushnikov: :P

PR #7721, opened 2021-07-19
[Proposal] chore: ship drivers with sha, instead of timestamp

avodovnik:
Instead of versioning our drivers with a timestamp, how about we version them with the commit sha from which they were built?

An advantage here would be, in combination with a submodule, an easy workflow for rolling and identifying API versions, etc. Basically, the roll would look at the submodule/package.json version to get the prefix, append the commit where it's at, and be able to download the correct driver.

### DIFF
--- a/utils/build/update_canary_version.js
+++ b/utils/build/update_canary_version.js
@@ -17,7 +17,7 @@
 
 const fs = require('fs');
 const path = require('path');
-const { execSync } = require('child_process');
+const { execSync, spawnSync } = require('child_process');
 
 const packageJSON = require('../../package.json');
 if (process.argv[2] === '--today-date') {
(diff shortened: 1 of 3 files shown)


utils/build/update_canary_version.js
```diff
@@ -31,6 +31,9 @@ if (process.argv[2] === '--today-date') {
     stdio: ['ignore', 'pipe', 'ignore']
   }).toString('utf8').trim();
   packageJSON.version = `${packageJSON.version}-${timestamp}000`;
+} else if (process.argv[2] === '--commit-sha') {
+  const commitSHA = spawnSync('git', ['rev-parse', 'HEAD'], {cwd: __dirname, encoding: 'utf8'}).stdout.trim();
+  packageJSON.version = `${commitSHA}`;
 } else {
   throw new Error('This script must be run with either --timestamp or --today-date parameter');
```
JoelEinbinder, line 38: Can we delete the `--commit-timestamp` code path?
```suggestion
  throw new Error('This script must be run with either --commit-sha or --today-date parameter');
```

PR #7725, opened 2021-07-19
feat(test-runner): rename Test to TestCase

dgozman:


pavelfeldman (approved):

JoelEinbinder:
I don't like TestType -> Test. It confuses "test" as a noun with "test" as a verb. I don't have a good alternative but I prefer the old way here.

Test -> TestSpec to avoid confusion is good to me, but I would prefer Test -> TestCase.

PR #7730, opened 2021-07-19
fix(test-runner): do not list tests to stdout when JSON reporter is used

mxschmitt:


tests/playwright-test/list-mode.spec.ts
```diff
@@ -30,7 +30,7 @@ test('should list tests', async ({ runInlineTest }) => {
         expect(1 + 1).toBe(2);
       });
     `
-  }, { 'list': true });
+  }, { 'list': true, 'reporter': 'line' });
```
dgozman, line 33: Let's undo here?

dgozman (approved):

PR #7734, opened 2021-07-20
fix(test-runner): allow specifying reporter from module

pavelfeldman:


src/test/loader.ts
```diff
@@ -132,7 +132,12 @@ export class Loader {
   }
 
   async loadReporter(file: string): Promise<new (arg?: any) => Reporter> {
-    let func = await this._requireOrImport(path.resolve(this._fullConfig.rootDir, file));
+    let func = '';
+    try {
+      func = await this._requireOrImport(path.resolve(this._fullConfig.rootDir, file));
```
JoelEinbinder, line 137: I think we need to be relative to the process.cwd(), not the rootDir. The config shouldn't change how my command line arguments are resolved.

src/test/loader.ts
```diff
@@ -132,7 +132,12 @@ export class Loader {
...
+    } catch (e) {
+      func = await this._requireOrImport(file);
+    }
```
JoelEinbinder, line 140:
```suggestion
    let resolvedPath = file;
    try {
     resolvedPath = require.resolve(file, { paths: [process.cwd()] });
    catch { }
    const func = await this._requireOrImport(resolvedPath);
```

dgozman:
Closing in favor of #7749.

Issue #7714, opened 2021-07-19
[Question]: Get `test() can only be called in a test file` error while I run `npx playwright test`


suedar:
### Your question

![image](https://user-images.githubusercontent.com/18492953/126142490-433ab70d-9927-4bcd-a9f0-55f4081ee08d.png)

I was running `npx playwright test` as tutorial saying.
![截屏2021-07-19 下午6 04 35](https://user-images.githubusercontent.com/18492953/126143750-9866c198-026e-4816-a3e8-9394c0bbfa4a.png)

But I got this error. Anyone know how to solve this. Thanks.

mxschmitt:
Do you have Jest or Jest-playwright installed?

suedar:
here is my package.json
``` json
{
  "name": "webs",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "test": "jest",
  },
  "dependencies": {
    "@playwright/test": "^1.12.3",
    "antd": "^4.11.2",
    "axios": "^0.18.0",
    "browser-detect": "^0.2.28",
    "classnames": "^2.2.6",
    "customize-cra": "^1.0.0",
    "dayjs": "^1.10.3",
    "emoji-regex": "^9.2.1",
    "htmr": "^0.10.1",
    "i18next": "^11.9.1",
    "jasmine-expect": "^5.0.0",
    "jest-playwright-preset": "^1.7.0",
    "js-sha256": "^0.9.0",
    "lodash": "^4.17.11",
    "mobile-detect": "^1.4.5",
    "react": "^16.8.5",
    "react-dom": "^16.8.5",
    "react-helmet-async": "^1.0.9",
    "react-i18next": "^11.8.8",
    "react-loadable": "^5.5.0",
    "react-pro-components": "^0.0.1-alpha.42",
    "react-redux": "^5.0.7",
    "react-router": "^5.2.0",
    "react-router-dom": "^5.1.7",
    "redux": "^4.0.0",
    "redux-thunk": "^2.3.0",
    "ssc-ui-react": "0.0.1-alpha.134",
    "styled-components": "^5.2.1",
    "tailwindcss": "^2.0.2",
    "ts-jest": "^27.0.3",
    "twin.macro": "^2.1.0",
    "uatu-lib": "^0.5.1",
    "uuid": "8.3.2",
    "uuidv4": "^6.2.6"
  },
  "devDependencies": {
    "@babel/core": "7.12.10",
[...]

Issue #7587, opened 2021-07-13
[Bug]: Devops Pipelines/Releases fail to run npx playwright test


wupaz:
### Playwright version

1.12.3

### Operating system

Windows

### What browsers are you seeing the problem on?

Chromium

### Other information

npm/6.14.13
node/v14.17.3 win32 x64

Azure Release : Agent.Version 2.188.4
Windows 2019

In package.json
"@playwright/test": "^1.12.3",
"playwright": "^1.12.3",

### What happened? / Describe the bug

When I try to run npx playwright test in a Pipelines/Releases I got this error 

If I run manually on same server same code (in release deploy directory) it work 

![2021-07-13_14-41-33](https://user-images.githubusercontent.com/14102653/125511324-513a8484-1b6e-4c2d-84b1-98d9abdd73e9.png)

### Code snippet to reproduce your bug

```shell
npx playwright test
```

### Relevant log output

```shell
021-07-13T17:35:20.4031005Z ##[section]Starting: Run Playwright Test 
2021-07-13T17:35:20.4155036Z ==============================================================================
2021-07-13T17:35:20.4155649Z Task         : Command Line
2021-07-13T17:35:20.4156001Z Description  : Run a command line with arguments
2021-07-13T17:35:20.4156302Z Version      : 1.1.3
2021-07-13T17:35:20.4156597Z Author       : Microsoft Corporation
2021-07-13T17:35:20.4157052Z Help         : [More Information](https://go.microsoft.com/fwlink/?LinkID=613735)
2021-07-13T17:35:20.4157535Z ==============================================================================
2021-07-13T17:35:20.4186740Z ##[debug]Working directory: 'C:\AzurePiplinesAgent_Extension\_work\r4\a\GMaster\DevDrop\test_e2e'
2021-07-13T17:35:20.4196097Z ##[debug]Fail on standard error: 'False'
2021-07-13T17:35:20.4196492Z ##[debug]Modify environment: 'False'
2021-07-13T17:35:20.4196878Z ##[debug]C:\windows\system32\cmd.exe /c "npx playwright test"
2021-07-13T17:35:20.4197250Z ##[command]npx playwright test
2021-07-13T17:35:22.5056679Z Error: Error while reading C:\AzurePiplinesAgent_Extension\_work\r4\a\GMaster\DevDrop\test_e2e\e2e\playwright\use-case\add-new-patient.uctest.spec.ts:
2021-07-13T17:35:22.5058090Z describe() can only be called in a test file
[...]

PR #7743, opened 2021-07-20
fix(test-runner): warn users who invoke the test-runner with Jest

mxschmitt:
Relates:
- https://github.com/microsoft/playwright/issues/7714
- https://github.com/microsoft/playwright/issues/7587

See here about the worker-index env var: https://jestjs.io/docs/environment-variables. Its probably the easiest way to detect if its Jest or not.

Before it lead to when using Jest on Playwright tests:

```
 FAIL  tests/foo.spec.js
  ● Test suite failed to run

    test() can only be called in a test file

      1 | const { test, expect } = require('@playwright/test');
      2 |
    > 3 | test('basic test', async ({ page }) => {
        | ^
      4 |   await page.goto('https://playwright.dev/');
      5 |   const name = await page.innerText('.navbar__title');
      6 |   expect(name).toBe('Playwright');

      at Object.<anonymous> (tests/foo.spec.js:3:1)
```

Not sure where to put, we can also put it on the top-level at the end of the file.

### DIFF
--- a/src/test/testType.ts
+++ b/src/test/testType.ts
@@ -55,6 +55,7 @@ export class TestTypeImpl {
   }
 
   private _createTest(type: 'default' | 'only', location: Location, title: string, fn: Function) {
+    throwIfRunningInsideJest();
     const suite = currentlyLoadingFileSuite();
     if (!suite)
       throw new Error(`test() can only be called in a test file`);
@@ -71,6 +72,7 @@ export class TestTypeImpl {
   }
 
   private _describe(type: 'default' | 'only', location: Location, title: string, fn: Function) {
+    throwIfRunningInsideJest();
     const suite = currentlyLoadingFileSuite();
     if (!suite)
       throw new Error(`describe() can only be called in a test file`);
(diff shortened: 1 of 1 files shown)


src/test/testType.ts
```diff
@@ -34,6 +34,9 @@ export class TestTypeImpl {
   constructor(fixtures: (FixturesWithLocation | DeclaredFixtures)[]) {
     this.fixtures = fixtures;
 
+    if (process.env.JEST_WORKER_ID)
```
dgozman, line 37: Let's maybe move this to the test() or describe() function so that I can require @playwright/test and use `chromium.launch` in jest? Seems rude to throw during require.

src/test/testType.ts
```diff
@@ -34,6 +34,9 @@ export class TestTypeImpl {
...
+      throw new Error(`Playwright Test needs to be invoked via 'npx playwright test' and excluded from Jest.`);
```
dgozman, line 38: Can we add something to the docs and point from here? This error does not tell me how to fix the issue.

src/test/testType.ts
mxschmitt: I personally think we should 100% prevent that from doing. If they want to use chromium.launch in Jest, then they should use the normal playwright* packages.

src/test/testType.ts
mxschmitt: should be better now.

JoelEinbinder:
I agree with dmitry that it's impolite to throw immediately if we detect Jest. Let's just throw a better error message if we detect Jest when we currently throw `test() can only be called in a test file`.

src/test/testType.ts
```diff
@@ -151,4 +153,14 @@ export class TestTypeImpl {
   }
 }
 
+function throwIfRunningInsideJest() {
+  if (process.env.JEST_WORKER_ID) {
+    throw new Error(
+        `Playwright Test needs to be invoked via 'npx playwright test' and excluded from Jest test runs.\n` +
+        `Creating one directory for Playwright tests and one for Jest is the recommended way of doing it.\n` +
+        `See https://playwright.dev/docs/test-intro/ for more information about Playwright Test.`,
```
dgozman, line 161: https://playwright.dev/docs/test-intro/ is now https://playwright.dev/docs/intro/

dgozman (approved):

PR #7748, opened 2021-07-20
fix(driver): do not only rely on module Node.js version

### LINKED ISSUE #7647: [Bug]: run-driver-posix.sh does not work under NixOS
### Playwright version

1.12.3

### Operating system

Linux

_No response_

### Other information

Linux distribution: [NixOS](https://nixos.org/)
Node.js version: v12.22.2

The [run-driver-posix.sh](https://github.com/microsoft/playwright/blob/07fc8cc77793aa4e014940f06e83bd29b655ab6b/utils/build/run-driver-posix.sh) shell script has a hard-coded `sh` path and uses its own `nodejs` binary, both things do not work under NixOS.

In a nutshell, NixOS does use standard paths like _/bin/_

```
~
❯ which sh
/run/current-system/sw/bin/sh

❯ ll /run/current-system/sw/bin/sh
lrwxrwxrwx 3 root root 75 Jan  1  1970 /run/current-system/sw/bin/sh -> /nix/store/zdk63pzvb0qgs5wy1rji3qj4h4kcl5hg-bash-interactive-4.4-p23/bin/sh
❯
```

`Nodejs` can not be executed because the binary contains hard-coded paths (same problem as described above).

The first problem is easily solvable instead of using _#!/bin/sh_  the right path / sh version could be picked by `env`  _#!/usr/bin/env sh_

For nodejs it would be an idea to check for the (right) nodejs version on system-level if the "normal" command failed?¿ :thinking:  

_No response_

### Relevant log output

_No response_

tuxinaut:
With this change _run-driver-posix.sh_ searches for a `Node.js` installation within the **$PATH** variable in the case the **module** `Node.js` version can't be executed (This is the case under `NixOS`)

The code is not 100% **POSIX** compatible, I hope this is not a big issue :thinking: 

Fixes #7647

### DIFF
--- a/utils/build/run-driver-posix.sh
+++ b/utils/build/run-driver-posix.sh
@@ -1,3 +1,30 @@
-#!/bin/sh
+#!/usr/bin/env sh
+# shellcheck disable=SC2164,SC3010,SC3020,SC3043
+
 SCRIPT_PATH="$(cd "$(dirname "$0")" ; pwd -P)"
-"$SCRIPT_PATH/node" "$SCRIPT_PATH/package/lib/cli/cli.js" "$@"
+
+is_node_js_available() {
+  local _full_path="${1}"
+
+  "${_full_path}" -v &>/dev/null
+  echo $?
+}
+
+# Test if the module Node.js works
+# Especially important for NixOS ;)
+nodejs="${SCRIPT_PATH}/node"
+status_code=$(is_node_js_available "${nodejs}")
+
+if [[ $status_code -ne 0 ]]; then
+  # Test for available Node.js installation within the PATH variable
+  nodejs="$(readlink -f "$(which "node" 2>/dev/null)")"
+  status_code=$(is_node_js_available "${nodejs}")
+
+  if [[ $status_code -ne 0 ]]; then
+    echo "No usable Node.js version has been found!"
+    echo "Goodbye"
+    exit 1
+  fi
+fi
+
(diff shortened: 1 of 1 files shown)


utils/build/run-driver-posix.sh
```diff
@@ -1,3 +1,30 @@
-#!/bin/sh
```
pavelfeldman, line 1: We are going from 3 line script to 30 line script for a platform we don't really support. We have no idea if this fails the moment you start browsers on nix, etc. So I would be reluctant to landing something like this.

pavelfeldman:
Closing as per above, please file a bug so that Nix users could upvote it. That way it gains priority and we can merge this.

utils/build/run-driver-posix.sh
tuxinaut: Hello @pavelfeldman :wave: 

> We are going from 3 line script to 30 line script for a platform we don't really support. 

Yea that is a fair argument.

> We have no idea if this fails the moment you start browsers on nix, etc. So I would be reluctant to landing something like this.

OK; fair enough it's "your" project if you say you can not test or support this code I will accept that and have to use my monkey patched version.

tuxinaut:
> please file a bug so that Nix users could upvote it. 

It already exists #7647

By the way, what I do not really understand is that @aslushnikov encouraged me to provide a pull request even with knowing that  `NixOS` is not one of the supported platforms :thinking:  ?¿

PR #7749, opened 2021-07-20
feat(test-runner): use require to resolve reporters

JoelEinbinder:
We import four different kinds of files:

config - has some very custom logic around folders and playwright.config.js
globalSetup/globalTeardown - Surprisingly allows relative paths without `./`
tests - custom logic with matches
reporters - updated in this patch to use require.resolve

I can update globalSetup/globalTeardown to use require.resolve. I think
it is a bug that they accept relative paths without `./`. But it would be
a breaking change, so I'll do it in another patch.

### DIFF
--- a/src/test/util.ts
+++ b/src/test/util.ts
@@ -162,3 +162,6 @@ export function formatLocation(location: Location) {
 export function errorWithFile(file: string, message: string) {
   return new Error(`${relativeFilePath(file)}: ${message}`);
 }
+
+export const builtInReporters = ['list', 'line', 'dot', 'json', 'junit', 'null'] as const;
+export type BuiltInReporters = typeof builtInReporters[number];
--- a/src/test/util.ts
+++ b/src/test/util.ts
@@ -162,6 +162,3 @@ export function formatLocation(location: Location) {
 export function errorWithFile(file: string, message: string) {
   return new Error(`${relativeFilePath(file)}: ${message}`);
 }
-
-export const builtInReporters = ['list', 'line', 'dot', 'json', 'junit', 'null'] as const;
-export type BuiltInReporters = typeof builtInReporters[number];
--- a/tests/playwright-test/reporter.spec.ts
+++ b/tests/playwright-test/reporter.spec.ts
@@ -16,6 +16,24 @@
 
 import { test, expect } from './playwright-test-fixtures';
 
+const smallReporterJS = `
+class Reporter {
+  onBegin(config, suite) {
+    console.log('\\n%%begin');
+  }
+  onTestBegin(test) {}
+  onStdOut() {}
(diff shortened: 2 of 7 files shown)


types/test.d.ts
```diff
@@ -213,7 +213,7 @@ interface ConfigBase {
    * It is possible to pass multiple reporters. A common pattern is using one terminal reporter
    * like `'line'` or `'list'`, and one file reporter like `'json'` or `'junit'`.
    */
-  reporter?: 'dot' | 'line' | 'list' | 'junit' | 'json' | 'null' | ReporterDescription[];
+  reporter?: string | ReporterDescription[];
```
dgozman, line 216: It was explicit before so that autocomplete works. Now it will not suggest builtin reporters, unfortunately.

What if we force `reporter: ['my-reporter-module']` for third-party reporters instead?

tests/playwright-test/reporter.spec.ts
```diff
@@ -91,3 +109,48 @@ test('should work with custom reporter', async ({ runInlineTest }) => {
     '%%reporter-end-end%%',
   ]);
 });
+
+
+test.only('should work without a file extension', async ({ runInlineTest }) => {
```
dgozman, line 114: Drop `test.only`

tests/playwright-test/reporter.spec.ts
```diff
@@ -91,3 +109,48 @@ test('should work with custom reporter', async ({ runInlineTest }) => {
...
+  }, { reporter: '', workers: 1 });
+
+  expect(result.exitCode).toBe(0);
+  expect(result.output.split('\n').filter(line => line.startsWith('%%'))).toEqual([
+    '%%begin',
+    '%%end',
+  ]);
+});
+
+test('should work load from node_modules', async ({ runInlineTest }) => {
```
dgozman, line 136: nit: 'should load reporter from node_modules'

src/test/util.ts
```diff
@@ -162,3 +162,6 @@ export function formatLocation(location: Location) {
 export function errorWithFile(file: string, message: string) {
   return new Error(`${relativeFilePath(file)}: ${message}`);
 }
+
+export const builtInReporters = ['list', 'line', 'dot', 'json', 'junit', 'null'] as const;
+export type BuiltInReporters = typeof builtInReporters[number];
```
dgozman, line 167: - nit: BulitInReporters -> BuiltInReporter?
- I'd move this to runner or cli, doesn't look like "util".

types/test.d.ts
JoelEinbinder: I brought back the autocomplete.

dgozman (approved):

PR #7752, opened 2021-07-20
feat(test-runner): use require.resolve for globalSetup and globalTeardown

JoelEinbinder:
This patch is done as to not be a breaking change.

### DIFF
--- a/src/test/loader.ts
+++ b/src/test/loader.ts
@@ -22,6 +22,7 @@ import { Suite } from './test';
 import { SerializedLoaderData } from './ipc';
 import * as path from 'path';
 import * as url from 'url';
+import * as fs from 'fs';
 import { ProjectImpl } from './project';
 import { Reporter } from '../../types/testReporter';
 import { LaunchConfig } from '../../types/test';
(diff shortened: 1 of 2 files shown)


src/test/loader.ts
```diff
@@ -437,3 +438,10 @@ const baseFullConfig: FullConfig = {
   workers: 1,
   launch: [],
 };
+
+function resolveScript(id: string, rootDir: string) {
+  const localPath = path.resolve(rootDir, id);
+  if (fs.existsSync(localPath))
+    return localPath;
+  return require.resolve(id, { paths: [rootDir] });
```
dgozman, line 446: So, this is going to work for `globalSetup: 'playwright-dev-server'` where you use a node module as a global setup?

tests/playwright-test/global-setup.spec.ts
```diff
@@ -21,7 +21,7 @@ test('globalSetup and globalTeardown should work', async ({ runInlineTest }) =>
     'playwright.config.ts': `
       import * as path from 'path';
       module.exports = {
-        globalSetup: 'globalSetup.ts',
+        globalSetup: './globalSetup',
```
dgozman, line 24: Since we preserve the old mode, let's leave a test that uses it without a relative path.

dgozman (approved):

tests/playwright-test/global-setup.spec.ts
JoelEinbinder: there are some down below that i left unchanged.

src/test/loader.ts
JoelEinbinder: yes

JoelEinbinder: added a test to confirm

PR #7753, opened 2021-07-20
feat(test runner): help when describe() is misused

dgozman:


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "playwright-internal",
-  "version": "1.13.0-next",
+  "version": "1.14.0-next",
```
JoelEinbinder, line 3: This is a roll or something?

src/test/testType.ts
```diff
@@ -91,11 +101,20 @@ export class TestTypeImpl {
...
     if (!suite)
-      throw new Error(`${name} hook can only be called in a test file`);
+      throw errorWithLocation(location, `${name} hook can only be called in a test file`);
     suite._hooks.push({ type: name, fn, location });
   }
 
   private _modifier(type: 'skip' | 'fail' | 'fixme' | 'slow', location: Location, ...modifierArgs: [arg?: any | Function, description?: string]) {
+    if (typeof modifierArgs[1] === 'function') {
+      throw errorWithLocation(location, [
+        `It looks like you are calling test.${type}() to declare a test. Call it inside the test instead:`,
```
JoelEinbinder, line 111: can we add todo and tell them to use todo? Skip inside the test is a bad pattern because it runs all my custom fixtures.

package-lock.json
dgozman: I guess someone updated package.json but not package-lock.json

src/test/testType.ts
dgozman: We can, but until we did it, I'd like to help our users.

package-lock.json
JoelEinbinder: https://github.com/microsoft/playwright/pull/7754

src/test/testType.ts
JoelEinbinder: https://github.com/microsoft/playwright/pull/7755

pavelfeldman (approved):

PR #7766, opened 2021-07-21
chore: support having a CLI target language for drivers

### LINKED ISSUE #7741: [Bug]: Playwright executable path not found message should be cross-language compatible
Problem: We currently show on Python etc. npx playwright install when the executables are not installed.

https://github.com/microsoft/playwright/blob/b56e2b713a6ec4b95faaa9d6e3ced9bb538c4c2e/src/utils/registry.ts#L252-L259

Needs protocol changes. Since we should notify the server early on which language we are. Either when we launch or better before. Currently its done on newContext.

Environment variables are not the best there, since if we do it over WS, then its not working.

mxschmitt:
I thought about the following cases which this change should cover. Python is used as an example language and "ffmpeg not installed" as an example when trying to record video.

1. Using Python and doing playwright.chromium.launch
  - We pass `python` over to the driver via an env var
  - Then the driver prints `playwright install` since its Python -> Python specific

2. Using Python and doing playwright.chromium.connect
  - The server can only be in Node.js (if we add launchServer support to the languages, we can also pass it via the env)
  - In `src/browserServerImpl.ts` we don't pass a custom language, so it fallbacks to JavaScript
  - Then the driver prints `npx playwright install ffmpeg`.
  - The reasoning behind this is that the host can be different, 10km far away, so it does not bring value if we promote Python (a different language) there

3. Using PlaywrightServer/Client
  - The server again can only be JS, so I'm printing JS CLI commands
  - Again it can be 10km far away
 
Side notes:

- I renamed sdkLanguage to cliLanguage, since its not the sdkLanguage anymore.
- I didn't implement sdkLanguage over WS http headers which would be the right way to do it, since we currently have no use-case for it.


pavelfeldman:
I'm not sure I agree with the spirit of this change. I would expect us to be interested in the `sdkLanguage` on the server side in general. Before reviewer needs to review a 25-file change, let's make sure we discussed the problem and potential solutions. It might be we arrive at yours, but maybe we won't.

PR #7779, opened 2021-07-21
feat(test runner): createContext fixture for multi-context scenarios

### LINKED ISSUE #7768: [Question]: Testing application with 1 host and 2 participants
### Your question

I would like to test the following scenario, but i am not sure if i can use playwright. I would be more than happy if anyone could help me out.

- I have one host, who is creating a meetingroom with several questions (the questions must be answered mutually by the participants)
- I need at least two participants, who can join the meetingroom in their own browser tabs / instances in order to answer the question 
- After the two participants gave an answer the host needs to click on "next question", the two participants need to answer these questions as well and so on

example:

Browser/Tab 1: Host clicks on "next question"
Browser/Tab 2: Participant A answer question no.1 and clicks on "ready"
Browser/Tab 3: Participants B answer question no.1 and clicks on "ready"
Browser/Tab 1: Host clicks on "next question"
Browser/Tab 2: Participants A answer question no 2 and clicks on "ready"
Browser/Tab 3: Participants B answer question no 2 and clicks on "ready"
and so on ... 

Is this kind of scenario possible to solve with playwright, in which each browser tab / instances need to be controlled separately? Thanks a lot

dgozman:
Fixes #7768.

### DIFF
--- a/docs/src/test-api/class-fixtures.md
+++ b/docs/src/test-api/class-fixtures.md
@@ -130,6 +130,66 @@ Learn how to [configure context](./test-configuration.md) through other fixtures
 
 The [`property: Fixtures.page`] belongs to this context.
 
+## property: Fixtures.createContext
+- type: <[function]\([BrowserContextOptions]|[void]\):[BrowserContext]>
+
+A function that creates a new context, taking into account all options set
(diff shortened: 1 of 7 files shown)


pavelfeldman (approved):

mxschmitt (approved):
at some point I was thinking if it makes sense to expose filled `contextOptions` and `browserOptions` fixture like we do internally. Because currently contextOptions is an empty object and does not contain the values from the use config fixtures.

PR #7784, opened 2021-07-22
docs(pyinstaller): added docs for pyinstaller

### LINKED ISSUE #513: [Feature]: Make it compatible with Pyinstaller
Hi. 
I'm bundling my app using PyInstaller.
But I'm facing an issue.
When I use the arg: --console. it works well, the browser opens itself when I call it....
But when I use the arg: --windowed. I'm not getting error (no timeout, that's why) but so we are sure that it finds the browser, but the "headless" mode to False isn't working. to explains briefly, I open the browser to solve captchas. so I need it to be "headless=False" but even If I use this arg, it doesn't shows up, without getting any error (if I set timeout to 3000 It will raise error of selector not found)

kumaraditya303:
Fixes https://github.com/microsoft/playwright-python/issues/513

### DIFF
--- a/docs/src/intro-python.md
+++ b/docs/src/intro-python.md
@@ -4,6 +4,7 @@ title: "Getting Started"
 ---
 
 <!-- TOC -->
+
 - [Release notes](./release-notes.md)
 
 ## Installation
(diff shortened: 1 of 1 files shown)


docs/src/intro-python.md
```diff
@@ -135,6 +136,52 @@ python -m asyncio
...
+```
+
+If you do not want to bundle browsers with the executables:
+
+```bash
+playwright install
+pyinstaller -F main.py
+```
+
+If you want to bundle browsers with the executables:
```
mxschmitt, line 162:
```suggestion
```

I would only tell them how to bundle with browsers.

docs/src/intro-python.md
```diff
@@ -135,6 +136,52 @@ python -m asyncio
...
+
+```bash
+# Linux/macOS
+PLAYWRIGHT_BROWSERS_PATH=0 playwright install
```
mxschmitt, line 166:
```suggestion
PLAYWRIGHT_BROWSERS_PATH=0 playwright install chromium
```
I would pass as an example a single browser name there.

mxschmitt (commented):
Looks good in general, lets make it a little bit more lightweight and then it should be good to go.

docs/src/intro-python.md
```diff
@@ -135,6 +136,45 @@ python -m asyncio
...
+If you want to bundle browsers with the executables:
+
+```bash
+# Linux/macOS
+PLAYWRIGHT_BROWSERS_PATH=0 playwright install chromium
+pyinstaller -F main.py
+
+# Windows with cmd.exe
+set PLAYWRIGHT_BROWSERS_PATH=0
+playwright install
```
mxschmitt, line 164:
```suggestion
playwright install chromium
```

docs/src/intro-python.md
```diff
@@ -135,6 +136,45 @@ python -m asyncio
...
+pyinstaller -F main.py
+
+# Windows with PowerShell
+$env:PLAYWRIGHT_BROWSERS_PATH="0"
+playwright install
```
mxschmitt, line 169:
```suggestion
playwright install chromium
```

mxschmitt (approved):

aslushnikov (approved):

PR #7801, opened 2021-07-22
docs(test-runner): fix option fixture typing issue

### LINKED ISSUE #7787: [docs]: option fixtures via the config lead to TS warnings
https://playwright.dev/docs/test-advanced

In the example PlaywrightTestConfig should be used as a generic and the workerFixtures type should be shared between the fixture and the config.

mxschmitt:
Fixes #7787

For JSDoc, we could probably also define an own type, but probably too much complexity.

### DIFF
--- a/docs/src/test-advanced.md
+++ b/docs/src/test-advanced.md
@@ -316,7 +316,12 @@ exports.test = base.test.extend({
 import { test as base } from '@playwright/test';
 import { startBackend } from './my-backend';
 
-export const test = base.extend<{ version: string, backendUrl: string }>({
+export type TestOptions = {
+  version: string;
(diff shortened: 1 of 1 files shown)


docs/src/test-advanced.md
```diff
@@ -316,7 +316,12 @@ exports.test = base.test.extend({
 import { test as base } from '@playwright/test';
 import { startBackend } from './my-backend';
 
-export const test = base.extend<{ version: string, backendUrl: string }>({
+export type TestOptions = {
+  version: string;
+  backendUrl: string;
+}
```
dgozman, line 322: Missing semicolon

dgozman (approved):

PR #7809, opened 2021-07-23
chore(typescript): update to 4.3

JoelEinbinder:


package.json
```diff
@@ -40,23 +40,23 @@
...
+    "@babel/plugin-proposal-private-methods": "^7.14.5",
+    "@babel/plugin-proposal-private-property-in-object": "^7.14.5",
     "@babel/plugin-syntax-async-generators": "^7.8.4",
     "@babel/plugin-syntax-json-strings": "^7.8.3",
     "@babel/plugin-syntax-object-rest-spread": "^7.8.3",
     "@babel/plugin-syntax-optional-catch-binding": "^7.8.3",
-    "@babel/plugin-transform-modules-commonjs": "^7.14.0",
-    "@babel/preset-typescript": "^7.13.0",
+    "@babel/plugin-transform-modules-commonjs": "^7.14.5",
+    "@babel/preset-typescript": "^7.14.5",
```
mxschmitt, line 59: is this still there because it is used inside the test-runner? (Might make sense to use the @babel/plugin-transform-typescript directly in both? But I might also be wrong.)

utils/generate_types/test/test.ts
```diff
@@ -572,15 +572,6 @@ playwright.chromium.launch().then(async browser => {
   await browser.close();
 })();
 
-(async () => {
-  const browser = await playwright.firefox.launch();
-  const page = await browser.newPage();
-  const context = page.context();
-  const oneTwoThree = ('pageTarget' in context) ? context['pageTarget'] : 123;
```
mxschmitt, line 579: what was the reason that this got removed?

mxschmitt (approved):

utils/generate_types/test/test.ts
JoelEinbinder: Yeah the new typescript doesn't figure this one out, but it wasn't a good test in the first place.

Our typescript tester has been using new typescript instead of old typescript since we upgraded to npm 7. But it will take a separate large patch to fix that.

package.json
JoelEinbinder: yes its still in the test runner and our webpacks. I will deal with them separately. If we want to add support for declarations to the default babel, it will technically be a breaking change. I am thinking about it.

PR #7810, opened 2021-07-23
chore(typescript): turn on noImplicitOverride

JoelEinbinder:
TypeScript added a feature to force all overrides to be marked as such. We use to do this in the DevTools days with `@override`. We can land it if we like how it looks.

### DIFF
--- a/src/test/reporters/dot.ts
+++ b/src/test/reporters/dot.ts
@@ -21,19 +21,19 @@ import { FullResult, TestCase, TestResult } from '../../../types/testReporter';
 class DotReporter extends BaseReporter {
   private _counter = 0;
 
-  onStdOut(chunk: string | Buffer, test?: TestCase, result?: TestResult) {
+  override onStdOut(chunk: string | Buffer, test?: TestCase, result?: TestResult) {
     super.onStdOut(chunk, test, result);
     if (!this.config.quiet)
       process.stdout.write(chunk);
   }
 
-  onStdErr(chunk: string | Buffer, test?: TestCase, result?: TestResult) {
+  override onStdErr(chunk: string | Buffer, test?: TestCase, result?: TestResult) {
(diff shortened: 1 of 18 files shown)


dgozman:
What's the reason for this one? Also, why does it update all our deps?

JoelEinbinder:
> What's the reason for this one? Also, why does it update all our deps?

Sorry I should've marked it as a draft because it depended on the the typescript roll. It should look a lot cleaner now. Added a description.

dgozman (commented):
Seems ok, but also not too useful. Let's check what others think.

src/test/reporters/dot.ts
```diff
@@ -21,7 +21,7 @@ import { FullResult, TestCase, TestResult } from '../../../types/testReporter';
 class DotReporter extends BaseReporter {
   private _counter = 0;
 
-  onTestEnd(test: TestCase, result: TestResult) {
+  override onTestEnd(test: TestCase, result: TestResult) {
```
mxschmitt, line 24: maybe we should say more "implements" instead of "extends"? (other reporters too)

mxschmitt (approved):
(VSCode does it too: https://github.com/microsoft/vscode/blob/b27a33cc43fa1e1b66de54ef105d1e541189f861/src/tsconfig.base.json#L7)

JoelEinbinder:
We have some code in utils.js where we override `_write`. The `override` keyword here makes it a lot more readable to me. 

```ts
class HashStream extends stream.Writable {
  private _hash = crypto.createHash('sha1');

  override _write(chunk: Buffer, encoding: string, done: () => void) {
    this._hash.update(chunk);
    done();
  }

  digest(): string {
    return this._hash.digest('hex');
  }
}
```

We can always turn it off if people don't like it.

PR #7813, opened 2021-07-23
feat(test runner): add baseURL option to the CLI

elaichenkov:
Added:
* `baseURL` option for the test runner cli

Updated:
* documentation regarding changes


src/test/cli.ts
```diff
@@ -98,6 +99,8 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
   const overrides = overridesFromOptions(opts);
   if (opts.headed)
     overrides.use = { headless: false };
+  if (opts.baseURL)
+    overrides.use = { baseURL: opts.baseURL };
```
pavelfeldman, line 103: Looks like headed above will get real upset if this code runs.

elaichenkov: Yeah, you're right. I've updated it to fix it.

mxschmitt:
Closing since we currently don't want to add another specific flag for that. If you need to make it depended from some external source, we recommend to use an environment variable for it. If this is not working for you, please open a feature request so people can upvote it and we can prioritise it.

We still appreciate the contribution!

PR #7871, opened 2021-07-27
feat(expect): implement toMatchText

pavelfeldman:


dgozman (approved):

src/test/matchers/toMatchText.ts
```diff
@@ -0,0 +1,135 @@
...
+    await pollUntilDeadline(async () => {
+      received = options?.useInnerText ? await locator.innerText() : await locator.textContent() || '';
+      if (exactMatch)
+        pass = expected === received;
+      else
+        pass = typeof expected === 'string' ? received.includes(expected) : new RegExp(expected).test(received);
+      return pass === !matcherOptions.isNot;
+    }, deadline, 100);
+  } catch (e) {
+    pass = false;
```
dgozman, line 78: Why swallow the exception here?

dgozman (approved):

src/test/matchers/toMatchText.ts
pavelfeldman: because we want to show a nice error message with expected and actual text values.

PR #7872, opened 2021-07-28
Added Video link to

alethcoe:
38 minute playwright/playwright test examples and how tos


mxschmitt:
Thank you for your contribution, we might rework the showcase section in the future and add there a video section. In the meantime I would close it.

PR #7875, opened 2021-07-28
docs: remove unused installation.md

mxschmitt:
On playwright.dev we redirect to `docs/library/` instead and keeping it results in a weird search state of un-referenced pages: https://playwright.dev/docs/installation


docs/src/installation.md
```diff
@@ -1,233 +0,0 @@
----
```
yury-s, line 1: How come this is unused? IIUC https://playwright.dev/python/docs/next/installation/ will break after this PR.

PR #7879, opened 2021-07-28
feat(test): write missing snapshots to `test-results`

aslushnikov:
By default, CI's are configured to upload `//test-results` folder to
artifacts storage. Storing missing snapshots in the `//test-results`
folder will simplify collecting new baselines for other platforms from the CI.

### DIFF
--- a/src/test/matchers/golden.ts
+++ b/src/test/matchers/golden.ts
@@ -90,6 +90,10 @@ export function compare(
   options?: { threshold?: number }
 ): { pass: boolean; message?: string; expectedPath?: string, actualPath?: string, diffPath?: string, mimeType?: string } {
   const snapshotFile = snapshotPath(name);
+  const outputFile = outputPath(name);
+  const expectedPath = addSuffix(outputFile, '-expected');
+  const actualPath = addSuffix(outputFile, '-actual');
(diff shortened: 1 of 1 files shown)


mxschmitt:
Out of curiosity, which CI's do that? (If I saw it correctly, only Jenkins?)

aslushnikov:
> Out of curiosity, which CI's do that? (If I saw it correctly, only Jenkins?)

We ourselves configure github actions to upload `//test-results` and nothing else, and that's what our clients do too. Without this, it's very hard to upload new snapshots from CI

src/test/matchers/golden.ts
```diff
@@ -101,6 +105,7 @@ export function compare(
     if (isWriteMissingMode) {
       fs.mkdirSync(path.dirname(snapshotFile), { recursive: true });
       fs.writeFileSync(snapshotFile, actual);
+      fs.writeFileSync(actualPath, actual);
```
dgozman, line 108: Let's add the actualPath to the `writing actual` message below so that I know where to look at?

dgozman (approved):

src/test/matchers/golden.ts
aslushnikov: Discussed offline and decided to keep as-is.

PR #7885, opened 2021-07-28
chore: enable linting with ESLint for TSX files

mxschmitt:


src/web/traceViewer/ui/snapshotTab.tsx
```diff
@@ -27,16 +27,18 @@ export const SnapshotTab: React.FunctionComponent<{
...
 
   const snapshotMap = new Map<string, { title: string, snapshotName: string }>();
   for (const snapshot of action?.metadata.snapshots || [])
     snapshotMap.set(snapshot.title, snapshot);
   const actionSnapshot = snapshotMap.get('action') || snapshotMap.get('after');
   const snapshots = [actionSnapshot ? { ...actionSnapshot, title: 'action' } : undefined, snapshotMap.get('before'), snapshotMap.get('after')].filter(Boolean) as { title: string, snapshotName: string }[];
 
-  if (snapshotIndex >= snapshots.length)
-    snapshotIndex = snapshots.length - 1;
+  React.useEffect(() => {
```
mxschmitt, line 38: we could also move this into the parameter of useState so its used as a default value. So we could get rid of one re-render.

yury-s (approved):

PR #7902, opened 2021-07-29
feat(expect): even more matchers

pavelfeldman:


src/test/expect.ts
```diff
@@ -28,12 +28,14 @@ import {
   toContainText,
   toHaveAttr,
   toHaveClass,
+  toHaveCount,
```
pavelfeldman, line 31: toHaveLength is sync

src/test/matchers/matchers.ts
```diff
@@ -65,178 +64,212 @@ export async function toBeEmpty(
...
     return await locator.getAttribute(name, { timeout }) || '';
   }, expected, options);
 }
 
-export async function toHaveClass(
+export function toHaveClass(
   this: ReturnType<Expect['getState']>,
   locator: Locator,
-  expected: string,
+  expected: string | RegExp | string[],
```
pavelfeldman, line 149: (string | RegExp)[] is substantial amount of work...

dgozman (approved):

PR #7906, opened 2021-07-29
feat(test-runner): re-enable web server

mxschmitt:


pavelfeldman:
I think we agreed to make it more conservative. Here is what Joel concluded it with: "removing baseURL and strict makes the feature reasonable to me". I.e. leave it as `launch` and remove the features we are unsure about.

types/test.d.ts
```diff
@@ -479,24 +479,28 @@ export interface Project<TestArgs = {}, WorkerArgs = {}> extends TestProject {
...
-export type LaunchConfig = {
+export type WebServerConfig = {
   /**
    * Shell command to start. For example `npm run start`.
    */
   command: string,
+  /**
+   * The port that your http server is using. Gets used for `setBaseURL` and `waitForPort`.
+   */
+  port?: number,
```
JoelEinbinder, line 490:
```suggestion
  port: number,
```

types/test.d.ts
```diff
@@ -479,24 +479,28 @@ export interface Project<TestArgs = {}, WorkerArgs = {}> extends TestProject {
...
   /**
    * The port that your http server is expected to appear on. If specified it does wait until it accepts connections.
    */
-  waitForPort?: number,
+  waitForPort?: boolean,
```
JoelEinbinder, line 494:
```suggestion
```

types/test.d.ts
```diff
@@ -505,6 +509,10 @@ export type LaunchConfig = {
    * Current working directory of the spawned process. Default is process.cwd().
    */
   cwd?: string,
+  /**
+   * Sets the http://localhost:{port} as a Playwright BaseURL on every context.
+   */
+  setBaseURL?: boolean,
```
JoelEinbinder, line 515:
```suggestion
```

types/test.d.ts
```diff
@@ -690,7 +698,7 @@ interface TestConfig {
    * Learn more about [snapshots](https://playwright.dev/docs/test-snapshots).
    */
   updateSnapshots?: UpdateSnapshots;
-  _launch?: LaunchConfig | LaunchConfig[];
+  webServer?: WebServerConfig | WebServerConfig[];
```
JoelEinbinder, line 701:
```suggestion
  webServer?: WebServerConfig;
```

JoelEinbinder:
> I think we agreed to make it more conservative. Here is what Joel concluded it with: "removing baseURL and strict makes the feature reasonable to me". I.e. leave it as `launch` and remove the features we are unsure about.

Discussed with Max offline:

`webServer` and `launch` are different features which solve different problems.

Today, launching a web server to go with the tests looks something like this:

1. start the process in global setup
2. wait for something to start listening on the port the server is expected on
3. return a teardown function to kill the process

This is ok, but most users don't realize they need to do step 2 and get flaky setups. So the first thing about `webServer` is that the port is mandatory to help users avoid this problem.

Then we are seeing people add extra features to their webServer setup. A common one I've seen is using an environment variable to not use a local web server and instead run tests against production. Another one that Max wants is to be able to run against an existing local web server. Some servers pick a random port and print it out, so in the future we could accept a regex instead of a port to find which one was picked.

Prior art: [jest-dev-server](https://github.com/smooth-code/jest-puppeteer/tree/master/packages/jest-dev-server)

docs/src/test-advanced-js.md
```diff
@@ -43,6 +43,7 @@ These options would be typically different between local development and CI oper
 - `shard: { total: number, current: number } | null` - [Shard](./test-parallel.md#shards) information.
 - `updateSnapshots: boolean` - Whether to update expected snapshots with the actual results produced by the test run.
 - `workers: number` - The maximum number of concurrent worker processes to use for parallelizing tests.
+- `webServer: { command: string, port: number, timeout?: number, reuseExistingServer?: boolean, cwd?: string, env?: object }` - Launch a process and wait that its ready before the tests will start. See [launch web server](#launching-a-development-web-server-during-the-tests) configuration for examples.
```
JoelEinbinder, line 46:
```suggestion
- `webServer: { command: string, port: number, timeout?: number, reuseExistingServer?: boolean, cwd?: string, env?: object }` - Launch a process and wait that it's ready before the tests will start. See [launch web server](#launching-a-development-web-server-during-the-tests) configuration for examples.
```

docs/src/test-advanced-js.md
```diff
@@ -200,6 +201,73 @@ export const test = base.extend<{ saveLogs: void }>({
...
+
+You can specify a port via `port` or additional environment variables, see [here](#configuration-object). The server will wait for it to be available before running the tests. For continuous integration, you may want to use the `reuseExistingServer: !process.env.CI` option which does not use an existing server on the CI.
+
+The port gets then passed over to Playwright as a [`param: baseURL`] when creating the context [`method: Browser.newContext`].
+
+```js js-flavor=ts
+// playwright.config.ts
+import { PlaywrightTestConfig } from '@playwright/test';
+const config: PlaywrightTestConfig = {
+  launch: {
```
JoelEinbinder, line 216:
```suggestion
  webServer: {
```

docs/src/test-advanced-js.md
```diff
@@ -200,6 +201,73 @@ export const test = base.extend<{ saveLogs: void }>({
...
+};
+export default config;
+```
+
+```js js-flavor=js
+// playwright.config.js
+// @ts-check
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+const config = {
+  launch: {
```
JoelEinbinder, line 231:
```suggestion
  webServer: {
```

docs/src/test-configuration-js.md
```diff
@@ -478,6 +478,7 @@ In addition to configuring [Browser] or [BrowserContext], videos or screenshots,
 - `testIgnore`: Glob patterns or regular expressions that should be ignored when looking for the test files. For example, `'**/test-assets'`.
 - `testMatch`: Glob patterns or regular expressions that match test files. For example, `'**/todo-tests/*.spec.ts'`. By default, Playwright Test runs `.*(test|spec)\.(js|ts|mjs)` files.
 - `timeout`: Time in milliseconds given to each test.
+- `webServer: { command: string, port: number, timeout?: number, reuseExistingServer?: boolean, cwd?: string, env?: object }` - Launch a process and wait that its ready before the tests will start. See [launch web server](./test-advanced.md#launching-a-development-web-server-during-the-tests) configuration for examples.
```
JoelEinbinder, line 481:
```suggestion
- `webServer: { command: string, port: number, timeout?: number, reuseExistingServer?: boolean, cwd?: string, env?: object }` - Launch a process and wait that it's ready before the tests will start. See [launch web server](./test-advanced.md#launching-a-development-web-server-during-the-tests) configuration for examples.
```

types/test.d.ts
```diff
@@ -479,24 +479,26 @@ export interface Project<TestArgs = {}, WorkerArgs = {}> extends TestProject {
...
    * How long to wait for the process to start up and be available in milliseconds. Defaults to 60000.
    */
-  waitForPortTimeout?: number,
+  timeout?: number,
   /**
-   * If true it will verify that the given port via `waitForPort` is available and throw otherwise.
-   * This should commonly set to !!process.env.CI to allow the local dev server when running tests locally.
+   * If true, it will re-use an existing server on the port when available. If no server is running
+   * on that port, it will run the command to start a new server.
+   * If false, it will throw when a port is running on the port.
```
JoelEinbinder, line 498:
```suggestion
   * If false, it will throw if an existing process is listening on the port.
```

utils/generate_types/overrides-test.d.ts
```diff
@@ -62,24 +62,26 @@ export interface Project<TestArgs = {}, WorkerArgs = {}> extends TestProject {
...
    * How long to wait for the process to start up and be available in milliseconds. Defaults to 60000.
    */
-  waitForPortTimeout?: number,
+  timeout?: number,
   /**
-   * If true it will verify that the given port via `waitForPort` is available and throw otherwise.
-   * This should commonly set to !!process.env.CI to allow the local dev server when running tests locally.
+   * If true, it will re-use an existing server on the port when available. If no server is running
+   * on that port, it will run the command to start a new server.
+   * If false, it will throw when a port is running on the port.
```
mxschmitt, line 81:
```suggestion
   * If false, it will throw if an existing process is listening on the port.
```

src/test/webServer.ts
```diff
@@ -33,19 +33,19 @@ const newProcessLogPrefixer = () => new stream.Transform({
   },
 });
 
-class LaunchServer {
+export class WebServer {
```
JoelEinbinder, line 36: This doesn't need to be exported right?
```suggestion
class WebServer {
```

src/test/webServer.ts
```diff
@@ -85,19 +84,17 @@ class LaunchServer {
...
-    if (this.config.waitForPort) {
-      await this._waitForAvailability(this.config.waitForPort);
-      const baseURL = `http://localhost:${this.config.waitForPort}`;
-      process.env.PLAYWRIGHT_TEST_BASE_URL = baseURL;
-      console.log(`Using baseURL '${baseURL}' from config.launch.`);
-    }
+    await this._waitForAvailability();
+    const baseURL = `http://localhost:${this.config.port}`;
+    process.env.PLAYWRIGHT_TEST_BASE_URL = baseURL;
+    console.log(`Using baseURL '${baseURL}' from config.launch.`);
```
JoelEinbinder, line 90: Let's not log here?
```suggestion
```

mxschmitt: Either not log or change it to: `console.log(`Using baseURL '${baseURL}' from config.webServer.`);` Both fine for me.

src/test/webServer.ts
```diff
@@ -54,12 +54,11 @@ class LaunchServer {
...
-        throw new Error(`Port ${this.config.waitForPort} is used, make sure that nothing is running on the port or set strict:false in config.launch.`);
-      if (portIsUsed)
+    const portIsUsed = !await canBindPort(this.config.port);
+    if (portIsUsed) {
+      if (this.config.reuseExistingServer)
         return;
+      throw new Error(`Port ${this.config.port} is used, make sure that nothing is running on the port or set strict:false in config.launch.`);
     }
 
     console.log(`Launching '${this.config.command}'...`);
```
JoelEinbinder, line 64: Let's not log until somebody complains. I don't want this polluting my logs every time I run my tests.

```suggestion
```

JoelEinbinder (approved):

src/test/webServer.ts
JoelEinbinder: No log please.

PR #7923, opened 2021-07-29
fix(test runner): make test order stable when fixtures are changing

dgozman:
We used to sort based on workerHash, and that changes depending on the exact worker fixtures list. Now we replace workerHash with an ordinal when constructing the TestGroup list to preserve the natural order.

### DIFF
--- a/src/test/runner.ts
+++ b/src/test/runner.ts
@@ -439,10 +439,27 @@ function buildItemLocation(rootDir: string, testOrSuite: Suite | TestCase) {
 }
 
 function createTestGroups(rootSuite: Suite): TestGroup[] {
-  const groupById = new Map<string, TestGroup>();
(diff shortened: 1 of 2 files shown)


src/test/runner.ts
```diff
@@ -439,10 +439,27 @@ function buildItemLocation(rootDir: string, testOrSuite: Suite | TestCase) {
 }
 
 function createTestGroups(rootSuite: Suite): TestGroup[] {
+  // We try to preserve the order of tests when they require different workers
+  // by ordering different worker hashes sequentially.
+  const workerHashToOrdinal = new Map<string, string>();
```
pavelfeldman, line 444: I can't not comment on this, but ordinal must be an integer number.

pavelfeldman (approved):

PR #7926, opened 2021-07-30
feat(test-runner): experiemental expect plumbing

pavelfeldman:


src/test/workerRunner.ts
```diff
@@ -263,6 +269,37 @@ export class WorkerRunner extends EventEmitter {
...
+            }).finally(() => {
+              this.emit('expectEnd', expectEndPayload);
+            });
+          }
+          return result;
+        } catch (e) {
+          expectEndPayload.error = e.message;
+          throw e;
+        } finally {
+          this.emit('expectEnd', expectEndPayload);
```
dgozman, line 300: It seems like we emit expectEnd twice for async expects, at line 300 and line 292.

tests/playwright-test/reporter.spec.ts
```diff
@@ -158,3 +158,85 @@ test('should load reporter from node_modules', async ({ runInlineTest }) => {
...
+    'playwright.config.ts': `
+      module.exports = {
+        reporter: './reporter',
+      };
+    `,
+    'a.test.ts': `
+      const { test } = pwt;
+      test('fail', async ({}) => {
+        expect(true).toBeTruthy();
+        expect(false).toBeTruthy();
```
dgozman, line 186: Please add async expect here as well.

dgozman (approved):

PR #7951, opened 2021-08-02
feat(inputValue): implement inputValue for select elements

fabianMendez:


src/server/dom.ts
```diff
@@ -884,7 +884,7 @@ export function throwFatalDOMError<T>(result: T | FatalDOMError): T {
   if (result === 'error:notinput')
     throw new Error('Node is not an HTMLInputElement');
   if (result === 'error:notinputvalue')
```
pavelfeldman, line 886: Now that it isn't input, we can rename `error:notinputvalue` to `error:hasnovalue`.

pavelfeldman (approved):
This looks great! Just one minor nit, but it can be fixed later.

src/server/dom.ts
fabianMendez: Done

PR #7974, opened 2021-08-03
feat(tracing): introduce _reset() and _export()

dgozman:
`tracing._export({ path })` exports current tracing state into a file
and does not require tracing to be stopped.

`tracing._reset()` resets current tracing state, but keeps resources
around so they can be referenced in the future snapshots. Does not stop.

The usage pattern is:
```js
await tracing.start({ screenshots: true, snapshots: true });
// ...
await tracing._reset();
// Do stuff, it will all be in the export below.
await tracing._export({ path });
// ...
await tracing.stop();
```

### DIFF
--- a/src/server/trace/recorder/tracing.ts
+++ b/src/server/trace/recorder/tracing.ts
@@ -17,6 +17,7 @@
 import fs from 'fs';
 import path from 'path';
 import yazl from 'yazl';
+import readline from 'readline';
 import { EventEmitter } from 'events';
 import { calculateSha1, createGuid, mkdirIfNeeded, monotonicTime } from '../../../utils/utils';
 import { Artifact } from '../../artifact';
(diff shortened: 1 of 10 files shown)


src/server/trace/recorder/tracing.ts
```diff
@@ -194,42 +238,58 @@ export class Tracing implements InstrumentationListener {
             height: params.height,
             timestamp: monotonicTime()
           };
-          this._appendTraceEvent(event);
-          this._appendEventChain = this._appendEventChain.then(async () => {
+          // Make sure to write the screencast frame before adding a reference to it.
+          this._appendTraceOperation(async () => {
             await fs.promises.writeFile(path.join(this._resourcesDir!, sha1), params.buffer).catch(() => {});
           });
```
yury-s, line 244: We should probably catch here to avoid unhandled promise rejections.

yury-s (approved):

src/server/trace/recorder/tracing.ts
dgozman: We do in line 243.

PR #7980, opened 2021-08-04
chore: PlaywrightClient/Server enhancements

mxschmitt:
- ensure timeout results in a meaningful message
- add onDisconnect handler

### DIFF
--- a/src/remote/playwrightServer.ts
+++ b/src/remote/playwrightServer.ts
@@ -33,6 +33,7 @@ export interface PlaywrightServerDelegate {
 
 export type PlaywrightServerOptions = {
   acceptForwardedPorts?: boolean
+  onDisconnect?: () => void;
 };
 
 export class PlaywrightServer {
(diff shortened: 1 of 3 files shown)


src/remote/playwrightServer.ts
```diff
@@ -33,14 +33,15 @@ export interface PlaywrightServerDelegate {
 
 export type PlaywrightServerOptions = {
   acceptForwardedPorts?: boolean
+  onDisconnect?: () => void;
```
yury-s, line 36: I don't see any usages, drop it for now?

mxschmitt: I use it for a demo, see later in the standup. Since it's only an internal API, it should be fine?

yury-s: If it's just for the demo you can use a local change? We normally stick to the rule of not adding unused code but if you really need it to be in the repo for the demo it's okay.

yury-s (approved):

src/remote/playwrightServer.ts
mxschmitt: (not exposed API, thats why I landed it for now. If we publish it at some point we re-consider the whole use of the API)

PR #7984, opened 2021-08-04
browser(webkit): roll to 04/08/21

q66:
3-way diff: https://github.com/q66/WebKit/commit/5d24e3b062fc40910b23e58fe590e1eb9c031af0
also a build fix: https://github.com/q66/WebKit/commit/e3d45214a407907b0e8f033dec16940ecc92d066

i had some sketchy output in tests, but dunno if a regression; i want to see the CI run before bisecting it

### DIFF
--- a/browser_patches/webkit/BUILD_NUMBER
+++ b/browser_patches/webkit/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1524
-Changed: yurys@chromium.org Mon 02 Aug 2021 03:47:03 PM PDT
+1525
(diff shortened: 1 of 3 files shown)


yury-s (approved):
> 3-way diff: [q66/WebKit@5d24e3b](https://github.com/q66/WebKit/commit/5d24e3b062fc40910b23e58fe590e1eb9c031af0)

Not a problem for this roll but it seems to be a "merge" rather than "diff3" output to me.

q66:
good point, the config probably got lost since i had to switch repos...

PR #7989, opened 2021-08-04
test: fix some failing locators tests

mxschmitt:


tests/page/locator-convenience.spec.ts
```diff
@@ -103,9 +103,9 @@ it('textContent should be atomic', async ({ playwright, page }) => {
       return result;
     }
   });
-  await playwright.selectors.register('textContent', createDummySelector);
+  await playwright.selectors.register('textContentFromLocators', createDummySelector);
```
mxschmitt, line 106: since Android shares the page, we need to use a different name. We already have this selector engine name in another test file.

dgozman (approved):

PR #7990, opened 2021-08-04
chore: cleanup locators implementation

mxschmitt:
Since _withElement internally calls .element which already does do `strict: true` there is no need of passing strict again. When we have a handle, we don't need strict in these option parameters.

### DIFF
--- a/src/client/locator.ts
+++ b/src/client/locator.ts
@@ -32,23 +32,21 @@ export class Locator implements api.Locator {
     this._selector = selector;
   }
 
-  private async _withElement<R, O extends TimeoutOptions>(task: (handle: ElementHandle<SVGElement | HTMLElement>, options?: O) => Promise<R>, options?: O): Promise<R> {
-    if (!options)
-      options = {} as O;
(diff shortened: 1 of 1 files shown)


src/client/locator.ts
```diff
@@ -48,7 +48,7 @@ export class Locator implements api.Locator {
   }
 
   async boundingBox(options?: TimeoutOptions): Promise<Rect | null> {
-    return this._withElement(h => h.boundingBox(), { strict: true, ...options });
+    return this._withElement(h => h.boundingBox(), options);
```
dgozman, line 51: Did you consider passing timeout to and from `_withElement` instead of options?

```js
async boundingBox(options: TimeoutOptions = {}): Promise<Rect | null> {
  return this._withElement((h, timeout) => h.boundingBox(), options.timeout);
}
```

mxschmitt: yeah doing it now like in Python. -> Less internal magic and only the new timeout calculation.

dgozman (approved):

Issue #7771, opened 2021-07-21
[Bug]: `device.launchBrowser()` stops on `about:blank` when changing the page URL


OxNinja:
### Playwright version

1.12.2

### Operating system

Linux

### What browsers are you seeing the problem on?

Chromium

### Other information

- NodeJS `v12.21.0`
-  `Linux thonkpad 5.10.0-kali9-amd64 #1 SMP Debian 5.10.46-1kali1 (2021-06-25) x86_64 GNU/Linux`

### What happened? / Describe the bug

Using physical devices and ADB I want to use Playwright's NodeJS API for browser automation.

I can get an AndroidDevice object and launch its browser using my script. But it stops here on a new page `about:blank`.

I use *experimental Android API functions* in my code, so maybe this is why I get this bug.

### Code snippet to reproduce your bug

```shell
const { _android: android, chromium } = require("playwright");

(async () => {
    // get the list of all connected devices
    const devices = await android.devices();
    for(device of devices) {
        // in real example I start chromium remote debugging with the flag `--remote-debugging-port=${port}`, but for reading purposes I simplified the thing
        device.shell(`am start -n com.android.chrome/com.google.android.apps.chrome.Main`);
        // launch the browser to automate it afterwards
        const browser = await device.launchBrowser(); // from here, no more logging output, even using `console.log`
        const page = await browser.newPage(); // launches a new page stuck on `about:blank`
        
        await page.goto("https://google.com"); // not working

        await browser.close();
    }
})();
```

pavelfeldman:
A couple of things to fix:

1. You are not supposed to be running `device.shell(`am start -n com.android.chrome/com.google.android.apps.chrome.Main`);`, `launchBrowser` does it for you:

2. Android browser is single-context, so you get context and page right away:

```js
const browserContext = await device.launchBrowser();
const [ page ] = browserContext.pages();
```

Please tell us if that helped.

OxNinja:
I am using `device.shell(am start ...)` in order to use Chromium's remote debug feature, does `launchBrowser` implements that?

I tried with your code and it does not change the problem, am I missing something important?

pavelfeldman:
Which Chromium's remote debug feature are you talking about? You mean chrome://insect to connect to the browser, yes, that works for all the browser instances.

OxNinja:
[...]

PR #8014, opened 2021-08-05
docs: add android requirements to the api docs page

pavelfeldman:
For https://github.com/microsoft/playwright/issues/7771

### DIFF
--- a/docs/src/api/class-android.md
+++ b/docs/src/api/class-android.md
@@ -3,6 +3,21 @@
 
 Playwright has **experimental** support for Android automation. See [here](./mobile.md) for more information. You can access android namespace via:
 
+*Requirements*
+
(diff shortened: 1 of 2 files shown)


docs/src/api/class-android.md
```diff
@@ -3,6 +3,21 @@
 
 Playwright has **experimental** support for Android automation. See [here](./mobile.md) for more information. You can access android namespace via:
 
+*Requirements*
```
mxschmitt, line 6: now its quite similar to https://playwright.dev/docs/mobile

can we nuke the mobile page?

pavelfeldman: Yes, we should, our mobile page is not indexed...

PR #8019, opened 2021-08-05
chore(protocol): do client hello instead of server hello

mxschmitt:
Relates #7741
Python implementation https://github.com/microsoft/playwright-python/pull/836

### DIFF
--- a/src/client/channelOwner.ts
+++ b/src/client/channelOwner.ts
@@ -78,7 +78,7 @@ export abstract class ChannelOwner<T extends channels.Channel = channels.Channel
         if (prop === 'debugScopeState')
           return (params: any) => this._connection.sendMessageToServer(this, prop, params, stackTrace);
         if (typeof prop === 'string') {
-          const validator = scheme[paramsName(this._type === '' ? 'Root' : this._type, prop)];
+          const validator = scheme[paramsName(this._type, prop)];
           if (validator)
             return (params: any) => this._connection.sendMessageToServer(this, prop, validator(params, ''), stackTrace);
         }
--- a/src/client/channelOwner.ts
+++ b/src/client/channelOwner.ts
@@ -78,7 +78,7 @@ export abstract class ChannelOwner<T extends channels.Channel = channels.Channel
         if (prop === 'debugScopeState')
           return (params: any) => this._connection.sendMessageToServer(this, prop, params, stackTrace);
         if (typeof prop === 'string') {
-          const validator = scheme[paramsName(this._type, prop)];
+          const validator = scheme[paramsName(this._type === '' ? 'Root' : this._type, prop)];
           if (validator)
             return (params: any) => this._connection.sendMessageToServer(this, prop, validator(params, ''), stackTrace);
         }
--- a/src/dispatchers/dispatcher.ts
+++ b/src/dispatchers/dispatcher.ts
@@ -23,6 +23,7 @@ import { tOptional } from '../protocol/validatorPrimitives';
 import { kBrowserOrContextClosedError } from '../utils/errors';
 import { CallMetadata, SdkObject } from '../server/instrumentation';
 import { rewriteErrorMessage } from '../utils/stackTrace';
+import { createPlaywright } from '../server/playwright';
 
 export const dispatcherSymbol = Symbol('dispatcher');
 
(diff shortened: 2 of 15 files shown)


src/dispatchers/dispatcher.ts
```diff
@@ -125,6 +126,12 @@ class Root extends Dispatcher<{ guid: '' }, {}> {
   constructor(connection: DispatcherConnection) {
     super(connection, { guid: '' }, '', {}, true);
   }
+
+  initialize(params: { language?: string }): import('./playwrightDispatcher').PlaywrightDispatcher {
```
kumaraditya303, line 130: You can use `import type` here

src/dispatchers/dispatcher.ts
```diff
@@ -121,15 +122,21 @@ export class Dispatcher<Type extends { guid: string }, Initializer> extends Even
...
 export type DispatcherScope = Dispatcher<any, any>;
-class Root extends Dispatcher<{ guid: '' }, {}> {
-  constructor(connection: DispatcherConnection) {
+export class Root extends Dispatcher<{ guid: '' }, {}> {
+  constructor(connection: DispatcherConnection, private readonly createPlaywright?: (scope: DispatcherScope) => Promise<PlaywrightDispatcher>) {
     super(connection, { guid: '' }, '', {}, true);
   }
+
+  async initialize(params: { language?: string }): Promise<channels.RootInitializeResult> {
+    assert(this.createPlaywright);
```
mxschmitt, line 131: note: for inprocess we don't need it, for everything else we use it.

src/dispatchers/dispatcher.ts
```diff
@@ -121,15 +122,21 @@ export class Dispatcher<Type extends { guid: string }, Initializer> extends Even
...
 
 export type DispatcherScope = Dispatcher<any, any>;
-class Root extends Dispatcher<{ guid: '' }, {}> {
-  constructor(connection: DispatcherConnection) {
+export class Root extends Dispatcher<{ guid: '' }, {}> {
+  constructor(connection: DispatcherConnection, private readonly createPlaywright?: (scope: DispatcherScope) => Promise<PlaywrightDispatcher>) {
     super(connection, { guid: '' }, '', {}, true);
   }
+
+  async initialize(params: { language?: string }): Promise<channels.RootInitializeResult> {
```
mxschmitt, line 130: note: will add the language wiring for the error message in a follow-up PR.

src/client/channelOwner.ts
```diff
@@ -78,7 +78,7 @@ export abstract class ChannelOwner<T extends channels.Channel = channels.Channel
         if (prop === 'debugScopeState')
           return (params: any) => this._connection.sendMessageToServer(this, prop, params, stackTrace);
         if (typeof prop === 'string') {
-          const validator = scheme[paramsName(this._type, prop)];
+          const validator = scheme[paramsName(this._type === '' ? 'Root' : this._type, prop)];
```
dgozman, line 81: Why don't we change Root constructor to pass 'Root' type instead?

src/dispatchers/dispatcher.ts
```diff
@@ -121,15 +122,21 @@ export class Dispatcher<Type extends { guid: string }, Initializer> extends Even
...
+    assert(this.createPlaywright);
+    return {
+      playwright: await this.createPlaywright(this),
```
dgozman, line 133: - Let's throw if this method is called twice.
- Why pass `scope` if everyone already has access to the `Root` anyway?

src/dispatchers/dispatcher.ts
```diff
@@ -175,6 +180,8 @@ export class DispatcherConnection {
     };
     const scheme = createScheme(tChannel);
     this._validateParams = (type: string, method: string, params: any): any => {
+      if (type === '')
```
dgozman, line 183: Similarly, make it a have a type 'Root' in the constructor?

dgozman (approved):
This looks pretty good, just a few minor comments.

src/client/channelOwner.ts
mxschmitt: yeah was a hacky workaround, calling it actually correctly is way better.

PR #8031, opened 2021-08-06
fix(docs): missing reference type

avodovnik:
This was causing the .NET API to have invalid XML comments.


mxschmitt (approved):
you need to re-generate the types. (npm run build / npm run watch)

Issue #7998, opened 2021-08-05
[internal]: fix firefox-beta video tests


mxschmitt:
[firefox] › video.spec.ts:361:3 › screencast should capture css transformation =================
    [firefox] › video.spec.ts:421:3 › screencast should scale frames down to the requested size  ===

PR #8033, opened 2021-08-06
browser(ff-beta): fix screencast

aslushnikov:
Both `libyuv::I420Copy` and `libyuv::I420Scale` support image cropping
by offsetting coordinates inside planes, but offsets must be even
numbers.

References #7998

### DIFF
--- a/browser_patches/firefox-beta/juggler/screencast/ScreencastEncoder.cpp
+++ b/browser_patches/firefox-beta/juggler/screencast/ScreencastEncoder.cpp
@@ -135,6 +135,9 @@ public:
 
         double src_width = src->width() - m_margin.LeftRight();
         double src_height = src->height() - m_margin.top;
+        // YUV offsets must be even.
+        int yuvTopOffset = m_margin.top & 1 ? m_margin.top + 1 : m_margin.top;
+        int yuvLeftOffset = m_margin.left & 1 ? m_margin.left + 1 : m_margin.left;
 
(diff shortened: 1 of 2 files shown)


browser_patches/firefox-beta/juggler/screencast/ScreencastEncoder.cpp
```diff
@@ -113,7 +113,11 @@ class ScreencastEncoder::VPXFrame {
     VPXFrame(rtc::scoped_refptr<webrtc::VideoFrameBuffer>&& buffer, const gfx::IntMargin& margin)
         : m_frameBuffer(std::move(buffer))
         , m_margin(margin)
-    { }
+    {
+      // YUV offsets must be even.
+      m_yuvTopOffset = m_margin.top & 1 ? m_margin.top + 1 : m_margin.top;
```
yury-s, line 118: nit: I'd use  `m_yuvTopOffset = m_margin & ~1;` and moved this inside `convertToVpxImage` where we compute all other yuv related stuff.

yury-s (approved):
Impressive!

browser_patches/firefox-beta/juggler/screencast/ScreencastEncoder.cpp
aslushnikov: I kept rounding to top so that it doesn't cut the top-left corner. Other then that, inlined it in

PR #8046, opened 2021-08-06
doc(test-runner): add example of compiling tests with typescript

JoelEinbinder:
I was having trouble figuring out how the docs are structured. Will this file be
automatically picked up by the search? Should it be linked from somewhere? Does it
need to be called `typescript-js.md`?﻿

### DIFF
--- a/docs/src/typescript-js.md
+++ b/docs/src/typescript-js.md
@@ -1,0 +1,34 @@
+## Manually compile tests with TypeScript
+
+Playwright Test supports TypeScript out the box. We automatically transform
+TypeScript code to javascript to run it.
+
+However if you find that the TypeScript code is not being transpiled correctly,
+you can perform your own TypeScript compilation before sending the tests to Playwright.
+
+First I add a `tsconfig.json` file inside my tests directory.
+```json
+{
+    "compilerOptions": {
+        "target": "ESNext",
+        "module": "commonjs",
+        "moduleResolution": "Node",
+        "sourceMap": true,
+        "outDir": "../tests-out",
+    }
+}
+```
+
+In my `package.json`, I have two scripts:
+```json
+{
+  "scripts": {
+    "pretest": "tsc --incremental -p tests/tsconfig.json",
+    "test": "playwright test -c tests-out"
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


mxschmitt:
> I was having trouble figuring out how the docs are structured. Will this file be
> automatically picked up by the search? Should it be linked from somewhere? Does it
> need to be called `typescript-js.md`?

It needs to be suffixed with "-js" since you want it only in the "Node.js variant". Does it get picked up by the search? Yes it does. But when rolling the docs on playwright.dev you need to add it to the sidebar.json.

docs/src/typescript.md
```diff
@@ -0,0 +1,34 @@
...
+  "scripts": {
+    "pretest": "tsc --incremental -p tests/tsconfig.json",
+    "test": "playwright test -c tests-out"
+  }
+}
+```
+
+The `pretest` script runs typescript on the tests.  `test` will run the tests that have been generated to the `tests-out` directory.
+
+Then `npm run test` will build the tests and run them.
```
mxschmitt, line 34: hmm, how to use an existing tsconfig? Why does the playwright.config.ts need to be compiled? What is the -c parameter? Why is the config a directory?

JoelEinbinder: I think using an existing tsconfig is pretty straightforward once you read this doc.

There isn't anything here about playwright.config.ts, but it needs to be compiled the same as the tests because it might use syntax we dont support through babel. Complain to Dmitry about the `-c` command line api :).

docs/src/typescript-js.md
```diff
@@ -0,0 +1,34 @@
...
+```json
+{
+  "scripts": {
+    "pretest": "tsc --incremental -p tests/tsconfig.json",
+    "test": "playwright test -c tests-out"
+  }
+}
+```
+
+The `pretest` script runs typescript on the tests.  `test` will run the tests that have been generated to the `tests-out` directory. The `-c` argument configures the test runner to look for tests inside the `tests-out` directory.
```
mxschmitt, line 32:
```suggestion
The `pretest` script runs typescript on the tests. `test` will run the tests that have been generated to the `tests-out` directory. The `-c` argument configures the test runner to look for tests inside the `tests-out` directory.
```

mxschmitt (approved):

PR #8047, opened 2021-08-06
feat(chromium): roll Chromium to M94

aslushnikov:


mxschmitt (approved):

src/server/chromium/chromium.ts
```diff
@@ -198,7 +198,7 @@ const DEFAULT_ARGS = [
   '--disable-dev-shm-usage',
   '--disable-extensions',
   // BlinkGenPropertyTrees disabled due to crbug.com/937609
-  '--disable-features=TranslateUI,BlinkGenPropertyTrees,ImprovedCookieControls,SameSiteByDefaultCookies,LazyFrameLoading,GlobalMediaControls,DestroyProfileOnBrowserClose',
+  '--disable-features=TranslateUI,BlinkGenPropertyTrees,ImprovedCookieControls,LazyFrameLoading,GlobalMediaControls,DestroyProfileOnBrowserClose',
```
dgozman, line 201: I cannot find TranslateUI and BlinkGenPropertyTrees features in Chromium source anymore...

tests/browsercontext-cookies.spec.ts
```diff
@@ -137,14 +137,17 @@ it('should get cookies from multiple urls', async ({context}) => {
     url: 'https://foo.com',
     name: 'doggo',
     value: 'woofs',
+    sameSite: 'Lax',
```
dgozman, line 140: Do I understand correctly that if we add sameSite: None here, we'll get None at line 161? Let's mix and match different SameSite values to see that all of them work?

tests/browsercontext-add-cookies.spec.ts
```diff
@@ -206,7 +206,8 @@ it('should set cookie with reasonable defaults', async ({context, server}) => {
   await context.addCookies([{
     url: server.EMPTY_PAGE,
     name: 'defaults',
-    value: '123456'
+    value: '123456',
+    sameSite: 'Lax',
```
dgozman, line 210: In this particular test we check "reasonable defaults", so I'd rather have different expectations per browser.

src/server/chromium/crBrowser.ts
```diff
@@ -361,7 +361,7 @@ export class CRBrowserContext extends BrowserContext {
   async _doCookies(urls: string[]): Promise<types.NetworkCookie[]> {
     const { cookies } = await this._browser._session.send('Storage.getCookies', { browserContextId: this._browserContextId });
     return network.filterCookies(cookies.map(c => {
-      const copy: any = { sameSite: 'None', ...c };
+      const copy: any = { sameSite: 'Lax', ...c };
```
dgozman, line 364: I don't understand this line at all. Do we lack sameSite property sometimes? Why do we assume Lax default now?

aslushnikov: lax is default now in chromium unless specified explicitly. (Used to be 'None')

src/server/chromium/chromium.ts
aslushnikov: cleaned up

tests/browsercontext-cookies.spec.ts
aslushnikov: done

tests/browsercontext-add-cookies.spec.ts
aslushnikov: Done

dgozman (approved):

PR #8058, opened 2021-08-06
fix(test runner): make sure options, trace and screenshot apply to all contexts

dgozman:


src/test/index.ts
```diff
@@ -47,46 +60,42 @@ export const test = _baseTest.extend<PlaywrightTestArgs & PlaywrightTestOptions,
...
-  timezoneId: undefined,
-  userAgent: undefined,
-  viewport: undefined,
-  actionTimeout: undefined,
-  navigationTimeout: undefined,
-  baseURL: async ({ }, use) => {
+  screenshot: ['off', { scope: 'file' }],
+  video: ['off', { scope: 'file' }],
+  trace: ['off', { scope: 'file' }],
+  acceptDownloads: [undefined, { scope: 'file' }],
```
mxschmitt, line 72: Can I still use it inside a describe times with test.use?

Issue #7189, opened 2021-06-17
[Doc] Introduce Protractor and Jest migration guides for Test Runner


pavelfeldman:
There are a lot of questions about global page object, global setup, passing data among those migrating. Let's help them with the guide.

https://github.com/microsoft/playwright/issues/7116
https://github.com/microsoft/playwright/issues/7137
#7128

Tallyb:
FWIW, I will leave it here for defining browser as global with typescript (I had some trouble getting there): 

```
//@types/node@16
// eslint-disable-next-line no-var
var browser: ChromiumBrowser | FirefoxBrowser | WebKitBrowser;

declare global {
  // eslint-disable-next-line no-var
  var browser: ChromiumBrowser | FirefoxBrowser | WebKitBrowser;
}
browser = await chromium.launch()
```

```
//@types/node@15
declare global {
  // eslint-disable-next-line @typescript-eslint/no-namespace
  namespace NodeJS {
    interface Global {
      browser: ChromiumBrowser | FirefoxBrowser | WebKitBrowser; //change to your favorite browser
    }
  }
}
global.browser = await webkit.launch(browserOptions);

aslushnikov:
- We're looking for the protractor test suite with no luck. 
- We decided not to follow-up on the jest-playwright migration guide. Those folks are already happy!

PR #8069, opened 2021-08-07
feat: introduce react selectors

aslushnikov:
This patch adds support for the `react` selector engine that allows
selecting DOM elements based on the component name.

> **NOTE**: in case of multi-root components (React.Fragment), `react`
engine will select all root DOM elements.

> **NOTE**: `react` engine supports react v15+.

References #7189

### DIFF
--- a/docs/src/selectors.md
+++ b/docs/src/selectors.md
@@ -179,6 +179,25 @@ methods accept [`param: selector`] as their first argument.
   await page.ClickAsync("xpath=//button");
   ```
   Learn more about [XPath selector][xpath].
+- React selector
+  ```js
+  await page.click('react=DatePickerComponent');
+  ```
+  ```java
+  page.click("react=DatePickerComponent");
+  ```
+  ```python async
+  await page.click("react=DatePickerComponent")
+  ```
+  ```python sync
+  page.click("react=DatePickerComponent")
+  ```
+  ```csharp
+  await page.ClickAsync("react=DatePickerComponent");
+  ```
+  Learn more about [React selector][react].
+
+
 
 ## Text selector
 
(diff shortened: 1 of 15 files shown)


docs/src/selectors.md
```diff
@@ -625,6 +644,25 @@ converts `'//html/body'` to `'xpath=//html/body'`.
 `xpath` does not pierce shadow roots
 :::
 
+## React selectors
+
+React selectors allow selecting elements by its component name. If component
+ has multiple root elements (see [React Fragments](https://reactjs.org/docs/fragments.html)),
+ `react` selector will return all of them.
```
dgozman, line 651: I don't quite understand "all of them". When I click on react=MultiRoot, will it click all of them? Let's be clear here. Also, I'd move this to a note, since it's an edge case.

tests/page/selectors-react.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+for (const [name, url] of Object.entries(reacts)) {
+  it.describe(name, () => {
+    it.beforeEach(async ({page, server}) => {
+      await page.goto(server.PREFIX + url);
+    });
+
+    it('should work with single-root elements', async ({page}) => {
+      expect(await page.$$eval(`react=BookList`, els => els.length)).toBe(1);
```
dgozman, line 33: Why only $$eval tests? Let's do some single-element queries like $.

tests/page/selectors-react.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      expect(await page.$$eval(`react=BookItem`, els => els.length)).toBe(3);
+      expect(await page.$$eval(`react=BookList >> react=BookItem`, els => els.length)).toBe(3);
+      expect(await page.$$eval(`react=BookItem >> react=BookList`, els => els.length)).toBe(0);
+    });
+
+    it('should work with multi-root elements (fragments)', async ({page}) => {
+      it.skip(name === 'react15', 'React 15 does not support fragments');
+      // App is a fragment.
+      expect(await page.$$eval(`react=App`, els => els.length)).toBe(5);
+      // App is a fragment.
```
dgozman, line 43: Same comment 3 times?

dgozman (approved):

docs/src/selectors.md
aslushnikov: dropped this until we release `_nth` engine

tests/page/selectors-react.spec.ts
aslushnikov: done

PR #8071, opened 2021-08-08
feat(expect): introduce explicit default async expect timeout

pavelfeldman:


tests/playwright-test/playwright.expect.misc.spec.ts
```diff
@@ -159,9 +159,9 @@ test('should support toHaveURL', async ({ runInlineTest }) => {
   expect(result.exitCode).toBe(1);
 });
 
-test('should support respect actionTimeout', async ({ runInlineTest }) => {
+test.only('should support respect actionTimeout', async ({ runInlineTest }) => {
```
dgozman, line 162: 'should respect expect timeout'

dgozman (approved):

Issue #7673, opened 2021-07-16
[Feature]: Case insensitive `--grep`


mskelton:
### Feature request

Currently, the `--grep` CLI command is case sensitive which makes matching tests more difficult if you don't know remember the exact casing offhand.  I think it would be beneficial to make the `--grep` command case insensitive by default.

Looking at the source, it would appear that the grep command accepts a regex expression, but since this is also a CLI flag it isn't as simple as just passing a regex argument.

aslushnikov:
it looks like it's working! Please re-file if it's not!

PR #8091, opened 2021-08-09
test(test-runner): add tests to show that grep is case insensitive

JoelEinbinder:
We were missing tests that ensured that `--grep` was case insensitive.

#7673﻿

### DIFF
--- a/tests/playwright-test/match-grep.spec.ts
+++ b/tests/playwright-test/match-grep.spec.ts
@@ -87,3 +87,15 @@ test('should grep invert test name', async ({ runInlineTest }) => {
   expect(result.skipped).toBe(0);
   expect(result.exitCode).toBe(0);
 });
+
+test('should be case insensitive by default', async ({ runInlineTest }) => {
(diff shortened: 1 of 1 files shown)


tests/playwright-test/match-grep.spec.ts
```diff
@@ -87,3 +87,15 @@ test('should grep invert test name', async ({ runInlineTest }) => {
   expect(result.skipped).toBe(0);
   expect(result.exitCode).toBe(0);
 });
+
+test('should be case insensitive by default', async ({ runInlineTest }) => {
+  const result = await runInlineTest(files, { 'grep': 'TesT Cc' });
```
yury-s, line 92: I wonder if it actually a regex or just a substring in this case?

yury-s (approved):

tests/playwright-test/match-grep.spec.ts
JoelEinbinder: The first test case in this file implies that it is a regex.

PR #8102, opened 2021-08-09
fix(test runner): ensure we run after hooks after failures

dgozman:
Instead of checking `this._isStopped` everywhere in the code for an early exit, we now always run `WorkerRunner._run` until the end, only skipping entire suites and tests after stop was requested. This ensures we run all the hooks after a failure or user interrupt. We also "terminate" the current `DeadlineRunner` so that we do not have to wait for the whole test to finish.

### DIFF
--- a/src/test/workerRunner.ts
+++ b/src/test/workerRunner.ts
@@ -43,7 +43,9 @@ export class WorkerRunner extends EventEmitter {
   private _fatalError: TestError | undefined;
   private _entries = new Map<string, TestEntry>();
   private _isStopped = false;
-  _currentTest: { testId: string, testInfo: TestInfoImpl, testFinishedCallback: () => void } | null = null;
+  private _runFinished = Promise.resolve();
+  private _currentDeadlineRunner: DeadlineRunner<any> | undefined;
+  _currentTest: { testId: string, testInfo: TestInfoImpl, type: 'test' | 'beforeAll' | 'afterAll' } | null = null;
 
   constructor(params: WorkerInitParams) {
     super();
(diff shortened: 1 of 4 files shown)


JoelEinbinder:
Can we add this test for max failures. Reading the code it looks like it should work, but who knows after the next refactor?

```ts
test('max-failures should still run afterEach/afterAll', async ({ runInlineTest }) => {
    'a.spec.js': `
      const { test } = pwt;
      test.afterAll(() => {
        console.log('\n%%afterAll');
      });
      test.afterEach(() => {
        console.log('\n%%afterAll');
      });
      test('failed', async () => {
        console.log('\\n%%test');
        test.expect(1).toBe(2);
      });
      test('skipped', async () => {
        console.log('\\n%%skipped');
      });
    `,
  }, { 'max-failures': 1 });
  expect(result.passed).toBe(0);
  expect(result.failed).toBe(1);
  expect(result.skipped).toBe(1);
  expect(stripAscii(result.output).split('\n').filter(line => line.startsWith('%%'))).toEqual([
    '%%test',
    '%%afterEach',
    '%%afterAll',
  ]);
});
```

src/test/workerRunner.ts
```diff
@@ -420,7 +410,7 @@ export class WorkerRunner extends EventEmitter {
       await this._runBeforeHooks(test, testInfo);
 
     // Do not run the test when beforeEach hook fails.
-    if (this._isStopped || testInfo.status === 'failed' || testInfo.status === 'skipped')
+    if (testInfo.status === 'failed' || testInfo.status === 'skipped')
```
JoelEinbinder, line 413: Do we run afterEach hooks when the beforeEach hook fails? I tested out mocha to verify what they do.

```js

before('before', () => {
    console.log('before');
    // throw new Error('before');
});
beforeEach('beforeEach', () => {
    console.log('beforeEach');
    // throw new Error('beforeEach');
});
after('after', () => {
    console.log('after');
    // throw new Error('after');
});
afterEach('afterEach', () => {
    console.log('afterEach');
    // throw new Error('afterEach');
});
it('test', () => {
    console.log('test');
    // throw new Error('test');
});
```

beforeAll throwing causes only afterAll to run
beforeEach throwing causes afterEach/afterAll to run
the test throwing causes everything to run

dgozman: We have `afterEach failure should not prevent other hooks and fixtures teardown` and `beforeEach failure should prevent the test, but not other hooks`. Let me add `beforeAll throwing causes only afterAll to run`.

JoelEinbinder (approved):

PR #8106, opened 2021-08-09
feat: make React and Vue selectors experimental

aslushnikov:
This way we'll be able to collect user feedback and fix certain things.


docs/src/selectors.md
```diff
@@ -181,38 +181,38 @@ methods accept [`param: selector`] as their first argument.
   Learn more about [XPath selector][xpath].
 - React selector
```
dgozman, line 182: Say "experimental" here as well?

dgozman (approved):

PR #8110, opened 2021-08-10
chore(tracing): do not produce cached resources maps

pavelfeldman:


src/server/snapshot/snapshotRenderer.ts
```diff
@@ -82,10 +82,12 @@ export class SnapshotRenderer {
...
 
     const resources: { [key: string]: { resourceId: string, sha1?: string } } = {};
-    for (const [url, contextResources] of this._contextResources) {
-      const contextResource = contextResources.find(r => r.frameId === snapshot.frameId) || contextResources[0];
-      if (contextResource)
-        resources[url] = { resourceId: contextResource.resourceId };
+    for (const resource of this._resources) {
+      if (resource.timestamp >= snapshot.timestamp)
+        break;
+      if (resource.frameId !== snapshot.frameId)
```
dgozman, line 88: I think resources can be reused between frames (e.g. from memory cache), so we should fallback to resource with any frameId if we didn't find one with specific frameId.

dgozman (approved):

PR #8121, opened 2021-08-10
browser(webkit): disable gpu process on macos

q66:
Keeping it enabled causes test crashes in local builds and CI since last week's roll (which seemingly enabled the functionality following some preferences code refactoring).

Therefore, disable until it can be fixed. On non-macos platforms, this is already always disabled.

This needs local testing on a Mac machine, @yury-s should be able to.

### DIFF
--- a/browser_patches/webkit/BUILD_NUMBER
+++ b/browser_patches/webkit/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1529
-Changed: yurys@chromium.org Fri 06 Aug 2021 12:34:04 PM PDT
+1530
+Changed: dkolesa@igalia.com Tue Aug 10 10:15:56 PM CEST 2021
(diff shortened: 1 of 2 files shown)


yury-s:
For the record, upstream commit that caused the failures: https://github.com/WebKit/WebKit/commit/57589b600368f5dd8f55784908b7081f8bba8e84

q66:
on my local machine, that is; on the CI, I believe it was this one: https://github.com/WebKit/WebKit/commit/0d4cc16381283e61dee1b8e7777bced359d02fe0

pavelfeldman:
I don't think we should be disabling GPU on WebKit macOS.

q66:
This does not disable GPU, it just disables separate GPUProcess (coincidentally, that it how it has been until last week)

q66:
that said, @yury-s appears to have found that there might be another way, so let's attempt that first

q66:
better fix in the linked PR

PR #8122, opened 2021-08-10
fix(trace viewer): no flickering for snapshots

dgozman:
Switch from `createRef` to `useRef` to retain ref identity, and stop
depending on `snapshots` array that is recreated every render.


pavelfeldman (approved):

dgozman:
Included in #8126.

PR #8127, opened 2021-08-10
browser(webkit): include GPU process in mac archive

q66:
As it turns out, the GPU process executable just wasn't included in the bundle (and I failed to notice that since the macOS names are different and when I considered this possibility I did not notice it while looking for it), thanks @yury-s for the hint.

Fixes Mac issues that started with https://github.com/microsoft/playwright/pull/7984.

Also closes https://github.com/microsoft/playwright/pull/8121.


yury-s (approved):
Let's also update BUILD_NUMBER to trigger new build.

q66:
done

yury-s (approved):

Issue #7447, opened 2021-07-02
[Question] How to use launchPersistentContext with Playwright Test


badsyntax:
Hi. I'm attempting to run lighthouse on browser instances managed by playwright. 

Lighthouse will launch a new window/session and the auth state is not persisted between windows. 

To allow persistent auth I can use `launchPersistentContext`. This works fine outside of playwright test:

```js
const userDataDir = '/path/to/data/dir';

const context = await chromium.launchPersistentContext(userDataDir, {
  args: ['--remote-debugging-port=9222'],
});
const page = await context.newPage();
await page.goto('http://localhost:3000/');

// Perform login steps here which will save to cookie or localstorage

// When lighthouse opens a new page the storage will be pesisted meaning the new page will have the same session (either via cookie or localstorage)
await lighthouseImpl({
  page: page,
  port: 9222,
});
```

How can I use `launchPersistentContext` with the Playwright Test Runner? I've attempted to override the browser context using fixtures, but I get the following error:

```
Fixture "context" has already been registered as a { scope: 'test' } fixture.
```

I am able to override the `browser` fixture but I don't have access to a `browser` instance when using `launchPersistentContext`. Looking for some tips or pointers, thanks in advance!

pavelfeldman:
There are several ways of addressing your issue. Could you give us a hint on how you use lighthouse? Do you want one of your tests to simply generate a lighthouse report? Do you do it more than once? Do you perform actions after report is generated?

badsyntax:
Hi @pavelfeldman

> Could you give us a hint on how you use lighthouse?

I'm using a [convenience library](https://github.com/abhinaba-ghosh/playwright-lighthouse#readme) to run lighthouse. The library doesn't do much, it essentially calls lighthouse like so:

```ts
import lighthouse from 'lighthouse';

const results = await lighthouse(
  url,
  { disableStorageReset: true, port: port },
  config
);
````

> Do you want one of your tests to simply generate a lighthouse report?

To both generate a report and to fail the tests if the thresholds don't match. 

> Do you do it more than once?

Yes, I'll run lighthouse in different tests. I achieve this without breaking the parallelism by overriding the browser fixture to pass in a random unused remote debugging port which lighthouse uses to connect to the correct browser instance:

```ts
import { chromium } from 'playwright';
import type { Browser } from 'playwright';
import { test as base } from '@playwright/test';
import getPort from 'get-port';

export const lighthouseTest = base.extend<
  {},
[...]

PR #8139, opened 2021-08-11
fix(types): allow specifying scope when overriding fixtures

dgozman:
Otherwise it shows a confusing error.

References #7447.


JoelEinbinder:
I get why we want to allow the scope here, but what does it mean to NOT have the scope for the worker parameters? Looks like a bug to me.

JoelEinbinder (approved):

PR #8140, opened 2021-08-11
fix(protocol): remove unnused strict in QSA and add it to waitForSelector docs

JoelEinbinder:
Strict doesn't make sense in querySelectorAll, but it was in the protocol.

And we had a strict option in ElementHandle.waitForSelector but it
was not surfaced in the docs/api.


docs/src/api/class-elementhandle.md
```diff
@@ -974,3 +974,5 @@ This method does not work across navigations, use [`method: Page.waitForSelector
 ### option: ElementHandle.waitForSelector.state = %%-wait-for-selector-state-%%
 
 ### option: ElementHandle.waitForSelector.timeout = %%-input-timeout-%%
+
+### option: ElementHandle.waitForSelector.strict = %%-input-strict-%%
```
mxschmitt, line 978: note: this is fixing a different docs bug.

mxschmitt (approved):

Issue #8113, opened 2021-08-10
[Feature] DOM Snapshot Testing


rwoll:
I'd like to author tests like

```
expect(await page.domSnapshot()).toMatchSnapshot()
```

where the snapshot serializes to a format similar to Chrome DevTools' Inspect > Elements
tab (which, presentationally, inlines all iframes—including OOPIFs—and adds pseudo elements).

I'd imagine `page.domSnapshot()` to have signature `Promise<playwright.DOMSnapshot>` where `DOMSnapshot` has
a `toString()` or `prettyPrint()` to get the pretty printed document (that looks like Chrome DevTools)
while `DOMSnapshot` itself maintains a walkable tree structure.

NB: The snapshot is a snapshot in time of the DOM and should not hold handles
back into the browser.

I attempted to create such a thing in user-land using DOM.getDocument CDP command
with `pierce = true` and `depth = -1` (no depth limit, IIRC), but OOPIFs were
missed. While you get the OOPIF node, it doesn't have a contentDocument (like
in process frames); it does have a frameId, but I could not find a way to take
this frameId and invoke DOM.getDocument on it recusively via the CDP functionality
expose by Playwright. (I was going to "hydrate" the missing contentDocument's for
the OOPIFs.)

Please let me know your thoughts and I'd be happy to collaborate on a PR!

Follow up from: https://playwright.slack.com/archives/CSUHZPVLM/p1628544879250600

rwoll:
Going to close this as I found tests with these turned out to be less useful than Image/Screenshot based snapshots, and more susceptible to breakage and maintenance overhead.

p01:
By themselves, DOM + CSS snapshot are too noisy, but they are extremely valuable to understand the reason behind visual changes.

dgozman:
@p01 Did you try [Playwright traces](https://playwright.dev/docs/trace-viewer)? They include DOM snapshots for exactly this reason.

p01:
Oh! Of course! Thanks @dgozman 

Reading the docs, my initial thinking is to use [tracing.startChunk(...)](https://playwright.dev/docs/api/class-tracing#tracing-start-chunk) to get a trace as small as possible right around the calls to `page.screenshot(...)` to easily associate the traces and screenshots together.

Or does that sound like a wrong approach ?

dgozman:
@p01 You could try that, but check the trace size to make sure it helps. You can also play with enabling/disabling screenshots.

Tallyb:
[...]

PR #8157, opened 2021-08-12
api(cdp): newCDPSession accepts frames, too

rwoll:
Without this, Playwright's CDP feature leaves unreachable
targets (namely OOPIFs).

This change allows for more advanced experimentation in user-land
without relying on out-of-band CDP connections and clients.

Now you can, for example, call `DOM.getDocument` on the
page OR main frame, observe there is an iframe node with
no `contentDocument` (i.e. OOPIF), make note of the referenced
`frameId`, and then iterate of page.frames() calling `Target.getInfo`
on each to link the Playwright Frame with the CDP `frameId` and
then recurse.

Relates #8113

### DIFF
--- a/src/protocol/protocol.yml
+++ b/src/protocol/protocol.yml
@@ -629,7 +629,8 @@ BrowserContext:
 
     newCDPSession:
       parameters:
-        page: Page
+        page: Page?
+        frame: Frame?
       returns:
         session: CDPSession
 
--- a/src/server/chromium/crBrowser.ts
+++ b/src/server/chromium/crBrowser.ts
@@ -20,6 +20,7 @@ import { assertBrowserContextIsNotOwned, BrowserContext, validateBrowserContextO
 import { assert } from '../../utils/utils';
 import * as network from '../network';
 import { Page, PageBinding, PageDelegate, Worker } from '../page';
+import { Frame } from '../frames';
 import { ConnectionTransport } from '../transport';
 import * as types from '../types';
 import { ConnectionEvents, CRConnection, CRSession } from './crConnection';
@@ -503,10 +504,18 @@ export class CRBrowserContext extends BrowserContext {
     return Array.from(this._browser._serviceWorkers.values()).filter(serviceWorker => serviceWorker._browserContext === this);
   }
 
-  async newCDPSession(page: Page): Promise<CRSession> {
-    if (!(page instanceof Page))
-      throw new Error('page: expected Page');
-    const targetId = (page._delegate as CRPage)._targetId;
+  async newCDPSession(page: Page | Frame): Promise<CRSession> {
+    let targetId: string | null = null;
+    if (page instanceof Page) {
+      targetId = (page._delegate as CRPage)._targetId;
+    } else if (page instanceof Frame) {
+      const session = (page._page._delegate as CRPage)._sessions.get(page._id);
+      if (!session) throw new Error(`This frame does not have a separate CDP session, it is a part of the parent frame's session`);
+      targetId = session._targetId;
+    } else {
+      throw new Error('page: expected Page or Frame');
+    }
+
     const rootSession = await this._browser._clientRootSession();
     const { sessionId } = await rootSession.send('Target.attachToTarget', { targetId, flatten: true });
     return this._browser._connection.session(sessionId)!;
--- a/tests/chromium/session.spec.ts
+++ b/tests/chromium/session.spec.ts
@@ -37,10 +37,15 @@ it('should send events', async function({page, server}) {
   expect(events.length).toBe(1);
 });
 
-it('should only accept a page', async function({page}) {
-  // @ts-expect-error newCDPSession expects a Page
+it('should only accept a page or frame', async function({page}) {
+  // @ts-expect-error newCDPSession expects a Page or Frame
   const error = await page.context().newCDPSession(page.context()).catch(e => e);
-  expect(error.message).toContain('page: expected Page');
+  expect(error.message).toContain('page: expected Page or Frame');
+
+  // non-channelable types hit validation at a different layer
+  // @ts-expect-error newCDPSession expects a Page or Frame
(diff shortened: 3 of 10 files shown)


src/server/chromium/crBrowser.ts
```diff
@@ -503,10 +504,18 @@ export class CRBrowserContext extends BrowserContext {
...
+    if (target instanceof Page) {
+      targetId = (target._delegate as CRPage)._targetId;
+    } else if (target instanceof Frame) {
+      const session = (target._page._delegate as CRPage)._sessions.get(target._id);
+      if (!session) throw new Error('No session for Target ID. This frame might be part of another target');
+      targetId = session._targetId;
+    } else {
+      throw new Error('target: expected Page or Frame');
+    }
+
```
rwoll, line 518: All this could be changed to:
```
const targetId = (target instanceof Page ? target.mainFrame() : target)._id
```
if we don't mind getting `browserContext.newCDPSession: Protocol error (Target.attachToTarget): No target with given id found` for frames that are part of another target. (Since the original code didn't do this, I went the more explicit _targetId finding route).

src/protocol/protocol.yml
```diff
@@ -629,7 +629,8 @@ BrowserContext:
 
     newCDPSession:
       parameters:
-        page: Page
+        page: Page?
+        frame: Frame?
```
rwoll, line 633: Is there a union type/notation to match the TS?

rwoll:
@pavelfeldman This is ready for a review pass! 🎉 Thanks!

vigneshshanmugam:
@rwoll This API looks interesting, Curious if the proposed API would also address the issue I mentioned here - https://github.com/microsoft/playwright/issues/6340#issuecomment-829647616

rwoll:
> @rwoll This API looks interesting, Curious if the proposed API would also address the issue I mentioned here - https://github.com/microsoft/playwright/issues/6340#issuecomment-829647616

@vigneshshanmugam Can you rephrase the question and explain the particular information you're still after? I skimmed through that issue thread, but wasn't quite sure which bits to answer about. Thanks!

vigneshshanmugam:
> Can you rephrase the question and explain the particular information you're still after? I skimmed through that issue thread, but wasn't quite sure which bits to answer about. Thanks!

@rwoll Sure, Currently all user level created CDP sessions are tied to page which means when new frames are created we will miss out on the events that are requested on the new page. Lets take an example
```ts
const context = await browser.newContext(playwrightOptions);
const client = await context.newCDPSession(page);

client.send('Network.enable'); 

client.on('Network.requestWillBeSent', () => {
    // This wont get logged for iframes or pages in new tabs
});
```

I was curios if the proposed API allows for monitoring these subsequent pages. Thanks.

src/server/chromium/crBrowser.ts
```diff
@@ -503,10 +504,18 @@ export class CRBrowserContext extends BrowserContext {
...
-    if (!(page instanceof Page))
-      throw new Error('page: expected Page');
-    const targetId = (page._delegate as CRPage)._targetId;
+  async newCDPSession(target: Page | Frame): Promise<CRSession> {
+    let targetId: string | null = null;
+    if (target instanceof Page) {
+      targetId = (target._delegate as CRPage)._targetId;
+    } else if (target instanceof Frame) {
+      const session = (target._page._delegate as CRPage)._sessions.get(target._id);
+      if (!session) throw new Error('No session for Target ID. This frame might be part of another target');
```
dgozman, line 513: `throw new Error('This frame does not have a separate CDP session, it is a part of the parent frame's session')`

rwoll:
> > Can you rephrase the question and explain the particular information you're still after? I skimmed through that issue thread, but wasn't quite sure which bits to answer about. Thanks!
> 
> 
> 
> @rwoll Sure, Currently all user level created CDP sessions are tied to page which means when new frames are created we will miss out on the events that are requested on the new page. Lets take an example
> 
> ```ts
> 
> const context = await browser.newContext(playwrightOptions);
> 
> const page = await context.newPage();
> 
> const client = await context.newCDPSession(page);
> 
> 
> 
> client.send('Network.enable'); 
> 
> 
> 
> client.on('Network.requestWillBeSent', () => {
> 
>     // This wont get logged for iframes or pages in new tabs
> 
> });
> 
> ```
> 
> 
> 
> I was curios if the proposed API allows for monitoring these subsequent pages. Thanks. 

As is, this won't work, but you could possibly set up sessions on each of the new frames/pages. Still, even that I don't think will work perfectly without race/timing issues so you might miss requests. You can experiment with `Target.setAutoAttach`'s `waitForDebuggerOnStart` to allow yourself time to set up handlers on the new frames before you send `Runtime.runIfWaitingForDebugger`, but that may deadlock coupled with the PW code.

tests/chromium/session.spec.ts
```diff
@@ -37,10 +37,10 @@ it('should send events', async function({page, server}) {
   expect(events.length).toBe(1);
 });
 
-it('should only accept a page', async function({page}) {
-  // @ts-expect-error newCDPSession expects a Page
+it('should only accept a page or frame', async function({page}) {
+  // @ts-expect-error newCDPSession expects a Page or Frame
   const error = await page.context().newCDPSession(page.context()).catch(e => e);
-  expect(error.message).toContain('page: expected Page');
+  expect(error.message).toContain('frame: expected Frame');
```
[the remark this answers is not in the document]
rwoll: I "fixed" this up. The error was automagically coming from channelOwner.ts since context is channelable. Looks like the error might depend on how the object props are iterated!

vigneshshanmugam:
> @vigneshshanmugam Can you file a separate request for this use case? (Or re-open your linked issue.) I'm happy to brainstorm further but want to separate that discussion from this PR. Thanks!

Absolutely, Apologies for opening up the discussion here, Let's chat in a separate issue. Thanks!

dgozman (approved):

PR #8158, opened 2021-08-12
fix(test runner): make sure we always teardown all fixtures

dgozman:
Even if one of the fixtures throws, we should teardown all of them so that we can run afterAll hooks.

### DIFF
--- a/src/test/fixtures.ts
+++ b/src/test/fixtures.ts
@@ -223,14 +223,23 @@ export class FixtureRunner {
   }
 
   async teardownScope(scope: FixtureScope) {
+    let error: Error | undefined;
     // Teardown fixtures in the reverse order.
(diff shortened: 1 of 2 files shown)


src/test/fixtures.ts
```diff
@@ -223,14 +223,23 @@ export class FixtureRunner {
...
+        } catch (e) {
+          if (error === undefined)
+            error = e;
+        }
+      }
     }
     if (scope === 'test')
       this.testScopeClean = true;
+    if (error !== undefined)
+      throw error;
```
JoelEinbinder, line 242: We need a Boolean HasError pattern in case they throw undefined.

JoelEinbinder (approved):

PR #8173, opened 2021-08-12
browser(firefox): support start/stop videos for context

dgozman:


yury-s (approved):
I feel like we are going to have pictures from previous test leaking into the video of the next test sharing the same page as the picture is updated asynchronously from the dom.

PR #8174, opened 2021-08-13
feat(tracing): support recording video through tracing

dgozman:
- `tracing.start()` gets a `video` option.
- When exporting, we generate a few video files next to the trace file and return them as `{ videoFiles: string[] }`.
- Video is stopped on any trace export, until the next `tracing.start()` call.
- This includes the ability to start/stop videos on the live pages, implemented differently in all engines.
- Tracing video is mutually exclusive with `recordVideo` context option.

### DIFF
--- a/docs/src/api/class-tracing.md
+++ b/docs/src/api/class-tracing.md
@@ -136,8 +136,6 @@ Whether to record video for all pages.
 Alternatively, specifies the dimensions of the recorded video. Actual picture of each page will be scaled down if necessary to fit the specified size.
 
 ## async method: Tracing.stop
-- returns: <[Object]>
-  - `videoFiles` <[Array]<[string]>> The list of video files created next to the exported trace file.
 
 Stop tracing.
 
@@ -145,5 +143,3 @@ Stop tracing.
 - `path` <[path]>
 
 Export trace into the file with the given name.
-
-If started with the `video` option, a few additional video files will be exported next to the trace file.
--- a/docs/src/api/class-tracing.md
+++ b/docs/src/api/class-tracing.md
@@ -3,7 +3,7 @@
 API for collecting and saving Playwright traces. Playwright traces can be opened using the Playwright CLI after
 Playwright script runs.
 
-Start with specifying the folder traces will be stored in:
+Start recording a trace, then perform actions, and later stop recording and export trace to a file.
 
 ```js
 const browser = await chromium.launch();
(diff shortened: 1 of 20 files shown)


docs/src/api/class-tracing.md
```diff
@@ -3,7 +3,7 @@
 API for collecting and saving Playwright traces. Playwright traces can be opened using the Playwright CLI after
 Playwright script runs.
 
-Start with specifying the folder traces will be stored in:
+Start with recording a trace, then perform actions, and later stop recording and export trace to a file.
```
yury-s, line 6: 'Start with recording' => 'Start recording'

docs/src/api/class-tracing.md
```diff
@@ -126,11 +126,24 @@ a timeline preview.
...
+  - `width` <[int]> Video frame width.
+  - `height` <[int]> Video frame height.
+
+Whether to record video for all pages.
+
+Alternatively, specifies the dimensions of the recorded video. Actual picture of each page will be scaled down if necessary to fit the specified size.
+
 ## async method: Tracing.stop
+- returns: <[Object]>
+  - `videoFiles` <[Array]<[string]>> The list of video files created next to the exported trace file.
```
yury-s, line 140: I anticipate users asking how to map the file names to the pages, do we have an answer?

src/client/tracing.ts
```diff
@@ -26,29 +28,46 @@ export class Tracing implements api.Tracing {
...
 
-  private async _doExport(channel: channels.BrowserContextChannel, path: string) {
+  private async _doExport(channel: channels.BrowserContextChannel, path: string): Promise<{ videoFiles: string[] }> {
     const result = await channel.tracingExport();
-    const artifact = Artifact.from(result.artifact);
+    const videoFiles: string[] = [];
+    await Promise.all([
+      this._saveAndDelete(result.trace, path),
+      ...result.video.map((artifact, index) => {
+        const videoPath = addSuffixToFilePath(path, '-video-' + (index + 1), '.webm');
```
yury-s, line 61: I hope the ordinal numbers correspond to start/stop order

src/server/browserContext.ts
```diff
@@ -380,6 +383,21 @@ export abstract class BrowserContext extends SdkObject {
     this.on(BrowserContext.Events.Page, installInPage);
     return Promise.all(this.pages().map(installInPage));
   }
+
+  async startVideoRecording(options: types.VideoOptions) {
+    if (this._videoOptions)
+      await this.stopVideoRecording();
```
yury-s, line 389: This will also stop videos started via context option, won't it?

docs/src/api/class-tracing.md
dgozman: We don't have the answer, but the intended use is for "artifacts for my test", where there is no such question. If anything comes up, we can return more fields here like `Video` objects.

src/client/tracing.ts
dgozman: Start/stop order of what? All these videos were started together with a single `tracing.start` command.

src/server/browserContext.ts
dgozman: Context option is mutually exclusive with tracing video.

PR #8176, opened 2021-08-13
fix(har): favicon redirect handling

### LINKED ISSUE #8177: [BUG] HAR recording throws when favicon is served with a redirect
**Context:**
- Playwright Version: v1.13.1
- Operating System: Windows
- Node.js version: v14.17.0
- Browser: Chromium

We use Playwright to integration test a site that serves a hashed favicon by redirecting (302) /favicon.ico to /favicon-[hash].ico.

Playwright HAR recorder filters out the first favicon request and then throws a TypeError when attempting to attach the first requests url to the redirected request (which is not detected as a favicon).

> TypeError: Cannot read property 'request' of undefined
>
>  at HarTracer._onResponse (node_modules/playwright/lib/server/supplements/har/harTracer.js:183:18)
>  at CRBrowserContext.<anonymous> (node_modules/playwright/lib/server/supplements/har/harTracer.js:68:88)
>  at FrameManager.requestReceivedResponse (node_modules/playwright/lib/server/frames.js:238:36)
>  at CRNetworkManager._onResponseReceived (node_modules/playwright/lib/server/chromium/crNetworkManager.js:314:34)
>  at node_modules/playwright/lib/server/chromium/crConnection.js:158:22
>      at runMicrotasks (<anonymous>)

cjsheets:
Handle log entries for redirects from favicons

Fixes #8177


yury-s (commented):
I believe it'd be better to filter out entire redirect chain that starts from `/favicon.ico` request to avoid problems like this in other places. See https://github.com/microsoft/playwright/blob/f4337ffc1d9b48462e90a3ce0592240cb7039a80/src/server/frames.ts#L262 and similar checks in the same file.

cjsheets:
Agreed, updated this PR to mark each request in the chain as `_isFavicon` so they can be filtered by existing conditions.

I tested the HAR recorder and this resolves the issue.

yury-s (commented):
The change looks good, let's add a test for this.

cjsheets:
Testing favicon is tricky 😄 Confirmed the added test would throw without the change in this PR.

tests/favicon.spec.ts
```diff
@@ -57,3 +73,48 @@ it('should load svg favicon with prefer-color-scheme', async ({page, server, bro
   // Text still being around ensures we haven't actually lost our browser to a crash.
   await page.waitForSelector('text=favicons');
 });
+
+it('should filter favicon and favicon redirects from request logs', async ({server, browserName, channel, headless, asset, contextFactory}, testInfo) => {
```
mxschmitt, line 77: since this is about the HAR feature, I would maybe move it into the har.spec.ts

cjsheets: 👍 good call

yury-s (approved):
Looks good. Let's also add a test that we don't fire 'request','response', `requestfinished` events for `/favicon.ico` and `/favicon-hashed.ico` ?

mxschmitt:
Headful Linux / Chromium seems failing on the bots: https://github.com/microsoft/playwright/pull/8176/checks?check_run_id=3437349359#step:7:68 the other failures are unrelated.

cc @cjsheets

yury-s (approved):

yury-s:
> Done. I re-used the existing strategy for testing favicons. Let me know if there's a more elegant approach.

Thanks and sorry for the delay!

mxschmitt:
@cjsheets it fails on one of our bots, would be awesome if you could take a look: https://github.com/microsoft/playwright/pull/8685/checks?check_run_id=3505959228#step:7:77

mxschmitt:
Fix is here: https://github.com/microsoft/playwright/pull/8738

PR #8195, opened 2021-08-14
fix: throw if intercepted response fails

yury-s:
Depends on #8194

#1774


tests/page/page-request-intercept.spec.ts
```diff
@@ -85,6 +85,30 @@ it('should support fulfill after intercept', async ({page, server}) => {
   expect(await response.text()).toBe('<title>Woof-Woof</title>' + os.EOL);
 });
 
+it.only('should intercept failures', async ({page, browserName, server}) => {
```
aslushnikov, line 88: drop only

aslushnikov (approved):

PR #8215, opened 2021-08-16
docs(js): fix syntax errors at intro page examples

kaangokdemir:
## What Changed?
- Examples (Both JS and TS) at Getting Started page were not compiling after a copy/paste since the JavaScript syntax is false due to missing closing parentheses at attribution selectors. They added.


mxschmitt:
Thank you for your contribution!

PR #8225, opened 2021-08-16
chore(dotnet): have options derive from IOptions

avodovnik:


avodovnik:
https://github.com/microsoft/playwright-dotnet/pull/1682

pavelfeldman:
It is unusual to have a struct / options bag implement an interface. Could you elaborate on your plan?

avodovnik:
Yeah, I'm aware, *however*, the reason I want to do it is, this gives me the option of having extension methods, like `.ConvertTo<T>(this IOptions origin)` which become a convenient way of converting between options, such as when calling `Frame` methods from `Page` (which is quite a lot of the cases).

mxschmitt:
Closing it seems stale and not needed.

PR #8234, opened 2021-08-16
feat(network): add Request.sizes()

### LINKED ISSUE #4469: [Feature] Report page assets size
Hi, 

It would be great if we can get transferred assets size with playwright api also the sizes can be grouped by request type.

When i use page.on('response') event, i can't get real encoded body size with this.

similiar repo: https://github.com/andersonba/puppeteer-assets

Thanks.

mxschmitt:
Fixes #4469

### DIFF
--- a/docs/src/api/class-request.md
+++ b/docs/src/api/class-request.md
@@ -183,6 +183,17 @@ following: `document`, `stylesheet`, `image`, `media`, `font`, `script`, `texttr
 
 Returns the matching [Response] object, or `null` if the response was not received due to error.
 
+## method: Request.sizes
+- returns: <[Object]>
+  - `requestBodySize` <[int]> Size of the request body (POST data payload) in bytes. Set to 0 if there was no body.
+  - `requestHeadersSize` <[float]> Total number of bytes from the start of the HTTP request message until (and including) the double CRLF before the body.
+  - `responseBodySize` <[int]> Size of the received response body in bytes.
+  - `responseHeadersSize` <[float]> Total number of bytes from the start of the HTTP response message until (and including) the double CRLF before the body.
+  - `responseTransferSize` <[float]> Total number of bytes received for the request.
+
+Returns resource size information for given request. Requires the response to be finished via [`method: Response.finished`]
+to ensure the info is available.
+
 ## method: Request.timing
 - returns: <[Object]>
   - `startTime` <[float]> Request start time in milliseconds elapsed since January 1, 1970 00:00:00 UTC
--- a/src/client/network.ts
+++ b/src/client/network.ts
@@ -59,6 +59,7 @@ export class Request extends ChannelOwner<channels.RequestChannel, channels.Requ
   _headers: Headers;
   private _postData: Buffer | null;
   _timing: ResourceTiming;
+  _sizes: RequestSizes = { requestBodySize: 0, requestHeadersSize: 0, responseBodySize: 0, responseHeadersSize: 0, responseTransferSize: 0 };
 
   static from(request: channels.RequestChannel): Request {
     return (request as any)._object;
@@ -165,6 +166,10 @@ export class Request extends ChannelOwner<channels.RequestChannel, channels.Requ
 
   timing(): ResourceTiming {
     return this._timing;
+  }
+
+  sizes(): RequestSizes {
+    return this._sizes;
   }
 
   _finalRequest(): Request {
@@ -366,6 +371,14 @@ export type ResourceTiming = {
   requestStart: number;
   responseStart: number;
   responseEnd: number;
+};
+
+export type RequestSizes = {
+  requestBodySize: number;
+  requestHeadersSize: number;
+  responseBodySize: number;
+  responseHeadersSize: number;
+  responseTransferSize: number;
 };
 
 export class Response extends ChannelOwner<channels.ResponseChannel, channels.ResponseInitializer> implements api.Response {
--- a/src/protocol/protocol.yml
+++ b/src/protocol/protocol.yml
@@ -756,6 +756,7 @@ BrowserContext:
       parameters:
         request: Request
         responseEndTiming: number
+        requestSizes: RequestSizes
         page: Page?
 
     response:
@@ -2231,6 +2232,15 @@ SecurityDetails:
     subjectName: string?
     validFrom: number?
     validTo: number?
+
+RequestSizes:
+  type: object
+  properties:
+    requestBodySize: number
+    requestHeadersSize: number
+    responseBodySize: number
+    responseHeadersSize: number
+    responseTransferSize: number
 
 
 RemoteAddr:
--- a/src/server/network.ts
+++ b/src/server/network.ts
@@ -215,6 +215,16 @@ export class Request extends SdkObject {
       headersSize += header.name.length + header.value.length + 4; // 4 = ': ' + '\r\n'
     return headersSize;
   }
+
+  sizes() {
+    return {
+      requestBodySize: this.bodySize(),
+      requestHeadersSize: this.headersSize(),
+      responseBodySize: this._sizes.responseBodySize,
+      responseHeadersSize: this._existingResponse()!.headersSize(),
+      responseTransferSize: this._sizes.transferSize,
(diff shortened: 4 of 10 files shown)


src/client/network.ts
```diff
@@ -239,6 +251,18 @@ export class InterceptedResponse implements api.Response {
   request(): Request {
     return this._request;
   }
+
+  async headersSize(): Promise<number> {
```
mxschmitt, line 255: not sure about this class, need to look into it.

src/server/webkit/wkPage.ts
```diff
@@ -1051,7 +1051,7 @@ export class WKPage implements PageDelegate {
         validTo: responseReceivedPayload?.response.security?.certificate?.validUntil,
       });
       const { responseBodyBytesReceived, responseHeaderBytesReceived } = event.metrics || {};
-      const transferSize = responseBodyBytesReceived ? responseBodyBytesReceived + (responseHeaderBytesReceived || 0) : undefined;
+      const transferSize = responseBodyBytesReceived !== undefined ? responseBodyBytesReceived + (responseHeaderBytesReceived || 0) : undefined;
```
mxschmitt, line 1054: fixes a bug when responseBodySize was 0.

src/client/network.ts
```diff
@@ -165,6 +165,18 @@ export class Request extends ChannelOwner<channels.RequestChannel, channels.Requ
     return this._timing;
   }
 
+  async bodySize(): Promise<number> {
```
pavelfeldman, line 168: Should we expose request.sizes instead next to timing to keep it easily accessible?

src/server/network.ts
```diff
@@ -193,6 +193,22 @@ export class Request extends SdkObject {
       this._headersMap.set('host', host);
     }
   }
+
+  bodySize(): number {
+    const postData = this.postDataBuffer();
+    if (!postData)
+      return 0;
+    return new TextEncoder().encode(postData.toString('utf8')).length;
```
pavelfeldman, line 201: Why do we encode utf8 here? I would imagine we just take the buffer size.

src/server/network.ts
```diff
@@ -193,6 +193,22 @@ export class Request extends SdkObject {
...
+  }
+
+  headersSize(): number {
+    if (!this._response)
+      return 0;
+    let rawHeaders = `${this._method} ${(new URL(this.url())).pathname} ${this._response.httpVersion()}\r\n`;
+    for (const header of this._headers)
+      rawHeaders += `${header.name}: ${header.value}\r\n`;
```
pavelfeldman, line 209: You don't need to allocate lots of memory to sum up the size.

src/server/network.ts
```diff
@@ -401,6 +417,32 @@ export class Response extends SdkObject {
...
+  httpVersion(): string {
+    if (!this._httpVersion)
+      return 'HTTP/1.1';
+    if (this._httpVersion === 'http/1.1')
+      return 'HTTP/1.1';
+    return this._httpVersion;
+  }
+
+  headersSize(): number {
+    let rawHeaders = `${this.httpVersion()} ${this.status()} ${this.statusText()}\r\n`;
```
pavelfeldman, line 440: ditto.

Unfortunately this is all approximation, I wonder if users are going to be happy about it. It'll be not worse than in DevTools, but different from real network traffic due to headers encoding.

docs/src/api/class-request.md
```diff
@@ -16,6 +16,11 @@ with `'requestfinished'` event.
 If request gets a 'redirect' response, the request is successfully finished with the 'requestfinished' event, and a new
 request is  issued to a redirected url.
 
+## method: Request.bodySize
```
pavelfeldman, line 19: I remember suggesting Request.sizes next to Request.timing

src/protocol/protocol.yml
```diff
@@ -731,6 +731,9 @@ BrowserContext:
       parameters:
         request: Request
         responseEndTiming: number
+        responseTransferSize: number
```
pavelfeldman, line 734: We should use intermediate structs, can be RequestSizes for example and we can update / augment them incrementally, but should be a separate struct.

docs/src/api/class-request.md
```diff
@@ -183,6 +183,17 @@ following: `document`, `stylesheet`, `image`, `media`, `font`, `script`, `texttr
 
 Returns the matching [Response] object, or `null` if the response was not received due to error.
 
+## method: Request.sizes
+- returns: <[Object]>
+  - `requestBodySize` <[int]> Size of the request body (POST data payload) in bytes. Set to 0 if there was no body.
+  - `requestHeadersSize` <[float]> Total number of bytes from the start of the HTTP request message until (and including) the double CRLF before the body.
+  - `responseBodySize` <[int]> Size of the received response body in bytes.
```
pavelfeldman, line 190: Do we have a good idea on why users could want original body size? And why they don't want the actual transfer size for the request body?

src/server/network.ts
```diff
@@ -401,6 +430,35 @@ export class Response extends SdkObject {
...
   }
+
+  transferSize(): number | undefined {
+    return this._transferSize;
+  }
+
+  bodySize(): number | undefined {
+    if (!this._transferSize)
+      return;
+    return this._transferSize - this.headersSize();
```
pavelfeldman, line 441: You are going to see some negative sizes here.

src/server/network.ts
```diff
@@ -193,6 +193,35 @@ export class Request extends SdkObject {
...
+    return postData.length;
+  }
+
+  headersSize(): number {
+    if (!this._response)
+      return 0;
+    let headersSize = 4; // 4 = 2 spaces + 2 line breaks (GET /path \r\n)
+    headersSize += this.method().length;
+    headersSize += (new URL(this.url())).pathname.length;
+    headersSize += this._response.httpVersion().length;
```
pavelfeldman, line 210: Just add 7

src/server/network.ts
```diff
@@ -401,6 +430,35 @@ export class Response extends SdkObject {
...
+      return 'HTTP/1.1';
+    if (this._httpVersion === 'http/1.1')
+      return 'HTTP/1.1';
+    return this._httpVersion;
+  }
+
+  headersSize(): number {
+    let headersSize = 4; // 4 = 2 spaces + 2 line breaks (HTTP/1.1 200 Ok\r\n)
+    headersSize += this.httpVersion().length;
+    headersSize += this.status().toString().length;
```
pavelfeldman, line 455: I'm going to predict this is 3

tests/config/checkCoverage.js
```diff
@@ -51,6 +51,11 @@ if (browserName !== 'chromium') {
 if (browserName === 'webkit')
   api.delete('browserContext.clearPermissions');
 
+// WebKit on Linux has no sizes caused by libsoup, waiting for libsoupv3
+if (browserName === 'webkit') {
+  api.delete('request.sizes');
```
pavelfeldman, line 56: I think we have body / transfer sizes through dataReceived in WebKit, don't we?

mxschmitt: Not on Linux unfortunately, there it's blocked by libsoup3.

src/server/network.ts
mxschmitt: In theory transfer size is always bigger, what would you change here?

docs/src/api/class-request.md
mxschmitt: It's from the HAR spec and I thought it's about exposing these values.

pavelfeldman: You should look at har spec and re-read my question!

src/server/network.ts
pavelfeldman: The logic that you use when computing sizes is fuzzy, so you should tolerate inaccurate numbers, capping this to zero. Does transfer size include headers?

tests/config/checkCoverage.js
pavelfeldman: So no data size in dataReceived?

pavelfeldman (approved):

PR #8241, opened 2021-08-16
fix(snapshots): account for malformed headers

### LINKED ISSUE #8095: trace-viewer doesn't show formatted (main screen)
### Playwright version

1.13.1

### Operating system

Windows

Chromium

### Other information

nodejs 14.1
windows 8
robotframework 4.1
robotframework-browser 7.0.0 (using playwright 1.13.1)

I created a script to create a trace with the below robot script. after getting to the trace it shows the following screen
image

What's strange about this screen, that the main screen doesn't contain any formatting. It looks like CSS is completely ignored. And the red dot indicating what has been pressed (the whole reason for this test), is not pointing to a clickable element.

I have seen this behavior in different versions of robotframework-browser (at least 6.0.0 and 7.0.0) which used playwright 1.12.3 and 1.13.1 (I think).

The first test I did using traces was in a corporate environment, and I thought that caused this problem. But now I can reproduce this problem on a private laptop, which has no limits.

Any reasons why it doesn't show formatted screens (but is does in the timeline) ?
I already created an issue with the robotframework-browser team, but they directed me here.
I would assume playwright trace viewer works as depicted in following [youtube](https://www.youtube.com/watch?v=Zutt7N6VvcU) clip. Which seems awesome

```shell
*** Settings ***
Library  Browser

*** Test Cases ***
Enable Tracing To File
    New Browser  chromium    headless=false
    New Context    tracing=trace.zip
    New Page    https://duckduckgo.com/
	
    Fill Text  id=search_form_input_homepage    tweakers.net
    Click    id=search_button_homepage
    Close Context
```

### Relevant log output

_No response_

dgozman:
When browser receives multiple header values for the same header name, we get them as LF-separated value. This is not considered valid in Node http, so we should split by LF when serving a snapshot.

There are more invalid characters in headers, so just in case we try/catch it.

Fixes #8095.

### DIFF
--- a/src/server/snapshot/snapshotServer.ts
+++ b/src/server/snapshot/snapshotServer.ts
@@ -183,8 +183,14 @@ export class SnapshotServer {
       if (isTextEncoding && !contentType.includes('charset'))
         contentType = `${contentType}; charset=utf-8`;
       response.setHeader('Content-Type', contentType);
-      for (const { name, value } of resource.responseHeaders)
-        response.setHeader(name, value);
+      for (const { name, value } of resource.responseHeaders) {
+        try {
+          response.setHeader(name, value.split('\n'));
+        } catch (e) {
+          // Browser is able to handle the header, but Node is not.
+          // Swallow the error since we cannot do anything meaningful.
+        }
+      }
 
       response.removeHeader('Content-Encoding');
       response.removeHeader('Access-Control-Allow-Origin');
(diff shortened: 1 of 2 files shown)


pavelfeldman:
This is browser-specific IIRC, webkit would have coma

pavelfeldman (approved):

dgozman:
> This is browser-specific IIRC, webkit would have coma

For the record, this is indeed the case. However, we do not have ways to detect that currently and, fortunately, it does not break snapshot rendering.

PR #8242, opened 2021-08-17
docs: add sections for describe.serial and reusing the page

dgozman:


docs/src/test-parallel-js.md
```diff
@@ -3,15 +3,27 @@ id: test-parallel
...
 ---
 
-Playwright Test runs tests in parallel by default, using multiple worker processes.
+Playwright Test uses multiple worker processes to run tests in parallel. Each worker process creates a new environment to run tests. By default, Playwright Test reuses the worker as much as it can to make testing faster.
 
-<!-- TOC -->
+Should any test fail, Playwright will discard entire worker process along with the browsers used and will start a new one. That way failing tests can't affect healthy ones.
 
-## Workers
+* By default, Playwright test [runs test files in parallel](#running-test-files-in-parallel).
```
yury-s, line 10: . Tests from the same file will run sequentially on one and the same worker.

Or just inline linked section here otherwise it's hard to read.

docs/src/test-parallel-js.md
```diff
@@ -57,7 +69,99 @@ You can control the maximum number of parallel worker processes via [command lin
...
+```js js-flavor=js
+// example.spec.js
+// @ts-check
+
+const { test } = require('@playwright/test');
+
+/** @type {import('@playwright/test').Page} */
+let page;
+
+test.beforeAll(async ({ browser }) => {
```
yury-s, line 119: having this outside of `test.describe.serial` blows my mind, can it be put inside describe block in this example?

yury-s (approved):

docs/src/test-parallel-js.md
```diff
@@ -3,61 +3,155 @@ id: test-parallel
 title: "Parallelism and sharding"
 ---
 
-Playwright Test runs tests in parallel by default, using multiple worker processes.
+Playwright Test uses multiple worker processes to run tests in parallel. Each worker process creates a new environment to run tests. By default, Playwright Test reuses the worker as much as it can to make testing faster.
```
pavelfeldman, line 6: I am not sure this is the language we want to speak to the @playwright/test customers. We just need to say that tests run in parallel, we don't need to talk about the worker processes right away?

pavelfeldman (approved):

PR #8245, opened 2021-08-17
docs(network): add notes, tests re: proxy loopback

rwoll:
Chromium's behavior diverges from the other browsers regarding implicit
proxy bypass of some loopback and local addresses.

While debugging tests, I sometimes use an HTTP Proxy debugger and/or
external HAR capture and Chrome was not forwarding requests I expected
to end up at the proxy.

This change documents this behavior as well as the obscure (to me) work
around derived from [Chromium source code][src-code].

[src-code]: https://source.chromium.org/chromium/chromium/src/+/main:net/proxy_resolution/proxy_bypass_rules.cc

### DIFF
--- a/docs/src/network.md
+++ b/docs/src/network.md
@@ -71,6 +71,19 @@ for the entire browser, or for each browser context individually.
 You can optionally specify username and password for HTTP(S) proxy, you can also specify hosts to
 bypass proxy for.
 
+:::note
+Chromium, unlike WebKit and Firefox, will **not** proxy localhost, link locals
+(i.e. 169.254/16), and loopback addresses by default (unless
+running in headful mode on Linux).
+
+If you need these requests (e.g. `page.goto("http://localhost:3030")`) to route
+through the configured proxy, add the special token `<-loopback>` to the `bypass`
+option.
+
+Do not pass `<-loopback>` to non-Chromium browsers as behavior will be undefined
+or broken.
+:::
+
 Here is an example of a global proxy:
 
 ```js
--- a/tests/proxy.spec.ts
+++ b/tests/proxy.spec.ts
@@ -17,6 +17,7 @@
 import { playwrightTest as it, expect } from './config/browserTest';
 import socks from 'socksv5';
 import net from 'net';
+import { LaunchOptions } from '../index';
 
 it('should throw for bad server value', async ({browserType, browserOptions}) => {
   const error = await browserType.launch({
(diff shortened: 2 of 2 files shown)


tests/proxy.spec.ts
```diff
@@ -75,6 +76,64 @@ it('should work with IP:PORT notion', async ({browserType, browserOptions, serve
...
+    socket.on('data', () => {
+      proxiedRequest = true;
+      socket.end();
+    });
+  });
+  await new Promise(f => server.listen(0, f));
+
+  const opts: LaunchOptions = {
+    ...browserOptions,
+    proxy: { server: `http://127.0.0.1:${(server.address() as any).port}` },
```
mxschmitt, line 91: You can also pass this directly in the [proxy options](https://playwright.dev/docs/api/class-browser#browser-new-context-option-proxy). Passing the arg directly is not needed.

```suggestion
    proxy: { server: `http://127.0.0.1:${(server.address() as any).port}`, bypass: '<-loopback>' },
```

tests/proxy.spec.ts
```diff
@@ -75,6 +76,64 @@ it('should work with IP:PORT notion', async ({browserType, browserOptions, serve
...
+  await page.goto('http://localhost:9999/request').catch(() => {});
+  await browser.close();
+  server.close();
+  if (browserType.name() === 'chromium')
+    expect(proxiedRequest).toBeFalsy();
+  else
+    expect(proxiedRequest).toBeTruthy();
+});
+
+it('should work with localhost', async ({ browserType, browserOptions }) => {
```
mxschmitt, line 107: not sure if a custom test for that is needed, since we are already testing this behaviour in the `portForwardingServer.spec.ts`

rwoll: I left this test in place as it actually had to get a bit more complicated because of platform quirks. The tests assert what the new documentation claims to be the state of the world.

docs/src/network.md
```diff
@@ -71,6 +71,19 @@ for the entire browser, or for each browser context individually.
...
+:::note
+Chromium, unlike WebKit and Firefox, will **not** proxy localhost, link locals
+(i.e. 169.254/16), and loopback addresses by default (unless
+running in headful mode on Linux).
+
+If you need these requests (e.g. `page.goto("http://localhost:3030")`) to route
+through the configured proxy, add the special token `<-loopback>` to the `bypass`
+option.
+
+Do not pass `<-loopback>` to non-Chromium browsers as behavior will be undefined
```
rwoll, line 83: I added this note since if you remove the following guard, you get a WK fail on macOS headless: https://github.com/microsoft/playwright/pull/8245/files#diff-6abdc33c61672400fa977184bbc3bc7f7ec0ea6873cc6e9d52e93a241c186633R122

So you must conditionally include the special bypass token. (I know the token is Chrome-specific, but I wish we didn't have to branch logic.)

tests/proxy.spec.ts
```diff
@@ -75,6 +76,62 @@ it('should work with IP:PORT notion', async ({browserType, browserOptions, serve
...
+  };
+
+  const browser = await browserType.launch(opts);
+
+  const page = await browser.newPage();
+
+  await page.goto('http://localhost:9999/request').catch(() => {});
+  await browser.close();
+  server.close();
+  if (browserType.name() !== 'chromium' || platform === 'linux' && !browserOptions.headless)
```
rwoll, line 101: From previous CI run, it turns out Headful Chrome on Linux is unique and does pass by default: https://github.com/microsoft/playwright/runs/3356777516?check_suite_focus=true

rwoll (commented):
@mxschmitt and @pavelfeldman : I updated the MR based on the full CI/CD run. Interestingly, there are platform differences. Heh! (See review comments.)

Instead of adding this documentation note that leaks Chromium implementation details and forces the user to branch based on platform+browser+headless, I think it might be cleaner to remove the documentation note proposed here, and fix behind the scenes so the browsers across platforms and headless/headful mode work identically.

Does that sound like a good idea? What might break for users by changing the default behavior (i.e. adding `<-loopback>` within the CR launch internal impl)?

mxschmitt:
@rwoll how do you feel about passing this flag by default to chromium when proxies are used?

rwoll:
> @rwoll how do you feel about passing this flag by default to chromium when proxies are used?
> 
> 
> 
> Or maybe assume that bypass includes localhost by default and make it work on all browsers?
> 
> 
> 
> Curious to hear your thoughts on that!

@mxschmitt I'm all for it! I had to discover chrome worked differently the hard way, so I'm all for making the browser behave similarly as possible out of the Playwright box! 

Regarding this, my question from https://github.com/microsoft/playwright/pull/8245#pullrequestreview-734687121 still remains: will we break any users current setups by changing the Playwright default at this point? If it's negligible or manageable, I'd vote to add this to the default args.

rwoll:
Closing for now for hygiene, but if this comes up again, I'd vote to internally change the default (not upstream, but in The default launch args we supply when launching) so Chromium matches the other browsers out of the box.

rwoll:
Re-opening since this came up with another user so a the doc change I had would at least be nice: https://github.com/microsoft/playwright/issues/10631

rwoll:
Closing! I'll make a new MR that proxies by default.

PR #8248, opened 2021-08-17
feat(types): export PageWaitForSelectorOptions

### LINKED ISSUE #8247: [Feature] Export PageWaitForSelectorOptions
I'm implementing simple wrapper over `page` object on TypeScript and it would be very useful if I could determine incoming types for `page.waitForSelector` options, so can you allow export of PageWaitForSelectorOptions?

zefirka:
Fixes #8247

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -12822,7 +12822,7 @@ export interface Cookie {
   sameSite: "Strict"|"Lax"|"None";
 }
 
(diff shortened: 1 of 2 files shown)


yury-s:
We are not planning to export options types, most of the other option types are actually defined inline. Having `PageWaitForSelectorOptions` as a named type in the api was a mistake and we'd rather align it with other option definitions by removing named type/interface.

PR #8259, opened 2021-08-17
fix(selectors): nicer errors if the selector engine returns a bad value

JoelEinbinder:
A selector engine returning something other than elements
would fail silently and return cryptic errors later on.﻿

### DIFF
--- a/tests/selectors-register.spec.ts
+++ b/tests/selectors-register.spec.ts
@@ -134,7 +134,7 @@ it('should not rely on engines working from the root', async ({ playwright, brow
   expect(await page.$eval('input >> __value=value2', e => e.id)).toBe('input2');
 });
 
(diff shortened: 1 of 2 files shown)


tests/selectors-register.spec.ts
```diff
@@ -133,3 +133,19 @@ it('should not rely on engines working from the root', async ({ playwright, brow
   await page.setContent(`<input id=input1 value=value1><input id=input2 value=value2>`);
   expect(await page.$eval('input >> __value=value2', e => e.id)).toBe('input2');
 });
+
+it.only('should throw a nice error if the selector returns a bad value', async ({ playwright, browser }) => {
```
pavelfeldman, line 137: it.only

pavelfeldman (approved):

mxschmitt (approved):

PR #8282, opened 2021-08-18
fix(test-runner): use describe name in output dir

### LINKED ISSUE #8276: [BUG] `describe` title not part of `outputDir`
**Context:**

## System:
 - OS: Linux 5.10 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
 - Memory: 17.90 GB / 31.09 GB
 - Container: Yes
## Binaries:
 - Node: 15.14.0 - ~/miniconda3/envs/jlab-dev/bin/node
 - Yarn: 1.21.1 - ~/projects/jupyterlab/node_modules/.bin/yarn
 - npm: 7.7.6 - ~/miniconda3/envs/jlab-dev/bin/npm
## Languages:
 - Bash: 5.1.8 - /usr/bin/bash

- @playwright/test 1.14.0

I got a question regarding the test naming. When using `test.describe` the test name does not include that title in the output directory name. This may result in name clashes.

will result in the a name test-fixture-should-return-mocked-settings.

I would expect something like test-fixture-mockSettings-should-return-mocked-settings

**Code Snippet**

For example, if in file named fixture.spec.ts:

```javascript
test.describe('mockSettings', () => {
  test('should return mocked settings', async ({ page }) => {
    ...
  });
});
```

will result in a output dir named `test-fixture-should-return-mocked-settings`.

I would expect something like `test-fixture-mockSettings-should-return-mocked-settings`

This can be seen in on that PR: https://github.com/jupyterlab/jupyterlab/pull/10796
Test artifacts can be found in the `Galata Tests` workflow.

JoelEinbinder:
Fixes #8276﻿

### DIFF
--- a/src/test/workerRunner.ts
+++ b/src/test/workerRunner.ts
@@ -201,7 +201,8 @@ export class WorkerRunner extends EventEmitter {
     const baseOutputDir = (() => {
       const relativeTestFilePath = path.relative(this._project.config.testDir, test._requireFile.replace(/\.(spec|test)\.(js|ts|mjs)$/, ''));
       const sanitizedRelativePath = relativeTestFilePath.replace(process.platform === 'win32' ? new RegExp('\\\\', 'g') : new RegExp('/', 'g'), '-');
-      let testOutputDir = sanitizedRelativePath + '-' + sanitizeForFilePath(test.title);
+      const fullTitleWithoutSpec = test.titlePath().slice(1).join(' ');
+      let testOutputDir = sanitizedRelativePath + '-' + sanitizeForFilePath(fullTitleWithoutSpec);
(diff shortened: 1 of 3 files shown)


mxschmitt (commented):
~~note: It's considered as a bug fix but might break people and lead to new generated snapshot files.~~

JoelEinbinder:
> note: It's considered as a bug fix but might break people and lead to new generated snapshot files.

We discussed offline and decided this is not a breaking change. It changes the `outputPath`, but not the `snapsshotPath` which is what people would be committing into their repos.

dgozman (approved):

mxschmitt (approved):

PR #8290, opened 2021-08-18
feat(strict): add context-level selector strictness

pavelfeldman:


docs/src/api/params.md
```diff
@@ -511,6 +511,51 @@ contexts override the proxy, global proxy will be never used and can be any stri
...
+- `strictSelectors` <[boolean]>
+
+It specified, enables strict selectors mode for this context. In the strict selectors mode all operations
+on selectors that imply single target DOM element will throw when more than one element matches the selector.
+
+```js
+// Throws if there are several buttons in DOM:
+await page.click('button');
+
+// Works because we explicitly tell locator to pick the first element:
```
dgozman, line 524: locator -> Playwright?

dgozman (approved):

PR #8299, opened 2021-08-19
test: proxy server fixture, new test for https via http proxy

yury-s:
Existing proxy tests relied on the browser to send HTTP requests to the proxy as regular GET requests, whereas HTTPS via HTTP proxy requires handling CONNECT request first to establish connection to the target server. The new test executes that path. The proxy server is also going to be used to test new proxy support for the new `BrowserContext.fetch()` method.

### DIFF
--- a/tests/browsercontext-proxy.spec.ts
+++ b/tests/browsercontext-proxy.spec.ts
@@ -115,6 +115,26 @@ it('should use proxy for second page', async ({contextFactory, server}) => {
   await context.close();
 });
 
+it('should use proxy for https urls', async ({ contextFactory, server, httpsServer, proxyServer }) => {
+  httpsServer.setRoute('/target.html', async (req, res) => {
+    res.end('<html><title>Served by the proxy</title></html>');
+  });
+  let connectedToProxy = false;
+  proxyServer.onConnect(req => {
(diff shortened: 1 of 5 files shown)


tests/browsercontext-proxy.spec.ts
```diff
@@ -115,6 +115,26 @@ it('should use proxy for second page', async ({contextFactory, server}) => {
...
+    res.end('<html><title>Served by the proxy</title></html>');
+  });
+  let connectedToProxy = false;
+  proxyServer.onConnect(req => {
+    req.url = `localhost:${httpsServer.PORT}`;
+    connectedToProxy = true;
+  });
+  const context = await contextFactory({
+    ignoreHTTPSErrors: true,
+    proxy: { server: `localhost:${proxyServer.PORT}` }
```
dgozman, line 129: Should we migrate more/all tests here to the new proxy?

dgozman (approved):

aslushnikov (approved):

tests/browsercontext-proxy.spec.ts
yury-s: Probably, will have a look.

Issue #8045, opened 2021-08-06
[Internal] Drop 32 bit Windows builds


mxschmitt:
Windows 32 bit is not used anymore. The measurements which we currently use (VSC, browser download stats) are not correct, because os.arch() returns also 32 bit on a 64 system. In Python or other languages this is e.g. not the case and returns 64 bit instead (for driver downloads e.g.).

![image](https://user-images.githubusercontent.com/17984549/128538258-7dbc853c-7440-49bf-bcc9-841e88571bee.png)

ref: https://www.pcbenchmarks.net/os-marketshare.html

mxschmitt:
Merged the drop of 32 bit and moved to 1.17 the removal of the infrastructure changes.

aslushnikov:
this is fixed

PR #8306, opened 2021-08-19
chore: pretend world is win64

aslushnikov:
It looks like we have a very small share of users on 32-bit windows.
This is a tiny change that stops Playwright from using 32-bit windows
binaries.

References #8045


mxschmitt (approved):
nice small experiment! and in 1-2 versions we can drop the build infra.

PR #8313, opened 2021-08-19
feat: support multiple roots in React and Vue selectors

### LINKED ISSUE #8230: Using React Selectors with multiple React trees
Hello,
Excited to see the new React Selector api! Our site uses two react trees and I wondering if there can be support for multiple trees.

I have my own custom react selector on top of the Selector api and this is what I do today:
image

Thanks!

aslushnikov:
Fixes #8230

### DIFF
--- a/src/server/injected/vueSelectorEngine.ts
+++ b/src/server/injected/vueSelectorEngine.ts
@@ -204,27 +204,34 @@ function filterComponentsTree(treeNode: ComponentNode, searchFn: (node: Componen
   return result;
 }
 
-function findVueRoot(): undefined|{version: number, root: VueVNode} {
-  const walker = document.createTreeWalker(document);
-  while (walker.nextNode()) {
-    // Vue3 root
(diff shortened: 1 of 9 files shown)


JoelEinbinder (approved):
cool!

src/server/injected/vueSelectorEngine.ts
```diff
@@ -233,9 +240,8 @@ export const VueEngine: SelectorEngine = {
         if (!checkComponentAttribute(treeNode.props, attr))
           return false;
       }
-
       return true;
-    });
+    })).flat();
```
dgozman, line 244: Is this available in all our browsers?

dgozman (approved):
Let's add a test where I can find something in a particular root, like `#root1 >> react=MyBook` or something like that.

src/server/injected/vueSelectorEngine.ts
aslushnikov: yes

PR #8327, opened 2021-08-19
fix(test runner): avoid internal error for step end without begin

dgozman:
Consider the following scenario:
- Test finishes and starts tearing down fixtures.
- Fixture teardown starts a step `S` and then times out.
- We declare the test finished (with timeout).
- Dispatcher shuts down the worker and spins a new one for a retry. Additionally, it clears steps information for the test to be ready for the new retry. Step `S` information is lost.
- Meanwhile, during worker teardown, the step `S` does actually finish (usually with an error), and we send `stepEnd` for `S`.
- Dispatcher does not know what to do with step `S` end and prints an internal error.

The fix is to ignore certain messages from the shutting down worker that failed.

### DIFF
--- a/src/test/dispatcher.ts
+++ b/src/test/dispatcher.ts
@@ -124,6 +124,7 @@ export class Dispatcher {
 
       // When worker encounters error, we will stop it and create a new one.
       worker.stop();
+      worker.didFail = true;
 
       const retryCandidates = new Set<string>();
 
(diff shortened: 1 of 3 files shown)


pavelfeldman (approved):

pavelfeldman (approved):

ematta:
Apologies if this is no the avenue to ask this. When will this be released as part of 1.14.x? I am experiencing this issue currently in my CircleCI builds.

mxschmitt:
You can try it out today via `npm i @playwright/test@next` or wait until 1.15.0 in around 1 week.

PR #8329, opened 2021-08-19
chore: move sdkLanguage over to the protocol init call

### LINKED ISSUE #7741: [Bug]: Playwright executable path not found message should be cross-language compatible

mxschmitt:
Fixes #7741

### DIFF
--- a/src/dispatchers/dispatcher.ts
+++ b/src/dispatchers/dispatcher.ts
@@ -125,16 +125,16 @@ export type DispatcherScope = Dispatcher<any, any>;
 export class Root extends Dispatcher<{ guid: '' }, {}> {
   private _initialized = false;
 
-  constructor(connection: DispatcherConnection, private readonly createPlaywright?: (scope: DispatcherScope) => Promise<PlaywrightDispatcher>) {
+  constructor(connection: DispatcherConnection, private readonly createPlaywright?: (scope: DispatcherScope, options: channels.RootInitializeParams) => Promise<PlaywrightDispatcher>) {
     super(connection, { guid: '' }, 'Root', {}, true);
   }
 
-  async initialize(params: { language?: string }): Promise<channels.RootInitializeResult> {
+  async initialize(params: channels.RootInitializeParams): Promise<channels.RootInitializeResult> {
     assert(this.createPlaywright);
     assert(!this._initialized);
     this._initialized = true;
     return {
-      playwright: await this.createPlaywright(this),
+      playwright: await this.createPlaywright(this, params),
     };
   }
 }
--- a/src/server/chromium/crPage.ts
+++ b/src/server/chromium/crPage.ts
@@ -36,7 +36,7 @@ import { CRPDF } from './crPdf';
 import { CRBrowserContext } from './crBrowser';
 import * as types from '../types';
 import { rewriteErrorMessage } from '../../utils/stackTrace';
-import { assert, headersArrayToObject, createGuid, canAccessFile } from '../../utils/utils';
+import { assert, headersArrayToObject, createGuid } from '../../utils/utils';
 import { VideoRecorder } from './videoRecorder';
 import { Progress } from '../progress';
 import { DragManager } from './crDragDrop';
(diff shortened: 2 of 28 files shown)


src/server/chromium/crPage.ts
```diff
@@ -848,18 +848,12 @@ class FrameSession {
     const ffmpegPath = registry.findExecutable('ffmpeg')!.executablePath();
     // TODO: use default error message once it's ready.
     if (!ffmpegPath || !canAccessFile(ffmpegPath)) {
```
dgozman, line 850: Why don't we use executablePathOrDie() here? We can append `Please install ffmpeg in order to record video.` to the error, or pass custom text to `executablePathOrDie` instead. It already has a nice `<3 Playwright Team` message.

src/server/playwright.ts
```diff
@@ -54,6 +55,6 @@ export class Playwright extends SdkObject {
   }
 }
 
-export function createPlaywright(isInternal = false) {
-  return new Playwright(isInternal);
+export function createPlaywright(isInternal: boolean = false, sdkLanguage: string = 'javascript') {
```
dgozman, line 58: I'd prefer sdkLanguage to be required.

src/utils/registry.ts
```diff
@@ -660,6 +660,18 @@ function markerFilePath(browserDirectory: string): string {
   return path.join(browserDirectory, 'INSTALLATION_COMPLETE');
 }
 
+export function buildPlaywrightCLICommand(cliLanguage: string, parameters: string): string {
+  const cliLanguageToCLIBuilder: Partial<Record<string, (parameters: string) => string>> = {
```
dgozman, line 664: Why Partial?

src/utils/registry.ts
```diff
@@ -660,6 +660,18 @@ function markerFilePath(browserDirectory: string): string {
...
+    'python': (parameters: string) => `playwright ${parameters}`,
+    'java': (parameters: string) => `mvn exec:java -e -Dexec.mainClass=com.microsoft.playwright.CLI -Dexec.args="${parameters}"`,
+    'csharp': (parameters: string) => `playwright ${parameters}`,
+    'javascript': (parameters: string) => `npx playwright ${parameters}`,
+  };
+  const build = cliLanguageToCLIBuilder[cliLanguage];
+  assert(build, `Unknown CLI language: ${cliLanguage}`);
```
dgozman, line 671: Let's fallback to javascript instead.

src/dispatchers/dispatcher.ts
```diff
@@ -125,16 +125,16 @@ export type DispatcherScope = Dispatcher<any, any>;
 export class Root extends Dispatcher<{ guid: '' }, {}> {
   private _initialized = false;
 
-  constructor(connection: DispatcherConnection, private readonly createPlaywright?: (scope: DispatcherScope) => Promise<PlaywrightDispatcher>) {
+  constructor(connection: DispatcherConnection, private readonly createPlaywright?: (scope: DispatcherScope, options: channels.RootInitializeParams) => Promise<PlaywrightDispatcher>) {
```
dgozman, line 128: Let's rename language to sdkLanguage for consistency.

src/server/chromium/crPage.ts
mxschmitt: good catch! one less exported function.

src/utils/registry.ts
mxschmitt: This ends up in a `string` type otherwise. I rewrote this part, so not needed anymore.

dgozman (approved):

PR #8336, opened 2021-08-20
chore: additional test mode

pavelfeldman:


tests/config/baseTest.ts
```diff
@@ -92,6 +93,19 @@ class ServiceMode {
   }
 }
 
+class GridMode {
+  private _client: any;
```
mxschmitt, line 97:
```suggestion
  private _client: import('../../src/remote/ghaGridClient');
```
nit

mxschmitt (approved):

tests/config/baseTest.ts
pavelfeldman: I don't think that's right...

Issue #5999, opened 2021-03-30
[Question] Is it possible to make an HttpRequest through Playwright?


symon-skelly:
I currently use Cypress for my automation testing framework, however I am toying with the idea of moving over to Playwright. One feature of cypress that really shines in the ability to make HttpRequests using the [cy.request()](https://docs.cypress.io/api/commands/request) function, I use this many times throughout the current system however the most important request is used to setup my randomly generated user application state before I even visit the web-app itself (it makes the request to our webServer which responds with data that is then used for the user-specific localStorage items), this allows me to skip the login screen of my application entirely, saving a lot of time for each test. 

I have read the docs about how Playwright can re-use application state but this is not really a viable solution for me because after each test I specifically use new application state. 

TLDR: Is it possible to make an HttpRequest through Playwright and get its response?

mxschmitt:
You can use the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) inside the browser to make http requests. Example for Playwright:

```js
// @ts-check
const playwright = require('playwright');

(async () => {
  const browser = await playwright.chromium.launch();
  const context = await browser.newContext();
  const page = await context.newPage();
  await page.goto('http://example.com');
  const response = await page.evaluate(async () => {
    return await fetch("http://worldtimeapi.org/api/timezone")
      .then(r => r.ok ? r.json() : Promise.reject(r))
  })
  console.log(response)
  await browser.close();
})();
```

Inside of evaluate you can access the LocalStorage etc. To pass arguments over to the browser process, you can pass it to `page.evaluate()` in the second argument. See [here](https://playwright.dev/docs/api/class-page#pageevaluatepagefunction-arg) for the docs.

symon-skelly:
That looks like a solution that could work for me, thank you for your fast reply!

pavelfeldman:
It looks like what you really want is this: https://playwright.dev/docs/auth#reuse-authentication-state. You can even create this state using cli once: https://playwright.dev/python/docs/cli#preserve-authenticated-state

pavelfeldman:
Now that I read you request to the end, I see that you use different state for each test. Is it something like incremental session id that the server keeps tracking of? Figuring out if we can get something out of the box for all the auth scenarios...

shirshak55:
I really don't like this fetch system. Getting raw bytes etc, cors issue (sometime even have to disable security feature) etc. 

[...]

PR #8349, opened 2021-08-21
feat: introduce BrowserContext._fetch

yury-s:
#5999


src/server/browserContext.ts
```diff
@@ -179,6 +183,63 @@ export abstract class BrowserContext extends SdkObject {
     await this._doExposeBinding(binding);
   }
 
+  async fetch(params: types.FetchOptions): Promise<{response?: network.FetchResponse, error?: string}> {
```
pavelfeldman, line 186: It'll probably grow, let's kick it all into a separate fetch.ts file.

src/server/browserContext.ts
```diff
@@ -179,6 +183,63 @@ export abstract class BrowserContext extends SdkObject {
...
+      if (cookieHeader) {
+        if (!params.headers)
+          params.headers = {};
+        params.headers['cookie'] = cookieHeader;
+      }
+      if (!params.method)
+        params.method = 'GET';
+      let agent;
+      if (this._options.proxy) {
+        const proxyOpts = url.parse(this._options.proxy.server);
```
pavelfeldman, line 203: Let's add a todo for proxy bypass here.

src/server/browserContext.ts
```diff
@@ -179,6 +183,63 @@ export abstract class BrowserContext extends SdkObject {
...
+
+      const request = new network.FetchRequest(this, params.url, params.method, params.headers, params.postData);
+      // TODO: set user agent
+      const response = await fetch(params.url, {
+        method: params.method,
+        headers: params.headers,
+        body: params.postData,
+        agent});
+      const setCookies = response.headers.raw()['set-cookie'];
+      if (setCookies) {
```
pavelfeldman, line 217: We will need a lot of tests :)

src/server/browserContext.ts
```diff
@@ -179,6 +183,63 @@ export abstract class BrowserContext extends SdkObject {
...
+          // Decode cookie value?
+          const cookie: types.SetNetworkCookieParam | null = parseCookie(header);
+          if (!cookie)
+            continue;
+          if (!cookie.domain)
+            cookie.url = params.url;
+          cookies.push(cookie);
+        }
+        if (cookies.length)
+          await this.addCookies(cookies);
```
pavelfeldman, line 229: Just as an idea, I think users might want to see this optional. Same goes for the cokie: headers.

src/protocol/protocol.yml
```diff
@@ -220,6 +220,52 @@ InterceptedResponse:
...
+    method: string
+    postData: binary?
+    headers:
+      type: array
+      items:
+        type: object
+        properties:
+          name: string
+          value: string
+    redirectedFrom: FetchRequest?
```
pavelfeldman, line 237: I don't see this being used and I don't really want to see it to keep things simple... Should we just follow redirects on the driver side w/o client knowing?

src/protocol/protocol.yml
```diff
@@ -220,6 +220,52 @@ InterceptedResponse:
           value: string
 
 
+FetchRequest:
```
pavelfeldman, line 223: Do you anticipate more commands on this one? If not, I would encourage you to have a context.request(payload) method instead, with no protocol objects for either request or response, just request and response payload to keep things simple.

src/server/network.ts
```diff
@@ -434,6 +434,47 @@ export class InterceptedResponse extends SdkObject {
...
 }
 
+export class FetchRequest extends SdkObject {
+  response?: FetchResponse;
+  url: string;
+  method: string;
+  headers: { [name: string]: string };
+  postData?: Buffer;
+
+  constructor(parent: SdkObject, url: string, method: string, headers?: { [name: string]: string }, postData?: Buffer) {
```
pavelfeldman, line 444: Don't make it an SdkObject

src/server/network.ts
```diff
@@ -434,6 +434,47 @@ export class InterceptedResponse extends SdkObject {
...
+    super(parent, 'fetchRequest');
+    this.url = url;
+    this.method = method;
+    this.headers = headers || {};
+    this.postData = postData;
+  }
+}
+
+export class FetchResponse extends SdkObject {
```
pavelfeldman, line 453: Ditto

src/dispatchers/networkDispatchers.ts
```diff
@@ -148,3 +149,55 @@ export class WebSocketDispatcher extends Dispatcher<WebSocket, channels.WebSocke
     webSocket.on(WebSocket.Events.Close, () => this._dispatchEvent('close', {}));
   }
 }
+
+export class FetchRequestDispatcher extends Dispatcher<FetchRequest, channels.FetchRequestInitializer> implements channels.FetchRequestChannel {
```
pavelfeldman, line 153: I would nuke these boilerplate fellows.

pavelfeldman (commented):
Looks good, suggesting to not over-complicate it in the initial patch, just make it a method on context.

src/server/browserContext.ts
yury-s: I'll add these options in a follow-up PR once we settle on the base API.

src/server/browserContext.ts
yury-s: yes

yury-s (commented):
As discussed earlier updated the code to use a new type (FetchResponse) for _fetch results, no information about request/redirects is provided, the response eagerly includes the body.

src/server/fetch.ts
```diff
@@ -0,0 +1,132 @@
...
+    const body = await response.buffer();
+    const setCookies = response.headers.raw()['set-cookie'];
+    if (setCookies) {
+      const cookies: types.SetNetworkCookieParam[] = [];
+      for (const header of setCookies) {
+        // Decode cookie value?
+        const cookie: types.SetNetworkCookieParam | null = parseCookie(header);
+        if (!cookie)
+          continue;
+        if (!cookie.domain)
```
pavelfeldman, line 63: The same should be done to `path`. It can be omitted.

You should also perform basic same origin checks here, otherwise server can poison your cookies.

yury-s: Done.

src/server/fetch.ts
```diff
@@ -0,0 +1,132 @@
...
+    const cookies = await context.cookies(params.url);
+    const valueArray = cookies.map(c => `${c.name}=${c.value}`);
+    const clientCookie = params.headers?.['cookie'];
+    if (clientCookie)
+      valueArray.unshift(clientCookie);
+    const cookieHeader = valueArray.join('; ');
+    if (cookieHeader) {
+      if (!params.headers)
+        params.headers = {};
+      params.headers['cookie'] = cookieHeader;
```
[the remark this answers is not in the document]
yury-s: Filed https://github.com/microsoft/playwright/issues/8381

pavelfeldman (approved):

PR #8353, opened 2021-08-21
feat(launchServer): accept wsPath option

### LINKED ISSUE #6291: [Feature] allow for starting a browser with a hard-coded web-socket path for easier use
I start a browser in a docker container and connect to it from the host that started the container. Unfortunately the path is always randomly generated - https://github.com/microsoft/playwright/blob/master/src/browserServerImpl.ts#L56 and thus requires some work to fetch it and configure it in the host's test code.

I propose a new option to be added to the launch options specifying a path. Something like:
```
  const browserServer = await chromium.launchServer({
	  port: 8080,
	  path: 'playwright'
  });
```
If that option is provided then the given path is used otherwise it behaves as it does at the moment (generate a random path).

Alternatively, one could think about environment variable that can be set and has the same effect (if that is easier to implement and / or matches the design of the rest of the code).

rwoll:
This change allows whoever is calling the launcher to pass
in a secret path to use, instead of having the PW code pick
for you.

This is particularly helpful in use cases where you have
a trusted process launching Docker containers containing the browser
server and you want the host launching the container to know what path
and port to access the running service at without grepping through logs
or resorting to other message passing mechanisms.

[Slack Thread][slack]

Closes #6291

[slack]: https://playwright.slack.com/archives/CSUHZPVLM/p1629509151068800?thread_ts=1629492900.067500&cid=CSUHZPVLM

### DIFF
--- a/docs/src/api/class-browsertype.md
+++ b/docs/src/api/class-browsertype.md
@@ -284,6 +284,18 @@ const { chromium } = require('playwright');  // Or 'webkit' or 'firefox'.
 
 Port to use for the web socket. Defaults to 0 that picks any available port.
 
+### option: BrowserType.launchServer.wsPath
+- `wsPath` <[string]>
+
+Path at which to serve the Browser Server. For security, this defaults to an
+unguessable string.
+
+:::warning
+Any process or web page (including those running in Playwright) with knowledge
+of the `wsPath` can take control of the OS user. For this reason, you should
+use an unguessable token when using this option.
+:::
+
 ## method: BrowserType.name
 - returns: <[string]>
 
--- a/src/browserServerImpl.ts
+++ b/src/browserServerImpl.ts
@@ -52,9 +52,13 @@ export class BrowserServerLauncherImpl implements BrowserServerLauncher {
       env: options.env ? envObjectToArray(options.env) : undefined,
     }, toProtocolLogger(options.logger));
 
+    let path = `/${createGuid()}`;
+    if (options.wsPath)
+      path = options.wsPath.startsWith('/') ? options.wsPath : `/${options.wsPath}`;
+
     // 2. Start the server
     const delegate: PlaywrightServerDelegate = {
-      path: '/' + createGuid(),
+      path,
       allowMultipleClients: true,
       onClose: () => {},
       onConnect: this._onConnect.bind(this, playwright, browser),
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -100,6 +100,7 @@ export type LaunchServerOptions = {
   downloadsPath?: string,
   chromiumSandbox?: boolean,
   port?: number,
+  wsPath?: string,
   logger?: Logger,
 } & FirefoxUserPrefs;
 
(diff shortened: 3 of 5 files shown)


docs/src/api/class-browsertype.md
```diff
@@ -284,6 +284,13 @@ const { chromium } = require('playwright');  // Or 'webkit' or 'firefox'.
 
 Port to use for the web socket. Defaults to 0 that picks any available port.
 
+### option: BrowserType.launchServer.path
+- `path` <[string]>
```
rwoll, line 288: Do we want to call this parameter something else to make it more explicit that it should be secret/unguessable (aside from the nodtes/comment)? Maybe `secretPath` or `unguessablePath`?

src/browserServerImpl.ts
```diff
@@ -52,9 +52,13 @@ export class BrowserServerLauncherImpl implements BrowserServerLauncher {
       env: options.env ? envObjectToArray(options.env) : undefined,
     }, toProtocolLogger(options.logger));
 
+    let path = `/${createGuid()}`;
+    if (options.path)
+      path = options.path.startsWith('/') ? options.path : `/${options.path}`;
```
rwoll, line 57: The leading slash is important. I opted to accept liberal input and fix it up. Let me know if we'd rather just throw an error if it doesn't start with `/`.

Likewise, do we want to do more validation or restriction on the path? Currently, you could put `/foo   bar` and we won't throw a validation error.

rwoll (commented):
@pavelfeldman Left inline comments regarding naming and validation, but other than that, this would work well for my use case.

Thanks!

docs/src/api/class-browsertype.md
pavelfeldman: I would call it `wsPath` and document that it should be unguessable.

src/browserServerImpl.ts
pavelfeldman: You code looks fine as is.

docs/src/api/class-browsertype.md
```diff
@@ -284,6 +284,13 @@ const { chromium } = require('playwright');  // Or 'webkit' or 'firefox'.
...
+
+Path at which to serve the Browser Server. For security, this defaults to an
+unguessable string. Any process or web page (including those running in
+Playwright) with knowledge of this path can take over control of the browser.
```
pavelfeldman, line 292: Rather control over the entire OS user.

pavelfeldman (approved):
Approved with a couple of nit comments.

rwoll:
> Approved with a couple of nit comments.

Thanks @pavelfeldman! All should be addressed now.

Did something change with CI? Looks like no checks are automatically running on my MR any more.

mxschmitt:
GitHub introduced a few months ago a feature that repository maintainers need to approve external contributors workflow runs to prevent spam. This was set to "all external contributors", I changed it to "first time contributors" so it should not happen for you in the future and pressed approve.

pavelfeldman:
I set it to "all external contributors" yesterday to see the impact. It did not last long as I can see :-)

rwoll:
> I set it to "all external contributors" yesterday to see the impact. It did not last long as I can see :-)

@pavelfeldman Phew! You almost foiled my plan to mine Bitcoin via PR! 💸💸😉

PR #8385, opened 2021-08-24
feat(har): Remotely accessible HAR file

### LINKED ISSUE #8355: [Feature] Access HAR Catpture from Remote Playwright Client
Under Playwright Server mode, `context.recoardHar.path` only saves the path on the server, but there is no way to access the HAR from the client.

Would you be open to a PR that saves the har to recordHar.path on the client?

Alternatively, it would be nice to have something like…

```js
context.har: (args: { path?: string }): Promise<HAR> // or Buffer if we don’t want it structured
```

As of now, I’ve worked around this by running a simple
HTTP file server exposing the directory the server is writing to so that the client
can request.

Example Server/Client:

`server.js`:

```js
const { chromium } = require(“playwright”);

(async () => {
  const server = await chromium.launchServer({
    port: 9090,
  });
  console.log(`Playwright Server Listening; PW_SERVER=${server.wsEndpoint()}`);
})()
```

`client.js`:

```js
// Filename: client.js
const { chromium } = require(“playwright”);

(async () => {
  const browser = await chromium.connect(http://process.env.PW_SERVER);
    recordHar: { path: “example.har” }
  })
  await page.goto(“https://playwright.dev/”);
  await context.close();
  await browser.close();
})()
```

rwoll:
This change ensure's the HAR file is saved at `recordHar.path` on the
client instead of the server.

NB: The goal was to make this change transparent to the user and NOT
introduce any new APIs. Namely, I want to leave the API open for
potential `context.har.start()` and `context.har.stop()`.

This does BREAK servers that expect the HAR to be at the `recordHar.path`
on the server, but I think that's OK since there haven't been reports
of missing HAR on client making me think not many users are getting
HAR with client and server on different hosts anyways.

Closes #8355

### DIFF
--- a/src/protocol/protocol.yml
+++ b/src/protocol/protocol.yml
@@ -707,6 +707,10 @@ BrowserContext:
       returns:
         artifact: Artifact
 
+    harExport:
+      returns:
+        artifact: Artifact
+
   events:
 
     bindingCall:
--- a/src/server/browserContext.ts
+++ b/src/server/browserContext.ts
@@ -61,7 +61,7 @@ export abstract class BrowserContext extends SdkObject {
   readonly _browserContextId: string | undefined;
   private _selectors?: Selectors;
   private _origins = new Set<string>();
-  private _harRecorder: HarRecorder | undefined;
+  readonly _harRecorder: HarRecorder | undefined;
   readonly tracing: Tracing;
 
   constructor(browser: Browser, options: types.BrowserContextOptions, browserContextId: string | undefined) {
@@ -74,7 +74,8 @@ export abstract class BrowserContext extends SdkObject {
     this._closePromise = new Promise(fulfill => this._closePromiseFulfill = fulfill);
 
     if (this._options.recordHar)
-      this._harRecorder = new HarRecorder(this, this._options.recordHar);
+      this._harRecorder = new HarRecorder(this, {...this._options.recordHar, path: path.join(this._browser.options.artifactsDir, `${createGuid()}.har`)});
+
     this.tracing = new Tracing(this);
   }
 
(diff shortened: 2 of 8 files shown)


rwoll:
@pavelfeldman / @dgozman: This is ready for review. Thanks!

src/server/supplements/har/harTracer.ts
```diff
@@ -263,6 +267,8 @@ export class HarTracer {
   }
 
   async flush() {
+    if (this._isFlushed) return;
```
pavelfeldman, line 270: nit: we put return on the next line

src/protocol/protocol.yml
```diff
@@ -684,6 +684,10 @@ BrowserContext:
       returns:
         artifact: Artifact
 
+    harExport:
```
pavelfeldman, line 687: nit: this is ok, but there was nothing wrong with returning optional artifact from close.

src/server/browserContext.ts
```diff
@@ -74,7 +74,8 @@ export abstract class BrowserContext extends SdkObject {
     this._closePromise = new Promise(fulfill => this._closePromiseFulfill = fulfill);
 
     if (this._options.recordHar)
-      this._harTracer = new HarTracer(this, this._options.recordHar);
+      this._harTracer = new HarTracer(this, {...this._options.recordHar, path: path.join(this._browser.options.artifactsDir, `${createGuid}.har`)});
```
pavelfeldman, line 77: I don't think this does what you think this does, which suggests we might be missing a test with 2 hars.

pavelfeldman (commented):
Looks good, has a small bug.

src/server/browserContext.ts
rwoll: Oops! Missed the actual invocation!

rwoll:
Thanks @pavelfeldman! It's all fixed up and I added a test for the duplicate/wrong filename.

The test failed before the change (8/10 times with `--repeat-each`), but now passes 🎉! (I couldn't get a more deterministic failure since once the context is closed, we don't call saveAs again.)

pavelfeldman (approved):

PR #8386, opened 2021-08-24
feat(route): rename method, add response option

yury-s:
Two changes to the response interception API:
* Renamed intercepting method to `Route._continueToRequest`
* Take intercepted response as an option in `Route.fulfill()`

#1774

### DIFF
--- a/tests/page/page-request-intercept.spec.ts
+++ b/tests/page/page-request-intercept.spec.ts
@@ -63,7 +63,6 @@ it('should override with defaults when intercepted response not provided', async
     res.end('my content');
   });
   await page.route('**/*', async route => {
-    route.continue();
     // @ts-expect-error
     await route._continueToResponse({});
     await route.fulfill({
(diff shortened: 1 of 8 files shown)


tests/page/page-request-intercept.spec.ts
```diff
@@ -55,6 +56,50 @@ it('should fulfill response with empty body', async ({page, server, browserName,
...
+
+it('should fulfill with any response', async ({page, server, browserName, browserMajorVersion}) => {
+  it.skip(browserName === 'chromium' && browserMajorVersion <= 91, 'Fails in Electron that uses old Chromium');
+
+  server.setRoute('/sample', (req, res) => {
+    res.setHeader('foo', 'bar');
+    res.end('Woo-hoo');
+  });
+  const page2 = await page.context().newPage();
+  const sampleResponse = await page2.goto(`${server.PREFIX}/sample`);
```
yury-s, line 87: We cannot use `page.goto()` here to create sampleResponse as in that case `sampleResponse.body()` would throw in because `page` already started navigating to another URL. Perhaps we should bail out if we cannot retrieve response body of the response which is passed or only accept intercepted response.

pavelfeldman (approved):

PR #8388, opened 2021-08-24
browser(firefox): disable enterprise policies

JoelEinbinder:
References #8141
I am not sure if this is the right way to ﻿solve this issue, but I just disabled the enterprise policy code.

### DIFF
--- a/browser_patches/firefox-beta/BUILD_NUMBER
+++ b/browser_patches/firefox-beta/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1280
-Changed: dgozman@gmail.com Wed Aug 18 14:34:33 PDT 2021
+1281
+Changed: joel.einbinder@gmail.com Tue 24 Aug 2021 12:02:01 PM PDT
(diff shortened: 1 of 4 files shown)


mxschmitt:
Don't forget firefox-beta.

yury-s (approved):

aslushnikov (approved):

jackyzy823:
@JoelEinbinder  since the issue #8141 still exists, can we revert this patch? since we all need group policies . see #10284

In my case i need to add self signed root certificate to test self-hosed self-signed Firefox sync service ( `ignoreHTTPSErrors` doest not work under this situation , because the request is not from HTML Pages but from Firefox  internally )

DashwoodIce9:
+1 to revert request. Folks over at #8141 have found a way forward with [`--pref='datareporting.policy.firstRunURL=""'`](https://github.com/microsoft/playwright/issues/8141#issuecomment-2008711137) and disabling enterprise policies disables a lot of customization features such loading custom certificates (#18115). Using custom certificates is necessary for my use case, so currently I have to write a script that undoes the changes made by this commit.

mxschmitt:
> +1 to revert request. Folks over at #8141 have found a way forward with [`--pref='datareporting.policy.firstRunURL=""'`](https://github.com/microsoft/playwright/issues/8141#issuecomment-2008711137) and disabling enterprise policies disables a lot of customization features such loading custom certificates (#18115). Using custom certificates is necessary for my use case, so currently I have to write a script that undoes the changes made by this commit.

Client Certificates are available natively in Playwright: https://playwright.dev/docs/api/class-testoptions#test-options-client-certificates since a few releases. Does this help?

DashwoodIce9:
Sorry, no client certificates are not what I was looking for. My use case is to add certificates to browser's trust store for server side certificate verification. This can be done via use of [Enterprise Policies feature on Firefox](https://mozilla.github.io/policy-templates/#certificates--install). Commonly used for self signed server side certificates or to test internal (to a particular organization) website whose certificates were issued by a private certificate authority.

PR #8405, opened 2021-08-24
fix(test-runner): allow matching against source mapped files

JoelEinbinder:
References https://github.com/microsoft/playwright/issues/16975

This is half the battle, matching the filename against the source map. But we also need to convert the line number from the source file to a line
number in the compiled file.

Maybe the source map code should have a try-catch? I'm not sure what broken source maps on our tests would mean.

### DIFF
--- a/src/test/runner.ts
+++ b/src/test/runner.ts
@@ -21,7 +21,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { promisify } from 'util';
 import { Dispatcher, TestGroup } from './dispatcher';
-import { createMatcher, FilePatternFilter, monotonicTime, raceAgainstDeadline } from './util';
+import { createMatcher, FilePatternFilter, Matcher, monotonicTime, raceAgainstDeadline } from './util';
 import { TestCase, Suite } from './test';
 import { Loader } from './loader';
 import { Reporter } from '../../types/testReporter';
@@ -36,6 +36,7 @@ import { ProjectImpl } from './project';
 import { Minimatch } from 'minimatch';
 import { Config, FullConfig } from './types';
 import { WebServer } from './webServer';
+import { mapSourcePosition, retrieveSourceMap } from 'source-map-support';
 
 const removeFolderAsync = promisify(rimraf);
 const readDirAsync = promisify(fs.readdir);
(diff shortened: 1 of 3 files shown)


tests/playwright-test/test-ignore.spec.ts
```diff
@@ -390,3 +390,28 @@ test('should only match files with JS/TS file extensions', async ({ runInlineTes
...
+    'src/foo.test.ts': `//@no-header
+import test from '@playwright/test';
+
+test('passes', async() => {
+});
+`
+  }, {
+    args: ['-c', 'out', 'foo.spec.ts']
+  });
+  expect(result.exitCode).toBe(0);
```
mxschmitt, line 415: The important part for the VSCode extension is that even if you have JS as input. it should use as output file names the TS file names.

JoelEinbinder: Where's the code that does this? Error messages will appear source mapped, but the suite titles will show the requireFile.

mxschmitt: Its not about error messages, we have two use cases which we need to solve:
a) find all tests via the test explorer. Its basically `npx playwright test out/` there we currently get in the JSON report the JS files, because in the out/ dir are only JS files. But to reference them correctly in vsc I need the TS references with line numbers.
b) the user opens a file, e.g. foo.test.ts With the current setup it won't work at all, because babel-typescript can't compile it, so its broken then.

tests/playwright-test/test-ignore.spec.ts
```diff
@@ -390,3 +390,28 @@ test('should only match files with JS/TS file extensions', async ({ runInlineTes
...
+    'out/foo.spec.js.map': `{"version":3,"file":"foo.spec.js","sourceRoot":"","sources":["../src/foo.spec.ts"],"names":[],"mappings":";;AAAA,YAAY;AACZ,2CAAoC;AAEpC,cAAI,CAAC,QAAQ,EAAE,KAAK,IAAG,EAAE;AACzB,CAAC,CAAC,CAAC"}`,
+
+    'src/foo.test.ts': `//@no-header
+import test from '@playwright/test';
+
+test('passes', async() => {
+});
+`
+  }, {
+    args: ['-c', 'out', 'foo.spec.ts']
```
dgozman, line 413: Can we do `src/foo.spec.ts`? Does it work?

src/test/runner.ts
```diff
@@ -321,6 +322,22 @@ export class Runner {
   }
 }
 
+
+export function createSourceMapMatcher(patterns: string | RegExp | (string | RegExp)[]): Matcher {
+  const matcher = createMatcher(patterns);
+  return (filePath: string) => {
+    const sourceMap = retrieveSourceMap(filePath);
```
dgozman, line 329: I see that we provide custom `retrieveSourceMap` in `transform.ts`. Do they play nicely together?

dgozman (commented):
Is this for "manually compile your tests" setup? If so, do we show source or output paths in the report?  Does this make it consistent?

mxschmitt:
Closing since its not needed for now.

Issue #8216, opened 2021-08-16
[Feature] Support emulating User Agent data like navigator.platform


mxschmitt:
Reference: https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#type-UserAgentMetadata

![image](https://user-images.githubusercontent.com/17984549/129541193-9927ee6c-7d7c-49ad-9258-e750ad6a2098.png)

qkdreyer:
- https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setUserAgentOverride
- https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setNavigatorOverrides

PR #8406, opened 2021-08-24
feat: add support for navigator.platform emulation

qkdreyer:
Relates #8216

### DIFF
--- a/tests/defaultbrowsercontext-1.spec.ts
+++ b/tests/defaultbrowsercontext-1.spec.ts
@@ -140,6 +140,11 @@ it('should support userAgent option', async ({ server, launchPersistent }) => {
   expect(request.headers['user-agent']).toBe('foobar');
 });
 
+it('should support platform option', async ({ launchPersistent }) => {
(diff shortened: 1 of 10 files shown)


mxschmitt (approved):

tests/defaultbrowsercontext-1.spec.ts
```diff
@@ -140,6 +140,11 @@ it('should support userAgent option', async ({server, launchPersistent}) => {
   expect(request.headers['user-agent']).toBe('foobar');
 });
 
+it('should support platform option', async ({launchPersistent, browserName}) => {
```
qkdreyer, line 143: If we want to push cleaner tests
```suggestion
it('should support platform option', async ({launchPersistent}) => {
```

JoelEinbinder:
@pavelfeldman what do you think about this api name? I think thats all we are blocked on.

gohoski:
Also still waiting, three years later… this would be a very useful feature, why did this PR get closed anyway? Why do we need to 'wait for more positive signals from the community'???

Issue #7029, opened 2021-06-10
globalSetup doesn't run before all tests


jsjoeio:
## Description

`globalSetup` should run before any tests run. However, it seems like `@playwright/test` is running the test (or checking the code) before running the test. This means that even though I define `globalSetup` in my `config.ts` file then set `process.env.STORAGE` and try to use it in a test, playwright thinks it's undefined and exits the runner. 

## Reproduction 

Here is the [repo](https://github.com/jsjoeio/playwright-test-repro). 

1. `git clone https://github.com/jsjoeio/playwright-test-repro.git`
2. `yarn`
3. `yarn test`

### Expected

Global setup should run _then_ tests should run 

### Actual

`@playwright/test` runs tests first:
```js
Error while reading /Users/jp/Dev/playwright-test-repro/hello.test.ts:
Unexpected token u in JSON at position 0
    at JSON.parse (<anonymous>)
    at /Users/jp/Dev/playwright-test-repro/hello.test.ts:8:29
    at describe (/Users/jp/Dev/playwright-test-repro/node_modules/folio/out/spec.js:127:9)
    at Object.<anonymous> (/Users/jp/Dev/playwright-test-repro/hello.test.ts:3:6)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Module._compile (/Users/jp/Dev/playwright-test-repro/node_modules/pirates/lib/index.js:99:24)
    at Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Object.newLoader [as .ts] (/Users/jp/Dev/playwright-test-repro/node_modules/pirates/lib/index.js:104:7)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Loader.loadTestFile (/Users/jp/Dev/playwright-test-repro/node_modules/folio/out/loader.js:60:13)
    at runTests (/Users/jp/Dev/playwright-test-repro/node_modules/folio/out/cli.js:130:16)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
error Command failed with exit code 1.
```

mxschmitt:
Closing for now. Small summary: Yes in your case it runs multiple times. its because in your global setup, you have 

`import { PASSWORD } from "./constants"`

which will call the constants.ts

[...]

PR #8414, opened 2021-08-24
fix(test-runner): do not start web-server when tests are listed

### LINKED ISSUE #35: Specifying the `webServer` in the config seems to partially initiate it
I have the `webServer` configuration specified in my playwright.config.ts file which is setup to run my local webpack dev server.  I'm noticing that when my webpack dev server has `open: true` set, opening the project or expanding a group of Playwright tests will open the URL in my browser.  This leads me to believe that the extension is partially initiating the web server when listing tests.  This also seems to negatively impact performance of the extension due to it doing something with the server.

mxschmitt:
Fixes https://github.com/microsoft/playwright-vscode/issues/35

### DIFF
--- a/src/test/runner.ts
+++ b/src/test/runner.ts
@@ -167,7 +167,7 @@ export class Runner {
       testFiles.forEach(file => allTestFiles.add(file));
     }
 
-    const webServer = (!list && config.webServer) && await WebServer.create(config.webServer);
+    const webServer = (!list && config.webServer) ? await WebServer.create(config.webServer) : undefined;
     let globalSetupResult: any;
     if (config.globalSetup)
       globalSetupResult = await (await this._loader.loadGlobalHook(config.globalSetup, 'globalSetup'))(this._loader.fullConfig());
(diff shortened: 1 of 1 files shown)


src/test/runner.ts
```diff
@@ -167,7 +167,7 @@ export class Runner {
       testFiles.forEach(file => allTestFiles.add(file));
     }
 
-    const webServer = config.webServer && await WebServer.create(config.webServer);
+    const webServer = (!list && config.webServer) && await WebServer.create(config.webServer);
     let globalSetupResult: any;
     if (config.globalSetup)
```
JoelEinbinder, line 172: We are still running globalSetup when we are in this mode. Is there a reason globalSetup runs before the tests are listed? I would have expected it to run right before we partition tests into work.

@dgozman

dgozman: https://github.com/microsoft/playwright/issues/7029
https://github.com/microsoft/folio/pull/222

JoelEinbinder (approved):
Let's land this for now, but think about some kind of watch mode to avoid re-running the globalSetup a million times.

PR #8429, opened 2021-08-25
cherry-pick(release-1.14): avoid the "accept incoming network connections" dialog

### LINKED ISSUE #7937: [Feature]: CodeSign chromium by default in macOS big sur
### Feature request

In macOS big sur , every time I get a popup to allow the Chromium to access network when I execute test.

So I googled a bit that it said it's because of the app is not signed , so I'm doing something like

```
    cd $HOME/Library/Caches/ms-playwright/
    # codesign for chromium in Macos
    for d in chromium-*; do
        echo Sign $HOME/Library/Caches/ms-playwright/${d}/chrome-mac/Chromium.app
        sudo codesign --force --deep --sign - ./${d}/chrome-mac/Chromium.app
    done
```

however is it possible to sign it by default when install add new Chromium inside of $HOME/Library/Caches/ms-playwright?

aslushnikov:
PR #8162 SHA 819287cc0eeb77b449e208364fe80aac47ec690f

Fixes #7937

### DIFF
--- a/src/server/chromium/chromium.ts
+++ b/src/server/chromium/chromium.ts
@@ -197,7 +197,7 @@ const DEFAULT_ARGS = [
   '--disable-default-apps',
   '--disable-dev-shm-usage',
   '--disable-extensions',
-  '--disable-features=ImprovedCookieControls,LazyFrameLoading,GlobalMediaControls,DestroyProfileOnBrowserClose',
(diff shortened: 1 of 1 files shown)


mxschmitt:
This is afaik no regression, I would not release 1.14.2 only for that.

mxschmitt (approved):

PR #8438, opened 2021-08-25
devops: drop build infrastructure for WebKit @ MacOS 10.14

aslushnikov:
We did not support it for a while now and had 0 complaints.


mxschmitt (approved):
don't forget the playwright-internal hooks.

PR #8483, opened 2021-08-26
chore: hide Route.fulfill._response from API

mxschmitt:


yury-s (commented):
Let's remove underscore prefix from the option name since the method is going to be hidden anyway

yury-s (approved):

PR #8485, opened 2021-08-26
docs: introduce overloads, generate JSDoc for overloads

dgozman:
- Each overload, e.g. for `page.evaluate`, shows a nice autocomplete doc, not only the first one.
- We can have multiple overloads directly on the docs page, e.g. `test.skip(title, fn)` and `test.skip(condition, description)`. These overloads are internally named `Test.skip#1` and are all aliased to `test.skip`.

### DIFF
--- a/docs/src/test-api/class-test.md
+++ b/docs/src/test-api/class-test.md
@@ -599,12 +599,9 @@ Timeout in milliseconds.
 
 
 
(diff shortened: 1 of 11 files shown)


JoelEinbinder (approved):
lgtm % pavel's comment about the params.

PR #8520, opened 2021-08-27
docs: expand "Strictness" examples

mrienstra:
Add a third example showing a case where "strictness" doesn't apply.

This may be unnecessary, but I thought it might be useful to some.

Also tempted to change this phrase:
>operations on locators that imply some target DOM element

To:
>operations on locators that expect a single target DOM element

Or:
>operations on locators that expect a single DOM element

Or:
>operations on locators that expect a single match

PS: also changed two instances of "you" to "we" for consistency.

### DIFF
--- a/docs/src/api/class-locator.md
+++ b/docs/src/api/class-locator.md
@@ -153,7 +153,7 @@ await page.Locator("button").ClickAsync();
 await page.Locator("button").First.ClickAsync();
 
 // Works because Count knows what to do with multiple matches:
-await page.Locator("button").Count();
+await page.Locator("button").CountAsync();
 ```
 
 ## async method: Locator.allInnerTexts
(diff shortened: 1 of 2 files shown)


docs/src/api/class-locator.md
```diff
@@ -123,21 +129,31 @@ page.locator('button').click()
...
 ```csharp
 // Throws if there are several buttons in DOM:
 await page.Locator("button").ClickAsync();
-// Works because you explicitly tell locator to pick the first element:
+
+// Works because we explicitly tell locator to pick the first element:
 await page.Locator("button").First.ClickAsync();
+
+// Works because Count knows what to do with multiple matches:
+await page.Locator("button").Count();
```
dgozman, line 156: It is `CountAsync` in .NET.

dgozman (commented):
Great PR, thank you! Just a minor issue with .NET 😄

mrienstra:
Thanks @dgozman & @pavelfeldman!

PR #8526, opened 2021-08-28
fix(cookies): read response headers off extra info event, if any

pavelfeldman:


src/server/chromium/crNetworkManager.ts
```diff
@@ -337,10 +342,16 @@ export class CRNetworkManager {
...
-    this._page._frameManager.requestFinished(request.request);
+    this._page._frameManager.requestFinished(request.request, response);
+  }
+
+  _onResponseReceivedExtraInfo(event: Protocol.Network.responseReceivedExtraInfoPayload) {
+    this._responseExtraInfoTracker.responseReceivedExtraInfo(event);
   }
 
   _onResponseReceived(event: Protocol.Network.responseReceivedPayload) {
+    this._responseExtraInfoTracker.responseReceived(event);
```
dgozman, line 353: I am a bit confused that we notify ResponseExtraInfoTracker here, and also in _createResponse called 6 lines below.

src/server/chromium/crNetworkManager.ts
```diff
@@ -542,3 +563,133 @@ const errorReasons: { [reason: string]: Protocol.Network.ErrorReason } = {
...
+    const info = this._requests.get(event.requestId);
+    if (info) {
+      // This is redirect.
+      if (!event.redirectResponse?.connectionId) {
+        // Bail out early for responses without connectionId.
+        // They won't be getting extra info, so there is nothing to track.
+        this._stopTracking(event.requestId);
+        return;
+      }
+      info.responseReceived = true;
```
dgozman, line 602: It looks like we set `responseReceived` to true only for requests that had at least one redirect. Perhaps rename it to 'hasRedirect'?

src/server/chromium/crNetworkManager.ts
```diff
@@ -542,3 +563,133 @@ const errorReasons: { [reason: string]: Protocol.Network.ErrorReason } = {
...
+      });
+    }
+  }
+
+  responseReceived(event: Protocol.Network.responseReceivedPayload) {
+    const info = this._requests.get(event.requestId);
+    if (!info)
+      return;
+
+    // Bail out early for responses without connectionId.
```
dgozman, line 618: Why don't we do this in processResponse?

src/server/chromium/crNetworkManager.ts
```diff
@@ -542,3 +563,133 @@ const errorReasons: { [reason: string]: Protocol.Network.ErrorReason } = {
...
+  private _requests = new Map<string, RequestInfo>();
+
+  constructor() {
+  }
+
+  requestWillBeSent(event: Protocol.Network.requestWillBeSentPayload) {
+    const info = this._requests.get(event.requestId);
+    if (info) {
+      // This is redirect.
+      if (!event.redirectResponse?.connectionId) {
```
dgozman, line 596: It seems like both connectionId checks could be done in processResponse method that is called for redirects and final response?

src/server/chromium/crNetworkManager.ts
```diff
@@ -542,3 +563,133 @@ const errorReasons: { [reason: string]: Protocol.Network.ErrorReason } = {
...
+    if (response && extraInfo)
+      response.extraHeadersReceived(headersObjectToArray(extraInfo.headers));
+  }
+
+  private _checkFinished(info: RequestInfo) {
+    if (!info.loadingFinished && !info.loadingFailed)
+      return;
+
+    // Loading finished, check that we have all ExtraInfo events in place.
+    if (!info.responseReceived) {
```
dgozman, line 677: According to my other comment where "responseReceived" actually means "has redirect", it seems we are not waiting for the extra info on responses without redirects.

src/server/chromium/crNetworkManager.ts
pavelfeldman: They are sending two distinct signals to tracker. As you can see, tracker repeats the CDP api. This is to make sure that it is easy to read it - CDP has well-understood contract. So our tracker is an engine that consumes two inputs: CDP on one side and created Playwright Responses on the other side.

src/server/chromium/crNetworkManager.ts
pavelfeldman: I should have assigned it in responseReceived as well.

src/server/chromium/crNetworkManager.ts
pavelfeldman: To follow a contract cleanly. If we've ever seen a CDP response w/o connectionId, we aren't interested in the services of this tracker.

src/server/chromium/crNetworkManager.ts
pavelfeldman: We could do that. But a) we would need to collect all the `responseReceived` events for that to be able to check for the connectionId and b) I think there is certain simplicity in following CDP rather than making sense of things as you are reading this code.

src/server/chromium/crNetworkManager.ts
pavelfeldman: Yeah, because "responseReceived" means "responseReceived", not "has redirect" as you suggested.

PR #8533, opened 2021-08-28
chore: introduce manual promise helper

pavelfeldman:


src/utils/utils.ts
```diff
@@ -392,3 +392,28 @@ export function wrapInASCIIBox(text: string, padding = 0): string {
     '╚' + '═'.repeat(maxLength + padding * 2) + '╝',
   ].join('\n');
 }
+
+export class ManualPromise<T> implements PromiseLike<T> {
```
dgozman, line 396: Why not extend Promise? Then you won't need PromiseLike throughout the code.

pavelfeldman: This thought did not cross my mind, but I feel like I should not be extending Promise.

pavelfeldman: > Then you won't need PromiseLike throughout the code.

Are you referring to a single instance where I needed to do it? :)

PR #8544, opened 2021-08-30
browser(webkit): fix compilation in Ubuntu 18.04

dpino:
This PR cherry-picks https://trac.webkit.org/changeset/281752/webkit, which introduces ICU 61 as a dependency in the JHBuild minimal dependencies module set.

It is necessary to make the CI Docker Bionic image to build successfully to upgrade the compiler to GCC-8.

### DIFF
--- a/browser_patches/webkit/patches/bootstrap.diff
+++ b/browser_patches/webkit/patches/bootstrap.diff
@@ -22550,6 +22550,73 @@ index 02fa9bff0ab4bd094fd9498b042e06f35c25e7fc..f33f29a9a26d1219aa5780b186079235
      ruby
  
      # These are dependencies necessary for running tests.
+diff --git a/Tools/jhbuild/jhbuild-minimal.modules b/Tools/jhbuild/jhbuild-minimal.modules
+index dc7ee36b969f6490d96df4d5310e6b81124a4a94..3ba44482a67059f7a6bd64241ff6f6f27fe8f2d5 100644
+--- a/Tools/jhbuild/jhbuild-minimal.modules
++++ b/Tools/jhbuild/jhbuild-minimal.modules
+@@ -6,6 +6,7 @@
+   <metamodule id="webkitgtk-minimal-dependencies">
+     <dependencies>
+       <dep package="wpebackend-fdo"/>
++      <dep package="icu"/>
+       <dep package="libsoup"/>
+       <dep package="manette"/>
+       <dep package="libvpx"/>
+@@ -15,6 +16,7 @@
+   <metamodule id="webkitwpe-minimal-dependencies">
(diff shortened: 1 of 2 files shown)


dpino:
Depends on https://github.com/microsoft/playwright/pull/8545

browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -22550,6 +22550,73 @@ index 02fa9bff0ab4bd094fd9498b042e06f35c25e7fc..f33f29a9a26d1219aa5780b186079235
...
++===================================================================
++--- icu-52~m1.orig/source/config/mh-linux	2013-09-14 18:53:23.284040467 -0400
+++++ icu-52~m1/source/config/mh-linux	2013-09-14 18:53:23.284040467 -0400
++@@ -21,7 +21,9 @@
++ LD_RPATH_PRE = -Wl,-rpath,
++ 
++ ## These are the library specific LDFLAGS
++-LDFLAGSICUDT=-nodefaultlibs -nostdlib
+++#LDFLAGSICUDT=-nodefaultlibs -nostdlib
+++# Debian change: linking icudata as data only causes too many problems.
```
yury-s, line 22615: I don't see this [upstream](https://trac.webkit.org/changeset/281752/webkit), why is this necessary?

yury-s:
> It is necessary to make the CI Docker Bionic image to build successfully to upgrade the compiler to GCC-8.

What's the story with upgrading the compiler, can you elaborate?

browser_patches/webkit/patches/bootstrap.diff
dpino: This patch was part of the upstream ChangeLog https://trac.webkit.org/changeset/281752/webkit (Files: 1 added, 2 edited). I think the ChangeLog diff only shows the edited files. It's much clearer to see this change in the patch's diff in WebKit's Bugzilla https://bugs.webkit.org/attachment.cgi?id=436693&action=prettypatch

This patch for the ICU libraries is already being used in the JHBuild general dependencies https://github.com/WebKit/WebKit/blob/main/Tools/gtk/jhbuild.modules#L449, so we imported to JHBuild minimal as it is. I don't know the background story of this patch but not to add it would mean to me a change over what's deployed when using the JHBuild general dependencies.

yury-s: > This patch was part of the upstream ChangeLog https://trac.webkit.org/changeset/281752/webkit (Files: 1 added, 2 edited). I think the ChangeLog diff only shows the edited files.

Oops, didn't see that. Thanks!

yury-s (approved):

PR #8545, opened 2021-08-30
devops(docker): upgrade Bionic to GCC-8

dpino:
This is necessary to build WebKit successfully in Ubuntu 18.04. GCC-8 is featured in Ubuntu 18.04.

### DIFF
--- a/utils/docker/Dockerfile.bionic
+++ b/utils/docker/Dockerfile.bionic
@@ -21,6 +21,11 @@ RUN apt-get update && apt-get install -y python3.8 python3-pip && \
     update-alternatives --install /usr/bin/python python /usr/bin/python3 1 && \
     update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 1
 
+# Install GCC-8.
+RUN apt-get update && apt-get install -y gcc-8 g++-8 && \
+    update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 700 --slave /usr/bin/g++ g++ /usr/bin/g++-7 && \
+    update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 800 --slave /usr/bin/g++ g++ /usr/bin/g++-8
+
 # === BAKE BROWSERS INTO IMAGE ===
 
 ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright


mxschmitt (commented):
The Docker files are used for customers to use Playwright and not for us when building browsers so I'd not add compiler suites there.
Is your intention to add it to our build infrastructure to fix the current failing builds?

dpino:
Yes, my main motivation was to switch the CI containers to GCC-8 so WebKit can actually build. I thought the CI containers were also generated out of these Dockerfiles.

Currently, WebKit doesn't build if using GCC-7. This is the first error I got:

```
/home/dpino/playwright/browser_patches/webkit/checkout/Source/WebCore/Modules/encryptedmedia/InitDataRegistry.cpp:151:12:   required from here
/home/dpino/playwright/browser_patches/webkit/checkout/WebKitBuild/GTK/Release/WTF/Headers/wtf/Vector.h:170:13: error: use of deleted function ~@~Xstd::unique_ptr<_Tp, _Dp>::unique_ptr(const std::uniquee
_ptr<_Tp, _Dp>&) [with _Tp = WebCore::ISOProtectionSystemSpecificHeaderBox; _Dp = std::default_delete<WebCore::ISOProtectionSystemSpecificHeaderBox>]~@~Y
             new (NotNull, dst) U(*src);
             ^~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/7/memory:80:0,
                 from /home/dpino/playwright/browser_patches/webkit/checkout/WebKitBuild/GTK/Release/WTF/Headers/wtf/StdLibExtras.h:30,
                 from /home/dpino/playwright/browser_patches/webkit/checkout/WebKitBuild/GTK/Release/WTF/Headers/wtf/FastMalloc.h:26,
                 from /home/dpino/playwright/browser_patches/webkit/checkout/Source/WebCore/config.h:56,
                 from /home/dpino/playwright/browser_patches/webkit/checkout/Source/WebCore/Modules/encryptedmedia/InitDataRegistry.cpp:26:
/usr/include/c++/7/bits/unique_ptr.h:383:7: note: declared here
       unique_ptr(const unique_ptr&) = delete;
       ^~~~~~~~~~
```

While we were maintaining the upstream Ubuntu 18.04 buildbot we actively fixed errors due to GCC-7. Usually these errors came from using compilers features not available on earlier version, or sometimes they came from using std libraries functions labeled as experimental on earlier version of GCC but were standard on later versions. At some point we decided to upgrade the GCC version to GCC-8 in the upstream Ubuntu 18.04 buildbot. We thought this was a reasonable change because actually GCC-8 comes with Ubuntu 18.04 (not need to add an external ppa). 

If this change doesn't make it to the codebase, users creating a Bionic image out this Dockerfile will find WebKit actually doesn't build. I think we should at least let users know about that, perhaps through a note on a README.

aslushnikov:
> If this change doesn't make it to the codebase, users creating a Bionic image out this Dockerfile will find WebKit actually doesn't build.

Yeah, folks don't build WebKit using this image, so there's no reason to add these deps.

I did install them on our buildbot though.

yury-s:
Closing per comment above and as r1538 built fine on Ubuntu 18.04

PR #8562, opened 2021-08-30
feat(fetch): support baseURL, add tests for invalid args

yury-s:
#5999

### DIFF
--- a/src/server/fetch.ts
+++ b/src/server/fetch.ts
@@ -47,8 +47,7 @@ export async function playwrightFetch(context: BrowserContext, params: types.Fet
       agent = new HttpsProxyAgent(proxyOpts);
     }
 
-    // TODO(https://github.com/microsoft/playwright/issues/8381): set user agent
-    const fetchResponse = await sendRequest(context, new URL(params.url), {
+    const fetchResponse = await sendRequest(context, new URL(params.url, context._options.baseURL), {
       method,
       headers,
       agent,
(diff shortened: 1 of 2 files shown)


dgozman (approved):

tests/browsercontext-fetch.spec.ts
```diff
@@ -349,3 +349,44 @@ it('should propagate extra http headers with redirects', async ({context, server
   expect(req3.headers['my-secret']).toBe('Value');
 });
 
+it('should throw on ivalid header value', async ({context, server}) => {
```
mxschmitt, line 352:
```suggestion
it('should throw on invalid header value', async ({context, server}) => {
```

src/server/fetch.ts
```diff
@@ -47,8 +47,7 @@ export async function playwrightFetch(context: BrowserContext, params: types.Fet
       agent = new HttpsProxyAgent(proxyOpts);
     }
 
-    // TODO(https://github.com/microsoft/playwright/issues/8381): set user agent
-    const fetchResponse = await sendRequest(context, new URL(params.url), {
+    const fetchResponse = await sendRequest(context, new URL(params.url, context._options.baseURL), {
```
mxschmitt, line 50: Let's use this function maybe? https://github.com/microsoft/playwright/blob/master/src/utils/utils.ts#L343

mxschmitt (approved):

src/server/fetch.ts
yury-s: It returns a string and swallows error, we need neither of those here.

tests/browsercontext-fetch.spec.ts
yury-s: fixed

PR #8566, opened 2021-08-30
chore: implement browserType.connect on the driver side

pavelfeldman:


src/dispatchers/browserTypeDispatcher.ts
```diff
@@ -47,4 +50,29 @@ export class BrowserTypeDispatcher extends Dispatcher<BrowserType, channels.Brow
...
+    });
+    const pipe = new JsonPipeDispatcher(this._scope);
+    ws.on('open', () => pipe.wasOpened());
+    ws.on('close', () => pipe.wasClosed());
+    ws.on('error', error => pipe.wasClosed(error));
+    pipe.on('close', () => ws.close());
+    pipe.on('message', message => ws.send(JSON.stringify(message)));
+    ws.addEventListener('message', event => {
+      waitForNextTask(() => {
+        pipe.dispatch(JSON.parse(event.data));
```
dgozman, line 73: Let's try/catch the JSON.parse and error the pipe.

src/dispatchers/browserTypeDispatcher.ts
```diff
@@ -47,4 +50,29 @@ export class BrowserTypeDispatcher extends Dispatcher<BrowserType, channels.Brow
...
+      maxPayload: 256 * 1024 * 1024, // 256Mb,
+      handshakeTimeout: params.timeout || 30000,
+      headers: paramsHeaders,
+    });
+    const pipe = new JsonPipeDispatcher(this._scope);
+    ws.on('open', () => pipe.wasOpened());
+    ws.on('close', () => pipe.wasClosed());
+    ws.on('error', error => pipe.wasClosed(error));
+    pipe.on('close', () => ws.close());
+    pipe.on('message', message => ws.send(JSON.stringify(message)));
```
dgozman, line 70: - Perhaps we should make the pipe send strings instead of objects to avoid extra parse/stringify?
- I'd try/catch here and send an error if something went wrong. For example, we had to check ws.readyState in some places, otherwise it throws.

src/dispatchers/jsonPipeDispatcher.ts
```diff
@@ -0,0 +1,54 @@
...
+    super(scope, { guid: 'jsonPipe@' + createGuid() }, 'JsonPipe', {});
+  }
+
+  async send(params: channels.JsonPipeSendParams): Promise<channels.JsonPipeSendResult> {
+    this.emit('message', params.message);
+  }
+
+  async close(): Promise<void> {
+    this.emit('close');
+    this._dispatchEvent('closed', {});
```
dgozman, line 33: We need `if (!this._disposed)` here as well, just in case `close` command comes shortly after the pipe has disposed. Perhaps the same with `send`?

src/client/browserType.ts
```diff
@@ -126,48 +123,30 @@ export class BrowserType extends ChannelOwner<channels.BrowserTypeChannel, chann
...
+    return await this._wrapApiCall(async (channel: channels.BrowserTypeChannel) => {
+      const timeoutPromise = new ManualPromise<Browser>();
+      const timer = params.timeout ? setTimeout(() => timeoutPromise.reject(new Error(`Timeout ${params.timeout}ms exceeded.`)), params.timeout) : undefined;
+
+      const { pipe } = await channel.connect({ wsEndpoint, headers: params.headers, timeout: params.timeout });
+      const connection = new Connection(() => pipe.close().catch(() => {}));
+      connection.onmessage = message => pipe.send({ message }).catch(() => { });
+      pipe.on('message', ({ message }) => {
+        try {
+          // Since we may slow down the messages, but disconnect
```
dgozman, line 138: I think this is not relevant anymore.

src/dispatchers/browserTypeDispatcher.ts
pavelfeldman: I can close ws on those.

src/dispatchers/browserTypeDispatcher.ts
pavelfeldman: - In JS there is no extra parse/stringify atm and for other languages, I don't think we can do anymore much more clever while staying a proxy
- So it'll throw in the driver event handler, that is probably ok when the socket is closed?

src/client/browserType.ts
pavelfeldman: The comment isn't, but I think the check still is.

src/client/browserType.ts
```diff
@@ -126,48 +123,27 @@ export class BrowserType extends ChannelOwner<channels.BrowserTypeChannel, chann
...
+      const connection = new Connection(() => pipe.close().catch(() => {}));
+      connection.onmessage = message => pipe.send({ message }).catch(() => { });
+      pipe.on('message', ({ message }) => {
+        try {
+          if (!connection!.isDisconnected())
+            connection!.dispatch(message);
+        } catch (e) {
+          console.error(`Playwright: Connection dispatch error`);
+          console.error(e);
+          pipe.close().catch(() => {});
```
yury-s, line 143: nit: `connection.close()`  here and below?

tests/browsertype-launch-server.spec.ts
```diff
@@ -53,11 +53,11 @@ it.describe('launch server', () => {
     await browserServer.close();
   });
 
-  it('should provide an error when ws endpoint is incorrect', async ({browserType, browserOptions}) => {
+  it.only('should provide an error when ws endpoint is incorrect', async ({browserType, browserOptions}) => {
```
yury-s, line 56: revert

PR #8569, opened 2021-08-31
fix(network): extra info can arrive before will send request

pavelfeldman:


src/client/network.ts
```diff
@@ -421,13 +422,8 @@ export class Response extends ChannelOwner<channels.ResponseChannel, channels.Re
     return { ...this._headers };
   }
 
-  async finished(): Promise<Error | null> {
-    return this._wrapApiCall(async (channel: channels.ResponseChannel) => {
-      const result = await channel.finished();
-      if (result.error)
-        return new Error(result.error);
```
dgozman, line 428: I don't see where we reject `this._finishedPromise`. Should we throw from somewhere? Perhaps we miss a test?

src/server/chromium/crNetworkManager.ts
```diff
@@ -620,9 +629,7 @@ class ResponseExtraInfoTracker {
   }
 
   responseReceivedExtraInfo(event: Protocol.Network.responseReceivedExtraInfoPayload) {
-    const info = this._requests.get(event.requestId);
-    if (!info)
-      return;
+    const info = this._getOrCreateEntry(event.requestId);
```
dgozman, line 632: What if we called `this._stopTracking()` for this info already, and then received extra info for the 10th redirect? It seems like we'll create a new info that will never clean up?

dgozman (approved):

src/server/chromium/crNetworkManager.ts
pavelfeldman: That is the case. It is unlikely though and I don't know how to mitigate it...

src/client/network.ts
pavelfeldman: We did not and we should not...

Issue #8512, opened 2021-08-27
[Feature] Make it easier to bundle playwright


mhils:
#### Problem Description

I have a use case where I want to bundle some code with playwright into a standalone JS file:

```typescript
import {chromium} from "playwright";
console.log(chromium);
```
Doing that with esbuild currently produces multiple warnings and one error:
```shell
esbuild --bundle --platform=node main.ts
 > node_modules/playwright/lib/server/chromium/crApp.js:28:67: warning: "./appIcon.png" should be marked as external for use with "require.resolve"
    28 │   const icon = await _fs.default.promises.readFile(require.resolve('./appIcon.png'));
       ╵                                                                    ~~~~~~~~~~~~~~~

 > node_modules/playwright/lib/server/electron/electron.js:168:51: error: Could not resolve "electron/index.js" (mark it as external to exclude it from the bundle)
    168 │         command: options.executablePath || require('electron/index.js'),
        ╵                                                    ~~~~~~~~~~~~~~~~~~~

 > node_modules/playwright/lib/utils/registry.js:207:23: warning: This call to "require" will not be bundled because the argument is not a string literal (surround with a try/catch to silence this warning)
    207 │   const browsersJSON = require(_path.default.join(packagePath, 'browsers.json'));
        ╵                        ~~~~~~~

2 warnings and 1 error
```

The first warning is fine and can optionally be suppressed with `--external:"*.png"`. 

The following error I don't quite understand (there is no `electron/index.js` anywhere in my playwright distribution), 
but it can also be fixed easily with `--external:"electron/index.js"`.

The final warning hints at the only serious problem, namely a runtime error that occurs when invoking the compiled `main.js`: 
```
internal/modules/cjs/loader.js:888
  throw err;
  ^

Cannot find module '/browsers.json'
```

If one manually patches `browsers.json` into the compiled `main.js`, the resulting standalone JS file works perfectly fine.

#### Proposal

It would be really useful if `browsers.json` could be imported more regularly so that bundling just works. For example, `require('../../browsers.json')` does just work on my platform. I assume there is some reason for not doing that? Alternatively, if `browsers.json` would be located in `src/utils`, `require('./browsers.json'))` could be used and that'd just work as esbuild (and other bundlers) can deal with simple string literals.

#### Version Information
```
$ esbuild --version
0.12.24
$ npx playwright --version
Version 1.14.1
```

aslushnikov:
>  I think all that needs to be changed is in mhils@7b3b918

[...]

PR #8579, opened 2021-08-31
chore(bundling): use a regular import for `browsers.json`

mhils:
As discussed in #8512: Use a regular `import` for the bundled `browsers.json` so that playwright can be bundled with esbuild et al.

I addressed @aslushnikov's comment. Let me know if you want me to merge the `BrowsersJSON` and `BrowsersJSONDescriptor` typedefs, I left them separate for now as they seem to be slightly different. :)

### DIFF
--- a/utils/generate_types/test/tsconfig.json
+++ b/utils/generate_types/test/tsconfig.json
@@ -3,6 +3,7 @@
     "strict": true,
     "target": "ES2019",
     "noEmit": true,
+    "resolveJsonModule": true,
     "moduleResolution": "node"
   },
   "include": [
--- a/utils/generate_types/test/tsconfig.json
+++ b/utils/generate_types/test/tsconfig.json
@@ -3,7 +3,6 @@
     "strict": true,
     "target": "ES2019",
     "noEmit": true,
-    "resolveJsonModule": true,
     "moduleResolution": "node"
   },
   "include": [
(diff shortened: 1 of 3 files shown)


aslushnikov:
@mhils something is not happy about your change! https://github.com/microsoft/playwright/pull/8579/checks?check_run_id=3470779520

utils/generate_types/test/tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "strict": true,
     "target": "ES2019",
     "noEmit": true,
+    "resolveJsonModule": true,
```
mxschmitt, line 6: Can we read it instead of import with require? Thats how we do it for deviceDescriptors

mhils: I've grepped the repo for `deviceDescriptors`, but only found requires/imports. Could you point me to what you are thinking of?

mxschmitt: (edited my message, I meant if require works instead of import)

aslushnikov:
Oof, tsconfig changes! I'm worried the scope of the change is way more now. 

@JoelEinbinder: what's your take on adding these flags to the tsconfigs?

mhils:
> Oof, tsconfig changes! I'm worried the scope of the change is way more now.

Yes, agreed. I've taken up @mxschmitt's suggestion from above to use `require`, that avoids the tsconfig changes.

 If you folks still feel this is getting out of hand please do close this PR, I can easily vendor and patch locally. :)

aslushnikov (approved):

aslushnikov:
Oh this one looks good! LGTM % passed bots

PR #8589, opened 2021-08-31
fix: alias reserved dotnet keyword

avodovnik:


yury-s (approved):
nit: maybe just `value` ?

PR #8601, opened 2021-08-31
chore: audit code for new Promise(async)

pavelfeldman:


src/server/frames.ts
```diff
@@ -927,14 +927,7 @@ export class Frame extends SdkObject {
...
     let cspMessage: ConsoleMessage | undefined;
-    const actionPromise = new Promise<void>(async resolve => {
-      try {
-        result = await func();
-      } catch (e) {
-        error = e;
-      }
-      resolve();
-    });
+    const actionPromise = func().catch(e => error = e);
```
dgozman, line 930: We never assign to "result" anymore?

dgozman (approved):

src/server/frames.ts
pavelfeldman: oh ah

Issue #8583, opened 2021-08-31
[Feature] Allow running multiple (but not all) projects via commandline


MindaugasMateika:
👋   Currently there is an option to run either a single project or all of them. It's very strict. Sometimes you need a combination of projects so it'd be nice if there would be an option to pass f.e. a comma-separated string of values to run multiple projects.

Example:

```
// playwright.config.ts
import { PlaywrightTestConfig } from '@playwright/test';
import { TestOptions } from './my-test';

const config: PlaywrightTestConfig<TestOptions> = {
  timeout: 20000,
  projects: [
    {
      name: 'v1',
      use: { version: '1.0' },
    },
    {
      name: 'v2',
      use: { version: '2.0' },
    },
    {
      name: 'v3',
      use: { version: '3.0' },
    },
    {
      name: 'v4',
      use: { version: '4.0' },
    },
  ]
};
export default config;
```

`npx playwright test --project="v2,v4"`

PR #8604, opened 2021-09-01
fix(runner): disallow duplicate project names

yury-s:
#8583


dgozman (approved):

mxschmitt:
In Playwright we use sometimes the same project name multiple times to share screenshots. This change might break us.

https://github.com/microsoft/playwright/blob/master/tests/config/electron.config.ts

PR #8607, opened 2021-09-01
fix(har): save cookies on request finished

pavelfeldman:


src/server/chromium/crNetworkManager.ts
```diff
@@ -358,6 +358,15 @@ export class CRNetworkManager {
       return;
     const response = this._createResponse(request, event.response);
     this._page._frameManager.requestReceivedResponse(response);
+
+    // fetch to 404 won't emit loadingFinished, do it ourselves.
+    if (response.status() >= 400) {
```
yury-s, line 363: Is it also true for 5xx responses? Let's add a test for this.

PR #8615, opened 2021-09-01
feat(testrunner): support channel values for `browserName` fixture

aslushnikov:


tests/playwright-test/playwright.spec.ts
```diff
@@ -141,6 +141,26 @@ test('should not override use:browserName without projects', async ({ runInlineT
   ]);
 });
 
+test('should support channel values in browserName', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'playwright.config.ts': `
+      module.exports = { use: { browserName: 'chrome' } };
```
mxschmitt, line 147: this would not work with TypeScript? https://github.com/microsoft/playwright/blob/bb5e44fbc48e84ed1f740ae9209148a3ab8ebc52/types/test.d.ts#L3368

PR #8624, opened 2021-09-01
fix(codegen): use new toHaveURL syntax

nygrenh:
Switches the code generation for the 'Playwright Test' target to use the new web-first assertion `expect(page).toHaveURL(url)`. Previously, the codegen used a `.toBe` assertion on a string which less reliable as it did not auto-wait in case of a failure.

### DIFF
--- a/src/server/supplements/recorder/javascript.ts
+++ b/src/server/supplements/recorder/javascript.ts
@@ -95,7 +95,7 @@ export class JavaScriptLanguageGenerator implements LanguageGenerator {
       formatter.add(`]);`);
     } else if (signals.assertNavigation) {
       if (this._isTest)
-        formatter.add(`  expect(${pageAlias}.url()).toBe(${quote(signals.assertNavigation.url)});`);
+        formatter.add(`  await expect(${pageAlias}).toHaveURL(${quote(signals.assertNavigation.url)});`);
       else
         formatter.add(`  // assert.equal(${pageAlias}.url(), ${quote(signals.assertNavigation.url)});`);
     }
--- a/tests/inspector/cli-codegen-1.spec.ts
+++ b/tests/inspector/cli-codegen-1.spec.ts
@@ -539,7 +539,7 @@ test.describe('cli codegen', () => {
     expect(sources.get('Playwright Test').text).toContain(`
   // Click text=link
   await page.click('text=link');
-  expect(page.url()).toBe('about:blank#foo');`);
+  await expect(page).toHaveURL('about:blank#foo');`);
 
     expect(sources.get('Java').text).toContain(`
       // Click text=link


tests/inspector/cli-codegen-1.spec.ts
```diff
@@ -539,7 +539,7 @@ test.describe('cli codegen', () => {
     expect(sources.get('Playwright Test').text).toContain(`
   // Click text=link
   await page.click('text=link');
-  expect(page.url()).toBe('about:blank#foo');`);
+  expect(page).toHaveURL('about:blank#foo');`);
```
yury-s, line 542: The previous assertion is more accurate as by the time click on the link finishes the page should already be on the new URL and there is no point in retrying that check (which is what toHaveURL would do).

nygrenh: Interesting. In one of my projects where I am generating tests for a Next.js application, I have encountered this problem multiple times. Maybe there is some other bug, then?

yury-s: `click()` will autowait for the first navigation if it happens, if there is a subsequent one the url may change again before the place where you check it, in which case toHaveURL would help but I'm not sure this should be default behavior.

nygrenh: In my case there has always been only one navigation after the click.

Here's a proof of concept of the problem: https://github.com/nygrenh/playwright-wait-for-url-problem-poc

It has a page with a link that navigates to another one (https://github.com/nygrenh/playwright-wait-for-url-problem-poc/blob/0b1bce60db4861a85eef53ff35dc438ac8273e24/pages/index.js#L10) and a test (https://github.com/nygrenh/playwright-wait-for-url-problem-poc/blob/0b1bce60db4861a85eef53ff35dc438ac8273e24/tests/example.spec.js). Instructions for running the POC in README.

src/server/supplements/recorder/javascript.ts
```diff
@@ -95,7 +95,7 @@ export class JavaScriptLanguageGenerator implements LanguageGenerator {
       formatter.add(`]);`);
     } else if (signals.assertNavigation) {
       if (this._isTest)
-        formatter.add(`  expect(${pageAlias}.url()).toBe(${quote(signals.assertNavigation.url)});`);
+        formatter.add(`  expect(${pageAlias}).toHaveURL(${quote(signals.assertNavigation.url)});`);
```
pavelfeldman, line 98: should we `await exect`

tests/inspector/cli-codegen-1.spec.ts
yury-s: In your case the navigation is asynchronous and the generated code would actually be:
```js
  await Promise.all([
    page.waitForNavigation(/*{ url: 'http://localhost:3000/page-2' }*/),
    page.click('text=Page 2')
  ]);
```
and this is what you should have in your test to make it pass. So this codegen change would not affect it at all.

src/server/supplements/recorder/javascript.ts
nygrenh: Fixed

yury-s (approved):

PR #8632, opened 2021-09-01
docs: add waitForSelector{visible,hidden}

mxschmitt:


docs/src/api/class-elementhandle.md
```diff
@@ -999,3 +999,27 @@ This method does not work across navigations, use [`method: Page.waitForSelector
...
+
+Returns when element specified by selector satisfies [`option: state`] option. See [`method: ElementHandle.waitForSelector`] for more information.
+
+### param: ElementHandle.waitForSelectorHidden.selector = %%-query-selector-%%
+### option: ElementHandle.waitForSelectorHidden.state = %%-wait-for-selector-state-hidden-%%
+### option: ElementHandle.waitForSelectorHidden.timeout = %%-input-timeout-%%
+### option: ElementHandle.waitForSelectorHidden.strict = %%-input-strict-%%
+
+
+## async method: ElementHandle.waitForSelectorVisible
```
pavelfeldman, line 1016: I thought we agreed to reuse existing method for this though?

mxschmitt: This ends up in typing vs implementation differences which is worse and which we should not have.

pavelfeldman: I understand, but your way ends up with 3 methods, and that might be worse than hacking the overloads for mypy in.

Issue #8490, opened 2021-08-27
[Feature] Allow .goto() to accept a relative URL based on baseURL defined in global-setup


c-kirkeby:
I run my tests using the following command, configured in `scripts` in `package.json`:

``` bash
$ playwright test
```

```typescript
//tests/global-setup.ts
import { chromium } from '@playwright/test'
import dotenv from 'dotenv'

const globalSetup = async () => {
  // Initialise environment variables
  dotenv.config()
  console.log(process.env)

  const browser = await chromium.launch()
  const page = await browser.newPage()
  await page.goto('/') // <-- This line yields the error.
  await page.fill('#okta-signin-username', process.env.AUTH_USERNAME)
  await page.fill('#okta-signin-password', process.env.AUTH_PASSWORD)
  await page.click('#okta-signin-submit')
  await page.context().storageState({ path: './tests/tmp/store_manager.json'})
}

export default globalSetup
```

```typescript
//playwright.config.ts
import { PlaywrightTestConfig } from '@playwright/test'

const config: PlaywrightTestConfig = {
  use: {
    baseURL: 'http://localhost:3000',
    bypassCSP: true,
    headless: false
  },
  globalSetup: './tests/global-setup'
}

export default config
```

```typescript
//tests/example.spec.ts
import { test, expect } from '@playwright/test'

test.use({ storageState: './tests/tmp/store_manager.json'})

test('Basic Auth test', async ({ page }) => {
  await page.goto('http://localhost:3000')
  await expect(page.locator('.menuItem.active > p')).toHaveText('dashboard')
})
```

I'm using Playwright 1.14.0. Trying to run this will yield the error: `page.goto: Protocol error (Page.navigate): Cannot navigate to invalid URL`. It would be good to allow the use of baseURL in global-setup so that I don't have to hardcode URLs. Even being able to somehow pass baseURL into global-setup would be better than hardcoding.

pavelfeldman:
> I can't find any information about this argument in the globalSetup function in the documentation though.

True, lets use this issue to track it.

axelson:
For anyone looking for an example of this, it was removed from the docs in https://github.com/dgozman/playwright/commit/d1a2803a577a5338ad99a971b17c77fdcf228c18
Here is the view right before it was removed https://github.com/dgozman/playwright/blob/759e1b6bdaaf1a50805b7f8066a2820585018d3a/docs/src/auth.md

PR #8637, opened 2021-09-01
docs: update global setup docs

dgozman:
- Changing example to "authenticate once".
- Similarly updating the auth doc.
- Adding a test with the same auth setup.

References #8490.

### DIFF
--- a/docs/src/test-advanced-js.md
+++ b/docs/src/test-advanced-js.md
@@ -270,50 +270,49 @@ test('test', async ({ page, baseURL }) => {
 
 ## Global setup and teardown
 
-To set something up once before running all tests, use `globalSetup` option in the [configuration file](#configuration-object).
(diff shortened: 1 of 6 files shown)


docs/src/test-advanced-js.md
```diff
@@ -270,57 +270,60 @@ test('test', async ({ page, baseURL }) => {
 
 ## Global setup and teardown
 
-To set something up once before running all tests, use `globalSetup` option in the [configuration file](#configuration-object).
+To set something up once before running all tests, use `globalSetup` option in the [configuration file](#configuration-object). Global setup file must export a single function that takes an optional config object. This function will be run once before all the tests.
```
JoelEinbinder, line 273: if the config is optional, when is it not passed?

JoelEinbinder (approved):

PR #8645, opened 2021-09-02
fix(size): fix some of the request.size() problems

pavelfeldman:
This fixes a bunch of issues (chromium had no body size, compression was giving negative number, etc. etc.), but there are more issues. for example:
- Firefox only has transfer size, there is no way of getting body size w/o fetching content. Same goes for WebKit Linux
- We don't need to estimate response headers size if we have content-length header & transfer size

I'd say we should carefully go over all of the scenarios and test them, probably expand 4 -> 12 tests.


pavelfeldman:
(the sync => async migration is unrelated to the issues with the numbers, it was due to the new async nature of headers, raw headers now arrive after `requestfinished` event, so we can't be sync.)

src/server/firefox/ffNetworkManager.ts
```diff
@@ -116,9 +116,7 @@ export class FFNetworkManager {
     if (!request)
       return;
     const response = request.request._existingResponse()!;
-
-    request.request._sizes.transferSize = event.transferSize;
-    request.request._sizes.responseBodySize = event.transferSize - response.headersSize();
```
mxschmitt, line 121: in theory we could get `encodedBodySize` from the httpChannel in FF which is a property there. But I did the calculation manually because it ended up in the same.

tests/har.spec.ts
```diff
@@ -276,18 +276,18 @@ it('should include sizes', async ({ contextFactory, server, asset }, testInfo) =
...
+  expect(log.entries[1].response._transferSize).toBeGreaterThanOrEqual(150);
 });
 
 it('should work with gzip compression', async ({ contextFactory, server, browserName }, testInfo) => {
-  it.fixme(browserName === 'webkit');
+  it.fixme(browserName !== 'chromium');
   const { page, getLog } = await pageWithHar(contextFactory, testInfo);
-  server.enableGzip('/simple.json');
-  const response = await page.goto(server.PREFIX + '/simple.json');
+  server.enableGzip('/simplezip.json');
```
mxschmitt, line 285: Is the reasoning behind changing the file because it contains repeating stuff and gzip internally will shows the compression more directly?

tests/har.spec.ts
```diff
@@ -276,18 +276,18 @@ it('should include sizes', async ({ contextFactory, server, asset }, testInfo) =
...
+  const response = await page.goto(server.PREFIX + '/simplezip.json');
   expect(response.headers()['content-encoding']).toBe('gzip');
   const log = await getLog();
   expect(log.entries.length).toBe(1);
-  expect(log.entries[0].response.content.compression).toBe(-20);
+  expect(log.entries[0].response.content.compression).toBeGreaterThan(4000);
```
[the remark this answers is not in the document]
pavelfeldman: All the header numbers are guesstimates, compression depends on them, so can't be strict.

src/server/firefox/ffNetworkManager.ts
pavelfeldman: headersSize is the problem, we don't have headers size here.

tests/har.spec.ts
pavelfeldman: It was showing compression as -20, while it is >= in the spec, so I needed a positive compression.

tests/har.spec.ts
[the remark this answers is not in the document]
mxschmitt: discussion offline: lets follow up afterwards when we have more tests.

tests/har.spec.ts
mxschmitt: discussed offline: seemed like DevTools did it wrong.

mxschmitt (approved):

Issue #8616, opened 2021-09-01
[BUG] timing of init-scripts?


mindplay-dk:
**Context:**
 - OS: Windows 10 10.0.19043
 - Memory: 15.98 GB / 31.73 GB
 - Node: 14.17.5 - C:\nodejs\node.EXE
 - npm: 7.21.1 - C:\nodejs\npm.CMD
 - Bash: 5.0.17 - C:\Windows\system32\bash.EXE
 - playwright: ^1.14.1 => 1.14.1

**Problem**

I've run into a curious case, which leaves me wondering, when do scripts run, when injected using `addInitScript`?

I'm trying to capture `<script>` tags at page startup - and I prototyped this script in Chrome, using local overrides, by simply injecting my script-tag as the first tag in `<head>`.

In that prototype, `MutationObserver` worked just fine - it captured all the `<script>` tags without problems.

But it appears that scripts injected using `addInitScript` run later than that - using `MutationObserver` does not work, presumably because the document has already been parsed and nodes already added by the time my injected script gets run. (?)

If that's the case though, I should be able to use `document.querySelectorAll("script")` to get the script-nodes that have already been added, right? But that doesn't work either.

This is baffling, to say the least - by the looks of it, the nodes have not already been added, but also do not appear to get added at any later time... I can't either find them in the document or catch them with `MutationObserver`.

It's as though `MutationObserver` doesn't really work? Which is odd. The same example, as explained, worked in Chrome.

I've put together an example script below:

* It prints out the number of scripts already on the page, which is `0`.
* It prints out a message if `MutationObserver` sees a script, which never happens.
* It prints out the URLs of every request - proving that e.g. Google Analytics script is loaded.
* It prints out the browser's cookie state - proving that Google Analytics script was run. (it sets cookies)

The "test subject" `todomvc.com` was chosen at random - I've tried other pages, including `google.com` and `github.com` etc. similarly with no results. Neither `MutationObserver` or `document.querySelectorAll` can find any script nodes.

When exactly are injected init-scripts supposed to run? I would have hoped before the document is populated - e.g. similar to injecting an inline script as the first `<script>` tag in `<head>`... but that may not be the case?

Either way, I can't understand how there could possibly be a time when `<script>` tags on the pages are somehow *not* in the document model yet, but still somehow manage to escape a `MutationObserver` attached within the same blocking call?

It would seem as though `MutationObserver` just doesn't work properly?

Any ideas?

**Code Snippet**

```ts
[...]

PR #8646, opened 2021-09-02
improve documentation for `addInitScript`

mindplay-dk:
* clarifies the fact that init-scripts are executed more than once (per frame)
* explains how to avoid that
* clarifies the fact that init-scripts are run, not only before other scripts, but before the document content is even loaded (see #8616)

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -457,10 +457,9 @@ page.
 
 Adds a script which would be evaluated in one of the following scenarios:
 * Whenever the page is navigated.
-* Whenever the child frame is attached or navigated. In this case, the script is evaluated in the context of the newly
-  attached frame.
+* Whenever a child frame is attached or navigated. In this case, the script is evaluated (repeatedly) in the context of every individual child frame.
 
-The script is evaluated after the document was created but before any of its scripts were run. This is useful to amend
+The script is evaluated immediately after the `document` object is created, but before any nodes are added to it - including the root `<html>` element. This is useful to amend
 the JavaScript environment, e.g. to seed `Math.random`.
 
 An example of overriding `Math.random` before the page loads:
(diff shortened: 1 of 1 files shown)


yury-s (commented):
Looks good, please ensure that `types/types.d.ts` and lint is happy, you can do `npm run watch` locally and add updated files to this PR.

docs/src/api/class-page.md
```diff
@@ -457,10 +457,9 @@ page.
 
 Adds a script which would be evaluated in one of the following scenarios:
 * Whenever the page is navigated.
-* Whenever the child frame is attached or navigated. In this case, the script is evaluated in the context of the newly
-  attached frame.
+* Whenever a child frame is attached or navigated. In this case, the script is evaluated (repeatedly) in the context of every individual child frame.
```
pavelfeldman, line 460: `(repeatedly) in the context` turns this doc from being formally correct to being incorrect. Execution context is an established term that describes the JavaScript world in the frame. addInitScript will only run once in each execution context. Note that when there are other worlds (content scripts), it'll run in them as well.

docs/src/api/class-page.md
```diff
@@ -494,6 +493,15 @@ page.add_init_script(path="./preload.js")
 await page.AddInitScriptAsync(new PageAddInitScriptOption { ScriptPath = "./preload.js" });
 ```
 
+Note that, if a page uses frames or iframes, the script is added to every frame in the document - to run your script only once, you can test of `window` is equal to `window.top`, for example:
```
pavelfeldman, line 496: Also formally incorrect, this is not about `how many times` (once), but about `where` (in the top level frame). So you could say: if you only want to run script inside the main frame of the page (as opposed to child frames), you can check `window === window.top` in the beginning of the script. That way the code runs only in the execution contexts of the main frame.

mindplay-dk:
Sorry, I don't know what any of these comments mean.

I was just trying to help others by describing what I saw, which is this: loading a page causes my init-script to run many times - once for every iframe on the page.

I understand why that's how it's supposed to work, but it confused me. The script I'm injecting was expected to run once, and doesn't care about iframes. I guess I don't know how to describe it "formally", but the point is, it took me a while to realize why the script was running more than once - and the documentation didn't help me understand.

mindplay-dk:
@mxschmitt  documentation for this is still lacking.

@pavelfeldman any chance you could suggest changes? I don't know what your feedback means, but you obviously do 🙂

PR #8651, opened 2021-09-02
fix(recorder): improve rendering of multiline selectors

mhils:
Call me crazy, but I like to use multi-line CSS selectors similar to this one:

```javascript
await page.click(`
    .foo,
    .bar,
    .baz
`);
```

Those break the recorder styling a bit, so here's a fix! 😃 

| before | after |
|----|----|
image | image

### DIFF
--- a/src/web/recorder/callLog.css
+++ b/src/web/recorder/callLog.css
@@ -79,6 +79,7 @@
 
 .call-log-selector {
   color: var(--orange);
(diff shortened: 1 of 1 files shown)


pavelfeldman (approved):

mhils:
Thanks for the super quick turnaround, you folks are fantastic! ❤️

PR #8665, opened 2021-09-02
fix(headers): bring headers closer to raw

pavelfeldman:


src/client/network.ts
```diff
@@ -652,7 +652,8 @@ export class RawHeaders implements api.Headers {
   }
 
   get(name: string): string | null {
-    return this.getAll(name)[0] || null;
+    const values = this.getAll(name) || [];
```
mxschmitt, line 655: personal opinion: this follows Headers.get from the Fetch API but seems not expected as a user. In Go e.g. I'm used that it returns [the first value](https://pkg.go.dev/net/http#Header.Get).

(feel free to resolve it)

mxschmitt (approved):

src/server/supplements/har/harTracer.ts
```diff
@@ -252,8 +252,9 @@ export class HarTracer {
       status: response.status(),
       statusText: response.statusText(),
       httpVersion: response.httpVersion(),
-      cookies: cookiesForHar(response.headerValue('set-cookie'), '\n'),
-      headers: response.headers().map(header => ({ name: header.name, value: header.value })),
+      // These are bad values that will be overwritten bellow.
+      cookies: [],
+      headers: [],
```
dgozman, line 257: I am afraid this will break tracing because we do not wait for the barriers. Can we bring best-effort headers back?

dgozman (approved):

src/server/supplements/har/harTracer.ts
pavelfeldman: Let's sort it out on the tracing side?

Issue #8636, opened 2021-09-01
[BUG] Inconsistent `responseEnd` timing in chromium for network events


vigneshshanmugam:
**Context:**
- Playwright Version: master
- Operating System: Mac
- Node.js version: 14.15
- Browser: Chromium

**Code Snippet**
```ts
it('should produce consistent RT', async ({ contextFactory, }) => {
  const context = await contextFactory();
  const page = await context.newPage();
  const map = new Map();
  context.on('requestfinished', req => {
    const timing  = req.timing();
    const dns = timing.domainLookupEnd - timing.domainLookupStart;
    const connect = timing.connectEnd - timing.connectStart;
    const ssl = timing.connectEnd - timing.secureConnectionStart;
    const wait = timing.responseStart - timing.requestStart;
    const receive = timing.responseEnd - timing.responseStart;
    map.set(req.url(), {
      dns,
      connect,
      ssl,
      wait,
      receive // would include the total time on chrome 
    });
  });
  await page.goto('https://vigneshh.in');
  for (const [key, value] of map.entries())
    console.log(key, value);

  await context.close();
});
```

**Describe the bug**

+ Resource timing data from `request.timing` API is exposing incorrect timing for `responseEnd` data in chromium browser. This results in `receive` part of the HAR denoting the overall duration of the request instead of the specific timing data. Works properly on FF. 

+ `responseEnd` timing actually denotes the complete request duration in chromium as opposed to only including the relative time from responseStart. https://github.com/microsoft/playwright/blob/9618a8477b33d5a6b69ec1a3abe43b193d1d8bc0/src/server/chromium/crNetworkManager.ts#L380

yury-s:
[...]

PR #8668, opened 2021-09-02
chore: test for consistent resource timing values

vigneshshanmugam:
+ adds tests for #8636

### DIFF
--- a/src/server/webkit/wkInterceptableRequest.ts
+++ b/src/server/webkit/wkInterceptableRequest.ts
@@ -79,7 +79,6 @@ export class WKInterceptableRequest {
       return Buffer.from(response.body, response.base64Encoded ? 'base64' : 'utf8');
     };
     const timingPayload = responsePayload.timing;
-    console.log('timingPayload', timingPayload);
     const timing: network.ResourceTiming = {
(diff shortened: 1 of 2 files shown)


vigneshshanmugam:
Thanks for the quick fix @yury-s, Added tests for the same since i already had them.

vigneshshanmugam:
Tests fail on WebKit, could this be a WebKit bug with RT dada? @yury-s @pavelfeldman ?

pavelfeldman:
It most likely is, let's see other platforms.

vigneshshanmugam:
Tried debugging this on webkit, event.timestamp values were wrong for the `loadingFinished` event. Even though the `responseReceived` fires before the `loadingFinished`, the event timestamp value for `responseReceived` is higher than that of `loadingFinished`. 

```
_onResponseReceived -  43.539 ms
_onDataReceived  - 44.039 ms
_onLoadingFinished -  42.445 ms
```

Should we skip this test on webkit?

aslushnikov:
Filed an issue if we ever want to tackle this: https://github.com/microsoft/playwright/issues/11986

Closing for now.

PR #8676, opened 2021-09-03
fix(types): remove `strict` param from Locator.setChecked

YusukeIwaki:
The API definition of `Locator#setChecked` has unused `strict` parameter.
Python client already have it and actually not used: https://github.com/microsoft/playwright-python/pull/879/files#diff-b080c6f27fac96dd688af70aaa7eeb919770d1f37777f9c6e9e1629965e51d28R448

This issue is very similar to #8653 where `ElementHandle#setChecked` is fixed.
I'm afraid I didn't find this issue when I report the issue above.


mxschmitt (approved):

mxschmitt (commented):
I guess it's actually as intended but let's wait for @pavelfeldman on that.

PR #8706, opened 2021-09-04
docs: test parallel/failures docs

### LINKED ISSUE #8505: [Question] Any way to have more than one worker for tests in the same spec file
I have a spec file with many tests and I was wondering if I can have more than one worker for the same spec file? Since all tests are separate browsers, I don't see an issue if I run the tests in one spec file in parallel.
I have no problem running parallel tests when there are multiple spec files.
Thank you!

dgozman:
Fixes #8505.


mxschmitt (approved):
Might make sense to also mention in the first example, that this means that `beforeAll` gets run multiple times (additional to the shared page example).

PR #8722, opened 2021-09-06
 docs(test-runner): add guide on how to typecheck before running tests

ysfaran:
https://playwright.dev/docs/test-typescript already contains a description on how to compile your code and run test afterwards. In my case I really don't care about the JavaScript code and have no problem with the default transpilation of Playwright. 

With the added docs I wanted to highlight that you don't need to compile to any JavaScript and can make the typecheck implementation much simpler.

Hope you like it 🙂

### DIFF
--- a/docs/src/test-typescript-js.md
+++ b/docs/src/test-typescript-js.md
@@ -39,3 +39,25 @@ In my `package.json`, I have two scripts:
 The `pretest` script runs typescript on the tests. `test` will run the tests that have been generated to the `tests-out` directory. The `-c` argument configures the test runner to look for tests inside the `tests-out` directory.
 
 Then `npm run test` will build the tests and run them.
+
+## Type checking
+
+Although Playwright Test is supporting TypeScript out of the box, it's not running any type checks before executing tests *by design*.  
+This allows faster development loops since you don't need to wait for type checking everytime you make some changes to your tests.
+
+It certainly makes sense to add type checks on demand, for example in CI environments:
+
+```bash
+tsc --noEmit
+```
+
+Which requires at least following `tsconfig.json` to be present:
+
+```json
+{
+  "compilerOptions": {
+    "target": "ESNext",
+    "moduleResolution": "Node"
+  }
+}
+```


mskelton:
This feels like a very specific use case to document in the docs which I feel like could add a lot of unnecessary confusion for beginners.  I feel like how type checks are run is something that ought to be left to the consumer to determine how they want to accomplish it.

ysfaran:
@mskelton

How is this very specific?

My thoughts that let me to this solution and creating this PR:

1. Playwright seems to be a cool tool. Oh it also works with TypeScript! I want to use it.
2. I write my tests and run them but after some time they stopped working
3. Going through my files I see some red underlines in my test indicating I have some typing issue. I broke a file that was not even open in the editor -> confusion? why TypeScript didn't complain?
4. Look in docs for TypeScript related stuff -> there is only a rather complicated solution to run playwright from compiled JavaScript code
5. Never the Iess tried it out and failed to get it to run -> more confusion
6. Go to Slack and ask -> get an answer from you, telling good reasons for it
7. I still want to typecheck my code so I don't run into unexpected typescript errors (which is the reason why I use TypeScript in the first place) and don't want to seek for red underlines somewhere
8. Investigating on my own on a solution -> after some time I get to know `--noEmit` and `--incremental` flag of `tsc` and `pretest` in `package.json`
9. Wanted to help other users to not get confused by TypeScript "not working" and created this PR

I don't think that this is really specific, especially the first few points which are actually relevant here. And I think it's especially for beginners very confusing that there is no typecheck at all.

What do you think?

P.S.:

The ideal solution would be an CLI option IMHO:

```ts
playwright test --typecheck
```

mskelton:
> I don't think that this is really specific, especially the first few points which are actually relevant here. And I think it's especially for beginners very confusing that there is no typecheck at all.

So, the docs certainly could be improved to give some guidance to beginners, no argument there.  The struggle you experienced trying to understand what was happening is definitely real.  I totally get that.  My recommendation for docs would be to point them in the direction of resources, perhaps with a few examples as well to give them a broader range of ideas.  This way you cut out many of the steps without prescribing an exact solution as the "right way" if you will.

> `pretest` in `package.json`

Honestly, `pretest` is probably one of my biggest concerns with a solution like this.  Running hooks like `pretest` is going to result in a lot of wasted time running type checks just to catch errors every once and a while.  In my experience, type errors are rarely something that results in the tests I am focused on not working, and a lot of times it is very helpful to actually have type errors while writing tests, since you can write slightly messy tests to experiment and then clean up the types if the experiment proves to be valid.

I think if anything should be recommended it would be running type checks and tests in CI.  That accomplishes the goal of ensuring code is type checked but doesn't slow you down when running tests.  Hopefully this makes sense 🙂 

> playwright test --typecheck

If Playwright was to add a flag like this, now they would have to bundle TypeScript as a dependency, as well as probably have a default tsconfig in case a user did not provide one which could lead to a lot of implied behavior, dependency mismatches, etc.

ysfaran:
@mskelton

Okay I see your point now and have to agree on that. Thanks for clearing things up.

I adapted my changes

JoelEinbinder:
Going to close this for now. We haven't seen a lot of questions about how to add type checking to `@playwright/test`, so I think everything is properly covered by the typescript docs and the red squiggly lines you get in your IDE. If people come with playwright specific questions we can revisit this PR.

PR #8749, opened 2021-09-07
feat(network): remove Headers class and add headersArray method

### LINKED ISSUE #8587: Header model semantic mismatch
Your header model generally looks something like this:

https://github.com/microsoft/playwright/blob/9da26f9534adcbd855e37b1584a94ca673e7f4bf/types/types.d.ts#L13228-L13231

Strictly speaking, headers are not a key/value map - and in practice, this causes some problems, for example with `set-cookie`, which frequently occurs more than once.

Apparently, the way you address this, is by concatenating multiple header-values with a `\n` (line-break) character.

It doesn't look like this detail is described in [documentation](https://playwright.dev/docs/api/class-request#request-headers) - and by the time you uncover this, it makes it quite difficult and clumsy to work with the model.

For example, the widely-used [set-cookie-parser](https://www.npmjs.com/package/set-cookie-parser) package has some extremely weird behavior when you pass it a multi-value header like this - and the `splitCookiesString` function doesn't help here either, as it expects multi-value headers to be separated by commas rather than line-breaks.

It's not obvious how to add a header either - doing so correctly is pretty clumsy:

```js
function addHeader(headers, name, value) {
  if (name in headers) {
    headers[name] = header[name] + "\n" + value;
  } else {
    headers[name] = value;
  }
}
```

It's not clear to me if the line-breaks detail is the only detail either, or just the one I've discovered

Is there anything else I need to know?

Either way, I would suggest a deprecation of the existing methods and a replacement with some sort of abstraction/type to properly deal with headers?

That or a wrapper-type for the current model, to take care of the the splitting and recombining, adding/removing headers, etc. - which I will most likely write under any circumstances to avoid these issues.

At the very least, these details should be documented.

I am happy to help with whatever course of action you prefer. 🙂

JoelEinbinder:
Fixes #8587


pavelfeldman (approved):

mindplay-dk:
@pavelfeldman If I might ask, what is the motivation for using tuples?

Tuples as a data structure are pretty unintuitive - `Array<Array<string>>` sounds like a list of list of strings, and I had to actually read through the source-code to understand how that could possibly represent list of headers.

Even if you know what a tuple is (many developers don't) the type-hint does not reflect the fact that the inner list of strings is in fact a two-element tuple.

I would strongly suggest you use objects proper, e.g. `Array<Array<{ name: string, value: string }>>`, probably with the inner type defined separately and named `Header`.

Or at least use a tuple type, e.g. `Array<Array<{ [string, string] }>>`, which at least constraints the inner array to two elements - not that this helps with the semantic problem or makes the code very readable, as you will see still things like `header[0]` etc. in client code.

But I don't believe you need tuples in modern JS engines - the language is highly optimized for objects, both in terms of performance and memory usage. If this was done as a micro-optimization, it's an extremely marginal one, and probably not really worth the confusion it's going to cause.

JoelEinbinder:
> @pavelfeldman If I might ask, what is the motivation for using tuples?
> 
> Tuples as a data structure are pretty unintuitive - `Array<Array<string>>` sounds like a list of list of strings, and I had to actually read through the source-code to understand how that could possibly represent list of headers.
> 
> Even if you know what a tuple is (many developers don't) the type-hint does not reflect the fact that the inner list of strings is in fact a two-element tuple.
> 
> I would strongly suggest you use objects proper, e.g. `Array<Array<{ name: string, value: string }>>`, probably with the inner type defined separately and named `Header`.
> 
> Or at least use a tuple type, e.g. `Array<Array<{ [string, string] }>>`, which at least constraints the inner array to two elements - not that this helps with the semantic problem or makes the code very readable, as you will see still things like `header[0]` etc. in client code.
> 
> But I don't believe you need tuples in modern JS engines - the language is highly optimized for objects, both in terms of performance and memory usage. If this was done as a micro-optimization, it's an extremely marginal one, and probably not really worth the confusion it's going to cause.

I like tuples because they match Object.entries and you can use them to immediately initialize an object with Object.fromEntries(headersArray) or new Map(headersArray). It's not for any performance reason. I would like the types to be `[string, string][]` to show that its a tuple properly in typescript, but our doc gen code doesn't support it yet.

tests/page/page-network-request.spec.ts
```diff
@@ -289,8 +288,8 @@ it('should report all headers', async ({ page, server, browserName, platform })
       ]
     }))
   ]);
-  const headers = await request.allHeaders();
-  expect(headers.headers()).toEqual(expectedHeaders);
+  const headers = await request.headersArray();
+  expect(headers.sort()).toEqual(expectedHeaders.sort());
```
YusukeIwaki, line 292: fetch API actually sends headers to server like below:

```
headers-a: value-a, value-a-1, value-a-2
headers-b: value-b
```

and headersArray() returns

```
[
  ['headers-a', 'value-a, value-a-1, value-a-2'],
  ['headers-b', 'value-b'],
]
```

Is it expected...?

I thought this test want to check the behavior with headers like

```
header-a: value-a
header-b: value-b
header-a: value-a-1
header-a: value-a-2
```

however it seems JS fetch API doesn't allow to send such request...

mindplay-dk:
> I like tuples because they match Object.entries and you can use them to immediately initialize an object with Object.fromEntries(headersArray) or new Map(headersArray).

Well, you'd have to convert the header-names to lowercase first - and, more importantly, you still have to deal with recurring headers, as using `Object.fromEntries` or `new Map` directly would randomly discard some headers.

So, not really.

I really think the best thing would be to have an actual header model - something like:

```ts
interface Header {
  name: string;
  value: string;
}

class HeaderMap {
  private headers: { [nameLower: string]: Header[] } = {};
  
  constructor(headers: Array<[string, string]>) {
    for (let [name, value] of headers) {
      this.add(name, value);
    }
  }

  get(name: string): string[] {
    return (this.headers[name.toLowerCase()] || []).map(({ value }) => value);
  }

  add(name: string, value: string): void {
    const nameLower = name.toLowerCase();

    this.headers[nameLower] = this.headers[nameLower]
      ? this.headers[nameLower].concat({ name, value })
      : [{ name, value }];
  }

  replace(name: string, value: string): void {
    this.headers[name.toLowerCase()] = [{ name, value }];
  }

  delete(name: string): void {
    delete this.headers[name.toLowerCase()];
  }

  values(): Header[] {
    const headers: Header[] = [];

    for (const name in this.headers) {
      headers.push(...this.headers[name]);
    }

    return headers;
  }
}
```

The standard object/map/set etc. types just weren't built for multi-valued maps. This has methods for adding or replacing multi-valued headers - it's optimized for case-insensitive lookups, while preserving the case of the individual headers, so it's completely lossless, and it lets you convert back to an array when needed.

I opted for a `Header` object type here, because I don't see the point of tuples - standard object/map/set types aren't really helpful because of multi-value headers, as well as the error prone operation of adding to a *(lossy)* multi-valued collection like `Map<string, string[]>` to support multi-valued headers, which this class encapsulates.

You could add more toThis/toThat methods, if necessary - but I really think this would be more elegantly handled with a single header-model built for the expected, normal use-cases, eliminating common mistakes, lossiness, and so on.

@pavelfeldman what do you think? You seemed to be moving in this direction earlier, but something made you change your mind?

PR #8761, opened 2021-09-07
fix(recorder): move recorder scripts into the main world

### LINKED ISSUE #7515: [Feature]: Edit a playwright script via codegen
### Feature request

I have an existing playwright script, but I want to take some additional complicated steps in the middle of that existing script.

As far as I understand, I have to manually write those steps or `codegen` from scratch.
I'd like to use the same great `codegen` tool to edit my existing script.

Some ideas of how to implement this:

* `codegen --edit existing-script.js --line 46`
* `codegen --edit existing-script.js`
  * (this would work by having some identifier to replace in the file. `// {{ insert point }}`

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/7515

### DIFF
--- a/src/server/supplements/recorderSupplement.ts
+++ b/src/server/supplements/recorderSupplement.ts
@@ -184,11 +184,11 @@ export class RecorderSupplement implements InstrumentationListener {
     // Input actions that potentially lead to navigation are intercepted on the page and are
     // performed by the Playwright.
     await this._context.exposeBinding('_playwrightRecorderPerformAction', false,
-        (source: BindingSource, action: actions.Action) => this._performAction(source.frame, action), 'utility');
+        (source: BindingSource, action: actions.Action) => this._performAction(source.frame, action));
 
     // Other non-essential actions are simply being recorded.
     await this._context.exposeBinding('_playwrightRecorderRecordAction', false,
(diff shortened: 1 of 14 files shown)


src/server/supplements/recorderSupplement.ts
```diff
@@ -181,14 +181,16 @@ export class RecorderSupplement implements InstrumentationListener {
       recorderApp.close().catch(() => {});
     });
 
+    // Everything moved into the main world to prevent from stalling evaluation.
+
     // Input actions that potentially lead to navigation are intercepted on the page and are
     // performed by the Playwright.
     await this._context.exposeBinding('_playwrightRecorderPerformAction', false,
-        (source: BindingSource, action: actions.Action) => this._performAction(source.frame, action), 'utility');
+        (source: BindingSource, action: actions.Action) => this._performAction(source.frame, action), 'main');
```
yury-s, line 189: Running this in the main world was slow on some web sites (namely cvs.com) that the recorder was unusable. The problem was that the page included a library that overrides String.trim() with an inefficient implementation. This change will definitely regress it. Why can't controls work in the util world?

pavelfeldman: See the bug attached.

PR #8768, opened 2021-09-08
fix: a human-readable error when lock cannot be acquired

### LINKED ISSUE #8709: Why does the compilation fail during installation
**Context:**
- Playwright Version: latest
- Operating System: big sur
- Node.js version: 16.7.0
- Browser: none
- Extra: none

image

As shown, I'm tried to install playwright. But this throws an error. Why?

aslushnikov:
image

Fixes #8709

### DIFF
--- a/src/utils/registry.ts
+++ b/src/utils/registry.ts
@@ -562,6 +562,25 @@ export class Registry {
         else
           throw new Error(`ERROR: Playwright does not support installing ${executable.name}`);
       }
+    } catch (e) {
+      if (e.code === 'ELOCKED') {
+        const rmCommand = process.platform === 'win32' ? 'rm -R' : 'rm -rf';
+        throw new Error('\n' + wrapInASCIIBox([
+          `An active lockfile is found at:`,
+          ``,
+          `  ${lockfilePath}`,
+          ``,
+          `Either:`,
+          `- wait a few minutes if other Playwright is installing browsers in parallel`,
+          `- remove lock manually with:`,
+          ``,
+          `    ${rmCommand} ${lockfilePath}`,
+          ``,
+          `<3 Playwright Team`,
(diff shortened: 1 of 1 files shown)


src/utils/registry.ts
```diff
@@ -562,6 +562,25 @@ export class Registry {
...
+          `- wait a few minutes if other Playwright is installing browsers in parallel`,
+          `- remove lock manually with:`,
+          ``,
+          `    ${rmCommand} ${lockfilePath}`,
+          ``,
+          `<3 Playwright Team`,
+        ].join('\n'), 1));
+      } else {
+        throw e;
+      }
```
mxschmitt, line 583:
```suggestion
      }
      throw e;
```
nit

src/utils/registry.ts
```diff
@@ -562,6 +562,25 @@ export class Registry {
         else
           throw new Error(`ERROR: Playwright does not support installing ${executable.name}`);
       }
+    } catch (e) {
+      if (e.code === 'ELOCKED') {
+        const rmCommand = process.platform === 'win32' ? 'rm -R' : 'rm -rf';
```
mxschmitt, line 567: nit: do we need `-r` there?

mxschmitt (approved):

src/utils/registry.ts
aslushnikov: has to be in parenthesis since the first `if` clause is multi-line.

aslushnikov: ah, i'd rather have it this way i guess!

src/utils/registry.ts
aslushnikov: just in case we have a lockfile actually inside the lock directory at some future point of time.

PR #8778, opened 2021-09-08
Update documentation to include pre-release install information

unlikelyzero:
A questions surrounding pre-release versions came up in slack recently. 

I've updated the readme to include basic instructions on how to install the latest development version or other pre-release versions.


mxschmitt:
Thank you for your contribution, but we usually don't recommend the `@next` tag to users. We advice it to specific user-groups when needed manually.

PR #8782, opened 2021-09-08
chore(expect): remove some dependencies on jest internals

JoelEinbinder:
As Simen suggests in https://github.com/facebook/jest/pull/11816, we can remove the
dependency on `jest-matcher-utils` by using `this.utils` in our matchers. I cleaned
up some other places were we use internal types that are not needed.

We still are requiring two unexported files: `expect/build/matchers` and `expect/build/print`

﻿

### DIFF
--- a/src/test/matchers/toMatchText.ts
+++ b/src/test/matchers/toMatchText.ts
@@ -19,14 +19,6 @@ import {
   printReceivedStringContainExpectedSubstring
 } from 'expect/build/print';
 
-import {
-  EXPECTED_COLOR,
-  matcherErrorMessage,
-  matcherHint, MatcherHintOptions,
-  printExpected,
-  printWithType,
-  printDiffOrStringify,
-} from 'jest-matcher-utils';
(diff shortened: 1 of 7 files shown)


src/test/matchers/toMatchText.ts
```diff
@@ -45,7 +37,7 @@ export async function toMatchText(
     throw new Error(`${matcherName} must be called during the test`);
   expectType(receiver, receiverType, matcherName);
 
-  const matcherOptions: MatcherHintOptions = {
```
dgozman, line 48: Should we copy this type definition for our own safety?

dgozman (approved):

src/test/matchers/toMatchText.ts
JoelEinbinder: It gets checked down below when we pass it in as an argument.

mrienstra:
@dgozman, @JoelEinbinder, is this ready to merge? Anything I can do to help if not?

PR #8786, opened 2021-09-08
fix(video): reduce buffering in ffmpeg, avoid overbooking cpu

dgozman:
This is an attempt to improve video performance when encoding
does not keep up with frames. This situation can be reproduced
by running multiple encoders at the same time.

Added `utils/video_stress.js` to manually reproduce this issue.

---

Observing ffmpeg logs, it does not do any encoding initially and
instead does "input analysis / probing" that detects fps and other
parameters. By the time it starts encoding (launches vpx and creates
the video file), we already have many frames in the buffer.

Reducing probing helps: `-avioflags direct -fpsprobesize 0 -probesize 32 -analyzeduration 0`

Note that using another suggested option `-fflags nobuffer` corrupts
the first ~5 seconds of the resulting video.

---

Another issue observed is questionable default `-threads` value.
We compile without threads support, so logs say "using thread emulation".
For some reason, setting explicit `-threads 1` (or any other value)
makes it better when cpu is loaded.

---

Local experiments with using vp9 show (surprisingly) 10-30%
file size increase, but also 10X less stalling when cpu was overbooked
(down from 100s to 10s post-processing time). It could be beneficial
to switch to vp9.

### DIFF
--- a/src/server/chromium/videoRecorder.ts
+++ b/src/server/chromium/videoRecorder.ts
@@ -63,31 +63,41 @@ export class VideoRecorder {
     //   $ ./third_party/ffmpeg/ffmpeg-mac -h encoder=vp8
     // 3. A bit more about passing vp8 options to ffmpeg.
     //   https://trac.ffmpeg.org/wiki/Encode/VP8
+    // 4. Tuning for VP9:
+    //   https://developers.google.com/media/vp9/live-encoding
+    //
+    // How to stress-test video recording (runs 10 recorders in parallel to book all cpus available):
+    //   $ node ./utils/video_stress.js
     //
     // We use the following vp8 options:
     //   "-qmin 0 -qmax 50" - quality variation from 0 to 50.
     //     Suggested here: https://trac.ffmpeg.org/wiki/Encode/VP8
     //   "-crf 8" - constant quality mode, 4-63, lower means better quality.
-    //   "-deadline realtime" - do not use too much cpu to keep up with incoming frames.
+    //   "-deadline realtime -speed 8" - do not use too much cpu to keep up with incoming frames.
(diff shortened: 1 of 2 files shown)


src/server/chromium/videoRecorder.ts
```diff
@@ -63,31 +63,41 @@ export class VideoRecorder {
...
+    //   "-f image2pipe -c:v mjpeg -i -" forces input to be read from standard input, and forces
+    //     mjpeg input image format.
+    //   "-avioflags direct" reduces general buffering.
+    //   "-fpsprobesize 0 -probesize 32 -analyzeduration 0" reduces initial buffering
+    //     while analyzing input fps and other stats.
     //
     // "-y" means overwrite output.
     // "-an" means no audio.
+    // "-threads 1" means using one thread. This drastically reduces stalling when
+    //   cpu is overbooked. By default vp8 tries to use all available threads?
```
yury-s, line 96: typo: '...all available threads' => '...all available cores' ?

src/server/chromium/videoRecorder.ts
```diff
@@ -63,31 +63,41 @@ export class VideoRecorder {
     //   $ ./third_party/ffmpeg/ffmpeg-mac -h encoder=vp8
     // 3. A bit more about passing vp8 options to ffmpeg.
     //   https://trac.ffmpeg.org/wiki/Encode/VP8
+    // 4. Tuning for VP9:
```
yury-s, line 66: We use vp8 not vp9, don't we? Maybe it's time to switch to vp9?

yury-s (approved):

src/server/chromium/videoRecorder.ts
dgozman: That's on the agenda.

PR #8803, opened 2021-09-09
fix(test-runner): allow underscore as fixture name

### LINKED ISSUE #8798: [Question] How to ignore first argument in test()
I'm using @playwright/test v1.14.1.
I followed the instruction in this page https://playwright.dev/docs/test-parallel/#reusing-single-page-between-tests. 
All seems fine, until I try to ignore first parameter in the `test()` by using _ as I normally doing in typescript.
When I try to run the test I got this error: `Error: First argument must use the object destructuring pattern: _`

```js
test.describe.serial('test case 1', () => {
  test('test step 1', async (_, testInfo) => {
    // Error: First argument must use the object destructuring pattern: _
  })
})
```

mxschmitt:
It came up a few times in the past, since ESLint does not like [empty object patterns](https://eslint.org/docs/rules/no-empty-pattern), we want to give users a better experience by allowing the underscore for them.

Fixes #8798

### DIFF
--- a/src/test/fixtures.ts
+++ b/src/test/fixtures.ts
@@ -309,6 +309,8 @@ function innerFixtureParameterNames(fn: Function, location: Location): string[] 
   if (!trimmedParams)
     return [];
   const [firstParam] = splitByComma(trimmedParams);
+  if (firstParam === '_')
+    return [];
   if (firstParam[0] !== '{' || firstParam[firstParam.length - 1] !== '}')
     throw errorWithLocations('First argument must use the object destructuring pattern: '  + firstParam, { location });
   const props = splitByComma(firstParam.substring(1, firstParam.length - 1)).map(prop => {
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
I don't particularly like how this fails for the user without them knowing whats going on. My autocomplete tells me I can do `_.page` and then when I try to I get page is undefined.

I don't have a good solution that isn't more involved though. We could construct some magic proxy object that throws a helpful error message when you do _.page, but I think that will require a bit of plumbing. So I guess I am fine with this.

There should be some docs component though.

dgozman:
> I don't have a good solution that isn't more involved though. We could construct some magic proxy object that throws a helpful error message when you do _.page, but I think that will require a bit of plumbing.

+1 for either ignoring this issue, or doing a proxy-like stuff. Another possibility is to contribute to eslint and allow `{}` as a function argument because why not?

Issue #8340, opened 2021-08-20
[BUG:Webkit] `page.goto() / page.waitForLoadState()` method does not wait for the `load` event


canonic-epicure:
**Context:**
- Playwright Version: 1.14.0
- Operating System: Ubuntu 18
- Node.js version: v12.22.5

Bug appears only with Webkit launcher.

Steps to reproduce (reduced to a bare minimum):
1) Clone this repo: `git clone https://github.com/canonic-epicure/playwright-8340-repro`
2) In the checkout directory, run `npm install`
3) In console, run `npx exec wds`. This will launch the dev web server on port 8000 (should be free, port is hardcoded)
4) Now, run the `node server.js` in the separate shell.
5) Observe the `TypeError: window.CLIENT is not a function` exception
6) Uncomment this line in `server.js`:
`await page.waitForFunction(() => document.readyState === 'complete')` 
This line performs a "manual" waiting for the `load` event
7) Run `node server.js` again and observe the exception is gone.
8) Comment that line back and make sure the exception appears again
9) Comment these lines in `client.js`:
```
import { it } from 'https://cdn.jsdelivr.net/npm/@bryntum/siesta@latest/index.js'
it('', () => {})
```
10) Observe the exception disappears (probably, since its a race condition).

The rootcause of the issue seems to be that `client.js` is importing some other big external module.

canonic-epicure:
Actually, I just checked the docs and it seems the `page.goto()` method itself is supposed to wait for `load` event: https://playwright.dev/docs/api/class-page#page-goto

Please feel free to change the title of the of ticket

pavelfeldman:
I started watching the video and noticed that your script has a `module` type. JavaScript modules are async, they imply `defer`, which I think explains the behavior you are observing.

You might want to use

```js
await page.waitForFunction(() => !!window.connect));
```

to ensure connect object is created.

canonic-epicure:
@pavelfeldman 

> JavaScript modules are async, they imply defer

I don't think so, can you provide a link to spec or MDN?

For this simple test file:
```html
<!DOCTYPE HTML>
<html>
    <head>
        <script type="module">
            window.addEventListener('load', () => console.log('LOAD'))
            window.addEventListener('DOMContentLoaded', () => console.log('DOMContentLoaded'))
        </script>

        <script type="module" src="test1.js"></script>
        <script type="module" src="test2.js"></script>
    </head>

    <body>
    </body>
</html>
```

[...]

PR #8810, opened 2021-09-09
browser(webkit): wait for parsing before emitting Page.frameStoppedLoading

JoelEinbinder:
The code here is very complicated, but experimentally this line works.

Maybe this can get upstreamed and somebody from the webkit team can take a look? There is some accessibility
code running with the progress controller so I think it could be a real bug.

References #8340﻿

### DIFF
--- a/browser_patches/webkit/BUILD_NUMBER
+++ b/browser_patches/webkit/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1542
-Changed: dpino@igalia.com Wed 08 Sep 2021 10:52:59 AM UTC
+1543
+Changed: joel.einbinder@gmail.com Thu 09 Sep 2021 08:50:37 AM PDT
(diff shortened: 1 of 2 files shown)


pavelfeldman:
That's pretty random, I'm sure you are breaking the browser.

JoelEinbinder:
> That's pretty random, I'm sure you are breaking the browser.

I'm not sure what to do with this comment.

canonic-epicure:
You could also just manually wait for the `document.readyState` property to have a required value?

canonic-epicure:
@JoelEinbinder A friendly bump for this PR, because it seems to be stalled and I'm interested in the original issue to be fixed.

Could it be fixed in the "user land" instead of "C++ land"? One can manually wait for the `document.readyState` to have correct value? This can be applied only for Webkit.

dgozman:
We discussed this one, and the way forward is to avoid using `Page.frameStoppedLoading` signal. @joeleinbinder suggests that we'll need a separate signal for `window.stop()` handling which is ok.

Issue #8806, opened 2021-09-09
[BUG] failure in BeforeEach does not block to run next BeforeEach block


eLeontev:
**Context:**
- Playwright Version: 1.14.1
- Playwright/test Version: 1.14.1
- Operating System: Windows
- Node.js version: 14.17.1
- Browser: -
- Extra: default startup with **npx playwright test** and **TypeScript**

**Code Snippet**

```typescript
import { expect, test } from '@playwright/test';

const bfeWithFailure = () => {
    throw new Error('123');
};

const bfeWithLog = () => console.log('should not be called');

test.describe.only('test', () => {
    test.beforeEach(bfeWithFailure);

    test.beforeEach(bfeWithLog);

    test('should not be called', () => {
        expect(1).toBe(1);
    });
});
```

**Describe the bug**
1. Prepare the flow with unhandled error in the first **beforeEach** block
2. Run test example

**Actual result:** **bfeWithLog** is called.
**Expected result:** **bfeWithLog** is not called.

eLeontev:
**Update:** Looks like I missed to put any test framework results.
Fortunately in Jest(tested with: 26.5.3) the behavior the same even if couple of tickets were directly addressed to **beforeAll** with **async** callback. 

Link to the issue: https://github.com/facebook/jest/issues/8688

Jest behavior:
![image](https://user-images.githubusercontent.com/15786916/132830510-7f2253a3-8305-415e-b1ab-19a44236e784.png)

JoelEinbinder:
It looks like our current behavior matches jest/jasmine. What you are describing is mocha style, where the failing hook stops the next hook from running. That gets tricky when you have a pattern like

```js
before(() => {
 // set up something important
});
after(() => {
 // throw an error
});
after(() => {
 // tear down the important thing
});
```

The first failing after causes the second not to run, but it might be very important that it runs. There are pros and cons of each approach, so let's stick with what we have.

PR #8812, opened 2021-09-09
test(test-runner): failing beforeEach should skip subsequent beforeEach

JoelEinbinder:
What do we want to do in this case?

I am checking the behavior of other test runners﻿

References #8806


JoelEinbinder:
This test matches the mocha behavior, but our current behavior matches jest/jasmine. Let's stick with the current behavior.

PR #8819, opened 2021-09-09
docs: no Response.finished() return value

mxschmitt:


pavelfeldman:
This is a breaking change!

pavelfeldman (approved):

PR #8822, opened 2021-09-09
chore(report): don't generate file per test

pavelfeldman:


src/test/reporters/raw.ts
```diff
@@ -199,44 +200,43 @@ class RawReporter {
...
           name: attachment.name,
           contentType: attachment.contentType,
-          path: toPosixPath(file)
+          body: attachment.body.toString('base64')
+        });
+      } else if (attachment.path) {
+        attachments.push({
+          name: attachment.name,
+          contentType: attachment.contentType,
+          path: toPosixPath(attachment.path)
```
dgozman, line 216: I think we should leave this a native path to be able to read from FS later.

src/test/html/htmlBuilder.ts
```diff
@@ -30,31 +31,44 @@ export class HtmlBuilder {
...
+        const testFile: TestFile = {
+          fileId,
+          path: file.location!.file,
+          tests: tests.map(t => this._createTestCase(t))
+        };
+        fs.writeFileSync(path.join(dataFolder, fileId + '.json'), JSON.stringify(testFile, undefined, 2));
+      }
+      projects.push({
+        name: projectJson.project.name,
         suits,
```
dgozman, line 52: typo: suits -> suites

dgozman (approved):

PR #8824, opened 2021-09-09
feat(network): request.getHeaderValue, response.getHeaderValue, response.getHeaderValue

JoelEinbinder:
Wdyt about getHeader/getHeaders instead of getHeaderValue/getHeaderValues?

### DIFF
--- a/docs/src/api/class-request.md
+++ b/docs/src/api/class-request.md
@@ -77,12 +77,12 @@ Headers with multiple entries, such as `Set-Cookie`, appear in the array multipl
 An array with all the request HTTP headers associated with this request. Unlike [`method: Request.allHeaders`], header names are not lower-cased.
 Headers with multiple entries, such as `Set-Cookie`, appear in the array multiple times.
 
-## async method: Request.getHeaderValue
+## async method: Request.headerValue
 - returns: <[null]|[string]>
 
 Returns the value of the header matching the name. The name is case insensitive.
 
-### param: Request.getHeaderValue.name
+### param: Request.headerValue.name
 - `name` <[string]>
 
 Name of the header.
--- a/docs/src/api/class-request.md
+++ b/docs/src/api/class-request.md
@@ -77,6 +77,17 @@ Headers with multiple entries, such as `Set-Cookie`, appear in the array multipl
 An array with all the request HTTP headers associated with this request. Unlike [`method: Request.allHeaders`], header names are not lower-cased.
 Headers with multiple entries, such as `Set-Cookie`, appear in the array multiple times.
 
+## async method: Request.getHeaderValue
+- returns: <[null]|[string]>
+
+Returns the value of the header matching the name. The name is case insensitive.
+
+### param: Request.getHeaderValue.name
+- `name` <[string]>
+
+Name of the header.
+
+
 ## method: Request.isNavigationRequest
 - returns: <[boolean]>
 
--- a/docs/src/api/class-response.md
+++ b/docs/src/api/class-response.md
@@ -46,7 +46,7 @@ Headers with multiple entries, such as `Set-Cookie`, appear in the array multipl
 ## async method: Response.headerValue
 - returns: <[null]|[string]>
 
-Returns the first value of the header matching the name. The name is case insensitive. If multiple headers have the same name, for example `set-cookie`, the first is returned. If no header is found, `null` is returned.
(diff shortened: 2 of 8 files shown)


docs/src/api/class-request.md
```diff
@@ -77,6 +77,17 @@ Headers with multiple entries, such as `Set-Cookie`, appear in the array multipl
 An array with all the request HTTP headers associated with this request. Unlike [`method: Request.allHeaders`], header names are not lower-cased.
 Headers with multiple entries, such as `Set-Cookie`, appear in the array multiple times.
 
+## async method: Request.getHeaderValue
```
pavelfeldman, line 80: We don't use `get` prefix.

docs/src/api/class-response.md
```diff
@@ -43,6 +43,26 @@ Headers with multiple entries, such as `Set-Cookie`, appear in the array multipl
 An array with all the request HTTP headers associated with this response. Unlike [`method: Response.allHeaders`], header names are not lower-cased.
 Headers with multiple entries, such as `Set-Cookie`, appear in the array multiple times.
 
+## async method: Response.getHeaderValue
+- returns: <[null]|[string]>
+
+Returns the first value of the header matching the name. The name is case insensitive. If multiple headers have the same name, for example `set-cookie`, the first is returned. If no header is found, `null` is returned.
```
pavelfeldman, line 49: We should concatenate values using `', '` as a separator instead.

JoelEinbinder: I thought the reason why set cookie can be specified more than once is because cookies can have ',' in their values. How would someone use this?

docs/src/api/class-request.md
JoelEinbinder: Done.

docs/src/api/class-response.md
mxschmitt: +1 to Joel but [`Headers.get`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/get) is also doing this weird comma separation logic :/

pavelfeldman: By the spec, accept-encoding: gzip, deflate is indistinguishable from accept-encoding: gzip\r\naccept-encoding: deflate. Yet you would want to get "gzip, deflate" as the value.

JoelEinbinder: done

docs/src/api/class-request.md
```diff
@@ -77,6 +77,17 @@ Headers with multiple entries, such as `Set-Cookie`, appear in the array multipl
 An array with all the request HTTP headers associated with this request. Unlike [`method: Request.allHeaders`], header names are not lower-cased.
 Headers with multiple entries, such as `Set-Cookie`, appear in the array multiple times.
 
+## async method: Request.headerValue
```
yury-s, line 80: Should it also have `Request.headerValues` ?

JoelEinbinder: no, because there are never multiple headers with the same name on a request.

src/client/network.ts
```diff
@@ -161,6 +161,16 @@ export class Request extends ChannelOwner<channels.RequestChannel, channels.Requ
     return (await this._getHeadersIfNeeded()).map(header => [header.name, header.value]);
   }
 
+  async headerValue(name: string): Promise<string | null> {
+    const headers = await this._getHeadersIfNeeded();
```
pavelfeldman, line 165: I would prefer to have the same implementation as with headerValue for response.

docs/src/api/class-response.md
```diff
@@ -43,6 +43,26 @@ Headers with multiple entries, such as `Set-Cookie`, appear in the array multipl
 An array with all the request HTTP headers associated with this response. Unlike [`method: Response.allHeaders`], header names are not lower-cased.
 Headers with multiple entries, such as `Set-Cookie`, appear in the array multiple times.
 
+## async method: Response.headerValue
+- returns: <[null]|[string]>
+
+Returns the first value of the header matching the name. The name is case insensitive. If multiple headers have the same name, for example `set-cookie`, the they are returned as a list separated by `, `. If no header is found, `null` is returned.
```
pavelfeldman, line 49: "Returns the first value of the header matching the name." seems wrong.

JoelEinbinder: done

src/client/network.ts
JoelEinbinder: done

PR #8828, opened 2021-09-10
fix(test): make `use` in config accept option values only

### LINKED ISSUE #8814: [BUG] config.projects[0].use is not typed anymore
https://playwright.dev/docs/next/auth/#reuse-authentication-in-playwright-test

image

Reported via: https://playwright.slack.com/archives/C01AG7QGW2F/p1631207091128300?thread_ts=1631201567.123900&cid=C01AG7QGW2F

dgozman:
Also include default options in FullConfig/FullProject.
Also make examples compile and add a test.

Fixes #8814.

### DIFF
--- a/docs/src/auth.md
+++ b/docs/src/auth.md
@@ -193,11 +193,11 @@ async function globalSetup(config: FullConfig) {
   const { baseURL, storageState } = config.projects[0].use;
   const browser = await chromium.launch();
   const page = await browser.newPage();
-  await page.goto(baseURL);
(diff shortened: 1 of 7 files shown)


mxschmitt:
Could you please remove the type cast here? Had it as a workaround: https://github.com/microsoft/playwright/blob/e40b805782757c932410b431f6fe905d9375af60/src/test/matchers/matchers.ts#L243

JoelEinbinder (approved):

PR #8829, opened 2021-09-10
docs(python): add docs about threading

### LINKED ISSUE #876: Docs about multithreading with Playwright

kumaraditya303:
Fixes https://github.com/microsoft/playwright-python/issues/876

### DIFF
--- a/docs/src/intro-python.md
+++ b/docs/src/intro-python.md
@@ -189,9 +189,9 @@ Playwright runs the driver in a subprocess, so it requires `ProactorEventLoop` o
 
 On Windows Python 3.7, Playwright sets the default event loop to `ProactorEventLoop` as it is default on Python 3.8+.
 
-### Threading and Playwright Sync API are incompatible
+### Threading
 
-Playwright's Sync API is not thread-safe. If you are using Playwright in a multi-threaded environment, you should create a playwright instance per thread. See [threading issue](https://github.com/microsoft/playwright-python/issues/623) for more details.
+Playwright's API is not thread-safe. If you are using Playwright in a multi-threaded environment, you should create a playwright instance per thread. See [threading issue](https://github.com/microsoft/playwright-python/issues/623) for more details.
 
 
 ## System requirements
--- a/docs/src/intro-python.md
+++ b/docs/src/intro-python.md
@@ -189,6 +189,11 @@ Playwright runs the driver in a subprocess, so it requires `ProactorEventLoop` o
 
 On Windows Python 3.7, Playwright sets the default event loop to `ProactorEventLoop` as it is default on Python 3.8+.
 
+### Threading and Playwright Sync API are incompatible
+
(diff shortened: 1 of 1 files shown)


docs/src/intro-python.md
```diff
@@ -189,6 +189,11 @@ Playwright runs the driver in a subprocess, so it requires `ProactorEventLoop` o
 
 On Windows Python 3.7, Playwright sets the default event loop to `ProactorEventLoop` as it is default on Python 3.8+.
 
+### Threading and Playwright Sync API are incompatible
+
+Playwright's Sync API is not thread-safe. If you are using Playwright in a multi-threaded environment, you should create a playwright instance per thread. See [threading issue](https://github.com/microsoft/playwright-python/issues/623) for more details.
```
mxschmitt, line 194: Isn't it the same for the async API?

kumaraditya303: No async API can be called from other thread with coroutine thread safe

mxschmitt: Please don't differentiate between sync and async and formulate it like: Playwright API's are not thread safe. We recommend to create new processes (to ensure no variables are shared) for concurrent usage. Threading also works when a new Playwright instance gets created in each thread.

kumaraditya303: That is what I have described but it is *only* applicable for sync api,not  in async api.

mxschmitt: I'm not sure at all if we need this in our docs, since if you try to use it in different threads and call functions on it, you'll get for async: `ValueError: The future belongs to a different loop than the one specified as the loop argument`
and for sync: `greenlet.error: cannot switch to a different thread`

kumaraditya303: See here https://docs.python.org/3/library/asyncio-task.html#asyncio.run_coroutine_threadsafe and it is possible with this method for async only @mxschmitt

kumaraditya303: Moreover threading is mainly used in sync hence there is no need of mentioning async in it and complicating it more IMO

kumaraditya303: @mxschmitt I have given sufficient info about async and why it is should be only for sync so any other changes required here ?

docs/src/intro-python.md
```diff
@@ -189,6 +189,11 @@ Playwright runs the driver in a subprocess, so it requires `ProactorEventLoop` o
 
 On Windows Python 3.7, Playwright sets the default event loop to `ProactorEventLoop` as it is default on Python 3.8+.
 
+### Threading and Playwright Sync API are incompatible
```
mxschmitt, line 192:
```suggestion
### Threading
```

docs/src/intro-python.md
mxschmitt:
```suggestion
Playwright's API is not thread-safe. If you are using Playwright in a multi-threaded environment, you should create a playwright instance per thread. See [threading issue](https://github.com/microsoft/playwright-python/issues/623) for more details.
```

mxschmitt (approved):

mxschmitt:
Thank you!

PR #8837, opened 2021-09-10
chore(electron): bump electron to v12

JoelEinbinder:
Also  needed to bump node types to v14﻿


package.json
```diff
@@ -91,7 +91,7 @@
     "@types/extract-zip": "^1.6.2",
     "@types/mime": "^2.0.3",
     "@types/minimatch": "^3.0.3",
-    "@types/node": "^10.17.28",
+    "@types/node": "^14.17.15",
```
mxschmitt, line 94: do we need 14.x? since our minimum right now which we support is 12.x.

tests/page/frame-evaluate.spec.ts
```diff
@@ -36,10 +36,13 @@ it('should have correct execution contexts', async ({ page, server }) => {
 });
 
 function expectContexts(pageImpl, count, browserName) {
-  if (browserName === 'chromium')
-    expect(pageImpl._delegate._mainFrameSession._contextIdToContext.size).toBe(count);
-  else
+  if (browserName === 'chromium') {
+    const contexts: Map<number, any> = pageImpl._delegate._mainFrameSession._contextIdToContext;
+    const notElectron = [...contexts.values()].filter(context => context._delegate._contextNameForTest !== 'Electron Isolated Context');
```
dgozman, line 41: You can just skip these tests in Electron and remove _contextNameForTest, since they seem to not test anything there.

dgozman (approved):

tests/page/frame-evaluate.spec.ts
JoelEinbinder: Given that electron creates its own isolated worlds, it seems like something could break here in a different way from stock chromium. I'd like to keep the tests.

mxschmitt (approved):

mxschmitt:
We should also bump this one:

https://github.com/microsoft/playwright/blob/40ae28e3bbd72b50e87a1d3c2a7be652812d62e5/packages/installation-tests/installation-tests.sh#L419

Or make it a range from 12...15?

PR #8844, opened 2021-09-10
chore: remove unused .idea folder

JoelEinbinder:
It looks like these files were added to the .gitignore, but not actaully deleted.
﻿


dgozman (commented):
Why don't we leave them and instead clean up the gitignore? I'd like to use this idea project if possible.

pavelfeldman (approved):

pavelfeldman:
I no longer remember if we need these files or not...

PR #8846, opened 2021-09-10
fix(sizes): responseBodySize on CR & WK macOS

mxschmitt:


src/server/webkit/wkPage.ts
```diff
@@ -1048,6 +1048,9 @@ export class WKPage implements PageDelegate {
       });
       if (event.metrics?.protocol)
         response._setHttpVersion(event.metrics.protocol);
+      if (event.metrics?.responseBodyBytesReceived)
+        request.request.responseSize.encodedBodySize = event.metrics.responseBodyBytesReceived;
```
pavelfeldman, line 1052: Try usinig responseHeadersSize too?

tests/page/page-network-sizes.spec.ts
```diff
@@ -88,7 +89,10 @@ it('should have the correct responseBodySize for chunked request', async ({ page
   });
   const response = await page.goto(server.PREFIX + '/chunked-simplezip.json');
   const sizes = await response.request().sizes();
-  expect(sizes.responseBodySize).toBe(fs.statSync(asset('simplezip.json')).size);
+  if (browserName === 'webkit')
+    expect(sizes.responseBodySize).toBe(5173);
```
pavelfeldman, line 93: Let's add a comment that it should be 5175, where the extra 75 are coming from. Also that webkit is not telling the truth here.

pavelfeldman (approved):

PR #8847, opened 2021-09-10
feat(har): add _requestref

rwoll:
This will be used by our traceviewer-like (userland) debug tools
that require correlating logging emitted from Playwright-dependent code
and the requests that end up in the HAR file after the run.

This can be utilized by Playwright’s own traceviewer some day, too. For
example, it can be used to link back code that transformed a request to
the actual request that ended up getting recorded in the HAR file.

### DIFF
--- a/src/server/supplements/har/harTracer.ts
+++ b/src/server/supplements/har/harTracer.ts
@@ -141,6 +141,7 @@ export class HarTracer {
     const pageEntry = this._ensurePageEntry(page);
     const harEntry: har.Entry = {
       pageref: pageEntry.id,
+      _requestref: request.guid,
       _frameref: request.frame().guid,
       _monotonicTime: monotonicTime(),
       startedDateTime: new Date(),
--- a/src/server/supplements/har/har.ts
+++ b/src/server/supplements/har/har.ts
@@ -55,6 +55,7 @@ export type Entry = {
   timings: Timings;
   serverIPAddress?: string;
   connection?: string;
+  _requestref: string;
   _frameref: string;
   _monotonicTime: number;
   _serverPort?: number;
--- a/tests/har.spec.ts
+++ b/tests/har.spec.ts
@@ -589,3 +589,42 @@ it('should have different hars for concurrent contexts', async ({ contextFactory
     expect(pageEntry.title).toBe('One');
   }
 });
+
+it('should include _requestref', async ({ contextFactory, server }, testInfo) => {
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+  const resp = await page.goto(server.EMPTY_PAGE);
+  const log = await getLog();
+  expect(log.entries.length).toBe(1);
+  const entry = log.entries[0];
+  expect(entry._requestref).toMatch(/^request@[a-f0-9]{32}$/);
+  expect(entry._requestref).toBe((resp.request() as any)._guid);
+});
+
+it('should include _requestref for redirects', async ({ contextFactory, server }, testInfo) => {
+  server.setRedirect('/start', '/one-more');
+  server.setRedirect('/one-more', server.EMPTY_PAGE);
+
+  const { page, getLog, context } = await pageWithHar(contextFactory, testInfo);
+
+  const requests = new Map<string, string>();
+  context.on('request', request => {
+    requests.set(request.url(), (request as any)._guid);
+  });
+
+  await page.goto(server.PREFIX + '/start');
+
+  const log = await getLog();
+  expect(log.entries.length).toBe(3);
+
+  const entryStart = log.entries[0];
+  expect(entryStart.request.url).toBe(server.PREFIX + '/start');
+  expect(entryStart._requestref).toBe(requests.get(entryStart.request.url));
+
+  const entryOneMore = log.entries[1];
+  expect(entryOneMore.request.url).toBe(server.PREFIX + '/one-more');
+  expect(entryOneMore._requestref).toBe(requests.get(entryOneMore.request.url));
+
+  const entryEmptyPage = log.entries[2];
+  expect(entryEmptyPage.request.url).toBe(server.EMPTY_PAGE);
+  expect(entryEmptyPage._requestref).toBe(requests.get(entryEmptyPage.request.url));
+});


src/server/network.ts
```diff
@@ -107,6 +108,7 @@ export class Request extends SdkObject {
     url: string, resourceType: string, method: string, postData: Buffer | null, headers: types.HeadersArray) {
     super(frame, 'request');
     assert(!url.startsWith('data:'), 'Data urls should not fire requests');
+    this._id = createGuid();
```
pavelfeldman, line 111: SdkObject already has `_guid`.

src/server/supplements/har/harTracer.ts
```diff
@@ -141,6 +141,7 @@ export class HarTracer {
     const pageEntry = this._ensurePageEntry(page);
     const harEntry: har.Entry = {
       pageref: pageEntry.id,
+      _id: request.id(),
```
pavelfeldman, line 144: If har had this property, i would be ok with it. But since it is an extension, given that there is no good use case for this story, i'm reluctant to letting it in.

src/server/network.ts
rwoll: Thanks! Updated

src/server/supplements/har/harTracer.ts
rwoll: @pavelfeldman I empathize with the hesitation. If it's helpful in your team's decision, here is what I was thinking while proposing this MR and why I felt it was appropriate:

1. Concrete, actionable use case. This isn't theoretical. With this field, we can delete buggy heuristic code in our codebase that we use to debug and do other slicing of the HAR data. Without this change, we either have to live with bugs or set out to re-implement all of the HarTracer code in user-land. (Another user/company (@vigneshshanmugam) sounds like they have a use case from Slack, too).
2. There's no way today, aside from heuristics, to tie a Request object you see in Node back with an entry in the HAR. (If there is, please let me know as that is the sole purpose of this PR. No need to add new API if already doable.)
3. HAR format is under spec'd. Every browser, HAR proxy, networking debug tool I've used adds fields that are useful to the user, so as long as they don't break people's code, I don't see the harm in adding them if they can be used within reason.
4. `playwright@next` (`playwright@1.15.0-next-alpha-sep-10-2021`) introduces 2 extension fields (that I don't see in Chromium [`_frameref`](https://source.chromium.org/search?q=_frameref&sq=&ss=chromium) and [`_monotonicTime`](https://source.chromium.org/search?q=_monotonicTime&ss=chromium)):
    ```
    "entries": [
      {
        "pageref": "page@702b0129b3d3993b3eb0fe38cae484ba",
        "_frameref": "frame@dfaf6341b6e76342a310147e06861f2b",     <-------------
        "_monotonicTime": 364665600.663, <---------------
    ```
    While these are useful for playwright internally, we'd find the `_id` extension field useful for us today. (Let me know if there's another place I should look for a standard.)

Would it make more sense to put it in the "connection" field?

> connection [string, optional] (new in 1.2) - Unique ID of the parent TCP/IP connection, can be the client or server port number. Note that a port number doesn't have to be unique identifier in cases where the port is shared for more connections. If the port isn't available for the application, any other unique connection ID can be used instead (e.g. connection index). Leave out this field if the application doesn't support this info.

source: http://www.softwareishard.com/blog/har-12-spec/

pavelfeldman:
Thank you for those additional details. First, here is where I think we see har differently: for me, har is a persistence format. It is written into file, you are not supposed to poke at it at a runtime. We think of it as a snapshot we dump.

But it looks like it is useful to you at runtime. I real it not as a feature request to har, but rather a feature request to our request api. We already have a rule that har should be composable in the user land. We are exposing timing, sizes, raw headers, everything you might see in har with our apis. So maybe we could look at it from another angle and give you convenience you miss in the request api?

rwoll:
> Thank you for those additional details. First, here is where I think we see har differently: to me, har is a persistence format. It is written into file, you are not supposed to poke at it at a runtime. We think of it as a snapshot we dump.
> 
> 
> 
> But it looks like it is useful to you at runtime. I read it not as a feature request to har, but rather a feature request to our request api. We already have a rule that har should be composable in the user land. We are exposing timing, sizes, raw headers, everything you might see in har with our apis. So maybe we could look at it from another angle and give you convenience you miss in the request api?

@pavelfeldman Let me go back to the problem we are trying to solve and we can work from there. I might have jumped to a solution without explaining what I'm trying to solve! 😄 Please let me know if this makes sense:

Here’s what we’re looking to solve:

1.	We run code (with Playwright), that does some stuff including Request transformations and captures (via context.on(“request”)). We have har recording enabled.
2.	Once that code is done running, we sometimes do nothing. Other times (after the fact i.e. hours to days later), in a separate tool, we analyze the HAR alongside logs from (1) if issues arise to debug.

The information from both (1) and (2) are essential in our debug flow, but we don’t have a perfect way to tie them together as is today. We have heuristics, but they are sometimes painfully off and make debugging and slicing the HAR up hard.

Anything that would allow us to tie a HAR entry (at step 2) back to a `console.log(request.someInformation())` (from step 1) would be extremely helpful and welcomed especially if it's exact like the proposed ID.

I think this flow also sees HAR as snapshot/artifact (and not a runtime thing).

Thanks for working through this! 😀

rwoll:
Added an edit to the end of last comment that modifies HAR but doesn't change public Request API. Don't love it, but it would solve the issue.

pavelfeldman:
Ok, so you have your own log + har log and you need to trace the request identify. You like HAR because some other tool opens it nicely. And your options are either expose id or fingerprint the request. First requires changes upstream, second is ugly. And you think you can convince us making this change upstream. This makes sense!

I guess we can justify this:
```js
_requestref: request.guid,
```

It'll help us correlating har with the protocol calls in the trace viewer. We don't need it now, but we might and it would make sense.

rwoll:
> Ok, so you have your own log + har log and you need to trace the request identify. You like HAR because some other tool opens it nicely. And your options are either expose id or fingerprint the request. First requires changes upstream, second is ugly. And you think you can convince us making this change upstream. This makes sense!
> 
> 
> 
> I guess we can justify this:
> 
> ```js
> 
> _requestref: request.guid,
> 
> ```
> 
> 
> 
> It'll help us correlating har with the protocol calls in the trace viewer. We don't need it now, but we might and it would make sense.

@pavelfeldman Exactly! perfect summary! Thanks! Amended the patch and the description to reflect the change.

pavelfeldman (approved):

PR #8859, opened 2021-09-11
fix(codegen): NUnit's Equal API should be Assert.AreEqual

doggy8088:


pavelfeldman:
Looks like it broke codegen tests

dgozman (approved):
Thank you for the PR!

PR #8863, opened 2021-09-11
docs(python): Specifying use of context element in multi-pages Python documentation

flavienbwk:
Current documentation for multi-pages doesn't explain how to get the `context` element in example.

This PR replaces `context.expect_page()` by `page.context.expect_page()` for example to be immediately usable.

### DIFF
--- a/docs/src/multi-pages.md
+++ b/docs/src/multi-pages.md
@@ -189,7 +189,7 @@ System.out.println(newPage.title());
 
 ```python async
 # Get page after a specific action (e.g. clicking a link)
-async with context.expect_page() as new_page_info:
+async with page.context.expect_page() as new_page_info:
     await page.click('a[target="_blank"]') # Opens a new tab
 new_page = await new_page_info.value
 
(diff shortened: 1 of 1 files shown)


pavelfeldman:
It looks backwards to me, you always have context. What is your use case?

flavienbwk:
> It looks backwards to me, you always have context. What is your use case?

Hi, I'm not sure about what you're referring to as "always having a context". See the following example : 

```python
with sync_playwright() as p:
    browser = p.chromium.launch(headless=False)
    page = browser.new_page()
    #...
    with page.context.expect_page() as page_info:
        page.wait_for_selector('.form-group > .description > a')
        page.click('.form-group > .description > a') # opens new tab
        new_page = page_info.value
```

If I put `context.expect_page()` instead of `page.context.expect_page()`, it raises an error : 

```
Traceback (most recent call last):
  File "scripts/crawler.py", line 39, in <module>
    with context.expect_page() as page_info:
NameError: name 'context' is not defined
```

mxschmitt:
>  page = browser.new_page()

this does internally:

```py
context = browser.new_context()
page = context.new_page()
return page
```

In your script you can either do what I did above or access the context via `page.context`

PR #8877, opened 2021-09-13
devops: rework client side changes bot

mxschmitt:
Outcome looks like that: https://github.com/microsoft/playwright-python/issues/900

We'll follow up by changing it to a different PAT then.


aslushnikov (approved):
lgtm % pavel's token

PR #8880, opened 2021-09-13
test: skip CLI tests on tracing bots

mxschmitt:
image

Broken since https://github.com/microsoft/playwright/pull/8761


pavelfeldman:
Do you know why they failed? (Maybe fix it if that is something simple?)

Issue #8865, opened 2021-09-11
[Internal] fetch API review


pavelfeldman:
Release blocking
- [x] `fetch` name does not make sense for API testing, they want `get`, `post`, `delete`, etc.
- [x] `postData` should accept key/value pairs and end up with `application/json` encoding
  - [x] there should be separate postData notation for `application/x-www-form-urlencoded` encoding
  - [x] read header `multipart/form-data` and encode accordingly
  - [x] `postData` is not a great name
- [x] should accept parsed query params
- [ ] needs top-level shortcuts for popular headers, at least `content-type` and `accept`
- [x] add validate status option

Nice to have:
- [x] make API call without context
- [x] request.create() -> FetchRequest w/ cookie jar, settings
- [x] needs explicit `ignoreHTTPSErrors`
- [ ] should retry requests (https://visionmedia.github.io/superagent/#retrying-requests)
- [x] make a series of API calls that produce storage state (cookies)
- [x] have default user-agent on the global fetch: Playwright/<version>

pavelfeldman:
Proposed API shape:
```js
import { request } from 'playwright';
await request.get('http://localhost:9222');
await request.post('http://localhost:9222', { a: b });

await context.request.get('/'); // uses baseURL
```

```js
test('API call', async ({ request }) => {
  // does not affect cookies
  const response = await request.get('/');
});
```

```js
test('API call', async ({ page }) => {
  // affects cookies
  const response = await page.request.get('/');
});
```

PR #8893, opened 2021-09-13
feat(fetch): support query params

yury-s:
#8865

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -288,6 +288,24 @@ class HashStream extends stream.Writable {
   }
 }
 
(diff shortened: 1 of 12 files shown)


src/utils/utils.ts
```diff
@@ -288,6 +288,24 @@ class HashStream extends stream.Writable {
   }
 }
 
+export function objectToArray(map?: { [key: string]: string }): { name: string, value: string }[] | undefined {
```
pavelfeldman, line 291: NameValue

pavelfeldman (approved):

src/utils/utils.ts
yury-s: Done.

Issue #8892, opened 2021-09-13
[internal] create-playwright release blockers


mxschmitt:
- [x] `end-to-end` vs. ~~`integration`~~ naming clarification
- [x] `Add GitHub Actions workflow` -> `Add a GitHub Actions workflow`
- [x] Add description of the commands which get executed which get reflected in the logs
- [x] Add `test-results/` to the .gitignore?
- [x] compare with create-react-app about a better rootDir default. -> CRA creates new applications we get applied on existing ones mostly.
- [x] Maybe use something different then `npm run playwright-tests`?
- [x] better debug command --debug CLI flag on CLI https://github.com/microsoft/playwright/pull/8938
- [x] - --project= use double quotes.
- [x] dont warn if gitignore gets overriden.
- [x] point to the config and test file which got created
- [x] ~~ Add a gif how it can be used on NPM/main readme to show which features it supports~~
- [x] ~~Add a README in the `e2e` directory with more information.~~
- [x] evaluate better CLI examples
- [x] reduce projects
- [x] add more examples of use properties

rwoll:
Some initial feedback:

* refuse to overwrite existing files. I (intentionally) had it overwrite an existing package.json but it should refuse to do this out of the box. This is P0/1 to me.
* take a directory as the first argument or prompt the user for the directory to init in. I didn't expect it to be CWD. (See create-react-app UX.)
* publish as `@playwright/create`. The official Playwright ecosystem has a bit of mixed naming (playwright, @playwright/test, playwright-chromium, create-playwright) making it hard to tell what exactly is official from  Playwright team (and there's new tools every week)
* include GitLab CI/CD option. We don't all use Microsoft's GitHub :)

Overall, awesome tool and will help folks getting started!

mxschmitt:
> refuse to overwrite existing files. I (intentionally) had it overwrite an existing package.json but it should refuse to do this out of the box. This is P0/1 to me.

I tried to repro this, but was not able to. If a package.json exists in the directory, it will only update it with adding @playwright/test there and add playwright to the scripts section. I can prevent the modification of the scripts section if there is already an entry for Playwright.

> take a directory as the first argument or prompt the user for the directory to init in. I didn't expect it to be CWD. (See create-react-app UX.)

The difference between CRA and us is that in their case a new application gets always created. In our case it often gets added to an existing project, so having . as a default is IMO fine. Also I've added a log message in the beginning which will print the path where it gets invoked in (or path passed) so its more clear.

[...]

PR #8897, opened 2021-09-13
chore(create-playwright): UX follow-ups

mxschmitt:
#8892

New outro is now in sync with what React & Next.js will print:

image

React:

```
Success! Created my-app at /Users/max/development/playwright/packages/create-playwright/my-app
Inside that directory, you can run several commands:

  npm run dev
    Starts the development server.

  npm run build
    Builds the app for production.

  npm start
    Runs the built app in production mode.

We suggest that you begin by typing:

  cd my-app
  npm run dev
```

Next.js:

```
✨  Done in 15.63s.

Success! Created test at /Users/max/development/playwright/packages/create-playwright/test
Inside that directory, you can run several commands:

  yarn start
    Starts the development server.

  yarn build
    Bundles the app into static files for production.

  yarn test
    Starts the test runner.

  yarn eject
    Removes this tool and copies build dependencies, configuration files
    and scripts into the app directory. If you do this, you can’t go back!

We suggest that you begin by typing:

  cd test
  yarn start

Happy hacking!
```

### DIFF
--- a/packages/create-playwright/assets/playwright.config.js
+++ b/packages/create-playwright/assets/playwright.config.js
@@ -19,7 +19,7 @@ const config = {
 
   projects: [
     {
-      name: 'Desktop Chromium',
+      name: 'Desktop Chrome',
       use: {
         ...devices['Desktop Chrome'],
       },
--- a/packages/create-playwright/assets/playwright.config.js
+++ b/packages/create-playwright/assets/playwright.config.js
@@ -21,9 +21,7 @@ const config = {
     {
       name: 'Desktop Chromium',
       use: {
-        browserName: 'chromium',
-        // Test against Chrome Beta channel.
-        channel: 'chrome-beta',
+        ...devices['Desktop Chrome'],
       },
     },
     {
--- a/packages/create-playwright/src/index.ts
+++ b/packages/create-playwright/src/index.ts
@@ -19,7 +19,7 @@ import fs from 'fs';
 import { prompt } from 'enquirer';
 import colors from 'ansi-colors';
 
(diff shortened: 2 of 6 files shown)


packages/create-playwright/assets/playwright.config.js
```diff
@@ -21,9 +21,7 @@ const config = {
     {
       name: 'Desktop Chromium',
       use: {
-        browserName: 'chromium',
-        // Test against Chrome Beta channel.
-        channel: 'chrome-beta',
+        ...devices['Desktop Chrome'],
```
mxschmitt, line 24: using chrome-* or normal Chrome will reduce the barrier of entry, since on e.g. on Codespaces this is not installed by default. If we would invoke it all the time (npx playwright install chrome) the browser would also get killed which is not a great user experience.

packages/create-playwright/src/index.ts
```diff
@@ -97,15 +97,28 @@ class Generator {
...
+
+    commands.push({
+      name: 'Downloading browsers',
+      command: 'npx playwright install --with-deps',
+    });
+
+    let gitIgnore = '';
+    if (fs.existsSync(path.join(this.rootDir, '.gitignore')))
+      gitIgnore = fs.readFileSync(path.join(this.rootDir, '.gitignore'), 'utf-8').trimEnd() + '\n';
+    gitIgnore += 'test-results/';
```
mxschmitt, line 120:
```suggestion
    gitIgnore += 'test-results/\n';
```

aslushnikov (approved):

PR #8903, opened 2021-09-14
feat(report): render attachment as a part of failure

pavelfeldman:


src/test/reporters/base.ts
```diff
@@ -153,17 +153,38 @@ export class BaseReporter implements Reporter  {
...
+        // Make this extensible
+        if (attachment.name === 'trace') {
+          lines.push(colors.cyan(`    Usage:`));
+          lines.push('');
+          lines.push(colors.cyan(`        npx playwright show-trace ${relativePath}`));
+          lines.push('');
+        }
+      } else {
+        if (attachment.contentType.startsWith('text/'))
+          lines.push(colors.cyan(`    ${attachment.body?.toString()}`));
```
dgozman, line 184: Should we trimEnd() this?

src/test/reporters/base.ts
```diff
@@ -153,17 +153,38 @@ export class BaseReporter implements Reporter  {
...
+      }
+      lines.push(colors.cyan('    ' + '-'.repeat(100 - 3)));
```
dgozman, line 186: I think we have a `pad` function in this file.

dgozman (approved):

PR #8920, opened 2021-09-14
Feature: Addition of Tags to JSON Reporter Output

shairozan:
# Background

I've used tools like Newman to do E2E testing for APIs, and have used their custom reporters and tagging to facilitate documentation / validation of a product in CI. While looking into doing something similar with Playwright, I realized the JSON reporter didn't expose the tags passed to a spec in the output (shy of still being in the title).

When going complex user acceptance testing, being able to map from User Story -> Requirement -> Test is important, and using tags to identify the tests for that purpose is extremely beneficial. While we can add tags with playwright, they're *absent* in the JSON reporter's output shy of being in the title of the spec itself. 

This pull request updates the reporter such that the spec serializer locates all tags by regex with word boundary, plops them into an array and then strips the preceding `@` since it's more of an identifier for a tag than the content of the tag.

### DIFF
--- a/src/test/reporters/json.ts
+++ b/src/test/reporters/json.ts
@@ -202,7 +202,7 @@ class JSONReporter implements Reporter {
 
   private _serializeTestSpec(test: TestCase): JSONReportSpec {
 
-    let fixedTags: string[] = []
+    const fixedTags: string[] = [];
     const tagMatches: RegExpMatchArray| null = test.title.match(/@[\S]+/g);
 
     if (tagMatches) {
--- a/src/test/reporters/json.ts
+++ b/src/test/reporters/json.ts
@@ -44,6 +44,7 @@ export interface JSONReportSuite {
   suites?: JSONReportSuite[];
 }
 export interface JSONReportSpec {
+  tags: string[],
   title: string;
   ok: boolean;
   tests: JSONReportTest[];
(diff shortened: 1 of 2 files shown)


src/test/reporters/json.ts
```diff
@@ -200,9 +201,20 @@ class JSONReporter implements Reporter {
   }
 
   private _serializeTestSpec(test: TestCase): JSONReportSpec {
+
```
pavelfeldman, line 204: My comments are mostly nits, not sure why eslint did not complain here:
- remove empty line

src/test/reporters/json.ts
```diff
@@ -200,9 +201,20 @@ class JSONReporter implements Reporter {
...
+    const fixedTags: string[] = [];
+    const tagMatches: RegExpMatchArray| null = test.title.match(/@[\S]+/g);
+
+    if (tagMatches) {
```
pavelfeldman, line 208:
```js
const tags = (test.title.match(/@[\S]+/g) || []).map(t => t.substring(1));
```

src/test/reporters/json.ts
shairozan: No problem. I'm not a JS / TS developer, so it's all relatively new to me. I'll update with a new commit shortly

src/test/reporters/json.ts
shairozan: Going that route didn't even require assignment up front. Could just set tags inline during the return statement. Got an updated commit in. Thanks for the help!

pavelfeldman (approved):

PR #8927, opened 2021-09-14
feat(fetch): introduce global fetch request

yury-s:
#5999

### DIFF
--- a/src/server/fetch.ts
+++ b/src/server/fetch.ts
@@ -22,209 +22,303 @@ import * as https from 'https';
 import { BrowserContext } from './browserContext';
 import * as types from './types';
 import { pipeline, Readable, Transform } from 'stream';
-import { monotonicTime } from '../utils/utils';
+import { createGuid, monotonicTime } from '../utils/utils';
+import { SdkObject } from './instrumentation';
+import { Playwright } from './playwright';
+import { HeadersArray, ProxySettings } from './types';
+import { HTTPCredentials } from '../../types/types';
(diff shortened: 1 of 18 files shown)


src/server/network.ts
```diff
@@ -228,7 +228,7 @@ export class Route extends SdkObject {
     if (body === undefined) {
       if (overrides.fetchResponseUid) {
         const context = this._request.frame()._page._browserContext;
-        const buffer = context.fetchResponses.get(overrides.fetchResponseUid);
+        const buffer = context.fetchRequest.fetchResponses.get(overrides.fetchResponseUid);
```
dgozman, line 231: This means I can only use context._request for response interception? I think we should be able to use the global one as well.

src/server/fetch.ts
```diff
@@ -22,209 +22,303 @@ import * as https from 'https';
...
-      if (proxy.username)
-        proxyOpts.auth = `${proxy.username}:${proxy.password || ''}`;
-      agent = new HttpsProxyAgent(proxyOpts);
-    }
+type FetchRequestOptions = {
+  userAgent: string;
+  extraHTTPHeaders?: HeadersArray;
+  httpCredentials?: HTTPCredentials;
+  proxy?: ProxySettings;
+  timeout: (options: {timeout?: number}) => number;
```
dgozman, line 38: Perhaps reuse TimeoutSettings?

dgozman (approved):

src/server/network.ts
yury-s: Will do in a separate PR.

PR #8932, opened 2021-09-15
chrome: don't nest pw:api steps

pavelfeldman:


dgozman (approved):

src/client/types.ts
```diff
@@ -27,7 +26,8 @@ export interface Logger {
 }
 
 export interface ClientSideInstrumentation {
-  onApiCall(stackTrace: ParsedStackTrace): ((error?: Error) => void) | undefined;
+  onApiCallBegin(apiCall: string): { userObject: any };
```
dgozman, line 29: I think we can replace `{ userObject: any }` in both methods with `any` and it should be equivalent?

tests/playwright-test/reporter.spec.ts
```diff
@@ -328,8 +330,8 @@ test('should report api step failure', async ({ runInlineTest }) => {
     `%% end {\"title\":\"Before Hooks\",\"category\":\"hook\",\"steps\":[{\"title\":\"browserContext.newPage\",\"category\":\"pw:api\"}]}`,
     `%% begin {\"title\":\"page.setContent\",\"category\":\"pw:api\"}`,
     `%% end {\"title\":\"page.setContent\",\"category\":\"pw:api\"}`,
-    `%% begin {\"title\":\"page.click\",\"category\":\"pw:api\"}`,
-    `%% end {\"title\":\"page.click\",\"category\":\"pw:api\",\"error\":{\"message\":\"page.click: Timeout 1ms exceeded.\\n=========================== logs ===========================\\nwaiting for selector \\\"input\\\"\\n============================================================\",\"stack\":\"<stack>\"}}`,
+    `%% begin {\"title\":\"page.click(input)\",\"category\":\"pw:api\"}`,
+    `%% end {\"title\":\"page.click(input)\",\"category\":\"pw:api\",\"error\":{\"message\":\"page.click(input): Timeout 1ms exceeded.\\n=========================== logs ===========================\\nwaiting for selector \\\"input\\\"\\n============================================================\",\"stack\":\"<stack>\"}}`,
```
dgozman, line 334: Where does the `input` comes from to the error message? I guess I missed that in the code, but it seems like a lot more of our tests should be rebased then.

dgozman (approved):

src/client/types.ts
pavelfeldman: the instrumentation end uses the fact that it is an envelope...

pavelfeldman: And more importantly, this documents it, right?

tests/playwright-test/reporter.spec.ts
pavelfeldman: It does not, fixed.

PR #8942, opened 2021-09-15
chore(create-playwright): enhance UX follow-ups and add more verbose Playwright Test config

mxschmitt:
- [x] Update JS examples + config

Relates #8892

### DIFF
--- a/packages/create-playwright/assets/.eslintrc.js
+++ b/packages/create-playwright/assets/.eslintrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+    "extends": "../../../.eslintrc.js",
+    "rules": {
+        "notice/notice": 0
+    }
+};
--- a/packages/create-playwright/assets/playwright.config.ts
+++ b/packages/create-playwright/assets/playwright.config.ts
@@ -3,8 +3,14 @@ import path from 'path';
 
 // Reference: https://playwright.dev/docs/test-configuration
 const config: PlaywrightTestConfig = {
+  // Timeout per test
   timeout: 30 * 1000,
+  // Test directory
   testDir: path.join(__dirname, '{{testDir}}'),
+  // If a test fails, retry it additional 2 times
+  retries: 2,
+  // Artifacts folder where screenshots, videos, and traces are stored.
+  outputDir: 'test-results/',
 
   // Run your local dev server before starting the tests:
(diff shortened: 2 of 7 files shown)


packages/create-playwright/assets/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
mxschmitt, line 1: drive-by: migrated instead of ignoring the assets I only disable the notice rule.

packages/create-playwright/assets/playwright.config.ts
```diff
@@ -3,8 +3,14 @@ import path from 'path';
 
 // Reference: https://playwright.dev/docs/test-configuration
 const config: PlaywrightTestConfig = {
+  // Timeout per test
   timeout: 30 * 1000,
+  // Test directory
   testDir: path.join(__dirname, '{{testDir}}'),
+  // If a test fails, retry it additional 2 times until it get marked as failing.
+  retries: 2,
+  // Artifacts folder where screenshots, videos, and traces are stored. Usually this gets uploaded on your CI artifacts store.
```
dgozman, line 12: Drop the "usually this gets ...." part.

packages/create-playwright/assets/example.spec.ts
```diff
@@ -1,7 +1,89 @@
-import { test, expect } from '@playwright/test';
+import { expect, Page, test as base } from '@playwright/test';
```
dgozman, line 1: - What happens to example.spec.js?
- I don't think this works as a "get started example". I'd rather make "1-get-started", "2-recipes", "3-pom" directories with examples or something like that.

mxschmitt: Will follow-up in more PRs dedicated to example test files.

packages/create-playwright/src/index.ts
```diff
@@ -171,6 +172,10 @@ We suggest that you begin by typing:
 
 ${colors.cyan(prefix + '  ' + commandToRunTests(this.packageManager))}
 
+And check out the following files:
+  - ./${pathToNavigate ? pathToNavigate + '/' : ''}${exampleSpecPath} - Example end-to-end test of a TODO application
```
dgozman, line 176: This is not a test of a TODO application anymore :)

dgozman (approved):

PR #8945, opened 2021-09-15
feat(fetch): introduce FetchRequest.dispose, fulfill with global fetch

yury-s:
#5999

### DIFF
--- a/docs/src/api/class-fetchrequest.md
+++ b/docs/src/api/class-fetchrequest.md
@@ -8,9 +8,9 @@ will be logged in and vice versa.
 
 ## async method: FetchRequest.dispose
 
-Disposes this request, it will release all responses received for this request. Call this method when you are done
-with the requests. If not called then all responses received for this object and not disposed via
-[`method: FetchResponse.dispose`] will stay in memory.
+All responses received through [`method: FetchRequest.fetch`], [`method: FetchRequest.get`], [`method: FetchRequest.post`]
+and other methods are stored in the memory, so that you can later call [`method: FetchResponse.body`]. This method
+discards all stored responses, and makes [`method: FetchResponse.body`] throw "Response disposed" error.
 
 ## async method: FetchRequest.fetch
 - returns: <[FetchResponse]>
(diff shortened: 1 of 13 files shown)


docs/src/api/class-fetchrequest.md
```diff
@@ -6,6 +6,12 @@ environment or the service to your e2e test. When used on [Page] or a [BrowserCo
 the cookies from the corresponding [BrowserContext]. This means that if you log in using this API, your e2e test
 will be logged in and vice versa.
 
+## async method: FetchRequest.dispose
+
+Disposes this request, it will release all responses received for this request. Call this method when you are done
```
dgozman, line 11: `All responses received through [fetch], [get], [post] and other methods are stored. in the memory, so that you can later call [FetchResponse.body]. This method discards all stored responses, and makes [FetchResponse.body] throw with "Response disposed" error.`

src/dispatchers/networkDispatchers.ts
```diff
@@ -173,6 +173,11 @@ export class FetchRequestDispatcher extends Dispatcher<FetchRequest, channels.Fe
     super(scope, request, 'FetchRequest', {}, true);
   }
 
+  async dispose(params?: channels.FetchRequestDisposeParams): Promise<void> {
+    this._object.dispose();
```
dgozman, line 177: I am a bit confused why are these two different?
- For example, I'd put `this._object.dispose()` directly into `_disposeDispatcher` - will that work?
- Alternatively, I'd only call `this._object.dispose()` everywhere, and emit a "disposed" event that dispatcher would listen to (similarly to context's "closed" event).

yury-s: > * For example, I'd put `this._object.dispose()` directly into `_disposeDispatcher` - will that work?

That works but it is indirect, i.e. FetchRequest is created by BrowserContext directly, owned by it but then disposed by an external dispatcher. This boils down to the same problem of separation between dispatchers and model objects, other than that disposing FetchRequest along with its dispatcher would work.

> * Alternatively, I'd only call `this._object.dispose()` everywhere, and emit a "disposed" event that dispatcher would listen to (similarly to context's "closed" event).

By everywhere you mean in FetchRequestDispatcher.dispose and BrowserContext._didCloseInternal ?

docs/src/api/class-fetchrequest.md
yury-s: Updated text.

dgozman (approved):

PR #8975, opened 2021-09-16
feat(fetch): support form data and json encodings

yury-s:
#5999
#8865

### DIFF
--- a/docs/src/api/class-fetchrequest.md
+++ b/docs/src/api/class-fetchrequest.md
@@ -39,9 +39,12 @@ If set changes the fetch method (e.g. PUT or POST). If not specified, GET method
 Allows to set HTTP headers.
 
 ### option: FetchRequest.fetch.data
-- `data` <[string]|[Buffer]>
+- `data` <[string]|[Buffer]|[Serializable]>
 
-Allows to set post data of the fetch.
+Allows to set post data of the fetch. If the data parameter is an object, it will be serialized the following way:
+* If `content-type` header is set to `application/x-www-form-urlencoded` the object will be serialized as html form using `application/x-www-form-urlencoded` encoding.
+* If `content-type` header is set to `multipart/form-data` the object will be serialized as html form using `multipart/form-data` encoding.
+* Otherwise the object will be serialized to json string and `content-type` header will be set to `application/json`.
 
 ### option: FetchRequest.fetch.timeout
 - `timeout` <[float]>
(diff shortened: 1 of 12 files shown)


src/client/fetch.ts
```diff
@@ -85,18 +89,89 @@ export class FetchRequest extends ChannelOwner<channels.FetchRequestChannel, cha
...
+                break;
+              }
+            }
+          }
+          if (contentType === 'application/json') {
+            const json = JSON.stringify(options.data);
+            postDataBuffer = Buffer.from(json, 'utf8');
+            if (!headersObj?.[contentTypeHeaderName]) {
+              headersObj ??= {};
+              headersObj[contentTypeHeaderName] = contentType;
```
dgozman, line 117: Can we move this logic to the backend?

src/dispatchers/networkDispatchers.ts
```diff
@@ -214,3 +216,12 @@ export class FetchRequestDispatcher extends Dispatcher<FetchRequest, channels.Fe
   }
 }
 
+function deserializeFormData(formData?: channels.FormField[]): FormField[] | undefined {
+  if (!formData)
+    return undefined;
+  for (const field of formData) {
+    if (field.file)
+      (field.file.buffer as any) = Buffer.from(field.file.buffer, 'base64');
```
dgozman, line 224: I'd suggest to create a new object here, just in case the old one is used somewhere in dispatcher/tracing/instrumentation.

src/client/fetch.ts
```diff
@@ -85,18 +89,89 @@ export class FetchRequest extends ChannelOwner<channels.FetchRequestChannel, cha
       const params = objectToArray(options.params);
       const method = options.method || request?.method();
       // Cannot call allHeaders() here as the request may be paused inside route handler.
-      const headersObj = options.headers || request?.headers() ;
+      let headersObj = options.headers || request?.headers() ;
+      let formData: channels.FormField[] | undefined;
+      let postDataBuffer: Buffer | undefined;
+      if (options.data) {
+        if (isString(options.data)) {
```
dgozman, line 96: Let's do some helper functions, so that it's easier to port to other clients.

src/client/fetch.ts
yury-s: Done

src/client/fetch.ts
yury-s: Done.

src/dispatchers/networkDispatchers.ts
yury-s: Removed this code.

docs/src/api/class-fetchrequest.md
```diff
@@ -108,9 +111,12 @@ Query parameters to be send with the URL.
 Allows to set HTTP headers.
 
 ### option: FetchRequest.post.data
-- `data` <[string]|[Buffer]>
+- `data` <[string]|[Buffer]|[Serializable]>
```
pavelfeldman, line 114: formData is a bit more complicated than that:
https://developer.mozilla.org/en-US/docs/Web/API/FormData/append

dgozman (approved):

PR #8979, opened 2021-09-17
feat(fetch): set content-length header if post data is present

yury-s:
#5999


dgozman (approved):

src/server/fetch.ts
```diff
@@ -136,7 +136,10 @@ export abstract class FetchRequest extends SdkObject {
         postData = params.formData ? serilizeFormData(params.formData, headers) : params.postData;
```
mxschmitt, line 136: nit: serializeFormData

yury-s: https://github.com/microsoft/playwright/pull/8991

PR #8988, opened 2021-09-17
feat(codegen): Add possibility to use custom codegen recorders

haikyuu:
Add possibility to use custom recorders for codegen

Relates #8992


JoelEinbinder:
Thanks for the PR. Unfortunately this kind of extensibility is outside the scope of this project. We don't want to commit to maintaining an api for custom recorders when our focus is on the languages we support ourselves.

Feel free to fork the codegen tool into your project.

PR #9005, opened 2021-09-18
chore: capture element of interest on the after snapshot

pavelfeldman:


src/server/injected/injectedScript.ts
```diff
@@ -105,7 +105,10 @@ export class InjectedScript {
       const result = this._querySelectorRecursively([{ element: root as Element, capture: undefined }], selector, 0, new Map());
       if (strict && result.length > 1)
         throw this.strictModeViolationError(selector, result.map(r => r.element));
-      return result[0]?.capture || result[0]?.element;
+      const element = result[0]?.capture || result[0]?.element;
+      if (element && selector.snapshotName)
+        element.setAttribute('__playwright_target__', selector.snapshotName);
```
dgozman, line 110: I am worried we'll mark multiple elements with the same snapshotName when polling retries. For example, waitForSelector(visible) can find an element, but retry because it's hidden and later resolve to a different one.

dgozman (approved):

PR #9021, opened 2021-09-20
fix(codegen): generate Asssert.AreEqual instead of Equals for csharp

avodovnik:
Using `Assert.Equals` throws an Exception in NUnit, which is our default.


mxschmitt:
Looks like this was already implemented in https://github.com/microsoft/playwright/pull/8859.

We could still fix it in the other places wdyt?

image

PR #9036, opened 2021-09-20
test: improve child process utilities in tests

dgozman:


tests/playwright-test/playwright-test-fixtures.ts
```diff
@@ -46,6 +47,76 @@ type Files = { [key: string]: string | Buffer };
...
+    this.process = spawn(params.command[0], params.command.slice(1), {
+      env: {
+        ...process.env,
+        ...params.env,
+      } as any,
+      cwd: params.cwd,
+      shell: params.shell,
+    });
+    if (process.env.PW_RUNNER_DEBUG)
+      process.stdout.write(`\n\nLaunching ${params.command.join(' ')}\n`);
```
yury-s, line 75: Shouldn't it write to stderr?

yury-s (approved):

tests/playwright-test/playwright-test-fixtures.ts
dgozman: Well, it's not an error but a debug log, so I'd say stdout is right.

PR #9042, opened 2021-09-21
docs: Rename playwright to @playwright/test

### LINKED ISSUE #9035: [BUG] Docs - Mismatched Typings `playwright` and `@playwright/test`
**Context:**
- Playwright Version: 1.14
- Operating System: Mac
- Node.js version: 14
- Browser: N/A

**Code Snippet**
The example listed in https://playwright.dev/docs/test-pom

```ts
// playwright-dev-page.ts
import type { Page } from 'playwright';
export class PlaywrightDevPage {  
    readonly page: Page;
```

```ts
// example.spec.ts
import { PlaywrightDevPage } from './playwright-dev-page';
test('Get Started table of contents', async ({ page }) => {  
    const playwrightDev = new PlaywrightDevPage(page);
```

**Describe the bug**

This causes issue with Typescript as the `Page` Type is from two different package.

Which is the correct package to use, `'playwright'` or `'@playwright/test'`?
I believe it is `'@playwright/test'`, I can raise a PR fixing the issue in docs.
It's present in multiple places.

sidharthv96:
Fixes #9035

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -23,7 +23,7 @@ Help us help you! Put down a short code snippet that illustrates your bug and
 that we can run and debug locally. For example:
 
 ```javascript
-const {chromium, webkit, firefox} = require('playwright');
+const {chromium, webkit, firefox} = require('@playwright/test');
 
 (async () => {
   const browser = await chromium.launch();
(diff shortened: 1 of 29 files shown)


mxschmitt:
@sidharthv96 we have two NPM packages which you can consume.

a) `playwright` - [the library](https://playwright.dev/docs/next/library). You can use it with different test-runners.
b) `@playwright/test` - [the test-runner](https://playwright.dev/docs/next/intro). You don't have to additionally instal `playwright` there.

I went over the changes and it seems like we don't have to change anything. In my comment before I meant that in the test-runner docs we should change it to playwright, but we did it already in all the places so we are golden there.

sidharthv96:
Oops, sorry for the mixup.

PR #9053, opened 2021-09-21
chore: move Frame.waitForTimeout to the server side

### LINKED ISSUE #9045: [BUG] waitForTimeout does not get shown in trace-viewer
```ts
import { test, expect } from '@playwright/test'

test('B', async ({ page }) => {
  await page.waitForTimeout(1234)
})

````

mxschmitt:
Fixes #9045


pavelfeldman (approved):
Ok to land because ports don't need to switch to it.

PR #9059, opened 2021-09-21
fix: make Response.json() return type any

### LINKED ISSUE #8976: [BUG] Response.json() should be unknown instead of Serializable
const data = resp.json() 
console.log(data.foo) // does not work

mxschmitt:
Making the return type more lax like Request.postDataJSON().

Fixes #8976

### DIFF
--- a/docs/src/api/class-request.md
+++ b/docs/src/api/class-request.md
@@ -102,7 +102,7 @@ Request's post body in a binary form, if any.
 
 ## method: Request.postDataJSON
 * langs: js, python
-- returns: <[null]|[any]>
+- returns: <[null]|[Serializable]>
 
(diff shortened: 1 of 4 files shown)


docs/src/api/class-response.md
```diff
@@ -58,7 +58,7 @@ Name of the header.
 
 ## async method: Response.json
 * langs: js, python
-- returns: <[Serializable]>
```
yury-s, line 61: Would it make sense to change `export type Serializable = {};` to `export type Serializable = any;` ?

JoelEinbinder: Yes I agree. I don't want to lose the concept of Serializable in the docs.

JoelEinbinder (approved):

yury-s (approved):

PR #9061, opened 2021-09-22
feat(fetch): support options in playwright._newRequest

yury-s:
#8865
#5999


docs/src/api/class-playwright.md
```diff
@@ -83,6 +83,57 @@ class PlaywrightExample
 }
 ```
 
+## async method: Playwright._newRequest
+* langs: js
+- returns: <[FetchRequest]>
+
+**experimental** Creates new instances of [FetchRequest].
+
+### option: Playwright._newRequest.useragent
```
dgozman, line 92: You can use references here to avoid copying the docs. See `docs/src/test-api/class-testoptions.md:126` for example.

docs/src/api/class-playwright.md
```diff
@@ -83,6 +83,57 @@ class PlaywrightExample
...
+### option: Playwright._newRequest.timeout
+- `timeout` <[float]>
+
+Maximum time in milliseconds to wait for the response. Defaults to
+`30000` (30 seconds). Pass `0` to disable timeout.
+
+### option: Playwright._newRequest.ignoreHTTPSErrors
+- `ignoreHTTPSErrors` <[boolean]>
+
+Whether to ignore HTTPS errors during navigation. Defaults to `false`.
```
dgozman, line 128: I think the place you've copied this form should be updated. It's not about navigation, but network requests.

utils/generate_types/index.js
```diff
@@ -463,7 +476,7 @@ class TypesGenerator {
 
   const apiDocumentation = parseApi(path.join(PROJECT_DIR, 'docs', 'src', 'api'));
   // Root module types are overridden.
-  apiDocumentation.classesArray = apiDocumentation.classesArray.filter(cls => cls.name !== 'Playwright');
+  // apiDocumentation.classesArray = apiDocumentation.classesArray.filter(cls => cls.name === 'Playwright');
```
dgozman, line 479: Please either remove or undo.

utils/generate_types/overrides.d.ts
```diff
@@ -343,12 +342,8 @@ export type AndroidKey =
   'Copy' |
   'Paste';
 
-export const chromium: BrowserType;
-export const firefox: BrowserType;
-export const webkit: BrowserType;
 export const _electron: Electron;
```
dgozman, line 345: I guess we can follow up with _electron and _android to follow _newRequest?

dgozman (approved):

utils/generate_types/overrides.d.ts
yury-s: Not sure if we want to expose them in the docs, do we?

utils/generate_types/index.js
yury-s: Removed.

docs/src/api/class-playwright.md
yury-s: So far they are only used in one place, I'll extract them into parmams.md if we decide to pass them to individual requests.

docs/src/api/class-playwright.md
yury-s: https://github.com/microsoft/playwright/pull/9081

PR #9062, opened 2021-09-22
chore: server-side expect impl harness

pavelfeldman:


src/server/progress.ts
```diff
@@ -70,10 +69,18 @@ export class ProgressController {
...
+      logEntry: entry => {
+        if ('message' in entry) {
+          const message = entry.message!;
+          if (this._state === 'running')
+            this.metadata.log.push(message);
+          // Note: we might be sending logs after progress has finished, for example browser logs.
+          this.instrumentation.onCallLog(this._logName, message, this.sdkObject, this.metadata);
+        }
+        if ('value' in entry)
+          this.metadata.lastActualValue = entry.value;
```
pavelfeldman, line 83: I might need to serialize/parse value here as well at some point.

src/protocol/serializers.ts
```diff
@@ -29,18 +29,24 @@ export function parseError(error: SerializedError): Error {
...
+  } else {
+    e = new Error(error.error.message);
+    e.name = error.error.name;
   }
-  const e = new Error(error.error.message);
   e.stack = error.error.stack || '';
-  e.name = error.error.name;
+
+  if (error.actualValue) {
+    // Both error and value means we have an expect error.
```
dgozman, line 43: Comment seems out-of-date.

src/protocol/serializers.ts
```diff
@@ -82,12 +88,14 @@ export function parseSerializedValue(value: SerializedValue, handles: any[] | un
...
 
 export type HandleOrValue = { h: number } | { fallThrough: any };
-export function serializeValue(value: any, handleSerializer: (value: any) => HandleOrValue, visited: Set<any>): SerializedValue {
-  const handle = handleSerializer(value);
-  if ('fallThrough' in handle)
-    value = handle.fallThrough;
-  else
-    return handle;
+export function serializeValue(value: any, handleSerializer?: (value: any) => HandleOrValue, visited: Set<any> = new Set()): SerializedValue {
+  if (handleSerializer && visited) {
```
dgozman, line 92: `visited` is always initialized here.

src/server/common/injectedTypes.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
dgozman, line 1: We also have `common/domErrors.ts`. Let's merge?

src/server/dom.ts
```diff
@@ -1066,18 +1067,22 @@ export function getAttributeTask(selector: SelectorInfo, name: string): Schedula
...
       const element = injected.querySelector(parsed, document, strict);
       if (!element)
         return continuePolling;
       progress.log(`  selector resolved to ${injected.previewNode(element)}`);
       if (element.nodeName !== 'INPUT' && element.nodeName !== 'TEXTAREA' && element.nodeName !== 'SELECT')
         return 'error:hasnovalue';
-      return (element as any).value;
+      const value = (element as any).value;
+      progress.setLastValue(value);
+      if (expected && !injected.matchesText(value, expected as any))
```
dgozman, line 1081: Shouldn't this be `expected.value as any`?

src/server/dom.ts
```diff
@@ -1066,18 +1067,22 @@ export function getAttributeTask(selector: SelectorInfo, name: string): Schedula
   }, { parsed: selector.parsed, strict: selector.strict, name });
 }
 
-export function inputValueTask(selector: SelectorInfo): SchedulableTask<string> {
-  return injectedScript => injectedScript.evaluateHandle((injected, { parsed, strict }) => {
+export function inputValueTask(selector: SelectorInfo, expected: channels.ExpectedTextValue | undefined): SchedulableTask<string> {
```
dgozman, line 1070: I do not see where we parse the SerializedValue inside expected...

pavelfeldman: Good point, I was hoping we can save on double serialization between protocol and injected script, but now I think it isn't a great idea.

src/server/dom.ts
```diff
@@ -1093,3 +1099,25 @@ export function elementStateTask(selector: SelectorInfo, state: ElementStateWith
 }
 
 export const kUnableToAdoptErrorMessage = 'Unable to adopt element handle from a different document';
+
+class TextMatcher {
```
dgozman, line 1103: I think this should move to injected to be callable from inside evaluates.

dgozman (approved):

PR #9069, opened 2021-09-22
docs: mention empty string in `userDataDir`

szmarczak:
https://github.com/microsoft/playwright/blob/c4eb2d4b1e17a0de1cea7220be45545c7c524614/src/server/browserType.ts#L146-L149

### DIFF
--- a/docs/src/api/class-browsertype.md
+++ b/docs/src/api/class-browsertype.md
@@ -244,8 +244,8 @@ this context will automatically close the browser.
 Path to a User Data Directory, which stores browser session data like cookies and local storage. More details for
 [Chromium](https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md#introduction) and
 [Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#User_Profile).
-Note that Chromium's user data directory is the **parent** directory of the "Profile Path" seen at `chrome://version`.
-Use an empty string to generate and destroy it automatically.
+Note that Chromium's user data directory is the **parent** directory of the "Profile Path" seen at `chrome://version`. Use an empty string to
+generate and destroy it automatically.
 
 ### option: BrowserType.launchPersistentContext.-inline- = %%-shared-browser-options-list-%%
 ### option: BrowserType.launchPersistentContext.slowMo = %%-browser-option-slowmo-%%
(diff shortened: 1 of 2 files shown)


docs/src/api/class-browsertype.md
```diff
@@ -244,7 +244,8 @@ this context will automatically close the browser.
 Path to a User Data Directory, which stores browser session data like cookies and local storage. More details for
 [Chromium](https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md#introduction) and
 [Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#User_Profile).
-Note that Chromium's user data directory is the **parent** directory of the "Profile Path" seen at `chrome://version`.
+Note that Chromium's user data directory is the **parent** directory of the "Profile Path" seen at `chrome://version`. Use an empty string to
+generate and destroy it automatically.
```
szmarczak, line 248:
```suggestion
set a temporary directory instead.
```

szmarczak, line 248:
```suggestion
Note that Chromium's user data directory is the **parent** directory of the "Profile Path" seen at `chrome://version`. Pass an empty string to
use a temporary directory instead.
```

types/types.d.ts
```diff
@@ -9656,7 +9656,8 @@ export interface BrowserType<Unused = {}> {
    * context will automatically close the browser.
    * @param userDataDir Path to a User Data Directory, which stores browser session data like cookies and local storage. More details for [Chromium](https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md#introduction) and
    * [Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Command_Line_Options#User_Profile). Note that Chromium's user
-   * data directory is the **parent** directory of the "Profile Path" seen at `chrome://version`.
+   * data directory is the **parent** directory of the "Profile Path" seen at `chrome://version`. Use an empty string to
+   * generate and destroy it automatically.
```
szmarczak, line 9660:
```suggestion
   * data directory is the **parent** directory of the "Profile Path" seen at `chrome://version`. Pass an empty string to
   * use a temporary directory instead.
```

JoelEinbinder (approved):

dgozman (approved):
Thank you for the PR!

PR #9079, opened 2021-09-22
chore(monorepo): utility to list packages in order of interdependence

JoelEinbinder:
Lists all the packages inside the `packages` directory in the order which
they depend on eachother. This is the order we would want to publish or install
these packages. Basically making sure that playwright-core is first.

Right now this only lists create-playwright, because thats the only package in that folder now.

### DIFF
--- a/utils/list_packages.js
+++ b/utils/list_packages.js
@@ -52,7 +52,7 @@ module.exports = {packages, packageNameToPath};
 /**
  * @param {string} packagePath
  */
-function * internalDependencies(packagePath) {
+function* internalDependencies(packagePath) {
   yield packagePath;
   for (const dependency of Object.keys(packagePathToJSON.get(packagePath).dependencies || [])) {
     const dependencyPath = packageNameToPath.get(dependency);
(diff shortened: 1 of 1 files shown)


utils/list_packages.js
```diff
@@ -0,0 +1,62 @@
...
+    return -1;
+  return 0;
+});
+
+module.exports = {packages, packageNameToPath};
+
+/**
+ * @param {string} packagePath
+ */
+function * internalDependencies(packagePath) {
```
yury-s, line 55: nit: using just package name as a parameter here would simplify the code a bit, we are not going to have a nested package structure anyway

yury-s (approved):

utils/list_packages.js
JoelEinbinder: The package name does not match the package path even without nested packages because of the package name name `@playwright/test` vs the directory name `playwright-test`.

yury-s: I guess what I was trying to say is that we could sort package names without name->path->json transition by using just name->json map and then convert list of names to list of paths if necessary as a separate pass.

JoelEinbinder: I tried that but it's not any smaller. https://gist.github.com/JoelEinbinder/a82496a327ca3c0f057a28ab00ed6ecd

PR #9087, opened 2021-09-22
chore(typescript): bump typescript to v4.4.3

JoelEinbinder:
Healthy roll to v4.4.3. New typescript treats caught variables as `unknown` instead of `any`, but it can be turned
off in the tsconfig.


tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "resolveJsonModule": true,
     "noEmit": true,
     "noImplicitOverride": true,
+    "useUnknownInCatchVariables": false
```
mxschmitt, line 13:
```suggestion
    "useUnknownInCatchVariables": false,
```

mxschmitt (approved):

Issue #7537, opened 2021-07-09
Debug mode for test runner


pavelfeldman:
### Feature request

Requested on Twitter: https://mobile.twitter.com/eliostruyf/status/1413392602418790401

```
The ideal would be a test runner
that shows the test progress, and
a separate browser instance.
```

- [x] My understanding is that this is a new mode

```bash
npx playwright test --debug
```

Which is a shortcut to

```bash
PWDEBUG=1 npx playwright test --workers=1 --headed
```

But in addition to that,

- [x] inspector and browser window are reused to reduce flashing
- [ ] elements of the new test status UI show the history of successful/failed runs in the inspector

This basically produces debug experience visually resembling Cypress for those relying only on core testing scenarios.

visortelle:
This feature done right can allow to build [storybook](https://storybook.js.org/)-like components catalogs.

There is no need to setup a separate storybook/styleguidist/etc. to display a UI components list with usage examples if you already have a list tests for each component and can run it in browser just by clicking a right test case.

mxschmitt:
This got partially implemented in https://github.com/microsoft/playwright/pull/8938

PR #9115, opened 2021-09-23
feat(test-runner): re-use context and page when debug mode is used

mxschmitt:
#7537

### DIFF
--- a/src/test/cli.ts
+++ b/src/test/cli.ts
@@ -16,7 +16,7 @@
 
 /* eslint-disable no-console */
 
-import { Command } from 'commander';
+import { Command, Option } from 'commander';
 import fs from 'fs';
 import path from 'path';
 import type { Config } from './types';
@@ -45,7 +45,7 @@ export function addTestCommand(program: Command) {
   command.option('--browser <browser>', `Browser to use for tests, one of "all", "chromium", "firefox" or "webkit" (default: "chromium")`);
   command.option('--headed', `Run tests in headed browsers (default: headless)`);
   command.option('--debug', `Run tests with Playwright Inspector. Shortcut for "PWDEBUG=1" environment variable and "--timeout=0 --maxFailures=1 --headed --workers=1" options`);
-  command.option('--reuse-context', `Re-uses the context, useful for running multiple tests after each-other.`);
+  command.addOption(new Option('--reuse-context').hideHelp());
   command.option('-c, --config <file>', `Configuration file, or a test directory with optional "${tsConfig}"/"${jsConfig}"`);
   command.option('--forbid-only', `Fail if test.only is called (default: false)`);
   command.option('-g, --grep <grep>', `Only run tests matching this regular expression (default: ".*")`);
--- a/src/test/cli.ts
+++ b/src/test/cli.ts
@@ -45,6 +45,7 @@ export function addTestCommand(program: Command) {
   command.option('--browser <browser>', `Browser to use for tests, one of "all", "chromium", "firefox" or "webkit" (default: "chromium")`);
   command.option('--headed', `Run tests in headed browsers (default: headless)`);
   command.option('--debug', `Run tests with Playwright Inspector. Shortcut for "PWDEBUG=1" environment variable and "--timeout=0 --maxFailures=1 --headed --workers=1" options`);
+  command.option('--reuse-context', `Re-uses the context, useful for running multiple tests after each-other.`);
   command.option('-c, --config <file>', `Configuration file, or a test directory with optional "${tsConfig}"/"${jsConfig}"`);
   command.option('--forbid-only', `Fail if test.only is called (default: false)`);
   command.option('-g, --grep <grep>', `Only run tests matching this regular expression (default: ".*")`);
@@ -96,15 +97,18 @@ async function createLoader(opts: { [key: string]: any }): Promise<Loader> {
   }
 
   const overrides = overridesFromOptions(opts);
-  if (opts.headed || opts.debug)
+  if (opts.headed || opts.debug || opts.reuseContext)
     overrides.use = { headless: false };
-  if (opts.debug) {
+  if (opts.debug || opts.reuseContext) {
     overrides.maxFailures = 1;
     overrides.timeout = 0;
     overrides.workers = 1;
-    process.env.PWDEBUG = '1';
-    process.env.PWTEST_REUSE_CONTEXT = '1';
   }
+  if (opts.debug)
+    process.env.PWDEBUG = '1';
+  if (opts.reuseContext)
+    process.env.PWTEST_REUSE_CONTEXT = '1';
+
   const loader = new Loader(defaultConfig, overrides);
 
   async function loadConfig(configFile: string) {
--- a/src/test/cli.ts
+++ b/src/test/cli.ts
@@ -103,6 +103,7 @@ async function createLoader(opts: { [key: string]: any }): Promise<Loader> {
     overrides.timeout = 0;
     overrides.workers = 1;
     process.env.PWDEBUG = '1';
+    process.env.PWTEST_REUSE_CONTEXT = '1';
   }
   const loader = new Loader(defaultConfig, overrides);
 
--- a/src/test/index.ts
+++ b/src/test/index.ts
@@ -37,6 +37,7 @@ type WorkerAndFileFixtures = PlaywrightWorkerArgs & PlaywrightWorkerOptions & {
 class ReuseBrowerContextStorage {
   private _browserContext?: BrowserContext;
   private _uniqueOrigins = new Set<string>();
+  private _options?: BrowserContextOptions;
 
   isEnabled(): boolean {
     return !!process.env.PWTEST_REUSE_CONTEXT;
@@ -49,6 +50,7 @@ class ReuseBrowerContextStorage {
   setContext(context: BrowserContext) {
     assert(!this._browserContext);
     this._browserContext = context;
+    this._options = (context as any)._options;
     this._browserContext.on('page', page => page.on('framenavigated', frame => this._uniqueOrigins.add(new URL(frame.url()).origin)));
   }
 
@@ -65,15 +67,18 @@ class ReuseBrowerContextStorage {
         pageWorker = await this._browserContext.newPage();
         pages = this._browserContext.pages();
       }
+      await Promise.all(pages.slice(2).map(page => page.close()));
+      await pageWorker.route(url => this._uniqueOrigins.has(url.origin), route => route.fulfill({body: `<html></html>`, contentType: 'text/html'}));
       for (const origin of this._uniqueOrigins) {
-        await pageWorker.route(origin, route => route.fulfill({body: `<html></html>`, contentType: 'text/html'}), { times: 1 });
         await pageWorker.goto(origin);
         await pageWorker.evaluate(() => window.localStorage.clear());
         await pageWorker.evaluate(() => window.sessionStorage.clear());
       }
+      await pageWorker.close();
+    } else {
+      await Promise.all(pages.slice(1).map(page => page.close()));
     }
-    for (const p of pages.slice(1))
-      await p.close();
+
     await this._browserContext.clearCookies();
     await this._applyNewContextOptions(page, newContextOptions);
     await page.goto('about:blank');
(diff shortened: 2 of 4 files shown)


src/test/index.ts
```diff
@@ -309,10 +344,17 @@ export const test = _baseTest.extend<TestFixtures, WorkerAndFileFixtures>({
     }));
   }, { auto: true }],
 
-  context: async ({ browser, video, _artifactsDir }, use, testInfo) => {
+  _reuseBrowerContext: [new ReuseBrowerContextStorage(), { scope: 'worker' }],
```
mxschmitt, line 347: maybe move it out of the fixtures to not pollute the fixtures namespace.

src/test/cli.ts
```diff
@@ -105,6 +105,7 @@ async function createLoader(opts: { [key: string]: any }): Promise<Loader> {
     overrides.timeout = 0;
     overrides.workers = 1;
     process.env.PWDEBUG = '1';
+    process.env.PWTEST_REUSE_CONTEXT = '1';
```
pavelfeldman, line 108: Not sure reuse === debug atm, how about `PWTEST_REUSE` for now.

src/test/index.ts
```diff
@@ -330,8 +415,14 @@ export const test = _baseTest.extend<TestFixtures, WorkerAndFileFixtures>({
     const allPages: Page[] = [];
     context.on('page', page => allPages.push(page));
 
+    if (_reuseBrowerContext.isEnabled())
+      _reuseBrowerContext.setContext(context);
+
     await use(context);
 
+    if (_reuseBrowerContext.isEnabled())
+      return;
```
pavelfeldman, line 424: We are not turning off video / tracing here, looks like the logic is off.

src/test/index.ts
```diff
@@ -364,7 +455,11 @@ export const test = _baseTest.extend<TestFixtures, WorkerAndFileFixtures>({
     }
   },
 
-  page: async ({ context }, use) => {
+  page: async ({ context, _reuseBrowerContext }, use) => {
+    if (_reuseBrowerContext.isEnabled() && _reuseBrowerContext.hasContext()) {
```
pavelfeldman, line 459: If _reuseBrowerContext.isEnabled() is true and given that we depend on context fixture, how can this be that hasContext is false?

src/test/index.ts
```diff
@@ -31,8 +31,86 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+      let pageWorker = null;
+      if (pages.length > 1) {
+        await page.bringToFront();
+        pageWorker = pages[1];
+      } else {
+        pageWorker = await this._browserContext.newPage();
+        pages = this._browserContext.pages();
+      }
+      for (const origin of this._uniqueOrigins) {
+        await pageWorker.route(origin, route => route.fulfill({body: `<html></html>`, contentType: 'text/html'}), { times: 1 });
```
pavelfeldman, line 69: You don't need to route each origin, just route everything once. Also, don't forget to remove the route, otherwise you keep adding listeners. Also revisit your route(times) implementation to remove listener after `times` expires. Otherwise we are leaking memory, that's probably for 1.15.1-2.

src/test/index.ts
```diff
@@ -31,8 +31,86 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+        await pageWorker.goto(origin);
+        await pageWorker.evaluate(() => window.localStorage.clear());
+        await pageWorker.evaluate(() => window.sessionStorage.clear());
+      }
+    }
+    for (const p of pages.slice(1))
```
pavelfeldman, line 75: You should have done it earlier - all these pages are polluting localStorage while they are alive. You clear it above, but it is not clean again.

src/test/index.ts
```diff
@@ -31,8 +31,86 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+    const emulateMediaOptions: Partial<Parameters<Page['emulateMedia']>[0]> = {};
+    if (oldOptions.colorScheme !== newOptions.colorScheme)
+      emulateMediaOptions.colorScheme = newOptions.colorScheme;
+    if (oldOptions.forcedColors !== newOptions.forcedColors)
+      emulateMediaOptions.forcedColors = newOptions.forcedColors;
+    if (oldOptions.reducedMotion !== newOptions.reducedMotion)
+      emulateMediaOptions.reducedMotion = newOptions.reducedMotion;
+    if (Object.keys(emulateMediaOptions).length > 0)
+      await page.emulateMedia(emulateMediaOptions);
+    (this._browserContext as any)._options = newOptions;
```
pavelfeldman, line 103: This does not really do nothing, only makes Playwright break inside. You should remember settings for your context and only reuse the context if the settings match.

pavelfeldman:
I think the direction is good, we just need to be a bit more careful in the implementation.

src/test/index.ts
mxschmitt: simplified that logic.

src/test/index.ts
mxschmitt: I changed the impl, since in this case we throw the worker page away, its not a big deal if we leak. (but yeah will create an issue for cleaning up if times exceed. Good catch).

src/test/index.ts
mxschmitt: done.

src/test/index.ts
```diff
@@ -31,8 +31,91 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+      } else {
+        pageWorker = await this._browserContext.newPage();
+        pages = this._browserContext.pages();
+      }
+      await Promise.all(pages.slice(2).map(page => page.close()));
+      await pageWorker.route(url => this._uniqueOrigins.has(url.origin), route => route.fulfill({body: `<html></html>`, contentType: 'text/html'}));
+      for (const origin of this._uniqueOrigins) {
+        await pageWorker.goto(origin);
+        await pageWorker.evaluate(() => window.localStorage.clear());
+        await pageWorker.evaluate(() => window.sessionStorage.clear());
```
mxschmitt, line 75: I guess since we navigate anyway we can drop this line.

src/test/cli.ts
```diff
@@ -45,6 +45,7 @@ export function addTestCommand(program: commander.CommanderStatic) {
   command.option('--browser <browser>', `Browser to use for tests, one of "all", "chromium", "firefox" or "webkit" (default: "chromium")`);
   command.option('--headed', `Run tests in headed browsers (default: headless)`);
   command.option('--debug', `Run tests with Playwright Inspector. Shortcut for "PWDEBUG=1" environment variable and "--timeout=0 --maxFailures=1 --headed --workers=1" options`);
+  command.option('--reuse-context', `Re-uses the context, useful for running multiple tests after each-other.`);
```
pavelfeldman, line 48: Let's make it hidden

src/test/index.ts
pavelfeldman: We can't - session storage is preserved across navigations.

src/test/index.ts
```diff
@@ -309,10 +392,23 @@ export const test = _baseTest.extend<TestFixtures, WorkerAndFileFixtures>({
...
 
-  context: async ({ browser, video, _artifactsDir }, use, testInfo) => {
+  _reuseBrowerContext: [new ReuseBrowerContextStorage(), { scope: 'worker' }],
+
+  context: async ({ browser, video, _artifactsDir, _reuseBrowerContext, _combinedContextOptions }, use, testInfo) => {
     const hook = hookType(testInfo);
     if (hook)
       throw new Error(`"context" and "page" fixtures are not supported in ${hook}. Use browser.newContext() instead.`);
+    if (_reuseBrowerContext.isEnabled()) {
+      let context: BrowserContext;
```
pavelfeldman, line 402: Let your utility handle the hasContext branching. This removes `hasContext` & `setContext`.

`const context = await _reuseBrowerContext.obtainContext(browser, _combinedContextOptions);`

src/test/index.ts
```diff
@@ -31,8 +31,91 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+      let pageWorker = null;
+      if (pages.length > 1) {
+        await page.bringToFront();
+        pageWorker = pages[1];
+      } else {
+        pageWorker = await this._browserContext.newPage();
+        pages = this._browserContext.pages();
+      }
+      await Promise.all(pages.slice(2).map(page => page.close()));
+      await pageWorker.route(url => this._uniqueOrigins.has(url.origin), route => route.fulfill({body: `<html></html>`, contentType: 'text/html'}));
```
pavelfeldman, line 71: You are still leaking listeners here.

src/test/index.ts
```diff
@@ -31,8 +31,91 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+      for (const origin of this._uniqueOrigins) {
+        await pageWorker.goto(origin);
```
pavelfeldman, line 73: It is most likely that you are already there. You should not navigate at all for the basic single-origin case.

src/test/index.ts
```diff
@@ -31,8 +31,91 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+        await pageWorker.evaluate(() => window.localStorage.clear());
+        await pageWorker.evaluate(() => window.sessionStorage.clear());
+      }
+      await pageWorker.close();
```
pavelfeldman, line 77: Oh no, you just closed our _only_ page?

src/test/index.ts
```diff
@@ -31,8 +31,91 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+
+  async obtainContext(newContextOptions: BrowserContextOptions): Promise<BrowserContext> {
+    assert(this._browserContext);
+    let pages = this._browserContext.pages();
+    const page = pages[0];
+    if (this._uniqueOrigins.size > 0) {
+      let pageWorker = null;
+      if (pages.length > 1) {
+        await page.bringToFront();
+        pageWorker = pages[1];
```
pavelfeldman, line 65: We should only use one page.

src/test/index.ts
```diff
@@ -31,8 +31,91 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+      await page.setViewportSize({width: newOptions.viewport?.width, height: newOptions.viewport?.height });
+    const emulateMediaOptions: Partial<Parameters<Page['emulateMedia']>[0]> = {};
+    if (this._options.colorScheme !== newOptions.colorScheme)
+      emulateMediaOptions.colorScheme = newOptions.colorScheme;
+    if (this._options.forcedColors !== newOptions.forcedColors)
+      emulateMediaOptions.forcedColors = newOptions.forcedColors;
+    if (this._options.reducedMotion !== newOptions.reducedMotion)
+      emulateMediaOptions.reducedMotion = newOptions.reducedMotion;
+    if (Object.keys(emulateMediaOptions).length > 0)
+      await page.emulateMedia(emulateMediaOptions);
```
pavelfeldman, line 107: This is a lot of code for a seemingly random property. Just remove it. Keeping viewport is probably ok.

src/test/index.ts
mxschmitt: Fixed in https://github.com/microsoft/playwright/pull/9234

src/test/index.ts
mxschmitt: My idea was to:
1. Close all pages > 2
2. focus the first page
3. use the second page as a worker page
4. on the worker page, I navigate to the hosts etc. and clear the storage

---

but for simplicity I changed it now back to use the pages[0] for that.

src/test/index.ts
```diff
@@ -31,8 +32,89 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+      await page.goto(nextOrigin);
+      await page.evaluate(() => window.localStorage.clear());
+      await page.evaluate(() => window.sessionStorage.clear());
+      await page.unroute(nextOrigin);
+    }
+    await Promise.all(pages.slice(1).map(page => page.close()));
+    await this._browserContext.clearCookies();
+    await this._applyNewContextOptions(page, newContextOptions);
+    await page.goto('about:blank');
+    this._pauseNavigationEventCollection = false;
```
pavelfeldman, line 84: you would typically try/finally those

src/test/index.ts
```diff
@@ -31,8 +32,89 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+    }));
+    return this._browserContext;
+  }
+
+  async _refurbishExistingContext(newContextOptions: BrowserContextOptions): Promise<BrowserContext> {
+    assert(this._browserContext);
+    const pages = this._browserContext.pages();
+    const page = pages[0];
+    this._pauseNavigationEventCollection = true;
+    const initialOrigin = new URL(page.url()).origin;
```
pavelfeldman, line 70: Can't page be null here?

src/test/index.ts
```diff
@@ -31,8 +32,89 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+    while (this._uniqueOrigins.size > 0) {
+      const nextOrigin = this._uniqueOrigins.has(initialOrigin) ? initialOrigin : this._uniqueOrigins.values().next().value;
+      this._uniqueOrigins.delete(nextOrigin);
+      await page.route(nextOrigin, route => route.fulfill({ body: `<html></html>`, contentType: 'text/html' }));
```
pavelfeldman, line 74: Why do you insist on routing in the loop?

pavelfeldman, line 74: Can we route origins? I would install one route and point all documents to empty html

src/test/index.ts
pavelfeldman: ping?

src/test/index.ts
mxschmitt: shall we handle the case when the user closes the default page?

src/test/index.ts
mxschmitt: ack: Previously I had a single route. Let me get back to that again.

mxschmitt: What do you mean with that? temporary **/* is probably the best if this is what you mean. No need to manually filter for origins there.

src/test/index.ts
mxschmitt: I thought its good to have, not much code and seems to work fine?

src/test/index.ts
```diff
@@ -31,8 +32,83 @@ type TestFixtures = PlaywrightTestArgs & PlaywrightTestOptions & {
...
+        this._uniqueOrigins.delete(nextOrigin);
+        await page.goto(nextOrigin);
+        await page.evaluate(() => window.localStorage.clear());
+        await page.evaluate(() => window.sessionStorage.clear());
+      }
+      await page.unroute('**/*');
+      await Promise.all(pages.slice(1).map(page => page.close()));
+      await this._browserContext.clearCookies();
+      await this._applyNewContextOptions(page, newContextOptions);
+      await page.goto('about:blank');
```
pavelfeldman, line 84: this needs to go above clearCookies.

pavelfeldman (approved):

src/test/index.ts
mxschmitt: good catch!

PR #9120, opened 2021-09-24
fix(webkit): support Cross-Origin-Opener-Policy navigation

### LINKED ISSUE #9065: [BUG] playwright 1.15.0 causes page.goto to throw with 'Frame load interrupted' in webkit
**Context:**
- Playwright Version: 1.15.0
- Operating System: Mac and Linux
- Node.js version: 16.9.1
- Browser: Webkit
- Extra: Our app uses service workers; that may or may not be relevant

**Code Snippet**
```javascript
const { webkit } = require('playwright');

(async () => {
  const browser = await webkit.launch();
  await page.goto('http://localhost:4000');
  // The previous line throws with
  // 'page.goto: Frame load interrupted'
})();
```

With playwright 1.14.1, this example worked fine. With playwright 1.15.0, `page.goto` rejects with 'page.goto: Frame load interrupted' only on webkit. Firefox and chromium work fine, as does the site when I load it in Safari 15.0 by hand.

Our app does use a service worker which has caused other issues with playwright in the past. I don't know if it's relevant in this case.

yury-s:
WebKit recently [enabled](https://trac.webkit.org/changeset/282007/webkit) support for  Cross-Origin-Opener-Policy / Cross-Origin-Embedder-Policy and decision about whether navigation should happen in a new process now made twice:
* navigation policy decision
* response policy decision

In the latter case the requestWillBeSent and responseReceivedEvents are fired in the original process and then (if process should be swapped) fired again in the new process (with new request id but same loaderId). Moreover, after the response policy decision the old process will send Page.frameStoppedLoading and Network.loadingFailed(canceled) events. We now ignore those events if there is ongoing navigation in a provisional page.

Fixes #9065

### DIFF
--- a/src/server/webkit/wkPage.ts
+++ b/src/server/webkit/wkPage.ts
@@ -1214,7 +1214,7 @@ class WKNavigation {
     this._frameId = event.frameId;
   }
 
-  shouldIgnoreFrameStoppedLoading(frameId: string) {
+  shouldIgnoreFrameStoppedLoading(frameId: string): boolean {
     if (frameId !== this._frameId)
       return false;
     // Navigation in the original frame is canceled and actually continues in the
@@ -1222,7 +1222,7 @@ class WKNavigation {
     return this._page.hasProvisionalPage();
   }
 
-  shouldIgnoreLoadingFailedEvent(event: Protocol.Network.loadingFailedPayload) {
+  shouldIgnoreLoadingFailedEvent(event: Protocol.Network.loadingFailedPayload): boolean {
     if (event.requestId !== this._originalRequestId)
       return false;
     if (!event.canceled)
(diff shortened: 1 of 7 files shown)


src/server/webkit/wkPage.ts
```diff
@@ -1160,3 +1195,62 @@ function isLoadedSecurely(url: string, timing: network.ResourceTiming) {
...
+    this._page = page;
+    this._loaderId = event.loaderId;
+    this._originalRequestId = event.requestId;
+    this._frameId = event.frameId;
+  }
+
+  shouldIgnoreFrameStoppedLoading(frameId: string): boolean {
+    if (frameId !== this._frameId)
+      return false;
+    // Navigation in the original frame is canceled and actually continues in the
```
dgozman, line 1220: I am afraid this might not be always the case. For example, we navigate to cross-origin 204 page that creates a provisional page but never commits?

src/server/webkit/wkPage.ts
```diff
@@ -1160,3 +1195,62 @@ function isLoadedSecurely(url: string, timing: network.ResourceTiming) {
...
+      return false;
+    this._didCheckFirstRequestInProvisionalPage = true;
+    if (event.loaderId !== this._loaderId)
+      return false;
+    this._provisionalPageRequestId = event.requestId;
+    return true;
+  }
+
+  checkIfDuplicateResponseEvent(event: Protocol.Network.responseReceivedPayload): boolean {
+    if (this._originalRequestId === event.requestId || this._provisionalPageRequestId === event.requestId) {
```
dgozman, line 1246: Can we ignore responses from the old web process only? This ensures that we get correct response from the new web process, in case WebKit fixes more security holes in the old web process and send bogus response from there (if any).

yury-s: Coincidentally that works because we generate loadingFinished event in WebKit ourselves when 204 response comes: https://github.com/microsoft/playwright/blob/45b365d95833e0ab61bd49e0b98293542fbd12c0/src/server/webkit/wkPage.ts#L1023-L1030

With network failures we will receive `Playwright.provisionalLoadFailed` and that will abort the navigation, so we are good in that case too. Added a test.

src/server/webkit/wkPage.ts
yury-s: Coincidentally that works because we generate loadingFinished event in WebKit ourselves when 204 response comes: https://github.com/microsoft/playwright/blob/45b365d95833e0ab61bd49e0b98293542fbd12c0/src/server/webkit/wkPage.ts#L1023-L1030

With network failures we will receive `Playwright.provisionalLoadFailed` and that will abort the navigation, so we are good in that case too. Added a test.

src/server/webkit/wkPage.ts
yury-s: At the moment response notification comes from old process we don't know if it's gonna be a cross-process or same process navigation. In case of same-process navigation we have to dispatch messages from the old process. Unless we introduce a separate notification that the navigation turned into cross-process and buffer up some messages from old web process before the notification arrives we cannot safely ignore responses from the old web process  .

mxschmitt:
Seems not needed anymore, closing.

PR #9121, opened 2021-09-24
chore: migrate remaining expect to protocol

pavelfeldman:


src/test/matchers/matchers.ts
```diff
@@ -14,198 +14,206 @@
...
   this: ReturnType<Expect['getState']>,
-  locator: Locator,
+  locator: LocatorEx,
   name: string,
   expected: string | RegExp,
   options?: { timeout?: number },
 ) {
-  return toMatchText.call(this, 'toHaveCSS', locator, 'Locator', async (expected, isNot, timeout) => {
-    return await (locator as any)._expect('to.have.css', { expected, isNot, timeout, data: { name } });
+  return toMatchText.call(this, 'toHaveCSS', locator, 'Locator', async (isNot, timeout) => {
```
dgozman, line 171: (for the follow up): I guess we do not have to pass `timeout`/`options` through `toMatchText` anymore, because timeout does not change.

dgozman (approved):

PR #9143, opened 2021-09-24
fix: increase recent logs buffer

yury-s:
Sometimes top part of the stack is trimmed in the log, like [here](https://github.com/microsoft/playwright/pull/9119/checks?check_run_id=3694112473)

### DIFF
--- a/src/utils/debugLogger.ts
+++ b/src/utils/debugLogger.ts
@@ -64,7 +64,7 @@ class DebugLogger {
 
 export const debugLogger = new DebugLogger();
 
-const kLogCount = 50;
+const kLogCount = 500;
 export class RecentLogsCollector {
   private _logs: string[] = [];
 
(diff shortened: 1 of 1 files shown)


src/utils/debugLogger.ts
```diff
@@ -64,7 +64,7 @@ class DebugLogger {
 
 export const debugLogger = new DebugLogger();
 
-const kLogCount = 50;
+const kLogCount = 500;
```
pavelfeldman, line 67: that's quite spammy, how about 150?

pavelfeldman (approved):

src/utils/debugLogger.ts
yury-s: We already set it to 500  before (https://github.com/microsoft/playwright/pull/6332) but I'm fine with 150 too :grin:

PR #9149, opened 2021-09-25
chore: fix PlaywrightClient disconnection logic

mxschmitt:
The reason was that the selector channels were not unregistered like for BrowserType.connect:

https://github.com/microsoft/playwright/blob/9b0e0c22735f86e049b174bcfd3cde185284a305/src/client/browserType.ts#L178-L181

This resulted that a method on a channel was called which was already closed (ws connection) (then the response was never received and it stalled):

https://github.com/microsoft/playwright/blob/34c1b338bed0872355096fb459ddd494392a13eb/src/client/selectors.ts#L30-L31

To prevent this in the future, I've added a check to verify that the WS connection (PlaywrightClient) is not CLOSED OR CLOSING when writing to the WS connection.

### DIFF
--- a/src/remote/playwrightClient.ts
+++ b/src/remote/playwrightClient.ts
@@ -32,14 +32,22 @@ export class PlaywrightClient {
     const { wsEndpoint, timeout = 30000 } = options;
     const connection = new Connection();
     const ws = new WebSocket(wsEndpoint);
-    connection.onmessage = message => ws.send(JSON.stringify(message));
+    connection.onmessage = message => {
+      if (ws.readyState === 2 /** CLOSING */ || ws.readyState === 3 /** CLOSED */)
+        throw new Error('PlaywrightClient: writing to closed WebSocket connection');
+      ws.send(JSON.stringify(message));
+    };
     ws.on('message', message => connection.dispatch(JSON.parse(message.toString())));
     const errorPromise = new Promise((_, reject) => ws.on('error', error => reject(error)));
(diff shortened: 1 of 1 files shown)


src/remote/playwrightClient.ts
```diff
@@ -32,15 +32,23 @@ export class PlaywrightClient {
     const { wsEndpoint, timeout = 30000 } = options;
     const connection = new Connection();
     const ws = new WebSocket(wsEndpoint);
-    connection.onmessage = message => ws.send(JSON.stringify(message));
+    connection.onmessage = message => {
+      if (ws.readyState === 2 /** CLOSING */ || ws.readyState === 3 /** CLOSED */)
```
dgozman, line 36: I think this can actually happen, and websocket is going to throw by itself. Should we reverse the check instead?

mxschmitt: The `ws` library does only throw if you pass a callback to the `send` method and/or promisify it which ends up like a silent faillure with the current state: 
https://github.com/websockets/ws/blob/055949fd23b0d7f6a23ba9b9532b7834909df192/lib/websocket.js#L940-L946

dgozman: > What is a use-case for that it can happen?

This code does not use `makeWaitForNextTask` currently, and it should. Once it does that, I _think_ we can end up dispatching a message after ws closes because we introduce asynchrony. I'd better be safe here and not throw.

mxschmitt: okay let me revert that part.

dgozman (approved):

Issue #9147, opened 2021-09-25
[Feature] Add GitHub Action failed test annotation for failed tests


sidharthv96:
I'm writing a Github Action which reports failures in Playwright tests using the JSON Reporter.

But the JSON Output does not have the exact location of the error, just the test which failed.

From the generated annotation:
![image](https://user-images.githubusercontent.com/10703445/134772245-ecc9af36-e6ac-4743-bcc2-19d3565866f8.png)

When running locally:
<img width="584" alt="image" src="https://user-images.githubusercontent.com/10703445/134771500-2aa5315a-0192-4396-9f66-4bc02435fe87.png">

If we could add an optional location parameter to `JSONReportTestResult`, that would be helpful in showing the exact location.

sidharthv96:
https://github.com/sidharthv96/playwright-report-action

This is my current implementation.

mxschmitt:
Workaround for now: You should be able to use `stack-utils` to parse the exception stack and place it at the corresponding position in the source-code: https://www.npmjs.com/package/stack-utils

sidharthv96:
Thanks, I was looking at https://github.com/microsoft/playwright/blob/master/src/test/reporters/base.ts and 
> Edit: Seems also possible without the GitHub SDK: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message

The cleaner solution than a separate github action would be a github-action-reporter like you suggested in #7038

I will try to play around with it and maybe create a POC with proper log format for annotations.

mxschmitt:
I think it would also make sense to enable such a reporter by default when you are using GitHub Actions with the test-runner. 
 A lot of users might benefit form it. So it might be a candidate for a contribution if you are interested in it. Also It would be great to not use any dependencies if possible to stay lightweight for our users.

sidharthv96:
Yup, the POC is working. No dependencies needed. 

![image](https://user-images.githubusercontent.com/10703445/134784899-01711b0c-6465-4194-ad8c-7eec8f9f8ac7.png)

sidharthv96:
I can raise a PR soon after cleaning it up.

All the warnings and errors in `epilogue` can be added as annotations.

PR #9151, opened 2021-09-26
feat(reporters): Add error position to JSON Report

sidharthv96:
@mxschmitt this is the fix for the initial problem with #9147

Will raise the PR for annotations soon.

### DIFF
--- a/src/test/reporters/base.ts
+++ b/src/test/reporters/base.ts
@@ -243,11 +243,13 @@ export function formatError(error: TestError, file?: string) {
   const stack = error.stack;
   const tokens = [''];
   if (stack) {
-    const { message, stackLines, codeFrame } = prepareErrorStack(
+    const { message, stackLines, position } = prepareErrorStack(
         stack,
         file
     );
     tokens.push(message);
+
+    const codeFrame = generateCodeFrame(file, position);
     if (codeFrame) {
       tokens.push('');
       tokens.push(codeFrame);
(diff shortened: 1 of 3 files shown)


src/test/reporters/json.ts
```diff
@@ -216,13 +224,24 @@ class JSONReporter implements Reporter {
...
+      results: test.results.map(r => this._serializeTestResult(r, test.location.file)),
       status: test.outcome(),
     };
   }
 
-  private _serializeTestResult(result: TestResult): JSONReportTestResult {
+  private _serializeTestResult(result: TestResult, file: string): JSONReportTestResult {
     const steps = result.steps.filter(s => s.category === 'test.step');
+    let position: { column?: number; line?: number } = {};
+    if (file && result.error?.stack !== undefined) {
```
dgozman, line 235: I think we should not duplicate this logic, but rather put it into `positionInFile` and rename it to `prepareErrorStack(stack: string, file?: string): { stackLines: string[], position: ...}`

src/test/reporters/json.ts
```diff
@@ -65,11 +66,18 @@ export interface JSONReportTestResult {
...
   retry: number;
   steps?: JSONReportTestStep[];
-  attachments: { name: string, path?: string, body?: string, contentType: string }[];
+  attachments: {
+    name: string;
+    path?: string;
+    body?: string;
+    contentType: string;
+  }[];
+  line?: number;
```
dgozman, line 79: Let's instead do `errorLocation?: { line: number, column: number }`.

src/test/reporters/base.ts
```diff
@@ -277,6 +272,38 @@ function indent(lines: string, tab: string) {
...
+  const lines = stack.split('\n');
+  let firstStackLine = lines.findIndex(line => line.startsWith('    at '));
+  if (firstStackLine === -1) firstStackLine = lines.length;
+  const message = lines.slice(0, firstStackLine).join('\n');
+  const stackLines = lines.slice(firstStackLine);
+  const position = file ? positionInFile(stackLines, file) : undefined;
+  let codeFrame;
+  if (position) {
+    const source = fs.readFileSync(file!, 'utf8');
+    codeFrame = codeFrameColumns(
```
dgozman, line 293: I wouldn't do `codeFrame` in this method - it is some expensive unnecessary work for the json reporter. Why don't we leave it in `formatError`?

src/test/reporters/json.ts
dgozman: I feel strongly that test result does not have a line/column because it is not defined in the source like suite or test. `errorLocation` works nicely here, in line with https://playwright.dev/docs/api/class-location.

src/test/reporters/base.ts
sidharthv96: Done

src/test/reporters/json.ts
sidharthv96: Done.

Just curious, shouldn't `JSONReportSuite` and `JSONReportSpec` be using the Location class as `file`, `line` and `column` are being used there?

dgozman (approved):
Looks great, thank you for the PR!

PR #9155, opened 2021-09-26
feat(reporter): Add github actions reporter

### LINKED ISSUE #9147: [Feature] Add GitHub Action failed test annotation for failed tests

sidharthv96:
This is a POC with minimal changes in base and a bit of duplication in logic.
I have another version where there are some changes in base to get additional data like position extracted from format functions.

Fixes: #9147
Example annotations can be found at: https://github.com/sidharthv96/jest-coverage-report-action-test/actions/runs/1275256978

TODO: 
- [ ] Unit Tests
- [ ] Doc Updates
- [ ] Annotate Flaky tests as warning

Needs Discussion if required:
- [x] ~~Running this reporter automatically when running in Actions. (This will only report from within actions, so users could simply add it to their config too..)~~ Land reporter first.
- [ ] Precisely mark slow tests (Might require additional changes in base/logic duplication)


sidharthv96:
Thanks for the review @pavelfeldman :)

> Please check out our other code to make sure your JS aligns with ours.

Can you please point to which places? I've changed the interfaces to types, and used private in the class.
Apart from that, almost all of it was copied from base.

> I don't think we should enable it based on the env for now, we can land the reporter first and enable it later

Sure. I added that for discussion based on https://github.com/microsoft/playwright/issues/9147#issuecomment-927176976

> The amount of copy-paste for epilogue and others looks on the high side, do we have different behavior there or we could align it with the base reporter?

Yup, as I noted, this version is one with almost no change in base. I have [a version](https://github.com/sidharthv96/jest-coverage-report-action-test/blob/master/app/base.ts) where there are changes in base to return details like error position from the formatting functions. That would change the signature of the functions in base. 
I can do that if that's ok. I could make that a separate PR so you can see the changes easily.

PR #9156, opened 2021-09-26
feat: add path option to `toMatchSnapshot`

### LINKED ISSUE #7792: [Feature]: Can I somehow change the directory for saving screenshots?
### Your question

I used this https://playwright.dev/java/docs/test-snapshots function
Can I somehow change the directory for saving screenshots?

nickofthyme:
allow `snapshotName` to accept an array of path segments in `toMathchSnapshot`.

```ts
// example.spec.ts

test('example test', async ({ page }) => {
  expect(await page.screenshot()).toMatchSnapshot(['test', 'path', 'landing.png']);
});
```

This would set and lookup the screenshot from `example.spec.ts-snapshots/test/path/landing-<suffix>.png`.

> The path segment is contained to within the base snapshot directory for the given test file (e.g. `example.spec.ts-snapshots/`). Any path resolving to outside of this directory is ignored.

The current functionality of `snapshotName` as a `'string'` is preserved. This string will continue to be sanitized for file path as is currently done.

```ts
// example.spec.ts

test('example test', async ({ page }) => {
  expect(await page.screenshot()).toMatchSnapshot('test/path/landing.png');
});
```
This would set and lookup the screenshot from `example.spec.ts-snapshots/test-path-landing-<suffix>.png`.

Related to #7792


src/test/util.ts
```diff
@@ -144,4 +144,14 @@ export function sanitizeForFilePath(s: string) {
...
 
+/**
+ * prevent parent directory traversal
+ */
+export function sanitizeForParentPath(parentPath: string, p?: string): string {
+  if (!p) return '';
+
+  const joinedPath = path.join(parentPath, path.normalize(p));
+  return /\.{2,}\//.test(path.relative(parentPath, joinedPath)) ? '' : path.normalize(p);
+}
```
nickofthyme, line 155: I am open to better ways of doing this.

src/test/util.ts
```diff
@@ -144,4 +144,14 @@ export function sanitizeForFilePath(s: string) {
...
+
```
nickofthyme, line 156: I am open to better ways of doing this.

tests/playwright-test/golden.spec.ts
```diff
@@ -488,6 +488,54 @@ test('should write missing expectations with sanitized snapshot name', async ({r
...
+  expect(result.exitCode).toBe(0);
+});
+
+test('should set snapshot with path', async ({ runInlineTest }, testInfo) => {
+  const result = await runInlineTest({
+    ...files,
+    'a.spec.js': `
+      const { test } = require('./helper');
+      test('is a test', ({}) => {
+        expect('Hello world').toMatchSnapshot('snapshot.txt', { path: 'test/path' });
```
dgozman, line 512: Should we instead partially revert #7620? The problem we were facing there was `path` going outside of the snapshots directory.

If we make `toMatchSnapshot('test/path/snapshot.txt')` to place snapshot at `a.spec.js-snapshots/test/path/snapshot.txt` that should just work, shouldn't it? This will save on extra params and APIs.

src/test/util.ts
dgozman: Perhaps we should just do `path.resolve()` instead and then check that `resultingPath.startsWith(baseSnapshotDir + path.sep)`?

dgozman (commented):
Thank you for the PR! See comments inline.

tests/playwright-test/golden.spec.ts
nickofthyme: Yeah good call I didn't notice that was a recent change. See changes in 9b9b677.

src/test/util.ts
nickofthyme: Yeah I like that approach better. See change in 0d0510a.

nickofthyme:
@dgozman I'm curious what your thoughts are on adding a `config` option much like [`outputDir`](https://playwright.dev/docs/api/class-testconfig#test-config-output-dir) for `snapshotDir` that would be the base directory for all snapshots seeing as this would the main request for #7792. For example...

```ts
// playwright.config.ts

  snapshotDir: 'snapshots',
};

export default config;
```

```
playwright.config.ts
tests/
└── foo.spec.ts
snapshots/
└── foo.spec.ts-snapshots/
     ├── basic-test-Chrome-linux.png
     └── basic-test-Safari-linux.png
```

Would there be any issue with this? I have started on these changes but think I'll put them in a follow up PR to this one.

src/test/util.ts
```diff
@@ -144,4 +144,13 @@ export function sanitizeForFilePath(s: string) {
   return s.replace(/[\x00-\x2F\x3A-\x40\x5B-\x60\x7B-\x7F]+/g, '-');
 }
 
+/**
+ * Get absolute path contained within parent directory, otherwise returns parent path
+ */
+export function getContainedPath(parentPath: string, subPath?: string): string {
+  if (!subPath) return parentPath;
+  const resolvedPath = path.resolve(parentPath, subPath);
+  return resolvedPath.startsWith(parentPath) ? resolvedPath : path.join(parentPath, path.basename(subPath));
```
dgozman, line 153: I think this needs to be `startsWith(parentPath  + path.sep)`, otherwise `toMatchSnapshot('../foo.spec.js-snapshots123/bar.txt')` will escape.

src/test/util.ts
```diff
@@ -144,4 +144,13 @@ export function sanitizeForFilePath(s: string) {
   return s.replace(/[\x00-\x2F\x3A-\x40\x5B-\x60\x7B-\x7F]+/g, '-');
 }
 
+/**
+ * Get absolute path contained within parent directory, otherwise returns parent path
```
dgozman, line 148: Why don't we throw a nice error instead of constructing an unexpected path? I think that would be appreciated by the user.

src/test/workerRunner.ts
```diff
@@ -242,20 +242,23 @@ export class WorkerRunner extends EventEmitter {
       outputDir: baseOutputDir,
       outputPath: (...pathSegments: string[]): string => {
         fs.mkdirSync(baseOutputDir, { recursive: true });
-        return path.join(baseOutputDir, ...pathSegments);
+        return getContainedPath(baseOutputDir, path.join(...pathSegments));
       },
       snapshotPath: (snapshotName: string): string => {
```
dgozman, line 247: Sorry for not spotting it earlier, but perhaps we want `...pathSegments: string[]` here, similar to `outputPath`? This way we can make it strictly-backwards-compatible (if we bring back the `sanitizeForFilePath()` call for each path segment), but also allow for the new usecase.

docs/src/test-snapshots-js.md
```diff
@@ -55,7 +55,8 @@ Sometimes you need to update the reference screenshot, for example when the page
 npx playwright test --update-snapshots
 ```
 
-Note that `snapshotName` is *not a path* relative to the test file, so don't try to use it like `expect(value).toMatchSnapshot('../../test-snapshots/snapshot.png')`.
+> Note that `snapshotName` also accepts a relative path to the snapshot file such as `expect(value).toMatchSnapshot('test-snapshots/snapshot.png')`.
+> However, this path is enfored to be within the snapshots directory for each test file (i.e. `a.spec.js-snapshots`) and will be ignored otherwise.
```
dgozman, line 59: - typo: enfored -> enforced
- I'd make it throw when it goes outside of the snapshots directory, and so change this to `However, this path must stay within the snapshots directory for each test file (i.e. a.spec.js-snapshots), otherwise it will throw.`
- A similar comment in `class-testinfo.md` for `outputPath` and `snapshotPath` methods would be great!

dgozman (commented):
I am sorry that this PR turns into a lot of work, but it is going to be a great feature!

src/test/workerRunner.ts
nickofthyme: This sound good to keep BC. Just to clarify your idea. So `snapshotPath` now takes `pathSegments` like `outputPath` and sanitizes each of them before joining. But `outputPath` will **not** sanitize the path segments which could make the two paths inconsistent. I realize they are independent directories but they mirror each other currently, Is this acceptable?

Also what is your thoughts about the `toMatchSnapshot(name)` api? Would it remain with `name` and then we would just need to do `snapshotPath(name.split(path.sep))`? Or do you have ideas on maintaining this BC as well?

dgozman: Great questions!

I did confuse `snapshotPath()` and `toMatchSnapshot()` apis. The latter one is important, and we cannot make `...pathSegments` there. But we can go with `toMatchSnapshot(name: string | string[], options?)` and let the user pass an array explicitly. That's not as convenient, but seems like a niche feature. Also, passing an array instead of `foo/bar/baz` makes it cross-platform out of the box because you don't need to worry about path separators. WDYT?

Re: different behavior of `outputPath` vs `snapshotPath` - I think that's fine. We can also move sanitization into `toMatchSnapshot` and only apply it when you pass a single string (as before), and avoid sanitization when given an array. This way, `TestInfo` methods will behave similarly, while `toMatchSnapshot` will be backwards compatible. Does this make sense?

docs/src/test-snapshots-js.md
nickofthyme: Good ideas, fixed in e899857.

BTW do I need to manually update `docs/src/test-api/class-testinfo.md` and `docs/src/test-snapshots-js.md` from `types/test.d.ts` or is there some auto generation script I am missing?

src/test/util.ts
nickofthyme: Wasn't sure what the protocol was for errors but I like the idea. Let me know what you think of df1e578

src/test/util.ts
nickofthyme: I don't exactly follow, are saying in the event someone passes a path the goes to the parent then back into the correct snapshot directory it will not escape the path? If I were to test something like...

```ts
test.only('should prevent reassigning snapshot test path', async ({ runInlineTest }) => {
    ...files,
    'a.spec.js-snapshots/snapshot.txt': `Hello world`,
    'a.spec.js': `
      const { test } = require('./helper');
      test('is a test', ({}) => {
        expect('Hello world').toMatchSnapshot(['..', 'a.spec.js-snapshots', 'snapshot.txt']);
      });
    `
  });

});
```

This test fails because `outputPath` path is expecting `/test-results/a-is-a-test` and not `/test-results/a.spec.js-snapshots` since the `outputPath` is formatted differently than snapshots.

```
  1) a.spec.js:6:7 › is a test =====================================================================

    Error: Path defined outside of parent directory
      path: ../a.spec.js-snapshots/snapshot.txt
      resolved path: /Users/nicholaspartridge/Documents/repos/playwright/test-results/golden-should-prevent-reassigning-snapshot-test-path-playwright-test/test-results/a.spec.js-snapshots/snapshot.txt
      parent: /Users/nicholaspartridge/Documents/repos/playwright/test-results/golden-should-prevent-reassigning-snapshot-test-path-playwright-test/test-results/a-is-a-test

      5 |       const { test } = require('./helper');
      6 |       test('is a test', ({}) => {
    > 7 |         expect('Hello world').toMatchSnapshot(['..', 'a.spec.js-snapshots', 'snapshot.txt']);
        |                               ^
      8 |       });
```

This seems like an edge case that could be resolved with the new error message. Is your intent for this case to throw on `snapshotPath`?

docs/src/test-snapshots-js.md
```diff
@@ -55,7 +55,8 @@ Sometimes you need to update the reference screenshot, for example when the page
 npx playwright test --update-snapshots
 ```
 
-Note that `snapshotName` is *not a path* relative to the test file, so don't try to use it like `expect(value).toMatchSnapshot('../../test-snapshots/snapshot.png')`.
+> Note that `snapshotName` also accepts a relative path to the snapshot file such as `expect(value).toMatchSnapshot('test-snapshots/snapshot.png')`.
```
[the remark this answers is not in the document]
nickofthyme: Fixed in e899857

src/test/workerRunner.ts
nickofthyme: Ok yeah that clears up my concerns, I think that all makes sense. Here are my changes based on my understanding 93d9632. Let me know if I missed something. Happy to update anything big or small 👍🏼.

nickofthyme:
I wasn't sure about the naming in the docs for `toMatchSnapshot` and changing it from `snapshotName` to `pathSegments`. Might be best to keep it as is with `snapshotName` with the note about using it as `pathSegments`. Also looks like I might need to fix some tests once you take another look.

tests/playwright-test/golden.spec.ts
```diff
@@ -472,22 +471,91 @@ test('should sanitize snapshot name', async ({ runInlineTest }) => {
...
+  }, { 'update-snapshots': true });
+
+  expect(result.exitCode).toBe(0);
+  const snapshotOutputPath = testInfo.outputPath('a.spec.js-snapshots/test/path/snapshot.txt');
   expect(result.output).toContain(`${snapshotOutputPath} is missing in snapshots, writing actual`);
   const data = fs.readFileSync(snapshotOutputPath);
   expect(data.toString()).toBe('Hello world');
 });
 
+test.only('should attach expected/actual/diff with snapshot path', async ({ runInlineTest }, testInfo) => {
```
dgozman, line 520: Don't forget to remove `test.only` 😄

tests/playwright-test/test-output-dir.spec.ts
```diff
@@ -216,6 +216,40 @@ test('should include the project name', async ({ runInlineTest }) => {
...
+      });
+      test('test with parent path', async ({}, testInfo) => {
+        console.log(testInfo.snapshotPath('../test/path/bar.txt').replace(/\\\\/g, '/'));
+      });
+    `,
+  });
+
+  expect(result.exitCode).toBe(0);
+  expect(result.results[0].status).toBe('passed');
+  expect(result.results[1].status).toBe('passed'); // reverts to base snapshot path
```
dgozman, line 247: So, this one should throw now, right?

dgozman (commented):
I think this works nicely, just a few code comments.

> BTW do I need to manually update docs/src/test-api/class-testinfo.md and docs/src/test-snapshots-js.md from types/test.d.ts or is there some auto generation script I am missing?

Files in `docs/src` are the source of truth, `types/test.d.ts` is generated from them by `npm run watch`.

> I wasn't sure about the naming in the docs for toMatchSnapshot and changing it from snapshotName to pathSegments. Might be best to keep it as is with snapshotName with the note about using it as pathSegments

I agree that it's better to keep `snapshotName`.

tests/playwright-test/golden.spec.ts
```diff
@@ -456,12 +456,11 @@ test('should respect project threshold', async ({ runInlineTest }) => {
   expect(result.exitCode).toBe(0);
 });
 
-test('should sanitize snapshot name', async ({ runInlineTest }) => {
+test('should sanitize snapshot name when passed as string', async ({ runInlineTest }) => {
   const result = await runInlineTest({
-    ...files,
-    'a.spec.js-snapshots/-snapshot-.txt': `Hello world`,
+    'a.spec.js-snapshots/-snapshot--darwin.txt': `Hello world`,
```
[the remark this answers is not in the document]
nickofthyme: Thanks I was wondering where that `-darwin` came from. These spec files took me a second to figure out.

src/test/matchers/toMatchSnapshot.ts
```diff
@@ -40,10 +42,11 @@ export function toMatchSnapshot(this: ReturnType<Expect['getState']>, received:
   if (options.threshold === undefined && projectThreshold !== undefined)
     options.threshold = projectThreshold;
 
+  const pathSegments = Array.isArray(options.name) ? options.name : [addSuffix(options.name, '', undefined, true)];
```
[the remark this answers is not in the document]
nickofthyme: Yeah I was trying to reuse the suffix code, but the sanitize makes it extra fancy 😄. Happy to put this back inline.

src/test/matchers/golden.ts
```diff
@@ -82,15 +83,15 @@ function compareText(actual: Buffer | string, expectedBuffer: Buffer): { diff?:
...
+  pathSegments: string[],
+  snapshotPath: TestInfoImpl['snapshotPath'],
+  outputPath: TestInfoImpl['outputPath'],
   updateSnapshots: UpdateSnapshots,
   withNegateComparison: boolean,
   options?: { threshold?: number }
 ): { pass: boolean; message?: string; expectedPath?: string, actualPath?: string, diffPath?: string, mimeType?: string } {
-  const snapshotFile = snapshotPath(name);
-  const outputFile = outputPath(name);
+  const snapshotFile = snapshotPath(...pathSegments);
```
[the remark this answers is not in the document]
nickofthyme: Ah gotcha I thought you wanted it the same as `outputPath` 👍🏼

tests/playwright-test/test-output-dir.spec.ts
nickofthyme: Yes correct, 311fa94.

tests/playwright-test/golden.spec.ts
nickofthyme: Whoops 😅  311fa94

tests/playwright-test/golden.spec.ts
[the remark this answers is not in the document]
nickofthyme: 311fa94

src/test/workerRunner.ts
```diff
@@ -255,20 +255,20 @@ export class WorkerRunner extends EventEmitter {
...
         if (testInfo.snapshotSuffix)
           suffix += '-' + testInfo.snapshotSuffix;
-        const ext = path.extname(snapshotName);
-        if (ext)
-          snapshotName = sanitizeForFilePath(snapshotName.substring(0, snapshotName.length - ext.length)) + suffix + ext;
-        else
-          snapshotName = sanitizeForFilePath(snapshotName) + suffix;
-        return path.join(test._requireFile + '-snapshots', snapshotName);
+
+        const snapshotPath = path.join(...pathSegments.map((segment, i, { length }) =>
```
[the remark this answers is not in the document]
nickofthyme: 311fa94

src/test/matchers/toMatchSnapshot.ts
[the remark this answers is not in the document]
nickofthyme: 311fa94

src/test/util.ts
```diff
@@ -144,4 +144,26 @@ export function sanitizeForFilePath(s: string) {
   return s.replace(/[\x00-\x2F\x3A-\x40\x5B-\x60\x7B-\x7F]+/g, '-');
 }
 
+export function addSuffix(filePath: string, suffix: string, customExtension?: string, sanitize = false): string {
```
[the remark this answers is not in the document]
nickofthyme: 311fa94

src/test/matchers/golden.ts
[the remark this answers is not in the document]
nickofthyme: I changed this and array in 311fa94 but lmk if you actually wanted the spread.

src/test/util.ts
```diff
@@ -144,4 +144,26 @@ export function sanitizeForFilePath(s: string) {
...
+
+/**
+ * Returns absolute path contained within parent directory.
+ */
+export function getContainedPath(parentPath: string, subPath: string = ''): string {
+  const resolvedPath = path.resolve(parentPath, subPath);
+
+  if (resolvedPath.startsWith(parentPath)) return resolvedPath;
+
+  throw new Error(`Path defined outside of parent directory
```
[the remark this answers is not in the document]
nickofthyme: That sounds good to me. I wasn't sure what the best way to do this so I just returned `null` if outside of the parent and throw an error at both callsites. See 07567bc. Let me know if you have any better ideas for doing this or the error message.

tests/playwright-test/golden.spec.ts
```diff
@@ -472,22 +472,91 @@ test('should sanitize snapshot name', async ({ runInlineTest }) => {
...
+      const { test } = pwt;
       test('is a test', ({}) => {
         expect('Hello world').toMatchSnapshot('../../snapshot!.txt');
       });
     `
   }, {}, { CI: '' });
 
   expect(result.exitCode).toBe(1);
-  const snapshotOutputPath = testInfo.outputPath('a.spec.js-snapshots/-snapshot-.txt');
+  const snapshotOutputPath = testInfo.outputPath('a.spec.js-snapshots/-snapshot--darwin.txt');
```
dgozman, line 484: Use `helper` and drop `darwin` here.

tests/playwright-test/golden.spec.ts
```diff
@@ -472,22 +472,91 @@ test('should sanitize snapshot name', async ({ runInlineTest }) => {
...
+  expect(result.output).toContain(`${snapshotOutputPath} is missing in snapshots, writing actual`);
+  const data = fs.readFileSync(snapshotOutputPath);
+  expect(data.toString()).toBe('Hello world');
+});
+
+test('should join array of snapshot path segments without sanitizing ', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.spec.js-snapshots/test/path/snapshot-darwin.txt': `Hello world`,
```
dgozman, line 492: Same here.

tests/playwright-test/test-output-dir.spec.ts
```diff
@@ -216,6 +216,61 @@ test('should include the project name', async ({ runInlineTest }) => {
...
+    'my-test.spec.js': `
+      const { test } = require('./helper');
+      test('test with parent path', async ({}, testInfo) => {
+        console.log(testInfo.snapshotPath(['..', 'test', 'path', 'bar.txt']).replace(/\\\\/g, '/'));
+      });
+    `,
+  });
+
+  expect(result.exitCode).toBe(1);
+  expect(result.results[0].status).toBe('failed');
```
dgozman, line 271: `expect(result.output).toContain('<insert the error message you throw in the method>')`

docs/src/test-advanced-js.md
```diff
@@ -120,7 +120,7 @@ In addition to everything from the [`workerInfo`](#workerinfo), the following in
 - `timeout: number` - Test timeout.
 - `annotations` - [Annotations](./test-annotations.md) that were added to the test.
 - `snapshotSuffix: string` - Suffix used to locate snapshots for the test.
-- `snapshotPath(snapshotName: string)` - Function that returns the full path to a particular snapshot for the test.
+- `snapshotPath(pathSegments: string[])` - Function that returns the full path to a particular snapshot for the test.
```
dgozman, line 123: > I changed this and array in 311fa94 but lmk if you actually wanted the spread.

Sorry, I was confused by some lint error. Spread is backwards compatible, so spread for sure! Luckily (or actually not!) there are still some tests failing, so this is not the only change to be made 😕

dgozman (commented):
Apart from the spread and some failing tests, this looks perfect!

docs/src/test-advanced-js.md
nickofthyme: No worries, simple fix. See 76efbc2.

docs/src/test-advanced-js.md
```diff
@@ -120,7 +120,7 @@ In addition to everything from the [`workerInfo`](#workerinfo), the following in
 - `timeout: number` - Test timeout.
 - `annotations` - [Annotations](./test-annotations.md) that were added to the test.
 - `snapshotSuffix: string` - Suffix used to locate snapshots for the test.
-- `snapshotPath(snapshotName: string)` - Function that returns the full path to a particular snapshot for the test.
+- `snapshotPath(...pathSegments: string[])` - Function that returns the full path to a particular snapshot for the test.
```
nickofthyme, line 123: This seems to control argument types for `TestInfo` inside `types/test.d.ts` but for some reason it keeps overriding the type for `snapshotPath` in `types/test.s.ts` to `snapshotPath: (pathSegments: string[]) => string;`. Any ideas?

tests/playwright-test/test-output-dir.spec.ts
nickofthyme: 76efbc2

tests/playwright-test/golden.spec.ts
nickofthyme: 76efbc2

tests/playwright-test/golden.spec.ts
nickofthyme: 76efbc2

docs/src/test-advanced-js.md
nickofthyme: Ugh nvm it was `utils/generate_types/overrides-test.d.ts` ac578b1

src/test/util.ts
```diff
@@ -144,4 +144,23 @@ export function sanitizeForFilePath(s: string) {
...
+  return (sanitize ? sanitizeForFilePath(base) : base) + suffix + (customExtension || ext);
+}
+
+/**
+ * Returns absolute path contained within parent directory.
+ */
+export function getContainedPath(parentPath: string, subPath: string = ''): string | null {
+  const resolvedPath = path.resolve(parentPath, subPath);
+
+  if (resolvedPath.startsWith(parentPath + path.sep)) return resolvedPath;
```
nickofthyme, line 161: Can you explain why `path.sep` is needed here? it's causing a test failure for an empty `''` path here https://github.com/microsoft/playwright/pull/9156/checks?check_run_id=3762066712.

dgozman: That's to avoid `getContainedPath('/foo/bar', '/foo/barbaz')` returning `true`. I guess this can be fixed with
```js
if (resolvedPath === parentPath || resolvedPath.startsWith(parentPath + path.sep)) return resolvedPath;
```

nickofthyme: 9bc8041

tests/page/page-request-fulfill.spec.ts
```diff
@@ -217,7 +217,7 @@ it('should fulfill with fetch result', async ({ page, server, isElectron }) => {
   expect(await response.json()).toEqual({ 'foo': 'bar' });
 });
 
-it('should fulfill with fetch result and overrides', async ({ page, server, isElectron }) => {
+it.only('should fulfill with fetch result and overrides', async ({ page, server, isElectron }) => {
```
dgozman, line 220:
```suggestion
it('should fulfill with fetch result and overrides', async ({ page, server, isElectron }) => {
```

tests/playwright-test/test-output-dir.spec.ts
```diff
@@ -167,7 +167,7 @@ test('should include the project name', async ({ runInlineTest }) => {
       });
       test2('test 2', async ({}, testInfo) => {
         console.log(testInfo.outputPath('bar.txt').replace(/\\\\/g, '/'));
-        console.log(testInfo.snapshotPath('bar.txt').replace(/\\\\/g, '/'));
+        console.log(testInfo.snapshotPath('bar.txt').replace(/\\\\/g, ''));
```
dgozman, line 170:
```suggestion
        console.log(testInfo.snapshotPath('bar.txt').replace(/\\\\/g, '/'));
```

dgozman (commented):
Two suggestions to fix failing tests, the rest is great!

dgozman (approved):

PR #9195, opened 2021-09-28
feat(fetch): send Playwright as default user-agent for global fetch

yury-s:
By default use 'Playwright' as user-agent in global fetch instead of empty string. I didn't make it `Playwright/1.16.0-next (x64/linux/5.11.0-34-generic)` which is what we use in our browserType.connect as there were no requests for that and we don't necessarily want to expose that much data.

#5999

### DIFF
--- a/src/server/fetch.ts
+++ b/src/server/fetch.ts
@@ -23,7 +23,7 @@ import zlib from 'zlib';
 import { HTTPCredentials } from '../../types/types';
 import { NameValue, NewRequestOptions } from '../common/types';
 import { TimeoutSettings } from '../utils/timeoutSettings';
-import { createGuid, isFilePayload, monotonicTime } from '../utils/utils';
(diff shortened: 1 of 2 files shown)


dgozman (approved):

src/server/fetch.ts
```diff
@@ -352,7 +352,7 @@ export class GlobalFetchRequest extends FetchRequest {
     }
     this._options = {
       baseURL: options.baseURL,
-      userAgent: options.userAgent || '',
+      userAgent: options.userAgent || 'Playwright',
```
mxschmitt, line 355: nit: to be more compliant we should maybe do Playwright/v1.15.0 instead: https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3

Since WAFs parse them and might find just Playwright too little, I would follow what the other http clients also do.

mxschmitt (approved):

src/server/fetch.ts
yury-s: Added version.

PR #9200, opened 2021-09-28
feat(api): introduce locator.waitFor

### LINKED ISSUE #8577: [Feature] Extend locator api with functions like waitForSelector()
Given the pseudo-POM below and the case where you would need to wait for an element for any reason, you'd have to repeat the selector in two places. It hurts reusability.

```
export class PageObject {
  private readonly page: Page
  private readonly someButton: Locator
  
  constructor(page: Page) {
    this.page = page
    this.someButton = this.page.locator('#someButtton')
  }
  
  async clickSomeButton() {
    await this.page.waitForSelector('#someButton')
    await this.someButton.click()
  }
}
```

While using the old method, you could set selector value once and then simply reuse the variable as much as you want:
<details>
  <summary>POM with page actions</summary>

```
export class PageObject {
  private readonly page: Page
  private readonly someButton: string

  constructor(page: Page) {
    this.page = page
    this.someButton = ('#someButtton')
  }

  async clickSomeButton() {
    await this.page.waitForSelector(this.someButton)
    await this.page.click(this.someButton)
  }
}
```
</details>

It'd be nice if locator API would have an option to use functions like waitForSelector(). My use case involves only this function but there might be others.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/8577

### DIFF
--- a/src/client/locator.ts
+++ b/src/client/locator.ts
@@ -213,6 +213,14 @@ export class Locator implements api.Locator {
     return this._frame.$$eval(this._selector, ee => ee.map(e => e.textContent || ''));
   }
 
+  waitFor(options: channels.FrameWaitForSelectorOptions & { state: 'attached' | 'visible' }): Promise<void>;
+  waitFor(options?: channels.FrameWaitForSelectorOptions): Promise<void>;
+  async waitFor(options?: channels.FrameWaitForSelectorOptions): Promise<void> {
+    return this._frame._wrapApiCall(async (channel: channels.FrameChannel) => {
+      await channel.waitForSelector({ selector: this._selector, strict: true, omitReturnValue: true, ...options });
+    });
+  }
+
   async _expect(expression: string, options: channels.FrameExpectOptions): Promise<{ pass: boolean, received?: any, log?: string[] }> {
     return this._frame._wrapApiCall(async (channel: channels.FrameChannel) => {
       const params: any = { selector: this._selector, expression, ...options };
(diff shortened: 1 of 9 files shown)


src/client/locator.ts
```diff
@@ -213,6 +213,14 @@ export class Locator implements api.Locator {
     return this._frame.$$eval(this._selector, ee => ee.map(e => e.textContent || ''));
   }
 
+  waitFor(options: channels.FrameWaitForSelectorOptions & { state: 'attached' | 'visible' }): Promise<void>;
+  waitFor(options?: channels.FrameWaitForSelectorOptions): Promise<void>;
+  async waitFor(options?: channels.FrameWaitForSelectorOptions): Promise<void> {
+    return this._frame._wrapApiCall(async (channel: channels.FrameChannel) => {
+      await channel.waitForSelector({ selector: this._selector, strict: true, ...options });
```
dgozman, line 220: Forgot to pass `omitReturnValue` here?

tests/page/locator-misc-2.spec.ts
```diff
@@ -82,3 +82,12 @@ it('should return bounding box', async ({ page, server, browserName, headless, i
   const box = await element.boundingBox();
   expect(box).toEqual({ x: 100, y: 50, width: 50, height: 50 });
 });
+
+it('should waitFor', async ({ page }) => {
```
dgozman, line 86: Let's add a test with `waitFor({ hidden })` or some other state passed, otherwise we lack coverage.

dgozman (approved):

sarweshkumar2018:
How to handle if the locator returns more than one elements. 
await this.page.locator('[aria-label="bene"]').waitFor(); 

await this.page.locator('[aria-label="bene"]:first-child').waitFor() --- is NOT working

xyide:
> How to handle if the locator returns more than one elements.

I don't think it would be necessary to handle this case.  If any element with a given selector us visible, then naturally the first-child is visible.

PR #9204, opened 2021-09-28
fix(expect): beautiful expect stacks

dgozman:
We now mark our wrapper as `__PWTRAP__[expect.toHaveText]` and
find it later in the stack trace.

Added trace/inspector tests to ensure this behavior in the future.

### DIFF
--- a/src/utils/stackTrace.ts
+++ b/src/utils/stackTrace.ts
@@ -33,8 +33,6 @@ export function rewriteErrorMessage<E extends Error>(e: E, newMessage: string): 
 const ROOT_DIR = path.resolve(__dirname, '..', '..');
 const CLIENT_LIB = path.join(ROOT_DIR, 'lib', 'client');
 const CLIENT_SRC = path.join(ROOT_DIR, 'src', 'client');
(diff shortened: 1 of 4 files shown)


src/utils/stackTrace.ts
```diff
@@ -65,14 +63,7 @@ export function captureStackTrace(): ParsedStackTrace {
...
       return null;
-    const inClient =
-      // Allow fixtures in the reported stacks.
-      (!fileName.includes('test/index') && !fileName.includes('test\\index')) && (
-        frame.file.includes(path.join('node_modules', 'expect'))
-        || fileName.startsWith(CLIENT_LIB)
-        || fileName.startsWith(CLIENT_SRC)
-        || fileName.startsWith(TEST_LIB)
-        || fileName.startsWith(TEST_SRC));
+    const inClient = fileName.startsWith(CLIENT_LIB) || fileName.startsWith(CLIENT_SRC);
```
mxschmitt, line 66: general observation: I have the feeling that we touch this part too often in the past.

pavelfeldman (approved):

PR #9207, opened 2021-09-28
feat(test runner): run multiple web servers

### LINKED ISSUE #8206: [Feature]: Support launching multiple web servers
**Context:**
- GOOD Playwright Version: 1.13
- BAD Playwright Version: 1.14
- Operating System: Mac (didn't test other OS)
- Extra: N/A

**Code Snippet**

This used to work on 1.13 (updated to 1.14 option names). 

```javascript
  webServer: [
    {
      command: 'yarn workspace @app/graphql start',
      port: 4000, 
      timeout: 60000,
      reuseExistingServer: !isCI,
    },
    {
      command: 'yarn workspace @app/frontend start',
      port: 3000, 
      timeout: 60000,
      reuseExistingServer: !isCI,
    }
  ]
};
```

**Describe the bug**

Previously we had the experimental `_launch` option to start a server for tests, and this options accepted an array of several servers to start. With 1.14 released it was renamed to `webServer` and only takes one server as a parameter. Maybe it's a weird use case but it was really helpful to be able to lift several servers. `baseUrl` fixture took the value of the last server launched.

Would it be possible to add this back?

osiux:
Bring back support to launch multiple web servers at once. Fixes #8206 

Mostly bring back things removed on #7906

### DIFF
--- a/src/test/loader.ts
+++ b/src/test/loader.ts
@@ -15,7 +15,7 @@
  */
 
 import { installTransform } from './transform';
-import type { FullConfig, Config, FullProject, Project, ReporterDescription, PreserveOutput } from './types';
+import type { FullConfig, Config, FullProject, Project, ReporterDescription, PreserveOutput, WebServerConfig } from './types';
 import { mergeObjects, errorWithFile } from './util';
 import { setCurrentlyLoadingFileSuite } from './globals';
 import { Suite } from './test';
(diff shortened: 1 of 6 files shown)


JoelEinbinder:
@mxschmitt I am fine with it. However if we want to make ports optional in the future I think we should have a separate `services` option.

PR #9221, opened 2021-09-29
feat(fetch): store cookies between requests

yury-s:
#5999

### DIFF
--- a/src/server/cookieStore.ts
+++ b/src/server/cookieStore.ts
@@ -112,7 +112,7 @@ export class CookieStore {
   }
 }
 
-function domainMatches(value: string, domain: string): boolean {
+export function domainMatches(value: string, domain: string): boolean {
   if (value === domain)
     return true;
   // Only strict match is allowed if domain doesn't start with '.' (host-only-flag is true in the spec)
--- a/src/server/cookieStore.ts
+++ b/src/server/cookieStore.ts
@@ -1,0 +1,133 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


src/server/cookieStore.ts
```diff
@@ -0,0 +1,133 @@
...
+    if (cookie.expired())
+      return;
+    let set = this._nameToCookies.get(cookie.name());
+    if (!set) {
+      set = new Set();
+      this._nameToCookies.set(cookie.name(), set);
+    }
+    CookieStore.pruneExpired(set);
+    // https://datatracker.ietf.org/doc/html/rfc6265#section-5.3
+    for (const other of set) {
```
dgozman, line 88: I think we are missing step 11.2:
`If the newly created cookie was received from a "non-HTTP"
            API and the old-cookie's http-only-flag is set, abort these
            steps and ignore the newly created cookie entirely.`

src/server/fetch.ts
```diff
@@ -388,14 +394,16 @@ function toHeadersArray(rawHeaders: string[]): types.HeadersArray {
 const redirectStatus = [301, 302, 303, 307, 308];
 
 function canSetCookie(cookieDomain: string, hostname: string) {
```
dgozman, line 396: Looks very similar to `domainMatches` - perhaps reuse one?

dgozman (approved):

src/server/cookieStore.ts
yury-s: Well we only set the cookies from network rather than JS and even if/when we expose an API for updating the cookies in the store we want to allow playwright clients to do so.

src/server/fetch.ts
yury-s: Done.

Issue #7822, opened 2021-07-23
Add PDF guide


mxschmitt:
Customers are confused, when a PDF results in a PDF viewer and when in a download event. We should explain how to workaround it in the relevant browsers.

corradin:
@mxschmitt Did you also take a look at the solution I proposed in case people are not able to change the source code of the application under test? https://github.com/microsoft/playwright/issues/6342#issuecomment-929112605

This uses the download event for headless chrome and the page event for all other browsers that use the integrated PDF reader.

mxschmitt:
@corradin yes I've seen that. Creative solution! Question from our side, what in your opinion gets mostly asserted / tested when working with PDF files?

waitForResponse() seems working in all the cases but it seems important to assert the suggested file name?

jvarghese2022:
Is there any simpler way to handle to pdf viewer where the url starts as blob.

mandras73:
Do you have a python version of the workaround?

vincenzo-gasparo:
Here's a python version of the workaround:

```python
import re
from playwright.sync_api import Page, expect


def test_download_pdf(page: Page):
    page.goto("https://www.example.com")

    def handle_pdf(route):
        response = page.context.request.get(route.request)
        route.fulfill(
            response,
            headers={**response.headers, "Content-Disposition": "attachment"},
        )

    page.route(re.compile(r".*\.pdf"), lambda route: handle_pdf(route))

    with page.expect_download() as download_info:
        page.get_by_text("click_here_to_download_pdf").click()
```

kinopop:
> Here's a python version of the workaround:
> 
> ```python
> import re
> from playwright.sync_api import Page, expect
> 
> 
> def test_download_pdf(page: Page):
>     page.goto("https://www.example.com")
> 
>     def handle_pdf(route):
>         response = page.context.request.get(route.request)
>         route.fulfill(
>             response,
>             headers={**response.headers, "Content-Disposition": "attachment"},
>         )
> 
>     page.route(re.compile(r".*\.pdf"), lambda route: handle_pdf(route))
> 
>     with page.expect_download() as download_info:
>         page.get_by_text("click_here_to_download_pdf").click()
> ```

What plug-ins do I lack?
>             response,
error 1 argument
>             response=response,
is that true？
my error ：waiting for event " download"

cobaltautomationdev:
@beegotsy  
[...]

PR #9235, opened 2021-09-30
test: add test for downloading PDF files

mxschmitt:
https://github.com/microsoft/playwright/issues/7822


tests/download.spec.ts
```diff
@@ -580,3 +581,65 @@ it.describe('download event', () => {
...
+  await page.close();
+});
+
+it('should be able to download a inline PDF file', async ({ browser, server, asset }) => {
+  it.fixme(true, 'Wrong download content gets returned');
+  const page = await browser.newPage({ acceptDownloads: true });
+  await page.goto(server.EMPTY_PAGE);
+  await page.route('**/empty.pdf', async route => {
+    // @ts-expect-error
+    const response = await route._continueToResponse({});
```
yury-s, line 605: Given that we may drop this response interception API I'd rather you set Content-Disposition in server.setRoute

tests/pdf.spec.ts
```diff
@@ -18,7 +18,7 @@ import { browserTest as it, expect } from './config/browserTest';
 import fs from 'fs';
 
 it('should be able to save file', async ({ contextFactory, headless, browserName }, testInfo) => {
-  it.skip(!headless || browserName !== 'chromium', 'Printing to pdf is currently only supported in headless chromium.');
+  it.skip(!headless || browserName !== 'chromium', 'Printing to pdf is currently only supported in headed chromium.');
```
yury-s, line 21: Revert? We only print in headless.

yury-s (approved):

tests/download.spec.ts
mxschmitt: discussed offline, using request.get for now.

PR #9245, opened 2021-09-30
docs: projects simplify & crosslink

mrienstra:
1. Remove extraneous bits from example code: This could mislead users into thinking they need to specify `browserName` when using `projects` for device emulation. If we want to demonstrate how to override some details, maybe we can find a more realistic example?
2. Link from [./test-advanced/#projects](https://playwright.dev/docs/test-advanced/#projects) to [./test-configuration/#emulation](https://playwright.dev/docs/test-configuration/#emulation).

### DIFF
--- a/docs/src/test-advanced-js.md
+++ b/docs/src/test-advanced-js.md
@@ -364,7 +364,7 @@ test('test', async ({ page }) => {
 
 ## Projects
 
-Playwright Test supports running multiple test projects at the same time. This is useful for running the same tests in multiple configurations. For example, consider running tests against multiple versions of some REST backend.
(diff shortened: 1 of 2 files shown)


mrienstra:
On a related note, is there any particular reason https://playwright.dev/docs/api/class-playwright#playwright-devices shows:
```js
  ...iPhone
});
```
... Rather than just:
```js
const context = await browser.newContext(iPhone);
```
(where `iPhone` is:
```js
const { webkit, devices } = require('playwright');
const iPhone = devices['iPhone 6'];
```
)

Is this just to make it obvious how to go about extending the device descriptor object?

pavelfeldman:
```
  ...iPhone
});
```

gives users a hint that they can add more options.

PR #9260, opened 2021-10-01
feat: add `snapshotDir` to set base snapshot directory

### LINKED ISSUE #7792: [Feature]: Can I somehow change the directory for saving screenshots?

### LINKED ISSUE #8912: [Feature] Option to save image snapshot in provided directory rather than close to tests spec files.
User case:
Got over 300 screenshots in tests. So far with jest image snapshot we just keep them in git lfs as submodule in the repository so the main repository app builded with webpack doesn't get too big. Over 300 image snapshots already take some considerable space. 

Overall option to keep snapshots in one directory makes it easier to manage them as one.

nickofthyme:
add `snapshotDir` option to `TestConfig` and `TestProject` to set base directory to save snapshot files.

```ts
// playwright.config.ts

  projects: [
    {
      name: 'foo',
    },
    {
      name: 'bar',
      snapshotDir: 'bar_snaps',
    },
  ],
  snapshotDir: 'snaps',
};

export default config;
```

Given the playwright above and as simple `toMatchSnapshot` test such as `a.spec.ts` below...

```ts
// a.spec.ts

test('should snap it', () => {
  expect('Hello world').toMatchSnapshot('snapshot.txt');
});
```

The resulting `TestInfo.snapshotDir` for the `foo` and `bar` projects would be...

- `foo` -> `snaps/a.spec.ts-snapshots/`
- `bar` -> `bar_snaps/a.spec.ts-snapshots/`

closes #8912 

Related to #7792, but I don't think this PR solves their request.

cc: @dgozman


nickofthyme:
@dgozman I have a few thoughts on way to enhance this but didn't want to get ahead of myself.

First I wonder if we can break away from the sub directory for each test file when providing an explicit `snapshotDir`. This would provide greater flexibility and for users to set the output directory as they see fit.

The current approach for nested files makes this strange at least to me. For example say we have a test file nested in `tests/foo.spec.ts`.

```ts
// tests/foo.spec.ts
test('should snap it', () => {
  expect('Hello world').toMatchSnapshot('snapshot.txt');
});
```

This would currently save the snapshots to `foo.spec.ts-snapshots/tests/snapshot.txt` and not `tests/foo.spec.ts-snapshots/snapshot.txt`.

Breaking away from the subdirectory would allow something like `my-snapshots/snapshot.txt`.

Secondly, I wonder if we could do something with [`executeTemplate`](https://github.com/microsoft/playwright/blob/b126a5685b0e047b19bbf54ec7bc753cba1efa8d/packages/create-playwright/src/utils.ts#L74-L78) to interpolate values from [`TestInfo`](https://playwright.dev/docs/api/class-testinfo) and [`TestProject `](https://playwright.dev/docs/api/class-testproject/) that would give the test context and power to the user to avoid collisions. For example `path/to/snapshots/{fileName}/output` would resolve to `path/to/snapshots/foo/output`. I haven't fully thought this part through but I'd love to get your thoughts.

dgozman:
> This would currently save the snapshot to `foo.spec.ts-snapshots/tests/snapshot.txt` and not `tests/foo.spec.ts-snapshots/snapshot.txt` as I would have expected.

I do not think it works this way. For example, [here](https://github.com/microsoft/playwright/tree/master/tests/chromium/oopif.spec.ts-snapshots) we have a `-snapshots` directory right next to the test file, somewhere inside our tests directory structure.

> First I wonder if we can break away from the sub directory for each test file when providing an explicit `snapshotDir`. This would provide greater flexibility for users to set the output directory as they see fit.

I think this is actually what original author of #7792 implied, but I could be wrong. However, the main idea behind current implementation is that sharing snapshots between test files is error-prone - it makes updating and managing them much harder. From our own experience, when we tried to share snapshots between tests, we usually ended up not reusing the snapshot, but copying/creating a new one for each test because that's just easier. With this in mid, `<customSnapshotDir>/my.spec.ts-snapshots/snapshot.txt` seems fine. WDYT?

> For example `path/to/snapshots/{fileName}/output` would resolve to `path/to/snapshots/foo/output`. I haven't fully thought this part through but I'd love to get your thoughts.

This is interesting, but looks very advanced and also incremental to whatever we do with `snapshotsDir`?

nickofthyme:
> I do not think it works this way

Ah yes you are correct, sorry must have transposed the path.

> I think this is actually what original author of #7792 implied, but I could be wrong. However, the main idea behind current implementation is that sharing snapshots between test files is error-prone - it makes updating and managing them much harder. From our own experience, when we tried to share snapshots between tests, we usually ended up not reusing the snapshot, but copying/creating a new one for each test because that's just easier. With this in mid, <customSnapshotDir>/my.spec.ts-snapshots/snapshot.txt seems fine. WDYT?

Yeah I think a direct path would have been their preferred solution but I'll ask if this would suffice. It makes sense to just do `<snapshotDir>/my.spec.ts-snapshots/snapshot.txt`, I keep conflating projects and test suites in my head thinking it would work.

I know that [`jest-image-snapshot`](https://github.com/americanexpress/jest-image-snapshot) dumps all screenshots into a single directory where the name is the concatenation of the test title tree. Then they just start appending a count to the end of the filename to avoid collisions (i.e. `my-test.png`, `my-test-1.png`, `my-test-2.png`) which to me is not a great solution. Right now I don't see a better way to do it so I'll keep the unique test file directory.

> This is interesting, but looks very advanced and also incremental to whatever we do with snapshotsDir?

Yeah good point.

dgozman (commented):
This looks very good, just a few comments around docs/tests.

dgozman (commented):
Looks great except for small documentation typo! Since this patch introduces new API, I will merge it a bit later once we settle down with 1.16 release.

dgozman (approved):

PR #9273, opened 2021-10-02
browser(webkit): support set-cookie in intercepted response(SOUP)

yury-s:
https://github.com/yury-s/WebKit/commit/77c0200938c85c125c1d05566a212837f816225f

#5999

### DIFF
--- a/browser_patches/webkit/patches/bootstrap.diff
+++ b/browser_patches/webkit/patches/bootstrap.diff
@@ -3960,7 +3960,7 @@ index 3386cb879f1178c1b9635775c9a0e864f5b94c52..d2350182f5f061855e8ca172779ad60e
  class Page;
  class SecurityOrigin;
 diff --git a/Source/WebCore/inspector/agents/InspectorNetworkAgent.cpp b/Source/WebCore/inspector/agents/InspectorNetworkAgent.cpp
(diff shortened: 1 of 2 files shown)


browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -4137,17 +4137,19 @@ index 37468432577f64e3d18bc3667ca298f43a3bf503..81e54d04c6142b29241ec5379470ef47
      ResourceRequest request = loader.request();
      if (!!url)
          request.setURL(URL({ }, url));
-@@ -1289,14 +1370,24 @@ Protocol::ErrorStringOr<void> InspectorNetworkAgent::interceptRequestWithRespons
+@@ -1289,14 +1370,25 @@ Protocol::ErrorStringOr<void> InspectorNetworkAgent::interceptRequestWithRespons
      response.setHTTPStatusCode(status);
      response.setHTTPStatusText(statusText);
      HTTPHeaderMap explicitHeaders;
 +    String setCookieValue;
```
pavelfeldman, line 4144: I seriously doubt this `frame->page()->cookieJar().setCookieFromResponse` code below works. You can't modify cookies from the page process, only network process can do that. Otherwise compromised renderer would be able to poison the httpOnly cookies, etc.

yury-s: As discussed offline this is inline with the rest of current network instrumentation.

PR #9284, opened 2021-10-04
feat: Add github to default reporters

### LINKED ISSUE #9281: [BUG] Enable GitHub reporter in default configs and internal tests.
Next step would be probably to automatically add it as a reporter in the default configuration if the `GITHUB_ACTION` environment variable is set + activate it on our configs so we can dogfood it.

_Originally posted by @mxschmitt in https://github.com/microsoft/playwright/issues/9191#issuecomment-933268500_

Relates: #9147

sidharthv96:
Fixes #9281


src/test/cli.ts
```diff
@@ -19,20 +19,19 @@
...
 
 const defaultTimeout = 30000;
-const defaultReporter: BuiltInReporter = process.env.CI ? 'dot' : 'list';
 const tsConfig = 'playwright.config.ts';
 const jsConfig = 'playwright.config.js';
 const mjsConfig = 'playwright.config.mjs';
 const defaultConfig: Config = {
   preserveOutput: 'always',
-  reporter: [ [defaultReporter] ],
+  reporter: getDefaultReporters(),
```
mxschmitt, line 34: lets store this inside a variable, change the function return type to []string so you can use defaultReporters.join(", ") in the CLI command description.

mxschmitt (approved):
LGTM, but lets wait for someone else to approve as-well. In the meantime you should be able to add it to our internal Playwright configs.

PR #9287, opened 2021-10-04
feat: Add github reporter to internal tests

sidharthv96:


mxschmitt:
@sidharthv96 seems like the ASCII bash annotations needs to get [removed properly](https://github.com/microsoft/playwright/blob/be30f9f1c44e59f7dd6ccf86b22d4d0b478d1dc4/src/test/reporters/base.ts#L411-L414): https://github.com/microsoft/playwright/pull/9287/files

pavelfeldman:
After seeing it in action on this and other PRs, I'd say that we don't want it enable for our repo.

Here is how I see it:

Pros:
- provides some context for the test failure source code

Cons:
- most of the time, the failure is in the 'unmodified files'
- code review is obstructed
- 'Files' tab becomes misused for the 'Test Results' experience.

The reporter itself is still useful to those that have separate repositories for tests and where the failure is likely to belong to the PR. Or it might work for those that don't practice PR code reviews.

sidharthv96:
But isn't the unmodified failures a code smell? Ideally there shouldn't be any right?
Also, they are tucked away at the bottom, not in between files to be reviewed.

PR #9300, opened 2021-10-04
fix(route): allow allHeaders call within route

### LINKED ISSUE #9298: [Question]: Should new Headers methods work in RouteAsync ?
### Your question

It seems when I call this new HeadersArrayAsync method like this in RouteAsync
```
await page.RouteAsync("**/*", async (route) =>
{
    var allHeaders= await route.Request.HeadersArrayAsync();
});
```
It throws 
`Microsoft.Playwright.PlaywrightException: 'Unable to fetch raw headers.'`
Are these new methods not supposed to be used in RouteAsync ?

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/9298

### DIFF
--- a/src/server/chromium/crNetworkManager.ts
+++ b/src/server/chromium/crNetworkManager.ts
@@ -154,6 +154,13 @@ export class CRNetworkManager {
   }
 
   _onRequestPaused(workerFrame: frames.Frame | undefined, event: Protocol.Fetch.requestPausedPayload) {
+    if (!event.responseStatusCode && !event.responseErrorReason) {
+      // Request intercepted, deliver signal to the tracker.
+      const request = this._requestIdToRequest.get(event.networkId!);
+      if (request)
+        this._responseExtraInfoTracker.requestPaused(request.request, event);
+    }
+
     if (!this._userRequestInterceptionEnabled && this._protocolRequestInterceptionEnabled) {
       this._client._sendMayFail('Fetch.continueRequest', {
         requestId: event.requestId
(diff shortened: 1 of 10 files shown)


src/server/chromium/crNetworkManager.ts
```diff
@@ -154,6 +154,13 @@ export class CRNetworkManager {
   }
 
   _onRequestPaused(workerFrame: frames.Frame | undefined, event: Protocol.Fetch.requestPausedPayload) {
+    if (!event.responseStatusCode) {
```
yury-s, line 157: && !event.responseErrorReason ? (in case you pause on failed request)

tests/page/page-route.spec.ts
```diff
@@ -672,3 +672,13 @@ it('should support the times parameter with route matching', async ({ page, serv
   await page.goto(server.EMPTY_PAGE);
   expect(intercepted).toHaveLength(1);
 });
+
+it('should contain raw header', async ({ page, server }) => {
+  let headers: any;
+  await page.route('**/*', async route => {
+    headers = await route.request().allHeaders();
```
yury-s, line 679: Check that it actually returns expected headers?

src/server/network.ts
```diff
@@ -153,8 +154,23 @@ export class Request extends SdkObject {
...
+      this._rawRequestHeadersPromise = new ManualPromise();
+  }
+
+  setRawRequestHeaders(headers: types.HeadersArray) {
+    if (!this._rawRequestHeadersPromise)
+      this._rawRequestHeadersPromise = new ManualPromise();
+    this._rawRequestHeadersPromise!.resolve(headers);
+  }
+
+  async rawRequestHeaders(): Promise<NameValue[]> {
```
yury-s, line 168: Please add a comment explaining 3 ways the raw headers could be received

yury-s (approved):

tests/page/page-route.spec.ts
pavelfeldman: I don't think we are returning raw headers here, so I only check that it `accept` one exists.

src/server/network.ts
pavelfeldman: Which 3 ways? :)

PR #9301, opened 2021-10-04
feat(test-runner): allow parametrized tests that don't extend base

pavelfeldman:


types/test.d.ts
```diff
@@ -2408,6 +2408,16 @@ export interface PlaywrightWorkerOptions {
 
 export type VideoMode = 'off' | 'on' | 'retain-on-failure' | 'on-first-retry' | /** deprecated */ 'retry-with-video';
 
+/**
+ * Namespace for additional global use parameters.
```
JoelEinbinder, line 2412: This file is generated, so this code should go into the overrides.

JoelEinbinder:
I am having trouble figuring out exactly what this does. Can you add a test to `tests/playwright-test/types.spec.ts`?

tests/playwright-test/types.spec.ts
```diff
@@ -182,3 +182,29 @@ test('config should allow void/empty options', async ({ runTSC }) => {
...
+    'playwright.config.ts': `
+      const configs: pwt.Config[] = [];
+      configs.push({ use: { myParam: 'text' }});
+    `,
+    'extraUseOptions.d.ts': `
+      export {};
+      declare global {
+        namespace PlaywrightTest {
+          export interface ExtraUseOptions {
+            myParam: string
```
JoelEinbinder, line 197: Does this cause myParam to be mandatory inside use? I think it does. We can convert ExtraUseOptions to an optional type like so:

```ts
type OptionalType = {[x in keyof PlaywrightTest.ExtraUseOptions]?: PlaywrightTest.ExtraUseOptions[x]};
```

JoelEinbinder (approved):

tests/playwright-test/types.spec.ts
pavelfeldman: Probably leave it up to the user? (What if all their tests have this param...)

PR #9326, opened 2021-10-05
fix(fetch): do not allow to call get/post with request

yury-s:
Otherwise the client may pass `post` request into `get` which will cause an error because get request can't have body.

#5999

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -12665,9 +12665,9 @@ export interface FetchRequest {
   /**
    * All responses received through
    * [fetchRequest.fetch(urlOrRequest[, options])](https://playwright.dev/docs/api/class-fetchrequest#fetch-request-fetch),
-   * [fetchRequest.get(urlOrRequest[, options])](https://playwright.dev/docs/api/class-fetchrequest#fetch-request-get),
-   * [fetchRequest.post(urlOrRequest[, options])](https://playwright.dev/docs/api/class-fetchrequest#fetch-request-post) and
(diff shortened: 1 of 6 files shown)


types/types.d.ts
```diff
@@ -12753,10 +12753,10 @@ export interface FetchRequest {
   /**
    * Sends HTTP(S) GET request and returns its response. The method will populate fetch cookies from the context and update
    * context cookies from the response. The method will automatically follow redirects.
-   * @param urlOrRequest Target URL or Request to get all fetch parameters from.
+   * @param urlOrRequest Target URL.
    * @param options
    */
-  get(urlOrRequest: string|Request, options?: {
+  get(urlOrRequest: string, options?: {
```
aslushnikov, line 12759: shall we change the argument name too?

types/types.d.ts
```diff
@@ -12786,10 +12786,10 @@ export interface FetchRequest {
   /**
    * Sends HTTP(S) fetch and returns its response. The method will populate fetch cookies from the context and update context
    * cookies from the response. The method will automatically follow redirects.
-   * @param urlOrRequest Target URL or Request to get all fetch parameters from.
+   * @param urlOrRequest Target URL.
    * @param options
    */
-  post(urlOrRequest: string|Request, options?: {
+  post(urlOrRequest: string, options?: {
```
aslushnikov, line 12792: ditto for the argument name

aslushnikov (approved):

pavelfeldman (approved):

PR #9332, opened 2021-10-06
feat(request): add global request fixture

pavelfeldman:


src/test/index.ts
```diff
@@ -456,7 +456,13 @@ export const test = _baseTest.extend<TestFixtures, WorkerAndFileFixtures>({
     }
     await use(await context.newPage());
   },
+
+  request: async ({ playwright, _combinedContextOptions }, use) => {
+    await use(await playwright.request.newContext(_combinedContextOptions as any));
```
yury-s, line 461: `.dispose()` call is missing. You can also drop `as any` now that #9333 landed.

tests/playwright-test/playwright-test-fixtures.ts
```diff
@@ -194,7 +195,7 @@ type Fixtures = {
 };
 
 const common = base.extend<CommonFixtures>(commonFixtures as any);
-export const test = common.extend<Fixtures>({
+export const test = common.extend<ServerFixtures>(serverFixtures as any).extend<Fixtures>({
```
yury-s, line 198: why do we need this?

pavelfeldman: Because we now run server in playwright tests.

PR #9339, opened 2021-10-06
docs(browsers): fix missing JS snippets

mxschmitt:
Currently there are no snippets in our docs for Node.js, see here: https://playwright.dev/docs/next/browsers/#install-behind-a-firewall-or-a-proxy

### DIFF
--- a/docs/src/browsers.md
+++ b/docs/src/browsers.md
@@ -146,21 +146,6 @@ dotnet tool install -g Microsoft.Playwright.CLI
 
 Playwright can install supported browsers by means of the CLI tool.
 
-```bash js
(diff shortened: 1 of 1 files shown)


docs/src/browsers.md
```diff
@@ -146,62 +146,20 @@ dotnet tool install -g Microsoft.Playwright.CLI
 
 Playwright can install supported browsers by means of the CLI tool.
 
-```bash js
```
mxschmitt, line 149: note: since this paragraph is .NET only, we can remove the snippets.

yury-s (approved):

PR #9376, opened 2021-10-07
fix: support cookies with empty value

yury-s:
Fixes #9367


aslushnikov (approved):

pavelfeldman:
My recollection is that roundtrip storage state in either webkit or firefox resulted in exception.

PR #9432, opened 2021-10-11
chore: add README's for packages

mxschmitt:


JoelEinbinder (approved):
For comparison, the old READMEs said something like

# playwright-chromium
This package contains the Chromium flavor of Playwright.

PR #9477, opened 2021-10-13
feat: mark PlaywrightClient's connection as remote

dgozman:
This makes artifacts work as expected for all remote scenarios.

- Ideally, we would just check `Connection.isRemote()` everywhere, but `connectOverCDP` shares the local connection and is still remote.
- Migrated  `isRemote` bit from `Browser` to `BrowserContext` to accommodate browser-less `launchPersistentContext`.


pavelfeldman (commented):
Seems like a new tricky bit is scattered across several core classes. Is three a way to implement this aspect in a more maintainable manner?

pavelfeldman (approved):

PR #9479, opened 2021-10-13
chore: update to webpack v5.58.2

JoelEinbinder:
This gets rid of annoying chokiar warning.﻿


dgozman (approved):
As long as this works!

PR #9489, opened 2021-10-14
Revert "chore: update to webpack v5.58.2 (microsoft#9479)"

pavelfeldman:


pavelfeldman:
I was getting this: https://stackoverflow.com/questions/43376869/webpack-ots-parsing-error-invalid-version-tag-while-loading-eot-font

PR #9503, opened 2021-10-14
feat(codgen): support positioned clicks in a canvas

### LINKED ISSUE #6737: [Feature] Add a way to include x/y coordinates during codegen
When testing canvas interactions (such as clicking on a virtual button visible in a 2D canvas), it is necessary to include a precise cursor position. For example:

```js
await page.click('canvas', {
  position: {
    x: 50,
    y: 100
  }
});
```

It would be very useful for that if the `codegen` feature was able to optionally add the position to the generated code.
Currently, I think there is no way to do it, and the click always happens in the middle of the canvas element.

mxschmitt:
Seems like its enough for now to only support clicks. Also thought about mouse down/up/move but seems not needed for now.

Fixes #6737

### DIFF
--- a/packages/playwright-core/src/server/supplements/injected/recorder.ts
+++ b/packages/playwright-core/src/server/supplements/injected/recorder.ts
@@ -249,6 +249,7 @@ export class Recorder {
     this._performAction({
       name: 'click',
       selector: this._hoveredModel!.selector,
+      position: positionForEvent(event),
       signals: [],
       button: buttonForEvent(event),
       modifiers: modifiersForEvent(event),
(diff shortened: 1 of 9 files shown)


pavelfeldman (approved):

SeanDuignan82:
Any progress on this feature? We need this for testing applications with leaflet maps. Also, can this be applied to div elements?

SeanDuignan82:
It would be really handy if we had a feature like fixtures for the codegen tool. That way we customise the page or browser contexts with all our desired functionality.

wonhyeongseo:
Hello @mxschmitt, may I please know how to use this feature in a codegen session?

mxschmitt:
> Hello @mxschmitt, may I please know how to use this feature in a codegen session?

Just click on a canvas and it should generate custom click statements with x and y included.

PR #9529, opened 2021-10-15
fix: separate Selectors instance per Playwright object

dgozman:
We used to share selectors for inprocess, remote and any internal playwright objects.


toiletrollen:
9529

pavelfeldman (approved):

PR #9535, opened 2021-10-15
(feat): Composition support for keyboard.imeSetComposition

trueadm:
Notes:
- Based off the work from https://github.com/microsoft/playwright/pull/7487. 
- Related feature thread: #5777
- Chromium changes have already been merged: https://chromium-review.googlesource.com/c/chromium/src/+/3009937

This PR implements the keyboard composition API `keyboard.imeSetComposition`. This makes it possible to test low-level IME related composition interactions.

Example:

```js
  await page.focus('textarea');
  await page.keyboard.imeSetComposition('ｓ', 1, 1);
  await page.keyboard.imeSetComposition('す', 1, 1);
  await page.keyboard.imeSetComposition('すｓ', 2, 2);
  await page.keyboard.imeSetComposition('すｓｈ', 3, 3);
  await page.keyboard.imeSetComposition('すし', 2, 2);
  await page.keyboard.insertText('すし');
```

You might be wondering what, what about triggering of keyboard events such as `keyboard.up`, `keyboard.press`, `keyboard.down` etc, Rather than couple this PR with other discussions (see https://github.com/microsoft/playwright/pull/7659) I think this can be worked on separately. Having `imeSetComposition` available sooner will be invaluable right now and help inform us of the discussions around how key events should work, and what their API should be.

### DIFF
--- a/docs/src/api/class-keyboard.md
+++ b/docs/src/api/class-keyboard.md
@@ -172,6 +172,71 @@ Modifier keys DO influence `keyboard.down`. Holding down `Shift` will type the t
 
 Name of the key to press or a character to generate, such as `ArrowLeft` or `a`.
 
+## async method: Keyboard.imeSetComposition
+
+If there is an active composition, it will update the composition. Otherwise it will start a new one. It will dispatch either
+a `compositionstart` or a `compositionupdate`. In order to cancel a composition: `keyboard.insertText('')`, and in order to commit/end a composition:
+`keyboard.insertText(<text to insert>)`.
+
+### param: Keyboard.imeSetComposition.text
+- `text` <[string]>
+
+Sets the text in the active composition
+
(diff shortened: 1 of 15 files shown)


trueadm:
@JoelEinbinder I've added Firefox support via some changes in `browser_patches`, but I was unsure how updating browser patches works. I wasn't able to get Webkit to build locally, so I wasn't able to take a stab at adding support for that. I'd be happy to catch up over VC and discuss the changes and maybe look into adding support to webkit together if you have time in the coming weeks.

pavelfeldman:
Closing as stale due to no progress (on our side). I'm not sure the action for this should be, but it seems like we aren't ready for ime API at this point.

trueadm:
@pavelfeldman Has there been any further interest in adding the IME/Composition API internally. I was of the understanding that this was regarded as important from the Microsoft Editor folks?

PR #9542, opened 2021-10-15
Fix typo: afterEach runs AFTER not before

pbiggar:
This looks like a typo in the docs. I assume that afterEach runs after each test. If this is intentionally "before", then some docs to clarify that would be helpful instead.

### DIFF
--- a/docs/src/test-api/class-test.md
+++ b/docs/src/test-api/class-test.md
@@ -72,7 +72,7 @@ Hook function that takes one or two arguments: an object with fixtures and optio
 
 ## method: Test.afterEach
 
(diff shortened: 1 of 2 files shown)


mxschmitt:
Hey! Thank you for your contribution. The file which you've edited is generated, so you need to change it in the source [here](https://github.com/microsoft/playwright/blob/0801a8c486d994b71aa3a525139241c5cfa246cf/docs/src/test-api/class-test.md) and run "npm ci" and "npm run build" locally to update it accordingly.

pbiggar:
@mxschmitt - thanks! Ready for review now.

PR #9546, opened 2021-10-15
feat(input): a mode that performs hit target check during input

### LINKED ISSUE #6238: [BUG] Wrong element is clicked due to lazy loading/UI shift
**Context:**

- Playwright Version: v1.10.0
- Operating System: Mac
- Node.js version: v15.4.0
- Browser: Webkit
- Extra: Docker image - v1.10.0-focal

**Describe the bug**
Scenario - Click "Contact Agent" button.

Issue - I am trying to click "Contact Agent" button on the page but there is a section "Popular Search" that loads later and shifts the UI. While the test is trying to click the button, the UI changes and instead it ends up clicking one of the popular searches. Please see the screenshot below captured using PWDEBUG=1 - the button is found and highlighted but pointer is on one of the searches.

webkit_pointer

dgozman:
Reincarnation of #6352.

This change replaces previous `checkHitTarget` heuristic that took place before the action with a new `setupHitTargetInterceptor` that works during the action:
- Before the action we set up capturing listeners on the window.
- During the action we ensure that event target is the element we expect to interact with.
- After the action we clear the listeners.

This should catch the "layout shift" issues where things move between action point calculation and the actual action.

Possible issues:
- **Risk:** `{ trial: true }` might dispatch move events like `mousemove` or `pointerout`,
because we do actually move the mouse but prevent all other events.
- **Timing**: The timing of "hit target check" has moved, so this may affect different web pages
in different ways, for example expose more races. In this case, we should retry the click as before.
- **No risk**: There is still a possibility of mis-targeting with iframes shifting around,
because we only intercept in the target frame. This behavior does not change.

There is an **opt-out** environment variable `PLAYWRIGHT_NO_LAYOUT_SHIFT_CHECK` that reverts to previous behavior.

Fixes #6238.

### DIFF
--- a/packages/playwright-core/src/server/dom.ts
+++ b/packages/playwright-core/src/server/dom.ts
@@ -19,7 +19,7 @@ import * as injectedScriptSource from '../generated/injectedScriptSource';
 import * as channels from '../protocol/channels';
 import { isSessionClosedError } from './common/protocolError';
 import * as frames from './frames';
-import type { InjectedScript, InjectedScriptPoll, LogEntry } from './injected/injectedScript';
+import type { InjectedScript, InjectedScriptPoll, LogEntry, HitTargetInterceptionResult } from './injected/injectedScript';
 import { CallMetadata } from './instrumentation';
 import * as js from './javascript';
 import { Page } from './page';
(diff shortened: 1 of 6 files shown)


pavelfeldman (approved):
Ship it!

PR #9583, opened 2021-10-18
chore: remove codicon from html report

pavelfeldman:


mxschmitt (approved):
Lets have a `<Icon variant="foobar..."/>` component instead? then the files does not get polluted with SVG.

PR #9584, opened 2021-10-18
fix(monorepo): explicitly list package.json exports instead of using glob patterns

JoelEinbinder:
In new versions of node, using `"./": "./"` in the package.json field is deprecated.

In old versions of node 12, using `"./src/**/*": "./lib/**/*.js"` is unsupported. Technically
we don't support old node 12, only current LTS node 12, but enough people seem to be using it
that I'd rather not break.

This patch lists the files that `@playwright/test` and `playwright-webkit` require from playwright-core.
They are:
- ./lib/grid/gridClient.js
- ./lib/utils/async.js
- ./lib/utils/httpServer.js
- ./lib/utils/processLauncher.js
- ./lib/utils/registry.js
- ./lib/utils/utils.js

Other files that are required by tests instead require the file directly like `../packages/playwright-core/lib/foo/bar`.

We had some tests that were testing requiring `./lib/utils/errors` and `./lib/server/deviceDescriptors`. These have
not been supported apis since the Puppeteer days, so I modified the tests.

### DIFF
--- a/install-from-github.js
+++ b/install-from-github.js
@@ -31,7 +31,7 @@ try {
 }
 
 console.log(`Downloading browsers...`);
-const { installDefaultBrowsersForNpmInstall } = require('./packages/playwright-core/lib/utils/registry');
+const { installDefaultBrowsersForNpmInstall } = require('playwright-core/src/utils/registry');
 installDefaultBrowsersForNpmInstall().catch(e => {
   console.error(`Failed to install browsers, caused by\n${e.stack}`);
   process.exit(1);
--- a/installation-tests/driver-client.js
+++ b/installation-tests/driver-client.js
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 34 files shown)


mxschmitt (approved):
~~It feels a little bit scary to me, that we sometimes import relative (..) and sometimes from playwright-core/src/*. Is there some validation - probably if we do top-level require/imports it will automatically yell at us if we do it wrong (e.g. import relative inside the src of pw-test).~~
otherwise LGTM.

JoelEinbinder:
> It feels a little bit scary to me, that we sometimes import relative (..) and sometimes from playwright-core/src/*. Is there some validation - probably if we do top-level require/imports it will automatically yell at us if we do it wrong (e.g. import relative inside the src of pw-test). otherwise LGTM.

The relative imports are just for tests, so we don't have to export all of the code we want to unit test. You can't do it wrong because it will throw an error message :)

PR #9597, opened 2021-10-18
feat: introduce `npx playwright install docker-image` command

aslushnikov:
This command pulls matching docker image for Playwright.


pavelfeldman (changes requested):
Looks like users can do that manually for now?

pavelfeldman (approved):

PR #9619, opened 2021-10-19
chore(test-runner): use test name as trace file name prefix

### LINKED ISSUE #9569: [Feature] Provide an ability for Playwright to configure the file names of the generate trace-viewer files to match the name of the test
Would it be possible to always use the test name as the file name of the trace.zip file?

Currently we upload numerous `trace.zip` files as build artefacts in our CI but when we download these files from the build server, we have no easy way of distinguishing which `trace.zip` belongs to which test?  It would be nice if we had a way to configure Playwright to generate the trace zip files names to match the actual test it ran against.

#trace-viewer #ux
Playwright 1.15

mxschmitt:
Fixes #9569

### DIFF
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -22,6 +22,7 @@ import { rootTestType } from './testType';
 import { createGuid, removeFolders } from 'playwright-core/lib/utils/utils';
 import { GridClient } from 'playwright-core/lib/grid/gridClient';
 import { Browser } from 'playwright-core';
+import { sanitizeForFilePath } from './util';
 export { expect } from './expect';
 export const _baseTest: TestType<{}, {}> = rootTestType.test;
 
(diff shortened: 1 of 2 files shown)


JoelEinbinder (approved):

dgozman:
What's the status of this one?

packages/playwright-test/src/index.ts
```diff
@@ -254,7 +255,7 @@ export const test = _baseTest.extend<TestFixtures, WorkerAndFileFixtures>({
 
     const traceAttachments: string[] = [];
     const addTraceAttachment = () => {
-      const tracePath = testInfo.outputPath(`trace${traceAttachments.length ? '-' + traceAttachments.length : ''}.zip`);
+      const tracePath = testInfo.outputPath(`${sanitizeForFilePath(testInfo.title)}${traceAttachments.length ? '-' + traceAttachments.length : ''}-trace.zip`);
```
dgozman, line 258: Let's do `trace-my-test.zip` instead? WDYT?

dgozman (approved):

PR #9620, opened 2021-10-19
fix(monorepo): support yarn2/pnp by specifying all deps and bins

JoelEinbinder:
`@playwright/test` was using some dependencies from `playwright-core` without explicitly depending on them itself. This doesn't work with yarn2 / pnp.

Also you can't use the `playwright` command with yarn2 unless you explictly depend on `playwright-core`. This is fixed by adding a `cli.js` file to each of our packages which forwards to the `playwright-core` `cli.js`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8183,9 +8183,9 @@
       "dev": true
     },
     "node_modules/stack-utils": {
-      "version": "2.0.3",
-      "resolved": "https://registry.npmjs.org/stack-utils/-/stack-utils-2.0.3.tgz",
-      "integrity": "sha512-gL//fkxfWUsIlFL2Tl42Cl6+HFALEaB1FU76I/Fy+oZjRreP7OPMXFlGbxM7NQsI0ZpUfw76sHnv0WNYuTb7Iw==",
+      "version": "2.0.5",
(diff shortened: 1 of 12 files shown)


pavelfeldman (approved):

FezVrasta:
I'm getting this when trying Playwright with mjs modules 

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@playwright/test' imported from /Users/federicozivolo/Developer/popper/popper-core/packages/dom/tests/functional/arrow.test.mjs
Did you mean to import @playwright-test-npm-1.16.3-3b65662da1-363f2f214d.zip/node_modules/@playwright/test/index.js?
```

mxschmitt:
> I'm getting this when trying Playwright with mjs modules and pnp
> 
> ```
> Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@playwright/test' imported from /Users/federicozivolo/Developer/popper/popper-core/packages/dom/tests/functional/arrow.test.mjs
> Did you mean to import @playwright-test-npm-1.16.3-3b65662da1-363f2f214d.zip/node_modules/@playwright/test/index.js?
> ```
> 
> `@playwright/test` is a devDependency

Please file a new GitHub issue, otherwise it would get lost.

PR #9629, opened 2021-10-19
feat(test runner): update scheduling logic to avoid unneeded stalls

dgozman:
In some circumstances, dispatcher was waiting for all existing jobs
to finish before scheduling a new one. This leads to unneeded stalls.
Instead, we can schedule jobs right away, if we have a worker
available.

Fixes #9609.


JoelEinbinder (approved):
Well I had no idea how the old code worked, and I have some idea how the new code works. So I say its great!

pavelfeldman (approved):
I can try following up with niceness!

PR #9665, opened 2021-10-20
feat(test runner): async describe

dgozman:


JoelEinbinder (approved):
yay!

JoelEinbinder:
> Nay! We are encouraging computation in the runner and in the worker processes, where things are supposed to be fast/declarative-ish.

This allows using new versions of `node-fetch` via `await import` without converting all of your tests to mjs.

PR #9723, opened 2021-10-22
docs(api): add locator assertions for java

yury-s:


dgozman (approved):
I wonder whether we should merge this doc and a similar doc for js expect.

PR #9767, opened 2021-10-25
docs(proxy): note that browser-wide proxy is only required for Cr+Win

dgozman:


mxschmitt (approved):
People think "per-context" is a constant. Was also thinking to rephrase it even more obvious. In the past 5-10 issues/Slack users reported it.

Issue #9733, opened 2021-10-23
[BUG] the config is TestProject?


KennethKinLum:
**Context:**
- Playwright Version: [what Playwright version do you use?]
1.16.0

- Operating System: [e.g. Windows, Linux or Mac]
Mac

This is about the docs.  I am reading https://playwright.dev/docs/test-assertions/#expectlocatortohavetextexpected-options

and so it looks like we can use the `options` as  `{ timeout: 20000 }` and that's clear enough.

Maybe I am missing something.  When I look at:

> defaults to timeout in testProject.expect.

and then https://playwright.dev/docs/api/class-testproject#test-project-expect

> TestProject encapsulates configuration specific to a single project.

and then the code below the line above does not mention `TestProject` whatsoever.
And below, it mentioned `testProject.expect`

So I tried:

```js
// playwright.config.js

module.exports = {
  testProject: {
    expect: {
      timeout: 20000,
    },
  },
};
```
and it didn't work.

and so I had to go by try and error and used:

```js
// playwright.config.js

module.exports = {
    expect: {
      timeout: 20000,
    },
};
```

and it worked.  So the object here (sometimes named `config`) is `TestProject`?  Can you just say that?  This is one case that I can read the docs for 10 minutes and I still didn't know what it was talking about and how to do it exactly.

Or maybe you can use

```js
// playwright.config.js

const testProject = {
    expect: {
      timeout: 20000,
    },
};

module.exports = testProject;
```
and it is crystal clear.  I wonder why often the docs don't try to make it so clear.

dgozman:
@KennethKinLum Documentation is hard! 😄 

If you go to https://playwright.dev/docs/api/class-testproject/, it says:
```md
Note that all properties of TestProject are available in the top-level TestConfig, in which case they are shared between all projects.
```

I can totally understand how this could be confusing. If you have any specific suggestions about how to improve things, that would be great! Anyway, we always try to improve the documentation to make things clear.

PR #9770, opened 2021-10-26
Making testProject more clear

KennethKinLum:
as per https://github.com/microsoft/playwright/issues/9733#issuecomment-951428289
If it is testProject, why not name it `testProject`?  Why name it `config` to add an extra confusion?


Adelgadillo711 (approved):
https://github.com/Adelgadillo711/supreme-octo-bassoon/commits/#tweet

PR #9772, opened 2021-10-26
add the general rule to expect the opposite of the matchers

KennethKinLum:


dgozman (commented):
Thank you for the PR! One small nit and this is good to go.

Issue #9720, opened 2021-10-22
[BUG] Move Windows Channel installation over to Node.js instead of PowerShell


mxschmitt:
PowerShell requires execution policy to be configured.

> Set-ExecutionPolicy -Scope Process -ExecutionPolicy Bypass

is not
digitally signed. You cannot run this script on the current system. For more information about running scripts and
setting execution policy, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170.
    + CategoryInfo          : SecurityError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : UnauthorizedAccess
Failed to install browsers

kokushkin:
Facing the same issue. It seems I can't override group polices to run the script. But what's the reason of not signing the scripts?

PR #9878, opened 2021-10-29
chore: move Chromium channel install scripts from ps1 to Node.js

mxschmitt:
- ProductInfo on Windows does not get displayed anymore since we can't get it easily from my research
- Moved the download from MS Edge Bash/PowerShell scripts over to the Node.js land
- Moved PowerShell scripts into Node.js land
- We should probably think about evaluating our privileges if the script gets run by a normal user

https://github.com/microsoft/playwright/issues/9720


aslushnikov (commented):
high-level comment: let's move all our scripts to js then?

PR #9880, opened 2021-10-29
allow to return values from inside test.step

### LINKED ISSUE #9742: [Feature] Allow test.step to return the value resolved in the test callback
Right now `test.step` resolves to `undefined`.

It would be great if it resolved to the value the callback resolves to, so we could have tests like this:

```ts
test('test something', async ({ pageObject, page }) => {
  await pageObject.navigateTo();
  
  const link = await test.step('is able to retrieve a link', async () => {
    const link = await pageObject.getLink();
    
    // tests...
    expect(link).toBe(blabla);
    
    return link;
  });
  
  await test.step('link is accessible', async () => {
    await page.goto(link)
    // tests...
    expect(page.locator('#my-element')).toBeVisible();
  });
})
```

Right now the workaround is to have `let` variables before the step (notice the `!` to avoid type errors):

```ts
test('test something', async ({ pageObject, page }) => {
  await pageObject.navigateTo();
  
  let link!: string
  await test.step('is able to retrieve a link', async () => {
    link = await pageObject.getLink();
    
    // tests...
    expect(link).toBe(blabla);
  });
  
  await test.step('link is accessible', async () => {
    await page.goto(link)
    // tests...
    expect(page.locator('#my-element')).toBeVisible();
  });
})
```

I'm free to work on this if it is something that would be accepted in the library.

JCMais:
Closes #9742

The test.step signature has been changed from:

```ts
step(title: string, body: () => Promise<any>): Promise<any>;
```

to:

```ts
step<T extends any = any>(title: string, body: () => Promise<T>): Promise<T>;
```

This is a backward-compatible change.

It allows the developer to use the return value from the `test.step` if they want to. This improves the DX in case you want to return some values from a `test.step` to use it again inside the current test.

### DIFF
--- a/docs/src/test-api/class-test.md
+++ b/docs/src/test-api/class-test.md
@@ -975,6 +975,32 @@ test('test', async ({ page }) => {
 });
 ```
 
+The `test.step` promise resolves to the value returned in the `body`:
+
+```js js-flavor=js
+const { test, expect } = require('@playwright/test');
+
+test('test', async ({ page }) => {
+  const result = await test.step('Log in', async () => {
+    // ...
+    return { a: true }
+  });
+  // result is { a: true }
+});
+```
+
+```js js-flavor=ts
+import { test, expect } from '@playwright/test';
+
+test('test', async ({ page }) => {
+  const result = await test.step('Log in', async () => {
+    // ...
+    return { a: true }
+  });
+  // result is { a: true }
+});
+```
+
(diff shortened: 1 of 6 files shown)


dgozman:
@JCMais Thank you for the PR. As I have explained in #9742, we are not yet ready for this API. Let me close the issue, and we can reopen when/if we decide to implement this.

PR #9892, opened 2021-10-29
feat: add "commit" to possible waitUntil options

yury-s:
https://github.com/microsoft/playwright-java/issues/678

### DIFF
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -453,7 +453,8 @@ export class Frame extends SdkObject {
     if (this._parentFrame)
       this._parentFrame._childFrames.add(this);
 
-    this._onLifecycleEvent('commit');
+    this._firedLifecycleEvents.add('commit');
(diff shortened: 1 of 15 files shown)


packages/playwright-core/src/server/frames.ts
```diff
@@ -471,6 +471,7 @@ export class Frame extends SdkObject {
     this._stopNetworkIdleTimer();
     if (this._inflightRequests.size === 0)
       this._startNetworkIdleTimer();
+    this._onLifecycleEvent('commit');
```
dgozman, line 474: Should we also do this once in the constructor?

dgozman (approved):

packages/playwright-core/src/server/frames.ts
yury-s: It already works on initial url. Added a test.

yury-s: As discussed offline fire 'commit' event in the constructor just in case.

PR #9893, opened 2021-10-29
feat(trace): throttle the screencast

pavelfeldman:


packages/playwright-core/src/server/page.ts
```diff
@@ -495,6 +496,16 @@ export class Page extends SdkObject {
 
   setScreencastOptions(options: { width: number, height: number, quality: number } | null) {
     this._delegate.setScreencastOptions(options).catch(e => debugLogger.log('error', e));
+    this._frameThrottler.setEnabled(!!options);
```
dgozman, line 499: Let's also cleanup (setEnabled(false)) when the page is closed, to avoid setInterval stuck forever.

packages/playwright-core/src/server/page.ts
```diff
@@ -631,3 +642,57 @@ function addPageBinding(bindingName: string, needsHandle: boolean) {
...
+  recharge() {
+    // Send all acks, reset budget.
+    for (const ack of this._acks)
+      ack();
+    this._acks = [];
+    this._budget = this._nonThrottledFrames;
+  }
+
+  ack(ack: () => void) {
+    // Either not engaged or video is also recording, don't throttle.
```
dgozman, line 679: I didn't get where we disable this when video is recording.

packages/playwright-core/src/server/page.ts
```diff
@@ -631,3 +642,57 @@ function addPageBinding(bindingName: string, needsHandle: boolean) {
...
+    if (!this._intervalId) {
+      ack();
+      return;
+    }
+
+    // Do we have enough budget to respond w/o throttling?
+    if (--this._budget > 0) {
```
dgozman, line 686: I don't see where we do `this._budget++`.

dgozman (approved):

packages/playwright-core/src/server/page.ts
pavelfeldman: We don't!

PR #9910, opened 2021-10-31
feat(html reporter): open test details page when running a single test

dgozman:


packages/playwright-test/src/reporters/html.ts
```diff
@@ -292,7 +298,9 @@ class HtmlBuilder {
       }
     }
 
-    return ok;
+    if (total !== 1)
```
pavelfeldman, line 301:
```suggestion
    const singleTestId = htmlReport.stats.total === 1 ? data.values().next().value.testFile.tests[0].testId : undefined;
    return { ok, singleTestId };
```

pavelfeldman (approved):

PR #9939, opened 2021-11-01
feat(html): bake report zip into the html report, allow opening from fs

pavelfeldman:


packages/playwright-test/src/reporters/html.ts
```diff
@@ -294,14 +292,31 @@ class HtmlBuilder {
       }
     }
 
+    // Inline report data.
+    const indexFile = path.join(this._reportFolder, 'index.html');
+    fs.appendFileSync(indexFile, '<script>\nwindow.playwrightReportBase64 = "data:application/zip;base64,');
```
dgozman, line 297: Should we `this._dataZipFile.on('error')` and print something nice?

dgozman (approved):

packages/playwright-test/src/reporters/html.ts
pavelfeldman: Let's do it when we hit it!

PR #9978, opened 2021-11-02
docs(java): message loop, Thread.sleep vs waitForTimeot

yury-s:
https://github.com/microsoft/playwright-java/issues/672

### DIFF
--- a/docs/src/threading-java.md
+++ b/docs/src/threading-java.md
@@ -56,3 +56,24 @@ public class PlaywrightThread extends Thread {
   }
 }
 ```
+
+## Synchronous API and event dispatching
+
+In the synchronous Playwright API all events are dispatched only when Playwright is running its message loop.
+This happens automatically when you call any of the API methods and doesn't happen if there are no active
+Playwright calls on the stack. If you need to wait for an event the best way to do this is via one of the
+`waitFor*` methods.
+
+### Page.waitForTimeout() vs. Thread.sleep()
+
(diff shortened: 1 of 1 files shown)


docs/src/threading-java.md
```diff
@@ -55,4 +55,25 @@ public class PlaywrightThread extends Thread {
...
+In the synchronous Playwright API all events are dispatched only when Playwright is running its message loop.
+This happens automatically when you call any of the API methods and doesn't happen if there are no active
+Playwright calls on the stack. If you need to wait for an event the best way to do this is via one of the
+`waitFor*` methods.
+
+### Page.waitForTimeout() vs. Thread.sleep()
+
+One consequence of the synchronous API is that if you for whatever reason call `Thread.sleep()` no events will
+be fired while the thread is sleeping. If you want events from the browser to be dispatched while the program
+execution is paused use [`method: Page.waitForTimeout`] or [`method: Frame.waitForTimeout`]:
```
mxschmitt, line 71: We have the same problem in Python and probably also in .NET: https://playwright.dev/python/docs/intro#timesleep-leads-to-outdated-state

mxschmitt (approved):

docs/src/threading-java.md
yury-s: Do you suggest moving this to the intro section? I'd rather keep that more concise and focused on the initial experience but perhaps we should restructure it.

Issue #9847, opened 2021-10-28
[Question] Why in CI there is no line with summary info about tests and workers?


krutoo:
Hi, first of all thanks for great testing framework

In my project i use Playwright in GitLab CI with recommender Docker image mcr.microsoft.com/playwright:focal

I ran my tests by npm-script:
```json
{
  "scripts": {
    "test:e2e": "playwright test" 
  }
}
```

Locally after run my tests i have summary info line in terminal like:
```bash
# ...
Running 105 tests using 2 workers
# ...
```

But i don't have this line in CI job log

What i am doing wrong or how can i turn it info on in CI jobs?

krutoo:
> I guess you'd like to see this line on CI?

@dgozman Yes, would be very helpful

mhatvan:
I will +1 this one, right now I am unsure if my tests got stuck or they just take a while ;)

PR #9992, opened 2021-11-02
fix(reporters): move more terminal output to terminal reporters

dgozman:
This makes 'Running X tests using Y workers' visible on CI as long as you specify terminal reporter, and generally makes the code cleaner.

Drive-by: refactor github repoter to not inherit from base, since it is not a terminal reporter.

Note: `base.ts` is just a code except for the `onBegin` method.
Note: tests changed because we now print 'Running X tests' in the tests no matter the `stdout.isTTY`.

References #9847.

### DIFF
--- a/packages/playwright-test/src/reporters/base.ts
+++ b/packages/playwright-test/src/reporters/base.ts
@@ -58,6 +58,7 @@ export class BaseReporter implements Reporter  {
   duration = 0;
   config!: FullConfig;
   suite!: Suite;
+  totalTestCount = 0;
   result!: FullResult;
   fileDurations = new Map<string, number>();
   monotonicStartTime: number = 0;
@@ -72,6 +73,10 @@ export class BaseReporter implements Reporter  {
     this.monotonicStartTime = monotonicTime();
     this.config = config;
     this.suite = suite;
+    this.totalTestCount = suite.allTests().length;
+    const jobs = Math.min(config.workers, (config as any).__testGroupsCount);
+    const shardDetails = config.shard ? `, shard ${config.shard.current} of ${config.shard.total}` : '';
+    console.log(`\nRunning ${this.totalTestCount} test${this.totalTestCount > 1 ? 's' : ''} using ${jobs} worker${jobs > 1 ? 's' : ''}${shardDetails}\n`);
   }
 
   onStdOut(chunk: string | Buffer, test?: TestCase, result?: TestResult) {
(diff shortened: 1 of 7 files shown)


packages/playwright-test/src/reporters/base.ts
```diff
@@ -72,6 +73,10 @@ export class BaseReporter implements Reporter  {
     this.monotonicStartTime = monotonicTime();
     this.config = config;
     this.suite = suite;
+    this.totalTestCount = suite.allTests().length;
+    const jobs = Math.min(config.workers, (config as any).__testGroupsCount);
+    const shardDetails = config.shard ? `, shard ${config.shard.current} of ${config.shard.total}` : '';
+    console.log(`\nRunning ${this.totalTestCount} test${this.totalTestCount > 1 ? 's' : ''} using ${jobs} worker${jobs > 1 ? 's' : ''}${shardDetails}\n`);
```
pavelfeldman, line 79: If this is now a terminal reporter, we should rename it to the terminal.ts and maybe extract all the utility functions that are used from non-terminal reporters into base.ts?

packages/playwright-test/src/reporters/base.ts
```diff
@@ -116,95 +121,80 @@ export class BaseReporter implements Reporter  {
...
-      failuresToPrint
-    };
-  }
-
-  epilogue(full: boolean) {
-    const summary = this.generateSummary();
-    const summaryMessage = this.generateSummaryMessage(summary);
-    if (full && summary.failuresToPrint.length && !this._omitFailures)
-      this._printFailures(summary.failuresToPrint);
-    this._printSlowTests();
```
pavelfeldman, line 178: Looks like it was more readable before? Let's bring it back.

packages/playwright-test/src/runner.ts
```diff
@@ -294,13 +294,7 @@ export class Runner {
         filterSuite(rootSuite, () => false, test => shardTests.has(test));
         total = rootSuite.allTests().length;
       }
-
-      if (process.stdout.isTTY) {
-        console.log();
-        const jobs = Math.min(config.workers, testGroups.length);
-        const shardDetails = shard ? `, shard ${shard.current} of ${shard.total}` : '';
-        console.log(`Running ${total} test${total > 1 ? 's' : ''} using ${jobs} worker${jobs > 1 ? 's' : ''}${shardDetails}`);
```
pavelfeldman, line 302: This was deliberately outside of the scope of the reporter. Why do we want to differentiate terminal and non-terminal reporters? Do we also want to differentiate stdio vs terminal/tty?

PR #9998, opened 2021-11-03
chore: introduce docker integration dogfood

aslushnikov:
This patch:
- rolls stable-test-runner to Nov 2, 2021 tip-of-tree
- introduces a new npm script, `npm run vtest`, to run Visual Regression
  Tests for our HTML reporter
- adds a single screenshot expectation for html reporter spec.

### DIFF
--- a/.github/workflows/tests_primary.yml
+++ b/.github/workflows/tests_primary.yml
@@ -71,3 +71,25 @@ jobs:
       if: matrix.os != 'ubuntu-latest'
     - run: xvfb-run npm run ttest
       if: matrix.os == 'ubuntu-latest'
+
+  test_runner_vrt:
+    name: Test Runner Visual Regression Testing
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 12
+    - run: npm i -g npm@7
+    - run: npm ci
+      env:
+        DEBUG: pw:install
+    - run: npm run build
+    - run: npx playwright install --with-deps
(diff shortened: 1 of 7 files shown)


tests/playwright-test/playwright-test.config.ts
```diff
@@ -22,9 +22,11 @@ const config: Config = {
   timeout: 30000,
   forbidOnly: !!process.env.CI,
   preserveOutput: process.env.CI ? 'failures-only' : 'always',
-  projects: [
-    { name: 'playwright-test' },
-  ]
+  projects: process.env.PLAYWRIGHT_DOCKER ? [
+    { name: 'vrt', testMatch: ['reporter-html.spec.ts'] },
```
dgozman, line 26: I'd instead use `foo.spec.ts` vs `foo.visual.ts` convention or something like that. You don't want to hardcode file names here.

dgozman (approved):

.github/workflows/tests_primary.yml
```diff
@@ -71,3 +71,20 @@ jobs:
...
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 12
+    - run: npm i -g npm@7
+    - run: npm ci
+      env:
+        DEBUG: pw:install
+    - run: npm run build
+    - run: npx playwright install --with-deps
```
mxschmitt, line 88: Do we need browsers and deps here since it's in docker iiuc?

aslushnikov: good point, probably not

tests/playwright-test/playwright-test.config.ts
aslushnikov: done

PR #10016, opened 2021-11-03
feat(test-runner): have type-based expect matchers

mxschmitt:
before: 

image
image

after: 
image
image


packages/playwright-test/src/matchers/matchers.ts
```diff
@@ -224,6 +224,18 @@ export function toHaveText(
   }
 }
 
+export function toHaveValue(
```
mxschmitt, line 227: note: re-ordered so the location matchers come first, then the page matchers.

PR #10038, opened 2021-11-04
fix(test-runner): tracing.stopChunk: Must start tracing before stopping

### LINKED ISSUE #9879: [BUG] tracing.stopChunk: Must start tracing before stopping
**Context:**
- Playwright Version: "@playwright/test": "^1.16.2"
- Node.js version: v14.17.4
- Browser: all

**Code Snippet**

```shell
  2) [chromium] › payment-flow.spec.js:152:1 › delete project ======================================

    tracing.stopChunk: Must start tracing before stopping
```

```shell
 4) [firefox] › payment-flow.spec.js:152:1 › delete project =======================================

    tracing.stopChunk: Must start tracing before stopping
```

```shell
  6) [webkit] › payment-flow.spec.js:152:1 › delete project ========================================

    tracing.stopChunk: Target page, context or browser has been closed
```

**Describe the bug**
Tracing does not work. I'm not doing anything special, it just fails after running some failed tests before.
It does not even start to actually run the test on which the error is thrown, except for the beforeEach part (this test is supposed to be skipped). The error message is different on webkit.
=> I don't have any trace available after running the tests

mxschmitt:
Not sure if this only hides the actual underlying error, since we do it the same way for screenshots it seems alright.

Fixes #9879

### DIFF
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -224,7 +224,7 @@ export const test = _baseTest.extend<TestFixtures, WorkerAndFileFixtures>({
         // after the test finishes.
         const tracePath = path.join(_artifactsDir(), createGuid() + '.zip');
         temporaryTraceFiles.push(tracePath);
-        await context.tracing.stopChunk({ path: tracePath });
+        await context.tracing.stopChunk({ path: tracePath }).catch(() => { });
       }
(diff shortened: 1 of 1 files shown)


dgozman:
> Not sure if this only hides the actual underlying error, since we do it the same way for screenshots it seems alright.

Yeah, I'd prefer to fix the underlying issue.

PR #10055, opened 2021-11-04
fix(trace-viewer): redirect, time and missing snapshot bugfixes

pavelfeldman:


packages/playwright-core/src/web/traceViewer/snapshotRenderer.ts
```diff
@@ -80,9 +80,10 @@ export class SnapshotRenderer {
     if (!html)
       return { html: '', pageId: snapshot.pageId, frameId: snapshot.frameId, index: this._index };
 
-    if (snapshot.doctype)
-      html = `<!DOCTYPE ${snapshot.doctype}>` + html;
-    html += `
+    // Hide the document in order to prevent flickering. We will unhide once script has processed shadow.
+    const hideAllStyle = '<style>*,*::before,*::after { visibility: hidden }</style>';
+    const prefix = snapshot.doctype ? `<!DOCTYPE ${snapshot.doctype}>` + hideAllStyle : hideAllStyle;
+    html = prefix + html + `
```
dgozman, line 86: I am a bit worried that we'll mess up with head/meta/etc.

dgozman (approved):

Issue #7982, opened 2021-08-04
[Feature]: Iframe selector that pierces iframes


uchagani:
### Feature request

Similar to how playwright pierces shadow DOMs, it would be great to have an `:iframe` selector that can do the same thing.  This would allow us to write code that interacts with elements inside iframes in a similar manner.

For example, if we have a document that looks like this:

```html
<html>
   <head></head>
   <body>
      <div id="outside-iframe"></div>
      <iframe id="frame1">
         #document
         <html>
            <head></head>
            <body>
               <div class="inside-iframe">This is the Element I want</div>
            </body>
         </html>
      </iframe>
   </body>
</html>
```

Maybe we can have a selector called `:iframe` to access element inside the iframe:

```js
page.textContext(":iframe(#frame1 div.inside-iframe)");
```

the first parameter inside the selector is how to find the iframe. 

For nested iframes maybe we an nest the selector like this:

```js
page.textContent(":iframe(#frame1 :iframe(#nestedIFrame #someElementInsideNestedIframe))")
```

splurgeop:
would be good to have this feature in playwright.

unlikelyzero:
This would be a key differentiator in the e2e testing world.

aslushnikov:
this is fixed!

PR #10063, opened 2021-11-04
feat(iframe): make iframe selectors work w/ element handles

pavelfeldman:
#7982

### DIFF
--- a/packages/playwright-core/src/server/dom.ts
+++ b/packages/playwright-core/src/server/dom.ts
@@ -691,48 +691,36 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
         this._page._timeoutSettings.timeout(options));
   }
 
-  async querySelector(metadata: CallMetadata, selector: string, options: types.StrictOptions): Promise<ElementHandle | null> {
-    const controller = new ProgressController(metadata, this);
-    return controller.run(async progress => {
-      const { frame, info } = await this._frame.resolveFrameForSelector(progress, selector, options, this);
-      // If we end up in the same frame => use the scope again, live above was noop.
-      return this._page.selectors.query(frame, info, this._frame === frame ? this : undefined);
-    });
+  async querySelector(selector: string, options: types.StrictOptions): Promise<ElementHandle | null> {
+    const { frame, info } = await this._frame.resolveFrameForSelectorNoWait(selector, options, this);
+    // If we end up in the same frame => use the scope again, line above was noop.
+    return this._page.selectors.query(frame, info, this._frame === frame ? this : undefined);
   }
 
-  async querySelectorAll(metadata: CallMetadata, selector: string): Promise<ElementHandle<Element>[]> {
-    const controller = new ProgressController(metadata, this);
-    return controller.run(async progress => {
-      const { frame, info } = await this._frame.resolveFrameForSelector(progress, selector, {}, this);
-      // If we end up in the same frame => use the scope again, live above was noop.
-      return this._page.selectors._queryAll(frame, info, this._frame === frame ? this : undefined, true /* adoptToMain */);
-    });
+  async querySelectorAll(selector: string): Promise<ElementHandle<Element>[]> {
+    const { frame, info } = await this._frame.resolveFrameForSelectorNoWait(selector, {}, this);
+    // If we end up in the same frame => use the scope again, line above was noop.
+    return this._page.selectors._queryAll(frame, info, this._frame === frame ? this : undefined, true /* adoptToMain */);
   }
 
-  async evalOnSelectorAndWaitForSignals(metadata: CallMetadata, selector: string, strict: boolean, expression: string, isFunction: boolean | undefined, arg: any): Promise<any> {
-    const controller = new ProgressController(metadata, this);
-    return controller.run(async progress => {
-      const { frame, info } = await this._frame.resolveFrameForSelector(progress, selector, { strict }, this);
-      // If we end up in the same frame => use the scope again, live above was noop.
-      const handle = await this._page.selectors.query(frame, info, this._frame === frame ? this : undefined);
-      if (!handle)
-        throw new Error(`Error: failed to find element matching selector "${selector}"`);
-      const result = await handle.evaluateExpressionAndWaitForSignals(expression, isFunction, true, arg);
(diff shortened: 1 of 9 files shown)


packages/playwright-core/src/server/dom.ts
```diff
@@ -689,28 +691,48 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
...
 
-  async querySelector(selector: string, options: types.StrictOptions): Promise<ElementHandle | null> {
-    return this._page.selectors.query(this._context.frame, selector, options, this);
+  async querySelector(metadata: CallMetadata, selector: string, options: types.StrictOptions): Promise<ElementHandle | null> {
+    const controller = new ProgressController(metadata, this);
+    return controller.run(async progress => {
+      const { frame, info } = await this._frame.resolveFrameForSelector(progress, selector, options, this);
+      // If we end up in the same frame => use the scope again, live above was noop.
+      return this._page.selectors.query(frame, info, this._frame === frame ? this : undefined);
+    });
```
dgozman, line 700: Pass a timeout from TimeoutSettings, otherwise it is going to be zero. Same for other places.

packages/playwright-core/src/server/dom.ts
```diff
@@ -689,28 +691,48 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
         this._page._timeoutSettings.timeout(options));
   }
 
-  async querySelector(selector: string, options: types.StrictOptions): Promise<ElementHandle | null> {
-    return this._page.selectors.query(this._context.frame, selector, options, this);
+  async querySelector(metadata: CallMetadata, selector: string, options: types.StrictOptions): Promise<ElementHandle | null> {
+    const controller = new ProgressController(metadata, this);
+    return controller.run(async progress => {
+      const { frame, info } = await this._frame.resolveFrameForSelector(progress, selector, options, this);
+      // If we end up in the same frame => use the scope again, live above was noop.
```
dgozman, line 698: typo: live => line

packages/playwright-core/src/server/dom.ts
```diff
@@ -760,21 +782,32 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
...
+          const result = await pollHandler.finishHandle();
+          if (!result.asElement()) {
+            result.dispose();
+            return null;
+          }
+          const handle = result.asElement() as ElementHandle<Element>;
+          return handle._adoptTo(await frame._mainContext());
+        } catch (e) {
+          // Navigated while trying to adopt the node.
+          if (js.isJavaScriptErrorInEvaluate(e) || isSessionClosedError(e))
```
dgozman, line 806: I think this now retries when |this| is detached, and it should not.

packages/playwright-core/src/server/frames.ts
```diff
@@ -1450,16 +1449,16 @@ export class Frame extends SdkObject {
...
+    let frame: Frame | null = this;
     const frameChunks = splitSelectorByFrame(selector);
 
     for (let i = 0; i < frameChunks.length - 1 && progress.isRunning(); ++i) {
       const info = this._page.parseSelector(frameChunks[i], options);
-      const task = dom.waitForSelectorTask(info, 'attached');
+      const task = dom.waitForSelectorTask(info, 'attached', false, i === 0 ? scope : undefined);
       const handle = await frame._scheduleRerunnableHandleTask(progress, info.world, task);
       const element = handle.asElement() as dom.ElementHandle<Element>;
       if (i < frameChunks.length - 1) {
```
dgozman, line 1460: This is always true now.

packages/playwright-core/src/server/dom.ts
pavelfeldman: That is intended for element handle...

packages/playwright-core/src/server/dom.ts
pavelfeldman: No change in behavior here, added a failing test.

packages/playwright-core/src/server/dom.ts
```diff
@@ -760,21 +782,27 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
...
+        const result = await pollHandler.finishHandle();
+        if (!result.asElement()) {
+          result.dispose();
+          return null;
+        }
+        const handle = result.asElement() as ElementHandle<Element>;
+        try {
+          return await handle._adoptTo(await frame._mainContext());
+        } catch (e) {
+          return continuePolling;
```
dgozman, line 803: This is a small behavior change - previously we would fail, but now we'll retry. I guess that's even good, we should add a test :)

packages/playwright-core/src/server/frames.ts
```diff
@@ -738,18 +741,15 @@ export class Frame extends SdkObject {
...
-          return adopted;
+          return await handle._adoptTo(await frame._mainContext());
         } catch (e) {
-          // Navigated while trying to adopt the node.
-          if (js.isJavaScriptErrorInEvaluate(e))
-            throw e;
-          result.dispose();
+          return continuePolling;
+        } finally {
+          result?.dispose();
```
dgozman, line 750: Should not call `dispose` in the case of success, because you could dispose the returned handle (you should be able to trigger that with any selector engine that requires main world).

tests/page/selectors-frame.spec.ts
```diff
@@ -145,3 +198,60 @@ it('should click in lazy iframe', async ({ page, server }) => {
...
+  await routeIframe(page);
+  await page.goto(server.PREFIX + '/iframe.html');
+  const promise = page.click('button:has-text("Hello nested iframe")');
+  await page.waitForTimeout(100);
+  await page.goto(server.PREFIX + '/iframe-2.html');
+  await promise;
+});
+
+it('click fail if waiting for selector on element handle', async ({ page, server }) => {
+  it.fixme();
```
dgozman, line 249: How does this test fail? Also, title has "click" but there is no click in the test.

dgozman (approved):
Mostly good!

tests/page/selectors-frame.spec.ts
pavelfeldman: renamed the test

PR #10071, opened 2021-11-05
browser(webkit): make sure frameStoppedLoading is fired after load event

yury-s:
https://github.com/yury-s/WebKit/commit/fedf5be91362eeb6968c78747a6d8f47cb26f400

We rely on this event in the cases when loading was cancelled (primarily by window.stop()). This still holds but additionally the event is not fired if the document is still loading (e.g. parsing or loading deferred js modules).

#8340


pavelfeldman (approved):
Let's give it a try!

PR #10080, opened 2021-11-05
chore: fix connectOverCDP on Windows when proxy is used

### LINKED ISSUE #10077: [Question] connectOverCDP on Windows machine with proxy
Hi there. I'm using script, which running real chrome, connect to it via connectOverCDP and use different proxies for contexts.
 ```
await exec(`chrome --remote-debugging-port=${process.env.REAL_BROWSER_PORT}`)
let browser = await chromium.connectOverCDP({wsEndpoint: .wsRealEndpoint, timeout: 20000});
contextOptiopns['proxy'] = proxy
let context = browser.newContext(contextOptiopns);
```
This code works well on Linux but have error message on Windows: 
**browser.newContext: Browser needs to be launched with the global proxy. If all contexts override the proxy, global proxy will be never used and can be any string, for example "launch({ proxy: { server: 'per-context' } })**
I tried to add --proxy-server to exec command or set proxy globally for system, browser uses proxy - checked it via ip checker. But it does not help me to solve that issue on playwright.
Can you advise something?

### PR DESCRIPTION
**Pull request by mxschmitt**
Before this was throwing on Windows:

https://github.com/microsoft/playwright/blob/56ca3a18f56f4759da0e3f7e7a698e83948c3e4d/packages/playwright-core/src/server/browserContext.ts#L424-L426

Alternatively we could introduce again a connection mode e.g. on the browser and don't validate if connectOverCDP is used.

Fixes #10077

### DIFF
--- a/packages/playwright-core/src/server/chromium/chromium.ts
+++ b/packages/playwright-core/src/server/chromium/chromium.ts
@@ -92,7 +92,11 @@ export class Chromium extends BrowserType {
       artifactsDir,
       downloadsPath: artifactsDir,
       tracesDir: artifactsDir,
-      // Chromium on Windows global proxy workaround
+      // On Windows context level proxies only work, if there isn't a global proxy
+      // set. This is currently a bug in the CR/Windows networking stack. By
+      // passing an arbitrary value we disable the check in PW land which warns
+      // users in normal (launch/launchServer) mode since otherwise connectOverCDP
+      // does not work at all with proxies on Windows.
       proxy: { server: 'per-context' },
     };
     progress.throwIfAborted();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/playwright-core/src/server/chromium/chromium.ts [hunk 1]
```diff
@@ -91,7 +91,9 @@ export class Chromium extends BrowserType {
       browserLogsCollector: new RecentLogsCollector(),
       artifactsDir,
       downloadsPath: artifactsDir,
-      tracesDir: artifactsDir
+      tracesDir: artifactsDir,
+      // Chromium on Windows global proxy workaround
```
pavelfeldman, line 95: Say more here...

pavelfeldman (approved):

PR #10081, opened 2021-11-05
chore(create-playwright): make it Node.js 12 compatible

mxschmitt:


packages/create-playwright/src/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
```
mxschmitt, line 1: note: NCC does not traverse tsconfig's up and they fallback to their default one. Thats why we need to reference it two times:
a) rootDir of create-playwright for VSC to pick it up
b) rootDir/src/ of create-playwright so NCC does pick it up

pavelfeldman (approved):

.github/workflows/package_create_playwright.yml
```diff
@@ -16,7 +18,7 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v2
       with:
-        node-version: 14
+        node-version: 12
```
pavelfeldman, line 21: Matrix it

PR #10083, opened 2021-11-05
devops: use NPM 8

mxschmitt:
Most of the folks in the team already use NPM 8 anyway (it does not introduce breaking changes).

### DIFF
--- a/.github/workflows/infra.yml
+++ b/.github/workflows/infra.yml
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: 12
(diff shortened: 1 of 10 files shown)


pavelfeldman:
There seems to be some kind of conflict here - we use node 12, but npm 8. If we don't care about user experience on these bots, why not node 16?

JoelEinbinder:
> There seems to be some kind of conflict here - we use node 12, but npm 8. If we don't care about user experience on these bots, why not node 16?

These are not our installation tests, so they are not the ones that test the npm installation.

JoelEinbinder (approved):

PR #10102, opened 2021-11-06
api(frame-locator): introduce frame locators

pavelfeldman:
https://github.com/microsoft/playwright/issues/7982

### DIFF
--- a/docs/src/api/class-framelocator.md
+++ b/docs/src/api/class-framelocator.md
@@ -1,0 +1,44 @@
+# class: FrameLocator
+
+FrameLocator represents a view to the `iframe` on the page. It captures the logic sufficient to retrieve the `iframe` and locate elements in that iframe. FrameLocator can be created with either [`method: Page.frameLocator`] or [`method: Locator.frameLocator`] method.
+
+```js
+const locator = page.frameLocator('#my-frame').locator('text=Submit');
+await locator.click();
+```
+
+```java
+Locator locator = page.frameLocator("#my-frame").locator("text=Submit");
+locator.click();
+```
+
+```python async
+locator = page.frame_locator("#my-frame").locator("text=Submit")
+await locator.click()
+```
+
+```python sync
+locator = page.frame_locator("my-frame").locator("text=Submit")
+locator.click()
+```
+
+```csharp
+var locator = page.FrameLocator("#my-frame").Locator("text=Submit");
+await locator.ClickAsync();
+```
+
+## method: FrameLocator.frameLocator
+- returns: <[FrameLocator]>
+
+When working with iframes, you can create a frame locator that will enter the iframe and allow selecting elements
+in that iframe.
+
+### param: FrameLocator.frameLocator.selector = %%-find-selector-%%
+
+
+## method: FrameLocator.locator
+- returns: <[Locator]>
+
+The method finds an element matching the specified selector in the FrameLocator's subtree.
+
+### param: FrameLocator.locator.selector = %%-find-selector-%%
--- a/docs/src/api/class-locator.md
+++ b/docs/src/api/class-locator.md
@@ -533,6 +533,41 @@ Calls [focus](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus
 
 ### option: Locator.focus.timeout = %%-input-timeout-%%
 
+
+## method: Locator.frameLocator
+- returns: <[FrameLocator]>
+
+When working with iframes, you can create a frame locator that will enter the iframe and allow selecting elements
+in that iframe:
+
+```js
+const locator = page.frameLocator('iframe').locator('text=Submit');
+await locator.click();
+```
+
+```java
+Locator locator = page.frameLocator("iframe").locator("text=Submit");
+locator.click();
+```
+
+```python async
+locator = page.frame_locator("iframe").locator("text=Submit")
+await locator.click()
+```
+
+```python sync
+locator = page.frame_locator("text=Submit").locator("text=Submit")
+locator.click()
+```
+
+```csharp
+var locator = page.FrameLocator("iframe").Locator("text=Submit");
+await locator.ClickAsync();
+```
+
+### param: Locator.frameLocator.selector = %%-find-selector-%%
+
+
 ## async method: Locator.getAttribute
 - returns: <[null]|[string]>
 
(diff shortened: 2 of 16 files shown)


docs/src/api/class-framelocator.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+## method: FrameLocator.frameLocator
+- returns: <[FrameLocator]>
+
+When working with iframes, you can create a frame locator that will enter the iframe and allow selecting elements
+in that iframe:
+
+```js
+const locator = page.frameLocator('iframe').locator('text=Submit');
```
dgozman, line 37: This should be something like `frameLocator('#my-frame').frameLocator('.preview-frame')`

docs/src/api/class-framelocator.md
```diff
@@ -0,0 +1,69 @@
...
+await locator.ClickAsync();
+```
+
+### param: FrameLocator.frameLocator.selector = %%-find-selector-%%
+
+
+## method: FrameLocator.locator
+- returns: <[Locator]>
+
+The method finds an element matching the specified selector in the FrameLocator's subtree.
```
dgozman, line 67: An example here would be nice!

docs/src/api/class-locator.md
```diff
@@ -533,6 +533,41 @@ Calls [focus](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus
...
 
+
+## method: Locator.frameLocator
+- returns: <[FrameLocator]>
+
+When working with iframes, you can create a frame locator that will enter the iframe and allow selecting elements
+in that iframe:
+
+```js
+const locator = page.frameLocator('iframe').locator('text=Submit');
```
dgozman, line 544: This should be `locator.frameLocator()` instead of a page.

dgozman (approved):

docs/src/api/class-framelocator.md
pavelfeldman: I don't think users would have a lot of nested frames, feels like me example is more relevant. I can use iframe#my-frame for clarity though.

docs/src/api/class-framelocator.md
pavelfeldman: Done

docs/src/api/class-locator.md
pavelfeldman: Why?!?

packages/playwright-core/src/server/common/selectorParser.ts
```diff
@@ -77,18 +77,20 @@ export function splitSelectorByFrame(selectorText: string): ParsedSelector[] {
...
 }
 
-
 export function stringifySelector(selector: string | ParsedSelector): string {
   if (typeof selector === 'string')
     return selector;
-  return selector.parts.map((p, i) => `${i === selector.capture ? '*' : ''}${p.name}=${p.source}`).join(' >> ');
+  return selector.parts.map((p, i) => {
+    const prefix = p.name === 'css' ? '' : p.name + '=';
+    return `${i === selector.capture ? '*' : ''}${prefix}${p.source}`;
```
dgozman, line 92: To be 100% correct, we have to escape quotes in the source, and also escape ">>" signs. If we never re-parse the result of this function, this might be ok.

packages/playwright-core/src/server/frames.ts
```diff
@@ -1476,6 +1506,7 @@ export class Frame extends SdkObject {
     for (let i = 0; i < frameChunks.length - 1 && progress.isRunning(); ++i) {
       const info = this._page.parseSelector(frameChunks[i], options);
       const task = dom.waitForSelectorTask(info, 'attached', false, i === 0 ? scope : undefined);
+      progress.log(`  waiting for frame "${stringifySelector(frameChunks[i])}"`);
```
dgozman, line 1509: Nice one!

packages/playwright-core/src/server/frames.ts
```diff
@@ -1326,9 +1355,10 @@ export class Frame extends SdkObject {
     const callbackText = body.toString();
 
     return controller.run(async progress => {
-      return this.retryWithProgress(progress, selector, options, async (frame, info) => {
+      return this.retryWithProgress(progress, selector, options, async selectorInFrame => {
         // Be careful, |this| can be different from |frame|.
         progress.log(`waiting for selector "${selector}"`);
+        const { frame, info } = selectorInFrame || { frame: this, info: { parsed: { parts: [{ name: 'control', body: 'return-empty', source: 'control-return-empty' }] }, world: 'utility', strict: !!options.strict } };
```
dgozman, line 1361: typo: 'control-return-empty' => 'control=return-empty'

packages/playwright-core/src/server/frames.ts
```diff
@@ -965,20 +972,25 @@ export class Frame extends SdkObject {
...
+    progress.log(`waiting for selector "${selector}"`);
     const continuePolling = Symbol('continuePolling');
     while (progress.isRunning()) {
-      const pair = await this._resolveFrameForSelector(progress, selector, options, scope);
-      if (!pair) {
-        // Missing content frame.
-        await new Promise(f => setTimeout(f, 100));
-        continue;
+      let selectorInFrame: SelectorInFrame | null;
+      if (options.omitAttached) {
```
dgozman, line 982: Somehow the name of this option is not clear to me. Perhaps it means "noWait" or "allowEmpty"?

packages/playwright-core/src/server/frames.ts
```diff
@@ -965,20 +972,25 @@ export class Frame extends SdkObject {
   async retryWithProgress<R>(
     progress: Progress,
     selector: string,
-    options: types.StrictOptions & types.TimeoutOptions,
-    action: (frame: Frame, info: SelectorInfo, continuePolling: symbol) => Promise<R | symbol>,
+    options: types.StrictOptions & types.TimeoutOptions & { omitAttached?: boolean },
+    action: (selector: SelectorInFrame | null, continuePolling: symbol) => Promise<R | symbol>,
     scope?: dom.ElementHandle): Promise<R> {
+    progress.log(`waiting for selector "${selector}"`);
```
dgozman, line 978: Now that we have "waiting for frame <selector>" and "waiting for selector <selector>" in the last frame, perhaps we do not need this one? It contains strange things like "control=" that users never entered :)

dgozman (approved):

packages/playwright-core/src/server/common/selectorParser.ts
pavelfeldman: We can rename the method yes, we never re-parse.

packages/playwright-core/src/server/frames.ts
pavelfeldman: The semantics of this option did not change - well are as we were omitting the 'attached' requirement.

packages/playwright-core/src/server/frames.ts
pavelfeldman: Sounds good, I might need to rebase expectations for this one, let me check.

PR #9451, opened 2021-10-12
ci(workflow): add cache to workflows using actions/setup-node

oscard0m:
## Description

Add `cache` to workflows using `actions/setup-node`

## Context

`setup-node` GitHub Action just released a new option to add cache to steps using it.

You can find the details here: https://github.blog/changelog/2021-07-02-github-actions-setup-node-now-supports-dependency-caching/

---

🤖 This PR has been generated automatically by [this octoherd script](https://github.com/oscard0m/octoherd-script-add-cache-to-node-github-action), feel free to run it in your GitHub user/org repositories! 💪🏾

PR #10116, opened 2021-11-07
ci(workflows): add 'npm' cache for actions/setup-node in .github/workflows

oscard0m:
## Description

Add `cache` to workflows using `actions/setup-node`

## Context

`setup-node` GitHub Action just released a new option to add cache to steps using it.

You can find the details here: https://github.blog/changelog/2021-07-02-github-actions-setup-node-now-supports-dependency-caching/

### How much better it is with caching?
It depends on
- The amount of dependencies the project has ([you can see a summarized table in this ADR](https://github.com/actions/setup-node/issues/271#issuecomment-871408337))
- The size of these dependencies ([if it exceeds the limit of 5GB, the cache starts to overwrite old entries](https://github.com/actions/cache#cache-limits))
- How often does these pipelines run? ([GitHub will remove cache entries not accessed in the last 7 days](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy))
---

Context: 
* https://github.com/microsoft/playwright/pull/9451
* https://github.com/microsoft/playwright/pull/9451#issuecomment-948770133

### DIFF
--- a/.github/workflows/infra.yml
+++ b/.github/workflows/infra.yml
@@ -19,6 +19,7 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: 12
+        cache: npm
     - run: npm i -g npm@7
     - run: npm ci
     - run: npm run build
--- a/.github/workflows/publish_canary.yml
+++ b/.github/workflows/publish_canary.yml
@@ -19,6 +19,7 @@ jobs:
       with:
         node-version: 12
         registry-url: 'https://registry.npmjs.org'
+        cache: npm
     - run: npm i -g npm@7
     - run: npm ci
     - run: npm run build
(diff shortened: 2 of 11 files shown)


aslushnikov:
I wouldn't bother caching npm - it might save us 10-20 seconds, but the moajority of time is still spend in our tests, which are like 30 minutes. So wins are negligible, whereas it does add complexity to what's going on on CI and diverges CI further from local setup.

oscard0m:
> I wouldn't bother caching npm - it might save us 10-20 seconds, but the moajority of time is still spend in our tests, which are like 30 minutes. So wins are negligible, whereas it does add complexity to what's going on on CI and diverges CI further from local setup.

Thanks for the feedback @aslushnikov . Has sense your explanation

PR #10121, opened 2021-11-08
feat(api): add explicit async testInfo.attach

### LINKED ISSUE #10113: [Feature] async test.attach({…}) API that Automagically Handles Temporary Paths
This is a follow up to [this Slack thread](https://playwright.slack.com/archives/C01AG7QGW2F/p1636232194179300?thread_ts=1636148004.173400&cid=C01AG7QGW2F). Please read that message for background on how I messed up (hehe) and why this proposal is useful (IMO)! We're in the middle of consolidating UI tests and API tests into @playwright/test that used to be in a mix of JS test runner and Python pytest. 

I'd like to propose adding:

```
test.attach: ({ name: string; contentType: string; path?: string; body?: … }, { move: boolean = False }) => Promise<void>;
```

so it can be explicitly awaited in a test, and the Playwright Test Runner internally can (1) confirm the intended attachment exists, and (2) copies (by default or moves, explicit setting) the attachment to a location that the reporters are guaranteed to have access to even after the test finishes and potentially cleans up temporary directories.

This would help ensure issues with attachments, which might only be looked at when tests suddenly start failing, are surfaced very early in authoring tests…not just when you debug a failing test months later!

The current implementation and reporters swallow errors around attachments and requires knowing about `testInfo.createTestOutput` and combining it with `testInfo.attachments.push`, but even with that there is room to accidentally mess up.

Let me know your thoughts! I'd be happy to contribute this or iterate on an alternative…even if it's just adding a warning log in the HTML Reporter about using `testInfo.testOutput` in the case of files not found! Ultimately, I'd like to surface issues with attachments as soon as possible to the user to ensure they are in fact there when the user thinks they will be!

🎉 Thanks, as always, for these awesome tools! We ❤️  the new HTML Report…you don't want to see what our legacy reports look like! 😄

rwoll:
We add an explicit async API for attaching file paths (and Buffers) to
tests that can be awaited to help users ensure they are attaching files
that actually exist at both the time of the invocation and later when
reporters (like the HTML Reporter) run and package up test artifacts.

This is intended to help surface attachment issues as soon as possible
so you aren't silently left with a missing attachment
minutes/days/months later when you go to debug a suddenly breaking test
expecting an attachment to be there.

NB: The current implemntation incurs an extra file copy compared to
manipulating the raw attachments array. If users encounter performance
issues because of this, we can consider an option parameter that uses
rename under the hood instead of copy. However, that would need to be
used with care if the file were to be accessed later in the test.

Resolves #10113

### DIFF
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -57,7 +57,7 @@ test('basic test', async ({ page }, testInfo) => {
   const path = testInfo.outputPath('screenshot.png');
   await page.screenshot({ path });
   await testInfo.attach(path);
-  // optionally override the name
+  // Optionally override the name.
   await testInfo.attach(path, { name: 'example.png' });
   // optionally override the contentType
   await testInfo.attach(path, { name: 'example.custom-file', contentType: 'x-custom-content-type' });
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -59,7 +59,7 @@ test('basic test', async ({ page }, testInfo) => {
   await testInfo.attach(path);
   // Optionally override the name.
   await testInfo.attach(path, { name: 'example.png' });
-  // optionally override the contentType
+  // Optionally override the contentType.
   await testInfo.attach(path, { name: 'example.custom-file', contentType: 'x-custom-content-type' });
 });
 ```
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -76,7 +76,7 @@ test('basic test', async ({ page }, testInfo) => {
   await testInfo.attach(path);
   // Optionally override the name.
   await testInfo.attach(path, { name: 'example.png' });
-  // optionally override the contentType
+  // Optionally override the contentType.
   await testInfo.attach(path, { name: 'example.custom-file', contentType: 'x-custom-content-type' });
 });
 ```
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -129,7 +129,7 @@ Attach data to the current test, either a `string` or a `Buffer`. Some reporters
 ### param: TestInfo.attach#2.name
 - `name` <[string]> Attachment name.
 
-### param: TestInfo.attach#2.contentType
+### option: TestInfo.attach#2.contentType
 - `contentType` <[void]|[string]> Optional content type of this attachment to properly present in the report, for example `'application/json'` or `'application/xml'`. If omitted, this falls back to an inferred type based on the [`param: name`]'s extension; if the type cannot be inferred from the name's extension, it will be set to `text/plain` (if [`param: body`] is a `string`) or `application/octet-stream` (if [`param: body`] is a `Buffer`).
 
 ## property: TestInfo.column
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -113,7 +113,7 @@ after awaiting the attach call.
 - `path` <[string]> Path on the filesystem to the attached file.
 
 ### option: TestInfo.attach#1.name
-- `name` <[void]|[string]> Optional attachment name. If omitted, this will be inferred from the [`param: name`] (if specified) or [`param: path`].
+- `name` <[void]|[string]> Optional attachment name. If omitted, this will be inferred from [`param: path`].
 
 ### option: TestInfo.attach#1.contentType
 - `contentType` <[void]|[string]> Optional content type of this attachment to properly present in the report, for example `'application/json'` or `'image/png'`. If omitted, this falls back to an inferred type based on the [`param: name`] (if set) or [`param: path`]'s extension; it will be set to `application/octet-stream` if the type cannot be inferred from the file extension.
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -121,7 +121,7 @@ after awaiting the attach call.
 
 ## method: TestInfo.attach#2
 
-Similar to [`method: TestInfo.attach#1`] but attaches inline `strings` or `Buffers` instead of files from disk.
+Attach data to the current test, either a `string` or a `Buffer`. Some reporters show test attachments.
 
 ### param: TestInfo.attach#2.body
 - `body` <[string]|[Buffer]> Attachment body.
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -74,7 +74,7 @@ test('basic test', async ({ page }, testInfo) => {
   const path = testInfo.outputPath('screenshot.png');
   await page.screenshot({ path });
   await testInfo.attach(path);
-  // optionally override the name
+  // Optionally override the name.
   await testInfo.attach(path, { name: 'example.png' });
   // optionally override the contentType
   await testInfo.attach(path, { name: 'example.custom-file', contentType: 'x-custom-content-type' });
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -38,7 +38,14 @@ Learn more about [test annotations](./test-annotations.md).
   - `path` <[void]|[string]> Optional path on the filesystem to the attached file.
   - `body` <[void]|[Buffer]> Optional attachment body used instead of a file.
 
-The list of files or buffers attached to the current test. Some reporters show test attachments. For example, you can attach a screenshot to the test.
+The list of files or buffers attached to the current test. Some reporters show test attachments.
+
+To safely add a file from disk as an attachment, please use [`method: TestInfo.attach#1`] instead of directly pushing onto this array. For inline attachments, use [`method: TestInfo.attach#1`].
+
+## method: TestInfo.attach#1
+Attach a file from disk to the current test. Some reporters show test attachments. The [`option: name`] and [`option: contentType`] will be inferred by default from the [`param: path`], but you can optionally override either of these.
+
+For example, you can attach a screenshot to the test:
 
 ```js js-flavor=js
 const { test, expect } = require('@playwright/test');
@@ -49,7 +56,11 @@ test('basic test', async ({ page }, testInfo) => {
   // Capture a screenshot and attach it.
   const path = testInfo.outputPath('screenshot.png');
   await page.screenshot({ path });
-  testInfo.attachments.push({ name: 'screenshot', path, contentType: 'image/png' });
+  await testInfo.attach(path);
+  // optionally override the name
+  await testInfo.attach(path, { name: 'example.png' });
+  // optionally override the contentType
+  await testInfo.attach(path, { name: 'example.custom-file', contentType: 'x-custom-content-type' });
 });
 ```
 
@@ -62,9 +73,64 @@ test('basic test', async ({ page }, testInfo) => {
   // Capture a screenshot and attach it.
   const path = testInfo.outputPath('screenshot.png');
   await page.screenshot({ path });
-  testInfo.attachments.push({ name: 'screenshot', path, contentType: 'image/png' });
+  await testInfo.attach(path);
+  // optionally override the name
+  await testInfo.attach(path, { name: 'example.png' });
+  // optionally override the contentType
+  await testInfo.attach(path, { name: 'example.custom-file', contentType: 'x-custom-content-type' });
 });
 ```
+
+Or you can attach files returned by your APIs:
+
+```js js-flavor=js
+const { test, expect } = require('@playwright/test');
+
+test('basic test', async ({}, testInfo) => {
+  const { download } = require('./my-custom-helpers');
+  const tmpPath = await download('a');
+  await testInfo.attach(tmpPath, { name: 'example.json' });
+});
+```
+
+```js js-flavor=ts
+import { test, expect } from '@playwright/test';
+
+test('basic test', async ({}, testInfo) => {
+  const { download } = require('./my-custom-helpers');
+  const tmpPath = await download('a');
+  await testInfo.attach(tmpPath, { name: 'example.json' });
+});
+```
+
+:::note
+[`method: TestInfo.attach#1`] automatically takes care of copying attachments to a
+location that is accessible to reporters, even if you were to delete the attachment
+after awaiting the attach call.
+:::
+
+### param: TestInfo.attach#1.path
+- `path` <[string]> Path on the filesystem to the attached file.
+
+### option: TestInfo.attach#1.name
+- `name` <[void]|[string]> Optional attachment name. If omitted, this will be inferred from the [`param: name`] (if specified) or [`param: path`].
+
+### option: TestInfo.attach#1.contentType
+- `contentType` <[void]|[string]> Optional content type of this attachment to properly present in the report, for example `'application/json'` or `'image/png'`. If omitted, this falls back to an inferred type based on the [`param: name`] (if set) or [`param: path`]'s extension; it will be set to `application/octet-stream` if the type cannot be inferred from the file extension.
+
+
+## method: TestInfo.attach#2
+
+Similar to [`method: TestInfo.attach#1`] but attaches inline `strings` or `Buffers` instead of files from disk.
+
+### param: TestInfo.attach#2.body
+- `body` <[string]|[Buffer]> Attachment body.
+
+### param: TestInfo.attach#2.name
+- `name` <[string]> Attachment name.
+
+### param: TestInfo.attach#2.contentType
+- `contentType` <[void]|[string]> Optional content type of this attachment to properly present in the report, for example `'application/json'` or `'application/xml'`. If omitted, this falls back to an inferred type based on the [`param: name`]'s extension; if the type cannot be inferred from the name's extension, it will be set to `text/plain` (if [`param: body`] is a `string`) or `application/octet-stream` (if [`param: body`] is a `Buffer`).
 
 ## property: TestInfo.column
 - type: <[int]>
--- a/packages/playwright-test/src/workerRunner.ts
+++ b/packages/playwright-test/src/workerRunner.ts
@@ -15,7 +15,7 @@
  */
 
 import fs from 'fs';
-import path, { basename } from 'path';
+import path from 'path';
 import rimraf from 'rimraf';
 import * as mime from 'mime';
 import util from 'util';
@@ -31,7 +31,6 @@ import { ProjectImpl } from './project';
 import { FixturePool, FixtureRunner } from './fixtures';
 import { DeadlineRunner, raceAgainstDeadline } from 'playwright-core/lib/utils/async';
 import { calculateFileSha1 } from 'playwright-core/lib/utils/utils';
-// import { calculateFileSha1 } from 'playwright-core/lib/utils/utils';
 
 const removeFolderAsync = util.promisify(rimraf);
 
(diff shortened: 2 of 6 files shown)


packages/playwright-test/src/testType.ts
```diff
@@ -202,6 +208,29 @@ export class TestTypeImpl {
...
+    if (attachment.body === undefined && attachment.path === undefined)
+      errorWithLocation(location, `test.attach() expects either body or path to be defined, but neither was`);
+
+    const tmpAttachment = { ...attachment };
+    if (attachment.path) {
+      const hash = await calculateFileSha1(attachment.path);
+      const dest = testInfo.outputPath('attachments', hash + path.extname(attachment.path));
+      await fs.promises.mkdir(path.dirname(dest), { recursive: true });
+      await fs.promises.copyFile(attachment.path, dest);
+      tmpAttachment.path = dest;
```
rwoll, line 228: NB: This is intentionally an absolute path. Reporters are responsible for "relativising" it as needed for their packaging formats.

rwoll:
Just removed the draft status as the tests are now all passing! (I think I won the battle with Windows path separators!) This is now ready for a review pass! Thanks!

cc: @pavelfeldman

tests/playwright-test/reporter-raw.spec.ts
```diff
@@ -110,6 +113,47 @@ test('should save attachments', async ({ runInlineTest }, testInfo) => {
   expect(path2).toBe('dummy-path');
 });
 
+test.extend<{ tmpDir: string }>({
```
dgozman, line 116: Why do we need this fixture? Might as well just create a file in the test, and then remove it manually?

dgozman (commented):
Thank you Ross for the PR! A few high-level comments:
- Why `test.attach` and not `testInfo.attach`?
- Should we separately change the html reporter (and any others) to not swallow the error, but instead show "Attachment file is missing" as attachment body, to give some indication? We can also rename the attachment to something like `ERROR`.
- Do you find the `move` option useful right away?

docs/src/test-api/class-test.md
```diff
@@ -1052,3 +1052,65 @@ test('test with locale', async ({ page }) => {
...
+  await test.attach.push({ name: 'screenshot', path, contentType: 'image/png' });
+});
+```
+
+Or you can attach files returned by your APIs:
+
+```js js-flavor=js
+const { test, expect } = require('@playwright/test');
+
+test('basic test', async ({ download }) => {
```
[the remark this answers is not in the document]
rwoll: I like your alternative! Keeps the example as simple as possible. I'll incorporate into the next changeset.

docs/src/test-api/class-testinfo.md
```diff
@@ -38,33 +38,9 @@ Learn more about [test annotations](./test-annotations.md).
...
-test('basic test', async ({ page }, testInfo) => {
-  await page.goto('https://playwright.dev');
-
-  // Capture a screenshot and attach it.
-  const path = testInfo.outputPath('screenshot.png');
-  await page.screenshot({ path });
-  testInfo.attachments.push({ name: 'screenshot', path, contentType: 'image/png' });
-});
-```
+To safely add an attachment to a test, please use [`method: Test.attach`] instead of directly pushing onto this array. Future versions of Playwright Test may mark this array as read-only.
```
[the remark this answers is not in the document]
rwoll: Here are my thoughts:

* Encouraging a canonical/uniform way of doing something in a lib is better than having multiple ways (and needing to debug multiple ways), so I think it would be best to move people onto the explicit method instead of magically pushing onto an array
* Pushing onto array as a user feels a little haphazard and like I'm doing something I shouldn't be doing with Playwright.
* Pushing onto array removes any opportunity for validation early (the whole point of this MR), so future-looking, it would be best to force things through the explicit method whose internal implementation, validation, and data-structure can change without breaking code

I'm trying to think of cases where I'd ever want to push onto the array directly as a user instead of using the async method and cannot. 

However, I don't want to include more words in the docs than necessary/true, so if TestInfo.attachments.push continues to be in the docs and encouraged, let's drop this note. Thoughts?

docs/src/test-api/class-test.md
```diff
@@ -1052,3 +1052,65 @@ test('test with locale', async ({ page }) => {
...
+```
+
+:::note
+[`method: Test.attach`] automatically takes care of copying attachments to a
+location that is accessible to reporters, even if you were to delete the attachment
+after awaiting the attach call.
+:::
+
+### param: Test.attach.attachment
+- type: <[TestAttachment]>
```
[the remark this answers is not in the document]
rwoll: Noted—it'll be updated in the next changeset.

tests/playwright-test/reporter-raw.spec.ts
rwoll: What if I write a buggy test or it gets interrupted while executing? Then I litter the FS with temp files I haven't removed! Playwright Test fixtures and hooks ensure these get cleaned up properly! 😄 

Let me know if that makes sense, otherwise, I can remove (since it would most likely get removed or worst case leave behind a temp file on the host machine).

rwoll:
@dgozman Thanks for the review! I've replied to the inline comments with any outstanding questions. For the higher-level comments, see replies below.

> Thank you Ross for the PR! A few high-level comments:
> 
> * Why `test.attach` and not `testInfo.attach`?

1. Convenience: We have quite a few parameterized tests, helpers, and fixtures. It's tedious to plumb through TestInfo object to each of them. If we can call `test.skip` (or any of the other `test.*` methods) anywhere (from within a test) without passing `test` through the call chain, I expect to be able to call `attach` from those same places with the same conveniences.
2. User Understanding: From an end-user perspective, what's the distinction between `test` and `testInfo`? They have a lot of overlap, yet they are different…without looking into them too deeply, it seems to me that `testInfo` could be deprecated and everything comes off of `test` (`test.info.*`). When looking through docs, sometimes I end up on TestInfo page, other times I end up on `test` page, they feel like they are the same.

> * Should we separately change the html reporter (and any others) to not swallow the error, but instead show "Attachment file is missing" as attachment body, to give some indication? We can also rename the attachment to something like `ERROR`.

Yes, before I arrived at the explicit attachment API, this is sort of what I was thinking (or a log line). I'll file a follow up ticket to do that in a distinct MR.

> * Do you find the `move` option useful right away?

No. I had it locally, and then realized it seems like a premature optimization (and complication) and instead opted to keep the API as minimal as possible for now. If it becomes an issue, we can add it, but no need to add it now. Most files we have are in the KB to MB size. We'll have some tests attaching files in the 100s of MB to <1GB, but on modern hardware+OSes, copy v. rename feels like a premature optimization.

Thanks! 🎉

tests/playwright-test/reporter-raw.spec.ts
dgozman: If you create a file inside `TestInfo.outputPath(...)` it will not be left behind :)

docs/src/test-api/class-testinfo.md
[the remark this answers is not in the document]
dgozman: We try to never break our APIs, so it's unlikely that it will become readonly, unless there is a **very** strong reason for it. But it is safe to change all usages of `testInfo.attachments` in the docs to `testInfo.attach()`.

dgozman:
> > * Why `test.attach` and not `testInfo.attach`?
> 
> 1. Convenience: We have quite a few parameterized tests, helpers, and fixtures. It's tedious to plumb through TestInfo object to each of them. If we can call `test.skip` (or any of the other `test.*` methods) anywhere (from within a test) without passing `test` through the call chain, I expect to be able to call `attach` from those same places with the same conveniences.
> 2. User Understanding: From an end-user perspective, what's the distinction between `test` and `testInfo`? They have a lot of overlap, yet they are different…without looking into them too deeply, it seems to me that `testInfo` could be deprecated and everything comes off of `test` (`test.info.*`). When looking through docs, sometimes I end up on TestInfo page, other times I end up on `test` page, they feel like they are the same.

We thought about `test.info` a few times already. However, we are not yet ready to expose it because that will ensure we'll never be able to run two tests in the same process at the same time. Same goes for `test.attach()` which is bound to `TestInfo` behind the scenes.

So for now, let's go with `TestInfo.attach()`, let the user plumb `TestInfo` where needed (or make their own "global test info" hack 😄 ) and live with it for a while. Does that sound good?

packages/playwright-test/types/test.d.ts
```diff
@@ -1310,12 +1319,28 @@ export interface TestInfo {
...
+   *   const { download } = require('./my-custom-helpers');
+   *   const tmpPath = await download('a');
+   *   await testInfo.attach({ name: 'example.json', path: tmpPath, contentType: 'application/json' });
+   * });
+   * ```
+   *
+   * > NOTE: [testInfo.attach(type:)](https://playwright.dev/docs/api/class-testinfo#test-info-attach) automatically takes
+   * care of copying attachments to a location that is accessible to reporters, even if you were to delete the attachment
+   * after awaiting the attach call.
+   * @param type:
```
rwoll, line 1341: @dgozman This doesn't look right! I tried to debug—albeit very briefly—and could not resolve. Can you give me some pointers? Maybe my syntax in docs/src/test-api/class-testinfo.md is incorrect. (I did inline the type compared to the last changeset where I had pulled it out to a TestAttachment type.)

packages/playwright-test/src/workerRunner.ts
```diff
@@ -337,6 +339,25 @@ export class WorkerRunner extends EventEmitter {
       },
     };
 
+    testInfo.attach = async attachment => {
```
rwoll, line 342: @dgozman Is there a better way to do this? It smells a little, but pulling out a shared reference to both `outputPath` and the `attachments` array felt equally gross…so this was the cleanest!

Aside—but may become relevant—why are some shared vars wrapped in a function that's immediately invoked (https://github.com/microsoft/playwright/blob/a51ac39275c33161f6d0f57751730afcce48850c/packages/playwright-test/src/workerRunner.ts#L230)? I'm assuming there is good reason related to scoping and worker processes, but not totally sure and piqued my interest!

rwoll (commented):
@dgozman Thanks for the feedback! I moved the API onto TestInfo and pushed all other updates from prior comments.

Thanks!

packages/playwright-test/src/workerRunner.ts
dgozman: I think you can just inline the method, similar to `_addStep`.

> why are some shared vars wrapped in a function that's immediately invoked

That's just stylistic, to give some piece of logic a name, and avoid introducing many more variables into the scope.

docs/src/test-api/class-testinfo.md
```diff
@@ -62,10 +71,46 @@ test('basic test', async ({ page }, testInfo) => {
...
+```
+
+:::note
+[`method: TestInfo.attach`] automatically takes care of copying attachments to a
+location that is accessible to reporters, even if you were to delete the attachment
+after awaiting the attach call.
+:::
+
+### param: TestInfo.attach.attachment
+- type: <[Object]>
```
dgozman, line 107:
```suggestion
- `attachment` <[Object]>
```

packages/playwright-test/types/test.d.ts
dgozman: I made a suggestion in the md file.

utils/generate_types/overrides-test.d.ts
```diff
@@ -207,6 +207,7 @@ export interface TestInfo {
   timeout: number;
   annotations: { type: string, description?: string }[];
   attachments: { name: string, path?: string, body?: Buffer, contentType: string }[];
+  attach(attachment: { name: string, path?: string, body?: Buffer, contentType: string }): Promise<void>;
```
dgozman, line 210: Looking at this, perhaps we can simplify the api?
```js
// Infers content type, defaults name to the basename(path)
attach(path: string, options: { contentType?: string, name?: string })
// Defaults to text/plain
attach(name: string, body: string)
// Manual buffer
attach(name: string, body: Buffer, contentType: string)
```

dgozman (commented):
Looks very good, mostly a question about nicer API - wdyt?

packages/playwright-test/src/workerRunner.ts
rwoll: Sweet! Forgot you can reference the object that you're actively building out from with itself…what a wild thing! :)

utils/generate_types/overrides-test.d.ts
rwoll: Yes! Love it! I iterated a bit while converting some local tests and have arrived at something slightly different since there are `string`s you want to have non `text/plain` content types. (JSON.stringify FTW).

Also made inference uniform across the various versions. `contentType` > inferred from `name` > inferred from `path` (which is applicable to more than just the file path variant).

rwoll:
@dgozman I'm **not** yet ready for another pass…but would like some help getting the overloaded API docs to be happy. 

`npm run lint` is complaining with:

```
> playwright-internal@1.17.0-next doc
> node utils/doclint/cli.js

TypeError: While processing "js"
Cannot read property 'sort' of undefined
    at Documentation.Member.index (/Users/rwoll/code/playwright/utils/doclint/documentation.js:336:29)
    at Documentation.Class.index (/Users/rwoll/code/playwright/utils/doclint/documentation.js:203:14)
    at Documentation.index (/Users/rwoll/code/playwright/utils/doclint/documentation.js:114:11)
    at new Documentation (/Users/rwoll/code/playwright/utils/doclint/documentation.js:61:10)
    at new ApiParser (/Users/rwoll/code/playwright/utils/doclint/api_parser.js:58:26)
    at parseApi (/Users/rwoll/code/playwright/utils/doclint/api_parser.js:320:10)
    at run (/Users/rwoll/code/playwright/utils/doclint/cli.js:105:37)
```

Looks like maybe `options` is a special keyword or something? (from `playwright/utils/doclint/documentation.js:336`)

Tried a few variations in the doc, but all failed in confusing ways!

dgozman:
> Looks like maybe `options` is a special keyword or something? (from `playwright/utils/doclint/documentation.js:336`)

It is, see https://github.com/microsoft/playwright/blob/master/docs/src/api/class-page.md?plain=1#L540 for an example.

rwoll:
> > Looks like maybe `options` is a special keyword or something? (from `playwright/utils/doclint/documentation.js:336`)
> 
> It is, see https://github.com/microsoft/playwright/blob/master/docs/src/api/class-page.md?plain=1#L540 for an example.

Ah, thanks! I was looking at https://github.com/microsoft/playwright/blob/7eec66d0f9ae20440280eb3aacfb5ddec1e4d132/docs/src/test-api/class-test.md?plain=1#L1049 but I guess that's different since they aren't inline!

rwoll:
@dgozman This is ready for another review! 😄 It now features the API overload for convenience including `contentType` inference. 

I added quite a few tests of positive test cases and some error cases (missing files) that prompted this new API.

The overloaded API diverges a bit from your initial suggestion, but I think it's convenient and sensible! In all cases—inline or file—we will automatically infer `name` and `contentType` unless they are specifically set and we have sensible default fallbacks.

Thanks! 🎉

rwoll:
I don't think the macos build failure is related to this change (https://github.com/microsoft/playwright/runs/4185131112?check_suite_focus=true). Assuming that's true, this is ready for `CQ1` suite!

tests/playwright-test/reporter-attachment.spec.ts
```diff
@@ -79,3 +79,36 @@ test('render trace attachment', async ({ runInlineTest }) => {
   expect(text).toContain('    ------------------------------------------------------------------------------------------------');
   expect(result.exitCode).toBe(1);
 });
+
+for (const { description, apiCall } of [
```
dgozman, line 83: Instead of doing this, you can put multiple test() calls in a single inner a.test.js file and check all expectations at the same time - makes for an easier test. Same for other tests below.

Here is an example: https://github.com/microsoft/playwright/blob/master/tests/playwright-test/playwright.expect.text.spec.ts#L146

packages/playwright-test/src/workerRunner.ts
```diff
@@ -29,6 +30,8 @@ import { Annotations, TestError, TestInfo, TestInfoImpl, TestStepInternal, Worke
 import { ProjectImpl } from './project';
 import { FixturePool, FixtureRunner } from './fixtures';
 import { DeadlineRunner, raceAgainstDeadline } from 'playwright-core/lib/utils/async';
+import { calculateFileSha1 } from 'playwright-core/lib/utils/utils';
+// import { calculateFileSha1 } from 'playwright-core/lib/utils/utils';
```
dgozman, line 34: Please remove this line

packages/playwright-test/src/workerRunner.ts
```diff
@@ -15,8 +15,9 @@
  */
 
 import fs from 'fs';
-import path from 'path';
+import path, { basename } from 'path';
```
dgozman, line 18: Please use `path.basename` instead.

docs/src/test-api/class-testinfo.md
```diff
@@ -62,10 +73,65 @@ test('basic test', async ({ page }, testInfo) => {
...
+[`method: TestInfo.attach#1`] automatically takes care of copying attachments to a
+location that is accessible to reporters, even if you were to delete the attachment
+after awaiting the attach call.
+:::
+
+### param: TestInfo.attach#1.path
+- `path` <[string]> Path on the filesystem to the attached file.
+
+### option: TestInfo.attach#1.name
+- `name` <[void]|[string]> Optional attachment name. If omitted, this will be inferred from the [`param: name`] (if specified) or [`param: path`].
```
dgozman, line 116:
```suggestion
- `name` <[void]|[string]> Optional attachment name. If omitted, this will be inferred from [`param: path`].
```

docs/src/test-api/class-testinfo.md
```diff
@@ -62,10 +73,65 @@ test('basic test', async ({ page }, testInfo) => {
...
+
+### option: TestInfo.attach#1.contentType
+- `contentType` <[void]|[string]> Optional content type of this attachment to properly present in the report, for example `'application/json'` or `'image/png'`. If omitted, this falls back to an inferred type based on the [`param: name`] (if set) or [`param: path`]'s extension; it will be set to `application/octet-stream` if the type cannot be inferred from the file extension.
+
+
+## method: TestInfo.attach#2
+
+Similar to [`method: TestInfo.attach#1`] but attaches inline `strings` or `Buffers` instead of files from disk.
```
dgozman, line 124:
```suggestion
Attach data to the current test, either a `string` or a `Buffer`. Some reporters show test attachments.
```

dgozman (commented):
API looks very nice!

rwoll:
@dgozman I've address all review feedback. However, I'm tempted to revert https://github.com/microsoft/playwright/pull/10121#discussion_r749909228 since overall compared to the previous commit I found it:

  1. harder to read
 2. harder to author
 3. harder to debug a failing test
 4. More error prone since you're dealing with lots of output at once when doing the matching
 5. lots of repeated code

I made this change in response to https://github.com/microsoft/playwright/pull/10121#discussion_r749909228 but I feel like maybe I'm just misunderstanding something :)

dgozman (commented):
Sorry for not spotting this earlier, but our policy is "all optional arguments go to the options bag". The rest looks great!

docs/src/test-api/class-testinfo.md
```diff
@@ -62,10 +73,65 @@ test('basic test', async ({ page }, testInfo) => {
...
+
+Attach data to the current test, either a `string` or a `Buffer`. Some reporters show test attachments.
+
+### param: TestInfo.attach#2.body
+- `body` <[string]|[Buffer]> Attachment body.
+
+### param: TestInfo.attach#2.name
+- `name` <[string]> Attachment name.
+
+### param: TestInfo.attach#2.contentType
```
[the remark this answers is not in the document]
rwoll: Good catch! All fixed 🎉

dgozman (approved):

rwoll:
Test failures look unrelated to this changeset, but let me know if you read otherwise!

PR #10133, opened 2021-11-08
docs: grammar and markdown updates

MarcusFelling:
- Misc grammar updates
- Updated markdown to align with linter rules here: https://github.com/DavidAnson/markdownlint/blob/v0.24.0/doc/Rules.md#rules

Maybe we should consider adding markdown linting to [lint script](https://github.com/microsoft/playwright/blob/7278fcffb803ba188856cfb9631e4175f3953713/package.json#L28)?


docs/src/actionability.md
```diff
@@ -7,6 +7,7 @@ Playwright performs a range of actionability checks on the elements before makin
 behave as expected. It auto-waits for all the relevant checks to pass and only then performs the requested action. If the required checks do not pass within the given `timeout`, action fails with the `TimeoutError`.
 
 For example, for [`method: Page.click`], Playwright will ensure that:
+
```
pavelfeldman, line 10: Here and below, lists and titles are rendered in a separate paragraph automatically

docs/src/assertions.md
```diff
@@ -3,7 +3,7 @@ id: assertions
 title: "Assertions"
 ---
 
-Playwright provides convenience APIs for common tasks, like reading the
+Playwright provides convenient APIs for common tasks, like reading the
```
pavelfeldman, line 6: It is a term: https://www.w3.org/TR/DOM-Level-2-Core/glossary.html

docs/src/auth.md
```diff
@@ -160,8 +160,9 @@ implement **login once and run multiple scenarios**. The lifecycle looks like:
 This approach will also **work in CI environments**, since it does not rely on any external state.
 
 ### API reference
-- [`method: BrowserContext.storageState`]
-- [`method: Browser.newContext`]
+
+* [`method: BrowserContext.storageState`]
```
pavelfeldman, line 164: Why is this changing? I think it renders the same way in html.

docs/src/browsers.md
```diff
@@ -106,27 +106,27 @@ class Program
 
 ### When to use Google Chrome & Microsoft Edge and when not to?
 
-**Defaults**
+#### Defaults
```
pavelfeldman, line 109: This might become smaller than the actual text.

pavelfeldman (commented):
Started reviewing it. I would recommend splitting the grammar from the formatting changes and to keep the size of the PR reasonable. My understanding is that changes related to added / removed blank lines are a noop (in the best case), same goes for the bullet type. Our markdown is not just the markdown, it is markdown that is processed programmatically several times on its way to the rendered docs, so we need to be careful managing it.

PR #10138, opened 2021-11-08
feat(test-runner): wait for a url before starting tests

divdavem:
This PR allows the `webServer` configuration in @playwright/test to accept a url in addition to a port number to wait not only for a port to be opened but also for a url to return a 2xx status code.

### DIFF
--- a/packages/playwright-test/src/webServer.ts
+++ b/packages/playwright-test/src/webServer.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import http from 'http';
+import https from 'https';
 import net from 'net';
 import os from 'os';
 import stream from 'stream';
@@ -36,9 +38,12 @@ const newProcessLogPrefixer = () => new stream.Transform({
 const debugWebServer = debug('pw:webserver');
 
 export class WebServer {
+  private _isAvailable: () => Promise<boolean>;
   private _killProcess?: () => Promise<void>;
   private _processExitedPromise!: Promise<any>;
-  constructor(private readonly config: WebServerConfig) { }
+  constructor(private readonly config: WebServerConfig) {
+    this._isAvailable = getIsAvailableFunction(config);
+  }
 
   public static async create(config: WebServerConfig): Promise<WebServer> {
     const webServer = new WebServer(config);
@@ -56,11 +61,11 @@ export class WebServer {
     let processExitedReject = (error: Error) => { };
     this._processExitedPromise = new Promise((_, reject) => processExitedReject = reject);
 
-    const portIsUsed = await isPortUsed(this.config.port);
-    if (portIsUsed) {
+    const isAlreadyAvailable = await this._isAvailable();
+    if (isAlreadyAvailable) {
       if (this.config.reuseExistingServer)
         return;
-      throw new Error(`Port ${this.config.port} is used, make sure that nothing is running on the port or set strict:false in config.webServer.`);
+      throw new Error(`${this.config.url ?? `http://localhost:${this.config.port}`} is already used, make sure that nothing is running on the port/url or set strict:false in config.webServer.`);
     }
 
     const { launchedProcess, kill } = await launchProcess({
@@ -86,7 +91,7 @@ export class WebServer {
 
   private async _waitForProcess() {
     await this._waitForAvailability();
-    const baseURL = `http://localhost:${this.config.port}`;
+    const baseURL = this.config.url ?? `http://localhost:${this.config.port}`;
     process.env.PLAYWRIGHT_TEST_BASE_URL = baseURL;
   }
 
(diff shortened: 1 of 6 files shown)


ExFlo:
Would be nice, I have issue with angular slow build.

divdavem:
Our goal with this PR is to better integrate the Playwright test runner with the Angular CLI. It is done in a way that is not Angular CLI-specific and could be useful in similar cases too.

When we use "ng serve" as the web server command, because the Angular CLI opens the port immediately even before the compilation is finished, the Playwright test runner will immediately consider that the server is ready and will start the tests, and the compilation time will be counted as part of the first test in every worker. Those tests will probably work fine if the compilation is quick, because, anyway, the Angular CLI server waits for the compilation to be finished before returning the page. However, if the compilation is too long one of the first tests may time out, and, anyway, it does not seem right to count the compilation time as part of a test.

To avoid this, we could create a globalSetup file containing a function that only does a first request, but I think it would be nice to simplify this use case with the ability to provide in the webServer config of the Playwright test runner a URL to wait for (with a 2xx status code) before starting the tests.

That's what I have implemented in my PR. What do you think?

mxschmitt:
Since this is pretty-similar to create-playwright, we could also think about:
- adding `test-results/` and `playwright-report/` to the gitignore file
- showing some helpful getting-started output for the users: https://github.com/microsoft/create-playwright/blob/5f1a4ae278ce8007c7ae7e40b8909335c4a58577/src/generator.ts#L164-L189

Good progress so far! I think there is not much missing to release a first version of it.

divdavem:
@mxschmitt Your comment seems to be related to [another PR](https://github.com/pkozlowski-opensource/playwright-schematic/pull/1)
I would be glad if you could review this PR too! Thank you in advance!

mxschmitt:
> @mxschmitt Your comment seems to be related to [another PR](https://github.com/pkozlowski-opensource/playwright-schematic/pull/1) I would be glad if you could review this PR too! Thank you in advance!

oh good point! Absolutely, I put this PR already on my list and will talk with the team later about it. Will keep you updated.

divdavem:
> > @mxschmitt Your comment seems to be related to [another PR](https://github.com/pkozlowski-opensource/playwright-schematic/pull/1) I would be glad if you could review this PR too! Thank you in advance!
> 
> oh good point! Absolutely, I put this PR already on my list and will talk with the team later about it. Will keep you updated.

Great! Thank you very much!

utils/generate_types/overrides-test.d.ts
```diff
@@ -77,6 +77,11 @@ export type WebServerConfig = {
    * The port that your http server is expected to appear on. It does wait until it accepts connections.
    */
   port: number,
+  /**
+   * The url on your http server that is expected to return a 2xx status code when the server is ready to accept connections.
+   * If the url is not specified, only the port is checked.
+   */
+  url?: string,
```
dgozman, line 84: - Let's make `port` optional.
- Please update comment for both port and url to mention that exactly one of them is required, and align the implementation.
- Please update reuseExistingServer comment to equally mention port and url.
- Please also update webServer section in `docs/src/test-api/class-testconfig.md`.

packages/playwright-test/src/webServer.ts
```diff
@@ -118,11 +119,32 @@ async function isPortUsed(port: number): Promise<boolean> {
   });
 }
 
-async function waitForSocket(port: number, delay: number, cancellationToken: { canceled: boolean }) {
+async function isURLAvailable(url: URL) {
+  return new Promise<boolean>(resolve => {
+    http.get(url, res => {
```
dgozman, line 124: Let's support https as well.

packages/playwright-test/src/webServer.ts
```diff
@@ -118,11 +119,32 @@ async function isPortUsed(port: number): Promise<boolean> {
...
-    const connected = await isPortUsed(port);
+    const connected = await waitFn();
     if (connected)
       return;
     await new Promise(x => setTimeout(x, delay));
   }
 }
+
+async function waitForSocketOrURL(port: number, path: string | undefined, delay: number, cancellationToken: { canceled: boolean }) {
+  if (path) {
```
dgozman, line 144: Let's check here that exactly one of `port` and `path` is undefined, and the other is not.

dgozman (commented):
Very nice! Looks good overall, just a few comments.

mxschmitt:
maybe we should get rid of the isPortUsed logic all together and move it fully to the URL based approach (2xx).
cc @dgozman

divdavem:
@dgozman Thank you for your review. I have rebased my PR and changed the code accordingly.

dgozman (approved):
Great PR, thank you!

divdavem:
@dgozman Thank you for merging my PR!

PR #10145, opened 2021-11-08
Chromium video duration fix

### LINKED ISSUE #9611: [Bug]: Video duration is different than real duration of the run (chromium)
### Playwright version

1.15.3

### Operating system

Linux

Chromium

### Other information

Repo to present bug:
https://github.com/qwark97/video_length_bug

Given command in the repo takes ~2min

Bug occurs also outside the docker image

To be honest I'm not sure if this is a bug or an expected behavior but videos recorded with playwright (by passing `record_video_dir`) at Chromium, at both headless/headed mode, with and without `slow_mo` parameter are longer than "real" execution time. For example, one of the executions of the test (from the example repo) took ~18s while recorded video took 36s (such 2x ratio is just a coincidence)

I haven't noticed such behavior with other browsers but on the other hand the same test took at them quite longer time (I don't know whether it is connected)

So my question is: if this is a bug? And if not, am I able to do anything to make execution and the recorded video the same duration?

_No response_

### Relevant log output

_No response_

gpiszczek:
Videos recorded with playwright using chromium engine are longer than what could be expected by measuring execution time. It is caused by more than one problem:
- chromium can generate frames faster than what is expected by playwright which uses hardcoded 25 fps,
- timestamp of last added frame is taken from timestamp reported by browser which is not in 1/fps intervals,
- the first screencast frame is send by browser with slight delay from the start of screencast.

Fixes #9611

### DIFF
--- a/packages/playwright-core/src/server/chromium/videoRecorder.ts
+++ b/packages/playwright-core/src/server/chromium/videoRecorder.ts
@@ -141,10 +141,12 @@ export class VideoRecorder {
       for (let i = 0; i < repeatCount; ++i)
         this._frameQueue.push(this._lastFrameBuffer);
       this._lastWritePromise = this._lastWritePromise.then(() => this._sendFrames());
+      this._lastFrameTimestamp += repeatCount / fps;
+    } else {
+      this._lastFrameTimestamp = timestamp;
     }
 
     this._lastFrameBuffer = frame;
-    this._lastFrameTimestamp = timestamp;
     this._lastWriteTimestamp = monotonicTime();
   }
 
--- a/packages/playwright-core/src/server/chromium/videoRecorder.ts
+++ b/packages/playwright-core/src/server/chromium/videoRecorder.ts
@@ -27,6 +27,7 @@ const fps = 25;
 export class VideoRecorder {
   private _process: ChildProcess | null = null;
   private _gracefullyClose: (() => Promise<void>) | null = null;
+  private _launchTimestamp: number = 0;
   private _lastWritePromise: Promise<void> = Promise.resolve();
   private _lastFrameTimestamp: number = 0;
   private _lastFrameBuffer: Buffer | null = null;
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/chromium/videoRecorder.ts
```diff
@@ -123,12 +124,17 @@ export class VideoRecorder {
...
     this._process = launchedProcess;
     this._gracefullyClose = gracefullyClose;
+    this._launchTimestamp = monotonicTime();
   }
 
   writeFrame(frame: Buffer, timestamp: number) {
     assert(this._process);
     if (this._isStopped)
       return;
+    if (timestamp - this._lastFrameTimestamp < 1 / fps) {
```
yury-s, line 134: With the new backpressure mechanism added in https://github.com/microsoft/playwright/pull/9893 this should not be necessary. If the throttler doesn't give desired effect we should adjust its parameters instead. The difference with this approach is that Chromium won't send new screencast frames until it gets acks for already sent once whereas with the approach in this PR the frames are still generated just to be dropped on the floor on the playwright end.

Can you check if the problem is still reproducible on @next version (which includes the fix mentioned above)?

packages/playwright-core/src/server/chromium/videoRecorder.ts
```diff
@@ -137,10 +143,12 @@ export class VideoRecorder {
       for (let i = 0; i < repeatCount; ++i)
         this._frameQueue.push(this._lastFrameBuffer);
       this._lastWritePromise = this._lastWritePromise.then(() => this._sendFrames());
+      this._lastFrameTimestamp += repeatCount / fps;
```
yury-s, line 146: This shouldn't make any difference as the number of frames we feed into ffmpeg stays the same or am I missing something?

packages/playwright-core/src/server/chromium/videoRecorder.ts
```diff
@@ -137,10 +143,12 @@ export class VideoRecorder {
...
+    } else {
+      this._lastFrameTimestamp = timestamp - (monotonicTime() - this._launchTimestamp) / 1000;
```
yury-s, line 148: This will only increase the duration of the resulting video by the time it took to launch screencast, why do we want this?

mxschmitt:
Closing since it looks stale and as part of the triage process! Feel free to re-open / create a new PR if you are still interested to work on it.

PR #10161, opened 2021-11-09
docs: expand core-concepts and multi-age

pavelfeldman:


docs/src/evaluating.md
```diff
@@ -0,0 +1,372 @@
...
+The [`method: Page.evaluate`] API can run a JavaScript function in the context
+of the web page and bring results back to the Playwright environment. Browser globals like
+`window` and `document` can be used in `evaluate`.
+
+```js
+const href = await page.evaluate(() => document.location.href);
+```
+
+```java
+String href = (String) page.evaluate("document.location.href");
```
dgozman, line 17: Why string and not a function like other languages?

docs/src/evaluating.md
```diff
@@ -0,0 +1,372 @@
...
+```python async
+href = await page.evaluate('() => document.location.href')
+```
+
+```python sync
+href = page.evaluate('() => document.location.href')
+```
+
+```csharp
+var href = await page.EvaluateAsync<string>("document.location.href");
```
dgozman, line 29: ditto

docs/src/pages.md
```diff
@@ -1,111 +1,88 @@
 ---
-id: multi-pages
-title: "Multi-page scenarios"
+id: pages
+title: "Pages and frames"
```
dgozman, line 3: This looks like just "Pages"?

dgozman (approved):

docs/src/evaluating.md
pavelfeldman: I don't know!

docs/src/pages.md
pavelfeldman: Fixed.

PR #10180, opened 2021-11-09
chore: api testing test nits

mxschmitt:


packages/playwright-core/src/server/fetch.ts
```diff
@@ -155,7 +155,7 @@ export abstract class APIRequestContext extends SdkObject {
         return { error: `${fetchResponse.status} ${fetchResponse.statusText}` };
       return { fetchResponse: { ...fetchResponse, fetchUid } };
     } catch (e) {
-      return { error: String(e) };
+      return { error: e instanceof Error ? e.message : String(e) };
```
mxschmitt, line 158: this fixes an issue that the actual error got wrapped as `Error: actual error` and landed by that not in the ideal form in the language ports.

tests/page/page-network-response.spec.ts
```diff
@@ -267,3 +268,12 @@ it('should behave the same way for headers and allHeaders', async ({ page, serve
   expect(allHeaders['name-b']).toEqual('v4');
 });
 
+it('should should provide a Response with a file URL', async ({ page, asset, isAndroid, browserName }) => {
```
yury-s, line 271: typo: should should

tests/global-fetch.spec.ts
```diff
@@ -41,27 +42,26 @@ it.afterAll(() => {
 for (const method of ['fetch', 'delete', 'get', 'head', 'patch', 'post', 'put']) {
   it(`${method} should work`, async ({ playwright, server }) => {
     const request = await playwright.request.newContext();
-    const response = await request[method](server.PREFIX + '/simple.json');
+    const response: APIResponse = await request[method](server.PREFIX + '/simple.json');
```
yury-s, line 45: why do we need this? isn't tsc smart enough?

packages/playwright-core/src/server/fetch.ts
yury-s: I believe we are losing backand's stack trace in this case and it often provides useful information about what's going on (at the very least for us).

packages/playwright-core/src/client/fetch.ts
```diff
@@ -214,7 +214,7 @@ export class APIResponse implements api.APIResponse {
   }
 
   ok(): boolean {
-    return this._initializer.status === 0 || (this._initializer.status >= 200 && this._initializer.status <= 299);
+    return (this._initializer.status >= 200 && this._initializer.status <= 299);
```
yury-s, line 217: add a test with file:// url ?

yury-s (approved):
I'm not sure about the change in error formatting, otherwise looks good.

tests/global-fetch.spec.ts
mxschmitt: good catch. I fixed it the other way around by adding `as const` to the for loop values.

packages/playwright-core/src/server/fetch.ts
mxschmitt: Seems like it never added the stack trace with the previous implementation: 
image

packages/playwright-core/src/client/fetch.ts
mxschmitt: added one. For our api testing we don't support the 'file:' protocol.

PR #10156, opened 2021-11-08
feat(webkit): roll to 1574

aslushnikov:

PR #10193, opened 2021-11-09
browser(webkit): disable GPU process on Windows only

aslushnikov:
- GPU Process was enabled by default on Windows recently: https://github.com/WebKit/WebKit/commit/2058f9454308a3a175cefcfbe71695721789e97e
- We tried it and it fails a bunch of tests: https://github.com/microsoft/playwright/pull/10156
- Curiously, disabling it on Mac fixes a mac-related bug: https://github.com/microsoft/playwright/pull/10170, however we're scared of the change


yury-s (approved):
Let's add some notes to the description of the PR so that we know why it was disabled.

PR #10220, opened 2021-11-10
docs(element-handle): discourage the element handle use

pavelfeldman:


yury-s (approved):

docs/src/assertions.md
```diff
@@ -131,27 +131,27 @@ Assert.True(checked);
 ## JS expression
 
 ```js
```
dgozman, line 133: Should we follow up with removing this file entirely from JavaScript docs? Same goes for other languages that implement web-first assertions?

dgozman (approved):

docs/src/assertions.md
pavelfeldman: Yeah, it is already hidden-ish, but yes.

Issue #9855, opened 2021-10-28
[BUG] page.click clicks incorrect item due to element reuse


sidharthv96:
**Context:**
## System:
 - OS: macOS 11.3
 - Memory: 216.95 MB / 16.00 GB
## Binaries:
 - Node: 14.17.6 - ~/.volta/tools/image/node/14.17.6/bin/node
 - Yarn: 1.22.11 - ~/.volta/tools/image/yarn/1.22.11/bin/yarn
 - npm: 6.14.15 - ~/.volta/tools/image/node/14.17.6/bin/npm
 - playwright: 1.16.2
## Languages:
 - Bash: 3.2.57 - /bin/bash

**Code Snippet**

Help us help you! Put down a short code snippet that illustrates your bug and
that we can run and debug locally. For example:

```javascript
const {chromium, webkit, firefox} = require('playwright');

(async () => {
  const browser = await chromium.launch();
  // ...
})();
```

**Describe the bug**

I found a race condition when clicking an element.
Should we be throwing an error/warning instead of clicking the wrong element?
When Page is Loaded, table is loaded unsorted.

| Item | Date      |
| ---- | --------- |
| One  | `<dateOne>` |
| Two  | `<dateTwo>` |

After Sort

| Item | Date      |
| ---- | --------- |
| Two  | `<dateTwo>` |
| One  | `<dateOne>`|

Problem 
```ts
await page.click('"Date"'); // To sort by date. This triggers a network call to fetch the data.
await page.click('"Two"'); // Expected to click "Two", but clicks "One", without any error/warning
```

Workaround

```ts
await Promise.all([
  this.page.waitForResponse(/-Date/), // Based on the URL
  this.page.click('"Date"'),
]);
await page.click('"Two"');
```

If it was clicking `Two` sorted/unsorted, it's not an issue. But it's clicking `One` after sort, without any kind of warning/error. This was happening in an unrelated area of the tests, so had to dig in to find the root cause.

karesztrk:
Hey guys. I'm experiencing the same issue inside a `<table>` rendered by React ^18.1.0
@sidharthv96 's workaround indeed solves the problem.
However im interested in about how to overcome with this in React since im already using `key=` for each row inside the table. Any advice @dgozman ?

pavelfeldman:
#### Why was this issue closed?

[...]

PR #10226, opened 2021-11-10
fix(actions): guard against element reuse

dgozman:
Query the selector when polling during stable check and hit target testing.
When strict mode violation is encountered, throw.
When selector resolves to a different element, retry.

This should prevent "element reuse" race where framework (e.g. React or Svelte)
reuses the same DOM element for a semantically different purpose, for example a different list item.

References #9855.


pavelfeldman:
Closing as obsolete

PR #10231, opened 2021-11-10
fix(html-report): fix command to open HTML report

aslushnikov:


packages/playwright-test/src/reporters/html.ts
```diff
@@ -147,7 +147,7 @@ class HtmlReporter implements Reporter {
       console.log('');
       console.log('To open last HTML report run:');
       console.log(colors.cyan(`
-  npx playwright show-report
+  npx playwright show-report ${this._outputFolder || ''}
```
pavelfeldman, line 150: Resolve outputFolder relative to cwd

pavelfeldman (approved):

pavelfeldman (approved):

PR #10234, opened 2021-11-11
docs/test-config: add github, html reporter

Nav-2d:
Update test config api docs to add html and github reporter


dgozman (approved):
Wonderful!

PR #10240, opened 2021-11-11
chore: fix API name capitalisation with APIRequestContext

### LINKED ISSUE #10182: [BUG] wrong API name casing with request API
```typescript 
 const error = await context.request.get(server.PREFIX + '/does-not-exist.html', {
      failOnStatusCode: true
    }).catch(e => e);
    expect(error.message).toBe('APIRequestContext.get: 404 Not Found');
```

results in: `aPIRequestContext.get: 404 Not Found`
expected: `APIRequestContext.get: 404 Not Found` 

cp: https://github.com/microsoft/playwright/blob/ceedf45d4a798fd0b1a10751d26d93679a9fe27d/packages/playwright-core/src/utils/stackTrace.ts#L117

mxschmitt:
Fixes #10182

### DIFF
--- a/tests/browsercontext-fetch.spec.ts
+++ b/tests/browsercontext-fetch.spec.ts
@@ -69,7 +69,7 @@ it('should throw on network error', async ({ context, server }) => {
     req.socket.destroy();
   });
   const error = await context.request.get(server.PREFIX + '/test').catch(e => e);
-  expect(error.message).toContain('socket hang up');
+  expect(error.message).toBe('apiRequestContext.get: socket hang up');
 });
 
 it('should throw on network error after redirect', async ({ context, server }) => {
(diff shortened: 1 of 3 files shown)


tests/browsercontext-fetch.spec.ts
```diff
@@ -78,7 +78,7 @@ it('should throw on network error after redirect', async ({ context, server }) =
     req.socket.destroy();
   });
   const error = await context.request.get(server.PREFIX + '/redirect').catch(e => e);
-  expect(error.message).toContain('socket hang up');
+  expect(error.message).toBe('APIRequestContext.get: socket hang up');
```
yury-s, line 81: this is inconsistent with our docs where we use `apiRequestContext.get`, let's apply the same logic as in the [docgen](https://github.com/microsoft/playwright/blob/38dcfd785cc1e47bd757d49f6e9ed69859456dc8/utils/doclint/documentation.js#L170) ?

yury-s (approved):

PR #10243, opened 2021-11-11
chore: add trace viewer file upload error handling

### LINKED ISSUE #10127: Trace viewer should print error when the npx playwright show-trace foo.zip does not exist
Got a few reports about that on Slack and on GitHub already.

mxschmitt:
This fixes: `npx playwright show-trace non-existing.zip`

Fixes #10127

### DIFF
--- a/packages/playwright-core/src/web/traceViewer/sw.ts
+++ b/packages/playwright-core/src/web/traceViewer/sw.ts
@@ -77,7 +77,7 @@ async function doFetch(event: FetchEvent): Promise<Response> {
         const traceFileName = url.searchParams.get('traceFileName')!;
         return new Response(JSON.stringify({
           error: traceFileName ? `Could not load trace from ${traceFileName}. Make sure to upload a valid Playwright trace.` :
-            `Could not load trace from ${traceUrl}. Make sure valid Playwright Trace is accessible over this url.`,
+            `Could not load trace from ${traceUrl}. Make sure a valid Playwright Trace is accessible over this url.`,
         }), {
           status: 500,
           headers: { 'Content-Type': 'application/json' }
--- a/packages/playwright-core/src/web/traceViewer/sw.ts
+++ b/packages/playwright-core/src/web/traceViewer/sw.ts
@@ -60,17 +60,29 @@ async function doFetch(event: FetchEvent): Promise<Response> {
       return new Response(null, { status: 200 });
     }
 
-    const traceUrl = new URL(url).searchParams.get('trace')!;
+    const traceUrl = url.searchParams.get('trace')!;
     const { snapshotServer } = loadedTraces.get(traceUrl) || {};
 
     if (relativePath === '/context') {
-      const traceModel = await loadTrace(traceUrl, event.clientId, (done: number, total: number) => {
-        client.postMessage({ method: 'progress', params: { done, total } });
-      });
-      return new Response(JSON.stringify(traceModel!.contextEntry), {
-        status: 200,
-        headers: { 'Content-Type': 'application/json' }
-      });
+      try {
+        const traceModel = await loadTrace(traceUrl, event.clientId, (done: number, total: number) => {
+          client.postMessage({ method: 'progress', params: { done, total } });
+        });
+        return new Response(JSON.stringify(traceModel!.contextEntry), {
+          status: 200,
+          headers: { 'Content-Type': 'application/json' }
+        });
+      } catch (error: unknown) {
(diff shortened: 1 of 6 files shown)


pavelfeldman:
Looks like #10127 can have a simpler fix. What exact scenario are we fixing here?

packages/playwright-core/src/web/traceViewer/ui/workbench.tsx
```diff
@@ -65,6 +68,12 @@ export const Workbench: React.FunctionComponent<{
     processTraceFile(event.target.files[0]);
   };
 
+  React.useEffect(() => {
```
pavelfeldman, line 71:
```suggestion
  // When a blob was open -> the blob URL ended up in the URL. If the user did a reload there the
  // blob got revoked and by that it ended up in an error. -> do not use blobs when pw trace opened initially
  React.useEffect(() => {
```

packages/playwright-core/src/web/traceViewer/ui/snapshotTab.tsx
```diff
@@ -65,6 +65,8 @@ export const SnapshotTab: React.FunctionComponent<{
         const info = await response.json();
         if (!info.error)
           setSnapshotInfo(info);
+      } else {
+        setSnapshotInfo(defaultSnapshotInfo);
```
mxschmitt, line 69: If an error occurred when a trace was already loaded before, this cleans up the old snapshot URL which was shown.

packages/playwright-core/src/web/traceViewer/sw.ts
```diff
@@ -37,11 +37,18 @@ async function loadTrace(trace: string, clientId: string, progress: (done: numbe
...
+  const fetchFromInbuiltTraceViewerServer = !trace.startsWith('http') && !trace.startsWith('blob');
+  let url = fetchFromInbuiltTraceViewerServer ? `file?path=${trace}` : trace;
   // Dropbox does not support cors.
   if (url.startsWith('https://www.dropbox.com/'))
     url = 'https://dl.dropboxusercontent.com/' + url.substring('https://www.dropbox.com/'.length);
-  await traceModel.load(url, progress);
+  try {
+    await traceModel.load(url, progress);
+  } catch (error) {
+    if (fetchFromInbuiltTraceViewerServer && error instanceof Error && error.message === 'HTTP error Not Found')
```
mxschmitt, line 48: since local file URLs also get loaded over HTTP, check if its done via the internal trace viewer web-server and rewrite it that the local file could not be found (instead of showing the http error).

packages/playwright-core/src/web/traceViewer/ui/snapshotTab.tsx
pavelfeldman: Put it in as a comment!

packages/playwright-core/src/web/traceViewer/sw.ts
```diff
@@ -64,13 +71,23 @@ async function doFetch(event: FetchEvent): Promise<Response> {
...
-      });
+      try {
+        const traceModel = await loadTrace(traceUrl, event.clientId, (done: number, total: number) => {
+          client.postMessage({ method: 'progress', params: { done, total } });
+        });
+        return new Response(JSON.stringify(traceModel!.contextEntry), {
+          status: 200,
+          headers: { 'Content-Type': 'application/json' }
+        });
+      } catch (error: unknown) {
```
pavelfeldman, line 82: This is the only try/catch you should need, remove all the others?

packages/playwright-core/src/web/traceViewer/ui/workbench.tsx
```diff
@@ -65,6 +68,13 @@ export const Workbench: React.FunctionComponent<{
...
+    const newTraceURL = new URL(window.location.href).searchParams.get('trace');
+    // Don't re-use blob file URLs on page load (results in Fetch error)
+    if (newTraceURL && !newTraceURL.startsWith('blob:'))
+      setTraceURL(newTraceURL);
+  }, [setTraceURL]);
```
pavelfeldman, line 76: What does it mean useEffect around setTraceURL, wouldn't it run only once?

mxschmitt: Yes, as the comment says its about if you open trace viewer with a blob URL. Usually old tab, forced reload etc.

pavelfeldman (approved):

PR #10260, opened 2021-11-11
chore: clean up docker image

aslushnikov:
This patch:
- removes the Visual Regression bits from the images. These will
  probably be added back when it's ready for the prime time.
    * removes VNC && noVNC
    * removes Playwright Agent
- cleans up apt caches
- uses openssh-client instead of a full ssh server

This drops image size from 2.03GB to 1.91GB.

### DIFF
--- a/utils/docker/Dockerfile.bionic
+++ b/utils/docker/Dockerfile.bionic
@@ -22,10 +22,10 @@ RUN apt-get update && \
     update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 1 && \
 
     # clean apt cache
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* && \
 
-# Create the pwuser (we internally create a symlink for the pwuser and the root user)
-RUN adduser pwuser
+    # Create the pwuser (we internally create a symlink for the pwuser and the root user)
+    adduser pwuser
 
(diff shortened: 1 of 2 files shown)


dgozman (approved):

utils/docker/Dockerfile.bionic
```diff
@@ -5,41 +5,27 @@ ARG TZ=America/Los_Angeles
...
-  && unzip -x noVNC.zip \
-  && mv noVNC-${NOVNC_REF} /opt/bin/noVNC \
-  && cp /opt/bin/noVNC/vnc.html /opt/bin/noVNC/index.html \
-  && rm noVNC.zip \
-  && curl -L -o websockify.zip "https://github.com/novnc/websockify/archive/v${WEBSOCKIFY_REF}.zip" \
-  && unzip -x websockify.zip \
-  && rm websockify.zip \
-  && mv websockify-${WEBSOCKIFY_REF} /opt/bin/noVNC/utils/websockify
+# Create the pwuser (we internally create a symlink for the pwuser and the root user)
+RUN adduser pwuser
```
mxschmitt, line 28: nit: maybe also put that into the last RUN statement. Then its 1 layer less.

mxschmitt (approved):

utils/docker/Dockerfile.bionic
aslushnikov: merged into the first RUN statement

PR #10292, opened 2021-11-12
devops: attempt to switch to LF by default

aslushnikov:


.gitattributes
```diff
@@ -1,3 +1,3 @@
 # text files must be lf for golden file tests to work
-*.txt eol=lf
-*.json eol=lf
+* text eol=lf
```
mxschmitt, line 2: I had back then `* text=auto eol=lf` not sure why, but it was passing for me: https://github.com/microsoft/playwright/pull/7987

mxschmitt (approved):

mxschmitt:
We should probably also align the new gitattributes file across the other repos. For Python e.g. we already use LF partly: https://github.com/microsoft/playwright-python/blob/main/.gitattributes

.gitattributes
pavelfeldman: Yep, reading the docs it looks like text=auto is what we want!

pavelfeldman (approved):

PR #10355, opened 2021-11-16
docs/test-configuration: Update code snippet

### LINKED ISSUE #10322: [BUG] Code snippet from documentation produces an error
When I run the code example provided in the documentation: https://playwright.dev/docs/test-configuration/ 
then it produces an error

```
// example.spec.js

test.describe('headed block', () => {
  // Run tests in this describe block in headed mode.
  test.use({ headless: false });

  test('my headed test', async ({ page }) => {
    // ...
  });
});
```

**Output:**
```
Error: Cannot use({ headless }) in a describe group, because it forces a new worker.
Make it top-level in the test file or put in the configuration file.
```
Reproduced on v1.16

Nav-2d:
Fixed #10322

### DIFF
--- a/docs/src/test-configuration-js.md
+++ b/docs/src/test-configuration-js.md
@@ -92,11 +92,11 @@ The same works inside describe.
 // example.spec.js
 const { test, expect } = require('@playwright/test');
 
-test.describe('headed block', () => {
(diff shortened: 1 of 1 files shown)


dgozman (approved):
👍

PR #10389, opened 2021-11-17
chore: use channel traits

pavelfeldman:


packages/playwright-core/src/dispatchers/playwrightDispatcher.ts
```diff
@@ -29,7 +29,8 @@ import { ElectronDispatcher } from './electronDispatcher';
 import { APIRequestContextDispatcher } from './networkDispatchers';
 import { SelectorsDispatcher } from './selectorsDispatcher';
 
-export class PlaywrightDispatcher extends Dispatcher<Playwright, channels.PlaywrightInitializer, channels.PlaywrightEvents> implements channels.PlaywrightChannel {
+export class PlaywrightDispatcher extends Dispatcher<Playwright, channels.PlaywrightChannel> implements channels.PlaywrightChannel {
+  _type_Playwright;
```
dgozman, line 33: Why not `_type_Playwright = true` like in other places?

dgozman (approved):

packages/playwright-core/src/dispatchers/playwrightDispatcher.ts
pavelfeldman: It does not compile :/

PR #10392, opened 2021-11-17
chore: fix small typo on https://playwright.dev/docs/test-parameterize

alisterscott:
Fixes a small typo in the doco https://playwright.dev/docs/test-parameterize

I came here to fix `for (const name in people) {` to `for (const name of people) {` but it was already fixed in master

### DIFF
--- a/docs/src/test-parameterize-js.md
+++ b/docs/src/test-parameterize-js.md
@@ -7,7 +7,7 @@ You can either parameterize tests on a test level or on a project level.
 
 <!-- TOC -->
 
-## Parametrized Tests
+## Parameterized Tests
 
 ```js js-flavor=js
 // example.spec.js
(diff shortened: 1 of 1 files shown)


docs/src/test-parameterize-js.md
```diff
@@ -7,7 +7,7 @@ You can either parameterize tests on a test level or on a project level.
 
 <!-- TOC -->
 
-## Parametrized Tests
+## Parameterized Tests
```
aslushnikov, line 10: Aren't both `Parametrized` and `Parameterized` legit?

alisterscott: I don't think `Parametrized` is actually a word - I tried googling it

mxschmitt: Parameterized seems more often used, so landing this patch.

PR #10412, opened 2021-11-18
chore: fix CLI --help when used in driver

### LINKED ISSUE #10411: [Bug]: Help text usage prints "npx playwright" instead of "playwright" for Python
### Playwright version

1.16.1

### Operating system

MacOS

_No response_

### Other information

_No response_

I installed Playwright for Python on macOS. I checked the help text for the `playwright` command, as shown below:

```
$ playwright --help
Usage: npx playwright [options] [command]

Options:
  -V, --version                          output the version number
  -h, --help                             display help for command

Commands:
  open [options] [url]                   open page in browser specified via -b, --browser
  codegen [options] [url]                open page and generate code for user actions
  install [options] [browser...]         ensure browsers necessary for this version of Playwright are installed
  install-deps [browser...]              install dependencies necessary to run browsers (will ask for sudo permissions)
  cr [options] [url]                     open page in Chromium
  ff [options] [url]                     open page in Firefox
  wk [options] [url]                     open page in WebKit
  screenshot [options] <url> <filename>  capture a page screenshot
  pdf [options] <url> <filename>         save page as pdf
  show-trace [options] [trace]           Show trace viewer
  help [command]                         display help for command
```

This line is the problem: ```Usage: npx playwright [options] [command]```. For Python, the `playwright` command should not have `npx`. I recognize that `npx` comes from JavaScript, but it could be confusing for a pure Python user.

I also noticed this problem for the `playwright install --help` command. I haven't tried other commands, but I would presume others would be affected, too.

_No response_

### Relevant log output

_No response_

mxschmitt:
- https://github.com/microsoft/playwright-dotnet/blob/b93078851f9ccbb88eeea3d1a3c5c4d7d9085dea/src/Playwright/Program.cs#L67
- https://github.com/microsoft/playwright-java/blob/4be749f045af2211fabda9d0f7f65710ef9b491d/playwright/src/main/java/com/microsoft/playwright/CLI.java#L36
- https://github.com/microsoft/playwright-python/blob/5b7bb1972222f0fc3b55a373cd0f4b8edb8eaa5f/playwright/__main__.py#L25

we set it in all the places already, so let's use it.

Fixes #10411

### DIFF
--- a/packages/playwright-core/src/cli/cli.ts
+++ b/packages/playwright-core/src/cli/cli.ts
@@ -40,7 +40,7 @@ const packageJSON = require('../../package.json');
 
 program
     .version('Version ' + packageJSON.version)
-    .name(process.env.PW_CLI_NAME || 'npx playwright');
+    .name(buildBasePlaywrightCLICommand(process.env.PW_CLI_TARGET_LANG));
 
 commandWithOpenOptions('open [url]', 'open page in browser specified via -b, --browser', [])
     .action(function(url, options) {
@@ -601,4 +601,17 @@ async function launchGridServer(factoryPathOrPackageName: string, port: number, 
   const gridServer = new GridServer(factory as GridFactory, authToken);
   await gridServer.start(port);
   console.log('Grid server is running at ' + gridServer.urlPrefix());
+}
+
+function buildBasePlaywrightCLICommand(cliTargetLang: string | undefined): string {
+  switch (cliTargetLang) {
+    case 'python':
+      return `playwright`;
+    case 'java':
+      return `mvn exec:java -e -Dexec.mainClass=com.microsoft.playwright.CLI -Dexec.args="...options.."`;
+    case 'csharp':
+      return `playwright`;
+    default:
+      return `npx playwright`;
+  }
 }


packages/playwright-core/src/utils/registry.ts
```diff
@@ -709,16 +709,17 @@ function markerFilePath(browserDirectory: string): string {
...
-export function buildPlaywrightCLICommand(sdkLanguage: string, parameters: string): string {
+export function buildPlaywrightCLICommand(sdkLanguage?: string, parameters?: string): string {
+  const parametersWithSpacePrefix = parameters ? (' ' + parameters) : '';
   switch (sdkLanguage) {
     case 'python':
-      return `playwright ${parameters}`;
+      return `playwright${parametersWithSpacePrefix}`;
     case 'java':
-      return `mvn exec:java -e -Dexec.mainClass=com.microsoft.playwright.CLI -Dexec.args="${parameters}"`;
+      return `mvn exec:java -e -Dexec.mainClass=com.microsoft.playwright.CLI -Dexec.args="${parameters ? parameters : '...options..'}"`;
```
dgozman, line 718: What is this '...options..'? Why don't we omit "exec.args" entirely?

mxschmitt: I thought about it. Then Maven users don't know where to put the args, since in Java its non-trivial in my mind.

Do you have a better proposal instead of `...options...`?

dgozman: Oh, I see. I'd prefer this logic to be in the CLI instead of the generic utility then.

dgozman (approved):

PR #10434, opened 2021-11-19
chore: support range requests in trace viewer http server

### LINKED ISSUE #10333: [BUG] cannot select playback time on html report video
**Context:**
- Playwright Version: 1.16.3
- Operating System: Mac
- Node.js version: 17.0.1
- Browser: Chromium

**Code Snippet**

```typescript

test("example", async ({ page }) => {
	await page.goto("https://www.google.com");
	const s = page.locator("input[name=q]");
	await s.fill("example");
	await page.waitForTimeout(5000);
	await s.fill("overwritten");
	await page.waitForTimeout(5000);
});

```

**Describe the bug**

When trying to select other times (rewind/forward) on the video, the clip continues to play from its current location.

When I play the video in `test-results`, I can rewind/forward it as expected

https://user-images.githubusercontent.com/20273647/141875252-7d8f47fe-f6d7-474b-83fb-fa12174f9dae.mp4

mxschmitt:
Fixes #10333

### DIFF
--- a/packages/playwright-core/src/utils/httpServer.ts
+++ b/packages/playwright-core/src/utils/httpServer.ts
@@ -106,30 +106,32 @@ export class HttpServer {
   _serveRangeFile(request: http.IncomingMessage, response: http.ServerResponse, absoluteFilePath: string) {
     const range = request.headers.range!;
     const size = fs.statSync(absoluteFilePath).size;
-    // Extracting Start and End value from Range Header
+    // Parse the range header: https://datatracker.ietf.org/doc/html/rfc7233#section-2.1
     const [startStr, endStr] = range.replace(/bytes=/, '').split('-');
     let start = parseInt(startStr, 10);
-    let end = endStr ? parseInt(endStr, 10) : size - 1;
+    let end = parseInt(endStr, 10);
 
+    // No end specified: use the whole file
     if (!isNaN(start) && isNaN(end)) {
       start = start;
       end = size - 1;
     }
+    // No start specified: calculate start manually
     if (isNaN(start) && !isNaN(end)) {
       start = size - end;
       end = size - 1;
     }
 
     // Handle unavailable range request
-    if (start >= size || end >= size) {
-      // Return the 416 Range Not Satisfiable.
+    if (start >= size || end >= size || start > end) {
+      // Return the 416 Range Not Satisfiable: https://datatracker.ietf.org/doc/html/rfc7233#section-4.4
       response.writeHead(416, {
         'Content-Range': `bytes */${size}`
       });
-      response.end();
+      return response.end();
     }
 
-    // Sending Partial Content With HTTP Code 206
+    // Sending Partial Content: https://datatracker.ietf.org/doc/html/rfc7233#section-4.1
     response.writeHead(206, {
       'Content-Range': `bytes ${start}-${end}/${size}`,
       'Accept-Ranges': 'bytes',
--- a/packages/playwright-core/src/utils/httpServer.ts
+++ b/packages/playwright-core/src/utils/httpServer.ts
@@ -104,26 +104,34 @@ export class HttpServer {
   }
 
   _serveRangeFile(request: http.IncomingMessage, response: http.ServerResponse, absoluteFilePath: string) {
-    const range = request.headers.range!;
-    const size = fs.statSync(absoluteFilePath).size;
+    const range = request.headers.range;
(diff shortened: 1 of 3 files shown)


JoelEinbinder:
Is it feasible for us to write a test for this? Code looks fine but quite tricky.

packages/playwright-core/src/utils/httpServer.ts
```diff
@@ -80,22 +80,69 @@ export class HttpServer {
...
+    response.setHeader('Content-Type', contentType);
+    response.setHeader('Content-Length', content.byteLength);
+    response.end(content);
+  }
+
+  _serveRangeFile(request: http.IncomingMessage, response: http.ServerResponse, absoluteFilePath: string) {
+    const range = request.headers.range!;
+    const size = fs.statSync(absoluteFilePath).size;
+    // Parse the range header: https://datatracker.ietf.org/doc/html/rfc7233#section-2.1
+    const [startStr, endStr] = range.replace(/bytes=/, '').split('-');
```
dgozman, line 110: There could be multiple ranges in the header, separated by commas. We should either  support it, or error out.

packages/playwright-core/src/utils/httpServer.ts
```diff
@@ -80,22 +80,69 @@ export class HttpServer {
...
+    let start = parseInt(startStr, 10);
```
dgozman, line 111: `parseInt` is happy to parse "10abc". Use `+startStr` and check for NaN instead.

packages/playwright-core/src/utils/httpServer.ts
```diff
@@ -80,22 +80,69 @@ export class HttpServer {
...
+
+    // Sending Partial Content: https://datatracker.ietf.org/doc/html/rfc7233#section-4.1
+    response.writeHead(206, {
+      'Content-Range': `bytes ${start}-${end}/${size}`,
+      'Accept-Ranges': 'bytes',
+      'Content-Length': end - start + 1,
+      'Content-Type': mime.getType(path.extname(absoluteFilePath))!,
+    });
+
+    const readable = fs.createReadStream(absoluteFilePath, { start, end });
```
dgozman, line 142: Let's comment that both "range header and createReadStream accept inclusive start/end".

packages/playwright-core/src/utils/httpServer.ts
```diff
@@ -80,22 +80,77 @@ export class HttpServer {
...
+    response.setHeader('Content-Type', contentType);
+    response.setHeader('Content-Length', content.byteLength);
+    response.end(content);
+  }
+
+  _serveRangeFile(request: http.IncomingMessage, response: http.ServerResponse, absoluteFilePath: string) {
+    const range = request.headers.range;
+    if (!range || !range.startsWith('bytes=') || range.includes(', ')) {
+      response.statusCode = 400;
+      return response.end('Bad request');
```
dgozman, line 110: Shouldn't this be `response.statusMessage = 'Bad request'`?

packages/playwright-core/src/utils/httpServer.ts
dgozman: There could also be multiple dashes, like '1-5-10'. Let's error out in this case as well.

packages/playwright-core/src/utils/httpServer.ts
mxschmitt: `statusMessage` is set automatically, having it in the response body is also common practise afaik.

packages/playwright-core/src/utils/httpServer.ts
```diff
@@ -80,22 +80,77 @@ export class HttpServer {
...
+    response.statusCode = 200;
+    const contentType = mime.getType(path.extname(absoluteFilePath)) || 'application/octet-stream';
+    response.setHeader('Content-Type', contentType);
+    response.setHeader('Content-Length', content.byteLength);
+    response.end(content);
+  }
+
+  _serveRangeFile(request: http.IncomingMessage, response: http.ServerResponse, absoluteFilePath: string) {
+    const range = request.headers.range;
+    if (!range || !range.startsWith('bytes=') || range.includes(', ') || [...range].filter(char => char === '-').length !== 1) {
```
dgozman, line 108: nit: I'd do something like `match = range.match(/^bytes=(\d*)-(\d*)$/)` and use `+match[1], +match[2]`, but this also works.

dgozman (approved):

PR #10445, opened 2021-11-19
chore: drop wrapApiCall (2)

pavelfeldman:


packages/playwright-core/src/client/tracing.ts
```diff
@@ -46,29 +46,25 @@ export class Tracing implements api.Tracing {
...
-      await channel.tracingStartChunk(options);
-    });
+    await this._context._channel.tracingStartChunk(options);
   }
 
   async stopChunk(options: { path?: string } = {}) {
-    await this._context._wrapApiCall(async (channel: channels.BrowserContextChannel) => {
-      await this._doStopChunk(channel, options.path);
-    });
+    await this._doStopChunk(this._context._channel, options.path);
```
dgozman, line 61: This one still needs wrapApiCall because doStopChunk makes multiple calls.

dgozman (approved):

PR #10447, opened 2021-11-19
fix(test runner): dont mangle test names with multiple dashes

### LINKED ISSUE #10371: Don't mangle dashes `-` in snapshots [BUG]
**Context:**
- Playwright Version: 1.16.3

**Code Snippet**
```
expect(await page.screenshot()).toMatchSnapshot('multiple---dashes');
```

**Describe the bug**
If the snapshot name or directory contains multiple dashes, they should not be changed.

This issue was introduced here: https://github.com/microsoft/playwright/pull/8093/files#diff-7fb51a655c0375df53db9955e390089ccbed6850cec52be73ba8f1ecf3795f15R220

Should be a simple fix to remove dash `-` (`0x2D`) from the list of characters that need replacing, and maybe adding a comment explaining why it's removed, and maybe a regression test too.

JoelEinbinder:
Fixes #10371﻿

### DIFF
--- a/packages/playwright-test/src/util.ts
+++ b/packages/playwright-test/src/util.ts
@@ -141,7 +141,7 @@ export function expectType(receiver: any, type: string, matcherName: string) {
 }
 
 export function sanitizeForFilePath(s: string) {
-  return s.replace(/[\x00-\x2F\x3A-\x40\x5B-\x60\x7B-\x7F]+/g, '-');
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

pavelfeldman:
While you are here: I can see people using long descriptive texts for their tests. There is no way we need all that text in the file name. Can we do something like windows did in the good old days? I.e. instead of 

/show-click-above-position-stable-whenn-runing-under-certain-conditions/

we would get

/show-click-above-position~64cafe/

PR #10449, opened 2021-11-19
docs(api): redo request API for java

yury-s:


docs/src/api/class-requestoptions.md
```diff
@@ -0,0 +1,123 @@
+# class: RequestOptions
+* langs: java
+
+The [RequestOptions] allows to create form data to be sent via [APIRequestContext].
```
dgozman, line 4: This sentence seems wrong.

docs/src/api/class-requestoptions.md
```diff
@@ -0,0 +1,123 @@
...
+this request body. If this parameter is specified `content-type` header will be set to `multipart/form-data`
+unless explicitly provided.
+
+### param: RequestOptions.setMultipart.form
+- `form` <[FormData]>
+
+Form data to be serialized as html form using `multipart/form-data` encoding and sent as
+this request body.
+
+## method: RequestOptions.setQueryParam
```
dgozman, line 83: addQueryParam?

docs/src/api/class-formdata.md
```diff
@@ -0,0 +1,37 @@
...
+    .set("age", 30);
+page.request().post("http://localhost/submit", RequestOptions.create().setForm(form));
+```
+
+## method: FormData.create
+- returns: <[FormData]>
+
+Creates new instance of [FormData].
+
+## method: FormData.set
```
dgozman, line 21: Can you do the same name multiple times? If so, I'd call this `add`.

docs/src/api/params.md
```diff
@@ -303,28 +303,39 @@ Emulates consistent window screen size available inside web page via `window.scr
 
 Target URL.
 
-## fetch-option-params
+## js-python-fetch-option-params
```
dgozman, line 306: I'd rename this and similar back to `fetch-option-param` since it's the default style.

dgozman (approved):

docs/src/api/class-requestoptions.md
yury-s: Currently it behaves as `set` we can add `add` method later if there is demand.

docs/src/api/class-formdata.md
yury-s: Same here, we can add `append` if need be.

docs/src/api/params.md
yury-s: They are now available only in js and python and for such cases we include languages into the prefix in other places, so this is consistent with existing approach.

PR #10454, opened 2021-11-20
fix(codegen): allow to set initial window position

### LINKED ISSUE #5696: [BUG] Playwright Inspector window is opened outside of visible screen

peterblazejewicz:
This commit removes hardcoded position of the recorder (codegen) window
position, which made this entire core feature non usable on machines
withuot external additional display or when using HDPi or resolution different from the original environment on which that position was added as fixed, that could work in one, specific setup scenario.

If required, window position can be set as
`PWTEST_RECORDER_WINDOW_POSITION` environment variable, for example to
restore original, fixed location:

```powershell
$env:PWTEST_RECORDER_WINDOW_POSITION="1280,10"
npx playwright codegen wikipedia.org
```

Thanks!

Fixes #5696

### DIFF
--- a/src/server/supplements/recorder/recorderApp.ts
+++ b/src/server/supplements/recorder/recorderApp.ts
@@ -97,11 +97,12 @@ export class RecorderApp extends EventEmitter {
     const recorderPlaywright = require('../../playwright').createPlaywright(true) as import('../../playwright').Playwright;
     const args = [
       '--app=data:text/html,',
-      '--window-size=600,600',
-      '--window-position=1280,10',
+      '--window-size=600,600'
     ];
     if (process.env.PWTEST_RECORDER_PORT)
       args.push(`--remote-debugging-port=${process.env.PWTEST_RECORDER_PORT}`);
+    if (process.env.PWTEST_RECORDER_WINDOW_POSITION)
+      args.push(`--window-position=${process.env.PWTEST_RECORDER_WINDOW_POSITION}`);
     let channel: types.BrowserChannel | undefined;
     let executablePath: string | undefined;
     if (inspectedContext._browser.options.isChromium) {


src/server/supplements/recorder/recorderApp.ts
```diff
@@ -97,11 +97,12 @@ export class RecorderApp extends EventEmitter {
...
     const args = [
       '--app=data:text/html,',
-      '--window-size=600,600',
-      '--window-position=1280,10',
+      '--window-size=600,600'
     ];
     if (process.env.PWTEST_RECORDER_PORT)
       args.push(`--remote-debugging-port=${process.env.PWTEST_RECORDER_PORT}`);
+    if (process.env.PWTEST_RECORDER_WINDOW_POSITION)
+      args.push(`--window-position=${process.env.PWTEST_RECORDER_WINDOW_POSITION}`);
```
pavelfeldman, line 105: Let's || it with the previous value.

pavelfeldman: I.e. 

```js
args.push(`--window-position=${process.env.PWTEST_RECORDER_WINDOW_POSITION || '1280,10'}`);
```

peterblazejewicz: this will change the intent. I'd like to propose to make it work everywhere, and optionally add a way to position the window on different location. That is, the intent is to have this working ok on HDPi (or HD) resolutions laptops without external display attached. Those numbers numbers used for positioning are not the best to use on HDPi screens, especially with different scaling set on system due to high density of the screen (4k or *.k screens).
I understand the codegen is displayed in same location as browser, but at least is controllable using system features (Win+arrow keys on Windows)

peterblazejewicz: @pavelfeldman let me know what you folks think about this, I can see the better approach would be via config property,
so it works when run from e.g. VSCode

PR #10465, opened 2021-11-22
docs: python assertion typings fixes

mxschmitt:


docs/src/api/class-locatorassertions.md
```diff
@@ -495,6 +507,12 @@ assertThat(page.locator("list > .component")).hasText(new String[] {"Text 1", "T
 ```
 
 ### param: LocatorAssertions.toHaveText.expected
+- `expected` <[string]|[RegExp]|[Array]<[string]|[RegExp]>>
```
yury-s, line 510: Don't you need to specify language filter in both overrides?

yury-s (approved):

docs/src/api/class-locatorassertions.md
mxschmitt: added.

PR #10486, opened 2021-11-23
fix(trace-viewer): multiple iframe and UX fixes

pavelfeldman:


dgozman (approved):

packages/playwright-core/src/web/traceViewer/ui/snapshotTab.tsx
```diff
@@ -124,3 +121,5 @@ function renderTitle(snapshotTitle: string): string {
     return 'Action';
   return snapshotTitle;
 }
+
+const kDefaultViewport = { width: 1280, height: 720 };
```
dgozman, line 125: This looks like a regression - we ignore default context viewport.

dgozman (approved):

PR #10494, opened 2021-11-23
Add sanity check for frames before removing it.

shirshak55:
Sometimes frame doesn't exist for various reasons. We add
some sanity check before removing the frame recursively.

### DIFF
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -319,8 +319,11 @@ export class FrameManager {
   }
 
   removeChildFramesRecursively(frame: Frame) {
-    for (const child of frame.childFrames())
-      this._removeFramesRecursively(child);
+    if(frame){
+      for (const child of frame.childFrames()){
+        this._removeFramesRecursively(child);
+      }
+    }
   }
 
   private _removeFramesRecursively(frame: Frame) {
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -319,10 +319,9 @@ export class FrameManager {
   }
 
   removeChildFramesRecursively(frame: Frame) {
-    if(frame){
-      for (const child of frame.childFrames()){
+    if (frame){
+      for (const child of frame.childFrames())
         this._removeFramesRecursively(child);
-      }
     }
   }
 


packages/playwright-core/src/server/frames.ts
```diff
@@ -319,8 +319,10 @@ export class FrameManager {
   }
 
   removeChildFramesRecursively(frame: Frame) {
-    for (const child of frame.childFrames())
-      this._removeFramesRecursively(child);
+    if (frame){
```
pavelfeldman, line 322: You are masking a problem we are hoping to fix some day. Do you have a repro case?

shirshak55: well, we already do similar checks in many places. eg. When we are detaching frames etc. 

And to answer your question without prevaricating I don't have a repro case but I am seeing an error that crashes everything because of this minor problem. It's a loss-loss situation but I think it is better if we don't crash the production server.  And the problem is try-catch from the caller doesn't work that's why I still think this PR will somehow mitigate the issue.

pavelfeldman: It'd be really nice to get a repro. Otherwise, something is broken and we don't know about it. With your change we will never know about it.

shirshak55: I can reproduce it but as it is related to company project I can't handover code or additional details. 
Browser: firefox
```
  pw:protocol ◀ RECV {"method":"Runtime.executionContextCreated","params":{"executionContextId":"id-93","auxData":{"frameId":"frame-4294967312","name":""}},"sessionId":"18c94280-280f-4350-96a6-af2f888616d9"} +0ms
  pw:protocol ◀ RECV {"method":"Runtime.executionContextCreated","params":{"executionContextId":"id-94","auxData":{"frameId":"frame-4294967312","name":"__playwright_utility_world__"}},"sessionId":"18c94280-280f-4350-96a6-af2f888616d9"} +1ms
  pw:protocol ◀ RECV {"method":"Page.navigationCommitted","params":{"frameId":"frame-4294967312","navigationId":"232173047120117","url":"https://tpc.googlesyndication.com/sodar/sodar2/224/runner.html","name":""},"sessionId":"18c94280-280f-4350-96a6-af2f888616d9"} +0ms


/home/quantum/Desktop/projects/world/hello/node_modules/playwright-core/lib/server/frames.js:339
    for (const child of frame.childFrames()) this._removeFramesRecursively(child);
                              ^
TypeError: Cannot read properties of undefined (reading 'childFrames')
    at FrameManager.removeChildFramesRecursively (/home/quantum/Desktop/projects/world/hello/node_modules/playwright-core/lib/server/frames.js:339:31)
    at FrameManager.frameCommittedNewDocumentNavigation (/home/quantum/Desktop/projects/world/hello/node_modules/playwright-core/lib/server/frames.js:171:10)
    at FFPage._onNavigationCommitted (/home/quantum/Desktop/projects/world/hello/node_modules/playwright-core/lib/server/firefox/ffPage.js:212:30)
    at FFSession.emit (node:events:390:28)
    at /home/quantum/Desktop/projects/world/hello/node_modules/playwright-core/lib/server/firefox/ffConnection.js:247:41
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
  pw:browser [pid=107985] <kill> +15s
  pw:browser [pid=107985] <will force kill> +0ms
  pw:browser [pid=107985] starting temporary directories cleanup +1ms
  pw:browser [pid=107985] finished temporary directories cleanup +15ms
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

mxschmitt:
Closing as part of the triage process and since without a repro we can't add such fixes to our code-base.

PR #10520, opened 2021-11-24
chore: add --dry-run to install-deps CLI command

### LINKED ISSUE #10232: [Feature] Dry run mode for install-deps
When running Playwright tests in a containerized environment, it is helpful to build a custom image with native dependencies preinstalled. It is possible to do this by manually examining the list of dependencies in the nativeDeps.ts file: https://github.com/microsoft/playwright/blob/d34f9972abf3e412c9a73ee7b524e529a89136f8/packages/playwright-core/src/utils/nativeDeps.ts#L228-L242

However, it would be convenient if there were a way to list all the dependencies that playwright would install, without actually running the install command. It looks like we can do this by separating the dependency list generation and installation commands, for example, create a function that returns uniqueLibraries and a separate function that accepts that as input, then spawns apt-get: https://github.com/microsoft/playwright/blob/d34f9972abf3e412c9a73ee7b524e529a89136f8/packages/playwright-core/src/utils/dependencies.ts#L49-L72

I'd be happy to create a pull request if this is a feature you'd consider! I'd propose something like `npx playwright install-deps --dry-run` to return the package list for the current system

mxschmitt:
This will print the command which gets executed before we combine it with && and add sudo/su. Seems to be the expected behaviour here instead of the sh -c -u root stuff.

Fixes https://github.com/microsoft/playwright/issues/10232
Re-landing partly: https://github.com/microsoft/playwright/pull/6901
Relates: https://github.com/playwright-community/heroku-playwright-buildpack/issues/14

### DIFF
--- a/packages/playwright-core/src/cli/innerCli.ts
+++ b/packages/playwright-core/src/cli/innerCli.ts
@@ -107,7 +107,7 @@ program
         if (!args.length) {
           const executables = registry.defaultExecutables();
           if (options.withDeps)
-            await registry.installDeps(executables);
+            await registry.installDeps(executables, false);
           await registry.install(executables);
         } else {
           const installDockerImage = args.some(arg => arg === 'docker-image');
@@ -123,7 +123,7 @@ program
 
           const executables = checkBrowsersToInstall(args);
           if (options.withDeps)
-            await registry.installDeps(executables);
+            await registry.installDeps(executables, false);
           await registry.install(executables);
         }
       } catch (e) {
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/cli/cli.ts
```diff
@@ -104,12 +104,13 @@ program
     .command('install [browser...]')
     .description('ensure browsers necessary for this version of Playwright are installed')
     .option('--with-deps', 'install system dependencies for browsers')
-    .action(async function(args: string[], options: { withDeps?: boolean }) {
+    .option('--dry-run', 'Print the command which would get executed when installing the system dependencies')
```
dgozman, line 107: I'd rather drop it from here.

packages/playwright-core/src/cli/cli.ts
```diff
@@ -144,13 +145,14 @@ Examples:
 
 program
     .command('install-deps [browser...]')
+    .option('--dry-run', 'Print the command which would get executed')
```
dgozman, line 148: "Do not execute installation commands, only print them"

dgozman (commented):
Ideally, we should do something similar for windows. At least, dry run must be a no-op.

> before we combine it with && and add sudo/su

I'd say that's lying to the user - let's print the whole thing :)

dgozman (approved):

PR #10522, opened 2021-11-24
feat(cli): add yarn to buildPlaywrightCLICommand

mrienstra:
Modify messages like this to say `yarn` instead of `npx` when applicable:
Screen Shot 2021-11-24 at 2 48 28 PM

`determinePackageManager` was copied from [packages/create-playwright/src/utils.ts](https://github.com/microsoft/playwright/blob/master/packages/create-playwright/src/utils.ts#L58-L64).

Prior to running across `determinePackageManager`, I was planning to use something a bit more conservative (in the sense of "more likely to default to npm"):
```js
if (!fs.existsSync(path.join(rootDir, 'package-lock.json')) && fs.existsSync(path.join(rootDir, 'yarn.lock')))
```

### DIFF
--- a/packages/playwright-core/src/utils/registry.ts
+++ b/packages/playwright-core/src/utils/registry.ts
@@ -21,7 +21,7 @@ import * as util from 'util';
 import * as fs from 'fs';
 import lockfile from 'proper-lockfile';
 import { getUbuntuVersion } from './ubuntuVersion';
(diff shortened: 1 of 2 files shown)


pavelfeldman:
We are consistently using npm / npx throughout the documentation. It is easy for our users to understand what they need to do when they use yarn.

PR #10532, opened 2021-11-25
docs(python): add assertion examples

mxschmitt:


docs/src/intro-java.md
```diff
@@ -59,7 +59,7 @@ public class Example {
     <dependency>
       <groupId>com.microsoft.playwright</groupId>
       <artifactId>playwright</artifactId>
-      <version>1.16.0</version>
+      <version>1.17.0</version>
```
pavelfeldman, line 62: Looks unrelated

mxschmitt: it is a drive-by.

pavelfeldman: Yes, but won't we be merging this into docs for 1.17?

mxschmitt: I don't think so that we will merge ToT docs with 1.17 docs. It would not work, since the ToT docs contain already too many unrelated changes.

pavelfeldman: I was not suggesting to merge ToT docs into 1.17 docs. I was suggesting to keep this separate to remember that we need to merge it into 1.17.

mxschmitt: extracted into https://github.com/microsoft/playwright/pull/10609

pavelfeldman (approved):

PR #10538, opened 2021-11-25
fix(test-runner): webServer port detection on Node.js 17

### LINKED ISSUE #10346: [BUG] WebServer does not work on Node.js 17+
**Context:**
- Playwright Version: ToT
- Operating System: Mac
- Node.js version: 17+

**Describe the bug**

If the user is using the Webserver in Node.js 17 it does not detect its readiness accordingly and ends up in an infinite timeout.

Repro: https://twitter.com/CristianEnguidn/status/1460509470728065028

1. Install pnpm
2. pnpm install
3. Migrate it to webServer
4. npm run e2e

mxschmitt:
Node.js >= 17 does return the ipv6 address (`::1`) when resolving localhost. Before it did return ipv4 as expected (`127.0.0.1`). With this change we verify now ipv4 and ipv6 localhost availability.

Fixes #10346

### DIFF
--- a/packages/playwright-test/src/webServer.ts
+++ b/packages/playwright-test/src/webServer.ts
@@ -105,9 +105,9 @@ export class WebServer {
 }
 
 async function isPortUsed(port: number): Promise<boolean> {
-  return new Promise<boolean>(resolve => {
+  const innerIsPortUsed = (host: string) => new Promise<boolean>(resolve => {
     const conn = net
-        .connect(port)
+        .connect(port, host)
         .on('error', () => {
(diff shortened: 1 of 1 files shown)


pavelfeldman (changes requested):
But the ports are still the same, why two checks?

mxschmitt:
> But the ports are still the same, why two checks?

one for the ipv4 and one for the ipv6 network interface.

pavelfeldman:
But those are different. Which one does web server start? Shouldn't we check the one that is going to be used?

PR #10542, opened 2021-11-25
chore: validate client/server version match for remote connections

mxschmitt:
- Keep the validation on the server side? Makes sense I guess
- Enforce it? I would probably, so the reported compatibility issues get resolved automatically. Maybe we can add an environment variable to opt-out of it. e.g. `Set the PLAYWRIGHT_DO_NOT_EXIT_ON_INCOMPATIBILITY to skip the validation`
- Tested with connect/and PlaywrightClient/PlaywrightServer

Related work: The language bindings need to implement it.

### DIFF
--- a/packages/playwright-core/src/dispatchers/dispatcher.ts
+++ b/packages/playwright-core/src/dispatchers/dispatcher.ts
@@ -18,7 +18,7 @@ import { EventEmitter } from 'events';
 import * as channels from '../protocol/channels';
 import { serializeError } from '../protocol/serializers';
 import { createScheme, Validator, ValidationError } from '../protocol/validator';
-import { assert, debugAssert, isUnderTest, monotonicTime } from '../utils/utils';
+import { assert, debugAssert, getPlaywrightVersion, isUnderTest, monotonicTime } from '../utils/utils';
 import { tOptional } from '../protocol/validatorPrimitives';
 import { kBrowserOrContextClosedError } from '../utils/errors';
 import { CallMetadata, SdkObject } from '../server/instrumentation';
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/dispatchers/dispatcher.ts
```diff
@@ -132,6 +132,8 @@ export class Root extends Dispatcher<{ guid: '' }, any> {
   async initialize(params: channels.RootInitializeParams): Promise<channels.RootInitializeResult> {
     assert(this.createPlaywright);
     assert(!this._initialized);
+    const expectedPlaywrightVersion = getPlaywrightVersion();
+    assert(params.version === expectedPlaywrightVersion, `Server (${expectedPlaywrightVersion}) does not match with client version (${params.version}).`);
```
pavelfeldman, line 136: ... does not match the client version ..

pavelfeldman: We should also match major/minor only and forgive patch.

packages/playwright-core/src/dispatchers/dispatcher.ts
```diff
@@ -320,3 +321,10 @@ function formatLogRecording(log: string[]): string {
 }
 
 let lastEventId = 0;
+
+function assertPlaywrightVersion(givenVersion: string): void {
+  const expectedVersion = getPlaywrightVersion();
+  const givenParts = givenVersion.split('.');
+  const expectedParts = givenVersion.split('.');
+  assert(givenParts.slice(0, 2) === expectedParts.slice(0, 2), `${expectedVersion} does not match the client version ${givenVersion}.`);
```
pavelfeldman, line 329: That's not how === works.

packages/playwright-core/src/dispatchers/dispatcher.ts
```diff
@@ -320,3 +321,10 @@ function formatLogRecording(log: string[]): string {
 }
 
 let lastEventId = 0;
+
+function assertPlaywrightVersion(givenVersion: string): void {
+  const expectedVersion = getPlaywrightVersion();
+  const givenParts = givenVersion.split('.');
+  const expectedParts = givenVersion.split('.');
+  assert(givenParts.slice(0, 2).join('') === expectedParts.slice(0, 2).join(''), `${expectedVersion} does not match the client version ${givenVersion}.`);
```
pavelfeldman, line 329: 1.17 == 11.7

pavelfeldman (approved):

Issue #8981, opened 2021-09-17
[Question] [Playwright Test] Run new browser instance for each test with --worker. 


karanshah-browserstack:
When running multiple tests with -workers, we can see that the number of browsers spawned is the same as the number of workers specified. Is there any way we can spawn a new browser for each test instead of new pages?

dgozman:
@karanshah-browserstack Tell us more about your usecase please. Why do you want a separate browser? Running multiple browsers is less efficient and usually does not give you any benefits.

karanshah-browserstack:
@dgozman We are trying to run this in remote mode(i.e client running on one machine and the server is running on a different machine) and users can spawn multiple browser connections in parallel. If different tests are using the same browser connection, the users find it unintuitive to locate a specific test within a bunch of tests that have run in the same browser. 

It would ideal if there was an option to specify(an optional parameter) that every playwright test runs in a different browser session.

karanshah-browserstack:
@dgozman Workers are recreated when a test fails to keep pristine environment, instead of doing this just on failures, we need this to happen for each and every spec. Currently we aren't able to do this as workers are reused. Can you please help if this is possible through a workaround or we need a config similar to "reuseWorker" as provided?

pavelfeldman:
#### Why was this issue closed?

We are prioritizing the features based on the upvotes, recency and activity in the issue thread. It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it. We are closing issues like this one to keep our bug database maintainable. Please feel free to open a new issue and link this one to it if you think this is a mistake.

PR #10573, opened 2021-11-29
Adding no reuseWorkers config option

francisf:
Changes to support creating a new worker per spec on cloud providers. Keeping this configurable - default is current behavior of reusing workers.


karanshah-browserstack:
fixes https://github.com/microsoft/playwright/issues/8981

pavelfeldman:
This looks premature, let's keep discussing it in the bug. So far the request does not make sense to us, so there probably is some API misuse.

Issue #10575, opened 2021-11-29
[Question] wait for console debug event using Firefox browser


cunhabruno:
Using the below code, I am waiting for a console message in **debug level** to be displayed in the browser, when I run with Chrome it works fine, but when I run with firefox, the console messages that are on debug level are not captured, is there any specific configuration I need to do, to be able to listen for debug level console messages using Firefox browser?

``` js
        try {
            await this.page.waitForEvent('console', {
                predicate: async (consoleMessage: ConsoleMessage) => {
                    return consoleMessage.text() === message
                },
                timeout: timeout
            })
        } catch (e) {
            throw new Error(`error waiting for message "${message}" Cause: ${e}`)
        }
```

mxschmitt:
Could you provide a full example including the code where you call console.debug? I've tested a few cases with Firefox and all are working fine for me.

cunhabruno:
Hello @mxschmitt thanks for the quick reply!

Actually, the `console.debug ` is called in the application side, not on my Playwright tests. On the Playwright side I am just using the code I provided in the previous message, to wait for the debug message. I have executed the tests using `PWDEBUG=1` and I could see the message in Firefox devtools, but Playwright wasn’t able to find the message.

mxschmitt:
My guess is that the `console.debug()` call happened already before calling waitForEvent. So you can do either:

````ts
const waitForYourSpecificConsoleEvent = page.waitForEvent('console');
// .... your logic which triggers the console event
await waitForYourSpecificConsoleEvent
```

or something like:

```ts
await Promise.all([
  page.waitForEvent('console'),
  page.click(".."), // replace with your logic which triggers the event
])
```

(Without a full repro its hard to debug)

cunhabruno:
I have tried using both approaches, I will try to come up with some better example to repro it and get back here once I have it :)

pavelfeldman:
We need more information to act on this report. Please file a new one and link to this issue when you get back to it!

PR #10577, opened 2021-11-29
test: add console.info/debug tests

mxschmitt:
tests #10575

### DIFF
--- a/tests/page/page-event-console.spec.ts
+++ b/tests/page/page-event-console.spec.ts
@@ -68,8 +68,12 @@ it('should work for different console API calls', async ({ page }) => {
     console.debug('calling console.debug');
     console.log(Promise.resolve('should not wait until resolved!'));
   });
+  // WebKit uses console.debug() to report binding calls, make sure they don't get reported.
+  await page.exposeBinding('foobar', async (_, value) => page.evaluate(value => console.log(value), value));
+  await page.evaluate(() => window['foobar']('Using bindings'));
+
   expect(messages.map(msg => msg.type())).toEqual([
-    'timeEnd', 'trace', 'dir', 'warning', 'error', 'info', 'debug', 'log'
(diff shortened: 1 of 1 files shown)


pavelfeldman (approved):

tests/page/page-event-console.spec.ts
```diff
@@ -64,17 +64,21 @@ it('should work for different console API calls', async ({ page }) => {
     console.dir('calling console.dir');
     console.warn('calling console.warn');
     console.error('calling console.error');
+    console.info('calling console.info');
+    console.debug('calling console.debug');
```
pavelfeldman, line 68: Actually, we can't use debug in WebKit - we use ut for bindings.

pavelfeldman (changes requested):
Clearing approve per comment above

tests/page/page-event-console.spec.ts
mxschmitt: Seems like we can? We emit custom information via console.debug but filter it out before we emit it to the client side so it seems fine?

https://github.com/microsoft/playwright/blob/f19864890fa6d4ecdadb2bba7def2f5bcde6bec5/packages/playwright-core/src/server/webkit/wkPage.ts#L495-L503

pavelfeldman: Ah, you are right. Their inspector is busted, but we don't report our staff. Good to land then.

pavelfeldman (approved):

tests/page/page-event-console.spec.ts
pavelfeldman: Actually, let's add a binding to that test, otherwise, you land this test, I remove that branch from the code and test keeps passing.

pavelfeldman (changes requested):
As per above

pavelfeldman (approved):

Issue #10565, opened 2021-11-28
[BUG] Firefox emitting console event for unhandled errors


PaperStrike:
**Context:**
- Playwright Version: 1.16.3, 1.18.0-alpha-nov-28-2021
- Operating System: Windows 11, Ubuntu 21.04
- Node.js version: 14.18.0 on Windows, 16.13.0 on Linux
- Browser: Firefox 93.0, Firefox 94.0.1
- Extra: Tested both versions on both Windows and Ubuntu (WSL)

**Code Snippet**

```javascript
const playwright = require('playwright');

const highlight = (text) => `\x1b[7m${text}\x1b[0m`;
const red = (text) => `\x1b[41m${text}\x1b[0m`;

['chromium', 'firefox', 'webkit'].reduce(async (last, browserName) => {
  await last;
  console.log(`=== test start ${browserName} ===`);

  const browser = await playwright[browserName].launch();
  const page = await browser.newPage();
  page.on('console', (message) => {
    console.log(red('console event'), message.text());
  });
  page.on('pageerror', (error) => {
    console.log(highlight('pageerror event'), error);
  });

  await page.addScriptTag({
    content: 'throw new Error();',
  });
  await browser.close();

  console.log(`=== test end ${browserName} ===`);
}, Promise.resolve());
```

**Describe the bug**

Running the snippet produces

```log
=== test start chromium ===
pageerror event Error
    at <anonymous>:1:7
    at addScriptContent (eval at evaluate (:3:2389), <anonymous>:9:21)
    at t.default.evaluate (<anonymous>:3:2412)
    at t.default.<anonymous> (<anonymous>:1:44)
=== test end chromium ===
=== test start firefox ===
pageerror event Error:
    at  (about:blank line 9 > injectedScript:1:7)
    at addScriptContent (debugger eval code line 3 > eval:9:21)
    at evaluate (debugger eval code:3:2412)
    at  (debugger eval code:1:44)
    at evaluateFunction (chrome://juggler/content/content/Runtime.js:381:60)
    at callFunction (chrome://juggler/content/content/Runtime.js:103:41)
    at _onMessage (chrome://juggler/content/SimpleChannel.js:165:37)
    at receiveMessage (chrome://juggler/content/SimpleChannel.js:16:42)
console event [JavaScript Error: "Error: " {file: "about:blank line 9 > injectedScript" line: 1}]
@about:blank line 9 > injectedScript:1:7
[...]

PR #10581, opened 2021-11-29
browser(firefox): ignore console events with attached exception

mxschmitt:
#10565

([ref usage](https://ffsearch.azurewebsites.net/#path=%2Fhome%2Fjoe%2Fplaywright%2Fbrowser_patches%2Ffirefox%2Fcheckout%2Fremote%2Fcdp%2Fdomains%2Fcontent%2FLog.jsm&line=75))

### DIFF
--- a/browser_patches/firefox-beta/BUILD_NUMBER
+++ b/browser_patches/firefox-beta/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1305
-Changed: aslushnikov@gmail.com Thu Nov 11 15:24:13 HST 2021
+1306
+Changed: max@schmitt.mx Mon 29 Nov 2021 13:10:11 CET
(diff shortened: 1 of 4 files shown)


pavelfeldman (commented):
Do we only see exceptions attached to thrown messages in Firefox? (This is not the case with Chrome, random messages cab have associated exception).

mxschmitt:
> Do we only see exceptions attached to thrown messages in Firefox? (This is not the case with Chrome, random messages cab have associated exception).

Yes, from my understanding see [here](https://ffsearch.azurewebsites.net/#path=%2Fhome%2Fjoe%2Fplaywright%2Fbrowser_patches%2Ffirefox%2Fcheckout%2Fdom%2Fbindings%2FnsIScriptError.idl&line=102) from the IDL:

> The hasException attribute is used to differentiate between no thrown exception and `throw undefined`.

pavelfeldman (approved):

PR #10589, opened 2021-11-29
osx monterey changes

rsouthpaw:
Adding mac-os-12


JoelEinbinder:
Does this fix something? Everything is working fine for me on Monterey

rsouthpaw:
> Does this fix something? Everything is working fine for me on Monterey

It works but the code is considering Monterey as BigSur during the setup.
Last stable version set is 11 (20/BigSur - 9). The darwin version getting returned for Monterey is 21, code picks minimum of 11 & 12, and follows the process for BigSur for monterey. While this has no side effects but adding it explicitly for organizing the code.

Yurii397:
> [JoelEinbinder](/JoelEinbinder)

doesn't work for me on Monterey now. Left couple of comments in this issue here - https://github.com/microsoft/playwright/issues/10594#issuecomment-982616105

and still blocked now

aslushnikov:
> I believe this code is organized around what versions we have explicit builds for rather than whatever the latest to be released was. 

That's right; things just work on Monterey. No need for these changes.

PR #10590, opened 2021-11-29
fix(expect): include selector errors in the log

### LINKED ISSUE #10471: [Feature] Better error messages/print selector
When I have a line in my test that fails when something doesn't go right with a page selector, I'd like to know exactly what selector failed. If my selector is composed using a template string, I want to know what the final string was, not the name of the variable I interpolated.

For example with this error message:

One question I have is, was something wrong with the `duplicateName` variable I interpolated ? If I could have the information about which selector failed above the failed line printout, then I could determine that much more quickly.

For example if my variable was `duplicateName: 'My Duplicate'` then I could receive the error message:

```console
Expected locator "text=My Duplicate" to be visible.
```

yury-s:
Fixes #10471

### DIFF
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -1279,6 +1279,8 @@ export class Frame extends SdkObject {
       // A: We want user to receive a friendly message containing the last intermediate result.
       if (js.isJavaScriptErrorInEvaluate(e))
         throw e;
+      if (!isSessionClosedError(e))
+        metadata.log.push('' + e);
       return { received: controller.lastIntermediateResult(), matches: options.isNot, log: metadata.log };
     });
   }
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/frames.ts
```diff
@@ -1279,6 +1279,8 @@ export class Frame extends SdkObject {
       // A: We want user to receive a friendly message containing the last intermediate result.
       if (js.isJavaScriptErrorInEvaluate(e))
         throw e;
+      if (!isSessionClosedError(e))
```
pavelfeldman, line 1282: This looks suspicious, what's the exact scenario?

yury-s: Selector with syntax error or non-existent engine.

Issue #10189, opened 2021-11-09
[Feature] expect(response).toBeOk() to print http response status and body


pavelfeldman:
### Discussed in https://github.com/microsoft/playwright/discussions/10175

<div type='discussions-op-text'>

<sup>Originally posted by **AndyGBC** November  9, 2021</sup>
Hi all, below is a code snippet that will use the APIContextRequest to make a post request and create a company.  This works fine, though I am sure someone could help improve this, if so, feel free :)

My question is, if this fails and returns a 403 status, the expect on the status will obviously catch it and fail, but how can I go about getting more information about why it failed?

By this I mean, if after the POST request has been made, I can console.log the response.json() which would show me some details about the failure but that won't appear in the HTML report which is what I would like to see if possible.

Any help appreciated.

`async createCompany(name: string, shortName: string, description: string): Promise<string> {

        expect(name.trim().length != 0).toBeTruthy();
        
        const response = await this.baseAPI.context.post(`https://some.endpoint/create`, { 
            headers:{
                'X-BC-JWT': this.baseAPI.api.getToken()
            },
            data: {
                kind: 'company',
                name: name,
                short_name: shortName,
                descr: description,
            }
        });

        expect(response.status()).toEqual(200);
        expect(await response.json()).toHaveProperty('artifact');

        let companyUri: string = '';

        await response.json().then(data => {
            companyUri = data.artifact;
        });

        expect(companyUri.length != 0).toBeTruthy();
        return companyUri;
    }`
</div>

yury-s:
@AndyGBC when the server returns 403 do you want to see the response body in the report or status text would be enough?

AndyGBC:
@yury-s Ideally the response body if possible as that would give us the fastest feedback on the failure.

Thank you :)

PR #10596, opened 2021-11-30
feat(expext): toBeOK for APIResponse

yury-s:
#10189


packages/playwright-test/src/matchers/matchers.ts
```diff
@@ -263,3 +268,15 @@ export function toHaveURL(
...
   }, expected, options);
 }
+
+export async function toBeOK(
+  this: ReturnType<Expect['getState']>,
+  response: APIResponseEx
+) {
+  const matcherName = 'toBeOK';
+  expectType(response, 'APIResponse', matcherName);
+  const log = await response._fetchLog();
```
dgozman, line 278: I think we should skip fetching the log when `this.isNot !== response.ok()`.

packages/playwright-test/types/testExpect.d.ts
```diff
@@ -103,6 +103,11 @@ declare global {
        */
       toBeHidden(options?: { timeout?: number }): Promise<R>;
 
+      /**
+       * Asserts given APIResponse's status is 200.
```
dgozman, line 107: between 200 and 299

dgozman (approved):

tests/playwright-test/playwright.expect.true.spec.ts
```diff
@@ -322,3 +322,41 @@ test('should print syntax error', async ({ runInlineTest }) => {
...
 });
+
+test('should support toBeOK', async ({ runInlineTest, server }) => {
+  const result = await runInlineTest({
+    'a.test.ts': `
+      const { test } = pwt;
+
+      test('pass with response', async ({ page }) => {
+        const res = await page.request.get('${server.EMPTY_PAGE}');
+        expect(res).toBeOK();
```
mxschmitt, line 333: You need await there afaik. Otherwise it ends up in an unhanded promise rejection.

mxschmitt (approved):
Should we add it to test-assertions-js.md and create a new class-responseassertions.md for it?a

tests/playwright-test/playwright.expect.true.spec.ts
yury-s: Yeah, good point.

yury-s:
Yes, I'll add it to the docs.

PR #10558, opened 2021-11-26
feat(screenshots): Add `captureBeyondViewport` option to screenshot

trotzig:
The default behavior for Playwright in chromium is to temporarily resize
the viewport before taking a screenshot. This happens both for
`page.screenshot({ fullPage: true })` and `elementHandle.screenshot()`.
Without it, screenshots would be cut at the viewport size. In general,
this behavior/workaround is okay. But when you start using CSS viewport
units like vh and vw, screenshots will look wrong. Consider the
following example:

<main style="min-height: 2000px">
  <div style="height: 50vh"></div>
</main>

For a viewport of size 800px, the div will render at 400px. But a
Playwright screenshot of a this page will show the div at 1000px. Of
course, one could argue that you should probably use vh heights together
with a fixed max-height or vmax or similar (with the reasoning usually
being something like "people shouldn't assume viewports are always
small") but the fact is that there are tons of websites out there that
use viewport units and they all sort of look wrong in Playwright
screenshots.

At happo.io, we've been using Puppeteer for an extended period of time,
and I've noticed that they don't have the same problem. Viewport units
work well even when content doesn't fit inside the viewport. After doing
some digging, I found that Puppeteer passes the `captureBeyondViewport`
flag in the call to `captureScreenshot()`.
https://github.com/puppeteer/puppeteer/blob/4d9dc8c0e613f22d4cdf237e8bd0b0da3c588edb/src/common/Page.ts#L2778
https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot

The `captureBeyondViewport` option is experiemental, so it's probably
best not to lock ourselves into the behavior by making it the default.
That's why this commit will add a new option to the screenshot methods
(page and elementHandle). That way, people can opt in to the
experimental behavior and people who are happy with the viewport resize
workaround can stay with the default.

This is a chromium only option. Other browsers won't be affected by this
change (they are also already doing "the right thing" for screenshots
with viewport units).

Issue #757, opened 2020-01-30
[BUG] Chromium fullPage screenshot mistreats vh and vw units


KartoffelToby:
I tested a very basic code with playwright. to take a fullpage screenshot of a webpage with a css typed header thats using `height: 65vh;`

The problem here with pupeeter is that the viewport haven't any relation to the vh. so the Screenshot has the wrong result. Playwright with chroimuim have the same problem.

But playwright with Firefox Headless Browser take the right result.

So i think the core bug is in Chroimuim.

Are there any Solutions that all 3 Browsers can handle CSS vh and vw with the right viewport?

My testing code is:

```
const playwright = require('playwright');

(async () => {
  for (const browserType of ['chromium', 'firefox']) {
    const browser = await playwright[browserType].launch({
      args: [
        '--disable-gpu',
        '--no-sandbox',
        '--disable-web-security',
        '--disable-dev-profile'
      ]
    });

    const  context = await browser.newContext({
      viewport: {
        width: 1920,
        height: 1080
      }
    });
    
    const page = await context.newPage('https://elements.t-haber.de/');
    await page.screenshot({ path: `example-${browserType}.png` , fullPage: true });
    await browser.close();
  }
})();
```

duytran-agilityio:
I saw the same issue. I see the results are different between Firefox and Chromium. Firefox has a better results.

Prior99:
I have the same issue with an even more basic site. Chromium just doesn't screenshot the whole page.

Lemaro86:
I have the same issue

mxschmitt:
Should be fixed in https://github.com/microsoft/playwright/pull/10606

PR #10606, opened 2021-11-30
chore(chromium): Capture off-screen content without resizing viewport

trotzig:
The default behavior for Playwright in chromium is to temporarily resize
the viewport before taking a screenshot. This happens both for
page.screenshot({ fullPage: true }) and elementHandle.screenshot().
Without it, screenshots would be cut at the viewport size. In general,
this behavior/workaround is okay. But when you start using CSS viewport
units like vh and vw, screenshots will look wrong. Consider the
following example:

For a viewport of size 800px, the div will render at 400px. But a
Playwright screenshot of a this page will show the div at 1000px. Of
course, one could argue that you should probably use vh heights together
with a fixed max-height or vmax or similar (with the reasoning usually
being something like "people shouldn't assume viewports are always
small") but the fact is that there are tons of websites out there that
use viewport units and they all sort of look wrong in Playwright
screenshots.

At happo.io, we've been using Puppeteer for an extended period of time,
and I've noticed that they don't have the same problem. Viewport units
work well even when content doesn't fit inside the viewport. After doing
some digging, I found that Puppeteer passes the captureBeyondViewport
flag in the call to captureScreenshot().
https://github.com/puppeteer/puppeteer/blob/4d9dc8c0e613f22d4cdf237e8bd0b0da3c588edb/src/common/Page.ts#L2778
https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot

In this commit, I'm making use of `captureBeyondViewport` so that we no
longer have to resize the viewport. I first worked on making this
opt-in, but after having things reviewed by @pavelfeldman we decided
it's safe to rely on this flag whenever content doesn't fit inside the
viewport.

This is a chromium only option. Other browsers won't be affected by this
change (they are also already doing "the right thing" for screenshots
with viewport units).

### DIFF
--- a/packages/playwright-core/src/server/screenshotter.ts
+++ b/packages/playwright-core/src/server/screenshotter.ts
@@ -62,79 +62,50 @@ export class Screenshotter {
   async screenshotPage(progress: Progress, options: types.ScreenshotOptions): Promise<Buffer> {
     const format = validateScreenshotOptions(options);
     return this._queue.postTask(async () => {
-      const { viewportSize, originalViewportSize } = await this._originalViewportSize(progress);
+      const { viewportSize } = await this._originalViewportSize(progress);
 
       if (options.fullPage) {
         const fullPageSize = await this._fullPageSize(progress);
         let documentRect = { x: 0, y: 0, width: fullPageSize.width, height: fullPageSize.height };
-        let overriddenViewportSize: types.Size | null = null;
         const fitsViewport = fullPageSize.width <= viewportSize.width && fullPageSize.height <= viewportSize.height;
-        if (!this._page._delegate.canScreenshotOutsideViewport() && !fitsViewport) {
-          overriddenViewportSize = fullPageSize;
-          progress.throwIfAborted(); // Avoid side effects.
-          await this._page.setViewportSize(overriddenViewportSize);
-          progress.cleanupWhenAborted(() => this._restoreViewport(originalViewportSize));
-        }
         if (options.clip)
           documentRect = trimClipToSize(options.clip, documentRect);
-        const buffer = await this._screenshot(progress, format, documentRect, undefined, options);
+        const buffer = await this._screenshot(progress, format, documentRect, undefined, fitsViewport, options);
         progress.throwIfAborted(); // Avoid restoring after failure - should be done by cleanup.
-        if (overriddenViewportSize)
-          await this._restoreViewport(originalViewportSize);
         return buffer;
       }
 
       const viewportRect = options.clip ? trimClipToSize(options.clip, viewportSize) : { x: 0, y: 0, ...viewportSize };
-      return await this._screenshot(progress, format, undefined, viewportRect, options);
+      return await this._screenshot(progress, format, undefined, viewportRect, true, options);
     });
   }
 
   async screenshotElement(progress: Progress, handle: dom.ElementHandle, options: types.ElementScreenshotOptions = {}): Promise<Buffer> {
     const format = validateScreenshotOptions(options);
     return this._queue.postTask(async () => {
-      const { viewportSize, originalViewportSize } = await this._originalViewportSize(progress);
+      const { viewportSize } = await this._originalViewportSize(progress);
 
       await handle._waitAndScrollIntoViewIfNeeded(progress);
 
       progress.throwIfAborted(); // Do not do extra work.
-      let boundingBox = await handle.boundingBox();
+      const boundingBox = await handle.boundingBox();
       assert(boundingBox, 'Node is either not visible or not an HTMLElement');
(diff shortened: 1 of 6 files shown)


trotzig:
This PR replaces https://github.com/microsoft/playwright/pull/10558 and is a fix for https://github.com/microsoft/playwright/issues/757.

packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -62,79 +62,50 @@ export class Screenshotter {
   async screenshotPage(progress: Progress, options: types.ScreenshotOptions): Promise<Buffer> {
     const format = validateScreenshotOptions(options);
     return this._queue.postTask(async () => {
-      const { viewportSize, originalViewportSize } = await this._originalViewportSize(progress);
+      const { viewportSize } = await this._originalViewportSize(progress);
 
+      const fullPageSize = await this._fullPageSize(progress);
```
pavelfeldman, line 67: Let's keep existing logic and only call _fullPageSize under the options.fullPage branch.

packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -62,79 +62,50 @@ export class Screenshotter {
...
+        const buffer = await this._screenshot(progress, format, documentRect, undefined, fitsViewport, options);
         progress.throwIfAborted(); // Avoid restoring after failure - should be done by cleanup.
-        if (overriddenViewportSize)
-          await this._restoreViewport(originalViewportSize);
         return buffer;
       }
 
       const viewportRect = options.clip ? trimClipToSize(options.clip, viewportSize) : { x: 0, y: 0, ...viewportSize };
-      return await this._screenshot(progress, format, undefined, viewportRect, options);
+      return await this._screenshot(progress, format, undefined, viewportRect, fitsViewport, options);
```
pavelfeldman, line 79: `fitsViewport` should be always `true` here, no matter what full page size is. We are screenshotting the viewport, so we trim to clip and to viewport size in the previous line.

tests/screenshot.spec.ts
```diff
@@ -97,9 +96,9 @@ browserTest.describe('page screenshot', () => {
       document.documentElement.style.margin = '0';
       document.documentElement.style.padding = '0';
       const div = document.createElement('div');
-      div.style.width = '1250px';
-      div.style.height = '8440px';
-      div.style.background = 'linear-gradient(red, blue)';
+      div.style.width = '100%';
```
pavelfeldman, line 99: Let's keep existing tests as they are, they should pass as is.

tests/screenshot.spec.ts
```diff
@@ -183,46 +211,7 @@ browserTest.describe('element screenshot', () => {
     await context.close();
   });
 
-  browserTest('should restore default viewport after fullPage screenshot', async ({ browser }) => {
```
pavelfeldman, line 186: Same here, `captureBeyondViewport` still resizes the viewport under the hood, so we should make sure things are normal after capture.

tests/screenshot.spec.ts
```diff
@@ -183,46 +211,7 @@ browserTest.describe('element screenshot', () => {
...
-    const context = await browser.newContext({ viewport: { width: 456, height: 789 } });
-    const page = await context.newPage();
-    await verifyViewport(page, 456, 789);
-    const screenshot = await page.screenshot({ fullPage: true });
-    expect(screenshot).toBeInstanceOf(Buffer);
-    await verifyViewport(page, 456, 789);
-    await context.close();
-  });
-
-  browserTest('should restore viewport after page screenshot and exception', async ({ browser, server, mode }) => {
```
pavelfeldman, line 196: Same here and below, old tests should stay.

tests/screenshot.spec.ts
```diff
@@ -135,6 +134,35 @@ browserTest.describe('element screenshot', () => {
     await context.close();
   });
 
+  browserTest('element screenshot should include content outside viewport', async ({ browserName, headless, platform, contextFactory }) => {
```
pavelfeldman, line 137: I liked previous naming that included vh more, also was nice to have two separate tests.

tests/screenshot.spec.ts
trotzig: Ah, I didn't know -- thought these would become redundant. I'll resurrect them.

tests/screenshot.spec.ts
trotzig: Will do.

trotzig:
@pavelfeldman I think I've addressed your feedback now. Thanks for helping me out!

pavelfeldman (approved):
Looks awesome!

PR #10608, opened 2021-11-30
chore: escape CSS selectors in codegen correctly

### LINKED ISSUE #10561: [BUG] codegen seems to trip over elements with namespace prefixes
**Context:**
- Playwright Version: 1.17.0-rc1
- Operating System: Windows
- Node.js version: v17.0.1
- Browser: Chromium
- Extra: Frontend with mixture of AngularJS (Legacy) and React

**Describe the bug**
When using Playwright codegen, I can log in successfully on our application, and that's recorded properly, but as soon as the main frontend is loaded, a lot of the elements generate errors in the console when hovering over them. Also such elements are not marked/highlighted, and when clicking, the wrong element is recorded, and the wrong element receives the event (usually resulting in nothing happening in the frontend).

I think it's related to elements with namespace prefixes. Examples of such elements are ng:switch, cr:desktop-view, twf:app-search-container, ui:icon, etc.

The error in the JavaScript console is: Uncaught DOMException: Failed to execute 'querySelectorAll' on 'Document': 'twf:app-search-container' is not a valid selector.

Unfortunately I don't have an easy way to describe a reproduction scenario since the application (and more importantly, the credentials) is private.
If needed, I might be able to arrange demo credentials though.

playwright-rc-1-17-codegen-console-errors

mxschmitt:
Fixes #10561

It seems like there was no reason that `toLocaleLowerCase` instead of `toLowerCase` was used.

### DIFF
--- a/packages/playwright-core/src/server/injected/selectorGenerator.ts
+++ b/packages/playwright-core/src/server/injected/selectorGenerator.ts
@@ -155,26 +155,26 @@ function buildCandidates(injectedScript: InjectedScript, element: Element): Sele
   if (element.hasAttribute('aria-label'))
     candidates.push({ engine: 'css', selector: `[aria-label=${quoteString(element.getAttribute('aria-label')!)}]`, score: 10 });
   if (element.getAttribute('alt') && ['APPLET', 'AREA', 'IMG', 'INPUT'].includes(element.nodeName))
-    candidates.push({ engine: 'css', selector: `${element.nodeName.toLowerCase()}[alt=${quoteString(element.getAttribute('alt')!)}]`, score: 10 });
+    candidates.push({ engine: 'css', selector: `${CSS.escape(element.nodeName.toLowerCase())}[alt=${quoteString(element.getAttribute('alt')!)}]`, score: 10 });
 
   if (element.hasAttribute('role'))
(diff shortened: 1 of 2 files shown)


tests/selector-generator.spec.ts
```diff
@@ -279,6 +279,6 @@ it.describe('selector generator', () => {
 
   it('should work with tricky ids', async ({ page }) => {
     await page.setContent(`<button id="this:is-my-tricky.id"><span></span></button>`);
-    expect(await generate(page, 'button')).toBe('[id="this:is-my-tricky.id"]');
+    expect(await generate(page, 'button')).toBe('[id="this\\:is-my-tricky\\.id"]');
```
pavelfeldman, line 282: Add a bunch of new tests instead.

packages/playwright-core/src/server/injected/selectorGenerator.ts
```diff
@@ -282,7 +282,7 @@ function escapeForRegex(text: string): string {
 }
 
 function quoteString(text: string): string {
```
pavelfeldman, line 284: This method is now misleading, inline `"CSS(...)"`

pavelfeldman (approved):

PR #10670, opened 2021-12-02
feat(test-runner): specific playwright types for expect

JoelEinbinder:
Fixes #10016

~~I am not sure what's different between our two approaches.~~

~~Let's write tests based on the differences.~~

### DIFF
--- a/packages/playwright-test/types/testExpect.d.ts
+++ b/packages/playwright-test/types/testExpect.d.ts
@@ -15,11 +15,20 @@
  */
 
 import type * as expect from 'expect';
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):

dgozman (commented):
Lint complains, so I guess something's not right?

JoelEinbinder:
@mxschmitt The test you added implies that
```ts
expect(page).toBe(otherPage)
```
will return an error. But it seems totally fine to me.

JoelEinbinder:
Discussed offline. Everyone seems cool with landing this as is.

PR #10697, opened 2021-12-03
test(mouse): add a test for pointerdown event with custom button

Andarist:
I was comparing my adjustments to [`cypress-real-events`](https://github.com/dmtrKovalenko/cypress-real-events/) with the code in Playwright. My code was not working correctly so I've gone here to check if the same thing works with Playwright - I've written this test and verified that it works in Playwright. This has allowed me to finally find a silly, silly mistake in my code.

I've noticed that there was no `button: "someType"` test in this file and since I've already added that as part of my own investigation I've figured out that it's best to just submit it as a PR.

### DIFF
--- a/tests/page/page-mouse.spec.ts
+++ b/tests/page/page-mouse.spec.ts
@@ -78,6 +78,38 @@ it('should dblclick the div', async ({ page, server }) => {
   expect(event.button).toBe(0);
 });
 
+it('should pointerdown the div with a custom button', async ({ page, server, browserName }) => {
+  await page.setContent(`<div style='width: 100px; height: 100px;'>Click me</div>`);
+  await page.evaluate(() => {
+    window['pointerdownPromise'] = new Promise(resolve => {
+      document.querySelector('div').addEventListener('pointerdown', event => {
+        resolve({
+          type: event.type,
+          detail: event.detail,
+          clientX: event.clientX,
+          clientY: event.clientY,
+          isTrusted: event.isTrusted,
+          button: event.button,
+          buttons: event.buttons,
+          pointerId: event.pointerId,
+        });
+      });
+    });
+  });
+  await page.mouse.click(50, 60, {
+    button: 'middle'
+  });
+  const event = await page.evaluate(() => window['pointerdownPromise']);
+  expect(event.type).toBe('pointerdown');
+  expect(event.detail).toBe(browserName === 'webkit' ? 1 : 0);
+  expect(event.clientX).toBe(50);
+  expect(event.clientY).toBe(60);
+  expect(event.isTrusted).toBe(true);
+  expect(event.button).toBe(1);
+  expect(event.buttons).toBe(4);
+  expect(event.pointerId).toBe(browserName === 'firefox' ? 0 : 1);
+});
+
 it('should select the text with mouse', async ({ page, server }) => {
   await page.goto(server.PREFIX + '/input/textarea.html');
   await page.focus('textarea');
--- a/packages/playwright-core/src/server/webkit/wkInput.ts
+++ b/packages/playwright-core/src/server/webkit/wkInput.ts
@@ -40,9 +40,9 @@ function toButtonsMask(buttons: Set<types.MouseButton>): number {
   let mask = 0;
   if (buttons.has('left'))
     mask |= 1;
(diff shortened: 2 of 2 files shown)


pavelfeldman:
Thanks for the patch! It appears that `pointerId` is 0 in Firefox, while `detail` is 1 in Safari. You can add a `browserName` fixture to your test and check those two in a browser-specific manner. Then the tests will pass and it'll reflect what those browsers do in reality.

Andarist:
> You can add a browserName fixture to your test and check those two in a browser-specific manner. 

Thank you for the suggestion, I've applied it based on other tests in the file - I hope this will now pass CI.

> It appears that pointerId is 0 in Firefox,

Out of curiosity, I've checked this in [the spec](https://w3c.github.io/pointerevents/#dom-pointerevent-pointerid) and it seems that they didn't get it wrong. Both 0 and 1 may be used for the primary mouse pointer:
"User agents MAY reserve a generic pointerId value of 0 or 1 for the primary mouse pointer"

> while detail is 1 in Safari. 

According to [the spec](https://w3c.github.io/pointerevents/#attributes-and-default-actions) they are doing it wrong cause it explicitly mentions this:
"For all pointer events in the table above the detail [UIEVENTS] attribute SHOULD be 0."

tests/page/page-mouse.spec.ts
```diff
@@ -78,6 +78,38 @@ it('should dblclick the div', async ({ page, server }) => {
...
+    button: 'middle'
+  });
+  const event = await page.evaluate(() => window['pointerdownPromise']);
+  expect(event.type).toBe('pointerdown');
+  expect(event.detail).toBe(browserName === 'webkit' ? 1 : 0);
+  expect(event.clientX).toBe(50);
+  expect(event.clientY).toBe(60);
+  expect(event.isTrusted).toBe(true);
+  expect(event.button).toBe(1);
+  expect(event.buttons).toBe(4);
```
Andarist, line 109: It seems that we get 2 in webkit here. I think this is because the buttons mask map middle button to 2 here:
https://github.com/microsoft/playwright/blob/ea9475d37d885abca7e604ba42642a2975f50f7f/packages/playwright-core/src/server/webkit/wkInput.ts#L43-L44

This is different from the mapping used for firefox:
https://github.com/microsoft/playwright/blob/ea9475d37d885abca7e604ba42642a2975f50f7f/packages/playwright-core/src/server/firefox/ffInput.ts#L52-L53
and from the mapping that I've implemented in my other PR here:
https://github.com/microsoft/playwright/pull/10698/files#diff-0d5748af7b9e00da36d86779f7daf1d0fb86b749910be5ab759f4ebaad232538R112-R113

It also doesn't match the expected mapping described in the MDN:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons#value

I've checked in Safari and it seems the right click maps correctly, as expected, to 2:
https://jsfiddle.net/8Lqyvs4m/1/

I can't currently verify the middle button cause I don't have access to the physical mouse right now but if right click maps to 2 then it would make sense that the middle button would map to 4.

So... I think this is a bug in the existing webkit mapping. Should I fix it as part of this PR?

pavelfeldman: Let's see what happens on the other platforms. The numbers that we send to different browsers are different, nothing wrong with it - they all have their own enums. But you might be right with this one.

pavelfeldman: > So... I think this is a bug in the existing webkit mapping. Should I fix it as part of this PR?

Yup, looks consistent on all ports, so safe to fix as a part of this PR!

pavelfeldman:
@Andarist do you want to take a stab at it?

Andarist:
@pavelfeldman sure, I've fixed the masking function and pushed it out - let's see if all existing tests pass now on CI (I'm just before hitting the bed and I didn't have time to run the full test suite locally). I will check the status in the morning and fix whatever might need to be fixed (if anything).

Andarist:
@pavelfeldman the CI is ✅ do you want me to add any other tests or is the introduced test case sufficient for this change?

pavelfeldman:
Awesome! Thanks!

pavelfeldman (approved):

PR #10698, opened 2021-12-03
feat(mouse): set `.buttons` correctly for basic mouse commands in Chrome

Andarist:
While working on https://github.com/microsoft/playwright/pull/10697 I've noticed that this is not handled correctly in `crInput` because the `buttons` argument was completely ignored. I've figured out that I will just add this since the change is trivial and might spare somebody doing an investigation why this doesn't work in the future 😉

### DIFF
--- a/tests/page/page-mouse.spec.ts
+++ b/tests/page/page-mouse.spec.ts
@@ -110,6 +110,44 @@ it('should pointerdown the div with a custom button', async ({ page, server, bro
   expect(event.pointerId).toBe(browserName === 'firefox' ? 0 : 1);
 });
 
+it('should mousedown the div with extra buttons', async ({ page, server }) => {
+  await page.setContent(
+      `<div style='width: 100px; height: 100px;'>Click me</div>`
+  );
+  await page.evaluate(() => {
+    window['mousedownPromise'] = new Promise(resolve => {
+      document.querySelector('div').addEventListener('mousedown', event => {
+        if (event.button !== 1 /* wheel button */) return;
+
+        resolve({
+          type: event.type,
+          detail: event.detail,
+          clientX: event.clientX,
+          clientY: event.clientY,
+          isTrusted: event.isTrusted,
+          button: event.button,
+          buttons: event.buttons,
+        });
+      });
+    });
+  });
+  await page.mouse.move(50, 60);
+  await page.mouse.down({
+    button: 'right',
+  });
+  await page.mouse.down({
+    button: 'middle',
+  });
+  const event = await page.evaluate(() => window['mousedownPromise']);
+  expect(event.type).toBe('mousedown');
+  expect(event.detail).toBe(1);
+  expect(event.clientX).toBe(50);
+  expect(event.clientY).toBe(60);
+  expect(event.isTrusted).toBe(true);
+  expect(event.button).toBe(1); // wheel button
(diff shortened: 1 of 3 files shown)


tests/page/page-mouse.spec.ts
```diff
@@ -78,6 +78,43 @@ it('should dblclick the div', async ({ page, server }) => {
...
+          buttons: event.buttons,
+        });
+      });
+    });
+  });
+  await page.mouse.move(50, 60);
+  await page.mouse.down({
+    button: 'right'
+  });
+  await page.mouse.down({
```
pavelfeldman, line 105: i'm trying to simulate this behavior in the real browser and am failing - I don't see the second event being dispatched, how real is this scenario?

Andarist: I’ve successfully reproduced this with a mouse:
- press right mouse button without releasing it -> context menu gets open
- press middle mouse button -> this closes the context menu
- press middle mouse button again -> this fires the event that is being tested here

The scenario might not be super realistic or useful (but possible!). It’s just a one that i’ve been sort-of debugging since at first I’ve configured buttons property in a wrong way for this command (when working with Cypress and `cypress-real-events`). So I’ve just reused this exact situation for the test case here, as I was sure how it works. However, I think the same idea applies for mixing other mouse buttons together. So maybe this could be converted to a „simpler” test - with a more common mouse button combination.

tests/page/page-mouse.spec.ts
```diff
@@ -78,6 +78,43 @@ it('should dblclick the div', async ({ page, server }) => {
   expect(event.button).toBe(0);
 });
 
+it('should mousedown the div with extra buttons', async ({ page, server }) => {
+  await page.setContent(`<div style='width: 100px; height: 100px;'>Click me</div>`);
+  await page.evaluate(() => {
+    window['mousedownPromise'] = new Promise(resolve => {
+      document.querySelector('div').addEventListener('mousedown', event => {
+        if (event.button !== 1)
```
mxschmitt, line 86:
```suggestion
        if (event.button !== 1 /* wheel button */)
```

tests/page/page-mouse.spec.ts
```diff
@@ -78,6 +78,43 @@ it('should dblclick the div', async ({ page, server }) => {
...
+    button: 'middle'
+  });
+  const event = await page.evaluate(() => window['mousedownPromise']);
+  expect(event.type).toBe('mousedown');
+  expect(event.detail).toBe(1);
+  expect(event.clientX).toBe(50);
+  expect(event.clientY).toBe(60);
+  expect(event.isTrusted).toBe(true);
+  expect(event.button).toBe(1);
+  expect(event.buttons).toBe(6);
```
mxschmitt, line 115:
```suggestion
  expect(event.button).toBe(1); // wheel button
  expect(event.buttons).toBe(6); // right + middle button
```

mxschmitt (approved):

Andarist:
@mxschmitt I've resolved the conflict and applied your suggestions

aslushnikov:
@Andarist I took the liberty to re-write the test to our liking and merge this PR. Hope you don't mind!

P.S. Thank you for the fix!!!

Andarist:
All good 👍

PR #10703, opened 2021-12-03
test(test-runner): add test for hiding conditional expect matchers

mxschmitt:
Added a test for #10016 and #10670

### DIFF
--- a/tests/playwright-test/expect.spec.ts
+++ b/tests/playwright-test/expect.spec.ts
@@ -50,6 +50,23 @@ test('should be able to call expect.extend in config', async ({ runInlineTest })
   expect(result.passed).toBe(1);
 });
 
+test('should not expand huge arrays', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'expect-test.spec.ts': `
+      const { test } = pwt;
+      test('numeric ranges', () => {
+        const a1 = Array(100000).fill(1);
+        const a2 = Array(100000).fill(1);
+        a2[500] = 2;
+        test.expect(a1).toEqual(a2);
+      });
+    `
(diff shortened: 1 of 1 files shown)


tests/playwright-test/expect.spec.ts
```diff
@@ -50,6 +50,23 @@ test('should be able to call expect.extend in config', async ({ runInlineTest })
   expect(result.passed).toBe(1);
 });
 
+test('should not expand huge arrays', async ({ runInlineTest }) => {
```
mxschmitt, line 53: drive-by: moved above the `runTSC` tests so its grouped.

tests/playwright-test/expect.spec.ts
```diff
@@ -126,19 +143,29 @@ test('should work with custom PlaywrightTest namespace', async ({ runTSC }) => {
...
+    'a.spec.ts': `
+    const { test } = pwt;
+    test('custom matchers', async ({ page }) => {
+      await test.expect(page).toHaveURL('https://example.com');
+      await test.expect(page).toBe(true);
+      // @ts-expect-error
+      await test.expect(page).toBeEnabled();
+
+      await test.expect(page.locator('foo')).toBeEnabled();
+      await test.expect(page.locator('foo')).toBe(true);
```
dgozman, line 158: I'd think that this one is also an error.

dgozman (approved):

tests/playwright-test/expect.spec.ts
mxschmitt: discussed offline: lets hide the default expect matchers for now. -> applied suggestion.

PR #10710, opened 2021-12-04
chore: fix utils/generate_dotnet_channels.js after monorepo

mxschmitt:


utils/generate_dotnet_channels.js
```diff
@@ -126,7 +126,7 @@ function objectType(props, indent, onlyOptional = false) {
   return { ts: `${indent}{${inner.ts}\n${indent}}`, scheme: `tObject({\n${inner.scheme}\n${indent}})` };
 }
 
-const yml = fs.readFileSync(path.join(__dirname, '..', 'src', 'protocol', 'protocol.yml'), 'utf-8');
+const yml = fs.readFileSync(path.join(__dirname, '../packages/playwright-core/src/protocol/protocol.yml'), 'utf-8');
```
dgozman, line 129: Keep it as before to work on Windows?

utils/generate_dotnet_channels.js
```diff
@@ -139,6 +139,9 @@ for (const [name, value] of Object.entries(protocol)) {
     mixins.set(name, value);
 }
 
+if (!process.argv[2])
+  throw new Error('.NET repository needs to be specified as an argument' + process.argv[2]);
```
dgozman, line 143: Let's provide an example, because it should point to some `src/Playwright` IIRC?

utils/generate_dotnet_channels.js
mxschmitt: afaik Node.js can deal with that. But let me revert it back to the old syntax.

dgozman (approved):

PR #10711, opened 2021-12-04
chore(dotnet): fix channel generation for 1.17

mxschmitt:


utils/generate_dotnet_channels.js
```diff
@@ -43,6 +42,10 @@ function mapType(type) {
     return 'int';
   if (type === 'ResourceTiming')
     return 'RequestTimingResult';
+  if (type === 'LifecycleEvent')
```
mxschmitt, line 45: these got renamed upstream (in TypeScript) so we need to map them to the previous names to not break users.

utils/generate_dotnet_channels.js
```diff
@@ -227,4 +230,10 @@ function processCustomType(type, optional) {
     && inlineType(type.properties.name).ts === 'string') {
     return { ts: 'DeviceDescriptorEntry', scheme: 'tObject()', optional };
   }
-}
\ No newline at end of file
+}
+
+function rewriteChannelType(type) {
+  if (['BrowserType'].includes(type))
```
mxschmitt, line 236: Before it generated it to `Core.Playwright` which works, with the current generator it leads to just `BrowserType` which conflicts with [this](https://github.com/microsoft/playwright-dotnet/blob/f1067ec2d20b4061ef0a60a96a2d144f254e3e12/src/Playwright/BrowserType.cs) BrowserType instead of [this](https://github.com/microsoft/playwright-dotnet/blob/8728aced9db702205ed175ad180ee66d603eec4b/src/Playwright/Core/BrowserType.cs).

So we need to rewrite it back to the more precise import.

utils/generate_dotnet_channels.js
dgozman: Let's add a big TOGO: "why do channels use public enums???"

utils/generate_dotnet_channels.js
dgozman: Let's do `Core.Foo` everywhere and remove using namespace.

utils/doclint/generateDotnetApi.js
```diff
@@ -122,7 +122,7 @@ function writeFile(kind, name, spec, body, folder, extendsName = null, namespace
   out.push(...body);
   out.push('}');
 
-  let content = template.replace('[NAMESPACE]', namespace).replace('[CONTENT]', out.join(EOL));
```
mxschmitt, line 125: drive-by: removed unused logic

dgozman (approved):

Issue #8345, opened 2021-08-20
test() can only be called in a test file


kdawg1981:
I'm needing some help with playwright. I started a new project and created one test to make sure that it was working, but I'm getting the following error message: 
`Error: tests\test.spec.js:3:1: test() can only be called in a test file
    at errorWithLocation (C:\Users\kevin.morales\source\repos\playwright_bridge\node_modules\@playwright\test\lib\test\util.js:220:10)
    at TestTypeImpl._createTest (C:\Users\kevin.morales\source\repos\playwright_bridge\node_modules\@playwright\test\lib\test\testType.js:76:51)
    at C:\Users\kevin.morales\source\repos\playwright_bridge\node_modules\@playwright\test\lib\test\transform.js:133:12
    at Object.<anonymous> (C:\Users\kevin.morales\source\repos\playwright_bridge\tests\test.spec.js:3:1)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Loader._requireOrImport (C:\Users\kevin.morales\Source\Repos\Playwright_Bridge\node_modules\@playwright\test\lib\test\loader.js:211:18)
    at Loader.loadTestFile (C:\Users\kevin.morales\Source\Repos\Playwright_Bridge\node_modules\@playwright\test\lib\test\loader.js:130:18)
    at Runner._run (C:\Users\kevin.morales\Source\Repos\Playwright_Bridge\node_modules\@playwright\test\lib\test\runner.js:210:59)`

I have done a uninstall and reinstall of playwright on my system through NPM and I'm not sure what is going on here. Any help would be greatly appreciated. 

Thank you, 

Kevin Morales

mxschmitt:
You could provide us your `tests\test.spec.js` or even better a small repro in form of a GithHub repository?

kdawg1981:
`const { test, expect } = require('@playwright/test');

test('basic test', async({ page }) => {
  await page.goto('https://playwright.dev/');
  const title = page.locator('.navbar__inner .navbar__title');
  await expect(title).toHaveText('Playwright');
});`

[...]

PR #10719, opened 2021-12-04
feat(proxy): unify local network proxy behavior

### LINKED ISSUE #10631: [BUG] HTTP requests to localhost not going through proxy with Chrome
**Context:**
- Playwright Version: 1.16.3
- Operating System: Mac and Linux
- Node.js version: v12.20.0
- Browser: Chromium
- Extra: Proxy configuration to test with OWASP

**Code Snippet**

I'm setting the proxy options in the `use` section of `playwright.config.js` like this:

```javascript
// playwright.config.js
import { PlaywrightTestConfig, devices } from '@playwright/test';

const proxyHost = process.env.LOCAL_PROXY_HOST || 'localhost';
const proxyPort = process.env.LOCAL_PROXY_PORT || '8080';

  testDir: 'tests',
  forbidOnly: !!process.env.CI,
  use: {
    baseURL: 'http://localhost:4200',
    viewport: { width: 1280, height: 1580 },
    ignoreHTTPSErrors: true,
    trace: 'on-first-retry',
    storageState: 'auth-state.json',
    proxy = {
      server: `http://${proxyHost}:${proxyPort}`
    }
  },
  webServer: {
    command: 'npx ng serve',
    port: 4200,
    timeout: 120 * 1000,
  },
  projects: [
    {
      name: 'chromium',
    }
  ],
};
export default config;
```

**Describe the bug**

When running playwright with a simple Angular application that is served from `localhost:4200`, requests to `localhost` are not going through the configured proxy (they are not logged in the proxy) but all other requests do. This problem only occurs with a Chromium browser, it works as expected with Firefox where I can see all requests logged in the proxy.

Is this a known behavior or are there any additional options we could set to enforce proxy for localhost? Browserstack has similar options (`forceProxy`).

rwoll:
When configuring a proxy, Chromium requires a magic tokens to get some
local network requests to go through the proxy. This has tripped up a
few users, so we make the behavior default to the expected: proxy
everything including the local requests. This matches the other vendors
as well.

NB: This can be disabled via
`PLAYWRIGHT_DISABLE_FORCED_CHROMIUM_PROXIED_LOOPBACK=1`

Supercedes: #8345
Fixes: #10631

### DIFF
--- a/tests/browsercontext-proxy.spec.ts
+++ b/tests/browsercontext-proxy.spec.ts
@@ -91,93 +91,55 @@ it('should use proxy', async ({ contextFactory, server, proxyServer }) => {
   await context.close();
 });
 
-it.describe('should proxy local network requests', () => {
+it.describe.parallel('should proxy local network requests', () => {
   for (const additionalBypass of [false, true]) {
-    it.describe(additionalBypass ? 'with other bypasses' : 'by default', () => {
-      it('localhost', async ({ platform, browserName, contextFactory, server, proxyServer }) => {
-        it.fail(platform !== 'linux' && browserName === 'webkit' && additionalBypass, 'WK fails to proxy localhost if additional bypasses are present');
+    it.describe.parallel(additionalBypass ? 'with other bypasses' : 'by default', () => {
+      for (const params of [
+        {
+          target: 'localhost',
+          description: 'localhost',
+        },
+        {
+          target: '127.0.0.1',
+          description: 'loopback address',
+        },
+        {
+          target: '169.254.3.4',
+          description: 'link-local'
+        }
+      ]) {
+        it(`${params.description}`, async ({ platform, browserName, contextFactory, server, proxyServer }) => {
+          it.fail(platform !== 'linux' && browserName === 'webkit' && additionalBypass && ['localhost', '127.0.0.1'].includes(params.target), 'WK fails to proxy 127.0.0.1 and localhost if additional bypasses are present');
 
-        const path = `/target-${additionalBypass}-localhost.html`;
-        server.setRoute(path, async (req, res) => {
-          res.end('<html><title>Served by the proxy</title></html>');
+          const path = `/target-${additionalBypass}-${params.target}.html`;
+          server.setRoute(path, async (req, res) => {
+            res.end('<html><title>Served by the proxy</title></html>');
+          });
+
+          const url = `http://${params.target}:${server.PORT}${path}`;
+          proxyServer.forwardTo(server.PORT);
+          const context = await contextFactory({
+            proxy: { server: `localhost:${proxyServer.PORT}`, bypass: additionalBypass ? '1.non.existent.domain.for.the.test' : undefined }
+          });
+
+          const page = await context.newPage();
+          await page.goto(url);
+          expect(proxyServer.requestUrls).toContain(url);
+          expect(await page.title()).toBe('Served by the proxy');
+
+          await page.goto('http://1.non.existent.domain.for.the.test/foo.html').catch(() => {});
+          if (additionalBypass)
+            expect(proxyServer.requestUrls).not.toContain('http://1.non.existent.domain.for.the.test/foo.html');
+          else
+            expect(proxyServer.requestUrls).toContain('http://1.non.existent.domain.for.the.test/foo.html');
+
+          await context.close();
         });
-
-        const url = `http://localhost:${server.PORT}${path}`;
-        proxyServer.forwardTo(server.PORT);
-        const context = await contextFactory({
-          proxy: { server: `localhost:${proxyServer.PORT}`, bypass: additionalBypass ? '1.non.existent.domain.for.the.test' : undefined }
-        });
-
-        const page = await context.newPage();
-        await page.goto(url);
-        expect(proxyServer.requestUrls).toContain(url);
-        expect(await page.title()).toBe('Served by the proxy');
-
-        await page.goto('http://1.non.existent.domain.for.the.test/foo.html').catch(() => {});
-        if (additionalBypass)
-          expect(proxyServer.requestUrls).not.toContain('http://1.non.existent.domain.for.the.test/foo.html');
-        else
-          expect(proxyServer.requestUrls).toContain('http://1.non.existent.domain.for.the.test/foo.html');
-
-        await context.close();
-      });
-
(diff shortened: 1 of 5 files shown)


rwoll:
Looks like the FF failures have to do with my test, and not PW impl. Looks like our `proxyServer` is too naïve for FF to startup properly (if used at the launch- but not context-level). The following blows up the `newPage` call in a funny way logging `Error: Parse Error: Invalid method encountered`. The only request I see hit the proxy before that is maybe an OCSP related one to http://ocsp.digicert.com/. Need to dig in later…

```
it.only('debug', async ({ browserType, server, proxyServer }) => {
  server.setRoute('/target.html', async (req, res) => {
    res.end('<html><title>Served by the proxy</title></html>');
  });
  proxyServer.forwardTo(server.PORT);
  const browser = await browserType.launch({
    proxy: { server: `localhost:${proxyServer.PORT}` }
  });
  await browser.close();
});
```

rwoll:
> Looks like the FF failures have to do with my test, and not PW impl. Looks like our `proxyServer` is too naïve for FF to startup properly (if used at the launch- but not context-level). The following blows up the `newPage` call in a funny way logging `Error: Parse Error: Invalid method encountered`. The only request I see hit the proxy before that is maybe an OCSP related one to http://ocsp.digicert.com/. Need to dig in later…
> 
> ```
> it.only('debug', async ({ browserType, server, proxyServer }) => {
>   server.setRoute('/target.html', async (req, res) => {
>     res.end('<html><title>Served by the proxy</title></html>');
>   });
>   proxyServer.forwardTo(server.PORT);
>   const browser = await browserType.launch({
>     proxy: { server: `localhost:${proxyServer.PORT}` }
>   });
>   const page = await browser.newPage();
>   await browser.close();
> });
> ```

Looks like the issue is with trying to proxy the HTTPS/connect requests in proxyServer which originate from browser startup.

rwoll:
> > Looks like the FF failures have to do with my test, and not PW impl. Looks like our `proxyServer` is too naïve for FF to startup properly (if used at the launch- but not context-level). The following blows up the `newPage` call in a funny way logging `Error: Parse Error: Invalid method encountered`. The only request I see hit the proxy before that is maybe an OCSP related one to http://ocsp.digicert.com/. Need to dig in later…
> > 
> > ```
> > it.only('debug', async ({ browserType, server, proxyServer }) => {
> >   server.setRoute('/target.html', async (req, res) => {
> >     res.end('<html><title>Served by the proxy</title></html>');
> >   });
> >   proxyServer.forwardTo(server.PORT);
> >   const browser = await browserType.launch({
> >     proxy: { server: `localhost:${proxyServer.PORT}` }
> >   });
> >   const page = await browser.newPage();
> >   await browser.close();
> > });
> > ```
> 
> Looks like the issue is with trying to proxy the HTTPS/connect requests in proxyServer which originate from browser startup.

Pushed a workaround in latest commit. (Skips rewriting some HTTPS requests FF requires for startup so we don't have to deal with MITMing them better just for the tests.)

tests/browsercontext-proxy.spec.ts
```diff
@@ -91,6 +91,56 @@ it('should use proxy', async ({ contextFactory, server, proxyServer }) => {
   await context.close();
 });
 
+it.describe.parallel('should proxy local network requests', () => {
```
pavelfeldman, line 94: nit: we are going to bulk-enable parallel for our suite, so no need to opt-in.

tests/browsercontext-proxy.spec.ts
```diff
@@ -91,6 +91,56 @@ it('should use proxy', async ({ contextFactory, server, proxyServer }) => {
...
+          target: '127.0.0.1',
+          description: 'loopback address',
+        },
+        {
+          target: '169.254.3.4',
+          description: 'link-local'
+        }
+      ]) {
+        it(`${params.description}`, async ({ platform, browserName, contextFactory, server, proxyServer }) => {
+          it.fail(platform !== 'linux' && browserName === 'webkit' && additionalBypass && ['localhost', '127.0.0.1'].includes(params.target), 'WK fails to proxy 127.0.0.1 and localhost if additional bypasses are present');
```
pavelfeldman, line 112: nit, style: given the complexity of the fail predicate, I'd copy-paste this test 3 times instead.

pavelfeldman (approved):

tests/browsercontext-proxy.spec.ts
rwoll: > nit, style: given the complexity of the fail predicate, I'd copy-paste this test 3 times instead.

@pavelfeldman how strong of a nit is this? Hehe 

I found the more complex fail predicate much easier to read/maintain/edit than trying to find the subtle differences in the otherwise lots of setup in each of the cases. 

Mind if I revert https://github.com/microsoft/playwright/commit/d307a872d3a5a432f82fb88cc71c63c50b03e01c before merging?

pavelfeldman: > how strong of a nit is this?

Not too strong, do it the way you like.

pavelfeldman: Happy to merge as is if this is ready.

rwoll: Thanks @pavelfeldman! Just reverted! Ready for a merge whenever assuming tests pass.

rwoll:
Fixed up the tests based on the CQ1 run. Ready for another run. (Ran all the proxy tests locally {macOS, Linux/Docker} x {headless, head full}. Don’t have a windows box to verifies this will pass now, but they should :)

rwoll:
@mxschmitt Thanks for toggling the label. Looks like `npm ci` exploded on a lot of the bots and didn’t even run the tests. Is that just runner flakiness, or should I rebase and try again?

mxschmitt:
> @mxschmitt Thanks for toggling the label. Looks like `npm ci` exploded on a lot of the bots and didn’t even run the tests. Is that just runner flakiness, or should I rebase and try again?

Seems like a temporary GitHub/NPM outage, I'd recommend to rebase.

rwoll:
Thanks for the re-trigger! Everything passed on the rebase except one unrelated test (https://github.com/microsoft/playwright/runs/4464958397?check_suite_focus=true)

So should be good to merge! 🎉 

cc: @pavelfeldman

PR #10730, opened 2021-12-06
chore(input): add missing `await` before `this.move` in the `click`

Andarist:
As far as I can tell this was missing by accident

### DIFF
--- a/packages/playwright-core/src/server/input.ts
+++ b/packages/playwright-core/src/server/input.ts
@@ -211,7 +211,7 @@ export class Mouse {
   async click(x: number, y: number, options: { delay?: number, button?: types.MouseButton, clickCount?: number } = {}) {
     const { delay = null, clickCount = 1 } = options;
     if (delay) {
-      this.move(x, y);
+      await this.move(x, y);
       for (let cc = 1; cc <= clickCount; ++cc) {
         await this.down({ ...options, clickCount: cc });
         await new Promise(f => setTimeout(f, delay));


pavelfeldman:
It is as designed. Is there a bug you are observing related to this code?

pavelfeldman:
We actually might be missing .catch(() => {}) there to mute potential unhandled rejections.

Andarist:
> Is there a bug you are observing related to this code?

No, I assume that there can't be a race condition and that commands are delivered/processed in order as the other branch doesn't await its "steps" but rather await all of the steps at once:
https://github.com/microsoft/playwright/blob/19ea0dfab05e733ff61a52c4133904a834f96830/packages/playwright-core/src/server/input.ts#L223-L229

It just felt weird to me that this other branch was actually awaiting `this.move` and the one that I've changed didn't. So even if I would adjust this PR to use your `.catch` suggestion the two main code branches in this `click` would handle this slightly differently. Would that be expected?

pavelfeldman:
As long as we are awaiting down/up after the move, it'll be awaited. Input domain dispatching is serialized.

Andarist:
> As long as we are awaiting down/up after the move, it'll be awaited. Input domain dispatching is serialized.

That sounds about right from what I can tell - it's just that usually, I expect async jobs to be explicitly awaited: either immediately or sometime later. This one felt looked like fired & forgotten and slightly inconsistent with the other branch in the same function. If you think that it's fine as it is - you can close this PR, I certainly won't mind it.

pavelfeldman:
If you look at the other branch, you'll see that we deliberately don't await for `move` before dispatching `down`/`up`. Those are added into array of promises, without awaiting. The idea is that we want to issue them immediately, in a serialized manner and then await for the whole bunch to resolve. The version with `delay` is trying to do the same thing, i.e. `move`, then `down` immediately, but it obviously needs to wait before the `up` in order to respect the delay. So awaiting `down` seems like not a big deal there. I don't think await before `move` hurts, but it diverges from the spirit of the move/down/up serialized pattern. Floating promise is a real problem though, so either `await` to `catch` should be there and I would be in favor of `catch`.

Andarist:
I haven't really meant that in the other branch you await that move immediately - but rather that it is awaited as part of that `promises` batch. So in a sense - it has to happen & succeed (at the protocol level at least).

I'm OK with changing the introduced `await` to `catch` but it feels to me like this would then act differently than the other branch. If that `move` can throw (for any reason) then the branch without the delay will also throw (as the error will propagate) but if we add a catch to the branch with the delay then "the same" error will just get silently swallowed.

This is the last time I'm raising concern about this though - I'm just a bystander and believe that you have far better context than me for making the final decision. Please only confirm that we are settling on the `catch`-based solution and I will adjust the PR accordingly.

aslushnikov:
This seems stale and looks like it works as intended.

PR #10742, opened 2021-12-07
fix: support shadow DOM with Vue and React selectors

### LINKED ISSUE #10123: [BUG] React selectors don't work inside shadow root
**Context:**
- Playwright Version: 1.16.3
- Operating System: Windows
- Node.js version: 16.13.0
- Browser: Chromium

**Code Snippet**

```javascript
const { chromium, expect } = require('@playwright/test');

(async () => {
  await page.goto('https://playwright-shadow-root.netlify.app/');
  const locator = page.locator('_react=ToolbarItem');
  await expect(locator).toHaveCount(6);
})();

```

**Describe the bug**

`react-shadow` is used to create the shadow root. You can see the source code here: https://github.com/cheminfo/analysis-ui-components/blob/a0f8ca8ff120f9fba155c42baccfdc0867a623c0/src/layout/RootLayout.tsx

If `<root.div>` is replaced with `<div>`, the test passes.

Note that the React developer tools are able to see and inspect these elements:
image

aslushnikov:
There were two issues:
- we did not find VDom roots inside shadow DOM
- we incorrectly relied on DOM's `contain` method to determine if
  VDom's rendered node belongs to requested scope.

Fixes #10123

### DIFF
--- a/packages/playwright-core/src/server/common/componentUtils.ts
+++ b/packages/playwright-core/src/server/common/componentUtils.ts
@@ -200,3 +200,13 @@ export function parseComponentSelector(selector: string): ParsedComponentSelecto
     throw new Error(`Error while parsing selector \`${selector}\` - selector cannot be empty`);
   return result;
 }
+
+export function isInsideScope(scope: Node, node: Node): boolean {
+  while (node) {
+    if (scope.contains(node))
+      return true;
+    node = node.parentNode || (node as ShadowRoot).host;
+  }
+  return false;
+}
+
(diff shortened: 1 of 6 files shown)


packages/playwright-core/src/server/common/componentUtils.ts
```diff
@@ -200,3 +200,13 @@ export function parseComponentSelector(selector: string): ParsedComponentSelecto
     throw new Error(`Error while parsing selector \`${selector}\` - selector cannot be empty`);
   return result;
 }
+
+export function isInsideScope(scope: Node, node: Node): boolean {
+  while (node) {
+    if (scope.contains(node))
+      return true;
+    node = node.parentNode || (node as ShadowRoot).host;
```
dgozman, line 208: There is an `HTMLAnchorElement.host` property  that breaks this code. Take a look at `parentNodeOrShadowHost` method we have somewhere.

dgozman (approved):

packages/playwright-core/src/server/common/componentUtils.ts
aslushnikov: Done

PR #10761, opened 2021-12-07
chore: validate versions only on remote connections

mxschmitt:
Motivation: The recently added (#10542) client/server validation does make much sense when using it with the driver. This patch moves it out into the WS server so its only used when using launchServer/connect OR PlaywrightClient/PlaywrightServer.

### DIFF
--- a/packages/playwright-core/src/protocol/protocol.yml
+++ b/packages/playwright-core/src/protocol/protocol.yml
@@ -423,7 +423,6 @@ Root:
     initialize:
       parameters:
         sdkLanguage: string
-        version: string
       returns:
         playwright: Playwright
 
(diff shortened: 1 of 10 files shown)


packages/playwright-core/src/protocol/protocol.yml
```diff
@@ -3032,4 +3031,5 @@ JsonPipe:
         message: json
 
     closed:
+      parameters:
```
mxschmitt, line 3034: this seemed to be like a bug in the YAML where the errors didn't get correctly plumped through.

pavelfeldman:
I'm a bit confused. Can we start with reverting the original patch that we are re-working and link to a bug that prompted it?

mxschmitt:
Reverting for now the initial implementation and waiting for more user reports.

PR #10769, opened 2021-12-07
devops: run docker test in headful mode too

yury-s:
Some test failures (like [this one](https://github.com/microsoft/playwright/pull/10741)) are reproducible in headful mode in docker while the same tests pass on GHA in headful mode.

### DIFF
--- a/.github/workflows/tests_docker.yml
+++ b/.github/workflows/tests_docker.yml
@@ -29,6 +29,7 @@ jobs:
       matrix:
         tag: [bionic, focal]
         user: [pwuser, root]
+        headless: ['headless', 'headful']
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/tests_docker.yml
```diff
@@ -22,13 +22,14 @@ env:
...
   test_linux_docker:
-    name: "Docker Ubuntu-${{ matrix.tag }}-${{ matrix.user }} Tests"
+    name: "Docker Ubuntu-${{ matrix.tag }}-${{ matrix.user }}-${{ matrix.headless }} Tests"
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
         tag: [bionic, focal]
         user: [pwuser, root]
+        headless: ['headless', 'headful']
```
mxschmitt, line 32:
```suggestion
        headless: [true, false]
```

.github/workflows/tests_docker.yml
```diff
@@ -59,7 +60,11 @@ jobs:
       if: matrix.user == 'root'
       run: docker exec playwright-docker-${{ matrix.tag }}-test npx playwright install chrome msedge
     - name: Run "npm run test" inside docker
+      if: matrix.headless == 'headless'
```
mxschmitt, line 63:
```suggestion
      if: matrix.headless
```

.github/workflows/tests_docker.yml
```diff
@@ -59,7 +60,11 @@ jobs:
...
       run: docker exec -e INSIDE_DOCKER=1 -e CI=1 playwright-docker-${{ matrix.tag }}-test xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" npm run test
+    - name: Run "HEADFUL=1 npm run test" inside docker
+      if: matrix.headless == 'headful'
```
mxschmitt, line 66:
```suggestion
      if: !matrix.headless
```
this could work, or maybe we need:

```suggestion
      if: matrix.headless == false
```

mxschmitt (approved):
(This increases the amount of Docker bots from 4 to 8 which usually take an average of 1 hour per job. So not sure if its worth it.)

.github/workflows/tests_docker.yml
mxschmitt: nevermind, you need it for the dynamic name.

aslushnikov:
Let's not land this to save on CPU load

PR #10778, opened 2021-12-08
fix(html): render text attachments as text

### LINKED ISSUE #10076: [BUG] HTML Reporter Renders Plaintext Attachments as Base64 Encoded Text
**Context:**
Playwright Version: `@playwright/test@1.16.3`, ToT (3c1aaa5338f9822d39a4c69c4ddd19acc12d0530)

**Code Snippet**

Currently, if you have the following in your tests:

```
  testInfo.attachments.push({
    name: "example.txt",
    contentType: "text/plain",
    body: Buffer.from("foo"),
  });

  testInfo.attachments.push({
    name: "example.json",
    body: Buffer.from(JSON.stringify({ foo: 1})),
  });

  console.log("stdout output example");
  console.error("stderr output example");
```

the HTML Reporter renders the `example.txt` and `example.json` as b64 encoded text (but, notably inconsistent, not `stdout`/`stderr`):

Screen Shot 2021-11-05 at 12 42 19 AM

It would be ideal if plaintext attachments were rendered plainly without the encoding so you can easily understand their contents while viewing the HTML Report in the browser. For each of our e2e smoketests, we attach metadata that allows a test reviewer to find items created in the e2e test run in our backend.) Here's what I expected it to look like:

Screen Shot 2021-11-05 at 12 40 08 AM

For now, we can log the info to stdout, but it's nice to put it in a standard attachment so it's easy to view.

It looks like stdout/stderr have special treatment, while all other `body` type attachments are base64 encoded:
https://github.com/microsoft/playwright/blob/3c1aaa5338f9822d39a4c69c4ddd19acc12d0530/packages/playwright-test/src/reporters/raw.ts#L248

There's a few alternative's I'd like us to consider:

1. Don't always do the above. If `body` Buffer is specified AND `contentType` is plaintext-ish (`text/*`, `application/json`), just call `toString()` instead of `toString('base64')`.
2. Leave as is, but change `_createAttachment` like so:
    ```diff
    diff --git a/packages/playwright-test/src/reporters/html.ts b/packages/playwright-test/src/reporters/html.ts
    index 2768b4cc..565bf6f5 100644
    --- a/packages/playwright-test/src/reporters/html.ts
    +++ b/packages/playwright-test/src/reporters/html.ts
    @@ -390,15 +390,18 @@ class HtmlBuilder {
              };
            }
    
    -        if ((a.name === 'stdout' || a.name === 'stderr') && a.contentType === 'text/plain') {
    -          if (lastAttachment &&
    +        if ((a.name === 'stdout' || a.name === 'stderr')) {
    +          if (a.contentType === 'text/plain' && lastAttachment &&
                lastAttachment.name === a.name &&
                lastAttachment.contentType === a.contentType) {
                lastAttachment.body += stripAnsiEscapes(a.body as string);
                return null;
              }
              a.body = stripAnsiEscapes(a.body as string);
    +        } else if (a.body && (a.contentType.toLowerCase().startsWith("text/") || a.contentType.toLowerCase() === "application/json")) {
    +          a.body = Buffer.from(a.body, "base64").toString();
            }
    ```
3. Similar to (2), but instead of in `_createAttachment`, conditionally decode it in the React Component around this existing line: https://github.com/microsoft/playwright/blob/3c1aaa5338f9822d39a4c69c4ddd19acc12d0530/packages/playwright-core/src/web/htmlReport/htmlReport.tsx#L326
4. Be more explicit. Change `TestInfo.body` from `Buffer | undefined` to `Buffer | string | undefined`. If the user supplies a `string` instead of a `Buffer`, update the following line to just use the raw string as the `body`: https://github.com/microsoft/playwright/blob/3c1aaa5338f9822d39a4c69c4ddd19acc12d0530/packages/playwright-test/src/reporters/raw.ts#L248

As an aside, it might be worth cleaning up the serialized attachments format so it's easy to tell if they've been base64 encoded or not. For instance, some attachments (`stderr` / `stdout`) store non-encoded data currently, while user-attachments are `base64` encoded, and there's not a programmatic way to tell the difference in this mix—except magically knowing. I think it would make sense to always b64 encode the body (event if it's stderr/stdout) so it's uniform, or introduce an `encoding` hint (`undefined | "base64"`) in the serialized format.

@pavelfeldman @dgozman : What's your take on the expected behavior/rendering here? If it should be plaintext for my examples above (which would greatly help us when going through our Test Reports), then what are your above thoughts on the correct fix? Happy to contribute an MR for one of the suggestions (or an alternative). Running (2) in a staging environment currently to get around the issue.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/10076

### DIFF
--- a/packages/playwright-test/src/reporters/raw.ts
+++ b/packages/playwright-test/src/reporters/raw.ts
@@ -71,7 +71,7 @@ export type JsonTestCase = {
 
 export type JsonAttachment = {
   name: string;
-  body?: string;
+  body?: string | Buffer;
   path?: string;
   contentType: string;
 };
@@ -245,7 +245,7 @@ class RawReporter {
         attachments.push({
           name: attachment.name,
           contentType: attachment.contentType,
-          body: attachment.body.toString('base64')
+          body: attachment.body
         });
       } else if (attachment.path) {
         attachments.push({
@@ -274,7 +274,7 @@ class RawReporter {
     return {
       name: type,
       contentType: 'application/octet-stream',
-      body: chunk.toString('base64')
+      body: chunk
     };
   }
 
--- a/packages/playwright-test/src/reporters/html.ts
+++ b/packages/playwright-test/src/reporters/html.ts
@@ -22,7 +22,7 @@ import { Transform, TransformCallback } from 'stream';
 import { FullConfig, Suite, Reporter } from '../../types/testReporter';
 import { HttpServer } from 'playwright-core/lib/utils/httpServer';
 import { calculateSha1, removeFolders } from 'playwright-core/lib/utils/utils';
-import RawReporter, { JsonReport, JsonSuite, JsonTestCase, JsonTestResult, JsonTestStep, JsonAttachment } from './raw';
+import RawReporter, { JsonReport, JsonSuite, JsonTestCase, JsonTestResult, JsonTestStep } from './raw';
 import assert from 'assert';
 import yazl from 'yazl';
 import { stripAnsiEscapes } from './base';
@@ -78,7 +78,13 @@ export type TestCase = TestCaseSummary & {
   results: TestResult[];
 };
 
-export type TestAttachment = JsonAttachment;
+export type TestAttachment = {
+  name: string;
+  body?: string;
+  path?: string;
+  contentType: string;
+};
+
 
 export type TestResult = {
   retry: number;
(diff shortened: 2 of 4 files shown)


packages/playwright-test/src/reporters/raw.ts
```diff
@@ -71,7 +71,7 @@ export type JsonTestCase = {
 
 export type JsonAttachment = {
   name: string;
-  body?: string;
+  body?: string | Buffer;
```
dgozman, line 74: Not quite "json" attachment anymore.

dgozman (approved):

packages/playwright-test/src/reporters/raw.ts
pavelfeldman: Agreed, but it is the right thing to do here...

PR #10781, opened 2021-12-08
test: skip input type=color test on WebKit/Windows

mxschmitt:
Its disabled by default, see here (git blame says since 6 years):

https://github.com/WebKit/WebKit/blob/873d43643d869e63be122e17cea53a7b4141c778/Source/cmake/OptionsWin.cmake#L36

and failing by that: https://devops.aslushnikov.com/flakiness2.html#filter_spec=uld+throw+on+incorrect+color+val

Probably fine to just skip it.


mxschmitt:
Superseded by #10792.

PR #10789, opened 2021-12-08
fix: omit page.route

houmingjie:
const routes = [1,2,3];
for(const v of routes){
    console.info(v);
    routes.splice(routes.indexOf(v),1)
}
got  1,3 but 2 has been omitted

### DIFF
--- a/packages/playwright-core/src/client/page.ts
+++ b/packages/playwright-core/src/client/page.ts
@@ -168,7 +168,7 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
   }
 
   private _onRoute(route: Route, request: Request) {
-    for (const routeHandler of this._routes) {
+    for (const routeHandler of [...this._routes]) {
       if (routeHandler.matches(request.url())) {
         if (routeHandler.handle(route, request)) {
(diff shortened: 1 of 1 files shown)


mxschmitt:
@houmingjie can you also add a test for the scenario which was broken?

The tests are located here:
- https://github.com/microsoft/playwright/blob/main/tests/browsercontext-route.spec.ts
- https://github.com/microsoft/playwright/blob/main/tests/page/page-route.spec.ts

Make sure to add your fix also to `BrowserContext.route()` where we have probably the same issue.

houmingjie:
> @houmingjie can you also add a test for the scenario which was broken?
> 
> The tests are located here:
> 
> * https://github.com/microsoft/playwright/blob/main/tests/browsercontext-route.spec.ts
> * https://github.com/microsoft/playwright/blob/main/tests/page/page-route.spec.ts
> 
> Make sure to add your fix also to `BrowserContext.route()` where we have probably the same issue.

gotcha, woking on it

houmingjie:
Never mind,just misunderstood route,it only can be matched once.
Sorry to bother.

PR #10796, opened 2021-12-08
docs(.net): update installation docs

### LINKED ISSUE #1834: Test against .NET 6
Currently only 3.1 and 5 are tested. Users already using it with .NET 6 and it seems not working.

pavelfeldman:
Fixes https://github.com/microsoft/playwright-dotnet/issues/1834

### DIFF
--- a/docs/src/cli.md
+++ b/docs/src/cli.md
@@ -98,7 +98,7 @@ playwright install --help
 ```
 
 ```bash csharp
-playwright install --help
+bin\Debug\netX\playwright.ps1 install --help
 ```
 
 ## Generate code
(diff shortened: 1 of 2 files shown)


docs/src/cli.md
```diff
@@ -98,7 +98,7 @@ playwright install --help
 ```
 
 ```bash csharp
-playwright install --help
+bin\Debug\netX\playwright.ps1 install --help
```
mxschmitt, line 101: note: netX might confuse users and it requires PowerShell which is not installed on most Linux machines. Or does .NET include PowerShell?

mxschmitt (approved):

docs/src/cli.md
pavelfeldman: Mac/Linux .net people know what to do with the powershell scripts.

PR #10797, opened 2021-12-08
    feat(test-runner): add trace.zip link to html report

### LINKED ISSUE #10790: [Feature] @playwright/test: add download trace.zip link to HTML report

Hi, thank you for awesome tool!  i've noticed that on some CIs (e.g. JENKINS) HTML report link to trace zip doesn't work (see screenshot below), so it would be useful to download trace.zip with link, same as screenshot link.

Jenkins after navigating to trace:
image

image

dartyuhov:
This patch adds download link for collected during test traces.zip into HTML report in playwright-test

    Fixes #10790


pavelfeldman:
I'm sorry, I did not notice your PR and landed the fix. Do you want to rebaseline it to only include the test?

dartyuhov:
> I'm sorry, I did not notice your PR and landed the fix. Do you want to rebaseline it to only include the test?

Hi, i've also added download link to HTML report, not only the test :)

more info in attached feature request #10790

dartyuhov:
> > I'm sorry, I did not notice your PR and landed the fix. Do you want to rebaseline it to only include the test?
> 
> Hi, i've also added download link to HTML report, not only the test :)
> 
> more info in attached feature request #10790

but as far as i can see someone already added it in https://github.com/microsoft/playwright/pull/10807
so this pr only for test then, if test is not required from your point of view, we can close this PR

PR #10798, opened 2021-12-08
fix: use correct name for React component case

pahan35:
https://techterms.com/definition/pascalcase


mxschmitt (approved):

mxschmitt:
Hey! Just FYI we had a discussion in the team about it and reverted it for now. We think that camelCase might be more appropriate for the target group who are mainly junior engineers. (even though that pascal case is formally correct).

pahan35:
> Hey! Just FYI we had a discussion in the team about it and reverted it for now. We think that camelCase might be more appropriate for the target group who are mainly junior engineers. (even though that pascal case is formally correct).

@mxschmitt I fixed the naming to avoid misleading our QA members with incorrect case naming. But if you think that this is better to continue to use the wrong name, let keep it as is. Frankly speaking, it was confusing to me to see **CamelCase**.

PR #10812, opened 2021-12-08
fix: use HTTP/2.0 instead of h2 as Response.httpVersion

### LINKED ISSUE #10751: [BUG] Invalid httpVersion (h2) in recorded har file
**Context:**
- Playwright Version: 1.13+
- Operating System: Windows, Linux
- Node.js version: 12
- Browser: Chromium

**Describe the bug**

We use recordHar option in our tests.
The har file is created but we are not able to import it in fiddler. It shows all sessions in red.
After investigation, we found that it is caused by an invalid value in httpVersion.
It has "h2" as a value, while it should be "HTTP/2.0"
When we rollback pw to v12 the value in har file is "HTTP/1.1"

mxschmitt:
Fixes #10751

### DIFF
--- a/packages/playwright-core/src/server/network.ts
+++ b/packages/playwright-core/src/server/network.ts
@@ -434,6 +434,8 @@ export class Response extends SdkObject {
       return 'HTTP/1.1';
     if (this._httpVersion === 'http/1.1')
       return 'HTTP/1.1';
+    if (this._httpVersion === 'h2')
+      return 'HTTP/2.0';
     return this._httpVersion;
   }
 
--- a/tests/har.spec.ts
+++ b/tests/har.spec.ts
@@ -512,8 +512,8 @@ it('should contain http2 for http2 requests', async ({ contextFactory, browserNa
   const { page, getLog } = await pageWithHar(contextFactory, testInfo);
   await page.goto(`https://localhost:${(server.address() as AddressInfo).port}`);
   const log = await getLog();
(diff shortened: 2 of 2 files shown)


packages/playwright-core/src/server/network.ts
```diff
@@ -430,11 +430,12 @@ export class Response extends SdkObject {
...
 
   httpVersion(): string {
-    if (!this._httpVersion)
+    const httpVersion = this._httpVersion?.toUpperCase();
+    if (!httpVersion)
       return 'HTTP/1.1';
-    if (this._httpVersion === 'http/1.1')
-      return 'HTTP/1.1';
-    return this._httpVersion;
+    if (['HTTP/2.0', 'H2'].includes(httpVersion))
```
mxschmitt, line 436: HTTP/2.0 seems to be the old variant, Nowadays its referred to it as HTTP/2.

packages/playwright-core/src/server/network.ts
```diff
@@ -430,11 +430,12 @@ export class Response extends SdkObject {
   }
 
   httpVersion(): string {
-    if (!this._httpVersion)
+    const httpVersion = this._httpVersion?.toUpperCase();
```
dgozman, line 433: Let's be careful and only convert 'h2/h2c' into 'HTTP/2.0'. There is stuff like spdy or quic.

dgozman (approved):

PR #10823, opened 2021-12-09
Use pixel buffer diff for 4x faster image diffng and reduced visual noise

p01:
Pixel-buffer-diff uses the same image diffing technique as Pixelmatch, but does not stop at individual pixels. It take into account the cumulated differences to reduce false positive.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6703,24 +6703,10 @@
         "node": ">= 6"
       }
     },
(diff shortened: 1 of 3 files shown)


aslushnikov:
Thank you for the heads-up. We'll refrain from any change here for now; as we look more closely into Visual Regression Testing, we might end up using something completely different.

So closing this PR for now.

PR #10828, opened 2021-12-09
feat(keyboard): support simple copy-pasting using meta+c/v

### LINKED ISSUE #12000: [Bug]: Copy content with Meta+A and Meta+C
### Discussed in https://github.com/microsoft/playwright/discussions/11930

<div type='discussions-op-text'>

<sup>Originally posted by **saguilarolmo** February  8, 2022</sup>
I'm doing a test where I have to simulate the user's action when doing `ctrl+A` (a paragraph is automatically selected) and then `ctrl+C` to check that the content of the clipboard is the same as the selected text.
To do this, I have tried to do this:

```
const codeText = await this.page.codeText.innerText()
  await this.page.context().grantPermissions(["clipboard-write"])
  await this.page.codeText.focus()

  await this.page.keyboard.press("Control+A")
  await this.page.keyboard.press("Control+C")
...
```

This test works when I run it inside a docker container using Linux.

The problem is when I try to run this same test in a macOS environment. To do so, the only thing that changes is `Control+A` for `Meta+A` and the same for C.

```
const codeText = await this.page.codeText.innerText()
  await this.page.context().grantPermissions(["clipboard-write"])
  await this.page.codeText.focus()

  await this.page.keyboard.press("Meta+A")
  await this.page.keyboard.press("Meta+C")
...
```

I have checked that `Meta+A` selects the text correctly, but `Meta+C` does not save the selected content to the clipboard.

Is there a reason for this or is it a bug?
</div>

Andarist:
It's a straightforward change to support new, common, keyboard commands

Note that I've tested this locally with Chrome on my Mac but it seems that CI doesn't want to pass Chrome tests - it's running on ubuntu though. Does this mean that I should introduce per-platform editing commands? At the moment there is only a single [`macEditingCommands`](https://github.com/microsoft/playwright/blob/0ed33522c572b974905f642408050d007860e33e/packages/playwright-core/src/server/macEditingCommands.ts) file.

Fixes https://github.com/microsoft/playwright/issues/12000

### DIFF
--- a/tests/page/page-keyboard.spec.ts
+++ b/tests/page/page-keyboard.spec.ts
@@ -464,7 +464,8 @@ it('should dispatch a click event on a button when Enter gets pressed', async ({
   expect((await actual.jsonValue()).clicked).toBe(true);
 });
 
-it('should support simple copy-pasting', async ({ page, isMac }) => {
+it('should support simple copy-pasting', async ({ page, isMac, browserName }) => {
+  it.fixme(browserName === 'webkit', 'https://github.com/microsoft/playwright/issues/12000');
   const modifier = isMac ? 'Meta' : 'Control';
   await page.setContent(`<div contenteditable>123</div>`);
   await page.focus('div');
--- a/tests/page/page-keyboard.spec.ts
+++ b/tests/page/page-keyboard.spec.ts
@@ -465,13 +465,13 @@ it('should dispatch a click event on a button when Enter gets pressed', async ({
 });
 
 it('should support simple copy-pasting', async ({ page, isMac }) => {
-  it.skip(!isMac);
+  const modifier = isMac ? 'Meta' : 'Control';
   await page.setContent(`<div contenteditable>123</div>`);
   await page.focus('div');
-  await page.keyboard.press('Meta+KeyA');
-  await page.keyboard.press('Meta+KeyC');
-  await page.keyboard.press('Meta+KeyV');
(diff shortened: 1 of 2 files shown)


yury-s:
>  CI doesn't want to pass Chrome tests - it's running on ubuntu though. Does this mean that I should introduce per-platform editing commands?

The keystrokes are macOS-specific, you just need to skip the test on other platforms, see [this test](https://github.com/microsoft/playwright/blob/4bbd38560a9a2ec91f63f2abee0cffb874357204/tests/page/page-keyboard.spec.ts#L434-L435) for example.

Andarist:
@yury-s I've applied the suggested change. One test suite is failing but it seems unrelated to this PR.

q: is there no support for non-mac editing commands at all? can those kinds of tests be only written for Macs at the moment?

tests/page/page-keyboard.spec.ts
```diff
@@ -464,6 +464,17 @@ it('should dispatch a click event on a button when Enter gets pressed', async ({
   expect((await actual.jsonValue()).clicked).toBe(true);
 });
 
+it('should support simple copy-pasting', async ({ page, isMac }) => {
+  it.skip(!isMac);
```
aslushnikov, line 468:
```suggestion
    const modifier = isMac ? 'Meta' : 'Control';
```

And let's use this modifier below to check on other platforms.

aslushnikov (approved):

PR #10834, opened 2021-12-09
docs(protracor): add a simpler "awaitForAngular"

nicojs:
Fixes: <issue #10548>

A small docs update. I reverse-engineered a better `awaitForAngular` 😎.

### DIFF
--- a/docs/src/protractor-js.md
+++ b/docs/src/protractor-js.md
@@ -117,7 +117,22 @@ Here's how to polyfill `waitForAngular` function in Playwright Test:
     ​await executeScriptAsync(page, clientSideScripts.waitForAngular, '');
     }
     ```
-
+   If you don't want to keep a version protractor around, you can also use this simpler approach using this function:
+   ```js
+   async function waitForAngular(page) {
+     await page.evaluate(async () => {
+       // @ts-expect-error
+       if (window.getAllAngularTestabilities) {
+         // @ts-expect-error
+         await Promise.all(window.getAllAngularTestabilities().map(whenStable));
+         // @ts-expect-error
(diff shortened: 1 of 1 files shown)


docs/src/protractor-js.md
```diff
@@ -117,7 +117,22 @@ Here's how to polyfill `waitForAngular` function in Playwright Test:
     ​await executeScriptAsync(page, clientSideScripts.waitForAngular, '');
     }
     ```
-
+   If you don't want to keep a version protractor around, you can also use this simpler approach using this function:
+   ```js
+   async function waitForAngular(page) {
+     await page.evaluate(async () => {
+       // @ts-expect-error
+       if (window.getAllAngularTestabilities) {
```
aslushnikov, line 125: do I understand correctly that this works only with Angular2?

nicojs: Yes, Angular 2+. I personally haven't seen an AngularJS project for a long time now. LTS for AngularJS runs until December 31, 2021, so I'd personally be happy not to include it.

aslushnikov: Awesome, let's clarify this in the docs, and it'll be good to go!

nicojs: Thanks! Just did that.

aslushnikov (approved):

nicojs:
Thanks for merging @aslushnikov 

When is this documentation published on the website? I want to use this for an Angular training I'm working on.

mxschmitt:
We publish the documentation every few days to the `/next` version. The stable one gets published once a new stable release of Playwright happens, for 1.18 in around one month.

PR #10846, opened 2021-12-10
fix(tracing): trace frame

### LINKED ISSUE #10845: [BUG] Snapshotter does not replaces frame src attr
**Context:**
- Playwright Version: 1.17.1(also reproduced on `main` branch)
- Operating System: Mac
- Node.js version: v14.15.3
- Browser: All

**Code Snippet**

```javascript

(async () => {
  const ctx = await browser.newContext();
  await ctx.tracing.start({ screenshots: true, snapshots: true });
  const page = await ctx.newPage();
  await page.goto(`http://abehiroshi.la.coocan.jp/`);
  await ctx.tracing.stop({ path: path.join(process.cwd(), "trace.zip") });
  await browser.close();
})();
```

**Describe the bug**
frame `src` attribute is not replaced by snapshotted version. so snapshotted frame content is not shown on trace viewer.

musou1500:
Fixes: #10845

### DIFF
--- a/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
+++ b/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
@@ -437,6 +437,8 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
               continue;
             if (nodeName === 'IFRAME' && (name === 'src' || name === 'sandbox'))
               continue;
+            if (nodeName === 'FRAME' && name === 'src')
+              continue;
             let value = element.attributes[i].value;
             if (nodeName === 'META')
               value = this.__sanitizeMetaAttribute(name, value, (node as HTMLMetaElement).httpEquiv);
--- a/packages/playwright-core/src/web/traceViewer/snapshotRenderer.ts
+++ b/packages/playwright-core/src/web/traceViewer/snapshotRenderer.ts
@@ -85,9 +85,9 @@ export class SnapshotRenderer {
       return { html: '', pageId: snapshot.pageId, frameId: snapshot.frameId, index: this._index };
 
     // Hide the document in order to prevent flickering. We will unhide once script has processed shadow.
-    const hideAllStyle = '<style>*,*::before,*::after { visibility: hidden }</style>';
-    const prefix = snapshot.doctype ? `<!DOCTYPE ${snapshot.doctype}>` + hideAllStyle : hideAllStyle;
+    const prefix = snapshot.doctype ? `<!DOCTYPE ${snapshot.doctype}>` : '';
     html = prefix + [
+      '<style>*,*::before,*::after { visibility: hidden }</style>',
       `<style>*[__playwright_target__="${this.snapshotName}"] { background-color: #6fa8dc7f; }</style>`,
       `<script>${snapshotScript()}</script>`
     ].join('') + html;
--- a/packages/playwright-core/src/web/traceViewer/snapshotRenderer.ts
+++ b/packages/playwright-core/src/web/traceViewer/snapshotRenderer.ts
@@ -87,10 +87,10 @@ export class SnapshotRenderer {
     // Hide the document in order to prevent flickering. We will unhide once script has processed shadow.
     const hideAllStyle = '<style>*,*::before,*::after { visibility: hidden }</style>';
     const prefix = snapshot.doctype ? `<!DOCTYPE ${snapshot.doctype}>` + hideAllStyle : hideAllStyle;
-    html = prefix + `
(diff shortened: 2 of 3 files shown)


packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
```diff
@@ -419,6 +419,8 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
               continue;
             if (nodeName === 'IFRAME' && name === 'src')
               continue;
+            if (nodeName === 'FRAME' && name === 'src')
```
dgozman, line 422: Let's combine this with previous check like `(nodeName === 'IFRAME' || nodeName === 'FRAME')`. If you rebase on top of the fresh "main" branch, that will make more sense.

packages/playwright-core/src/web/traceViewer/snapshotRenderer.ts
```diff
@@ -87,10 +87,10 @@ export class SnapshotRenderer {
...
     const hideAllStyle = '<style>*,*::before,*::after { visibility: hidden }</style>';
     const prefix = snapshot.doctype ? `<!DOCTYPE ${snapshot.doctype}>` + hideAllStyle : hideAllStyle;
-    html = prefix + html + `
-      <style>*[__playwright_target__="${this.snapshotName}"] { background-color: #6fa8dc7f; }</style>
-      <script>${snapshotScript()}</script>
-    `;
+    html = prefix + [
+      `<style>*[__playwright_target__="${this.snapshotName}"] { background-color: #6fa8dc7f; }</style>`,
+      `<script>${snapshotScript()}</script>`
+    ].join('') + html;
```
dgozman, line 93: Why this change?

packages/playwright-core/src/web/traceViewer/snapshotRenderer.ts
```diff
@@ -252,7 +251,11 @@ function snapshotScript() {
       }
       document.styleSheets[0].disabled = true;
     };
+
+    const onDOMContentLoaded = () => visit(document);
```
dgozman, line 255: Why this change?

tests/snapshotter.spec.ts
```diff
@@ -211,14 +231,32 @@ it.describe('snapshots', () => {
...
 });
 
+
+async function waitUntilPass(fn: (counter: number) => Promise<unknown>, sleep = 250) {
+  let counter = 0;
+  for (; ; ++counter) {
+    let err: unknown;
+    try {
+      await fn(counter);
+    } catch (e) {
```
dgozman, line 241: I don't really like that we rely on the exception - who knows what else could throw in this function? Can we do `captureSnapshotWithFrames` helper that would only do the frames trickery?

dgozman (commented):
Very nice! A few comments below.

tests/snapshotter.spec.ts
musou1500: Is `captureSnapshotWithFrames` like following?

```typescript
async function captureSnapshotWithFrames(snapshotter: InMemorySnapshotter, page: Page, toImpl:(rpcObject: any) => any) {
  for (let counter = 0; ; ++counter) {
    const snapshot = await snapshotter.captureSnapshot(toImpl(page), 'snapshot' + counter);
    const text = distillSnapshot(snapshot).replace(/frame@[^"]+["]/, '<id>"');
    if (text === '<IFRAME __playwright_src__=\"/snapshot/<id>\"></IFRAME>')
      break;
    await page.waitForTimeout(250);
  }
}
```

musou1500 (commented):
Thank you for reviewing. I wrote some comments.

tests/snapshotter.spec.ts
dgozman: Yeah, something like that. Or you can just duplicate this code in both tests, that's totally fine and probably easier than doing some helpers.

packages/playwright-core/src/web/traceViewer/snapshotRenderer.ts
dgozman: I see, thank you for the explanation. You can now merge hideAllStyle with this `<style>` tag - would be easier to follow.

dgozman (commented):
Looks good, let's figure our the tests and simplify a bit.

musou1500:
thank you. I fixed.
considering `distilSnapshot` implementation, I added new `hideAllStyle` style tag to array instead of add styles into existing style tag.

dgozman:
@musou1500 This looks great, but many bot are unhappy. I suspect that's not your changes. Could you please rebase on top of current `main` branch and push, so I'll rerun the tests.

musou1500:
@dgozman I rebased.

dgozman:
Everything looks good, merging.

Issue #10847, opened 2021-12-10
[BUG] meta tag specifies charset remains in dom snapshot


musou1500:
**Context:**
- Playwright Version: 1.17.1(also reproduced on `main` branch)
- Operating System: Mac
- Node.js version: v14.15.3
- Browser: All

**Code Snippet**

```javascript
const {chromium, webkit, firefox} = require('playwright');

(async () => {
  const browser = await chromium.launch();
  const ctx = await browser.newContext();
  await ctx.tracing.start({ screenshots: true, snapshots: true });
  const page = await ctx.newPage();
  await page.goto(`https://www.shtml.jp/mojibake/meta.html`);
  await ctx.tracing.stop({ path: path.join(process.cwd(), "trace.zip") });
  await browser.close();
})();
```

**Describe the bug**

meta tag specifies charset(for example: `<meta http-equiv="Content-Type" content="text/html; charset=Shift_JIS">`) remains in dom snapshot.
so garbled text is appears on trace viewer.

PR #10848, opened 2021-12-10
fix(tracing): overwrite attr value specifies charset other than utf-8

musou1500:
Fixes: <issue #10847>

### DIFF
--- a/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
+++ b/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
@@ -420,7 +420,11 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
             if (nodeName === 'IFRAME' && name === 'src')
               continue;
             let value = element.attributes[i].value;
-            if (name === 'src' && (nodeName === 'IMG'))
+            if (name === 'charset' && nodeName === 'META')
+              value = 'utf-8';
+            else if (nodeName === 'META' && (node as HTMLMetaElement).httpEquiv.toLowerCase() === 'content-type' && name === 'content')
+              value = 'text/html; charset=utf-8';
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
```diff
@@ -420,7 +420,11 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
             if (nodeName === 'IFRAME' && name === 'src')
               continue;
             let value = element.attributes[i].value;
-            if (name === 'src' && (nodeName === 'IMG'))
+            if (name === 'charset' && nodeName === 'META')
+              value = 'utf-8';
+            else if (nodeName === 'META' && (node as HTMLMetaElement).httpEquiv.toLowerCase() === 'content-type' && name === 'content')
+              value = 'text/html; charset=utf-8';
```
dgozman, line 426: Let's only rewrite the charset inside the content? I'm worried about non-html frames, for example `<meta http-equiv="Content-Type" content="image/svg+xml">`.

I'd also make a separate function like `value = this._sanitizeMetaAttribute(name, value, (node as HTMLMetaElement).httpEquiv)`

dgozman (commented):
Thank you for the great PR!

packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
musou1500: I implemented `_sanitizeMetaAttribute`.

musou1500: but `_sanitizeMetaAttribute` may not be needed because `meta` tag is html specific.

dgozman (approved):

PR #10868, opened 2021-12-10
chore: create trace.zip in driver for local runs

yury-s:


packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -219,18 +219,19 @@ export class Tracing implements InstrumentationListener, SnapshotterDelegate, Ha
         entries.push({ name: path.join('resources', sha1), value: path.join(this._resourcesDir, sha1) });
 
       const sourceEntries: NameValue[] = [];
+      const targetForSources = (params.mode === 'compressTraceAndSources') ? entries : sourceEntries;
       for (const value of state.sources)
-        sourceEntries.push({ name: 'resources/src@' + calculateSha1(value) + '.txt', value });
+        targetForSources.push({ name: 'resources/src@' + calculateSha1(value) + '.txt', value });
```
dgozman, line 224: Let's put an if statement here - I find it hard to understand that we push into one of the two arrays and that affects the semantics.

packages/playwright-core/src/protocol/protocol.yml
```diff
@@ -833,13 +833,14 @@ BrowserContext:
...
+          literals:
+          - doNotSave
+          - compressTrace
+          - compressTraceAndSources
       returns:
         artifact: Artifact?
-        entries:
-          type: array
-          items: NameValue
         sourceEntries:
```
dgozman, line 844: Perhaps this should be optional now? We only return it when `mode === compressTrace`, right?

dgozman (approved):

packages/playwright-core/src/server/trace/recorder/tracing.ts
yury-s: Done.

PR #10891, opened 2021-12-13
test: wait for screenshots to get into trace before stopping

yury-s:
The test has been flaky [recently](https://devops.aslushnikov.com/flakiness2.html#filter_spec=should+not+include+trace+resources+from+the+provious+chunks&timestamp=1639424793673).

### DIFF
--- a/tests/tracing.spec.ts
+++ b/tests/tracing.spec.ts
@@ -139,6 +139,8 @@ test('should not include trace resources from the provious chunks', async ({ con
   await page.goto(server.EMPTY_PAGE);
   await page.setContent('<button>Click</button>');
   await page.click('"Click"');
+  // Give it enough time for both screenshots to get into the trace.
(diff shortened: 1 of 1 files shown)


dgozman (approved):

tests/tracing.spec.ts
```diff
@@ -139,6 +139,8 @@ test('should not include trace resources from the provious chunks', async ({ con
   await page.goto(server.EMPTY_PAGE);
   await page.setContent('<button>Click</button>');
   await page.click('"Click"');
+  // Give it enough time for both screenshots to get into the trace.
+  await new Promise(f => setTimeout(f, 1000));
```
pavelfeldman, line 143: what about our users?

yury-s: We should probably address that but it's not the subject of this test which is just checking that the artifacts are properly isolated between chunks.

PR #10899, opened 2021-12-14
chore: move setting PW User-Agent on connectOverCDP to the server side

mxschmitt:
This moves the default UA from the client to the server side. On the languages we didn't set the default UA yet, but we do it in Node.js. If there is a request in the future for adding the language name, then we can do it by using the env var.


packages/playwright-core/src/server/chromium/chromium.ts
```diff
@@ -63,6 +63,11 @@ export class Chromium extends BrowserType {
     if (options.headers)
       headersMap = headersArrayToObject(options.headers, false);
 
+    if (!headersMap)
```
mxschmitt, line 66: what if the header name is lower-case?

yury-s: headersMap ??= {}; 
if (!headersMap['User-Agent'])
  headersMap['User-Agent'] = getUserAgent();

yury-s (approved):

Issue #10259, opened 2021-11-11
[BUG] Firefox tracing crash


keevie:
**Context:**
- Playwright Version: official docker image/1.16.3
- Operating System: official docker image
- Node.js version: official docker image
- Browser:Firefox

Firefox crashes quite consistently with tracing turned on (in retain-on-failure mode) in this simple case. Opening the extra tabs seems to make it worse, but I'm not sure if that is the problem, or if it is the individual tests that finish really quickly combined with the shared page object across serial tests.

The error is: page.goto: Target page, context or browser has been closed

```
import { test, expect, Page } from "@playwright/test";

test.describe.serial('does some tests', () => {
  let page: Page;

  test.beforeAll(async ({ browser }) => {
    page = await browser.newPage();
    for (let i=0;i<5;i++) {
      const newCtx = await browser.newContext();
      await newCtx.newPage();
    }
  })

  test.afterAll(async () => {
    await page.close();
  })

  test('does a thing', async () => {
    await page.goto('localhost:3000/one')
    const resultUrl = new URL(page.url());
    expect(resultUrl.pathname.split('/').slice(-1)[0]).not.toEqual('exit');
  })

  test('clicks on the page', async() => {
    await page.click('body')
  })

  test('does another thing',  async () => {
    const resultUrl = new URL(page.url());
    expect(resultUrl.pathname.split('/').slice(-1)[0]).toEqual('one');
  })

})
```

Repo with instructions for reproducing + docker stuff is here https://github.com/keevie/playwright-firefox-tracing-crash

aslushnikov:
@keevie awesome repro! Reproduces for me both inside container and on my linux workstation. Looking!

aslushnikov:
So far I noticed at least two crashes:

- `Document::PreferredColorScheme` cannot use `GetDocShell()` method:
    * `GetDocShell()` is a weakref
    * `PreferredColorScheme` is called from different threads (main & style-compute thread)
    * WeakRefs cannot be accessed from multiple threads

- Headless screencast crash (with a partially transcribed stack)

```
# +0ms
# Fatal error in /home/aslushnikov/firefox/third_party/libwebrtc/webrtc/rtc_base/criticalsection.cc, line 109 +0ms
# last system error: 0 +0ms
# Check failed: CurrentThreadIsOwner() +0ms
#  +1ms
# +0ms
Redirecting call to abort() to mozalloc_abort +0ms
 +0ms
[...]

PR #10904, opened 2021-12-14
browser(firefox): fix Firefox crashes

aslushnikov:
Review URL: https://github.com/aslushnikov/juggler/commit/aff16fc8e4ae4133bf82efce8ae3c660f8ffdaf9

This patch fixes 2 firefox crashers:
- color scheme override code was not used, but was called
  from multiple threads, which caused a weakptr use violation (cannot
  be used from multiple threads)
- snapshot listener callback was reset asynchronously, so when
  `HeadlessWindowCapturer` was destroyed, it was still occasionally
  called (see `HeadlessWindowCapturer::~HeadlessWindowCapturer`)

With this patch, I no londer experience tracing crashes in firefox.

References #10259

### DIFF
--- a/browser_patches/firefox-beta/patches/bootstrap.diff
+++ b/browser_patches/firefox-beta/patches/bootstrap.diff
@@ -290,7 +290,7 @@ index 0be8f84cd6c818e13a8d7049a3e8db7010a10a3c..6766416dc62715549f8f04ef30a30f6b
  
    bool CanSet(FieldIndex<IDX_SuspendMediaWhenInactive>, bool, ContentParent*) {
 diff --git a/docshell/base/nsDocShell.cpp b/docshell/base/nsDocShell.cpp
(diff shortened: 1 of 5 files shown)


browser_patches/firefox-beta/patches/bootstrap.diff
```diff
@@ -2789,17 +2749,16 @@ index 7f91de9e67d7ffa02de3eef1d760e5cfd05e7ad6..684293dab3e81e8a60d245f979f2051d
...
  
   private:
-+  void SetSnapshotListenerOnCompositorThread(
-+      HeadlessWidget::SnapshotListener&& listener);
 +  void UpdateDrawTarget(const LayoutDeviceIntSize& aClientSize);
 +  void PeriodicSnapshot();
 +  void TakeSnapshot();
 +
    HeadlessWidget* mWidget;
++  mozilla::ReentrantMonitor mMon;
```
yury-s, line 2761: Not sure why it has to be reentrant but I assume it doesn't hurt either.

yury-s (approved):

PR #10915, opened 2021-12-14
feat: export tagged pdf by default

### LINKED ISSUE #745: [Question] Can we generate accessible/tagged PDF using  playwright?
Hi,

Chrome has functionality to generate accessible PDFs when printing a webpage Ref - https://blog.chromium.org/2020/07/using-chrome-to-generate-more.html .

Do we have this capability with playwright as well? i.e. Can page.pdf() option of playwright generate tagged pdf?

yury-s:
Since v85 Chrome does export pdf with accessibility tags [by default](https://blog.chromium.org/2020/07/using-chrome-to-generate-more.html). Enable this in Playwright too.

Fixes: https://github.com/microsoft/playwright-java/issues/745


aslushnikov (approved):
LGTM % tagged PDF's aren't drastically larger

PR #10944, opened 2021-12-15
feat(test runner): disable test timeout on page.pause() call

### LINKED ISSUE #10353: [BUG] Playwright Inspector exits after 30 seconds if run with page.pause()
**Context:**
- Playwright Version: 1.16
- Operating System: Windows
- Node.js version: 14.18.1
- Browser: All

**Overview**
There are multiple ways to turn on the Inspector, as explained in https://playwright.dev/docs/inspector
As per docs, when we set "PWDEBUG", it also disables the timeout, by setting it to 0. This is a good idea, as it allows the user to click around the inspector without a time limit.

However, this is not the case with `page.pause()`. This option also opens the inspector, but the timeout is not set to 0, thus the inspector will force exit after 30s with e.g. 

` Slow test:  tests\my_test.test.js (30s)`

The same behavior is mentioned in a Feature request here: https://github.com/microsoft/playwright/issues/10132
And I believe it should be fixed to match the default behaviour of other options, at least for consistency reasons. For example, in Java binding, `page.pause()` does not require timeOut to be explicitly disabled.

**Code Snippet**

```javascript
test.only('Inspector Demo', async ({ page }) => {
  test.setTimeout(0); // this helps, but I believe it shouldn't be needed

  await page.pause();
});
```

dgozman:
Fixes #10353.

### DIFF
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -207,6 +207,8 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
         onApiCallBegin: (apiCall: string, stackTrace: ParsedStackTrace | null, userData: any) => {
           if (apiCall.startsWith('expect.'))
             return { userObject: null };
+          if (apiCall === 'page.pause')
+            testInfo.setTimeout(0);
           const testInfoImpl = testInfo as any;
           const step = testInfoImpl._addStep({
             location: stackTrace?.frames[0],


mxschmitt (approved):

packages/playwright-test/src/index.ts
```diff
@@ -207,6 +207,8 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
         onApiCallBegin: (apiCall: string, stackTrace: ParsedStackTrace | null, userData: any) => {
           if (apiCall.startsWith('expect.'))
             return { userObject: null };
+          if (apiCall === 'page.pause')
```
pavelfeldman, line 210: Can you do the same to expect timeout?

pavelfeldman (approved):

packages/playwright-test/src/index.ts
dgozman: Will do.

PR #10983, opened 2021-12-17
chore: add download statistics

mxschmitt:


pavelfeldman:
Collecting telemetry requires:
- formal compliance with the existing regulations
- careful planning and transparency around the process

Something like: https://code.visualstudio.com/docs/getstarted/telemetry#_gdpr-and-vs-code

PR #10984, opened 2021-12-17
chore: add blink-diff third party library

pavelfeldman:


packages/playwright-test/src/matchers/golden.ts
```diff
@@ -66,7 +68,16 @@ function compareImages(actualBuffer: Buffer | string, expectedBuffer: Buffer, mi
...
   }
   const diff = new PNG({ width: expected.width, height: expected.height });
-  const count = pixelmatch(expected.data, actual.data, diff.data, expected.width, expected.height, { threshold: 0.2, ...options });
+  const thresholdOptions = { threshold: 0.2, ...options };
+  if (process.env.PW_USE_BLINK_DIFF && mimeType === 'image/png') {
+    const diff = new BlinkDiff({
+      imageA: new PNGImage(expected as any),
+      imageB: new PNGImage(actual as any),
+    });
+    const result = diff.runSync();
```
dgozman, line 77: `{ delta: thresholdOptions.threshold * 256 }` or something like that

dgozman (approved):

packages/playwright-test/src/matchers/golden.ts
pavelfeldman: I'll follow up.

PR #11004, opened 2021-12-17
fix(test runner): test.setTimeout should not break debugging

dgozman:
We ignore test.setTimeout() when timeout is already zero for debugging.


yury-s (approved):

yury-s (commented):
Consider using an explicit flag for the debug modes.

dgozman:
> Consider using an explicit flag for the debug modes.

Yep, this is in my TODO list, as soon as we understand our debugging story :)

PR #11006, opened 2021-12-18
chore: add browser like UA to browser fetcher

mxschmitt:
Drive-by: unify all Playwright user agents across the board.

### DIFF
--- a/packages/playwright-core/src/utils/browserFetcher.ts
+++ b/packages/playwright-core/src/utils/browserFetcher.ts
@@ -20,9 +20,8 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 import { execSync } from 'child_process';
-import { existsAsync, download, getPlaywrightVersion } from './utils';
+import { existsAsync, download, getPlaywrightVersion, getUserAgent } from './utils';
 import { debugLogger } from './debugLogger';
-import { parseOSReleaseText } from './ubuntuVersion';
 
 export async function downloadBrowserWithProgressBar(title: string, browserDirectory: string, executablePath: string, downloadURL: string, downloadFileName: string): Promise<boolean> {
   const progressBarName = `Playwright build of ${title}`;
@@ -38,7 +37,7 @@ export async function downloadBrowserWithProgressBar(title: string, browserDirec
     await download(url, zipPath, {
       progressBarName,
       log: debugLogger.log.bind(debugLogger, 'install'),
-      userAgent: getDownloadUserAgent(),
+      userAgent: getUserAgent(),
     });
     debugLogger.log('install', `extracting archive`);
     debugLogger.log('install', `-- zip: ${zipPath}`);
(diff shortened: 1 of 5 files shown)


aslushnikov (approved):
lgtm % pavel's lgtm

packages/playwright-core/src/utils/browserFetcher.ts
```diff
@@ -63,3 +66,63 @@ export function logPolitely(toBeLogged: string) {
...
+  try {
+    const { identifier: osIdentifier, version: osVersion } = getOSInfo();
+    const userAgentTokens = [
+      getLanguage(),
+      osIdentifier,
+      osVersion,
+      os.arch(),
+      getAsBooleanFromENV('CI') ? '1' : '0',
+    ];
+    return `Playwright/${getPlaywrightVersion()} (${userAgentTokens.join('; ')})`;
```
pavelfeldman, line 88: Let's follow the established UA practice and include Platform information (Node, Python, Java, .Net version)
```
Playwright/1.17.2 (amd64; macOS 10.15.7) Node/1.16.8
Playwright/1.17.2 (aarch64; macOS 12.0.1) Python/3.9
```

Also make sure params are easy to understand (you are using 1 for CI). Something like this:
```
Playwright/1.17.2 (amd64; macOS 10.15.7) Node/1.16.8 CI/1 (GitHubAction)
```

Otherwise in your user agent, `CI` is hard to decipher - I'm just getting `;1` and that means `CI`.

aslushnikov:
With the new patch version, we get user agents like this:
- `Playwright/1.17.2 (aarch64; macOS 12.0.1) Python/3.9`
- `Playwright/1.17.2 (x86_64; windows 10.0) node/16.3`
- `Playwright/1.17.2 (x86_64; linux unknown) Java/unknown`

There's some work required to get lang version in language ports.

packages/playwright-core/src/utils/browserFetcher.ts
```diff
@@ -63,3 +66,45 @@ export function logPolitely(toBeLogged: string) {
...
+  if (process.platform === 'win32') {
+    const version = os.release().split('.');
+    osIdentifier = 'windows';
+    osVersion = `${version[0]}.${version[1]}`;
+  } else if (process.platform === 'darwin') {
+    const version = execSync('sw_vers -productVersion').toString().trim().split('.');
+    osIdentifier = 'macOS';
+    osVersion = `${version[0]}.${version[1]}`;
+  } else if (process.platform === 'linux') {
+    const version = getUbuntuVersionSync();
```
pavelfeldman, line 94: I would generalize it to reading NAME + VERSION off /etc/os-release. That way we get all vendors. Not sure why getUbuntuVersionSync is doing more than that btw, /etc/os-release has all the info we need. While looking at it, I thought that maybe putting OS + architecture as a Playwright property is not what we want, maybe we want

Playwright/1.18 Ubuntu/20.04 (cpu=aarch)

packages/playwright-core/src/utils/browserFetcher.ts
pavelfeldman: Thinking out loud: Playwright/1.17.2 Python/3.9 macOS/12.0.1 (aarch64)

packages/playwright-core/src/utils/browserFetcher.ts
aslushnikov: > I would generalize it to reading NAME + VERSION off /etc/os-release. That way we get all vendors. 

Turns out to be mostly true! https://gist.github.com/aslushnikov/8ceddb8288e4cf9db3039c02e0f4fb75

packages/playwright-core/src/utils/utils.ts
```diff
@@ -445,8 +459,50 @@ export function canAccessFile(file: string) {
...
+    const version = execSync('sw_vers -productVersion').toString().trim().split('.');
+    osIdentifier = 'macOS';
+    osVersion = `${version[0]}.${version[1]}`;
+  } else if (process.platform === 'linux') {
+    try {
+      const osReleaseText = fs.readFileSync('/etc/os-release', 'utf8');
+      const fields = parseOSReleaseText(osReleaseText);
+      osIdentifier = fields.get('id') || 'unknown';
+      osVersion = fields.get('version_id') || 'unknown';
+    } catch (e) {
```
mxschmitt, line 493: Maybe set to "unknown-linux" in the catch?

packages/playwright-core/src/utils/utils.ts
```diff
@@ -445,8 +459,54 @@ export function canAccessFile(file: string) {
...
+    return cachedUserAgent;
+  try {
+    cachedUserAgent = determineDownloadUserAgent();
+  } catch (e) {
+    cachedUserAgent = 'Playwright/unknown';
+  }
+  return cachedUserAgent;
+}
+
+function determineDownloadUserAgent(): string {
```
dgozman, line 474: nit: rename to `determineUserAgent`?

tests/global-fetch.spec.ts
```diff
@@ -177,13 +178,23 @@ it('should resolve url relative to gobal baseURL option', async ({ playwright, s
...
     request.get(server.EMPTY_PAGE)
   ]);
-  expect(serverRequest.headers['user-agent']).toBe('Playwright/' + getPlaywrightVersion());
+  const userAgent = serverRequest.headers['user-agent'];
+  expect(userAgent).toContain('Playwright/' + getPlaywrightVersion());
+  expect(userAgent).toContain(os.arch());
+  const NODEJS_VERSION = process.version.substring(1).split('.').slice(0, 2).join('.');
+  expect(userAgent).toContain('node/' + NODEJS_VERSION);
+  if (isWindows)
+    expect(userAgent).toContain('windows');
```
dgozman, line 193:
```js
expect(userAgent.replace(/\d+/g, 'X').toBe(`Playwright/X.X.X (${os.arch()}; windows X.X) node X.X`);
```

dgozman (approved):

Issue #11031, opened 2021-12-20
[infra] build webkit on MacOS 12


aslushnikov:

unlikelyzero:
@aslushnikov FYI this _should_ solve the webkit "bug" with basic auth and websockets according to the safari dev team.

https://github.com/microsoft/playwright/issues/10417

aslushnikov:
This is fixed now; should be available in canary tomorrow.

PR #11066, opened 2021-12-22
devops: download new MacOS 12 WebKit builds

aslushnikov:
Drive-by: simplify management of executable paths since they are
the same for each platform.

References #11031

### DIFF
--- a/packages/playwright-core/src/utils/registry.ts
+++ b/packages/playwright-core/src/utils/registry.ts
@@ -30,52 +30,24 @@ const BIN_PATH = path.join(__dirname, '..', '..', 'bin');
 
 const EXECUTABLE_PATHS = {
   'chromium': {
-    'ubuntu18.04': ['chrome-linux', 'chrome'],
-    'ubuntu20.04': ['chrome-linux', 'chrome'],
-    'ubuntu18.04-arm64': ['chrome-linux', 'chrome'],
-    'ubuntu20.04-arm64': ['chrome-linux', 'chrome'],
-    'mac10.13': ['chrome-mac', 'Chromium.app', 'Contents', 'MacOS', 'Chromium'],
-    'mac10.14': ['chrome-mac', 'Chromium.app', 'Contents', 'MacOS', 'Chromium'],
-    'mac10.15': ['chrome-mac', 'Chromium.app', 'Contents', 'MacOS', 'Chromium'],
-    'mac11': ['chrome-mac', 'Chromium.app', 'Contents', 'MacOS', 'Chromium'],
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/utils/registry.ts
```diff
@@ -114,6 +90,8 @@ const DOWNLOAD_PATHS = {
     'mac10.15': 'builds/firefox/%s/firefox-mac-11.zip',
     'mac11': 'builds/firefox/%s/firefox-mac-11.zip',
     'mac11-arm64': 'builds/firefox/%s/firefox-mac-11-arm64.zip',
+    'mac12': 'builds/firefox/%s/firefox-mac-11.zip',
```
dgozman, line 93: Just to double check - we reuse 11 build for 12?

packages/playwright-core/src/utils/registry.ts
```diff
@@ -30,52 +30,24 @@ const BIN_PATH = path.join(__dirname, '..', '..', 'bin');
...
-    'mac11-arm64': ['pw_run.sh'],
-    'win64': ['Playwright.exe'],
+    'linux': ['pw_run.sh'],
+    'mac': ['pw_run.sh'],
+    'win': ['Playwright.exe'],
   },
   'ffmpeg': {
-    'ubuntu18.04': ['ffmpeg-linux'],
-    'ubuntu20.04': ['ffmpeg-linux'],
-    'ubuntu18.04-arm64': ['ffmpeg-linux-arm64'],
```
dgozman, line 71: Seems like we lost this one?

packages/playwright-core/src/utils/registry.ts
```diff
@@ -126,6 +104,8 @@ const DOWNLOAD_PATHS = {
     'mac10.15': 'builds/firefox-beta/%s/firefox-beta-mac-11.zip',
     'mac11': 'builds/firefox-beta/%s/firefox-beta-mac-11.zip',
     'mac11-arm64': 'builds/firefox-beta/%s/firefox-beta-mac-11-arm64.zip',
+    'mac12': 'builds/firefox-beta/%s/firefox-beta-mac-11.zip',
```
yury-s, line 107: Should be -mac12.zip right?

aslushnikov: for everything but webkit we re-use macos 11 builds

packages/playwright-core/src/utils/registry.ts
aslushnikov: :'( good catch. Let me change this to `ffmpeg-linux` for consistency.

packages/playwright-core/src/utils/registry.ts
aslushnikov: yes we do

PR #11081, opened 2021-12-23
feat(ffmpeg): roll ffmpeg to 1007

aslushnikov:
This build:
- Includes a native build for Mac Arm64
- Changes executable name for the linux arm64 build

Drive-by: start running docker tests as part of CQ1.


mxschmitt:
Would be cool if we can fix the other workflow too, so we don't need to open/close it for new rolls

PR #11131, opened 2021-12-29
[Draft] Check accessible name when calculating enabled/disabled

### LINKED ISSUE #10562: [Feature] Option to timeout if target is missing accessibility name
## Summary

I'd like an option to extend [actionability checks](https://playwright.dev/docs/actionability/) to check if the target has an [accessible name](https://www.w3.org/TR/accname-1.1/), so that my tests will time out if they are not accessible to assistive technology.

## Background

Actionability checks assert that a number of conditions are true about an element before attempting to fire interactions. For example, if the element is disabled or obscured, the test will timeout on actionability checks. I propose and optional actionability check to assert that the control has an accessible name (and potentially also interactive role), so that test will fail if the control is not accessible to screen readers and other assistive technology.

Accessible name and role are synthesized by the browser into an accessibility tree, and normally (with the exception of [experimental AOM](https://wicg.github.io/aom/caniuse.html) in Chrome) aren't available to JS. However, there's work to recreate the browser accessible name and role computation in JS, e.g. https://github.com/eps1lon/dom-accessibility-api, that could be used here.

### Current Implementation
```javascript
const browser = await firefox.launch();
await page.setContent(`<div><input placeholder='First Name' /></div>`); // no name exposed to a11y (missing a label)
await page.fill('input', 'hello'); // works
```

With this proposed feature enabled, this example would timeout on the last line:

```javascript
const browser = await firefox.launch();
const page = await browser.newPage({ actionability: { a11y: true }); // turn on the actionability step
await page.setContent(`<div><input placeholder='First Name' /></div>`); // no name exposed to a11y (missing a label)
await page.fill('input', 'hello'); // Times out
```

I'm proposing extending the actionability checks to assert on the accessible name and roles of the element using something like https://github.com/eps1lon/dom-accessibility-api, so that the above example would timeout, prompting me to update the markup since it doesn't work for screen reader users, dictation technology, and so on:

```javascript
const browser = await firefox.launch();
const page = await browser.newPage({ actionability: { a11y: true });
await page.setContent(`<div><label>First Name <input placeholder='John Doe' /></label></div>`); // fixed! has an accessible name "First Name"
await page.fill('input', 'hello'); // works
```

Or using `force`:

```javascript
const browser = await firefox.launch();
const page = await browser.newPage({ actionability: { a11y: true }); // turn on the actionability step
await page.setContent(`<div><input placeholder='First Name' /></div>`); // no name exposed to a11y (missing a label)
await page.fill('input', 'hello', { force: true }); // works
```

## Why?

Enabling this feature would allow me to quickly and easily surface basic but critical accessibility failures in my existing suite, without needing to load external libraries or helpers. In fact, **I wouldn't change my test examples at all**. It would instead require a page or browser level configuration option to enable accessibility in actionability checks. This low touch approach would be a big win for massive codebases trying to shake out accessibility issues in their examples.

## Approach

1. Add an option `a11y: boolean;` to [FrameClickOptions](https://github.com/microsoft/playwright/blob/d70e37de80f99c3fe953921abd5ad7ba1a7f1da8/packages/playwright-core/src/protocol/channels.ts#L1868)
1. When the option is passed in as true from `page.click('#my-selector', { a11y: true })`, run an additional check in the timeout checks.
1.  Add support for global configuration option to default `a11y: true`, to make it unnecessary to change the test files at all to get the expected timeouts.

What do y'all think?

ckundo:
- When a control is missing an accessible name, it is effectively missing for assistive technology and therefore in-actionable.
- Developers can use `force` to override.
- Draft / PoC pending issue triage

Fixes #10562

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3222,6 +3222,11 @@
         "node": ">=6.0.0"
       }
     },
+    "node_modules/dom-accessibility-api": {
+      "version": "0.5.10",
+      "resolved": "https://registry.npmjs.org/dom-accessibility-api/-/dom-accessibility-api-0.5.10.tgz",
+      "integrity": "sha512-Xu9mD0UjrJisTmv7lmVSDMagQcU9R5hwAbxsaAE/35XPnPLJobbuREfV/rraiSaEj/UOvgrzQs66zyTWTlyd+g=="
+    },
     "node_modules/dom-converter": {
       "version": "0.2.0",
       "resolved": "https://registry.npmjs.org/dom-converter/-/dom-converter-0.2.0.tgz",
@@ -9084,6 +9089,7 @@
       "dependencies": {
         "commander": "^8.2.0",
         "debug": "^4.1.1",
+        "dom-accessibility-api": "^0.5.10",
         "extract-zip": "^2.0.1",
         "https-proxy-agent": "^5.0.0",
         "jpeg-js": "^0.4.2",
@@ -11684,6 +11690,11 @@
       "requires": {
         "esutils": "^2.0.2"
       }
+    },
+    "dom-accessibility-api": {
+      "version": "0.5.10",
+      "resolved": "https://registry.npmjs.org/dom-accessibility-api/-/dom-accessibility-api-0.5.10.tgz",
+      "integrity": "sha512-Xu9mD0UjrJisTmv7lmVSDMagQcU9R5hwAbxsaAE/35XPnPLJobbuREfV/rraiSaEj/UOvgrzQs66zyTWTlyd+g=="
     },
     "dom-converter": {
       "version": "0.2.0",
@@ -14420,6 +14431,7 @@
       "requires": {
         "commander": "^8.2.0",
         "debug": "^4.1.1",
+        "dom-accessibility-api": "^0.5.10",
         "extract-zip": "^2.0.1",
         "https-proxy-agent": "^5.0.0",
         "jpeg-js": "^0.4.2",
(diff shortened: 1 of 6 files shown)


WestonThayer:
I like the approach of putting under the enabled check! There is a very tiny risk of false negatives with using `dom-accessibility-api` for any case where a browser has decided to go off-spec with their accname calculations. Their deviation would have to be drastic, choosing to not surface a name when they spec says they should. I'm not aware of any situations where this actually happens intentionally, but there have been browser-specific bugs.

That said, Playwright does have the ability to directly ask the browser for a node's calculated accname via [accessibility.snapshot({ root: element })](https://playwright.dev/docs/api/class-accessibility#accessibility-snapshot). Might be worth replacing `dom-accessibility-api` with it?

`accessibility.snapshot()` does retrieve the element's children, which could maybe cause a perf issue? If that ends up being the case, it does seem fixable. CDP has [getPartialAXTree](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#method-getPartialAXTree), where `fetchRelatives` can be set to `false`. Playwright's [Webkit patch](https://github.com/microsoft/playwright/blob/fba29aa31942e81ebf7c70ef53eda97aeab24017/browser_patches/webkit/patches/bootstrap.diff#L4826) can be modified.

ckundo:
> That said, Playwright does have the ability to directly ask the browser for a node's calculated accname via [accessibility.snapshot({ root: element })](https://playwright.dev/docs/api/class-accessibility#accessibility-snapshot). Might be worth replacing `dom-accessibility-api` with it?

I like this suggestion... I assumed anecdotally that getting the browser a11y tree would be more expensive than a pure JS calculation. If that's not true then that'd be preferable, I agree. Still, this patch adds incremental value with a small diff, which is nice.

PR #11115, opened 2021-12-28
fix(html reporter): fix too much strikethrough in diffs

rwoll:
Textual snapshot diffs were previously broken in the HTML Report. The strikethrough'd text extended beyond the intended region.

HTML Report Before: 
Screen Shot 2021-12-27 at 4 43 35 PM

HTML Report After:
Screen Shot 2021-12-27 at 4 48 37 PM

This now matches what's expected and shown in the terminal (which has always been correct):

Screen Shot 2021-12-27 at 4 36 29 PM

NB: This MR is a workaround, but not a root cause fix. It works, but I never fully got to the root cause so a bug upstream may be required. It's unclear whether it's (1) in [`colors`](https://www.npmjs.com/package/colors), (2) in [`ansi-to-html`](https://www.npmjs.com/package/ansi-to-html), or (3) Playwright's use of the two. Since the terminal output is correct, I suspect it is in `ansi-to-html`. For example:

```js
const colors = require("colors");
const Convert = require('ansi-to-html');
const convert = new Convert();

// original (strike incorrectly wraps everything in the HTML)
console.log(convert.toHtml(colors.strikethrough("crossed out") + ' ' + colors.red("red")))
// prints: <strike>crossed out <span style="color:#A00">red<span style="color:#FFF"></span></span></strike>

// workaround
console.log(convert.toHtml(colors.reset(colors.strikethrough("crossed out")) + ' ' + colors.red("red")))
// prints: <strike>crossed out</strike> <span style="color:#A00">red<span style="color:#FFF"></span></span>
```

Fixes #11116

PR #11132, opened 2021-12-30
feat(test runner): add `snapshotSuffix` config

### LINKED ISSUE #11134: [Feature] API Testing: Easy Opt-Out of Platform Name in Snapshots
We use Playwright Test for API testing and make heavy use of snapshots on textual data that's been normalized from JSON API responses.

Currently, Playwright Test will embed the platform name by default in the snapshot, but this doesn't make sense for API Testing.

I stumbled a bit trying to find out how to disable it. (I saw `snapshotDir` could be overriden in the playwright.config.ts, but not `snapshotSuffix`.)

Ultimately, I figured out to use a `beforeEach` hook and/or an `auto: true` fixture. However, it would be nice to have an easy/ergonomic way to opt out of the platform-embedding behavior globally.

rwoll:
Adds `snapshotSuffix` to be configured in `playwright.config.ts` overall and per-project. Also adds test cases for currently documented but untested behavior (like platform and/or project being present in snapshot names).

**Context**:

We're using Playwright Test for API testing. Under this use case, all of our snapshots are textual (normalized JSON, YAML, etc.), so the default behavior of having the platform embedded in the snapshot name does not make sense. I went to disable it in the config, but realized there does not exist an option currently, although there does exist a top-level option for `snapshotDir`.

We worked around this with a `beforeEach` hook, then a more robust `auto` fixture; however, it felt awkward compared to a top-level config like there are for related behavior/configs (`snapshotDir`).

Resolves #11134
Relates #8912
Relates #11115

### DIFF
--- a/docs/src/test-snapshots-js.md
+++ b/docs/src/test-snapshots-js.md
@@ -41,6 +41,10 @@ drwxr-xr-x  3 user  group   96 Jun  4 11:46 example.spec.ts-snapshots
 
 Note the `chromium-darwin` in the generated snapshot file name - it contains the browser name and the platform. Screenshots differ between browsers and platforms due to different rendering, fonts and more, so you will need different snapshots for them. If you use multiple projects in your [configuration file](./test-configuration.md), project name will be used instead of `chromium`.
 
+:::note
+If you're snapshotting platform-independent data, like JSON or YAML from an API response, you can opt out of embedding the platform name (e.g. `linux`, `darwin`, etc.). See [Snapshot Suffix Overrides](#snapshot-suffix-overrides) for more details.
+:::
+
 If you are not on the same operating system as your CI system, you can use Docker to generate/update the screenshots:
 
 ```bash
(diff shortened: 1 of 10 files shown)


tests/playwright-test/golden.spec.ts
```diff
@@ -790,3 +790,112 @@ test('should allow comparing text with text without file extension', async ({ ru
...
+        const test = pwt.test;
+        test('is a test', ({}) => {
+          expect('Hello world').toMatchSnapshot('example.txt');
+        });
+      `
+    });
+    expect(result.exitCode).toBe(0);
+  });
+
+  test('includes project name', async ({ runInlineTest }) => {
```
rwoll, line 808: This behavior isn't new, but I didn't see an existing test for it.

tests/playwright-test/golden.spec.ts
```diff
@@ -790,3 +790,112 @@ test('should allow comparing text with text without file extension', async ({ ru
   });
   expect(result.exitCode).toBe(0);
 });
+
+test.describe('snapshotSuffix', () => {
+  test('default to platform', async ({ runInlineTest }) => {
```
rwoll, line 795: This behavior isn't new, but I didn't see an existing test for it.

docs/src/test-snapshots-js.md
```diff
@@ -125,3 +129,70 @@ test('example test', async ({ page }) => {
 ```
 
 Snapshots are stored next to the test file, in a separate directory. For example, `my.spec.ts` file will produce and store snapshots in the `my.spec.ts-snapshots` directory. You should commit this directory to your version control (e.g. `git`), and review any changes to it.
+
+
+## Snapshot Suffix Overrides
```
rwoll, line 134: I couldn't get the docs to work locally (seem to be missing a docusaurus config?), so I have no clue if all the syntax in the docs is correct and working as intended. 😄 

We're using Plawyright Test more and more for API testing. It seems like others (based on some Slack convos) are also interested in this and/or doing it already. In the future, a dedicated preset-config and/or docs page with tips-and-tricks (which go beyond this setting) may be nice. Happy to contribute it once I have a better sense of what is and isn't working for us in the way we build our API tests.

pavelfeldman:
Thanks for the PR, Ross! As we are focusing on the visual regression testing, we are revisiting the strategies we use for snapshot organization and storage. Let's keep this PR open as we are considering the overall picture. Just setting expectations here - it might take more time to land than usually!

rwoll:
Closing as this will be resolve with other VRT work. cc: @aslushnikov

Issue #9217, opened 2021-09-29
[Feature] Expose browser/platform info through reporter API


FriggaHel:
Hi !

I am working currently working on a reporter for playwright and want to integrate browser & platform informations as part of the report. Currently, only configuration and tests related info are available.

Would it be possible to provide browser and platform info as:
- Playwright version
- Platform/OS name and version
- Browser name and version (that are already available out of the reporter through _**Browser**_ and **_BrowserType_** types)

Thanks !

dgozman:
@FriggaHel When you run multiple projects, you might have tests that run in different browsers. Do you need this information for each test?

Note that you can integrate platform as easy as `process.platform`, and get the browser name from `config.projects[i].use.browserName`. Note that multiple projects may have different browser names, so it is up to you how to surface this.

If you could share more details, like a config + test example and what you'd like to see in your report, we'll be able to help better!

FriggaHel:
Hello @dgozman !

**_A bit of context:_**

I'm working on a reporter that will push tests results into Sauce Labs. 
A job representation (in Sauce Labs) contains tuple of OS+Browser, so in case you have multiple browsers it will results in different jobs from SL perspective. 

**_Regarding platform:_**

That was my plan-B approach to rely on `process.platform`, I was wondering if it wasn't better to have a value directly provided by Playwright (I agree this one is easy to address).

**_Regarding browser:_**

Currently, there is no access to project configuration from `onTestEnd` hook, only the project name.

Only way to make coincide project configuration + test is to rely on project name from `TestCase` object , and find the correct on from the list received in `onBegin`. Knowing. that the project name may not be unique / defined in `playwright.config.js` there is no reliable way to find the matching configuration.

Also, if `browserName` is not provided though project config or cli-parameter, tests are still executed, and there is no way to know which browser was used from within the reporter.

Next point is `browserVersion`. Through project config we may select to use `chrome-beta` that may result in using an already installed version of Chrome Beta. From the reporter perspective, there is no clue to know what version is installed, just that this is a beta one. Eventually, it can be considered to find a way to resolve it from the reporter, but knowing that playwright is likely to know which version it is, it would be more convenient to gather the version from Playwright API directly.

**_Final thoughts:_**

In my opinion, the main pain point, is not to have the config attached to the `project suite`[[1](https://playwright.dev/docs/api/class-suite/#suite-suites)] and/or not enforcing unique names for project, so it can be access differently.
[...]

PR #11159, opened 2022-01-03
feat(test): annotate tests with browser version

dgozman:
Example:

Screen Shot 2022-01-03 at 3 54 39 PM

References #9217.


packages/playwright-test/src/index.ts
```diff
@@ -188,6 +188,9 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
     const createdContexts = new Set<BrowserContext>();
 
     const onDidCreateContext = async (context: BrowserContext) => {
+      const browser = context.browser();
+      if (browser && !testInfo.annotations.find(a => a.type === 'browserVersion'))
+        testInfo.annotations.push({ type: 'browserVersion', description: browser.version() });
```
pavelfeldman, line 193: That's going to spam our HTML report a lot :/

PR #11188, opened 2022-01-05
fix(html reporter): skip img bits with non-images

### LINKED ISSUE #11183: [BUG] HTML Reporter: Image Mismatch Section Shows Even on non-Image Types
**Context:**
- Playwright Version: ToT, 1.17.1

The HTML Reporter always shows an Image Mismatch section, even if the snapshot is not a common image format:

Screen Shot 2022-01-04 at 5 08 33 PM

In this example, our use case is API testing, so we are not doing visual regressions.

**Would it be OK for me to suppress this block in the reporter if the snapshot is not a common image format?**

I'd imagine we could infer the content type based on (1) sniffed mimetype from the contents, (2) the filename extension, and/or (3) an explicit (new) contentType option similar to what we have in https://github.com/microsoft/playwright/pull/10121/files. 

For backwards compatibility (if we don't have (1)), we might have to inference in some cases and only NOT show it if the name explicitly has a textual extension (e.g. `.json`, `.yml`) or explicit (new) non-image `contentType` option. For instance, if a user is currently supplying their image snapshot name as `foo` (no extension), we wouldn't want to break the reporter so we would always show the Image mismatch block in this case.

(I know there is other work/conversations regarding snapshot testing, and I'm happy to wait on implementing this to see if it makes sense with whatever other changes land. Thanks!)

rwoll:
Regarding backwards compatibility: The image block will no longer show
up in cases where no extension is used (see test case), but I think this
is OK since you would have never hit the actual image comparison in the
test if you omitted an extension: https://github.com/microsoft/playwright/blob/328f3e265efe9c431ebae1ed3d7fd47dcd547b29/packages/playwright-test/src/matchers/golden.ts#L44 before.

Depending on how Visual Testing evolves, a follow up PR is due to clean
up references to images.

Fixes #11183

### DIFF
--- a/packages/html-reporter/src/testResultView.tsx
+++ b/packages/html-reporter/src/testResultView.tsx
@@ -49,6 +49,7 @@ export const TestResultView: React.FC<{
   const expected = attachmentsMap.get('expected');
   const actual = attachmentsMap.get('actual');
   const diff = attachmentsMap.get('diff');
+  const hasImages = [actual?.contentType, expected?.contentType, diff?.contentType].some(v => v && /^image\//i.test(v));
   return <div className='test-result'>
     {result.error && <AutoChip header='Errors'>
       <ErrorMessage key='test-result-error-message' error={result.error}></ErrorMessage>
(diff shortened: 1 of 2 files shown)


packages/html-reporter/src/testResultView.tsx
```diff
@@ -57,8 +58,8 @@ export const TestResultView: React.FC<{
       {result.steps.map((step, i) => <StepTreeItem key={`step-${i}`} step={step} depth={0}></StepTreeItem>)}
     </AutoChip>}
 
-    {expected && actual && <AutoChip header='Image mismatch'>
-      <ImageDiff actual={actual} expected={expected} diff={diff}></ImageDiff>
+    {expected && actual && <AutoChip header={`${hasImages ? 'Image' : 'Snapshot'} mismatch`}>
```
rwoll, line 61: We leave this block since it is convenient to be able to download the expected/actual/diff attachments regardless of their content type.

pavelfeldman (approved):

PR #11212, opened 2022-01-06
fix(selectors): properly determine visibility of display:contents

### LINKED ISSUE #11202: [BUG] Elements with display:contents are considered invisible
**Context:**
- Playwright Version: 1.17.1
- Operating System: Linux 
- Node.js version: 14.18.1
- Browser: Chromium

**Code Snippet**
I have a repro: https://github.com/basilgood/pw-repro

**Describe the bug**
A checkbox, when is checked a text (paragraph) appear in the footer with text ```Checkbox is CHECKED!```
The footer it's a parent selector with ```display: none``` style.
The paragraph is the child of the footer and has ```display: contents```

Test: ```npx playwright test --headed```
```js
  await page.goto('http://localhost:3000/');
  await page.click('#myCheck');
  await page.waitForSelector('"Checkbox is CHECKED!"');
```
It works well, the footer with text appear so the text is visible in page.
The bug it's that waitForSelector fails to find it.
```
waiting for selector ""Checkbox is CHECKED!"" to be visible
  selector resolved to hidden <p>Checkbox is CHECKED!</p>
```
The same behavior is if I change ```waitForSelector``` with ```click```
Surprisingly is that the test pass with:
```js
await page.waitForSelector('"Checkbox is CHECKED!"', { state: 'hidden' });
```
```waitForSelector``` behave like the text (selector) it's hidden and find it.
It also pass with ```{ state: "attached" }``` , that makes me think of: #5949

### PR DESCRIPTION
**Pull request by dgozman**
Fixes #11202.

### DIFF
--- a/packages/playwright-core/src/server/injected/selectorEvaluator.ts
+++ b/packages/playwright-core/src/server/injected/selectorEvaluator.ts
@@ -660,7 +660,25 @@ export function isVisible(element: Element): boolean {
   const style = element.ownerDocument.defaultView.getComputedStyle(element);
   if (!style || style.visibility === 'hidden')
     return false;
+  if (style.display === 'contents') {
+    // display:contents is not rendered itself, but its child nodes are.
+    for (let child = element.firstChild; child; child = child.nextSibling) {
+      if (child.nodeType === 1 /* Node.ELEMENT_NODE */ && isVisible(child as Element))
+        return true;
+      if (child.nodeType === 3 /* Node.TEXT_NODE */ && isVisibleTextNode(child as Text))
+        return true;
+    }
+    return false;
+  }
   const rect = element.getBoundingClientRect();
+  return rect.width > 0 && rect.height > 0;
+}
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/playwright-core/src/server/injected/selectorEvaluator.ts [hunk 1]
```diff
@@ -660,10 +660,28 @@ export function isVisible(element: Element): boolean {
   const style = element.ownerDocument.defaultView.getComputedStyle(element);
   if (!style || style.visibility === 'hidden')
     return false;
+  if (style.display === 'contents') {
+    // display:contents is not rendered itself, but its child nodes are.
+    for (let child = element.firstChild; child; child = child.nextSibling) {
+      if (child.nodeType === 1 /* Node.ELEMENT_NODE */ && isVisible(child as Element))
```
pavelfeldman, line 666: Inline Node.ELEMENT_NODE, etc?

pavelfeldman (approved):

PR #11236, opened 2022-01-07
docs: fix colorized errors in GitLab CI

rwoll:
Without this, HTML Reports (which we make available via S3 in Slack alerts) were missing the pretty green/red colorization on `expect` failures/errors that you can see here (e.g. the red `Snapshot comparison failed`):

Screen Shot 2022-01-06 at 5 42 50 PM

Turns out, the colors weren't being outputted in GitLab CI logs either. For the snapshot diffs, the colorization is critical as it shows what's added/removed.

### DIFF
--- a/docs/src/ci.md
+++ b/docs/src/ci.md
@@ -334,6 +334,8 @@ stages:
 tests:
   stage: test
   image: mcr.microsoft.com/playwright:focal
+  variables:
+    FORCE_COLOR: 1 # ensure pretty, colorized errors from `expect` (courtesy of https://www.npmjs.com/package/colors)
   script:
   ...
 ```
--- a/docs/src/ci.md
+++ b/docs/src/ci.md
@@ -335,7 +335,7 @@ tests:
   stage: test
   image: mcr.microsoft.com/playwright:focal
   variables:
-    FORCE_COLOR: 1 # ensure pretty, colorized errors from `expect` (courtesy of https://www.npmjs.com/package/colors)
+    FORCE_COLOR: 1 # Ensure colorized errors appear in all reports like the HTML Reporter. (Follow https://github.com/microsoft/playwright/issues/11314 to be notified of when this is no longer necessary.)
   script:
   ...
 ```


yury-s (approved):

pavelfeldman (changes requested):
I don't think you should be fixing it here. HTML report should not respect TTY and always produce color information.

rwoll:
> I don't think you should be fixing it here. HTML report should not respect TTY and always produce color information.

Agreed. Looking into it quickly, looks like it gets a bit tricky as these colors are coming directly from global use of `colors` and baked/absent directly into the error message expect throws (https://github.com/microsoft/playwright/blob/a26646a811cd59a48366bee8dc7fd3d41a61afff/packages/playwright-test/src/matchers/golden.ts#L94).

Will have to think about the path forward. Perhaps we globally force colors and then just run them through strip wherever appropriate in the various reporters and stdout/err streams. Without this or changing our use of colors, I think we lose the color information before we have a chance to keep it or not. @pavelfeldman let me know if you have any initial thoughts or if I'm missing something major, but looks like currently HTML Report colors for this case are tied with  the other reporters/system (unlike  codeframes which have fine grained control over highlightCode which we might end up being able to use as an indication for when to strip or not).

pavelfeldman:
@rwoll: pretty much, but it should not be too hard:
- All the expect / golden matchers run in the worker processes. So we can unconditionally FORCE_COLOR: '1' in dispatcher.ts as we spawn those to make sure all the information is generated.
- We can then introduce Reporter::supportsColors?() and based on the return value, strip the ascii information from the errors.
- Terminal-based reporters will return TTY there, HTML reporter will return true there and all the third party reporters won't have it implemented => will keep existing behavior.

rwoll:
Thanks @pavelfeldman! 

I have a busier week than usual, but I'll take a pass at it next week!

pavelfeldman:
@dgozman is suggesting to go w/o opt in, considering that all reporters should have been exposed to operation in TTY already.

rwoll:
Closing this since it looks like it's fixed on @next (courtesy of https://github.com/microsoft/playwright/pull/11799).

Issue #11126, opened 2021-12-29
[BUG] Test Reporter Gets Confused on File Name


rwoll:
**Context:**
- Playwright Version: 1.17 and ToT
- Operating System: macOS
 - Node: 14.17.6 - ~/.nvm/versions/node/v14.17.6/bin/node
 - Yarn: 1.22.11 - ~/.nvm/versions/node/v14.17.6/bin/yarn
 - npm: 7.24.1 - ~/.nvm/versions/node/v14.17.6/bin/npm

**Code Snippet**

See the following for a full description and repro: https://github.com/rwoll/playwright-parameterized-tests-repro/blob/main/README.md 

**Describe the bug**

We define a test parametizer/helper `parameterizeTestCase` and call it in various other test files to get shared behavior.

(In reality the actual `parameterizeTestCase` has a much more extensive implementation to get templated test cases for API testing.)

If you run this sample, you will see the following output:

## Actual

```
$ npm run test:e2e

> plawyright-parameterized-tests-repro@1.0.0 test:e2e
> playwright test


Running 3 tests using 3 workers

  ✓  helpers/index.ts:4:3 › a (6ms)
  ✓  helpers/index.ts:4:3 › suite › b-0 (5ms)
[ 'e2e/a.spec.ts', 'a' ]
  ✓  helpers/index.ts:4:3 › suite › b-1 (5ms)
[ 'e2e/b.spec.ts', 'suite', 'b-0' ]
[ 'e2e/b.spec.ts', 'suite', 'b-1' ]
```

## Expected

```
$ npm run test:e2e

> plawyright-parameterized-tests-repro@1.0.0 test:e2e
> playwright test


Running 3 tests using 3 workers

  ✓  e2e/a.spec.ts:?:? › a (6ms)
  ✓  e2e/b.spec.ts:?:? › suite › b-0 (5ms)
[ 'e2e/a.spec.ts', 'a' ]
  ✓  e2e/b.spec.ts:?:? › suite › b-1 (5ms)
[ 'e2e/b.spec.ts', 'suite', 'b-0' ]
[ 'e2e/b.spec.ts', 'suite', 'b-1' ]
```

I expected the reporter to not reference `helpers/index.ts` at all.

A few notes:
* The `testInfo.titlePath` looks correct in the actual output.
* It looks like the runner chose to run `a.spec.ts` and `b.spec.ts` in parallel (based on worker counts).
[...]

PR #11238, opened 2022-01-07
[experimental] feat(reporter): verbose codeframe stacks 

rwoll:
This is a WIP experiment to try to get more complete stack traces and code previews into the HTML Report. We use helpers and parameterized tests and find that sometimes the report does not provide enough context to quickly understand where exactly the code failed.

We run smoketests via Playwright Test, so it's important for us to quickly triage directly from the report without opening the source code.

Given the following code:

`base.ts`:
```ts
import { test as _test, expect } from "@playwright/test";

export const test = _test.extend<{ run: () => void }>({
  run: async ({}, use) => {
    const run = () => expect(1).toBe(2);
    await use(run);
  },
});
```

`example.spec.ts`:
```ts
import { test } from "./base";

test("a", async ({ run }) => {
  run();
});
```

**v1.17.1** will output:

Screen Shot 2022-01-06 at 10 00 56 PM

It shows which `test` (of potentially many) in a file failed, but the codeframe doesn't show us which `expect` call site failed (of which there can be many).

**v1.18.0-alpha-jan-7-2022** will outputs something a bit different (which we just picked up in a staging environment today):

Screen Shot 2022-01-06 at 10 03 21 PM

While the codeframe shows which specific `expect` call site failed, we no longer see which `test` context it was called in.

Neither paint a complete picture, so the PR is an experiment with a verbose stack codeframe where we get codeframes for each line item on the stack. I think this provides the best of both worlds 😄 

Here's the sample output for the placeholder test examples checked into this MR:

`base.ts`:
```ts
const { test: _test, expect } = pwt;
  helpers: {
    runParameterizedTest: (v: number) => Promise<void>;
    exampleFixtureThatThrows: (v: number) => Promise<void>;
  };
}>({
  helpers: async ({}, use) => {
    const helpers = {
      runParameterizedTest: async (v: number) => {
        expect(v).toBe(2);
      },
      exampleFixtureThatThrows: async (v: number) => {
        throw new Error("oops! this is a def broken fixture helper!");
      },
    };
    await use(helpers);
  },
});

export const parameterizeTest = async (v: number) => {
  await test.step("a", () => expect(1).toBe(3));
};
```

`a.spec.ts`:
```ts
import { test, parameterizeTest } from "./base.ts";

test("basic expect example", () => {
  expect(1).toBe(2);
});

test("parameterized test helper fixture expect error", async ({ helpers }) => {
  await helpers.runParameterizedTest(5);
});

test("example fixture that throws", async ({ helpers }) => {
  await helpers.exampleFixtureThatThrows(5);
});

test("non-fixture parameterizer example", async () => {
  await parameterizeTest(47);
});
```
**Before** this change:
Screen Shot 2022-01-06 at 10 17 29 PM

**After** this change:
Screen Shot 2022-01-06 at 10 15 25 PM
Screen Shot 2022-01-06 at 10 15 42 PM

The last screenshots, while verbose, provide us with all the context we need even if there are errors/expects triggered in any helper functions that out test consumes.

Perhaps to find a happy balance between details vs. conciseness, there is an option for the reporter that controls whether just the top stack line gets a code frame or all of them within the stack do (like this experiment proposes). If it doesn't make sense to amend the built in reporter, this could also be a supplemental, non-default official (or user-land reporter).

### DIFF
--- a/packages/playwright-test/src/reporters/base.ts
+++ b/packages/playwright-test/src/reporters/base.ts
@@ -346,27 +346,30 @@ export function formatError(config: FullConfig, error: TestError, highlightCode:
   const tokens = [''];
   let location: Location | undefined;
   if (stack) {
-    const parsed = prepareErrorStack(stack);
-    tokens.push(parsed.message);
-    location = parsed.location;
-    if (location) {
-      try {
-        // Stack will have /private/var/folders instead of /var/folders on Mac.
-        const realFile = fs.realpathSync(location.file);
-        const source = fs.readFileSync(realFile, 'utf8');
-        const codeFrame = codeFrameColumns(source, { start: location }, { highlightCode });
-        if (!file || file !== realFile) {
+    const stackLines = prepareErrorStack(stack);
+    for (let i = 0; i < stackLines.length; i++) {
+      const stackLine = stackLines[i];
+      if (i == 0) location = stackLine.location;
+      if (stackLine.message) tokens.push(stackLine.message);
+      if (stackLine.location) {
+        try {
+          // Stack will have /private/var/folders instead of /var/folders on Mac.
+          const realFile = fs.realpathSync(stackLine.location.file);
+          const source = fs.readFileSync(realFile, 'utf8');
+          const codeFrame = codeFrameColumns(source, { start: stackLine.location }, { highlightCode });
+          if (!file || file !== realFile) {
+            tokens.push('');
+            tokens.push(colors.gray(`   at `) + `${relativeFilePath(config, realFile)}:${stackLine.location.line}`);
+          }
           tokens.push('');
-          tokens.push(colors.gray(`   at `) + `${relativeFilePath(config, realFile)}:${location.line}`);
+          tokens.push(codeFrame);
+          tokens.push('');
+        } catch (e) {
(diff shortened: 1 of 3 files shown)


rwoll:
Please ignore the actual implementation for now. If the end behavior is desired, I can clean stuff up, add tests, etc. I wanted to draft this quickly to see what the output would feel like for some of our company's internal tests and reports.

rwoll:
cc: @dgozman Let me know if you like how the screenshots look and if you think this is a good evolution of the reporter/codeframes! (Dogfooded some of these in our smoketest reports and find the extra codeframes/context extremely useful in practice.)

This is sort of what I was thinking about with https://github.com/microsoft/playwright/issues/11126#issuecomment-1004442183. I'm not sure if the change in behavior/codeframe site of **v1.17.1** vs. **v1.18.0-alpha-jan-7-2022** was intentional, but the proposal in this MR would give us the best of both 😄 🎉 

Thanks!

rwoll:
Closing this as it was just a little experiment. An issue/feature request for now will be better to discuss and not clutter the PR screen. Filed here: https://github.com/microsoft/playwright/issues/11315

PR #11285, opened 2022-01-10
Add installing project dependencies / uploading test results

vthommeret:
It took me a bit of time to figure out an initial working GitHub Actions workflow (using the documented one as a starting point) and then I found a message about uploading test results in the Slack but felt nice if it would be documented / to save people time in the future.

I'm not a node expert but it seems like the example workflow doesn't include actually installing your project, although the Python example below it does? This might be obvious to others but I've added it as well as an example for uploading artifacts to GitHub for people that want to.

### DIFF
--- a/docs/src/ci.md
+++ b/docs/src/ci.md
@@ -51,8 +51,16 @@ steps:
       node-version: '14'
   - name: Install operating system dependencies
     run: npx playwright install-deps
+  - name: Install dependencies
+    run: npm install
   - name: Run your tests
     run: npm test
+  - name: Upload test results
+    if: always()
+    uses: actions/upload-artifact@v2
+    with:
+      name: playwright-results
+      path: test-results
 ```
 
 ```yml python
--- a/docs/src/ci.md
+++ b/docs/src/ci.md
@@ -49,10 +49,10 @@ steps:
   - uses: actions/setup-node@v2
     with:
       node-version: '14'
-  - name: Install operating system dependencies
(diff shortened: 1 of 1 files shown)


docs/src/ci.md
```diff
@@ -51,8 +51,16 @@ steps:
       node-version: '14'
   - name: Install operating system dependencies
     run: npx playwright install-deps
+  - name: Install dependencies
```
aslushnikov, line 54: this step should go before `Install operating system dependencies`: this way you'll use the Playwright binary version that your project depends on.

docs/src/ci.md
```diff
@@ -51,8 +51,16 @@ steps:
...
+    run: npm install
```
mxschmitt, line 55:
```suggestion
    run: npm ci
```
https://docs.npmjs.com/cli/v8/commands/npm-ci

vthommeret:
Thanks for feedback! I've flipped the order and also switched from `npm install` to `npm ci`.

One note I had to switch `npx playwright install-deps` to `npx playwright install --with-deps` after flipping the order.

Another thing I noticed is that the `GitHub Actions on deployment​` section below actually has the expected commands — `npm ci` before installing Playwright using the `--with-deps` command, so I basically aligned the first example with the second.

There are some other minor differences (step names, test run command, node version) but I believe this should be a minimal working example now.

aslushnikov (approved):

PR #11316, opened 2022-01-11
chore(test runner): refactor worker runner parts

dgozman:
This will make it easier to change in the future.

### DIFF
--- a/packages/playwright-test/src/workerRunner.ts
+++ b/packages/playwright-test/src/workerRunner.ts
@@ -469,7 +469,6 @@ export class WorkerRunner extends EventEmitter {
     let error1: TestError | undefined;
     if (test._type === 'test')
       error1 = await this._runFn(() => this._runBeforeHooks(test, testInfo), testInfo, 'allowSkips');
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/workerRunner.ts
```diff
@@ -481,69 +466,42 @@ export class WorkerRunner extends EventEmitter {
       canHaveChildren: true,
       forceNoParent: true
     });
+    let error1: TestError | undefined;
     if (test._type === 'test')
-      await this._runBeforeHooks(test, testInfo);
+      error1 = await this._runFn(() => this._runBeforeHooks(test, testInfo), testInfo, 'allowSkips');
+    // Continue running afterEach hooks even after the failure.
```
yury-s, line 472: This line is just confusing, perhaps drop it?

yury-s (approved):

packages/playwright-test/src/workerRunner.ts
dgozman, line 472:
```suggestion
```

PR #11339, opened 2022-01-12
chore: implement locator._highlight / playwright._hideHighlight

pavelfeldman:


packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -856,6 +856,33 @@ export class InjectedScript {
     return error;
   }
 
+  highlight(selector: ParsedSelector) {
+    this.hideHighlight();
+    (window as any).pw_highlights = [];
```
dgozman, line 861: Why not `this.highlights` ?

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -856,6 +856,33 @@ export class InjectedScript {
...
+    const elements = this.querySelectorAll(selector, document.documentElement);
+    for (const element of elements) {
+      const box = element.getBoundingClientRect();
+      const div = document.createElement('div');
```
dgozman, line 865: Use `x-pw-debug-highlight` instead to avoid accidental styling from the page's css?

dgozman (approved):

packages/playwright-core/src/server/injected/injectedScript.ts
pavelfeldman: historic reason, fixing

dgozman (approved):

PR #11342, opened 2022-01-12
fix(test runner): hangs in calculateFileSha

### LINKED ISSUE #11341: [BUG] testInfo.attach(…) hangs with paths in fixtures
**Context:**
- Playwright Version: `@playwright/test@1.18.0-alpha-jan-12-2022`
- Operating System: macOS
- Node.js version: v14.17.6

**Code Snippet**

```ts
export const exampleTest = _test.extend<{ aNumber: number }>({
  aNumber: async ({}, use, testInfo) => {
    await use(10);
    console.log("use completed");
    await testInfo.attach("example.txt", {
      contentType: "text/plain",
      path: "/path/to/nowhere",
    });
    console.log("attach completed");
  },
});

exampleTest("foo", ({ aNumber }) => {
  expect(aNumber).toBe(10);
  console.log("done");
});
```

**Describe the bug**

The above code hangs when you run `npx playwright --reporter line`.  I expect it to error immediately. If there is an attachment that exists, I'd expect it to be included, but this case should at least error.

---

The above repro is a simplified example. I hit it when trying to add a per-test logger fixture which I want to add since we have too many `console.log`s spamming our test output (but who's output I don't want to lose).

Here's the fixture I was writing:

```ts
  logger: Logger;
  pollAPI: (id: string) => Promise<string>;
}>({
  logger: async ({}, use, testInfo) => {
    const outFile = testInfo.outputPath("test-logs.log");
    const transports: winston.transport[] = [
      new winston.transports.File({
        filename: outFile,
      }),
    ];

    // Force console output. This is sometimes nice to have locally.
    if (process.env["PLAYWRIGHT_TEST_FORCE_CONSOLE_LOGS"]) {
      transports.push(new winston.transports.Console());
    }

    const logger = winston.createLogger({
      level: "debug",
      transports,
    });

    const finished = new Promise((res, rej) => {
      logger.on("finish", res);
      logger.on("error", rej);
    });

    await use(logger);
    logger.end();
    await finished;
    await testInfo.attach("debug.log", {
      path: outFile,
      contentType: "text/plain",
    });
  },
  pollAPI: async ({ logger }, use) => {
    const pollAPI = async (id: string) => {
      logger.debug(`Polling for ${id}`);
      const cycles = Math.random() * 100;
      for (let i = 0; i < cycles; i++) {
        logger.debug(`still polling`);
      }
      logger.debug("finished polling");
      return "example";
    };
    await use(pollAPI);
  },
});

test("spam example", async () => {
  console.log("spams the console…");
});

test("per-test logging, without the spam!", async ({ pollAPI }) => {
  const result = await pollAPI("foo");
  expect(result).toBe("example");
});

```

I'll open an PR with a test case in a few.

rwoll:
Fixes #11341

See runs of 2bc8ffdec9342402d459601c765bb1c33b2cc0df where it was originally failing from before the calculateFileSha fix.

### DIFF
--- a/packages/playwright-core/src/utils/utils.ts
+++ b/packages/playwright-core/src/utils/utils.ts
@@ -405,12 +405,14 @@ export function arrayToObject(array?: NameValue[]): { [key: string]: string } | 
 export async function calculateFileSha1(filename: string): Promise<string> {
   const hashStream = new HashStream();
   const stream = fs.createReadStream(filename);
-  stream.once('error', err => hashStream.emit('error', err));
   stream.on('open', () => stream.pipe(hashStream));
-  await new Promise((f, r) => {
-    hashStream.on('finish', f);
-    hashStream.on('error', r);
-  });
+  await Promise.race([
+    new Promise((_, r) => stream.on('error', r)),
+    new Promise((f, r) => {
+      hashStream.on('finish', f);
+      hashStream.on('error', r);
+    }),
+  ]);
   return hashStream.digest();
 }
 
--- a/packages/playwright-core/src/utils/utils.ts
+++ b/packages/playwright-core/src/utils/utils.ts
@@ -405,6 +405,7 @@ export function arrayToObject(array?: NameValue[]): { [key: string]: string } | 
 export async function calculateFileSha1(filename: string): Promise<string> {
   const hashStream = new HashStream();
   const stream = fs.createReadStream(filename);
+  stream.once('error', err => hashStream.emit('error', err));
   stream.on('open', () => stream.pipe(hashStream));
   await new Promise((f, r) => {
     hashStream.on('finish', f);
--- a/tests/playwright-test/reporter-attachment.spec.ts
+++ b/tests/playwright-test/reporter-attachment.spec.ts
@@ -105,7 +105,6 @@ test(`testInfo.attach errors`, async ({ runInlineTest }) => {
 });
 
 test(`testInfo.attach doesn't hang fixture`, async ({ runInlineTest }) => {
-  test.fail();
   const result = await runInlineTest({
     'a.test.js': `
       const { test: _test } = pwt;
--- a/tests/playwright-test/reporter-attachment.spec.ts
+++ b/tests/playwright-test/reporter-attachment.spec.ts
@@ -104,6 +104,35 @@ test(`testInfo.attach errors`, async ({ runInlineTest }) => {
   expect(result.exitCode).toBe(1);
 });
 
+test(`testInfo.attach works in fixture`, async ({ runInlineTest }, testInfo) => {
+  const result = await runInlineTest({
+    'a.test.js': `
+      const fs = require('fs');
+      const { test: _test } = pwt;
+      const test = _test.extend({
+        aFixture: async ({}, use, testInfo) => {
+          const outFile = testInfo.outputPath('foo.txt');
+          await fs.promises.writeFile(outFile, 'Hello world\\n');
+          await use(10);
+          await testInfo.attach('example.txt', {
+            path: outFile,
+          });
+        }
+      });
+
+      test('working', ({ aFixture }) => {
+        expect(aFixture).toBe(2);
+      });
+    `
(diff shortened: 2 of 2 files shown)


tests/playwright-test/reporter-attachment.spec.ts
```diff
@@ -103,3 +103,27 @@ test(`testInfo.attach errors`, async ({ runInlineTest }) => {
   expect(result.failed).toBe(3);
   expect(result.exitCode).toBe(1);
 });
+
+test(`testInfo.attach doesn't hang fixture`, async ({ runInlineTest }) => {
+  test.fail();
```
rwoll, line 108:
```suggestion
  test.fixme();
```

packages/playwright-core/src/utils/utils.ts
```diff
@@ -405,6 +405,7 @@ export function arrayToObject(array?: NameValue[]): { [key: string]: string } |
 export async function calculateFileSha1(filename: string): Promise<string> {
   const hashStream = new HashStream();
   const stream = fs.createReadStream(filename);
+  stream.once('error', err => hashStream.emit('error', err));
```
rwoll, line 408: Looks like this is needed as error streams don't always forward their errors. Although, something smells since this case looks eerily similar to https://github.com/microsoft/playwright/blob/359d523ec318106bf79d650b2b4458a1a388ebdf/tests/playwright-test/reporter-attachment.spec.ts#L89

I don't know all the intricicies of Node.js streams, so HashStream implementation should also be looked at.

rwoll: Reading up more on streams, this looks to be the correct fix for `calculateFileSha1 ` so we don't have an unhandled error.

Before this change, we were triggering https://github.com/microsoft/playwright/blob/da1f39fb29dd3c45a8f26de77f029c2bfc09a790/packages/playwright-test/src/workerRunner.ts#L94 since we weren't forwarding the error and it was surfacing as an unhandled error in both the new test fixture cases and the old (but seemingly "working")  https://github.com/microsoft/playwright/blob/359d523ec318106bf79d650b2b4458a1a388ebdf/tests/playwright-test/reporter-attachment.spec.ts#L89. 

The discrepancy in behavior likely has to do with how unhandled errors are depending on where/when they come: https://github.com/microsoft/playwright/blob/da1f39fb29dd3c45a8f26de77f029c2bfc09a790/packages/playwright-test/src/workerRunner.ts#L94

So, an additional fix is likely needed since I _think_ we'd want the fixture case to also immediately fail and not hang even if there is an unhandled error.

rwoll: > So, an additional fix is likely needed since I think we'd want the fixture case to also immediately fail and not hang even if there is an unhandled error.

`calculateFileSha1` will no longer be the culprit with this fix, but other code could intro unhandled errors in the future.

rwoll: cc: @dgozman since I think you'll be able to explain to me the expectations around how unhandled errors in fixtures vs. tests should behave overall.

pavelfeldman, line 408: Node convention for streams is to dispatch an error event upon failures. And if error handler is not attached, event target will generate unhandled exception. The right way to handle it is to register 'error' event handler forever (on, not once), and upon error, in this case, reject the resulting promise. You can use Promise.race for it here, no need to dispatch it on the foreign hash stream object.

rwoll: Updated in https://github.com/microsoft/playwright/commit/83e988b9b0574c541adce3e3a00f9dd0a32297da.

Let me know if I created a race with the error listener. 😀 I think I'm fine since there are no awaits between where we create the readstream and I register err listener promise.

@pavelfeldman For my own education, when does Node actually go to open the file? Is it as soon as we give over the thread while waiting for the settling of thee `race` race promise?

rwoll:
Thanks, @dgozman! Closing this in favor of https://github.com/microsoft/playwright/pull/11365

PR #11358, opened 2022-01-12
chore: cut v1.18.0

aslushnikov:
Drive-by: fix `//utils/bump_package_version.js` to take care of `html-reporter` package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "playwright-internal",
-  "version": "1.18.0-next",
+  "version": "1.19.0-next",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "playwright-internal",
-      "version": "1.18.0-next",
+      "version": "1.19.0-next",
       "license": "Apache-2.0",
       "workspaces": [
         "packages/*"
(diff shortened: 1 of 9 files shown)


package-lock.json
```diff
@@ -9049,9 +9049,11 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "packages/html-reporter": {},
+    "packages/html-reporter": {
+      "version": "v1.19.0-next"
```
dgozman, line 9053: What's going on here?

package-lock.json
```diff
@@ -9065,7 +9067,7 @@
       }
     },
     "packages/playwright-chromium": {
-      "version": "1.18.0-next",
+      "version": "v1.19.0-next",
```
mxschmitt, line 9070: Seems like there shouldn't be a 'v' prefix.

mxschmitt (approved):

dgozman (approved):

utils/bump_package_versions.js
```diff
@@ -20,10 +20,12 @@ const path = require('path');
 const fs = require('fs');
 const { execSync } = require('child_process');
 
-const { packages } = require('./list_packages.js');
+const { packages, packagesToPublish } = require('./list_packages.js');
 
 (async () => {
   const version = process.argv[2];
+  if (version.startsWith('v'))
+    throw new Error('Version must not start with "v"');
```
campersau, line 28: I guess it would be better to move this after the existing `if (!version)` check. Otherwise it might get a reference error.

aslushnikov: very fair; thank you for pointing out! https://github.com/microsoft/playwright/pull/11448

PR #11365, opened 2022-01-12
fix(test runner): fixture teardown double error, testInfo.attach()

### LINKED ISSUE #11341: [BUG] testInfo.attach(…) hangs with paths in fixtures

dgozman:
- Use file path, not content to calculate the attachment hash.
- Always cleanup fixture from the list on teardown, to avoid reporting
  teardown error multiple times: from the test, and from the cleanup.

Fixes #11341.

### DIFF
--- a/packages/playwright-core/src/utils/utils.ts
+++ b/packages/playwright-core/src/utils/utils.ts
@@ -371,19 +371,6 @@ export function monotonicTime(): number {
   return seconds * 1000 + (nanoseconds / 1000 | 0) / 1000;
 }
 
-class HashStream extends stream.Writable {
-  private _hash = crypto.createHash('sha1');
-
-  override _write(chunk: Buffer, encoding: string, done: () => void) {
-    this._hash.update(chunk);
-    done();
-  }
-
-  digest(): string {
-    return this._hash.digest('hex');
-  }
-}
-
 export function objectToArray(map?:  { [key: string]: any }): NameValue[] | undefined {
   if (!map)
     return undefined;
@@ -400,17 +387,6 @@ export function arrayToObject(array?: NameValue[]): { [key: string]: string } | 
   for (const { name, value } of array)
     result[name] = value;
   return result;
-}
-
-export async function calculateFileSha1(filename: string): Promise<string> {
-  const hashStream = new HashStream();
-  const stream = fs.createReadStream(filename);
-  stream.on('open', () => stream.pipe(hashStream));
-  await new Promise((f, r) => {
-    hashStream.on('finish', f);
-    hashStream.on('error', r);
-  });
-  return hashStream.digest();
 }
 
 export function calculateSha1(buffer: Buffer | string): string {
--- a/packages/playwright-test/src/workerRunner.ts
+++ b/packages/playwright-test/src/workerRunner.ts
@@ -30,7 +30,7 @@ import { Annotations, TestCaseType, TestError, TestInfo, TestInfoImpl, TestStepI
 import { ProjectImpl } from './project';
 import { FixtureRunner } from './fixtures';
 import { DeadlineRunner, raceAgainstDeadline } from 'playwright-core/lib/utils/async';
-import { calculateFileSha1 } from 'playwright-core/lib/utils/utils';
+import { calculateSha1 } from 'playwright-core/lib/utils/utils';
 
 const removeFolderAsync = util.promisify(rimraf);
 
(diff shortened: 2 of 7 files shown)


packages/playwright-core/src/utils/utils.ts
```diff
@@ -402,17 +389,6 @@ export function arrayToObject(array?: NameValue[]): { [key: string]: string } |
   return result;
 }
 
-export async function calculateFileSha1(filename: string): Promise<string> {
```
rwoll, line 405: Can't have a bug in it if the code no longer exists! 😃

tests/playwright-test/fixture-errors.spec.ts
```diff
@@ -435,3 +435,23 @@ test('should not teardown when setup times out', async ({ runInlineTest }) => {
   expect(result.output.split('\n').filter(line => line.startsWith('%%'))).toEqual([
   ]);
 });
+
+test('should not report fixture teardown error twice', async ({ runInlineTest }) => {
```
rwoll, line 439: Can you bring over the other test, too? 

https://github.com/microsoft/playwright/pull/11342/files#diff-7baf8520adc91e87042c746455f6e8a73b78771e9ca03095c9b15437894a6b4aR107

I intentionally had two tests so I ensured the positive test case actually does work alongside the negative test case. (Afterall, the positive case is what lead us here 😀 so I'd want to know if we broke that in the future, and it feels different enough from the other attachment cases since it happens in fixture teardown as opposed to during the test.)

packages/playwright-test/src/workerRunner.ts
```diff
@@ -268,7 +268,7 @@ export class WorkerRunner extends EventEmitter {
         if ((options.path !== undefined ? 1 : 0) + (options.body !== undefined ? 1 : 0) !== 1)
           throw new Error(`Exactly one of "path" and "body" must be specified`);
         if (options.path) {
-          const hash = await calculateFileSha1(options.path);
+          const hash = calculateSha1(options.path);
```
rwoll, line 271: What's the reason for the switch to filename instead of contents? Regardless, filename hash seems much simpler (and I suppose negligibly faster depending on what was in the files).

rwoll (changes requested):
Left some questions for my own education/curiosity/understanding, and a request to bring over one more of the test cases.

Thanks!

packages/playwright-test/src/workerRunner.ts
dgozman: You got it right 😄

tests/playwright-test/fixture-errors.spec.ts
dgozman: Sure, why not.

pavelfeldman (approved):

Issue #11272, opened 2022-01-09
[Feature] Remove dependency on color.js


lilithium-hydride:
color.js' maintainer had a fit and sabotaged the package, and it will now print out the American flag, Abraham Lincoln, and an endless stream of zalgo text (see: `https://github.com/Marak/colors.js/commit/074a0f8ed0c31c35d13d28632bd8a049ff136fb6`, `https://github.com/Marak/colors.js/issues/285`). I have avoided linking them directly as to avoid having this issue bombed.

This is problematic and should probably be rectified.

See https://github.com/Marak/colors.js/issues/285

DABH:
Please see https://github.com/Marak/colors.js/issues/285#issuecomment-1008212640 for latest update - thanks

sindresorhus:
If you're looking to switch to a well-maintained coloring package you can trust. Check out [Chalk](https://github.com/chalk/chalk). I'm one of the maintainers.

btd:
Marak added change to color/safe, so every install now affected

pavelfeldman:
> If you're looking to switch to a well-maintained coloring package you can trust. Check out Chalk. I'm one of the maintainers.

@sindresorhus that'd be nice, is there a way for us to use it from our TS code? Import is compiled into require('chalk'), which fails due to ESM vs CommonJS. I was going to raise the issue w/ question in your repo, but it won't let me.

pavelfeldman:
Fixed in #11274

sindresorhus:
> is there a way for us to use it from our TS code? 

TS is lagging behind on ESM support, so for now, you can just use [version 4](https://github.com/chalk/chalk/tree/v4.1.2) (which has been stable for years). TS is scheduled to fully support ESM in TS 4.6 which will be out in February.

PR #11366, opened 2022-01-13
fix: update to @dabh/colors for safe alternative

mannyluvstacos:
This PR updates the color package to using [@dabh/colors](https://www.npmjs.com/package/@dabh/colors) as stated on this [colors issue #317](https://github.com/Marak/colors.js/issues/317#issue-1098535594) which is a safe alternative.


EthanHindmarsh:
Hello Manny,

Colors.js has already removed the offending version ( https://www.npmjs.com/package/colors )
There is no need for this change 👍

This issue has also been resolved in issue #11272

Issue #10942, opened 2021-12-15
[Feature] Junit Reporter - Adding custom testsuite properties


JuanCruzTrinidad:
Hi! This time we come to propose custom properties in test suite return to xml file.
Use the fixture in jest-junit reporter. 
https://www.npmjs.com/package/jest-junit
![image](https://user-images.githubusercontent.com/53027244/146187765-086f3fd2-ff81-4e6f-8ce7-77ca9d17679d.png)

My team use datadog for ci analytics test and they read properties in suite and properties for test case.

https://docs.datadoghq.com/continuous_integration/setup_tests/junit_upload/#providing-metadata-through-property-elements

![image](https://user-images.githubusercontent.com/53027244/146209058-d661ffeb-0d9b-4b2f-b785-ad3bb04f963c.png)

pavelfeldman:
#### Why was this issue closed?

We are prioritizing the features based on the upvotes, recency and activity in the issue thread. It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it. We are closing issues like this one to keep our bug database maintainable. Please feel free to open a new issue and link this one to it if you think this is a mistake.

PR #11374, opened 2022-01-13
feat(junit reporter): link testcases to Xray test issues and provide additional metadata for Xray Test Management

bitcoder:
Provides support for specific enhancements on the junit report, allowing deeper integration with Xray Test Management to show additional information on Jira side, or do some related operations upon the time of importing tesults.
The test metainformation may be specified using annotations on the `TestInfo` object.

This PR gives users the ability to:
- link a Playwright test to an existing Test in Xray, by referring its issue key using an annotation with type `test_key`
- link a Playwright test to an existing Test in Xray, by referring its issue id using an annotation with type `test_id`
- link a Playwright test to an existing story/requirement kinda of issue in Jira, by referring its key using an annotation with type `requirements`; multiple requirements may be specified, using comma as a delimiter on the annotation's description
- specify the Test issue' summary, to be used during provisioning of the Test in Xray if it doesn't exist yet, using an annotation with type 'test_summary'
- specify the Test issue's description, to be used during provisioning of the Test in Xray if it doesn't exist yet, using an annotation with type 'test_description'
- embed attachments, including its contents, on the report, by flagging it using a `embed_attachments_in_report` annotation with 'true' as its description
 
Resolves #9348

### DIFF
--- a/docs/src/test-reporters-js.md
+++ b/docs/src/test-reporters-js.md
@@ -393,7 +393,12 @@ The JUnit reporter provides support for embedding additional information on the 
 
 In configuration file, a set of options can be used to configure this behavior. A full example, in this case for Xray, follows ahead.
 
-```js js-flavor=ts
+```js js-flavor=js
+// playwright.config.js
+// @ts-check
+
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+
 // JUnit reporter config for Xray
 const xrayOptions = {
   // Whether to add <properties> with all annotations; default is false
@@ -411,15 +416,60 @@ const xrayOptions = {
   outputFile: './xray-report.xml'
 };
 
-module.exports = {
-  reporter: [['junit', xrayOptions]]
+const config = {
+  reporter: [ ['junit', xrayOptions] ]
 };
+
+module.exports = config;
+```
+
+```js js-flavor=ts
+// playwright.config.ts
+import { PlaywrightTestConfig } from '@playwright/test';
+
+// JUnit reporter config for Xray
+const xrayOptions = {
+  // Whether to add <properties> with all annotations; default is false
+  embedAnnotationsAsProperties: true,
+
+  // By default, annotation is reported as <property name='' value=''>.
+  // These annotations are reported as <property name=''>value</property>.
+  textContentAnnotations: ['test_description'],
+
+  // This will create a "testrun_evidence" property that contains all attachments. Each attachment is added as an inner <item> element.
+  // Disables [[ATTACHMENT|path]] in the <system-out>.
+  embedAttachmentsAsProperty: 'testrun_evidence',
+
+  // Where to put the report.
+  outputFile: './xray-report.xml'
+};
+
+const config: PlaywrightTestConfig = {
+  reporter: [ ['junit', xrayOptions] ]
+};
+
+export default config;
 ```
 
 In the previous configuration sample, all annotations will be added as `<property>` elements on the JUnit XML report. The annotation type is mapped to the `name` attribute of the `<property>`, and the annotation description will be added as a `value` attribute. In this case, the exception will be the annotation type `testrun_evidence` whose description will be added as inner content on the respective `<property>`.
 Annotations can be used to, for example, link a Playwright test with an existing Test in Xray or to link a test with an existing story/requirement in Jira (i.e., "cover" it).
 
 ```js js-flavor=js
+// @ts-check
+const { test } = require('@playwright/test');
+
+test('using specific annotations for passing test metadata to Xray', async ({}, testInfo) => {
+  testInfo.annotations.push({ type: 'test_id', description: '1234' });
+  testInfo.annotations.push({ type: 'test_key', description: 'CALC-2' });
+  testInfo.annotations.push({ type: 'test_summary', description: 'sample summary' });
+  testInfo.annotations.push({ type: 'requirements', description: 'CALC-5,CALC-6' });
+  testInfo.annotations.push({ type: 'test_description', description: 'sample description' });
+});
+```
+
+```js js-flavor=ts
+import { test } from '@playwright/test';
+
 test('using specific annotations for passing test metadata to Xray', async ({}, testInfo) => {
   testInfo.annotations.push({ type: 'test_id', description: '1234' });
   testInfo.annotations.push({ type: 'test_key', description: 'CALC-2' });
@@ -434,6 +484,8 @@ Please note that the semantics of these properties will depend on the tool that 
 If the configuration option `embedAttachmentsAsProperty` is defined, then a `property` with its name is created. Attachments, including their contents, will be embeded on the JUnit XML report inside `<item>` elements under this `property`. Attachments are obtained from the `TestInfo` object, using either a path or a body, and are added as base64 encoded content.
 Embedding attachments can be used to attach screenshots or any other relevant evidence; nevertheless, use it wisely as it affects the report size.
 
+The following configuration sample enables embedding attachments by using the `testrun_evidence` element on the JUnit XML report:
+
 ```js js-flavor=js
 // playwright.config.js
 // @ts-check
@@ -442,14 +494,43 @@ Embedding attachments can be used to attach screenshots or any other relevant ev
 const config = {
   reporter: [ ['junit', { embedAttachmentsAsProperty: 'testrun_evidence', outputFile: 'results.xml' }] ],
 };
+
+module.exports = config;
 ```
 
+```js js-flavor=ts
+// playwright.config.js
+
+import { PlaywrightTestConfig } from '@playwright/test';
+const config: PlaywrightTestConfig = {
+  reporter: [ ['junit', { embedAttachmentsAsProperty: 'testrun_evidence', outputFile: 'results.xml' }] ],
+};
+
+export default config;
+```
+
+The following test adds attachments:
+
 ```js js-flavor=js
+// @ts-check
+const { test } = require('@playwright/test');
+
 test('embed attachments, including its content, on the JUnit report', async ({}, testInfo) => {
   const file = testInfo.outputPath('evidence1.txt');
   require('fs').writeFileSync(file, 'hello', 'utf8');
-  testInfo.attachments.push({ name: 'evidence1.txt', path: file, contentType: 'text/plain' });
-  testInfo.attachments.push({ name: 'evidence2.txt', body: Buffer.from('world'), contentType: 'text/plain' });
+  await testInfo.attach('evidence1.txt', { path: file, contentType: 'text/plain' });
+  await testInfo.attach('evidence2.txt', { body: Buffer.from('world'), contentType: 'text/plain' });
+});
+```
+
+```js js-flavor=ts
+import { test } from '@playwright/test';
+
+test('embed attachments, including its content, on the JUnit report', async ({}, testInfo) => {
+  const file = testInfo.outputPath('evidence1.txt');
+  require('fs').writeFileSync(file, 'hello', 'utf8');
+  await testInfo.attach('evidence1.txt', { path: file, contentType: 'text/plain' });
+  await testInfo.attach('evidence2.txt', { body: Buffer.from('world'), contentType: 'text/plain' });
 });
 ```
 
--- a/docs/src/test-reporters-js.md
+++ b/docs/src/test-reporters-js.md
@@ -442,6 +442,7 @@ Embedding attachments can be used to attach screenshots or any other relevant ev
 const config = {
   reporter: [ ['junit', { embedAttachmentsAsProperty: 'testrun_evidence', outputFile: 'results.xml' }] ],
 };
+```
 
 ```js js-flavor=js
 test('embed attachments, including its content, on the JUnit report', async ({}, testInfo) => {
--- a/docs/src/test-reporters-js.md
+++ b/docs/src/test-reporters-js.md
@@ -389,6 +389,31 @@ const config: PlaywrightTestConfig = {
 export default config;
 ```
 
+The JUnit reporter also provides support for Xray Test Management, which uses an evolved JUnit XML format to provide additional information on the `testcase` elements using inner `properties`. Specific annotations can be used to, for example, link a Playwright test with an existing Test in Xray or to link a test with an existing story/requirement in Jira (i.e., "cover" it). For more information about the available properties and their semantics, please check [Xray Test Management documentation](https://docs.getxray.app/display/XRAYCLOUD/Taking+advantage+of+JUnit+XML+reports).
+
+```js js-flavor=js
+test('using specific annotations for passing test metadata to Xray', async ({}, testInfo) => {
+  testInfo.annotations.push({ type: 'test_id', description: '1234' });
+  testInfo.annotations.push({ type: 'test_key', description: 'CALC-2' });
+  testInfo.annotations.push({ type: 'test_summary', description: 'sample summary' });
+  testInfo.annotations.push({ type: 'requirements', description: 'CALC-5,CALC-6' });
+  testInfo.annotations.push({ type: 'test_description', description: 'sample description' });
+});
+```
+
+A more advanced scenario is supported by Xray: the ability to embed attachments, including their contents, on the JUnit XML report. For that, the `embed_attachments_in_report` annotation must be specified, having the `true` value as its description. Attachments are obtained from the `TestInfo` object, using either a path or a body, and are embed as base64 encoded content on a custom `property` on the JUnit report.
+Embedding attachments can be used to attach screenshots or any other relevant evidence; nevertheless, use it wisely as it affects the report size.
+
+```js js-flavor=js
+test('embed attachments on the JUnit report processable by Xray', async ({}, testInfo) => {
+  testInfo.annotations.push({ type: 'embed_attachments_in_report', description: 'true' });
+  const file = testInfo.outputPath('evidence1.txt');
+  require('fs').writeFileSync(file, 'hello', 'utf8');
+  testInfo.attachments.push({ name: 'evidence1.txt', path: file, contentType: 'text/plain' });
+  testInfo.attachments.push({ name: 'evidence2.txt', body: Buffer.from('world'), contentType: 'text/plain' });
+});
+```
+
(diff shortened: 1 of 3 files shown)


btd:
I would say we need generic way to set properties in junit reporter, like described there: https://github.com/microsoft/playwright/issues/10942
As you can see in mentioned issue, a lot of different properties could be added. I do not think it is right idea to add support for specific tool, but just a generic way to set properties

bitcoder:
@btd thanks for your feedback; I understand your point but this PR and feature request  is targeted for adding support for Xray Test Management and make that process as clean as possible for these users (by curiosity and to have just a more clear idea, Xray is used by several millions of users in more than 17K orgs). 
Making it more generic is possible *but* would making the integration with Xray more cumbersome; for example, with Xray the values of some properties are embed as "value" attribute, while others (because of their nature) are embed as inner cdata content. In the case of attachments, these have a specific mapping on the JUnit XML, with multiple inner `<item>` elements, with specific attributes and inner content. The current implementation hides this logic and makes it straighforward to use the attachments on the `TestInfo` object.
From what I saw, the datadog feature request #10942, also mentions `properties` at `testsuite` level, which the current PR doesn't address nor is supported by the built-in junit reporter on playwright-test.

As I've invested quite some time on the PR, I'll wait for feedback from the playwright team before doing additional changes.

pdinhgia:
Hello, any updates about this issue ? I'm very interested about this feature

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -133,6 +133,94 @@ class JUnitReporter implements Reporter {
...
     entries.push(entry);
 
+    // Xray Test Management supports testcase level properties, where additional metadata may be provided
+    // some annotations are encoded as value attributes, other as cdata content
+    const properties: XMLEntry = {
+      name: 'properties',
+      children: [] as XMLEntry[]
+    };
+
+    const textValueBasedAnnotations = ['test_id', 'test_key', 'test_summary', 'requirements'];
```
dgozman, line 143: I am hesitant supporting only some annotations. Why don't we put all annotations as properties? Should be safe enough, as long as Xray ignores unknown properties, which it probably does?

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -133,6 +133,94 @@ class JUnitReporter implements Reporter {
...
+      const annotation = test.annotations.find(annotation => {
+        return annotation.type === annotationType;
+      });
+      if (annotation !== undefined) {
+        const property: XMLEntry = {
+          name: 'property',
+          attributes: {
+            name: annotationType
+          },
+          text: annotation.description
```
dgozman, line 171: Why do we have to put this one as text, and not as `value=` as others? I'd assume that these are interchangeable, it's just a matter of escaping.

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -133,6 +133,94 @@ class JUnitReporter implements Reporter {
...
+        };
+        properties.children?.push(property);
+      }
+    }
+
+    // attachments are optionally embed as base64 encoded content on inner elements
+    const embedAttachments = test.annotations.find(annotation => {
+      return annotation.type === 'embed_attachments_in_report';
```
dgozman, line 179: I think it's better to have an option on the whole junit reporter, to say "embed attachments". Even in this case, we probably should not embed files, wdyt?

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -133,6 +133,94 @@ class JUnitReporter implements Reporter {
...
+    });
+    if (embedAttachments?.description === 'true') {
+      const evidence: XMLEntry = {
+        name: 'property',
+        attributes: {
+          name: 'testrun_evidence'
```
dgozman, line 185: Let's add an option to the junit reporter: `{ embedAttachmentsAsProperty: 'testrun_evidence' }`? This seems a very specific name to put here.

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -133,6 +133,94 @@ class JUnitReporter implements Reporter {
...
+
+          if (contents) {
+            const item: XMLEntry = {
+              name: 'item',
+              attributes: {
+                name: attachment.name
+              },
+              text: contents
+            };
+            evidence.children?.push(item);
```
dgozman, line 213: When embedding attachments, we should skip the "reference attachments" logic on the line 245.

dgozman (commented):
Sorry for being late! This PR looks very good overall, but I think we should make it a bit more generic - see comment. Thank you for contributing ❤️

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -205,7 +293,7 @@ function escape(text: string, stripANSIControlSequences: boolean, isCharacterDat
   const escapeRe = isCharacterData ? /[&<]/g : /[&"<>]/g;
   text = text.replace(escapeRe, c => ({ '&': '&amp;', '"': '&quot;', '<': '&lt;', '>': '&gt;' }[c]!));
   if (isCharacterData)
-    text = text.replace(/]]>/g, ']]&gt;');
+    text = '<![CDATA[' + text.replace(/]]>/g, ']]&gt;') + ']]>';
```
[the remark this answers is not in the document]
bitcoder: Well, I remember that whenever I was testing I wasn't getting some errors because it would produce invalid XML on the parser I was using. Nevertheless, if that block is character data it should be in a [CDATA block](https://www.w3.org/TR/REC-xml/#sec-cdata-sect). It doesn't make sense in my perspective, know that is character data and then not properly identify as such, right? :)
https://www.w3.org/TR/REC-xml/#sec-cdata-sect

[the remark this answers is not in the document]
bitcoder: It's a bit unrelated indeed but I think it should be the standard behaviour (I don't see a need to create an option for it unless there's a known use case where having CDATA is going to break anything for any parser out there, which I doubt but sometimes there are basic parsers...)

packages/playwright-test/src/reporters/junit.ts
bitcoder: Well, this annotations are the ones Xray understands as text based values. We could make it more generic (please see next comment)

packages/playwright-test/src/reporters/junit.ts
bitcoder: That `testrun_evidence` name is the one that Xray parses but I'm ok having it configurable, just let me know if you want to proceed that way.
Just for information, JUnit doesn't specify how to embed attachments in the report. This is something specified by the Xray team (even though I've validated it with JUnit5 team leader at high-level just to know if it would hurt the "schema" or go against some principles), that eventually other tools can take advantage of :)

packages/playwright-test/src/reporters/junit.ts
bitcoder: I agree

packages/playwright-test/src/reporters/junit.ts
bitcoder: Embeding attachments is in my perspective something that users will need to opt-in explicitly and should be disabled by default as it will make reports larger.
I'm not sure if I fully understood you, but please feel free to clairify/correct.
If you're suggesting to add a global configuration for the whole reporter, I agree, having it as false by default. I'm not sure if you were suggesting to remove this logic at the test level using the 'embed_attachments_in_report' annotation type that I've implemented.. or if you were suggesting having that global configuration plus this logic... or something else :)
Well, in my perspective maybe a global option for the junit reporter would be enough... not sure if users having that option as true, would want to disable this behavior in some test(s). If this makes sense, then the existing logic combined with the global option could make sense. What are your thoughts?

packages/playwright-test/src/reporters/junit.ts
bitcoder: For the time being, Xray processes some properties from a `value` attribute and other ones from inner text, due to their nature. It's not interchangeable.
So we have a set of well-known properties are are either mapped as "myproperty1" or as shown for "myproperty2".

`
<property name="myproperty1" value="dummyvalue"/>
<property name="myproperty2">
 a very long description .....
</property>
`

To make this mechanism generic and process either from the value or from the inner text,  we would need to change Xray. In XML there are limitations for the usage of the `value` attribute, as AFAIK though.

One option here moving forward, in order to make it more agnostic and not have any hardcoding of the annotations that should be either mapped to a value or to inner text, would be to have some sort of explicitly differentiate that whenever using the annotations. However, in playwright we just have the `type` and the `description` fields, correct? 

`
testInfo.annotations.push({ type: 'test_id', description: '1234' });
`

I don't see here a way of signalling it from the outside.. unless we would use a prefix on the `type` or on the  `description` or something like that.
Any ideas?

dgozman: I see, thank you for clarifying. Let's go with options for now, and provide a nice snippet in the docs:

```js
// JUnit reporter config for XRay.
const xrayOptions = {
  // Whether to add <properties> with all annotations.
  embedAnnotationsAsProperties: true,

  // By default, annotation is reported as <property name='' value=''>.
  // These annotations are reported as <property name=''>value</property>.
  textContentAnnotations: ['test_description'],

  // This will create a "testrun_evidence" property that contains all attachments.
  // Disables [[ATTACHMENT|path]] in the <system-out>.
  embedAttachmentsAsProperty: 'testrun_evidence',

  // Where to put the report.
  outputFile: './xray-report.xml',
};

module.exports = {
  reporter: [['list'],  ['junit', xrayOptions]]
};
```

packages/playwright-test/src/reporters/junit.ts
dgozman: I agree this should be disabled by default. I'm inclined for a single global option on the reporter for now, unless someone explicitly asks for more fine-grained control per each test. See my other comment with suggested options.

packages/playwright-test/src/reporters/junit.ts
dgozman: Let's go with a single global option for the junit reporter. See my other comment with suggested options.

packages/playwright-test/src/reporters/junit.ts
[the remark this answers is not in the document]
dgozman: Yeah, I am hesitant changing it by default - just in case some of the parsers in the wild breaks and we regress existing users. An option is a safe bet, and we can recommend it in the docs for all new users. A separate PR with this option would be ideal 😄

dgozman (commented):
I was on vacations, so didn't get to this before, unfortunately. Sorry!

Let's go with adding some options, and this will be a great change! Note that you'll need to sign CLA to be able to contribute.

packages/playwright-test/src/reporters/junit.ts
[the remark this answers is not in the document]
bitcoder: I've made a new PR  for that #12741

packages/playwright-test/src/reporters/junit.ts
bitcoder: done, updated the PR, including the docs

packages/playwright-test/src/reporters/junit.ts
bitcoder: done, with a global option for now. updated the PR

dgozman (commented):
I think we are on the right track, just a few small details left.

docs/src/test-reporters-js.md
```diff
@@ -389,6 +389,70 @@ const config: PlaywrightTestConfig = {
 export default config;
 ```
 
+The JUnit reporter provides support for embedding additional information on the `testcase` elements using inner `properties`. This is based on an [evolved JUnit XML format](https://docs.getxray.app/display/XRAYCLOUD/Taking+advantage+of+JUnit+XML+reports) from Xray Test Management, but can also be used by other tools if they support this way of embedding additonal information for test results; please check it first.
+
+In configuration file, a set of options can be used to configure this behavior. A full example, in this case for Xray, follows ahead.
+
+```js js-flavor=ts
```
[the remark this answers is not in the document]
bitcoder: thanks @dgozman  for the feedback, we'll have a look at it

tests/playwright-test/reporter-junit.spec.ts
```diff
@@ -255,3 +255,140 @@ function parseXML(xml: string): any {
...
+  expect(testcase['properties'][0]['property'][1]['$']['value']).toBe('unknown');
+  expect(result.exitCode).toBe(0);
+});
+
+test('should render all annotations to testcase value based properties, if requested', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'playwright.config.ts': `
+      const xrayOptions = {
+        embedAnnotationsAsProperties: true,
+        textContentAnnotations: ['test_description']
```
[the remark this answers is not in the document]
bitcoder: right, it shouldn't be there. removed

mxschmitt:
@bitcoder thank you for your contribution! 💯 don't forget to sign the CLA https://github.com/microsoft/playwright/pull/11374#issuecomment-1012083148, otherwise we can't merge PRs.

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -133,6 +140,82 @@ class JUnitReporter implements Reporter {
...
+          let contents;
+          if (attachment.body) {
+            contents = attachment.body.toString('base64');
+          } else {
+            if (!attachment.path)
+              continue;
+            try {
+              const attachmentPath = path.relative(this.config.rootDir, attachment.path);
+              if (fs.existsSync(attachmentPath))
+                contents = fs.readFileSync(attachmentPath, { encoding: 'base64' });
```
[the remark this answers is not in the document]
bitcoder: given the previous comment, where I mentioned it would make sense to add the warning to the `<system-err>` element, then I'll keep this approach and the `existsSync` as also exists in the other code block

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -133,6 +140,82 @@ class JUnitReporter implements Reporter {
...
+            } catch (e) {
+            }
```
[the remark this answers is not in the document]
dgozman: Sounds good.

packages/playwright-test/src/reporters/junit.ts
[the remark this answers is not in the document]
dgozman: Sure, although this would require handling both "does not exist" and "error reading file" situations separately. In the latter case we can even print the error message to `<system-err>`.

docs/src/test-reporters-js.md
```diff
@@ -389,6 +389,70 @@ const config: PlaywrightTestConfig = {
...
+
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+const config = {
+  reporter: [ ['junit', { embedAttachmentsAsProperty: 'testrun_evidence', outputFile: 'results.xml' }] ],
+};
+```
+
+```js js-flavor=js
+test('embed attachments, including its content, on the JUnit report', async ({}, testInfo) => {
+  const file = testInfo.outputPath('evidence1.txt');
```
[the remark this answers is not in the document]
bitcoder: sure, makes sense. The branch I have still has a old implementation of testInfo.attach which is not compatible with the latest one. In the doc I'll refer `testInfo.attach`; in the related test I use `testInfo.attachments.push` (and I've left the new testInfo.attach commented..)

bitcoder:
@dgozman  thanks for your patience reviewing this. Hope the latest PR code is closer for merging, but please review it once again.

docs/src/test-reporters-js.md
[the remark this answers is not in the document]
bitcoder: updated in the latest PR code

packages/playwright-test/src/reporters/junit.ts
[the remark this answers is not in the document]
bitcoder: I have kept the same login as the existing oneon the initial code, where if file does not exist then it will add it to <system-err> (as it means the user provided a wrong path); if it fails reading the file, it is catch and doesn't do anything. 
This can be changed though, but I would say that then we would also need to change the existing logic on the other code.

dgozman (approved):
Looks great! I will be able to merge after you sign the CLA - see the comment from microsoft-cla bot.

mxschmitt:
@bitcoder any updates on the CLA? :) (want to reduce risks with merge conflicts)

bitcoder:
hey @mxschmitt , thanks for reaching out. Unfortunately not yet (I have no ETA for it also). I've asked for this to the legal team a few times and it is taking  more than I've anticipated.. :( thanks for your patience in advance

dgozman:
@bitcoder Just a friendly reminder to ping the legal team about this issue 😄 I hope to merge you great PRs!

bitcoder:
> @bitcoder Just a friendly reminder to ping the legal team about this issue 😄 I hope to merge you great PRs!

on it... I also want to move this forward as fast as possible (doing what I can here..)

Alje001:
@bitcoder  @dgozman
Do you have any update on this task? This is a very important feature, which we like to use. 

Thank you for supporting

Alje001:
do you have any update on this task? it is an important feature and we need it for test traceability.

thank you for supporting.

raghavthind2005 (approved):

bitcoder:
> @bitcoder @dgozman Do you have any update on this task? This is a very important feature, which we like to use.
> 
> Thank you for supporting

not yet, sorry for that. this is taking way longer than I would expect but as Microsoft requires the signed CLA and my company is still going over it, I cannot do anything els right now. Hope for your understanding

mxschmitt:
@bitcoder friendly ping, that you ping your legal team about the CLA, thank you in advance!

bitcoder:
@mxschmitt on it; even talked about that yesterday to upper mgnt

Alje001:
I hope that you @bitcoder will get the Cla soon, because we would like to use this feature

bitcoder:
CLA has been signed; I got the approval from my current company (Xblend).

bitcoder:
@mxschmitt  @dgozman  fyi, this should now be able to be merged

PR #11410, opened 2022-01-14
docs(dotnet): enable collecting tracing sources

mxschmitt:


yury-s (approved):

yury-s (commented):
In Java it requires setting an environment variable with path to source files. We may have to do something similar in C#, in which case you'll need a separate definition for it (similar to java).

mxschmitt:
> In Java it requires setting an environment variable with path to source files. We may have to do something similar in C#, in which case you'll need a separate definition for it (similar to java).

When building .NET programs the absolute stack-traces get attached as symbols. If they are available, we use them, if not, we skip. I'd for not rewrite it with a custom path.

PR #11414, opened 2022-01-14
fix(test runner): dispatcher transformation of empty attachments

### LINKED ISSUE #11413: [BUG] empty inline attachments crash reporter
**Context:**
- Playwright Version: ToT

**Code Snippet**

(I'll open an MR with failing test cases after filing this.)

```
      test('success', async ({}, testInfo) => {
        await testInfo.attach('name', { body: '', contentType: 'text/plain' });
        expect(0).toBe(1);
      });
```

**Describe the bug**

Crashes test runner:

```
Running 1 test using 1 worker
F

TypeError: Cannot read property 'toString' of undefined
(exit)
```

💥 occurs here: https://github.com/microsoft/playwright/blob/205969e95827f04068ec460a849e2fe6a0cbb4ca/packages/playwright-test/src/reporters/base.ts#L260

At that line `attachment` has the value of:

```
attachments: {
  name: 'name',
  path: undefined,
  contentType: 'text/plain',
  body: undefined
}
```

so it looks like somewhere `''` is getting turned into an `undefined`. (Same thing happens with `Buffer.from('')`).

cc: @dgozman

rwoll:
Fixes #11413

### DIFF
--- a/packages/playwright-test/src/dispatcher.ts
+++ b/packages/playwright-test/src/dispatcher.ts
@@ -204,7 +204,7 @@ export class Dispatcher {
         name: a.name,
         path: a.path,
         contentType: a.contentType,
-        body: a.body ? Buffer.from(a.body, 'base64') : undefined
+        body: typeof a.body !== 'undefined' ? Buffer.from(a.body, 'base64') : undefined
       }));
       result.status = params.status;
       test.expectedStatus = params.expectedStatus;
--- a/tests/playwright-test/reporter-attachment.spec.ts
+++ b/tests/playwright-test/reporter-attachment.spec.ts
@@ -143,3 +143,35 @@ test(`testInfo.attach success in fixture`, async ({ runInlineTest }) => {
   expect(result.failed).toBe(1);
   expect(stripAscii(result.output)).toContain('attachment #1: name (text/plain)');
 });
+
+test(`testInfo.attach allow empty string body`, async ({ runInlineTest }) => {
+  test.fail(true, `Currently throws TypeError: Cannot read property 'toString' of undefined: https://github.com/microsoft/playwright/issues/11413`);
+  const result = await runInlineTest({
+    'a.test.js': `
+      const { test } = pwt;
+      test('success', async ({}, testInfo) => {
+        await testInfo.attach('name', { body: '', contentType: 'text/plain' });
+        expect(0).toBe(1);
+      });
+    `,
+  });
+  expect(result.exitCode).toBe(1);
+  expect(result.failed).toBe(1);
+  expect(stripAscii(result.output)).toContain('attachment #1: name (text/plain)');
+});
+
+test(`testInfo.attach allow empty buffer body`, async ({ runInlineTest }) => {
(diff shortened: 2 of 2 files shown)


rwoll:
Looking the other reporters to see if we need to handle '' better elsewhere, too.

tests/playwright-test/reporter-raw.spec.ts
```diff
@@ -238,6 +238,24 @@ test(`testInfo.attach should save attachments via inline attachment`, async ({ r
...
+      });
+  `,
+  }, { reporter: 'dot,' + kRawReporterPath, workers: 1 }, {}, { usesCustomOutputDir: true });
+  const json = JSON.parse(fs.readFileSync(testInfo.outputPath('test-results', 'report', 'project.report'), 'utf-8'));
+  const result = json.suites[0].tests[0].results[0];
+  expect(result.attachments[0].name).toBe('example.txt');
+  expect(result.attachments[0].contentType).toBe('text/plain');
+  expect(Buffer.from(result.attachments[0].body, 'base64')).toEqual(Buffer.from(''));
+});
+
```
rwoll, line 258:
```suggestion
test(`testInfo.attach should handle empty body attachments`, async ({ runInlineTest }, testInfo) => {
  await runInlineTest({
    'a.test.js': `
      const path = require('path');
      const fs = require('fs');
      const { test } = pwt;
      test('empty', async ({}, testInfo) => {
        await testInfo.attach('example.txt', { body: '' });
      });
  `,
  }, { reporter: 'dot,' + kRawReporterPath, workers: 1 }, {}, { usesCustomOutputDir: true });
  const json = JSON.parse(fs.readFileSync(testInfo.outputPath('test-results', 'report', 'project.report'), 'utf-8'));
  const result = json.suites[0].tests[0].results[0];
  expect(result.attachments[0].name).toBe('example.txt');
  expect(result.attachments[0].contentType).toBe('text/plain');
  expect(Buffer.from(result.attachments[0].body, 'base64')).toEqual(Buffer.from(''));
});
```

Unlike the other test cases, this one may be unnecessary since raw reporter sees body's as Buffers, not strings. Let me know if you this case deleted!

dgozman: Yep, I don't think we should duplicate all attachment tests for raw reporter as well 😄

dgozman (approved):
Great find and fix! Thank you, Ross.

tests/playwright-test/reporter-raw.spec.ts
rwoll: 👍 Just removed it. (Had to do it on the CLI. First attempt via the GitHub suggestions PR UI just resulted in a single line deletion…guess it doesn't like deleting the whole chunk! hehe)

packages/playwright-test/src/dispatcher.ts
```diff
@@ -204,7 +204,7 @@ export class Dispatcher {
         name: a.name,
         path: a.path,
         contentType: a.contentType,
-        body: a.body ? Buffer.from(a.body, 'base64') : undefined
+        body: typeof a.body !== 'undefined' ? Buffer.from(a.body, 'base64') : undefined
```
pavelfeldman, line 207: nit: as of ~7 years ago, you can compare to undefined.

PR #11421, opened 2022-01-15
fix: falsey behavior in route.continue, page.post, testInfo.attach

rwoll:
In several of the Playwright APIs, falsey values were not handled correctly. This changeset adds tests (and some fixes):

1. `route.continue`: If `options.postData` was the empty string, the `continue` failed to override the post data.
2. `page.post` (`application/json` with `options.data: false|''|0|null`): Raw falsey values were getting dropped (i.e. you can't do the equivalent of `curl --header application/json … -d 'false'`). This has been fixed with most values across all browsers, but an additional fix is needed for `'null'` which the channel serializer treats extra specially.
3. `testInfo.attach`: This didn't get reported as an error when `options.path` was the empty string, but should have been.

#11413 (and its fix #11414) inspired this search as they are the same
class of bug.

Unlike #11413, this set was programmatically found with static analysis (ESLint's
`strict-boolean-expressions`) + grep filtering + manual review.
(Given JS conventions, that rule itself is too noisy to be useful on its
own as most of it would be FPs from a functional point.)

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -25,7 +25,7 @@ import zlib from 'zlib';
 import { HTTPCredentials } from '../../types/types';
 import * as channels from '../protocol/channels';
 import { TimeoutSettings } from '../utils/timeoutSettings';
-import { assert, createGuid, getUserAgent, isString, monotonicTime } from '../utils/utils';
+import { assert, createGuid, getUserAgent, monotonicTime } from '../utils/utils';
 import { BrowserContext } from './browserContext';
 import { CookieStore, domainMatches } from './cookieStore';
 import { MultipartFormData } from './formData';
@@ -583,7 +583,7 @@ function serializePostData(params: channels.APIRequestContextFetchParams, header
     }
     headers['content-type'] ??= formData.contentTypeHeader();
     return formData.finish();
-  } else if (isString(params.postData)) {
+  } else if (params.postData !== undefined) {
     headers['content-type'] ??= 'application/octet-stream';
     return Buffer.from(params.postData, 'base64');
   }
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -563,7 +563,7 @@ function isJsonParsable(value: any) {
 
 function serializePostData(params: channels.APIRequestContextFetchParams, headers: { [name: string]: string }): Buffer | undefined {
   assert((params.postData ? 1 : 0) + (params.jsonData ? 1 : 0) + (params.formData ? 1 : 0) + (params.multipartData ? 1 : 0) <= 1, `Only one of 'data', 'form' or 'multipart' can be specified`);
-  if (isString(params.jsonData)) {
+  if (params.jsonData !== undefined) {
     const json = isJsonParsable(params.jsonData) ? params.jsonData : JSON.stringify(params.jsonData);
     headers['content-type'] ??= 'application/json';
     return Buffer.from(json, 'utf8');
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -25,7 +25,7 @@ import zlib from 'zlib';
 import { HTTPCredentials } from '../../types/types';
 import * as channels from '../protocol/channels';
 import { TimeoutSettings } from '../utils/timeoutSettings';
-import { assert, createGuid, getUserAgent, monotonicTime } from '../utils/utils';
+import { assert, createGuid, getUserAgent, isString, monotonicTime } from '../utils/utils';
 import { BrowserContext } from './browserContext';
 import { CookieStore, domainMatches } from './cookieStore';
 import { MultipartFormData } from './formData';
@@ -563,7 +563,7 @@ function isJsonParsable(value: any) {
 
 function serializePostData(params: channels.APIRequestContextFetchParams, headers: { [name: string]: string }): Buffer | undefined {
   assert((params.postData ? 1 : 0) + (params.jsonData ? 1 : 0) + (params.formData ? 1 : 0) + (params.multipartData ? 1 : 0) <= 1, `Only one of 'data', 'form' or 'multipart' can be specified`);
-  if (params.jsonData) {
+  if (isString(params.jsonData)) {
     const json = isJsonParsable(params.jsonData) ? params.jsonData : JSON.stringify(params.jsonData);
     headers['content-type'] ??= 'application/json';
     return Buffer.from(json, 'utf8');
@@ -583,7 +583,7 @@ function serializePostData(params: channels.APIRequestContextFetchParams, header
     }
     headers['content-type'] ??= formData.contentTypeHeader();
     return formData.finish();
-  } else if (params.postData) {
+  } else if (isString(params.postData)) {
     headers['content-type'] ??= 'application/octet-stream';
     return Buffer.from(params.postData, 'base64');
   }
--- a/tests/browsercontext-route.spec.ts
+++ b/tests/browsercontext-route.spec.ts
@@ -212,3 +212,70 @@ it('should support the times parameter with route matching', async ({ context, p
   await page.goto(server.EMPTY_PAGE);
   expect(intercepted).toHaveLength(1);
 });
+
+it(`should overwrite post body`, async ({ context, server, page }) => {
+  server.setRoute('/load', (req, res) => {
+    res.write(`
+      <script>
+        (async () => {
+          await Promise.all([
+            fetch('/empty.html', {
+              method: 'POST',
+              body: 'original',
+            }),
+          ]);
+        })()
+      </script>
+    `);
+    res.end();
+  });
+
+  await context.route('**/empty.html', route => {
+    route.continue({
+      postData: 'replaced',
+    });
+  });
+
+  const [req] = await Promise.all([
+    server.waitForRequest('/empty.html'),
+    page.goto(server.PREFIX + '/load'),
+  ]);
+
+  const body = (await req.postBody).toString();
+  expect(body).not.toBe('original');
+  expect(body).toBe('replaced');
+});
+
+it(`should overwrite post body with empty string`, async ({ context, server, page }) => {
+  it.fail(true, 'Search for "postData ?" for likely fix locations');
+  server.setRoute('/load', (req, res) => {
+    res.write(`
+      <script>
+        (async () => {
+          await Promise.all([
+            fetch('/empty.html', {
+              method: 'POST',
+              body: 'original',
+            }),
+          ]);
+        })()
+      </script>
+    `);
+    res.end();
+  });
+
+  await context.route('**/empty.html', route => {
+    route.continue({
+      postData: '',
+    });
+  });
+
+  const [req] = await Promise.all([
+    server.waitForRequest('/empty.html'),
+    page.goto(server.PREFIX + '/load'),
+  ]);
+
+  const body = (await req.postBody).toString();
+  expect(body).not.toBe('original');
+  expect(body).toBe('');
+});
--- a/tests/browsercontext-route.spec.ts
+++ b/tests/browsercontext-route.spec.ts
@@ -245,7 +245,6 @@ it(`should overwrite post body`, async ({ context, server, page }) => {
 });
 
 it(`should overwrite post body with empty string`, async ({ context, server, page, browserName }) => {
-  it.fail(browserName === 'firefox', `There's likely a FF-specific bug in Juggler or how we ser/de to Juggler.`);
   server.setRoute('/load', (req, res) => {
     res.write(`
       <script>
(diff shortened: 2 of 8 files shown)


rwoll:
@dgozman / @pavelfeldman : Please review and let me know which of the three are legitimate test cases. Whichever end up being valid, I will file corresponding issues to follow up with fix(es) post-merge. (Didn't want to spam Issues if some of these were invalid 😄 since they aren't that high priority.)

I think the `page.post` one is definitely a bug, but the other 2 depending on how you look at it may be expected behavior (but would make an argument that they are unexpected behavior).

dgozman:
These all look like real bugs to me. Let's fix them right away - do you want to take a stab?

rwoll:
> These all look like real bugs to me. Let's fix them right away - do you want to take a stab?

Sure! I'll take a pass later today or this evening.

tests/global-fetch.spec.ts
```diff
@@ -266,16 +266,23 @@ it('should remove content-length from reidrected post requests', async ({ playwr
...
   ['number', 2021],
+  ['number (falsey)', 0],
+  ['null', null],
+  ['literal string null', 'null'],
+  ['literal string undefined', 'undefined'],
 ];
 for (const [type, value] of serialization) {
   const stringifiedValue = JSON.stringify(value);
   it(`should json stringify ${type} body when content-type is application/json`, async ({ playwright, server }) => {
+    it.fail(type === 'literal string null', `Some where isn't escaping this. Perhaps either at the following link or before: https://github.com/microsoft/playwright/blob/bfe7b7cc20ab9fdbf63dc84986369bf0f3f5a2dd/packages/playwright-core/src/protocol/serializers.ts#L53`);
```
rwoll, line 285: @dgozman This is a new test case since the last review. It remains unfixed. `'null'` (the string) gets treated by the channel serializer (https://github.com/microsoft/playwright/blob/bfe7b7cc20ab9fdbf63dc84986369bf0f3f5a2dd/packages/playwright-core/src/protocol/serializers.ts#L53) extra specially, so we'll need to do something different somewhere. I didn't dig into this particular failure enough to determine where the appropriate fix for this new case would be. Let's tackle it in a follow up Issue and PR? If you know the correct fix off the top of your head, feel free to do patch it post merge, otherwise let me know and I can dig in.

tests/browsercontext-route.spec.ts
```diff
@@ -212,3 +212,66 @@ it('should support the times parameter with route matching', async ({ context, p
...
+    page.goto(server.PREFIX + '/load'),
+  ]);
+
+  const body = (await req.postBody).toString();
+  expect(body).not.toBe('original');
+  expect(body).toBe('replaced');
+});
+
+it(`should overwrite post body with empty string`, async ({ context, server, page, browserName }) => {
+  it.fail(browserName === 'firefox', `There's likely a FF-specific bug in Juggler or how we ser/de to Juggler.`);
```
rwoll, line 248: @dgozman I don't have an FF checkout currently, so feel free to tackle this one in a follow up MR. It might end up not being Juggler, but my gut says it's in Juggler or code that is unique to talking to Juggler. The empty string is present for this `sendMayFail` call (so I suspect it's after that somewhere):

https://github.com/microsoft/playwright/blob/db0cda9df5c9ee2d28eb2cbeef3e0c04d7abb2a7/packages/playwright-core/src/server/firefox/ffNetworkManager.ts#L216

rwoll (commented):
@dgozman I had a chance to put in some fixes over lunch. I've updated the PR description, and left a few inline comments for review.

Feel free to merge and we can address the newly discovered issue with the channel serializer post-merge as well as fixup FF. (If you know those fixes off the top of your head, feel free to do, otherwise let me know and I will dig in.)

Thanks!

rwoll:
hmmm…bots look very unhappy. I must have botched something in my rebase/sqaush.

tests/browsercontext-route.spec.ts
rwoll: This should now be fixed. I coincidentally fixed while fixing something else.

tests/browsercontext-route.spec.ts
```diff
@@ -212,3 +212,65 @@ it('should support the times parameter with route matching', async ({ context, p
...
+    server.waitForRequest('/empty.html'),
+    page.goto(server.PREFIX + '/load'),
+  ]);
+
+  const body = (await req.postBody).toString();
+  expect(body).not.toBe('original');
+  expect(body).toBe('replaced');
+});
+
+it(`should overwrite post body with empty string`, async ({ context, server, page, browserName }) => {
```
rwoll, line 247:
```suggestion
it(`should overwrite post body with empty string`, async ({ context, server, page, browserName }) => {
  it.fail(browserName === 'firefox', `There's likely a FF-specific bug in Juggler or how we ser/de to Juggler.`);
```

FF is still broken here.

tests/browsercontext-route.spec.ts
rwoll: Think I now fixed this. 'twas juggler after all. Patch here: 342a49b6e4bfbc132c6858637ddc7e7570ae6748

browser_patches/firefox/BUILD_NUMBER
```diff
@@ -1,2 +1,2 @@
-1314
-Changed: lushnikov@chromium.org Tue 18 Jan 2022 02:42:08 AM PST
+1315
```
rwoll, line 1: Ran this patched build via `npm run ftest` and all passed.

rwoll:
> hmmm…bots look very unhappy. I must have botched something in my rebase/sqaush.

This looks good locally now. Needed to patch juggler for FF. Filed https://github.com/microsoft/playwright/issues/11470 to follow up with a FF rev roll once this lands.

Other than that roll, the only outstanding thing here is fixing the special treatment of `'null'` by the channel serializer (https://github.com/microsoft/playwright/pull/11421/files#r787247463), but that should be done in a post-merge follow up if deemed appropriate.

rwoll:
@dgozman This PR is now stable and ready for another review pass. Hopefully the pushes weren't too noisy!


Thanks!

tests/global-fetch.spec.ts
dgozman: I think [this function](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/fetch.ts#L550-L562) parses `'null'` as `null` and converts from a string to a null. I've discussed this with Yury and we decided to ignore this case, so let's remove it from the test :) We can leave the comment in the `isJsonParsable` function though.

tests/browsercontext-route.spec.ts
```diff
@@ -212,3 +212,66 @@ it('should support the times parameter with route matching', async ({ context, p
   await page.goto(server.EMPTY_PAGE);
   expect(intercepted).toHaveLength(1);
 });
+
+it(`should overwrite post body`, async ({ context, server, page }) => {
```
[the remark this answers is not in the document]
rwoll: Agreed! Removed! Thanks!

tests/browsercontext-route.spec.ts
```diff
@@ -212,3 +212,66 @@ it('should support the times parameter with route matching', async ({ context, p
...
+  ]);
+
+  const body = (await req.postBody).toString();
+  expect(body).not.toBe('original');
+  expect(body).toBe('replaced');
+});
+
+it(`should overwrite post body with empty string`, async ({ context, server, page, browserName }) => {
+  it.fail(browserName === 'firefox', `There's likely a FF-specific bug in Juggler or how we ser/de to Juggler.`);
+  server.setRoute('/load', (req, res) => {
```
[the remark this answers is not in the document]
rwoll: Thanks! Fixed up!

packages/playwright-core/src/server/fetch.ts
```diff
@@ -583,7 +583,7 @@ function serializePostData(params: channels.APIRequestContextFetchParams, header
     }
     headers['content-type'] ??= formData.contentTypeHeader();
     return formData.finish();
-  } else if (params.postData) {
+  } else if (isString(params.postData)) {
```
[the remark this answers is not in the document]
rwoll: 'Twas a flawed "consistency" line of thought on my behalf: I thought I was being consistent with some other [existing `isString`](https://github.com/microsoft/playwright/blob/7d924470d397975a74a19184c136b3573a974e13/packages/playwright-core/src/client/network.ts) usage in thinking it was some stylistic thing, but it turns out (as you point out) in this context functionally worse since we are deciding between `string | undefined`, not `string | object | etc.`.  Oops! Thanks!

rwoll:
@dgozman Thanks for the review, all should be addressed except w.r.t.

> I think [this function](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/fetch.ts#L550-L562) parses 'null' as null and converts from a string to a null. I've discussed this with Yury and we decided to ignore this case, so let's remove it from the test :) We can leave the comment in the isJsonParsable function though.

I'm not sure if that's the root of the issue, but I may not be following. My read is: that function looks to parse the value but only return a boolean. And in [`serializePostData`](https://github.com/microsoft/playwright/blob/7d924470d397975a74a19184c136b3573a974e13/packages/playwright-core/src/server/fetch.ts#L567), where that function is used, we then call JSON.stringify, but not JSON.parse, and I think it would have to be a `JSON.parse` call that actually returns the result that would be messing us up:

```
> JSON.stringify('\'null\'')
`"'null'"`
> JSON.stringify('null')
'"null"'
> JSON.parse(null)
null
> JSON.parse('null')
null
```

Does that hold? I can still stick a comment somewhere but wanted to make sure I wasn't misreading. Thanks!

tests/browsercontext-route.spec.ts
```diff
@@ -212,3 +212,30 @@ it('should support the times parameter with route matching', async ({ context, p
...
+              method: 'POST',
+              body: 'original',
+            });
+        })()
+      </script>
+    `),
+  ]);
+
+  const body = (await req.postBody).toString();
+  expect(body).not.toBe('original');
```
yury-s, line 239: nit: this is redundant as we check in the next line that the body is empty

dgozman:
> @dgozman Thanks for the review, all should be addressed except w.r.t.
> 
> > I think [this function](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/fetch.ts#L550-L562) parses 'null' as null and converts from a string to a null. I've discussed this with Yury and we decided to ignore this case, so let's remove it from the test :) We can leave the comment in the isJsonParsable function though.
> 
> I'm not sure if that's the root of the issue, but I may not be following. My read is: that function looks to parse the value but only return a boolean. And in [`serializePostData`](https://github.com/microsoft/playwright/blob/7d924470d397975a74a19184c136b3573a974e13/packages/playwright-core/src/server/fetch.ts#L567), where that function is used, we then call JSON.stringify, but not JSON.parse, and I think it would have to be a `JSON.parse` call that actually returns the result that would be messing us up:
> 
> ```
> > JSON.stringify('\'null\'')
> `"'null'"`
> > JSON.stringify('null')
> '"null"'
> > JSON.parse(null)
> null
> > JSON.parse('null')
> null
> ```
> 
> Does that hold? I can still stick a comment somewhere but wanted to make sure I wasn't misreading. Thanks!

As I understand it, test passes `"null"` string as post data. This string is json-parsable, so we pass it directly. Later on the test compares post data to `JSON.stringify('"null"')` which is `"\"null\""` and not `"null"`.

dgozman (approved):
The change looks good to me.

mxschmitt (changes requested):
Please extract the browser patches in an own PR which commit message starts like that:

`browser(firefox): fixed ....`

The other changes wait then until the browser was rolled, so we don't need the `it.fail` in the meantime and it's easier to track a browser only change.

rwoll:
> Please extract the browser patches in an own PR which commit message starts like that:
> 
> `browser(firefox): fixed ....`
> 
> The other changes wait then until the browser was rolled, so we don't need the `it.fail` in the meantime and it's easier to track a browser only change.

@mxschmitt Extracted into https://github.com/microsoft/playwright/pull/11529

mxschmitt:
@rwoll I landed your patch and it's actual roll into pw. After you rebase it should be all green.

rwoll:
> @rwoll I landed your patch and its actual roll into pw. After you rebase it should be all green.

Thanks @mxschmitt! Rebased and pushed. At first glance, looks like the bots window s-latest and macos-latest bots had some flakes on unrelated bits.

I think this is ready for CQ1 but if those failures are actually related, let me know and I'll take a deeper look.

Cheers! 🎉

rwoll:
@mxschmitt or @dgozman Can either of you take a look at the CQ1 results? I think the failures there are unrelated, but don't want to dismiss in case they are legit/related failures. Thanks!!

dgozman:
> @mxschmitt or @dgozman Can either of you take a look at the CQ1 results? I think the failures there are unrelated, but don't want to dismiss in case they are legit/related failures. Thanks!!

Yep, seems unrelated.

mxschmitt (approved):

PR #11468, opened 2022-01-19
fix(windows): hide child console window when running PrintDeps

### LINKED ISSUE #10802: [BUG] PrintDeps.exe ran in visible console window
**Context:**
- Playwright Version: 1.17.1
- Node.js version: 16
- Browser: Firefix

**Describe the bug**

Running some script that uses Firefox, headless, as admin user in Windows 10, and this makes playwrite pop up visible console windows running PrintDeps.exe on the screen like 10 times when a new browser is opened. I'll try to investigate further, but you might know why. Perhaps something around the util.spawnAsync() function.

yury-s:
Hide the subprocess console window when running PrintDeps and other utils on Windows

Fixes #10802

### DIFF
--- a/packages/playwright-core/src/utils/utils.ts
+++ b/packages/playwright-core/src/utils/utils.ts
@@ -241,7 +241,7 @@ function toMegabytes(bytes: number) {
 }
 
 export function spawnAsync(cmd: string, args: string[], options: SpawnOptions = {}): Promise<{stdout: string, stderr: string, code: number | null, error?: Error}> {
-  const process = spawn(cmd, args, options);
+  const process = spawn(cmd, args, Object.assign({ windowsHide: true }, options));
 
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):

54995610:
Hello Team,

We are using Playwright with Javascript for our UI automation where we are also establishing a connection to IBM Data base by using following two :-

When I am executing the script I am getting 8 different PrintDeps.exe bad image pop-ops where I have to close all those 8 pop-ups manually and that's when the execution is starting.

We are using version :- "@playwright/test": "^1.40.1".

Any help is much appreciated.

Images which we are getting
Bad image Print deps

mxschmitt:
Please file a new issue with steps on how we can reproduce it. We usually don't monitor old issues / PRs.

PR #11478, opened 2022-01-19
docs: add Python 1.18 release notes

aslushnikov:


docs/src/release-notes-python.md
```diff
@@ -5,6 +5,76 @@ title: "Release notes"
...
+
+- test your server API
+- prepare server side state before visiting the web application in a test
+- validate server side post-conditions after running some actions in the browser
+
+To do a request on behalf of Playwright's Page, use **new [`property: Page.request`] API**:
+
+```python
+// Do a GET request on behalf of page
+res = await page.request.get("http://example.com/foo.json");
```
mxschmitt, line 23:
```suggestion
res = page.request.get("http://example.com/foo.json")
```
Sync is more common than Async. We should probably provide both examples.
let's show a full example on how to read the response body and assert the status code?

docs/src/release-notes-python.md
```diff
@@ -5,6 +5,76 @@ title: "Release notes"
...
+
+Consider the following example:
+
+```python
+from playwright.async_api import Page, expect
+
+async def test_status_becomes_submitted(page: Page) -> None:
+    # ..
+    await page.click("#submit-button")
+    await expect(page.locator(".status")).to_have_text("Submitted")
```
mxschmitt, line 40: the vast majority of people uses sync, so we should also provide a sync example.

docs/src/release-notes-python.md
```diff
@@ -5,6 +5,76 @@ title: "Release notes"
...
+reached. You can pass this timeout as an option.
+
+Read more in [our documentation](./api/class-playwrightassertions).
+
+### Locator Improvements
+
+- [`method: Locator.dragTo`]
+- Each locator can now be optionally filtered by the text it contains:
+    ```python
+    await page.locator("li", has_text="my item")).locator("button").click();
```
mxschmitt, line 55: we need sync there too.

aslushnikov: done

docs/src/release-notes-python.md
aslushnikov: done

docs/src/release-notes-python.md
aslushnikov: done

mxschmitt (approved):

PR #11516, opened 2022-01-20
fix: playwright.locator.hasText RegExp flag serialisation in Console API

mxschmitt:


aslushnikov (approved):
lgtm % tests (if we have any here)

PR #11537, opened 2022-01-21
move install script to `@playwright/core`

### LINKED ISSUE #11531: [BUG] `playwright` package has `install` script but `@playwright/test` doesn't
the `install` script [defined here](https://github.com/microsoft/playwright/blob/main/packages/playwright/package.json#L26) is useful because it [gets automatically run when you run `npm install playwright`](https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-ci)

however `@playwright/test` doesn't seem to have an `install` script, meaning you have to manually run `playwright install` after installing the package, which is annoying

DetachHead:
Fixes #11531


mxschmitt:
Thank you for your contribution! This is working as intended see here https://github.com/microsoft/playwright/issues/11531#issuecomment-1018368760.

PR #11579, opened 2022-01-23
fix(prepare): don't change package files on prepare

pavelfeldman:


.github/workflows/infra.yml
```diff
@@ -23,6 +23,7 @@ jobs:
     - run: npm ci
     - run: npm run build
     - run: npx playwright install-deps
+    - run: npx playwright install
```
mxschmitt, line 26:
```suggestion
    - run: npx playwright install --with-deps
```
nit: remove the file above it.

packages/playwright-webkit/install.js
```diff
@@ -14,6 +14,13 @@
  * limitations under the License.
  */
 
-const { installBrowsersForNpmInstall } = require('playwright-core/lib/utils/registry');
+let installDefaultBrowsersForNpmInstall;
 
-installBrowsersForNpmInstall(['webkit']);
+try {
+  installDefaultBrowsersForNpmInstall = require('playwright-core/lib/utils/registry');
```
mxschmitt, line 20: This looks different to how it was before. Install... is a named export rather than a default export.

.github/workflows/infra.yml
pavelfeldman: I actually want to see groups for those in the UI!

packages/playwright-webkit/install.js
pavelfeldman: Ouch, good catch!

mxschmitt (approved):

mxschmitt:
Might be a cherry-pick 1.18 candidate?

PR #11607, opened 2022-01-25
fix `PWDEBUG=0` being treated as true

### LINKED ISSUE #11606: [BUG] setting `PWDEBUG` to `0` is treated as true
https://github.com/microsoft/playwright/blob/c510819407d91975340bee1249502c51c6a3bde9/packages/playwright-core/src/utils/utils.ts#L324

[the docs](https://playwright.dev/docs/next/debug#run-in-debug-mode) say setting `PWDEBUG` to `1` enables the playwright inspector, so one would assume setting it to `0` disables it

DetachHead:
Fixes #11606

### DIFF
--- a/packages/playwright-core/src/utils/utils.ts
+++ b/packages/playwright-core/src/utils/utils.ts
@@ -327,7 +327,7 @@ export function debugMode() {
     return 'console';
   if (!debugEnv || debugEnv === '0')
     return '';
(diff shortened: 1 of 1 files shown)


mxschmitt:
Thank you for your contribution! One of our peers unfortunately already went ahead and fixed it in #11611.

PR #11613, opened 2022-01-25
fix(server): handle `=` in cookie values correctly

### LINKED ISSUE #11612: [BUG] APIRequestContext omits trailing `=` in cookie values
**Context:**
- Playwright Version: 1.17.2
- Operating System: macOS
- Node.js version: v17.4.0
- Browser: Chromium

**Code Snippet**

I am making two requests using a new APIRequestContext:

```javascript
export async function loginAccountWithPassword(
  user: IdentityWithPassword,
  fixture: string,
  baseUrl: string = routes.kratos.publicBaseUrl
): Promise<void> {
  const r = await request.newContext()
  const result = await r
    .get(baseUrl + '/api/kratos/public/self-service/login/browser', {
      headers: {
        Accept: 'application/json'
      }
    })
    .then<SelfServiceLoginFlow>(async (res) => {
      expect(res.status()).toEqual(200)
      await r.storageState({ path: fixture })
      return res.json()
    })
    .then((flow) => {
      const form = {
        password_identifier: user.traits.email,
        password: user.password,
        method: 'password',
        csrf_token: findCsrfToken(flow.ui)
      }
      console.log('sending', form)
      return r
        .post(flow.ui.action, {
          headers: {
            Accept: 'application/json'
          },
          data: form
        })
        .then(async (res) => {
          const result = await res.json()
          console.log(result)
          expect(res.status()).toEqual(200)
          expect(result.session.identity.traits.email).toEqual(
            user.traits.email
          )
          return result
        })
    })

  await r.storageState({ path: fixture })
  await r.dispose()

  return result
}
```

**Describe the bug**

The second requests fails because the cookie sent as part of the second request is dropping the trailing `=`. Using `DEBUG=pw:api`, I can see the following trace:

```
  pw:api => aPIRequestContext.get started +0ms
  pw:api → GET https://angry-nash-aaaxsdan5t.projects.oryapis:8080/api/kratos/public/self-service/login/browser +1ms
  pw:api   user-agent: Playwright/1.17.2 +0ms
  pw:api   accept: application/json +0ms
  pw:api   accept-encoding: gzip,deflate,br +0ms
  pw:api ← 200 OK +46ms
  pw:api   cache-control: private, no-cache, no-store, must-revalidate +0ms
  pw:api   ...
  pw:api   set-cookie: csrf_token_dc5a55bab395766eaeba4dc60f1e5f254ef2d35fcb7c7b10c96136a0148cab56=Ixd5OPbRNqvtuQl47GKuMdubeJKgUsYlpDuMD7AjawY=; Path=/; Domain=angry-nash-aaaxsdan5t.projects.oryapis; Max-Age=31536000; HttpOnly; SameSite=Lax +0ms
  pw:api   vary: Origin,Cookie +0ms
  pw:api   ...

  pw:api <= aPIRequestContext.get succeeded +1ms
  pw:api => aPIRequestContext.storageState started +1ms
  pw:api <= aPIRequestContext.storageState succeeded +4ms
  pw:api => aPIResponse.json started +0ms
  pw:api <= aPIResponse.json succeeded +0ms

  pw:api => aPIRequestContext.post started +3ms
  pw:api → POST https://angry-nash-aaaxsdan5t.projects.oryapis:8080/api/kratos/public/self-service/login?flow=fb00b320-fc3a-479a-ada7-6757ce77dcbe +1ms
  pw:api   ...
  pw:api   cookie: csrf_token_dc5a55bab395766eaeba4dc60f1e5f254ef2d35fcb7c7b10c96136a0148cab56=Ixd5OPbRNqvtuQl47GKuMdubeJKgUsYlpDuMD7AjawY +0ms
  pw:api ← 403 Forbidden +11ms
  pw:api   ...
```

As you can see, the `Set-Cookie` header of the first request sets the following cookie (notice the trailing `=`):

```
set-cookie: csrf_token_dc5a55bab395766eaeba4dc60f1e5f254ef2d35fcb7c7b10c96136a0148cab56=Ixd5OPbRNqvtuQl47GKuMdubeJKgUsYlpDuMD7AjawY=
```

The `Cookie` header of the second request however omits that trailing `=`:

```
cookie: csrf_token_dc5a55bab395766eaeba4dc60f1e5f254ef2d35fcb7c7b10c96136a0148cab56=Ixd5OPbRNqvtuQl47GKuMdubeJKgUsYlpDuMD7Ajaw
```

The JSON being stored using `r.storageState({ path: fixture })` is also missing the trailing `=` in the value:

```
{
  "cookies": [
    {
      "name": "csrf_token_dc5a55bab395766eaeba4dc60f1e5f254ef2d35fcb7c7b10c96136a0148cab56",
      "value": "Ixd5OPbRNqvtuQl47GKuMdubeJKgUsYlpDuMD7AjawY",
      "domain": ".angry-nash-aaaxsdan5t.projects.oryapis",
      "path": "/",
      "expires": 1674658967.931,
      "httpOnly": true,
      "secure": false,
      "sameSite": "Lax"
    }
  ],
  "origins": []
}
```

I assume that there's an issue with potentially string splitting here. Please note that this is not an issue in regular browser contexts.

aeneasr:
This patch resolves an issue where using `=` in cookie values would cause the cookie value to contain only the partial value.

Closes https://github.com/microsoft/playwright/issues/11612

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -517,7 +517,21 @@ function toHeadersArray(rawHeaders: string[]): types.HeadersArray {
 const redirectStatus = [301, 302, 303, 307, 308];
 
 function parseCookie(header: string): types.NetworkCookie | null {
-  const pairs = header.split(';').filter(s => s.trim().length > 0).map(p => p.split('=').map(s => s.trim()));
+  const pairs = header.split(';').filter(s => s.trim().length > 0).map(p => {
+    let key = '';
+    let value = '';
+    const separatorPos = p.indexOf('=');
+    if (separatorPos === -1) {
+      // If only a key is specified, the value is left undefined.
+      key = p.trim();
+    } else {
+      // Otherwise we assume that the key is the element before the first `=`
+      key = p.slice(0, separatorPos).trim();
+      // And the value is the rest of the string.
+      value = p.slice(separatorPos + 1).trim();
+    }
+    return [key, value];
+  });
   if (!pairs.length)
     return null;
   const [name, value] = pairs[0];
--- a/tests/global-fetch-cookie.spec.ts
+++ b/tests/global-fetch-cookie.spec.ts
@@ -166,6 +166,33 @@ it('should remove expired cookies', async ({ request, server }) => {
   expect(serverRequest.headers.cookie).toBe('a=v');
 });
 
+it('should store cookie from Set-Cookie header even if it contains equal signs', async ({ request, server }) => {
+  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/11612' });
+
+  server.setRoute('/setcookie.html', (req, res) => {
+    res.setHeader('Set-Cookie', ['f=value == value=; secure; httpOnly; path=/some=value']);
+    res.end();
+  });
+
+  await request.get(`http://a.b.one.com:${server.PORT}/setcookie.html`);
+  const state = await request.storageState();
+  expect(state).toEqual({
+    'cookies': [
+      {
+        domain: 'a.b.one.com',
+        expires: -1,
+        name: 'f',
+        path: '/some=value',
+        sameSite: 'Lax',
+        httpOnly: true,
+        secure: true,
+        value: 'value == value=',
+      }
+    ],
+    'origins': []
+  });
+});
+
 it('should export cookies to storage state', async ({ request, server }) => {
   const expires = new Date('12/31/2100 PST');
   server.setRoute('/setcookie.html', (req, res) => {


aeneasr:
Not sure why the tests fails, they do not seem related to this change?

packages/playwright-core/src/server/fetch.ts
```diff
@@ -517,7 +517,11 @@ function toHeadersArray(rawHeaders: string[]): types.HeadersArray {
 const redirectStatus = [301, 302, 303, 307, 308];
 
 function parseCookie(header: string): types.NetworkCookie | null {
-  const pairs = header.split(';').filter(s => s.trim().length > 0).map(p => p.split('=').map(s => s.trim()));
+  const pairs = header.split(';').filter(s => s.trim().length > 0).map(p => {
+    const value = p.slice(p.indexOf('=') + 1).trim();
+    const key = p.slice(0, p.indexOf('=')).trim();
```
yury-s, line 522: This will not work for the attributes that have only name (no value and no '='), e.g. `HttpOnly`. Please fix this and extend your test case to cover this.

tests/global-fetch-cookie.spec.ts
```diff
@@ -67,6 +67,20 @@ it('should store cookie from Set-Cookie header', async ({ request, server }) =>
   expect(serverRequest.headers.cookie).toBe('c=d');
 });
 
+it('should store cookie from Set-Cookie header with trailing equal sign', async ({ request, server }) => {
+  // https://github.com/microsoft/playwright/issues/11612
```
yury-s, line 71: let's convert it to the following annotation(we started using this form recently) :

```js
 it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/11612' });
```

yury-s (changes requested):
Thanks for the comment, the fix looks good modulo the problem with name-only attributes.

packages/playwright-core/src/server/fetch.ts
aeneasr: Thank you for the fast review! I will!

Do you have a tip for me where I can execute this test only? Running the full test suite locally causes a lot of test failures. Thank you :)

aeneasr: Nevermind, this does the trick:

```
npm run ctest -- tests/global-fetch-cookie.spec.ts
```

aeneasr: Would it be acceptable to export `parseCookie` and test it individually? It seems a bit difficult to test it well given how it is integrated

yury-s: We normally don't do unit tests. You can read parsed cookie via `request.storageState()`, similar to this test: https://github.com/microsoft/playwright/blob/6a7a2971f2e40442d720f42412d299c504fcbd71/tests/global-fetch-cookie.spec.ts#L175-L188

packages/playwright-core/src/server/fetch.ts
```diff
@@ -517,7 +517,10 @@ function toHeadersArray(rawHeaders: string[]): types.HeadersArray {
 const redirectStatus = [301, 302, 303, 307, 308];
 
 function parseCookie(header: string): types.NetworkCookie | null {
-  const pairs = header.split(';').filter(s => s.trim().length > 0).map(p => p.split('=').map(s => s.trim()));
+  const pairs = header.split(';').filter(s => s.trim().length > 0).map(p => {
+    const [key, ...rest] = p.split('=');
+    return [key.trim(), rest.map(s => s.trim()).join('=')];
```
yury-s, line 522: This will break on `a=b == c`, I also liked the previous version of this code more as it didn't do unnecessary split/join, perhaps you could resurrect that and just add a check if '=' is present?

yury-s (changes requested):

packages/playwright-core/src/server/fetch.ts
aeneasr: Sorry, I just pushed a change which fixes the trim issue. I also added a test for it! Is that solution still acceptable? It's basically like the one we had before, it just handles the rest of the values a bit differently.

packages/playwright-core/src/server/fetch.ts
aeneasr: Thank you, that was very helpful!

packages/playwright-core/src/server/fetch.ts
yury-s: I liked the version without extra `.split` calls more as the intent of the code was more clear, I'd rather you just modified it to something like the following
```js
  let key;
  let value;
  const separatorPos = p.indexOf('=');
  if (separatorPos === -1) {
    key = p.trim();
  } else {
    key = p.slice(0, separatorPos).trim();
    value = p.slice(separatorPos + 1).trim();
  }
  return [key, value];
```
but it is up to you.

yury-s (approved):

packages/playwright-core/src/server/fetch.ts
aeneasr: Ah indeed, it's definitely more readable! I adjusted your version and added a few comments to make clear what this is for

yury-s (approved):
lgtm, but looks like now the lint is unhappy (not sure why it didn't complain before your change), initializing key/value with empty string should help I believe.

aeneasr:
Thanks, resolved :)

yury-s (approved):

aeneasr:
Thank you @yury-s for help, rewview & merge :) I looked around a bit but couldn't find some type of nightly build that might include this change. Is there anything available to using nightly / builds from main? If not it would be awesome if this fix could be in the next patch release :)

aeneasr:
Nevermind, it appears there are nightlies on NPM :)

PR #11614, opened 2022-01-25
fix(chromium): close all javascript dialogs when closing context

### LINKED ISSUE #11581: [BUG] context.close() doesn't close stray onbeforeunload dialog in chromium
**Context:**
- Playwright Version: 1.17.2
- Operating System: Mac
- Python version: Python 3.9.10
- Browser: All

**Code Snippet**

```python

import pytest

HEADLESS = True
HEADLESS = False

WAIT = 5000

@pytest.fixture(scope='session', params=('firefox', 'chromium', 'webkit'))
def browser_vendor(request):
    yield request.param

@pytest.fixture(scope='session')
def browser(browser_vendor):
    with sync_playwright() as playwright:
        browser = getattr(playwright, browser_vendor)
        instance = browser.launch(headless=HEADLESS)
        yield instance
        instance.close()

@pytest.fixture
def context(browser):
    context = browser.new_context()
    yield context
    context.close()

@pytest.fixture
def page(context):
    page = context.new_page()
    page.set_default_timeout(WAIT)
    yield page
    page.close()

html = '''
<form>
    <label for=input_id id=label>input_label</label>
    <input id=input_id class=input_class name=input_name value=input_value 
        title=input_title placeholder=input_placeholder aria-label=input_aria_label>
    <div id=div_id>div_text</div>
</form>
'''

ask_to_leave_script = '''
// interestingly Firefox webdriver doesn't show thes dialogs at all, even though this code works in normal Firefox
window.addEventListener('beforeunload', function (e) {
    // Cancel the event
    e.preventDefault(); // mozilla will now always show dialog
    // Chrome requires returnValue to be set
    e.returnValue = 'Fnord';
});
'''

def test_before_unload(page):
    # delay unload
    page.set_content(html)

    page.evaluate(ask_to_leave_script)
    # page needs a change otherwise the onbeforeunload doesn't trigger
    page.fill('text=input_label', value='fnord')
    
    page.click("body")

    with page.expect_event("dialog") as dialog_info:
        page.close(run_before_unload=True)
    dialog = dialog_info.value
    assert dialog.type == "beforeunload"

    # with page.expect_event("close"):
    #     dialog.accept()
    
    context = page.context
    context.close()
    # browser = context.browser
    # browser.close()
```

**Describe the bug**

When dealing with stray leftover dialogs in (often background windows) Chromium doesn't behave the same as the other browsers, in that it hangs forever in `context.close()` while Safari and FF close the context just fine. 

It is my understanding that context.close() is what should give test isolation and is what should happen between tests. In any case, the three browsers should behave the same way.

The workaround of course, is to just close the dialog programmatically. But this is still annoying.

aslushnikov:
Fixes #11581

### DIFF
--- a/packages/playwright-core/src/server/chromium/crBrowser.ts
+++ b/packages/playwright-core/src/server/chromium/crBrowser.ts
@@ -471,15 +471,16 @@ export class CRBrowserContext extends BrowserContext {
 
   async _doClose() {
     assert(this._browserContextId);
+    // We have to close all pending beforeUnload dialogs
     while (true) {
-      const pendingDialogs = [];
+      const pendingBeforeUnloadDialogs = [];
       for (const crPage of this._browser._crPages.values()) {
         if (crPage._browserContext === this)
-          pendingDialogs.push(...[...crPage._pendingDialogs]);
+          pendingBeforeUnloadDialogs.push(...[...crPage._pendingBeforeUnloadDialogs]);
       }
-      if (!pendingDialogs.length)
+      if (!pendingBeforeUnloadDialogs.length)
         break;
-      await Promise.all(pendingDialogs.map(dialog => dialog.dismiss()));
+      await Promise.all(pendingBeforeUnloadDialogs.map(dialog => dialog.dismiss()));
     }
     await this._browser._session.send('Target.disposeBrowserContext', { browserContextId: this._browserContextId });
     this._browser._contexts.delete(this._browserContextId);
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/chromium/crBrowser.ts
```diff
@@ -471,6 +472,14 @@ export class CRBrowserContext extends BrowserContext {
 
   async _doClose() {
     assert(this._browserContextId);
+    const openedBeforeUnloadDialogs: Dialog[] = [];
```
dgozman, line 475: Let's comment that headed chrome does not close with beforeunload open, but not with any other dialog types, and we do not expect any new beforeunload dialogs because our close does not trigger them.

dgozman (approved):

PR #11643, opened 2022-01-26
docs(dotnet): bundle driver for different platforms

### LINKED ISSUE #1969: [docs]: document bundled driver changes
Ref: https://github.com/microsoft/playwright-dotnet/pull/1955

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/1969
Relates https://github.com/microsoft/playwright-dotnet/pull/1955

cc @campersau

### DIFF
--- a/docs/src/intro-csharp.md
+++ b/docs/src/intro-csharp.md
@@ -122,6 +122,26 @@ dotnet test -- NUnit.NumberOfTestWorkers=5
 pwsh bin\Debug\netX\playwright.ps1 codegen
 ```
 
+## Bundle drivers for different platforms
+
+Playwright by default does bundle only the driver for the .NET publish target runtime. If you want to bundle for additional platforms, you can
+override this behavior by using either `all`, `none` or `linux`, `win`, `osx` in your project file.
+
+```xml
+<PropertyGroup>
+  <PlaywrightPlatform>all</PlaywrightPlatform>
+</PropertyGroup>
+```
+
+or:
+
+```xml
+<PropertyGroup>
+  <PlaywrightPlatform>osx</PlaywrightPlatform>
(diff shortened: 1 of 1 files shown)


docs/src/intro-csharp.md
```diff
@@ -122,6 +122,26 @@ dotnet test -- NUnit.NumberOfTestWorkers=5
...
+  <PlaywrightPlatform>all</PlaywrightPlatform>
+</PropertyGroup>
+```
+
+or:
+
+```xml
+<PropertyGroup>
+  <PlaywrightPlatform>osx</PlaywrightPlatform>
+  <PlaywrightPlatform>linux</PlaywrightPlatform>
```
campersau, line 141: Properties are just strings, so something like this works, [because the checks are using `Contains`](https://github.com/microsoft/playwright-dotnet/pull/1955/files#diff-e3082407b006296971300d96ce43253f11b1754c334c6b7e7b81191d0a5a37fbR11):

```suggestion
  <PlaywrightPlatform>osx;linux</PlaywrightPlatform>
```

docs/src/intro-csharp.md
```diff
@@ -122,6 +122,25 @@ dotnet test -- NUnit.NumberOfTestWorkers=5
 pwsh bin\Debug\netX\playwright.ps1 codegen
 ```
 
+## Bundle drivers for different platforms
```
pavelfeldman, line 125: I don't think it needs to me this high in the list, probably closer to the end, say that "By default, Playwright will only bundle..."

pavelfeldman (approved):
lgtm % comments

docs/src/intro-csharp.md
mxschmitt: it can't be lower.

PR #11712, opened 2022-01-28
fix: run reporter.onEnd after webserver teardown

### LINKED ISSUE #11647: [Question] Should launched webservers close after auto viewing html report?
I'm new to playwright, so just getting started, but I seem to have run into an issue, and I don't know if it's a bug or not.

I'm running the first test provided by the auto setup tool and as per the getting started guide, checking the title.

This fails (as expected) on my app because the title is different.

I have set in the config for a web server to start, which works as expected.

However, when a test fails and an HTML report is created, which automatically opens... when closing the report, and ending the test process as instructed, it doesn't tear down the web server it spun up. This blocks the port, and so the tests cannot be re-run.

As a work around, I've set the html reporter to never open, as I can open it manually if required.

Is this expected behaviour or an oversight?
I'd have expected that when the report is generated, the web server would be shut down, unless `reuseExistingServer` is set.

I had to manually kill the node processes for the server to free up the port.

aslushnikov:
Fixes #11647

### DIFF
--- a/packages/playwright-test/src/runner.ts
+++ b/packages/playwright-test/src/runner.ts
@@ -144,6 +144,8 @@ export class Runner {
         else
           this._reporter.onError?.(createStacklessError(`Timed out waiting ${config.globalTimeout / 1000}s for the entire test run`));
         return actualResult;
+      } else if (this._didBegin) {
+        await this._reporter.onEnd?.(result.result);
       }
       return result.result;
     } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/runner.ts
```diff
@@ -179,7 +179,10 @@ export class Runner {
 
   private async _run(list: boolean, testFileReFilters: FilePatternFilter[], projectNames?: string[]): Promise<FullResult> {
     const filesByProject = await this._collectFiles(testFileReFilters, projectNames);
-    return await this._runFiles(list, filesByProject, testFileReFilters);
+    const result = await this._runFiles(list, filesByProject, testFileReFilters);
+    if (this._didBegin)
+      await this._reporter.onEnd?.(result);
```
dgozman, line 184: - Let's move this to `_runAllTests` instead, because it has one more onEnd so it would be better to handle all onEnd calls in one place.
- A test would also be nice (just console.log from globalteardown and onEnd).

aslushnikov: done

aslushnikov:
For the record: Dmitry LGTM'ed this in telegram chat :)

image

PR #11735, opened 2022-01-29
feat: support clarification message for expect

### LINKED ISSUE #7816: [Question]: Is there any way to define custom assertion error message with 'expect' function in Playwright?
### Your question

I want to define custom assertion error with 'expect' function so that whenever the assertion is getting failed then I can use that custom error message. Is there any way to do this in Playwright?

Linking:
- https://github.com/microsoft/playwright/pull/4778
- https://github.com/microsoft/playwright/issues/4665

aslushnikov:
The clarification message is displayed in the HTML report as the name of the step:

image

It is also shown in terminal output:

image

Fixes #7816

### DIFF
--- a/packages/playwright-test/src/expect.ts
+++ b/packages/playwright-test/src/expect.ts
@@ -92,37 +92,12 @@ export const printReceivedStringContainExpectedResult = (
 
 // #endregion
 
-class ExpectMetaInfoProxy {
-  private _message: string;
-
-  constructor(message: string) {
-    this._message = message;
-  }
-
-  get(target: any, prop: any, receiver: any): any {
-    const value = Reflect.get(target, prop, receiver);
-    if (typeof value !== 'function')
-      return new Proxy(value, this);
-    return (...args: any[]) => {
-      try {
-        expectCallMetaInfo = {
-          message: this._message,
-        };
-        const result = value.call(target, ...args);
-        return result;
-      } finally {
-        expectCallMetaInfo = undefined;
-      }
-    };
-  }
-}
-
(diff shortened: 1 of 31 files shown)


pavelfeldman:
I'd suggest following UX when this syntax is used:

```
Error: must be six!

Expected: 6
Received: 5
```

I.e. drop the `// Object.is equality` line altogether

Same goes for HTML reporter, let's drop expect.toBe, I already named this check.

packages/playwright-test/src/expect.ts
```diff
@@ -92,7 +92,40 @@ export const printReceivedStringContainExpectedResult = (
...
+        expectCallMetaInfo = undefined;
+      }
+    };
+  }
+}
+
+export const expect: Expect = new Proxy(expectLibrary as any, {
+  apply: function(target: any, thisArg: any, argumentsList: [actual: unknown, message: string|undefined]) {
+    const message: string = argumentsList[1] || '';
+    if (typeof message !== 'string')
```
dgozman, line 123: Check for undefined and string explicitly, otherwise I can pass e.g. null.

packages/playwright-test/src/expect.ts
```diff
@@ -92,7 +92,40 @@ export const printReceivedStringContainExpectedResult = (
...
+      throw new Error('Second `expect` argument must be string!');
```
dgozman, line 124: `expect(actual, optionalErrorMessage): optional error message must be a string.`

dgozman (approved):

PR #11750, opened 2022-01-31
feat: compute md5 hash for CDN uploads.

### LINKED ISSUE #10173: [BUG] Installing firefox fails because maintenanceservice.exe gets flagged as malware
**Context:**
- Playwright Version: 1.16.3
- Operating System: Windows Server 2019 (17763.2268)
- Node.js version: v16.13.0
- Browser: Firefox
- Extra: System has Bitdefender Endpoint Security Tools installed

**Code Snippet**

```powershell
❯ playwright install firefox
Removing unused browser at ~~redacted~~\pw_browsers\firefox-1297
Downloading Playwright build of firefox v1297 - 74 Mb [====================] 100% 0.0s
Failed to install browsers
Error: Failed to download firefox v1297, caused by
Error: EPERM: operation not permitted, open '~~redacted~~\pw_browsers\firefox-1297\firefox\maintenanceservice.exe'
```

**Describe the bug**

Since Firefox-1297, I was unable to complete the installation for Firefox binaries because Bitdefender is flagging it as malware. I didn't have this problem in binary packages Firefox-1295 and earlier.

To further verify this issue, I grabbed the maintenanceservice binary from another system with no Bitdefender and uploaded it to VirusTotal, and the [results](https://www.virustotal.com/gui/file/109e1bb86f1e2793efdfea1267792cfb42851bdc83b3d53e21644f4eb2304d5e/detection) show it is flagged by 9 other antiviruses as `Gen:Variant.Razy` malware.

I also tried sending the similar binary from the actual Firefox installation and the [results](https://www.virustotal.com/gui/file/333196c33bae6a2b513e39e24bf2404a77bf93c470d685c1e2e43a0351026b67) are clean.

It looks like a case of false-positive detection, but are there options where I can verify the integrity of the downloaded binaries? I would need it to make a case to our IT security that it is fine to whitelist this file/folder.

aslushnikov:
By default, azure auto-computes MD5 for all blobs < 64MBs.
However, many of our binaries exceed 64MB so md5 has to be computed
manually.

With this patch, MD5 hash will be set to all our CDN uploads and could
be retrieved as `content-md5` header, e.g.:

```bash
$ curl -I https://playwright.azureedge.net/builds/ffmpeg/1001/ffmpeg-win32.zip
```

Fixes #10173

### DIFF
--- a/browser_patches/upload.sh
+++ b/browser_patches/upload.sh
@@ -43,6 +43,16 @@ if [[ ("$2" == '--check') ]]; then
   fi
 fi
 
+GENERATE_MD5_HASH=$(cat <<EOF
+  const crypto = require('crypto');
+  const buffer = fs.readFileSync(process.argv[1]);
+  console.log(crypto.createHash('md5').update(buffer).digest("base64"));
+EOF
+)
+
+MD5_HASH=$(node -e "${GENERATE_MD5_HASH}" "${ZIP_PATH}")
+echo "MD5 hash: ${MD5_HASH}"
+
 if ! [[ -f $ZIP_PATH ]]; then
   echo "ERROR: ${ZIP_PATH} does not exist"
   exit 1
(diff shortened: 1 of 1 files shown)


browser_patches/upload.sh
```diff
@@ -43,6 +43,16 @@ if [[ ("$2" == '--check') ]]; then
   fi
 fi
 
+GENERATE_MD5_HASH=$(cat <<EOF
+  const crypto = require('crypto');
+  const buffer = fs.readFileSync(process.argv[1]);
```
mxschmitt, line 48: you forgot to import 'fs'.

browser_patches/upload.sh
```diff
@@ -43,6 +43,16 @@ if [[ ("$2" == '--check') ]]; then
...
+  console.log(crypto.createHash('md5').update(buffer).digest("base64"));
```
mxschmitt, line 49:
```suggestion
  console.log(crypto.createHash('md5').update(buffer).digest('base64'));
```

yury-s (approved):

PR #11758, opened 2022-01-31
feat(breaking): always report onBegin/onEnd, report file errors

pavelfeldman:


docs/src/test-reporter-api/class-suite.md
```diff
@@ -24,6 +24,11 @@ Reporter is given a root suite in the [`method: Reporter.onBegin`] method.
 
 Returns the list of all test cases in this suite and its descendants, as opposite to [`property: Suite.tests`].
 
+## property: Suite.loadErrors
```
dgozman, line 27: I don't see who uses this, perhaps a test would be nice?

packages/playwright-test/src/runner.ts
```diff
@@ -237,148 +227,194 @@ export class Runner {
...
-    if (!list) {
-      for (const internalGlobalSetup of this._internalGlobalSetups)
-        internalGlobalTeardowns.push(await internalGlobalSetup());
+
+    const fatalErrors: TestError[] = [];
+
+    // 1. Add all tests.
+    const preprocessRoot = new Suite('');
+    for (const file of allTestFiles) {
+      const fileSuite = await this._loader.loadTestFile(file);
```
dgozman, line 236: It is somewhat scary that we continue loading files after an error, because this could trigger cascading errors.

For example:
```js
// When "base" module did not load because of syntax error or something,
// "Base" will be undefined and extending it will fail with a cascading error.

const { Base } = require('./base');
const Derived extends Base {}
```

Also, if I have a syntax error in the fixtures file, every single test file will get that error reported, won't it?

PR #11793, opened 2022-02-01
chore: pin production dependencies

mxschmitt:


aslushnikov (approved):

aslushnikov:
Let's make sure that built packages actually do not have any carets in their package.json's. (this is to make sure we covered all places where we do automation for package.json's)

mxschmitt:
drive-by: fixing service bots after the upgrade of the `ws` module.

barak007:
Hey guys. 
What is the reason behind pinning all the versions?
It makes things worse for us by forcing duplicate dependencies. I acknowledge the intention to make the **direct dependencies** more stable, but it also has some downsides:
* Pining the the first level of dependencies does not work for transitive dependencies. 
* Consumers of this library stop getting bugs/security fixes from upstream until you release version bumping for those deps

Would you be willing to re-consider this approach?

@mxschmitt 
@aslushnikov

mxschmitt:
Hey, the answer is pretty similar to https://github.com/vercel/next.js/issues/6465#issuecomment-467813088 since Playwright got broken a bunch of times already (3 times which we know) because of it's dependencies we decided to pin it for now. In the future we are considering a bundled approach to fully include all dependencies into the NPM package itself instead of relying on NPM for that. We ensure that the dependencies are staying up to date.

barak007:
So you still "welcome" transitive dependencies breakages for now, so I'm not sure how different is it to pin you likely have more of them then direct dependencies.

`bundled approach` - That's to me worse solution, just hiding what's being used.
`We ensure that the dependencies are staying up to date.` - that's already got outdated.

barak007:
I'm not sure how it sounds but I think for the eco system it's better keep the semver approach.

https://i.redd.it/tfugj4n3l6ez.png

PR #11800, opened 2022-02-02
feat(test-runner): support expect.soft

### LINKED ISSUE #7819: [Feature]: Soft/queued asserts/expects so that it does not fail after the first one
### Your question

I have a scenario where we have multiple scenarios to assert. I want to finish off all assertions irrespective of their status and need to see what all assertions got passed or failed once all validations steps were done. 
Is there any way to do this in Playwright?

aslushnikov:
Soft expects will still fail the test, but will not abort it's execution. As a consequence of this:
-  `TestResult` now might have multiple errors, which is reflected with a new `testResult.erros: TestError[]` field.
- `TestInfo` now might have multiple errors as well, which is reflected with a new `testInfo.errors: TestError[]` field.

This is how multiple `expect.soft` look in HTML report:

image

These are the same soft errors in terminal:

image
 
Fixes #7819

### DIFF
--- a/docs/src/test-assertions-js.md
+++ b/docs/src/test-assertions-js.md
@@ -19,10 +19,21 @@ expect(value).not.toEqual(0);
 await expect(locator).not.toContainText("some text");
 ```
 
-You can also specify a custom error message as a second argument to the `expect` function, for example:
+By default, failed assertion will terminate test execution. Playwright also
+supports *soft assertions*: failed soft assertions **do not** terminate test execution,
+but mark the test as failed.
+
+```js
+expect.soft(1+1).toBe(3);
+expect.soft(2*2).toBe(5);
+```
+
+
+You can specify a custom error message as a second argument to the `expect` function, for example:
 
 ```js
 expect(value, 'my custom error message').toBe(42);
+expect.soft(value, 'my soft assertion').toBe(56);
 ```
 
 <!-- TOC -->
--- a/docs/src/test-assertions-js.md
+++ b/docs/src/test-assertions-js.md
@@ -24,10 +24,14 @@ supports *soft assertions*: failed soft assertions **do not** terminate test exe
 but mark the test as failed.
 
 ```js
-expect.soft(1+1).toBe(3);
-expect.soft(2*2).toBe(5);
-```
+// Make a few checks that will not stop the test when failed...
+await expect.soft(page.locator('#status')).toHaveText('Success');
+await expect.soft(page.locator('#eta')).toHaveText('1 day');
 
+// ... and continue the test to check more things.
+await page.locator('#next-page').click();
+await expect.soft(page.locator('#title')).toHaveText('Make another order');
+```
 
 You can specify a custom error message as a second argument to the `expect` function, for example:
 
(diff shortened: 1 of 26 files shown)


packages/playwright-test/src/util.ts
```diff
@@ -202,7 +202,7 @@ export function getContainedPath(parentPath: string, subPath: string = ''): stri
 
 export const debugTest = debug('pw:test');
 
-export function prependToTestError(testError: TestError | undefined, message: string | undefined, location?: Location) {
+export function prependToTestError(testError: TestError, message: string | undefined, location?: Location): TestError {
   if (!message)
     return testError;
   if (!testError) {
```
dgozman, line 208: We are missing this  branch now.

packages/playwright-test/src/test.ts
```diff
@@ -191,7 +191,11 @@ export class TestCase extends Base implements reporterTypes.TestCase {
       stderr: [],
       attachments: [],
       status: 'skipped',
-      steps: []
+      steps: [],
+      errors: [],
+      get error(): reporterTypes.TestError | undefined {
```
dgozman, line 196: Let's just assign at the same we assign to `errors`.

packages/playwright-test/src/expect.ts
```diff
@@ -128,28 +136,45 @@ const customMatchers = {
...
   get(target: any, prop: any, receiver: any): any {
     const value = Reflect.get(target, prop, receiver);
     if (typeof value !== 'function')
       return new Proxy(value, this);
     return (...args: any[]) => {
+      const testInfo = currentTestInfo();
+      if (!testInfo)
+        return value.call(target, ...args);
+      const handleError = (e: Error) => {
+        if (this._isSoft || (prop === 'toMatchSnapshot' && e.message.includes('missing')))
```
dgozman, line 162: Let's think about this issue.

aslushnikov: did alternative solution!

docs/src/test-assertions-js.md
```diff
@@ -19,10 +19,25 @@ expect(value).not.toEqual(0);
 await expect(locator).not.toContainText("some text");
 ```
 
-You can also specify a custom error message as a second argument to the `expect` function, for example:
+By default, failed assertion will terminate test execution. Playwright also
```
dgozman, line 22: I now think that we need a separate #h3 section, so that I see `expect.soft()` and `custom error message` in the right sidebar. WDYT?

packages/playwright-test/src/matchers/golden.ts
```diff
@@ -129,8 +129,8 @@ export function compare(
       return { pass: true, message };
     }
     if (updateSnapshots === 'missing') {
-      testInfo._appendErrorMessage(message);
-      return { pass: true, message };
+      testInfo._failWithError(serializeError(new Error(message)), false /* isHardError */);
```
dgozman, line 132: Does this get a nice stack trace?

dgozman (approved):
All good modulo expect.soft vs expect.defer. Let's settle on the name :)

packages/playwright-test/src/matchers/golden.ts
aslushnikov: It does, thanks to a new stack filter that filters out expect stack frames.

docs/src/test-assertions-js.md
aslushnikov: I'm happy to iterate on the docs in the follow-up!

PR #11822, opened 2022-02-03
feat(parallel): allow setting enclosing scope parallel mode

pavelfeldman:


packages/playwright-test/types/test.d.ts
```diff
@@ -1706,6 +1708,7 @@ export interface TestType<TestArgs extends KeyValue, WorkerArgs extends KeyValue
    */
   only: SuiteFunction;
     };
+    configure(options: { mode?: 'parallel' | 'serial' }): void;
```
dgozman, line 1711: This is missing docs. I think it would be fixed if we do `configure: (...) => void` instead.

dgozman (approved):

PR #11826, opened 2022-02-03
chore: use ipc transport for out-of-process driver

pavelfeldman:


packages/playwright-core/src/cli/driver.ts
```diff
@@ -38,7 +38,7 @@ export function runDriver() {
     const playwright = createPlaywright(sdkLanguage);
     return new PlaywrightDispatcher(rootScope, playwright);
   });
-  const transport = new Transport(process.stdout, process.stdin);
+  const transport = process.env.PW_DRIVER_USE_IPC ? new IpcTransport(process) : new PipeTransport(process.stdout, process.stdin);
```
dgozman, line 41: We can probably replace env with `!!process.send` here.

dgozman (approved):

PR #11838, opened 2022-02-03
chore: Add pixel-buffer-diff behind env variable

p01:
Prove alternative image comparison using [pixel_buffer_diff](https://www.npmjs.com/package/pixel-buffer-diff) behind an env variable, just like for blinkDiff. This image comparison weighs ~1.2kb, is 4x faster than pixelMatch and yields better results as it takes both individual pixel differences and their sum into account to avoid false positives due to anti-aliasing differences.

Test authors, can opt-in by setting the `PW_USE_PIXEL_BUFFER_DIFF` env variable.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5775,6 +5775,11 @@
         "node": ">=6"
       }
     },
+    "node_modules/pixel-buffer-diff": {
+      "version": "1.3.2",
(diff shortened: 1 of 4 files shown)


aslushnikov:
@p01 thanks! This is too early for us to add this as a dependency yet. We don't know what we want yet; we'll reconsider this as our VRT story develops!

PR #11926, opened 2022-02-08
chore: disable page.pause() when JS debugger is attached

pavelfeldman:


andrisi:
This should be optional I think. page.pause() is very useful, even if the scrio is run with a debugger attached. For example you can "help" the script, and then change it accordingly in the next run. Being paused with the debugger makes the browser freeze, while page.pause() does not and that's a big (helpful) difference.

PR #11951, opened 2022-02-08
fix(webServer): do not set baseURL equal to webServer.url

dgozman:


mxschmitt (approved):
lets also follow-up on the docs that this correlation has never existed?

pavelfeldman (approved):

PR #11953, opened 2022-02-09
feat(debug): transpile js as well

pavelfeldman:


dgozman (approved):

packages/playwright-test/src/transform.ts
```diff
@@ -135,39 +144,50 @@ export function transformHook(code: string, filename: string, isModule = false):
...
     plugins.unshift([require.resolve('@babel/plugin-transform-react-jsx')]);
 
-  if (!isModule) {
-    plugins.push([require.resolve('@babel/plugin-transform-modules-commonjs')]);
-    plugins.push([require.resolve('@babel/plugin-proposal-dynamic-import')]);
-  }
 
-  if (process.env.PW_TEST_SOURCE_TRANSFORM)
-    plugins.push([process.env.PW_TEST_SOURCE_TRANSFORM]);
+  if (scriptPreprocessor)
```
dgozman, line 189: Only add this one when file is in scope? Otherwise we transform ts files outside the scope.

dgozman (approved):

packages/playwright-test/src/transform.ts
pavelfeldman: Done!

PR #11961, opened 2022-02-09
WIP: implement screenshot blackout

aslushnikov:
This is a WIP approach.

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -2683,6 +2683,11 @@ An object which specifies clipping of the resulting image. Should have the follo
 Hides default white background and allows capturing screenshots with transparency. Not applicable to `jpeg` images.
 Defaults to `false`.
 
+### option: Page.screenshot.blackout
+- `blackout` <[Array]<[Locator]>>
+
+Blackouts given set of elements for the screenshot.
+
 ### option: Page.screenshot.timeout = %%-input-timeout-%%
 
 ## async method: Page.selectOption
--- a/packages/playwright-core/src/client/page.ts
+++ b/packages/playwright-core/src/client/page.ts
@@ -457,10 +457,13 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
     await this._channel.setNetworkInterceptionEnabled({ enabled: false });
   }
 
-  async screenshot(options: channels.PageScreenshotOptions & { path?: string } = {}): Promise<Buffer> {
+  async screenshot(options: channels.PageScreenshotOptions & { path?: string, blackout?: api.Locator[] } = {}): Promise<Buffer> {
     const copy = { ...options };
     if (!copy.type)
       copy.type = determineScreenshotType(options);
+    if (options.blackout) {
+      await Promise.all(options.blackout.map(loc => (loc as any)._blackout().catch(() => {})));
+    }
     const result = await this._channel.screenshot(copy);
     const buffer = Buffer.from(result.binary, 'base64');
     if (options.path) {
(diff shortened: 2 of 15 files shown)


docs/src/api/class-page.md
```diff
@@ -2683,6 +2683,11 @@ An object which specifies clipping of the resulting image. Should have the follo
 Hides default white background and allows capturing screenshots with transparency. Not applicable to `jpeg` images.
 Defaults to `false`.
 
+### option: Page.screenshot.blackout
+- `blackout` <[Array]<[Locator]>>
```
dgozman, line 2687: Locator|FrameLocator ?

packages/playwright-core/src/client/page.ts
```diff
@@ -457,10 +457,13 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
     await this._channel.setNetworkInterceptionEnabled({ enabled: false });
   }
 
-  async screenshot(options: channels.PageScreenshotOptions & { path?: string } = {}): Promise<Buffer> {
+  async screenshot(options: channels.PageScreenshotOptions & { path?: string, blackout?: api.Locator[] } = {}): Promise<Buffer> {
```
dgozman, line 460: Use Locator|FrameLocator here

p01:
Can we use a more inclusive term, e.g. `redact` ? The term "blackout" sounds terrible.

docs/src/api/class-page.md
p01, line 2687: s/blackout/hide|redact|cover|overlay|...

packages/playwright-core/src/server/injected/highlight.ts
```diff
@@ -113,6 +113,27 @@ export class Highlight {
     this._actionPointElement.hidden = true;
   }
 
+  blackoutElements(elements: Element[]) {
+    const boxes = elements.map(e => e.getBoundingClientRect());
+    const pool = this._highlightElements;
+    this._highlightElements = [];
+    for (const box of boxes) {
+      const highlightElement = pool.length ? pool.shift()! : this._createHighlightElement();
+      highlightElement.style.backgroundColor = '#000';
```
p01, line 122: As dark and light themes are common place, we can't hardcode a `#000` overlay color.

Instead we should add an `overlayColor` ( or whatever more inclusive name we pick for this feature ) property that defaults to a neutral or less common color e.g. `#808080` or `#F0F`

docs/src/api/class-page.md
aslushnikov: We discussed this last week and so far settled upon `mask`.

packages/playwright-core/src/server/injected/highlight.ts
aslushnikov: I like the `#F0F`!

aslushnikov:
Technical tidbits:

This PR explored the implementation via re-using the locators highlighting functionality in the protocol.
We decided against this approach and instead will explore serializing locators over the protocol to the server side.

The alternative implementation lives here: https://github.com/microsoft/playwright/pull/12072

p01:
Thank you.

Good thinking about handling this scenario. 
```typescript
await Promise.all([
  page.locator('foo').screenshot({ mask: [ page.locator('.mask-1') ] }),
  page.locator('bar').screenshot({ mask: [ page.locator('.mask-2') ] }),
]);a
```
😦 TBH I would have no idea what to expect ( in terms of execution order ) from that code with the disabling of animations, the masks, ... if it wasn't serialized on the server side. Just never saw/wrote tests with this scenario.

PR #11964, opened 2022-02-09
docs: remove webServer/baseURL mentionings

mxschmitt:
Follow-up on #11951.


dgozman (commented):
I have mixed feelings about this. Not mentioning this in docs would produce a surprise. I think we should mention the actual behavior, but recommend specifying it explicitly. WDYT?

PR #11970, opened 2022-02-09
feat: add key support on react engine

kblok:
I've got [this question](https://stackoverflow.com/questions/71050193/react-locator-example/71052432#71052432) on StackOverflow. And although, in that case, the value used as a `key` was part of the `props` attributes (it was `job.id`), that might not always be true.

I am bringing this to the tell to see what you think about this.
I'm also fixing a typo :)

### DIFF
--- a/docs/src/selectors.md
+++ b/docs/src/selectors.md
@@ -755,6 +755,7 @@ In react selectors, component names are transcribed with **CamelCase**.
 Selector examples:
 
 - match by **component**: `_react=BookItem`
+- match by component and **key**: `_react=BookItem[key = '2']`
 - match by component and **exact property value**, case-sensitive: `_react=BookItem[author = "Steven King"]`
 - match by property value only, **case-insensitive**: `_react=[author = "steven king" i]`
 - match by component and **truthy property value**: `_react=MyButton[enabled]`
--- a/packages/playwright-core/src/server/injected/reactSelectorEngine.ts
+++ b/packages/playwright-core/src/server/injected/reactSelectorEngine.ts
@@ -19,6 +19,7 @@ import { isInsideScope } from './selectorEvaluator';
 import { checkComponentAttribute, parseComponentSelector } from './componentUtils';
 
 type ComponentNode = {
+  key?: any,
   name: string,
   children: ComponentNode[],
   rootElements: Element[],
@@ -26,6 +27,7 @@ type ComponentNode = {
 };
 
 type ReactVNode = {
+  key?: any,
   // React 16+
   type: any,
   child?: ReactVNode,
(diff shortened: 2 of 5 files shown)


packages/playwright-core/src/server/injected/reactSelectorEngine.ts
```diff
@@ -165,12 +172,17 @@ export const ReactEngine: SelectorEngine = {
     const reactRoots = findReactRoots(document);
     const trees = reactRoots.map(reactRoot => buildComponentsTree(reactRoot));
     const treeNodes = trees.map(tree => filterComponentsTree(tree, treeNode => {
+      const props = treeNode.props ?? {};
+
+      if (treeNode.key !== undefined)
+        props.key = treeNode.key;
```
aslushnikov, line 178: can I have a react prop named "key" that would clash with this?

kblok: you can't :)

aslushnikov: oh nice! Would you mind also adding a line to the docs on the react selectors regarding `key` filtering?

kblok: Done @aslushnikov!

aslushnikov (approved):

aslushnikov:
@kblok awesome, thank you! Would you like to do the same for vuejs? If not I'll follow-up.

> And although, in that case, the key was part of the props attributes. That might not always be true.

Also, I got confused: react documentation say that you cannot have property named "key". What do you mean then?

kblok:
@aslushnikov I updated the description. What I meant is that the value they were using as a `key` was inside an object in the `props`.

kblok:
> @kblok awesome, thank you! Would you like to do the same for vuejs? If not I'll follow-up.

Sure! I will try to cook an example.

docs/src/selectors.md
```diff
@@ -755,6 +755,7 @@ In react selectors, component names are transcribed with **CamelCase**.
 Selector examples:
 
 - match by **component**: `_react=BookItem`
+- match by component and **key**: `_react=BookItem[key = '2']`
```
rwoll, line 758: @aslushnikov / @kblok For the docs, should the `key` example go last (and/or come with a note/caveat linking to https://playwright.dev/docs/selectors#best-practices or mentioning React selector best practices)?

In most cases, I think the other React selector examples here are preferable (alone or chained with Playwright's other selectors) as they will be less fragile than `key` (at least anecdotally thinking through how often I've changed `key`s vs. other props like e.g. `author`.)

aslushnikov: fair. Feel free to amend!

PR #11995, opened 2022-02-10
fix(test-runner): undefined body crash with manual attachments

### LINKED ISSUE #11565: [BUG] Old Attachment API Crashes Base Reporter With Undefineds
**Context:**
- Playwright Version: ToT

**Code Snippet**

```ts
test.only('example', ({}, testInfo) => {
  testInfo.attachments.push({
    body: undefined,
    contentType: "text/plain",
    name: "example.txt",
  });

  expect(1).toBe(2);
});
```

**Describe the bug**

If a user explicitly pushes onto the attachments array like above (instead of using the new [async attachments API](https://playwright.dev/docs/api/class-testinfo#test-info-attach) introduced in v1.18), the base reporter currently crashes:

```
…/packages/playwright-test/lib/reporters/base.js:279
            let text = attachment.body.toString();
                                       ^

TypeError: Cannot read property 'toString' of undefined
    at formatFailure (…/packages/playwright-test/lib/reporters/base.js:279:40)
    at LineReporter.onTestEnd (…/packages/playwright-test/lib/reporters/line.js:85:43)
    at Multiplexer.onTestEnd (…/playwright/packages/playwright-test/lib/reporters/multiplexer.js:69:124)
    at Dispatcher._reportTestEnd (…/playwright/packages/playwright-test/lib/dispatcher.js:531:186)
    at Worker.onTestEnd (…/playwright/packages/playwright-test/lib/dispatcher.js:218:12)
    at Worker.emit (events.js:400:28)
    at Worker.emit (domain.js:470:12)
    at ChildProcess.<anonymous> (…/playwright/packages/playwright-test/lib/dispatcher.js:585:12)
    at ChildProcess.emit (events.js:400:28)
```

The new [v1.18 async attachments API](https://playwright.dev/docs/api/class-testinfo#test-info-attach) method handles these cases gracefully (and reports the errors nicely), so the recommendation for user-land code should be to use the new async API (I think).

However, the crash that can be triggered with the older push strategy is undesirable as it crashes the whole test runner, so it might be desirable to have more error handling in the reporter (or elsewhere).

cc: @dgozman What are your thoughts? I'm inclined to say "manually pushing onto the attachments array is deprecated, only use the new [v1.18 async attach API](https://playwright.dev/docs/api/class-testinfo#test-info-attach)" which was introduced to handle the fact that with the manual push onto the array, the state of the world can change by the time the reporter needs to use it, plus pushing onto the array defers possibility of validation until later.

rwoll:
Fixes #11565

The new (as of 1.18) `async testInfo.attach(…)` API handles this
gracefully (and is part of the reason for the new API's existence).
However, for the foreseeable future, it's still possible to manually
push onto the attachments array where we can't validate the contents
until it's too late, so this change ensures more graceful handling in
that case.

### DIFF
--- a/packages/playwright-test/src/reporters/base.ts
+++ b/packages/playwright-test/src/reporters/base.ts
@@ -258,8 +258,8 @@ export function formatFailure(config: FullConfig, test: TestCase, options: {inde
             resultLines.push('');
           }
         } else {
-          if (attachment.contentType.startsWith('text/')) {
-            let text = attachment.body!.toString();
+          if (attachment.contentType.startsWith('text/') && attachment.body) {
+            let text = attachment.body.toString();
             if (text.length > 300)
               text = text.slice(0, 300) + '...';
             resultLines.push(colors.cyan(`    ${text}`));
--- a/tests/playwright-test/reporter-base.spec.ts
+++ b/tests/playwright-test/reporter-base.spec.ts
@@ -270,3 +270,22 @@ test('should print "no tests found" error', async ({ runInlineTest }) => {
   expect(result.exitCode).toBe(1);
   expect(result.output).toContain('no tests found.');
 });
+
+test('should not crash on undefined body with manual attachments', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.test.js': `
+      const { test } = pwt;
+      test('one', async ({}, testInfo) => {
+        testInfo.attachments.push({
+          name: 'foo.txt',
+          body: undefined,
+          contentType: 'text/plain'
+        });
+        expect(1).toBe(2);
+      });
+    `,
+  });
+  expect(stripAnsi(result.output)).not.toContain('Error in reporter');
+  expect(result.failed).toBe(1);
+  expect(result.exitCode).toBe(1);
+});


rwoll:
cc: @dgozman for review

packages/playwright-test/src/reporters/base.ts
```diff
@@ -258,8 +258,8 @@ export function formatFailure(config: FullConfig, test: TestCase, options: {inde
             resultLines.push('');
           }
         } else {
-          if (attachment.contentType.startsWith('text/')) {
-            let text = attachment.body!.toString();
+          if (attachment.contentType.startsWith('text/') && attachment.body) {
```
rwoll, line 261: Let me know if fixing this in base reporter is too tactical. The alternative I considered was filtering out any attachments where the `body XOR path` precondition failed to be met which would happen here https://github.com/microsoft/playwright/blob/1f6b84f445597655b015df52cef0784d9963492c/packages/playwright-test/src/workerRunner.ts#L415 and then guarantee any reporters that deal with attachments can always depend on their either being a defined path or body (like the new `async testInfo.attach` guarantees).

dgozman: Since our attachment type has `{  path?: string, body?: Buffer }`, I think that reporters should handle the case where both are missing.

dgozman (approved):

PR #12051, opened 2022-02-11
chore: add support private methods in TS

### LINKED ISSUE #12027: [Feature] Update babel config to include private methods proposal plugin
Could private class fields be added to the babel config for tests? It is included in `@babel/preset-env` and is a [stage 4 proposal now](https://github.com/tc39/proposal-class-fields#class-field-declarations-for-javascript) which should be integrated into the standard soon.

Moved from comment in #11440.

```
$ playwright test
Using config at .../playwright.config.ts

SyntaxError: .../index.ts: Class private methods are not enabled.
  81 |   }
  82 |
> 83 |   get #log() {
     |   ^
  84 |     return this.#logger
  85 |   }
  86 |
```

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/12027

### DIFF
--- a/packages/playwright-test/src/transform.ts
+++ b/packages/playwright-test/src/transform.ts
@@ -158,6 +158,7 @@ export function transformHook(code: string, filename: string, isModule = false):
         [require.resolve('@babel/plugin-proposal-logical-assignment-operators')],
         [require.resolve('@babel/plugin-proposal-nullish-coalescing-operator')],
         [require.resolve('@babel/plugin-proposal-optional-chaining')],
+        [require.resolve('@babel/plugin-proposal-private-methods')],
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/transform.ts
```diff
@@ -158,6 +158,7 @@ export function transformHook(code: string, filename: string, isModule = false):
         [require.resolve('@babel/plugin-proposal-logical-assignment-operators')],
         [require.resolve('@babel/plugin-proposal-nullish-coalescing-operator')],
         [require.resolve('@babel/plugin-proposal-optional-chaining')],
+        [require.resolve('@babel/plugin-proposal-private-methods')],
```
dgozman, line 161: Missing package.json change?

dgozman (approved):

PR #12062, opened 2022-02-12
fix(test runner): shutdown redundant workers

dgozman:


packages/playwright-test/src/dispatcher.ts
```diff
@@ -141,6 +144,15 @@ export class Dispatcher {
     this._finished.resolve();
   }
 
+  private _isWorkerRedundant(worker: Worker) {
```
pavelfeldman, line 147: `_willWorkerGetTasks`?
`_shouldBeDecomissioned`?

pavelfeldman (approved):

PR #12070, opened 2022-02-13
fix: propagate exit code in experimental mode

pierscowburn:
In experimental ESM mode a child process is forked in order to run the tests. Currently the exit code of this child process is not propagated to the exit code of the parent process, which means that the process exits with a status code of `0` even if some of the tests failed.

This makes it difficult to use Playwright in CI in experimental mode, as the CI pipeline as a whole will pass despite the test failures.

This change addresses this by propagating the exit code in the case where it is non-zero.


pavelfeldman (approved):

aslushnikov:
@pierscowburn thanks! I'll cherry-pick this fix to 1.19.1

pierscowburn:
You’re very welcome! Love the project.

PR #12072, opened 2022-02-14
feat: support `mask` option in screenshot methods

### LINKED ISSUE #10162: Hide/remove Elements from Screenshots
### Discussed in https://github.com/microsoft/playwright/discussions/5068

<div type='discussions-op-text'>

<sup>Originally posted by **rollin-cogers** January 20, 2021</sup>
Hey everyone! I've been very excited to use this tool for testing, as I've previously been using protractor for all my testing.

In the past I've used modules like webdriver-image-comparison that allow for options like [hiding and removing elements](https://github.com/wswebcreation/webdriver-image-comparison/blob/master/docs/OPTIONS.md#hideelements-5) from a screenshot by adding `visibility: hidden` or `display: none` to the styling for the passed elements.

Is this a feature that we might see in the future with playwright? Would anyone else benefit from the ability to exclude elements from screenshots?

I would love to see something like this:
```javascript
page.screenshot({fullpage: true, hideElements: ['div.dynamic_content', 'div#profilePicture']})
```</div>

### PR DESCRIPTION
**Pull request by aslushnikov**
Fixes #10162

### CODE REVIEW

File: docs/src/api/params.md [hunk 1]
```diff
@@ -913,11 +913,17 @@ saved to the disk.
 
 Specify screenshot type, defaults to `png`.
 
+## screenshot-option-mask
+- `mask` <[Array]<[Locator|FrameLocator]>>
```
mxschmitt, line 917: c# and java will hate it :/

tests/page/page-screenshot.spec.ts
```diff
@@ -615,3 +615,45 @@ it.describe('page screenshot animations', () => {
     ]);
   });
 });
+
+it.only('should work', async ({ page, server }) => {
```
mxschmitt, line 619:
```suggestion
it.only('should work', async ({ page, server }) => {
```
Quick tip for development:

npm run ctest foo.spec.ts:42

Instead of test.only

tests/page/page-screenshot.spec.ts
```diff
@@ -615,3 +615,45 @@ it.describe('page screenshot animations', () => {
...
+
+  const loc = page.locator('body');
+  await loc.screenshot({
+    path: '/Users/andreylushnikov/prog/playwright/aaaa-3.png',
+    mask: [
+      page.locator('div').nth(16),
+    ],
+  });
+  await page.screenshot({
+    path: '/Users/andreylushnikov/prog/playwright/aaaa-4.png',
```
mxschmitt, line 647: They should not be absolute!

aslushnikov: That's draft for a reason!

tests/page/page-screenshot.spec.ts
aslushnikov: Sometimes I do fallback to line numbers, but do find .only more convenient in the end of the day

docs/src/api/params.md
aslushnikov: Let's discuss on a meeting today!

packages/playwright-core/src/server/types.ts
```diff
@@ -16,6 +16,7 @@
  */
 
 import { Size, Point, Rect, TimeoutOptions } from '../common/types';
+import * as frames from './frames';
```
dgozman, line 19: I don't think types should depend on frames. Can we avoid that?

tests/page/page-screenshot.spec.ts
```diff
@@ -338,6 +338,77 @@ it.describe('page screenshot', () => {
...
+      await page.goto(server.PREFIX + '/grid.html');
+      const bodyLocator = page.locator('body');
+      expect(await bodyLocator.screenshot({
+        mask: [ page.locator('div').nth(5) ],
+      })).toMatchSnapshot('mask-should-work-with-locator.png');
+    });
+
+    it('should work with elementhandle', async ({ page, server }) => {
+      await page.goto(server.PREFIX + '/grid.html');
+      const bodyLocator = page.locator('body');
```
dgozman, line 360: Did you mean `await page.$('body')`?

dgozman (approved):

PR #12101, opened 2022-02-15
browser(firefox): manage scripts to evaluate on load on front-end

pavelfeldman:


examples/svgomg/tests/example.spec.ts
```diff
@@ -3,6 +3,8 @@
 import { test, expect } from '@playwright/test';
 import fs from 'fs';
 
+test.describe.configure({ mode: 'parallel' });
```
dgozman, line 6: Looks like this includes previous PR.

dgozman (approved):

PR #12143, opened 2022-02-16
chore: parse-skip .tsx

pavelfeldman:


packages/playwright-test/src/runner.ts
```diff
@@ -221,7 +221,7 @@ export class Runner {
       const allFiles = await collectFiles(project.config.testDir);
       const testMatch = createFileMatcher(project.config.testMatch);
       const testIgnore = createFileMatcher(project.config.testIgnore);
-      const extensions = ['.js', '.ts', '.mjs', ...(process.env.PW_COMPONENT_TESTING ? ['.tsx', '.jsx'] : [])];
+      const extensions = ['.js', '.ts', '.mjs', '.tsx', '.jsx'];
```
dgozman, line 224: This could be a breaking behavior, for example if someone has jest tests in the same repo.

packages/playwright-test/src/transform.ts
```diff
@@ -251,10 +249,7 @@ export function wrapFunctionWithLocation<A extends any[], R>(func: (location: Lo
   };
 }
 
-// Experimental components support for internal testing.
 function isComponentImport(filename: string): boolean {
-  if (!process.env.PW_COMPONENT_TESTING)
-    return false;
   if (filename.endsWith('.tsx') && !filename.endsWith('spec.tsx') && !filename.endsWith('test.tsx'))
```
dgozman, line 253: I think we should plumb testMatch here, or even the list of test files.

packages/playwright-test/src/tsxTransform.ts
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ */
+
+import { types as t } from '@babel/core';
+import { declare } from '@babel/helper-plugin-utils';
+ 
+export default declare(api => {
+  api.assertVersion(7);
+
+  return {
```
dgozman, line 23: I guess this works :)

packages/playwright-test/src/runner.ts
pavelfeldman: Not sure how this is breaking, we are under our test folder with our test filter.

dgozman (approved):

PR #12146, opened 2022-02-16
chore: refactor internal cli commands to use commander

dgozman:


tests/config/default.playwright.config.ts
```diff
@@ -71,7 +71,7 @@ if (mode === 'service') {
 
 if (mode === 'service2') {
   config.webServer = {
-    command: 'npx playwright run-server 3333',
```
mxschmitt, line 74: not sure if this breaks people, some are using this unofficial `run-server` command already.

mxschmitt (approved):

yury-s (approved):

tests/config/default.playwright.config.ts
dgozman: Well, they should not 😄

PR #12186, opened 2022-02-17
fix(trace-viewer): don't crash when argument is null

mxschmitt:
Previously the following was crashing the trace viewer with `cannot read 'guid' of null`.

Repro:
```ts

(async () => {
  await context.tracing.start();
  await page.evaluate("1+1", null)
  await context.tracing.stop({path: "foo.zip"})
  await browser.close();
})();
```

This was because `null` is typeof `object`.


pavelfeldman (approved):
Let's add a test!

Issue #11670, opened 2022-01-27
[Feature] browser API mocks


yury-s:
In some cases (e.g. #7339, #11580, #11288) the clients may want to call into experimental browser APIs which normally don't instrument until they become standard and supported by all browser vendors. In cases like that where Playwright doesn't provide dedicated testing API out of the box the clients would benefit from a mechanism that allows adding javascript stubs for such APIs and mocking them.

judofyr:
Interesting. Would these mocks actually be able to support the real functionality (i.e. accessing the host file system)?

PR #12244, opened 2022-02-19
docs: add mock API project examples

yury-s:
* The battery API example uses https://github.com/pazguille/demo-battery-api app.
* Filesystem API example uses its own test page.

#11670


dgozman (approved):

examples/mock-battery/tests/update-battery-status.spec.js
```diff
@@ -0,0 +1,80 @@
+// @ts-check
+const { test, expect } = require('@playwright/test');
+
+let log = [];
+
+test.beforeEach(async ({page}) => {
+  log = [];
+  await page.exposeFunction('logCall', msg => log.push(msg));
```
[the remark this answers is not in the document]
yury-s: > * wdyt about making this a fixture?

It's not a fixture in the docs which is why I keep it as `beforeEach` here.

> * some comments would be nice here.

Done.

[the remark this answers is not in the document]
yury-s: Also wrapping both logCall and logs in a fixture would require including `logs` into beforeEach parameters even though it's not actually used.

examples/mock-filesystem/tests/directory-reader.spec.js
```diff
@@ -0,0 +1,64 @@
+// @ts-check
+const { test, expect } = require('@playwright/test');
+
+test.beforeEach(async ({page}) => {
+  await page.addInitScript(() => {
+    class FileSystemHandleMock {
+      constructor({name, children}) {
+        this.name = name;
+        children ??= [];
```
mxschmitt, line 9: Do all browsers understand that already? Or does it go through babel?

yury-s: dunno, tested with chrome only and the example doesn't run other browsers, we can change it to the old syntax

yury-s: works in all browsers

PR #12250, opened 2022-02-19
fix(electron): do not attach external debugger when running Electron …

pavelfeldman:
…tests

### DIFF
--- a/packages/playwright-core/src/server/electron/electron.ts
+++ b/packages/playwright-core/src/server/electron/electron.ts
@@ -127,10 +127,15 @@ export class Electron extends SdkObject {
       const artifactsDir = await fs.promises.mkdtemp(ARTIFACTS_FOLDER);
 
       const browserLogsCollector = new RecentLogsCollector();
+      const env = options.env ? envArrayToObject(options.env) : process.env;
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -127,10 +127,15 @@ export class Electron extends SdkObject {
       const artifactsDir = await fs.promises.mkdtemp(ARTIFACTS_FOLDER);
 
       const browserLogsCollector = new RecentLogsCollector();
+      const env = options.env ? envArrayToObject(options.env) : process.env;
+      // When debugging Playwright test that runs Electron, NODE_OPTIONS
+      // will make the debugger attach to Electron's Node. But Playwright
+      // also needs to attach to drive the automation. Disable external debugging.
+      delete env.NODE_OPTIONS;
```
mxschmitt, line 134: Not sure if we always want to do that or only if it's used with the vsc extension.

dgozman (approved):

packages/playwright-core/src/server/electron/electron.ts
pavelfeldman: Debugging is mutually exclusive with driving Electron because they are using the same debugging channel.

PR #12253, opened 2022-02-20
npm audit fix

ayroblu:
I ran npm audit fix
image


pavelfeldman:
Thanks for the report! We don't apply changes to the system solely based on the `npm audit fix` recommendations.

PR #12255, opened 2022-02-20
Add prettier

ayroblu:
Maybe you have a good reason for this, but just putting it out there, prettier is great, would love to have it. See #12256 for how it would run and change things


pavelfeldman:
Thanks for putting it together! We like it the way it is at the moment. Maybe some other time!

Issue #8161, opened 2021-08-12
feat: better support for visual regression testing


aslushnikov:
Playwright Test has a built-in `toMatchSnapshot()` method to power Visual Regression Testing (VRT).

However, VRT is still challenging due to variances in the host environments. There's a bunch of measures we can do right away to drastically improve experience in @playwright/test
- [ ] support for `docker` test fixture to run browsers inside docker image.
- [ ] support for `blur` in matching snapshot to counteract antialiasing

Interesting context:
- [migration from backstopjs to @playwright/test](https://github.com/microsoft/playwright/issues/7548#issuecomment-881897256)

florianbepunkt:
I think https://github.com/americanexpress/jest-image-snapshot provides a nice suite of options for various VRT scenarios. Test scenarios vary widely, depending on the context (testing components, whole pages, text-heavy or not, etc).

Besides bluring which helps a lot with antialiasing it would be nice if multiple image comparisons (e. g. SSM) would be possible. Alternative image comparison algorithms could be left to userland, if they can be plugged into `toMatchSnapshot` via a common interface.

aslushnikov:
> Besides bluring which helps a lot with antialiasing it would be nice if multiple image comparisons (e. g. SSM) would be possible.

@florianbepunkt What's SSM? Is it structural similarity measurement (SSIM)?

kevinmpowell:
Solid integration with Storybook would be beneficial for the work I do. [Chromatic](https://www.chromatic.com) and [Percy](https://percy.io) do this really well.

Also a UI for reviewing the diffs would be great.

kevinmpowell:
Slider is rarely useful for me. An onion-skin (transparency overlay) would be more useful.

AlexNetman:
@aslushnikov Why `toMatchSnapshot()` is not available in the documentation?
It can not be found in API list.
And the article that was in 1.13  https://playwright.dev/python/docs/1.13.0/test-snapshots
is not available for 1.14 anymore.

Thanks for thinking about Visual Regression testing. Thats important!

florianbepunkt:
On a related note: It would be great if tests could be run cross-plattform. Currently the os platform name is baked into the snapshot filename, so our CI tests sometime fail due to name miss-match. https://github.com/microsoft/playwright/issues/7575

lo1tuma:
> support for blur in matching snapshot to counteract antialiasing

[...]

PR #12258, opened 2022-02-20
Add SSIM

ayroblu:
SSIM has long been a feature of jest-image-snapshot
This ports the functionality to playwright
Attempts to address some of #8161

Note: I wasn't sure where to put the code so I just put it in the comparators file, happy for it to be in a separate file etc


pavelfeldman:
It is unclear whether ssim is the solution we'd like to use to improve our visual regression testing story!

PR #12271, opened 2022-02-21
docs(python): fix set_viewport_size in emulation.md

### LINKED ISSUE #1166: [BUG] Inconsistent documentation for page.set_viewport_size()
**Context:**
* https://playwright.dev/python/docs/emulation#viewport
* https://playwright.dev/python/docs/api/class-page#page-set-viewport-size

**Describe the bug**

[emulation#viewport](https://playwright.dev/python/docs/emulation#viewport) describes `page.set_viewport_size(height:int, width:int)`, while the method is [documented itself](https://playwright.dev/python/docs/api/class-page#page-set-viewport-size) as `page.set_viewport_size(size:dict[height:int, width:int])`, while the description itself also mentions `page.setViewportSize()`.

I think it makes a lot of sense (and would be intuitive) to allow  `page.set_viewport_size(height:int, width:int)` in addition to the backwards compatible (but rather js centric) `page.set_viewport_size(size:dict[height:int, width:int])`. But in any case, the docs being inconsistent is rather annoying. :-)

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/1166

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -2841,7 +2841,7 @@ are resolved relative to the the current working directory. For empty array, cle
 In the case of multiple pages in a single browser, each page can have its own viewport size. However,
 [`method: Browser.newContext`] allows to set viewport size (and more) for all pages in the context at once.
 
-`page.setViewportSize` will resize the page. A lot of websites don't expect phones to change size, so you should set the
+[`method: Page.setViewportSize`] will resize the page. A lot of websites don't expect phones to change size, so you should set the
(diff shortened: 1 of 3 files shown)


dwt:
This misses the typo at https://playwright.dev/python/docs/api/class-page#page-set-viewport-size where the docs talk about `page.setViewportSize` instead of `page.set_viewport_size`

dwt:
Are you guys considering also allowing the more pythonic `page.set_viewport_size(height:int, width:int)` as an alternative or is that out because of backwards compatibility?

mxschmitt:
> 

There is no plan to also allow it, it's the `dict` variant for now. It was only different before 1.x of Playwright for Python.

dwt:
With that: lgtm for what it's worth.

aslushnikov (approved):

Issue #12231, opened 2022-02-18
[Bug] Can't see post data payload for window.sendBacon events


manuelriveragax:
Hey, i been trying to get the payload for some events we are tracking in analytics, i need to verify we are sending the correct data in the payload but no luck so far. 

```javascript
import { test, expect } from '@fixtures/myFixtures'

test.describe('SSU Analytics events', () => {
  test('Free trial modal events', async ({ ssuProductionPage }) => {
    await ssuProductionPage.goToHome()
    
    // When this button gets clicked, it triggers the POST call
    await ssuProductionPage.signUpButtonMobile.click()
    
    
    // Example of a payload:
    // buttonLocation: "header"
    // buttonTitle: "START FREE TRIAL"
    // url: "https://www.homePage.com/"
    
    Assert the correct data in the payload
  })
})
```

Not sure if this is possible or how to, thanks for the help in advance.

manuelriveragax:
Similar issue: https://github.com/microsoft/playwright/issues/4037

```javascript

 test('Trying to get payload', async ({}) => {
    const browser = await chromium.launch({
      devtools: true,
      headless: false
    })

    const page = await browser.newPage()
    await page.goto('https://www.somepage.com/')

    const [loadRes] = await Promise.all([
      page.waitForRequest((req) =>
        req.url().startsWith('https://api/something/v1')
      ),
      await page.click('a.styles__Button-sc-1v0jnie-0.eHxhuY')
    ])

    console.log(loadRes.url())
    console.log(loadRes.postDataJSON())
    await browser.close()
  })
  
  ```
  
  Console output:
  https://api/something/v1
  null

**Comment by dgozman** [comment]
The snippet with `loadRes.postDataJSON()` looks good. If you have a repro that we can run locally to reproduce this problem, we would be able to debug it.

**Comment by manuelriveragax** [comment]
Thanks for the response!
```javascript
test('Test Case Name', async ({ page }) => {
  await page.goto('https://www.us-central1-dev.gcp.maestro.io/')
  const [loadRes] = await Promise.all([
    page.waitForRequest((req) =>
      req.url().startsWith('https://insights.maestro.io/ingest/v1')
    ),
    await page.click('a.styles__Button-sc-1v0jnie-0.eHxhuY')
  ])

  console.log(loadRes.url())
  console.log(loadRes.postDataJSON())
})
```

mxschmitt:
Thank you for the repro! Was able to reproduce and added a test for it.

[...]

PR #12274, opened 2022-02-21
test: add test for sendBeacon and asserting request body

mxschmitt:
Test for https://github.com/microsoft/playwright/issues/12231

`hasPostData` in https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-requestWillBeSent is true, so probably an upstream bug.

### DIFF
--- a/tests/page/network-post-data.spec.ts
+++ b/tests/page/network-post-data.spec.ts
@@ -107,3 +107,17 @@ it('should get post data for file/blob', async ({ page, server, browserName }) =
   ]);
   expect(request.postData()).toBe('file-contents');
 });
+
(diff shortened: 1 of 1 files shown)


tests/page/network-post-data.spec.ts
```diff
@@ -107,3 +107,17 @@ it('should get post data for file/blob', async ({ page, server, browserName }) =
   ]);
   expect(request.postData()).toBe('file-contents');
 });
+
+it('should get post data for navigator.sendBacon api calls', async ({ page, server, browserName }) => {
+  it.skip(browserName === 'webkit', 'not supported in WebKit');
```
Smrtnyk, line 112: sendBeacon does not work in webkit?
it does work in safari

aslushnikov (approved):

PR #12291, opened 2022-02-22
fix(list mode): keep outputDir intact

dgozman:


mxschmitt (approved):

packages/playwright-test/src/runner.ts
```diff
@@ -303,9 +303,11 @@ export class Runner {
       fatalErrors.push(createNoTestsError());
 
     // 8. Fail when output fails.
-    await Promise.all(Array.from(outputDirs).map(outputDir => removeFolderAsync(outputDir).catch(e => {
-      fatalErrors.push(serializeError(e));
-    })));
+    if (!list) {
```
pavelfeldman, line 306: Move it to line 360 instead?

pavelfeldman (approved):

PR #12325, opened 2022-02-23
chore: fix Android tests

### LINKED ISSUE #12214: [BUG] Android APP Testing: androidWebView.page: tracesDir
**Context:**
- Playwright Version: v1.19.1
- Operating System: Linux
- Node.js version: v14.18
- Browser: Android APP

## System:
 - OS: Linux 5.4 Ubuntu 20.04.3 LTS (Focal Fossa)
 - Memory: 196.99 MB / 15.45 GB
 - Container: Yes
## Binaries:
 - Node: 14.18.0 - /usr/local/bin/node
 - Yarn: 1.22.5 - /usr/bin/yarn
 - npm: 8.5.0 - /usr/local/bin/npm
## Languages:
 - Bash: 5.0.17 - /bin/bash

**Code Snippet**

```typescript
async () => {
    // Connect to the device.
    const [device] = await android.devices();
    console.log(`Model: ${device.model()}`);
    console.log(`Serial: ${device.serial()}`);

    // --------------------- WebView -----------------------

    // Launch an application with WebView.
    const app_package = 'com.my.app_package';
    await device.shell(`am force-stop ${app_package}`);
    await device.shell(`am start ${app_package}/.MainActivity`);
    // Get the WebView.
    const webview = await device.webView({ pkg: app_package }); // <- Error: androidWebView.page: tracesDir must be specified for BrowserContext

    // Work with WebView's page as usual.
    const page = await webview.page();
    console.log(await page.title()); 
}
```

**Describe the bug**

Well, we are working with the experimental Android feature. The aim is to test an Ionic APP (webview) and we are getting this error:

```
Error: androidWebView.page: tracesDir must be specified for BrowserContext
```

Following the trace we got into the `/node_modules/playwright-core/lib/server/android/android.js` file, line 285. It seems like a validation requires the tracesDir value filled in the BrowserOptions object, but in that file [(line 280)](https://github.com/microsoft/playwright/blob/1215057ca1cf756f35e045a34fd340838bee81d8/packages/playwright-core/src/server/android/android.ts#L267) was setted an empty value hardcoded.

If we fill that variable with some value in that file (inside node_modules), even if the folder doesn't exist the tests are working properly. The APP is launched in the physical device and we can use the `page` object to interact with the webview page. But we don't find any way to pass it from outside as its value is hardcoded.

Is there anything we are missing?

Note: we are not using any kind of trace recording, at least to our knowledge hehe

Thank you so much.

### LINKED ISSUE #12312: [internal] android is broken after tracing changes

mxschmitt:
Proof that its working!

image
image

Fixes #12312
Fixes #12214

Example run: https://github.com/mxschmitt/playwright/runs/5311299825?check_suite_focus=true


.github/workflows/tests_fyi.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - release-*
+      - bugfix/android-tests
```
mxschmitt, line 8: will delete it before merge.

utils/avd_recreate.sh
```diff
@@ -8,7 +8,7 @@ if [[ -z "${ANDROID_HOME}" ]]; then
     export ANDROID_SDK_ROOT=${SDKDIR}
 fi
 
-${ANDROID_HOME}/tools/bin/avdmanager delete avd --name android31 || true
-echo "y" | ${ANDROID_HOME}/tools/bin/sdkmanager --install "system-images;android-31;google_apis;x86_64"
-echo "no" | ${ANDROID_HOME}/tools/bin/avdmanager create avd --force --name android31 --device "Nexus 5X" --package "system-images;android-31;google_apis;x86_64"
+${ANDROID_HOME}/tools/bin/avdmanager delete avd --name android32 || true
+echo "y" | ${ANDROID_HOME}/tools/bin/sdkmanager --install "system-images;android-32;google_apis;x86_64"
```
mxschmitt, line 12: reason to bump to android-32: it uses chrome 98 instead of 91.

packages/playwright-core/src/server/android/android.ts
```diff
@@ -256,15 +260,23 @@ export class AndroidDevice extends SdkObject {
     await androidBrowser._init();
     this._browserConnections.add(androidBrowser);
 
+    const artifactsDir = await fs.promises.mkdtemp(ARTIFACTS_FOLDER);
```
mxschmitt, line 263: we don't launch a process, so we have to implement a naive cleanup logic ourselves.

packages/playwright-core/src/server/android/android.ts
```diff
@@ -256,15 +260,23 @@ export class AndroidDevice extends SdkObject {
...
       name: 'clank',
       isChromium: true,
       slowMo: 0,
       persistent: { ...options, noDefaultViewport: true },
-      artifactsDir: '',
-      downloadsPath: '',
-      tracesDir: '',
+      artifactsDir,
+      downloadsPath: artifactsDir,
+      tracesDir: artifactsDir,
```
dgozman, line 279: - I think this fixes https://github.com/microsoft/playwright/issues/12214!
- While you are here, perhaps fix the TODO at line 294?

packages/playwright-core/src/server/android/android.ts
```diff
@@ -256,15 +260,23 @@ export class AndroidDevice extends SdkObject {
     await androidBrowser._init();
     this._browserConnections.add(androidBrowser);
 
+    const artifactsDir = await fs.promises.mkdtemp(ARTIFACTS_FOLDER);
+    socket.on('close', async () => {
```
dgozman, line 264: - Maybe `androidBrowser.onclose = () => { ... }` is better?
- You could also add to  "gracefully close set" to cleanup on process exit like this: https://github.com/microsoft/playwright/blob/71edad327b7bd07bf00eea510a1edd5e258048bf/packages/playwright-core/src/server/chromium/chromium.ts#L192-L194

dgozman (approved):

packages/playwright-core/src/server/android/android.ts
mxschmitt: `androidBrowser.onclose` gets used internally by the transport, gets set there so we can't override it. Added it to the gracefully close set for safety, this gives us additional protection when running e.g. in the driver mode.

PR #12329, opened 2022-02-24
chore(test runner): run hooks/modifiers as a part of the test 

dgozman:
This moves `beforeAll`, `afterAll` and some modifiers from running as a separate entity into running inside a test.

Pros:
- All errors are reported as test errors.
- All artifacts are collected as test artifacts.
- Reporters support this out of the box.

Details:
- Each test computes the necessary hooks to run and runs them.
- Teardown is usually performed during the test (on test failure or worker stop).
- `skipRemaining` is added to `DonePayload` to preserve the behavior where `beforeAll` hook failure skips subsequent tests. This behavior can now be improved to only target tests affected by this hook.


aslushnikov (approved):

packages/playwright-test/src/workerRunner.ts
```diff
@@ -129,89 +140,43 @@ export class WorkerRunner extends EventEmitter {
...
         return true;
       });
       if (suite) {
-        const firstPool = suite.allTests()[0]._pool!;
-        this._fixtureRunner.setPool(firstPool);
-        await this._runSuite(suite, []);
+        this._extraSuiteAnnotations = new Map();
+        this._activeSuites = new Set();
+        this._didFullCleanup = false;
+        const testsAndRetries: { test: TestCase, retry: number }[] = [];
```
aslushnikov, line 155: this can be simplified!

packages/playwright-test/src/workerRunner.ts
```diff
@@ -284,98 +264,197 @@ export class WorkerRunner extends EventEmitter {
...
+            } catch (e) {
+              // Always run all the hooks, and capture the first error.
+              beforeAllError = beforeAllError || e;
+            }
+          }
+          if (beforeAllError)
+            throw beforeAllError;
+        }
+
+        // This will inform dispatcher that we should not run more tests from this group
```
aslushnikov, line 327: the comment should be inversed

packages/playwright-test/src/workerRunner.ts
```diff
@@ -284,98 +264,197 @@ export class WorkerRunner extends EventEmitter {
...
+        // because we had a beforeAll error.
+        // This behavior avoids getting the same common error for each test.
+        shouldSkipRemainingTests = false;
+
+        // Run "beforeEach" modifiers. Once started with "beforeEach", we must run all "afterEach" hooks as well.
+        shouldRunAfterEachHooks = true;
```
aslushnikov, line 333: this should move after modifier initialization

packages/playwright-test/src/workerRunner.ts
```diff
@@ -284,98 +264,197 @@ export class WorkerRunner extends EventEmitter {
...
+
+        // Now run the test itself.
+        const fn = test.fn; // Extract a variable to get a better stack trace ("myTest" vs "TestCase.myTest [as fn]").
+        await fn(params, testInfo);
+      }, 'allowSkips');
+
+      beforeHooksStep.complete(maybeError); // Second complete is a no-op.
+    });
+
+    if (shouldSkipRemainingTests)
```
aslushnikov, line 360: s/shouldSkipRemainingTests/didFailBeforeAll/

aslushnikov (approved):

aslushnikov (approved):

PR #12337, opened 2022-02-24
chore: download new browsers in watch script

mxschmitt:
Not sure if I like it but I often run into the issue that we have a new chromium and I need to install it...


dgozman (commented):
Not sure I like this as well :) Existing setup brings us close to our users.

PR #12341, opened 2022-02-24
devops: Docker publishing cleanups

mxschmitt:
Reduce diff to make it easier to re-use across languages.

Thats where the warning was shown: https://github.com/microsoft/playwright/runs/5315750627?check_suite_focus=true#step:14:125

### DIFF
--- a/.github/workflows/publish_release_docker.yml
+++ b/.github/workflows/publish_release_docker.yml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       is_release:
+        required: false
         type: boolean
         description: "Is this a release image?"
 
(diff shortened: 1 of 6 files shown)


.github/workflows/publish_release_docker.yml
```diff
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       is_release:
+        required: false
```
aslushnikov, line 7: can we keep this as a required field?

aslushnikov (approved):

.github/workflows/publish_release_docker.yml
mxschmitt, line 7:
```suggestion
        required: true
```

Issue #12307, opened 2022-02-23
[BUG] Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'


frederikhors:
**Context:**

- Playwright Version: `@playwright/test: 1.19.1`
- Browser: Chrome
- OS: Windows 10 10.0.19044
- Node: 16.14.0 - C:\Program Files\nodejs\node.EXE
- npm: 8.5.1 - C:\Program Files\nodejs\npm.CMD

**REPL:**

https://github.com/frederikhors/iss-playwright-typescript

**Describe the bug**

@mxschmitt asked me to open a new issue here: https://github.com/microsoft/playwright/issues/12138#issuecomment-1048777901.

The issue is:

@bbigras [here](https://github.com/microsoft/playwright/issues/12138#issuecomment-1045137524) suggests how to work with Typescript using `PW_EXPERIMENTAL_TS_ESM`.

On Windows 10 this doesn't work:

```
(node:12372) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/errors:464
    ErrorCaptureStackTrace(err);
    ^

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at new NodeError (node:internal/errors:371:5)
    at throwIfUnsupportedURLProtocol (node:internal/modules/esm/resolve:1033:11)
    at defaultResolve (node:internal/modules/esm/resolve:1103:3)
    at ESMLoader.resolve (node:internal/modules/esm/loader:530:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:251:18)
    at ESMLoader.import (node:internal/modules/esm/loader:332:22)
    at initializeLoader (node:internal/process/esm_loader:74:49)
    at loadESM (node:internal/process/esm_loader:87:11)
    at runMainESM (node:internal/modules/run_main:51:21)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:74:5) {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}
```

WSL is not an option because I cannot install it and I wanna use Windows 10 natively.

[...]

PR #12359, opened 2022-02-25
devops: run test-runner tests on Node 16

mxschmitt:
Turns out we run some tests, especially the fancy ESM tests only on Node.js 16. So we never ran them.

https://github.com/microsoft/playwright/issues/12307

### DIFF
--- a/.github/workflows/tests_primary.yml
+++ b/.github/workflows/tests_primary.yml
@@ -49,33 +49,6 @@ jobs:
         name: ${{ matrix.browser }}-${{ matrix.os }}-test-results
         path: test-results
 
-  test_test_runner:
-    name: Test Runner
-    strategy:
-      fail-fast: false
-      matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests_primary.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        node-version: [12]
+        node-version: [12, 16]
```
dgozman, line 58: Should we move half of these bots to secondary? I am afraid our primary is getting big again.

.github/workflows/tests_primary.yml
```diff
@@ -55,7 +55,7 @@ jobs:
...
         include:
```
dgozman, line 59: We can now remove this section, can't we?

dgozman (approved):

Issue #12319, opened 2022-02-23
docs get rid of /docs/assertions


mxschmitt:
- https://playwright.dev/docs/assertions -> we should reference web-first assertions

PR #12361, opened 2022-02-25
docs: delete assertions.md

mxschmitt:
https://github.com/microsoft/playwright/issues/12319


dgozman (commented):
Maybe we should merge the contents into "verification" doc, changing all examples to use locators?

docs/src/verification.md
```diff
@@ -3,8 +3,353 @@ id: verification
...
+```csharp
+var checked = await page.Locator("input").IsCheckedAsync();
+Assert.True(checked);
+```
+
+### API reference
+- [`method: Page.isChecked`]
+- [`method: ElementHandle.isChecked`]
+
+## JS expression
```
dgozman, line 132: Snippets in this section are about text content :)

dgozman (approved):

PR #12277, opened 2022-02-21
browser(webkit): rebase to 02/18/22

dpino:
Rebase `webkit/patches/boostrap.diff` to [r290138](https://trac.webkit.org/changeset/290138/webkit).

This likely solves the reported missing libOpenGL library in the CI bot.

PR #12368, opened 2022-02-25
browser(webkit): Fix build for Ubuntu 18.04

dpino:
Follow-up to https://github.com/microsoft/playwright/pull/12277

### DIFF
--- a/browser_patches/webkit/patches/bootstrap.diff
+++ b/browser_patches/webkit/patches/bootstrap.diff
@@ -2047,6 +2047,35 @@ index 447a02d3cd2802e028ec16d549654a06adcaaae5..bf576a79f103fe43e375df9be49e0f6e
  				4131C3CF234B98420028A615 /* rtc_stats.cc in Sources */,
  				41323A1D2665288B00B38623 /* packet_sequencer.cc in Sources */,
  				4131BF2D234B88200028A615 /* rtc_stats_collector.cc in Sources */,
+diff --git a/Source/WTF/Scripts/GeneratePreferences.rb b/Source/WTF/Scripts/GeneratePreferences.rb
+index de0f2268b178cd2672408d21ab18659d093208e1..4555f317b5b0158d9ce53c492f7a53384a90e109 100644
+--- a/Source/WTF/Scripts/GeneratePreferences.rb
++++ b/Source/WTF/Scripts/GeneratePreferences.rb
+@@ -228,11 +228,23 @@ class Preferences
+     result
+   end
+ 
++  def createTemplate(templateString)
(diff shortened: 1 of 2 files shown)


browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -2047,6 +2047,35 @@ index 447a02d3cd2802e028ec16d549654a06adcaaae5..bf576a79f103fe43e375df9be49e0f6e
...
  				4131BF2D234B88200028A615 /* rtc_stats_collector.cc in Sources */,
+diff --git a/Source/WTF/Scripts/GeneratePreferences.rb b/Source/WTF/Scripts/GeneratePreferences.rb
+index de0f2268b178cd2672408d21ab18659d093208e1..4555f317b5b0158d9ce53c492f7a53384a90e109 100644
+--- a/Source/WTF/Scripts/GeneratePreferences.rb
++++ b/Source/WTF/Scripts/GeneratePreferences.rb
+@@ -228,11 +228,23 @@ class Preferences
+     result
+   end
+ 
++  def createTemplate(templateString)
```
yury-s, line 2058: Are you planning to upstream this?

yury-s (approved):

browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -2047,6 +2047,35 @@ index 447a02d3cd2802e028ec16d549654a06adcaaae5..bf576a79f103fe43e375df9be49e0f6e
...
++    # Newer versions of ruby deprecate and/or drop passing non-keyword
++    # arguments for trim_mode and friends, so we need to call the constructor
```
yury-s, line 2060: I'm a bit confused if newer versions deprecate non-keyword args then we should be able to use keyword args in both old and new versions, what am I missing?

browser_patches/webkit/patches/bootstrap.diff
dpino: The change is already merged https://bugs.webkit.org/show_bug.cgi?id=237035

browser_patches/webkit/patches/bootstrap.diff
dpino: I think the issue was that in modern versions of Ruby the former code prints a warning. That was fixed in https://bugs.webkit.org/show_bug.cgi?id=236684. But in older versions of Ruby the new code doesn't work. The change could have been reverted to its original form but since Ubuntu 18.04 is not actively supported anymore, I think the changed proposed by https://bugs.webkit.org/show_bug.cgi?id=237035 is the right way to go.

PR #12371, opened 2022-02-25
test: mark failing android tests as fixme / fix them

mxschmitt:
This should reduce failing android tests to only the following three:
```
    [android] › page/page-screenshot.spec.ts:481:3 › page screenshot animations › should not capture css animations in shadow DOM 
    [android] › page/page-screenshot.spec.ts:495:3 › page screenshot animations › should stop animations that happen right before screenshot 
    [android] › page/page-screenshot.spec.ts:609:3 › page screenshot animations › should not change animation with playbackRate equal to 0 
```

### DIFF
--- a/tests/page/page-network-response.spec.ts
+++ b/tests/page/page-network-response.spec.ts
@@ -235,9 +235,9 @@ it('should report multiple set-cookie headers', async ({ page, server }) => {
   expect(await response.headerValues('set-cookie')).toEqual(['a=b', 'c=d']);
 });
 
-it('should behave the same way for headers and allHeaders', async ({ page, server, browserName, channel, platform }) => {
+it('should behave the same way for headers and allHeaders', async ({ page, server, browserName, channel, platform, isAndroid }) => {
   it.fixme(browserName === 'webkit' && platform === 'win32', 'libcurl does not support non-set-cookie multivalue headers');
-  it.skip(!!channel, 'Stable chrome uses \n as a header separator in non-raw headers');
+  it.fixme(isAndroid, 'Android uses \n as a header separator in non-raw headers');
   server.setRoute('/headers', (req, res) => {
     const headers = {
       'Set-Cookie': ['a=b', 'c=d'],
--- a/tests/page/page-request-intercept.spec.ts
+++ b/tests/page/page-request-intercept.spec.ts
@@ -16,12 +16,28 @@
  */
 
 import type { Route } from 'playwright-core';
-import { expect, test as it } from './pageTest';
+import { expect, test as base } from './pageTest';
 import fs from 'fs';
 import path from 'path';
 
-it('should fulfill intercepted response', async ({ page, server, isElectron }) => {
+const it = base.extend<{
+  // We access test servers at 10.0.2.2 from inside the browser on Android,
+  // which is actually forwarded to the desktop localhost.
+  // To use request such an url with apiRequestContext on the desktop, we need to change it back to localhost.
+  rewriteAndroidLoopbackURL(url: string): string
+      }>({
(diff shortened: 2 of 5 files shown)


tests/page/wheel.spec.ts
```diff
@@ -110,22 +109,6 @@ it('should work when the event is canceled', async ({ page }) => {
   expect(await page.evaluate('window.scrollY')).toBe(0);
 });
 
-contextTest('should scroll when emulating a mobile viewport', async ({ contextFactory, server, browserName }) => {
```
mxschmitt, line 113: we don't want contextTests inside `tests/page/**`

tests/page/page-network-response.spec.ts
```diff
@@ -235,9 +235,9 @@ it('should report multiple set-cookie headers', async ({ page, server }) => {
   expect(await response.headerValues('set-cookie')).toEqual(['a=b', 'c=d']);
 });
 
-it('should behave the same way for headers and allHeaders', async ({ page, server, browserName, channel, platform }) => {
+it('should behave the same way for headers and allHeaders', async ({ page, server, browserName, channel, platform, isAndroid }) => {
   it.fixme(browserName === 'webkit' && platform === 'win32', 'libcurl does not support non-set-cookie multivalue headers');
-  it.skip(!!channel, 'Stable chrome uses \n as a header separator in non-raw headers');
+  it.skip(!!channel || isAndroid, 'Stable chrome uses \n as a header separator in non-raw headers');
```
dgozman, line 240: "While we are here" rubric 😄  I think we can drop `!!channel` already?

tests/android/androidTest.ts
```diff
@@ -65,4 +65,24 @@ export const androidTest = baseTest.extend<PageTestFixtures, AndroidWorkerFixtur
     const page = await androidContext.newPage();
     await run(page);
   },
+
+  rewriteRequestToLocalhost: [async ({ isAndroid, rewriteRequestURLToLocalhost }, use) => {
+    await use(request => {
+      if (!isAndroid)
+        return request;
+      const initializer = (request as any)._initializer;
+      initializer.url = rewriteRequestURLToLocalhost(initializer.url);
```
dgozman, line 74: Let's disable all tests that need this. I don't think that tampering with internals is worth it.

tests/android/androidTest.ts
```diff
@@ -65,4 +65,24 @@ export const androidTest = baseTest.extend<PageTestFixtures, AndroidWorkerFixtur
...
+      return request;
+    });
+  }, { scope: 'worker' }],
+
+  rewriteRequestURLToLocalhost: [async ({ isAndroid }, use) => {
```
dgozman, line 79: Let's make this a local fixture `rewriteAndroidLoopbackURL` in the interception test file and add comments:
```js
// We access test servers at 10.0.2.2 from inside the browser on Android,
// which is actually forwarded to the desktop localhost.
// To use request such an url with apiRequestContext on the desktop, we need to change it back to localhost.
```

tests/page/page-request-intercept.spec.ts
```diff
@@ -41,10 +41,9 @@ it('should fulfill intercepted response', async ({ page, server, isElectron }) =
   expect(await page.evaluate(() => document.body.textContent)).toBe('Yo, page!');
 });
 
-it('should fulfill response with empty body', async ({ page, server, browserName, browserMajorVersion }) => {
-  it.skip(browserName === 'chromium' && browserMajorVersion <= 91, 'Fails in Electron that uses old Chromium');
```
dgozman, line 45: Nice!

dgozman (approved):

PR #12403, opened 2022-02-28
chore: hide unsupported expect methods

### LINKED ISSUE #11734: [BUG] Hide unsupported expect methods
`toBeCalledWith`, `toBeCalledTimes`, etc.

mxschmitt:
Fixes #11734

### DIFF
--- a/packages/playwright-test/types/testExpect.d.ts
+++ b/packages/playwright-test/types/testExpect.d.ts
@@ -52,6 +52,7 @@ type Awaited<T> = T extends PromiseLike<infer U> ? U : T;
 
 /**
  * Removed methods require the jest.fn() integration from Jest to spy on function calls which we don't support:
+ * - lastCalledWith()
  * - lastReturnedWith()
  * - nthCalledWith()
  * - nthReturnedWith()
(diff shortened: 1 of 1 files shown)


packages/playwright-test/types/testExpect.d.ts
```diff
@@ -48,18 +42,34 @@ export declare type Expect = {
...
+  'toMatch' |
+  'toMatchObject' |
+  'toStrictEqual' |
+  'toThrow' |
+  'toThrowError'
 
 declare global {
   export namespace PlaywrightTest {
-    export interface Matchers<R> extends Omit<expect.Matchers<R>, OverriddenExpectProperties> {
+    export interface Matchers<R> extends Pick<expect.Matchers<R>, SupportedExpectProperties> {
```
pavelfeldman, line 72: Do we have an idea on how many users we break and how? Should this have a large 'Breaking changes' section somewhere?

mxschmitt: It was never working from the implementation side (no access to the jest.fn() function), so it should be fine.

pavelfeldman: Let's get ourselves a list in the comments in this file where we explain why we ban everything else!

pavelfeldman (approved):

PR #12413, opened 2022-02-28
fix(hooks): separate test timeout from beforeAll/afterAll timeouts

dgozman:
This makes it possible to have longer `beforeAll`/`afterAll` and not
affect first/last test timeout.


aslushnikov (approved):

packages/playwright-test/src/testInfo.ts
```diff
@@ -281,19 +295,24 @@ export class TestInfoImpl implements TestInfo {
...
   setTimeout(timeout: number) {
-    if (!this.timeout)
-      return; // Zero timeout means some debug mode - do not set a timeout.
-    this.timeout = timeout;
-    this._timeoutRunner.updateTimeout(timeout);
+    if (this._currentRunnable.timeout !== undefined) {
+      if (!this._currentRunnable.timeout)
+        return; // Zero timeout means some debug mode - do not set a timeout.
+      this._currentRunnable.timeout = timeout;
+      this._timeoutRunner.updateTimeout(timeout);
```
yury-s, line 302: nit: move this outside of if block

PR #12417, opened 2022-03-01
feat(test runner): on beforeAll failure, precisely skip the tests

dgozman:
Previously, we used to skip all the tests from the same file when any `beforeAll` fails in the file.

Now, we only skip the rest of the tests affected by this particular `beforeAll` and continue with other tests in the new worker.


tests/playwright-test/max-failures.spec.ts
```diff
@@ -34,10 +34,10 @@ test('max-failures should work', async ({ runInlineTest }) => {
         });
       }
     `
-  }, { 'max-failures': 8 });
```
aslushnikov, line 37: revert

aslushnikov (approved):

PR #12467, opened 2022-03-02
docs: restore assertions timeout option in js

yury-s:


docs/src/api/params.md
```diff
@@ -703,12 +703,18 @@ The default value can be changed by using the [`method: BrowserContext.setDefaul
...
 using the [`method: AndroidDevice.setDefaultTimeout`] method.
 
-## assertions-timeout
+## js-assertions-timeout
 * langs: js
 - `timeout` <[float]>
 
 Time to retry the assertion for. Defaults to `timeout` in [`property: TestConfig.expect`].
 
+## csharp-java-python-assertions-timeout
```
mxschmitt, line 712: Not 100% sure but I think we don't need to call it differently, we can just define it another time and keep your new langs property which should handle the rest.

yury-s: In the current implementation we do, template substitution is dumb. I added a check to the generator so that throws if there are duplicate entries. We can change the approach altogether but currently I'm just fixing existing bug.

mxschmitt (approved):

PR #12473, opened 2022-03-02
docs: fix empty Docker images on languages

mxschmitt:
Turns out this was empty on the language ports since we never supported multiple langs on the markdown `code` elements: https://playwright.dev/python/docs/docker

### DIFF
--- a/docs/src/docker.md
+++ b/docs/src/docker.md
@@ -13,7 +13,15 @@ This image is published on [Docker Hub].
 
 ### Pull the image
 
-```bash js python csharp
+```bash js
+docker pull mcr.microsoft.com/playwright:focal
+```
+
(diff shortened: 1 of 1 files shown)


docs/src/docker.md
```diff
@@ -13,22 +13,38 @@ This image is published on [Docker Hub].
 
 ### Pull the image
 
-```bash js python csharp
```
yury-s, line 16: Should we support multiple languages in the filter instead?

yury-s (approved):

docs/src/docker.md
mxschmitt: we'll have dedicated images soon - so the next one who needs it adds it.

PR #12479, opened 2022-03-03
docs: correct instructions for port reuse

basarat:
At least according to typescript there is no `strict:false` 

image

But I do know of `reuseExistingServer:true` so that's what was probably meant :rose:


mxschmitt (approved):

mxschmitt:
Thanks!

PR #12509, opened 2022-03-04
browser(firefox): roll Firefox stable to M98

aslushnikov:
The release of Firefox 98 is scheduled for Mar 8, 2022.

### DIFF
--- a/browser_patches/firefox/preferences/playwright.cfg
+++ b/browser_patches/firefox/preferences/playwright.cfg
@@ -14,6 +14,12 @@ pref("browser.tabs.remote.useCrossOriginOpenerPolicy", false);
 
 pref("browser.tabs.remote.separatePrivilegedMozillaWebContentProcess", false);
 
+pref("pdfjs.disabled", true);
+
+pref("fission.autostart", false);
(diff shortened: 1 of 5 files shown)


browser_patches/firefox/preferences/playwright.cfg
```diff
@@ -14,6 +14,12 @@ pref("browser.tabs.remote.useCrossOriginOpenerPolicy", false);
 
 pref("browser.tabs.remote.separatePrivilegedMozillaWebContentProcess", false);
 
+pref("pdfjs.disabled", true);
```
mxschmitt, line 17: I'm curious, why did do we disable it?

aslushnikov: This way navigating to PDF triggers download instead of opening the inline plugin.

mxschmitt: sure, but did it broke tests or are we just randomly adding it?

aslushnikov: it did!

PR #12555, opened 2022-03-07
fix(install): always display missing dependencies

bodinsamuel:
Just a proposal I saw that it was output when the script did not find the distribution but even on ubuntu (or on docker in my case) sometimes it's still missing some deps and it's hard to say which one.

### DIFF
--- a/packages/playwright-core/src/utils/dependencies.ts
+++ b/packages/playwright-core/src/utils/dependencies.ts
@@ -188,6 +188,7 @@ export async function validateDependenciesLinux(sdkLanguage: string, linuxLddDir
       ``,
       `    ${maybeSudo}${buildPlaywrightCLICommand(sdkLanguage, 'install-deps')}`,
       ``,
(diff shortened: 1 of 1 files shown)


mxschmitt:
Which Ubuntu or what Docker image where you using? If install-deps was not working, then most likely something in our deps mapping was wrong.

bodinsamuel:
It was my bad, but I was in the blind since I wasn't sure what was missing.
In the end it was just `ttf-ubuntu-font-family`. 

Feel free to dismiss if you think it's not relevant ☺️

mxschmitt:
Which browser did you use?

bodinsamuel:
I initially wanted to leverage docker caching by doing `apt-get` manually before `npm install` but since playwright have its own list that could change I reverted and used what you did until I figure out a better way. I did not commit the prev docker version but if you are curious this is there: https://github.com/algolia/renderscript/pull/572

aslushnikov:
This should be fixed on tip-of-tree as part of the recent refactoring. Please check it out on tip-of-tree!

PR #12577, opened 2022-03-08
feat: support multiple screenshots in HTML report

aslushnikov:
This patch adds support to multiple diffs. These are possible
due to soft assertions.

Drive-by: rename second screenshot in `toHaveScreenshot` failure when
re-generating expectations from "expected" to "previous".

### DIFF
--- a/packages/html-reporter/src/testResultView.tsx
+++ b/packages/html-reporter/src/testResultView.tsx
@@ -26,30 +26,68 @@ import { AttachmentLink } from './links';
 import { statusIcon } from './statusIcon';
 import './testResultView.css';
 
-const imageDiffNames = ['diff', 'expected', 'actual'];
+type DiffTab = {
+  id: string,
+  title: string,
+  attachment: TestAttachment,
+};
+
+function classifyAttachments(attachments: TestAttachment[]) {
+  const screenshots = new Set(attachments.filter(a => a.contentType.startsWith('image/')));
+  const videos = attachments.filter(a => a.name === 'video');
+  const traces = attachments.filter(a => a.name === 'trace');
+
+  const otherAttachments = new Set<TestAttachment>(attachments);
+  [...screenshots, ...videos, ...traces].forEach(a => otherAttachments.delete(a));
+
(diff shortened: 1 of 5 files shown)


packages/html-reporter/src/testResultView.tsx
```diff
@@ -117,38 +152,22 @@ const StepTreeItem: React.FC<{
...
 
 const ImageDiff: React.FunctionComponent<{
- actual: TestAttachment,
- expected: TestAttachment,
- diff?: TestAttachment,
-}> = ({ actual, expected, diff }) => {
-  const [selectedTab, setSelectedTab] = React.useState<string>('actual');
+ tabs: DiffTab[],
+}> = ({ tabs }) => {
+  // Pre-select a tab called "actual", if any.
```
dgozman, line 157: Why not "diff"?

packages/html-reporter/src/testResultView.tsx
```diff
@@ -26,30 +26,65 @@ import { AttachmentLink } from './links';
...
-      attachmentsMap.set(a.name, a);
+
+    const otherAttachments = new Set<TestAttachment>(attachments);
     [...screenshots, ...videos, ...traces].forEach(a => otherAttachments.delete(a));
-    return { attachmentsMap, screenshots, videos, otherAttachments, traces };
+
+    const snapshotNameToDiffTabs = new Map<string, DiffTab[]>();
+    let tabId = 0;
+    for (const attachment of attachments) {
+      const match = attachment.name.match(/^(.*)-(actual|diff|expected|previous)(\.[^.]+)?$/);
```
dgozman, line 52: Hmm.... I thought that if we find `foo-diff.png`, we would also pick up `foo-grayscale.png`, `foo-overlayed.png`, `foo-actual.png` and similar. wdyt?

aslushnikov: - if there's one from diff, actual or expected, then pick everybody.
- extract this all thing to `classifyAttachments`

dgozman (approved):

Mone71:
How to say my life. Now I live War Zone.

PR #12597, opened 2022-03-08
fix: tune down pinkness of element masks

aslushnikov:
Instead of using `#FF00FF` as a mask color, this patch starts using
Deep Pink color - `#FF1493`.


aslushnikov:
As per offline discussion, decided to keep it as-is.

PR #12601, opened 2022-03-08
docs: add name for screenshot animations option

yury-s:
Also ensure all enums have names.


mxschmitt (approved):
lgtm but we need to follow up since the docs generator does not recognize this as an enum (only if length > 1 as of today).

dgozman (approved):

yury-s:
> lgtm but we need to follow up since the docs generator does not recognize this as an enum (only if length > 1 as of today).

I've fixed it.

PR #12606, opened 2022-03-09
docs: clarify multiple before/after hooks execution order

### LINKED ISSUE #12604: [Question] after* hook execution order
Hi there!

The [documention](https://playwright.dev/docs/intro#using-test-hooks) on test hooks doesn't appear to specify a defined order for them; that is, given:
```ts
test.afterAll(() => console.log('1'));
test.afterAll(() => console.log('2'));
test.afterAll(() => console.log('3'));
```
Should the output be in the order `1` / `2` / `3`, or `3` / `2` / `1`, or is the order (explicitly!) undefined? This can be important if some of the hooks have dependencies (that is, one hooks needs another to _not_ have run yet to be in a state to run).

If possible, this should be in the documentation (either in the [guide part](https://playwright.dev/docs/intro#using-test-hooks), or the [API reference](https://playwright.dev/docs/api/class-test#test-after-all), but ideally both).

Thanks for reading!

yury-s:
Fixes #12604

### DIFF
--- a/docs/src/intro-js.md
+++ b/docs/src/intro-js.md
@@ -250,6 +250,7 @@ Here is a list of the pre-defined fixtures that you are likely to use most of th
 
 You can use `test.beforeAll` and `test.afterAll` hooks to set up and tear down resources shared between tests.
 And you can use `test.beforeEach` and `test.afterEach` hooks to set up and tear down resources for each test individually.
+If multiple hooks of the same kind are added, they will run in the order of their registration.
 
 ```js js-flavor=js
 // example.spec.js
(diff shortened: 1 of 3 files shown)


docs/src/intro-js.md
```diff
@@ -250,6 +250,7 @@ Here is a list of the pre-defined fixtures that you are likely to use most of th
 
 You can use `test.beforeAll` and `test.afterAll` hooks to set up and tear down resources shared between tests.
 And you can use `test.beforeEach` and `test.afterEach` hooks to set up and tear down resources for each test individually.
+If multiple hooks of the same kind are added, they will run in the order of their registration.
```
dgozman, line 253: I'd drop from here, it's the intro!

dgozman (approved):

PR #12615, opened 2022-03-09
fix(test-runner): ESM compatibly on Windows 

mxschmitt:
Lets keep ensuring that ESM and its friends are working because currently they are broken and users are affected.

Relates: 
- https://github.com/microsoft/playwright/discussions/11624#discussioncomment-2069146
- https://github.com/microsoft/playwright/issues/12307
- https://github.com/sveltejs/kit/issues/4194

### DIFF
--- a/.github/workflows/tests_primary.yml
+++ b/.github/workflows/tests_primary.yml
@@ -56,9 +56,6 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         node-version: [12]
-        include:
-        - os: ubuntu-latest
-          node-version: 16
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


dgozman (commented):
Can we run just the ESM tests on win/linux node16 bots?

rwoll (approved):

.github/workflows/tests_primary.yml
```diff
@@ -78,6 +78,22 @@ jobs:
...
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 16
+    - run: npm i -g npm@8.3
+    - run: npm ci
+      env:
+        DEBUG: pw:install
+    - run: npm run build
+    - run: npx playwright install --with-deps
+    - run: npm run ttest -- --grep=esm
```
dgozman, line 95: Let's follow https://playwright.dev/docs/test-annotations#tag-tests:
`npm run ttest -- --grep=@esm`

.github/workflows/tests_primary.yml
```diff
@@ -78,6 +78,22 @@ jobs:
     - run: ./utils/upload_flakiness_dashboard.sh ./test-results/report.json
       if: always()
 
+  test_test_runner_windows_esm:
```
dgozman, line 81: Let's move ubuntu-latest 16 here as well!

.github/workflows/tests_primary.yml
mxschmitt: was doing it with `#` because we used it for our smoke tests: `#smoke`. Shall I change #smoke -> @esm in a follow-up too.

dgozman: Yeah, why not.

PR #12657, opened 2022-03-10
chore: add experimental mount

pavelfeldman:


packages/html-reporter/playwright.config.ts
```diff
@@ -27,6 +27,7 @@ const config: PlaywrightTestConfig = {
     ['html', { open: 'on-failure' }]
   ],
   use: {
+    baseURL: 'file://' + path.join(__dirname, 'out-ct', 'index.html').replace(path.sep, '/'),
```
mxschmitt, line 30:
```suggestion
    baseURL: 'file://' + path.join(__dirname, 'out-ct', 'index.html').replace(path.sep, '/'),
```

-> url.pathToFileURL()

packages/playwright-test/src/mount.ts
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+
+import { Page } from '@playwright/test';
+
+export async function _mount(page: Page, component: any): Promise<string> {
+  const callbacks: Function[] = [];
+  wrapFunctions(component, page, callbacks);
+
+  await page.exposeFunction('__pw_dispatch_', (ordinal: number, args: any[]) => {
```
dgozman, line 23: Make this work multiple times on a single page.

dgozman (approved):

packages/playwright-test/src/tsxTransform.ts
```diff
@@ -23,6 +23,42 @@ export default declare(api => {
...
+        const components = path.parentPath.getData('pw-components');
+        for (const specifier of importNode.specifiers) {
+          if (t.isImportDefaultSpecifier(specifier))
+            components.set(specifier.local.name, specifier.local.name);
+          if (t.isImportSpecifier(specifier) && t.isIdentifier(specifier.imported))
+            components.set(specifier.local.name, specifier.imported.name);
+        }
+
+        const importNodes = (path.parentPath.node as t.Program).body.filter(i => t.isImportDeclaration(i));
+        const isLast = importNodes.indexOf(importNode) === importNodes.length - 1;
```
dgozman, line 50: I think this will not work when last import is not a vue/svelte import or is not a string literal.

dgozman (approved):

PR #12660, opened 2022-03-10
feat(test-runner): more default workers for M1

rwoll:
New default is based on the very (unscientific) output of the following repeated a few
times:

```
$ for i in `seq 5 13`; do time npm run ctest -- --reporter=line --workers=$i 'page/*' 'selector*' 'channels*' 'resource-timing*'; done
```

NB: Depending on the user's version of Node, their platform will be
reported as x86 and not arm, so this won't make a difference for them.

### DIFF
--- a/packages/playwright-test/src/cli.ts
+++ b/packages/playwright-test/src/cli.ts
@@ -27,7 +27,7 @@ import { FilePatternFilter } from './util';
 import { showHTMLReport } from './reporters/html';
 import { GridServer } from 'playwright-core/lib/grid/gridServer';
 import dockerFactory from 'playwright-core/lib/grid/dockerGridFactory';
-import { createGuid } from 'playwright-core/lib/utils/utils';
+import { createGuid, hostPlatform } from 'playwright-core/lib/utils/utils';
 import { fileIsModule } from './loader';
 
 const defaultTimeout = 30000;
@@ -109,8 +109,7 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
 
   const os = require('os');
   const cpus = os.cpus().length;
-  const isLikelyM1 = os.arch() === 'arm64' && os.platform() === 'darwin';
-  const workers = isLikelyM1 ? cpus : Math.ceil(cpus / 2);
+  const workers = hostPlatform.startsWith('mac') && hostPlatform.endsWith('arm64') ? cpus : Math.ceil(cpus / 2);
 
   const defaultConfig: Config = {
     preserveOutput: 'always',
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):
LGTM

packages/playwright-test/src/cli.ts
```diff
@@ -107,13 +107,18 @@ Examples:
 async function runTests(args: string[], opts: { [key: string]: any }) {
   await startProfiling();
 
+  const os = require('os');
+  const cpus = os.cpus().length;
+  const isLikelyM1 = os.arch() === 'arm64' && os.platform() === 'darwin';
```
dgozman, line 112: We have a version that is not fooled by Rosetta: https://github.com/microsoft/playwright/blob/10bf5f3e493d25a72eda0f530590bafbab46edc5/packages/playwright-core/src/utils/utils.ts#L515-L535

packages/playwright-test/src/cli.ts
```diff
@@ -107,13 +107,17 @@ Examples:
 async function runTests(args: string[], opts: { [key: string]: any }) {
   await startProfiling();
 
+  const os = require('os');
```
aslushnikov, line 110: why isn't this a global import like everything else?

aslushnikov (approved):

packages/playwright-test/src/cli.ts
rwoll: Let me change that!

It was initially written like that (I'm assuming to be lazily loaded): https://github.com/microsoft/playwright/pull/6816/files#diff-b8be30c0cfa92ae9e4ace7585d93f8e19d3cb5896683821d6c83878f00329124

But looking into it further…since we import `playwright-core/lib/utils/utils` (even without this change), that's no longer relevant since that transitively always imports `os`.

scottopherson:
Is there more information about why M1's can supposedly have more workers vs all other architectures? Ever since this change I've had to manually limit the number of workers on my local machine because my test suite has random failures if all my cores are being used. Perhaps there's something wrong with my test suite but I haven't pinned down the issue yet so not sure.

scottopherson:
Other architectures don't benefit from 100% CPU core utilization? I presume the initial default of dividing the core count in half was chosen for a reason 🤔

aslushnikov:
> I presume the initial default of dividing the core count in half was chosen for a reason 🤔

@scottopherson it was based on measurements we did on different hardware we had around. More measurements showed that M1 Max performs much better than average, thus the change. I myself run M1 Max and enjoy the extra speed :)

But this all is not very scientific. What kind of M1 do you have? Is it the original M1, M1 Pro or M1 Max?

PR #12663, opened 2022-03-10
chore: add `"allow"` as a default value for animations.

aslushnikov:


docs/src/api/params.md
```diff
@@ -902,12 +902,14 @@ Note that outer and inner locators must belong to the same frame. Inner locator
...
 ## screenshot-option-animations
-- `animations` <[ScreenshotAnimations]<"disabled">>
+- `animations` <[ScreenshotAnimations]<"disabled"|"allow">>
 
 When set to `"disabled"`, stops CSS animations, CSS transitions and Web Animations. Animations get different treatment depending on their duration:
 * finite animations are fast-forwarded to completion, so they'll fire `transitionend` event.
 * infinite animations are canceled to initial state, and then played over after the screenshot.
 
+Defaults to `"allow"`.
+
```
dgozman, line 912: Add a test that uses `allow` and you'll see that we have to change protocol.yml for it to work :)

dgozman (approved):

PR #12664, opened 2022-03-10
Fixing example code in `accessibility` API docs

andrewhayward:
The example code for finding the focused node doesn't work as described. While this is not a particularly big deal, it should probably do what it says it does!

This patch updates the existing JS/Python examples to work as intended, but does't add any new code for missing languages.


yury-s (approved):

aslushnikov:
@andrewhayward could you please sign CLA one more time? The bot is unhappy

mxschmitt (approved):

PR #12673, opened 2022-03-11
chore: dogfood dotenv

rwoll:
Before this change, I had to use the CLI version of Playwright instead
of fully dogfooding the VSCode extension since I was running with a
custom `CRPATH`.

### DIFF
--- a/tests/config/android.config.ts
+++ b/tests/config/android.config.ts
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+import { config as loadEnv } from 'dotenv';
+loadEnv({ path: path.join(__dirname, '..', '..', '.env') });
+
 import type { Config, PlaywrightTestOptions, PlaywrightWorkerOptions } from '@playwright/test';
 import * as path from 'path';
 import { ServerWorkerOptions } from './serverFixtures';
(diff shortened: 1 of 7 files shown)


mxschmitt:
not sure if this is worth it, I'd rather add it to the docs section instead, wdyt?

rwoll:
> not sure if this is worth it, I'd rather add it to the docs section instead, wdyt?

I think it is! Prior to this I could use the extension, or I needed to look through our configs, find the correct one, (temporarily) add a process.env line.

This makes it much easier, IMO, since you can edit the .env in one place, it works with extension and command line, and your temp overrides don't get checked in.

tests/config/android.config.ts
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+import { config as loadEnv } from 'dotenv';
+loadEnv({ path: path.join(__dirname, '..', '..', '.env') });
```
dgozman, line 18: Does this work if I don't have the `.env` file?

dgozman (approved):
I like this!

tests/config/android.config.ts
mxschmitt: Yes, otherwise the bots wouldn't have passed!

mxschmitt (approved):

PR #12677, opened 2022-03-11
feat(toHaveScreenshot): enhance messaging in case of failures

### LINKED ISSUE #12563: [BUG] toHaveScreenshot: more information in call log when screenshots differ
See https://playwright.dev/docs/next/test-assertions#custom-expect-message

aslushnikov:
This patch:
- adds call logs to track screenshot timeouts, e.g. due to
  waiting for web fonts
- makes sure all snapshot expectations have `.png` extension
- throws a polite error when given a buffer or a string instead of a
  page or a locator
- removes stray NL between error description and call log

image

Fixes #12563

### DIFF
--- a/packages/playwright-core/src/server/page.ts
+++ b/packages/playwright-core/src/server/page.ts
@@ -458,16 +458,24 @@ export class Page extends SdkObject {
       errorMessage?: string,
       diff?: Buffer,
     } | undefined = undefined;
+    const callTimeout = this._timeoutSettings.timeout(options);
     return controller.run(async progress => {
       let actual: Buffer | undefined;
       let previous: Buffer | undefined;
       const pollIntervals = [0, 100, 250, 500];
+      progress.log(`${metadata.apiName}${callTimeout ? ` with timeout ${callTimeout}ms` : ''}`);
+      if (isGeneratingNewScreenshot)
+        progress.log(`  generating new screenshot expectation: waiting for 2 consecutive screenshots to match`);
+      else
+        progress.log(`  waiting for screenshot to match expectation`);
       while (true) {
         progress.throwIfAborted();
         if (this.isClosed())
           throw new Error('The page has closed');
         let comparatorResult: ComparatorResult | undefined;
         const screenshotTimeout = pollIntervals.shift() ?? 1000;
+        if (screenshotTimeout)
+          progress.log(`waiting ${screenshotTimeout}ms before taking screenshot`);
         if (isGeneratingNewScreenshot) {
           previous = actual;
           actual = await rafrafScreenshot(progress, screenshotTimeout).catch(e => undefined);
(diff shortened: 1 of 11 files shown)


packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -138,6 +140,10 @@ export class Screenshotter {
   }
 
   async _preparePageForScreenshot(progress: Progress, disableAnimations: boolean, waitForFonts: boolean) {
+    if (disableAnimations)
+      progress.log('  disabled all CSS animations because "animations" option is set to "disabled"');
```
dgozman, line 144: `"disabled"` does not play well with languages - let's drop it? Same for `"ready"`.

packages/playwright-test/src/expect.ts
```diff
@@ -257,6 +257,7 @@ function wrap(matcherName: string, matcher: any) {
       reportStepError(e);
     }
   };
+  Object.defineProperty(matcher, 'name', { value: '__PWTRAP__[expect.' + matcherName + ']' });
```
dgozman, line 260: Why this change?

packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -203,11 +203,12 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
     if (diffError) {
       output.push(...[
```
dgozman, line 204:
```suggestion
      output.push(indent(diffError, '  '));
```

packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -289,6 +290,10 @@ export async function toHaveScreenshot(
       testInfo, testInfo._screenshotPath.bind(testInfo), 'png',
       testInfo.project.expect?.toHaveScreenshot || {},
       nameOrOptions, optOptions);
+  if (!helper.snapshotPath.toLowerCase().endsWith('.png'))
+    throw new Error(`Screenshot name "${path.basename(helper.snapshotPath)}" must have '.png' extension`);
+  if (Buffer.isBuffer(pageOrLocator) || (typeof pageOrLocator === 'string'))
```
dgozman, line 295: There is an `expectType` helper in `playwright-test/src/util.ts`.

packages/playwright-core/src/server/page.ts
```diff
@@ -458,16 +458,24 @@ export class Page extends SdkObject {
...
     } | undefined = undefined;
+    const callTimeout = this._timeoutSettings.timeout(options);
     return controller.run(async progress => {
       let actual: Buffer | undefined;
       let previous: Buffer | undefined;
       const pollIntervals = [0, 100, 250, 500];
-      while (true) {
+      progress.log(`${metadata.apiName}${callTimeout ? ` with timeout ${callTimeout}ms` : ''}`);
+      if (isGeneratingNewScreenshot)
+        progress.log(`  generating new screenshot expectation: waiting for 2 consecutive screenshots to match.`);
```
dgozman, line 468: This is the only line that has a full stop at the end :)

packages/playwright-core/src/server/page.ts
```diff
@@ -458,16 +458,24 @@ export class Page extends SdkObject {
...
+      else
+        progress.log(`  waiting for screenshot to match expectation`);
+      for (let iteration = 0; true; ++iteration) {
```
dgozman, line 471: I guess `iteration` was used at some point, but not anymore?

packages/playwright-test/src/matchers/toMatchSnapshot.ts
aslushnikov: check for page or locator explicitly

packages/playwright-test/src/expect.ts
aslushnikov: i'll extract separately.

dgozman (approved):

PR #12697, opened 2022-03-12
chore: prepare to publish component packages

pavelfeldman:


dgozman (approved):

packages/playwright-ct-vue/package.json
```diff
@@ -1,8 +1,22 @@
 {
-    "name": "@playwright/ct-vue",
+    "name": "@playwright/experimental-ct-vue",
```
mxschmitt, line 2: The last time we prefixed a package with experimental it went wrong and people kept using it for years in blog articles (allure). Also they didn't known about it's successor.

Let's use the right name but a 0.x version which is more common?

tests-components/ct-react-vite/package.json
```diff
@@ -17,5 +17,9 @@
     "@vitejs/plugin-react": "^1.0.7",
     "typescript": "^4.5.4",
     "vite": "^2.8.0"
+  },
+  "@standaloneDevDependencies": {
```
mxschmitt, line 21: Who invented that?

packages/playwright-ct-vue/package.json
pavelfeldman: The problem is that we don't yet have the 'right name'.

tests-components/ct-react-vite/package.json
pavelfeldman: me!

PR #12706, opened 2022-03-12
fix: properly define apiName for web-first assertions

aslushnikov:
Turns out relying on PWTRAP in stack is not reliable: depending on the
call structure, the stack might be cut unpredictably by Node.js.

This patch removes PWTRAP and instead plumbs explicit stack and
pre-set `apiName` all the way down to `wrapApiCall`.

### DIFF
--- a/packages/playwright-test/src/expect.ts
+++ b/packages/playwright-test/src/expect.ts
@@ -47,8 +47,7 @@ import { toMatchSnapshot, toHaveScreenshot, getSnapshotName } from './matchers/t
 import type { Expect, TestError } from './types';
 import matchers from 'expect/build/matchers';
 import { currentTestInfo } from './globals';
-import { serializeError } from './util';
-import { captureStackTrace } from 'playwright-core/lib/utils/stackTrace';
+import { serializeError, captureStackTrace } from './util';
 
 // #region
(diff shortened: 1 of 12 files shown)


packages/playwright-test/src/expect.ts
```diff
@@ -48,9 +48,7 @@ import type { Expect, TestError } from './types';
 import matchers from 'expect/build/matchers';
 import { currentTestInfo } from './globals';
 import { serializeError } from './util';
-import StackUtils from 'stack-utils';
-
-const stackUtils = new StackUtils();
+import { captureStackTrace } from 'playwright-core/lib/utils/stackTrace';
```
dgozman, line 51: Suggested wrapper in `playwright-test/src/util` that removes notion of `expect` from core entirely.

```js
function captureStackTrace() {
  const stack = require('...core').captureStackTrace();
  for (let i = stack.frames.length - 1; i >= 0; i--) {
    if (stack.frames[i] is EXPECT) {
      stack.frames.splice(i, 1, 0);
      stack.frameTexts.splice(i, 1, 0);
    }
  }
  return stack;
}
```

dgozman (approved):

Issue #9888, opened 2021-10-29
[Feature] Automagically Run xvfb-run if no DISPLAY set and head full


rwoll:
In slack, it's fairly common to see folks get the DISPLAY error in CI/CD, so it might be friendlier to auto detect this and automatically wrap the browser launch in xvfb-run internally in Playwright so it's transparent for the user.

Since our setup has a mix of headful and headless projects, we just always wrap our test launches with xvfb-run so we don't have to have conditional logic dealing with each of the different projects. If playwright auto detected the need for xvfb-run wrap, we wouldn't need waste calling it AND we wouldn't have to remember to add xvfb-run wrap when we add new CI/CD workflows.

mxschmitt:
Relates https://github.com/microsoft/playwright/issues/6037

Issue #6037, opened 2021-04-01
[Feature] Link to docs page when "Unable to open X display" gets logged.


mxschmitt:
Link to that: https://playwright.dev/docs/ci#running-headful

Relates to https://github.com/microsoft/playwright-python/issues/607

Would probably help users that they fix it themselves instead of opening an issue.

(not sure if the output on all the browsers is the same.)

mxschmitt:
Maybe `xvfb-run` should be integrated into Playwright when using headful? That would add more magic but also more comfort to the user-experience.

dgozman:
@yury-s, what do you think about this?

mxschmitt:
Linking https://github.com/microsoft/playwright/issues/8393

pavelfeldman:
#### Why was this issue closed?

We are prioritizing the bugs based on the upvotes, recency and our ability to act. It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it. We are closing issues like this one to keep our bug database maintainable. Please feel free to open a new issue and link this one to it if you think this is a mistake.

SgtPooki:
Not needing any action from playwright here, but FYI that playwright, service workers, and iframes caused me a lot of pain in https://github.com/ipfs-examples/js-ipfs-examples/pull/527 and I had to remove the iframes entirely to get CI to pass in github CI. If anyone knows a solution that I missed, please update me =)

PR #12724, opened 2022-03-14
chore: improve handling with missing xserver

mxschmitt:
We could improve it in different ways:
- a) if `headless: false` and no `process.env.DISPLAY` configured and os is linux , then yell
- b) if `headless: false` and no `process.env.DISPLAY` configured and os is linux , then automatically try to start a xserver
- c) if browser does not start because of missing xserver (we compare the error message), then yell 

Currently c) is implemented in this PR, but I'm probably more for a). Not sure how reliable this is tho.

Relates: 
- https://github.com/microsoft/playwright/issues/9888
- https://github.com/microsoft/playwright/discussions/12704
- https://github.com/microsoft/playwright/issues/6037

### DIFF
--- a/packages/playwright-core/src/server/browserType.ts
+++ b/packages/playwright-core/src/server/browserType.ts
@@ -31,6 +31,9 @@ import { helper } from './helper';
 import { RecentLogsCollector } from '../utils/debugLogger';
 import { CallMetadata, SdkObject } from './instrumentation';
 
+export const kNoXServerRunningError = 'Looks like you launched a headed browser without having a XServer running.\n' +
+  'Set either \'headless: false\' or use \'xvfb-run <your-playwright-app>\' before running Playwright.\n\n<3 Playwright Team';
+
 export abstract class BrowserType extends SdkObject {
   private _name: BrowserName;
   readonly _playwrightOptions: PlaywrightOptions;
@@ -76,7 +79,7 @@ export abstract class BrowserType extends SdkObject {
 
   async _innerLaunchWithRetries(progress: Progress, options: types.LaunchOptions, persistent: types.BrowserContextOptions | undefined, protocolLogger: types.ProtocolLogger, userDataDir?: string): Promise<Browser> {
     try {
-      return this._innerLaunch(progress, options, persistent, protocolLogger, userDataDir);
+      return await this._innerLaunch(progress, options, persistent, protocolLogger, userDataDir);
     } catch (error) {
       // @see https://github.com/microsoft/playwright/issues/5214
       const errorMessage = typeof error === 'object' && typeof error.message === 'string' ? error.message : '';
--- a/packages/playwright-core/src/server/chromium/chromium.ts
+++ b/packages/playwright-core/src/server/chromium/chromium.ts
@@ -22,12 +22,12 @@ import { CRBrowser } from './crBrowser';
 import { Env, gracefullyCloseSet } from '../../utils/processLauncher';
 import { kBrowserCloseMessageId } from './crConnection';
 import { rewriteErrorMessage } from '../../utils/stackTrace';
-import { BrowserType } from '../browserType';
+import { BrowserType, kNoXServerRunningError } from '../browserType';
 import { ConnectionTransport, ProtocolRequest, WebSocketTransport } from '../transport';
 import { CRDevTools } from './crDevTools';
 import { Browser, BrowserOptions, BrowserProcess, PlaywrightOptions } from '../browser';
 import * as types from '../types';
-import { debugMode, fetchData, getUserAgent, headersArrayToObject, HTTPRequestParams, removeFolders, streamToString } from '../../utils/utils';
+import { debugMode, fetchData, getUserAgent, headersArrayToObject, HTTPRequestParams, removeFolders, streamToString, wrapInASCIIBox } from '../../utils/utils';
 import { RecentLogsCollector } from '../../utils/debugLogger';
 import { Progress, ProgressController } from '../progress';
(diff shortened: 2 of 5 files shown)


packages/playwright-core/src/server/browserType.ts
```diff
@@ -76,14 +77,28 @@ export abstract class BrowserType extends SdkObject {
...
+        // Chromium
+        errorMessage.includes('Missing X server') ||
+        // WebKit
+        errorMessage.includes('cannot open display') ||
+        // Firefox
+        errorMessage.includes('no DISPLAY environment variable specified')
+      ) {
+        rewriteErrorMessage(error, '\n' + wrapInASCIIBox(
+            'Looks like you launched a headed browser without having a XServer running.\n' +
+          'Set either \'headless: false\' or use \'xvfb-run <your-playwright-app>\' before running Playwright'
```
rwoll, line 98: Is it possible to get here on Windows? If so the Xvfb tip might be misleading.

rwoll (commented):
Overall looks good and nice to have a more explicit/friendly error for users to debug.

Option (b) of auto-adding Xvfb when needed would be particularly helpful on GitHub Codespaces.

yury-s:
> (a) and (b) rely on DISLPAY env variable. Do all window servers (e.g. wayland) expose it?

Looks like at least sometimes it can be [WAYLAND_DISPLAY](https://wayland-book.com/wayland-display/creation.html).

packages/playwright-core/src/server/browserType.ts
mxschmitt: there is no such a thing in Windows, so I doubt it.

aslushnikov:
If https://github.com/microsoft/playwright/pull/12806 lands, we can auto-run xvfb inside of our docker container.

packages/playwright-core/src/server/browserType.ts
```diff
@@ -76,14 +77,28 @@ export abstract class BrowserType extends SdkObject {
...
       // @see https://github.com/microsoft/playwright/issues/5214
-      const errorMessage = typeof error === 'object' && typeof error.message === 'string' ? error.message : '';
+      const errorMessage: string = typeof error === 'object' && typeof error.message === 'string' ? error.message : '';
       if (errorMessage.includes('Inconsistency detected by ld.so')) {
         progress.log(`<restarting browser due to hitting race condition in glibc>`);
         return this._innerLaunch(progress, options, persistent, protocolLogger, userDataDir);
       }
+      if (
+        // Chromium
+        errorMessage.includes('Missing X server') ||
```
aslushnikov, line 90: Shouldn't this all go to `_rewriteStartupError`?

mxschmitt: done

mxschmitt:
Error looks now like this:

```
root@abc4a32a047f:/work# node test.js
node:internal/process/promises:265
            triggerUncaughtException(err, true /* fromPromise */);
            ^

browserType.launch: 
╔════════════════════════════════════════════════════════════════════════════════════════════════╗
║ Looks like you launched a headed browser without having a XServer running.                     ║
║ Set either 'headless: false' or use 'xvfb-run <your-playwright-app>' before running Playwright ║
║                                                                                                ║
║ <3 Playwright Team                                                                             ║
╚════════════════════════════════════════════════════════════════════════════════════════════════╝
<launching> /ms-playwright/chromium-980066/chrome-linux/chrome --disable-background-networking --enable-features=NetworkService,NetworkServiceInProcess --disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-breakpad --disable-client-side-phishing-detection --disable-component-extensions-with-background-pages --disable-default-apps --disable-dev-shm-usage --disable-extensions --disable-features=ImprovedCookieControls,LazyFrameLoading,GlobalMediaControls,DestroyProfileOnBrowserClose,MediaRouter,AcceptCHFrame,AutoExpandDetailsElement --allow-pre-commit-input --disable-hang-monitor --disable-ipc-flooding-protection --disable-popup-blocking --disable-prompt-on-repost --disable-renderer-backgrounding --disable-sync --force-color-profile=srgb --metrics-recording-only --no-first-run --enable-automation --password-store=basic --use-mock-keychain --no-service-autorun --export-tagged-pdf --no-sandbox --user-data-dir=/tmp/playwright_chromiumdev_profile-nzgcMS --remote-debugging-pipe --no-startup-window
<launched> pid=306
[pid=306][err] [306:306:0318/180622.264244:ERROR:ozone_platform_x11.cc(247)] Missing X server or $DISPLAY
[pid=306][err] [306:306:0318/180622.264665:ERROR:env.cc(225)] The platform failed to initialize.  Exiting.
    at /work/test.js:4:34
    at Object.<anonymous> (/work/test.js:7:3) {
  name: 'Error'
}
```

packages/playwright-core/src/server/browserType.ts
```diff
@@ -76,7 +79,7 @@ export abstract class BrowserType extends SdkObject {
 
   async _innerLaunchWithRetries(progress: Progress, options: types.LaunchOptions, persistent: types.BrowserContextOptions | undefined, protocolLogger: types.ProtocolLogger, userDataDir?: string): Promise<Browser> {
     try {
-      return this._innerLaunch(progress, options, persistent, protocolLogger, userDataDir);
+      return await this._innerLaunch(progress, options, persistent, protocolLogger, userDataDir);
```
mxschmitt, line 82: I have the assumption that the actual stuff underneath with checking the error was never working. This should fix it!

aslushnikov (approved):
awesome! Can we have a test?

PR #12734, opened 2022-03-14
feat(html-reporter): add report context header

### LINKED ISSUE #11318: [Feature] HTML Report: Commit Context Header/Sidebar
**tl;dr:** We produce a lot of HTML Reports and as we share links to them in tickets or with teammates, it would be nice for them to have some additional context in the header of the report so we know who/what/when/etc the report was produced (most of them come from CI bots with each commit we push). If you link the report to a ticket and come back to it days/weeks later[^1], or have lots of tabs open this information becomes especially useful in understanding when/if the report is still relevant or under what conditions it was created (e.g. staging v. prod envs).

The following info would be nice to see somewhere prominently in the report (and as always, I'm happy to contribute a PR):

| Information | Info |
| ----------- | ------ |
| Start Time (Wall Time) | Generated by HTML Reporter. |
| End Time (Wall Time) | Generated by HTML Reporter. |
| Commit Author | Sourced from `PLAYWRIGHT_REPORT_ENV_COMMIT_AUTHOR` (Explicitly set, or inferred from `CI_COMMIT_AUTHOR` in GitLab, and `??` in GitHub) |
| Commit Message | Sourced from `PLAYWRIGHT_REPORT_ENV_COMMIT_MESSAGE` (Explicitly set, or inferred from `CI_COMMIT_MESSAGE` in GitLab, and `??` in GitHub) |
| Commit Hash | Sourced from `PLAYWRIGHT_REPORT_ENV_COMMIT_HASH` (Explicitly set, or inferred from `CI_COMMIT_SHA` in GitLab, and `GITHUB_SHA` in GitHub) |
| CI/CD Link | Sourced from `PLAYWRIGHT_REPORT_ENV_CI_CD_LINK` (Explicitly set, or inferred from `CI_JOB_URL` in GitLab, and `??` in GitHub) |
| Notes | Freeform notes for the user to supply via `PLAYWRIGHT_REPORT_ENV_NOTES`. |

Here's an example of what it might look like (modulo some better styling[^2]):

Screen Shot 2022-01-10 at 7 33 05 PM

I picked these fields as they seem universally helpful and not unique to us. We receive test failure information via Slack notifications that also link to the (hosted) HTML Report, so the HTML Report is quickly becoming our entry point into test triaging etc.

If this is something the team/community supports, **please let me know and I will be happy to make a quick PR to the HTML Report** so the information in these variables gets inlined and presented in the UI like is done today with the the other test-time inlined report data/json.

:tada: :heart: 😄 

[^1]: We embed the timestamp and some metadata in the path to the hosted report, so we can sort of infer some minimal context now. We addionally capture context in test attachments, but that info is not discoverable when you first click on a report.

[^2]: I'd probably also make it collapsable.

rwoll:
Resolves #11318.

* Adds `TestConfig.attachments` public API. (We opted to not implement an analog to the async `TestInfo.attach(…)` API.)
* Adds `TestConfig.attachments` to common reporters.
* Dogfoods some git and CI-info inference to generate useful atttachments
* Updates HTML Reporter to include a side bar to present a pre-defined set of attachments (a.k.a git/commit context sidebar)

Here's what it looks like:

Screen Shot 2022-03-21 at 3 23 28 PM

See `tests/playwright-test/reporter-html.spec.ts` for an example of usage (for dogfood-ing only). In the future, if this becomes user-facing, there the Global Setup bit would likely become unnecessary (as would interaction with attachments array); there would likely just be a nice top-level config and/or CLI flag to enable collecting of info.

### DIFF
--- a/packages/html-reporter/src/index.tsx
+++ b/packages/html-reporter/src/index.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import type { HTMLReport } from '@playwright/test/src/reporters/html';
+import type { HTMLReport, TestAttachment } from '@playwright/test/src/reporters/html';
 import type zip from '@zip.js/zip.js';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
@@ -22,24 +22,40 @@ import './colors.css';
 import { LoadedReport } from './loadedReport';
 import { ReportView } from './reportView';
 
-export type Metadata = {
-  generatedAt: number;
-  git: {
-      commit: {
-          author: {
-              name: string;
-              email: string;
-          };
-          sha: string;
-          subject: string;
-          timestamp: number;
-          link: string | undefined;
-      };
-      clean: boolean;
+export type Metadata = Partial<{
+  'generatedAt': number;
+  'revision.id': string;
+  'revision.author': string;
+  'revision.email': string;
+  'revision.subject': string;
+  'revision.timestamp': number;
+  'revision.link': string;
+  'revision.localPendingChanges': boolean;
+  'ci.link': string;
+}>;
+
+const extractMetadata = (attachments: TestAttachment[]): Metadata | undefined => {
+  const field = (name: string) => attachments.find(({ name: n }) => n === name)?.body;
+  const fieldAsJSON = (name: string) => {
+    const raw = field(name);
+    if (raw !== undefined)
+      return JSON.parse(raw);
   };
-  ci: {
-      link: string | undefined;
+
+  const out = {
+    'generatedAt': fieldAsJSON('generatedAt'),
+    'revision.id': field('revision.id'),
+    'revision.author': field('revision.author'),
+    'revision.email': field('revision.email'),
+    'revision.subject': field('revision.subject'),
+    'revision.timestamp': fieldAsJSON('revision.timestamp'),
+    'revision.link': field('revision.link'),
+    'revision.localPendingChanges': fieldAsJSON('revision.localPendingChanges'),
+    'ci.link': field('ci.link'),
   };
+
+  if (Object.entries(out).filter(([_, v]) => v !== undefined).length)
+    return out;
 };
 
 const zipjs = (self as any).zip;
@@ -68,9 +84,7 @@ class ZipReport implements LoadedReport {
     for (const entry of await zipReader.getEntries())
       this._entries.set(entry.filename, entry);
     this._json = await this.entry('report.json') as HTMLReport;
-    const rawMetadata = this._json.attachments.find(v => v.name === 'ci-info')?.body;
-    if (rawMetadata)
-      this._json.metadata = JSON.parse(rawMetadata);
+    this._json.metadata = extractMetadata(this._json.attachments);
   }
 
   json(): HTMLReport {
--- a/packages/html-reporter/src/index.tsx
+++ b/packages/html-reporter/src/index.tsx
@@ -22,6 +22,26 @@ import './colors.css';
 import { LoadedReport } from './loadedReport';
 import { ReportView } from './reportView';
 
+export type Metadata = {
+  generatedAt: number;
+  git: {
+      commit: {
+          author: {
+              name: string;
+              email: string;
+          };
+          sha: string;
+          subject: string;
+          timestamp: number;
+          link: string | undefined;
+      };
+      clean: boolean;
+  };
+  ci: {
+      link: string | undefined;
+  };
+};
+
 const zipjs = (self as any).zip;
 
 const ReportLoader: React.FC = () => {
@@ -41,13 +61,16 @@ window.onload = () => {
 
 class ZipReport implements LoadedReport {
   private _entries = new Map<string, zip.Entry>();
-  private _json!: HTMLReport;
+  private _json!: HTMLReport & { metadata?: Metadata };
 
   async load() {
     const zipReader = new zipjs.ZipReader(new zipjs.Data64URIReader(window.playwrightReportBase64), { useWebWorkers: false }) as zip.ZipReader;
     for (const entry of await zipReader.getEntries())
       this._entries.set(entry.filename, entry);
     this._json = await this.entry('report.json') as HTMLReport;
+    const rawMetadata = this._json.attachments.find(v => v.name === 'ci-info')?.body;
+    if (rawMetadata)
+      this._json.metadata = JSON.parse(rawMetadata);
   }
 
   json(): HTMLReport {
--- a/packages/html-reporter/src/index.tsx
+++ b/packages/html-reporter/src/index.tsx
@@ -35,22 +35,39 @@ export type Metadata = Partial<{
 }>;
 
 const extractMetadata = (attachments: TestAttachment[]): Metadata | undefined => {
+  // The last plugin to register for a given key will take precedence
+  attachments = [...attachments];
+  attachments.reverse();
   const field = (name: string) => attachments.find(({ name: n }) => n === name)?.body;
   const fieldAsJSON = (name: string) => {
     const raw = field(name);
     if (raw !== undefined)
       return JSON.parse(raw);
   };
+  const fieldAsNumber = (name: string) => {
+    const v = fieldAsJSON(name);
+    if (v !== undefined && typeof v !== 'number')
+      throw new Error(`Invalid value for field '${name}'. Expected type 'number', but got ${typeof v}.`);
+
+    return v;
+  };
+  const fieldAsBool = (name: string) => {
+    const v = fieldAsJSON(name);
+    if (v !== undefined && typeof v !== 'boolean')
+      throw new Error(`Invalid value for field '${name}'. Expected type 'boolean', but got ${typeof v}.`);
+
+    return v;
+  };
 
   const out = {
-    'generatedAt': fieldAsJSON('generatedAt'),
+    'generatedAt': fieldAsNumber('generatedAt'),
     'revision.id': field('revision.id'),
     'revision.author': field('revision.author'),
     'revision.email': field('revision.email'),
     'revision.subject': field('revision.subject'),
-    'revision.timestamp': fieldAsJSON('revision.timestamp'),
+    'revision.timestamp': fieldAsNumber('revision.timestamp'),
     'revision.link': field('revision.link'),
-    'revision.localPendingChanges': fieldAsJSON('revision.localPendingChanges'),
+    'revision.localPendingChanges': fieldAsBool('revision.localPendingChanges'),
     'ci.link': field('ci.link'),
   };
 
--- a/packages/html-reporter/src/reportView.css
+++ b/packages/html-reporter/src/reportView.css
@@ -24,7 +24,6 @@ html, body {
 
 body {
   overflow: auto;
-  max-width: 1024px;
   margin: 0 auto;
   width: 100%;
 }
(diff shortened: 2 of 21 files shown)


packages/playwright-test/src/reporters/html.ts
```diff
@@ -529,4 +537,16 @@ function isTextContentType(contentType: string) {
...
 }
 
+function metadata(): Metadata {
+  const githubJobLink = process.env.GITHUB_SERVER_URL ? `${process.env.GITHUB_SERVER_URL}/${process.env.GITHUB_REPOSITORY}/actions/runs/${process.env.GITHUB_RUN_ID}` : undefined;
+  const gitlabJobLink = process.env.CI_JOB_URL;
+  const jenkinsJobLink = process.env.JOB_URL;
+
+  return {
+    generatedAt: Date.now(),
+    title: process.env.PLAYWRIGHT_HTML_REPORT_TITLE || 'Playwright Test Report',
```
rwoll, line 547: Environment variables are used (instead of JS reporter options) since (1) it's nice to be able to easily call git from within your shell:

```
PLAYWRIGHT_HTML_REPORT_TITLE="$(git log --pretty=format:'%h: %s' -n 1)" npm run ctest
```

(2) `playwright.config.ts` reporter options do not work nicely with command line `--reporter` flag.

rwoll:
@pavelfeldman [PR description](https://github.com/microsoft/playwright/pull/12734#issue-1168571706) has updated examples.

Let me know if it looks OK to merge, and then we can make pixel-level adjustments! 😄 It's based on GitHub but diverges slightly around link coloring/behavior.

packages/html-reporter/src/reportView.css
```diff
@@ -33,8 +33,57 @@ body {
...
+  line-height: 1.5;
+}
+
 @media only screen and (max-width: 600px) {
   .report {
     padding: 0 !important;
   }
 }
+
+@media only screen and (max-width: 900px) {
```
rwoll, line 77: @pavelfeldman To not have anything above search box, should we just `display: none` the header at this point? If you look at the Break Point in the description, the header sticks around for a while so I'd totally be OK with it just being gone when we no longer have room for sidebar.

docs/src/test-reporters-js.md
```diff
@@ -302,6 +302,19 @@ Or if there is a custom folder name:
...
 
+A custom title for the report can be set via the `PLAYWRIGHT_HTML_REPORT_TITLE` environment variable. For example, in CI/CD it could be helpful to set it to the commit hash and subject:
+
+```
+PLAYWRIGHT_HTML_REPORT_TITLE="$(git log --pretty=format:'%h: %s' -n 1)" npx playwright test
+```
+
+When run in a CI/CD system, the HTML Reporter will attempt to infer and include a link to the CI/CD job. If it cannot be inferred, the `PLAYWRIGHT_HTML_REPORT_JOB_LINK` environment variable can be set explicitly.
+
+To opt out of including the link entirely, set `PLAYWRIGHT_HTML_REPORT_JOB_LINK` to the empty string:
```
aslushnikov, line 313: if we were to use `TestProject.metadata`, then we probably would not need this!

tests/playwright-test/reporter-html.spec.ts
```diff
@@ -592,3 +592,89 @@ test('open tests from required file', async ({ runInlineTest, showReport, page }
...
+
+    const [ jobPage ] = await Promise.all([
+      context.waitForEvent('page'),
+      metadata.locator('a').click(),
+    ]);
+
+    expect(jobPage.url()).toContain('job-link');
+  });
+
+  test('should infer job link', async ({ runInlineTest, showReport, page, context, server }) => {
```
aslushnikov, line 641:
```suggestion
  test('should infer job link for GitLab', async ({ runInlineTest, showReport, page, context, server }) => {
```

packages/html-reporter/src/reportView.css
```diff
@@ -33,8 +33,57 @@ body {
   border-top: 1px solid var(--color-border-default);
 }
 
+.htmlreport {
+  gap: 24px;
```
aslushnikov, line 37: non-existent!!

aslushnikov (approved):
maybe we can put all the meta information in `TestProject.meta`?

We already do it ourselves!

We can also call into `git` to infer all the commit details!

rwoll:
Preview of latest report (not yet pushed here; pending some config discussions):

Screen Shot 2022-03-16 at 7 03 03 PM

packages/html-reporter/src/reportView.css
rwoll: It actually does something in flex context! Confirmed via toggling it in DevTools, and [MDN agrees](https://developer.mozilla.org/en-US/docs/Web/CSS/gap?msclkid=21de9013a66c11ecbcbb669f896b9dfe):

> <html>
<body>


Applies to | multi-column elements, flex containers, grid containers
-- | --



</body>
</html>

rwoll:
@aslushnikov / @dgozman I just pushed updates based on today's discussion. There's still a bit of work to finish it up, but wanted to get feedback.

Visually, it's the same as before (and the screenshots in the PR description still apply), but the implementation is totally different. It's now attachment-based instead of metadata-based. Additionally, since we are going to dogfood this, if you're not dogfooding, the HTML Report is identical (functionally/visually) to what's in 1.20.

The PR introduces three overall changes:

1. `FullConfig.attachments` (which you can populate in Global Setup)
2. Includes these attachments in the reporters.
3. Adds a self contained helper in lib to populate `FullConfig.attachments` based on git inference and environment variable inference. (See our internal globalSetup.ts` for example usage.)
4. Updates the HTML Reporter to show a side bar with the git/ci info if (2) is used.

Notably, the PR lacks:

- [ ] JUnit Reporter Updates (I updated the other reporters to support these new top-level attachments, but need to do JUnit depending on how we proceed.)
- [ ] Docs: I'll add these before it lands, but not until we like the direction.
- [ ] Tests: Same as above (although I'd just bring back the tests from the previous version of this PR, plus add some around the git business).

As part of the review, we should decide how we want to land this. (i.e. which of 1-4 above will be part of the public API.)

As you review, there's some places where we should think about schema/layout in the reporters, so please comment there if you have thoughts. Additionally, I think we need `async FullConfig.attach(...)` method (not just the array) to mirror the API we  have for `testInfo`. wdyt?

packages/html-reporter/src/index.tsx
```diff
@@ -14,14 +14,50 @@
...
+  'revision.author': string;
+  'revision.email': string;
+  'revision.subject': string;
+  'revision.timestamp': number;
+  'revision.link': string;
+  'revision.localPendingChanges': boolean;
+  'ci.link': string;
+}>;
+
+const extractMetadata = (attachments: TestAttachment[]): Metadata | undefined => {
```
rwoll, line 37: This is a result of deciding to do an attachment per-field (instead of the previous iteration where we just put everything in one attachment and just did JSON.parse(…) once here).

We chose one-field-per-file so "plugins" don't have to be concerned with merging.

rwoll (commented):
@dgozman and @aslushnikov This is ready for a review pass. It's been updated based on the feedback from this morning. I brought my tests back, so it's a candidate to land.

Cheers! 🎉

packages/html-reporter/src/index.tsx
```diff
@@ -14,14 +14,50 @@
...
+  const field = (name: string) => attachments.find(({ name: n }) => n === name)?.body;
+  const fieldAsJSON = (name: string) => {
+    const raw = field(name);
+    if (raw !== undefined)
+      return JSON.parse(raw);
+  };
+
+  const out = {
+    'generatedAt': fieldAsJSON('generatedAt'),
```
aslushnikov, line 46: Maybe use `fieldAsNumber` and `fieldAsBoolean` to enforce types as well?

packages/html-reporter/src/reportView.css
aslushnikov: TIL!

packages/playwright-test/src/ci.ts
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { spawnAsync } from 'playwright-core/lib/utils/utils';
+
+const GIT_OPERATIONS_TIMEOUT_MS = 1_500;
```
aslushnikov, line 18: Historic context: last time we had issues with number separators since our transpiler didn't touch them, and they made code crash on Node 12 I think. I ended up reverting my usages of them.

Maybe something has changed in our pipelines so it works now?

packages/playwright-test/src/ci.ts
```diff
@@ -0,0 +1,91 @@
...
+    { name: 'revision.id', body: Buffer.from(sha), contentType: kContentTypePlainText },
+    { name: 'revision.author', body: Buffer.from(authorName), contentType: kContentTypePlainText },
+    { name: 'revision.email', body: Buffer.from(authorEmail), contentType: kContentTypePlainText },
+    { name: 'revision.subject', body: Buffer.from(subject), contentType: kContentTypePlainText },
+    { name: 'revision.timestamp', body: Buffer.from(JSON.stringify(timestamp)),  contentType: kContentTypeJSON },
+    { name: 'revision.localPendingChanges', body: Buffer.from(!!status + ''), contentType: kContentTypePlainText },
+  ];
+};
+
+export const linksFromEnv = async (): Promise<Attachment[]> => {
```
aslushnikov, line 60: can we have three methods, one per CI provider? It'll be easy to add more later

packages/playwright-test/types/test.d.ts
```diff
@@ -1268,6 +1268,15 @@ export interface FullConfig<TestArgs = {}, WorkerArgs = {}> {
...
   webServer: WebServerConfig | null;
+  /**
+   * > NOTE: This does not include test-level attachments. See
+   * [testInfo.attach(name[, options])](https://playwright.dev/docs/api/class-testinfo#test-info-attach) and
+   * [testInfo.attachments](https://playwright.dev/docs/api/class-testinfo#test-info-attachments) for working with test-level
+   * attachments.
+   *
+   * The list of files or buffers attached for the overall Playwright Test run. Some reporters show attachments.
+   */
+  attachments: { name: string, path?: string, body?: Buffer, contentType: string }[];
```
aslushnikov, line 1279: fundamental question: we started with `testInfo.attachments` array initially, but IIRC we now consider using `testInfo.attach()` a better practice? Will the same happen here too? Shall we have a method right away?

tests/playwright-test/reporter-html.spec.ts
```diff
@@ -624,3 +627,63 @@ test('open tests from required file', async ({ runInlineTest, showReport, page }
...
+  const beforeRunPlaywrightTest = async ({ baseDir }: { baseDir: string }) => {
+    const execGit = async (args: string[]) => {
+      const { code, stdout, stderr } = await spawnAsync('git', args, { stdio: 'pipe', cwd: baseDir });
+      if (!!code)
+        throw new Error(`Non-zero exit of:\n$ git ${args.join(' ')}\nConsole:\nstdout:\n${stdout}\n\nstderr:\n${stderr}\n\n`);
+      return;
+    };
+
+    await execGit(['init']);
+    await execGit(['config', '--local', 'user.email', 'shakespeare@example.local']);
```
aslushnikov, line 641: : D oh crazy test!

packages/playwright-test/src/reporters/html.ts
```diff
@@ -228,7 +229,7 @@ class HtmlBuilder {
     this._dataZipFile = new yazl.ZipFile();
   }
 
-  async build(rawReports: JsonReport[]): Promise<{ ok: boolean, singleTestId: string | undefined }> {
+  async build(runLevelReport: { attachments: JsonAttachment[] }, rawReports: JsonReport[]): Promise<{ ok: boolean, singleTestId: string | undefined }> {
```
aslushnikov, line 232: curious: why isn't the first argument just a `testReportAttachments: JsonAttachment[]`?

packages/html-reporter/src/index.tsx
```diff
@@ -14,14 +14,50 @@
...
+  'revision.email': string;
+  'revision.subject': string;
+  'revision.timestamp': number;
+  'revision.link': string;
+  'revision.localPendingChanges': boolean;
+  'ci.link': string;
+}>;
+
+const extractMetadata = (attachments: TestAttachment[]): Metadata | undefined => {
+  const field = (name: string) => attachments.find(({ name: n }) => n === name)?.body;
```
aslushnikov, line 38: nit: maybe `findLast` would be more natural vs `find` to prefer last plugin attachment?

aslushnikov (approved):

packages/playwright-test/src/reporters/html.ts
rwoll: I named it poorly! Updated! Thanks!

packages/playwright-test/types/test.d.ts
rwoll: Mayb—but let's see if this gets used! I mentioned adding it yesterday, but we decided to punt on it for now.

rwoll:
Looking into the failed HTML Reporter test; I can repro locally if I replicate the checkout strategy the runners use.

rwoll:
Merged in. Reviewed CQ1 failures manually and all look unrelated; confirmed with @aslushnikov

PR #12744, opened 2022-03-14
feat (junit reporter): add option to force usage of CDATA sections for content in XML elements

### LINKED ISSUE #12741: [Feature] add new option on junit reporter to use CDATA sections to encode content
Provide a new option on the junit reporter to control whether to use CDATA sections to encode text content in the JUnit XML elements.
This provides a safe mechanism to add content on elements, such as `<system-out>` and other that use inner text, where we don't control what the users provide. Parsers usually skip (i.e., don't pase) the contents of CDATA sections, taking out the start and ending CDATA sections delimiters.

bitcoder:
Provides a new configuration option on the junit reporter to control whether to use CDATA sections on XML elements such as `<system-out>`. CDATA sections are not parsed by XML parsers, and thus represent a safe way to embed user defined content.

- adds the `useCDATAsections` configuration option for the junit reporter (false, by default)
- if true, for example if `Hello world` is written to the STDOUT, it would be persisted on the JUnit report as `<system-out><![CDATA[Hello world]]></system-out>` instead of  `<system-out>Hello world</system-out>`
- the special characters `&"'<>` are always escaped, except if they are part of a CDATA section
- on the CDATA section, all characters are valid, except the sequence `]]>` which is the CDATA section end delimiter; if this exists on the text, then the `>` is escaped to `&gt;` and the sequence becomes  `]]&gt;`

Additional info on the [XML specification](https://www.w3.org/TR/REC-xml/#sec-cdata-sect).

 Fixes #12741 and references #11374


dgozman (commented):
This looks great!

I thought about this more, and discussed internally, and now think we don't have to do an option. Sorry for the troubles, but could you please remove the option now, and leave the implementation and tests?

Sorry for being indecisive!

bitcoder:
> This looks great!
> 
> I thought about this more, and discussed internally, and now think we don't have to do an option. Sorry for the troubles, but could you please remove the option now, and leave the implementation and tests?
> 
> Sorry for being indecisive!

@dgozman updated the PR :)

dgozman (approved):
Looks great! I will be able to merge after you sign the CLA - see the comment from microsoft-cla bot.

raghavthind2005 (approved):

PR #12750, opened 2022-03-14
docs: mention locator.nth is zero based

mxschmitt:
We get the question over and over again.

### DIFF
--- a/docs/src/api/class-framelocator.md
+++ b/docs/src/api/class-framelocator.md
@@ -128,7 +128,7 @@ The method finds an element matching the specified selector in the FrameLocator'
 ## method: FrameLocator.nth
 - returns: <[FrameLocator]>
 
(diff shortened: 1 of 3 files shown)


docs/src/api/class-framelocator.md
```diff
@@ -128,7 +128,7 @@ The method finds an element matching the specified selector in the FrameLocator'
 ## method: FrameLocator.nth
 - returns: <[FrameLocator]>
 
-Returns locator to the n-th matching frame.
+Returns locator to the n-th matching frame. It's zero based and starts by that at `0`.
```
pavelfeldman, line 131:
```suggestion
Returns locator to the n-th matching frame. It's zero based, `nth(0)` selects the first element.
```

pavelfeldman (approved):

PR #12751, opened 2022-03-14
feat(fixtures): per-fixture timeout

dgozman:
By default, fixtures share timeout with the test they are instantiated for. However, for more heavy fixtures, especially worker-scoped ones, it makes sense to have a separate timeout.

This introduces `{ timeout: number }` option to the list of fixture options that opts the fixture into a dedicated timeout rather than sharing it with the test.

### DIFF
--- a/docs/src/test-api/class-test.md
+++ b/docs/src/test-api/class-test.md
@@ -880,7 +880,7 @@ Test function that takes one or two arguments: an object with fixtures and optio
 
 ## method: Test.setTimeout
 
-Changes the timeout for the test. Learn more about [various timeouts](./test-timeouts.md).
+Changes the timeout for the test. Zero means no timeout. Learn more about [various timeouts](./test-timeouts.md).
 
 ```js js-flavor=js
 const { test, expect } = require('@playwright/test');
--- a/packages/playwright-test/src/fixtures.ts
+++ b/packages/playwright-test/src/fixtures.ts
@@ -16,11 +16,13 @@
 
 import { formatLocation, debugTest } from './util';
 import * as crypto from 'crypto';
-import { FixturesWithLocation, Location, WorkerInfo, TestInfo } from './types';
+import { FixturesWithLocation, Location, WorkerInfo } from './types';
 import { ManualPromise } from 'playwright-core/lib/utils/async';
+import { TestInfoImpl } from './testInfo';
+import { FixtureDescription, TimeoutManager } from './timeoutManager';
 
 type FixtureScope = 'test' | 'worker';
-type FixtureOptions = { auto?: boolean, scope?: FixtureScope, option?: boolean };
+type FixtureOptions = { auto?: boolean, scope?: FixtureScope, option?: boolean, timeout?: number | undefined };
 type FixtureTuple = [ value: any, options: FixtureOptions ];
 type FixtureRegistration = {
   location: Location;  // Fixutre registration location.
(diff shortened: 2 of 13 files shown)


docs/src/test-api/class-test.md
```diff
@@ -920,6 +920,42 @@ test.beforeEach(async ({ page }, testInfo) => {
...
 
+Changing timeout for a fixture:
+
+```js js-flavor=js
+const { test: base, expect } = require('@playwright/test');
+
+const test = base.extend({
+  slowFixture: async ({}, use) => {
+    // Set a separate timeout for this fixture.
+    test.setTimeout(60000);
```
rwoll, line 931: Without looking at the docs, it seems like this would control the Test-level and not Fixture-level timeout. I think calling it `test.setFixtureTimeout(…)` would be more clear (or even `use.setFixtureTimeout` so it's not hanging off of `test` and is more coupled with the Fixture API).

packages/playwright-test/src/testInfo.ts
```diff
@@ -177,13 +192,85 @@ export class TestInfoImpl implements TestInfo {
...
+
+  _setTimeout(location: Location, timeout: number) {
+    // NOTE: make sure _updateRunnables and _setTimeout are aligned!
+
+    if (this._fixtureRunnable && this._fixtureRunnable.timeout === undefined) {
+      if (!this.timeout)
+        return; // Zero timeout means some debug mode - do not set a timeout.
+
+      // Update from shared timeout to per-fixture timeout, reset elapsed to zero.
+      this._elapsedTestTime = this._timeoutRunner.elapsed();
```
aslushnikov, line 226: consider fixture during beforeAll

packages/playwright-test/src/testInfo.ts
```diff
@@ -166,7 +164,7 @@ export class TestInfoImpl implements TestInfo {
     const description = modifierArgs[1];
     this.annotations.push({ type, description });
     if (type === 'slow') {
-      this.setTimeout(this.timeout * 3);
+      this._timeoutManager.setDefaultTimeout(this.timeout * 3);
```
aslushnikov, line 167: I'd expect that `test.slow()` inside fixture / hooks affect the context behavior similarly to `test.setTimeout`?

packages/playwright-test/src/timeoutManager.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  setTimeout(location: Location, timeout: number) {
+    const slot = this._currentSlot();
+    if (!slot.timeout)
+      return; // Zero timeout means some debug mode - do not set a timeout.
+
+    // First setTimeout() call in a fixture - move it to a dedicated time slot.
+    if (this._fixture && !this._fixture.slot) {
+      this._currentSlot().elapsed = this._timeoutRunner.elapsed();
```
aslushnikov, line 90: slot.elapsed =

packages/playwright-test/src/fixtures.ts
```diff
@@ -79,32 +82,41 @@ class Fixture {
     };
     const workerInfo: WorkerInfo = { config: testInfo.config, parallelIndex: testInfo.parallelIndex, workerIndex: testInfo.workerIndex, project: testInfo.project };
     const info = this.registration.scope === 'worker' ? workerInfo : testInfo;
+    this._runnableDescription = {
```
aslushnikov, line 85: move to constructor

aslushnikov (approved):
🌈🥑

healthy patch!!

aslushnikov (approved):

packages/playwright-test/src/fixtures.ts
```diff
@@ -131,12 +147,13 @@ export class FixturePool {
       for (const entry of Object.entries(fixtures)) {
         const name = entry[0];
         let value = entry[1];
-        let options: Required<FixtureOptions> | undefined;
+        let options: { auto: boolean, scope: FixtureScope, option: boolean, timeout: number | undefined } | undefined;
```
mxschmitt, line 150: what was the reason that this got changed? I thought TypeScript-wise this is the same.

dgozman: Nope, it insists on `timeout` being a number, not undefined.

PR #12755, opened 2022-03-14
fix(test-runner): resolve outputDir wrt rootDir

aslushnikov:
Currently, running `npx playwright test` in a subfolder of your project
will result in a `test-results` folder created in your `cwd`.

This is unexpected; instead, we should always resolve all paths
against `rootDir` - directory that contains config.


yury-s (commented):
Should we update how html outputFolder is resolved too?  see https://playwright.slack.com/archives/CSUHZPVLM/p1647040630957209

dgozman (approved):
Let's mention this in "behavior changes" section!

aslushnikov:
@yury-s very fair; I'll follow up with HTML reporter output dir

PR #12759, opened 2022-03-14
feat(vrt): new option `"caret"` for taking screenshots

### LINKED ISSUE #12643: [Feature] Option to not automatically hide caret in screenshots
Version 1.19 contained the great contribution to automatically hide blinking carets in screenshots (https://github.com/microsoft/playwright/pull/11854). Had a similar workaround with overriding styles for the `caret-color` in a project before, the generic integration avoids to manually disable it for visual tests and simplifies the tests a lot.

However I'd still like to request an option to *not* automatically hide blinking carets via CSS, ideally via a [TestOption](https://playwright.dev/docs/api/class-testoptions). The reason for this is that in particular for Firefox, a more powerful option to get similar results exists:

```typescript
  use: {
    launchOptions: {
      firefoxUserPrefs: { 'ui.caretBlinkTime': 0 },
    },
  },
};
```

Instead of hiding the caret, this just disables the blink, thus also guaranteeing a stable screenshot, but with the added value that the caret is still visible. This allows to verify caret styling in visual tests, or check its position, etc., at least in Firefox. Thus, if an option to avoid insertion of the CSS would be available, one could setup projects where it remains visible but doesn't blink in Firefox, but is removed in other browsers.

aslushnikov:
This has two values:
- `"hide"` to hide input caret for taking screenshot
- `"blink"` to keep caret blinking

This defaults to `"hide"`.

Fixes #12643


pavelfeldman:
side note: I don't think anyone would want blinking

aslushnikov:
Closing this since the API doesn't make much sense. Ideally we'd like to have `caret: "show"` or `caret: "hide"` 

Chromium code pointers: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/layout/layout_theme.cc;l=531;drc=6b1959a71454901a7e158820c08bc1e3535956e9;bpv=1;bpt=1

PR #12761, opened 2022-03-15
fix(test-runner): use `configDir` for base in `screenshotsDir` and `outputDir`

aslushnikov:
Turns out `fullConfig.rootDir` is a `testDir` in fact; many places
in our source didn't expect this. Instead, what we want is to have
a `configDir`.

This patch:
- introduces `fullConfig.configDir` that points to the `configDir`
- starts using `configDir` as a default base for `outputDir` and
  `screenshotsDir`


aslushnikov:
closing in favor of https://github.com/microsoft/playwright/pull/12798

PR #12800, opened 2022-03-15
docs: drop release notes and refer to GitHub

mxschmitt:
Closes https://github.com/microsoft/playwright.dev/pull/517


pavelfeldman:
I'm looking at https://code.visualstudio.com/updates/v1_65 + https://github.com/microsoft/vscode/releases and I really like it, should we do the other way around?

PR #12806, opened 2022-03-16
feat: detect docker version and Playwright version mismatch

aslushnikov:
This patch prints a friendly instructions in case Docker image version
mismatches Playwright version and there are missing browser
dependencies.

With this patch, Playwright will yield the following error:

```
root@f0774d2b2097:~# node a.mjs
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

browserType.launch:
╔════════════════════════════════════════════════════════════════════════════════════════════╗
║ Host system is missing dependencies to run browsers.                                       ║
║ This is most likely due to docker image version not matching Playwright version:           ║
║ - Playwright: 1.22.0                                                                       ║
║ -     Docker: 1.21.0                                                                       ║
║                                                                                            ║
║ Either:                                                                                    ║
║ - (recommended) use docker image "mcr.microsoft.com/playwright:v1.22.0-focal"              ║
║ - (alternative 1) run the following command inside docker to install missing dependencies: ║
║                                                                                            ║
║     npx playwright install-deps                                                            ║
║                                                                                            ║
║ - (alternative 2) use Aptitude inside docker:                                              ║
║                                                                                            ║
║     apt-get install libgbm1                                                                ║
║                                                                                            ║
║ <3 Playwright Team                                                                         ║
╚════════════════════════════════════════════════════════════════════════════════════════════╝
    at file:///root/a.mjs:3:10 {
  name: 'Error'
}```

References #12796

### DIFF
--- a/packages/playwright-core/src/utils/dependencies.ts
+++ b/packages/playwright-core/src/utils/dependencies.ts
@@ -207,7 +207,7 @@ export async function validateDependenciesLinux(sdkLanguage: string, linuxLddDir
   ];
   // Ignore patch versions when comparing docker container version and Playwright version:
   // we **NEVER** roll browsers in patch releases, so native dependencies do not change.
-  if (dockerInfo && !dockerInfo.driverVersion.startsWith(utils.getPlaywrightVersion(true /* majorMinorOnly */))) {
+  if (dockerInfo && !dockerInfo.driverVersion.startsWith(utils.getPlaywrightVersion(true /* majorMinorOnly */) + '.')) {
     // We are running in a docker container with unmatching version.
     // In this case, we know how to install dependencies in it.
     const pwVersion = utils.getPlaywrightVersion();
--- a/packages/playwright-core/src/utils/dependencies.ts
+++ b/packages/playwright-core/src/utils/dependencies.ts
@@ -24,6 +24,24 @@ import { deps } from './nativeDeps';
 import { getUbuntuVersion } from './ubuntuVersion';
 
 const BIN_DIRECTORY = path.join(__dirname, '..', '..', 'bin');
+const packageJSON = require('../../package.json');
+
+const dockerVersionFilePath = '/ms-playwright/.docker-info';
+export async function writeDockerVersion(dockerImageNameTemplate: string) {
+  await fs.promises.mkdir(path.dirname(dockerVersionFilePath), { recursive: true });
+  await fs.promises.writeFile(dockerVersionFilePath, JSON.stringify({
+    driverVersion: packageJSON.version,
+    dockerImageName: dockerImageNameTemplate.replace('%version%', packageJSON.version),
+  }, null, 2), 'utf8');
+  // Make sure version file is globally accessible.
+  await fs.promises.chmod(dockerVersionFilePath, 0o777);
+}
+
+async function readDockerVersion(): Promise<null | { driverVersion: string, dockerImageName: string }> {
+  return await fs.promises.readFile(dockerVersionFilePath, 'utf8')
+      .then(text => JSON.parse(text))
+      .catch(e => null);
+}
 
 const checkExecutable = (filePath: string) => fs.promises.access(filePath, fs.constants.X_OK).then(() => true).catch(e => false);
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: utils/docker/build.sh [hunk 1]
```diff
@@ -37,4 +37,8 @@ else
   exit 1
 fi
 
-docker build --platform "${PLATFORM}" -t "$3" -f "Dockerfile.$2" .
+PW_VERSION=$(node ../../utils/workspace.js --get-version)
```
mxschmitt, line 40: we do the same also here, lets replace it also with --get-version? https://github.com/microsoft/playwright/blob/76c37f1977bfdb31165e09db61b4c43d95a9ac7c/utils/docker/publish_docker.sh#L10

packages/playwright-core/src/utils/dependencies.ts
```diff
@@ -179,6 +179,28 @@ export async function validateDependenciesLinux(sdkLanguage: string, linuxLddDir
...
   const maybeSudo = (process.getuid() !== 0) && os.platform() !== 'win32' ? 'sudo ' : '';
+  // Check if we are running in a docker container which version does not match.
+  // Ignore patch versions: we never roll browsers in patch releases.
+  if (process.env.__PLAYWRIGHT_DOCKER_DISPLAY_NAME && process.env.__PLAYWRIGHT_DOCKER_VERSION && !process.env.__PLAYWRIGHT_DOCKER_VERSION.startsWith(utils.getPlaywrightVersion(true /* majorMinorOnly */))) {
+    const pwVersion = utils.getPlaywrightVersion();
+    const requiredDockerImage = process.env.__PLAYWRIGHT_DOCKER_DISPLAY_NAME.replace(process.env.__PLAYWRIGHT_DOCKER_VERSION, pwVersion);
+    throw new Error('\n' + utils.wrapInASCIIBox([
+      `Host system is missing a few dependencies to run browsers.`,
+      ``,
+      `This is due to docker image version not matching Playwright version:`,
```
dgozman, line 190: "most likely"

packages/playwright-core/src/utils/dependencies.ts
```diff
@@ -179,6 +179,28 @@ export async function validateDependenciesLinux(sdkLanguage: string, linuxLddDir
...
+      `- Playwright: ${pwVersion}`,
+      `-     Docker: ${process.env.__PLAYWRIGHT_DOCKER_VERSION}`,
+      ``,
+      `Either:`,
+      `- (recommended) use docker image "${requiredDockerImage}"`,
+      `- (alternative) run the following command inside docker to install missing dependencies:`,
+      ``,
+      `    ${maybeSudo}${buildPlaywrightCLICommand(sdkLanguage, 'install-deps')}`,
```
dgozman, line 198: I see "Unhappy path" below, so this command is not always the right one? Let's be careful.

dgozman (commented):
Nice!

mxschmitt:
Instead of making the docker build process more complicated, we could think about a hidden command in our CLI: "mark-pw-version-to-os" which will store the current version on the file system, e.g. `ms-playwright` folder. 

On the languages. like .NET or Java etc. its often quite complicated to get the version, you need to read XML. And in Python there is no version in the actual code.

This would enable us to just add `npx playwright mark-pw-version-to-os` in our docker images and we are golden!

packages/playwright-core/src/utils/dependencies.ts
```diff
@@ -183,11 +201,36 @@ export async function validateDependenciesLinux(sdkLanguage: string, linuxLddDir
...
   const maybeSudo = (process.getuid() !== 0) && os.platform() !== 'win32' ? 'sudo ' : '';
+  const dockerInfo = await readDockerVersion();
   const errorLines = [
     `Host system is missing dependencies to run browsers.`,
   ];
-  if (missingPackages.size && !missingDeps.size) {
-    // Happy path: known dependencies are missing for browsers.
+  // Ignore patch versions when comparing docker container version and Playwright version:
+  // we **NEVER** roll browsers in patch releases, so native dependencies do not change.
+  if (dockerInfo && !dockerInfo.driverVersion.startsWith(utils.getPlaywrightVersion(true /* majorMinorOnly */))) {
```
dgozman, line 210:
```suggestion
  if (dockerInfo && !dockerInfo.driverVersion.startsWith(utils.getPlaywrightVersion(true /* majorMinorOnly */) + '.')) {
```

To avoid 1.200 vs 1.20 mismatch 😄

dgozman (approved):

Issue #12814, opened 2022-03-16
[Question] how to use expect.any in playwright


FrankWang117:
I want to :
```typescript
expect(id).toEqual(expect.any(String))
```

aslushnikov:
You can do 

```js
expect(typeof id).toBe('string')
```

pavelfeldman:
This looks like a regression from https://github.com/microsoft/playwright/pull/12403, does it?

FrankWang117:
@aslushnikov maybe.
but if I want to compare object:
```typescript
expect(request).toEqual({
    foo: 1,
    id: expect.any(String)
})
```
how to do?
Now I use:
```typescript
 expect(request).toMatchObject({
    foo: 1
 });
expect(typeof request.id).toBe('string');
```

pavelfeldman:
```
import { test, expect } from '@playwright/test';

test('foo', async () => {
  expect('id').toEqual(expect.any(String));
});
```

works for me, it is just that the types aren't right, we'll fix it right away.

FrankWang117:
@pavelfeldman ok,thanks a lot.

pavelfeldman:
Fixed, awaiting the 1.20.1 push.

PR #12820, opened 2022-03-16
chore: restore expect.any()/expect.anything()

mxschmitt:
This got previously removed in https://github.com/microsoft/playwright/pull/12403, this patch restores it (see [here](https://github.com/facebook/jest/blob/main/packages/expect/build/asymmetricMatchers.d.ts) for all) and adds a tsc and real test (previously it was only tsc) to it.

Relates https://github.com/microsoft/playwright/issues/12814

### DIFF
--- a/tests/playwright-test/playwright-test-fixtures.ts
+++ b/tests/playwright-test/playwright-test-fixtures.ts
@@ -241,7 +241,8 @@ const TSCONFIG = {
     'esModuleInterop': true,
     'allowSyntheticDefaultImports': true,
     'rootDir': '.',
-    'lib': ['esnext', 'dom', 'DOM.Iterable']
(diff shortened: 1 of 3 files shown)


tests/playwright-test/playwright-test-fixtures.ts
```diff
@@ -241,7 +241,8 @@ const TSCONFIG = {
     'esModuleInterop': true,
     'allowSyntheticDefaultImports': true,
     'rootDir': '.',
-    'lib': ['esnext', 'dom', 'DOM.Iterable']
+    'lib': ['esnext', 'dom', 'DOM.Iterable'],
+    'noEmit': true,
```
mxschmitt, line 245: this was previously emitting the files on the disk, noEmit prevents that and the TSC validation still happens.

dgozman (approved):

Issue #12836, opened 2022-03-16
[REGRESSION]: Tests not detected as ES module in v1.20


henrikstorck:
**Context:**
- GOOD Playwright Version: 1.19.2
- BAD Playwright Version: 1.20.0
- Operating System: Windows

**Code Snippet**

[https://github.com/henrikstorck/playwright-test](https://github.com/henrikstorck/playwright-test)

**Describe the bug**

When running `npm test` in the above repository, I get the following error message with playwright v1.20:

```
> npm test

Running 0 test using 0 worker


Error [ERR_REQUIRE_ESM]: require() of ES Module D:\Projects\playwright-test\tests\test.js from D:\Projects\playwright-test\node_modules\@playwright\test\lib\loader.js not supported.
Instead change the require of test.js in D:\Projects\playwright-test\node_modules\@playwright\test\lib\loader.js to a dynamic import() which is available in all CommonJS modules.

   at node_modules\@playwright\test\lib\loader.js:244

  242 |
  243 |       if (isModule) return await esmImport();
> 244 |       return require(file);
      |              ^
  245 |     } catch (error) {
  246 |       if (error.code === 'ERR_MODULE_NOT_FOUND' && error.message.includes('Did you mean to import')) {
  247 |         var _$exec;

    at Loader._requireOrImport (D:\Projects\playwright-test\node_modules\@playwright\test\lib\loader.js:244:14)
    at Loader.loadTestFile (D:\Projects\playwright-test\node_modules\@playwright\test\lib\loader.js:142:18)
    at Runner._runFiles (D:\Projects\playwright-test\node_modules\@playwright\test\lib\runner.js:295:44)
    at Runner._run (D:\Projects\playwright-test\node_modules\@playwright\test\lib\runner.js:231:23)
    at async TimeoutRunner.run (D:\Projects\playwright-test\node_modules\playwright-core\lib\utils\async.js:38:14)
    at async raceAgainstTimeout (D:\Projects\playwright-test\node_modules\playwright-core\lib\utils\async.js:85:15)
    at async Runner.runAllTests (D:\Projects\playwright-test\node_modules\@playwright\test\lib\runner.js:188:20)
    at async runTests (D:\Projects\playwright-test\node_modules\@playwright\test\lib\cli.js:184:18)
    at async Command.<anonymous> (D:\Projects\playwright-test\node_modules\@playwright\test\lib\cli.js:79:7)

=================
 no tests found.
=================
```

[...]

PR #12837, opened 2022-03-16
fix: fix folder traversal for isModule check (#12836)

henrikstorck:
This change fixed the error shown in #12836, where tests in an ES module don't get detected as such and therefore break test execution in v1.20 because they are loaded with `require()` instead of `import`.

By using `path.dirname(folder)` instead of `path.basename(folder)`, `folderIsModule` is able to correctly iterate through all parent folders and detect the correct project type.


pavelfeldman (approved):

pavelfeldman:
Awesome, thanks! Let me follow up with a test.

PR #12838, opened 2022-03-16
test: add a test for #12837

pavelfeldman:


tests/playwright-test/loader.spec.ts
```diff
@@ -198,6 +198,26 @@ test('should load esm when package.json has type module', async ({ runInlineTest
   expect(result.passed).toBe(1);
 });
 
+test('should load nested as esm when package.json has type module', async ({ runInlineTest }) => {
```
mxschmitt, line 201:
```suggestion
test('should load nested as esm when package.json has type module @esm', async ({ runInlineTest }) => {
```

so it will also runs on our esm bots (they run on windows and node.js 16)

mxschmitt (approved):

PR #12864, opened 2022-03-17
chore: fix modification dates in driver

mxschmitt:


campersau:
`tar` also supports `--touch` so this
https://github.com/microsoft/playwright/blob/70603d76858d5a62e7bd725309d3aa03bef186c6/utils/build/build-playwright-driver.sh#L33
could be change to:
```bash
tar -xzf ./output/playwright-core.tgz -C ./output/playwright-${SUFFIX}/ --touch
```
instead.

yury-s (approved):

yury-s:
> [`tar`](https://linux.die.net/man/1/tar) also supports `--touch` so this

Seem like it is not supported on macOS :(

aslushnikov:
How come this was working before? Is this a regression? If it is, what triggered it?

campersau:
To support different publish modes in `playwright-dotnet` the files are now included via `msbuild`, which checks for the modified date, instead of manually copying them every time.
See https://github.com/microsoft/playwright-dotnet/pull/1955#issuecomment-1019148707

PR #12866, opened 2022-03-18
chore: disambiguate internal flags

### LINKED ISSUE #12809: [BUG] Infinite loop when calling `storageState` on page close
**Context:**
- Playwright Version: 1.19.2
- Operating System: Linux
- Node.js version: v16.13.0
- Browser: Chromium, Firefox

**Code Snippet**

```javascript

require('http').createServer((request, response) => {
        response.end('');
}).listen(8888);

(async () => {
        const browser = await playwright.chromium.launchPersistentContext('', {
                headless: false,
        });

        browser.on('page', (page) => {
                page.on('close', async () => {
                        const state = await browser.storageState();

                        // console.log(state);
                });
        });

        await page.goto('http://localhost:8888/');
        await page.close();
})();
```

**Describe the bug**

New pages keep opening and closing indefinitely. That's because of the current implementation: https://github.com/microsoft/playwright/blob/964eeba0dc3934f0d57f7ed0e209992c53ba9027/packages/playwright-core/src/server/browserContext.ts#L349

pavelfeldman:
Fixes #12809

### DIFF
--- a/packages/playwright-core/src/server/browser.ts
+++ b/packages/playwright-core/src/server/browser.ts
@@ -103,10 +103,10 @@ export abstract class Browser extends SdkObject {
   _videoStarted(context: BrowserContext, videoId: string, path: string, pageOrError: Promise<Page | Error>) {
     const artifact = new Artifact(context, path);
     this._idToVideo.set(videoId, { context, artifact });
-    context.emit(BrowserContext.Events.VideoStarted, artifact);
     pageOrError.then(page => {
       if (page instanceof Page) {
         page._video = artifact;
+        page.emitOnContext(BrowserContext.Events.VideoStarted, artifact);
         page.emit(Page.Events.Video, artifact);
       }
     });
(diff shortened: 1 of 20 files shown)


packages/playwright-core/src/server/browser.ts
```diff
@@ -103,10 +103,10 @@ export abstract class Browser extends SdkObject {
   _videoStarted(context: BrowserContext, videoId: string, path: string, pageOrError: Promise<Page | Error>) {
     const artifact = new Artifact(context, path);
     this._idToVideo.set(videoId, { context, artifact });
-    context.emit(BrowserContext.Events.VideoStarted, artifact);
```
dgozman, line 106: We probably have a test that ensures all videos are reported, even for those popups that were closed very fast.

dgozman (approved):

packages/playwright-core/src/server/browser.ts
pavelfeldman: I don't think we do! But this makes me think I should try skipping video recording there!

PR #12873, opened 2022-03-18
devops: apply patch failsafe

dpino:
When running prepare_checkout.sh on a newer BASE_REVISION "git apply"
fails if there are conflicts. This use case is useful when rebasing
bootstrap.patch to a new commit.

Now when there is an environment variable APPLY_PATCH_FAILSAFE which
when defined makes "git apply" not to fail if there are conflicts.
In addition, conflicts are stored in a file called 'conflicts' in CHECKOUT_PATH.

The patch also allows to pass a second argument to
./prepare_checkout.sh, which will overwrite the commit id value of
BASE_REVISION in UPSTREAM_CONFIG.sh.

This feature is only support for WebKit browser.

### DIFF
--- a/browser_patches/prepare_checkout.sh
+++ b/browser_patches/prepare_checkout.sh
@@ -27,6 +27,35 @@ if [[ $# == 0 ]]; then
   exit 1
 fi
 
+update_base_revision() {
+    local browser="$1"
+    local base_revision="$2"
+
+    sed -i -r "s|^BASE_REVISION=\"[^\"]+\"$|BASE_REVISION=\"$base_revision\"|" "./$browser/UPSTREAM_CONFIG.sh"
+}
+
+apply_patch_failsafe() {
+    set +e;
+    git apply --3way --whitespace=nowarn "$PATCHES_PATH"/*
+    if [[ "$?" -ne 0 ]]; then
+        echo -e "\033[31mWARNING\033[m: Patch apply failed!"
+        dump_conflicts
+    fi
+    set -e;
+}
+
+dump_conflicts() {
+    local conflicts="$CHECKOUT_PATH/conflicts"
+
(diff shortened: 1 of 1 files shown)


aslushnikov:
@dpino This patch suggests browser rolling to be driven by `./prepare_checkout.sh wk <NEW_BASE_SHA>` and adds some code to support it. While the code itself seems to be fine, I'd like to better understand the motivation.

---

As of today, we use the following procedure to roll browsers that worked for us. Say, for example, you have Playwright checkout at `~/playwright` and WebKit checkout at `~/webkit`. 

Then to roll WebKit on new tip-of-tree, we'd do the following:

1. prepare current playwright build checkout:
    ```bash
    ~/playwright(git: main)$ ./browser_patches/prepare_checkout.sh wk
    ```
2. go to webkit checkout and fetch from `browser_upstream`:
    ```bash
    ~/playwright(git: main)$ cd ~/webkit
    ~/webkit(git: playwright-build)$ git fetch browser_upstream
    ```
3. rebaseline `playwright-build` branch atop of `browser_upstream/main`:
    ```bash
    ~/webkit(git: playwright-build)$ git rebase browser_upstream/main
    ```
    Resolve all the conflicts e.t.c.
4. finally, re-export a new `playwright-build` branch:
    ```bash
    ~/webkit(git: playwright-build)$ cd ~/playwright
    ~/playwright(git: main)$ ./browser_patches/export.sh wk
    ```

Would this workflow work for you? If not, why?

dpino:
@aslushnikov Sure, the workflow you described works for me, so there's no need to actually modify `prepare_checkout.sh`. To give some background about the proposed changes, when I started doing the rolls I added these tweaks to `prepare_checkout.sh` which I have been using so far. I thought it would be handy to have these changes upstream.

PR #12874, opened 2022-03-18
browser(webkit): skip generate-bundle step if Minibrowser not defined

dpino:
Script 'browser_patches/webkit/build.sh' allows to build WebKitGTK only,
WPE only or both browsers. In case of building only one port, the
'archive.sh' script should skip the generate-bundle step for the port not
built.


yury-s (approved):
lgtm, but this script doesn't affect build.sh as we don't perform archive locally. As far as I can tell it is only called from checkout_build_archive_upload.sh which is only used on the bots.

raghavthind2005 (approved):

PR #12880, opened 2022-03-18
fix(codegen): do not consider empty attributes for selector generation

brhoten:
fixes codegen preferring attributes even if they are empty.


yury-s (commented):
The change looks good, is there a bug describing an example that was affected by the previous behavior?

brhoten:
I did not see one. I encountered the issue while trying to automate an OAuth2 log in with a provider that uses empty aria-labels on their input fields.

yury-s (approved):

yury-s:
> Inspector would create an action based on the empty aria-label which would later cause an error when trying to run the same test.

Do you know why it caused an error on second run, is it because the aria-label was populated with some other value or there were multiple elements with such attribute or something else?

brhoten:
Oh, it wasn't a second run, it was the first run of the test that was recorded.

dgozman (approved):
Great PR! 🚀

PR #12882, opened 2022-03-18
docs: automatically test examples

rwoll:
This change intends to help us make our docs and examples:

1. more `@playwright/test` centric since many users consume the APIs via
   test runner
2. self-contained: many of our docs show example API usage, but don't
   show the corresponding DOM that the API is being exercised against
3. working out of the box: there is nothing more frustrating than trying
   an example only to get an error

This introduces:

```
$ npm run dtest
```

which will run all (annotated) code samples from the docs.

See the diffs of the `*.md` files for examples.

The current annotation/generation assumes the example is completely
self-contained, but since importing `test`/`expect` into each example
can get verbose (for both doc contributors and users/readers), we might
consider having that part implied and just including it as part of the
generation. Currently, runnable blocks are annotated with `RUNNABLE`,
so maybe we could introduce variations of this that indicate to the
generation step whether the example is self-container or requires
auto-adding of the common imports.

### DIFF
--- a/docs/src/api/class-download.md
+++ b/docs/src/api/class-download.md
@@ -10,7 +10,7 @@ Download event is emitted once the download starts. Download path becomes availa
 ```js js-flavor=js RUNNABLE
 const { test, expect } = require('@playwright/test');
 
-test('downlad example', async ({ page }) => {
+test('download example', async ({ page }) => {
   await page.setContent(`<a href='data:application/json;base64,ImV4YW1wbGUgZGF0YSIK' download>Download file</a>`);
   // Note that Promise.all prevents a race condition
   // between clicking and waiting for the download.
@@ -28,7 +28,7 @@ test('downlad example', async ({ page }) => {
 ```js js-flavor=ts RUNNABLE
 import { test, expect } from '@playwright/test';
 
-test('downlad example', async ({ page }) => {
+test('download example', async ({ page }) => {
   await page.setContent(`<a href='data:application/json;base64,ImV4YW1wbGUgZGF0YSIK' download>Download file</a>`);
   // Note that Promise.all prevents a race condition
   // between clicking and waiting for the download.
--- a/docs/src/api/class-download.md
+++ b/docs/src/api/class-download.md
@@ -7,17 +7,40 @@ browser context is closed.
 
 Download event is emitted once the download starts. Download path becomes available once download completes:
 
-```js
-// Note that Promise.all prevents a race condition
-// between clicking and waiting for the download.
-const [ download ] = await Promise.all([
-  // It is important to call waitForEvent before click to set up waiting.
-  page.waitForEvent('download'),
-  // Triggers the download.
-  page.locator('text=Download file').click(),
-]);
-// wait for download to complete
-const path = await download.path();
+```js js-flavor=js RUNNABLE
+const { test, expect } = require('@playwright/test');
+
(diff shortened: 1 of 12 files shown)


docs/src/api/class-download.md
```diff
@@ -7,17 +7,40 @@ browser context is closed.
...
-  page.waitForEvent('download'),
-  // Triggers the download.
-  page.locator('text=Download file').click(),
-]);
-// wait for download to complete
-const path = await download.path();
+```js js-flavor=js RUNNABLE
+const { test, expect } = require('@playwright/test');
+
+test('downlad example', async ({ page }) => {
```
yury-s, line 13: typo: download

docs/src/api/class-download.md
```diff
@@ -7,17 +7,40 @@ browser context is closed.
...
+  ]);
+  // wait for download to complete
+  await expect(download.path()).resolves.toBeTruthy();
+});
+```
+
+```js js-flavor=ts RUNNABLE
+import { test, expect } from '@playwright/test';
+
+test('downlad example', async ({ page }) => {
```
yury-s, line 31: ditto

docs/src/api/class-download.md
```diff
@@ -7,17 +7,40 @@ browser context is closed.
...
-  // Triggers the download.
-  page.locator('text=Download file').click(),
-]);
-// wait for download to complete
-const path = await download.path();
+```js js-flavor=js RUNNABLE
+const { test, expect } = require('@playwright/test');
+
+test('download example', async ({ page }) => {
+  await page.setContent(`<a href='data:application/json;base64,ImV4YW1wbGUgZGF0YSIK' download>Download file</a>`);
```
rwoll, line 14: This set content snippet is distracting; point at live (stable) demo site.

docs/src/input.md
```diff
@@ -9,21 +9,50 @@ title: "Input"
...
-```js
-// Text input
-await page.fill('#name', 'Peter');
+```js js-flavor=ts RUNNABLE
+import { test, expect } from '@playwright/test';
 
-// Date input
-await page.fill('#date', '2020-02-02');
+test('form examples', async ({ page }) => {
+  await page.goto('https://rwoll.github.io/gh-pages-examples/');
```
rwoll, line 16: All these examples will point at official domain; this site is currently a placeholder.

docs/src/input.md
```diff
@@ -9,21 +9,50 @@ title: "Input"
...
 
-// Time input
-await page.fill('#time', '13:15');
+  // Text input
+  await page.fill('#input-text', 'Peter');
```
rwoll, line 19: Let's amend the example with more natural ids, like `#author-name`, `#input-text` could be confusing since it coincidentally contains the name of an HTML Element.

rwoll (commented):
Pushed an update with more examples for the Input Guide in our docs. This points at a live site so we don't have distracting bits in our examples like I had before; however, in the future it would be nice to see the HTML inline somewhere in the docs so the selectors make more sense without visiting the external site.

rwoll:
Closing—it's a nice idea, but at the moment I don't see us converting all our docs to this style. Furthermore, after converting one of the guides to this style, I think as-is this proposal would add noise to the docs for users (since there's extra code that would accompany each example). This is fixable (e.g. we could skip rendering some bits), but this requires some additional thought.

Issue #12835, opened 2022-03-16
[BUG] `npx playwright install-deps` and `npx playwright install` does not install necessary dependencies for 1.19.2 after 1.20 released


unlikelyzero:
I debated whether I should file this as a bug or question, so I'll include both statements :)

[Bug] Is it possible for the  `npx playwright install` to be backwards compatible with N-1 installs to reduce the overall problem of dependency management while teams "vet" each new release?

[Question] is there a way we could reduce our playwright dependency bumps to a single referenced of a given playwright version? I think we're looking for guidance on how to better manage the dependency changes across our different CI systems and projects.

**Context:**
- Playwright Version: 1.19.2 _after_ 1.20 was released
- Operating System: ubuntu/mac11.6/windows
- Node.js version: 12/14/16 @lts
- Browser: Chromium
- Extra: CircleCI and Github Actions

**Describe the bug**

When playwright/test@1.20 was released, we were still on 1.19.2. It is no longer possible to perform a `npx playwright install` or `npx playwright install-deps` on `ubuntu-latest` or `windows-latest`.

We have been riding N-1 for some time now just due to rapid pace of development on playwright and will likely never run in an unpinned, bleeding edge version.

Because we want to use windows and mac OS installations for running playwright tests on github actions, we are using `npx playwright install-deps` on `macos-latest` and `windows-latest` instead of the base `:focal` images. We **do** use the versioned `:focal` images in Circle CI.

Keeping our systems up to date with playwright/test changes now involve the following:
**Open Source**
1. Update playwright/test dependency in package.json
2. Update playwright base image in circle ci
**Downstream Closed Source**
1. Update playwright/test dependency in package.json
2. Update playwright base image in CI

**Now we will need to add:**
3. Update `npx playwright install` in 2/3 different GHA Files.

pavelfeldman:
Interesting, I already answered this one, but I don't see my reply.

> When playwright/test@1.20 was released, we were still on 1.19.2. It is no longer possible to perform a npx playwright install or npx playwright install-deps on ubuntu-latest or windows-latest.

Could you elaborate on what is no longer possible?

> We have been riding N-1 for some time now just due to rapid pace of development on playwright and will likely never run in an unpinned, bleeding edge version.

This sounds reasonable.

> Because we want to use windows and mac OS installations for running playwright tests on github actions, we are using npx playwright install-deps on macos-latest and windows-latest instead of the base :focal images. We do use the versioned :focal images in Circle CI.

[...]

PR #12893, opened 2022-03-18
[CI] Test examples folder

unlikelyzero:
This is a simple GHA to test the existing examples folder to try to repro https://github.com/microsoft/playwright/issues/12835

By doing this, I've been unable to reproduce the issue we saw, but hope that testing the examples might prove beneficial to the project longterm!

### DIFF
--- a/.github/workflows/n-minus-one-install.yml
+++ b/.github/workflows/n-minus-one-install.yml
@@ -5,6 +5,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        key: ${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('package.json') }}
     - uses: actions/setup-node@v2
       with:
         node-version: 16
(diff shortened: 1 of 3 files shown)


.github/workflows/n-minus-one-install.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  test_n_minus_one: #test that an older, pinned playwright/test dependency is able to use npx playwright install
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '16'
+    - run: cd examples/todomvc; npm install
+    - run: cd examples/todomvc; npx playwright install
```
pavelfeldman, line 17: You are missing the `npx playwright install-deps`, so things are not supposed to work.

ankur-cvinfotech: i am facing same issue and does not how to resolve it my docker container is keep failing due to this error i have done all the step still non of them worked.

please let know if there is any specific solutions to solve this.

PR #12909, opened 2022-03-20
chore: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.


pavelfeldman:
I don't think `substr` is going anywhere, any time. Calling it deprecated sounds like an uphill battle.

rwoll:
Per the above (https://github.com/microsoft/playwright/pull/12909#issuecomment-1074725148), closing as part of triage!

Thanks for your time!

PR #12917, opened 2022-03-21
docs(docker.md): correct seccomp_profile.json

weeix:
seccomp_profile.json should be a JSON object. A JSON array would produce this error message:

```
docker: Error response from daemon: Decoding seccomp profile failed: json: cannot unmarshal array into Go value of type seccomp.Seccomp.
```

CR: https://codereviewvideos.com/blog/how-i-fixed-docker-error-response-from-daemon-decoding-seccomp-profile-failed-json-cannot-unmarshal-array-into-go-value-of-type-seccomp-seccomp/


aslushnikov:
@weeix the actual seccomp profile to be used is https://github.com/microsoft/playwright/blob/main/utils/docker/seccomp_profile.json

The JSON is just an extra syscall that added to the profile. I see there's confusion though; I'd remove the whole snippet altogether.

weeix:
Thank you for the clarification. I'll close this PR.

PR #12933, opened 2022-03-21
fix(selenium): Use ms:edgeOptions with Edge

bwalderman:
Updating Selenium Grid support to use the `ms:edgeOptions` capability when connecting to Microsoft Edge. Edge WebDriver doesn't recognize `goog:chromeOptions`.


dgozman (approved):
Good catch! Thank you for fixing 😄

PR #12942, opened 2022-03-22
chore: use `package.json` directory to resolve output directory

aslushnikov:
This patch:
- starts using directory of `package.json` to resolve default
  output directory path
- starts using either `package.json` directory or configuration
  directory to resolve all relative paths

References #12970

### DIFF
--- a/packages/playwright-test/src/loader.ts
+++ b/packages/playwright-test/src/loader.ts
@@ -37,7 +37,7 @@ export class Loader {
   private _defaultConfig: Config;
   private _configOverrides: Config;
   private _fullConfig: FullConfig;
-  private _config: Config = {};
+  private _configDir: string = '';
   private _configFile: string | undefined;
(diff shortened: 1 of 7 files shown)


packages/playwright-test/src/loader.ts
```diff
@@ -37,7 +37,8 @@ export class Loader {
   private _defaultConfig: Config;
   private _configOverrides: Config;
   private _fullConfig: FullConfig;
-  private _config: Config = {};
+  private _configDir: string = '';
+  private _packageJsonDir: string | undefined;
```
yury-s, line 41: drop it as it's unused

dgozman (commented):
- [x] As discussed, let's undo the rootDir change.
- [x] Change docs to mention this.
- [x] Let's add a test for the new behavior.
- [ ] Might as well put html report change here.

dgozman (approved):

Issue #12929, opened 2022-03-21
[Feature] improve or clarify cross origin request mocking


pngwn:
**Problem**

It is common to have an app and a server running on different origins. Users may wish to mock such network requests (using `page.route` and `page.fulfill`) with something like the following:

```ts
test("basic test", async ({ page }) => {
  await page.route("http://localhost:1234/config", (route) =>
    route.fulfill({
      path: "./test.json",
    })
  );

  await page.goto("http://localhost:3000");

  // assertions, etc.
});
```

Although it looks okay at a glance, it will fail because access control headers have not been set. Adding `"Access-Control-Allow-Origin": "*"` to the headers option object fixes this.

**Possible Solutions**

- **Document it and do nothing else.** At the very least adding an entry to the documentation would be good, it is one of those issues that is obvious when you realise it but it can take some time for your brain to get there. This exacerbated by the fact that the request just hangs rather than erroring in any helpful way. _Is there a way to improve the error (or generate one at all)?_
- **Enable CORS by default.** Bit of an assumption, but most people probably want their mock requests to go through. Could be disabled via an option.
- **Enable CORS via an option.** A `cors: true` option passed to `route.fulfill`

My _personal_ opinion is that Enabling CORS by default for all mocked requests and allowing users to disable this behaviour by passing `cors: false` would be the best solution that caters to the most common case. I'm not sure I can come up with a common reason you would want a mock request to fail due to CORS issues. The only reason would be if you were specifically testing CORS, which seems like a very niche case, if it exists at all. 

The feature-y solutions, have the added benefit of reducing test boilerplate, as every mocked request will need the headers setting.

**Would this be helpful for others?**

I believe so.

PR #12943, opened 2022-03-22
feat(route): add cors header in route.fulfill

yury-s:
#12929

### DIFF
--- a/docs/src/api/class-route.md
+++ b/docs/src/api/class-route.md
@@ -227,11 +227,12 @@ is resolved relative to the current working directory.
 [APIResponse] to fulfill route's request with. Individual fields of the response (such as headers) can be overridden using fulfill options.
 
 ### option: Route.fulfill.cors
-- `cors` <[boolean]>
+- `cors` <[CorsMode]<"allow"|"none">>
 
 By default fulfilled response will have ["Access-Control-Allow-Origin"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin)
-header set to either request's origin (if specified) or '*'. If the is option is set to fals then the [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)
-headers won't be added. Note that all CORS headers configured via `headers` option will take precedence.
+header set to either request's origin (if specified) or '*'. If the is option is set to "none" then the
+[CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) headers won't be added. Note that all CORS
+headers configured via `headers` option will take precedence.
 
 ## method: Route.request
 - returns: <[Request]>
--- a/docs/src/api/class-route.md
+++ b/docs/src/api/class-route.md
@@ -229,11 +229,11 @@ is resolved relative to the current working directory.
 ### option: Route.fulfill.cors
 - `cors` <[CorsMode]<"allow"|"none">>
 
-If the otions is not specifed or is "allow" the fulfilled response will have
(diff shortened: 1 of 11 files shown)


packages/playwright-core/src/server/network.ts
```diff
@@ -265,9 +265,18 @@ export class Route extends SdkObject {
         isBase64 = false;
       }
     }
+    const headers = [...(overrides.headers || [])];
+    if (overrides.cors !== false) {
+      const corsHeader = headers.find(({ name, value }) => name === 'access-control-allow-origin');
+      // See https://github.com/microsoft/playwright/issues/12929
+      if (!corsHeader) {
+        const origin = this._request.headerValue('origin') || '*';
+        headers.push({ name: 'access-control-allow-origin', value: origin });
```
dgozman, line 274: MDN suggests to also send `Vary: Origin`: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#access-control-allow-origin.

packages/playwright-core/src/server/network.ts
```diff
@@ -265,9 +265,18 @@ export class Route extends SdkObject {
         isBase64 = false;
       }
     }
+    const headers = [...(overrides.headers || [])];
+    if (overrides.cors !== false) {
```
dgozman, line 269: I guess we never arrive here for preflights? If we do, we might need some special handling.

packages/playwright-core/src/server/network.ts
```diff
@@ -265,9 +265,18 @@ export class Route extends SdkObject {
...
+      const corsHeader = headers.find(({ name, value }) => name === 'access-control-allow-origin');
+      // See https://github.com/microsoft/playwright/issues/12929
+      if (!corsHeader) {
+        const origin = this._request.headerValue('origin') || '*';
+        headers.push({ name: 'access-control-allow-origin', value: origin });
+      }
```
dgozman, line 275: Should we also set `Access-Control-Allow-Credentials: true`?

docs/src/api/class-route.md
```diff
@@ -226,6 +226,13 @@ is resolved relative to the current working directory.
 
 [APIResponse] to fulfill route's request with. Individual fields of the response (such as headers) can be overridden using fulfill options.
 
+### option: Route.fulfill.cors
+- `cors` <[boolean]>
```
dgozman, line 230: WDYT about `cors: 'allow' | 'deny' | 'none'`?

Also, here is an example of configuring cors on the server side (similar to our use case): https://expressjs.com/en/resources/middleware/cors.html

packages/playwright-core/src/server/network.ts
yury-s: Added a test, we don't currently intercept preflights in any browser.

packages/playwright-core/src/server/network.ts
```diff
@@ -264,9 +265,24 @@ export class Route extends SdkObject {
...
+      const corsHeader = headers.find(({ name, value }) => name === 'access-control-allow-origin');
+      // See https://github.com/microsoft/playwright/issues/12929
+      if (!corsHeader) {
+        const origin = this._request.headerValue('origin');
+        if (origin) {
+          headers.push({ name: 'access-control-allow-origin', value: origin });
+          headers.push({ name: 'access-control-allow-credentials', value: 'true' });
+          headers.push({ name: 'vary', value: 'Origin' });
+        } else {
+          headers.push({ name: 'access-control-allow-origin', value: '*' });
```
dgozman, line 279: Let's drop this one.

dgozman (approved):

Issue #12145, opened 2022-02-16
Why does `TestInfo.attach` hash the filename?


dgozman:
### Discussed in https://github.com/microsoft/playwright/discussions/12129

<div type='discussions-op-text'>

<sup>Originally posted by **mskelton** February 15, 2022</sup>
When passing a file to [`TestInfo.attach`](https://playwright.dev/docs/api/class-testinfo#test-info-attach), it will copy it to `test-results/attachments` and hash the filename.  The hashed filename is not a great user experience as it looses context of where it came from or what it is.  Could it instead use the provided filename with a postfix?  For example:

```
my-report.json
```

becomes

```
my-report-3eba252.json
```

Or even a full hash.</div>

pavelfeldman:
#### Why was this issue closed?

We are prioritizing the features based on the upvotes, recency and activity in the issue thread. It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it. We are closing issues like this one to keep our bug database maintainable. Please feel free to open a new issue and link this one to it if you think this is a mistake.

PR #12950, opened 2022-03-22
feat(testinfo): enhance attach function

Yexiaoxing:
- Add ability to save body as file, this is useful for reporters like JUnit and other.
- Add filename or name to saved filename (#12145)

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -33,6 +33,52 @@ const config: PlaywrightTestConfig = {
 export default config;
 ```
 
+## property: TestConfig.attachmentConfig
+- type: <[Object]>
+  - `useNameAsPrefix` <[boolean]>: Whether to use the attachment name as prefix for the attachment filename, defaults to false.
+  - `useOriginalFilenameAsPrefix` <[boolean]>: Whether to use original filename as saved filename prefix, defaults to false.
+  - `saveBodyAsFile` <[boolean]>: Whether to save body as file. This is useful for reporters such as junit, defaults to false.
+
+Configuration for the [`method: TestInfo.attach`] behaviour.
+
+By default, attachments are saved to the test result directory with hashed path as filename. `useNameAsPrefix` and `useOriginalFilenameAsPrefix` control whether to append the attachment name or the original filename to the filename.
+
+If `useNameAsPrefix` is true, the filename will be `${attach name}-${hashed path}.${extension}`. If `useOriginalFilenameAsPrefix` is true, the filename will be `${original filename}-${hashed path}.${extension}`.
+
+If `saveBodyAsFile` is set to true, the attachment body will be saved to a file. This is useful for reporters such as junit, and if you need to upload all attachments to another service. Be noted that if you have this enabled, the reporter will no longer report the body, but the saved path instead.
+
+
+```js js-flavor=js
+// playwright.config.js
+// @ts-check
+
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+const config = {
+  attachmentConfig: {
+    useNameAsPrefix: true,
+    useOriginalFilenameAsPrefix: true,
+    saveBodyAsFile: true,
+  },
+};
+
+module.exports = config;
+```
+
+```js js-flavor=ts
+// playwright.config.ts
+import { PlaywrightTestConfig } from '@playwright/test';
+
+const config: PlaywrightTestConfig = {
+  attachmentConfig: {
+    useNameAsPrefix: true,
+    useOriginalFilenameAsPrefix: true,
+    saveBodyAsFile: true,
+  },
+};
+
+export default config;
+```
+
 ## property: TestConfig.expect
 - type: <[Object]>
   - `timeout` <[int]> Default timeout for async expect matchers in milliseconds, defaults to 5000ms.
--- a/docs/src/test-api/class-testproject.md
+++ b/docs/src/test-api/class-testproject.md
@@ -104,6 +104,20 @@ const config: PlaywrightTestConfig = {
 export default config;
 ```
 
+## property: TestProject.attachmentConfig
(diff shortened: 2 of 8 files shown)


rwoll (commented):
Thanks for your contribution! Before I review, can you elaborate more on your use case? What are you trying to do today where the base/default attachment layout is causing issues?

Ideally, we keep the built in attachment API and config as slim as possible (i.e. we do not add more options unless necessary).

If a user or reporter wants the attachments in a specific layout, that can be done via a reporter (much like how the HTML Reporter iterates over the `TestResult['attachments']` and copies over and re-arranges what it needs). By not exposing too many options, we are free to change locations and internal file layout, and the implied contract is: use the `TestResult['attachments']` array to determine where your attachments are on disk and what filename they have (which can be separate from the presentational name that appears alongside).

rwoll:
> Thanks for your contribution! Before I review, can you elaborate more on your use case? What are you trying to do today where the base/default attachment layout is causing issues?
> 
> Ideally, we keep the built in attachment API and config as slim as possible (i.e. we do not add more options unless necessary).
> 
> If a user or reporter wants the attachments in a specific layout, that can be done via a reporter (much like how the HTML Reporter iterates over the `TestResult['attachments']` and copies over and re-arranges what it needs). By not exposing too many options, we are free to change locations and internal file layout, and the implied contract is: use the `TestResult['attachments']` array to determine where your attachments are on disk and what filename they have (which can be separate from the presentational name that appears alongside).

@Yexiaoxing For this, can you file a GitHub issue so we can triage and have a log of the discussion there (and possibly gauge upvotes if required)?

Yexiaoxing:
Hi @rwoll, thanks for taking time to review. Let me explain more on the cases,

1. Add ability to save body as file. Currently some reporters only save attachment already on local filesystem, but not for body only attachment, e.g. JUnit.
2. Add filename or name to saved filename. This is based on a discussion [here](https://github.com/microsoft/playwright/discussions/12129).

WIthout the 2# change, the filename will look like this, which is nothing readable and make no sense to users.
CleanShot 2022-04-28 at 13 24 48@2x

Can this`TestResult['attachments']` be a solution? Maybe or maybe not, as you can see, after testinfo.attach function, the original path name is already lost. The current change is only to change the attachment destination, which is the final location of the attachments.

rwoll:
Thanks for the additional info, @Yexiaoxing! Let me sync with the team tomorrow, and get back to you with a review and or recommendations. 

I'm thinking of a slightly different change than you have here with the goal of (1) making the out of the box default experience better for everyone than what we're currently shipping in 1.21, and (2) reduce the need to introduce more options into the PW config. (The more options we have in our config, the more confusion there potentially is on what's best practice, the more docs there are to maintain and read, and the harder it is to troubleshoot user's experience since there's more possibilities of codepaths they are hitting.)

My proposal is to change the default behavior to:

(1) Always save the attachment on disk as `<friendly name as provided to name in attach>-<unique hash>.<ext>`. (This makes `useNameAsPrefix` not needed.) Our current pattern of just using the hash makes debugging difficult, and skimming through a fileviewer difficult. NB: We'll want some sanitization since if someone provided a name like `../../../../../foo.txt`, we wouldn't want that to effect the path where the file gets saved to.

(2) Don't allow full customization. (i.e. drop `useOriginalFilenameAsPrefix`), unless there is a compelling use case where (1) wouldn't work for folks. Although, if they absolutely need a different layout/naming convention, that can still be done in a reporter by iterating over `TestResult.attachments`.

(3) Don't add `saveBodyAsFile`; instead, this should be done at the reporter level. i.e. Change should be made in JUnit Reporter to just always convert inline attachments to on disk attachments.

@dgozman: wdyt?

Yexiaoxing:
@rwoll The reason why I am introducing more options is not to change current behavior and to prevent blocking existing users, while I do agree that too many customizations will make more confusion. Will leave it to the team for discussion.

rwoll:
> @rwoll The reason why I am introducing more options is not to change current behavior and to prevent blocking existing users, while I do agree that too many customizations will make more confusion. Will leave it to the team for discussion.

Makes sense! The intent is that https://github.com/microsoft/playwright/pull/12950#issuecomment-1111831271 only proposes backwards-compatible changes, so they'd be safe to do by default without options.

(`TestResults['attachments']` should always be the source of truth. I don't think we want filenames or disk layout to have significant meaning (unless a Reporter decides to), otherwise we can't make changes to the default in the future without fear of breaking workflows.)

dgozman:
> (1) Always save the attachment on disk as `<friendly name as provided to name in attach>-<unique hash>.<ext>`. (This makes `useNameAsPrefix` not needed.) Our current pattern of just using the hash makes debugging difficult, and skimming through a fileviewer difficult. NB: We'll want some sanitization since if someone provided a name like `../../../../../foo.txt`, we wouldn't want that to effect the path where the file gets saved to.

This is a good change. I think it should go as a separate PR. Also don't forget to sanitize/trim the friendly name like we [do here](https://github.com/microsoft/playwright/blob/1ffd18f131c0bf2fe7fe8030a11d08c9c98e0fa2/packages/playwright-test/src/testInfo.ts#L128).

> (2) Don't allow full customization. (i.e. drop `useOriginalFilenameAsPrefix`), unless there is a compelling use case where (1) wouldn't work for folks. Although, if they absolutely need a different layout/naming convention, that can still be done in a reporter by iterating over `TestResult.attachments`.

I agree, the name+hash should work for everyone.

> (3) Don't add `saveBodyAsFile`; instead, this should be done at the reporter level. i.e. Change should be made in JUnit Reporter to just always convert inline attachments to on disk attachments.

Moving to this to JUnit makes sense. We can have an option for the JUnit to do this. However, I did not find a corresponding issue - could you please share the usecase?

mxschmitt:
Closing as part of the triage process since it seemed stale. Please create an issue first if you interested in contributing new features to the project, thanks!

PR #12960, opened 2022-03-22
feat(selectors): support regular expressions in attribute selectors

dgozman:
Supports inline regex in addition to string: `_react=BookItem[author = /Ann?a/i]`.

This is similar to `text=` selector, but applies to `_react` and `_vue` selectors. In the future will also apply to `role=` selector.

### DIFF
--- a/docs/src/selectors.md
+++ b/docs/src/selectors.md
@@ -765,7 +765,7 @@ Selector examples:
 - match by component and property value **prefix**: `_react=BookItem[author ^= "Steven"]`
 - match by component and property value **suffix**: `_react=BookItem[author $= "Steven"]`
 - match by component and **key**: `_react=BookItem[key = '2']`
-
+- match by property value **regex**: `_react=[author = /Steven(\\s+King)?/i]`
 
 
 To find React element names in a tree use [React DevTools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi).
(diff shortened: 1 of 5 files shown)


docs/src/selectors.md
```diff
@@ -765,7 +765,7 @@ Selector examples:
 - match by component and property value **prefix**: `_react=BookItem[author ^= "Steven"]`
 - match by component and property value **suffix**: `_react=BookItem[author $= "Steven"]`
 - match by component and **key**: `_react=BookItem[key = '2']`
-
+- match by component and property value **regex**: `_react=BookItem[author = /Steven(\\s+King)?/i]`
```
aslushnikov, line 768:
```suggestion
- match by property value **regex**: `_react=[author = /Steven(\\s+King)?/i]`
```

docs/src/selectors.md
```diff
@@ -801,6 +801,7 @@ Selector examples:
 - match by **nested** property value: `_vue=[some.nested.value = 12]`
 - match by component and property value **prefix**: `_vue=book-item[author ^= "Steven"]`
 - match by component and property value **suffix**: `_vue=book-item[author $= "Steven"]`
+- match by component and property value **regex**: `_vue=book-item[author = /Steven(\\s+King)?/i]`
```
aslushnikov, line 804:
```suggestion
- match by property value **regex**: `_vue=[author = /Steven(\\s+King)?/i]`
```

aslushnikov (approved):

PR #12966, opened 2022-03-22
fix: print all missing libraries when used on unsupported linux distro

aslushnikov:
This patch:
- Adds 3 new host platform types:
  * `generic-linux` and `generic-linux-arm64` for the unsupported
    linux distributions
  * `<unknown>` for non-supported OS versions
- Prints a warning when downloading Ubuntu browser builds on
  unsupported Linux distribution
- Makes sure launch doctor prints all missing shared libraries
  on unknown Linux distributions
- Also prints an `apt` command as an alternative to Playwright CLI
  dependency installation.


yury-s (approved):

dgozman (approved):

aslushnikov:
TODO:
- introduce `generic-linux` that downloads ubuntu20.04 by default
- introduce `generic-linux-arm64` that downloads ubuntu20.04-arm64 by default
- introduce `unknown` that just throws

dgozman (approved):

PR #12968, opened 2022-03-22
fix(show-trace): traceModel was removed by mistake

### LINKED ISSUE #12725: [BUG] Online traceviewer random 404s
I've created a sample trace that I've been sharing to show off how cool Playwright is, however people are reporting that things like image thumbnails are randomly failing. After a little digging I can see that some of the service worker requests result in 404s.

You can try this yourself by refreshing this sample a few times and watching the service worker network requests:
https://trace.playwright.dev/?trace=https://pw-trace-demo.s3.eu-west-1.amazonaws.com/Search_For_DotNet_Core_chromium_macos_2022-03-08-00-24-44.zip

I'm unsure where exactly the problem lies, it could be in zip.js for example 🤔 Could be worth upgrading to the latest as there are a couple of fixes (not sure if they would resolve what we are seeing here).

image
image

liuxingbaoyu:
fixes: https://github.com/microsoft/playwright/issues/12725

### DIFF
--- a/packages/trace-viewer/src/sw.ts
+++ b/packages/trace-viewer/src/sw.ts
@@ -32,12 +32,13 @@ const scopePath = new URL(self.registration.scope).pathname;
 
 const loadedTraces = new Map<string, { traceModel: TraceModel, snapshotServer: SnapshotServer }>();
 
-const loadedMap = new Map<string, string>();
+const clientIdToTraceUrl = new Map<string, string>();
 
 async function loadTrace(trace: string, clientId: string, progress: (done: number, total: number) => void): Promise<TraceModel> {
   const entry = loadedTraces.get(trace);
-  loadedMap.set(clientId, trace);
-  if (entry) return entry.traceModel;
+  clientIdToTraceUrl.set(clientId, trace);
+  if (entry)
+    return entry.traceModel;
   const traceModel = new TraceModel();
   await traceModel.load(trace, progress);
   const snapshotServer = new SnapshotServer(traceModel.storage());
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/web/traceViewer/sw.ts
```diff
@@ -30,16 +30,18 @@ self.addEventListener('activate', function(event: any) {
 
 const scopePath = new URL(self.registration.scope).pathname;
 
-const loadedTraces = new Map<string, { traceModel: TraceModel, snapshotServer: SnapshotServer, clientId: string }>();
+const loadedTraces = new Map<string, { traceModel: TraceModel, snapshotServer: SnapshotServer }>();
+
+const loadedMap = new Map<string, string>();
```
pavelfeldman, line 35:
```suggestion
const clientIdToTraceUrl = new Map<string, string>();
```

packages/playwright-core/src/web/traceViewer/sw.ts
```diff
@@ -30,16 +30,18 @@ self.addEventListener('activate', function(event: any) {
...
 
 async function loadTrace(trace: string, clientId: string, progress: (done: number, total: number) => void): Promise<TraceModel> {
   const entry = loadedTraces.get(trace);
-  if (entry)
-    return entry.traceModel;
+  loadedMap.set(clientId, trace);
+  if (entry) return entry.traceModel;
```
pavelfeldman, line 40: style:

```suggestion
  if (entry)
    return entry.traceModel;
```

packages/playwright-core/src/web/traceViewer/sw.ts
```diff
@@ -117,10 +119,12 @@ async function doFetch(event: FetchEvent): Promise<Response> {
...
 async function gc() {
+  const clients = await self.clients.matchAll();
   const usedTraces = new Set<string>();
-  for (const [traceUrl, entry] of loadedTraces) {
-    const client = await self.clients.get(entry.clientId);
-    if (client)
+
+  for (const [clientId, traceUrl] of loadedMap) {
+    // @ts-ignore
+    if (!clients.find(c => c.id === clientId)) loadedMap.delete(clientId); else
```
pavelfeldman, line 127:
```suggestion
    if (!clients.find(c => c.id === clientId))
      loadedMap.delete(clientId);
    else
      usedTraces.add(traceUrl);
```

pavelfeldman (commented):
Great job! Some style comments and it is good to go!

pavelfeldman (approved):

liuxingbaoyu:
finished!

raghavthind2005 (approved):

Issue #11182, opened 2022-01-05
[Feature] Add byRole selector


pavelfeldman:
Extracted from https://github.com/microsoft/playwright/issues/11145

https://testing-library.com/docs/queries/byrole/

sebinsua:
[This comment describing the implementation](https://github.com/microsoft/playwright/issues/11145#issuecomment-1007647083) might be helpful:

> `ByRole` follows ARIA and related specifications e.g. the `name` option in `ByRole(role, { name })` implements [ACCNAME](https://www.w3.org/TR/accname-1.1/).
> 
> `role` is tied to [HTML-ARIA](https://www.w3.org/TR/html-aria/).
> 
> The [documentation for `ByRole`](https://testing-library.com/docs/queries/byrole/) goes into more detail.

kevin940726:
Just found this, and I want to shamelessly plug my experimental library [`role-selector`](https://github.com/kevin940726/role-selector) here. Instead of using `@testing-library/dom` as the underlying engine, it uses [`axe-core`](https://github.com/dequelabs/axe-core) to compute the accessible roles and attributes. An example selector will look like following:

```js
page.locator('role=button[name="My button"]');
```

Although the library will technically work in puppeteer or even jsdom, it's designed with playwright in mind, thus intentionally not implementing duplicated features like `byText` or `visible`.

The selector syntax is similar to puppeteer's [aria selector](https://developer.chrome.com/blog/puppetaria/), but prioritizing `role` more than `name` to follow the philosophy of `@testing-library/*`'s `byRole`.

The library is still under development, but I plan to experiment with it in [gutenberg](https://github.com/WordPress/gutenberg)'s e2e test suites when we migrate to playwright (soon).

kevin940726:
Just want to update that we're now using the official built-in role selectors in Playwright and it works perfectly fine [migrating](https://github.com/WordPress/gutenberg/pull/41389) from the `role-selector` package!

mxschmitt:
Closing as per above, seems to be working as it is for a large customer.

PR #12999, opened 2022-03-23
feat(selectors): role selector engine

dgozman:
This introduces `role=button[name="Click me"][pressed]` attribute-style role selector. It is only available under `env.PLAYWRIGHT_EXPERIMENTAL_FEATURES`.

Supported attributes:
- `role` is required, for example `role=button`;
- `name` is accessible name, supports matching operators and regular expressions: `role=button[name=/Click(me)?/]`;
- `checked` boolean/mixed, for example `role=checkbox[checked=false]`;
- `selected` boolean, for example `role=option[selected]`;
- `expanded` boolean, for example `role=button[expanded=true]`;
- `disabled` boolean, for example `role=button[disabled]`;
- `level` number, for example `role=heading[level=3]`;
- `pressed` boolean/mixed, for example `role=button[pressed="mixed"]`;
- `includeHidden` - by default, only non-hidden elements are considered. Passing `role=button[includeHidden]` matches hidden elements as well.

References #11182.

### DIFF
--- a/packages/playwright-core/src/server/dom.ts
+++ b/packages/playwright-core/src/server/dom.ts
@@ -28,7 +28,7 @@ import { Progress, ProgressController } from './progress';
 import { SelectorInfo } from './selectors';
 import * as types from './types';
 import { TimeoutOptions } from '../common/types';
-import { isUnderTest } from '../utils/utils';
+import { experimentalFeaturesEnabled, isUnderTest } from '../utils/utils';
 
 type SetInputFilesFiles = channels.ElementHandleSetInputFilesParams['files'];
 export type InputFilesItems = { files?: SetInputFilesFiles, localPaths?: string[] };
@@ -104,6 +104,7 @@ export class FrameExecutionContext extends js.ExecutionContext {
           ${isUnderTest()},
           ${this.frame._page._delegate.rafCountForStablePosition()},
           "${this.frame._page._browserContext._browser.options.name}",
+          ${experimentalFeaturesEnabled()},
           [${custom.join(',\n')}]
         );
         })();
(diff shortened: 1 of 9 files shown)


aslushnikov (approved):

kevin940726:
Nice work!

Now that we have all these accessible attributes helpers, is it worth making them available under locator and assertions as well?

For instance, we want to know whether an element is expanded or not:

```js
const button = page.locator('role=button[name=Expand me]');
const isExpanded = await button.expanded(); // `true` or `false`
await expect(button).toBeExpanded(); // Error if the button is not expanded
```

I think it's only natural to make them available both via querying and accessing.

Currently, we only have `page.accessibility`, but that uses browsers' API and it's platform-specific, different from the JS implementation injected into the browsers.

dgozman:
@kevin940726 That's a great suggestion! Could you please file an issue with your thoughts?

Also, we are releasing v1.21 soon that will have `role=` selectors available under an environment variable. Would you like to give them a try and provide feedback?

kevin940726:
@dgozman Awesome! Opened in https://github.com/microsoft/playwright/issues/13517.

> Also, we are releasing v1.21 soon that will have `role=` selectors available under an environment variable. Would you like to give them a try and provide feedback?

I already gave them a try before posting that comment using the `next` release 😆. It worked perfectly without having me change any of the code when switching from [`role-selector`](https://github.com/kevin940726/role-selector)! Glad to know that it's supported natively 🎉.

We'll make the switch from `role-selector` to using the native role selectors when it's ready in the stable release at [gutenberg](https://github.com/WordPress/gutenberg) :)

PR #13013, opened 2022-03-23
docs(parallel): tell about fully parallel mode

dgozman:


aslushnikov (approved):

docs/src/test-parallel-js.md
```diff
@@ -88,6 +88,30 @@ test('runs in parallel 1', async ({ page }) => { /* ... */ });
...
+```js js-flavor=ts
+// playwright.config.ts
+import { PlaywrightTestConfig } from '@playwright/test';
+
+const config: PlaywrightTestConfig = {
+  fullyParallel: true,
+};
+export default config;
+```
+
```
cliffordfajardo, line 114: leaving myself a TODO: to show an example of fullyParallel used for only 1 test file
essentially co-locating an example here so its easier not to miss this awesome feature

## References
- https://playwright.dev/docs/api/class-test#test-describe-configure

PR #13032, opened 2022-03-24
chore(test-runner): support self signed certificate in webServer

### LINKED ISSUE #12976: [Feature] Add option to ignoreHTTPSErrors in webServer
I noticed that playwright cannot correctly specify if the server is running or not and starts tests too fast when there are problems with https. When I have added check for URL and port using opening new window and ignoring HTTPS errors it works fine, so adding an option to ignoreHTTPSErrors in webserver config would do the job.

mxschmitt:
Resolves https://github.com/microsoft/playwright/issues/12976

### DIFF
--- a/packages/playwright-test/src/webServer.ts
+++ b/packages/playwright-test/src/webServer.ts
@@ -35,7 +35,7 @@ export class WebServer {
   private _processExitedPromise!: Promise<any>;
 
   constructor(private readonly config: WebServerConfig, private readonly reporter: Reporter) {
-    this._isAvailable = getIsAvailableFunction(config);
+    this._isAvailable = getIsAvailableFunction(config, reporter.onStdErr?.bind(reporter));
   }
 
   public static async create(config: WebServerConfig, reporter: Reporter): Promise<WebServer> {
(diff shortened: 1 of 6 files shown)


packages/playwright-test/src/webServer.ts
```diff
@@ -122,13 +122,16 @@ async function isPortUsed(port: number): Promise<boolean> {
   return await innerIsPortUsed('127.0.0.1') || await innerIsPortUsed('::1');
 }
 
-async function isURLAvailable(url: URL) {
+async function isURLAvailable(url: URL, ignoreHTTPSErrors?: boolean) {
   return new Promise<boolean>(resolve => {
-    (url.protocol === 'https:' ? https : http).get(url, res => {
+    (url.protocol === 'https:' ? https : http).get(url, {
+      rejectUnauthorized: !ignoreHTTPSErrors,
```
mxschmitt, line 128: in theory we only need to pass it when its 'http:' protocol, but probably still fine.

packages/playwright-test/src/webServer.ts
```diff
@@ -122,13 +122,21 @@ async function isPortUsed(port: number): Promise<boolean> {
...
   return new Promise<boolean>(resolve => {
-    (url.protocol === 'https:' ? https : http).get(url, res => {
+    (isHttps ? https : http).get(url, requestOptions, res => {
       res.resume();
       const statusCode = res.statusCode ?? 0;
       resolve(statusCode >= 200 && statusCode < 300);
-    }).on('error', () => {
+    }).on('error', error => {
+      if ((error as NodeJS.ErrnoException).code === 'DEPTH_ZERO_SELF_SIGNED_CERT')
+        onStdErr?.(`[WebServer] Self-signed certificate detected, Try adding ignoreHTTPSErrors: true to config.webServer.`);
```
yury-s, line 137: nit: "detected. Try"

yury-s (approved):

PR #13041, opened 2022-03-24
feat(reporter): support different attachment root for junit reporter

Yexiaoxing:
This resolves part of #13026.


pavelfeldman:
Closing as stale and conflicting, not sure what the action on this one is, but it looks like it does not progress. Let's start with filing a bug.

pesimeao:
I still believe this PR is very useful. It covers a case that is generating wrong reports (screenshot with wrong path)

PR #13048, opened 2022-03-24
chore: add warning about FullConfig

rwoll:


rwoll:
FullConfig is used internally all over, so we'll need to think through this in order to (1) not break public API, but move off of it from internal usage.

PR #13064, opened 2022-03-25
Show how to set video save path

jameshfisher:


docs/src/videos.md
```diff
@@ -82,13 +82,20 @@ var context = await browser.NewContextAsync(new BrowserNewContextOptions
 await context.CloseAsync();
 ```
 
-Saved video files will appear in the specified folder. They all have generated unique names.
-For the multi-page scenarios, you can access the video file associated with the page via the
-[`method: Page.video`].
-
+By default, the video file is saved in the `dir` specified in the context configuration, with a unique random filename.
+This can be viewed or overridden with [`method: Page.video`]:
```
pavelfeldman, line 86: It is not really overridden by saveAs, it actually is copied over to the file with the new name instead.

PR #13074, opened 2022-03-25
chore: remove internal dep on FullConfig

### LINKED ISSUE #13058: [internal] Remove Internal Dep on FullConfig
FullConfig is part of the public API. We want to stop adding to it. However, this type is currently used throughout all the internal runner plumbing, so internal refactoring of the runner is required so we can freeze the public API but still plumb configs through the runner.

cc: @dgozman @pavelfeldman

rwoll:
Resolves #13058.

This blocks adding #8206 since we don't want the new launch configs
publicly available on FullConfig.


packages/playwright-test/src/types.ts
```diff
@@ -33,3 +34,10 @@ export interface TestStepInternal {
   forceNoParent: boolean;
   location?: Location;
 }
+
+/**
+ * FullConfigInternal allows the plumbing of configuration details throughout the Test Runner without
+ * increasing the surface aread of the public API type called FullConfig.
```
yury-s, line 40: nit: aread -> area

utils/generate_types/overrides-test.d.ts
```diff
@@ -195,6 +195,9 @@ export interface Config<TestArgs = {}, WorkerArgs = {}> extends TestConfig {
   use?: UseOptions<TestArgs, WorkerArgs>;
 }
 
+// [internal] !!! DO NOT ADD TO THIS !!!
+// [internal] It is part of the public API and is computed from the user's config.
+// [internal] If you need new fields internally, add them to InternalFullConfig instead.
```
yury-s, line 200: typo: add them to FullConfigInternal

packages/playwright-test/src/loader.ts
```diff
@@ -15,7 +15,8 @@
  */
 
 import { installTransform, setCurrentlyLoadingTestFile } from './transform';
-import type { FullConfig, Config, FullProject, Project, ReporterDescription, PreserveOutput } from './types';
+import type { Config, FullProject, Project, ReporterDescription, PreserveOutput } from './types';
+import type { FullConfigInternal as FullConfig } from './types';
```
yury-s, line 19: Let's update the code to explicitly use FullConfigInternal in the signature and avoid possible confusion?

yury-s (approved):

PR #13083, opened 2022-03-25
feat(test-runner): introduce GlobalInfo

rwoll:
This change introduces GlobalInfo: an API to collect runner-level data (like attachments). It's analogous to TestInfo, but at the global level.

The API is kept as small as possible for now.

Some changes to note:
- GlobalInfo is available to user's globalSetup; it's **not** available in globalTeardown
- `async GlobalInfo.attach(…)` is backed by a similar impl. to `async TestInfo.attach(…)`, and is preferred over pushing directly onto the attachments array.
- We no longer have FullConfig.attachments
- The HTML Reporter CI plugin has been migrated to use this new API
- The attachments are available in Reporters (read-only-ish), via the root suite
- Many of our reporters don't expose the root Suite, so attachments are promoted to a top-level field
- Global attachments use a throwaway directory that re-uses our existing directory structure; however, project level overrides are NOT respected 
- This API is not currently documented; let's continue dogfooding it and decide if we want to release it publicly.


tests/playwright-test/reporter-raw.spec.ts
```diff
@@ -238,6 +238,53 @@ test(`testInfo.attach should save attachments via inline attachment`, async ({ r
...
+    const attachment = json.attachments[0];
+    expect(attachment.name).toBe('inline.txt');
+    expect(attachment.contentType).toBe('text/plain');
+    expect(Buffer.from(attachment.body, 'base64').toString()).toEqual('inline');
+  }
+  {
+    const attachment = json.attachments[1];
+    expect(attachment.name).toBe('external.txt');
+    expect(attachment.contentType).toBe('text/plain');
+    const contents = fs.readFileSync(attachment.path);
```
aslushnikov, line 283: shall we check for path to be inside outptu dir?

tests/playwright-test/reporter-json.spec.ts
```diff
@@ -235,3 +235,50 @@ test('should add line in addition to file json without CI', async ({ runInlineTe
   expect(stripAnsi(result.output)).toContain('[1/1] a.test.js:6:7 › one');
   expect(fs.existsSync(testInfo.outputPath('a.json'))).toBeTruthy();
 });
+
+test(`GlobalInfo.attach works`, async ({ runInlineTest }, testInfo) => {
```
aslushnikov, line 239: will not be needed

aslushnikov (approved):

utils/generate_types/overrides-test.d.ts
```diff
@@ -291,6 +291,10 @@ export interface TestInfo {
   outputPath: (...pathSegments: string[]) => string;
 }
 
+export interface GlobalInfo {
+  attachments(): { name: string, path?: string, body?: Buffer, contentType: string }[];
+  attach(name: string, options?: { contentType?: string, path?: string, body?: string | Buffer }): Promise<void>;
+}
```
rwoll, line 297:
```suggestion
}

```

packages/playwright-test/src/runner.ts
```diff
@@ -59,9 +60,11 @@ export class Runner {
   private _loader: Loader;
   private _reporter!: Reporter;
   private _internalGlobalSetups: Array<InternalGlobalSetupFunction> = [];
+  private _globalInfo: GlobalInfo;
```
dgozman, line 63:
```suggestion
  private _globalInfo: GlobalInfoImpl;
```

packages/playwright-test/src/globalInfo.ts
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+import { GlobalInfo } from './types';
+import { normalizeAndSaveAttachment } from './util';
+import fs from 'fs/promises';
+export class GlobalInfoImpl implements GlobalInfo {
+  private _outputDir: string;
+  private _attachments: { name: string; path?: string | undefined; body?: Buffer | undefined; contentType: string; }[] = [];
+
+  constructor(outputDir: string) {
```
dgozman, line 23: I think we can safely pass `FullConfigInternal` here, we'll need it later.

packages/playwright-test/src/types.ts
```diff
@@ -40,6 +40,7 @@ export interface TestStepInternal {
  * increasing the surface area of the public API type called FullConfig.
  */
 export interface FullConfigInternal extends FullConfigPublic {
+  globalOutputDir: string;
```
dgozman, line 43: Everything here should be underscored, to not confuse folks printing their full configs. Might as well add a comment for this.

packages/playwright-test/types/testReporter.d.ts
```diff
@@ -84,6 +84,7 @@ export interface Suite {
    * listed in the child [suite.suites](https://playwright.dev/docs/api/class-suite#suite-suites).
    */
   tests: TestCase[];
+  attachments: { name: string, path?: string, body?: Buffer, contentType: string }[];
```
dgozman, line 87: If this is public, add it to the docs!

packages/playwright-test/types/test.d.ts
```diff
@@ -1727,6 +1727,11 @@ export interface TestInfo {
   outputPath: (...pathSegments: string[]) => string;
 }
 
+export interface GlobalInfo {
```
dgozman, line 1730: If this is public, add it to the docs!

packages/playwright-test/src/loader.ts
```diff
@@ -168,7 +169,7 @@ export class Loader {
     return suite;
   }
 
-  async loadGlobalHook(file: string, name: string): Promise<(config: FullConfigInternal) => any> {
+  async loadGlobalHook(file: string, name: string): Promise<(config: FullConfigInternal, globalInfo?: GlobalInfo) => any> {
```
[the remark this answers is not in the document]
rwoll: This function produces the globalSetup, so it's used: https://github.com/microsoft/playwright/pull/13083/files#diff-b6a8a37f7e3925fe6761b813b8b9350cf8755afe66dabf1e34020e8030fc0751R461

packages/playwright-test/src/types.ts
rwoll: Agreed—when this PR was first authored, there were no other members here! But now that it has friends, definitely should be underscored!

packages/playwright-test/src/reporters/raw.ts
```diff
@@ -129,21 +130,22 @@ class RawReporter {
...
         throw new Error('Internal error, could not create report file');
-      const report = this.generateProjectReport(this.config, suite);
+      const report = this.generateProjectReport(this.config, suite, globalAttachments);
       fs.writeFileSync(reportFile, JSON.stringify(report, undefined, 2));
     }
   }
 
-  generateAttachments(config: FullConfig): JsonAttachment[] {
-    return this._createAttachments((config as FullConfigInternal)._attachments);
+  generateAttachments(attachments: Suite['attachments']): JsonAttachment[] {
```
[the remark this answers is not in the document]
rwoll: oops! updated—removed this, and renamed `_createAttachments` to `generateAttachments()`; it's called into from HTML Reporter, too.

dgozman (approved):

docs/src/test-api/class-globalinfo.md
```diff
@@ -0,0 +1,123 @@
+# class: GlobalInfo
+* langs: js
+
+`GlobalInfo` contains information on the overall test run. The information spans projects and tests.
+
+You can write to GlobalInfo via your Global Setup hook, and read from it in a [Custom Reporter](../test-reporters.md):
```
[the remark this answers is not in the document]
rwoll: Amended! However, note that JSON reporter no longer does (a75207d7cab5a90090dc7521b102ce2f99dfdfc9 per review with @aslushnikov) and HTML Reporter only knows about our magic git attachments, which won't be there for general users until we settle land plugins (and have an official plugin to do the git magic).

rwoll:
All feedback addressed! This should be ready to land! 

There are conflicts, but I will wait to resolve until right before we land. I'm waiting on 1.21 to be branched as this is not intended to go into 1.21, and have already resolved conflicts a few times.

PR #13105, opened 2022-03-27
fix: resolve ts compilerOptions.paths with prefixes and suffixes

### LINKED ISSUE #12851: [BUG] ts compilerOptions.paths with * in middle of value are not being resolved properly
**Context:**
- Playwright Version: 1.19.2
- Operating System: Mac
- Node.js version: v16.14.0

**Description**

[Recently added](https://github.com/microsoft/playwright/issues/7121#issuecomment-1009140849) support of typescript `compilerOptions.paths` config property doesn't work for values with `*` char in the middle. 
Example:

```
"paths": {
  "@company/prefix-*": [
    "packages/prefix-*/src"
  ]
}
```

The config above with error:

```sh
$ playwright test

Running 0 test using 0 worker


Error: Cannot find module '<path-to-project>project/packages/prefix-*/src'
Require stack:
...
```

Current support of paths works fine for values like `foo/*` or `foo`. But not for cases like `foo-*/bar`. However `tsc` is able to handle cases like this one.

**Reproduction repo**

https://github.com/ivankoleda/playwright-ts-paths-bug-reproduction

ivankoleda:
Resolves #12851
In addition to handling prefixes/suffixes in `compilerOptions.paths` added the support for matching the longest prefix.

### DIFF
--- a/packages/playwright-test/src/transform.ts
+++ b/packages/playwright-test/src/transform.ts
@@ -100,23 +100,52 @@ export function resolveHook(filename: string, specifier: string): string | undef
     return;
   const tsconfig = loadAndValidateTsconfigForFile(filename);
   if (tsconfig) {
+    let longestPrefixLength = -1;
+    let pathMatchedByLongestPrefix: string | undefined;
+
     for (const { key, values } of tsconfig.paths) {
-      const keyHasStar = key[key.length - 1] === '*';
-      const matches = specifier.startsWith(keyHasStar ? key.substring(0, key.length - 1) : key);
-      if (!matches)
-        continue;
+      let matchedPartOfSpecifier = specifier;
+
+      const [keyPrefix, keySuffix] = key.split('*');
+      if (key.includes('*')) {
+        // * If pattern contains '*' then to match pattern "<prefix>*<suffix>" module name must start with the <prefix> and end with <suffix>.
+        // * <MatchedStar> denotes part of the module name between <prefix> and <suffix>.
+        // * If module name can be matches with multiple patterns then pattern with the longest prefix will be picked.
+        // https://github.com/microsoft/TypeScript/blob/f82d0cb3299c04093e3835bc7e29f5b40475f586/src/compiler/moduleNameResolver.ts#L1049
+        if (keyPrefix) {
+          if (!specifier.startsWith(keyPrefix))
+            continue;
+          matchedPartOfSpecifier = matchedPartOfSpecifier.substring(keyPrefix.length, matchedPartOfSpecifier.length);
+        }
+        if (keySuffix) {
+          if (!specifier.endsWith(keySuffix))
+            continue;
+          matchedPartOfSpecifier = matchedPartOfSpecifier.substring(0, matchedPartOfSpecifier.length - keySuffix.length);
+        }
+      } else {
+        if (specifier !== key)
+          continue;
+        matchedPartOfSpecifier = specifier;
+      }
+
       for (const value of values) {
-        const valueHasStar = value[value.length - 1] === '*';
-        let candidate = valueHasStar ? value.substring(0, value.length - 1) : value;
-        if (valueHasStar && keyHasStar)
-          candidate += specifier.substring(key.length - 1);
+        let candidate: string = value;
+
+        if (value.includes('*'))
+          candidate = candidate.replace('*', matchedPartOfSpecifier);
         candidate = path.resolve(tsconfig.absoluteBaseUrl, candidate.replace(/\//g, path.sep));
         for (const ext of ['', '.js', '.ts', '.mjs', '.cjs', '.jsx', '.tsx']) {
-          if (fs.existsSync(candidate + ext))
-            return candidate;
+          if (fs.existsSync(candidate + ext)) {
+            if (keyPrefix.length > longestPrefixLength) {
+              longestPrefixLength = keyPrefix.length;
+              pathMatchedByLongestPrefix = candidate;
+            }
+          }
         }
       }
     }
+    if (pathMatchedByLongestPrefix)
+      return pathMatchedByLongestPrefix;
   }
   if (specifier.endsWith('.js')) {
     const resolved = path.resolve(path.dirname(filename), specifier);
--- a/tests/playwright-test/resolver.spec.ts
+++ b/tests/playwright-test/resolver.spec.ts
@@ -163,3 +163,110 @@ test('should respect baseurl w/o paths', async ({ runInlineTest }) => {
   expect(result.passed).toBe(1);
   expect(result.output).not.toContain(`Could not`);
 });
+
+test('should respect complex path resolver', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'playwright.config.ts': `
+      export default {
+        projects: [{name: 'foo'}],
+      };
+    `,
+    'tsconfig.json': `{
+      "compilerOptions": {
+        "target": "ES2019",
+        "module": "commonjs",
+        "lib": ["esnext", "dom", "DOM.Iterable"],
+        "baseUrl": ".",
+        "paths": {
+          "prefix-*": ["./prefix-*/bar"],
+          "prefix-*-suffix": ["./prefix-*-suffix/bar"],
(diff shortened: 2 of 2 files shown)


packages/playwright-test/src/transform.ts
```diff
@@ -101,23 +101,55 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
+      // * If module name can be matches with multiple patterns then pattern with the longest prefix will be picked.
+      // https://github.com/microsoft/TypeScript/blob/f82d0cb3299c04093e3835bc7e29f5b40475f586/src/compiler/moduleNameResolver.ts#L1049
+      const [prefix, suffix] = key.split('*');
+      if (prefix) {
+        matches = matches && specifier.startsWith(prefix);
+        matchedPartOfSpecifier = matchedPartOfSpecifier.replace(prefix, '');
+      }
+      if (suffix) {
+        matches = matches && specifier.endsWith(suffix);
+        matchedPartOfSpecifier = matchedPartOfSpecifier.replace(suffix, '');
```
pavelfeldman, line 124: I think following input yields a bug. Let's add a test for it.

```js
key === '*bar';
specifier === 'barfoobar';
// should be 
matchedPartOfSpecifier === 'barfoo'
// actual 
matchedPartOfSpecifier === 'foobar'
```

packages/playwright-test/src/transform.ts
```diff
@@ -101,23 +101,55 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
       continue;
     for (const value of values) {
-      const valueHasStar = value[value.length - 1] === '*';
-      let candidate = valueHasStar ? value.substring(0, value.length - 1) : value;
-      if (valueHasStar && keyHasStar)
-        candidate += specifier.substring(key.length - 1);
+      const valueHasStar = value.includes('*');
+
+      let candidate: string = value;
+      if (valueHasStar) candidate = candidate.replace('*', matchedPartOfSpecifier);
```
pavelfeldman, line 137: style: we put body on the next line.

packages/playwright-test/src/transform.ts
```diff
@@ -101,23 +101,55 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
       candidate = path.resolve(tsconfig.absoluteBaseUrl, candidate.replace(/\//g, path.sep));
       for (const ext of ['', '.js', '.ts', '.mjs', '.cjs', '.jsx', '.tsx']) {
-        if (fs.existsSync(candidate + ext))
-          return candidate;
+        if (fs.existsSync(candidate + ext)) {
+          const [keyPrefix] = key.split('*');
```
pavelfeldman, line 141: Don't split in the loop. Btw, you also have it computed once above.

packages/playwright-test/src/transform.ts
```diff
@@ -101,23 +101,55 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
+          // This comparassion is `>`, not `>=` to not override matches with same prefix length.
```
pavelfeldman, line 142: No need for this comment.

pavelfeldman (commented):
Nice! Some comments below.

packages/playwright-test/src/transform.ts
```diff
@@ -100,23 +100,53 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
+        // * <MatchedStar> denotes part of the module name between <prefix> and <suffix>.
+        // * If module name can be matches with multiple patterns then pattern with the longest prefix will be picked.
+        // https://github.com/microsoft/TypeScript/blob/f82d0cb3299c04093e3835bc7e29f5b40475f586/src/compiler/moduleNameResolver.ts#L1049
+        if (keyPrefix) {
+          matches = specifier.startsWith(keyPrefix);
+          matchedPartOfSpecifier = matchedPartOfSpecifier.replace(keyPrefix, '');
+        }
+        if (keySuffix && matches) {
+          matches = specifier.endsWith(keySuffix);
+          matchedPartOfSpecifier = matchedPartOfSpecifier.replace(new RegExp(`${keySuffix}$`), '');
```
pavelfeldman, line 122: I believe that this now has a new bug:
```js
keySuffix === '[abc]';
specifier === 'foo/[abc]';
```

use `matchedPartOfSpecifier.substring` instead.

packages/playwright-test/src/transform.ts
```diff
@@ -100,23 +100,53 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
+
+      const keyHasStar = key.includes('*');
+      const [keyPrefix, keySuffix] = key.split('*');
+      if (keyHasStar) {
+        // * If pattern contains '*' then to match pattern "<prefix>*<suffix>" module name must start with the <prefix> and end with <suffix>.
+        // * <MatchedStar> denotes part of the module name between <prefix> and <suffix>.
+        // * If module name can be matches with multiple patterns then pattern with the longest prefix will be picked.
+        // https://github.com/microsoft/TypeScript/blob/f82d0cb3299c04093e3835bc7e29f5b40475f586/src/compiler/moduleNameResolver.ts#L1049
+        if (keyPrefix) {
+          matches = specifier.startsWith(keyPrefix);
```
pavelfeldman, line 117: you can continue right away if it does not match, that way you don't need `matches`:

```js
if (!specifier.startsWith(keyPrefix))
  continue;
```

packages/playwright-test/src/transform.ts
```diff
@@ -100,23 +100,52 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
+        }
+        if (keySuffix) {
+          if (specifier.endsWith(keySuffix))
+            matchedPartOfSpecifier = matchedPartOfSpecifier.substring(0, matchedPartOfSpecifier.lastIndexOf(keySuffix));
+          else continue;
+        }
+      } else {
+        if (specifier === key)
+          matchedPartOfSpecifier = specifier;
+        else continue;
```
pavelfeldman, line 128: style: ditto

packages/playwright-test/src/transform.ts
```diff
@@ -100,23 +100,52 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
+        // https://github.com/microsoft/TypeScript/blob/f82d0cb3299c04093e3835bc7e29f5b40475f586/src/compiler/moduleNameResolver.ts#L1049
+        if (keyPrefix) {
+          if (specifier.startsWith(keyPrefix))
+            matchedPartOfSpecifier = matchedPartOfSpecifier.replace(keyPrefix, '');
+          else continue;
+        }
+        if (keySuffix) {
+          if (specifier.endsWith(keySuffix))
+            matchedPartOfSpecifier = matchedPartOfSpecifier.substring(0, matchedPartOfSpecifier.lastIndexOf(keySuffix));
+          else continue;
```
pavelfeldman, line 123: style: ditto

pavelfeldman (commented):
Looks good, a couple of style nits and it is good to go!

packages/playwright-test/src/transform.ts
```diff
@@ -100,23 +100,52 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
+      const [keyPrefix, keySuffix] = key.split('*');
+      if (key.includes('*')) {
+        // * If pattern contains '*' then to match pattern "<prefix>*<suffix>" module name must start with the <prefix> and end with <suffix>.
+        // * <MatchedStar> denotes part of the module name between <prefix> and <suffix>.
+        // * If module name can be matches with multiple patterns then pattern with the longest prefix will be picked.
+        // https://github.com/microsoft/TypeScript/blob/f82d0cb3299c04093e3835bc7e29f5b40475f586/src/compiler/moduleNameResolver.ts#L1049
+        if (keyPrefix) {
+          if (!specifier.startsWith(keyPrefix))
+            continue;
+          matchedPartOfSpecifier = matchedPartOfSpecifier.replace(keyPrefix, '');
```
ivankoleda, line 118: Actually this `replace` can also be replaced with `substring` for the sake of speed, let me do it as well

pavelfeldman (approved):
Approved. Will wait for a substring and merge!

PR #13124, opened 2022-03-28
feat(touchscreen): add Touchscreen move down up methods

kaivean:
### Why

Some business logics is valid when a page is pressed down  by user and Once the page is pressed up, the logics changes. So  I need to validate this logic under `down`. 

On touch move, I need to record user scroll behavior and send log.

### How 
```javascript
await page.touchscreen.down(100, 100);

await page.touchscreen.up();

await page.touchscreen.move(200, 200, 200, 180);
```

### DIFF
--- a/docs/src/api/class-touchscreen.md
+++ b/docs/src/api/class-touchscreen.md
@@ -3,14 +3,13 @@
 The Touchscreen class operates in main-frame CSS pixels relative to the top-left corner of the viewport. Methods on the
 touchscreen can only be used in browser contexts that have been initialized with `hasTouch` set to true.
 
-## async method: Touchscreen.tap
+## async method: Touchscreen.down
+Dispatches a `touchstart` event with a single touch at the position ([`param: x`],[`param: y`]).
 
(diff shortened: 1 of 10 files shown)


pavelfeldman:
Let's file a bug that explains the functionality being added, scenarios that need it, etc. We would also need to add this for WebKit so that we could test iPhones, etc. But it can be a separate patch.

kaivean:
> Let's file a bug that explains the functionality being added, scenarios that need it, etc. We would also need to add this for WebKit so that we could test iPhones, etc. But it can be a separate patch.

@pavelfeldman  I updated the description

pavelfeldman:
Does not page.mouse.down / move / up not work for your scenario? Could you be specific about the events you intend to process. You should file a bug to make sure we are on the same page.

dgozman:
@kaivean Are you still interested in this functionality? If so, could you please file an issue and explain your usecases?

pavelfeldman:
Respectfully closing this. Please feel free to re-file should you get back to this effort!

Issue #8740, opened 2021-09-07
[Feature] add expect(locator).isInViewport() / isIntersectingViewport


cardinalX:
Now we have to use this method, the implementation of which is taken from puppeteer. It would be more convenient if it was in the playwright itself
```ts
  /**
   * @returns {!Promise<boolean>}
   * @summary Проверка, что нужный элемент находится в текущей области видимости(viewPort)
   */
  isIntersectingViewport(selector: string): Promise<boolean> {
    return this.page.$eval(selector, async element => {
      const visibleRatio: number = await new Promise(resolve => {
        const observer = new IntersectionObserver(entries => {
          resolve(entries[0].intersectionRatio);
          observer.disconnect();
        });
        observer.observe(element);
        // Firefox doesn't call IntersectionObserver callback unless
        // there are rafs.
        requestAnimationFrame(() => {});
      });
      return visibleRatio > 0;
    });
  }
```

Context: https://playwright.slack.com/archives/CSUHZPVLM/p1631008178339400

pavelfeldman:
What's your use case for the method? (Helps us prioritize)

cardinalX:
> What's your use case for the method? (Helps us prioritize)

We have some card elements, on clicking on which another subcards appears, and this appearance should occur in the user's field of view(there may be bugs that the subcards appeared below and the user will not even notice that something appeared)

cardinalX:
I can, in principle, use the above method, which you actually suggested, thank you:) 
Therefore, I would not give this task a high priority if there are no new applicants

federico-moretti:
I also would like this feature as we have an anchor tag and on mobile we want to check that after the click the page scrolls down to the anchor correctly.

NikkTod:
We have long list of vertical pages of a report. When we select a particular page ( e.g. 10) we want to verify if the scroll has been performed and the page is now in the viewport.

cliffordfajardo:
**Use case from our app**
Our page is loaded to a specific section of the page (essentially your pretty scrolled down)
User clicks the breadcrumb, and it scrolls them up to the first container of their page

I would like to assert the top container is in the viewport

https://user-images.githubusercontent.com/6743796/160018972-d69db8f3-4d31-4da0-8229-9d0c72d0a114.mp4

cliffordfajardo:
Hi @pavelfeldman  @dgozman  @mxschmitt ,

**Use case from our app** (greetings from MSFT as well, well technically LinkedIn 😄 )
[...]

PR #13134, opened 2022-03-28
test: out-of-viewport should be considered visible

### LINKED ISSUE #13131: [BUG] waitFor({ state: 'hidden' }) does not work if element animates out of the viewport
**Context:**
- Playwright Version: `1.20.0`
- Operating System: Mac
- Node.js version: `16.14.2`
- Browser: Chrome

System:
 - OS: macOS 12.0.1
 - Memory: 164.48 MB / 16.00 GB

Binaries:
 - Node: 16.14.2
 - npm: 8.5.5

**Describe the bug**

I have an app where the sidebar animates out of the viewport when it is collapsed using CSS, something similar to this:

```
.sidebar {
  position: fixed
  left: 0
  transition: all .5s
  transform: translateX(-100%) // moves the sidebar out of the viewport on the left
}

.sidebar-trigger:hover .sidebar {
  transform: translateX(0) // when the user hovers near the left edge, animate the sidebar in
}
```

I'm trying to test this behavior using Playwright. The test looks something like this:

```
await page.locator('.sidebar-trigger').hover()
await page.locator('.sidebar').waitFor({ state: 'visible' })

await page.locator('.something-else').hover() // sidebar looses hover and animates away
await page.locator('.sidebar').waitFor({ state: 'hidden' }) // this FAILS even though the sidebar is no longer visible
```

The `waitFor({ state: 'hidden' })` part always times out, even though I can clearly see that the sidebar animates out of the viewport and is no longer visible.

I'm very new to Playwright so I can't tell, is this a bug or by design? If the latter, how am I supposed to test this behavior properly?

Thanks!

rwoll:
Closes #13131.

Per the visibility spec on https://playwright.dev/docs/next/actionability#visible:

> Element is considered visible when it has non-empty bounding box and does not have visibility:hidden computed style. Note that elements of zero size or with display:none are not considered visible.

✅ non-empty bounding box
✅ does not have visibility:hidden

Given the above conditions are satisfied, the locator is considered visible.

https://github.com/microsoft/playwright/issues/8740 proposes something like `isInViewport()` that would be better suited for checking if an element is offscreen.

### DIFF
--- a/tests/page/page-wait-for-selector-2.spec.ts
+++ b/tests/page/page-wait-for-selector-2.spec.ts
@@ -100,6 +100,29 @@ it('hidden should wait for removal', async ({ page, server }) => {
   expect(divRemoved).toBe(true);
 });
 
+it('hidden should wait for out-of-viewport', async ({ page }) => {
+  it.fail(true, 'https://github.com/microsoft/playwright/issues/13131');
+  await page.setContent(`
+    <style>
+      .cover {
+        position: fixed;
+        left: 0;
+        top: 0;
+        width: 100px;
+        height: 100px;
+        background-color: red;
+        transform: translateX(-200px);
+      }
+    </style>
+    <div class="cover">cover</div>
+  `);
+
+  const cover = page.locator('.cover');
+  await expect.soft(cover).not.toBeVisible();
+  // Below is not web-first, but is included to mirror original snippet from https://github.com/microsoft/playwright/issues/13131
+  await cover.waitFor({ state: 'hidden' });
+});
+
 it('should return null if waiting to hide non-existing element', async ({ page, server }) => {
   const handle = await page.waitForSelector('non-existing', { state: 'hidden' });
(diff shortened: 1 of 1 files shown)


tests/page/page-wait-for-selector-2.spec.ts
```diff
@@ -100,6 +100,29 @@ it('hidden should wait for removal', async ({ page, server }) => {
...
+        transform: translateX(-200px);
+      }
+    </style>
+    <div class="cover">cover</div>
+  `);
+
+  const cover = page.locator('.cover');
+  await expect.soft(cover).not.toBeVisible();
+  // Below is not web-first, but is included to mirror original snippet from https://github.com/microsoft/playwright/issues/13131
+  await cover.waitFor({ state: 'hidden' });
```
yury-s, line 123: nit: reduce timeout or skip the test for now?

yury-s (approved):

Issue #13078, opened 2022-03-25
[BUG] Extension required when importing other files with type="module"


IanVS:
**Context:**

```md
## System:
 - OS: macOS 12.2.1
 - Memory: 802.27 MB / 32.00 GB
## Binaries:
 - Node: 16.13.1 - ~/.volta/tools/image/node/16.13.1/bin/node
 - Yarn: 1.22.17 - ~/.volta/bin/yarn
 - npm: 8.3.0 - ~/.volta/tools/image/npm/8.3.0/bin/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash
```

**Describe the bug**

I am using type="module" in my package.json, and running with `PW_EXPERIMENTAL_TS_ESM=1`.  I found that when I attempt to import another of my files from my global setup file, I need to specify the full filename with extension, which I think is a node esm requirement, but then typescript complains with `An import path cannot end with a '.ts' extension. Consider importing './generateTOTP' instead.`.  If I don't include the extension, playwright will not start and I get an error of:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/xxx/e2e/config/generateTOTP' imported from /xxx/e2e/config/global-setup.ts
```

dgozman:
@IanVS Are you using typescript for your config file, e.g. `playwright.config.ts`? Could you please share a full repro that we can run locally?

IanVS:
Sure, I'll put one together soon, thanks.

frederikhors:
On Windows too with 1.20.1. With 1.20.0 it works.

IanVS:
Here's a reproduction, @dgozman.  https://github.com/IanVS/playwright-issue-13078

pavelfeldman:
I can repro on 1.20, but I can't repro on 1.19. So it is a regression. Starting with 1.20 we require file extensions when importing Typescript files.

dgozman:
We discussed this issue and arrived to the following conclusion. We follow [TypeScript's experimental support for ESM](https://www.typescriptlang.org/docs/handbook/esm-node.html) in the test runner. TypeScript insists on using the extension of the output file for imports:
 ```ts
 // example.spec.ts
 import { foo } from './bar.js';
 ```
PR #13137 made sure that this syntax works. It is also supported by VSCode, so you should not get any errors when compiling or using an IDE.

We also allow using `import { foo } from './bar.ts'` because you are not usually compiling the tests and never actually see the `bar.js` file.

[...]

PR #13137, opened 2022-03-28
fix(esm): make sure import from './foo.js' is supported

dgozman:
Drive-by: migrate all `@esm` tests to `esm.spec.ts`.

References #13078.

### DIFF
--- a/packages/playwright-test/src/transform.ts
+++ b/packages/playwright-test/src/transform.ts
@@ -99,23 +99,31 @@ export function resolveHook(filename: string, specifier: string): string | undef
   if (!isTypeScript)
     return;
   const tsconfig = loadAndValidateTsconfigForFile(filename);
-  if (!tsconfig)
-    return;
-  for (const { key, values } of tsconfig.paths) {
-    const keyHasStar = key[key.length - 1] === '*';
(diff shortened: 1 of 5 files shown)


packages/playwright-test/src/transform.ts
```diff
@@ -99,25 +99,34 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
+        for (const ext of ['', '.js', '.ts', '.mjs', '.cjs', '.jsx', '.tsx']) {
+          if (fs.existsSync(candidate + ext))
+            return candidate;
+        }
       }
     }
   }
+  if (specifier.endsWith('.js')) {
+    const resolved = path.resolve(path.dirname(filename), specifier);
+    console.log(filename, specifier, resolved);
```
pavelfeldman, line 123: remove logging

pavelfeldman (approved):

frederikhors:
I'm having this issue using both `1.20.1` and `next`:

image

Only a VSCode issue. The tests are working.

If I use `./constants` (without `.ts`) I get error in pw at runtime: `[ERR_MODULE_NOT_FOUND]` using `PW_EXPERIMENTAL_TS_ESM=1 playwright test`.

How can I fix?

PR #13140, opened 2022-03-28
cherry-pick(#13138): fix(parallel): minimize the number of beforeAll/afterAll hooks in parallel mode

dgozman:
Previously, we always formed groups consisting of a single test.
Now, we group tests that share `beforeAll`/`afterAll` hooks into
`config.workers` equally-sized groups.


pavelfeldman:
Looks broken!

dgozman:
> Looks broken!

Indeed!

pavelfeldman (approved):

PR #13143, opened 2022-03-29
chore: migrate injected scripts to esbuild

pavelfeldman:


package.json
```diff
@@ -100,8 +97,6 @@
     "style-loader": "^3.3.1",
```
dgozman, line 97: Do we still need `style-loader`, `css-loader` and `file-loader`?

dgozman (approved):

PR #13164, opened 2022-03-29
feat(vrt): new option `"caret"` for taking screenshots

### LINKED ISSUE #12643: [Feature] Option to not automatically hide caret in screenshots

aslushnikov:
This has two values:
- `"hide"` to hide input caret for taking screenshot
- `"initial"` to keep caret behavior unchanged

Defaults to `"hide"`.

Fixes #12643


dgozman (approved):
Do we have to copy to `toHaveScreenshot` somewhere, or does it happen auto-magically?

PR #13166, opened 2022-03-29
chore(html-report): render metainfo as a chip

pavelfeldman:


packages/html-reporter/src/metadataView.tsx
```diff
@@ -0,0 +1,100 @@
...
+          icon='commit'
+        />
+      }
+      {(metadata['revision.author'] || metadata['revision.email']) &&
+        <MetadatViewItem
+          content={(
+            metadata['revision.author'] && metadata['revision.email']
+              ? <>{metadata['revision.author']} {metadata['revision.email']}</>
+              : (metadata['revision.author'] || metadata['revision.email'])
+            )!}
```
rwoll, line 56:
```suggestion
          content={`${metadata['revision.author'] ${metadata['revision.email']}`}
```

There's no more <br/> being conditionally inserted, so this can be simplified.

rwoll (approved):
Thanks! LGTM, modulo one minor simplification

packages/html-reporter/src/metadataView.tsx
pavelfeldman: Following up with you offline!

PR #13168, opened 2022-03-30
test(components): add more tests

pavelfeldman:


packages/playwright-ct-react/register.mjs
```diff
@@ -30,6 +30,10 @@ function render(component) {
     if (typeof child === 'string')
       return child;
     return render(child);
+  }).filter(child => {
```
dgozman, line 33: Is this react specific, or should we copy this to other ct packages?

packages/web/src/components/splitView.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+test('should hide sidebar', async ({ mount }) => {
+  const component = await mount(<SplitView sidebarSize={100} orientation={'horizontal'} sidebarHidden={true}>
+    <div id='main' style={{ border: '1px solid blue', flex: 'auto' }}>main</div>
+    <div id='sidebar' style={{ border: '1px solid red', flex: 'auto' }}>sidebar</div>
+  </SplitView>);
+  const mainBox = await component.locator('#main').boundingBox();
+  expect.soft(mainBox).toEqual({ x: 0, y: 0, width: 500, height: 500 });
+});
+
+test('drag resize', async ({ page, mount }) => {
```
dgozman, line 65: Nice one!!!

dgozman (approved):

packages/playwright-ct-react/register.mjs
pavelfeldman: Other frameworks use slots that operate fragments containing text nodes.

Issue #5905, opened 2021-03-22
[Feature] console events for electron app


sirtimid:
It would be great if we could use the the pluggable logger sink for Electron tests as well. :)

pavelfeldman:
We are actually thinking about going the other way around elsewhere. How do you use the sink? Does DEBUG= not work for you?

sirtimid:
Hey @pavelfeldman I would like to use logger in order to assert things within tests, so I would push the logs in an array and assert that the expected string exists in that array, The DEBUG flag won't help me with this

sirtimid:
Issue is that currently we can only get the logs only from renderer with the `console` event, I need a way to take the logs from main process as well, spectron has a `app.client.getMainProcessLogs()` method

pavelfeldman:
@sirtimid: would it work if you had something like [page.on('console')](https://playwright.dev/docs/api/class-page#pageonconsole) on the electronApp object?

sirtimid:
@pavelfeldman yeah, for sure, it would be even better TBH!

rssk:
Is there any way to get at electron main logs from the launched process? It seems like otherwise it'd be hard to debug fails on headless tests

imolorhe:
I also need this API to retrieve logs from the main process, for the same reasons as @sirtimid. Are there any timelines for this?

cameron-martin:
It does seem to [collect the stdout and stderr](https://github.com/microsoft/playwright/blob/807f70bccf975a007f68a3faf7e4a949d58b1ab7/packages/playwright-core/src/utils/processLauncher.ts#L102-L110) of the launched electron process, but I can't work out how to access them via the public APIs.

Prinzhorn:
> It does seem to [collect the stdout and stderr](https://github.com/microsoft/playwright/blob/807f70bccf975a007f68a3faf7e4a949d58b1ab7/packages/playwright-core/src/utils/processLauncher.ts#L102-L110) of the launched electron process, but I can't work out how to access them via the public APIs.

To safe anyone from going through the same rabbit hole I just did, this works with the latest playwright 1.24.1:

```js
const app = await electron.launch({ args: ['main.js'] });

app.process().stdout.on('data', (data) => console.log(`stdout: ${data}`));
app.process().stderr.on('data', (error) => console.log`stderr: ${error}`));
```

From https://github.com/microsoft/playwright/issues/13267#issuecomment-1086786960 and https://github.com/microsoft/playwright/pull/13280

[...]

PR #13192, opened 2022-03-30
fix: add stdio option to make it possible to debug electron applications

SimonSiefke:
Currently it is very hard to debug when an electron test fails because no stdout or stderr is printed (see also https://github.com/microsoft/playwright/issues/5905#issuecomment-966598268)

This adds an stdio option which prints to stdout and stderr just like a normal electron app would. 

## before
it is very unclear why the test is failing:
before

## after
it is very clear why and where the test is failing:
after

### DIFF
--- a/docs/src/api/class-electron.md
+++ b/docs/src/api/class-electron.md
@@ -77,6 +77,11 @@ Current working directory to launch application from.
 
 Specifies environment variables that will be visible to Electron. Defaults to `process.env`.
 
+### option: Electron.launch.stdio
+- `stdio` <[string]>
+
+Specifies where output is printed to.
+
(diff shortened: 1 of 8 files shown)


pavelfeldman:
Do you think this will address your use case: https://github.com/microsoft/playwright/commit/8232497c884e16d8846c6a8585c27691969c7739?

Also, you can always set env variable DEBUG=pw:browser and it'll log electron process output.

SimonSiefke:
Yes. I think https://github.com/microsoft/playwright/issues/13267#issuecomment-1086786960 would work really well. Thanks for your comment and for implementing that!

PR #13194, opened 2022-03-30
chore: support msys2

liuxingbaoyu:


aslushnikov:
@liuxingbaoyu thanks! How did you end up here? Do you build with msys?

aslushnikov (approved):

PR #13196, opened 2022-03-30
chore: warn against using globally scoped install

### LINKED ISSUE #13188: [BUG] npx playwright install does not install appropriate browsers
**Context:**
- Playwright Version: 1.19.2
- Operating System: ubuntu-latest, windows-latest
- Node.js version: 12,14,16,17
- Browser:  Webkit, Chrome
- Extra: Github Workflows

**Code Snippet**

https://github.com/nasa/openmct/blob/ae553e2b82eea88e26c78a23abc8664f2a412e95/.github/workflows/e2e-pr.yml

```yaml
     - uses: actions/checkout@v3
      - uses: actions/setup-node@v3
        with:
          node-version: '16'
          check-latest: true
      - name: Cache node modules
        uses: actions/cache@v3
        env:
          cache-name: cache-node-modules
        with:
          path: ~/.npm
          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package.json') }}
          restore-keys: |
            ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package.json') }}
      - run: npx playwright install
      - run: npm install
      - run: npm run test:e2e:full
```

**Describe the bug**

`npx playwright install` Does not appear to download necessary browsers if using an older playwright/test version.

```
browserType.launch: Executable doesn't exist at /home/runner/.cache/ms-playwright/chromium-965416/chrome-linux/chrome
    ╔═════════════════════════════════════════════════════════════════════════╗
    ║ Looks like Playwright Test or Playwright was just installed or updated. ║
    ║ Please run the following command to download new browsers:              ║
    ║                                                                         ║
    ║     npx playwright install                                              ║
    ║                                                                         ║
    ║ <3 Playwright Team                                                      ║
    ╚═════════════════════════════════════════════════════════════════════════╝
```

Note that specifying npx playwrigh@1.19.2 works.

rwoll:
Resolves #13188

The message looks like this:

```
╔═══════════════════════════════════════════════════════════════════════════════╗
║ WARNING: It looks like you are running 'npx playwright install' without first ║
║ installing your project's dependencies.                                       ║
║                                                                               ║
║ To avoid unexpected behavior, please install your dependencies first, and     ║
║ then run Playwright's install command:                                        ║
║                                                                               ║
║     npm install                                                               ║
║     npx playwright install                                                    ║
║                                                                               ║
║ If your project does not yet depend on Playwright, first install the          ║
║ applicable npm package (most commonly @playwright/test), and                  ║
║ then run Playwright's install command to download the browsers:               ║
║                                                                               ║
║     npm install @playwright/test                                              ║
║     npx playwright install                                                    ║
║                                                                               ║
║ <3 Playwright Team                                                            ║
╚═══════════════════════════════════════════════════════════════════════════════╝
```

The cross-platform, cross-node version output of https://github.com/rwoll/where-am-i-executing/runs/5795454928
was used to determine the isLikelyNpmGlobal heuristic.


packages/playwright-core/src/utils/registry.ts
```diff
@@ -738,7 +739,7 @@ export function buildPlaywrightCLICommand(sdkLanguage: string, parameters: strin
     case 'csharp':
       return `pwsh bin\\Debug\\netX\\playwright.ps1 ${parameters}`;
     default:
-      return `npx playwright ${parameters}`;
+      return `npx playwright@${packageJSON.version} ${parameters}`;
```
rwoll, line 742: The version is not always necessary, but depending on where you call this from (and when) it can be.

aslushnikov: use `getPlaywrightVersion` instead!

rwoll: Updated! Let me know if it's good to merge!

mxschmitt:
I'm not sure about this change since "npx playwright install" is primarily for the Playwright Test use-case where no browsers are installed by default.

If a user invokes it before the deps are installed, it would install the playwright npm module, which already would install all browsers by default, so the CLI use-case to install a single browser is meaningless.

Maybe we should instead check if the CLI argv[0] is inside the npx tmp direcory, and if so yell out "that the dependencies need to be installed before (npm ci/npm install etc.)", otherwise it ends up in a version conflict and unwanted behavior (all browsers are getting installed).

rwoll:
> npx tmp direcory, and if so yell out "that the dependencies need to be installed before (npm ci/npm install etc.)", otherwise it ends up in a version conflict and unwanted behavior (all browsers are getting installed).

I like this idea, and since it aligns with the majority of the use case, it should be the route we take.

There are some rare circumstances where you want do want to run `npx playwright@pinned install` before `npm i` for Docker-related workflows, but I think these are rare enough to not warrant potentially creating version confusion that this current patch is liable to make.

packages/playwright-core/src/utils/registry.ts
```diff
@@ -738,7 +738,7 @@ export function buildPlaywrightCLICommand(sdkLanguage: string, parameters: strin
     case 'csharp':
       return `pwsh bin\\Debug\\netX\\playwright.ps1 ${parameters}`;
     default:
-      return `npx playwright ${parameters}`;
+      return `npx playwright@${getPlaywrightVersion()} ${parameters}`;
```
rwoll, line 741: Don't do this. If we detect running `npx playwright install` from non-local, print a nice error message (but do not crash), that warns the user and includes a hit at the more future-proof way.

rwoll:
The error message could use some wording help! ❤️

packages/playwright-core/src/cli/cli.ts
```diff
@@ -115,6 +115,27 @@ program
...
+          ``,
+          `To avoid unexpected behavior, please install your dependencies first, and`,
+          `then run Playwright's install command:`,
+          ``,
+          `    npm install`,
+          `    npx playwright install`,
+          ``,
+          `If your project does not yet depend on Playwright, first install the`,
+          `applicable npm package (most commonly @playwright/test), and`,
+          `then run Playwright's install command to download the browsers:`,
```
rwoll, line 132: This case is confusing or just plain wrong. It only applies for @playwright/test. If all they are after is `playwright` package, that will install the browsers for them, the the `npx playwright install` command is not needed and confusing.

aslushnikov:
```
ATTENTION: running `npx playwright install` using global installation!

If you have Playwright in your `package.json`, then this command might use 
a different Playwright version and might not install browsers that your
package require.

To use Playwright version your package depends on, make sure to run this
command after installing your project dependencies:

  npm install
  npx playwright install

Read more about NPX: https://nodejs.dev/learn/the-npx-nodejs-package-runner

<3 Playwright Team          
```

pavelfeldman:
I like the wording in the PR description

pavelfeldman (approved):

PR #13197, opened 2022-03-31
chore: prepare image diff for refactoring

pavelfeldman:


packages/html-reporter/src/testResultView.tsx
```diff
@@ -26,57 +26,45 @@ import { AttachmentLink } from './links';
...
-  const traces = attachments.filter(a => a.name === 'trace');
-
-  const otherAttachments = new Set<TestAttachment>(attachments);
-  [...screenshots, ...videos, ...traces].forEach(a => otherAttachments.delete(a));
-
-  const snapshotNameToDiffTabs = new Map<string, DiffTab[]>();
-  let tabId = 0;
-  for (const attachment of attachments) {
-    const match = attachment.name.match(/^(.*)-(\w+)(\.[^.]+)?$/);
+function groupImageDiffs(screenshots: Set<TestAttachment>): ImageDiff[] {
```
aslushnikov, line 36: Maybe rename since `screenshots` is actually an out parameter?

```suggestion
function extractImageDiffs(screenshots: Set<TestAttachment>): ImageDiff[] {
```

aslushnikov (approved):

packages/html-reporter/src/testResultView.tsx
pavelfeldman: will follow up!

PR #13199, opened 2022-03-31
chore: allow matchers decorate step title

pavelfeldman:


packages/playwright-test/src/dispatcher.ts
```diff
@@ -262,6 +262,8 @@ export class Dispatcher {
         this._reporter.onStdErr?.('Internal error: step end without step begin: ' + params.stepId, data.test, result);
         return;
       }
+      if (params.refinedTitle)
```
dgozman, line 265: Let's add a comment that `onStepBegin` and `onStepEnd` will get different titles, and we are fine with that.

dgozman (approved):

packages/playwright-test/src/dispatcher.ts
pavelfeldman: I am not sure we should do this, let's wait for it to be reported as a concern.

PR #1623, opened 2020-04-01
docs(api): update ordering for $eval and $$eval

arjunattam:
...to be consistent with how we order `$` and `$$` - the inconsistency has been bothering me. However, this fails on doc lint. I could update the linter along with this.

```
DocLint Failures:
  1) [MarkDown] Bad alphabetic ordering of Page members: Page.$eval() should go after Page.$$eval()
  2) [MarkDown] Bad alphabetic ordering of Frame members: Frame.$eval() should go after Frame.$$eval()
  3) [MarkDown] Bad alphabetic ordering of ElementHandle members: ElementHandle.$eval() should go after ElementHandle.$$eval()
```

PR #13223, opened 2022-03-31
browser(webkit): rebase to 03/23/22 (r291756)

dpino:
Rebase `webkit/patches/boostrap.diff` to [r291756](https://trac.webkit.org/changeset/291756/webkit)

Changes:

* [chore(webkit): bootstrap build #1623](https://github.com/dpino/WebKit/commit/e2c2e508c1ae38ecd29684da200d661f0e126f9d)
* [Resolve conflicts](https://github.com/dpino/WebKit/commit/3a9f8c19ed0bb16d9e851a5ac4425d9ba01f6456)
* [[GLIB] Build error in GraphicsContextState.cpp (error: no match for 'operator!=') in Ubuntu LTS after r291696](https://github.com/dpino/WebKit/commit/e4445f38ce3b6bf7648fe9e300a536de843397f2). Cherry-pick [r291970](https://commits.webkit.org/r291970).
* [Add mising header](https://github.com/dpino/WebKit/commit/6fb76eec7af0654932f99fe1b7bed34c03d8d221): Add missing header to fix build error. This change should go upstream.
* [Work around function 'gst_video_format_info_component', available since >=GStreamer 1.18](https://github.com/dpino/WebKit/commit/f5a6bd9443aca7827e282cf6e4c1cfec2a4c9178): [r291343](https://trac.webkit.org/changeset/291343/webkit) introduces new code using function `gst_video_format_info_component`, only available since >=GStreamer 1.18. The patch replaces that function for a local alternative implementation called `webkitGstVideoFormatInfoComponent`. This change should be reverted once we use the new bundle system.
* [Revert '[GTK][WPE] Remove the ATK implementation](https://github.com/dpino/WebKit/commit/05fd3bbb031ad650a43b9051415329a22d1314d2): This patch re-introduces the ATK code path, which was removed in [r291613](https://commits.webkit.org/r291613). This change should be reverted once we use the new bundle system.
* [Build using ATK](https://github.com/dpino/WebKit/commit/d92d9c1e245af8fec8152a366d7fa9a12bea44d2): Ubuntu 18.04 cannot build using ATSPI and still requires ATK. This change should be reverted once we use the new bundle system.


yury-s:
> [Work around function 'gst_video_format_info_component', available since >=GStreamer 1.18](https://github.com/dpino/WebKit/commit/f5a6bd9443aca7827e282cf6e4c1cfec2a4c9178): [r291343](https://trac.webkit.org/changeset/291343/webkit) introduces new code using function gst_video_format_info_component, only available since >=GStreamer 1.18. The patch replaces that function for a local alternative implementation called webkitGstVideoFormatInfoComponent. This change should be reverted once we use the new bundle system.

Is it possible to include gstreamer into jhbundle instead? Is this an issue only for ubuntu 18.04?

> [Revert '[GTK][WPE] Remove the ATK implementation](https://github.com/dpino/WebKit/commit/05fd3bbb031ad650a43b9051415329a22d1314d2): This patch re-introduces the ATK code path, which was removed in [r291613](https://commits.webkit.org/r291613). This change should be reverted once we use the new bundle system.
[Build using ATK](https://github.com/dpino/WebKit/commit/d92d9c1e245af8fec8152a366d7fa9a12bea44d2): Ubuntu 18.04 cannot build using ATSPI and still requires ATK. This change should be reverted once we use the new bundle system.

This moves us away from the upstream and modifies the core login which we try to avoid as much as possible (we try to add only missing instrumentation bits in the fork). Can we include the new dependency into jhbuild? Alternatively, what will break if we build with USE_ATK=0?

dpino:
I addressed the comments from the last review.

Instead of re-introduced the ATK code path I added ATSPI, and other required libraries, as a dependency to the JHBuild module. This changes goes as a separated patch in `browser_webkit/webkit/patches`. Using ATSPI in Ubuntu 18.04 also requires cherry-picking a patch that eventually didn't go upstream ([webkit.org/b/236237](https://github.com/microsoft/playwright/pull/13223/commits/9cd91710133db3d9ebf2a5c33e70fb8212de1778)). This change also goes as a separacted patch in `browser_webkit/webkit/patches`.

As for the GStreamer change, we want this change to go upstream. The upstream patch is at [webkit.org/b/238685](https://github.com/microsoft/playwright/pull/13223/commits/c5b8ab68dc945edae6ccb4cf21f5a3ddd0b016e1). I pulled this patch and also stored as a separated patch in `browser_webkit/webkit/patches`.

So basically that go in `bootstrap.diff` are the following:
* [chore(webkit): bootstrap build #1623](https://github.com/dpino/WebKit/commit/e2c2e508c1ae38ecd29684da200d661f0e126f9d)
* [Resolve conflicts](https://github.com/dpino/WebKit/commit/3a9f8c19ed0bb16d9e851a5ac4425d9ba01f6456)
* [[GLIB] Build error in GraphicsContextState.cpp (error: no match for 'operator!=') in Ubuntu LTS after r291696](https://github.com/dpino/WebKit/commit/e4445f38ce3b6bf7648fe9e300a536de843397f2). Cherry-pick [r291970](https://commits.webkit.org/r291970).
* [Add mising header](https://github.com/dpino/WebKit/commit/6fb76eec7af0654932f99fe1b7bed34c03d8d221): Add missing header to fix build error. This change has already landed as part of https://trac.webkit.org/changeset/292244/webkit.

yury-s:
> Instead of re-introduced the ATK code path I added ATSPI, and other required libraries, as a dependency to the JHBuild module.

Good, will we be able to upstream those dependency changes? Ideally, we'd have jhbuild mode which allows building on Ubuntu 18 without applying changes to the core logic. My understanding is that on Ubuntu 18.04 there are way more old dependencies that need to be pulled in by jhbuild compared to Ubuntu 20.04, this is why I'm saying it's better to keep track of them separately from ubuntu 20.

> This changes goes as a separated patch in browser_webkit/webkit/patches.

We discussed in the meeting today approach with separate patches and decided that we want to keep single diff file for 2 reasons:
* Rolling new WebKit version would require updating N patch files separately. Folks on our team tried such approach in another project and it proved to have too much overhead compared to single patch.
* We don't want to have non-trivial patches to the core logic except for missing instrumentation. So basically we don't want our port to ever behave differently compared to upstream. By trivial changes I mean build fixes (mostly unified build related), sometimes adding stub implementations (empty platform clients) etc.

> Using ATSPI in Ubuntu 18.04 also requires cherry-picking a patch that eventually didn't go upstream ([webkit.org/b/236237](https://github.com/microsoft/playwright/pull/13223/commits/9cd91710133db3d9ebf2a5c33e70fb8212de1778)). This change also goes as a separacted patch in browser_webkit/webkit/patches.

Can this be fixed by including newer gbus stuff into jhbuild on ubuntu 18.04? Alternatively I'd like to explore compiling with both ATSPI and ATK disabled but instead stubbing and those platforms hooks. AFAIU since we only need the accessibility stuff in our port to work for the automation clients, we should be able to stub the platform hooks at the expense of  possibly breaking actual behavior for human interaction. I'll try to do this today and see if that works.

> As for the GStreamer change, we want this change to go upstream. The upstream patch is at [webkit.org/b/238685](https://github.com/microsoft/playwright/pull/13223/commits/c5b8ab68dc945edae6ccb4cf21f5a3ddd0b016e1). I pulled this patch and also stored as a separated patch in browser_webkit/webkit/patches.

Sounds good, let's merge it into the bootstrap.diff per the above comment.

yury-s:
> I'll try to do this today and see if that works.

How about doing something like this https://github.com/yury-s/WebKit/commit/ce7dad80af05a4187f356d819d1d5d482c1826bf where the platform hooks are stubbed to allow compiling with ENABLE(ACCESSIBILITY) but !USE(ATK) and !USE(ATSPI) ? Our existing tests pass with that and it should eliminate required platform dependencies that are causing troubles. I believe it should also be possible to land this upstream as there is a clear automation use case where this is used. WDYT?

dpino:
> > Instead of re-introduced the ATK code path I added ATSPI, and other required libraries, as a dependency to the JHBuild module.
> 
> Good, will we be able to upstream those dependency changes? Ideally, we'd have jhbuild mode which allows building on Ubuntu 18 without applying changes to the core logic. My understanding is that on Ubuntu 18.04 there are way more old dependencies that need to be pulled in by jhbuild compared to Ubuntu 20.04, this is why I'm saying it's better to keep track of them separately from ubuntu 20.
> 

Yes, there are more required dependencies in Ubuntu 18.04 than Ubuntu 20.04. JHBuild only pulls dependencies if they cannot be satisfied via system libraries. So, for instance, in the case o thef ATSPI/glib dependencies this PR adds to `jhbuild.minimal`, they do not affect Ubuntu 20.04 because Ubuntu 20.04 can satisfy these dependencies system libraries. If I recall correctly, at this moment the number of pulled dependencies for Ubuntu 20.04 is 9 whereas for Ubuntu 18.04 is 13. We have never had different jhbuild module sets for different systems (although is possible).

> We discussed in the meeting today approach with separate patches and decided that we want to keep single diff file for 2 reasons:
> 
> * Rolling new WebKit version would require updating N patch files separately. Folks on our team tried such approach in another project and it proved to have too much overhead compared to single patch.
> * We don't want to have non-trivial patches to the core logic except for missing instrumentation. So basically we don't want our port to ever behave differently compared to upstream. By trivial changes I mean build fixes (mostly unified build related), sometimes adding stub implementations (empty platform clients) etc.
> 

Ok, no problem. I can merge all the patches into `bootstrap.diff`.

> > Using ATSPI in Ubuntu 18.04 also requires cherry-picking a patch that eventually didn't go upstream ([webkit.org/b/236237](https://github.com/microsoft/playwright/pull/13223/commits/9cd91710133db3d9ebf2a5c33e70fb8212de1778)). This change also goes as a separacted patch in browser_webkit/webkit/patches.
> 
> Can this be fixed by including newer gbus stuff into jhbuild on ubuntu 18.04? Alternatively I'd like to explore compiling with both ATSPI and ATK disabled but instead stubbing and those platforms hooks. AFAIU since we only need the accessibility stuff in our port to work for the automation clients, we should be able to stub the platform hooks at the expense of possibly breaking actual behavior for human interaction. I'll try to do this today and see if that works.
> 

Ok, providing a newer version of `gdbus-codegen` is another way to go. 

> Sounds good, let's merge it into the bootstrap.diff per the above comment.

Ok.

dpino:
> How about doing something like this [yury-s/WebKit@ce7dad8](https://github.com/yury-s/WebKit/commit/ce7dad80af05a4187f356d819d1d5d482c1826bf) where the platform hooks are stubbed to allow compiling with ENABLE(ACCESSIBILITY) but !USE(ATK) and !USE(ATSPI) ? Our existing tests pass with that and it should eliminate required platform dependencies that are causing troubles. I believe it should also be possible to land this upstream as there is a clear automation use case where this is used. I'm sure the Source/cmake/Options{GTK,WPE}.cmake could be done in a more elegant way and ideally just USE_ATK/ATSPI configured in our build scripts, I hope you can help straighten that part.

This looks interesting, I can discuss it with Carlos Garcia who is the main developer working on A11y in WebKitGTK/WPE and try to merge it upstream.

dpino:
I squashed the relevant individual patches into `bootstrap.diff`. I also cherry-picked your patch to build without ATSPI.

Changes:

  - [chore(webkit): bootstrap build #1623](https://github.com/dpino/WebKit/commit/5d3198ca90d0781f99ae2952526f2967e091dea6)
  - [Resolve conflicts](https://github.com/dpino/WebKit/commit/a70b1474101d8bf964ca244e39a25e9955d4396a)
  - [[GLIB] Fix build error 'not match operator !=' in Ubuntu LTS after...](https://github.com/dpino/WebKit/commit/465b6175154613e145f6805556a750be8727b3d9)
  - [[GStreamer] gst_video_format_info_component not defined in GStreamer...](https://github.com/dpino/WebKit/commit/640857c46f196f6638bc9016231e2b41f2d60968)
  - [Compile accessibility without ATK/ATSPI](https://github.com/dpino/WebKit/commit/ca170bc385165b8eebf003741d0e54ca000b2352)

yury-s (approved):

PR #13238, opened 2022-04-01
update cookie type definitions for `addCookies`

### LINKED ISSUE #13237: [Feature] improved typescript definitions for `BrowserContext.addCookies`
[`BrowserContext.addCookies`][1] does not have a typed parameter for what it accepts as a cookie.

The existing `Cookie` definition requires more fields than necessary:
* path
* httpOnly
* expires
* secure
* sameSite

It would be helpful to introduce a definition to more closely match `addCookies` as well as a type for `sameSite`

Current definition is: https://github.com/microsoft/playwright/blob/7d7fe3c618464fd6d9d01afd4e679ba336ddd1ac/packages/playwright-core/types/types.d.ts#L6515

```typescript
interface {
    name: string;

    value: string;

    /**
     * either url or domain / path are required. Optional.
     */
    url?: string;

    /**
     * either url or domain / path are required Optional.
     */
    domain?: string;

    /**
     * either url or domain / path are required Optional.
     */
    path?: string;

    /**
     * Unix time in seconds. Optional.
     */
    expires?: number;

    /**
     * Optional.
     */
    httpOnly?: boolean;

    /**
     * Optional.
     */
    secure?: boolean;

    /**
     * Optional.
     */
    sameSite?: "Strict"|"Lax"|"None";
  }
```

[1]: https://github.com/microsoft/playwright/blob/7d7fe3c618464fd6d9d01afd4e679ba336ddd1ac/packages/playwright-core/types/types.d.ts#L6515

davidbielik:
Fixes https://github.com/microsoft/playwright/issues/13237

* added `NewCookie` to replace inline type
* added `CookieSameSite` to replace inline type

### DIFF
--- a/packages/playwright-core/types/types.d.ts
+++ b/packages/playwright-core/types/types.d.ts
@@ -6512,46 +6512,7 @@ export interface BrowserContext {
    *
    * @param cookies
    */
-  addCookies(cookies: Array<{
-    name: string;
-
-    value: string;
-
-    /**
-     * either url or domain / path are required. Optional.
-     */
-    url?: string;
-
-    /**
-     * either url or domain / path are required Optional.
-     */
-    domain?: string;
-
-    /**
-     * either url or domain / path are required Optional.
-     */
-    path?: string;
-
-    /**
-     * Unix time in seconds. Optional.
-     */
-    expires?: number;
-
-    /**
-     * Optional.
-     */
-    httpOnly?: boolean;
-
-    /**
-     * Optional.
-     */
-    secure?: boolean;
-
-    /**
-     * Optional.
-     */
-    sameSite?: "Strict"|"Lax"|"None";
-  }>): Promise<void>;
+  addCookies(cookies: Array<NewCookie>): Promise<void>;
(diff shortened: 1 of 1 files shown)


davidbielik:
I see `SetNetworkCookieParam` also exists within '@playwright-core/client/network'. is there a better way to re-use that instead?

packages/playwright-core/types/types.d.ts
```diff
@@ -6512,46 +6512,7 @@ export interface BrowserContext {
...
-     * Optional.
-     */
-    secure?: boolean;
-
-    /**
-     * Optional.
-     */
-    sameSite?: "Strict"|"Lax"|"None";
-  }>): Promise<void>;
+  addCookies(cookies: Array<NewCookie>): Promise<void>;
```
pavelfeldman, line 6515: This file is generated, so you should not edit it.

davidbielik: I'll close this out for now then and wait to hear on #13237

PR #13248, opened 2022-04-01
Android socket name

kaivean:
### Why
Some apps includes custom chromium version and socket names. Can not get webviews only by `pid`. eg:
```
0000000000000000: 00000002 00000000 00010000 0001 01 17073592 @webview_devtools_remote_zeus
0000000000000000: 00000002 00000000 00010000 0001 01 17123419 @webview_devtools_remote_32327_zeus
0000000000000000: 00000002 00000000 00010000 0001 01 17194878 @chrome_devtools_remote
0000000000000000: 00000002 00000000 00010000 0001 01 22537480 @content_shell_devtools_remote
```
Chromedriver also supports `androidDeviceSocket` param to set custom socket name.

### How
```
const webview = await device.webView({ socketName: 'chrome_devtools_remote' });
```


docs/src/api/class-androidwebview.md
```diff
@@ -21,3 +21,8 @@ WebView process PID.
 - returns: <[string]>
 
 WebView package identifier.
+
+## method: AndroidWebView.socketName
```
pavelfeldman, line 25: Let's not expose it for now.

packages/playwright-core/src/client/android.ts
```diff
@@ -101,13 +101,26 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel> i
...
-  async webView(selector: { pkg: string }, options?: types.TimeoutOptions): Promise<AndroidWebView> {
-    const webView = [...this._webViews.values()].find(v => v.pkg() === selector.pkg);
+  async webView(selector: { pkg?: string; socketName?: string; }, options?: types.TimeoutOptions): Promise<AndroidWebView> {
+    const webView = [...this._webViews.values()].find(v => {
+      if (selector.pkg)
+        return v.pkg() === selector.pkg;
+      if (selector.socketName)
+        return v.socketName() === selector.socketName;
+      return false;
+    });
```
pavelfeldman, line 111:
```js
const predicate = v => {
  if (selector.pkg)
    return v.pkg() === selector.pkg;
  if (selector.socketName)
    return v.socketName() === selector.socketName;
  return false;
};
const webView = [...this._webViews.values()].find(predicate);
...
return this.waitForEvent('webview', { ...options, predicate });
```

packages/playwright-core/src/server/android/android.ts
```diff
@@ -238,20 +238,19 @@ export class AndroidDevice extends SdkObject {
   async launchBrowser(pkg: string = 'com.android.chrome', options: types.BrowserContextOptions): Promise<BrowserContext> {
     debug('pw:android')('Force-stopping', pkg);
     await this._backend.runCommand(`shell:am force-stop ${pkg}`);
-
-    const socketName = 'playwright-' + createGuid();
+    const socketName = isUnderTest() ? 'chrome_devtools_remote' : ('playwright-' + createGuid());
```
pavelfeldman, line 241: 'chrome_devtools_remote' will clash with Chrome, let's get a unique prefix for Playwright test.

packages/playwright-core/src/server/android/android.ts
```diff
@@ -336,42 +335,59 @@ export class AndroidDevice extends SdkObject {
   }
 
   private async _refreshWebViews() {
-    const sockets = (await this._backend.runCommand(`shell:cat /proc/net/unix | grep webview_devtools_remote`)).toString().split('\n');
+    // possible socketName, eg: webview_devtools_remote_32327, webview_devtools_remote_32327_zeus, chrome_devtools_remote, content_shell_devtools_remote
```
pavelfeldman, line 338: `chrome_devtools_remote` and `content_shell_devtools_remote` aren't web views though, why do we want to include them?

kaivean: I just expect to support more socketNames. 

Additionally. I cannot find  chromium webview app (package: org.chromium.webview_shell). I downloaded the app latest version from https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Android/ .  But it't chromium version is 77 which is not supported version  so that I cannot write and run test file.

docs/src/api/class-androidwebview.md
kaivean: OK. But check error: `Missing documentation for "AndroidWebView.socketName"`. How to handle it

packages/playwright-core/src/server/android/android.ts
pavelfeldman: > I just expect to support more socketNames.

But where would they be coming from? It'd be nice to understand what you are trying to achieve.

kaivean: > But where would they be coming from? It'd be nice to understand what you are trying to achieve.

Sorry, I mean I just find the old version of chromium webview app, it is not work , it cannot run test files . So I use android chrome to test webview.

kaivean: > `chrome_devtools_remote` and `content_shell_devtools_remote` aren't web views though, why do we want to include them?

Yes, I agree with you. I already add the "webview" prefix. Please code review.

docs/src/api/class-androidwebview.md
kaivean: How to handle the lint error

pavelfeldman: Call is `_socketName`

packages/playwright-core/src/server/android/android.ts
```diff
@@ -238,20 +238,19 @@ export class AndroidDevice extends SdkObject {
...
   }
 
-  async connectToWebView(pid: number): Promise<BrowserContext> {
-    const webView = this._webViews.get(pid);
+  async connectToWebView(socketName: string): Promise<BrowserContext> {
+    const webView = this._webViews.get(socketName);
     if (!webView)
       throw new Error('WebView has been closed');
-    return await this._connectToBrowser(`webview_devtools_remote_${pid}`);
+    return await this._connectToBrowser(socketName, { hasTouch: true });
```
pavelfeldman, line 253: Let's sort `{ hasTouch: true }` out in a different patch.

packages/playwright-core/src/server/android/android.ts
```diff
@@ -336,42 +335,58 @@ export class AndroidDevice extends SdkObject {
...
-        const p = match[1];
-        if (+p !== pid)
+        const webView = { pid: -1, pkg: '', socketName };
+        this._webViews.set(socketName, webView);
+        this.emit(AndroidDevice.Events.WebViewAdded, webView);
+        continue;
+      }
+
+      const pid = +match[1];
+      if (!newPids.has(pid)) {
```
pavelfeldman, line 366: Now that your map is sourceName => webView the `newPids` map is no longer relevant. Everything should operate sockets instead. So I would:
- extract a function that computes pkg from pid and move it below in this file
- list all sockets, manage Added/Removed events based on sockets. you'll get `newSockets` map.
- optionally resolve pkgs for those that end with `_\d+` so that user could fitler by pkg.

should be less code and much easier to follow.

kaivean: I removed newPids map. Now everything would operate sockets.  Please code review.

packages/playwright-core/src/server/android/android.ts
```diff
@@ -336,22 +335,35 @@ export class AndroidDevice extends SdkObject {
...
     const sockets = (await this._backend.runCommand(`shell:cat /proc/net/unix | grep webview_devtools_remote`)).toString().split('\n');
     if (this._isClosed)
       return;
 
-    const newPids = new Set<number>();
+    const socketNames = new Set<string>();
     for (const line of sockets) {
-      const match = line.match(/[^@]+@webview_devtools_remote_(\d+)/);
-      if (!match)
+      const match = line.match(/[^@]+@.*?webview_devtools_remote_?(\d*)/);
```
pavelfeldman, line 345: nit: you only need it on line 359.

packages/playwright-core/src/server/android/android.ts
```diff
@@ -336,22 +335,35 @@ export class AndroidDevice extends SdkObject {
...
+      socketNames.add(socketName);
+
+      if (this._webViews.has(socketName))
         continue;
 
+      // possible line: 0000000000000000: 00000002 00000000 00010000 0001 01 5841881 @webview_devtools_remote_zeus
+      // the result: match[1] = ''
+      if (!match || !match[1]) {
+        const webView = { pid: -1, pkg: '', socketName };
+        this._webViews.set(socketName, webView);
```
pavelfeldman, line 361: It'd be really nice if we could only do it in one place. Extracting a method that computes pkg would really help.

pavelfeldman (commented):
As per my last review, lets extract the function that computes pkg to keep webview bookkeeping in one place. Otherwise looks great!

packages/playwright-core/src/server/android/android.ts
kaivean: > As per my last review, lets extract the function that computes pkg to keep webview bookkeeping in one place. Otherwise looks great!

Thank you for code review. I agree with you. I extracted a function as a class method because it use `this.backend`.

pavelfeldman (approved):

pavelfeldman:
Looks good, but it looks like you have a conflict that you need to resolve.

mxschmitt:
Thank you @kaivean!

kaivean:
> Thank you @kaivean!

Thank you for merging the pr

PR #13249, opened 2022-04-01
feat(android): add androidDevice.options.omitDriverInstall

kaivean:
## Control installing drivers
Drivers are installed on every execution. But some phones must input password to install apps and can not be bypassed.
So Let user install driver for the first time ,  and providing  ` omitDriverInstall: true` avoid installing drivers
```
const [device] = await _android.devices({
      omitDriverInstall: true
});
```

### DIFF
--- a/docs/src/api/class-android.md
+++ b/docs/src/api/class-android.md
@@ -87,6 +87,11 @@ Returns the list of detected Android devices.
 
 Optional port to establish ADB server connection.
 
+### option: Android.devices.omitDriverInstall
+- `omitDriverInstall` <[boolean]>
+
+Optional install latest playwright driver automatically. if true, firstly you should install android-driver.apk and android-driver-target.apk which are in bin directory of playwright-core.  Default to false.
+
 ## method: Android.setDefaultTimeout
 
 This setting will change the default maximum time for all the methods accepting [`param: timeout`] option.
--- a/docs/src/api/class-android.md
+++ b/docs/src/api/class-android.md
@@ -90,7 +90,7 @@ Optional port to establish ADB server connection.
 ### option: Android.devices.omitDriverInstall
 - `omitDriverInstall` <[boolean]>
 
-Optional install latest playwright driver automatically. if true, firstly you should install android-driver.apk and android-driver-target.apk which are in bin directory of playwright-core.  Default to false.
+Prevents automatic playwright driver installation on attach. Assumes that the drivers have been installed already.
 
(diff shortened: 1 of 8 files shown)


docs/src/api/class-android.md
```diff
@@ -87,6 +87,11 @@ Returns the list of detected Android devices.
 
 Optional port to establish ADB server connection.
 
+### option: Android.devices.omitDriverInstall
+- `omitDriverInstall` <[boolean]>
+
+Optional install latest playwright driver automatically. if true, firstly you should install android-driver.apk and android-driver-target.apk which are in bin directory of playwright-core.  Default to false.
```
pavelfeldman, line 93:
```suggestion
Prevents automatic playwright driver installation on attach. Assumes that the drivers have been installed already.
```

packages/playwright-core/src/server/android/android.ts
```diff
@@ -169,13 +171,16 @@ export class AndroidDevice extends SdkObject {
...
+      await this.shell(`cmd package uninstall com.microsoft.playwright.androiddriver`);
+      await this.shell(`cmd package uninstall com.microsoft.playwright.androiddriver.test`);
 
-    debug('pw:android')('Installing the new driver');
-    for (const file of ['android-driver.apk', 'android-driver-target.apk'])
-      await this.installApk(await fs.promises.readFile(require.resolve(`../../../bin/${file}`)));
+      debug('pw:android')('Installing the new driver');
+      for (const file of ['android-driver.apk', 'android-driver-target.apk'])
+        await this.installApk(await fs.promises.readFile(require.resolve(`../../../bin/${file}`)));
+    }
```
pavelfeldman, line 183:
```suggestion
    } else {
      debug('pw:android')('Skipping the driver installation');
    }
```

pavelfeldman (approved):
Looks good module review comments.

kaivean:
> @kaivean the tests are failing when the whole test file runs. Would be amazing if you could take a look. I could repro it locally with: `npm run atest device.spec.ts`
> 
> Seems like the transport pipe is closing for some reason.

I just tested omitDriverInstall function. 
Could you give a a download link of chromium webview app? I only found a old version and it could not run tests

mxschmitt:
We have two scripts which should yield to the same android emulator which we have running on the CI:

```
    - name: Create Android Emulator
      run: bash utils/avd_recreate.sh
    - name: Start Android Emulator
      run: bash utils/avd_start.sh
```

PR #13270, opened 2022-04-02
fix: wait for worker to be ready before initializing

geigerzaehler:
The test dispatcher waits until a worker is ready before sending the `init` message. This guarantees that the worker process is listening for the message. Otherwise, the worker process might miss the `init` message and the subsequent `run` message would crash the worker.

I ran into this bug when I set a custom ESM loader with `NODE_OPTIONS=--loader=./esm-loader.mjs`. The following loader delays loading files and can be used to reproduce the issue:
```javascript
// ./esm-loader.mjs
export async function resolve(specifier, context, defaultResolver) {
  await new Promise((resolve) => setTimeout(resolve, 10))
  return defaultResolver(specifier, context)
}
```

### DIFF
--- a/packages/playwright-test/src/dispatcher.ts
+++ b/packages/playwright-test/src/dispatcher.ts
@@ -466,7 +466,7 @@ class Worker extends EventEmitter {
   private _didSendStop = false;
   private _didFail = false;
   private didExit = false;
-  private readonly _ready: Promise<void>;
+  private _ready: Promise<void>;
 
   constructor(hash: string, parallelIndex: number) {
     super();
--- a/packages/playwright-test/src/dispatcher.ts
+++ b/packages/playwright-test/src/dispatcher.ts
@@ -466,7 +466,7 @@ class Worker extends EventEmitter {
   private _didSendStop = false;
   private _didFail = false;
   private didExit = false;
-  private readonly ready: Promise<void>;
+  private readonly _ready: Promise<void>;
 
   constructor(hash: string, parallelIndex: number) {
     super();
@@ -496,14 +496,14 @@ class Worker extends EventEmitter {
       this.emit(method, params);
     });
 
-    this.ready = new Promise((resolve, reject) => {
+    this._ready = new Promise((resolve, reject) => {
       this.process.once('exit', () => reject(new Error('worker exited before it became ready')));
       this.once('ready', () => resolve());
     });
   }
 
   async init(testGroup: TestGroup, loaderData: SerializedLoaderData) {
-    await this.ready;
+    await this._ready;
     const params: WorkerInitParams = {
       workerIndex: this.workerIndex,
       parallelIndex: this.parallelIndex,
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/worker.ts
```diff
@@ -88,6 +87,8 @@ process.on('message', async message => {
   }
 });
 
+sendMessageToParent('ready');
```
pavelfeldman, line 90: I don't think this changes anything, cross-process communication is async.

packages/playwright-test/src/dispatcher.ts
```diff
@@ -466,6 +466,7 @@ class Worker extends EventEmitter {
   private _didSendStop = false;
   private _didFail = false;
   private didExit = false;
+  private readonly ready: Promise<void>;
```
pavelfeldman, line 469: style:

```suggestion
  private _ready: Promise<void>;
```

packages/playwright-test/src/dispatcher.ts
```diff
@@ -494,9 +495,15 @@ class Worker extends EventEmitter {
       const { method, params } = message;
       this.emit(method, params);
     });
+
+    this.ready = new Promise((resolve, reject) => {
+      this.process.once('exit', () => reject(new Error('worker exited before it became ready')));
```
pavelfeldman, line 500: This assumes that failure to create a worker is considered a non-recoverable error and we crash the entirety of the test runner upon it. @dgozman, would we want to handle it and retry?

packages/playwright-test/src/worker.ts
geigerzaehler: Reverted in https://github.com/microsoft/playwright/pull/13270/commits/39a0a52c33b816018056ee12f387e3e99674ffa4

packages/playwright-test/src/dispatcher.ts
geigerzaehler: Fixed in https://github.com/microsoft/playwright/pull/13270/commits/39a0a52c33b816018056ee12f387e3e99674ffa4 and https://github.com/microsoft/playwright/pull/13270/commits/44297b77fdfef941e59e36f6f8b9340bf45308cf

packages/playwright-test/src/dispatcher.ts
dgozman: Yep, we should handle this case without crashing 😄 

I propose the following changes:
- make `_ready` resolve with a boolean (success or failed);
- make `init` return immediately if `_ready` failed, so that it does not stall.
- make `_runJob` check `worker.didExit`, and [call `onExit`](https://github.com/microsoft/playwright/blob/3636d8548f5ad7740fe9db53cb0bca2e0de649c0/packages/playwright-test/src/dispatcher.ts#L393) immediately if the worker did exit already. This will mark scheduled tests as failed, and enqueue a retry if needed.

dgozman: Scratch that, let's crash instead, since something is very much broken in this case.

dgozman (approved):

Issue #13265, opened 2022-04-02
[Question] typescript import enum


hyusetiawan:
I am using next.js / react and inside the page.tsx I export an enum called `TestID`, I use it to mark specific component with test id that I will then share the same test id in jest. I want to do the same thing with playwright but it seems I am getting `undefined` as the enum, as a matter of fact, during runtime `TestID`is a type of `string` how come? Is there a way to share the test ids this way? Trying to avoid hard coding.

rwoll:
Let me make sure I understand the setup. Is this what your Playwright Test import structure/file layout looks like?

```ts
// @filename example.spec.ts
import { MyEnum } from './component'; // THIS IS A TSX FILE!

test('works', () => {
  expect(MyEnum.SomeID).toBe("value");
});
```

```ts
// @filename component.tsx
export enum MyEnum {
    SomeID = "value";
}
```

If so, I repro'd the behavior: https://github.com/microsoft/playwright/pull/13272. For now, as a work around, you can move the enums into a .ts file (`shared.ts`) and import it into your test as well as your component.

hyusetiawan:
@rwoll that setup is correct, glad you are able to reproduce it, is the issue caused because it's tsx?

rwoll:
> is the issue caused because it's tsx

Yes—I received confirmation: "[w]e don't compile TSX, we only stub those altogether for component testing. So this test documented present behavior." [^1]

If you move your enum's into a `.ts` file and import that, you should be good. For now, avoid import'ing `.tsx` files into your tests. They won't do what you expect. (We are experimenting with component testing, so at some point in the future, there may be legitimate reasons to import a `.tsx` file into your test, but not atm.)

[^1]: https://github.com/microsoft/playwright/pull/13272#discussion_r841322948

PR #13272, opened 2022-04-03
test: repro enums in tsx undefined

rwoll:
The enums imported from the .tsx file are ending up as "undefined";
from the .ts file they are defined.

Repro for #13265.

This likely has to do with our component testing/transform; before fixing, I want to
understand expected behavior.

### DIFF
--- a/tests/playwright-test/babel.spec.ts
+++ b/tests/playwright-test/babel.spec.ts
@@ -36,3 +36,43 @@ test('should succeed', async ({ runInlineTest }) => {
   expect(result.passed).toBe(1);
   expect(result.failed).toBe(0);
 });
+
+test('should treat enums equally', async ({ runInlineTest }) => {
+  test.fail(true, 'https://github.com/microsoft/playwright/issues/13265');
+  const result = await runInlineTest({
+    'component.tsx': `
+      export enum MyEnum {
+        Value = "value",
+      };
+
+      export const enum MyConstEnum {
+        Value = "value",
+      }
+    `,
+    'regular.ts': `
+      export enum MyEnum {
+        Value = "value",
+      };
+
+      export const enum MyConstEnum {
+        Value = "value",
+      }
+    `,
+    'example.spec.ts': `
+      const { test } = pwt;
+
+      import * as components from './component';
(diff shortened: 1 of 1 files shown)


tests/playwright-test/babel.spec.ts
```diff
@@ -36,3 +36,43 @@ test('should succeed', async ({ runInlineTest }) => {
...
+    `,
+    'example.spec.ts': `
+      const { test } = pwt;
+
+      import * as components from './component';
+      import * as regular from './regular';
+
+      test('works', () => {
+        expect.soft(components.MyEnum.Value).toBe("value");
+        expect.soft(components.MyConstEnum.Value).toBe("value");
```
rwoll, line 69: @pavelfeldman What should we expect `components.MyConstEnum.Value` to be? It's currently "undefined", while those coming out of `regular.ts` (below) are defined as "value".

tests/playwright-test/babel.spec.ts
```diff
@@ -36,3 +36,43 @@ test('should succeed', async ({ runInlineTest }) => {
   expect(result.passed).toBe(1);
   expect(result.failed).toBe(0);
 });
+
+test('should treat enums equally', async ({ runInlineTest }) => {
+  test.fail(true, 'https://github.com/microsoft/playwright/issues/13265');
```
mxschmitt, line 41: We started to do it differently: https://github.com/microsoft/playwright/blob/02cac8a0668c59078eb35029ed88c9f88e139428/tests/library/browsercontext-page-event.spec.ts#L189

tests/playwright-test/babel.spec.ts
pavelfeldman: We don't compile TSX, we only stub those altogether for component testing. So this test documented present behavior.

pavelfeldman (approved):

tests/playwright-test/babel.spec.ts
rwoll, line 41:
```suggestion
  test.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/13265' }); 
```

PR #13287, opened 2022-04-04
docs(showcase): add odottaa to the tools list

elaichenkov:
Add the new [odottaa](https://github.com/elaichenkov/odottaa) matchers library to the tools list.


pavelfeldman:
Let's wait a bit for the project to get some validation. We don't have any formal criteria in terms of number of stars or users at the moment, but adding every new projects that extends Playwright's capabilities does not scale. Btw, did you consider filing a bug and adding those upstream in Playwright itself?

pavelfeldman:
> the main concern is that it is a new project or that it has not enough stars/users or because it just extends Playwright's capabilities

Every time we add (any) project to this list, we need to check if it is trustworthy. We don't want to point users to something that might potentially compromise them. So we look at the existing community for this project, for people that starred the repository, etc. If the project looks real and used, we gladly add it to the list. But if this is a repo that was created few days ago without users and community behind it, we prefer to wait and revisit it later.

> I'd love to have such solution as built-in functionality. Does Playwright need it?

Filing an issue with the proposal would be a great way to start. It'll gain some stars and we'll have a better idea. You don't need to many stars because you are ready to send a PR - we don't need to prioritize it ourselves, we just need to assess the cost of maintenance after the initial contribution.

elaichenkov:
Thank you for explaining.

PR #13299, opened 2022-04-04
chore: don't auto-install browsers if global install

### LINKED ISSUE #13252: [BUG] npx playwright --help triggers browser downloads
We can be smarter about when to download browsers v. not.

rwoll:
Fixes #13252.


rwoll:
@mxschmitt / @aslushnikov Let's think through this to make sure we won't regress anyone.

The intent is:

```
npm i playwright
```

should still install the browsers (otherwise we break people).

but, `npx playwright <any command other than explicit install>` won't download browsers.

packages/playwright/install.js
```diff
@@ -17,7 +17,8 @@
 let install;
 
 try {
-  install = require('playwright-core/lib/utils/registry').installDefaultBrowsersForNpmInstall;
+  if (!require('playwright-core/lib/utils').isLikelyNpxGlobal())
```
mxschmitt, line 20: maybe also relevant to playwright-chromium/firefox/webkit packages?

mxschmitt (approved):
LGTM but lets get another approval.

rwoll:
> LGTM but lets get another approval.

@mxschmitt Removed your approval; looks like it's [not working as intended](https://github.com/microsoft/playwright/runs/5823572432?check_suite_focus=true). Will debug.

aslushnikov:
looks good, but let's add an installation test for this!

rwoll:
@mxschmitt @aslushnikov Tests added (including for https://github.com/microsoft/playwright/pull/13196)! The relevant bits are `test_npx_`.

You might notice there's some other code to make these tests work! For our installation tests, we now run a local npm registry that reads through to the official one. Our packages are served from their local copies. The registry and cache are isolated per test run. 

You can now technically do `npm i playwright` in an installation test and it will use our local packages! You don't have to install a .tgz yourself. You'll still want to use the `npm_i` helper if you care about capturing installation output.

Having the registry makes `npx` happy!

rwoll:
> @mxschmitt @aslushnikov Tests added (including for #13196)! The relevant bits are `test_npx_`.
> 
> You might notice there's some other code to make these tests work! For our installation tests, we now run a local npm registry that reads through to the official one. Our packages are served from their local copies. The registry and cache are isolated per test run.
> 
> You can now technically do `npm i playwright` in an installation test and it will use our local packages! You don't have to install a .tgz yourself. You'll still want to use the `npm_i` helper if you care about capturing installation output.
> 
> Having the registry makes `npx` happy!

Now that we have @playwright/test, I'm going to take a pass at replacing the installation tests's bash testing rig with @playwright/test. I think it will be much easier to make these kinds of changes in the future (and, when these current tests inevitably break), it will be easier to fix. This will be a separate PR.

rwoll:
Per offline discussion, we'll punt on porting the installation test runner over from bash to @playwright/test. I'll leave https://github.com/rwoll/playwright/commit/5f6e59fb8858195257b35f7105d6c0d4b13986c0 around as it shows my port of these new tests (they become much simpler and isolated), but in the interest of not having two different setups for the installation tests, let's leave this out.

installation-tests/bin/local-playwright-registry
```diff
@@ -0,0 +1,282 @@
...
+    process.exit(0);
+};
+
+const kill = async () => {
+    const pid = await fs.promises.readFile(PID_FILE);
+    process.kill(Number.parseInt(pid, 10));
+    await fs.promises.unlink(PID_FILE).catch(() => { });
+};
+
+const assertLocalPkg = async (pkg) => {
```
aslushnikov, line 224: let's make sure local package installation is also of a proper version

aslushnikov (approved):

aslushnikov (approved):

PR #13300, opened 2022-04-04
chore: unrelease 'fonts' screenshot option

dgozman:


packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -200,6 +200,7 @@ export class ElementHandle<T extends Node = Node> extends JSHandle<T> implements
         selector: locator._selector,
       }));
     }
+    copy.fonts = (options as any)._fonts;
```
rwoll, line 203: When we do experiments feature, we should think about how we can avoid needing to do the `as any` bits. Perhaps something similar to FullConfig and FullConfigInternal.

rwoll (approved):

aslushnikov (approved):

PR #13305, opened 2022-04-04
chore: rewrite non-web protocol for Electron tracing

### LINKED ISSUE #13289: [BUG] electron tracing with `snapshot: true` not very functional
**Context:**
- Playwright Version: `1.20.2`
- Operating System: macOS
- Node.js version: `16.x`
- Browser: `Chromium`

**Code Snippet**

This is from our use of playwright in VSCodes Electron smoke tests, so I cannot really share a snippet. We setup playwright for Electron here (not using `snapshot: true`, `sources: true` in that link, but had it enabled for the attached trace): 

https://github.com/microsoft/vscode/blob/2ada17080cd3e351e093e5fce40d36f16db3124a/test/automation/src/playwrightElectronDriver.ts#L30-L61

**Describe the bug**

See attached trace and open it on https://trace.playwright.dev/. Try to operate in the DOM and notice how that is not possible. Screenshots seem to work fine though.

image

[playwright-trace-1-searches_for_body___checks_for_correct_result_number.zip](https://github.com/microsoft/playwright/files/8408238/playwright-trace-1-searches_for_body___checks_for_correct_result_number.zip)

//cc @pavelfeldman

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/13289

### DIFF
--- a/packages/trace-viewer/src/snapshotRenderer.ts
+++ b/packages/trace-viewer/src/snapshotRenderer.ts
@@ -62,8 +62,9 @@ export class SnapshotRenderer {
           // Never set relative URLs as <iframe src> - they start fetching frames immediately.
           const isFrame = n[0] === 'IFRAME' || n[0] === 'FRAME';
           for (const [attr, value] of Object.entries(n[1] || {})) {
-            const attrToSet = isFrame && attr.toLowerCase() === 'src' ? '__playwright_src__' : attr;
-            builder.push(' ', attrToSet, '="', escapeAttribute(value as string), '"');
+            const attrName = isFrame && attr.toLowerCase() === 'src' ? '__playwright_src__' : attr;
+            const attrValue = attr.toLowerCase() === 'href' || attr.toLowerCase() === 'src' ? rewriteURLForCustomProtocol(value) : value;
+            builder.push(' ', attrName, '="', escapeAttribute(attrValue as string), '"');
           }
           builder.push('>');
           for (let i = 2; i < n.length; i++)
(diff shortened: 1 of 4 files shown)


packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -273,3 +274,33 @@ function snapshotScript() {
...
+    href = href.substring(kLegacyBlobPrefix.length);
+
+  try {
+    const url = new URL(href);
+    // Sanitize URL.
+    if (url.protocol === 'javascript:')
+      return 'javascript:void(0)';
+
+    // Pass through if possible.
+    const isBlob = url.protocol === 'blob:';
```
dgozman, line 294: Can just  remove "blob" from the list, and everything should work?

packages/trace-viewer/src/snapshotServer.ts
```diff
@@ -103,13 +100,3 @@ declare global {
     showSnapshot: (url: string, point?: Point) => Promise<void>;
   }
 }
-
-function removeHash(url: string) {
```
dgozman, line 107: Why dropped this one? We intercept without a hash, but page requests with a hash, so we make sure url matches.

dgozman (approved):

packages/trace-viewer/src/snapshotRenderer.ts
pavelfeldman: Done.

packages/trace-viewer/src/snapshotServer.ts
pavelfeldman: Ouch

PR #13343, opened 2022-04-06
fix: remove tmp files

liuxingbaoyu:
There is a bug in rimraf, maxBusyTries does not take effect, resulting in several gb temporary files on my computer......

### DIFF
--- a/packages/playwright-core/src/utils/processLauncher.ts
+++ b/packages/playwright-core/src/utils/processLauncher.ts
@@ -172,7 +172,7 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
       // Force kill the browser.
       try {
         if (process.platform === 'win32') {
-          const taskkillProcess = childProcess.spawnSync(`chcp 65001>nul&&taskkill /pid ${spawnedProcess.pid} /T /F /FI "MEMUSAGE gt 0"`, { shell: true });
+          const taskkillProcess = childProcess.spawnSync(`chcp 65001>nul&&taskkill /pid ${spawnedProcess.pid} /T /F`, { shell: true });
           const [stdout, stderr] = [taskkillProcess.stdout.toString(), taskkillProcess.stderr.toString()];
           if (stdout)
             options.log(`[pid=${spawnedProcess.pid}] taskkill stdout: ${stdout}`);
--- a/packages/playwright-core/src/utils/processLauncher.ts
+++ b/packages/playwright-core/src/utils/processLauncher.ts
@@ -172,8 +172,8 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
       // Force kill the browser.
       try {
         if (process.platform === 'win32') {
-          const taskkillProcess = childProcess.spawnSync(`taskkill /pid ${spawnedProcess.pid} /T /F /FI "MEMUSAGE gt 0"`, { shell: true });
-          const [stderr, stdout] = [taskkillProcess.stdout.toString(), taskkillProcess.stderr.toString()];
+          const taskkillProcess = childProcess.spawnSync(`chcp 65001>nul&&taskkill /pid ${spawnedProcess.pid} /T /F /FI "MEMUSAGE gt 0"`, { shell: true });
+          const [stdout, stderr] = [taskkillProcess.stdout.toString(), taskkillProcess.stderr.toString()];
           if (stdout)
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/utils/processLauncher.ts
```diff
@@ -190,14 +190,23 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
...
     killProcess();
     options.log(`[pid=${spawnedProcess.pid || 'N/A'}] starting temporary directories cleanup`);
     for (const dir of options.tempDirectories) {
-      try {
-        rimraf.sync(dir, { maxBusyTries: 10 });
-      } catch (e) {
-        options.log(`[pid=${spawnedProcess.pid || 'N/A'}] exception while removing ${dir}: ${e}`);
+      for (let i = 0; i < 10; i++) {
+        try {
+          // maxBusyTries has bugs
```
pavelfeldman, line 203: It does not have bugs, but it only works in async mode, so the code here was simply wrong. Doing it 10 times synchronously does not help though, we need a proper async solution.

packages/playwright-core/src/utils/processLauncher.ts
```diff
@@ -190,14 +190,23 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
     }
   }
 
+  function sleep(ms: Number) {
+    childProcess.spawnSync(process.execPath, ['-e', 'setTimeout(function(){},' + ms + ')']);
```
pavelfeldman, line 194: I can see what you did here. Let's try finding a smaller hammer.

packages/playwright-core/src/utils/processLauncher.ts
liuxingbaoyu: Changed from async to sync in the last commit.
https://github.com/microsoft/playwright/commit/e8c512dbeb3f69ba8dde7fc22d7ee27843f1629a
I don't know the full logic, for smaller changes I chose this solution.

packages/playwright-core/src/utils/processLauncher.ts
liuxingbaoyu: Yes, I found a lot of methods, the most stable one is a loop, but it will consume a lot of cpu, so I chose this relatively simple and relatively stable method.

packages/playwright-core/src/utils/processLauncher.ts
```diff
@@ -190,14 +190,23 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
...
+
   function killProcessAndCleanup() {
     killProcess();
     options.log(`[pid=${spawnedProcess.pid || 'N/A'}] starting temporary directories cleanup`);
     for (const dir of options.tempDirectories) {
-      try {
-        rimraf.sync(dir, { maxBusyTries: 10 });
-      } catch (e) {
-        options.log(`[pid=${spawnedProcess.pid || 'N/A'}] exception while removing ${dir}: ${e}`);
+      for (let i = 0; i < 10; i++) {
```
dgozman, line 201: @liuxingbaoyu Could you please `console.log(new Error().stack)` here to see which specific codepath triggers this sync removal for you? I'd like to better understand the situation before landing a fix.

liuxingbaoyu:
```
    at process.emit (node:events:394:28)    at processEmit [as emit] (F:\playwright\node_modules\signal-exit\index.js:155:35)
    at process.exit (node:internal/process/per_thread:184:15)
    at logErrorAndExit (F:\playwright\packages\playwright-core\lib\cli\cli.js:590:11)
```

dgozman: Thank you, this is interesting! What did you do to trigger an error in CLI? Any specific command that you run?

liuxingbaoyu: Just use codegen or trace-viewer normally

aslushnikov:
@liuxingbaoyu is this still relevant after @dgozman patches?

liuxingbaoyu:
I haven't tested the new version, but the sync cleanup is still there.

https://sourcegraph.com/github.com/microsoft/playwright/-/blob/packages/playwright-core/src/utils/processLauncher.ts?L191:12#tab=references

mxschmitt:
@liuxingbaoyu is this still relevant, did you run into any problems with 1.22?

liuxingbaoyu:
It should have been fixed.

PR #13369, opened 2022-04-07
docs: mark optional properties/arguments explicitly with "?"

dgozman:
Previously, we were guessing whether something is optional from the
comment (looking for stuff like "defaults to" or "optional").

To remove this error-prone technique, every optional property or
argument (but not option) is explicitly marked with "?".


pavelfeldman (approved):

rwoll (commented):
🎉  I like this (hence https://github.com/microsoft/playwright/pull/13358#discussion_r844278585). I prefer explicit over trying to infer from natural language.

PR #13371, opened 2022-04-07
Update class-coverage.md

stevez:
Based on the api, v8toIstanbul is a function instead of class. When I compile this code snippet, I got compiler error, after I removed the `new` keyword, then build is successful


mxschmitt (approved):

mxschmitt:
@stevez make sure to sign the CLA otherwise we can't merge.

PR #13375, opened 2022-04-07
fix: extendInjectedScript

### LINKED ISSUE #12988: [BUG] Codegen via iframes does not work all the time
Repro:

1. `npx playwright codegen https://support-shop.astoreprocurement.com/access/unauthenticated`
2. Wait for 5 seconds - do not focus the site
3. Try to fill something in the email page

Expected: Selector does get generated
Actual: It keeps saying just `iframe` and does not recognise the interactions inside the iframe.

Note: It seems its related to the lazy initiated iframe - when you reload just the iframe it then works correctly.

liuxingbaoyu:
fixes: #12988

### DIFF
--- a/packages/playwright-core/src/server/browserContext.ts
+++ b/packages/playwright-core/src/server/browserContext.ts
@@ -426,7 +426,10 @@ export abstract class BrowserContext extends SdkObject {
   }
 
   async extendInjectedScript(source: string, arg?: any) {
-    const installInFrame = (frame: frames.Frame) => frame.extendInjectedScript(source, arg).catch(() => {});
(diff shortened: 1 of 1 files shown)


liuxingbaoyu:
This is a terrible workaround, does anyone have a better way?

mxschmitt:
Indeed! Let's continue digging into the CDP networking calls, to understand why the main utility context is destroyed. Maybe its an upstream chromium bug too. Having a repro for it would be a good start for it, based on that we can write a test and then continue digging into the actual root-cause.

liuxingbaoyu:
Very bad find!

Searching the repository for `catch(() => {})` can find many such codes that ignore exceptions.

And also because the exception was ignored, I spent a lot of time to solve this issue.

This means that we may encounter many unstable problems and cannot find the cause.

PR #13399, opened 2022-04-07
chore: use utils via index export (2)

pavelfeldman:


packages/playwright-core/src/utils/netUtils.ts
```diff
@@ -23,3 +27,85 @@ export async function createSocket(host: string, port: number): Promise<net.Sock
     socket.on('error', error => reject(error));
   });
 }
+
+// `https-proxy-agent` v5 is written in TypeScript and exposes generated types.
```
mxschmitt, line 31: potential drive-by: we could migrate it over to import, we have esModuleInterop now enabled.

PR #13401, opened 2022-04-07
feat(test-runner): support filtering by columns

mxschmitt:
Motivation: Inside stack traces or html report, or flakiness dashboard etc. we usually print `file-name:file-line:file-column`. When double clicking on it, it selects it usually with the column. When passing this to the test-runner it did not work. After this patch it works.


dgozman (approved):
This probably needs a documentation update.

PR #13404, opened 2022-04-07
chore: use utils via index export (4)

pavelfeldman:


packages/playwright-core/src/utils/debugLogger.ts
```diff
@@ -1,83 +0,0 @@
...
-  'browser': 0, // reset
-  'proxy': 92, // purple
-  'error': 160, // red,
-  'channel:command': 33, // blue
-  'channel:response': 202, // orange
-  'channel:event': 207, // magenta
-};
-export type LogName = keyof typeof debugLoggerColorMap;
-
-class DebugLogger {
```
mxschmitt, line 34: where did it go?

pavelfeldman: Ouch!

mxschmitt (approved):

PR #13411, opened 2022-04-07
fix: wait for cleanup on double SIGINT

dgozman:
This is a speculative fix to leftover tmp directories. When users issues SIGINT twice, we enter `gracefullyClose()` twice, and shortcut the second time. It turns out, we do not wait for directories removal.

Note: it is unknown how often we reach this codepath in practice.


dgozman:
@liuxingbaoyu, @aslushnikov Could you please take a look at this? I suspect this could be one of the reasons of letfover tmp directories. See also #13343 for more context.

aslushnikov (approved):

liuxingbaoyu:
https://github.com/microsoft/playwright/blob/5e2a4a039471b691f91f621e42e39e126995969d/packages/playwright-core/src/utils/processLauncher.ts#L125

Use codegen, traceviewer, or manually close while developing, will go here

Issue #13362, opened 2022-04-06
[Feature] Test Plugins


rwoll:
Hook into Playwright Test's lifecycle to customize behavior easily, with minimal config. Similar to Custom Reporter, but spans entire run lifecycle, including modifying state (pushing attachments, influencing sharding/scheduling decisions, adding CLI flags, etc.).

Some uses for plugins:
- generating tests / data for use in tests (e.g. yaml file -> test cases run via Playwright, Gherkin style tests)
- test compilation and/or transformation (e.g. custom tsx)
- scheduling/sharding: currently, PWT's scheduling and sharding logic is static
- set multiple configurations at once (e.g. worker count, reporter, connection, webServers/launchServices)
- run test on file change
- dynamically set timeouts
- add default fixtures (like `test.extend`)

Some quick ideas for API:

```ts
import { webServer } from '@rwoll/playwright-plugins';
import { PlaywrightTestConfig } from '@playwright/test';
const config: PlaywrightTestConfig = {
  plugins: [
     compile('./src'),
     webServer('./html/files'),
     remoteConnection(),
     testsFromDB(dbConnectionString),
     // …
  ],
};
export default config;
```

```ts
interface Plugin {
    setup(config);
    collectFiles();
    schedule();
    …
}
```

Some open questions?

- how do plugins interact with one another? For example, if we run the `webServer` plugin, can it output a port/URL for other plugins to consume? Can we get typesafety and dependency ordering by make plugins like fixtures?

ryanjduffy:
I think this would be very useful for our replay.io integration. We've added some integration via [action-playwright](https://github.com/Replayio/action-playwright) and a [custom reporter](https://github.com/Replayio/replay-cli/tree/main/packages/playwright) but we need to go one step deeper to be able to set environment variables for the browser for each worker.

I've explored using a fixture for this which I think can work but seems to require the consumers explicitly use the fixture from their tests which is not ideal. I'd rather rely on declarative config so the test files stay the same but they can opt in via config to using the Replay browser.

denis-domanskii:
Probably it may be useful for services initialisations, see here: #7445 . Because fixtures, as I understand, don't guarantee that the code will be called once. Even with a worker scope.

dgozman:
Closing for now.

PR #13444, opened 2022-04-09
feat(test-runner): add proof-of-concept WebServerPlugin

rwoll:
This patch introduces the concept of Plugins, and introduces two plugins:

- `versionControlSystemInfo`: attaches VCS/git metadata that the HTML Reporter renders nicely
- `webServer`: shares underlying implementation with existing our existing [Web Server feature](https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests), but since you can have as many plugins as you like, you can now launch multiple Web Servers. The Plugin implementation differs in logging: logs will _**NOT**_ be captured to reporters; they are just dumped to stdout/stderr depending on the log-level enabled (via `DEBUG=pw:webserver`).

```js
// @filename:playwright.config.ts
      const { webServer } = require('@playwright/test');
      module.exports = {
        plugins: [
            webServer({
                command: 'node ${JSON.stringify(path.join(__dirname, 'assets', 'simple-server.js'))} ${port}',
                port: ${port},
            }),
            webServer({
                command: 'node ${JSON.stringify(path.join(__dirname, 'assets', 'simple-server.js'))} ${port2}',
                port: ${port3},
            }),
        ],
      };
```

Per-discussion, for this first pass we've decided to punt on:
- captured logging: we likely need to put the plugins in their own process to do this properly
- configure hook: there are too many initial open questions from an earlier commit we'd like to think through (e.g. do we mutate or not? when does it run? what can be changed? etc.)

Relates #13362
Relates #8206


docs/src/test-api/class-testplugin.md
```diff
@@ -0,0 +1,12 @@
...
+
+## optional async method: TestPlugin.globalSetup
+
+### param: TestPlugin.globalSetup.config
+- `config` <[FullConfig]>
+
+### param: TestPlugin.globalSetup.globalInfo
+- `info` <[GlobalInfo]>
+
+## optional async method: TestPlugin.globalTeardown
```
rwoll, line 12: This needs to be filled out 😄

packages/playwright-test/src/cli.ts
```diff
@@ -153,7 +153,7 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
     return;
 
   const runner = new Runner(overrides, { defaultConfig });
-  const config = resolvedConfigFile ? await runner.loadConfigFromResolvedFile(resolvedConfigFile) : runner.loadEmptyConfig(configFileOrDirectory);
+  const config = resolvedConfigFile ? await runner.loadConfigFromResolvedFile(resolvedConfigFile) : await runner.loadEmptyConfig(configFileOrDirectory);
```
rwoll, line 156:
```suggestion
  const config = resolvedConfigFile ? await runner.loadConfigFromResolvedFile(resolvedConfigFile) : runner.loadEmptyConfig(configFileOrDirectory);
```

packages/playwright-test/src/index.ts
```diff
@@ -25,6 +25,7 @@ import { GridClient } from 'playwright-core/lib/grid/gridClient';
 export { expect } from './expect';
 export const _baseTest: TestType<{}, {}> = rootTestType.test;
 import * as outOfProcess from 'playwright-core/lib/outofprocess';
+export { webServer } from './webServerPlugin';
```
rwoll, line 28: What do we want the public export to be? It's currently:

```ts
import { webServer } from '@playwright/test';
```

I don't mind this now, but if we add more official plugins, I think we'd want a sub-package:

```ts
import { webServer } from '@playwright/test/plugins';
```

packages/playwright-test/src/loader.ts
```diff
@@ -66,7 +65,7 @@ export class Loader {
       config = config['default'];
     this._configFile = file;
     const rawConfig = { ...config };
-    this._processConfigObject(config, path.dirname(file));
+    await this._processConfigObject(config, path.dirname(file));
```
rwoll, line 68:
```suggestion
    this._processConfigObject(config, path.dirname(file));
```

docs/src/test-api/class-testconfig.md
```diff
@@ -652,6 +652,9 @@ test('test', async ({ page }) => {
 });
 ```
 
+## property: TestConfig.plugins
+- type: <[Array]<[TestPlugin]>>
```
rwoll, line 656: This should be marked as experimental.

rwoll (commented):
@aslushnikov I've left some open questions inline that we should resolve before trying to land; otherwise, please take a look and leave responses/comments/questions.

Thanks!

rwoll:
Closing as we landed the base work for this in https://github.com/microsoft/playwright/pull/13717

PR #13462, opened 2022-04-10
feat(android) add webview pages method

kaivean:
## why 
In mobile phone,  in order to save cpus and memory,  only a webview process exist and load many pages.   there are also some preloaded pages in a app webview.  so I need get all pages to test business logics.

## how
```
const pages = await webview.pages();
```

### DIFF
--- a/tests/android/webview.spec.ts
+++ b/tests/android/webview.spec.ts
@@ -78,3 +78,21 @@ test('select webview from socketName', async function({ androidDevice }) {
   await newPage.close();
   await context.close();
 });
+
+test('webview.pages', async function({ androidDevice }) {
+  test.slow();
+  // open first page
+  const context = await androidDevice.launchBrowser();
+  // open second page
+  const newPage = await context.newPage();
+  newPage.goto('about:blank');
+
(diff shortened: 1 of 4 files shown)


mxschmitt:
Hey @kaivean we'll review it next week, since the person who is responsible for Android is currently OOO.

tests/android/webview.spec.ts
```diff
@@ -77,4 +77,22 @@ test('select webview from socketName', async function({ androidDevice }) {
...
+
+test('webview.pages', async function({ androidDevice }) {
+  test.slow();
+  // open first page
+  const context = await androidDevice.launchBrowser();
+  // open second page
+  const newPage = await context.newPage();
+  newPage.goto('about:blank');
+
+  const webview = await androidDevice.webView({ socketName: 'webview_devtools_remote_playwright_test' });
```
pavelfeldman, line 90: The pages in lines 85 and 87 are created in the browser, those are not web views. I don't think this tests will pass because there are no webviews with the webview_devtools_remote_playwright_test socket name.

pavelfeldman:
Respectfully closing this. Please feel free to re-file should you get back to this effort!

PR #13469, opened 2022-04-11
chore: set minimum version to Node.js 14

mxschmitt:
Node.js 12 EOL is on the 2022-04-30, this is before our next release. So in our next release we can stop supporting Node.js 12 officially.

### DIFF
--- a/docs/src/library-js.md
+++ b/docs/src/library-js.md
@@ -97,7 +97,7 @@ let page: import('playwright').Page;
 
 ## System requirements
 
-Playwright requires Node.js version 12 or above. The browser binaries for Chromium,
+Playwright requires Node.js version 14 or above. The browser binaries for Chromium,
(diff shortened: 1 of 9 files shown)


mxschmitt:
Lets wait until Node.js 18 comes out and then bump it all-together.

Meir017:
missed https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/stackTrace.ts#L70-L72 :)

pavelfeldman:
Does it make anyone happy?

mxschmitt:
As per offline discussion with @pfeldman we'll drop official support for it but don't break the users for now. So we'll keep the Node.js 12 hacks in there for now.

pavelfeldman:
Let's add a new Node right away so that it was clear that we were saving bots for it!

dgozman (approved):

PR #12877, opened 2022-03-18
chore: print error if install-deps is used != ubuntu

mxschmitt:
We've seen a plenty of users trying to install-deps on != ubuntu like debian. This prints a pretty error for them instead of "package xyz not found".

Fixes https://github.com/microsoft/playwright/issues/12845
Fixes https://github.com/microsoft/playwright/issues/12859

PR #13536, opened 2022-04-13
revert(#12877): chore: print error if install-deps is used != ubuntu

### LINKED ISSUE #13530: [REGRESSION]: Pull request #12877 prevents the library from being used on any linux distro that is not Ubuntu
**Context:**
- GOOD Playwright Version: < 1.21.0 (1.20.2, etc)
- BAD Playwright Version: 1.21.0
- Operating System: Linux
- Extra: while we are using Docker images based on buster slim, this is applicable to any Linux distro that isn't ubuntu based (debian, possibly arch too)

**Code Snippet**

Run `npx playwright install-deps chrome` in a non-ubuntu environment (debian for instance)

**Describe the bug**

https://github.com/microsoft/playwright/pull/12877 was merged in order to "solve" issues related to users of Debian that couldn't install the required dependencies due to missing packages (usually fonts which can be solved by including the debian non-free source list), however this prevents any users of a Linux distro that is not ubuntu based from using the library.

I believe the PR should be reverted or changed to emit an actual warning instead of a hard error, or a way to ignore the check should be added, as artificially limiting the library just for Ubuntu/Linux Mint users doesn't sound right.

aslushnikov:
Reason: turns out Debian Buster requires just one source list to
install `ttf-ubuntu-font-family` font.

All other dependencies are satisfied.

Fixes #13530

### DIFF
--- a/packages/playwright-core/src/server/registry/dependencies.ts
+++ b/packages/playwright-core/src/server/registry/dependencies.ts
@@ -23,7 +23,6 @@ import { spawnAsync } from '../../utils/spawnAsync';
 import { hostPlatform } from '../../utils/hostPlatform';
 import { buildPlaywrightCLICommand } from '.';
 import { deps } from './nativeDeps';
-import { getUbuntuVersion } from '../../utils/ubuntuVersion';
 import { getPlaywrightVersion } from '../../common/userAgent';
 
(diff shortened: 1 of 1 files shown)


rwoll (approved):
Assuming it's verified to work (unofficially) on Debian, LGTM

PR #13544, opened 2022-04-13
chore: force localPaths to be resolved

### LINKED ISSUE #13537: [internal] validate localPaths is absolute, fully resolved
Enforce that clients send absolute paths to the driver, otherwise renderers crash in unexpected ways.

rwoll:
Relative paths can cause the renderer to crash.

Fixes #13537


yury-s (approved):
Can we have a test for this ?

yury-s (approved):

PR #13549, opened 2022-04-14
feat: restore toHaveScreenshot as experimental feature

dgozman:
- Restore docs.
- Make `TestConfig.expect` generated.
- Allow experimental properties with "e" marker: `- foo e<float>`.
- Use experimental types in test runner implementation to avoid `as any` casts.

Note: `toHaveScreenshot` was unreleased in #13304.

### DIFF
--- a/packages/playwright-test/src/matchers/toMatchSnapshot.ts
+++ b/packages/playwright-test/src/matchers/toMatchSnapshot.ts
@@ -17,7 +17,7 @@
 import type { Locator, Page } from 'playwright-core';
 import type { Page as PageEx } from 'playwright-core/lib/client/page';
 import type { Locator as LocatorEx } from 'playwright-core/lib/client/locator';
-import type { Expect } from '../types';
+import type { Expect, UpdateSnapshots } from '../types';
 import { currentTestInfo } from '../globals';
 import type { ImageComparatorOptions, Comparator } from 'playwright-core/lib/utils/comparators';
 import { getComparator } from 'playwright-core/lib/utils/comparators';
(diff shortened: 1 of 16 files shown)


packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -308,7 +309,7 @@ export async function toHaveScreenshot(
   const [page, locator] = pageOrLocator.constructor.name === 'Page' ? [(pageOrLocator as PageEx), undefined] : [(pageOrLocator as Locator).page() as PageEx, pageOrLocator as LocatorEx];
   const screenshotOptions = {
     animations: config?.animations ?? 'disabled',
-    fonts: process.env.PLAYWRIGHT_EXPERIMENTAL_FEATURES ? (config?.fonts ?? 'ready') : undefined,
```
aslushnikov, line 311: let's keep it?

utils/generate_types/index.js
```diff
@@ -531,7 +545,15 @@ class TypesGenerator {
     const documentation = coreDocumentation.mergeWith(testDocumentation);
     const generator = new TypesGenerator({
       documentation,
-      classNamesToGenerate: new Set(['TestError', 'TestInfo', 'WorkerInfo', ...assertionClasses]),
+      doNotGenerate: new Set([
+        ...coreDocumentation.classesArray.map(cls => cls.name).filter(name => !assertionClasses.has(name)),
+        'PlaywrightAssertions',
```
aslushnikov, line 550: not needed?

aslushnikov (approved):
let's revert types part as discussed offline

PR #13567, opened 2022-04-14
docs: expand TypeScript doc with modules and path mappings

### LINKED ISSUE #13515: [BUG] ESM: imports don't work [ERR_MODULE_NOT_FOUND] and [ERR_UNSUPPORTED_DIR_IMPORT]
**Context:**
- Playwright Version: 1.21.0
- Operating System: Mac
- Node.js version: 16
- New empty project created with `create-playwright`
- set type="module" in package.json

**Code Snippet**

```typescript
// import from ./file.ts
import { b } from './file' // [ERR_MODULE_NOT_FOUND]
// import from a folder/index.ts
import { a } from './folder' // [ERR_UNSUPPORTED_DIR_IMPORT]
```

**Describe the bug**

In ESM project (TypeScript) imports don't work

Reproduction repo: https://github.com/ebeloded/playwright-13515

dgozman:
Fixes #13515.

### DIFF
--- a/docs/src/test-typescript-js.md
+++ b/docs/src/test-typescript-js.md
@@ -1,17 +1,109 @@
 ---
 id: test-typescript
-title: "Advanced: TypeScript"
+title: "TypeScript"
 ---
+
+Playwright Test supports TypeScript out of the box. You just write tests in TypeScript and Playwright Test will read them, transform to JavaScript and run. This works both with [CommonJS modules](https://nodejs.org/api/modules.html) and [ECMAScript modules](https://nodejs.org/api/esm.html).
+
+## TypeScript with CommonJS
+
+[Node.js](https://nodejs.org/en/) works with CommonJS modules **by default**. Unless you use `'.mjs'` or `'.mts'` extensions, or specify `type: "module"` in your `pacakge.json`, Playwright Test will treat all TypeScript files as CommonJS. You can then import as usual without an extension.
+
+Consider this helper module written in TypeScript:
+
+```js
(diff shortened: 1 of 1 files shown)


docs/src/test-typescript-js.md
```diff
@@ -1,17 +1,109 @@
...
+  await page.locator('#username').fill(username);
+  await page.locator('#password').fill(password);
+});
+```
+
+## TypeScript with ESM
+
+You can opt into using [ECMAScript modules](https://nodejs.org/api/esm.html) by setting `type: "module"` in your `package.json` file. Playwright Test will switch to the ESM mode once it reads the `playwright.config.ts` file, so make sure you have one.
+
+Playwright Test [follows the experimental support for ESM in TypeScript](https://www.typescriptlang.org/docs/handbook/esm-node.html) and **requires an extension** when importing from a module, either `'.js'` or `'.ts'`.
```
rwoll, line 37:
```suggestion
Playwright Test [follows the experimental support for ESM in TypeScript](https://www.typescriptlang.org/docs/handbook/esm-node.html) and (per the spec) **requires an extension** when importing from a module, either `'.js'` or `'.ts'`.
```

to make it more clear that this isn't Playwright quirk, but rather a function of the upstream spec.

rwoll (approved):
nit: consider using `:` instead of `.` with each direction leading to a code block.

i.e. 

```patch
- Consider this helper module written in TypeScript.
+ Consider this helper module written in TypeScript:
```

PR #13588, opened 2022-04-16
chore: babel, expect, zip bundles

pavelfeldman:


packages/playwright-core/src/server/registry/browserFetcher.ts
```diff
@@ -24,6 +23,8 @@ import { existsAsync } from '../../utils/fileUtils';
 import { debugLogger } from '../../common/debugLogger';
 import { download } from './download';
 
+const extract: typeof import('../../../bundles/zip/node_modules/extract-zip') = require('../../zipBundle').extract;
```
dgozman, line 26: Why not `typeof import('extract-zip')` similar to `yazl`?

packages/playwright-test/.npmignore
```diff
@@ -1 +1,2 @@
-src
\ No newline at end of file
+src
```
dgozman, line 1: We should switch this file to allowlist!

package.json
```diff
@@ -45,6 +45,12 @@
   ],
   "devDependencies": {
     "@babel/cli": "^7.16.8",
+    "@babel/plugin-proposal-class-properties": "^7.16.7",
+    "@babel/plugin-proposal-export-namespace-from": "^7.16.7",
+    "@babel/plugin-proposal-logical-assignment-operators": "^7.16.7",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.16.7",
+    "@babel/plugin-proposal-optional-chaining": "^7.16.7",
+    "@babel/plugin-transform-modules-commonjs": "^7.17.9",
```
dgozman, line 53: Why do we have some of these modules here? Is it to build our own source?

packages/playwright-test/src/util.ts
```diff
@@ -57,7 +57,6 @@ function filterStackTrace(e: Error) {
         return true;
       return !fileName.startsWith(PLAYWRIGHT_TEST_PATH) &&
              !fileName.startsWith(PLAYWRIGHT_CORE_PATH) &&
-             !fileName.startsWith(EXPECT_PATH) &&
```
dgozman, line 60: Fancy!

utils/generate_third_party_notice.js
```diff
@@ -0,0 +1,73 @@
...
+
+This project incorporates components from the projects listed below. The original copyright notices and the licenses under which Microsoft received such components are set forth below. Microsoft reserves all rights not expressly granted herein, whether by implication, estoppel or otherwise.
+`);
+
+    const allPackages = {};
+    const projectDir = path.join(__dirname, '..', 'packages', project);
+    const bundlesDir = path.join(projectDir, 'bundles');
+    for (const bundle of fs.readdirSync(bundlesDir)) {
+      const dir = path.join(bundlesDir, bundle);
+      execSync('npm ci', { cwd: dir });
```
dgozman, line 37: Why this one? I see every bundle package.json has `npm ci` in every command.

packages/playwright-core/src/server/registry/browserFetcher.ts
pavelfeldman: extract-zip does not have a definitely-typed module, so i'd need a code devDependency in package.json

packages/playwright-test/.npmignore
pavelfeldman: Yeah, in a separate patch

package.json
pavelfeldman: Yes, our build has a babel step.

utils/generate_third_party_notice.js
pavelfeldman: It was so that i could just run that one alone at any moment

PR #13615, opened 2022-04-18
chore: add core utils bundle

pavelfeldman:


package.json
```diff
@@ -99,6 +99,7 @@
     "proxy": "^1.0.2",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "rimraf": "^3.0.2",
```
yury-s, line 102: Looks like we already have it https://github.com/microsoft/playwright/blob/4f0103fef0e70000ee89bf7503891d57e805e6b9/packages/playwright-test/package.json#L42

yury-s: Cosider using rimrafBundle here?

yury-s (approved):

package.json
pavelfeldman: Yes, it is already using bundle in the src, was a stray dependency.

PR #13644, opened 2022-04-20
chore: port installation tests

rwoll:
- [x] port remaining .todo tests
- [x] flip CI to new tests


tests/installation/playwright-electron-should-work.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { test } from './npmTest';
+
+test('electron should work', async ({ npm, exec, envOverrides }) => {
+  envOverrides['-PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD'] = '1';
```
mxschmitt, line 19: what does the minus do here?

tests/installation/playwright-should-work-with-relative-home-path.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+import { test } from './npmTest';
+
+test('playwright should work with relative home path', async ({ npm, npx, exec, envOverrides, nodeVersion }) => {
+  envOverrides['PLAYWRIGHT_BROWSERS_PATH'] = '0';
+  envOverrides['HOME'] = '.';
+
+  await npm('i', '--foreground-scripts', 'playwright');
+  // Firefox does not work with relative HOME.
+  await exec('node', ['sanity.js', 'playwright', 'chromium', 'webkit']);
```
mxschmitt, line 24: sounds like we can move sanity.js to a JS function?

tests/installation/npmTest.ts
```diff
@@ -0,0 +1,380 @@
...
+      };
+    }
+
+    return {
+      pass: true,
+    };
+  }
+});
+
+interface CustomExpect extends Expect {
```
mxschmitt, line 95: We could extend the PlaywrightTest namespace see [here](https://playwright.dev/docs/test-advanced#add-custom-matchers-using-expectextend) instead of a custom interface, might be better and follow the official way.

tests/installation/npmTest.ts
```diff
@@ -0,0 +1,387 @@
...
+
+              args = argsAndOrOptions as string[];
+
+              let result!: Awaited<ReturnType<typeof spawnAsync>>;
+              await test.step(`exec: ${[cmd, ...args].join(' ')}`, async () => {
+                result = await spawnAsync(cmd, args, {
+                  shell: true,
+                  cwd: options.cwd ?? tmpWorkspace,
+                  env: {
+                    'PATH': process.env.PATH,
```
rwoll, line 345:
```suggestion
                  env: {
                     // NOTE(rwoll): We end up running npm-in-npm, so passing along all of process.env would be problematic. Instead, we cherry-pick the minimal vars that we need.
                    'PATH': process.env.PATH,
```

tests/installation/npmTest.ts
```diff
@@ -0,0 +1,387 @@
...
+                  const message = options.message ? ` Message: ${options.message}` : '';
+                  throw new Error(`Expected the command to exit with an error, but exited cleanly.${message}`);
+                }
+              } else if (result.code !== 0) {
+                const message = options.message ? ` Message: ${options.message}` : '';
+                throw new Error(`Expected the command to exit cleanl (0 status code), but exited with ${result.code}.${message}`);
+              }
+
+              return stdio;
+
```
rwoll, line 377:
```suggestion
              return stdio;
```

tests/installation/npmTest.ts
```diff
@@ -0,0 +1,387 @@
...
+              // User expects command to fail
+              if (options.expectToExitWithError) {
+                if (result.code === 0) {
+                  const message = options.message ? ` Message: ${options.message}` : '';
+                  throw new Error(`Expected the command to exit with an error, but exited cleanly.${message}`);
+                }
+              } else if (result.code !== 0) {
+                const message = options.message ? ` Message: ${options.message}` : '';
+                throw new Error(`Expected the command to exit cleanl (0 status code), but exited with ${result.code}.${message}`);
```
rwoll, line 373:
```suggestion
                throw new Error(`Expected the command to exit cleanly (0 status code), but exited with ${result.code}.${message}`);
```

tests/installation/playwright.config.ts
```diff
@@ -0,0 +1,32 @@
...
+loadEnv({ path: path.join(__dirname, '..', '..', '.env') });
+
+const config: PlaywrightTestConfig = {
+  testIgnore: '**\/fixture-scripts/**',
+  globalSetup: path.join(__dirname, 'globalSetup'),
+  timeout: 5 * 60 * 1000,
+  retries: 0,
+  reporter: process.env.CI ? 'dot' : [['list'], ['html', { open: 'on-failure' }]],
+  forbidOnly: !!process.env.CI,
+  workers: 1,
```
rwoll, line 29: Parallel doesn't quite work…so we pin to `workers: 1`…for now! Post-merge, we can dig in!

rwoll (commented):
🎉 @aslushnikov / @mxschmitt : This is now ready for a more formal review!

Everything is ported over! Since last time, there have been some ergonomic changes (based on my own experience and feedback from @aslushnikov):

Screen Shot 2022-04-21 at 11 36 08 AM


## Improved Debugging 

* Set `DEBUG=itest:*` for realtime output from any commands executing.
* Commands being executed are now turned into Playwright Test Steps. This means they show up in the CLI reporter as they each run, as well as in the HTML Report. This makes it much easier to know what the test is doing/waiting on in realtime on the CLI.
 * Exec output is **always** attached to the test run. (1 attach per `exec` call), so you can easily see the output in the HTML Report of each command. This is helpful in cases when things go wrong, but it's also helpful in writing more assertion against the output. Attachment includes: command executed (which is also its name), exit code, stderr/stdout.
 * All `tmpWorkspace`s are rooted in `/tmp/pwt/workspaces`. This dir is left in tact after each test run and only removed at the start (via globalSetup), so you can debug its contents. Set `DEBUG=itest` to see which specific sub-dirs correspond to each test within there. (We still use a random subdir per-test…maybe we want to change that? But we do want it to put out of the checkout tree.)

## Ergonomics

* npm and npx helpers are dropped
* exec can now be used like: `await exec('npm i playwright')` or `await exec('npm', 'i', 'playwright')`. The last argument (even if using variadic sig), are now options. See below.
* `envOverrides` fixture is also dropped. Instead, supply options as the last argument to `exec`.
* ExecOptions (all optional) take:
  * `env`: environment variables merged in with fixture's
  * `expectToExitWithError`: Set this if you expect `exec` to exit with non-zero exit code. `exec` will throw by default if the command exits with a non-zero code. (You no longer need to check exit codes explicitly.) If this option is true, we'll assert the command doesn't exit cleanly.
  * `cwd`
  * `message`: custom failure message

tests/installation/typescript-types.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { test } from './npmTest';
+
+test('typescript types should work', async ({ exec, tsc, writeFiles }) => {
+  const env = { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1' };
```
aslushnikov, line 19:
```suggestion
```

tests/installation/typescript-types.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  await exec('npm i -D @types/node@14.18.9');
```
aslushnikov, line 20: Originally, this test checks against typescript 3.7.5 & thus installs a very particular version of node types. I don't think this is important though, and with this PR we'll switch typechecking to ts 3.8 altogether.

I'll need someone to confirm typechecking everything against ts3.8 is good!

```suggestion
```

tests/installation/typescript-types.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const libraryPackages = [
+    'playwright',
+    'playwright-core',
+    'playwright-firefox',
+    'playwright-webkit',
+    'playwright-chromium',
+  ];
+  await exec('npm i @playwright/test', ...libraryPackages, { env });
```
aslushnikov, line 29:
```suggestion
  await exec('npm i @playwright/test', ...libraryPackages, { env: { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1' }});
```

tests/installation/typescript-types.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  for (const libraryPackage of libraryPackages) {
+    const filename = libraryPackage + '.ts';
+    await writeFiles({
+      [filename]: `import { Page } from '${libraryPackage}';`,
+    });
+    await tsc(filename, { env });
```
aslushnikov, line 36:
```suggestion
    await tsc(filename);
```

tests/installation/typescript-types.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  await tsc('playwright-test-types.ts', { env });
```
aslushnikov, line 39:
```suggestion
  await tsc('playwright-test-types.ts');
```

tests/installation/skip-browser-download-inspect-with-custom-executable.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { test, expect } from './npmTest';
+
+test('should skip download', async ({ exec }) => {
+  const env = { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1' };
```
aslushnikov, line 19:
```suggestion
```

tests/installation/skip-browser-download-inspect-with-custom-executable.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  const installOutput = await exec('npm i --foreground-scripts playwright', { env });
```
aslushnikov, line 20:
```suggestion
  const installOutput = await exec('npm i --foreground-scripts playwright', { env: { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1' }});
```

tests/installation/skip-browser-download-inspect-with-custom-executable.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  expect(installOutput).toContain('Skipping browsers download because');
+  if (process.platform === 'linux') {
+    const output = await exec('node inspector-custom-executable.js', { env });
```
aslushnikov, line 23: Original test launched with the `PWDEBUG` env variable:
```suggestion
    const output = await exec('node inspector-custom-executable.js', { env: { PWDEBUG: '1' }});
```

aslushnikov (commented):
looks awesome!

The biggest change seems to be the `PLAYWRIGHT_BROWSERS_PATH` change:
- we used to default to `PLAYWRIGHT_BROWSERS_PATH=0` in our installation tests
- now we'll default to `PLAYWRIGHT_BROWSERS_PATH=./browsers`

I'd add a test to make sure hermetic installation (with PLAYWRIGHT_BROWSERS_PATH=0) actually works.

tests/installation/npmTest.ts
```diff
@@ -0,0 +1,387 @@
...
+            await spawnAsync('npm', ['init', '-y'], {
+              cwd: tmpWorkspace,
+            });
+
+            await use(tmpWorkspace);
+            // NOTE WELL: We do not remove for easier debugging
+            // await promisify(rimraf)(tmpWorkspace);
+          },
+          registry: async ({}, use, testInfo) => {
+            const port = testInfo.workerIndex + 16123;
```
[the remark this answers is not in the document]
rwoll: Not lazy! My old code did that, so it's very possible. However, this is simpler (and we do it for `server` fixture in other projects). If it's dynamic, we have to make `registry.url()` a promise, and rewrite the NPM Metadata's tarball field response at some point since that's generated _before_ we start the server. (We could change it so the server is started before we publish our packages, but I'd rather the server to be down until it's actually ready.)

tests/installation/npmTest.ts
```diff
@@ -0,0 +1,387 @@
...
+          tmpWorkspace: async ({}, use) => {
+            // We want a location that won't have a node_modules dir anywhere along its path
+            const tmpWorkspace = await fs.promises.mkdtemp(path.join('/tmp/pwt/workspaces', 'playwright-installation-tests-workspace-'));
+            debug(`Workspace Folder: ${tmpWorkspace}`);
+            await spawnAsync('npm', ['init', '-y'], {
+              cwd: tmpWorkspace,
+            });
+
+            await use(tmpWorkspace);
+            // NOTE WELL: We do not remove for easier debugging
```
rwoll, line 316:
```suggestion
```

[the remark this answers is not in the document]
rwoll: Will do. Oh! Normally I write: `NB: some comment`. The _NB_ stands for _nota bene_ which is a comment convention (https://dictionary.cambridge.org/dictionary/english/nb), and apparently I just subconsciously translated it in this instance!

I picked up the `NB: …` convention from academia somewhere along the way.

.gitignore
```diff
@@ -25,3 +25,5 @@ playwright-report
 /packages/playwright/README.md
 /packages/playwright-core/api.json
 .env
+/tests/installation/output/
```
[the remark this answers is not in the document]
rwoll: It's where the tgz pack files from build_package.js go (https://github.com/microsoft/playwright/pull/13644/files#diff-91e819f6a2ca67ef1e1d34c98a0124fa5bd32bd02009247d9266aecec7fe11c2R36)

tests/installation/globalSetup.ts
```diff
@@ -0,0 +1,61 @@
...
+  const builds = await Promise.all([
+    build('playwright-core'),
+    build('playwright-test', '@playwright/test'),
+    build('playwright'),
+    build('playwright-chromium'),
+    build('playwright-firefox'),
+    build('playwright-webkit'),
+  ]);
+
+  await fs.promises.writeFile(path.join(__dirname, './registry.json'), JSON.stringify(Object.fromEntries(builds)));
```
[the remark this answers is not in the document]
rwoll: I'm going to leave as is. We still need the file on disk for globalSetup since it stores which packages its built already.

tests/installation/npm-installs-local-packages.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+test('installs local packages', async ({ registry, exec, tmpWorkspace }) => {
+  const packages = ['playwright', 'playwright-core', 'playwright-chromium', 'playwright-firefox', 'playwright-webkit', '@playwright/test'];
+  await exec('npm i --foreground-scripts', ...packages, { env: { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1' } });
+
+  const output = await exec('node', path.join(__dirname, '..', '..', 'utils', 'workspace.js'), '--get-version');
+  const expectedPlaywrightVersion = output.trim();
+  for (const pkg of packages) {
+    await test.step(`check version and installation location of ${pkg}`, async () => {
+      registry.assertLocalPackage(pkg);
```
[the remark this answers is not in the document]
rwoll: We do this on lines 29-33. `registry.assertLocalPackage()` just tracks that the registry fulfilled the HTTP Request. Does that make sense? Should we rename this function to reflect that?

tests/installation/driver-should-work.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { test } from './npmTest';
+
+test('codegen should work', async ({ exec }) => {
+  await exec('npm i --foreground-scripts playwright', { env: { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1' } });
+  await exec('npx playwright install', { env: { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '0' } });
+  await exec('node driver-client.js');
```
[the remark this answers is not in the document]
aslushnikov: this should be OK

tests/installation/driver-should-work.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { test } from './npmTest';
+
+test('codegen should work', async ({ exec }) => {
+  await exec('npm i --foreground-scripts playwright', { env: { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1' } });
+  await exec('npx playwright install', { env: { PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '0' } });
```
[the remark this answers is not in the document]
aslushnikov: this should be OK

tests/installation/npmTest.ts
```diff
@@ -0,0 +1,387 @@
...
+    _autoCopyScripts: void,
+    tmpWorkspace: string,
+    nodeMajorVersion: number,
+    installedBrowsers: () => Promise<string[]>;
+    writeFiles: (nameToContents: Record<string, string>) => Promise<void>,
+    exec: (cmd: string, ...argsAndOrOptions: ArgsOrOptions) => Promise<string>
+    tsc: (...argsAndOrOptions: ArgsOrOptions) => Promise<string>,
+    registry: Registry,
+        }>({
+          _autoCopyScripts: [async ({ tmpWorkspace }, use) => {
```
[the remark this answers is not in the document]
rwoll: Yes! Although, I prefer fixtures since (1) they get auto-applied, and (2) don't suffer from scoping issues if you have helpers/parameterized tests.

tests/installation/connect-to-selenium.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import path from 'path';
+import { test } from './npmTest';
+
+test('connect to selenium', async ({ exec, tmpWorkspace }) => {
+  await exec('npm i --foreground-scripts playwright-core');
+  await exec(`node ./download-chromedriver.js ${path.join(tmpWorkspace)}`);
+  await exec('npm run test -- --reporter=list selenium.spec', { cwd: path.join(__dirname, '..', '..'), env: { ...process.env, PWTEST_CHROMEDRIVER: path.join(tmpWorkspace, 'chromedriver') } });
```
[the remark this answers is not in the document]
rwoll: At the start of each test, a registry object store is created in itest's outputDir. When we run line 22, that also uses the same outputDir, and since it's a new process/test run, it goes to remove outputDir at the start (i.e. on line 22).

aslushnikov (approved):
shipit!

tests/installation/playwright-global-installation.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { test, expect } from './npmTest';
+
+test('global installation', async ({ exec, installedSoftwareOnDisk }) => {
+  const result = await exec('npm i --foreground-scripts playwright');
+  expect(result).toHaveLoggedSoftwareDownload(['chromium', 'ffmpeg', 'firefox', 'webkit']);
+  expect(await installedSoftwareOnDisk()).toEqual(['chromium', 'ffmpeg', 'firefox', 'webkit']);
```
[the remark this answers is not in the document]
rwoll: Added: https://github.com/microsoft/playwright/issues/13711

tests/installation/playwright-cli-install-should-work.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    expect(await installedSoftwareOnDisk()).toEqual(['chromium', 'ffmpeg']);
+  });
+
+  await test.step('playwright install', async () => {
+    const result = await exec('npx playwright install');
+    expect(result).toHaveLoggedSoftwareDownload(['firefox', 'webkit']);
+    expect(await installedSoftwareOnDisk()).toEqual(['chromium', 'ffmpeg', 'firefox', 'webkit']);
+  });
+
+  await exec('node sanity.js playwright none', { env: {  PLAYWRIGHT_BROWSERS_PATH: undefined } });
```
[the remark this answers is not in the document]
rwoll: https://github.com/microsoft/playwright/issues/13711

tests/installation/npmTest.ts
```diff
@@ -0,0 +1,172 @@
...
+                throw new Error(`Expected the command to exit cleanly (0 status code), but exited with ${result.code}.${message}`);
+              }
+
+              return stdio;
+
+            });
+          },
+          tsc: async ({ exec }, use) => {
+            await exec('npm i --foreground-scripts typescript@3.8 @types/node@14');
+            await use((...args: ArgsOrOptions) => exec('npx', '-p', 'typescript@3.8', 'tsc', ...args));
```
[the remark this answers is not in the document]
rwoll: Converged on versions.

tests/installation/playwright-xyz-should-work.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+import { test, expect } from './npmTest';
+
+for (const pkg of ['playwright-chromium', 'playwright-firefox', 'playwright-webkit']) {
+  test(`${pkg} should work`, async ({ exec, nodeMajorVersion, installedSoftwareOnDisk }) => {
+    const result = await exec('npm i --foreground-scripts', pkg);
+    const browserName = pkg.split('-')[1];
+    const expectedSoftware = [browserName];
+    if (browserName === 'chromium')
+      expectedSoftware.push('ffmpeg');
+    expect(result).toHaveLoggedSoftwareDownload(expectedSoftware as any);
```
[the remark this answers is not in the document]
rwoll: `expectedSoftware` is `string[]` while  `toHaveLoggedSoftwareDownload` expects an array of literals (`"chromium" | …`). The cast could have been more specific, but in this context, `any` is fine since the expect would start (runtime) yelling if the any is not suitale.

PR #13655, opened 2022-04-20
browser(webkit): roll to 2022/04/20

q66:
No 3-way diff, as the patch rebased cleanly and therefore there was no conflict resolution.

Build fixes:

* https://github.com/q66/WebKit/commit/3cab59a5a1babf1c4825acd37256f7a995f64e15 for ubuntu 18.04 (pending upstream)
* https://github.com/q66/WebKit/commit/55871e6f681fe28939e815c8011f5419413fb7dd (not upstreamable as it's a part of the playwright patch; just reflecting upstream API changes)

### DIFF
--- a/browser_patches/webkit/BUILD_NUMBER
+++ b/browser_patches/webkit/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1631
-Changed: dpino@igalia.com Thu Apr 14 06:07:27 UTC 2022
+1632
+Changed: dkolesa@igalia.com Wed Apr 20 03:58:57 PM CEST 2022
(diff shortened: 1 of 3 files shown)


q66:
It seems there is a build failure on ubuntu 18.04; do not merge yet, I have a local fix that I need to apply.

q66:
ok, now the build is successful; had to incorporate https://github.com/q66/WebKit/commit/c6723cbc4be3be4128a61ae84c99b6e8053989ee

q66:
(the issue is that the § character in https://github.com/q66/WebKit/blob/main/Source/WebKit/UIProcess/Automation/atoms/FindNodes.js#L123 is non-ascii, and it would seem older versions of python default to ascii encoding when reading/writing files in text mode)

q66:
upstream patch opened: https://bugs.webkit.org/show_bug.cgi?id=239542

yury-s (approved):

PR #13666, opened 2022-04-21
fix(runner): fail if worker cannot find some of the tests

### LINKED ISSUE #13611: [BUG] Playwright stalls before half of the test finished.
**Context:**
- Playwright Version: 1.21.0
- Operating System: Mac
- Node.js version: v14.19.0
- Browser: All

## System:
 - OS: macOS 12.3.1
 - Memory: 14.22 GB / 64.00 GB
## Binaries:
 - Node: 14.19.0 - ~/.nvm/versions/node/v14.19.0/bin/node
 - Yarn: 1.22.18 - /usr/local/bin/yarn
 - npm: 8.5.5 - ~/.nvm/versions/node/v14.19.0/bin/npm
## Languages:
 - Bash: 5.1.16 - /usr/local/bin/bash

**Code Snippet**

```javascript
// @ts-check
const {
  test,
  expect
} = require('@playwright/test');

const path = require('path');

const patternTool = require('../tools/get-test');
const patternsToTest = patternTool.getPatterns();
const testConfigJson = require('./localStorage.json');

let currentPattern;

const currentTheme = 'yellow'


test.beforeEach(async ({
  page
}) => {

  await page.goto(`${currentPattern.path}`);

  await page.addInitScript(value => {
    window.localStorage.setItem('availableThemes', value);
    window.sessionStorage.setItem('currentTheme', currentTheme);
  }, testConfigJson);

});



for (const currentTest in patternsToTest) {
  // console.log(currentPattern);

  currentPattern = patternsToTest[currentTest];

  test.describe(`Theme ${currentTheme}`, () => {

    test(`Test: ${currentTest} - ${currentPattern.name}`, async ({
      page
    }) => {

      // await page.waitForLoadState('domcontentloaded');

      const screenshotPath = `./screenshots/${currentTheme}/${path.basename(currentPattern.path, '.html')}.png`

      expect(await page.screenshot()).toMatchSnapshot(screenshotPath, {
        threshold: 0.01,
        maxDiffPixelRatio: 0.1
      });

    });


  });

};
```

**Describe the bug**

After nearly having the test seems to stall and I can wait for hours. Nothing moving forward anymore.

Screenshot 2022-04-18 at 18 15 46

Is the bug on my side or is it really a bug.

Try to test a style guide and various themes. The style guide is based on https://patternlab.io
Assets I like to test can be found here: https://lab.n8d.studio/htwoo/htwoo-core/?p=all

yury-s:
Fixes #13611

### DIFF
--- a/packages/playwright-test/src/workerRunner.ts
+++ b/packages/playwright-test/src/workerRunner.ts
@@ -157,6 +157,7 @@ export class WorkerRunner extends EventEmitter {
   async runTestGroup(runPayload: RunPayload) {
     this._runFinished = new ManualPromise<void>();
     const entries = new Map(runPayload.entries.map(e => [ e.testId, e ]));
+    let fatalUnknownTestIds;
     try {
       await this._loadIfNeeded();
       const fileSuite = await this._loader.loadTestFile(runPayload.file, 'worker');
@@ -179,7 +180,8 @@ export class WorkerRunner extends EventEmitter {
           await this._runTest(tests[i], entry.retry, tests[i + 1]);
         }
       } else {
-        this.emit('unknowTest', runPayload);
+        fatalUnknownTestIds = runPayload.entries.map(e => e.testId);
+        this.stop();
       }
     } catch (e) {
       // In theory, we should run above code without any errors.
@@ -190,6 +192,7 @@ export class WorkerRunner extends EventEmitter {
       const donePayload: DonePayload = {
         fatalErrors: this._fatalErrors,
         skipTestsDueToSetupFailure: [],
+        fatalUnknownTestIds
       };
       for (const test of this._skipRemainingTestsInSuite?.allTests() || []) {
         if (entries.has(test._id))
--- a/packages/playwright-test/src/workerRunner.ts
+++ b/packages/playwright-test/src/workerRunner.ts
@@ -178,6 +178,8 @@ export class WorkerRunner extends EventEmitter {
           entries.delete(tests[i]._id);
           await this._runTest(tests[i], entry.retry, tests[i + 1]);
         }
+      } else {
+        this.emit('unknowTest', runPayload);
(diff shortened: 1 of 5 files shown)


packages/playwright-test/src/workerRunner.ts
```diff
@@ -179,6 +179,8 @@ export class WorkerRunner extends EventEmitter {
           entries.delete(tests[i]._id);
           await this._runTest(tests[i], entry.retry, tests[i + 1]);
         }
+      } else {
```
dgozman, line 182: - Why don't we report an error if any of the tests is missing, instead of when all the tests are missing?
- This should stop testing and trigger teardowns, similar to fatal error. See [this block](https://github.com/microsoft/playwright/blob/8c4fc62e8f601c81c16fc382f37999dfd4ce1213/packages/playwright-test/src/workerRunner.ts#L141-L146). Btw, perhaps we can just issue a fatal error instead of a new ipc?

yury-s: > Why don't we report an error if any of the tests is missing, instead of when all the tests are missing?

We could do that but it would require maintaining another set of not-found tests every time we perform filtering and the end result would be the same as currently the process converges with only unknown tests in the worker and we fail at that point. Given that it is very rare edge case let's keep the code simple.


> This should stop testing and trigger teardowns, similar to fatal error. See [this block](https://github.com/microsoft/playwright/blob/8c4fc62e8f601c81c16fc382f37999dfd4ce1213/packages/playwright-test/src/workerRunner.ts#L141-L146). Btw, perhaps we can just issue a fatal error instead of a new ipc?

Added `fatalUnknownTestIds` to `DonePayload` so that we can format test names in the runner (the worker doesn't know about those tests).

dgozman (approved):

PR #13678, opened 2022-04-21
chore: make tests pass on Node.js 17+

mxschmitt:


packages/playwright-core/src/common/socksProxy.ts
```diff
@@ -375,7 +375,7 @@ export class SocksProxyHandler extends EventEmitter {
 
   async socketRequested({ uid, host, port }: SocksSocketRequestedPayload): Promise<void> {
     if (host === 'local.playwright')
-      host = 'localhost';
```
yury-s, line 378: This may backfire if the host does custom mapping for `localhost` name, maybe we can update our test server instead?

packages/playwright-core/src/server/fetch.ts
```diff
@@ -360,15 +360,15 @@ export abstract class APIRequestContext extends SdkObject {
...
-              reject(new Error(`failed to decompress '${encoding}' encoding: ${e}`));
-          });
+          body = response.pipe(emptyStreamTransform).pipe(transform);
         }
 
         body.on('data', chunk => chunks.push(chunk));
         body.on('end', notifyBodyFinished);
-        body.on('error', reject);
+        body.on('error', e => {
+          if (e)
```
yury-s, line 369: Let's revert changes in this file as now  we say 'Failed to decompress' on any error even if the response is not compressed.

Does Node 17 remove `pipeline` from its API?

tests/library/global-fetch-cookie.spec.ts
```diff
@@ -143,10 +143,10 @@ it('should send secure cookie over http for localhost', async ({ request, server
     res.setHeader('Set-Cookie', ['a=v; secure', 'b=v']);
     res.end();
   });
-  await request.get(`${server.PREFIX}/setcookie.html`);
+  await request.get(`${server.CROSS_PROCESS_PREFIX}/setcookie.html`);
   const [serverRequest] = await Promise.all([
     server.waitForRequest('/empty.html'),
-    request.get(server.EMPTY_PAGE)
+    request.get(server.CROSS_PROCESS_PREFIX + '/empty.html')
```
yury-s, line 149: This should fail I believe as we only check for 'localhost' in [the code](https://github.com/microsoft/playwright/blob/8c4fc62e8f601c81c16fc382f37999dfd4ce1213/packages/playwright-core/src/server/cookieStore.ts#L31).

packages/playwright-test/src/webServer.ts
```diff
@@ -93,7 +93,7 @@ export class WebServer {
     debugWebServer(`Waiting for availability...`);
     await this._waitForAvailability();
     debugWebServer(`WebServer available`);
-    if (this.config.port !== undefined)
+    if (this.config.port !== undefined && !process.env.PLAYWRIGHT_TEST_DONT_SET_BASE_URL)
```
yury-s, line 96: Can we check instead if PLAYWRIGHT_TEST_BASE_URL is already set?

yury-s (approved):

PR #13701, opened 2022-04-22
feat: support tip-of-tree chromium channel

aslushnikov:


.github/workflows/tests_secondary.yml
```diff
@@ -304,6 +304,83 @@ jobs:
         name: chrome-stable-mac-test-results
         path: test-results
 
+  chromium_tot_linux:
```
yury-s, line 307: Can we make it a matrix?

yury-s (approved):

.github/workflows/tests_secondary.yml
aslushnikov: in a follow-up!

PR #13703, opened 2022-04-22
tests: use ::1 instead of 127.0.0.2

yury-s:
127.0.0.2+ is not working on macos.


aslushnikov (approved):

rwoll:
@yury-s Will this work in Docker for mac? I don't think it supports v6 stack fully (although, since this is a special address it may work).

yury-s:
> @yury-s Will this work in Docker for mac? I don't think it supports v6 stack fully (although, since this is a special address it may work).

I didn't check, closing this PR in favor of just deleting the test https://github.com/microsoft/playwright/pull/13705

PR #13707, opened 2022-04-22
chore: use helper functions to define platforms

aslushnikov:
This is to make code less error-prone due to involved windows
detection.

### DIFF
--- a/browser_patches/firefox/build.sh
+++ b/browser_patches/firefox/build.sh
@@ -67,7 +67,7 @@ else
   echo "ac_add_options --enable-release" >> .mozconfig
 fi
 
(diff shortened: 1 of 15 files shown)


browser_patches/firefox/build.sh
```diff
@@ -67,7 +67,7 @@ else
   echo "ac_add_options --enable-release" >> .mozconfig
 fi
 
-if [[ "$(uname)" == MINGW* || "$(uname)" == "Darwin" || "$(uname)" == MSYS* ]]; then
+if is_win || is_linux; then
```
aslushnikov, line 70:
```suggestion
if is_win || is_mac; then
```

PR #13717, opened 2022-04-24
chore: use plugin story for components

pavelfeldman:


packages/playwright-ct-vue/register.mjs
```diff
@@ -29,6 +29,7 @@ export default (components, options) => {
 const allListeners = [];
 
 function render(component) {
+console.log(component);
```
dgozman, line 32:
```suggestion
```

packages/playwright-test/src/runner.ts
```diff
@@ -450,15 +444,23 @@ export class Runner {
       }, result);
 
       await this._runAndReportError(async () => {
-        for (const internalGlobalTeardown of internalGlobalTeardowns)
-          await internalGlobalTeardown();
+        await Promise.all(pluginTeardowns.map(t => t()));
```
dgozman, line 447: Why run setup serially, but teardown in parallel?

dgozman (approved):

Issue #10177, opened 2021-11-09
[Feature] WithAttribute function to chain elements on the same level with different attributes


dz1dz1on:
Right now we need to repeat lines of code to get to the element with more than one attribute.

We would like to have an implementation that will help us with defining for example elements and then filter them by additional attribute:

**[Example of what we want]**:

![image](https://user-images.githubusercontent.com/62831238/140914358-4d43bb6a-e212-49b1-84f4-3d6023561272.png)

**[What we have?]**
Right now we need to duplicate part of the standardProducts:

```
async chooseStandardProduct(dataProductName: string): Promise<void> {
   await this.page.locator(`[e2e-target-name="standard"][e2e-target-name="${dataProductName}"`]).click()
}
```

Thanks to that solution we can decrease code duplication. Similar solution already have testcafe.

pavelfeldman:
#### Why was this issue closed?

We are prioritizing the features based on the upvotes, recency and activity in the issue thread. It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it. We are closing issues like this one to keep our bug database maintainable. Please feel free to open a new issue and link this one to it if you think this is a mistake.

PR #13731, opened 2022-04-25
feat: locator.that

### LINKED ISSUE #12116: [Feature] Improvment of has option from the build 1.19
***Let us know what functionality you'd like to see in Playwright and what your use case is.***

Actual implementation of `has` is strictly connected with passed locator. 
We need to write page.locator and then pass option has. It feels like repetitive work to do. 

`page.locator('.cssClass', {has:'[e2e-target="search"]'})`

I think the better solution would be to add `has` (or as hasAttribute) as function passed by dot notation to the locator.(`page.locator().has()`)

Thanks to that solution we've could get such syntax without repeating every time :

```
Right now: 

get() addressButtonElement {
    page.locator('.cssClass', {has: '#button [e2e-target="address"]'} )
}

get() exitButtonElement {
        page.locator('.cssClass', {has: '#button [e2e-target="exit"]'} )
}
----------------------------------------------------------------
Proposition:

const parentElement =  page.locator('.cssClass').has('#button')

get() addressButtonElement {
    this.parentElement.has('[e2e-target="address"]')
}

get() exitButtonElement {
    this.parentElement.has('[e2e-target="exit"]')
}
```

***Do you think others might benefit from this as well?***

Thanks to `has function` for locator we can easily apply this method everywhere in the page object (where we have already declared locators as a variables or like below). It also can help with distinguishing between elements in the same div but with different attributes.

```
type InputTextType = 'color' | 'description'

class Example {

$ = {
    modalExitButton: this.parentElement.has('[e2e-target="exit"]'), <- elements in the same div
    modalAcceptButton:  this.parentElement.has('[e2e-target="accept"]'),  <- elements in the same div
    input: this.page.locator('#text'),
}

constructor(page: Page, page.locator('.abcModal')) {
    super(page)
}

async function fillInput(type: InputTextType, text): Promise<void> {
   await this.$.input.has('[e2e-target="${type}"]').fill(text)
}

}
```

dgozman:
Filters existing locator by options, currently `has` and `hasText`.

Fixes #12116. References #10177.

### DIFF
--- a/docs/src/api/class-locator.md
+++ b/docs/src/api/class-locator.md
@@ -770,6 +770,13 @@ Returns the `node.textContent`.
 
 ### option: Locator.textContent.timeout = %%-input-timeout-%%
 
+## method: Locator.that
+- returns: <[Locator]>
+
+This method narrows existing locator according to the options, for example filters by text.
+
+### option: Locator.that.-inline- = %%-locator-options-list-%%
+
 ## async method: Locator.type
 
 Focuses the element, and then sends a `keydown`, `keypress`/`input`, and `keyup` event for each character in the text.
--- a/docs/src/locators.md
+++ b/docs/src/locators.md
@@ -67,10 +67,69 @@ await locator.HoverAsync();
 await locator.ClickAsync();
 ```
 
+## Creating Locators
+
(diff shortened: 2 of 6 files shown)


docs/src/api/class-locator.md
```diff
@@ -770,6 +770,13 @@ Returns the `node.textContent`.
 
 ### option: Locator.textContent.timeout = %%-input-timeout-%%
 
+## method: Locator.that
```
rwoll, line 773: This word was quite confusing to me until I saw the examples later on in the PR. It seems like this just means `and`. Let's consider renaming it to `and` since I think Boolean logic words will have more straightforward meaning/explanations.

pavelfeldman (approved):

docs/src/api/class-locator.md
dgozman: TBD at API review!

PR #13744, opened 2022-04-25
fix: installation tests on Node.js 18

rwoll:


rwoll:
@aslushnikov Looks like we missed a spot! Merged so we don't have extra failures on main.

PR #13746, opened 2022-04-25
feat(codegen): pytest support in codegen

roniemartinez:
This is a proposal to add support for `pytest` in the code generator

NOTE: 
- ~~Everything here are roughly a copy-pasted from the `PythonLanguageGenerator`. Hope it gains approval from community.~~
- Pytest async is not currently supported ~~so this can be removed/turned off~~ (https://github.com/microsoft/playwright-pytest/issues/74)

image

### DIFF
--- a/packages/playwright-core/src/server/recorder/python.ts
+++ b/packages/playwright-core/src/server/recorder/python.ts
@@ -156,9 +156,19 @@ export class PythonLanguageGenerator implements LanguageGenerator {
   generateHeader(options: LanguageGeneratorOptions): string {
     const formatter = new PythonFormatter();
     if (this._isTest) {
-      formatter.add(`
+      formatter.add(`${options.deviceName ? 'import pytest\n' : ''}
 from playwright.sync_api import Page, expect
+${options.deviceName ? `
 
+@pytest.fixture(scope="session")
+def browser_context_args(browser_context_args, playwright) {
+    device = playwright.devices["${options.deviceName}"]
+    return dict(
+        **browser_context_args,
+        **device,
+    )
+}
+` : ''}
 
 def test_example(page: Page) -> None {`);
     } else if (this._isAsync) {
--- a/packages/playwright-core/src/server/recorder/python.ts
+++ b/packages/playwright-core/src/server/recorder/python.ts
@@ -33,17 +33,22 @@ export class PythonLanguageGenerator implements LanguageGenerator {
   private _awaitPrefix: '' | 'await ';
   private _asyncPrefix: '' | 'async ';
   private _isAsync: boolean;
+  private _isTest: boolean;
 
-  constructor(isAsync: boolean) {
-    this.id = isAsync ? 'python-async' : 'python';
-    this.fileName = isAsync ? 'Python Async' : 'Python';
+  constructor(isAsync: boolean, isTest: boolean) {
+    this.id = isTest ? 'pytest' : (isAsync ? 'python-async' : 'python');
+    this.fileName = isTest ? 'Pytest' : (isAsync ? 'Python Async' : 'Python');
     this._isAsync = isAsync;
+    this._isTest = isTest;
     this._awaitPrefix = isAsync ? 'await ' : '';
     this._asyncPrefix = isAsync ? 'async ' : '';
   }
 
   generateAction(actionInContext: ActionInContext): string {
     const action = actionInContext.action;
+    if (this._isTest && (action.name === 'openPage' || action.name === 'closePage'))
+      return '';
+
     const pageAlias = actionInContext.frame.pageAlias;
     const formatter = new PythonFormatter(4);
     formatter.newLine();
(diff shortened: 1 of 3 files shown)


pavelfeldman:
I'm supportive for adding this kind of the codegen. I would not copy and paste though - take a look at a sibling `javascript.ts` with its `_isTest` variable. You merely need to adjust the header and footer based on whether you generate the test, and the rest of the logic will stay the same.

roniemartinez:
@pavelfeldman I updated in 56ebcaf20483e423283a0b4950d5f29e6f283a9e based on your recommendation. Should I also add tests? The Javascript test does not have this either.

dgozman:
> Should I also add tests? The Javascript test does not have this either.

Take a look at [this file](https://github.com/microsoft/playwright/blob/3aba94dbef8b8d29901ec13d441cd94e071b6dcc/tests/library/inspector/cli-codegen-test.spec.ts) - it is testing the javascript implementation.

packages/playwright-core/src/server/recorder/python.ts
```diff
@@ -150,7 +155,13 @@ export class PythonLanguageGenerator implements LanguageGenerator {
 
   generateHeader(options: LanguageGeneratorOptions): string {
     const formatter = new PythonFormatter();
-    if (this._isAsync) {
+    if (this._isTest) {
+      formatter.add(`
```
dgozman, line 159: Let's also configure device emulation by overriding `browser_context_args` fixture, similar to javascript? [Here is an example](https://playwright.dev/python/docs/test-runners#device-emulation).

dgozman (commented):
This looks great! If you would make device emulation configurable, that would be just wonderful! If you don't have cycles for that, just let us know and we'll follow up.

packages/playwright-core/src/server/recorder/python.ts
roniemartinez: @dgozman I find it tricky to create the curly braces for dictionaries and the double indentation (8 spaces). Any advice.

image

packages/playwright-core/src/server/recorder/python.ts
```diff
@@ -33,17 +33,22 @@ export class PythonLanguageGenerator implements LanguageGenerator {
   private _awaitPrefix: '' | 'await ';
   private _asyncPrefix: '' | 'async ';
   private _isAsync: boolean;
+  private _isTest: boolean;
 
-  constructor(isAsync: boolean) {
-    this.id = isAsync ? 'python-async' : 'python';
-    this.fileName = isAsync ? 'Python Async' : 'Python';
+  constructor(isAsync: boolean, isTest: boolean) {
```
mxschmitt, line 38:
```suggestion
  constructor(isAsync: boolean, isPyTest: boolean) {
```

packages/playwright-core/src/server/recorder/python.ts
```diff
@@ -33,17 +33,22 @@ export class PythonLanguageGenerator implements LanguageGenerator {
...
+    this.id = isTest ? 'pytest' : (isAsync ? 'python-async' : 'python');
+    this.fileName = isTest ? 'Pytest' : (isAsync ? 'Python Async' : 'Python');
     this._isAsync = isAsync;
+    this._isTest = isTest;
```
mxschmitt, line 42:
```suggestion
    this._isPyTest = isPyTest;
```

packages/playwright-core/src/server/recorder/python.ts
```diff
@@ -173,7 +194,9 @@ def run(playwright: Playwright) -> None {
   }
 
   generateFooter(saveStorage: string | undefined): string {
-    if (this._isAsync) {
+    if (this._isTest) {
+      return '';
+    } else if (this._isAsync) {
```
mxschmitt, line 199:
```suggestion
    if (this._isTest)
      return '';
    if (this._isAsync) {
```

packages/playwright-core/src/server/recorder/python.ts
dgozman: Yeah, that's indeed tricky. Let me merge this PR and I can follow up, because it might require some internal refactoring to make that work.

dgozman (approved):
Thank you for the PR!

tests/library/inspector/cli-codegen-pytest.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+def test_example(page: Page) -> None:
+
+    # Go to ${emptyHTML}
+    page.goto("${emptyHTML}")
+`);
+});
+
+test('should save the codegen output to a file if specified', async ({ runCLI }, testInfo) => {
+  const tmpFile = testInfo.outputPath('script.js');
```
mxschmitt, line 62:
```suggestion
  const tmpFile = testInfo.outputPath('test_foo.py');
```
nit

mxschmitt (approved):

PR #13776, opened 2022-04-26
feat(poll): expose custom poll interval

### LINKED ISSUE #13775: [Feature] Allow to define an interval / wait period between expect.poll assertions
In one of the latest versions of Playwright the `expect.poll`-function was introduced to allow to poll for a specific assertion. This is really a great addition to Playwright. 

Only I would like to request an enhancement to this functionality to delay a breathing period between assertion checks. For example, when you are querying for a status change via an API endpoint then it can be beneficial to wait a short period between assertion checks to not hit the API server too often or it might not beneficial to continuously execute this assertion.

In my case I would like to check if a revocation of permissions has completed, and it doesn't make sense to check the assertion of a period of 60 seconds continuously. A bit of breathing or waiting time of like 5 seconds would help to ease everything

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/13775

### DIFF
--- a/packages/playwright-test/src/expect.ts
+++ b/packages/playwright-test/src/expect.ts
@@ -96,7 +96,7 @@ export const printReceivedStringContainExpectedResult = (
 
 // #endregion
 
-type ExpectMessageOrOptions = undefined | string | { message?: string, timeout?: number };
+type ExpectMessageOrOptions = undefined | string | { message?: string, timeout?: number, intervals?: number[] };
 
 function createExpect(actual: unknown, messageOrOptions: ExpectMessageOrOptions, isSoft: boolean, isPoll: boolean, generator?: Generator) {
   return new Proxy(expectLibrary(actual), new ExpectMetaInfoProxyHandler(messageOrOptions, isSoft, isPoll, generator));
@@ -153,6 +153,7 @@ type ExpectMetaInfo = {
   isSoft: boolean;
   isPoll: boolean;
   pollTimeout?: number;
+  pollIntervals?: number[];
   generator?: Generator;
 };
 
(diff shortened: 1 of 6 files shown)


packages/playwright-test/src/expect.ts
```diff
@@ -248,10 +250,11 @@ class ExpectMetaInfoProxyHandler {
   }
 }
 
-async function pollMatcher(matcherName: any, isNot: boolean, timeout: number, generator: () => any, ...args: any[]) {
+async function pollMatcher(matcherName: any, isNot: boolean, pollIntervals: number[] | undefined, timeout: number, generator: () => any, ...args: any[]) {
   let matcherError;
   const startTime = monotonicTime();
-  const pollIntervals = [100, 250, 500];
+  pollIntervals = pollIntervals || [100, 250, 500, 1000];
```
aslushnikov, line 256:
```suggestion
  pollIntervals = pollIntervals ?? [100, 250, 500, 1000];
```

packages/playwright-test/src/expect.ts
```diff
@@ -248,10 +250,11 @@ class ExpectMetaInfoProxyHandler {
...
+  const lastPollInterval = pollIntervals[pollIntervals.length - 1] || 1000;
```
aslushnikov, line 257: I assume if someone passes [0] as poll intervals we want to respect it
```suggestion
  const lastPollInterval = pollIntervals[pollIntervals.length - 1] ?? 1000;
```

aslushnikov (approved):

packages/playwright-test/src/expect.ts
pavelfeldman: This should be the same thing.

packages/playwright-test/src/expect.ts
pavelfeldman: I'm not sure we ever want to see 0 here? That'd be a denial of service more or less...

PR #13807, opened 2022-04-28
feat(plugins): expose suite to plugin setup

pavelfeldman:


yury-s (approved):

rwoll (approved):
LGTM! We'll need `GlobalInfo` in setup for the VCS Plugin, so setup will also have an `info: GlobalInfo` param.

PR #13825, opened 2022-04-28
test: log more phases of plugins

rwoll:


tests/playwright-test/plugins.spec.ts
```diff
@@ -64,10 +69,15 @@ test('event order', async ({ runInlineTest }, testInfo) => {
   expect(result.passed).toBe(1);
   const logLines = await fs.promises.readFile(log, 'utf8');
   expect(logLines.split('\n')).toEqual([
+    'a configure',
+    'b configure',
     'a setup',
     'b setup',
     'globalSetup',
+    'a configure',
+    'b configure',
```
rwoll, line 78: I believe we knew this duplicate configure call would be invoked since the config is loaded first in the Runner Process and then in each Worker.

@pavelfeldman: I think we should fix the Runner/Worker so it only calls configure once and only in Runner process.

pavelfeldman: Yes, we should fix this

pavelfeldman (approved):

PR #13831, opened 2022-04-28
fix(core): use import for whatwg url usage

redabacha:
Useful when working with playwright-core in an environment like jsdom

### DIFF
--- a/packages/playwright-core/src/client/clientHelper.ts
+++ b/packages/playwright-core/src/client/clientHelper.ts
@@ -17,6 +17,7 @@
 
 import type * as types from './types';
 import fs from 'fs';
+import { URL } from 'url';
 import { isString, isRegExp, constructURLBasedOnBaseURL } from '../utils';
 
 export function envObjectToArray(env: types.Env): { name: string, value: string }[] {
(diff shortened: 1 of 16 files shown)


dgozman (commented):
Thank you for the PR! Could you please tell us more about your usecase?

I am worried that this is not sustainable going forward, because any PR could start using `URL` without an import again and there are no checks to prevent that.

redabacha:
> Thank you for the PR! Could you please tell us more about your usecase?
> 
> 
> 
> I am worried that this is not sustainable going forward, because any PR could start using `URL` without an import again and there are no checks to prevent that.
> 
> 

hi @dgozman,

essentially trying to use the playwright-core package in jest can be quite tricky under the jsdom environment. one particular use case i have for this currently is using storybook's storyshots solution w/ playwright to create a visual regression testing platform. 

yes that's a very valid point, i was thinking to add the no-restricted-global eslint rule to catch this https://eslint.org/docs/rules/no-restricted-globals. does that sound like an okay idea?

dgozman:
@redabacha We've seen many folks successfully running Playwright in Jest. Could you please file an issue with a repro? This is probably some edge case or some quirky setup issue.

pavelfeldman:
You can't use Playwright in jsdom environment, it requires Node environment. As per the [docs](https://nodejs.org/api/url.html#class-url), URL is available on the global object as of Node 10, so we need environment that respects that.

PR #13835, opened 2022-04-28
chore: pass config w/ cli overrides to plugins

pavelfeldman:


packages/playwright-test/src/cli.ts
```diff
@@ -132,9 +132,10 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
     return;
 
   const runner = new Runner(overrides);
-  const config = resolvedConfigFile ? await runner.loadConfigFromResolvedFile(resolvedConfigFile) : await runner.loadEmptyConfig(configFileOrDirectory);
-  if (('projects' in config) && opts.browser)
-    throw new Error(`Cannot use --browser option when configuration file defines projects. Specify browserName in the projects instead.`);
```
rwoll, line 137: Why did this go away? I don't see us doing it in `validateConfig`, so this changes behavior.

rwoll (approved):
Overall, LGTM. See one comment to make sure we are intentionally changing error condition.

packages/playwright-test/src/cli.ts
pavelfeldman: It was moved, this logic is still there.

PR #13836, opened 2022-04-29
feat: rewrite VCS Plugin

rwoll:
- uses metadata instead of GlobalInfo/attachments
- allows user to manually specify info via the plugin
- drops GlobalInfo+Attachments from the codebase (yes, this probably should have been a distinct PR)


rwoll:
Need to do some major re-basing it looks like!

PR #13837, opened 2022-04-29
feat: rewrite gitCommitInfo plugin, drop GlobalInfo & attachments

rwoll:
- plumb top-level metadata through
- use metadata instead of GlobalInfo/attachments for gitCommitInfo
- drop GlobalInfo/attachments from the codebase (Yes, this should have been a distinct PR…)
- rewrite the gitCommitInfo Plugin, which either uses git CLI if no manual values are provided, or uses manual value exclusively
- add typing to metadata


packages/html-reporter/src/hideIfErrors.tsx
```diff
@@ -13,19 +13,20 @@
  * See the License for the specific language governing permissions and
```
rwoll, line 13: This diff is odd. GitHub thinks the file moved. But really: `globalSetup.ts` was deleted, and `hideIfErrors.tsx` was added. They are not related at all, and hideIfErrors.tsx is net new, so just look at the file, not the diff.

packages/playwright-test/src/loader.ts
```diff
@@ -131,6 +131,7 @@ export class Loader {
     this._fullConfig.workers = takeFirst(this._configOverrides.workers, config.workers, baseFullConfig.workers);
     this._fullConfig.webServer = takeFirst(this._configOverrides.webServer, config.webServer, baseFullConfig.webServer);
     this._fullConfig._plugins = takeFirst(this._configOverrides.plugins, config.plugins, baseFullConfig._plugins);
+    this._fullConfig.metadata = mergeObjects(this._configOverrides.metadata, mergeObjects(config.metadata, baseFullConfig.metadata));
```
rwoll, line 134: `metadata` is an `any`-typed thing in our public API. We can't change that. So, better hope this doesn't try to merge a number and an object or array!

Not sure what to do here…other then maybe some nicer error message in the case of explosions. reviewers: wdyt?

tests/library/har.spec.ts
```diff
@@ -38,6 +38,7 @@ async function pageWithHar(contextFactory: (options?: BrowserContextOptions) =>
 }
 
 it('should throw without path', async ({ browser }) => {
+  expect(1).toBe(2);
```
rwoll, line 41:
```suggestion
```

packages/html-reporter/src/hideIfErrors.tsx
```diff
@@ -13,19 +13,20 @@
...
  * limitations under the License.
  */
-import type { FullConfig, GlobalInfo } from '@playwright/test';
+import * as React from 'react';
 
-// We're dogfooding this, so the …/lib/… import is acceptable
-import * as ci from '@playwright/test/lib/ci';
+export class HideIfErrors extends React.Component<{}, { hitError: boolean }> {
+  state = {
```
pavelfeldman, line 19: We don't use legacy components, only FCs w/ useState.

packages/html-reporter/src/index.tsx
```diff
@@ -26,58 +26,24 @@ import { ReportView } from './reportView';
...
+  link: string;
+}
 
-  if (Object.entries(out).filter(([_, v]) => v !== undefined).length)
-    return out;
-};
+export type Metadata = {
+  generatedAt?: number;
+  ci?: Partial<CI>;
+  revision?: Partial<Revision>;
```
pavelfeldman, line 45: Unless you use full Revision somewhere, I'd put ? everywhere inside it.

packages/html-reporter/src/metadataView.tsx
```diff
@@ -24,59 +24,61 @@ import './reportView.css';
 import './theme.css';
 
 export const MetadataView: React.FC<Metadata> = metadata => {
+  if (!metadata.ci?.link && !metadata.revision?.id && !metadata.revision?.author && !metadata.revision?.email && !metadata.revision?.subject && !metadata.revision?.timestamp && !metadata.revision?.link && !metadata.generatedAt)
```
pavelfeldman, line 27: My gut tells me that we should make these flat with the named keys:
metadata['commit.author'] = ...

I think the list will grow beyond type-able once all kinds of Jira IDs get there...

packages/html-reporter/src/reportView.tsx
```diff
@@ -46,7 +47,7 @@ export const ReportView: React.FC<{
   return <div className='htmlreport vbox px-4 pb-4'>
     <main>
       {report?.json() && <HeaderView stats={report.json().stats} filterText={filterText} setFilterText={setFilterText}></HeaderView>}
-      {report?.json().metadata && <MetadataView {...report?.json().metadata!} />}
+      {report?.json().metadata && <HideIfErrors><MetadataView {...report?.json().metadata!} /></HideIfErrors>}
```
pavelfeldman, line 50: What is <HideIfErrors> used for here?

packages/playwright-test/src/loader.ts
pavelfeldman, line 134: this._configOverrides can't override the metadata, so we don't need this line. My recent patches make it clear via explicit type for the overrides.

packages/playwright-test/src/loader.ts
```diff
@@ -478,6 +479,7 @@ export const baseFullConfig: FullConfigInternal = {
   grep: /.*/,
   grepInvert: null,
   maxFailures: 0,
+  metadata: undefined,
```
pavelfeldman, line 482: Let's give them an empty object.

packages/playwright-test/src/plugins/index.ts
```diff
@@ -14,3 +14,4 @@
  * limitations under the License.
  */
 export { webServer } from './webServerPlugin';
+export { vcs } from './vcsPlugin';
```
pavelfeldman, line 17: I don't think your plugin helps those that are not using Git much. How about call it GitCommitInfo (or GitCommitInfo)?

packages/playwright-test/src/plugins/vcsPlugin.ts
```diff
@@ -0,0 +1,130 @@
...
+      if (shouldRunGit(options.revision, options.mode)) {
+        revision = {
+          ...await gitStatusFromCLI(options.directory || configDir),
+          ...revision,
+        };
+      }
+      const links = linksFromEnv();
+      // Merge it all together minding the order of precedence
+      // Be sure to mutate the config passed in and not return a new copy
+      config.metadata = config.metadata || {};
```
pavelfeldman, line 38: Yeah, definitely better to have {} there by default!

packages/playwright-test/src/plugins/vcsPlugin.ts
```diff
@@ -0,0 +1,130 @@
...
+import { createGuid } from 'playwright-core/lib/utils';
+import { spawnAsync } from 'playwright-core/lib/utils/spawnAsync';
+
+const GIT_OPERATIONS_TIMEOUT_MS = 1500;
+
+export const vcs = (options?: VCSPluginOptions): TestPlugin => {
+  return {
+    name: 'playwright-vcs-plugin',
+
+    configure: async (config: PlaywrightTestConfig, configDir: string) => {
```
pavelfeldman, line 26: You don't need to be PlaywrightTestConfig-specific, make it Config.

packages/playwright-test/src/plugins/vcsPlugin.ts
```diff
@@ -0,0 +1,130 @@
...
+      config.metadata.ci = {
+        ...config.metadata?.ci,
+        link: links.ci,
+        ...options.ci,
+      };
+
+      // Normalize dates
+      const timestamp = config.metadata.revision?.timestamp;
+      if (timestamp instanceof Date)
+        config.metadata.revision.timestamp = timestamp.getTime();
```
pavelfeldman, line 54: Unix time in millis is not terribly popular, so needs to be documented as such or changed.

packages/playwright-test/src/plugins/vcsPlugin.ts
```diff
@@ -0,0 +1,130 @@
...
+};
+
+export type VCSPluginOptions = {
+    mode?: 'disable-cli',
+    directory?: string,
+    revision?: Partial<Revision>,
+    ci?: Partial<CI>,
+};
+
+interface Revision {
```
pavelfeldman, line 67: This looks familiar, but it makes sense here! Also not a fan of partial.

packages/playwright-test/src/plugins/vcsPlugin.ts
```diff
@@ -0,0 +1,130 @@
...
+const GIT_OPERATIONS_TIMEOUT_MS = 1500;
+
+export const vcs = (options?: VCSPluginOptions): TestPlugin => {
+  return {
+    name: 'playwright-vcs-plugin',
+
+    configure: async (config: PlaywrightTestConfig, configDir: string) => {
+      options = options || {};
+      let revision: Partial<Revision> | undefined = options.revision;
+      if (shouldRunGit(options.revision, options.mode)) {
```
pavelfeldman, line 29: Without this, plugin does not add much value. Let's make it git-only and allow room for non-git plugins. Otherwise we claim vcs and fall short.

packages/playwright-test/types/test.d.ts
```diff
@@ -1055,6 +1055,10 @@ export interface FullConfig<TestArgs = {}, WorkerArgs = {}> {
    *
    */
   maxFailures: number;
+  /**
+   * Any JSON-serializable metadata that will be put directly to the test report.
+   */
+  metadata: any;
```
pavelfeldman, line 1061: Ah, so it is new, make it { [key: string]: string | number | boolean | { content: string|Buffer, contentType } } ?

packages/html-reporter/src/hideIfErrors.tsx
rwoll: Can you do React Error Boundaries with an FC? (The `componentDidCatch` is the special bit. The official docs only show class example.)

packages/playwright-test/types/test.d.ts
rwoll: It's not new in TestConfig which is where it come from. We can change it there, too, though.

packages/html-reporter/src/reportView.tsx
rwoll: https://github.com/microsoft/playwright/pull/13837/files#diff-85123fb439c927ffac44a1e35b5fea6df9e24ecbe03b2575f65f41dea1a752bcR841

rwoll: If someone manually puts weirdness into the metadata field via the config (which they can already use and access today), we don't want to crash the HTML Report.

packages/playwright-test/src/plugins/vcsPlugin.ts
rwoll: I flattened it here, too to see if that made sense. We'll judge in the next patchset.

packages/html-reporter/src/hideIfErrors.tsx
rwoll: > Only class components can be error boundaries.

(https://reactjs.org/docs/error-boundaries.html)

so we must use a class-based component here.

pavelfeldman: Overall, throughout the code, we try avoiding silent exceptions. We also don't try/catch our own code, we prefer to handle the error where it happens instead. This is basically a silent exception in the React world, let's not use it.

packages/html-reporter/src/metadataView.tsx
```diff
@@ -18,21 +18,34 @@ import * as React from 'react';
...
+  'revision.subject'?: string;
+  'revision.timestamp'?: number | Date;
+  'revision.link'?: string;
+  'ci.link'?: string;
+}
+
+export type GitCommitInfo =  (Info & { generatedAt?: number }) | undefined;
+
+export const MetadataView: React.FC<GitCommitInfo> = metadata => {
+  if (!metadata['ci.link'] && !metadata['revision.id'] && !metadata['revision.author'] && !metadata['revision.email'] && !metadata['revision.subject'] && !metadata['revision.timestamp'] && !metadata['revision.link'])
```
pavelfeldman, line 38:
```suggestion
  if (!Object.keys(metadata).find(k => k.startsWith('revision.') || k.startsWith('ci.')))
```

packages/html-reporter/src/metadataView.tsx
```diff
@@ -73,10 +86,10 @@ export const MetadataView: React.FC<Metadata> = metadata => {
           icon='externalLink'
         />
       }
-      {metadata['generatedAt'] &&
+      {metadata.generatedAt &&
```
pavelfeldman, line 89: Let's keep them uniform, always access as string keys, all lower-case, '.' for namespaces. Core can be no namespace, 'timestamp' in this case.

packages/html-reporter/src/reportView.tsx
```diff
@@ -23,10 +23,12 @@ import { HeaderView } from './headerView';
 import { Route } from './links';
 import type { LoadedReport } from './loadedReport';
 import './reportView.css';
+import type { GitCommitInfo } from './metadataView';
```
pavelfeldman, line 26: So this is inside HTML report, so it can't know user has Git.

packages/html-reporter/src/reportView.tsx
```diff
@@ -46,7 +48,7 @@ export const ReportView: React.FC<{
   return <div className='htmlreport vbox px-4 pb-4'>
     <main>
       {report?.json() && <HeaderView stats={report.json().stats} filterText={filterText} setFilterText={setFilterText}></HeaderView>}
-      {report?.json().metadata && <MetadataView {...report?.json().metadata!} />}
+      {report?.json().metadata && <HideIfErrors><MetadataView {...report?.json().metadata['git-commit-info'] as GitCommitInfo} /></HideIfErrors>}
```
pavelfeldman, line 51: I thought metadata would be flat?

packages/playwright-test/src/plugins/gitCommitInfoPlugin.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+import type { Config, Metadata, TestPlugin } from '../types';
+import { createGuid } from 'playwright-core/lib/utils';
+import { spawnAsync } from 'playwright-core/lib/utils/spawnAsync';
+
+const GIT_OPERATIONS_TIMEOUT_MS = 1500;
+
+export const gitCommitInfo = (options?: GitCommitInfoPluginOptions): TestPlugin => {
+  return {
+    name: 'playwright-git-commit-info-plugin',
```
pavelfeldman, line 24:
```suggestion
    name: 'playwright:git-commit-info',
```

packages/playwright-test/src/plugins/gitCommitInfoPlugin.ts
```diff
@@ -0,0 +1,97 @@
...
+        ...options?.info ? options.info : await gitStatusFromCLI(options?.directory || configDir),
+        generatedAt: Date.now(),
+      };
+        // Normalize dates
+      const timestamp = info['revision.timestamp'];
+      if (timestamp instanceof Date)
+        info['revision.timestamp'] = timestamp.getTime();
+
+      config.metadata = config.metadata || {};
+      config.metadata['git-commit-info'] = info as Metadata;
```
pavelfeldman, line 38: Yeah, so I was assuming we are using global namespace for revisions. I.e. metadata['revision.timestamp'] so that non-git could use the same namespace.

packages/playwright-test/types/test.d.ts
```diff
@@ -921,6 +921,8 @@ export interface Config<TestArgs = {}, WorkerArgs = {}> extends TestConfig {
   use?: UseOptions<TestArgs, WorkerArgs>;
 }
 
+export type Metadata = { [key: string]: string | number | boolean | Metadata };
```
pavelfeldman, line 924:
```suggestion
export type Metadata = { [key: string]: string | number | boolean };
```

packages/html-reporter/src/reportView.tsx
rwoll: Correct—one of the tests covers the case they either (1) don't have info/git, (2) manually set metadata to something that conflicts with our schema.

packages/html-reporter/src/metadataView.tsx
```diff
@@ -18,21 +18,62 @@ import * as React from 'react';
...
+  'revision.id'?: string;
+  'revision.author'?: string;
+  'revision.email'?: string;
+  'revision.subject'?: string;
+  'revision.timestamp'?: number | Date;
+  'revision.link'?: string;
+  'ci.link'?: string;
+}
+
+export type GitCommitInfo =  (Info & { timestamp?: number }) | undefined;
```
pavelfeldman, line 35:
```suggestion
export type CommitInfo =  (Info & { timestamp?: number }) | undefined;
```

pavelfeldman: You should merge it into the above Info, rename to Metainfo (Info is too generic)?

packages/html-reporter/src/metadataView.tsx
```diff
@@ -18,21 +18,62 @@ import * as React from 'react';
...
+
+export class ErrorBoundary extends React.Component<{}, { error: Error | null, errorInfo: React.ErrorInfo | null }> {
```
pavelfeldman, line 37: I would not want to encourage this kind of try/catch behavior, what's wrong with a local try/catch where you are validating user-provided data?

pavelfeldman (approved):

packages/html-reporter/src/metadataView.tsx
```diff
@@ -18,21 +18,61 @@ import * as React from 'react';
...
+  'revision.author'?: string;
+  'revision.email'?: string;
+  'revision.subject'?: string;
+  'revision.timestamp'?: number | Date;
+  'revision.link'?: string;
+  'ci.link'?: string;
+  'timestamp'?: number
+} | undefined;
+
+export class ErrorBoundary extends React.Component<{}, { error: Error | null, errorInfo: React.ErrorInfo | null }> {
```
rwoll, line 36:
```suggestion
class ErrorBoundary extends React.Component<{}, { error: Error | null, errorInfo: React.ErrorInfo | null }> {
```

Issue #13106, opened 2022-03-27
[BUG] Cannot remove header in `route.continue`


szmarczak:
**Context:**
- Playwright Version: 1.20.1
- Operating System: Linux
- Node.js version: 17.8.0
- Browser: Chromium

**Code Snippet**

```javascript
const playwright = require('playwright');

require('http').createServer((request, response) => {
        response.end('');
}).listen(8888);

(async () => {
        const browser = await playwright.chromium.launchPersistentContext('', {
                headless: false,
        });

        const page = await browser.newPage();

        await page.route('**/*', (route, request) => {
          // Override headers
          const headers = {
            ...request.headers(),
            foo: 'bar', // set "foo" header
            origin: undefined, // remove "origin" header
          };
          route.continue({headers});
        });

        await page.goto('http://localhost:8888/', { waitUntil: 'load' });
})();
```

**Describe the bug**

The example https://playwright.dev/docs/next/api/class-route#route-continue is broken.

```
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

route.continue: headers[7].value: expected string, got undefined
    at RouteHandler.handler (/home/szm/Desktop/cache/bug6.js:21:18)

Node.js v17.8.0
```

rwoll:
Thanks for the report! I repro'd locally. 

Does using `delete` built-in before passing along the headers work for your use case?

```
const headers = {
  ...request.headers(),
  foo: 'bar'
};

delete headers['x-some-header'];
…
```

Regardless, we'll either update the docs or make the doc example work!

---

Note from repro: The error is likely from our protocol type def: https://github.com/microsoft/playwright/blob/1471bb71776cf84223190e1552ee4e055db1547a/packages/playwright-core/src/protocol/protocol.yml#L2442

szmarczak:
> Does using delete built-in before passing along the headers work for your use case?

It doesn't. I guess the headers are merged.

Horsty80:
I opened an issue #13527 , maybe it's the same issue than here with the `route.continue()`

yury-s:
> It doesn't. I guess the headers are merged.

I can only reproduce it in WebKit, the `origin` header is deleted fine in Chromium and Firefox, adding a test for that.

Can you share a code snippet that reproduces it in Chromium with `delete headers['origin']`?

szmarczak:
@yury-s Please note that I'm removing the `origin` header

yury-s:
[...]

PR #13880, opened 2022-05-02
fix: delete headers with undefined value

yury-s:
https://github.com/microsoft/playwright/issues/13106

Instead of throwing 'route.continue: headers[5].value: expected string, got undefined' on the server side, just treat such headers as if they were deleted from the map via `delete headers['name']`


rwoll (approved):
Provisional LGTM. I was expecting to see a change to protocol.yml change (and ensure clients can send undefined/None etc. so this fix will apply to all lang ports).

yury-s:
Ports have different logic how to delete the values from the map and I'd rather keep current protocol and let them handle the values correctly on their end (note that there is normally no null vs undefined distinction in other languages).

PR #13881, opened 2022-05-02
chore: standardize plugin naming

rwoll:


pavelfeldman:
Obsolete now, but we'll restore it at some point

PR #13883, opened 2022-05-02
chore: debug log plugin phases

rwoll:


packages/playwright-test/src/util.ts
```diff
@@ -294,3 +294,5 @@ export async function normalizeAndSaveAttachment(outputPath: string, name: strin
     return { name, contentType, body: typeof options.body === 'string' ? Buffer.from(options.body) : options.body };
   }
 }
+
+export const pluginLogger = (plugin: TestPlugin) => debug(`pw:plugin:${plugin.name}`);
```
mxschmitt, line 298: lets use pw-test instead of pw? Since its more test-runner specific.

rwoll, line 298: This is keeping it simple. I almost wrapped our plugins with a Proxy so we would log accesses to the interesting methods, but that became silly.

rwoll: Is there a precedent for this? I frequently just do `pw:*` expecting to get all the relevant logs

rwoll: There is a `export const debugTest = debug('pw:test');` in a few places, so I will scope it from there

rwoll:
Closing. We'll revive when we want to land user-facing plugin API.

rwoll:
Closing. We'll revive when we want to land user-facing plugin API.

PR #13889, opened 2022-05-03
docs(dotnet): add API testing

mxschmitt:


docs/src/api/params.md
```diff
@@ -310,26 +310,32 @@ Target URL.
 
 Query parameters to be sent with the URL.
 
-## java-fetch-params
+## csharp-fetch-option-params
+* langs: csharp
+- `params` <[Object]<[string], [any]>>
```
yury-s, line 315: Why `any` and not only primitive types like in other languages?

docs/src/api/params.md
```diff
@@ -355,16 +369,25 @@ this request body. If this parameter is specified `content-type` header will be
...
+Provides an object that will be serialized as html form using `multipart/form-data` encoding and sent as
+this request body. If this parameter is specified `content-type` header will be set to `multipart/form-data`
+unless explicitly provided. File values can be passed either as [`fs.ReadStream`](https://nodejs.org/api/fs.html#fs_class_fs_readstream)
+or as file-like object containing file name, mime-type and its content.
+
 ## js-python-fetch-option-data
-* langs: js, python
-- `data` <[string]|[Buffer]|[Serializable]>
+* langs: js, python, csharp
+- `data` <[string]|[Buffer]|[any]>
```
yury-s, line 383: Why not Serializable? Whatever we prefer here, let's keep it consistent with `RequestOptions.setData.data` (update both to use same type).

docs/src/api/class-requestoptions.md
```diff
@@ -22,7 +23,7 @@ Creates new instance of [RequestOptions].
 Sets the request's post data.
 
 ### param: RequestOptions.setData.data
-- `data` <[string]|[Buffer]|[Serializable]>
+- `data` <[string]|[Buffer]|[Object]>
```
yury-s, line 26: Why did it change?

utils/doclint/generateDotnetApi.js
```diff
@@ -407,7 +407,7 @@ function generateNameDefault(member, name, t, parent) {
           attemptedName = `${parent.name}BoundingBoxResult`;
         if (attemptedName === 'BrowserContextCookie')
           attemptedName = 'BrowserContextCookiesResult';
-        if (attemptedName === 'File')
+        if (attemptedName === 'File' || attemptedName === 'SetValue')
```
yury-s, line 410: SetValue may become non-unique in the future, consider checking parent name too.

yury-s (approved):

docs/src/api/params.md
mxschmitt: Because .NET does not allow unions, having it IDictionary<string, object> seems to be the best?

docs/src/api/params.md
mxschmitt: Serializable ends up in .NET to a generic, I can probably revert and massage the generator to prevent that in some cases.

utils/doclint/generateDotnetApi.js
mxschmitt: SG

PR #13894, opened 2022-05-03
chore: get rid of ProjectImpl

dgozman:


packages/playwright-test/src/loader.ts
```diff
@@ -324,6 +333,105 @@ ${'='.repeat(80)}\n`);
...
+      const fixtures = this._applyConfigUseOptions(testType, this._config.use);
+      const pool = new FixturePool(fixtures);
+      this._testTypePools.set(testType, pool);
+    }
+    return this._testTypePools.get(testType)!;
+  }
+
+  // TODO: we can optimize this function by building the pool inline in cloneSuite
+  private _buildPool(test: TestCase): FixturePool {
+    if (!this._testPools.has(test)) {
```
mxschmitt, line 358:
```suggestion
    if (this._testPools.has(test))
      return this._testPools.get(test)!;
    ...
```
nit: reduces nesting

mxschmitt (approved):

PR #13912, opened 2022-05-03
devops: more compact commits for the flakiness dashboard

aslushnikov:
- do not collect commit info since we don't use
- do not store CI plugin information
- do not store test timings

### DIFF
--- a/utils/flakiness-dashboard/processing/dashboard_compressed_v1.js
+++ b/utils/flakiness-dashboard/processing/dashboard_compressed_v1.js
@@ -43,14 +43,12 @@ function compressReports(reports) {
           delete project.metadata.mode;
         if (project.metadata.platform && project.metadata.platform.toLowerCase() !== 'android')
           delete project.metadata.platform;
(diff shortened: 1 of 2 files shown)


utils/flakiness-dashboard/processing/dashboard_compressed_v1.js
```diff
@@ -43,6 +43,16 @@ function compressReports(reports) {
...
           delete project.metadata.platform;
+        // Cleanup a bunch of data from report.
+        delete project.metadata['ci.link'];
+        delete project.metadata['revision.id'];
+        delete project.metadata['revision.author'];
+        delete project.metadata['revision.email'];
+        delete project.metadata['revision.subject'];
+        delete project.metadata['revision.timestamp'];
+        delete project.metadata['revision.link'];
+        delete project.metadata['timestamp'];
```
rwoll, line 54: We're likely to add more metadata in the future, so instead of deleting, let's do a safelisted approach.

PR #13922, opened 2022-05-04
fix windows electron launch enoent error

JaosnHsieh:
Without `shell:true` option on node.js `spawn` option, windows would show `electron. launch: Failed to launch: Error: spawn ..... ENOENT` error

reference:

https://stackoverflow.com/questions/37459717/error-spawn-enoent-on-windows

### DIFF
--- a/packages/playwright-core/src/protocol/channels.ts
+++ b/packages/playwright-core/src/protocol/channels.ts
@@ -3589,6 +3589,7 @@ export type ElectronLaunchParams = {
   },
   strictSelectors?: boolean,
   timezoneId?: string,
+  shell?: boolean;
 };
 export type ElectronLaunchOptions = {
   executablePath?: string,
--- a/packages/playwright-core/src/server/electron/electron.ts
+++ b/packages/playwright-core/src/server/electron/electron.ts
@@ -178,7 +178,7 @@ export class Electron extends SdkObject {
         handleSIGTERM: true,
         handleSIGHUP: true,
         onExit: () => {},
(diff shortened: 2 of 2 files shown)


pavelfeldman:
So I just checked it and our Electron tests pass on Windows.

pavelfeldman:
https://github.com/microsoft/playwright/runs/6295993378?check_suite_focus=true - windows looks clear here. Is there a bug this fixes?

JaosnHsieh:
Is the path to your Electron absolute? 

No, I use relative path like `path.join(__dirname,'..', 'node_modules', '.bin', 'electron')`.

Put `shell:true` fixed ENOENT error on windows10 worked for me, I suggest to make it an optional param so it won't change any behaviour by default.

Windows 10 version

```
>> [System.Environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      19044  0

```


Powershell version

```
>> $PSversionTable

Name                           Value
----                           -----
PSVersion                      5.1.19041.1645
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.19041.1645
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

liuxingbaoyu:
executablePath:path.join(__dirname,'..', 'node_modules', '.bin', 'electron.cmd')

`electron` is a linux script, windows should use `electron.cmd`.

Issue #11755, opened 2022-01-31
[BUG] svelte + 1.18.0-alpha: SyntaxError: The requested module '@playwright/test' does not provide an export named 'PlaywrightTestConfig'


bbigras:
**Context:**
- Playwright Version: @playwright/test@1.18.0-alpha-jan-10-2022
- Operating System: Linux
- Node.js version: v16.13.2
- Browser: <didn't reach this part>
- Extra: 

## System:
 - OS: Linux 5.15 NixOS 22.05 (Quokka) 22.05 (Quokka)
 - Memory: 740.97 MB / 15.51 GB
 - Container: Yes
## Binaries:
 - Node: 16.13.2 - /nix/store/1s5l90kdzz5ii4bafi2dnyzv3smziq49-nodejs-16.13.2/bin/node
 - npm: 8.1.2 - /nix/store/1s5l90kdzz5ii4bafi2dnyzv3smziq49-nodejs-16.13.2/bin/npm
## Languages:
 - Bash: 5.1.12 - /run/current-system/sw/bin/bash

**Code Snippet**

```sh
# Skeleton project, Typescript, Eslint, Prettier
npm init svelte@next svelte-playwright
cd svelte-playwright
npm install

# Typescript, "tests", true, false
npm init playwright
npx playwright test
```

I get: `Error: playwright.config.ts: Cannot import a typescript file from an esmodule.`

try 1.18.0-alpha:
```
npm i --save-dev @playwright/test@1.18.0-alpha-jan-10-2022
PW_EXPERIMENTAL_TS_ESM=1 npx playwright test
```

I get:
```
(node:62468) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
Using config at /home/bbigras/dev/svelte-playwright/playwright.config.ts
file:///home/bbigras/dev/svelte-playwright/playwright.config.ts:1
import { PlaywrightTestConfig, devices } from '@playwright/test';
         ^^^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module '@playwright/test' does not provide an export named 'PlaywrightTestConfig'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
    at ModuleJob.run (node:internal/modules/esm/module_job:181:5)
    at async Promise.all (index 0)
    at ESMLoader.import (node:internal/modules/esm/loader:281:24)
[...]

PR #13928, opened 2022-05-04
docs: in playwright.config.ts, make importing PlaywrightTestConfig as a type

jsakamoto:
# Summary

In all documents, use the `import type` syntax on "playwright.config.ts" for importing `PlaywrightTestConfig` to avoid the error if ESModule mode is enabled.

# Description

I know the following "playwright.config.ts" works well.

```ts
// playwright.config.ts
```

However, for some reason, if the "packahe.json" was configured like this:

```js
// package.json
{
  ...
  "type": "module",
  ...
```

then the "playwright.config.ts" above will cause the following error when the `px playwright test` command is executed.

```
SyntaxError: The requested module '@playwright/test' does not provide an export named 'PlaywrightTestConfig'
```

Fortunately , we can fix that error by importing the `PlaywrightTestConfig` as a type, like this:

```ts
// playwright.config.ts
import type { PlaywrightTestConfig } from '@playwright/test';
```

Moreover, the `import type` syntax above will work fine even if the `type` of the "package.json" is not defined (the default value of the `type` is `commonjs`). 

In my understanding, the initial code that will be generated from `npm init playwright` was changed to use the `import type` syntax to import the `PlaywrightTestConfig` for the exact same reason.
(See also [the comment in #11755](https://github.com/microsoft/playwright/issues/11755#issuecomment-1027307529))

Therefore, I propose changing all documents to use the `import type` syntax wherever importing the `PlaywrightTestConfig` in the "playwright.config.ts" file.

# References 
#11755

### DIFF
--- a/docs/src/intro-js.md
+++ b/docs/src/intro-js.md
@@ -138,7 +138,8 @@ module.exports = config;
 
 ```js js-flavor=ts
 // playwright.config.ts
-import { PlaywrightTestConfig, devices } from '@playwright/test';
+import type { PlaywrightTestConfig } from '@playwright/test';
+import { devices } from '@playwright/test';
 
 const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
--- a/docs/src/browsers.md
+++ b/docs/src/browsers.md
@@ -48,7 +48,7 @@ module.exports = config;
 ```
 
 ```js js-flavor=ts
-import { PlaywrightTestConfig } from '@playwright/test';
+import type { PlaywrightTestConfig } from '@playwright/test';
 const config: PlaywrightTestConfig = {
   use: {
     channel: 'chrome',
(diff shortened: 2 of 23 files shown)


docs/src/intro-js.md
```diff
@@ -138,7 +138,8 @@ module.exports = config;
 
 ```js js-flavor=ts
 // playwright.config.ts
-import { PlaywrightTestConfig, devices } from '@playwright/test';
+import type { PlaywrightTestConfig } from '@playwright/test';
```
pavelfeldman, line 141: You can do `import { type PlaywrightTestConfig, devices }`

pavelfeldman: @jsakamoto How do you feel about using inline notation instead?

jsakamoto: @pavelfeldman 
Yes, I feel that the inline notation is a bit smarter than my pull request. 👍
However, I just followed the code that the `npm init playwright` command generated.

Anyway, I strongly hope this pull request is accepted **rapidly**, even if my pull request is not the best.
Because there might be some people who are wasting time and energy due to that syntax error at this moment.
I was one of those people.

After doing that, it would be better to keep improving the documents to be smarter by using inline notation.

mxschmitt:
Seems implemented in https://github.com/microsoft/playwright/pull/14451#event-6695014916.

jsakamoto:
🎉

PR #13942, opened 2022-05-04
chore: widen Metadata type

rwoll:
@dgozman pointed out it was a bit odd we didn't allow for array's as values.
We discsussed and decided that while we Metadta to at least be Key-Value (object)
at the top-level, its inner shape shall be less constrained.

### DIFF
--- a/utils/generate_types/overrides-test.d.ts
+++ b/utils/generate_types/overrides-test.d.ts
@@ -40,7 +40,7 @@ export interface Project<TestArgs = {}, WorkerArgs = {}> extends TestProject {
 export interface FullProject<TestArgs = {}, WorkerArgs = {}> {
   grep: RegExp | RegExp[];
   grepInvert: RegExp | RegExp[] | null;
-  metadata: any;
+  metadata: Metadata;
   name: string;
   snapshotDir: string;
   outputDir: string;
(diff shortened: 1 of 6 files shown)


utils/generate_types/overrides-test.d.ts
```diff
@@ -66,7 +66,7 @@ export interface Config<TestArgs = {}, WorkerArgs = {}> extends TestConfig {
   use?: UseOptions<TestArgs, WorkerArgs>;
 }
 
-export type Metadata = { [key: string]: string | number | boolean };
+export type Metadata = { [key: string]: any };
```
mxschmitt, line 69: Out of curiosity, why not

```suggestion
export type Metadata = Record<string, any>;
```

rwoll: Stylistic pref+habit. We use a mix internally in overrides-test.d.ts.

(Some `Record`, some `{…}` variants. I find for simple types, not using parameterized helpers easier to read. When I was new to TS, if I saw `{ [key: string]: any }` I knew exactly what it meant.  When I saw `Record<string, any>`, I was like…"What's this Record type? I didn't know TS has Record. Is this different than an object?!")

pavelfeldman (approved):

PR #13943, opened 2022-05-04
chore(electron): add Win+Mac bots

pavelfeldman:


pavelfeldman:
Adding bots for #13922

.github/workflows/tests_electron.yml
```diff
@@ -0,0 +1,53 @@
...
+    - run: npm run build
+    - run: npx playwright install --with-deps chromium
+    - run: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- npm run etest
+      if: matrix.os == 'ubuntu-latest'
+    - run: npm run etest
+      if: matrix.os != 'ubuntu-latest'
+    - run: node tests/config/checkCoverage.js electron
+      if: always() && matrix.os == 'ubuntu-latest'
+    - run: ./utils/upload_flakiness_dashboard.sh ./test-results/report.json
+      if: always() && matrix.os == 'ubuntu-latest'
```
aslushnikov, line 48:
```suggestion
      shell: bash
      if: always()
```

aslushnikov (approved):

.github/workflows/tests_electron.yml
pavelfeldman: Let me start Linux-only and then follow up with multi-platform flakiness dashboard

PR #13950, opened 2022-05-05
chore: fixtures-via-plugin implementation

pavelfeldman:


yury-s (approved):
Looks reasonable, would be nice if Ross had a look at it too.

utils/generate_types/overrides-test.d.ts
```diff
@@ -56,9 +56,14 @@ export interface FullProject<TestArgs = {}, WorkerArgs = {}> {
 
 type LiteralUnion<T extends U, U = string> = T | (U & { zz_IGNORE_ME?: never });
 
+export interface TestPlugin {
+  fixtures?: Fixtures;
```
rwoll, line 60:
```suggestion
  fixtures?: Fixtures;
  // [internal] NB: other fields are generated from the docs
```

tests/playwright-test/plugins.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+      import plugin from './plugin.ts';
+      module.exports = {
+        plugins: [ plugin ],
+      };
+    `,
+    'plugin.ts': `
+      export default {
+        fixtures: {
+          foo: 42,
+          myBrowserName: async ({ browserName }, use) => { await use(browserName) }
```
rwoll, line 131: We don't need to do this now, but we should make sure the various fixture options work and/or still validate nicely.

packages/playwright-test/src/reporters/raw.ts
```diff
@@ -317,4 +317,12 @@ function dedupeSteps(steps: JsonTestStep[]): JsonTestStep[] {
   return result;
 }
 
+function filterOutPrivateFields(object: any): any {
+  if (!object || typeof object !== 'object')
+    return object;
+  if (Array.isArray(object))
+    return object.map(filterOutPrivateFields);
+  return Object.fromEntries(Object.entries(object).filter(entry => !entry[0].startsWith('_')).map(entry => [entry[0], filterOutPrivateFields(entry[1])]));
+}
```
rwoll, line 326: Let's make sure this change lands regardless of the fate of plugins.

rwoll (approved):

packages/playwright-test/src/reporters/raw.ts
pavelfeldman: I think we'll just _plugins

PR #13952, opened 2022-05-05
chore: allow global type overrides for Fixtures and Configs

rwoll:
See 6f04adc9c278db6bfaf3d489b7e6775aa19627e9 for example usage.

This allows plugins to influence the types for `PlaywrightTestConfig` and `test` (from `@playwright/test`).

This is intended to land with https://github.com/microsoft/playwright/pull/13950, and then be followed up with cleaning up our CT plugins to use both changes to get (1) nice typings, (2) avoid needing to repackage the runner.


utils/generate_types/overrides-test.d.ts
```diff
@@ -312,6 +312,24 @@ declare global {
...
+    }
+
+    export interface WorkerFixtures extends KeyValue {
+    }
+
+    export interface ConfigTestOptions extends KeyValue {
+    }
+
+    export interface ConfigWorkerOptions extends KeyValue {
+    }
```
rwoll, line 326: In an ideal world, I'd use `Global` instead of using `Config` v. `Project` prefix to imply the same thing, but we already have a convention.

utils/generate_types/overrides-test.d.ts
```diff
@@ -245,8 +245,8 @@ export interface PlaywrightTestArgs {
   request: APIRequestContext;
 }
 
-export type PlaywrightTestProject<TestArgs = {}, WorkerArgs = {}> = Project<PlaywrightTestOptions & TestArgs, PlaywrightWorkerOptions & WorkerArgs>;
-export type PlaywrightTestConfig<TestArgs = {}, WorkerArgs = {}> = Config<PlaywrightTestOptions & TestArgs, PlaywrightWorkerOptions & WorkerArgs>;
+export type PlaywrightTestProject<TestArgs = {}, WorkerArgs = {}> = Project<PlaywrightTestOptions & TestArgs & PlaywrightTest.TestFixtures & PlaywrightTest.ProjectTestOptions, PlaywrightWorkerOptions & WorkerArgs & PlaywrightTest.WorkerFixtures & PlaywrightTest.ProjectWorkerOptions>;
+export type PlaywrightTestConfig<TestArgs = {}, WorkerArgs = {}, ProjectTestArgs = {}, ProjectWorkerArgs = {}> = Config<PlaywrightTestOptions & TestArgs & PlaywrightTest.WorkerFixtures & PlaywrightTest.ConfigTestOptions, PlaywrightWorkerOptions & WorkerArgs & PlaywrightTest.WorkerFixtures & PlaywrightTest.ConfigWorkerOptions, PlaywrightTestOptions & ProjectTestArgs & PlaywrightTest.TestFixtures & PlaywrightTest.ProjectTestOptions, PlaywrightWorkerOptions & ProjectWorkerArgs & PlaywrightTest.WorkerFixtures & PlaywrightTest.ProjectWorkerOptions>;
```
rwoll, line 249: This is _really_ gross, so hopefully there is a more idiomatic way. However, prior to adding worker-scoped type parameters via https://github.com/microsoft/playwright/pull/13952/files#diff-69583b434e25c5c454472fca02f61a2f7caa0233032e79934e6e69525c622721R64, our projects array was incorrectly types.

Prior to df15f56622913819e6eacfb14d685686f196693a, the following would fail type check:

```ts
const p: PlaywrightTestProject = { use: { vitePort: 1 } }
```

while this would pass:

```ts
    projects: [
        { …, 
            use: { vitePort: 1 },
        } 
    ]
} 
```

rwoll:
Closing in favor of https://github.com/microsoft/playwright/pull/13977

PR #13971, opened 2022-05-05
tests: remove outdated redundant skips

mxschmitt:
- Stable Chrome is now 101
- drive-by removed some unused imported fixtures


yury-s (approved):

rwoll (changes requested):
Please leave the skips that have version guards. IMO, they serve as documentation and are helpful when debugging—especially when users come to us with issues where they are using a custom executable.

PR #13976, opened 2022-05-05
tests: Enable "SharedArrayBuffer should work @smoke"

Meir017:
SharedArrayBuffer is enabled in version https://webkit.org/blog/12140/new-webkit-features-in-safari-15-2/

also in MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#browser_compatibility

### DIFF
--- a/tests/library/capabilities.spec.ts
+++ b/tests/library/capabilities.spec.ts
@@ -19,7 +19,6 @@ import url from 'url';
 import { contextTest as it, expect } from '../config/browserTest';
 
 it('SharedArrayBuffer should work @smoke', async function({ contextFactory, httpsServer, browserName }) {
-  it.fail(browserName === 'webkit', 'no shared array buffer on webkit');
   const context = await contextFactory({ ignoreHTTPSErrors: true });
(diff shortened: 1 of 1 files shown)


Meir017:
test is still failing, maybe this was not fixed yet 🤔
image
https://github.com/microsoft/playwright/runs/6311922527?check_suite_focus=true

mxschmitt:
Hi, let's file an issue for it first so we can track it as a feature request? Then we can mark it as a specific version and dig into the cause why its not working on WebKit.

yury-s:
I debugged it a bit and believe it's because we disabled COOP handling in https://github.com/microsoft/playwright/pull/9185 as it broke some fundamental functionality at that point. We can try reebling it and see if the tests pass.

yury-s:
I any case SharedArrayBuffer only works when WebAssembly is supported and it is currently not on Windows so the test is not going to work on windows.

yury-s:
Filed https://github.com/microsoft/playwright/issues/14043

PR #13977, opened 2022-05-05
chore: use plugins for component testing again

pavelfeldman:


docs/src/test-components-js.md
```diff
@@ -62,7 +62,7 @@ npm i @playwright/experimental-ct-react
 ### Create a test `src/App.spec.tsx`
 
 ```js
-import { test, expect } from '@playwright/experimental-ct-react';
```
rwoll, line 65: I expected to see more changes in this doc mentioning including the plugin and the typings. They'll see it in the `examples/*`, but it's worth pointing out which few lines in there are the wiring.

packages/playwright-test/types/test.d.ts
```diff
@@ -475,7 +474,7 @@ interface TestConfig {
    *
    */
   webServer?: TestConfigWebServer;
-  plugins?: TestPlugin[],
+  plugins?: (TestPlugin | string | [string, any])[],
```
rwoll, line 477: All that nice typing…and then an `any` for the string load case?!

packages/playwright-test/types/test.d.ts
```diff
@@ -2620,7 +2619,7 @@ export type VideoMode = 'off' | 'on' | 'retain-on-failure' | 'on-first-retry';
  * ```
  *
  */
-export interface PlaywrightTestOptions {
+export interface PlaywrightTestOptions extends PlaywrightTest.TestOptions {
```
rwoll, line 2622: This is an odd direction of extension. If user overrides `acceptDownload: string` in their global, what will TS compiler say? I think it will yell about `PlaywrightTestOptions`, but we want it to say `PlaywrightTest.TestOptions` is wrong.

tests/components/ct-react/tsconfig.json
```diff
@@ -21,6 +21,7 @@
     "jsx": "react-jsx"
   },
   "include": [
-    "src"
+    "src",
```
rwoll, line 24: nit: we mix styling in tsconfig.json. Looks like other cases we put everything on one line, and others we do this.

utils/generate_types/index.js
```diff
@@ -93,7 +93,7 @@ class TypesGenerator {
 
     let overrides = await parseOverrides(overridesFile, className => {
       const docClass = this.docClassForName(className);
-      if (!docClass)
+      if (!docClass || !this.shouldGenerate(className))
```
rwoll, line 96: I think this is fine, but double check with @dgozman

utils/generate_types/overrides-test.d.ts
```diff
@@ -57,13 +57,14 @@ export interface FullProject<TestArgs = {}, WorkerArgs = {}> {
 type LiteralUnion<T extends U, U = string> = T | (U & { zz_IGNORE_ME?: never });
 
 export interface TestPlugin {
+  name: string;
```
rwoll, line 60:
```suggestion
  // [internal] other fields are added via docs
  name: string;
```

rwoll (approved):
Looks good to land…some nits, and I want to play around more with the typings post-land to see how TS compiler behaves in some edge cases with its error messaging.

docs/src/test-components-js.md
pavelfeldman: I'll follow up when the story settles

packages/playwright-test/types/test.d.ts
pavelfeldman: I'll follow up with simplification!

packages/playwright-test/types/test.d.ts
pavelfeldman: Error would still be actionable i think

utils/generate_types/index.js
pavelfeldman: It affected the output in the expected manner, so should be good. We commit the results of the run.

utils/generate_types/overrides-test.d.ts
pavelfeldman: will follow up

PR #14001, opened 2022-05-06
test(components): add test for doc example and React 18

rwoll:
Ensures we have a test for the example in our docs (and React 18)!

NB: The docs were missing a `poll`.


tests/components/ct-react/src/Button.spec.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+test.use({ viewport: { width: 500, height: 500 } });
+
+test('should work', async ({ mount }) => {
+  let clicked = false;
+  const component = await mount(<Button onClick={() => clicked = true }></Button>);
+  await expect(component).toContainText('click me');
+  expect(clicked).toBe(false);
+  await component.click();
+  expect.poll(() => clicked).toBe(true);
```
mxschmitt, line 12: Did we figure out why it was not working during the demo?

rwoll: Per-discussion:
1. we should make this work without poll
2. we still need to resolve the complaint in the console that we saw this AM about circular objects

PR #14003, opened 2022-05-06
wip: fix cycle errors in ct code

rwoll:
Without this change, the newly added ct-react test does not work, and
react-dom logs a complaint about cyclic JSON.

The code here, however:

1. creates regressions in the event tests (so it's not viable)
2. is likely an attempt to fix the issue in the incorrect place

### DIFF
--- a/packages/playwright-test/src/mount.ts
+++ b/packages/playwright-test/src/mount.ts
@@ -23,6 +23,54 @@ export async function mount(page: Page, jsxOrType: any, options: any, baseURL: s
   }, true);
 }
 
+function serialize(value: any, visited: Set<any> = new Set()): any {
+  if (visited.has(value))
+    return undefined; // we found a cycle
+  if (typeof value === 'symbol')
+    return value;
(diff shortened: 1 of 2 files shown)


rwoll:
@pavelfeldman / @mxschmitt : This is not intended to land, so don't review :) We'll need to fix, but this isn't quite it. Just opening this draft to save off the work/convo.

packages/playwright-test/src/mount.ts
```diff
@@ -23,6 +23,54 @@ export async function mount(page: Page, jsxOrType: any, options: any, baseURL: s
...
+  if (typeof value === 'number')
+    return value;
+  if (typeof value === 'string')
+    return value
+  if (value instanceof Error)
+    return value;
+  if (value instanceof Date)
+    return value;
+  if (value instanceof RegExp)
+    return value;
```
rwoll, line 54: My brain was tired…most of this function could actually be squashed down if we end up needing it.

PR #14006, opened 2022-05-07
feat: move `toHaveScreenshot` to use old snapshot paths by default

aslushnikov:
Note: all toHaveScreenshot tests still use `__screenshots__` directory
for their expectations. One more test was added to make sure that
by default, `toHaveScreenshot` uses old snapshots.

### DIFF
--- a/tests/playwright-test/to-have-screenshot.spec.ts
+++ b/tests/playwright-test/to-have-screenshot.spec.ts
@@ -97,6 +97,27 @@ test('should fail with proper error when unsupported argument is given', async (
   expect(stripAnsi(result.output)).toContain(`Expected options.clip.width not to be 0`);
 });
 
+test('should use match snapshot paths by default', async ({ runInlineTest }, testInfo) => {
+  const result = await runInlineTest({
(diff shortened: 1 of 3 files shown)


tests/playwright-test/to-have-screenshot.spec.ts
```diff
@@ -922,6 +943,9 @@ test('should update expectations with retries', async ({ runInlineTest }, testIn
 });
 
 function playwrightConfig(obj: any) {
+  obj.expect ??= {};
+  obj.expect.toHaveScreenshot ??= {};
+  obj.expect.toHaveScreenshot._useScreenshotsDir ??= true;
```
dgozman, line 948: I guess this flag is temporary, just for tests?

dgozman (approved):

tests/playwright-test/to-have-screenshot.spec.ts
aslushnikov: Yes, unless we decide to release screenshotsDir somehow later in the future.

PR #14008, opened 2022-05-07
feat(bindings): best-effort serialize circular structures

pavelfeldman:


packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
```diff
@@ -23,129 +23,142 @@ export type SerializedValue =
...
 
-function isRegExp(obj: any): obj is RegExp {
-  return obj instanceof RegExp || Object.prototype.toString.call(obj) === '[object RegExp]';
-}
+export type HandleOrValue = { h: number } | { fallThrough: any };
 
-function isDate(obj: any): obj is Date {
-  return obj instanceof Date || Object.prototype.toString.call(obj) === '[object Date]';
-}
+export function source(ignoreCycles: boolean = false) {
```
yury-s, line 28: source -> `createSerializers()` ?

packages/playwright-core/src/server/page.ts
```diff
@@ -718,12 +726,12 @@ export class PageBinding {
   constructor(name: string, playwrightFunction: frames.FunctionWithSource, needsHandle: boolean) {
     this.name = name;
     this.playwrightFunction = playwrightFunction;
-    this.source = `(${addPageBinding.toString()})(${JSON.stringify(name)}, ${needsHandle})`;
+    this.source = `(${addPageBinding.toString()})(${JSON.stringify(name)}, ${needsHandle}, (${source})(true))`;
```
yury-s, line 729: Why can't we send `(${source})(true)` once into the page?

yury-s (approved):

packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
```diff
@@ -23,129 +23,142 @@ export type SerializedValue =
...
+  function serialize(value: any, handleSerializer: (value: any) => HandleOrValue, visited: Set<any>): SerializedValue {
+    const result = handleSerializer(value);
+    if ('fallThrough' in result)
+      value = result.fallThrough;
+    else
       return result;
+
+    if (visited.has(value)) {
+      if (ignoreCycles)
+        return '[Circular Ref]';
```
dgozman, line 92: - I am somewhat worried about performance, now that you can pass whatever to the binding.
- On the other hand, we can go further and restore cycles like we do with handles.

packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
dgozman: Yeah, things like `typeof source` look suspicious :)

dgozman (approved):

packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
pavelfeldman: Aliasing Window / Document / Node

packages/playwright-core/src/server/page.ts
pavelfeldman: Same applies to addPageBinding.

PR #14018, opened 2022-05-07
fix(debug): do not pause on internal calls, update ct docs

pavelfeldman:


packages/playwright-test/src/plugins/vitePlugin.ts
```diff
@@ -69,7 +69,7 @@ export function createPlugin(
           },
         },
         sourcemap: true,
-        outDir: viteConfig?.build?.outDir || path.join(viteConfig.root, './dist-pw/')
+        outDir: viteConfig?.build?.outDir || path.join(viteConfig.root, 'playwright', 'out')
```
mxschmitt, line 72: Should we add this to the create-playwright defaults?
(Maybe already done)

mxschmitt (approved):

packages/playwright-test/src/plugins/vitePlugin.ts
pavelfeldman: It is done, yes

PR #14019, opened 2022-05-08
chore: do not reset internal bindings for reuse

pavelfeldman:


mxschmitt (approved):

packages/playwright-test/src/mount.ts
```diff
@@ -58,24 +63,18 @@ async function innerMount(page: Page, jsxOrType: any, options: any, viewport: Vi
...
     component = jsxOrType;
 
-  const callbacks: Function[] = [];
-  wrapFunctions(component, page, callbacks);
-
-  const dispatchMethod = `__pw_dispatch_${createGuid()}`;
-  await page.exposeFunction(dispatchMethod, (ordinal: number, args: any[]) => {
-    callbacks[ordinal](...args);
-  });
+  wrapFunctions(component, page, boundCallbacksForMount);
```
dgozman, line 66: So, we are leaking all bound callbacks across the tests? Could be a problem if we end running thousands on tests in the same process.

dgozman (approved):

PR #14026, opened 2022-05-08
feat(ct): only rebuild when necessary

pavelfeldman:


packages/playwright-test/src/plugins/vitePlugin.ts
```diff
@@ -37,42 +40,68 @@ export function createPlugin(
...
-      viteConfig.root = viteConfig.root || configDir;
-      viteConfig.plugins = viteConfig.plugins || [
-        frameworkPluginFactory()
-      ];
-      const files = new Set<string>();
-      for (const project of suite.suites) {
-        for (const file of project.suites)
-          files.add(file.location!.file);
+      const rootDir = viteConfig.root || configDir;
+      const outDir = viteConfig?.build?.outDir || path.join(rootDir, 'playwright', '.cache');
```
mxschmitt, line 47: maybe `.playwright/cache` is better than `playwright/.cache`.

dgozman (approved):

packages/playwright-test/src/plugins/vitePlugin.ts
pavelfeldman: It is, but there is /playwright next to it anyways...

PR #14058, opened 2022-05-10
fix(test runner): perform duplicates check before filtering

dgozman:
Previously, this check happened after filtering by only and lines, but it should happen before that.
Also, fix the error message when multiple duplicate issues are present.

Fixes #14034.


packages/playwright-test/src/runner.ts
```diff
@@ -272,25 +273,25 @@ export class Runner {
...
+    // 2. Complain about duplicate titles.
+    const duplicateTitlesError = createDuplicateTitlesError(config, preprocessRoot);
+    if (duplicateTitlesError)
+      fatalErrors.push(duplicateTitlesError);
+
+    // 3. Filter tests to respect line/column filter.
     filterByFocusedLine(preprocessRoot, testFileReFilters);
 
-    // 3. Complain about only.
+    // 4. Complain about only.
```
pavelfeldman, line 284: Why do you check for `only` after the filtration, but check dupes before the filtration?

Issue #14047, opened 2022-05-10
[BUG] navigation to about:blank in headed hangs


pavelfeldman:
Repro steps:

1. ```npm init -y && npm i playwright@1.22.0-alpha-1652115973000```

2. ```vim test.mjs```

```js
import { chromium } from 'playwright';

(async () => {
  const browser = await chromium.launch({
    headless: false,
  });
  const page = await browser.newPage();
  await page.goto('about:blank');
  await browser.close();
})();
```

3. ```node test.mjs```

Bisect: https://chromium.googlesource.com/chromium/src/+log/c932654a16474815d92f582dd474386cc28770c3..2ec610f785e814658ed1ddef125bca1decc523c7

Offending: https://chromium.googlesource.com/chromium/src/+/4a5fef2495c176ca02e73fbc98a7787e5d5bde44

Review: https://chromium-review.googlesource.com/c/chromium/src/+/3341866

Kolic822:
I still have this problem with    "@playwright/test": "^1.29.2" in chromium

PR #14068, opened 2022-05-10
fix(chromium): work around about:blank issue on Chromium

dgozman:
We don't receive the `loaderId` which translates to `newDocumentId`,
so we expect the same-document navigation. Instead, we can wait
for any new-document navigation as a workaround, only for `about:blank`.

References #14047.

### DIFF
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -636,6 +636,14 @@ export class Frame extends SdkObject {
         }
         if (event.error)
           throw event.error;
+      } else if (isAboutBlank(url)) {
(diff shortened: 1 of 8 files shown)


packages/playwright-core/src/server/frames.ts
```diff
@@ -1748,3 +1756,7 @@ function verifyLifecycle(name: string, waitUntil: types.LifecycleEvent): types.L
   return waitUntil;
 }
 
+function isAboutBlank(url: string) {
+  return url === 'about:blank' || url.startsWith('about:blank#');
```
jameskip, line 1760:
```suggestion
  return url.includes('about:blank')
```

jameskip (approved):

pavelfeldman (approved):

Issue #13920, opened 2022-05-04
[BUG] - test case fails even before running the test case code/statements for Automatic fixtures


autTestUser:
**Context:**
- Playwright Version: 1.20.0
- Operating System: Windows
- Node.js version: 14.6.1
- Browser: Chromium/Chrome

**Code Snippet**

```javascript
export const test = base.extend<{ testFixture: void }>({

    testFixture: [async ({}, use) => {
        await use();
        throw new Error('MyError')
    }, { auto: true }],

});
```

**Describe the bug**

For the code snippet given, the automatic fixture doesn't work correctly. If I use *throw new Error()" it fails the test cases even before executing the contents/code in the "test" block. The test.beforeAll() or test.afterAll() would run, but the test itself fails without running the code inside it. Of course, the failure of the test case is expected, but it should not fail in the beginning, instead, it should fail after the test ends. Instead of an automatic fixture, if I use a normal fixture, then it works. Meaning, that the test case fails after executing the code inside the "test" block.

dgozman:
@autTestUser Could you please share the whole test? I think this could be related to exact hooks (`beforeAll` and others) that you have, so sharing a repro will help greatly!

autTestUser:
> @autTestUser Could you please share the whole test? I think this could be related to exact hooks (`beforeAll` and others) that you have, so sharing a repro will help greatly!

import { test } from "../../fixture";

let page: Page;
let myInstance;

test.describe("describeName", () => {

    test.beforeAll(async ({ browser }) => {
        page = await browser.newPage();
		myInstance= await new MyDefaultPage(page)
    });
	
    test.afterAll(async () => {
        await page.close()
    });

    test.beforeEach(async () => {
        myInstance.common()
    });

    test('menu displayed', async () => {     
		myInstance.isMenuDisplayed()

    });

});

autTestUser:
@dgozman When I ran the exact code you shared, I'm getting below output. Notice that "test works?" is not displayed and "will throw right here" is displayed twice. Also, the error/exception is thrown twice.

```
npx playwright test a.spec.ts

Running 1 test using 1 worker

will throw right here
will throw right here
  x  a.spec.ts:26:5 › describeName › menu displayed (0ms)


  1) a.spec.ts:26:5 › describeName › menu displayed ================================================

    Error: MyError

       6 |         await use();
       7 |         console.log('will throw right here');
    >  8 |         throw new Error('MyError');
         |               ^
[...]

PR #14104, opened 2022-05-11
fix(test runner): do not run automatic test fixtures for beforeAll hooks

### LINKED ISSUE #13186: [BUG] Unable to use test.BeforeAll when using text fixture with auto: true to run before every test
created a sample test with fixtures Refer the code below:

```ts
import { test as baseTest, expect} from '@playwright/test'

const test = baseTest.extend<{
    sampleFixture: void
}>({
    sampleFixture: [async({page}, use) => {
        console.log('before each')
        await use();
        console.log('after each')
    }, {auto: true}]
})

test.beforeAll(async({}) => {
    console.log('test before all')
})


test('basic test', async ({ page }) => {
    const title = page.locator('.navbar__inner .navbar__title');
    await expect(title).toHaveText('Playwright');
    console.log('test')
});
```

on executing the above code getting the following error:
Error: "context" and "page" fixtures are not supported in beforeAll. Use browser.newContext() instead.

Note: but there is no "page" and "context" used in the  beforeAll.
 

Playwright Version : 1.20.1

dgozman:
There are a few issues this covers:
- Some fixtures like `page` and `context` are not allowed in `beforeAll` hooks, so using them in automatic fixture makes it throw.
- Running automatic fixture solely for `afterAll` is unexpected. This currently happens when `afterAll` is run for cleanup after fixture timeout/throw.

For built-in playwright fixture, we keep `'all-hooks-included'` auto mode.
This does not affect automatic worker fixtures - these run first thing before any user callback in the worker.

Fixes #13186.
References #13920.

### DIFF
--- a/docs/src/test-fixtures-js.md
+++ b/docs/src/test-fixtures-js.md
@@ -609,3 +609,144 @@ const config: PlaywrightTestConfig<MyOptions> = {
 };
 export default config;
 ```
+
+## Execution order
+
+Each fixture has a setup and teardown phase separated by the `await use()` call in the fixture. Setup is executed before the fixture is used by the test/hook, and teardown is executed when the fixture will not be used by the test/hook anymore.
+
+Fixtures follow these rules to determine the execution order:
+* When fixture A depends on fixture B: B is always set up before A and teared down after A.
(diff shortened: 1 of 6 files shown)


pavelfeldman (approved):
Let's get ourselves a public doc.

PR #14113, opened 2022-05-11
test: add more layout selector tests

pavelfeldman:


docs/src/api/params.md
```diff
@@ -900,37 +900,54 @@ Note that outer and inner locators must belong to the same frame. Inner locator
...
+You are likely to use [`method: Locator.first`] on the outer locator when using this option. Otherwise you end up selecting all the elements to the left of the anchor and the strict checks trigger.
+:::
+
+Inner locator is queried against the same root as the outer one, it can't contain frame locators. See more details in the [layout selectors](../selectors.md#selecting-elements-based-on-layout) guide.
 
 ## locator-option-right-of
 - `rightOf` <[Locator]>
 
-Matches elements that are to the right of any element matching the inner locator, at any vertical position. Inner locator is queried against the same root as the outer one. More details in [layout selectors](../selectors.md#selecting-elements-based-on-layout) guide.
+Matches elements that are to the right of any element matching the inner locator (anchor), at any vertical position, sorted by the proximity to the anchor.
```
rwoll, line 914: I think these will all be made much clearer with concrete, inline examples. For example:

Given the following page:

Screen Shot 2022-05-11 at 4 46 38 PM


```
await expect(page.locator('css=img').filter({ rightOf: page.locator('text=My Profile') })).toBeVisible()
```

will produce an error since there are two (2) images to the right-of the _My Profile_ text.

Assuming you want to select the picture of the Bird (that has a solid green border in the above example), you'll need to disambiguate by using `first`:

```
await expect(page.locator('css=img').filter({ rightOf: page.locator('text=My Profile') }).first()).toBeVisible()
```

---

NB: Or should it be `last()`?! Regardless, the point still stands with using examples. Pictures are worth a 1000 words—especially in the case of layouts.

rwoll: On further thinking, `first()` and `last()` seem out of place and confusing when thinking about layout selectors. I think `lowest()`, `highest()`, `leftmost()` and `rightmost()` make more sense than `first()`/`last()` in contexts where you are using layout selector and thinking in pixels.

PR #14138, opened 2022-05-13
fix wrong command on -g option

maztak:
It worked with space not `=`.


rwoll (approved):
Thanks! Repro'd and confirmed the example was broken. CLA is signed, so this is good to merge. Cheers! 🎉

PR #14166, opened 2022-05-15
fix(recorder): remove waitForNavigation from codegen

pavelfeldman:


aslushnikov:
@pavelfeldman  We discussed this awhile ago with Dmitry, and here's the concern we had:

Consider automating a page that uses cashbusting URLs, like `https://example.com/libs?cachebust=ab23a23`

* Today, codegen script will not hard-code the URL.
* With this change, however, it'll put the cache busting hash in, so the script won't be replay-able.

pavelfeldman:
Replacing it with `https://example.com/libs?cachebust=*` would make it work though?

aslushnikov (approved):

PR #14230, opened 2022-05-17
fix: include @types/node in pwt deps

pavelfeldman:
Considering it to be the right thing to do since @playwright/test is itself being a dev dependency in the recommended setting.

### DIFF
--- a/packages/playwright-test/package.json
+++ b/packages/playwright-test/package.json
@@ -31,6 +31,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@types/node": "*",
     "playwright-core": "1.23.0-next"
   }
 }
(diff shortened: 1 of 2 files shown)


packages/playwright-test/package.json
```diff
@@ -31,6 +31,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@types/node": "^17.0.34",
```
mxschmitt, line 34:
```suggestion
    "@types/node": "*",
```

package.json
```diff
@@ -55,7 +55,7 @@
     "@babel/plugin-transform-typescript": "^7.16.8",
     "@babel/preset-react": "^7.16.7",
     "@types/formidable": "^2.0.4",
-    "@types/node": "=14.17.15",
+    "@types/node": "*",
```
mxschmitt, line 58:
```suggestion
    "@types/node": "14.17.15",
```

pavelfeldman: Hm, why 14?

pavelfeldman: Let me make it 12!

mxschmitt (approved):

PR #14243, opened 2022-05-18
fix(ct): recreate context on option change

pavelfeldman:


packages/playwright-test/src/mount.ts
```diff
@@ -100,3 +108,28 @@ function wrapFunctions(object: any, page: Page, callbacks: Function[]) {
...
+    javaScriptEnabled: context.javaScriptEnabled,
+    locale: context.locale,
+    offline: context.offline,
+    permissions: context.permissions,
+    proxy: context.proxy,
+    storageState: context.storageState,
+    timezoneId: context.timezoneId,
+    userAgent: context.userAgent,
+    deviceScaleFactor: context.deviceScaleFactor,
+  };
```
rwoll, line 133: I worry we will add a context option and never remember to add here making for a gnarly debug. (Especially given most of these are upstream browser options, they will rarely change and we will forget to edit here.)

Can we change this to be defensive so by default new context options will create a cache miss.

Something like:

```
const EXCLUDE_FROM_HASH = […];
const entriesToHash = Object.entries(context).filter(([key]) => !EXCLUDE_FROM_HASH.includes(key));
```

rwoll (approved):
LGTM, modulo some inline comments w.r.t. readability.

However, we should consider making optimization/reuse of the page opt-in (like we do for fullyParallel). For smaller suites, newPage is likely plenty fast.



I ran:

```
  const start = hrtime.bigint();
  for (let i = 0; i < 150; i++) {
    await page.close();
  }

  const end = hrtime.bigint();
  console.log('END', browserName, `${end - start} nanoseconds`);
```

for each of the browsers, and the results were:

| browser | time per iteration (ms) |
| --- | --- |
| chromium | ~40 |
| webkit | ~100 |
| firefox | ~150 |

PR #14267, opened 2022-05-19
fix: page.locator.focus() and page.locator(…).type(…) 

### LINKED ISSUE #14254: [BUG] focus() function in version 1.22 closes dropdown (not of select type) instead of just focus on the option
**Context:**
- Playwright Version: 1.22.1
- Operating System: Windows
- Node.js version: 16.15
- Browser: Chromium
- Extra: Works as well with version 1.21.1

**Code Snippet**

```javascript

await page.locator('locator').focus();

```

**Describe the bug**
I try to handle the dropdown (not of the `<select>` type with two playwright functions:
focus and click.

In a similar way, we handle dropdowns in our entire project.

The version of the 1.22.1 - fails due to the fact that the focus() function closes the dropdown. See the video:

https://user-images.githubusercontent.com/52656401/169029036-635a9487-b519-4517-ab64-8e2372df1e4f.mp4

You can find the branch with the reproduced test in the following link:
[branch with version 1.22.1](https://github.com/AlexKomanov/playwright-version-1-22-issue/tree/version-1.22)

The same test works as well with version 1.21.1:

https://user-images.githubusercontent.com/52656401/169026710-3466eb1e-a3c7-43b6-b616-5f35af8cab11.mp4

[master branch with version 1.21.1](https://github.com/AlexKomanov/playwright-version-1-22-issue/tree/master)

Also, you can see the same results from the GitHub Actions run results in the repo.

image

In my case it was reproduced in the following way:
First - run the branch with the 1.21.1 version - should work.
Second - try the branch with 1.22.1 version - should fail.

In reverse order, both of the tests will fail. 

But use the results of the workflow.

Maybe the focus function is not so necessary before click, but as you can see - something in the function behavior was changed between these two versions.

rwoll:
Fixes focus and blur management when `page.locator(…).focus()`  and  `page.locator(…).type(…)` are used which was regressed by 7a5b070 (#13510).

#13510 relied on an implicit assumption that this (conditional) [`blur`](https://github.com/microsoft/playwright/blob/7a5b070e9507b622877a2af010373585f2184196/packages/playwright-core/src/server/injected/injectedScript.ts#L672) call would always be followed by a call that resulted in a newly focused element via this [`focus`](https://github.com/microsoft/playwright/blob/7a5b070e9507b622877a2af010373585f2184196/packages/playwright-core/src/server/injected/injectedScript.ts#L674) call.

However, some elements are [not focusable](https://html.spec.whatwg.org/multipage/interaction.html#focusable-area), so we were blurring incorrectly, and losing focus that we should have maintained.

Two regression tests were added that pass on the commit prior to 7a5b070e9507b622877a2af010373585f2184196 (and match manual testing/expectations):

* `page.locator(…).focus()`: _keeps focus on element when attempting to focus a non-focusable element_
* `page.locator(…).type(…)`: _should type repeatedly in input in shadow dom_

Additionally, a third test (_should type repeatedly in input in shadow dom_) was added to check the invariant from #13510 that states:

> This affects [contenteditable] elements, but not input elements.

and allows us to introduce the targeted fix (contenteditble check before blur) without breaking FF again.

And _should type repeatedly in contenteditable in shadow dom with nested elements_ was added to ensure the above fix works with nest contenteditble detection.

Fixes #14254.

### DIFF
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -674,7 +674,7 @@ export class InjectedScript {
 
     const activeElement = (node.getRootNode() as (Document | ShadowRoot)).activeElement;
     const wasFocused = activeElement === node && node.ownerDocument && node.ownerDocument.hasFocus();
-    if (!wasFocused && activeElement && (activeElement as HTMLElement | SVGElement).blur) {
+    if ((node as HTMLElement).isContentEditable && !wasFocused && activeElement && (activeElement as HTMLElement | SVGElement).blur) {
       // Workaround the Firefox bug where focusing the element does not switch current
       // contenteditable to the new element. However, blurring the previous one helps.
       (activeElement as HTMLElement | SVGElement).blur();
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -671,7 +671,14 @@ export class InjectedScript {
       return 'error:notconnected';
     if (node.nodeType !== Node.ELEMENT_NODE)
       throw this.createStacklessError('Node is not an element');
-    const wasFocused = (node.getRootNode() as (Document | ShadowRoot)).activeElement === node && node.ownerDocument && node.ownerDocument.hasFocus();
+
+    const activeElement = (node.getRootNode() as (Document | ShadowRoot)).activeElement;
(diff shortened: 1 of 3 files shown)


rwoll:
Converted to a draft while I have some realizations offline and potentially keep a variation of the FF workaround.

rwoll:
Found another bug that's in 1.22 while working on this! That's fixed as well! Description of PR updated, so please read that for review.

pavelfeldman (approved):

PR #14316, opened 2022-05-20
chore: render titles on all matching nodes

pavelfeldman:


packages/playwright-core/src/server/injected/highlight.ts
```diff
@@ -113,90 +114,119 @@ export class Highlight {
...
-    if (boxes.length) {
-      const primaryBox = boxes[0];
-      let anchorLeft = primaryBox.left;
+    if (this._highlightIsUpToDate(elements))
+      return;
+
+    // 1. Destroy the layout
+    this.clearHighlight();
+
+    for (let i = 0; i < elements.length; ++i) {
```
yury-s, line 148: Add a separate comment for this step too ?

yury-s (approved):

packages/playwright-core/src/server/injected/highlight.ts
pavelfeldman: This is the same step!

PR #14317, opened 2022-05-21
fix(route): support route w/ async handler & times

### LINKED ISSUE #14313: [BUG] route.fulfill() fails when times:1 is used when it is not the first awaited call
**Context:**
- Playwright Version: 1.18.1
- Node.js version: 16.13.0
- Browser: Chromium

**Code Snippet**

The bug can be reproduced with the following snippet. Remove `, {times: 1}` to see the expected behavior.

```javascript
(async () => {
  const browser = await playwright.chromium.launch();
  await page.route("**/cars?**", async (route, request) => {
    await request.allHeaders();
    await route.fulfill({
      headers: { "access-control-allow-origin": "*" },
      status: 200,
      body: JSON.stringify(["BMW", "Tesla"]),
    });
  }, {times: 1});
  const resp = await page.evaluate(
    async () =>
      await (
        await fetch("https://example.com/cars?type=foobar", {
          method: "GET"
        })
      ).json()
  );
  console.log(resp);
  await browser.close();
})();
```

**Describe the bug**

If a `{times: 1}` route handler `await`s a function before calling `route.fulfill()`, the request is not mocked. Instead, the browser continues the request and the `fulfill` call seems to be ignored. Tracing with `DEBUG="pw:protocol"` points to the cause being a `Fetch.disable` request being sent before the `Fetch.fulfillRequest` message.

This makes intercepting exactly once anything where an async call is needed to determine the value to pass to fulfill impossible. Even the example code for intercepting a request and modifying the result fails because `page.request.fetch()` returns a promise.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14313

### DIFF
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -145,19 +145,26 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
 
   _onRoute(route: network.Route, request: network.Request) {
     for (const routeHandler of this._routes) {
-      if (routeHandler.matches(request.url())) {
+      if (!routeHandler.matches(request.url()))
+        continue;
+      // Immediately deactivate based on |times|.
+      if (routeHandler.willExpire())
+        this._routes.splice(this._routes.indexOf(routeHandler), 1);
+
+      (async () => {
         try {
-          routeHandler.handle(route, request);
+          // Let async callback work prior to disabling interception.
+          await routeHandler.handle(route, request);
         } finally {
-          if (!routeHandler.isActive()) {
-            this._routes.splice(this._routes.indexOf(routeHandler), 1);
-            if (!this._routes.length)
-              this._wrapApiCall(() => this._disableInterception(), true).catch(() => {});
-          }
+          if (!this._routes.length)
+            await this._wrapApiCall(() => this._disableInterception(), true).catch(() => {});
         }
-        return;
-      }
+      })();
+
+      // There is no chaining, first handler wins.
+      return;
     }
+
     // it can race with BrowserContext.close() which then throws since its closed
     route._internalContinue();
   }
--- a/packages/playwright-core/src/client/page.ts
+++ b/packages/playwright-core/src/client/page.ts
@@ -180,18 +180,24 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
 
   private _onRoute(route: Route, request: Request) {
     for (const routeHandler of this._routes) {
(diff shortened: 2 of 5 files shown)


packages/playwright-core/src/client/browserContext.ts
```diff
@@ -145,19 +145,26 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
 
   _onRoute(route: network.Route, request: network.Request) {
     for (const routeHandler of this._routes) {
-      if (routeHandler.matches(request.url())) {
+      if (!routeHandler.matches(request.url()))
+        continue;
+      // Immediately deactivate based on |times|.
+      if (routeHandler.willExpire())
```
yury-s, line 151: Are you sure it works fine with the for/of iteration?

tests/page/page-route.spec.ts
```diff
@@ -772,6 +772,19 @@ it('should support the times parameter with route matching', async ({ page, serv
   expect(intercepted).toHaveLength(1);
 });
 
+it('should support async handler w/ times', async ({ page, server }) => {
+  await page.route('**/empty.html', async route => {
```
yury-s, line 776: Please add a test for browser context too.

packages/playwright-core/src/client/page.ts
```diff
@@ -180,18 +180,24 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
...
+          // Let async callback work prior to disabling interception.
+          await routeHandler.handle(route, request);
         } finally {
-          if (!routeHandler.isActive()) {
-            this._routes.splice(this._routes.indexOf(routeHandler), 1);
-            if (!this._routes.length)
-              this._wrapApiCall(() => this._disableInterception(), true).catch(() => {});
-          }
+          if (!this._routes.length)
+            this._wrapApiCall(() => this._disableInterception(), true).catch(() => {});
```
yury-s, line 195: This is still not ideal as the handler may call request.continue()  asynchronously, maybe just leave it enabled or check if there are no inflight interceptions?

yury-s (approved):

packages/playwright-core/src/client/browserContext.ts
pavelfeldman: 1) yes 2) we return after a single iteration.

packages/playwright-core/src/client/page.ts
pavelfeldman: I think awaiting for the hook is a reasonable middle ground.

PR #12565, opened 2022-03-07
feat: add Service Worker network interception

rwoll:
Resolves #1090.

Some known TODOs:
- [ ] Code Organization
   - [x] move SW code into own CRServiceWorker 
   - [ ] clean up import cycles (and general haphazard importing)
   - [ ] Intro NetworkedTarget to try to mitigate all the `this._page || this._serviceWorker` bits
- [ ] More Tests!
   - [ ] CORS
   - [ ] Native Service Worker Fetch
   - [ ] Service Worker Cache Intercaction
   - [ ] Multi-Page Scenarios
   - [ ] WebSockets
   - [ ] setOffline
   - [ ] multiple
- [x] Toggle Service Worker Interception based on current routeing
- [ ] Review/resolve all `FIXME(raw):`/`TODO(raw):`
- [ ] (maybe) Add option to opt-out of SW interception. The current patch may break users who are currently doing interception and don't expect the SW bits to show up (or they assume `request.frame()` is always valid…it was before!)
- [x] ~(maybe) Add `worker.on('request',...)` etc. events to mirror page events~

PR #14321, opened 2022-05-21
feat: add Service Worker interception

### LINKED ISSUE #1090: [Feature] support route aka request interception for service workers
Hi,

I'd like to be able to include markup directly instead of using `page.goto` to a remote address. I currently have a small example here of what I've tried out but it appears that `'serviceWorker' in navigator` is returning false and registering a service worker fails.

I've tried `page.setContent` as well and I had the same results.

Can anyone tell me what I'm missing? The example I am running:

```js

const msg = 'hi';

const serviceWorkerPath = '/sw.js';
const serviceWorkerScope = '/';
const serviceWorker = `console.log("${msg}");`;

const html = `<!doctype html>
<html>
  <head>
    <title>One Service Worker</title>
  </head>

  <body>
    <script type="text/javascript">
      console.log('serviceWorker' in navigator);

      try {
        navigator.serviceWorker.register('${serviceWorkerPath}', {
          scope: '${serviceWorkerScope}'
        });
      } catch (e) {
        console.error(e);
      }
    </script>
  </body>
</html>
`.trim();

(async () => {

  await page.route('**/*.html', async request =>
    request.fulfill({
      status: 200,
      contentType: 'text/html',
      body: html,
    }),
  );

  await page.route(serviceWorkerPath, request =>
    request.fulfill({
      status: 200,
      contentType: 'text/javascript',
      body: serviceWorker,
    }),
  );

  await page.goto('http://example.com/index.html', {
    waitUntil: 'load',
  });

  const sw = await page.evaluate(`window.navigator.serviceWorker`);
  
  console.log(sw);

  if (sw) {
    const swTarget = await browser.waitForTarget(target => target.type() === 'service_worker');
    const worker = await browser.serviceWorker(swTarget);
    // use worker target/handle
  }

  await browser.close();
})();
```

### LINKED ISSUE #14666: [internal] Response.fullfilledByServiceWorker for FF

rwoll:
Adds Service Worker Main Script Interception (CR-only) and Service Worker related utils:

- (all browsers) `Response.fulfilledByServiceWorker`
- (all browsers) `serviceWorkers: "allow" | "block"` context option (and `test.use` option)
- (CR-only) instruments Service Worker Networking (allows for interception/routing, event listening, offline mode, etc.)

Resolves #1090
Resolves #14666
Supercedes #12565

Landed upstream patches:
- https://chromium-review.googlesource.com/c/chromium/src/+/3510917
- https://chromium-review.googlesource.com/c/chromium/src/+/3526571
- https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/3566669
- https://chromium-review.googlesource.com/c/chromium/src/+/3544685
- https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/3610924

Pending upstream patches (to resolve the failing test cases in this PR):
- https://chromium-review.googlesource.com/c/chromium/src/+/3689949

Since this Playwright change opts all users in by default, but the upstream pending patch is required to fix some rare failure modes, this Playwright PR should be followed up with a minor change that guards the installation of the new instrumentation based on CR major revision.


packages/playwright-core/src/server/chromium/crServiceWorker.ts
```diff
@@ -0,0 +1,74 @@
...
+  constructor(browserContext: CRBrowserContext, session: CRSession, url: string) {
+    super(browserContext, url);
+    this._browserContext = browserContext;
+    this._networkManager = new CRNetworkManager(session, null, this, null);
+    session.once('Runtime.executionContextCreated', event => {
+      this._createExecutionContext(new CRExecutionContext(session, event.context));
+    });
+
+    // These might fail if the target is closed before we receive all execution contexts.
+    this._networkManager.initialize().catch(() => {});
```
rwoll, line 37: We'll need to think about how to opt people in/out. Doing this unconditionally would risk old code triggering an error by calling `request.frame()`.

packages/playwright-core/src/client/network.ts
```diff
@@ -172,13 +173,25 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
   }
 
   frame(): Frame {
+    if (!this._initializer.frame) {
+      assert(this.isServiceWorkerRequest());
+      throw new Error('Service Worker requests do not have an associated frame.');
```
rwoll, line 178: Frame had been nullable, but then we made it non-nullable at some point in the past. We can't go back, and SW's don't have a frame associated with them.

Meir017:
Intercepting only requests sent from the service worker is not possible through the first condition argument, I was able to do something like this:
```js
test('should intercept only serviceworker request, not page', async ({ context, page, }) => {
  await context.route('**/sw.js', route =>
    route.fulfill({
      contentType: 'text/javascript',
      status: 200,
      body: `
        self.addEventListener('fetch', event => {
          if (event.request.url.endsWith('/data.json')) {
            event.respondWith(fetch(event.request));
          }
        });
      `,
    })
  );
  await context.route('**/data.json', async route => {
    if (route.request().isServiceWorkerRequest()) {
      return route.fulfill({
        contentType: 'text/plain',
        status: 200,
        body: 'from sw',
      });
    } else {
      return route.continue();
    }
  })

  const [ sw ] = await Promise.all([
    context.waitForEvent('serviceworker'),
    page.goto(server.PREFIX + '/serviceworkers/empty/sw.html'),
  ]);
  const response = await page.evaluate(() => fetch('/data.json').then(r => r.text()));
  expect(response).toBe('from sw');
});
```

rwoll:
@Meir017 See this newly added test: https://github.com/microsoft/playwright/pull/14321/commits/92e28e0e7269597d58e4062f2329050b3a8b575f#diff-f48a7787820cb49396256c1234745152eb05ca5a3f43b6d128ce0ee45eb6dff6R253

It aims to cover the intent of your scenario and passes. Notably it: (1) waits for the SW to be active, and (2) asserts the SW actually intercepted the request (according to the SW). Let me know if this covers your case!

tl;dr: I think https://github.com/microsoft/playwright/pull/14321#issuecomment-1133755994 was invalid since it didn't wait for the SW to be active, so SW wasn't yet handling your request (which was still a page req at the time and was reported as such in the context.route).

tests/library/chromium/chromium.spec.ts
```diff
@@ -30,6 +30,233 @@ test('should create a worker from a service worker', async ({ page, server }) =>
...
+  await page.evaluate(() => window['activationPromise']);
+  {
+    const [inner] = await Promise.all([
+      context.waitForEvent('request', r => r.url().endsWith('/inner.txt')),
+      page.evaluate(() => fetch('/inner.txt')),
+    ]);
+    // FIXME: This doesn't seem to be the correct assetion.
+    // Ensure the reques is actually intercepted by SW.
+    expect(inner.isServiceWorkerRequest()).toBe(false);
+    expect(inner.serviceWorker()).toBe(null);
```
rwoll, line 80: There's a bug somewhere! This is wrong. It should be considered service worker. Interestingly, if you add `context.route('**', (route) => route.continue());` to the top of the test, it passes. (Having the no-op context route does enable more CDP events.)

rwoll: For SW with `fetch` listener, it looks like we are firing two requestWillBeSentEvents. One on Page and one on SW.

rwoll:
The following test has different behavior when the `context.route` line is commented vs. uncommented:

```ts
test.only('isServiceWorker() and serviceWorker() work', async ({ context, page, server }) => {
    // context.route('**', (route) => route.continue());
    const [worker] = await Promise.all([
      context.waitForEvent('serviceworker'),
      page.goto(server.PREFIX + '/serviceworkers/fetch/sw.html')
    ]);
    await page.evaluate(() => window['activationPromise']);
    const [swReq, frameReq] = await Promise.all([
      context.waitForEvent('request', r => r.url().endsWith('/inner.txt') && r.isServiceWorkerRequest()),
      context.waitForEvent('request', r => r.url().endsWith('/inner.txt') && !r.isServiceWorkerRequest()),
      page.evaluate(() => fetch('/inner.txt')),
    ]);
    const [ url ] = await worker.evaluate(() => self['intercepted']);
    expect(url).toMatch(/\/inner\.txt$/);
    expect(swReq.frame).toThrow();
    expect(swReq.isServiceWorkerRequest()).toBe(true);
    expect(swReq.serviceWorker()).toBe(worker);

    expect(frameReq.frame()).toBe(page.mainFrame());
    expect(frameReq.isServiceWorkerRequest()).toBe(false);
    expect(frameReq.serviceWorker()).toBe(null);
});
```

With the following log line:

```patch
diff --git a/packages/playwright-core/src/server/chromium/crNetworkManager.ts b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
index fe1716c65..759a6c58d 100644
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -348,6 +348,7 @@ export class CRNetworkManager {
     this._responseExtraInfoTracker.responseReceived(event);
 
     const request = this._requestIdToRequest.get(event.requestId);
+    console.log('eventLocation', this._serviceWorker ? 'sw' : 'page', 'fromServiceWorker:', event.response.fromServiceWorker, request?.request.url());
     // FileUpload sends a response without a matching request.
     if (!request)
       return;
```

We see the following with the test run as shown above:

```
[chromium] › library/chromium/chromium.spec.ts:42:6 › isServiceWorker() and serviceWorker() work
eventLocation page fromServiceWorker: false http://localhost:8907/serviceworkers/fetch/sw.html
eventLocation page fromServiceWorker: false http://localhost:8907/serviceworkers/fetch/style.css
eventLocation sw fromServiceWorker: false http://localhost:8907/serviceworkers/fetch/sw.js
eventLocation sw fromServiceWorker: false http://localhost:8907/request-from-within-worker.txt
eventLocation sw fromServiceWorker: false http://localhost:8907/inner.txt
eventLocation page fromServiceWorker: true http://localhost:8907/inner.txt
```

and the following when the context.route is added:

```
[chromium] › library/chromium/chromium.spec.ts:42:6 › isServiceWorker() and serviceWorker() work
eventLocation page fromServiceWorker: false http://localhost:8907/serviceworkers/fetch/sw.html
eventLocation page fromServiceWorker: false http://localhost:8907/serviceworkers/fetch/style.css
eventLocation sw fromServiceWorker: false http://localhost:8907/serviceworkers/fetch/sw.js
eventLocation sw fromServiceWorker: false http://localhost:8907/request-from-within-worker.txt
eventLocation sw fromServiceWorker: false http://localhost:8907/inner.txt
eventLocation page fromServiceWorker: true undefined
```

The issue likley starts earlier in `_onRequestWillBeSent`, but it highlights a larger fundamental issue to resolve around this feature: if a SW's `fetch` event handler handles a frame's request, how do we report the request(s). In the case above, the fetch requests seem "the same", but there's only one request that hits the network. Also, recall the Frame could call `fetch('/foo')`, and the SW could decide to:

1. serve from cache
2. run the fetch request as is (like in the example above)
3. rewrite it completely and hit the network as `/bar`, or multiple endpoints that it then merges into one Request, etc.

In our simple example, how many times do we expect to see `context.on('request', …)` fire for `/inner.txt`? I think the answer has to be 2 times since there is no guarantee frame and SW requests map 1:1 to eachother. Given this, we have to provide enough info on our Request object:

```
Request
   isServiceWorkerRequest(): the network was actually engaged within the SW.
   isFromServiceWorker(): this says nothing about if the network was hit or not. It just indicates the request came from the SW (i.e. via [`respondWith`](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/respondWith)).  
```

^ This is a bit of word soup for me to look at on Monday 😄

rwoll:
Currently working on some of the opt-(in|out) logic. Once that's in, I'll update the description and open for review as that will be an actual candidate for landing and API review.

packages/playwright-core/src/server/dispatchers/networkDispatchers.ts
```diff
@@ -39,13 +40,15 @@ export class RequestDispatcher extends Dispatcher<Request, channels.RequestChann
...
-      frame: FrameDispatcher.from(scope, request.frame()),
+      frame: FrameDispatcher.fromNullable(scope, request.frame()),
+      serviceWorker: WorkerDispatcher.fromNullable(scope, request.serviceWorker()),
       url: request.url(),
       resourceType: request.resourceType(),
       method: request.method(),
       postData: postData === null ? undefined : postData.toString('base64'),
       headers: request.headers(),
       isNavigationRequest: request.isNavigationRequest(),
+      isServiceWorkerRequest: request.isServiceWorkerRequest(),
```
dgozman, line 51: Why send this one over the protocol? It can be easily computed on the client side.

packages/playwright-core/src/server/network.ts
```diff
@@ -193,10 +199,18 @@ export class Request extends SdkObject {
...
-  frame(): frames.Frame {
+  frame(): frames.Frame | null {
     return this._frame;
   }
 
+  serviceWorker(): pages.Worker | null {
+    return this._serviceWorker;
+  }
+
+  isServiceWorkerRequest(): boolean {
```
dgozman, line 210: I'd remove this method, and instead have it on the client side.

packages/playwright-core/src/server/types.ts
```diff
@@ -373,3 +374,8 @@ export type AndroidDeviceOptions = {
   port?: number,
   omitDriverInstall?: boolean,
 };
+
+export type ServiceWorkerPolicy = {
```
dgozman, line 378: I'd suggest an enum with 3 values, something like default/disabled/inspected.

packages/playwright-core/types/types.d.ts
```diff
@@ -14616,6 +14649,14 @@ export interface Response {
    */
   headerValues(name: string): Promise<Array<string>>;
 
+  /**
+   * > NOTE: This field is Chromium only. It's safe to call when using other browsers, but it will always be `null`.
+   *
+   * If set, indicates whether this Response was fullfilled by a Service Worker's Fetch Handler (i.e. via
+   * [FetchEvent.respondWith](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/respondWith)).
+   */
+  isFromServiceWorker(): null|boolean;
```
dgozman, line 14658: Why `null` here? I'd return `false` for other browsers.

docs/src/api/class-request.md
```diff
@@ -85,6 +85,15 @@ Name of the header.
 
 Whether this request is driving frame's navigation.
 
+## method: Request.isServiceWorkerRequest
```
dgozman, line 88: Are we sure we want this one? It is equivalent to `!!request.serviceWorker()` after all.

docs/src/api/class-request.md
```diff
@@ -207,6 +216,15 @@ following: `document`, `stylesheet`, `image`, `media`, `font`, `script`, `texttr
...
 Returns the matching [Response] object, or `null` if the response was not received due to error.
 
+## method: Request.serviceWorker
+- returns: <[null]|[Worker]>
+
+:::note
+This field is Chromium only. It's safe to call when using other browsers, but it will always be `null`.
+:::
+
+The Service [`Worker`] that is performing the request.
```
dgozman, line 226:
```suggestion
The Service [Worker] that is performing the request.
```

For the proper link.

docs/src/api/class-response.md
```diff
@@ -51,6 +51,15 @@ Name of the header.
 
 Returns all values of the headers matching the name, for example `set-cookie`. The name is case insensitive.
 
+## method: Response.isFromServiceWorker
```
dgozman, line 54: Bikeshedding: `fulfilledByServiceWorker`?

docs/src/api/params.md
```diff
@@ -636,10 +636,15 @@ contexts override the proxy, global proxy will be never used and can be any stri
...
 
-It specified, enables strict selectors mode for this context. In the strict selectors mode all operations
+If specified, enables strict selectors mode for this context. In the strict selectors mode all operations
 on selectors that imply single target DOM element will throw when more than one element matches the selector.
 See [Locator] to learn more about the strict mode.
 
+## context-option-service-worker-policy
+- `serviceWorkerPolicy` <[ServiceWorkerPolicy]>
+
+If set to `disabled`, all Service Worker registrations will be blocked.
```
dgozman, line 646: This needs explanation for all other values as well.

packages/playwright-core/src/client/network.ts
```diff
@@ -225,6 +238,8 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
   }
 
   private _raceWithPageClose(promise: Promise<any>): Promise<void> {
+    if (this.request().isServiceWorkerRequest())
+      return Promise.resolve();
```
dgozman, line 242: I think we should race with sw close event here.

packages/playwright-core/src/server/chromium/crBrowser.ts
```diff
@@ -485,6 +470,8 @@ export class CRBrowserContext extends BrowserContext {
   async doUpdateRequestInterception(): Promise<void> {
     for (const page of this.pages())
       await (page._delegate as CRPage).updateRequestInterception();
+    for (const sw of this.serviceWorkers())
```
dgozman, line 473: I think we should also cover `setExtraHTTPHeaders`, `setOffline` and `doSetHTTPCredentials`.

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -241,7 +250,7 @@ export class CRNetworkManager {
       return;
     }
 
-    if (!frame) {
+    if (!frame && !this._serviceWorker) {
```
dgozman, line 253: Let's add a comment that non-service-worker requests must have a frame. If we didn't find one, pretend there was not request :)

packages/playwright-core/src/server/chromium/crServiceWorker.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    session.send('Runtime.enable', {}).catch(e => { });
+    session.send('Runtime.runIfWaitingForDebugger').catch(e => { });
+  }
+
+  updateRequestInterception(): Promise<void> {
+    return this._networkManager.setRequestInterception(this._needsRequestInterception()).catch(e => { });
+  }
+
+  _needsRequestInterception(): boolean {
```
dgozman, line 49: nit: I think this is called from CRNetworkManager, so remove underscore?

utils/generate_types/overrides.d.ts
```diff
@@ -336,5 +336,9 @@ export type AndroidKey =
 export const _electron: Electron;
 export const _android: Android;
 
+export type ServiceWorkerPolicy = 'default' | 'disabled' | {
```
dgozman, line 339: We prefer not to export auxilary types unless we really need that.

docs/src/api/class-request.md
rwoll: Agreed—removed.

packages/playwright-core/types/types.d.ts
rwoll: Unresolving—we're going to see if we can get the correct information here across all the browsers.

packages/playwright-core/src/server/types.ts
rwoll: Updating—this was an artifact of an evolving policy where there were more options.

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -122,6 +122,20 @@ export abstract class BrowserContext extends SdkObject {
...
     if (debugMode() === 'console')
       await this.extendInjectedScript(consoleApiSource.source);
+    if (this._options.serviceWorkerPolicy === 'disabled') {
+      await this.addInitScript(`
+
+        // Service Worker Policy Injected Script
+
+        navigator.serviceWorker.register = () => {
+          console.warn('Service Worker registration disabled by Playwright');
+          return Promise.reject('Service Worker registration disabled by Playwright');
```
pavelfeldman, line 132: I would probably stall this one.

packages/playwright-core/types/types.d.ts
rwoll: Per-offline comments, we'll land as it and follow up.

packages/playwright-core/src/server/browserContext.ts
rwoll: For stall here, do you mean return (A) a Promise that never resolves or (B) immediately return a resolved promise?

tests/library/browsercontext-service-worker-policy.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+import { browserTest as it, expect } from '../config/browserTest';
+
+it('should allow service workers by default', async ({ page, server }) => {
+  await page.goto(server.PREFIX + '/serviceworkers/empty/sw.html');
+  await expect(page.evaluate(() => window['registrationPromise'])).resolves.toBeTruthy();
+});
+
+it.describe('disabled', () => {
+  it.use({ serviceWorkerPolicy: 'disabled' });
```
rwoll, line 24: s/serviceWorkerPolicy/serviceWorkers

docs/src/api/class-response.md
```diff
@@ -22,6 +22,15 @@ Waits for this response to finish, returns always `null`.
 
 Returns the [Frame] that initiated this response.
 
+## method: Response.fulfilledByServiceWorker
+- returns: <[boolean]>
+
+:::note
+This field is Chromium only. It's safe to call when using other browsers, but it will always be `false` (for now).
+:::
```
rwoll, line 30:
```suggestion
```

docs/src/api/params.md
```diff
@@ -636,10 +636,19 @@ contexts override the proxy, global proxy will be never used and can be any stri
...
 
+## context-option-service-worker-policy
+- `serviceWorkerPolicy` <[ServiceWorkerPolicy]<"default"|"disabled"|"enable-network-events">>
+
+* `"default"`: Sites can register [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API). However, the Network requests from within the Service Worker (as well as the main Service Worker script request itself) are not reported in [`method: BrowserContext.route`] nor [`event: BrowserContext.request`] event listeners.
+* `"disabled"`: Blocks all registration of Service Workers.
+* `"enable-network-events"`: Service Workers are enabled, and their Network Requests will show up in [`method: BrowserContext.route`] and the [`event: BrowserContext.request`] event listeners as well as the HAR.
+
+Defaults to `"default"`.
+
```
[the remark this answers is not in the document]
rwoll: Per-offline discussion:

* change `serviceWorkersPolicy` to `serviceWorkers`
* options are `"allowed" | "blocked"`, defaults to `"allowed"`
* opt everyone in by default (based on some GH searches, we don't think the frame() throwing case will be hit)
* fix upstream bug with .response() hanging for errored SW Main Request scripts (e.g. 404, wrong content-type)

rwoll:
Closing this to chop it up into a few different PRs

PR #14325, opened 2022-05-21
fix: sanitize URLs with vbscript:

ElijahPepe:
The vbscript: protocols can be used to run scripts in much the same way as the javascript: protocol. This PR adds in validation for those aforementioned protocols in snapshotterInjected.ts and snapshotRenderer.ts.

### DIFF
--- a/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
+++ b/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
@@ -218,7 +218,7 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
     }
 
     private _sanitizeUrl(url: string): string {
-      if (url.startsWith('javascript:'))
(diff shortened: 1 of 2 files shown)


dgozman (commented):
Could you please share an example of `data:` url that runs a script?

ElijahPepe:
`data:text/html,<script>alert('alert!')</script>` is one.

dgozman:
> `data:text/html,<script>alert('alert!')</script>` is one.

Is that an iframe src, or something else? I think that there is no chance for the script to be executed in the trace, because we handle all the frames. At the same time, data urls are essential for correct traces, so blocking them is not an option.

If you have a trace that suffers from script execution due to a data url, I would really like to understand what's happening.

ElijahPepe:
> I think that there is no chance for the script to be executed in the trace, because we handle all the frames. At the same time, data urls are essential for correct traces, so blocking them is not an option.

Fair enough.

dgozman (approved):
Looks good, thank you for the PR!

PR #14326, opened 2022-05-22
feat(ct): support cra w/ .js

pavelfeldman:


rwoll (approved):
Contingent on #14369 landing (and removing fixmes's in some of those tests), LGTM.

As discussed, once this lands, let's late it bake and ask some folks to try it out in some real projects (both folks using Component Testing, and also just PW Test as this hits both).

Issue #6319, opened 2021-04-26
[BUG] Memory increases when same context is used


Romain-P:
**Context:**
- Playwright Version: Latest (today is 26/044/2021)
- Operating System: Linux
- Node.js version: tested on both node.js version
- Browser: chromium

**Describe the bug**

I'm watching full-js apps (e.g react/angular websites). I initialize one instance, 1 browser and 1 page.
I keep the page in cache & retrieving content every 2 seconds.

After 1/2hours, the memory goes crazy. I tried to `reload()` the page every 30 minutes. It doesn't free the memory.
Only way to free the memory is closing the page and recreating a new one.

What could be the source of this memory leak? I suppose `reload()` frees the javascript-vm so it must be a leak internally to the page

mxschmitt:
Currently the objects for e.g. request/response/route get only flushed when a new context is created. For testing you typically create a new context for each test. I folded a few issues into this one of users who also faced into that, so we can better keep track of it and might find a workaround for it in the future.

Romain-P:
I am recreating the page every 1 hour this is the only work around I have found to keep the memory stable.
Why would recreating the context actually helps? I mean doing `page.close()` and the re-allocating a page using `page.newPage()` is fixing the issue

mxschmitt:
`browser.newPage` does internally also create a new context for you and close it once you close the page, so its basically a helper wrapper to simplify its usage.

VikramTiwari:
Hey @mxschmitt  Are there some other similar quirks that can be used to lower memory/CPU usage? I am starting to look into performance and parallelization of our tests and things like this would be helpful. Thanks!

mxschmitt:
The Playwright test runner gets released soon, stay tuned! It handels all that for you.
Headless requires mostly less cpu/memory than headed.

LDubya:
> Currently the objects for e.g. request/response/route get only flushed when a new context is created

Is this intended functionality? While testing is one use case for a headless browser, data science related activities such as scraping are another, and that field is growing. In those instances, you may not want to create an entirely new context, as the cookies may be important. But if you're retrieving content every two seconds, you can hit 1,800 requests in an hour, and that can easily amass to over 1GB of additional memory usage within an hour if all of their requests and responses are being stored. I think being able to have the page flush requests and responses when new ones are sent would be a useful feature. While headless browsers have their origins in browser testing, scraping is a major selling point for using headless browsers and this memory leak leads programs to break.

[...]

PR #13991, opened 2022-05-06
fix: leaking server side objects

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2132

PR #14340, opened 2022-05-23
fix(leak): do not retain test fixtures in worker fixtures

### LINKED ISSUE #14336: pageshot case memory leak
context: 
- playwright version: 1.19.2
- operation System: Linux
- node.js version:  14 & 16
- browser: chromium

I am using pageshot to save images, but the memory increased obvious~
when check code line by line, I find this: It seems pageshot method case a memory leak?

1. **the heap keeps even I didn't use the Buffer**
image
image

2. **and when delete this code, the Retained Size clear**:
image
image

_Originally posted by @roy-k in https://github.com/microsoft/playwright/issues/6319#issuecomment-1133694599_

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/14336
Affects: https://github.com/microsoft/playwright/issues/6319


mxschmitt:
I don't think it fixes https://github.com/microsoft/playwright/issues/6319 because it is about playwright-library and the issue if you keep using the same context over and over again that we don't flush all the objects.

There was https://github.com/microsoft/playwright/pull/13991 which fixed leaking BrowserContext and Page instances, yours fixes a different issue I guess in the test-runner.

PR #14342, opened 2022-05-23
chore(dotnet): generate object types from protocol

mxschmitt:


utils/generate_dotnet_channels.js
```diff
@@ -45,8 +70,6 @@ function mapType(type) {
     return 'RequestTimingResult';
   if (type === 'LifecycleEvent')
     return 'WaitUntilState';
-  if (type === 'NameValue')
```
mxschmitt, line 48: we now use the generated type internally.

yury-s (approved):

PR #14353, opened 2022-05-23
feat(trace-viewer): show request time in trace viewer network tab

### LINKED ISSUE #11893: [Feature] Trace Viewer better if network page includes elapsed time for each api call
The Trace Viewer is an indispensable tool when trying to find out why delays and timeouts occur in Playwright tests.
This is important in order to know whether it is an API call that is slow or plain networking, so that the tests can be written in a more robust manner.
However, I currently have to run both Playwright test ( with trace turned on ) and Fiddler in order to track down the issues.
In the picture I captured below, there is a 3 second gap ( green box ) on the Trace Viewer that I needed to map to the timeline of Fiddler. I had to first identify key events ( red boxes, like styles.3fbef etc ) on both Trace Viewer and Fiddler to find the right area of timeline to look. This is quite tedious and exacting.
For example, is it possible in Trace Viewer to list the elapsed time as part of the network log so that each GET, POST time can also be seen, without having to look at Fiddler's timeline. This would make it easier to find calls that take a long time or cause the timeout. 
TraceViewToFiddler

nil1511:
Fixes: #11893. Added resource time in the expanded details view. 

Screenshot 2022-05-23 at 10 02 16 PM

### DIFF
--- a/packages/trace-viewer/src/ui/networkResourceDetails.css
+++ b/packages/trace-viewer/src/ui/networkResourceDetails.css
@@ -96,3 +96,7 @@
   margin: 3px 0;
   font-weight: bold;
 }
+
+.network-request-details-time {
+  float: right;
+}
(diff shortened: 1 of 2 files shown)


nil1511:
@pavelfeldman can you review this pr?

nil1511:
@pavelfeldman can you review this pr?

pavelfeldman (approved):

Issue #14348, opened 2022-05-23
[BUG] Types not compatible with TS4.7 Node ESM support


Jamesernator:
**Context:**
- Playwright Version: 1.22.2
- Operating System: Linux
- Node.js version: 18.2.0
- Browser: N/A
- Extra: 

**Code Snippet**

Occurs in an ESM node file:

```ts
// Could not find a declaration file for module 'playwright'. '/home/jamesernator/scripts/node_modules/playwright/index.mjs' implicitly has an 'any' type.
//  Try `npm i --save-dev @types/playwright` if it exists or add a new declaration (.d.ts) file containing `declare module 'playwright';`ts(7016)
import playwright from "playwright";
```

**Describe the bug**

Basically the types as declared don't work with the new [`Node16`/`NodeNext` ESM support in TypeScript 4.7](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-rc/#ecmascript-module-support-in-node-js).

This happens because `playwright` only specifies a `index.d.ts` file which in TS4.7 ESM support is **only associated to index.js**. As such when importing via ESM, the resolution to `index.mjs` has no corresponding `.d.mts` file.

One resolution to this is just to include `index.d.mts` with the following line:

```ts
export * from "./index.js";
```

yury-s:
Can you share a project where this happens?

Jamesernator:
> Can you share a project where this happens?

Minimum reproduction is here: https://github.com/Jamesernator/playwright-ts-4.7-issue

Run `npm install` and `npx tsc` to see the error (or see the error in text editor).

Note that while TypeScript 4.7 is an RC, it is expected to release in the next couple days.

pavelfeldman:
Fix is on its way, the workaround would be adding:

```
    "include": [
        "./node_modules/playwright/index.d.ts",
        ...
```
in your tsconfig.json

PR #14358, opened 2022-05-23
chore: bump TypeScript to 4.6

mxschmitt:
I tried TS 4.7-rc which was working, since we don't want RC's using the latest stable for now.

relates https://github.com/microsoft/playwright/issues/14348

We'll follow-up digging into the new TS ESM support since we have then a TS version which supports it.


yury-s (approved):
Why does it fix the issue from the bug though?

PR #14360, opened 2022-05-23
docs: use selector-based methods in examples

rwoll:
Deprecates `page.*` and `frame.*` methods that have a corresponding `Locator` method. Updates doc examples accordingly.

### DIFF
--- a/docs/src/api/class-frame.md
+++ b/docs/src/api/class-frame.md
@@ -178,6 +178,8 @@ Raw CSS content to be injected into frame.
 
 ## async method: Frame.check
 
+**DEPRECATED**: Use [`async method: Frame.locator`] combined with [`async method: Locator.check`] instead. [`Locator`]'s are strict by default, so they help avoid selecting an unexpected element due to an ambiguous selector.
+
 This method checks an element matching [`param: selector`] by performing the following steps:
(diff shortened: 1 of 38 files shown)


docs/src/api/class-frame.md
```diff
@@ -207,6 +209,8 @@ When all steps combined have not finished during the specified [`option: timeout
 
 ## async method: Frame.click
 
+**DEPRECATED**: Use [`async method: Frame.locator`] combined with [`async method: Locator.click`] instead. [`Locator`]'s are strict by default, so they help avoid selecting an unexpected element due to an ambiguous selector.
```
mxschmitt, line 212: I'm not sure if we should call it deprecated, more like not recommended?

rwoll: I think those are the same here. Plus, our doc generator turns these into @deprecated tags in the JSDoc so editors call it out more specifically. (It's an existing convention in our docs.)

Put another way: when would you want to use these over the Locator variants? I think never (since you can always do `first()` and it makes it explicit in the code that you are breaking from strictness). So the deprecated notice is warranted. Using these methods introduce subtle flakes and bugs.

rwoll:
Closing for now—we'll address this as part of the larger Best Practice overhaul!

/cc @debs-obrien

PR #14362, opened 2022-05-23
fix(ct): do not break if there are no components

pavelfeldman:


rwoll (approved):
Did this manifest as the `manifest.json` not existing error?

pavelfeldman:
> Did this manifest as the manifest.json not existing error?

We were not producing the bundle altogether.

PR #14363, opened 2022-05-23
fix(leak): do not retain test fixtures in worker fixtures

### LINKED ISSUE #14336: pageshot case memory leak

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/14336
Affects: https://github.com/microsoft/playwright/issues/6319

### DIFF
--- a/packages/playwright-test/src/fixtures.ts
+++ b/packages/playwright-test/src/fixtures.ts
@@ -53,18 +53,18 @@ type FixtureRegistration = {
 class Fixture {
   runner: FixtureRunner;
   registration: FixtureRegistration;
-  usages: Set<Fixture>;
   value: any;
 
   _useFuncFinished: ManualPromise<void> | undefined;
   _selfTeardownComplete: Promise<void> | undefined;
   _teardownWithDepsComplete: Promise<void> | undefined;
   _runnableDescription: FixtureDescription;
+  _deps = new Set<Fixture>();
+  _usages = new Set<Fixture>();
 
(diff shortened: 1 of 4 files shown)


packages/playwright-test/src/fixtures.ts
```diff
@@ -125,9 +130,13 @@ class Fixture {
     if (typeof this.registration.fn !== 'function')
       return;
     try {
-      for (const fixture of this.usages)
+      for (const fixture of this._usages)
         await fixture.teardown(timeoutManager);
-      this.usages.clear();
+      if (this._usages.size !== 0) {
+        // TODO: replace with assert.
```
yury-s, line 136: Why not do it right away?

yury-s (approved):

packages/playwright-test/src/fixtures.ts
pavelfeldman: I was thinking we merge it into 1.22

PR #14369, opened 2022-05-24
test(ct): start covering components with tests

pavelfeldman:


packages/playwright-test/src/plugins/vitePlugin.ts
```diff
@@ -113,6 +113,10 @@ export function createPlugin(
       if (hasNewTests || hasNewComponents || sourcesDirty)
         await fs.promises.writeFile(buildInfoFile, JSON.stringify(buildInfo, undefined, 2));
       previewServer = await preview(viteConfig);
+      const isAddressInfo = (x: any): x is AddressInfo => x?.address;
+      const address = previewServer.httpServer.address();
+      if (isAddressInfo(address))
```
rwoll, line 118: And if it's not AddressInfo, we silently do nothing? Under what circumstances do we not get an address from .address()?

tests/playwright-test/playwright.ct-react.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+      test('pass', async ({ mount }) => {
+        const component = await mount(<Button></Button>);
+        await expect(component).toHaveText('Button');
+      });
+    `,
+  }, { workers: 1 });
+
+  expect(result.exitCode).toBe(0);
+  expect(result.passed).toBe(1);
+});
```
rwoll, line 91: Can you add one more test that mixes JS, JSX, and TSX? For brand new projects, this combo is less likely but for projects in the middle of migration from JS -> TS, this is common.

packages/playwright-test/src/plugins/vitePlugin.ts
pavelfeldman: That would be an imaginary use case.

tests/playwright-test/playwright.ct-react.spec.ts
pavelfeldman: I will, but I need to land JS/JSX change first.

packages/playwright-test/src/plugins/vitePlugin.ts
rwoll: > Under what circumstances do we not get an address from .address()?

So none! It should always return an address—and the code assumes that—so why the fancy `is` type discriminator, instead of just a cast? Stylistic or am I missing something?

Issue #14385, opened 2022-05-24
[BUG] toHaveScreenshot types


rwoll:
From a user in Slack:

Hi, I started using toHaveScreenshot (1.22.2), I think that there could be a misspelling in animations  allowed values:
- In playwright.config.ts allowed values are: "allow" | "disable" | undefined
- In expect options allowed values are: "disabled"|"allow"
So if I set animations: "disable"  in my config file I obtain screenshotOptions.animations: expected one of (disabled|allow)

PR #14386, opened 2022-05-24
docs: fix toHaveScreenshot.animations doc

yury-s:
#14385


hw53 (commented):
If someone could help me figure out what these exactly mean I have about 4 medallions for someone that was from code they submitted and I'm not sure if it was an exploiting what I was posting to him and I am trying to understand it's just I fear it will be a while until I can fully comprehend the meaning of someone sending me these pull requests that were submitted as I only am a beginner and just have a personal account on GitHub and am an aspiring dev

PR #14412, opened 2022-05-25
fix(jsx): allow passing variables into mount

### LINKED ISSUE #14401: Issue with component testing / possibly intended behaviour
I have recently set up the experimental component testing feature of Playwright Test 1.22.2

I am using it with a React repo and have noticed some interesting behaviour when you pass in extracted components stored as consts.

Example

This works:

```
const alert = () => (
  <Alert type="info" title={'test'} id="alert" message={'hello'} />
)

test('should work', async ({ mount }) => {
  const component = await mount(alert())
  await expect(component).toContainText(alertMsg)
})
```

This fails:

```
const alert = (
  <Alert type="info" title={'test'} id="alert" message={'hello'} />
)

test('should work', async ({ mount }) => {
  const component = await mount(alert)
  await expect(component).toContainText(alertMsg)
})
```

In both cases the const appears to be a valid JSX.Element? The error is: 

```
undefined: TypeError: Cannot read properties of undefined (reading 'map')
    at render (http://localhost:3100/assets/index.22f7df41.js:2257:84)
    at window.playwrightMount (http://localhost:3100/assets/index.22f7df41.js:2274:19)
    at eval (eval at evaluate (:178:30), <anonymous>:19:25)
    at UtilityScript.evaluate (<anonymous>:180:17)
    at UtilityScript.<anonymous> (<anonymous>:1:44)
    at src/components/Alert/Alert.spec.pw.tsx:18:27
```

Interestingly, both of the above work with Cypress component testing. So wondering if Playwright is enforcing something specific here?

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14401

### DIFF
--- a/tests/playwright-test/playwright.ct-react.spec.ts
+++ b/tests/playwright-test/playwright.ct-react.spec.ts
@@ -240,3 +240,33 @@ test.fixme('should work with stray JS import', async ({ runInlineTest }) => {
   expect(result.exitCode).toBe(0);
   expect(result.passed).toBe(1);
 });
+
+test('should work with JSX in variable', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'playwright/index.html': `<script type="module" src="/playwright/index.js"></script>`,
+    'playwright/index.js': `
+      //@no-header
+    `,
+
+    'src/button.jsx': `
+      //@no-header
+      export const Button = () => <button>Button</button>;
+    `,
+
+    'src/button.test.jsx': `
+      //@no-header
+      import { test, expect } from '@playwright/experimental-ct-react';
+      import { Button } from './button';
+
+      const button = <Button></Button>;
+
+      test('pass button', async ({ mount }) => {
+        const component = await mount(button);
+        await expect(component).toHaveText('Button'); 
+      });
+    `,
+  }, { workers: 1 });
+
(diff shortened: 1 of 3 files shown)


tests/playwright-test/playwright.ct-react.spec.ts
```diff
@@ -240,3 +240,33 @@ test.fixme('should work with stray JS import', async ({ runInlineTest }) => {
...
+      import { test, expect } from '@playwright/experimental-ct-react';
+      import { Button } from './button';
+
+      const button = <Button></Button>;
+
+      test('pass button', async ({ mount }) => {
+        const component = await mount(button);
+        await expect(component).toHaveText('Button'); 
+      });
+    `,
```
rwoll, line 267: Let's add some tests that show the limits. I.e. cases where we can't do the correct thing (short of solving the halting problem). This will be nice to use for reference in docs.

rwoll (approved):

tests/playwright-test/playwright.ct-react.spec.ts
pavelfeldman: I need to start adding Vue / Svelte tests to show the limits. Otherwise, import-wise, React is somewhat limitless.

PR #14423, opened 2022-05-26
browser(webkit): rebase to 05/26/22 (r294871)

dpino:
Rebase `webkit/patches/bootstrap.diff` to [r294871](https://trac.webkit.org/changeset/r294871/webkit). Changes:

  * [r294820](https://trac.webkit.org/changeset/r294820/webkit) removed old WebStorage implementation (https://bugs.webkit.org/show_bug.cgi?id=240833). There were some downstream changes in several deleted files:
    - Source/WebKit/NetworkProcess/WebStorage/LocalStorageNamespace.cpp
    - Source/WebKit/NetworkProcess/WebStorage/LocalStorageNamespace.h
    - Source/WebKit/NetworkProcess/WebStorage/StorageArea.cpp
    - Source/WebKit/NetworkProcess/WebStorage/StorageArea.h
  * I kept the upstream changes, in other words, deleted those files and get rid of the downstream changes. AFAICT the functions added downstream were not used anywhere else. Tests pass so I believe that was the right thing to do.


yury-s:
> * I kept the upstream changes, in other words, deleted those files and get rid of the downstream changes. AFAICT the functions added downstream were not used anywhere else. Tests pass so I believe that was the right thing to do.

That implementation is incomplete and not yet used, so should be safe to remove.

yury-s (approved):
Assuming there were no other conflicts but the deleted storage files.

dpino:
Tha'ts correct, there were no conflicts except for the deleted files. I thought about committing the deleted files and remove them in a following commit, but it was acutally harder to read than the new boostrap.patch diff.

PR #14428, opened 2022-05-26
fix(ts): export types explicitly

### LINKED ISSUE #14348: [BUG] Types not compatible with TS4.7 Node ESM support

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14348

### DIFF
--- a/packages/playwright-chromium/package.json
+++ b/packages/playwright-chromium/package.json
@@ -13,6 +13,7 @@
   "license": "Apache-2.0",
   "exports": {
     ".": {
+      "types": "./index.d.ts",
       "import": "./index.mjs",
       "require": "./index.js"
     },
--- a/packages/playwright-core/package.json
+++ b/packages/playwright-core/package.json
@@ -13,6 +13,7 @@
   "license": "Apache-2.0",
   "exports": {
     ".": {
+      "types": "./index.d.ts",
       "import": "./index.mjs",
       "require": "./index.js"
     },
--- a/packages/playwright-ct-react/package.json
+++ b/packages/playwright-ct-react/package.json
@@ -12,8 +12,14 @@
   },
   "license": "Apache-2.0",
   "exports": {
-    "./register": "./register.mjs",
-    ".": "./index.js"
+    ".": {
+      "types": "./index.d.ts",
+      "default": "./index.js"
+    },
+    "./register": {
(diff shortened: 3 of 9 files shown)


yury-s (approved):
Please add a test for this.

rwoll (approved):

pavelfeldman:
I won't need a test once we move to TS 4.7. We have the opposite problem though, we might need bots for earlier versions of TS.

AviVahl:
any chance for a bugfix release with this change? just noticed it didn't resolve types in native esm mode (node16 module/moduleResolution)... :+1: 

was going to report it using the following reproduction steps:
```
mkdir playwright-node16-bug
cd playwright-node16-bug
npm init -y
npm i playwright-core typescript -D
npx tsc --init --module node16 --moduleResolution node16
echo "import playwright from 'playwright-core'; console.log(playwright);" > bug.mts
npx tsc
```

AviVahl:
thanks for the tip. :+1: would be useful in own projects, but my team is using some dependencies with stable playwright peerDeps. 

I'm currently using `"module": "esnext", "moduleResolution": "node"`, but wanted to switch to `"node16"`, as it actually works as I'd expect (picks up "exports").

PR #14472, opened 2022-05-29
Minor updates in the JSDoc of two files. 

AlexKomanov:
Adding JSDoc in two files.


pavelfeldman:
I don't think any types help us much!

PR #14521, opened 2022-05-31
feat: add serviceWorkerPolicy option

### LINKED ISSUE #14522: [Feature] Disable Service Workers
Easily allow disabling Service Workers which has been reported to be helpful when testing Web Apps.

rwoll:
Anecdotally, it's common for users to want to disable Service Workers
while testing, so this provides a simple context option to do so.

It also prepares us to land #14321 which requires us to have a point for
folks to enable seeing Service Worker Requests as events (and allow for
routing).

Resolves #14522


rwoll:
Closing—this will land together with https://github.com/microsoft/playwright/pull/14321

PR #14534, opened 2022-06-01
feat(expect): add ignoreCase option to toHaveText and toContainText

### LINKED ISSUE #13927: [Feature] Expect toHaveText & toContainText have  ignoreCase option
I think the title explains itself

```js
await expect(page.locator('body')).toHaveText('text', { ignoreCase: true });
await expect(page.locator('body')).toContainText('text', { ignoreCase: true });
```

I know there's an option to use regex, but anyway
```js
await expect(page.locator('body')).toHaveText(new RegExp('text', 'i'));
```

dgozman:
Fixes #13927.

### DIFF
--- a/docs/src/api/class-locatorassertions.md
+++ b/docs/src/api/class-locatorassertions.md
@@ -158,6 +158,11 @@ The opposite of [`method: LocatorAssertions.toContainText`].
 
 Expected substring or RegExp or a list of those.
 
+### option: LocatorAssertions.NotToContainText.ignoreCase
+- `ignoreCase` <[boolean]>
+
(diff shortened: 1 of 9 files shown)


rwoll (changes requested):
Let's either make it also apply to regex (approx `new RegExp(original, original.flags + 'i')`), or throw an error if used in combo.

This is already a redundant option (to regex), so let's minimize how it can be mis-understood/used.

rwoll (approved):
nit: consider changing _Overrides corresponding flag in regular expressions._ to _`ignoreCase` option takes precedence over the corresponding regex flag if specified._

Issue #14533, opened 2022-05-31
[BUG] array setup to use in a forEach loop


bshostak:
**Context:**
- Playwright Version: 1.21.1
- Operating System: Mac
- Node.js version: 16.14.2
- Browser: not applicable
- Extra: I use axios as the node module to make api calls as mentioned below.

**Code Snippet**

Help us help you! Put down a short code snippet that illustrates your bug and
that we can run and debug locally. For example:

```js
// helperFunctions.ts

export function usernameGen(length) {
    let result = '';
    const characters = `'你好喂èíàóえこれúñç@들은서ระกาศถ่ายทอดสดศึกฟุตบอล비스호출에추èящий%用語ăをțどん#ね.[]{}`;
    const charactersLength = characters.length;
    for (let i = 0; i< length; i++) {
        result += characters.charAt(Math.floor(Math.random() * charactersLength))
    }

    return result;
}

export function usernameArr(arrLength, usernameLength) {
    let arr = [];
    for (let i = 0; i < arrLength; i++) {
        arr.push(usernameGen(usernameLength));
    }
    return arr;
}
```

```js
// test.spec.ts

test.describe("Custom Username Login Tests", async () => {


   let names = usernameArr(4, 18);


     names.forEach(function (username) {
        test(`save a special character username, username saved is: ${username}`, async () => {
            
            let response = await putSaveUsername(username);

            await expect(response.status).toBe(200);
            await expect(response.data.status).toBe("SUCCESS");
          
        })
    })
})
```

**Describe the bug**

Hey guys, this might just be something Playwright can't do at execution or maybe I just don't understand but wanted to see if this was a bug or if this is something not supported/won't be a feature.

I'm doing a test that's a sort of hybrid API/UI test where I randomly generate some strings and then send those strings to an API to ensure it saves correctly and then on the UI side I ensure that string that was saved can also be used to login with. If you look at the code example above, you can see the setup where within the test file I am calling the function to setup the usernames I want to save. What happens when this starts running is it just constantly runs through this function and doesn't stop so it never gets to the test execution. If I do NOT call that usernameArr function and instead pass in hardcoded strings into the array, then the test works and runs successfully. However, having the ability to have these randomly generated would be nice but I think Playwright is limiting me doing that with the way it executes. 

Let me know if I can do this another way or if this is just something Playwright may not support / won't support.

Thank you!

kryaksy:
[...]

PR #14538, opened 2022-06-01
fix(playwright-test): prohibit async functions passed to describe

aslushnikov:
References #14533


mxschmitt (approved):

rwoll (approved):

dgozman (commented):
I wonder whether this is a breaking change? I can imagine someone has an async function, but does not actually do anything async today, and we'll start complaining.

PR #14543, opened 2022-06-01
tests: attempt to fix WebKit headed tests

aslushnikov:
It looks like GitHub Actions have default screen resolution of 1176 x 694, so
we can't open a browser window larger than this size.

This patch sets screen resolution to be at least 1280 x 1024.


aslushnikov:
Closing in favor of https://github.com/microsoft/playwright/pull/14578

PR #14555, opened 2022-06-01
feat: support multi-select/combo box with toHaveValue

### LINKED ISSUE #13824: [Feature] Have the ability to validate multiple options selected on a select element
I asked the following question on StackOverflow and received an answer to use page.$eval to save the selected options to a variable and verify that variable against the expected result. This feature request is to remove that necessity for validating a select element that allows multiple selections.

Given a page element:
```
<select id="colors" multiple name="selectedColors">
  <option value="R">Red</option>
  <option value="Y">Yellow</option>
  <option value="B">Blue</option>
</select>
```

I wish to validate that multiple items are selected, such that something similar to:
```
await expect(page.locator('#colors')).toHaveValue(['R', 'B']);
```
would work to validate that both R and B have been selected.

Whether it is "toHaveValue", or some other helper function, I'll leave that up to the developers.

rwoll:
Resolves #13824.

### DIFF
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -1051,6 +1051,31 @@ export class InjectedScript {
       }
     }
 
+    // Multi-Select/Combobox
+    {
+      if (expression === 'to.have.value' && options.expectedText?.length && options.expectedText.length >= 2) {
+        if (element.nodeName !== 'SELECT' || !(element as HTMLSelectElement).multiple)
+          throw this.createStacklessError('Not a select element with a multiple attribute');
+
+        const received = [...(element as HTMLSelectElement).selectedOptions].map(o => o.value);
+        if (received.length !== options.expectedText.length)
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1015,6 +1015,31 @@ export class InjectedScript {
...
+
+        const received = [...(element as HTMLSelectElement).selectedOptions].map(o => o.value);
+        if (received.length !== options.expectedText.length)
+          return { received, matches: false };
+
+        let i = 0;
+        const matchers = options.expectedText.map(e => new ExpectedTextMatcher(e));
+        let allMatchesFound = true;
+        for (const matcher of matchers) {
+          while (i < received.length && !matcher.matches(received[i]))
```
dgozman, line 1032: This seems to implement `toContain` logic, but I'd expect the exact match logic instead.

dgozman (approved):

PR #14565, opened 2022-06-01
chore: simplify context/page reuse in ct

dgozman:


packages/playwright-test/src/mount.ts
```diff
@@ -21,51 +21,48 @@ let boundCallbacksForMount: Function[] = [];
...
-  PlaywrightWorkerArgs & PlaywrightWorkerOptions & { _ctWorker: { page: Page | undefined, context: BrowserContext | undefined, hash: string, isolateTests: boolean } },
+  PlaywrightWorkerArgs & PlaywrightWorkerOptions & { _ctWorker: { context: BrowserContext | undefined, hash: string } },
   { _contextFactory: (options?: BrowserContextOptions) => Promise<BrowserContext> }> = {
 
-    _ctWorker: [{ page: undefined, context: undefined, hash: '', isolateTests: false }, { scope: 'worker' }],
-
-    context: async ({ _contextFactory, playwright, browser, _ctWorker, video, trace, viewport }, use, testInfo) => {
-      _ctWorker.isolateTests = shouldCaptureVideo(normalizeVideoMode(video), testInfo) || shouldCaptureTrace(normalizeTraceMode(trace), testInfo);
-      if (_ctWorker.isolateTests) {
-        await use(await _contextFactory());
```
pavelfeldman, line 32: I believe you need this line.

packages/playwright-test/src/mount.ts
```diff
@@ -21,51 +21,47 @@ let boundCallbacksForMount: Function[] = [];
...
-        const page = _ctWorker.page;
-        await (page as any)._wrapApiCall(async () => {
-          await (page as any)._resetForReuse();
-          await (page.context() as any)._resetForReuse();
-          await page.goto('about:blank');
-          await page.setViewportSize(viewport || { width: 1280, height: 800 });
-          await page.goto(process.env.PLAYWRIGHT_VITE_COMPONENTS_BASE_URL!);
-        }, true);
-        await use(page.context());
+        await (_ctWorker.context as any)._resetForReuse();
```
pavelfeldman, line 47: This will show up in tracing now.

It feels like starting with ToT and moving addInitScript/exposeFunction into context would already be as good as what you are doing here.

pavelfeldman: Actually, this is now mutually exclusive w/ tracing, so all good!

pavelfeldman (approved):

PR #14568, opened 2022-06-02
browser(firefox): force opener tab to be active

yury-s:
Pretty-diff: https://github.com/yury-s/gecko-dev/commit/d242fc7f2efe6325102a3d68d29c2edde00d0494
References: #14557


aslushnikov (approved):
lgtm % null-check for the docshell

yury-s:
> lgtm % null-check for the docshell

The code is inside the following if condition which guarantees that the doc shell is not null:

```
if (auto* browserChild = BrowserChild::GetFrom(root->GetDocShell())) {
```

PR #14600, opened 2022-06-02
feat(ct): support Vue2

### LINKED ISSUE #14143: [Feature] support Vue 2 in component testing
It looks like the current Vue component testing plugin targets Vue 3 only. It would be nice to have a Vue 2 version as well. Vue 3 only became the default version in the past 6 months or so, so I would venture that the majority of Vue apps are still on Vue 2. That ratio will, of course, begin to shift over the next few years, but to see significant Vue user adoption right now will probably need Vue 2 support.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14143


dgozman (approved):

rwoll:
Post-merge LGTM

PR #14606, opened 2022-06-02
browser(firefox): expose Response.fromServiceWorker

dgozman:
This property is "true" when network response was fulfilled by the service worker.

### DIFF
--- a/browser_patches/firefox-beta/juggler/NetworkObserver.js
+++ b/browser_patches/firefox-beta/juggler/NetworkObserver.js
@@ -533,6 +533,9 @@ class NetworkRequest {
       // remoteAddress is not defined for cached requests.
     }
 
+    const fromServiceWorker = this._networkObserver._channelIdsFulfilledByServiceWorker.has(this.requestId);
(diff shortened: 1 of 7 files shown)


browser_patches/firefox-beta/juggler/NetworkObserver.js
```diff
@@ -533,6 +533,9 @@ class NetworkRequest {
       // remoteAddress is not defined for cached requests.
     }
 
+    const fromServiceWorker = this._networkObserver._channelIdsFulfilledByServiceWorker.has(this.requestId);
+    this._networkObserver._channelIdsFulfilledByServiceWorker.delete(this.requestId);
+
```
yury-s, line 538: Do we need to remove request id from the set on failure as well?

yury-s (approved):

browser_patches/firefox-beta/juggler/NetworkObserver.js
dgozman: Since we listen to "service-worker-synthesized-response", I think it only happens when response is created, not on complete failures.

PR #14638, opened 2022-06-04
fix: hide DeprecationWarning of fd-slicer

### LINKED ISSUE #14065: Bug: DeprecationWarning: Buffer() is deprecated due to security and usability issues
```
(node:88279) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

During `npx playwright install`.

mxschmitt:
This patch fixes the deprecation warning which got emitted by `fd-slicer` during installation phase:

```
(node:336) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

Instead of rewriting all the `new Buffer(...)` usages of the bundle, only the `new Buffer(...)` usages inside `fd-slicer` gets rewritten. This is better since `new Buffer(...)` accepts `number` and `array` and replacing it all with `Buffer.alloc` or `Buffer.from` does not work. Needs to be done case by case.

Fixes https://github.com/microsoft/playwright/issues/14065

### DIFF
--- a/packages/playwright-core/bundles/zip/build.js
+++ b/packages/playwright-core/bundles/zip/build.js
@@ -1,0 +1,33 @@
+// @ts-check
+const path = require('path');
+const fs = require('fs');
+const esbuild = require('esbuild');
+
+// Can be removed once https://github.com/thejoshwolfe/yauzl/issues/114 is fixed.
+/** @type{import('esbuild').Plugin} */
+let patchFdSlicerToHideBufferDeprecationWarning = {
+  name: 'patch-fd-slicer-deprecation',
+  setup(build) {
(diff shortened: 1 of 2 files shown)


packages/playwright-core/bundles/zip/build.js
```diff
@@ -0,0 +1,33 @@
+// @ts-check
```
pavelfeldman, line 1: Add a license header

pavelfeldman (approved):

PR #14639, opened 2022-06-04
fix(ct): return locator to root for fragments

pavelfeldman:


packages/playwright-ct-react/registerSource.mjs
```diff
@@ -55,11 +69,5 @@ function render(component) {
 }
 
 window.playwrightMount = component => {
-  if (!document.getElementById('root')) {
-    const rootElement = document.createElement('div');
-    rootElement.id = 'root';
-    document.body.append(rootElement);
-  }
   ReactDOM.render(render(component), document.getElementById('root'));
```
mxschmitt, line 72: btw. `ReactDOM.render` is btw. deprecated in React 18, not sure which React version gets picked, but I think it will yield to console warnings if used with React 18.

packages/playwright-ct-svelte/registerSource.mjs
```diff
@@ -14,21 +14,25 @@
...
  */
 
+// @ts-check
+
 // This file is injected into the registry as text, no dependencies are allowed.
 
+/** @typedef {import('../playwright-test/types/component').Component} Component */
+/** @typedef {any} FrameworkComponent */
+
+/** @type { Map<string, FrameworkComponent> } */
```
mxschmitt, line 24:
```suggestion
/** @type {Map<string, FrameworkComponent>} */
```

packages/playwright-test/src/mount.ts
```diff
@@ -102,7 +103,17 @@ async function innerMount(page: Page, jsxOrType: any, options: any): Promise<str
...
     unwrapFunctions(component);
-    return await (window as any).playwrightMount(component);
+    let rootElement = document.getElementById('root');
+    if (!rootElement) {
+      rootElement = document.createElement('div');
+      rootElement.id = 'root';
+      document.body.appendChild(rootElement);
+    }
+
+    window.playwrightMount(component, rootElement);
```
mxschmitt, line 113: this was awaited before but not anymore. But seems intentional.

mxschmitt (approved):

packages/playwright-ct-react/registerSource.mjs
pavelfeldman: Thanks! I'll follow up!

packages/playwright-test/src/mount.ts
pavelfeldman: It is

PR #14645, opened 2022-06-04
chore: update to react18, but use react17 mode

pavelfeldman:


mxschmitt (approved):
Just a note: affected by that is for the end user trace-viewer and recorder. It's not about component testing if I understand it correctly.

Issue #14236, opened 2022-05-17
[BUG] Unable change snapshot naming for webkit


amrsa1:
**Context:**
- Playwright Version: 1.22
- Operating System: linux 'docker'
- Node.js version: 14.6
- Browser: webkit

**Code Snippet**

```
test.describe('Landing page Test Suite', () => {

    test.beforeEach(async ({ page, landingPage },testinfo) => {
        testinfo.snapshotSuffix = ''
        await page.goto(path.en.landingPage, { waitUntil: "load", timeout: 30000 })
    });

    test('Landing page is loading properly with all exact contents in EN lang', async ({ browserName, page }) => {
        expect(await page.screenshot({ fullPage: true, timeout: 4000 })).toMatchSnapshot(`${browserName}-EN.png`,{maxDiffPixels: 30,});
    });

```

**Describe the bug**

Test is failing in beforeEach hook for webkit when using  testinfo.snapshotSuffix = ''

<img width="1042" alt="Screenshot 2022-05-17 at 23 46 48" src="https://user-images.githubusercontent.com/32042806/168915522-1bfb0033-3fc8-4684-b7bb-7c07deb21ede.png">

rwoll:
I attempted to repro, but have been unable. The following works for me (changing the URL of the page you are visiting to https://example.com):

```
import { test, expect } from '@playwright/test';

test.describe('Landing page Test Suite', () => {
  test.beforeEach(async ({ page },testinfo) => {
      testinfo.snapshotSuffix = ''
      await page.goto('https://example.com', { waitUntil: "load", timeout: 30000 })
  });

  test('Landing page is loading properly with all exact contents in EN lang', async ({ browserName, page }) => {
      expect(await page.screenshot({ fullPage: true, timeout: 4000 })).toMatchSnapshot(`${browserName}-EN.png`,{maxDiffPixels: 30,});
  });
});
```

Can you try the following:

1. Run your snippet with the `testinfo.snapshotSuffix = ''` line removed. Does it still crash?
2. Run the snippet I included above as another test. Does it still crash? 

If (1) is yes, and (2) is no, it likely means Webkit doesn't like something about the site you are visiting. Please set `DEBUG` env var to `*` while running and repro'ing the crash to see if it shows any additional errors. If possible, can you share the URL you are testing against?

Thanks!

amrsa1:
@rwoll 
[...]

PR #14669, opened 2022-06-06
browser(webkit): do not use openssl on Linux for networking

aslushnikov:
Looks like OpenSSL doesn't work in Ubuntu aarch64: any navigation
to SSL site results in an "Internal WebKit Error".

References #14236

### DIFF
--- a/browser_patches/webkit/BUILD_NUMBER
+++ b/browser_patches/webkit/BUILD_NUMBER
@@ -1,2 +1,2 @@
-1655
-Changed: dpino@igalia.com Mon Jun  6 23:21:50 HKT 2022
+1656
+Changed: lushnikov@chromium.org Mon Jun  6 22:08:13 +03 2022
--- a/browser_patches/webkit/patches/bootstrap.diff
+++ b/browser_patches/webkit/patches/bootstrap.diff
@@ -22871,6 +22871,20 @@ index 017c992209e39859a0b56c1145bb79d9fdc6e939..5bf0744cfdcca58f6ecab6a9838abacf
      ruby
  
      # These are dependencies necessary for running tests.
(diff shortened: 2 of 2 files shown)


mxschmitt (commented):
Should we upstream that?

yury-s:
Reading the description of https://github.com/WebKit/WebKit/commit/f73c49acba44755c0404c162c236a6c7f6db3e14 I'd expect this change to break Ubuntu 18.04. Perhaps this dependency should be conditional? @dpino, wdyt?

aslushnikov:
@dpino some context: this PR fixes WebKit on **Ubuntu 20.04 aarch64** where any navigation to any HTTPS website results in "WebKit Internal Error".

dpino:
That's correct, this change will break Ubuntu 18.04 compilation because glib-networking requires a version of gnutls that is not available in Ubuntu 18.04. https://github.com/WebKit/WebKit/commit/f73c49acba44755c0404c162c236a6c7f6db3e14 upstreamed a change that was already present in Playwright's downstream patches (as it can be seen in this recent [roll conflict resolution](https://github.com/dpino/webkit/commit/9b66a5604e5c028cc534e070933c5515983d92c3)).

One solution, as Yury mentioned, is to use the openssl conditionally (only use it for Ubuntu 18.04, for example).

The ideal solution is to include the required gnutls version in `jhbuild.minimal` and build it together with the other libraries. We tried this solution a while ago but we had some trouble building gnutls and integrate it in jhbuild. I discussed this with @clopez and likely we will try this approach again.

dpino:
Something like this could work https://github.com/dpino/webkit/commit/ce2fae90076dc64d5165bd4c4a80b50b1b356c99 (not tested yet).

aslushnikov:
Closing in favor of https://github.com/microsoft/playwright/pull/14753

Issue #14190, opened 2022-05-16
[Question] Usage of test.slow() if beforeAll hook is slow


jithinjosejacob:
Can we use test.slow() annotation in beforeAll hook  if before all execution is slow. or is it mandatorily needs to be on test blocks. I am trying to verify few errors in nth page. Where I need to fill out all pages until page n in beforeAll hook

john-cyndx:
similarly, could there be a way to toggle or set the timeout countdown to begin only once the before hooks are done? I have tests where webkit is slow to load the page in the set up fixture and it means that my tests timeout randomly depending on the fixture load time. Could be a good feature

dgozman:
@john-cyndx Do you use `beforeAll` or `beforeEach`? For `beforeAll`, you can set a separate timeout with `test.setTimeout()`. For `beforeEach`, you can increate test timeout if you'd like. Take a look [at the documentation](https://playwright.dev/docs/next/api/class-test#test-set-timeout).

dgozman:
Sounds like everything is fixed here. Feel free to open a new issue and link to this one if there's still an issue.

PR #14670, opened 2022-06-06
docs: describe test.setTimeout inside beforeAll/afterAll

dgozman:
Adding documentation that `test.setTimeout()` call from the `beforeAll`/`afterAll` hook changes the timeout of the hook itself.

References #14190.

### DIFF
--- a/docs/src/test-api/class-test.md
+++ b/docs/src/test-api/class-test.md
@@ -943,7 +943,7 @@ test('very slow test', async ({ page }) => {
 });
 ```
 
-Changing timeout from a slow hook:
+Changing timeout from a slow `beforeEach` or `afterEach` hook. Note that this affects the test timeout that is shared with `beforeEach`/`afterEach` hooks.
 
 ```js js-flavor=js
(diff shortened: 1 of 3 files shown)


docs/src/test-api/class-test.md
```diff
@@ -943,7 +943,7 @@ test('very slow test', async ({ page }) => {
 });
 ```
 
-Changing timeout from a slow hook:
+Changing timeout from a slow `beforeEach` or `afterEach` hook. Note that this affects the test timeout that is shared with `beforeEach`/`afterEach` hooks.
```
rwoll, line 946: what does "test timeout" mean in this line? I think it's still a bit confusing.

rwoll (approved):

docs/src/test-api/class-test.md
rwoll: I think a small note pointing out this is different (and has different behavior) to `testInfo.setTimeout` could be helpful. They are named so similarly so I think that's where I was also getting lost.

PR #14696, opened 2022-06-08
feat(cli): option to record har

dgozman:
`npx playwright open --save-har=file.har` produces a HAR file that can be later inspected or used for mocks.
`--save-har-glob="**/*.{png,woff2}"` option can be used to produce a smaller har that only contains entries matching the glob.


pavelfeldman (approved):
I would match against glob and use glob as a part of the flag name.

Issue #13738, opened 2022-04-25
[Feature] Support Ubuntu 22.04


mxschmitt:
Blocked by:
- https://github.com/actions/virtual-environments/issues/5428

Requires the following changes so far:
- [x] Chromium
- [x] WebKit
- [x] Firefox
- [x] Docs changes that we support it https://github.com/microsoft/playwright/pull/14588
- [x] Regenerate required dependencies https://github.com/microsoft/playwright/pull/14588
- [x] Add tests on e.g. GitHub Actions https://github.com/microsoft/playwright/pull/14741
- [x] Evaluate if we want to add Ubuntu 22 Docker images as well
- [x] add Docker image arm64 amd64 tests
- [x] ~~don't suggest focal in the docs anymore, suggest 22.~~
- [x] Support GitHub Action on Ubuntu 22 https://github.com/microsoft/playwright-github-action/pull/66

jgilbert2017:
fyi- ubuntu 22.x install-deps fails. it seems as some packages were renamed:

ttf-unifont -> fonts-unifont
ttf-ubuntu-font-family -> fonts-freefont-ttf

that's as far as i get before the script aborts.

ddelange:
For the failures on jammy (22.04LTS):

```
#15 7.794 E: Unable to locate package libenchant1c2a
#15 7.794 E: Unable to locate package libicu66
#15 7.794 E: Unable to locate package libvpx6
#15 7.794 E: Unable to locate package libwebp6
#15 7.794 E: Package 'ttf-unifont' has no installation candidate
#15 7.794 E: Package 'ttf-ubuntu-font-family' has no installation candidate
```

I did some digging:

libenchant1c2a [>](https://packages.ubuntu.com/search?suite=jammy&arch=any&searchon=names&keywords=libenchant) libenchant-2-2
libicu66 [>](https://packages.ubuntu.com/search?suite=jammy&arch=any&searchon=names&keywords=libicu) libicu70
libvpx6 [>](https://packages.ubuntu.com/search?suite=jammy&section=all&arch=any&keywords=libvpx&searchon=names) libvpx7
libwebp6 [>](https://packages.ubuntu.com/search?suite=jammy&section=all&arch=any&keywords=libwebp&searchon=names) libwebp7
ttf-unifont [>](https://packages.ubuntu.com/search?suite=jammy&section=all&arch=any&keywords=unifont&searchon=names) fonts-unifont
[...]

PR #14710, opened 2022-06-08
browser(webkit): fix compilation on Ubuntu 22

mxschmitt:
https://github.com/microsoft/playwright/issues/13738


yury-s (approved):
I assume this is going to be upstreamed.

Also please attach a link to pretty diff, so that it's

PR #14729, opened 2022-06-09
chore: fix error message typo

joshbwlng:


rwoll (approved):
Thanks for the fix! 🎉

PR #14770, opened 2022-06-09
chore: fix WK ubuntu 22 deps on Docker

mxschmitt:
Fixes Docker tests in the pw-internal repo.

https://github.com/microsoft/playwright/issues/13738

### DIFF
--- a/packages/playwright-core/src/server/registry/nativeDeps.ts
+++ b/packages/playwright-core/src/server/registry/nativeDeps.ts
@@ -503,6 +503,7 @@ export const deps: any = {
       'libxtst6'
     ],
     webkit: [
+      'libenchant-2-2',
       'gstreamer1.0-libav',
       'gstreamer1.0-plugins-bad',
       'gstreamer1.0-plugins-base',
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/registry/nativeDeps.ts
```diff
@@ -642,7 +643,7 @@ export const deps: any = {
       'libicuuc.so.66': 'libicu70',
       'libicui18n.so.66': 'libicu70',
       'libwebp.so.6': 'libwebp6',
-      'libenchant.so.1': 'libenchant-2-2',
+      'libenchant-2.so.2': 'libenchant-2-2',
```
yury-s, line 646: Should we keep both of them? I don't think there is libenchant-2.so.2 on Ubuntu 20 or is there?

yury-s (approved):

packages/playwright-core/src/server/registry/nativeDeps.ts
yury-s: Oh, nm, it's 22.

PR #14771, opened 2022-06-09
feat(route): chain routes

### LINKED ISSUE #14625: [Feature] Allow request interception chaining (was 'based on request method')
Currently `page.route(url, handler, options)`/`browserContext.route(url, handler, options)` intercept routes based on url alone. It would be great if we have the ability to intercept based on any request properties like `method`, `header`, etc.

## Use case
A use case would be the common REST scenario where we want to intercept both `GET /items` and `POST /items`. 

Very simplified example:
```js
test('items', async ({ page }) => {
  // Intercept GET /api/items
  await page.route('/api/items', route => route.fulfill({ body: JSON.stringify([{ id: 1, name: 'Item 1' }]) }))
  
  // Intercept POST /api/items
  await page.route('/api/items', route => route.fulfill({ body: JSON.stringify({ id: 2, name: 'Item 2' }) }))
  
  await page.goto('/items')
  //  expect "Item 1" to be in the list
  
  await page.click('role=button[name="New item"]')
  // expect "Item 2" to be created in the list
})
```

The "matcher"(`url<string|RegExp|function(URL):boolean>`) doesn't allow for any matching on the request itself - only the request url. 
The "response"(`handler<function(Route, Request)>`) doesn't allow us to "skip" to the next interceptor - as the comment says in the source code `// There is no chaining, first handler wins.`. 

## Existing workarounds

1. Add single interceptor with conditional handler:

```js
await page.route('/api/items', (route, request) => { 
  if (request.method() === 'GET') {
    return route.fulfill({ body: JSON.stringify([{ id: 1, name: 'Item 1' }]) }) 
  } 
  
  if (request.method() === 'POST') {
    return route.fulfill({ body: JSON.stringify({ id: 2, name: 'Item 2' })
  } 
  
  return route.continue()
})
```
While this is enough in some cases, it's not very atomic and reusable. My idea is to make these a bit more generic and use them in fixtures - making sure that i can easily target just the request i need. Also, the idea of having the matching logic in the handler instead of the matcher just seems wrong.

2. Use `options.times` and order the interceptors based on their runtime execution. This seems very fragile and really not ideal.

## Proposal
Even though the contribution guide says "Expose as little information as needed. When in doubt, don’t expose new information.", i feel like we should have an escape hatch in the "matcher" function. Two options:
* Non-breaking change: adding the request as a second parameter to the url function:
   `url<string|RegExp|function(URL, request):boolean>`
   Not the best API but that's the price of non-breaking i guess.
* Breaking change: replacing the URL param with the request param in the url function:
    `url<string|RegExp|function(request):boolean>`
    This seems like a very robust and flexible way of route matching but it does break BC.

It's a bit unfortunate that we can't reuse the standard glob/regex url matching when we use the function. 
This is how i imagine it could look like:

```js
await page.route(
  (url, request) => request.method() === 'GET' && minimatch('/api/items', url.toSting()), 
  route => route.fulfill({ body: JSON.stringify([{ id: 1, name: 'Item 1' }]) }) 
)

await page.route(
  (url, request) => request.method() === 'POST' && minimatch('/api/items', url.toSting()), 
  route => route.fulfill({ body: JSON.stringify({ id: 2, name: 'Item 2' })
)
```

I think that having a convenient way to match by method(in addition to url) would be helpful in a lot of situations. 
What do you guys think? Do you have other APIs in mind to achieve this or do you think the current workaround methods are good enough?

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14625

### DIFF
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -216,7 +216,17 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
   }
 }
 
+type OverridesForContinue = {
+  url?: string;
+  method?: string;
+  headers?: Headers;
+  postData?: string | Buffer;
+};
+
 export class Route extends ChannelOwner<channels.RouteChannel> implements api.Route {
+  private _pendingContinueOverrides: OverridesForContinue | undefined;
+  private _routeChain: ((done: boolean) => Promise<void>) | null = null;
+
   static from(route: channels.RouteChannel): Route {
     return (route as any)._object;
   }
@@ -240,11 +250,21 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
     ]);
   }
 
+  _startHandling(routeChain: (done: boolean) => Promise<void>) {
+    this._routeChain = routeChain;
+  }
+
   async abort(errorCode?: string) {
     await this._raceWithPageClose(this._channel.abort({ errorCode }));
+    await this._followChain(true);
   }
 
   async fulfill(options: { response?: api.APIResponse, status?: number, headers?: Headers, contentType?: string, body?: string | Buffer, path?: string, har?: string } = {}) {
+    await this._innerFulfill(options);
+    await this._followChain(true);
+  }
+
+  private async _innerFulfill(options: { response?: api.APIResponse, status?: number, headers?: Headers, contentType?: string, body?: string | Buffer, path?: string, har?: string } = {}) {
     let fetchResponseUid;
     let { status: statusOption, headers: headersOption, body } = options;
 
@@ -314,15 +334,21 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
     }));
   }
 
-  async continue(options: { url?: string, method?: string, headers?: Headers, postData?: string | Buffer } = {}) {
-    await this._continue(options);
+  async continue(options: OverridesForContinue = {}) {
+    if (!this._routeChain)
+      throw new Error('Route is already handled!');
+    this._pendingContinueOverrides = { ...this._pendingContinueOverrides, ...options };
+    await this._followChain(false);
   }
 
-  async _internalContinue(options: { url?: string, method?: string, headers?: Headers, postData?: string | Buffer } = {}) {
-    await this._continue(options, true).catch(() => {});
+  async _followChain(done: boolean) {
+    const chain = this._routeChain!;
+    this._routeChain = null;
+    await chain(done);
   }
 
-  private async _continue(options: { url?: string, method?: string, headers?: Headers, postData?: string | Buffer }, isInternal?: boolean) {
+  async _finalContinue() {
+    const options = this._pendingContinueOverrides || {};
     return await this._wrapApiCall(async () => {
       const postDataBuffer = isString(options.postData) ? Buffer.from(options.postData, 'utf8') : options.postData;
       await this._raceWithPageClose(this._channel.continue({
@@ -331,11 +357,11 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
         headers: options.headers ? headersObjectToArray(options.headers) : undefined,
         postData: postDataBuffer ? postDataBuffer.toString('base64') : undefined,
       }));
-    }, isInternal);
+    }, !this._pendingContinueOverrides);
   }
 }
 
-export type RouteHandlerCallback = (route: Route, request: Request) => void | Promise<void>;
+export type RouteHandlerCallback = (route: Route, request: Request) => void;
 
 export type ResourceTiming = {
   startTime: number;
@@ -548,9 +574,10 @@ export class RouteHandler {
     return urlMatches(this._baseURL, requestURL, this.url);
   }
 
-  public handle(route: Route, request: Request): Promise<void> | void {
+  public handle(route: Route, request: Request, routeChain: (done: boolean) => Promise<void>) {
     ++this.handledCount;
-    return this.handler(route, request);
+    route._startHandling(routeChain);
+    this.handler(route, request);
   }
 
   public willExpire(): boolean {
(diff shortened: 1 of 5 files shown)


tests/library/browsercontext-route.spec.ts
```diff
@@ -249,3 +249,74 @@ it('should overwrite post body with empty string', async ({ context, server, pag
   const body = (await req.postBody).toString();
   expect(body).toBe('');
 });
+
+it('should chain continue', async ({ context, page, server }) => {
```
dgozman, line 253: Seems to be covered entirely by the previous test.

tests/library/browsercontext-route.spec.ts
```diff
@@ -249,3 +249,74 @@ it('should overwrite post body with empty string', async ({ context, server, pag
...
+  await context.route('**/empty.html', route => {
+    intercepted.push(3);
+    route.continue();
+  });
+  await page.goto(server.EMPTY_PAGE);
+  expect(intercepted).toEqual([3, 2, 1]);
+});
+
+it('should not chain fulfill', async ({ context, page, server }) => {
+  await context.route('**/empty.html', route => {
```
dgozman, line 272: Can we check that we also did not call this route function at all?

tests/library/browsercontext-route.spec.ts
```diff
@@ -249,3 +249,74 @@ it('should overwrite post body with empty string', async ({ context, server, pag
...
+    route.continue();
+  });
+  const response = await page.goto(server.EMPTY_PAGE);
+  const body = await response.body();
+  expect(body.toString()).toEqual('fulfilled');
+});
+
+it('should not chain abort', async ({ context, page, server }) => {
+  await context.route('**/empty.html', route => {
```
dgozman, line 287: Same here.

packages/playwright-core/src/client/page.ts
```diff
@@ -179,26 +179,22 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
...
   }
 
-  private _onRoute(route: Route, request: Request) {
+  private async _onRoute(route: Route, request: Request) {
     for (const routeHandler of this._routes) {
       if (!routeHandler.matches(request.url()))
         continue;
       // Immediately deactivate based on |times|.
       if (routeHandler.willExpire())
         this._routes.splice(this._routes.indexOf(routeHandler), 1);
```
dgozman, line 188: I think that splicing an array while iterating it will mess up with the iteration. Now that we do not break after first route, this matters.

dgozman (approved):

packages/playwright-core/src/client/network.ts
```diff
@@ -216,7 +216,17 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
   }
 }
 
+type OverridesForContinue = {
+  url?: string | undefined;
```
mxschmitt, line 220: nit: aren't all these `| undefined` unnecessary?

tests/page/page-route.spec.ts
```diff
@@ -837,3 +837,76 @@ for (const method of ['fulfill', 'continue', 'abort'] as const) {
...
+    intercepted.push(3);
+    route.continue();
+  });
+  await page.goto(server.EMPTY_PAGE);
+  expect(intercepted).toEqual([3, 2, 1]);
+});
+
+it('should not chain fulfill', async ({ page, server }) => {
+  await page.route('**/empty.html', route => {
+    route.continue();
```
mxschmitt, line 861: throw here or asserting something?

tests/page/page-route.spec.ts
```diff
@@ -837,3 +837,76 @@ for (const method of ['fulfill', 'continue', 'abort'] as const) {
...
+  });
+  const response = await page.goto(server.EMPTY_PAGE);
+  const body = await response.body();
+  expect(body.toString()).toEqual('fulfilled');
+});
+
+it('should not chain abort', async ({ page, server }) => {
+  await page.route('**/empty.html', route => {
+    route.continue();
```
mxschmitt, line 876: ditto

tests/page/page-route.spec.ts
```diff
@@ -837,3 +837,76 @@ for (const method of ['fulfill', 'continue', 'abort'] as const) {
...
+    route.abort();
+  });
+  await page.route('**/empty.html', route => {
+    route.continue();
+  });
+  const e = await page.goto(server.EMPTY_PAGE).catch(e => e);
+  expect(e).toBeTruthy();
+});
+
+it('should continue after exception', async ({ page, server }) => {
```
mxschmitt, line 888: what happens if the "first page.route" throws", we don't reach the "second page.route"?

mxschmitt (approved):

tests/page/page-route.spec.ts
pavelfeldman: Sure. I'm doing that in a browser context code, the same code serves it.

packages/playwright-core/src/client/network.ts
```diff
@@ -314,15 +334,21 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
     }));
   }
 
-  async continue(options: { url?: string, method?: string, headers?: Headers, postData?: string | Buffer } = {}) {
-    await this._continue(options);
+  async continue(options: OverridesForContinue = {}) {
+    if (!this._routeChain)
+      throw new Error('Route is already handled!');
+    this._pendingContinueOverrides = { ...this._pendingContinueOverrides, ...options };
```
PaperStrike, line 340: Is chaining the continue options an expected feature? I don't see where the the overrides get passed from the last route nor `continue` can be called multiple times in a route. 👀

PaperStrike, line 340: Is chaining the continue options an expected feature? I don't see where the the overrides get passed from the last route nor `continue` can be called multiple times in a route. 👀

Issue #11532, opened 2022-01-21
[BUG] Firefox DevTools does not decode Playwright's HAR response bodies of text resources


DetachHead:
**Context:**
## System:
 - OS: Windows 10 10.0.18363
 - Memory: 2.46 GB / 15.78 GB
## Binaries:
 - Node: 16.9.1 - C:\Program Files\nodejs\node.EXE
 - npm: 7.24.1 - C:\Program Files\nodejs\npm.CMD
## Languages:
 - Bash: 5.0.17 - C:\WINDOWS\system32\bash.EXE
## npmPackages:
 - playwright: ^1.18.0 => 1.18.0

**Code Snippet**

```javascript
const context = await browser.newContext({ recordHar: { path: 'asdf.har' } })
const page = await context.newPage()
await page.goto('https://google.com/')
await context.close()
```

**Describe the bug**

har file exported by playwright:
![image](https://user-images.githubusercontent.com/57028336/150447483-ecb30a41-4fd4-4544-97ee-4eb69777df9e.png)

har file exported manually:
![image](https://user-images.githubusercontent.com/57028336/150447712-c2d1e7f0-eed0-48df-8be0-2e0c0876ed3e.png)

mxschmitt:
This works when you load the HAR file into Chromium and does not work in Firefox. Seems like Firefox does not decode the text resources, even tho we say its base64. The [spec](http://www.softwareishard.com/blog/har-12-spec/) does not give any advice when to use base64 and when not, so we use it all the time.

DetachHead:
the reason i used firefox for those screenshots was because in chrome the base64 responses were not being displayed at all, which i assume is somehow related.

when exported from playwright:
![image](https://user-images.githubusercontent.com/57028336/150702360-ffed1aa0-cab2-4fc0-83e9-dc767d9679a6.png)

when exported manually:
![image](https://user-images.githubusercontent.com/57028336/150702408-039a2c77-9dda-4309-b782-b772b943a469.png)

but now that i try this on the google.com example i can't seem to reproduce it, only on the application i'm testing. it only seems to happen with sap batch requests

moeinsa:
good

alone-breecher:
![SVG_XSS](https://user-images.githubusercontent.com/76893049/176115009-e53ae94d-c80a-4e06-9424-a30644bbedfe.svg)

alone-breecher:
[...]

PR #14772, opened 2022-06-09
feat(har): store textual content without base64 encoding

dgozman:


mxschmitt:
Maybe this fixes https://github.com/microsoft/playwright/issues/11532 as well.

tests/library/har.spec.ts
```diff
@@ -248,21 +248,26 @@ it('should include secure set-cookies', async ({ contextFactory, httpsServer },
...
+  expect(log.entries[1].response.content.encoding).toBe(undefined);
   expect(log.entries[1].response.content.mimeType).toBe('text/css; charset=utf-8');
-  expect(Buffer.from(log.entries[1].response.content.text, 'base64').toString()).toContain('pink');
+  expect(log.entries[1].response.content.text).toContain('pink');
   expect(log.entries[1].response.content.size).toBeGreaterThanOrEqual(37);
   expect(log.entries[1].response.content.compression).toBe(0);
+
+  expect(log.entries[2].response.content.encoding).toBe('base64');
+  expect(log.entries[2].response.content.mimeType).toBe('image/png');
+  expect(Buffer.from(log.entries[2].response.content.text, 'base64')).toBeTruthy();
```
mxschmitt, line 268: isn't an empty Buffer also truthy and it would be better to assert the size here?

mxschmitt (approved):

DetachHead:
unfortunately it didn't. the batch requests from my issue are still base64 encoded and do not display correctly in either chrome or firefox

PR #14789, opened 2022-06-10
feat(fulfill): improve fulfilling from har

dgozman:
- `har` option is now an object `{ path, fallback }`.
- Allows falling back to `abort()`, `continue()` or throwing.
- Matches based on url + method.
- Follows redirects in the HAR file.
- Nice error/stack when throwing.
- Tests.

### DIFF
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -55,6 +55,11 @@ export type SetNetworkCookieParam = {
   sameSite?: 'Strict' | 'Lax' | 'None'
 };
 
+type RouteHAR = {
+  fallback?: 'abort' | 'continue' | 'throw';
+  path: string;
+};
+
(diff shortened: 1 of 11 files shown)


packages/playwright-core/src/client/network.ts
```diff
@@ -244,74 +249,89 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
...
-      length = buffer.length;
-    } else if (isString(body)) {
-      isBase64 = false;
-      length = Buffer.byteLength(body);
-    } else if (body) {
-      length = body.length;
-      body = body.toString('base64');
-      isBase64 = true;
-    }
+      let isBase64 = false;
```
pavelfeldman, line 301: Looks like lines 301 - 327 should be in the local util.

pavelfeldman (approved):

PR #14799, opened 2022-06-10
feat(server): consider 3XX responses as started server

### LINKED ISSUE #14793: [Feature] WebServerPlugin: add support for servers that have auth
Hi,

We're trying to use playwright/test's [testConfig.webServer](https://playwright.dev/docs/api/class-testconfig#test-config-web-server) option to start our application for testing.

This is working great except for the fact that only 2XX status codes are allowed.

Our application will respond with a 302 to unauthorised requests. This is because it redirects to another page so that the user can login. The consequence of this is that our server is never detected as up and playwright will timeout waiting for it to start, when in reality it has already started. 

I confirmed this was happening by setting DEBUG= pw:webserver and seeing that it was stuck in a loop of 302s.

```
  pw:webserver HTTP Status: 302 +5ms
  pw:webserver Waiting 1000ms +0ms
```

I suppose this could happen with other status codes such as 307, 401, 403 too.

I looked through your [code](https://github.com/microsoft/playwright/blob/29fd1d86df7d4d9e9b3e57d3e6b3f4d52231db0a/packages/playwright-test/src/plugins/webServerPlugin.ts#L156) and found that indeed only 200<=x<300 codes are allowed.

I can think of a few solutions to this:
A. Allow more codes by default
B. Make acceptable codes configurable (e.g add new option to testConfig.webServer)
C. Add a way for the availability check to get through the auth

Thoughts?

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14793

### DIFF
--- a/tests/playwright-test/web-server.spec.ts
+++ b/tests/playwright-test/web-server.spec.ts
@@ -546,3 +546,26 @@ test.describe('baseURL with plugins', () => {
     expect(result.passed).toBe(1);
   });
 });
+
+test('should treat 3XX as available server', async ({ runInlineTest }, { workerIndex }) => {
+  const port = workerIndex + 10500;
+  const result = await runInlineTest({
+    'test.spec.ts': `
+      const { test } = pwt;
+      test('pass', async ({}) => {});
+    `,
+    'playwright.config.ts': `
+      module.exports = {
+        webServer: {
+          command: 'node ${JSON.stringify(SIMPLE_SERVER_PATH)} ${port}',
+          url: 'http://localhost:${port}/redirect',
+        }
+      };
+    `,
+  }, {}, { DEBUG: 'pw:webserver' });
+  expect(result.exitCode).toBe(0);
+  expect(result.passed).toBe(1);
+  expect(result.output).toContain('[WebServer] listening');
+  expect(result.output).toContain('[WebServer] error from server');
+});
+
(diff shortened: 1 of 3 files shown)


tests/playwright-test/web-server.spec.ts
```diff
@@ -35,7 +35,7 @@ test('should create a server', async ({ runInlineTest }, { workerIndex }) => {
     'playwright.config.ts': `
       module.exports = {
         webServer: {
-          command: 'node ${JSON.stringify(SIMPLE_SERVER_PATH)} ${port}',
+          command: 'node ${JSON.stringify(SIMPLE_SERVER_PATH)} ${port} ${301}',
```
dgozman, line 38: I don't understand why 301 here.

dgozman (approved):

rwoll (changes requested):
Let's let user define what status code is "success" (or let them pass in a isReady function).

While this patch helps this specific report, what if another user reports—in a similar scenario—their server returns 401 or 403 (which are the very appropriate for saying unauthorized and forbidden)? I don't think Playwright should decide beyond the 200 range by default.

tests/playwright-test/web-server.spec.ts
pavelfeldman: Ah, I don't need this, left it by accident

pavelfeldman:
> Let's let user define what status code is "success" (or let them pass in a isReady function).

I was debating between <400 and <500 and considered that I'll wait for request about 401 before I allow it. Could you point me to the request?

PR #14805, opened 2022-06-11
chore: respect sigint in global setup

### LINKED ISSUE #14794: [BUG] WebServerPlugin: some processes are not killed properly during cleanup
**Context:**
- Playwright Version: 1.22.2
- Operating System: macOS 12.4
- Node.js version: 14.16.0
- Browser: Chromium
- Extra: We are using testConfig.webServer to start a java server via maven (openjdk version "17.0.1" / Apache Maven 3.8.4)

**Code Snippet**

```javascript
config: PlaywrightTestConfig = {
  webServer: {
    command: "mvn spring-boot:run",
    url: "http://localhost:9000",
    reuseExistingServer: true,
  },
```

**Describe the bug**

We are using testConfig.webServer to start a java server via maven, but it is never seems to be killed at the end of the test run or if the tests are cancelled half way through.

When we do this with an express server, the process is correctly killed strangely.

The way I am checking whether the server was killed is by running `lsof -i tcp:9000` and seeing whether there are any java processes left on that port.

Before we were using playwright, we wrote our own server setup / kill code and also had the exact same issue using the various nodejs APIs for killing processes. We switched to the [tree-kill](https://www.npmjs.com/package/tree-kill) library and that has worked flawlessly. 

Side note: There is another smaller inconsistency we've noticed between our express and java servers, which is that only the express server logs anything during startup. Setting DEBUG=pw:webserver makes the java server log stuff to console, but it's not ideal since it also logs playwright internals which we don't care too much about.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14794

### DIFF
--- a/packages/playwright-test/src/runner.ts
+++ b/packages/playwright-test/src/runner.ts
@@ -397,11 +397,10 @@ export class Runner {
     }
 
     // 13. Run Global setup.
-    const globalTearDown = await this._performGlobalSetup(config, rootSuite);
-    if (!globalTearDown)
-      return { status: 'failed' };
-
     const result: FullResult = { status: 'passed' };
+    const globalTearDown = await this._performGlobalSetup(config, rootSuite, result);
+    if (result.status !== 'passed')
+      return result;
 
     // 14. Run tests.
     try {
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/runner.ts
```diff
@@ -445,34 +445,49 @@ export class Runner {
...
+        ]);
+      }
     }, result);
 
-    if (result.status !== 'passed') {
+    sigintWatcher.disarm();
+
+    if (result.status !== 'passed' || sigintWatcher.hadSignal()) {
       await tearDown();
+      result.status = sigintWatcher.hadSignal() ? 'interrupted' : 'failed';
```
dgozman, line 490: Shouldn't this keep previous `result.status` if it was not `'passed'`?

dgozman (approved):

packages/playwright-test/src/runner.ts
pavelfeldman: That's consistent with the way it was.

PR #14809, opened 2022-06-12
Added excludeDefaultArguments to remove default args from electron launch

dquak:
I've added the `excludeDefaultArguments` argument to allow launching electron process without the default argument list.
This will give better versatile option to test electron-based projects, 
For example [Electron-Forge](https://www.electronforge.io/) which will otherwise crash if any of the default arguments it passed..

Creating electron-forge instance can be done by doing:
```
const electronForgePath = resolve(basePath, 'node_modules', '.bin', 'electron-forge.cmd');
return await electron.launch({
  args: ['start', '--inspect-electron', '--', '--remote-debugging-port=0'],
  executablePath: electronForgePath,
});
```

### DIFF
--- a/packages/playwright-core/src/protocol/channels.ts
+++ b/packages/playwright-core/src/protocol/channels.ts
@@ -3609,6 +3609,7 @@ export type ElectronLaunchParams = {
   },
   strictSelectors?: boolean,
   timezoneId?: string,
+  excludeDefaultArguments?: boolean
 };
 export type ElectronLaunchOptions = {
   executablePath?: string,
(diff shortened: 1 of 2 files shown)


mxschmitt:
Hey! We usually call it `ignoreDefaultArgs` like when launching a normal browser. The GitHub Actions are red, because you probably don't made enough changes.

You need to update: protocol.yml as well as docs/src/api/class-electron.md to add the param there, you can probably mostly copy it from: docs/src/api/params.md

Which exact args do you want to hide? I'm curious.

mxschmitt:
Are you still interested in working on this? Otherwise we'd close it since it looks stale.

PR #14829, opened 2022-06-13
chore: remove esm error interception, it only confuses users

pavelfeldman:


tests/playwright-test/loader.spec.ts
```diff
@@ -238,7 +238,7 @@ test('should fail to load ts from esm when package.json has type module', async
   });
 
   expect(result.exitCode).toBe(1);
-  expect(result.output).toContain('Cannot import a typescript file from an esmodule');
+  expect(result.output).toContain('Unknown file extension ".ts"');
```
dgozman, line 241: It would be ideal to point to our docs here, wdyt?

dgozman (approved):

tests/playwright-test/loader.spec.ts
pavelfeldman: I checked that searching exact error brings me to a good stack overflow page.

PR #14840, opened 2022-06-13
Add constraint to `DummyChannelOwner` for (upcoming) TypeScript 4.8

DanielRosenwasser:
Hi there! 👋

TypeScript recently added some stricter checking around unconstrained generics being incompatible with emptyish object types. The changes are only in nightly versions of TypeScript right now, but we've been exploring how open-source projects have been impacted [here](https://github.com/microsoft/TypeScript/issues/49461). In TypeScript 4.8, you'll get the following error without this change.

```
error TS2344: Type 'T' does not satisfy the constraint 'Channel'.
```

### DIFF
--- a/packages/playwright-core/src/client/connection.ts
+++ b/packages/playwright-core/src/client/connection.ts
@@ -56,7 +56,7 @@ class Root extends ChannelOwner<channels.RootChannel> {
   }
 }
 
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Deferring to #14847.

PR #14849, opened 2022-06-13
feat(fallback): allow falling back w/ overrides

pavelfeldman:


packages/playwright-core/src/client/network.ts
```diff
@@ -142,7 +157,9 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
    * @deprecated
    */
   headers(): Headers {
-    return this._provisionalHeaders.headers();
+    if (this._fallbackOverrides.headers)
```
yury-s, line 160: Remove the `if` block , `return this._fallbackOverrides.headers || this._provisionalHeaders.headers();` should be sufficient

packages/playwright-core/types/types.d.ts
```diff
@@ -14859,8 +14879,44 @@ export interface Route {
...
+     */
+    method?: string;
+
+    /**
+     * If set changes the post data of request
+     */
+    postData?: string|Buffer;
+
+    /**
+     * If set changes the request URL. New URL must have same protocol as original one.
```
yury-s, line 14916: Also explain how next routes will be matched?

tests/library/defaultbrowsercontext-2.spec.ts
```diff
@@ -223,35 +223,3 @@ it('should connect to a browser with the default page', async ({ browserType,cre
   expect(context.pages().length).toBe(1);
   await context.close();
 });
-
-it('route.continue should delete the origin header', async ({ launchPersistent, server, isAndroid, browserName }) => {
```
yury-s, line 227: Please revert this, see corresponding bug, the origin header is not deletable in some browsers.

yury-s (approved):

packages/playwright-core/src/client/network.ts
pavelfeldman: I need to normalize those (lower case names, etc)

packages/playwright-core/types/types.d.ts
```diff
@@ -14834,8 +14834,28 @@ export interface Route {
...
+   * ```js
+   * await page.route('**\/*', route => {
+   *   // Runs last.
+   *   route.abort();
+   * });
+   * await page.route('**\/*', route => {
+   *   // Runs second.
+   *   route.fallback();
+   * });
+   * await page.route('**\/*', route => {
```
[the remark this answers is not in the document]
pavelfeldman: Our generator escapes those...

tests/library/defaultbrowsercontext-2.spec.ts
pavelfeldman: It is a dupe of "should delete the origin header" in tests/page/page-request-continue.spec.ts

PR #14875, opened 2022-06-14
docs: explain how to test multiple users in the same test

### LINKED ISSUE #14872: [Question] Config doesn't accept multiple storageState files
**Context:**
- Playwright Version: 1.22.2
- Operating System: Manjaro Linux
- Node.js version: 16.5
- Browser: Chromium

According to https://playwright.dev/docs/test-auth#multiple-signed-in-roles, it should be possible to use multiple storageState files for different roles (need a total of 5 different roles, probably more later).
However, that doesn't seem to be the case, as `playwright.config.ts` only accepts one storageStage file argument.
It is neither possible to list multiple files, nor is it possible to have multiple storageState entries (not that I expected that to work).

Screenshot_20220614_175145
Screenshot_20220614_175244

```
  globalSetup: require.resolve('./globalSetup'),

  use: {
    storageState: 'AdminStorageState.json', 'userStorageState.json',
    ...
  }
```

I have also tried it with brackets but that doesn't seem to be correct either:
```
  use: {
    storageState: {'adminStorageState.json', 'userStorageState.json'},
    ...
  }
```

https://playwright.dev/docs/api/class-testoptions#test-options-storage-state also doesn't indicate a usage of multiple storageState files.

If it is indeed possible to use multiple storageState files, the doc is missing the required information on how to set them up in the config.

dgozman:
Fixes #14872.

### DIFF
--- a/docs/src/test-auth-js.md
+++ b/docs/src/test-auth-js.md
@@ -278,6 +278,181 @@ test.describe(() => {
 });
 ```
 
+### Testing multiple roles together
+
+If you need to test how multiple authenticated roles interact together, use multiple [BrowserContext]s and [Page]s with different storage states in the same test. Any of the methods above to create multiple storage state files would work.
+
+```js tab=js-ts
+import { test } from '@playwright/test';
+
+test('admin and user', async ({ browser }) => {
+  // adminContext and all pages inside, including adminPage, are signed in as "admin".
+  const adminContext = await browser.newContext({ storageState: 'adminStorageState.json' });
+  const adminPage = await adminContext.newPage();
+
+  // userContext and all pages inside, including userPage, are signed in as "user".
+  const userContext = await browser.newContext({ storageState: 'userStorageState.json' });
+  const userPage = await userContext.newPage();
+
+  // ... interact with both adminPage and userPage ...
+});
+```
+
+```js tab=js-js
+const { test } = require('@playwright/test');
+
+test('admin and user', async ({ browser }) => {
+  // adminContext and all pages inside, including adminPage, are signed in as "admin".
+  const adminContext = await browser.newContext({ storageState: 'adminStorageState.json' });
+  const adminPage = await adminContext.newPage();
+
+  // userContext and all pages inside, including userPage, are signed in as "user".
+  const userContext = await browser.newContext({ storageState: 'userStorageState.json' });
+  const userPage = await userContext.newPage();
+
+  // ... interact with both adminPage and userPage ...
+});
+```
+
+### Testing multiple roles with POM fixtures
+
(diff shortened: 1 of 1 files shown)


rwoll (commented):
This looks useful for one or two tests, but if you are testing a full app, we'd recommend introducing `page` and `context` fixtures for each of the users (e.g. `{ standardUserPage, standardUserContext, adminUserPage, adminUserContext }`). Do you think it's worth either elaborating on a fuller/advanced example below what you have or simply alluding to is and linking to the [Advanced Fixtures](https://playwright.dev/docs/test-fixtures) page?

rwoll:
> we'd recommend introducing `page` and `context` fixtures for each of the users (e.g. `{ standardUserPage, standardUserContext, adminUserPage, adminUserContext }`). 

Or introducing POM fixtures (`adminUser` and `standardUser` that each have a page/context hanging off of them).

To keep things simple, I think simply adding one line at the end of the section you added would suffice:

_If many of your tests require multiple user role sessions from within the same test, you can introduce fixtures for each role. See [Advanced Fixture](https://playwright.dev/docs/test-fixtures) and [Page Object Model](https://playwright.dev/docs/test-pom) for further reading._

docs/src/test-auth-js.md
```diff
@@ -278,6 +278,38 @@ test.describe(() => {
...
+
+```js tab=js-ts
+import { test } from '@playwright/test';
+
+test('two users chatting', async ({ browser }) => {
+  // page1 is signed in as user-1.
+  const page1 = await browser.newPage({ storageState: 'userStorageState-1.json' });
+
+  // page2 is signed in as user-2.
+  const page2 = await browser.newPage({ storageState: 'userStorageState-2.json' });
```
rwoll, line 293: Instead of using `browser.newPage` which has implicit `context` creation, let's make the example help the user understand `context`s and be a bit more verbose showing the `context`s created explicitly:


```ts
const adminUserContext = await browser.newContext({ ...contextOptions, storageState: 'userStorageState-admin.json' })
const adminUserPage = await browser.newContext({ ...contextOptions, storageState: 'userStorageState-adminUser.json' });

const standardUserContext = await browser.newContext({ ...contextOptions, storageState: 'userStorageState-standardUser.json' })
const standardUserPage = await browser.newPage();
```

azad-derakhshani-GS:
> Do you think it's worth either elaborating on a fuller/advanced example below?

Oh that's interesting, because I'm already using POM and user (and other) fixtures and I was wondering if I could somehow implement the storageStates as fixtures too. So I think a fuller example in the docs might actually be very useful for people. :)

rwoll (approved):
LGTM—thanks so for adding the advance examples, too!

PR #14893, opened 2022-06-15
chore: allow joining async handlers

pavelfeldman:


packages/playwright-core/src/utils/multimap.ts
```diff
@@ -38,6 +38,16 @@ export class MultiMap<K, V> {
     return this._map.has(key);
   }
 
+  delete(key: K, value: V) {
+    const values = this._map.get(key);
+    if (!values)
+      return;
+    const index = values.indexOf(value);
+    if (index !== -1)
+      values.splice(index, 1);
```
dgozman, line 47: I wonder why `values` is not a set...

dgozman (approved):

packages/playwright-core/src/utils/multimap.ts
pavelfeldman: I guess we did not want to clash the values. I can theoretically delete all matching.

PR #14897, opened 2022-06-15
fix(page): "load" event should fire before "waitForLoadState" resolves

dgozman:
Currently, `loadstate` and `load` are two separate events in the protocol, and are fired in this order. As a result, `waitForLoadState()` sometimes resolves before the `'load'` event is fired, which is unexpected.

Also fixes a flaky test that assumed `load` event comes after `domcontentloaded` for the empty page, which is not always a case in Chromium.


pavelfeldman (approved):
Let's remove server-side events as well.

PR #14902, opened 2022-06-16
docs: replace serviceWorkers workaround

### LINKED ISSUE #14901: [docs] replace all old references to Service Worker block workaround
Suggest new `serviceWorkes: 'block'` option instead of the navigator code snippet that's lingering in the docs.

rwoll:
Fixes #14901.

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -854,7 +854,7 @@ Routing provides the capability to modify network requests that are made by any 
 is enabled, every request matching the url pattern will stall unless it's continued, fulfilled or aborted.
 
 :::note
-[`method: Page.route`] will not intercept requests intercepted by Service Worker. See [this](https://github.com/microsoft/playwright/issues/1090) issue. We recommend disabling Service Workers when using request interception. Via `await context.addInitScript(() => delete window.navigator.serviceWorker);`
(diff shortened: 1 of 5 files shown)


docs/src/api/class-browsercontext.md
```diff
@@ -854,7 +854,7 @@ Routing provides the capability to modify network requests that are made by any
 is enabled, every request matching the url pattern will stall unless it's continued, fulfilled or aborted.
 
 :::note
-[`method: Page.route`] will not intercept requests intercepted by Service Worker. See [this](https://github.com/microsoft/playwright/issues/1090) issue. We recommend disabling Service Workers when using request interception. Via `await context.addInitScript(() => delete window.navigator.serviceWorker);`
+[`method: BrowserContext.route`] will not intercept requests intercepted by Service Worker. See [this](https://github.com/microsoft/playwright/issues/1090) issue. We recommend disabling Service Workers when using request interception by setting [`option: Browser.newContext.serviceWorkers`] to `'block'`.
 :::
```
rwoll, line 858: There's two problems:

1. Our linkRenderer doesn't handle this properly: 
```
[`option: Browser.newContext.serviceWorkers`]
```
2. If it's the JS-version, we should point to `TestOptions.serviceWorkers`, too.

(2) we shall punt since this applies to many of our docs, (1) we should fix. Likely somewhere around: https://github.com/microsoft/playwright/blob/main/utils/doclint/documentation.js#L774

yury-s: Does it at least render as `Browser.newContext.serviceWorkers` in the docs (I only see non-qualified references to options like [`option: timeout`])? Let's update the generator to linkify fully qualified options

yury-s (approved):

PR #14920, opened 2022-06-16
docs: mention chaining in locator.filter

### LINKED ISSUE #14881: Element selection, based on multiple inner selectors (that might not be in a parent/child relation)
Hi all,

I have a mat-table:
Column1 | Column2
a | 1
a | 2
b | 1
b | 2
b | 3
c | 1

If I want to check that there is a row that has a cell with value 'a', I can simply use:
await page.waitForSelector('mat-row >> mat-cell >> text="a"');
or
await page.locator(`mat-row >> mat-cell >> text="a"`).waitFor();
or
await page.locator('mat-row', { has: page.locator(`mat-cell >> text="a"`) }).waitFor();
and a few other ways.

But what if I want to check that there is a row that has a cell with value 'a' and another cell with value '1'?
I didn't find an easy solution for this.
It would be nice to be able to do something like:
await page.waitForSelector('mat-row >> (mat-cell >> text="a" && mat-cell >> text="1")');

It might be that there is already a simple way to do this, but I did not find it.

dgozman:
Fixes #14881.

### DIFF
--- a/docs/src/api/class-locator.md
+++ b/docs/src/api/class-locator.md
@@ -388,6 +388,52 @@ Value to set for the `<input>`, `<textarea>` or `[contenteditable]` element.
 - returns: <[Locator]>
 
 This method narrows existing locator according to the options, for example filters by text.
+It can be chained to filter multiple times.
+
+```js
+const rowLocator = page.locator('tr');
+// ...
+await rowLocator
+    .filter({ hasText: 'text in column 1' })
+    .filter({ has: page.locator('button', { hasText: 'column 2 button' }) })
+    .screenshot();
+```
+```java
+Locator rowLocator = page.locator("tr");
+// ...
+rowLocator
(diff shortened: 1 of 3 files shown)


docs/src/api/class-locator.md
```diff
@@ -388,6 +388,52 @@ Value to set for the `<input>`, `<textarea>` or `[contenteditable]` element.
...
+row_locator
+    .filter(has_text="text in column 1")
+    .filter(has=page.locator("tr", has_text="column 2 button"))
+    .screenshot()
+```
+```csharp
+var rowLocator = page.Locator("tr");
+// ...
+await rowLocator
+    .Filter(new LocatorFilterOptions { HasText = "text in column 1" })
```
yury-s, line 431: new LocatorFilterOptions => new () ?

yury-s (approved):

PR #14926, opened 2022-06-16
Add example for running tests against a project

BillBrower-Shopify:
This PR updates the `Playwright Test > Command line` documentation to add an example command for running all tests against a specific project, as defined in `playwright.config.ts`.

### DIFF
--- a/docs/src/test-cli-js.md
+++ b/docs/src/test-cli-js.md
@@ -14,11 +14,6 @@ Here are the most common options available in the command line.
   npx playwright test
   ```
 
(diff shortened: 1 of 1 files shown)


docs/src/test-cli-js.md
```diff
@@ -14,6 +14,11 @@ Here are the most common options available in the command line.
   npx playwright test
   ```
 
+- Run all the tests against a specific [project](./test-configuration-js.md#multiple-browsers)
```
pavelfeldman, line 17: Let's move it below, where --browser options are. You can also remove --browser versions while you are here.

pavelfeldman (approved):

PR #14933, opened 2022-06-16
chore: add CDN download fallbacks

mxschmitt:


packages/playwright-core/src/server/registry/download.ts
```diff
@@ -88,6 +88,7 @@ export async function download(
     log(
         `downloading ${progressBarName} - attempt #${attempt}`
     );
+    const url = Array.isArray(urls) ? urls[(attempt - 1) % urls.length] : urls;
```
pavelfeldman, line 91: This line above would make it a bit easier to read:
```js
if (!Array.isArray(urls))
  urls = [urls];
```

pavelfeldman (approved):

packages/playwright-core/src/server/registry/download.ts
rwoll, line 91: As is, if our primary is down, all our clients will fallback to Verizon. Is that what we want, or should the fallback order be randomized so the fallbacks (assuming they themselves are not down), see partial load?

packages/playwright-core/src/server/registry/index.ts
```diff
@@ -36,6 +36,12 @@ export { writeDockerVersion } from './dependencies';
 const PACKAGE_PATH = path.join(__dirname, '..', '..', '..');
 const BIN_PATH = path.join(__dirname, '..', '..', '..', 'bin');
 
+const PLAYWRIGHT_CDN_MIRRORS = [
```
rwoll, line 39: Let's consider adding install tests that induce testing fallback CDNs! Sort of like testing to make sure a database backup can actually be used if needed! :) 

Maybe a separate PR?

packages/playwright-core/src/server/registry/download.ts
```diff
@@ -99,14 +100,10 @@ export async function download(
...
-    if (
-      attempt < retryCount &&
-      (errorMessage.includes('ECONNRESET') ||
-        errorMessage.includes('ETIMEDOUT'))
-    ) {
+    if (attempt < retryCount) {
       // Maximum default delay is 3rd retry: 1337.5ms
       const millis = Math.random() * 200 + 250 * Math.pow(1.5, attempt);
-      log(`sleeping ${millis}ms before retry...`);
+      log(`sleeping ${millis.toFixed(2)}ms before retry...`);
```
aslushnikov, line 106: Sleeping was introduced as a heuristic: we noticed that our primary CDN was sometimes returning ETIMEDOUT or ECONNRESET for just-uploaded browser builds. So the idea was to wait a little in this case and then retry.

Since you have a bunch of different URLs now, sleeping does not make much sense to me and probably can be nuked

packages/playwright-core/src/server/registry/index.ts
```diff
@@ -701,12 +707,9 @@ export class Registry {
...
     if (hostPlatform === 'generic-linux' || hostPlatform === 'generic-linux-arm64')
       logPolitely('BEWARE: your OS is not officially supported by Playwright; downloading Ubuntu build as a fallback.');
-    const downloadHost =
-        (downloadHostEnv && getFromENV(downloadHostEnv)) ||
-        getFromENV('PLAYWRIGHT_DOWNLOAD_HOST') ||
-        'https://playwright.azureedge.net';
     const downloadPath = util.format(downloadPathTemplate, descriptor.revision);
-    const downloadURL = `${downloadHost}/${downloadPath}`;
+    const customHostOverride = (downloadHostEnv && getFromENV(downloadHostEnv)) || getFromENV('PLAYWRIGHT_DOWNLOAD_HOST');
+    const downloadURLs = PLAYWRIGHT_CDN_MIRRORS.map(mirror => `${customHostOverride || mirror}/${downloadPath}`) ;
```
aslushnikov, line 712: This would imply 3 attempts for the same custom Host url. Do we want this?

I'd just have an array of a single url if there's a custom host.

packages/playwright-core/src/server/registry/index.ts
rwoll: I'll follow up with the test! https://github.com/microsoft/playwright/issues/14958

❤️ 😄 🎭

PR #14934, opened 2022-06-16
feat(har): allow storing content as separate files

pavelfeldman:


packages/playwright-core/src/client/browserContext.ts
```diff
@@ -383,7 +383,7 @@ function prepareRecordHarOptions(options: BrowserContextOptions['recordHar']): c
     return;
   return {
     path: options.path,
-    omitContent: options.omitContent,
+    content: options.content,
```
dgozman, line 386: Support old `omitContent` as well.

dgozman (approved):

PR #14937, opened 2022-06-16
test: fix multiple focus highlights in ff codegen tests

dgozman:


tests/library/inspector/inspectorTest.ts
```diff
@@ -131,21 +131,27 @@ class Recorder {
...
+    let callback: Function | undefined;
+
+    const listener = async (msg: ConsoleMessage) => {
       const prefix = 'Highlight updated for test: ';
       if (msg.text().startsWith(prefix)) {
-        this.page.off('console', listener);
-        callback(msg.text().substr(prefix.length));
+        generatedSelector = msg.text().substr(prefix.length);
+        if (callback) {
+          this.page.off('console', listener);
```
yury-s, line 144: the listener should be removed unconditionally here

yury-s (approved):

tests/library/inspector/inspectorTest.ts
dgozman: The whole idea of this patch is to not remove the listener while `action()` is in progress.

PR #14964, opened 2022-06-18
chore: warn users on clashing test output and html reporter folders

### LINKED ISSUE #14957: Throw if html report folder clashes with test results folder

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14957

### DIFF
--- a/packages/playwright-test/src/reporters/html.ts
+++ b/packages/playwright-test/src/reporters/html.ts
@@ -131,9 +131,12 @@ class HtmlReporter implements Reporter {
   private config!: FullConfigInternal;
   private suite!: Suite;
   private _options: HtmlReporterOptions;
+  private _outputFolder!: string;
+  private _open: string | undefined;
 
   constructor(options: HtmlReporterOptions = {}) {
     this._options = options;
+    console.log('HTML REPORTER OPTIONS', options);
   }
 
   printsToStdio() {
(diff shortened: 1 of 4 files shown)


packages/playwright-test/src/reporters/html.ts
```diff
@@ -142,6 +144,25 @@ class HtmlReporter implements Reporter {
...
+    this._outputFolder = outputFolder;
+    this._open = open;
+    const reportedWarnings = new Set<string>();
+    for (const project of config.projects) {
+      if (outputFolder.startsWith(project.outputDir) || project.outputDir.startsWith(outputFolder)) {
+        const key = outputFolder + '|' + project.outputDir;
+        if (reportedWarnings.has(key))
+          continue;
+        reportedWarnings.add(key);
+        console.log(colors.red(`Configuration Error: HTML reporter output folder clashes with the tests output folder:`));
```
dgozman, line 157: A test for this should be easy, let's add one.

dgozman (approved):

packages/playwright-test/src/reporters/html.ts
```diff
@@ -142,6 +145,25 @@ class HtmlReporter implements Reporter {
 
   onBegin(config: FullConfig, suite: Suite) {
     this.config = config as FullConfigInternal;
+    const { outputFolder, open } = this._resolveOptions();
+    this._outputFolder = outputFolder;
+    this._open = open;
+    const reportedWarnings = new Set<string>();
+    for (const project of config.projects) {
+      if (outputFolder.startsWith(project.outputDir) || project.outputDir.startsWith(outputFolder)) {
```
thekbn, line 153: can someone explain the condition there? from what I understand the default test result is `test-results` and if I configure the html reporter with ` outputFolder: 'test-results/html-report'` then I should not see this error: 

image

because in the `onEnd` method [here](https://github.com/microsoft/playwright/pull/14964/files#diff-16ac82cfcd6cc32aee0a06c1f9f86c2c26d3cfc20857225da0fd3395e324cc06R187) we are removing the html folder and not the parent `test-results` folder

kalinkrustev:
It looks like `.startsWith` is not a good way to check. For example it will give incorrect warning for:
html reporter: /test/playwright-html
test results: /test/playwright

mxschmitt:
> It looks like `.startsWith` is not a good way to check. For example it will give incorrect warning for: html reporter: /test/playwright-html test results: /test/playwright

I'd recommend filing a bug on the repo.

PR #14967, opened 2022-06-19
chore: move har router into local utils

pavelfeldman:


packages/playwright-core/src/protocol/protocol.yml
```diff
@@ -478,6 +478,36 @@ LocalUtils:
...
+        disposition:
+          type: enum
+          literals:
+          - error
+          - redirect
+          - fulfill
+        message: string?
+        redirectURL: string?
+        status: number?
+        headers: json?
```
yury-s, line 503:
```
        headers:
          type: array?
          items: NameValue
```

packages/playwright-core/src/client/browser.ts
```diff
@@ -61,7 +61,7 @@ export class Browser extends ChannelOwner<channels.BrowserChannel> implements ap
 
   async newContext(options: BrowserContextOptions = {}): Promise<BrowserContext> {
     options = { ...this._browserType._defaultContextOptions, ...options };
-    const harRouter = options.har ? await HarRouter.create(options.har) : null;
+    const harRouter = options.har ? await HarRouter.create(this._connection.localUtils(), options.har) : null;
```
yury-s, line 64: Close it when the context closes.

yury-s (approved):

packages/playwright-core/src/client/browser.ts
pavelfeldman: This is already happening in addRoute.

PR #14996, opened 2022-06-20
fix(chromium): disable back-forward cache

dgozman:
Otherwise, back/forward navigation does not intercept requests.


rwoll (commented):
Discussed offline to better understand what this option means. It sounds safe to add this, but please defer to the others for actual approval on this one!

pavelfeldman (approved):

PR #15004, opened 2022-06-21
browser(webkit): fix Ubuntu arm64 build

aslushnikov:
Patch version is double-digit, so regex should account for this.

pretty diff: https://github.com/aslushnikov/WebKit/commit/f5183cc210de9c8f5a9ebf5928094c6dd6c09e1a

### DIFF
--- a/browser_patches/webkit/patches/bootstrap.diff
+++ b/browser_patches/webkit/patches/bootstrap.diff
@@ -2224,7 +2224,7 @@ diff --git a/Source/WebCore/Modules/speech/cocoa/WebSpeechRecognizerTask.mm b/So
 index a941d76a4f748718df1e3cff2a6c5e0827f48891..f62db5a27ac0e4c12430e7d19e60c83d768ace22 100644
 --- a/Source/WebCore/Modules/speech/cocoa/WebSpeechRecognizerTask.mm
 +++ b/Source/WebCore/Modules/speech/cocoa/WebSpeechRecognizerTask.mm
-@@ -198,6 +198,7 @@ - (void)sendEndIfNeeded
+@@ -198,6 +198,7 @@ NS_ASSUME_NONNULL_BEGIN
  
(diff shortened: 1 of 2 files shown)


aslushnikov:
cc @dpino

browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -20767,7 +20767,7 @@ diff --git a/Source/WebKitLegacy/mac/WebView/WebView.mm b/Source/WebKitLegacy/ma
 index 59cecf9242ab834dadc904ef295365e1476f47f9..ca4cc96e62df62e92c22c3535f5972cc1fdc4cba 100644
 --- a/Source/WebKitLegacy/mac/WebView/WebView.mm
 +++ b/Source/WebKitLegacy/mac/WebView/WebView.mm
-@@ -4039,7 +4039,7 @@ + (void)_doNotStartObservingNetworkReachability
+@@ -4039,7 +4039,7 @@ IGNORE_WARNINGS_END
```
mxschmitt, line 20770: can we dig into fixing this? We did this change like 6 times already and reverted it again because diffs on some computers produce different diffs.

mxschmitt (approved):

dpino (approved):

browser_patches/webkit/patches/bootstrap.diff
aslushnikov: So this is called a "hunk header", and according to [this answer](https://stackoverflow.com/a/28111535/314883) it is being constantly improved in almost every Git version. Specifically, C++ got improvements in Git 2.34, while I use Git 2.31. 

We can either:
- force everybody to use the same git version (meh)
- play with custom hunk headers, using the [`git diff --function-context`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---function-context)
- have a script that nukes all hunk headers from diff

I don't see this as a big-enough issue to do anything, so I'd do nothing. Let me know if I'm missing something
and it actually causes discomfort of some sort!

PR #15007, opened 2022-06-21
fix: Add Ubuntu 22.04 deps for arm64

Kurre:
Adds missing arm64/aarch64 deps for Ubuntu 22.04.
Tested with M1 Macbook + Docker.


mxschmitt:
Thanks!

Issue #13983, opened 2022-05-06
[Question] Reduce test name during execution (in console)


A-ZC-Lau:
Right now, when executing tests, it'll show the whole build up as one line, e.g.

`[browser] › filename:line › describe › describe › test_name`

The problem with this, is that if the console width isn't large enough, it'll cut off.

Is there a way to reduce the displayed name to just the test name? Or maybe even have indentation/grouping in the test display?

e.g.

```
filename:line
    > describe 
        > describe 
            > test_name
```

A-ZC-Lau:
@dgozman I'm just using playwright's default reporter.

But Jest's reporting basically does what I mentioned

MiguelSOliveira:
I would also love this feature.. I like to use nesting describes and everything ends up in the same line getting cut in the terminal.

Kranael:
Maybe a marker could be added like:

`test.describe(true|false //this here - true would be including the name, false would be excluding the name?, 'name', async ....)`

default would be like it is now?

PR #15052, opened 2022-06-22
fix(reporters): truncate long test titles from the start

dgozman:
Most useful information is at the end - test name, current step, retry. We truncate the repetitive project + suites at the start.

References #13983.


aslushnikov (approved):

diegohaz:
With this change, I can't differentiate the same tests in different projects:

Screenshot of terminal

Is this expected? If so, is there a way to revert it to the old style?

PR #15065, opened 2022-06-23
devops: use Android SDK 33 instead of 32

mxschmitt:
This should give us a more recent Chrome.

Blocked by an upstream bug: https://issuetracker.google.com/issues/236862869


pavelfeldman:
https://issuetracker.google.com/issues/236862869 bounced.

PR #15074, opened 2022-06-23
Update docs/src/intro-js.md

adg08101:
Update docs/src/intro-js.md
For selector 'text=Get Started' will not do it since
actual text is 'text=Get started'
Minor fix

### DIFF
--- a/docs/src/intro-js.md
+++ b/docs/src/intro-js.md
@@ -208,9 +208,9 @@ test('my test', async ({ page }) => {
   await expect(page).toHaveTitle(/Playwright/);
 
   // Expect an attribute "to be strictly equal" to the value.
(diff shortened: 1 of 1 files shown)


rwoll:
Thanks for the PR! However, the test passes as is.

`text=Get started` is the same as `text=Get Started`; however, `text="Get started"`  is different from `text="Get Started"`. The difference is the quotes _inside_ the `text=` selector. Without quotes, it's case insensitive; with quotes it's case sensitive. See https://playwright.dev/docs/next/selectors#text-selector for more info—although this subtle distinction is a bit buried in the explanation there! 

I'm going to close this, but we'll think about how we can make this more clear in the docs!

Thanks! 🎉 

/cc @debs-obrien

adg08101:
Awesome. Thanks.

PR #2177, opened 2020-05-11
feat(default context): support selected options for default context

dgozman:
- The only unsupported option is acceptDownloads.
- isMobile and locale require workaround to re-create the first page.

PR #15115, opened 2022-06-24
fix(default context): remove first page re-creation workaround for isMobile and locale

eran-rom:
This patch removes a workaround that was added in https://github.com/microsoft/playwright/pull/2177, about 2 years ago. The workaround fixed a couple of bugs in chrome and webkit:

> Workaround for:
> - chromium fails to change isMobile for existing page;
> - webkit fails to change locale for existing page.

No reference to the bugs in the browsers' bug-trackers was provided so it is impossible to know if the bugs were fixed by now so I simply removed that workaround code and ran the full tests suite, and no regression was observed.


eranr-talon:
Obviously the CI shows a regression in tests, I will investigate it and decide how to continue later.

PR #15119, opened 2022-06-25
fix(test runner): ignore undefined values in fixtures definitions

### LINKED ISSUE #14989: [BUG] Setting the use.trace option to `undefined` causes TypeError: Cannot read property 'mode' of undefined
## Context:
- Playwright Version: 1.22.2
- Operating System: Windows
- Node.js version: 14.18
- Browser: All

### Test file
```ts
// tests/example.test.ts
import { test } from '@playwright/test'


test('test', async () => {})
```

### Playwright Config

```ts
// playwright.config.ts
import { PlaywrightTestConfig } from "@playwright/test";

const playwrightTestConfig: PlaywrightTestConfig = {
  use: {
    trace: undefined,
  },
};

export default playwrightTestConfig;
```

**Describe the bug**

Running the above throws this error:

```
Running 1 test using 1 worker

  ✘  test/example.test.ts:4:1 › test (4ms)


  1) test/example.test.ts:4:1 › test ==================================================================

    TypeError: Cannot read property 'mode' of undefined




  1 failed
    test/example.test.ts:4:1 › test ===================================================================
```

This is caused by having `use,trace` as undefined in the playwright config.

This used to work with Playwright `v1.19`, and given the types of the option (`trace?: ...`), it should be allowed.

_After_ you figure out what the problem actually is, it is not a big deal as you can simply set the option to `off` instead of `undefined`. But it did take me a while to find that the trace option was causing it, as there is no stack trace in the error.

I got this right after upgrading to `v1.21`, and the reason the option is undefined is that I was setting it conditionally.

dgozman:
These mean "I don't want to specify this fixture/option" instead of "I want the value of undefined", aligned with how TypeScript works. We already do similar things in the config.

Fixes #14989.

### DIFF
--- a/tests/playwright-test/test-extend.spec.ts
+++ b/tests/playwright-test/test-extend.spec.ts
@@ -225,3 +225,43 @@ test('test._extendTest should print nice message when used as extend', async ({ 
   expect(result.passed).toBe(0);
   expect(result.output).toContain('Did you mean to call test.extend() with fixtures instead?');
 });
+
+test('fixture options should ignore undefined value', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'playwright.config.ts': `
+      module.exports = {
+        use: { option: undefined },
+      };
+    `,
+    'a.test.js': `
+      const test = pwt.test.extend({ option: [ 'default', { option: true } ] });
+      test('test1', async ({ option }) => {
+        console.log('test1-' + option);
+      });
+
+      test.describe('', () => {
+        test.use({ option: 'foo' });
+        test('test2', async ({ option }) => {
+          console.log('test2-' + option);
+        });
+
+        test.describe('', () => {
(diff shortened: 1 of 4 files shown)


tests/playwright-test/test-extend.spec.ts
```diff
@@ -225,3 +225,43 @@ test('test._extendTest should print nice message when used as extend', async ({
...
+      });
+
+      test.describe('', () => {
+        test.use({ option: 'foo' });
+        test('test2', async ({ option }) => {
+          console.log('test2-' + option);
+        });
+
+        test.describe('', () => {
+          test.use({ option: undefined });
```
yury-s, line 249: Shouldn't we throw in this case? I would expect such config to delete previously set value.

yury-s (approved):

Issue #14112, opened 2022-05-11
[Feature] "Accessibility testing" guidance docs


dbjorge:
My team ([Accessibility Insights](https://accessibilityinsights.io)) would be interested in contributing an official "Accessibility testing" page to the Guides section of https://playwright.dev/docs, comparable in size/scope/structure to the existing "API testing" guides available for [Python](https://playwright.dev/python/docs/api-testing) and [Java](https://playwright.dev/java/docs/api-testing).

We wanted to open this issue to build consensus on whether this would be reasonable to include among Playwright's docs and what the desired structure/placement/etc should be.

## Motivation

We are hopeful that providing official, first-party guidance will make it more likely that folks might think to use Playwright to incorporate accessibility testing earlier in the development cycle.

We think that these issues are good examples demonstrating community interest for such a guide:

* https://github.com/microsoft/playwright/issues/13338
* https://github.com/microsoft/playwright/issues/11621
* https://github.com/microsoft/playwright/issues/4750
* https://github.com/microsoft/playwright/issues/1638

## Using axe-core

We propose that the guidance should be based on using Deque Labs' [axe-core](https://github.com/dequelabs/axe-core), an industry-standard accessibility testing engine used by Accessibility Insights and Lighthouse, among [many others](https://github.com/dequelabs/axe-core/blob/develop/doc/projects.md).

`axe-core` itself is a testing engine which runs in the context of (each frame of) the page being scanned. Most folks using `axe-core` with browser automation frameworks do so via a language- and framework-specific "integration library", which encapsulates the process of using the browser automation framework to inject `axe-core` into each frame of the page under test, pass through any necessary configuration options, invoke it, and retrieve results back to the test context in a format suitable for writing test assertions and/or producing standalone reports.

Today, we are aware of the following major integration libraries for `axe-core` and Playwright:

* (Node.js) `@axe-core/playwright` ([src](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/README.md), [npm](https://www.npmjs.com/package/@axe-core/playwright)): Official integration library published by Deque
* (Node.js) `axe-playwright` ([src](https://github.com/abhinaba-ghosh/axe-playwright), [npm](https://www.npmjs.com/package/axe-playwright)): Unofficial third-party integration library which predates `@axe-core/playwright` and has [slightly more downloads](https://www.npmtrends.com/@axe-core/playwright-vs-axe-playwright) as of writing
[...]

Issue #13545, opened 2022-04-13
[Question] toMatchSnapshot to tests objects


diego-s-lopez:
Hi there! I'm trying to use "toMatchSnapshot" to test an object, but i receive "TypeError: file.slice is not a function".
Is there a way to accomplish this? It would be great to have the hability to do this, like with Jest, https://jestjs.io/docs/snapshot-testing#property-matchers

Thanks!

``` javascript
const { test, expect } = require("@playwright/test");

test("basic test", async ({ page }) => {
  expect({ a: 1, b: "Test" }).toMatchSnapshot();
});

```

aslushnikov:
@diego-s-lopez right now you'll need to do `JSON.stringify` first:

```js
const { test, expect } = require("@playwright/test");

test("basic test", async ({ page }) => {
  expect(JSON.stringify({ a: 1, b: "Test" })).toMatchSnapshot();
});
```

But that's a neat feature to have.

diego-s-lopez:
Hi there, thanks for your response!
Is there a way to use the "toMatchSnapshot()" function from Jest library?

pavelfeldman:
#### Why was this issue closed?

Thank you for your involvement. This issue was closed due to limited engagement (upvotes/activity), lack of recent activity, and insufficient actionability. To maintain a manageable database, we prioritize issues based on these factors.

If you disagree with this closure, please open a new issue and reference this one. More support or clarity on its necessity may prompt a review. Your understanding and cooperation are appreciated.

PR #15154, opened 2022-06-27
docs(accessibility-testing): create accessibility testing guide for js

dbjorge:
This PR partially implements #14112 by adding a new "Accessibility testing" guide for JavaScript, similar in size/scope to the existing API testing guide.

Per the discussion in #14112, this PR starts out by creating a js-only version of the guide based on Deque's `@axe-core/playwright` library. Once we're satisfied with the js version, I'll send a PR in microsoft/playwright.dev to add this to the sidebar and start working on a separate PR to create a similar guide for java based on Deque's `com.deque.html.axe-core:playwright` package.

The other similar docs I was using as examples were mixed about whether to provide just js-js example code vs both js-js and js-ts. In this initial commit, I've included only js-js where the code would be identical between the two, and both when there would be notable differences (particularly, the examples that include import vs require differences, and the custom fixture example where the js-ts version includes typings).

Per `CONTRIBUTING.md`, I verified that `npm run doc` passes without issue.

### DIFF
--- a/docs/src/accessibility-testing-js.md
+++ b/docs/src/accessibility-testing-js.md
@@ -190,24 +190,22 @@ function violationFingerprints(accessibilityScanResults) {
 }
 ```
 
-## Exporting a report as a test attachment
+## Exporting scan results as a test attachment
 
-There are many tools and libraries which support generating standalone reports for axe scan results. You can use [`testInfo.attach()`](./api/class-testinfo#test-info-attach) to attach these reports to your test results, which [reporters](./test-reporters) can embed or link as part of your test output.
+Most accessibility tests are primarily concerned with the `violations` property of the axe scan results. However, the scan results contain more than just `violations`. For example, the results also contain information about rules which passed and about elements which axe found to have inconclusive results for some rules. This information can be useful for debugging tests that aren't detecting all the violations you expect them to.
 
-The following example uses the [`axe-sarif-converter` package](http://npmjs.com/package/axe-sarif-converter) to produce a [SARIF report](https://sarifweb.azurewebsites.net/) and attach it to the test result:
+To include *all* of the scan results as part of your test results for debugging purposes, you can add the scan results as a test attachment with [`testInfo.attach()`](./api/class-testinfo#test-info-attach). [Reporters](./test-reporters) can then embed or link the full results as part of your test output.
+
+The following example demonstrates attaching scan results to a test:
 
 ```js
-import { test, expect } from '@playwright/test';
-import { convertAxeToSarif } from 'axe-sarif-converter';
-
-test('example with SARIF report attachment', async ({ page }, testInfo) => {
+test('example with attachment', async ({ page }, testInfo) => {
   await page.goto('https://your-site.com/');
 
   const accessibilityScanResults = await new AxeBuilder({ page }).analyze();
 
-  const sarifReport = convertAxeToSarif(accessibilityScanResults);
-  await testInfo.attach('sarif-report', {
-    body: JSON.stringify(sarifReport),
+  await testInfo.attach('accessibility-scan-results', {
+    body: JSON.stringify(accessibilityScanResults, null, 2),
     contentType: 'application/json'
   });
 
--- a/docs/src/accessibility-testing-js.md
+++ b/docs/src/accessibility-testing-js.md
@@ -39,8 +39,8 @@ This example demonstrates how to test an entire page for automatically detectabl
 import { test, expect } from '@playwright/test';
 import AxeBuilder from '@axe-core/playwright'; // 1
 
-test.describe('homepage', function() { // 2
-  test('should not have any automatically detectable accessibility issues', async function({ page }) {
+test.describe('homepage', () => { // 2
+  test('should not have any automatically detectable accessibility issues', async ({ page }) => {
     await page.goto('https://your-site.com/'); // 3
 
     const accessibilityScanResults = await new AxeBuilder({ page }).analyze(); // 4
@@ -54,8 +54,8 @@ test.describe('homepage', function() { // 2
 const { test, expect } = require('@playwright/test');
 const AxeBuilder = require('@axe-core/playwright').default; // 1
 
-test.describe('homepage', function() { // 2
-  test('should not have any automatically detectable accessibility issues', async function({ page }) {
+test.describe('homepage', () => { // 2
+  test('should not have any automatically detectable accessibility issues', async ({ page }) => {
     await page.goto('https://your-site.com/'); // 3
 
     const accessibilityScanResults = await new AxeBuilder({ page }).analyze(); // 4
@@ -74,7 +74,7 @@ For example, you can use [`AxeBuilder.include()`](https://github.com/dequelabs/a
 `AxeBuilder.analyze()` will scan the page *in its current state* when you call it. To scan parts of a page that are revealed based on UI interactions, use [Locators](./locators.md) to interact with the page before invoking `analyze()`:
 
 ```js
-test('navigation menu flyout should not have automatically detectable accessibility violations', async function({ page }) {
+test('navigation menu flyout should not have automatically detectable accessibility violations', async ({ page }) => {
   await page.goto('https://your-site.com/');
 
   await page.locator('button[aria-label="Navigation Menu"]').click();
@@ -98,7 +98,7 @@ You can constrain an accessibility scan to only run those rules which are "tagge
 Note that [automated testing cannot detect all types of WCAG violations](#disclaimer).
 
 ```js
-test('should not have any automatically detectable WCAG A or AA violations', async function({ page }) {
+test('should not have any automatically detectable WCAG A or AA violations', async ({ page }) => {
   await page.goto('https://your-site.com/');
 
   const accessibilityScanResults = await new AxeBuilder({ page })
@@ -126,7 +126,7 @@ This is usually the simplest option, but it has some important downsides:
 Here is an example of excluding one element from being scanned in one specific test:
 
 ```js
-test('should not have any accessibility violations outside of elements with known issues', async function({ page }) {
+test('should not have any accessibility violations outside of elements with known issues', async ({ page }) => {
   await page.goto('https://your-site.com/page-with-known-issues');
 
   const accessibilityScanResults = await new AxeBuilder({ page })
@@ -146,7 +146,7 @@ If your application contains many different pre-existing violations of a specifi
 You can find the rule IDs to pass to `disableRules()` in the `id` property of the violations you want to suppress. A [complete list of axe's rules](https://github.com/dequelabs/axe-core/blob/master/doc/rule-descriptions.md) can be found in `axe-core`'s documentation.
 
 ```js
-test('should not have any accessibility violations outside of rules with known issues', async function({ page }) {
+test('should not have any accessibility violations outside of rules with known issues', async ({ page }) => {
   await page.goto('https://your-site.com/page-with-known-issues');
 
   const accessibilityScanResults = await new AxeBuilder({ page })
@@ -197,7 +197,7 @@ The following example uses the [`axe-sarif-converter` package](http://npmjs.com/
 import { test, expect } from '@playwright/test';
 import { convertAxeToSarif } from 'axe-sarif-converter';
 
-test('example with SARIF report attachment', async function({ page }, testInfo) {
+test('example with SARIF report attachment', async ({ page }, testInfo) => {
   await page.goto('https://your-site.com/');
 
   const accessibilityScanResults = await new AxeBuilder({ page }).analyze();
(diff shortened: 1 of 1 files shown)


docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+- UI controls and form elements without labels that a screen reader could identify
+- Interactive elements with duplicate IDs which can confuse assistive technologies
+
+The following examples rely on the [`@axe-core/playwright`](https://npmjs.org/@axe-core/playwright) package which adds support for running the [axe accessibility testing engine](https://www.deque.com/axe/) as part of your Playwright tests.
+
+<!-- TOC -->
+
+## Disclaimer
+
+Automated accessibility tests can detect some common accessibility problems such as missing or invalid properties. But most accessibility problems can only be discovered through manual testing. We recommend using a combination of automated testing, manual accessibility assessments, and inclusive user testing.
```
WilcoFiers, line 19:
```suggestion
Automated accessibility tests can detect some common accessibility problems such as missing or invalid properties. But many accessibility problems can only be discovered through manual testing. We recommend using a combination of automated testing, manual accessibility assessments, and inclusive user testing.
```

The "most" thing is kind of disputed. I don't think anyone disputes "many" though.

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+// my-test-utils.js
+function violationFingerprints(accessibilityScanResults) {
+    const violationFingerprints = accessibilityScanResults.violations.map(violation => ({
+        rule: violation.id,
+        // These are CSS selectors which uniquely identify each element with
+        // a violation of the rule in question.
+        targets: violation.nodes.map(node => node.target),
+    }));
+
+    return JSON.stringify(violationFingerprints, null, 2);
```
dbjorge, line 186: N.B. this function can be simplified to just return the fingerprint object directly if/when #13545 is implemented

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+3. Uses normal Playwright syntax to navigate to the page under test
+4. Awaits `AxeBuilder.analyze()` to run the accessibility scan against the page
+5. Uses normal Playwright Test [expect] syntax to verify that there are no violations in the returned scan results
+
+```js tab=js-ts
+import { test, expect } from '@playwright/test';
+import AxeBuilder from '@axe-core/playwright'; // 1
+
+test.describe('homepage', function() { // 2
+  test('should not have any automatically detectable accessibility issues', async function({ page }) {
```
dgozman, line 43: All our example use arrow functions, let's be consistent.

```suggestion
  test('should not have any automatically detectable accessibility issues', async ({ page }) => {
```

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+  });
+});
+```
+
+```js tab=js-js
+const { test, expect } = require('@playwright/test');
+const AxeBuilder = require('@axe-core/playwright').default; // 1
+
+test.describe('homepage', function() { // 2
+  test('should not have any automatically detectable accessibility issues', async function({ page }) {
```
dgozman, line 58:
```suggestion
  test('should not have any automatically detectable accessibility issues', async ({ page }) => {
```

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+
+By default, axe checks against a wide variety of accessibility rules. Some of these rules correspond to specific success criteria from the [Web Content Accessibility Guidelines (WCAG)](https://www.w3.org/TR/WCAG21/), and others are "best practice" rules that are not specifically required by any WCAG criteron.
+
+You can constrain an accessibility scan to only run those rules which are "tagged" as corresponding to specific WCAG success criteria by using [`AxeBuilder.withTags()`](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/README.md#axebuilderwithtagstags-stringarray). For example, [Accessibility Insights for Web's Automated Checks](https://accessibilityinsights.io/docs/web/getstarted/fastpass/?referrer=playwright-accessibility-testing-js) only include axe rules that test for violations of WCAG A and AA success criteria; to match that behavior, you would use the tags `wcag2a`, `wcag2aa`, `wcag21a`, and `wcag21aa`.
+
+Note that [automated testing cannot detect all types of WCAG violations](#disclaimer).
+
+```js
+test('should not have any automatically detectable WCAG A or AA violations', async function({ page }) {
```
dgozman, line 101:
```suggestion
test('should not have any automatically detectable WCAG A or AA violations', async ({ page }) => {
```

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+### Example 2: Configuring axe to scan a specific part of a page
+
+`@axe-core/playwright` supports many configuration options for axe. You can specify these options by using a Builder pattern with the `AxeBuilder` class.
+
+For example, you can use [`AxeBuilder.include()`](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/README.md#axebuilderincludeselector-string--string) to constrain an accessibility scan to only run against one specific part of a page.
+
+`AxeBuilder.analyze()` will scan the page *in its current state* when you call it. To scan parts of a page that are revealed based on UI interactions, use [Locators](./locators.md) to interact with the page before invoking `analyze()`:
+
+```js
+test('navigation menu flyout should not have automatically detectable accessibility violations', async function({ page }) {
```
dgozman, line 77:
```suggestion
test('navigation menu flyout should not have automatically detectable accessibility violations', async ({ page }) => {
```

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+  await page.goto('https://your-site.com/');
+
+  await page.locator('button[aria-label="Navigation Menu"]').click();
+
+  await page.locator('#navigation-menu-flyout').waitFor();
```
dgozman, line 82: Let's note that it is important to `waitFor()` - otherwise we can run axe too early and it won't find the target element.

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+If your application contains a few specific elements with known issues, you can use [`AxeBuilder.exclude()`](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/README.md#axebuilderexcludeselector-string--string) to exclude them from being scanned until you're able to fix the issues.
+
+This is usually the simplest option, but it has some important downsides:
+* `exclude()` will exclude the specified elements *and all of their descendants*. Avoid using it with components that contain many children.
+* `exclude()` will prevent *all* rules from running against the specified elements, not just the rules corresponding to known issues.
+
+Here is an example of excluding one element from being scanned in one specific test:
+
+```js
+test('should not have any accessibility violations outside of elements with known issues', async function({ page }) {
```
dgozman, line 129:
```suggestion
test('should not have any accessibility violations outside of elements with known issues', async ({ page }) => {
```

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+If the element in question is used repeatedly in many pages, consider [using a test fixture](#using-a-test-fixture-for-common-axe-configuration) to reuse the same `AxeBuilder` configuration across multiple tests.
+
+### Disabling individual scan rules
+
+If your application contains many different pre-existing violations of a specific rule, you can use [`AxeBuilder.disableRules()`](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/README.md#axebuilderdisablerulesrules-stringarray) to temporarily disable individual rules until you're able to fix the issues.
+
+You can find the rule IDs to pass to `disableRules()` in the `id` property of the violations you want to suppress. A [complete list of axe's rules](https://github.com/dequelabs/axe-core/blob/master/doc/rule-descriptions.md) can be found in `axe-core`'s documentation.
+
+```js
+test('should not have any accessibility violations outside of rules with known issues', async function({ page }) {
```
dgozman, line 149:
```suggestion
test('should not have any accessibility violations outside of rules with known issues', async ({ page }) => {
```

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+
+    return JSON.stringify(violationFingerprints, null, 2);
+}
+```
+
+## Exporting a report as a test attachment
+
+There are many tools and libraries which support generating standalone reports for axe scan results. You can use [`testInfo.attach()`](./api/class-testinfo#test-info-attach) to attach these reports to your test results, which [reporters](./test-reporters) can embed or link as part of your test output.
+
+The following example uses the [`axe-sarif-converter` package](http://npmjs.com/package/axe-sarif-converter) to produce a [SARIF report](https://sarifweb.azurewebsites.net/) and attach it to the test result:
```
dgozman, line 194: I think this needs a bit more details. This is the first time SARIF is mentioned in our docs. Let's say that it's an easy way to see axe results, and tell how to see the results from the produced SARIF json with some nice UI.

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+
+```js
+import { test, expect } from '@playwright/test';
+import { convertAxeToSarif } from 'axe-sarif-converter';
+
+test('example with SARIF report attachment', async function({ page }, testInfo) {
```
dgozman, line 200:
```suggestion
test('example with SARIF report attachment', async ({ page }, testInfo) => {
```

docs/src/accessibility-testing-js.md
```diff
@@ -0,0 +1,292 @@
...
+  makeAxeBuilder: () => AxeBuilder;
+};
+
+// Extend base test by providing "makeAxeBuilder"
+//
+// This new "test" can be used in multiple test files, and each of them will get
+// a consistently configured AxeBuilder instance.
+export const test = base.extend<AxeFixture>({
+  makeAxeBuilder: async ({ page }, use, testInfo) => {
+    const makeAxeBuilder = () => new AxeBuilder({ page })
```
dgozman, line 243: I think this could be the AxeBuilder instance to make it even easier. WDYT?

```js
const axeBuilder = new AxeBuilder({ page }).....;
await use(axeBuilder);
```

Or, if you make it a function, let it have some options that I can pass from a test. Do you think that makes sense?

dgozman (commented):
What a wonderful PR! Just a few comments, and I think this is good to go.

docs/src/accessibility-testing-js.md
dbjorge: The original version of this example did actually provide an `AxeBuilder` instance directly instead of a factory function, but when I tested out the code in a working project, I found that doing that led to the following gotcha during usage:

```ts
test('example fixture usage with multiple scans per test', async ({ myPageModel, axeBuilder }) => {
  const dialogResults = await axeBuilder
    .include('#welcome-dialog')
    .disableRules(['color-contrast']) // known issue #123
    .analyze();
  expect(dialogResults.violations).toEqual([]);

  await myPageModel.dismissWelcomeDialog();

  // Oops: this looks like it's doing a new scan, but actually `color-contrast` is still disabled because it's the same axeBuilder instance
  const mainContentResults = await axeBuilder
    .include('main')
    .analyze();
  expect(mainContentResults.violations).toEqual([]);
});
```

I wanted the example fixture to be a pit of success, so I thought it felt a bit too dangerous to have the easiest way to run multiple scans per test lead to that gotcha. I considered explaining the reasoning behind using a factory function instead of a direct AxeBuilder within the example, but it felt like it required so much space to explain reasonably that it took away from the rest of the example.

As far as providing options, in general the factory still returns an axeBuilder instance and the usage example demonstrates how to apply additional options on top of it from within a test. Are you asking about providing options that pass-through to builder methods, or something else?

dgozman: Great point on the reuse, I didn't realize that! Factory it is then.

> Are you asking about providing options that pass-through to builder methods, or something else?

I thought maybe you have any ideas for the options, because I don't have anything specific in mind. But I guess if any option comes up later, we'll easily add it. Sorry for the noise.

docs/src/accessibility-testing-js.md
dbjorge: This is a great point. I tried adding more context, but it felt like I was shoehorning in a concept that was only tangentially related to the guide as a whole. I also found that for the specific use case of a SARIF report, I don't actually think that using a test attachment like this is a great idea; most other tools that integrate with SARIF expect to read a standalone SARIF file, which attaching as a test result like this doesn't actually necessarily produce.

Instead, I rewrote the section in terms of attaching the axe scan results without going through a separate report library. I think that's a more realistic use case for a test attachment; that information can be useful for debugging test cases that aren't finding all the violations a user expects.

docs/src/accessibility-testing-js.md
dbjorge: No worries, I appreciate the feedback! I definitely agree that if it weren't for the gotcha, avoiding the unnecessary layer of factory would be cleaner/better.

dgozman (approved):

PR #15162, opened 2022-06-28
docs: fix routing and har examples

rwoll:
This will also be cherry-picked to the 1.23 release.

### DIFF
--- a/docs/src/network.md
+++ b/docs/src/network.md
@@ -740,7 +740,7 @@ await context.close();
 ```java
 BrowserContext context = browser.newContext(new Browser.NewContextOptions()
     .setRecordHarPath(Paths.get("example.har"))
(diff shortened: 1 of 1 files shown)


docs/src/network.md
```diff
@@ -619,7 +619,7 @@ page.route("**/title.html", route -> {
 ```
 
 ```python async
-async def handle_route(route: Route) -> None:
+async def handle_route(route: Route, _: Request) -> None:
```
yury-s, line 622: Is it actually required in python?

yury-s (approved):

docs/src/network.md
rwoll: Actually, no. Just dropped this "fix" from the CL.

PR #15173, opened 2022-06-28
docs: fix previos -> previous

regseb:
Fix a typo in [`route.fallback([options])`](https://playwright.dev/docs/api/class-route#route-fallback) : _previos_.


mxschmitt:
you need to run npm run build, since it affects generated files.

rwoll:
Landed! Thanks for the fix! Cheer! 🎉

PR #15188, opened 2022-06-28
fix(evaluate): fallback to toJSON if it exists when serializing

### LINKED ISSUE #15106: [Question] Why is window.performance/performance undefined?
Hey there! 👋

I just created a new playwright project and called the following test. Can somebody explain why I can't use the performance api?

```bash
npm init playwright@latest new-project
```

```typescript
//example.spec.ts

test.beforeEach(async ({ page }) => {
  await page.goto("https://demo.playwright.dev/todomvc");
});

test.describe("New Todo", () => {
  test("should allow me to add todo items", async ({ page }) => {
    console.log(
      await page.evaluate(() => {
        window.performance.mark("start");
        window.performance.mark("end");
        return window.performance.measure("my-measure", "start", "end");
      })
    );
  });
});
```

This logs 'undefined' in every browser. Calling this within Chrome for example gives this expected result:

```javascript
{
    "name": "my-measure",
    "entryType": "measure",
    "startTime": 224071,
    "duration": 3577.699999988079
}
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/15106

### DIFF
--- a/tests/page/page-evaluate.spec.ts
+++ b/tests/page/page-evaluate.spec.ts
@@ -334,13 +334,14 @@ it('should properly serialize PerformanceMeasure object', async ({ page }) => {
   expect(await page.evaluate(() => {
     window.performance.mark('start');
     window.performance.mark('end');
-    return window.performance.measure('my-measure', 'start', 'end');
-  })).toEqual({
+    window.performance.measure('my-measure', 'start', 'end');
+    return performance.getEntriesByType('measure');
+  })).toEqual([{
     duration: expect.any(Number),
     entryType: 'measure',
     name: 'my-measure',
     startTime: expect.any(Number),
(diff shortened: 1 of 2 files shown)


tests/page/page-evaluate.spec.ts
```diff
@@ -330,6 +330,20 @@ it('should properly serialize null fields', async ({ page }) => {
   expect(await page.evaluate(() => ({ a: null }))).toEqual({ a: null });
 });
 
+it('should properly serialize PerformanceMeasure object', async ({ page }) => {
```
pavelfeldman, line 333: nit: add a test for `window.performance` itself as well

pavelfeldman (approved):

PR #15197, opened 2022-06-28
chore: allow updating har while routing

pavelfeldman:


packages/playwright-core/src/server/har/harTracer.ts
```diff
@@ -252,10 +256,12 @@ export class HarTracer {
   private async _onRequestFinished(request: network.Request, response: network.Response | null) {
     if (!response)
       return;
-    const page = request.frame()._page;
     const harEntry = this._entryForRequest(request);
     if (!harEntry)
       return;
+    const page = request.frame()._page;
+    if (this._page && page !== this._page)
```
dgozman, line 263: This check is probably not needed, given the one in `_onRequest`.

packages/playwright-core/src/server/har/harTracer.ts
```diff
@@ -108,7 +110,7 @@ export class HarTracer {
     return (request as any)[this._entrySymbol];
   }
 
-  private _ensurePageEntry(page: Page): har.Page | undefined {
+  private _createPageEntryIfNeeded(page: Page): har.Page | undefined {
```
dgozman, line 113: Let's add a `page !== this._page` check in this method to be extra careful.

dgozman (approved):

yury-s (approved):
As long as we are ok with 2 ways to record har the change looks good to me.

rwoll (commented):
LGTM, assuming update option as a `bool` is enough. I expected something like:

- `"always"`: create a new HAR from or overwrite the existing one
- `"never"`: use existing HAR
- `"on-missing"`: only update if it's missing on disk

However, maybe this is better handled at the @playwright/test level.

PR #15201, opened 2022-06-28
fix(test-runner): quoted fixtures

### LINKED ISSUE #15189: [BUG] [playwright-test] Fixture can't be wrapped by quotes
**Context:**
- Playwright Version: 1.22.2
- Operating System: Ubuntu
- Node.js version: 16.15.1
- Browser: All

**Code Snippet**
```js
const test2 = test.extend({
  "as-df": async ({}, test) => await test(123),
});
test2("should use asdf", ({ "as-df": asdf }) => {
  expect(asdf).toBe(123);
});
```

Get error: `Test has unknown parameter ""as-df""`

**Describe the bug**

Params in the test function can not be wrapped by quotes.

rwoll:
Fixes #15189.


rwoll:
This won't work with lots of things (e.g a fixture named `foo:bar`), so unless we bring in a proper parser and walk an AST, let's keep it restrictive as is for simplicity.

Issue #15202, opened 2022-06-28
[BUG] paths of test files in "playwright test --list" on Windows is screwed up


iva2k:
**Context:**
- Playwright Version: 1.23.0 (latest)
- Browser: All
- Extra:

## System:
 - OS: Windows 10 10.0.19044
 - Memory: 956.04 MB / 15.74 GB
## Binaries:
 - Node: 18.1.0 - C:\nodejs\node.EXE
 - Yarn: 1.22.15 - C:\nodejs\yarn.CMD
 - npm: 8.10.0 - C:\nodejs\npm.CMD
## Languages:
 - Bash: 4.4.23 - C:\Program Files\Git\usr\bin\bash.EXE (Windows CMD has same issue, so it is not git/bash)
 
**Describe the bug**

Paths of test files are screwed up in `playwright test --list`.

**Code Snippet**

```bash
playwright test --list
# Listing tests:
#  [chromium] › ..\..\..\..\C:\dev\playwright\playwright-ts1\tests\example.spec.ts:14:3 › New Todo should allow me to add todo items
#  [chromium] › ..\..\..\..\C:\dev\playwright\playwright-ts1\tests\example.spec.ts:37:3 › New Todo should clear text input field when an item is added
# ...
```
Expected normal path without `..\..\..\..\C:\` part.

Additional info:

- Was working correctly in playwright 1.22.2
- playwright-vscode can't run individual test files or open the test with errors, caused by path like "C:\c:\playwright\try\tests\example.spec.ts"

rwoll:
Thanks for the report—this looks like a regression, and will be triaged accordingly.

rwoll:
@iva2k **Can you link me to a repo I can clone and run to see the behavior?** I'll be able to debug more quickly. I tried repro'ing via:

```
> npm init playwright@latest
> npx playwright test --list
```

but the output is expected and the tests run.

Additionally, please elaborate on your file structure:

- Where is the (full) path to your package.json?
- Where is the (full) path to your test files?
- Where is the(full) path where you are running `npx playwright test --list` from?
- What does your playwright config look like if any?
- What is the (full) path to the config, if any?

Thanks!

rwoll:
Ah—just saw the repro you filed in https://github.com/microsoft/playwright/issues/15200. I'm not yet able to repro, but will keep trying. I've closed that issue to keep things centralized here.

rwoll:
I'm able to repro via:

```
> npm init playwright@latest repro
[...]

PR #15206, opened 2022-06-28
test: match more of output to ensure source map support is good

rwoll:
Tests issues in #15202, #15202 which are fixed by #15237.


tests/playwright-test/esm.spec.ts
```diff
@@ -143,5 +143,5 @@ test('should use source maps w/ ESM', async ({ runInlineTest, nodeVersion }) =>
   const output = stripAnsi(result.output);
   expect(result.exitCode).toBe(0);
   expect(result.passed).toBe(1);
-  expect(output).toContain('a.test.ts:7:7');
+  expect(output).toContain('[foo] › a.test.ts:7:7 › check project name');
```
rwoll, line 146: Looks like this is broken on windows with https://github.com/microsoft/playwright/pull/14561

tests/playwright-test/esm.spec.ts
```diff
@@ -135,13 +137,22 @@ test('should use source maps w/ ESM', async ({ runInlineTest, nodeVersion }) =>
...
-  expect(result.exitCode).toBe(0);
-  expect(result.passed).toBe(1);
-  expect(output).toContain('a.test.ts:7:7');
+  expect(result.exitCode).toBe(1);
+  expect(result.failed).toBe(1);
+  expect.soft(output, 'error carrot—via source maps—is positioned appropriately').toContain(
+      [
+        `    >  8 |         expect(1).toBe(2);`,
+        `         |                   ^`
+      ].join('\n'));
```
rwoll, line 153: The above output is what we get if we don't have `"type": "module"`, so—assuming source map support is working—I'd expect to see the same output. However, with `"type": "module"` I just see:

```
  1) [foo] › a.test.ts:7:7 › check project name ====================================================

    Expected: 2
    Received: 1

        at file:///Users/rosswollman/code/microsoft/playwright/test-results/esm-should-use-source-maps-w-ESM-playwright-test/a.test.ts:8:19
```

i.e. no code frame indicating some misalignment with source mappings. This can be observed on both macOS and Windows.

rwoll: ^ cc: @pavelfeldman What's expected behavior here? Is this a valid test?

mxschmitt (approved):

pavelfeldman (approved):

PR #15208, opened 2022-06-29
fix(network): remove races from sizes calculation

### LINKED ISSUE #15017: [internal]: fix CR protocol ordering handling when responseReceivedExtraInfo comes after loadingFinished
repro:

` FORCE_COLOR=0 DEBUG=pw:protocol npm run ctest -- --headed tests/library/har.spec.ts:358 --repeat-each 1 --reporter=list --max-failures 1 --workers 1 &> test.txt `

telegram-cloud-photo-size-2-5300952528143367152-y

`Network.responseReceivedExtraInfo` comes in after `Network.loadingFinished`

dgozman:
- Do not resolve raw headers upon `loadingFinished`, since they may still come later in `responseReceivedExtraInfo`.
- Introduce separate promises for `encodedBodySize`, `transferSize` and `responseHeadersSize`.
- Make sure we resolve each of the above with data available from the browser, or fallback to manual calculation.
- Set raw response headers for redirects on WebKit.
- Do not stall on cached responses in Chromium, they have erroneously set `hasExtraInfo` flag.
- Use `transferSize` that is available in Firefox protocol.

Fixes #15017.

### DIFF
--- a/packages/playwright-core/src/server/webkit/wkPage.ts
+++ b/packages/playwright-core/src/server/webkit/wkPage.ts
@@ -18,7 +18,7 @@
 import path from 'path';
 import { PNG, jpegjs } from '../../utilsBundle';
 import { splitErrorMessage } from '../../utils/stackTrace';
-import { assert, createGuid, debugAssert, headersArrayToObject, headersObjectToArray } from '../../utils';
+import { assert, createGuid, debugAssert, headersArrayToObject } from '../../utils';
 import { hostPlatform } from '../../utils/hostPlatform';
 import type * as accessibility from '../accessibility';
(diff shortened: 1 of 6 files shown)


packages/playwright-core/src/server/webkit/wkPage.ts
```diff
@@ -1080,6 +1073,9 @@ export class WKPage implements PageDelegate {
     if (!request)
       return;
 
+    // Use provisional headers if we didn't have the response with raw headers.
+    request.request.setRawRequestHeaders(null);
```
yury-s, line 1077: move it to else branch below for clarity

yury-s (approved):

PR #15226, opened 2022-06-29
fix(reporter): add truthy check for attachment.contentType in formatFailure

gabkeib:
When some tests fail, report generation crashes if one of the attachments has the contentType equal to the 'undefined'.


dgozman:
@gabkeib I think `contentType` should never be `undefined`. Do you have an example repro where this is the case?

gabkeib:
@dgozman After deeper investigation we found a bug on our side. Thank you for pointing this out.

PR #15229, opened 2022-06-29
docs: fix code block example

michael-siek:


rwoll:
Thanks for the PR! However, with the change, the example is broken. `devices` is used a few lines below the fold as more than just a type, so this change would cause it to be missing.

PR #15254, opened 2022-06-30
devops: adapt repack-juggler script to work with win

aslushnikov:
This patch:
- Uses some folder in `CWD` instead of `/tmp` on win32 to store
  builds
- Drops usage of `find`, `zip` and `unzip` posix tools. Instead, rely
  on `adm-zip` package.

### DIFF
--- a/browser_patches/repack-juggler.mjs
+++ b/browser_patches/repack-juggler.mjs
@@ -193,7 +193,7 @@ async function repackageJuggler(browserName, buildInfo) {
     zip.writeZip(omniWithJugglerPath);
   }
 
-  const module = await import(path.join(__dirname, browserName, 'install-preferences.js'));
+  const module = await import(URL.pathToFileURL(path.join(__dirname, browserName, 'install-preferences.js')));
   await module.default.installFirefoxPreferences(path.join(BUILD_DIRECTORY, 'firefox'));
 
   // Output executable path to be used in test.
--- a/browser_patches/repack-juggler.mjs
+++ b/browser_patches/repack-juggler.mjs
@@ -23,6 +23,7 @@ import * as https from 'https';
 import * as os from 'os';
 import * as util from 'util';
 import * as child_process from 'child_process';
+import AdmZip from 'adm-zip';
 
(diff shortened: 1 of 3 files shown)


browser_patches/repack-juggler.mjs
```diff
@@ -49,7 +50,7 @@ if (process.argv[2] === 'firefox' || process.argv[2] === 'ff') {
 // Path to jar.mn in the juggler
 const JARMN_PATH = path.join(__dirname, browserName, 'juggler', 'jar.mn');
 // Workdir for Firefox repackaging
-const BUILD_DIRECTORY = `/tmp/repackaged-firefox`;
+const BUILD_DIRECTORY = os.platform() === 'win32' ? path.join(__dirname, '__repackaged_firefox__') : `/tmp/repackaged-firefox`;
```
mxschmitt, line 53:
```suggestion
const BUILD_DIRECTORY = path.join(os
tmpdir(), '__repackaged_firefox__');
```

Why not like that?

browser_patches/repack-juggler.mjs
```diff
@@ -108,27 +109,48 @@ async function ensureFirefoxBuild(browserName, buildNumber, buildPlatform) {
...
+  let downloadedPercentage = 0;
+  await downloadFile(url, buildZipPath, (downloaded, total) => {
+    const percentage = Math.round(downloaded / total * 10) * 10;
+    if (percentage === downloadedPercentage)
+      return;
+    downloadedPercentage = percentage;
+    console.log(`Downloaded: ${downloadedPercentage}%`);
+  });
+
+  const zip = new AdmZip(buildZipPath);
```
mxschmitt, line 121: Don't we have extract-zip already for it as a dependency or does this do a better job?

mxschmitt (approved):

browser_patches/repack-juggler.mjs
aslushnikov: extract-zip only extracts; this also compresses & lists files.

dgozman (approved):

PR #15262, opened 2022-06-30
fix(esm): show codeframe when errors get reported

### LINKED ISSUE #15244: Codeframe is not shown when ESM mode is used
fixme with one of these tests: https://github.com/microsoft/playwright/pull/15206/files

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/15244

Abstract: stack-utils return file locations as `file://` in ESM mode. We handled this case already correctly inside `playwright-core`, and this patch applies the same fix inside `playwright-test`.

drive-by: replace `new URL(frame.file).pathname` with `url.fileURLToPath`

### DIFF
--- a/packages/playwright-core/src/utils/stackTrace.ts
+++ b/packages/playwright-core/src/utils/stackTrace.ts
@@ -15,6 +15,7 @@
  */
 
 import path from 'path';
+import url from 'url';
 import { StackUtils } from '../utilsBundle';
 import { isUnderTest } from './';
 
@@ -87,12 +88,8 @@ export function captureStackTrace(rawStack?: string): ParsedStackTrace {
       return null;
     if (isInternalFileName(frame.file, frame.function))
       return null;
-    // Workaround for https://github.com/tapjs/stack-utils/issues/60
-    let fileName: string;
-    if (frame.file.startsWith('file://'))
-      fileName = new URL(frame.file).pathname;
-    else
-      fileName = path.resolve(process.cwd(), frame.file);
+    // ESM files return file:// URLs, see here: https://github.com/tapjs/stack-utils/issues/60
+    const fileName = frame.file.startsWith('file://') ? url.fileURLToPath(frame.file) : path.resolve(process.cwd(), frame.file);
     if (isTesting && fileName.includes(COVERAGE_PATH))
       return null;
     const inCore = fileName.startsWith(CORE_LIB) || fileName.startsWith(CORE_SRC);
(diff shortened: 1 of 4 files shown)


tests/playwright-test/esm.spec.ts
```diff
@@ -148,7 +148,6 @@ test('should use source maps', async ({ runInlineTest, nodeVersion }) => {
 });
 
 test('should show the codeframe in errors', async ({ runInlineTest, nodeVersion }) => {
-  test.fixme();
```
rwoll, line 151: 😎 Nice! I know I added this test specifically for the ESM case, but do we have one testing the non-ESM case, too?

rwoll (approved):
Assuming our new test passes, LGTM!

packages/playwright-core/src/utils/stackTrace.ts
```diff
@@ -87,12 +88,8 @@ export function captureStackTrace(rawStack?: string): ParsedStackTrace {
       return null;
     if (isInternalFileName(frame.file, frame.function))
       return null;
-    // Workaround for https://github.com/tapjs/stack-utils/issues/60
-    let fileName: string;
-    if (frame.file.startsWith('file://'))
-      fileName = new URL(frame.file).pathname;
```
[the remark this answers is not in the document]
mxschmitt: lets see what the bots say, on my mac it passed. it was fixed in 2019, so its fine, since we only support Node.js 14 officially which was released in 2020.

tests/playwright-test/esm.spec.ts
mxschmitt: Yes, we do have: tests/playwright-test/reporter-base.spec.ts:66

packages/playwright-core/src/utilsBundle.ts
```diff
@@ -28,10 +31,26 @@ export const program: typeof import('../bundles/utils/node_modules/commander').p
...
+
+const StackUtils: typeof import('../bundles/utils/node_modules/@types/stack-utils') = require('./utilsBundleImpl').StackUtils;
+const stackUtils = new StackUtils();
+
+export function parseStackTraceLine(line: string): { frame: import('../bundles/utils/node_modules/@types/stack-utils').StackLineData | null, fileName: string | null } {
+  const frame = stackUtils.parseLine(line);
+  if (!frame)
+    return { frame: null, fileName: null };
+  let fileName = null;
+  if (frame.file)
```
dgozman, line 49: Doesn't this need curly braces?

dgozman (approved):

PR #15269, opened 2022-06-30
feat(matchers): support null/undefined with toHaveProperty

### LINKED ISSUE #14909: [internal]: toHaveJSProperty null -> should accept undefined as well
Motivation: Language ports like Python, .NET, or Java can't test for undefined when using toHaveJSProperty.

As discussed on 15.06 if the user passes null, it should also match undefined in that case.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/14909

Inside the language ports we then use this token when null gets passed over.


mxschmitt:
Waiting for actual user reports running into it.

PR #15286, opened 2022-06-30
test: unflake "should truncate long test names"

dgozman:
Adding `\n` to test output to easier check per-line expectations.


packages/playwright-test/src/reporters/list.ts
```diff
@@ -149,6 +149,8 @@ class ListReporter extends BaseReporter {
     // Go down if needed.
     if (testRow !== this._lastRow)
       process.stdout.write(`\u001B[${this._lastRow - testRow}E`);
+    if (process.env.PWTEST_TTY_WIDTH)
+      process.stdout.write('\n');  // For testing.
```
yury-s, line 153: Add an explanation why it is needed (in pr description would be fine I think).

yury-s (approved):

PR #15289, opened 2022-06-30
chore: port 1.23 Python release notes

rwoll:


docs/src/release-notes-js.md
```diff
@@ -96,12 +96,12 @@ Read more about [component testing with Playwright](./test-components).
...
   });
   ```
-* If you intend to edit HAR by hand, consider using the `"minimal"` HAR recording mode 
+* If you intend to edit HAR by hand, consider using the `"minimal"` HAR recording mode
   that only records information that is essential for replaying:
   ```ts
     recordHar: {
-      path: 'github.har.zip',
+      path: 'github.har',
```
rwoll, line 104: I made this change since this snippet is for `If you intend to edit HAR by hand`, so I think it doesn't make sense to have .zip. Is that correct?

yury-s: Yeah, makes sense

yury-s (approved):
I'd rather we used sync api snippets, this is what we did in previous releases.

PR #15304, opened 2022-07-01
Remove `chromium` switch

IEvangelist:
Removing hardcoded `chromium` CLI switch/option, as it comes from the `${{ matrix.browser }}` expression. Otherwise, at runtime this would evaluate as follows:

- 0: `npx playwright install --with-deps chromium chromium`
- 1: `npx playwright install --with-deps firefox chromium`
- 2: `npx playwright install --with-deps webkit chromium`


mxschmitt:
This is expected, because Chromium is required to run e.g. inspector tests so we always need a Chromium browser installed.

PR #15314, opened 2022-07-01
test: fix some flaky failures

dgozman:
- Never use open shadow root for highlight. This messes up our selectors that accidentally match internal preview elements.
- Remove failing electron test that we do not care about.
- Skip `channels.spec.ts` in non-default mode.

### DIFF
--- a/packages/playwright-core/src/server/injected/highlight.ts
+++ b/packages/playwright-core/src/server/injected/highlight.ts
@@ -45,8 +45,9 @@ export class Highlight {
     this._actionPointElement = document.createElement('x-pw-action-point');
     this._actionPointElement.setAttribute('hidden', 'true');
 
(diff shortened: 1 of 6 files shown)


packages/playwright-core/src/server/injected/highlight.ts
```diff
@@ -45,8 +45,9 @@ export class Highlight {
     this._actionPointElement = document.createElement('x-pw-action-point');
     this._actionPointElement.setAttribute('hidden', 'true');
 
-    // Use a closed shadow root to prevent selectors matching our internal previews.
-    this._glassPaneShadow = this._glassPaneElement.attachShadow({ mode: isUnderTest ? 'open' : 'closed' });
+    // NOTE: do not use an open shadow root, event for test.
```
yury-s, line 48: typo: event -> even

yury-s (approved):

Issue #15086, opened 2022-06-23
[Question] Is the Page `load` event supposed to be triggered multiple times?


rinogo:
In my project, I'm seeing `page.on('load', handleEvent)` generate multiple calls to `handleEvent()` for a single `page.goto()`. Is that the intended behavior? I don't think there are any redirects present, although I suppose that could have something to do with it. 🤷‍♂️

I feel like I'm missing something obvious. This behavior seems to be somewhat consistent depending on the page being loaded. One page only generates one `load` event while another might generate 2, 3, or 4 events (and resulting calls).

Am I missing something obvious? I've confirmed that I'm only registering for the event once, but I still see multiple responses to a page load.

I'm happy to create a new Issue if there's a bug present, but something tells me that there's a Playwright-specific nuance here that's not present in the [brief documentation for `page.on('load')`](https://playwright.dev/docs/api/class-page#page-event-load).

mxschmitt:
It should only fire a single time per navigation, can you provide us a repro which reflects the issue?

rinogo:
Hi, Max! Thanks for your response. This should reproduce on Playwright 1.22.2.

```javascript
const playwright = require('playwright');

(async () => {
	const url = 'https://www.prweek.com/us';
	const browser = await playwright.chromium.launch();
	const context = await browser.newContext();
	const page = await context.newPage();
	page.on('load', (p) => console.log(`Loaded ${p.url()}`));
	console.log(`Loading ${url}`);
	await page.goto(url);
	await page.waitForTimeout(5000); //If this line is commented out, Playwright seems to behave as we'd expect. However, I suspect it's simply because we're exiting before the issue manifests.
	process.exit();
})();
```

Output:
```
rinogo$ node repro.js 
Loading https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
Loaded https://www.prweek.com/us
```

rinogo:
Alternatively, [see the repro on `try.playwright.tech`](https://try.playwright.tech/?l=javascript&s=gf6batj).

rwoll:
[...]

PR #15320, opened 2022-07-01
test: repro for load event fired too often

rwoll:
Repro derived from #15086. This does not look like a recent regression (if any at all) as it reproduces in both 1.22 and 1.23.

### DIFF
--- a/tests/page/page-event-load.spec.ts
+++ b/tests/page/page-event-load.spec.ts
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


tests/page/page-event-load.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+
+it('should fire once', async ({ page, server, browserName }) => {
+  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/15086' });
+  it.fixme(browserName === 'firefox', 'Firefox sometimes double fires.');
+
+  let count = 0;
+  page.on('load', () => count++);
+  await page.goto(server.PREFIX + '/one-style.html');
+  await page.waitForTimeout(1000);
```
rwoll, line 27: What's an alternative here? Asserting it fired at least once is easy, but not sure of a great way to assert **exactly** since we can't see into the future!

tests/page/page-event-load.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    res.end();
+  });
+  let count = 0;
+  page.on('load', () => count++);
+  await page.goto(server.PREFIX + '/home');
+  const trackingFrame = page.frameLocator('[name=tracker]');
+  await expect(trackingFrame.locator(':scope')).toContainText('request count: 1');
+  const loadFired = Promise.race([
+    page.waitForEvent('load').then(() => 'loadfired'),
+    page.waitForTimeout(1000).then(() => 'timeout'),
```
rwoll, line 69: Same comment as above.

dgozman (approved):

PR #15359, opened 2022-07-04
fix(junit reporter): embedding attachments on report didnt work for tests outside root folder

### LINKED ISSUE #15360: [BUG] junit_reporter: embed attachments only works for tests in project root folder
**Context:**
- Playwright Version: 1.23.1
- Operating System: Mac
- Node.js version: v12.18.2
- Browser: All
- Extra: 

## System:
 - OS: macOS 12.4
 - Memory: 1.26 GB / 16.00 GB
## Binaries:
 - Node: 12.18.2 - ~/.nvm/versions/node/v12.18.2/bin/node
 - Yarn: 1.22.10 - /usr/local/bin/yarn
 - npm: 8.3.0 - ~/.nvm/versions/node/v12.18.2/bin/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash

**Code Snippet**

playwright.config.js
```

```

sample test:

```javascript
    test('Login with valid credentials', async({ page }, testInfo) => {
       await page.goto('https://example.com/');

        // Capture a screenshot and attach it.
        const path = testInfo.outputPath('tmp_screenshot.png');
        await page.screenshot({ path });
        testInfo.attachments.push({ name: 'screenshot.png', path, contentType: 'image/png' });
    });

```

Expected Junit XML report (after running `npx playwright test`):
```xml
<testcase name="Login validations Login with valid credentials" classname="login.spec.js:13:5 › Login validations › Login with valid credentials" time="2.226">
<properties>
<property name="testrun_evidence">
<item name="screenshot.png">
<![CDATA[iVBORw0KGgoAAAANSUhEUgAABQAAAALQCAYAAADPfd1WAAAAAXNSR0IArs4c6QAAIABJREFUeJzs3Xv81/PdP/BHB+TwdbhQDjW09FVR5ComYiGHhGrNYWziYphTOeW6nMIcZpfDLlNsl5nM2ZrFZcmQQ5pDKkSlDNloLHwlJfX7o9/3s759Pt/65rB4736/3brt83m9X+/3+/V+vz9u6/bo9Xo/G02fPn1RAAAAAIBCaryyBwAAAAAAfHkEgAAAAABQYAJAAAAAACgwASAAAAAAFJgAEAAAAAAKTAAIAAAAAAUmAAQAAACAAhMAAgAAAECBCQABAAAAoMAEgAAAAABQYAJAAAAAACgwASAAAAAAFJgAEAAAAAAKTAAIAAAAAAUm....
....
```
**Describe the bug**

The junit reporter has the possibility of embedding attachments as base64. However, the previous code checked for the existence of the attachment based on a wrong check and was not reading it also from the correct path.
The previous code worked fine, if the test was in the project's root folder. If the test was somewhere else, like in ".tests/", then it wouldn't work as expected.

bitcoder:
There was a relative path for attachments that could be embed on the junit XML report; it was obtained for logging purposes; however, the actual attachment path is the one to consider (and it's independent from the relative location of the tests).

fixes #15360 by properly checking the attachment path and reading from the correct path

### DIFF
--- a/packages/playwright-test/src/reporters/junit.ts
+++ b/packages/playwright-test/src/reporters/junit.ts
@@ -191,11 +191,10 @@ class JUnitReporter implements Reporter {
             if (!attachment.path)
               continue;
             try {
-              const attachmentPath = path.relative(this.config.rootDir, attachment.path);
               if (fs.existsSync(attachment.path))
                 contents = fs.readFileSync(attachment.path, { encoding: 'base64' });
               else
-                systemErr.push(`\nWarning: attachment ${attachmentPath} is missing`);
+                systemErr.push(`\nWarning: attachment ${attachment.path} is missing`);
             } catch (e) {
             }
           }
--- a/packages/playwright-test/src/reporters/junit.ts
+++ b/packages/playwright-test/src/reporters/junit.ts
@@ -192,8 +192,8 @@ class JUnitReporter implements Reporter {
               continue;
             try {
               const attachmentPath = path.relative(this.config.rootDir, attachment.path);
-              if (fs.existsSync(attachmentPath))
-                contents = fs.readFileSync(attachmentPath, { encoding: 'base64' });
+              if (fs.existsSync(attachment.path))
+                contents = fs.readFileSync(attachment.path, { encoding: 'base64' });
               else
                 systemErr.push(`\nWarning: attachment ${attachmentPath} is missing`);
             } catch (e) {


packages/playwright-test/src/reporters/junit.ts
```diff
@@ -192,8 +192,8 @@ class JUnitReporter implements Reporter {
               continue;
             try {
               const attachmentPath = path.relative(this.config.rootDir, attachment.path);
```
aslushnikov, line 194: I wonder if this was supposed to be `path.resolve` instead. cc @dgozman

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -192,8 +192,8 @@ class JUnitReporter implements Reporter {
...
-              if (fs.existsSync(attachmentPath))
-                contents = fs.readFileSync(attachmentPath, { encoding: 'base64' });
+              if (fs.existsSync(attachment.path))
+                contents = fs.readFileSync(attachment.path, { encoding: 'base64' });
               else
                 systemErr.push(`\nWarning: attachment ${attachmentPath} is missing`);
```
dgozman, line 198: Let's print `attachment.path` here and drop `attachmentPath` variable altogether?

dgozman (commented):
Thank you for the fix, this is indeed an issue.

packages/playwright-test/src/reporters/junit.ts
bitcoder: done

dgozman (approved):

PR #15427, opened 2022-07-06
docs: explain how to sign in once per worker

### LINKED ISSUE #14659: [Question] Is there a way to use different authentication state for different workers
I am running tests against a website using 3 workers

I am reusing the same authentication state to login

so when tests are run in parallel , same user gets logged in at 3 places which slows down the app.

Is there a chance that I can create 3 different authentication states using 3 different users and use each of them on 3 seperate workers so that they can login simultaneously

dgozman:
Fixes #14659.

### DIFF
--- a/docs/src/test-auth-js.md
+++ b/docs/src/test-auth-js.md
@@ -200,7 +200,7 @@ async function globalSetup() {
 export default globalSetup;
 ```
 
-### Avoding multiple sessions per account at a time
+### Avoiding multiple sessions per account at a time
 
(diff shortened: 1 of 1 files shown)


docs/src/test-auth-js.md
```diff
@@ -200,6 +200,87 @@ async function globalSetup() {
 export default globalSetup;
 ```
 
+### Avoding multiple sessions per account at a time
```
rwoll, line 203:
```suggestion
### Avoiding multiple sessions per account at a time
```

rwoll (approved):

PR #15428, opened 2022-07-06
fix(test runner): collect artifacts when calling `browser.close()`, attempt 2

### LINKED ISSUE #14773: [BUG] Trace file is not generated when the last test fails in serial mode
**Context:**
- Playwright Version:  1.22.2
- Node.js version:  17.7.1
- Browser: chromium  102.0.5005.40
- Extra: We handle the creation of the `browser`, `page` and `context` ourselves.  We do not use fixtures. This has been simplified in the snippet below.

```
## System:
 - OS: Windows 10 10.0.22000
 - Memory: 2.93 GB / 15.79 GB
## Binaries:
 - Node: 17.7.1 - C:\Program Files\nodejs\node.EXE
 - Yarn: 1.22.17 - C:\Program Files\nodejs\yarn.CMD
 - npm: 8.5.2 - C:\Program Files\nodejs\npm.CMD
## Languages:
 - Bash: 5.0.17 - C:\Windows\system32\bash.EXE
```

**Code Snippet**

*playwright.config.ts*
```typescript
const config = {
  testDir: 'tests',
  reporter: [
    ['dot'],
    ['html'],
  ],
  timeout: 10000,
  forbidOnly: true,
  use: {
    headless: true,
    viewport: { width: 1920, height: 1080 },
    video: 'retain-on-failure',
    trace: 'retain-on-failure',
  },
};

module.exports = config;
```

*example.spec.ts*
```typescript
import {
  test,
  Page,
  Browser,
  chromium,
  BrowserContext,
} from "@playwright/test";

let page: Page;
let browser: Browser;
let context: BrowserContext;

test.describe.serial("serial mode - trace file does not generate when the last test fails", () => {
  test.beforeAll(async ({}) => {
    browser = await chromium.launch();
    context = await browser.newContext()
    page = await context.newPage()
  });

  test.afterAll(async () => {
    await browser.close();
  });

  test("first test passes", async () => {
    await page.goto("https://playwright.dev/");
  });

  test("second test passes", async () => {
    await page.goto("https://playwright.dev/");
  });

  test("last test fails", async () => {
    await page.click("text=Get Started fail"); // << deliberate fail, when this last test fails - no trace file is generated.
  });
});

```

**Describe the bug**

If the last test fails in a serial mode execution, no trace file is generated.

dgozman:
Drive-by: call `onDidCreateContext` for contexts created with `connectOverCDP`. This was exposed by tracing bot.

Fixes #14773.

### DIFF
--- a/packages/playwright-core/src/client/browser.ts
+++ b/packages/playwright-core/src/client/browser.ts
@@ -104,6 +104,8 @@ export class Browser extends ChannelOwner<channels.BrowserChannel> implements ap
 
   async close(): Promise<void> {
     try {
+      for (const context of this.contexts())
+        await this._browserType?._onWillCloseContext?.(context);
       if (this._shouldCloseConnectionOnClose)
         this._connection.close(kBrowserClosedError);
       else
--- a/packages/playwright-core/src/client/browserType.ts
+++ b/packages/playwright-core/src/client/browserType.ts
@@ -211,6 +211,8 @@ export class BrowserType extends ChannelOwner<channels.BrowserTypeChannel> imple
       browser._contexts.add(BrowserContext.from(result.defaultContext));
     browser._logger = params.logger;
     browser._setBrowserType(this);
+    for (const context of browser.contexts())
+      await this._onDidCreateContext?.(context);
     return browser;
   }
 }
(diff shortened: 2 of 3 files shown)


yury-s (approved):
Let's discuss this in the meeting tomorrow, I'm not sure we want to encourage people rely on this behavior while they could handle that in a fixture and given that playwright library also doesn't do it.

DavertMik:
Fixes #15430

PR #15430, opened 2022-07-06
chore: respect event name in removeAllListeners

### LINKED ISSUE #15431: [BUG] Regression: page.on('console') is ignored in 1.23
**Context:**
- Playwright Version: 1.23
- Operating System: Linux
- Node.js version: 14 (checked on 12 too)
- Browser: Chromium
- Extra: CodeceptJS

If I use `page.removeAllListeners('dialog');` it removes listener to `page.on('console')`.
Check the example:

```

(async () => {
  // Create pages, interact with UI elements, assert values

  page.on('console', async msg => {
        console.log(msg)
  });
  // comment out next and see that console log message is printed
  page.removeAllListeners('dialog');
  await page.goto('http://127.0.0.1:8000/');
  await page.evaluate(() => console.log('aaaaaa'));
  await browser.close();
})();

```
In Playwright 1.22 it passes correctly but started failing since 1.23

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/15431

### DIFF
--- a/packages/playwright-core/src/client/joiningEventEmitter.ts
+++ b/packages/playwright-core/src/client/joiningEventEmitter.ts
@@ -54,7 +54,7 @@ export class JoiningEventEmitter implements EventEmitter {
   }
 
   removeAllListeners(event?: string | symbol | undefined): this {
-    this._emitterDelegate.removeAllListeners();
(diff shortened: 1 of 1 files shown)


DavertMik:
Fixes: #15431

PR #15476, opened 2022-07-08
fix(test runner): more friendly test duration

### LINKED ISSUE #15338: [BUG] Worker-scope fixture setup skewing test duration
Hi! 👋 

**Context:**
- Playwright Version: 1.22.2
- Operating System: macOS 12.1 Monterey
- Node.js version: v18.4.0
- Browser: Chrome
- Extra: I doubt the environment plays a role

**Describe the bug**
The duration of any worker-scoped fixture setup is attributed to the first test that the worker runs afterwards.
i.e. if there's a worker-scoped fixture that takes 20s to setup, and the worker runs 3 tests that each take 5 seconds, the report will say that:
1. test 1 (25s) -> so: counting fixture setup.
2. test 2 (5s)
3. test 3 (5s)

This:
a) greatly skews the metrics for tests duration
b) is non-deterministic, as the fixture setup is not done consistently previous to the same test across runs (if e.g. there's a test failure and the worker restarts)

IMO, duration of worker-scoped fixtures setup should be reported (still, but) completely outside of individual test duration, that'd be my user expectation.

dgozman:
Reported test duration now does not include time spent in `beforeAll`, `afterAll` and fixtures that have a separate timeout.
This is to avoid different reported test execution time, depending on the test execution order.

Fixes #15338.

### DIFF
--- a/tests/playwright-test/timeout.spec.ts
+++ b/tests/playwright-test/timeout.spec.ts
@@ -340,3 +340,41 @@ test('test timeout should still run hooks before fixtures teardown', async ({ ru
     '%%after-auto',
   ]);
 });
+
+test('should not include fixtures with own timeout and beforeAll in test duration', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'c.spec.ts': `
+      const test = pwt.test.extend({
+        foo: [async ({}, use) => {
+          await new Promise(f => setTimeout(f, 1000));
+          await use('foo');
+        }, { timeout: 0 }],
+
+        bar: async ({}, use) => {
+          await new Promise(f => setTimeout(f, 300));
+          await use('bar');
+        },
+      });
+
+      test.beforeAll(async () => {
+        await new Promise(f => setTimeout(f, 1000));
+      });
+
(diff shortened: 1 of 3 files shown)


tests/playwright-test/timeout.spec.ts
```diff
@@ -340,3 +340,41 @@ test('test timeout should still run hooks before fixtures teardown', async ({ ru
...
+
+      test('works', async ({ foo, bar }) => {
+        await new Promise(f => setTimeout(f, 300));
+      });
+    `
+  }, { timeout: 5000 });
+  expect(result.exitCode).toBe(0);
+  const duration = result.results[0].duration;
+  expect(duration).toBeGreaterThanOrEqual(300 * 4);  // Includes test, beforeEach, afterEach and bar.
+  expect(duration).toBeLessThan(300 * 4 + 1000);  // Does not include beforeAll and foo.
```
yury-s, line 379: I got a feeling this it will be flaky depending on how loaded CI bots are.

yury-s (approved):

PR #15482, opened 2022-07-08
tests(chromium): service worker main request/response should be marked correctly

Meir017:
This can be seen in the devtools (navigate to https://meir017.github.io/service-worker-examples/ and open the devtools)
image

* The main service-worker request should be marked as sent from service-worker
* The response should not be served from a service worker

### DIFF
--- a/tests/library/chromium/chromium.spec.ts
+++ b/tests/library/chromium/chromium.spec.ts
@@ -101,6 +101,8 @@ test('should intercept service worker requests (main and within)', async ({ cont
 
   const [ sw ] = await Promise.all([
     context.waitForEvent('serviceworker'),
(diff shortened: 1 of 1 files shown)


rwoll (approved):
Thanks! LGTM

Meir017:
BTW, @rwoll is the correct behavior? Now playwright is consistent with the devtools but the main sw request is sent before the sw is evaluated

rwoll:
> BTW, @rwoll is the correct behavior? Now playwright is consistent with the devtools but the main sw request is sent before the sw is evaluated

Yes, it's now documented :) You also get the SW event before the main script is even requested (by design), so the fact it's not fully evaluated is OK.

PR #15496, opened 2022-07-08
devops: support universal webkit builds

aslushnikov:


browser_patches/webkit/archive.sh
```diff
@@ -28,6 +28,11 @@ if ! [[ -d $(dirname "$ZIP_PATH") ]]; then
   exit 1
 fi
 
+IS_UNIVERSAL=""
+if [[ $2 == "--universal" ]]; then
```
yury-s, line 32: The code below assumes that this flag can be combined with others hence doesn't not necessarily come as second argument. Let's move argument parsing into a single location?

yury-s (approved):

PR #15503, opened 2022-07-08
fix(webkit): use new protocol methods for bindings

### LINKED ISSUE #14745: [BUG] exposeBinding() breaks in Webkit on a certain page
**Context:**
- Playwright Version: 1.22.2
- Operating System: Mac
- Node.js version: v17.8.0
- Browser: WebKit
- Extra: nothing special

**Code Snippet**

```typescript
import { test, expect, Page, Frame } from "@playwright/test";

test.describe('xxx', async () => {
  test(`test exposeBinding`, async ({ page }) => {
    const url = "https://www.dastelefonbuch.de/"; // fails with this url
    // const url = "https://example.com/"; // works with this url
    await page.exposeBinding("exposedFn", messageCallback);
    await page.goto(url, { waitUntil: "commit" });

    let received = null;

    async function messageCallback({ frame }: { frame: Frame }, msg: any) {
      console.log("RECEIVED MESSAGE", msg);
      received = msg;
    }

    await page.waitForTimeout(1000); // seems to be important. I suspect the page is messing up the global context upon initialization

    console.log('trying to call the exposed function...');
    await page.evaluate(
      'exposedFn({ hello: "there" })'
    ); // hangs here
    console.log('...done');

    expect(received).toEqual({ hello: "there" });
  });
});

```

**Describe the bug**

The callback exposed with `Page.exposeBinding()` stops working on a specific page: https://www.dastelefonbuch.de/. It does seem to exist in page context, but the payload is never delivered to playwright scope.
It works when I change the URL or the browser, so it's really specific to WebKit and this particular website.

I run the above code with `npx playwright test doit.spec.ts --browser webkit`, so there's no extra config.

yury-s:
Fixes #14745

### DIFF
--- a/packages/playwright-core/src/server/webkit/wkPage.ts
+++ b/packages/playwright-core/src/server/webkit/wkPage.ts
@@ -46,7 +46,6 @@ import { debugLogger } from '../../common/debugLogger';
 import { ManualPromise } from '../../utils/manualPromise';
 
 const UTILITY_WORLD_NAME = '__playwright_utility_world__';
-const BINDING_CALL_MESSAGE = '__playwright_binding_call__';
 
 export class WKPage implements PageDelegate {
   readonly rawMouse: RawMouseImpl;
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/webkit/wkPage.ts
```diff
@@ -522,6 +524,15 @@ export class WKPage implements PageDelegate {
     this._contextIdToContext.set(contextPayload.id, context);
   }
 
+  private async _onBindingCalled(contextId: any, name: any, argument: any) {
```
dgozman, line 527: Change `any` to actual types?

tests/page/page-expose-function.spec.ts
```diff
@@ -300,3 +300,13 @@ it('should serialize cycles', async ({ page }) => {
...
   expect(object).toEqual(a);
 });
+
+it('should work with overridden console object', async ({ page }) => {
+  await page.evaluate(() => window.console = null);
+  expect(page.evaluate(() => window.console === null)).toBeTruthy();
+  await page.exposeFunction('__pw_add', function(a, b) {
+    return Promise.resolve(a + b);
+  });
+  await (page as any)._removeExposedBindings();
```
dgozman, line 310: Let's drop this line.

dgozman (approved):

Issue #15474, opened 2022-07-07
[BUG] Page Request Served from SW Cache: Missing Page-Request Event


rwoll:
See https://gist.github.com/rwoll/100f361aca78838aad3d138eac30f35a#file-example-spec-ts-L90.

If a page makes a request, even if the SW fetch handler responds from https://developer.mozilla.org/en-US/docs/Web/API/Cache, we should still see the request on the page itself.

rwoll:
Logs for the Network event(s) triggered by this line: https://github.com/microsoft/playwright/pull/15544/files#diff-f48a7787820cb49396256c1234745152eb05ca5a3f43b6d128ce0ee45eb6dff6R335

```
pw:protocol ◀ RECV {"method":"Network.requestWillBeSent","params":{"requestId":"1314.4","loaderId":"0116D4046F08D37E8CF343E9294D70A1","documentURL":"http://localhost:8907/test-page.html","request":{"url":"http://localhost:8907/scripts/resource-1.js","method":"GET","headers":{"sec-ch-ua":"","Referer":"http://localhost:8907/test-page.html","Accept-Language":"en-US","sec-ch-ua-mobile":"?0","User-Agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/104.0.5112.39 Safari/537.36","sec-ch-ua-platform":""},"mixedContentType":"none","initialPriority":"Low","referrerPolicy":"strict-origin-when-cross-origin","isSameSite":true},"timestamp":963717.502395,"wallTime":1657573789.014687,"initiator":{"type":"script","stack":{"callFrames":[{"functionName":"eval","scriptId":"33","url":"","lineNumber":5,"columnNumber":22},{"functionName":"eval","scriptId":"33","url":"","lineNumber":1,"columnNumber":13},{"functionName":"evaluate","scriptId":"6","url":"","lineNumber":192,"columnNumber":16},{"functionName":"","scriptId":"13","url":"","lineNumber":0,"columnNumber":43}]}},"redirectHasExtraInfo":false,"type":"Script","frameId":"DD4EFEDD63D262067A2BC752F1C30CC4","hasUserGesture":true},"sessionId":"6F9602EE99F11C04FEAADEF7D420B018"}
[...]

PR #15544, opened 2022-07-11
test: page-level request/response events with SW fetch handlers and interception

rwoll:
Repro for #15474 derived from use-case by @Meir017!


tests/library/chromium/chromium.spec.ts
```diff
@@ -250,6 +250,99 @@ test('setOffline', async ({ context, page, server }) => {
...
+  ]);
+
+  await Promise.all([
+    context.waitForEvent('request', request => !!request.serviceWorker()),
+    sw.evaluate(async () => {
+      const cache = await caches.open('v1');
+      await cache.add('scripts/resource-1.js');
+    })
+  ]);
+  console.info('wait for request from window intercepted by service-worker');
```
dgozman, line 329: Drop this one.

tests/library/chromium/chromium.spec.ts
```diff
@@ -250,6 +250,99 @@ test('setOffline', async ({ context, page, server }) => {
...
+  });
+  const [ sw ] = await Promise.all([
+    context.waitForEvent('serviceworker'),
+    page.goto(server.PREFIX + '/test-page.html'),
+    page.waitForFunction('activationPromise'),
+  ]);
+
+  await Promise.all([
+    context.waitForEvent('request', request => !!request.serviceWorker()),
+    sw.evaluate(async () => {
```
dgozman, line 324: Perhaps, just reloading the page would trigger the cache response?

dgozman (approved):

tests/library/chromium/chromium.spec.ts
```diff
@@ -250,6 +250,99 @@ test('setOffline', async ({ context, page, server }) => {
...
+      const cache = await caches.open('v1');
+      await cache.add('scripts/resource-1.js');
+    })
+  ]);
+  console.info('wait for request from window intercepted by service-worker');
+  await Promise.all([
+    // currently times out here
+    page.waitForRequest('**/scripts/resource-1.js'),
```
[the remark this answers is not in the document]
rwoll: I marked with `fixme` so we don't actually run this.

tests/library/chromium/chromium.spec.ts
rwoll: Good point—let me see if it repros like that instead.

rwoll: Oh—we don't load this script on page-load, so this is the first time we're requesting it.

PR #15549, opened 2022-07-12
fix: Service Workers+Interception: missing page-level Network events

### LINKED ISSUE #15474: [BUG] Page Request Served from SW Cache: Missing Page-Request Event

rwoll:
Fixes #15474.

Notes:

* page-level requests that are also handled by a SW's fetch handler, should not be interceptable at the page-level
* `Network.requestWillBeSent` does not provide enough metadata for Playwright to fire the `request` event at that time, so it does it as soon as it gets to the end of the request lifecycle

### DIFF
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -367,7 +367,7 @@ export class CRNetworkManager {
     // For frame-level Requests that are handled by a Service Worker's fetch handler, we'll never get a requestPaused event, so we need to
     // manually create the request. In an ideal world, crNetworkManager would be able to know this on Network.requestWillBeSent, but there
     // is not enough metadata there.
-    if (this._protocolRequestInterceptionEnabled && event.response.fromServiceWorker) {
+    if (!request && event.response.fromServiceWorker) {
       const requestWillBeSentEvent = this._requestIdToRequestWillBeSentEvent.get(event.requestId);
       const frame = requestWillBeSentEvent?.frameId ? this._page?._frameManager.frame(requestWillBeSentEvent.frameId) : null;
       if (requestWillBeSentEvent && frame) {
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -363,7 +363,18 @@ export class CRNetworkManager {
   }
 
   _onResponseReceived(event: Protocol.Network.responseReceivedPayload) {
-    const request = this._requestIdToRequest.get(event.requestId);
+    let request = this._requestIdToRequest.get(event.requestId);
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -363,7 +363,18 @@ export class CRNetworkManager {
   }
 
   _onResponseReceived(event: Protocol.Network.responseReceivedPayload) {
-    const request = this._requestIdToRequest.get(event.requestId);
+    let request = this._requestIdToRequest.get(event.requestId);
+    // For frame-level Requests that are handled by a Service Worker's fetch handler, we'll never get a requestPaused event, so we need to
+    // manually create the request. In an ideal world, crNetworkManager would be able to know this on Network.requestWillBeSent, but there
+    // is not enough metadata there.
+    if (this._protocolRequestInterceptionEnabled && event.response.fromServiceWorker) {
```
dgozman, line 370:
```suggestion
    if (!request && event.response.fromServiceWorker) {
```

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -363,7 +363,18 @@ export class CRNetworkManager {
...
+      const requestWillBeSentEvent = this._requestIdToRequestWillBeSentEvent.get(event.requestId);
+      const frame = requestWillBeSentEvent?.frameId ? this._page?._frameManager.frame(requestWillBeSentEvent.frameId) : null;
+      if (requestWillBeSentEvent && frame) {
+        this._onRequest(frame, requestWillBeSentEvent, null /* requestPausedPayload */);
+        request = this._requestIdToRequest.get(event.requestId);
```
dgozman, line 375:
```suggestion
        request = this._requestIdToRequest.get(event.requestId);
        this._requestIdToRequestWillBeSentEvent.delete(event.requestId);
```

tests/library/chromium/chromium.spec.ts
```diff
@@ -390,6 +393,7 @@ test('should emit page-level request event for respondWith when interception ena
   expect(pageReq.url()).toMatch(/fetchdummy\/foo$/);
   expect(pageReq.serviceWorker()).toBe(null);
   expect((await pageReq.response()).fromServiceWorker()).toBe(true);
+  expect(markFailureIfPageRoutesARequestAlreadyHandledByServiceWorker).toBe(false);
```
dgozman, line 396: Let's also test that we get `response` and `requestfinished` events.

rwoll: updating. also adding requestfailed case, too.

dgozman (approved):

PR #15552, opened 2022-07-12
fix(test runner): make sure worker cleans up if last test was skipped

### LINKED ISSUE #15344: [BUG] `test.AfterAll` block is ignored if last test of the describe has attribute `.skip`
**Context:**
- Playwright Version: 1.22.2
- Operating System: win 10
- Node.js version: 16.6.2
- Browser: Chromium

**Code Snippet**

```javascript

(async () => {
	const browser = await chromium.launch()
	const page = await context.newPage()

	await page.setContent(`
	<div>
	  …
	</div>
	`)

	test.describe('First describe', () => {
		test.beforeAll(async () => {
			console.log('Before test')
		})
		test.afterAll(async () => {
			console.log('After test') // won't be executed
		})

		test.describe('Second describe', () => {
			test('First test', async () => {
				console.log('First test')
				expect(1).toEqual(1)
			})
			test.skip('Second test', async () => {
				console.log('Second test')
				expect(1).toEqual(1)
			})
		})
	})
})()
```

**Describe the bug**

If last test of the describe has attribute `.skip` then `.AfterAll` block won't be executed. If you move test with `.skip` up and last test won't have this attribute then `.AfterAll` block will work.

dgozman:
Fixes #15344.

### DIFF
--- a/packages/playwright-test/src/workerRunner.ts
+++ b/packages/playwright-test/src/workerRunner.ts
@@ -286,7 +286,9 @@ export class WorkerRunner extends EventEmitter {
     setCurrentTestInfo(testInfo);
     this.emit('testBegin', buildTestBeginPayload(testInfo));
 
-    if (testInfo.expectedStatus === 'skipped') {
+    const isSkipped = testInfo.expectedStatus === 'skipped';
+    if (isSkipped && nextTest) {
+      // Fast path - this test and skipped, and there are more tests that will handle cleanup.
       testInfo.status = 'skipped';
       this.emit('testEnd', buildTestEndPayload(testInfo));
       return;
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/workerRunner.ts
```diff
@@ -286,7 +286,9 @@ export class WorkerRunner extends EventEmitter {
     setCurrentTestInfo(testInfo);
     this.emit('testBegin', buildTestBeginPayload(testInfo));
 
-    if (testInfo.expectedStatus === 'skipped') {
+    const isSkipped = testInfo.expectedStatus === 'skipped';
+    if (isSkipped && nextTest) {
+      // Fast path - this test and skipped, and there are more tests that will handle cleanup.
```
yury-s, line 291: typo: this test is skipped

yury-s (approved):

PR #15565, opened 2022-07-12
devops: do not dedupe webkit archives

aslushnikov:
The 2e331715ff7b304529a7ce65a98184681eceaab9 introduced
universal webkit build and tried re-using archive deduping technique
across all our regular JHBuild-based builds.

However, this turns out to be too aggressive and doesn't work. At least
on Ubuntu 18.04:
- `minibrowser-gtk/minibrowser` is identical to
  `minibrowser-wpe/minibrowser`
- WPE gets symlinked into GTK
- Thus we now always start headed


aslushnikov:
TBR: @yury-s

PR #15595, opened 2022-07-12
docs: Service Worker examples and explanations

### LINKED ISSUE #15582: [docs] service workers and network interception

rwoll:
Resolves #15582.


rwoll:
@dgozman Can you take a skim through and let me know your thoughts? If we like the section breakdown and examples, I'll port to the other languages. Thanks!

docs/src/service-worker-network-interception.md
```diff
@@ -0,0 +1,101 @@
...
+});
+```
+
+[`event: BrowserContext.request`] would be emitted each time `fetch("/example.txt")`, `fetch("/a.txt")` and `fetch("/b.txt")` are called.
+
+A [Route] can be setup to intercept these requests via [`method: BrowserContext.route`], too.
+
+If a `fetch` call is made in a page where a Service Worker is registered with a [`FetchEvent` handler](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent#examples), a [`event: Page.request`] ***and*** [`event: BrowserContext.request`] will be emitted for the request associated with the page, in addition to the two [`event: BrowserContext.request`] events that are emitted on behalf of the `fetch("/a.txt")` and `fetch("/b.txt")`. 
+
+The request coming from the page will not be interceptable or routeable at the page- or context-level.
```
dgozman, line 40: What's the difference between interceptable and routeable? I think we should use "route" everywhere.

docs/src/service-worker-network-interception.md
```diff
@@ -0,0 +1,101 @@
...
+Most Service Worker's will just intercept a Page's `fetch` events, and then make the `fetch` themselves:
+
+```js
+// service-worker.js
+self.addEventListener("fetch", async event => {
+  event.respondWith(fetch(event.request));
+});
+```
+
+In this case, calling `fetch('/data.json')` in the page, would trigger emit context-level events for the request within the Service Worker, plus a page- and context-level set of events.
```
dgozman, line 53: - typo: 'trigger emit'
- This section isn't really clear. I'd suggest a list of events, something similar to this: https://playwright.dev/docs/test-retries#failures

docs/src/service-worker-network-interception.md
```diff
@@ -0,0 +1,101 @@
...
+      status: 200,
+      body: 'from sw',
+    });
+  } else {
+    return route.continue();
+  }
+});
+```
+
+## Intercepting the Main Service Worker Script
```
dgozman, line 80: I'd omit this section just to avoid extra complexity. Previous section covers it already.

docs/src/service-worker-network-interception.md
```diff
@@ -0,0 +1,101 @@
...
+
+The initial request for the Service Worker's script can be intercepted via [`method: BrowserContext.route`].
+
+## Disabling Service Workers
```
dgozman, line 84: - Let's rename this doc to "Service Workers Guide"
- Start with this option explained, tell when you would want to use it (most of the time?). Explain the caveat that routing all network requires this option in non-Chromium.
- Then tell something about a chromium-only ServiceWorker object we expose, and waiting for activation (section below), and evaluating in SW context. An example with examining cache entries seems appropriate to me.
- Then explain network events and routing.

docs/src/service-worker-network-interception.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Waiting for Activation
+
+If you want want to ensure a Service Worker is activated before continuing some actions, you can use:
+
+```js
+await page.waitForFunction('navigator.serviceWorker.ready');
+```
+
+The [`BrowserContext.serviceworker`] event is emitted ***before*** the Service Worker main script has been evaluated, so before calling [ServiceWorker.evaluate] you should wait on its activation.
```
dgozman, line 96: It seems like waiting in the `page` before being able to use `ServiceWorker` object is kinda wrong. Can we instead evaluate in the service worker itself? Or add a method for it?

docs/src/service-worker-network-interception.md
```diff
@@ -0,0 +1,101 @@
...
+title: "Service Worker Network Interception (Chromium)"
+---
+
+:::note
+Service Worker Network instrumentation only works in Chrome / Chromium. If you want full network interception in Firefox and WebKit, you can disable Service Workers by setting [`option: Browser.newContext.serviceWorkers`] to `'block'`.
+:::
+
+If a site uses [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API), Playwright's standard set of Network Events and Interception/Routing will work. Service Workers have complex lifecycles and can intercept Network Requests themselves, so there are a few things to note.
+
+## Event Targets
```
dgozman, line 12: This section explains how service worker fetch works. Let's name it accordingly?

docs/src/service-worker-network-interception.md
rwoll: Yes, updated. A method/event would be nice too.

tests/library/chromium/chromium.spec.ts
```diff
@@ -233,6 +233,68 @@ test('should intercept only serviceworker request, not page', async ({ context,
...
+    '[SW   ] fallthrough.txt           fromServiceWorker: false',
+  ]);
+  expect(await Promise.all(pageRouted.map(formatRequest))).toEqual([
+    '[FRAME] index.html                fromServiceWorker: false',
+    '[FRAME] addressbook.json          fromServiceWorker: false',
+    '[FRAME] example.jpg               fromServiceWorker: false',
+    '[FRAME] tracker.js                fromServiceWorker: false',
+    '[FRAME] fallthrough.txt           fromServiceWorker: false',
+  ]);
+});
```
rwoll, line 296: So `fromServiceWorker` seems to (1) be wrong here, (2) disagree with the docs, (3) disagree with out other tests.

I'm a bit confused though, since we have other cases of it working:

https://github.com/microsoft/playwright/blob/3436e64b755776108907c6e365dac86dffd1f6be/tests/library/chromium/chromium.spec.ts#L378

https://github.com/microsoft/playwright/blob/3436e64b755776108907c6e365dac86dffd1f6be/tests/library/chromium/chromium.spec.ts#L421

Raw CDP logs for this test look like it's always `false`.

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,159 @@
...
+* [`method: Request.serviceWorker`] will be set to the [Service Worker] instance, and [`method: Request.frame`] will **throw**
+* [`method: Response.fromServiceWorker`] will return `false`
+
+A Network Request made within a page will have:
+
+* [`event: BrowserContext.request`] event emitted
+* [`event: Page.request`] event emitted
+* [`method: BrowserContext.route`] and [`method: Page.route`] will **not** see the request (if a Service Worker's fetch handler was registered)
+* [`method: Request.serviceWorker`] will be set to `null`, and [`method: Request.frame`] will return the [Frame]
+* [`method: Response.fromServiceWorker`] will return `true` (if a Service Worker's fetch handler was registered)
```
rwoll, line 54: Hopefully this (coupled with the tables below) make it clearer and more closely align with the link you shared.

Given enough context, the test case I added also seems elucidating, but I don't think it belongs in the docs.

rwoll (commented):
@dgozman Took another pass at it.

I haven't yet edited it for spelling/proper rendering, and tomorrow I'll need to debug the new test case.

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,159 @@
+---
+id: service-workers-guide
+title: "Service Workers Guide"
+---
+
+## Disabling Service Workers
+
+[Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) provide a browser-native way to handle requests made with the native [Fetch API (`fetch`)](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
+
+If you're using Playwright's [Network-related features](./network.md) like the routing APIs ([`method: BrowserContext.route`], [`method: BrowserContext.routeFromHAR`], [`method: Page.route`], [`method: Page.routeFromHAR`]), or rely heavily on Network Events (e.g. [`event: Page.request`], [`event: BrowserContext.request`], etc.), we recommend disabling Service Workers by setting [`option: Browser.newContext.serviceWorkers`] to `'block'`.
```
rwoll, line 10: I'm not sure I agree with this recommendation. The qualification helps, but there is a point to be made this potentially changes the behavior of the app under test s.t. it no longer reflects what a user would experience.

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+| Context | Frame          | addressbook.json          |        | Yes               |
+| Page    | Frame          | addressbook.json          |        | Yes               |
+| Context | Service Worker | bar                       | Yes    |                   |
+| Context | Frame          | foo                       |        | Yes               |
+| Page    | Frame          | foo                       |        | Yes               |
+| Context | Frame          | tracker.js                |        | Yes               |
+| Page    | Frame          | tracker.js                |        | Yes               |
+| Context | Service Worker | fallthrough.txt           | Yes    |                   |
+| Context | Frame          | fallthrough.txt           |        | Yes               |
+| Page    | Frame          | fallthrough.txt           |        | Yes               |
```
rwoll, line 140: I think this table makes it much clearer (and it's generate from a Test nearly verbatim):

Screen Shot 2022-07-13 at 2 12 17 AM

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+You can use [`method: BrowserContext.serviceWorkers`] to list the Service Workers, or specifically watch for it:
+
+```js
+const [ serviceWorker ] = await Promise.all([
+  context.waitForEvent('serviceworker'),
+  page.goto('example-with-a-service-worker.html'),
+]);
+```
+
+The [`BrowserContext.serviceworker`] event is emitted ***before*** the Service Worker main script has been evaluated, so before calling [ServiceWorker.evaluate] you should wait on its activation:
```
rwoll, line 33:
```suggestion
The [`event: BrowserContext.serviceWorker`] event is emitted ***before*** the Service Worker main script has been evaluated, so before calling [`method: ServiceWorker.evaluate`] you should wait on its activation:
```

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+* [`event: Page.request`] event emitted
+* [`method: BrowserContext.route`] and [`method: Page.route`] will **not** see the request (if a Service Worker's fetch handler was registered)
+* [`method: Request.serviceWorker`] will be set to `null`, and [`method: Request.frame`] will return the [Frame]
+* [`method: Response.fromServiceWorker`] will return `true` (if a Service Worker's fetch handler was registered)
+
+
+Consider the following Service Worker Main Script:
+
+```js
+// filename: service-worker-main.js
```
Meir017, line 60: this is duplicated in `service-worker-main.js`
can you use here `<!-- filename: service-worker-main.js -->` ?

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Disabling Service Workers
+
+[Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) provide a browser-native way to handle requests made with the native [Fetch API (`fetch`)](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
+
+If you're using Playwright's [Network-related features](./network.md) like the routing APIs ([`method: BrowserContext.route`], [`method: BrowserContext.routeFromHAR`], [`method: Page.route`], [`method: Page.routeFromHAR`]), or rely heavily on Network Events (e.g. [`event: Page.request`], [`event: BrowserContext.request`], etc.), we recommend disabling Service Workers by setting [`option: Browser.newContext.serviceWorkers`] to `'block'`.
+
+Unless you're testing the Service Worker behavior itself, blocking ensures all requests are routed to your code via Playwright. It also simplifies Network-related APIs within Playwright.
+
+If you need to visibility or routing into all Network events in Firfox or WebKit, currently `'block'`ing Service Workers is required.
```
rwoll, line 14:
```suggestion
If you need visibility or routing into all Network events in Firfox or WebKit, currently `'block'`ing Service Workers is required.
```

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Service Worker Fetch
+
+:::note
+The next sections are only currently supported when using Playwright with Chrome/Chromium.
+:::
+
+### Accessing Service Workers and Waiting for Activation
+
+You can use [`method: BrowserContext.serviceWorkers`] to list the Service Workers, or specifically watch for it:
```
rwoll, line 24:
```suggestion
You can use [`method: BrowserContext.serviceWorkers`] to list the Service Workers, or specifically watch for the creation of a Service Worker if you expect one to be created:
```

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+const [ serviceWorker ] = await Promise.all([
+  context.waitForEvent('serviceworker'),
+  page.goto('example-with-a-service-worker.html'),
+]);
+```
+
+The [`BrowserContext.serviceworker`] event is emitted ***before*** the Service Worker main script has been evaluated, so before calling [ServiceWorker.evaluate] you should wait on its activation:
+
+```js
+await expect.poll(() => serviceWorker.evaluate(() => self.registration.active?.state)).toBe('activated');
```
rwoll, line 36: We probably ought to add an event/PW API for this: something like `serviceWorker.waitForLifecycle('active')`, and/or `serviceWorker.waitForActive()`.

Meir017: can this be achieved using something like this?
```ts
expect.poll(() => page.evaulaute(() => navigator.serviceWorker.controller));
```

docs/src/service-workers-guide.md
dgozman: I think the paragraph about FF/WK is the most important one. If you use networking APIs and test in these browsers, we strongly recommend to disable SW.

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+## Network Events and Routing
+
+A Network Request made within a Service Worker (along with the main script request itself) will have:
```
dgozman, line 41:
```suggestion
Any network request made by the Service Worker will have:
```

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+* [`event: BrowserContext.request`] event emitted
+* [`method: BrowserContext.route`] will see the request
+* [`method: Request.serviceWorker`] will be set to the [Service Worker] instance, and [`method: Request.frame`] will **throw**
+* [`method: Response.fromServiceWorker`] will return `false`
+
+A Network Request made within a page will have:
```
dgozman, line 48: I'd start this section with something like this:

In a typical service worker that has a fetch handler and responds with a fetch (an example source you had would be handy), each page request generates an additional request issued by the Service Worker.

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
...
+<script>
+  window.registrationPromise = navigator.serviceWorker.register('./service-worker-main.js');
+</script>
+```
+
+On the first visit to the page via [`method: Page.goto`], the following Request/Response events would be emitted:
+
+| Scope   | Target         | URL                       | Routed | fromServiceWorker |
+| -       | -              | -                         | -      | -                 |
+| Context | Frame          | index.html                | Yes    |                   |
```
dgozman, line 112: "Scope" is a bit unclear, can we do it like `page.on('request')` vs `context.on('request')`?

tests/library/chromium/chromium.spec.ts
```diff
@@ -233,6 +233,55 @@ test('should intercept only serviceworker request, not page', async ({ context,
   expect(response).toBe('from sw');
 });
 
+test('doc claims should be accurate', async ({ page, context, server }) => {
```
dgozman, line 236: nit: docs will change, and test will not :)

docs/src/service-workers-guide.md
```diff
@@ -0,0 +1,161 @@
+---
+id: service-workers-guide
+title: "Service Workers Guide"
+---
+
```
dgozman, line 5: Suggesting the following intro (rough sketch):

[Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) provide a browser-native way to handle requests made with the native [Fetch API (`fetch`)](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).

Depending on your testing needs, here are the scenarios supported by Playwright:
1. Test your page with a service worker. This works out of the box in all supported browsers.
2. Test your page without a service worker. Use `serviceWorkers: 'block'` option.
3. Route network traffic, whether it comes from a service worker or not. In Firefox and WebKit, use `serviceWorkers: 'block'` option to avoid service worker network traffic entirely. In Chromium, either block service workers or use [`BrowserContext.route()`].
4. Test your service worker implementation. Only in Chromium, use [`BrowserContext.serviceWorkers()`] to get access to the service worker and evaluate there.

docs/src/service-workers-guide.md
rwoll: It's a JS example, so want to keep it as JS syntax. The test code is actually independent from this.

docs/src/api/class-browsercontext.md
```diff
@@ -1102,7 +1102,6 @@ A glob pattern, regular expression or predicate to match the request URL. Only r
 
 ## method: BrowserContext.serviceWorkers
 * since: v1.11
-* langs: js, python
```
rwoll, line 1105: This will be dropped from the changeset once https://github.com/microsoft/playwright/pull/15635 lands

docs/src/service-workers.md
```diff
@@ -0,0 +1,349 @@
...
+Many sites that use Service Workers simply use them as a transparent optimization technique. While users might notice a faster experience, the app's implementation is unaware of their existence. Running the app with or without Service Workers enabled appears functionally equivalent.
+
+**If your app uses Service Workers**, here's the scenarios that Playwright supports:
+
+1. Testing the page exactly like a user would experience it. This works out of the box in all supported browsers.
+1. Test your page without a Service Worker. Set [`option: Browser.newContext.serviceWorkers`] to `'block'`. You can test your page as if no Service Worker was registered.
+1. Listen for and route network traffic via Playwright, whether it comes from a Service Worker or not. In Firefox and WebKit, set [`option: Browser.newContext.serviceWorkers`] to `'block'` to avoid Service Worker network traffic entirely. In Chromium, either block Service Workers or use [`method: BrowserContext.route`].
+1. (Chromium-only) Test your Service Worker implementation itself. Use [`method: BrowserContext.serviceWorkers`] to get access to the Service Worker and evaluate there.
+
+## Disabling Service Workers
```
rwoll, line 20: This repeats the info above; however, it's nice to have a linkable to send to people when all they are after is this snippet.

docs/src/service-workers.md
```diff
@@ -0,0 +1,349 @@
...
+
+[`event: BrowserContext.serviceWorker`] is fired ***before*** the Service Worker's main script has been evaluated, so ***before*** calling service[`method: Worker.evaluate`] you should wait on its activation:
+
+```js tab=js-ts
+await expect.poll(() => serviceworker.evaluate(() => (self as any).registration.active?.state)).toBe('activated');
+```
+
+```js tab=js-js
+await expect.poll(() => serviceworker.evaluate(() => self.registration.active?.state)).toBe('activated');
+```
```
rwoll, line 75: Post-landing the doc change, let's sort out an API to add to do this idiomatically. I don't like poll to begin with, and Python/NET/Java don't seem to have the equivalent.

If we want to hold of on introducing an API, but still want to provide a workaround, we can share the activationPromise strategy on Page instead of Service Worker.

docs/src/service-workers.md
```diff
@@ -0,0 +1,349 @@
...
+```html
+<!-- filename: index.html -->
+<script>
+  window.registrationPromise = navigator.serviceWorker.register('/transparent-service-worker.js');
+</script>
+```
+
+On the first visit to the page via [`method: Page.goto`], the following Request/Response events would be emitted (along with the corresponding network lifecycle events):
+
+| Event                             | Owner            | URL                            | Routed | [`method: Response.fromServiceWorker`] |
```
rwoll, line 121: Looks like this:

Screen Shot 2022-07-13 at 2 20 51 PM

docs/src/service-workers.md
```diff
@@ -0,0 +1,349 @@
...
+| [`event: BrowserContext.request`] | [Frame]          | index.html                     | Yes    |                                        |
+| [`event: Page.request`]           | [Frame]          | index.html                     | Yes    |                                        |
+| [`event: BrowserContext.request`] | Service [Worker] | transparent-service-worker.js  | Yes    |                                        |
+| [`event: BrowserContext.request`] | Service [Worker] | data.json                      | Yes    |                                        |
+| [`event: BrowserContext.request`] | [Frame]          | data.json                      |        | Yes                                    |
+| [`event: Page.request`]           | [Frame]          | data.json                      |        | Yes                                    |
+
+
+
+It's important to note that, since the example Service Worker just acts a basic transparent "proxy":
```
rwoll, line 132:
```suggestion
Since the example Service Worker just acts a basic transparent "proxy":
```

docs/src/service-workers.md
```diff
@@ -0,0 +1,349 @@
...
+
+* There's 2 [`event: BrowserContext.request`] events for `data.json`; one [Frame]-owned, the other Service [Worker]-owned.
+* Only the Service [Worker]-owned request for the resource was routable via [`method: BrowserContext.route`]; The [Frame]-owned events for `data.json` are not routeable, as they would not have even had the possibility to hit the external network since the Service Worker has a fetch handler registered.
```
rwoll, line 135:
```suggestion
* Only the Service [Worker]-owned request for the resource was routable via [`method: BrowserContext.route`]; the [Frame]-owned events for `data.json` are not routeable, as they would not have even had the possibility to hit the external network since the Service Worker has a fetch handler registered.
```

docs/src/service-workers.md
```diff
@@ -0,0 +1,349 @@
...
+      contentType: 'text/plain',
+      status: 200,
+      body: 'from sw',
+    });
+  } else {
+    return route.continue();
+  }
+});
+
+await context.RouteAsync("**", async route => {
```
rwoll, line 333:
```suggestion
await context.RouteAsync("**", async route => {
```

docs/src/service-workers.md
```diff
@@ -0,0 +1,349 @@
...
+[`event: BrowserContext.serviceWorker`] is fired ***before*** the Service Worker's main script has been evaluated, so ***before*** calling service[`method: Worker.evaluate`] you should wait on its activation:
+
+```js tab=js-ts
+await expect.poll(() => serviceworker.evaluate(() => (self as any).registration.active?.state)).toBe('activated');
+```
+
+```js tab=js-js
+await expect.poll(() => serviceworker.evaluate(() => self.registration.active?.state)).toBe('activated');
+```
+
```
rwoll, line 76:
```suggestion
[`event: BrowserContext.serviceWorker`] is fired ***before*** the Service Worker's main script has been evaluated, so ***before*** calling service[`method: Worker.evaluate`] you should wait on its activation.

```

Let's address this with a proper API post-landing the docs.

docs/src/service-workers.md
```diff
@@ -0,0 +1,327 @@
...
+1. Testing the page exactly like a user would experience it. This works out of the box in all supported browsers.
+1. Test your page without a Service Worker. Set [`option: Browser.newContext.serviceWorkers`] to `'block'`. You can test your page as if no Service Worker was registered.
+1. Listen for and route network traffic via Playwright, whether it comes from a Service Worker or not. In Firefox and WebKit, set [`option: Browser.newContext.serviceWorkers`] to `'block'` to avoid Service Worker network traffic entirely. In Chromium, either block Service Workers or use [`method: BrowserContext.route`].
+1. (Chromium-only) Test your Service Worker implementation itself. Use [`method: BrowserContext.serviceWorkers`] to get access to the Service Worker and evaluate there.
+
+## Disabling Service Workers
+
+Set [`option: Browser.newContext.serviceWorkers`] to `'block'`.
+
```
rwoll, line 23:
```suggestion
## Disabling Service Workers

Set [`option: Browser.newContext.serviceWorkers`] to `'block'`.

```

Actually, we already have a link directly to the option, so no need for this. It would just create an extra click

docs/src/service-workers-guide.md
rwoll: Agreed, just updated. See screenshot for what it looks like. I also changed the `Scope` heading to `Event` since we then put events underneath it.

docs/src/service-workers-guide.md
rwoll: A variant of that could be used. I was hoping to keep it on the SW PW object instead of the Page, but looks like we don't have generic poll in other language ports, so I've updated the workaround for now.

docs/src/service-workers.md
```diff
@@ -0,0 +1,370 @@
...
+```csharp
+var waitForServiceWorkerTask = page.WaitForServiceWorkerAsync();
+await page.GotoAsync('/example-with-a-service-worker.html');
+var serviceworker = await waitForServiceWorkerTask;
+```
+
+[`event: BrowserContext.serviceWorker`] is fired ***before*** the Service Worker's main script has been evaluated, so ***before*** calling service[`method: Worker.evaluate`] you should wait on its activation.
+
+There are more iodiomatic methods of waiting for a Service Worker to be activated, but the following is an implementation agnostic method (which will be replaced by the [Playwright feature](https://github.com/microsoft/playwright/issues/15636)):
```
dgozman, line 65: There's no need to make commitments in the documentation :)

```suggestion
There are more iodiomatic methods of waiting for a Service Worker to be activated, but the following is an implementation agnostic method:
```

docs/src/service-workers.md
```diff
@@ -0,0 +1,370 @@
...
+page.evaluate("fetch('/fallthrough.txt')")
+```
+
+```csharp
+await page.EvaluateAsync("fetch('/addressbook.json')");
+await page.EvaluateAsync("fetch('/foo')");
+await page.EvaluateAsync("fetch('/tracker.js')");
+await page.EvaluateAsync("fetch('/fallthrough.txt')");
+```
+
```
dgozman, line 280: Missing java examples everywhere :)

dgozman (approved):

PR #15644, opened 2022-07-14
docs: Add note about memory leaking in context

amokmen:


amokmen:
It seems that my PR is invisible )

Some explanation then.

mem_leak.js
```js
/*
PoW: memory leaking in context while long using page.goto() (increment meta-info in page).

2022-07-14T11:54:03.782Z Total memory consumed: 44884 KiB, heapTotal is: 30060 KiB, heapUsed is: 13270 KiB
...
2022-07-14T12:06:20.210Z Total memory consumed: 109548 KiB, heapTotal is: 80140 KiB, heapUsed is: 77701 KiB
*/

const { chromium } = require("playwright");

let browser;
let context;
let page;

(async () => {
  browser = await chromium.launch({
    headless: true,
  });
  context = await browser.newContext();
  page = await context.newPage();

  CycleGoto();

  // Start heartbeat
  HeartBeat();
})();

async function CycleGoto() {
  await page.goto("https://ya.ru");
  setTimeout(
    CycleGoto,
    3000, // ms
  );
}

function MemWatcher() {
  const objProcessMemoryUsage = process.memoryUsage();

  return `${new Date().toISOString()} Total memory consumed: ${
    objProcessMemoryUsage.rss / 1024
  } KiB, heapTotal is: ${
    objProcessMemoryUsage.heapTotal / 1024
  } KiB, heapUsed is: ${parseInt(
    objProcessMemoryUsage.heapUsed / 1024,
    0,
  )} KiB`;
}

function HeartBeat() {
  console.debug(MemWatcher());

  setTimeout(
    HeartBeat,
    5000, // ms
  );
}
```

docs/src/api/class-browsercontext.md
```diff
@@ -10,6 +10,8 @@ context.
 Playwright allows creating "incognito" browser contexts with [`method: Browser.newContext`] method. "Incognito" browser
 contexts don't write any browsing data to disk.
 
+:::note Attention! There would be memory consuming/leak around 150 MiB per 1 hour when using same context for a long time for operations page.goto(). Closing context would free memory. :::
```
pavelfeldman, line 13: That might be misleading - it can be 0 or it can be 150Mb per second, depending on the page. Also technically it is not a leak. What we should say is that context should be created per unit of work, because it retains information about the pages.

amokmen:
@pavelfeldman Updated note text.

docs/src/api/class-browsercontext.md
jfp1992: During my test run I get up to ~1.6GB after 10 minutes, then 2.5GB after another 10 minutes.

pavelfeldman:
As per my comment, 150 MiB/hour still does not make sense to me, I think it is misleading. I'll close this for now.

PR #15685, opened 2022-07-14
feat(expect): add 'subset' option for toHaveText and toContainText

### LINKED ISSUE #15666: [BUG] Explicitly control subset of elements in toContainText (and alike)

dgozman:
This removes the default 'subset' behavior from `toContainText` and adds an explicit option.

Fixes #15666.

### DIFF
--- a/docs/src/api/class-locatorassertions.md
+++ b/docs/src/api/class-locatorassertions.md
@@ -192,6 +192,9 @@ Expected substring or RegExp or a list of those.
 
 Whether to perform case-insensitive match. [`option: ignoreCase`] option takes precedence over the corresponding regular expression flag if specified.
 
+### option: LocatorAssertions.NotToContainText.subset = %%-assertions-subset-%%
+* since: v1.24
+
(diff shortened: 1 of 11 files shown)


mxschmitt (approved):

docs/src/api/class-locatorassertions.md
```diff
@@ -1316,6 +1325,40 @@ var locator = Page.Locator("list > .component");
 await Expect(locator).toHaveTextAsync(new string[]{ "Text 1", "Text 2", "Text 3" });
 ```
 
+You can also check just a subset of the elements matching the [Locator]. For example, in the list of ten items from `Item 1` to `Item 10`, only check the text of `Item 2`, `Item 4` and `Item 5`:
```
pavelfeldman, line 1328: Let's explicitly specify that the order matters.

pavelfeldman (approved):

pavelfeldman:
Looks like we aren't quite ready to break this.

PR #15700, opened 2022-07-15
cherry-pick(#15694): fix(esm loader): support Node 18.6

rwoll:
A small change in the API requires `shortCircuit: true` when returning the
transformed source.
Another API change removes third argument from the chained resolve/load,
but passing three instead of two still works.


rwoll:
oops dupe of https://github.com/microsoft/playwright/pull/15698

PR #15714, opened 2022-07-15
browser(webkit): rebase to 07/15/22

q66:
3-way diff: https://github.com/q66/WebKit/commit/b9a8cf3176c909241afa8734e0dcb33c15dca023

attempts to fix some API refactoring fallout after https://github.com/WebKit/WebKit/commit/55390aa216e406feeb3abdd5735de3da1130a8a2 - mac only, not sure if my fix builds (as far as i can tell, it should)

currently building on linux locally


q66:
Added build fixes:

- https://github.com/q66/WebKit/commit/208fc4dd1d887c7c611cb130f1e5d575da49b5bf - necessary after https://github.com/WebKit/WebKit/commit/fb205fb01a0863848066ea78dbfb0e2bc67b44a1 , quickest/shortest fix but playwright code should probably be adjusted after this
- https://github.com/q66/WebKit/commit/ee7d8d34763d10244c417bc52d9783a4338c86a9 - needed to stop the compiler from complaining about duplicate definitions (this exact block is already present upstream later in the file)

browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -15338,7 +15330,7 @@ index 0000000000000000000000000000000000000000..d9f566a6cce6433c2cfbb25fd6793f65
 +        items.append(language);
 +    }
 +
-+    browserContext->processPool->configuration().setOverrideLanguages(WTFMove(items));
++    WebKit::setOverrideLanguages(WTFMove(items));
```
yury-s, line 15333: This is going to break our tests, we need language override to be per context so that different tests with different language settings can run in parallel in a browser.

yury-s (approved):

browser_patches/webkit/patches/bootstrap.diff
q66: I dropped the API workaround and instead added the bare minimum revert of the changes that broke it: https://github.com/q66/WebKit/commit/ec4c7e874e004be2fffcacdca77d36dc54304e6b

It's not an exact revert, as that would make the patch grow quite a bit more (there are at least two commits and they are both bigger than this). Instead, I picked the changes from both while taking care of having to move around as little as possible.

I think we could have this for now, and drop those patch bits once the situation is solved upstream? I marked them with comments to make it clear what will need dropping.

mxschmitt:
TBR Yury once he is back

PR #15719, opened 2022-07-15
chore: reuse context in the innerloop mode

pavelfeldman:


packages/playwright-test/src/index.ts
```diff
@@ -507,7 +507,7 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
       testInfo.errors.push({ message: prependToError });
   }, { scope: 'test',  _title: 'context' } as any],
 
-  _contextReuseEnabled: !!process.env.PW_REUSE_CONTEXT,
+  _contextReuseEnabled: !!process.env.PW_REUSE_CONTEXT || true,
```
dgozman, line 510: You don't want to commit this line :)

packages/playwright-core/src/server/browser.ts
```diff
@@ -90,6 +91,17 @@ export abstract class Browser extends SdkObject {
     return context;
   }
 
+  async newContextForReuse(params: channels.BrowserNewContextForReuseParams, metadata: CallMetadata): Promise<{ context: BrowserContext, needsReset: boolean }> {
+    const hash = BrowserContext.reusableContextHash(params);
+    if (!this._contextForReuse || hash !== this._contextForReuse.hash || !this._contextForReuse.context.canResetForReuse()) {
+      if (this._contextForReuse)
+        await this._contextForReuse.context.close(metadata);
+      this._contextForReuse = { context: await this.newContext(metadata, params), hash };
+      return { context: this._contextForReuse.context, needsReset: false };
```
dgozman, line 100: I don't see where would we do the following code in this `needsReset: false` case.

```js
    for (const key of paramsThatAllowContextReuse)
      (this._options as any)[key] = params[key];
```

Issue #11143, opened 2021-12-31
[Feature] TRX Reporter


creage:
We are using Azure DevOps for our CI/CD, and report our test results using JUnit reporter for now. But this reporter is very limited in functionality - it cannot deliver attachments directly as a part of the report itself.

Would be nice if you guys there at Microsoft create a Visual Studio TRX reporter for Playwright, which has greater support in Azure, so we can deliver attachments (traces, screenshots, logs) directly to our pipeline summary page.

https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=trx%2Cyaml#attachments-support

bigredjoe:
I second this feature. We use Azure DevOps and a TRX reporter with attachments for playwright would really help us debug our e2e tests in our pipeline.

jwalakom:
@creage @bigredjoe 

We too want to generate videos as attachments for the failed functional test cases on our ADO Pipelines using Playwright. Can you please let me know if there is any other way to do this? Thanks!

creage:
@jwalakom well, you can zip your attachments and download them using PowerShell task in your pipeline, and then publish them wherever you want. You can't attach your logs per test result, but you'll have them all available at least.

Camo30:
Any news about this issue? I would highly appreciate a TRX reporter for Playwright.

murraymichal:
> @jwalakom well, you can zip your attachments and download them using PowerShell task in your pipeline, and then publish them wherever you want. You can't attach your logs per test result, but you'll have them all available at least.

that's not what we are after. we'd like attachments to be available right in Tests section, as they should.

jwalakom:
> @jwalakom well, you can zip your attachments and download them using PowerShell task in your pipeline, and then publish them wherever you want. You can't attach your logs per test result, but you'll have them all available at least.

We have authored the tests in Azure Test plans with Test IDs. Then we "link" them to our functional tests. Playwright has a feature to generate test videos for failed test functional cases. Now we want to only attach these results ( videos in this case ) to our ADO pipeline as attachments. 

Any thoughts on this? Thanks! @murraymichal @Camo30 @creage

ivan-xue-prospa:
It will be awesone if Playwright has trx reporter built.

[...]

PR #15736, opened 2022-07-16
Azure DevOps reporter

alexneo2003:
Hi all.

I'm have my own package with [Azure DevOps reporter](https://github.com/alexneo2003/playwright-azure-reporter) and @pavelfeldman suggested that I publish a reporter as part of the project
You can look at package [README](https://github.com/alexneo2003/playwright-azure-reporter/blob/master/README.md) but the main code will have a lot of fixes compared to the package code
The main purpose of a report is to publish test results, including attachments, to Azure DevOps based on the text case ID, run ID, project name and organisation url

The code was written based on my own experience with Azure DevOps and the needs of my particular project. Ready to modify for additional conditions and needs

Compared to the code of my project, I added the ability to specify what type of attachment to send through the parameters
```ts 
    uploadAttachments: true,
    attachmentsType: ['screenshot', 'trace', 'video']
```

When writing tests i'm used the [nock](https://github.com/nock/nock) package to mock network requests
Reporter use [azure-devops-node-api](https://github.com/microsoft/azure-devops-node-api) to make requests to Azure DevOps

Feature request: https://github.com/microsoft/playwright/issues/11143


packages/playwright-test/package.json
```diff
@@ -33,6 +33,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@types/node": "*",
+    "azure-devops-node-api": "^11.2.0",
```
weyert, line 36: Maybe could be a optional peer dependency?

alexneo2003: if it's better, i can do it

mxschmitt, line 36: Would be amazing if we can get rid of that dependency, this would mean to write our own API client which calls to the azure devops API.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+    this.planId = this._options.planId;
+    this.logging = (typeof this._options.logging === 'undefined') ? true : this._options.logging;
+    this.isDisabled = this._options.isDisabled || false;
+    this.testRunTitle = `${this.environment ? `[${this.environment}]:` : ''} ${this._options.testRunTitle || 'Playwright Test Run'}` ||
+      `${this.environment ? `[${this.environment}]:` : ''}Test plan ${this.planId}`;
+    this.uploadAttachments = this._options.uploadAttachments || false;
+    this.attachmentsType = this._options.attachmentsType;
+    this.token = this._options.token;
+
+    if (!this.orgUrl || this.orgUrl.length === 0) {
```
mxschmitt, line 145:
```suggestion
    if (!this.orgUrl) {
```
an empty string is always falsy, so here and below you can just do !foo.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+      this.log(colors.yellow("'orgUrl' is not set. Reporting is disabled."));
+      this.isDisabled = true;
+      return;
+    }
+    if (!this.projectName || this.projectName.length === 0) {
```
mxschmitt, line 150: I'd maybe extract this validation logic out into an own function called "isDisabled" which you then call from the constructor.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+  private token: string = '';
+
+  public constructor(options: AzureReporterOptions = {} as AzureReporterOptions) {
+    this._options = options;
+
+    this.orgUrl = this._options.orgUrl;
+    this.projectName = this._options.projectName;
+    this.environment = this._options.environment || undefined;
+    this.planId = this._options.planId;
+    this.logging = (typeof this._options.logging === 'undefined') ? true : this._options.logging;
```
mxschmitt, line 137: I'd disable logging by default.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+    this.testApi = await this.connection.getTestApi();
+
+    this.createRun(this.testRunTitle).then(run => {
+      if (run) {
+        this.runId = run.id;
+        this.log(colors.green(`Using run ${this.runId} to publish test results`));
+      } else {
+        this.isDisabled = true;
+      }
+    });
```
mxschmitt, line 186:
```suggestion
    const run = await this.createRun(this.testRunTitle)
    if (run) {
      this.runId = run.id;
      this.log(colors.green(`Using run ${this.runId} to publish test results`));
    } else {
      this.isDisabled = true;
    }
```

try always avoid `.then` uses since it makes it not readable.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+
+  async onEnd(): Promise<void> {
+    await this.awaitForRunId().catch(e => {
+      this.log(colors.red(e));
+    });
+    if (this.isDisabled)
+      return;
+
+    let prevCount = this.resultsToBePublished.length;
+    while (this.resultsToBePublished.length > 0) {
```
mxschmitt, line 208: instead of having this logic here, I'd collect all the stuff in onTestEnd and then upload all the stuff on onEnd. So you don't need to wait for uploads to happen etc. and it makes it easier to understand.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+      await new Promise(resolve => setTimeout(resolve, 250));
+    }
+
+    if (this.publishedResultsCount === 0 && !this.runId) {
+      this.log('No testcases were matched. Ensure that your tests are declared correctly.');
+      return;
+    }
+
+    try {
+      const runUpdate: TestInterfaces.RunUpdateModel = {
```
mxschmitt, line 225: it's not used anywhere else, I'd put it right into line 230, instead of storing it inside a variable.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+      const runUpdatedResponse = await this.testApi.updateTestRun(runUpdate, this.projectName, this.runId as number);
+      this.log(colors.green(`Run ${this.runId} - ${runUpdatedResponse.state}`));
+    } catch (err) {
+      this.log(`Error on completing run ${err as string}`);
+    }
+  }
+
+  private log(message?: any) {
+    if (this.logging)
+      console.log(colors.magenta(`azure: ${message}`));
```
mxschmitt, line 239: since you are writing to the console, you need to override `printsToStdio` function and return either true.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+
+  }
+
+  private getCaseIds(test: TestCase): string {
+    const regexp = /\[([\d,]+)\]/;
+    const results = regexp.exec(test.title);
```
mxschmitt, line 245:
```suggestion
    const results = /\[([\d,]+)\]/.exec(test.title);
```

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+    };
+    if (testResult.status)
+      this.log(map[testResult.status]);
+
+  }
+
+  private async createRun(runName: string): Promise<TestInterfaces.TestRun> {
+    return new Promise<TestInterfaces.TestRun>(async (resolve, reject) => {
+      try {
+        this.checkProject(this.projectName).catch(e => {
```
mxschmitt, line 269: you need to await this one.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+          reject(e);
+        });
+        const runModel: TestInterfaces.RunCreateModel = {
+          name: runName,
+          automated: true,
+          configurationIds: [1],
+          plan: { id: `${this.planId}` }
```
mxschmitt, line 276:
```suggestion
          plan: { id: this.planId }
```

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+      disabled: colors.gray(`Test ${test.title} - ${testResult.status}`),
+      timedOut: colors.gray(`Test ${test.title} - ${testResult.status}`)
+    };
+    if (testResult.status)
+      this.log(map[testResult.status]);
+
+  }
+
+  private async createRun(runName: string): Promise<TestInterfaces.TestRun> {
+    return new Promise<TestInterfaces.TestRun>(async (resolve, reject) => {
```
mxschmitt, line 267: just thinking if a rewrite here makes sense without using "new Promise". I think it would make it easier to follow.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+  private planId = 0;
+  private runId!: number;
+  private logging = true;
+  private isDisabled = false;
+  private testRunTitle = '';
+  private uploadAttachments = false;
+  private attachmentsType?: TAttachmentType;
+  private token: string = '';
+
+  public constructor(options: AzureReporterOptions = {} as AzureReporterOptions) {
```
mxschmitt, line 130:
```suggestion
  public constructor(options: Partial<AzureReporterOptions> = {}) {
```
I think thats what you actually want.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+  'x-cache': string;
+  'x-msedge-ref': string;
+  'date': string;
+  'connection': string;
+}
+
+class AzureDevOpsReporter implements Reporter {
+  private testApi!: Test.ITestApi;
+  private coreApi!: ICoreApi;
+  private _options: AzureReporterOptions;
```
mxschmitt, line 114: you don't need this since you already "preprocess" the options and store them as properties inside the class.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,481 @@
...
+        if (project)
+          resolve(project);
+        else
+          reject(`Project ${projectName} does not exist. Reporting is disabled.`);
+
+      } catch (e) {
+        reject(e);
+      }
+    }).catch(e => {
+      this.log(e);
```
mxschmitt, line 313: I don't like the general way of error handling, that it gets logged, then isDisabled is true and then after the actual function call you check if its disabled to return otherwise. Try to avoid using .catch, new Promise, either wrap the errors into new ones to make them readable, if some are expected or don't handle them with too much precaution.

alexneo2003: done

packages/playwright-test/src/reporters/azure.ts
alexneo2003: done

packages/playwright-test/package.json
alexneo2003: it's a great idea but i think it's "out of budget"))

packages/playwright-test/src/reporters/azure.ts
alexneo2003: let's i'm trying to explain why i'm using this
1. if i'm upload results in `onTestEnd` user does not affected and results published parallel with main process, includes all attachments and the user will not even see how the test results are loaded.  If i'm try collect results and publish in `onEnd` - it would be have extra time. if it be only test results - it does not matter but if it includes attachments - it will be much more time to upload, because it's not local logic - it's network activity.
2. `resultsToBePublished` and `awaitForRunId`. if you test scope is more then `expect(1).toBe(1)`  - it does not matter. But if you have case like `expect(1).toBe(1); `- you need to await network activities like `createRun`, `publishCaseResult` because `onTestEnd` and `onEnd` hooks trigger much earlear and faster then network methods would be receive results. And we will receive result when main test run process will be ended but network requests not yet completed. If i'm remove this logic reporter will not work as expected of him, he will not have time to publish the results

PavanMudigondaTR:
@alexneo2003 i have used your reporter earlier and it worked well. I faced one limitation. All test cases need to be present under one suiteID. Actually we keep test cases in different suites for each Feature.

alexneo2003:
> @alexneo2003 i have used your reporter earlier and it worked well. I faced one limitation. All test cases need to be present under one suiteID. Actually we keep test cases in different suites for each Feature. 

Maybe you mean planId? As i seen planId is highest level at test cases what includes test suites.

PavanMudigondaTR:
> > @alexneo2003 i have used your reporter earlier and it worked well. I faced one limitation. All test cases need to be present under one suiteID. Actually we keep test cases in different suites for each Feature. 
> 
> Maybe you mean planId? As i seen planId is highest level at test cases what includes test suites.
> 

Azure DevOps instance could have multiple projects. One one particular project there could be multiple Test Plans (test plan IDs). If you take one Test Plan it can have multiple test suites. If you would provide option to specify testPlanID,  suiteID and testCaseID for each test case that would be awesome. Rest of variables such as PAT token, project id, ADO Org URL can be common.

PavanMudigondaTR:
> > > @alexneo2003 i have used your reporter earlier and it worked well. I faced one limitation. All test cases need to be present under one suiteID. Actually we keep test cases in different suites for each Feature.
> > 
> > 
> > Maybe you mean planId? As i seen planId is highest level at test cases what includes test suites.
> 
> Azure DevOps instance could have multiple projects. One one particular project there could be multiple Test Plans (test plan IDs). If you take one Test Plan it can have multiple test suites. If you would provide option to specify testPlanID, suiteID and testCaseID for each test case that would be awesome. Rest of variables such as PAT token, project id, ADO Org URL can be common.

To add to my above feedback, for those people who are storing all test cases in single Test Plan and single Test Suite,  you can continue your current capability as well. So if the testPlanID, TestSuiteID are not found at test case level then it will search and use the same from playwright.config.ts

packages/playwright-test/package.json
mxschmitt: Was checking the bundle size, seems definitely a lot, putting it maybe under an extra package, aka `@playwright/ado-reporter` might make sense.

```
npx esbuild --bundle test.js --outfile=out.js --platform=node

out.js  1.7mb ⚠️
```

weyert: What about making it an optional peer dependency that people manually need to add to the project when using this reporter?

tests/playwright-test/reporter-azure.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+      test('[3] foobar', async () => {
+        expect(1).toBe(1);
+      });
+      `
+  }, { reporter: '' });
+  expect(stripAnsi(result.output)).not.toContain('Failed request: (401)');
+  expect(stripAnsi(result.output)).toContain('azure: Using run 150 to publish test results');
+  expect(stripAnsi(result.output)).toContain('azure: Test [3] foobar - passed');
+  expect(stripAnsi(result.output)).toContain('azure: Start publishing: [3] foobar');
+  expect(stripAnsi(result.output)).toContain('Could not find test point for test case [3] associated with test plan 44. Check, maybe testPlanId, what you specifiyed, is incorrect');
```
mxschmitt, line 331:
```suggestion
  expect(stripAnsi(result.output)).toContain('Could not find test point for test case [3] associated with test plan 44. Check, maybe testPlanId, what you specified, is incorrect');
```

tests/playwright-test/reporter-azure.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+  expect(stripAnsi(result.output)).not.toContain('azure: Result published: [3] foobar');
+  expect(stripAnsi(result.output)).toContain('azure: Run 150 - Completed');
+  expect(result.exitCode).toBe(0);
+});
+
+test('05 upload attachments, attachmentsType in not defined - default "screenshot" @azure', async ({ runInlineTest }) => {
```
mxschmitt, line 337: I think we can drop the `@azure` everywhere, since you can run it instead by the file.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,475 @@
...
+    }
+    if (!options?.planId) {
+      this.warning("'planId' is not set. Reporting is disabled.");
+      this.isDisabled = true;
+      return;
+    }
+    if (!options?.token) {
+      this.warning("'token' is not set. Reporting is disabled.");
+      this.isDisabled = true;
+    }
```
mxschmitt, line 156: why don't we return here?

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,475 @@
...
+  }
+
+  private async awaitForRunId(): Promise<number | void> {
+    // need wait runId variable to be initialised in onBegin() hook
+    if (this.isDisabled)
+      return;
+    try {
+      const timeout = 10_000;
+      const startTime = Date.now();
+      while (this.runId === undefined && Date.now() - startTime < timeout)
```
mxschmitt, line 414: this can be done nicer in the following way.

1. inside constructor do the following:
```js
this.runIdPromise = new Promise((resolve, reject) => {
  this.resolveRunId = resolve;
  this.rejectRunId = reject;
});
```
2. inside `onBegin()` remove that it returns a Promise, since its a sync method. and do something like that:
```js
  onBegin(): void {
    if (this.isDisabled)
      return;
    this.connection = new azdev.WebApi(this.orgUrl, azdev.getPersonalAccessTokenHandler(this.token));

    (async () => {
      this.testApi = await this.connection.getTestApi();
      const run = await this.createRun(this.testRunTitle);
      if (run) {
        this.resolveRunId(run.id)
        this.log(colors.green(`Using run ${this.runId} to publish test results`));
      } else {
        this.isDisabled = true;
      }
    })(error => {
      this.isDisabled = true;
      this.warning(`Failed to create test run: ${error.message}. Check your token and orgUrl. Reporting is disabled.`);
      rejectRunId(error);
    });
  }
```
3. in the places you need the run id, just do the following:
```
const runId = await this.runIdPromise;
```

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,475 @@
...
+      `${this.environment ? `[${this.environment}]:` : ''}Test plan ${this.planId}`;
+    this.uploadAttachments = options?.uploadAttachments || false;
+  }
+
+  async onBegin(): Promise<void> {
+    if (this.isDisabled)
+      return;
+
+    try {
+      this.connection = new azdev.WebApi(this.orgUrl, azdev.getPersonalAccessTokenHandler(this.token));
```
mxschmitt, line 182: move this line into the constructor.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,475 @@
...
+      this.log(`Error on completing run ${err as string}`);
+    }
+  }
+
+  printsToStdio() {
+    return true;
+  }
+
+  private log(message?: any) {
+    if (this.logging)
```
mxschmitt, line 244: I'm thinking if we should drop all that logging and use our common way of logging with the [`debug`](https://www.npmjs.com/package/debug) library, e.g. `pw:test:azure` as a key and document it.

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,475 @@
...
+      pending: colors.blue(`Test ${test.title} - ${testResult.status}`),
+      disabled: colors.gray(`Test ${test.title} - ${testResult.status}`),
+      timedOut: colors.gray(`Test ${test.title} - ${testResult.status}`)
+    };
+    if (testResult.status)
+      this.log(map[testResult.status]);
+
+  }
+
+  private async createRun(runName: string): Promise<TestInterfaces.TestRun | void> {
```
mxschmitt, line 274: lets prefix all the private methods with an underscore?

packages/playwright-test/src/reporters/azure.ts
```diff
@@ -0,0 +1,475 @@
...
+      }
+      return pointsIds;
+    } catch (e) {
+      this.log(colors.red(e.message));
+    }
+    return pointsIds;
+  }
+
+  private addReportingOverride = (api: Test.ITestApi): Test.ITestApi => {
+    // https://github.com/microsoft/azure-devops-node-api/issues/318#issuecomment-498802402
```
mxschmitt, line 347: having a link is a good start, but lets also add a few words why we need it exactly before the link.

package.json
```diff
@@ -83,6 +83,7 @@
     "license-checker": "^25.0.1",
     "mime": "^3.0.0",
     "ncp": "^2.0.0",
+    "nock": "^13.2.7",
```
mxschmitt, line 86: would be also nice if we could drop the nock dependency. I'd do it the following way:

1. when an env var is set, use a localhost instead of the official ado host
2. spawn a http server inside our tests and return the relevant responses

alexneo2003: @mxschmitt i'm have some situation with this step
i'm using [TestServer](https://github.com/microsoft/playwright/tree/main/utils/testserver) to write ADO mock server where i'm defining all i needed routs with response
I took one first and faced some difficulties
i'm connect test server as pw config
```ts
webServer: {
  command: 'node ${JSON.stringify(SIMPLE_SERVER_PATH)} ${port}',
  port: ${port},
},
```
and when test runned it failed because last request from reporter does not intercepting with test server - test server will be destroyed with pw test-runner (i think so), and last request from azure reporter onEnd hook does not be responding
i solved it using `spawn` from `child_process` package:
1. spawn node process with server
2. run test with 'runInlineTest' method 
3. kill spawned process with `child.kill('SIGINT')` command
4. check test results

this variant working
else maybe you have some idea?

mxschmitt: We have an inline fixture for our server see here: https://github.com/microsoft/playwright/blob/d12afa7ff399c8a9479394bcb1fc492d192e8173/tests/playwright-test/web-server.spec.ts#L434-L437

If you want to server static assets, I recommend to put them into `tests/assets/playwright-test/reporters/azure` and use then `server.serveFile` inside your actual test.

I think using this makes the other questions irrelevant? Feel free to ping me on Slack as well if you have further questions: https://aka.ms/playwright-slack

alexneo2003:
@mxschmitt 
at last tests run receive some errors at `ubuntu-latest, 14` and `ubuntu-latest, 16` at same test
i can't reproduce this at local, i'm run it on macOS, windows, ubuntu 14-16 node version but not receive the same result what in PR :-(

PavanMudigondaTR:
Any chance this can come in next release?

ShuiRuTian:
Hey guys, one small question about Token.

If it's an open source repo, should I store the token in the repo? It feels like should not, but then how others run this report?

PavanMudigondaTR:
> Hey guys, one small question about Token.
> 
> If it's an open source repo, should I store the token in the repo? It feels like should not, but then how others run this report?

Azure DevOps PAT Token should be coming from Environment variable and not to be stored directly in config files. If I am running in CI I would set env variable in GitHub Action from GitHub Secrets. In JavaScript/TypeScript "process.env.ADO_PAT" would retrieve token from environment variables set in local computer or in GH Action Runner Context/Env.

pomi:
Hi, Is there a way to add test configuration to test results?

alexneo2003:
> Hi, Is there a way to add test configuration to test results?

did you mean to pass [this](https://docs.microsoft.com/en-us/rest/api/azure/devops/test/runs/create?view=azure-devops-rest-7.1&tabs=HTTP#request-body) data through config to created test run?
like this?
Monosnap Runs - Test Plans 2022-09-13 15-21-27

pomi:
> > Hi, Is there a way to add test configuration to test results?
> 
> did you mean to pass [this](https://docs.microsoft.com/en-us/rest/api/azure/devops/test/runs/create?view=azure-devops-rest-7.1&tabs=HTTP#request-body) data through config to created test run? like this? Monosnap Runs - Test Plans 2022-09-13 15-21-27



> > Hi, Is there a way to add test configuration to test results?
> 
> did you mean to pass [this](https://docs.microsoft.com/en-us/rest/api/azure/devops/test/runs/create?view=azure-devops-rest-7.1&tabs=HTTP#request-body) data through config to created test run? like this? Monosnap Runs - Test Plans 2022-09-13 15-21-27

Yes, exactly that. I am really interested in build and configurationids parameters.

alexneo2003:
> Yes, exactly that. I am really interested in build and configurationids parameters.

I think it's possible

mechandan:
If a test case fails the error message should be there in the test run right??

 I am actually not getting the error , its blank , any leads ??

pavelfeldman:
This has been open for many months w/o reviews. I read it as a lack of the reviewing capacity for the priority of the change, so I'm closing it based on that. Let's continue the conversation in the Issue. Sorry about the hassle.

woeterman94:
@pavelfeldman Can  you please reconsider this? We need this feature

alexneo2003:
> @pavelfeldman Can you please reconsider this? We need this feature

Try using this https://www.npmjs.com/package/@alex_neo/playwright-azure-reporter

PR #15777, opened 2022-07-19
chore: fix Android 13 compatibility

### LINKED ISSUE #15776: [BUG] Playwright for Android does not work with Android 13
Caused by the "Intent filters block non-matching intents"

See here: https://developer.android.com/about/versions/13/behavior-changes-13

This needs to be adjusted: https://github.com/mxschmitt/playwright/blob/1471c526addb1c66a4ae1a259ce045333940a90a/packages/playwright-core/src/server/android/android.ts#L253

mxschmitt:
Fixes #15776
Relates #15065

This also increases the Chrome version from 93 to 101.


utils/avd_test.js
```diff
@@ -1,63 +0,0 @@
-/**
```
mxschmitt, line 1: This script was unused, 2 years old and not working. We use our tests instead to ensure Android is working.

utils/avd_start.sh
```diff
@@ -8,16 +8,15 @@ if [[ -z "${ANDROID_HOME}" ]]; then
     export ANDROID_SDK_ROOT=${SDKDIR}
 fi
 
-echo "Killing previous emulators"
```
mxschmitt, line 11: Moved this out, since its useful to just stop the emulator.

packages/playwright-core/src/server/android/android.ts
```diff
@@ -250,7 +250,12 @@ export class AndroidDevice extends SdkObject {
     const commandLine = `_ --disable-fre --no-default-browser-check --no-first-run --remote-debugging-socket-name=${socketName}`;
     debug('pw:android')('Starting', pkg, commandLine);
     await this._backend.runCommand(`shell:echo "${commandLine}" > /data/local/tmp/chrome-command-line`);
-    await this._backend.runCommand(`shell:am start -n ${pkg}/com.google.android.apps.chrome.Main about:blank`);
+    const canBecomeRoot = await this._backend.runCommand(`shell:su root whoami`).then(result => result.toString().trim() === 'root');
+    // Android 13 introduced "Intent filters block non-matching intents" which prevent non-root users from launching different apps with data_uri.
```
mxschmitt, line 254: https://developer.android.com/about/versions/13/behavior-changes-13#intent-filters

PR #15797, opened 2022-07-19
test: android test enhancements

mxschmitt:
Extracted from #15777.


tests/page/locator-highlight.spec.ts
```diff
@@ -16,16 +16,34 @@
...
 });
+
+function roundBox(box: BoundingBox): BoundingBox {
+  return {
+    x: Math.round(box.x),
+    y: Math.round(box.y),
+    width: Math.round(box.width),
+    height: Math.round(box.height),
+  };
+}
```
rwoll, line 49: Will this deflake it? Might be better to check they are mostly overlapping instead. `overlapping(box1, box2, tolerance)`

utils/avd_stop.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+if [[ -z "${ANDROID_HOME}" ]]; then
+    SDKDIR=$PWD/.android-sdk
+    export ANDROID_HOME=${SDKDIR}
+    export ANDROID_SDK_ROOT=${SDKDIR}
+fi
+
+echo "Killing previous emulators"
+${ANDROID_HOME}/platform-tools/adb devices | grep emulator | cut -f1 | while read line; do ${ANDROID_HOME}/platform-tools/adb -s $line emu kill; done
```
rwoll, line 12: Why is this a separate file? Looks only used on one place and makes it hard to follow `utils/avd_start.sh`?

rwoll (approved):

tests/page/locator-highlight.spec.ts
mxschmitt: yes, it will deflate, its 0.00001 off on Android.

utils/avd_stop.sh
mxschmitt: It's helpful when running avd locally.

PR #15804, opened 2022-07-20
devops: send Telegram messages via Node.js

mxschmitt:
This should fix the broken emojis: 
image

Relates https://stackoverflow.com/questions/41383336/curl-doesnt-encode-utf-8


browser_patches/send_telegram_message.js
```diff
@@ -0,0 +1,79 @@
...
+    console.log('Text not set!');
+    console.log('Usage: node send_telegram_message.js <text>');
+    return;
+  }
+
+  await sendTelegramMessage(TELEGRAM_BOT_KEY, text);
+  console.log('Telegram message sent successfully!');
+})().catch(error => {
+  console.log(`Failed to send Telegram message. Error: ${error}`);
+  process.exit(1);
```
rwoll, line 40: This looks like a change in behavior that's undesirable. Now, if Telegram is down or flaky in a request, our build bot script will exit early in a number of places. 

Let's either exit with 0, so we still get a build; or wherever you call this script in the build script do `… || true`.

It's OK if we don't get a telegram message, but we want our builds!

While Telegram is unlikley to be down, I said the same thing about Slack in build scripts in other places, and the day Slack was down—actually several times in a year—I had to modify a build script just to get the important build artifact out.

rwoll (changes requested):

rwoll (approved):

PR #15812, opened 2022-07-20
fix(lifecycle): recalculate lifecycle on iframe detach

### LINKED ISSUE #15084: [BUG] waitForNavigation fails to complete intermittently on Chromium
**Context:**
- Playwright Version: 1.22.2
- Operating System: Linux
- Node.js version: 16.15.0
- Browser: Chromium
- Extra: Running under Jenkins with Docker image mcr.microsoft.com/playwright:v1.22.2-focal

## System:
 - OS: Linux 5.10 Ubuntu 20.04.4 LTS (Focal Fossa)
 - Memory: 13.82 GB / 15.45 GB
 - Container: Yes
## Binaries:
 - Node: 16.15.0 - /usr/bin/node
 - Yarn: 1.22.18 - /usr/bin/yarn
 - npm: 8.5.5 - /usr/bin/npm
## Languages:
 - Bash: 5.0.17 - /usr/bin/bash

**Describe the bug**
We are seeing random random failures an almost every run of our tests. The tests run 6 times per day and maybe one run be successful per day, but mostly we have at least one failure. It is a different test that fails each time. We have not seen the problem when running locally - only on the Jenkins server, but we don't run it locally as often. Our web application is an Angular application that logs into the server and receives an access token. After some time the access token is refreshed in the background.

The failure is that the test does a Promise.all for a page.waitForNavigation (waitUntil = networkidle) and a button click to start the navigation. When the test failed we get a network timeout waiting for the networkidle event. We have tried using other waituntil values but the same thing happens. It seems that under some circumstance the waitForNavigation is not completing for some reason.

The access token refresh seems to be involved in the problem. It doesn't cause a failure every time the token refresh runs during a test but could be the timing between how the refresh interleaves with the browser application and the playwright test execution. After capturing different logs with DEBUG=pw:api and DEBUG=pw:channel* we see the token refresh in all the failed tests and spot checking the same tests that succeeded in other runs the token refresh is not present.

**Failure 1**
This a a portion of the log for DEBUG=pw:api for a test that failed with some comments added. In a successful run we see a message for "pw:api <= page.waitForNavigation succeeded" that is not in the failed test.
[silent-refresh.txt](https://github.com/microsoft/playwright/files/8970347/silent-refresh.txt)

**Failure 2**
This scenario failed with DEBUG=pw:channel*. One file is for the failed run of one of the tests. The other file is the same test when it succeeded in another run. 
[mappingdata-fail.txt](https://github.com/microsoft/playwright/files/8970364/mappingdata-fail.txt)
[mappingdata-success.txt](https://github.com/microsoft/playwright/files/8970365/mappingdata-success.txt)

The access token refresh creates a new frame to perform the refresh and then deletes it. Could the temporary frame be interfering with the events on the main frame the application is in?

dgozman:
It could be that iframe was blocking some event, like load or networkidle, and we never updated the lifecycle after the iframe was detached. This lead to goto and other navigation commands to never resolve.

Fixes #15084.

### DIFF
--- a/tests/page/page-goto.spec.ts
+++ b/tests/page/page-goto.spec.ts
@@ -663,3 +663,35 @@ it('should return when navigation is committed if commit is specified', async ({
   const response = await page.goto(server.EMPTY_PAGE, { waitUntil: 'commit' });
   expect(response.status()).toBe(200);
 });
+
+it('should wait for load when iframe attaches and detaches', async ({ page, server }) => {
+  server.setRoute('/empty.html', (req, res) => {
+    res.writeHead(200, { 'content-type': 'text/html' });
+    res.end(`
+      <body>
+        <script>
+          const iframe = document.createElement('iframe');
+          iframe.src = './iframe.html';
+          document.body.appendChild(iframe);
+          setTimeout(() => iframe.remove(), 1000);
+        </script>
+      </body>
+    `);
+  });
+
+  server.setRoute('/iframe.html', (req, res) => {
+    res.writeHead(200, { 'content-type': 'text/html' });
+    res.end(`
+      <link rel="stylesheet" href="./style2.css">
+    `);
+  });
+
+  // Stall the css so that 'load' does not fire.
+  server.setRoute('/style2.css', () => {});
+
+  const frameDetached = page.waitForEvent('framedetached');
+  const done = page.goto(server.EMPTY_PAGE, { waitUntil: 'load' });
+  await frameDetached; // Make sure that iframe is gone.
+  await done;
+  expect(await page.$('iframe')).toBe(null);
+});
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -277,8 +277,11 @@ export class FrameManager {
 
   frameDetached(frameId: string) {
     const frame = this._frames.get(frameId);
-    if (frame)
+    if (frame) {
       this._removeFramesRecursively(frame);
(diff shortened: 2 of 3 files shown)


mxschmitt (approved):

tests/page/page-goto.spec.ts
```diff
@@ -663,3 +664,42 @@ it('should return when navigation is committed if commit is specified', async ({
...
+    res.writeHead(200, { 'content-type': 'text/html' });
+    res.end(`
+      <body>
+        <script>
+          setTimeout(() => {
+            const iframe = document.createElement('iframe');
+            iframe.src = './iframe.html';
+            document.body.appendChild(iframe);
+            setTimeout(() => iframe.remove(), 1000);
+          }, 100);
```
aslushnikov, line 679: remove 100 ms timeout? otherwise `load` event might fire!

tests/page/page-goto.spec.ts
```diff
@@ -663,3 +664,42 @@ it('should return when navigation is committed if commit is specified', async ({
...
+  });
+
+  const cssPromise2 = new Promise<ServerResponse>(f => {
+    server.setRoute('/style2.css', (req, res) => f(res));
+  });
+
+  const done = page.goto(server.EMPTY_PAGE, { waitUntil: 'load' });
+
+  const css2 = await cssPromise2;
+  await new Promise(f => setTimeout(f, 2000));  // Make sure iframe is gone.
```
aslushnikov, line 699: wait for `framedetached` instead?

aslushnikov (approved):

PR #15854, opened 2022-07-21
docs: playwright test v. library

rwoll:
Currently, our Playwright Test and Playwright Library pages acknowledges each exist, but don't really spell out the difference between the two.

The goal with this page is:

1. Clarify which package a user should be using
2. If using Playwright Library, show what's required

Depending on the content of this page, it may be possible to make our docs completely `@playwright/test`-first (including the examples), and then just have one doc that answers "if you're using Playwright Library, here's the few unique bits to it that you didn't have to think about in Playwright Test".

The less duplication we have across Library vs. Test docs, the less room there is for confusion and maintenance burden.

An alternate version of this that doesn't compare the two can be viewed here: #15888


rwoll:
Here's a preview of the table (but read the source or render it locally as I will **NOT** keep this screenshot comment up to date):

Screen Shot 2022-07-21 at 12 47 10 PM

docs/src/playwright-test-versus-library-js.md
```diff
@@ -0,0 +1,127 @@
+---
+id: playwright-test-vs-library
+title: "Playwright Test vs. Library"
+---
+
+
+Playwright Library provides unified APIs for launching and interacting with browsers, while Playwright Test provides all this plus a fully managed end-to-end Test Runner and experience.
+
+Under most circumstances, for end-to-end testing, you'll want to use `@playwright/test` (Playwright Test), and not `playwright` (Playwright Library) directly. If you are integrating Playwright with a third-party test runner, or are working on scraping/automation, you will use Playwright Library directly.
```
rwoll, line 9: I'm not sure yet how a user arrives at this doc; however, if you get here, I want to quickly get the user out and back the testing docs unless they *really* want to use the lib directly and therefore need this info.

rwoll, line 9:
```suggestion
Under most circumstances, for end-to-end testing, you'll want to use `@playwright/test` (Playwright Test), and not `playwright` (Playwright Library) directly.

## When Should Playwright Library Be Used Directly?

- creating an integration for a third party test runner (e.g. the third-party runner plugins listed [here](./test-runners.md) are built on top of Playwright Library)
- automation and scraping
```

rwoll:
Leaving open for now. @mxschmitt plans to do some work making the docs more test-use case focused in general, so I'll adapt this once I see some of those changes.
/cc @debs-obrien

mxschmitt (approved):
- double quotes -> single quotes
- put it inside the library-js.md - not a strong feeling on it, but I'd put it into the beginning of it.
- add a small note why web-first assertions are better, (auto-retry until the condition is met)

debs-obrien (commented):
Looks good. adding to top of library page as per Max´s sugguestion

docs/src/library-js.md
```diff
@@ -3,11 +3,136 @@ id: library
 title: "Library"
 ---
 
-Playwright can either be used as a part of the [Playwright Test](./intro.md), or as a Playwright Library (this guide). If you are working on an application that utilizes Playwright capabilities or you are using Playwright with another test runner, read on.
+Playwright Library provides unified APIs for launching and interacting with browsers, while Playwright Test provides all this plus a fully managed end-to-end Test Runner and experience.
+
+Under most circumstances, for end-to-end testing, you'll want to use `@playwright/test` (Playwright Test), and not `playwright` (Playwright Library) directly.
```
mxschmitt, line 8: Add a link to the intro here.

mxschmitt (approved):

docs/src/library-js.md
rwoll, line 8:
```suggestion
Under most circumstances, for end-to-end testing, you'll want to use `@playwright/test` (Playwright Test), and not `playwright` (Playwright Library) directly. To get started with Playwright Test, follow its [Getting Started Guide](./intro.md).
```

PR #15892, opened 2022-07-23
fix(chromium): Add error message to asset when there is one.

ghost:
[195 - 200] -- Prevent an error from being thrown when the user manually closes the sesstion. This is something which happens when you launch the browser with persistent context.

[202] -- If the object has an error message it should pass that on to asset so that the end user gets something useful instead of 'Assertion Error'. This line will add the error message to the assert call which will then repsent it to the user.

### DIFF
--- a/packages/playwright-core/src/server/chromium/crConnection.ts
+++ b/packages/playwright-core/src/server/chromium/crConnection.ts
@@ -193,7 +193,7 @@ export class CRSession extends EventEmitter {
     } else if (object.id && object.error?.code === -32001) {
       // Message to a closed session, just ignore it.
     } else {
-      assert(!object.id);
+      assert(!object.id, object?.error?.message || undefined);
       Promise.resolve().then(() => {
         if (this._eventListener)
           this._eventListener(object.method!, object.params);
--- a/packages/playwright-core/src/server/chromium/crConnection.ts
+++ b/packages/playwright-core/src/server/chromium/crConnection.ts
@@ -192,6 +192,12 @@ export class CRSession extends EventEmitter {
         callback.resolve(object.result);
     } else if (object.id && object.error?.code === -32001) {
       // Message to a closed session, just ignore it.
+    } else if (
+      !!object.error &&
+      !!object.error.message &&
+      /Session with given id not found/i.test(object.error.message)
+    ) {
+      // Ignore the error as the session was closed by the user
     } else {
       assert(!object.id, object?.error?.message || undefined);
       Promise.resolve().then(() => {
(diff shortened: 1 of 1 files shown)


dgozman:
@MakingStuffs Do you mind filing an issue that explains the problem? Otherwise, I am not sure what are we fixing. A repro that we could run would be ideal.

MakingStuffs:
Hi @dg

> @MakingStuffs Do you mind filing an issue that explains the problem? Otherwise, I am not sure what are we fixing. A repro that we could run would be ideal.
Hi mate, sorry I should have probably labelled it as a chore. The update is just ensuring two things: 

1. That the user closing the browser manually is ignored when in launched with persistent context, as with using the normal launch function
2. That any error message provided in to the if block is passed to assert as the current implementation of launchPersistentContext just gives a very unhelpful error message of `Assertion error` due to the fact that the message is not being passed, only the `!object.id` is.

MakingStuffs:
Here is the repo I forked: https://github.com/MakingStuffs/playwright

dgozman:
> That the user closing the browser manually is ignored when in launched with persistent context, as with using the normal launch function

For this part, we would love to see a repro script, to be aligned on the issue. I don't think that ignoring `Session with given id not found` errors is the right solution, but we will gladly help with the right fix if we know the problem exactly.

> That any error message provided in to the if block is passed to assert as the current implementation of launchPersistentContext just gives a very unhelpful error message of Assertion error due to the fact that the message is not being passed, only the !object.id is.

This is a good point! If you'd make a PR with just this change, I'll merge it right away.

MakingStuffs:
> > That the user closing the browser manually is ignored when in launched with persistent context, as with using the normal launch function
> 
> For this part, we would love to see a repro script, to be aligned on the issue. I don't think that ignoring `Session with given id not found` errors is the right solution, but we will gladly help with the right fix if we know the problem exactly.
> 
> > That any error message provided in to the if block is passed to assert as the current implementation of launchPersistentContext just gives a very unhelpful error message of Assertion error due to the fact that the message is not being passed, only the !object.id is.
> 
> This is a good point! If you'd make a PR with just this change, I'll merge it right away.

No probs I will do that today :) Should I close this one and start fresh?

dgozman:
> Should I close this one and start fresh?

That's up to you. If you repurpose this PR that would also be fine.

MakingStuffs:
> > Should I close this one and start fresh?
> 
> That's up to you. If you repurpose this PR that would also be fine.

Nice one mate I have removed the regex check for lost session ID and just kept the change passing the error message to assert :) thank you for your help

dgozman (approved):

PR #15904, opened 2022-07-24
chore(default context): Apply context workaround only when browser and scenario match

### LINKED ISSUE #15965: [BUG] Page reopen workaround for WebKit locale and Chromium isMobile bugs is applied even when not needed
**Context:**
- Playwright Version: 1.25.0
- Operating System: macOS 12.4
- Node.js version: 18.6.0
- Browser: Chromium-based browser

**Describe the bug**

The workaround to isMobile and locale that was introduced in https://github.com/microsoft/playwright/pull/2177 is applied to all instances, regardless of the browser type and options.

This uncommon behavior, re-opening the page, that is caused by the workaround, is causing some of our Chromium-based browser features to misbehave and breaks our end-to-end tests, that are based on Playwright, even though we do not fall into the criteria where the workaround should be fixed.

I uncommonly started off with a pull request (https://github.com/microsoft/playwright/pull/15904), suggesting a change that will not change the expected behavior for any Playwright user, but will fix the issue on our end. I opened this issue to continue the discussion as per @dgozman's request.

eran-rom:
The workaround to isMobile and locale that was introduced in https://github.com/microsoft/playwright/pull/2177 was applied to all instances, regardless of the browser type and options.

This uncommon behavior, re-opening the page, that is caused by the workaround, may cause unexpected behavior by applications that are leveraging Playwright for testings.

In this pull request I suggest narrowing the amount of instances that are affected by the workaround by only applying it when it is required, and let instances that doesn't match the criteria run naturally, as expected.

Fixes #15965.

### DIFF
--- a/packages/playwright-core/src/server/browserContext.ts
+++ b/packages/playwright-core/src/server/browserContext.ts
@@ -295,7 +295,8 @@ export abstract class BrowserContext extends SdkObject {
 
   async _loadDefaultContext(progress: Progress) {
     const pages = await this._loadDefaultContextAsIs(progress);
-    if (this._options.isMobile || this._options.locale) {
+    const browserName = this._browser.options.name;
+    if ((this._options.isMobile && browserName === 'chromium') || (this._options.locale && browserName === 'webkit')) {
       // Workaround for:
       // - chromium fails to change isMobile for existing page;
       // - webkit fails to change locale for existing page.
--- a/tests/library/har.spec.ts
+++ b/tests/library/har.spec.ts
@@ -79,7 +79,7 @@ it('should have pages', async ({ contextFactory, server }, testInfo) => {
   expect(pageEntry.pageTimings.onLoad).toBeGreaterThan(0);
 });
 
-it('should have pages in persistent context', async ({ launchPersistent }, testInfo) => {
+it('should have pages in persistent context', async ({ launchPersistent, browserName }, testInfo) => {
   const harPath = testInfo.outputPath('test.har');
   const { context, page } = await launchPersistent({ recordHar: { path: harPath } });
   await page.goto('data:text/html,<title>Hello</title>');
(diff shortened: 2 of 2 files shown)


packages/playwright-core/src/server/browserContext.ts
```diff
@@ -295,7 +295,8 @@ export abstract class BrowserContext extends SdkObject {
 
   async _loadDefaultContext(progress: Progress) {
     const pages = await this._loadDefaultContextAsIs(progress);
-    if (this._options.isMobile || this._options.locale) {
+    const browserName = this._browser.options.name;
```
thaipham96, line 298: gh pr checkout 15904

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -295,7 +295,8 @@ export abstract class BrowserContext extends SdkObject {
...
+    if ((this._options.isMobile && browserName === 'chromium') || (this._options.locale && browserName === 'webkit')) {
```
thaipham96, line 299: f1

thaipham96 (approved):
gh pr checkout 

> 15904`
`

dgozman:
@eranr-talon Could you please provide more details? Also, we usually start with an issue, and then link the PR to the issue. This way it's easier to get the conversation going.

> This uncommon behavior, re-opening the page, that is caused by the workaround, may cause unexpected behavior by applications that are leveraging Playwright for testings.

What would be an example where this is unexpected? Do you have a repro script that suffers from this behavior?

eranr-talon:
> @eranr-talon Could you please provide more details? Also, we usually start with an issue, and then link the PR to the issue. This way it's easier to get the conversation going.
> 

Understood. Would you like me to open an issue now so we could continue the conversation over there?

> > This uncommon behavior, re-opening the page, that is caused by the workaround, may cause unexpected behavior by applications that are leveraging Playwright for testings.
> 
> What would be an example where this is unexpected? Do you have a repro script that suffers from this behavior?

We're using Playwright to end-to-end test a Chromium based browser. One of its' features includes postponing opening of new pages, so when calling `oldPage.close()` in `_loadDefaultContext()`, eventually the driver tries to close a postponed page and fails. The unexpected behavior I was referring to was that first page rotation.

Eventually, this PR does not change any functionality for every other Playwright user, and helps us avoid the workaround behavior as we do not fall into the criteria where we need the first page to be reopened, which breaks some of our end-to-end tests.

dgozman:
> Would you like me to open an issue now so we could continue the conversation over there?

That would be awesome!

> Eventually, this PR does not change any functionality for every other Playwright user, it just fine tunes the hack and helps us avoid the workaround behavior, as we do not fall into the criteria where we need the first page to be reopened, which breaks some of our end-to-end tests.

I see. So, it seems like you are using `locale` emulation in chromium, because that's the only case this PR affects? You can workaround the issue for now by changing locale through command line `args` when calling `launchPersistentContext()`. Take a look at [this stack overflow question](https://stackoverflow.com/questions/24992240/start-google-chrome-with-a-specific-locale-using-a-command-line-argument) for example.

dgozman (approved):

dgozman:
Thank you for the PR!

eranr-talon:
> Thank you for the PR!

My pleasure!

PR #15922, opened 2022-07-25
feat(webkit): roll to r1687

playwrightmachine:


mxschmitt:
folding into https://github.com/microsoft/playwright/pull/15943

PR #15923, opened 2022-07-25
docs: fix .net code in js and add link

debs-obrien:


docs/src/troubleshooting.md
```diff
@@ -81,5 +81,5 @@ LTS releases.
 
 ## Please file an issue
 
-Playwright is a new project, and we are watching the issues very closely. As we solve common issues, this document will grow to include the common answers.
+Playwright is a new project, and we are watching the [issues](https://github.com/microsoft/playwright/issues) very closely. As we solve common issues, this document will grow to include the common answers.
```
mxschmitt, line 84: the best would be if it's a dynamic link. What does it mean? Either repeat "Please file an issue" 4 times or create a new link [here](https://github.com/microsoft/playwright.dev/tree/main/common) in each file, which you then can reference in the troubleshooting.md.

mxschmitt (approved):

PR #15933, opened 2022-07-25
fix: accomodate to changed MSEdge download format

### LINKED ISSUE #15932: [BUG] - Install MS Edge on CI Fails
On CI machines (MacOS and Windows) running:
`yarn dlx playwright@1.23.2 install chrome webkit msedge --with-deps`

Fails with an error installing Edge. We see linux passing this step just fine.

MacOS Error:
image

Windows Error:
image

I believe the script is failing here: https://github.com/microsoft/playwright/blob/c11f594c2921f98c4e72e47f9e819d2a40787b5e/packages/playwright-core/src/server/registry/index.ts#L760

Not sure why this is happening all of a sudden but we started seeing it in our CI ~5 hours ago and it is still happening.

aslushnikov:
Fixes #15932

### DIFF
--- a/packages/playwright-core/src/server/registry/index.ts
+++ b/packages/playwright-core/src/server/registry/index.ts
@@ -910,15 +910,12 @@ export function findChromiumChannel(sdkLanguage: string): string | undefined {
 }
 
 function lowercaseAllKeys(json: any): any {
-  if (typeof json !== 'object')
+  if (typeof json !== 'object' || !json)
     return json;
 
-  if (Array.isArray(json)) {
-    const result = [];
-    for (const entry of json)
-      result.push(lowercaseAllKeys(entry));
-    return result;
-  }
+  if (Array.isArray(json))
+    return json.map(lowercaseAllKeys);
+
   const result: any = {};
   for (const [key, value] of Object.entries(json))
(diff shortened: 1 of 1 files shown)


pavelfeldman (approved):

packages/playwright-core/src/server/registry/index.ts
```diff
@@ -908,4 +909,20 @@ export function findChromiumChannel(sdkLanguage: string): string | undefined {
   return channel;
 }
 
+function lowercaseAllKeys(json: any): any {
+  if (typeof json !== 'object')
```
dgozman, line 913: To accomodate null:

```suggestion
  if (typeof json !== 'object' || !json)
```

packages/playwright-core/src/server/registry/index.ts
```diff
@@ -908,4 +909,20 @@ export function findChromiumChannel(sdkLanguage: string): string | undefined {
...
+    return json;
+
+  if (Array.isArray(json)) {
```
dgozman, line 916:
```suggestion
  if (Array.isArray(json))
    return json.map(lowercaseAllKeys);
```

dgozman (approved):

charlesanim:
When can we expect a package update? @aslushnikov

PR #15939, opened 2022-07-26
fix(runner): friendly error message instead of "digests do not match"

### LINKED ISSUE #14504: [BUG] Error: Digests do not match
**Context:**
- Playwright Version: 1.21.1
- Operating System: Mac
- Node.js version: 18.2.0
- Browser: Chromium

**Describe the bug**

I've got this error in one of my test case when running multiple test suites with multiple workers. 
"Error: Digests do not match"

What could be the problem here? I'm using playwright for couple of months and this is the first time encountering this issue after setting up some custom fixtures. 

**Some Observations**
- It isn't related with the multiple workers. I've also tried with 1 worker and still got the error. 
- It is however connected with worker count somehow since the error is thrown in different suites with changing worker count.
- It is most probably related to custom fixtures.
- Error is thrown when starting the first case of the test suite and all other cases in the suite is skipped. 
- Also tried in 1.22.2 and got the same error

dgozman:
Fixes #14504.

### DIFF
--- a/packages/playwright-test/src/fixtures.ts
+++ b/packages/playwright-test/src/fixtures.ts
@@ -300,8 +300,14 @@ export class FixtureRunner {
   setPool(pool: FixturePool) {
     if (!this.testScopeClean)
       throw new Error('Did not teardown test scope');
-    if (this.pool && pool.digest !== this.pool.digest)
-      throw new Error('Digests do not match');
+    if (this.pool && pool.digest !== this.pool.digest) {
+      throw new Error([
+        `Playwright detected inconsistent test.use() options.`,
+        `Most common mistakes that lead to this issue:`,
+        `  - Calling test.use() outside of the test file, for example in a common helper.`,
+        `  - One test file imports from another test file.`,
+      ].join('\n'));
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/fixtures.ts
```diff
@@ -300,8 +300,14 @@ export class FixtureRunner {
...
     if (!this.testScopeClean)
       throw new Error('Did not teardown test scope');
-    if (this.pool && pool.digest !== this.pool.digest)
-      throw new Error('Digests do not match');
+    if (this.pool && pool.digest !== this.pool.digest) {
+      throw new Error([
+        `Playwright detected inconsistent test.use() options.`,
+        `Most common mistakes that lead to this issue:`,
+        `  - Calling test.use() outside of the test file, for example in a common helper.`,
+        `  - One test file imports from another test file.`,
```
yury-s, line 308: Would it make sense to add docs section on this and have a link from the message?

yury-s (approved):

mxschmitt (approved):

PR #15966, opened 2022-07-26
devops: another fix for Docker build script

aslushnikov:
* git configuration should be system wide to propagate to all
  users
* rust should be installed under the `pwuser`


aslushnikov:
TBR @mxschmitt

Issue #15919, opened 2022-07-25
[Feature] component testing mount result


sand4rt:
Can the `update ` and the `unmount` functions be added? This is useful for Vue, React and Svelte:

```typescript
test('update props', ({ mount }) => {
  const component = await mount(Component, {
    props: { name: 'test' }
  });
  await component.update({ props: { name: 'test2' }});
  await expect(component).toHaveText('test2');
})
```

```typescript
test('display a pending changes conformation', ({ mount }) => {
  const component = await mount(Component);
  await component.unmount();
  await expect(component.locator('#conformation-dialog')).toBeVisible();
})
```

## TODO
- [x] update Svelte
- [ ] update Solid
- [x] update React
- [x] update Vue
- [x] update Vue2
- [x] unmount Svelte
- [x] unmount Solid
- [x] unmount React
- [x] unmount Vue
- [x] unmount Vue2

pavelfeldman:
Thanks for another useful use case. I meant to ask a while ago, would you be interested in contributing patches for the components story?

pavelfeldman:
> I'm somewhat limited in time, but I'd love to! How would you like to proceed?

I'm adding unmount in the PR above, you can follow the pattern and implement `setProps`. I'm not sure it will work for all the frameworks, but present fixtures API allows framework-specific signatures, so I'm sure something useful can be done for each.

sand4rt:
Awesome😄 I will have a look at ’setProps’!

Wondering why you would choose for '({ unmount }) => {}' and 'unmount(compoment)' instead of just calling 'component.unmount()'? The last option is easier to write and read right?

pavelfeldman:
> Wondering why you would choose for ({ unmount }) => {} and unmount(compoment) instead of just calling component.unmount()? The last option is easier to write and read.

There is no `Locator.unmount`, so we would need a separate type for it. Let's see how many methods we will need and if we get a bunch, we can transition.

sand4rt:
Have you taken into account that the `setProps` will not automatically be type safe as well if i follow the current implementation of `unmount`?

```typescript
test('work', ({ mount, setProps }) => {
  const component = mount<Props>({ 
    props: { title: 'test' }
  })
  
  setProps<Props>(component, { title: 'test' }) // Must specifically specify the type here
})
```

instead of

```typescript
test('work', ({ mount }) => {
  const component = mount<Props>({ 
    props: { title: 'test' }
  })
  
[...]

PR #15974, opened 2022-07-26
feat(ct): allow unmounting components

pavelfeldman:
Partial fix for: https://github.com/microsoft/playwright/issues/15919

### DIFF
--- a/packages/playwright-ct-svelte/registerSource.mjs
+++ b/packages/playwright-ct-svelte/registerSource.mjs
@@ -20,6 +20,7 @@
 
 /** @typedef {import('../playwright-test/types/component').Component} Component */
 /** @typedef {any} FrameworkComponent */
+/** @typedef {import('svelte').SvelteComponent} SvelteComponent */
 
 /** @type {Map<string, FrameworkComponent>} */
 const registry = new Map();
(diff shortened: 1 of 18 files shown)


packages/playwright-ct-svelte/registerSource.mjs
```diff
@@ -54,14 +55,22 @@ window.playwrightMount = async (component, rootElement, hooksConfig) => {
...
+    await hook({ hooksConfig, svelteComponent });
 
   for (const [key, listener] of Object.entries(component.options?.on || {}))
-    wrapper.$on(key, event => listener(event.detail));
+    svelteComponent.$on(key, event => listener(event.detail));
+};
+
+window.playwrightUnmount = async (element, rootElement) => {
+  const svelteComponent = /** @type {SvelteComponent} */ (rootElement[svelteComponentKey]);
+  svelteComponent.$destroy();
```
dgozman, line 73: Should we throw a meaningful error when I call `unmount(someRandomLocator)` instead of passing the component that I mounted before?

dgozman (approved):

tests/components/ct-react-vite/src/App.spec.tsx
```diff
@@ -18,3 +18,10 @@ test('should configure app', async ({ page, mount }) => {
   });
   expect(messages).toEqual(['Before mount: {\"route\":\"A\"}', 'After mount']);
 });
+
+test('should unmount', async ({ page, mount, unmount }) => {
+  const component = await mount(<App></App>);
+  await expect(page.locator('#root')).toContainText('Hello Vite + React!');
+  unmount(component);
```
Meir017, line 25: should this be awaited? the `await expect(page.locator('#root'))` implicitly waits for this. I think this can be confusing

PR #15975, opened 2022-07-26
feat(test runner): show the number of fatal errors at the end

dgozman:
Example output from the `dot` reporter:

```bash
Running 8 tests using 1 worker
···
Worker teardown timeout of 2000ms exceeded while tearing down "worker".

Failed worker ran 4 tests:
output/a.spec.ts:11:1 › test1
output/a.spec.ts:17:3 › good0
output/a.spec.ts:17:3 › good1
output/a.spec.ts:22:1 › test2
·····

  1 fatal error
  8 passed (3s)
```

### DIFF
--- a/packages/playwright-test/src/reporters/base.ts
+++ b/packages/playwright-test/src/reporters/base.ts
@@ -42,6 +42,7 @@ type TestSummary = {
   unexpected: TestCase[];
   flaky: TestCase[];
   failuresToPrint: TestCase[];
+  fatalErrors: TestError[];
 };
 
 export class BaseReporter implements Reporter  {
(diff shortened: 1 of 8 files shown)


yury-s (approved):

packages/playwright-test/src/reporters/base.ts
```diff
@@ -96,7 +98,7 @@ export class BaseReporter implements Reporter  {
   }
 
   onError(error: TestError) {
-    console.log('\n' + formatError(this.config, error, colors.enabled).message);
+    this._fatalErrors.push(error);
```
pavelfeldman, line 101: This can be perceived as a regression by some!

rwoll:
@dgozman when does this come up during usage?

packages/playwright-test/src/reporters/base.ts
dgozman: This code has moved to dot, line and list to not interfere with their internal line counting details.

dgozman:
> @dgozman when does this come up during usage?

On our bots you can see `Worker teardown timeout during "browser" teardown` and similar errors that are not reflected at the summary.

PR #15976, opened 2022-07-27
devops: fix WebKit Linux arm64 build

aslushnikov:


aslushnikov:
TBR @yury-s @mxschmitt

PR #16003, opened 2022-07-27
feat(testId): expose persistent test id

### LINKED ISSUE #14310: [Question] How can I get value for "testId" query-parametr for each test in HTML-report
### Problem:

We push a message into the slack that contains a link to the playwright HTML-report. My colleagues asked that this message contain a link not only to the full report, but also to each test.

Example of a link to one of the tests:
`https://test.com/index.html#?testId=ec870af7b8935a30dd92c181a150c5a3d014e16b-0671b0d0053685bb8fe9ac82667f6d117cb6faf3`

I read the documentation, and also looked at the structure of the playwright JSON-report and didn't find that testId was specified for each test in it. How can I get testId query for each test in the suite?

_____
P.S. Thank you for such a wonderful tool! <3

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/14310

### DIFF
--- a/packages/playwright-test/src/loader.ts
+++ b/packages/playwright-test/src/loader.ts
@@ -152,12 +152,28 @@ export class Loader {
     }
     this._fullConfig.metadata = takeFirst(config.metadata, baseFullConfig.metadata);
     this._fullConfig.projects = (config.projects || [config]).map(p => this._resolveProject(config, this._fullConfig, p, throwawayArtifactsPath));
+    this._assignUniqueProjectIds(this._fullConfig.projects);
+  }
+
+  private _assignUniqueProjectIds(projects: FullProjectInternal[]) {
+    const usedNames = new Set();
+    for (const p of projects) {
+      const name = p.name || '';
+      for (let i = 0; i < projects.length; ++i) {
+        const candidate = name + (i ? i : '');
+        if (usedNames.has(candidate))
+          continue;
+        p._id = candidate;
+        usedNames.add(candidate);
+        break;
+      }
+    }
   }
 
   async loadTestFile(file: string, environment: 'runner' | 'worker') {
     if (cachedFileSuites.has(file))
       return cachedFileSuites.get(file)!;
-    const suite = new Suite(path.relative(this._fullConfig.rootDir, file) || path.basename(file));
+    const suite = new Suite(path.relative(this._fullConfig.rootDir, file) || path.basename(file), 'file');
     suite._requireFile = file;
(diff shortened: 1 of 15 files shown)


packages/playwright-test/src/loader.ts
```diff
@@ -363,19 +379,21 @@ class ProjectSuiteBuilder {
         }
       } else {
         const test = entry._clone();
+        to._addTest(test);
         test.retries = this._project.retries;
-        // We rely upon relative paths being unique.
-        // See `getClashingTestsPerSuite()` in `runner.ts`.
-        test._id = `${calculateSha1(relativeTitlePath + ' ' + entry.title)}@${entry._requireFile}#run${this._index}-repeat${repeatEachIndex}`;
+        const repeatEachIndexSuffix = repeatEachIndex ? ` (repeat:${repeatEachIndex})` : '';
```
dgozman, line 384: You can remove `relativeTitlePath` argument now.

packages/playwright-test/src/loader.ts
```diff
@@ -363,19 +379,21 @@ class ProjectSuiteBuilder {
         }
       } else {
         const test = entry._clone();
+        to._addTest(test);
         test.retries = this._project.retries;
-        // We rely upon relative paths being unique.
```
dgozman, line 367: This comment is still relevant. Also, to be exactly correct, this code and `getClashingTestsPerSuite` must use the same delimiter (space vs ">").

dgozman (approved):

packages/playwright-test/src/loader.ts
```diff
@@ -152,12 +152,28 @@ export class Loader {
...
+      for (let i = 0; i < projects.length; ++i) {
+        const candidate = name + (i ? '#' + i : '');
+        if (usedNames.has(candidate))
+          continue;
+        p._id = candidate;
+        usedNames.add(candidate);
+        break;
+      }
+    }
   }
```
rwoll, line 171: I don't love how sensitive these IDs are. If the user wants to rename a project (or a test…or nest in a describe, etc.)—maybe fix a typo—all the IDs change, so the notion these are historical IDs are pretty fragile.

I think this is fine for now, but I suggest being more explicit in the docs. If requested, we could always add an ID generation function that can be customized (or allow more explicitly setting IDs for projects, tests, etc.).

pavelfeldman: Yeah, we can always add a project.historyID!

PR #16012, opened 2022-07-28
chore(test-runner): better error message if page gets used inside beforeAll

### LINKED ISSUE #15798: [Feature] Change wording in error output for page fixture in beforeall to browser fixture
When I run `beforeAll(async ({page}) => {}` I get the message 

> Error: "context" and "page" fixtures are not supported in beforeAll. Use browser.newContext() instead.

This is a bit misleading. Most users, including myself, will be using a page fixture in beforeAll if they are trying to reuse a page. This had me blocked for 2 days trying to get the same behavior as I have in Mocha. I was going down the rabbit hole trying to make browser initialize and global in globalSetup. I also was under the impression if `{page}` was not avail in beforeAll then all fixtures were not avail. 

Instead, the error output should be something like :

> Error: "context" and "page" fixtures are not supported in beforeAll. Use beforeAll({browser}) to reuse a page. Please see https://playwright.dev/docs/test-retries#reuse-single-page-between-tests

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/15798

### DIFF
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -463,8 +463,13 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
 
     await use(async options => {
       const hook = hookType(testInfo);
-      if (hook)
-        throw new Error(`"context" and "page" fixtures are not supported in ${hook}. Use browser.newContext() instead.`);
+      if (hook) {
+        throw new Error([
+          `"context" and "page" fixtures are not supported in "${hook}" since they are created on a per-test basis (test fixtures).`,
+          `Either use "${hook.replace('All', 'Each')}" or create a new context/page instance inside "${hook}" by using the "browser" fixture (browser.newContext() and context.newPage())`,
+          'Please see https://playwright.dev/docs/test-retries#reuse-single-page-between-tests for more information.',
+        ].join('\n'));
+      }
       const videoOptions: BrowserContextOptions = captureVideo ? {
         recordVideo: {
           dir: _artifactsDir(),
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -466,7 +466,7 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
       if (hook) {
         throw new Error([
           `"context" and "page" fixtures are not supported in "${hook}" since they are created on a per-test basis.`,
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/index.ts
```diff
@@ -463,8 +463,13 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
 
     await use(async options => {
       const hook = hookType(testInfo);
-      if (hook)
-        throw new Error(`"context" and "page" fixtures are not supported in ${hook}. Use browser.newContext() instead.`);
+      if (hook) {
+        throw new Error([
+          `"context" and "page" fixtures are not supported in "${hook}" since they are created on a per-test basis (test fixtures).`,
```
dgozman, line 468: Should we do something like this? WDYT?

"context" and "page" fixtures are not supported in "${hook}" since they are created on a per-test basis.
- If you would like to reuse a single page between tests, create context manually with browser.newContext(). See https://playwright.dev/docs/test-retries#reuse-single-page-between-tests for details.
- If you would like to configure your page before each test, do that in beforeEach hook instead.

packages/playwright-test/src/index.ts
```diff
@@ -463,8 +463,13 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
 
     await use(async options => {
       const hook = hookType(testInfo);
-      if (hook)
-        throw new Error(`"context" and "page" fixtures are not supported in ${hook}. Use browser.newContext() instead.`);
+      if (hook) {
+        throw new Error([
+          `"context" and "page" fixtures are not supported in "${hook}" since they are created on a per-test basis.`,
+          `If you would like to reuse a single page between tests, create context manually with browser.newContext(). See https://playwright.dev/docs/test-retries#reuse-single-page-between-tests for details.`,
+          `If you would like to configure your page before each test, do that in beforeEach hook instead.`,
```
rwoll, line 470: I like the idea of linking to the docs. It should be noted, it's our first time doing it. I don't recall why, but last time I had a link in an PW Error message, we decided we didn't want the link.

dgozman (approved):

packages/playwright-test/src/index.ts
mxschmitt: We probably should use an aka.ms link instead and figure out a naming schema.
e.g.  https://aka.ms/playwright/reusable-context

PR #16022, opened 2022-07-28
devops: fix Firefox build on Linux

aslushnikov:
Prebuilt WASI sysroot is not availabe any more, so this patch
disables sandboxing via WASI for now in our builds.

I tried compiling WASI manually, but it didn't work out for me
for some reason.


aslushnikov:
TBR @mxschmitt

PR #16026, opened 2022-07-28
fix(fixtures): make undefined option mean "default value"

### LINKED ISSUE #14989: [BUG] Setting the use.trace option to `undefined` causes TypeError: Cannot read property 'mode' of undefined

dgozman:
In the following example, `locale` inside the `describe` would be reverted to the default value:

```js
test.use({ locale: 'en-GB' });

test.decsribe(() => {
  test.use({ locale: undefined });
});
```

Fixes #14989.

### DIFF
--- a/packages/playwright-test/src/fixtures.ts
+++ b/packages/playwright-test/src/fixtures.ts
@@ -203,7 +203,7 @@ export class FixturePool {
           };
           value = value[0];
         }
-        const fn = value as (Function | any);
+        let fn = value as (Function | any);
 
         const previous = this.registrations.get(name);
         if (previous && options) {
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/fixtures.ts
```diff
@@ -222,6 +222,15 @@ export class FixturePool {
         if (options.scope === 'worker' && disallowWorkerFixtures)
           throw errorWithLocations(`Cannot use({ ${name} }) in a describe group, because it forces a new worker.\nMake it top-level in the test file or put in the configuration file.`, { location, name });
 
+        // Overriding option with "undefined" value means setting it to the default value
+        // from the original declaration of the option.
+        if (fn === undefined && options.option && previous) {
```
yury-s, line 227: `fn` is a bit weird in this context, consider a more descriptive name `valueOrFn` maybe?

yury-s (approved):

PR #16028, opened 2022-07-28
docs(selectors): avoid nth, first, last

rwoll:
An attempt to add some more best practices, and make them concrete through examples.

### DIFF
--- a/docs/src/selectors.md
+++ b/docs/src/selectors.md
@@ -1323,7 +1323,7 @@ await page.Locator("data-test-id=directions").ClickAsync();
 ### Avoid selectors tied to implementation
 
 [xpath] and [css] can be tied to the DOM structure or implementation. These selectors can break when
-the DOM structure changes.
+the DOM structure changes. Similarly, [`method: Locator.nth`], [`method: Locator.first`], and [`method: Locator.last`] are tied to implementation and the structure of the DOM, and will target the incorrect element if the DOM changes.
 
 ```js
 // avoid long css or xpath chains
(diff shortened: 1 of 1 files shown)


docs/src/selectors.md
```diff
@@ -1355,6 +1355,48 @@ await page.Locator("#tsf > div:nth-child(2) > div.A8SBwf > div.RNNXgb > div > di
...
+await page.locator("xpath=/html/body/div/main/form/label/input").fill("jordan@example.com"); // 👎 works today, BUT breaks in the future with any small DOM change, hard to read (i.e. what input is this locator trying to target?) 
+await page.locator("form >> text=Email").fill("jordan@example.com"); // 👌 OK: readable and captures how a user will know what to click and fill. Specific enough to avoid strict violation.
+```
+
+```js tab=js-ts
+await page.locator("input").fill("jordan@example.com"); // 👎 locator.fill: Error: strict mode violation: "input" resolved to 2 elements:
+await page.locator("input").first().fill("jordan@example.com"); // 👎 works today, BUT easily breaks in the future (e.g. an input to capture name, or a search bar is added at the top of the page)
+await page.locator("xpath=/html/body/div/main/form/label/input").fill("jordan@example.com"); // 👎 works today, BUT breaks in the future with any small DOM change, hard to read (i.e. what input is this locator trying to target?) 
+await page.locator("form >> text=Email").fill("jordan@example.com"); // 👌 OK: readable and captures how a user will know what to click and fill. Specific enough to avoid strict violation.
```
rwoll, line 1395: This works, but I'm not sure I like this selector. It's better than the others, though!

docs/src/selectors.md
```diff
@@ -1355,6 +1355,48 @@ await page.Locator("#tsf > div:nth-child(2) > div.A8SBwf > div.RNNXgb > div > di
...
+    Additional Comments
+    <input name="comments" type="text" />
+  </label>
+  <button id="submit" type="submit">Submit</button>
+</form>
+```
+
+Let's look at a few possible attempts to fill in the Email address:
+
+```js tab=js-js
```
rwoll, line 1384: Other langs will be added if we like the content of this change!

docs/src/selectors.md
```diff
@@ -1355,6 +1355,48 @@ await page.Locator("#tsf > div:nth-child(2) > div.A8SBwf > div.RNNXgb > div > di
...
+await page.locator("input").fill("jordan@example.com"); // 👎 locator.fill: Error: strict mode violation: "input" resolved to 2 elements:
+await page.locator("input").first().fill("jordan@example.com"); // 👎 works today, BUT easily breaks in the future (e.g. an input to capture name, or a search bar is added at the top of the page)
+await page.locator("xpath=/html/body/div/main/form/label/input").fill("jordan@example.com"); // 👎 works today, BUT breaks in the future with any small DOM change, hard to read (i.e. what input is this locator trying to target?) 
+await page.locator("form >> text=Email").fill("jordan@example.com"); // 👌 OK: readable and captures how a user will know what to click and fill. Specific enough to avoid strict violation.
```
Meir017, line 1388: isn't selecting the input element better using an attribute selector? for sites supporting i18n the label text can change as well.
I think
```js
await page.locator("form >> input[name=email]").fill("jordan@example.com");
```
would be better

docs/src/selectors.md
Meir017: isn't selecting the input element better using an attribute selector? for sites supporting i18n the label text can change as well.
I think
```js
await page.locator("form >> input[name=email]").fill("jordan@example.com");
```
would be better

docs/src/selectors.md
```diff
@@ -1355,6 +1355,48 @@ await page.Locator("#tsf > div:nth-child(2) > div.A8SBwf > div.RNNXgb > div > di
 await page.Locator("//*[@id='tsf']/div[2]/div[1]/div[1]/div/div[2]/input").ClickAsync();
 ```
 
+### Example
+
+The best selector to use is dependent on the context and the app under test. Here's some questions to keep in mind when authoring selectors:
```
pavelfeldman, line 1360: Looks like you are pulling the best practices doc into the example section. As in we had nice prioritized talking points and all of a sudden Example decided to take over with its own wisdom. Let's keep points structure as is, we can add more points. And we can provide example with dos and donts for each if you feel like example is helpful!

pavelfeldman (approved):

PR #16040, opened 2022-07-29
feat: print response text when toBeOK fails

### LINKED ISSUE #15950: [Feature] toBeOK should print the response if it is short enough
I have code like this, I think it is quite typical:
```js
const response = await request.post('/accounts/', { data: { … … … })
await expect(response).toBeOK()
```
So, I'm just making a request and checking if it's OK (if the status code is in 200..299 range).

On failure, Playwright will print something like this:
```js
    Error: expect(received).toBeOK()
    Call log:
      - → POST http://localhost:8042/accounts/
      -   user-agent: Playwright/1.23.2 (x64; debian unknown) node/16.14
      -   accept: */*
      -   accept-encoding: gzip,deflate,br
      -   authorization: Bearer ……………………………………………………………………………………………….……………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………………….…………………………………………………………………………………………………………………
      -   content-type: application/json
      -   content-length: 124
      - ← 400 Bad Request
      -   connection: close
      -   content-type: application/json
      -   vary: Accept, Origin
      -   allow: GET, POST, HEAD, OPTIONS
      -   x-djangoquerycount-count: 9
      -   content-length: 48
      -   x-content-type-options: nosniff
      -   referrer-policy: same-origin
```
That's a lot of data, but none of it is actually useful. Usually, on 400, the backend returns a more detailed reason in the response body. The actual reason for failure can be shown by adding console.log like this:
```js
const response = await request.post('/accounts/', { data: { … … … })
console.log(await response.json())
await expect(response).toBeOK()
```
Which will print:
```js
{ account: [ 'You already created this account' ] }
```
Ah, OK, now it's clear why the backend is unhappy.

## Proposal
I think `.toBeOK` should print the response body if it is not too large (it has no issues printing a huge access token in headers, so relatively large bodies should not be a problem too I guess).

yury-s:
Fixes #15950

### DIFF
--- a/packages/playwright-test/src/matchers/matchers.ts
+++ b/packages/playwright-test/src/matchers/matchers.ts
@@ -290,13 +290,18 @@ export async function toBeOK(
 ) {
   const matcherName = 'toBeOK';
   expectTypes(response, ['APIResponse'], matcherName);
+
+  const contentType = response.headers()['content-type'];
+  const isTextEncoding = contentType && /^text\/|^application\/.*json/.test(contentType);
   const [log, text] = (this.isNot === response.ok()) ? await Promise.all([
     response._fetchLog(),
-    response.text()
+    isTextEncoding ? response.text() : null
   ]) : [];
+
   const message = () => this.utils.matcherHint(matcherName, undefined, '', { isNot: this.isNot }) +
     callLogText(log) +
-    `\nResponse text:\n${colors.dim(text?.substring(0, 1000) || '')}`;
+    (text === null ? '' : `\nResponse text:\n${colors.dim(text?.substring(0, 1000) || '')}`);
+
   const pass = response.ok();
   return { message, pass };
 }
--- a/packages/playwright-test/src/matchers/matchers.ts
+++ b/packages/playwright-test/src/matchers/matchers.ts
@@ -24,6 +24,7 @@ import { toBeTruthy } from './toBeTruthy';
 import { toEqual } from './toEqual';
 import { toExpectedTextValues, toMatchText } from './toMatchText';
 import type { ParsedStackTrace } from 'playwright-core/lib/utils/stackTrace';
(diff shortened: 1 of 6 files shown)


packages/playwright-test/src/matchers/matchers.ts
```diff
@@ -289,8 +290,13 @@ export async function toBeOK(
...
   expectTypes(response, ['APIResponse'], matcherName);
-  const log = (this.isNot === response.ok()) ? await response._fetchLog() : [];
-  const message = () => this.utils.matcherHint(matcherName, undefined, '', { isNot: this.isNot }) + callLogText(log);
+  const [log, text] = (this.isNot === response.ok()) ? await Promise.all([
+    response._fetchLog(),
+    response.text()
+  ]) : [];
+  const message = () => this.utils.matcherHint(matcherName, undefined, '', { isNot: this.isNot }) +
+    callLogText(log) +
+    `\nResponse text:\n${colors.dim(text?.substring(0, 1000) || '')}`;
```
dgozman, line 299: - Should we only print response if it has `textualMimeType` (we have a helper like that somewhere)? Otherwise, it's probably not even safe to call `response.text()` as it might not be utf8.
- 1000 seems to be a long string, but maybe not 😄

yury-s: > * Should we only print response if it has `textualMimeType` (we have a helper like that somewhere)? Otherwise, it's probably not even safe to call `response.text()` as it might not be utf8.

Agreed, but what if there is no content-type? It's probably good enough to print it if there is no content type or text content-type but that's not ideal either.

> * 1000 seems to be a long string, but maybe not smile

What number do you suggest instead?

dgozman (approved):

dgozman (approved):

PR #16060, opened 2022-07-29
feat: return value from step

### LINKED ISSUE #15614: [Feature] Return value from test step
It would be useful to be able to return a value from a test step and use it in a subsequent step.
This way we could avoid mutable state (`let`).

```typescript
    const invitationLink = await test.step('Create invitation link', async () => {
      await page.goto('create-invitation-link');
      …
      return await page.locator('#link').getAttribute('href');
    });

    await test.step('Accept invitation', async () => {
      await page.goto(invitationLink);
    });
```

The `test.step` method would probably need a generic signature:

```typescript
step<A>(title: string, body: () => Promise<A>): Promise<A>;
```

yury-s:
Fixes #15614

### DIFF
--- a/tests/playwright-test/test-step.spec.ts
+++ b/tests/playwright-test/test-step.spec.ts
@@ -379,3 +379,24 @@ test('should report custom expect steps', async ({ runInlineTest }) => {
     },
   ]);
 });
+
+test('should return value from step', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.test.ts': `
+      const { test } = pwt;
+      test('timeout', async ({ page }) => {
+        const v1 = await test.step('my step', () => {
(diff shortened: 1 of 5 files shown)


tests/playwright-test/test-step.spec.ts
```diff
@@ -379,3 +379,26 @@ test('should report custom expect steps', async ({ runInlineTest }) => {
     },
   ]);
 });
+
+test('should return value from step', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
```
rwoll, line 384: Can you add a TSC test to ensure the typing all checks out?

rwoll (approved):
Once a TSC test is added to check the typing is all wired properly, LGTM.

Issue #15163, opened 2022-06-28
[BUG] `browser.close` doesn't close contexts properly


DetachHead:
**Context:**
## System:
 - OS: Windows 10 10.0.19044
 - Memory: 1.41 GB / 15.78 GB
## Binaries:
 - Node: 18.0.0 - C:\Program Files\nodejs\node.EXE
 - npm: 8.7.0 - C:\Program Files\nodejs\npm.CMD
## Languages:
 - Bash: 5.0.17 - C:\WINDOWS\system32\bash.EXE
## npmPackages:
 - playwright: ^1.23.0 => 1.23.0

**Code Snippet**

```javascript
const { chromium } = require('playwright')

;(async () => {
    const browser = await chromium.launch()
    const context = await browser.newContext({
        recordHar: { path: 'asdf.har' },
    })
    const page = await context.newPage()
    await page.goto('https://google.com/')
    await browser.close()
})()
```

**Describe the bug**

when closing the browser with `browser.close()` instead of `context.close()`, it does not close the context properly. in this case, no HAR file is created.

i would expect `browser.close()` to also call `context.close()` on all of its contexts before closing the browser

rwoll:
Thanks for the report! This is expected behavior. You should close all contexts you create before calling `browser.close()`. Playwright doesn't want `browser.close()` to hang/not work, so it can't wait on too much and instead requires explicit `context.close()` to happen before it.

KotlinIsland:
That's absolute insanity

DetachHead:
@rwoll i think in that case it should at least output a warning when you call `browser.close` without closing all the contexts first

rwoll:
I feel your pain! Is there somewhere in the docs you think seeing this information would be useful? The warning/log line is a good idea; I'll raise with the team, but there's not a precedent of outputting logs in Playwright (other than when using the DEBUG env/mode).

It is indeed slightly extra bookkeeping if you are using Playwright directly as a Library. However, we expect most users are using Playwright Test and are therefore not managing contexts directly. (The Test Runner handles their creation _and_ closing.)

KotlinIsland:
What about users of playwright-<other language>, that has no support for playwright test at all?

rwoll:
> It is indeed slightly extra bookkeeping if you are using Playwright directly as a Library. However, we expect most users are using Playwright Test and are therefore not managing contexts directly. (The Test Runner handles their creation and closing.)

[...]

PR #16073, opened 2022-07-30
feat(browser): close contexts when close browser

regseb:
With this pull request, when the [`browser.close()`](https://playwright.dev/docs/api/class-browser#browser-close) method is called: the contexts are closed cleanly. If you don't want to wait or to hang: you can pass the `force` option.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      <pre lang="JavaScript">for (const context of browser.contexts()) {
    await context.close();
}
await browser.close();</pre>
    </td>
    <td><pre lang="JavaScript">await browser.close();</pre></td>
  </tr>
  <tr>
    <td><pre lang="JavaScript">await browser.close();</pre></td>
    <td><pre lang="JavaScript">await browser.close({ force: true });</pre></td>
  </tr>
</table>

Issue related: [_[BUG] `browser.close` doesn't close contexts properly_ #15163](https://github.com/microsoft/playwright/issues/15163)

---

## Edit

Looking at the tests that fail, maybe we need to change the default value of the parameter. Feedback is welcome.

```TypeScript
  async close(options: { closeContexts?: boolean; } = {}): Promise<void> {
    try {
      if (options.closeContexts) {
        for (const context of this._contexts)
          await context.close();
      }
      // ...
    }
  }
```

### DIFF
--- a/docs/src/api/class-browser.md
+++ b/docs/src/api/class-browser.md
@@ -96,11 +96,13 @@ were opened).
 In case this browser is connected to, clears all created contexts belonging to this browser and disconnects from the
 browser server.
 
-:::note
-This is similar to force quitting the browser. Therefore, you should call [`method: BrowserContext.close`] on any [BrowserContext]'s you explicitly created earlier with [`method: Browser.newContext`] **before** calling [`method: Browser.close`].
-:::
-
 The [Browser] object itself is considered to be disposed and cannot be used anymore.
+
+### option: Browser.close.force
+* since: v1.25
+- `force` <boolean>
+
+Whether to bypass the closure of the contexts. Defaults to `false`.
 
 ## method: Browser.contexts
 * since: v1.8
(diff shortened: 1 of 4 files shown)


rwoll:
Thanks for the PR! However, at this time, Playwright is not comfortable changing the default behavior of `browser.close()` to include additional work that can potentially slow it down and/or cause it to hit some internal forceful timeouts. This could break existing users in unknown/unexpected/unanticipated ways.

As follow up to https://github.com/microsoft/playwright/issues/15163 we amended the docs in an attempt to clarify the behavior. However, perhaps it would be helpful to include the code snippet for folks to paste to achieve the graceful behavior:

```js
for (const context of browser.contexts())
    await context.close();
await browser.close();
```

Feel free to amend the docs to include this snippet for others. Thanks!

PR #16085, opened 2022-08-01
devops: use pip3 to install azure-cli instead of pip

mxschmitt:


mxschmitt:
Verified on u18/u20/u22

TBR Lusha.

PR #16142, opened 2022-08-02
devops: fix pip3 for webkit ubuntu18.04 build

aslushnikov:
The canonical script to install pip3 doesn't work with Python3.6
that is used to compile WebKit on Ubuntu 18.04, so we fallback
to installing pip3 from apt.


aslushnikov:
TBR @mxschmitt

PR #16162, opened 2022-08-02
feat(browser): close contexts when close browser

regseb:
I reversed the default value of the [`browser.close()`](https://playwright.dev/docs/api/class-browser#browser-close) option so as not to change the default behavior. I added the `closeContexts` option which simplifies closing the browser and its contexts.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td><pre lang="JavaScript">await browser.close();</pre></td>
    <td><pre lang="JavaScript">await browser.close();</pre></td>
  </tr>
  <tr>
    <td>
      <pre lang="JavaScript">for (const context of browser.contexts()) {
    await context.close();
}
await browser.close();</pre>
    </td>
    <td><pre lang="JavaScript">await browser.close({ closeContexts: true });</pre></td>
  </tr>
</table>

- Issue related: [_[BUG] `browser.close` doesn't close contexts properly_ #15163](https://github.com/microsoft/playwright/issues/15163)
- Pull request related: #16073. I pushed on this pull request, but my modification does not go up (probably because it is closed). So I created a new pull request. 😕

### DIFF
--- a/docs/src/api/class-browser.md
+++ b/docs/src/api/class-browser.md
@@ -97,10 +97,16 @@ In case this browser is connected to, clears all created contexts belonging to t
 browser server.
 
 :::note
-This is similar to force quitting the browser. Therefore, you should call [`method: BrowserContext.close`] on any [BrowserContext]'s you explicitly created earlier with [`method: Browser.newContext`] **before** calling [`method: Browser.close`].
+This is similar to force quitting the browser. Therefore, you should add `closeContexts` option. Or you should call [`method: BrowserContext.close`] on any [BrowserContext]'s you explicitly created earlier with [`method: Browser.newContext`] **before** calling [`method: Browser.close`].
 :::
 
 The [Browser] object itself is considered to be disposed and cannot be used anymore.
+
+### option: Browser.close.closeContexts
+* since: v1.25
+- `force` <boolean>
+
+If the contexts have to be closed. Defaults to `false`.
 
 ## method: Browser.contexts
 * since: v1.8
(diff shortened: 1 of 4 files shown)


rwoll:
We're happy to accept a doc change, but do not want to touch the API itself at this time. We can't change the default behavior for fear of breaking folks (per other PR comment). Non-default options are unlikley to be used known to folks unless they carefully read the docs, and in this case only saves from two lines:

```
for (const context of browser.contexts())
    await context.close();
```

so we'd prefer this to just be a documentation change at this point. Thanks!

regseb:
My new proposal doesn't change the default behavior. Without option, the `closeContexts` property is `false` and the content of the _if_ isn't executed.

```TypeScript
  async close(options: { closeContexts?: boolean; } = {}): Promise<void> {
    try {
      if (options.closeContexts) {
        for (const context of this._contexts)
          await context.close();
      }
      // ...
    }
  }
```

rwoll:
We do not want to touch the API itself. This includes non-default behavior and any change that would change the signature or behavior of `browser.close()`.

rwoll:
Closing as part of triage per https://github.com/microsoft/playwright/pull/16162#issuecomment-1203108642

regseb:
@rwoll To not modify the API of `close()`, we could add a new method :

```TypeScript
  async cleanClose(): Promise<void> {
    for (const context of this._contexts)
      await context.close();
    }
    return this.close();
  }
```

PR #16166, opened 2022-08-02
feat: expose BrowserType.default{Context,Launch}Options

dgozman:


pavelfeldman:
I don't think we have a clear signal for exposing those atm.

PR #16204, opened 2022-08-03
chore(ct): change setProps to rerender

sand4rt:
partial fix for: https://github.com/microsoft/playwright/issues/15919

### DIFF
--- a/packages/playwright-ct-vue/index.d.ts
+++ b/packages/playwright-ct-vue/index.d.ts
@@ -34,25 +34,22 @@ export type PlaywrightTestConfig = Omit<BasePlaywrightTestConfig, 'use'> & {
   }
 };
 
-interface MountResult<Props = { [key: string]: any }> extends Locator {
+export interface MountOptions<Props = Record<string, unknown>> {
+  props?: Props,
+  slots?: Record<string, unknown>,
+  on?: Record<string, Function>,
+  hooksConfig?: any,
+}
+
+interface MountResult<Props = Record<string, unknown>> extends Locator {
   unmount(): Promise<void>;
-  setProps(props: Partial<Props>): Promise<void>
+  rerender(options: { props: Props }): Promise<void>
 }
 
 export interface ComponentFixtures {
   mount(component: JSX.Element): Promise<MountResult>;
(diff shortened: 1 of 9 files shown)


tests/components/ct-vue-vite/src/notation-jsx.spec.tsx
```diff
@@ -1,20 +1,31 @@
...
-  await expect(component).toContainText('Loading');
+test('renderer and keep the component instance intact', async ({ mount }) => {
+  const component = await mount<{ count: number }>(Counter, {
+    props: { 
+      count: 9001
+    }
+  });
+  await expect(component.locator('#rerender-count')).toContainText('9001')
+  
+  await component.rerender({ props: { count: 1337 } })
```
sand4rt, line 18: NOTE: JSX syntax is not supported yet. API Should be:

```typescript
 await component.rerender(<Counter count={1337} />)
```

packages/playwright-ct-vue/index.d.ts
```diff
@@ -34,25 +34,22 @@ export type PlaywrightTestConfig = Omit<BasePlaywrightTestConfig, 'use'> & {
...
+  props?: Props,
+  slots?: Record<string, unknown>,
+  on?: Record<string, Function>,
+  hooksConfig?: any,
+}
+
+interface MountResult<Props = Record<string, unknown>> extends Locator {
   unmount(): Promise<void>;
-  setProps(props: Partial<Props>): Promise<void>
+  rerender(options: { props: Props }): Promise<void>
```
pavelfeldman, line 46: We are nearing the branch point, so we should probably call this `_rerender` for the time being, until we make it work for jsx + react + svelte.

pavelfeldman (approved):

pooledge:
I'd love to have `rerender` for svelte very much, for example, in order to test some functional components e.g. accordion expandable etc.

sand4rt:
> I'd love to have `rerender` for svelte very much, for example, in order to test some functional components e.g. accordion expandable etc.

Hi @pooledge, i've just created a PR for that. If you want to track it see: https://github.com/microsoft/playwright/pull/17940.

Issue #11568, opened 2022-01-22
[BUG] cannot view test report from example test


SteveALee:
*Context:

    Playwright Version: current
    Operating System: ubuntu 20.04.3 on WSL (windows 11 version)
    Node.js version: [e.g. 12.22, 14.6] 16.13.0
    Browser: [e.g. All, Chromium, Firefox, WebKit] Firefox
    Extra: [any specific details about your environment] pretty much standard WSL Ubuntu - though been on my machine a long time

In a new empty folder I did:

    npm init playwright (chose default false for os deps install)
    npx playwright test

```
wsl@OD-X1-CARBON:~/spikes/playright$  npx playwright show-report

  Serving HTML report at http://127.0.0.1:9323. Press Ctrl+C to quit.
```

A new Firefox tab is opened at `http://127.0.0.1:9323/` but the server is not found
```
Unable to connect

Firefox can’t establish a connection to the server at 127.0.0.1:9323.

    The site could be temporarily unavailable or too busy. Try again in a few moments.
    If you are unable to load any pages, check your computer’s network connection.
    If your computer or network is protected by a firewall or proxy, make sure that Firefox is permitted to access the Web.
```

I run servers in WSL and connect without any problem - eg 'netlify dev' so it must be a bug with the playwright report server.

SteveALee:
Oh. I just thought that these days Firefox is set for https only. 

Can we do this?

SteveALee:
I cannot find any docs or useful --help for `npx playwright show-report`

mxschmitt:
This seems like a WSL specific problem, which you need to workaround to have access to the WSL localhost.

This might help: https://superuser.com/questions/1594420/cant-access-127-0-0-180-outside-of-wsl2-ubuntu-20-04

Closing since this is not a bug caused by Playwright.

SteveALee:
@mxschmitt ah, that could be the case. Everything else that works fine uses a specific port number other than the default 80. Eg Netlify dev.

Shame I can't specify it on the command line.

SteveALee:
@mxschmitt there is a REALLY simple solution

Rather that `127.0.01` use `localhost`. Of course this assumes it set in a systems hosts file or equivalent. But if Netlify think it's good enough to use....

HTH

SteveALee:
Actually, given the simplicity of the page would opening using the `file://` protocol not be good enough and not bother with a server?

mxschmitt:
[...]

PR #16205, opened 2022-08-03
chore: listen on localhost for html report

jfgreffier:
I came across issue #11568 on my machine. I didn't realize it was fixed, even after reading the code.

This MR is for only for readibility and to harmonize with the rest of the code in `httpServer.ts`
It follows this suggestion https://github.com/microsoft/playwright/issues/11568#issuecomment-1173566316

Also I think localhost is more universally understood than `127.0.0.1` or `::1` for users

### DIFF
--- a/packages/playwright-core/src/utils/httpServer.ts
+++ b/packages/playwright-core/src/utils/httpServer.ts
@@ -58,7 +58,7 @@ export class HttpServer {
       this._activeSockets.add(socket);
       socket.once('close', () => this._activeSockets.delete(socket));
     });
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):
[Vite](https://github.com/vitejs/vite/blob/ccb3449edcbe08948a2d01e5d37dbca0d12b582a/packages/vite/src/node/utils.ts#L779) + Webpack Dev Server do it as well. Also tested myself the change on WSL, so lgtm from my side.

PR #16210, opened 2022-08-03
fix(Entire Browser): make sure input recording with japanese IME Work

YA2KM:
fix(Entire Browser): make sure input recording with japanese IME Work

Input events are not  recoreded collectry with japanese IME. 

thir problem caused by invaliid key event "Prosess".
"_performAction" receives fuge amount of "Prosess Key  event " in a short moment with japanese IME.

This patch fixes this problem.


YA2KM:
https://user-images.githubusercontent.com/106018566/182643940-1eaabdc3-5614-4807-95f2-74ac15108e8e.mp4

yury-s (commented):
Could you also add a test for this?

YA2KM:
sure, why not!

YA2KM:
@yury-s 

I changed my patch with an advice by my boss like this.

packages/playwright-core/src/server/injected/recorder.ts
`
if (['Shift', 'Control', 'Meta', 'Alt', 'Process'].includes(event.key))
      return false;
`

It works fine, But I am not sure where the test file is.

Could you kindly tell me where the file which test "injected/recorder.ts" is?

I suppose that I should just add  'Process' to the test file.

Thank you.

mxschmitt:
Here would be a nice place to add a test / test this special case: https://github.com/microsoft/playwright/blob/4bec6309df2c45ecbbb2767201a98f89f1cf8f5e/tests/library/inspector/cli-codegen-1.spec.ts#L228

YA2KM:
> Here would be a nice place to add a test / test this special case:

Thank you for your replying.

I heard about the file from my boss.
But, I would like to test the 'Process Key event ' rejection behavior.

I can fill input element by japanese and test will be passed  without this patch.

But I can't record japanese Input with playwright codegen.

So ,I this i should modify here.
https://github.com/microsoft/playwright/blob/4bec6309df2c45ecbbb2767201a98f89f1cf8f5e/tests/library/inspector/cli-codegen-1.spec.ts#L281

I wrote test code like below, But It doesnt work due to timeout.

```
 const [, sources] = await Promise.all([
      recorder.waitForActionPerformed(),
      page.press('input', 'Process') //timeout.
    ]);

```

How do I test?
Sorry for your inconvinience.
I am newbie of playwright and not fluent eng speaker.
thank you.

packages/playwright-core/src/server/recorder.ts
```diff
@@ -484,6 +484,7 @@ class ContextRecorder extends EventEmitter {
   }
 
   private async _performAction(frame: Frame, action: actions.Action) {
+
```
mxschmitt, line 487:
```suggestion
```

yury-s (approved):

mxschmitt (approved):

yury-s:
@YA2KM  heads up, I had to revert the change as the test was failing on all bots. Feel free to open another PR with a passing test.

mxschmitt:
I think I broke it with my "formatting fixes", let me reland it.

PR #16212, opened 2022-08-03
devops: fix Firefox build on Ubuntu 22.04 arm64

aslushnikov:
Turns out the weird compilation error is fixed if we use Clang 14
to compile.

This is the same clang that would've been shipped in mozilla build
toolchain.


mxschmitt (approved):

Mone71:
Thanks

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Issue #16180, opened 2022-08-02
[BUG] 'page screenshot › should capture canvas changes' is failing on macOS 11


yury-s:
With [the recent](https://github.com/microsoft/playwright/pull/16151) roll it started failing again on macOS 11. It passes on macOS 12 (locally).

PR #16218, opened 2022-08-03
devops: add macos-12 bots, rearrange others

yury-s:
* Added macos-12 bots to the secondary workflow
* Moved ubuntu 20.04 from primary to secondary workflow
* For the bots where we don't care about macos version (Chrome Stable, Edge Dev etc.) switched to macos-latest

References #16180


aslushnikov (approved):

yury-s:
cc @dgozman

PR #16225, opened 2022-08-03
chore: preserve window while reusing window

pavelfeldman:


dgozman (approved):

packages/playwright-core/src/server/firefox/ffPage.ts
```diff
@@ -377,6 +377,10 @@ export class FFPage implements PageDelegate {
     });
   }
 
+  async updateUserAgent(): Promise<void> {
```
dgozman, line 380: I don't see anyone calling this method.

packages/playwright-core/src/server/page.ts
```diff
@@ -247,10 +247,11 @@ export class Page extends SdkObject {
     this._extraHTTPHeaders = undefined;
     this._interceptFileChooser = false;
 
-    await this._delegate.updateEmulatedViewportSize();
-    await this._delegate.updateEmulateMedia();
-    await this._delegate.updateExtraHTTPHeaders();
```
dgozman, line 252: Where did extra http headers go?

dgozman (approved):

packages/playwright-core/src/server/page.ts
pavelfeldman: Those are handled via the browser.

PR #16229, opened 2022-08-03
feat(reuse): account for the browser launch args when reusing the bro…

pavelfeldman:
…wsers

### DIFF
--- a/packages/playwright-core/src/remote/playwrightConnection.ts
+++ b/packages/playwright-core/src/remote/playwrightConnection.ts
@@ -24,11 +24,12 @@ import { registry } from '../server';
 import { SocksProxy } from '../common/socksProxy';
 import type { Mode } from './playwrightServer';
 import { assert } from '../utils';
+import type { LaunchOptions } from '../server/types';
 
 type Options = {
   enableSocksProxy: boolean,
   browserAlias: string | null,
-  headless: boolean,
+  launchOptions: LaunchOptions,
 };
 
 type PreLaunched = {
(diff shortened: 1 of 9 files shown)


packages/playwright-core/src/remote/playwrightServer.ts
```diff
@@ -120,17 +121,28 @@ export class PlaywrightServer {
...
+      const launchOptionsHeader = request.headers['x-playwright-launch-options'] || '';
+      let launchOptions: LaunchOptions = {};
+      try {
+        launchOptions = JSON.parse(Array.isArray(launchOptionsHeader) ? launchOptionsHeader[0] : launchOptionsHeader);
+      } catch (e) {
+      }
+
+      const headlessHeader = request.headers['x-playwright-headless'];
+      const headlessValue = url.searchParams.get('headless') || (Array.isArray(headlessHeader) ? headlessHeader[0] : headlessHeader);
+      if (headlessValue !== '0')
```
dgozman, line 137: I think this does not work when `x-playwright-headless` is not present, and `x-playwright-launch-options` has `headless: false`.

```suggestion
      if (headlessValue !== '0' && headlessValue)
```

packages/playwright-core/src/remote/playwrightConnection.ts
```diff
@@ -99,7 +100,7 @@ export class PlaywrightConnection {
     const socksProxy = this._options.enableSocksProxy ? await this._enableSocksProxy(playwright) : undefined;
     const browser = await playwright[executable.browserName!].launch(serverSideCallMetadata(), {
       channel: executable.type === 'browser' ? undefined : executable.name,
-      headless: this._options.headless,
+      headless: this._options.launchOptions?.headless,
```
dgozman, line 103: Pass `launchOptions` instead?

dgozman (approved):

packages/playwright-core/src/remote/playwrightConnection.ts
pavelfeldman: We should not!

PR #16231, opened 2022-08-03
docs(python): pytest-xdist reference

### LINKED ISSUE #1442: [docs] add refer to running in parallel
- refer to either pytest-xdist or pytest-parallel
- make sure we work with both of them
- add docs for them

rwoll:
Partially resolves https://github.com/microsoft/playwright-python/issues/1442.

Tests: https://github.com/microsoft/playwright-pytest/pull/128.

### DIFF
--- a/docs/src/running-tests-python.md
+++ b/docs/src/running-tests-python.md
@@ -47,6 +47,14 @@ You can run a single test, a set of tests or all tests. Tests can be run on one 
   pytest test_login.py --browser webkit --browser firefox
   ```
 
+- Running Tests in parallel
+
+  ```bash
+  pytest --numprocesses auto
+  ```
+
+  (This assumes `pytest-xdist` is installed. For more information see [here](./test-runners.md#parallelism-running-multiple-tests-at-once).)
+
 For more information see [Playwright Pytest usage](./test-runners.md) or the Pytest documentation for [general CLI usage](https://docs.pytest.org/en/stable/usage.html).
 
 ## Running Tests
--- a/docs/src/test-runners-python.md
+++ b/docs/src/test-runners-python.md
@@ -71,10 +71,12 @@ If your tests are running on a machine with a lot of CPUs, you can speed up the 
 # install dependency
 pip install pytest-xdist
 # use the --numprocesses flag
-pytest --numprocesses=auto
+pytest --numprocesses auto
 ```
 
 Depending on the hardware and nature of your tests, you can set `numprocesses` to be anywhere from `2` to the number of CPUs on the machine. If set too high, you may notice random flakes in your tests.
+
+See [Running Tests](./running-tests.md) for general information on `pytest` options.
 
 ## Examples
 
--- a/docs/src/test-runners-python.md
+++ b/docs/src/test-runners-python.md
@@ -74,7 +74,7 @@ pip install pytest-xdist
 pytest --numprocesses auto
 ```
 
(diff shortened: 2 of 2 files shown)


docs/src/test-runners-python.md
```diff
@@ -63,6 +63,19 @@ def test_my_app_is_working(fixture_name):
 - `browser_type_launch_args`: Override launch arguments for [`method: BrowserType.launch`]. It should return a Dict.
 - `browser_context_args`: Override the options for [`method: Browser.newContext`]. It should return a Dict.
 
+## Parallelism: Running Multiple Tests at Once
```
rwoll, line 66: In the future, we can improve even more by running multiple tests/contexts at once within a worker like we do in official Test Runner, but this at least improves runtime for machines that have multiple CPUs and are currently just doing everything serially.

docs/src/test-runners-python.md
```diff
@@ -63,6 +63,19 @@ def test_my_app_is_working(fixture_name):
...
+
+```bash
+# install dependency
+pip install pytest-xdist
+# use the --numprocesses flag
+pytest --numprocesses=auto
+```
+
+Depending on the hardware and nature of your tests, you can set `numprocesses` to be anywhere from `2` to the number of CPUs on the machine. If set too high, you may notice random flakes in your tests.
+
```
mxschmitt, line 78: lets maybe add a link to it inside the "running-tests-py.md"?

rwoll: added

docs/src/running-tests-python.md
```diff
@@ -47,6 +47,14 @@ You can run a single test, a set of tests or all tests. Tests can be run on one
...
   ```
 
+- Running Tests in parallel
+
+  ```bash
+  pytest --numprocesses auto
+  ```
+
+  (This assumes `pytest-xdist` is installed. For more information see [here](./test-runners.md#parallelism-running-multiple-tests-at-once).)
+
```
rwoll, line 57: @mxschmitt wdyt about adding this section?

Since it's a drop-in speed up, I like making sure users will find it, so having it here too is nice.

docs/src/test-runners-python.md
```diff
@@ -63,6 +63,21 @@ def test_my_app_is_working(fixture_name):
...
+If your tests are running on a machine with a lot of CPUs, you can speed up the overall execution time of your test suite by using [`pytest-xdist`](https://pypi.org/project/pytest-xdist/) to run multiple tests at once:
+
+```bash
+# install dependency
+pip install pytest-xdist
+# use the --numprocesses flag
+pytest --numprocesses auto
+```
+
+Depending on the hardware and nature of your tests, you can set `numprocesses` to be anywhere from `2` to the number of CPUs on the machine. If set too high, you may notice random flakes in your tests.
```
mxschmitt, line 77: why does it get flaky when its too high?

mxschmitt (approved):

docs/src/running-tests-python.md
mxschmitt: sgtm

docs/src/test-runners-python.md
rwoll: Same thing in the JS Test Runner world. We usually have 1 browser per worker, and if you run 50 browsers and only have 2 CPUs, the computer will melt! :)

rwoll, line 77:
```suggestion
Depending on the hardware and nature of your tests, you can set `numprocesses` to be anywhere from `2` to the number of CPUs on the machine. If set too high, you may notice unexpected behavior.
```

PR #16251, opened 2022-08-04
feat(plugins): send accept header in webServer url checking

ghost91-:
This PR sets the Accept HTTP header to `*/*` for HTTP requests made by the webServer plugin when using a URL to check if the server has already started. So far, the Accept header was not being sent at all, which should be semantically equivalent to sending `*/*`, according to [RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#section-5.3.2). However, there are some web servers that handle this differently. Most notably, [Vite](https://vitejs.dev/)'s server just returns a 404 if the `Accept` header is missing from the request. I created https://github.com/vitejs/vite/issues/9520 in order to address the issue on their side, but I think it's also fine to just add the header here, which is why I created this PR.

### DIFF
--- a/tests/playwright-test/web-server.spec.ts
+++ b/tests/playwright-test/web-server.spec.ts
@@ -312,7 +312,7 @@ test('should be able to specify a custom baseURL with the server', async ({ runI
   await new Promise(resolve => server.close(resolve));
 });
 
-test('should be able to use an existing server when reuseExistingServer:true ', async ({ runInlineTest }, { workerIndex }) => {
+test('should be able to use an existing server when reuseExistingServer:true', async ({ runInlineTest }, { workerIndex }) => {
   const port = workerIndex + 10500;
   const server = http.createServer((req: http.IncomingMessage, res: http.ServerResponse) => {
     res.end('<html><body>hello</body></html>');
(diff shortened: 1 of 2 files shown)


ghost91-:
The test failures seem to be completely unrelated, from what I can tell.

Is there anything I can do to help this PR to get a review?

tests/playwright-test/web-server.spec.ts
```diff
@@ -429,6 +429,37 @@ test(`should suport self signed certificate`, async ({ runInlineTest, httpsServe
   expect(result.exitCode).toBe(0);
 });
 
+test('should send Accept header', async ({ runInlineTest }, { workerIndex }) => {
+  const port = workerIndex + 10500;
+  let acceptHeader: string | undefined | null = null;
+  const server = http.createServer((req: http.IncomingMessage, res: http.ServerResponse) => {
+    if (acceptHeader === null) acceptHeader = req.headers.accept;
+    res.end('<html><body>hello</body></html>');
+  });
```
rwoll, line 438: There's a built-in `server` fixture for testing, so that should be used instead of creating a server inline here. See https://github.com/microsoft/playwright/blob/9795a4ba60479fc75de7187e681ba9e90891bb54/tests/page/page-goto.spec.ts#L72 for an example.

rwoll (changes requested):
Thanks for the PR. Minor change needed, but otherwise LGTM.

The CI failures are unrelated. (If you re-base—although not necessary—they should go away.)

tests/playwright-test/web-server.spec.ts
ghost91-: Thanks for the feedback, I'll adjust it accordingly. However, I just copied it from other tests in the same file. Should the other tests also be adjusted to use the `server` fixture`?

rwoll (approved):

tests/playwright-test/web-server.spec.ts
rwoll: Oops—thanks for pointing that out.

> Should the other tests also be adjusted to use the server fixture?

I'll have to take a closer look before cleaning those up, but I've merged your change.

PR #16254, opened 2022-08-04
docs: Trace-viewer-intro

debs-obrien:
image

### DIFF
--- a/docs/src/trace-viewer-intro-csharp-java-python.md
+++ b/docs/src/trace-viewer-intro-csharp-java-python.md
@@ -5,13 +5,11 @@ title: "Trace Viewer"
 
 Playwright Trace Viewer is a GUI tool that lets you explore recorded Playwright traces of your tests meaning you can go back and forward though each action of your test and visually see what was happening during each action.
 
-:::info You will learn
-<ul>
-<li>How to record a trace on CI</li>
-<li>How to open the HTML report</li>
-<li>How to open the trace viewer</li>
-</ul>
-:::
+**You will learn**
+
+- How to record a trace
+- How to open the HTML report
+- How to open the trace viewer
 
 ## Recording a trace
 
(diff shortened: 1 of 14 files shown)


docs/src/trace-viewer-intro-csharp-java-python.md
```diff
@@ -0,0 +1,120 @@
...
+context.tracing().stop(new Tracing.StopOptions()
+  .setPath(Paths.get("trace.zip")));
+```
+
+```python async
+browser = await chromium.launch()
+context = await browser.new_context()
+
+# Start tracing before creating / navigating a page.
+await context.tracing.start(screenshots=True, snapshots=True, sources=True)
```
mxschmitt, line 41: We should mention the pytest plugin way there instead of the library approach.

docs/src/trace-viewer.md
```diff
@@ -52,13 +52,14 @@ await page.goto('https://playwright.dev');
 await context.tracing.stop({ path: 'trace.zip' });
 ```
 
-You can also use `trace: 'retain-on-failure'` if you do not enable retries but still want traces for failed tests.
-
 Available options to record a trace:
+- `'on-first-retry'` - Record a trace only when retrying a test for the first time.
```
mxschmitt, line 56: Mention to make sure to have retries >= 1?

debs-obrien: This doc needs a full rework which will be part of another pr

docs/src/trace-viewer-intro-csharp-java-python.md
debs-obrien: you will have to add what you mean here as no idea what the pytest plugin way is

mxschmitt: https://playwright.dev/python/docs/test-runners

`pytest --trace on`

mxschmitt (approved):

PR #16269, opened 2022-08-04
feat(reuse): retain different browser types when reusing

pavelfeldman:


packages/playwright-core/src/remote/playwrightServer.ts
```diff
@@ -120,7 +120,7 @@ export class PlaywrightServer {
       }
       const url = new URL('http://localhost' + (request.url || ''));
       const browserHeader = request.headers['x-playwright-browser'];
-      const browserAlias = url.searchParams.get('browser') || (Array.isArray(browserHeader) ? browserHeader[0] : browserHeader) || null;
+      const browserName = url.searchParams.get('browser') || (Array.isArray(browserHeader) ? browserHeader[0] : browserHeader) || null;
```
dgozman, line 123: I think this breaks existing `connect()` contract - it should come with some changes in fixtures.

dgozman (approved):

PR #16277, opened 2022-08-04
feat(test runner): add --trace cli option

dgozman:


pavelfeldman (approved):

Mone71:
Thanks

packages/playwright-test/src/cli.ts
```diff
@@ -56,6 +57,7 @@ function addTestCommand(program: Command) {
   command.option('--shard <shard>', `Shard tests and execute only the selected shard, specify in the form "current/all", 1-based, for example "3/5"`);
   command.option('--project <project-name...>', `Only run tests from the specified list of projects (default: run all projects)`);
   command.option('--timeout <timeout>', `Specify test timeout threshold in milliseconds, zero for unlimited (default: ${defaultTimeout})`);
+  command.option('--trace <mode>', `Force tracing mode, can be ${kTraceModes.map(mode => `"${mode}"`).join(', ')}`);
```
rwoll, line 60: Why is `<mode>` required? If `<mode>`. I'd expect it to default to `on`. Doing `--trace on` seems redundant for a command line switch.

rwoll (changes requested):

packages/playwright-test/src/cli.ts
rwoll: Comment edited as something got lost when I hit save!

rwoll (approved):
We can't different between `--trace a.spec.ts` and `--trace <mode>` nicely so ignore my last review!

PR #16283, opened 2022-08-05
docs: explain toHaveText and toContainText in more details

### LINKED ISSUE #15666: [BUG] Explicitly control subset of elements in toContainText (and alike)

dgozman:
Drive-by: fix dotnet snippets.

Closes #15666.

### DIFF
--- a/docs/src/api/class-locatorassertions.md
+++ b/docs/src/api/class-locatorassertions.md
@@ -762,36 +762,95 @@ await Expect(locator).ToContainTextAsync("substring");
 await Expect(locator).ToContainTextAsync(new Regex("\\d messages"));
 ```
 
-Note that if array is passed as an expected value, entire lists of elements can be asserted:
+If you pass an array as an expected value, the expectations are:
+1. Locator resolves to a list of elements.
+1. Elements from a **subset** of this list contain text from the expected array, respectively.
+1. The matching subset of elements has the same order as the expected array.
+1. Each text value from the expected array is matched by some element from the list.
+
+For example, consider the following list:
+```html
+<ul>
+  <li>Item Text 1</li>
+  <li>Item Text 2</li>
+  <li>Item Text 3</li>
+</ul>
+```
+
+Let's see how we can use the assertion:
 
 ```js
-const locator = page.locator('list > .list-item');
-await expect(locator).toContainText(['Text 1', 'Text 4', 'Text 5']);
+// ✓ Contains the right items in the right order
+await expect(page.locator('ul > li')).toContainText(['Text 1', 'Text 3']);
+
+// ✖ Wrong order
+await expect(page.locator('ul > li')).toContainText(['Text 3', 'Text 2']);
+
+// ✖ No item contains this text
+await expect(page.locator('ul > li')).toContainText(['Some 33']);
+
+// ✖ Locator points to the outer list element, not to the list items
+await expect(page.locator('ul')).toContainText(['Text 3']);
 ```
 
 ```java
-assertThat(page.locator("list > .list-item")).containsText(new String[] {"Text 1", "Text 4", "Text 5"});
+// ✓ Contains the right items in the right order
+assertThat(page.locator("ul > li")).containsText(new String[] {"Text 1", "Text 3", "Text 4"});
+
+// ✖ Wrong order
+assertThat(page.locator("ul > li")).containsText(new String[] {"Text 3", "Text 2"});
+
+// ✖ No item contains this text
+assertThat(page.locator("ul > li")).containsText(new String[] {"Some 33"});
+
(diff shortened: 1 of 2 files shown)


docs/src/api/class-locatorassertions.md
```diff
@@ -60,8 +60,8 @@ public class ExampleTests : PageTest
     public async Task StatusBecomesSubmitted()
     {
         // ..
-        await Page.Locator("#submit-button").ClickAsync();
-        await Expect(Page.Locator(".status")).ToHaveTextAsync("Submitted");
+        await page.Locator("#submit-button").ClickAsync();
```
mxschmitt, line 63: The fix is wrong, inside .NET page is an instance variable called `Page`.

docs/src/api/class-locatorassertions.md
```diff
@@ -757,41 +757,91 @@ expect(locator).to_contain_text(re.compile(r"\d messages"))
...
 ```csharp
-var locator = Page.Locator(".title");
+var locator = page.Locator(".title");
 await Expect(locator).ToContainTextAsync("substring");
 await Expect(locator).ToContainTextAsync(new Regex("\\d messages"));
 ```
 
-Note that if array is passed as an expected value, entire lists of elements can be asserted:
+If you pass an array as an expected value, the expectations are:
+1. Locator matches a list of elements.
```
rwoll, line 766: I had to re-read this a few times. I think it comes down to the fact that:

1. Locators are associated with the idea of "strictness" (that's their main marketing), but these assertions all rely on them resolving to lists. I don't think it's always clear to users that Locators are lazy and it's only when you make an action or assertion involving them that they get resolved and that's when the strictness comes into play.
2. "matches" vs. "resolves": This says _Locator_ **matches** list of _elements_. But, I think _Locator_ **resolves to** a list of _elements_. And _elements_ **match** an array of text.

I think this is all improved by carefully using the words **resolve** and **match**. Reserve **resolve** for Locators since _resolve to_ implies laziness and taking from abstract to concrete elements (plus it mirrors the language in our error messages _locator.click: Error: strict mode violation: "li" resolved to 2 elements:_), and reserve **match** for once you have concrete elements, they can match some text.

rwoll (changes requested):
Let me know if you have questions about the comments. I think they are nuanced and important!

Overall, the illustrative examples are helpful.

docs/src/api/class-locatorassertions.md
```diff
@@ -1286,34 +1345,94 @@ await Expect(locator).ToHaveTextAsync(new Regex("Welcome, .*"));
...
+1. The number of elements equals the number of expected values in the array.
+1. Elements from the list have text matching expected array values, one by one, in order.
+
+For example, consider the following list:
+```html
+<ul>
+  <li>Text 1</li>
+  <li>Text 2</li>
+  <li>Text 3</li>
+</ul>
```
rwoll, line 1359:
```suggestion
<ul>
  <li>Item Text 1</li>
  <li>Item Text 2</li>
  <li>Item Text 3</li>
</ul>
```

Make this the same example list as the previous section.

docs/src/api/class-locatorassertions.md
```diff
@@ -1286,34 +1345,94 @@ await Expect(locator).ToHaveTextAsync(new Regex("Welcome, .*"));
...
+await expect(page.locator('ul > li')).toHaveText(['Text 1', 'Text 2', 'Text 3']);
+
+// ✖ Wrong order
+await expect(page.locator('ul > li')).toHaveText(['Text 3', 'Text 2', 'Text 1']);
+
+// ✖ Last item does not match
+await expect(page.locator('ul > li')).toHaveText(['Text 1', 'Text 2', 'Text']);
+
+// ✖ Locator points to the outer list element, not to the list items
+await expect(page.locator('ul')).toHaveText(['Text 1', 'Text 2', 'Text 3']);
```
rwoll, line 1375:
```suggestion
await expect(page.locator('ul > li')).toHaveText(['Item Text 1', 'Item Text 2', 'Item Text 3']);

// ✖ Wrong order
await expect(page.locator('ul > li')).toHaveText(['Item Text 3', 'Item Text 2', 'Item Text 1']);

// ✖ Last item does not match
await expect(page.locator('ul > li')).toHaveText(['Item Text 1', 'Item Text 2', 'Text']);

// ✖ Locator points to the outer list element, not to the list items
await expect(page.locator('ul')).toHaveText(['Item Text 1', 'Item Text 2', 'Item Text 3']);
```

rwoll (approved):

docs/src/api/class-locatorassertions.md
dgozman: This makes expect examples too long.

PR #16324, opened 2022-08-05
chore: allow inspecting element from server

pavelfeldman:


packages/playwright-core/src/protocol/protocol.yml
```diff
@@ -937,7 +937,12 @@ BrowserContext:
       experimental: True
       parameters:
         language: string?
-        startRecording: boolean?
+        mode:
+          type: enum?
+          literals:
+          - inspecting
+          - recording
+        startInspecting: boolean?
```
dgozman, line 945: I don't see this flag being used.

dgozman (approved):

PR #16331, opened 2022-08-06
docs(dotnet): consistent access modifiers

timdeschryver:
Use `public async` instead of `async public` to use the same access modifiers in all examples.
My IDE (Rider) was also giving a warning on using `async public`. 

While I was in the files making the changes, I also caught a few typos.

### DIFF
--- a/docs/src/test-runners-csharp.md
+++ b/docs/src/test-runners-csharp.md
@@ -42,7 +42,7 @@ namespace PlaywrightTests;
 public class MyTest : PageTest
 {
     [Test]
-    async public Task ShouldHaveTheCorrectSlogan()
+    public async Task ShouldHaveTheCorrectSlogan()
     {
         await Page.GotoAsync("https://playwright.dev");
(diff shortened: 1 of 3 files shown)


docs/src/test-runners-csharp.md
```diff
@@ -299,7 +299,7 @@ dotnet test
 You can also choose specifically which tests to run, using the [filtering capabilities](https://docs.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=mstest):
 
 ```bash
-dotnet test --filter "Name~ShouldAdd"
+dotnet test --filter "Name~Slogan"
```
timdeschryver, line 302: Running this command against the examples result that no tests are executed.
I thought it would be better to filter a test from the examples while still having one test executed, if this is unwanted I can revert this change.

mxschmitt (approved):

mxschmitt:
Thank you Tim, appreciate your contribution to the project!

PR #16334, opened 2022-08-06
chore: minor inspect tweaks

pavelfeldman:


packages/playwright-core/src/cli/driver.ts
```diff
@@ -90,13 +93,40 @@ function wireController(playwright: Playwright, wsEndpoint: string) {
...
+            recorder.setHighlightedSelector('');
+            recorder.setMode('none');
+          }
+          return;
+        }
+
+        // Create browser if none.
+        const browsers = playwright.allBrowsers();
+        if (!browsers.length) {
+          process.send!({ method: 'error', params: { error: 'NEW BROWSER' } });
```
dgozman, line 107: Is this some kind of logging? Perhaps it's better to `console.log` and wire that on the client side?

dgozman (approved):

PR #16335, opened 2022-08-07
feat(chromium): roll to r1019

playwrightmachine:


mxschmitt:
After https://chromium-review.googlesource.com/c/chromium/src/+/3703191 it seems fine ignoring unrelated events for now without a backendNodeId.

Issue #16345, opened 2022-08-08
[BUG] Package.json exports should include "default" fallback key


jeremy-daley-kr:
**Context:**
- Playwright Version: 1.24.0
- Operating System: Mac
- Node.js version: 16.13.1
- Browser: NA
- Extra: NA

**Describe the bug**

The `eslint-plugin-import` plugin uses a [`import/no-unresolved`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unresolved.md) directive which fails on `@playwright/test` unless a [`default`](https://nodejs.org/api/packages.html#conditional-exports) export is defined in the `package.json` file.

Does this seem like a reasonable inclusion? I can't say if it's considered best practice to include the fallback, but it _is_ explicitly mentioned in Node's docs.

rwoll:
Fixed by #16350 and #16360

PR #16350, opened 2022-08-08
Add fallback "default" key to "." exports object

jeremy-daley-kr:
Adds a [`default`](https://nodejs.org/api/packages.html#conditional-exports) export in the `package.json` file.

See https://github.com/microsoft/playwright/issues/16345

### DIFF
--- a/packages/playwright-test/package.json
+++ b/packages/playwright-test/package.json
@@ -12,7 +12,8 @@
     ".": {
       "types": "./index.d.ts",
       "import": "./index.mjs",
-      "require": "./index.js"
+      "require": "./index.js",
+      "default": "./index.js"
     },
     "./cli": "./cli.js",
     "./package.json": "./package.json",


rwoll (approved):
Reading through the docs https://nodejs.org/api/packages.html#conditional-exports and https://nodejs.org/api/packages.html#dual-commonjses-module-packages, this seems OK. And it also seems appropriate to default to CJS (since it can be loaded from ESM envs but not the other way around).

jeremy-daley-kr:
Thanks so much @rwoll 
Is there documentation around Playwright's typical release schedule, so I might know when this is publicly available?

rwoll:
@jeremy-daley-kr We don't publish a release schedule, but you can watch Releases of this repo on GitHub to get notifications:

Screen Shot 2022-08-10 at 11 21 02 AM

Within ~1hr of getting the release notification, CI publishes it to npm.

Playwright also publishes a canary daily, so you can try it out today by doing `npm i @playwright/test@next`.

1.25 is branched, so the release will more than likley be within the next few days.

jeremy-daley-kr:
@rwoll Just noticed it's in the latest release! Thanks, again!

PR #16354, opened 2022-08-08
docs: add migration from Puppeteer guide

### LINKED ISSUE #13646: [Feature] docs - Migrate puppeteer to playwright
similar to #13608

Having docs for migrating from puppeteer to playwright and handling common scenarios would be helpful
some scenarios can be taken from the official puppeteer docs to show the API change.

for example:
```ts
import * as playwright from 'playwright';
import * as puppeteer from 'puppeteer';

async function puppeteerMain() {
    const browser = await puppeteer.launch(); // defaults to chrome
    await page.goto('https://example.com');
    await page.screenshot({ path: 'example.png' });

    await browser.close();
}

async function playwrightMain() {
    const browser = await playwright.chromium.launch();
    await page.goto('https://example.com');
    await page.screenshot({ path: 'example.png' });

    await browser.close();
}
```

jfgreffier:
Migration guide from Puppeteer. 

The cheat sheet is partially built on https://github.com/checkly/checklyhq.com/blob/main/site/content/guides/puppeteer-to-playwright.md from checkly (MIT license)

Example is my own, tested and both are functionals.

I tried to fit with the style and level of details of the other migrations docs.

Fixes #13646

### DIFF
--- a/docs/src/puppeteer-js.md
+++ b/docs/src/puppeteer-js.md
@@ -1,0 +1,178 @@
+---
+id: puppeteer
+title: "Migrating from Puppeteer"
+---
+
+<!-- TOC -->
+
+## Migration Principles
+
+This guide describes migration to [Playwright Library](./library) and [Playwright Test](./intro.md) from Puppeteer. The APIs have similarities, but Playwright offers much more possibilities for web testing and cross-browser automation.
+
+- Most Puppeteer APIs can be used as is
+- The use of [ElementHandle] is discouraged, use [Locator] objects and web-first assertions instead.
+- Playwright is cross-browser
+- You probably don't need explicit wait
+
+## Cheat Sheet
+
+| Puppeteer                                          | Playwright Library                          |
+|----------------------------------------------------|---------------------------------------------|
+| `await puppeteer.launch()`                         | `await playwright.chromium.launch()`        |
+| `puppeteer.launch({product: 'firefox'})`           | `await playwright.firefox.launch()`         |
+|  WebKit is not supported by Puppeteer              | `await playwright.webkit.launch()`          |
+| `await browser.createIncognitoBrowserContext(...)` | `await browser.newContext(...)`             |
+| `await page.setViewport(...)`                      | `await page.setViewportSize(...)`           |
+| `await page.waitForXPath(XPathSelector)`           | `await page.waitForSelector(XPathSelector)` |
+| `await page.waitForNetworkIdle(...)`               | `await page.waitForLoadState({ state: 'networkidle' }})` |
+| `await page.$eval(...)`                            | [Assertions](./test-assertions) can often be used instead to verify text, attribute, class... |
+| `await page.$(...)`                                | Discouraged, use [Locators](./api/class-locator) instead |
+| `await page.$x(xpath_selector)`                    | Discouraged, use [Locators](./api/class-locator) instead |
+| No methods dedicated to checkbox or radio input    | `await page.locator(selector).check()`<br/>`await page.locator(selector).uncheck()` |
+| `await page.click(selector)`                       | `await page.locator(selector).click()`      |
+| `await page.focus(selector)`                       | `await page.locator(selector).focus()`      |
+| `await page.hover(selector)`                       | `await page.locator(selector).hover()`      |
+| `await page.select(selector, values)`              | `await page.locator(selector).selectOption(values)` |
+| `await page.tap(selector)`                         | `await page.locator(selector).tap()`        |
+| `await page.type(selector, ...)`                   | `await page.locator(selector).type(...)`<br/> Please also consider [`method: Locator.fill`] |
+| `await page.waitForFileChooser(...)`<br/>`await elementHandle.uploadFile(...)` | `await page.locator(selector).setInputFiles(...)` |
+| `await page.cookies([...urls])`                    | `await browserContext.cookies([urls])`      |
+| `await page.deleteCookie(...cookies)`              | `await browserContext.clearCookies()`       |
+| `await page.setCookie(...cookies)`                 | `await browserContext.addCookies(cookies)`  |
+| `page.on(...)`                                     | `page.on(...)`<br/>In order to intercept and mutate requests, see [`method: Page.route`] |
+
+`page.waitForNavigation` and `page.waitForSelector` remain, but in many cases will not be necessary due to [auto-waiting](./actionability).
+
+The use of [ElementHandle] is discouraged , use [Locator] objects and web-first assertions instead.
+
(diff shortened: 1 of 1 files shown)


docs/src/puppeteer-js.md
```diff
@@ -0,0 +1,110 @@
...
+| `await page.setViewport(...)`                      | `await page.setViewportSize(...)`           |
+| `await page.waitForSelector(selector)`<br/>`await page.click(selector);` | `await page.click(selector)` |
+| `await page.waitForXPath(XPathSelector)`           | `await page.waitForSelector(XPathSelector)` |
+| `await page.$(...)`                                | Discouraged, use [Locators](./api/class-locator) instead |
+| `await page.$x(xpath_selector)`                    | `await page.$(xpath_selector)`              |
+| `await page.waitForNetworkIdle(...)`               | `await page.waitForLoadState({ state: 'networkidle' }})` |
+| `await page.cookies([...urls])`                    | `await browserContext.cookies([urls])`      |
+| `await page.deleteCookie(...cookies)`              | `await browserContext.clearCookies()`       |
+| `await page.setCookie(...cookies)`                 | `await browserContext.addCookies(cookies)`  |
+| `page.on('request', ...)`                          | Handled through [`method: Page.route`]. |
```
Meir017, line 31: I'm not sure this is entirely true, in puppeteer the `request` event is not only for request interception, maybe request interception should be in it's own dedicated section below the table?

docs/src/puppeteer-js.md
```diff
@@ -0,0 +1,110 @@
...
+
+- Most Puppeteer APIs can be used as is
+- Playwright is cross-browser
+- You probably don't need explicit wait
+
+## Cheat Sheet
+
+| Puppeteer                                          | Playwright Library                          |
+|----------------------------------------------------|---------------------------------------------|
+| `await puppeteer.launch(...)`                      | `await playwright.chromium.launch(...)`     |
```
Meir017, line 20: consider describing how to start a specific browser (firefox / chromium) and maybe noting that webkit is only supported in playwright

docs/src/puppeteer-js.md
```diff
@@ -0,0 +1,110 @@
...
+| `await page.waitForNetworkIdle(...)`               | `await page.waitForLoadState({ state: 'networkidle' }})` |
+| `await page.cookies([...urls])`                    | `await browserContext.cookies([urls])`      |
+| `await page.deleteCookie(...cookies)`              | `await browserContext.clearCookies()`       |
+| `await page.setCookie(...cookies)`                 | `await browserContext.addCookies(cookies)`  |
+| `page.on('request', ...)`                          | Handled through [`method: Page.route`]. |
+| `await page.waitForFileChooser(...)`<br/>`await elementHandle.uploadFile(...)`              | `await elementHandle.setInputFiles(...)`    |
+
+`page.waitForNavigation` and `page.waitForSelector` remain, but in many cases will not be necessary due to [auto-waiting](./actionability).
+
+## Example
```
Meir017, line 36: Adding a test comparison will highlight the improvement that `@playwright/test` adds

jfgreffier: My intent here was to offer a simple automation example. Indeed it can be useful to demonstrate `@playwright/test`

Do you think I should add a sample in this "Example" section, or in the "Testing" one ?

Meir017: I think we can have just a example using a test, similar to https://playwright.dev/docs/protractor#example or maybe keep the current script as well, I'm not sure

docs/src/puppeteer-js.md
jfgreffier: Right, page.on() is mostly the same. I've added a mention to `Page.route` that explains it is meant to change the requests

rwoll:
Requesting review from either @dgozman or @aslushnikov since they know the puppeteer internals quite well! 😄

jfgreffier:
> When we simply migrate users from one syntax to another, we don't really add much value. With Playwright, you are not supposed to use any of the suggested methods, you aren't even supposed to use `page.click()`, should rather use the strict `page.locator.click()` instead, should only use web-first assertions for content verification, etc. Would you be interested in putting that together?

Thanks for the feedback. Sure, I'll update the PR.

docs/src/puppeteer-js.md
```diff
@@ -0,0 +1,168 @@
...
+- Most Puppeteer APIs can be used as is
+- The use of [ElementHandle] is discouraged, use [Locator] objects and web-first assertions instead.
+- Playwright is cross-browser
+- You probably don't need explicit wait
+
+## Cheat Sheet
+
+| Puppeteer                                          | Playwright Library                          |
+|----------------------------------------------------|---------------------------------------------|
+| `await puppeteer.launch()`                         | `await playwright.chromium.launch()`        |
```
pavelfeldman, line 21: This looks much better, thanks!

Can we list the actions here?
- page.click => page.locator.click
- ...

And I would add a separate list of assertions such as `exepect(locator).toHaveText()`. The reality is that Pptr scripts will be heavy on `page.evaluate` and `page.waitForFunction` for things like getting and waiting for the `.textContent`, `.getAttribute`, etc., etc. I think it would be useful to mention it here since it would make the migrated script so much better (more readable and not flaky).

docs/src/puppeteer-js.md
```diff
@@ -0,0 +1,168 @@
...
+    await page.goto('https://playwright.dev/');
+    const titleLocator = page.locator('.hero__title'); // 4
+    await expect(titleLocator).toContainText( // 5
+      'Playwright enables reliable end-to-end testing'
+    );
+  });
+});
+```
+
+1. Each Playwright Test file has explicit import of the `test` and `expect` functions
```
pavelfeldman, line 130: I think the major distinction would be that in this set up each test is isolated. I.e. if you pollute the localStorage in the first test, second test has clean environment. There are use cases when you don't want that and in those cases you could basically use pptr script as is.

pavelfeldman (commented):
I think this is a great start. Having the migration for `page.click` would make it immediately useful.

docs/src/puppeteer-js.md
jfgreffier: I've listed several actions : `check`, `click`, `focus`, `hover`, `select`, `tap`, `type`

There are now mentions of `$eval` vs web-first assertions in the cheat sheet, test example and the test section.

docs/src/puppeteer-js.md
jfgreffier: I added mention of it line 140, with page fixture to emphasis that you get a new page on every test.

pavelfeldman (approved):

PR #16362, opened 2022-08-08
fix: toBeFocused should match shadow elements

### LINKED ISSUE #16268: [BUG] toBeFocused() matcher isn't piercing Shadow DOM
**Context:**
- Playwright Version: [1.24.2]
- Operating System: [Mac]
- Node.js version: [16.13.2]
- Browser: [any]
- Extra: []

**Code Snippet**

Unfortunately I have a web component built with Svelte and it will be hard to include it all, but the gist is:

Web component named `<my-switch>` with a control inside of it `<input>`. The only element with focus is the `<input>` in this component, e.g:

```
<my-switch>
   shadowroot
      <input />
<my-switch>
```

```javascript
    await page.locator('my-switch  input').focus();
    await expect(page.locator('my-switch')).toBeFocused(); // passes but it isn't really focused
    await expect(page.locator('my-switch input')).toBeFocused(); // fails assertion but it is focused
```

When I run in headed mode and pause after focus() is run,
`document.activeElement` returns `my-switch` element.
`document.activeElement.shadowRoot.activeElement` returns `input` element.

This is because input is really what has focus, but the browser doesn't pierce the shadow DOM. 

**Describe the bug**

It appears Playwright is not piercing shadow DOM for this matcher and is returning the parent element within the browser's DOM that is focused. Playwright does a great job piercing shadow DOM for other functionality, I'd expect toBeFocused to also pierce shadow DOM to get the right focused element.

Call log:
```
      - expect.toBeFocused with timeout 5000ms
      - waiting for selector "input"
      -   selector resolved to <input type="checkbox" id="50260257-8885-4b84-b894-8d6b…/>
      -   unexpected value "false"
      -   selector resolved to <input type="checkbox" id="50260257-8885-4b84-b894-8d6b…/>
      -   unexpected value "false"
```

rwoll:
Fixes #16268.

### DIFF
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -668,14 +668,19 @@ export class InjectedScript {
     return 'done';
   }
 
+  private _activelyFocused(node: Node): { activeElement: Element | null, isFocused: boolean } {
+    const activeElement = (node.getRootNode() as (Document | ShadowRoot)).activeElement;
+    const isFocused = activeElement === node && !!node.ownerDocument && node.ownerDocument.hasFocus();
+    return { activeElement, isFocused };
+  }
+
   focusNode(node: Node, resetSelectionIfNotFocused?: boolean): 'error:notconnected' | 'done' {
     if (!node.isConnected)
       return 'error:notconnected';
     if (node.nodeType !== Node.ELEMENT_NODE)
       throw this.createStacklessError('Node is not an element');
 
-    const activeElement = (node.getRootNode() as (Document | ShadowRoot)).activeElement;
-    const wasFocused = activeElement === node && node.ownerDocument && node.ownerDocument.hasFocus();
+    const { activeElement, isFocused: wasFocused } = this._activelyFocused(node);
     if ((node as HTMLElement).isContentEditable && !wasFocused && activeElement && (activeElement as HTMLElement | SVGElement).blur) {
       // Workaround the Firefox bug where focusing the element does not switch current
       // contenteditable to the new element. However, blurring the previous one helps.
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1029,7 +1029,7 @@ export class InjectedScript {
       } else if (expression === 'to.be.enabled') {
         elementState = progress.injectedScript.elementState(element, 'enabled');
       } else if (expression === 'to.be.focused') {
-        elementState = document.activeElement === element;
+        elementState = document.activeElement === element || document.activeElement?.shadowRoot?.activeElement === element;
```
rwoll, line 1032: I don't think this contradicts our doc spec (https://playwright.dev/docs/test-assertions#locator-assertions-to-be-focused):

```
Ensures the [Locator](https://playwright.dev/docs/api/class-locator) points to a focused DOM node.
```

since I'd include _shadow_ DOM node in "focused DOM node".

tests/playwright-test/playwright.expect.true.spec.ts
```diff
@@ -352,6 +352,35 @@ test('should support toBeFocused', async ({ runInlineTest }) => {
...
+            const input = document.createElement('input');
+            input.id = "my-input"
+            shadowRoot.appendChild(input);
+          </script>
+        \`);
+
+        await page.locator("input").focus();
+        expect(await page.evaluate(() => document.activeElement.shadowRoot.activeElement.id)).toBe("my-input");
+        await expect(page.locator("#app")).toBeFocused();
+        await expect(page.locator("input")).toBeFocused();
```
rwoll, line 376: NB: These are two different elements that are focused. I think this aligns with our definition of this assertion, and intuitively makes sense.

packages/playwright-core/src/server/injected/injectedScript.ts
dgozman: This sounds good, but I think we should copy implementation from: https://github.com/microsoft/playwright/blob/57fcb590f7384e50c05671a5d8467da1bfd5fd32/packages/playwright-core/src/server/injected/injectedScript.ts#L677-L678

rwoll: Updated. Thanks! I copied those two lines, then decided to put in separate function that can be shared. I don't like the latter much due to the need to return the `activeElement` for the first use-case, but fine with either. wdyt?

dgozman (approved):

PR #16365, opened 2022-08-08
chore: trace viewer fallback error

### LINKED ISSUE #16349: [BUG] no error message when clicking trace icon from downloaded playwright report
**Context:**
- Playwright Version: latest
- Operating System: Windows
- Node.js version: [e.g. 12.22, 14.6]
- Browser: [e.g. All, Chromium, Firefox, WebKit]

**Code Snippet**

**Describe the bug**

when clicking on trace icon from CI no error message is given saying that you need to deploy to view, instead you just get a blank page and presume it's broken or you did something wrong.

Download html report from CI, extracting the zip and double clicking the index.html to open and click on trace icon.

rwoll:
Fixes #16349.

### DIFF
--- a/packages/trace-viewer/index.html
+++ b/packages/trace-viewer/index.html
@@ -28,6 +28,7 @@
     <script type="module" src="/src/index.tsx"></script>
     <dialog id="fallback-error">
       <p>The Playwright Trace Viewer must be loaded over the <code>http://</code> or <code>https://</code> protocols.</p>
+      <p>For more information, please see the <a href="https://aka.ms/playwright/err-trace-viewer-file-protocol">docs</a>.</p>
     </dialog>
     <script>
       if (!/^https?:/.test(window.location.protocol))
--- a/packages/trace-viewer/index.html
+++ b/packages/trace-viewer/index.html
@@ -26,12 +26,12 @@
   <body>
     <div id="root"></div>
     <script type="module" src="/src/index.tsx"></script>
-    <dialog id="fallback-error" open="false">
+    <dialog id="fallback-error">
       <p>The Playwright Trace Viewer must be loaded over the <code>http://</code> or <code>https://</code> protocols.</p>
     </dialog>
     <script>
       if (!/^https?:/.test(window.location.protocol))
-        document.getElementById("fallback-error").open = "true";
+        document.getElementById("fallback-error").show();
     </script>
   </body>
 </html>
--- a/packages/trace-viewer/index.html
+++ b/packages/trace-viewer/index.html
@@ -26,5 +26,12 @@
   <body>
     <div id="root"></div>
     <script type="module" src="/src/index.tsx"></script>
+    <dialog id="fallback-error" open="false">
+      <p>The Playwright Trace Viewer must be loaded over the <code>http://</code> or <code>https://</code> protocols.</p>
+    </dialog>
+    <script>
+      if (!/^https?:/.test(window.location.protocol))
+        document.getElementById("fallback-error").open = "true";
+    </script>
   </body>
 </html>
--- a/packages/trace-viewer/index.html
+++ b/packages/trace-viewer/index.html
@@ -28,7 +28,7 @@
     <script type="module" src="/src/index.tsx"></script>
     <dialog id="fallback-error">
       <p>The Playwright Trace Viewer must be loaded over the <code>http://</code> or <code>https://</code> protocols.</p>
-      <p>For more information, please see the <a href="https://aka.ms/playwright/err-trace-viewer-file-protocol">docs</a>.</p>
+      <p>For more information, please see the <a href="https://aka.ms/playwright/trace-viewer-file-protocol">docs</a>.</p>
     </dialog>
     <script>
       if (!/^https?:/.test(window.location.protocol))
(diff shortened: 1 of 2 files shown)


packages/trace-viewer/index.html
```diff
@@ -26,5 +26,12 @@
   <body>
     <div id="root"></div>
     <script type="module" src="/src/index.tsx"></script>
+    <dialog id="fallback-error" open="false">
+      <p>The Playwright Trace Viewer must be loaded over the <code>http://</code> or <code>https://</code> protocols.</p>
+    </dialog>
+    <script>
+      if (!/^https?:/.test(window.location.protocol))
+        document.getElementById("fallback-error").open = "true";
+    </script>
```
rwoll, line 35: We can probably be more helpful and make suggestions of running either `npx playwright show-report`, `npx playwright show-trace`, using `trace.playwright.dev`, or running/hosting on static server.

Additionally, we could inline the `index.tsx` and re-use the logic/components in there. The current issue is, `index.tsx` is violates CORS policy.

packages/trace-viewer/index.html
```diff
@@ -26,5 +26,12 @@
   <body>
     <div id="root"></div>
     <script type="module" src="/src/index.tsx"></script>
+    <dialog id="fallback-error">
+      <p>The Playwright Trace Viewer must be loaded over the <code>http://</code> or <code>https://</code> protocols.</p>
+    </dialog>
+    <script>
+      if (!/^https?:/.test(window.location.protocol))
+        document.getElementById("fallback-error").show();
+    </script>
```
rwoll, line 35: We can probably be more helpful—although there's a few ways to get here. Realistically, telling people to run `npx playwright show-report` would probably be the most helpful or running/hosting through static server.

packages/trace-viewer/index.html
```diff
@@ -26,5 +26,12 @@
   <body>
     <div id="root"></div>
     <script type="module" src="/src/index.tsx"></script>
+    <dialog id="fallback-error">
```
mxschmitt, line 29: Why not move it into the index.tsx or below?

rwoll: If it's in `index.tsx` it would never run. `index.tsx` is bundled as an external file (`main.[hash].js`) which is blocked when opening the trace viewer app itself with the `file://` protocol:

Screen Shot 2022-08-08 at 4 25 39 PM

The Trace Viewer does not inline all assets unlike the HTML Report which—except for the traceviewer—inlines its bundle into the file instead of having it be external.

packages/trace-viewer/index.html
debs-obrien: yes adding documentation in here as a message would be fantastic. Telling them why and how to solve it so  and linking to the docs page would be great

packages/trace-viewer/index.html
```diff
@@ -26,5 +26,13 @@
...
+      <p>The Playwright Trace Viewer must be loaded over the <code>http://</code> or <code>https://</code> protocols.</p>
+      <p>For more information, please see the <a href="https://aka.ms/playwright/err-trace-viewer-file-protocol">docs</a>.</p>
```
rwoll, line 31: @mxschmitt what do you think about this link? In the interest of time, since we don't think folks will likely end up on this error anyway, we can also just omit the second paragraph for now and let folks file and issue or ask for help. (It at least won't be a blank screen any more and has minimal copy.)

tests/playwright-test/reporter-html.spec.ts
```diff
@@ -473,6 +474,37 @@ test('should show multi trace source', async ({ runInlineTest, page, server, sho
   await expect(page.locator('.source-line-running')).toContainText('request.get');
 });
 
+test('should warn user when viewing via file:// protocol', async ({ runInlineTest, page, showReport }, testInfo) => {
```
rwoll, line 477: Added tests since—it would be pretty unexpected if the dialog showed up when it wasn't supposed to!

rwoll (commented):
@mxschmitt Can you take another look?

packages/trace-viewer/index.html
mxschmitt:
```suggestion
      <p>For more information, please see the <a href="https://aka.ms/playwright/trace-viewer-file-protocol">docs</a>.</p>
```
lgtm

tests/playwright-test/reporter-html.spec.ts
```diff
@@ -473,6 +474,37 @@ test('should show multi trace source', async ({ runInlineTest, page, server, sho
...
+    `,
+  }, { reporter: 'dot,html' }, { PW_TEST_HTML_REPORT_OPEN: 'never' });
+  expect(result.exitCode).toBe(0);
+  expect(result.passed).toBe(1);
+
+  await test.step('view via server', async () => {
+    await showReport();
+    await page.locator('[title="View trace"]').click();
+    await expect(page.locator('body')).toContainText('Action does not have snapshots', { useInnerText: true });
+    await expect(page.locator('dialog')).not.toBeVisible();
```
mxschmitt, line 496:
```suggestion
    await expect(page.locator('dialog')).toBeHidden();
```

mxschmitt (approved):

PR #16382, opened 2022-08-09
feat: introduce docker integration

aslushnikov:
This patch introduces the following commands:
- `npx playwright install docker-image` that builds a VRT docker image
  locally that is based off the `mcr.microsoft.com/playwright:focal`
- `npx playwright start-docker` that launches a docker container with
  browsers.
- `npx playwright stop-docker` that stops given docker container.
- `npx playwright test --docker` that runs all the tests inside a
  launched docker container.


rwoll:
@aslushnikov When we add docs, make sure we tell people to increase default CPU and mem limit on their docker install (https://playwright.slack.com/archives/C01AG7QGW2F/p1659463173880389?thread_ts=1659428827.669649&cid=C01AG7QGW2F) otherwise they'll likely hit unexpected perf issues if they have a freshly installed Docker-for-Desktop.

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,282 @@
...
+  RepoDigests: null | string[];
+  RepoTags: null | string[];
+  SharedSize: number;
+  Size: number;
+  VirtualSize: number;
+}
+
+const PW_CONTAINER_NAME = 'playwright-' + getPlaywrightVersion();
+const BASE_IMAGE_NAME = getPlaywrightVersion().includes('next') ?
+    'mcr.microsoft.com/playwright:next-focal' :
```
mxschmitt, line 36: use jammy so we save 2 years with updating?

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,282 @@
...
+    // 1. Pull base docker image
+    console.log(`Pulling ${BASE_IMAGE_NAME}...`);
+    await callDockerAPI('post', `/images/create?fromImage=${encodeURIComponent(BASE_IMAGE_NAME)}`);
+  }
+  // 2. Find pulled docker image
+  const dockerImage = await findDockerImage(baseImageName);
+  if (!dockerImage)
+    throw new Error(`Failed to pull ${baseImageName}`);
+  // 3. Launch container and install VNC in it
+  const NOVNC_REF = '1.2.0';
```
mxschmitt, line 55:
```suggestion
  const NOVNC_REF = '1.3.0';
```

rwoll:
Can you add user docs (either directly here or in another PR) and then I'll review both docs and impl in one?

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,386 @@
...
+      [end]
+      [include] (/etc/X11/fluxbox/fluxbox-menu)
+    [end]
+  \`);
+`;
+
+const CONTAINER_ENTRY_POINT = `#!/bin/bash
+  set -e
+  SCREEN_WIDTH=1360
+  SCREEN_HEIGHT=1020
```
mxschmitt, line 58: just curious, why these numbers?

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,386 @@
...
+  chmod 755 /start.sh
+
+  export DEBIAN_FRONTEND=noninteractive
+
+  # Install FluxBox, VNC & noVNC
+  mkdir -p /opt/bin && chmod +x /dev/shm \
+      && apt-get update && apt-get install -y unzip fluxbox x11vnc \
+      && curl -L -o noVNC.zip "https://github.com/novnc/noVNC/archive/v${NOVNC_REF}.zip" \
+      && unzip -x noVNC.zip \
+      && mv noVNC-${NOVNC_REF} /opt/bin/noVNC \
```
mxschmitt, line 103:
```suggestion
      && rm -rf noVNC-${NOVNC_REF}/{docs,tests} \
      && mv noVNC-${NOVNC_REF} /opt/bin/noVNC \
```

not sure if we should!

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,386 @@
...
+      && cp /opt/bin/noVNC/vnc.html /opt/bin/noVNC/index.html \
+      && rm noVNC.zip \
+      && curl -L -o websockify.zip "https://github.com/novnc/websockify/archive/v${WEBSOCKIFY_REF}.zip" \
+      && unzip -x websockify.zip \
+      && rm websockify.zip \
```
mxschmitt, line 108:
```suggestion
      && rm websockify.zip \
      && rm -rf  websockify-${WEBSOCKIFY_REF}/{docs,tests} \
```

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,386 @@
...
+
+export async function ensureDockerEngineIsRunningOrDie() {
+  try {
+    await callDockerAPI('get', '/info');
+  } catch (e) {
+    console.error(utils.wrapInASCIIBox([
+      `Docker is not running!`,
+      `Please install if necessary and launch docker first:`,
+      ``,
+      `    https://docs.docker.com/`,
```
mxschmitt, line 271:
```suggestion
      `    https://docs.docker.com/get-docker/`,
```

packages/playwright-core/src/cli/docker.ts
aslushnikov: sure why not!

packages/playwright-core/src/cli/docker.ts
aslushnikov: Sure why not!

docs/src/docker.md
```diff
@@ -145,3 +145,44 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
 ```
 docker run --rm -it playwright:localbuild /bin/bash
 ```
+
+## (Experimental) Playwright Test Docker Integration
```
mxschmitt, line 149: This paragraph should be js only?

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,389 @@
...
+  Id: string;
+}
+
+async function findRunningDockerContainerId(): Promise<string|undefined> {
+  const containers: (Container[]|undefined) = await getJSON('/containers/json');
+  if (!containers)
+    return undefined;
+  // 1. Try findind a container with our name. This happens if the container is launched
+  // automatically with `npx playwright docker start`.
+  let container = containers.find((container: Container) => container.Names.some(name => name.includes(VRT_CONTAINER_NAME)));
```
dgozman, line 299: Might as well remove this case?

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,389 @@
...
+interface ContainerOptions {
+  image: DockerImage;
+  autoRemove: boolean;
+  command?: string[];
+  ports?: Number[];
+  name?: string;
+}
+
+async function launchContainer(options: ContainerOptions): Promise<string> {
+  const ExposedPorts: any = {};
```
dgozman, line 318: Follow-up ideas:
- `docker stop` and `docker delete-image` should stop all versions of pw image;
- bind to available ports on the host instead of 5400 and 7900;

aslushnikov:
Note for self: also add docker desktop recommended settings that @rwoll recommended

docs/src/docker.md
```diff
@@ -145,3 +145,45 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
...
+* langs: js
+
+Playwright Test now ships an **experimental** docker integration.
```
mxschmitt, line 152:
```suggestion
Playwright Test now ships an **experimental** Docker integration.
```

docs/src/docker.md
```diff
@@ -145,3 +145,45 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
...
+With this integration, **only** browser binaries are running inside a docker container,
```
mxschmitt, line 153:
```suggestion
With this integration, browser binaries are running **only** inside a docker container,
```

docs/src/docker.md
```diff
@@ -145,3 +145,45 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
...
+while all the code is still running on the host operating system.
+
+Docker container provides a consistent environment, eliminating browser rendering
+differences across platforms. Playwright Test will automatically proxy host network
```
mxschmitt, line 157:
```suggestion
differences across platforms. Playwright Test will automatically proxy host network traffic
```

docs/src/docker.md
```diff
@@ -145,3 +145,45 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
...
+
+:::note
+Docker integration requires Docker installed & running on your computer.
+See https://docs.docker.com/get-docker/
+:::
+
+Docker integration usage:
+
+1. Build a local docker image that will be used to run containers. This step
+   should be done only once.
```
mxschmitt, line 168:
```suggestion
   needs to be done only once.
```

needs seems to fit better?

docs/src/docker.md
```diff
@@ -145,3 +145,45 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
...
+   ```bash js
+   npx playwright docker build
+   ```
+
+2. Run docker container in the background.
```
mxschmitt, line 173:
```suggestion
2. Run Docker container in the background.
```

docs/src/docker.md
```diff
@@ -145,3 +145,45 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
...
+   ```bash js
+   npx playwright docker start
+   ```
+
+3. Run tests inside docker container. Note that this command accepts all the same arguments
```
mxschmitt, line 178:
```suggestion
3. Run tests inside Docker container. Note that this command accepts all the same arguments
```

docs/src/docker.md
```diff
@@ -145,3 +145,45 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
...
+   as a regular `npx playwright test` command.
+   ```bash js
+   npx playwright docker test
+   ```
+
+   Note that this command will detect running docker container, and auto-launch it if needed.
+
+4. Finally, stop docker container when it is no longer needed.
```
mxschmitt, line 186:
```suggestion
4. Finally, stop Docker container when it is no longer needed.
```

docs/src/docker.md
```diff
@@ -145,3 +145,45 @@ The image will be tagged as `playwright:localbuild-focal` and could be run as:
...
+   ```bash js
+   npx playwright docker stop
+   ```
```
mxschmitt, line 189: I feel like we should mention here the new "ignore-snapshots" CLI and config option, which is helpful to be used in parallel when running them locally.

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,380 @@
...
+
+const DOCKER_API_VERSION = '1.41';
+
+function callDockerAPI(method: 'post'|'get'|'delete', url: string, body: Buffer|string|undefined = undefined): Promise<string> {
+  const dockerSocket = process.platform === 'win32' ? '\\\\.\\pipe\\docker_engine' : '/var/run/docker.sock';
+  return new Promise((resolve, reject) => {
+    const request = http.request({
+      socketPath: dockerSocket,
+      path: `/v${DOCKER_API_VERSION}${url}`,
+      method,
```
mxschmitt, line 354:
```suggestion
      timeout: NET_DEFAULT_TIMEOUT
      method,
```
import it from this file: https://github.com/mxschmitt/playwright/blob/fix-doclint/packages/playwright-core/src/common/netUtils.ts#L42

packages/playwright-core/src/cli/docker.ts
```diff
@@ -0,0 +1,380 @@
...
+    });
+    request.on('error', function(e){
+      reject(e);
+    });
+    if (body) {
+      request.setHeader('Content-Type', 'application/json');
+      request.setHeader('Content-Length', body.length);
+      request.write(body);
+    } else {
+      request.setHeader('Content-Type', 'text/plain');
```
mxschmitt, line 375: why set the content type if there is no body?

utils/docker/Dockerfile.jammy
```diff
@@ -25,8 +25,8 @@ ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
 
 # 1. Add tip-of-tree Playwright package to install its browsers.
 #    The package should be built beforehand from tip-of-tree Playwright.
-COPY ./playwright-core.tar.gz /tmp/playwright-core.tar.gz
 
+COPY ./playwright-core.tar.gz /tmp/playwright-core.tar.gz
 # 2. Bake in browsers & deps.
```
mxschmitt, line 30:
```suggestion
COPY ./playwright-core.tar.gz /tmp/playwright-core.tar.gz

# 2. Bake in browsers & deps.
```

revert changes in this file.

mxschmitt (approved):

tests/installation/docker-integration.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+test.skip(() => process.env.CI && process.platform !== 'linux');
+
+test.beforeAll(async ({ exec }) => {
+  // Delete any previous docker image to ensure clean run.
+  await exec('npx playwright docker delete-image', {
+    cwd: path.join(__dirname, '..', '..'),
+  });
+});
+
+test('make sure it tells to run `npx playwright docker build` when image is not instaleld', async ({ exec }) => {
```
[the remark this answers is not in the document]
aslushnikov: This is one of the follow-ups.

packages/playwright-core/src/cli/docker.ts
aslushnikov: This is to get plain response

dgozman (approved):

docs/src/docker.md
aslushnikov: I'd disagree with this change since it changes the semantics of the sentence!

My point was to highlight that unlike the regular docker approach that mounts the whole project into docker & executes all tests & test runner & browser binaries inside, this approach keeps test runner on the host system, while only launching browsers inside the container.

packages/playwright-core/src/cli/docker.ts
aslushnikov: Let me hard code 30 seconds timeout here so that I don't need to export more things from `playwright-core` (the `docker.ts` file got moved into playwright-test)

docs/src/docker.md
aslushnikov: Both docker and ignore-snapshots are parts of Visual Regression Testing story that's being baked. We'll gradually get there!

tests/installation/docker-integration.spec.ts
mxschmitt, line 32:
```suggestion
test('make sure it tells to run `npx playwright docker build` when image is not installed', async ({ exec }) => {
```

PR #16400, opened 2022-08-10
fix(codegen): make sure input recording with japanese IME Work

mxschmitt:
As per https://github.com/microsoft/playwright/pull/16210#issuecomment-1210263618 reland with formatting fixes.


mxschmitt:
Merging without approval since already lgtm on the actual PR by Yury.

PR #16419, opened 2022-08-10
devops: add cherry-pick bot

mxschmitt:


rwoll (commented):
This looks fine, but I'm not sure it saves us much time and if it breaks, it's hard to debug. In any given release, we individually don't CP too much, and the commands to do it are minimal and straightforward.

That being said, we can use this action or manually do the CP, so seems fine to land.

mxschmitt:
TBR Ross.

PR #16431, opened 2022-08-10
test: unflake 'should fill japanese text'

mxschmitt:


yury-s (approved):
I wonder why the old way is more flaky, do you have a clue?

Issue #16458, opened 2022-08-11
[BUG] Breaking Change in v1.25.0 release from v1.24.2


audiolion:
https://github.com/microsoft/playwright/compare/v1.24.2...v1.25.0#diff-6070243a9a8ea57709af102355d6cc733767fd59e931c354efc75f47114f2349L30

The exported function used to be `addTestCommand`, now that has been made private, and a new function `addTestCommands` is exported, so code relying on the exported API surface is breaking on the new version.

The fix is I guess to change to use `addTestCommands` but this is not documented in the release notes, not sure if there are any differences we should be aware of.

rwoll:
Can you include your package.json and package-lock.json? Thanks! I will be looking at this issue later, but this info will possibly be helpful here.

Leaving triaging label off for now until I circle back to look into this. Thanks!

mxschmitt:
This usually happens when playwright-core and @olaywright/test have different versions. npm ci/npm i after aligning the versions should fix it.

rwoll:
> Maybe I am misunderstanding something but my package.json / lock are irrelevant. 

This request is to reduce the debug loop on our end if it becomes relevant when we triage this issue later today. PW has a few packages that are part of it.

Sorry you're facing issues and thanks for hunting down the commit! I'll be looking into this in detail later today when triaging.

mxschmitt:
> The issue arises because the recommendation is to do npx playwright test which always will grab the latest playwright core

This is not the case, `@playwright/test` will expose a bin called `playwright` so if you do `npx playwright test` it will automatically go to the `@playwright/test` NPM module which will internally call `playwright-core`. `@playwright/test`
 depends strictly on `playwright-core` with a strong version match.

We consider `addTestCommand` an implementation detail, it's not an official API which is exposed somewhere. We agree that the current behaviour is not great as well, so we should most likely add version validation to throw in the beginning if the versions diverge like afaik babel or other projects are doing.

audiolion:
> We consider addTestCommand an implementation detail, it's not an official API which is exposed somewhere. 

[...]

PR #16475, opened 2022-08-11
chore: validate pw-core version when pwtest CLI gets executed

mxschmitt:
https://github.com/microsoft/playwright/issues/16458

### DIFF
--- a/packages/playwright-core/src/cli/cli.ts
+++ b/packages/playwright-core/src/cli/cli.ts
@@ -309,13 +309,28 @@ Examples:
 
 if (!process.env.PW_LANG_NAME) {
   let playwrightTestPackagePath = null;
+  const resolvePwTestPaths = [__dirname, process.cwd()];
   try {
     playwrightTestPackagePath = require.resolve('@playwright/test/lib/cli', {
-      paths: [__dirname, process.cwd()]
+      paths: resolvePwTestPaths,
     });
   } catch {}
 
   if (playwrightTestPackagePath) {
+    const pwTestVersion = require(require.resolve('@playwright/test/package.json', {
+      paths: resolvePwTestPaths,
+    })).version;
+    const pwCoreVersion = require(path.join(__dirname, '../../package.json')).version;
+    if (pwTestVersion !== pwCoreVersion) {
+      console.error(wrapInASCIIBox([
+        `Playwright Test compatibility check failed:`,
+        `@playwright/test version '${pwTestVersion}' does not match playwright-core version '${pwCoreVersion}'!`,
+        `To fix this either align the Playwright versions or only keep @playwright/test since it depends on playwright-core.`,
+        `If you still receive this error, execute 'npm ci' or delete 'node_modules' and do 'npm install' again.`,
+      ].join('\n')));
+      process.exit(1);
+    }
+
     require(playwrightTestPackagePath).addTestCommands(program);
   } else {
     {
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/cli/cli.ts
```diff
@@ -309,13 +309,28 @@ Examples:
...
+      paths: resolvePwTestPaths,
+    })).version;
+    const pwCoreVersion = require(path.join(__dirname, '../../package.json')).version;
+    if (pwTestVersion !== pwCoreVersion) {
+      console.error(wrapInASCIIBox([
+        `Playwright Test compatibility check failed:`,
+        `@playwright/test version '${pwTestVersion}' does not match playwright-core version '${pwCoreVersion}'!`,
+        `To fix this either align the Playwright versions or only keep @playwright/test since it depends on playwright-core.`,
+        `If you still receive this error, execute 'npm ci' or delete 'node_modules' and do 'npm install' again.`,
+      ].join('\n')));
```
rwoll, line 330: What are some scenarios the user ends up here? Like how do they end up with mismatched versions in the first place. It happens, and I'd like to understand why since maybe we can catch it on the other end, too.

mxschmitt: Most common scenario is if they have pw-core 1.24 and pwt 1.25

Maybe also playwright the package 1.24 and pwt 1.25.

rwoll: That's the issue, but it doesn't explain how they end up with pw-core 1.24 and pwt 1.25. Like what sequence of command would have had to be run to get into this situation?

mxschmitt: > That's the issue, but it doesn't explain how they end up with pw-core 1.24 and pwt 1.25. Like what sequence of command would have had to be run to get into this situation?

In both my my examples I imply that they have these in the package.json.

rwoll (commented):
I think this looks good—but can you please add an installation test to make sure this fires when we want it to, and doesn't fire when we don't want it too?

I think you can copy-and-paste an existing test I've added.

mxschmitt:
Status update: We are waiting for user's responses to understand why they run into this actual issue.

packages/playwright-core/src/cli/cli.ts
```diff
@@ -309,13 +309,28 @@ Examples:
...
+  const resolvePwTestPaths = [__dirname, process.cwd()];
   try {
     playwrightTestPackagePath = require.resolve('@playwright/test/lib/cli', {
-      paths: [__dirname, process.cwd()]
+      paths: resolvePwTestPaths,
     });
   } catch {}
 
   if (playwrightTestPackagePath) {
+    const pwTestVersion = require(require.resolve('@playwright/test/package.json', {
```
pavelfeldman, line 320: Site note: all of it isn't quite right it seems. Consider the case with the global installation of playwright, that led to the global installation of playwright-core, and a local installation of @playwright-test. Our logic here does not work then.

packages/playwright-core/src/cli/cli.ts
```diff
@@ -309,13 +309,28 @@ Examples:
...
+      paths: resolvePwTestPaths,
+    })).version;
+    const pwCoreVersion = require(path.join(__dirname, '../../package.json')).version;
```
pavelfeldman, line 323: This should check should go the other way around, it should be in @playwright/test that upon initialization checks its dependency.

pavelfeldman (approved):

mxschmitt:
For further questions, how to reproduce it. It does not reproduce if you have `@playwright/test` and `playwright`/`playwright-core` inside your package.json. You need to have a pwt version first, and then install pw/core so it will override the .bin entry.

Issue #15934, opened 2022-07-25
[BUG] `page.locator.isVisible` returns `true` on hidden element


DetachHead:
**Context:**
## System:
 - OS: Windows 10 10.0.19044
 - Memory: 926.38 MB / 15.78 GB
## Binaries:
 - Node: 18.0.0 - C:\Program Files\nodejs\node.EXE
 - npm: 8.7.0 - C:\Program Files\nodejs\npm.CMD   
## Languages:
 - Bash: 5.0.17 - C:\WINDOWS\system32\bash.EXE    
## npmPackages:
 - playwright: ^1.23.0 => 1.23.0

**Code Snippet**

```javascript
test('test', async ({ page }) => {
    await page.goto('C:\\page.html')
    const selector = '.sapUiLocalBusyIndicatorAnimStandard'
    const isVisible = await page.locator(selector).isVisible()
    console.log(isVisible) // true
    await page.waitForSelector(selector, {state: 'hidden'}) // success
  })
```

**Describe the bug**
this SAPUI5 busy indicator is not visible, and `page.waitForSelector(selector, {state: 'hidden'})` succeeds, however `page.locator.isVisible` incorrectly returns `true`. the bounding box seems to also be empty:

![image](https://user-images.githubusercontent.com/57028336/180892867-b4600c22-d49d-4aed-84a4-e401ba6d3286.png)
![image](https://user-images.githubusercontent.com/57028336/180892830-4fcbdabf-733f-4238-9bc0-d61713e897d2.png)

unfortunately i'm not able to share the full html as it's for an internal site, though i'm open to suggestions to try and provide a minimal example.

pavelfeldman:
I suspect that at the moment when the check was made, it was visible, but then it became hidden. It explains everything you observe.

DetachHead:
@pavelfeldman i thought so too, but this was tested on a static html file that I saved from a trace file. Nothing on the page is Changing at all here

pavelfeldman:
We would need exact steps to follow in order to address this.

DetachHead:
@pavelfeldman 

ok i was able to reproduce it with the example elements on the sapui5 docs

```ts
test('test', async ({ page }) => {
    await page.goto('https://sapui5.hana.ondemand.com/1.98.0/#/entity/sap.m.PullToRefresh/sample/sap.m.sample.PullToRefresh')
    await page.click('"Reject All"')
    await page.waitForSelector("span:text-is('Refresh')")
[...]

PR #16480, opened 2022-08-11
fix(actions): pass retarget behavior matching the action

dgozman:
We used to determine retarget behavior in various places, and for some APIs it was not determined correctly. Now every API method explicitly passes the desired retarget behavior through helper methods to avoid surprises.

- Fixes retarget behavior from "follow label and enclosing button" to "follow label" for the following methods: `toHaveValue`, `toHaveValues`, `fill`, `selectOptions`, `inputValue`, `setInputFiles`, `toBeEditable`, `isEditable`, `selectText`, `waitForElementState('editable')`.

- Fixes `scrollIntoView` retarget from "enclosing button" to "none".

- Does not fix `check`/`setChecked` retarget behavior, leaving it partially "follow label and enclosing button", and partially "enclosing button or link". Fixing this requires more changes.

References #15934.

### DIFF
--- a/packages/playwright-core/src/server/dom.ts
+++ b/packages/playwright-core/src/server/dom.ts
@@ -20,7 +20,7 @@ import type * as channels from '../protocol/channels';
 import { isSessionClosedError } from './protocolError';
 import type { ScreenshotOptions } from './screenshotter';
 import type * as frames from './frames';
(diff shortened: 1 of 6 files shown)


pavelfeldman (approved):

pavelfeldman:
GC

PR #16481, opened 2022-08-11
fix(ct): pass local config to preview

### LINKED ISSUE #16460: [BUG] Component test fails on 2nd run when SSL is used
**Context:**
- Playwright Version: `"playwright/experimental-ct-svelte": "^1.24.2"`
- Operating System: macOS
- Node.js version: 16.14.2
- Browser: All
- Extra: ARM chip device

**Code Snippet**

Here's the reduced test case which reproduces the issue:
https://github.com/vhscom/playwright-ct-ssl-reduced-test

I've left the reduced test case configured with the Vite plugin for SSL but the issue can be reproduced without plugins specified in the Playwright component config file.

**Describe the bug**

This is an extension of https://github.com/microsoft/playwright/issues/16424. When testing a Svelte component using `@vitejs/plugin-basic-ssl` plugin test runner fails with error message when cache is present and unless the component under test has a timestamp modification (read: it was saved, even without changes):

```
undefined: net::ERR_EMPTY_RESPONSE at http://localhost:3100/playwright/index.html
navigating to "http://localhost:3100/playwright/index.html", waiting until "load"
```

I narrowed it down to a TLS/vite config/playwright lib corner case. It seems the URL under test does not work when the `@vitejs/plugin-basic-ssl` plugin is used—but only on the second run if the component under test doesn't have a timestamp modification since the last cached run.

Ideally tests will either fail wholesale if SSL is configured or there will be a way to run component tests over the `https` scheme—which would be expected when the plug-in is configured via Playwright experimental `ctViteConfig` prop. Having test runs once, fail on the second run, then work again when the cache is cleared or the component timestamp changed is unexpected behavior.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/16460

### DIFF
--- a/packages/playwright-test/src/plugins/vitePlugin.ts
+++ b/packages/playwright-test/src/plugins/vitePlugin.ts
@@ -110,32 +110,37 @@ export function createPlugin(
         };
       }
       const { build, preview } = require('vite');
-      if (sourcesDirty) {
-        viteConfig.plugins = viteConfig.plugins || [
-          frameworkPluginFactory()
-        ];
+      // Build config unconditionally, either build or build & preview will use it.
+      viteConfig.plugins = viteConfig.plugins || [
+        frameworkPluginFactory()
+      ];
+      // But only add out own plugin when we actually build / transform.
+      if (sourcesDirty)
         viteConfig.plugins.push(vitePlugin(registerSource, relativeTemplateDir, buildInfo, componentRegistry));
-        viteConfig.configFile = viteConfig.configFile || false;
-        viteConfig.define = viteConfig.define || {};
-        viteConfig.define.__VUE_PROD_DEVTOOLS__ = true;
-        viteConfig.css = viteConfig.css || {};
-        viteConfig.css.devSourcemap = true;
-        viteConfig.build = {
-          ...viteConfig.build,
-          target: 'esnext',
-          minify: false,
-          rollupOptions: {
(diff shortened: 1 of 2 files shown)


rwoll (approved):
Bump your

```
const VERSION = 6;
```

to `7` for good measure?

rwoll:
Shall we also add a test?

dgozman (approved):

dgozman (approved):

packages/playwright-core/src/cli/driver.ts
```diff
@@ -197,10 +197,16 @@ class ProtocolHandler {
...
+  id: number;
+  method: string;
+  params: Record<string, any>
+};
+
 function wireController(playwright: Playwright, wsEndpoint: string) {
   process.send!({ method: 'ready', params: { wsEndpoint } });
   const handler = new ProtocolHandler(playwright);
-  process.on('message', async message => {
+  process.on('message', async (message: DriverProtocolMessage) => {
```
rwoll, line 209: Is this and the @types/node change needed in this PR?

rwoll (approved):
LGTM! Thanks for adding the test!

packages/playwright-core/src/cli/driver.ts
rwoll: Ignore—we chatted offline

PR #16488, opened 2022-08-12
support array headers (eg: set-cookie)

kadirgun:
If the headers option is entered as an array in the route.fulfill function, it will parse all the headers entered.

Example:
```
route.fulfill({
  body: JSON.stringify({message: 'ok'}),
  contentType: 'application/json',
  status: 200,
  headers: {
    'Content-Type': 'application/json',
    'Content-Length': 16,
    'Set-Cookie': [
      'session=xxxxxxx; expires=Fri, 12 Aug 2055 00:00:00 GMT; Max-Age=7200; path=/; httponly; samesite=lax',
      'xsrf-token=xxxxxxx; expires=Fri, 12 Aug 2055 00:00:00 GMT; Max-Age=7200; path=/; httponly; samesite=lax',
    ]
  }
})
```

### DIFF
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -342,7 +342,7 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
 
     const headers: Headers = {};
     for (const header of Object.keys(headersOption || {}))
-      headers[header.toLowerCase()] = String(headersOption![header]);
+      headers[header.toLowerCase()] = Array.isArray(headersOption![header]) ? headersOption![header] : String(headersOption![header]);
     if (options.contentType)
       headers['content-type'] = String(options.contentType);
     else if (options.path)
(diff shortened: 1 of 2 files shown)


okanfidan (approved):
This works great. I think it should be updated for everyone.

packages/playwright-core/src/client/network.ts
```diff
@@ -342,7 +342,7 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
 
     const headers: Headers = {};
     for (const header of Object.keys(headersOption || {}))
-      headers[header.toLowerCase()] = String(headersOption![header]);
+      headers[header.toLowerCase()] = Array.isArray(headersOption![header]) ? headersOption![header] : String(headersOption![header]);
```
pavelfeldman, line 345: That's not how multi-value headers are concatenated. Please start with filing a bug and we'll help you navigate it!

PR #16494, opened 2022-08-12
chore: add id for each test to JSONReporter

bezyakina:
Since there is a [problem](https://github.com/microsoft/playwright/issues/14310) in the release 1.25 was partially solved, then I decided to try adding the test id to the json reporter on my own

I couldn't run the tests locally, because during the build, my edits in the JSONReportSpec interface were [overwritten](https://github.com/microsoft/playwright/runs/7805545343?check_suite_focus=true#step:6:145)


mxschmitt:
You need to add the "id" there: https://github.com/microsoft/playwright/blob/e725b5f8a9787107c04609f4dd0ccca6a998d3c9/utils/generate_types/overrides-testReporter.d.ts#L78 and then run "npm run build" / generate types then it should not be overriden anymore.

pavelfeldman (approved):

PR #16530, opened 2022-08-15
devops: fix firefox builds

aslushnikov:
Turns out Firefox mach does not respect `DEBIAN_FRONTEND=noninteractive`
command and insists on interactive input for the mozbuild folder
creation.

Pre-creation of the folder avoid this codepath.


aslushnikov:
TBR @mxschmitt

PR #16536, opened 2022-08-15
feat(webkit): roll to r1701

playwrightmachine:


mxschmitt:
```
page/page-set-content.spec.ts:126:1 › should return empty content there is no iframe src 
page/page-wait-for-load-state.spec.ts:184:1 › should work with javascript: iframe ===
```

these two fail across all the platforms.

yury-s:
The failures is a regression introduced by https://github.com/WebKit/WebKit/commit/7de67f9a0ccc4fc8ce195e32f61e7873ad81345e

yury-s:
> ```
> page/page-set-content.spec.ts:126:1 › should return empty content there is no iframe src 
>```
I'm going to mark the test as `fixme` in webkit as it's already marked so in the other two browsers for the same reason: there is not utility context.

> ```
> page/page-wait-for-load-state.spec.ts:184:1 › should work with javascript: iframe ===
> ```
> 
This one should be fixed by https://github.com/microsoft/playwright/pull/16591

yury-s:
Superseded by https://github.com/microsoft/playwright/pull/16603

PR #16525, opened 2022-08-13
chore(ct): react vite tests

sand4rt:
added tests and changed the structure, similar to the other ct projects

PR #16549, opened 2022-08-15
feat(ct): react rerender

sand4rt:
https://github.com/microsoft/playwright/pull/16525 needs to be merged first


packages/playwright-test/src/mount.ts
```diff
@@ -60,25 +60,41 @@ export const fixtures: Fixtures<
...
-  let component: Component;
-  if (typeof jsxOrType === 'string')
-    component = { kind: 'object', type: jsxOrType, options };
-  else
-    component = jsxOrType;
+async function innerRerender(page: Page, jsxOrType: JsxComponent | string, options: Omit<MountOptions, 'hooksConfig'> = {}): Promise<void> {
+  const component = createComponent(jsxOrType, options);
+  wrapFunctions(component, page, boundCallbacksForMount);
+
+  await page.evaluate(async ({ component }) => {
```
pavelfeldman, line 76: It looks like we have a nice code reuse opportunity here, bu we can follow up.

pavelfeldman (approved):

pavelfeldman:
Could you re-push it so that we could maintain your authorship?

sand4rt:
> Could you re-push it so that we could maintain your authorship?

@pavelfeldman see: https://github.com/microsoft/playwright/pull/16560

PR #16565, opened 2022-08-16
docs: merge emulation with parameterize

debs-obrien:


mxschmitt:
decided to keep it for now as it is.

PR #16584, opened 2022-08-16
chore(ct): react tests

sand4rt:
align the folder structure + tests with the other ct projects


pavelfeldman (approved):

pavelfeldman:
Bots should be fixed, so i'll land

PR #16592, opened 2022-08-16
feat(chromium): use Element.checkVisibility in isElementVisible

josepharhar:
Element.checkVisibility is a new browser API that was shipped in
chromium 105:
https://bugs.chromium.org/p/chromium/issues/detail?id=1309533

Using checkVisibility accounts for the content-visibility:hidden in the
user-agent ShadowRoot of the details element, which means we can remove
the usage of the AutoExpandDetailsElementFlag (I am trying to remove the
flag in chromium).

This behavior is covered by the existing "isVisible and isHidden should
work with details" test in locator-convenience.spec.ts.

### DIFF
--- a/packages/playwright-core/src/server/injected/domUtils.ts
+++ b/packages/playwright-core/src/server/injected/domUtils.ts
@@ -75,11 +75,8 @@ export function isElementVisible(element: Element): boolean {
   // Element.checkVisibility checks for content-visibility and also looks at
   // styles up the flat tree including user-agent ShadowRoots, such as the
   // details element for example.
-  if (Element.prototype.checkVisibility) {
-    if (!element.checkVisibility()) {
-      return false;
-    }
-  }
+  if (Element.prototype.checkVisibility && !element.checkVisibility())
+    return false;
   const rect = element.getBoundingClientRect();
   return rect.width > 0 && rect.height > 0;
 }
--- a/packages/playwright-core/src/server/injected/domUtils.ts
+++ b/packages/playwright-core/src/server/injected/domUtils.ts
@@ -60,8 +60,6 @@ export function isElementVisible(element: Element): boolean {
   if (!element.ownerDocument || !element.ownerDocument.defaultView)
     return true;
   const style = element.ownerDocument.defaultView.getComputedStyle(element);
-  if (!style || style.visibility === 'hidden')
-    return false;
   if (style.display === 'contents') {
     // display:contents is not rendered itself, but its child nodes are.
     for (let child = element.firstChild; child; child = child.nextSibling) {
@@ -76,7 +74,9 @@ export function isElementVisible(element: Element): boolean {
   // styles up the flat tree including user-agent ShadowRoots, such as the
   // details element for example.
   // @ts-ignore Typescript doesn't know that checkVisibility exists yet.
-  if (Element.prototype.checkVisibility && !element.checkVisibility())
+  if (Element.prototype.checkVisibility && !element.checkVisibility({ checkOpacity: false, checkVisibilityCSS: false }))
+    return false;
+  if (!style || style.visibility === 'hidden')
     return false;
   const rect = element.getBoundingClientRect();
   return rect.width > 0 && rect.height > 0;
--- a/packages/playwright-core/src/server/injected/domUtils.ts
+++ b/packages/playwright-core/src/server/injected/domUtils.ts
@@ -75,6 +75,7 @@ export function isElementVisible(element: Element): boolean {
   // Element.checkVisibility checks for content-visibility and also looks at
   // styles up the flat tree including user-agent ShadowRoots, such as the
   // details element for example.
+  // @ts-ignore Typescript doesn't know that checkVisibility exists yet.
   if (Element.prototype.checkVisibility && !element.checkVisibility())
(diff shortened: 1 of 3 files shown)


JoelEinbinder:
Does passing a non existent feature flag to chromium cause it to break? It seems like we should keep passing the flag and just also do the `isElementVisible` check. That way it will keep working with old Chromium/Electron as well as new Chromium that has the flag removed.

josepharhar:
> Does passing a non existent feature flag to chromium cause it to break? It seems like we should keep passing the flag and just also do the isElementVisible check. That way it will keep working with old Chromium/Electron as well as new Chromium that has the flag removed.

Sounds good, I put the flag back in

pavelfeldman:
@josepharhar 👋
is there a spec for checkVisibility? our isVisible() is strictly defined and I'd like to make sure these definition match. In particular, elements are visible even though they are absolutely positioned outside the window bounds in Playwright (negative offset).

josepharhar:
> is there a spec for checkVisibility?

https://drafts.csswg.org/cssom-view-1/#dom-element-checkvisibility

> elements are visible even though they are absolutely positioned outside the window bounds in Playwright

Using checkVisibility like this should only consider the element as hidden if it's in a display:none subtree or if there is a content-visibility:hidden ancestor up the flat tree: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/element.cc;l=8688;drc=13ea55663d225cb9362af057035a2bce914d7d72
The opacity and visibility property checks don't happen by default: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/element.idl;l=28-31;drc=b5a3f9828f94b9a48b4e91107bbd874c947d2541

The existing getBoundingClientRect check should already consider display:none, so this patch should only make a difference when there is a content-visibility:hidden ancestor. Having a content-visibility:hidden ancestor will prevent the element from being rendered, similarly to display:none.

Alternatively, I could have tried to make content-visibility:hidden return a zero getBoundingClientRect in chromium, but there was some resistance to changing that so here we are.

pavelfeldman:
Oh, it looks exactly like we want it to be. Are there signals from other vendors? How likely do you think it'll be able to preserve its semantics?

pavelfeldman:
@josepharhar ^^

josepharhar:
Other browsers probably won't implement until they have content-visibility implemented which I don't think is done yet.

WebKit had no response, but they didn't object in the csswg as far as I can tell: https://lists.webkit.org/pipermail/webkit-dev/2022-May/032218.html

Mozilla had some feedback about the api shape which I implemented: https://github.com/mozilla/standards-positions/issues/634

I don't think that the api shape or semantics will change since there were plenty of opportunities for feedback which I implemented in the api before I shipped it:
https://groups.google.com/a/chromium.org/g/blink-dev/c/fvpRgcE8_Yg
https://github.com/w3c/csswg-drafts/issues/6850

packages/playwright-core/src/server/injected/domUtils.ts
```diff
@@ -72,6 +72,12 @@ export function isElementVisible(element: Element): boolean {
     }
     return false;
   }
+  // Element.checkVisibility checks for content-visibility and also looks at
+  // styles up the flat tree including user-agent ShadowRoots, such as the
+  // details element for example.
+  // @ts-ignore Typescript doesn't know that checkVisibility exists yet.
+  if (Element.prototype.checkVisibility && !element.checkVisibility())
```
dgozman, line 79: @josepharhar Sorry for the long delay!

I looked at the implementation and I think this is not 100% compatible, but pretty close to it.
- Browser looks at `opacity: 0`, while playwright does not. Can we pass `ignoreOpacity` option?
- Browser checks some kind of "inert" criteria - what is that?
- Browser checks that `visibility === visible`, while Playwright only checks for `hidden`. I think browser's behavior is actually better, but I'd like to make a separate change for that.

Other comments:
- We should probably move line 63 after this check, to not be redundant, wdyt?
- What about display:contents elements? Are there any plans to address those?

josepharhar: > Browser looks at opacity: 0, while playwright does not. Can we pass ignoreOpacity option?

> Browser checks that visibility === visible, while Playwright only checks for hidden. I think browser's behavior is actually better, but I'd like to make a separate change for that.

Opacity and visibility are not looked at by default: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/element.idl;l=29-30;drc=8b23a425b4a5e75caa5b91c78b062a38b1f92aea

I'll add the parameters and set them to false just to make it clearer though.

> Browser checks some kind of "inert" criteria - what is that?

The first iteration of checkVisibility had inert, but I actually removed it. It shouldn't exist anywhere.

> We should probably move line 63 after this check, to not be redundant, wdyt?

Done

> What about display:contents elements? Are there any plans to address those?

Hmm... I suppose that if an element it display:contents, then it shouldn't have a LayoutObject, which would make checkVisibility return false. The code earlier in this function already checks for display:contents though and won't continue to checkVisibility.

dgozman (approved):
Looks good, thank you for the PR!

PR #16631, opened 2022-08-18
docs(advanced): improve docs about global.d.ts

### LINKED ISSUE #16599: [Docs] Further explain global.d.ts usage for custom assertions
The docs section for [creating custom assertions](https://playwright.dev/docs/test-advanced#add-custom-matchers-using-expectextend) mentions a `global.d.ts` file, but doesn't give any information where it's located or where/whether it should be possibly created.  
So it's unclear to the reader what to do with the information currently given in the docs.  

In my case, I assumed the file would exist somewhere in the `node_modules` folder, but it doesn't. I only found a `globals.d.ts` under `node_modules/@types/`, but that doesn't seem to be it.  
And just placing the file in the Playwright root folder doesn't seem to work.  

So it would be great if this section of the docs could be expanded by adding the necessary information on where that file should be located and whether it already exists or the user should create that file.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/16599

### DIFF
--- a/docs/src/test-advanced-js.md
+++ b/docs/src/test-advanced-js.md
@@ -681,7 +681,9 @@ test('numeric ranges', () => {
 });
 ```
 
-For TypeScript, also add the following to `global.d.ts`. You don't need it for JavaScript.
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):

azad-derakhshani-GS:
Thanks Max, much appreciated! 🙏

PR #16647, opened 2022-08-18
chore: align more ESLint rules with VSCode formatting

mxschmitt:


rwoll (approved):
The more auto-fixable rules the better! I don't particularly care what the code gets fixed/formatted too, as long as we don't have to think about it. 😄 

🚢 it!

rwoll:
@mxschmitt When you land this, please immediately notify everyone since it may create some annoying merge conflicts for a second for pending work items. Thanks!

PR #16696, opened 2022-08-19
fix(trace): preserve line breaks in console messages

### LINKED ISSUE #16608: [BUG] Trace viewer does not display carriage returns when logging to the console using page.evaluate
**Context:**
- Playwright Version: 1.23.0
- Operating System: Windows
- Extra: Python

```
self.context.tracing.start(title=test_title, sources=True, screenshots=True, snapshots=True)
page.evaluate(f"console.error(String.raw`{traceback.format_exc()}`)")
context.tracing.stop()
```

When sending a log (or trace back in this instance) it shows normally in the browser. i.e with carriage returns
When viewing the console log in the trace viewer, it is just one line without carriage returns.

I have tried many, many things. But no carriage returns, so the formatting is broken.

I did check the trace.trace file and it appears that \ is replaced with \\\\ if that helps.
Formatting:
Pycharm console:
image

Browser:
image

Trace viewer:
image

yury-s:
Fixes #16608

### DIFF
--- a/packages/web/src/third_party/vscode/codicon.css
+++ b/packages/web/src/third_party/vscode/codicon.css
@@ -23,7 +23,7 @@
 	user-select: none;
 }
 
-.codicon-blank:before { content: '\81'; }
+.codicon-blank:before { content: '\2003'; }
 .codicon-add:before { content: '\ea60'; }
 .codicon-plus:before { content: '\ea60'; }
 .codicon-gist-new:before { content: '\ea60'; }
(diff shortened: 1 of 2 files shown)


packages/web/src/third_party/vscode/codicon.css
```diff
@@ -23,7 +23,7 @@
 	user-select: none;
 }
 
-.codicon-blank:before { content: '\81'; }
+.codicon-blank:before { content: '\2003'; }
```
pavelfeldman, line 26: not the best place to modify it (third party supposed to be rollable), but ok

pavelfeldman (approved):

Issue #16467, opened 2022-08-11
[Question] How to run serial and parallel tests within single test file


aakash3771:
I created a test file with 2 groups of tests using test.describe. For one group I set mode to serial and other to parallel. When I ran the test the tests in the test file, the parallel tests ran together but one serial test also started with them. My requirement is to run parallel tests in parallel, once they are finished then start the serial tests.

Also, I noted one thing in documentation that serial mode assumes the tests are dependent on each other which is not the case with me. What is the best way using features built in to Playwright?

rwoll:
Thanks for the issue. I'd like to make sure I understand the scenario before answering. Given the following tests:

- A
- B
- C
- D
- E

Which ones can run at the same time as eachother? And which need to run one after the other, if any?

aakash3771:
// feature 1.js
test.describe.parallel('parallel tests', () => {
  test('parallel one', async ({ page }) => {});

  test('parallel two', async ({ page }) => {});
});

test.describe.serial('serial tests', () => {
  test('serial one', async ({ page }) => {});

  test('serial two', async ({ page }) => {});
});

// feature 2.js
test.describe.parallel('parallel tests', () => {
  test('parallel one', async ({ page }) => {});

  test('parallel two', async ({ page }) => {});
});

test.describe.serial('serial tests', () => {
  test('serial one', async ({ page }) => {});

  test('serial two', async ({ page }) => {});
});

Requirements:
1. Run the parallel tests first, complete them. Once they are done, serial tests start. My serial tests have pre-condition and they are not inter-dependent.
2. I want a single Junit report for all above tests belonging to a feature. This will help in the import to a test management tool.

rwoll:
Relates https://github.com/microsoft/playwright/pull/16698

davidAtInleague:
is something like this possible:
```
test.describe.parallel(() => {
  test.describe.serial(() => {
    test(...) // A
    test(...) // B
  });
  test(...) // C
  test(...) // D
});
```
so B always follows A; but C can go before A or during A or after A or before D or after D, etc -- there's no ordering requirement on C or D.

lsokolova:
Hello! 
We have a similar need:
- we have 3 test groups (test.describe) in my test file
- we need to run one of them first, and then 2 others can run in parallel.

Will the same [PR 16698](https://github.com/microsoft/playwright/pull/16698) be re-opened for the new API proposal implementation? I wonder if we can follow up on it or if there will be a new PR?

[...]

Issue #16232, opened 2022-08-03
[Feature] file order with multiple workers


TJHdev:
From looking at the documentation there doesn't seem to be a way to specify the order that tests are run while also allowing multiple workers. 

I'm currently using playwright to test a fairly complex web hosting application that has a lot of platform based settings. These settings can take a long time to update (over 1 minute) and can cause other tests to fail as they affect all users. At the moment we're just using the single worker in the CI pipeline but as the number of tests grow this will mean the test suite will take a very long time to complete.

I was thinking of a more fully featured "test list" file where you can specify items to run sequentially and items that ok to run in parallel.

```ts
// test.list.ts
import './feature-b.spec.ts';
import './feature-a.spec.ts';
```

```ts
// new.test.list.ts
[
   './global-setting-a.spec.ts',  // changes to global setting
   ['./depends-on-a-1.spec.ts', './depends-on-a-2.spec.ts', '.depends-on-a-3.spec.ts', ...etc], // runs in parallel
   './global-setting-b.spec.ts'
   // runs any other tests in parallel that are not specified.
]
```

mxschmitt:
You can use "test list files" would that work for you? https://playwright.dev/docs/test-parallel#use-a-test-list-file

It should allow creating multiple of them which then get ran in parallel.

TJHdev:
That could help us split up the tests but they're still all going to race one another. Each test list file will need to be pointed at a separate backend instance which isn't ideal.

mxschmitt:
Yes, but how do you want to solve it? All the tests inside a list file will run sequentially so they don't race. I'm not following!

TJHdev:
It looks like the documentation was recently updated for "test list" files.

I was thinking something along the lines of syntax like this, but it's just a rough idea.

```ts
import { test } from '@playwright/test';

import addEmailRole from 'components/platform/addEmailRole.spec'
import addBackupRole from 'components/platform/addBackupRole.spec'
import removeEmailRole from 'components/platform/removeEmailRole.spec'
import removeBackupRole from 'components/platform/removeBackupRole.spec'

import addEmails from 'components/addEmails.spec';
import emailsBulkActionBar from 'components/emailsBulkActionBar.spec';

import addWebsite from 'components/addWebsite.spec';
import websitesBulkActionBar from 'components/websitesBulkActionBar.spec';
import backups from 'components/website/backups.spec'

test.describe(addEmailRole);  // won't start the next line until all tests have completed.
[...]

Issue #16215, opened 2022-08-03
[Feature] Allow Fixtures to be used in global-setup


azad-derakhshani-GS:
Currently, when trying to setup a `global-setup.ts`, it is not possible to use any of the Fixtures from Playwright or the custom Fixtures the user created.  
Instead, the pages and other variables have to be defined & instantiated from scratch there, which is not only redundant but also makes things more complicated.

So it would be a great improvement if the Fixtures could be utilized in `global-setup`.

rwoll:
Folding into https://github.com/microsoft/playwright/issues/14895#issuecomment-1203272743

Issue #16043, opened 2022-07-29
[Feature] Ability to attach screenshots / videos / traces captured in globalSetup / teardown to html report


vijaydubber:
we have testInfo.attach(), but there won't be any tests when globalSetup is running. So won't be able to attach at that level.

This would be especially useful when running tests in CI, because the HTML will show all the tests as skipped, and won't have any information on why it was skipped.

vijaydubber:
Hi @pavelfeldman , @rwoll  I would also like to request the steps executed in the global setup be available in the html report (like how each steps in a regular test is reported with code line and time taken for execution).

Do I need to raise a new feature request for that.
I understand the requirement may not fit nicely with the existing html reporting, but these features would be very helpful.

rwoll:
@vijaydubber I think we'll end up covering that in https://github.com/microsoft/playwright/issues/14895 (which is a new idea as of this morning)

vijaydubber:
For anyone who's looking for a workaround:
Make your global scripts save videos or traces in results/global, then follow the below steps:
circle:
```
      - run:
          name: Copy global trace if exist
          command: bash .circleci/scripts/copyGlobalTrace.sh
          when: always 

```
.circleci/scripts/copyGlobalTrace.sh
```
if [ -d "results/global" ]; then
    # This means global step has generated video or trace files.. we would copy it over to html-report directory..

    # Install tree and generate directory listing for global video or trace files
    apt install tree -y
    /usr/bin/tree results/global -H . -o results/global/index.html

    #copy them over to html-report and add a link in html-report/index.html
    cp -r results/global results/html-report/global
    echo "<a href='global/index.html'>global/index.html</a>" >> results/html-report/index.html
fi
```

rwoll:
Likely solved by https://github.com/microsoft/playwright/issues/14895#issuecomment-1207665459 which would relocate this setup code to a regular test which can take attachments as of today.

pavelfeldman:
Starting with v1.31 you'll be able to shape your global setup as a Playwright project: https://playwright.dev/docs/next/api/class-testproject#test-project-dependencies. You can set up tracing for your setup project and see your global setup in the trace viewer!

PR #16698, opened 2022-08-20
feat(test-runner): minimal proof-of-concept scheduler API

rwoll:
Proof-of-concept for https://github.com/microsoft/playwright/issues/14895#issuecomment-1219986634.

## About

This PR introduces a proof-of-concept, minimal user-facing scheduling API that attempts to address a variety of longstanding issues and feature requests.

At this point, the implementation details should be ignored (although look if you want a puzzle 😄 ). However, the PR serves to allow us to concretely discuss a proposed solution, the general test strategy (which will apply regardless of the solution/API), articulate pros/cons/problems, etc. (The implementation is currently enough so that we can play around with the proposed public API to get a feel for it in the context of the use cases it attempts to solve.)

## API Overview

This introduces a `projectSchedule?: string[][]` configuration option at the top-level, which can be used in combination with the already existing [`projects`](https://playwright.dev/docs/api/class-testproject) configuration option to guarantee which projects run in relation to others.

Using it looks something like:

```ts
      projects: [
        { name: 'My Setup', testDir: './setup',  },
        { name: 'Desktop', testDir: './desktop', use: devices['Desktop Chrome'] },
        { name: 'Mobile', testDir: './mobile', use: devices['iPhone 11'] },
        { name: 'My Teardown', testDir: './teardown', },
      ],
      projectSchedule: [
        ['My Setup'],
        ['Desktop', 'Mobile'],
        ['My Teardown']
      ],
    },
```

In this case, this simply means: run all tests in the _My Setup_ project, when that's done run all the tests in _Desktop_ and _Mobile_ project in parallel, then run _My Teardown_ project.

## Issues Addressed

1. User wants to run global setup/teardown code and have a trace in the HTML Report
    - https://github.com/microsoft/playwright/issues/14895
    - https://github.com/microsoft/playwright/issues/16043
1. Trace, Retry, Device, etc. Settings from Config Should Apply to Global Setup/Teardown
    - https://github.com/microsoft/playwright/issues/14895
1. Different Global Setup Per-Project
    - https://github.com/microsoft/playwright/issues/16008
1. Allow Fixtures to Be Used in Global Setup
    - https://github.com/microsoft/playwright/issues/16215
1. How to Use Create and Use Storage State for Tests in Global Setup
    - https://github.com/microsoft/playwright/issues/14759
    - https://github.com/microsoft/playwright/issues/12063
1. Run Files/Projects in Certain Order AND Include in One HTML Report
    - https://github.com/microsoft/playwright/issues/14084
   - https://github.com/microsoft/playwright/issues/15788
   - https://github.com/microsoft/playwright/issues/14535
   - https://github.com/microsoft/playwright/issues/10290
   - https://github.com/microsoft/playwright/issues/8295
   - https://github.com/microsoft/playwright/issues/16232
   - https://github.com/microsoft/playwright/issues/16467

From memory, there are other issues (and Slack conversations) that fall into some of the buckets above, too.

## Pros

1. minimal, but effective(?) API: less is more. It's much easier to explain and program than some [fancy DAG-based API](https://github.com/microsoft/playwright/issues/14895#issuecomment-1207665459)
2. everything's just a test! we get config, tracing, retries, fixtures, etc. for free — no more trying to figure out how to use globalSetup/globalTeardown

## Cons/Limitations/Questions

When using `projectSchedule`:

- cannot use `shard` option: if you're using the schedule, you can quasi-shard with the schedule itself
- project names must be unique

In the above example, if I want to run _Desktop_ project, but not _Mobile_, how do we express this? Presumably we need to run _My Setup_ and _My Teardown_, too.

We're not exposing a full-featured DAG, so dependency expression is limited (I think for the better, for now — although we can do some really neat scheduling optimizations in the future with 1 timing run + the DAG).

We don't guarantee projects run on the same worker/process, so you can't, for instance, start a webserver in memory in _My Setup_ and kill it in _My Teardown_. Although, in this case you can just use the `webServer` config option, or write a PID in your setup, then read and kill that PID in your teardown.

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -349,6 +349,12 @@ Whether to preserve test output in the [`property: TestConfig.outputDir`]. Defau
 Playwright Test supports running multiple test projects at the same time. See [TestProject] for more information.
 
 
+## property: TestConfig.projectSchedule
+* since: v1.26
+- type: ?<[Array]<[[Array]<[string]>]>>
+
+TODO
+
 ## property: TestConfig.quiet
 * since: v1.10
 - type: ?<[boolean]>
--- a/packages/playwright-test/src/dispatcher.ts
+++ b/packages/playwright-test/src/dispatcher.ts
@@ -26,6 +26,7 @@ import { ManualPromise } from 'playwright-core/lib/utils/manualPromise';
 import { TestTypeImpl } from './testType';
 
 export type TestGroup = {
+  priority: number;
   workerHash: string;
   requireFile: string;
   repeatEachIndex: number;
@@ -53,6 +54,7 @@ type WorkerExitData = {
 export class Dispatcher {
   private _workerSlots: { busy: boolean, worker?: Worker }[] = [];
   private _queue: TestGroup[] = [];
+  private _runningJobs: Set<TestGroup> = new Set();
   private _queuedOrRunningHashCount = new Map<string, number>();
   private _finished = new ManualPromise<void>();
   private _isStopped = false;
@@ -88,9 +90,17 @@ export class Dispatcher {
     if (index === -1)
       return;
 
-    // 3. Claim both the job and the worker, run the job and release the worker.
+    // 3. Attempt to claim both the job and the worker
     this._queue.shift();
     this._workerSlots[index].busy = true;
+    // 3.1. If the dispatcher is not ready to run this priority, release the worker and job for the future.
+    const inProgressPriority = () => this._runningJobs.size ? this._runningJobs.values().next().value.priority : undefined;
+    if (inProgressPriority() !== undefined && job.priority !== inProgressPriority()) {
+      this._workerSlots[index].busy = false;
+      this._queue.unshift(job);
+      setTimeout(() => this._scheduleJob(), 0);
+      return;
+    }
     await this._startJobInWorker(index, job);
     this._workerSlots[index].busy = false;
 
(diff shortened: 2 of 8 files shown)


rwoll:
Initial feedback from @dgozman:

> 1. Is seems equivalent to introducing priority field on the project - is that right?

Yes, it's functionally equivalent. However, I think if we omitted `projectSchedule` and just did `priority` on project, you'd force the user to build up schedule array you see in `projectSchedule` anyway. Priority of an individual project is not helpful, since it really only matters in relation to the other project priorities, so you end up building this in your head.

> 2. We do not address "global setup per project"

We do, (although it's not necessarily pretty). It would look something like:

```
      projects: [
        { name: 'Desktop Setup', testDir: './desktop/setup', use: devices['Desktop Chrome']  },
        { name: 'Desktop', testDir: './desktop/tests', use: devices['Desktop Chrome'] },
        { name: 'Mobile Setup', testDir: './mobile/setup', use: devices['iPhone 11'] },
        { name: 'Mobile', testDir: './mobile/tests', use: devices['iPhone 11'] },
      ],
      projectSchedule: [
        ['Desktop Setup', 'Mobile Setup'],
        ['Desktop', 'Mobile'],
      ],
```

> 3. `--project` and `--shard` kinda mess things up

`--shard` I need to think more on (and intentionally punted per the limitations section), but I think the `--project` issue is resolvable by using a project-level-DAG instead of a test level-DAG.

i.e. scrap the `projectSchedule` list. And instead allow a each project to specify `dependsOn: […]`:

```
      projects: [
        { name: 'Desktop Setup', testDir: './desktop/setup', use: devices['Desktop Chrome']  },
        { name: 'Desktop', testDir: './desktop/tests', use: devices['Desktop Chrome'], dependsOn: ['Desktop Setup'] },
        { name: 'Mobile Setup', testDir: './mobile/setup', use: devices['iPhone 11'] },
        { name: 'Mobile', testDir: './mobile/tests', use: devices['iPhone 11'], dependsOn: ['Mobile Setup'] },
      ],
      projectSchedule: [
        ['Desktop Setup', 'Mobile Setup'],
        ['Desktop', 'Mobile'],
      ],
```

So if a user does, `npx playwright test --project=Mobile` it would run any dependencies, too (e.g. `Mobile Setup` project) in this case.

> 4. Great job on the detailed description and finding related issues!

😄

rwoll:
Closing for now — after much offline discussion, we have a new API proposal, so we will re-open with that implementation.
.

Issue #15057, opened 2022-06-23
[Feature] Components: support for Solid.js


floer32:
Since there is support for Svelte, Vue, and React, I imagine the components feature could also support [Solid.js](https://solidjs.com/) components eventually. Solid.js projects are built with Vite.

I'll try it myself soon, but wanted to open the issue first, in case it takes me a while before getting around to it.

mxschmitt:
Since there is a vite plugin for it, it should be doable: https://github.com/solidjs/vite-plugin-solid

Marking it as a feature request.

bjoluc:
To anyone coming across this issue looking for SolidJS component testing support: `@playwright/experimental-ct-solid` has been deprecated again in #33523, sadly. The last published version (1.48.2) still works, but it lacks any Playwright features released after Playwright 1.48.2, unfortunately.

sand4rt:
> To anyone coming across this issue looking for SolidJS component testing support: `@playwright/experimental-ct-solid` has been deprecated again in [#33523](https://github.com/microsoft/playwright/pull/33523), sadly. The last published version (1.48.2) still works, but it lacks any Playwright features released after Playwright 1.48.2, unfortunately.

I'm going bring the library back to life soon under my own namespace—I'll post an update here when it's done

bjoluc:
That's wonderful news – thanks @sand4rt! I'm currently overriding my Playwright version (still works fine with Playwright 1.53.0), but that's quite scary and guaranteed to break sooner or later, so I appreciate your efforts! 🙌

sand4rt:
@bjoluc see: https://github.com/sand4rt/playwright-ct-solid

PR #16703, opened 2022-08-20
feat(ct): initial solid

sand4rt:
Very early stage, only a simple mount without props is possible. Related to https://github.com/microsoft/playwright/issues/15057


pavelfeldman:
This looks pretty slick. Do you think we'll be able to do properties and other features? If you are optimistic, I'm totally fine with landing it.

pavelfeldman (approved):

sand4rt:
@pavelfeldman Cool, i believe so yes😄

Issue #16736, opened 2022-08-22
[BUG] `Firefox` `route.continue({postData})` overrides `content-type` to `application/octet-stream`


panmenghan:
**Context:**
- Playwright Version: 1.25.0
- Operating System: Windows
- Node.js version: 16.17.0
- Browser: Firefox
- Extra:

**Code Snippet**

```javascript
import {firefox} from 'playwright'

void main()

async function main() {
  const browser = await firefox.launch({
    headless: false,
  })
  const context = await browser.newContext()
  const page = await context.newPage()
  
  // goto any site
  await page.goto('https://www.example.com/')

  // setup route
  await page.route('https://httpbin.org/anything', (route, request) => {
    const postData = request.postDataJSON()
    postData.a = 2
    route.continue({postData: JSON.stringify(postData)}).catch(() => {})
  })

  // send post request
  const data = await page.evaluate(async () => {
    const response = await fetch('https://httpbin.org/anything', {
      method: 'POST',
      body: '{"a":1}',
      headers: {'content-type': 'application/json'}, 
    return response.json()
  })

  console.log(data) // <- data.headers['Content-Type'] is `application/octet-stream`, but should be `application/json`
}
```

**Describe the bug**

This bug is only in Firefox.

PR #16739, opened 2022-08-23
browser(firefox): preserve content-type from original request if it i…

yury-s:
…s not overridden

Reference #16736

### DIFF
--- a/browser_patches/firefox-beta/juggler/NetworkObserver.js
+++ b/browser_patches/firefox-beta/juggler/NetworkObserver.js
@@ -878,7 +878,7 @@ function setPostData(httpChannel, postData, headers) {
   let contentType = overriddenHeader('content-type');
   if (contentType === undefined) {
     try {
-      contentType = httpChannel.getRequestHeader('content-encoding');
+      contentType = httpChannel.getRequestHeader('content-type');
     } catch (e) {
       if (e.result == Cr.NS_ERROR_NOT_AVAILABLE)
         contentType =  'application/octet-stream';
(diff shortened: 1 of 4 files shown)


browser_patches/firefox-beta/juggler/NetworkObserver.js
```diff
@@ -863,19 +863,30 @@ function setPostData(httpChannel, postData, headers) {
...
-    return defaultValue;
+    return undefined;
   }
   // Clear content-length, so that upload stream resets it.
   httpChannel.setRequestHeader('content-length', '', false /* merge */);
-  httpChannel.explicitSetUploadStream(synthesized, overriddenHeader('content-type', 'application/octet-stream'), -1, httpChannel.requestMethod, false);
+  let contentType = overriddenHeader('content-type');
+  if (contentType === undefined) {
+    try {
+      contentType = httpChannel.getRequestHeader('content-encoding');
```
rwoll, line 881: Is this correct? What happens with gzipped req/resp? Would we report content-type as `gzip` b/c of this line? That doesn't sound correct.

yury-s: Yes this is the intended behavior unless content-type is explicitly overridden we use the one from the original request. That's what we do in other browsers too.

yury-s: Do you suggest something different?

yury-s: Oh, you meant the typo 'content-encoding' instead of 'content-type', fixed.

dgozman (approved):

PR #16810, opened 2022-08-24
test: update "page.goBack should work for file urls" to match status quo

dgozman:
Chromium works correctly, WebKit mac has a security error, Firefox fails to go back.

### DIFF
--- a/tests/page/page-history.spec.ts
+++ b/tests/page/page-history.spec.ts
@@ -53,20 +53,29 @@ it('page.goBack should work with HistoryAPI', async ({ page, server }) => {
 });
 
 it('page.goBack should work for file urls', async ({ page, server, asset, browserName, platform, isAndroid }) => {
-  it.fail(browserName === 'webkit' && platform === 'darwin');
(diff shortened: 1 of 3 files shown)


tests/page/page-history.spec.ts
```diff
@@ -53,28 +53,40 @@ it('page.goBack should work with HistoryAPI', async ({ page, server }) => {
 });
 
 it('page.goBack should work for file urls', async ({ page, server, asset, browserName, platform, isAndroid }) => {
-  it.fail(browserName === 'webkit' && platform === 'darwin');
+  it.fixme(browserName === 'firefox', 'Firefox pretends, but does not complete goBack to the file url');
+  it.fail(browserName === 'webkit' && platform === 'darwin', 'WebKit embedder fails to go back/forward to the file url.');
```
yury-s, line 57: make it fixme, otherwise it always times out?

yury-s (approved):

tests/page/page-history.spec.ts
dgozman: WebKit is failing with an exception, so we are good. Leaving as it was.

Issue #16545, opened 2022-08-15
[Feature] WebView2 Desktop testing support with Playwright


mastrzyz:
Hello Folks, is there any guidance on how to perform testing on Playwright for standalone WebView 2 desktop applications?

We have been able to connect to via `connectOverCDP` BUT!

I am running into an issue where there are no pages in the browserContext and I cannot create a new page or a new context, leaving me unable to test various parts of the application. What can I do to resolve this issue?

OR is there any Webview2 documentation or examples for a standalone application?

The same exact application setup works for Selenium.

---

Open tasks:

- [x] Make sure that we work with webview2, add test infra on our repo to guarantee, that we work overtime and don't regress (run essentially page tests inside it, like we do for Android and Electron) https://github.com/microsoft/playwright/pull/16827
- [x] Add a guide for it on our documentation site (all the language bindings) https://github.com/microsoft/playwright/pull/17190
- [ ] Add a guide on the msedge documentation site: https://docs.microsoft.com/en-us/microsoft-edge/webview2/how-to/webdriver

aslushnikov:
We should add some docs for this!

mastrzyz:
Appreciate the work so far on this!

mxschmitt:
Starting in the next release, we'll have [a dedicated doc](https://playwright.dev/docs/next/webview2) for it.

If you encounter any issues with the doc, or things which don't work, I recommend filing them in separate issues.

[This](https://playwright.dev/docs/next/webview2) doc explains the manual approach using CDP and connecting to webview2 controls, for having an easier integration, I filed this feature request: https://github.com/microsoft/playwright/issues/17617

Since the Playwright part is done by that, I'm closing it.

PR #16827, opened 2022-08-25
test: add initial webview2 tests

mxschmitt:
Relates https://github.com/microsoft/playwright/issues/16545


tests/webview2/webView2Test.ts
```diff
@@ -0,0 +1,69 @@
...
+export { expect } from '@playwright/test';
+import { TestChildProcess } from '../config/commonFixtures';
+import { DEFAULT_ARGS } from '../../packages/playwright-core/lib/server/chromium/chromium';
+
+type WebView2WorkerFixtures = PageWorkerFixtures & {
+  webView2CdpPort: number;
+};
+
+process.env.PW_TEST_REUSE_CONTEXT = '1';
+
```
mxschmitt, line 31:
```suggestion
```

tests/webview2/webView2Test.ts
```diff
@@ -0,0 +1,69 @@
...
+      if (data.toString().includes('WebView2 initialized'))
+        resolve();
+    }));
+    await use(cdpPort);
+    await spawnedProcess.close();
+  }, { scope: 'worker' }],
+
+  page: async ({ context }, run) => {
+    const page = context.pages()[0];
+    await page.goto('about:blank');
```
mxschmitt, line 60: not super happy about that, the best would be if we navigate either in a separate session so the history is still null or if we navigate after the test.

tests/webview2/webView2Test.ts
```diff
@@ -0,0 +1,67 @@
...
+    await run(page);
+  },
+
+  context: async ({ playwright, webView2CdpPort }, run) => {
+    const browser = await playwright.chromium.connectOverCDP(`http://127.0.0.1:${webView2CdpPort}`);
```
yury-s, line 63: Reconnecting doesn't help much with cleaning up page state in the browser (like history, storage etc), I think we can keep using the same connection but clear context state. But it'd be much easier if we could create a new browser context per test. As discussed offline let's ask webview2 folks and if they think creating a new context is an option.

yury-s (approved):

PR #16839, opened 2022-08-25
feat(ct): solid default child

sand4rt:
Related to https://github.com/microsoft/playwright/issues/15057


packages/playwright-ct-solid/registerSource.mjs
```diff
@@ -54,7 +54,10 @@ function render(component) {
   if (component.kind !== 'jsx')
     throw new Error('Object mount notation is not supported');
 
-  return createComponent(componentFunc, component.props);
+  return createComponent(componentFunc, {
+    children: component.children,
+    ...component.props
```
pavelfeldman, line 59: nit: trim children from here?

pavelfeldman (approved):

packages/playwright-ct-solid/registerSource.mjs
sand4rt: Can I fix this in the next PR? I'm working on the multi-root children feature which touches this part

pavelfeldman: Sadly they always require rebaselines

PR #16871, opened 2022-08-27
chore(ct): remove example project

sand4rt:
We already have examples here: https://github.com/microsoft/playwright/tree/main/tests/components


pavelfeldman:
We do, but I'd like to keep it there for discoverability so that people could get started with the component testing. In particular, I'd like dependencies to be like the users would have them (not assuming that it runs within our monorepo). And it should also be user-friendly, our tests don't need to. I realize that examples/ is now behind in terms of the available features and that it requires duplicate effort, but I have no better ideas than keeping it around atm.

PR #16885, opened 2022-08-29
chore: added starting time of test to the base json reporter

### LINKED ISSUE #16617: [Question] getting date and time of when a test was run into the default playwright report
Hello,

` 
reporter: [

    ['line'],

    ['allure-playwright'],

    ['json', { outputFile: 'test-results/json/results.json' }],

    ['junit', { outputFile: 'test-results/xml/results.xml' }]

  ],`

With the above in playwright.config.ts at the end of a run, playwright generates a JSON with all the results.
It does not log the timestamp when the test was actually run.

How would one get this to be shown in the report?

PS: we have found allure does it, but that leads to other problems and we need a way to do it without allure.

Any tips will be appreciated.

Thanks!

nate-ro:
Fixes: #16617

### DIFF
--- a/packages/playwright-test/types/testReporter.d.ts
+++ b/packages/playwright-test/types/testReporter.d.ts
@@ -493,6 +493,7 @@ export interface JSONReportTestResult {
   stderr: JSONReportSTDIOEntry[];
   retry: number;
   steps?: JSONReportTestStep[];
+  startTime?: Date;
   attachments: {
     name: string;
     path?: string;
--- a/packages/playwright-test/types/testReporter.d.ts
+++ b/packages/playwright-test/types/testReporter.d.ts
@@ -493,7 +493,7 @@ export interface JSONReportTestResult {
   stderr: JSONReportSTDIOEntry[];
   retry: number;
   steps?: JSONReportTestStep[];
-  startTime?: Date;
+  startTime: Date;
   attachments: {
     name: string;
     path?: string;
--- a/tests/playwright-test/reporter-json.spec.ts
+++ b/tests/playwright-test/reporter-json.spec.ts
@@ -235,3 +235,20 @@ test('should add line in addition to file json without CI', async ({ runInlineTe
   expect(stripAnsi(result.output)).toContain('[1/1] a.test.js:6:7 › one');
   expect(fs.existsSync(testInfo.outputPath('a.json'))).toBeTruthy();
 });
+test('should have starting time in results', async ({ runInlineTest }, testInfo) => {
+  const result = await runInlineTest({
+    'a.test.js': `
+      const { test } = pwt;
+      test('math works!', async ({}) => {
+        expect(1 + 1).toBe(2);
+      });
+    `
+  },  {reporter: 'json' });
+  expect(result.exitCode).toBe(0);
+  if (result.report.suites[0].specs[0].tests[0].results[0].startTime) {
+    expect(new Date(result.report.suites[0].specs[0].tests[0].results[0].startTime).getTime).toBeTruthy();
+  } 
+  else {
+    expect(0).toBe(1);
+  }
+});
--- a/tests/playwright-test/reporter-json.spec.ts
+++ b/tests/playwright-test/reporter-json.spec.ts
@@ -243,12 +243,10 @@ test('should have starting time in results', async ({ runInlineTest }, testInfo)
         expect(1 + 1).toBe(2);
       });
     `
-  },  {reporter: 'json' });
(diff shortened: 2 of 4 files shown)


pavelfeldman:
Make sure you always include the link to the issue you are addressing. Edit the body of this PR and say "Fixes: #XXXXX". GitHub will link it.

tests/playwright-test/reporter-json.spec.ts
```diff
@@ -235,3 +235,18 @@ test('should add line in addition to file json without CI', async ({ runInlineTe
...
+    'a.test.js': `
+      const { test } = pwt;
+      test('math works!', async ({}) => {
+        expect(1 + 1).toBe(2);
+      });
+    `
+  },   { reporter: 'json' });
+  expect(result.exitCode).toBe(0);
+  if (result.report.suites[0].specs[0].tests[0].results[0].startTime)
+    expect(new Date(result.report.suites[0].specs[0].tests[0].results[0].startTime).getTime).toBeTruthy();
```
pavelfeldman, line 249: Note: this will always pass.

tests/playwright-test/reporter-json.spec.ts
```diff
@@ -235,3 +235,18 @@ test('should add line in addition to file json without CI', async ({ runInlineTe
...
+  else
+    expect(0).toBe(1);
```
pavelfeldman, line 251: We always try to make assertions easy to follow. This one would require user to look into the code when it fails.

tests/playwright-test/reporter-json.spec.ts
```diff
@@ -235,3 +235,18 @@ test('should add line in addition to file json without CI', async ({ runInlineTe
...
+  const result = await runInlineTest({
+    'a.test.js': `
+      const { test } = pwt;
+      test('math works!', async ({}) => {
+        expect(1 + 1).toBe(2);
+      });
+    `
+  },   { reporter: 'json' });
+  expect(result.exitCode).toBe(0);
+  if (result.report.suites[0].specs[0].tests[0].results[0].startTime)
```
pavelfeldman, line 248: So you want to check if this field a) exists and b) can be parsed as the Date object from JSON and c) has reasonable value. Something like this:

```js
const startTime = result.report.suites[0].specs[0].tests[0].results[0].startTime;
expect(new Date(startTime).getTime()).toBeGreaterThan(new Date('1/1/2000').getTime());
```

It still allows startTime to be lax (not ISO format), but that's fine for the purpose of this test.

nate-ro: The issue with this code is what I had earlier and why I wrote that mess with an if statement, 
image

Is it alright to leave an error like that?

tests/playwright-test/reporter-json.spec.ts
nate-ro: I see, during my testing it showed the code around that when it failed
image
It definitely could be clearer, though.

tests/playwright-test/reporter-json.spec.ts
pavelfeldman: Your startTime is nullable, which does not matter in the scope of this test, you can always cast to non-nullable via appending !

packages/playwright-test/types/testReporter.d.ts
```diff
@@ -493,6 +493,7 @@ export interface JSONReportTestResult {
   stderr: JSONReportSTDIOEntry[];
   retry: number;
   steps?: JSONReportTestStep[];
+  startTime?: Date;
```
pavelfeldman, line 496: Btw, can we make it required here or is it optional on the result itself?

nate-ro: I believe we can make it required, I left it optional because I didn't want this change to break anyone's reporter if they used those types.

pavelfeldman: This should not break the call sites since clients are the consumer of this api. You can add options and change optional to required, but not the other way.

pavelfeldman: So I'm suggesting that you make this change and convert `startTime` into a required field. Our main public API is `testReporter.d.ts`, and it is already required there. What you are editing here is a JSON-serialized format, which is still public, but much less important. If we managed to commit to this field being in test reporter API, committing to it in JSON is an easy sell.

nate-ro: I see, I'll make that change

pavelfeldman (approved):

PR #16898, opened 2022-08-29
fix: include headlessness as a launch argument to browser

aslushnikov:


packages/playwright-test/src/index.ts
```diff
@@ -116,7 +116,7 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
     const browser = await playwright[browserName].connect(connectOptions.wsEndpoint, {
       headers: {
         'x-playwright-browser': browserName,
-        'x-playwright-launch-options': JSON.stringify(launchOptions),
+        'x-playwright-launch-options': JSON.stringify({...launchOptions, headless }),
```
dgozman, line 119: `channel` is also missing.

packages/playwright-test/src/index.ts
```diff
@@ -116,7 +116,7 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
     const browser = await playwright[browserName].connect(connectOptions.wsEndpoint, {
       headers: {
         'x-playwright-browser': browserName,
-        'x-playwright-launch-options': JSON.stringify(launchOptions),
+        'x-playwright-launch-options': JSON.stringify({ ...launchOptions, headless, channel }),
```
dgozman, line 119: We should not override user values with undefined here. Take a look at [this code](https://github.com/microsoft/playwright/blob/main/packages/playwright-test/src/index.ts#L97-L100).

dgozman (approved):

PR #16900, opened 2022-08-29
fix(click): allow clicking inside closed shadow root

### LINKED ISSUE #16828: [Feature] Toggle check for frame receives events
Hello,
with recent bump of playwright from 1.22 to 1.24 / 1.25, it was noticed and new bug-fix/feature for validation of click over frames.

This change was introduced with this two commits:
```
[BUG] Misclick when some element overlays the target's iframe
https://github.com/microsoft/playwright/issues/14752
```
```
Playwright missclicks on iframe instead on actual element
https://github.com/microsoft/playwright/issues/13762
```

More/less this is an test design/implementation issue, but also normally nobody wants to click outside the main port/frame.

With the quoted fixes above, this is a breaking change for existing tests, running on a website that have some different implementation, from the normal behavior or other element intercepts the frame actions.

Sometimes the tester is out of control, how the website is designed/implemented, and/or cannot "fix" the website due 3rd party implementation/integration/other reason.

So my point is this has to be under control with feature flag.

There is a similar topics already created as a "bug", where the "solution" is to use "force" flag for operations as click.
But the "force" disables all other auto-waiting actionability checks.

Proposing a feature flag to be introduced only for the check whether element is outside the main-frame, or other element receive events in iframe.

Code ref: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/dom.ts#L439

Thanks!

dgozman:
Although Playwright selectors do not pierce closed shadow roots, one can still obtain a reference to an element inside a closed shadow root:
- through `page.evaluateHandle()`;
- through `handle.$()` where `handle` is inside the shadow root;
- through `frame.locator()` by choosing an iframe that belongs to a closed shadow root.

In this case, `click()` action fails during the hit check test, but it is possible to make it work by going bottom up from the target rather than top down from the document.

Fixes #16828.

### DIFF
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -23,7 +23,7 @@ import type { NestedSelectorBody, ParsedSelector, ParsedSelectorPart } from '../
 import { allEngineNames, parseSelector, stringifySelector } from '../isomorphic/selectorParser';
 import { type TextMatcher, elementMatchesText, createRegexTextMatcher, createStrictTextMatcher, createLaxTextMatcher, elementText } from './selectorUtils';
 import { SelectorEvaluatorImpl } from './selectorEvaluator';
-import { isElementVisible, parentElementOrShadowHost } from './domUtils';
+import { enclosingShadowRootOrDocument, isElementVisible, parentElementOrShadowHost } from './domUtils';
 import type { CSSComplexSelectorList } from '../isomorphic/cssParser';
 import { generateSelector } from './selectorGenerator';
 import type * as channels from '../../protocol/channels';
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -726,14 +726,59 @@ export class InjectedScript {
...
+      }
+      const innerElement = elements[0] as Element | undefined;
+      if (!innerElement)
+        break;
+      hitElement = innerElement;
+      if (index && innerElement !== (roots[index - 1] as ShadowRoot).host)
+        break;
+    }
+
+    // console.log(`expectHitTarget hitElement=${hitElement ? this.previewNode(hitElement) : '<none>'} targetElement: ${this.previewNode(targetElement)}`);
```
yury-s, line 771: remove?

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -726,14 +726,59 @@ export class InjectedScript {
...
+          // Workaround a case where elementsFromPoint misses the inner-most element with display:contents.
+          // https://bugs.chromium.org/p/chromium/issues/detail?id=1342092
+          elements.unshift(singleElement);
+        }
+      }
+      const innerElement = elements[0] as Element | undefined;
+      if (!innerElement)
+        break;
+      hitElement = innerElement;
+      if (index && innerElement !== (roots[index - 1] as ShadowRoot).host)
```
yury-s, line 767: Let's add a test with two shadow roots overlaying each other (with the one containing target being overlayed).

yury-s (approved):

PR #16902, opened 2022-08-29
feat: --pass-with-no-tests option

### LINKED ISSUE #16737: [Feature] Parameter or config option to NOT fail with exit code 1 when no tests are found
When running tests (for example with grep filters or similar) and no tests are found to match, Playwright terminates with exit code 1.

This has a significant impact in CI where builds fail as a result. Sometimes a pipeline will run tests if they exist (for example if they are part of a feature toggle), but such tests come and go.

In Jest this is already possible with "--passWithNoTests". A similar feature in Playwright would be very useful.

yury-s:
Fixes #16737

### DIFF
--- a/packages/playwright-test/src/runner.ts
+++ b/packages/playwright-test/src/runner.ts
@@ -338,8 +338,7 @@ export class Runner {
 
     // 7. Fail when no tests.
     let total = rootSuite.allTests().length;
-    const noTests = !total;
-    if (!total)
+    if (!total && !options.passWithNoTests)
       fatalErrors.push(createNoTestsError());
 
     // 8. Compute shards.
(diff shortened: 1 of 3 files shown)


packages/playwright-test/src/runner.ts
```diff
@@ -381,6 +383,8 @@ export class Runner {
     if (fatalErrors.length) {
       for (const error of fatalErrors)
         this._reporter.onError?.(error);
+      if (options.passWithNoTests && noTests && fatalErrors.length === 1)
```
dgozman, line 386: I would rather not add the error in the first place.

tests/playwright-test/exit-code.spec.ts
```diff
@@ -156,6 +156,16 @@ test('should exit with code 1 if passed a file name', async ({ runInlineTest })
   expect(result.output).toContain(`no tests found.`);
 });
 
+test.only('should exit with code 0 with --pass-with-no-tests', async ({ runInlineTest }) => {
```
dgozman, line 159:
```suggestion
test('should exit with code 0 with --pass-with-no-tests', async ({ runInlineTest }) => {
```

packages/playwright-test/src/runner.ts
yury-s: Done.

dgozman (approved):

PR #16907, opened 2022-08-30
fix: inject caret-hiding style in every shadow tree

### LINKED ISSUE #16732: [Question] Should screenshot option to hide caret work with Custom Elements (Shadow DOM)?
I love the fairly new options for the screenshotter to fast forward animations and hide carets. Implementing these will remove manual workarounds we had to implement for our tests.

I'm not sure if this is a bug or intentional, but screenshot.caret's implementation (caret: "hide") does not work with Shadow DOM for custom elements. Conversely, disabling animations does work with Shadow DOM (animations: "disabled").

You can see in the source code these are next to each other, disableAnimations clearly traverses the DOM tree but hideCaret just applies a style at the document.

disableAnimations
https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/screenshotter.ts#L162

hideCaret
https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/screenshotter.ts#L151

Before filing a bug I wanted to check whether this was intentional. It seems like they should both work with custom elements for consistency.

aslushnikov:
Style inheritance disregards selector specificity, so we can't
dominate local shadow dom styles.

To mitigate this, we inject the style tag with caret-hiding
style in every shadowDom tree.

Fixes #16732

### DIFF
--- a/packages/playwright-core/src/server/screenshotter.ts
+++ b/packages/playwright-core/src/server/screenshotter.ts
@@ -147,30 +147,38 @@ export class Screenshotter {
       progress.log('  disabled all CSS animations');
     await Promise.all(this._page.frames().map(async frame => {
       await frame.nonStallingEvaluateInExistingContext('(' + (async function(hideCaret: boolean, disableAnimations: boolean) {
-        const styleTag = document.createElement('style');
+        const collectRoots = (root: Document | ShadowRoot, roots: (Document|ShadowRoot)[] = []): (Document|ShadowRoot)[] => {
+          roots.push(root);
+          const walker = document.createTreeWalker(root, NodeFilter.SHOW_ELEMENT);
+          do {
+            const node = walker.currentNode;
+            const shadowRoot = node instanceof Element ? node.shadowRoot : null;
+            if (shadowRoot)
+              collectRoots(shadowRoot, roots);
+          } while (walker.nextNode());
+          return roots;
+        };
+
+        const styleTags: Element[] = [];
         if (hideCaret) {
-          styleTag.textContent = `
-            *:not(#playwright-aaaaaaaaaa.playwright-bbbbbbbbbbb.playwright-cccccccccc.playwright-dddddddddd.playwright-eeeeeeeee) {
-              caret-color: transparent !important;
-            }
-          `;
-          document.documentElement.append(styleTag);
+          for (const root of collectRoots(document)) {
+            const styleTag = document.createElement('style');
+            styleTag.textContent = `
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -147,30 +147,38 @@ export class Screenshotter {
...
-          styleTag.textContent = `
-            *:not(#playwright-aaaaaaaaaa.playwright-bbbbbbbbbbb.playwright-cccccccccc.playwright-dddddddddd.playwright-eeeeeeeee) {
-              caret-color: transparent !important;
-            }
-          `;
-          document.documentElement.append(styleTag);
+          for (const root of collectRoots(document)) {
+            const styleTag = document.createElement('style');
+            styleTag.textContent = `
+              *:not(#playwright-aaaaaaaaaa.playwright-bbbbbbbbbbb.playwright-cccccccccc.playwright-dddddddddd.playwright-eeeeeeeee) {
```
yury-s, line 167: I wonder why we need it to be such a looong id?

yury-s (approved):

packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -147,30 +147,38 @@ export class Screenshotter {
...
+
+        const styleTags: Element[] = [];
         if (hideCaret) {
-          styleTag.textContent = `
-            *:not(#playwright-aaaaaaaaaa.playwright-bbbbbbbbbbb.playwright-cccccccccc.playwright-dddddddddd.playwright-eeeeeeeee) {
-              caret-color: transparent !important;
-            }
-          `;
-          document.documentElement.append(styleTag);
+          for (const root of collectRoots(document)) {
```
dgozman, line 164: Can we call `collectRoots()` just once instead of once for caret and once for animations?

packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -147,30 +147,38 @@ export class Screenshotter {
       progress.log('  disabled all CSS animations');
     await Promise.all(this._page.frames().map(async frame => {
       await frame.nonStallingEvaluateInExistingContext('(' + (async function(hideCaret: boolean, disableAnimations: boolean) {
```
dgozman, line 149: nit: this function is getting complicated, can we extract it somewhere?

packages/playwright-core/src/server/screenshotter.ts
aslushnikov: this makes it very specific so that it overtakes any other selectors!

Issue #16745, opened 2022-08-23
[BUG] CSS are not loaded using Playwright/Webkit in some cases


mbouchotsuricate:
**Context:**
- Playwright Version: 1.25.0
- Operating System: Linux (Lubuntu 18.04 on VirtuaBox VM)
- Node.js version: 16.17.0
- Browser: Webkit
- Extra: NA

**Issued Code Snippet**

```javascript
const { webkit, devices } = require('playwright');

(async () => {
    const options = {
        devtools: false,
        headless: false,
        dumpio: true,
        ignoreHTTPSErrors: true,
        args: []
    };
    const browser = await webkit.launch(options);
    const context = await browser.newContext({
        ...devices['iPhone 11'],
        ignoreHTTPSErrors: true,
        serviceWorkers: 'block'
    });

    const page = await context.newPage();

    await context.route('**', route => {
        route.continue();
    });

    await page.goto('https://www.picard.fr/');
    await page.locator('#onetrust-accept-btn-handler').tap();

    await page.locator('[href="https://www.picard.fr/on/demandware.store/Sites-picard-Site/fr_FR/Account-Show"]').tap();
    //await browser.close();
})();
```
**Working Code Snippet**

```javascript
const { webkit, devices } = require('playwright');

(async () => {
    const options = {
        devtools: false,
        headless: false,
        dumpio: true,
        ignoreHTTPSErrors: true,
        args: []
    };
    const browser = await webkit.launch(options);
    const context = await browser.newContext({
        ...devices['iPhone 11'],
        ignoreHTTPSErrors: true,
        serviceWorkers: 'block'
    });

    const page = await context.newPage();

    /*await context.route('**', route => {
        route.continue();
    });*/

    await page.goto('https://www.picard.fr/');
    await page.locator('#onetrust-accept-btn-handler').tap();

    await page.locator('[href="https://www.picard.fr/on/demandware.store/Sites-picard-Site/fr_FR/Account-Show"]').tap();
    //await browser.close();
})();
```

**Describe the bug**

The bug is that when we declare instruction "context.route" or "page.route" on our playwright scripts, on Webkit, all css declared with <link rel="preload"... are not loaded on the page but css declared with <link rel="stylesheet"... are properly loaded (as expected). if we remove the instruction context.route (as written on the working snippet), all css are properly loaded.
NB : You will see the obvious difference of style between the working and issued snippets.

yury-s:
[...]

PR #16908, opened 2022-08-30
test: link preload interception

yury-s:
Reference #16745

### DIFF
--- a/tests/page/page-request-fulfill.spec.ts
+++ b/tests/page/page-request-fulfill.spec.ts
@@ -353,3 +353,31 @@ function findResponse(har: har.HARFile, url: string): har.Response {
   expect(entry, originalUrl).toBeTruthy();
   return entry?.response;
 }
+
+it('should continue preload link requests', async ({ page, server, browserName }) => {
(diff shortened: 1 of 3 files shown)


tests/page/page-request-fulfill.spec.ts
```diff
@@ -353,3 +353,31 @@ function findResponse(har: har.HARFile, url: string): har.Response {
   expect(entry, originalUrl).toBeTruthy();
   return entry?.response;
 }
+
+it('should continue preload link requests', async ({ page, server, browserName }) => {
```
dgozman, line 357:
```suggestion
it('should fulfill preload link requests', async ({ page, server, browserName }) => {
```

dgozman (approved):

Meir017:
consider adding matching tests that uses the `link` response header - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link

yury-s:
> consider adding matching tests that uses the `link` response header - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link

We can add that. Looking at the specification and browser compatibility sections on that page it's hard to tell what the adoption of the feature is and whether it's/going to be standardized, so probably less of a problem.

PR #16917, opened 2022-08-30
explain that the `Selectors.register` script is executed in the page context

### LINKED ISSUE #16915: [BUG] `selectors.register` documentation doesn't explain that the script is executed in the browser context
https://playwright.dev/docs/api/class-selectors#selectors-register-option-script

the script gets run in the brwoser context, not nodejs. this can lead to confusion so it should be mentioned in the documentation where the script is executed.

for example, the [`Page.evaluate` documentation](https://playwright.dev/docs/api/class-page#page-evaluate-option-expression) states that the expression is "evaluated in the page context."

DetachHead:
fixes #16915

### DIFF
--- a/docs/src/api/class-selectors.md
+++ b/docs/src/api/class-selectors.md
@@ -45,7 +45,7 @@ const { selectors, firefox } = require('playwright');  // Or 'chromium' or 'webk
 ```
 
 ```java
-// Script that evaluates to a selector engine instance.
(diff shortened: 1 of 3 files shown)


DetachHead:
i don't think the failing ci is my fault since i only changed documentation

pavelfeldman (approved):

PR #16943, opened 2022-08-31
docs: override headers typo

younes-zeboudj:
headers' object must be nested as a value for the key `headers`

### DIFF
--- a/docs/src/api/class-route.md
+++ b/docs/src/api/class-route.md
@@ -49,7 +49,7 @@ await page.route('**/*', (route, request) => {
     foo: 'foo-value', // set "foo" header
     bar: undefined, // remove "bar" header
   };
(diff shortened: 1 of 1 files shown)


docs/src/api/class-route.md
```diff
@@ -49,7 +49,7 @@ await page.route('**/*', (route, request) => {
     foo: 'foo-value', // set "foo" header
     bar: undefined, // remove "bar" header
   };
-  route.continue({headers});
+  route.continue({headers:{...headers});
```
yury-s, line 52: I believe the only difference is that your code creates a copy of the headers map. Since we don't modify I'd leave the more concise version.

mxschmitt:
Closing as per above since there is no user facing change in this pull request.

PR #17004, opened 2022-08-31
feat: introduce the `--ignore-snapshots` option

aslushnikov:
This patch introduces `--ignore-snapshots` Playwright Test CLI option,
as long as `ignoreSnapshots` configuration argument.

### DIFF
--- a/packages/playwright-test/src/matchers/toMatchSnapshot.ts
+++ b/packages/playwright-test/src/matchers/toMatchSnapshot.ts
@@ -251,6 +251,10 @@ export function toMatchSnapshot(
     throw new Error(`toMatchSnapshot() must be called during the test`);
   if (received instanceof Promise)
     throw new Error('An unresolved Promise was passed to toMatchSnapshot(), make sure to resolve it by adding await to it.');
+
+  if (testInfo.config._ignoreSnapshots)
+    return { pass: !this.isNot, message: () => '' };
+
   const helper = new SnapshotHelper(
       testInfo, testInfo.snapshotPath.bind(testInfo), determineFileExtension(received),
       testInfo.project._expect?.toMatchSnapshot || {},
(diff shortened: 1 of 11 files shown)


packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -292,6 +296,10 @@ export async function toHaveScreenshot(
   const testInfo = currentTestInfo();
   if (!testInfo)
     throw new Error(`toHaveScreenshot() must be called during the test`);
+
+  if (testInfo.config._ignoreSnapshots)
+    return { pass: !this.isNot, message: () => '' };
```
dgozman, line 301: I wonder if we could wait for a screenshot or two consecutive matching screenshots to better match real timings.

tests/playwright-test/to-have-screenshot.spec.ts
```diff
@@ -555,6 +555,20 @@ test('should fail on same snapshots with negate matcher', async ({ runInlineTest
   expect(result.output).toContain('Expected result should be different from the actual one.');
 });
 
+test('should not fail if --ignore-snapshots is passed', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    ...playwrightConfig({ screenshotsDir: '__screenshots__' }),
+    '__screenshots__/a.spec.js/snapshot.png': redImage,
+    'a.spec.js': `
+      pwt.test('is a test', async ({ page }) => {
+        await expect(page).not.toHaveScreenshot('snapshot.png', { timeout: 2000 });
```
dgozman, line 564: I think this does not test anything, since screenshot does not match the red image anyway.

dgozman (approved):

packages/playwright-test/src/matchers/toMatchSnapshot.ts
aslushnikov: I'd refrain from this for now.

tests/playwright-test/config.spec.ts
```diff
@@ -373,6 +373,25 @@ test('should inerhit use options in projects', async ({ runInlineTest }) => {
...
+  const result = await runInlineTest({
+    'playwright.config.ts': `
+      module.exports = {
+        ignoreSnapshots: true,
+      };
+    `,
+    'a.test.ts': `
+      const { test } = pwt;
+      test('pass', async ({}, testInfo) => {
+        expect('foo').toMatchSnapshot();
```
mxschmitt, line 386:
```suggestion
        expect('foo').toMatchSnapshot();
        expect('foo').not.toMatchSnapshot();
```

PR #17007, opened 2022-09-01
fix: stop har recording when APIRequestContext is disposed

### LINKED ISSUE #16910: [BUG] ENOENT error for timed out tests with page and request when tracing on failure in Firefox with the html reporter
**Context:**
- Playwright Version: 1.25.1
- Operating System: Mac
- Node.js version: 18.8
- Browser: Firefox

**System:**
 - OS: macOS 12.5.1
 - Memory: 1.79 GB / 64.00 GB

**Binaries:**
 - Node: 18.8.0 - /usr/local/bin/node
 - npm: 8.18.0 - /usr/local/bin/npm

**Languages:**
 - Bash: 3.2.57 - /bin/bash

**Code Snippet**

```typescript

test.describe("reproduces the ENOENT error", async () => {
    test("poll API requests until they time out", async ({ page, request }, testInfo) => {
        await page.goto("https://github.com/microsoft/playwright");
        await expect.poll(
            async () => {
                const searchResults = await request.get(
                    'https://api.github.com/repos/microsoft/playwright/issues',
                );
                return ((await searchResults.json()).length);
            },
            {
                intervals: [1000],
                timeout: testInfo.timeout,
            }
        ).toEqual(0);
    })

})
```

**Describe the bug**

The reporter outputs an ENOENT error to the terminal that looks like `Error: ENOENT: no such file or directory, open '/var/folders/h0/j6c1j6ms36n8g62fsvpt5djm0000gn/T/playwright-tracing-zoHPvZ/f1002bb38a0e7417e5eeb8d6e0c78da5.network'`. It's unclear if there's any impact to the actual HTML report or if the error is spurious.

The provided snippet can be saved to a file and run with `npx playwright test --trace retain-on-failure --browser firefox --reporter html` to reproduce. Repro case may not be 100%, but is fairly high. On my system I have about an 80% reproduction rate. `--timeout` can be used to make the iteration time faster. I've successfully gone as low as a 5 second timeout while keeping a decently high reproduction rate.

yury-s:
Fixes https://github.com/microsoft/playwright/issues/16910

### DIFF
--- a/packages/playwright-core/src/server/browserContext.ts
+++ b/packages/playwright-core/src/server/browserContext.ts
@@ -389,7 +389,7 @@ export abstract class BrowserContext extends SdkObject {
 
       for (const harRecorder of this._harRecorders.values())
         await harRecorder.flush();
-      await this.tracing.flush();
+      await this.tracing.dispose();
 
       // Cleanup.
       const promises: Promise<void>[] = [];
--- a/packages/playwright-core/src/server/browserContext.ts
+++ b/packages/playwright-core/src/server/browserContext.ts
@@ -216,7 +216,7 @@ export abstract class BrowserContext extends SdkObject {
     this._closedStatus = 'closed';
     this._deleteAllDownloads();
     this._downloads.clear();
-    this.tracing.dispose();
+    this.tracing.dispose().catch(() => {});
     if (this._isPersistentContext)
       this.onClosePersistent();
     this._closePromiseFulfill!(new Error('Context closed'));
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/server/browserContext.ts
```diff
@@ -389,7 +389,7 @@ export abstract class BrowserContext extends SdkObject {
 
       for (const harRecorder of this._harRecorders.values())
         await harRecorder.flush();
-      await this.tracing.flush();
+      await this.tracing.dispose();
```
dgozman, line 392: Should we `catch()` the dispose call in `_didCloseInternal`?

dgozman (approved):

Issue #12911, opened 2022-03-20
[Feature] Option to not follow redirects in the requests API


Meemaw:
The `request` fixture methods seems to follow the redirects by default. It would be great if there was an option to disable this behaviour.

```js
const response = await request.get('http://localhost:3000/url', { followRedirects: false })
expect(response.status()).toBe(302)
expect(response.headers()['Location']).toBe('/login')
```

mxschmitt:
@Meemaw what is the use-case for it?

Meemaw:
@mxschmitt We're using Playwright to test an application with server side rendering. Some of the tests are testing that our server correctly redirects you based on your permissions (e.g redirection to `/login` when you visit some authenticated page).

We could test that by just navigating to that page using `page.goto`, but that is not the same. That would have passed even if the navigation client side. In our case, we want to test the server returns status code 302 and appropriate redirect. There tests are also much "cheaper" as there might be a lot of them.

raimohanska:
We have a similar use case and would greatly appreciate this. We mostly use `page.request` which is an `APIRequestContext` object.

franzvezuli:
One use case that I use is:

-- Inspect the first call has a status of 302
-- Inspect the `location` header is correct

vincenzo-gasparo:
I have similar use cases too, I often need to check the redirect location for some auth routes and this feature would be very useful. Currently we are using another request module to handle this use case.

jessfraz:
Yes want to check the set-cookie header on a 302 redirect to intercept it and set it via playwright since chromium blocks third-party cookies in incognito mode (aka playwright default settings)

mcdehaan:
Same as the people above me. I'm testing for the 302 status, location in the header and correct cookies. My workaround is using Axios but I try to replace it entirely with Playwright.

PR #17033, opened 2022-09-02
feat: added follow and redirect arguments to fetch

vincenzo-gasparo:
Reference issue: #12911 
This could be an example of this feature implementation, the behavior is the same as node-fetch library

### DIFF
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -398,6 +398,18 @@ set to `application/octet-stream` if not explicitly set.
 
 Whether to ignore HTTPS errors when sending network requests. Defaults to `false`.
 
+## js-python-fetch-option-follow
+* langs: js
+- `follow` <[int]>
+
+Maximum number of request allowed redirects. Defaults to `20`.
+
+## js-python-fetch-option-redirect
+* langs: js
+- `redirect` ?<[RedirectMode]<"manual"|"follow"|"error">>
+
+A string indicating whether request follows redirects, results in an error upon encountering a redirect, or returns the redirect.
+
 ## evaluate-expression
 - `expression` <[string]>
 
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -398,17 +398,11 @@ set to `application/octet-stream` if not explicitly set.
 
 Whether to ignore HTTPS errors when sending network requests. Defaults to `false`.
 
-## js-python-fetch-option-follow
+## js-python-fetch-option-maxredirects
 * langs: js
-- `follow` <[int]>
+- `maxRedirects` <[int]>
 
-Maximum number of request allowed redirects. Defaults to `20`.
-
-## js-python-fetch-option-redirect
-* langs: js
-- `redirect` ?<[RedirectMode]<"manual"|"follow"|"error">>
-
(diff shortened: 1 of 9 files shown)


docs/src/api/params.md
```diff
@@ -398,6 +398,18 @@ set to `application/octet-stream` if not explicitly set.
 
 Whether to ignore HTTPS errors when sending network requests. Defaults to `false`.
 
+## js-python-fetch-option-follow
+* langs: js
+- `follow` <[int]>
+
+Maximum number of request allowed redirects. Defaults to `20`.
+
+## js-python-fetch-option-redirect
```
yury-s, line 407: I don't think we need 2 separate options for this and also 'error' mode can be turned into an assert on the call site. Let's introduce `maxRedirects` option instead, it would mean max number of redirect responses followed automatically. If set to 0 then the first response is always returned and that would cover the scenario requested in the bug.

tests/library/global-fetch.spec.ts
```diff
@@ -96,6 +96,47 @@ it('should propagate extra http headers with redirects', async ({ playwright, se
...
 });
 
+it('should throw an error when max redirects are exceeded', async ({ playwright, server }) => {
+  server.setRedirect('/a/redirect1', '/b/c/redirect2');
+  server.setRedirect('/b/c/redirect2', '/b/c/redirect3');
+  server.setRedirect('/b/c/redirect3', '/b/c/redirect4');
+  server.setRedirect('/b/c/redirect4', '/simple.json');
+
+  const request = await playwright.request.newContext();
+  for (const follow of [1, 2]) await expect.soft(async () => request.get(`${server.PREFIX}/a/redirect1`, { follow: follow })).rejects.toThrow('Max redirect count exceeded');
```
yury-s, line 106: All the asserts should be hard, we don't need to continue if any operation fails.

yury-s (changes requested):

docs/src/api/params.md
vincenzo-gasparo: Hi, yes my first attempt was using only a `maxRedirects` option, but the default behavior is to fail when maxRedirects counter reach 0:

- When I set the option to a value > 0 then the right behavior is to throw an exception when `maxRedirects === 0`
- What if I want to obtain the exact initial 3xx request ('manual' case)? If I pass 0 as maxRedirect the same logic should apply and an exception is thrown without returning the desired response

Should I handle the zero case as 'don't follow the request' and still manage to let the function throw when I pass maxRedirects > 0 and the counter reach 0 or should I change the type to `'manual'  |  number` for example?

yury-s: Let's only throw in case `maxRedirects` option is not set. This will be equivalent to the current behavior and will throw in case of an infinite redirect. If `maxRedirects` is set the method will not throw, instead it will return last response in the chain even if it is another redirect. The client can alway assert that it has certain status code. This change won't break existing clients that don't set the option and will address the use case with redirect inspection.

vincenzo-gasparo: I added the `maxRedirects` option in the latest commit, let me know what you think, thanks

yury-s (approved):

PR #17046, opened 2022-09-02
docs: merge auth docs

debs-obrien:


docs/src/auth.md
```diff
@@ -5,16 +5,618 @@ title: "Authentication"
 
 Playwright can be used to automate scenarios that require authentication.
```
dgozman, line 7: Sketch for a new doc:
- copy stuff from test-auth.md as is;
- Add the following text somewhere in "reuse" section:
```md
Web apps use cookie-based or token-based authentication, where authenticated state is stored as [cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies) or in [local storage](https://developer.mozilla.org/en-US/docs/Web/API/Storage). Playwright provides [browserContext.storageState([options])](https://playwright.dev/docs/api/class-browsercontext#browser-context-storage-state) method that can be used to retrieve storage state from authenticated contexts and then create new contexts with prepopulated state.

Cookies and local storage state can be used across different browsers. They depend on your application's authentication model: some apps might require both cookies and local storage.
```
- copy "session storage" and "persistent authentication" from auth.md;
- copy https://playwright.dev/docs/auth#reuse-authentication-state as "reuse authentication state with playwright test runner" at the end.

dgozman (approved):

PR #17070, opened 2022-09-04
test(ct): testing typescript types with tsd

sand4rt:
@pavelfeldman do you agree that this could be helpful? otherwise i'll not continue with this


pavelfeldman:
We are running tsc to check our types in other tests. Overall, I don't think this is too useful at this point. Maybe closer to when we approach stable semver.

PR #17082, opened 2022-09-05
fix(test-runner): correct currentStep

vovsemenv:
We have `currentStep` field inside `TestInfo` but we set this filed only for `ExpectMetaInfoProxy`.

This PR fix this behavior and set correct currentTest for all addStep calls.

### DIFF
--- a/packages/playwright-test/src/expect.ts
+++ b/packages/playwright-test/src/expect.ts
@@ -204,7 +204,6 @@ class ExpectMetaInfoProxyHandler {
         canHaveChildren: true,
         forceNoParent: false
       });
-      testInfo.currentStep = step;
 
       const reportStepError = (jestError: Error) => {
         const message = jestError.message;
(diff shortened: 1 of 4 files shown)


mxschmitt:
Would it be possible to add a test for it?

vovsemenv:
Yessir. I'll do it.

vovsemenv:
@mxschmitt is it ok if we add `currentStepInfo` util for testsSteps. it we be the same as `currentTestInfoValue` ?

vovsemenv:
ok, ci passed when I asked very nicely 😅

vovsemenv:
@mxschmitt waiting for your review

dgozman:
@vovsemenv It seems like you are using `testInfo.currentStep` somewhere? As far as I can see, it is not in the public api reference. I think it was mistakenly put on `TestInfo` without and underscore. Could you please file an issue that explains your usage, and we'll consider making it public?

vovsemenv:
@dgozman actually i want to use it in internal feature that allow attach files to steps.

https://github.com/microsoft/playwright/pull/17084

pavelfeldman:
My understanding was that https://github.com/microsoft/playwright/pull/17084#issuecomment-1240897142 was sufficient for your needs. Closing this one in favor of pushing attachments in time.

PR #17084, opened 2022-09-05
feat(test-runner) correct testResult onStepEnd and onStepBegin

### LINKED ISSUE #11592: [Feature] Step-scoped attachments
In my current project, we starting the migration to Playwright (with playwright-test runner).
I've managed to build a decorator function that marks all methods of my Page Object as a test step, but it would be nice if I could add arguments that were passed to the corresponding method as attachments to the step in the report(ideally if it will be also displayed same way in allure HTML report).

### LINKED ISSUE #14364: [BUG] Can't get actual test result info in onStepEnd
**Context:**
- Playwright Version: 1.22.2
- Operating System: Mac
- Node.js version: 16
- Browser: Chromium
- 
**Code Snippet**

I have custom reporter like this
```javascript
class CustomReporter {
    onStepEnd(_test, _result, step) {
        console.log(_result.attachments)   
    }
    onTestEnd(_test, _result, step) {
        console.log(_result.attachments)   
    }
}
```
and test like this
```javascript
test('testing attachments', async ({}, testInfo) => {
    testInfo.attach('4', {body:'444'})
});
```

output
```
❯ npm t

> new-project@1.0.0 test
> playwright test

[]
[]
[
  {
    name: '4',
    path: undefined,
    contentType: 'text/plain',
    body: <Buffer 34 34 34>
  }
]
```

**Describe the bug**
in onStepEnd hook i have `test` but when try to check current attachments always get empty array and attachments appear only in onTestEnd step.

### LINKED ISSUE #17038: [Feature] Step-scoped attachments
Request from Allure Team (approved by @pavelfeldman , please do not close)

Related issues: https://github.com/microsoft/playwright/issues/14364 and https://github.com/microsoft/playwright/issues/11592

PS we have multiple requests for such feature from our paid customers, for us this issue is really high priority as it makes attachments feature unusable in most cases

vovsemenv:
## How it works:
If you use `testInfo.attach()` inside step attachment will be flagged as `isStepAttachment` and appended to `stepAttachments` field inside `step`.

Reporters can access current step attachments
```ts
onStepEnd(test: TestCase, result: TestResult, step: TestStep) {
  console.log(step.stepAttachments)
}
```

depends on #17082

fixes #17038 
fixes #14364
fixes #11592

### DIFF
--- a/packages/playwright-test/src/dispatcher.ts
+++ b/packages/playwright-test/src/dispatcher.ts
@@ -17,7 +17,7 @@
 import child_process from 'child_process';
 import path from 'path';
 import { EventEmitter } from 'events';
-import type { RunPayload, TestBeginPayload, TestEndPayload, DonePayload, TestOutputPayload, WorkerInitParams, StepBeginPayload, StepEndPayload, SerializedLoaderData, TeardownErrorsPayload, WatchTestResolvedPayload, WorkerIsolation } from './ipc';
+import type { RunPayload, TestBeginPayload, TestEndPayload, DonePayload, TestOutputPayload, WorkerInitParams, StepBeginPayload, StepEndPayload, SerializedLoaderData, TeardownErrorsPayload, WatchTestResolvedPayload, WorkerIsolation, SerializedTestResult } from './ipc';
 import type { TestResult, Reporter, TestStep, TestError } from '../types/testReporter';
 import type { Suite } from './test';
 import type { Loader } from './loader';
@@ -263,7 +263,7 @@ export class Dispatcher {
       (parentStep || result).steps.push(step);
       if (params.canHaveChildren)
         stepStack.add(step);
-      this._reporter.onStepBegin?.(data.test, result, step);
+      this._reporter.onStepBegin?.(data.test, testResultFromSerializedInfo(params.serializedTestResult, runData.result.steps), step);
     };
     worker.on('stepBegin', onStepBegin);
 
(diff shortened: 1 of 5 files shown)


pavelfeldman:
I believe the plan was to make sure testInfo.attachments are up-to-date in the onStepBegin / onStepEnd calls.

@baev ^^

vovsemenv:
@pavelfeldman i changed PR and i did what we originally planned.

vovsemenv:
But i think we also need to add attachments in `TestStep`. This will enable Reporters to check `step.attachments` field directly rather than implementing their own logic with `TestResult.attachments` diffs between multiple `onStepEnd` calls.

packages/playwright-test/src/dispatcher.ts
```diff
@@ -17,13 +17,14 @@
...
 import { EventEmitter } from 'events';
-import type { RunPayload, TestBeginPayload, TestEndPayload, DonePayload, TestOutputPayload, WorkerInitParams, StepBeginPayload, StepEndPayload, SerializedLoaderData, TeardownErrorsPayload, WatchTestResolvedPayload, WorkerIsolation } from './ipc';
+import type { RunPayload, TestBeginPayload, TestEndPayload, DonePayload, TestOutputPayload, WorkerInitParams, StepBeginPayload, StepEndPayload, SerializedLoaderData, TeardownErrorsPayload, WatchTestResolvedPayload, WorkerIsolation, SerializedTestResult } from './ipc';
 import type { TestResult, Reporter, TestStep, TestError } from '../types/testReporter';
 import type { Suite } from './test';
 import type { Loader } from './loader';
 import { TestCase } from './test';
 import { ManualPromise } from 'playwright-core/lib/utils/manualPromise';
 import { TestTypeImpl } from './testType';
+import { stdioEntryDecode } from './reporters/json';
```
pavelfeldman, line 27: We don't want dependencies to go from the code of the test runner to the reporter, consider moving required method into the utilities file if you need it here. I'm not sure why you would need it though.

packages/playwright-test/src/dispatcher.ts
```diff
@@ -606,3 +607,19 @@ function chunkFromParams(params: TestOutputPayload): string | Buffer {
     return params.text;
   return Buffer.from(params.buffer!, 'base64');
 }
+
+function testResultFromSerializedInfo(serializedTestResult: SerializedTestResult, steps: TestStep[]): TestResult {
+  return {
```
pavelfeldman, line 612: TestResult identity is important, you should always pass the same instance to the user.

packages/playwright-test/src/ipc.ts
```diff
@@ -77,6 +77,19 @@ export type StepBeginPayload = {
   forceNoParent: boolean;
   wallTime: number;  // milliseconds since unix epoch
   location?: { file: string, line: number, column: number };
+  serializedTestResult: SerializedTestResult;
+};
+
+export type SerializedTestResult  = Omit<TestResult, 'attachments' | 'startTime' | 'stderr' | 'stdout' | 'steps' > & {
```
pavelfeldman, line 83: It seems like you have these new types and utilities that were not necessary. I'd expect that the only thing this PR does is makes sure that TestResult has up-to-date collection of attachments when onStepBegin/End are invoked.

pavelfeldman:
Closing as per my comments above.

PR #17175, opened 2022-09-08
fix(reuse): restore storage state

### LINKED ISSUE #17160: [BUG] StorageState is lost after first test (with VSCode ☑️ Open Browsers)
**Context:**
- Playwright Version: [what Playwright version do you use?]
- Operating System: [e.g. Windows, Linux or Mac]
- Node.js version: [e.g. 12.22, 14.6]
- Browser: Chromium
- Extra: Could also reproduce this on Windows

## System:
 - OS: macOS 12.4
 - Memory: 1.09 GB / 32.00 GB
## Binaries:
 - Node: 14.20.0 - ~/.nvm/versions/node/v14.20.0/bin/node
 - Yarn: 1.22.19 - ~/.nvm/versions/node/v14.20.0/bin/yarn
 - npm: 6.14.17 - ~/.nvm/versions/node/v14.20.0/bin/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash
## npmPackages:
 - playwright: ^1.25.1 => 1.25.1 

## Code

```ts
// global-setup.ts
import { chromium, FullConfig } from '@playwright/test';

async function globalSetup(config: FullConfig) {
  await page.goto(`localhost:4242/login`);

  await page.evaluate(() => localStorage.setItem('auth', 'true'));

  await page.context().storageState({ path: 'storage.json' });
  await browser.close();
}

export default globalSetup;
```

```ts
// tests/first.spec.ts
import { expect, test } from '@playwright/test';

test('should login', async ({ page }) => {
  await page.goto('/');
  expect(await page.evaluate(() => localStorage.getItem('auth'))).toEqual('true');
});

test('should login 2', async ({ page }) => {
  await page.goto('/');
  expect(await page.evaluate(() => localStorage.getItem('auth'))).toEqual('true');
});

// repeat the above multiple times
```

```ts
// playwright.config.ts
import type { PlaywrightTestConfig } from '@playwright/test';

  use: {
    viewport: { width: 1280, height: 720 },
    ignoreHTTPSErrors: true,
    baseURL: 'http://localhost:4242/',
    storageState: require.resolve('./storage.json'),
  },
};
export default config;
```

## Bug

The tests above can be executed fine if different browser windows are used (VSCode extension):
image

But when the VSCode extension is set to "☑️ Show Browsers" only the first test will succeed, and all others will fail. Make sure to hit "❎ Close Browsers" before trying this:
image

---

I already tried to debug it, and it seems that the browser is not restoring the storage state for subsequent runs. It took quite a while  to narrow this down to this use-case, but the code above is exactly what our test-suite is containing.

We also have the same problem on weaker machines even without the VSCode extension, but I'm not yet sure if that is connected to this.

**Addition:** When I single out any of these tests they work fine, but all together it's not working (given that the show browsers option is set). Additionally, we also have the same issue occur in our real, more complex test-suite even without the  "Show Browsers" option ticked.

Thanks for this great framework 👏

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/17160


packages/playwright-core/src/server/browserContext.ts
```diff
@@ -484,15 +486,35 @@ export abstract class BrowserContext extends SdkObject {
...
+
     await page._setServerRequestInterceptor(undefined);
+
+    // No longer interested in these origins.
+    this._origins.clear();
     // It is safe to not restore the URL to about:blank since we are doing it in Page::resetForReuse.
   }
 
+  async _resetCookies() {
+    this.clearCookies();
```
dgozman, line 513: await here and below?

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -484,15 +486,35 @@ export abstract class BrowserContext extends SdkObject {
     await page._setServerRequestInterceptor(handler => {
       handler.fulfill({ body: '<html></html>' }).catch(() => {});
     });
-    for (const origin of this._origins) {
+
+    // Remove/replace old origins.
```
dgozman, line 490: Could be a single loop?

```ts
for (const origin of new Set([...oldOrigins, ...newOrigins.keys()])) {
  ...
  await frame.resetStorageForCurrentOriginBestEffort(newOrigins.get(origin));
}
```

dgozman (approved):

PR #17187, opened 2022-09-08
chore: have a socket timeout when downloading browsers

mxschmitt:
This would have prevented the following: https://github.com/microsoft/playwright.dev/runs/8235096615?check_suite_focus=true

Should we instead of manually doing it inside download.ts maybe put a default timeout if not set to all our http.get/http.request requests?

### DIFF
--- a/tests/installation/playwright-cdn-should-race-with-timeout.spec.ts
+++ b/tests/installation/playwright-cdn-should-race-with-timeout.spec.ts
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


tests/installation/playwright-cdn-should-race-with-timeout.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+import http from 'http';
+import type { AddressInfo } from 'net';
+import { test, expect } from './npmTest';
+
+test(`playwright cdn should race with a timeout`, async ({ exec }) => {
+  test.slow(); // This test will timeout on all the 3 fallback CDNs -> 30 seconds duration.
+  const server = http.createServer(() => {});
+  await new Promise<void>(resolve => server.listen(0, resolve));
```
aslushnikov, line 23: shall we close the server in the end of the test?

aslushnikov (approved):

Issue #14790, opened 2022-06-10
[BUG] Error: Failed to download chromium v1000


lszomoru:
We have started seeing intermittent issues while trying to download Playwright in our Azure Pipelines:

```
yarn run v1.22.19
$ /__w/1/s/node_modules/.bin/npm-run-all -lp 'electron x64' playwright-install
[electron x64      ] $ node build/lib/electron x64
[playwright-install] $ node build/azure-pipelines/common/installPlaywright.js
[playwright-install] /__w/1/s/node_modules/playwright-core/lib/server/registry/index.js:613
[playwright-install]       throw new Error(`Failed to download ${title}, caused by\n${e.stack}`);
[playwright-install]             ^
[playwright-install] 
[playwright-install] Error: Failed to download chromium v1000, caused by
[playwright-install] Error: Client network socket disconnected before secure TLS connection was established
[playwright-install]     at connResetException (node:internal/errors:692:14)
[playwright-install]     at TLSSocket.onConnectEnd (node:_tls_wrap:1587:19)
[playwright-install]     at TLSSocket.emit (node:events:539:35)
[playwright-install]     at endReadableNT (node:internal/streams/readable:1345:12)
[playwright-install]     at processTicksAndRejections (node:internal/process/task_queues:83:21)
[playwright-install]     at /__w/1/s/node_modules/playwright-core/lib/server/registry/index.js:613:13
[playwright-install]     at async Registry._downloadExecutable (/__w/1/s/node_modules/playwright-core/lib/server/registry/index.js:612:5)
[playwright-install]     at async Registry.install (/__w/1/s/node_modules/playwright-core/lib/server/registry/index.js:590:9)
[playwright-install]     at async installBrowsersForNpmInstall (/__w/1/s/node_modules/playwright-core/lib/server/registry/index.js:780:3)
[playwright-install]     at async retry (/__w/1/s/build/azure-pipelines/common/retry.js:12:20)
[playwright-install]     at async install (/__w/1/s/build/azure-pipelines/common/installPlaywright.js:10:5)
[playwright-install] error Command failed with exit code 1.
[playwright-install] info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: "playwright-install" exited with 1.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Are there any known issues with the infrastructure hosting Playwright? Thank you!

bpasero:
[...]

PR #17196, opened 2022-09-08
chore: improve browser download failure message

s-h-a-d-o-w:
I realize that this is an unsolicited PR and maybe you would want to use a different URL but given that this happens [here and there](https://github.com/microsoft/playwright/issues/14790), I thought I would just go for it.
At least I would assume that the outage scenario is the most likely and so I think it would be great to give users this hint instead of making them search the GitHub issues. 😅 

Related to: https://github.com/microsoft/playwright/issues/14790

### DIFF
--- a/packages/playwright-core/src/server/registry/index.ts
+++ b/packages/playwright-core/src/server/registry/index.ts
@@ -748,7 +748,7 @@ export class Registry {
 
     const downloadFileName = `playwright-download-${descriptor.name}-${hostPlatform}-${descriptor.revision}.zip`;
     await downloadBrowserWithProgressBar(title, descriptor.dir, executablePath, downloadURLs, downloadFileName).catch(e => {
-      throw new Error(`Failed to download ${title}, caused by\n${e.stack}`);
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/registry/index.ts
```diff
@@ -748,7 +748,7 @@ export class Registry {
 
     const downloadFileName = `playwright-download-${descriptor.name}-${hostPlatform}-${descriptor.revision}.zip`;
     await downloadBrowserWithProgressBar(title, descriptor.dir, executablePath, downloadURLs, downloadFileName).catch(e => {
-      throw new Error(`Failed to download ${title}, caused by\n${e.stack}`);
+      throw new Error(`Failed to download ${title}, caused by\n${e.stack}\nIf this persists, please check here for a possible outage: https://azure.status.microsoft/en-us/status`);
```
pavelfeldman, line 751: As of 1.22 we are going to try 3 different CDN networks, so Microsoft's one being down should not affect Playwright's ability to fetch the browsers. If you are still experiencing this issue, chances are you are on an old Playwright.

s-h-a-d-o-w: Thanks for the update, I'm glad to hear that you've taken steps to improve resilience!

(We currently happen to be on 1.21.0 in the affected project. 😅)

PR #17232, opened 2022-09-10
feat(ct): type-safe hook config

sand4rt:
`JsonObject` is from: https://github.com/sindresorhus/type-fest/blob/main/source/basic.d.ts


pavelfeldman:
Lint issue, otherwise looks good

pavelfeldman (approved):

pavelfeldman (approved):

PR #17238, opened 2022-09-10
feat(ct): solid type-safe hook config

sand4rt:


pavelfeldman:
Let's merge all remaining type-safes into one PR to save on the bots hours

PR #17244, opened 2022-09-11
feat: allow wildcards in the project option

### LINKED ISSUE #15128: [Feature] Allow to use wildcard in --project <name>
I have a `playwright.config.ts` in which I define 16 projects.

- 5x Tests running in Chrome in 5 different view port sizes
- 5x Tests running in Firefox in 5 different view port sizes
- 5x Tests running in Webkit in 5 different view port sizes
- 1x API test

Each test has a different name but follows a specific schema. Example:

- `ui-small-chrome`
- `ui-large-firefox`
- `api`

To run all tests, expect the `api` tests I need to use `npx playwright --project=ui-small-chrome --project=ui-large-firefox --project=...`.

What I would like to do is run `npx playwright --project=ui-*`.

I would expect this command to run all projects whose name start with `ui`.

nate-ro:
Fixes: #15128


packages/playwright-test/src/runner.ts
```diff
@@ -243,6 +243,13 @@ export class Runner {
         return true;
       const name = project.name.toLocaleLowerCase();
       unknownProjects!.delete(name);
```
pavelfeldman, line 245: This comment is not about your code, but rather the code that used to be here.

I have to admit that I'm struggling with following the logic here. Maybe the naming of the variables is off, maybe the logic itself is unnecessarily complex. Could you go over the large blocks here and add comments explaining what the code is doing? Ideally we should refactor it to make sense, but adding comments is a good start.

packages/playwright-test/src/runner.ts
```diff
@@ -243,6 +243,13 @@ export class Runner {
...
+      for (const value of projectsToFind) {
+        const match = new RegExp('^' + value.replace(/\*/g, '.*').replace(/\?/g, '.?') + '$');
+        if (match.test(name)) {
+          unknownProjects!.delete(value);
```
pavelfeldman, line 249: So I guess `unknownProjects` means `unmatchedProjectFilters`, let's rename it.

packages/playwright-test/src/runner.ts
```diff
@@ -243,6 +243,13 @@ export class Runner {
...
+          return true;
+        }
+      }
       return projectsToFind.has(name);
```
pavelfeldman, line 253: This seems to be no longer needed.

packages/playwright-test/src/runner.ts
```diff
@@ -243,6 +243,13 @@ export class Runner {
         return true;
       const name = project.name.toLocaleLowerCase();
       unknownProjects!.delete(name);
+      for (const value of projectsToFind) {
+        const match = new RegExp('^' + value.replace(/\*/g, '.*').replace(/\?/g, '.?') + '$');
```
pavelfeldman, line 247: How do you feel about using regex here from the get go? The users would need to put `foo.*` to include all projects with prefix foo, but they get rich matching consistent with our test filter and `--grep` treatment. There is a `forceRegExp` method that is used for command like args that are considered regex.

packages/playwright-test/src/runner.ts
```diff
@@ -243,6 +243,13 @@ export class Runner {
         return true;
       const name = project.name.toLocaleLowerCase();
       unknownProjects!.delete(name);
+      for (const value of projectsToFind) {
```
pavelfeldman, line 246: `projectsToFind` used to normalize the names for case-insensitive processing. Now that the project names are regex, you can make regex matching `i`, case-insensitive, so you should not need this `projectsToFind` normalization.

packages/playwright-test/src/runner.ts
nate-ro: Also, I don't quite understand what the purpose of `forceRegExp` is. Won't that require users to do `--project /foo.*/`?

pavelfeldman: > Should I still do ^ and $? Is it okay for --project r to match chromium and firefox?

Oh, that's a good point - we don't want to run `non-basic` when `--project basic` is passed.

> Won't that require users to do --project /foo.*/?

I don't think it will, the way I read it is that it'll drop surrounding slashes if they were provided while constructing the regex

nate-ro: > Oh, that's a good point - we don't want to run `non-basic` when `--project basic` is passed.

`const match = forceRegExp('/^' + value + '$/i');`
Would this be an acceptable solution?

pavelfeldman: Yes, you might want to strip the surrounding // if they are present, other than that it looks fair.

nate-ro: I think that making the user use \ if their project is `/foo/` is worse than them having to remove their // if they try to do a normal regex search.

pavelfeldman: I'm not sure I follow this

nate-ro: If you strip surrounding //, then if a user has a project that's called `/foo/` they won't be able to find it unless they do `--project //foo//` or `--project \/foo\/`.

If you don't strip them, then if a user is trying to use `--project /foo/` to find project `foo`, they'll have to remove the `//` to find it.

Both cases are unlikely, but I didn't strip them because I think the second case is easier for the user to handle. I might be wrong, though.

tests/playwright-test/runner.spec.ts
```diff
@@ -449,3 +449,18 @@ test('should not crash with duplicate titles and line filter', async ({ runInlin
...
+  const result = await runInlineTest({
+    'playwright.config.ts': `
+    module.exports = { name: 'project-name' };
+    `,
+    'a.test.js': `
+      const { test } = pwt;
+      test('one', async ({}) => {
+        expect(1).toBe(1);
+      });    `
+  }, { project: '.*oj.*t-na.?e' }, {}, {});
```
yury-s, line 463: Let's also add a test where one of the filters doesn't match any projects.

yury-s, line 463: Nit: '.*oj.*t-Na.?e' or `name: 'Project-Name'` to make sure the match is case-insensitive.

yury-s (approved):

yury-s (approved):
You'll need to rebase to resolve the conflict in packages/playwright-test/src/runner.ts so that we could merge this PR.

packages/playwright-test/src/runner.ts
```diff
@@ -619,11 +618,15 @@ function filterByFocusedLine(suite: Suite, focusedTestFileLines: TestFileFilter[
   if (!filterWithLine)
     return;
 
-  const testFileLineMatches = (testFileName: string, testLine: number, testColumn: number) => focusedTestFileLines.some(filter => {
-    const lineColumnOk = (filter.line === testLine || filter.line === null) && (filter.column === testColumn || filter.column === null);
+  const testFileLineMatches = (testFileName: string, testLine: number, testColumn: number) => focusedTestFileLines.some(({ re, exact, line, column }) => {
+    const lineColumnOk = (line === testLine || line === null) && (column === testColumn || column === null);
     if (!lineColumnOk)
       return false;
-    return createFileMatcherFromFilters([filter])(testFileName);
```
yury-s, line 626: Why did this change? It was introduced in a recent bug fix that unified the matchers, unless there is a need for that let's revert this part of the PR.

tests/playwright-test/runner.spec.ts
yury-s: It can be something similar to [this test](https://github.com/microsoft/playwright/blob/76cffad023cc01f5f85d16b05d48cc32cd7fa302/tests/playwright-test/runner.spec.ts#L31-L32) which would assert the output contains expected error message.

yury-s (changes requested):
Please add a test with non matching filter and revert unrelated changes.

packages/playwright-test/src/runner.ts
nate-ro: sorry, I messed up trying to rebase things

yury-s (changes requested):
"Files changed 538" - looks like your rebase didn't work well and picked up som unrelated changes to the PR. Please fix.

pavelfeldman:
Seems like we got stuck with this one! Let me close as stale.

PR #17249, opened 2022-09-11
feat(ct): vue mount type

### LINKED ISSUE #17206: [Feature] Derive the prop type from the Vue/Svelte component
Would be great if the type of prop could be derived from the component _(`mount(Component)`)_  instead of explicitly defining it through the generic _(`mount<Props>`)_. Vue test utils also managed to do this: https://github.com/vuejs/test-utils/blob/main/test-dts/mount.d-test.ts#L60.

```vue
<script lang="ts" setup>
  defineProps<{ title: string }>();
</script>
<template>
  <button>{{ title }}</button>
</template>
```

```typescript
test('should not work', async ({ mount }) => {
  await mount(Button, { props: {} }) // Should throw a type error because title is a required prop
})
```

- [x] Svelte
- [x] Vue 3
- [x] Vue 2

sand4rt:
closes: https://github.com/microsoft/playwright/issues/17206


pavelfeldman:
This looks too involved to maintain.

sand4rt:
@pavelfeldman 
Deriving the types from each component style _(class, functional, define and setup component)_ in Vue is just hard to do, but would be really great to have it in place. Note that i copied the types from Vue-Test-Utils: https://github.com/vuejs/test-utils/blob/main/src/mount.ts#L97. They have this in production for a while and we can use this as an example which makes maintenance a bit easier.

packages/playwright-ct-vue/index.d.ts
```diff
@@ -41,26 +60,181 @@ type JsonObject = { [Key in string]?: JsonValue };
...
+ * The types for mount have been copied directly from the VTU mount
+ * https://github.com/vuejs/vue-test-utils-next/blob/master/src/mount.ts.
+ *
+ * There isn't a good way to make them generic enough that we can extend them.
+ *
+ * In addition, we modify the types slightly.
+ *
+ * Changed MountingOptions<...> to MountOptions<...>
+ * and VueWrapper<...> to MountResult<...>.
+ */
```
sand4rt, line 111: @pavelfeldman 
Added a comment to keep it a little easier to maintain. I'm not sure if there's a better way.. Seems like Cypress has taken the same approach.

sand4rt:
@lmiller1990 This is the simplest thing i could think of. It doesn't support class components _(think this is not supported with Vue3 anyway)_. Do you see any problems with this? I tested it with:

- Functional components
- Component declared with defineComponent
- Component declared by vue-tsc ScriptSetup
- Component declared with no props
- Component declared with { props: { ... } }

```typescript
type Slot = string | string[];

type VueComponent = abstract new (...args: any) => { $props: any };

type ComponentProps<Component extends VueComponent | any> = Component extends VueComponent ? InstanceType<Component>['$props'] : any;

export interface MountOptions<Component extends VueComponent | any> {
  props?: ComponentProps<Component>;
  slots?: Record<string, Slot> & { default?: Slot };
}
````

pavelfeldman:
Let's wait with it for now?

Issue #14241, opened 2022-05-18
[Feature] Web Component (Lit) support for test-ct


alexkrolick:
Feature request - support for web components in "component test" mode.

Looking into the Svelte/React/Vue [implementation](https://github.com/microsoft/playwright/blob/main/packages/playwright-ct-svelte/index.js#L26), I think it would be even simpler to do for native web components since there is no framework-specific Vite plugin needed, even if using a library like Lit.  You might even consider web components the base case for component tests, since all you need is the environment setup/teardown; this would open up experimentation with lesser-known frameworks that aren't going to get an official plugin since they can do some work in userspace for "mount" style commands.

bjarkihall:
If I understand the current architecture, the `test` function is extended with a mount fixture on import.
So each plugin makes sure that this signature:
```ts
const locator = await mount(comp: T, opts?: U);
```
uses `comp` and `opts` to render something to the `#root` element and return a locator to it.
Wouldn't a more general "base case" be a (primitive | function | HTMLElement | Array) -> HTML?

```tsx
// Primitive value -> root.innerHTML = (comp ?? '').toString():
mount(undefined) -> <div id="root"></div>
mount(0) -> <div id="root">0</div>
mount('hello') -> <div id="root">hello</div>
mount('<button>Click me</button>') -> <div id="root"><button>Click me</button></div>

// Object instanceof HTMLElement -> root.appendChild(comp):
mount(document.createElement('div')) -> <div id="root"><div></div></div>
// That should also work for Custom Elements (Web Components) if you've registered it:
mount(new MyComponent()) -> <div id="root"><my-component></my-component></div>
// A base-plugin could also auto-register custom elements if they're not already:
//     if(!customElements.get(compName)) customElements.define(compName, comp);

// Functions could just mount the awaited return value:
// Note: if return value is also a function, throw Error to prevent recursion
mount(() => 'hello') -> <div id="root">hello</div>
mount(async () => 'hello') -> <div id="root">hello</div>

// Array could just be iterated/joined:
mount([1, 1]) -> <div id="root">11</div>

// If an Object isn't supported, it can just fail the test.
mount({}) -> Error('Unknown component type, did you forget to import test from a plugin?')
mount(new Date()) -> Error('Unknown component type, did you forget to import test from a plugin?')
```

[...]

PR #17272, opened 2022-09-12
feat(ct): initial web components

sand4rt:
Only a simple mount without props is possible. Partial fix for https://github.com/microsoft/playwright/issues/14241


pavelfeldman:
I wonder if we should stabilize the existing set before growing it further.

sand4rt:
Yeah the same thought crossed my mind. Was too curious if it would work..

If we merge, we don't have to release it right? Maybe the other people who ask for it in the issue can start working on it too if there is a base.

Not sure if this is too much hassle. Otherwise i can leave it open or close it, it's up to you

pavelfeldman:
We have the code in the PR, so it won't go anywhere. We can close it for now and wait for the critical mass of lit to appear.

prashantpalikhe:
Would be great to have support for Lit components too with the same simple mount API. 

We are thinking of testing this via Storybook's play function and Playwright. But default Playwright CT will be much faster.

sand4rt:
@prashantpalikhe Made a package for Lit and Web components with the same simple mount API until official support: https://github.com/microsoft/playwright/issues/14241#issuecomment-1368124560. 

BTW. good luck with the rebuild at KPN!😄

PR #17273, opened 2022-09-12
chore(dotnet): no nested namespace in transport channels

mxschmitt:


utils/doclint/generateDotnetApi.js
```diff
@@ -226,11 +226,6 @@ for (const [name, type] of modelTypes)
 for (const [name, literals] of enumTypes)
   renderEnum(name, literals);
 
-if (process.argv[3] !== '--skip-format') {
```
yury-s, line 229: Why was this removed?

yury-s (approved):

utils/doclint/generateDotnetApi.js
mxschmitt: it's done here now: https://github.com/microsoft/playwright-dotnet/blob/main/build.sh#L69

PR #17277, opened 2022-09-12
test: improve debugability of installation tests

dgozman:
Put more details into the error message. This way it is immediately seen on the bots.


rwoll (commented):
Please leave the attachments as well. If it doesn't error, it's still useful to have the output per-command as an attachment when authoring and modifying these tests. I used the captured output when adding new assertions to these tests.

rwoll (approved):

PR #17288, opened 2022-09-13
fix: make evaluate work with overridden Window/Document/Node

### LINKED ISSUE #17287: [BUG] Process exited due to  TypeError: log is not iterable
**Context:**
- Playwright Version: 1.25.0
- .NET version: .NET 6
- Browser: Chromium

**Code Snippet**

```cs
using System.Threading.Tasks;
using Microsoft.Playwright;

class Program
{
    public static async Task Main()
    {
        using var playwright = await Playwright.CreateAsync();
        var sclocatorLocatorOptions = new SelectorsRegisterOptions()
        {
             Script = @"{
query(root, selector) {
    if (typeof isc !== 'undefined' && typeof isc.AutoTest !== 'undefined') {
        return isc.AutoTest.getElement(selector);
    }
    return null;
},
queryAll(root, selector) {
    if (typeof isc !== 'undefined' && typeof isc.AutoTest !== 'undefined') {
        var element = isc.AutoTest.getElement(selector);
        if (element !== null) {
            return [element];
        }
    }
    return [];
}
}"
};
    await playwright.Selectors.RegisterAsync("sclocator", sclocatorLocatorOptions); 
    var chromium = await playwright.Chromium.LaunchAsync();
    var page = await chromium.NewPageAsync();
    await page.GotoAsync("http://localhost:23730/index.html");
    await page.Locator("sclocator=//autoID[Class=HLayout||index=1||length=2||classIndex=0||classLength=1]/fieldRow[Class=FieldRow||scClass=HLayout||index=0||length=1||classIndex=0||classLength=1||scClassIndex=0||scClassLength=1]/field/field/input/item[name=username||title=username||index=0||Class=TextItem]/element")
          .TypeAsync("UserName");
    }
}
```

**Describe the bug**

I'm not sure if this is more of a dotnet bug or Node bug but I'm having issues using the latest version of playwright-dotnet. I'm trying to run a login test and on the call to type the username I'm running into the following error.

```
Microsoft.Playwright.PlaywrightException : Process exited
    C:\tests\bin\Debug\net6.0\.playwright\package\lib\server\dom.js:1025
      for (const entry of log) this._progress.logEntry(entry);
                          ^
TypeError: log is not iterable
    at InjectedScriptPollHandler._streamLogs (C:\tests\bin\Debug\net6.0\.playwright\package\lib\server\dom.js:1025:27)
    at runNextTicks (node:internal/process/task_queues:61:5)
    at processImmediate (node:internal/timers:437:9)
```

The only thing I'd say that is different about the usual setup is I'm using a custom selector since the JS library that the app that I'm testing uses does not make traditional locator methods easy and instead provides its own custom element selectors. I provided the script that I'm using for the custom selector in case that error could be related to the custom selector. I've encountered this error on multiple Playwright versions after 1.21.0 and I'm using that version as a workaround.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/17287

### DIFF
--- a/packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
+++ b/packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
@@ -107,11 +107,11 @@ export function source() {
 
   function serialize(value: any, handleSerializer: (value: any) => HandleOrValue, visitorInfo: VisitorInfo): SerializedValue {
     if (value && typeof value === 'object') {
-      if (globalThis.Window && typeof globalThis.Window === 'function' && value instanceof globalThis.Window)
+      if (typeof globalThis.Window === 'function' && value instanceof globalThis.Window)
         return 'ref: <Window>';
-      if (globalThis.Document && typeof globalThis.Document === 'function' && value instanceof globalThis.Document)
+      if (typeof globalThis.Document === 'function' && value instanceof globalThis.Document)
         return 'ref: <Document>';
-      if (globalThis.Node && typeof globalThis.Node === 'function' && value instanceof globalThis.Node)
+      if (typeof globalThis.Node === 'function' && value instanceof globalThis.Node)
         return 'ref: <Node>';
     }
     return innerSerialize(value, handleSerializer, visitorInfo);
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
```diff
@@ -107,11 +107,11 @@ export function source() {
 
   function serialize(value: any, handleSerializer: (value: any) => HandleOrValue, visitorInfo: VisitorInfo): SerializedValue {
     if (value && typeof value === 'object') {
-      if (globalThis.Window && value instanceof globalThis.Window)
+      if (globalThis.Window && typeof globalThis.Window === 'function' && value instanceof globalThis.Window)
```
yury-s, line 110: you can now drop `globalThis.Window &&` etc as `typeof` will take care of that

yury-s (approved):

Issue #17217, opened 2022-09-09
[Feature] Add ability to specify video params (like fps) for video recordings


rachitm022:
Would be good to specify params for video recordings like fps 

```js
const context = await browser.newContext({
  recordVideo: {
    dir: 'videos/',
    size: { width: 640, height: 480 },
    fps: 25
  }
});
```

HeDo88TH:
This is a missing piece, we need this.

riethmue:
Hi, I need this feature. When do you plan to roll it out? :blush:

pavelfeldman:
#### Why was this issue closed?

Thank you for your involvement. This issue was closed due to limited engagement (upvotes/activity), lack of recent activity, and insufficient actionability. To maintain a manageable database, we prioritize issues based on these factors.

If you disagree with this closure, please open a new issue and reference this one. More support or clarity on its necessity may prompt a review. Your understanding and cooperation are appreciated.

pavelfeldman:
#### Why was this issue closed?

Thank you for your involvement. This issue was closed due to limited engagement (upvotes/activity), lack of recent activity, and insufficient actionability. To maintain a manageable database, we prioritize issues based on these factors.

If you disagree with this closure, please open a new issue and reference this one. More support or clarity on its necessity may prompt a review. Your understanding and cooperation are appreciated.

terion-labs:
years pass...

angel1254mc:
This would be awesome. Our team has been experimenting with using Playwright to automate demo recordings.  The idea is to generate scripted browser walkthroughs that could be regenerated with every release. Configurable FPS would let us produce smoother, more professional output for these kinds of non-test use cases.

Totally acknowledge that Playwright is at the mercy of the browser's devtools protocol when it comes to the count/timing of frames it receives, so we can't really guarantee that a user's desired FPS will be respected.

One approach could be adding an optional `fps` param to `recordVideo` that lets users set a _target_ fps for the output video (defaults to today's 25 fps). If the browser sends enough frames, awesome. Otherwise the user just gets repeated frames to fill the gap, which matches how it works today at the hardcoded 25 fps. Thoughts?

MuTsunTsai:
Hi @rachitm022, @angel1254mc, @terion-name — for the FPS / demo recording use case described here, [`playwright-recorder-plus`](https://github.com/MuTsunTsai/playwright-recorder-plus) might be useful:

[...]

PR #17293, opened 2022-09-13
Feature(video): allow create video with custom fps

vitalics:
Implement video record feature which allows to provide custom frames per second 

Issue link: #17217


vitalics:
close since undiscussed

PR #17307, opened 2022-09-13
feat(docker): auto-bind container ports to host ports

aslushnikov:
Drive-by: make sure docker container does not expose ports on `0.0.0.0`
and instead registers to localhost. This way websocket and vnc ports
are not exposed to the public internet.

### DIFF
--- a/packages/playwright-test/src/docker/docker.ts
+++ b/packages/playwright-test/src/docker/docker.ts
@@ -99,14 +99,14 @@ export async function containerInfo(): Promise<ContainerInfo|undefined> {
   const container = await findRunningDockerContainer();
   if (!container)
     return undefined;
-  const vncBinding = container.portBindings.find(binding => binding.containerPort === 7900);
-  const wsBinding = container.portBindings.find(binding => binding.containerPort === 5400);
-  if (!vncBinding || !wsBinding)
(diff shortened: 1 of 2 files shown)


yury-s (approved):

packages/playwright-test/src/docker/docker.ts
```diff
@@ -96,17 +96,29 @@ interface ContainerInfo {
...
+  const vncBinding = container.portBindings.find(binding => binding.containerPort === 7900);
+  const wsBinding = container.portBindings.find(binding => binding.containerPort === 5400);
+  if (!vncBinding || !wsBinding)
+    return undefined;
+  const logLines = await dockerApi.getContainerLogs(container.containerId);
+
+  const containerUrlToHostUrl = (address: string, portBinding: dockerApi.PortBinding) => {
+    const url = new URL(address);
+    url.host = portBinding.ip;
+    url.port = portBinding.hostPort + '';
```
dgozman, line 111:
```suggestion
    url.port = container.portBindings.find(binding => biding.containerPort === +url.port).hostPort + '';
```

This way you don't need to hardcode 5400 and 7900.

dgozman (approved):

PR #17315, opened 2022-09-13
fix(docker): do not pollute stdout when used with JSON reporter

aslushnikov:
This patch moves parts of docker configuration to a plugin so that
it can rely on repoter for stdout.

Drive-by: provide all plugins with reporter in the `setup` callback.

### DIFF
--- a/packages/playwright-test/src/docker/docker.ts
+++ b/packages/playwright-test/src/docker/docker.ts
@@ -22,7 +22,7 @@ import { spawnAsync } from 'playwright-core/lib/utils/spawnAsync';
 import * as utils from 'playwright-core/lib/utils';
 import { getPlaywrightVersion } from 'playwright-core/lib/common/userAgent';
 import * as dockerApi from './dockerApi';
-import { addRunnerPlugin } from '../plugins';
+import type { TestRunnerPlugin } from '../plugins';
 import type { FullConfig, Reporter, Suite } from '../../types/testReporter';
 
 const VRT_IMAGE_DISTRO = 'focal';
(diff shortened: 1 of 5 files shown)


packages/playwright-test/src/runner.ts
```diff
@@ -603,14 +603,14 @@ export class Runner {
...
-    this._plugins.push(...webServerPluginsForConfig(config, this._reporter));
+    this._plugins.push(...webServerPluginsForConfig(config));
 
     await this._runAndReportError(async () => {
       // First run the plugins, if plugin is a web server we want it to run before the
       // config's global setup.
       for (const plugin of this._plugins) {
         await Promise.race([
-          plugin.setup?.(config, config._configDir, rootSuite),
+          plugin.setup?.(config, config._configDir, rootSuite, this._reporter),
```
dgozman, line 613: You should change the `TestRunnerPlugin` interface to include this new argument.

packages/playwright-test/src/docker/docker.ts
```diff
@@ -129,25 +131,32 @@ export async function buildPlaywrightImage() {
...
-    console.log(colors.dim('Stop manually using:'));
-    console.log(colors.dim('    npx playwright docker stop'));
-  }
-  console.log(colors.dim(`View screen: ${info.vncSession}`));
-  process.env.PW_TEST_CONNECT_WS_ENDPOINT = info.wsEndpoint;
-  process.env.PW_TEST_CONNECT_HEADERS = JSON.stringify({
-    'x-playwright-proxy': '*',
-  });
-  process.env.PLAYWRIGHT_DOCKER = '1';
+  addRunnerPlugin(() => ({
```
dgozman, line 134: Perhaps we should add the plugin unconditionally, and just check the env variable here? This way `cli.ts` could probably be simplified.

dgozman (approved):

PR #17317, opened 2022-09-14
fix(ignoreSnapshots): print a notice when ignoreSnapshots option is on

aslushnikov:
image


packages/playwright-test/src/runner.ts
```diff
@@ -603,6 +604,15 @@ export class Runner {
       }
     };
 
+    if (config._ignoreSnapshots) {
```
dgozman, line 607: Move this out of the global setup please.

dgozman (approved):

PR #17386, opened 2022-09-16
chore: replace __proto__ by getPrototypeOf

kapouer:
Some users might want to run `--disable-proto=throw|delete` option for a hardened Node.
It's easy to fix and doesn't prevent `{ __proto__: null }` pattern.

### DIFF
--- a/packages/playwright-core/src/protocol/serializers.ts
+++ b/packages/playwright-core/src/protocol/serializers.ts
@@ -184,5 +184,6 @@ function isURL(obj: any): obj is URL {
 }
 
 function isError(obj: any): obj is Error {
-  return obj instanceof Error || obj?.__proto__?.name === 'Error' || (obj?.__proto__ && isError(obj.__proto__));
+  const proto = obj ? Object.getPrototypeOf(obj) : null;
+  return obj instanceof Error || proto?.name === 'Error' || (proto && isError(proto));
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -71,6 +71,7 @@ module.exports = {
         "valid-typeof": 2,
         "no-implicit-globals": [2],
         "no-unused-expressions": [2, { "allowShortCircuit": true, "allowTernary": true, "allowTaggedTemplates": true}],
+        "no-proto": 2,
 
         // es2015 features
         "require-yield": 2,
--- a/packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
+++ b/packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts
@@ -48,7 +48,7 @@ export function source() {
 
   function isError(obj: any): obj is Error {
     try {
-      return obj instanceof Error || (obj && obj.__proto__ && obj.__proto__.name === 'Error');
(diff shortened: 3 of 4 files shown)


jfgreffier:
Should we add the `no-proto` ESLint rule in `.eslintrc.js` to enforce this ?

https://eslint.org/docs/latest/rules/no-proto

kapouer:
Let's see how it goes

mxschmitt:
The CI is failing: `page/page-expose-function.spec.ts:77:1 › should support throwing "null"`

kapouer:
Oh ! fixed and rebased.

mxschmitt (approved):

packages/playwright-core/src/protocol/serializers.ts
```diff
@@ -184,5 +184,6 @@ function isURL(obj: any): obj is URL {
 }
 
 function isError(obj: any): obj is Error {
-  return obj instanceof Error || obj?.__proto__?.name === 'Error' || (obj?.__proto__ && isError(obj.__proto__));
+  const proto = obj ? Object.getPrototypeOf(obj) : null;
```
yury-s, line 187: This will fail if there is a user code like `Object.getPrototypeOf = () => {};`

yury-s: Oh, this is a server side code, we don't run random js here so it should be ok.

yury-s (commented):
Could you add a test for this?

packages/playwright-core/src/protocol/serializers.ts
kapouer: Indeed, and `obj.__proto__` was an order of magnitude less safe than Object.getPrototypeOf :)

kapouer:
> Could you add a test for this?

What do you mean ? this is already tested. The behavior didn't change at all ?

kapouer:
For the record, the more modules are going to fix this, the more node is likely to get rid of __proto__ ! https://github.com/nodejs/node/issues/31951

yury-s:
> > Could you add a test for this?
> 
> What do you mean ? this is already tested. The behavior didn't change at all ?

I mean a test that would fail before your change and pass after, this way we can ensure that this is documented and will not break in the future during refactoring of the code.

jfgreffier:
> > > Could you add a test for this?
> > 
> > 
> > What do you mean ? this is already tested. The behavior didn't change at all ?
> 
> I mean a test that would fail before your change and pass after, this way we can ensure that this is documented and will not break in the future during refactoring of the code.

Not a test, but the `no-proto` ESLint rule that was added will guarantee that `__proto__` will not be used again

kapouer:
Maybe `process.env.NODE_OPTIONS = (process.env.NODE_OPTIONS || "") + " --disable-proto=throw"` should be added but honestly I'm not sure where.

kapouer:
> I'm fine with the change, but I don't get the idea behind the option above. We can't guarantee Playwright operation in the mode above, because we use third party modules and `__proto__` is a part of JavaScript. So we can land it, but it won't mean anything in the context of the above request.

`__proto__` is a *deprecated* part of js, and, while still used here and there, it is not so far away from being abandoned by modules authors.
Thanks !

PR #17400, opened 2022-09-16
feat(test runner): workers as percentage

### LINKED ISSUE #17204: [Feature] Allow percentage values for number of workers
As has been recently discussed on the Playwright Slack, running PW tests with the default number of workers (8 on a CPU with 8 physical cores and 16 logical cores) is not stable with a growing testsuite (in our case 37 test cases, another user reported the same with 50 test cases), test execution slows down considerably and leads to fails due to teardown timeouts.

Limiting the number of workers to 4 alleviates this, the tests run stable and don't timeout.
`workers: process.env.CI ? 1 : 4,`

However, it would be better to be able to set this as a percentage value, so it would be flexible and use the proper amount of workers depending on the machine's hardware.  
Currently the setting only accepts integer values.  

Also here's an article that was shared on the PW Slack with some interesting findings relating to this topic: https://ivantanev.com/make-jest-faster/

jfgreffier:
Allows to set workers as a percentage of logical CPUs, for example "50%".

Examples :
```bash
npx playwright test --workers 3
npx playwright test --workers 50%
```

```js
  // ...
  workers: '33%',
}
```

Fixes #17204

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -791,13 +791,13 @@ module.exports = config;
 
 ## property: TestConfig.workers
 * since: v1.10
-- type: ?<[int]>
+- type: ?<[int]|[string]>
 
-The maximum number of concurrent worker processes to use for parallelizing tests.
+The maximum number of concurrent worker processes to use for parallelizing tests. Can also be set as percentage of logical CPU cores, e.g. `'50%'.`
 
 Playwright Test uses worker processes to run tests. There is always at least one worker process, but more can be used to speed up test execution.
 
-Defaults to one half of the number of CPU cores. Learn more about [parallelism and sharding](../test-parallel.md) with Playwright Test.
+Defaults to half of the number of logical CPU cores. Learn more about [parallelism and sharding](../test-parallel.md) with Playwright Test.
 
 ```js tab=js-js
 // playwright.config.js
--- a/docs/src/test-configuration-js.md
+++ b/docs/src/test-configuration-js.md
@@ -516,7 +516,7 @@ In addition to configuring [Browser] or [BrowserContext], videos or screenshots,
 - `testMatch`: Glob patterns or regular expressions that match test files. For example, `'**/todo-tests/*.spec.ts'`. By default, Playwright Test runs `.*(test|spec)\.(js|ts|mjs)` files.
 - `timeout`: Time in milliseconds given to each test. Learn more about [various timeouts](./test-timeouts.md).
 - `webServer: { command: string, port?: number, url?: string, ignoreHTTPSErrors?: boolean, timeout?: number, reuseExistingServer?: boolean, cwd?: string, env?: object }` - Launch a process and wait that it's ready before the tests will start. See [launch web server](./test-advanced.md#launching-a-development-web-server-during-the-tests) configuration for examples.
-- `workers`: The maximum number of concurrent worker processes to use for parallelizing tests.
+- `workers`: The maximum number of concurrent worker processes to use for parallelizing tests. Can also be set as percentage of logical CPU cores, e.g. `'50%'.`
 
 You can specify these options in the configuration file. Note that testing options are **top-level**, do not put them into the `use` section.
 
--- a/packages/playwright-test/src/cli.ts
+++ b/packages/playwright-test/src/cli.ts
@@ -100,7 +100,7 @@ function addTestCommand(program: Command) {
   command.option('-gv, --grep-invert <grep>', `Only run tests that do not match this regular expression`);
   command.option('--global-timeout <timeout>', `Maximum time this test suite can run in milliseconds (default: unlimited)`);
   command.option('--ignore-snapshots', `Ignore screenshot and snapshot expectations`);
-  command.option('-j, --workers <workers>', `Number of concurrent workers, use 1 to run in a single worker (default: number of CPU cores / 2)`);
+  command.option('-j, --workers <workers>', `Number of concurrent workers or percentage of logical CPU cores, use 1 to run in a single worker (default: 50%)`);
   command.option('--list', `Collect all the tests and report them, but do not run`);
   command.option('--max-failures <N>', `Stop after the first N failures`);
   command.option('--output <dir>', `Folder for output artifacts (default: "test-results")`);
--- a/packages/playwright-test/src/cli.ts
+++ b/packages/playwright-test/src/cli.ts
@@ -272,7 +272,7 @@ function overridesFromOptions(options: { [key: string]: any }): ConfigCLIOverrid
     timeout: options.timeout ? parseInt(options.timeout, 10) : undefined,
     ignoreSnapshots: options.ignoreSnapshots ? !!options.ignoreSnapshots : undefined,
     updateSnapshots: options.updateSnapshots ? 'all' as const : undefined,
-    workers: options.workers ? parseInt(options.workers, 10) : undefined,
+    workers: options.workers,
   };
 }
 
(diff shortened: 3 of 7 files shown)


packages/playwright-test/src/cli.ts
```diff
@@ -100,7 +100,7 @@ function addTestCommand(program: Command) {
   command.option('-gv, --grep-invert <grep>', `Only run tests that do not match this regular expression`);
   command.option('--global-timeout <timeout>', `Maximum time this test suite can run in milliseconds (default: unlimited)`);
   command.option('--ignore-snapshots', `Ignore screenshot and snapshot expectations`);
-  command.option('-j, --workers <workers>', `Number of concurrent workers, use 1 to run in a single worker (default: number of CPU cores / 2)`);
+  command.option('-j, --workers <workers>', `Number of concurrent workers, use 1 to run in a single worker (default: number of logical CPU cores / 2). Can be percentage of logical CPU cores, for example "50%"`);
```
dgozman, line 103:
```suggestion
  command.option('-j, --workers <workers>', `Number of concurrent workers or percentage of logical CPU cores, use 1 to run in a single worker (default: 50%)`);
```

packages/playwright-test/src/loader.ts
```diff
@@ -143,7 +143,19 @@ export class Loader {
     this._fullConfig.shard = takeFirst(config.shard, baseFullConfig.shard);
     this._fullConfig._ignoreSnapshots = takeFirst(config.ignoreSnapshots, baseFullConfig._ignoreSnapshots);
     this._fullConfig.updateSnapshots = takeFirst(config.updateSnapshots, baseFullConfig.updateSnapshots);
-    this._fullConfig.workers = takeFirst(config.workers, baseFullConfig.workers);
+
+    const workers = takeFirst(config.workers, baseFullConfig.workers);
```
dgozman, line 147: Let's make `baseFullConfig.workers` be `50%`, remove cpu logic from there and only leave it here. Note that today it has some special considerations for M1 that we decided to remove.

packages/playwright-test/src/loader.ts
```diff
@@ -143,7 +143,19 @@ export class Loader {
...
+    if (typeof workers === 'string') {
+      if (workers.endsWith('%')) {
+        const cpus = os.cpus().length;
+        this._fullConfig.workers = Math.floor(cpus * (parseInt(workers, 10) / 100));
```
dgozman, line 151:
```suggestion
        this._fullConfig.workers = Math.min(1, Math.floor(cpus * (parseInt(workers, 10) / 100)));
```

utils/generate_types/overrides-test.d.ts
```diff
@@ -60,6 +60,7 @@ type LiteralUnion<T extends U, U = string> = T | (U & { zz_IGNORE_ME?: never });
 interface TestConfig {
   reporter?: LiteralUnion<'list'|'dot'|'line'|'github'|'json'|'junit'|'null'|'html', string> | ReporterDescription[];
   webServer?: TestConfigWebServer | TestConfigWebServer[];
+  workers?: number | string;
```
dgozman, line 63: I think you don't need to add this line, types generator should pick it up automatically.

packages/playwright-test/src/cli.ts
```diff
@@ -272,7 +272,7 @@ function overridesFromOptions(options: { [key: string]: any }): ConfigCLIOverrid
     timeout: options.timeout ? parseInt(options.timeout, 10) : undefined,
     ignoreSnapshots: options.ignoreSnapshots ? !!options.ignoreSnapshots : undefined,
     updateSnapshots: options.updateSnapshots ? 'all' as const : undefined,
-    workers: options.workers ? parseInt(options.workers, 10) : undefined,
+    workers: options.workers ? options.workers : undefined,
```
dgozman, line 275:
```suggestion
    workers: options.workers,
```

docs/src/test-configuration-js.md
```diff
@@ -516,7 +516,7 @@ In addition to configuring [Browser] or [BrowserContext], videos or screenshots,
 - `testMatch`: Glob patterns or regular expressions that match test files. For example, `'**/todo-tests/*.spec.ts'`. By default, Playwright Test runs `.*(test|spec)\.(js|ts|mjs)` files.
 - `timeout`: Time in milliseconds given to each test. Learn more about [various timeouts](./test-timeouts.md).
 - `webServer: { command: string, port?: number, url?: string, ignoreHTTPSErrors?: boolean, timeout?: number, reuseExistingServer?: boolean, cwd?: string, env?: object }` - Launch a process and wait that it's ready before the tests will start. See [launch web server](./test-advanced.md#launching-a-development-web-server-during-the-tests) configuration for examples.
-- `workers`: The maximum number of concurrent worker processes to use for parallelizing tests.
+- `workers`: The maximum number of concurrent worker processes to use for parallelizing tests. Can also be set as percentage of logical CPU cores. e.g. `'50%'`
```
dgozman, line 519:
```suggestion
- `workers`: The maximum number of concurrent worker processes to use for parallelizing tests. Can also be set as percentage of logical CPU cores, e.g. `'50%'.`
```

docs/src/test-api/class-testconfig.md
```diff
@@ -791,13 +791,13 @@ module.exports = config;
 
 ## property: TestConfig.workers
 * since: v1.10
-- type: ?<[int]>
+- type: ?<[int]|[string]>
 
-The maximum number of concurrent worker processes to use for parallelizing tests.
+The maximum number of concurrent worker processes to use for parallelizing tests. Can also be set as percentage of logical CPU cores. e.g. `'50%'`
```
dgozman, line 796:
```suggestion
The maximum number of concurrent worker processes to use for parallelizing tests. Can also be set as percentage of logical CPU cores, e.g. `'50%'.`
```

dgozman (commented):
Thank you for the PR. I left some comments, but overall looks great.

packages/playwright-test/src/loader.ts
jfgreffier: Done.

utils/generate_types/overrides-test.d.ts
jfgreffier: Done.

packages/playwright-test/src/loader.ts
jfgreffier: I set the `50%` on this line instead of `baseFullConfig`, to don't mess up with `FullConfigInternal.workers` typing for no reason. I believe it is not useful to have `FullConfigInternal.workers` as `number|string`  since it actually represent the number of workers internally.

What do you think ?

packages/playwright-test/src/loader.ts
```diff
@@ -143,7 +142,19 @@ export class Loader {
     this._fullConfig.shard = takeFirst(config.shard, baseFullConfig.shard);
     this._fullConfig._ignoreSnapshots = takeFirst(config.ignoreSnapshots, baseFullConfig._ignoreSnapshots);
     this._fullConfig.updateSnapshots = takeFirst(config.updateSnapshots, baseFullConfig.updateSnapshots);
-    this._fullConfig.workers = takeFirst(config.workers, baseFullConfig.workers);
+
+    const workers = takeFirst(config.workers, '50%');
```
jfgreffier, line 146: I set the `50%` on this line instead of `baseFullConfig`, to don't mess up with `FullConfigInternal.workers` typing for no reason. I believe it is not useful to have `FullConfigInternal.workers` as `number|string`  since it actually represent the number of workers internally.

What do you think ?

dgozman: Yep, this is good.

dgozman (approved):
Looks good, merging. Thank you for the PR!

PR #17413, opened 2022-09-16
fix(html-reporter): resolve output dir from given path

### LINKED ISSUE #17412: [BUG] HTML reporter otuput dir always resolves from the playwright config location
**Context:**
- Playwright Version: 1.25.2
- Operating System: Mac
- Node.js version: 16.16
- Browser: All

**Bug Description**

Given the following report configuration

```javascript

   //...
reporter: [
    [
      'html',
      {
        outputFolder: '../../dist/apps/playwright-project/playwright-report',
      },
    ],
    [
      'json',
      {
        outputFile:
          'dist/apps/playwright-project/playwright-report/test-results.json',
      },
    ],
  ],
// ....
}
```

and running `npx playwright --config apps/playwright-project/playwright.config.ts` one would expect to get an output like this

```
.
├── apps
│   └── playwright-project
│       ├── playwright.config.ts
│       └── src
│           └── e2e
│               └── my-test.spec.ts
└── dist
    └── apps
        └── playwright-project
            └── playwright-results
                ├── index.html
                └── test-results.json
```

but the html report location is always resolving relative to the playwright.config.ts file location, so the current result is:

```
.
├── apps
│   └── playwright-project
│       ├── dist
│       │   └── apps
│       │       └── playwright-project
│       │           └── playwright-report
│       │               └── index.html
│       ├── playwright.config.ts
│       └── src
│           └── e2e
│               └── my-test.spec.ts
└── dist
    └── apps
        └── playwright-project
            └── playwright-results
                └── test-results.json
```

The HTML reporter should resolve the output dir according to what it received in the `outputFolder` config, just like the JSON reporter does.

sebastiandg7:
Resolve the HTML reporter outputFolder location according to the received outputFolder path.

Closes #17412

### DIFF
--- a/packages/playwright-test/src/reporters/html.ts
+++ b/packages/playwright-test/src/reporters/html.ts
@@ -173,8 +173,6 @@ class HtmlReporter implements ReporterInternal {
 
   _resolveOptions(): { outputFolder: string, open: HtmlReportOpenOption } {
     let { outputFolder } = this._options;
-    if (outputFolder)
-      outputFolder = path.resolve(this.config._configDir, outputFolder);
     return {
       outputFolder: reportFolderFromEnv() ?? outputFolder ?? defaultReportFolder(this.config._configDir),
       open: process.env.PW_TEST_HTML_REPORT_OPEN as any || this._options.open || 'on-failure',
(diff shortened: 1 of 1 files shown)


sebastiandg7:
@rwoll will do! I got a bit lost in the `tests` directory. There is a spec file for the HTML reporter under `tests/playwright-test` but only spec files under `tests/library` and `tests/page` are being tested. Should I create a new spec file for the html reporter inside `tests/library`? 

PD. I think the contributing guide can get an update to make this clearer.

sebastiandg7:
What about adding a config option to the html reporter like {..., relativeToConfig: false (true by default), ....}, with the proper documentation so that users can opt-in to have that path resolution consistent with the JSON reporter.

PR #17424, opened 2022-09-18
fix(docker): do not sync primary selection into clipboard

pavelfeldman:


packages/playwright-test/src/docker/build_docker_image.sh
```diff
@@ -17,6 +17,54 @@ mkdir -p /opt/bin && chmod +x /dev/shm \
     && rm -rf websockify-${WEBSOCKIFY_REF}/{docs,tests} \
     && mv websockify-${WEBSOCKIFY_REF} /opt/bin/noVNC/utils/websockify
 
+# Patch noVNC
+
+cat <<'EOF' > /opt/bin/noVNC/clip.patch
```
mxschmitt, line 22: make a PR upstream so we can get rid of the patch in the future? https://github.com/novnc/noVNC/blob/master/app/ui.js

aslushnikov (approved):

PR #17426, opened 2022-09-18
Documentation about multiple root folders and monorepos

### LINKED ISSUE #17240: [FEATURE][VS CODE EXTENSION] Use all projects for starting the tree of tests
**Context:**
- Playwright Version: 1.25.2
- Operating System: Mac
- Node.js version: 16
- Visual Studio Code version: 1.66
- Playwright for VSCode extension version: 0.2.10 Preview
- Browser: All
- Extra: monorepo setup managed by TurboRepo

I imagine that there are many things planned for the Playwright extension for VS Code. But here is one suggestion.

I use a monorepo setup managed by TurboRepo. Ideally, to me, I would have Playwright working independently on each one of my packages, and the extension would detect the many files named `playwright.config.ts` at the root of each package.

Unfortunately, that does not seem to work with the VS extension.

Because of this, I have many projects listed in one single config file at the root of my monorepo. Each project uses a `testMatch` property to funnel the tests files of a single project, as such:
```
// Please nevermind the strings. I know they are supposed to be regexp.
projects: [
  {
    name: 'all'
    testMatch: '**/*.test.ts'
  },
  {
    name: 'client'
    testMatch: 'apps/client/**/*.test.ts'
  },
  {
    name: 'server'
    testMatch: 'apps/server/**/*.test.ts'
  },
  {
    name: 'package1'
    testMatch: 'packages/package-1/**/*.test.ts'
  },
  ...
]
```

But did you notice the first project named `all`? I do that because that is the only way to have all tests listed by the extension.

What I would like to do is to remove that project named `all` and have the extension listing all projects as starting points for their own tree of tests. That way, the extension would look like this (without the project named `all`):

```
> client
> server
  src/
    a.test.ts
    b.test.ts
> package1
```

I think this would be nice because each project has its own configuration, like `outputDir` and `snapshotDir`.

Please advise if you think this is a good approach or if I should have one config file for each project.

phcoliveira:
docs(vs_code_extension) Documentation about multiple root folders and monorepos

This PR introduces some explanation about how the VS Code extension can be used with multiple root folders added to a workspace. It also suggests how the extension can work with a monorepo.

Closes #17240

### DIFF
--- a/docs/src/getting-started-vscode-js.md
+++ b/docs/src/getting-started-vscode-js.md
@@ -71,11 +71,19 @@ However, a single Playwright configuration file can still be used to test many p
 
 For example, suppose that a given root folder has a root level package, in which Playwright was installed, and three others packages in nested folders: "apps/client", "apps/server" and "libs/common".
 
+<img width="355" height="458" alt="A file explorer with a root level package and three nested packages: 'apps/client', 'apps/server' and 'libs/common'." src="https://user-images.githubusercontent.com/594605/190917713-c51da6a0-7bde-49fd-a78a-0274b712367a.png" />
+
 The Playwright configuration file at the root level package can define projects for each package by restricting the test files belonging to any given package.
+
+<img width="432" height="754" alt="The property 'projects' of the Playwright configuration file describing three projects due to the package 'apps/client', one project due to the package 'apps/server' and one project due to the package 'libs/common'." src="https://user-images.githubusercontent.com/594605/190918095-035aa6fb-3efe-46ac-8ee8-4269ffe8dd7d.png" />
 
 For more information about which options can be used when defining a project, please see the [options available for a project](./test-api/class-testproject.md).
 
-The dropdown menu **Select Configuration** will account for all projects.
+The testing tree and the dropdown menu **Select Configuration** will account for all projects.
+
+<img width="355" height="373" alt="The testing tree showing all tests belonging to all five projects." src="https://user-images.githubusercontent.com/594605/190918243-063936ea-e070-4297-8f40-776d9bd339ce.png" />
+
+<img width="504" height="199" alt="The dropdown menu 'Select Configuration' showing all five projects." src="https://user-images.githubusercontent.com/594605/190918293-1b1e5f39-9178-4906-897d-a256ca48c147.png" />
 
 ### Run Tests on Specific Browsers
 
--- a/docs/src/getting-started-vscode-js.md
+++ b/docs/src/getting-started-vscode-js.md
@@ -105,7 +105,7 @@ Pick a selector and copy it into your test file by clicking the **Pick selector*
 
 ## What's next
 
-- [See how to work with the VS Code Extension in different scenarios](./vscode-extension-js.md)
+- [See how to work with the VS Code Extension in different scenarios](./vscode-extension.md)
 - [Write tests using web first assertions, page fixtures and locators](./writing-tests.md)
 - [See test reports](./running-tests.md#test-reports)
 - [See a trace of your tests](./trace-viewer-intro.md)
--- a/docs/src/getting-started-vscode-js.md
+++ b/docs/src/getting-started-vscode-js.md
@@ -65,9 +65,9 @@ And the dropdown menu **Select Configuration** will account for all the projects
 
 ### Run Tests from Multiple Packages - Monorepos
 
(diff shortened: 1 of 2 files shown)


phcoliveira:
image

phcoliveira:
image

phcoliveira:
image

phcoliveira:
image

phcoliveira:
image

phcoliveira:
image

phcoliveira:
image

phcoliveira:
image

phcoliveira:
image

debs-obrien (changes requested):
Hi, Thanks for the great contribution. 

I would love to see this in a new doc on VS Code that covers more complex things rather than in getting started which is aimed at people trying out Playwright for the first time. This doc hasn't be created yet so feel free to create it or I can do it and transfer your content to it.
Then the getting started guide should link to this section.

For the examples I would like to refrain from using bar and foo and use more real world examples and explanations. Again happy to use what you created and work on it if you don't have the time to make the changes.

debs-obrien:
Hi, no worries @phcoliveira 

I can sort it out tomorrow. 

But to answer your question it would have the js extension because it is only relevant for VS code extension which only works in node. Adding the js extension means that doc will only get shown in the node docs. All docs are created here and then rolled to playwright. dev so the only manual thing we need to do in that repo is add the page to the sidebar so it can be discovered once the pr has been merged.

phcoliveira:
> Hi, no worries @phcoliveira
> 
> I can sort it out tomorrow.
> 
> But to answer your question it would have the js extension because it is only relevant for VS code extension which only works in node. Adding the js extension means that doc will only get shown in the node docs. All docs are created here and then rolled to playwright. dev so the only manual thing we need to do in that repo is add the page to the sidebar so it can be discovered once the pr has been merged.

Thank you for your readiness, @debs-obrien. Then I will proceed with a new doc named `VS Code extension` with the suffix `-js`. If you have a suggestiong about title of such doc, please let me know.

debs-obrien:
You can give it the title of VS Code Extension.

phcoliveira:
image

phcoliveira:
image

phcoliveira:
image

debs-obrien:
Looks great. I will later add to this doc to add more VS Code extension things that are not covered in getting started but this is a great start and much appreciated. Once merged I will take care of adding to the sidebar.

@mxschmitt will give you it the once over to make sure we haven't missed anything.

debs-obrien (approved):

phcoliveira:
Just a friendly reminder, @mxschmitt.

debs-obrien:
No need for reminder @phcoliveira. Max and I are discussing what else needs to go into the doc before it goes live. If we need to rename some sections, add others etc so consider it still a work in progress. We don’t want to make something live and then remove it so we want to make sure it’s right, it’s solid, and it’s there to stay. These decisions always take time. Thanks for your understanding 🙏

phcoliveira:
> I was testing the mentioned edge-cases inside your guide and it was working for me as expected which would make this page unnecessary. I tested it with TypeScript and JavaScript. In both cases it recognized multiple Playwright configurations inside sub folders correctly.
> 
> Can you share your exact file structure maybe as a git repo?

I agree that this page would be unnecessary, perhaps even prejudicial, if Playwright configurations were detected in nested folders.

I will try that right away. I think I will be able to answer you today or tomorrow morning.
@mxschmitt

phcoliveira:
> I was testing the mentioned edge-cases inside your guide and it was working for me as expected which would make this page unnecessary. I tested it with TypeScript and JavaScript. In both cases it recognized multiple Playwright configurations inside sub folders correctly.
> 
> Can you share your exact file structure maybe as a git repo?

@mxschmitt, you are right. I changed my project from one root installation to multiple nested ones, and the VS Code Extension is working.

As of now, I can't say why it didn't work on my first attempt. I started with multiple installations, but I didn't find any tests in the extension. I read somewhere (will try to find it again) that the extension would look for Playwright as a dependency inside a `package.json` file and a configuration file.

Therefore, I think the instructions I wrote would be detrimental because I do believe that having a Playwright configuration per package is a better approach than having a single one at the root with multiple projects.

On a unrelated (or tangent) matter, I do think that project "Profile" from the extension is confusing and could use a single selection instead of multiple ones. And also that the testing tree should only show the tests matched by that single project.

I believe Playwright can be much more than it was intended to do: an e2e testing suite. I am using it for testing Firebase's rules and functions in one package, and React components in another. So it is quite strange to see the server's tests (Firebase) from one package when I have only selected the client's tests performed by Firefox, from another package. So, to summarise, I replaced Jest, JsDom and Testing Library React with Playwright in all packages.

@debs-obrien

debs-obrien:
closing as not needed anymore with new explanations in vs getting started guide

Issue #17448, opened 2022-09-12
[BUG] Playwright throws exception "Browser closed" when testing against MS Edge


arc95:
**Context:**
- Playwright Version: 1.25.0
- Operating System: Windows 10 x64
- .NET version: .NET 6
- Browser: MS Edge
- Extra: Visual Studio Enterprise 17.3.2

**Code Snippet**
```cs
[TestMethod]
public async Task LoadPlaywrightPage()
{
    var playwright = await Playwright.CreateAsync();
    var browser = await playwright.Chromium.LaunchAsync(new BrowserTypeLaunchOptions
    {
        Headless = false,
        Channel = "msedge"
    });
    var context = await browser.NewContextAsync();
    var page = await context.NewPageAsync();
    await page.GotoAsync("https://playwright.dev/dotnet/");

    Assert.IsTrue((await page.TitleAsync()).Contains("Playwright .NET"));
}
```

**Describe the bug**

Running Visual Studio Enterprise in non-admin, the above code throws this exception:

```
Test method PlaywrightTests.PlaywrightTests.LoadPlaywrightPage threw exception: 
   Microsoft.Playwright.PlaywrightException: Browser closed. 
==================== Browser output: ==================== 
C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe --disable-field-trial-config --disable-background-networking 
--enable-features=NetworkService,NetworkServiceInProcess --disable-background-timer-throttling 
--disable-backgrounding-occluded-windows --disable-back-forward-cache --disable-breakpad 
--disable-client-side-phishing-detection --disable-component-extensions-with-background-pages 
--disable-default-apps --disable-dev-shm-usage --disable-extensions 
--disable-features=ImprovedCookieControls,LazyFrameLoading,GlobalMediaControls,DestroyProfileOnBrowserClose,
MediaRouter,DialMediaRouteProvider,AcceptCHFrame,AutoExpandDetailsElement,
CertificateTransparencyComponentUpdater,AvoidUnnecessaryBeforeUnloadCheckSync,Translate 
--allow-pre-commit-input --disable-hang-monitor --disable-ipc-flooding-protection --disable-popup-blocking 
--disable-prompt-on-repost --disable-renderer-backgrounding --disable-sync --force-color-profile=srgb 
--metrics-recording-only --no-first-run --enable-automation --password-store=basic --use-mock-keychain 
--no-service-autorun --export-tagged-pdf --no-sandbox 
[...]

Issue #17299, opened 2022-06-27
[BUG] Chrome Policy ExtensionInstallForcelist prevents Browser from closing


gemiusz:
**Context:**
- Playwright Version: 1.22 (JAVA)
- Operating System: Windows 10 Pro
- Browser: Chrome (103.0.5060.53)
- Extra: Browser managed by the organization.

**Code Snippet**

```java
import com.microsoft.playwright.*;
import org.testng.annotations.Test;

public class PlayWrightTestNGExample {

    @Test
    public void testName() {
        Playwright playwright = Playwright.create();
        Browser browser;
        BrowserContext context;
        Page page;
        browser = playwright.chromium().launch(new BrowserType.LaunchOptions().setHeadless(false).setChannel("chrome"));
        context = browser.newContext();
        page = context.newPage();

        page.navigate("chrome://chrome-urls/");

        page.close();
        context.close();
        browser.close();
    }
}

```

**Describe the bug**

It took me a week of looking how to reproduce Bug mentioned and closed here [https://github.com/microsoft/playwright-java/issues/347](https://github.com/microsoft/playwright-java/issues/347)

I announce that: **I have reproduced it xD**

Below long story short:
I asked about this issue on Polish Facebook Playright group and one person have the same issue but in Playwright JS version. 
In my work environment Chrome is managed by organisation and have appiled some polices - the same have person from microsoft/playwright-java#347 and facebook group so this was place to search.
After some analyses and tirals with Workplace Team who managed polices we found that Extensions making problem - when all disabled everything working fine.
Today I added Chrome polices to my private laptop and found that when Policy [ExtensionInstallForcelist](https://chromeenterprise.google/policies/?policy=ExtensionInstallForcelist) is enabled and added for example [Webex Extension](https://chrome.google.com/webstore/detail/cisco-webex-extension/jlhmfgmfgeifomenelglieieghnjghma) Playwright driver have mentioned above problems with "Gracefully Close".

Logs when Policy is enabled:
![image](https://user-images.githubusercontent.com/30750006/176017274-43f731fd-de54-4d50-808e-ad9a5ae69134.png)

Policy settings:
![image](https://user-images.githubusercontent.com/30750006/176017425-e6b0c37c-971f-4826-9483-6efb3f9470d0.png)

I think the find above should help in preparing the fix.

dgozman:
Closing this one, because Playwright does not work with any Chrome Policies.

gemiusz:
[...]

PR #17449, opened 2022-09-19
docs: add note about unsupported group policies

mxschmitt:
https://github.com/microsoft/playwright/issues/17299
https://github.com/microsoft/playwright/issues/17448

### DIFF
--- a/docs/src/troubleshooting.md
+++ b/docs/src/troubleshooting.md
@@ -40,10 +40,10 @@ Running TypeScript tests in `"type": "module"` project requires Node.js 16 or hi
 
 ## Incompatible Microsoft Edge / Google Chrome policy settings
 
-It's known that Playwright is not working as expected when some Microsoft Edge / Google Chrome policy settings are set. The following shows some of them, there might be more that affect Playwright's functionalities. If you find any other policy settings that break functionality, please file an issue. You can see a list of applied policies by running `chrome://policy` in the browser.
+It's known that Playwright is not working as expected when some Microsoft Edge / Google Chrome policy settings are set. The following shows some of them, there might be more that affect Playwright's functionalities. If you find any other policy settings that break functionality, please file an issue and we'll add it to this document. You can see a list of applied policies by running `chrome://policy` in the browser.
 
-- `UserDataDir` - This policy is used to specify the location of the user data directory. Playwright uses a temporary directory for user data, so this policy is not compatible with Playwright see [here](https://github.com/microsoft/playwright/issues/17448).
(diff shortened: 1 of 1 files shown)


docs/src/troubleshooting.md
```diff
@@ -38,6 +38,13 @@ Playwright requires Node.js 14 or higher.
 
 Running TypeScript tests in `"type": "module"` project requires Node.js 16 or higher.
 
+## Incompatible Microsoft Edge / Google Chrome policy settings
+
+It's known that Playwright is not working as expected when some Microsoft Edge / Google Chrome policy settings are set. The following shows some of them, there might be more that affect Playwright's functionalities. If you find any other policy settings that break functionality, please file an issue. You can see a list of applied policies by running `chrome://policy` in the browser.
```
yury-s, line 43:
```suggestion
It's known that Playwright is not working as expected when some Microsoft Edge / Google Chrome policy settings are set. The following shows some of them, there might be more that affect Playwright's functionalities. If you find any other policy settings that break functionality, please file an issue and we'll add it to this document. You can see a list of applied policies by running `chrome://policy` in the browser.
```

docs/src/troubleshooting.md
```diff
@@ -38,6 +38,13 @@ Playwright requires Node.js 14 or higher.
...
+
+- `UserDataDir` - This policy is used to specify the location of the user data directory. Playwright uses a temporary directory for user data, so this policy is not compatible with Playwright see [here](https://github.com/microsoft/playwright/issues/17448).
```
yury-s, line 45:
```suggestion
- `UserDataDir` - This policy is used to specify the location of the user data directory. Playwright uses a temporary directory for user data, so this policy is not compatible with Playwright. See discussion in [this bug](https://github.com/microsoft/playwright/issues/17448).
```

docs/src/troubleshooting.md
```diff
@@ -38,6 +38,13 @@ Playwright requires Node.js 14 or higher.
...
+- `ExtensionInstallForcelist` - This policy is used to specify a list of extensions that should be installed. Playwright's browser close will not work if this policy is set see [here](https://github.com/microsoft/playwright/issues/17299).
```
yury-s, line 46:
```suggestion
- `ExtensionInstallForcelist` - This policy is used to specify a list of extensions that should be installed. Playwright's browser close will not work if this policy is set. See discussion in [this bug](https://github.com/microsoft/playwright/issues/17299).
```

yury-s (approved):

PR #17450, opened 2022-09-20
chore: migrate codegen to controller events

pavelfeldman:


packages/recorder/src/callLog.tsx
```diff
@@ -54,7 +54,7 @@ export const CallLogView: React.FC<CallLogProps> = ({
             { message.trim() }
           </div>;
         })}
-        { !!callLog.error && <div className='call-log-message error' hidden={!isExpanded}>{ callLog.error.error?.message }</div> }
+        { !!callLog.error && <div className='call-log-message error' hidden={!isExpanded}>{ callLog.error }</div> }
```
dgozman, line 57: Stray change?

packages/playwright-core/src/server/reuseController.ts
```diff
@@ -61,9 +62,9 @@ export class ReuseController extends SdkObject {
         onPageNavigated: () => this._emitSnapshot(),
         onPageClose: () => this._emitSnapshot(),
       };
-      this.instrumentation.addListener(this._trackHierarchyListener, null);
+      this._playwright.instrumentation.addListener(this._trackHierarchyListener, null);
```
dgozman, line 65: Ouch!

dgozman (approved):

packages/recorder/src/callLog.tsx
pavelfeldman: It is not stray, I needed to simplify the interface to remove channel dependencies.

packages/playwright-core/src/server/reuseController.ts
pavelfeldman: Ouch indeed!

PR #17475, opened 2022-09-20
fix(ct): support empty fragments

### LINKED ISSUE #17231: [BUG] MountResult functions hang indefinitely when called on an empty React.Fragment, but not an empty div
**Context:**
- Playwright Version: 1.25.1
- Operating System: Linux 5.17 Pop!_OS 20.04 LTS
- Node.js version: v16.14.2 LTS
- Browser: All
- Extra: N/A

**Code Snippet**

When testing a list component in a test using the `mount` function from the `@playwright/experimental-ct-react` `test` function, I ran into this problem.

I have not yet stubbed the `listElements` variable in my test, so the `MyListContainer` component renders an empty JSX Fragment. 

This code can be tested successfully:

```javascript
const MyListContainer = () => {
  return (
    <div>
      {listElements.map(() => <SomeListElement />)}
    </div>
  )
}
```

But with this code:

```javascript
const MyListContainer= () => {
  return (
    <>
      {listElements.map(() => <SomeListElement />)}
    </>
  )
}
```

The `textContent`, `innerText`, and `innerHTML` functions will hang a test indefinitely after called on the mounted component. The `allInnerTexts` and `allTextContents` functions understand the JSX and correctly return an empty array.

Test code:

```javascript
import React from 'react'
import { test, expect } from '@playwright/experimental-ct-react'

import MyListContainer from '.'

test('component mounts', async ({ mount }) => {
  const component = await mount(<MyListContainer />)

  console.log('allInnerTexts():', await component.allInnerTexts()) // Correctly prints: allInnerTexts(): []
  console.log('allTextContents():', await component.allTextContents()) // Correctly prints: allTextContents(): []
  console.log('textContent():', await component.textContent()) // Will hang the test indefinitely, nothing is logged
  console.log('innerText():', await component.innerText()) // Will hang the test indefinitely if called first
  console.log('innerHTML():', await component.innerHTML()) // Will hang the test indefinitely if called first

  await expect(component).toBeTruthy()
})

```

**Describe the bug**

Described above.

dgozman:
Currently, we ues `#root` vs `#root > *` selector for component roots depending on the number of root children. This heuristic detects fragments that render multiple elements inside the root.

However, this does not work with empty fragments that do not render anything.

The fix is to make the `#root >> control=component` selector that would dynamically detect the root. This supports empty fragments and also allows for dynamic updates of the fragments.

Fixes #17231.

### DIFF
--- a/tests/components/ct-react-vite/src/components/EmptyFragment.tsx
+++ b/tests/components/ct-react-vite/src/components/EmptyFragment.tsx
@@ -1,0 +1,3 @@
+export default function EmptyFragment() {
+  return <>{[]}</>;
+}
--- a/tests/components/ct-react/src/components/EmptyFragment.tsx
+++ b/tests/components/ct-react/src/components/EmptyFragment.tsx
@@ -1,0 +1,3 @@
+export default function EmptyFragment() {
+  return <>{[]}</>;
+}
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -278,6 +278,13 @@ export class InjectedScript {
           return [];
         if (body === 'return-empty')
           return [];
+        if (body === 'component') {
+          if (root.nodeType !== 1 /* Node.ELEMENT_NODE */)
+            return [];
+          // Usually, we return the mounted component that is a single child.
+          // However, when mounting fragments, return the root instead.
+          return [root.childElementCount === 1 ? root.firstElementChild! : root as Element];
+        }
         throw new Error(`Internal error, unknown control selector ${body}`);
       }
     };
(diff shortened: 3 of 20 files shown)


tests/components/ct-react-vite/src/components/EmptyFragment.tsx
```diff
@@ -0,0 +1,3 @@
+export default function EmptyFragment() {
+  return <>{[]}</>;
```
mxschmitt, line 2: Note that this is not only about fragments from my understanding. In theory it's the same for:

- null
- undefined
- Fragments

Since just nothing gets rendered to the root element.

tests/components/ct-react/src/components/EmptyFragment.tsx
```diff
@@ -0,0 +1,3 @@
+export default function EmptyFragment() {
+  return <>{[]}</>;
```
mxschmitt, line 2:
```suggestion
  return <></>;
```

Why not that?

pavelfeldman (approved):

Issue #16517, opened 2022-08-12
[BUG] toHaveAttribute always passes when given a blank value string


radium-v:
**Context:**
- Playwright Version: 1.25
- Operating System: Windows 11, Ubuntu 22.04 via WSL2
- Node.js version: 16.16.0
- Browser: All

**Code Snippet**

```javascript
import { expect, test } from "@playwright/test";

test("should return false for nonexistent attributes", async ({ page }) => {
    await page.setContent(`<div></div>`);

    const element = page.locator("div");

    // always passes
    await expect(element).toHaveAttribute("", "");

    // always passes
    await expect(element).toHaveAttribute("any-arbitrary-attribute", "");

    // always fails, `value` must be a string
    await expect(element).not.toHaveAttribute("any-attribute", "");

    // always fails (Matcher error: expected value must be a string or regular expression)
    await expect(element).not.toHaveAttribute("attr");

    // workaround: passes when the attribute is not present
    expect(await element.getAttribute("attr")).toBeNull();

    await element.evaluate(node => node.toggleAttribute("attr"));

    // workaround: this solution checks for the existence of boolean attributes with no value, but doesn't cover every valid boolean attribute situation
    expect(await element.getAttribute("attr")).toBe("");
});
```

**Describe the bug**

The `toHaveAttribute` matcher doesn't work as expected with the checks above, since `await expect(element).toHaveAttribute("anything", "")` will always pass. Likewise, `await expect(element).not.toHaveAttribute("anything", "")` will always fail for any attribute when used as a check for the presence of attributes. In our case, this primarily affects the use of boolean attributes like `checked`, `disabled`, `hidden`, `open`, which are considered `true` when they're present, regardless of value.

I think the `value` parameter for `toHaveAttribute` could be made optional, which would allow checking for the existence of attributes. The always-failing `not.toHaveAttribute` matcher would then be able to pass. This may also help to address #16270, though boolean attributes likely would need their own logic to match the allowed validity rules described in [this MDN article](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes).

aslushnikov:
> I think the value parameter for toHaveAttribute could be made optional, which would allow checking for the existence of attributes

@radium-v Sounds legit to me, let me bring it up on the team meeting.

aslushnikov:
@radium-v workaround: could you please try with empty slashes? 🤷‍♂️

```ts
[...]

PR #17477, opened 2022-09-20
fix(expect): toHaveAttribute with empty value should not match missing attribute

yury-s:
Reference #16517

### DIFF
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -1101,7 +1101,7 @@ export class InjectedScript {
       if (expression === 'to.have.attribute') {
         const value = element.getAttribute(options.expressionArg);
         if (value === null)
-          return { received, matches: false };
+          return { received: null, matches: false };
         received = value;
       } else if (expression === 'to.have.class') {
         received = element.classList.toString();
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1099,7 +1099,10 @@ export class InjectedScript {
       // Single text value.
       let received: string | undefined;
       if (expression === 'to.have.attribute') {
-        received = element.getAttribute(options.expressionArg) || '';
+        const value = element.getAttribute(options.expressionArg);
```
pavelfeldman, line 1102: I would drop '' and leave the rest as is. That way users get 'got null' in the error message

pavelfeldman (approved):

rwoll (approved):

packages/playwright-core/src/server/injected/injectedScript.ts
yury-s: Received cannot be null, so we'd have to covert it to undefined in that case which would result in 'Unknown expect matcher: to.have.attribute' which is not what we want to see. Additionally it should return true for `.not.toHaveAttribute('foo', '');` rather than throw when the attribute is missing.

pavelfeldman: > Received cannot be null

Why?

yury-s: Updated

PR #17486, opened 2022-09-20
chore: move protocol and trace types into the top-level packages

pavelfeldman:


utils/generate_types/test/tsconfig.json
```diff
@@ -3,7 +3,14 @@
     "strict": true,
     "target": "ES2019",
     "noEmit": true,
-    "moduleResolution": "node"
+    "moduleResolution": "node",
+    "baseUrl": "../../..",
+    "paths": {
```
dgozman, line 8: Hmm... It would be better for this test to have a setup like the end user would.

dgozman (approved):

PR #17518, opened 2022-09-21
fix(reporters): output relative to config

### LINKED ISSUE #17412: [BUG] HTML reporter otuput dir always resolves from the playwright config location

rwoll:
Fixes #17412.
Supercedes #17413.

- if configured via playwright.config.ts, relative paths should be relative to the config.
- if configured via env var, should be relative to `cwd`

### DIFF
--- a/packages/playwright-test/src/reporters/json.ts
+++ b/packages/playwright-test/src/reporters/json.ts
@@ -19,13 +19,14 @@ import path from 'path';
 import type { FullConfig, TestCase, Suite, TestResult, TestError, TestStep, FullResult, Location, Reporter, JSONReport, JSONReportSuite, JSONReportSpec, JSONReportTest, JSONReportTestResult, JSONReportTestStep } from '../../types/testReporter';
 import { prepareErrorStack } from './base';
 import { MultiMap } from 'playwright-core/lib/utils/multimap';
+import type { FullConfigInternal } from '../types';
 
 export function toPosixPath(aPath: string): string {
   return aPath.split(path.sep).join(path.posix.sep);
 }
 
 class JSONReporter implements Reporter {
-  config!: FullConfig;
+  config!: FullConfigInternal;
   suite!: Suite;
   private _errors: TestError[] = [];
   private _outputFile: string | undefined;
@@ -38,7 +39,7 @@ class JSONReporter implements Reporter {
     return !this._outputFile;
   }
 
-  onBegin(config: FullConfig, suite: Suite) {
+  onBegin(config: FullConfigInternal, suite: Suite) {
     this.config = config;
     this.suite = suite;
   }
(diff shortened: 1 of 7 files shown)


packages/playwright-test/src/reporters/json.ts
```diff
@@ -19,13 +19,14 @@ import path from 'path';
...
 import { MultiMap } from 'playwright-core/lib/utils/multimap';
+import type { FullConfigInternal } from '../types';
 
 export function toPosixPath(aPath: string): string {
   return aPath.split(path.sep).join(path.posix.sep);
 }
 
 class JSONReporter implements Reporter {
-  config!: FullConfig;
+  config!: FullConfigInternal;
```
pavelfeldman, line 29: This suggests that reporters need internal APIs, which does not sound right.

pavelfeldman: If config is used, you have configFile. If config is not used, I don't think you can even configure these reporters.

dgozman (approved):

packages/playwright-test/src/reporters/json.ts
```diff
@@ -19,6 +19,7 @@ import path from 'path';
 import type { FullConfig, TestCase, Suite, TestResult, TestError, TestStep, FullResult, Location, Reporter, JSONReport, JSONReportSuite, JSONReportSpec, JSONReportTest, JSONReportTestResult, JSONReportTestStep } from '../../types/testReporter';
 import { prepareErrorStack } from './base';
 import { MultiMap } from 'playwright-core/lib/utils/multimap';
+import assert from 'assert';
```
mxschmitt, line 22: this assert seems wrong, its not our custom assert.

mxschmitt (changes requested):

PR #17524, opened 2022-09-22
docs: update POM test example

depapp:
hi all,
since `Page Object Models` docs is no longer under `Playwright Test`. there's failed test example in https://playwright.dev/docs/test-pom related to wrong selector.

- after this PR:
playwright - pom - after

- before this PR:
playwright - pom - before


debs-obrien:
Great, thanks :)

PR #17537, opened 2022-09-22
[Feature] Implement launchServer and connect for Android

pranavj1001:
Detailed description can be found here: https://github.com/microsoft/playwright/issues/17538


pavelfeldman:
This change requires extensive review and is not high in the priority list for Playwright. I would recommend filing it as a feature request and tracking its progress. It is likely to be triaged as P3-collecting the feedback and will sit there for ~6 months while we are getting upvotes.

pranavj1001:
Hi @pavelfeldman, yes, we have filed a [Feature Request](https://github.com/microsoft/playwright/issues/17538) for this and it has received some upvotes.

Issue #17543, opened 2022-09-22
[Docs] Clarify the default number of workers based on CPU cores


azad-derakhshani-GS:
The current version of the [docs](https://playwright.dev/docs/api/class-testconfig#test-config-workers) about the default number of workers can be a bit confusing for users who are not well-versed in the domain of CPUs.

> Defaults to half of the number of logical CPU cores.

More specifically, some users don't know the difference between physical and logical CPU cores (or that there is a difference at all).
They might misinterpret this as simply "half of the number of CPU cores" (see example [here](https://playwright.slack.com/archives/CSUHZPVLM/p1663873162598759?thread_ts=1663860454.381469&cid=CSUHZPVLM)).

So I would like to update the documentation to further clarify this and avoid any confusion.

PR #17544, opened 2022-09-22
docs(class-testconfig.md): further clarify the number of workers Playwright defaults to

azad-derakhshani-GS:
This further clarifies the docs on how many workers Playwright defaults to depending on the CPU, as mentioned in #17543.

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -797,7 +797,10 @@ The maximum number of concurrent worker processes to use for parallelizing tests
 
 Playwright Test uses worker processes to run tests. There is always at least one worker process, but more can be used to speed up test execution.
 
-Defaults to half of the number of logical CPU cores. Learn more about [parallelism and sharding](../test-parallel.md) with Playwright Test.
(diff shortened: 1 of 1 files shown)


azad-derakhshani-GS:
I'm not sure why the docs & lint check failed, could anyone help with that?

rwoll (commented):
Thanks for the PR! However, this information is wrong (logical cores != physical cores * 2), and I'm left as a user with more questions than answers once seeing discussions of logical vs. physical.

We'll opt to leave this as is for now. Thanks!

azad-derakhshani-GS:
Interesting, that's news to me. Can you show me an example where this would not apply?
As far as I know, consumer CPUs usually have twice as many logical cores as physical cores.
Of course this could be different for processors that are designed for servers etc.

But that's not really important, I just wanted to make the docs less mistakable for others (not even myself, I knew already) who don't know what logical cores are (which I'm assuming are most people).  
Because they will simply read this as "CPU cores" and configure Playwright based on wrong assumptions, as seen in the Slack example.

So I would be fine with a different wording that doesn't include the wrong info (logical cores = physical cores * 2), if you have any suggestions. :)

PR #17552, opened 2022-09-23
docs: fix failed POM test example

depapp:
hi all,

there's failed test example in https://playwright.dev/docs/pom related to wrong selector & wrong text assertion


debs-obrien:
great thanks :)

PR #17558, opened 2022-09-23
chore: custom error if HTML report gets uploaded in Trace Viewer

### LINKED ISSUE #17309: [BUG] Print helpful message when a playwright report gets uploaded in the trace viewer
Relates https://github.com/microsoft/playwright/issues/17245

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/17309

Not sure if its worth adding a test.

### DIFF
--- a/packages/trace-viewer/src/sw.ts
+++ b/packages/trace-viewer/src/sw.ts
@@ -35,13 +35,20 @@ const loadedTraces = new Map<string, { traceModel: TraceModel, snapshotServer: S
 
 const clientIdToTraceUrls = new MultiMap<string, string>();
 
+class PrettyFetchError extends Error { }
+
 async function loadTrace(trace: string, clientId: string, progress: (done: number, total: number) => void): Promise<TraceModel> {
   const entry = loadedTraces.get(trace);
   clientIdToTraceUrls.set(clientId, trace);
   if (entry)
     return entry.traceModel;
   const traceModel = new TraceModel();
-  await traceModel.load(trace, progress);
+  try {
+    await traceModel.load(trace, progress);
+  } catch (error) {
+    if (error.message?.includes('Cannot find .trace file') && await traceModel.hasEntry('index.html'))
+      throw new PrettyFetchError('Could not load trace. Did you upload a Playwright HTML report instead? Make sure to extract the archive first and then double-click the index.html file or put it on a web server.');
+  }
   const snapshotServer = new SnapshotServer(traceModel.storage());
   loadedTraces.set(trace, { traceModel, snapshotServer });
   return traceModel;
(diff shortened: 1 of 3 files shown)


packages/trace-viewer/src/traceModel.ts
```diff
@@ -58,6 +59,8 @@ export class TraceModel {
         networkEntry = entry;
       if (entry.filename.includes('src@'))
         this.contextEntry.hasSource = true;
+      if (entry.filename === 'index.html')
+        throw new CustomFetchError('Could not load trace. Did you upload a Playwright HTML report instead? Make sure to extract the archive first and then double-click the index.html file or put it on a web server.');
```
dgozman, line 63: This is a wrong place to have this error. Should be somewhere in the UI, call `traceModel.entry('index.html')` or similar.

packages/trace-viewer/src/sw.ts
```diff
@@ -35,12 +35,19 @@ const loadedTraces = new Map<string, { traceModel: TraceModel, snapshotServer: S
...
+    throw new PrettyFetchError('Could not load trace. Did you upload a Playwright HTML report instead? Make sure to extract the archive first and then double-click the index.html file or put it on a web server.');
+}
+
 async function loadTrace(trace: string, clientId: string, progress: (done: number, total: number) => void): Promise<TraceModel> {
   const entry = loadedTraces.get(trace);
   clientIdToTraceUrls.set(clientId, trace);
   if (entry)
     return entry.traceModel;
-  const traceModel = new TraceModel();
+  const traceModel = new TraceModel(validateZipEntry);
```
dgozman, line 50: I'd do the following:
- Make `TraceModel.load` throw a specific error when `traceEntry` is not found.
- Catch that error, search for the `index.html` entry in `traceModel` and provide a helpful error message.

That said, current implementation is ok, if it wouldn't be failing for valid trace files that also  have a stray `index.html` file inside.

packages/trace-viewer/src/sw.ts
```diff
@@ -35,13 +35,24 @@ const loadedTraces = new Map<string, { traceModel: TraceModel, snapshotServer: S
...
   const entry = loadedTraces.get(trace);
   clientIdToTraceUrls.set(clientId, trace);
   if (entry)
     return entry.traceModel;
   const traceModel = new TraceModel();
-  await traceModel.load(trace, progress);
+  try {
+    await traceModel.load(trace, progress);
+  } catch (error) {
+    if (error instanceof TraceModelNotFoundError && traceModel.zipReader) {
```
dgozman, line 49:
```js
if (error.message?.includes('Cannot find .trace file') && traceModel.hasEntry('index.html')
  throw new Error(...)
```

packages/trace-viewer/src/traceModel.ts
```diff
@@ -60,10 +67,13 @@ export class TraceModel {
         this.contextEntry.hasSource = true;
       this._entries.set(entry.filename, entry);
     }
+    if (!traceEntry)
+      throw new TraceModelNotFoundError();
```
dgozman, line 71:
```suggestion
      throw new Error('Cannot find .trace file');
```

dgozman (approved):

PR #17567, opened 2022-09-24
chore: add snippet to the json report

pavelfeldman:


packages/playwright-test/src/reporters/json.ts
```diff
@@ -182,6 +183,7 @@ class JSONReporter implements Reporter {
       status: result.status,
       duration: result.duration,
       error: result.error,
+      errors: this._collectStepErrors(result),
```
dgozman, line 186: What about `result.errors`? There could be multiple.

dgozman (approved):

PR #17577, opened 2022-09-26
feat(api): introduce get/getByText/getByRole

pavelfeldman:


sand4rt:
Cool! May i ask here if other queries will be added as well?: 

- [ByLabelText](https://testing-library.com/docs/queries/bylabeltext)
- [ByPlaceholderText](https://testing-library.com/docs/queries/byplaceholdertext)
- [ByDisplayValue](https://testing-library.com/docs/queries/bydisplayvalue)
- [ByAltText](https://testing-library.com/docs/queries/byalttext)
- [ByTitle](https://testing-library.com/docs/queries/bytitle)
- [ByTestId](https://testing-library.com/docs/queries/bytestid)

Then we probably don't have to use https://github.com/testing-library/playwright-testing-library

packages/playwright-core/src/client/page.ts
```diff
@@ -564,10 +564,26 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
     return this.mainFrame().locator(selector, options);
   }
 
+  get(selector: string, options?: LocatorOptions): Locator {
+    return this.mainFrame().locator(selector, options);
```
Meir017, line 568: consider aligning all functions to use `this.mainFrame()` instead of `this._mainFrame` or vise-versa

pavelfeldman:
> Cool! May i ask here if other queries will be added as well?:

- `getByTestId` is planned
- `byLabelText` - playwright will resolve it for you on its own, so you can just use `getByText`
- `byPlaceholderText` - we might add `{ placeholder: true }` in `getByText`
- `byDisplayValue` - we have not seen this one used, would be curious about the use case
- `byAltText` - same as placeholder
- `byTitle` - same as placeholder

docs/src/api/class-frame.md
```diff
@@ -887,6 +887,18 @@ await locator.ClickAsync();
 * since: v1.17
 
 
+## method: Frame.get
+* since: v1.27
+- returns: <[Locator]>
+
+The method finds an element matching the specified selector in the `Locator`'s subtree. It also accepts filter options, similar to [`method: Locator.filter`] method.
```
dgozman, line 894:
```suggestion
The method finds an element matching the specified selector inside the frame. It also accepts filter options, similar to [`method: Locator.filter`] method.
```

docs/src/api/class-framelocator.md
```diff
@@ -114,6 +114,39 @@ in that iframe.
 * since: v1.17
 
 
+## method: FrameLocator.get
+* since: v1.27
+- returns: <[Locator]>
+
+The method finds an element matching the specified selector in the `Locator`'s subtree. It also accepts filter options, similar to [`method: Locator.filter`] method.
```
dgozman, line 121:
```suggestion
The method finds an element matching the specified selector inside the frame. It also accepts filter options, similar to [`method: Locator.filter`] method.
```

docs/src/api/class-page.md
```diff
@@ -2162,6 +2162,18 @@ await locator.ClickAsync();
 
 An array of all frames attached to the page.
 
+## method: Page.get
+* since: v1.27
+- returns: <[Locator]>
+
+The method finds an element matching the specified selector in the `Locator`'s subtree. It also accepts filter options, similar to [`method: Locator.filter`] method.
```
dgozman, line 2169:
```suggestion
The method finds an element matching the specified selector. It also accepts filter options, similar to [`method: Locator.filter`] method.
```

packages/playwright-core/src/client/frame.ts
```diff
@@ -298,6 +299,18 @@ export class Frame extends ChannelOwner<channels.FrameChannel> implements api.Fr
...
 
+  get(selector: string, options?: LocatorOptions): Locator {
+    return this.locator(selector, options);
+  }
+
+  getByText(text: string | RegExp, options?: { exact?: boolean }): Locator {
+    return this.locator(Locator.getByTextSelector(text, options));
+  }
+
+  getByRole(role: string, options: ByRoleOptions = {}): api.Locator {
```
dgozman, line 310:
```suggestion
  getByRole(role: string, options: ByRoleOptions = {}): Locator {
```

packages/playwright-core/src/client/locator.ts
```diff
@@ -31,10 +31,47 @@ export type LocatorOptions = {
...
+
 export class Locator implements api.Locator {
   _frame: Frame;
   _selector: string;
 
+  static getByTextSelector(text: string | RegExp, options?: { exact?: boolean }): string {
+    if (!isString(text))
+      return `text=${text}`;
+    const escaped = JSON.stringify(text);
+    const selector = options?.exact ? `text=${escaped}` : `text=${escaped.substring(1, escaped.length - 1)}`;
```
dgozman, line 52: I am not exactly sure that non-quoted escaped text is what we need. We should double check whether `text=foo\"bar` actually parses with slash or without it.

packages/playwright-core/src/client/locator.ts
```diff
@@ -31,10 +31,47 @@ export type LocatorOptions = {
...
+    if (options.disabled !== undefined)
+      props.push(['disabled', String(options.disabled)]);
+    if (options.expanded !== undefined)
+      props.push(['expanded', String(options.expanded)]);
+    if (options.includeHidden !== undefined)
+      props.push(['include-hidden', String(options.includeHidden)]);
+    if (options.level !== undefined)
+      props.push(['level', String(options.level)]);
+    if (options.name !== undefined)
+      props.push(['name', String(options.name)]);
```
dgozman, line 69: I think this should get the same treatment as `getByText` with regex vs string?

packages/playwright-core/src/client/page.ts
```diff
@@ -564,6 +564,18 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
...
+  }
+
+  getByText(text: string | RegExp, options?: { exact?: boolean }): Locator {
+    return this.mainFrame().getByText(text, options);
+  }
+
+  getByRole(role: string, options: ByRoleOptions = {}): api.Locator {
```
dgozman, line 575:
```suggestion
  getByRole(role: string, options: ByRoleOptions = {}): Locator {
```

utils/generate_types/overrides.d.ts
```diff
@@ -335,6 +335,7 @@ export type AndroidKey =
 
 export const _electron: Electron;
 export const _android: Android;
+export type BooleanWithMixed = true | false | 'mixed';
```
dgozman, line 338: I don't think we use this type in generated types, do we? We should remove it.

dgozman (approved):

sand4rt:
byDisplayValue returns the input, textarea, or select element that has the matching display value:

```html
<input type="text" id="lastName" />
```

```typescript
document.getElementById('lastName').value = 'Norris'
```

```typescript
component.getByDisplayValue('Norris')
```

pavelfeldman:
> Thanks for the reply! byDisplayValue returns the input, textarea, or select element that has the matching display value:

@sand4rt: yes, I get what it does, but I'm not sure why :)

sand4rt:
"playwright will resolve it for you on its own, so you can just use `getByText`"

Wonder if the `getByText` will be to greedy if it also matches label texts? When label is matched, does it also return the element associated with that label? For example:

```html

// for/htmlFor relationship between label and form element id
<label for="username-input">Username</label>
<input id="username-input" />

// The aria-labelledby attribute with form elements
<label id="username-label">Username</label>
<input aria-labelledby="username-label" />

// Wrapper labels
<label>Username <input /></label>

// Wrapper labels where the label text is in another child element
<label>
  <span>Username</span>
  <input />
</label>

// aria-label attributes
// Take care because this is not a label that users can see on the page,
// so the purpose of your input must be obvious to visual users.
<input aria-label="Username" />
```

```typescript
const input = component.getByLabelText('Username')
```

pavelfeldman:
> Wonder if the getByText will be to greedy if it also matches label texts? When label is matched, does it also return the element associated with that label? For example:

Yes, it will throw due to strictness.

pavelfeldman:
@sand4rt feel free to file those separately. The goal is to allow typed api for role and text selectors, not to clone the TL api. We expect them to be consistent, but not the same.

PR #17587, opened 2022-09-26
docs(nodejs): extend geolocation context examples

timdeschryver:
The [geolocation](https://playwright.dev/docs/emulation#geolocation) docs mention that the location can be overwritten with `context.setGeolocation`. These examples are easy to follow for library usages, but within a test environment, the example doesn't show how to access the `context`.

This PR, extend the examples to make it aware if it's run in a test environment, or in a library environment.

Before:
image

After:
image

### DIFF
--- a/docs/src/emulation.md
+++ b/docs/src/emulation.md
@@ -553,7 +553,35 @@ await using var context = await browser.NewContextAsync(new()
 
 Change the location later:
 
-```js
+```js tab=js-ts
+import { test, expect } from '@playwright/test';
+
+test.use({ 
+  geolocation: { longitude: 48.858455, latitude: 2.294474 },
+  permissions: ['geolocation'],
+});
+
+test('my test with geolocation', async ({ page, context }) => {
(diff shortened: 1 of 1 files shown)


docs/src/emulation.md
```diff
@@ -553,7 +553,35 @@ await using var context = await browser.NewContextAsync(new()
...
+import { test, expect } from '@playwright/test';
+
+test.use({ 
+  geolocation: { longitude: 48.858455, latitude: 2.294474 },
+  permissions: ['geolocation'],
+});
+
+test('my test with geolocation', async ({ page, context }) => {
+  // overwrite the location for this test
+  context.setGeolocation({ longitude: 29.979097, latitude: 31.134256 });
```
pavelfeldman, line 566: We would not really recomment making this call to override the geolocation of the test, test.use should be used for that. But you can still use this method once (or several times) to actually update the location and trigger respective web events.

pavelfeldman (approved):

PR #17595, opened 2022-09-26
fix: ignore timing data when request served from memory cache

### LINKED ISSUE #1080: [Question]If a interface used memory cache the request.timing().responseEnd ‘s  value is always  same.
helllo，I find if a interface used memory cache then  I use page.reload() and collect the requests's info . then I find two question.

the first question is the interface used memory cache  ,I  view the interface  in devtools（ f12） , the time is 0ms.then view the request.timing().responseEnd in playwright .It shows dozens of milliseconds.
I'm confused why caused the dozens of milliseconds.
image
image
image

the second question is  if Used the page.reload and the interface used memory cache  . I find the value of request.timing().responseEnd is the same. No matter how many times the page is refreshed.

image

I sincerely look forward to your answer. thank you.

yury-s:
`Response.timing` contains stale data when the request is served from memory cache, in that we should ignore it and return -1 where we don't know the value.

Fixes https://github.com/microsoft/playwright-java/issues/1080


tests/library/resource-timing.spec.ts
```diff
@@ -96,6 +96,34 @@ it('should work for redirect', async ({ contextFactory, browserName, server }) =
...
+  const [response] = await Promise.all([
+    page.waitForResponse('**/one-style.css'),
+    page.reload()
+  ]);
+  await response.finished();
+
+  const timing = response.request().timing();
+  verifyConnectionTimingConsistency(timing);
+
+  expect(timing.responseStart).toBe(-1);
```
dgozman, line 122: I'd prefer that both `responseStart` and `responseEnd` are either present or missing.

dgozman (approved):

PR #17634, opened 2022-09-27
feat: expose transferSize for all responses

vigneshshanmugam:
+ expose the missing `transferSize` information on all network data when `request.sizes()` is invoked.

### DIFF
--- a/packages/playwright-core/types/types.d.ts
+++ b/packages/playwright-core/types/types.d.ts
@@ -15299,6 +15299,11 @@ export interface Request {
      * Total number of bytes from the start of the HTTP response message until (and including) the double CRLF before the body.
      */
     responseHeadersSize: number;
+
+    /**
+     * Total number of bytes received from the start of the HTTP response including the header and body sizes.
+     */
+    transferSize: number;
   }>;
 
   /**
--- a/docs/src/api/class-request.md
+++ b/docs/src/api/class-request.md
@@ -244,6 +244,7 @@ The Service [Worker] that is performing the request.
   - `requestHeadersSize` <[int]> Total number of bytes from the start of the HTTP request message until (and including) the double CRLF before the body.
   - `responseBodySize` <[int]> Size of the received response body (encoded) in bytes.
   - `responseHeadersSize` <[int]> Total number of bytes from the start of the HTTP response message until (and including) the double CRLF before the body.
+  - `transferSize` <[int]> Total number of bytes received from the start of the HTTP response including the header and body sizes.
 
 Returns resource size information for given request.
 
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -416,6 +416,7 @@ export type RequestSizes = {
   requestHeadersSize: number;
   responseBodySize: number;
   responseHeadersSize: number;
+  transferSize: number;
 };
 
 export class Response extends ChannelOwner<channels.ResponseChannel> implements api.Response {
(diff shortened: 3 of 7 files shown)


packages/playwright-core/types/types.d.ts
```diff
@@ -15299,6 +15299,11 @@ export interface Request {
      * Total number of bytes from the start of the HTTP response message until (and including) the double CRLF before the body.
      */
     responseHeadersSize: number;
+
+    /**
+     * Total number of bytes received from the start of the HTTP response including the header and body sizes.
+     */
+    transferSize: number;
```
yury-s, line 15306: It can be easily computed using existing API, so I don't think we want to complicate the API by adding another field that just a some of the other two. Can you start with filing an issue describing the use cases which are not addressed by current API? We usually use issue tracker to decide if a feature is worth implementing.

vigneshshanmugam: I did try using the existing API and the values were bit mixed up. The reason why I added this to the API was
1. The value of `transferSize = resHeaderSize + responseBodySize` was not always correct, you can also see from this failed test - https://github.com/microsoft/playwright/actions/runs/3137989206/jobs/5096813728. Maybe it could be a Firefox bug alone. 
2. The API was inconsistent between the `server/network` vs `client/network` which makes it difficult for users relying on their own tracer instead of using HarTracer which uses the other API - https://github.com/microsoft/playwright/blob/2a85806cc0a903f676036af7f5fab603ac45205e/packages/playwright-core/src/server/har/harTracer.ts#L339

Happy to close the PR, if you feel its not worth it.

yury-s: By definition `transferSize = resHeaderSize + responseBodySize` from what you describe it seems that there is a bug and this invariant doesn't hold for Firefox. We should fix the instead of exposing another value which is inconsistent with the other two.

yury-s: The header size is computed manually on the client end if browser doesn't provide exact values. This is the case in Firefox and believe we can improve that by changing the code like this because 2 out of 3 value do come from the browser there:

```diff
diff --git a/packages/playwright-core/src/server/firefox/ffNetworkManager.ts b/packages/playwright-core/src/server/firefox/ffNetworkManager.ts
index 2fdc8850b..164e038b0 100644
--- a/packages/playwright-core/src/server/firefox/ffNetworkManager.ts
+++ b/packages/playwright-core/src/server/firefox/ffNetworkManager.ts
@@ -115,8 +115,6 @@ export class FFNetworkManager {
     });
     // "raw" headers are the same as "provisional" headers in Firefox.
     response.setRawResponseHeaders(null);
-    // Headers size are not available in Firefox.
-    response.setResponseHeadersSize(null);
     this._page._frameManager.requestReceivedResponse(response);
   }
 
@@ -127,6 +125,8 @@ export class FFNetworkManager {
     const response = request.request._existingResponse()!;
     response.setTransferSize(event.transferSize);
     response.setEncodedBodySize(event.encodedBodySize);
+    // Headers size are not available in Firefox.
+    response.setResponseHeadersSize(event.transferSize - event.encodedBodySize);
 
     // Keep redirected requests in the map for future reference as redirectedFrom.
     const isRedirected = response.status() >= 300 && response.status() <= 399;
```

vigneshshanmugam: Thanks for the details around the FF implementation. 

> The API was inconsistent between the server/network vs client/network which makes it difficult for users relying on their own tracer instead of using HarTracer which uses the other API/ 

Do you think if the above is a valid use case? if not, we can close this PR and fix the Firefox issue separately.

pavelfeldman:
Reviewing this change seems to require a larger investment than we can grant, so I'll take a liberty of closing it. We have 700 issues to prioritize, so we need to respect the overall project priorities.

PR #17636, opened 2022-09-27
feat: allow opening multiple html reporters and trace viewers

### LINKED ISSUE #17201: [Feature] Allow multiple local trace viewers to run
The trace viewer displays a lot of information in a compact manner. Currently, when running locally, it uses a 
specific port. This means only one trace viewer can be opened on a computer and a second invocation will get an
error :

   > npx playwright show-trace
   > events.js:352
   >       throw er; // Unhandled 'error' event
   >       ^
   > 
   > Error: listen EADDRINUSE: address already in use 127.0.0.1:9322
   >     at Server.setupListenHandle [as _listen2] (net.js:1320:16)
   >     at listenInCluster (net.js:1368:12)
   >     at doListen (net.js:1505:7)
   >     at processTicksAndRejections (internal/process/task_queues.js:83:21)
   > Emitted 'error' event on Server instance at:
   >     at emitErrorNT (net.js:1347:8)
   >     at processTicksAndRejections (internal/process/task_queues.js:82:21) {
   >   code: 'EADDRINUSE',
   >   errno: -4091,
   >   syscall: 'listen',
   >   address: '127.0.0.1',
   >   port: 9322
   > }

In order to investigate a test failure, I generated traces for both a passing and failing (same) test. It would be useful
to be able to view both traces in their own viewers to narrow down the issues. The current workaround is to view 
one trace locally and the second on the web with trace.playwright.dev.

It would be useful to update the trace viewer to use any available local port so as to allow multiple instances of 
the viewer.

Thanks

dgozman:
This makes `HttpServer` accept `preferredPort` option that will first try to listen on that port, and if that port is already in use, listen on some available port instead.

Fixes #17201.

### DIFF
--- a/packages/playwright-core/src/grid/gridServer.ts
+++ b/packages/playwright-core/src/grid/gridServer.ts
@@ -393,7 +393,7 @@ export class GridServer {
   }
 
   async start(port?: number) {
-    await this._server.start(port);
+    await this._server.start({ port });
   }
 
   gridURL(): string {
--- a/packages/playwright-core/src/server/trace/viewer/traceViewer.ts
+++ b/packages/playwright-core/src/server/trace/viewer/traceViewer.ts
@@ -26,7 +26,7 @@ import { serverSideCallMetadata } from '../../instrumentation';
 import { createPlaywright } from '../../playwright';
 import { ProgressController } from '../../progress';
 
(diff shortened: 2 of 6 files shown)


yury-s (approved):
Do we actually need preferredPort ?

PR #17645, opened 2022-09-27
feat(api): introduce getByTestId

pavelfeldman:


packages/playwright-core/src/client/locator.ts
```diff
@@ -46,6 +46,15 @@ export class Locator implements api.Locator {
   _frame: Frame;
   _selector: string;
 
+  static _testIdAttributeName = 'data-testid';
+  static _setTestIdAttribute(attributeName: string) {
+    Locator._testIdAttributeName = attributeName;
+  }
+
+  static getByTestIdSelector(testId: string): string {
+    return `css=[${Locator._testIdAttributeName}=${testId}]`;
```
dgozman, line 55: I think this needs some escaping :)

dgozman (approved):

packages/playwright-core/src/client/locator.ts
pavelfeldman: What do you mean?

PR #17651, opened 2022-09-28
feat(core): type addInitScript function arg

Trinovantes:
Allows autocomplete on the arg passed to the init script

```ts
await page.addInitScript((arg) => {
    window.localStorage.setItem('init_user_state', JSON.stringify(arg.initUserState)) // arg was previously 'any'
}, {
    initUserState: { foo: 'bar' },
})
```

### DIFF
--- a/utils/generate_types/overrides.d.ts
+++ b/utils/generate_types/overrides.d.ts
@@ -34,6 +34,8 @@ export interface Page {
   evaluateHandle<R, Arg>(pageFunction: PageFunction<Arg, R>, arg: Arg): Promise<SmartHandle<R>>;
   evaluateHandle<R>(pageFunction: PageFunction<void, R>, arg?: any): Promise<SmartHandle<R>>;
 
+  addInitScript<Arg>(script: PageFunction<Arg, any> | { path?: string, content?: string }, arg?: Arg): Promise<void>;
+
   $<K extends keyof HTMLElementTagNameMap>(selector: K, options?: { strict: boolean }): Promise<ElementHandleForTag<K> | null>;
   $(selector: string, options?: { strict: boolean }): Promise<ElementHandle<SVGElement | HTMLElement> | null>;
 
@@ -97,6 +99,8 @@ export interface Frame {
 export interface BrowserContext {
   exposeBinding(name: string, playwrightBinding: (source: BindingSource, arg: JSHandle) => any, options: { handle: true }): Promise<void>;
   exposeBinding(name: string, playwrightBinding: (source: BindingSource, ...args: any[]) => any, options?: { handle?: boolean }): Promise<void>;
+
+  addInitScript<Arg>(script: PageFunction<Arg, any> | { path?: string, content?: string }, arg?: Arg): Promise<void>;
 }
 
 export interface Worker {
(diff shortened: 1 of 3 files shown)


Trinovantes:

packages/playwright-core/src/client/browserContext.ts
```diff
@@ -237,7 +237,7 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
     await this._channel.setHTTPCredentials({ httpCredentials: httpCredentials || undefined });
   }
 
-  async addInitScript(script: Function | string | { path?: string, content?: string }, arg?: any): Promise<void> {
+  async addInitScript<Arg>(script: structs.PageFunction<Arg, void> | { path?: string, content?: string }, arg?: Arg): Promise<void> {
```
dgozman, line 240: The change should instead go [here](https://github.com/microsoft/playwright/blob/main/utils/generate_types/overrides.d.ts#L30) and [here](https://github.com/microsoft/playwright/blob/main/utils/generate_types/overrides.d.ts#L97).

Ideally, a test snippet [like this](https://github.com/microsoft/playwright/blob/main/utils/generate_types/test/test.ts#L574-L576) would be great to ensure types work as expected.

Make sure to run `npm run watch` and `npm run lint` to check that everything works.

dgozman (commented):
Thank you for the PR!

packages/playwright-core/src/client/browserContext.ts
Trinovantes: I've made the changes in the overrides file and added a test

I had to add line breaks in existing tests that use `addInitScript` to ensure their script function returned void

utils/generate_types/overrides.d.ts
```diff
@@ -34,6 +34,8 @@ export interface Page {
   evaluateHandle<R, Arg>(pageFunction: PageFunction<Arg, R>, arg: Arg): Promise<SmartHandle<R>>;
   evaluateHandle<R>(pageFunction: PageFunction<void, R>, arg?: any): Promise<SmartHandle<R>>;
 
+  addInitScript<Arg>(script: PageFunction<Arg, void> | { path?: string, content?: string }, arg?: Arg): Promise<void>;
```
dgozman, line 37:
```suggestion
  addInitScript<Arg>(script: PageFunction<Arg, any> | { path?: string, content?: string }, arg?: Arg): Promise<void>;
```

This way you will avoid issues where init function returns something. Otherwise it would be a breaking change, as seen by our own tests.

utils/generate_types/overrides.d.ts
```diff
@@ -97,6 +99,8 @@ export interface Frame {
 export interface BrowserContext {
   exposeBinding(name: string, playwrightBinding: (source: BindingSource, arg: JSHandle) => any, options: { handle: true }): Promise<void>;
   exposeBinding(name: string, playwrightBinding: (source: BindingSource, ...args: any[]) => any, options?: { handle?: boolean }): Promise<void>;
+
+  addInitScript<Arg>(script: PageFunction<Arg, void> | { path?: string, content?: string }, arg?: Arg): Promise<void>;
```
dgozman, line 103:
```suggestion
  addInitScript<Arg>(script: PageFunction<Arg, any> | { path?: string, content?: string }, arg?: Arg): Promise<void>;
```

dgozman (commented):
Looks very good, but I think we should avoid the regression.

dgozman (approved):
Looks great, merging in. Thank you for the PR!

PR #17666, opened 2022-09-28
[Feature]: Locator.toArray method

vitalics:
As a part of https://github.com/microsoft/playwright/issues/12336 crates locator new api: toArray and for JS implementation - make Locator async iterable. See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of)

TODO: complete examples for python, java and C#

### DIFF
--- a/docs/src/api/class-locator.md
+++ b/docs/src/api/class-locator.md
@@ -823,6 +823,19 @@ Returns locator to the n-th matching element. It's zero based, `nth(0)` selects 
 * since: v1.14
 - `index` <[int]>
 
+
+## async method: Locator.toArray
+* since: v1.27
+- returns: <[Array]<[Locator]>>
+
+Returns shallow array of locators
+
+## property: Locator.[Symbol.asyncIterator]()
+* since: v1.27
+- type: <[AsyncIterator]<[Locator]>>
+
+Returns iterator of array representation
+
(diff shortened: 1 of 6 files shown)


jfgreffier:
I don't see this as necessary, as it can probably be done in the user land. Also, there are several ways to manipulate Locator lists https://playwright.dev/docs/locators#lists

> TODO: complete examples for python, java and C#

You can't do something in core only for TypeScript. That would be... bad

vitalics:
@jfgreffier In my case of using playwright is reading a tons of DOM elements and working with such ways is disgusting and you should a lot of spagetti code to do the same things. As for me it will be quite useful to add this API for locator in JS implementation, since such API can reduce the code

pavelfeldman:
Let's publish some real-world examples into that issue first.

vitalics:
@pavelfeldman You mean examples in docs guide, or provide separate repo where `toArray` will be helpful?

pavelfeldman:
> @pavelfeldman You mean examples in docs guide, or provide separate repo where toArray will be helpful?

I meant let's update the issue (or better yet create a new one) with the proposal and the use cases it helps with! I'll close this PR while we are getting the proposal sorted!

PR #17684, opened 2022-09-29
feat(api): add getByLabelText

pavelfeldman:


docs/src/api/params.md
```diff
@@ -1180,6 +1180,10 @@ Locate element by the test id. By default, the `data-testid` attribute is used a
 
 Allows locating elements that contain given text.
 
+## template-locator-get-by-label-text
+
+Allows locating input elements with the labels containinng given text.
```
dgozman, line 1185: Allows locating input elements by the text of the associated label. For example, this method will find the `input` by label text `Password` in the following DOM:

```html
<label for="password-input">Password:</label>
<input id="password-input">
```

dgozman (approved):

PR #17716, opened 2022-09-29
fix: --grep and --grep-invert should intersect with config

### LINKED ISSUE #17405: [BUG] Command line `--grep-invert` option overrides project-level `grepInvert` configuration
**Context:**
- Playwright Version: 1.25.2
## System:
 - OS: macOS 12.5.1
 - Memory: 31.85 MB / 16.00 GB
## Binaries:
 - Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
 - npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash

**Describe the bug**

We are running into an issue where the `--grep-invert` option on the command line is overriding the `grepInvert` for our individual project configurations.

Our use case is as such:

- We are executing our nightly test run, which runs against all of our project configurations.
- We have multiple projects, but we only want to run tests annotated with `@snapshot` on one of them. So we have configured all projects except one to have `grepInvert: /@snapshot/`
- We have tests annotated with `@couchdb` which we don't want to run as part of our full nightly test run. So we have added the `--grep-invert='@couchdb'` command line option to our npm script for triggering a full nightly run.

But it seems that the command line option is overriding the individual project configurations as `@snapshot` tests are being run by all projects. One solution which would work for us is if the regexes can be compounded together as if it were an 'AND'. If this is not possible, it would be nice to have this behavior documented so it's clear you should use one or the other.

I have a repro repo located here: https://github.com/ozyx/grepInvert-repro
Run `npm run test` to see the behavior.

yury-s:
Fixes https://github.com/microsoft/playwright/issues/17405


packages/playwright-test/src/runner.ts
```diff
@@ -289,7 +288,7 @@ export class Runner {
 
   private _runPhaseFromOptions(options: RunOptions): RunPhase {
     const testFileMatcher = fileMatcherFrom(options.testFileFilters);
-    const testTitleMatcher = () => true;
+    const testTitleMatcher = options.testTitleMatcher;
```
dgozman, line 291: Do we take `options.testTitleMatcher` into account when running groups? I think a similar TODO as we have to `testFileFilters`?

dgozman (approved):

packages/playwright-test/src/runner.ts
yury-s: Currently we don't,  added TODO.

PR #17721, opened 2022-09-30
chore: turn RunPhase in a class with helper methods

yury-s:


packages/playwright-test/src/runner.ts
```diff
@@ -251,14 +273,13 @@ export class Runner {
         throw new Error(`Cannot find project group '${projectGroup}' in the config`);
       for (const entry of group) {
```
dgozman, line 274: This whole method could move to `RunPhase`?

packages/playwright-test/src/runner.ts
```diff
@@ -424,6 +444,15 @@ export class Runner {
       const testGroups = createTestGroups(projectSuites, config.workers);
       concurrentTestGroups.push(testGroups);
     }
+    return { rootSuite, concurrentTestGroups };
+  }
+
+  private async _run(options: RunOptions): Promise<FullResult> {
```
dgozman, line 450: Why this method split? Now `7. Fail when no tests` seems strange :)

yury-s: For better readability. Removed numbers.

packages/playwright-test/src/runner.ts
yury-s: Done.

dgozman (approved):

PR #17757, opened 2022-10-01
feat(codegen) - codeceptjs - playwright

kobenguyent:
Codegen for CodeceptJS

Screen Shot 2022-10-01 at 4 04 50 PM


gkushang:
looking forward to this 🎉

basalona:
This would be a great addition to CodeceptJS features 👏

pavelfeldman:
This would sadly be outside of the scope for Playwright. We can start with making codegen extensible, but the point of Playwright codegen is really to generate Playwright scripts only. It would be easier to maintain a separate codegen that emits Codeceptjs!

Issue #15889, opened 2022-07-22
[Feature] Component testing and page.goto


sand4rt:
**Context:**
- Playwright Version: 1.24.0
- Operating System: All
- Node.js version:  v17.3.0
- Browser: Chromium
- Extra: Vue component testing

I've been playing with the Vue component testing feature and it seems like the `page.goto()` is not working properly.

For example 
[i have a failing test that should be passing](https://github.com/sand4rt/TVmaze/blob/master/src/Features/NotFound/NotFoundPage.test.ts).

It should navigate to the 404 page because i have a catch all specified [`/:pathMatch(.*)*`](https://router.vuejs.org/guide/essentials/dynamic-matching.html#catch-all-404-not-found-route) inside Vue router [see the code here](https://github.com/sand4rt/TVmaze/blob/master/src/Router/Index.ts#L52).

The Vue router plugin is registered [here](https://github.com/sand4rt/TVmaze/blob/master/playwright/index.ts).

[Here](https://github.com/sand4rt/TVmaze/blob/master/src/Features/Browse/BrowsePage.test.ts#L61) you can see another example that also uses the `page.goto` function that i can't get to work.

sand4rt:
[Here](https://github.com/sand4rt/TVmaze/blob/master/src/Features/Browse/BrowsePage.test.ts#L61) you can see another example that also uses the `page.goto` function that i can it goes to work.

pavelfeldman:
I believe `page.goto` works as expected here. What it does is it navigates the page to the specified URL and navigation is resetting the world along with all the JS state, as it is supposed to. No wonder none of the installed routers work. Non-anchor navigation is equivalent to going to a new URL via `about:blank`. In order to perform the navigation that router can handle, you should actually click on a router-aware link or you should know the mechanics that the framework uses to support routing. In most cases it'll be history API, so instead of `page.goto` you should do

```js
await page.evaluate(() => {
  window.history.pushState('', '', '/route-not-exsist-5b45f7cd193ecc56339c64428079706a');
});
```

It is saying "Perform SPA navigation to the new URL that history controller will handle".

pavelfeldman:
For the second use case, it would be something like this:

```js
// Typically mount() navigates to the component registry page (playwright/index.html)
// I'm doing it manually here, but we can probably make it work automatically as the test starts. It will install the router.
await page.goto(process.env.PLAYWRIGHT_VITE_COMPONENTS_BASE_URL!);
[...]

PR #17767, opened 2022-10-02
fix(ct): vite base url

sand4rt:
Vue router and probably others are not aware that playwright opens the page in a subfolder/url. So when a page is rendered, Vue router doesn't know what the url is and will throw a page not found.

Usually this can be resolved by specifying a base url with the subfolder/url included. For example trough [`createWebHistory`](https://github.com/microsoft/playwright/pull/17767/files#diff-cd3339bf4bd46b9793e091c73ef565d763439eae202bb579ba22626667d62e6cR6). 

However it's quite annoying to set this base URL including subfolder for the tests only with environment variables. Hence why i think it would be easier if playwright doesn't open the page in a subfolder/url.

**Warning**: this is will be a breaking change
       1. The url is changed from http://localhost:3100/playwright/index.html to http://localhost:3100
<s>2. The imports in `/playwright/index.html` doesn't need to include `/playwright` subfolder any more. `<script type="module" src="/playwright/index.js" />` is now: `<script type="module" src="/index.js" />`</s>


sand4rt:
@pavelfeldman before i look at the failing tests, what do you think about this?

sand4rt:
I forgot to mention that https://github.com/microsoft/playwright/issues/15889 is blocked until this PR is resolved. Not sure yet if there is a way without heaving a breaking change. The only thing left todo was mounting the component on page load:

```typescript
await use((component: JsxComponent | string, options?: MountOptions) => {
  async function mount() {
    const selector = await(page as any)._wrapApiCall(async () => {
      return await innerMount(page, component, options);
    }, true);
    const locator = page.locator(selector);
    return Object.assign(locator, {
      unmount: async () => {
        await locator.evaluate(async () => {
          const rootElement = document.getElementById('root')!;
          await window.playwrightUnmount(rootElement);
        });
      },
      rerender: async (component: JsxComponent | string, options?: Omit<MountOptions, 'hooksConfig'>) => {
        await innerRerender(page, component, options);
      }
    });
  }
  page.addListener('load', mount);
  return mount();
});```

**Comment by pavelfeldman** [comment]
I'm trying to figure out if route-aware components are rather a subject of the e2e tests. It does not look like we are testing the components at this point, would it make sense to recommend using Playwright proper for everything with routes? Thinking out loud here.

File: packages/playwright-test/src/plugins/vitePlugin.ts [hunk 1]
```diff
@@ -92,7 +92,8 @@ export function createPlugin(
       // 3. Check component sources.
       const sourcesDirty = !buildExists || hasNewComponents || await checkSources(buildInfo);
 
-      viteConfig.root = rootDir;
+      viteConfig.root = templateDir;
```
pavelfeldman, line 95: At least when I started with it (Vite 2), it did not work - Vite seemed to only process / transpile files under the root. In this case, component files become outside of the /playwright/ root, so I wonder if things will keep working by the spec...

pavelfeldman: If things are indeed working with this new setup, let's leave old behavior behind environment variable, so that it was easy for the existing users to keep it working without intrusive changes.

sand4rt:
> I'm trying to figure out if route-aware components are rather a subject of the e2e tests. It does not look like we are testing the components at this point, would it make sense to recommend using Playwright proper for everything with routes? Thinking out loud here.

These are the arguments i could think of: 
- Would say that testing full user flows with component tests is not recommended. Testing simpler things like: 
`test('display a custom 404 page if the url does not exist')` or
`test('typing in the search input component changes the url /search?query=...')`
Do involve the router and are useful to component test. They are faster than e2e tests and pinpoint the errors better when something goes wrong.
- Having we half broken api's like `page.goto()`, `expect(page).toHaveUrl()`, `page.reload()` is kind of odd.
- For what it's worth, the other test libraries are supporting this and they have specific tests written for this.

packages/playwright-test/src/plugins/vitePlugin.ts
sand4rt: > At least when I started with it (Vite 2), it did not work - Vite seemed to only process / transpile files under the root. In this case, component files become outside of the /playwright/ root, so I wonder if things will keep working by the spec...

Will test it further but ATM i think there is one thing left todo:
Vite doesn't pick up the config files that are in the root of the project. So config files like `svelte.config.js` and `tailwind.config.js` have to be duplicated and placed inside the `/playwright/` as well.

sand4rt: > If things are indeed working with this new setup, let's leave old behavior behind environment variable, so that it was easy for the existing users to keep it working without intrusive changes.

Curious if you ever thought about a post-install script that migrates old API's to the new ones? As in this case that would be finding and replacing the import path's

sand4rt: Found a way without breaking existing users by moving `index.html` one directory down after build

packages/playwright-test/src/plugins/vitePlugin.ts
```diff
@@ -139,8 +139,10 @@ export function createPlugin(
         sourcemap: true,
       };
 
-      if (sourcesDirty)
+      if (sourcesDirty) {
         await build(viteConfig);
+        fs.renameSync(`${outDir}/${relativeTemplateDir}/index.html`, `${outDir}/index.html`);
```
pavelfeldman, line 144:
```suggestion
        await fs.promises.rename(`${outDir}/${relativeTemplateDir}/index.html`, `${outDir}/index.html`);
```

pavelfeldman (approved):
Looks good modulo baseURL question.

tests/components/ct-vue-vite/playwright.config.ts
```diff
@@ -23,8 +23,8 @@ const config: PlaywrightTestConfig = {
   reporter: 'html',
   use: {
     ctTemplateDir: 'playwright',
-    ctCacheDir: 'playwright/.cache',
     trace: 'on-first-retry',
+    baseURL: 'http://localhost:3100'
```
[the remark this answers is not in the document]
sand4rt: Otherwise `await expect(page).toHaveURL('/');` will throw an error with something along the lines of: URL is `http://localhost:3100/` instead of `/`. 

Maybe it's better to set a default `baseURL` in the `vitePlugin.ts`? Can i do that in a follow-up PR?

packages/playwright-test/src/plugins/vitePlugin.ts
sand4rt: fixed

packages/playwright-test/src/plugins/vitePlugin.ts
```diff
@@ -150,7 +153,8 @@ export function createPlugin(
       const isAddressInfo = (x: any): x is AddressInfo => x?.address;
       const address = previewServer.httpServer.address();
       if (isAddressInfo(address))
-        process.env.PLAYWRIGHT_VITE_COMPONENTS_BASE_URL = `http://localhost:${address.port}/${relativeTemplateDir}/index.html`;
+        process.env.PLAYWRIGHT_VITE_COMPONENTS_BASE_URL = `http://localhost:${address.port}`;
```
pavelfeldman, line 156: Where is this used?

pavelfeldman: For baseURL to work, you can either assign it in mount.ts or say PLAYWRIGHT_TEST_BASE_URL = ... here and the packages/playwright-test/src/index.ts will pick it up. Latter is a bit of a hack, but given that it is covered with the test, we can do that.

sand4rt: Oh, that's a lot easier than i thought it would be. I replaced `PLAYWRIGHT_VITE_COMPONENTS_BASE_URL` with `PLAYWRIGHT_TEST_BASE_URL` and used it in `mount()`. If I understand correctly, there is already a test for this, right?

PR #17791, opened 2022-10-03
fix: missing proxy in ws connection

wodyjowski:
This adds missing proxy to websocket connection
when using Selenium Grid 4

#17792

### DIFF
--- a/packages/playwright-core/src/server/transport.ts
+++ b/packages/playwright-core/src/server/transport.ts
@@ -18,7 +18,7 @@
 import { ws } from '../utilsBundle';
 import type { WebSocket } from '../utilsBundle';
 import { getProxyForUrl, HttpsProxyAgent } from '../utilsBundle';
-import * as URL from 'url';
+import { URL } from 'url';
 import type { ClientRequest, IncomingMessage, Agent } from 'http';
 import type { Progress } from './progress';
 import { makeWaitForNextTask } from '../utils';
@@ -93,11 +93,11 @@ export class WebSocketTransport implements ConnectionTransport {
   constructor(progress: Progress|undefined, url: string, headers?: { [key: string]: string; }, followRedirects?: boolean) {
     this.wsEndpoint = url;
 
-    const endpointURL = URL.parse(url);
+    const endpointURL = new URL(url);
     const proxyURL = url.startsWith('ws:') ? getProxyForUrl(`http://${endpointURL.hostname}`) : getProxyForUrl(`https://${endpointURL.hostname}`);
     let agent: Agent | undefined = undefined;
     if (proxyURL) {
-      const parsedProxyURL = URL.parse(proxyURL);
+      const parsedProxyURL = new URL(proxyURL);
       (parsedProxyURL as any).secureProxy = parsedProxyURL.protocol === `https:`;
 
       agent = new HttpsProxyAgent(parsedProxyURL);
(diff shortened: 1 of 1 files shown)


wodyjowski:

packages/playwright-core/src/server/transport.ts
```diff
@@ -90,13 +92,26 @@ export class WebSocketTransport implements ConnectionTransport {
 
   constructor(progress: Progress, url: string, headers?: { [key: string]: string; }, followRedirects?: boolean) {
     this.wsEndpoint = url;
+
+    const endpointURL = URL.parse(url);
+    const proxyURL = url.startsWith('ws') ? getProxyForUrl(`http://${endpointURL.hostname}`) : getProxyForUrl(`https://${endpointURL.hostname}`);
```
luki530, line 97: It should be `ws:` instead of `ws`, because `wss` also starts with `ws`

luki530 (approved):

packages/playwright-core/src/server/transport.ts
```diff
@@ -17,7 +17,9 @@
 
 import { ws } from '../utilsBundle';
 import type { WebSocket } from '../utilsBundle';
-import type { ClientRequest, IncomingMessage } from 'http';
+import { getProxyForUrl, HttpsProxyAgent } from '../utilsBundle';
+import * as URL from 'url';
```
pavelfeldman, line 21: Use plain URL object, url module is deprecated.

packages/playwright-core/src/server/transport.ts
```diff
@@ -17,7 +17,9 @@
 
 import { ws } from '../utilsBundle';
 import type { WebSocket } from '../utilsBundle';
-import type { ClientRequest, IncomingMessage } from 'http';
+import { getProxyForUrl, HttpsProxyAgent } from '../utilsBundle';
+import { URL } from 'url';
```
pavelfeldman, line 21: As per above, url module is deprecated, use URL object instead.

jlcd:
1. When will this be merged?
2. And also, will websocket proxies affect any Playwright+Chrome automations or only when using it together with Selenium Grid 4?

packages/playwright-core/src/server/transport.ts
```diff
@@ -90,13 +91,26 @@ export class WebSocketTransport implements ConnectionTransport {
...
+
     this._ws = new ws(url, [], {
       perMessageDeflate: false,
       maxPayload: 256 * 1024 * 1024, // 256Mb,
       // Prevent internal http client error when passing negative timeout.
       handshakeTimeout: Math.max(progress?.timeUntilDeadline() ?? 30_000, 1),
       headers,
       followRedirects,
+      agent: agent,
+      rejectUnauthorized: false,
```
pavelfeldman, line 113: Why did this change?

wodyjowski: Because you've added the same line in [here](https://github.com/microsoft/playwright/blob/b7a49fb3714705c3b487179012c778c5d2922391/packages/playwright-core/src/common/netUtils.ts#L66)

pavelfeldman: I don't think ignoring TLS in this case makes sense, let's remove this.

pavelfeldman:
Closing as stale. Sorry we were slow with it.

Issue #17804, opened 2022-10-04
[BUG] Can't use component testing in existing project


piotrpalek:
**Context:**
- Playwright Version: 1.26.1
- Operating System: Mac
- Node.js version: 16.13.0
- Browser: All

CodeSandbox link showing bug (running below repo): https://codesandbox.io/p/github/piotrpalek/template-hcp/playwright
Github link to example repo: https://github.com/piotrpalek/template-hcp/tree/playwright

**Describe the bug**

Hey, I'm trying to run playwright, on an existing webpack project, but I'm getting an issue when trying to import modules without adding a file extension. It's saying "Error: Cannot find module" when trying to run it. I've succeeded in making an example repo showing the exact same issue.

It would be pretty cumbersome (and against our current styleguide) to change all imports to have file extensions. Is there a workaround I could use here?

yury-s:
> Github link to example repo: https://github.com/piotrpalek/template-hcp/tree/playwright

I cloned this repo and run the commands you mentioned but I got the error:

```
$ npm run dev
npm ERR! Missing script: "dev"
```

piotrpalek:
@yury-s sorry the instructions were from a template, I've updated the repo with the correct one, fwiw you just need to run `yarn run test-ct` (or `yarn run start` in that repo which does the same)

yury-s:
> @yury-s sorry my bad, the instructions were from a template, I've updated the repo with the correct one :) you need to run `yarn run test-ct` (or `yarn run start` in that repo which does the same)

Still not working:
```
error Command "test-ct" not found.
```

yury-s:
> edit: I've pushed a proof of concept PR that seems to fix my issue over here: #17839

We need to be able to reproduce the problem before accepting a fix, maybe you can add a test to our suite that demonstrates the bug if it's easier?

yury-s:
> @yury-s that's weird I just tried the repo from scratch and `yarn run test-ct` works. Did you clone it from scratch? I've changed the link to point to a fresh repo to make sure.

Oh, you updated the link for CodeSandbox, I was using the second one. I can reproduce it with https://github.com/piotrpalek/playwright-issue

pavelfeldman:
@piotrpalek I can see you are using the type: module. This mode requires that all the imported paths have extensions, by the specification. Furthermore, for imported TS paths you should be using JS extensions. Any reason you want to use type: module?

See https://www.typescriptlang.org/docs/handbook/esm-node.html

[...]

PR #17839, opened 2022-10-05
fix: support imports without file extensions

piotrpalek:
POC fix for supporting imports without file extenions

related issue: https://github.com/microsoft/playwright/issues/17804


pavelfeldman:
That's not how modules are solved in Node, ESM and Typescript ESM interop mode.

PR #17853, opened 2022-10-05
fix(api): hide `TestConfig.groups`

dgozman:
Also fix types for `FullConfig.configFile`.


packages/playwright-test/src/loader.ts
```diff
@@ -165,7 +165,10 @@ export class Loader {
     this._fullConfig.metadata = takeFirst(config.metadata, baseFullConfig.metadata);
     this._fullConfig.projects = (config.projects || [config]).map(p => this._resolveProject(config, this._fullConfig, p, throwawayArtifactsPath));
     this._assignUniqueProjectIds(this._fullConfig.projects);
-    this._fullConfig.groups = config.groups;
+
+    // TODO: restore or remove once groups are decided upon.
+    this._fullConfig.groups = undefined;
```
yury-s, line 170: This will break group tests.

yury-s (approved):

PR #17910, opened 2022-10-07
Fix invalid JS/TS code examples

jfgreffier:


docs/src/emulation.md
```diff
@@ -425,18 +425,18 @@ Allow notifications for a specific domain.
...
-    permissions: ['notifications'], {origin: 'https://skype.com'},
+    permissions: ['notifications'],
   },
 };
 
 module.exports = config;
 ```
 
-```js tab=js-ts
+```ts tab=js-ts
```
pavelfeldman, line 435:
```suggestion
```js tab=js-ts
```

docs/src/test-fixtures-js.md
```diff
@@ -362,7 +362,7 @@ exports.test = base.test.extend({
 exports.expect = base.expect;
 ```
 
-```js tab=js-ts
+```ts tab=js-ts
```
pavelfeldman, line 365: All these should be js tab=js-ts

jfgreffier: These code blocks are actually TypeScript, not JavaScript. From my understanding and my tests, it does not mess up with the tabs.

I don't see any downside, am I missing something ?

jfgreffier: Hi @mxschmitt  ! Sorry for the ping, what do you think about this change ? Tag the code block with `ts` is more correct in terms of syntax, and will allow linting the code example in the future.
Debbie told me you're familiar with the docs generations. Do you see any risks or downside ?

pavelfeldman: We use ```js for both consistently and our extensive codegen in this and other projects rely on it.

jfgreffier:
Stale PR...

PR #17931, opened 2022-10-08
chore(ct): type checking

sand4rt:
Fixed all type errors and run typecheck in CI


tests/components/ct-react/src/tests.spec.tsx
```diff
@@ -137,7 +137,7 @@ test('get textContent of the empty fragment', async ({ mount }) => {
 
 const testWithServer = test.extend(serverFixtures);
 testWithServer('components routing should go through context', async ({ mount, context, server }) => {
-  server.setRoute('/hello', (req, res) => {
+  server.setRoute('/hello', (req: any, res: any) => {
```
pavelfeldman, line 140: I would imagine that these are typed, why require instead of import?

pavelfeldman (approved):
looks good modulo http server types

tests/components/ct-react/src/tests.spec.tsx
sand4rt: Made `serverFixtures` a `require` instead of `import` to ignore the type errors inside:
```code
6  ../../config/proxy.ts:19
4  ../../config/serverFixtures.ts:19
16  ../../../utils/testserver/index.ts:21
```

The errors are popping up because React's `tsconfig.json` has `strict: true`.

Don't know if it's about the same but is it an idea rewrite the test from:

```typescript
const testWithServer = test.extend(serverFixtures);
testWithServer('components routing should go through context', async ({ mount, context, server }) => {
  server.setRoute('/hello', (req, res) => {
    res.write('served via server');
    res.end();
  });

  let markRouted: (url: string) => void;
  const routedViaContext = new Promise(res => markRouted = res);
  await context.route('**/hello', async (route, request) => {
    markRouted(`${request.method()} ${request.url()}`);
    await route.fulfill({
      body: 'intercepted',
    });
  });

  const whoServedTheRequest = Promise.race([
    server.waitForRequest('/hello').then((req) => `served via server: ${req.method} ${req.url}`),
    routedViaContext.then(req => `served via context: ${req}`),
  ]);

  const component = await mount(<Fetch url={server.PREFIX + '/hello'} />);
  await expect.soft(whoServedTheRequest).resolves.toMatch(/served via context: GET.*\/hello.*/i);
  await expect.soft(component).toHaveText('intercepted');
});
```

to:

```typescript
test("intercept routing", async ({ page, mount }) => {
  await page.route(`**/hello`, async (route) => {
    await route.fulfill({
      body: JSON.stringify({ body: "intercepted" }),
    });
  });
  const component = await mount(<Fetch url="/hello" />);
  await expect(component).toHaveText("intercepted");
});
```

pavelfeldman: Ok, we can follow-up with this.

PR #17935, opened 2022-10-09
feat(ct): typed hooks

sand4rt:


packages/playwright-ct-react/hooks.d.ts
```diff
@@ -14,5 +14,9 @@
  * limitations under the License.
  */
 
-export declare function beforeMount(callback: (params: { hooksConfig: any }) => Promise<void>): void;
-export declare function afterMount(callback: (params: { hooksConfig: any }) => Promise<void>): void;
+type JsonPrimitive = string | number | boolean | null;
+type JsonValue = JsonPrimitive | JsonObject | JsonArray;
+type JsonArray = JsonValue[];
+type JsonObject = { [Key in string]?: JsonValue };
```
sand4rt, line 20: This `JsonObject` type is in a lot of places. Is there a central place where i can put it and reuse it across the different projects?

pavelfeldman: I would keep it as is for now

pavelfeldman (approved):

PR #17940, opened 2022-10-09
feat(ct): svelte rerender

sand4rt:
API is `component.rerender({ props: {}, on: {} })` 
instead of `component.rerender({ props: {}, on: {}, slots: {} })` because Svelte doesn't support slots: https://github.com/sveltejs/svelte/issues/2588. I think they are gonna add this in the future: https://github.com/pngwn/svelte-adapter/issues/13#issuecomment-1221602073.

partial fix for: https://github.com/microsoft/playwright/issues/15919

### DIFF
--- a/packages/playwright-ct-svelte/index.d.ts
+++ b/packages/playwright-ct-svelte/index.d.ts
@@ -42,22 +42,23 @@ type JsonObject = { [Key in string]?: JsonValue };
 
 type Slot = string | string[];
 
-export interface MountOptions<Component extends SvelteComponent> {
+export interface MountOptions<Component extends SvelteComponent = SvelteComponent> {
   props?: ComponentProps<Component>;
   slots?: Record<string, Slot> & { default?: Slot };
   on?: Record<string, Function>;
   hooksConfig?: JsonObject;
 }
 
-interface MountResult extends Locator {
+interface MountResult<Component extends SvelteComponent> extends Locator {
   unmount(): Promise<void>;
+  rerender(options: Omit<MountOptions<Component>, 'hooksConfig'|'slots'>): Promise<void>
 }
 
 interface ComponentFixtures {
   mount<Component extends SvelteComponent>(
     component: new (...args: any[]) => Component,
     options?: MountOptions<Component>
-  ): Promise<MountResult>;
+  ): Promise<MountResult<Component>>;
 }
 
 export const test: TestType<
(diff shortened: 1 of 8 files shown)


packages/playwright-ct-svelte/index.d.ts
```diff
@@ -42,22 +42,23 @@ type JsonObject = { [Key in string]?: JsonValue };
...
   props?: ComponentProps<Component>;
   slots?: Record<string, Slot> & { default?: Slot };
   on?: Record<string, Function>;
   hooksConfig?: JsonObject;
 }
 
-interface MountResult extends Locator {
+interface MountResult<Component extends SvelteComponent> extends Locator {
   unmount(): Promise<void>;
+  rerender(options: Omit<MountOptions<Component>, 'hooksConfig'|'slots'>): Promise<void>
```
caenguidanos, line 54: The name of the method associated with the update has already been discussed?

I say this because the name `rerender` doesn't convince me. Perhaps something more expressive like `update` would be a better option and improve the intuitive understanding of the API.

```ts
let component = mount(<App></App>, {
   props: {
      count: 0 
   }
});

component.update({
   props: {
      count: 1,
   }
});
```

sand4rt: Hi, thanks for the review! Why doesn't `rerender` convince you?

pavelfeldman: I'm fine with the `update`, it sounds simpler to me as well.

pavelfeldman (approved):

packages/playwright-ct-svelte/index.d.ts
sand4rt: I created a draft PR for it: https://github.com/microsoft/playwright/pull/17972. Shall we discuss it further there?

pooledge:
Hmm, I might have lost track here, has this ever been released/documented? I am on v1.31

sand4rt:
@pooledge yes, it is called `update` instead of `rerender`

pooledge:
> @pooledge yes, it is called `update` instead of `rerender`

Am I supposed to call it alongside `mount` like this:
`test('name', async ({ mount, update }) => {})`?

Because `Error: Test has unknown parameter "update".`

sand4rt:
@pooledge An overview of all API's can be found here: https://github.com/microsoft/playwright/tree/main/tests/components/ct-svelte-vite/tests. Checkout `update.spec.ts` for the `update` API.

PR #17972, opened 2022-10-10
feat(ct): rerender to update

sand4rt:
@pavelfeldman @caenguidanos

Now that this is coming up, i wonder if it wouldn't be better to be a little less flexible and only support `updateProps({ option: 'yes' })` instead of `update({ props: {}, on: {}, slots: {} })`. Updating the slots in Svelte and Solid programaticly is quite tricky.. What do you think?


pavelfeldman:
We are tailoring apis in a framework-specific manner, so I think `update` is a safe bet. React and Vue would have slots, while Svelte and Solid might be only updating the props using the same method name.

pavelfeldman (approved):

PR #17981, opened 2022-10-11
fix: Allow setting process.env.PWDEBUG programatically

eeroan:
Currently setting process.env.PWDEBUG programatically is not possible, because it is read once code loaded. After this change use can set it i.e. before launching browser.

### DIFF
--- a/packages/playwright-core/src/utils/index.ts
+++ b/packages/playwright-core/src/utils/index.ts
@@ -90,8 +90,8 @@ export function isError(obj: any): obj is Error {
   return obj instanceof Error || (obj && Object.getPrototypeOf(obj)?.name === 'Error');
 }
 
-const debugEnv = getFromENV('PWDEBUG') || '';
 export function debugMode() {
+  const debugEnv = getFromENV('PWDEBUG') || '';
   if (debugEnv === 'console')
     return 'console';
   if (debugEnv === '0' || debugEnv === 'false')


eeroan:
Sure.

I had an idea that when debugging mocha tests with I would always get headed browser at the same time, like this:

```
if (require('inspector').url()) {
  process.env.PWDEBUG = 'console'
}
chromium.launch(...)
```

But currently setting `process.env.PWDEBUG`  in code happens too late.

Our project uses only playwright browser API and mocha as test engine.

pavelfeldman:
> when debugging mocha tests with Intellij Idea

I'm sure Intellij Idea has a way of adding env variable for the debug launch configurations. Can you use it instead?

eeroan:
> > when debugging mocha tests with Intellij Idea
> 
> I'm sure Intellij Idea has a way of adding env variable for the debug launch configurations. Can you use it instead?

Probably it would be possible, but every developer would have to do the same configuration by hand. Is there a problem reading the variable inside the execution?

pavelfeldman:
I'd recommend starting with filing a bug for this one and seeing if it gets some traction instead!

eeroan:
is there a drawback having this change? 

Because I can’t see any problems. The performance change is irrelevant. 

Unfortunately there is no way configuring this in my IDE.

PR #17995, opened 2022-10-11
fix(ct): vue2 do not destroy events on update

sand4rt:


pavelfeldman (approved):

pavelfeldman:
A follow-up with the test would be nice!

PR #17998, opened 2022-10-11
fix: override cookie even if new is expired

### LINKED ISSUE #17997: [BUG] Set-Cookie is discarded if the expiry is in the past
**Context:**
- Playwright Version: 1.23.0
- Operating System: Windows
- Node.js version: 16.15.1
- Browser: Chromium

**Code Snippet**

```javascript
import { test, APIRequestContext, expect } from "@playwright/test";

let authCookieName = "authCookieName";
let baseUrl = "";

test(`Logout test`, async ({ playwright }) => {
    let apiContext = <APIRequestContext>await playwright.request.newContext({
        baseURL: baseUrl,
    });

    await apiContext.post(`/login`, { data: {/* Login data */ } });     // Sets the authCookie - Set-Cookie: token=tokenvalue ...

    await apiContext.post(`/logout`);                                   // Should remove the authCookie - Set-Cookie: token=; expires=Thu, 01 Jan 1970 00:00:00 GMT

    let storageState = await apiContext.storageState();
    let authCookie = storageState.cookies.filter(cookie => cookie.name == authCookieName);

    expect(authCookie).toHaveLength(0);                                 // Error: Expected length: 0, Received length: 1
});
```

**Describe the bug**

Our backend sets the cookie when the user logs in. When the user logs out we would like to delete that cookie and to do this the backend sends a response with a header like this:

`Set-Cookie: token=; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT`

In browsers the cookie is properly cleared by this. But with playwright the cookie doesn’t get updated / removed, it stays as it was before.
I think that the correct behavior would be to remove the cookie.
See this example in RFC 6265:

- https://datatracker.ietf.org/doc/html/rfc6265#section-3.1

> Finally, to remove a cookie, the server returns a Set-Cookie header with an expiration date in the past.`

barbaralocsi:
This fixes #17997

### DIFF
--- a/packages/playwright-core/src/server/cookieStore.ts
+++ b/packages/playwright-core/src/server/cookieStore.ts
@@ -83,8 +83,6 @@ export class CookieStore {
   }
 
   private _addCookie(cookie: Cookie) {
-    if (cookie.expired())
-      return;
     let set = this._nameToCookies.get(cookie.name());
     if (!set) {
       set = new Set();
@@ -93,12 +91,11 @@ export class CookieStore {
     CookieStore.pruneExpired(set);
     // https://datatracker.ietf.org/doc/html/rfc6265#section-5.3
     for (const other of set) {
-      if (other.equals(cookie)) {
-        cookie.updateExpiresFrom(other);
+      if (other.equals(cookie))
         set.delete(other);
-      }
     }
     set.add(cookie);
+    CookieStore.pruneExpired(set);
   }
 
   private *_cookiesIterator(): IterableIterator<Cookie> {
--- a/packages/playwright-core/src/server/cookieStore.ts
+++ b/packages/playwright-core/src/server/cookieStore.ts
@@ -88,7 +88,6 @@ export class CookieStore {
       set = new Set();
       this._nameToCookies.set(cookie.name(), set);
     }
-    CookieStore.pruneExpired(set);
     // https://datatracker.ietf.org/doc/html/rfc6265#section-5.3
     for (const other of set) {
       if (other.equals(cookie))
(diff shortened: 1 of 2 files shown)


barbaralocsi:
@microsoft-github-policy-service agree [company="{GenTwo}"]

packages/playwright-core/src/server/cookieStore.ts
```diff
@@ -93,12 +91,11 @@ export class CookieStore {
     CookieStore.pruneExpired(set);
```
yury-s, line 91: Remove this line, it's now done in line 98.

tests/library/global-fetch-cookie.spec.ts
```diff
@@ -206,6 +206,28 @@ it('should store cookie from Set-Cookie header even if it contains equal signs',
   });
 });
 
+it('should override cookie from Set-Cookie header even if it expired', async ({ request, server }) => {
+  server.setRoute('/setcookie.html', (req, res) => {
+    res.setHeader('Set-Cookie', [`a=ok`]);
```
yury-s, line 211: Please also add a test where the cookie is persistent, i.e. has explicit `Max-Age` or `Expires` attribute in the future.

barbaralocsi: I added a test with expiries in the future. This would also fail before the changes, because `cookieStore` used to keep the old  cookie's expiry.

yury-s: Nice! This does cover the change in the behavior with the expiry time.

What I meant though is to add another test where the original cookie had an explicit expiry time and then new cookie is received with expiry time in the past which should delete the cookie from the storage. Maybe you can just extend the newly added test to cover that?

yury-s (approved):

tests/library/global-fetch-cookie.spec.ts
barbaralocsi: Okay, I extended this test with another cookie that initally has an explicit expiry in the future and later the expiry is set to the past.

yury-s (approved):

tests/library/global-fetch-cookie.spec.ts
yury-s: Thanks!

Issue #8379, opened 2021-08-23
[Feature] Add BrowserType.connect support in @playwright/test


archlichking:
I'm trying to follow up https://www.browserstack.com/docs/automate/playwright, is there a way to setup `wsEndpoint` as a configuration in config file or in fixtures?

---

Edit: from the Playwright Team:

```js
import base, { chromium } from '@playwright/test';

// caps is copied from browserstack guide
const caps = {
  browser: 'chrome',
  os: 'osx',
  os_version: 'catalina',
  name: 'My first playwright test',
  build: 'playwright-build-1',
  'browserstack.username': process.env.BROWSERSTACK_USERNAME
  'browserstack.accessKey': process.env.BROWSERSTACK_ACCESS_KEY,
  'client.playwrightVersion': '1.13.1',
};

export const visitorTestFixture = base.extend<{}>({
  browser: async ({ browser }, use, workerInfo) => {
    const vBrowser = await chromium.connect({
      wsEndpoint:
        `wss://cdp.browserstack.com/playwright?caps=` +
        `${encodeURIComponent(JSON.stringify(caps))}`,
    });
    await use(vBrowser);
    await vBrowser.close();
  },
});

```

JoelEinbinder:
You can override the browser fixture to connect via browserless, but we do not currently have simple way to configure a remote browser.

archlichking:
Good to know, can you provide me an example?

unlikelyzero:
@JoelEinbinder given the number of browser-hosting providers out there AND the number of teams migrating from selenium-based frameworks, I think this feature / functionality would be critical for widespread adoption of playwright-test.

https://aerokube.com/moon/
https://www.browserless.io/
https://www.browserstack.com/
https://saucelabs.com/
https://crossbrowsertesting.com/
https://headlesstesting.com/
etc.

karanshah-browserstack:
@mxschmitt Any timelines on when this would be picked up?

AndreyMelnichenko:
When I am using page fixture with browser into Moon 
`
import { BrowserContext, BrowserContextOptions, chromium, devices, Page, test as base } from '@playwright/test';
import { MainPage } from '../../../src/pages/MainPage';

let customCfg: BrowserContextOptions = {
    viewport: { width: 1905, height: 977 },
    locale: 'en-US',
};
let context: BrowserContext;
let page: Page;

export const test = base.extend<{ mainPage: MainPage }>({
[...]

PR #18052, opened 2022-10-13
feat(playwright-test): support `connectOverCDP` option in `connectOptions`

tjenkinson:
Open to suggestions for a better way to do this. We just need to be able to use the CDP protocol to connect to chrome using the test library.

Don't suppose there's a way of achieving this on the current version? I couldn't find any

### DIFF
--- a/packages/playwright-core/src/client/browserType.ts
+++ b/packages/playwright-core/src/client/browserType.ts
@@ -89,7 +89,7 @@ export class BrowserType extends ChannelOwner<channels.BrowserTypeChannel> imple
 
   private async _connectInsteadOfLaunching(): Promise<Browser> {
     const connectOptions = this._defaultConnectOptions!;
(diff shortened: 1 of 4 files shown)


unlikelyzero:
https://github.com/microsoft/playwright/issues/8379 was originally filed for this usecase. This is something we need to be compatible with browserless

dgozman:
@tjenkinson Could you please file an issue or reference an existing one? We prefer to have conversations in the issue so that it's available to everyone searching.

That said, if you'd like to use `connectOverCDP` with `@playwright/test` today, the easiest way would be to [override the `browser` fixture](https://playwright.dev/docs/test-fixtures#overriding-fixtures).

```ts

export const test = base.extend({
  browser: async ({ playwright }, use) => {
    const browser = await playwright.chromium.connectOverCDP('ws://myhost:port');
    await use(browser);
    await browser.close()l
  },
});
```

PR #18132, opened 2022-10-17
fix(tracing): do not serialize buffers into the trace file

### LINKED ISSUE #18126: [BUG] "Tracing stop" is taking a very long time (Longer than 5 minutes)
**Context:**
- Playwright Version: 1.27.0
- Operating System: Windows
- Node.js version: 16.13.0
- Browser: Edge

**Code Snippet**

We are calling the `tracing.stop()` method at the end of our test and it is hanging in `await tracingContext.stop({
          path: `${this.baseLogDirectory}/playwright`,
        });`

```typescript
      this.log("🛑 Stopping Playwright UI Driver");
      const tracingContext = this.playwrightContext?.tracing;
      if (tracingContext) {
        await tracingContext.stop({
          path: `${this.baseLogDirectory}/playwright`,
        });
      }
      await this.browser.close();
      await this.playwrightContext?.close();

      this.log("🛑 Stopped Playwright UI Driver");
```

**Describe the bug**
I ran this with a debugger and see the "hanging" is happening in the `visitSha1s` function in tracing.js of Playwright Core
image

For some reason there is a huge object its trying to iterate over -> 

image

dgozman:
Fixes #18126.

### DIFF
--- a/packages/playwright-core/src/server/trace/recorder/tracing.ts
+++ b/packages/playwright-core/src/server/trace/recorder/tracing.ts
@@ -367,8 +367,8 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
   onEntryFinished(entry: har.Entry) {
     const event: trace.ResourceSnapshotTraceEvent = { type: 'resource-snapshot', snapshot: entry };
     this._appendTraceOperation(async () => {
-      visitSha1s(event, this._state!.networkSha1s);
-      await fs.promises.appendFile(this._state!.networkFile, JSON.stringify(event) + '\n');
+      const visited = visitTraceEvent(event, this._state!.networkSha1s);
+      await fs.promises.appendFile(this._state!.networkFile, JSON.stringify(visited) + '\n');
     });
   }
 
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -452,22 +452,24 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
...
 
-function visitSha1s(object: any, sha1s: Set<string>) {
-  if (Array.isArray(object)) {
-    object.forEach(o => visitSha1s(o, sha1s));
-    return;
-  }
+function visitTraceEvent(object: any, sha1s: Set<string>): any {
+  if (Array.isArray(object))
+    return object.map(o => visitTraceEvent(o, sha1s));
+  if (object instanceof Buffer)
```
yury-s, line 458: Should we also skip it in the parent instead of inserting undefined?

yury-s (approved):

packages/playwright-core/src/server/trace/recorder/tracing.ts
dgozman: Luckily, `JSON.stringify()` does it for us.

PR #18135, opened 2022-10-17
chore: migrate to the internal:text selector

pavelfeldman:


dgozman (approved):

packages/playwright-core/src/server/injected/selectorGenerator.ts
```diff
@@ -205,12 +205,12 @@ function buildTextCandidates(injectedScript: InjectedScript, element: Element, i
...
 
-  const escaped = escapeForTextSelector(text, false, true);
+  const escaped = escapeForTextSelector(text, false);
 
   if (isTargetNode)
-    candidates.push({ engine: 'text', selector: escaped, score: 10 });
+    candidates.push({ engine: 'internal:text', selector: escaped, score: 10 });
 
-  if (escaped === text) {
+  if (!escaped.startsWith('/')) {
```
dgozman, line 213: This check does not guarantee that `has-text("{text}")` is a valid CSS selector now. Perhaps we should `cssEscape(text)` in line 217? Or replace it all with `internal:has="internal:text=..."` like we do with `hasText` option?

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1294,19 +1301,25 @@ function unescape(s: string): string {
...
-  }
-  if (selector.length > 1 && selector[0] === "'" && selector[selector.length - 1] === "'") {
-    selector = unescape(selector.substring(1, selector.length - 1));
+  } else if (internal && selector.length > 1 && selector[0] === '"' && selector[selector.length - 2] === '"' && selector[selector.length - 1] === 'i') {
+    selector = unquote(selector.substring(0, selector.length - 1));
+    strict = false;
+  } else if (internal && selector.length > 1 && selector[0] === '"' && selector[selector.length - 2] === '"' && selector[selector.length - 1] === 's') {
+    selector = unquote(selector.substring(0, selector.length - 1));
+    strict = true;
+  } else if (selector.length > 1 && selector[0] === "'" && selector[selector.length - 1] === "'") {
```
dgozman, line 1321: We do not expect single quotes with `internal`, do we?

```suggestion
  } else if (!internal && selector.length > 1 && selector[0] === "'" && selector[selector.length - 1] === "'") {
```

dgozman (approved):

packages/playwright-core/src/server/injected/injectedScript.ts
pavelfeldman: This change seems to move us from one invalid state to another should internal happen with (')

dgozman (approved):

PR #18152, opened 2022-10-18
fix(ct): setting ctTemplateDir does not work

sand4rt:
closes: https://github.com/microsoft/playwright/issues/18116


pavelfeldman (approved):

packages/playwright-test/src/plugins/vitePlugin.ts
```diff
@@ -291,7 +291,11 @@ function vitePlugin(registerSource: string, relativeTemplateDir: string, buildIn
         return { code, map: { mappings: '' } };
       }
 
-      if (!id.endsWith(`${relativeTemplateDir}/index.ts`) && !id.endsWith(`${relativeTemplateDir}/index.tsx`) && !id.endsWith(`${relativeTemplateDir}/index.js`))
+      const indexTs = path.join(relativeTemplateDir, 'index.ts');
```
pavelfeldman, line 294: If relativeTemplateDir contains forward slashes, this still won't work on Windows I am afraid.

sand4rt: Tested it on Windows and it works. I'm using `path.resolve` and `path.join` to convert all slashes to either `\` on Windows or `/` on POSTFIX before the `endsWith` comparison

pavelfeldman (approved):

PR #18157, opened 2022-10-18
fix(test runner): remove unused type imports during transform

dgozman:
Fixes #18117.


packages/playwright-test/bundles/babel/src/babelBundleImpl.ts
```diff
@@ -64,7 +64,7 @@ export function babelTransform(filename: string, isTypeScript: boolean, isModule
       setPublicClassFields: true,
     },
     presets: [
-      [require('@babel/preset-typescript'), { onlyRemoveTypeImports: true }],
+      [require('@babel/preset-typescript'), { onlyRemoveTypeImports: false }],
```
pavelfeldman, line 67: I would be a default value now?

pavelfeldman (approved):

PR #18160, opened 2022-10-18
feat(runner): project run: "always"

yury-s:
Projects marked with `run: 'always'` are non shard-able and run after failures.

### DIFF
--- a/docs/src/test-api/class-testproject.md
+++ b/docs/src/test-api/class-testproject.md
@@ -105,12 +105,6 @@ const config: PlaywrightTestConfig = {
 export default config;
 ```
 
-## property: TestProject.canShard
-* since: v1.28
-- type: ?<[boolean]>
-
-If set to false and the tests run with --shard command line option, all tests from this project will run in every shard. If not specified, the project can be split between several shards.
-
 ## property: TestProject.expect
 * since: v1.10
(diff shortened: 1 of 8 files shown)


docs/src/test-api/class-testproject.md
```diff
@@ -265,20 +265,20 @@ The maximum number of retry attempts given to failed tests. Learn more about [te
 
 Use [`property: TestConfig.retries`] to change this option for all projects.
 
+## property: TestProject.run
```
dgozman, line 268: - Should you remove `TestProject.canShard` from this file?
- Let's also mention how this property affects sharding.

packages/playwright-test/src/runner.ts
```diff
@@ -448,7 +448,26 @@ export class Runner {
       let sigintWatcher;
 
       let hasWorkerErrors = false;
-      for (const testGroups of concurrentTestGroups) {
+      let previousStageFailed = false;
+      for (let testGroups of concurrentTestGroups) {
+        if (previousStageFailed) {
+          const runAlwaysGroups = [];
```
dgozman, line 454: Let's extract a method `skipTestsNotMarkedAsRunAlways()` or something like that?

dgozman (approved):

PR #18219, opened 2022-10-20
fix(ts): resolve .js to .ts in non-ESM mode too

### LINKED ISSUE #18077: [BUG] Regression in 1.27 for ESM file extensions
**Context:**
- Playwright Version: 1.27.1
- Operating System: macOS
- Node.js version: 18.10
- Browser: N/A
- Extra: N/A

**Code Snippet**

Trying to import `test-utils.ts` in a Playwright test:

```ts
// App.test.ts
import { setupTest } from './test-utils.js';

// Error: Cannot find module './test-utils.js'
```

**Describe the bug**

When trying to import `.ts` files, ESM requires an extension. This was working in version `1.26`. However, when upgrading to `1.27`, the “`Cannot find module`” error occurs. Downgrading back to `1.26` again fixes the problem and the file is imported just fine.

Worth highlighting that `test-utils.ts` (TypeScript) is the extension, not `.js`, but following [convention](https://github.com/microsoft/playwright/issues/13078#issuecomment-1082271101) `.js` is preferred.

Also worth noting that my repo is **not** `"type": "module"` which may be causing an issue (we are using ESM throughout and are using `"module": "ESNext"` in TSConfig).

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/18077

### DIFF
--- a/packages/playwright-test/src/transform.ts
+++ b/packages/playwright-test/src/transform.ts
@@ -97,15 +97,14 @@ const scriptPreprocessor = process.env.PW_TEST_SOURCE_TRANSFORM ?
   require(process.env.PW_TEST_SOURCE_TRANSFORM) : undefined;
 const builtins = new Set(Module.builtinModules);
 
-export function resolveHook(isModule: boolean, filename: string, specifier: string): string | undefined {
+export function resolveHook(filename: string, specifier: string): string | undefined {
   if (builtins.has(specifier))
     return;
-  // In real life, playwright-test is under node_modules, but in the tests it isn't.
-  if (filename.startsWith(kPlaywrightInternalPrefix))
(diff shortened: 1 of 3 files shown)


packages/playwright-test/src/transform.ts
```diff
@@ -100,12 +100,11 @@ const builtins = new Set(Module.builtinModules);
 export function resolveHook(isModule: boolean, filename: string, specifier: string): string | undefined {
   if (builtins.has(specifier))
     return;
-  // In real life, playwright-test is under node_modules, but in the tests it isn't.
```
dgozman, line 103: Do we not need this anymore?

dgozman (approved):

dgozman (approved):

PR #18220, opened 2022-10-20
feat: send SIGTERM to webserver before SIGKILL'ing it.

### LINKED ISSUE #18209: [BUG] Killing webserver process should use SIGTERM on Linux, with SIGKILL as fallback
**Context:**
- Playwright Version: 1.20.0-alpha-feb-23-2022
- Operating System: Linux
- Node.js version: v16.10.0
- Browser: Chromium
- Extra: Using docker-compose to launch webserver

**Config Snippets**

Playwright config:
```javascript
  // ...
  webServer: {
    command: 'docker-compose -f docker/docker-compose.yml up app-for-e2e',
    port: 3238,
    timeout: 10 * 1000,
    reuseExistingServer: !process.env.CI,
  }
};
```

docker-compose.yml:
```yaml
app-for-e2e:
  # ...
  container_name: app-for-e2e
  platform: linux/amd64
  ports:
    - 3238:80
```

**Describe the bug**

We use a Docker container to launch our dev webserver and run Playwright tests against it. [The `docker-compose up` command](https://docs.docker.com/compose/reference/up/) responds to SIGINT and SIGTERM by gracefully terminating the container(s) that it launched, but if it receives SIGKILL then it cannot terminate the containers, and they stay running. At the moment, the WebServer class [does not provide any way to gracefully close the launched process](https://github.com/microsoft/playwright/blob/26d65371195936ff553e0515023e866037b6392a/packages/playwright-test/src/webServer.ts#L64), instead using the `killAndWait` function from processLauncher.ts, [which sends SIGKILL to the process](https://github.com/microsoft/playwright/blob/26d65371195936ff553e0515023e866037b6392a/packages/playwright-core/src/utils/processLauncher.ts#L182) without ever trying SIGTERM.

The result is that after I run `npx playwright test`, the `app-for-e2e` container is still running, because Playwirght never attempted to send SIGTERM to the process and just sent SIGKILL instead. I then have to stop the `app-for-e2e` container manually, since I don't want it running and taking up system resources when I'm not using it.

**Proposed solution**

To solve this bug, I would suggest changing `killProcess` in processLauncher.ts to take a boolean argument `gracefully`, defaulting to `false` for backwards compatibility. If `gracefully` is true, send SIGTERM rather than SIGKILL on Unix-like platforms (Linux and MacOS), or on Windows, run `taskkill` without the `/F` flag. Add the same parameter to `killProcessAndCleanup` and `killAndWait`. Then change the `this._killProcess()` method in the WebServer class to look something like this:

```javascript
this._killProcess = async () => {
  const _hasExited = async () => launchedProcess.exitCode != null;
  const killTimeout = this.config.timeout || 60 * 1000;
  kill(true);
  const cancellationToken = { canceled: false };
  const { timedOut } = (await Promise.race([
    raceAgainstTimeout(() => waitFor(_hasExited, 100, cancellationToken), killTimeout),
    this._processExitedPromise,
  ]));
  cancellationToken.canceled = true;
  if (timedOut)
    kill(false);
};
```

**Additional Info**

This is #12299 reopened, since #12299 was closed without a good reason for closing it. (The reasons given were "It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it", **none** of which are good reasons to close a bug **without discussion in the issue first**. If a bug is going to be closed for any of those reasons, a comment to that effect should be made *in the bug itself* first. Closing bugs with no discussion like that is bad issue-management practice.)

aslushnikov:
We now will send `SIGTERM` to the webserver and wait for the `timeout`
before sending `SIGKILL` to it.

Fixes #18209

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -656,7 +656,7 @@ export default config;
   - `port` ?<[int]> The port that your http server is expected to appear on. It does wait until it accepts connections. Exactly one of `port` or `url` is required.
   - `url` ?<[string]> The url on your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the server is ready to accept connections. Exactly one of `port` or `url` is required.
   - `ignoreHTTPSErrors` ?<[boolean]> Whether to ignore HTTPS errors when fetching the `url`. Defaults to `false`.
-  - `timeout` ?<[int]> How long to wait for the process to start up and be available in milliseconds. Defaults to 60000.
+  - `timeout` ?<[int]> How long to wait for the process to start up and be available in milliseconds. The same timeout is also used to terminate the process. Defaults to 60000.
   - `reuseExistingServer` ?<[boolean]> If true, it will re-use an existing server on the `port` or `url` when available. If no server is running on that `port` or `url`, it will run the command to start a new server. If `false`, it will throw if an existing process is listening on the `port` or `url`. This should be commonly set to `!process.env.CI` to allow the local dev server when running tests locally.
   - `cwd` ?<[string]> Current working directory of the spawned process, defaults to the directory of the configuration file.
   - `env` ?<[Object]<[string], [string]>> Environment variables to set for the command, `process.env` by default.
--- a/packages/playwright-core/src/utils/processLauncher.ts
+++ b/packages/playwright-core/src/utils/processLauncher.ts
@@ -188,24 +188,15 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
     }
   }
 
-  function sendSigTERM() {
+  function sendPosixSIGTERM() {
     if (spawnedProcess.pid && !spawnedProcess.killed && !processClosed) {
       options.log(`[pid=${spawnedProcess.pid}] <will sigterm>`);
-      // Request process to terminate.
       try {
-        if (process.platform === 'win32') {
-          const taskkillProcess = childProcess.spawnSync(`taskkill /pid ${spawnedProcess.pid} /T`, { shell: true });
-          const [stdout, stderr] = [taskkillProcess.stdout.toString(), taskkillProcess.stderr.toString()];
-          if (stdout)
-            options.log(`[pid=${spawnedProcess.pid}] taskkill stdout: ${stdout}`);
-          if (stderr)
-            options.log(`[pid=${spawnedProcess.pid}] taskkill stderr: ${stderr}`);
-        } else {
-          process.kill(-spawnedProcess.pid, 'SIGTERM');
-        }
+        // Send SIGTERM to process tree.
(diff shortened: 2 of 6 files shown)


docs/src/test-api/class-testconfig.md
```diff
@@ -656,7 +656,7 @@ export default config;
   - `port` ?<[int]> The port that your http server is expected to appear on. It does wait until it accepts connections. Exactly one of `port` or `url` is required.
   - `url` ?<[string]> The url on your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the server is ready to accept connections. Exactly one of `port` or `url` is required.
   - `ignoreHTTPSErrors` ?<[boolean]> Whether to ignore HTTPS errors when fetching the `url`. Defaults to `false`.
-  - `timeout` ?<[int]> How long to wait for the process to start up and be available in milliseconds. Defaults to 60000.
+  - `timeout` ?<[int]> How long to wait for the process to start up and be available and terminate, in milliseconds. Defaults to 60000.
```
dgozman, line 659:
```suggestion
  - `timeout` ?<[int]> How long to wait for the process to start up and be available in milliseconds. Defaults to 60000. The same timeout is also used to terminate the process.
```

packages/playwright-test/src/plugins/webServerPlugin.ts
```diff
@@ -72,7 +74,8 @@ export class WebServerPlugin implements TestRunnerPlugin {
   }
 
   public async teardown() {
-    await this._killProcess?.();
+    // Send SIGTERM and wait 10 seconds for it to gracefully close.
```
dgozman, line 77:
```suggestion
    // Send SIGTERM and wait for it to gracefully close.
```

tests/playwright-test/web-server.spec.ts
```diff
@@ -19,6 +19,7 @@ import path from 'path';
 import { test, expect } from './playwright-test-fixtures';
 
 const SIMPLE_SERVER_PATH = path.join(__dirname, 'assets', 'simple-server.js');
+const SIMPLE_SERVER_THAT_IGNORES_SIGTERM_PATH = path.join(__dirname, 'assets', 'simple-server-ignores-sigterm.js');
```
dgozman, line 22: Missing this file in the PR.

packages/playwright-core/src/utils/processLauncher.ts
```diff
@@ -188,6 +188,31 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
     }
   }
 
+  function sendSigTERM() {
+    options.log(`[pid=${spawnedProcess.pid}] <sigterm>`);
```
dgozman, line 192: I'd drop this line, we have either 194 or 212.

dgozman (approved):

Issue #18190, opened 2022-10-18
[BUG]page.request.get() hangs with the timeout parameter


shuguang101:
**Context:**
- Playwright Version: [playwright==1.25.2]
- Operating System: [Mac]
- Python version: [python3.8.13]
- Browser: [Chromium]

**Code Snippet**

```python
from playwright.sync_api import sync_playwright

def run(playwright):
    chromium = playwright.chromium # or "firefox" or "webkit".
    browser = chromium.launch(headless=False, channel="chrome")
    page = browser.new_page()
    page.set_default_timeout(10000)
    page.set_default_navigation_timeout(10000)

    page.goto("https://juejin.cn/post/6844903846766968845")
    page.wait_for_load_state(state="networkidle", timeout=10000)
    # the page.request.get() will hang, but it should timeout
    page.request.get("https://lf3-cdn-tos.bytescm.com/obj/static/xitu_juejin_web//static/favicons/site.webmanifest", timeout=5000, headers={
        "Accept": "*/*",
        "Pragma": "no-cache",
        "Cache-Control": "no-cache",
        "Range": "bytes=0-1"
    })
    
    browser.close()

with sync_playwright() as playwright:
    run(playwright)
```

**Describe the bug**

This url "https://lf3-cdn-tos.bytescm.com/obj/static/xitu_juejin_web//static/favicons/site.webmanifest" does not support range headers, so a request with the "Range: bytes=0-1" header will caused http request hang. Because I set the timeout parameter(timeout=5000), so page.request.get() should timeout after 5s. But in fact, page.request.get() will hang.

mxschmitt:
Looks like response does only emit `close` but not `end` when its a partial response.

cp: https://github.com/microsoft/playwright/blob/9b35a8071f2afbd6fb123bf418b8c734f8daa2e8/packages/playwright-core/src/server/fetch.ts#L368

mxschmitt:
Looks like the server doesn't properly encode range response with Brotli - apparently it cuts selected range from already encoded body whereas it should first cut out the specified range and then encode it. You can see that the response received with curl is also corrupted: `curl -O -v -X GET -H "range: bytes=0-1" -H "Accept-Encoding: br" https://lf3-cdn-tos.bytescm.com/obj/static/xitu_juejin_web//static/favicons/site.webmanifest`
If you don't care about the response encoding you can pass headers:
```
{
    'Accept-Encoding': '',
    'Range': 'bytes=0-1',
}
```
[...]

PR #18223, opened 2022-10-20
fix(fetch): throw on unexpected end of file brotli requests

mxschmitt:
https://github.com/microsoft/playwright/issues/18190

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -360,6 +360,7 @@ export abstract class APIRequestContext extends SdkObject {
             if (e)
               reject(new Error(`failed to decompress '${encoding}' encoding: ${e}`));
           });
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/fetch.ts
```diff
@@ -360,6 +360,7 @@ export abstract class APIRequestContext extends SdkObject {
             if (e)
               reject(new Error(`failed to decompress '${encoding}' encoding: ${e}`));
           });
+          body.on('error', e => reject(new Error(`failed to decompress '${encoding}' encoding: ${e}`)));
```
mxschmitt, line 363: This was required otherwise it ended up in an unhandledException:

```
 Error: unexpected end of file

        at BrotliDecoder.zlibOnError [as onerror] (node:zlib:189:17)
```

yury-s (approved):

Issue #18143, opened 2022-10-18
[BUG] vscode extension fails to list any test if config file contains a console.log


speray:
**Context:**
- Playwright Version: 1.20.1
- Operating System:  Windows
- Node.js version: 14
- Visual Studio Code version: 1.72.2
- Playwright for VSCode extension version: 0.2.13

**Code Snippet**

`console.log('hello')`

**Describe the bug**
In vscode, the playwright extension isn't listing any test in the "Testing" section if the config file executes a console.log

After investigations:
Our playwright config file is containing a console.warn for some internal logic after checking an env variable.
The vscode output window named Log(Window) shows the following:
[2022-10-18 11:52:28.786] [renderer1] [error] [Extension Host] SyntaxError: Unexpected token e in JSON at position 0
	at JSON.parse (<anonymous>)
	at Ke.listFiles (c:\Users\xxxx\.vscode\extensions\ms-playwright.playwright-0.2.13\out\extension.js:14:4435)
	at process.processTicksAndRejections (node:internal/process/task_queues:96:5)
	at async ct.listFiles (c:\Users\xxxx\.vscode\extensions\ms-playwright.playwright-0.2.13\out\extension.js:18:8538)
	at async ht._rebuildModel (c:\Users\xxxx\.vscode\extensions\ms-playwright.playwright-0.2.13\out\extension.js:18:21097)

Looking at [the source of the extension](https://github.com/microsoft/playwright-vscode/blob/1de0859cef3e3370ce8d6100dd687299e4ee3918/src/playwrightTest.ts#L105), it seems that the list-files CLI command is used, and the result is parsed as json.
Running the command `npx playwright list-files` in the CLI indeed prints the message from our console.log, which is preventing the `JSON.parse` to work.

EDIT: Somehow, when creating a minimal project, this doesn't repro. In the small project, the console.log text also shows with the list-files CLI but doesn't bother the vscode extension. I would guess our main project has some other vscode settings which could somehow influence the playwright extension? Any idea where to look at?

EDIT2: On our original project, only console.log breaks the test listing, console.warn is ok...

EDIT3: My minimal repro project was on a different version of playwright 1.23 where it doesn't repro. Using 1.20 is indeed reproeing. 1.21 is the first version where it doesn't repro anymore. So I guess this bug has already been fixed, but I'll let the playwright team confirm.

aslushnikov:
[...]

PR #18227, opened 2022-10-20
chore: output JSON report on a single last line

aslushnikov:
References #18143


aslushnikov:
Turns out extension is already using IPC to list tests, so no need to change JSON format!

PR #18308, opened 2022-10-25
chore: switch to the new debug controller harness

pavelfeldman:


packages/playwright-core/src/cli/cli.ts
```diff
@@ -682,8 +682,8 @@ function logErrorAndExit(e: Error) {
   process.exit(1);
 }
 
-function language(): string {
-  return process.env.PW_LANG_NAME || 'test';
+function codegenId(): string {
+  return process.env.PW_LANG_NAME || 'javascript-test';
```
mxschmitt, line 686: isn't it by definition more TS? Or did we make "import" work in JS well?

mxschmitt (approved):

packages/playwright-core/src/cli/driver.ts
```diff
@@ -55,8 +52,6 @@ export async function runServer(port: number | undefined, path = '/', maxConnect
   process.on('exit', () => server.close().catch(console.error));
   console.log('Listening on ' + wsEndpoint);  // eslint-disable-line no-console
   process.stdin.on('close', () => selfDestruct());
-  if (reuseBrowser && process.send)
```
dgozman, line 58: `reuseBrowser` is unused now?

tests/library/debug-controller.spec.ts
```diff
@@ -153,3 +143,33 @@ test('should highlight all', async ({ backend, connectedBrowser }) => {
...
+  backend.on('sourceChanged', event => events.push(event));
+
+  await backend.setMode({ mode: 'recording' });
+
+  const context = await connectedBrowser._newContextForReuse();
+  const [page] = context.pages();
+
+  await page.locator('body').click();
+
+  await new Promise(f => setTimeout(f, 2000));
```
dgozman, line 158: Don't need this one probably?

dgozman (approved):

packages/playwright-core/src/cli/cli.ts
pavelfeldman: Hm, I did not think it about it. import does work in JS, but I wonder if we should differentiate between JS and TS now...

tests/library/debug-controller.spec.ts
pavelfeldman: Oh yes.

packages/playwright-core/src/cli/cli.ts
mxschmitt: tested with js/ts/js-esm all good, lgtm.

dtinth:
This changes breaks `playwright-ruby-client` as it is sending `sdkLanguage: 'ruby'`

- https://github.com/YusukeIwaki/playwright-ruby-client/issues/228

Maybe the schema should allow experimental languages (perhaps identified by some prefix like `x-`)? Otherwise the alternative is that 3rd party clients have to masquerade as one of (javascript|python|java|csharp).

PR #18314, opened 2022-10-25
docs: fix library-js

### LINKED ISSUE #18145: [Question] `node ./my-script.ts` in the docs, does it work?
On the the [Library page under Guides](https://playwright.dev/docs/library), there is an instruction that tells to run the following code:

```bash
node ./my-script.ts
```

I think `node` can't directly run a TypeScript file, can it?

dgozman:
Fixes #18145.

### DIFF
--- a/docs/src/library-js.md
+++ b/docs/src/library-js.md
@@ -5,17 +5,14 @@ title: "Library"
 
 Playwright Library provides unified APIs for launching and interacting with browsers, while Playwright Test provides all this plus a fully managed end-to-end Test Runner and experience.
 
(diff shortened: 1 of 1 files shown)


docs/src/library-js.md
```diff
@@ -5,30 +5,27 @@ title: "Library"
...
+## Differences when using library
 
 ### Library Example
 
 The following is an example of using the Playwright Library directly to launch Chromium, go to a page, and check its title:
 
 
 ```js tab=js-ts
-import playwright, { devices } from 'playwright';
+import { chromium, devices } from 'playwright';
```
mxschmitt, line 23: whats the reason for this change? Just curious. Since tree-shaking isn't a thing anyway.

mxschmitt (approved):

PR #18323, opened 2022-10-25
fix: make reoccurring networkidle work

### LINKED ISSUE #18283: [BUG] `networkidel` does not work on second or later `goto`
**Context:**
- Playwright Version: 1.27.1
- Operating System: macOS(M1)
- Node.js version: 16.15.0
- Browser: All

```
## System:
 - OS: macOS 12.6
 - Memory: 11.90 GB / 32.00 GB
## Binaries:
 - Node: 16.15.0 - ~/.asdf/installs/nodejs/16.15.0/bin/node
 - npm: 8.5.5 - ~/.asdf/plugins/nodejs/shims/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash
## npmPackages:
 - playwright: ^1.27.1 => 1.27.1
 ```

**Code Snippet**
```javascript
test('`networkidel` does not work on second or later `goto`', async ({ page }) => {
  let sampleApiRequestCount = 0

  await page.route("**/test.html", (route) => {
    route.fulfill({
      status: 200,
      contentType: 'text/html',
      body: `
  <script>
    fetch("http://localhost:8000/sample").then(res => console.log(res.json()))
  </script>
`
    })
  })

  await page.route("**/sample", (route) => {
    sampleApiRequestCount += 1;
    route.fulfill({
      status: 200,
      body: JSON.stringify({
        content: "sample"
      })
    })
  })

  await page.goto("http://localhost:8000/test.html", {
    waitUntil: "networkidle"
  })
  expect(sampleApiRequestCount).toBe(1)

  await page.goto("http://localhost:8000/test.html", {
    waitUntil: "networkidle"
  })
  expect(sampleApiRequestCount).toBe(2)
});

```

**Describe the bug**

Hi, I'm looking for a new library. Thank you for the great library!
When I run the above test, it always fails at the `expect(sampleApiRequestCount).toBe(2)` part.
Also, if I use `goto` a second time to move to a different page than the first, the test fails in the same way.
Is this a bug? Or am I using it wrong? Sorry if this is hard to understand since it is a machine translation.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/18283

### DIFF
--- a/tests/page/page-network-idle.spec.ts
+++ b/tests/page/page-network-idle.spec.ts
@@ -177,7 +177,7 @@ it('should wait for networkidle when iframe attaches and detaches', async ({ pag
   expect(done).toBe(true);
 });
 
-it('should work multiple times after each other', async ({ page, server }) => {
+it('should work after repeated navigations in the same page', async ({ page, server }) => {
   it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/18283' });
 
   let requestCount = 0;
--- a/tests/page/page-network-idle.spec.ts
+++ b/tests/page/page-network-idle.spec.ts
@@ -176,3 +176,33 @@ it('should wait for networkidle when iframe attaches and detaches', async ({ pag
   await promise;
   expect(done).toBe(true);
 });
+
+it('should work multiple times after each other', async ({ page, server }) => {
+  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/18283' });
+
+  let requestCount = 0;
+  await page.route('**/empty.html', route => {
+    route.fulfill({
+      contentType: 'text/html',
+      body: `
+        <script>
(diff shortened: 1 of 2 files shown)


tests/page/page-network-idle.spec.ts
```diff
@@ -176,3 +176,33 @@ it('should wait for networkidle when iframe attaches and detaches', async ({ pag
   await promise;
   expect(done).toBe(true);
 });
+
+it('should work multiple times after each other', async ({ page, server }) => {
```
yury-s, line 180: nit: should work after repeated navigations in the same page ?

yury-s (approved):

tests/page/page-network-idle.spec.ts
mxschmitt, line 180:
```suggestion
it('should work after repeated navigations in the same page', async ({ page, server }) => {
```

PR #18334, opened 2022-10-26
chore: use Node.js 18 for Docker images

mxschmitt:
Node.js 18 is now LTS and we follow LTS schedule.


mxschmitt:
TBR Lusha (merged by accident the wrong PR).

aslushnikov:
LGTM + needs follow-up with bionic

mxschmitt:
Leaving bionic out for now, as per offline discussion, since they also dropped support for it:

#6 350.8 node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)

PR #18352, opened 2022-10-26
docs: update locators doc to use new APIs

dgozman:


docs/src/locators.md
```diff
@@ -71,115 +71,113 @@ await locator.ClickAsync();
...
 
 
 ## Locating elements
 
-Use [`method: Page.locator`] method to create a locator. This method takes a selector that describes how to find an element in the page. The choice of selectors determines the resiliency of the test when the underlying web page changes. To reduce the maintenance burden, we recommend prioritizing user-facing attributes and explicit contracts.
+Playwright comes with multiple built-in ways to create a locator. To make tests resilient, we recommend prioritizing user-facing attributes and explicit contracts, and provide dedicated methods for them, such as [`method: Page.getByText`]. It is often convenient to use the [code generator](./codegen.md) to generate a locator, and then edit it as you'd like.
+
+[`method: Page.locator`] is the most generic way to create a locator that takes a [selector](./selectors.md) describing how to find an element in the page, such as a CSS or XPath selector. That said, we highly recommend to use dedicated methods when you can, and rely on [`method: Page.locator`] as a fallback when more resilient locators are not possible. When in doubt, consider using [code generator](./codegen.md) to generate a locator for you.
```
pavelfeldman, line 126: "is the most generic way" suggests that this is a preferred option.

```suggestion
If you absolutely must use CSS or XPath locators, you can use [`method: Page.locator`] to create a locator that takes a [selector](./selectors.md) describing how to find an element in the page.
```

pavelfeldman (approved):

docs/src/locators.md
```diff
@@ -188,104 +186,98 @@ await page.getByRole('checkbox', { checked: true, name: "Check me" }).check();
...
+page.get_by_test_id('directions').click()
 ```
 
 ```csharp
-await page.Locator("data-test-id=directions").ClickAsync();
+await page.GetByTestId("directions").ClickAsync();
 ```
 
-### Locate by label text
+By default, [`method: Page.getByTestId`] will locate elements baed on the `data-test-id` attribute, but you can configure it in your test config or calling [`method: Selectors.setTestIdAttribute`].
```
debs-obrien, line 244: same as above data-testid or data-test-id

docs/src/locators.md
```diff
@@ -188,104 +186,98 @@ await page.getByRole('checkbox', { checked: true, name: "Check me" }).check();
...
+Note that role locators **do not replace** accessibility audits and conformance tests, but rather give early feedback about the ARIA guidelines.
 
-### Define explicit contract and use `data-test-id=`
+### Define explicit contract and use [`method: Page.getByTestId`]
 
-User-facing attributes like text or accessible name can change frequently. In this case it is convenient to define explicit test ids, for example with a `data-test-id` attribute. Playwright has dedicated support for `id`, `data-test-id`, `data-test` and `data-testid` attributes.
+User-facing attributes like text or accessible name can change over time. In this case it is convenient to define explicit test ids.
 
 ```html
 <button data-test-id="directions">Itinéraire</button>
```
debs-obrien, line 221: our api docs say: By default, the data-testid attribute is used as a test id. so should we be using data-testid or data-test-id` or are both acceptable?

docs/src/locators.md
```diff
@@ -188,104 +186,98 @@ await page.getByRole('checkbox', { checked: true, name: "Check me" }).check();
...
 ```
 
-However, other methods will target the label itself, for example `textContent` will return the text content of the label, not the input field.
-
 ### Locate in a subtree
 
-You can chain [`method: Page.locator`] and [`method: Locator.locator`] calls to narrow down the search to a particular part of the page.
+You can chain methods that create a locator, like [`method: Page.getByText`] or [`method: Locator.getByRole`], to narrow down the search to a particular part of the page.
 
 For example, consider the following DOM structure:
```
debs-obrien, line 283: The example here has data-test-id in the html so if we need to change this to data-testid then keep this in mind.

docs/src/locators.md
```diff
@@ -673,30 +661,30 @@ var texts = await rows.EvaluateAllAsync("list => list.map(element => element.tex
...
 ```csharp
-await page.Locator("data-test-id=product-card").Nth(3).ClickAsync();
+await page.GetByTestId("product-card").Nth(3).ClickAsync();
 ```
 
+However, use these methods with caution. Often times, the page might change, and locator will point to a completely different element from the one you expected. Instead, try to come up with a unique locator that will pass the [strictness criteria](#strictness).
+
 ## Locator vs ElementHandle
 
 :::caution
```
debs-obrien, line 690: suggestion to move locator vs elementHandle to handles doc so that it isnt something people read or find unless they look for it (as discussed with Pavel earlier)

debs-obrien:
doc is great and is huge improvement. i would love to see an overview of all locators  or a list of them like in release notes even just a code block with an example of them all but this is an improvement I can work on later and we can merge this one without it. eg we dont mention getByPlaceholder at all in this doc only in the release notes, same for getByAlt and getByTitle so unless you read the release notes you have no idea these exist

docs/src/locators.md
```diff
@@ -71,115 +71,113 @@ await locator.ClickAsync();
...
 
 
 ## Locating elements
 
-Use [`method: Page.locator`] method to create a locator. This method takes a selector that describes how to find an element in the page. The choice of selectors determines the resiliency of the test when the underlying web page changes. To reduce the maintenance burden, we recommend prioritizing user-facing attributes and explicit contracts.
+Playwright comes with multiple built-in ways to create a locator. To make tests resilient, we recommend prioritizing user-facing attributes and explicit contracts, and provide dedicated methods for them, such as [`method: Page.getByText`]. It is often convenient to use the [code generator](./codegen.md) to generate a locator, and then edit it as you'd like.
+
+If you absolutely must use CSS or XPath locators, you can use [`method: Page.locator`] to create a locator that takes a [selector](./selectors.md) describing how to find an element in the page.
+
+Note that all methods that create a locator, such as [`method: Page.getByLabel`], are also available on the [Locator] and [FrameLocator] classes, so you can chain them and iteratively narrow down your locator.
```
debs-obrien, line 128: would love an example of this as I am not quite sure what it means. isn't getByLabel a locator so ... framelocator ? API docs has a nice example we could possibly use here:
```js
const locator = page.frameLocator('#my-frame').getByText('Submit');
await locator.click();
```

docs/src/locators.md
dgozman: Ouch, my bad!

docs/src/locators.md
dgozman: Sounds good.

Issue #18326, opened 2022-10-25
[BUG] overridden window.eval breaks Playwright evaluate and other methods


bashmish:
**Context:**
- Playwright Version: mine is 1.23.1 (but I think it's any version)
- Operating System: any
- Node.js version: any
- Browser: any
- Extra: website with overridden `window.eval`

**Code Snippet**

```javascript
const { chromium } = require('playwright');

(async () => {
  const browser = await chromium.launch();
  const context = await browser.newContext();
  const page = await context.newPage();
  await page.goto('https://taplink.cc/alognv');
  const title = await page.evaluate(() => {
    return document.title;
  });
  console.log('title', title);
})();
```

```
page.evaluate: TypeError: result2 is not a function
at UtilityScript.evaluate (<anonymous>:180:17)
at UtilityScript.<anonymous> (<anonymous>:1:44)
```

**Describe the bug**

Sometimes you need to control a webpage where `window.eval` is overridden with smth which breaks it, e.g. with a noop method like this:

```javascript
window.eval = () => {};
```

As soon as this code is executed on the page, all Playwright methods which interact with the page's runtime like `page.evaluate`, `page.dispatchEvent` and so on get broken and throw the error `TypeError: result2 is not a function` from `UtilityScript`, because this is not working

```ts
export class UtilityScript {
  evaluate(isFunction: boolean | undefined, returnByValue: boolean, expression: string, argCount: number, ...argsAndHandles: any[]) {
    const args = argsAndHandles.slice(0, argCount);
    const handles = argsAndHandles.slice(argCount);
    const parameters = args.map(a => parseEvaluationResultValue(a, handles));


    // returns null since eval is noop, result is `result2` in the compiled code
    let result = globalThis.eval(expression);


    if (isFunction === true) {
      // throws
      result = result(...parameters);
    } else if (isFunction === false) {
      result = result;
    } else {
      // auto detect.
      if (typeof result === 'function')
        result = result(...parameters);
    }
    return returnByValue ? this._promiseAwareJsonValueNoThrow(result) : result;
  }
```

pavelfeldman:
#### Why was this issue closed?

Thank you for your contribution to our project. This issue has been closed due to its limited upvotes and recent activity, and insufficient feedback for us to effectively act upon. Our priority is to focus on bugs that reflect higher user engagement and have actionable feedback, to ensure our bug database stays manageable.

[...]

PR #18356, opened 2022-10-26
fix: do not use globalThis.eval in utility script

mxschmitt:
https://github.com/microsoft/playwright/issues/18326


bashmish:
I tried a different approach, using `addInitScript` internally to backup `window.eval` in e.g. `window.__pw_eval__` on page load before any user script and then use the backup version in `UtilityScript`, but my knowledge was not sufficient to make it work in all cases, it's very different per browser and recent feature which resets init scripts interferes with that logic in unclear manner to me. I hoped this could be an approach you do, just you know better at which moment to call `addInitScript` to have it working consistently across all browsers and in all corner cases.

PR #18357, opened 2022-10-26
chore: make handle beforeunload when reusing the context

### LINKED ISSUE #17903: [BUG] browser._newContextForReuse: Navigation failed because page was closed
**Context:**
- Playwright Version: 1.24.0 and above
- Operating System: Linux and Mac
- Node.js version: 16
- Browser: All

**Code Snippet**

```
2) [webkit] › components/testComponent/testComponent.test.tsx:17:1 › TestComponent is accessible

Test timeout of 10000ms exceeded while setting up "context".

  Pending operations:
  - browser._newContextForReuse

browser._newContextForReuse: Navigation failed because page was closed!
navigating to "about:blank", waiting until "load"
```

**Describe the bug**
There seem to have been introduced a race condition in version 1.24.0. (This bug is not present in version 1.23.4)

When I run a significant amount of component tests (over 100), some of them fail with the above error. It is always a random test.
If I run the given test in isolation, it works fine.

I have looked through the commits in the 1.24.0 release, and I think that the issue might have been introduced in this PR
https://github.com/microsoft/playwright/pull/15432

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/17903

### DIFF
--- a/packages/playwright-core/src/server/browserContext.ts
+++ b/packages/playwright-core/src/server/browserContext.ts
@@ -180,10 +180,13 @@ export abstract class BrowserContext extends SdkObject {
       page = undefined;
     }
 
-    // Unless I do this early, setting extra http headers below does not respond.
+    // Unless dialogs are dismissed, setting extra http headers below does not respond.
+    page?._frameManager.setCloseAllOpeningDialogs(true);
     await page?._frameManager.closeOpenDialogs();
     // Navigate to about:blank first to ensure no page scripts are running after this point.
     await page?.mainFrame().goto(metadata, 'about:blank', { timeout: 0 });
+    page?._frameManager.setCloseAllOpeningDialogs(false);
+
     await this._resetStorage();
     await this._removeExposedBindings();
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/server/browserContext.ts
```diff
@@ -180,10 +180,13 @@ export abstract class BrowserContext extends SdkObject {
       page = undefined;
     }
 
-    // Unless I do this early, setting extra http headers below does not respond.
+    // Unless dialogs are dismissed, setting extra http headers below does not respond.
+    page?._frameManager.setCloseAllOpeningDialogs(true);
```
dgozman, line 184: This call might as well close all currently open dialogs.

dgozman (approved):

packages/playwright-core/src/server/browserContext.ts
pavelfeldman: Separate semantics -> separate call?

PR #18360, opened 2022-10-26
chore: print where the browsers are downloaded from

### LINKED ISSUE #16926: [BUG] Print the URL browser is downloaded from
Often times browser can't be downloaded due to proxy / firewall / security system on the client machine. Printing the url of the browser would help clients fetching it manually and would help them talk to the security departments while explaining the issue.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/16926

title

### DIFF
--- a/packages/playwright-core/src/server/registry/browserFetcher.ts
+++ b/packages/playwright-core/src/server/registry/browserFetcher.ts
@@ -24,6 +24,7 @@ import { existsAsync } from '../../utils/fileUtils';
 import { debugLogger } from '../../common/debugLogger';
 import { extract } from '../../zipBundle';
 import { ManualPromise } from '../../utils/manualPromise';
+import { colors } from '../../utilsBundle';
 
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):
Make sure that the experience with terminals with a smaller width does not get worse.

aswatsankar:
Hey @pavelfeldman and @mxschmitt  

When I try to download the browser binaries directly using the path mentioned above https://playwright.azureedge.net/builds/chromium/1030/chromium-linux.zip it is downloaded successfully. However when I try to run a simple program, I am still getting the Unable to get local issuer certificate error. 
I have tried setting NODE_EXTRA_CA_CERTS=file but no luck
image

Kindly throw some light on this.

hp77-creator:
@pavelfeldman I have installed driver from browser on my own, where should I keep this driver so that playwright discovers it?

Issue #10825, opened 2021-12-09
[Question] Specific test retries


awnclaim:
Hi!

Is there any way I could set the maximum number of retries to a specific test function?

My case is the following: in my test, I create an object via a POST request and then I want to verify that the response body of a GET request contains some info about it. The matter is that our backend will return no info within 2-4 minutes after the object is created, so I need to send one request after another, retrying this test, to ultimately reach the point when the response contains the info I need. At the same time, I don't want my other tests / test files to be retried X times as well. If they fail, they fail.

Is it possible to make such an adjustment to one test only? Thank you very much!

dgozman:
@awnclaim There is no such feature as of today. As a workaround, you can put this test into a separate file and use a separate project for it:

```js
// playwright.config.ts

const config = {
  projects: [
    {
       name: 'default',
       testIgnore: ['retried-test.spec.ts'],
    },
    {
       name: 'retried',
       testMatch: /retried-test.spec.ts/,
       retries: 2,
    },
  ],
};
```

awnclaim:
Hello,

Thanks for your reply. I would like to express my strong desire to **include this in your backlog**, as in the workaround you suggested all test functions will be retried in case our backend does not respond, and this is not something we would want.

Could I please ask another question? To apply this workaround, I would need to create separate authentication setups for the suits not to impede each other (i. e., make several suits log in under different credentials). Right now, as I understand, only one `globalSetup` is possible. What would you suggest to implement to start all test suits with one command, so that they could be run in parallel with several workers but under different login & email?

dgozman:
> I would like to express my strong desire to include this in your backlog

Of course 👍 

> Right now, as I understand, only one globalSetup is possible. What would you suggest to implement to start all test suits with one command, so that they could be run in parallel with several workers but under different login & email?

Yeah, that's not an easy problem, but we are working on a nice ergonomic solution for this. For now, you'll have to log in twice in the single `globalSetup` and save multiple storage states, as [described here](https://playwright.dev/docs/test-auth#multiple-signed-in-roles).

awnclaim:
Hello,

Thank you so much for drawing my attention to this instruction, I hadn't noticed it before.

I'll stay tuned to the updates from you guys!

dgozman:
Let's keep this issue open to track the "set retries per single test" feature request.

johno-stripe:
[...]

PR #18388, opened 2022-10-27
feat(test runner): test.describe.configure({ retries, timeout })

dgozman:
References #10825.

### DIFF
--- a/docs/src/test-retries-js.md
+++ b/docs/src/test-retries-js.md
@@ -134,6 +134,42 @@ test('my test', async ({ page }, testInfo) => {
 });
 ```
 
+You can specify retries for a specific group of tests or a single file with [`method: Test.describe.configure`].
+
+```js tab=js-js
+const { test, expect } = require('@playwright/test');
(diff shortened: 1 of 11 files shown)


docs/src/test-retries-js.md
```diff
@@ -134,6 +134,42 @@ test('my test', async ({ page }, testInfo) => {
 });
 ```
 
+You can specify retries for a specific group of tests or a single file with [`method: Test.describe.configure`].
+
+```js tab=js-js
+
+test.describe(() => {
+  // All tests in this describe group will get 2 retry attempts.
```
yury-s, line 143: One of the snippets should contain top-level tests instead of test.describe?

yury-s (approved):

docs/src/test-retries-js.md
dgozman: These are js and ts snippets. I decided to show the describe version, because it's more useful based on the filed issues.

PR #18413, opened 2022-10-28
chore: allow not overriding media when creating context

pavelfeldman:


packages/playwright-core/src/client/browserContext.ts
```diff
@@ -416,6 +416,9 @@ export async function prepareBrowserContextParams(options: BrowserContextOptions
     storageState: await prepareStorageState(options),
     serviceWorkers: options.serviceWorkers,
     recordHar: prepareRecordHarOptions(options.recordHar),
+    colorScheme: options.colorScheme ? options.colorScheme : 'no-override',
```
dgozman, line 419: This turns `undefined` into `no-override`, but it should leave our default `light` instead. We should only convert nulls into `no-override`.

packages/playwright-core/src/client/types.ts
```diff
@@ -47,32 +47,36 @@ export type SetStorageState = {
...
+  };
   recordHar?: {
     path: string,
     omitContent?: boolean,
     content?: 'omit' | 'embed' | 'attach',
     mode?: 'full' | 'minimal',
     urlFilter?: string | RegExp,
-  },
+  };
+  colorScheme?: 'dark' | 'light' | 'no-preference' | null,
```
dgozman, line 69: nit: why change all commas to semicolons, just to introduce more commas? 😄

packages/playwright-core/src/server/webkit/wkPage.ts
```diff
@@ -627,27 +627,34 @@ export class WKPage implements PageDelegate {
...
-  private static async _setEmulateMedia(session: WKSession, mediaType: types.MediaType | null, colorScheme: types.ColorScheme | null, reducedMotion: types.ReducedMotion | null, forcedColors: types.ForcedColors | null): Promise<void> {
+  private static async _setEmulateMedia(session: WKSession, mediaType: types.MediaType, colorScheme: types.ColorScheme, reducedMotion: types.ReducedMotion, forcedColors: types.ForcedColors): Promise<void> {
     const promises = [];
-    promises.push(session.send('Page.setEmulatedMedia', { media: mediaType || '' }));
+    promises.push(session.send('Page.setEmulatedMedia', { media: mediaType === 'no-override' ? '' : mediaType }));
     let appearance: any = undefined;
-    switch (colorScheme) {
-      case 'light': appearance = 'Light'; break;
-      case 'dark': appearance = 'Dark'; break;
+    if (colorScheme !== 'no-override') {
```
dgozman, line 634: If I had an override, and then call `page.emulateMedia({ colorScheme: null })`, it will be ignored here. Same for other browsers maybe?

packages/playwright-test/types/test.d.ts
```diff
@@ -2855,11 +2855,6 @@ export interface PlaywrightTestOptions {
    * Toggles bypassing page's Content-Security-Policy.
    */
   bypassCSP: boolean | undefined;
-  /**
```
dgozman, line 2858: Why did this disappear?

tests/page/page-emulate-media.spec.ts
```diff
@@ -64,7 +64,7 @@ it('should throw in case of bad colorScheme argument', async ({ page }) => {
   let error = null;
   // @ts-expect-error 'bad' is not a valid media type
   await page.emulateMedia({ colorScheme: 'bad' }).catch(e => error = e);
-  expect(error.message).toContain('colorScheme: expected one of (dark|light|no-preference|null)');
+  expect(error.message).toContain('colorScheme: expected one of (dark|light|no-preference|no-override)');
```
dgozman, line 67: This is actually wrong now, because user never passes `no-override`, they pass `null`.

pavelfeldman: Yeah, but they also never see this message because of the type checks.

dgozman (approved):

PR #18429, opened 2022-10-30
chore: add explore locator parser

pavelfeldman:


dgozman (approved):
From offline discussion:
- probably can support `.locator('foo', { hasText: 'bar' })`;
- probably can `JSON.parse()` the quoted string to support `\u0020` and similar escapes;
- probably can call `asLocator()` and compare to the original to be 100% sure.

PR #18440, opened 2022-10-31
feat(testinfo): add name to attachment output name

Yexiaoxing:
Per discussion in #12950, adding sanitized name to the output filename prefix. This can make debugging easier, and the filename structure more human friendly,

### DIFF
--- a/packages/playwright-test/src/util.ts
+++ b/packages/playwright-test/src/util.ts
@@ -21,7 +21,7 @@ import path from 'path';
 import url from 'url';
 import { colors, debug, minimatch } from 'playwright-core/lib/utilsBundle';
 import type { TestError, Location } from './types';
-import { calculateSha1, isRegExp } from 'playwright-core/lib/utils';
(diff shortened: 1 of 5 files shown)


packages/playwright-test/src/util.ts
```diff
@@ -287,7 +287,10 @@ export async function normalizeAndSaveAttachment(outputPath: string, name: strin
     throw new Error(`Exactly one of "path" and "body" must be specified`);
   if (options.path !== undefined) {
     const hash = calculateSha1(options.path);
-    const dest = path.join(outputPath, 'attachments', hash + path.extname(options.path));
+    let sanitizedNamePrefix = '';
+    if (isString(name))
```
dgozman, line 291: We should not ignore non-strings, but rather throw an error.

Yexiaoxing: Done.

dgozman (commented):
Looking good, merging in.

dgozman (approved):

PR #18449, opened 2022-10-31
chore: compress trace-viewer images

Fdawgs:
Losslessly compress images using [Pngcrush](https://pmt.sourceforge.io/pngcrush/). Should reduce the build size by a tiny amount.


pavelfeldman:
We seem to be only saving 26K of the multi-megabyte NPM package that downloads a gigabyte of the browsers. Reducing the package size is not a priority at the moment, and when it becomes a priority, we'll probably start with analyzing where the footprint is coming from and start slicing larger chunks.

PR #18461, opened 2022-10-31
feat(ct): navigate with page.goto

### LINKED ISSUE #15889: [Feature] Component testing and page.goto

sand4rt:
closes: https://github.com/microsoft/playwright/issues/15889

### DIFF
--- a/packages/playwright-test/src/mount.ts
+++ b/packages/playwright-test/src/mount.ts
@@ -48,23 +48,30 @@ export const fixtures: Fixtures<
     },
 
     mount: async ({ page }, use) => {
-      await use(async (component: JsxComponent | string, options?: MountOptions) => {
-        const selector = await (page as any)._wrapApiCall(async () => {
-          return await innerMount(page, component, options);
-        }, true);
-        const locator = page.locator(selector);
-        return Object.assign(locator, {
-          unmount: async () => {
-            await locator.evaluate(async () => {
-              const rootElement = document.getElementById('root')!;
-              await window.playwrightUnmount(rootElement);
-            });
-          },
-          update: async (options: JsxComponent | Omit<MountOptions, 'hooksConfig'>) => {
-            if (isJsxApi(options)) return await innerUpdate(page, options);
-            await innerUpdate(page, component, options);
-          }
+      await use((component: JsxComponent | string, options?: MountOptions) => {
+        async function mount() {
+          const selector = await (page as any)._wrapApiCall(async () => {
+            return await innerMount(page, component, options);
+          }, true);
+          const locator = page.locator(selector);
+          return Object.assign(locator, {
+            unmount: async () => {
+              await locator.evaluate(async () => {
+                const rootElement = document.getElementById('root')!;
+                await window.playwrightUnmount(rootElement);
+              });
+            },
+            update: async (options: JsxComponent | Omit<MountOptions, 'hooksConfig'>) => {
+              if (isJsxApi(options)) return await innerUpdate(page, options);
+              await innerUpdate(page, component, options);
+            }
+          });
+        }
+        page.addListener('load', async page => {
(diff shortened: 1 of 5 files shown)


packages/playwright-test/src/mount.ts
```diff
@@ -48,23 +48,27 @@ export const fixtures: Fixtures<
...
+                await window.playwrightUnmount(rootElement);
+              });
+            },
+            update: async (options: JsxComponent | Omit<MountOptions, 'hooksConfig'>) => {
+              if (isJsxApi(options)) return await innerUpdate(page, options);
+              await innerUpdate(page, component, options);
+            }
+          });
+        }
+        page.addListener('load', mount);
```
pavelfeldman, line 70: This change suggests that it is Ok to navigate in the component tests. But in reality, only 'good' navigations to a magic URL will result in the harness result (the one with window.playwrightMount available. If we want to support it, we should ensure that the page navigated to any URL allows mounting?

sand4rt: > If we want to support

I have some router logic which i think is useful to test with `page.goto` and `page.refresh`: https://github.com/sand4rt/vue-router-test/blob/master/src/router/index.ts#L6

Also wrote some tests for this logic: https://github.com/sand4rt/vue-router-test/blob/master/src/notation-vue.test.ts#L4

Wonder what you think and if there is a better way to test this logic?

> We should ensure that the page navigated to any URL allows mounting?

When testing components, navigating to pages that don't start with the base URL `process.env.PLAYWRIGHT_TEST_BASE_URL` seems not useful to me. A URL that starts with the base URL but doesn't exist/is not registered in [`router/index.ts`](https://github.com/sand4rt/vue-router-test/blob/master/src/router/index.ts#L8), like: `http://localhost:3100/page-that-does-not-exists` is in my opinion useful to keep.

tests/components/ct-vue-vite/src/notation-jsx.spec.tsx
```diff
@@ -143,11 +143,20 @@ test('get textContent of the empty template', async ({ mount }) => {
...
   await component.getByRole('link', { name: 'Dashboard' }).click();
   await expect(component.getByRole('main')).toHaveText('Dashboard');
   await expect(page).toHaveURL('/dashboard');
 });
+
+test('navigate to a page with page.goto', async ({ page, mount }) => {
+  const component = await mount(<App />);
+  await expect(component.getByRole('main')).toHaveText('Login');
+  await expect(page).toHaveURL('/');
+  await page.goto('/dashboard');
```
pavelfeldman, line 159: I'm not sure I understand why App is mounted in /dashboard after the navigation, this looks confusing to me.

pavelfeldman:
Let me push back to this one until I understand the use case better.

PR #18466, opened 2022-10-31
chore: download android driver from the cdn

pavelfeldman:


packages/playwright-core/src/server/android/android.ts
```diff
@@ -186,8 +187,13 @@ export class AndroidDevice extends SdkObject {
       await this.shell(`cmd package uninstall com.microsoft.playwright.androiddriver.test`);
 
       debug('pw:android')('Installing the new driver');
-      for (const file of ['android-driver.apk', 'android-driver-target.apk'])
-        await this.installApk(await fs.promises.readFile(require.resolve(`../../../bin/${file}`)));
+      const executable = registry.findExecutable('android')!;
+      for (const file of ['android-driver.apk', 'android-driver-target.apk']) {
+        const fullName = path.join(executable.directory!, file);
+        if (!fs.existsSync(fullName))
+          throw new Error('Please install Android driver apk using `npx playwright install android`');
```
dgozman, line 194:
```suggestion
          throw new Error('Please install Android driver apk using `${buildPlaywrightCLICommand(sdkLanguage, 'install android')}`');
```

packages/playwright-core/src/server/registry/index.ts
```diff
@@ -743,7 +761,8 @@ export class Registry {
   }
 
   private _downloadURLs(descriptor: BrowsersJSONDescriptor): string[] {
-    const downloadPathTemplate: string|undefined = (DOWNLOAD_PATHS as any)[descriptor.name][hostPlatform];
+    const paths = (DOWNLOAD_PATHS as any)[descriptor.name];
+    const downloadPathTemplate: string|undefined = paths[hostPlatform] || paths['<unknown>'];
```
dgozman, line 765: I think this has a side-effect when downloading browsers, probably for unknown linux versions?

tests/android/playwright.config.ts
```diff
@@ -67,6 +67,7 @@ config.projects.push({
     browserName: 'chromium',
   },
   testDir: path.join(testDir, 'page'),
+  testMatch: /page-check/,
```
mxschmitt: revert that

packages/playwright-core/src/server/registry/index.ts
pavelfeldman: I don't think so?

packages/playwright-core/src/server/android/android.ts
pavelfeldman: I don't think your suggestion compiles.

mxschmitt (approved):
lgtm but unknown -> default/fallback

PR #18484, opened 2022-11-01
chore(ct): refactor tests

sand4rt:
Changed:

- grouped test with `test.describe`
- use dark mode
- replaced `locators` with `getBy`
- code formatting


pavelfeldman:
This change pretty much  destroys the blame history. We don't typically land something like this unless it adds a lot of value. In this case, locators -> getBy is approaching this criteria.

sand4rt:
ah alright, that's fine

Issue #15620, opened 2022-07-13
[BUG] Component tests. Cannot compile test if component passed as property


pashidlos:
**Context:**
  "@playwright/experimental-ct-react": "^1.22.2",

**Code Snippet**
https://github.com/pashidlos/pw-component-test-example/tree/component-as-prop

**Describe the bug**
Getting error trying to pass icon as property in component 
Locally made icons are throwing different error than the ones from external libraries (ex `@mui/icons-material`)

## Local icon error
`yarn test-ct1`
```
Running 0 test using 0 worker


/pw-component-test-example/src/DisableIcon.tsx:14
const DisableIconOriginal = (0, _material.createSvgIcon)(<>
                                                         ^

SyntaxError: Unexpected token '<'

    at wrapSafe (internal/modules/cjs/loader.js:984:16)
    at Module._compile (internal/modules/cjs/loader.js:1032:27)
    at Module.A._compile (/Users/pstrunkin/work/pw-component-test-example/node_modules/@playwright/test/lib/utilsBundleImpl.js:16:994)
    at Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
    at Object.t.<computed>.tu._extensions.<computed> [as .tsx] (/Users/pstrunkin/work/pw-component-test-example/node_modules/@playwright/test/lib/utilsBundleImpl.js:16:1010)
    at Module.load (internal/modules/cjs/loader.js:933:32)
    at Function.Module._load (internal/modules/cjs/loader.js:774:14)
    at Module.require (internal/modules/cjs/loader.js:957:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/Users/pstrunkin/work/pw-component-test-example/src/TestComponent.pw-spec.tsx:3:1)
    at Module._compile (internal/modules/cjs/loader.js:1068:30)
    at Module.A._compile (/Users/pstrunkin/work/pw-component-test-example/node_modules/@playwright/test/lib/utilsBundleImpl.js:16:994)
    at Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
    at Object.t.<computed>.tu._extensions.<computed> [as .tsx] (/Users/pstrunkin/work/pw-component-test-example/node_modules/@playwright/test/lib/utilsBundleImpl.js:16:1010)
    at Module.load (internal/modules/cjs/loader.js:933:32)

=================
 no tests found.
```

## External icon error
`yarn test-ct2`
```
[...]

PR #18505, opened 2022-11-02
fix(mount): ensure unwrapped functions return their value when called

darrylblake:
If a callback prop returned a value you should be able to access it. We have components that fail because they expect to receive a promise when invoking one of the callback props (probably not best practice).


pavelfeldman:
This won't work for sync functions, while the users' expectations will now be that those functions can return values.

darrylblake:
Oh right, I didn't realise a promise was always returned - I was only testing with our components that expects one of its props to return a promise.

I suppose this related to this issue too then: https://github.com/microsoft/playwright/issues/15620, React's complaining as it can't render a promise?

pavelfeldman:
> I suppose this related to this issue too then: https://github.com/microsoft/playwright/issues/15620, React's complaining as it can't render a promise?

Yes. I'll close this PR for now!

PR #18560, opened 2022-11-03
Fix typo: "it's" -> "its"

mwiemer-microsoft:


pavelfeldman:
We can consolidate this into one of the larger fixes. It does not make sense to run 50 bots with all the checks and tests over this fix.

mwiemer-microsoft:
Are there ways to exclude `.md` files from code-related checks? Is work being tracked for that side of things? I'd be happy to contribute there

PR #18564, opened 2022-11-04
fix(webserver): do not spawn webserver as a new process group

aslushnikov:
This patch stops using `processLauncher` to launch web servers. Process Launcher
will spawn a new process group which is separate from test runner.

This might result in unexpected behavior, e.g. `kill -sigkill -<testrunnerpid>` will orphan web server process.

Instead, this patch simply spawns web server and sends `SIGTERM` to it
when ready.

### DIFF
--- a/packages/playwright-test/src/plugins/webServerPlugin.ts
+++ b/packages/playwright-test/src/plugins/webServerPlugin.ts
@@ -17,7 +17,7 @@ import http from 'http';
 import https from 'https';
 import path from 'path';
 import net from 'net';
-import { spawn, execSync, spawnSync } from 'child_process';
+import { spawn, spawnSync } from 'child_process';
 import type { ChildProcess } from 'child_process';
 
 import { debug } from 'playwright-core/lib/utilsBundle';
@@ -47,7 +47,7 @@ const debugWebServer = debug('pw:webserver');
 
 export class WebServerPlugin implements TestRunnerPlugin {
   private _isAvailable?: () => Promise<boolean>;
-  private _processExitedPromise?: Promise<void>;
+  private _processExitedPromise?: Promise<{ code: number|null, signal: NodeJS.Signals|null }>;
   private _childProcess?: ChildProcess;
   private _options: WebServerPluginOptions;
   private _checkPortOnly: boolean;
@@ -115,12 +115,7 @@ export class WebServerPlugin implements TestRunnerPlugin {
       shell: true,
     });
     this._processExitedPromise = new Promise((resolve, reject) => {
-      this._childProcess!.once('exit', (code, signal) => {
-        if (code)
-          reject(new Error(`Process from config.webServer terminated with exit code "${code}" and signal "${signal}"`));
-        else
-          resolve(undefined);
-      });
+      this._childProcess!.once('exit', (code, signal) => resolve({ code, signal }));
     });
 
     debugWebServer(`Process started`);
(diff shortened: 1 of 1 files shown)


pavelfeldman:
Let's revert separately and then carefully look at the actual bug and the right fix.

packages/playwright-test/src/plugins/webServerPlugin.ts
```diff
@@ -46,17 +47,15 @@ const debugWebServer = debug('pw:webserver');
 
 export class WebServerPlugin implements TestRunnerPlugin {
   private _isAvailable?: () => Promise<boolean>;
-  private _killProcess?: (presendSigtermBeforeSigkillTimeout?: number) => Promise<void>;
-  private _processExitedPromise!: Promise<any>;
+  private _processExitedPromise?: Promise<any>;
```
yury-s, line 50: Can it have a more specific type than `any`?

packages/playwright-test/src/plugins/webServerPlugin.ts
```diff
@@ -66,22 +65,27 @@ export class WebServerPlugin implements TestRunnerPlugin {
...
   public async teardown() {
-    // Send SIGTERM and wait for it to gracefully close.
-    await this._killProcess?.(this._launchTerminateTimeout);
+    if (!this._childProcess || !this._childProcess.pid || this._childProcess.killed)
+      return;
+    try {
+      if (process.platform === 'win32')
+        execSync(`taskkill /pid ${this._childProcess.pid} /T /F /FI "MEMUSAGE gt 0"`, { stdio: 'ignore' });
+      else
+        process.kill(this._childProcess.pid, 'SIGTERM');
```
yury-s, line 81: Let's keep the comment about graceful termination ?

packages/playwright-test/src/plugins/webServerPlugin.ts
```diff
@@ -92,47 +96,48 @@ export class WebServerPlugin implements TestRunnerPlugin {
...
-      attemptToGracefullyClose: async () => {},
-      log: () => {},
-      onExit: code => processExitedReject(new Error(code ? `Process from config.webServer was not able to start. Exit code: ${code}` : 'Process from config.webServer exited early.')),
-      tempDirectories: [],
     });
-    this._killProcess = kill;
+    this._processExitedPromise = new Promise((resolve, reject) => {
+      this._childProcess!.once('exit', code => {
+        if (code)
+          reject(new Error(`Process from config.webServer was not able to start. Exit code: ${code}`));
```
yury-s, line 113: I [believe](https://nodejs.org/api/child_process.html#event-exit) the exit code will be `null` when the process was killed by a signal. Should we check for both here?

tests/playwright-test/web-server.spec.ts
```diff
@@ -602,27 +601,3 @@ test('should treat 3XX as available server', async ({ runInlineTest }, { workerI
...
-  const port = workerIndex + 10500;
-  const result = await runInlineTest({
-    'test.spec.ts': `
-      const { test } = pwt;
-      test('pass', async ({}) => {});
-    `,
-    'playwright.config.ts': `
-      module.exports = {
-        webServer: {
-          command: 'node ${JSON.stringify(SIMPLE_SERVER_THAT_IGNORES_SIGTERM_PATH)} ${port}',
```
yury-s, line 617: Should we replace it with a test that checks graceful termination in response to sigterm or is it covered elsewhere already?

yury-s (approved):

tests/playwright-test/web-server.spec.ts
aslushnikov: All other tests check it!

PR #18569, opened 2022-11-04
feat(containers): implement reverse proxy for novnc and server

aslushnikov:
This patch adds a reverse proxy in front of novnc and playwright
server inside the container.

As a result:
- container exposes a single HTTP port to the host
- all Playwright clients can connect using this exposed port, e.g.
  `http://127.0.0.1:5400`
- navigating to the exposed port in the browser lands on a nice HTML page
  with a link to view container screen
- users can also manually navigate to `http://127.0.0.1:5400/screen` to
  view screen

### DIFF
--- a/packages/playwright-core/bin/container_novnc_proxy.js
+++ b/packages/playwright-core/bin/container_novnc_proxy.js
@@ -1,0 +1,69 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const http = require('http');
+const fs = require('fs');
+const path = require('path');
+const { debug, program } = require('../lib/utilsBundle');
+const { ProxyServer } = require('../lib/third_party/http_proxy');
+
+const debugLog = debug('pw:proxy');
+
+program
+    .command('start')
+    .description('reverse proxy for novnc and playwright server')
+    .option('--port <number>', 'port number')
+    .option('--server-endpoint <url>', 'Playwright Server endpoint')
+    .option('--novnc-endpoint <url>', 'novnc server endpoint')
+    .option('--novnc-ws-path <string>', 'novnc websocket path')
+    .action(async function(options) {
+      launchReverseProxy(options.port, options.serverEndpoint, options.novncEndpoint, options.novncWsPath);
+    });
+
+program.parse(process.argv);
+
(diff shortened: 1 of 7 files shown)


packages/playwright-core/bundles/utils/package.json
```diff
@@ -13,6 +13,7 @@
     "commander": "8.3.0",
     "debug": "^4.3.4",
     "glob": "^7.1.3",
+    "http-proxy": "^1.18.1",
```
dgozman, line 16: Can we only install this in the container? Playwright does not use it.

packages/playwright-core/bin/container_novnc_proxy.js
```diff
@@ -0,0 +1,103 @@
...
+      }).end(`
+        <style>${CSS_TYPOGRAPHY}</style>
+        <body style='display: flex; align-items: center; justify-content: center; flex-direction: column;'>
+          <h1>Playwright Container</h1>
+          <a href="/screen">View Screen</a>
+        </body>
+      `);
+    } else if ((request.url === '/screen' || request.url === '/screen/') && request.method === 'GET') {
+      response.writeHead(307, {
+        Location: `/screen/?resize=scale&autoconnect=1&path=${novncWSPath}`,
```
dgozman, line 84: I want to understand whether this is still secure with the unguessable vnc path, or not :)

packages/playwright-core/bin/container_novnc_proxy.js
```diff
@@ -0,0 +1,103 @@
...
+
+async function launchReverseProxy(port, serverEndpoint, novncEndpoint, novncWSPath) {
+  const vncProxy = httpProxy.createProxyServer({ ws: true, target: novncEndpoint });
+  const serverProxy = httpProxy.createProxyServer({ ws: true, target: serverEndpoint });
+
+  const httpServer = http.createServer((request, response) => {
+    if (request.url === '/' && request.method === 'GET') {
+      response.writeHead(200, {
+        'content-type': 'text/html',
+      }).end(`
```
dgozman, line 75: nit: I'd extract the whole landing page as a constant above. Might as well `fs.readFileSync()` it at the start.

packages/playwright-core/bin/container_novnc_proxy.js
```diff
@@ -0,0 +1,103 @@
...
+  });
+  httpServer.on('error', error => debugLog(error));
+  httpServer.on('upgrade', (request, socket, head) => {
+    if (request.url === '/' + novncWSPath)
+      vncProxy.ws(request, socket, head);
+    else
+      serverProxy.ws(request, socket, head);
+  });
+  httpServer.listen(port, () => {
+    console.log('Reverse proxy listening on', port);
```
dgozman, line 101: Let's make it more user-friendly: `Playwright Container listening on`.

packages/playwright-core/bundles/utils/package.json
aslushnikov: As discussed offline with @dgozman, there are two options: either have all of container story inside of playwright-core, or extract it into a separate package, e.g. @playwright/container

Originally, we have docker implemented in playwright-test, but moved it to the playwright-core so that language ports would be able to benefit from it.

So having this as part of playwright-core is an intentionatl design decision.

packages/playwright-core/bin/container_novnc_proxy.js
aslushnikov: We played with this offline and seems to be secure!

dgozman (approved):

PR #18587, opened 2022-11-04
feat(runner): TestOptions.storageStateName

yury-s:


packages/playwright-test/src/index.ts
```diff
@@ -218,13 +220,12 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
       options.permissions = permissions;
     if (proxy !== undefined)
       options.proxy = proxy;
-    if (storageState !== undefined) {
+    if (storageStateName !== undefined) {
+      const value = await test.info().storage().get(storageStateName);
+      if (value)
```
dgozman, line 225: Let's throw otherwise? Now, with an explicit option, we can be strict to avoid surprises.

packages/playwright-test/src/loader.ts
```diff
@@ -642,6 +642,8 @@ function validateProject(file: string, project: Project, title: string) {
   if ('use' in project && project.use !== undefined) {
     if (!project.use || typeof project.use !== 'object')
       throw errorWithFile(file, `${title}.use must be an object`);
+    if ((project.use as any).storageState !== undefined && (project.use as any).storageStateName !== undefined)
```
dgozman, line 645: We should not have this check in the `loader`, but rather in the fixtures themselves. However, I'd just prioritize one over the other and say that in the docs.

dgozman (approved):

PR #18589, opened 2022-11-05
docs(python): Remove not-needed semicolons from codes

attakei:
There are semicolons in some python codes in docs that is not needed to run.

This PR is to remove these.


attakei:

mxschmitt (approved):

PR #18595, opened 2022-11-05
chore: expose new locator apis on window.playwright

pavelfeldman:


packages/playwright-core/src/server/injected/consoleApi.ts
```diff
@@ -14,40 +14,54 @@
...
+    self.getByLabel = (text: string | RegExp, options?: { exact?: boolean }): Locator => self.locator(getByLabelSelector(text, options));
+    self.getByPlaceholder = (text: string | RegExp, options?: { exact?: boolean }): Locator => self.locator(getByPlaceholderSelector(text, options));
+    self.getByText = (text: string | RegExp, options?: { exact?: boolean }): Locator => self.locator(getByTextSelector(text, options));
+    self.getByTitle = (text: string | RegExp, options?: { exact?: boolean }): Locator => self.locator(getByTitleSelector(text, options));
+    self.getByRole = (role: string, options: ByRoleOptions = {}): Locator => self.locator(getByRoleSelector(role, options));
   }
-  return new Locator(initial, options);
 }
 
 type ConsoleAPIInterface = {
```
dgozman, line 59: This type is now missing `getBy...` and `locator` methods. Do we even need it?

dgozman (approved):

packages/playwright-core/src/server/injected/consoleApi.ts
pavelfeldman: I don't think we need it, but it does not seem to hurt either.

PR #18596, opened 2022-11-06
feat(ct): vue2 plugins

sand4rt:
There is no support for Vue2 plugins _(probably because there was no test yet describing/documenting this behaviour)_ and did some minor refactoring.


packages/playwright-ct-vue2/hooks.d.ts
```diff
@@ -14,14 +14,20 @@
...
-import { CombinedVueInstance, Vue } from 'vue/types/vue';
-
+import { ComponentOptions } from 'vue';
+import { CombinedVueInstance, Vue, VueConstructor } from 'vue/types/vue';
+ 
 type JsonPrimitive = string | number | boolean | null;
 type JsonValue = JsonPrimitive | JsonObject | JsonArray;
 type JsonArray = JsonValue[];
 type JsonObject = { [Key in string]?: JsonValue };
+type VueOptions = ComponentOptions<Vue> & { router?: unknown, store?: unknown, [key: string]: unknown };
```
pavelfeldman, line 24: Why do we special-case `router?: unknown, store?: unknown, [key: string]: unknown` here?

packages/playwright-ct-vue2/hooks.d.ts
```diff
@@ -14,14 +14,20 @@
...
+
 export declare function beforeMount<HooksConfig extends JsonObject>(
-  callback: (params: { hooksConfig: HooksConfig }) => Promise<void>
+  callback: (params: { 
+    hooksConfig: HooksConfig, 
+    Vue: VueConstructor<Vue>, 
+    options: VueOptions }) => Promise<void>
```
pavelfeldman, line 30: `componentOptions` to be consistent with the Vue types?

pavelfeldman (approved):

packages/playwright-ct-vue2/hooks.d.ts
sand4rt: Ah, i think i forgot to delete that type. BTW i can either return the `VueOptions` or update it directly:

```tsx
beforeMount(async ({ Vue, options, hooksConfig }) => {
  Vue.use(Router);
  options.router = router;
});
```

vs

```tsx
beforeMount(async ({ Vue, hooksConfig }) => {
  Vue.use(Router);
  return { router }
});
```

I think i have a slight preference for returning `VueOptions` as updating props directly feels a kind of odd. Also the return is a bit more in line with:
```tsx
beforeMount(async ({ App }) => {
  return <Router><App /></Router>
});
```

Which one do you prefer?

packages/playwright-ct-vue2/hooks.d.ts
sand4rt: Fixed

pastelsky:
@sand4rt Looks like you have a failing build — might be a flake though. Do you mind re-running?

sand4rt:
> @sand4rt Looks like you have a failing build — might be a flake though. Do you mind re-running?

I don't think that was related to my changes but sure

pastelsky:
I know — was probably flake. Just giving it a gentle nudge since this seems like a pre-req for https://github.com/microsoft/playwright/pull/18616

pavelfeldman (approved):

PR #18631, opened 2022-11-08
chore: use provided value for the generated test id

pavelfeldman:


packages/playwright-core/src/server/recorder.ts
```diff
@@ -383,6 +389,14 @@ class ContextRecorder extends EventEmitter {
     this._generator = generator;
   }
 
+  testIdAttributeName() {
+    return this._testIdAttributeName;
+  }
+
+  settestIdAttributeName(testIdAttributeName: string) {
```
dgozman, line 396:
```suggestion
  setTestIdAttributeName(testIdAttributeName: string) {
```

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -132,7 +135,12 @@ export class InjectedScript {
     return globalThis.eval(expression);
   }
 
-  parseSelector(selector: string): ParsedSelector {
+  testIdAttributeNameForStrictErrorAndConsoleCodegen(): string {
+    return this._testIdAttributeNameForErrorAndConsoleCodegen;
+  }
+
+  parseSelectorBestEffortForTools(selector: string): ParsedSelector {
+    // Don't use this for strict parsing, only server can parse reliably.
```
dgozman, line 143: Can we explain here why? Otherwise it is unclear, because the method parses everything perfectly fine.

dgozman (approved):

PR #18688, opened 2022-11-09
fix(codegen): update priorites in selector generator

dgozman:
- prefer `role=checkbox` over `input[type=checkbox]`
- prefer `#id` over `input[type=checkbox]` and `role=checkbox`
- prefer `text=foo` over `internal:has-text=foo`
- ignore `none` and `presentation` roles
- remove non-strict support

### DIFF
--- a/tests/library/selector-generator.spec.ts
+++ b/tests/library/selector-generator.spec.ts
@@ -78,7 +78,7 @@ it.describe('selector generator', () => {
       <select><option>foo</option></select>
       <select mark=1><option>bar</option></select>
     `);
-    expect(await generate(page, '[mark="1"]')).toBe('select >> nth=1');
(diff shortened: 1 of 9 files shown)


tests/library/selector-generator.spec.ts
```diff
@@ -397,4 +401,9 @@ it.describe('selector generator', () => {
     await page.setContent(`<label for=target>Coun"try</label><input id=target>`);
     expect(await generate(page, 'input')).toBe('internal:label="Coun\\\"try"i');
   });
+
+  it('should prefer role other input[type]', async ({ page }) => {
```
aslushnikov, line 405:
```suggestion
  it('should prefer role over input[type]', async ({ page }) => {
```

aslushnikov (approved):

PR #18701, opened 2022-11-10
chore: report paused signal to the debug controller clients

pavelfeldman:


packages/playwright-core/src/server/debugController.ts
```diff
@@ -47,6 +48,7 @@ export class DebugController extends SdkObject {
   constructor(playwright: Playwright) {
     super({ attribution: { isInternalPlaywright: true }, instrumentation: createInstrumentation() } as any, undefined, 'DebugController');
     this._playwright = playwright;
+    Recorder.setAppFactory(async () => new InspectingRecorderApp(this));
```
dgozman, line 51: Move this to `initialize`, otherwise default recorder app is always overwritten.

dgozman (approved):

PR #18703, opened 2022-11-10
feat(clear): remove `clear()` methods from all classes except `Locator`

dgozman:
The preferred way is to use the locator, so we should not encourage the use of `ElementHandle` or `Page` in the new APIs.

### DIFF
--- a/docs/src/api/class-androiddevice.md
+++ b/docs/src/api/class-androiddevice.md
@@ -16,20 +16,6 @@ Emitted when the device connection gets closed.
 
 Emitted when a new WebView instance is detected.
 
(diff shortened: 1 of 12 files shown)


mxschmitt (approved):

dgozman:
> My bad, I had no idea. So this basically means no new methods on `ElementHandle` or `Page` ?

Well, the answer is "it depends" 😄  But as a rule of thumb, no new methods on `Page` that take a `selector`, or new methods on `ElementHandle`.

jfgreffier:
Fair enough 😀
Thanks

PR #18714, opened 2022-11-10
chore: support narrow terminal windows for messages

aslushnikov:
This patch starts using a message box that's not really a box and thus
is better behaving on a narrow-width terminals.

Before:
image

After:
image

### DIFF
--- a/packages/playwright-core/src/utils/index.ts
+++ b/packages/playwright-core/src/utils/index.ts
@@ -192,6 +192,13 @@ export function constructURLBasedOnBaseURL(baseURL: string | undefined, givenURL
 export function wrapInASCIIBox(text: string, padding = 0): string {
   const lines = text.split('\n');
   const maxLength = Math.max(...lines.map(line => line.length));
+  if (process.stdout.isTTY && process.stdout.columns < maxLength + padding * 2 + 2) {
+    return [
+      '═'.repeat(process.stdout.columns),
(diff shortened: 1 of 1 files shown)


dgozman (approved):

packages/playwright-core/src/utils/index.ts
```diff
@@ -191,12 +191,10 @@ export function constructURLBasedOnBaseURL(baseURL: string | undefined, givenURL
...
 export function wrapInASCIIBox(text: string, padding = 0): string {
   const lines = text.split('\n');
-  const maxLength = Math.max(...lines.map(line => line.length));
-  return [
-    '╔' + '═'.repeat(maxLength + padding * 2) + '╗',
-    ...lines.map(line => '║' + ' '.repeat(padding) + line + ' '.repeat(maxLength - line.length + padding) + '║'),
-    '╚' + '═'.repeat(maxLength + padding * 2) + '╝',
-  ].join('\n');
+  const maxLineLength = Math.max(...lines.map(line => line.length));
+  const separatorLength = process.stdout.isTTY ? process.stdout.columns : maxLineLength;
```
dgozman, line 195: I think `columns` could be zero in some cases, perhaps the following would work better?

```suggestion
  const separatorLength = process.stdout.columns || maxLineLength;
```

dgozman (approved):

PR #18719, opened 2022-11-11
fix(getByRole): name and exact

dgozman:
Following the `getByText()` and other methods:

- By default, matching is substring and case-insensitive. Before, it was only case-insensitive, but not substring.
- With new option `exact: true`, matching is full string and case-sensitive.
- Matching always normalizes whitespace.
- Codegen generates `exact: false` by default.
- `internal:role` treats `[name="foo"i]` as non-exact match.

Various fixes:
- Updated `getByRole` docs to match the reality.
- Locator generator edge cases.

### DIFF
--- a/packages/playwright-core/src/server/injected/roleSelectorEngine.ts
+++ b/packages/playwright-core/src/server/injected/roleSelectorEngine.ts
@@ -107,8 +107,8 @@ function validateAttributes(attrs: AttributeSelectorPart[], role: string) {
   }
 }
 
-export const RoleEngine: SelectorEngine = {
-  queryAll(scope: SelectorRoot, selector: string): Element[] {
+export function createRoleEngine(internal: boolean): SelectorEngine {
+  const queryAll = (scope: SelectorRoot, selector: string): Element[] => {
     const parsed = parseAttributeSelector(selector, true);
     const role = parsed.name.toLowerCase();
     if (!role)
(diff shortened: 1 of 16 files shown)


packages/playwright-core/src/utils/isomorphic/locatorUtils.ts
```diff
@@ -71,9 +72,14 @@ export function getByRoleSelector(role: string, options: ByRoleOptions = {}): st
...
     props.push(['level', String(options.level)]);
-  if (options.name !== undefined)
-    props.push(['name', isString(options.name) ? escapeForAttributeSelector(options.name, false) : String(options.name)]);
   if (options.pressed !== undefined)
     props.push(['pressed', String(options.pressed)]);
-  return `internal:role=${role}${props.map(([n, v]) => `[${n}=${v}]`).join('')}`;
+  let result = `internal:role=${role}${props.map(([n, v]) => `[${n}=${v}]`).join('')}`;
+  if (options.name !== undefined) {
+    if (isString(options.name))
+      result += `[name${options.exact ? '=' : '*='}${escapeForAttributeSelector(options.name, !!options.exact)}]`;
```
pavelfeldman, line 80: We already encode non-exact as "foo"i, why additional *= here?

pavelfeldman (commented):
Could you add locator-generator tests?

packages/playwright-core/src/server/injected/roleSelectorEngine.ts
```diff
@@ -149,7 +149,13 @@ export const RoleEngine: SelectorEngine = {
           return;
       }
       if (nameAttr !== undefined) {
-        const accessibleName = getElementAccessibleName(element, includeHidden, hiddenCache);
+        // Always normalize whitespace in the accessible name.
+        const accessibleName = getElementAccessibleName(element, includeHidden, hiddenCache).trim().replace(/\s+/g, ' ');
+        if (typeof nameAttr.value === 'string')
+          nameAttr.value = nameAttr.value.trim().replace(/\s+/g, ' ');
+        // internal:role assumes that [name="foo"i] also means substring.
```
pavelfeldman, line 156: We can follow up and rename all 'i' to 'lax'

packages/playwright-core/src/server/isomorphic/locatorGenerators.ts
```diff
@@ -22,8 +22,9 @@ export type Language = 'javascript' | 'python' | 'java' | 'csharp';
 export type LocatorType = 'default' | 'role' | 'text' | 'label' | 'placeholder' | 'alt' | 'title' | 'test-id' | 'nth' | 'first' | 'last' | 'has-text' | 'has' | 'frame';
 export type LocatorBase = 'page' | 'locator' | 'frame-locator';
 
+type LocatorOptions = { attrs?: { name: string, value: string | boolean | number}[], exact?: boolean, name?: string | RegExp };
```
pavelfeldman, line 25: attrs.name and name are clashing for me.

pavelfeldman (approved):

PR #18748, opened 2022-11-11
fix: ensure write permission bit is set on HTML report index.html

### LINKED ISSUE #18747: [BUG] HTML Report index.html is created as read-only and can't have playwrightReportBase64 written
**Context:**
- Playwright Version: Version 1.25.0
- Operating System: macOS
- Node.js version: 16.18.1
- Browser: N/A
- Extra: Using Nix Package manager

**Describe the bug**

If the Playwright node module is installed into a read-only location, the HTML reporter creates an index.html file that is read-only and then subsequently fails with the error:

```
Error in reporter Error: EACCES: permission denied, open '<redacted>/playwright-report/index.html'
    at Object.openSync (node:fs:590:3)
```

...when trying to append the report base64 data.

I think this is probably down to the file simply being copied here https://github.com/microsoft/playwright/blob/765ac5f0a7a0949ebc48abd5cbde294ea7c93f63/packages/html-reporter/bundle.js#L48 perhaps the file should have the write permission bit explicitly set.

madjam002:
Fixes #18747

### DIFF
--- a/packages/playwright-test/src/reporters/html.ts
+++ b/packages/playwright-test/src/reporters/html.ts
@@ -272,7 +272,10 @@ class HtmlBuilder {
 
     // Copy app.
     const appFolder = path.join(require.resolve('playwright-core'), '..', 'lib', 'webpack', 'htmlReport');
-    fs.copyFileSync(path.join(appFolder, 'index.html'), path.join(this._reportFolder, 'index.html'));
+    const destIndexPath = path.join(this._reportFolder, 'index.html');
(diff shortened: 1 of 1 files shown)


pavelfeldman:
Ouch, I got my PR approved and I automatically merged it. We can follow-up to it with inheriting the flags.

> Is there a reason why the group bit should be set as well?

That's what would happen if we created the file, so I think we should be consistent with the default behavior.

pavelfeldman:
Closing as stale

Issue #18755, opened 2022-11-12
[BUG] Meta+X not working on macOS for Chromium


luin:
Meta+X doesn't work whereas Meta+C/Meta+A work as expected. Not sure whether it's due to https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/macEditingCommands.ts doesn't include cut.

**Context:**
Playwright Version: 1.27.1
Operating System: Mac
Node.js version: 16
Browser: Chromium

**Code Snippet**

```javascript
const { chromium } = require('playwright');

(async () => {
  const browser = await chromium.launch({
    headless: false,
  });
  const context = await browser.newContext();
  await context.grantPermissions(['clipboard-read']);
  const page = await context.newPage();
  await page.goto('https://output.jsbin.com/zifated');
  await page.locator('input').click();
  await page.keyboard.type(String(Date.now()));
  await page.keyboard.press('Meta+KeyA');
  await page.keyboard.press(`Meta+KeyX`); // works if changing KeyX to KeyC.

  console.log(
    await page.evaluate(async () => {
      return (await (await navigator.clipboard.read())[0].getType('text/plain')).text();
    }),
  );

  await browser.close();
})();
```

**Describe the bug**

With `Meta+KeyX`, the Hello*** text is not written into the clipboard, whereas `Meta+KeyC` works. Make sure to clear your clipboard before running the test.

LiutskoOlga:
Hello,
I have a problem with next shortcut that added on our site -> 
'Control+Z'/'Meta+Z' - for undo and 'Control+Y'/'Control+Shift+Z' for redo.
It is working in debug mode, but it is not working on headless mode
Thank you for helping to resolve

luin:
The original Meta+X issue was fixed with https://github.com/microsoft/playwright/pull/18756 so I'm okay to close it. Not sure we need to keep this open for tracking the issue mentioned by @LiutskoOlga?

skogsmaskin:
I want to mention that `Meta+KeyV` has also stopped working when triggering pasting into a contentEditable div on MacOS in Chromium. Works fine on Linux, and used to work on MacOS (I think before v. `chromium-1055`).

StanislavIvanovQA:
I have the following function:
`export const pasteFromClipboard = async (page: Page) => {
  await test.step('Pasting value from clipboard', async () => {
    await page.keyboard.press(`${process.env.CI ? 'Control' : 'Meta'}+V`);
  });
};`
[...]

PR #18756, opened 2022-11-12
feat(keyboard): support simple cut-pasting using meta+x/v

luin:
Similar to how we handle Meta+c, in this PR we support Meta+x so users can cut text on the page.

References #18755

### DIFF
--- a/tests/page/page-keyboard.spec.ts
+++ b/tests/page/page-keyboard.spec.ts
@@ -482,7 +482,7 @@ it('should support simple copy-pasting', async ({ page, isMac, browserName }) =>
   expect(await page.evaluate(() => document.querySelector('div').textContent)).toBe('123123');
 });
 
-it('should support simple cut-pasting', async ({ page, isMac, browserName }) => {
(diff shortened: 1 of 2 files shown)


luin:

tests/page/page-keyboard.spec.ts
```diff
@@ -482,6 +482,17 @@ it('should support simple copy-pasting', async ({ page, isMac, browserName }) =>
   expect(await page.evaluate(() => document.querySelector('div').textContent)).toBe('123123');
 });
 
+it('should support simple cut-pasting', async ({ page, isMac, browserName }) => {
```
ErnstdeJong, line 485: why is browserName called here? it's not used

luin: Good catch. Updated.

luin: cc @dgozman not sure why re-requesting review removed you from the list. Don't think I have permission to add you so ping manually.

pavelfeldman (approved):

PR #18764, opened 2022-11-13
test(ct): intercept http request

sand4rt:


pavelfeldman:
What prompted these if you don't mind me asking? Playwright has many features and we don't test them in the CT context because we assume they just work...

pavelfeldman:
Let's start with slicing the test!

PR #18765, opened 2022-11-13
test(ct): vue3 store

sand4rt:


pavelfeldman:
Same here, is this more like a test or as an example?

sand4rt:
> Same here, is this more like a test or as an example?

Would really like to known if i'm wrong but IMO both. A test for every Vue plugin seems overkill to me but Pinia is the default state manager of Vue and will be widely used. I think a test that makes sure this keeps working end to end _(with refactoring and maybe with the third hook you talked about earlier)_ is useful. If the test is written as the user uses the software, it automatically becomes a form of documentation.

I really like the quote from Testing Library on this: "The more your tests resemble the way your software is used, the more confidence they can give you."

pavelfeldman:
At this point we are mindful of the test efficiency - they are taking time and eating CPUs, so we want them to be as efficient as possible. The reason I was asking was more in the context of #18764 - we should not test Playwright's features that are already covered.

I agree it is important to make sure Playwright interoperates with Pinia and pin it with the test. For this purpose, I would recommend introducing a self-contained test called pinia-interop.spec.ts that would cover that aspect of the API and the feature. I.e. instead of a single kitchen sink test that tests everything, we would have a modular approach for interop with other modules. Also makes it easier to locate it for those looking for an example.

tests/components/ct-vue-vite/src/notation-jsx.spec.tsx
```diff
@@ -151,3 +152,17 @@ test('render app and navigate to a page', async ({ page, mount }) => {
   await expect(component.getByRole('main')).toHaveText('Dashboard');
   await expect(page).toHaveURL('/dashboard');
 });
+
+test('render global store state', async ({ mount }) => {
```
pavelfeldman, line 156: Same as in #18765, let's start slicing tests by feature rather than putting everything in one test!

sand4rt:
Awesome! Yeah agree, will slice them up by feature! Does this mean we can get rid of the example project?: https://github.com/microsoft/playwright/pull/16871. The tests contained therein also correspond to the tests already written in the other ct test project.

pavelfeldman:
Closing as per above, let's start with slicing the tests.

PR #18768, opened 2022-11-14
chore: verify tab groups in docs during lint

dgozman:
This extracts the logic from playwright.dev so that we get early warnings.


mxschmitt (commented):
Instead of that I thought about instead triggering a docs roll+build if files in docs/ get changed.
Should not require much compute and not hurt much.

dgozman:
> Instead of that I thought about instead triggering a docs roll+build if files in docs/ get changed. Should not require much compute and not hurt much.

I don't think this will work nicely on my local machine, but `npm run doc` just works.

yury-s (approved):

PR #18799, opened 2022-11-14
chore: allow `py` code blocks for python

dgozman:
Turns out, we have some snippets that use `py` instead of `python`.


yury-s (approved):

docs/src/chrome-extensions-js-python.md
```diff
@@ -99,17 +99,20 @@ with sync_playwright() as playwright:
 
 To have the extension loaded when running tests you can use a test fixture to set the context. You can also dynamically retrieve the extension id and use it to load and test the popup page for example.
 
+First, add fixtures that will load the extension:
+
 ```ts
-import { test as base, expect, BrowserContext } from "@playwright/test";
+// fixtures.ts
```
mxschmitt, line 105: this file has mixed kind of quotes.

PR #18817, opened 2022-11-15
docs: Html rendering for locators instead of images

debs-obrien:
Web capture_15-11-2022_173712_localhost


pavelfeldman:
This PR is now stale

PR #18890, opened 2022-11-17
test(cli): setup files are in list-files and test --list output

yury-s:


pavelfeldman:
This makes my VS Code extension code harder for no good reason. I also need to align and release a new version of extension for that...

yury-s:
> This makes my VS Code extension code harder for no good reason. I also need to align and release a new version of extension for that...

what do you suggest instead?

pavelfeldman (approved):

PR #18900, opened 2022-11-18
feat(runner): allow filtering setup and tests

yury-s:
Running `npx playwright test file:123` will have the following behavior
- if only test files match then only matching subset of tests will run but all setup files will run as well
- if only setup files match the filter then only those setup tests will run
- if both setup and test files match an error will be thrown

### DIFF
--- a/packages/playwright-test/src/runner.ts
+++ b/packages/playwright-test/src/runner.ts
@@ -196,7 +196,7 @@ export class Runner {
 
   async listTestFiles(projectNames: string[] | undefined): Promise<any> {
     const projects = this._collectProjects(projectNames);
-    const { filesByProject } = await this._collectFiles(projects, () => true);
+    const { filesByProject } = await this._collectFiles(projects, []);
     const report: any = {
       projects: []
     };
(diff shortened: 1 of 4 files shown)


packages/playwright-test/src/runner.ts
```diff
@@ -270,13 +271,41 @@ export class Runner {
...
+          if (!testFileFilter(file))
+            setupFiles.delete(file);
+        }
+      } else {
+        const setupFile = [...setupFiles].find(testFileFilter);
+        // If the filter is not empty and it matches both setup and tests then it's an error: we allow
+        // to run either subset of tests with full setup or partial setup without any tests.
+        if (setupFile) {
+          const testFile = Array.from(fileToProjectName.keys()).find(f => !setupFiles.has(f));
+          throw new Error(`Both setup and test files match command line filter.\n  Setup file: ${setupFile}\n  Test file: ${testFile}`);
```
dgozman, line 297: - Let's print relative to `config.rootDir`. 
- We can also print the number of matched files if more than one.

packages/playwright-test/src/runner.ts
```diff
@@ -270,13 +271,41 @@ export class Runner {
...
       filesByProject.set(project, testFiles);
     }
-    return { filesByProject, setupFiles };
+
+    let applyFilterToSetup = false;
+    if (testFileFilters.length && setupFiles.size) {
+      // If the filter didn't match any tests apply it to the setup files.
+      if (setupFiles.size === fileToProjectName.size) {
+        applyFilterToSetup = true;
+        for (const [project, files] of filesByProject) {
```
dgozman, line 280:
```suggestion
        // We now have only setup files in filesByProject, because all test files were filtered out.
        for (const [project, setupFiles] of filesByProject) {
```

packages/playwright-test/src/runner.ts
```diff
@@ -235,12 +235,13 @@ export class Runner {
...
   }
 
-  private async _collectFiles(projects: FullProjectInternal[], testFileFilter: Matcher): Promise<{filesByProject: Map<FullProjectInternal, string[]>; setupFiles: Set<string>}> {
+  private async _collectFiles(projects: FullProjectInternal[], testFileFilters: TestFileFilter[]): Promise<{filesByProject: Map<FullProjectInternal, string[]>; setupFiles: Set<string>, applyFilterToSetup: boolean}> {
     const extensions = ['.js', '.ts', '.mjs', '.tsx', '.jsx'];
     const testFileExtension = (file: string) => extensions.includes(path.extname(file));
     const filesByProject = new Map<FullProjectInternal, string[]>();
     const setupFiles = new Set<string>();
     const fileToProjectName = new Map<string, string>();
+    const testFileFilter = fileMatcherFrom(testFileFilters);
```
dgozman, line 244:
```suggestion
    const commandLineFileFilter = fileMatcherFrom(testFileFilters);
```

packages/playwright-test/src/runner.ts
```diff
@@ -270,13 +271,41 @@ export class Runner {
       });
       filesByProject.set(project, testFiles);
     }
-    return { filesByProject, setupFiles };
+
+    let applyFilterToSetup = false;
+    if (testFileFilters.length && setupFiles.size) {
```
dgozman, line 276: Let's move this condition to the place where we throw.

dgozman (approved):

PR #18926, opened 2022-11-18
fix(click): account for transformed iframes

### LINKED ISSUE #18245: [BUG] interacting with iframe that has webkit-transform: scale() set, causes iframe content to jump around
**Context:**
- Playwright Version: [1.27.1]
- Operating System: [Mac]
- Node.js version: [e.g. 17.9.0]
- Browser: [All]

## System:
 - OS: macOS 12.5.1
 - Memory: 98.22 MB / 16.00 GB
## Binaries:
 - Node: 17.9.0 - ~/.nvm/versions/node/v17.9.0/bin/node
 - Yarn: 1.22.19 - /opt/homebrew/bin/yarn
 - npm: 8.5.5 - ~/.nvm/versions/node/v17.9.0/bin/npm

**Code Snippet**

Here is the content of a test for the below HTML.
Please see videos down below.
```javascript
    await page.goto("..../iframe-test.html")

    await page.frameLocator('#iframe-test').getByRole('link', { name: 'Contents' }).click();
    await page.frameLocator('#iframe-test').getByRole('link', { name: 'Contents' }).click();
    await page.frameLocator('#iframe-test').getByRole('link', { name: 'Contents' }).click();
    await page.frameLocator('#iframe-test').getByRole('link', { name: 'Contents' }).click();
    await page.frameLocator('#iframe-test').getByRole('link', { name: 'Contents' }).click();
    await page.frameLocator('#iframe-test').getByRole('link', { name: 'Contents' }).click();
```

``` html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>test</title>
</head>
<body>
<div>
    <p>before iframe</p>

    <iframe id="iframe-test" src="https://wikipedia.org/wiki/Main_Page" width="640" height="480" style="-webkit-transform: scale(0.82);"></iframe>

    <p>after iframe</p>
</div>
</body>
</html>


```

**Describe the bug**

It seems that by adding `style="-webkit-transform: scale(0.82);"` to an `<iframe>` it causes some weird behaviours with the playwright web driver.

In my applications it causes an endless cycle of jumping around without interacting with the iframe. Please see videos at the bottom.
My case is weird, I have about 3 nested Iframes that can not be changed.

But here are three simpler tests that uses the html and test above.

1. a manual test with -webkit-transform: scale(0.82);
2. a Playwright test with -webkit-transform: scale(0.82);
3. a Playwright test without the scale set

using `npx playwright codegen <above html localhost url>` will have the same jumping around when clicking on buttons.

# 1. Manual Test with -webkit-transform: scale(0.82);

I'm manually clicking the `Contents` link. There is no jumping.

https://user-images.githubusercontent.com/24510543/197189071-62135473-d597-4190-9627-49a8c74bae46.mov

----

# 2. Playwright Test with -webkit-transform: scale(0.82);

Playwright is clicking `Contents` link inside the iframe

The test is clicking the link a bunch of times.

There is jumping around.

https://user-images.githubusercontent.com/24510543/197189664-174fee83-d92c-4775-ad4b-fd8572a296ef.mov

---

# 3. Playwright Test with NO scale

Playwright is clicking `Contents` link inside the iframe
There is no jumping around.

https://user-images.githubusercontent.com/24510543/197190978-4f7a7df7-cd26-4e18-9552-f0119d22f1b8.mov

---

# My Actual application with -webkit-transform: scale(0.82);

Endless jumping and not pressing the `next` button
Everything under and including the `Payment Security` block is in an iframe.

My test only interacts with the `next` button once and this happens.

https://user-images.githubusercontent.com/24510543/197191311-090f13c7-9859-4c7c-9661-f7797971b2ce.mov

---

# My Actual application with no scale

https://user-images.githubusercontent.com/24510543/197191854-c456489e-ce40-49dc-916e-5c9efd2ffc57.mov

dgozman:
- Properly convert coordinates for iframes with non-zero borders.
- IFrames that have `transform` anywhere in the ancestors skip `hitPoint`-based check because we cannot reliably translate the viewport point into frame document's coordinates.

Fixes #18245.

### DIFF
--- a/packages/playwright-core/src/server/dom.ts
+++ b/packages/playwright-core/src/server/dom.ts
@@ -869,16 +869,23 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
     return result;
   }
 
-  async _checkFrameIsHitTarget(point: types.Point): Promise<{ framePoint: types.Point } | 'error:notconnected' | { hitTargetDescription: string }> {
+  async _checkFrameIsHitTarget(point: types.Point): Promise<{ framePoint: types.Point | undefined } | 'error:notconnected' | { hitTargetDescription: string }> {
     let frame = this._frame;
     const data: { frame: frames.Frame, frameElement: ElementHandle<Element> | null, pointInFrame: types.Point }[] = [];
     while (frame.parentFrame()) {
       const frameElement = await frame.frameElement() as ElementHandle<Element>;
       const box = await frameElement.boundingBox();
-      if (!box)
+      const style = await frameElement.evaluateInUtility(([injected, iframe]) => injected.describeIFrameStyle(iframe), {}).catch(e => 'error:notconnected' as const);
+      if (!box || style === 'error:notconnected')
         return 'error:notconnected';
+      if (style === 'transformed') {
+        // We cannot translate coordinates when iframe has any transform applied.
+        // The best we can do right now is to skip the hitPoint check,
+        // and solely rely on the event interceptor.
+        return { framePoint: undefined };
+      }
       // Translate from viewport coordinates to frame coordinates.
-      const pointInFrame = { x: point.x - box.x, y: point.y - box.y };
+      const pointInFrame = { x: point.x - box.x - style.borderLeft, y: point.y - box.y - style.borderTop };
       data.push({ frame, frameElement, pointInFrame });
       frame = frame.parentFrame()!;
     }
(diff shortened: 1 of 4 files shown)


pavelfeldman (approved):

PhilipStubbs:
Awesome, thanks @dgozman and @pavelfeldman 
Looking forward to 1.29!

PR #18938, opened 2022-11-20
Document precedence of route definitions

aryzing:
Adds small section on route matching precedence.

### DIFF
--- a/docs/src/network.md
+++ b/docs/src/network.md
@@ -421,6 +421,17 @@ await page.RouteAsync("**/api/fetch_data", async route => {
 await page.GotoAsync("https://example.com");
 ```
 
+#### Precedence
+
+When a request matches multiple routes, the last matched route is used to handle the request. As an example, the routes below will return data for `useful-resource`, and block all other requests to the API.
+
+```js
+page.route("**/api", route => route.abort());
+page.route("**/api/useful-resource", route => route.fulfill(/* data */));
+```
+
+It is also possible to explicitly fall back to another matching route with [`method: Route.fallback`].
+
 ## Modify requests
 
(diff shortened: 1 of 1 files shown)


docs/src/network.md
```diff
@@ -421,6 +421,17 @@ await page.RouteAsync("**/api/fetch_data", async route => {
...
+#### Precedence
+
+When a request matches multiple routes, the last matched route is used to handle the request.
+
+```js
+page.route("**/api", route => route.abort());
+page.route("**/api/useful-resource", route => route.fulfill(/* data */));
+```
+
+The above routes will return data for `useful-resource`, and block all other requests to the API.
```
pavelfeldman, line 433: Do you want to also mention the `route.fallback` here?

aryzing: Great idea. Added, and reworded the text a little. 

Not sure the link is formatted correctly, I looked at how other links are defined and tried to follow the same convention, let me know if it needs changing.

docs/src/network.md
```diff
@@ -421,6 +421,17 @@ await page.RouteAsync("**/api/fetch_data", async route => {
 await page.GotoAsync("https://example.com");
 ```
 
+#### Precedence
+
+When a request matches multiple routes, the last matched route is used to handle the request. As an example, the routes below will return data for `useful-resource`, and block all other requests to the API.
+
+```js
```
pavelfeldman, line 428: Sorry for the slow review. Now we need the snippets for the other languages as well, as below.

aryzing: Thanks for the review. I'm not familiar with the other languages, so I'll close the issue for now.

PR #18967, opened 2022-11-21
fix(runner): run all setup files when there is test.only

yury-s:


packages/playwright-test/src/runner.ts
```diff
@@ -671,9 +673,9 @@ export class Runner {
   }
 }
 
-function filterOnly(suite: Suite) {
+function filterOnly(suite: Suite, setupFiles: Set<string>) {
```
dgozman:
```suggestion
function filterOnly(suite: Suite, doNotFilterFiles: Set<string>) {
```

dgozman (approved):

PR #18972, opened 2022-11-21
fix(electron): consistently emit ready event after app is loaded

### LINKED ISSUE #18928: [BUG] Electron firstWindow times out after upgrading to 1.28.0
**Context:**
- Playwright Version: 1.28.0
- Operating System: Tested on Windows and Mac
- Node.js version: 16
- Browser: Electron
- Extra: Wasn't a problem in 1.27.1 or below. Repros 100% of the time after upgrading.

**Code Snippet**

```javascript
import { ElectronApplication, Page, _electron as electron } from "playwright";

(async () => {
  const electronApp = await electron.launch({
            args: [path.join(__dirname, "../dist/App.js"),
            '--enable-logging',
            '--skip-welcome',
            '--disable-telemetry',
            '--no-cached-data',
            ],
            env: {
                automation: "1",
                authConfig: JSON.stringify(config),
            }
        });
  const page = await electronApp.firstWindow(); // times out after 30 seconds
})();
```

**Describe the bug**

After upgrading to 1.28.0 Electron tests are all timing out. Downgrading to 1.27.1 resolves the issue. Example test to assist in repro can be found [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/msal-node-samples/ElectronTestApp/test/electron-code-aad.spec.ts)

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/18928

### DIFF
--- a/packages/playwright-core/src/server/electron/loader.ts
+++ b/packages/playwright-core/src/server/electron/loader.ts
@@ -32,6 +32,25 @@ for (const arg of chromiumSwitches) {
   app.getAppPath = () => path.dirname(appPath);
 }
 
-(globalThis as any).__playwright_run = () => {
+let launchInfoEventPayload: any;
+app.on('ready', launchInfo => launchInfoEventPayload = launchInfo);
+
+(globalThis as any).__playwright_run = async () => {
+  // Wait for app to be ready to avoid browser initialization races.
+  await app.whenReady();
+
+  // Override isReady pipeline.
+  let isReady = false;
+  let whenReadyCallback: () => void;
+  const whenReadyPromise = new Promise<void>(f => whenReadyCallback = f);
+  app.isReady = () => isReady;
+  app.whenReady = () => whenReadyPromise;
+
   require(appPath);
+
+  // Trigger isReady.
+  isReady = true;
+  whenReadyCallback!();
+  app.emit('will-finish-launching');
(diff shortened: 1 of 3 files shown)


dgozman (approved):

packages/playwright-core/src/server/electron/loader.ts
```diff
@@ -32,6 +32,25 @@ for (const arg of chromiumSwitches) {
   app.getAppPath = () => path.dirname(appPath);
 }
 
-(globalThis as any).__playwright_run = () => {
+let launchInfoEventPayload: any;
+app.on('ready', launchInfo => launchInfoEventPayload = launchInfo);
+
+(globalThis as any).__playwright_run = async () => {
+  // Wait for app to be ready to avoid browser initialization races.
+  await app.whenReady();
```
andersk, line 40: This has caused a regression for apps using Electron APIs that need to be called before the `ready` event fires, such as [`protocol.registerSchemesAsPrivileged`](https://www.electronjs.org/docs/latest/api/protocol#protocolregisterschemesasprivilegedcustomschemes) and [`app.disableHardwareAcceleration`](https://www.electronjs.org/docs/latest/api/app#appdisablehardwareacceleration).

We need to give the user a chance to run synchronous code before the `ready` event fires.

- #19412

PR #18975, opened 2022-11-22
docs(python): expand assertion List types

mxschmitt:
https://github.com/microsoft/playwright-python/issues/1524

The following was before:

```txt
typing.Union[
    typing.List[typing.Union[typing.Pattern[str], str]],
    typing.Pattern[str],
    str,
]
```

but it should be:

```txt
typing.Union[
    typing.List[typing.Pattern[str]],
    typing.List[str],
    typing.List[typing.Union[typing.Pattern[str], str]],
    typing.Pattern[str],
    str,
]
```

Otherwise mypy yells at us.

### DIFF
--- a/docs/src/api/class-locatorassertions.md
+++ b/docs/src/api/class-locatorassertions.md
@@ -202,7 +202,7 @@ The opposite of [`method: LocatorAssertions.toContainText`].
 
 ### param: LocatorAssertions.NotToContainText.expected
 * since: v1.18
-- `expected` <[string]|[RegExp]|[Array]<[string]|[RegExp]>>
+- `expected` <[string]|[RegExp]|[Array]<[string]>|[Array]<[RegExp]>|[Array]<[string]|[RegExp]>>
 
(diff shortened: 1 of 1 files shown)


docs/src/api/class-locatorassertions.md
```diff
@@ -202,7 +202,7 @@ The opposite of [`method: LocatorAssertions.toContainText`].
 
 ### param: LocatorAssertions.NotToContainText.expected
 * since: v1.18
-- `expected` <[string]|[RegExp]|[Array]<[string]|[RegExp]>>
+- `expected` <[string]|[RegExp]|[Array]<[string]>|[Array]<[RegExp]>|[Array]<[string]|[RegExp]>>
```
pavelfeldman, line 205: How does this affect the ports? Does it make them better or worse? I understand that this makes TS typing unnecessary more complex (i.e. makes it a bit worse for the vast majority of our users)

pavelfeldman (approved):

PR #18976, opened 2022-11-22
fix(chromium): detach from shared workers

### LINKED ISSUE #18932: [BUG] SharedWorker is broken on Chromium since 1.27.0
**Context:**
## System:
 - OS: macOS 12.5.1
 - Memory: 141.05 MB / 64.00 GB
## Binaries:
 - Node: 18.9.1 - ~/.nvm/versions/node/v18.9.1/bin/node
 - Yarn: 3.1.0 - ~/.nvm/versions/node/v18.9.1/bin/yarn
 - npm: 8.19.1 - ~/.nvm/versions/node/v18.9.1/bin/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash

**Code Snippet**

See this repo: https://github.com/flunderpero/playwright-sharedworker-bug

If you run the tests with `npx playwright test` every second test will fail. If you change the Playwright version to 1.26.0 all tests run just fine.

Using a Worker instead of a SharedWorker works just fine. Running each test in its own worker (via `fullyParallel` for instance) works too. But as soon as two tests run in the same worker and browser, the SharedWorker does not answer. It seems as though a failure of a test somehow resets something so the next test works again,

dgozman:
This prevents shared workers from stalling upon restart.

We receive `Inspector.targetCrashed` and `Inspector.targetReloadedAfterCrash` events that assume `Runtime.runIfWaitingForDebugger` from any attached client. It is easier and more stable to just detach from shared workers, because we do not inspect them.

For service workers, we should actually issue `Runtime.runIfWaitingForDebugger` in such cases, because we attach to them.

Fixes #18932.

### DIFF
--- a/packages/playwright-core/src/server/chromium/crBrowser.ts
+++ b/packages/playwright-core/src/server/chromium/crBrowser.ts
@@ -196,6 +196,18 @@ export class CRBrowser extends Browser {
       context.emit(CRBrowserContext.CREvents.ServiceWorker, serviceWorker);
       return;
     }
+
+    // Detach from any targets we are not interested in, to avoid side-effects.
+    //
+    // One example of a side effect: upon shared worker restart, we receive
+    // Inspector.targetReloadedAfterCrash and backend waits for Runtime.runIfWaitingForDebugger
+    // from any attached client. If we do not resume, shared worker will stall.
+    //
+    // Ideally, detaching should resume any target, but there is a bug in the backend,
+    // so we must Runtime.runIfWaitingForDebugger first.
+    session._sendMayFail('Runtime.runIfWaitingForDebugger').then(() => {
+      this._session._sendMayFail('Target.detachFromTarget', { sessionId });
+    });
   }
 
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/chromium/crBrowser.ts
```diff
@@ -196,6 +196,12 @@ export class CRBrowser extends Browser {
       context.emit(CRBrowserContext.CREvents.ServiceWorker, serviceWorker);
       return;
     }
+
+    // Detach from any targets we are not interested in.
+    // Ideally, detaching should resume any target, but there is a bug in the backend.
+    session._sendMayFail('Runtime.runIfWaitingForDebugger').then(() => {
```
pavelfeldman, line 202: I don't think the severity of this issue justifies the workaround. At the very least it should link to an upstream bug, but ideally, we should close it once chromium is fixed.

pavelfeldman (approved):

PR #18978, opened 2022-11-22
chore(runner): add all setup files in separate pass

yury-s:


packages/playwright-test/src/runner.ts
```diff
@@ -307,7 +317,7 @@ export class Runner {
       const fileSuite = await this._loader.loadTestFile(file, 'runner');
       if (fileSuite._loadError)
         fatalErrors.push(fileSuite._loadError);
-      preprocessRoot._addSuite(fileSuite);
+      preprocessRoot._addSuite(fileSuite._deepClone());
```
dgozman, line 320: In theory, we can avoid cloning a) when calling the second time, b) when calling the first time but no `setupFiles` are defined.

dgozman (approved):

packages/playwright-test/src/runner.ts
```diff
@@ -346,6 +350,8 @@ export class Runner {
       const grepInvertMatcher = project.grepInvert ? createTitleMatcher(project.grepInvert) : null;
 
       const titleMatcher = (test: TestCase) => {
+        if (doNotFilterFiles.has(test._requireFile))
```
dgozman, line 353: It seems like we apply `project.grep` to setup files when calling this function for the first time.

packages/playwright-test/src/runner.ts
```diff
@@ -677,25 +646,25 @@ export class Runner {
...
-    if (!lineColumnOk)
-      return false;
-    return createFileMatcherFromFilters([filter])(testFileName);
-  });
-  const suiteFilter = (suite: Suite) => {
-    return !!suite.location && testFileLineMatches(suite.location.file, suite.location.line, suite.location.column);
-  };
+  const matchers = focusedTestFileLines.map(createFileMatcherFromFilter);
+  const testFileLineMatches = (testFileName: string, testLine: number, testColumn: number) => matchers.some(m => m(testFileName, testLine, testColumn));
+  const suiteFilter = (suite: Suite) => !!suite.location && testFileLineMatches(suite.location.file, suite.location.line, suite.location.column);
```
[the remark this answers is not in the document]
yury-s: Added, but it actually works even without it (nested tests will be in doNotFilterFiles set so their parent suite will be included).

PR #18990, opened 2022-11-22
chore: simplify slowmo implementation

pavelfeldman:


dgozman (approved):

packages/playwright-core/src/server/debugger.ts
```diff
@@ -63,6 +59,15 @@ export class Debugger extends EventEmitter implements InstrumentationListener {
       await this.pause(sdkObject, metadata);
   }
 
+  async _doSlowMo() {
+    await new Promise(f => setTimeout(f, this._slowMo));
```
dgozman, line 63: Let's fast return when `slowMo` is zero? I am afraid of unnecessary slow downs.

pavelfeldman: It happens on line 67 new code.

PR #18996, opened 2022-11-22
fix(inspector): support custom test id attribute

### LINKED ISSUE #18959: [BUG] Debug mode doesn't highlight elements on the page
**Context:**
- Playwright Version: 1.28.0
- Operating System: macOS 12.3
- Node.js version: 16.15.1
- Browser: Chromium

**Code Snippet**
Any code, related to UI automation.

**Describe the bug**
1. Run test case in debug mode `PWDEBUG=1 yarn pw:run --project=chromium`
2. Step over to the step that operates with UI element
3. Check the element is highlighted on the page.

In playwright 1.27.1: we can see the exact element that used in the test step:
image

In playwright 1.28.0: no highlighting on the same test step.

Why this behavior is expected: it's a good way to verify that selector was chosen correctly.

**Workaround**: I can click "explore", but clicking it on each test step is a bit annoying.

dgozman:
Fixes #18959.

### DIFF
--- a/packages/playwright-core/src/server/recorder.ts
+++ b/packages/playwright-core/src/server/recorder.ts
@@ -172,7 +172,7 @@ export class Recorder implements InstrumentationListener {
         actionPoint,
         actionSelector,
         language: this._currentLanguage,
-        testIdAttributeName: this._contextRecorder.testIdAttributeName(),
+        testIdAttributeName: this._context.selectors().testIdAttributeName(),
       };
       return uiState;
     });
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/recorder.ts
```diff
@@ -398,11 +394,7 @@ class ContextRecorder extends EventEmitter {
   }
 
   testIdAttributeName() {
-    return this._testIdAttributeName;
-  }
-
-  setTestIdAttributeName(testIdAttributeName: string) {
-    this._testIdAttributeName = testIdAttributeName;
+    return this._context.selectors().testIdAttributeName();
```
pavelfeldman, line 397: This getter looks weird now.

pavelfeldman (approved):

PR #19005, opened 2022-11-23
docs: annotate deprecated methods

pavelfeldman:


docs/src/api/class-frame.md
```diff
@@ -1345,6 +1362,7 @@ Time to wait between `keydown` and `keyup` in milliseconds. Defaults to 0.
 
 ## async method: Frame.querySelector
 * since: v1.9
+* discouraged: Use locator-based [`method: Locator.locator`] instead.
```
dgozman, line 1365: I think `Frame.locator` is a more suitable replacement. WDYT?

docs/src/api/class-frame.md
```diff
@@ -1368,6 +1386,7 @@ returns `null`.
 
 ## async method: Frame.querySelectorAll
 * since: v1.9
+* discouraged: Use locator-based [`method: Locator.locator`] instead.
```
dgozman, line 1389: ditto

dgozman (approved):

docs/src/api/class-frame.md
pavelfeldman: Yes, that's what I did in Page, should be same here!

PR #19060, opened 2022-11-24
docs: change listItem to article in locators doc

debs-obrien:
changed the div with role of list-item for 2 reasons:

1. the article tag is a better use case for product cards.[ see mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) for explanation
2. It is a bad practice to add the role of list to an item when a `<ul>`  or `<ol>` could be used instead.[ see here ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/list_role#best_practices)for explanation

Therefore I think article is a better for this example plus gives us something different to locate by role as we already have list examples in the doc


docs/src/locators.md
```diff
@@ -806,14 +806,14 @@ await Expect(page.Locator("x-details")).ToContainTextAsync("Details");
 Consider the following DOM structure where we want to click on the buy button of the second product card. We have a few options in order to filter the locators to get the right one.
 
 ```html card
-<div role="listitem">
+<article>
```
pavelfeldman, line 809: I think the list of products in a show is a list, not an article.

pavelfeldman:
Check out https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/article_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role

debs-obrien:
Yes but if you further study the doc it suggests you should use native elements rather than adding roles. So why add a role to a div when we can use correct html element?

5B42ADCF-AB03-4B3A-A02E-34C734673D7C
DC2DA0BF-F631-42E0-ACF0-C52604F6FC5F

docs/src/locators.md
debs-obrien: The docs also said that product cards should be articles. They can be a list of articles for sure but the card itself an article cause that card can be placed on any page. It’s independent of the other cards. It can be in a list but the card is an independent item.

debs-obrien:
Article tag for product cards according to this page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

pavelfeldman (approved):

PR #19138, opened 2022-11-29
feat(connect): support special headers for debug/attachments

dgozman:
`x-playwright-debug-log: value` headers are printed to `pw:browser` debug log.
`x-playwright-attach-string: name=value` headers are attached to each test.


pavelfeldman:
I think we can safely wait with this!

pavelfeldman:
GC

PR #19143, opened 2022-11-29
feat: allow options in automatic screenshots

### LINKED ISSUE #9983: [Feature] Allow custom options when making screenshots with Playwright Test (e.g. fullscreen)
Automatic Screenshots appear in the test output directory, typically test-results, but they are not FullPage.

You don't have the same options as in `page.screenshot([options])` , like "FullPage", which will be really useful to see some screenshot errors that otherwise you can't see (full page = full screenshot information)
```
  use: {
  },
};
export default config;
```

garciallorente-fer:
Closes #9983 

Allowing Automatic Screenshots in `PlaywrightTestConfig` to have `fullPage` & `omitBackground` as optional parameters.


garciallorente-fer:

utils/generate_types/overrides-test.d.ts
```diff
@@ -208,11 +208,12 @@ export interface PlaywrightWorkerOptions {
   channel: BrowserChannel | undefined;
   launchOptions: LaunchOptions;
   connectOptions: ConnectOptions | undefined;
-  screenshot: 'off' | 'on' | 'only-on-failure';
+  screenshot: ScreenshotMode | { mode: ScreenshotMode } & Pick<PageScreenshotOptions, 'fullPage' | 'omitBackground'>;
```
dgozman, line 211: In addition to this change, please also update the [api docs here](https://github.com/microsoft/playwright/blob/main/docs/src/test-api/class-testoptions.md?plain=1#L191-L200). Take a look at [the similar `trace` option here](https://github.com/microsoft/playwright/blob/main/docs/src/test-api/class-testoptions.md?plain=1#L221-L237) for an example.

tests/playwright-test/playwright.artifacts.spec.ts
```diff
@@ -189,6 +189,71 @@ test('should work with screenshot: only-on-failure', async ({ runInlineTest }, t
   ]);
 });
 
+test('should work with screenshot: on & fullPage & omitBackground', async ({ runInlineTest }, testInfo) => {
```
dgozman, line 192: Instead of these tests, I'd rather test that passing `fullPage` does indeed take a full page screenshot.

I'd recommend to combine the page setup of [this test](https://github.com/microsoft/playwright/blob/main/tests/page/page-screenshot.spec.ts#L214-L219) and `expect.toMatchSnapshot` the generated screenshot in the output folder. You can find the generated snapshot at something like `testInfo.outputPath('test-results', 'filename-testname', 'test-failed-1.png')`.

packages/playwright-test/src/index.ts
```diff
@@ -253,6 +253,8 @@ export const test = _baseTest.extend<TestFixtures, WorkerFixtures>({
     if (debugMode())
       testInfo.setTimeout(0);
 
+    const screenshotOptions = typeof screenshot !== 'string' ? screenshot : undefined;
```
dgozman, line 256: To avoid unexpected properties being passed to `page.screenshot`:

```suggestion
    const screenshotOptions = typeof screenshot !== 'string' ? { fullPage: screenshot.fullPage, omitBackground: screenshot.omitBackground } : undefined;
```

dgozman (commented):
Thank you for the great PR! Please address the comments and I'll merge it in.

utils/generate_types/overrides-test.d.ts
garciallorente-fer: Thanks! Added

tests/playwright-test/playwright.artifacts.spec.ts
garciallorente-fer: Makes sense, thanks! Just updated the test.

packages/playwright-test/src/index.ts
garciallorente-fer: Added thanks!

tests/playwright-test/reporter-html.spec.ts
```diff
@@ -321,6 +321,35 @@ test('should include screenshot on failure', async ({ runInlineTest, page, showR
   expect(src).toBeTruthy();
 });
 
+test('should include screenshot on failure with options', async ({ runInlineTest, page, showReport }) => {
```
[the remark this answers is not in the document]
garciallorente-fer: Makes sense, just deleted the test, thanks!

davidglezz (approved):

tests/page/pageTestApi.ts
```diff
@@ -26,6 +26,7 @@ export type PageTestFixtures = {
 export type PageWorkerFixtures = {
   headless: boolean;
   channel: string;
+  screenshot: ScreenshotMode | { mode: ScreenshotMode } & Pick<PageScreenshotOptions, 'fullPage' | 'omitBackground'>;
```
garciallorente-fer, line 29: @dgozman I also added a new `screenshot` property in `PageWorkerFixtures`

tests/playwright-test/playwright.artifacts.spec.ts
```diff
@@ -189,6 +189,35 @@ test('should work with screenshot: only-on-failure', async ({ runInlineTest }, t
...
+  expect(result.failed).toBe(1);
+  expect(listFiles(testInfo.outputPath('test-results'))).toEqual([
+    'artifacts-should-fail-and-take-fullPage-screenshots',
+    '  test-failed-1.png',
+    'report.json',
+  ]);
+  const screenshotFailure = fs.readFileSync(
+      testInfo.outputPath('test-results', 'artifacts-should-fail-and-take-fullPage-screenshots', 'test-failed-1.png')
+  );
+  expect.soft(screenshotFailure).toMatchSnapshot('screenshot-grid-fullpage.png');
```
dgozman, line 218: I think this snapshot file missing from the PR.

dgozman (commented):
Could you please add the missing snapshot? Other than that, this looks great and ready to merge!

tests/playwright-test/playwright.artifacts.spec.ts
garciallorente-fer: True, sorry about that. Just added

dgozman: Good question, I forgot about that! Since our grid page is platform-independent, we can have just a single snapshot. However, to remove platform suffix from the name, you have to add `snapshotPathTemplate` to [this file](https://github.com/microsoft/playwright/blob/main/tests/playwright-test/playwright.config.ts):

```ts
const config: Config = {
  snapshotPathTemplate: '__screenshots__/{testFilePath}/{arg}{ext}',
};
```

This should produce a snapshot in a different location, without the platform suffix, so that we can run the new test on linux/win/mac CI.

garciallorente-fer: It seems that the `Config` type from `'./stable-test-runner'` (the one from [the file you mentioned](https://github.com/microsoft/playwright/blob/main/tests/playwright-test/playwright.config.ts)) does not have a `snapshotPathTemplate` parameter.

dgozman: That's right! I have [just rolled](https://github.com/microsoft/playwright/commit/d294c5da33dfe60bca2b76ca6893969002cbd62e) `stable-test-runner` to the new version and landed `snapshotPathTemplate`. You should be good after rebase.

garciallorente-fer: Thanks @dgozman ! I have just updated the screenshot

dgozman (approved):
Looks great, merging in.

PR #19154, opened 2022-11-30
docs: remove Promise.all from examples

dgozman:
Replaced with explicit `popupPromise` variable.


yury-s (approved):

brantphoto:
Hi all, any context or links to why Promise.all example was removed from download example? I know this is an old PR so my apologies. Thanks.

mxschmitt:
> Hi all, any context or links to why Promise.all example was removed from download example? I know this is an old PR so my apologies. Thanks. 

We learned that customers often use them incorrectly, by adding e.g. an await inside the Promise array. Having it the new way is essentially the same and simpler to understand.

PR #19166, opened 2022-11-30
feat: implement a new image comparison function

aslushnikov:
This patch implements a new image comparison function, codenamed
"ssim-cie94". The goal of the new comparison function is to cancel out
browser non-determenistic rendering.

To use the new comparison function:

```ts
await expect(page).toHaveScreenshot({
  comparator: 'ssim-cie94',
});
```

As of Nov 30, 2022, we identified the following sources of
non-determenistic rendering for Chromium:
- Anti-aliasing for certain shapes might be different due to the
  way skia rasterizes certain shapes.
- Color blending might be different on `x86` and `aarch64` architectures.

The new function employs a few heuristics to fight these
differences.

Consider two non-equal image pixels `(r1, g1, b1)` and `(r2, g2, b2)`:
1. If the [CIE94] metric is less then 1.0, then we consider these pixels
   **EQUAL**. (The value `1.0` is the [just-noticeable difference] for
   [CIE94].). Otherwise, proceed to next step.
1. If all the 8 neighbors of the first pixel match its color, or
   if the 8 neighbors of the second pixel match its color, then these
   pixels are **DIFFERENT**. (In case of anti-aliasing, some of the
   direct neighbors have to be blended up or down.) Otherwise, proceed
   to next step.
1. If SSIM in some locality around the different pixels is more than
   0.99, then consider this pixels to be **EQUAL**. Otherwise, mark them
   as **DIFFERENT**. (Local SSIM for anti-aliased pixels turns out to be
   very close to 1.0).

[CIE94]: https://en.wikipedia.org/wiki/Color_difference#CIE94
[just-noticeable difference]: https://en.wikipedia.org/wiki/Just-noticeable_difference


pastelsky:
This is amazing! It might be worth also exploring multi-scale SSIM —
https://ece.uwaterloo.ca/~z70wang/publications/msssim.pdf
and ssimulacra2 — 
https://github.com/cloudinary/ssimulacra2

as further improvements to SSIM.

aslushnikov:
@pastelsky thank you for the recommendation! I'm familiar with MS-SSIM, but we don't need it here since we're only inspecting a small locality of the image around the anti-aliased pixel.

I haven't seen ssimulacra2 before - thank you for pointing it out. I'll explore it.

docs/src/test-api/class-testproject.md
```diff
@@ -110,14 +110,16 @@ export default config;
...
     - `maxDiffPixels` ?<[int]> an acceptable amount of pixels that could be different, unset by default.
     - `maxDiffPixelRatio` ?<[float]> an acceptable ratio of pixels that are different to the total amount of pixels, between `0` and `1` , unset by default.
     - `animations` ?<[ScreenshotAnimations]<"allow"|"disabled">> See [`option: animations`] in [`method: Page.screenshot`]. Defaults to `"disabled"`.
     - `caret` ?<[ScreenshotCaret]<"hide"|"initial">> See [`option: caret`] in [`method: Page.screenshot`]. Defaults to `"hide"`.
     - `scale` ?<[ScreenshotScale]<"css"|"device">> See [`option: scale`] in [`method: Page.screenshot`]. Defaults to `"css"`.
   - `toMatchSnapshot` ?<[Object]> Configuration for the [`method: SnapshotAssertions.toMatchSnapshot#1`] method.
-    - `threshold` ?<[float]> an acceptable perceived color difference in the [YIQ color space](https://en.wikipedia.org/wiki/YIQ) between the same pixel in compared images, between zero (strict) and one (lax). Defaults to `0.2`.
+    - `comparator` ?<[string]> a comparator function to use, either `"pixelmatch"` or `"ssim_v1"`. Defaults to `"pixelmatch"`.
```
dgozman, line 121:
```suggestion
    - `comparator` ?<[string]> a comparator function to use, either `"pixelmatch"` or `"ssim-cie94"`. Defaults to `"pixelmatch"`.
```

packages/playwright-core/src/image_tools/compare.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const drawRedPixel = (x: number, y: number) => drawPixel(width, diff, x, y, 255, 0, 0);
+  const drawYellowPixel = (x: number, y: number) => drawPixel(width, diff, x, y, 255, 255, 0);
+  const drawGrayPixel = (x: number, y: number) => {
+    const gray = rgb2gray(r1.get(x, y), g1.get(x, y), b1.get(x, y));
+    const value = blendWithWhite(gray, 0.1);
+    drawPixel(width, diff, x, y, value, value, value);
+  };
+
+  const fastR = new FastStats(r1, r2);
```
dgozman, line 51: I'd make these 3 lazy, upon first non-matching pixel, because they probably take the most time of the happy path.

packages/playwright-core/src/utils/comparators.ts
```diff
@@ -17,26 +17,17 @@
 
 import { colors, jpegjs } from '../utilsBundle';
 import pixelmatch from '../third_party/pixelmatch';
+import { compare } from '../image_tools/compare';
```
dgozman, line 20: Perhaps we should move this file (`comparators.ts`) to `image_tools` as well?

packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -145,6 +145,7 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
       maxDiffPixels: options.maxDiffPixels,
       maxDiffPixelRatio: options.maxDiffPixelRatio,
       threshold: options.threshold,
+      comparator: options.comparator,
```
dgozman, line 148: I think you should update the `new SnapshotHelper()` call in `toHaveScreenshot` to include `comparator` option.

dgozman (approved):

packages/playwright-core/src/utils/comparators.ts
aslushnikov: I'd keep `comparators.ts` where they are since they combine all kinds of comparators (i.e. text comparators + different image comparators).

kolodny:
Was this released in `1.30.0`? I'm still not seeing the `comparator` option after pulling in the latest stable release

Jovan27:
@aslushnikov  Can we get this included in some of the newer alpha versions or it's going to be available only with `@playwright/test@1.29.0-alpha-dec-9-2022` ?

aslushnikov:
@Jovan27 it is available as a `_comparator` option in the latest 1.32 version. It would help a ton if you can let us know if it works / doesn't work for you.

PR #19170, opened 2022-11-30
fix: Fixes #19169 Issue with moving up a directory in snapshotPathTemplate

### LINKED ISSUE #19169: [BUG] `snapshotPathTemplate` doesn't always work if trying to move up a directory using `..`
**Context:**
- Playwright Version: 1.28.1
- Operating System: Mac
- Node.js version: 14.6
- Browser: All

**Describe the bug**

I want to write my snapshots one directory above where the test file lies  —

```js
tests
├── __snapshots__ // Here
├── topic-a
│   └── test.spec.ts
└── topic-b
    └── test.spec.ts
```

But when I specify the following snapshotPathTemplate —

```ts
snapshotPathTemplate: '{testFileDir}/../__snapshots__'
```
playwright uses the `..` to remove the `{testFileDir}` symbol entirely, instead of resolving the `testFileDir` and then moving up a path.

This happens because here —

https://github.com/microsoft/playwright/blob/307a0b6ea7508306dd99c5819c47a23081760531/packages/playwright-test/src/testInfo.ts#L249-L256 

`testFileDir` is kept outside the resolve function for some reason. 

## Expectation

Template tokens such as `{testFileDir}` are substituted before the path is normalized.

pastelsky:
Fixes #19169

Note, I'm not sure why it was done this way before — e.g. Why `testDir` were considered for resolution, but `testFileDir` wasn't.

Looking for guidance on the approach here, because there are still some template tokens outside `path.resolve`

### DIFF
--- a/packages/playwright-test/src/testInfo.ts
+++ b/packages/playwright-test/src/testInfo.ts
@@ -246,13 +246,11 @@ export class TestInfoImpl implements TestInfo {
     const parsedRelativeTestFilePath = path.parse(relativeTestFilePath);
     const projectNamePathSegment = sanitizeForFilePath(this.project.name);
 
-    const snapshotPath = path.resolve(
-        this.config._configDir, this.project.snapshotPathTemplate
-            .replace(/\{(.)?testDir\}/g, '$1' + this.project.testDir)
-            .replace(/\{(.)?snapshotDir\}/g, '$1' + this.project.snapshotDir)
-            .replace(/\{(.)?snapshotSuffix\}/g, this.snapshotSuffix ? '$1' + this.snapshotSuffix : '')
-            .replace(/\{(.)?testFileDir\}/g, '$1' + parsedRelativeTestFilePath.dir)
-    )
+    const snapshotPath = this.project.snapshotPathTemplate
+        .replace(/\{(.)?testDir\}/g, '$1' + this.project.testDir)
+        .replace(/\{(.)?snapshotDir\}/g, '$1' + this.project.snapshotDir)
+        .replace(/\{(.)?snapshotSuffix\}/g, this.snapshotSuffix ? '$1' + this.snapshotSuffix : '')
+        .replace(/\{(.)?testFileDir\}/g, '$1' + parsedRelativeTestFilePath.dir)
         .replace(/\{(.)?platform\}/g, '$1' + process.platform)
         .replace(/\{(.)?projectName\}/g, projectNamePathSegment ? '$1' + projectNamePathSegment : '')
         .replace(/\{(.)?testName\}/g, '$1' + this._fsSanitizedTestName())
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/testInfo.ts
```diff
@@ -246,14 +246,16 @@ export class TestInfoImpl implements TestInfo {
...
-        .replace(/\{(.)?testDir\}/g, '$1' + this.project.testDir)
-        .replace(/\{(.)?snapshotDir\}/g, '$1' + this.project.snapshotDir)
-        .replace(/\{(.)?snapshotSuffix\}/g, this.snapshotSuffix ? '$1' + this.snapshotSuffix : ''))
+    const snapshotPath = path.resolve(
+        this.config._configDir, this.project.snapshotPathTemplate
+            .replace(/\{(.)?testDir\}/g, '$1' + this.project.testDir)
+            .replace(/\{(.)?snapshotDir\}/g, '$1' + this.project.snapshotDir)
+            .replace(/\{(.)?snapshotSuffix\}/g, this.snapshotSuffix ? '$1' + this.snapshotSuffix : '')
+            .replace(/\{(.)?testFileDir\}/g, '$1' + parsedRelativeTestFilePath.dir)
+    )
```
aslushnikov, line 255: oh thank you! Nice catch!

The idea is to actually perform all the substitutions before `path.resolve`. To avoid this confusion in future - let's move substitutions away from `path.resolve`, like this:

```ts
  snapshotPath(...pathSegments: string[]) {
    const subPath = path.join(...pathSegments);
    const parsedSubPath = path.parse(subPath);
    const relativeTestFilePath = path.relative(this.project.testDir, this._test._requireFile);
    const parsedRelativeTestFilePath = path.parse(relativeTestFilePath);
    const projectNamePathSegment = sanitizeForFilePath(this.project.name);

    const snapshotPath = this.project.snapshotPathTemplate
        .replace(/\{(.)?testDir\}/g, '$1' + this.project.testDir)
        .replace(/\{(.)?snapshotDir\}/g, '$1' + this.project.snapshotDir)
        .replace(/\{(.)?snapshotSuffix\}/g, this.snapshotSuffix ? '$1' + this.snapshotSuffix : '')
        .replace(/\{(.)?platform\}/g, '$1' + process.platform)
        .replace(/\{(.)?projectName\}/g, projectNamePathSegment ? '$1' + projectNamePathSegment : '')
        .replace(/\{(.)?testName\}/g, '$1' + this._fsSanitizedTestName())
        .replace(/\{(.)?testFileDir\}/g, '$1' + parsedRelativeTestFilePath.dir)
        .replace(/\{(.)?testFileName\}/g, '$1' + parsedRelativeTestFilePath.base)
        .replace(/\{(.)?testFilePath\}/g, '$1' + relativeTestFilePath)
        .replace(/\{(.)?arg\}/g, '$1' + path.join(parsedSubPath.dir, parsedSubPath.name))
        .replace(/\{(.)?ext\}/g, parsedSubPath.ext ? '$1' + parsedSubPath.ext : '');
    return path.normalize(path.resolve(this.config._configDir, snapshotPath));
  }
```

pastelsky: Done!

aslushnikov (approved):

PR #19182, opened 2022-11-30
fix(codegen): do not generate Promise.all

dgozman:
Drive-by: fix `download`, `dialog` and `popup` signals to use different variable names in all languages.

### DIFF
--- a/tests/library/inspector/cli-codegen-2.spec.ts
+++ b/tests/library/inspector/cli-codegen-2.spec.ts
@@ -222,39 +222,53 @@ test.describe('cli codegen', () => {
       page.waitForEvent('download'),
       page.click('a')
     ]);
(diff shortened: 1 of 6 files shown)


yury-s (approved):

tests/library/inspector/cli-codegen-2.spec.ts
```diff
@@ -326,7 +340,7 @@ test.describe('cli codegen', () => {
     }
   });
 
-  test('should record open in a new tab with url', async ({ page, openRecorder, browserName, platform }) => {
+  test.only('should record open in a new tab with url', async ({ page, openRecorder, browserName, platform }) => {
```
debs-obrien, line 343: I'm guessing .only shouldn't be here

PR #19218, opened 2022-12-01
changed option

### LINKED ISSUE #15075: [Feature] run tests only against changed files
## Description

As a user, I want to run my component tests only on changed files, as it would be a use case in a git hook ([`pre-push`](https://git-scm.com/docs/githooks#_pre_push))

## Alternatives

- https://jestjs.io/docs/cli#--onlychanged
- https://vitest.dev/guide/cli.html#changed

## Preferred Solution

Enhancement of the cli commands, e.g.

```sh
npx playwright test --only-changed
```

jfgreffier:
Run tests that are uncommited , or have changes compared to a commit.

```bash
npx playwright test --changed
npx playwright test --changed d819f97f40386c9bd31a53525240c6e9997a10f4
npx playwright test --changed main
```

Can also be passed in TestConfig
* `'main'` - run new tests e.g. from a Pull Request
* `true` - run tests with uncommited changes
* `'HEAD~1'` - run tests new compared to the last commit

Fixes #15075

### DIFF
--- a/.github/workflows/tests_primary.yml
+++ b/.github/workflows/tests_primary.yml
@@ -112,6 +112,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
     - uses: actions/setup-node@v3
       with:
         node-version: ${{matrix.node-version}}
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -39,7 +39,7 @@ export default config;
 - type: ?<[Object]>
   - `timeout` ?<[int]> Default timeout for async expect matchers in milliseconds, defaults to 5000ms.
   - `toHaveScreenshot` ?<[Object]> Configuration for the [`method: PageAssertions.toHaveScreenshot#1`] method.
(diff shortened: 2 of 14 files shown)


pavelfeldman:
Sorry for the late review. A couple of high level comments:
- I'm not too keen on taking a dependency on `jest-changed-files`, it pulls transitive dependencies that it can do the job without.
- To those working a full stack, running the changed tests only could be misleading. Imagine I touched a component. I actually want to run all the affected tests, not all the changed tests. If this is inspired by vitest, it should be based on the dependency tree rather than modified test files.

jfgreffier:
> Sorry for the late review. A couple of high level comments:
> 
> * I'm not too keen on taking a dependency on `jest-changed-files`, it pulls transitive dependencies that it can do the job without.

Thanks for the comments.
Understandable for `jest-changed-files`. Writing the necessary code to do the job is definitely feasible.

> * To those working a full stack, running the changed tests only could be misleading. Imagine I touched a component. I actually want to run all the affected tests, not all the changed tests. If this is inspired by vitest, it should be based on the dependency tree rather than modified test files.

To be precise, it is inspired by Vitest's `changed` CLI option https://vitest.dev/guide/cli.html#changed

Let's discuss further the use case in the #15075 issue 🙂

pavelfeldman:
> I have no idea on how to achieve that 

Let's keep collecting ideas in the issue. I think Vite bundler that we use under the hood exposes enough information to know which files are needed to build a component. And we know which component files each test needs. So theoretically we have all the information.

pavelfeldman:
I'll close this for now so that we could discuss it in the issue.

PR #19220, opened 2022-12-01
docs: auth in project setup

yury-s:


docs/src/auth.md
```diff
@@ -260,58 +268,64 @@ test('test', async ({ page }) => {
...
 ```
 
-:::note
-If you can log in once and commit the `storageState.json` into the repository, you won't need the global setup at all, just specify the `storageState.json` in Playwright Config as above and it'll be picked up.
+### Committing storage state int the repository
+* langs: js
 
-However, periodically, you may need to update the `storageState.json` file if your app requires you to re-authenticate after some amount of time. For example, if your app prompts you to sign in every week even if you're on the same computer/browser, you'll need to update `storageState.json` at least this often.
-:::
+When you set an entry on [`method: TestInfo.storage`] Playwright will store it in a separate file under `.playwright-storage/`. Playwright does not delete those files automatically. You can leverage this fact to persist storage state between test runs and only sign in if the entry is not in the storage yet.
```
dgozman, line 284: A small two-lines example would be nice:

```ts
test('sign in', async ({ page, context }) => {
  if (test.info().storage().get('github-test-user'))
    return;
  // ... login here ...
  await test.info().storage().set('github-test-user', await context.storageState());
});
```

docs/src/auth.md
```diff
@@ -260,58 +268,64 @@ test('test', async ({ page }) => {
...
+
+If you can log in once and commit the storage state into the repository, you won't need the setup step test that signs in at all, just specify the name of the storage entry as above and it'll be picked up.
```
dgozman, line 286:
```suggestion
If you can log in once and commit the storage state into the repository, you won't need the setup that signs in at all, just specify the name of the storage entry as above and it'll be picked up.
```

docs/src/auth.md
```diff
@@ -260,58 +268,64 @@ test('test', async ({ page }) => {
...
+
+However, you may need to periodically update the storage state entry if your app requires you to re-authenticate after some amount of time. For example, if your app prompts you to sign in every week even if you're on the same computer/browser, you'll need to update saved storage state at least this often.
```
dgozman, line 288: I'd swap this and previous paragraph, and recommend removing `./playwright-storage` directory for an easy referesher.

docs/src/auth.md
```diff
@@ -322,74 +336,87 @@ By default, Playwright Test runs tests in parallel. If you reuse a single signed
...
+  mode: 'parallel'
 });
-exports.expect = base.expect;
+
+// Sign in all test users duing project setup and save their state
+// to be used in the tests.
+for (let i = 0; i < users.length; i++) {
+  test(`login user ${i}`, async ({ page }) => {
+    await page.goto('https://github.com/login');
+    // Use a unique username for each worker.
```
dgozman, line 358:
```suggestion
```

docs/src/auth.md
```diff
@@ -322,74 +336,87 @@ By default, Playwright Test runs tests in parallel. If you reuse a single signed
...
+  });
+}
 
 // example.spec.js
-const { test, expect } = require('./fixtures');
+const { test } = require('@playwright/test');
+
+test.use({
+  // User different user for each worker.
+  storageStateName: `test-user-${test.info().parallelIndex}`
```
dgozman, line 374:
```suggestion
  storageStateName: ({}, use) => use(`test-user-${test.info().parallelIndex}`)
```

dgozman (approved):

PR #19281, opened 2022-12-06
feat: use setup() instead of test() for project setup

yury-s:


packages/playwright-test/src/dispatcher.ts
```diff
@@ -32,6 +32,7 @@ export type TestGroup = {
   projectId: string;
   tests: TestCase[];
   watchMode: boolean;
+  projectSetup: boolean;
```
dgozman, line 35:
```suggestion
  isProjectSetup: boolean;
```

packages/playwright-test/src/test.ts
```diff
@@ -23,6 +23,7 @@ class Base {
   title: string;
   _only = false;
   _requireFile: string = '';
+  _projectSetup: boolean = false;
```
dgozman, line 26: Since we are doing this, let's switch all `setupFiles.has(_requireFile)` checks to `_isProjectSetup`? Could be a follow up.

dgozman, line 26:
```suggestion
  _isProjectSetup: boolean = false;
```

dgozman (approved):

packages/playwright-test/src/test.ts
yury-s: will do in a follow-up

PR #19316, opened 2022-12-07
chore: match selected options by both value and label

pavelfeldman:


docs/src/api/class-locator.md
```diff
@@ -1007,56 +1011,59 @@ Triggers a `change` and `input` event once all the provided options have been se
...
 ```js
-// single selection matching the value
+// single selection matching the value or label
 element.selectOption('blue');
 
 // single selection matching the label
 element.selectOption({ label: 'Blue' });
 
-// multiple selection
+// multiple selection for blue, red and second option
```
dgozman, line 1029:
```suggestion
// multiple selection for red, green and blue options
```

dgozman (approved):

PR #19356, opened 2022-12-08
Add a flag to build a docker image if needed when starting container

pastelsky:
If I need to connect playwright to a local docker, I need to —
1. Ensure that there is a local image with `wss` server exposed
2. Run that docker image

before I can execute `playwright test`.

While `playwright docker build` and `playwright docker start` are helpful as setup steps, you don't always want to build un-necessarily if there's already an image available on disk (when iterating locally, say).

In this PR I add a `--build-if-needed` flag to `playwright docker start` so image is built only if it doesn't already exists.

An alternative  way to solve this problem would would be to provide a command to check for docker image's existence `playwright docker exists` but I felt that the flag implementation was neater.

### DIFF
--- a/packages/playwright-core/src/containers/docker.ts
+++ b/packages/playwright-core/src/containers/docker.ts
@@ -340,6 +340,7 @@ export function addDockerCLI(program: Command) {
   dockerCommand.command('start')
       .description('start docker container')
       .option('--port <port>', 'port to start container on. Auto-pick by default')
+      .option('--build-if-needed', 'builds docker image for the container if needed before starting')
       .action(async function(options) {
         try {
           const port = options.port ? +options.port : 0;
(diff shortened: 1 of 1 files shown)


aslushnikov:
Hey @pastelsky! We're still iterating on this story; instead of building the image, we'll actually pull it from container registry in future. This should address your usability concern.

pastelsky:
Thanks, @aslushnikov, that'd be great! I'm going to close my PR in that case. 
I'm interested in following developments w.r.t docker/playwright server — is there a public roadmap?

aslushnikov:
> I'm interested in following developments w.r.t docker/playwright server — is there a public roadmap?

@pastelsky there's no exact roadmap for container support, this is currently in the exploration phase. The best option to follow the development is to keep an eye on the pull requests that land to the repo

PR #19426, opened 2022-12-13
chore: fix Android tests

mxschmitt:


packages/playwright-core/src/server/android/android.ts
```diff
@@ -264,9 +264,13 @@ export class AndroidDevice extends SdkObject {
     debug('pw:android')('Force-stopping', pkg);
     await this._backend.runCommand(`shell:am force-stop ${pkg}`);
     const socketName = isUnderTest() ? 'webview_devtools_remote_playwright_test' : ('playwright-' + createGuid());
-    const commandLine = this._defaultArgs(options, socketName).join(' ');
+    const commandLine = this._defaultArgs(options, socketName).map(arg => {
```
mxschmitt, line 267: see [here](https://source.chromium.org/chromium/chromium/src/+/main:base/android/java/src/org/chromium/base/CommandLine.java;l=175;drc=38321ee39cd73ac2d9d4400c56b90613dee5fe29;bpv=1;bpt=1) for the upstream command-line parser

tests/android/browser.spec.ts
```diff
@@ -32,12 +32,20 @@ test('androidDevice.launchBrowser', async function({ androidDevice }) {
   await context.close();
 });
 
-test('androidDevice.launchBrowser should pass args with spaces', async ({ androidDevice }) => {
-  const context = await androidDevice.launchBrowser({ args: ['--user-agent=I am Foo'] });
-  const page = await context.newPage();
-  const userAgent = await page.evaluate(() => navigator.userAgent);
-  await context.close();
-  expect(userAgent).toBe('I am Foo');
+test('androidDevice.launchBrowser should treat args correctly', async ({ androidDevice }) => {
```
mxschmitt, line 35: - We need to escape the double quotes before we send it to the Android backend (linux bash)

utils/avd_recreate.sh
```diff
@@ -8,7 +8,14 @@ if [[ -z "${ANDROID_HOME}" ]]; then
     export ANDROID_SDK_ROOT=${SDKDIR}
 fi
 
+ANDROID_ARCH="x86_64"
+
+# on MacOS M1 we need to use arm64 (can't emulate x86_64)
+if [[ "$(uname -m)" == "arm64" ]]; then
```
mxschmitt, line 14: Makes the Android Emulator script working on M1 Macs

tests/android/browser.spec.ts
```diff
@@ -32,12 +32,20 @@ test('androidDevice.launchBrowser', async function({ androidDevice }) {
   await context.close();
 });
 
-test('androidDevice.launchBrowser should pass args with spaces', async ({ androidDevice }) => {
-  const context = await androidDevice.launchBrowser({ args: ['--user-agent=I am Foo'] });
-  const page = await context.newPage();
-  const userAgent = await page.evaluate(() => navigator.userAgent);
-  await context.close();
-  expect(userAgent).toBe('I am Foo');
```
mxschmitt, line 40: this test was not passing before. 
Expected: I am Foo
Actual: I

packages/playwright-core/src/server/android/android.ts
```diff
@@ -264,9 +264,13 @@ export class AndroidDevice extends SdkObject {
...
+      if (!arg.includes('\'') && !arg.includes('"'))
+        return `"${arg}"`;
```
pavelfeldman, line 269: What are you trying to do here, quote the strings? If so, you should quote and escape.

packages/playwright-core/src/server/android/android.ts
```diff
@@ -264,9 +264,13 @@ export class AndroidDevice extends SdkObject {
...
+      return arg;
+    }).join(' ');
     debug('pw:android')('Starting', pkg, commandLine);
-    await this._backend.runCommand(`shell:echo "${commandLine}" > /data/local/tmp/chrome-command-line`);
+    await this._backend.runCommand(`shell:echo "${commandLine.replace(/"/g, '\\"')}" > /data/local/tmp/chrome-command-line`);
```
pavelfeldman, line 273: You seem to be double-escaping here?

mxschmitt: We need to bash escape in the end, since we `echo` so quotes would not have ended up inside the `/data/local/tmp/chrome-command-line`.

mxschmitt: using base64 now and added a comment.

pavelfeldman (approved):

PR #19428, opened 2022-12-13
test: fix transport tests

mxschmitt:
Transport tests were failing since https://github.com/microsoft/playwright/commit/256e9fd443c0e67dc6a42d3f1e0c40680dc7a5f1. I found two ways of fixing it:
a) dc6ed838cf9e1cf7642e288d10125f8bff9f1b43 aka. use env vars to set connect options and network filter
b) b3748d44ee31fa6e2f65666b910dafb382fbb443 aka. pass a private connect option for the network filter

~used b) since it has a smaller diff.~ did a) now

Can be repro'd via `WTEST_MODE=service  npm run test -- --project='Chromium page tests' --reporter=list page/elementhandle-click.spec.ts:20`.

### DIFF
--- a/tests/library/playwright.config.ts
+++ b/tests/library/playwright.config.ts
@@ -81,11 +81,8 @@ if (mode === 'service') {
       PWTEST_UNSAFE_GRID_VERSION: '1',
     },
   };
-  config.use.connectOptions = {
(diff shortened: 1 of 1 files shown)


yury-s (commented):
Let's use env variables instead as they are easier to track in the code? I can only guess what _exposeNetwork does.

mxschmitt:
Not relevant anymore.

PR #19442, opened 2022-12-14
chore: hide setup, store, TestProject.setupMatch, storageStateName

yury-s:


packages/playwright-test/types/test.d.ts
```diff
@@ -3393,7 +3351,6 @@ export default test;
 
 export const _baseTest: TestType<{}, {}>;
 export const expect: Expect;
-export const store: TestStore;
```
dgozman, line 3396: Notably, we did not export `setup` from these types - we should make sure we do after reland.

dgozman (approved):

PR #19463, opened 2022-12-15
feat(api): introduce expect().toPass

pavelfeldman:


packages/playwright-test/src/matchers/matchers.ts
```diff
@@ -309,3 +311,51 @@ export async function toBeOK(
...
+  options: {
+    intervals?: number[];
+    timeout?: number,
+  } = {},
+) {
+  let matcherError: Error | undefined;
+  const startTime = monotonicTime();
+  const pollIntervals = options.intervals || [100, 250, 500, 1000];
+  const lastPollInterval = pollIntervals[pollIntervals.length - 1] || 1000;
+  const timeout = currentExpectTimeout(options);
```
dgozman, line 327: As discussed, this should be something else.

docs/src/test-assertions-js.md
```diff
@@ -150,3 +150,27 @@ await expect.poll(async () => {
...
+You can retry blocks of code until they are passing successfully.
+
+```js
+await expect(async () => {
+  const response = await page.request.get('https://api.example.com');
+  expect(response.status()).toBe(200);
+}).toPass();
+```
+
+You can also specify custom retry intervals:
```
dgozman, line 165:
```suggestion
You can also specify custom timeout or retry intervals:
```

utils/generate_types/overrides-test.d.ts
```diff
@@ -293,14 +293,20 @@ type MakeMatchers<R, T> = BaseMatchers<R, T> & {
...
   } & SnapshotAssertions &
   ExtraMatchers<T, Page, PageAssertions> &
   ExtraMatchers<T, Locator, LocatorAssertions> &
-  ExtraMatchers<T, APIResponse, APIResponseAssertions>;
+  ExtraMatchers<T, APIResponse, APIResponseAssertions> &
+  ExtraMatchers<T, Function, {
+    /**
+     * Retries the callback until it passes.
+     */
+     toPass(options?: { timeout?: number, intervals?: number[] }): Promise<void>;
```
dgozman, line 308: It would be nice to add a test, see `expect-poll.spec:44` for example.

packages/playwright-test/src/matchers/matchers.ts
```diff
@@ -309,3 +311,51 @@ export async function toBeOK(
   const pass = response.ok();
   return { message, pass };
 }
+
+export async function toPass(
+  this: ReturnType<Expect['getState']>,
+  callback: () => Promise<void>,
+  options: {
+    intervals?: number[];
```
dgozman, line 319: Missing custom `message`?

dgozman (approved):

packages/playwright-test/src/matchers/matchers.ts
pavelfeldman: Custom messages are handled on the higher level.

PR #19500, opened 2022-12-16
docs(class-page.md): fixed mistake to use API

kostyabushuev:
Fixed mistake to use API in `class-page.md`


docs/src/api/class-page.md
```diff
@@ -1463,7 +1463,7 @@ Console.WriteLine(await page.EvaluateAsync<int>("1 + 2")); // prints "3"
 [ElementHandle] instances can be passed as an argument to the [`method: Page.evaluate`]:
 
 ```js
-const bodyHandle = await page.evaluate('document.body');
```
pavelfeldman, line 1466: It was right before.

PR #19577, opened 2022-12-20
feat: introduce toHaveCountGreaterThan and toHaveCountLessThan

### LINKED ISSUE #19381: [Feature] add `toHaveCountGreaterThan` and `toHaveCountLessThan` locator assertions
Playwright currently offers the `LocatorAssertions.toHaveCount` function, that allows to expect -with wait- for a locator to have an exact count.
```ts
await expect(someLocator).toHaveCount(1);
```
However, it doesn't provide an out of the box way for asserting on non-exact counts.

The proposal here is to add the following additional functions:
```ts
await expect(todoLocator).toHaveCountGreaterThan(2);
await expect(recentTodoLocator).toHaveCountLessThan(4);
```

See related: https://github.com/microsoft/playwright/issues/14278#issuecomment-1343970254

aslushnikov:
These are used like this:

```ts
await expect(list).toHaveCountGreaterThan(3);
await expect(list).toHaveCountLessThan(3);
```

Note: these are strict assertions. We do not add the
similar `toHaveCountGreaterOrEqualThan` yet since these
seem to be too verbose.

Fixes #19381


docs/src/api/class-locatorassertions.md
```diff
@@ -1162,6 +1163,51 @@ Expected count.
...
+- `count` <[int]>
+
+Expected count.
+
+### option: LocatorAssertions.toHaveCountGreaterThan.timeout = %%-js-assertions-timeout-%%
+* since: v1.30
+
+## async method: LocatorAssertions.toHaveCountLessThan
+* since: v1.30
+* langs: js
```
mxschmitt, line 1190: Language ports?

PR #19579, opened 2022-12-20
fix(recoder): get iframe selector while inspeting

csbun:
Let's say we have a web page like:

```html
<html>
<body>
  <div>MY_ITEM</div>
  <iframe>
    <html>
    <body>
      <div>MY_ITEM_IN_IFRAME</div>
    </body>
    </html>
  </iframe>
</body>
</html>
```

When using `Pick locator` with playwright-vscode, selecting `MY_ITEM_IN_IFRAME` we get a locator `getByText('MY_ITEM_IN_IFRAME')`, which can not be found in

```js
await page.getByText('MY_ITEM_IN_IFRAME');
```

what we want is `frameLocator('iframe').first().getByText('MY_ITEM_IN_IFRAME')`, so we can use as

```js
await page.frameLocator('iframe').first().getByText('MY_ITEM_IN_IFRAME');
```


packages/playwright-core/src/server/recorder.ts
```diff
@@ -178,7 +178,8 @@ export class Recorder implements InstrumentationListener {
     });
 
     await this._context.exposeBinding('__pw_recorderSetSelector', false, async (_, selector: string) => {
-      await this._recorderApp?.setSelector(selector, true);
+      const selectorWithFrame = _.frame.parentFrame() ? `iframe >> nth=0 >> internal:control=enter-frame >> ${selector}` : selector;
```
pavelfeldman, line 181: > iframe >> nth=0

That's not accurate - we would need exact frame locator along with the tests for this feature!

csbun: Yes, since I can not find any locator for iframe, I just hard code `iframe >> nth=0`.
Will we pass the frame locator within `selector` or a new argument?

pavelfeldman: I'm not sure I understand the question!

pavelfeldman:
Closing as stale

PR #19628, opened 2022-12-21
fix: bring back some browser builds on Ubuntu 18.04

aslushnikov:
Despite end-of-lifing Ubuntu 18.04 support, we can do a best effort
to bring *some* browser builds.

This patch:
- Brings back Chromium on Ubuntu 18.04
- Brings back Firefox on Ubuntu 18.04
- Sets up a stab for WebKit on Ubuntu 18.04, similar to the stabs we use
  for MacOS 10.14 and MacOS 10.15


aslushnikov:
Closing in favor of Max's patch

Issue #19615, opened 2022-12-21
[BUG]  TypeError: Cannot read properties of undefined (reading 'fullPage')


jhuh0821:
**Context:**
- Playwright Version: 1.29.0 (and @playwright/test version is also 1.29.0)
- Operating System: Linux
- Node.js version: 16
- Browser: Chromium
- Extra: GitHub Actions

It works well in my local.
But running playwright tests failed on GitHub actions with following message:

<img width="752" alt="image" src="https://user-images.githubusercontent.com/64876995/208871322-7eea083f-57cd-4f9b-a9fe-5b36673debce.png">

pavelfeldman:
Is there something we could reproduce?

pavelfeldman:
Do you set screenshots: property in config or anywhere else?

jhuh0821:
Yes, there is screenshots option in config.
![image](https://user-images.githubusercontent.com/64876995/209035434-3a934207-8aac-40f0-a7f4-0bd64a9ad611.png)

pavelfeldman:
As a workaround, you can replace `undefined` with `off`. This is fixed in 1.29.1

PR #19632, opened 2022-12-21
fix(test runner): make sure undefined options in config result in default values

dgozman:
Note: this keeps existing behavior of `undefined` in `test.use()` reverting to the config value and not to the original default value.

References #19615.


pavelfeldman (approved):
Curious why this applies to type option only.

PR #19640, opened 2022-12-22
chore: report more fatal errors via reporter

pavelfeldman:


packages/playwright-test/src/loader.ts
```diff
@@ -72,6 +78,10 @@ export class Loader {
     return {};
   }
 
+  addLoadError(message: string, location: Location) {
```
dgozman, line 81: Unused?

dgozman (approved):

PR #19657, opened 2022-12-22
test(ct): vue vite slice by feature

sand4rt:
@pavelfeldman Could you provide feedback before i proceed slicing the tests for the other frameworks? And do you prefer everything in one PR or in separate ones?

related to https://github.com/microsoft/playwright/pull/18765


tests/components/ct-vue-vite/tests/events/events.spec.js
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from '@playwright/experimental-ct-vue';
+import Button from '@/components/Button.vue';
+
+test('emit an submit event when the button is clicked', async ({ mount }) => {
```
pavelfeldman, line 4:
```suggestion
test('emit a submit event when the button is clicked', async ({ mount }) => {
```

tests/components/ct-vue-vite/tests/events/events.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from '@playwright/experimental-ct-vue';
+import Button from '@/components/Button.vue';
+
+test('emit an submit event when the button is clicked', async ({ mount }) => {
```
pavelfeldman, line 4:
```suggestion
test('emit a submit event when the button is clicked', async ({ mount }) => {
```

tests/components/ct-vue-vite/tests/events/events.spec.js
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from '@playwright/experimental-ct-vue';
```
pavelfeldman, line 1: I don't think we need to go all the way in ts/js and tsx/jsx parity. It would be sufficient to check that all notations work, but no need for functional changes for all of them.

pavelfeldman (approved):
I like the groups that you outlined, this looks fine to land and we can mirror this for others.

tests/components/ct-vue-vite/tests/events/events.spec.js
sand4rt: Thank you for the feedback! 

I'm not completely sure how you envision it. Was thinking of keeping all the `TS`/`TSX` tests both for documentation and to make sure everything works as expected. Perhaps keep `render.js` tests to make sure that works as expected as well?

sand4rt: BTW. there are also other ways to speed up the pipeline if they need to be removed for performance reasons. Like using pnpm and or maybe a tool like turborepo. When one of the component test projects is modified, the entire pipeline is executed, which i don't think is necessary.

pavelfeldman: We don't use tests for documentation - doing that typically compromises both in the end. So we should have tests for coverage and examples for the documentation. Since this is about the tests, all we care about is coverage. Which means we don't need to repeat all the tests in all the dialects, unless that triggers paths we want to test.

pavelfeldman: Considering your router tests as an example - once we've established that we can send hooks config in both JS and JSX notations, we don't need to test the routes twice. I.e. testing that config is plumbed in both notations and then testing routes using only one notation is sufficient. It is not about the speed of execution, more about the identical tests diverging over time.

Happy New Year btw!

sand4rt: Thank you for the replies! Happy new year!

pavelfeldman (approved):

PR #19659, opened 2022-12-22
docs: test runner first inside Node.js docs

mxschmitt:


docs/src/emulation.md
```diff
@@ -425,7 +425,7 @@ Allow notifications for a specific domain.
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
 const config = {
   use: {
-    permissions: ['notifications'], {origin: 'https://skype.com'},
+    permissions: ['notifications'],
```
dgozman, line 428: This is now not an example for "Allow notifications for a specific domain." :)

dgozman (approved):

docs/src/emulation.md
mxschmitt: maybe we should add a beforeAll with context.grandPermissions instead?

dgozman: Sure!

Issue #19655, opened 2022-12-22
[BUG] 1.28+ breaks tests for fancytree


vincchan:
**Context:**
- Playwright Version: 1.28+
- Operating System: Windows
- Node.js version: 16.17
- Browser: Chromium

**Codegen generated following test for 1.27.1**

```javascript
import { test, expect } from '@playwright/test';

test('test', async ({ page }) => {
  await page.goto('http://wwwendt.de/tech/fancytree/demo/sample-api.html');
  await page.getByRole('treeitem', { name: 'This simple node (and the following) have been created from html.' }).getByRole('checkbox').click();
  await page.getByRole('treeitem', { name: 'item 4. Note that also non-folders (i.e. \'documents\') may have child nodes' }).getByRole('button').click();
  await page.getByRole('treeitem', { name: 'Sub-item 4.1' }).getByRole('checkbox').click();
  await page.getByRole('treeitem', { name: 'Sub-item 4.3' }).getByRole('button').click();
  await page.getByRole('treeitem', { name: 'Sub-item 4.3.1' }).getByRole('checkbox').click();
  await page.getByRole('treeitem', { name: 'Advanced examples item5.1: Using data attribute as an alternative way to specify a key. item5.1: Using data attribute as an alternative way to specify a folder. Sub-item 5.2 Item without a key. Keys are optional (generated automatically), but may be used in the callbacks' }).getByRole('checkbox').first().click();
  await page.getByRole('treeitem', { name: 'item5.1: Using data attribute as an alternative way to specify a key.' }).getByRole('checkbox').click();
  await page.getByRole('treeitem', { name: 'item5.1: Using data attribute as an alternative way to specify a folder.' }).getByRole('checkbox').click();

});
```

**Codegen generated following test for 1.28**

```javascript
import { test, expect } from '@playwright/test';

test('test', async ({ page }) => {
  await page.goto('http://wwwendt.de/tech/fancytree/demo/sample-api.html');
  await page.getByRole('treeitem', { name: 'This simple node (and the following) have been created from html.' }).getByRole('checkbox').click();
  await page.getByRole('treeitem', { name: 'item 4. Note that also non-folders (i.e. \'documents\') may have child nodes' }).getByRole('button').click();
[...]

PR #19672, opened 2022-12-23
fix(generator): do not contribute to aria name, penalize for length

pavelfeldman:
Affects https://github.com/microsoft/playwright/issues/19655

Screenshot 2022-12-22 at 5 24 14 PM

### DIFF
--- a/packages/playwright-core/src/server/injected/selectorGenerator.ts
+++ b/packages/playwright-core/src/server/injected/selectorGenerator.ts
@@ -28,19 +28,33 @@ type SelectorToken = {
 const cacheAllowText = new Map<Element, SelectorToken[] | null>();
 const cacheDisallowText = new Map<Element, SelectorToken[] | null>();
 
+const kTextScoreRange = 10;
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/injected/roleUtils.ts
```diff
@@ -590,6 +617,11 @@ function getElementAccessibleNameInternal(element: Element, options: AccessibleN
       if (skipSlotted && (node as Element | Text).assignedSlot)
         return;
       if (node.nodeType === 1 /* Node.ELEMENT_NODE */) {
+        // Do not contribute from selected roles.
+        const role = getAriaRole(node as Element);
+        if (role && doNotContributeToParentName.has(role))
```
dgozman, line 622: I think we should move this around line 612 instead. Otherwise, IIUC, we entirely skip children that have one of these roles and `aria-label` or `aria-labelledby`, while Blink does not.

dgozman (approved):

PR #19690, opened 2022-12-24
chore: update to vite 4

sand4rt:


packages/playwright-test/src/plugins/vitePlugin.ts
```diff
@@ -117,7 +117,7 @@ export function createPlugin(
       const { build, preview } = require('vite');
       // Build config unconditionally, either build or build & preview will use it.
       viteConfig.plugins = viteConfig.plugins || [
-        frameworkPluginFactory()
+        await frameworkPluginFactory()
```
sand4rt, line 120: Made it asynchronous because of: https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#how-can-i-use-vite-plugin-svelte-from-commonjs and to keep it consistant

pavelfeldman (approved):

PR #19691, opened 2022-12-24
feat: support sharding in HTML Reporter

### LINKED ISSUE #10437: [Feature] Combine sharding with html reporter
Currently there is no way to combine test results into one [html report,](https://playwright.dev/docs/test-reporters#html-reporter) while using [sharding](https://playwright.dev/docs/test-parallel#shard-tests-between-multiple-machines) on multiple machines. (v1.17.0-rc1)

I end up with multiple reports which makes it harder for sharing with others in case of failures.

It would be nice to be able to combine multiple executions into 1 report.

I believe anyone who wants to scale tests horizontally might benefit from such feature

kevin940726:
Close https://github.com/microsoft/playwright/issues/10437.

This feature aims to bring official support to combine sharded reports in the HTML reporter. This is probably not the best solution, but I think it works with minimal changes.

## How it works

1. We inline the `shard` config into a global variable along with other metadata for future usage called `window.playwrightMetadata`.
2. Instead of inlining the ZIP file into a base64 string, we write the ZIP into a `report.zip` file. If there's sharding, we instead write them into a `report-{shardIndex}.zip` file.
3. After each run, we typically upload the generated report directory somewhere to an artifact. Once every part is finished, we download and extract them to a single place. Everything except for the `report-{shardIndex}.zip` files should be overridden and replaced.
4. In the web app, instead of reading the ZIP from a base64 string, we read them from HTTP. If there's sharding, we read the reports one by one and merge them together.

## Comparing to [`playwright-merge-html-reports`](https://github.com/Rippling/playwright-merge-html-reports)

The `mergeStats` helper is largely inspired by `playwright-merge-html-reports`, but I think this approach has some advantages:

1. Official support.
2. No extra steps are needed, no need to run another script to combine the reports. This can be tedious for CI as they have to initialize an environment, npm install, download reports, run the scripts, then finally upload the report.
3. Works out of the box, no need for another API.

## Example setup

GitHub Actions with artifacts

```yml
steps:
  playwright:
    name: 'Playwright Tests - ${{ matrix.project }} - Shard ${{ matrix.shardIndex }} of ${{ matrix.shardTotal }}'
    runs-on: ubuntu-latest
    container:
      image: mcr.microsoft.com/playwright:v1.29.0-focal
    strategy:
      fail-fast: false
      matrix:
        project: [chromium, webkit]
        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
        shardTotal: [10]
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-node@v3
        with:
          node-version: '18'
      - name: Install dependencies
        run: npm ci
      - name: Run your tests
        run: npx playwright test --project=${{ matrix.project }} --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }}
      - uses: actions/upload-artifact@v3
        with:
          name: playwright-report
          path: playwright-report
```

WDYT? Once we're happy with the approach we can start writing some docs and tests.

### DIFF
--- a/packages/html-reporter/src/index.tsx
+++ b/packages/html-reporter/src/index.tsx
@@ -23,6 +23,7 @@ import * as ReactDOM from 'react-dom';
 import './colors.css';
 import type { LoadedReport } from './loadedReport';
 import { ReportView } from './reportView';
+import { mergeReports } from './mergeReports';
 // @ts-ignore
 const zipjs = zipImport as typeof zip;
 
@@ -46,10 +47,27 @@ class ZipReport implements LoadedReport {
   private _json!: HTMLReport;
 
   async load() {
-    const zipReader = new zipjs.ZipReader(new zipjs.Data64URIReader((window as any).playwrightReportBase64), { useWebWorkers: false }) as zip.ZipReader;
-    for (const entry of await zipReader.getEntries())
-      this._entries.set(entry.filename, entry);
-    this._json = await this.entry('report.json') as HTMLReport;
+    const metadata = (window as any).playwrightMetadata;
+    if (metadata.shard) {
+      const reports: HTMLReport[] = [];
+      for (let index = 1; index <= metadata.shard.total; index += 1) {
+        try {
+          const zipReader = new zipjs.ZipReader(new zipjs.HttpReader(`/report/report-${index}.zip`), { useWebWorkers: false }) as zip.ZipReader;
+          for (const entry of await zipReader.getEntries())
+            this._entries.set(entry.filename, entry);
+          const currentJson = await this.entry(`report-${index}.json`) as HTMLReport;
+          reports.push(currentJson);
+        } catch (error) {
+          // Ignore not found error for viewing individual shard report.
+        }
+      }
+      this._json = mergeReports(reports);
+    } else {
+      const zipReader = new zipjs.ZipReader(new zipjs.HttpReader('/report/report.zip'), { useWebWorkers: false }) as zip.ZipReader;
+      for (const entry of await zipReader.getEntries())
+        this._entries.set(entry.filename, entry);
+      this._json = await this.entry('report.json') as HTMLReport;
+    }
   }
 
   json(): HTMLReport {
--- a/packages/playwright-test/src/reporters/html.ts
+++ b/packages/playwright-test/src/reporters/html.ts
@@ -18,8 +18,6 @@ import { colors } from 'playwright-core/lib/utilsBundle';
 import fs from 'fs';
 import { open } from '../utilsBundle';
 import path from 'path';
-import type { TransformCallback } from 'stream';
-import { Transform } from 'stream';
 import type { FullConfig, Suite } from '../../types/testReporter';
 import { HttpServer } from 'playwright-core/lib/utils/httpServer';
 import { assert, calculateSha1, monotonicTime } from 'playwright-core/lib/utils';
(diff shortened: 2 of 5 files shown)


kevin940726:

packages/html-reporter/src/index.tsx
```diff
@@ -46,10 +47,27 @@ class ZipReport implements LoadedReport {
...
-    const zipReader = new zipjs.ZipReader(new zipjs.Data64URIReader((window as any).playwrightReportBase64), { useWebWorkers: false }) as zip.ZipReader;
-    for (const entry of await zipReader.getEntries())
-      this._entries.set(entry.filename, entry);
-    this._json = await this.entry('report.json') as HTMLReport;
+    const metadata = (window as any).playwrightMetadata;
+    if (metadata.shard) {
+      const reports: HTMLReport[] = [];
+      for (let index = 1; index <= metadata.shard.total; index += 1) {
+        try {
+          const zipReader = new zipjs.ZipReader(new zipjs.HttpReader(`/report/report-${index}.zip`), { useWebWorkers: false }) as zip.ZipReader;
```
pavelfeldman, line 55: Let's keep those files siblings in order to be less intrusive, hard-code less names. So it'll be
```js
new zipjs.HttpReader(`report-${index}.zip`)
```

packages/html-reporter/src/index.tsx
```diff
@@ -46,10 +47,27 @@ class ZipReport implements LoadedReport {
...
+          for (const entry of await zipReader.getEntries())
+            this._entries.set(entry.filename, entry);
+          const currentJson = await this.entry(`report-${index}.json`) as HTMLReport;
+          reports.push(currentJson);
+        } catch (error) {
+          // Ignore not found error for viewing individual shard report.
```
pavelfeldman, line 61: Let's make sure these errors end up in devtools so that it was clear what is going on.

packages/html-reporter/src/index.tsx
```diff
@@ -46,10 +47,27 @@ class ZipReport implements LoadedReport {
...
-    for (const entry of await zipReader.getEntries())
-      this._entries.set(entry.filename, entry);
-    this._json = await this.entry('report.json') as HTMLReport;
+    const metadata = (window as any).playwrightMetadata;
+    if (metadata.shard) {
+      const reports: HTMLReport[] = [];
+      for (let index = 1; index <= metadata.shard.total; index += 1) {
+        try {
+          const zipReader = new zipjs.ZipReader(new zipjs.HttpReader(`/report/report-${index}.zip`), { useWebWorkers: false }) as zip.ZipReader;
+          for (const entry of await zipReader.getEntries())
```
pavelfeldman, line 56: Looks like you could benefit from fetching them all concurrently instead.

packages/playwright-test/src/reporters/html.ts
```diff
@@ -289,17 +289,19 @@ class HtmlBuilder {
       }
     }
 
-    // Inline report data.
+    // Inline metadata.
     const indexFile = path.join(this._reportFolder, 'index.html');
-    fs.appendFileSync(indexFile, '<script>\nwindow.playwrightReportBase64 = "data:application/zip;base64,');
+    fs.appendFileSync(indexFile, `<script>\nwindow.playwrightMetadata = ${JSON.stringify(metadata)}\n</script>`);
```
pavelfeldman, line 294: We would need both, the new data and the base64 placeholder.

pavelfeldman (commented):
Looks great! Sorry for the late review, just returned from the break.

dgozman:
@kevin940726 Heads up: I will pick this up, create a new PR and attribute existing work in the description. This way it will be easier to manage and land the feature.

kevin940726:
@dgozman Sure! I'll be away until next week anyway (LNY), so feel free to do anything with this PR :). Also thank you @pavelfeldman for the review!

mxschmitt:
Looks like got implemented in https://github.com/microsoft/playwright/pull/20700. Closing for now then and thanks for the contribution!

Issue #19437, opened 2022-12-13
[BUG] Unable to resolve TsConfig Aliases in Playwright Component Testing for React


bdlb77:
Currently we are unable to resolve our `tsconfig.json` Alias paths that we set. 

The error we receive is:

<img width="1218" alt="Screenshot 2022-12-13 at 2 57 54 PM" src="https://user-images.githubusercontent.com/32174912/207463125-0e4bd1ed-86fe-49b9-95f5-2565614e22ab.png">

**Context:**
- Playwright Version: v.1.28.1
- Operating System: Mac Ventura 13.0.1
- Node.js version: 16.17.1
- Browser: All

**Extra**

We are using both e2e Playwright tests, and component Testing. Here are the playwright packages being used:

```json
    "@playwright/experimental-ct-react": "^1.28.1",
    "@playwright/test": "^1.28.1",
    "playwright": "^1.28.1",
```

**Code Snippet**
Component to be tested (it's a basic sample one)

```tsx
// src/renderer/components/sample.tsx
import React from "react";

const MyComponent: React.FC = () => {
    return <h1>Hello World</h1>;
};

export default MyComponent;
```
This works:

```tsx
import MyComponent from "../../../src/renderer/components/sample";
test("should work", async ({ mount }) => {
    const component = await mount(<MyComponent />);
    await expect(component).toContainText("Greetings");
});
```

This Does Not work:

```tsx
import MyComponent from "~Components/sample";

test("should work", async ({ mount }) => {
    const component = await mount(<MyComponent />);
    await expect(component).toContainText("Greetings");
});
```
Here is the `tsconfig.json` paths definition

```json
    "paths": {                                    
      "~Components/*": ["src/renderer/components/*"],
    },
```
**Describe the bug**

As Stated above, it seems to be an issue with resolving `tsconfig.json` specifically in the Experimental Testing Components library. 
Our aliases are resolving fine in our `e2e` testing suite using playwright. 

We have 2 playwright configs:
-  1 for `e2e`
- 1 for **react component testing**. The generated one through `yarn create playwright --ct` is what we use, besides changes to the `testDir`

```ts
const config: PlaywrightTestConfig = {
    testDir: "./__tests__/renderer",
   // ...
}

**Comment by sand4rt** [comment]
It seems you forgot to tell vite that you are using import aliases:

```
// playwright-ct.config.ts
import { type PlaywrightTestConfig, devices } from '@playwright/experimental-ct-react'

import { resolve } from 'path'

[...]

Issue #17846, opened 2022-10-05
[Question] Component Tests perform a different bundling? Failed to resolve asset import


Nithos:
I have a simple component for a header which just includes a logo and button.
But running the component test fails as the vite bundler gets invoked and complains due to the reference to the image.
Does it not resolve aliases properly?
When using a relative path it works.

Fails
`

	<nav>
		<img src="@assets/images/logo.png" height="32" width="128" alt="logo" />
		{{ auth.name }}
		<Button v-if="isAuthenticated" @click="logout">Logout <Logout /></Button>
		<Button v-else @click="login">Login <Login /></Button>
	</nav>
`
ERROR:
[vite]: Rollup failed to resolve import "@assets/images/logo.png" from "src/components/Header.vue?vue&type=script&setup=true&lang.ts".

Works with a relative path
`

	<nav>
		<img src="../assets/images/logo.png" height="32" width="128" alt="logo" />
		{{ auth.name }}
		<Button v-if="isAuthenticated" @click="logout">Logout <Logout /></Button>
		<Button v-else @click="login">Login <Login /></Button>
	</nav>
`

sand4rt:
Hi @Nithos, you have to specify your import alias in `playwright-ct-config.ts` as well:

```typescript
import { resolve } from 'path';
const config: PlaywrightTestConfig = {
  use: {
    trace: 'on-first-retry',
    ctPort: 3100,
    ctViteConfig: {
      resolve: {
        alias: {
          '@': resolve(__dirname, './src'),
        },
      },
    },
  },
};
```

Playwright doesn't reuse the Vite config yet: https://github.com/microsoft/playwright/issues/14749.

Nithos:
Thank you @sand4rt that was an omission on my part. When I first configured it I added our main alias but forgot about the additional ones :(. That solves my issue.

PR #19693, opened 2022-12-24
feat(ct): merge vite config if exists

### LINKED ISSUE #14749: [Feature]Component testing: reuse the same vite server and same vite config file
Since my application already uses vite server, I wonder if we can use the same vite sever for both local development and component testing.
and It would be great if we can put the component test configuration in the same vite config file.

sand4rt:
closes https://github.com/microsoft/playwright/issues/14749
related https://github.com/microsoft/playwright/issues/17846, https://github.com/microsoft/playwright/issues/19437, https://github.com/microsoft/playwright/issues/19934, https://github.com/microsoft/playwright/issues/19829

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "eslint": "^8.8.0",
     "eslint-plugin-notice": "^0.9.10",
     "eslint-plugin-react-hooks": "^4.3.0",
+    "find-up": "^6.3.0",
     "formidable": "^2.1.1",
     "license-checker": "^25.0.1",
     "mime": "^3.0.0",
--- a/packages/playwright-test/src/plugins/vitePlugin.ts
+++ b/packages/playwright-test/src/plugins/vitePlugin.ts
@@ -53,16 +53,18 @@ export function createPlugin(
       configDir = configDirectory;
       const use = config.projects[0].use as CtConfig;
       const port = use.ctPort || 3100;
-      const viteConfig: InlineConfig = use.ctViteConfig || {};
+      const ctViteConfig: InlineConfig = use.ctViteConfig || {};
       const relativeTemplateDir = use.ctTemplateDir || 'playwright';
 
-      const rootDir = viteConfig.root || configDir;
+      const rootDir = ctViteConfig.root || configDir;
       const templateDir = path.join(rootDir, relativeTemplateDir);
-      const outDir = viteConfig?.build?.outDir || (use.ctCacheDir ? path.resolve(rootDir, use.ctCacheDir) : path.resolve(templateDir, '.cache'));
+      const outDir = ctViteConfig?.build?.outDir || (use.ctCacheDir ? path.resolve(rootDir, use.ctCacheDir) : path.resolve(templateDir, '.cache'));
 
       const buildInfoFile = path.join(outDir, 'metainfo.json');
       let buildExists = false;
       let buildInfo: BuildInfo;
+
+      const viteConfig = await mergeViteConfig(rootDir, ctViteConfig);
 
       const registerSource = await fs.promises.readFile(registerSourceFile, 'utf-8');
       const registerSourceHash = calculateSha1(registerSource);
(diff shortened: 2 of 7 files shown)


tests/components/ct-react-vite/vite.config.ts
```diff
@@ -1,7 +1,13 @@
...
+import { fileURLToPath, URL } from 'url';
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [react()]
-})
+  plugins: [react()],
+  resolve: {
+    alias: {
+      '@': fileURLToPath(new URL('./src', import.meta.url))
```
sand4rt, line 10: Note: works without specifying in `playwright.config.ts`

sand4rt:
Is the failing test correct? Not sure why not to use global config for preview?

package.json
```diff
@@ -83,6 +83,7 @@
     "eslint": "^8.8.0",
     "eslint-plugin-notice": "^0.9.10",
     "eslint-plugin-react-hooks": "^4.3.0",
+    "find-up": "^6.3.0",
```
pavelfeldman, line 86: It looks like this is your prod dependency, not dev. It would need to go to packages/playwright-test/bundles/utils/package.json. Better yet implement it in fileUtils, it is a two-liner.

packages/playwright-test/src/plugins/vitePlugin.ts
```diff
@@ -53,17 +53,19 @@ export function createPlugin(
...
+      const rootDir = ctViteConfig.root || configDir;
       const templateDir = path.join(rootDir, relativeTemplateDir);
-      const outDir = viteConfig?.build?.outDir || (use.ctCacheDir ? path.resolve(rootDir, use.ctCacheDir) : path.resolve(templateDir, '.cache'));
+      const outDir = ctViteConfig?.build?.outDir || (use.ctCacheDir ? path.resolve(rootDir, use.ctCacheDir) : path.resolve(templateDir, '.cache'));
 
       const buildInfoFile = path.join(outDir, 'metainfo.json');
       let buildExists = false;
       let buildInfo: BuildInfo;
 
+      const viteConfig = await mergeViteConfig(rootDir, ctViteConfig);
```
pavelfeldman, line 67: I am not sure doing this unconditionally is a good idea, my production config could have steps that are not relevant (destructive) to my testing.

sand4rt: Hi thank you for the reviews!:) Was thinking about that too. Then i thought perhaps missing configuration can even be more destructive and people will probably run a lot quicker into that. Like missing plugins, import aliases and so on _(see related tickets)_. It might also be better to keep the build config as close as possible to the real build configuration unless that causes problems. Then it's possible to override the `vite.config.ts` by defining it in `playwright.config.ts`. 

Vitest and Cypress also read the `vite.config.ts` if it exists. However it's only possible to define/use the config in one place. If the config is put their own config file, the `vite.config.ts` will be ignored. Perhaps that's a better option than working with overrides like i've done now?

pavelfeldman: > Then i thought perhaps missing configuration can even be more destructive and people will probably run a lot quicker into that. Like missing plugins, import aliases and so on (see related tickets).

They way I think about it is as follows:
- We build components using vite under the hood. It is an implementation detail of Playwright component testing story. Users could be using vite 3 with 3-specific config and we will be using Vite 4, and vice versa. Or they will be using non-vite altogether. We can't force users to bump their vite version and we can't be dependent on their vite version.
- Now that we've established this baseline, we start working on the optimizations that improve our ability to work out of the box. It means that we should now be able to cherry-pick important bits of the configuration from the existing configs. So we should respect babel/webpack aliases, cherry pick bits from vite config, etc. And while doing so, we should be mindful that the user can have any bundler. That's why I think that while bulk-importing production vite config brings up closer to the vite user, conceptually, that's not how we want to organize better support for the existing configuration of the generic user.

pavelfeldman: In the light of the above, this whole ctViteConfig is a layering violation that was used to jump-start non-trivial amount of the users that could benefit from it immediately. But overall, we should be able to jump the bundler ship and allow users to do so.

sand4rt: Thank you for the explanation! I get your point. I'm still a bit concerned about having a separate bundler and config for test and for prod. For example, what if a Webpack plugin is used which is not supported in Vite? What if the tests are passing while prod is broken? I have not encountered this yet and i think time will tell if this is a real problem. Will close the PR.

Issue #15239, opened 2022-06-29
[Feature] Amount of  `fixme` vs `skip` in report


dimkin-eu:
Tests marked with `.skip` and `.fixme` 
- both are skipped
- both are counted at the end as `skipped` 
- hides the number of tests to be fixed.

Having a big permutation of tests ( this mostly affects screenshot testing ) we skip a lot of them ( like some combinations don't make sense to test), but skipped tests are shown as `skipped`. 

Could be nice to not include the `.skip`-ed tests at the end, but only show tests with `.fixme` which are failing/flaky.
Showing both in separate lines also could be an option.

![Screenshot 2022-06-29 at 19 29 54](https://user-images.githubusercontent.com/7899764/176488690-9ec8364c-26f4-476f-b541-18a6f556b45e.png)

As a workaround we are doing this way, but meh
```js
    someStorybook.forEach((story) => {
        test(story.name, async ({}, testInfo) => {
            if (testInfo.project.name.startsWith("Desktop-") && story.name.endsWith("foobar Story")) {
                return
            }
            await page.goto(storybookLocalhost + story.url)
            await compareScreenshots(page, story.name)
        })
    })
```

rwoll:
Thanks for the report! I've included a workaround below that I think gives the info you need. (We'll have to think more deeply about changing the built-in reporter as some folks may already depend on the format of it's output in, for example, CI/CD reporting. But hopefully this helps in the interim!)

**Let me know if the below unblocks you!**

With the workaround, the output looks like:

```
$ npx playwright test --reporter=line,./DetailedSkipInfoReporter.ts

Running 6 tests using 6 workers

  6 skipped

  Skipped Breakdown: 3 skipped, 3 fixme
```

To get this output (the additional line with _Skipped Breakdown_), you can do the following:

1. Save/check-in the [Custom Reporter](https://playwright.dev/docs/test-reporters#custom-reporters) I've included at the end of this comment called `DetailedSkipInfoReporter.ts`.
2. Add the custom reporter at the end of the reporters array in your playwright config (or on the command line like `npx playwright test --reporter=line,./DetailedSkipInfoReporter.ts`)

```ts
// @filename: DetailedSkipInfoReporter.ts
import type { FullConfig, FullResult, Reporter, Suite } from '@playwright/test/reporter';

class DetailedSkipInfoReporter implements Reporter {
  private _suite!: Suite;

  onBegin(config: FullConfig, suite: Suite) {
    this._suite = suite;
  }

[...]

PR #19729, opened 2022-12-27
feat(html): skipped v. fixme filter

### LINKED ISSUE #19336: [Feature] separate `fixme` and `skip` in html report
Both test groups are shown inside 1 column. but could be nice to see them independent

```js
  176 skipped
  1912 passed (25m)
  Skipped Breakdown: 19 skipped, 157 fixme
```

related to this https://github.com/microsoft/playwright/issues/15239

rwoll:
Resolves #19336.
Relates #15239.

Visual Preview:

Screenshot 2022-12-27 at 1 11 37 PM

### DIFF
--- a/packages/html-reporter/src/filter.ts
+++ b/packages/html-reporter/src/filter.ts
@@ -97,8 +97,12 @@ export class Filter {
         status = 'failed';
       if (test.outcome === 'flaky')
         status = 'flaky';
-      if (test.outcome === 'skipped')
-        status = 'skipped';
+      if (test.outcome === 'skipped') {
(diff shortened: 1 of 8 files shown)


pavelfeldman:
Not convinced we want to land this. Here is my thinking: from the test run perspective, this test did not run, it was skipped. It could be skipped, or it could run. If it ran it could fail or not, it could fail as flaky.

Fixme vs Skipped is not a property of a test run, it is a property of your test suite, so as long as our html report just serves the test results purposes, it does not seem like we should surface it there.

I did not close the original issue because if a lot of people would want that, we would do that. But if not, maybe we could wait a little longer?

PR #19744, opened 2022-12-28
feat(reactSelectorEngine): extract names of React.memo/lazy/forwardRe…

hanayashiki:
…f for getComponentName

Previously, playwright's `_react` selector fails to match the following components:

```js
// in browser
const MemoizedForwardRef = React.memo(React.forwardRef(() => {
  return e('div', null, 'MemoizedForwardRef');
}));

MemoizedForwardRef.displayName = 'MemoizedForwardRef';

const Lazy = React.lazy(() => import('./react-lazy.js'));
```

```js
// ./react-lazy
export default function ReactLazy() {
  return e('div', null, 'ReactLazy');
};
```

```
"_react=MemoizedForwardRef"
"_react=Memoized"
"_react=ReactLazy"
```

That said, even if the user has the correct `displayName` displayed in the devtool, the user still cannot match the React element with `_react` selector.

This PR looks into the element's VDOM's `type` field and handles when it is not `class` or `function`. It does that according to the [source code of react](https://github.com/facebook/react/blob/main/packages/shared/getComponentNameFromType.js).

The selector is tested on react 16/17/18.

### DIFF
--- a/packages/playwright-core/src/server/injected/reactSelectorEngine.ts
+++ b/packages/playwright-core/src/server/injected/reactSelectorEngine.ts
@@ -31,6 +31,7 @@ type ReactVNode = {
   key?: any,
   // React 16+
   type: any,
+  elementType: any,
   child?: ReactVNode,
   sibling?: ReactVNode,
   stateNode?: Node,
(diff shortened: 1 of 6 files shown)


hanayashiki:

@microsoft-github-policy-service agree [company="{your company}"]

hanayashiki:
> @microsoft-github-policy-service agree

pavelfeldman:
Could you file a bug and describe the use case there? We are considering evicting react and vue selectors into a separate package due to limited use and cross-cutting aspect that looks at the implementation, so it is important for us to understand why users use them.

hanayashiki:
> Could you file a bug and describe the use case there? We are considering evicting react and vue selectors into a separate package due to limited use and cross-cutting aspect that looks at the implementation, so it is important for us to understand why users use them.

Hi, I think the reason is clearly stated in the PR.
Simply put, the original implementation doesn't consider exotic elements resulting from React.memo/forwardRef/lazy, so the user cannot match them like ordinary components, even if using there name just showing in the devtools.

pavelfeldman:
As per my suggestion, lets start with the issue.

PR #19753, opened 2022-12-28
docs: add instructions on msedge & chrome installation

### LINKED ISSUE #18511: [Feature] Add explanation about installation of MSEdge and Chrome
The guideline for the first installation and usage missing an explanation about the installation of MSEdge and Chrome browsers in case we need to add them manually. 
Also, I didn't find something convenient section that explains it.

I suggest adding this section, and in case I missed it - to add the explanation to the more convenient section.

yury-s:
Fixes #18511

### DIFF
--- a/docs/src/browsers.md
+++ b/docs/src/browsers.md
@@ -29,8 +29,9 @@ other WebKit-based browsers. This gives a lot of lead time to react on the poten
 ## Google Chrome & Microsoft Edge
 
 While Playwright can download and use the recent Chromium build, it can operate against the stock Google
-Chrome and Microsoft Edge browsers available on the machine. In particular, current Playwright version will
-support Stable and Beta channels of these browsers. Here is how you can opt into using the stock browser:
+Chrome and Microsoft Edge browsers available on the machine (note that Playwright doesn't install them by
+default). In particular, current Playwright version will support Stable and Beta channels of these browsers.
+Here is how you can opt into using the stock browser:
 
 ```js tab=js-js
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):

docs/src/browsers.md
```diff
@@ -95,6 +96,17 @@ var chromium = playwright.Chromium;
...
+### Installing Google Chrome & Microsoft Edge
+
+If stock Google Chrome or Microsoft Edge is not available on your machine, you can install
+them using Playwright command line tool:
+
+```bash
+npx playwright install msedge
+```
+
+Run `npx playwright install --help` to see full list of the browsers that can be installed this way.
```
aslushnikov, line 108: **DANGER**: Google Chrome and Microsoft Edge installations will not be isolated and will be installed in the default location on your operating system.

docs/src/browsers.md
```diff
@@ -95,6 +96,22 @@ var chromium = playwright.Chromium;
 var browser = await chromium.LaunchAsync(new BrowserTypeLaunchOptions { Channel = "chrome" });
 ```
 
+### Installing Google Chrome & Microsoft Edge
+
+If stock Google Chrome or Microsoft Edge is not available on your machine, you can install
+them using Playwright command line tool:
+
+```bash
```
mxschmitt, line 104: This should include the snippets for the other language ports

yury-s: good point!

PR #19754, opened 2022-12-28
chore: render readable title separator in errors

pavelfeldman:


packages/playwright-test/src/reporters/base.ts
```diff
@@ -380,14 +379,14 @@ export function formatError(config: FullConfig, error: TestError, highlightCode:
   const message = error.message || error.value || '';
   const stack = error.stack;
   if (!stack && !error.location)
-    return { message };
+    return { message: `Error: ${message}` };
```
dgozman, line 382: I am worried this will add unnecessary `Error` prefix to some errors:
- `Test timeout of 30000ms exceeded.`
- `Pending operations: ...`
- Probably more.

Also, isn't it common practice for the `Error:` prefix to be included? Especially when it's a custom error name like `TimeoutError: ...`.

PR #19755, opened 2022-12-28
fix: make sure jsx fragments are transformed

### LINKED ISSUE #19142: Playwright Typescript Monorepo JSX issue
**Context:**
- Playwright Version: 1.28.1
- Operating System: Mac
- Node.js version: v16.16.0
- Browser: All
- Extra: Typescript Monorepo

**Code Snippet**

Here is an example repo which demonstrates the issue 
https://github.com/rubber-duck-software/playwright-typescript-monorepo-tsx

**Describe the bug**

This is a followup to #17964. Playwright can now do the jsx transform in general, but it cannot handle jsx fragments. As far as I can tell `babel-plugin-syntax-jsx` does support fragments, so I am not sure about the underlying cause of this issue.

rubber-duck-software:
This patch fixes jsx fragment transforms

fixes: #19142

### DIFF
--- a/packages/playwright-test/ThirdPartyNotices.txt
+++ b/packages/playwright-test/ThirdPartyNotices.txt
@@ -53,11 +53,12 @@ This project incorporates components from the projects listed below. The origina
 -	@babel/plugin-syntax-private-property-in-object@7.14.5 (https://github.com/babel/babel)
 -	@babel/plugin-syntax-typescript@7.20.0 (https://github.com/babel/babel)
 -	@babel/plugin-transform-modules-commonjs@7.19.6 (https://github.com/babel/babel)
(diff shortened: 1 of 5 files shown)


rubber-duck-software:

pavelfeldman:
Could you add a test to `tests/playwright-test/playwright.ct-react.spec.ts`? You can run them as `npm run ttest playwright.ct`.

rubber-duck-software:
`playwright.ct-react` looks like its related to component testing, could you explain more about what test should be added there?

I just added a test to `loader.spec.ts` based on your work in #18123

pavelfeldman:
> playwright.ct-react looks like its related to component testing, could you explain more about what test should be added there?

I assumed you were using those fragments and it did not work. If those are just stray code imported from your code, test in loader is what we need.

pavelfeldman (approved):

PR #19820, opened 2023-01-02
feat: show diff tab by default

### LINKED ISSUE #19803: [Feature] select `diff` by default when it's available
When using html-reporter for visual testing, I would love to check `diff` view first when it's available,

Which means I can tell the difference between Actual and Expected at the first sight. Selecting Actual in the beginning isn't helpful IMO.

Let me know if you would accept a pull request for it.

chenxsan:
Close https://github.com/microsoft/playwright/issues/19803

I had to manually inject `window.playwrightReportBase64` data into `playwright/packages/html-reporter/index.html` when developing, wondering if there's any method I don't know.


packages/html-reporter/src/index.tsx
```diff
@@ -17,7 +17,7 @@
 import type { HTMLReport } from './types';
 import type zip from '@zip.js/zip.js';
 // @ts-ignore
-import zipImport from '@zip.js/zip.js/dist/zip-no-worker-inflate.min.js';
+import * as zipImport from '@zip.js/zip.js/lib/zip-no-worker-inflate.js';
```
chenxsan, line 20: According to https://github.com/gildas-lormeau/zip.js/tree/master/dist:

> Warning: These files are not compatible with ES modules, i.e. they cannot be imported with import. Instead, import index.js in the root folder of the project or one of the files prefixed with zip- in the [/lib](https://github.com/gildas-lormeau/zip.js/blob/master/lib) folder (e.g. /lib/zip-no-worker-inflate.js).

chenxsan:
The CI failure doesn't seem related to the changes.

aslushnikov:
Over to @pavelfeldman since this is a UX-related change.

packages/html-reporter/src/index.tsx
pavelfeldman: Could you share the report sizes before and after this change?

pavelfeldman (approved):
Sorry for the slow review. It looks good, but it now requires rebase.

dgozman (approved):

dgozman:
Thank you for the PR, merging in!

PR #19840, opened 2023-01-03
docs: fix example code of Route.fetch for Python

YusukeIwaki:
Fix typo in Route.fetch for Python. `route.fetch()` would be correct.

image


rwoll (approved):
Thanks!

PR #19841, opened 2023-01-03
docs(ct): pnpm install

sand4rt:


mxschmitt:
I think the vscode extension is not compatible with Yarn3 or PNPM thats why we are careful with adding it.

sand4rt:
@mxschmitt Thank you for the reply. Not sure about Yarn3, but PNPM in combination with the vscode extension works fine on my end. Please feel free to close the PR if necessary.

docs/src/test-components-js.md
```diff
@@ -61,6 +62,14 @@ yarn create playwright --ct
 ```
 
 </TabItem>
+
+<TabItem value="pnpm">
+
+ ```bash
+ pnpm dlx create-playwright --ct
```
pavelfeldman, line 69: Unlike npx, `pnpm dlx` won't be using the installed version of the package, will always be getting the newest one. While it works great in this use case, the rest of the documentation with `pnpm dlx playwright install` instead of `npx playwright install` will be completely broken.

pavelfeldman: on, so I learned that `pnpx exec` does a fine job with install, while dlx does a fine job with create-playwright.

pavelfeldman: https://github.com/microsoft/create-playwright/pull/76/files has landed, we need to push create-playwright and all is good.

pavelfeldman (approved):

docs/src/test-components-js.md
sand4rt: Ahh alright, nice!

PR #19853, opened 2023-01-04
feat(chromium-tip-of-tree): roll to r1074

playwrightmachine:
#### ` page/jshandle-to-string.spec.ts:42:3 › should work with different subtypes @smoke`

- a `Proxy` instance toString is now `Proxy(Object)` instead of `Proxy`
- bisect range: https://chromium.googlesource.com/chromium/src/+log/bca6d94b1730f8bef2996ca35ab7f28fcf087aff..adbb84441bb8bbe26ecf52a301e5b72de744fd50
- specific CL https://chromium.googlesource.com/v8/v8/+/57b1dc9acffc464205ea728706f355a32fa086dd

#### `library/proxy.spec.ts:180:3 › should exclude patterns` headless / win + mac

- `Error: Execution context was destroyed, most likely because of a navigation` gets thrown
- bisect range: https://chromium.googlesource.com/chromium/src/+log/50c936ec97b4ecc5fcb8025adb184943a25e9521..9a081ada6bfc9625f04d13bf62042f330f7dfe84
- Filed upstream: https://bugs.chromium.org/p/chromium/issues/detail?id=1405237

### DIFF
--- a/packages/playwright-core/browsers.json
+++ b/packages/playwright-core/browsers.json
@@ -15,9 +15,9 @@
     },
     {
       "name": "chromium-tip-of-tree",
-      "revision": "1069",
+      "revision": "1074",
       "installByDefault": false,
-      "browserVersion": "110.0.5479.0"
+      "browserVersion": "111.0.5515.0"
     },
     {
       "name": "firefox",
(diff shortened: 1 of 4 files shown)


dgozman:
Filed upstream issue. Meanwhile, we can change tests to include the following after any navigation that fails:

```ts
    // Failing navigation.
    const error = await page.goto('http://foo.is.the.another.test/target.html').catch(e => e);
    expect(error.message).toBeTruthy();
  
    // ADD THIS:
    // Make sure error page commits.
    if (browserName === 'chromium')
      await page.waitForURL('chrome-error://chromewebdata/');
```

PR #19856, opened 2023-01-04
fix(network): Accommodate async route handler callbacks

### LINKED ISSUE #19855: [BUG] RouteHandlerCallback type incompatible with async route.* methods
**Context:**
- Playwright Version: 1.29.1
- Operating System: Mac
- Node.js version: 14.21
- Browser: Chromium

**Code Snippet**

Consider the example from https://playwright.dev/docs/network#handle-requests
```ts
    await page.route('**/api/fetch_data', route => route.fulfill({
      status: 200,
      body: testData,
    }));
```

**Describe the bug**

`RouteHandlerCallback` expects that the callback only returns `void`:
https://github.com/microsoft/playwright/blob/3f333a8/packages/playwright-core/src/client/network.ts#L425

However, most of the `route` apis are async and thus return `Promise<void>`:
`route.abort`: https://github.com/microsoft/playwright/blob/3f333a8/packages/playwright-core/src/client/network.ts#L310
`route.continue`: https://github.com/microsoft/playwright/blob/3f333a8/packages/playwright-core/src/client/network.ts#L394
`route.fallback`: https://github.com/microsoft/playwright/blob/3f333a8/packages/playwright-core/src/client/network.ts#L304
`route.fetch`: https://github.com/microsoft/playwright/blob/3f333a8/packages/playwright-core/src/client/network.ts#L322
`route.fulfill`: https://github.com/microsoft/playwright/blob/3f333a8/packages/playwright-core/src/client/network.ts#L329

`Promise<void>` is not assignable to `<void>` and raises a flag with [@typescript-eslint/no-misused-promises](https://typescript-eslint.io/rules/no-misused-promises/)
Screen Shot 2023-01-03 at 6 15 57 PM

To address this should be as simple as amending the return type of `RouteHandlerCallback` to include `Promise<void>` (see https://github.com/microsoft/playwright/pull/19856)

carmanchris31:
fixes #19855

Signed-off-by: Christopher Carman <carmanchris31@gmail.com>

### DIFF
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -422,7 +422,7 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
   }
 }
 
-export type RouteHandlerCallback = (route: Route, request: Request) => void;
+export type RouteHandlerCallback = (route: Route, request: Request) => void | Promise<void>;
 
 export type ResourceTiming = {
   startTime: number;


mxschmitt:
The Playwright project has a [generated type layer](https://github.com/microsoft/playwright/pull/19691) in between which is actually exposed to the public, it is generated based off the [docs](https://github.com/microsoft/playwright/blob/afc1774a2ba7e89a6a8dd108d98f2ec4ae83378f/docs/src/api/class-page.md?plain=1#L3138) and the [override-types](https://github.com/microsoft/playwright/blob/3afd83c8cce5b036ad93b0dca188e8a5bd59abbf/utils/generate_types/overrides.d.ts).

Let us know if you need help, I'd try massaging the docs first, by doing something like `- `handler` <[function]\([Route], [Request]\):[Promise|void]>` and run `node ./utils/generate_types`.

carmanchris31:
I'll take a look, thanks!

Issue #15684, opened 2022-07-14
[Feature] Experimental Service Worker Network Events 


rwoll:

joshbrooks:
I tried this, our use case is in being able to mock 404's and 500's from service worker requests, to test an offline-first experience handling both GET and PUT/POST requests from service workers. Using the "experimental" flag "PW_EXPERIMENTAL_SERVICE_WORKER_NETWORK_EVENTS" from Python with `1.25.2` seems to have no effect:

```python
os.environ["PW_EXPERIMENTAL_SERVICE_WORKER_NETWORK_EVENTS"] = "1"
```
```python
  route.fulfill(
      status=HTTPStatus.INTERNAL_SERVER_ERROR,
      content_type="text/plain",
      body="not found!"
  )
```

This still returns a 200 from the service worker. Other (non-service-worker) requests return a 500 though

As a workaround, for some tests it's sufficient just to block service workers

`context = browser.new_context(service_workers='block')`

wentokay:
Thanks for adding this experimental feature. I'm using it when testing a chrome browser extension.

Network requests are made from the background script (which is a serviceworker due to [manifest v3](https://developer.chrome.com/docs/extensions/mv3/intro/) requirements) 

It seems to be doing what I need right now but will report back if there are any issues.

smokir:
@wentokay could you please share some configuration examples? I've been doing the same thing (chrome extension on v3 manifest) and using `playwright` with `jest-playwright-preset` for e2e testing. But got no luck mocking API requests coming from a service worker.

damienforster:
Interested in this also for Playwright testing a manifest v3 extension.  Currently I'm not seeing requests from the service worker getting picked up, even with the PW_EXPERIMENTAL_SERVICE_WORKER_NETWORK_EVENTS flag

smokir:
@damienforster I had the same problem because I was using `jest-playwright-preset` instead of `@playwright/test`

MSOwnExperiment:
> Thanks for adding this experimental feature. I'm using it when testing a chrome browser extension.
> 
> Network requests are made from the background script (which is a serviceworker due to [manifest v3](https://developer.chrome.com/docs/extensions/mv3/intro/) requirements)
> 
> It seems to be doing what I need right now but will report back if there are any issues.

I also have the same use case. I could fetch the background page for Manifest V3. However, this part is not working for me. 
const [request] = await Promise.all([
[...]

PR #14714, opened 2022-06-08
feat: add allow/block serviceWorkers option

rwoll:
Adds cross-browser support for easily allowing/blocking Service Workers via a Context option.

Includes plumbing for Playwright Test's `use`.

Resolves #14522.

Relates #1090.
Supercedes #14321.

PR #19870, opened 2023-01-04
fix(serviceworker): network inspection works without options.serviceWorkers set

orenyomtov:
In the implementation of adding the serviceWorkers option (#14714), the default value was not properly set.

This PR fixes the workaround needed here:
https://github.com/microsoft/playwright/issues/15684#issuecomment-1371127580

As you can see, it is documented to default to `"allow"`: https://github.com/rwoll/playwright/blob/main/docs/src/api/params.md#context-option-service-worker-policy

An alternative fix suggested is to edit the documentation:
https://github.com/microsoft/playwright/pull/19871

Signed-off-by: Oren <orenyomtov@users.noreply.github.com>

### DIFF
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -421,7 +421,7 @@ export async function prepareBrowserContextParams(options: BrowserContextOptions
     noDefaultViewport: options.viewport === null,
     extraHTTPHeaders: options.extraHTTPHeaders ? headersObjectToArray(options.extraHTTPHeaders) : undefined,
     storageState: await prepareStorageState(options),
-    serviceWorkers: options.serviceWorkers === null ? 'allow' : options.serviceWorkers,
+    serviceWorkers: options.serviceWorkers,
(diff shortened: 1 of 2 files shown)


orenyomtov:
Hey @rwoll, appreciate the super quick review.

I'm using the `playwright` npm package.

This is the test script I was running:

The requests sent from the extension's background page (actually service worker nowadays) don't show up unless I explicitly set `serviceWorkers` to `"allow"`:
https://we.tl/t-Qch7BaWQBW

dgozman:
I think the right fix is to allow `undefined` in this line: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/chromium/crServiceWorker.ts#L124. Perhaps even checking that `options.serviceWorkers !== 'block'` instead. We usually tolerate undefined values for the options and assume the default value across the codebase.

orenyomtov:
Implemented @dgozman 's approach

dgozman (approved):
Looks good, merging in.

PR #19891, opened 2023-01-05
feat: add pnpm to Docker image

nielsdB97:
Install the [pnpm](https://pnpm.io/) package manager in the Docker images.

### DIFF
--- a/utils/docker/Dockerfile.focal
+++ b/utils/docker/Dockerfile.focal
@@ -13,7 +13,7 @@ RUN apt-get update && \
     apt-get install -y nodejs && \
     # Feature-parity with node.js base images.
     apt-get install -y --no-install-recommends git openssh-client && \
-    npm install -g yarn && \
+    npm install -g yarn pnpm && \
     # clean apt cache
     rm -rf /var/lib/apt/lists/* && \
     # Create the pwuser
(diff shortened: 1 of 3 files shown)


nielsdB97:

aslushnikov:
@nielsdB97 we already have NPM installed inside the docker image; if you want to customize the image, please create your own image that inherits from the `mcr.microsoft.com/playwright`!

nielsdB97:
@aslushnikov thanks for checking out my PR!
Seeing that next to NPM also Yarn is installed by default, I figured that we could also add PNPM to the list of supported package managers.
It has seen quite an uptake in adoption over 2022 and on my current project we're very happy using it.

On [this certain date in December](https://npm-stat.com/charts.html?package=pnpm&package=yarn&from=2022-01-01&to=2022-12-30) it even had 524.000 downloads, which compared to 702.000 downloads for Yarn comes pretty close:
Screenshot 2023-01-19 at 09 57 00

I'd like to hear your thoughts on when a package manager is qualified to be supported in the project.

mxschmitt:
(The reason why we install Yarn is so that we are compatible with the [upstream Node.js Docker image](https://github.com/nodejs/docker-node/blob/main/18/buster/Dockerfile).)

PR #19902, opened 2023-01-05
feat(fetch): happy eyeballs

### LINKED ISSUE #18790: [Feature] implement happy eyeballs for API testing
After switching from  Node.js 16 to 18 API tests started failing in Java because it only binds ipv4 127.0.0.1 and node resolves localhost to ::1 by default. To mitigate that we'd like to implement happy eyeballs on our end before node itself fixes it (see https://github.com/nodejs/node/issues/41625).

yury-s:
Fixes #18790


packages/playwright-core/src/server/happy-eyeballs.ts
```diff
@@ -0,0 +1,123 @@
...
+      // TODO: Cache the result?
+      // Close other outstanding sockets.
+      for (const s of sockets) {
+        if (s !== socket)
+          s.destroy();
+      }
+    });
+    socket.on('timeout', () => {
+      handleError(new Error('Connection timeout'));
+      socket.destroy();
```
aslushnikov, line 90: let's put them in a Set so that we don't call `socket.destroy()` twice

aslushnikov (approved):

PR #19905, opened 2023-01-05
chore(html): show project name always for the test file

### LINKED ISSUE #19874: [BUG] Browser tag is not displayed in HTML Report When Running in single browser
**Context:**
- Playwright Version: [what Playwright version do you use?]1.29.1
- Operating System: [e.g. Windows, Linux or Mac] Windows
- Node.js version: [e.g. 12.22, 14.6] v16.13.0
- Browser: [e.g. All, Chromium, Firefox, WebKit] NA

Description:
I have a few tests which I ran in Chromium browser using CLI command " npx playwright test login --project chromium"
Next I ran the command for Report Generation and I got the HTML report where "chromium" tag is missing.
image

However, when I ran the same tests in all browsers (using npx playwright test login )
It shows the browser tags.
image

It seems like a bug to me. Please have a look.

mxschmitt:
This shows the project name in the header if `new Set(projectNames).size === 0`. In the other cases it gets shown as before next to the test file.

Preview: 
image

Fixes https://github.com/microsoft/playwright/issues/19874

### DIFF
--- a/packages/html-reporter/src/headerView.spec.tsx
+++ b/packages/html-reporter/src/headerView.spec.tsx
@@ -96,14 +96,4 @@ test('should show the project names', async ({ mount }) => {
     await expect(component.getByText('my-project')).toBeHidden();
     await expect(component.getByText('great-project')).toBeHidden();
   });
-  await test.step('with more than 1 project but same name', async () => {
-    const component = await mount(<HeaderView
-      stats={stats}
-      filterText=''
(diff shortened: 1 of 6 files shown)


pavelfeldman:
We could probable have a global section that would state the project name if that's the only project we report on. That way we mention chromium only once.

pavelfeldman (approved):

PR #19954, opened 2023-01-09
docs: add Electron troubleshooting note

### LINKED ISSUE #19854: [BUG] Electron app launches correctly, but await electron.launch is never resolved
**Context:**
- Playwright Version: 1.29.1
- Operating System: Linux (Ubuntu 22.04)
- Node.js version: 16.17.0
- Browser: Electron 

**Code Snippet**

```javascript
import { _electron as electron } from 'playwright';

test.beforeAll(async () => {
  const electronApp = await electron.launch({
    args: ['ABSOLUTE_PATH/release/dev/linux-unpacked/resources/app.asar/app/dist/main.prod.js'],
    executablePath: 'ABSOLUTE_PATH/release/dev/linux-unpacked/resources/app.asar/app/dist/main.prod.js'
  });
})
```

**Describe the bug**

I adapted this code from [electron-playwright-example](https://github.com/spaceagetv/electron-playwright-example) and tested it with [electron-react-boilerplate](https://github.com/electron-react-boilerplate/electron-react-boilerplate) and [secure-electron-template](https://github.com/reZach/secure-electron-template) and it works in all three of those projects - but for our project (which is adapted from the latter two) it hangs indefinitely.

I tried to locate the code responsible for `electron.launch` to investigate possible causes.

Our Electron app is loaded via the `app:` protocol, disables `nodeIntegration` and enables a `preload` script - but all the same is true in `secure-electron-template` as well - so I can't figure out why it fails to resolve `await electron.launch(...)`

The strangest thing to me is that the app runs entirely correctly, the Playwright code appears to run correctly, but still we get stuck on the `await`.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/19854

### DIFF
--- a/docs/src/api/class-electron.md
+++ b/docs/src/api/class-electron.md
@@ -51,6 +51,12 @@ PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm i -D playwright
 * v13.4.0+
 * v14+
 
+**Trouble shooting:**
+
+If you are not able to launch Electron and it will end up in timeouts during launch, try the following:
+
+- Ensure that `nodeCliInspect` ([FuseV1Options.EnableNodeCliInspectArguments](https://www.electronjs.org/docs/latest/tutorial/fuses#nodecliinspect)) fuse is **not** set to `false`.
+
 ## async method: Electron.launch
 * since: v1.9
 - returns: <[ElectronApplication]>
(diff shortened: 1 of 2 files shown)


docs/src/api/class-electron.md
```diff
@@ -51,6 +51,12 @@ PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm i -D playwright
 * v13.4.0+
 * v14+
 
+**Trouble shooting:**
```
pavelfeldman, line 54:
```suggestion
**Troubleshooting:**
```

pavelfeldman: Actually, I would call it Known issues.

pavelfeldman (approved):

PR #19964, opened 2023-01-09
fix: use no internal selectors for frame locators

### LINKED ISSUE #19406: [BUG] Codegen with frame locators uses internal selectors
Repro
1. `npx playwright codegen https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_iframe`
2. Click on e.g. "Log in"
image

Expected: selector without internal: prefix
Actual: Selector with internal:pref

`await page.frameLocator('#iframeResult').frameLocator('internal:attr=[title="W3Schools Free Online Web Tutorials"i]').getByRole('link', { name: 'Log in' }).dblclick();`

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/19406

### DIFF
--- a/packages/playwright-core/src/server/injected/selectorGenerator.ts
+++ b/packages/playwright-core/src/server/injected/selectorGenerator.ts
@@ -34,6 +34,7 @@ const kExactPenalty = kTextScoreRange / 2;
 const kTestIdScore = 1;        // testIdAttributeName
 const kOtherTestIdScore = 2;   // other data-test* attributes
 
+const kIframeByAttributeScore = 10;
 
 const kBeginPenalizedScore = 50;
 const kPlaceholderScore = 100;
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/injected/selectorGenerator.ts
```diff
@@ -216,8 +216,12 @@ function buildCandidates(injectedScript: InjectedScript, element: Element, testI
     candidates.push({ engine: 'css', selector: `${cssEscape(element.nodeName.toLowerCase())}[name=${quoteAttributeValue(element.getAttribute('name')!)}]`, score: kCSSInputTypeNameScore });
 
   if (element.getAttribute('title')) {
-    candidates.push({ engine: 'internal:attr', selector: `[title=${escapeForAttributeSelector(element.getAttribute('title')!, false)}]`, score: kTitleScore });
-    candidates.push({ engine: 'internal:attr', selector: `[title=${escapeForAttributeSelector(element.getAttribute('title')!, true)}]`, score: kTitleScoreExact });
+    if (element.nodeName === 'IFRAME') {
```
pavelfeldman, line 219: Do we need a nodeName === IFRAME check as a first thing here given that we won't be using any of the non-CSS engines to locate it?

packages/playwright-core/src/server/injected/selectorGenerator.ts
```diff
@@ -182,6 +183,13 @@ function buildCandidates(injectedScript: InjectedScript, element: Element, testI
       candidates.push({ engine: 'css', selector: `[${attr}=${quoteAttributeValue(element.getAttribute(attr)!)}]`, score: kOtherTestIdScore });
   }
 
+  if (element.nodeName === 'IFRAME') {
+    for (const attribute of ['name', 'title']) {
+      if (element.getAttribute(attribute))
+        candidates.push({ engine: 'css', selector: `${cssEscape(element.nodeName.toLowerCase())}[${attribute}=${quoteAttributeValue(element.getAttribute(attribute)!)}]`, score: kIframeScore });
```
pavelfeldman, line 189: Should be kCSSTagNameScore since you are using css. Scorer will be comparing it to kTestIdScore.

pavelfeldman (approved):

Issue #19746, opened 2022-12-28
[BUG] VSCode extension does not update testInfo object


lucasbit:
**Context:**
- Playwright Version: 1.28.1
- Operating System: [e.g. Windows 11
- Node.js version: 18.12.1
- Visual Studio Code version: 1.74.2
- Playwright for VSCode extension version: 1.0.3
- Browser: Chromium
- Extra: 

**Code Snippet**
![image](https://user-images.githubusercontent.com/29520131/209802785-d39392d3-f406-4afd-a6a1-b2cf7280acfe.png)
![image](https://user-images.githubusercontent.com/29520131/209803191-c48ba0d9-3828-4bdc-981c-ca77eb971333.png)
![image](https://user-images.githubusercontent.com/29520131/209803607-4284743b-87b2-42cb-a543-a5f9caa851a4.png)

Help us help you! Put down a short code snippet that illustrates your bug and
that we can run and debug locally. For example:

**Describe the bug**
I'm trying to implement visual comparison, but I have a lot of differences between headfull and headless testing with Chromium ( 12912 pixels (ratio 0.04 of all image pixels) are different) or perhaps that is not a lot off difference? 
![image](https://user-images.githubusercontent.com/29520131/209804283-3a1a237c-4847-4d8e-8f3a-e72a6b79cb5b.png)

It works fine when testing only in headless mode. So my idea is that, I can read value of testInfo object and run visual checks only when browser is headless. But that solution does not work when running tests from the VSCode extension. And that a big issue as then I can't use your amazing locator selector and other cool features from VSCode extension. 

Or perhaps there is another value in testInfo object I can check out if test run is triggered from terminal vs VSCode extension? 

Add any other details about the problem here.

dgozman:
@lucasbit Instead of looking into `testInfo.project.use.headless` that reflects your config file and not the final value, examine the `headless` fixture:

```ts
test('my test', async ({ page, headless }) => {
  console.log(headless);
});
```

Let me know if that helps.

lucasbit:
@dgozman, thanks for quick replay, sadly your suggestion didn't work. I will paste my code snippet in case I'm doing something wrong(doubt it):
![image](https://user-images.githubusercontent.com/29520131/209966235-8f0631aa-f676-4231-a0ff-3058f440a945.png)
![image](https://user-images.githubusercontent.com/29520131/209966356-9da7b9d7-622a-4887-b235-86a0395af4b0.png)

dgozman:
[...]

Issue #19881, opened 2023-01-04
[BUG] Cli after "Run Test" with on-toggled "Show Browser" does not start  with --headed parameter


joma74:
**Context:**
- Playwright Version: 1.29.1
- Operating System: Ubuntu 16
- Node.js version: 16.14.2
- Visual Studio Code version: 1.74.2
- Playwright for VSCode extension version: 1.0.3
- Browser: All
- Extra: No

**Retrace**

- Open Test Explorer
- Toggle "Show Browser" on
- Write test that assumes the `headless` parameter to the args of the test function is false
- "Run Test"
- See the test fail

**Describe the bug**

Playwright Cli after "Start Test" with on-toggled "Show Browser" does not start  with --headed parameter. 

Here is my Playwright Cli process call:

`joma     24808 19258  9 22:41 ?        00:00:00 /home/joma/.nvm/versions/node/v16.14.2/bin/node /home/joma/entwicklung/nodews/tallpad-headlessui-sidebar-vite/node_modules/playwright-core/lib/cli/cli test -c playwright.config.ts --project=Desktop Firefox /home/joma/entwicklung/nodews/tallpad-headlessui-sidebar-vite/tests/e2e/basic\.spec\.ts:12 --repeat-each 1 --retries 0 --workers 1`

It shows no `--headed` parameter.

Hence the `headless` parameter to the args of the test function is set to true. On the contrary, any browser is started headed, because of the on-toggled "Show Browser".

This issue bothers me as screenshots of at least different sizes(t.i. heights) are created in headed mode vs headless mode(poss. related  https://github.com/microsoft/playwright/issues/12683). Have to maintain two screenshots for each mode. Therefore, on image comparison, i have to take the`headless` parameter from the args to the test function into account, which fails for matching the headless golden snapshot image with the headfull browser snapshot image. Which gives following Test Report snippet
```
...
1) [Desktop Firefox] › basic.spec.ts:12:5 › looks like expected screenshot =======================

    Error: Screenshot comparison failed:

      Expected an image 1280px by 3602px, received 1280px by 3630px. 

    Call log:
      - expect.toHaveScreenshot with timeout 5000ms
      -   verifying given screenshot expectation
      - taking page screenshot
      -   disabled all CSS animations
      - Expected an image 1280px by 3602px, received 1280px by 3630px. 
      - waiting 100ms before taking screenshot
      - taking page screenshot
      -   disabled all CSS animations
      - captured a stable screenshot
[...]

PR #19980, opened 2023-01-10
chore: respect headless option when reusing browser

dgozman:
References #19746.


mxschmitt:
Looks the same as https://github.com/microsoft/playwright/issues/19881 which we can close by landing this.

pavelfeldman (approved):

PR #20009, opened 2023-01-10
fix: support import assertions

### LINKED ISSUE #19928: Playwright fails to import json files for transpiled typeScript code when type=module
## Context:
- Playwright Version: `1.29.1`
- Operating System: `Mac`
- Node.js version: `18.12.1`
- Browser: All
- Extra: TypeScript 4.9.4

## Reproducible Code
Here's a reproducible [repo](https://github.com/cristobal/experiment-playwright-import-json) 

1. Checkout the code
2. install the dependencies

#### ESM Variant
Navigate into the `esm-variant` directory and run `npx playwright test`, which yields error as in the picture below: 
Screenshot 2023-01-06 at 21 19 50

Since the typescript code is transpiled to ESM compatible code given that `type=module` is specified in the package.json, however babel is not configured with support for __importAssertions__.

#### CommonJS Variant
Navigate into `commonjs-variant` directory and run `npx playwright`, which yields no error as in the picture below: 
Screenshot 2023-01-06 at 21 10 27

Since the typescript code is transpiled to CommonJS compatible code given that `type=commonjs` is specified in the package.json.

## Describe the bug

The problem is that the underlying babel used by playwright to transpile TypeScript, infers the type from the module field in the package.json from where the playwright is run. Whenever babel infers that the output type should be of type module aka ESM it then transpiles the code to esm compatible code that will not run via the current babel setup in Playwright since:
>Support for the experimental syntax 'importAssertions' isn't currently enabled.

The [@babel/plugin-syntax-import-assertions](ttps://github.com/babel/babel/tree/main/packages/babel-plugin-syntax-import-assertions) needs to be added to the current playwright setup.

aslushnikov:
Fixes #19928

### DIFF
--- a/packages/playwright-test/ThirdPartyNotices.txt
+++ b/packages/playwright-test/ThirdPartyNotices.txt
@@ -42,6 +42,7 @@ This project incorporates components from the projects listed below. The origina
 -	@babel/plugin-syntax-async-generators@7.8.4 (https://github.com/babel/babel/tree/master/packages/babel-plugin-syntax-async-generators)
 -	@babel/plugin-syntax-dynamic-import@7.8.3 (https://github.com/babel/babel/tree/master/packages/babel-plugin-syntax-dynamic-import)
 -	@babel/plugin-syntax-export-namespace-from@7.8.3 (https://github.com/babel/babel/tree/master/packages/babel-plugin-syntax-export-namespace-from)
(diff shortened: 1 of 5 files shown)


dgozman (approved):

mxschmitt (approved):

cristobal:
Thanks for the fix 🙏🏽 👋🏽

PR #20012, opened 2023-01-10
fix: do not write to stdout on cssTokenizer parse error

### LINKED ISSUE #20074: [BUG] Playwright silently crashes with newline characters in selectors
**Context:**
- Playwright Version: 1.29.1
- Operating System: Mac
- Python version: 3.9
- Browser: Chromium

**Code Snippet**

```Python
import asyncio
from playwright.async_api import async_playwright

non_sanitised_string = "Hello \nWorld"
sanitised_text = non_sanitised_string.replace("\n", " ").replace("\r", " ").strip()

async def try_error():
    async with async_playwright() as pw:
        browser = await pw.chromium.launch(headless=False)
        context = await browser.new_context(ignore_https_errors=True)

        page = await context.new_page()
        await page.goto("https://any-website.com")
        count = await page.locator(f'div:has-text("{non_sanitised_string}")').first.count()
        print(count)
        
loop = asyncio.get_event_loop()
loop.run_until_complete(try_error())
```

**Describe the bug**

When a selector contains a newline e.g. with `non_sanitised_string` the Playwright debugger just goes blank and no errors are printed to the terminal.  When `sanitised_string` is used instead the code works correctly.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/20074

### DIFF
--- a/packages/playwright-core/src/server/isomorphic/cssTokenizer.ts
+++ b/packages/playwright-core/src/server/isomorphic/cssTokenizer.ts
@@ -130,7 +130,6 @@ export function tokenize(str1: string): CSSTokenInterface[] {
   };
   const donothing = function() { };
   const parseerror = function() {
-    // eslint-disable-next-line no-console
     // Language bindings don't like writing to stdout!
     // console.log('Parse error at index ' + i + ', processing codepoint 0x' + code.toString(16) + '.'); return true;
   };
--- a/packages/playwright-core/src/server/isomorphic/cssTokenizer.ts
+++ b/packages/playwright-core/src/server/isomorphic/cssTokenizer.ts
@@ -131,7 +131,8 @@ export function tokenize(str1: string): CSSTokenInterface[] {
   const donothing = function() { };
   const parseerror = function() {
     // eslint-disable-next-line no-console
-    console.log('Parse error at index ' + i + ', processing codepoint 0x' + code.toString(16) + '.'); return true;
+    // Language bindings don't like writing to stdout!
+    // console.log('Parse error at index ' + i + ', processing codepoint 0x' + code.toString(16) + '.'); return true;
   };
 
   const consumeAToken = function(): CSSTokenInterface {


packages/playwright-core/src/server/isomorphic/cssTokenizer.ts
```diff
@@ -131,7 +131,8 @@ export function tokenize(str1: string): CSSTokenInterface[] {
   const donothing = function() { };
   const parseerror = function() {
     // eslint-disable-next-line no-console
```
pavelfeldman, line 133: Let's also remove this line.

packages/playwright-core/src/server/isomorphic/cssTokenizer.ts
```diff
@@ -131,7 +131,8 @@ export function tokenize(str1: string): CSSTokenInterface[] {
...
-    console.log('Parse error at index ' + i + ', processing codepoint 0x' + code.toString(16) + '.'); return true;
+    // Language bindings don't like writing to stdout!
+    // console.log('Parse error at index ' + i + ', processing codepoint 0x' + code.toString(16) + '.'); return true;
```
pavelfeldman, line 135: Can we still print those when under test?

mxschmitt: would be a lot of plumbing of `isUnderTest` and introduce a `logErrors: boolean` based on `isUnderTest` since `cssTokenizer` is isomorphic.

As of today we'd throw the following: `"Unsupported token \\"BADSTRING\\" while parsing selector \\"div:has-text(\\"A\\nA\\")\\"`

pavelfeldman: Let's ask @dgozman if he is comfortable dropping these or we need to wire them...

dgozman: Seems safe, because we always return something like `BadStringToken()` and throw later on.

dgozman (approved):

PR #20040, opened 2023-01-11
docs: update route.fetch example to use new api

yury-s:


docs/src/network.md
```diff
@@ -575,7 +575,7 @@ To modify a response use [APIRequestContext] to get the original response and th
 ```js
 await page.route('**/title.html', async route => {
   // Fetch original response.
-  const response = await page.request.fetch(route.request());
+  const response = await page.request.fetch();
```
mxschmitt, line 578:
```suggestion
  const response = await route.fetch()
```
shouldn't it be like that and in the other places?

dgozman: +1

dgozman (approved):

mxschmitt (approved):

PR #20053, opened 2023-01-11
docs(dotnet): add support for Chrome DevTools Protocol

hutchcodes:
Implemented in https://github.com/microsoft/playwright-dotnet/pull/2448

### DIFF
--- a/docs/src/api/class-cdpsession.md
+++ b/docs/src/api/class-cdpsession.md
@@ -48,9 +48,9 @@ client.send("Animation.setPlaybackRate", {
 ```csharp
 var client = await Page.Context.NewCDPSessionAsync(Page);
 await client.SendAsync("Runtime.enable");
-client.Event("Animation.animationCreated").OnEvent += (_, _) => Console.WriteLine("Animation created!"));
+client.Event("Animation.animationCreated").OnEvent += (_, _) => Console.WriteLine("Animation created!");
 var response = await client.SendAsync("Animation.getPlaybackRate");
-var playbackRate = response.Value.Deserialize<JsonNode>()["result"]["playbackRate"].GetValue<decimal>();
+var playbackRate = response.Value.GetProperty("playbackRate").GetDouble();
 Console.WriteLine("playback rate is " + playbackRate);
 await client.SendAsync("Animation.setPlaybackRate", new() { { "playbackRate", playbackRate / 2 } });
 ```
@@ -70,6 +70,7 @@ send messages.
 * since: v1.30
 * langs: csharp
 - returns: <[JsonElement?]>
+
 ### param: CDPSession.send.method
 * since: v1.8
 * langs: js, python, csharp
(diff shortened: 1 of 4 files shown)


docs/src/api/class-cdpsession.md
```diff
@@ -45,6 +45,16 @@ client.send("Animation.setPlaybackRate", {
     playbackRate: response["playbackRate"] / 2
 })
 ```
+```csharp
+var client = await Page.Context.NewCDPSessionAsync(Page);
```
mxschmitt, line 49: this example is not working. 

line 51: maybe we need an overload for callbacks without arguments.
line 54/55/56 should be `playbackRate` instead of `response.playbackRate`.

docs/src/api/class-cdpsession.md
```diff
@@ -64,6 +74,15 @@ Protocol method name.
 
 ### param: CDPSession.send.params
 * since: v1.8
+* langs: js, python
 - `params` ?<[Object]>
 
 Optional method parameters.
+
+### param: CDPSession.send.params
+* since: v1.8
```
mxschmitt, line 83:
```suggestion
* since: v1.30
```

docs/src/api/class-cdpsession.md
```diff
@@ -64,6 +74,15 @@ Protocol method name.
...
+* langs: csharp
+  - alias-csharp: args
+- `params` ?<[Dictionary<string, Object>]>
+
+Optional method parameters.
```
mxschmitt, line 88: We might need to declare in this file "On" as well.

docs/src/api/class-cdpsession.md
hutchcodes: Adding an overload for callbacks without arguments would mean maintaining 2 dictionaries of callbacks, one for Action<JsonElement> and one for Action. This might be a nice to have in the future, but for now I'm just going to use a discard for the unwanted arguments.

docs/src/api/class-cdpsession.md
hutchcodes: Added AddEventListener and RemoveEventListener

docs/src/api/class-cdpsession.md
```diff
@@ -54,16 +63,63 @@ send messages.
...
+
+Removes a listener for a named CDP event
+
+### param: CDPSession.removeEventListener.eventName
+* since: v1.30
+* langs: csharp
+- `eventName` <[string]>
+
+### param: CDPSession.removeEventListener.eventHandler
+* since: v1.830
```
mxschmitt, line 123:
```suggestion
* since: v1.30
```

docs/src/api/class-cdpsession.md
```diff
@@ -54,16 +63,63 @@ send messages.
...
+* langs: js, python
 - `params` ?<[Object]>
 
 Optional method parameters.
+
+### param: CDPSession.send.params
+* since: v1.30
+* langs: csharp
+  - alias-csharp: args
+- `params` ?<[Dictionary<string, Object>]>
```
pavelfeldman, line 91: I believe we map `Map` to a `Dictionary`.

docs/src/api/class-cdpsession.md
```diff
@@ -54,16 +63,63 @@ send messages.
...
+
+Optional method parameters.
+
+## method: CDPSession.addEventListener
```
pavelfeldman, line 95: Please see my comments on the C# patch for this.

PR #20062, opened 2023-01-12
feat: config.globalScripts

yury-s:
Introduce config.globalScripts. Tests from the matching files will run before all projects. We'll only allow beforeAll/afterAll instead of tests in such files (next PR).

Global scripts are executed as part of 'Global Scripts' project which is not present in FullConfig.projects but may be referenced by corresponding global setup Suites.

### DIFF
--- a/packages/playwright-test/src/runner.ts
+++ b/packages/playwright-test/src/runner.ts
@@ -238,13 +238,29 @@ export class Runner {
     return projects;
   }
 
-  private async _collectFiles(projects: FullProjectInternal[], commandLineFileFilters: TestFileFilter[]): Promise<{filesByProject: Map<FullProjectInternal, string[]>; setupFiles: Set<string>}> {
+  private async _collectFiles(projects: FullProjectInternal[], commandLineFileFilters: TestFileFilter[]): Promise<{filesByProject: Map<FullProjectInternal, string[]>; setupFiles: Set<string>, globalSetupFiles: Set<string>}> {
     const extensions = ['.js', '.ts', '.mjs', '.tsx', '.jsx'];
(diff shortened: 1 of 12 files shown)


packages/playwright-test/src/runner.ts
```diff
@@ -492,15 +516,23 @@ export class Runner {
...
-      let dispatchResult = await this._dispatchToWorkers(projectSetupGroups);
+      // TODO: run only setups, keep workers alive
+      let dispatchResult = await this._dispatchToWorkers(globalSetupGroups);
       if (dispatchResult === 'success') {
-        const failedSetupProjectIds = new Set<string>();
-        for (const testGroup of projectSetupGroups) {
-          if (testGroup.tests.some(test => !test.ok()))
-            failedSetupProjectIds.add(testGroup.projectId);
+        if (globalSetupGroups.some(group => group.tests.some(test => !test.ok()))) {
+          this._skipTestsFromMatchingGroups(testGroups, () => true);
```
dgozman, line 523:
```suggestion
          this._skipTestsFromMatchingGroups([...testGroups, ...projectSetupGroups], () => true);
```

packages/playwright-test/src/runner.ts
```diff
@@ -492,15 +516,23 @@ export class Runner {
 
     // Run tests.
     try {
-      let dispatchResult = await this._dispatchToWorkers(projectSetupGroups);
+      // TODO: run only setups, keep workers alive
```
dgozman, line 519:
```suggestion
      // TODO: run only setups, keep workers alive, inherit process.env from global setup workers
```

dgozman (approved):

PR #20092, opened 2023-01-13
fix(tsx): resolve .js imports to .tsx as well

### LINKED ISSUE #20039: [BUG] Cannot import TSX components with ESM
**Context:**
- Playwright Version: 1.29
- Operating System: Mac
- Node.js version: 18.12.1, 16.8.1
- Browser: N/A

**Code Snippet**

Trying to call `import { TestComponent } from "./Component.js";` in a Playwright test blows up with a `Cannot find module` error.

**Describe the bug**

It seems that [tsxTransform](https://github.com/microsoft/playwright/blob/main/packages/playwright-test/src/tsxTransform.ts#L46) has a different codepath when it detects JSX, so the ESM resolution doesn't work.

I _think_ the problem is that [gathering the component info](https://github.com/microsoft/playwright/blob/main/packages/playwright-test/src/tsxTransform.ts#L176) uses `require.resolve` but the `.js` file doesn't exist (yet?). 

[Here is an example repo showing the problem](https://github.com/drewloomer/playwright-import-repro).

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/20039

### DIFF
--- a/packages/playwright-test/src/transform.ts
+++ b/packages/playwright-test/src/transform.ts
@@ -169,11 +169,14 @@ export function resolveHook(filename: string, specifier: string): string | undef
 
 export function js2ts(resolved: string): string | undefined {
   const match = resolved.match(/(.*)(\.js|\.jsx|\.mjs)$/);
-  if (match) {
-    const tsResolved = match[1] + match[2].replace('j', 't');
-    if (!fs.existsSync(resolved) && fs.existsSync(tsResolved))
-      return tsResolved;
-  }
+  if (!match || fs.existsSync(resolved))
+    return;
+  const tsResolved = match[1] + match[2].replace('js', 'ts');
+  if (fs.existsSync(tsResolved))
+    return tsResolved;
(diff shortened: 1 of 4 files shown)


packages/playwright-test/src/transform.ts
```diff
@@ -169,11 +169,14 @@ export function resolveHook(filename: string, specifier: string): string | undef
...
-    const tsResolved = match[1] + match[2].replace('j', 't');
-    if (!fs.existsSync(resolved) && fs.existsSync(tsResolved))
-      return tsResolved;
-  }
+  if (!match || fs.existsSync(resolved))
+    return;
+  const tsResolved = match[1] + match[2].replace('js', 'ts');
+  if (fs.existsSync(tsResolved))
+    return tsResolved;
+  const tsxResolved = match[1] + match[2].replace('js', 'tsx');
```
dgozman, line 177: This produces `mtsx` from `mjs`.

dgozman (approved):

Issue #10369, opened 2021-11-17
[Question] Attaching playwright to an existing electron window?


sunmeixia2005:
Hi guys,

I am new on playwright, currently investigate if playwright can connect to an existing electron application?
I try to some code:
`browser_aut = playwright.chromium.connect("ws://127.0.0.1:8768/devtools/browser/b62e618c-27e9-456a-a0d2-363a27d46755")`

it error when try to execute it, error not special the:

Task exception was never retrieved
future: <Task finished name='Task-11' coro=<Connection.run.<locals>.init() done, defined at C:\Users\sunme\AppData\Local\Programs\Python\Python310\lib\site-packages\playwright\_impl\_connection.py:174> exception=Error('Playwright connection closed')>
Traceback (most recent call last):
  File "C:\Users\sunme\AppData\Local\Programs\Python\Python310\lib\site-packages\playwright\_impl\_connection.py", line 175, in init
    self.playwright_future.set_result(await self._root_object.initialize())
  File "C:\Users\sunme\AppData\Local\Programs\Python\Python310\lib\site-packages\playwright\_impl\_connection.py", line 135, in initialize
    await self._channel.send(
  File "C:\Users\sunme\AppData\Local\Programs\Python\Python310\lib\site-packages\playwright\_impl\_connection.py", line 39, in send
    return await self.inner_send(method, params, False)
  File "C:\Users\sunme\AppData\Local\Programs\Python\Python310\lib\site-packages\playwright\_impl\_connection.py", line 63, in inner_send
    result = next(iter(done)).result()
  File "C:\Users\sunme\AppData\Local\Programs\Python\Python310\lib\site-packages\playwright\_impl\_browser_type.py", line 203, in connect
    playwright: "Playwright" = next(iter(done)).result()
playwright._impl._api_types.Error: Playwright connection closed

my electron is start with argument --remote-debugging-port=8768, I want to know if playwright has such feature to attach existing electron app? if not, is any plan to support in future?

aslushnikov:
@sunmeixia2005 try using [`chromium.connect_over_cdp`](https://playwright.dev/python/docs/api/class-browsertype/#browser-type-connect-over-cdp) instead

sunmeixia2005:
> @sunmeixia2005 try using [`chromium.connect_over_cdp`](https://playwright.dev/python/docs/api/class-browsertype/#browser-type-connect-over-cdp) instead
Thanks for replying.

I try 3 different ways, all of them give some errors:
[...]

PR #20094, opened 2023-01-13
feat(electron): add connectOverCDP function to playwright._electron

ZFail:
Added connectOverCDP functionality to playwright._electron, requested at https://github.com/microsoft/playwright/issues/10369


ZFail:

mastrzyz:
Love This!

packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -114,6 +120,28 @@ export class ElectronApplication extends SdkObject {
   }
 }
 
+async function urlToNodeWSEndpoint(progress: Progress, endpointURL: string) {
+  if (endpointURL.startsWith('ws'))
+    return endpointURL;
+  progress.log(`<ws preparing> retrieving websocket url from ${endpointURL}`);
+  const httpURL = endpointURL.endsWith('/') ? `${endpointURL}json/` : `${endpointURL}/json/`;
```
macabeus, line 127: Please, remove the last `/`.
```suggestion
  const httpURL = endpointURL.endsWith('/') ? `${endpointURL}json` : `${endpointURL}/json`;
```

For some reason, the `json/` endpoint here returns 404: 
image

But without the last slash works:
image

macabeus:
I tried to use this branch, but I'm always receiving the following error when trying to evaluate anything on the page:
```
JavaScriptErrorInEvaluate: ReferenceError: process is not defined
```

The following commands returned it:
```js
const result = await electronApp.evaluate(() => {
  return 1 + 2;
})
```
```js
const window = await electronApp.firstWindow();
console.log(await window.title());
```

Despite that, `electronApp.windows()` returned the right amount of pages.

ZFail:
> ```js
> const result = await electronApp.evaluate(() => {
>   return 1 + 2;
> })
> ```

on my machine this code work and returns 3, i`m add this to my test https://github.com/microsoft/playwright/pull/20094/files#diff-25a018cac50f8f5e3a9f2a7691a61391be2e5995e74d9cc03a058be43163a795R230

macabeus:
I'm working on macOS

ZFail:
> I'm working on macOS

on macos, the test also passes, can you please run the test with the command `npx playwright test electron-app.spec.ts -g "^(.*[\\\/])?electron-app.spec.ts.should connect to external electron$"`

macabeus:
The test passes. Perhaps it's failing here because of the electron application itself

jiby-aurum:
@ZFail just tried it, works well, thanks for the work.

jiby-aurum:
@dgozman @aslushnikov is this something the core team would be willing to land. Would love to use it from upstream

pavelfeldman:
I don't think we are comfortable committing to the over-CDP support for Electron. Playwright needs to connect to both Node and the browser in a specific order and we are deferring the electron activation to make sure there are no internal races. Connecting over CDP requires us to add similar dance around the races that we would not want to maintain.

PR #20110, opened 2023-01-13
chore: merge the util exports

pavelfeldman:


packages/playwright-core/src/client/fetch.ts
```diff
@@ -336,4 +336,17 @@ function isJsonContentType(headers?: HeadersArray): boolean {
       return value === 'application/json';
   }
   return false;
-}
\ No newline at end of file
+}
+
+function objectToArray(map?:  { [key: string]: any }): NameValue[] | undefined {
```
dgozman, line 341: There is `envObjectToArray` that does pretty much the same.

packages/playwright-core/src/utils/sha1.ts
```diff
@@ -0,0 +1,57 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as crypto from 'crypto';
+import * as URL from 'url';
+
+export { makeWaitForNextTask } from './task';
```
dgozman, line 20: These exports seem strange!

packages/playwright-core/src/utils/sha1.ts
```diff
@@ -0,0 +1,57 @@
...
+export { isString, isRegExp, isObject, isError, isLikelyNpxGlobal } from './rtti';
+export { getFromENV, getAsBooleanFromENV } from './env';
+export { headersObjectToArray, headersArrayToObject } from './headers';
+
+export function monotonicTime(): number {
+  const [seconds, nanoseconds] = process.hrtime();
+  return seconds * 1000 + (nanoseconds / 1000 | 0) / 1000;
+}
+
+export function calculateSha1(buffer: Buffer | string): string {
```
dgozman, line 31: Isn't this one in `cyrpto.ts`? Overall, looks like an accidental file.

dgozman (approved):

packages/playwright-core/src/client/fetch.ts
pavelfeldman: But this one is used for fetch params...

Issue #20121, opened 2023-01-14
[BUG] TypeError when passing array as children in a component test


cassus:
### System 
- Playwright Version: 
  - "@playwright/experimental-ct-react": "1.29.0",
  - "@playwright/test": "1.29.0",
 - OS: macOS 13.1
 - Memory: 56.48 MB / 16.00 GB
 - Node: 18.13.0 - ~/.nvm/versions/node/v18.13.0/bin/node
- command: `rm -rf playwright/.cache/ ; yarn playwright:test -- src/components/ArrayDisplay/ArrayDisplay.spec.tsx`

### ArrayDisplay.tsx
```typescript
export function ArrayDisplay({ children }: { children: unknown }) {
  return <div>{JSON.stringify({ children })}</div>
}
```

### ArrayDisplay.spec.tsx
```typescript
import { test, expect } from "@playwright/experimental-ct-react"

import { ArrayDisplay } from "./ArrayDisplay"

// 👍🏻 works fine 
test("ArrayDisplay", async ({ mount }) => {
  const comp = await mount(
    <div>
      <ArrayDisplay>[1, 2, 3]</ArrayDisplay>
    </div>,
  )

  await expect(comp).toContainText("3")
})

// 💥 fails with TypeError (note that the passed children is an actual array, not a string)
test("ArrayDisplay 2", async ({ mount }) => {
  const comp = await mount(
    <div>
      <ArrayDisplay>{[1, 2, 3]}</ArrayDisplay>
    </div>,
  )

  await expect(comp).toContainText("3")
})
```

### Error message in command line
```
TypeError: Cannot read properties of undefined (reading 'endsWith')
    at render (http://localhost:3100/assets/index.930b025c.js:602:26
    at http://localhost:3100/assets/index.930b025c.js:620:12
    at Array.map (<anonymous>)
    at render (http://localhost:3100/assets/index.930b025c.js:617:92
    at http://localhost:3100/assets/index.930b025c.js:620:12
    at Array.map (<anonymous>)
    at render (http://localhost:3100/assets/index.930b025c.js:617:92
    at window.playwrightMount (http://localhost:3100/assets/index.930b025c.js:632:19
    at eval (eval at evaluate (:197:30), <anonymous>:24:18)
    at UtilityScript.evaluate (<anonymous>:199:17)
    at src/components/ArrayDisplay/ArrayDisplay.spec.tsx:18:22
```

mxschmitt:
[...]

PR #20125, opened 2023-01-15
fix(ct): solid and react JS as child

sand4rt:
Related to: https://github.com/microsoft/playwright/issues/20121. Think this should also be fixed in Svelte and Vue _(both JS and JSX API)_.

### DIFF
--- a/packages/playwright-ct-react/registerSource.mjs
+++ b/packages/playwright-ct-react/registerSource.mjs
@@ -36,9 +36,11 @@ export function register(components) {
 
 /**
  * @param {Component} component
- * @returns {JSX.Element}
  */
 function render(component) {
+  if (typeof component !== 'object' || Array.isArray(component))
+    return component;
+
   let componentFunc = registry.get(component.type);
   if (!componentFunc) {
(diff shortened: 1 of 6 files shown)


pavelfeldman:
Could you help me understand what this does?

sand4rt:
It was related to this issue: https://github.com/microsoft/playwright/issues/20121. Previously it was not possible to pass js primitives as slot/child:

```typescript
// 👍🏻 works fine 
test("ArrayDisplay", async ({ mount }) => {
  const comp = await mount(
    <div>
      <ArrayDisplay>[1, 2, 3]</ArrayDisplay>
    </div>,
  )

  await expect(comp).toContainText("3")
})

// 💥 fails with TypeError (note that the passed children is an actual array, not a string)
test("ArrayDisplay 2", async ({ mount }) => {
  const comp = await mount(
    <div>
      <ArrayDisplay>{[1, 2, 3]}</ArrayDisplay>
    </div>,
  )

  await expect(comp).toContainText("3")
})
```

pavelfeldman (approved):

PR #20132, opened 2023-01-16
test: fix toIntersectViewport test on WebKit @ Ubuntu 22.04

aslushnikov:
For some reason scrolling inline element into view doesn't
always work.


dgozman:
> For some reason scrolling inline element into view doesn't always work.

Perhaps we need a fix in WK to be able to scroll and click inline elements?

PR #20135, opened 2023-01-16
chore: return Promise<any> in {page,context}.unroute

mxschmitt:
Follow-up to https://github.com/microsoft/playwright/issues/20005.

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -1287,7 +1287,7 @@ A glob pattern, regex pattern or predicate receiving [URL] used to register a ro
 ### param: BrowserContext.unroute.handler
 * since: v1.8
 * langs: js, python
-- `handler` ?<[function]\([Route], [Request]\)>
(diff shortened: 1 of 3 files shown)


docs/src/api/class-browsercontext.md
```diff
@@ -1287,7 +1287,7 @@ A glob pattern, regex pattern or predicate receiving [URL] used to register a ro
 ### param: BrowserContext.unroute.handler
 * since: v1.8
 * langs: js, python
-- `handler` ?<[function]\([Route], [Request]\)>
+- `handler` ?<[function]\([Route], [Request]\): [Promise<any>|any]>
```
yury-s, line 1290: Why is return type not just `any` ?

yury-s (approved):

docs/src/api/class-browsercontext.md
mxschmitt: See https://github.com/microsoft/playwright/pull/20005

PR #20148, opened 2023-01-16
docs: improve intro screenshots

debs-obrien:

PR #20153, opened 2023-01-17
docs(running-tests): show command to enable the HTML reporter

### LINKED ISSUE #20152: [Docs] Introduce the `test --reporter=html` command prior to `show-report` command
### Issue I ran into

While going through the running-tests-js.md file, I tried to run the `show-report` command. I got an error:

```
npx playwright show-report
No report found at "/src/experiment-playwright-trace-viewer/playwright-report"
```

The instructions did not introduce the `--reporter=html` option  prior to running `show-report` command.

### What should happen

Show the `npx playwright test --reporter=html` command prior to the `npx playwright show-report` command.

cnishina:
fixes #20152

### DIFF
--- a/docs/src/running-tests-js.md
+++ b/docs/src/running-tests-js.md
@@ -90,11 +90,20 @@ Check out our [debugging guide](./debug.md) to learn more about the [Playwright 
 
 ## Test Reports
 
-The [HTML Reporter](././test-reporters.md#html-reporter) shows you a full report of your tests allowing you to filter the report by browsers, passed tests, failed tests, skipped tests and flaky tests. By default, the HTML report is opened automatically if some of the tests failed.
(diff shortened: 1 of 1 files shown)


cnishina:
@microsoft-github-policy-service agree [company="Google"]

debs-obrien:
Hi @cnishina 

as per comment on issue posting here and closing:

For projects created with `create-playwright` the HTML report is enabled by default. If you have a config which does not use the HTML report or running the show-report command yields to no reports, you can run it with `--reporter=HTML`.

I have added this comment to the docs at the very end as this is a getting started guide the majority of people should be following this guide from the start and will therefore have Playwright installed with create-playwright and therefore won't run into this issue. They can just use `npx playwright show-report` with no flag needed. I prefer not to have everyone use this flag as default when it is not necessary.

For those that are more advanced and have played with the config or have an already installed project that they are working on then they may come across this and therefore have added the comment above. 

You can see pr here which also has some other minor improvements: https://github.com/microsoft/playwright/pull/20148/files

PR #20158, opened 2023-01-17
fix: filter out pwt stacks correctly

### LINKED ISSUE #20155: [BUG] Codegen jumping to captureStackTrace on assertion steps
**Context:**
- Playwright Version: 1.29.2
- Operating System: Mac
- Node.js version: 14
- Browser: All

**Describe the bug**

When I am on any assertion, it takes me to `captureStackTrace()`. I would prefer if it showed the execution line in my source/test file.

Please clone the following repo: https://github.com/Nav-2d/pw-codegen-repro and run the test using `npx playwright test --debug`. Once on the assertion step, notice that the inspector jumps to `captureStackTrace()`
Screen Shot 2023-01-16 at 7 26 41 PM

Related discussion in https://github.com/microsoft/playwright/issues/19083 & https://github.com/microsoft/playwright/issues/19087

cc - @aslushnikov

mxschmitt:
Not sure where to write a test for that.

The issue was that `TEST_DIR_LIB` resolved to:

```
C:\Users\maxschmitt\development\tmp\pw-codegen-repro\node_modules\@playwright\test\node_modules\@playwright\test
```

instead of:

```
C:\Users\maxschmitt\development\tmp\pw-codegen-repro\node_modules\@playwright\test\lib
```

So `C:\Users\maxschmitt\development\tmp\pw-codegen-repro\node_modules\@playwright\test\lib\util.js` did not get filtered out.

Fixes https://github.com/microsoft/playwright/issues/20155

### DIFF
--- a/packages/playwright-core/src/utils/stackTrace.ts
+++ b/packages/playwright-core/src/utils/stackTrace.ts
@@ -30,11 +30,9 @@ export function rewriteErrorMessage<E extends Error>(e: E, newMessage: string): 
 const CORE_DIR = path.resolve(__dirname, '..', '..');
 const CORE_LIB = path.join(CORE_DIR, 'lib');
 const CORE_SRC = path.join(CORE_DIR, 'src');
-const TEST_DIR_SRC = path.resolve(CORE_DIR, '..', 'playwright-test');
 const COVERAGE_PATH = path.join(CORE_DIR, '..', '..', 'tests', 'config', 'coverage.js');
 
 const stackFilters = [
-  (frame: StackFrame) => frame.file.startsWith(TEST_DIR_SRC),
   (frame: StackFrame) => frame.file.startsWith(CORE_DIR),
 ];
 export const addStackFilter = (filter: (frame: StackFrame) => boolean) => stackFilters.push(filter);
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/utils/stackTrace.ts
```diff
@@ -31,7 +31,12 @@ const CORE_DIR = path.resolve(__dirname, '..', '..');
 const CORE_LIB = path.join(CORE_DIR, 'lib');
 const CORE_SRC = path.join(CORE_DIR, 'src');
 const TEST_DIR_SRC = path.resolve(CORE_DIR, '..', 'playwright-test');
-const TEST_DIR_LIB = path.resolve(CORE_DIR, '..', '@playwright', 'test');
+let TEST_DIR_LIB = '';
```
pavelfeldman, line 34: I find it strange that core depends on the @playwright/test assumptions. It would be better for @playwright/test to use semi-private core api to set its location instead. We already export stackTrace to test, so pwt could say addStackFilter(require.resolve('./')) in the index.ts fixtures file.

packages/playwright-test/src/index.ts
```diff
@@ -29,6 +29,8 @@ export { addRunnerPlugin as _addRunnerPlugin } from './plugins';
 export const _baseTest: TestType<{}, {}> = rootTestType.test;
 export const store = _baseStore;
 
+addStackIgnoreFilter((frame: StackFrame) => frame.file.startsWith(path.join(__dirname, '..')));
```
pavelfeldman, line 32: nit: require.resolve('../package.json') would be easier to reason about if this file is moved into a subfolder.

pavelfeldman (approved):

PR #20178, opened 2023-01-18
docs(test-parameterize): fix typos

rfdonnelly:
Fixed additional typos of parameterized missed by 80235c47a5fb142c3b21fbbb9350de81b60e6faf

Signed-off-by: Rob Donnelly <rfdonnelly@gmail.com>

### DIFF
--- a/docs/src/test-parameterize-js.md
+++ b/docs/src/test-parameterize-js.md
@@ -1,9 +1,9 @@
 ---
 id: test-parameterize
-title: "Parametrize tests"
+title: "Parameterize tests"
 ---
 
-You can either parametrize tests on a test level or on a project level.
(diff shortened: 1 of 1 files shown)


pavelfeldman:
Both are correct and recognized by the dictionaries, also https://trends.google.com/trends/explore?geo=US&q=parametrize,parameterize. So does not seem to be a big issue!

rfdonnelly:
TIL! In that case, the PR should read, "Use parameterize consistently instead of mixing equally valid spellings parameterize and parametrize." 😉

pavelfeldman:
TIL we used "parameterize" somewhere :)

pavelfeldman (approved):

PR #20191, opened 2023-01-18
docs: Added types to custom reporter typescript example

ValorHeart:
Signed-off-by: Adi <61405630+ValorHeart@users.noreply.github.com>

### DIFF
--- a/docs/src/test-reporter-api/class-reporter.md
+++ b/docs/src/test-reporter-api/class-reporter.md
@@ -34,22 +34,22 @@ module.exports = MyReporter;
 
 ```js tab=js-ts
 // my-awesome-reporter.ts
-import { Reporter } from '@playwright/test/reporter';
+import { Reporter, FullConfig, Suite, TestCase, TestResult, FullResult } from '@playwright/test/reporter';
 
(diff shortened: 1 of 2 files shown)


ValorHeart:
Not sure what is the reason regarding 'tree is dirty' error, will be glad for assistance with that.

mxschmitt:
You need to run `npm ci && npm run build` to re-generate types.

ValorHeart:

ValorHeart:
> You need to run `npm ci && npm run build` to re-generate types.

Thanks! Fixed

pavelfeldman:
Aren't all these types inferred? We don't annotate inferred type names, IDEs and compilers will do the right thing.

pavelfeldman (approved):

pavelfeldman:
Actually, I approved it too soon. We don't like explicit type annotation when they can be inferred. Could you show the entirety of the file that WebStrom could not infer?

ValorHeart:
@pavelfeldman Attached example project, notice how the auto-complete doesn't work
[example.zip](https://github.com/microsoft/playwright/files/10527657/example.zip)

Issue #20190, opened 2023-01-18
[BUG] The method 'isChecked' return incorrect value


Sapendo:
**Context:**
- Playwright Version: 1.29.2
- Operating System: Windows
- Node.js version: 16.15.0
- Browser: WebKit

**Code Snippet**
```javascript
expect(await page.locator('stj-checkbox-demo').locator('stj-checkbox').nth(0).locator('input').isChecked()).toBe(true);
```

**Describe the bug**

There is a checkbox that has both a checked and an indeterminate state. Before updating the version of the playwright to version 1.29.2 the test worked correctly and returned the value as 'true' if we checked with 'locator.isChecked()'. After updating returned value is 'false'. However, if we look at the state of the element in devtools we see that the checked state equals 'true'.
This happens only if the checkbox has both states as 'true'. If, for example, the indeterminate state equals 'false' and the checked state equals 'true' the test will pass.

![image](https://user-images.githubusercontent.com/24610132/213172270-e1ad6695-2b76-4cf8-a07d-6c10c839a872.png)

yury-s:
According to the accessibility tree the element's state is `checked: mixed`:

![image](https://user-images.githubusercontent.com/9798949/213331339-a35d6926-b5de-44e1-b452-feff819fc46e.png)

So the current behavior is actually correct.

Sapendo:
According to the accessibility tree the element's state is `checked: true`:

![image](https://user-images.githubusercontent.com/24610132/213385156-c238fc2e-66f4-4a30-bfab-a57ba0540305.png)

yury-s:
I wonder why we see different value in devtools, can you share a repro so that we can try it locally?

> Moreover, I do not understand why the method isChecked look at the accessibility property. If you take value from the accessibility property, the method isChecked should return one of the next values: 'checked (true)', 'unchecked (false)', or 'mixed'.

If isChecked returned 3-state value we'd return mixed in this case. Since indeterminate element is a grey area, we consider everything that is not true to be false. What you suggest is basically consider everything that is not false in the ax tree to be true.

yury-s:
> I wonder why we see different value in devtools, can you share a repro so that we can try it locally?

@dgozman pointed out that if you are inspecting the element in the trace then it most likely doesn't have indeterminate property as we don't restore it.

Sapendo:
[...]

PR #20209, opened 2023-01-19
fix: indeterminate input is unchecked

yury-s:
* Also, prioritize 'aria-checked' attribute above all if it's present, this aligns the behavior with what DevTools show.

https://github.com/microsoft/playwright/issues/20190

### DIFF
--- a/tests/page/locator-convenience.spec.ts
+++ b/tests/page/locator-convenience.spec.ts
@@ -146,11 +146,11 @@ it('isChecked should work', async ({ page }) => {
 
 it('isChecked should work for indeterminate input', async ({ page }) => {
   it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/20190' });
-  it.fixme()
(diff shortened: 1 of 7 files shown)


tests/page/locator-convenience.spec.ts
```diff
@@ -144,6 +144,15 @@ it('isChecked should work', async ({ page }) => {
   expect(error.message).toContain('Not a checkbox or radio button');
 });
 
+it('isChecked should work for indeterminate input', async ({ page }) => {
+  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/20190' });
+  await page.setContent(`<input type="checkbox" checked >`);
+  expect(await page.locator('input').isChecked()).toBe(true);
```
pavelfeldman, line 150: Are we documenting that isChecked returns false for mixed?

yury-s: Done.

PR #20235, opened 2023-01-19
chore: extract pool builder, simplify project suite cloning

pavelfeldman:


packages/playwright-test/src/suiteUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+* limitations under the License.
+*/
+
+import path from 'path';
+import { calculateSha1 } from 'playwright-core/lib/utils';
+import type { Suite, TestCase } from './test';
+import type { FullProjectInternal } from './types';
+
+export function filterTests(suite: Suite, filter: (test: TestCase) => boolean): boolean {
+  for (const child of suite.suites) {
```
dgozman, line 23:
```js
suite.suites = suite.suites.filter(child => filterTests(child, filter));
```

dgozman (approved):

PR #20241, opened 2023-01-20
chore: use fake pool on the runner side

pavelfeldman:


packages/playwright-test/src/testLoader.ts
```diff
@@ -34,7 +34,7 @@ export class TestLoader {
     this._fullConfig = fullConfig;
   }
 
-  async loadTestFile(file: string, environment: 'runner' | 'worker'): Promise<Suite> {
+  async loadTestFile(file: string, environment: 'loader' | 'worker'): Promise<Suite> {
```
dgozman, line 37: Unrelated, but perhaps we should take `loadErrors` array here as in other places, instead of setting `suite._loadError` flag?

tests/playwright-test/fixtures.spec.ts
```diff
@@ -714,30 +714,3 @@ test('worker teardown errors reflected in timed-out tests', async ({ runInlineTe
   expect(result.output).toContain('Test timeout of 1000ms exceeded.');
   expect(result.output).toContain('Rejecting!');
 });
-
-test('should not complain about fixtures of non-focused tests', async ({ runInlineTest }) => {
```
dgozman, line 718: We can keep this functionality by splitting `buildPools` and `validateSignatures` steps and filtering in-between.

packages/playwright-test/src/runner.ts
```diff
@@ -346,13 +338,29 @@ export class Runner {
...
+  private async _loadTests(testFiles: Set<string>): Promise<{ rootSuite: Suite, loadErrors: TestError[] }> {
+    const config = this._configLoader.fullConfig();
+    const testLoader = new TestLoader(config);
+    const loadErrors: TestError[] = [];
+    const rootSuite = new Suite('', 'root');
+    for (const file of testFiles) {
+      const fileSuite = await testLoader.loadTestFile(file, 'loader');
+      if (fileSuite._loadError)
+        loadErrors.push(fileSuite._loadError);
+      // We have to clone only if there maybe subsequent calls of this method.
```
dgozman, line 356: Not sure what this comment means today :)

pavelfeldman: Not mine :)

packages/playwright-test/src/testLoader.ts
pavelfeldman: yeah, i can follow up

tests/playwright-test/fixtures.spec.ts
pavelfeldman: This won't work when loader is out of process sadly.

dgozman (approved):

PR #20255, opened 2023-01-20
feat: add clipbord copy to test title

thohe973:
When a test fails, you often want to search for it in your code base.
This adds a copy icon to the test title so you can quickly copy it.


thohe973:

pavelfeldman:
The code looks good, please attach the screenshots of the title w/ and w/o hover to verify UI!

thohe973:
image

thohe973:
Updated to the latest codicon version that has a copy icon.

image

pavelfeldman:
There is a test failure at https://github.com/microsoft/playwright/actions/runs/4025083649/jobs/6917856418, could you take a look? The web components one.

mahtabnejad90 (commented):
Can we add some context on this feature please in the PR description.

pavelfeldman:
@thohe973 sorry for the slow review. Could you remove the new codicon and only add an icon? We'll need to add the fonts file ourselves in order to ensure its authenticity.

thohe973:
@pavelfeldman I reverted the codicon commit and went back to the old clippy icon. I also fixed the test failure.

packages/html-reporter/src/testCaseView.tsx
```diff
@@ -14,6 +14,7 @@
   limitations under the License.
 */
 
+import { CopyToClipboard } from '@web/components/copyToClipboard';
```
pavelfeldman, line 17: Oh-ah, sorry for not spotting this earlier. I thought the change modifies trace viewer, but in fact it changes the html report. Trace Viewer uses VS Code design language (@web), while HTML Reporter uses GitHub design language, does not depend on @web. So if you are interested in the change to the html report, you should use svg for images, similarly to everything else in the HTML report code.

thohe973: Ah, didn't know that two different design languages where used. I thought @web was a "common area".
I'll close this PR and open a new one since I will need to undo most of the changes.

PR #20258, opened 2023-01-20
feat(trace-viewer): allow host and port to be specified

gobeli:
Currently the `--host` and `--port` cli options are only available for `show-report`. 
This change would allow the two options to be passed to the `show-trace` command.

### DIFF
--- a/packages/playwright-core/src/cli/cli.ts
+++ b/packages/playwright-core/src/cli/cli.ts
@@ -298,6 +298,8 @@ program
 program
     .command('show-trace [trace...]')
     .option('-b, --browser <browserType>', 'browser to use, one of cr, chromium, ff, firefox, wk, webkit', 'chromium')
+    .option('-h, --host <host>', 'Host to serve trace on', 'localhost')
+    .option('-p, --port <port>', 'Port to serve trace on', '9323')
     .description('show trace viewer')
     .action(function(traces, options) {
       if (options.browser === 'cr')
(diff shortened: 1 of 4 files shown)


gobeli:

packages/playwright-core/src/cli/cli.ts
```diff
@@ -298,6 +298,8 @@ program
 program
     .command('show-trace [trace...]')
     .option('-b, --browser <browserType>', 'browser to use, one of cr, chromium, ff, firefox, wk, webkit', 'chromium')
+    .option('-h, --host <host>', 'Host to serve trace on', 'localhost')
+    .option('-p, --port <port>', 'Port to serve trace on', '9323')
```
pavelfeldman, line 302: This is a different value from what it used to be.

packages/playwright-core/src/server/trace/viewer/traceViewer.ts
```diff
@@ -26,7 +26,7 @@ import { serverSideCallMetadata } from '../../instrumentation';
 import { createPlaywright } from '../../playwright';
 import { ProgressController } from '../../progress';
 
-export async function showTraceViewer(traceUrls: string[], browserName: string, headless = false, preferredPort?: number): Promise<BrowserContext | undefined> {
+export async function showTraceViewer(traceUrls: string[], browserName: string, headless = false, host?: string, preferredPort?: number): Promise<BrowserContext | undefined> {
```
pavelfeldman, line 29: Let's turn this into the options: { headless?: boolean, host?: string, port?: number }

pavelfeldman (commented):
Looks good, a couple of small comments.

packages/playwright-core/src/cli/cli.ts
gobeli: At `9322` now

packages/playwright-core/src/server/trace/viewer/traceViewer.ts
gobeli: Options object is now used for `headless`, `host` and `port`

pavelfeldman (approved):

PR #20272, opened 2023-01-21
feat: expect toBeFast

JoelEinbinder:
```ts
await expect(async () => {
  // some code that might be fast
}).toBeFast({ threshold: 3000 });
```

Throws if the code doesn't run in the `threshold` amount of time. Extra options of `warnThreshold` and `description` control if and how the timing shows up in the test report.


JoelEinbinder:

pavelfeldman:
How about we add a timeout to the `test.step` instead? That would add the feature you need, but won't encourage broader audience to think about performance testing this way.

PR #20280, opened 2023-01-22
chore: add type checking

wopperer:


wopperer:

packages/html-reporter/src/types.ts
```diff
@@ -59,12 +59,19 @@ export type TestCaseSummary = {
   projectName: string;
   location: Location;
   annotations: { type: string, description?: string }[];
-  outcome: 'skipped' | 'expected' | 'unexpected' | 'flaky';
+  outcome: keyof typeof Outcome;
```
dgozman, line 62: We are using [string literal types](https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types) instead of enums, which already gives us type checking.

dgozman (commented):
Thank you for the PR! However, we prefer to not use enums, while keep everything type-checked with string literal types.

packages/html-reporter/src/types.ts
wopperer: You're welcome, but actually its more about the language SourceProp which is not type checked properly. Only for a string.

packages/web/src/components/codeMirrorWrapper.tsx
```diff
@@ -27,9 +27,16 @@ export type SourceHighlight = {
...
+  javascript = 'javascript',
+  python = 'python',
+  java = 'java',
+  csharp = 'csharp'
+}
+
 export interface SourceProps {
   text: string;
-  language: string;
+  language: keyof typeof Language;
```
pavelfeldman, line 39: Our preferred way would be:

```js
type Language = 'javascript' | 'python' | 'java' | 'csharp';
```

and then use string literals in the code

wopperer: yes makes it more aligned with the rest, I changed it!

pavelfeldman (approved):

Issue #19694, opened 2022-12-25
[Feature] defineConfig helper for playwright config file


sand4rt:
Can we have a `defineConfig` helper? A lot of libraries _[(like vite)](https://vitejs.dev/config/#config-intellisense)_ have this and seems to be the standard. It's a little cleaner and provides intellisense without the need for jsdoc annotations when js is used:

```ts
type PlaywrightTestConfigProps = {
    mode: 'CI';
}

function defineConfig(config: PlaywrightTestConfig): PlaywrightTestConfig;
function defineConfig((props: PlaywrightTestConfigProps) => PlaywrightTestConfig): PlaywrightTestConfig;
```

```typescript
// playwright.config.ts
import { devices, defineConfig } from '@playwright/experimental-ct-solid';

export default defineConfig({
  retries: process.env.CI ? 2 : 0,
  projects: [
    {
      name: 'chromium',
      use: { ...devices['Desktop Chrome'] },
    },
  ],
});
```

Conditional Config:

```typescript
// playwright.config.ts
import { devices, defineConfig } from '@playwright/experimental-ct-solid';

export default defineConfig(({ mode }) => ({
  retries: mode === 'CI' ? 2 : 0,
  projects: [
    {
      name: 'chromium',
      use: { ...devices['Desktop Chrome'] },
    },
  ],
}));
```

Old way can still exist, but will be good to modify the playwright-create project if this gets accepted:

```typescript
// playwright.config.ts
import { type PlaywrightTestConfig, devices } from '@playwright/experimental-ct-solid';

const config: PlaywrightTestConfig = {
  retries: process.env.CI ? 2 : 0,
  projects: [
    {
      name: 'chromium',
      use: { ...devices['Desktop Chrome'] },
    },
  ],
};

export default config;
```

dgozman:
@sand4rt How is the conditional config going to be used? Do you have an example in mind? Who is passing the arguments to the function? What could be there?

Apart from conditionals, this seems to only save a single line with the type declaration, in the config file that is already generated by `create-playwright`, so doesn't look worthwhile.

sand4rt:
Feature request was mainly about the first `defineConfig`, but it could retrieve the flag if it's running in CI or not _(aka `process.env.CI`)_. Any config option can be injected into it in the future if needed. 

```ts
// Simple usage
export default defineConfig(({ mode }) => ({
  retries: mode === 'CI' ? 2 : 0,
}));

// Advanced usage
[...]

PR #20288, opened 2023-01-23
Ct vite config async

yeion7:
This PR adds the ability to `ctViteConfig` be async 

This allows read the vite config file by importing the `loadConfigFromFile` function from the `vite` package and using it within the `ctViteConfig` function. The config object is then returned, in this way it's possible to fully reuse the existing vite config

```
import { loadConfigFromFile } from "vite"

const config = {
    use: {
        async ctViteConfig() {
            const { config } = await loadConfigFromFile(
                resolve(__dirname, "./vite.config.js")
            );

            return config
        }
    }
}

export default config
```

### DIFF
--- a/packages/playwright-ct-react/index.d.ts
+++ b/packages/playwright-ct-react/index.d.ts
@@ -30,7 +30,7 @@ export type PlaywrightTestConfig<T = {}, W = {}> = Omit<BasePlaywrightTestConfig
     ctPort?: number;
     ctTemplateDir?: string;
     ctCacheDir?: string;
-    ctViteConfig?: InlineConfig;
+    ctViteConfig?: InlineConfig | (() => Promise<InlineConfig>);
   };
 };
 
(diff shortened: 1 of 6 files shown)


yeion7:
> please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.

sand4rt:
Related to: https://github.com/microsoft/playwright/issues/19694

pavelfeldman:
You should be able to do the following:

```js
import { loadConfigFromFile } from "vite"

const config = {
    use: {
        ctViteConfig: async ({}, use) => {
            const { config } = await loadConfigFromFile(
                resolve(__dirname, "./vite.config.js")
            );
            await use(config);
        });
    }
}

export default config
```

yeion7:
> You should be able to do the following:

Interesting thing, I would not have imagined that it could be done reading the type definition of `ctViteConfig`

I think part of the inspiration of this is, other tools that have integration with vite, follow this kind of format, for example storybook

https://github.com/storybookjs/builder-vite#customize-vite-config

pavelfeldman (approved):

PR #20305, opened 2023-01-23
fix(test runner): show deep strack traces during imports

### LINKED ISSUE #20239: [Question] How to show import resolution stack?
I have made some changes to a project which causes playwright to be unable to run tests, due to importing a dependency that can not be run in node.

I run playwright with `npx playwright test someFolder/`, the error is: 
```
ReferenceError: document is not defined

   at ../src/modules/izitoast.ts:2

  1 | import { boot } from 'quasar/wrappers'
> 2 | import iziToast from 'izitoast'
    | ^
  3 | import 'izitoast/dist/css/iziToast.min.css'
  4 | import {IziToastPosition} from 'izitoast/types/index'
  5 |

    at /home/david/dev-environment/quasar/node_modules/izitoast/dist/js/iziToast.js:23:10
    at /home/david/dev-environment/quasar/node_modules/izitoast/dist/js/iziToast.js:10:20
    at Object.<anonymous> (/home/david/dev-environment/quasar/node_modules/izitoast/dist/js/iziToast.js:14:3)
    at Object.<anonymous> (/home/david/dev-environment/quasar/src/modules/izitoast.ts:2:1)

```

This error is almost helpful, but it doesn't seem to go "all the way", to telling me why the `izitoast` module was imported. Is there a way to convince playwright to output the full chain of files that led to this undesired import?

dgozman:
Fixes #20239.

### DIFF
--- a/packages/playwright-test/src/transform.ts
+++ b/packages/playwright-test/src/transform.ts
@@ -38,8 +38,7 @@ type ParsedTsConfigData = {
 };
 const cachedTSConfigs = new Map<string, ParsedTsConfigData | undefined>();
 
-const kStackTraceLimit = 15;
-Error.stackTraceLimit = kStackTraceLimit;
+Error.stackTraceLimit = 200;
 
 sourceMapSupport.install({
   environment: 'node',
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/transform.ts
```diff
@@ -219,13 +219,17 @@ export function transformHook(code: string, filename: string, moduleUrl?: string
 export async function requireOrImport(file: string) {
   const revertBabelRequire = installTransform();
   const isModule = fileIsModule(file);
+  const oldStackTraceLimit = Error.stackTraceLimit;
   try {
+    // Increase stack trace limit because import stacks are very deep.
+    Error.stackTraceLimit = 200;
```
yury-s, line 225: Can we consolidate it with the other place where we set it to 15?

yury-s (approved):

PR #20346, opened 2023-01-24
fix(test runner): do not send entirely skipped test groups to a worker

### LINKED ISSUE #20156: [Question] test.skip() still creates separate thread
Hi,
Is there a good reason why tests skipped with `test.skip()` still get allocated a separate worker thread?
I get that the test gets marked as skipped in the report, but that feels like a post run activity that updates the test results rather than an action requiring a separate worker thread.

This becomes very apparent, and very inefficient, if you have a spec file with, for example:
```
test('test1', () => {})
test.skip('test2', () => {})
```

And then execute it with `--repeat-each`. If you repeat it 6 times, you end up with 12 worker threads. So this grows (linearly) as the number of repeats increases.

This may turn into a feature request or bug ticket.

dgozman:
Move TestCase properties calculation from WorkerMain to suite building phase.

Fixes #20156.

### DIFF
--- a/packages/playwright-test/src/dispatcher.ts
+++ b/packages/playwright-test/src/dispatcher.ts
@@ -65,7 +65,34 @@ export class Dispatcher {
     }
   }
 
+  private _processFullySkippedJobs() {
+    // If all the tests in a group are skipped, we report them immediately
+    // without sending anything to a worker. This avoids creating unnecessary worker processes.
+    //
+    // However, if there is at least one non-skipped test in a group, we'll send
+    // the whole group to the worker process and report tests in the natural order,
+    // with skipped tests mixed in-between non-skipped. This makes
+    // for a better reporter experience.
+    while (!this._isStopped && this._queue.length) {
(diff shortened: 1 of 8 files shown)


packages/playwright-test/src/dispatcher.ts
```diff
@@ -65,7 +65,33 @@ export class Dispatcher {
...
+    // with skipped tests mixed in-between non-skipped. This makes
+    // for a better reporter experience.
+    while (!this._isStopped && this._queue.length) {
+      const group = this._queue[0];
+      const allTestsSkipped = group.tests.every(test => test.expectedStatus === 'skipped');
+      if (!allTestsSkipped)
+        break;
+
+      for (const test of group.tests) {
+        const result = test._appendTestResult();
```
yury-s, line 83: call `result.status = 'skipped';` explicitly?

yury-s (approved):

yury-s (approved):

PR #20355, opened 2023-01-25
fix(timeout): add additional validation for timeout

sohbat17:
- Check if the provided timeout is greater than 0 before returning static timeout.
- Check if the provided timeout is greater than 0 before returning timeout.
- Check if the provided timeout is greater than 0 before returning launchTimeout.


mxschmitt:
How can it ever be negative?

pavelfeldman:
Please start with filing your request as an issue and explaining the use case in detail!

PR #20374, opened 2023-01-25
feat: Restrain sending http credentials on a specific origin

### LINKED ISSUE #20464: [Feature] Restrain sending http credentials on a specific hostname

SebastienRichert:
Fixes https://github.com/microsoft/playwright/issues/20464

Hi there.

For security purpose, we would like to restrain sending HTTP credentials to only the specified server. The idea is to give the ability to specify a origin (scheme://host:port) additionally to current pair username/password. When an authorization response is received from servers, the credentials are sent only if the server origin in the request matches case insensitive the specified origin.

If this PR is OK for you, we are also interested into improving the Java wrapper as a next step.


SebastienRichert:
@microsoft-github-policy-service agree [company="Tricentis"]

SebastienRichert:

pavelfeldman:
Please start with filing your issue and explaining the use case in detail!

packages/playwright-core/src/server/firefox/ffBrowser.ts
```diff
@@ -331,7 +331,7 @@ export class FFBrowserContext extends BrowserContext {
 
   async doSetHTTPCredentials(httpCredentials?: types.Credentials): Promise<void> {
     this._options.httpCredentials = httpCredentials;
-    await this._browser._connection.send('Browser.setHTTPCredentials', { browserContextId: this._browserContextId, credentials: httpCredentials || null });
+    await this._browser._connection.send('Browser.setHTTPCredentials', { browserContextId: this._browserContextId, credentials: httpCredentials && { username: httpCredentials.username, password: httpCredentials.password } || null });
```
pavelfeldman, line 334: Could you elaborate on the semantics of this expression?

```js
httpCredentials && { username: httpCredentials.username, password: httpCredentials.password } || null
```

browser_patches/firefox/juggler/protocol/Protocol.js
```diff
@@ -305,7 +306,11 @@ const Browser = {
     'setHTTPCredentials': {
       params: {
         browserContextId: t.Optional(t.String),
-        credentials: t.Nullable(networkTypes.HTTPCredentials),
+        credentials: t.Nullable({
```
pavelfeldman, line 309: I think you could leave it as is.

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -153,19 +153,34 @@ export class CRNetworkManager {
...
+    const shouldProvideCredentials = this._shouldProvideCredentials(event.request.url);
     if (this._attemptedAuthentications.has(event.requestId)) {
       response = 'CancelAuth';
-    } else if (this._credentials) {
+    } else if (shouldProvideCredentials) {
       response = 'ProvideCredentials';
       this._attemptedAuthentications.add(event.requestId);
     }
-    const { username, password } = this._credentials || { username: undefined, password: undefined };
+    const { username, password } =  shouldProvideCredentials && this._credentials ? this._credentials : { username: undefined, password: undefined };
```
pavelfeldman, line 163: You don't need `&& this._credentials` check here

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -153,19 +153,34 @@ export class CRNetworkManager {
...
     this._client._sendMayFail('Fetch.continueWithAuth', {
       requestId: event.requestId,
       authChallengeResponse: { response, username, password },
     });
   }
 
+  _shouldProvideCredentials(url: string): boolean {
+    if (this._credentials){
```
pavelfeldman, line 171: In JavaScript, we would say

```js
_shouldProvideCredentials(url: string): boolean {
  if (!this._credentials)
    return false;
  return !this._credentrials.hostname || new URL(url).hostname.toLowerCase() === this._credentials.hostname.toLowerCase();
}


return this._credentials;
```

tests/library/browsercontext-credentials.spec.ts
```diff
@@ -64,6 +89,41 @@ it('should fail with wrong credentials', async ({ browser, server }) => {
   await context.close();
 });
 
+it('should fail with wrong credentials and matching hostname', async ({ browser, server }) => {
+  server.setAuth('/empty.html', 'user', 'pass');
+  const hostname = URL.parse(server.PREFIX).hostname;
+  const context = await browser.newContext({
+    httpCredentials: { hostname: hostname, username: 'foo', password: 'bar' }
```
pavelfeldman, line 96: All these tests will pass today, w/o any changes that you were adding. You should assert that the credentials are not sent, in here you actually send wrong credentials and that's why your tests pass. Should you fix it, the tests will start failing. Once Chromium is fixed, you would need to implement it for Firefox and WebKit as well.

SebastienRichert: Right, I fixed the test "should fail with correct credentials and mismatching hostname". Current code wouldn't work with this additional test case.
This test assert that the credentials are not send (because they are correct, and we still got a 401 as response).
 
Then agree the couple other test are not very relevant since it fails anyway because of wrong credentials... do you want me to remove them to keep tests cleaner ?

packages/playwright-core/src/server/chromium/crNetworkManager.ts
SebastienRichert: I agree, this is more condensed, I just fixed it.

packages/playwright-core/src/server/chromium/crNetworkManager.ts
SebastienRichert: Yeah, I did not have it at first, but I ended up adding it because without it the lint check fails (even though the build work). My understanding was that on that line, this._credentials can be null, so in that case it cannot map the username and password from this._credentials to the const {username, password}.

packages/playwright-core/src/server/firefox/ffBrowser.ts
SebastienRichert: My mistake. I'll fix that.

SebastienRichert:
@pavelfeldman thanks for the code review. Really appreciate it !

As discussed, I fixed the tests and reached the point where it works fine with Chromium, and now need to implement the same logic with other browsers. 
Unlike Chromium that has the logic of credential management within the playwright core (crNetworkManager.ts), Firefox and WebKit don't seem to have such capability exposed in playwright core directly, but instead relies on code in browser_patches (in the Firefox juggler for example).
For Firefox, I believe the implementation needs to be in NetworkObserver.js (in the Juggler). I found the specific entry point in charge of re-transmitting request with credentials (after the 401) and did the fix with the hostname verification
https://github.com/microsoft/playwright/pull/20374/commits/640df2d1a1643514efd218fec125e24d27a95513

However, I'm struggling to execute that code. Any changes does not seem to be taken into account... I may be missing an indus step to force the test execution in Firefox to use these updated scripts of the browser_patches. My understanding was that for firefox, we may not need patch the mozilla/gecko-dev (file bootstrap.diff) but only customizing Juggler on the side may be enough... 

Thought ? Is there any documentation on how to deal with this browser_patches ?

Thanks.

pavelfeldman:
We expect contributors to be comfortable building Firefox w/ Juggler and also compiling WebKit on all three platforms, making necessary changes w/ minimal help. Otherwise it is much easier for us to make the changes ourselves.

Otherwise, we'd keep the issue open and wait for a contributor to emerge or for the feature to gain substantial amount of the upvotes.

SebastienRichert:
Make sense yes. 

We had good progress with Firefox, we believe we succeed in building it with juggler, now able to pass the Firefox tests after updating Juggler souce code (in file omni.ja embedded in Firefox folder). We pushed the fix on the PR since all firefox tests pass in local env. But it still fails on Github, I believe because the browser_patches indus is on internal repo (#16845), the update of firefox release is something only you guys can do, right @pavelfeldman ?

Last piece is the Webkit implementation. We'll continue working on that.

pavelfeldman:
Yes, so the juggler file should be landed in a separate PR, we'll help you landing it.

Firehawk99 (approved):

SebastienRichert:
@pavelfeldman We are now all set with the compilation of Webkit patch. We landed that in a separate PR: https://github.com/microsoft/playwright/pull/21196

All tests run successfully locally. Hopefully it will pass here too once Firefox & Webkit are rebuilt on your end.

docs/src/api/params.md
```diff
@@ -530,6 +530,7 @@ Whether to emulate network being offline. Defaults to `false`.
 - `httpCredentials` <[Object]>
   - `username` <[string]>
   - `password` <[string]>
+  - `origin` ?<[string]>
```
mxschmitt, line 533: We need some docs here as well.

packages/playwright-core/src/server/firefox/ffBrowser.ts
```diff
@@ -331,7 +331,12 @@ export class FFBrowserContext extends BrowserContext {
 
   async doSetHTTPCredentials(httpCredentials?: types.Credentials): Promise<void> {
     this._options.httpCredentials = httpCredentials;
-    await this._browser._connection.send('Browser.setHTTPCredentials', { browserContextId: this._browserContextId, credentials: httpCredentials || null });
+    let credentials = null;
+    if (httpCredentials){
+      credentials = { username: httpCredentials.username, password: httpCredentials.password };
+      if (httpCredentials.origin) (credentials as any).origin = httpCredentials.origin;
```
mxschmitt, line 337: lets get rid of this as any hack? It should be now in the types. Maybe revert the changes in this file and only add the origin in line 334.

tests/library/browsercontext-credentials.spec.ts
```diff
@@ -76,3 +77,64 @@ it('should return resource body', async ({ browser, server }) => {
...
+  });
+  const page = await context.newPage();
+  const response = await page.goto(server.EMPTY_PAGE);
+  expect(response.status()).toBe(401);
+  await context.close();
+});
+
+it('should fail with correct credentials and mismatching hostname', async ({ browser, server }) => {
+  server.setAuth('/empty.html', 'user', 'pass');
+  const hostname = URL.parse(server.PREFIX).hostname;
```
mxschmitt, line 119: lets not use URL.parse use `new URL(server.PREFIX).hostname` instead and remove the import.

tests/library/browsercontext-credentials.spec.ts
```diff
@@ -76,3 +77,64 @@ it('should return resource body', async ({ browser, server }) => {
   expect((await response.body()).toString()).toContain('Playground');
   await context.close();
 });
+
+it('should work with correct credentials and matching origin', async ({ browser, server }) => {
+  server.setAuth('/empty.html', 'user', 'pass');
+  const origin = server.PREFIX;
+  const context = await browser.newContext({
+    httpCredentials: { username: 'user', password: 'pass', origin: origin }
```
mxschmitt, line 85:
```suggestion
    httpCredentials: { username: 'user', password: 'pass', origin: server.PREFIX }
```

tests/library/browsercontext-credentials.spec.ts
```diff
@@ -76,3 +77,64 @@ it('should return resource body', async ({ browser, server }) => {
...
+  const response = await page.goto(server.EMPTY_PAGE);
+  expect(response.status()).toBe(200);
+  await context.close();
+});
+
+it('should work with correct credentials and matching origin case insensitive', async ({ browser, server }) => {
+  server.setAuth('/empty.html', 'user', 'pass');
+  const origin = server.PREFIX.toUpperCase();
+  const context = await browser.newContext({
+    httpCredentials: { username: 'user', password: 'pass', origin: origin }
```
mxschmitt, line 97:
```suggestion
    httpCredentials: { username: 'user', password: 'pass', origin: server.PREFIX.toUpperCase() }
```

tests/library/browsercontext-credentials.spec.ts
```diff
@@ -76,3 +77,64 @@ it('should return resource body', async ({ browser, server }) => {
...
+  const response = await page.goto(server.EMPTY_PAGE);
+  expect(response.status()).toBe(200);
+  await context.close();
+});
+
+it('should fail with correct credentials and mismatching scheme', async ({ browser, server }) => {
+  server.setAuth('/empty.html', 'user', 'pass');
+  const origin = server.PREFIX.includes('http://') ? server.PREFIX.replace('http://', 'https://') : server.PREFIX.replace('https://', 'http://');
+  const context = await browser.newContext({
+    httpCredentials: { username: 'user', password: 'pass', origin: origin }
```
mxschmitt, line 109:
```suggestion
    httpCredentials: { username: 'user', password: 'pass', origin: server.PREFIX.replace('http://', 'https://') }
```
server has always `http:` protocol. Drop again the `origin` variable.

packages/playwright-core/src/server/firefox/ffBrowser.ts
SebastienRichert: You're right. I reverted that code. Not sure what you mean by still add the origin in line 334. It seems to work fine without it.

dgozman (commented):
This looks very good, just a few comments!

docs/src/api/class-browsercontext.md
```diff
@@ -1243,6 +1243,9 @@ its geolocation.
 - `httpCredentials` <[null]|[Object]>
   - `username` <[string]>
   - `password` <[string]>
+  - `origin` ?<[string]> Restrain sending http credentials on specific origin (scheme://host:port).
```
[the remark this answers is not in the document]
dgozman: A list of credentials seems feasible, I don't see any blockers.

packages/playwright-core/src/server/webkit/wkPage.ts
```diff
@@ -731,8 +731,10 @@ export class WKPage implements PageDelegate {
   }
 
   async updateHttpCredentials() {
-    const credentials = this._browserContext._options.httpCredentials || { username: '', password: '' };
-    await this._pageProxySession.send('Emulation.setAuthCredentials', { username: credentials.username, password: credentials.password });
+    const credentials = this._browserContext._options.httpCredentials || { username: '', password: '', origin: '' };
+    if (!credentials.origin) credentials.origin = '';
```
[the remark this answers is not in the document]
SebastienRichert: My mistake. Was initially thinking the next call to Emulation.setAuthCredentials needed to have origin initialized to empty if not present, but that's not needed. I removed that line.

docs/src/api/params.md
```diff
@@ -530,8 +530,10 @@ Whether to emulate network being offline. Defaults to `false`.
 - `httpCredentials` <[Object]>
   - `username` <[string]>
   - `password` <[string]>
+  - `origin` ?<[string]> Restrain sending http credentials on specific origin (scheme://host:port).
```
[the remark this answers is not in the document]
SebastienRichert: All set. I updaded the tests as well.

dgozman (approved):
Perfect! Merging in.

dgozman:
@SebastienRichert Driver is usually ready by the next day. However, this will need a roll which requires all the new features to be implemented in Java. You can take a stab at it: https://github.com/microsoft/playwright-java/blob/main/ROLLING.md, but I'd recommend waiting until we roll ourselves.

docs/src/api/class-browsercontext.md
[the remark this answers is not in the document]
SebastienRichert: Idea for list of credentials: https://github.com/microsoft/playwright/issues/22013

SebastienRichert:
Ok, I'll wait for the roll on your side on playwright-java (driver 1.33). 
I've already prepared some test (TestBrowserContextCredentials & TestGlobalFetch), for playwright-java... it is needed or not ? If so should I push now (PR test will fail) or wait until driver are rolled to 1.33 on your end ?

dgozman:
@SebastienRichert Sending a PR never hurts 😄

SebastienRichert:
@dgozman Sounds good 😄. 
https://github.com/microsoft/playwright-java/pull/1253

PR #20379, opened 2023-01-25
fix: better formatting for sparse arrays

### LINKED ISSUE #20347: [BUG]
**Context:**
- Playwright Version: 1.29.2
- Operating System: Mac
- Node.js version: v18.11.0
- Browser: Chromium
- Extra: 

**Code Snippet**

```javascript

(async () => {
    let browser = await playwright.chromium.launch({
        headless: false
    })
    let page = await browser.newPage();
    await page.goto('https://easyimages.net');
}) ();
```

**Describe the bug**

The browser keeps crashing with the following error
```javascript
/Users/johndoe/node_modules/playwright-core/lib/server/chromium/crExecutionContext.js:153
    return '[' + String(result) + ']';
                 ^

RangeError: Invalid string length
    at Array.join (<anonymous>)
    at Array.toString (<anonymous>)
    at String (<anonymous>)
    at renderPreview (/Users/johndoe/node_modules/playwright-core/lib/server/chromium/crExecutionContext.js:153:18)
    at CRExecutionContext.createHandle (/Users/johndoe/node_modules/playwright-core/lib/server/chromium/crExecutionContext.js:108:80)
    at FrameExecutionContext.createHandle (/Users/johndoe/node_modules/playwright-core/lib/server/javascript.js:67:27)
    at FrameExecutionContext.createHandle (/Users/johndoe/node_modules/playwright-core/lib/server/dom.js:84:18)
    at /Users/johndoe/node_modules/playwright-core/lib/server/chromium/crPage.js:692:50
    at Array.map (<anonymous>)
    at FrameSession._onConsoleAPI (/Users/johndoe/node_modules/playwright-core/lib/server/chromium/crPage.js:692:31)
```

I have tried the same code with Puppeteer and the issue was not there. The problem seems to be related to iframes that are loaded within the particular site.

aslushnikov:
Right now arrays preview yields all array elements. In case
of a sparse array with a single element on index 10000000,
this results in a large string that OOM Node.js.

This patch changes pretty-printing. For example:

```ts
// Given this array
const a = [];
a[10] = 1;
// Before this patch, pretty printing will yield:
"[,,,,,,,,1]"
// With this patch, pretty printing yields:
"[empty x 9, 1]"
```

The new array pretty-printing is equal to what Chrome DevTools
do to render sparse arrays.

Fixes #20347

### DIFF
--- a/packages/playwright-core/src/server/javascript.ts
+++ b/packages/playwright-core/src/server/javascript.ts
@@ -329,18 +329,26 @@ export function isJavaScriptErrorInEvaluate(error: Error) {
   return error instanceof JavaScriptErrorInEvaluate;
 }
 
-export function sparseArrayToString(entries: { index: number, value: any }[]): string {
-  let lastIndex = 0;
+export function sparseArrayToString(entries: { key: string, value: any }[]): string {
+  const arrayEntries = [];
+  for (const { name, value } of entries) {
+    const index = +name;
+    if (isNaN(index) || index < 0)
+      continue;
+    arrayEntries.push({ index, value });
+  }
+  arrayEntries.sort((a, b) => a.index - b.index);
+  let lastIndex = -1;
   const tokens = [];
-  for (const { index, value } of entries) {
+  for (const { index, value } of arrayEntries) {
     const emptyItems = index - lastIndex - 1;
     if (emptyItems === 1)
-      tokens.push(`<1 empty item>`);
+      tokens.push(`empty`);
     else if (emptyItems > 1)
-      tokens.push(`<${emptyItems} empty items>`);
+      tokens.push(`empty x ${emptyItems}`);
     tokens.push(String(value));
     lastIndex = index;
   }
 
-  return '[' + tokens.join(',') + ']';
+  return '[' + tokens.join(', ') + ']';
 }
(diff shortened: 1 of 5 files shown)


tests/page/jshandle-to-string.spec.ts
```diff
@@ -32,6 +32,22 @@ it('should work for complicated objects', async ({ page, browserName }) => {
     expect(aHandle.toString()).toBe('JSHandle@object');
 });
 
+it('should beutifully render sparse arrays', async ({ page, browserName }) => {
```
mxschmitt, line 35:
```suggestion
it('should beautifully render sparse arrays', async ({ page, browserName }) => {
```

packages/playwright-core/src/server/javascript.ts
```diff
@@ -328,3 +328,19 @@ export class JavaScriptErrorInEvaluate extends Error {
 export function isJavaScriptErrorInEvaluate(error: Error) {
   return error instanceof JavaScriptErrorInEvaluate;
 }
+
+export function sparseArrayToString(entries: { index: number, value: any }[]): string {
```
dgozman, line 332: Why don't we accept `{ name, value }[]` array? Then cr/wk wouldn't need to additionally convert, since they share the same format.

PR #20383, opened 2023-01-26
chore: remove addFatalError

pavelfeldman:


packages/playwright-test/src/poolBuilder.ts
```diff
@@ -15,25 +15,29 @@
...
 
-  static buildForLoader(suite: Suite) {
-    new PoolBuilder('loader').buildPools(suite);
+  static buildForLoader(suite: Suite, loadErrors: LoadError[]) {
+    new PoolBuilder('loader', loadErrors).buildPools(suite);
   }
 
   static createForWorker(project: FullProjectInternal) {
-    return new PoolBuilder('worker', project);
+    return new PoolBuilder('worker', undefined, project);
```
yury-s, line 35: who is supposed to handle the errors in the worker? are they propagated to the runner with a global catch or something?

pavelfeldman: Errors in the worker throw in place, they are not collected.

dgozman (approved):

PR #20396, opened 2023-01-26
chore: make connectOverCDP work with localhost

### LINKED ISSUE #20364: [BUG] use happy eyeballs for connectOverCDP

mxschmitt:
This wraps happy eyeballs in two places, the place where we make the JSON request to Chromium and the actual CDP WebSocket request.

It required changes inside our happy eyeballs implementation since the [websocket library does not set](https://github.com/websockets/ws/blob/master/lib/websocket.js#L714) the `clientRequestOptions.hostname` field, it just sets the `host` field where we then fall back to when its not set.

This test would pass before Node.js 18 and fail after Node.js 18 without my changes.

Fixes https://github.com/microsoft/playwright/issues/20364

### DIFF
--- a/packages/playwright-core/src/server/happy-eyeballs.ts
+++ b/packages/playwright-core/src/server/happy-eyeballs.ts
@@ -31,7 +31,7 @@ const connectionAttemptDelayMs = 300;
 class HttpHappyEyeballsAgent extends http.Agent {
   createConnection(options: http.ClientRequestArgs, oncreate?: (err: Error | null, socket?: net.Socket) => void): net.Socket | undefined {
     // There is no ambiguity in case of IP address.
-    if (net.isIP(options.hostname!))
+    if (net.isIP(clientRequestArgsToHostName(options)))
       return net.createConnection(options as net.NetConnectOpts);
     createConnectionAsync(options, oncreate).catch(err => oncreate?.(err));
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/server/happy-eyeballs.ts
```diff
@@ -126,3 +127,11 @@ async function lookupAddresses(hostname: string): Promise<dns.LookupAddress[]> {
   return result;
 }
 
+function clientRequestArgsToHostName(options: http.ClientRequestArgs): string {
+  if (options.hostname)
+    return options.hostname;
+  if (options.host)
```
yury-s, line 133: There is some dubious logic around hostname / host etc inside http client option processing. Let's double check that it doesn't break with urls that have ip as the host.

yury-s (approved):

packages/playwright-core/src/server/happy-eyeballs.ts
mxschmitt: Looks similar to how we have it:

https://github.com/nodejs/node/blob/f69e84c75f8fbf48e5019af6d8b59e949c5cd653/lib/_http_client.js#L191-L193

yury-s (approved):

PR #20412, opened 2023-01-26
chore: truncate long protocol logs

aslushnikov:
This affects the logs in the `DEBUG=pw:protocol` mode so that they
never span more then 10 lines of 80-character-width terminal.


aslushnikov:
As per discussion: do env to limit max log line

dgozman (approved):

PR #20435, opened 2023-01-27
fix(windows): enable socks5 hostname resolution on WebKit

### LINKED ISSUE #20451: [BUG] WebKit Windows does not resolve DNS through the socks server
WebKit as of today on Windows not resolve DNS queries from the socks server side. It gets resolved locally instead.

This is different to WebKit linux/mac or Firefox and Chromium.

https://bugs.webkit.org/show_bug.cgi?id=251285

Relates https://github.com/microsoft/playwright/pull/20435

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/20451

### DIFF
--- a/packages/playwright-core/src/server/webkit/webkit.ts
+++ b/packages/playwright-core/src/server/webkit/webkit.ts
@@ -75,7 +75,9 @@ export class WebKit extends BrowserType {
         if (proxy.bypass)
           webkitArguments.push(...proxy.bypass.split(',').map(t => `--ignore-host=${t}`));
       } else if (process.platform === 'win32') {
-        webkitArguments.push(`--curl-proxy=${proxy.server}`);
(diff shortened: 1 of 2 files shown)


tests/library/browsercontext-proxy.spec.ts
```diff
@@ -367,19 +367,25 @@ it('should exclude patterns', async ({ contextFactory, server, browserName, head
   await context.close();
 });
 
-it('should use socks proxy', async ({ contextFactory, socksPort }) => {
+it('should use socks proxy', async ({ contextFactory, socksPort, platform, browserName }) => {
+  // libcurl by default resolves hostnames locally, which is not what we want.
+  const socks5Protocol = (platform === 'win32' && browserName === 'webkit') ? 'socks5h' : 'socks';
```
yury-s, line 372: let's move this logic inside wkBrowser.ts so so that the clients are fixed too?

yury-s (approved):

PR #20492, opened 2023-01-30
docs(test-components-js.md): Fix mount parameter for Vue example

sergemazille:
The example for mounting a Vue js component seems to be a duplicate of the React example.
For a Vue component we don't need the brackets syntax. (See also the difference between the two in [another part of the documentation](https://playwright.dev/docs/test-components#q-whats-the-difference-between-playwrighttest-and-playwrightexperimental-ct-reactsveltevuesolid))

### DIFF
--- a/docs/src/test-components-js.md
+++ b/docs/src/test-components-js.md
@@ -135,7 +135,7 @@ import App from './App.vue';
 test.use({ viewport: { width: 500, height: 500 } });
 
 test('should work', async ({ mount }) => {
-  const component = await mount(<App />);
+  const component = await mount(App);
   await expect(component).toContainText('Vite + Vue');
 });
 ```


mahtabnejad90 (approved):

sand4rt:
@sergemazille this valid code. Vue and Playwright CT both support JSX

sergemazille:
@sand4rt
> this valid code. Vue and Playwright CT both support JSX

On a bare Vue project `await mount(<App />)` throws an error while `await mount(App)` does not.
Might be something related to the project config, even though it's pretty much out of the box. I don't really know how to provide a reproducible environment...

sand4rt:
> @sand4rt
> 
> > this valid code. Vue and Playwright CT both support JSX
> 
> 
> 
> On a bare Vue project `await mount(<App />)` throws an error while `await mount(App)` does not.
> 
> Might be something related to the project config, even though it's pretty much out of the box. I don't really know how to provide a reproducible environment...
> 
> 

Did you change the filename to `.tsx`?

sergemazille:
> Did you change the filename to .tsx?

I've renamed the test file `welcome.cpt.spec.tsx` and the error is gone. :+1:

I think my initial Pull Request comes from the fact that this format is not common on Vue projects, it was a bit confusing to me.

Thanks for your help @sand4rt and sorry for the wasted time.

sand4rt:
Nice! Yeah tsx is not very common in Vue. Its mostly used in React/Solid and some other frameworks. No worries!

PR #20525, opened 2023-01-31
devops: re-add Chromium symbols build

mxschmitt:
Note: This the workflow how it previously was, just instead of sending the ref, its sending the CR revision.

### DIFF
--- a/.github/workflows/trigger_build_chromium_with_symbols.yml
+++ b/.github/workflows/trigger_build_chromium_with_symbols.yml
@@ -10,11 +10,21 @@ jobs:
     name: "trigger"
     runs-on: ubuntu-22.04
     steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '16'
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger_build_chromium_with_symbols.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '16'
+    - name: Get Chromium revision
+      id: chromium-version
+      run: |
+        REVISION=$(node -e "console.log(require('./packages/playwright-core/browsers.json').browsers.find(b => b.name === 'chromium').revision)")
```
aslushnikov, line 20:
```suggestion
        REVISION=$(node -e "console.log(require('./packages/playwright-core/browsers.json').browsers.find(b => b.name === 'chromium-with-symbols').revision)")
```

aslushnikov (approved):

PR #20557, opened 2023-02-01
chore: load tests file by file

pavelfeldman:


packages/playwright-test/src/common/testLoader.ts
```diff
@@ -29,64 +27,44 @@ export const defaultTimeout = 30000;
...
+  suite.location = { file, line: 0, column: 0 };
 
-  constructor(rootDir: string) {
-    this._rootDir = rootDir;
+  setCurrentlyLoadingFileSuite(suite);
+  try {
+    await requireOrImport(file);
+    cachedFileSuites.set(file, suite);
+  } catch (e) {
+    if (environment === 'worker')
```
dgozman, line 42: nit: instead of passing `environment` we can make `loadErrors` optional, or provide an error sink.

dgozman (approved):

PR #20645, opened 2023-02-05
chore: collect test dependencies

pavelfeldman:


packages/playwright-test/src/common/transform.ts
```diff
@@ -238,20 +256,6 @@ export function wrapFunctionWithLocation<A extends any[], R>(func: (location: Lo
   };
 }
 
-// This will catch the playwright-test package as well
-const kPlaywrightInternalPrefix = path.resolve(__dirname, '../../../playwright');
```
pavelfeldman, line 242: I'm not sure how this was supposed to work in production.

dgozman: I think this is for dev mode only, because usually any playwright* package is under `node_modules`.

packages/playwright-test/src/test.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
dgozman, line 1: Is this file for internal testing? Let's be more specific with the name, like `utilitiesForTest`, because `test` is too overloaded 😄

dgozman (approved):

packages/playwright-test/src/common/transform.ts
pavelfeldman: Ah, makes sense.

packages/playwright-test/src/test.ts
pavelfeldman: True

PR #20647, opened 2023-02-06
chore: watch mode first cut

pavelfeldman:


examples/todomvc/tests/integration.spec.ts
```diff
@@ -18,6 +18,7 @@ const TODO_ITEMS = [
 
 test.describe('New Todo', () => {
   test('should allow me to add todo items', async ({ page }) => {
+    expect(1).toBe(2);
```
mxschmitt, line 21: revert this

examples/todomvc/tests/integration.spec.ts
```diff
@@ -44,6 +45,7 @@ test.describe('New Todo', () => {
   });
 
   test('should clear text input field when an item is added', async ({ page }) => {
+    expect(1).toBe(2);
```
mxschmitt, line 48: ditto

packages/playwright-test/package.json
```diff
@@ -33,6 +33,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@types/node": "*",
+    "chokidar": "3.5.3",
```
mxschmitt, line 36: I think instead we want to say: ^3.5 to not break the ecosystem when other tools like vite also already depend on it.

packages/playwright-test/src/cli.ts
```diff
@@ -175,7 +169,9 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
   config._internal.passWithNoTests = !!opts.passWithNoTests;
 
   const runner = new Runner(config);
-  const status = await runner.runAllTests();
+  if (opts.watch)
+    process.stdout.write('\x1Bc');
```
mxschmitt, line 173:
```suggestion
    process.stdout.write('\x1Bc'); // Reset the terminal
```

Not sure if its worth mentioning with the other commands.

packages/playwright-test/package.json
pavelfeldman: I am actually thinking the opposite, I'd put it into bundle if I could.

dgozman (approved):

PR #20684, opened 2023-02-06
fix(chromium): do not pre-populate non-preflight OPTIONS requests

### LINKED ISSUE #20469: [BUG] Failing to intercept API requests
**Context:**
- Playwright Version: 1.29.2
- Operating System: Windows
- Node.js version: v17.0.1
- Browser: Chromium

## System:
 - OS: Windows 10 10.0.19044
 - Memory: 16.19 GB / 31.85 GB
## Binaries:
 - Node: 17.0.1 - C:\Program Files\nodejs\node.EXE
 - npm: 8.5.0 - C:\Program Files\nodejs\npm.CMD   
## Languages:
 - Bash: 5.0.16 - C:\WINDOWS\system32\bash.EXE (but really Powershell Core)

**Code Snippet**

```javascript
const { chromium } = require('@playwright/test');
const readline = require('node:readline/promises');
const { stdin : input, stdout : output } = require('node:process');

(async () => {
  const browser = await chromium.launch({ headless: false });
  const context = await browser.newContext({ serviceWorkers: 'block' });

   await context.route(url => true, (route, request) => {
     console.log(`Logging handler for ${request.method()} ${request.url()}`);
     route.continue();
   });

  // context.on('request', request => { console.log(`${request.method()} ${request.url()}`) });

    await page.goto('https://login.master.dev.twinfield.com/'); // Sorry, it's not openly accessible.
  
  const rl = readline.createInterface({ input, output });
  await rl.question('Please press ENTER to close Playwright browser ...');
  rl.close();

  console.log('Closing browser ...');

  await context.close();
  await browser.close();
})()
```

**Describe the bug**

Hi, I'm trying to mock all API responses to see if we can use it to do (pure) frontend testing.
First attempts with routeFromHAR I ran into some trouble handling API requests (didn't seem to get handled).

So to troubleshoot, I created a logging route handler, which only logs to the console.
Strangely, all requests, except the API requests are logged.
Also API requests in the frontend don't receive a proper response.
After logging in, the OPTIONS request on /api/ returns a 204 No content instead of a 200 with Hypermedia links to all other accessible APIs.

I thought it might have to do with service workers, so I blocked them, but it makes no difference.

If I disable the context.route block and enable the context.on('request', ...) it does log everything and API responses from the server are received properly by the frontend.

Any idea what could be the cause? Did I make a mistake in my code? Could there be something in the frontend application interfering with API request interception from Playwright?
If access to the application is needed, I might be able to arrange it (access to a public demo environment) next week.

dgozman:
Fixes #20469.


tests/page/page-route.spec.ts
```diff
@@ -688,6 +688,61 @@ it('should respect cors overrides', async ({ page, server, browserName, isAndroi
   }
 });
 
+it('should not auto-intercept non-preflight OPTIONS', async ({ page, server, browserName }) => {
+  await page.goto(server.EMPTY_PAGE);
```
yury-s, line 692: nit: add issue number annotation?

yury-s (approved):

mxschmitt:
Shall we also make sure that `page.on('request')` is able to read the response body or not related / important?

Simpso1 (approved):

PR #20691, opened 2023-02-06
feat: introduce Dockerfile.remote image

aslushnikov:
When this image is launched, it exposes a single endpoint
that can be used to connect to and to launch browsers.

### DIFF
--- a/packages/playwright-core/src/cli/cli.ts
+++ b/packages/playwright-core/src/cli/cli.ts
@@ -38,6 +38,7 @@ import type { GridFactory } from '../grid/gridServer';
 import { GridServer } from '../grid/gridServer';
 import type { Executable } from '../server';
 import { registry, writeDockerVersion } from '../server';
+import { addContainerCLI } from '../containers/';
 
 const packageJSON = require('../../package.json');
 
@@ -311,6 +312,8 @@ Examples:
 
   $ show-trace https://example.com/trace.zip`);
 
+addContainerCLI(program);
+
 if (!process.env.PW_LANG_NAME) {
   let playwrightTestPackagePath = null;
   const resolvePwTestPaths = [__dirname, process.cwd()];
@@ -371,6 +374,7 @@ type Options = {
   channel?: string;
   colorScheme?: string;
   device?: string;
+  disableRecorder?: boolean;
   geolocation?: string;
   ignoreHttpsErrors?: boolean;
   lang?: string;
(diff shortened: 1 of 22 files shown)


packages/playwright-core/src/cli/driver.ts
```diff
@@ -76,7 +76,7 @@ export async function launchBrowserServer(browserName: string, configFile?: stri
   console.log(server.wsEndpoint());
 }
 
-function selfDestruct() {
+export function selfDestruct() {
```
dgozman, line 79: I'd just copy this to containers instead of reusing.

packages/playwright-core/src/cli/cli.ts
```diff
@@ -701,6 +707,7 @@ function commandWithOpenOptions(command: string, description: string, options: a
       .option('--geolocation <coordinates>', 'specify geolocation coordinates, for example "37.819722,-122.478611"')
       .option('--ignore-https-errors', 'ignore https errors')
       .option('--load-storage <filename>', 'load context storage state from the file, previously saved with --save-storage')
+      .option('--disable-recorder', 'do not open Playwright Inspector')
```
dgozman, line 710: Since this is a private option, let's use env instead?

tests/config/platformFixtures.ts
```diff
@@ -24,7 +24,7 @@ export type PlatformWorkerFixtures = {
 };
 
 export const platformTest = test.extend<{}, PlatformWorkerFixtures>({
-  platform: [process.platform as 'win32' | 'darwin' | 'linux', { scope: 'worker' }],
+  platform: [process.env.DOCKER_REMOTE ? 'linux' : process.platform as 'win32' | 'darwin' | 'linux', { scope: 'worker' }],
```
dgozman, line 27:
```suggestion
  platform: [process.env.PWTEST_DOCKER_REMOTE ? 'linux' : process.platform as 'win32' | 'darwin' | 'linux', { scope: 'worker' }],
```

packages/playwright-core/src/containers/index.ts
```diff
@@ -0,0 +1,96 @@
...
+const debugLog = debug('pw:proxy');
+
+export function addContainerCLI(program: Command) {
+  const ctrCommand = program.command('container', { hidden: true })
+      .description(`Manage container integration (EXPERIMENTAL)`);
+
+  ctrCommand.command('install-services', { hidden: true })
+      .description('install services required to run container agent')
+      .action(async function() {
+        const { code } = await spawnAsync('bash', [path.join(__dirname, 'container_install_deps.sh')], { stdio: 'inherit' });
```
dgozman, line 38: `shell: true` ?

dgozman (approved):

packages/playwright-core/src/containers/index.ts
aslushnikov: this already runs bash, so `shell: true` should not be necessary

PR #20700, opened 2023-02-07
feat: sharded html report

yury-s:
This implementation is based on the [original PR](https://github.com/microsoft/playwright/pull/19691)  by @kevin940726. It makes the reporter produce single file when there is no sharding and multiple out-of-line report-x-of-y.zip reports which are automatically merged together when put in one folder.

References https://github.com/microsoft/playwright/issues/10437

Co-authored-by: Kai Hao <kevin830726@gmail.com>

### DIFF
--- a/packages/html-reporter/src/index.tsx
+++ b/packages/html-reporter/src/index.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import type { HTMLReport } from './types';
+import type { HTMLReport, Stats } from './types';
 import type zip from '@zip.js/zip.js';
 // @ts-ignore
 import * as zipImport from '@zip.js/zip.js/lib/zip-no-worker-inflate.js';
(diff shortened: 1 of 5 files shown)


packages/html-reporter/src/index.tsx
```diff
@@ -63,3 +81,35 @@ class ZipReport implements LoadedReport {
     return JSON.parse(await writer.getData());
   }
 }
+
+function mergeReports(reports: HTMLReport[]): HTMLReport {
```
dgozman, line 85: nit: I'd move this somewhere to utils, maybe make it a static function of `LoadedReport` or something.

dgozman (approved):

packages/html-reporter/src/index.tsx
yury-s: Moved to mergeReports.ts

PR #20755, opened 2023-02-08
chore(docs): sort members instead of validating

dgozman:


packages/playwright-core/types/types.d.ts
```diff
@@ -4620,7 +4591,36 @@ export interface Page {
...
+  /**
+   * API testing helper associated with this page. This method returns the same instance as
+   * [browserContext.request](https://playwright.dev/docs/api/class-browsercontext#browser-context-request) on the
+   * page's context. See
+   * [browserContext.request](https://playwright.dev/docs/api/class-browsercontext#browser-context-request) for more
+   * details.
+   */
+  request: APIRequestContext;
+
+  touchscreen: Touchscreen;}
```
mxschmitt, line 4623: nit: this bracket is weird.

packages/playwright-core/types/types.d.ts
```diff
@@ -4620,7 +4591,36 @@ export interface Page {
...
+  accessibility: Accessibility;
+
+  /**
+   * **NOTE** Only available for Chromium atm.
+   *
+   * Browser-specific Coverage implementation. See [Coverage] for more details.
+   */
+  coverage: Coverage;
+
+  keyboard: Keyboard;
```
mxschmitt, line 4610: nit: there are no comments for them, but they were also not there before.

mxschmitt (approved):

PR #20758, opened 2023-02-08
chore: do not run all on watch

pavelfeldman:


packages/playwright-test/src/runner/taskRunner.ts
```diff
@@ -45,66 +45,70 @@ export class TaskRunner<Context> {
   }
 
   async run(context: Context, deadline: number): Promise<FullResult['status']> {
+    const { status, cleanup } = await this.runDeferCleanup(context, deadline);
+    await cleanup();
```
dgozman, line 49: It seems that we will lose the `status` from cleanup here? In the situation where setup+tests succeeded, but teardown failed/interrupted for whatever reason.

packages/playwright-test/src/runner/watchMode.ts
```diff
@@ -63,24 +63,41 @@ class FSWatcher {
...
+  const context: TaskRunnerState = {
+    config,
+    reporter,
+    phases: [],
+  };
+  const taskRunner = createTaskRunnerForWatchSetup(config, reporter);
+  reporter.onConfigure(config);
+  const { status, cleanup: globalCleanup } = await taskRunner.runDeferCleanup(context, 0);
+  if (status !== 'passed') {
+    await globalCleanup();
```
dgozman, line 78: Same thing here?

packages/playwright-test/src/runner/watchMode.ts
```diff
@@ -167,11 +184,16 @@ Waiting for file changes. Press ${colors.bold('h')} for help or ${colors.bold('q
...
 
-    if (command === 'interrupted')
-      return 'interrupted';
+    if (command === 'interrupted') {
+      result = 'interrupted';
+      break;
+    }
   }
+
+  await globalCleanup();
```
dgozman, line 195: - Maybe same `status` issue here?
- Perhaps we should create another reporter to report teardown's output/errors?

dgozman (approved):

PR #20759, opened 2023-02-08
chore: set filters and run tests separately

pavelfeldman:


packages/playwright-test/src/runner/watchMode.ts
```diff
@@ -114,27 +109,36 @@ Waiting for file changes. Press ${colors.bold('a')} to run all, ${colors.bold('q
       continue;
     }
 
-    if (command === 'all') {
+    if (command === 'run') {
       // All means reset filters.
```
dgozman, line 113: Outdated comment.

dgozman (approved):

packages/playwright-test/src/runner/watchMode.ts
pavelfeldman: Will follow up.

PR #20785, opened 2023-02-09
chore: remove unnecessary extra options from docs

dgozman:
Python-only methods should not reference js-only options.


mxschmitt (commented):
good finding.

mxschmitt (approved):

PR #20790, opened 2023-02-09
chore: remove experimental dockerfile.remote 

aslushnikov:
We didn't find a compelling-enough use case to release this.


dgozman (approved):

pastelsky:
@aslushnikov having a docker command / image around was quite helpful in powering a local devloop if you wanted to connect to browsers in docker in order to maintain a consistent test environment. Moreover, it'd been nice to use remote docker as a potential playwright-as-a-service backend in the future for other users of playwright.

Was this removed due to lack of interest from the community?

aslushnikov:
@pastelsky yes, precisely. We decided to hold off releasing this and revisit the decision in ~6 months.

pastelsky:
That's a bit of a bummer since this was quite useful for us — even though I acknowledge this was never officially released or documented. Is there a recommended workflow for running local browser-as-a-service dockers now? Docker seems like a must have for us for capturing visual regressions.

aslushnikov:
@pastelsky have you been using this already at Atlassian? If you file a new issue, describing your usecase and how this would be helpful to you, then we would be able to see some real clients & real use cases filed. Part of the reason this got removed is the lack of this kind of feedback.

> Is there a recommended workflow for running local browser-as-a-service dockers now? 

Looks like you can bake your own solution similarly to what was done here, but relying on publicly available [`BrowserType.launchServer`](https://playwright.dev/docs/next/api/class-browsertype#browser-type-launch-server).

PR #20792, opened 2023-02-09
chore: inline docs from python.md

dgozman:
These are currently messed up, with non-python options all over the place.


docs/src/api/class-selectors.md
```diff
@@ -202,13 +202,27 @@ Script that evaluates to a selector engine instance. The script is evaluated in
 
 Script that evaluates to a selector engine instance. The script is evaluated in the page context.
 
+### option: Selectors.register.script
```
mxschmitt, line 205:
```suggestion
### option: Selectors.register.path
```

I feel like this was broken before and should be like this.

mxschmitt (approved):

docs/src/api/class-selectors.md
dgozman: Updated.

PR #20802, opened 2023-02-10
feat (reporter-junit): Allow users to customize the class name for results in JUNIT reporter

### LINKED ISSUE #20380: [Feature] Junit ClassName Template support
We want a way to configure a template function for the junit classname.

https://github.com/microsoft/playwright/blob/147bb6b292cb5d4051688eb0535eccd89a372ced/packages/playwright-test/src/reporters/junit.ts
Currently its set via -> 
```ts

  private _addTestCase(test: TestCase, entries: XMLEntry[]) {
    const entry = {
      name: 'testcase',
      attributes: {
        classname: formatTestTitle(this.config, test, undefined, true),
        time: (test.results.reduce((acc, value) => acc + value.duration, 0)) / 1000
      },
      children: [] as XMLEntry[]
    };
    entries.push(entry);
```

Which leads to a className of -> ` classname="messaging\verify_giphy.playwright.ui.test.ts › able to see giphy › should user able to see giphy and add in chat"`

This includes the file path and the testname.

Unfortunately, this doesn't work nicely with Azure Devops as it expects a filePath -> 

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-test-results-v2?view=azure-pipelines&tabs=junit%2Ctrxattachments%2Cyaml

image

## Proposal:

Can we have a `classNameTemplate` configuration option where we have access to the current data of the test that allows us to pick what we want in the classname? Like jest-junit does -> https://github.com/jest-community/jest-junit#configuration

```
  reporter: [
    [
      "junit",
      {
        outputFile: "logs/test-results/e2e-ui/results.xml",
        classNameTemplate: vars => vars.filepath.replaceAll(".js", ".ts"),
      },
    ],
  ],
```

caaldrid:
feat (reporter-junit): Allow users to customize the class name for results in JUNIT reporter

Introduce a `classNameTemplate` property into the junit configuration who's value is a function that takes in a TestClass object and returns a string. By using properties of the TestClass Object Users can set the reported class name in the test junit results to whatever they want. Updated doc with information and an example of this.

Fixes #20380


mastrzyz:
Awesome!
Yeah we were seeing this in Azure Devops previously and it was causing confusion in our scripts on the test file containing the testcase name with it.

```
 📁File           │ packages\foo\bar\baz\edge-tests\engsys\webpack:\playwright_smoke.playwright.ui.test.ts › Native Smoke Test › should user able to see giphy and add in chat                                                                               ║
```

packages/playwright-test/types/test.d.ts
```diff
@@ -16,14 +16,18 @@
  */
 
 import type { APIRequestContext, Browser, BrowserContext, BrowserContextOptions, Page, LaunchOptions, ViewportSize, Geolocation, HTTPCredentials, Locator, APIResponse, PageScreenshotOptions } from 'playwright-core';
+import type { TestCase } from '@playwright/test/reporter';
+
 export * from 'playwright-core';
 
+export type junitClassNameTemplate = (test: TestCase) => string;
```
yury-s, line 23: `TestCase` parameter is not sufficient to get e.g. a relative path of the test file, without it you won't be able to implement default classname format in the user code. Having the config as another parameter would help with that.

packages/playwright-test/types/test.d.ts
```diff
@@ -16,14 +16,18 @@
...
+
 export type ReporterDescription =
   ['dot'] |
   ['line'] |
   ['list'] |
   ['github'] |
-  ['junit'] | ['junit', { outputFile?: string, stripANSIControlSequences?: boolean }] |
+  ['junit'] | ['junit', { outputFile?: string, stripANSIControlSequences?: boolean, classNameTemplate?: junitClassNameTemplate }] |
```
yury-s, line 30: I'm not sure we want to allow functions in the configuration, we don't do it in other places today, maybe we can have a declarative config similar to [snapshotPathTemplate](https://playwright.dev/docs/api/class-testproject#test-project-snapshot-path-template). Let me discuss this in the team meeting tomorrow.

yury-s:
Closing this PR as it was superseded by https://github.com/microsoft/playwright/pull/20927

PR #20830, opened 2023-02-10
fix(reporters): preserve the order between deferred errors and stdio

dgozman:


packages/playwright-test/src/reporters/multiplexer.ts
```diff
@@ -93,7 +88,7 @@ export class Multiplexer implements Reporter {
   async onEnd() { }
 
   async onExit(result: FullResult) {
-    if (this._deferredErrors) {
+    if (this._deferred) {
       // onBegin was not reported, emit it.
       this.onBegin(this._config, new Suite('', 'root'));
```
pavelfeldman, line 93: another wrap?

pavelfeldman (approved):

PR #20839, opened 2023-02-11
browser(firefox): Restrain sending http credentials on a specific origin

### LINKED ISSUE #20464: [Feature] Restrain sending http credentials on a specific hostname

SebastienRichert:
Fixes #20464

For security purpose, we would like to restrain sending HTTP credentials to only the specified server. The idea is to give the ability to specify a hostname additionally to current pair username/password. When an authorization response is received from servers, the credentials are sent only if the server hostname in the request matches case insensitive the specified hostname.

This is a separate PR following #20374 in order to update juggler files separatly.

### DIFF
--- a/browser_patches/firefox/juggler/NetworkObserver.js
+++ b/browser_patches/firefox/juggler/NetworkObserver.js
@@ -300,6 +300,9 @@ class NetworkRequest {
     }
     if (!credentials)
       return false;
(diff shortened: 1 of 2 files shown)


aslushnikov (approved):
Awesome, thank you for contribution. I pulled this downstream so that we can produce new firefox builds; should be part of Firefox 1379

aslushnikov:
(closing this PR now since it's migrated to our internal build repo)

PR #20883, opened 2023-02-14
feat: resolve storageState relative to the config

dgozman:


dgozman:
Closing since this is not better than current behavior.

PR #20892, opened 2023-02-14
fix(doc): electron evaluate evaluates in electron not worker

JoelEinbinder:


JoelEinbinder:
Types say the first argument here is actually the electron application? Should that also be documented here? I'm a bit scared of how I changed one line and a dozen method docs changed.

dgozman (approved):

dgozman:
> Types say the first argument here is actually the electron application? Should that also be documented here?

Yep, first argument is the `app` inside the electron main context. It makes sense to document it.

> I'm a bit scared of how I changed one line and a dozen method docs changed.

Infrastructure 😄

PR #20908, opened 2023-02-15
fix(docs): command to install chrome

laqieer:
image


aslushnikov:
"msedge" or "chrome" is just an example value; either would work in the docs, so I don't think changing one to another is beneficial.

mxschmitt:
I followed-up in https://github.com/microsoft/playwright/pull/21027 to fix it. Thanks!

PR #20932, opened 2023-02-16
feat: key value store backed by filesystem

yury-s:
Keys are used as a relative file path without any sanitization assuming that the underlying fs will throw on error.

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -749,6 +749,34 @@ export default defineConfig({
 });
 ```
 
+## property: TestConfig.storeDir
+* since: v1.32
+- type: ?<[string]>
+
+Directory where the values accessible via [TestStore] are persisted. Each value is saved in its own file. The key of the value is its path relative to the `storeDir`. Defaults to `./playwright`.
+
+**Usage**
+
+```js tab=js-js
+// playwright.config.js
+// @ts-check
+
+const { defineConfig } = require('@playwright/test');
+
+module.exports = defineConfig({
+  storeDir: './playwright-store',
+});
+```
+
+```js tab=js-ts
+// playwright.config.ts
+import { defineConfig } from '@playwright/test';
+
+export default defineConfig({
+  storeDir: './playwright-store',
+});
+```
+
 ## property: TestConfig.testDir
 * since: v1.10
 - type: ?<[string]>
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -753,7 +753,7 @@ export default defineConfig({
 * since: v1.32
 - type: ?<[string]>
 
-Directory where the values accessible via [TestStore] are persisted. Each value is saved in its own file. The key of the value is its path relative to the `storeDir`. Defaults to `./playwright`.
+Directory where the values accessible via [TestStore] are persisted. All pahts in [TestStore] are relative to `storeDir`. Defaults to `./playwright`.
 
 **Usage**
 
--- a/docs/src/test-api/class-teststore.md
+++ b/docs/src/test-api/class-teststore.md
@@ -17,6 +17,17 @@ test('get user name', async ({ page, context }) => {
 });
 ```
 
+## async method: TestStore.delete
+* since: v1.32
+
+Delete named item from the store. Does nothing if the path is not in the store.
+
+### param: TestStore.delete.path
+* since: v1.32
+- `path` <[string]>
+
+Item path.
+
 ## async method: TestStore.get
 * since: v1.32
 - returns: <[any]>
@@ -40,6 +51,12 @@ Returns absolute path of the corresponding store entry on the file system.
 - `path` <[string]>
 
 Path of the item in the store.
+
+## method: TestStore.root
+* since: v1.32
+- returns: <[string]>
+
+Returns absolute path of the store root directory.
 
 ## async method: TestStore.set
 * since: v1.32
--- a/docs/src/test-api/class-teststore.md
+++ b/docs/src/test-api/class-teststore.md
@@ -2,7 +2,7 @@
 * since: v1.32
 * langs: js
 
-Playwright Test provides a global `store` object that can be used read/write values to the filesystem. Each value is stored in its own file with the key used as a path relative to `TestConfig.storeDir` (it is `./playwright` by default).
+Playwright Test provides a global `store` object that can be used to read/write values on the filesystem. Each value is stored in its own file inside './playwright' directory, configurable with [`property: TestConfig.storeDir`].
 
 ```ts
 import { test, store } from '@playwright/test';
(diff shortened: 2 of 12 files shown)


docs/src/test-api/class-testconfig.md
```diff
@@ -749,6 +749,34 @@ export default defineConfig({
...
+**Usage**
+
+```js tab=js-js
+// playwright.config.js
+// @ts-check
+
+const { defineConfig } = require('@playwright/test');
+
+module.exports = defineConfig({
+  storeDir: './playwright-store',
```
dgozman, line 767:
```suggestion
  storeDir: './playwright',
```

docs/src/test-api/class-testconfig.md
```diff
@@ -749,6 +749,34 @@ export default defineConfig({
 });
 ```
 
+## property: TestConfig.storeDir
+* since: v1.32
+- type: ?<[string]>
+
+Directory where the values accessible via [TestStore] are persisted. Each value is saved in its own file. The key of the value is its path relative to the `storeDir`. Defaults to `./playwright`.
```
dgozman, line 756:
```suggestion
Directory where the values accessible via [TestStore] are persisted. All pahts in [TestStore] are relative to `storeDir`. Defaults to `./playwright`.
```

docs/src/test-api/class-testconfig.md
```diff
@@ -749,6 +749,34 @@ export default defineConfig({
...
+  storeDir: './playwright-store',
+});
+```
+
+```js tab=js-ts
+// playwright.config.ts
+import { defineConfig } from '@playwright/test';
+
+export default defineConfig({
+  storeDir: './playwright-store',
```
dgozman, line 776:
```suggestion
  storeDir: './playwright',
```

docs/src/test-api/class-teststore.md
```diff
@@ -0,0 +1,59 @@
+# class: TestStore
+* since: v1.32
+* langs: js
+
+Playwright Test provides a global `store` object that can be used read/write values to the filesystem. Each value is stored in its own file with the key used as a path relative to `TestConfig.storeDir` (it is `./playwright` by default).
```
dgozman, line 5:
```suggestion
Playwright Test provides a global `store` object that can be used to read/write values on the filesystem. Each value is stored in its own file inside './playwright' directory, configurable with [`property: TestConfig.storeDir`].
```

docs/src/test-api/class-teststore.md
```diff
@@ -0,0 +1,59 @@
...
+  // Check that the name matches mock data.
+  await expect(page.getByLabel('Name')).toHaveText('John');
+});
+```
+
+## async method: TestStore.get
+* since: v1.32
+- returns: <[any]>
+
+Get named item from the store. Returns undefined if there is no value with given path.
```
dgozman, line 24: We should figure out:
- whether we support binary values like images;
- whether we treat json specially and JSON.parse it;
- whether we detect mime type from extension or pass it explicitly.

docs/src/test-api/class-teststore.md
```diff
@@ -0,0 +1,59 @@
...
+test('get user name', async ({ page, context }) => {
+  await page.goto('/');
+  // Return mock user info from the store.
+  await page.route('**/info/user', route => route.fulfill({ path: store.path('mocks/user.json')}))
+  await page.getByText('My Profile');
+  // Check that the name matches mock data.
+  await expect(page.getByLabel('Name')).toHaveText('John');
+});
+```
+
```
dgozman, line 19: Suggestion: `method: TestStore.root` to get the root?

docs/src/test-api/class-teststore.md
```diff
@@ -0,0 +1,59 @@
...
+* since: v1.32
+- `path` <[string]>
+
+Item path.
+
+### param: TestStore.set.value
+* since: v1.32
+- `value` <[any]>
+
+Item value. The value must be serializable to JSON. Passing `undefined` deletes the entry with given path.
```
dgozman, line 59: I would also do `delete` right away.

packages/playwright-test/src/runner/runner.ts
```diff
@@ -25,12 +25,14 @@ import type { TaskRunnerState } from './tasks';
...
 import { colors } from 'playwright-core/lib/utilsBundle';
 import { runWatchModeLoop } from './watchMode';
+import { setCurrentConfig } from '../common/globals';
 
 export class Runner {
   private _config: FullConfigInternal;
 
   constructor(config: FullConfigInternal) {
     this._config = config;
+    setCurrentConfig(this._config);
```
dgozman, line 35: We probably should move this to ConfigLoader.

packages/playwright-test/src/store.ts
```diff
@@ -0,0 +1,52 @@
...
+    const config = currentConfig();
+    if (!config)
+      throw new Error('Cannot access store before config is loaded');
+    return path.join(config._internal.storeDir, name);
+  }
+
+  async set<T>(name: string, value: T | undefined) {
+    const file = this.path(name);
+    if (value === undefined) {
+      await fs.promises.rm(file, { force: true });
```
dgozman, line 43: Should we support `name` that is actually a directory? I'd say that would be convenient.

tests/playwright-test/store.spec.ts
```diff
@@ -216,115 +179,100 @@ test('should load context storageState from store', async ({ runInlineTest, serv
...
     `,
   }, { workers: 1 });
   expect(result.exitCode).toBe(0);
   expect(result.passed).toBe(2);
+  const file = path.join(test.info().outputPath(), 'my/store/dir/foo/bar.json');
+  expect(JSON.parse(await fs.promises.readFile(file, 'utf-8'))).toEqual({ 'a': 2023 });
 });
 
-test('should throw on unknown storageStateName value', async ({ runInlineTest, server }) => {
+test('fulfill with return path of the entry', async ({ runInlineTest, page }) => {
```
dgozman, line 261: This test should be moved to `playwright.spec.ts` or similar one.

dgozman (approved):

docs/src/test-api/class-teststore.md
yury-s: > whether we support binary values like images;
We should, I'll address this separately.


> whether we treat json specially and JSON.parse it;
I'd say yes, as this is expected to be the most common use case.

> whether we detect mime type from extension or pass it explicitly.
If the keys include extension like in this implementation to we should infer the mime type from it I believe. In any case I'll tackle this in a separate PR.

docs/src/test-api/class-teststore.md
yury-s: Let's wait for now?

packages/playwright-test/src/store.ts
yury-s: Don't know, although convenient it maybe confusing.

packages/playwright-test/src/runner/runner.ts
yury-s: done

docs/src/test-api/class-teststore.md
yury-s: Added.

dgozman (approved):

PR #20971, opened 2023-02-17
docs: delete the URI /wd/hub Selenium Grid addresses.

evertones:
Selenium 4 does not use `/wd/hub` as part of the Hub URI.

As seen in the example https://www.selenium.dev/documentation/grid/getting_started/#metadata-in-tests, it registers the Selenium Grid URL simply as `http://gridUrl:4444` (no `/wd/hub`).

When trying to access the URL ending with `/wd/hub`, it returns the message 'Unable to find handler for (GET) \u002fwd\u002fhub'.

### DIFF
--- a/docs/src/selenium-grid.md
+++ b/docs/src/selenium-grid.md
@@ -17,7 +17,7 @@ If you run distributed Selenium Grid, Playwright needs selenium nodes to be regi
 
 ```bash
 # Start selenium node
-SE_NODE_GRID_URL="http://<selenium-hub-ip>:4444/wd/hub" java -jar selenium-server-<version>.jar node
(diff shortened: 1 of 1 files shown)


evertones:

dgozman:
@evertones Did you encounter issues when using `/wd/hub` url? Playwright does not GET this url, but rather POST to `/wd/hub/session` to create a new session. From my local testing, it works both with and without `/wd/hub`, quite surprisingly. So I wonder whether this is just a cosmetic change, or does it actually fix an issue on your end?

evertones:
@dgozman: This is just a cosmetic change to be consistent with the changes in Selenium 4.

dgozman (approved):

dgozman:

PR #20979, opened 2023-02-17
Make annotations searchable in HTML reporter

pastelsky:
Currently, annotations are only visible in the test details view.

This PR allows tests to be searched by annotations in the test listing view of the HTML reporter. 
Entering `type: description` allows for searching by annotation.

This is helpful because teams can tag tests (either manually or through common fixtures) with certain metadata, and get a shareable URL to a subset of test cases shown in the HTML report filtered by annotation.

The assumption here is that `type: description` is the key value pair for an annotation — which may need to expand to properties other than `description` in the future.

Ideally, we'd use tagging for this, but currently, tags appear in the test description which makes it noisy when using multiple tags at once.

### DIFF
--- a/docs/src/test-annotations-js.md
+++ b/docs/src/test-annotations-js.md
@@ -211,7 +211,11 @@ test('user profile', async ({ page }) => {
 
 ## Custom annotations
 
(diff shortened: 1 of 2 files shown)


pavelfeldman:
Please start your contribution with filing an issue so that we could have a chance to discuss it!

pastelsky:
Ah, to be clear I expected the implementation to change, and this was to start a conversation on tags/annotations — but I see why creating an issue might have been a more standard way to go about it.

PR #20991, opened 2023-02-17
Add explainer for CLI output of `DotReporter`

Sheraff:
After running tests, playwright outputs something like:
```sh
········································································×····±··
·×··×··××····F·±·······················
```

And it was never clear to me what each of these symbols meant, and I couldn't find the explanation anywhere until I looked at the source code itself. I propose having the labels directly in the terminal. For reference, Next.js and Gatsby both do this for their build output and it's very pleasing.


pavelfeldman:
Users refer this reporter because it is brief. It fits a hundreds of tests in one line, we can't be adding more lines to it.

PR #20998, opened 2023-02-17
fix: stricter condition for Firefox codegen tests workaround

aslushnikov:
This patch fixes the following tests on WebKit @ Darwin:
- library/inspector/cli-codegen-2.spec.ts:197:7 › cli codegen › should download files
- library/inspector/cli-codegen-2.spec.ts:428:7 › cli codegen › should update hover model on action


aslushnikov:
TBR: @dgozman

Issue #20955, opened 2023-02-16
[Feature] Show current step/fixture in list reporter


JoelEinbinder:
Some of my tests take a long time to run, and they have a few slow steps and slow fixtures which they instantiate. It would be nice to show the current fixture/step being worked on while the test is running so I can get a sense of how it is doing. I envision this going in gray where the elapsed time eventually appears.

I expect one fixture/step name to be shown. In the case of more than one active, I'd expect the most recently started one to be shown. Maybe add 250ms throttle to avoid too much noise in the terminal for quick tests.

I am happy to implement it if this is something that could land.

aslushnikov:
There's already a "printSteps" option in list reporter!  https://playwright.dev/docs/next/test-reporters#list-reporter


```ts// playwright.config.ts
import { defineConfig } from '@playwright/test';

export default defineConfig({
  reporter: [['list', { printSteps: true }]],
});
```

note for self: update doc

JoelEinbinder:
1. Docs wouldn't have helped me. I didn't know reporters even had options because `npm init playwright` just gave me a string for reporter.

2. Print steps doesn't seem to show fixtures, which is the main place my tests are slow. Consider a `loggedInPage` fixture which loads and logs into the page. I can add steps directly into my fixtures, but that gets ugly pretty quickly. Needs a begin and end step, and all the variables have to be outside the step so that they can be used across steps. Would much rather it be built in.

3. Why not turn printSteps on by default? Who wants it off?

aslushnikov:
Hey Joel!

As per the team meeting discussion:

> Why not turn printSteps on by default? Who wants it off?

We felt uncomfortable to toggle the option on by default.

> Docs wouldn't have helped me. I didn't know reporters even had options because npm init playwright just gave me a string for reporter.

This is still an issue with docs, probably. Debbie is re-working documentation, and it'll be part of her re-work.

With this, closing!

acidoxee:
Hi, as mentioned in #20534 unfortunately the `printSteps` option doesn't seem to work in GitHub Actions. Is there any trick to force it anyway? The lack of display of steps in CI is quite painful for long-lived tests as we don't know what's happening unless we enable debug mode with `DEBUG=pw:api`, which is way too verbose.

PR #18269, opened 2022-10-22
chore: render steps in list reporter

pavelfeldman:

PR #21001, opened 2023-02-17
feat(test): print steps by default in list reporter

JoelEinbinder:
Updates #18269 to be on by default. Let's be bold!

#20955

### DIFF
--- a/docs/src/test-reporters-js.md
+++ b/docs/src/test-reporters-js.md
@@ -137,7 +137,7 @@ Running 124 tests using 6 workers
 10    should respect enclosing .gitignore:
 ```
 
-You can opt into the step rendering via passing the following config option:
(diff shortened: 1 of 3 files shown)


pavelfeldman:
> Let's be bold!

List is our default reporter, I'm slightly worried that we should not be breaking something that works for the majority of users. Wdyt?

JoelEinbinder:
I'm just wondering when it switches if ever? It doesn't seem like I ever want this feature to be off as long as it works.

pavelfeldman:
> I'm just wondering when it switches if ever? It doesn't seem like I ever want this feature to be off as long as it works.

I think it starts creating row-per step on CI, doesn't it?

dgozman:
Closing as stale, since no consensus has been reached and branch has conflicts.

PR #21003, opened 2023-02-17
fix(snapshot): render srcdoc iframe snapshots

kalmi:
This PR fixes DOM snapshots for iframes that have an `srcdoc` attribute.

Currently `srcdoc` attributes are left as is. They should be stripped just like `src` attributes currently are.

Leaving it as is results in the `srcdoc` being used for the iframe when viewing traces. That prevents the proper iframe DOM snapshot from being present and visible.


kalmi:

dgozman (approved):

PR #21035, opened 2023-02-20
chore(ct): vue2 import h from vue

sand4rt:
It is supported since v2.7

### DIFF
--- a/packages/playwright-ct-vue2/registerSource.mjs
+++ b/packages/playwright-ct-vue2/registerSource.mjs
@@ -18,7 +18,7 @@
 
 // This file is injected into the registry as text, no dependencies are allowed.
 
-import Vue from 'vue';
+import Vue, { h } from 'vue';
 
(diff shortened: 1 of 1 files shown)


pavelfeldman:
I was doing that and it resulted in 2 copies of vue when used with vue-cli. If we have reasons to believe that is fixed, we can go ahead with your change. Not sure how many users are using vue-cli.

sand4rt:
Did you try this after Dec 1st, 2022? The `h()` was recently added. I tested it again manually and i see that it works as expected. Would also mean there would be a bug in Vue2 if this does not work.

Not much i think, 2,250 downloads a week _([by NPM](https://www.npmjs.com/package/@playwright/experimental-ct-vue2))_. Vue2 will reach end of life on December 31st, 2023 btw. See no reason why people want to start a new project in Vue2 and those who already have a Vue2 project do not migrate so quickly to Playwright component testing, i guess.

pavelfeldman (approved):

PR #21040, opened 2023-02-20
Prerequisites added #641

surfmuggle:
This helps to address [#641](https://github.com/microsoft/playwright.dev/issues/641)


debs-obrien:
Hi @surfmuggle, thanks for your contribution. We don't really want to recommend different ways of downloading node and as we are on the node version of the docs we do presume that people have node on their machines or know that node is required and will therefore find the node docs and install node depending on the system they are on and the most up to date version at the time. For now we won't add this to the getting started docs but really appreciate your suggestion.

PR #21062, opened 2023-02-21
fix: use the JUnit schema's specified ISO format for timestamp

lsklyut:
JUnit uses an ISO 8601 format for the timestamp attribute of the testsuite element. This updates the code to use that format, rather than milliseconds from epoch.

https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd#L181

### DIFF
--- a/packages/playwright-test/src/reporters/junit.ts
+++ b/packages/playwright-test/src/reporters/junit.ts
@@ -116,7 +116,7 @@ class JUnitReporter implements Reporter {
       name: 'testsuite',
       attributes: {
         name: suite.title,
-        timestamp: this.timestamp?.toISOString(),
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -116,7 +116,7 @@ class JUnitReporter implements Reporter {
       name: 'testsuite',
       attributes: {
         name: suite.title,
-        timestamp: this.timestamp,
+        timestamp: this.timestamp?.toISOString(),
```
mxschmitt, line 119:
```suggestion
        timestamp: this.timestamp.toISOString(),
```

PR #21106, opened 2023-02-22
fix(loader): experimentalLoader with node@18

nowells:
There is currently a bug when running `node@18.14.2` when running with experimentalLoader

```
TypeError: The URL must be of scheme file
  at new NodeError (node:internal/errors:399:5)
  at Object.fileURLToPath (node:internal/url:1492:11)
  at resolve (./node_modules/@playwright/test/lib/experimentalLoader.js:39:48)
  at nextResolve (node:internal/modules/esm/loader:163:22)
  at ESMLoader.resolve (node:internal/modules/esm/loader:838:24)
  at ESMLoader.getModuleJob (node:internal/modules/esm/loader:424:7)
  at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:79:21)
```

This came from my test `vrt.spec.ts` which had a non package import inside of it

```ts
import fs from "node:fs/promises"
```

The test run failed due to node imports not returning fileUrls when resolved.

### DIFF
--- a/tests/playwright-test/loader.spec.ts
+++ b/tests/playwright-test/loader.spec.ts
@@ -612,3 +612,29 @@ test('should import export assignment from ts', async ({ runInlineTest }) => {
   expect(result.passed).toBe(1);
   expect(result.exitCode).toBe(0);
 });
+
+test('should support node imports', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'package.json': JSON.stringify({
+      type: 'module'
+    }),
+    'test.json': 'test data',
+    'utils.mjs': `
+      import fs from "node:fs/promises";
+
+      export async function utilityModuleThatImportsNodeModule() {
+        return await fs.readFile('test.json', 'utf8');
+      }
+    `,
+    'a.test.ts': `
+      import { test, expect } from '@playwright/test';
+      import { utilityModuleThatImportsNodeModule } from './utils.mjs';
+
+      test('pass', async () => {
+        expect(await utilityModuleThatImportsNodeModule()).toBe('test data');
+      });
+    `
+  });
+  expect(result.passed).toBe(1);
+  expect(result.exitCode).toBe(0);
+});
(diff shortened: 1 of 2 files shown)


Smrtnyk:
I am running into the same issue after update to 1.31

nowells:
The current test failure appears to be unrelated to this change, but I am not familiar enough with the overall stability of the test suite to know if this is a common flakey test, or if this is indeed related. I am happy to dig in and debug and fix this failure, if someone more familiar can let me know if this looks related or a common flakey test.

klh:
can confirm this bug

nowells:

tests/playwright-test/loader.spec.ts
```diff
@@ -612,3 +612,29 @@ test('should import export assignment from ts', async ({ runInlineTest }) => {
   expect(result.passed).toBe(1);
   expect(result.exitCode).toBe(0);
 });
+
+test('should support node imports', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
```
dgozman, line 617: You'll probably need the following two lines: https://github.com/microsoft/playwright/blob/main/tests/playwright-test/loader.spec.ts#L254-L255.

dgozman (commented):
Thank you for the PR! I can reproduce the issue and this fixes it. As long as tests pass, I'll merge it in.

tests/playwright-test/loader.spec.ts
dgozman: Also, to engage esm loader, you need `playwright.config.ts: 'export default {}'`, otherwise ts files do not load as ES modules.

nowells: thanks for the feedback! I previously had `playwright.config.ts: 'export default {}'` but when I removed it the tests still failed when I undid my fix, but I'll add in for explicitness.

dgozman (approved):
Looks great, merging in.

Smrtnyk:
I now get a new error after update to 1.31
`SyntaxError: Unexpected token 'export'`
I am using es modules, and I am importing from esm package, and it worked on 1.30

PR #21127, opened 2023-02-23
test: fix installation tests on win32

aslushnikov:
Since we now use a special build for win32, the
`installedSoftwareOnDisk` reports a different installed browser.


aslushnikov:
We decided to not land this (as per discussion with @dgozman )

PR #21131, opened 2023-02-23
fix(snapshots): define dummy custom elements

### LINKED ISSUE #21030: [BUG] custom elements never `:defined` in trace viewer
**Context:**

## Playwright 
  - 1.30.0
## System:
 - OS: macOS 12.5.1
 - Memory: 5.29 GB / 32.00 GB
## Binaries:
 - Node: 19.0.1 - /opt/homebrew/bin/node
 - Yarn: 1.22.19 - /opt/homebrew/bin/yarn
 - npm: 8.19.2 - /opt/homebrew/bin/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash

**Code Snippet**

If you run the following and capture a trace:

```javascript

(async () => {
  
  // define a custom element
  await page.addScriptTag({
    content: `
    customElements.define("my-element", class extends HTMLElement {
      constructor() {
        super()
        const shadow = this.attachShadow({ mode: "open" })
        shadow.innerHTML = \`<p style="color: red">hello <slot></slot></p>\`
      }
    })
    `,
  })

  // render the custom element
  // and add style to hide the element until it is defined,
  // preventing a flash of unregistered component. this is a very common rule used with custom elements
  await page.setContent(`
    <style>
      :not(:defined) {
        visibility: hidden;
      }
    </style>
    <my-element>world</my-element>
  `)
})();
```

**Describe the bug**

Whilst the trace correctly captures the shadow root and all styles, it doens't go as far as registering a custom element. So when the above is viewed in the trace viewer, the component will be `hidden` since the custom element is not defined. 

This can be proven by running the following code in the context of the iframe:

```js
customElements.get("my-element") // undefined
```

One possible fix would be to register a stub/dummy custom element during the trace capture phase. Something like:

```js
cusotmElements.define("my-element", class extends HTMLElement)
```

A user-space workaround would be to inject a css rule in some global before hook:

```css
await page.addStyleTag({
  content: `:not(:defined) { visibility: visible; }`
})
```

Though given the prevalance of hiding elements until defined, it would be nicer if this just worked out of the box.

dgozman:
For all custom elements defined in the page, we preserve their names and define them in the rendered snapshot.
This makes things like `:defined` css pseudo work.

Fixes #21030.

### DIFF
--- a/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
+++ b/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
@@ -45,6 +45,7 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
   const kScrollLeftAttribute = '__playwright_scroll_left_';
   const kStyleSheetAttribute = '__playwright_style_sheet_';
   const kTargetAttribute = '__playwright_target__';
+  const kCustomElementsAttribute = '__playwright_custom_elements__';
 
   // Symbols for our own info on Nodes/StyleSheets.
   const kSnapshotFrameId = Symbol('__playwright_snapshot_frameid_');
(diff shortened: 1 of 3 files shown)


yury-s (approved):

WickyNilliams:
Thanks for the quick fix! I tried diving into the code to make a PR myself, but wasn't sure where best to make the fix. Appreciate you taking the time

PR #21165, opened 2023-02-23
feat(html): render annotations as links if needed

### LINKED ISSUE #20584: [Feature] Add parameter 'link' to Annotations
In most cases in my practice annotations using for linking bugs for Jira (as an example).
But now we have only `type` and `description`. 

```
    test('Test is red, bug in product!', async () => {
        test.info().annotations.push({ type: 'Bug is created', description: 'https://jira.com/BUG-123' });
        // ...
    });
```

image

The suggestion is: to add one new parameter `link`
And show it to the reporter as a clickable link. Link should be opened in new tab (target: _blank): 

image

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/20584

### DIFF
--- a/packages/html-reporter/src/testCaseView.tsx
+++ b/packages/html-reporter/src/testCaseView.tsx
@@ -14,7 +14,7 @@
   limitations under the License.
 */
 
-import type { TestCase } from './types';
+import type { TestCase, TestCaseAnnotation } from './types';
 import * as React from 'react';
 import { TabbedPane } from './tabbedPane';
 import { AutoChip } from './chip';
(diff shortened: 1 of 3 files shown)


packages/html-reporter/src/testCaseView.tsx
```diff
@@ -52,6 +49,21 @@ export const TestCaseView: React.FC<{
   </div>;
 };
 
+function renderAnnotationDescription(description: string) {
+  if (description.startsWith('http://') || description.startsWith('https://'))
```
dgozman, line 53: For proper URL validation:

```suggestion
  try {
    if (['http:', 'https:'].includes(new URL(description).protocol))
      ...
  } catch {}
  return description;
```

packages/html-reporter/src/testCaseView.tsx
```diff
@@ -52,6 +49,21 @@ export const TestCaseView: React.FC<{
...
+    return <a href={description} target='_blank' rel='noopener noreferrer'>{description}</a>;
+  return description;
+}
+
+function TestAnnotation({ annotation: { type, description } }: { annotation: TestCaseAnnotation }) {
```
dgozman, line 58:
```suggestion
function TestCaseAnnotationView({ annotation: { type, description } }: { annotation: TestCaseAnnotation }) {
```

dgozman (approved):

PR #21178, opened 2023-02-24
chore: properly use primitive type over wrapper

Andarist:
The problem was caught here: https://github.com/microsoft/TypeScript/pull/52947#issuecomment-1443147890


mxschmitt (approved):

mxschmitt:
Thanks!

PR #21196, opened 2023-02-24
browser(webkit): Restrain sending http credentials on a specific origin

### LINKED ISSUE #20464: [Feature] Restrain sending http credentials on a specific hostname

SebastienRichert:
Fixes https://github.com/microsoft/playwright/issues/20464

For security purpose, we would like to restrain sending HTTP credentials to only the specified server. The idea is to give the ability to specify a origin (scheme://host:port) additionally to current pair username/password. When an authorization response is received from servers, the credentials are sent only if the server origin in the request matches case insensitive the specified origin.

This is a separate PR following https://github.com/microsoft/playwright/pull/20374 in order to update the Webkit patch separatly.

### DIFF
--- a/browser_patches/webkit/patches/bootstrap.diff
+++ b/browser_patches/webkit/patches/bootstrap.diff
@@ -684,7 +684,7 @@ new file mode 100644
 index 0000000000000000000000000000000000000000..347a01b3fdd1a8277cb4104558e8bbfa63539374
 --- /dev/null
 +++ b/Source/JavaScriptCore/inspector/protocol/Emulation.json
-@@ -0,0 +1,51 @@
+@@ -0,0 +1,52 @@
 +{
 +    "domain": "Emulation",
 +    "availability": ["web"],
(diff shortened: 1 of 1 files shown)


browser_patches/webkit/patches/bootstrap.diff
```diff
@@ -16598,12 +16599,12 @@ index 0000000000000000000000000000000000000000..ae45b4212bdb3f6a004cc80a1d91146b
 +    return { };
 +}
 +
-+Inspector::Protocol::ErrorStringOr<void> WebPageInspectorEmulationAgent::setAuthCredentials(const String& username, const String& password)
++Inspector::Protocol::ErrorStringOr<void> WebPageInspectorEmulationAgent::setAuthCredentials(const String& username, const String& password, const String& hostname)
 +{
-+    if (!!username && !!password)
-+        m_page.setAuthCredentialsForAutomation(WebCore::Credential(username, password, CredentialPersistencePermanent));
++    if (!!username && !!password && !!hostname)
```
yury-s, line 16604: We should not check for !!hostname here as it will break the code that doesn't restrict it to a particular domain.

yury-s (approved):
It would be nice if you could also post a link to the original webkit diff (similar to e.g. https://github.com/yury-s/WebKit/commit/850b1e6ffc6f6ce4b1c2e64c15c7fd4b198ca63c), it would make it easier for us to apply the change.

yury-s:
I edited the PR to accept origin and merged this upstream so that we can build new webkit with it. Should be included in webkit 1806. Closing this PR now since it's migrated to our internal build repo.

mxschmitt:
Note: its 1807 now instead of 1806. Currently compiling.

SebastienRichert:
Webkit diff: https://github.com/yury-s/WebKit/commit/2d0832cbebde4696a6d706d13ea6fa061baf7b46

PR #21199, opened 2023-02-24
 chore(trace): include expect steps in a trace

pavelfeldman:


packages/playwright-test/src/matchers/expect.ts
```diff
@@ -236,22 +241,32 @@ class ExpectMetaInfoProxyHandler {
...
+        }
         if (this._info.isSoft)
           testInfo._failWithError(serializerError, false /* isHardError */);
         else
           throw jestError;
       };
 
+      const finalizer = () => {
+        if (traceEvent)
+          traceEvent.metadata.endTime = monotonicTime();
```
dgozman, line 257: We should probably update trace event title from `step.refinedTitle`.

packages/playwright-core/src/utils/traceUtils.ts
```diff
@@ -37,3 +42,43 @@ export function serializeClientSideCallMetadata(metadatas: ClientSideCallMetadat
...
+      zipFile.outputStream.pipe(fs.createWriteStream(fileName)).on('close', f);
+    });
+  });
+}
+
+export function createTraceEventForExpect(apiName: string, expected: any, stack: StackFrame[], wallTime: number): ActionTraceEvent {
+  return {
+    type: 'action',
+    metadata: {
+      id: 'expect@' + wallTime,
```
dgozman, line 71: Couldn't there be two expect calls at the same `wallTime`? Do I misunderstand it?

packages/trace-viewer/src/ui/modelUtil.ts
```diff
@@ -74,6 +75,41 @@ function indexModel(context: ContextEntry) {
...
+    const offet = callActions[0].metadata.startTime - callActions[0].metadata.wallTime!;
+    for (const expectAction of expectActions) {
+      const duration = expectAction.metadata.endTime - expectAction.metadata.startTime;
+      expectAction.metadata.startTime = expectAction.metadata.wallTime! + offet;
+      expectAction.metadata.endTime = expectAction.metadata.startTime + duration;
+    }
+  }
+  const callActionsByTime = new Map<number, ActionTraceEvent>();
+  for (const action of callActions)
+    callActionsByTime.set(action.metadata.wallTime!, action);
```
dgozman, line 95: What if there are two call actions at the same time?

packages/trace-viewer/src/ui/modelUtil.ts
```diff
@@ -74,6 +75,41 @@ function indexModel(context: ContextEntry) {
...
+
+  const result = [...callActions];
+  for (const expectAction of expectActions) {
+    const callAction = callActionsByTime.get(expectAction.metadata.wallTime!);
```
dgozman, line 99: Can we also check `callAction.apiName`? If I understand correctly, it would be `expect.toBeVisible` or something like that? Otherwise, we'll probably pair up `expect(await locator.isVisible()).toBeTruthy()` because both `expect` and `isVisible` could have the same wall time.

packages/trace-viewer/src/ui/callTab.tsx
```diff
@@ -39,10 +40,8 @@ export const CallTab: React.FunctionComponent<{
   const wallTime = action.metadata.wallTime ? new Date(action.metadata.wallTime).toLocaleString() : null;
   const duration = action.metadata.endTime ? msToString(action.metadata.endTime - action.metadata.startTime) : 'Timed Out';
   return <div className='call-tab'>
-    <div className='call-error' key='error' hidden={!error}>
-      <div className='codicon codicon-issues'/>
-      {error}
-    </div>
+    {!!error && <ErrorMessage error={error}></ErrorMessage>}
+    {!!error && <div className='call-section'>Call</div>}
```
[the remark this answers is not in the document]
pavelfeldman: So that it was either
<Call body>
or
<Error>
<Call header>
<Call body>

packages/trace-viewer/src/ui/callTab.tsx
```diff
@@ -146,3 +145,40 @@ function parseSerializedValue(value: SerializedValue, handles: any[] | undefined
...
+const ErrorMessage: React.FC<{
+  error: string;
+}> = ({ error }) => {
+  const html = React.useMemo(() => {
+    const config: any = {
+      bg: 'var(--vscode-panel-background)',
+      fg: 'var(--vscode-foreground)',
+    };
+    config.colors = ansiColors;
+    return new ansi2html(config).toHtml(escapeHTML(error));
```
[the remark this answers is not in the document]
pavelfeldman: Our deps checker won't be happy - there is no package.json in packages/web...

packages/trace-viewer/src/ui/modelUtil.ts
pavelfeldman: Hm, I ended up matching by time based on your recommendation.

packages/playwright-test/src/matchers/expect.ts
pavelfeldman: I match by apiName + wallTime now, so can't do.

packages/playwright-core/src/utils/traceUtils.ts
```diff
@@ -92,3 +95,51 @@ export async function mergeTraceFiles(fileName: string, temporaryTraceFiles: str
...
+    params: { 'expected': bestEffortJsonStringify(expected) },
+    snapshots: [],
+    log: [],
+    stack,
+  };
+}
+
+function bestEffortJsonStringify(object: any): string {
+  try {
+    return JSON.stringify(object);
```
dgozman, line 141: `Buffer`s are stringifiable, so any `toMatchSnapshot` call will be very large. Let's skip them?

packages/playwright-core/src/utils/traceUtils.ts
```diff
@@ -92,3 +95,51 @@ export async function mergeTraceFiles(fileName: string, temporaryTraceFiles: str
...
+  zipFile.addBuffer(Buffer.from(lines.join('\n')), 'trace.trace');
+
+  if (saveSources) {
+    const sourceFiles = new Set<string>();
+    for (const event of traceEvents) {
+      for (const frame of event.stack || [])
+        sourceFiles.add(frame.file);
+    }
+    for (const sourceFile of sourceFiles) {
+      const source = await fs.promises.readFile(sourceFile, 'utf8');
```
dgozman, line 111: - In `LocalUtilsDispatcher` we `calculateSha1` based on the filename, not file content. I think we should do the same thing to avoid duplication.
- File might not exist for whatever reason, so we should try/catch here.

packages/playwright-test/src/index.ts
```diff
@@ -426,7 +426,18 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
       await stopTracing(tracing);
     })));
 
-    // 6. Either remove or attach temporary traces and screenshots for contexts closed
+
+    // 6. Save test trace.
+    if (captureTrace) {
```
dgozman, line 431:
```suggestion
    if (preserveTrace) {
```

packages/playwright-test/src/matchers/expect.ts
```diff
@@ -236,22 +241,35 @@ class ExpectMetaInfoProxyHandler {
         }
 
         const serializerError = serializeError(jestError);
-        step.complete({ error: serializerError });
+        if (traceEvent) {
+          traceEvent.error = { name: jestError.name, message: jestError.message, stack: jestError.stack };
+          traceEvent.apiName = traceEvent.apiName;
```
dgozman, line 246: This line looks suspicious 😄

packages/playwright-test/src/matchers/expect.ts
```diff
@@ -236,22 +241,35 @@ class ExpectMetaInfoProxyHandler {
...
+        }
         if (this._info.isSoft)
           testInfo._failWithError(serializerError, false /* isHardError */);
         else
           throw jestError;
       };
 
+      const finalizer = () => {
+        if (traceEvent) {
+          traceEvent.apiName = traceEvent.apiName;
```
dgozman, line 258: This one as well.

packages/playwright-test/src/worker/testInfo.ts
```diff
@@ -23,6 +23,7 @@ import type { TestCase } from '../common/test';
 import { TimeoutManager } from './timeoutManager';
 import type { Annotation, FullConfigInternal, FullProjectInternal, Location } from '../common/types';
 import { getContainedPath, normalizeAndSaveAttachment, sanitizeForFilePath, serializeError, trimLongString } from '../util';
+import type * as trace from '../../../trace/src/trace';
```
dgozman, line 26:
```suggestion
import type * as trace from '@trace/trace';
```

packages/trace-viewer/src/ui/modelUtil.ts
```diff
@@ -74,6 +75,41 @@ function indexModel(context: ContextEntry) {
...
+    const callAction = callActionsByKey.get(expectAction.apiName + '@' + expectAction.wallTime);
+    if (callAction) {
+      if (expectAction.error)
+        callAction.error = expectAction.error;
+      continue;
+    }
+    result.push(expectAction);
+  }
+
+  // Prefer client-side wall time when available.
```
dgozman, line 108: I don't really understand this comment here.

dgozman (approved):

packages/playwright-core/src/utils/traceUtils.ts
pavelfeldman: I think it is ok

PR #21203, opened 2023-02-25
chore: simplify the trace collection fixtures

pavelfeldman:


packages/playwright-test/src/index.ts
```diff
@@ -307,7 +307,9 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
     };
 
     const startedCollectingArtifacts = Symbol('startedCollectingArtifacts');
-    const stopTraceChunkOnContextClosure = async (tracing: Tracing) => {
+    const stopTracing = async (tracing: Tracing) => {
+      if ((tracing as any)[startedCollectingArtifacts])
```
dgozman, line 311: This will break for "reuse page between tests" scenario, because we will not collect traces for the second test due to this flag already being set.

dgozman (approved):

packages/playwright-test/src/index.ts
pavelfeldman: I don't think this change affects the behavior, so it keeps things the way they were.

PR #21212, opened 2023-02-26
feat(html-report): support objects as description

gbasto:
Currently with an object is set as annotation's description, the report won't even render that test.

With this change, if the annotation's description is an object, it will be stringified and rendered correctly.


pavelfeldman:
Please start with filing an issue and describing the problem along with the proposed solution!

gbasto:
> Please start with filing an issue and describing the problem along with the proposed solution!

https://github.com/microsoft/playwright/issues/21253

Happy to reopen this PR if we agree on the problem/solution.

Issue #18288, opened 2022-10-24
[Question] How to handle subsequent GET requests returning server state during HAR replay?


leomelzer:
Hello there! I have a question about HAR replay and subsequent (matching) GET requests.

Some details:

**Context:**
- Playwright Version: 1.27.1
- Operating System: macOS 12.5.1
- Node.js version: 16.15.0
- Browser: All
- Extra: -

**Code Snippet**

- Using `npx playwright codegen` I captured a test scenario + matching HAR file
- When replaying, I noticed the same subsequent GET requests always **only return the first match form the HAR file**
- This will be wrong, if a UI interaction modifies the server state
- Example:

```ts
import path from "path";
import { test, expect } from "@playwright/test";

test("test", async ({ page }) => {
	await page.routeFromHAR(
		path.resolve(process.cwd(), "tests", "scenario.har"),
		{
			url: "https://api.local.dev/**",
		}
	);

	// When loading `/`, some client side javascript fetches from the api.local.dev
	// First GET returns initial value `{text: "Hello"}`
	await page.goto("/");
	await expect(page.locator(".message")).toHaveText("Hello");

	// Some POST requests modifying server state, setting `text = "World"`
	await page.getByRole("button", { name: "Submit" }).click();

	// Second GET still returns initial value `{text: "Hello"}`
	// I'd expect `{text: "World"}` though
	await page.goto("/");
	await expect(page.locator(".message")).toHaveText("World"); // ❌ This fails
});
```

- The current behavior is in line with the documentation (matching headers etc, see https://playwright.dev/docs/network#replaying-from-har) but I find it weird to model a real application scenario with this
- I believe the issue (?) lies within the way playwright decides which request to match here: https://github.com/microsoft/playwright/blob/4ed2a01d9c4a372b7aee7931d32003261cb0351b/packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts#L282-L291
- How do I get playwright to return the "next" GET response from the HAR file? I'm not sure if it would be within the semantics of HAR replay to only ever return a HAR log _once_ (hence removing it once returned)?

Thank you for any pointers!

olivierbeaulieu:
Hi @leomelzer! I was running into the same problem, and found this issue.

I've figured out a way to make this work - although I'd love to see the Playwright team come up with a recommended approach.

Rather than having a single HAR file for the entire test, I create multiple HARs by creating "checkpoints". Here's a simplified version of my solution:

```ts

[...]

PR #21223, opened 2023-02-27
feat(routeFromHAR): add arguments

NoamGaash:
When using `routeFromHAR` to record traffic for specific test cases, the user has less control over the recording mechanism (compared to recording HARs in the browser context).
This commit suggests passing more arguments, giving the user more control.

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -1128,9 +1128,9 @@ A glob pattern, regular expression or predicate to match the request URL. Only r
 * since: v1.32
 - `mode` <[HarMode]<"full"|"minimal">>
 
-When set to `minimal`, only record information necessary for routing from HAR. This omits sizes, timing, page, cookies, security and other types of HAR information that are not used when replaying from HAR. Defaults to `full`.
+When set to `minimal`, only record information necessary for routing from HAR. This omits sizes, timing, page, cookies, security and other types of HAR information that are not used when replaying from HAR. Defaults to `minimal`.
 
-### option: BrowserContext.routeFromHAR.url
+### option: BrowserContext.routeFromHAR.content
 * since: v1.32
 - `content` <[HarContentPolicy]<"omit"|"embed"|"attach">>
 
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -1121,8 +1121,17 @@ If specified, updates the given HAR with the actual network information instead 
 ### option: BrowserContext.routeFromHAR.url
 * since: v1.23
 - `url` <[string]|[RegExp]>
-
 A glob pattern, regular expression or predicate to match the request URL. Only requests with URL matching the pattern will be served from the HAR file. If not specified, all requests are served from the HAR file.
+
+### option: BrowserContext.routeFromHAR.mode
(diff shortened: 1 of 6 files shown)


NoamGaash:

pavelfeldman:
Your change looks good, but I'd like to better understand your use case. Could it be that you don't really need routeFromHAR and you use it to get a manual control on when har starts / stops recording? We might expose a different API for that on sat context.tracing.startHAR / stopHAR.

docs/src/api/class-browsercontext.md
```diff
@@ -1124,6 +1124,18 @@ If specified, updates the given HAR with the actual network information instead
 
 A glob pattern, regular expression or predicate to match the request URL. Only requests with URL matching the pattern will be served from the HAR file. If not specified, all requests are served from the HAR file.
 
+### option: BrowserContext.routeFromHAR.mode
+* since: v1.32
+- `mode` <[HarMode]<"full"|"minimal">>
+
+When set to `minimal`, only record information necessary for routing from HAR. This omits sizes, timing, page, cookies, security and other types of HAR information that are not used when replaying from HAR. Defaults to `full`.
+
+### option: BrowserContext.routeFromHAR.url
```
pavelfeldman, line 1133:
```suggestion
### option: BrowserContext.routeFromHAR.content
```

docs/src/api/class-browsercontext.md
```diff
@@ -1124,6 +1124,18 @@ If specified, updates the given HAR with the actual network information instead
 
 A glob pattern, regular expression or predicate to match the request URL. Only requests with URL matching the pattern will be served from the HAR file. If not specified, all requests are served from the HAR file.
 
+### option: BrowserContext.routeFromHAR.mode
+* since: v1.32
+- `mode` <[HarMode]<"full"|"minimal">>
+
+When set to `minimal`, only record information necessary for routing from HAR. This omits sizes, timing, page, cookies, security and other types of HAR information that are not used when replaying from HAR. Defaults to `full`.
```
pavelfeldman, line 1131:
```suggestion
When set to `minimal`, only record information necessary for routing from HAR. This omits sizes, timing, page, cookies, security and other types of HAR information that are not used when replaying from HAR. Defaults to `minimal`.
```

NoamGaash:
Thanks!
In my company, we love using Playwright to test our client side.
We want to isolate some of our tests from the backend and mock the network traffic using HAR files.

We have some tests that change the state on the server side, and in order to mock this behavior and make sequential GET requests serve different outputs, we have used [this solution](https://github.com/microsoft/playwright/issues/18288#issuecomment-1324056041) by @olivierbeaulieu.

our code seems like this:
```js
await harCheckpoint('test-checkpoint-1')
// go to the page
//get data
await harCheckpoint('test-checkpoint2')
// perform action that changes the data
// get updated data
```
When the HAR recorder operates in `attach` mode, many different files are created, and it's very tricky to follow which file belongs to each checkpoint.

I want to record my HAR filed in `embed` mode and in this pull request I found a way to do it as a non-breaking change :)

Thank you!

pavelfeldman:
Ok, sounds good. I was verifying you were using recordHAR as recordHAR, which it looks like what you are doing.

PR #21324, opened 2023-03-02
fix(oopif): get response body from adopting frame

### LINKED ISSUE #20809: [BUG] Intercepting response body from oopif throws error
**Context:**
- Playwright Version: 1.30.0
- Operating System: Linux
- Node.js version: 18.4.0
- Browser: Chromium
- Extra: n/a

**Code Snippet**
```javascript

test('route can access response body of oopif requests', async ({ page }) => {
  await page.on('response', async response => console.log(await response.body()));
  await page.goto(server.PREFIX + '/dynamic-oopif.html');
});
```
In this case dynamic-oopif.html is the standard playwright test file. This also reproduces for a simple static page with an oopif like:
`<html><body><iframe src="https://example.com/" width="400" height="400"></body></html>`

**Describe the bug**

Intercepting the response body of the out-of-process iframe results in this error:
`response.body: Protocol error (Network.getResponseBody): No resource with given identifier found`

For some reason this seems to work in _headless_, but running this test headed throws.

I've seen a few similar issues that mention there's sometimes a race condition where the Network stack will dispose of response body on a redirect, but there is no redirect here. Am I missing something obvious?

Thanks!

yury-s:
Fixes https://github.com/microsoft/playwright/issues/20809

### DIFF
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -284,18 +284,19 @@ export class CRNetworkManager {
       const contentLengthHeader = Object.entries(responsePayload.headers).find(header => header[0].toLowerCase() === 'content-length');
       const expectedLength = contentLengthHeader ? +contentLengthHeader[1] : undefined;
 
-      const response = await this._client.send('Network.getResponseBody', { requestId: request._requestId });
+      const client = request._adoptingOopifSession ?? this._client;
+      const response = await client.send('Network.getResponseBody', { requestId: request._requestId });
       if (response.body || !expectedLength)
         return Buffer.from(response.body, response.base64Encoded ? 'base64' : 'utf8');
 
       // For <link prefetch we are going to receive empty body with non-emtpy content-length expectation. Reach out for the actual content.
-      const resource = await this._client.send('Network.loadNetworkResource', { url: request.request.url(), frameId: this._serviceWorker ? undefined : request.request.frame()!._id, options: { disableCache: false, includeCredentials: true } });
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -474,6 +476,7 @@ class InterceptableRequest {
   readonly _wallTime: number;
   private _route: RouteImpl | null;
   private _redirectedFrom: InterceptableRequest | null;
+  _adoptingOopifSession: CRSession | undefined;
```
dgozman, line 479: nit: I'd rather make this `owningNetworkManager: CRNetworkManager`.

tests/library/chromium/oopif.spec.ts
```diff
@@ -336,6 +336,16 @@ it('should not throw on exposeFunction when oopif detaches', async ({ page, brow
   expect(await page.evaluate(() => (window as any).myFunc())).toBe(2022);
 });
 
+it.only('should intercept response body from oopif', async function({ page, browser, server }) {
```
dgozman, line 339:
```suggestion
it('should intercept response body from oopif', async function({ page, browser, server }) {
```

tests/library/chromium/oopif.spec.ts
```diff
@@ -336,6 +336,16 @@ it('should not throw on exposeFunction when oopif detaches', async ({ page, brow
...
+  const [response] = await Promise.all([
+    page.waitForResponse('**/grid.html'),
+    page.goto(server.PREFIX + '/dynamic-oopif.html')
+  ]);
+  console.log('XXX');
```
dgozman, line 344:
```suggestion
```

dgozman (approved):

PR #21325, opened 2023-03-02
chore: ui mode ui improvements

pavelfeldman:


packages/playwright-test/src/runner/uiMode.ts
```diff
@@ -85,8 +85,13 @@ class Controller {
     this._page.exposeBinding('binding', false, (source, data) => {
       const { method, params } = data;
       if (method === 'run') {
-        const { location } = params;
-        config._internal.cliArgs = [location];
+        const { location, testIds } = params;
+        if (location)
+          config._internal.cliArgs = [location];
```
dgozman, line 90: Should we reset `cliArgs` and `testIdMatcher` from previous run when `!location` and `!testIds` respectively?

dgozman (approved):

PR #21332, opened 2023-03-02
docs: add projects page

debs-obrien:


docs/src/test-projects-js.md
```diff
@@ -0,0 +1,118 @@
...
+id: test-projects
+title: "Projects"
+---
+
+
+Playwright supports multiple **projects** so you can run your tests in multiple browsers such as chromium, webkit and firefox browsers as well as branded browsers such as Google Chrome and Microsoft Edge. Playwright can also run on emulated tablet and mobile devices. See the [registry of device parameters](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/deviceDescriptorsSource.json) for a complete list of selected desktop, tablet and mobile devices. 
+
+Each project is given a name and a set of options which are configured in the `use` object.
+
+## Project Configuration
```
dgozman, line 11: Here is a high-level structure suggestion. Let me know what you think.

1. What is a project? - A logical group of tests running in the same configuration. Here we talk about most common usecase that is running tests in a specific browser. We can also explain other benefits of projects: running groups of tests with different timeout or retries, against different environment, splitting tests per package/functionality and more.
2. Configure projects for multiple browsers. - This section.
3. Run projects.
4. Configure projects for multiple environments. - An example with baseURL and retries.
5. Splitting tests into projects. - Here we can talk about testMatch and other filters. Also `@smoke` example that uses `grep` would fit here.
6. Dependencies.

debs-obrien: ok re worked, let me know what you think

docs/src/test-projects-js.md
```diff
@@ -0,0 +1,180 @@
...
+
+```js
+import { defineConfig } from '@playwright/test';
+
+export default defineConfig({
+  timeout: 60000, // Timeout is shared between all tests.
+  projects: [
+    {
+      name: 'staging',
+      baseURL: 'staging.example.com',
```
dgozman, line 106: `baseURL` goes into `use` section.

dgozman (approved):

Issue #11122, opened 2021-12-28
[Feature] `install-deps` should be more friendly to users of other Linux distros


Luc45:
Currently, only Ubuntu 18.04 and 20.04 are officially supported by Playwright.

Out of the box on Manjaro, playwright is able to be installed and run a tests both in headless and headed mode:

1. npm i -D @playwright/test
2. npx playwright install
3. npx playwright test

Code under test:

```js
const { test, expect } = require('@playwright/test');

test('basic test', async ({ page }) => {
  await page.goto('https://foo.loc/');
  const title = page.locator('.site-title a');
  await expect(title).toHaveText('Foo Test');
});
```

It works.

However, when I try to execute the same test with more browsers after specifying `chromium`, `firefox` and `webkit` on `playwright.config.js`, playwright prompts me to install dependencies:

![image](https://user-images.githubusercontent.com/9341686/147602962-83cb3bab-06af-4e69-af93-085fd9a7032b.png)

When running `npx playwright install-deps`, I get:

```
Installing Ubuntu dependencies...
sh: line 1: apt-get: command not found
```

The error is clear, it tries to install some package with `apt-get`, which does not exist in Arch Linux. However, it doesn't tell me **which one**, so that I can install it on my own.

To get around this, I was trying to figure out what I needed to install on my own. The closest I could get was to this list: https://github.com/microsoft/playwright/blob/d34f9972abf3e412c9a73ee7b524e529a89136f8/packages/playwright-core/src/utils/nativeDeps.ts

But surely I don't need all of that, as the Chromium works out of the box both headless and headed, so there must be some package missing in my environment that I need to install.

My request is that `install-deps` be specific about what dependencies are missing, so that users of other distros of Linux can have a chance to manually install the dependencies themselves.

aslushnikov:
@Luc45 fair enough. Would you like to submit PR to have a more friendly message with the list of packages for non-ubuntu distros?

You'd need to enhance this function: https://github.com/microsoft/playwright/blob/fba29aa31942e81ebf7c70ef53eda97aeab24017/packages/playwright-core/src/utils/dependencies.ts#L148

Luc45:
> @Luc45 fair enough. Would you like to submit PR to have a more friendly message with the list of packages for non-ubuntu distros?
> 
> You'd need to enhance this function:
> 
[...]

Issue #8100, opened 2021-08-09
[Feature]: support Arch linux


aslushnikov:
**Edit by maintainers**

Hey everybody!

There's now an experimental support for all linux distributions using our browser container image  `mcr.microsoft.com/playwright`.
If you can run this container image on your linux distribution, then you can instruct `@playwright/test` to connect to it and run browsers inside the container.

Please give it a try and let us know what you think here: https://github.com/microsoft/playwright/issues/26482

---

### Feature request

This is a tracking bug to gauge public interest in Arch linux support.

keevie:
Not a big deal because I can just run in docker, but it would be nice!

pavelfeldman:
#### Why was this issue closed?

We are prioritizing the features based on the upvotes, recency and activity in the issue thread. It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it. We are closing issues like this one to keep our bug database maintainable. Please feel free to open a new issue and link this one to it if you think this is a mistake.

ShivamJoker:
Why can't playwright work on all OS? Why only support ubuntu?

joe733:
Okay, so I install `playwright` in a clean [`chroot`](https://wiki.archlinux.org/title/DeveloperWiki:Building_in_a_clean_chroot) which had only `base-devel` package.

```console
$ mkdir ~/.chroot
$ export $CHROOT=$HOME/.chroot
$ mkarchroot $CHROOT/root base-devel
$ arch-nspawn $CHROOT/root pacman -Syu
$ git clone 'https://aur.archlinux.org/playwright.git'
$ cd playwright
$ makechrootpkg -c -r $CHROOT
$ ls -la       
Permissions Size User  Date Modified Name
drwxr-xr-x     - us-er 19 May 19:53  .git
.rw-r--r--   534 us-er 19 May 19:53  .SRCINFO
.rw-r--r--   940 us-er 19 May 19:53  PKGBUILD
.rw-r--r--  1.6M us-er 19 May 19:59  playwright-1.33.0-1-any.pkg.tar.zst
.rw-r--r--    24 us-er 19 May 19:59  playwright-1.33.0-1-x86_64-package.log
.rw-r--r--  7.8k us-er 19 May 19:53  playwright-1.33.0.tar.gz
$ sudo cp playwright-1.33.0-1-any.pkg.tar.zst $CHROOT/root/home/
[...]

PR #21336, opened 2023-03-02
feat: add support for "arch" and Arch-like distro to `install-deps`

ndom91:
Added support for `arch` and `arch-arm64` to the `install-deps` command, including the helper functions for grabbing `distroInfo`, etc.

The `id_like` `/etc/os-release` field is used (in the arch ecosystem at least) for Arch forks like Endeavour and Manjaro. These should still define `ID_LIKE=arch` in their `os-release` :+1: 

I saw the discussion in https://github.com/microsoft/playwright/issues/11122 and based the required packages and required symlinks off of that.

I can understand that Arch isn't something you do testing on normally, but having them in here at least will make it easier for arch users to get up and running now, and community members can much more easily just update the package name(s) in `nativeDeps.ts` in the future should the `pacman`/`yay` package names change or anythign like that! 

Sidenote for those not familiar, arch uses `pacman` as the primary package manager, but `yay` is the defacto standard pkg manager wrapper / helper that also gives access to the AUR (Arch User Repository) which all Arch users will most likely have installed.

### DIFF
--- a/packages/playwright-core/src/server/registry/dependencies.ts
+++ b/packages/playwright-core/src/server/registry/dependencies.ts
@@ -105,15 +105,15 @@ export async function installDependenciesLinux(targets: Set<DependencyGroup>, dr
   if (platform.includes('arch')) {
     const { code } = await spawnAsync('command', ['-v', 'yay']);
     if (code !== 0) {
-      console.log('Installing playwright deps in Arch distros requires the pacman helper "yay"')
-      return
+      console.log('Installing playwright deps in Arch distros requires the pacman helper "yay"'); // eslint-disable-line no-console
+      return;
     }
 
     commands.push('pacman -Syy');
     commands.push(['yay', '-S', '--noconfirm', '--removemake', '--noprovides', '--answerdiff', 'None', '--answerclean', 'None', '--mflags', '"--noconfirm"',
       ...uniqueLibraries,
     ].join(' '));
-    commands.push('sudo ln -s /usr/lib/libpcre.so /usr/lib/libpcre.so.3')
+    commands.push('sudo ln -s /usr/lib/libpcre.so /usr/lib/libpcre.so.3');
   } else {
     commands.push('apt-get update');
     commands.push(['apt-get', 'install', '-y', '--no-install-recommends',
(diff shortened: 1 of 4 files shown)


ndom91:

mxschmitt:
Hey, thank you for your contribution to the project! I think we are not ready to add any Arch Linux support to the project, since it would require compiling browsers for Arch (and different versions), otherwise the dynamically linked dependencies won't resolve and their versions don't match. Or does it work out of the box if you just install some OS dependencies via yay/pacman? If I remember correctly you had to create some symlinks somewhere and str replace some stuff inside some of the Playwright browsers.

What about creating a third-party guide / gist instead with the steps which hacks are required to use it on Arch Linux so it won't pollute the core?

ndom91:
> Hey, thank you for your contribution to the project! I think we are not ready to add any Arch Linux support to the project, since it would require compiling browsers for Arch (and different versions), otherwise the dynamically linked dependencies won't resolve and their versions don't match. Or does it work out of the box if you just install some OS dependencies via yay/pacman? If I remember correctly you had to create some symlinks somewhere and str replace some stuff inside some of the Playwright browsers.
> 
> What about creating a third-party guide / gist instead with the steps which hacks are required to use it on Arch Linux so it won't pollute the core?

Yeah so there was 1 symlink I had to create, but I've added this here as well (see: https://github.com/microsoft/playwright/pull/21336/files#diff-32f93435358b9404d9fb48826a9f1983a994e9fdf437afe6b0a56b87873ffba1R116). Otherwise, simply `npx playwright install chrome` works out of the box.

The only thing that didn't was the `install-deps` because obviously the package names were different and we had Ubuntu stuff like `apt-get` hardcoded.

mxschmitt:
Does Firefox and WebKit work?

ndom91:
> Does Firefox and WebKit work?

Firefox also worked out of the box, WebKit did not however. See the following error, missing symbol `cmulex_init` :shrug:

image

ndom91:
Also, when trying to download webkit (`npx playwright install webkit`), it says the OS isn't officially supported and will try to download the fallback build, but then just exits immediately after printing that

image

mxschmitt:
It's just exiting because the WebKit executable is already downloaded on your system.

Since we only add features which work across all 3 browsers, I'm going to close it for now. There was not enough public interest in the past for adding Arch Linux support see https://github.com/microsoft/playwright/issues/8100. It would require engineering effort from our side to support another linux distribution, add browser compilation targets and test it accordingly on the new operating system. Maybe even arm64 there.

I re-opened the feature request so we can reconsider it once it gains more interest.

romeerez:
For other ArchiLinux users who got here by searching for the "cmulex_init" problem:

downgrading to 1.30.0 helps: `npx playwright@1.30.0 install`

PR #21337, opened 2023-03-02
docs: update test reporters

debs-obrien:


dgozman (commented):
- Let's not only update config snippets, but test snippets as well, to be consistent.
- In examples that include both config and test, it might be beneficial to leave file names in the comment.

dgozman (approved):

PR #21367, opened 2023-03-03
chore: fix xpath-in-iframe

### LINKED ISSUE #21364: [BUG] locator picker in trace viewer behaves differently to the one in the playwright inspector
### System info
- Playwright Version: 1.32.0-alpha-mar-3-2023
- Operating System: windows 10
- Browser: chromium
- Other info:

### Source code

**Config file**

```ts
// playwright.config.ts
import type { PlaywrightTestConfig } from '@playwright/test'
import { devices } from '@playwright/test'

    testDir: './tests',
    timeout: 50 * 1000,
    expect: {
        timeout: 50000,
    },
    forbidOnly: !!process.env['CI'],
    retries: process.env['CI'] ? 2 : 0,
    reporter: 'html',
    use: {
        actionTimeout: 0,
        trace: 'on',
    },
    projects: [
        {
            name: 'chromium',
            use: {
                ...devices['Desktop Chrome'],
            },
        },
    ],
}

export default config
```

```js
test('asdf', async ({ page }) => {
    await page.goto('https://google.com')
    await page.pause()
})
```

**Steps**
1. run `playwright test`
1. hit run button when the inspector appears
1. when it pauses, input the locator, eg. `//span`
1. hit resume button
1. run `playwright show-report` then click the trace button on the test
1. input the same locator as before, eg. `//span`

**Expected**
the same elements are highlighted in both the inspector and the trace viewer

**Actual**
image
image

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/21364

### DIFF
--- a/packages/playwright-core/src/server/injected/xpathSelectorEngine.ts
+++ b/packages/playwright-core/src/server/injected/xpathSelectorEngine.ts
@@ -21,7 +21,7 @@ export const XPathEngine: SelectorEngine = {
     if (selector.startsWith('/'))
       selector = '.' + selector;
     const result: Element[] = [];
-    const document = root instanceof Document ? root : root.ownerDocument;
+    const document = root.ownerDocument || root;
     if (!document)
       return result;
     const it = document.evaluate(selector, root, null, XPathResult.ORDERED_NODE_ITERATOR_TYPE);


packages/playwright-core/src/server/injected/xpathSelectorEngine.ts
```diff
@@ -21,7 +21,7 @@ export const XPathEngine: SelectorEngine = {
     if (selector.startsWith('/'))
       selector = '.' + selector;
     const result: Element[] = [];
-    const document = root instanceof Document ? root : root.ownerDocument;
+    const document = root.ownerDocument || root;
```
dgozman, line 24: Should we add `Document/Node` to the list of prohibited globals?

dgozman (approved):

mxschmitt:
Test would be nice!

packages/playwright-core/src/server/injected/xpathSelectorEngine.ts
pavelfeldman: I did at first, but I could not find a way to ban it only as a runtime object - it yells at all the type declarations...

PR #21378, opened 2023-03-03
docs: add doc on webserver

debs-obrien:


mxschmitt (approved):
LGTM from my side. Changes to the test-advanced guide are the following:

1. we don't mention `port` option anymore.
2. we don't mention the `DEBUG=pw:webserver` env var anymore

I think we should fix 2 and add it back.

Also might be good if we can explain the logic [how baseURL creates a URL](https://playwright.dev/docs/api/class-testoptions#test-options-base-url) since it created a lot questions from users in the past.

debs-obrien:
ok added back the debug option and explanation for base url

Issue #21182, opened 2023-02-24
[BUG] Browser context closed promise doesn't resolve if chunked request is still active upon test completion on webkit with recordHar


DerGernTod:
**Context:**
- Playwright Version: 1.31.1
- Operating System: Windows
- Node.js version: 16.17.1
- Browser: WebKit

**Code Snippet**
Example playwright test
```javascript
test.only("webkit doesn't resolve .waitForSelector if chunked request is in progress", async ({ page }) => {
    page.on("console", (log) => {
        console.log(log.text());
    });
    await page.goto("http://localhost:8090/ajax/testframework/typescript/tests/modules/webkit-test.jsp", { waitUntil: "load" });
    console.log("goto resolved, waiting for selector...");
    await page.waitForSelector("#pw-finished", { state: "attached", timeout: 15000 });
});
```

Example html
```html
<!DOCTYPE html>
<html>
<head>
    <script type="application/javascript">
        function loadXhr() {
            console.log(performance.now().toFixed(2), "sending xhr...");
            fetch(`chunked-response.jsp`);
            
            setTimeout(function() {
                var div = document.createElement("div");
                div.id = "pw-finished";
                document.body.appendChild(div);
                console.log(performance.now().toFixed(2), "div attached!");
            }, 1000);
        }
    </script>
</head>
<body onload="loadXhr()">
loads some long request onload
</body>
</html>
```

In case you want to verify that element is really attached:
```javascript
var m = new MutationObserver(function () {
    console.log(performance.now().toFixed(2), "mutation happened");
});
m.observe(document.body, {subtree: true, childList: true});
```

An example of how a chunked-response.jsp could look like (it flushes in-between in order to fire progress events):
```jsp
<%--
  responds after 30 seconds, but sends updates every 100ms
--%><%
    for (int i = 0; i < 10; i++) {
        response.getWriter().print(i + " ");
        response.getWriter().flush();
        try {
            Thread.sleep(100);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
    }
    try {
        Thread.sleep(30000);
    } catch (InterruptedException e) {
        e.printStackTrace();
    }
%>

```

**Describe the bug**
On webkit, playwright doesn't resolve the `waitForSelector` promise when the respective element shows up, but only when all chunked requests are done. To me these two things are completely independent, and this doesn't happen for other browsers. It also only seems to happen for chunked requests, for normal requests it resolves in time.

[...]

PR #21392, opened 2023-03-03
test: har does not hang on slow chunked response

yury-s:
#21182

### DIFF
--- a/tests/library/har.spec.ts
+++ b/tests/library/har.spec.ts
@@ -652,7 +652,7 @@ it('should return server address directly from response', async ({ page, server,
     expect(port).toBe(server.PORT);
 });
 
-it('should return security details directly from response', async ({ contextFactory, httpsServer, browserName, platform }) => {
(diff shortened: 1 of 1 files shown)


dgozman (approved):

tests/library/har.spec.ts
```diff
@@ -828,3 +828,37 @@ it('should not hang on resources served from cache', async ({ contextFactory, se
...
+
+it('should not hang on slow chunked response', async ({ browserName, browser, contextFactory, server }, testInfo) => {
+  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/21182' });
+  it.fixme(browserName === 'webkit');
+  server.setRoute('/empty.html', (req, res) => {
+    res.writeHead(200, {
+      'Content-Type': 'text/html',
+    });
+    res.end(`<script>
+    let receivedFirstData = new Promise(f => {
```
vigneshshanmugam, line 840: we are not actually waiting for the progress event correctly when waiting on this promise, since `receivedFirstData` would be undefined. needs to be changed from `let ->  var`

PR #21396, opened 2023-03-03
fix(ct): react17 primitive as slot

sand4rt:
The JS primitive check which was added recently was not added to react17. Related to https://github.com/microsoft/playwright/pull/19814 & https://github.com/microsoft/playwright/pull/20125.

@pavelfeldman Are there any tests need for react17? There are currently no tests for it, only for react18-vite and react18-webpack.

### DIFF
--- a/packages/playwright-ct-react17/registerSource.mjs
+++ b/packages/playwright-ct-react17/registerSource.mjs
@@ -36,9 +36,11 @@ export function register(components) {
 
 /**
  * @param {Component} component
- * @returns {JSX.Element}
  */
 function render(component) {
+  if (typeof component !== 'object' || Array.isArray(component))
+    return component;
+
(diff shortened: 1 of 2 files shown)


packages/playwright-ct-react17/registerSource.mjs
```diff
@@ -36,9 +36,11 @@ export function register(components) {
 
 /**
  * @param {Component} component
- * @returns {JSX.Element}
  */
 function render(component) {
+  if (typeof component !== 'object' || Array.isArray(component))
+    return component;
+
```
sand4rt, line 43: This is the same check as in react18: https://github.com/microsoft/playwright/blob/main/packages/playwright-ct-react/registerSource.mjs#L43

pavelfeldman (approved):

PR #21398, opened 2023-03-04
chore(ui): make ui-side drive things

pavelfeldman:


packages/playwright-test/src/runner/uiMode.ts
```diff
@@ -14,112 +14,144 @@
...
+    clearCompilationCache();
+    reporter.onConfigure(this._config);
+    const stop = new ManualPromise();
+    const run = taskRunner.run(context, 0, stop).then(async status => {
+      await reporter.onExit({ status });
+      return status;
+    });
+    this._testRun = { run, stop };
+    await run;
+    this._testRun = undefined;
```
dgozman, line 120: I'd move this line to 116.5, so that when `stop` awaits `testRun.run` we would be sure there is no run anymore.

dgozman (approved):

PR #21400, opened 2023-03-04
chore(ct): vue prop type of JsonObject

sand4rt:
The `props` type has been changed so that code below throws a type error:

```typescript
test('type error', async ({ mount }) => {
  const component = await mount(Button, {
    props: {
      submit: () => {} // Type '() => void' is not assignable to type 'JsonValue | undefined'
    },
  });
});
```

### DIFF
--- a/packages/playwright-ct-vue/index.d.ts
+++ b/packages/playwright-ct-vue/index.d.ts
@@ -41,19 +41,14 @@ type JsonObject = { [Key in string]?: JsonValue };
 
 type Slot = string | string[];
 
-export interface MountOptions<
-HooksConfig extends JsonObject,
-Props extends Record<string, unknown>
-> {
+export interface MountOptions<HooksConfig extends JsonObject, Props extends JsonObject> {
   props?: Props;
   slots?: Record<string, Slot> & { default?: Slot };
   on?: Record<string, Function>;
   hooksConfig?: HooksConfig;
 }
 
-interface MountResult<
-  Props extends Record<string, unknown>
-> extends Locator {
+interface MountResult<Props extends JsonObject> extends Locator {
   unmount(): Promise<void>;
   update(options: Omit<MountOptions<never, Props>, 'hooksConfig'>): Promise<void>;
 }
@@ -67,12 +62,9 @@ export interface ComponentFixtures {
   mount(component: JSX.Element): Promise<MountResultJsx>;
   mount<HooksConfig extends JsonObject>(
     component: any,
-    options?: MountOptions<HooksConfig, Record<string, unknown>>
-  ): Promise<MountResult<Record<string, unknown>>>;
-  mount<
-    HooksConfig extends JsonObject,
-    Props extends Record<string, unknown> = Record<string, unknown>
-  >(
+    options?: MountOptions<HooksConfig, JsonObject>
+  ): Promise<MountResult<JsonObject>>;
+  mount<HooksConfig extends JsonObject, Props extends JsonObject = JsonObject>(
     component: any,
     options: MountOptions<HooksConfig, never> & { props: Props }
   ): Promise<MountResult<Props>>;
--- a/packages/playwright-ct-vue2/index.d.ts
+++ b/packages/playwright-ct-vue2/index.d.ts
@@ -41,19 +41,14 @@ type JsonObject = { [Key in string]?: JsonValue };
 
 type Slot = string | string[];
 
-export interface MountOptions<
-  HooksConfig extends JsonObject,
-  Props extends Record<string, unknown>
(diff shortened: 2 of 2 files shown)


pavelfeldman (approved):

dospunk:
If we can only pass JsonObjects to props, the how are we supposed to test components that take things like functions or more complex objects as props?

sand4rt:
Vue works with events/emits instead of callbacks/functions as props right? The `await mount(Button, { on: { submit: () => {} })` is used for that.

What do you mean by complex objects?

dospunk:
@sand4rt While the events/emits pattern is typical for Vue, it also does allow taking functions as props. I prefer to use functions as props because I find that it gives me more control through typing with typescript. You can only provide types for events/emits in Vue if you're using the <script setup> syntactic sugar. While it may not be the established best practice, I feel that a complete testing library for Vue should support it because Vue itself supports it. I can open an issue for further discussion if you'd like!

As for more complex objects, I mean objects that may contain functions. For example, an instance of a class.

unitydynamics:
I'd like to add my voice to reverse this change.  You'll find that people cannot move from 1.31 to 1.32 because of this change specifically.  I don't think it had the desired effect in terms of making code "safer"--this new restriction does not match what Vue considers good props.  You really should let Vue decide what a bad prop is (and leave the type to `unknown`).  If you need examples, I can show how this change essentially broke most Typescript code that uses props in their Vue components.

unitydynamics:
@sand4rt The primary issue (at least for me) is that I define all of my interesting props with Typescript interfaces, and the `JsonObject` type doesn't allow for it...  This isn't a full blown component test, but you can get the idea [here](https://www.typescriptlang.org/play?#code/PTAEGEHsDsGcBcBOBDAltetQDNKNAFKwwDyARgFYCmAxvKNJPQO54DW6A5qM6vABagADokhCcqKgBsAJlhlVs6KjJ59BAFQCeQqrBqJUQ+unhVE2ZDT2gAtLZ79UNQckRVQAKgBu5rZ9AaSABbYJgpLVAAV1gVUHRQADUojwBmABpQPV0aVGQpCLUBUAAifUNjUFj4KKESgCh603NLa1AAWS1tXVAAb3rQUDMEAC4qpC4AbnqAX3qguHpgrr14AAVRITHO7o8AXj6h1bGAcn5pKUgT0Dn6+B0PIhgN1GC+VF9QA4RDaG4AHwYUWCZHMoEBZEgkCkVGQ0HBQIK03uPSe0ES+RSX0IxGgLze8A+HkBaPI1DoCLRAEFECgtMiHjiYDS6di0RipCkANoAXQZqNxZNo9AOvVAXIA0lRIgkflweQB+MbszEeGbTeowpZaMnKwWUYUahYIYSbWAjUkGimigam17vXwAZQmf1OrEQbCwSmgVBO6VtzQsVioADFJLIxssNKsNmJMiBQDJIHpoCd6CiPC5aGwAISzRpAA).

sand4rt:
@unitydynamics the example seem to work as expected, do I miss something?:

```typescript
interface MyType {
  test: string;
}
const myTestProp = { test: 'hello' }

type JsonPrimitive = string | number | boolean | null;
type JsonValue = JsonPrimitive | JsonObject | JsonArray;
type JsonArray = JsonValue[];
type JsonObject = { [Key in string]?: JsonValue };

const props: JsonObject = {
  primitiveStringProp: 'works fine',
  objectProp: { test: 'also works fine', test2: 4 }, // there was a missing comma here
  interfaceProp: myTestProp, // doesn't type check!
}
```

dospunk:
@sand4rt you're missing the type definition on myTestProp, line 4 should be
```ts
const myTestProp: MyType = { test: 'hello' }
```

unitydynamics:
You've simply removed Typescript from the relevant code.  
What I have in my code:
`const myTestProp:MyType = {`
is significantly different than what you have in your code: 
`const myTestProp = {`

In your code `myTestProp` could be assigned to anything.  In my code, Typescript will help me ensure that I've created my test prop appropriately.  This is *critical* to writing and maintaining good tests.  Developers usually want *durable* tests, and it's nice when we can lean on Typescript to ensure their test props conform to the expected type.  

So in my opinion, the 1.32 release didn't add anything to type safety of testing--but it made it much worse!

sand4rt:
Ah i see, accidentally remove that part. Hadn't thought of this before.. I usually use `type` instead of `interface`. With `type` you don't have this problem: https://www.youtube.com/watch?v=zM9UPcIyyhQ.

I guess we can't force everyone to use `type` instead of `interface`. @pavelfeldman shall i revert this?

PR #21415, opened 2023-03-06
chore: split code mirror and xterm modules

pavelfeldman:


packages/trace-viewer/src/watch.tsx
```diff
@@ -0,0 +1,41 @@
...
+      await new Promise<void>(f => {
+        navigator.serviceWorker.oncontrollerchange = () => f();
+      });
+    }
+
+    // Keep SW running.
+    setInterval(function() { fetch('ping'); }, 10000);
+  }
+
+  ReactDOM.render(<WatchModeView></WatchModeView>, document.querySelector('#root'));
```
mxschmitt, line 40: lets do the new way? 

```ts
const root = ReactDOM.createRoot(document.getElementById('root')).render(<WatchModeView></WatchModeView>);
```

pavelfeldman: Why?

PR #21454, opened 2023-03-07
fix(types): accept string in expect().toMatch()

### LINKED ISSUE #21453: [REGRESSION]: Type of toMatch assertion changed in Playwright v1.31.0
**Context:**
- GOOD Playwright Version: v1.30.0
- BAD Playwright Version: v1.31.2
- Operating System: Ubuntu 22.04
- Extra: n/a

**Code Snippet**

```javascript
expect('dfblah():<>?df').toMatch('blah():<>?')
```

**Describe the bug**

The type of the `toMatch` assertion has changed such that it longer accepts a string (only a regex). This appears to have changed in f46883e58e12847db71cadfc9364289e71278ee3.

Previously, passing a string did a substring match (avoiding the need to do any escaping for such use cases). This still works in v1.31.2 if the type error is ignored.

(Such a change to the type is also really a breaking change.)

Jest also accepts a string as a comparison: https://jestjs.io/docs/expect#tomatchregexp--string

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/21453

### DIFF
--- a/docs/src/api/class-genericassertions.md
+++ b/docs/src/api/class-genericassertions.md
@@ -418,7 +418,7 @@ expect(value).toMatch(/Is \d+ enough/);
 
 ### param: GenericAssertions.toMatch.expected
 * since: v1.9
-- `expected` <[RegExp]>
+- `expected` <[RegExp]|[string]>
 
 Regular expression to match against.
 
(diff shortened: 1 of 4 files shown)


pavelfeldman (approved):

pavelfeldman:
Let's merge into the branch

Issue #14487, opened 2022-05-30
[Feature] Add label to HTML-report from test for test filtering


bezyakina:
Hi! 
Is it possible to add such beautiful labels to the HTML report directly from the test, even if it is already in the scope of the project or does not have a project at all. It is very convenient to filter the tests in the report using these tags.

Thank you for your attention!

![Screenshot 2022-05-30 at 19 34 20](https://user-images.githubusercontent.com/35337277/171033635-e29d19e8-6511-4c17-8381-67727dba29bc.png)

alexneo2003:
@bezyakina hi
does it look like you wanted?

<img width="1074" alt="Monosnap Playwright Test Report 2023-02-23 21-19-33" src="https://user-images.githubusercontent.com/11705902/221009971-bcc772a7-7992-4c92-9abf-3a6dc9c4742f.png">
<img width="1128" alt="Monosnap Playwright Test Report 2023-02-23 21-20-51" src="https://user-images.githubusercontent.com/11705902/221009965-074a1553-ff2f-4352-9017-eae3450b00c7.png">
<img width="1141" alt="Monosnap Playwright Test Report 2023-02-23 21-21-49" src="https://user-images.githubusercontent.com/11705902/221009961-8327592f-9be1-40a0-801c-48236ac36a81.png">

@pavelfeldman if it does i can make PR
(it will only be necessary to cover it with tests :-))

bezyakina:
> @bezyakina hi
> 
> does it look like you wanted?

@alexneo2003 hi, these tags look nice. I can add yet another custom tag from test? (maybe this feature already implemented as part of another issue)

alexneo2003:
> @alexneo2003 hi, these tags look nice. I can add yet another custom tag from test? (maybe this feature already implemented as part of another issue)

yes, you may write tags in test name like `@smoke` `@regression `etc... and than tests will be filtered about it
now i make some feature for this filters - severals tags filter and after it i make PR

PaulMEdwards:
First, thanks for this functionality, it's pretty good.

However, we have noticed a couple issues with the implementation in #21468... Specifically, tags in a `test.describe` wrapper of other `test`s are ignored and tags with hyphens are broken.

Attached is a screenshot showing both issues. The `test.describe('@Monitoring', ...` tag is ignored and the tag `@call-details` is incorrectly rendered as `call`.

[...]

PR #21468, opened 2023-03-07
feat(html): added labels to test cases and test files

alexneo2003:
Related to #14487 

- match labels from test title by @ symbol, e.g. @smoke, @regression
- added labels to test case view
- added labels to test file view
- handling Meta and Ctrl keys for label selection, if no label is selected then add label to search input, if label is selected then remove label
- simple click on label will add it to search input, click another - will change search input to new label
- combining labels with search input and status filters

https://user-images.githubusercontent.com/11705902/223535701-f9971b3d-5eda-4a09-b6af-c1d8bba6ef03.mp4


packages/html-reporter/src/filter.ts
```diff
@@ -118,12 +125,16 @@ export class Filter {
...
         return false;
     }
-
     if (this.text.length) {
       const matches = this.text.filter(t => searchValues.text.includes(t)).length === this.text.length;
       if (!matches)
         return false;
     }
+    if (this.labels.length) {
+      const matches = this.labels.filter(l => searchValues.text.match(new RegExp(`(\\s|^)${l}(\\s|$)`, 'g'))).length === this.labels.length;
```
pavelfeldman, line 134: `searchValues.text` has just been constructed to contain status, project name, path, title. You don't really want to regex against it. What you probably want it to parse out labels from the test.path + test.title into `searchValues.labels` and check that all the filter labels are present in the `searchValues.labels`.

packages/html-reporter/src/headerView.tsx
```diff
@@ -31,12 +31,25 @@ export const HeaderView: React.FC<React.PropsWithChildren<{
   projectNames: string[],
 }>> = ({ stats, filterText, setFilterText, projectNames }) => {
   React.useEffect(() => {
-    (async () => {
-      window.addEventListener('popstate', () => {
-        const params = new URLSearchParams(window.location.hash.slice(1));
-        setFilterText(params.get('q') || '');
-      });
-    })();
+    const popstateFn = () => {
```
pavelfeldman, line 34: Is this necessary? Looks like this code will end up in the Link/Router infrastructure rather than scattered across the components. Let's do one thing at a time.

packages/html-reporter/src/labels.css
```diff
@@ -0,0 +1,6 @@
+.labels {
```
pavelfeldman, line 1: Missing copyright

packages/html-reporter/src/labels.tsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import { navigate } from './links';
+import type { TestCaseSummary } from './types';
+
+/**
```
pavelfeldman, line 5: Misplaced copyright

packages/html-reporter/src/labels.css
pavelfeldman, line 1: Let's call this `labelsView` to make it a little less ambiguous.

packages/html-reporter/src/labels.tsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
```
pavelfeldman, line 1: labelsView as well

packages/html-reporter/src/labels.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+export const Labels: React.FC<React.PropsWithChildren<{
+  testCase: TestCaseSummary,
+  style?: React.CSSProperties,
+}>> = ({ testCase, style }) => {
+
+  const labels = React.useMemo(() => {
+    return matchTags(testCase.title).sort((a, b) => a.localeCompare(b));
```
pavelfeldman, line 26: nit: Feels like preprocessing to filter out tags would make code look nicer and more efficient.

packages/html-reporter/src/labels.tsx
```diff
@@ -0,0 +1,105 @@
...
+      navigate(`#?q=${searchParams.get('q')?.toString()}` || '');
+    }
+
+    // if metaKey or ctrlKey is pressed, add tag to search query without replacing existing tags
+    // if metaKey or ctrlKey is pressed and tag is already in search query, remove tag from search query
+    if (e.metaKey || e.ctrlKey) {
+      if (!q.includes('@')) {
+        searchParams.set('q', `${q} @${tag}`.trim());
+      } else if (!q.includes(`@${tag}`)) {
+        searchParams.set('q', `${q} @${tag}`.trim());
```
pavelfeldman, line 63: Looks same as in line 61, why different branches?

packages/html-reporter/src/labels.tsx
```diff
@@ -0,0 +1,105 @@
...
+      } else {
+        const re = new RegExp(`@${escapeRegExp(tag)}`, 'g');
```
pavelfeldman, line 65: Any reason to not replace it as a string?

packages/html-reporter/src/labels.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const encoded = encodeURIComponent(tag);
+  const value = tag === encoded ? tag : `"${encoded.replace(/%22/g, '%5C%22')}"`;
+
+  const onClickHandle = (e: React.MouseEvent, tag: string) => {
+    e.preventDefault();
+
+    const searchParams = new URLSearchParams(window.location.hash.slice(1));
+    const q = searchParams.get('q') || '';
+
+    if (searchParams.has('testId')){
```
pavelfeldman, line 51: Looks like we have two different types of links: simple links like `ProjectLink` that you use in tests and in this branch and a smart ctrl+clickable link. Let's split them?

packages/html-reporter/src/testCaseView.tsx
```diff
@@ -36,7 +37,8 @@ export const TestCaseView: React.FC<{
     {test && <div className='test-case-path'>{test.path.join(' › ')}</div>}
     {test && <div className='test-case-title'>{test?.title}</div>}
     {test && <div className='test-case-location'>{test.location.file}:{test.location.line}</div>}
-    {test && !!test.projectName && <ProjectLink projectNames={projectNames} projectName={test.projectName}></ProjectLink>}
+    {test && !!test.projectName && <><span style={{ margin: '2px 8px' }}>Project: <ProjectLink projectNames={projectNames} projectName={test.projectName}></ProjectLink></span></>}
```
pavelfeldman, line 40: Looks like you are looking for `.mx-2` style, we use github style here.

pavelfeldman (commented):
Looks mostly good! Some behavior observations:

- Labels should be rendered similarly, we can remove the "Project" text here and just inline them together:
Screenshot 2023-03-09 at 7 42 50 AM
- Clicking back after navigating by a link takes two "Back" clicks. You should use regular <Link> / <Route> where possible to have it just work
- Project and @ labels are rendered inconsistently in list, should all be one line, like project labels:
Screenshot 2023-03-09 at 7 46 14 AM


---

If it helps, I would remove the ctrl+click for now to make the patch smaller and simpler in terms of simply using hrefs. We can always follow-up to add complexity one step at a time.

alexneo2003:
@pavelfeldman hi, i have some question
i'm moving @ labels aside project label and now we have some issue - as you see labels overflowing test title - main part of test title.
screenshot - it's my real project

Screenshot 2023-03-11 at 16 51 15

i can try do some like this
- for test title using style for wrap text into two lines and ellipsis it
```css
  display: -webkit-box;
  -webkit-line-clamp: 2;
  -webkit-box-orient: vertical;
```
- labels render together with project, if it exists (labels + project zone occupy 40% of row or maybe less)
- if no one - title render for all width

Screenshot 2023-03-11 at 18 24 16

pavelfeldman:
I wonder if we should keep it simple and do the GitHub style regular overflow, nothing fancy:

Screenshot 2023-03-11 at 10 57 25 AM

alexneo2003:
@pavelfeldman i have some problem)
i have no idea how but `imageDiffView` tests will failed.
i have no changing nothing at this component but it failed.
report attached.
[html-reporter-ct-test.zip](https://github.com/microsoft/playwright/files/10951368/html-reporter-ct-test.zip)

packages/html-reporter/src/headerView.tsx
alexneo2003: i'm removed inputFn with his listeners but left `popstateFn` with added `removeEventListener` to avoiding duplication listerers

old implementation 

https://user-images.githubusercontent.com/11705902/224674539-39413e69-9184-44d4-aaf2-a596290d00f1.mp4

new

https://user-images.githubusercontent.com/11705902/224674669-18f4e15e-df5e-48f0-b039-a3701fda5a45.mp4

alexneo2003:
in `labelsView` i'm duplicating from `reportView`
```ts
const testFilesRoutePredicate = (params: URLSearchParams) => !params.has('testId');
const testCaseRoutePredicate = (params: URLSearchParams) => params.has('testId');
```
explain: 
1. in `labelsView` using `Route` and need `testFilesRoutePredicate` functions
2. when i'm run `imageDiffView.spec.tsx` it chaining styles from` ImageDiffView --> TestResultView --> TestCaseView --> LabelsView --> ReportView `(because testFilesRoutePredicate was imported from they) 
3. `ReportView` styles was applied to `ImageDiffView` and style `body { margin: 8px; }` form `user agent style` which was before, will be replaced by a new one --> `body { margin: 0 auto; }`, and three test from `imageDiffView.spec.tsx` will be failed.

we can remove duplication of `testFilesRoutePredicate` and `testCaseRoutePredicate` but then we need to change `imageDiffView.spec.tsx` with new styles.

alexneo2003:
https://user-images.githubusercontent.com/11705902/224677684-bc08172a-ee42-4f02-a5c0-c604235175b8.mp4

packages/html-reporter/src/headerView.tsx
```diff
@@ -47,7 +50,7 @@ export const HeaderView: React.FC<React.PropsWithChildren<{
       <form className='subnav-search' onSubmit={
         event => {
           event.preventDefault();
-          navigate(`#?q=${filterText ? encodeURIComponent(filterText) : ''}`);
+          navigate(`#?q=${filterText ? decodeURIComponent(filterText) : ''}`);
```
pavelfeldman, line 53: nit: i'm not sure we need to encode/decode it here at all since it is a part of the anchor

packages/html-reporter/src/labelsView.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <>
+      {encodedTags.map(tag => (
+        <React.Fragment key={tag}>
+          <Route predicate={testFilesRoutePredicate}>
+            <LabelClick key={tag} tag={tag} />
+          </Route>
+          <Route predicate={testCaseRoutePredicate}>
+            <LabelLink key={tag} tag={tag} />
```
pavelfeldman, line 39: These are seemingly the same, what is the difference and why do you use route here?

packages/html-reporter/src/labelsView.tsx
```diff
@@ -0,0 +1,112 @@
...
+const testFilesRoutePredicate = (params: URLSearchParams) => !params.has('testId');
+const testCaseRoutePredicate = (params: URLSearchParams) => params.has('testId');
+
+export const LabelsView: React.FC<React.PropsWithChildren<{
+  labels: string[],
+}>> = ({ labels }) => {
+  if (!labels.length)
+    return null;
+
+  const encodedTags = labels.map(tag => encodeTag(tag));
```
pavelfeldman, line 29: What's up with this all encode/decode here? :)

packages/html-reporter/src/testFileView.css
```diff
@@ -18,7 +18,7 @@
   line-height: 32px;
   align-items: center;
   padding: 2px 10px;
-  white-space: nowrap;
+  white-space: normal;
```
pavelfeldman, line 21: just drop it.

packages/html-reporter/src/testFileView.css
```diff
@@ -67,3 +67,7 @@
 .test-file-test-outcome-skipped {
   color: var(--color-fg-muted);
 }
+
+.test-file-test-status-icon {
+  flex-shrink: 0;
```
pavelfeldman, line 72: nit: we say flex: none of those.

packages/html-reporter/src/labelsView.tsx
alexneo2003: There are two labels
 LabelLink - 'simple', inside <a/> and href - used for testCaseView
 LabelClick - 'smart', internal onClick handler that works in conjunction with filters by project or status (status AND label) and is used in testFileView

packages/html-reporter/src/headerView.tsx
alexneo2003: encodeURIComponent is not correct method to use it
```
   Expected pattern: /@smoke/
    Received string:  "http://localhost:50414/#?q=%40smoke"
```

packages/html-reporter/src/filter.ts
```diff
@@ -102,7 +110,8 @@ export class Filter {
       const searchValues: SearchValues = {
         text: (status + ' ' + test.projectName + ' ' + test.path.join(' ') + test.title).toLowerCase(),
         project: test.projectName.toLowerCase(),
-        status: status as any
+        status: status as any,
+        labels: (test.path.join(' ') + test.title).toLowerCase(),
```
pavelfeldman, line 114: At this point, it is not distinguishable from the text in terms of the search, so just use `.text`?

packages/html-reporter/src/filter.ts
```diff
@@ -118,12 +127,16 @@ export class Filter {
...
         return false;
     }
-
     if (this.text.length) {
       const matches = this.text.filter(t => searchValues.text.includes(t)).length === this.text.length;
       if (!matches)
         return false;
     }
+    if (this.labels.length) {
+      const matches = this.labels.filter(l => searchValues.labels?.match(new RegExp(`(\\s|^)${escapeRegExp(l)}(\\s|$)`, 'g'))).length === this.labels.length;
```
pavelfeldman, line 136:
```suggestion
      const matches = this.labels.every(l => searchValues.text?.match(new RegExp(`(\\s|^)${escapeRegExp(l)}(\\s|$)`, 'g')));
```

packages/html-reporter/src/filter.ts
```diff
@@ -133,5 +146,6 @@ type SearchValues = {
   text: string;
   project: string;
   status: 'passed' | 'failed' | 'flaky' | 'skipped';
+  labels?: string;
```
pavelfeldman, line 149: Drop?

packages/html-reporter/src/headerView.tsx
pavelfeldman: We parse hash as follows: ```const searchParams = new URLSearchParams(window.location.hash.slice(1));```, so if the value of `filterText` is `foo&bar=baz`, you end up with `q==foo`, which is not expected.

packages/html-reporter/src/labelsView.tsx
pavelfeldman: I see. It looks like it is adding unnecessary complexity: labels control should not depend on the route. It would read simpler if you inlined smart and non-smart labels in the code that needs them respectively. Otherwise, you have 2 different clients that need 2 different controls, but you make them go through a single location that makes sense of the context via route and gives away the right control.

packages/html-reporter/src/testCaseView.tsx
```diff
@@ -84,3 +93,19 @@ function retryLabel(index: number) {
     return 'Run';
   return `Retry #${index}`;
 }
+
+const LabelsLinkView: React.FC<React.PropsWithChildren<{
+  labels: string[],
+}>> = ({ labels }) => {
+  return labels.length > 0 ? (
```
pavelfeldman, line 100: nit: you don't seem to need length check.

pavelfeldman (approved):

PR #21496, opened 2023-03-08
feat(docker): add pnpm to docker images

gierschv:
Hello,

We use your official Docker image in our CI and it works great! But having [pnpm](https://pnpm.io/) inside the image would save an npm install every time we run our tests :)

Thank you!


gierschv:
@microsoft-github-policy-service agree company="Tutteo SAS"

mxschmitt:
Thank you for your contribution! We try to mimic the official Node.js Docker image, which does not have pnpm installed as of today. As you already pointed out it can be installed in the user-land very quickly by a single line (`npm i -g pnpm`) which saves most of the users the 15MB storage footprint which they would not use otherwise.

Issue #19992, opened 2023-01-10
[Question] Is it possible to hide secrets from trace viewer?


ValorHeart:
While logging into my application, the input field contains a password which is a secret that I prefer to hide, is it possible to hide it in the trace viewer?
<img width="475" alt="image" src="https://user-images.githubusercontent.com/61405630/211558455-b26d201d-80cf-43ea-b533-90ed4b8e83f0.png">

MindaugasMateika:
This is not the first issue regarding this issue and I would very much need such feature myself. An option for fill and type to mask the inputted value from reports and traces would be a life saver. Something like:

```
await someLocator.fill('secret text', { maskInReport: true })
```

aslushnikov:
> While logging into my application, the input field contains a password which is a secret that I prefer to hide, is it possible to hide it in the trace viewer?

@ValorHeart there's no way to do so. Even if we mask if from the step itself, we cannot guarantee that
- there's no DOM element that holds the sensitive value (e.g. form input value)
- there's no HTTP request that carried the sensitive value over
- there's no console message that dumped the value

and so on. Traces also store login cookies that can be used to login to the service.

All in all, filtering traces for sensitive information is an uphill battle. I'd recommend against sharing sensitive traces with untrusted parties.

MindaugasMateika:
The mentioned cases are out of scope for Playwright but what is within the scope is to not leak sensitive data in .fill or .type logs. It's a bit dissapointis that these issues are simply closed (this is not the first) because it would certainly get upvotes and is needed by community.

chuanqisun:
Agree with @MindaugasMateika

Just an idea: grep all environment variable (by their values) and remove them from the output folder. It isn't too hard to implement and will make the trace sharing more secure for everyone. Please re-consider this feature request.

carlnoval:
Posted about something similar in discussion: https://github.com/microsoft/playwright/discussions/20902 before finding this thread. Though implementation would be an uphill battle, this feature would greatly help with security especially when sharing reports with steak holders.

Please re-consider this feature request.

jpinheiro27:
This feature should be a must have in terms of security, please could you re-consider it?

p01:
> All in all, filtering traces for sensitive information is an uphill battle. I'd recommend against sharing sensitive traces with untrusted parties.

We don't even have to go as far as sharing traces featuring passwords for them to be problematic.

[...]

PR #21547, opened 2023-03-09
feat: hide secret values passed into type and fill

yury-s:
#19992

### DIFF
--- a/packages/playwright-core/src/server/dispatchers/dispatcher.ts
+++ b/packages/playwright-core/src/server/dispatchers/dispatcher.ts
@@ -330,7 +330,7 @@ function formatLogRecording(log: string[]): string {
 function maybeMaskSensitiveData(params: any) {
   if (!params || !isObject(params))
     return params;
-  if (!(params as any).redactFromLogs)
+  if (!(params as any).secret)
     return params;
   const redacted: any = { ...params };
   for (const name of ['text', 'value']) {
--- a/packages/playwright-core/src/server/dispatchers/dispatcher.ts
+++ b/packages/playwright-core/src/server/dispatchers/dispatcher.ts
@@ -18,7 +18,7 @@ import { EventEmitter } from 'events';
 import type * as channels from '@protocol/channels';
 import { serializeError } from '../../protocol/serializers';
 import { findValidator, ValidationError, createMetadataValidator, type ValidatorContext } from '../../protocol/validator';
-import { assert, isUnderTest, monotonicTime } from '../../utils';
+import { assert, isObject, isString, isUnderTest, maskString, monotonicTime } from '../../utils';
 import { kBrowserOrContextClosedError } from '../../common/errors';
 import type { CallMetadata } from '../instrumentation';
 import { SdkObject } from '../instrumentation';
@@ -257,7 +257,7 @@ export class DispatcherConnection {
       endTime: 0,
       type: dispatcher._type,
       method,
-      params: params || {},
+      params: maybeMaskSensitiveData(params) || {},
       log: [],
       snapshots: []
     };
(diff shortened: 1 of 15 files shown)


packages/playwright-core/src/server/dispatchers/dispatcher.ts
```diff
@@ -326,3 +326,16 @@ function formatLogRecording(log: string[]): string {
...
   return `\n${'='.repeat(leftLength)}${header}${'='.repeat(rightLength)}\n${log.join('\n')}\n${'='.repeat(headerLength)}`;
 }
+
+function maybeMaskSensitiveData(params: any) {
+  if (!params || !isObject(params))
+    return params;
+  if (!(params as any).secret)
+    return params;
+  const redacted: any = { ...params };
+  for (const name of ['text', 'value']) {
```
dgozman, line 336: Let's make this declarative? Like we have `tracing: snapshot: true` in the `protocol.yml`.

packages/playwright-core/src/server/dispatchers/dispatcher.ts
```diff
@@ -257,7 +257,7 @@ export class DispatcherConnection {
       endTime: 0,
       type: dispatcher._type,
       method,
-      params: params || {},
+      params: maybeMaskSensitiveData(params) || {},
```
dgozman, line 260: Is this the right place? Perhaps better do it in the trace recorder?

packages/playwright-core/src/server/dom.ts
```diff
@@ -582,17 +582,35 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
       progress.log('  element is visible, enabled and editable');
       if (filled === 'needsinput') {
         progress.throwIfAborted();  // Avoid action that has side-effects.
-        if (value)
+        if (value) {
+          if (options.secret)
+            await this._addSecret(value);
```
dgozman, line 587: We should `addSecret` not only in the case `needsinput`, but in all cases.

packages/playwright-core/src/server/dom.ts
```diff
@@ -582,17 +582,35 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
...
           await this._page.keyboard.press('Delete');
+        }
       } else {
         assertDone(filled);
       }
       return 'done';
     }, 'input');
   }
 
+  private async _addSecret(secret: string) {
```
dgozman, line 599: Should this be a part of tracing/snapshotting instead?

packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
```diff
@@ -59,6 +60,12 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
     cssRef?: number, // Previous snapshotNumber for overridden stylesheets.
   };
 
+  function maskIfSecret(value: string) {
+    if ((window as any).__playwright_secrets_?.has(value))
+      return maskString(value);
```
dgozman, line 65: I don't think we can use imported functions in `snapshotterInjected`, because we do not build+bundle it.

packages/playwright-core/src/utils/traceUtils.ts
```diff
@@ -120,6 +120,10 @@ export async function saveTraceFile(fileName: string, traceEvents: ActionTraceEv
   });
 }
 
+export function maskString(str: string) {
+  return '*'.repeat(str.length);
```
dgozman, line 124: We should not leak the secret's length as well:

```suggestion
  return '****';
```

tests/library/tracing.spec.ts
```diff
@@ -584,6 +584,42 @@ test('should store postData for global request', async ({ request, server }, tes
...
+  expect(actions).toContain('page.fill');
+  expect(actions).toContain('page.type');
+  expect(actions).toContain('locator.fill');
+  expect(actions).toContain('locator.type');
+  expect(actions).toContain('elementHandle.fill');
+  expect(actions).toContain('elementHandle.type');
+
+  for (const [name, content] of resources) {
+    if (name.endsWith('.trace'))
+      expect(content.toString()).not.toContain(secret);
```
dgozman, line 620: Let's also check the rendered snapshot, see trace-viewer.spec.

yury-s:
Closing it for now as it is too fragile and doesn't cover all the cases.

PR #21580, opened 2023-03-10
feat(html reporter): render multiple annotations of the same type together

### LINKED ISSUE #21253: [Feature] Objects in custom annotations & html report
**Issue**
Currently only strings are accepted as custom annotations' description. If an object is set as description the html report won't even render. Being able to pass other types than string to annotations would be helpful. Some examples:
* when a single test covers more than one issue, we could set an array of affected issues
* when E2E testing, we could save an object containing the current context

**Proposed solution**
On html reporter, before rendering the annotation description, check if the value is an object and if so stringify it. [Implementation example](https://github.com/microsoft/playwright/compare/main...gbasto:playwright:html-reporter-annotations).

Result:
image

dgozman:
Multiple annotations are rendered as comma-separated lists. Fixes #21253.

IMAGE 2023-03-10 14:24:10

### DIFF
--- a/packages/html-reporter/src/testCaseView.tsx
+++ b/packages/html-reporter/src/testCaseView.tsx
@@ -14,7 +14,7 @@
   limitations under the License.
 */
 
-import type { TestCase, TestCaseAnnotation } from './types';
+import type { TestCase } from './types';
 import * as React from 'react';
 import { TabbedPane } from './tabbedPane';
 import { AutoChip } from './chip';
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

packages/html-reporter/src/testCaseView.tsx
```diff
@@ -32,13 +32,20 @@ export const TestCaseView: React.FC<{
 }> = ({ projectNames, test, run, anchor }) => {
   const [selectedResultIndex, setSelectedResultIndex] = React.useState(run);
 
+  const annotations = new Map<string, (string | undefined)[]>();
+  test?.annotations.forEach(annotation => {
+    const list = annotations.get(annotation.type) || [];
+    list.push(annotation.description);
```
gbasto, line 38: Might be worth to force string here, otherwise if an object is passed by mistake as description, the html report won't render with no indication of what the problem is.

```javascript
    list.push(new String(annotation.description));
```

PR #21590, opened 2023-03-11
chore: hide timeline bars and stack from ui mode

pavelfeldman:


packages/trace-viewer/src/ui/timeline.tsx
```diff
@@ -50,6 +51,10 @@ export const Timeline: React.FunctionComponent<{
 
   const { boundaries, offsets } = React.useMemo(() => {
     const boundaries = { minimum: model?.startTime || 0, maximum: model?.endTime || 30000 };
+    if (boundaries.minimum > boundaries.maximum) {
```
dgozman, line 54: How is this possible?

packages/trace-viewer/src/ui/timeline.tsx
```diff
@@ -95,41 +100,34 @@ export const Timeline: React.FunctionComponent<{
...
       const xMiddle = (bar.leftTime + bar.rightTime) / 2;
       const xd = Math.abs(time - xMiddle);
-      const yd = Math.abs(y - yMiddle);
       if (left > right)
         continue;
       // Prefer closest yDistance (the same bar), among those prefer the closest xDistance.
-      if (index === undefined ||
-          (yd < yDistance!) ||
-          (Math.abs(yd - yDistance!) < 1e-2 && xd < xDistance!)) {
+      if (index === undefined || xd < xDistance!) {
```
dgozman, line 118: - Why this change?
- The comment does not match anymore.

pavelfeldman: I don't think users could use it / were using it.

packages/trace-viewer/src/ui/timeline.tsx
pavelfeldman: Initial values min is maxint, max is minint.

PR #21631, opened 2023-03-13
fix(reuse): only reuse the default browser, not explicitly launched ones

### LINKED ISSUE #20399: [BUG] Test execution gets stuck when running through Playwright extension in Vscode
**Context:**
- Playwright Version: 1.30.0
- Playwright Test for VSCode 1.0.6
- Operating System: Mac
- Node.js version: v16.15.0
- Browser: Chromium

**Code Snippet**

Test:

```javascript
import { test, chromium } from '@playwright/test';
import path from 'path';

test('Reproduce defect', async function ({ page }) {
  await page.goto('https://google.com');

  const browser = await chromium.launch({
    headless: false,
    args: [
      '--use-fake-device-for-media-stream',
      '--use-fake-ui-for-media-stream'
    ],
  });
  const pageNewContext = await context.newPage();

  await pageNewContext.goto('https://www.microsoft.com');
});
```

Playwright config:
```javascript
import { type PlaywrightTestConfig } from '@playwright/test';
import dotenv from 'dotenv';

dotenv.config();

  name: 'tests',
  use: {
    ignoreHTTPSErrors: true,
    acceptDownloads: true,
    headless: true,
    viewport: { width: 1280, height: 720 },
  },
  testDir: 'test/tests/',
  reporter: [['list'], ['html']],
  reportSlowTests: null,
  timeout: 200_000,
};

export default config;

```

**Describe the bug**

When I run this test through Playwright extension in Vscode test execution gets stuck on await chromium.launch(). 
Screenshot 2023-01-26 at 17 28 18

But when I run it through: **_playwright test --grep 'Reproduce defect'_** the test works as expected and opens new chromium inside the test

dgozman:
Fixes #20399.

### DIFF
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -17,6 +17,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import type { APIRequestContext, BrowserContext, BrowserContextOptions, LaunchOptions, Page, Tracing, Video } from 'playwright-core';
+import type { BrowserType as BrowserTypeImpl } from 'playwright-core/lib/client/browserType';
 import * as playwrightLibrary from 'playwright-core';
 import { createGuid, debugMode, removeFolders, addInternalStackPrefix, mergeTraceFiles, saveTraceFile } from 'playwright-core/lib/utils';
 import type { Fixtures, PlaywrightTestArgs, PlaywrightTestOptions, PlaywrightWorkerArgs, PlaywrightWorkerOptions, ScreenshotMode, TestInfo, TestType, TraceMode, VideoMode } from '../types/test';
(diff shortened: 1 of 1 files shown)


packages/playwright-test/src/index.ts
```diff
@@ -109,20 +104,31 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
...
+    if (connectOptions && process.env.PW_TEST_REUSE_CONTEXT) {
+      browser = await (playwright[browserName] as BrowserTypeImpl)._connectInsteadOfLaunching({
+        ...connectOptions,
+        headers: {
+          ...connectOptions.headers,
+          'x-playwright-reuse-context': '1',
+        },
+      }, {});
+    } else {
+      browser = await playwright[browserName].launch();
```
pavelfeldman, line 130: I'm not sure I follow what is happening here since launch also goes into connectInsteadOfLaunching when we have connect options. Looks like you are building a mode where only fixture's browser is using the connection. If seems like not setting `_defaultConnectOptions` in that case would achieve that.

pavelfeldman (approved):

PR #21638, opened 2023-03-13
fix(shards): shard test groups instead of files

### LINKED ISSUE #21497: [BUG] Sharding is not dividing the tests number equally for each shard
### System info
- Playwright Version: [v1.31.2]

**Steps**
Let's suppose we have 10 test files - 

Test File 1 - 2 tests
Test File 2 - 10 tests
Test File 3 - 10 tests
Test File 4 - 1 tests
Test File 5 - 1 tests
Test File 6 - 2 tests
Test File 7 - 1 tests
Test File 8 - 1 tests
Test File 9 - 1 tests
Test File 10 - 1 tests

Now if we are executing the tests using 3 / 4 shards, ~equal number tests are not getting executed for each shard and it's causing a huge problem for us as we are having 1000+ cases across ~180 files. We are currently using 6 shards, but we see huge number of tests difference for each shard when using latest playwright 1.31.2.

Observed the comment from #21226 that, with new playwright version sharding will be performed on test files.

But I believe, it may cause a big problem for existing projects / new projects, where some feature may have more number of tests than other features. And every time, it will be very difficult for maintaining the count of tests in the test files from user side and instead I believe it should be automatically handled by the playwright framework itself as maintaining / re-factoring tests numbers for each test file will be a big challenge, where multiple users will be working on the repository.

**Expected**

Playwright sharding should divide the tests ~equally for each shard by the playwright framework itself and user don't need to maintain the framework seperately.

**Actual**

Playwright sharding is not dividing the tests ~equally for each shard when using 1.31.2.

dgozman:
Fixes #21497.

### DIFF
--- a/packages/playwright-test/src/runner/tasks.ts
+++ b/packages/playwright-test/src/runner/tasks.ts
@@ -22,27 +22,21 @@ import { Dispatcher } from './dispatcher';
 import type { TestRunnerPluginRegistration } from '../plugins';
 import type { Multiplexer } from '../reporters/multiplexer';
 import type { TestGroup } from '../runner/testGroups';
-import { createTestGroups } from '../runner/testGroups';
 import type { Task } from './taskRunner';
 import { TaskRunner } from './taskRunner';
 import type { Suite } from '../common/test';
 import type { FullConfigInternal, FullProjectInternal } from '../common/types';
-import { loadAllTests, loadGlobalHook } from './loadUtils';
+import { loadAllTests, loadGlobalHook, type ProjectWithTestGroups } from './loadUtils';
 import type { Matcher } from '../util';
 
 const removeFolderAsync = promisify(rimraf);
 const readDirAsync = promisify(fs.readdir);
 
-type ProjectWithTestGroups = {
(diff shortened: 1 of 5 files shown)


packages/playwright-test/src/runner/tasks.ts
```diff
@@ -22,27 +22,21 @@ import { Dispatcher } from './dispatcher';
 import type { TestRunnerPluginRegistration } from '../plugins';
 import type { Multiplexer } from '../reporters/multiplexer';
 import type { TestGroup } from '../runner/testGroups';
-import { createTestGroups } from '../runner/testGroups';
 import type { Task } from './taskRunner';
 import { TaskRunner } from './taskRunner';
 import type { Suite } from '../common/test';
 import type { FullConfigInternal, FullProjectInternal } from '../common/types';
-import { loadAllTests, loadGlobalHook } from './loadUtils';
+import { loadAllTests, loadGlobalHook, type ProjectWithTestGroups } from './loadUtils';
```
pavelfeldman, line 29: lint error

pavelfeldman (approved):

Issue #21566, opened 2023-03-10
[BUG] Test fixtures do not support args which are not destructured or destructured args which use object spread


enagy27:
### System info
- Playwright Version: [v1.31.2]
- Operating System: [macOS Ventura 13.2.1]
- Browser: [All]

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**The following fixture code works**
The comment in this file provides some helpful context for some workarounds that I have tried which have failed
```ts
import { test as base } from "@playwright/test"

import { page as browserStackPage } from "./browserStackFixture"

export const test = base.extend({
	/**
	 * SUPER IMPORTANT NOTE DO NOT IGNORE!
	 *
	 * It looks like VSCode/Playwright are doing something meta and inspecting
	 * these fixture arguments potentially. That's causing some interesting
	 * issues which are outlined below.
	 *
	 * Our goal with this function is to choose the right fixture and forward
	 * all of the necessary arguments. Unfortunately we must destructure the args
	 * object or else VSCode's test runner will break and we cannot run tests.
	 * Not only must we destructure, but `page` must be in the destructure and we
	 * cannot use object spreads.
	 *
	 * Things we have tried:
	 * - Spreading the args: `{ page, ...args }`
	 * - Using `args` and destructuring from that object: `const { page } = args`
	 * - using `arguments` on a `function` while destructuring in the declaration.
	 *   This works but will only pick up the arguments which have been
	 *   destructured, which defeats the point. Going this route makes debugging
	 *   even more difficult because there is no type-safety indicating the missing
	 *   data
	 */
	page: async function ({ page, browserName }, use, testInfo) {
		const { project } = testInfo
		if (project.name.endsWith("@browserstack")) {
			// Read the JSDoc above to learn why we are only passing these and not
			// using object spreading
			await browserStackPage({ page, browserName }, use, testInfo)
			return
		}

		await use(page)
	},
})
```

**Config file**

```ts
import type {
	PlaywrightTestConfig,
	Locator,
	Project,
	PlaywrightTestOptions,
	PlaywrightWorkerOptions,
} from "@playwright/test"
import { devices, expect } from "@playwright/test"
import dotenv from "dotenv"

/**
 * Read environment variables from file.
 * https://github.com/motdotla/dotenv
 */
dotenv.config()

const defaultSnapshotPathTemplate =
[...]

PR #21659, opened 2023-03-14
chore: throw pretty error if rest parameters is used inside fixtures

mxschmitt:
https://github.com/microsoft/playwright/issues/21566

### DIFF
--- a/packages/playwright-test/src/common/fixtures.ts
+++ b/packages/playwright-test/src/common/fixtures.ts
@@ -239,15 +239,15 @@ function innerFixtureParameterNames(fn: Function, location: Location, onError: L
     onError({ message: 'First argument must use the object destructuring pattern: '  + firstParam, location });
     return [];
   }
-  const firstParamWithoutBrackets = firstParam.substring(1, firstParam.length - 1);
-  if (firstParamWithoutBrackets.includes('...')) {
-    onError({ message: 'Fixtures cannot use rest parameters: ' + firstParamWithoutBrackets.trim(), location });
-    return [];
-  }
-  const props = splitByComma(firstParamWithoutBrackets).map(prop => {
+  const props = splitByComma(firstParam.substring(1, firstParam.length - 1)).map(prop => {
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/common/fixtures.ts
```diff
@@ -239,7 +239,12 @@ function innerFixtureParameterNames(fn: Function, location: Location, onError: L
     onError({ message: 'First argument must use the object destructuring pattern: '  + firstParam, location });
     return [];
   }
-  const props = splitByComma(firstParam.substring(1, firstParam.length - 1)).map(prop => {
+  const firstParamWithoutBrackets = firstParam.substring(1, firstParam.length - 1);
+  if (firstParamWithoutBrackets.includes('...')) {
+    onError({ message: 'Fixtures cannot use rest parameters: ' + firstParamWithoutBrackets.trim(), location });
```
dgozman, line 244: It is called "rest property": https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#rest_property.

packages/playwright-test/src/common/fixtures.ts
```diff
@@ -239,7 +239,12 @@ function innerFixtureParameterNames(fn: Function, location: Location, onError: L
     onError({ message: 'First argument must use the object destructuring pattern: '  + firstParam, location });
     return [];
   }
-  const props = splitByComma(firstParam.substring(1, firstParam.length - 1)).map(prop => {
+  const firstParamWithoutBrackets = firstParam.substring(1, firstParam.length - 1);
+  if (firstParamWithoutBrackets.includes('...')) {
```
dgozman, line 243: Can we check `props` at the line 251, to be more specific in the error message: `Rest property "...foo" is not supported. List all used fixtures explicitly, separated by comma.`

dgozman (approved):

enagy27 (approved):

PR #21669, opened 2023-03-14
fix(routeFromHar): prefer valid responses

NoamGaash:
In case:
a. few HAR entries are valid routing match candidates and
b. some of them have no response
I suggest we should always prefer to route to the entry with a valid response over the entries having exceptions.

It will solve a problem that we have with unstable services - occasionally when requests are failing we make retries, and we need a way to prefer the successful try over the failing ones.

### DIFF
--- a/packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
+++ b/packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
@@ -306,9 +306,8 @@ class HarBackend {
           const matchingHeaders = countMatchingHeaders(candidate.request.headers, headers);
           list.push({ candidate, matchingHeaders });
         }
+        list.reverse();
         list.sort((a, b) => b.matchingHeaders - a.matchingHeaders);
-        while (list[0].candidate.time < 0 && list.some(entry => entry.candidate.time >= 0))
-          list.shift();
         entry = list[0].candidate;
       }
 
--- a/packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
+++ b/packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
@@ -307,6 +307,8 @@ class HarBackend {
           list.push({ candidate, matchingHeaders });
         }
         list.sort((a, b) => b.matchingHeaders - a.matchingHeaders);
+        while (list[0].candidate.time < 0 && list.some(entry => entry.candidate.time >= 0))
+          list.shift();
         entry = list[0].candidate;
       }
 
--- a/tests/assets/har-unstable-network.har
+++ b/tests/assets/har-unstable-network.har
@@ -1,0 +1,105 @@
+{
+  "log": {
+    "version": "1.2",
+    "creator": {
+      "name": "Playwright",
+      "version": "1.32.3"
+    },
+    "browser": {
+      "name": "chromium",
+      "version": "112.0.5615.29"
+    },
+    "entries": [
+      {
+        "startedDateTime": "2022-06-16T21:41:23.951Z",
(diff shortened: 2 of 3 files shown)


mxschmitt:
Can you file an issue first with the feature / bug this PR is fixing? Also having a test which fails before your change and passes after your change would be amazing.

NoamGaash:
sure! on it.
should the test be added inside this PR. or on a separate one?

mxschmitt:
Same PR usually. Also please file an issue first which describes your bug / behaviour change.

NoamGaash:
Thanks for your waiting!
I have added a test and filed issue #22325 

:smiley:

packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
```diff
@@ -307,6 +307,8 @@ class HarBackend {
           list.push({ candidate, matchingHeaders });
         }
         list.sort((a, b) => b.matchingHeaders - a.matchingHeaders);
+        while (list[0].candidate.time < 0 && list.some(entry => entry.candidate.time >= 0))
```
mxschmitt, line 310: Discussed with the team offline: let's remove the filtering logic and always use the last request instead of the first request. Test should continue to pass then.

NoamGaash: just to make sure I understand -
1. our "ideal entry" is last, has a maximal amount of matching headers, and is valid.
2. if we have two entries, let's name those `a` and `b`, where `a` is before `b` and has a smaller amount of matching headers, but `b` is invalid, do we want `b` to be returned?

the use case that I gave as an example is very simple, and the approach you suggested will solve it, but in reality, things can be more complicated - for example, consider chat notifications being periodically polled. if the har recording will end soon after one failing request, retrieving the last result will make the endpoint consistently fail

mxschmitt: As of today we don't return the network requests ordered, aka. we don't have web replay, we just do a best effort in what we return or not. Since we're not going to add web replay capabilities and network request counting in the near future, we came to the conclusion to always return the last one for now until we revisit this again.

Oh and yes if its like that:

request1 5 matching headers
request2 5 matching headers
request3 4 matching headers

we should return request 2, so just returning the last one might not work. We could push the `occurenceIndex` into the list and then adjust our list.sort to reflect the index as well which should work.

NoamGaash: It sounds a little odd, because if request 2 is aborted (for example, due to network condition, or due to navigation), I believe it make sense to ignore it.

also, consider the use case where I have some service being periodically fetched -

request1, 5 matching headers, status 200
request2, 5 matching headers, status 200
request3, 5 matching headers, network unavailable
request4, 5 matching headers, status 200
request5, 5 matching headers, status 200
request6, 5 matching headers, status 200
request7, 5 matching headers, canceled due change in application state

preferring request 6 over 7 makes more sense IMO, but I'll accept any decision

mxschmitt: we decided for now to not filter them. So yes, lets continue with request7 for now. Thanks!

NoamGaash: done :)

NoamGaash:
it seems like I have some failing tests.
converting to draft - will fix it soon

pavelfeldman:
Looks like we dropped the ball on this one, sorry, closing as obsolete.

PR #21696, opened 2023-03-16
chore: split trace events into phases

pavelfeldman:


packages/trace-viewer/src/versions/traceV3.ts
```diff
@@ -97,6 +97,8 @@ export type ResourceOverride = {
 };
 
 export type FrameSnapshot = {
+  // There was no callId in the original, we are intentionally regressing it.
+  callId: string;
```
dgozman, line 101: We can turn `snapshotName` into `callId` for old traces, to preserve the highlight.

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -123,7 +122,7 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
 
     if (this._state) {
       const o = this._state.options;
-      if (!o.screenshots !== !options.screenshots || !o.snapshots !== !options.snapshots)
+      if (o.name !== options.name || !o.screenshots !== !options.screenshots || !o.snapshots !== !options.snapshots)
```
dgozman, line 125: With my change, this is not needed.

dgozman (approved):

packages/trace-viewer/src/versions/traceV3.ts
pavelfeldman: Yes, but given how backwards support worked, I think it is ok.

PR #21729, opened 2023-03-16
feat: Allow for passing --disable-gpu to Chromium

fdgonthier:
Playwright works under Microsoft WSL *except* where Chromium is involved. Running Chromium in WSL involves passing the --disable-gpu argument. Without this, any Chromium window remain black or entirely transparent. This patch adds support for the PWTEST_DISABLE_GPU environment variable, which, if set to any value, automatically passes the argument whenever Chromium is invoked.

### DIFF
--- a/packages/playwright-core/src/server/chromium/chromium.ts
+++ b/packages/playwright-core/src/server/chromium/chromium.ts
@@ -271,6 +271,8 @@ export class Chromium extends BrowserType {
       chromeArguments.push('about:blank');
     else
       chromeArguments.push('--no-startup-window');
+    if (process.env.PWTEST_DISABLE_GPU)
+      chromeArguments.push('--disable-gpu');
     return chromeArguments;
   }
 


mxschmitt:
Why does the following not work for you which you can put into your config?

```
...
    {
      name: 'chromium',
      use: {
        ...devices['Desktop Chrome'],
        launchOptions: {
          args: ['--disable-gpu']
        }
      },
    },
...
```

under the projects section.

fdgonthier:
Will that work for the Recorder/Debugger and the Trace Viewer? We had a solution for the tests themselves (we use Python), by passing the "--disable-gpu" argument when the browser context is created, but that doesn't work for the tools.

mxschmitt:
I saw https://github.com/puppeteer/puppeteer/commit/d1c667ba1e4419191378cca0c844bd476c7aa771 and it looks like you can use the GPU if you install the required dependencies.

Do you have all the Playwright dependencies installed? `npx playwright install-deps` Or in Python: `playwright install-deps`

fdgonthier:
Maybe it's more a specificity of my system than I thought at first as what you linked me makes it sounds like it should work, but I've tried everything suggested, to no avail. The official Chrome .deb starts with a black screen exactly the same way yet works when I pass --disable-gpu. I've seen quite a few people reporting problems with Chrome based tools on WSL. I've seen no resolution. On my side, Gitkraken, another Chrome based app, also requires this flag to start.

fdgonthier:
> Could you please file an issue before sending a PR next time? It is beneficial to have a discussion in the issue rather than in the PR, because later on it is easily searchable and available to everyone having the same problem. PR discussions are much harder to follow and find.

Okay fair I didn't think this 2 line patch would require discussion.

> 
> > The official Chrome .deb starts with a black screen exactly the same way yet works when I pass --disable-gpu.
> 
> This means that Chrome does not work with your WSL configuration. Try following the steps from here: https://learn.microsoft.com/en-us/windows/wsl/tutorials/gui-apps#install-google-chrome-for-linux.
> 
> If that does not work, the solution would be somewhere in Chromium or in WSL. I'd recommend to file a Chromium issue at https://crbug.com/new and see what they say. I don't think Playwright should introduce a workaround for this issue right now. Sorry for the inconvenience!

The problem is that without passing "--disable-gpu" somehow, Playwright tools UI are flat-out not usable for me on my WSL but forcing "--disable-gpu" makes them work. I'm not aware of any workaround. If there was a general environment variable , or any other central configuration that worked for all Chromium-based apps, I would certainly use it. 

I admit this is a specific solution though, so perhaps this environment variable could be turned into a more generic one, to pass command line arguments to Chromium before it starts?

Anyway, I'll file a bug...

fdgonthier:
Thanks to you insisting, I ended up finding a central solution! I swear I had been looking for it a while. I'll file a bug about it still, with the solution. I think it could be beneficial to mention the solution in the Playwright documentation to prevent another idiot from sending the same patch I did.

PR #21737, opened 2023-03-17
chore(ui): store selected projects in settings

pavelfeldman:


dgozman (approved):

packages/playwright-core/src/server/chromium/crApp.ts
```diff
@@ -29,23 +30,21 @@ export async function installAppIcon(page: Page) {
...
-    setInterval(() => {
-      const value = JSON.stringify(localStorage);
-      if (value !== lastValue) {
-        lastValue = value;
-        window.saveSettings(value);
-      }
-    }, 2000);
-  })})(${settings})`);
+  await page.addInitScript(
+      `(${String((settings: any) => {
```
mxschmitt, line 43: how do types work inside a JavaScript script `string`?

PR #21742, opened 2023-03-17
feat: Add clipboard copy to test title in html reporter

thohe973:
When a test fails, you often want to search for it in your code base.
This adds a copy icon to the test title so you can quickly copy it.
copyTitle


packages/html-reporter/src/testCaseView.css
```diff
@@ -35,11 +35,23 @@
...
   padding: 8px;
   font-weight: 400;
-  font-size: 32px !important;
+  font-size: var(--title-size);
   line-height: 1.25 !important;
+  cursor: pointer;
+}
+
+.test-case-title .octicon {
+  width: var(--title-size);
```
pavelfeldman, line 48: It is unusual to use the same values for the width / height and the font size.

packages/html-reporter/src/testCaseView.tsx
```diff
@@ -45,9 +47,18 @@ export const TestCaseView: React.FC<{
     return matchTags(test.title).sort((a, b) => a.localeCompare(b));
   }, [test]);
 
+  const handleCopy = (value: string) => {
+    navigator.clipboard.writeText(value).then(() => {
+      setCopyIcon(icons.check());
+      setTimeout(() => setCopyIcon(icons.copy()), 3000);
```
pavelfeldman, line 53: You probably want to clear this timeout from an effect?

thohe973: Fixed

packages/html-reporter/src/testCaseView.css
thohe973: Yeah I know, but this was the best solution I could figure out to ensure that the icon matches the font size if someone changes it in the future.

mm0hammadi:
this is awsome feature! :+1:

mahtabnejad90 (approved):

packages/html-reporter/src/testCaseView.css
gskierk: These values can surely depend on the title size, but new variables should be added for the icon for consistency.

```css
--title-size: 32px !important;

--octicon-width: var(--title-size);
--octicon-height: var(--title-size);
```

pavelfeldman:
Sorry we dropped the ball on this one. I don't think we want this UX improvement at this point!

PR #21748, opened 2023-03-17
test: do not pollute stderr with trace-viewer wsEndpoint

mxschmitt:


packages/playwright-core/src/server/trace/viewer/traceViewer.ts
```diff
@@ -99,7 +99,7 @@ export async function showTraceViewer(traceUrls: string[], browserName: string,
   await context.extendInjectedScript(consoleApiSource.source);
   const [page] = context.pages();
 
-  if (isUnderTest())
+  if (process.env.PWTEST_PRINT_WS_ENDPOINT)
```
pavelfeldman, line 102: I would still spam sometimes, let me see what we can do to change that.

pavelfeldman: Or does not it spam because we hide process out in ui-mode early?

pavelfeldman: ah, that's because it only spams in browser tests, i get it now.

pavelfeldman (approved):

PR #21769, opened 2023-03-17
chore: fix and test ui mode stop

pavelfeldman:


packages/trace-viewer/src/ui/watchMode.tsx
```diff
@@ -122,6 +122,12 @@ export const WatchModeView: React.FC<{}> = ({
     setProgress({ total: testIds.length, passed: 0, failed: 0, skipped: 0 });
     setRunningState({ testIds: new Set(testIds) });
     sendMessage('run', { testIds }).then(() => {
+      // Clear pending tests in case of interrupt.
+      for (const test of testModel.rootSuite?.allTests() || []) {
+        if (test.results[0]?.duration === -1)
```
dgozman, line 127: For actually skipped tests with `test.skip()`, wouldn't the `duration` be -1 as well? If so, this seems to turn test status from "skipped" into "none".

dgozman (approved):

packages/trace-viewer/src/ui/watchMode.tsx
pavelfeldman: Let me add a test.

PR #21863, opened 2023-03-22
feat: add timeout to `electron.firstWindow()` method

### LINKED ISSUE #21846: [BUG] `timeout: 0` option is not respected as documented
### System info
- Playwright Version: 1.27.1
- Operating System: macOS
- Browser: Electron
- Other info:

We are starting Electron in our CI with this:

```ts
const electron = await playwright._electron.launch({
	executablePath: configuration.electronPath,
	args: configuration.args,
	env: configuration.env as { [key: string]: string },
	timeout: 0
});

const window = await electron.firstWindow();
```

The `timeout: 0` option was only added recently to address timeout errors we would hit where the log shows:

```
[2023-03-21T13:16:29.991Z] Finished operation 'launch playwright (electron)' with error TimeoutError: electronApplication.firstWindow: Timeout 30000ms exceeded while waiting for event "window" after 31298ms
```

So it seems to me `timeout: 0` is not disabling the timeout. We still see a timeout after 30 seconds.

aslushnikov:
Fixes https://github.com/microsoft/playwright/issues/21846

### DIFF
--- a/docs/src/api/class-electronapplication.md
+++ b/docs/src/api/class-electronapplication.md
@@ -135,12 +135,13 @@ Convenience method that waits for the first application window to be opened.
   // ...
 ```
 
-### param: ElectronApplication.firstWindow.options
+### option: ElectronApplication.firstWindow.timeout
 * since: v1.33
-* langs: js
-- `options` ?<[Object]>
-  - `timeout` ?<[float]> maximum time to wait for in milliseconds. Defaults to `30000` (30 seconds). Pass `0` to
-    disable timeout. The default value can be changed by using the [`method: BrowserContext.setDefaultTimeout`].
+- `timeout` ?<[float]>
+
+Maximum time to wait for in milliseconds. Defaults to `30000` (30 seconds).
+Pass `0` to disable timeout. The default value can be changed by using the
+[`method: BrowserContext.setDefaultTimeout`].
 
 ## method: ElectronApplication.process
 * since: v1.21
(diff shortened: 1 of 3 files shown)


docs/src/api/class-electronapplication.md
```diff
@@ -135,6 +135,13 @@ Convenience method that waits for the first application window to be opened.
   // ...
 ```
 
+### param: ElectronApplication.firstWindow.options
+* since: v1.33
+* langs: js
+- `options` ?<[Object]>
+  - `timeout` ?<[float]> maximum time to wait for in milliseconds. Defaults to `30000` (30 seconds). Pass `0` to
+    disable timeout. The default value can be changed by using the [`method: BrowserContext.setDefaultTimeout`].
```
mxschmitt, line 143: BrowserContext.setDefaultTimeout?

docs/src/api/class-electronapplication.md
```diff
@@ -135,6 +135,13 @@ Convenience method that waits for the first application window to be opened.
   // ...
 ```
 
+### param: ElectronApplication.firstWindow.options
```
dgozman, line 138:
```suggestion
### option: ElectronApplication.firstWindow.timeout
```

And change everything below! Copy from other options when in doubt 😄

docs/src/api/class-electronapplication.md
aslushnikov: what about it?

mxschmitt: There is no browsercontext in electron?

aslushnikov: What about https://playwright.dev/docs/next/api/class-electronapplication#electron-application-context ?

mxschmitt: You won

PR #21903, opened 2023-03-23
change codeblocks in issue template to typescript

DetachHead:
the example config says it's `playwright.config.ts` so the codeblock should probably be typescript


pavelfeldman:
We use js codeblocks for consistency.

DetachHead:
@pavelfeldman consistency with what? It seems inconsistent to me to have a comment that says a file is a `.ts` while it's in a `.js` codeblock

PR #21947, opened 2023-03-24
fix: make sure transform cache is world-writable

### LINKED ISSUE #21859: [BUG] permission denied accessing '/tmp/playwright-transform-cache'
### System info
- Playwright Version: 1.28.1
- Operating System: Ubuntu 20
- Browser: All
- Other info:

**Steps**
- Run playwright on a system where multiple users have been using it

**Actual**

```
Running 651 tests using 2 workers

Error: EACCES: permission denied, open '/tmp/playwright-transform-cache/3f/connectionFailurespec_3fcb2d59d7f1ac9abe23f187eeeed5065482bbfa.map'

  651 skipped
  1 error was not a part of any test, see above for details

```

The permissions on /tmp/playwright-transform-cache:

```
drwxrwxrwx 162 BrentBaccala BrentBaccala 4096 Mar 21 23:28 /tmp/playwright-transform-cache/
```

`BrentBaccala` is one of the users on the system.  Permissions on some subdirectories:

```
drwxrwxrwx 2 BrentBaccala    BrentBaccala    4096 Feb 21 12:04 3e
drwxr-xr-x 2 MaximKhlobystov MaximKhlobystov 4096 Mar  1 13:44 3f
drwxrwxrwx 2 BrentBaccala    BrentBaccala    4096 Feb 21 12:04 42
```

`MaximKhlobystov` is also a user on the system, and I'm running this test (that's trying to access `3f`) as user `baccala`, so obviously there's a problem.

It looks like Maxim is also running Playwright 1.28.1.

What explanation is there for how this directory `3f` was created with these restrictive permissions?  It surely happened on a previous playwright run, and I really don't know what the exact test configuration was that caused this.  The directory is dated March 1, and today is March 21, so reproducing exactly what caused this might be problematic (if it isn't obvious to somebody).

aslushnikov:
Fixes https://github.com/microsoft/playwright/issues/21859

### DIFF
--- a/packages/playwright-test/src/common/compilationCache.ts
+++ b/packages/playwright-test/src/common/compilationCache.ts
@@ -78,10 +78,15 @@ export function getFromCompilationCache(filename: string, code: string, moduleUr
 
   return {
     addToCache: (code: string, map: any) => {
-      fs.mkdirSync(path.dirname(cachePath), { recursive: true });
+      // Make sure we create cache folders with 777 permissions.
+      // Based on the stackoveflow answer: https://stackoverflow.com/a/34721366/314883
+      // See https://github.com/microsoft/playwright/issues/21859
+      const oldmask = process.umask(0);
+      fs.mkdirSync(path.dirname(cachePath), { recursive: true, mode: 0o777 });
       if (map)
         fs.writeFileSync(sourceMapPath, JSON.stringify(map), 'utf8');
       fs.writeFileSync(codePath, code, 'utf8');
+      process.umask(oldmask);
       _innerAddToCompilationCache(filename, { codePath, sourceMapPath, moduleUrl });
     }
   };


packages/playwright-test/src/common/compilationCache.ts
```diff
@@ -78,10 +78,15 @@ export function getFromCompilationCache(filename: string, code: string, moduleUr
 
   return {
     addToCache: (code: string, map: any) => {
-      fs.mkdirSync(path.dirname(cachePath), { recursive: true });
+      // Make sure we create cache folders with 777 permissions.
+      // Based on the stackoveflow answer: https://stackoverflow.com/a/34721366/314883
+      // See https://github.com/microsoft/playwright/issues/21859
+      const oldmask = process.umask(0);
```
dgozman, line 84: Why don't we do this once at the start of the test runner, and not only here? We create quite a few directories.

packages/playwright-test/src/common/compilationCache.ts
```diff
@@ -78,10 +78,15 @@ export function getFromCompilationCache(filename: string, code: string, moduleUr
...
+      fs.mkdirSync(path.dirname(cachePath), { recursive: true, mode: 0o777 });
```
dgozman, line 85: Perhaps a helper that we will use everywhere?

mxschmitt:
I'm worried that this makes it possible that user1 can modify the transpiled code of user2.

aslushnikov:
> I'm worried that this makes it possible that user1 can modify the transpiled code of user2.

 oh year, we decided against this and use different folders, prefixed with username. PR is coming soon!

PR #21972, opened 2023-03-24
chore: pass config to multiplexer

yury-s:


pavelfeldman:
Why is this changing?

PR #21978, opened 2023-03-25
feat(webkit): roll to r1817

playwrightmachine:


mxschmitt:
macos-11 seems failing even after a restart: [macos-11 (webkit)](https://github.com/microsoft/playwright/actions/runs/4518182759/jobs/7957880542#logs)

PR #21980, opened 2023-03-25
refactor(routeFromHar): content to updateContent

NoamGaash:
During PR #21764, the behavior of `routeFromHar` changed.

before #21764 the following code:
```ts
page.routeFromHAR(
        `/networks_cache/file.har`,
        {
            url: /api/,
            content: "embed",
            update: true
        }
    );
```
was used to create a single file.
now, to achieve the same behavior, the user has to do the following:
```ts
page.routeFromHAR(
        `/networks_cache/file.har`,
        {
            url: /api/,
            updateContent: "embed",
            update: true
        } as any
    );
```

by finishing the incomplete refactor, `updateContent: "embed"` will be functional again.

### DIFF
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -624,8 +624,8 @@ Logger sink for Playwright logging.
 - `recordHar` <[Object]>
   - `omitContent` ?<[boolean]> Optional setting to control whether to omit request content from the HAR. Defaults to
     `false`. Deprecated, use `content` policy instead.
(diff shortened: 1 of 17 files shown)


pavelfeldman:
Could you point to a broken use case or add a failing test so that we could better understand the problem?

yury-s:
Closing this PR as part of housekeeping, it has an unresolved problem pointed out above. Feel free to send another PR with the concern addressed.

PR #21985, opened 2023-03-26
feat: Adding trace option 'on-all-retries'

### LINKED ISSUE #21860: [Feature] add a trace option in the playwright conf to trace on all retries
Hi!
Today we use the option 'retain-on-failure' to get traces on all of our retries (actually all of the failing runs) but this also mean that we are tracing the whole time and this consumes resources. We run our tests in a pipeline with limited resources. In these cases it would be very handy with an option that only enables retries on retries but also traces in all the retries. 'on-first-retry' and 'retry-with-trace' only do tracing in the first retry. It would be ideal if 'retry-with-trace' did trace all the retries. Also since it's not mentioned here:
https://playwright.dev/docs/test-configuration#record-test-trace
I feel I can't file a bug for it.

I hope to see this anytime soon.

Cheers!
// Marcus

marcusNumminen:
Fixes: #21860

### DIFF
--- a/docs/src/trace-viewer.md
+++ b/docs/src/trace-viewer.md
@@ -168,7 +168,7 @@ await context.tracing.stop({ path: 'trace.zip' });
 
 Available options to record a trace:
 - `'on-first-retry'` - Record a trace only when retrying a test for the first time.
-- `'on-all-retries'` - Record traces only when retrying for all retries.
+- `'on-all-retries'` - Record traces for all test retries.
 - `'off'` - Do not record a trace.
 - `'on'` - Record a trace for each test. (not recommended as it's performance heavy)
 - `'retain-on-failure'` - Record a trace for each test, but remove it from successful test runs.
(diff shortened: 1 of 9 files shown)


marcusNumminen:

marcusNumminen:
Ok, the test ui-mode-test-progress.spec.ts runs successfully green when I run it locally....

mxschmitt:
> Ok, the test ui-mode-test-progress.spec.ts runs successfully green when I run it locally....

This one is an unrelated flake, I re-triggered the bots, and now they are green.

mxschmitt (approved):

marcusNumminen:
Ok. I can't merge the PR due to 'Required status check "license/cla" was not set by the expected GitHub app.' But Under checks it states "All CLA requirements met". See:
image
What should I do?

docs/src/trace-viewer.md
```diff
@@ -168,6 +168,7 @@ await context.tracing.stop({ path: 'trace.zip' });
 
 Available options to record a trace:
 - `'on-first-retry'` - Record a trace only when retrying a test for the first time.
+- `'on-all-retries'` - Record traces only when retrying for all retries.
```
pavelfeldman, line 171:
```suggestion
- `'on-all-retries'` - Record traces for all test retries.
```

pavelfeldman (approved):

PR #21991, opened 2023-03-27
chore: add snippet to the error message

yury-s:


packages/playwright-test/src/reporters/base.ts
```diff
@@ -409,15 +422,14 @@ export function formatError(config: FullConfig, error: TestError, highlightCode:
       // Failed to read the source file - that's ok.
     }
   }
+
+  error.message = tokens.join('\n');
```
dgozman, line 426: I think we should also remove the name prefix `Error:` because by default message does not contain it.

packages/playwright-test/src/reporters/base.ts
```diff
@@ -409,15 +422,14 @@ export function formatError(config: FullConfig, error: TestError, highlightCode:
...
   if (parsedStack) {
     tokens.push('');
```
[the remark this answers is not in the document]
yury-s: Moved to formatError

packages/playwright-test/src/reporters/base.ts
```diff
@@ -390,36 +390,52 @@ export function formatError(config: FullConfig, error: TestError, highlightCode:
...
+
   return {
     location,
     message: tokens.join('\n'),
   };
 }
 
+export function addSnippetToError(config: FullConfig, error: TestError, file?: string) {
+  let location = error.location;
+  if (error.stack)
```
dgozman, line 418:
```suggestion
  if (error.stack && !location)
```

packages/playwright-test/types/testReporter.d.ts
```diff
@@ -584,6 +584,11 @@ export interface TestError {
    */
   message?: string;
 
+  /**
+   * Source code snippet with highlighted error.
+   */
+  snippet?: string;
```
dgozman, line 590: Let's use this snippet for raw/html right away, to make sure we like it.

tests/playwright-test/reporter.spec.ts
```diff
@@ -635,3 +638,89 @@ test('parallelIndex is presented in onTestEnd', async ({ runInlineTest }) => {
 
   expect(result.output).toContain('parallelIndex: 0');
 });
+
+test('test and step error should have code snippet', async ({ runInlineTest }) => {
+  const testErrorFile = test.info().outputPath('testError.txt');
+  const stepErrorFile = test.info().outputPath('testError.txt');
```
dgozman, line 644:
```suggestion
  const stepErrorFile = test.info().outputPath('stepError.txt');
```

dgozman (approved):

packages/playwright-test/types/testReporter.d.ts
yury-s: As discussed offline, the only part where we read files to create snippets in raw reporter is for steps snippets and the plans is to ditch step snippets altogether in the merged reports.

PR #21996, opened 2023-03-27
feat(webkit): roll to r1818, freeze webkit version on macos 11

playwrightmachine:
https://github.com/WebKit/WebKit/commit/1267923bb07eea806df5bf31091c7f6e8d08834b broke the WebSocket functionality on macOS 11. They don't support macOS 11 on trunk anymore, hence we freeze it for now.

### DIFF
--- a/packages/playwright-core/browsers.json
+++ b/packages/playwright-core/browsers.json
@@ -33,7 +33,7 @@
     },
     {
       "name": "webkit",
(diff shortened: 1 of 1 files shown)


yury-s (approved):
Lgtm, provided bots pass. Let's also add to the description that we freeze webkit for macos 11 so that's it's easier to search in the logs.

PR #22000, opened 2023-03-27
feat: Locator.parent()

dgozman:


docs/src/api/class-locator.md
```diff
@@ -1546,6 +1546,14 @@ Alternative locator to match.
 
 A page this locator belongs to.
 
+
+## method: Locator.parent
```
pavelfeldman, line 1550: I don't think we should have a raw parent w/o selector / locator.

PR #22005, opened 2023-03-27
fix(ct): vue revert json object as prop

sand4rt:
related to: https://github.com/microsoft/playwright/pull/21400#issuecomment-1485698616.

### DIFF
--- a/packages/playwright-ct-vue/index.d.ts
+++ b/packages/playwright-ct-vue/index.d.ts
@@ -41,14 +41,17 @@ type JsonObject = { [Key in string]?: JsonValue };
 
 type Slot = string | string[];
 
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

pavelfeldman:
@sand4rt do we want it in 1.32.2?

sand4rt:
Would be nice, if it is not to much of a hassle.

It did not effect to many users tho. Only the ones that where using interfaces in combination with separate variables as props:

```Typescript
interface Props { test: string }

const props: Props = { test: 'boop' }

test('test', async ({ mount }) => {
  await mount(Component, { props }); // throws type error, props is not a JsonObject
});
```

Issue #22003, opened 2023-03-27
[Feature] Make Vue component mount props less restrictive


dospunk:
Moving discussion from #21400 

right now, the `mount` fixture for Vue components only allows a props object of type JsonObject to be passed. This does not match the actual behavior of Vue, which allows anything to be passed as props. The current typing prevents passing functions, objects containing functions (for example, instances of a class), BigInts, Symbols, or objects containing any of those from being passed to the Vue component, as well as objects with non-string keys.

sand4rt:
related to: https://github.com/microsoft/playwright/issues/18346, https://github.com/microsoft/playwright/issues/15620, https://github.com/microsoft/playwright/issues/19151, https://github.com/microsoft/playwright/issues/19320

pavelfeldman:
@sand4rt could you help me understand what the action items are for this issue?

sand4rt:
@dospunk correct me if i'm wrong, but i think this one can be closed.

The prop ts type is now less strict and there is type checking at runtime.

The props still need to be JSON serializable, but that's all what's needed in almost all cases. There are already outstanding issues for the cases where non-JSON serializable props are desired.

PR #22025, opened 2023-03-28
chore(ct): throw error when props are not json serializable

sand4rt:
Related to: https://github.com/microsoft/playwright/pull/21400, https://github.com/microsoft/playwright/issues/22003

### DIFF
--- a/packages/playwright-test/src/mount.ts
+++ b/packages/playwright-test/src/mount.ts
@@ -82,12 +82,12 @@ const jsonType: Record<string, Function> = {
   array: (value: any) => Array.isArray(value) && value.every(isJson),
   object: (value: any) => typeof value === 'object' && value !== null && !Array.isArray(value)
     && Object.values(value).every(isJson),
-}
+};
 
 function isJson(value: any): boolean {
   const valueType = Array.isArray(value) ? 'array' : typeof value;
   const validate = jsonType[valueType];
-  if (validate) 
+  if (validate)
     return validate(value);
   return false;
 }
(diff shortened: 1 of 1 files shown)


packages/playwright-test/src/mount.ts
```diff
@@ -72,6 +74,24 @@ export const fixtures: Fixtures<
     },
   };
 
+const jsonType: Record<string, Function> = {
+  string: (value: any) => typeof value === 'string',
+  number: (value: any) => typeof value === 'number',
+  boolean: (value: any) => typeof value === 'boolean',
+  null: (value: any) => value === null,
+  array: (value: any) => Array.isArray(value) && value.every(isJson),
+  object: (value: any) => typeof value === 'object' && value !== null && !Array.isArray(value)
```
pavelfeldman, line 83: JSON will have a default serializing conventions for types like Date, but I guess it is Ok to be more restrictive here.

pavelfeldman (approved):
Fix those nits and we are good.

sand4rt:
> Fix those nits and we are good.

Done!

sand4rt:
> Fix those nits and we are good.

Done, also added the check for `await component.update({ props: { boop: 'boop' }})` : https://github.com/microsoft/playwright/pull/22025/files#diff-f6515604ba6e73e5ba509d2d713faacc6e09f0a06b1db6ca30d0ca7c6ffb4e55R68

packages/playwright-test/src/mount.ts
pavelfeldman: Look at what `JSON.stringify({ data: new Date() })` produces, but don't bother.

pavelfeldman (approved):

PR #22033, opened 2023-03-28
feat: BrowserContext.on('dialog')

dgozman:
Dialogs created early during page initialization are only reported on the context, with `page()` being `null`.

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -154,6 +154,42 @@ context.Console += async (_, msg) =>
 await page.EvaluateAsync("console.log('hello', 5, { foo: 'bar' })");
 ```
 
+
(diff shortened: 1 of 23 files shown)


pavelfeldman (approved):

docs/src/api/class-browsercontext.md
```diff
@@ -154,6 +154,42 @@ context.Console += async (_, msg) =>
...
+  dialog.accept();
+});
+```
+
+```python
+context.on("dialog", lambda dialog: dialog.accept())
+```
+
+```csharp
+context.RequestFailed += (_, request) =>
```
campersau, line 183: This does not look correct, right?

mxschmitt: added to https://github.com/microsoft/playwright/pull/23140

PR #22035, opened 2023-03-28
fix(plugins): follow redirects when checking the webServer url

### LINKED ISSUE #22144: [BUG] WebServer only starting after timeout
### System info
- Playwright Version: v1.32.1
- Operating System: macOS 13.2
- Browser: Chromium
- Other info:

### Source code

https://github.com/pdyck/short-link-management/blob/main/playwright.config.ts

**Config file**

```js
// playwright.config.ts
import { defineConfig, devices } from "@playwright/test";

const baseURL = "http://localhost:3000";

export default defineConfig({
  timeout: 30 * 1000,
  testDir: "./tests",
  testMatch: /.*\.e2e\.tsx?/,
  retries: 2,

  webServer: {
    command: "yarn start",
    timeout: 10 * 1000,
    url: baseURL,
    reuseExistingServer: false,
  },

  use: {
    baseURL,
    trace: "retry-with-trace",
  },

  projects: [
    {
      name: "chromium",
      use: { ...devices["Desktop Chrome"] },
    },
  ],
});

```

```js

test("has title", async ({ page }) => {
    await page.goto("/");

    await expect(page).toHaveTitle(/Test/);
});
```

**Steps**
- `yarn playwright test`

**Expected**

The web server starts and the tests run.

**Actual**

Playwright pings the `baseURL` until timeout and only starts the web server after that. Tested it with timeouts from 10s to 120s.

```
$ DEBUG=pw:webserver yarn playwright test
yarn run v1.22.19
$ .../node_modules/.bin/playwright test
  pw:webserver HTTP GET: http://localhost:3000/ +0ms
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:3000 +309ms
  pw:webserver Starting WebServer process yarn start... +0ms
  pw:webserver Process started +16ms
  pw:webserver Waiting for availability... +0ms
  pw:webserver HTTP GET: http://localhost:3000/ +0ms
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:3000 +309ms
  pw:webserver Waiting 100ms +0ms
  pw:webserver HTTP GET: http://localhost:3000/ +101ms
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:80 +967ms
  pw:webserver Waiting 250ms +0ms
  pw:webserver HTTP GET: http://localhost:3000/ +253ms
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:80 +616ms
  pw:webserver Waiting 500ms +1ms
  pw:webserver HTTP GET: http://localhost:3000/ +501ms
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:80 +617ms
  pw:webserver Waiting 1000ms +0ms
  pw:webserver HTTP GET: http://localhost:3000/ +1s
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:80 +615ms
  pw:webserver Waiting 1000ms +0ms
  pw:webserver HTTP GET: http://localhost:3000/ +1s
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:80 +615ms
  pw:webserver Waiting 1000ms +0ms
  pw:webserver HTTP GET: http://localhost:3000/ +1s
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:80 +618ms
  pw:webserver Waiting 1000ms +0ms
  pw:webserver HTTP GET: http://localhost:3000/ +1s
  pw:webserver Error while checking if http://localhost:3000/ is available: connect ECONNREFUSED ::1:80 +615ms
  pw:webserver Waiting 1000ms +0ms
[WebServer] $ next start
[WebServer] ready - started server on 0.0.0.0:3000, url: http://localhost:3000
[WebServer] info  - Loaded env from .../.env.local
Error: Timed out waiting 10000ms from config.webServer.
``` 

#16834

ghost91-:
Fixes https://github.com/microsoft/playwright/issues/22144

This change implements following redirects when a 3xx HTTP status code is received, when probing the `url` of a `webServer` config.

Without this change (i.e., currently), playwright still _tries_ to follow the redirect, but often fails, since it incorrectly uses the received `Location` header as new URL, instead of combining it with the URL from the previous request.

For example, if the URL is `http://localhost:8080/some-path` and a redirect with `Location: /some-other-path` is received, playwright incorrectly tries to query `http://localhost:80/some-other-path`, which of course fails, due to the incorrect port.

An alternative solution would be to not follow redirects at all when probing the URL, and just consider a 3xx status code as "server is available". This is actually what the documentation claims will happen right now, but it's incorrect. I chose this solution over not following the redirect, since it makes sense, in my opinion, and not following the redirect would require more code changes (`httpRequest` would need some kind of option that prevents it from following redirects). But if that solution would be preferred, I am happy to change the implementation.

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -595,7 +595,7 @@ export default defineConfig({
 - type: ?<[Object]|[Array]<[Object]>>
   - `command` <[string]> Shell command to start. For example `npm run start`..
   - `port` ?<[int]> The port that your http server is expected to appear on. It does wait until it accepts connections. Exactly one of `port` or `url` is required.
-  - `url` ?<[string]> The url on your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the server is ready to accept connections. Exactly one of `port` or `url` is required.
+  - `url` ?<[string]> The url on your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the server is ready to accept connections. Redirects (3xx status codes) are being followed and the new location is checked. Exactly one of `port` or `url` is required.
   - `ignoreHTTPSErrors` ?<[boolean]> Whether to ignore HTTPS errors when fetching the `url`. Defaults to `false`.
   - `timeout` ?<[int]> How long to wait for the process to start up and be available in milliseconds. Defaults to 60000.
   - `reuseExistingServer` ?<[boolean]> If true, it will re-use an existing server on the `port` or `url` when available. If no server is running on that `port` or `url`, it will run the command to start a new server. If `false`, it will throw if an existing process is listening on the `port` or `url`. This should be commonly set to `!process.env.CI` to allow the local dev server when running tests locally.
--- a/packages/playwright-core/src/utils/network.ts
+++ b/packages/playwright-core/src/utils/network.ts
@@ -71,7 +71,7 @@ export function httpRequest(params: HTTPRequestParams, onResponse: (r: http.Inco
   const requestCallback = (res: http.IncomingMessage) => {
     const statusCode = res.statusCode || 0;
     if (statusCode >= 300 && statusCode < 400 && res.headers.location)
-      httpRequest({ ...params, url: res.headers.location }, onResponse, onError);
+      httpRequest({ ...params, url: new URL.URL(res.headers.location, params.url).toString() }, onResponse, onError);
     else
       onResponse(res);
   };
--- a/tests/playwright-test/web-server.spec.ts
+++ b/tests/playwright-test/web-server.spec.ts
@@ -457,6 +457,32 @@ test('should send Accept header', async ({ runInlineTest, server }) => {
   expect(acceptHeader).toBe('*/*');
 });
 
+test('should follow redirects', async ({ runInlineTest, server }) => {
+  server.setRedirect('/redirect', '/redirected-to');
+  server.setRoute('/redirected-to', (req, res) => {
+    res.end('<html><body>hello</body></html>');
+  });
+  const result = await runInlineTest({
+    'test.spec.ts': `
+      import { test, expect } from '@playwright/test';
+      test('connect to the server', async ({baseURL, page}) => {
+        await page.goto('http://localhost:${server.PORT}/redirect');
+        expect(await page.textContent('body')).toBe('hello');
+      });
+    `,
+    'playwright.config.ts': `
+      module.exports = {
+        webServer: {
+          command: 'node ${JSON.stringify(SIMPLE_SERVER_PATH)} ${server.PORT}',
+          url: 'http://localhost:${server.PORT}/redirect',
+          reuseExistingServer: true,
+        }
+      };
+    `,
+  });
+  expect(result.exitCode).toBe(0);
+});
+
 test('should create multiple servers', async ({ runInlineTest }, { workerIndex }) => {
   const port = workerIndex * 2 + 10500;
   const result = await runInlineTest({
--- a/packages/playwright-test/types/test.d.ts
+++ b/packages/playwright-test/types/test.d.ts
@@ -5891,7 +5891,8 @@ interface TestConfigWebServer {
 
   /**
    * The url on your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the
-   * server is ready to accept connections. Exactly one of `port` or `url` is required.
+   * server is ready to accept connections. Redirects (3xx status codes) are being followed and the new location is
+   * checked. Exactly one of `port` or `url` is required.
    */
   url?: string;
 


tests/playwright-test/web-server.spec.ts
```diff
@@ -457,6 +457,32 @@ test('should send Accept header', async ({ runInlineTest, server }) => {
...
+      import { test, expect } from '@playwright/test';
+      test('connect to the server', async ({baseURL, page}) => {
+        await page.goto('http://localhost:${server.PORT}/redirect');
+        expect(await page.textContent('body')).toBe('hello');
+      });
+    `,
+    'playwright.config.ts': `
+      module.exports = {
+        webServer: {
+          command: 'node ${JSON.stringify(SIMPLE_SERVER_PATH)} ${server.PORT}',
```
mxschmitt, line 476:
```suggestion
          command: 'exit 1',
```
I think we can do this to make clear that we should never run the command, since reuseExistingServer is true.

mxschmitt (approved):

tests/playwright-test/web-server.spec.ts
ghost91-: This could also be done for some of the other tests, e.g., `should send Accept header` or `should be able to use an existing server when reuseExistingServer:true`. Should I also add it for those tests? If so, should that be done in the same, or in a separate commit?

docs/src/test-api/class-testconfig.md
```diff
@@ -595,7 +595,7 @@ export default defineConfig({
 - type: ?<[Object]|[Array]<[Object]>>
   - `command` <[string]> Shell command to start. For example `npm run start`..
   - `port` ?<[int]> The port that your http server is expected to appear on. It does wait until it accepts connections. Exactly one of `port` or `url` is required.
-  - `url` ?<[string]> The url on your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the server is ready to accept connections. Exactly one of `port` or `url` is required.
+  - `url` ?<[string]> The url on your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the server is ready to accept connections. Redirects (3xx status codes) are being followed and the new location is checked. Exactly one of `port` or `url` is required.
```
pavelfeldman, line 598: I like the simple formula that we currently document:

```
The url on your http server that is expected to return a
2xx, 3xx, 400, 401, 402, or 403 status code when the
server is ready to accept connections.
```

Does that not work? We can prevent probe from following the redirects if current behavior does not much the doc above.

ghost91-: Yeah, I mentioned that in the PR description: At the moment, the docs don’t match reality: redirects are _attempted_ to be followed, but it’s buggy. I am fine with both not following redirects, or fixing redirects. But not following redirects is more work, because the underlying utility (httpRequest) does not support that yet. And I wanted to keep things as simple as possible. But if that’s the preferred solution, I’m also Ok with that. I’d probably need some guidance on how to add that „don’t follow redirects“ option to `httpRequest`.

pavelfeldman: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/network.ts#L78 we are using http.request for that, which should not follow redirects. Could you add a failing tests that surfaces the bug?

ghost91-: `http.request` may not follow redirects, but there is dedicated code in `httpRequest` to do so: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/network.ts#L73

But this code is broken, which is the whole point of this PR. At the moment it fixes the broken redirect implementation. But I’m also happy to remove the redirect behavior, if that’s preferred.

Regarding a test case that fails, here is the simplest one I can think of:
```ts

test('should follow redirects', async ({ runInlineTest, server }) => {
  server.setRedirect('/redirect', '/redirected-to');
    'test.spec.ts': `
      test('connect to the server', async () => {
        expect(true).toBe(true);
      });
    `,
    'playwright.config.ts': `
      module.exports = {
        webServer: {
          command: 'exit 1',
          url: 'http://localhost:${server.PORT}/redirect',
          reuseExistingServer: true,
        }
      };
    `,
  });
});
```

packages/playwright-core/src/utils/network.ts
```diff
@@ -71,7 +71,7 @@ export function httpRequest(params: HTTPRequestParams, onResponse: (r: http.Inco
   const requestCallback = (res: http.IncomingMessage) => {
     const statusCode = res.statusCode || 0;
     if (statusCode >= 300 && statusCode < 400 && res.headers.location)
-      httpRequest({ ...params, url: res.headers.location }, onResponse, onError);
+      httpRequest({ ...params, url: new URL.URL(res.headers.location, params.url).toString() }, onResponse, onError);
```
mxschmitt, line 74:
```suggestion
      httpRequest({ ...params, url: new URL(res.headers.location, params.url).toString() }, onResponse, onError);
```

change this to that and `import * as URL from 'url';` to `import * as url from 'url';`.

Clarified offline with Pavel that we want to continue to follow redirects.

mxschmitt: actually its fine, we used to do this before. Merging in.

PR #22046, opened 2023-03-29
fix: per-user cache folders

### LINKED ISSUE #21859: [BUG] permission denied accessing '/tmp/playwright-transform-cache'

aslushnikov:
Fixes https://github.com/microsoft/playwright/issues/21859

### DIFF
--- a/packages/playwright-test/src/common/compilationCache.ts
+++ b/packages/playwright-test/src/common/compilationCache.ts
@@ -19,6 +19,7 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 import { sourceMapSupport } from '../utilsBundle';
+import { sanitizeForFilePath } from '../util';
 
 export type MemoryCache = {
   codePath: string;
@@ -29,7 +30,7 @@ export type MemoryCache = {
 const version = 13;
 
 const DEFAULT_CACHE_DIR_WIN32 = path.join(os.tmpdir(), `playwright-transform-cache`);
-const DEFAULT_CACHE_DIR_POSIX = path.join(os.tmpdir(), `playwright-transform-cache-` + os.userInfo().username);
+const DEFAULT_CACHE_DIR_POSIX = path.join(os.tmpdir(), `playwright-transform-cache-` + sanitizeForFilePath(os.userInfo().username));
 
 const cacheDir = process.env.PWTEST_CACHE_DIR || (process.platform === 'win32' ? DEFAULT_CACHE_DIR_WIN32 : DEFAULT_CACHE_DIR_POSIX);
 
--- a/packages/playwright-test/src/common/compilationCache.ts
+++ b/packages/playwright-test/src/common/compilationCache.ts
@@ -27,7 +27,11 @@ export type MemoryCache = {
 };
 
 const version = 13;
-const cacheDir = process.env.PWTEST_CACHE_DIR || path.join(os.tmpdir(), 'playwright-transform-cache');
+
+const DEFAULT_CACHE_DIR_WIN32 = path.join(os.tmpdir(), `playwright-transform-cache`);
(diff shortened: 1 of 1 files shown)


packages/playwright-test/src/common/compilationCache.ts
```diff
@@ -27,7 +27,11 @@ export type MemoryCache = {
 };
 
 const version = 13;
-const cacheDir = process.env.PWTEST_CACHE_DIR || path.join(os.tmpdir(), 'playwright-transform-cache');
+
+const DEFAULT_CACHE_DIR_WIN32 = path.join(os.tmpdir(), `playwright-transform-cache`);
+const DEFAULT_CACHE_DIR_POSIX = path.join(os.tmpdir(), `playwright-transform-cache-` + os.userInfo().username);
```
dgozman, line 32: Are we sure `username` is safe to use in file paths?

dgozman (approved):

packages/playwright-test/src/common/compilationCache.ts
aslushnikov: According to https://unix.stackexchange.com/a/157431/323684 - seems to be safe on Debian. However, I added sanitization - just in case.

daydayhappychao:
Same issue as https://github.com/microsoft/playwright/issues/21859.
I really want to use this update, when will it be merged into the online version ?

packages/playwright-test/src/common/compilationCache.ts
```diff
@@ -27,7 +28,11 @@ export type MemoryCache = {
 };
 
 const version = 13;
-const cacheDir = process.env.PWTEST_CACHE_DIR || path.join(os.tmpdir(), 'playwright-transform-cache');
+
+const DEFAULT_CACHE_DIR_WIN32 = path.join(os.tmpdir(), `playwright-transform-cache`);
+const DEFAULT_CACHE_DIR_POSIX = path.join(os.tmpdir(), `playwright-transform-cache-` + sanitizeForFilePath(os.userInfo().username));
```
pei-wang, line 33: Hi @aslushnikov 
I encountered some issue which caused by `os.userInfo()`. u may want to use the alternative what Cypress use.
https://github.com/cypress-io/cypress/pull/17488

mxschmitt: Please file an GitHub issue, thanks!

PR #22067, opened 2023-03-29
feat(expect): expect(locator).toBeAttached()

### LINKED ISSUE #13467: [Feature] Add isAttached or isExisting to element
### Summary

Implement an easy way to check whether an element is attached to the DOM.

```ts
test('Awardee institution country/region', async ({ page }) => {
  const isExisting = await page.locator('body').isExisting() // returns true or false
})
```

It's really not convenient to do something like this every time
```ts
const els = await page.locator('body').elementHandles()
const isExisting = els.length > 0
```

#### Existing functionality 

Please don't suggest waiters (wait for state to be attached/not attached, etc), visibility checks (isVisible, isHidden) or similar. This is completely different requests.
Thank you

dgozman:
Fixes #13467.

### DIFF
--- a/docs/src/api/class-locatorassertions.md
+++ b/docs/src/api/class-locatorassertions.md
@@ -86,6 +86,20 @@ assertThat(locator).not().containsText("error");
 await Expect(locator).Not.ToContainTextAsync("error");
 ```
 
+## async method: LocatorAssertions.NotToBeAttached
+* since: v1.33
+* langs: python
+
+The opposite of [`method: LocatorAssertions.toBeAttached`].
+
+### option: LocatorAssertions.NotToBeAttached.attached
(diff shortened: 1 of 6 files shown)


pavelfeldman (approved):

joszo:
I've got the feeling that there is some kind o flakiness currently:

`    Error: locator.scrollIntoViewIfNeeded: Element is not attached to the DOM
      waiting for element to be  and stable
        element is not stable - waiting...
      element is  and stable

       at test/e2e/playwright/pages/browser/sections/search.section.ts:245

      243 |
      244 |         await expect(condition).toBeAttached();
    > 245 |         await condition.scrollIntoViewIfNeeded();
          |                         ^`

PR #22101, opened 2023-03-30
feat: rename Locator.filter(locator) to Locator.and

dgozman:


angelo-loria:
This `and` locator looks handy. I'm curious about the examples in the README, though. Correct me if I'm wrong, but I would think that `page.getByRole('button').and(page.getByTitle('Subscribe'));` would return the same locator as `getByRole('button', { name: 'Subscribe' })`, making the `and` useless in this case. I suppose you could have a `button` element with a specific `title` tag for a tooltip or something but a more clear use case might be helpful here.

pavelfeldman (approved):

PR #22105, opened 2023-03-30
chore: allow force reload in traceviewer/watch mode

### LINKED ISSUE #22097: [BUG] Using CMD + SHIFT + R in UI Mode makes it blank
### System info
- Playwright Version: [v1.32.1]
- Operating System: [macOS 13.2.1]
- Other info: UI Mode

**Steps**
- Open UI Mode
- Press CMD + SHIFT + R

**Expected**

CMD + R reloads the UI Mode. CMD + SHIFT + R should reload it too

**Actual**

UI Mode becomes blank.
Image 30-03-2023 at 15 06

mxschmitt:
According to the [spec](https://w3c.github.io/ServiceWorker/#navigator-service-worker-controller) if a user does a force refresh, navigator.serviceWorker.controller is `null` which is why it was hanging before. So we need to first unregister the SW and the register a new one to be controlled and have access to a SW.

Fixes https://github.com/microsoft/playwright/issues/22097

### DIFF
--- a/packages/trace-viewer/src/index.tsx
+++ b/packages/trace-viewer/src/index.tsx
@@ -26,6 +26,13 @@ import { WorkbenchLoader } from './ui/workbenchLoader';
   if (window.location.protocol !== 'file:') {
     if (window.location.href.includes('isUnderTest=true'))
       await new Promise(f => setTimeout(f, 1000));
+    // If a user does a force reload (Cmd+R), there might be an old service worker
+    // that is still running BUT it is NOT controlling the page which is by spec.
+    // In this case, we unregister all service workers first.
+    if (!navigator.serviceWorker.controller) {
+      const registrations = await navigator.serviceWorker.getRegistrations();
+      await Promise.all(registrations.map(r => r.unregister()));
+    }
     navigator.serviceWorker.register('sw.bundle.js');
     if (!navigator.serviceWorker.controller) {
       await new Promise<void>(f => {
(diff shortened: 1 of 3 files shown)


packages/trace-viewer/src/index.tsx
```diff
@@ -26,11 +26,10 @@ import { WorkbenchLoader } from './ui/workbenchLoader';
   if (window.location.protocol !== 'file:') {
     if (window.location.href.includes('isUnderTest=true'))
       await new Promise(f => setTimeout(f, 1000));
-    navigator.serviceWorker.register('sw.bundle.js');
+    await navigator.serviceWorker.register('sw.bundle.js');
     if (!navigator.serviceWorker.controller) {
-      await new Promise<void>(f => {
```
pavelfeldman, line 31: I believe this construct was used for a reason.

packages/trace-viewer/src/index.tsx
```diff
@@ -26,12 +26,11 @@ import { WorkbenchLoader } from './ui/workbenchLoader';
...
-    navigator.serviceWorker.register('sw.bundle.js');
-    if (!navigator.serviceWorker.controller) {
-      await new Promise<void>(f => {
-        navigator.serviceWorker.oncontrollerchange = () => f();
-      });
-    }
+    await navigator.serviceWorker.register('sw.bundle.js');
+    // navigator.serviceWorker.controller is null for a force refresh, hence
+    // we do a soft refresh to be controlled by the service worker again.
+    if (!navigator.serviceWorker.controller)
```
pavelfeldman, line 32: There is no relation between the service worker installation / activation and its registration, so the old code was good.

PR #22127, opened 2023-03-31
chore: implement jsHandle._objectCount in chromium

pavelfeldman:


dgozman (approved):
Let's add a test?

pavelfeldman:
The test is there, I just needed to git add it.

PR #22129, opened 2023-03-31
chore: do not leak on clicks

### LINKED ISSUE #22115: [BUG] Simple click leaks the element
```js
test('should not leak', async ({ page }) => {
  for (let i = 0; i < 100; ++i) {
    await page.evaluate((i) => {
      const element = document.createElement('button');
      element.textContent = 'BUTTON ' + i;
      document.body.appendChild(element);
    }, i);
    await page.click('button');
    await page.evaluate(() => {
      document.body.textContent = '';
    });
  }
});
```

pavelfeldman:
Fixes #22115

### DIFF
--- a/packages/playwright-core/src/server/dom.ts
+++ b/packages/playwright-core/src/server/dom.ts
@@ -47,7 +47,7 @@ export class FrameExecutionContext extends js.ExecutionContext {
   readonly world: types.World | null;
 
   constructor(delegate: js.ExecutionContextDelegate, frame: frames.Frame, world: types.World|null) {
-    super(frame, delegate);
+    super(frame, delegate, world || 'content-script');
     this.frame = frame;
     this.world = world;
   }
@@ -114,7 +114,7 @@ export class FrameExecutionContext extends js.ExecutionContext {
         );
         })();
       `;
-      this._injectedScriptPromise = this.rawEvaluateHandle(source).then(objectId => new js.JSHandle(this, 'object', undefined, objectId));
+      this._injectedScriptPromise = this.rawEvaluateHandle(source).then(objectId => new js.JSHandle(this, 'object', 'InjectedScript', objectId));
     }
     return this._injectedScriptPromise;
   }
@@ -455,6 +455,7 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
         // Do not await here, just in case the renderer is stuck (e.g. on alert)
         // and we won't be able to cleanup.
         hitTargetInterceptionHandle!.evaluate(h => h.stop()).catch(e => {});
+        hitTargetInterceptionHandle!.dispose();
       });
     }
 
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/dom.ts
```diff
@@ -455,6 +455,7 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
         // Do not await here, just in case the renderer is stuck (e.g. on alert)
         // and we won't be able to cleanup.
         hitTargetInterceptionHandle!.evaluate(h => h.stop()).catch(e => {});
+        hitTargetInterceptionHandle!.dispose();
```
yury-s, line 458: Does it mean we'll call a disposed handle at line 474 below?

tests/page/page-leaks.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+import { MultiMap } from '../../packages/playwright-core/lib/utils/multimap';
+import { test, expect } from './pageTest';
+
+function leakedJSHandles(): string {
+  const map = new MultiMap();
+  for (const [h, e] of (globalThis as any).leakedJSHandles.entries()) {
+    const name = `[${h.worldName()}] ${h.preview()}`;
+    if (name === '[main] UtilityScript' || '[utility] UtilityScript' || '[main] InjectedScript' || '[utility] InjectedScript')
```
yury-s, line 24: [...].includes(name)

tests/page/page-leaks.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MultiMap } from '../../packages/playwright-core/lib/utils/multimap';
+import { test, expect } from './pageTest';
+
+function leakedJSHandles(): string {
+  const map = new MultiMap();
+  for (const [h, e] of (globalThis as any).leakedJSHandles.entries()) {
```
dgozman, line 22: Let's cast to the server's JSHandle type here, for future code refactorings.

dgozman (approved):

packages/playwright-core/src/server/javascript.ts
```diff
@@ -64,9 +65,11 @@ export class ExecutionContext extends SdkObject {
   private _delegate: ExecutionContextDelegate;
   private _utilityScriptPromise: Promise<JSHandle> | undefined;
   private _contextDestroyedRace = new ScopedRace();
+  readonly worldName: string;
```
yury-s, line 68: why is it not types.World?

packages/playwright-core/src/server/dom.ts
pavelfeldman: Line 454 is saying this code is exiting.

packages/playwright-core/src/server/javascript.ts
pavelfeldman: Because it is wider than that, I'll rename

yury-s (approved):

PR #22131, opened 2023-03-31
fix(ui mode): preserve manually selected action in live trace

dgozman:


packages/trace-viewer/src/ui/watchMode.tsx
```diff
@@ -514,6 +517,8 @@ const TraceView: React.FC<{
     hideStackFrames={true}
     showSourcesFirst={true}
     rootDir={rootDir}
+    selectedAction={selectedActionCallId ? model?.actions.find(a => a.callId === selectedActionCallId) : undefined}
+    onActionSelected={action => setSelectedActionCallId(action.callId)}
```
pavelfeldman, line 521: This will trigger on each re-render.

pavelfeldman (approved):

PR #22138, opened 2023-04-01
chore(ct): dedupe json object

sand4rt:
Also useful for 3rd party component testing libs..

### DIFF
--- a/packages/playwright-test/types/component.d.ts
+++ b/packages/playwright-test/types/component.d.ts
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+type JsonPrimitive = string | number | boolean | null;
+type JsonValue = JsonPrimitive | JsonObject | JsonArray;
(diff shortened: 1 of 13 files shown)


packages/playwright-test/types/component.d.ts
```diff
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+type JsonPrimitive = string | number | boolean | null;
```
pavelfeldman, line 17: I'm a little worried as this becomes a subject to the public semver, but then again, entire file does. We should rename it to experimentalComponent.d.ts in a follow-up patch to set expectations.

pavelfeldman (approved):

packages/playwright-test/types/component.d.ts
sand4rt: Alright, will do.

Wonder if have you considered a separate folder/package for the shared ct code?

PR #22139, opened 2023-04-01
chore(types): Add export keyword to MakeMatchers type

mikeedwards83:
Add the export keyword to the MakeMatchers type. This allows more flexibility to users create their own test frameworks bases on playwright.

Example usage:

    import { Locator, MakeMatchers } from "@playwright/test"; 
    import { expect as exp } from "@playwright/test";

    export class HtmlElement {
       protected _element: Locator;

        expect(message?: string): MakeMatchers<void, Locator> {
            return exp(this._element, { message: message });
        }

    }

### DIFF
--- a/packages/playwright-test/types/test.d.ts
+++ b/packages/playwright-test/types/test.d.ts
@@ -4189,7 +4189,7 @@ interface GenericAssertions<R> {
 
 type BaseMatchers<R, T> = GenericAssertions<R> & PlaywrightTest.Matchers<R, T>;
 
-type MakeMatchers<R, T> = BaseMatchers<R, T> & {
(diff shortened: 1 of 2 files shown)


mxschmitt:
For me it [inferred](https://www.typescriptlang.org/docs/handbook/type-inference.html) already it automatically, would that work for you?

image

See also here:

image

mikeedwards83:

mikeedwards83:
@mxschmitt 

I would also prefer to have type exposed so that it could be passed around and have a defined return type.

pavelfeldman:
Every export increases our exposure to semver, so we prefer to not do that.

mikeedwards83:
@pavelfeldman, I don't understand this. What is the point of a type system if you don't expose the types?

PR #22164, opened 2023-04-03
test: add regression test for screencast bug

aslushnikov:
Fixed firefox is already rolled in.

Fixes https://github.com/microsoft/playwright/issues/22069

### DIFF
--- a/tests/library/video.spec.ts
+++ b/tests/library/video.spec.ts
@@ -292,6 +292,24 @@ it.describe('screencast', () => {
     expect(fs.existsSync(path)).toBeTruthy();
   });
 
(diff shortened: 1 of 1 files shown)


tests/library/video.spec.ts
```diff
@@ -292,6 +292,24 @@ it.describe('screencast', () => {
     expect(fs.existsSync(path)).toBeTruthy();
   });
 
+  it.only('should work with weird screen resolution', async ({ browser }, testInfo) => {
```
mxschmitt, line 295:
```suggestion
  it('should work with weird screen resolution', async ({ browser }, testInfo) => {
```

mxschmitt (approved):

dgozman (approved):

PR #22174, opened 2023-04-03
fix(expect): report expect "Timed out" when it actually does

### LINKED ISSUE #21664: [BUG] Misleading expect timeout in reports
**Context:**

- Playwright Version: 1.31.2
- Node.js version: v18.14.1
- Browser: Chromium
- Extra: Base on default project after running `npm init playwright@latest`

**Test**
Configuration:
```

export default defineConfig({
  testDir: './tests',
  timeout: 5 * 1000,
  expect: {
    timeout: 3000
  },
  fullyParallel: true,
  reporter: 'html',

  use: {
    actionTimeout: 0,
  },

  projects: [
    {
      name: 'chromium',
    },
  ],
});
```

Sample test:
```
import { test, expect } from '@playwright/test'

test.describe('has title', () => {
    test.only('create item', async ({ page }) => {
        await page.goto('https://playwright.dev/')
        await expect(page).toHaveTitle(/Playwright/)
        // Below line triggers wrong error message about expected timeout in HTLM report
        await expect.soft(page.getByRole('link', { name: 'Get started BUG' })).toBeEnabled({timeout: 7*1000}); 
    })
})
```

**Describe the bug**
In report there is info about timeout 7000ms during waiting for element, while there was timeout after 5000ms (taken from test timeout from config):
image_2023_03_14T14_41_44_801Z

**Expected**
Information that test failed after test timeout, not after expect timeout (that was set in test to greater value than test timeout).

dgozman:
Previously, it would say "Timed out" when page was closed at test timeout, or not say "Timed out" when at least one element matched.

Fixes #21664.

### DIFF
--- a/tests/playwright-test/ui-mode-test-progress.spec.ts
+++ b/tests/playwright-test/ui-mode-test-progress.spec.ts
@@ -53,7 +53,7 @@ test('should update trace live', async ({ runUITest, server }) => {
   ).toHaveText([
     /browserContext.newPage[\d.]+m?s/,
     /page.gotohttp:\/\/localhost:\d+\/one.html/
-  ]);
+  ], { timeout: 15000 });
 
   await expect(
       listItem.locator(':scope.selected'),
@@ -138,7 +138,7 @@ test('should preserve action list selection upon live trace update', async ({ ru
     /browserContext.newPage[\d.]+m?s/,
     /page.gotoabout:blank[\d.]+m?s/,
     /page.setContent[\d.]+m?s/,
-  ]);
+  ], { timeout: 15000 });
 
   // Manually select page.goto.
   await page.getByTestId('action-list').getByText('page.goto').click();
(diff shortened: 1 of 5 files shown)


pavelfeldman (approved):

tests/playwright-test/ui-mode-test-progress.spec.ts
```diff
@@ -53,7 +53,7 @@ test('should update trace live', async ({ runUITest, server }) => {
   ).toHaveText([
     /browserContext.newPage[\d.]+m?s/,
     /page.gotohttp:\/\/localhost:\d+\/one.html/
-  ]);
+  ], { timeout: 15000 });
```
pavelfeldman, line 56: slowExpect!

pavelfeldman (approved):

tests/playwright-test/ui-mode-test-progress.spec.ts
dgozman: Indeed!

PR #22212, opened 2023-04-05
feat(html-report): add attachmentsBaseURL option

### LINKED ISSUE #21636: [Feature] Overriding html report attachment paths
Hello,

In CI, once report is generated, I upload all artifacts (images, videos, traces) from `data/` to external source for storage, and thus, I would like to update paths in the actual index.html report to use `https://someurl.com/filename.png` instead of `data/filename.png`.

I though about doing just replace like `htmlAsString.replace(/src="data\//g, 'src="my-url"')` but that does not seem to work.

Could you suggest any way I could replace attachment paths?

MindaugasMateika:
Fixes https://github.com/microsoft/playwright/issues/21636


MindaugasMateika:
I'm not sure how to make test work since it uses stable test runner which does not have the option that I added

packages/playwright-test/src/reporters/html.ts
```diff
@@ -197,11 +201,13 @@ class HtmlBuilder {
   private _testPath = new Map<string, string[]>();
   private _dataZipFile: ZipFile;
   private _hasTraces = false;
+  private _attachmentsBaseURL: string | undefined;
 
-  constructor(outputDir: string) {
+  constructor(outputDir: string, attachmentsBaseURL: string | undefined) {
```
pavelfeldman, line 206: Sounds like 'data/' could be a default value?

packages/playwright-test/src/reporters/html.ts
```diff
@@ -89,13 +92,14 @@ class HtmlReporter implements Reporter {
...
 
-  _resolveOptions(): { outputFolder: string, open: HtmlReportOpenOption } {
+  _resolveOptions(): { outputFolder: string, open: HtmlReportOpenOption, attachmentsBaseURL: string | undefined } {
     let { outputFolder } = this._options;
     if (outputFolder)
       outputFolder = path.resolve(this.config._internal.configDir, outputFolder);
     return {
       outputFolder: reportFolderFromEnv() ?? outputFolder ?? defaultReportFolder(this.config._internal.configDir),
       open: process.env.PW_TEST_HTML_REPORT_OPEN as any || this._options.open || 'on-failure',
+      attachmentsBaseURL: this._options.attachmentsBaseURL
```
pavelfeldman, line 102:
```suggestion
      attachmentsBaseURL: this._options.attachmentsBaseURL || 'data/'
```

packages/playwright-test/src/reporters/html.ts
```diff
@@ -429,7 +435,7 @@ class HtmlBuilder {
         return {
           name: a.name,
           contentType: a.contentType,
-          path: 'data/' + sha1,
+          path: this._attachmentsBaseURL ? this._attachmentsBaseURL + sha1 : 'data/' + sha1,
```
pavelfeldman, line 438: path.join to allow no trailing slash.

pavelfeldman:
I wonder if addressing it at runtime would be more sound. I.e. you would provide a query parameter that sets the baseURL for artifacts. That way you can generate the report, it'll be readable locally, and it will also be readable with the overridden artifacts location.

packages/playwright-test/src/reporters/html.ts
MindaugasMateika: Tried to but it seems that path.join doesn't work well with URLs and it removes one slash from `https://` and gives `https:/`. Should I leave it as it is or sanitize path in some other way?

MindaugasMateika:
> > I wonder if addressing it at runtime would be more sound. I.e. you would provide a query parameter that sets the baseURL for artifacts. That way you can generate the report, it'll be readable locally, and it will also be readable with the overridden artifacts location.
> 
> @MindaugasMateika what are your thoughts on this? Looks like an overall better approach to me.

It may sound better overall but it's not gonna work for my case. I would rather stick with this solution if that's okay for you.

pavelfeldman:
> It may sound better overall but it's not gonna work for my case.

Could you elaborate on the use case?

MindaugasMateika:
> > It may sound better overall but it's not gonna work for my case.
> 
> Could you elaborate on the use case?

Report is loaded in an iframe in CI so can't add queries to the URL while the rest of the artifacts (due to their size) are uploaded to an external storage so report can't reach them via `data/...` path.

pavelfeldman:
> Report is loaded in an iframe in CI 

And there is no way to specify the URL for that iframe? I.e. it is opened as a 'file' from some explorer? Which CI is that?

MindaugasMateika:
> > Report is loaded in an iframe in CI
> 
> And there is no way to specify the URL for that iframe? I.e. it is opened as a 'file' from some explorer? Which CI is that?

There isn't. And if you manipulate main url like adding `?q=s:passed` query (which normally works) it does not change report view at all. Report is loaded via plugin (kind of html viewing plugin) on Jenkins.

pavelfeldman (approved):

PR #22218, opened 2023-04-05
chore: return promise from tele receiver dispatch

yury-s:


packages/trace-viewer/src/ui/watchMode.tsx
```diff
@@ -618,7 +618,7 @@ const refreshRootSuite = (eraseResults: boolean): Promise<void> => {
     return;
   }
 
-  receiver?.dispatch(message);
+  receiver?.dispatch(message).catch(() => {});
```
dgozman, line 621: Why catch? I think it would be better to see errors in the console, like before.

dgozman (approved):

packages/trace-viewer/src/ui/watchMode.tsx
yury-s: This is inline with line 602 above.

packages/playwright-test/src/isomorphic/teleReceiver.ts
```diff
@@ -151,7 +156,11 @@ export class TeleReporterReceiver {
       return;
     }
     if (method === 'onEnd') {
-      this._onEnd(params.result);
+      await this._onEnd(params.result);
```
pavelfeldman, line 159: let's return instead and keep dispatch sync to keep its dispatching spirit.

PR #22232, opened 2023-04-06
devops: do not download electron on non-electron workflows

mxschmitt:


pavelfeldman:
This looks good, but maybe we could opt-in instead of opting out? I.e. playwright install will set this environment variable itself, unless some other 'ELECTRON_TEST' variable is set.

pavelfeldman (approved):

PR #22234, opened 2023-04-06
chore(ct): rename component.d.ts

sand4rt:
@pavelfeldman Wonder if you've considered a separate folder/package for the shared CT code instead?

related to: https://github.com/microsoft/playwright/pull/22138#discussion_r1158962859


pavelfeldman:
> @pavelfeldman Wonder if you've considered a separate folder/package for the shared CT code instead?

I'd be fine with that. If you think having a @playwright/experimental-ct-core that other packages depend on is beneficial, I can create one.

pavelfeldman (approved):

sand4rt:
> > @pavelfeldman Wonder if you've considered a separate folder/package for the shared CT code instead?
> 
> I'd be fine with that. If you think having a @playwright/experimental-ct-core that other packages depend on is beneficial, I can create one.

I think it should be definitely given a chance. 

Moving the CT-related code from the `/playwright/packages/playwright-core` directory and the duplicate code from the `playwright/packages/playwright-ct-{framework}` directories to `@playwright/experimental-ct-core` would also facilitate third-party library usage.

pavelfeldman:
I'll do that when I have a chance!

pavelfeldman:
@sand4rt: Actually, you should be able to do that on your own, as long as it stays in our monorepo, a new package in`packages/playwright-ct-core` would do the trick. Our publishing system should pick it up, we should name it as @playwright/experimental-ct-core for consistency. You are already familiar with the `utils/workspace.js` structure, other than that, you might need to add / edit some DEPS files.

sand4rt:
@pavelfeldman Cool, will do!

Issue #21654, opened 2023-03-14
[internal] make WebKit Windows bot green


mxschmitt:
Looks like its failing with 4x `Worker teardown timeout of 30000ms exceeded while tearing down "childProcess".` constantly.

mxschmitt:
Looks good now, windows headless/headed is passing.

PR #22237, opened 2023-04-06
test: kill CLI if test kept running after timeout

mxschmitt:
What happens was the following:
1. CLI codegen test was running (does not play a role if it passed or failed)
2. `runCLI` fixture after the `run` was waiting for `cli.exited`. But the process was never exiting. Reasoning unclear, most likely some bug, but should be fixed separately imo.
3. This ended up in a `Test finished within timeout of 30000ms, but tearing down "runCLI" ran out of time. Please allow more time for the test, since teardown is attributed towards the test timeout budget.` error

How to fix? Still wait for the CLI process to exit, race against a timeout tho, if the process was not exited afterwards kill the process (to cleanup) and throw an Error that there was a leaking process.

https://github.com/microsoft/playwright/issues/21654

### DIFF
--- a/tests/config/commonFixtures.ts
+++ b/tests/config/commonFixtures.ts
@@ -87,19 +87,19 @@ export class TestChildProcess {
   }
 
   async close() {
-    if (!this.process.killed)
+    if (this.process.kill(0))
       this._killProcessGroup('SIGINT');
     return this.exited;
   }
 
   async kill() {
-    if (!this.process.killed)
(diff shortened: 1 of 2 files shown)


pavelfeldman:
You probably want to improve the `childProcess` common fixture instead. After the use, it should first gracefully terminate all the processes it created (it already does that), but after that it should be killing them within the test budget timeout.

pavelfeldman (approved):

PR #22244, opened 2023-04-06
feat: blob reporter

yury-s:
#10437

### DIFF
--- a/packages/playwright-test/src/reporters/blob.ts
+++ b/packages/playwright-test/src/reporters/blob.ts
@@ -151,7 +151,9 @@ function mergeBeginEvents(beginEvents: JsonEvent[]): JsonEvent {
   if (!beginEvents.length)
     throw new Error('No begin events found');
   const projects: JsonProject[] = [];
+  let totalWorkers = 0;
   for (const event of beginEvents) {
+    totalWorkers += event.params.config.workers;
     const shardProjects: JsonProject[] = event.params.projects;
     for (const shardProject of shardProjects) {
       const mergedProject = projects.find(p => p.id === shardProject.id);
@@ -161,10 +163,15 @@ function mergeBeginEvents(beginEvents: JsonEvent[]): JsonEvent {
         mergeJsonSuites(shardProject.suites, mergedProject);
     }
   }
+  const config = {
+    ...beginEvents[0].params.config,
+    workers: totalWorkers,
+    shard: undefined
+  };
   return {
     method: 'onBegin',
     params: {
-      config: beginEvents[0].params.config,
+      config,
       projects,
     }
   };
--- a/packages/playwright-test/src/reporters/blob.ts
+++ b/packages/playwright-test/src/reporters/blob.ts
@@ -20,11 +20,9 @@ import path from 'path';
 import { ManualPromise, ZipFile } from 'playwright-core/lib/utils';
 import { yazl } from 'playwright-core/lib/zipBundle';
 import { Readable } from 'stream';
-import type { Reporter } from '../../types/testReporter';
-import type { FullConfig, FullResult } from '../../types/testReporter';
-import type { BuiltInReporter } from '../common/configLoader';
+import type { FullConfig, FullResult, Reporter } from '../../types/testReporter';
+import type { BuiltInReporter, FullConfigInternal } from '../common/config';
 import type { Suite } from '../common/test';
-import type { FullConfigInternal } from '../common/types';
 import { TeleReporterReceiver, type JsonEvent, type JsonProject, type JsonSuite } from '../isomorphic/teleReceiver';
 import DotReporter from '../reporters/dot';
 import EmptyReporter from '../reporters/empty';
@@ -33,9 +31,9 @@ import JSONReporter from '../reporters/json';
 import JUnitReporter from '../reporters/junit';
 import LineReporter from '../reporters/line';
 import ListReporter from '../reporters/list';
+import { loadReporter } from '../runner/loadUtils';
 import HtmlReporter, { defaultReportFolder } from './html';
 import { TeleReporterEmitter } from './teleEmitter';
-import { loadReporter } from '../runner/loadUtils';
 
 
 type BlobReporterOptions = {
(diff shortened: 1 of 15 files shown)


packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export class BlobReporter extends TeleReporterEmitter {
+  private _messages: any[] = [];
+  private _options: BlobReporterOptions;
+  private _outputFile: string | undefined;
+
+  constructor(options: BlobReporterOptions = {}) {
+    super(message => this._messages.push(message));
+    this._options = options;
+  }
```
pavelfeldman, line 42: lint blank line below.

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+  override onBegin(config: FullConfig<{}, {}>, suite: Suite): void {
+    super.onBegin(config, suite);
+    const outputDir = this._resolveOutputDir(config as FullConfigInternal);
+    let shardSuffix = '';
+    if (config.shard) {
+      const paddedNumber = String(config.shard.current).padStart(String(config.shard.total).length, '0');
```
pavelfeldman, line 48: ChatGPT is suggesting

```js
const paddedNumber = `${config.shard.current}`.padStart(`${config.shard.total}`.length, '0');
```

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+  private _options: BlobReporterOptions;
+  private _outputFile: string | undefined;
+
+  constructor(options: BlobReporterOptions = {}) {
+    super(message => this._messages.push(message));
+    this._options = options;
+  }
+  override onBegin(config: FullConfig<{}, {}>, suite: Suite): void {
+    super.onBegin(config, suite);
+    const outputDir = this._resolveOutputDir(config as FullConfigInternal);
```
pavelfeldman, line 45: Extract output file calculation into a separate method, inline resolveOutputDir there.

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (config.shard) {
+      const paddedNumber = String(config.shard.current).padStart(String(config.shard.total).length, '0');
+      shardSuffix = `-${paddedNumber}-of-${config.shard.total}`;
+    }
+    this._outputFile = path.join(outputDir, `report${shardSuffix}.zip`);
+  }
+
+  override async onEnd(result: FullResult): Promise<void> {
+    await super.onEnd(result);
+    if (!this._outputFile)
```
pavelfeldman, line 56: How can this happen? Looks like we should have thrown when calculating the output file?

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+      return;
+
+    const reportString = JSON.stringify(this._messages, undefined, 2);
```
pavelfeldman, line 59: Let's use the JSONL format for serialization / parsing.

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+    fs.mkdirSync(path.dirname(this._outputFile), { recursive: true });
+    await zipReport(this._outputFile, reportString);
```
pavelfeldman, line 61: ... and stream the messages into the zip, preferably live.

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+}
+
+export async function createMergedReport(config: FullConfigInternal, dir: string, reporterName?: string) {
+  const shardFiles = await sortedShardFiles(dir);
+  console.log(`Merging ${shardFiles}`);
+  const events = await mergeEvents(dir, shardFiles);
+  console.log(`Writing ${reporterName} report to ${dir}`);
+
+  // TODO: use reporterName.
+  const reporter = new HtmlReporter({ _internalResolvedOutputFolder: dir });
```
pavelfeldman, line 79: Looks like you should implement this todo right away

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+  const receiver = new TeleReporterReceiver(path.sep, reporter);
+  // TODO: handle onError.
```
pavelfeldman, line 81: Ditto

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -0,0 +1,184 @@
...
+  return {
+    method: 'onBegin',
+    params: {
+      config: beginEvents[0].params.config,
+      projects,
+    }
+  };
+}
+
+function mergeJsonSuites(jsonSuites: JsonSuite[], parent: JsonSuite | JsonProject) {
```
pavelfeldman, line 138: There is similar code in the telereceiver, what is the difference.

packages/playwright-test/src/reporters/html.ts
```diff
@@ -42,6 +42,7 @@ const kMissingContentType = 'x-playwright/missing';
 type HtmlReportOpenOption = 'always' | 'never' | 'on-failure';
 type HtmlReporterOptions = {
   outputFolder?: string,
+  _internalResolvedOutputFolder?: string,
```
pavelfeldman, line 45: Why can't outputFolder be resolved?

packages/playwright-test/src/reporters/blob.ts
yury-s: This one operates on Json* structures while that works with Tele* and the logic is different in many ways.

packages/playwright-test/src/reporters/blob.ts
yury-s: Let's add it separately.

pavelfeldman (approved):

PR #22257, opened 2023-04-06
feat(chromium): allow to adjust output video quality

SomaticIT:
Hello Team,

Thank you for your amazing work on this project. I use it on different projects and it's very powerful.

### Short description

This PR allows to configure output video quality on Chromium.

### Background

Recently, I tried to generate videos based on HTML5 animation. I tried to use playwright to generate the video but unfortunately, the quality is not sufficient.

I also plan to use playwright to generate "support" videos for a product knowledge base.

### Use cases

- Generate videos from HTML5 animation
- Automate "how-to" videos
- Automate "demonstration" videos
- Reduce quality when recording tests

### Additional notes

- This feature is only available for Chromium. It's OK for my use case but maybe we can add it on other platforms.

- It's my first PR on playwright. Do not hesitate to tell me if everything is correct.

### DIFF
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -679,6 +679,11 @@ When set to `minimal`, only record information necessary for routing from HAR. T
     Actual picture of each page will be scaled down if necessary to fit the specified size.
     - `width` <[int]> Video frame width.
     - `height` <[int]> Video frame height.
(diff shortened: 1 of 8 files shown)


SomaticIT:
@microsoft-github-policy-service agree [company="Touchify"]

SomaticIT:
@microsoft-github-policy-service agree company="Touchify"

yury-s (commented):
Thanks for the PR. Can you start with filing a feature request describing the use case so that we can evaluate what API we may need and how it will work in other browsers (you likely noticed that we don't use ffmpeg in webkit and firefox and these options won't work there as is).

yury-s:
Closing this PR for now as part of housekeeping. Per the comment above we need to figure out a coherent story working across browsers first.

Issue #21425, opened 2023-03-06
[Feature] Update `npx playwright show-report` for yarn & pnpm users


segevfiner:
The code prints the following message: https://github.com/microsoft/playwright/blob/e222874445f69af4ff389f29e322d0a98e2d8c20/packages/playwright-test/src/reporters/html.ts#L128

For yarn & pnpm, you can just use `yarn playwright show-report` & `pnpm playwright show-report`, which is more correct. It would be nice to update this message for such users depending on the used package manager.

mxschmitt:
We could do something like done in create-playwright: [the place where the package manager gets determined](https://github.com/microsoft/create-playwright/blob/f6087f59b04c3b000e326d9edc9ddd28687ce919/src/packageManager.ts#L99-L108).

Happy to get a PR for it.

segevfiner:
This actually might also appear in other messages, so we likely want to fix those as well, I guess a grep on `npx` might reveal those.

PR #22259, opened 2023-04-07
feat: show message depending user package manager

brunocroh:
### Issue

#21425

### Short description

This PR implements, a better message when user, run the command
`playwright test --reporter=html `

Show a message with package manager of user instead of npx to everyone

if running with `yarn` output is 
```
To open last HTML report run:

yarn playwright show-report
```

if running with `npm` output is
```
To open last HTML report run:

npx playwright show-report
```

if running with `pnpm` output is
```
To open last HTML report run:

pnpm exec playwright show-report
```


brunocroh:

mxschmitt (changes requested):
Please revert all the unrelated formatting changes.

brunocroh:
@mxschmitt done, sorry for the mistake

packages/playwright-test/src/util.ts
```diff
@@ -305,3 +305,12 @@ export function envWithoutExperimentalLoaderOptions(): NodeJS.ProcessEnv {
     result.NODE_OPTIONS = result.NODE_OPTIONS.replace(substring, '').trim() || undefined;
   return result;
 }
+
+export function determinePackageManager() {
+  if (process.env.npm_config_user_agent) {
+    if (process.env.npm_config_user_agent.includes('yarn')) return 'yarn';
```
pavelfeldman, line 311: We can use process.env.npm_config_user_agent? onstruct for brevity. Also I don't think `pnpm playwright` does the right thing, you probably want `pnpm exec playwright`.

brunocroh: I believe the right way to get package manage of use is this.
I changed the code to use `getFromENV` like another files of project
And change the message from `pnpm` to `pnpm exec`

If you have some suggest or believe this is not a better way to do that, let me know

pavelfeldman:
Sorry we dropped the ball on this one and now it is obsolete.

PR #22280, opened 2023-04-07
chore: remove FullProjectInternal.from

pavelfeldman:


packages/playwright-test/src/reporters/base.ts
```diff
@@ -523,6 +523,23 @@ function fitToWidth(line: string, width: number, prefix?: string): string {
   return taken.reverse().join('');
 }
 
+export function uniqueProjectIds(projects: FullProject[]): Map<FullProject, string> {
```
yury-s, line 526: This repeats ConfigLoader._assignUniqueProjectIds, can you extract it into a common util?

yury-s (approved):

packages/playwright-test/src/reporters/base.ts
pavelfeldman: I'll follow up

Issue #20093, opened 2023-01-13
[BUG]  Button click with window close not working for Firefox


Rom4ig:
**Context:**
- Playwright Version: 1.29.1
- Operating System: Windows/docker(linux)
- Node.js version: 18.13.0
- Browser: Firefox

**Code Snippet**

```typescript
  await page.getByRole('button', { name: 'Выбрать' }).click()
```
**Console output**
![image](https://user-images.githubusercontent.com/43033763/212257593-73d31bc6-c57c-4623-a2b6-4f99f5842510.png)

After this click, the window closes.
This click works for **Chrome** but doesn't work for **Firefox**.
I also tried this with **delay** and **force** options.

mxschmitt:
Could you provide us a repro for this? Or try to set the `DEBUG=pw:browser` environment variable, since this looks like a browser crash.

aslushnikov:
@Rom4ig thank you for the report! I can repro this:

```ts
// a.spec.ts
import { test } from '@playwright/test';

test('should work', async ({ page }) => {
  await page.setContent('<button>clickme</button>');
  await page.$eval('button', button => button.addEventListener('click', () => {
    window.open('');
  }, false));
  const [popup] = await Promise.all([
    page.waitForEvent('popup'),
    page.locator('button').click(),
  ]);
  await popup.setContent(`<button>clickme</button>`);
  await popup.$eval('button', body => body.addEventListener('click', () => {
    window.close();
  }));
  await Promise.all([
    popup.locator('button').click(), // throws in Firefox, but not in Chromium or WebKit
    popup.waitForEvent('close'),
  ]);
});
```

Rom4ig:
Why if you change to `window.open(document.URL, '_blank', 'location=yes');` then it will work?

aslushnikov:
@Rom4ig heads-up: it turns out this is how Firefox actually works. My "fix" for this particular bug was just masking the problem. 

So this might throw in Firefox in future as well, similarly to https://github.com/microsoft/playwright/issues/21225

PR #22321, opened 2023-04-11
test: remove faulty test

aslushnikov:
Turns out we cannot guarantee order of events in Firefox: it might
close the page before we get a response from renderer.

Thus removing a faulty test.

References https://github.com/microsoft/playwright/issues/20093

### DIFF
--- a/tests/page/page-click.spec.ts
+++ b/tests/page/page-click.spec.ts
@@ -38,22 +38,6 @@ it('should click button inside frameset', async ({ page, server }) => {
   expect(await frame.evaluate('result')).toBe('Clicked');
 });
 
-it('should click a button that closes popup', async ({ page }) => {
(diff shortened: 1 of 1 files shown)


tests/page/page-click.spec.ts
```diff
@@ -38,22 +38,6 @@ it('should click button inside frameset', async ({ page, server }) => {
   expect(await frame.evaluate('result')).toBe('Clicked');
 });
 
-it('should click a button that closes popup', async ({ page }) => {
-  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/20093' });
```
yury-s, line 42: Should we just skip it in Firefox instead?

yury-s (approved):

tests/page/page-click.spec.ts
aslushnikov: I don't think we want to commit to this behavior in any of the browsers.

PR #22361, opened 2023-04-12
docs: update linked words

Aleff13:
Update words to include the "s" character inside de link in the [auth doc](https://playwright.dev/docs/auth#testing-multiple-roles-together)

Before 
image


mxschmitt:
Looks like it contains a lot of unrelated changes.

Aleff13:
> [mxschmitt](/mxschmitt)

Fixed, sorry the prettier was enabled

mxschmitt:
This will break their links, so it was expected before that only the Page/BrowserContext word was linkified.

PR #22366, opened 2023-04-12
test: make sure process killing logic works without `ps` on Linux

aslushnikov:
The bare-bones `debian` distribution we use for testing doesn't have
`ps`. This patch switches to reading `/proc` file system directly on
Linux instead of relying on `ps`.

Performance measurements for the 20000 active processes on Debian Docker
container, tested on my M1 Max Mac:
- the `ps -eo pid,ppid,pgid` call + parsing takes 293ms
- the manual synchronous `/proc` traversal + parsing takes 326ms

So this is ~10% perf penalty.

Drive-by: rename `pgid` into `pgrp` so that it stands out from `pid`
(process ID) and `ppid` (parent process ID).

### DIFF
--- a/tests/config/commonFixtures.ts
+++ b/tests/config/commonFixtures.ts
@@ -18,6 +18,7 @@ import type { Fixtures } from '@playwright/test';
 import type { ChildProcess } from 'child_process';
 import { execSync, spawn } from 'child_process';
 import net from 'net';
+import fs from 'fs';
 import { stripAnsi } from './utils';
 
 type TestChildParams = {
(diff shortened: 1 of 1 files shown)


tests/config/commonFixtures.ts
```diff
@@ -32,28 +33,59 @@ import childProcess from 'child_process';
...
+      const match = statFile.match(/^(?<pid>\d+)\s+\((?<comm>.*)\)\s+(?<state>R|S|D|Z|T|t|W|X|x|K|W|P)\s+(?<ppid>\d+)\s+(?<pgrp>\d+)/);
+      if (match) {
+        result.push({
+          pid: +match.groups.pid,
+          ppid: +match.groups.ppid,
+          pgrp: +match.groups.pgrp,
+        });
+      }
+    } catch (e) {
+      // We don't have access to some /proc/<pid>/stat file.
```
yury-s, line 58: should we log this for easier debugging?

yury-s (approved):

tests/config/commonFixtures.ts
aslushnikov: I don't think we should: all the descendants should be accessible, and all the other processes (e.g. root-spawned) might happen regularly. This is not an edge-case.

PR #22375, opened 2023-04-13
feat(firefox): roll to r1397

playwrightmachine:


aslushnikov:
Seems to be crashing on headed; follow-up coming.

PR #22398, opened 2023-04-14
Change snippet for the "Use web first assertions" chapter

bvanderneut:
Change the example snippet so it's not the same as the snippet below.


mxschmitt:
The point is to be "the same" as below, just the first variant is bad practise and below its best practise.

Closing by that.

PR #22413, opened 2023-04-14
fix(deps): install fonts-liberation2

remcohaszing:
Everywhere where `fonts-liberation` is installed, now also `fonts-liberation2` is installed too.

This package affects the rendering of certain fonts in Chromium.

I found out adding this dependency fixes the font rendering of FontAwesome icons in [`mermaid-isomorphic`](https://github.com/remcohaszing/mermaid-isomorphic). In particular this fixes the errors shown in [this CI job](https://github.com/remcohaszing/mermaid-isomorphic/actions/runs/4701982182/jobs/8338682513).


remcohaszing:
Based on https://askubuntu.com/questions/965002/difference-between-fonts-liberation-and-fonts-liberation2-packages I’m not sure if Playwright should include `fonts-liberation2` or if I should uninstall it from my system though. :thinking:

`google-chrome-*` depend on `fonts-liberarion`, but not `fonts-liberation2`.

remcohaszing:
Installing `fonts-liberation2` makes a Playwright behave the same as my machine in a Docker container, but uninstalling `fonts-liberation2` from my machine didn’t make it behave the same as a Docker container. I find this peculiar, but I don’t know if installing `fonts-liberation2` is the way to go.

In the end I had another issue, which was solved by loading all fonts using:

```js
await Promise.all(Array.from(document.fonts, (font) => font.load()))
```

Issue #22002, opened 2023-03-27
[BUG] Playwright trace should format console messages with % sequences properly


segevfiner:
### System info
- Playwright Version: 1.31.2
- Operating System: macOS 13.2
- Browser: Chromium
- Other info:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Link to the GitHub repository with the repro**

[https://github.com/your_profile/playwright_issue_title]

or

**Config file**

```js
// playwright.config.ts
import { defineConfig, devices } from '@playwright/test';

export default defineConfig({
  use: {
    trace: 'on',
  },
  projects: [
    {
      name: 'chromium',
      use: { ...devices['Desktop Chrome'], },
    },
});
```

**Test file (self-contained)**

```js
it('should check the box using setChecked', async ({ page }) => {
  await page.evaluate(() => {
    console.log('%s', 'Hello, World!');
  });
});
```

**Steps**
- Run the test 
- Open the trace and check the console tab.

**Expected**

The message is formatted like displayed in the browser dev tools: "Hello, World!"

**Actual**

You get "%s    Hello, World!" instead.

aditya-sriram:
Hello. I am a student in EECS 481 at the university of Michigan and I have an assignment to contribute to an open source project on GitHub. I have programming experience in typescript and JavaScript. I have read the issue closely, familiarized myself with the codebase and read the contribution guidelines. I would like to work on this issue.

pavelfeldman:
Should be fixed

PR #22424, opened 2023-04-15
feat(tracing): string formatting in console (#22002)

aditya-sriram:
This is a patch for the bug identified in issue (https://github.com/microsoft/playwright/issues/22002) in the console tab in the trace viewer. It formats substitution strings according to the substitution sequence.

See my comment in the issue thread about my background and my motivation for choosing this project. Any feedback is appreciated and valued.

Thank you.


package.json
```diff
@@ -88,6 +88,7 @@
     "mime": "^3.0.0",
     "ncp": "^2.0.0",
     "node-stream-zip": "^1.15.0",
+    "printj": "^1.3.1",
```
pavelfeldman, line 91: I'm not comfortable bringing it as a dependency, we should either perform a clean room implementation or pull it as a third party. a new implementation might be better.

pavelfeldman:
Closing as per above

Issue #22122, opened 2023-03-31
[Feature] In remote browser mode retain-on-failure for trace and video can avoid bringing content from remote server


puagarwa:
Let us know what functionality you'd like to see in Playwright and what your use case is.
When we use tracing with remote browser connect, video and traces takes some extra time to get downloaded at client side. As a possible optimization when user uses retain-on-failure option, can we avoid bringing this content from remote server to client? It could save extra time being spent in downloading for passed test scenarios. 

Do you think others might benefit from this as well?
Yes, it would help other user using remote browser mode and using tracing as data transfer would cause extra latency.

puagarwa:
Update: During my testing, it seems video is already optimized but problem is only with trace.

Issue #22120, opened 2023-03-31
Do not send trace file for passed tests when retain-on-failure is set.


pavelfeldman:
<eom>

PR #22442, opened 2023-04-17
fix: preserve trace if context if closed manually during test run

yury-s:
Reference #22122, #22120

### DIFF
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -24,6 +24,7 @@ import type { TestInfoImpl } from './worker/testInfo';
 import { rootTestType } from './common/testType';
 import { type ContextReuseMode } from './common/config';
 import { artifactsFolderName } from './isomorphic/folders';
+import { runningTestBody } from './common/globals';
 export { expect } from './matchers/expect';
 export { store as _store } from './store';
 export const _baseTest: TestType<{}, {}> = rootTestType.test;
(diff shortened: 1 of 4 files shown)


packages/playwright-test/src/worker/workerMain.ts
```diff
@@ -369,8 +369,13 @@ export class WorkerMain extends ProcessRunner {
       await testInfo._runFn(async () => {
         // Now run the test itself.
         debugTest(`test function started`);
+        setRunningTestBody(true);
```
dgozman, line 372: Let's assign a property on `TestInfoImpl` instead.

packages/playwright-test/src/index.ts
```diff
@@ -331,7 +332,10 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
       (tracing as any)[startedCollectingArtifacts] = true;
       if (captureTrace) {
         let tracePath;
-        if (preserveTrace()) {
+        // Create a trace file if we know that:
+        // - it is's going to be used due to the config setting and the test status or
+        // - we are inside a test and the user manually closed the context.
+        if (preserveTrace() || runningTestBody()) {
```
dgozman, line 338: I guess we can also have a new context created and closed in `beforeEach`/`afterEach` hook, and we should preserve its trace as well? Perhaps just check whether we have started the fixture teardown phase or not.

packages/playwright-test/src/index.ts
```diff
@@ -24,6 +24,7 @@ import type { TestInfoImpl } from './worker/testInfo';
 import { rootTestType } from './common/testType';
 import { type ContextReuseMode } from './common/config';
 import { artifactsFolderName } from './isomorphic/folders';
+import { runningTestBody } from './common/globals';
```
pavelfeldman, line 27: This looks like a layering violation to me: index.ts is a part of Playwright, globals are a part of folio. Why do we need to know the stage of the test if we are the ones that start and end the tracing?

packages/playwright-test/src/worker/workerMain.ts
yury-s: Added one on the context instead.

packages/playwright-test/src/index.ts
yury-s: We don't.

dgozman (approved):

PR #22443, opened 2023-04-17
chore(junit): fallback to path extension for attachment name

### LINKED ISSUE #22396: [BUG] xray-report does not provide evidence data types
### System info
- Playwright Version: v1.31.2
- Operating System:  Windows 10

I configured my test system as described here; https://docs.getxray.app/display/XRAYCLOUD/Testing+web+applications+using+Playwright

I noticed that in the xray xml-report the data types are not provided (e.g. ".png", ".webm"). So I have to edit the xml content beforehand to be able to open the evidence in the browser. 
In the default junit reports on the other hand the data types are apparently provided.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/22396

Before this change we were always using `attachment.name` when putting the attachments into the junit report. In the case of a trace, this is e.g. `trace`. After this change we try a best effort there, since attach the trace via a path to the junit report, we put the extension from the path to the attachment name, if there is none.

### DIFF
--- a/packages/playwright-test/src/reporters/junit.ts
+++ b/packages/playwright-test/src/reporters/junit.ts
@@ -205,10 +205,11 @@ class JUnitReporter implements Reporter {
           }
 
           if (contents) {
+            const attachmentName = attachment.name + ((attachment.path && !path.extname(attachment.name)) ? path.extname(attachment.path) : '');
             const item: XMLEntry = {
               name: 'item',
               attributes: {
-                name: attachment.name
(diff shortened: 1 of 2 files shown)


pavelfeldman:
It looks like xray has a richer format for evidence / attachments with filename, content type and content. Making junit strange to fit the junit / xray interoperability looks wrong to me. Maybe someone wants to contribute a proper xray reporter?

PR #22459, opened 2023-04-18
Fix leak in Locator.waitFor

### LINKED ISSUE #22458: [BUG] Locator.waitFor is leaking DOM nodes
### System info
- Playwright Version: v1.31
- Browser: Edge/WebView2
- Other info: It is probably for all system and browsers, but checked only in this one

### Source code

**Test file (pseudocode)**

```js
it('should check element is not leaking', async ({ page }) => {
  await Promise.all([
    () => page.locator('//*[@id="should_not_leak'div"]').waitFor({ state: 'attached' }), 
    () => page.setContent(`<div id='should_not_leak'div></div>`) 
  ])
});
```

**Steps**
- [Run the test]
- [...]

**Expected**

There is no detached div element at the end of the test

**Actual**

Playwright leaks element(s) found by locators, if there is `waitFor` called on them, due to internally used Handles and not disposing them.

dolezel:
Fixes #22458

### DIFF
--- a/packages/playwright-core/src/client/locator.ts
+++ b/packages/playwright-core/src/client/locator.ts
@@ -328,7 +328,11 @@ export class Locator implements api.Locator {
   waitFor(options: channels.FrameWaitForSelectorOptions & { state: 'attached' | 'visible' }): Promise<void>;
   waitFor(options?: channels.FrameWaitForSelectorOptions): Promise<void>;
   async waitFor(options?: channels.FrameWaitForSelectorOptions): Promise<void> {
-    await this._frame._channel.waitForSelector({ selector: this._selector, strict: true, omitReturnValue: true, ...options });
+    return this._frame._wrapApiCall<void>(async () => {
+      const result = await this._frame._channel.waitForSelector({ selector: this._selector, strict: true, omitReturnValue: true, ...options });
+      const handle = ElementHandle.fromNullable(result.element) as ElementHandle<SVGElement | HTMLElement> | null;
+      await handle?.dispose();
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/client/locator.ts
```diff
@@ -328,7 +328,11 @@ export class Locator implements api.Locator {
   waitFor(options: channels.FrameWaitForSelectorOptions & { state: 'attached' | 'visible' }): Promise<void>;
   waitFor(options?: channels.FrameWaitForSelectorOptions): Promise<void>;
   async waitFor(options?: channels.FrameWaitForSelectorOptions): Promise<void> {
-    await this._frame._channel.waitForSelector({ selector: this._selector, strict: true, omitReturnValue: true, ...options });
+    return this._frame._wrapApiCall<void>(async () => {
+      const result = await this._frame._channel.waitForSelector({ selector: this._selector, strict: true, omitReturnValue: true, ...options });
```
yury-s, line 332: I believe omitReturnValue was supposed to avoid handle allocation but got lost during refactorings. Let's either remove 'omitReturnValue' and all dispose here as you do or make 'omitReturnValue' work.

yury-s (commented):
Let's add a test in https://github.com/microsoft/playwright/blob/main/tests/page/page-leaks.spec.ts to make sure this does not regress.

pavelfeldman:
Closing as I landed the one with the backend fix. Thanks though!

dolezel:
Sure, thanks!

Issue #22041, opened 2023-03-28
[BUG] Playwright codegen performance issue with many elements on the page


danielmhair:
### System info
- Playwright Version: v1.31.2
- Operating System: Windows, but most likely in others
- Browser: Chromium, but most likely in others

### Source code

- [ ] I Here is a [repo](https://github.com/danielmhair/playwright-recorder-lag-repro) that reproduces the issue locally. Just run `npx playwright codegen <local-html-file-in-repo>` and then start moving the mouse around in the site to see the lag in performance.
- [ ] | https://www.youtube.com and https://www.amazon.com also has a lag (not as much as my example above)

**Link to the GitHub repository with the repro**

https://github.com/danielmhair/playwright-recorder-lag-repro

**Steps**
- npm ci
- npx playwright codegen "file://location-to-local-version-of-my-repro-repo/open-15000-elements.html"

**Expected**

The HTML files I provide in the repository I link above, it has 15000 elements with very little depth. I initially expected this to run quite fast and not slow down to highlight the elements.

**Actual**
If you notice the codegen runs very slow. We have quite a few use cases that will have 15000 elements or more (I know, not the best in performance for the site itself). I wish I could provide those use cases, but they have login credentials that I can't give out. So hopefully this repo is enough to work through. I understand that this is just one use case that deals with only divs. If needed, I can provide other examples where it can handle cases like text, many ids, placeholders and others to give a better example. But hopefully, this shows the issue enough to improve performance.

---

**Willing to contribute if needed**

Also, I'm happy to help out if needed. If you have any tips on what might help improve the performance, I'd be happy to create a PR for a fix for the improvement.

**Thank you**

I also wanted to mention how awesome playwright is. I have used Cypress for many years, but time and time again, Playwright proves to be better, so I'm planning to complete switch over in my company very soon.

retroace:
Yeah I am facing the same issues. It would be awesome to see it get solved soon

shuyanzhou:
Facing the same issue on my website that lists many products. Hope this could be resolved soon

mxschmitt:
**Investigation notes:**

After looking into it I found two issues:

a) that we clear most of our good caches after each generateSelector call
b) that we in two places not put DOM Elements into our _cached function, we pass a always newly constructed object instead. This gives us always negative cache hits.

When applying fixes for these two issues, the performance is significantly better. Gist: https://gist.github.com/mxschmitt/98936e03196cc06c243cf618007743bc

[...]

PR #22468, opened 2023-04-18
fix(codegen): do not use re-computed arrays as cache keys

mxschmitt:
https://github.com/microsoft/playwright/issues/22041


dgozman:
Closing because this regresses query performance 3x-20x as measured by `should work with large DOM @smoke` test.

PR #22494, opened 2023-04-19
feat(ct): code split for better performance and isolation

sand4rt:
> "Code splitting ensures that the test JS / CSS is the minimal required for a test to run in isolation, which should ease up memory and CPU when executing tests.
>
> Code splitting also introduces script boundaries, which means that a runtime error in the global code of another component would not lead to the entire test suite blowing up.
>
> This comes at a price of a slight increase in bundling time if your module graph is large and you have a lot of components, but this should be offset by the fact that once compilation happens, running individual tests is a lot faster.
>
> When running with an without code-splitting on a large real-world test suite at Atlassian with 900+ tests, I noticed that the average time per test dropped from ~5 seconds / test to < 2 seconds / test when code-splitting. Browsers also consumed less memory and CPU which meant we could push for better parallelisation on the same hardware."

~ https://github.com/pastelsky

### DIFF
--- a/packages/playwright-ct-react/registerSource.mjs
+++ b/packages/playwright-ct-react/registerSource.mjs
@@ -23,13 +23,17 @@ import { createRoot as __pwCreateRoot } from 'react-dom/client';
 /** @typedef {import('../playwright-ct-core/types/component').Component} Component */
 /** @typedef {import('react').FunctionComponent} FrameworkComponent */
 
-/** @type {Map<string, FrameworkComponent>} */
+/** @type {Map<string, Promise<FrameworkComponent>} */
 const __pwRegistry = new Map();
+
+/** @type {Map<string, FrameworkComponent} */
+const __pwRegistryResolved = new Map();
+
 /** @type {Map<Element, import('react-dom/client').Root>} */
 const __pwRootRegistry = new Map();
 
 /**
- * @param {{[key: string]: FrameworkComponent}} components
+ * @param {{[key: string]: () => Promise<FrameworkComponent>}} components
  */
 export function pwRegister(components) {
   for (const [name, value] of Object.entries(components))
@@ -39,14 +43,42 @@ export function pwRegister(components) {
 /**
  * @param {Component} component
  */
+async function __pwResolveAllComponents(component) {
+  if (typeof component !== 'object' || Array.isArray(component))
+    return
+
+  let __pwRegistryKey = component.type
+
+  let componentLoaderFunc = __pwRegistry.get(component.type);
+
+  if (!componentLoaderFunc) {
+    // Lookup by shorthand.
+    for (const [name, value] of __pwRegistry) {
+      if (component.type.endsWith(`_${name}`)) {
+        __pwRegistryKey = name
+        componentLoaderFunc = value;
+        break;
+      }
+    }
+  }
+
+  if(componentLoaderFunc) {
+    __pwRegistryResolved.set(__pwRegistryKey, await componentLoaderFunc())
+  } else {
+    __pwRegistryResolved.set(__pwRegistryKey, component.type)
+  }
+
+  await Promise.all(component.children.map(child => __pwResolveAllComponents(child)))
+}
+
 function __pwRender(component) {
   if (typeof component !== 'object' || Array.isArray(component))
     return component;
 
-  let componentFunc = __pwRegistry.get(component.type);
+  let componentFunc = __pwRegistryResolved.get(component.type);
   if (!componentFunc) {
     // Lookup by shorthand.
-    for (const [name, value] of __pwRegistry) {
+    for (const [name, value] of __pwRegistryResolved) {
       if (component.type.endsWith(`_${name}`)) {
         componentFunc = value;
         break;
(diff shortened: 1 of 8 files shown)


packages/playwright-ct-react/registerSource.mjs
```diff
@@ -23,22 +23,48 @@ import { createRoot as __pwCreateRoot } from 'react-dom/client';
...
 /**
  * @param {Component} component
  */
+async function __pwResolveComponent(component) {
+  if (typeof component !== 'object' || Array.isArray(component))
+    return
+
+  let componentFuncLoader = __pwLoaderRegistry.get(component.type);
+  if (!componentFuncLoader) {
+    // Lookup by shorthand.
```
pavelfeldman, line 52: Carrying over my comment from the original PR: Now that you are resolving it here, do you want to remove it from the line 73 (new code)?

sand4rt: Before I change everything, is this what you had in mind?: https://github.com/microsoft/playwright/pull/22494/commits/7755cc98043808917841e8834a4daa799001a20e

pavelfeldman: Not exactly.

pavelfeldman:
I guess I'm missing something. I thought we had 2-stage process:
- resolve components (including shorthands), populate registry
- render using registry where everything is already resolved (including the shorthands)

packages/playwright-ct-react/registerSource.mjs
sand4rt: Thank you for the feedback. Made another proposal: https://github.com/microsoft/playwright/pull/22494/commits/e8456360e8bd55e5601fd003cda4953078ebe01a what do you think?

packages/playwright-ct-react/registerSource.mjs
```diff
@@ -21,48 +21,79 @@ import * as __pwReact from 'react';
...
+    __pwLoaderRegistry.set(name, value);
+}
+
+/**
+ * @template Component
+ * @param {string} type
+ * @param {Map<string, Component>} componentRegistry
+ */
+function findComponent(type, componentRegistry) {
+  const Component = componentRegistry.get(type);
```
pavelfeldman, line 51: This one seems to be looking up the registration, I.e. within the _pwLoaderRegistry. Name the method respectively and don't pass componentRegistry here, it is global. Otherwise I can get an idea that you can be passing __pwRegistry here sometimes, which would make things very confusing.

packages/playwright-ct-react/registerSource.mjs
```diff
@@ -21,48 +21,79 @@ import * as __pwReact from 'react';
...
+ * @param {{[key: string]: () => Promise<FrameworkComponent>}} components
  */
 export function pwRegister(components) {
   for (const [name, value] of Object.entries(components))
-    __pwRegistry.set(name, value);
+    __pwLoaderRegistry.set(name, value);
+}
+
+/**
+ * @template Component
```
pavelfeldman, line 46: The name Component clashes with the import on line 23.

packages/playwright-ct-react/registerSource.mjs
```diff
@@ -21,48 +21,79 @@ import * as __pwReact from 'react';
...
-  }
+/**
+ * @param {Component} component
+ */
+async function __pwResolveComponent(component) {
+  if (!isComponent(component))
+    return
 
-  if (!componentFunc && component.type[0].toUpperCase() === component.type[0])
+  const Component = findComponent(component.type, __pwLoaderRegistry);
```
pavelfeldman, line 76: Our variable names are camelCase.

pavelfeldman, line 76:
```suggestion
  const componentFactory = findComponentFactory(component.type);
```

packages/playwright-ct-solid/registerSource.mjs
```diff
@@ -23,15 +23,43 @@ import __pwH from 'solid-js/h';
...
+}
+
+/**
+ * @param {Component} component
+ */
+async function __pwResolveComponent(component) {
+  if (typeof component !== 'object' || Array.isArray(component))
+    return
+
+  let componentFuncLoader = __pwLoaderRegistry.get(component.type);
```
pavelfeldman, line 47: I like how you inline it here more than when you extract it in react.

pavelfeldman (approved):
This looks good now. I am adding some comments, but they are nits. I really like how the render code does not change with the new 2-step resolution. You have the old sync code running, plus in you have some pre-work that makes it work with the async registrations.

sand4rt:
> This looks good now. I am adding some comments, but they are nits. I really like how the render code does not change with the new 2-step resolution. You have the old sync code running, plus in you have some pre-work that makes it work with the async registrations.

Thanks again, i think i resolved the comments if i understood you correctly. II'll double check it again tomorrow with some fresh eyes.

BTW. congrats on the 50k GitHub stars!:)

pavelfeldman:
I think your shorthand management went backwards, i.e. it was good on the patch that I reviewed and now it is again resolving shorthands in the pwRender.

PR #22514, opened 2023-04-20
chore: fix ui mode w/ multiple contexts

### LINKED ISSUE #21895: [BUG] UI mode doesn't show live trace when beforeEach is used to make an API call
### System info
- Playwright Version: 1.32
- Operating System: macOS
- Browser: Chromium
- Other info:

### Source code

**Config file**

```js
// playwright.config.ts

export default defineConfig({
  projects: [
    {
      name: 'chromium',
      use: { ...devices['Desktop Chrome'], },
    },
});
```

```js

test.beforeEach(async ({ request }) => {
  const test = await request.get(`https://catfact.ninja/fact`);
  expect(test.ok()).toBeTruthy();
});

test('some steps', async ({ page }) => {
  await page.waitForTimeout(5000)
  await expect(page).toHaveTitle(/Playwright/);
});
```

**Steps**
- Run the test using UI watch mode and observe the live trace

**Expected**
The live trace is rendered as the test runs

**Actual**
The live trace is not rendered as the test runs.

In our framework implementation, we make api call to create a new workspace (in beforeEach block) in every test

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/21895

### DIFF
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -397,7 +397,8 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
     {
       (playwright.request as any)._onDidCreateContext = onDidCreateRequestContext;
       (playwright.request as any)._onWillCloseContext = onWillCloseRequestContext;
-      (playwright.request as any)._defaultContextOptions = _combinedContextOptions;
+      (playwright.request as any)._defaultContextOptions = { ..._combinedContextOptions };
+      (playwright.request as any)._defaultContextOptions.tracesDir = path.join(_artifactsDir(), 'traces');
       const existingApiRequests: APIRequestContext[] =  Array.from((playwright.request as any)._contexts as Set<APIRequestContext>);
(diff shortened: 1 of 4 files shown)


packages/playwright-test/src/index.ts
```diff
@@ -397,7 +397,8 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
     {
       (playwright.request as any)._onDidCreateContext = onDidCreateRequestContext;
       (playwright.request as any)._onWillCloseContext = onWillCloseRequestContext;
-      (playwright.request as any)._defaultContextOptions = _combinedContextOptions;
+      (playwright.request as any)._defaultContextOptions = { ..._combinedContextOptions };
+      (playwright.request as any)._defaultContextOptions.tracesDir = path.join(_artifactsDir(), 'traces');
```
dgozman, line 401: Is this for persistent contexts only? Let's add a comment.

dgozman (approved):

PR #22531, opened 2023-04-20
feat(ct): error message when navigating with page.go

sand4rt:
related to: https://github.com/microsoft/playwright/issues/15889.

Would still prefer `page.goto` to work, but at least an error message is thrown now.

### DIFF
--- a/packages/playwright-ct-core/src/mount.ts
+++ b/packages/playwright-ct-core/src/mount.ts
@@ -40,7 +40,7 @@ export const fixtures: Fixtures<
 
     page: async ({ page }, use, info) => {
       if (!((info as any)._configInternal as FullConfigInternal).defineConfigWasUsed)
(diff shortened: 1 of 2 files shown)


pavelfeldman:
Is there a user bug or a test so that it was easier to reason about it.

sand4rt:
@pavelfeldman think you missed the PR description.

In shorthand:
- `page.go...` is not working in component testing ATM, so a proper error message is shown instead of timing out with a blank screen. Related ticket: https://github.com/microsoft/playwright/issues/15889 
- I can write tests for this!
- Please ignore if you see draft PR. I try to avoid it, but i saw afterwards that some tests are failing.

pavelfeldman:

sand4rt:
> :wave: @sand4rt what should we do with this one?

👋 @pavelfeldman will close this one, its is just annoying that you will get a blank screen when using `page.goto` in component testing.. An error message that this is not supported would be nice but not that important i guess

PR #22533, opened 2023-04-20
chore: introduce expect.configure

pavelfeldman:


packages/playwright-test/src/matchers/expect.ts
```diff
@@ -106,27 +107,45 @@ export const printReceivedStringContainExpectedResult = (
...
     const [actual, messageOrOptions] = argumentsList;
-    return createExpect(actual, messageOrOptions, false /* isSoft */, false /* isPoll */);
+    const message = isString(messageOrOptions) ? messageOrOptions : messageOrOptions?.message;
+    return createExpect(actual, { message });
   }
 });
 
-expect.soft = (actual: unknown, messageOrOptions: ExpectMessageOrOptions) => {
-  return createExpect(actual, messageOrOptions, true /* isSoft */, false /* isPoll */);
+expect.soft = (actual: unknown, messageOrOptions?: ExpectMessage) => {
```
dgozman, line 124: I wonder whether `expect.soft = expect.configure({ soft: true })` is the same?

packages/playwright-test/src/matchers/expect.ts
```diff
@@ -106,27 +107,45 @@ export const printReceivedStringContainExpectedResult = (
...
+  const message = isString(messageOrOptions) ? messageOrOptions : messageOrOptions?.message;
+  return createExpect(actual, { message, isSoft: true });
 };
 
-expect.poll = (actual: unknown, messageOrOptions: ExpectMessageOrOptions) => {
+expect.poll = (actual: unknown, messageOrOptions?: ExpectMessage & { timeout?: number, intervals?: number[] }) => {
   if (typeof actual !== 'function')
```
dgozman, line 130: I guess this check should now move somewhere else, because you can have `expect.configure({ poll: true })` that would completely avoid this?

packages/playwright-test/src/matchers/expect.ts
```diff
@@ -106,28 +107,57 @@ export const printReceivedStringContainExpectedResult = (
...
+    return expect.configure({ soft: true })(actual, messageOrOptions) as any;
+  };
 
-expect.poll = (actual: unknown, messageOrOptions: ExpectMessageOrOptions) => {
-  if (typeof actual !== 'function')
-    throw new Error('`expect.poll()` accepts only function as a first argument');
-  return createExpect(actual, messageOrOptions, false /* isSoft */, true /* isPoll */, actual as any);
-};
+  expect.poll = (actual: unknown, messageOrOptions?: ExpectMessage & { timeout?: number, intervals?: number[] }) => {
+    if (typeof actual !== 'function')
```
dgozman, line 133: I still think we should move this somewhere deeper, because `configure({ poll })` now misses this logic.

packages/playwright-test/types/test.d.ts
```diff
@@ -4654,6 +4654,12 @@ export type Expect = {
      not: BaseMatchers<Promise<void>, T>;
   };
   extend(matchers: any): void;
+  configure: (configuration: {
+    message?: string,
+    timeout?: number,
+    soft?: boolean,
+    poll?: boolean | { timeout?: number, intervals?: number[] },
+  }) => any;
```
dgozman, line 4662: Forgot the type?

dgozman (approved):

PR #22543, opened 2023-04-21
chore: add step attachments into the expect trace event

pavelfeldman:


dgozman (approved):

packages/trace/src/trace.ts
```diff
@@ -79,6 +79,12 @@ export type AfterActionTraceEvent = {
   afterSnapshot?: string;
   log: string[];
   error?: SerializedError['error'];
+  attachments?: {
+    name: string;
+    contentType: string;
+    path?: string;
+    body?: Buffer;
```
dgozman, line 86: Shouldn't we convert Buffers to strings to store them in json?

dgozman (approved):

PR #22563, opened 2023-04-21
chore: better zone stack preservation in expect

pavelfeldman:


packages/playwright-test/src/matchers/expect.ts
```diff
@@ -267,17 +272,28 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
...
+              await matcher.call(target, ...args);
+            });
+            finalizer();
+          } catch (e) {
+            reportStepError(e);
+          }
+        })();
+      } else {
+        try {
+          zones.run<ExpectZone, any>('expectZone', expectZone, () => {
```
dgozman, line 290: I guess we don't really need the zone here, since we don't run our own matchers?

packages/playwright-test/src/matchers/expect.ts
```diff
@@ -159,6 +159,11 @@ const customMatchers = {
   toPass,
 };
 
+const customMatchers = {
+  ...customAsyncMatchers,
+  toBeOK,
```
dgozman, line 164: toBeOK seems to be async as well.

dgozman (approved):

packages/playwright-test/src/matchers/expect.ts
pavelfeldman: You are right!

PR #22605, opened 2023-04-24
feat: preserve attachments in blob reports

yury-s:
Attachments with `path` are stored as `resources/sha1.ext` zip entries and extracted under `report-dir/tmp` when merging. This way normal fs.readFile keeps working as before even thought the file path is different. The clients should rely on `attachment.name` instead of `attachment.path` when deriving user visible titles in the UI. If this turns out not to be the case we can reconsider later.

#10437

### DIFF
--- a/packages/playwright-test/src/reporters/blob.ts
+++ b/packages/playwright-test/src/reporters/blob.ts
@@ -17,6 +17,7 @@
 import type { EventEmitter } from 'events';
 import fs from 'fs';
 import path from 'path';
+import os from 'os';
 import { ManualPromise, ZipFile, calculateSha1, removeFolders } from 'playwright-core/lib/utils';
 import { mime } from 'playwright-core/lib/utilsBundle';
 import { yazl } from 'playwright-core/lib/zipBundle';
@@ -112,7 +113,7 @@ export class BlobReporter extends TeleReporterEmitter {
 
 export async function createMergedReport(config: FullConfigInternal, dir: string, reporterName?: string) {
   const shardFiles = await sortedShardFiles(dir);
-  const resourceDir = path.join(dir, 'temp');
+  const resourceDir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'playwright-report-'));
   await fs.promises.mkdir(resourceDir, { recursive: true });
   try {
     const shardReports = await extractReports(dir, shardFiles, resourceDir);
@@ -136,7 +137,6 @@ export async function createMergedReport(config: FullConfigInternal, dir: string
     const options = {
       ...arg,
       configDir: process.cwd(),
-      outputFolder: path.join(dir, 'merged-report')
     };
 
     let reporter: Reporter | undefined;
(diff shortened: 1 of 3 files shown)


packages/playwright-test/src/reporters/blob.ts
```diff
@@ -83,66 +112,103 @@ export class BlobReporter extends TeleReporterEmitter {
...
-    list: ListReporter,
-    github: GitHubReporter,
-    json: JSONReporter,
-    junit: JUnitReporter,
-    null: EmptyReporter,
-    html: HtmlReporter,
-    blob: BlobReporter,
-  };
-  reporterName ??= 'list';
+  const resourceDir = path.join(dir, 'temp');
```
dgozman, line 115: Let's do this inside `os.tmpdir()`

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -83,66 +112,103 @@ export class BlobReporter extends TeleReporterEmitter {
...
+        const fileName = path.join(resourceDir, entryName);
+        await fs.promises.mkdir(path.dirname(fileName), { recursive: true });
+        await fs.promises.writeFile(fileName, content);
+      }
+    }
+  }
+  return reports;
+}
+
+function patchAttachmentPaths(events: JsonEvent[], rootDir: string) {
```
dgozman, line 178:
```suggestion
function patchAttachmentPaths(events: JsonEvent[], resourceDir: string) {
```

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -83,66 +112,103 @@ export class BlobReporter extends TeleReporterEmitter {
...
-    reporter = new defaultReporters[reporterName as keyof typeof defaultReporters](options);
-  } else {
-    const reporterConstructor = await loadReporter(config, reporterName);
-    reporter = new reporterConstructor(options);
-  }
+    const arg = config.config.reporter.find(([reporter, arg]) => reporter === reporterName)?.[1];
+    const options = {
+      ...arg,
+      configDir: process.cwd(),
+      outputFolder: path.join(dir, 'merged-report')
```
dgozman, line 139: This should come from the config instead, and it should just work (tm).

packages/playwright-test/src/reporters/blob.ts
```diff
@@ -83,66 +112,103 @@ export class BlobReporter extends TeleReporterEmitter {
...
+    const events = mergeEvents(shardReports);
+    patchAttachmentPaths(events, resourceDir);
 
-  const arg = config.config.reporter.find(([reporter, arg]) => reporter === reporterName)?.[1];
-  const options = {
-    ...arg,
-    configDir: process.cwd(),
-    outputFolder: dir
-  };
+    const defaultReporters: {[key in BuiltInReporter]: new(arg: any) => Reporter} = {
```
dgozman, line 122: Let's try to reuse `createReporter` function.

dgozman (approved):

packages/playwright-test/src/reporters/blob.ts
yury-s: I'll try to do that in a follow-up as agreed offline.

PR #22607, opened 2023-04-24
devops: bump Node.js versions 16->18

### LINKED ISSUE #22582: [Bug] add Node 20 to CI matrix
Node 20 is the latest stable release of Node. As a result, we're trying to add Node 20 to our CI matrix. However, it is causing our Playwright tests to fail in an unexpected and difficult to diagnose manner. I would like if Playwright itself were tested on Node 20 to give me some confidence in whether this might be an issue in Playwright or our own code.

mxschmitt:
Its like https://github.com/microsoft/playwright/pull/20800 2 months ago.

Fixes https://github.com/microsoft/playwright/issues/22582.

### DIFF
--- a/.github/workflows/infra.yml
+++ b/.github/workflows/infra.yml
@@ -21,7 +21,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-node@v3
       with:
-        node-version: 16
+        node-version: 18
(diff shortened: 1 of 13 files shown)


mxschmitt:
Discussed offline, we stay for a bit more on 16 everywhere until the majority moves to 18 LTS.

Issue #22576, opened 2023-04-23
[BUG] Error is thrown when trying to save a file that has some special characters in its name


ahmed-0011:
### System info
- Playwright Version: v1.32.3
- Operating System: Windows 10
- Browser: All
- Other info: Node v18.16.0

### Source code

```js
const { firefox, selectors } = require("playwright");

const URL = "https://unsplash.com/backgrounds";
const DIR = "C:\\Users\\MyUser\\Desktop";

(async () => {
    const browser = await firefox.launch({ headless: false, downloadsPath: DIR });
    const context = await browser.newContext({ acceptDownloads: true });
    const page = await context.newPage();
    await page.goto(URL, { timeout: 500000 });
    await page.waitForLoadState("domcontentloaded");

    const images = await page.locator("img").all();

    const firstImage = await images[0].getAttribute("src");

    await page.goto(firstImage);
    await page.waitForLoadState("networkidle");

    await page.evaluate(async firstImage => {
        const link = document.createElement("a");
        link.href = firstImage;
        link.download = "";
        link.textContent = "Download file";
        document.body.appendChild(link);
    }, firstImage);

    const downloadPromise = page.waitForEvent("download");
    await page.getByText("Download file").click();
    const download = await downloadPromise;
    await download.saveAs(`${DIR}\\image\?.jpg`);

    await browser.close();
})();
```

**Expected**
The file should be downloaded with no problem, at least when escaping the special character * and ?
other special characters such as [ and { seems to work just fine without even escaping by a backslash.

**Actual**
Browser terminates and the following error is thrown:

```js
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

download.saveAs: ENOENT: no such file or directory, copyfile 'C:\Users\MyUser\Desktop\5a59b49a-ec6f-46c2-bc07-34a0d2403442' -> 'C:\Users\MyUser\Desktop\image?.jpg'
    at C:\Users\MyUser\Desktop\test.js:31:20 {
  name: 'Error'
}
```

aslushnikov:
@ahmed-0011 so it looks like on Windows you cannot have files with question mark in name; see "naming conventions" in https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

You'll have to just avoid using question marks in your file names. If you have some automation, than you should make your file name fs-safe by replacing `?` and other restricted characters with some other ones. 

Hope this helps!

PR #22619, opened 2023-04-25
test: add regression test

aslushnikov:
References: https://github.com/microsoft/playwright/issues/22576


aslushnikov:
Turns out this works as expected.

Issue #22596, opened 2023-04-24
[BUG] Xray-Report: Tests within a skipped test group unnecessarily create new Xray test issues


theokoester:
### System info
- Playwright Version: [v1.32.3]
- Operating System: [All]
- Other info: GitLab pipeline

### Source code

For importing the results from my GitLab pipeline I use the extended Junit reporter which was fine so far.
All tests do have an test_key annotation with an already existing Test-Issue in Xray. 

Config according to [docs](https://playwright.dev/docs/test-reporters#junit-reporter)
```js
// playwright.config.ts
import { defineConfig, devices } from '@playwright/test';

const xrayOptions = {
    textContentAnnotations: ['test_description'],
    outputFile: './xray-report.xml'
};

export default defineConfig({
    reporter: [['junit', xrayOptions]]
});
```

**Test file**

```js
test.describe.skip('ui tests @gui', () => {
    test('Some test', async ({ page }) => {
        test.info().annotations.push({ type: 'test_key', description: 'ABCD-166' })
        // ..test code
    })
})
```

**Expected**

When the group is being skipped, all the contained tests are transferred with status ToDo to an Xray execution and no new test issues are created. No creation because of the existing test_key annotation and already existing corresponding test issues in Xray is the expected [flow](https://docs.getxray.app/display/XRAY/Working+with+Automated+Tests)

**Actual**

All the tests within an skipped test.describe() group unnecessarily generate **new** test issues in Xray and an execution with run status ToDo. The existing test issues are being ignored instead of using them for attaching and linking the execution results, in this case *ToDo*.

pavelfeldman:
We don't really support XRay. If you have reasons to believe JUnit that we generate is invalid, please file a specific bug!

aslushnikov:
Closing this for now; given the discussion in https://github.com/microsoft/playwright/pull/22624, a dedicated xray support might be added sometime in future.

Issue #18379, opened 2022-10-27
[Feature] JUnit reporter: optionally embed links instead of encoded files for attachments.


raulparada:
A bit of background on https://github.com/microsoft/playwright/issues/9348#issuecomment-1293387867

The JUnit report generated with embedded attachments (option `embedAttachmentsAsProperty: "testrun_evidence"`, described on https://playwright.dev/docs/test-reporters#junit-reporter, thought mostly for Xray integration) easily creeps in the hundreds of MBs for regular runs with a few failures. Though one may try to keep artifacts to a minimum, it's difficult to limit this, and the size will likely affect the transport, import, usage, etc. of the report.

It would be ideal having an **option** to include URLs/paths to the attachments **instead** of the base64-encoded version of them. If done, further having an easy way to pass a base URL to be prepended to each attachment URL would be imo very user-friendly. This would allow Test Management frameworks (e.g. Xray) to consume light, portable reports that link to this evidence where stored on a CI server.

Not aware of appropriate workarounds using annotations.

pavelfeldman:
#### Why was this issue closed?

Thank you for your involvement. This issue was closed due to limited engagement (upvotes/activity), lack of recent activity, and insufficient actionability. To maintain a manageable database, we prioritize issues based on these factors.

If you disagree with this closure, please open a new issue and reference this one. More support or clarity on its necessity may prompt a review. Your understanding and cooperation are appreciated.

PR #22624, opened 2023-04-25
chore: purify the junit reporter

pavelfeldman:
https://github.com/microsoft/playwright/pull/11374 introduced a number of configuration options for JUnit reporter that was targeting a better alignment with the [Xray JUnit dialect](https://docs.getxray.app/display/XRAYCLOUD/Taking+advantage+of+JUnit+XML+reports).

We are seeing more Xray-specific requests to JUnit:
- https://github.com/microsoft/playwright/issues/22396
- https://github.com/microsoft/playwright/issues/22596

As well as the requests to counter the Xray changes:
- https://github.com/microsoft/playwright/issues/18379

When looking at https://github.com/microsoft/playwright/issues/22396, it is clear that we are setting up Xray user to failure: instead of generating a full-fledged Xray report, we are going through an intermediate JUnit format where we are hacking information in a lossy manner. What's worse is that we are polluting JUnit with Xray-specific options that pretend to be generic ones, without those actually being requested by the JUnit customers.

I'm suggesting that we prune the Xray-specificity from JUnit and allow Xray team or third parties to implement a full Xray reporter for Playwright.


pavelfeldman:
// cc @bitcoder

bitcoder:
Hi, sorry to hear that but I totally understand and respect your decision.
We'll need a bit of time to move this to a separate reporter. To make it easier right now, maybe the best is to replicate the current behaviour along with the respective code with minor tweaks if needed.
I'll let you know about the progress.

pavelfeldman:
> Hi, sorry to hear that but I totally understand and respect your decision. We'll need a bit of time to move this to a separate reporter. To make it easier right now, maybe the best is to replicate the current behaviour along with the respective code with minor tweaks if needed. I'll let you know about the progress.

I won't be landing this for now and will ping you in a couple of weeks!

bitcoder:
@pavelfeldman can I kindly ask you to quickly review the following project?
I created it using mostly the existing code and did some subtle changes to make it compile and have the tests passing.
I kept the license notice on the files copied from this project.
It's not something perfect but I think it works but I appreciate if you can give it a look

https://github.com/Xray-App/playwright-junit-reporter

pavelfeldman:
> Would it be possible to maintain the API that supports adding these properties in the JUnit reporter? I.e.:

@d-rede, that makes sense, so we would make it work like `embedAnnotationsAsProperties` is true by default. Are you also interested in attachments or annotations only?

packages/playwright-test/src/reporters/junit.ts
```diff
@@ -31,17 +31,10 @@ class JUnitReporter implements Reporter {
   private totalSkipped = 0;
   private outputFile: string | undefined;
   private stripANSIControlSequences = false;
-  private embedAnnotationsAsProperties = false;
-  private textContentAnnotations: string[] | undefined;
-  private embedAttachmentsAsProperty: string | undefined;
 
-
-  constructor(options: { outputFile?: string, stripANSIControlSequences?: boolean, embedAnnotationsAsProperties?: boolean, textContentAnnotations?: string[], embedAttachmentsAsProperty?: string } = {}) {
+  constructor(options: { outputFile?: string, stripANSIControlSequences?: boolean, embedAnnotationsAsProperties?: boolean, textContentAnnotations?: string[] } = {}) {
```
dgozman, line 35:
```suggestion
  constructor(options: { outputFile?: string, stripANSIControlSequences?: boolean } = {}) {
```

dgozman (approved):

bitcoder:
> @pavelfeldman can I kindly ask you to quickly review the following project? I created it using mostly the existing code and did some subtle changes to make it compile and have the tests passing. I kept the license notice on the files copied from this project. It's not something perfect but I think it works but I appreciate if you can give it a look
> 
> https://github.com/Xray-App/playwright-junit-reporter

Did you have opportunity to give it a look? Would appreciate your feedback :)

pavelfeldman:
I looked briefly and it seemed fine!

PR #22636, opened 2023-04-26
chore: reuse createReporters in the merger

yury-s:
#10437

### DIFF
--- a/packages/playwright-test/src/runner/reporters.ts
+++ b/packages/playwright-test/src/runner/reporters.ts
@@ -25,13 +25,13 @@ import JSONReporter from '../reporters/json';
 import JUnitReporter from '../reporters/junit';
 import LineReporter from '../reporters/line';
 import ListReporter from '../reporters/list';
-import { Multiplexer } from '../reporters/multiplexer';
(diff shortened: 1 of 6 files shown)


pavelfeldman (approved):

packages/playwright-test/src/runner/reporters.ts
```diff
@@ -44,7 +44,10 @@ export async function createReporter(config: FullConfigInternal, mode: 'list' |
     blob: BlobReporter,
   };
   const reporters: Reporter[] = [];
-  for (const r of config.config.reporter) {
+  const descriptions: ReporterDescription[] = reporterNames ?
```
dgozman, line 47: What about something like this at line 51.5?

```
if (reporterNames && !reporterNames.includes(name)) continue;
```

dgozman (approved):

packages/playwright-test/src/runner/reporters.ts
yury-s: This won't work, some of `reporterNames` may not be in the config at all and we still want to create corresponding reports.

PR #22651, opened 2023-04-26
chore: test on Node.js 20

### LINKED ISSUE #22582: [Bug] add Node 20 to CI matrix

mxschmitt:
As per offline discussion we keep the majority of our bots on Node.js 16 for now, since thats where most of our users are.

This patch does remove Node.js 14 bots, since 14 goes EOL before our next release, see [here](https://github.com/nodejs/Release#release-schedule) and adds Node.js 20 bots.

Node.js 20 tests are failing since https://github.com/nodejs/node/commit/5ec0f39a7a565b5a82fe90ba9f095731a7b8b005 so we had to adjust TestServer a bit.

Fixes https://github.com/microsoft/playwright/issues/22582

### DIFF
--- a/.github/workflows/tests_secondary.yml
+++ b/.github/workflows/tests_secondary.yml
@@ -102,9 +102,9 @@ jobs:
       matrix:
         include:
         - os: ubuntu-latest
-          node_version: "^18.0.0"
+          node_version: 18
         - os: ubuntu-latest
-          node_version: "^16.0.0"
+          node_version: 20
     timeout-minutes: 30
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/tests_components.yml
+++ b/.github/workflows/tests_components.yml
@@ -29,7 +29,6 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-node@v3
       with:
-        # Component tests require Node.js 16+ (they require ESM via TS)
         node-version: 16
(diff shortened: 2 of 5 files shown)


dgozman (approved):

benmccann:
Just FYI, the SvelteKit CI is failing on Node 20.0.0, but one of the maintainers figured out it works with the latest nightly release, so perhaps the Node 20 tests here will start passing after the next release of Node 20.

mxschmitt:
Node.js 20 tests are failing since https://github.com/nodejs/node/compare/2dd6d76c89...cdd20cfd71.

.github/workflows/tests_secondary.yml
```diff
@@ -102,9 +102,9 @@ jobs:
       matrix:
         include:
         - os: ubuntu-latest
-          node_version: "^18.0.0"
+          node_version: 18
         - os: ubuntu-latest
-          node_version: "^16.0.0"
```
mxschmitt, line 107: note: 16 is already tested in tests_primary

dgozman (approved):

PR #22662, opened 2023-04-26
chore: make client-side instrumentation non-nullable

pavelfeldman:


packages/playwright-core/src/client/fetch.ts
```diff
@@ -80,7 +77,7 @@ export class APIRequest implements api.APIRequest {
     this._contexts.add(context);
     context._request = this;
     context._tracing._tracesDir = tracesDir;
-    await this._onDidCreateContext?.(context);
+    await this._playwright._instrumentation.onDidCreateRequestContext(context);
```
dgozman, line 80: To be similar to other places:
```suggestion
    await this._instrumentation.onDidCreateRequestContext(context);
```

packages/playwright-core/src/client/page.ts
```diff
@@ -691,6 +691,9 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
   async pause() {
     if (require('inspector').url())
       return;
+    this._browserContext.setDefaultNavigationTimeout(0);
+    this._browserContext.setDefaultTimeout(0);
```
dgozman, line 695: I wonder if we should restore the timeout back, now that this code triggers with just a library?

packages/playwright-test/src/index.ts
```diff
@@ -361,46 +370,26 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
...
-      (browserType as any)._onWillCloseContext = onWillCloseContext;
       (browserType as any)._defaultContextOptions = _combinedContextOptions;
       const promises: Promise<void>[] = [];
       const existingContexts = Array.from((browserType as any)._contexts) as BrowserContext[];
       for (const context of existingContexts) {
         if ((context as any)[kIsReusedContext])
           reusedContexts.add(context);
         else
-          promises.push(onDidCreateBrowserContext(context));
+          promises.push((playwright as any)._instrumentation.onDidCreateBrowserContext(context));
```
dgozman, line 384:
```suggestion
          promises.push(csiListener.onDidCreateBrowserContext(context));
```

packages/playwright-test/src/index.ts
```diff
@@ -361,46 +370,26 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
...
       await Promise.all(promises);
     }
     {
-      (playwright.request as any)._onDidCreateContext = onDidCreateRequestContext;
-      (playwright.request as any)._onWillCloseContext = onWillCloseRequestContext;
       (playwright.request as any)._defaultContextOptions = { ..._combinedContextOptions };
       (playwright.request as any)._defaultContextOptions.tracesDir = path.join(_artifactsDir(), 'traces');
       const existingApiRequests: APIRequestContext[] =  Array.from((playwright.request as any)._contexts as Set<APIRequestContext>);
-      await Promise.all(existingApiRequests.map(onDidCreateRequestContext));
+      await Promise.all(existingApiRequests.map(c => (playwright as any)._instrumentation.onDidCreateRequestContext(c)));
```
dgozman, line 392:
```suggestion
      await Promise.all(existingApiRequests.map(c => csiListener.onDidCreateRequestContext(c)));
```

packages/playwright-test/src/index.ts
```diff
@@ -361,46 +370,26 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
...
-      await stopTracing(tracing, (context as any)[kStartedContextTearDown]);
-    };
-
     // 1. Setup instrumentation and process existing contexts.
+    const instrumentation = (playwright as any)._instrumentation as ClientInstrumentation;
+    instrumentation.addListener(csiListener);
     for (const browserType of [playwright.chromium, playwright.firefox, playwright.webkit]) {
-      (browserType as any)._onDidCreateContext = onDidCreateBrowserContext;
-      (browserType as any)._onWillCloseContext = onWillCloseContext;
       (browserType as any)._defaultContextOptions = _combinedContextOptions;
```
dgozman, line 377: Perhaps we can migrate this one as well, to something like `onWillCreateContext` or explicit `modifyNewContextOptions`?

packages/playwright-core/src/client/clientInstrumentation.ts
```diff
@@ -15,17 +15,30 @@
...
+  onWillCloseBrowserContext(context: BrowserContext): Promise<void>;
+  onWillCloseRequestContext(context: APIRequestContext): Promise<void>;
 }
 
 export interface ClientInstrumentationListener {
-  onApiCallBegin?(apiCall: string, stackTrace: ParsedStackTrace | null, wallTime: number, userData: any): any;
-  onApiCallEnd?(userData: any, error?: Error): any;
+  onApiCallBegin?(apiCall: string, stackTrace: ParsedStackTrace | null, wallTime: number, userData: any): void;
+  onApiCallEnd?(userData: any, error?: Error): void;
+  onDidCreateBrowserContext?(context: BrowserContext): Promise<void>;
```
dgozman, line 37: I think the PR title is now a bit off 😄

dgozman (approved):

packages/playwright-core/src/client/fetch.ts
pavelfeldman: This unfortunately won't compile. You probably mean context._instrumentation. But it won't be similar.

packages/playwright-core/src/client/page.ts
pavelfeldman: We could, we don't have means to doing that though, also does not seem to be important?

packages/playwright-test/src/index.ts
pavelfeldman: I wanted to follow up

packages/playwright-core/src/client/clientInstrumentation.ts
pavelfeldman: yeah

PR #22665, opened 2023-04-26
feat(test runner): implement beforeAll/afterAll as tests

dgozman:
We store `beforeAll`/`afterAll` hook as a `TestCase`, and handle it differently in a few places.
1. `TestGroup` only contains proper tests, which leads to:
    - Dispatcher only schedules proper tests.
    - Sharding only counts proper tests.
2. Filtering by `only`, `grep`, `shard`, `testId` and `file:line` ignore hooks.
3. Worker automatically runs relevant hooks when asked to run a `TestGroup`.

Remaining issues:
- Figure out whether we need any reporter changes.
- `should report error from beforeAll timeout` test fails.
- Filtering by `testId` and `file:line` are not implemented for hooks.
- Perhaps we need `TestCase.kind` in the reporter api.

Note that #12518 was moving in the opposite direction, but hooks were completely hidden from the reporter api.


packages/playwright-test/src/common/suiteUtils.ts
```diff
@@ -27,17 +27,18 @@ export function filterSuite(suite: Suite, suiteFilter: (suites: Suite) => boolea
...
 
 export function filterTestsRemoveEmptySuites(suite: Suite, filter: (test: TestCase) => boolean): boolean {
   const filteredSuites = suite.suites.filter(child => filterTestsRemoveEmptySuites(child, filter));
-  const filteredTests = suite.tests.filter(filter);
+  const filteredTests = suite.tests.filter(test => test._kind === 'test' && filter(test));
   const entries = new Set([...filteredSuites, ...filteredTests]);
-  suite._entries = suite._entries.filter(e => entries.has(e)); // Preserve the order.
-  return !!suite._entries.length;
+  const hooks = new Set<TestCase | Suite>(suite.tests.filter(test => test._kind !== 'test'));
+  suite._entries = suite._entries.filter(e => entries.has(e) || hooks.has(e)); // Preserve the order.
```
pavelfeldman, line 40: Let's also remove the hooks from the filtered out suites.

PR #22695, opened 2023-04-27
chore: split merge and blob reporter

yury-s:


packages/playwright-test/src/reporters/blob.ts
```diff
@@ -16,21 +16,15 @@
...
+import { ManualPromise, calculateSha1 } from 'playwright-core/lib/utils';
 import { mime } from 'playwright-core/lib/utilsBundle';
 import { yazl } from 'playwright-core/lib/zipBundle';
 import { Readable } from 'stream';
 import type { FullConfig, FullResult, TestResult } from '../../types/testReporter';
-import type { FullConfigInternal } from '../common/config';
 import type { Suite } from '../common/test';
-import { TeleReporterReceiver, type JsonEvent, type JsonProject, type JsonSuite, type JsonTestResultEnd } from '../isomorphic/teleReceiver';
-import { createReporters } from '../runner/reporters';
 import { defaultReportFolder } from './html';
```
dgozman, line 26: Why do we use report folder of the html report here? Sounds strange.

dgozman (approved):

Issue #22700, opened 2023-04-28
[BUG] HTML report - unable show the tag with dash


xyngfei:
### System info
- Playwright Version: v.1.30.0
- Operating System: macOS 13.2.1
- Browser: All
- Other info:

### Source code

NA

**Test file (self-contained)**

```js
test("'@GCC-1505'|As the requestor, able to create a new tenant account successfully", async ({ }) => {
    console.log("@GCC-1505")
  });
test("@GCC-1508 @api|As the CIO, I want to REJECT the tenant creation request", async ({ }) => {
    console.log("@GCC-1508")
  });
```

**Steps**
- define a tag with dash '-'

**Expected**

tag label should reflect together with '-', which is @GCC-1505, rather @gcc

**Actual**
<img width="823" alt="image" src="https://user-images.githubusercontent.com/17619539/235033681-cbb1864b-581c-4674-9791-3396b4332ee9.png">

PR #22709, opened 2023-04-28
feat(html): matching labels with dash&spec symbol

alexneo2003:
related to #22700 

https://user-images.githubusercontent.com/11705902/235174845-bceaaa07-7ea2-4c9f-839b-041bdd21cd6d.mp4

### DIFF
--- a/packages/html-reporter/src/labelUtils.tsx
+++ b/packages/html-reporter/src/labelUtils.tsx
@@ -25,7 +25,7 @@ export function escapeRegExp(string: string) {
 
 // match all tags in test title
 export function matchTags(title: string): string[] {
-  return title.match(/@(\w+)/g)?.map(tag => tag.slice(1)) || [];
+  return title.match(/@([\w-]+)/g)?.map(tag => tag.slice(1)) || [];
 }
 
 // hash string to integer in range [0, 6] for color index, to get same color for same tag
--- a/packages/html-reporter/src/labelUtils.tsx
+++ b/packages/html-reporter/src/labelUtils.tsx
@@ -25,7 +25,7 @@ export function escapeRegExp(string: string) {
 
 // match all tags in test title
 export function matchTags(title: string): string[] {
-  return title.match(/@([\w-]+)/g)?.map(tag => tag.slice(1)) || [];
+  return title.match(/@([\S]+)/g)?.map(tag => tag.slice(1)) || [];
 }
 
 // hash string to integer in range [0, 6] for color index, to get same color for same tag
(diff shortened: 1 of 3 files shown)


packages/html-reporter/src/labelUtils.tsx
```diff
@@ -25,7 +25,7 @@ export function escapeRegExp(string: string) {
 
 // match all tags in test title
 export function matchTags(title: string): string[] {
-  return title.match(/@(\w+)/g)?.map(tag => tag.slice(1)) || [];
+  return title.match(/@([\w-]+)/g)?.map(tag => tag.slice(1)) || [];
```
pavelfeldman, line 28: what about underscore?

alexneo2003: `\w` - matches any word characters, including underscore `_`
`-` - is exceptional character what using somebody for tags
`# $ % ^ & * + = < > |` - are more special symbols, and i think what almost nobody not using it for tags
If tag will containing special symbol it should be escaped by search engine and no longer work

pavelfeldman: What is the conceptual difference between `-` and `+` when it comes to the tab value, why do you want to treat them differently?

alexneo2003: `[\w-]+` - it means
`\w` - any one of [a-zA-Z0-9_]
`-` - one of `-`
`+` - as much as possible
It match like `@CBC-123`, `@flaky`
Or I'm not understand you what you want)
I offered such a solution to the "problem", you have a different vision of it.  we still won't be able to cover all possible options) but we should try to do so.  and it is necessary to exclude options that, on the contrary, can harm.

pavelfeldman: I am suggesting to be more lax in matching the tags. Clearly we have already underestimated the expressiveness of the tags, hence this follow up. So no considerations like "more special" stand at this point. I'm thinking about tags like:

`@smoke:p1`
`@issue[123]`
`@issue#123`
`@$$$`
`@tl/dr`

etc.

Hence my question of why not treat any uninterrupted sequence as a tag and minimize the number of follow up patches like this one.

pavelfeldman: (backslash escaping space-separated tags would also be fine)

alexneo2003: i'm try using `escapeRegExp` when matching tags from test title but after they looks like `@issue\[123\]`, `@\$\$\$` and after be clicked this string not matching anything


https://user-images.githubusercontent.com/11705902/235366845-6139f600-8045-41f6-9c8b-72d6ea68a96a.mp4

pavelfeldman (approved):

Issue #22328, opened 2023-04-11
[BUG] "Cannot read properties of null (reading '$options')" in Vue 2 Component Tests


oliver-freudrich-packwise:
Hey there,
I've got an issue since the 1.32. Currently we have the 1.30.0 and there everything works fine. I wanted to upgrade and now the `component.update` logic fails in vue 2 component tests.

### System info
- Playwright Version: v1.32.3
- Previous working Playwright Version: v1.30.0
- Operating System: macOS 13.3
- Browser: All
- Other info:
- @playwright/experimental-ct-vue2: v1.32.3

**Config file**

```js
// playwright.config.ts
import { devices, defineConfig } from '@playwright/experimental-ct-vue2';
import viteConfig from "./vite.config";

export default defineConfig({
    testDir: 'src',
    testMatch: /.*\.spec\.[t|j]sx/,
    expect: {
        timeout: 15 * 1000,
    },
    timeout: 30 * 1000,
    globalTimeout: 30 * 60 * 1000,
    forbidOnly: true,
    workers: 1,
    use: {
        viewport: { width: 1920, height: 1080 },
        trace: 'on',
        video: {
            mode: 'on',
            size: { width: 1920, height: 1080 },
        },
        screenshot: 'on',
        ignoreHTTPSErrors: true,
        ctPort: 10443,
        ctViteConfig: viteConfig,
    },
    globalSetup: './playwright/setup.js',
    globalTeardown: './playwright/teardown.js',
    reporter: [['html', { open: 'never' }]],
    fullyParallel: false,
    projects: process.env.CI
        ? [
            {
                name: 'component-chromium',
                use: {
                    ...devices['Desktop Chrome'],
                },
            },
            {
                name: 'component-firefox',
                use: {
                    ...devices['Desktop Firefox'],
                },
            },
            {
                name: 'component-webkit',
                use: {
                    ...devices['Desktop Safari'],
                },
            },
        ]
        : undefined,
});
```

**Test file (self-contained)**

```js
test('quick test', async ({ page, mount }) => {
  const component = await mount(<div id="test" />);
  await component.update(<div id="test2" />);
});
```

**Steps**
- Run the test (ran it with the --ui flag)
- Errors while updating the component

**Expected**

Test should pass without any issues.

**Actual**

[...]

PR #22712, opened 2023-04-28
test(ct): update child

sand4rt:
partial fix for: https://github.com/microsoft/playwright/issues/22328

Updating child props, slots or events in vue2 and vue3 does not work yet. Added the tests for the frameworks where it already worked and i'm planning to reuse those tests for vue2 and vue3.

### DIFF
--- a/tests/components/ct-react-vite/src/components/Counter.tsx
+++ b/tests/components/ct-react-vite/src/components/Counter.tsx
@@ -18,8 +18,8 @@ import { useLayoutEffect, useRef, useState } from "react"
      }
    }, [])
    return <button onClick={() => props.onClick?.('hello')}>
-     <div data-testid="props">{ props.count }</div>
-     <div data-testid="remount-count">{ remountCount }</div>
+     <span data-testid="props">{ props.count }</span>
+     <span data-testid="remount-count">{ remountCount }</span>
      { props.children }
    </button>
  }
--- a/tests/components/ct-react-vite/src/components/Counter.tsx
+++ b/tests/components/ct-react-vite/src/components/Counter.tsx
@@ -17,9 +17,9 @@ import { useLayoutEffect, useRef, useState } from "react"
        _remountCount++;
      }
    }, [])
-   return <div onClick={() => props.onClick?.('hello')}>
+   return <button onClick={() => props.onClick?.('hello')}>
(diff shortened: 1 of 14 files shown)


tests/components/ct-react-vite/src/components/Counter.tsx
```diff
@@ -17,9 +17,9 @@ import { useLayoutEffect, useRef, useState } from "react"
        _remountCount++;
      }
    }, [])
-   return <div onClick={() => props.onClick?.('hello')}>
+   return <button onClick={() => props.onClick?.('hello')}>
```
pavelfeldman, line 20: button with divs inside does not sound right - what are you trying to do here?

sand4rt: Yeah agree, i can change it to `span` instead.

Svelte complained about having a click event on a `div` so i changed everything to `button`. This also worked out well for the added tests.

pavelfeldman: I'm not entirely sure where this is coming from (this does not sound like the right a11y advice to me). Maybe add a role="button" on your fix to make a11y tree happy?

sand4rt: The warning is not that clear but it is correct. I believe it's about having a click event on divs and divs can't be triggered with keyboard events like `enter`. Adding a `role="button"` does not solve this problem.

Is there something wrong with a `span` in a `button`?

By the way, the a11y warning can be seen here: https://github.com/microsoft/playwright/actions/runs/4817238735/jobs/8577703732?pr=22678#step:8:360

pavelfeldman: > I believe it's about having a click event on divs and divs can't be triggered with keyboard events like enter. Adding a role="button" does not solve this problem.

Yes, but that's a fairly naive take on a11y. Can these warnings be disabled?

> Is there something wrong with a span in a button?

Either way is acceptable, div was a bit unusual. The meta-point is: you are modifying the code due to the audit warning generated by the third party, implying that you trust the warning. But that warning does not sound right to me (and I've spent some time dealing with a11y). Let alone that our code is for testing. So I'd just disable it.

PR #22718, opened 2023-04-29
fix: call onExit when merging reports

yury-s:


packages/playwright-test/src/reporters/merge.ts
```diff
@@ -93,11 +93,13 @@ function mergeEvents(shardReports: string[]) {
         beginEvents.push(event);
       else if (event.method === 'onEnd')
         endEvents.push(event);
+      else if (event.method === 'onExit')
+        continue;
```
pavelfeldman, line 97: Waht was the point of plumbing if it is not handled?

yury-s: To have complete report with all the events but if you think it's better to just special case it in the merger I can change.

dgozman (approved):

PR #22741, opened 2023-05-01
chore: unflake ui tests (3)

pavelfeldman:


utils/limits.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+sudo sysctl -w fs.inotify.max_user_instances=1024
```
mxschmitt, line 2: Seems unused?

dgozman (approved):

utils/limits.sh
pavelfeldman: Very much used when running locally on Ubuntu with > 50 cores!

PR #22750, opened 2023-05-02
fix: React selectors should work after unmount

### LINKED ISSUE #22729: [BUG] any react locators stop working after any react 18 root unmounted
### System info
- Operating System: [All, Windows 10.]
- Browser: [All]
- Other info: React 18.2

### Source code

```
<!DOCTYPE html>
<html lang="en" xmlns="http://www.w3.org/1999/html" xmlns="http://www.w3.org/1999/html">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <meta name="theme-color" content="#000000" />
    <meta
      name="description"
      content="Web site"
    />
  </head>
  <body>
    <div id="root1"></div>
    <div id="root2"></div>
    <script src="script-bellow.js"></script>
  </body>
</html>
```

```
import {createRoot} from "react-dom/client";

function MyReactComponent() {
    return <div></div>;
}

const container1 = document.getElementById("root1");
const root1 = createRoot(container1);
root1.render(<MyReactComponent />);

root1.unmount(); // important

const container2 = document.getElementById("root2");
const root2 = createRoot(container2);
root2.render(<MyReactComponent />);
```

```
test("mycase", async ({page}) => {
  await page.goto("http://localhost:3000/");
  await page.locator("_react=MyReactComponent").click();
})
```

**Config file**
default

**Steps**

- [Run the test]

**Expected**

`MyReactComponent` of the `root2` receives a click.

**Actual**

Test fails with error:
```
    Error: locator.click: TypeError: null is not an object (evaluating 'node[rootKey].stateNode')
    waiting for locator('_react=MyReactComponent')
```

After React unmounts at a DOM element that DOM element still contains property with `__reactContainer` prefix but it's value [becomes](https://github.com/facebook/react/blob/0b931f90e8964183f08ac328e7350d847abb08f9/packages/react-dom-bindings/src/client/ReactDOMComponentTree.js#L77) `null`. So `findReactRoots` fails [here](https://github.com/microsoft/playwright/blob/cffb6ac269ab97ac63ac98d88244603e3349b793/packages/playwright-core/src/server/injected/reactSelectorEngine.ts#L155).

As a result usage of any React locator throws an error.

alexandersorokin:
Fixes #22729

### DIFF
--- a/packages/playwright-core/src/server/injected/reactSelectorEngine.ts
+++ b/packages/playwright-core/src/server/injected/reactSelectorEngine.ts
@@ -147,16 +147,20 @@ function findReactRoots(root: Document | ShadowRoot, roots: ReactVNode[] = []): 
   do {
     const node = walker.currentNode;
 
+    const reactNode = node as { readonly [customKey: string]: any };
     // React 17+
     // React sets rootKey when mounting
     // @see https://github.com/facebook/react/blob/a724a3b578dce77d427bef313102a4d0e978d9b4/packages/react-dom/src/client/ReactDOMComponentTree.js#L62-L64
-    const rootKey = Object.keys(node).find(key => key.startsWith('__reactContainer'));
+    const rootKey = Object.keys(reactNode).find(key => key.startsWith('__reactContainer') && reactNode[key] !== null);
     if (rootKey) {
-      roots.push((node as any)[rootKey].stateNode.current);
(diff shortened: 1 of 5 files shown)


alexandersorokin:
@microsoft-github-policy-service agree company="AO SKB Kontur"

aslushnikov (approved):

PR #22773, opened 2023-05-03
fix(steps): step should not get an unrelated error

dgozman:
Previously, we would use any error that was added during the step execution as an error for this particular step.

This produces false positives, for example failing `page.click` call that happened during `context` teardown was producing an error and marking teardown is failed. However, in reality, the test itself has failed, while teardown has not.

New approach uses test step hierarchy to inherit errors from child steps to the parent step. This does not regress the original fix where `expect.soft` errors are surfaced in the parent step.

See also #19973 that introduced the original logic.

### DIFF
--- a/tests/playwright-test/test-step.spec.ts
+++ b/tests/playwright-test/test-step.spec.ts
@@ -720,3 +720,117 @@ test('should nest steps based on zones', async ({ runInlineTest }) => {
     }
   ]);
 });
+
+test('should not mark page.close as failed when page.click fails', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'reporter.ts': stepHierarchyReporter,
+    'playwright.config.ts': `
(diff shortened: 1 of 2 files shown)


tests/playwright-test/test-step.spec.ts
```diff
@@ -720,3 +720,117 @@ test('should nest steps based on zones', async ({ runInlineTest }) => {
...
+  const result = await runInlineTest({
+    'reporter.ts': stepHierarchyReporter,
+    'playwright.config.ts': `
+      module.exports = {
+        reporter: './reporter',
+      };
+    `,
+    'a.test.ts': `
+      import { test, expect } from '@playwright/test';
+      let page: Page;
```
yury-s, line 734: nit: should Page be imported?

yury-s (approved):

Issue #22640, opened 2023-04-26
[BUG] Reload pushes a new history entry on Firefox


gtm-nayan:
### System info
- Playwright Version: v1.32.3
- Operating System: Windows 10, Ubuntu 22
- Browser: Firefox
- Other info: The reload button is broken on older versions as well, specifically on 1.29.2, calling `await page.reload()` behaves as expected but manually pressing the reload button in debug mode causes a similar bug.

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Config file**

```js
// playwright.config.ts
import { defineConfig, devices } from '@playwright/test';

export default defineConfig({
  projects: [
    {
      name: 'firefox',
      use: { ...devices['Desktop Firefox'], },
    },
});
```

**Test file (self-contained)**

```js
it('should reload properly', async ({ page }) => {
	await page.goto('https://www.iana.org/domains/reserved');
	await page.mouse.wheel(0, 200);

	const history = await page.evaluate('history.length');
	await page.pause(); // back button disabled
	await page.reload(); 
	await page.pause(); // back button is enabled at this point 
	expect(await page.evaluate('history.length')).toBe(history); // new entry to the same page pushed

	expect(await page.evaluate('scrollY')).toBe(200); // scroll is reset to top
});
```

**Steps**
- Run the test

**Expected**
Scroll is restored, history remains same as in the consumer versions of firefox or other PW browsers.

**Actual**
New history entry with null state is pushed on. Scroll is reset to the top.

aslushnikov:
For the record: this is bisected to the `pref("fission.webContentIsolationStrategy", 0);` preference that we use today.

aslushnikov:
Filed upstream: https://bugzilla.mozilla.org/show_bug.cgi?id=1832341

benmccann:
Thanks for tracking that down! Would it be possible to remove the use of that preference until the issue is resolved upstream?

aslushnikov:
> Would it be possible to remove the use of that preference until the issue is resolved upstream?

@benmccann I don't think so: without this preference, we won't be able to interact with out-of-process iframes in firefox.

If you're sure you don't have any OOPIFs in your tests, you can try overriding this preference with another value using `firefoxUserPrefs` option in `firefox.launch()` method.

benmccann:
Thanks for the suggestion @aslushnikov. However, it looks like https://github.com/microsoft/playwright/issues/15405 probably needs to be addressed before that's a possibility.

[...]

PR #22824, opened 2023-05-04
test: add regression test for page.reload() and history entries

aslushnikov:
References https://github.com/microsoft/playwright/issues/22640

### DIFF
--- a/tests/page/page-history.spec.ts
+++ b/tests/page/page-history.spec.ts
@@ -110,6 +110,14 @@ it('page.reload should work', async ({ page, server }) => {
   expect(await page.evaluate(() => window['_foo'])).toBe(undefined);
 });
 
+it('page.reload should not push history entry', async ({ page, server }) => {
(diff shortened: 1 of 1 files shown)


tests/page/page-history.spec.ts
```diff
@@ -110,6 +110,14 @@ it('page.reload should work', async ({ page, server }) => {
   expect(await page.evaluate(() => window['_foo'])).toBe(undefined);
 });
 
+it('page.reload should not push history entry', async ({ page, server }) => {
+  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/22640' });
+  await page.goto(server.EMPTY_PAGE);
+  console.log(await page.evaluate('history.length'));
```
mxschmitt, line 116: assert?

gtm-nayan: IIRC the Chromium build actually counts the initial blank page as one entry so you can't directly hardcode the length in. Should instead assert that history lengths are equal before and after the reload.

PR #22837, opened 2023-05-04
chore: render test steps in the trace

pavelfeldman:


dgozman (approved):

packages/playwright-test/src/worker/testInfo.ts
```diff
@@ -257,6 +261,8 @@ export class TestInfoImpl implements TestInfo {
           error,
         };
         this._onStepEnd(payload);
+        const errorForTrace = error ? { name: '', message: error.message || '', stack: error.stack } : undefined;
+        this._traceEvents.push(createAfterActionTraceEventForStep(stepId, serializeAttachments(this.attachments, initialAttachments), errorForTrace));
```
dgozman, line 265: This will pick up the same attachment for both steps:

```js
await test.step('outer', async () => {
  await test.step('inner', async () => {
    await test.info().attach(...);
  });
});
```

tests/playwright-test/playwright.trace.spec.ts
```diff
@@ -87,11 +87,59 @@ test('should record api trace', async ({ runInlineTest, server }, testInfo) => {
   expect(result.failed).toBe(1);
   // One trace file for request context and one for each APIRequestContext
   const trace1 = await parseTrace(testInfo.outputPath('test-results', 'a-pass', 'trace.zip'));
-  expect(trace1.actions).toEqual(['browserContext.newPage', 'page.goto', 'apiRequestContext.get']);
+  expect(trace1.apiNames).toEqual([
+    'Before Hooks',
+    'apiRequest.newContext',
+    'tracing.start',
```
dgozman, line 93: Why double `tracing.start` here?

tests/playwright-test/playwright.reuse.spec.ts
```diff
@@ -144,22 +144,31 @@ test('should reuse context with trace if mode=when-possible', async ({ runInline
...
   const trace1 = await parseTrace(testInfo.outputPath('test-results', 'reuse-one', 'trace.zip'));
-  expect(trace1.actions).toEqual([
+  expect(trace1.apiNames).toEqual([
+    'Before Hooks',
+    'browserType.launch',
     'browserContext.newPage',
     'page.setContent',
     'page.click',
+    'After Hooks',
+    'tracing.stopChunk',
```
dgozman, line 154: I wonder why two `stopChunk` calls?

dgozman (approved):

packages/playwright-test/src/worker/testInfo.ts
pavelfeldman: Let's start like that, this way we can render 'attachment' in the report against the step that has attachments.

PR #22852, opened 2023-05-05
updated dotnet container jammy image

rmjoia:
the yaml example was using the node image version
since it didn't have powershell it failed the pipeline additionally the path was using backslashes which also errored in ubuntu

### DIFF
--- a/docs/src/ci.md
+++ b/docs/src/ci.md
@@ -194,7 +194,7 @@ jobs:
     name: 'Playwright Tests'
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.34.0-jammy
+      image: mcr.microsoft.com/playwright/python:v1.34.0-jammy
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python
@@ -218,7 +218,7 @@ jobs:
     name: 'Playwright Tests'
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.34.0-jammy
+      image: mcr.microsoft.com/playwright/java:v1.34.0-jammy
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
@@ -239,7 +239,7 @@ jobs:
     name: 'Playwright Tests'
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright/dotnet:v1.33.0-jammy
+      image: mcr.microsoft.com/playwright/dotnet:v1.34.0-jammy
     steps:
       - uses: actions/checkout@v3
       - name: Setup dotnet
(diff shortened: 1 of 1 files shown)


docs/src/ci.md
```diff
@@ -239,7 +239,7 @@ jobs:
     name: 'Playwright Tests'
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.34.0-jammy
+      image: mcr.microsoft.com/playwright/dotnet:v1.33.0-jammy
```
pavelfeldman, line 242: 1.34?

rmjoia:
Doesn't exist, i tried, fails.

docs/src/ci.md
rmjoia: 1.34 doesn't work.
` Error response from daemon: manifest for mcr.microsoft.com/playwright/dotnet:v1.34.0-jammy not found: manifest unknown: manifest tagged by "v1.34.0-jammy" is not found`

mxschmitt: It needs to be 1.34 since the stable release for 1.33 already happened. We can backport, but to make lint happy, this needs to be 1.34.

Do you mind making the same change for Java and Python?

rmjoia: @mxschmitt I can update the docs, but if anyone tries to use it now, will get that error as I did. I tried to use it and it failed. 
My intention was to make sure the documentation had an example that actually worked.

mxschmitt: I will backport it afterwards and then your fix is live if you want. Do you mean with not working that its not the right version?

rmjoia: I mean, at least for dotnet, version 34 is not available yet, I pasted the error above on this thread.

mxschmitt: Oh maybe you didn't know. These docs which you are editing are not intended to be consumed via GitHub. They get published on playwright.dev with the right version, thats the code snippets are weird, versions off etc.

See here the outcome: https://playwright.dev/dotnet/docs/ci

So if we would backport, the version gets fixed (changed from 1.34 to 1.33).

rmjoia: image

rmjoia: Thanks @mxschmitt! I was wondering if the docs wouldn't be improved if some examples following the templates were added to git repos with sample code and yaml... where the community could see it working and a paragraph with something along the lines of always check [https://mcr.microsoft.com](https://mcr.microsoft.com/en-us/catalog?search=playwright) for the latest version and ensure that your dependency matches the container version. coz.. won't work if it doesn't (I had that issue... and was wondering what was the point of the container if I had to download the browsers anyway, which... btw.. the docs suggest you do, but you don't need because they are in the container already)

mxschmitt (approved):

rmjoia:

docs/src/ci.md
mxschmitt: sure! the ci.md is definitely one of our guides which needs some more improvements. I updated your PR so it does not contain the install anymore, since the Docker image already contains the browsers. Feel free to file an issue with improvements you suggest.

PR #22858, opened 2023-05-05
fix(deps): `--repeat-each` should not apply to dependencies

### LINKED ISSUE #21778: [BUG] using setup in project dependencies --repeat-each will repeat the setup
### System info
- Playwright Version: v1.32
- Operating System: Windows 10, Ubuntu 20
- Browser: Chromium
- Other info:

When using a "setup" in project dependencies and you run a test with `--repeat-each` the setup step is repeated as well.
I don't know if this is bug or "works as design" but it would be nice if the setup code only ran once. We can work around this my just running the setup one and then run the test with a project that do not have the setup as dependency, like this (see the  playwright.config.ts below):
```js
npx playwright test ./tests/setup.ts && npx playwright test ./tests/myTest.spec.ts --project local-chromium --repeat-each 25 --workers 1
```
But this is not optimal since the second playwright run will overwrite the report from the setup.

### Source code

- [*] I provided exact source code that allows reproducing the issue locally.

**Config file**

```js
// playwright.config.ts

export default defineConfig({
  projects: [
   {
      name: 'setup',
      testMatch: /setup\.ts/,
    },
    {
      name: 'chromium',
      use: {
        ...devices['Desktop Chrome'],
      },
      dependencies: ['setup'],
    },
    {
      name: 'local-chromium',
      use: {
        ...devices['Desktop Chrome'],
      },
    },
});
```

```js
// setup.ts
import { test as setup } from '@playwright/test';
setup('setup', async ({}) => {
  // setup code that takes long time
});
```

```js
// myTest.spec.ts
test('some fine test', async ({ page }) => {
  // some excellent test code
});
```

**Steps**
- Run the test: npx playwright test ./tests/myTest.spec.ts --repeat-each 3 --workers 1

**Expected**
```
  ✓  1 [setup] › setup.ts:11:6 › setup (11.0s)
  ✓  2 [chromium] › myTest.spec.ts › some fine test (740ms)
  ✓  3 [chromium] › myTest.spec.ts › some fine test (647ms)
  ✓  4 [chromium] › myTest.spec.ts› some fine test (656ms)
```
the setup code should only run once

**Actual**

```
  ✓  1 [setup] › setup.ts:11:6 › setup (12.7s)
  ✓  2 [setup] › setup.ts:11:6 › setup (10.1s)
  ✓  3 [setup] › setup.ts:11:6 › setup (11.0s)
  ✓  4 [chromium] › myTest.spec.ts › some fine test (740ms)
  ✓  5 [chromium] › myTest.spec.ts › some fine test (647ms)
  ✓  6 [chromium] › myTest.spec.ts› some fine test (656ms)
```
The set up is ran the "repeat-each" number of times

dgozman:
Fixes #21778.

### DIFF
--- a/packages/playwright-test/src/common/config.ts
+++ b/packages/playwright-test/src/common/config.ts
@@ -164,7 +164,9 @@ export class FullProjectInternal {
       grep: takeFirst(projectConfig.grep, config.grep, defaultGrep),
       grepInvert: takeFirst(projectConfig.grepInvert, config.grepInvert, null),
       outputDir: takeFirst(configCLIOverrides.outputDir, pathResolve(configDir, projectConfig.outputDir), pathResolve(configDir, config.outputDir), path.join(throwawayArtifactsPath, 'test-results')),
-      repeatEach: takeFirst(configCLIOverrides.repeatEach, projectConfig.repeatEach, config.repeatEach, 1),
+      // Note: we either apply the cli override for repeatEach or not, depending on whether the
+      // project is top-level vs dependency. See collectProjectsAndTestFiles in loadUtils.
+      repeatEach: takeFirst(projectConfig.repeatEach, config.repeatEach, 1),
       retries: takeFirst(configCLIOverrides.retries, projectConfig.retries, config.retries, 0),
       metadata: takeFirst(projectConfig.metadata, config.metadata, undefined),
       name: takeFirst(projectConfig.name, config.name, ''),
--- a/packages/playwright-test/src/runner/loadUtils.ts
+++ b/packages/playwright-test/src/runner/loadUtils.ts
@@ -73,6 +73,12 @@ export async function collectProjectsAndTestFiles(testRun: TestRun, additionalFi
     }
   }
 
+  // Apply overrides that are only applicable to top-level projects.
+  for (const [project, type] of projectClosure) {
+    if (type === 'top-level')
+      project.project.repeatEach = project.fullConfig.configCLIOverrides.repeatEach ?? project.project.repeatEach;
+  }
+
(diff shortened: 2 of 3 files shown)


packages/playwright-test/src/runner/loadUtils.ts
```diff
@@ -73,6 +73,12 @@ export async function collectProjectsAndTestFiles(testRun: TestRun, additionalFi
     }
   }
 
+  // Apply overrides that are only applicable to top-level projects.
```
pavelfeldman, line 76: We'll probably be looking for this code everywhere, except for this method.

packages/playwright-test/src/common/config.ts
```diff
@@ -164,7 +164,7 @@ export class FullProjectInternal {
       grep: takeFirst(projectConfig.grep, config.grep, defaultGrep),
       grepInvert: takeFirst(projectConfig.grepInvert, config.grepInvert, null),
       outputDir: takeFirst(configCLIOverrides.outputDir, pathResolve(configDir, projectConfig.outputDir), pathResolve(configDir, config.outputDir), path.join(throwawayArtifactsPath, 'test-results')),
-      repeatEach: takeFirst(configCLIOverrides.repeatEach, projectConfig.repeatEach, config.repeatEach, 1),
+      repeatEach: takeFirst(projectConfig.repeatEach, config.repeatEach, 1),
```
pavelfeldman, line 167: Add a comment and point to where the counterpart is? Otherwise, when I'll refactor it, I'll normalize to include cli overrides.

pavelfeldman (approved):

PR #22887, opened 2023-05-08
feat(typescript): align with `--moduleResolution=bundler`

### LINKED ISSUE #22169: [Feature] Support TypeScript's `--moduleResolution bundler`
Copying from https://github.com/microsoft/playwright/issues/19710#issuecomment-1454132730 by @danieldiekmeier:

---

In TypeScript 5.0, there (probably) will be a new module resolution option: `--moduleResolution bundler`: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#moduleresolution-bundler

I installed the TypeScript 5.0 Beta because I also want to use imports without extensions in Playwright. Sadly, it still does not seem to work. Are there plans to update Playwright to also support `--moduleResolution bundler`?

dgozman:
This relaxes import requirements and allows importing `.ts` files without an extension in CJS and ESM modes.

Fixes #22169.


pavelfeldman (approved):

packages/playwright-test/src/util.ts
```diff
@@ -307,14 +307,26 @@ export function envWithoutExperimentalLoaderOptions(): NodeJS.ProcessEnv {
...
+  ['', ['.js', '.jsx', '.ts', '.tsx', '.cjs', '.mjs', '.cts', '.mts']],
+]);
+export function resolveImportSpecifierExtension(resolved: string): string | undefined {
+  if (fs.existsSync(resolved))
+    return resolved;
+  for (const [ext, others] of kExtLookups) {
+    if (!resolved.endsWith(ext))
+      continue;
+    for (const other of others) {
+      const modified = resolved.substring(0, resolved.length - ext.length) + other;
```
pavelfeldman, line 326: Looks like you could move it outside of the loop: `resolved.substring(0, resolved.length - ext.length)`

pavelfeldman (approved):

kristojorg:
Is there any comment on when this will be released?

dgozman:
@kristojorg Should be already available in v1.34.

kristojorg:
Ah I see it is, didn't see it in the changelog but it's working. I am wondering, however, if directory imports are meant to be supported by this as well? Outside Playwright in my ESModules app I can import directories `/dir` instead of `/dir/index`, but in playwright I'm not able to. I no longer have to import `/dir/index.ts` but I do have to import `/dir/index`. Is this intended?

kristojorg:
It does look like TS intends to enable the directory imports according to the `bundler` PR: https://github.com/microsoft/TypeScript/pull/51669

kristojorg:
I went ahead and added a PR: https://github.com/microsoft/playwright/pull/23254

PR #22888, opened 2023-05-08
chore: render parse errors in the UI mode

### LINKED ISSUE #22863: [Feature]: In UI mode, syntax errors fail silently
If you have a test file with a syntax error and you start UI mode you don't get any indication that there was a problem. You simply get the app showing there are no tests:

image

The terminal doesn't show there are any issues:

image

To find where the issue is you must run non-UI mode:

image

Would be nice to get errors like this in the terminal, and also cool to get something in the app as well.

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/22863

### DIFF
--- a/packages/trace-viewer/src/ui/sourceTab.tsx
+++ b/packages/trace-viewer/src/ui/sourceTab.tsx
@@ -22,15 +22,14 @@ import './sourceTab.css';
 import { StackTraceView } from './stackTrace';
 import { CodeMirrorWrapper } from '@web/components/codeMirrorWrapper';
 import type { SourceHighlight } from '@web/components/codeMirrorWrapper';
-import type { SourceModel } from './modelUtil';
-import type { StackFrame } from '@protocol/channels';
+import type { SourceLocation, SourceModel } from './modelUtil';
 
 export const SourceTab: React.FunctionComponent<{
   action: ActionTraceEvent | undefined,
   sources: Map<string, SourceModel>,
   hideStackFrames?: boolean,
   rootDir?: string,
-  fallbackLocation?: StackFrame,
+  fallbackLocation?: SourceLocation,
 }> = ({ action, sources, hideStackFrames, rootDir, fallbackLocation }) => {
   const [lastAction, setLastAction] = React.useState<ActionTraceEvent | undefined>();
   const [selectedFrame, setSelectedFrame] = React.useState<number>(0);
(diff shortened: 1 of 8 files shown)


packages/trace-viewer/src/ui/sourceTab.tsx
```diff
@@ -43,35 +42,37 @@ export const SourceTab: React.FunctionComponent<{
...
-    const fileName = rootDir && location.file.startsWith(rootDir) ? location.file.substring(rootDir.length + 1) : location.file;
-    const highlight: SourceHighlight[] = source.errors.map(e => ({ type: 'error', line: e.location.line, message: e.error!.message }));
+    const targetLine = shoulduseFallback ? fallbackSource?.errors[0]?.line || 0 : actionLocation.line;
+    const fileName = rootDir && file.startsWith(rootDir) ? file.substring(rootDir.length + 1) : file;
+    const highlight: SourceHighlight[] = source.errors.map(e => ({ type: 'error', line: e.line, message: e.message }));
     highlight.push({ line: targetLine, type: 'running' });
 
-    if (source.content === undefined || fallbackLocation) {
-      const sha1 = await calculateSha1(location.file);
+    if (source.content === undefined || fallbackSource) {
```
dgozman, line 63: I wonder why fetch when `fallbackSource` is there, but `!shoulduseFallback`?

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -607,11 +620,13 @@ const refreshRootSuite = (eraseResults: boolean): Promise<void> => {
         ++progress.failed;
       else
         ++progress.passed;
-      throttleUpdateRootSuite(config, rootSuite, progress);
+      throttleUpdateRootSuite(config, rootSuite, loadErrors, progress);
     },
 
     onError: (error: TestError) => {
       xtermDataSource.write((error.stack || error.value || '') + '\n');
+      loadErrors.push(error);
```
dgozman, line 628: This will keep pushing new (probably duplicate) errors after the initial load.

dgozman (approved):

packages/trace-viewer/src/ui/sourceTab.tsx
pavelfeldman: They are interchangeable in this context.

packages/trace-viewer/src/ui/uiModeView.tsx
pavelfeldman: Let me see if I can mute it.

pavelfeldman: Not getting a good idea about it. But willing to wait for a report.

PR #22909, opened 2023-05-09
docs: improve soft assertions

debs-obrien:


docs/src/best-practices-js.md
```diff
@@ -287,3 +273,17 @@ Playwright can [shard](./test-parallel.md#shard-tests-between-multiple-machines)
 ```js
 npx playwright test --shard=1/3
 ```
+
+## Productivity tips
+
+### Use Soft assertions
+
+If your test fails, Playwright will give you an error message showing what part of the test failed which you can see either in VS Code, the terminal, the HTML report, or the trace viewer. However, you can also use [soft assertions](/test-assertions.md#soft-assertions) which do not terminate test execution but mark the test as failed.
```
mxschmitt, line 281:
```suggestion
If your test fails, Playwright will give you an error message showing what part of the test failed which you can see either in VS Code, the terminal, the HTML report, or the trace viewer. However, you can also use [soft assertions](/test-assertions.md#soft-assertions) these do not immediately terminate the test execution, but rather compile and display a list of failed assertions once the test finshes.
```

mxschmitt (approved):

mupkoo:
Why did you remove the `Write fewer tests but longer tests` section?

mxschmitt:
Because this is not a best practise we want to recommend to our users. Tests should be small, concise and test a specific feature in isolation.

PR #22910, opened 2023-05-09
docs: Add-title-to-config

debs-obrien:
Docs PR: https://github.com/microsoft/playwright.dev/pull/1039


docs/src/test-api/class-testconfig.md
```diff
@@ -6,8 +6,7 @@ Playwright Test provides many options to configure how your tests are collected
 
 Playwright Test supports running multiple test projects at the same time. Project-specific options should be put to [`property: TestConfig.projects`], but top-level [TestConfig] can also define base options shared between all projects.
 
-```js
-// playwright.config.ts
+```js title="playwright.config.ts"
```
mxschmitt, line 9: either we need to adjust the type generator for the typescript types or not touch the api/ test-api/ folders. Because they generate invalid code snippets otherwise. See the generated types test.d.ts and testReporter.d.ts.

debs-obrien: it makes sense to be consistent across docs and avoid having to manually put comments for example snippets for titles but can revert back and we can look into if we also change api docs too and how we would do it

utils/markdown.js
```diff
@@ -327,7 +328,9 @@ function innerRenderMdNode(indent, node, lastNode, result, options) {
 
   if (node.type === 'code') {
     newLine();
-    result.push(`${indent}\`\`\`${node.codeLang}${node.title ? ' title="' + node.title + '"' : ''}`);
+    result.push(`${indent}\`\`\`${node.codeLang}${(options?.renderCodeBlockTitlesInHeader && node.title) ? ' title="' + node.title + '"' : ''}`);
+    if (!options?.renderCodeBlockTitlesInHeader && node.title)
```
mxschmitt, line 332: no super happy about this name yet.

yury-s (approved):

Issue #9682, opened 2021-10-21
[Feature] allow multiple filters in --grep with -g tag1 -g tag2


ychaudhari:
We have multiple tags in our test suites that we want to combine and run together in one CD stage. It would be great if we have an ability to provide multiple parameters like `-g tag1 -g tag2 -g tag3` in playwright/test CLI. 

Currently there are some issues in using `-g tag1|tag2` as discussed in this thread: 
https://playwright.slack.com/archives/C01AG7QGW2F/p1634755054418000

pavelfeldman:
There are no issues with -g tag1|tag2, you just need to escape it if you use Windows cdm as discussed in that thread.

`-g tag1^|tag2`

ychaudhari:
@pavelfeldman aha cool... thanks for that. Happy to close this
I'll comment on #9667 because that fix didn't work for me for some reason :(

ffluk3:
@pavelfeldman Please forgive if this is discussed in the Slack thread, but I think there is still a valid use case to have the multiple grep/grepInvert tags. For example:

```json
{
  "scripts": {
    "e2e": "playwright --grep-invert \"@wip\""
  }
}
```

However, if I wanted to omit another tag on the fly, I wouldn't be able to run `npm run e2e -- --grep-invert @flaky`. Is there a recommended way to approach this use case?

PR #22911, opened 2023-05-09
docs: add example to run multiple tags at a time

kevin-brotcke:
I checked the [docs](https://playwright.dev/docs/test-annotations#tag-tests) for how to do this but I didn't see anything. I first tried adding multiple grep params but that didn't work, only the last grep param was applied. I had to search for a workaround and found [this](https://github.com/microsoft/playwright/issues/9682) issue showing it's already supported so I've updated the docs to reflect that.


kevin-brotcke:
I also needed the intersection (`AND`) of tags, so posted a new pull request including it and will close this one:

https://github.com/microsoft/playwright/pull/22913

PR #22916, opened 2023-05-09
fix(trace): merge step hierarchies

pavelfeldman:


packages/trace-viewer/src/ui/modelUtil.ts
```diff
@@ -122,6 +122,8 @@ function mergeActions(contexts: ContextEntry[]) {
           existing.error = action.error;
         if (action.attachments)
           existing.attachments = action.attachments;
+        if (action.parentId)
+          existing.parentId = action.parentId;
```
yury-s, line 126: Should it be unconditional, so that the action could be removed from the parent?

yury-s (approved):

PR #22991, opened 2023-05-12
docs: minor auth.md update

johnGachihi:
This might save a playwright newb like me a few minutes

### DIFF
--- a/docs/src/auth.md
+++ b/docs/src/auth.md
@@ -68,7 +68,7 @@ setup('authenticate', async ({ page }) => {
 });
 ```
 
-Create a new `setup` project in the config and declare it as a [dependency](./test-projects-js.md#dependencies) for all your testing projects. This project will always run and authenticate before all the tests. All testing projects should use the authenticated state as `storageState`.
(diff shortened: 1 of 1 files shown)


docs/src/auth.md
```diff
@@ -68,7 +68,7 @@ setup('authenticate', async ({ page }) => {
 });
 ```
 
-Create a new `setup` project in the config and declare it as a dependency for all your testing projects. This project will always run and authenticate before all the tests. All testing projects should use the authenticated state as `storageState`.
+Create a new `setup` project in the config and declare it as a [dependency](./test-projects-js.md#dependencies) for all your testing projects. This project will always run and authenticate before all the tests. All testing projects should use the authenticated state as `storageState`.
```
mxschmitt, line 71:
```suggestion
Create a new `setup` project in the config and declare it as a [dependency](./test-projects.md#dependencies) for all your testing projects. This project will always run and authenticate before all the tests. All testing projects should use the authenticated state as `storageState`.
```

mxschmitt (approved):

mxschmitt (approved):

PR #23002, opened 2023-05-12
feat(locator): relaxed locator parsing

### LINKED ISSUE #22965: [BUG] finding locators in Inspector ignores locator options param
### System info
- Operating System: macOS 13.3.1
- Browser: Chromium

### Steps to reproduce

1. Run `npx playwright codegen playwright.dev`
2. Click 'Record' button in Inspector to stop recording
3. In the `Pick locator` box, enter `locator('.navbar__item', {hasText: 'API'})`, notice it does not highlight any elements
4. Now enter `locator('.navbar__item').filter({hasText: 'API'})`, notice it highlights the nav bar 'API' link

**Expected**

Both locators should find an element, and the same element, as they should be equivalent.

**Actual**

The first locator does not find any elements

dgozman:
Includes support for `locator(selector, { has, hasText })`.

Fixes #22965.

### DIFF
--- a/packages/playwright-core/src/utils/isomorphic/locatorParser.ts
+++ b/packages/playwright-core/src/utils/isomorphic/locatorParser.ts
@@ -15,7 +15,6 @@
  */
 
 import { escapeForAttributeSelector, escapeForTextSelector } from '../../utils/isomorphic/stringUtils';
-import { asLocator } from './locatorGenerators';
 import type { Language } from './locatorGenerators';
 import { parseSelector } from './selectorParser';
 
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/utils/isomorphic/locatorParser.ts
```diff
@@ -202,14 +204,8 @@ export function locatorOrSelectorAsSelector(language: Language, locator: string,
   } catch (e) {
   }
   try {
-    const selector = parseLocator(locator, testIdAttributeName);
-    if (digestForComparison(asLocator(language, selector)) === digestForComparison(locator))
-      return selector;
+    return parseLocator(locator, testIdAttributeName);
```
pavelfeldman, line 207: Let's generate all the valid locators in `asLocatorOptions` and validate?

PR #23004, opened 2023-05-12
fix: unique resource names between shards

yury-s:


packages/playwright-test/src/reporters/blob.ts
```diff
@@ -81,7 +84,7 @@ export class BlobReporter extends TeleReporterEmitter {
       const paddedNumber = `${config.shard.current}`.padStart(`${config.shard.total}`.length, '0');
       shardSuffix = `${paddedNumber}-of-${config.shard.total}-`;
     }
-    return path.join(this._outputDir, `report-${shardSuffix}${createGuid()}.jsonl`);
+    return path.join(this._outputDir, `report-${shardSuffix}${this._runGuid}.jsonl`);
```
aslushnikov, line 87: this we can safely revert

aslushnikov (approved):

PR #23005, opened 2023-05-12
chore: partially hide built-in fixture steps

pavelfeldman:


packages/playwright-test/src/util.ts
```diff
@@ -40,15 +40,21 @@ export function filterStackTrace(e: Error): { message: string, stack: string } {
   };
 }
 
+export function filterStackFile(file: string) {
```
dgozman, line 43:
```suggestion
export function isInternalStackFile(file: string) {
```

dgozman (approved):

packages/playwright-test/src/util.ts
pavelfeldman: follow up...

PR #23030, opened 2023-05-15
chore: validate expected image buffer when comparing images

### LINKED ISSUE #23012: [Bug] Getting 'unrecognised content at end of stream' when Git LFS snapshots are not fetched.
Issue: I'm using Git LFS, the files are not fetched yet and I'll get `unrecognised content at end of stream` when running Playwright toHaveSnapshot tests.

Solution: Install Git LFS, and pull the snapshots, then its working: `git lfs pull`

Maybe we can on our side throw a better error if the fetched image is e.g. not a PNG file when we try to parse it. Took me a while to figure out what was actually happening.

When using Git LFS the file PNG files look like the following when the repository is fetched without Git LFS:

image

```
version https://git-lfs.github.com/spec/v1
oid sha256:d6281b717a082f21e24f82cd65df5a2d859f0893169d8fe75d39c19c7865e7e4
size 49232
```

Relates https://github.com/microsoft/playwright/issues/22903

We could do something like:

```ts
  if (mimeType.startsWith('image/') && expectedBuffer.indexOf(Buffer.from('version ')) !== -1)
    return {
      errorMessage: 'The expected image is a Git LFS object. Did you fetch the Git LFS objects before cloning the repository?'
    }
```

here: https://github.com/microsoft/playwright/blob/17ab777db5ae8b24c6654e1c8ecb0b9672538c86/packages/playwright-core/src/utils/comparators.ts#L56

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/23012

### DIFF
--- a/packages/playwright-core/src/utils/comparators.ts
+++ b/packages/playwright-core/src/utils/comparators.ts
@@ -52,9 +52,9 @@ type ImageData = { width: number, height: number, data: Buffer };
 function compareImages(mimeType: string, actualBuffer: Buffer | string, expectedBuffer: Buffer, options: ImageComparatorOptions = {}): ComparatorResult {
   if (!actualBuffer || !(actualBuffer instanceof Buffer))
     return { errorMessage: 'Actual result should be a Buffer.' };
+  validateBuffer(expectedBuffer, mimeType);
 
   let actual: ImageData = mimeType === 'image/png' ? PNG.sync.read(actualBuffer) : jpegjs.decode(actualBuffer, { maxMemoryUsageInMB: JPEG_JS_MAX_BUFFER_SIZE_IN_MB });
-  validateBuffer(expectedBuffer, mimeType);
   let expected: ImageData = mimeType === 'image/png' ? PNG.sync.read(expectedBuffer) : jpegjs.decode(expectedBuffer, { maxMemoryUsageInMB: JPEG_JS_MAX_BUFFER_SIZE_IN_MB });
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/utils/comparators.ts
```diff
@@ -92,6 +93,14 @@ function compareImages(mimeType: string, actualBuffer: Buffer | string, expected
   return null;
 }
 
+function validateBuffer(buffer: Buffer, mimeType: string): void {
+  if (mimeType === 'image/png') {
```
mxschmitt, line 97: Should I add JPEG as well?

packages/playwright-core/src/utils/comparators.ts
```diff
@@ -54,6 +54,7 @@ function compareImages(mimeType: string, actualBuffer: Buffer | string, expected
     return { errorMessage: 'Actual result should be a Buffer.' };
 
   let actual: ImageData = mimeType === 'image/png' ? PNG.sync.read(actualBuffer) : jpegjs.decode(actualBuffer, { maxMemoryUsageInMB: JPEG_JS_MAX_BUFFER_SIZE_IN_MB });
+  validateBuffer(expectedBuffer, mimeType);
```
aslushnikov, line 57: is there a reason why this is after decoding the `afterBuffer`?

packages/playwright-core/src/utils/comparators.ts
aslushnikov: yes please!

aslushnikov (approved):

PR #23044, opened 2023-05-16
docs: fix `testConfig.testMatch` markdown escaping

sam-super:
Because the `|` is meaningful in the markdown table, it causes truncated docs:
image

### DIFF
--- a/docs/src/test-configuration-js.md
+++ b/docs/src/test-configuration-js.md
@@ -85,7 +85,7 @@ export default defineConfig({
 | Option | Description |
 | :- | :- |
 | [`property: TestConfig.testIgnore`] | Glob patterns or regular expressions that should be ignored when looking for the test files. For example, `'*test-assets'` |
-| [`property: TestConfig.testMatch`] | Glob patterns or regular expressions that match test files. For example, `'*todo-tests/*.spec.ts'`. By default, Playwright runs `.*(test\|spec)\.(js\|ts\|mjs)` files. |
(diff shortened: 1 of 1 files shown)


mxschmitt:
If we do this, then it looks like the following which is also unexpected:

image

This seems to be caused by the code block aka. the backticks, if I remove them, then it looks fine:

image

docs/src/test-configuration-js.md
```diff
@@ -85,7 +85,7 @@ export default defineConfig({
 | Option | Description |
 | :- | :- |
 | [`property: TestConfig.testIgnore`] | Glob patterns or regular expressions that should be ignored when looking for the test files. For example, `'*test-assets'` |
-| [`property: TestConfig.testMatch`] | Glob patterns or regular expressions that match test files. For example, `'*todo-tests/*.spec.ts'`. By default, Playwright runs `.*(test|spec)\.(js|ts|mjs)` files. |
+| [`property: TestConfig.testMatch`] | Glob patterns or regular expressions that match test files. For example, `'*todo-tests/*.spec.ts'`. By default, Playwright runs `.*(test\|spec)\.(js\|ts\|mjs)` files. |
```
mxschmitt, line 88:
```suggestion
| [`property: TestConfig.testMatch`] | Glob patterns or regular expressions that match test files. For example, `'*todo-tests/*.spec.ts'`. By default, Playwright runs <code>.*(test&#124;spec)\.(js&#124;ts&#124;mjs)</code> files. |
```

this seems like a good workaround as per https://github.com/facebook/docusaurus/issues/2004#issuecomment-561136550

PR #23048, opened 2023-05-16
docs: maxDiffPixelRatio example value should be between 0 and 1

adrien-fichet:
On the page https://playwright.dev/docs/test-configuration, the `maxDiffPixelRatio` example value is 10 instead of being between 0 and 1. Set it to 0.1 instead.

Fixes https://github.com/microsoft/playwright/pull/23049


adrien-fichet:

adrien-fichet:

mxschmitt (approved):

PR #23062, opened 2023-05-16
chore: don't install gstreamer for WebKit Ubuntu 20

mxschmitt:


packages/playwright-core/src/server/registry/nativeDeps.ts
```diff
@@ -254,10 +254,8 @@ export const deps: any = {
     ],
     webkit: [
       'libenchant-2-2',
-      'gstreamer1.0-libav',
-      'gstreamer1.0-plugins-bad',
-      'gstreamer1.0-plugins-base',
-      'gstreamer1.0-plugins-good',
+      'libflite1',
```
yury-s, line 257: I don't see any entries for these two in `lib2package` map, is that expected?

yury-s (approved):

yury-s (approved):

PR #23073, opened 2023-05-17
chore(trace): do not flush console events

### LINKED ISSUE #20626: [Feature] More granular support of what Playwright trace file can gather
Currently seeing a problem with gathering a Playwright trace file for our tests.

We start our Playwright tracer with :
```
    await this.playwrightWindow?.tracing.start({
      screenshots: true,
      snapshots: true,
      sources: false,
    });
````
and then ~4 minutes ( on average ) later we stop it 
```
            await tracingContext.stop({
              path: traceDirectory,
            }),
````

We are seeing that a lot of our tests are timing out from this, so we had to add a top level 45s timeout.

```
2023-02-01T19:19:34.1641975Z      🛑 Stopping Playwright UI Driver
2023-02-01T19:20:19.0624297Z      📩❌ Timed out trying to stop Playwright trace. (took longer than 45000ms)
2023-02-01T19:20:19.2285355Z      🛑 Stopped Playwright UI Driver
```

From my investigation, it looks like our product is VERY verbose in its console messages.

image

Anyway we can batch these? or report only the important console message events like Error?

It also looks like "JSHandle" is extremely verbose here : 

image

### PR DESCRIPTION
**Pull request by pavelfeldman**
Fixes https://github.com/microsoft/playwright/issues/20626

### DIFF
--- a/packages/playwright-core/src/server/trace/recorder/tracing.ts
+++ b/packages/playwright-core/src/server/trace/recorder/tracing.ts
@@ -54,8 +54,8 @@ export type TracerOptions = {
 type RecordingState = {
   options: TracerOptions,
   traceName: string,
-  networkFile: string,
-  traceFile: string,
+  networkFile: { file: string, buffer: trace.TraceEvent[] },
+  traceFile: { file: string, buffer: trace.TraceEvent[] },
   tracesDir: string,
   resourcesDir: string,
   chunkOrdinal: number,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/playwright-core/src/server/trace/recorder/tracing.ts [hunk 1]
```diff
@@ -132,15 +132,26 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
...
+      resourcesDir: '',
+      chunkOrdinal: 0,
+      traceSha1s: new Set(),
+      networkSha1s: new Set(),
+      recording: false
+    };
     const state = this._state;
 
-    state.tracesDir = await this._createTracesDirIfNeeded();
     state.resourcesDir = path.join(state.tracesDir, 'resources');
```
dgozman, line 153: nit: move to line 145 right away?

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -201,10 +216,16 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
...
       const oldNetworkFile = state.networkFile;
       state.traceName = name;
-      state.traceFile = path.join(state.tracesDir, name + '.trace');
-      state.networkFile = path.join(state.tracesDir, name + '.network');
+      state.traceFile = {
+        file: path.join(state.tracesDir, name + '.trace'),
+        buffer: [],
+      };
+      state.networkFile = {
+        file: path.join(state.tracesDir, name + '.network'),
```
dgozman, line 224: I think we should flush files here, otherwise we might be copying an uncompleted old network file at least.

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -272,10 +293,10 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
       // We can use <traceName>-<guid>.network, but "-pwnetcopy-0" suffix is more readable
       // and makes it easier to debug future issues.
       const networkFile = path.join(state.tracesDir, state.traceName + `-pwnetcopy-${state.chunkOrdinal}.network`);
-      await fs.promises.copyFile(state.networkFile, networkFile);
+      await fs.promises.copyFile(state.networkFile.file, networkFile);
```
dgozman, line 296: Same here, we need to force flush both files.

dgozman (approved):

packages/playwright-core/src/server/trace/recorder/tracing.ts
pavelfeldman: I thought we only call it in the startChunk?

PR #23100, opened 2023-05-17
feat(webkit): roll to r1850

playwrightmachine:


mxschmitt:
WPE tap tests are failing:

```
    [webkit] › library/tap.spec.ts:23:3 › should send all of the correct events @smoke ─────────────
    [webkit] › library/tap.spec.ts:43:3 › trial run should not tap ─────────────────────────────────
    [webkit] › library/tap.spec.ts:55:3 › should not send mouse events touchstart is canceled ──────
    [webkit] › library/tap.spec.ts:71:3 › should not send mouse events when touchend is canceled ───
```

yury-s:
https://github.com/microsoft/playwright/pull/23267

PR #23124, opened 2023-05-18
chore(internal): generate code in jsonl

pavelfeldman:


dgozman (approved):

lharries:
Hi @pavelfeldman, hope you don't mind me commenting on this PR.

Is the jsonl format likely to be supported going forward as I'm considering building on it for a project?

(Context if useful - I'm considering using the jsonl format from codegen as an abstraction of the generated test that could be edited and then later used to generate python/javascript playwright tests, and maybe even cypress tests)

PR #23128, opened 2023-05-18
feat(cli): introduce npx playwright uninstall

### LINKED ISSUE #23072: [Feature] npx playwright uninstall
Based on our customers feedback (https://github.com/microsoft/playwright/issues/15049#issuecomment-1163980204) we should make it easy to uninstall the Playwright browsers from the system.

mxschmitt:
As per offline discussion:

- removes the backlink to the current folder
- runs prune
- if more browsers are there, print message and run with --all

Fixes https://github.com/microsoft/playwright/issues/23072


mahtabnejad90 (commented):
would be nice to have some documentation around this actually.

dgozman (approved):

PR #23185, opened 2023-05-21
feat(screenshot): Add customizable color option for masked elements

Darkensses:
I added a new option to the screenshot method to customize the color of the box when we want to mask some elements for the screenshot.

The default color is pink `#FF00FF`, but with this new option you can specify the color you like the most, like a nice green `#00FF00`:

```js
await page.screenshot({
  mask: [page.locator('div').nth(5)],
  maskColor: "#00FF00",
})
```

ss


Darkensses:
> @microsoft-github-policy-service agree

docs/src/api/params.md
```diff
@@ -1135,6 +1140,7 @@ When set to `"hide"`, screenshot will hide text caret. When set to `"initial"`,
 - %%-screenshot-option-caret-%%
 - %%-screenshot-option-type-%%
 - %%-screenshot-option-mask-%%
+- %%-screenshot-option-mask-color-%%
```
dgozman, line 1143: Since this option was not added in v1.8, you'll have to manually add it everywhere `screenshot-options-common-list-v1.8` is mentioned, remove it from this list, and insert `since: v1.34` before line 1096.

dgozman (commented):
Thank you for the PR! Looks pretty good, just a few comments.

Darkensses:
> Thank you for the PR! Looks pretty good, just a few comments.

Done! I made changes to the files according to your suggestions. Let me know if you need something else :) thanks!

dgozman (approved):
Looks great, merging in.

PR #23194, opened 2023-05-21
feat(ct): vue2 type safe mount props

sand4rt:
partial fix for: https://github.com/microsoft/playwright/issues/17206

https://www.npmjs.com/package/vue-component-type-helpers has been released last month and is maintained by Vue. It's also used in Vue Test Utils: https://github.com/vuejs/test-utils/blob/main/package.json#L26

related: https://github.com/microsoft/playwright/pull/23151

### DIFF
--- a/packages/playwright-ct-vue2/index.d.ts
+++ b/packages/playwright-ct-vue2/index.d.ts
@@ -36,20 +36,30 @@ export type PlaywrightTestConfig<T = {}, W = {}> = Omit<BasePlaywrightTestConfig
 };
 
 type Slot = string | string[];
+type ComponentSlots = Record<string, Slot> & { default?: Slot };
 
-export interface MountOptions<
-  HooksConfig extends JsonObject,
-  Props extends Record<string, unknown>
-> {
-  props?: Props;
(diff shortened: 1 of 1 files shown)


packages/playwright-ct-vue2/index.d.ts
```diff
@@ -25,6 +25,7 @@ import type {
 } from '@playwright/test';
 import type { JsonObject } from '@playwright/experimental-ct-core/types/component';
 import type { InlineConfig } from 'vite';
+import type { ComponentProps } from 'vue-component-type-helpers';
```
pavelfeldman, line 28: same as in vue, let's inline these

pavelfeldman:

sand4rt:
> 👋 @sand4rt what should we do with this one?

I resolved the merge conflicts and comment. Would really like to see this merged. [`@vue/test-utils`](https://github.com/vuejs/test-utils/blob/main/src/mount.ts#L21) decided to do this as well and [`playwright-ct-svelte`](https://github.com/microsoft/playwright/blob/main/packages/playwright-ct-svelte/index.d.ts#L44) also has typed props.

pavelfeldman (approved):

PR #23260, opened 2023-05-24
chore: move to npx playwright-core install when using core

pavelfeldman:


dgozman (approved):

dominikg:
@pavelfeldman  i think this is a breaking change. Prior to this, you could `pnpm playwright install chromium`  in ci scripts because the playwright bin was automatically resolved.

PR #23295, opened 2023-05-26
fix: patch project names and ids when merging reports

yury-s:
Project name is used in testId calculation, so patching it in the reporter is too late. Instead we now save project suffix to the blob report file and patch all project and test ids as well as project names in the report merger

### DIFF
--- a/packages/playwright-test/src/reporters/merge.ts
+++ b/packages/playwright-test/src/reporters/merge.ts
@@ -19,7 +19,8 @@ import path from 'path';
 import type { ReporterDescription } from '../../types/test';
 import type { FullResult } from '../../types/testReporter';
 import type { FullConfigInternal } from '../common/config';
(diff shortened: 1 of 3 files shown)


packages/playwright-test/src/reporters/merge.ts
```diff
@@ -153,3 +156,61 @@ async function sortedShardFiles(dir: string) {
...
+    }
+  }
+
+  private _updateTestIds(suite: JsonSuite) {
+    suite.tests.forEach(test => test.testId = this._mapTestId(test.testId));
+    suite.suites.forEach(suite => this._updateTestIds(suite));
+  }
+
+  private _mapTestId(testId: string): string {
+    return testId + '-' + this._projectNameSuffix;
```
dgozman, line 214: The only thing I don't really like is that we are patching test ids. These are supposed to be "history ids", so it seems fragile that changing the suffix changes test ids. However, I don't see a way around it.

dgozman (approved):

PR #23317, opened 2023-05-26
chore: replace sigint handler per browser with a single one

dgozman:
Otherwise, multiple sigint handlers (one from each browser) would try to `process.exit(130)` each.

### DIFF
--- a/packages/playwright-core/src/utils/processLauncher.ts
+++ b/packages/playwright-core/src/utils/processLauncher.ts
@@ -18,7 +18,6 @@
 import * as childProcess from 'child_process';
 import * as readline from 'readline';
 import * as path from 'path';
-import { eventsHelper } from './eventsHelper';
 import { isUnderTest } from './';
 import { removeFolders } from './fileUtils';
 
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/utils/processLauncher.ts
```diff
@@ -51,16 +50,63 @@ type LaunchResult = {
...
+    // Give tests a chance to see that launched process did exit and dispatch any async calls.
+    if (isUnderTest())
+      setTimeout(() => process.exit(130), 1000);
+    else
+      process.exit(130);
+  };
+
+  if (sigintHandlerCalled) {
+    // Resort to default handler from this point on, just in case we hang/stall.
+    process.off('SIGINT', sigintHandler);
```
aslushnikov, line 76: shall we clear the `processHandlers.SIGINT` here as well?

packages/playwright-core/src/utils/processLauncher.ts
```diff
@@ -116,34 +162,25 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
   spawnedProcess.once('exit', (exitCode, signal) => {
     options.log(`[pid=${spawnedProcess.pid}] <process did exit: exitCode=${exitCode}, signal=${signal}>`);
     processClosed = true;
-    eventsHelper.removeEventListeners(listeners);
     gracefullyCloseSet.delete(gracefullyClose);
+    killSet.delete(killProcessAndCleanup);
```
aslushnikov, line 166: to retain old logic, we would need to remove event handlers if `killSet` and `gracefullyCloseSet` sets are empty.

aslushnikov (approved):

packages/playwright-core/src/utils/processLauncher.ts
dgozman: Let's try without this, and we'll bring it back if things break.

PR #23341, opened 2023-05-29
feat: support selenoid for chromium browsers

DudaGod:
## What's done

Selenoid session doesn't return capability `se:cdp` which is specific only for Selenium and not described in specification (w3c, cdp). So I implemented `SELENIUM_WS_ENDPOINT` env variable to specify selenoid ws endpoint (looks like - `ws://selenoid.example.com:4444/devtools/<session-id>`). Moreover env variable `SELENIUM_WS_ENDPOINT` can be used for another selen-projects which not support `se:cdp` capability.

More info about selenoid here - https://aerokube.com/selenoid/latest/

### DIFF
--- a/packages/playwright-core/src/server/chromium/chromium.ts
+++ b/packages/playwright-core/src/server/chromium/chromium.ts
@@ -51,6 +51,7 @@ import { validateBrowserContextOptions } from '../browserContext';
 import { chromiumSwitches } from './chromiumSwitches';
 
 const ARTIFACTS_FOLDER = path.join(os.tmpdir(), 'playwright-artifacts-');
+const SESSION_ID_PLACEHOLDER = /{{SESSION_ID}}/;
 
 export class Chromium extends BrowserType {
   private _devtools: CRDevTools | undefined;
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/chromium/chromium.ts
```diff
@@ -217,7 +218,19 @@ export class Chromium extends BrowserType {
       const capabilities = value.capabilities;
       let endpointURL: URL;
 
-      if (capabilities['se:cdp']) {
+      if (process.env.SELENIUM_WS_ENDPOINT) {
+        // support selenium successors (like selenoid)
+        progress.log(`<selenium> using some selen-project`);
+        let wsEndpoint = process.env.SELENIUM_WS_ENDPOINT;
+
+        if (SESSION_ID_PLACEHOLDER.test(wsEndpoint))
```
DudaGod, line 226: Placeholders implementation is not needed by selenoid itself, but can be used by another selen-project in which `sessionId` might be in the middle of the url (like for selenium4 - `ws://localhost:4444/session/{{SESSION_ID}}/se/cdp`).

DudaGod:

packages/playwright-core/src/server/chromium/chromium.ts
```diff
@@ -217,7 +218,19 @@ export class Chromium extends BrowserType {
       const capabilities = value.capabilities;
       let endpointURL: URL;
 
-      if (capabilities['se:cdp']) {
+      if (process.env.SELENIUM_WS_ENDPOINT) {
```
pavelfeldman, line 221: You should instead fix selenoid to return the endpoint in the se:cdp capability!

DudaGod: You agree that this capability is not described in any specification and is a specific only for selenium (this is indicated by the `se` prefix)? This means that all services that provide access to browsers must support this capability.

I can support it for selenoid if you don't want support additional environment variable.

pavelfeldman:
Closing as per above, please feel free to open a new issue if this does not cover your use case.

PR #23348, opened 2023-05-29
feat: ability to pass additional headers to selenium

DudaGod:
## What's done

Implement ability to pass additional headers to selenium using `SELENIUM_REMOTE_HEADERS` endpoint.

### Use case

I have browsers pool in some cloud. I want that only users with access will be able to connect to them. So they must provide token through headers. But playwright does not allow me to send headers when using pwt as framework. So I added this feature.


pavelfeldman (approved):

pavelfeldman:
Needs rebaseline

DudaGod:
@pavelfeldman when will release with these changes come out?

mxschmitt:
With 1.36 or https://playwright.dev/docs/canary-releases

PR #23360, opened 2023-05-30
fix: should be able to upload multiple large files

### LINKED ISSUE #23258: [BUG] Several files of 50 megabytes are not added to the input
### System info
- Operating System: [All, Windows 11, Ubuntu 20, macOS 13.2, etc.]
- Other info:

### Source code

[/pull/23158]

**Expected**
10 files of 50MB or 30MB should be added to input

**Actual**

files are not added to input

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/23258

### DIFF
--- a/packages/playwright-core/src/client/elementHandle.ts
+++ b/packages/playwright-core/src/client/elementHandle.ts
@@ -271,8 +271,9 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
     throw new Error('Cannot set buffer larger than 50Mb, please write it to a file and pass its path instead.');
 
   const stats = await Promise.all(items.filter(isString).map(item => fs.promises.stat(item as string)));
+  const totalExceedsMaxSize = stats.reduce((acc, stat) => acc + stat.size, 0) > sizeLimit;
   const hasLargeFile = !!stats.find(s => s.size > sizeLimit);
-  if (hasLargeFile) {
+  if (hasLargeFile || totalExceedsMaxSize) {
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -271,8 +271,9 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
     throw new Error('Cannot set buffer larger than 50Mb, please write it to a file and pass its path instead.');
 
   const stats = await Promise.all(items.filter(isString).map(item => fs.promises.stat(item as string)));
+  const totalExceedsMaxSize = stats.reduce((acc, stat) => acc + stat.size, 0) > sizeLimit;
```
mxschmitt, line 274: note: if the size is too big, over multiple files, we fail to convert it to JSON (when it gets stringified, and are not able to send it to the Chromium CDP transport)

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -271,8 +271,9 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
...
   const hasLargeFile = !!stats.find(s => s.size > sizeLimit);
-  if (hasLargeFile) {
+  if (hasLargeFile || totalExceedsMaxSize) {
```
dgozman, line 276: It looks like `hasLargeFile` is redundant now.

dgozman (approved):

PR #23384, opened 2023-05-30
docs: add option clarifications

### LINKED ISSUE #23355: [BUG] Documentation missing defaults for some BrowserContext options
The following applies to the API docs for all languages.

For new Browser Context options, many of the options don't have documented defaults or the behaviour is not clear:

- Base URL 
- Bypass CSP
- Extra HTTP headers (defaults to None)
- Locale
- No viewport
- Permissions
- Proxy
- Record HAR URL filter
- Strict selectors
- Timezone ID
- User agent

Some of the formatting is inconsistent, for example in Python the message for `ignore_https_errors` is "defaults to `false`" and for `has_touch` is "defaults to false". Neither are valid as the type in Python is `False`.
There is inconsistency in the use of quotes for string literal defaults in the Python API. e.g., in `record_har_content` the default is "`full`" and in `forced_colors` it is `'none'`

aslushnikov:
Fixes https://github.com/microsoft/playwright/issues/23355

### DIFF
--- a/packages/playwright-core/src/utils/glob.ts
+++ b/packages/playwright-core/src/utils/glob.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-const escapeGlobChars = new Set(['/', '$', '^', '+', '.', '(', ')', '=', '!', '|', '?']);
+const escapeGlobChars = new Set(['/', '$', '^', '+', '.', '(', ')', '=', '!', '|']);
 
 export function globToRegex(glob: string): RegExp {
   const tokens = ['^'];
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/utils/glob.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-const escapeGlobChars = new Set(['/', '$', '^', '+', '.', '(', ')', '=', '!', '|']);
+const escapeGlobChars = new Set(['/', '$', '^', '+', '.', '(', ')', '=', '!', '|', '?']);
```
mxschmitt, line 17: looks like a stray change!

dgozman: +1

dgozman (approved):

PR #23417, opened 2023-06-01
fix(pause): ignore `page.pause` on the server

### LINKED ISSUE #23051: [Feature] Disable pause() when running playwright tests on remote browsers

Sometimes 'pause()' is left unintentionally in the code. When the tests are run on remote browser, they don't get executed. 
If tests  are run on remote browser, automatically disable pause()

Yes

dgozman:
Fixes #23051.

### DIFF
--- a/packages/playwright-core/src/server/debugger.ts
+++ b/packages/playwright-core/src/server/debugger.ts
@@ -120,6 +120,8 @@ export class Debugger extends EventEmitter implements InstrumentationListener {
 }
 
 function shouldPauseOnCall(sdkObject: SdkObject, metadata: CallMetadata): boolean {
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/client/page.ts
```diff
@@ -681,6 +681,8 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
   async pause() {
     if (require('inspector').url())
       return;
+    if (this._connection.isRemote())
+      return;
```
gskierk, line 685: Wouldn't any console.log about skipped execution be helpful in this case? Then the user will know that there are some unexpected pauses in the code.

yury-s (approved):

pavelfeldman (approved):

PR #23447, opened 2023-06-01
chore: add Playwright to attribution

dgozman:
This makes it easier to plumb all kinds of options around.


packages/playwright-core/src/androidServerImpl.ts
```diff
@@ -24,7 +24,7 @@ import { PlaywrightServer } from './remote/playwrightServer';
 
 export class AndroidServerLauncherImpl {
   async launchServer(options: LaunchAndroidServerOptions = {}): Promise<BrowserServer> {
-    const playwright = createPlaywright('javascript');
+    const playwright = createPlaywright({ sdkLanguage: 'javascript', isRemote: true });
```
pavelfeldman, line 27: `isServer` ?

pavelfeldman (approved):

PR #23475, opened 2023-06-02
docs: Soft expect example + tobeAttached

debs-obrien:
- add an example for soft expect
- add toBeAttached to table

### DIFF
--- a/docs/src/test-webserver-js.md
+++ b/docs/src/test-webserver-js.md
@@ -18,6 +18,8 @@ export default defineConfig({
     command: 'npm run start',
     url: 'http://127.0.0.1:3000',
     reuseExistingServer: !process.env.CI,
+    stdout: 'ignore',
+    stderr: 'pipe',
   },
 });
 ```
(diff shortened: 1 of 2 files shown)


docs/src/test-webserver-js.md
```diff
@@ -18,6 +18,8 @@ export default defineConfig({
     command: 'npm run start',
     url: 'http://127.0.0.1:3000',
     reuseExistingServer: !process.env.CI,
+    stdout: 'ignore',
```
aslushnikov, line 21: this is part of https://github.com/microsoft/playwright/pull/23474 - so probably should be reverted from this PR

aslushnikov (approved):

aslushnikov:
Note: you mentioned "toBeAttached" in the table, but I didn't find it

debs-obrien:
> Note: you mentioned "toBeAttached" in the table, but I didn't find it

we don't seem to give many examples for our assertions. I guess we presume people either understand them or will read api docs so we should discuss what is best here as in just table or add examples for them all

debs-obrien:
as discussed in meeting we will leave table with link to api docs for example

PR #23493, opened 2023-06-03
fix: scroll timeline with multiple pages in context

### LINKED ISSUE #23490: [BUG] Reproducible film strip layout bug in trace viewer and UI mode
Re-filing with a repro as suggested in #23304. I can only reproduce this when analyzing the page contents using [@axe-core/playwright](https://github.com/dequelabs/axe-core-npm).

### System info
- Playwright Version: v1.34.3
- Operating System: macOS
- Browser: All
- Other info:

### Source code

[scitech/playwright-filmstrip-issue](https://github.com/scitech/playwright-filmstrip-issue)

or

**Steps**
- `yarn && yarn playwright test --ui`

**Expected**

The timeline is laid out in a single row

**Actual**

Blank thumbnail images are displayed below the main timeline. Here is an example from the repro case:
image

When more pages are analyzed with axe-core, the filmstrip panel obstructs the rest of the panels making them unusable.

scitech:
Fixes #23490 

Before:
image

After:
image

### DIFF
--- a/packages/trace-viewer/src/ui/timeline.css
+++ b/packages/trace-viewer/src/ui/timeline.css
@@ -23,6 +23,8 @@
   cursor: text;
   user-select: none;
   margin-left: 10px;
+  max-height: 20vh;
+  overflow-y: scroll;
 }
 
 .timeline-divider {
--- a/packages/trace-viewer/src/ui/timeline.css
+++ b/packages/trace-viewer/src/ui/timeline.css
@@ -23,7 +23,7 @@
   cursor: text;
   user-select: none;
   margin-left: 10px;
-  max-height: 20vh;
+  max-height: 200px;
   overflow-y: scroll;
 }
 


packages/trace-viewer/src/ui/timeline.css
```diff
@@ -23,6 +23,8 @@
   cursor: text;
   user-select: none;
   margin-left: 10px;
+  max-height: 20vh;
```
aslushnikov, line 26:
```suggestion
  max-height: 200px;
```

aslushnikov:
Thank you for the fix! I'd suggest `200px` instead of `20vh`, and it's good to go.

scitech:
> Thank you for the fix! I'd suggest `200px` instead of `20vh`, and it's good to go.

Update made in 27ec925cf43cbfdc6488d495abc97ece39e7d55e. Thanks!

aslushnikov (approved):

PR #23500, opened 2023-06-03
chore: wait for downloads getting removed on context.close()

### LINKED ISSUE #22525: [internal]: unflake 'should delete downloads when persistent context closes' on WebKit

mxschmitt:
Before there was a race, that we ran into this code:

https://github.com/microsoft/playwright/blob/9cd49d5dd5b4074fdfa66b463b0a3379e73bcd69/packages/playwright-core/src/server/browserContext.ts#L236-L237

and then into this code:

https://github.com/microsoft/playwright/blob/9cd49d5dd5b4074fdfa66b463b0a3379e73bcd69/packages/playwright-core/src/server/browserContext.ts#L429-L431

which had the side effect, that the first call did not wait. Then immediately clears the downloads Set and then the second call is a NOOP. This ends up that the the removal of the downloads can happen after the context is closed, hence the test is flaky.

Relates to https://github.com/microsoft/playwright/pull/6151 where it got introduced. So something for @yury-s.

Fixes https://github.com/microsoft/playwright/issues/22525

### DIFF
--- a/packages/playwright-core/src/server/browserContext.ts
+++ b/packages/playwright-core/src/server/browserContext.ts
@@ -232,9 +232,12 @@ export abstract class BrowserContext extends SdkObject {
       // at the same time.
       return;
     }
+    const gotClosedGracefully = this._closedStatus === 'closing';
     this._closedStatus = 'closed';
-    this._deleteAllDownloads();
-    this._downloads.clear();
+    if (!gotClosedGracefully) {
+      this._deleteAllDownloads();
+      this._downloads.clear();
+    }
     this.tracing.dispose().catch(() => {});
     if (this._isPersistentContext)
       this.onClosePersistent();


packages/playwright-core/src/server/browserContext.ts
```diff
@@ -232,9 +232,12 @@ export abstract class BrowserContext extends SdkObject {
       // at the same time.
       return;
     }
+    const gotClosedGracefully = this._closedStatus === 'closing';
     this._closedStatus = 'closed';
-    this._deleteAllDownloads();
-    this._downloads.clear();
+    if (!gotClosedGracefully) {
```
yury-s, line 237: This way if the browser gets killed when browser.close() started (e.g. during tracing.dispose()) we may end up not deleting the files at all.

I don't understand how it helps with the test flakiness. AFAIU the problem is that we don't wait for the deletions to complete here but assume it in the test.

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -232,9 +232,12 @@ export abstract class BrowserContext extends SdkObject {
       // at the same time.
       return;
     }
+    const gotClosedGracefully = this._closedStatus === 'closing';
     this._closedStatus = 'closed';
-    this._deleteAllDownloads();
```
pavelfeldman, line 236: Should we just delete this line?

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -232,9 +232,12 @@ export abstract class BrowserContext extends SdkObject {
...
-    this._downloads.clear();
```
pavelfeldman, line 237: (well, and this one)

packages/playwright-core/src/server/browserContext.ts
yury-s: As discussed offline let's remove these two lines completely. If the browser is killed by a signal we'll leak the artifacts but it should be fine as it may happen in other scenarios where the browser is killed too.

yury-s (approved):

yury-s (approved):

PR #23511, opened 2023-06-04
Playwright as a Chromium extension

ruifigueira:
I've been working on a PoC to compile playwright into a chromium extension.

I was able to make it work (with some limitations, of course).

Here is a small video of the recorder as a chrome devtools panel:

https://github.com/microsoft/playwright/assets/1374559/ad9ae9b7-a4f5-4abf-b04e-3356f2f6f777

The core of this integration is a `ConnectionTransport` implementation (see [crxTransport.ts](https://github.com/ruifigueira/playwright-crx/blob/crx/packages/crx/src/crx/crxTransport.ts)) that mostly uses [chrome.debugger API](https://developer.chrome.com/docs/extensions/reference/debugger).

This API is per tab (we need to provide chrome tab ID to attach it), so for now it can only handle one tab at a time.

Besides, some CDP methods are not supported under `chrome.debugger`, so I had to emulate them.

To build the chrome extension:

```
npm ci
npm run build
```

Then, In chrome, just add an unpacked extension pointing to `packages/playwright-core/lib/webpack/crx`.

image

After that, it should be available in devtools as a panel with name **Playwright Recorder**.
Try to record and inspect some elements to ensure everything works fine.

Let me know yout thoughts!


ruifigueira:

ruifigueira:
Well, it doesn't make much sense to have a PR at this point, I opened a feature request instead #23514.

Closing.

PR #23536, opened 2023-06-06
chore: allow opening ui mode over http

pavelfeldman:


dgozman (approved):

packages/playwright-test/src/runner/uiMode.ts
```diff
@@ -81,49 +81,58 @@ class UIMode {
...
+        queue = queue.then(() => this._queueListOrRun(method, params));
+        await queue;
+      },
+
+      onclose: () => exitPromise.resolve(),
+    };
+    await showTraceViewer([], 'chromium', {
+      app: 'uiMode.html',
+      headless: isUnderTest() && process.env.PWTEST_HEADED_FOR_TEST !== '1',
+      transport: this._transport,
```
mxschmitt, line 122: we should directly have `--ui-web-host` here since Codespaces requires to have `0.0.0.0` as a listening one.

PR #23547, opened 2023-06-06
chore: do not show the open error

pavelfeldman:


packages/playwright-core/src/server/trace/viewer/traceViewer.ts
```diff
@@ -185,7 +185,7 @@ async function openTraceInBrowser(traceUrls: string[], options?: Options) {
   // eslint-disable-next-line no-console
   console.log('\nListening on ' + url);
   // eslint-disable-next-line no-console
```
Smrtnyk, line 187: I guess this lint exclusion is redundant now since you removed log?

PR #23555, opened 2023-06-06
feat: add `maskColor` option to the `toHaveScreenshot` method

aslushnikov:


docs/src/api/params.md
```diff
@@ -1093,7 +1093,7 @@ Specify locators that should be masked when the screenshot is taken. Masked elem
 a pink box `#FF00FF` that completely covers its bounding box.
```
dgozman, line 1093: While you are here, let's mention `maskColor` here instead of a "pink box".

dgozman (approved):

Issue #23540, opened 2023-06-06
[BUG] HTML elements with >> break Playwright


mkawumba:
### System info
- Playwright Version: [v1.32]
- Operating System: [Windows 11, macOS 13.2]
- Browser: [All]

**Config file**

```js
// playwright.config.ts
import { PlaywrightTestConfig, devices } from '@playwright/test';
import { join } from 'path';
import { isCI } from './utils';
const CI = isCI();

// Use process.env.PORT by default and fallback to port 3000
const PORT = process.env.PORT || 3000;

// Set webServer.url and use.baseURL with the location of the WebServer respecting the correct set port
const baseURL = `http://localhost:${PORT}`;

// Reference: https://playwright.dev/docs/test-configuration
const config: PlaywrightTestConfig = {
  globalSetup: require.resolve('./global-setup'),
  timeout: 30 * 1000,
  testDir: join(__dirname, 'tests'),
  forbidOnly: CI,
  retries: 1,
  workers: 1,
  outputDir: 'e2e-results/',
  reporter: [['list'], ['junit', { outputFile: './e2e-results/results.xml' }]],

  use: {
    baseURL,
    trace: 'retain-on-failure',
    screenshot: 'only-on-failure'
  },

  projects: [
    {
      name: 'API',
      testDir: 'tests/api',
    },
    {
      name: 'Desktop Chrome',
      testDir: 'tests/ui',
      use: {
        ...devices['Desktop Chrome'],
      },
    },
    {
      name: 'Desktop Safari',
      testDir: 'tests/ui',
      use: {
        ...devices['Desktop Safari'],
      },
    },
    // Test against mobile viewports.
    {
      name: 'Mobile Chrome',
      testDir: 'tests/ui',
      use: {
        ...devices['Pixel 5'],
      },
    },
    {
      name: 'Mobile Safari iPad',
      testDir: 'tests/ui',
      use: devices['iPad (gen 6)'],
    },
    {
      name: 'Mobile Safari iPhone',
      testDir: 'tests/ui',
      use: devices['iPhone 13'],
    },
  ],
};
export default config;

```

**Test file (self-contained)**

```js
```

**Steps**
- Run any test on the page with >> attributes

**Expected**

The HTML elements with any form of >> should works fine with Playwright including:
```js
&gt;&#x200B;&gt;
&gt;&gt; 
```
**Actual**

[...]

PR #23595, opened 2023-06-08
fix: regex selectors containing >> (#23540)

ruifigueira:
Only quotes were considered when splitting chained selectors

### DIFF
--- a/docs/src/other-locators.md
+++ b/docs/src/other-locators.md
@@ -906,7 +906,7 @@ document
   .querySelector('span[attr=value]')
 ```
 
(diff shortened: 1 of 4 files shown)


ruifigueira:
Hmm, it fails in selectors like `xpath=/html/body/section >> css=div >> text="hello"`.

Looking into `parseSelectorString`, which splits chaining selectors into its parts:

https://github.com/microsoft/playwright/blob/a40fa1d86c2a5bcd7be938da55d3f833e36caa82/packages/playwright-core/src/utils/isomorphic/selectorParser.ts#L180-L212

It handle the expression character by character, and this approach would be impossible to use to distinguish a normal xpath starting with `/` from a regular expression.

I'm closing this PR because this approach won't work,

PR #23609, opened 2023-06-09
feat(goto): better navigation error message

hoffination:
While this log message is only a sanity check, it is still beneficial to have a message that can better inform what is happening when this occurs.

### DIFF
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -682,7 +682,7 @@ export class Frame extends SdkObject {
       if (event.newDocument!.documentId !== navigateResult.newDocumentId) {
         // This is just a sanity check. In practice, new navigation should
         // cancel the previous one and report "request cancelled"-like error.
-        throw new NavigationAbortedError(navigateResult.newDocumentId, 'Navigation interrupted by another one');
+        throw new NavigationAbortedError(navigateResult.newDocumentId, `Navigation to "${url}" is interrupted by another navigation to "${event.url}"`);
       }
       if (event.error)
         throw event.error;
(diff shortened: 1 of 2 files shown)


hoffination:

packages/playwright-core/src/server/frames.ts
```diff
@@ -682,7 +682,7 @@ export class Frame extends SdkObject {
       if (event.newDocument!.documentId !== navigateResult.newDocumentId) {
         // This is just a sanity check. In practice, new navigation should
         // cancel the previous one and report "request cancelled"-like error.
-        throw new NavigationAbortedError(navigateResult.newDocumentId, 'Navigation interrupted by another one');
+        throw new NavigationAbortedError(navigateResult.newDocumentId, `Navigation interrupted by another one\n\n${url} changed to ${event.newDocument!.request!.url()}`);
```
dgozman, line 685:
```suggestion
        throw new NavigationAbortedError(navigateResult.newDocumentId, `Navigation to "${url}" is interrupted by another navigation to "${event.url}"`);
```

dgozman (commented):
Thank you for the PR! Could you please undo all the whitespace-only changes for cleaner history? See also for error message recommendation and a way to get the new `url` even without a request.

hoffination:
Thanks @dgozman ! I've updated the PR to remove the whitespace updates

dgozman:
@hoffination The code looks good, but there are more tests with the wrong error message that needs to be fixed. Take a look at the test report linked above.

hoffination:
@dgozman fixed the related tests but looks like other tests are failing.

dgozman (approved):
> @dgozman fixed the related tests but looks like other tests are failing.

The failures look unrelated to your change, so we'll handle them ourselves. Thank you for the PR, looks great, merging in.

PR #23626, opened 2023-06-09
chore: run html reporter tests with merged report

yury-s:


.github/workflows/tests_primary.yml
```diff
@@ -171,6 +171,28 @@ jobs:
         name: pull-request
         path: pull_request_number.txt
 
+  reporter_tests_with_blob_report:
```
dgozman, line 174: Let's do two describes instead of two bots.

tests/playwright-test/reporter-html.spec.ts
```diff
@@ -351,7 +351,6 @@ test('should use different path if attachments base url option is provided', asy
     `
   }, {}, { PW_TEST_HTML_REPORT_OPEN: 'never' });
   expect(result.exitCode).toBe(0);
-  expect(result.passed).toBe(1);
```
aslushnikov, line 354: let's automatically add `line` instead!

aslushnikov (approved):

PR #23642, opened 2023-06-11
docs(dotnet): add note about using assertions with library

mxschmitt:
Common question.

### DIFF
--- a/docs/src/library-csharp.md
+++ b/docs/src/library-csharp.md
@@ -58,7 +58,7 @@ await using var browser = await playwright.Firefox.LaunchAsync(new()
 
 ## Using Assertions
 
-You can do the following to leverage Playwright's assertions when you are using your own test framework:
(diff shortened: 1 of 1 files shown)


docs/src/library-csharp.md
```diff
@@ -56,6 +56,21 @@ await using var browser = await playwright.Firefox.LaunchAsync(new()
 });
 ```
 
+## Using Assertions
+
+You can do the following to leverage Playwright's assertions when you are using your own test framework:
```
yury-s, line 61: It needs a bit more context describing why the assertions are useful.

mxschmitt, line 61:
```suggestion
You can do the following to leverage Playwright's web-first assertions when you are using your own test framework. These will automatically retry until the condition is met, e.g. an element has a certain text or the timeout is reached:
```

mxschmitt: done

yury-s (approved):

PR #23691, opened 2023-06-13
resolves microsoft/playwright#23690

### LINKED ISSUE #23690: [Feature] make text-based snapshots cross-platform friendly
`toMatchSnapshot` fails to compare successfully on Windows machines when snapshots have been created on *NIX operating systems due to newline inequalities.

spcoder:
i think this code might do the trick to fix #23690.

normalizing both strings by replacing `\r\n` with `\n` should eliminate newline differences.


aslushnikov:
We can't land this as-is since it will break current clients.

Let's continue discussion in the issue first.

PR #23692, opened 2023-06-13
chore: optional root for generateSelector

dgozman:


packages/playwright-core/src/server/injected/domUtils.ts
```diff
@@ -46,9 +46,11 @@ function enclosingShadowHost(element: Element): Element | undefined {
   return parentElementOrShadowHost(element);
 }
 
-export function closestCrossShadow(element: Element | undefined, css: string): Element | undefined {
+export function closestCrossShadow(element: Element | undefined, css: string, scope?: Document | Element): Element | undefined {
   while (element) {
     const closest = element.closest(css);
+    if (scope && closest !== scope && closest?.contains(scope))
+      return;
     if (closest)
```
aslushnikov, line 54: given the function name, I'd expect a check that `closest` actually belongs to `scope`

aslushnikov (approved):

PR #23698, opened 2023-06-14
Allow for manual downloads

amosjyng:
Setting the `PLAYWRIGHT_DOWNLOAD_CONNECTION_TIMEOUT` environment variable does not work as expected for me because the download still stalls after approximately 30 seconds. I have not managed to debug why this is, but I figure it would be good to allow for manual workarounds on slow internet connections.

At the very least, I would appreciate some way of making it easier for me to fix the download issue myself by including better logging messages.

### DIFF
--- a/packages/playwright-core/src/utils/network.ts
+++ b/packages/playwright-core/src/utils/network.ts
@@ -82,7 +82,7 @@ export function httpRequest(params: HTTPRequestParams, onResponse: (r: http.Inco
   request.on('error', onError);
   if (timeout !== undefined) {
     const rejectOnTimeout = () =>  {
-      onError(new Error(`Request to ${params.url} timed out after ${timeout}ms`));
+      onError(new Error(`Request to ${params.url} timed out after ${timeout}ms. Consider setting PLAYWRIGHT_DOWNLOAD_CONNECTION_TIMEOUT if you haven't already.`));
       request.abort();
     };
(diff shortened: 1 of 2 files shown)


amosjyng:

packages/playwright-core/src/utils/network.ts
```diff
@@ -82,7 +82,7 @@ export function httpRequest(params: HTTPRequestParams, onResponse: (r: http.Inco
   request.on('error', onError);
   if (timeout !== undefined) {
     const rejectOnTimeout = () =>  {
-      onError(new Error(`Request to ${params.url} timed out after ${timeout}ms`));
+      onError(new Error(`Request to ${params.url} timed out after ${timeout}ms. Consider setting PLAYWRIGHT_DOWNLOAD_CONNECTION_TIMEOUT if you haven't already.`));
```
victoriaquasar, line 85: Is this a really necessary message? In my opinion, information like that is more suitable in a documentation of FAQ

yury-s:
@amosjyng could you start with filing a bug that explains the problem you are trying to solve?

aslushnikov:
All in all, let's do a step back and start a discussion with an issue.

As for the change it self, I don't believe it is necessary since the relevant bits are already covered by documentation: https://playwright.dev/docs/next/browsers#install-behind-a-firewall-or-a-proxy

PR #23713, opened 2023-06-14
Improve type safety of `Serializable` type

OliverJAsh:
I need to fix the build but first I wanted to make sure that we're happy to make this change?

### DIFF
--- a/packages/playwright-core/types/structs.d.ts
+++ b/packages/playwright-core/types/structs.d.ts
@@ -19,7 +19,7 @@ import { JSHandle, ElementHandle, Frame, Page, BrowserContext } from './types';
 /**
  * Can be converted to JSON
  */
-export type Serializable = any;
+export type Serializable = unknown;
 /**
  * Can be converted to JSON, but may also contain JSHandles.
  */


packages/playwright-core/types/structs.d.ts
```diff
@@ -19,7 +19,7 @@ import { JSHandle, ElementHandle, Frame, Page, BrowserContext } from './types';
 /**
  * Can be converted to JSON
  */
-export type Serializable = any;
+export type Serializable = unknown;
```
OliverJAsh, line 22: Alternatively, we can make this more precise, but I'm not sure if there's really any benefit:

```ts
export type Json = boolean | number | string | null | JsonArray | JsonRecord
export interface JsonRecord {
  readonly [key: string]: Json
}
export interface JsonArray extends ReadonlyArray<Json> {}
/**
 * Can be converted to JSON
 */
export type Serializable = Json;
```

scarletquasar: @OliverJAsh I think `JsonRecord` could be typed as:

```ts
type JsonRecord = Record<string, Json>
```

yury-s (commented):
Can you start with filing a bug that would explain the problem and why we need this change?

OliverJAsh:
> Can you start with filing a bug that would explain the problem and why we need this change?

I can file a bug but it's pretty simple. It's to protect against bugs like this:

```ts
const request = await page.waitForRequest('foo');
// ❌ No type error but at runtime we get an exception!
request.postDataJSON().oops.i.do.not.exist;
```

Using `unknown` forces you to validate/parse/narrow the type.

yury-s:
> 
> Using `unknown` forces you to validate/parse/narrow the type.

This indeed would be nice to fix, but I'm afraid that the ship has sailed because with this change we'll break compilation for existing clients that currently rely on the type to be `any` and don't have explicit type hast. Unless we can update the type without breaking existing clients we'd rather not change it.

OliverJAsh:
Perhaps you could consider this change for the next major version?

yury-s:
> Perhaps you could consider this change for the next major version?

Absolutely. There are no current plans for v2 though.

mxschmitt:
Closing as per above. I put this change into our internal notes for a potential v2.

PR #23717, opened 2023-06-14
chore(codegen): do not use accessible name for non-text selectors

dgozman:
Accessible name usually includes text, so we don't want it for non-text selectors, e.g. for `expect()` selectors.


packages/playwright-core/src/server/injected/selectorGenerator.ts
```diff
@@ -175,18 +179,19 @@ function generateSelectorFor(injectedScript: InjectedScript, targetElement: Elem
 function buildCandidates(injectedScript: InjectedScript, element: Element, options: GenerateSelectorOptions, accessibleNameCache: Map<Element, boolean>): SelectorToken[] {
```
aslushnikov, line 179: remove `accessibleNameCache`

aslushnikov (approved):

PR #23720, opened 2023-06-14
feat(blob): zip .jsonl report files

yury-s:
For linux tests without tracing blob-report-1.zip takes 19M, while unpacked size is 228 MB. That size is counted for GitHub artifact billing:

image

We zip individual .jsonl reports so that they still have unique names and can be easily uploaded into the same artifacts directory without name collisions.


dgozman (approved):

Chantellfarrell66 (commented):
#24534

Issue #23729, opened 2023-06-15
[BUG] Browsers do not download 


simon-greatrix:
This is a follow-on from https://github.com/microsoft/playwright/issues/23569 which has been closed.

As requested in issue 23569, here is the debug output from attempting an install:

```
Removing unused browser at /Users/simongreatrix/Library/Caches/ms-playwright/chromium-1067
  pw:install downloading Chromium 115.0.5790.24 (playwright build v1067) - attempt #1 +0ms
Downloading Chromium 115.0.5790.24 (playwright build v1067) from https://playwright.azureedge.net/builds/chromium/1067/chromium-mac.zip
  pw:install running download: +207ms
  pw:install -- from url: https://playwright.azureedge.net/builds/chromium/1067/chromium-mac.zip +0ms
  pw:install -- to location: /var/folders/fn/rj2mq_9947q0v4bbywbrgslh0000gr/T/playwright-download-chromium-mac13-1067.zip +0ms
  pw:install -- response status code: 200 +247ms
  pw:install -- total bytes: 136140840 +0ms
129.8 Mb [====================] 98% 0.5s  pw:install SUCCESS downloading Chromium 115.0.5790.24 (playwright build v1067) +22s
  pw:install extracting archive +1ms
  pw:install -- zip: /var/folders/fn/rj2mq_9947q0v4bbywbrgslh0000gr/T/playwright-download-chromium-mac13-1067.zip +0ms
  pw:install -- location: /Users/simongreatrix/Library/Caches/ms-playwright/chromium-1067 +0ms
  pw:install FAILED installation Chromium 115.0.5790.24 (playwright build v1067) with error: Error: end of central directory record signature not found +7ms
Failed to install browsers
Error: Failed to download Chromium 115.0.5790.24 (playwright build v1067), caused by
Error: end of central directory record signature not found
    at /Users/simongreatrix/tmp/node_modules/playwright-core/lib/zipBundleImpl.js:1:24033
    at /Users/simongreatrix/tmp/node_modules/playwright-core/lib/zipBundleImpl.js:1:31712
    at /Users/simongreatrix/tmp/node_modules/playwright-core/lib/zipBundleImpl.js:1:17288
    at FSReqCallback.wrapper [as oncomplete] (node:fs:682:5)
```

By editing the `browserFetch.js` to remove the clean-up I was able to check the file contents. In each case the file was too short. The bytes in the file were correct, there were just too few of them.

Over 8 downloads the actual file size on disk was:

[...]

PR #23739, opened 2023-06-15
chore: move extracting of browser out-of-process

mxschmitt:
https://github.com/microsoft/playwright/issues/23729

### DIFF
--- a/packages/playwright-core/src/server/registry/browserFetcher.ts
+++ b/packages/playwright-core/src/server/registry/browserFetcher.ts
@@ -44,12 +44,14 @@ export async function downloadBrowserWithProgressBar(title: string, browserDirec
         debugLogger.log('install', `SUCCESS installing ${title}`);
         break;
       }
+      if (await existsAsync(zipPath))
+        await fs.promises.unlink(zipPath);
+      if (await existsAsync(browserDirectory))
+        await fs.promises.rmdir(browserDirectory, { recursive: true });
       const errorMessage = error?.message || '';
       debugLogger.log('install', `attempt #${attempt} - ERROR: ${errorMessage}`);
       if (attempt >= retryCount)
         throw error;
-      if (await existsAsync(zipPath))
-        await fs.promises.unlink(zipPath);
     }
   } catch (e) {
     debugLogger.log('install', `FAILED installation ${title} with error: ${e}`);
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/server/registry/browserFetcher.ts
```diff
@@ -19,15 +19,14 @@ import fs from 'fs';
...
-import { getUserAgent } from '../../utils/userAgent';
 import { existsAsync } from '../../utils/fileUtils';
 import { debugLogger } from '../../common/debugLogger';
-import { extract } from '../../zipBundle';
 import { ManualPromise } from '../../utils/manualPromise';
 import { colors } from '../../utilsBundle';
+import { browserDirectoryToMarkerFilePath } from '.';
 
 export async function downloadBrowserWithProgressBar(title: string, browserDirectory: string, executablePath: string | undefined, downloadURLs: string[], downloadFileName: string, downloadConnectionTimeout: number): Promise<boolean> {
-  if (await existsAsync(browserDirectory)) {
```
mxschmitt, line 30: this looked like a bug before, we didn't check for the marker file for some reason.

packages/playwright-core/src/server/registry/browserFetcher.ts
```diff
@@ -49,14 +48,8 @@ export async function downloadBrowserWithProgressBar(title: string, browserDirec
...
         throw error;
-    }
-    debugLogger.log('install', `extracting archive`);
-    debugLogger.log('install', `-- zip: ${zipPath}`);
-    debugLogger.log('install', `-- location: ${browserDirectory}`);
-    await extract(zipPath, { dir: browserDirectory });
-    if (executablePath) {
-      debugLogger.log('install', `fixing permissions at ${executablePath}`);
-      await fs.promises.chmod(executablePath, 0o755);
+      if (await existsAsync(zipPath))
```
mxschmitt, line 51: upper code got moved into out-of-process. The additional check in line 51 got added for safety precautions, since for a retry the file could potentially exist.

packages/playwright-core/src/server/registry/browserFetcher.ts
aslushnikov: Fair point. However, if the marker doesn't exist, you should remove the `browserDirectory` before trying to download something there.

aslushnikov (approved):

PR #23763, opened 2023-06-16
fix(trace viewer): prefer latest resource with the same url

### LINKED ISSUE #23709: [BUG] DOM snapshot does not collect styles
Playwright version: 1.36 TOT
Platform: macOS

Repro:

1. Start an Epic Stack website on `http://localhost:3000` website:
    ````
    git clone https://github.com/epicweb-dev/epic-stack.git
    cd epic-stack
    git checkout f21bbe93ce26b3ff391874a102e8a7c83319be11 # I verified that the issue reproduces on this commit
    cp .env.example .env
    npm install
    npm run setup -s
    npm run dev
    ````
  This should start a webserver you can navigate to.

2. Run the following Playwright snippet:

    ```ts
    import { chromium } from '@playwright/test';

    const browser = await chromium.launch({ headless: false });
    await context.tracing.start({ screenshots: true, snapshots: true });
    await page.goto('http://localhost:3000');
    await context.tracing.stop({ path: 'trace.zip' });
    await browser.close();
    ```

3. Open recorded trace:

    ```
    npx playwright show-trace trace.zip
    ```

The DOM snapshot for the `page.goto` does not record styles:

image

For the record: this is as per discussion in https://github.com/microsoft/playwright/issues/23613#issuecomment-1590258388

dgozman:
When rendering snapshot, disregard earlier resources with the same url, because it's most likely that the latest one was used for rendering.

An example would be reloading the page before the stylesheet has finished loading. In this case, the stylesheet will be requested twice, and the second copy that was not aborted should be used for the snapshot.

Fixes #23709.

### DIFF
--- a/tests/library/trace-viewer.spec.ts
+++ b/tests/library/trace-viewer.spec.ts
@@ -850,3 +850,38 @@ test('should open trace-1.31', async ({ showTraceViewer }) => {
   const snapshot = await traceViewer.snapshotFrame('locator.click');
   await expect(snapshot.locator('[__playwright_target__]')).toHaveText(['Submit']);
 });
+
+test('should prefer later resource request', async ({ page, server, runAndTrace }) => {
+  const html = `
+    <body>
+      <script>
+        const link = document.createElement('link');
+        link.rel = 'stylesheet';
+        link.href = 'style.css';
+        document.head.appendChild(link);
+
+        if (!window.location.href.includes('reloaded'))
+          window.location.href = window.location.href + '?reloaded';
+      </script>
+    </body>
+  `;
+
+  let reloadStartedCallback = () => {};
+  const reloadStartedPromise = new Promise<void>(f => reloadStartedCallback = f);
+  server.setRoute('/style.css', async (req, res) => {
+    // Make sure reload happens before style arrives.
+    await reloadStartedPromise;
+    res.end('body { background-color: rgb(123, 123, 123) }');
+  });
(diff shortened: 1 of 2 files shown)


tests/library/trace-viewer.spec.ts
```diff
@@ -850,3 +850,31 @@ test('should open trace-1.31', async ({ showTraceViewer }) => {
...
+      <script>
+        if (!window.location.href.includes('reloaded'))
+          window.location.href = window.location.href + '?reloaded';
+      </script>
+    </body>
+  `;
+
+  server.setRoute('/style.css', async (req, res) => {
+    // Make sure reload happens before style arrives.
+    await new Promise(f => setTimeout(f, 2000));
```
yury-s, line 869: just use a counter instead?

yury-s (approved):

andrecasal:
Wohoo, great work @dgozman 🎉 Thanks for this!

BryantUng-at:
I am seeing a similar issue to https://github.com/microsoft/playwright/issues/23709 for what I'm testing with playwright and was wondering if this change is in playwright-test@next?

mxschmitt:
Yes it should be live with our [canary releases](https://playwright.dev/docs/canary-releases).

BryantUng-at:
Thanks! I tried this for an application I'm testing and still seeing this issue but can confirm that the bug that reported it was fixed.  I'll file a new bug and try to provide info where I can.

PR #23759, opened 2023-06-16
docs: improve example for mock API

debs-obrien:
- improve example by adding full working example that someone can copy and paste into their code and run the test
- renamed the test data to be more unique and make it easier to understand
- moved links to advanced guide to after examples
- minor text improvements

PR #23791, opened 2023-06-19
docs: improve HAR example

debs-obrien:
improve the HAR example and explanation:
- other langs not added yet as will depend on feedback from js examples first

### DIFF
--- a/docs/src/network.md
+++ b/docs/src/network.md
@@ -706,10 +706,10 @@ await Page.RouteAsync("**/title.html", async route =>
 
 ## Record and replay requests
 
-You can record network activity as an HTTP Archive file (HAR). Later on, this archive can be used to mock responses to the network requests. You'll need to:
+A HAR file is an HTTP Archive file that contains a record of all the network requests that are made when a page is loaded. It contains information about the request and response headers, cookies, content, timings, and more. You can use HAR files to mock network requests in your tests. You'll need to:
 1. Record a HAR file.
-1. Commit the HAR file alongside the tests.
-1. Route requests using the saved HAR files in the tests.
+2. Commit the HAR file alongside the tests.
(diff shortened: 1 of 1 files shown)


docs/src/network.md
```diff
@@ -791,6 +791,28 @@ var context = await browser.NewContextAsync(new() {
...
+### Recording HAR from a test
+
+To record a HAR file we use the `routeFromHAR` method. This method takes in the path to the HAR file and an optional object of options.
+
+The options object can contain the URL so that only requests with the URL matching the specified glob pattern will be served from the HAR File. If not specified, all requests will be served from the HAR file.
+
+The `update` option updates the given HAR file with the actual network information instead of serving from the file. In order to record the HAR file, you need to set `update` to true.
+
+```js
+test('Record a HAR', async ({
```
dgozman, line 803: We don't usually split test header into multiple lines in other examples. Let's be consistent.

docs/src/network.md
```diff
@@ -826,6 +848,25 @@ page.route_from_har("example.har")
 await context.RouteFromHARAsync("example.har");
 ```
 
+### Mocking with HAR
+
+You can modify the response from the HAR file and then run your tests against the modified response.
```
dgozman, line 853: Not sure how the snippet below is different from the previous sections, nor how does it highlight "you can modify the response" part?

dgozman (commented):
Should we move HAR record/replay into mocking APIs doc?

docs/src/network.md
debs-obrien: yeh I need to add that they need to modify the .txt file to modify the response, will fix

debs-obrien:
closing as this is now dealt with in this pr: https://github.com/microsoft/playwright/pull/23759

PR #23796, opened 2023-06-20
fix(chromium): can get correct orientation angle on non-mobile devices

### LINKED ISSUE #23772: [BUG] Chromium screen orientation angle is incorrect
### System info
- Playwright Version: [v1.35.1]
- Operating System: [macOS 13.2]
- Browser: [Chromium]
- Other info:

**Description**
I used `MobileDatePicker`, a component from [MUI](https://mui.com/x/react-date-pickers/date-picker/), in a desktop website. It displays differently in local Chrome and in Playwright Chromium (use `Desktop Chrome`, and `viewport = { width: 1200, height: 800 }`).

**Expected**
It is the same in Playwright Chromium as in the local Chrome, as below:

image

**Actual**
It displays in Playwright Chromium as below:

image

**Investigate**
`MobileDatePicker` displays differently according to `screen.orientation.angle`, and Playwright Chromium set the incorrect angle:

https://github.com/microsoft/playwright/blob/e171194c86a8c86a598e86a7f04abc0c3e786d34/packages/playwright-core/src/server/chromium/crPage.ts#L1011-L1021

According to [Screen Orientation API Spec](https://w3c.github.io/screen-orientation/#the-current-screen-orientation-type-and-angle), the angle is also associated with the [natural](https://w3c.github.io/screen-orientation/#dfn-natural) orientation.

image

image

Besides, in my understanding, `isLandscape` depends on screen size, not viewport size, is it right? 🤔

Refer to the solution for Webkit, the `screenOrientation` should be always `{ angle: 0, type: 'landscapePrimary' }` for not mobile devices, but the logic for mobile devices is similar as below:

https://github.com/microsoft/playwright/blob/e171194c86a8c86a598e86a7f04abc0c3e786d34/packages/playwright-core/src/server/webkit/wkPage.ts#L718-L721

If you think it makes sense, I'd like to submit a PR to apply it.

zhengjitf:
Fix #23772

### DIFF
--- a/packages/playwright-core/src/server/chromium/crPage.ts
+++ b/packages/playwright-core/src/server/chromium/crPage.ts
@@ -1008,7 +1008,7 @@ class FrameSession {
       return;
     const viewportSize = emulatedSize.viewport;
     const screenSize = emulatedSize.screen;
-    const isLandscape = viewportSize.width > viewportSize.height;
+    const isLandscape = screenSize.width > screenSize.height;
     const metricsOverride: Protocol.Emulation.setDeviceMetricsOverrideParameters = {
       mobile: !!options.isMobile,
       width: viewportSize.width,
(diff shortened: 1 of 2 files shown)


zhengjitf:

dgozman (approved):
Great work! Thank you for the PR, merging in.

PR #23825, opened 2023-06-21
chore: publish markdown report as a comment

yury-s:


.github/workflows/create_test_report.yml
```diff
@@ -48,6 +48,10 @@ jobs:
       run: |
         npx playwright merge-reports --reporter html --attachments missing blob-report
 
+    - name: Merge into Markdown Report
```
dgozman, line 51: Why not combine with the previous command?
```
npx playwright merge-reports --reporter markdown,html --attachments missing blob-report
```

dgozman (approved):

PR #23876, opened 2023-06-23
chore: keep UI Mode running when used with browser mode

### LINKED ISSUE #23801: UI Mode: don't kill process when WS connections gets closed

mxschmitt:
This will keep UI Mode running in browser mode. When launched in normal persistent context mode, we know when the persistent context closes, so we can run the project teardown code.

Fixes https://github.com/microsoft/playwright/issues/23801

### DIFF
--- a/packages/playwright-core/src/server/trace/viewer/traceViewer.ts
+++ b/packages/playwright-core/src/server/trace/viewer/traceViewer.ts
@@ -33,25 +33,16 @@ export type Transport = {
   onclose: () => void;
 };
 
-type Options = {
+export type OpenTraceViewerOptions = {
   app?: string;
   headless?: boolean;
   host?: string;
   port?: number;
   isServer?: boolean;
-  openInBrowser?: boolean;
   transport?: Transport;
 };
 
-export async function showTraceViewer(traceUrls: string[], browserName: string, options?: Options): Promise<void> {
-  if (options?.openInBrowser) {
-    await openTraceInBrowser(traceUrls, options);
-    return;
-  }
-  await openTraceViewerApp(traceUrls, browserName, options);
-}
-
-async function startTraceViewerServer(traceUrls: string[], options?: Options): Promise<{ server: HttpServer, url: string }> {
+async function startTraceViewerServer(traceUrls: string[], options?: OpenTraceViewerOptions): Promise<{ server: HttpServer, url: string }> {
   for (const traceUrl of traceUrls) {
     let traceFile = traceUrl;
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/trace/viewer/traceViewer.ts
```diff
@@ -171,8 +179,6 @@ export async function openTraceViewerApp(traceUrls: string[], browserName: strin
 
   if (isUnderTest())
     page.on('close', () => context.close(serverSideCallMetadata()).catch(() => {}));
-  else
```
mxschmitt, line 174: we should probably guard that behind an option?

packages/playwright-test/src/runner/uiMode.ts
```diff
@@ -82,18 +82,13 @@ class UIMode {
...
   async showUI(options: { host?: string, port?: number }) {
-    const exitPromise = new ManualPromise();
     let queue = Promise.resolve();
 
     this._transport = {
       dispatch: async (method, params) => {
         if (method === 'ping')
           return;
 
-        if (method === 'exit') {
```
mxschmitt, line 93: This was unused before.

mxschmitt:
- inline showTraceViewer into both places
- do teardown when CTRL + C gets called

packages/playwright-core/src/server/trace/viewer/traceViewer.ts
mxschmitt: moved over to program.ts

packages/playwright-test/src/runner/uiMode.ts
```diff
@@ -117,16 +112,23 @@ class UIMode {
...
       transport: this._transport,
       host: options.host,
       port: options.port,
-      openInBrowser: options.host !== undefined || options.port !== undefined,
-    });
+    };
+    const exitPromise = new ManualPromise<void>();
+    if (options.host !== undefined || options.port !== undefined) {
+      await openTraceInBrowser([], openOptions);
+      process.on('SIGINT', () => exitPromise.resolve());
```
dgozman, line 127: This doesn't really work nicely with running tests. Perhaps not do this and file a bug instead? I think missing teardown is worse than interfering with test run.

dgozman (approved):

packages/playwright-test/src/runner/uiMode.ts
mxschmitt: https://github.com/microsoft/playwright/issues/23907 for further reference.

Issue #5501, opened 2021-02-19
[BUG] Can't run Playwright in Nix


VaasRamsay:
**Context:**
- Playwright Version: 1.8.0
- Operating System: Linux
- Node.js version: 12
- Browser: All

**Describe the bug**

Running Playwright in Nix Shell throws `missing dependencies` error. Even after explicitly adding the missing dependencies suggested by "launch doctor", the same error is thrown. 

```
Error: browserType.launch: Host system is missing dependencies!
--
  | Missing libraries we didn't find packages for:
  | libexpat.so.1
  | libxshmfence.so.1
```

Probably, I'm hitting this error from [here](https://github.com/microsoft/playwright/blob/a9faa9c941fa0b0383d68dff2fb7ef825bd258ea/src/server/validateDependencies.ts#L250). Nix has it's own `ldconfig` and it's not found in `sbin`.

aslushnikov:
@SamBreaksThings we don't support Nix at the moment and this is the first request we've received regarding it so far. 

So until there's a huge demand for Nix support, this will be low-pri for us. Is there something easy we can do to enable you use Playwright on Nix?

uri-canva:
Is it possible to make the linked workaround configurable: https://github.com/microsoft/playwright/blob/a9faa9c941fa0b0383d68dff2fb7ef825bd258ea/src/server/validateDependencies.ts#L247-L250?

Note the commit message of https://github.com/aslushnikov/playwright/commit/aa3019bc8225abd463602481e3362fc61ea33b40

> To the best of my knowledge, `ldconfig` is always located at `/sbin`.

This is one case where it's not located there, and loading it from there leads to broken behaviour.

joscha:
> Is it possible to make the linked workaround configurable:

maybe via an env var `await spawnAsync(process.env.LDCONFIG_PATH || '/sbin/ldconfig', ['-p'], {});`

? Or if that is an unwanted API it would also be thinkable to make it a fallback from the generic `ldconfig` (w/o qualified path, e.g. `command -v ldconfig` or `which ldconfig`) and if that one is not found to fall back to `/sbin/ldconfig`. Or if we want to make the speed sacrifice for more exotic users of this package if could be in the catch block of the `/sbin/ldconfig` spawn and introspect the error and retry with `ldconfig` from path?

uri-canva:
[...]

PR #23889, opened 2023-06-25
fix: Allow portable bash paths for systems including NixOS

PatrickShaw:
Although very common, bash is not guaranteed to be located at `/bin/bash`. NixOS is an example of this.

More commonly, `/bin/bash` can be quite out of date. An example of this is MacOS's version of `bash`. This realistically won't affect Playwright but it's worth noting. You can technically update MacOS's system version of bash but you need elevated permissions to do so.

By using `/usr/bin/env bash` instead of `/bin/bash` we can execute Playwright's bash scripts in like NixOS and generally improve the selection behaviour for bash in other systems too.

Some discussion of why it's worth favouring `/usr/bin/env bash` over `/bin/bash`:
- Discusses `/bin/bash` missing in NixOS: https://discourse.nixos.org/t/add-bin-bash-to-avoid-unnecessary-pain/5673
- Some general commentary on why `/usr/bin/env bash` is favoured: https://askubuntu.com/a/1402721
- Points out how old bash is in MacOS: https://itnext.io/upgrading-bash-on-macos-7138bd1066ba

Improves situation at #5501

### DIFF
--- a/browser_patches/roll_from_upstream.sh
+++ b/browser_patches/roll_from_upstream.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # A script to roll browser patches from internal repository.
 
 set -e
--- a/browser_patches/webkit/pw_run.sh
+++ b/browser_patches/webkit/pw_run.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 function runOSX() {
   # if script is run as-is
--- a/browser_patches/winldd/archive.sh
+++ b/browser_patches/winldd/archive.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 set +x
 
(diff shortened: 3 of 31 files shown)


PatrickShaw:

AfroEuro:
> Although very common, bash is not guaranteed to be located at `/bin/bash`. NixOS is an example of this.
> 
> More commonly, `/bin/bash` can be quite out of date. An example of this is MacOS's version of `bash`. This realistically won't affect Playwright but it's worth noting. You can technically update MacOS's system version of bash you need elevated permissions to do so.
> 
> By using `/usr/bin/env bash` instead of `/bin/bash` we can execute Playwright's bash scripts in like NixOS and generally improve the selection behaviour for bash in other systems too.
> 
> Some discussion of why it's worth favouring `/usr/bin/env bash` over `/bin/bash`:
> - Discusses `/bin/bash` missing in NixOS: https://discourse.nixos.org/t/add-bin-bash-to-avoid-unnecessary-pain/5673
> - Some general commentary on why `/usr/bin/env bash` is favoured: https://askubuntu.com/a/1402721
> - Points out how old bash is in MacOS: https://itnext.io/upgrading-bash-on-macos-7138bd1066ba
> 
> Improves situation at #5501

- [ ] []()

aslushnikov (approved):

PR #23904, opened 2023-06-26
docs(python): fix syntactically incorrect snippets

mxschmitt:


docs/src/api/class-locator.md
```diff
@@ -966,19 +966,13 @@ rowLocator
 ```python async
 row_locator = page.locator("tr")
 # ...
-await row_locator
-    .filter(has_text="text in column 1")
-    .filter(has=page.get_by_role("button", name="column 2 button"))
-    .screenshot()
+await row_locator.filter(has_text="text in column 1").filter(has=page.get_by_role("button", name="column 2 button")).screenshot()
```
dgozman, line 969: Did you run this through the formatter? We should set something like a 80-char limit for it.

mxschmitt: Without its just not valid Python: http://black.readthedocs.io/en/latest/

debs-obrien (approved):

PR #23919, opened 2023-06-27
docs: fix missing broken Node.js code snippets

mxschmitt:


docs/src/protractor-js.md
```diff
@@ -98,26 +98,26 @@ Here's how to polyfill `waitForAngular` function in Playwright Test:
...
+        `);
+      }
+
+      await executeScriptAsync(page, clientSideScripts.waitForAngular, '');
     }
     ```
-    If you don't want to keep a version protractor around, you can also use this simpler approach using this function (only works for Angular 2+):
+    If you don't want to keep a version protractor around, you can also use this simpler approach using this function (only works for Angular 2+):/
     ```js
     async function waitForAngular(page) {
```
debs-obrien, line 122: Why the / here?

docs/src/service-workers-experimental-network-events-js.md
```diff
@@ -180,40 +180,41 @@ Consider the code snippets below to understand Playwright's view into the Reques
...
-      if (event.request.url.endsWith('tracker.js'))
-        return new Response('console.log('no trackers!')', {
-          status: 200,
-          headers: { 'Content-Type': 'text/javascript' },
-        });
-
-      // 4. Otherwise, fallthrough, perform the fetch and respond
-      return fetch(event.request);
-    })()
+        const response = await caches.match(event.request);
```
debs-obrien, line 197: Does indent here make sense?

mxschmitt: fixed

docs/src/protractor-js.md
mxschmitt: fixed, by accident.

Issue #7275, opened 2021-06-23
[Feature] expose @playwright/test runner API


bgotink:
We're currently using jest as a test runner for testing angular applications using playwright. In the angular ecosystem tools usually integrate with the `@angular/cli` by providing a builder, so our `@ngx-playwright/jest` package is actually what powers the `ng e2e` command in our repositories.

It would be great if we could write such an integration for the `@playwright/test` runner as well, but that'd require access to the runner APIs.
While it would be possible to execute the playwright CLI in a new process, this would make certain integrations harder or hacky. For instance passing configuration from the angular.json into the test runner would require turning those into CLI arguments and passing the base URL the angular app is hosted at into the runner would require a workaround using the child process's environment to pass this along.

dgozman:
Is there an example of what this API would look like? I see that you'd like to pass "base URL", so this seems like somewhat deeper integration than just "run tests based on this configuration file". Maybe you could point to your existing integration with jest for us to take a look?

bgotink:
Jest has quite a lot of extension points, our test integration provides a `TestEnvironment`, which then sets up the global environment the test has access to (e.g. we provide `page`, `browserContext`, `browser` etc as globals). This environment
 itself can access the jest config, which contains the base URL.
The test environment can be found here: https://github.com/bgotink/ngx-playwright/blob/main/packages/jest/src/environment.ts

Jest actually has a configuration value called "testURL" they use in their jsdom test environment, we reused that one and pass it to jest here: https://github.com/bgotink/ngx-playwright/blob/main/packages/jest/src/builder/builder.ts#L54-L64

For `@playwright/test` things could probably be a lot easier, as the test runner already takes care of setting up the entire browser, context and page.
A function where we pass in the configuration (from the config file) and overrides (the CLI options) which then gets executed goes a long way to providing the necessary API. A separate function to read the config, so the caller of the test API doesn't have to tackle typescript conversion, would be useful to add.
This setup would allow us to add an extra fixture with the base URL in it to the tests, giving our other fixtures access to this variable.

If a watch-mode would be added to the test runner, being able to hook into that and rerun the tests when the angular app finishes rebuilding on a change would be nice to have.

dgozman:
@bgotink Thank you, we'll take a look.

bgotink:
The addition of the `baseURL` fixture made implementing this via child_process actually very easy to do. We've now switched from jest to playwright-test and it has been a breeze.

[...]

PR #23921, opened 2023-06-27
feat(page): Pick best locator for existing element.

### LINKED ISSUE #13900: [Feature] SelectorGenerator as standalone package
Hi,
The selector-generator component which is injected as part of the codegen mode has significant utility outside of playwright.
I'd really like to use it stand-alone. I think there's a lot of projects which would like to also.

Would you consider publishing this independely, or providing a code-sample which uses just this alone?

Thanks!

Germandrummer92:
Improve programmatic API of accessing the "pick locator" functionality as exposed by the playwright trace viewer/codegenerator.

Fixes #13900

I am not sure if this is the both the best way to implement it (sorry in advance, first time contributing here) nor the best API, so would love some feedback on both.

Overall: it would be super helpful for us (octomind) to programmatically be able to pick the "best" locator/selector for a certain HTMLElement we have in-memory already. Our use-case is a little different than the one referenced in the issue: we want to be able to auto-generate playwright tests using some form of AI/smart exploration. To do that effectively we would like to be able to choose the "best" selector/locator for a certain element that we already hold in-memory from our exploration process so that the resulting test case is as robust as possible. 

If you need more information we can go also go into more detail. We in general would love to help give playwright a programmatic API if allowed , e.g. test execution and introspection of test results is something we are very interested in as well (see https://github.com/microsoft/playwright/issues/7275).

Re: this specific change: I was looking to build an API that works on a certain HTMLElement directly, but found it a) difficult to get that through the serialization between client/server and b) thought that a selector -> "better" selector API might be more generic anyways.

### DIFF
--- a/tests/page/locator-convenience.spec.ts
+++ b/tests/page/locator-convenience.spec.ts
@@ -17,6 +17,37 @@
 
 import { test as it, expect } from './pageTest';
 
+it.describe('picking the best available locator for an existing element', () => {
+
+  const html = '<div><input id="inputId" placeholder="somePlaceholder"/><button id="buttonId">Some Text</button></div>';
+
+  const testCases = [
+    {
+      originalSelector: '#buttonId',
+      expectedSelector: 'internal:role=button[name="Some Text"i]',
+      expectedLocator: "getByRole('button', { name: 'Some Text' })"
+    },
+    {
+      originalSelector: '#inputId',
+      expectedSelector: 'internal:attr=[placeholder="somePlaceholder"i]',
+      expectedLocator: "getByPlaceholder('somePlaceholder')"
+    }];
+
+  for (const testCase of testCases) {
+    it(`should allow selecting the best locator from an existing selector ${testCase.originalSelector}`, async ({ page }) => {
+      await page.setContent(html);
+
+      const locatorReference = await page.pickBestLocator(testCase.originalSelector);
+
+      expect(locatorReference.locator).toEqual(testCase.expectedLocator);
+      expect(locatorReference.selector).toEqual(testCase.expectedSelector);
+      await page.close();
+    });
+  }
+
+});
+
+
 it('should have a nice preview', async ({ page, server }) => {
   await page.goto(`${server.PREFIX}/dom.html`);
   const outer = page.locator('#outer');
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -845,6 +845,39 @@ Gets the full HTML contents of the page, including the doctype.
 
 Get the browser context that the page belongs to.
 
+## method: Page.pickBestLocator
+* since: v1.36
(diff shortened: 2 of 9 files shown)


Germandrummer92:
@microsoft-github-policy-service agree company="OctoMind GmbH"

Germandrummer92:
> How do you determine the actual element on the page and how do you integrate it with Playwright?
> 
> When you set the `PWDEBUG=console` env var, we expose the Playwright scripts to the frames. Then you can do:
> 
> ```ts
> const element = document.querySelector('body > div > h1');
> const locator = playwright.generateLocator(element);
> console.log(locator);
> ```

Calling it from the browser-controlling process we could just "evaluate" it correct?
```ts
await page.setContent(
  '<html lang="en"><body><button id="button">Some Text</button></body></html>'
);
const locator = await page.locator("#button").evaluate((e) => {
  return playwright.generateLocator(e);
});
```

For my understanding: this only works with starting the browser in ui-mode (so non-headless), correct? While the injected script/this PR implementation could run on a headless one, right? 

We were hoping to not really have to run the browser with UI during test generation/discovery, but we will see if that works for us, I will leave this open for now and comment back later if it works.


Re your question: we want to control the browser during test discovery/generation and predict elements we will interact with, that means of course we have the full DOM and elements in-memory and can relatively easily come up with "some" selector that works ok on our own (currently by just injecting test-id attributes) but want the "best"/"most reliable" selector for the generated test case in the end -> we would like to use playwright code for that since rolling our own "selector hierachy" will likely be worse and get outdated from upstream than what you do.

pavelfeldman:
Closing as per above, please feel free to open a new issue if this does not cover your use case.

Germandrummer92:
Hi @pavelfeldman @mxschmitt sorry for the late reply, just as a fyi:

unfortunately the solution as provided by max does not seem to work in headless mode, it requires setting PWDEBUG=console which will launch the browser WITH ui. It's somewhat okay for us, but does require quite a bit of workarounds to make it work in a server-/serverless environment.

tests/page/locator-convenience.spec.ts
```diff
@@ -17,6 +17,37 @@
 
 import { test as it, expect } from './pageTest';
 
+it.describe('picking the best available locator for an existing element', () => {
```
dopecoder, line 20: How does using a selector which depends on the content of the website such as name of buttons, placeholders more stable than id, data-testid etc.?

Germandrummer92: well, ids are often changing, in some frameworks even for each reload.

data-testids are of course the most stable, but in our use case (see [octomind](www.octomind.dev)) we don't control the actual source of the website we're testing and don't want our customer to have to make changes.

also, but not a very strong argument over testids, if accesibility is of concern to you it of course helps to see if people are breaking the accessible aria-labels etc. if you use them in your tests.

PR #23928, opened 2023-06-27
fix: do not wait for navigations after certain actions

dgozman:
Do not wait for navigations anymore:
- `check`/`uncheck`/`setChecked`
- `fill`/`type`/`press`
- `selectOption`
- `setInputFiles`
- `dragTo`/`dragAndDrop`

Keep waiting for navigations:
- `click`/`hover`/`dblclick`
- `tap`


pavelfeldman (commented):
I'd make it easier to revert. I'm anticipating a pushback to the change, so we should have an environment variable readily available at least.

PR #23932, opened 2023-06-27
fix(route): handle escapes in the glob pattern

### LINKED ISSUE #23303: [BUG] Escaping ? in glob pattern does not work
### System info
- Playwright Version: [v1.33.2]
- Browser: Chromium

### Source code
Run the below code on https://try.playwright.tech/
The test will fail but it should pass.

```js

test('basic test', async ({ page }) => {
  // Set up the route stub
  await page.route('**/task*', (route) => {
    route.fulfill({
      status: 200,
      contentType: 'application/json',
      body: JSON.stringify({ message: 'YES' }),
    });
  });

  // Set up the route stub that matches query params only
  //  (escaping \? should match a literal question mark)
  // This stub should not be triggered when calling the route below
  // as the route called has no question mark
  await page.route('**/task\?*', (route) => {
    route.fulfill({
      status: 200,
      contentType: 'application/json',
      body: JSON.stringify({ message: 'NO' }),
    });
  });

  await page.evaluate(() => {
    fetch('https://api.example.com/task/test');
  });

  const response = await page.waitForResponse(/.+task.*/);
  const responseBody = await response.json();
  expect(responseBody.message).toEqual('YES')
});
```

dgozman:
Fixes #23303.

### DIFF
--- a/packages/playwright-core/src/utils/glob.ts
+++ b/packages/playwright-core/src/utils/glob.ts
@@ -14,15 +14,17 @@
  * limitations under the License.
  */
 
-const escapeGlobChars = new Set(['/', '$', '^', '+', '.', '(', ')', '=', '!', '|']);
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions#escaping
+const escapedChars = new Set(['$', '^', '+', '.', '*', '(', ')', '|', '\\', '?', '{', '}', '[', ']']);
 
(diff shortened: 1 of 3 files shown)


mxschmitt (commented):
Looks like a change that needs to be done in the language ports. Please add it to the corresponding backporting issues.

pavelfeldman (approved):

PR #23945, opened 2023-06-28
fix(route): make sure fallback overrides are recorded in har/trace

### LINKED ISSUE #23793: [BUG] `route.fulfill` discards request headers overrides from previous `route.fallback` calls
### System info
- Playwright Version: v1.35
- Operating System: All
- Browser: All
- Other info:

### Source code

I used the test suite at `tests/library/har.spec.ts` to reproduce the issue, so you can run it with `npm run ctest -- --grep "should record request overrides on fulfill"`. Here is the failing test:

```js
it('should record request overrides on fulfill', async ({ contextFactory, server }, testInfo) => {
  const { page, getLog } = await pageWithHar(contextFactory, testInfo);

  await page.route('**/foo', async route => {
    await route.fulfill({
      status: 200,
      json: {
        fulfilled: true
      },
    });
  });

  await page.route('**/foo', route => {
    void route.fallback({
      headers: {
        'fallback-request-header': '1'
      },
    });
  });

  await page.goto(server.PREFIX + '/foo');
  const log = await getLog();

  const request = log.entries[0].request;
  expect(request.headers).toContainEqual({ name: 'fallback-request-header', value: '1' });
});
```

**Steps**
- `npm run ctest -- --grep "should record request overrides on fulfill"`

**Expected**

- `'fallback-request-header': '1'` should be part of the request headers

**Actual**

- All request headers overrides are discarded

dgozman:
Fixes #23793.

### DIFF
--- a/tests/library/har.spec.ts
+++ b/tests/library/har.spec.ts
@@ -860,3 +860,32 @@ it('should not hang on slow chunked response', async ({ browserName, browser, co
   expect(log.browser.name.toLowerCase()).toBe(browserName);
   expect(log.browser.version).toBe(browser.version());
 });
+
+it('should record request overrides on fulfill', async ({ contextFactory, server }, testInfo) => {
+  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/23793' });
+  const { page, getLog } = await pageWithHar(contextFactory, testInfo);
+
+  await page.route('**/foo', async route => {
+    await route.fulfill({ status: 200, json: { fulfilled: true } });
+  });
+  await page.route('**/bar', async route => {
+    await route.abort();
+  });
+
+  await page.route('**/*', route => {
+    void route.fallback({
+      headers: {
+        'fallback-request-header': '1'
+      },
+    });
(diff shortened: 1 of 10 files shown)


tests/library/har.spec.ts
```diff
@@ -860,3 +860,32 @@ it('should not hang on slow chunked response', async ({ browserName, browser, co
...
+  });
+
+  await page.goto(server.PREFIX + '/foo');
+  await page.evaluate(() => fetch('/bar').catch(() => {}));
+  const log = await getLog();
+
+  expect(log.entries[0].request.url).toBe(server.PREFIX + '/foo');
+  expect(log.entries[0].request.headers).toContainEqual({ name: 'fallback-request-header', value: '1' });
+  expect(log.entries[1].request.url).toBe(server.PREFIX + '/bar');
+  expect(log.entries[1].request.headers).toContainEqual({ name: 'fallback-request-header', value: '1' });
```
pavelfeldman, line 890: I'm not sure I follow - I ended up aborting the request, why am I seeing anything in the har log?

PR #23953, opened 2023-06-29
docs(dotnet): add-cancellation-feature for to code generation

maw136:
supporting this PR: https://github.com/microsoft/playwright-dotnet/pull/2597


mxschmitt:
Would having the CancellationToken in the normal options work as well? We are following these semantics in the other places where we have optional values.

maw136:
I have pushed the changes, generates CancellationToken that is non-nullable (with proper default) value

mxschmitt:
Closing as per https://github.com/microsoft/playwright/issues/24083#issuecomment-1640761745.

PR #23985, opened 2023-06-30
chore: migrate builtin reporters to ReporterV2

### LINKED ISSUE #23539: [BUG] WebServer stdout and stderr "pipe" does not stream output
[BUG] WebServer stdout and stderr "pipe" does not stream output as per the documentation which makes debugging impossible.

### System info
- Playwright Version: 1.34.3
- Operating System: osx
- Browser: All
- Other info:

### Source code

https://github.com/andrewmclagan/playwright-std-out

```
pnpm test:e2e
```

You will notice output is only displayed _**after**_ the web-server's URL is reachable and is not streamed/piped to `stdout` as per the configuration setting. This makes it very difficult to debug when the web-server has a possible long startup, or there are errors. Forcing the developer to wait for the `webServer.timeout`

dgozman:
This allows builtin reporters to handle stdio between onConfigure and onBegin.

Fixes #23539.

### DIFF
--- a/packages/playwright-test/src/reporters/base.ts
+++ b/packages/playwright-test/src/reporters/base.ts
@@ -16,9 +16,10 @@
 
 import { colors, ms as milliseconds, parseStackTraceLine } from 'playwright-core/lib/utilsBundle';
 import path from 'path';
-import type { FullConfig, TestCase, Suite, TestResult, TestError, FullResult, TestStep, Location, Reporter } from '../../types/testReporter';
+import type { FullConfig, TestCase, Suite, TestResult, TestError, FullResult, TestStep, Location } from '../../types/testReporter';
 import type { SuitePrivate } from '../../types/reporterPrivate';
 import { monotonicTime } from 'playwright-core/lib/utils';
+import type { ReporterV2 } from './reporterV2';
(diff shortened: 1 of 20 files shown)


packages/playwright-test/src/reporters/base.ts
```diff
@@ -104,6 +115,19 @@ export class BaseReporter implements Reporter {
     this.result = result;
   }
 
+  onStepBegin(test: TestCase, result: TestResult, step: TestStep): void {
```
pavelfeldman, line 118: Why do we have these, aren't they optional?

packages/playwright-test/src/reporters/empty.ts
```diff
@@ -14,9 +14,50 @@
  * limitations under the License.
  */
 
-import type { Reporter } from '../../types/testReporter';
+import type { ReporterV2 } from './reporterV2';
+import type { FullConfig, TestCase, TestError, TestResult, FullResult, TestStep, Suite } from '../../types/testReporter';
 
-class EmptyReporter implements Reporter {
+class EmptyReporter implements ReporterV2 {
+  onConfigure(config: FullConfig) {
```
pavelfeldman, line 21: ditto

pavelfeldman (approved):

Issue #23514, opened 2023-06-04
[Feature] Playwright as a Chromium extension


ruifigueira:
I've been working on a PoC to compile playwright into a chromium extension. You can find my fork here: https://github.com/ruifigueira/playwright-crx.

I was able to make it work (with some limitations, of course).

Here is a small video of the recorder as a chrome devtools panel:

https://github.com/microsoft/playwright/assets/1374559/ad9ae9b7-a4f5-4abf-b04e-3356f2f6f777

The core of this integration is a `ConnectionTransport` implementation (see [crxTransport.ts](https://github.com/ruifigueira/playwright-crx/blob/crx/packages/crx/src/crx/crxTransport.ts)) that mostly uses [chrome.debugger API](https://developer.chrome.com/docs/extensions/reference/debugger).

This API is per tab (we need to provide chrome tab ID to attach it), so for now it can only handle one tab at a time.

Besides, some CDP methods are not supported under `chrome.debugger`, so I had to emulate them.

To build the chrome extension:

```
npm ci
npm run build
```

Then, In chrome, just add an unpacked extension pointing to `packages/playwright-core/lib/webpack/crx`.

![image](https://github.com/microsoft/playwright/assets/1374559/0bad5bc5-9f10-4cd7-adca-c26d01ec2115)

After that, it should be available in devtools as a panel with name **Playwright Recorder**.
Try to record and inspect some elements to ensure everything works fine.

Let me know yout thoughts!

pavelfeldman:
@ruifigueira this looks very impressive! What kind of feedback are you looking for?

In terms of priorities, we like to generate code into the editor (VS Code), but given how ubiquitous Chrome DevTools are, it feels like a nice addition to the suite. So if we can figure out a good maintenance story and cover it with enough tests, we can accept the contribution!

I think the major obstacle is the collection of the polyfills (vite-plugin-node-polyfills). Your approach is great - it is minimally intrusive to Playwright code, but at the same time it is the most brittle. I wonder if there is a middle ground where we don't pollute our code with too many unnecessary abstractions, yet we can compile our code with the explicit platform externs. Think explicit platformBundle that is implemented via the node internals and mocked by the browser). It might easily get out of hand, but I would at least be interested in what it could look like.

ruifigueira:
@pavelfeldman thanks for the feedback.

Honestly, I think this approach for shimming / polyfilling those thirdparty dependencies is not that bad.
[...]

PR #24020, opened 2023-07-04
feat: playwright-crx, the chrome extensions flavor of playwright

ruifigueira:
As discussed in #23514, here's the PR for **playwright-crx**, a version of playwright that can run in a chrome extension service worker.

For that, it relies on `chrome.debugger.sendCommand` / `chrome.debugger.onEvent` to implement the `ConnectionTransport` interface.

## API

It's possible to use `@playwright/experimental-crx` as a library to create new chrome extensions.

Here's a simple example of a background service worker for a chrome extension using **playwright-crx**:

```ts
import { _crx } from '@playwright/experimental-crx';

const crxPromise = _crx.start({ slowMo: 500 });

chrome.action.onClicked.addListener(async ({ id: tabId }) => {
  const crx = await crxPromise;

  // tries to connect to the active tab, or creates a new one
  const page = await crx.attach(tabId!).catch(() => crx.newPage());

  await page.goto('https://demo.playwright.dev/todomvc/#/');
  await page.getByPlaceholder('What needs to be done?').click();
  await page.getByPlaceholder('What needs to be done?').fill('Hello World!');
  await page.getByPlaceholder('What needs to be done?').press('Enter');

  // page stays open, but no longer controlled by playwright
  await crx.detach(page);
});
```

A more complete example can be found in `example/todomvc-crx`.

### API Tests

Most of **playwright-crx** tests run in a chrome extension found in `tests/crx/test-extension`.
Its background service worker receives a serialized test method with some fixtures (limited at the time, see `crxFixtureRunner.ts`) and runs it.

To run tests in `crx` mode, I created a new npm script: `npm run xtest`.

Even with several limitations (it's similar to `page.evaluate`), it already passes more than 1k tests.

Nevertheless, this is something that requires a big improvement:

- tests won't run if a function or variable is declared outside the test function
- tests run slowly because it needs to relaunch a new persistent context with our testing chrome extension for each test
- fixtures are very limited at the moment

## Recorder

I decided to integrate the recorder chrome extension into `packages/recorder`.

This way, the app will keep working as before, but it can now also be installed as a chrome extension.

The only change to the original app is just a content script for communication with the extension background service worker when it detects it's running in a chrome extension context.

In terms of chrome extension functionality, it provides:

- [action button](https://developer.chrome.com/docs/extensions/reference/action/) for attaching current tab into recorder (it opens the recorder if it's closed)
- [context menu](https://developer.chrome.com/docs/extensions/reference/contextMenus/) for the same purpose
- pages must be explicitly attached to be recordable, except if they are opened from already attached pages
- closing the recorder window will detach all pages and uninstall injected scripts (highlights and event listeners)

Here's an updated video (the app I used to record the video has a bug with the mouse position):

https://github.com/microsoft/playwright/assets/1374559/3c67d830-3eb2-48f5-b409-28354c101b87

### Recorder Tests

Tests at `tests/library/inspector` also run against recorder as a chrome extension when in `crx` mode.

It launches a persistent context with our recorder chrome extension and then attach to its tab using a chrome playwright page.

Note that it uses playwright chrome flavor and not crx flavor to automate these specific tests, but recorded instructions are captured by `crx`. See `inspectorTests.ts` for more details.

## Library Bundles / Require

Most of `require`s were replaced with corresponding `import`s, except playwright bundles. For those, I created a vite plugin to replace requires from corresponding `BundleImpl` with imports in compile time

See `packages/playwright-crx/vite.config.ts` for more details.

## Shims

For **playwright-crx** to compile and run, I had to rely on shims / polyfills.

Here is the list of current shim libraries it uses (declared in `packages/playwright-crx/package.json`):

- [@isomorphic-git/lightning-fs](https://github.com/isomorphic-git/lightning-fs)
- [assert](https://github.com/browserify/commonjs-assert)
- [browserify-zlib](https://github.com/browserify/browserify-zlib)
- [buffer](https://github.com/feross/buffer)
- [crypto-browserify](https://github.com/browserify/crypto-browserify)
- [events](https://github.com/browserify/events)
- [https-browserify](https://github.com/substack/https-browserify) - unavailable
- [os-browserify](https://github.com/CoderPuppy/os-browserify)
- [path](https://github.com/jinder/path)
- [process](https://github.com/defunctzombie/node-process)
- [readable-stream](https://github.com/nodejs/readable-stream)
- [setimmediate](https://github.com/yuzujs/setImmediate)
- [stream-http](https://github.com/jhiesey/stream-http)
- [url](https://github.com/defunctzombie/node-url)
- [util](https://github.com/browserify/node-util)

I also created some shims under `packages/playwright-crx/src/shims`, mostly with `noop` functions.

Refs: #23514


ruifigueira:
@pavelfeldman I didn't address the explicit platformBundle yet because I wanted to keep code changes minimal at this point, and with some decent code coverage.

I noticed (the hard way, CI not passing) that you already have a mechanism to control code dependencies, probably we can extend it to also take into account node libraries. Nevertheless, those changes will always be very disruptive (splitting and moving code around) so I think those changes should come in future PRs.

Let me know your thoughts

ruifigueira:
Meanwhile, I'm doing a vite plugin to avoid the `utilsBundleRequire` mess, so stay tuned.

ruifigueira:
vite plugin done. I also did a major rebase and combined / dropped commits that no longer made sense.

pavelfeldman:
As I mentioned in https://github.com/microsoft/playwright/issues/23514#issuecomment-1583077762 I don't think we'll be comfortable shipping something that uses https://github.com/niksy/node-stdlib-browser#package-contents in production. But again, as I mentioned in the same comment, we will accept the code health patches that might enable you implementing your functionality in a third party repo!

ruifigueira:
Thanks @pavelfeldman, I totally understand your concerns. And my goal is to avoid using any node shim / polyfill, but that appears to be harder than I initially thought, so this is my first approach to have things working and tested, to ensure I'm not breaking anything in both sides, and then starting to address that decoupling, one step at a time.

But just to manage my expectations and if it's worth the effort, if I figure out a clean way to only rely on isomorphic code, would this be acceptable to merge?

ruifigueira:
Just an update: I found out a way to remove dependencies using tree shaking and making certain code unreachable. For instance:

```ts
function writeFile(file: string, text: string) {
  if (process.env.PW_CRX) throw new Error(`Operation allowed in CRX mode`);
  fs.writeFileSync(this._file, this._text);
}
``` 

will not include the `fs.writeFileSync` function after compilation with `process.env.PW_CRX=1`. I expect code changes will be minimal this way.

I just did this `http` and `https` modules, and I no longer need those `https-browserify` and `stream-http`. Most of the others are easier to replace.

I also look into `utilsBundle` and `zipBundle` libraries, and several ones are isomorphic and the ones that are not are used in very specific places, so a similar approach will do.

Modules that will be a challenge:

- [fs](https://github.com/isomorphic-git/lightning-fs)
- [buffer](https://github.com/feross/buffer)
- [events](https://github.com/browserify/events)
- [path](https://github.com/jinder/path)
- [stream](https://github.com/nodejs/readable-stream)

Do you think any of those libraries are acceptable to include? For instance, `readable-stream` is from **NodeJS** team and some of the others (`events` and `path`) are copies from the node version.

ruifigueira:
First iteration on removing polyfill dependencies.

I didn't remove all of them yet, but I already implemented the strategy:

- I created a shim generator for node modules that we don't want to include. The idea of it is to clearly mark in the vite compiled library that calls for those modules are being made (see `utils/crx/generate_node_shims.js`)
- I also create a checker that ensures that no excluded node module is called in the final compiled library (see `utils/crx/check_deps.js`)
- based on the outcome of the checker, I forced node related code to be unreachable (see my previous coment https://github.com/microsoft/playwright/pull/24020#issuecomment-1626348338)
- I also excluded non-isomorphic libraries from the crx version of existing bundles

I tried to keep changes on current code base as small as possible.

pavelfeldman:
Closing as non-actionable

573dave:
> the app I used to record the video has a bug with the mouse position

May I suggest ShareX? https://github.com/ShareX/ShareX

ruifigueira:
I did use ShareX 😁

PR #24073, opened 2023-07-06
feat: ability to use selenium settings via launch options

DudaGod:
## What's done

Implement ability to specify selenium settings via launch options in pwt config or standalone script. It will allow to configure the project more conveniently and in one place without the need to use environment variables. For backwards compatibility, I didn't delete env vars and give them the highest priority (to be able to redefine some options in CI for example).

### What problems does it solve

- can't launch several chromium with different capabilities (each browser use the same `SELENIUM_REMOTE_CAPABILITIES`);
- using environment variables is not convenient, especially if you need to pass objects.

### DIFF
--- a/docs/src/selenium-grid.md
+++ b/docs/src/selenium-grid.md
@@ -44,6 +44,36 @@ You don't have to change your code, just use your testing harness or [`method: B
 
 When using Selenium Grid Hub, you can [skip browser downloads](./browsers.md#skip-browser-downloads).
 
+Besides this, you can specify the url using the launch options (has a lower priority than the environment variable).
+
+```js tab=js-test title="playwright.config.ts"
+import { defineConfig } from '@playwright/test';
+
+export default defineConfig({
+  projects: [
+    {
+      name: 'chromium',
+      use: {
+        launchOptions: {
+          selenium: {
+            url: 'http://localhost:4444/wd/hub',
+          },
+        },
+      },
+    },
+  ],
+});
+```
+
+```js tab=js-library
+const { chromium } = require('playwright');
+const browser = await chromium.launch({
+  selenium: {
+    url: 'http://localhost:4444/wd/hub',
+  },
+});
+```
+
 ### Passing additional capabilities
 
 If your grid requires additional capabilities to be set (for example, you use an external service), you can set `SELENIUM_REMOTE_CAPABILITIES` environment variable to provide JSON-serialized capabilities.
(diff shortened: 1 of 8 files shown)


pavelfeldman:
Selenium grid support is a highly experimental feature guarded by the environment flags in Playwright. It was added there to encourage experimentation and see if Selenium grid can be used as a generic browser backend for Playwright. It is unlikely that we'll be releasing it out of experimental though, so we can't introduce any APIs or configuration options that would be referencing it. Furthermore, the latest signals from the Selenium grid are that it positions itself as a Selenium grid, not a browser grid and as such it should not be used by Playwright (see https://github.com/SeleniumHQ/selenium/issues/11590#issuecomment-1436113950). So we are going to be unenrolling this experiment gragually.

DudaGod:
> Selenium grid support is a highly experimental feature guarded by the environment flags in Playwright

This is strange, because in [the documentation](https://playwright.dev/docs/selenium-grid) there is nothing about the fact that this is an experimental feature.

> Furthermore, the latest signals from the Selenium grid are that it positions itself as a Selenium grid, not a browser grid and as such it should not be used by Playwright (see https://github.com/SeleniumHQ/selenium/issues/11590#issuecomment-1436113950). So we are going to be unenrolling this experiment gragually.

But they said that they are ready for collaboration. Why we can't just support BiDi protocol when it will be released instead of just removing everything related to selenium?

sipayRT:
> So we are going to be unenrolling this experiment gragually.

For what? To lose your clients? This functionality greatly expands the possibilities of your tool. Our team can implement this feature ourselves (in pw or in selenium). We only need confirmation from your team that you will not remove this functionality

DudaGod:
@pavelfeldman can you answer our questions?

sipayRT:
@pavelfeldman @aslushnikov ping

pavelfeldman:
> This is strange, because in [the documentation](https://playwright.dev/docs/selenium-grid) there is nothing about the fact that this is an experimental feature.

We should have made it more clear that support for Selenium is experimental. We introduced it silently, but should have added a better disclaimer for it, we'll fix that.

> Why we can't just support BiDi protocol when it will be released instead of just removing everything related to selenium?

BiDi is not a CDP replacement, so you can't say "make Playwright work with BiDi". It would take several years of engineering work done by the browser vendors, counting from the time when BiDi considers supporting feature set sufficient for Playwright, if that even happens. So we are talking about ~5-10 years timeframe, if ever.

> This functionality greatly expands the possibilities of your tool. Our team can implement this feature ourselves (in pw or in selenium).

As long as Selenium Grid exposes CDP in a generic manner, we should be able to keep Playwright working against it. But given the public plans to remove CDP, we can't suggest our users to onboard the stack that is likely facing a dead end. So given their plans, we are hesitant growing the feature set, accepting patches like this, because it'll be sending the wrong signal to the customers.

Did you look into implementing Playwright Grid instead (and open sourcing it)? I know that Moon folks did something like that, running Playwright Servers instead of the Selenium Grid Agents should be straightforward.

link89:
@pavelfeldman I think it is unwise to reject this feature. Even though the Selenium community claims that they will replace BiDi in the future, I still doubt that they will give up CDP . Even though they dare to do it, there are other Selenium compatible solutions like [Selenoid](https://aerokube.com/selenoid/latest/) and [selenium-federation](https://github.com/link89/selenium-federation) that will support this feature. I believe that as long as someone needs it, there will definitely be solutions to support it.

diemol:
Removing CDP from Selenium Grid will take at least 1-2 years. 

As I mentioned in the comment referenced, the Selenium project is always open to collaboration. If leaving CDP enabled in Selenium Grid is a first step to enabling collaboration between projects, we can look into it. We would just love to hear from the Playwright folks and have a chat.

PR #24113, opened 2023-07-08
docs(ct): hooks config import

sand4rt:


pavelfeldman:

sand4rt:
👋 @pavelfeldman this should be merged for sure, the `hooksConfig` import path in the documentation is wrong. It was fine before but someone changed it. See:
`import type { HooksConfig } from '@playwright/test';`

docs/src/test-components-js.md
```diff
@@ -234,7 +234,7 @@ This includes any config that needs to be run before or after mounting the compo
 
   ```js title="src/pages/ProductsPage.spec.tsx"
   import { test, expect } from '@playwright/experimental-ct-react';
-  import type { HooksConfig } from '@playwright/test';
+  import type { HooksConfig } from 'playwright';
```
pavelfeldman, line 237: Most of our customers don't have `playwright` installed and also `playwright` has nothing to do with the test runner, so should not be aware of hooks configs. What are we fixing here?

sand4rt: This is referencing to `/{root-of-the-project/playwright` eg: https://github.com/microsoft/playwright/blob/main/tests/components/ct-react-vite/playwright/index.tsx#L5

But i see now how this is confusing, will adjust it

docs/src/test-components-js.md
```diff
@@ -234,7 +234,7 @@ This includes any config that needs to be run before or after mounting the compo
 
   ```js title="src/pages/ProductsPage.spec.tsx"
   import { test, expect } from '@playwright/experimental-ct-react';
-  import type { HooksConfig } from '@playwright/test';
+  import type { HooksConfig } from '../playwright';
```
pavelfeldman, line 237: There is no ../playwright in user's project, so this won't work either.

sand4rt: I think this is correct:

When initialising Playwright component testing, a `playwright/index.ts` file is created. In the documentation there is a `type HooksConfig` in this file.

Screenshot 2023-08-03 at 21 38 20

sand4rt:
Fixed by: https://github.com/microsoft/playwright/pull/26707/files

PR #24117, opened 2023-07-09
fix(ct-vue): allow use of props with null

tibnor:
During validation of props typeof null was assumed to be "null". typeof null is "object", therefore validation of props with null failed.

The PR adds tests in vue to capture the error.


tibnor:

sand4rt:
Hi @tibnor, i already removed all the type checks in this PR: https://github.com/microsoft/playwright/pull/24114

mxschmitt:
Looks not relevant anymore after https://github.com/microsoft/playwright/pull/24117#issuecomment-1627833890.

PR #24142, opened 2023-07-10
fix(trace viewer): break out of loop when finding a resource that matches url

fadezak100:
### Description:
I have been getting missed up styles in my iframes ever since that #23763 was merged. When I inspected the network I found that some of the styles resources requests come back with empty CSS files. 

**Reproduce the problem:** 
1. clone the playwright repo
2. `npm ci` and `npm run build`
3. cd packages/trace-viewer and `npm run preview`
4. drop and drag a trace.zip file

---

### Before:

missedupstyles

https://github.com/microsoft/playwright/assets/43090330/808a0c69-67d6-4f46-908b-630e9375520c

---

### After

Screen Shot 2023-07-11 at 1 12 16 AM

https://github.com/microsoft/playwright/assets/43090330/ea98d67a-193f-47de-8921-406551e7bee3

### DIFF
--- a/packages/trace-viewer/src/snapshotRenderer.ts
+++ b/packages/trace-viewer/src/snapshotRenderer.ts
@@ -123,9 +123,8 @@ export class SnapshotRenderer {
       if (resource._frameref !== snapshot.frameId)
         continue;
       if (resource.request.url === url) {
-        // Pick the last resource with matching url - most likely it was used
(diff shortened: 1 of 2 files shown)


dgozman:
@fadezak100 Could you please start with filing an issue, probably a regression?

dgozman:
> @fadezak100 Could you please start with filing an issue, probably a regression?

If you could share a trace file, and ideally a repro, that would be awesome.

PR #24169, opened 2023-07-12
chore: do not leak internal page handles after closing page

pavelfeldman:
Partial fix for https://github.com/microsoft/playwright/issues/6319

After this fix, the following scenario won't leak and the context state (cookies, storage, etc) can be reused by the new page sessions:

```js
  for (let i = 0; i < 1000; ++i) {
    await page.goto('...');
    await page.close('...');
  }
```

### DIFF
--- a/packages/playwright-core/src/server/dispatchers/elementHandlerDispatcher.ts
+++ b/packages/playwright-core/src/server/dispatchers/elementHandlerDispatcher.ts
@@ -26,7 +26,8 @@ import type { CallMetadata } from '../instrumentation';
 import type { WritableStreamDispatcher } from './writableStreamDispatcher';
 import { assert } from '../../utils';
 import path from 'path';
-import type { PageDispatcher } from './pageDispatcher';
+import { BrowserContextDispatcher } from './browserContextDispatcher';
+import { PageDispatcher, WorkerDispatcher } from './pageDispatcher';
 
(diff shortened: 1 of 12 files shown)


packages/playwright-core/src/server/dispatchers/elementHandlerDispatcher.ts
```diff
@@ -223,4 +224,20 @@ export class ElementHandleDispatcher extends JSHandleDispatcher implements chann
   async waitForSelector(params: channels.ElementHandleWaitForSelectorParams, metadata: CallMetadata): Promise<channels.ElementHandleWaitForSelectorResult> {
     return { element: ElementHandleDispatcher.fromNullable(this.parentScope(), await this._elementHandle.waitForSelector(metadata, params.selector, params)) };
   }
+
+  private _browserContextDispatcher(): BrowserContextDispatcher {
+    const scope = this.parentScope();
```
dgozman, line 229: Maybe a loop?

`while (scope && !(scope instanceof BrowserContextDispatcher)) scope = scope.parentScope()`

dgozman (approved):

packages/playwright-core/src/server/dispatchers/elementHandlerDispatcher.ts
pavelfeldman: The idea is that we keep it strongly-typed - it was wrong before.

PR #24194, opened 2023-07-12
test: request cookie order is same as in set-cookie

yury-s:


tests/library/browsercontext-fetch.spec.ts
```diff
@@ -218,6 +218,33 @@ it('should add cookies from Set-Cookie header', async ({ context, page, server }
...
+it('should preserve cookie order from Set-Cookie header', async ({ context, page, server, browserName, isLinux }) => {
+  it.info().annotations.push({ type: 'issue', description: 'https://github.com/microsoft/playwright/issues/23390' });
+  it.fixme(browserName === 'webkit' && isLinux);
+  server.setRoute('/setcookie.html', (req, res) => {
+    res.setHeader('Set-Cookie', ['cookie.0=foo', 'cookie.1=bar']);
+    res.end();
+  });
+  await page.request.get(server.PREFIX + '/setcookie.html');
+  const cookies = await context.cookies();
+  expect(cookies.map(c => ({ name: c.name, value: c.value }))).toEqual([
```
dgozman, line 230: I don't think we should guarantee the order in `context.cookies()`.

tests/library/browsercontext-fetch.spec.ts
```diff
@@ -218,6 +218,33 @@ it('should add cookies from Set-Cookie header', async ({ context, page, server }
...
+      name: 'cookie.1',
+      value: 'bar'
+    },
+  ]);
+  await page.goto(server.EMPTY_PAGE);
+  expect(await page.evaluate(() => document.cookie)).toEqual('cookie.0=foo; cookie.1=bar');
+  const requestPromise = server.waitForRequest('/empty.html');
+  await page.request.get(server.EMPTY_PAGE);
+  const request = await requestPromise;
+  expect(request.headers.cookie).toEqual('cookie.0=foo; cookie.1=bar');
```
dgozman, line 245: I guess this line does not work on webkit?

dgozman (approved):

tests/library/browsercontext-fetch.spec.ts
yury-s: Yeah, the test is failing there.

tests/library/browsercontext-fetch.spec.ts
yury-s: We don't provide such guarantees publicly, it's just an internal detail covered in this test.

PR #24222, opened 2023-07-14
devops: add upload report action

yury-s:
* Fix report downloading from Azure (reports are now zipped)
* Extracted upload logic into an action
* Extracted PR number file generation into its own job

### DIFF
--- a/.github/workflows/tests_primary.yml
+++ b/.github/workflows/tests_primary.yml
@@ -61,25 +61,12 @@ jobs:
     - run: ./utils/upload_flakiness_dashboard.sh ./test-results/report.json
       if: always()
       shell: bash
-    - name: Upload blob report to Azure
-      if: always() && github.event_name == 'push'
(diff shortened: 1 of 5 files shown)


.github/workflows/tests_primary.yml
```diff
@@ -61,25 +61,12 @@ jobs:
...
-        name: blob-report-${{ github.run_attempt }}
-        path: test-results/blob-report
-        retention-days: 30
-    - name: Write the pull request number in an file
-      if: always() && github.event_name == 'pull_request'
-      run: echo '${{ github.event.number }}' > pull_request_number.txt;
-    - name: Upload artifact with pull request number
-      uses: actions/upload-artifact@v3
-      if: always() && github.event_name == 'pull_request'
+    - name: Upload blob report
```
dgozman, line 64: I think we should also migrate `PWTEST_BLOB_REPORT` things inside our config, like our users would do.

.github/workflows/tests_primary.yml
```diff
@@ -245,3 +221,16 @@ jobs:
       if: always()
       shell: bash
 
+  write-pull-request-number:
```
dgozman, line 224: It would be ideal to make this part of the action, so that users would not need this additional job.

dgozman (approved):

PR #24224, opened 2023-07-14
docs(): add cache step to CI

ShaMan123:
Add a cache step to CI makes tests a lot faster!

### DIFF
--- a/docs/src/ci-intro-js.md
+++ b/docs/src/ci-intro-js.md
@@ -39,7 +39,15 @@ jobs:
           node-version: 18
       - name: Install dependencies
         run: npm ci
+      - name: Cache Playwright Browsers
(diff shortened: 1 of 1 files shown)


mxschmitt:
Thank you for your contribution to the project. We strongly discourage doing it so, see https://playwright.dev/docs/ci#caching-browsers and https://github.com/epicweb-dev/epic-stack/pull/204. Closing by that.

ShaMan123:
Thanks
Seems time is actually significant when caching, but I might be wrong - the errors sound bad

mxschmitt:
I responded to your PR, that you can do npx playwright install --with-deps chromium since you only run your tests on Chromium. This should give you a speed boost.

PR #24245, opened 2023-07-15
feat(ui): run deps in UI mode if dep projects are checked

### LINKED ISSUE #22045: [Feature] UI mode doesn't respect test dependencies
Running playwright test with a test that have a dependency. For example the below snippet will run the `setup` dependency when running `npx playwright test`. In UI mode I would reasonably expect the same - in that when running a test that has a dependency it would run the dependency first and then run the test. This is currently not the case.

```{
      name: 'loggedIn',
      use: {
        ...devices['Desktop Chrome'],
        storageState: STORAGE_STATE,
      },
      testIgnore: ['**/*login.spec.ts', '**/*.setup.ts'],
      dependencies: ['setup'],
    },```

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/22045


utils/limits.sh
dgozman, line 1: Intended change?

dgozman (approved):

utils/limits.sh
pavelfeldman: Very much so

amitasil:
Thanks!!

mikelhamer:
Yay!!! Thank you!

lattebread:
Thank you!

sarah-gelt:
Fantastic, thank you!

beig:
where can I find the option?
or what does 'dep projects are checked' mean?

currently my setup dependency is not run in UI mode

sarah-gelt:
> where can I find the option? or what does 'dep projects are checked' mean?
> 
> currently my setup dependency is not run in UI mode

Click in the area highlighted to expand the Projects list. Ensure your dependency projects are checked as well as the projects the tests are in. When you run a test, UI mode will then run the dependencies as well. Hope that helps!
image

scamden:
Thanks for this! One small follow up if you have time would be that if I specify `--project chromium` on the cli, ideally the dependent projects and teardowns would also be checked when I enter UI mode

 (this would mirror the non UI mode functionality of automatically running those projects if I select only the one)

scamden:
Also this line seems to be a little out of date since this was merged:
>  The UI mode will not take into consideration the setup tests and therefore you will have to manually run them first.

https://playwright.dev/docs/test-ui-mode#filtering-tests

Issue #7396, opened 2021-06-30
[BUG] element.type does not trigger key events with russian text


alex-popov-tech:
**Context:**
- Playwright Version: [1.12.2]
- Operating System: [Linux & Mac]
- Node.js version: [14.17.0]
- Browser: [Chromium]
- Extra: [any specific details about your environment]

```
## System:
 - OS: Linux 5.8 Ubuntu 20.04.2 LTS (Focal Fossa)
 - Memory: 15.31 GB / 31.30 GB
 - Container: Yes
## Binaries:
 - Node: 14.17.0 - /usr/bin/node
 - npm: 7.5.2 - /usr/local/bin/npm
## Languages:
 - Bash: 5.0.17 - /usr/bin/bash
## npmPackages:
 - playwright: 1.12.2 => 1.12.2 
```

```
## System:
 - OS: macOS 11.4
 - Memory: 670.91 MB / 32.00 GB
## Binaries:
 - Node: 15.3.0 - ~/.asdf/installs/nodejs/15.3.0/bin/node
 - Yarn: 1.22.10 - /usr/local/bin/yarn
 - npm: 7.0.14 - ~/.asdf/installs/nodejs/15.3.0/bin/npm
## Languages:
 - Bash: 3.2.57 - /bin/bash
## npmPackages:
 - playwright: 1.12.2 => 1.12.2 

```
**Code Snippet**

Help us help you! Put down a short code snippet that illustrates your bug and
that we can run and debug locally. For example:

```javascript
const {chromium, webkit, firefox} = require('playwright');

(async () => {
  const browser = await chromium.launch();
  const page = await browser.newPage();
      await page.goto('https://learn.javascript.ru/article/keyboard-events/keyboard-dump/');
      const checkbox = await page.$('input[name="keyupStop"]');
      await checkbox.click();
      const input = await page.$('#kinput');
      await input.type(' working ', { delay: 500, noWaitAfter: false, timeout: 15000 });
      await input.type(' не работает ', { delay: 500, noWaitAfter: false, timeout: 15000 });
})();
```

**Describe the bug**

When I try to type russian characters keyup/down events are not triggered

JoelEinbinder:
We currently hard code in a [US keyboard layout](https://github.com/microsoft/playwright/blob/master/src/server/usKeyboardLayout.ts). We can add in more layouts, but its a lot of work for each one so we've been waiting for a flood of user requests to prioritize it.

alexsorinsky:
[...]

PR #24249, opened 2023-07-16
feat: support for different keyboard layouts

ruifigueira:
Layouts were generated from scraping [kbdlayout.info](https://kbdlayout.info/) (see `generate_keyboard_layouts.js`).

For locale mapping, see https://github.com/microsoft/playwright/issues/7396#issuecomment-1637150391

There are some things to improve:

- how to select the keyboard layout (I left an environment variable `PW_KEYBOARD_LAYOUT` but I think it probably should be a `browserContext` option
- better testing of non-trivial keyboard layouts
- documentation

Refs: #7396


docs/src/api/class-keyboard.md
```diff
@@ -374,3 +374,12 @@ Dispatches a `keyup` event.
 - `key` <[string]>
 
 Name of the key to press or a character to generate, such as `ArrowLeft` or `a`.
+
+## async method: Keyboard.changeLayout
+* since: v1.**
+
+Changes keyboard layout.
```
mxschmitt, line 381: We should document here a list about all our supported layout names.

docs/src/api/params.md
```diff
@@ -734,6 +734,10 @@ Whether to allow sites to register Service workers. Defaults to `'allow'`.
 * `'allow'`: [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) can be registered.
 * `'block'`: Playwright will block all registration of Service Workers.
 
+## context-option-keyboard-layout
+- `keyboardLayout` <[string]>
+
+Keyboard layout.
```
mxschmitt, line 740: ditto

tests/android/browser.spec.ts
```diff
@@ -117,3 +117,14 @@ test('should be able to pass context options', async ({ androidDevice, httpsServ
...
+
+test('should set keyboard layout on android', async function({ androidDevice, server }) {
+  const context = await androidDevice.launchBrowser({ keyboardLayout: 'el-GR' });
+  const [page] = context.pages();
+  await page.goto(server.PREFIX + '/input/keyboard.html');
+  await page.keyboard.press('α');
+  expect(await page.evaluate('getResult()')).toBe(
+      ['Keydown: α KeyA 65 []',
+        'Keypress: α KeyA 945 945 []',
+        'Keyup: α KeyA 65 []'].join('\n'));
```
mxschmitt, line 129:
```suggestion
        'Keyup: α KeyA 65 []'].join('\n'));
    await context.close();
```

tests/electron/electron-app.spec.ts
```diff
@@ -221,3 +221,14 @@ test('should return app name / version from manifest', async ({ launchElectronAp
...
+test('should set keyboard layout on electron app', async ({ launchElectronApp, server }) => {
+  const electronApp = await launchElectronApp('electron-window-app.js', [], { keyboardLayout: 'el-GR' });
+  const page = await electronApp.firstWindow();
+  await page.goto(server.PREFIX + '/input/keyboard.html');
+  await page.keyboard.press('α');
+  expect(await page.evaluate('getResult()')).toBe(
+      ['Keydown: α KeyA 65 []',
+        'Keypress: α KeyA 945 945 []',
+        'Keyup: α KeyA 65 []'].join('\n'));
+});
```
mxschmitt, line 234:
```suggestion
  await electronApp.close();
});
```

tests/page/page-keyboard-layouts.spec.ts
```diff
@@ -0,0 +1,363 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
```
mxschmitt, line 2: this should be just Microsoft copyright. E.g. from [here](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/fetch.ts#L2).

tests/page/page-keyboard-layouts.spec.ts
```diff
@@ -0,0 +1,363 @@
...
+    await page.keyboard.press('Backslash');
+    expect(await page.evaluate('getResult()')).toBe(
+        ['Keydown: \\ Backslash 220 []',
+          'Keypress: \\ Backslash 92 92 []',
+          'Keyup: \\ Backslash 220 []'].join('\n'));
+  });
+});
+
+it(`should fallback to us on invalid layout format`, async ({ page, toImpl, server }) => {
+  await page.keyboard.changeLayout('invalid');
```
mxschmitt, line 103: I'd throw instead so users are not confused.

tests/page/page-keyboard-layouts.spec.ts
```diff
@@ -0,0 +1,363 @@
...
+  'tk_tm': { keyCode: 65, key: 'a' },
+  'uk_ua': { keyCode: 65, key: 'ф' },
+  'ug_cn': { keyCode: 65, key: 'ھ' },
+  'uz_cyrl_uz': { keyCode: 65, key: 'ф' },
+  'cy_gb': { keyCode: 65, key: 'a' },
+  'yo_ng': { keyCode: 65, key: 'a' },
+};
+
+for (const [locale, { key, keyCode }] of Object.entries(testData)) {
+  it(`should fire events on KeyA for ${locale} locale`, async ({ page, server, toImpl }) => {
```
mxschmitt, line 352:
```suggestion
  it(`should fire events on KeyA for ${locale} locale`, async ({ page, server }) => {
```

tests/library/browsercontext-keyboard-layout.spec.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
```
mxschmitt, line 2: license ditto

tests/library/browsercontext-keyboard-layout.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+it('should set keyboard layout on Browser.newContext @smoke', async ({ browser, server }) => {
+  const context = await browser.newContext({ keyboardLayout: 'el-GR' });
+  const page = await context.newPage();
+  await page.goto(server.PREFIX + '/input/keyboard.html');
+  await page.keyboard.press('α');
+  expect(await page.evaluate('getResult()')).toBe(
+      ['Keydown: α KeyA 65 []',
+        'Keypress: α KeyA 945 945 []',
+        'Keyup: α KeyA 65 []'].join('\n'));
+});
```
mxschmitt, line 29:
```suggestion
  await context.close();
});
```

tests/library/browsercontext-keyboard-layout.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+it('should set keyboard layout on Browser.newPage @smoke', async ({ browser, server }) => {
+  const page = await browser.newPage({ keyboardLayout: 'el-GR' });
+  await page.goto(server.PREFIX + '/input/keyboard.html');
+  await page.keyboard.press('α');
+  expect(await page.evaluate('getResult()')).toBe(
+      ['Keydown: α KeyA 65 []',
+        'Keypress: α KeyA 945 945 []',
+        'Keyup: α KeyA 65 []'].join('\n'));
+});
```
mxschmitt, line 39:
```suggestion
  await page.close();
});
```

tests/library/browsercontext-keyboard-layout.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  const userDataDir = await createUserDataDir();
+  const context = await browserType.launchPersistentContext(userDataDir, { keyboardLayout: 'el-GR' });
+  const page = await context.newPage();
+  await page.goto(server.PREFIX + '/input/keyboard.html');
+  await page.keyboard.press('α');
+  expect(await page.evaluate('getResult()')).toBe(
+      ['Keydown: α KeyA 65 []',
+        'Keypress: α KeyA 945 945 []',
+        'Keyup: α KeyA 65 []'].join('\n'));
+});
```
mxschmitt, line 51:
```suggestion
  await context.close();
});
```

packages/playwright-core/src/server/keyboards/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  key?: string;
+  keyCode?: number;
+  keyCodeWithoutLocation?: number;
+  shiftKey?: string;
+  shiftKeyCode?: number;
+  text?: string;
+  location?: number;
+};
+
+export type KeyboardLayout = { [s: string]: KeyDefinition; };
```
mxschmitt, line 27:
```suggestion
export type KeyboardLayout = Record<string, KeyDefinition>;
```

utils/generate_keyboard_layouts.js
```diff
@@ -0,0 +1,458 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @ts-check
+
+/**
+ * @typedef { import("../packages/playwright-core/src/server/keyboards/types").KeyDefinition } KeyDefinition
```
mxschmitt, line 21: We prefer single quotes, please update the whole files to single quotes only.

utils/generate_keyboard_layouts.js
```diff
@@ -0,0 +1,458 @@
...
+  "00000442": { layoutName: "Turkmen keyboard", locales: ["tk-TM"] },
+  "00020422": { layoutName: "Ukrainian (Enhanced) keyboard", locales: ["uk-UA"] },
+  "00010480": { layoutName: "Uyghur keyboard", locales: ["ug-CN"] },
+  "00000843": { layoutName: "Uzbek Cyrillic keyboard", locales: ["uz-Cyrl-UZ"] },
+  "00000452": { layoutName: "United Kingdom Extended keyboard", locales: ["cy-GB"] },
+  "0000046A": { layoutName: "Yoruba keyboard", locales: ["yo-NG"] },
+};
+
+const copyrightHeader = `
+/**
```
mxschmitt, line 156:
```suggestion
const copyrightHeader = `/**
```

utils/generate_keyboard_layouts.js
```diff
@@ -0,0 +1,458 @@
...
+ * Copyright 2017 Google Inc. All rights reserved.
+ * Modifications copyright (c) Microsoft Corporation.
```
mxschmitt, line 158: This should be the normal copyright.

utils/generate_keyboard_layouts.js
```diff
@@ -0,0 +1,458 @@
...
+  return mappings;
+}
+
+/**
+ * @param { Page } page
+ * @param { string } url
+ * @returns { Promise<Object.<string, KeyDefinition>> }
+ */
+async function generate(page, url) {
+  const locatorToVirtualKeys = await extractLocatorToVirtualKeys(page, url);
```
mxschmitt, line 364: Instead of using Playwright to scrape the keyboard layouts it would be better to find their source or find out how they generate it. So we don't depend on selectors which might change over time.

utils/generate_keyboard_layouts.js
```diff
@@ -0,0 +1,458 @@
...
+    `export const localeMapping = new Map<string, string>([`,
+    ...mapEntries,
+    `]);`,
+    ``,
+    `export const keypadLocation = 3;`,
+  ].join('\n');
+
+  fs.writeFileSync(path.resolve(keyboardsDir, 'index.ts'), index, 'utf-8');
+
+  browser.close();
```
mxschmitt, line 457:
```suggestion
  await browser.close();
```

packages/playwright-core/src/server/keyboards/index.ts
```diff
@@ -0,0 +1,267 @@
+
```
mxschmitt, line 1: remove new line + update copyright

mxschmitt (dismissed):
Hey! Thank you for your great contribution!

We discussed it in our team meeting and would like to just go with the 5 most common keyboard layouts instead of all of them for the beginning. We can follow up for more (bundle size is our reasoning for now). Also it would be cool if its possible to find another way of getting the data for our generation. Since scraping is not ideal.

utils/generate_keyboard_layouts.js
ruifigueira: I understand. I did it that way because it was easier for a poc, and not knowing at the time too much about keyboard layouts, it was the fastest way to have some keyboard layouts, but I agree that's not ideal.

I was already looking into the XML files they reference to try to figure the mapping, because actually we only need `key`, `shiftKey` and `virtualKeyCode` (all other fields, like `position` are constant, so we can have those hardcoded). 

Honestly, for `key` and `shiftKey` we should rely on CLDR:

https://github.com/unicode-org/cldr/blob/main/keyboards/windows/en-t-k0-windows.xml

However, I didn't find how to get the virtual key code from there, but I'll investigate that a bit more

tests/page/page-keyboard-layouts.spec.ts
ruifigueira: I thought it also, I only decided for the fallback instead of throwing an exception because there's no standard layout IDs. But since we're going to explicitly document all supported layouts, that makes sense.

Another thing: I don't particularly like the `keyboard.changeLayout` API, it's a destructive call, and I was thinking that maybe I could just create new keyboard instances per layout. Something like:

```js
const ptKeyboard = page.keyboardFor('pt-PT');
await ptKeyboard.press('Tab');
```

utils/generate_keyboard_layouts.js
ruifigueira: Just an update: I couldn't find how to get the virtual key code from CLDR, so I'm using the XML files provided at kbdlayout.info. For instance, for portuguese keyboard:

https://kbdlayout.info/kbdpo/download/kbdtables
https://kbdlayout.info/kbdpo/download/xml

I also found out that I was handling ligature characters properly. Nevertheless, current `input.ts` doesn't handle them properly neither, so I suggest that we don't include keyboard layouts with ligatures for now, like arabic ones. The ones containing ligatures have `LIGATURE` XML elements:

https://kbdlayout.info/kbda1/download/kbdtables

ruifigueira: addressed in a2d1e0bb5d0a752a2a61866f35c233f0391dbc48

tests/page/page-keyboard-layouts.spec.ts
ruifigueira: I did a branch to test the alternative API proposal, you can find it here: https://github.com/microsoft/playwright/commit/eea892fd6368295d1e1cf178a87d6db10418181f

ruifigueira:
@mxschmitt do you already know which keyboard layouts you want to start with, so that I can adjust everything? (tests. documentation, etc.)

mxschmitt:
> @mxschmitt do you already know which keyboard layouts you want to start with, so that I can adjust everything? (tests. documentation, etc.)

This would be our preferred list:

```
US English
Arabic
British
Danish
French
German
Italian
Portuguese
Russian
Ukrainian
Spanish & Catalan (Spain)
Spanish (Latin America)
Swiss
```

thanks!

ruifigueira:
Arabic is tricky, because it contains ligatures (see https://github.com/microsoft/playwright/pull/24249#discussion_r1267773057). The generator extracts them, but as they are not a multi-unicode characters, they are not yet handled properly by current playwright implementation.

I did some tests and it doesn't seem to have a trivial solution, so I suggest we open an issue for ligature characters and address arabic keyboard layouts after that.

For reference, here are the actual key events sent when pressing `Shift+KeyT` with an arabic layout at https://w3c.github.io/uievents/tools/key-event-viewer.html (notice the two keypress events, one for each unicode character of the ligature):

image

mxschmitt:
Great investigation! Actually I added it by mistake, since as you said blocked by https://github.com/microsoft/playwright/issues/5777. Feel free to leave it out.

ruifigueira:
Here is the list, with their codes (I'm using both `locale` and [linux layout codes](https://www.apt-browse.org/browse/debian/wheezy/main/all/xkb-data/2.5.1-3/file/usr/share/X11/xkb/rules/evdev.lst)).

| Values | Name |
| :- | :- |
| `us`, `en-US` | [US English](https://learn.microsoft.com/en-us/globalization/keyboards/kbdus_7)  |
| `gb`, `en-GB` | [British](https://learn.microsoft.com/en-us/globalization/keyboards/kbduk)  |
| `dk`, `da-DK` | [Danish](https://learn.microsoft.com/en-us/globalization/keyboards/kbdda)  |
| `fr`, `fr-FR` | [French](https://learn.microsoft.com/en-us/globalization/keyboards/kbdfr)  |
| `de`, `de-DE` | [German](https://learn.microsoft.com/en-us/globalization/keyboards/kbdgr)  |
| `it`, `it-IT` | [Italian](https://learn.microsoft.com/en-us/globalization/keyboards/kbdit)  |
| `pt`, `pt-PT` | [Portuguese (Portugal)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdpo)  |
| `br`, `pt-BR` | [Portuguese (Brazil)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdpo)  |
| `ru`, `ru-RU` | [Russian](https://learn.microsoft.com/en-us/globalization/keyboards/kbdru)  |
| `ua`, `uk-UA` | [Ukrainian](https://learn.microsoft.com/en-us/globalization/keyboards/kbdur1)  |
| `es`, `es-ES`, `ca-ES` | [Spanish & Catalan (Spain)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdsp)  |
| `latam` | [Spanish (Latin America)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdla)  |
| `ch`, `de-CH` | [German (Switzerland)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdsg)  |
| `fr-CH`, `it-CH` | [French and Italian (Switzerland)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdsf_2)  |

ruifigueira:
I'm doing some deadkey tests. There are some discrepancies among browsers: for instance, with Portuguese keyboard layout and Windows 11, key `BracketRight`, which is a deadkey, has a `keyCode=186` in both chromium and webkit, but has `keyCode=192` in firefox.

**chromium**

image

**firefox**

image

Nevertheless, `keyCode` is a [deprecated property](
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), so maybe we can leave it this way?

ruifigueira:
I already generated the first set of keyboard layouts. As I said, dead keys were not yet handled, so I had to change `input.ts` to handle them.

I tried to test all cases with dead keys on all keyboard layouts, I think that it has a good coverage.

ruifigueira:
@mxschmitt, did you have a chance to review mylatest commits and comments?

dgozman:
@ruifigueira Thank you for the PR! We are sorry that reviewing this takes a long time. You can be sure this is on our radar on a daily basis. I hope we'll get back to you tomorrow.

ruifigueira:
Actually, I don't have a use case specific for dead keys, but as they are part of some keyboard layouts and they were not handled, it felt incomplete. That is, pressing `BracketRight` in a portuguese layout should generate a `Dead` key event:

image

I also started checking IME support (see https://github.com/microsoft/playwright/issues/5777#issuecomment-1651784399), but my understanding is that it is more related with [CompositionEvents](https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent). In my case, I'm only trying to mimick events fired when I press my physical keyboard using those different layouts on [Keyboard Event Viewer](https://w3c.github.io/uievents/tools/key-event-viewer.html), for example.

dgozman:
@ruifigueira

> Actually, I don't have a specific use case for dead keys, but as they are part of some keyboard layouts and they were not handled, it felt incomplete.

Interesting. How would you expect to type `é` or any accented letter in this case? Do you need to? If not, why is keyboard layout important to you?
Also, it is still not clear why would you need to `type` instead of `fill`? Could you please explain in more details?

ruifigueira:
> How would you expect to type é or any accented letter in this case?

I added tests for those cases of accented letters, and it basically presses de dead key for the accent followed by the `KeyE` key. And it is generating realistic events.

> Also, it is still not clear why would you need to type instead of fill

For instance, using `type` for testing a typeahead search component with accented keys will be more realistic in terms of generated events.

ruifigueira:
Just to complement a bit more my previous responses, here are the outputs typing the word `não`:

## Using `fill`

```js
test(`fill`, async ({ page }) => {
  await page.goto('https://w3c.github.io/uievents/tools/key-event-viewer.html');
  await page.locator('#input').fill('não');
  await page.locator('#output').screenshot({ path: 'fill.png' });
});
```

**Result:**

fill

## Using `type`, US keyboard

```js
test('type with us keyboard', async ({ page }) => {
  await page.keyboard.changeLayout('us');
  await page.goto('https://w3c.github.io/uievents/tools/key-event-viewer.html');
  await page.locator('#input').click();
  await page.keyboard.type('não');
  await page.locator('#output').screenshot({ path: 'type_us.png' });
});
```

**Result:**

type_us

## Using `type`, portuguese keyboard

```js
test('type with portuguese keyboard', async ({ page }) => {
  await page.keyboard.changeLayout('pt');
  await page.goto('https://w3c.github.io/uievents/tools/key-event-viewer.html');
  await page.locator('#input').click();
  await page.keyboard.type('não');
  await page.locator('#output').screenshot({ path: 'type_pt.png' });
});
```

**Result:**

type_pt

## Manually typing, portuguese keyboard

**Result:**

type_pt_manual

This exactly matches the playwright test using `type` on portuguese keyboard

ruifigueira:
@dgozman I got a chance to do a few more tests, and dead keys are more complex than I thought. For instance, pressing a dead key followed by a key without accents will generate some odd events.

For example, with a portuguese layout, typing `BracketRight` followed by `KeyT` should generate `´t`  with the following events:

image
Notice that there's two `(keypress, beforeinput, input)` blocks while `KeyT` was pressed down and released.

The only way I found to replicate that behaviour is to use CDP directly:

```js
test('dead key followed by a key without accent', async ({ page }) => {
  await page.goto('https://w3c.github.io/uievents/tools/key-event-viewer.html');
  const client = await page.context().newCDPSession(page);
  await client.send('Input.dispatchKeyEvent', { type: 'rawKeyDown', key: 'Dead', code: 'BracketRight', windowsVirtualKeyCode: 186 });
  await client.send('Input.dispatchKeyEvent', { type: 'keyUp', key: 'Dead', code: 'BracketRight', windowsVirtualKeyCode: 186 });
  await client.send('Input.dispatchKeyEvent', { type: 'keyDown', key: 't', code: 'KeyT', text: '´', windowsVirtualKeyCode: 84 });
  await client.send('Input.dispatchKeyEvent', { type: 'char', key: 't', code: 'KeyT', text: 't', windowsVirtualKeyCode: 84 });
  await client.send('Input.dispatchKeyEvent', { type: 'keyUp', key: 't', code: 'KeyT', windowsVirtualKeyCode: 84 });
});
```

That being said, do you have any suggestion on how to procede? I don't mind to try to add better support for dead keys, but I understand your concerns and maybe we should address that seperately.

The following keyboards layouts have dead keys:

| Values | Name | Dead Keys |
| :- | :- | :- |
| `us`, `en-US` | [US English](https://learn.microsoft.com/en-us/globalization/keyboards/kbdus_7)  |  |
| `gb`, `en-GB` | [British](https://learn.microsoft.com/en-us/globalization/keyboards/kbduk)  |  |
| `dk`, `da-DK` | [Danish](https://learn.microsoft.com/en-us/globalization/keyboards/kbdda)  | :heavy_check_mark: |
| `fr`, `fr-FR` | [French](https://learn.microsoft.com/en-us/globalization/keyboards/kbdfr)  | :heavy_check_mark: |
| `de`, `de-DE` | [German](https://learn.microsoft.com/en-us/globalization/keyboards/kbdgr)  | :heavy_check_mark: |
| `it`, `it-IT` | [Italian](https://learn.microsoft.com/en-us/globalization/keyboards/kbdit)  |  |
| `pt`, `pt-PT` | [Portuguese (Portugal)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdpo)  | :heavy_check_mark: |
| `br`, `pt-BR` | [Portuguese (Brazil)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdpo)  | :heavy_check_mark: |
| `ru`, `ru-RU` | [Russian](https://learn.microsoft.com/en-us/globalization/keyboards/kbdru)  |  |
| `ua`, `uk-UA` | [Ukrainian](https://learn.microsoft.com/en-us/globalization/keyboards/kbdur1)  |  |
| `es`, `es-ES`, `ca-ES` | [Spanish & Catalan (Spain)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdsp)  | :heavy_check_mark: |
| `latam` | [Spanish (Latin America)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdla)  | :heavy_check_mark: |
| `ch`, `de-CH` | [German (Switzerland)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdsg)  | :heavy_check_mark: |
| `fr-CH`, `it-CH` | [French and Italian (Switzerland)](https://learn.microsoft.com/en-us/globalization/keyboards/kbdsf_2)  | :heavy_check_mark: |

ruifigueira:
More news: I just added 5345b87fa8946169f52e88ada6cd40c194eedcf9 that provides support for multichar text input via keyDown for both firefox and chromium. I'm almost certain that this will also allow ligatures (see https://github.com/microsoft/playwright/pull/24249#issuecomment-1648593806).

In chromium, I make use of `char` type in `Input.dispatchKeyEvent` command, and in firefox I found out that `key` can be a multicharacter string and it will generate all `keyPress` / `beforeInput` / `input` events (see https://github.com/mozilla/gecko-dev/blob/f1f50693655c093d974f026bd37860d939cd5529/dom/base/test/chrome/window_nsITextInputProcessor.xhtml#L2615-L2638).

Only webkit is not supported yet because playwright patch doesn't handle `RawKeyDown` and `Char` event types, which exist in webkit source code:

https://github.com/WebKit/WebKit/blob/9b47de8376bfeceb41323030c879fd3c7e5d64b3/Source/WebKit/Shared/WebEventType.h#L45-L48

I found out that compiling webkit with those patches is out of scope of this repository, but this seems to be a fairly simple change in `browser_patches/webkit/patches/bootstrap.diff`.

ruifigueira:
I rebased with `main` branch, and also added arabic with ligature support (except for webkit) as explained in my previous comment. This will be easy to remove if we decide not to follow this path, but it's to demostrate that both ligatures and deadkeys can be implemented with just a few code changes (for webkit, it will require a patch nevertheless).

ruifigueira:
Hi @dgozman and @mxschmitt , did you get a chance to look into my comments and recents commits?

dgozman:
@ruifigueira I was on vacation last week. I will get to this PR after getting back up to speed.

dgozman:
@ruifigueira Thank you for your patience, I am back to this PR.

Interestingly, when I use `pt` layout on my Mac, the situation is a bit different. I guess, Mac tries to composite right away and shows pre-composited "`" upon hitting a dead key.

n-dead-a-o-pt

dead-t-pt

Given the above, I think we'll be fine either way, unless we do proper IME composition that is not planned yet.

---  

> More news: I just added https://github.com/microsoft/playwright/commit/5345b87fa8946169f52e88ada6cd40c194eedcf9 that provides support for multichar text input via keyDown for both firefox and chromium. This way I was able to implement the scenarios above of a deadkey followed by a key without accent.

I am not sure we should take this approach without IME support. I think for this PR we should just ignore the extra dead key symbol ("\`" for `pt` layout) and proceed with a single `t` symbol and call this a known limitation. It can only arise if you manually `press('Dead')` and then `press('t')`, which I do not expect anyone to do. Then we can follow up with using `insertText` for inserting the dead key symbol, see how much complexity that adds and decide whether it's worth it.

---

Overall, this looks very good! We should proceed with the taken approach. See my comments for some details that I think we should figure out, and it should be good to go. Again, thank you for spending your time to contribute this!

ruifigueira:
Thanks @dgozman. I just rebased my branch,removed those extra commits trying to fix the unmapped deadkey, and adapted the code not to fail in that scenario. Now, pressing `BracketRight` followed by `KeyT` will just type  `t` and ignore the deadkey, as you proposed.

dgozman (commented):
Sorry, I forgot to publish review comments!

packages/playwright-core/src/server/input.ts
```diff
@@ -44,40 +50,58 @@ export class Keyboard {
...
+    this._keyboardLayout = getByKeyboardLayoutName(page._browserContext._options.keyboardLayout);
+  }
+
+  changeLayout(layoutName: string) {
+    this._keyboardLayout = getByKeyboardLayoutName(layoutName);
   }
 
   async down(key: string) {
     const description = this._keyDescriptionForString(key);
+    if (description.key !== 'Shift') this._deadKeyMappings = undefined;
```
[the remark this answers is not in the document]
ruifigueira: That logic handles uppercase accented keys. For instance, in `pt` layout, to get `Á`, we need to press `BracketRight` (which sets `_deadKeyMappings`), then press down `Shift` followed by `KetA`. If we reset `_deadKeyMappings` after pressing down the `Shift` key, we would not be able to access the proper accent.

packages/playwright-core/src/server/input.ts
```diff
@@ -286,9 +346,24 @@ function buildLayoutClosure(layout: keyboardLayout.KeyboardLayout): Map<string,
     if (description.key.length === 1)
       result.set(description.key, description);
 
-    // Map from shiftKey, no shifted
-    if (shiftedDescription)
+    // Map from accented keys
+    if (definition.deadKeyMappings) {
+      for (const [k, v] of Object.entries(definition.deadKeyMappings))
+        // if there's a dedicated accented key, we don't want to replace them
+        if (!result.has(v)) result.set(v, [code, k]);
```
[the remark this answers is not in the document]
ruifigueira: You're right, I'll keep a temporary structure to put all accented candidates, and I'll add those entries to the result in the second pass.

packages/playwright-core/src/server/input.ts
```diff
@@ -286,9 +346,24 @@ function buildLayoutClosure(layout: keyboardLayout.KeyboardLayout): Map<string,
...
+    }
+
+    if (shiftedDescription) {
+      // Map from shiftKey, no shifted
       result.set(shiftedDescription.key, { ...shiftedDescription, shifted: undefined });
+
+      // Map from shifted accented keys
+      if (definition.shiftDeadKeyMappings) {
```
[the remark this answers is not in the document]
ruifigueira: I aggree, and I changed it accordingly.

packages/playwright-core/src/server/input.ts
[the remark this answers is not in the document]
ruifigueira: I also tried in windows and linux, but it's not very consistent. For instance, arrow keys in windows don't reset deadkeys, but in linux they do. But I'll include that set.

packages/playwright-core/src/server/input.ts
```diff
@@ -44,40 +50,66 @@ export class Keyboard {
...
+
     const shift = this._pressedModifiers.has('Shift');
     description = shift && description.shifted ? description.shifted : description;
 
     // if any modifiers besides shift are pressed, no text should be sent
     if (this._pressedModifiers.size > 1 || (!this._pressedModifiers.has('Shift') && this._pressedModifiers.size === 1))
       return { ...description, text: '' };
-    return description;
+
+    if (!this._deadKeyMappings || description.key === 'Shift') return description;
```
dgozman, line 93: Why special-casing `Shift` here? I guess this should be the same list as in `down()` method?

packages/playwright-core/src/server/input.ts
```diff
@@ -44,40 +50,66 @@ export class Keyboard {
...
 
   async up(key: string) {
     const description = this._keyDescriptionForString(key);
     if (kModifiers.includes(description.key as types.KeyboardModifier))
       this._pressedModifiers.delete(description.key as types.KeyboardModifier);
     this._pressedKeys.delete(description.code);
-    await this._raw.keyup(this._pressedModifiers, description.code, description.keyCode, description.keyCodeWithoutLocation, description.key, description.location);
+    const descKey = description.deadKeyMappings ? 'Dead' : description.key;
+    await this._raw.keyup(this._pressedModifiers, description.code, description.keyCode, description.keyCodeWithoutLocation, descKey, description.location);
+    if (description.key !== 'Shift') this._deadKeyMappings = description.deadKeyMappings;
```
dgozman, line 112: Same thing here, I guess.

packages/playwright-core/src/server/input.ts
```diff
@@ -44,40 +50,66 @@ export class Keyboard {
...
+    return { ...description, text: deadKeyText, key: deadKeyText };
   }
 
   async up(key: string) {
     const description = this._keyDescriptionForString(key);
     if (kModifiers.includes(description.key as types.KeyboardModifier))
       this._pressedModifiers.delete(description.key as types.KeyboardModifier);
     this._pressedKeys.delete(description.code);
-    await this._raw.keyup(this._pressedModifiers, description.code, description.keyCode, description.keyCodeWithoutLocation, description.key, description.location);
+    const descKey = description.deadKeyMappings ? 'Dead' : description.key;
```
dgozman, line 110: With the new layouts, dead keys already have `key === 'Dead'`, don't they?

packages/playwright-core/src/server/input.ts
```diff
@@ -241,10 +279,38 @@ const aliases = new Map<string, string[]>([
...
+
+  const normalizedLayoutName = normalizeLayoutName(layoutName);
+  const klid = keyboardLayoutNamesMapping.get(normalizedLayoutName);
+  if (!klid) throw new Error(`Keyboard layout name "${layoutName}" not found`);
+
+  const cached = cache.get(klid);
+  if (cached) return cached;
+
+  const layout: KeyboardLayout = require(`./keyboards/layouts/${klid}`).default;
+  assert(layout, `No layout found for klid ${klid}`);
```
dgozman, line 299: This should never happen, so I'd remove the `assert`. If anything, there would be an NPE down the road.

dgozman (commented):
This looks great, just a few more comments.

packages/playwright-core/src/server/input.ts
ruifigueira: That's correct.

packages/playwright-core/src/server/input.ts
ruifigueira: I'm fixing that, and `_deadKeyMappings` will be updated in `keyDown` only. This way it will handle cases like pressing down the deadkey while pressing the key to accent, e.g. `BracketRight+KeyA`. I'm also including a test for that.

docs/src/api/class-keyboard.md
```diff
@@ -374,3 +374,14 @@ Dispatches a `keyup` event.
 - `key` <[string]>
 
 Name of the key to press or a character to generate, such as `ArrowLeft` or `a`.
+
+## async method: Keyboard.changeLayout
+* since: v1.**
```
mxschmitt, line 379: I think we should change them to 1.38 since this PR is now quite ready.

tests/page/page-keyboard-layouts.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+type AccentedKeyTest = [...SimpleKeyTest, string, number];
+
+const testData: Record<string, (SimpleKeyTest | AccentedKeyTest)[]> = {
+  'us': [['`', 'Backquote', 192], ['~', 'Shift+Backquote', 192]],
+  'gb': [['`', 'Backquote', 223], ['¬', 'Shift+Backquote', 223]],
+  'dk': [['ø', 'Quote', 222], ['ä', 'BracketRight', 186, 'KeyA', 65], ['â', 'Shift+BracketRight', 186, 'KeyA', 65], ['Â', 'Shift+BracketRight', 186, 'Shift+KeyA', 65]],
+  'fr': [['q', 'KeyA', 81], ['â', 'BracketLeft', 221, 'KeyQ', 65], ['ä', 'Shift+BracketLeft', 221, 'KeyQ', 65], ['Ä', 'Shift+BracketLeft', 221, 'Shift+KeyQ', 65]],
+  'de': [['#', 'Backslash', 191], ['á', 'Equal', 221, 'KeyA', 65], ['à', 'Shift+Equal', 221, 'KeyA', 65], ['À', 'Shift+Equal', 221, 'Shift+KeyA', 65]],
+  'it': [['è', 'BracketLeft', 186], ['é', 'Shift+BracketLeft', 186]],
+  'pt': [['\\', 'Backquote', 220], ['á', 'BracketRight', 186, 'KeyA', 65], ['à', 'Shift+BracketRight', 186, 'KeyA', 65], ['À', 'Shift+BracketRight', 186, 'Shift+KeyA', 65]],
```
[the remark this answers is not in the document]
ruifigueira: Added:

https://github.com/microsoft/playwright/blob/ee6134cbd0c25d213869735f064937daefc26188/tests/page/page-keyboard-layouts.spec.ts#L39-L46

https://github.com/microsoft/playwright/blob/ee6134cbd0c25d213869735f064937daefc26188/tests/page/page-keyboard-layouts.spec.ts#L48-L57

ruifigueira:
@dgozman @mxschmitt did you had a chance to review this? I noticed that 1.38 was released already, so at least I'll have to change versions in documentation.

dgozman:
@ruifigueira This looks good. I was waiting for v1.38 to ship, so this can go to v1.39. Please update `since` fields and I'll merge this in. Thank you!

ruifigueira:
Done, updated to 1.39.

ruifigueira:
@dgozman any updates?

dgozman:
**Short summary**

We've discussed this feature once more, and we'd prefer to postpone it for now, until we gather more feedback on `type()` vs `fill()`. @ruifigueira, I am sorry for being indecisive, and very much thankful for all the work you've put in!

**More details**

Your work has prompted us to make changes to the `type()` vs `fill()` methods in v1.38, and stronger encourage everyone to use `fill()` whenever possible. That's what we recommended from the start, but we never articulated it clearly enough. Now, with the `pressSequentially()` instead of `type()`, we expect much more usage of `fill()` instead.

Given that keyboard layouts feature is a smaller subset of `type()` usage, that we already think should be small enough, we'd like a strong signal before committing to maintaining the new API. This does not mean that your PR or proposed API are in some way flawed. In general, the more niche and complicated the API is, the less inclined we are to introduce it.

**Next steps**

With the next 1.39 release, I'll ask for more feedback on the usecases for `type/pressSequentially` usage and keyboard layouts support among our users. Then we'll re-asses whether the feature would be worth releasing. For now, I'd recommend to park this PR so that we can revisit it later. Again, thank you for the work you've done, and sorry for all the time you have put into the feature that we are not releasing yet.

ruifigueira:
@dgozman I noticed that API change and I actually thought what would be the impact in this PR, but I wasn't expecting this, after all the work I had 😔.

I understand your explanation, but that still leaves a void on trusted key events for any character not available in US keyboard layout, similar to [IME](https://github.com/microsoft/playwright/issues/5777) events, but with the difference that playwright internally already has a way to have full keyboard events support.

Currently, we can do it low level by creating a CDP session and use the [Input.dispatchKeyEvent](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchKeyEvent) method, but that only works on chromium. Both webkit and firefox sessions are not available in client API (and even if they were, their protocols appear to be internal), and so there's no chance to fire trusted events there for non-US characters there.

SI honestly don't have an alternative proposal, but playwright is able to do so many incredible things like intercepting requests and change them on the fly, that it just feels odd that it cannot fire some basic trusted events for non-US keys.

dgozman:
> That been said, it would be great if there was a way to generate custom key / composition events

I think discussion about composition events should be in #5777.

Issue #23907, opened 2023-06-26
[Bug] UI Mode via browser does not run worker teardown on CTRL + C


mxschmitt:
See https://github.com/microsoft/playwright/pull/23876#discussion_r1242443903.

marcin-bojar:
Hi @mxschmitt, thanks for creating this issue.

I just realised that I have 247 users left in AWS Cognito due to not running global teardown when closing UI mode (chromium, macOS Ventura). But for me global teardown is also not run when closing the UI mode via context menu (Chromium --> Quit Chromium). 

I'm using `global-setup` file, which returns teardown function. How do I close UI mode, so the global teardown is run? I can't see any way to do that at the moment. Thanks in advance!

PR #24265, opened 2023-07-17
fix(test-runner): don't double emit SIGINT in long running test

mxschmitt:
https://github.com/microsoft/playwright/issues/23907

This fixes the following scenario. You press Command + C during the test-execution, then a global teardown will execute, which takes a long time and before this on macOS did sometimes end up in an unwanted SIGINT received inside the globalTeardown handler. After this change this won't happen anymore.

### DIFF
--- a/packages/playwright-test/src/runner/sigIntWatcher.ts
+++ b/packages/playwright-test/src/runner/sigIntWatcher.ts
@@ -22,25 +22,11 @@ export class SigIntWatcher {
     let sigintCallback: () => void;
     this._sigintPromise = new Promise<void>(f => sigintCallback = f);
     this._sigintHandler = () => {
-      // We remove the handler so that second Ctrl+C immediately kills the runner
-      // via the default sigint handler. This is handy in the case where our shutdown
-      // takes a lot of time or is buggy.
-      //
-      // When running through NPM we might get multiple SIGINT signals
-      // for a single Ctrl+C - this is an NPM bug present since at least NPM v6.
-      // https://github.com/npm/cli/issues/1591
-      // https://github.com/npm/cli/issues/2124
-      //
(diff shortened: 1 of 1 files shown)


packages/playwright-test/src/runner/sigIntWatcher.ts
```diff
@@ -52,6 +38,36 @@ export class SigIntWatcher {
...
+  static on(handler: () => void) {
+    this._handlers.push(handler);
+    if (this._handlers.length === 1)
+      process.on('SIGINT', this._dispatch);
+  }
+
+  static off(handler: () => void) {
+    this._handlers = this._handlers.filter(h => h !== handler);
+    if (!this._handlers.length)
+      process.off('SIGINT', this._dispatch);
```
dgozman, line 71: Should not do this if we had a SIGINT during the last second.

packages/playwright-test/src/runner/sigIntWatcher.ts
```diff
@@ -52,6 +38,56 @@ export class SigIntWatcher {
...
+      // https://github.com/npm/cli/issues/1591
+      // https://github.com/npm/cli/issues/2124
+      // https://github.com/npm/cli/issues/5021
+      //
+      // Therefore, removing the handler too soon will just kill the process
+      // with default handler without printing the results.
+      // We work around this by giving NPM 1000ms to send us duplicate signals.
+      // The side effect is that slow shutdown or bug in our process will force
+      // the user to hit Ctrl+C again after at least a second.
+      setTimeout(() => {
```
dgozman, line 84: I'd rather check `this._handlers.length` at line 55 instead of two timers.

dgozman (approved):

PR #24216, opened 2023-07-13
chore: resolve top-level vs dependency after cli filtering

pavelfeldman:

PR #24268, opened 2023-07-17
chore: apply only filtration to deps

pavelfeldman:


dgozman (approved):

sosullivanIDBS:
I believe this or https://github.com/microsoft/playwright/pull/24216 has caused a bug where `test.only` no longer works if a project has dependencies.

When I use `test.only` and run `npx playwright test`, I get the following error:

`TypeError: Cannot read properties of undefined (reading '_hasTests')`

mxschmitt:
Do you mind filing a bug for this? We usually don't monitor closed issues / PRs, thanks!

PR #24284, opened 2023-07-18
feat(api): expose step stack

### LINKED ISSUE #24283: [Feature] Expose the stacks of test steps
At [Replay](https://www.replay.io/) we want to improve our support for recordings of Playwright tests. In particular we'd like to be able to show a test step's stack and its locations in the test's source, similar to Playwright's trace viewer. To do that we need Playwright to send the test step's stack to our reporter.

hbenl:
Fixes #24283

### DIFF
--- a/packages/playwright-test/src/runner/dispatcher.ts
+++ b/packages/playwright-test/src/runner/dispatcher.ts
@@ -267,6 +267,8 @@ export class Dispatcher {
       }
       const { result, steps } = runData;
       const parentStep = params.parentStepId ? steps.get(params.parentStepId) : undefined;
+      // Sanitize location that comes from user land, it might have extra properties.
+      const location = params.stack?.length && !params.stack[0].file.includes('@playwright') ? { file: params.stack[0].file, line: params.stack[0].line, column: params.stack[0].column } : undefined;
       const step: TestStep = {
         title: params.title,
         titlePath: () => {
@@ -278,7 +280,8 @@ export class Dispatcher {
         startTime: new Date(params.wallTime),
         duration: -1,
         steps: [],
-        location: params.location,
+        location,
+        stack: params.stack,
       };
       steps.set(params.stepId, step);
       (parentStep || result).steps.push(step);
--- a/packages/playwright-test/src/runner/dispatcher.ts
+++ b/packages/playwright-test/src/runner/dispatcher.ts
@@ -281,7 +281,7 @@ export class Dispatcher {
         duration: -1,
         steps: [],
         location,
-        stack: params.stack,
+        stack: params.stack?.length ? params.stack : undefined,
       };
       steps.set(params.stepId, step);
       (parentStep || result).steps.push(step);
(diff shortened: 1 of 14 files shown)


jasonLaster:
Here's some a loom of what we're going for
https://www.loom.com/share/46192b219a9a44418262543f5f6f5733

For context, our current playwright panel is a list of locators which we feel like is difficult to map back to the test itself

packages/playwright-test/src/runner/dispatcher.ts
```diff
@@ -267,6 +267,8 @@ export class Dispatcher {
       }
       const { result, steps } = runData;
       const parentStep = params.parentStepId ? steps.get(params.parentStepId) : undefined;
+      // Sanitize location that comes from user land, it might have extra properties.
```
pavelfeldman, line 270: I can see where you toot it from, but I am not sure it is necessary, should be Ok to pass it in as is.

packages/playwright-test/src/runner/dispatcher.ts
```diff
@@ -278,7 +280,8 @@ export class Dispatcher {
         startTime: new Date(params.wallTime),
         duration: -1,
         steps: [],
-        location: params.location,
+        location,
```
pavelfeldman, line 283: You probably don't need it any longer.

packages/playwright-test/src/worker/testInfo.ts
```diff
@@ -295,20 +295,20 @@ export class TestInfoImpl implements TestInfo {
...
       testId: this._test.id,
       stepId,
       parentStepId: parentStep ? parentStep.stepId : undefined,
       title: data.title,
       category: data.category,
       wallTime: data.wallTime,
-      location,
+      stack: data.stack?.map(frame => ({
+        ...frame,
+        file: path.relative(process.cwd(), frame.file)
```
pavelfeldman, line 307: I think we leave it up to the reporters to perform the cosmetic resolution.

packages/playwright-test/src/worker/testInfo.ts
```diff
@@ -295,20 +295,20 @@ export class TestInfoImpl implements TestInfo {
...
+      })),
     };
     this._onStepBegin(payload);
-    this._traceEvents.push(createBeforeActionTraceEventForStep(stepId, parentStep?.stepId, data.apiName || data.title, data.params, data.wallTime, data.location ? [data.location] : []));
+    this._traceEvents.push(createBeforeActionTraceEventForStep(stepId, parentStep?.stepId, data.apiName || data.title, data.params, data.wallTime, data.stack?.length ? [data.stack[0]] : []));
```
pavelfeldman, line 311: It wants stack, you have stack, pass the stack?

packages/playwright-test/src/runner/dispatcher.ts
hbenl: Do you mean I just don't have to remove the `function` property or do you also mean I don't need to filter out a location whose `file` includes `'@playwright'`?

pavelfeldman:
Closing to move the discussion into the issue.

Issue #9667, opened 2021-10-20
Add ability to use AND and OR expression via grep from command line


humbienri:
To bring Windows usage up to par with usage on Mac OS X or Linux, 
`npx playwright test -g "tag1|tag2"`

should also work from the Windows command line. Please.

---

Edit by maintainers: Here are Git Bash and PowerShell solutions:

**PowerShell:**

```PowerShell
npx playwright test --grep  "`"@test1|@test2"`"
```

**Git Bash**

```bash
 npx playwright test --grep "@test1^|@test2"
```

pavelfeldman:
@humbienri : on Windows do this:

`npx playwright test -g "tag1^|tag2"`

jatin-karla11:
how to use and condition in grep?

-g "xy&yz" is not working

timtait:
> how to use and condition in grep?
> 
> -g "xy&yz" is not working

Grep is just matching the test title, so you could do something like
```-g "xy.+yz"```

btadros:
I also get 
'commission' is not recognized as an internal or external command, operable program or batch file.

my full command on Powershell:
npx playwright test --grep -g "register|commission" --config=playwright.config.js --workers=3 --retries=1

P.S.: tried escaping it, but still didn't help

i-savitski:
In Powershell the [stop-parsing token](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parsing?view=powershell-7.2#the-stop-parsing-token) helps:  
`npx playwright test -g --% "tag1|tag2"`

imartinflores:
Could someone execute this from the package.json script ?

I m trying:
 "test:textbox": "npx playwright test --config=./configs/playwright-web.config.js --project=chromium --grep text1 | text2"
 "test:textbox": "npx playwright test --config=./configs/playwright-web.config.js --project=chromium -g text1|text2"
 
 And also using the stop-parsing token mentioned above. Always getting the same kind of error
 
 "'text2' is not recognized as an internal or external command"

shashankkvs309668:
> p-parsing token mentioned above. Always getting the same kind of erro

Still facing the issue.

GustavoBernardi:
@imartinflores - same here...

"test:form": "npx playwright test --project=chromium -g @payx|@ezlm"

mihalqesko:
[...]

PR #24304, opened 2023-07-19
docs: add grep scripts examples for windows terminals

jaktestowac:
Windows Playwright users struggling to find a good command for grouping tags with OR and AND operations, so some examples in official docs can help with this problem.
Adding pull request as discussed here https://github.com/microsoft/playwright/issues/9667#issuecomment-1641818170


jaktestowac:

mxschmitt (approved):

PR #24313, opened 2023-07-20
test: generate code coverage for ttest

dgozman:


.github/workflows/tests_primary.yml
```diff
@@ -125,6 +125,8 @@ jobs:
         DEBUG: pw:install
     - run: npm run build
     - run: npx playwright install --with-deps
+    - run: set PWTEST_COVERAGE=1
```
mxschmitt, line 128: Should be this instead, since every step has its own terminal session.

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-writing-an-environment-variable-to-github_env

aslushnikov (approved):

PR #24315, opened 2023-07-20
chore: refactor timeout manager to use scopes (1)

pavelfeldman:


packages/playwright-test/src/worker/workerMain.ts
```diff
@@ -403,82 +403,84 @@ export class WorkerMain extends ProcessRunner {
...
-      if (testInfo._isFailure())
-        this._isStopped = true;
-
-      if (this._isStopped) {
-        // Run all remaining "afterAll" hooks and teardown all fixtures when worker is shutting down.
-        // Mark as "cleaned up" early to avoid running cleanup twice.
-        this._didRunFullCleanup = true;
-
-        // Give it more time for the full cleanup.
+    await testInfo._timeoutManager.runRunnable({ type: 'afterEach', slot: afterHooksSlot }, async () => {
```
dgozman, line 407: Pavel says: `runWithTimeout`, then `runRunnable`, then `runAsStep`.

PR #24345, opened 2023-07-21
feat(trace): allow serving HTML report from Trace Viewer

### LINKED ISSUE #24219: [Feature] Make it possible to open playwright-report on trace.playwright.dev

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/24219


packages/html-reporter/src/links.tsx
```diff
@@ -76,7 +76,15 @@ export const AttachmentLink: React.FunctionComponent<{
 };
 
 export function generateTraceUrl(traces: TestAttachment[]) {
-  return `trace/index.html?${traces.map((a, i) => `trace=${new URL(a.path!, window.location.href)}`).join('&')}`;
+  const queryString = new URLSearchParams();
+  for (const trace of traces)
+    queryString.append('trace', new URL(trace.path!, window.location.href).toString());
+
+  // If served from a Trace Viewer, we re-use the trace viewer and not use the bundled one.
```
pavelfeldman, line 83: We don't really want to "open" html report in the trace viewer. I wonder if it is easier to solve the same problem via having a drag'n'drop loader that can redirect to either html report or to the trace viewer app. That way you have a single facade, but the tools still don't know about each other more than they need to know.

mxschmitt:
We decided to not go forward with that. I moved it to the Community org: https://github.com/playwright-community/html.playwright.tech

PR #24352, opened 2023-07-21
feat(browser download): add env option to disable progress bar in situations like CI

### LINKED ISSUE #22074: [Feature] Ability to disable progress output from install
👋 We call `npx playwright install` in CI quite a bit, and it spams the logs quite a bit with progress bars. It'd be nice if there was a `--no-progress` flag or similar like many other dependency installation commands that suppresses the progress bar printing, allowing for cleaner logs. Cheers.

mastrzyz:
## What?

Adds the option to disable the progress bar when downloading browsers:

```
Downloading Chromium 115.0.5790.98 (playwright build v1045) from https://playwright.azureedge.net/builds/chromium/1045/chromium-mac-arm64.zip
123.8 Mb [==                  ] 11% 10.5s^C
```

## Why?

In situations like CI , this can lead to a lot of console spam so its better to have this functionality configurable. Fixes #22074 

The below is actually from this PR's Build! ( Let me know if you want me to also enable this ENV in the YAML :) )
image

## How did I test this?

I leveraged the `roll_browser.js` NodeJS script to verify this functionalilty.

`PLAYWRIGHT_SHOW_DOWNLOAD_PROGRESS=0 node roll_browser.js chromium 1045 build` -> Doesn't show progress bar

`PLAYWRIGHT_SHOW_DOWNLOAD_PROGRESS=1 node roll_browser.js chromium 1045 build` -> Shows progress bar

`PLAYWRIGHT_SHOW_DOWNLOAD_PROGRESS=0 node roll_browser.js chromium 1045 build` -> Default , shows progress bar

### DIFF
--- a/packages/playwright-core/src/server/registry/oopDownloadBrowserMain.ts
+++ b/packages/playwright-core/src/server/registry/oopDownloadBrowserMain.ts
@@ -146,7 +146,7 @@ async function main() {
   const log = (message: string) => process.send?.({ method: 'log', params: { message } });
   const [title, browserDirectory, url, zipPath, executablePath, downloadConnectionTimeout, downloadShowProgress] = process.argv.slice(2);
   await downloadFile(url, zipPath, {
-    progressCallback:downloadShowProgress === "true" ?  getDownloadProgress() : () => {},
+    progressCallback: downloadShowProgress === 'true' ?  getDownloadProgress() : () => {},
     userAgent: getUserAgent(),
     log,
     connectionTimeout: +downloadConnectionTimeout,
--- a/packages/playwright-core/src/server/registry/oopDownloadBrowserMain.ts
+++ b/packages/playwright-core/src/server/registry/oopDownloadBrowserMain.ts
@@ -144,9 +144,9 @@ function toMegabytes(bytes: number) {
 
 async function main() {
   const log = (message: string) => process.send?.({ method: 'log', params: { message } });
-  const [title, browserDirectory, url, zipPath, executablePath, downloadConnectionTimeout] = process.argv.slice(2);
+  const [title, browserDirectory, url, zipPath, executablePath, downloadConnectionTimeout, downloadShowProgress] = process.argv.slice(2);
   await downloadFile(url, zipPath, {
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/server/registry/oopDownloadBrowserMain.ts
```diff
@@ -144,9 +144,9 @@ function toMegabytes(bytes: number) {
 
 async function main() {
   const log = (message: string) => process.send?.({ method: 'log', params: { message } });
-  const [title, browserDirectory, url, zipPath, executablePath, downloadConnectionTimeout] = process.argv.slice(2);
+  const [title, browserDirectory, url, zipPath, executablePath, downloadConnectionTimeout, downloadShowProgress] = process.argv.slice(2);
   await downloadFile(url, zipPath, {
-    progressCallback: getDownloadProgress(),
+    progressCallback: downloadShowProgress === 'true' ?  getDownloadProgress() : () => {},
```
pavelfeldman, line 149: There already is a special case for `process.stdout.isTTY` (line 94 of this file). I think you are seeing the results of the `getBasicDownloadProgress` here, which is as intended. Sometimes downloading will fail and it is important to see when this happens.

mastrzyz: `isTTY` can still spam the console I think right? 

image


Current method shows an error if download happens even with `() => {}`

image


Anyway I might be misunderstanding the original goal so please let me know if you would like me to do this a different way.

pavelfeldman: To make sure we are on the same page, we have two versions of the download progress reporting: `getBasicDownloadProgress` and animated `getAnimatedDownloadProgress`. The idea is that we use animated one whenever we detect TTY so that it was nice and interactive. And for the other cases (CI), we use the basic one.

I read you request as `getBasicDownloadProgress` is still too noisy, I want it to be more concise. Is that correct assessment? If so, the changes should be localized to `getBasicDownloadProgress`.

pavelfeldman:
Closing as stale.

PR #24367, opened 2023-07-24
docs(annotations): grep snippets follow-ups #24304

mxschmitt:
Follow-ups to #24304.

Wasn't rendered otherwise: https://playwright.dev/docs/next/test-annotations

### DIFF
--- a/docs/src/test-annotations-js.md
+++ b/docs/src/test-annotations-js.md
@@ -96,12 +96,14 @@ npx playwright test --grep "@fast|@slow"
 On Windows shells:
 
 - PowerShell
-  ```powershell
+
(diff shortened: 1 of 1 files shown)


docs/src/test-annotations-js.md
```diff
@@ -96,12 +96,14 @@ npx playwright test --grep "@fast|@slow"
 On Windows shells:
 
 - PowerShell
-  ```powershell
+
+  ```bash
```
dgozman, line 100: Some tooling should complain instead of silently not rendering the snippet.

dgozman (approved):

PR #24383, opened 2023-07-24
chore: move fs operation in trace recorder to a separate class

dgozman:


packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -570,18 +528,75 @@ function createAfterActionTraceEvent(metadata: CallMetadata): trace.AfterActionT
...
+    await this._writeChain;
+  }
+
+  exportZip(entries: NameValue[], zipFileName: string, onError: (error: Error) => void) {
+    // Chain the export operation against write operations,
+    // so that files do not change during the export.
+    this._appendOperation(async () => {
+      const zipFile = new yazl.ZipFile();
+      const result = new ManualPromise<Error | void>();
+      (zipFile as any as EventEmitter).on('error', error => result.resolve(error));
```
mxschmitt, line 582:
```suggestion
      (zipFile as any as EventEmitter).on('error', () => result.resolve());
```

I feel like this is not used here anymore.

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -570,18 +528,75 @@ function createAfterActionTraceEvent(metadata: CallMetadata): trace.AfterActionT
...
+      for (const entry of entries)
+        zipFile.addFile(entry.value, entry.name);
+      zipFile.end();
+      zipFile.outputStream
+          .pipe(fs.createWriteStream(zipFileName))
+          .on('close', () => result.resolve())
+          .on('error', error => result.resolve(error));
```
mxschmitt, line 589:
```suggestion
          .on('error', () => result.resolve());
```

ditto

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -570,18 +528,75 @@ function createAfterActionTraceEvent(metadata: CallMetadata): trace.AfterActionT
...
+    // This way tracing never have to write the same resource twice.
+    this._appendOperation(async () => {
+      if (skipIfExists)
+        await fs.promises.writeFile(file, content, { flag: 'wx' }).catch(() => {});
+      else
+        await fs.promises.writeFile(file, content);
+    });
+  }
+
+  appendFile(file: string, text: string, flush?: boolean) {
```
pavelfeldman, line 551: One would think that write, append and flush are serialized.

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -570,18 +529,90 @@ function createAfterActionTraceEvent(metadata: CallMetadata): trace.AfterActionT
...
 
-async function flushTraceFile(file: { file: string, buffer: trace.TraceEvent[] }) {
-  const data = file.buffer.map(e => Buffer.from(JSON.stringify(e) + '\n'));
-  await fs.promises.appendFile(file.file, Buffer.concat(data));
-  file.buffer = [];
+  private _flushFile(file: string) {
+    const buffer = this._buffers.get(file);
+    if (buffer === undefined)
+      return;
+    const text = buffer.join('');
```
pavelfeldman, line 567: nit: stream instead of allocating 2x?

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -570,18 +529,90 @@ function createAfterActionTraceEvent(metadata: CallMetadata): trace.AfterActionT
...
+    });
+  }
+
+  private _appendOperation(cb: () => Promise<unknown>): void {
+    // This method serializes all writes to the trace.
+    this._writeChain = this._writeChain.then(async () => {
+      try {
+        await cb();
+      } catch (e) {
+        this._error = e;
```
pavelfeldman, line 614: Do you want to stop the chain in case of the error? Otherwise all these operations will be happening and you would only capture the last error.

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -570,18 +529,90 @@ function createAfterActionTraceEvent(metadata: CallMetadata): trace.AfterActionT
   };
 }
 
-async function appendEventAndFlushIfNeeded(file: { file: string, buffer: trace.TraceEvent[] }, event: trace.TraceEvent) {
-  file.buffer.push(event);
+class SerializedFS {
+  private _writeChain = Promise.resolve();
+  private _buffers = new Map<string, string[]>();
```
pavelfeldman, line 534: Add a comment // Never access from within appendOperation

pavelfeldman (approved):

PR #24406, opened 2023-07-25
fix(artifacts): only attach screenshot when it succeeds

### LINKED ISSUE #24378: [BUG] Can't record trace in a test that opens a download/pdf file
### System info
- Playwright Version: v1.36.0
- Operating System: Windows 11 (WSL Ubuntu)
- Browser: Chromium
- Other info:

**Config file**

```js
import { defineConfig } from '@playwright/test'
import { devices } from '@playwright/test'

/**
 * See https://playwright.dev/docs/test-configuration.
 */
export default defineConfig({
  testDir: './playwright-e2e',

  /**
   Maximum time one test can run for => 120 seconds.
   This timeout is also the reason, why the playwright generator exits after 120 seconds (120 * 1000).
   You need to temporarily adjust this value if you want to use the generator for a longer time.
   */
  // timeout: 120 * 1000 * 1000,
  timeout: 120 * 1000,

  expect: {
    /**
     * Maximum time expect() should wait for the condition to be met.
     * For example in `await expect(locator).toHaveText();`
     */
    timeout: 100000,
  },

  /* Report the slowest tests (up to `max`) slower than `threshold` (in ms) */
  reportSlowTests: { max: 5, threshold: 30 * 1000 },

  /* Fail the build on CI if you accidentally left test.only in the source code. */
  forbidOnly: !!process.env.CI,

  /* Retry on CI only */

  /*
   * Use only as many workers as there are CPUs, but at most 4.
   *
   * Ambient Gitlab-CI workers in May 2022 crashed consistently with more than 2 worker nodes.
   * On an AMD machine with 6 cores and 12 threads using more than 6 workers slowed the tests significantly
   * Because 4-cores/8-threads is the most common dev machine, using 4 workers at most seems reasonable
   */
  workers: process.env.CI ? 2 : 4,

  /*
   * Run tests inside a Testcase in parallel by default.
   * Use test.describe.configure({ mode: 'serial' }) to avoid that,
   * but at this point in time all our tests were  manually marked as parallel mode anyway.
   */
  fullyParallel: true,

  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
  reporter: process.env.CI
    ? [['junit', { outputFile: 'test-results/junit.xml' }], ['list']]
    : 'list',

  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
  use: {
    /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). Limit set to 30s to stop
     infinite tests */
    actionTimeout: 30 * 1000,

    /* Base URL to use in actions like `await page.goto('/')`. */
    baseURL: 'http://localhost:3000',

    screenshot: process.env.PLAYWRIGHT_SLOW_MO ? 'on' : 'only-on-failure',
    video: process.env.PLAYWRIGHT_SLOW_MO ? 'on' : 'retry-with-video',
    trace: process.env.PLAYWRIGHT_SLOW_MO ? 'on' : 'on-all-retries',
  },

  /* Configure projects for major browsers */
  projects: [
    {
      name: 'empto',

      /*
      Project-specific settings.

      launchOptions: {slowMo: 1000,} controls how fast playwright executes the specified commands in a headed run.
      Uncomment this option, if you want to have time when watching how playwright executes tests
      */
      use: {
        ...devices['Desktop Chrome'],
        viewport: { width: 1440, height: 900 },
        launchOptions: {
          slowMo: parseInt(process.env.PLAYWRIGHT_SLOW_MO || '0'),
        },
      },
    },
  ],

  /* Folder for test artifacts such as screenshots, videos, traces, etc. */
  outputDir: 'test-results/',
})

```

```js
test.describe('Try valid users', () => {
    test.use({
      storageState: `loginContexts/loggedin.json`,
    })
    test(`are the internal downloads visible for loggedin`, async ({
      page,
      headless,
    }) => {
      await page.goto('http://localhost:3000/')
      await page.locator('.dropdown-toggle-icon >> nth=-1').click()
      await page.getByRole('link', { name: 'Interne Downloads' }).click()
      await page.getByRole('link', { name: 'Testseite' })

      if (headless) {
        // PDF Files are downloaded in headless mode
        const downloadPromise = page.waitForEvent('download')
        await page
          .locator('a:text("Testseite"):above(a:text("epd Logo vertical"))')
          .click()
        const download = await downloadPromise
        const filename = download.suggestedFilename()
        await expect(filename).toBe('testseite.pdf')
      } else {
        // PDF files are shown in a new browser tab using the built-in PDF viewer in headed mode
        const popupPromise = page.waitForEvent('popup')
        await page
          .locator('a:text("Testseite"):above(a:text("epd Logo vertical"))')
          .click()
        const popup = await popupPromise
        // Wait for the popup to load.
        await popup.waitForLoadState()
        const url = popup.url()
        await expect(url).toContain('testseite.pdf')
      }
    })
  })
```

**Steps**
- Run the test with traces set to always record (`on`)

**Expected**

- The test should work just fine even if I tell it to record traces on successful runs

**Actual**

- The test fails, complaining it can't find screenshot #2 `Error: ENOENT: no such file or directory, open '/home/benedikt/develop/empto/test-results/general-emptoInternalDownloads-Try-valid-users-6acb2-ads-visible-for-mp-makler-premium-dashboard-epd-empto/test-finished-2.png'`

- If I turn off the trace-recording and only save a video and screenshots the test runs fine

- If I run the test in headed mode trace-recording works just fine as well. This leads me to believe the problem is with taking a screenshot of a Download (headless mode) vs taking a screenshot of the PDF-Viewer Tab that chrome opens in headed mode instead.

I can't share the site the test is running against (proprietary code, not hosted on GitHub) - If absolutely necessary to reproduce, I can try and whip up a minimal sample site.    
The testcase explicitly has separate paths for headless and headed mode because Playwright has no way to turn on/off browser features (here: the PDF-Viewer) in either headless or headed mode. (AFAIK there is no way to enable the PDF viewer in chromium headless)  
I'm aware this isn't a great solution, but it's one that works (well, except for the trace-recorder, which seems to have problems here)

dgozman:
Fixes #24378.

### DIFF
--- a/packages/playwright-test/src/index.ts
+++ b/packages/playwright-test/src/index.ts
@@ -532,6 +532,7 @@ class ArtifactsRecorder {
   private _temporaryScreenshots: string[] = [];
   private _reusedContexts = new Set<BrowserContext>();
   private _traceOrdinal = 0;
+  private _screenshotOrdinal = 0;
   private _screenshottedSymbol: symbol;
   private _startedCollectingArtifacts: symbol;
 
@@ -551,6 +552,10 @@ class ArtifactsRecorder {
     this._testInfo = testInfo;
     testInfo._onDidFinishTestFunction = () => this.didFinishTestFunction();
     this._captureTrace = shouldCaptureTrace(this._traceMode, testInfo) && !process.env.PW_TEST_DISABLE_TRACING;
+
+    // Since beforeAll(s), test and afterAll(s) reuse the same TestInfo, make sure we do not
+    // overwrite previous screenshots.
+    this._screenshotOrdinal = testInfo.attachments.filter(a => a.name === 'screenshot').length;
 
     // Process existing contexts.
     for (const browserType of [this._playwright.chromium, this._playwright.firefox, this._playwright.webkit]) {
@@ -617,9 +622,15 @@ class ArtifactsRecorder {
         await Promise.all(context.pages().map(async page => {
           if ((page as any)[this._screenshottedSymbol])
             return;
-          // Pass caret=initial to avoid any evaluations that might slow down the screenshot
-          // and let the page modify itself from the problematic state it had at the moment of failure.
-          await page.screenshot({ ...this._screenshotOptions, timeout: 5000, path: this._addScreenshotAttachment(), caret: 'initial' }).catch(() => {});
+          try {
+            const screenshotPath = this._createScreenshotAttachmentPath();
+            // Pass caret=initial to avoid any evaluations that might slow down the screenshot
+            // and let the page modify itself from the problematic state it had at the moment of failure.
+            await page.screenshot({ ...this._screenshotOptions, timeout: 5000, path: screenshotPath, caret: 'initial' });
+            this._testInfo.attachments.push({ name: 'screenshot', path: screenshotPath, contentType: 'image/png' });
+          } catch {
+            // Screenshot may fail, just ignore.
+          }
         }));
       }
     }).concat(leftoverApiRequests.map(async context => {
@@ -630,10 +641,16 @@ class ArtifactsRecorder {
     // Either remove or attach temporary screenshots for contexts closed before
     // collecting the test trace.
     await Promise.all(this._temporaryScreenshots.map(async file => {
-      if (captureScreenshots)
-        await fs.promises.rename(file, this._addScreenshotAttachment()).catch(() => {});
-      else
+      if (!captureScreenshots) {
         await fs.promises.unlink(file).catch(() => {});
+        return;
+      }
+      try {
+        const screenshotPath = this._createScreenshotAttachmentPath();
+        await fs.promises.rename(file, screenshotPath);
+        this._testInfo.attachments.push({ name: 'screenshot', path: screenshotPath, contentType: 'image/png' });
+      } catch {
+      }
     }));
 
     // Collect test trace.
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/index.ts
```diff
@@ -617,9 +618,15 @@ class ArtifactsRecorder {
...
             return;
-          // Pass caret=initial to avoid any evaluations that might slow down the screenshot
-          // and let the page modify itself from the problematic state it had at the moment of failure.
-          await page.screenshot({ ...this._screenshotOptions, timeout: 5000, path: this._addScreenshotAttachment(), caret: 'initial' }).catch(() => {});
+          try {
+            const screenshotPath = this._createScreenshotAttachmentPath();
+            // Pass caret=initial to avoid any evaluations that might slow down the screenshot
+            // and let the page modify itself from the problematic state it had at the moment of failure.
+            await page.screenshot({ ...this._screenshotOptions, timeout: 5000, path: screenshotPath, caret: 'initial' });
+            this._testInfo.attachments.push({ name: 'screenshot', path: screenshotPath, contentType: 'image/png' });
```
pavelfeldman, line 626: While I appreciate the strictless here, I think the actual bug is in the tracing processing that assumes that all the attachments always exist. I think we should fix it there first.

pavelfeldman (approved):

PR #24415, opened 2023-07-26
fix(tracing): make sure resetForReuse does not throw

dgozman:
When trace chunk recording is in progress, calling `stop()` throws `Error: Must stop trace file before stopping tracing`.


pavelfeldman (approved):
A test would be nice.

Issue #24310, opened 2023-07-19
[Feature] "Accept new snapshot" button in html-reporter


nikolay-yavorovskiy:
Hi all. Recently, on our project, we introduced screenshot tests using Playwright.
These tests have shown themselves very well - every release they catch UI bugs. We began to increase their number but rather quickly ran into a problem.

Our workflow is as follows: we created a separate project in the Playwright config - purely UI tests, which run separately from functional tests. After the run - we look at the results in the standard Playwright html-reporter. In general, it is quite convenient. It highlights the diffs, you can see both the original and the new screenshot to view - everything is cool.

But relatively often there is such a situation that the test fails for valid reasons. Not because of some bug, but simply because some UI was changed, this is, so to speak, the expected failure. But it is important when looking at failed tests not to miss those that really fail due to bugs.

The problem is that now, after running the tests it is necessary, let's say, to update 20 that fell "validly" and accept new originals for them, but for 5 - it is not necessary, because they fell due to a bug - there is no tool to do so.
You have to run the tests again with `--update-snapshots`, and the diffs will not be generated in this case. Then you need to sit in your IDE and watch and check which pictures to commit and which not.
Plus, the stability is not always 100%, sometimes there are flake tests, and they also need to be looked at, so they are not accidentally committed, and so on.
In short, there is a lot of trouble with this.

The presence of the "Accept new" button for each failed test separately - would completely solve this problem.
Therefore, I really want to have a button like "Accept new golden snapshot", so that for each failed test, a new screenshot can be accepted as the new original, after which, accordingly, it will be replaced on disk in the directory with screenshots, it will appear in the diff in the git, and you will only need to commit it.

Also, it is desirable that pressing this button turns the test into "green", that is, it would turn the test from failed to pass.

I don't know how many people use screenshot tests using Playwright, but I think everyone who does has the same problem. Playwright essentially provides everything you need for screenshot testing, the only thing missing is a convenient way to manage screenshots.

shadowusr:
Hey. I'm currently working on implementing this feature. Planning on sending a PR today. I'm not a member of playwright team though, so I'll have to get approval and discuss details with them.

shadowusr:
Okay, the accept screenshot buttons themselves are mostly ready:
![](https://i.imgur.com/vTofeNu.gif)

However, I've stumbled into some issues while implementing the aforementioned test status update and need some more time to work things out.

d-ryabtsev:
[...]

PR #24446, opened 2023-07-26
feat(ui): add accept screenshot button to the UI mode

shadowusr:
### Overview & Demo
This PR implements an ability to accept screenshots in UI mode.

### What's done
- A tick appears next to a failed screenshot. If you click on it, the actual screenshot is accepted and the button is changed to "undo".
- Save buttons appear next to actual and expected attachments of a failed screenshot. Button is disabled if the click won't change anything (screenshot's already accepted)
- If some screenshots are modified, corresponding tests are marked as modified in the test tree with yellow circle
- A "modified" filter has been added

### Why
This PR partially addresses #24310.
In short, it's highly inconvenient to switch between CLI mode and UI mode to accept screenshots. As of now, you can't interact with screenshots state in UI mode which makes its use very limited

### DIFF
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -31,6 +31,7 @@ Learn more about [test annotations](../test-annotations.md).
   - `contentType` <[string]> Content type of this attachment to properly present in the report, for example `'application/json'` or `'image/png'`.
   - `path` ?<[string]> Optional path on the filesystem to the attached file.
   - `body` ?<[Buffer]> Optional attachment body used instead of a file.
+  - `targetPath` ?<[string]> Path on the filesystem to the original file in project to which this attachment relates to.
 
 The list of files or buffers attached to the current test. Some reporters show test attachments.
 
--- a/packages/playwright-test/src/isomorphic/teleReceiver.ts
+++ b/packages/playwright-test/src/isomorphic/teleReceiver.ts
@@ -526,7 +526,7 @@ export class TeleTestCase implements reporterTypes.TestCase {
 
 export type TeleTestResult = reporterTypes.TestResult & {
   stepMap: Map<string, reporterTypes.TestStep>;
-  statusEx: reporterTypes.TestResult['status'] | 'scheduled' | 'running';
+  statusEx: reporterTypes.TestResult['status'] | 'scheduled' | 'running' | 'modified';
 };
 
 export type TeleFullProject = FullProject & { id: string };
--- a/packages/playwright-test/src/matchers/toMatchSnapshot.ts
+++ b/packages/playwright-test/src/matchers/toMatchSnapshot.ts
@@ -212,19 +212,20 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
     else
       output.push('');
 
+    const targetPath = path.relative(process.cwd(), this.snapshotPath);
     if (expected !== undefined) {
       writeFileSync(this.expectedPath, expected);
-      this.testInfo.attachments.push({ name: addSuffixToFilePath(this.snapshotName, '-expected'), contentType: this.mimeType, path: this.expectedPath });
+      this.testInfo.attachments.push({ name: addSuffixToFilePath(this.snapshotName, '-expected'), contentType: this.mimeType, path: this.expectedPath, targetPath });
       output.push(`Expected: ${colors.yellow(this.expectedPath)}`);
     }
     if (previous !== undefined) {
       writeFileSync(this.previousPath, previous);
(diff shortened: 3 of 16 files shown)


shadowusr:

pavelfeldman:
The patch is much larger than a typical contribution we get and upon looking at it, I have review comments regarding pretty much every other line, starting with the reporter bits you used to capture the original path all the way to the changes in the UI, leaving alone the UX you chose. I don't think the review will converge, so I'd suggest to either wait for us to implement it or to split it into sensible review-able patches that we would be able to land one by one.

shadowusr:
@pavelfeldman, I'm interested in implementing this feature myself, because we need it now and it's completely unclear when playwright team will be able to implement it.

> I have review comments regarding pretty much every other line

I'd appreciate if you could list at least several high-level comments regarding this PR, what specifically needs to be changed.

> leaving alone the UX you chose

If you don't think the chosen UX is the best choice, let's discuss what issues you see with it and make sure we're on the same page before I dive into further edits.

> split it into sensible review-able patches that we would be able to land one by on

Okay, I can do that, just want to make sure those PRs won't be silently closed as this one and discuss aforementioned points in advance. It takes time, and I don't want it to go to waste.

As for splitting into multiple PRs, we can do it like this:
1. Implement the most basic screenshots accepting experience by adding "save" buttons near the attachments
2. Add "Accept actual" / "Undo" button near the `toHaveScreenshot` command
3. Add test status updating and filtering capabilities

Sounds good to you?

docs/src/test-api/class-testinfo.md
```diff
@@ -31,6 +31,7 @@ Learn more about [test annotations](../test-annotations.md).
   - `contentType` <[string]> Content type of this attachment to properly present in the report, for example `'application/json'` or `'image/png'`.
   - `path` ?<[string]> Optional path on the filesystem to the attached file.
   - `body` ?<[Buffer]> Optional attachment body used instead of a file.
+  - `targetPath` ?<[string]> Path on the filesystem to the original file in project to which this attachment relates to.
```
aslushnikov, line 34: We treat public API as a huge maintenance cost; all changes or additions to the surface have to be justified and approved by all core team members.

In practice, only core functionality and highest pressing concerns result in public API changes. The "Accept screenshot button" is not important enough to justify this change, so this can't go in.

packages/playwright-test/src/isomorphic/teleReceiver.ts
```diff
@@ -526,7 +526,7 @@ export class TeleTestCase implements reporterTypes.TestCase {
 
 export type TeleTestResult = reporterTypes.TestResult & {
   stepMap: Map<string, reporterTypes.TestStep>;
-  statusEx: reporterTypes.TestResult['status'] | 'scheduled' | 'running';
+  statusEx: reporterTypes.TestResult['status'] | 'scheduled' | 'running' | 'modified';
```
aslushnikov, line 529: Telereceiver is used in a few places around; we're not comfortable adding a new execution status for such a niche usecase.

Niche features should be implemented in a well-isolated manner. An easy heuristic that could be used is "could this feature be removed by only deleting lines of code"?

packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -212,19 +212,20 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
     else
       output.push('');
 
+    const targetPath = path.relative(process.cwd(), this.snapshotPath);
```
aslushnikov, line 215: `process.cwd()` should never be used in matchers (and generally, we prefer to never use is at all).

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -166,75 +175,77 @@ export const UIModeView: React.FC<{}> = ({
...
-      <div className='vbox'>
-        <div className={'vbox' + (isShowingOutput ? '' : ' hidden')}>
-          <Toolbar>
-            <div className='section-title' style={{ flex: 'none' }}>Output</div>
-            <ToolbarButton icon='circle-slash' title='Clear output' onClick={() => xtermDataSource.clear()}></ToolbarButton>
-            <div className='spacer'></div>
-            <ToolbarButton icon='close' title='Close' onClick={() => setIsShowingOutput(false)}></ToolbarButton>
-          </Toolbar>
-          <XtermWrapper source={xtermDataSource}></XtermWrapper>
+  return <AppContext.Provider value={ { isUIMode: true } }>
```
aslushnikov, line 178: The UX: accepting the screenshot should happen next to the screenshot diff view since that's where the decision happens.

PR #24452, opened 2023-07-27
feat: add support for browserContext.on('pageerror')

### LINKED ISSUE #24466: [Feature] Expose browserContext.on('pageerror') for capturing exceptions
Currently when page issues requests, we can use the `pageerror` event on the page to capture the unhandled exceptions that happen on that given page. But if the page opens a new popup or window, we would have to manually listen to these `pageerror` on those next set of pages to be able to capture these exceptions. 

In an ideal scenario, It would be better to expose this information via the BrowserContext which would hold all the Unhandled exceptions that happen on the pages that are created within this context. 

```ts

context.on("pageerror", (error, page?) => {
   // error -> unhandled exception
});
```

It also sync with other API's like the `context.on("error", (error) => {})` where we can log all the error's that happen on all the pages created by this context.

vigneshshanmugam:
+ fix #24466
+ Adds support for exposing the `pageerror` events via `browserContext` API. 
+ Helps with capturing the overall exceptions that are thrown outside of the the current page and also captures the exceptions happens on other windows/popups. 
+ Keeps the API in sync with `context.on('request)', context.on('console'), etc..`


vigneshshanmugam:
@dgozman Thoughts? I am still trying to fix the docs and tests.

dgozman:
> @dgozman Thoughts? I am still trying to fix the docs and tests.

We considered this, but so far there is no way to get the `Page` object from the `Error`, so that one can know from which page a particular error has originated. Without page attribution, we would need a very strong reason to expose this event. Would you mind filing an issue with the usecase and linking this PR to it?

packages/playwright-core/src/client/browserContext.ts
```diff
@@ -100,6 +101,12 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
       if (page)
         page.emit(Events.Page.Console, consoleMessage);
     });
+    this._channel.on('pageError', ({ error, page }) => {
+      this.emit(Events.BrowserContext.PageError, parseError(error));
```
pavelfeldman, line 105: So I guess @dgozman wants to see 

```js
this.emit(Events.BrowserContext.PageError, { error: parseError(error), page });
```

here and wants to make sure Page is not nullable.

vigneshshanmugam: makes sense, @dgozman Do you think is there any path forward to get this in?

packages/protocol/src/protocol.yml
```diff
@@ -1183,6 +1183,11 @@ BrowserContext:
       parameters:
         page: Page
 
+    pageError:
+      parameters:
+        error: SerializedError
+        page: Page?
```
dgozman, line 1189:
```suggestion
        page: Page
```

tests/library/browsercontext-pageerror-event.spec.ts
```diff
@@ -1,5 +1,6 @@
+
```
dgozman, line 1: Let's keep `page-event-pageerror.spec.ts` as is, and instead add a test to `browsercontext-events.spec.ts`.

packages/playwright-core/src/client/browserContext.ts
vigneshshanmugam: Is the intention here to expose a `PageError` class, that has client/pageerror and server/pageerror and uses channel dispatcher to be able to be consumed in this way. 

```
context.on("pageerror", pageerror => {
   page.error.page() //  page where error happened
   pageerror.error() // exeception
});
```

Just want to make sure before I make the changes. Thank you in advance.

dgozman: Just `client/pageError.ts`. We don't need server-side `pageError` nor `pageErrorDispatcher`, the way you plumbed through `protocol.yml` looks good. The usage snippet above looks great!

utils/doclint/missingDocs.js
```diff
@@ -22,7 +22,7 @@ const path = require('path');
 
 /** @typedef {import('../../markdown').MarkdownNode} MarkdownNode */
 
-const IGNORE_CLASSES = ['PlaywrightAssertions', 'GenericAssertions', 'LocatorAssertions', 'PageAssertions', 'APIResponseAssertions', 'SnapshotAssertions'];
+const IGNORE_CLASSES = ['PlaywrightAssertions', 'GenericAssertions', 'LocatorAssertions', 'PageAssertions', 'APIResponseAssertions', 'SnapshotAssertions', 'PageError'];
```
vigneshshanmugam, line 25: @dgozman Any idea how to solve this erorr from docs lint? 
```
> node utils/doclint/cli.js

ERROR: missing documentation:
Documented "PageError" not found in sources
```

packages/playwright-core/src/client/browserContext.ts
vigneshshanmugam: Great. Added the necessary changes.

docs/src/api/class-pageerror.md
```diff
@@ -0,0 +1,23 @@
+# class: PageError
+* since: v1.38
+
+[PageError] class represents objects created by context when there are unhandled
+execeptions thrown on the pages and dispatched via the [`event: BrowserContext.pageError`] event
+
+```js
```
mxschmitt, line 7: We should add here snippets for the other language ports before merging.

packages/playwright-core/src/client/page.ts
```diff
@@ -23,7 +23,7 @@ import { isSafeCloseError, kBrowserOrContextClosedError } from '../common/errors
 import { urlMatches } from '../utils/network';
 import { TimeoutSettings } from '../common/timeoutSettings';
 import type * as channels from '@protocol/channels';
-import { parseError, serializeError } from '../protocol/serializers';
+import {  serializeError } from '../protocol/serializers';
```
mxschmitt, line 26:
```suggestion
import { serializeError } from '../protocol/serializers';
```

utils/doclint/missingDocs.js
mxschmitt: You can add `export type { PageError } from './pageError';` into `packages/playwright-core/src/client/api.ts` and then the error is gone.

packages/playwright-core/src/server/fetch.ts
```diff
@@ -730,4 +729,4 @@ function shouldBypassProxy(url: URL, bypass?: string): boolean {
   });
   const domain = '.' + url.hostname;
   return domains.some(d => domain.endsWith(d));
-}
\ No newline at end of file
+}
```
mxschmitt, line 732: nit: line endings have changed.

mxschmitt (changes requested):

docs/src/api/class-pageerror.md
vigneshshanmugam: Added https://github.com/microsoft/playwright/pull/24452/commits/636e710a01ff8e3982c372459fe96372dbfd0d06

packages/playwright-core/src/client/api.ts
```diff
@@ -42,3 +42,4 @@ export { Video } from './video';
 export { Worker } from './worker';
 export { CDPSession } from './cdpSession';
 export { Playwright } from './playwright';
+export type { PageError } from './pageError';
```
dgozman, line 45:
```suggestion
export { PageError } from './pageError';
```

packages/playwright-core/src/server/chromium/crPage.ts
```diff
@@ -859,7 +859,7 @@ class FrameSession {
   }
 
   _handleException(exceptionDetails: Protocol.Runtime.ExceptionDetails) {
-    this._page.firePageError(exceptionToError(exceptionDetails));
+    this._page.emitOnContext(BrowserContext.Events.PageError, exceptionToError(exceptionDetails), this._page);
```
dgozman, line 862: Let's use `emitOnContextOnceInitialized` everywhere instead.

packages/playwright-core/src/client/browserContext.ts
```diff
@@ -100,6 +102,13 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
       if (page)
         page.emit(Events.Page.Console, consoleMessage);
     });
+    this._channel.on('pageError', ({ error, page }) => {
+      const pageObject = Page.fromNullable(page);
```
dgozman, line 106: Since it's non-nullable in the `protocol.yml`:

```suggestion
      const pageObject = Page.from(page);
```

dgozman (commented):
Looks very good, just a few comments.

I think we should leave page as non-nullable in the `protocol.yml` and as nullable in the api, like you already have.

vigneshshanmugam:
@dgozman :yay Thank you

docs/src/api/class-pageerror.md
```diff
@@ -0,0 +1,62 @@
...
+  System.out.println("Uncaught exception: " + pagerror.error());
+});
+
+// Navigate to a page with an exception.
+page.navigate("data:text/html,<script>throw new Error('Test')</script>");
+```
+
+```python async
+# Log all uncaught errors to the terminal
+context.on("pageerror", lambda pageerror: print(f"uncaught exception: {pageerror.error()}"))
```
mxschmitt, line 29:
```suggestion
context.on("pageerror", lambda pageerror: print(f"uncaught exception: {pageerror.error}"))
```

docs/src/api/class-pageerror.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Navigate to a page with an exception.
+await page.goto("data:text/html,<script>throw new Error('test')</script>")
+```
+
+```python sync
+# Log all uncaught errors to the terminal
+context.on("pageerror", lambda pageerror: print(f"uncaught exception: {pageerror.error()}"))
```
mxschmitt, line 37:
```suggestion
context.on("pageerror", lambda pageerror: print(f"uncaught exception: {pageerror.error}"))
```

docs/src/api/class-pageerror.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Navigate to a page with an exception.
+page.goto("data:text/html,<script>throw new Error('test')</script>")
+```
+
+```csharp
+// Log all uncaught errors to the terminal
+context.PageError += (_, pageerror) =>
+{
+  Console.WriteLine("Uncaught exception: " + pageerror.error());
```
mxschmitt, line 47:
```suggestion
  Console.WriteLine("Uncaught exception: " + pageerror.Error);
```

mxschmitt (approved):

docs/src/api/class-pageerror.md
```diff
@@ -0,0 +1,62 @@
...
+// Log all uncaught errors to the terminal
+context.PageError += (_, pageerror) =>
+{
+  Console.WriteLine("Uncaught exception: " + pageerror.Error);
+};
+```
+
+## method: PageError.page
+* since: v1.38
+- returns: <[null]|[Page]>
```
yury-s, line 53: If it's a page error it implies that there is a page where the error was thrown, let's change it to be non nullable?

packages/playwright-core/src/client/browserContext.ts
```diff
@@ -100,6 +102,13 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
       if (page)
         page.emit(Events.Page.Console, consoleMessage);
     });
+    this._channel.on('pageError', ({ error, page }) => {
+      const pageObject = Page.from(page);
+      const parsedError = parseError(error);
+      this.emit(Events.BrowserContext.PageError, new PageError(pageObject, parsedError));
+      if (pageObject)
```
yury-s, line 109: This is never null as otherwise Page.from would throw, remove this check.

docs/src/api/class-pageerror.md
yury-s: Never mind, I see it was suggested [above](https://github.com/microsoft/playwright/pull/24452#pullrequestreview-1557970923), I'll discuss this with Dmitry.

PR #24473, opened 2023-07-27
devops: fix installation tests on low disk space machines

mxschmitt:
Before all the browser downloads were stored in the workspace folder for the corresponding test. After this change, we clean it up once the test has finished to save disk space.

~800 MB per test before, now 30MB.

### DIFF
--- a/tests/installation/npmTest.ts
+++ b/tests/installation/npmTest.ts
@@ -101,8 +101,10 @@ export const test = _test
         const contents = await fs.promises.readdir(sourceDir);
         await Promise.all(contents.map(f => fs.promises.copyFile(path.join(sourceDir, f), path.join(tmpWorkspace, f))));
         await use();
-        // Browsers are large, we remove them after each test to save disk space.
-        await fs.promises.rm(_browsersPath, { recursive: true, force: true });
+        if (test.info().status === test.info().expectedStatus) {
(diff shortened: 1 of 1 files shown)


tests/installation/npmTest.ts
```diff
@@ -93,12 +94,15 @@ export type NPMTestFixtures = {
...
-      _auto: [async ({ tmpWorkspace, exec }, use) => {
+      _browsersPath: async ({tmpWorkspace}, use) => use(path.join(tmpWorkspace, 'browsers')),
+      _auto: [async ({ tmpWorkspace, exec, _browsersPath }, use) => {
         await exec('npm init -y');
         const sourceDir = path.join(__dirname, 'fixture-scripts');
         const contents = await fs.promises.readdir(sourceDir);
         await Promise.all(contents.map(f => fs.promises.copyFile(path.join(sourceDir, f), path.join(tmpWorkspace, f))));
         await use();
+        // Browsers are large, we remove them after each test to save disk space.
+        await fs.promises.rm(_browsersPath, { recursive: true, force: true });
```
aslushnikov, line 105: Can we remove only for successful tests? This way debugging unsuccessful tests would be easier

mxschmitt: done

aslushnikov (approved):

Issue #24451, opened 2023-07-27
[internal] merge report action items


dgozman:
- [x] Document `--config` option.
- [x] When each shard runs the same `setup` project, it is merged incorrectly because of clashing test ids.
- [x] (internal) Do not put `blob-report` inside `test-results`.
- [x] Copy default `blob-report` location logic from html reporter - next to `package.json`.
- [x] Copy default `report.md` markdown report location logic from html reporter - next to `package.json`.
- [x] `blob-report` output directory should be removed at the start.
- [x] Document sharding+blob+merge yaml files.
- [x] Change the `report.zip` naming scheme with suffix and shard number.
- [x] OOM
- [x] Remove `run_attempt` from all yamls and docs.
- [x] Check playwright or blob versions from each shard and throw if they do not match.
- [x] Replace "projectSuffix" with "reportName".
- [x] Split `onBegin` per project (https://github.com/microsoft/playwright/pull/26329)
---
- [x] Update yaml in the docs to print the final report url from Azure.
- [x] Add GHA screenshots to the doc.

PR #24481, opened 2023-07-28
fix(blob report): default location relative to package.json

dgozman:
Also:
- remove `blob-report` directory at the start;
- markdown's `report.md` next to package.json;
- use default location in playwright's workflows.

References #24451.

### DIFF
--- a/packages/playwright-test/src/reporters/blob.ts
+++ b/packages/playwright-test/src/reporters/blob.ts
@@ -16,15 +16,15 @@
 
 import fs from 'fs';
 import path from 'path';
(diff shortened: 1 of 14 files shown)


packages/playwright-test/src/reporters/blob.ts
```diff
@@ -51,8 +51,18 @@ export class BlobReporter extends TeleReporterEmitter {
     this._salt = createGuid();
   }
 
+  private _resolveOutputDir(): string {
+    if (this._options.outputDir)
```
pavelfeldman, line 55: Surprised this to be not sharing code with html reporter.

pavelfeldman (approved):

PR #24500, opened 2023-07-28
chore: emit goto upon omnibox navigation during page.fill

### LINKED ISSUE #24499: [BUG] Navigation actions not recorded after some actions
### System info
- Playwright Version: v1.36.1
- Browser: Chromium
- Other info:

Some navigation actions are not recorded. For instance, if I start filling an input field and then enter a new URL in the navigation bar, it won't record the navigation action. I provide a video and a test (it needs to run with `inspectorTest` fixtures).

https://github.com/microsoft/playwright/assets/1374559/bd0eb261-4dda-4f04-a821-8c31dd477c62

### Source code

https://github.com/ruifigueira/playwright-crx/pull/new/bug/navigation-not-recorded

**Steps**
- Run:

```
npm run ctest -- tests/library/inspector/recorder.spec.ts
```

**Expected**

```
Array [
  "  const page = await context.newPage();",
  "  await page.goto('http://localhost:8907/input/textarea.html');",
  "  await page.locator('textarea').fill('hello world');",
  "  await page.goto('http://localhost:8907/input/button.html');",
  "  await page.getByRole('button', { name: 'Click target' }).click();",
]
```

**Actual**

Navigation to `http://localhost:8907/input/button.html` is lost:

```
Array [
  "  const page = await context.newPage();",
  "  await page.goto('http://localhost:8907/input/textarea.html');",
  "  await page.locator('textarea').fill('hello world');",
  "  await page.getByRole('button', { name: 'Click target' }).click();",
]
```

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/24499

### DIFF
--- a/packages/playwright-core/src/server/recorder/codeGenerator.ts
+++ b/packages/playwright-core/src/server/recorder/codeGenerator.ts
@@ -130,7 +130,9 @@ export class CodeGenerator extends EventEmitter {
       this._currentAction.action.signals.push(signal);
       return;
     }
-    if (this._lastAction && !this._lastAction.committed) {
+
+    // Consider fill action as committed for the purposes of navigation signal processing.
+    if (this._lastAction && !this._lastAction.committed && this._lastAction.action.name !== 'fill') {
       const signals = this._lastAction.action.signals;
       if (signal.name === 'navigation' && signals.length && signals[signals.length - 1].name === 'download')
         return;


dgozman (approved):

packages/playwright-core/src/server/recorder/codeGenerator.ts
```diff
@@ -130,7 +130,9 @@ export class CodeGenerator extends EventEmitter {
       this._currentAction.action.signals.push(signal);
       return;
     }
-    if (this._lastAction && !this._lastAction.committed) {
+
+    // Consider fill action as committed for the purposes of navigation signal processing.
+    if (this._lastAction && !this._lastAction.committed && this._lastAction.action.name !== 'fill') {
```
ruifigueira, line 135: It also happens in a click followed by a navigation.
In my test, if you reorder the calls, you'll have the same issue:

```ts
  await page.goto(server.PREFIX + '/input/button.html');
  await page.click('button');
  await page.goto(server.PREFIX + '/input/textarea.html');
  await page.fill('textarea', 'hello world');
```

I understand that the navigation can be interpreted as the effect of clicking the button if they're almost at the same time, so I'm not sure if there's a way to distinguish a navigation caused by a click from a navigation caused by entering a URL in the navigation bar.

pavelfeldman: You are right, I thought that we commit clicks sooner than other actions. I don't want to reduce the commit timer and I am not sure I want to commit on mouse move, so I guess existing imperfection is the lesser evil. I'll close the issue and the PR.

PR #24528, opened 2023-07-31
docs: improve auto-retrying vs generic assertions

### LINKED ISSUE #24460: [Feature] Generic assertion to check Object against structure (keys name and theirs type)
By investigate: https://playwright.dev/docs/api/class-genericassertions#generic-assertions-to-match-object

I tried every single assertion but I can't assert 2 object against only structure (I mean theirs keys and type). The closes to my expectation is `toMatchObject` but this function also check detailed value of key.

Below example of my problem:

```
let receivedObject = {
name:'Andrew',
age: 15
 };
 
let templateObject =  {
name: 'Jack',
age: 21
};

expect(receivedObject).toMatchObject(templateObject) // False
```
I would like to be able to call an assertion that will check the **types of keys** (string -> string, number -> number) and their **names** - that is, in the above case, return true.
Problem I can solve by using additional `chai` library, but do not want to install additional node package.

The only thing I managed is to bring the comparison by key names only, but this is not a full test in my case. By using empty object as a keys values in `templateObject`:

```
let receivedObject = {
name:'Andrew',
age: 15
 };
 
let templateObject =  
{ 
name: {},
age: {}
};

expect(receivedObject).toMatchObject(templateObject) // True
```

dgozman:
Fixes #24460.

### DIFF
--- a/docs/src/test-assertions-js.md
+++ b/docs/src/test-assertions-js.md
@@ -20,37 +20,44 @@ Playwright will be re-testing the element with the test id of `status` until the
 
 By default, the timeout for assertions is set to 5 seconds. Learn more about [various timeouts](./test-timeouts.md).
 
-## Popular assertions
+## Auto-retrying assertions
+
+The following assertions will retry until the assertion passes, or the assertion timeout is reached.
+Note that retrying assertions are async, so you must `await` them.
 
 | Assertion | Description |
 | :- | :- |
-| [`method: LocatorAssertions.toBeAttached`] | Element is attached |
-| [`method: LocatorAssertions.toBeChecked`] | Checkbox is checked |
-| [`method: LocatorAssertions.toBeDisabled`] | Element is disabled |
-| [`method: LocatorAssertions.toBeEditable`] | Element is editable |
-| [`method: LocatorAssertions.toBeEmpty`] | Container is empty |
-| [`method: LocatorAssertions.toBeEnabled`] | Element is enabled |
-| [`method: LocatorAssertions.toBeFocused`] | Element is focused |
-| [`method: LocatorAssertions.toBeHidden`] | Element is not visible |
-| [`method: LocatorAssertions.toBeInViewport`] | Element intersects viewport |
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):

docs/src/test-assertions-js.md
```diff
@@ -20,37 +20,44 @@ Playwright will be re-testing the element with the test id of `status` until the
...
+| [await expect(locator).toHaveValue()](./api/class-locatorassertions.md#locator-assertions-to-have-value) | Input has a value |
+| [await expect(locator).toHaveValues()](./api/class-locatorassertions.md#locator-assertions-to-have-values) | Select has options selected |
+| [await expect(page).toHaveScreenshot()](./api/class-pageassertions.md#page-assertions-to-have-screenshot-1) | Page has a screenshot |
+| [await expect(page).toHaveTitle()](./api/class-pageassertions.md#page-assertions-to-have-title) | Page has a title |
+| [await expect(page).toHaveURL()](./api/class-pageassertions.md#page-assertions-to-have-url) | Page has a URL |
+| [await expect(response).toBeOK()](./api/class-apiresponseassertions.md#api-response-assertions-to-be-ok) | Response has an OK status |
+
+## Non-retrying assertions
+
+These assertions allow to test any conditions, but do no auto-retry. Most of the time, web pages show information asynchronously, and using non-retrying assertions can lead to a flaky test.
```
aslushnikov, line 58:
```suggestion
These assertions allow to test any conditions, but do not auto-retry. Most of the time, web pages show information asynchronously, and using non-retrying assertions can lead to a flaky test.
```

docs/src/test-assertions-js.md
```diff
@@ -20,37 +20,44 @@ Playwright will be re-testing the element with the test id of `status` until the
 
 By default, the timeout for assertions is set to 5 seconds. Learn more about [various timeouts](./test-timeouts.md).
 
-## Popular assertions
+## Auto-retrying assertions
```
aslushnikov, line 23:
```suggestion
## (Recommended) Auto-retrying assertions
```

PR #16733, opened 2022-08-22
chore: do not use experimental loader for web server

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/16319

PR #24539, opened 2023-08-01
chore: allow calling spawnSync on Node.js file inside test

### LINKED ISSUE #24516: Running node execSync not working in Playwright
### System info
- Playwright Version: 1.36.2
- Operating System: macOS 13.5
- Browser: [All]
- Other info:

### Source code

Here is the [repo](https://github.com/21RISK/playwright-issue)

[https://github.com/your_profile/playwright_issue_title]

**Steps**
- Download repo and run tests with `npx playwright test`

**Expected**

I would expect "bente" in the console

**Actual**

The fixture halts at line 15, `console.log(execSync('node ./test-js-file.js').toString());` - and never continues...

**comments**

Please notice that the project has configured `type: "module"` in the root package.json. With this removed, it actually does work - so I think it has something to do with cjs/mjs, but I'm a bit unsure.

Running `node ./test-js-file.js` from the cmd line works without issues.

mxschmitt:
The issue was that when `spawnSync` gets used, `NODE_OPTIONS` is set with our experimental loader, I think we can't prevent that from happening.
~~So then a Node.js script gets executed with the loader but we never ack the messages we send inside the loader IPC.
This change adds a small `init` method, if its not called, we never initialize the transport.~~

Fixes https://github.com/microsoft/playwright/issues/24516 

Relates https://github.com/microsoft/playwright/pull/16733

### DIFF
--- a/packages/playwright-test/src/common/process.ts
+++ b/packages/playwright-test/src/common/process.ts
@@ -51,6 +51,7 @@ process.on('disconnect', gracefullyCloseAndExit);
 process.on('SIGINT', () => {});
 process.on('SIGTERM', () => {});
 
+// Clear execArgv immediately, so that the user-code does not inherit our loader.
 process.execArgv = execArgvWithoutExperimentalLoaderOptions();
 
 let processRunner: ProcessRunner;
--- a/packages/playwright-test/src/common/process.ts
+++ b/packages/playwright-test/src/common/process.ts
@@ -18,7 +18,7 @@ import type { WriteStream } from 'tty';
 import type { EnvProducedPayload, ProcessInitParams, TtyParams } from './ipc';
 import { startProfiling, stopProfiling } from 'playwright-core/lib/utils';
 import type { TestInfoError } from '../../types/test';
-import { serializeError } from '../util';
+import { execArgvWithoutExperimentalLoaderOptions, serializeError } from '../util';
 
 export type ProtocolRequest = {
   id: number;
@@ -50,6 +50,8 @@ sendMessageToParent({ method: 'ready' });
 process.on('disconnect', gracefullyCloseAndExit);
 process.on('SIGINT', () => {});
 process.on('SIGTERM', () => {});
+
+process.execArgv = execArgvWithoutExperimentalLoaderOptions();
 
 let processRunner: ProcessRunner;
 let processName: string;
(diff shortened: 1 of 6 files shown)


packages/playwright-test/src/util.ts
```diff
@@ -307,16 +307,8 @@ function folderIsModule(folder: string): boolean {
...
-  return ` --no-warnings --experimental-loader=${url.pathToFileURL(require.resolve('@playwright/test/lib/transform/esmLoader')).toString()}`;
-}
-
-export function envWithoutExperimentalLoaderOptions(): NodeJS.ProcessEnv {
-  const substring = experimentalLoaderOption();
-  const result = { ...process.env };
-  if (result.NODE_OPTIONS)
-    result.NODE_OPTIONS = result.NODE_OPTIONS.replace(substring, '').trim() || undefined;
-  return result;
+export function experimentalLoaderOptions() {
```
dgozman, line 310: Shouldn't we also use this in `processHost` when spawning workers? How do our tests pass?

mxschmitt: main process restarts with it, so all it's child processes inherit the same NODE_OPTIONS.

packages/playwright-test/src/runner/processHost.ts
```diff
@@ -50,6 +51,7 @@ export class ProcessHost extends EventEmitter {
       detached: false,
       env: { ...process.env, ...this._extraEnv },
       stdio: inheritStdio ? ['ignore', 'inherit', 'inherit', 'ipc'] : ['ignore', 'ignore', process.env.PW_RUNNER_DEBUG ? 'inherit' : 'ignore', 'ipc'],
+      execArgv: experimentalLoaderOptions(),
```
dgozman, line 54: We do not need experimental options in non-TSM mode. I think we should check `process.env.PW_TS_ESM_ON` here to make it conditional.

packages/playwright-test/src/common/process.ts
```diff
@@ -51,6 +51,8 @@ process.on('disconnect', gracefullyCloseAndExit);
 process.on('SIGINT', () => {});
 process.on('SIGTERM', () => {});
 
+process.execArgv = execArgvWithoutExperimentalLoaderOptions();
```
dgozman, line 54: Same comment here.

packages/playwright-test/src/util.ts
```diff
@@ -307,16 +307,17 @@ function folderIsModule(folder: string): boolean {
...
 
-export function envWithoutExperimentalLoaderOptions(): NodeJS.ProcessEnv {
-  const substring = experimentalLoaderOption();
-  const result = { ...process.env };
-  if (result.NODE_OPTIONS)
-    result.NODE_OPTIONS = result.NODE_OPTIONS.replace(substring, '').trim() || undefined;
-  return result;
+export function execArgvWithoutExperimentalLoaderOptions() {
+  const execArgv = execArgvWithExperimentalLoaderOptions();
+  return process.execArgv.filter(arg => !execArgv.includes(arg));
```
dgozman, line 320: This is going to remove everything from `process.execArgv` because `execArgvWithExperimentalLoaderOptions()` includes `process.execArgv` inside.

dgozman (approved):

PR #24560, opened 2023-08-02
chore(test): expose expect matchers from @playwright/test

### LINKED ISSUE #24501: [Feature] export @playwright/test expect matchers
+ `@playwright/test` has great set of assertions built in top of some of the built in asserts from `jest mactchers` which can be consumed from the library via `import {expect} from "@playwright/test"`. Examples are https://playwright.dev/docs/test-assertions#list-of-assertions

+ Would it be possible for exporting the `expect` matchers from Playwright test to be consumed by other test runners like `jest, @elastic/synthetics, etc..`. I am a maintainer on the later package and would be interested in swapping the jest expect matchers with the one from the PW test. 

+ Even if not a standalone package which might require structural change in the PW library itself, would it possible to add the `lib/matchers` to the export list on `playwright/test/package.json` ? https://github.com/microsoft/playwright/blob/d92fe16b76272afb19e7af5a2496f7efce45441d/packages/playwright-test/package.json#L11
```ts
exports: {
 "./lib/expect": "./lib/matchers/expect.js"
}
```

I am not sure if there are consequences in doing so, Since PW test relies on some of the functionalities associated with the PW-core. If this PW team thinks this approach is good enough, I am happy to do a PR.

vigneshshanmugam:
+ fix https://github.com/microsoft/playwright/issues/24501
+ Allows libraries building on top of the `playwright-core` to use the extended expect matchers from the `@playwright/test`.

### DIFF
--- a/packages/playwright-test/package.json
+++ b/packages/playwright-test/package.json
@@ -23,6 +23,7 @@
     "./lib/transform/esmLoader": "./lib/transform/esmLoader.js",
     "./lib/internalsForTest": "./lib/internalsForTest.js",
     "./lib/plugins": "./lib/plugins/index.js",
+    "./lib/matchers/expect": "./lib/matchers/expect.js",
     "./jsx-runtime": {
       "import": "./jsx-runtime.mjs",
       "require": "./jsx-runtime.js",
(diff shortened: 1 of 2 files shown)


pavelfeldman:
Marking it out of scope for now. I'm not sure we want to commit to supporting this mode.

PR #24567, opened 2023-08-02
fix: use `snapshotPath` instead of `expectedPath` for image diffs

aslushnikov:
This opens a road to the "accept new screenshot" button in
Playwright tools.

References https://github.com/microsoft/playwright/issues/24310

### DIFF
--- a/packages/playwright-test/src/matchers/toMatchSnapshot.ts
+++ b/packages/playwright-test/src/matchers/toMatchSnapshot.ts
@@ -43,7 +43,7 @@ type SnapshotNames = {
 class SnapshotHelper<T extends ImageComparatorOptions> {
   readonly testInfo: TestInfoImpl;
   readonly snapshotName: string;
-  readonly expectedPath: string;
+  readonly legacyExpectedPath: string;
   readonly previousPath: string;
   readonly snapshotPath: string;
   readonly actualPath: string;
@@ -126,7 +126,7 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
     this.snapshotPath = snapshotPathResolver(...inputPathSegments);
     const inputFile = testInfo._getOutputPath(...inputPathSegments);
     const outputFile = testInfo._getOutputPath(...outputPathSegments);
-    this.expectedPath = addSuffixToFilePath(inputFile, '-expected');
+    this.legacyExpectedPath = addSuffixToFilePath(inputFile, '-expected');
     this.previousPath = addSuffixToFilePath(outputFile, '-previous');
     this.actualPath = addSuffixToFilePath(outputFile, '-actual');
     this.diffPath = addSuffixToFilePath(outputFile, '-diff');
@@ -213,12 +213,11 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
       output.push('');
 
     if (expected !== undefined) {
-      writeFileSync(this.expectedPath, expected);
-      // Note: use actual snapshot path in the snapshots directory instead of the one
-      // copied over to the `test-results/` folder.
-      // This way we can use this path later on when we need to update snapshot expectation.
+      // Copy the expectation inside the `test-results/` folder for backwards compatibility,
+      // so that one can upload `test-results/` directory and have all the data inside.
+      writeFileSync(this.legacyExpectedPath, expected);
       this.testInfo.attachments.push({ name: addSuffixToFilePath(this.snapshotName, '-expected'), contentType: this.mimeType, path: this.snapshotPath });
-      output.push(`Expected: ${colors.yellow(this.expectedPath)}`);
+      output.push(`Expected: ${colors.yellow(this.snapshotPath)}`);
     }
     if (previous !== undefined) {
       writeFileSync(this.previousPath, previous);
--- a/packages/playwright-test/src/matchers/toMatchSnapshot.ts
+++ b/packages/playwright-test/src/matchers/toMatchSnapshot.ts
@@ -214,7 +214,10 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
 
     if (expected !== undefined) {
       writeFileSync(this.expectedPath, expected);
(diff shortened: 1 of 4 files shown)


packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -214,7 +214,10 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
 
     if (expected !== undefined) {
       writeFileSync(this.expectedPath, expected);
-      this.testInfo.attachments.push({ name: addSuffixToFilePath(this.snapshotName, '-expected'), contentType: this.mimeType, path: this.expectedPath });
+      // Note: use actual snapshot path in the snapshots directory instead of the one
+      // copied over to the `test-results/` folder.
+      // This way we can use this path later on when we need to update snapshot expectation.
+      this.testInfo.attachments.push({ name: addSuffixToFilePath(this.snapshotName, '-expected'), contentType: this.mimeType, path: this.snapshotPath });
       output.push(`Expected: ${colors.yellow(this.expectedPath)}`);
```
dgozman, line 221: - Perhaps change output to `snapshotPath` as well?
- More broadly, I'd remove `this.expectedPath` and `snapshotPath` everywhere, only calculating it here with a comment "additionally copy the expectation to the old location for backwards compatibility, so that one can upload test-results directory and have all the data inside"

dgozman (approved):

packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -213,9 +213,11 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
       output.push('');
 
     if (expected !== undefined) {
-      writeFileSync(this.expectedPath, expected);
-      this.testInfo.attachments.push({ name: addSuffixToFilePath(this.snapshotName, '-expected'), contentType: this.mimeType, path: this.expectedPath });
-      output.push(`Expected: ${colors.yellow(this.expectedPath)}`);
+      // Copy the expectation inside the `test-results/` folder for backwards compatibility,
+      // so that one can upload `test-results/` directory and have all the data inside.
+      writeFileSync(this.legacyExpectedPath, expected);
+      this.testInfo.attachments.push({ name: addSuffixToFilePath(this.snapshotName, '-expected'), contentType: this.mimeType, path: this.snapshotPath });
```
shadowusr, line 219: I see a few issues here:
- `this.snapshotPath` as far as I remember stores an absolute path, meaning everything will break once trace file is moved to another host
- the `path` property will get erased [here](https://github.com/microsoft/playwright/blob/a0211924d328c78d374b3b79255296efe5f4d718/packages/playwright-core/src/utils/traceUtils.ts#L132). So while this PR won't break anything, it still leaves a few problems unsolved on the way to accept screenshot button
- even though it shouldn't break anything because of resolution via sha1 hashes in service worker, personally, I find it weird that expected path points to one directory while all other attachments (previous, actual, diff) point to the other (if I'm not mistaken, this way expected path will lead us inside user's project while other paths will point to results directory)

aslushnikov: > this.snapshotPath as far as I remember stores an absolute path, meaning everything will break once trace file is moved to another host

@shadowusr All of the paths here are absolute, including the `this.legacyExpectedPath`. What do you mean with "trace files breaking"? 

> the path property will get erased [here](https://github.com/microsoft/playwright/blob/a0211924d328c78d374b3b79255296efe5f4d718/packages/playwright-core/src/utils/traceUtils.ts#L132). So while this PR won't break anything, it still leaves a few problems unsolved on the way to accept screenshot button

This PR is not directly targeted towards the UI mode button you're interested in :) however, it might help there as well!

 > I find it weird that expected path points to one directory while all other attachments (previous, actual, diff) point to the other (if I'm not mistaken, this way expected path will lead us inside user's project while other paths will point to results directory)

That's a fair concern and that's why it was implemented this way originally. However, it's also confusing that "expectedPath" does NOT point to the actual snapshot. To the best of our knowledge, no one relies upon `expectedPath` location on disk (and should not), so changing one to another should be safe, while enabling some of the use cases we're interested in.

shadowusr: > All of the paths here are absolute

Ah, my bad, have just tried tinkering with playwright and confirmed what you said. For some reason I thought otherwise. I see you don't rely on those paths anywhere and use sha1 hashes instead.

However, using absolute path will most likely become an issue when you'll need it for the accept screenshot button.

> This PR is not directly targeted towards the UI mode button you're interested in :)

Ah, okay. I was confused by your PR's description "opens a road to the accept button".

As for the last point, I of course don't insist, but I still believe that all of those paths should point to the `outputDir` and don't see any disadvantages of introducing a new property that will point to the user's project.

aslushnikov: > As for the last point, I of course don't insist, but I still believe that all of those paths should point to the outputDir and don't see any disadvantages of introducing a new property that will point to the user's project.

We consider public API to be of *enormous* maintenance cost, so adding a new one requires strong justification and unanimous approval from all core team members. Small features like this one don't justify public API changes.

shadowusr: It makes sense to touch public API as little as possible, however I'm not sure I understand how modifying existing public API behaviour (technically, a breaking change) can be justified while adding a new field (not a breaking change) can't

aslushnikov: > however I'm not sure I understand how modifying existing public API behaviour (technically, a breaking change)

@shadowusr any bugfix is a breaking change to the current behavior, but not necessarily breaking any explicitly defined contracts (aka "public API" / documentation). 

Here, we do change current behavior, but we don't break any contracts that we uphold.

If this change regresses our clients, and their usecases are legit, then we will revert the change and enhance our documentation to explicitly state the contract they rely upon.

The foregoing risk is the reason why this change is landed separately.

PR #24568, opened 2023-08-02
chore: prefer import over require

benmccann:
The next major version of Vite may start producing warnings when Vite is required to encourage people to move off of the legacy CJS syntax and onto modern ESM. We should make a change now to get ahead of it and ensure users don't see this warning

### DIFF
--- a/packages/playwright-ct-core/src/vitePlugin.ts
+++ b/packages/playwright-ct-core/src/vitePlugin.ts
@@ -123,7 +123,7 @@ export function createPlugin(
           }
         };
       }
-      const { build, preview } = require('vite');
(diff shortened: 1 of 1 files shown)


bluwy:
Line 78 can also be updated:

```diff
- const { version: viteVersion } = require('vite/package.json')
+ const { version: viteVersion } = await import('vite')
```

Vite exports the `version` since Vite 3

pavelfeldman (approved):

pavelfeldman:
Unfortunately, this is a major performance regression due to importing the entire module vs parsing a single json file in a fast path, so I'll need to revert it.

PR #24571, opened 2023-08-02
bug: fixed `PW_TEST_HTML_REPORT_OPEN`  + more type safe + doc

mastrzyz:
# What?
- Added typing guards for using `PW_TEST_HTML_REPORT_OPEN` 
- Added `PW_TEST_HTML_REPORT_OPEN` entry in documentation

# Why?

You could accidentally put `foo` as part of PW_TEST_HTML_REPORT_OPEN and never been the wiser it was wrong since we had to console output.

# Test
Added a test and pushed the recent reporter code locally into stable-test-runner

image


packages/playwright-test/src/reporters/html.ts
```diff
@@ -140,6 +147,20 @@ function reportFolderFromEnv(): string | undefined {
...
+    if (isHtmlReportOption(htmlOpenEnv)){
+      return htmlOpenEnv;
+    } else {
+      console.log(colors.red(`Configuration Error: HTML reporter Invalid value for ${processKey}: ${htmlOpenEnv}. Valid values are: ${htmlReportOptions.join(', ')}`));
+      return undefined;
+    }
+  }
+  return undefined;
+}
+
```
mxschmitt, line 163:
```suggestion
function getHtmlReportOptionProcessEnv(): HtmlReportOpenOption | undefined {
  const processKey = 'PW_TEST_HTML_REPORT_OPEN';
  const htmlOpenEnv = process.env[processKey];
  if (!htmlOpenEnv)
    return undefined;
  if (!isHtmlReportOption(htmlOpenEnv)) {
    console.log(colors.red(`Configuration Error: HTML reporter Invalid value for ${processKey}: ${htmlOpenEnv}. Valid values are: ${htmlReportOptions.join(', ')}`));
    return undefined
  }
  return htmlOpenEnv;
}

```


Negating it is usually more readable and less nested.

mastrzyz: Sounds good!

pavelfeldman:
I'm not sure I follow the change - you can always use your own env variable in your config to control the HTML report behavior, why are the changes to Playwright necessary?

mastrzyz:
> I'm not sure I follow the change - you can always use your own env variable in your config to control the HTML report behavior, why are the changes to Playwright necessary?

@pavelfeldman  sorry for the confusion, I made this change to guard against cases where you use the existing `PW_TEST_HTML_REPORT_OPEN` env variable but add a typo or a mistake, for example;


`PW_TEST_HTML_REPORT_OPEN=allways yarn test`

Normally this would of just went through and lead to confusion, I added a warning instead.

Hope this explains the goal of this PR.

const htmlReportOptions = ['always', 'never', 'on-failure'];

mxschmitt (approved):

PR #24577, opened 2023-08-02
chore: ensure the file operations are async in the junit reporter

mastrzyz:
# What ?

Converts the file system operations in the json reporter to be async.
  
# Why?

Although I doubt it how profound this will ever be , there could be some performance bottleneck with multiple reporters at once here eventually.

## existsSync

There is no equivalent in async so I leveraged the reccomended `access` one, see doc string for more info -> 

```
     * Tests a user's permissions for the file or directory specified by `path`.
     * The `mode` argument is an optional integer that specifies the accessibility
     * checks to be performed. Check `File access constants` for possible values
     * of `mode`. It is possible to create a mask consisting of the bitwise OR of
     * two or more values (e.g. `fs.constants.W_OK | fs.constants.R_OK`).
     *
     * If the accessibility check is successful, the promise is resolved with no
     * value. If any of the accessibility checks fail, the promise is rejected
     * with an [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) object. The following example checks if the file`/etc/passwd` can be read and
     * written by the current process.
     *
     * ```js
     * import { access } from 'fs/promises';
     * import { constants } from 'fs';
     *
     * try {
     *   await access('/etc/passwd', constants.R_OK | constants.W_OK);
     *   console.log('can access');
     * } catch {
     *   console.error('cannot access');
     * }
```

# Verified?

I ran `npm run build` and then `CI=1 npm run test` with `    ['junit', { outputFile: path.join(outputDir, 'report.xml') }],` and once finished I see the report
image

### DIFF
--- a/packages/playwright-test/src/reporters/junit.ts
+++ b/packages/playwright-test/src/reporters/junit.ts
@@ -96,7 +96,7 @@ class JUnitReporter extends EmptyReporter {
     let duration = 0;
     const children: XMLEntry[] = [];
 
-    for(const test of suite.allTests()){
+    for (const test of suite.allTests()){
       ++tests;
       if (test.outcome() === 'skipped')
         ++skipped;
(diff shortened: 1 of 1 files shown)


packages/playwright-test/src/reporters/junit.ts
```diff
@@ -189,13 +190,14 @@ class JUnitReporter extends EmptyReporter {
...
-          if (fs.existsSync(attachment.path))
-            systemOut.push(`\n[[ATTACHMENT|${attachmentPath}]]\n`);
-          else
-            systemErr.push(`\nWarning: attachment ${attachmentPath} is missing`);
-        } catch (e) {
+          await fs.promises.access(attachment.path);
+          systemOut.push(`\n[[ATTACHMENT|${attachmentPath}]]\n`);
+        } catch {
+          systemErr.push(`\nWarning: attachment ${attachmentPath} is missing`);
+          continue;
```
pavelfeldman, line 200: I don't think you need this `continue`.

pavelfeldman (approved):

packages/playwright-test/src/reporters/junit.ts
mastrzyz: agreed!

PR #24593, opened 2023-08-03
fix(blob): throw if blob report has newer version than the merger

yury-s:


packages/playwright-test/src/reporters/merge.ts
```diff
@@ -102,7 +102,11 @@ async function extractAndParseReports(dir: string, shardFiles: string[], printSt
 function findMetadata(events: JsonEvent[], file: string): BlobReportMetadata {
   if (events[0]?.method !== 'onBlobReportMetadata')
     throw new Error(`No metadata event found in ${file}`);
-  return events[0].params;
+  const metadata = (events[0].params as BlobReportMetadata);
+  console.log('metadata', metadata.version);
```
dgozman, line 106: Stray log?

dgozman (approved):

PR #24599, opened 2023-08-04
Include (optional) fullPage into TestConfig - expect 

autonomobil:


mxschmitt:
I recommend starting with filing an issue first before trying to implement a feature / fixing a bug, thanks!

PR #24615, opened 2023-08-04
fix: handle worker process sigint before initialization

dgozman:
When happening during a test, shows as a test error. When happening during stop, this error is ignored.

Drive-by: send SIGINT in tests to the whole tree, to better emulate Ctrl+C behavior.

### DIFF
--- a/packages/playwright-test/src/runner/dispatcher.ts
+++ b/packages/playwright-test/src/runner/dispatcher.ts
@@ -25,6 +25,7 @@ import { WorkerHost } from './workerHost';
 import type { TestGroup } from './testGroups';
 import type { FullConfigInternal } from '../common/config';
 import type { ReporterV2 } from '../reporters/reporterV2';
+import { serializeError } from '../util';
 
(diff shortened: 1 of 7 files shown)


packages/playwright-test/src/runner/dispatcher.ts
```diff
@@ -131,7 +132,15 @@ export class Dispatcher {
...
       this._workerSlots[index].worker = worker;
       worker.on('exit', () => this._workerSlots[index].worker = undefined);
-      await worker.start();
+      try {
+        await worker.start();
+      } catch (error) {
+        // When failed to start after stopping, most likely not-yet-intialized subprocess
+        // got SIGINT'ed. It is safe to ignore.
+        if (!this._isStopped)
+          this._markAllTestsInJobAsFailedOrSkipped(job, error);
```
pavelfeldman, line 141: It feels wrong, i.e. looks like it should be handled the same way as a crashed worker.

PR #24618, opened 2023-08-04
chore: intern test ids and abs paths

yury-s:
Created named classes for TeleTestStep and TeleTestResult which makes it easier to analyze in heap snapshot. Also changing `titlePath` from a closure to a method on TeleTestStep saved ~100Mb out of 2300Mb.

Intern test ids and absolute paths which showed up a lot of duplicate strings in heap snapshot.

Clear stepMap after processing onTestEnd.

### DIFF
--- a/packages/playwright-test/src/isomorphic/teleReceiver.ts
+++ b/packages/playwright-test/src/isomorphic/teleReceiver.ts
@@ -250,6 +250,7 @@ export class TeleReporterReceiver {
     result.error = result.errors?.[0];
     result.attachments = this._parseAttachments(payload.attachments);
     this._reporter.onTestEnd?.(test, result);
+    // Free up the memory as won't see these step ids.
     result.stepMap = new Map();
   }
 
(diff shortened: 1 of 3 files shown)


packages/playwright-test/src/isomorphic/teleReceiver.ts
```diff
@@ -248,26 +250,16 @@ export class TeleReporterReceiver {
     result.error = result.errors?.[0];
     result.attachments = this._parseAttachments(payload.attachments);
     this._reporter.onTestEnd?.(test, result);
+    result.stepMap = new Map();
```
dgozman, line 253: Let's comment about saving memory here.

packages/playwright-test/src/isomorphic/teleReceiver.ts
```diff
@@ -307,6 +299,7 @@ export class TeleReporterReceiver {
   }
 
   private _onExit(): Promise<void> | void {
+    this._stringPool = new StringInternPool();
```
dgozman, line 302: ditto

dgozman (approved):

PR #25017, opened 2023-08-05
fix(blob): replace projectSuffix with reportName 

dgozman:
Always ensure unique project/test ids across blobs.
Show `reportName` as a label in the html report.

References #24451.

### DIFF
--- a/.github/workflows/tests_secondary.yml
+++ b/.github/workflows/tests_secondary.yml
@@ -42,7 +42,7 @@ jobs:
     - run: npx playwright install --with-deps ${{ matrix.browser }} chromium
     - run: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- npm run test -- --project=${{ matrix.browser }}
       env:
-        PWTEST_BLOB_SUFFIX: "-${{ matrix.browser }}-${{ matrix.os }}"
+        PWTEST_BLOB_REPORT_NAME: "${{ matrix.browser }}-${{ matrix.os }}"
     - run: node tests/config/checkCoverage.js ${{ matrix.browser }}
     - run: ./utils/upload_flakiness_dashboard.sh ./test-results/report.json
       if: always()
@@ -75,7 +75,7 @@ jobs:
     - run: npx playwright install --with-deps ${{ matrix.browser }} chromium
     - run: npm run test -- --project=${{ matrix.browser }}
       env:
-        PWTEST_BLOB_SUFFIX: "-${{ matrix.browser }}-${{ matrix.os }}"
+        PWTEST_BLOB_REPORT_NAME: "${{ matrix.browser }}-${{ matrix.os }}"
     - run: ./utils/upload_flakiness_dashboard.sh ./test-results/report.json
       if: always()
       shell: bash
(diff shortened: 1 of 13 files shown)


.github/workflows/tests_secondary.yml
```diff
@@ -535,7 +535,7 @@ jobs:
       shell: bash
       env:
         PWTEST_CHANNEL: msedge-beta
-        PWTEST_BLOB_SUFFIX: "-edge-beta-windows"
+        PWTEST_BLOB_REPORT_NAME: "-edge-beta-windows"
```
yury-s, line 538: Let's drop leading dash from '-edge...' and in other places too. We can prepend it where needed in html report and in the blob file name but it's a bit odd to have report-name always start with `-`.

packages/playwright-test/src/reporters/merge.ts
```diff
@@ -114,25 +115,34 @@ async function mergeEvents(dir: string, shardReportFiles: string[], printStatus:
   const configureEvents: JsonEvent[] = [];
   const beginEvents: JsonEvent[] = [];
   const endEvents: JsonEvent[] = [];
-  const shardEvents = await extractAndParseReports(dir, shardReportFiles, printStatus);
-  shardEvents.sort((a, b) => {
```
yury-s, line 118: Let's keep it. We've changed between different naming schemes for files several times and there is a change we'll do it again in the future, so it's better if the merger does not depend on particular naming scheme.

packages/playwright-test/src/reporters/merge.ts
```diff
@@ -114,25 +115,34 @@ async function mergeEvents(dir: string, shardReportFiles: string[], printStatus:
...
+  const saltSet = new Set<string>();
+
   printStatus(`merging events`);
-  for (const { parsedEvents } of shardEvents) {
+
+  for (const { file, parsedEvents } of blobs) {
+    const metadataEvent = parsedEvents.find(e => e.method === 'onBlobReportMetadata')!;
+
+    // Generate unique salt for each blob.
+    // Prefer metadata.reportName, but fallback to reportFileName because it's unique.
```
yury-s, line 129: I guess this only makes sense in case of shards as shard number is not a part of reportName but is included in the file name. In this case maybe just add the shard id to the name?

packages/html-reporter/src/testFileView.tsx
```diff
@@ -93,33 +91,34 @@ const LabelsClickView: React.FC<React.PropsWithChildren<{
...
+        q = `${q} ${label}`.trim();
       else
-        q = q.split(' ').filter(t => t !== `@${tag}`).join(' ').trim();
-      // if metaKey or ctrlKey is not pressed, replace existing tags with new tag
+        q = q.split(' ').filter(t => t !== label).join(' ').trim();
     } else {
+      // if metaKey or ctrlKey is not pressed, replace existing tags with new tag
       if (!q.includes('@'))
-        q = `${q} @${tag}`.trim();
+        q = `${q} ${label}`.trim();
```
yury-s, line 110: Don't you need to prepend `@` to the labels?

yury-s (approved):

packages/playwright-test/src/reporters/merge.ts
dgozman: Since `reportName` is optional, we need something stable and unique to use for the salt.

PR #26322, opened 2023-08-07
feat: print message if maxFailures has reached

### LINKED ISSUE #24239: [BUG] Make it more clear when maxFailures got reached
As a user its often not directly when maxFailures got reached. We should print it to the console inside list/line reporter.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/24239

### DIFF
--- a/packages/playwright-test/src/reporters/base.ts
+++ b/packages/playwright-test/src/reporters/base.ts
@@ -55,6 +55,7 @@ export class BaseReporter implements ReporterV2 {
   private _omitFailures: boolean;
   private readonly _ttyWidthForTest: number;
   private _fatalErrors: TestError[] = [];
+  private _failureCount: number = 0;
 
   constructor(options: { omitFailures?: boolean } = {}) {
     this._omitFailures = options.omitFailures || false;
@@ -94,6 +95,8 @@ export class BaseReporter implements ReporterV2 {
   }
 
   onTestEnd(test: TestCase, result: TestResult) {
+    if (result.status !== 'skipped' && result.status !== test.expectedStatus)
+      ++this._failureCount;
     // Ignore any tests that are run in parallel.
     for (let suite: Suite | undefined = test.parent; suite; suite = suite.parent) {
       if ((suite as SuitePrivate)._parallelMode === 'parallel')
@@ -232,6 +235,7 @@ export class BaseReporter implements ReporterV2 {
     if (full && summary.failuresToPrint.length && !this._omitFailures)
       this._printFailures(summary.failuresToPrint);
     this._printSlowTests();
+    this._printMaxFailuresReached();
     this._printSummary(summaryMessage);
   }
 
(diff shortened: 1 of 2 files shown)


packages/playwright-test/src/runner/runner.ts
```diff
@@ -87,12 +87,13 @@ export class Runner {
     }
 
     const taskStatus = await taskRunner.run(testRun, deadline);
+    const maxFailuresReached = testRun.phases.some(p => p.dispatcher.hasReachedMaxFailures());
```
dgozman, line 90: Now that I see this line, there is a bug where we can have up to 99 failures in each dispatcher 😄  Let's fix it as well.

packages/playwright-test/src/runner/dispatcher.ts
```diff
@@ -512,7 +512,7 @@ export class Dispatcher {
     this._checkFinished();
   }
 
-  private _hasReachedMaxFailures() {
+  hasReachedMaxFailures(): boolean {
     const maxFailures = this._config.config.maxFailures;
     return maxFailures > 0 && this._failureCount >= maxFailures;
```
dgozman, line 517: Looking at this line, it seems straightforward to calculate `maxFailuresReached` in the reporter, instead of plumbing it. What do you think?

packages/playwright-test/src/runner/runner.ts
mxschmitt: Filed as https://github.com/microsoft/playwright/issues/26344.

packages/playwright-test/src/runner/dispatcher.ts
mxschmitt: we use it in various other places outside the reporter world.

packages/playwright-test/src/reporters/base.ts
```diff
@@ -99,6 +100,8 @@ export class BaseReporter implements ReporterV2 {
       if ((suite as SuitePrivate)._parallelMode === 'parallel')
         return;
     }
+    if (result.status !== 'skipped' && result.status !== test.expectedStatus)
```
dgozman, line 103: This should move before the return on line 101.

packages/playwright-test/src/reporters/base.ts
```diff
@@ -253,6 +257,12 @@ export class BaseReporter implements ReporterV2 {
       console.log(colors.yellow('  Consider splitting slow test files to speed up parallel execution'));
   }
 
+  private _printMaxFailuresReached(skipped: number) {
+    if (this._failureCount < this.config.maxFailures)
```
dgozman, line 261:
```suggestion
    if (this.config.maxFailures && this._failureCount < this.config.maxFailures)
```

packages/playwright-test/src/reporters/base.ts
```diff
@@ -253,6 +257,12 @@ export class BaseReporter implements ReporterV2 {
...
+      return;
+    console.log(colors.yellow(`  Maximum number of allowed failures (${this.config.maxFailures}) reached. ${skipped} remaining tests were skipped.`));
```
dgozman, line 263: This is not accurate, because `skipped` includes those tests that were marked as "skipped" as well. Let's not print this number :)

I'd recommend something like `Testing stopped early after 23 maximum allowed failures`.

dgozman (commented):
The tests are failing!

dgozman (approved):

PR #26329, opened 2023-08-07
chore(blob): split `onBegin` into multiple `onProject` events

dgozman:


packages/playwright-test/src/reporters/teleEmitter.ts
```diff
@@ -45,7 +45,9 @@ export class TeleReporterEmitter implements ReporterV2 {
 
   onBegin(suite: Suite) {
     const projects = suite.suites.map(projectSuite => this._serializeProject(projectSuite));
-    this._messageSink({ method: 'onBegin', params: { projects } });
+    for (const project of projects)
```
yury-s, line 48: Add a comment explaining why we split it into onProject events.

yury-s (approved):

PR #26333, opened 2023-08-07
chore(merge): read test events from one file at a time

yury-s:
Reference https://github.com/microsoft/playwright/issues/24451


packages/playwright-test/src/reporters/merge.ts
```diff
@@ -165,15 +199,27 @@ async function mergeEvents(dir: string, shardReportFiles: string[], printStatus:
...
-    mergeEndEvents(endEvents),
-    { method: 'onExit', params: undefined },
-  ];
+
+  return {
+    prologue: [
+      mergeConfigureEvents(configureEvents),
+      ...projectEvents,
+      { method: 'onBegin', params: undefined },
+      ...events,
```
dgozman, line 215: What else could be here? I'd expect `events.length` to be zero.

dgozman (approved):

packages/playwright-test/src/reporters/merge.ts
yury-s: Good point, removed.

PR #26337, opened 2023-08-08
chore(blob): store playwright user agent

yury-s:


packages/playwright-test/src/reporters/blob.ts
```diff
@@ -35,6 +35,7 @@ export const currentBlobReportVersion = 1;
 
 export type BlobReportMetadata = {
   version: number;
+  playwrightVersion: string;
```
dgozman, line 38: Might as well record the playwright user agent.

dgozman (approved):

PR #26352, opened 2023-08-08
fix: restore portable bash paths

rummik:
This PR restores the portable paths that were included in #23889 but were later broken by #24079

### DIFF
--- a/browser_patches/webkit/pw_run.sh
+++ b/browser_patches/webkit/pw_run.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 function runOSX() {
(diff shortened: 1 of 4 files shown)


rummik:

aslushnikov:
@rummik yep, we upstreamed these changes to our internal repository, so no need to update paths in this mirror.

rummik:
@aslushnikov So the next time the browser patches are updated the paths for these four files will be fixed automatically?

aslushnikov:
@rummik yes, exactly

aslushnikov:
@rummik actually, will happen on the next roll of the browser patches - when we will be releasing 1.38.

PR #26382, opened 2023-08-09
docs(merge): use azcopy for uploads

yury-s:


docs/src/test-sharding-js.md
```diff
@@ -135,7 +146,12 @@ We can utilize Azure Storage's static websites hosting capabilities to easily an
           shell: bash
           run: |
             REPORT_DIR='run-${{ github.run_id }}-${{ github.run_attempt }}'
-            az storage blob upload-batch -s playwright-report -d "\$web/$REPORT_DIR" --connection-string "${{ secrets.AZURE_CONNECTION_STRING }}"
+            azcopy cp --recursive "./playwright-report/*" "https://<STORAGE_ACCOUNT_NAME>.blob.core.windows.net/\$web/$REPORT_DIR"
```
mxschmitt, line 149: the backslash of $ is weird, but I guess you have reasons for it.

mxschmitt (approved):

docs/src/test-sharding-js.md
yury-s: Yeah, the reason is otherwise bash will substitute `$web`  with the env variable `web` value which is empty in this case. Since we want to expand `$REPORT_DIR` in the URL, we cannot say azcopy cp --recursive '... /$web/$REPORT_DIR' either.

PR #26407, opened 2023-08-10
fix: download of attachments in UI Mode

### LINKED ISSUE #26326: [BUG] Can't download attachments from UI mode
Related to https://github.com/microsoft/playwright/issues/22736

### System info
- Playwright Version: [v1.36.2]
- Operating System: [Windows 11, macOS 13.5]
- Browser: [Chromium]

### Source code

**Config file**

```js
// playwright.config.ts

export default defineConfig({
  projects: [
    {
      name: 'chromium',
      use: { ...devices['Desktop Chrome'], },
    },
});
```

```js

test('download file', async ({ page }, testInfo) => {
  const payload = "data:application/zip;base64,UEsDBAoAAAAAAJCNB1eCidH3BQAAAAUAAAAJAAAASGVsbG8udHh0SGVsbG9QSwECPwMKAAAAAACQjQdXgonR9wUAAAAFAAAACQAkAAAAAAAAACCApIEAAAAASGVsbG8udHh0CgAgAAAAAAABABgAm6rbDUbJ2QEAAAAAAAAAAAAAAAAAAAAAUEsFBgAAAAABAAEAWwAAACwAAAAAAA==";
  await page.setContent(`<a href="${payload}" download="myFile.zip">My Link</a>`);
  const downloadPromise = page.waitForEvent('download');
  await page.locator('a').click();
  const download = await downloadPromise;
  const path = download.suggestedFilename();
  await download.saveAs(path);
  if (path)
    testInfo.attachments.push({ name: 'myFile.zip', path: path, contentType: 'application/zip' });
  expect(path).toBeTruthy();
});
```

**Steps**
- `npx playwright test --ui`
- Run the test
- See the attached file and try to actually view it / download it / open it

image

**Expected**

Can download the attachment and check what it contains, no matter the mode.
Also expected would be to get the right filename (the one specified in `attachments.push`) which probably means the server needs to set proper Content-Disposition headers (?).

**Actual**

- ❌ can't be downloaded from UI mode
- (✅) can be downloaded from HTML mode but thas a random filename instead of the expected name
- ❌ can't be downloaded from VS Code extension

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/26326.

### DIFF
--- a/packages/playwright-core/src/server/chromium/crBrowser.ts
+++ b/packages/playwright-core/src/server/chromium/crBrowser.ts
@@ -343,7 +343,7 @@ export class CRBrowserContext extends BrowserContext {
   override async _initialize() {
     assert(!Array.from(this._browser._crPages.values()).some(page => page._browserContext === this));
     const promises: Promise<any>[] = [super._initialize()];
-    if (this._browser.options.name !== 'electron' && this._browser.options.name !== 'clank' && !this._options.keepDefaultDownloadBehavior) {
+    if (this._browser.options.name !== 'electron' && this._browser.options.name !== 'clank' && this._options.acceptDownloads !== 'internal-browser-default') {
       promises.push(this._browser._session.send('Browser.setDownloadBehavior', {
         behavior: this._options.acceptDownloads ? 'allowAndName' : 'deny',
         browserContextId: this._browserContextId,
--- a/packages/playwright-core/src/server/chromium/crBrowser.ts
+++ b/packages/playwright-core/src/server/chromium/crBrowser.ts
@@ -343,7 +343,7 @@ export class CRBrowserContext extends BrowserContext {
   override async _initialize() {
     assert(!Array.from(this._browser._crPages.values()).some(page => page._browserContext === this));
     const promises: Promise<any>[] = [super._initialize()];
-    if (this._browser.options.name !== 'electron' && this._browser.options.name !== 'clank') {
+    if (this._browser.options.name !== 'electron' && this._browser.options.name !== 'clank' && !this._options.keepDefaultDownloadBehavior) {
       promises.push(this._browser._session.send('Browser.setDownloadBehavior', {
         behavior: this._options.acceptDownloads ? 'allowAndName' : 'deny',
         browserContextId: this._browserContextId,
--- a/packages/playwright-core/src/server/chromium/crBrowser.ts
+++ b/packages/playwright-core/src/server/chromium/crBrowser.ts
@@ -345,7 +345,7 @@ export class CRBrowserContext extends BrowserContext {
     const promises: Promise<any>[] = [super._initialize()];
     if (this._browser.options.name !== 'electron' && this._browser.options.name !== 'clank' && this._options.acceptDownloads !== 'internal-browser-default') {
       promises.push(this._browser._session.send('Browser.setDownloadBehavior', {
-        behavior: this._options.acceptDownloads ? 'allowAndName' : 'deny',
+        behavior: this._options.acceptDownloads === 'accept' ? 'allowAndName' : 'deny',
         browserContextId: this._browserContextId,
         downloadPath: this._browser.options.downloadsPath,
         eventsEnabled: true,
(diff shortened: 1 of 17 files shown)


packages/playwright-core/src/server/trace/viewer/traceViewer.ts
```diff
@@ -146,6 +146,7 @@ export async function openTraceViewerApp(traceUrls: string[], browserName: strin
     ignoreDefaultArgs: ['--enable-automation'],
     colorScheme: 'no-override',
     useWebSocket: isUnderTest(),
+    keepDefaultDownloadBehavior: isUnderTest() ? false : true,
```
dgozman, line 149: Perhaps it's time to introduce `launchApp()` internal method that would also replace `installAppIcon()` and unify the usages? Because I assume all other apps we launch would also need this download behavior.

packages/playwright-core/src/server/chromium/crBrowser.ts
```diff
@@ -343,7 +343,7 @@ export class CRBrowserContext extends BrowserContext {
   override async _initialize() {
     assert(!Array.from(this._browser._crPages.values()).some(page => page._browserContext === this));
     const promises: Promise<any>[] = [super._initialize()];
-    if (this._browser.options.name !== 'electron' && this._browser.options.name !== 'clank') {
+    if (this._browser.options.name !== 'electron' && this._browser.options.name !== 'clank' && !this._options.keepDefaultDownloadBehavior) {
```
dgozman, line 346: - What about other browsers? This option should work in all of them.
- Perhaps it makes sense to replace `options.acceptDownloads` with an enum `downloads: 'accept' | 'deny' | 'internal-browser-default'`?

packages/trace-viewer/src/sw.ts
```diff
@@ -145,6 +146,49 @@ async function doFetch(event: FetchEvent): Promise<Response> {
...
+  const nameHasExt = name.includes('.');
+  if (!nameHasExt)
+    name += extensionForContentType(attachment.contentType);
+  headers.set('Content-Disposition', `attachment; filename="${name}"`);
+  if (attachment.contentType)
+    headers.set('Content-Type', attachment.contentType);
+  return headers;
+}
+
+function extensionForContentType(contentType: string): string {
```
dgozman, line 164: Don't we have an utility like this already?

packages/trace-viewer/src/sw.ts
```diff
@@ -122,9 +122,10 @@ async function doFetch(event: FetchEvent): Promise<Response> {
         // We will accept explicit ?trace= value as well as the clientId associated with the trace.
         if (traceUrl !== trace && !traceUrls.includes(trace))
           continue;
-        const blob = await traceModel!.resourceForSha1(relativePath.slice('/sha1/'.length));
+        const sha1 = relativePath.slice('/sha1/'.length);
```
dgozman, line 125: Let's extract this into `serveResource(traceModel, sha1)` method?

packages/trace-viewer/src/sw.ts
```diff
@@ -145,6 +146,49 @@ async function doFetch(event: FetchEvent): Promise<Response> {
...
 
+function headersForResource(traceModel: TraceModel, sha1: string): Headers | undefined {
+  const attachment = traceModel.attachmentForSha1(sha1);
+  if (!attachment)
+    return;
+  const headers = new Headers();
+  let name = attachment.name;
+  const nameHasExt = name.includes('.');
+  if (!nameHasExt)
+    name += extensionForContentType(attachment.contentType);
```
dgozman, line 157: Was this explicitly requested? I'd leave `attachment.name` without adding anything.

packages/playwright-core/src/server/trace/viewer/traceViewer.ts
mxschmitt: https://github.com/microsoft/playwright/pull/26499

packages/trace-viewer/src/sw.ts
mxschmitt: did not find any.

yury-s: mime.getExtension() should do what you need

packages/trace-viewer/src/sw.ts
yury-s: @mxschmitt add a comment explaining why we add it here?

yury-s (approved):

mxschmitt:
Note: I left the content-type to file extension out for now, since its rare that customers look at the screenshot/video attachments inside UI Mode (these are just named: `screenshot` and `video`). Trace won't be present there, since we use it internally for UI Mode.

PR #26421, opened 2023-08-10
chore: make html report produce named attachments

mxschmitt:
https://github.com/microsoft/playwright/issues/26326

### DIFF
--- a/packages/html-reporter/src/links.tsx
+++ b/packages/html-reporter/src/links.tsx
@@ -68,12 +68,22 @@ export const AttachmentLink: React.FunctionComponent<{
 }> = ({ attachment, href, linkName }) => {
   return <TreeItem title={<span>
     {attachment.contentType === kMissingContentType ? icons.warning() : icons.attachment()}
-    {attachment.path && <a href={href || attachment.path} target='_blank'>{linkName || attachment.name}</a>}
+    {attachment.path && <a href={href || attachment.path} download={downloadFileNameForAttachment(attachment)}>{linkName || attachment.name}</a>}
     {attachment.body && <span>{attachment.name}</span>}
   </span>} loadChildren={attachment.body ? () => {
     return [<div className='attachment-body'>{attachment.body}</div>];
   } : undefined} depth={0} style={{ lineHeight: '32px' }}></TreeItem>;
 };
+
+function downloadFileNameForAttachment(attachment: TestAttachment): string {
+  if (attachment.name.includes('.'))
+    return attachment.name;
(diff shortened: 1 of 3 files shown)


packages/html-reporter/src/links.tsx
```diff
@@ -68,13 +68,23 @@ export const AttachmentLink: React.FunctionComponent<{
...
     {attachment.body && <span>{attachment.name}</span>}
   </span>} loadChildren={attachment.body ? () => {
     return [<div className='attachment-body'>{attachment.body}</div>];
   } : undefined} depth={0} style={{ lineHeight: '32px' }}></TreeItem>;
 };
 
+function downloadFileNameForAttachment(attachment: TestAttachment): string {
+  if (attachment.name.includes('.'))
+    return attachment.name;
+  const path = attachment.path!;
```
dgozman, line 81: I think there could be no path, and if there is one it would be sha1. I'd always use `attachment.name` - that's what user sees. We do not make any promises about keeping the original attachment path anywhere.

mxschmitt: Then we produce a "bad filename" which folks can't open for our default trace, screenshot and video which are by default "trace", "screenshot", and "video". Thats why I decided to go this route.

packages/html-reporter/src/links.tsx
```diff
@@ -68,13 +68,22 @@ export const AttachmentLink: React.FunctionComponent<{
...
     {attachment.body && <span>{attachment.name}</span>}
   </span>} loadChildren={attachment.body ? () => {
     return [<div className='attachment-body'>{attachment.body}</div>];
   } : undefined} depth={0} style={{ lineHeight: '32px' }}></TreeItem>;
 };
 
+function downloadFileNameForAttachment(attachment: TestAttachment): string {
+  if (attachment.name.includes('.') || !attachment.path)
+    return attachment.name;
+  const firstDotIndex = attachment.path.indexOf('.');
```
yury-s, line 81: nit: use `path.extname` and check for emptiness instead?

yury-s, line 81: I believe it should be lastDotIndex.

yury-s (approved):

packages/html-reporter/src/links.tsx
mxschmitt: We are in the browser land.

mxschmitt: clarified offline, to keep it to support e.g. tar.gz files.

PR #26425, opened 2023-08-10
docs: examples that save download with suggested filename

### LINKED ISSUE #26341: [Feature] Playwright should save file with name and extension while downloading file
As of now when we automate download functionality scenario, then file is getting save with some GUID number without any extension. So as a result if we wanted to verify the content of downloaded file we can't do it . 

We have also tried implementing steps giving under  url https://playwright.dev/docs/downloads recommended by many post , but with no success 

There are many open issue for this bug few on them I am listing #24048 #18139. 

Adding this feature will be beneficial for download functionality scenario and verifying its content.

Let me know if you needed more input

dgozman:
Fixes #26341.

### DIFF
--- a/docs/src/api/class-download.md
+++ b/docs/src/api/class-download.md
@@ -6,7 +6,7 @@
 All the downloaded files belonging to the browser context are deleted when the
 browser context is closed.
 
-Download event is emitted once the download starts. Download path becomes available once download completes:
+Download event is emitted once the download starts. Download path becomes available once download completes.
 
 ```js
 // Start waiting for download before clicking. Note no await.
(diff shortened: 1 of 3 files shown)


docs/src/api/class-download.md
```diff
@@ -6,49 +6,59 @@
...
 Download download = page.waitForDownload(() -> {
-  page.getByText("Download file").click();
+    // Perform the action that initiates download
+    page.getByText("Download file").click();
 });
-// wait for download to complete
-Path path = download.path();
+
+// Wait for the download process to complete and save the downloaded file somewhere
+download.saveAs(Paths.get("/path/to/save/at/", download.suggestedFilename()));
```
yury-s, line 29:
```suggestion
download.saveAs(Paths.get("/path/to/save/at/" + download.suggestedFilename()));
```

yury-s (approved):

PR #26437, opened 2023-08-11
fix(junit reporter): Properly report actual errors in the junit failure message property

### LINKED ISSUE #12944: [BUG] Incorrect serialization failure message in Junit reporter
**Context:**
- Playwright Version: [1.20.0]
- Operating System: [Windows 10]
- Node.js version: [v14.17.5]
- Browser: [All]

**Code Snippet**

Hello everyone 🎭✋🏻! I found incorrect junit reporter serialization work;
when the test fails I expect the "failure message" field in the xml file to contain the error text, but it contains the test header:

```xml
<failure message="example.spec.ts:5:3 Check equal" type="FAILURE">
```

test code:

```typescript
test.describe('Example', () => {

  test('Check equal', async () => {
    throw new Error('Error message');
  });
});
```

how it looks in json reporter:
```json
"error": {
  "message": "Error message",
  "stack": "Error: Error message\n    at C:\\Users\\username\\Desktop\\issue\\new-project\\tests\\example.spec.ts:6:11"
},
```

I prepared a [repository ](https://github.com/Gornostalev/playwright_issue) with npm scripts:

```npm run testJson``` - for run with json reporter
```npm run testJunit``` - for run with junit reporter

I hope this makes your job easier, thanks for Playwright!❤

### PR DESCRIPTION
**Pull request by caaldrid**
Currently junit test failure are reported with the failing test file's name, failure location, and test name in the message property, while the actual failure information is stored in the CDATA text. This can break tools that look at the error message to do failure analysis and deduplication as they will not have the actual failure cause.

This change makes it so we report all unique errors hit by the tests in the xml failure entry's message property.

Fixes: #12944

### CODE REVIEW

File: packages/playwright-test/src/reporters/junit.ts [hunk 1]
```diff
@@ -175,7 +188,7 @@ class JUnitReporter extends EmptyReporter {
       entry.children.push({
         name: 'failure',
         attributes: {
-          message: `${path.basename(test.location.file)}:${test.location.line}:${test.location.column} ${test.title}`,
+          message: this._extractErrorMessage(test.results),
```
yury-s, line 191: This way the message does not contain `file:line` information anymore which will likely break more tools than it will fix?

caaldrid: I don't really see what value of including the file given that you already know what test file and the test name are in the `testcase` parent node of the result entry. I would even argue that tools should be using that instead of relying on the file to be in the error message. 
image

For the line and column though, I would agree it's still useful to have in the error message and can add that back. Maybe enhance it with the actual function that threw?

yury-s: Re: test name, if we remove the test location from the error, I believe it would make sense to add it to the `classname` in `<testcase classname="a.test.js:10:13">` instead. This will aid tools in navigating to the source, particularly when the test name is templated, such as "test(`test ${i}`...". In cases like these, a specific substring like 'test 5' may not exist in the file. What do you think?

> For the line and column though, I would agree it's still useful to have in the error message and can add that back. Maybe enhance it with the actual function that threw?

Yes, I think having text of the line that threw along with its location would be ideal, i.e. instead of the test location it would be actual error location. How about for the error that is currently formatted like:

```xml
<testcase name="one" classname="a.test.js" time="0.133">
<failure message="expect(received).toBe(expected) // Object.is equality  Expected: 0 Received: 1
New failure on retry 2" type="FAILURE">
<![CDATA[  a.test.js:3:15 › one ─────────────────────────────────────────────────────────────────────────────

    Expected: 0
    Received: 1

      5 |               throw new Error('New failure on retry 2');
      6 |             }
    > 7 |             expect(1).toBe(0);
        |                       ^
      8 |           });
      9 |         

        at fn (/Users/yurys/playwright/test-results/reporter-junit-created-should-render-all-unique-errors-after-retry-playwright-test/a.test.js:7:23)
...
```

We change it to the following?

```xml
<testcase name="one" classname="a.test.js:3:15" time="0.133">
<failure message="a.test.js:7:23 expect(received).toBe(expected) // Object.is equality  Expected: 0 Received: 1
New failure on retry 2" type="FAILURE">
<![CDATA[  a.test.js:3:15 › one ─────────────────────────────────────────────────────────────────────────────

    Expected: 0
    Received: 1

      5 |               throw new Error('New failure on retry 2');
      6 |             }
    > 7 |             expect(1).toBe(0);
        |                       ^
      8 |           });
      9 |         

        at fn (/Users/yurys/playwright/test-results/reporter-junit-created-should-render-all-unique-errors-after-retry-playwright-test/a.test.js:7:23)

```

Or alternatively add the location to the end of the message, e.g. `message="a.test.js:7:23 expect(received).toBe(expected) // Object.is equality  Expected: 0 Received: 1
New failure on retry 2 at a.test.js:7:23"`

pavelfeldman:
Closing as stale

PR #26473, opened 2023-08-14
docs(merge): add screenshot, print url, fix name

### LINKED ISSUE #24451: [internal] merge report action items

yury-s:
* Renamed test job to `playwright-tests` to match docs and rererences
* Added screenshot with github jobs and artifacts
* Print Azure URL

Fixes: https://github.com/microsoft/playwright/issues/24451

### DIFF
--- a/docs/src/test-sharding-js.md
+++ b/docs/src/test-sharding-js.md
@@ -55,7 +55,7 @@ on:
   pull_request:
     branches: [ main, master ]
 jobs:
-  test:
+  playwright-tests:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


docs/src/test-sharding-js.md
```diff
@@ -55,7 +55,7 @@ on:
   pull_request:
     branches: [ main, master ]
 jobs:
-  test:
+  playwright-tests:
```
mxschmitt, line 58: can we change the references instead and keep test here?

Reason: Have it as much aligned as [this](https://github.com/microsoft/create-playwright/blob/main/assets/github-actions.yml) as possible, since thats what the users most likely have as a foundation.

mxschmitt (approved):

docs/src/test-sharding-js.md
yury-s: That was renamed to `playwright-tests` to disambiguate references.

PR #26499, opened 2023-08-16
chore: move launchApp into separate function

mxschmitt:
https://github.com/microsoft/playwright/pull/26407#discussion_r1290727547

### DIFF
--- a/packages/playwright-core/src/server/browserType.ts
+++ b/packages/playwright-core/src/server/browserType.ts
@@ -20,7 +20,7 @@ import path from 'path';
 import type { BrowserContext } from './browserContext';
 import { normalizeProxySettings, validateBrowserContextOptions } from './browserContext';
 import type { BrowserName } from './registry';
-import { registry } from './registry';
+import { findChromiumChannel, registry } from './registry';
 import type { ConnectionTransport } from './transport';
 import { WebSocketTransport } from './transport';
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/server/browserType.ts
```diff
@@ -84,6 +85,37 @@ export abstract class BrowserType extends SdkObject {
     return browser._defaultContext!;
   }
 
+  async launchApp(options: {
```
yury-s, line 88: Should it be inside crApp.ts instead?

packages/playwright-core/src/server/browserType.ts
```diff
@@ -84,6 +85,37 @@ export abstract class BrowserType extends SdkObject {
...
+    sdkLanguage: string,
+    windowSize: types.Size,
+    windowPosition?: types.Point,
+    persistentContextOptions?: Parameters<BrowserType['launchPersistentContext']>[2];
+  }) {
+    const args = [...options.persistentContextOptions?.args ?? []];
+
+    if (this._name === 'chromium') {
```
yury-s, line 96: Do we upport non-chromium as recorder host?

yury-s (approved):

packages/playwright-core/src/server/browserType.ts
mxschmitt: up we do.

PR #26508, opened 2023-08-16
docs: blob reporter options

### LINKED ISSUE #26481: Blob report custom output directory
With the latest addition of the report merger I was trying to switch to blob reports and then merge those.

Unfortunately we need to store those reports in a custom directory.

Is it possible to configure the output directory for blob reports like HTML reports?

yury-s:
Fixes #26481

### DIFF
--- a/docs/src/test-reporters-js.md
+++ b/docs/src/test-reporters-js.md
@@ -205,8 +205,23 @@ Or if there is a custom folder name:
 npx playwright show-report my-report
 ```
 
-> The `html` reporter currently does not support merging reports generated across multiple [`--shards`](./test-parallel.md#shard-tests-between-multiple-machines) into a single report. See [this](https://github.com/microsoft/playwright/issues/10437) issue for available third party solutions.
+### Blob reporter
 
+Blob reports contain all the details about the test run and can be used later to produce any other report. Their primary function is to facilitate the merging of reports from [sharded tests](./test-sharding.md).
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/src/test-reporters-js.md
```diff
@@ -205,8 +205,23 @@ Or if there is a custom folder name:
...
-> The `html` reporter currently does not support merging reports generated across multiple [`--shards`](./test-parallel.md#shard-tests-between-multiple-machines) into a single report. See [this](https://github.com/microsoft/playwright/issues/10437) issue for available third party solutions.
+### Blob reporter
 
+Blob reports contain all the details about the test run and can be used later to produce any other report. Their primary function is to facilitate the merging of reports from [sharded tests](./test-sharding.md).
+
+```bash
+npx playwright test --reporter=blob
+```
+
+By default, the report is written into the `blob-report` directory in the package.json parent directory or current working directory (if no package.json is found). The output directory can be overridden in the configuration file:
```
mxschmitt, line 216:
```suggestion
By default, the report is written into the `blob-report` directory in the package.json directory or current working directory (if no package.json is found). The output directory can be overridden in the configuration file:
```

Was the parent here appropriate logic wise?

mxschmitt (approved):

docs/src/test-reporters-js.md
yury-s: I think it's ok without 'parent', the explanation is already quite verbose.

PR #26516, opened 2023-08-17
devops: hide previous Test results comments

mxschmitt:
There is no REST call for it, so we need to use GraphQL.

References:
- https://github.com/actions/github-script#run-custom-graphql-queries
- https://docs.github.com/en/graphql/reference/mutations#minimizecomment
- https://github.com/isaacs/github/issues/1480

### DIFF
--- a/.github/workflows/create_test_report.yml
+++ b/.github/workflows/create_test_report.yml
@@ -73,6 +73,24 @@ jobs:
             core.error('No pull request found for commit ' + context.sha + ' and workflow triggered by: ' + context.payload.workflow_run.event);
             return;
           }
+          {
+            // Mark previous comments as outdated by minimizing them.
+            const { data: comments } = await github.rest.issues.listComments({
+              ...context.repo,
+              issue_number: prNumber,
+            });
+            for (const comment of comments) {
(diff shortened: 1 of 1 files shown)


.github/workflows/create_test_report.yml
```diff
@@ -73,6 +73,24 @@ jobs:
...
+            // Mark previous comments as outdated by minimizing them.
+            const { data: comments } = await github.rest.issues.listComments({
+              ...context.repo,
+              issue_number: prNumber,
+            });
+            for (const comment of comments) {
+              if (comment.user.login === 'github-actions[bot]' && comment.body.includes('Test results')) {
+                await github.graphql(`
+                  mutation {
+                    minimizeComment(input: {subjectId: "${comment.node_id}", classifier: OUTDATED}) {
```
yury-s, line 86: should it be quoted 'OUTDATED' ?

yury-s (approved):

.github/workflows/create_test_report.yml
mxschmitt: according to some examples, no.

PR #26526, opened 2023-08-17
feat: allow absolute ctTemplateDir

tmkx:
When passing an absolute directory to `ctTemplateDir`, Playwright doesn't work as expected.


packages/playwright-ct-core/src/vitePlugin.ts
```diff
@@ -62,7 +62,8 @@ export function createPlugin(
       const use = config.projects[0].use as CtConfig;
       const port = use.ctPort || 3100;
       const viteConfig = typeof use.ctViteConfig === 'function' ? await use.ctViteConfig() : (use.ctViteConfig || {});
-      const relativeTemplateDir = use.ctTemplateDir || 'playwright';
+      const templateDirConfig = use.ctTemplateDir || 'playwright';
```
pavelfeldman, line 65: We should not use process.cwd in build and test systems. If you look at the line 69, you'll see that instead of join you probably want to call resole.

tmkx:

pavelfeldman (approved):

PR #26552, opened 2023-08-18
chore: migrate minimatch -> micromatch

### LINKED ISSUE #24122: [Feature] Use `micromatch` or better for faster file matching
This isn't really a feature, but I wasn't sure I should file this as a bug.

Currently, playwright depends on a very old version of `minimatch` (v3) for file matching. 

The file matching operation is done a few times during playwright startup. 
`micromatch` [performs significantly better](https://www.npmjs.com/package/micromatch#running-benchmarks) in glob matching, and this can have a small but noticeable impact on startup times for large projects that use playwright.

I was about to send across a PR for this when I noticed that playwright [uses some API bits](https://github.com/microsoft/playwright/blob/b4ffb848de1b00e9a0abad6dacdccce60cce9bed/packages/playwright-test/src/runner/projectUtils.ts#L171-L192) from `minimatch` that aren't compatible with `micromatch`. e.g. Parsing comments in patterns.

Overall, this might _technically_ be a breaking change since there are a few esoteric features like comments in globs that `micromatch` doesn't support (does anyone really use them?), but it is otherwise largely a self-stated replacement for minimatch.
A 3-10x performance jump might justify such a change.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/24122


packages/playwright-test/src/runner/projectUtils.ts
```diff
@@ -181,17 +181,24 @@ async function collectFiles(testDir: string, respectGitIgnore: boolean): Promise
...
+          if (!line || line[0] === '#')
             return;
-          // Use flipNegate, because we handle negation ourselves.
-          const rule = new minimatch.Minimatch(s, { matchBase: true, dot: true, flipNegate: true }) as any;
-          if (rule.comment)
-            return;
-          rule.dir = dir;
-          return rule;
-        }).filter(rule => !!rule);
+          const negate = line[0] === '!' || line[0] === '^';
```
pavelfeldman, line 188: What is ^?

packages/playwright-test/src/runner/projectUtils.ts
```diff
@@ -181,17 +181,24 @@ async function collectFiles(testDir: string, respectGitIgnore: boolean): Promise
...
+          // Remove the negation from the rule, because we handle negation ourselves.
+          if (negate)
+            line = line.substring(1);
+          return {
+            dir,
+            negate,
+            match: (s: string, partial?: boolean) => {
+              if (partial)
```
pavelfeldman, line 196: I don't think we should ever need partial mode for the gitignore, should we use this as an opportunity to reimplement the filtering logic? It seems like it was created by layers with each layer adding into the logic.

PR #26555, opened 2023-08-19
chore: format console message from page

pavelfeldman:


packages/trace-viewer/src/ui/consoleTab.tsx
```diff
@@ -139,3 +148,75 @@ function iconClass(message: channels.ConsoleMessageInitializer): string {
...
+      styleObject['color'] = 'var(--vscode-debugTokenExpression-number)';
+    tokens.push(<span style={styleObject}>{value?.preview || ''}</span>);
+  }
+  return formatted;
+}
+
+function parseCSSStyle(cssFormat: string): Record<string, string | number> {
+  try {
+    const styleObject: Record<string, string | number> = {};
+    const cssText = cssFormat.replace(/;$/, '').replace(/: /g, ':').replace(/; /g, ';');
```
yury-s, line 203: Can we just trim the values after splitting?

yury-s (approved):

packages/trace/src/trace.ts
```diff
@@ -101,10 +101,19 @@ export type EventTraceEvent = {
   pageId?: string;
 };
 
-export type ObjectTraceEvent = {
+export type ConsoleMessageTraceEvent = {
   type: 'object';
```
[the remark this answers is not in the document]
pavelfeldman: I'm not changing the actual file format, so it should stay.

packages/trace-viewer/src/ui/consoleTab.tsx
pavelfeldman: I'll follow up

PR #26559, opened 2023-08-19
chore: remove webpack reference in readme/comments.

mastrzyz:
Removed the reference to webpack since it is no longer up to date reference.


pavelfeldman:
We bundle changes like this with meaningful changes to code to spare the bots.

PR #26562, opened 2023-08-19
chore: make console stream live in ui mode

pavelfeldman:


tests/playwright-test/ui-mode-test-output.spec.ts
```diff
@@ -151,3 +151,28 @@ test('should format console messages in page', async ({ runUITest }, testInfo) =
...
+  const { page } = await runUITest({
+    'a.spec.ts': `
+      import { test, expect } from '@playwright/test';
+      test('print', async ({ page }) => {
+        await page.setContent('<button>Click me</button>');
+        const button = page.getByRole('button', { name: 'Click me' });
+        await button.evaluate(node => node.addEventListener('click', () => {
+          console.log('I was clicked');
+        }));
+        await button.click();
```
mxschmitt, line 165:
```suggestion
        setTimeout(() => console.log('from node'), 100)
        await button.click();
```

Shall we add a test for that as well?

pavelfeldman: Yes, good catch

mxschmitt (approved):

PR #26564, opened 2023-08-20
chore: use SVG icons for web apps

mxschmitt:
- SVG favicons are supported in Firefox + Chrome (we could add fallback for Safari via a small .ico file)


packages/trace-viewer/src/ui/workbenchLoader.css
```diff
@@ -100,6 +100,13 @@ body.dark-mode .drop-target {
 .workbench-loader .logo {
   font-size: 20px;
   margin-left: 16px;
+  display: flex;
```
pavelfeldman, line 103: I don't think you need font-size anymore. Also, do you need to align the image in flex?

packages/trace-viewer/src/ui/workbenchLoader.css
```diff
@@ -100,6 +100,13 @@ body.dark-mode .drop-target {
...
+}
+
+.workbench-loader .logo img {
+  height: 32px;
+  align-self: center;
+  pointer-events: none;
```
pavelfeldman, line 109: why no pointer events? also, why no width?

packages/trace-viewer/src/ui/workbenchLoader.css
```diff
@@ -100,6 +100,13 @@ body.dark-mode .drop-target {
 .workbench-loader .logo {
   font-size: 20px;
   margin-left: 16px;
+  display: flex;
+}
+
+.workbench-loader .logo img {
+  height: 32px;
+  align-self: center;
```
pavelfeldman, line 108: prefer align-items in container to overriding the alignment

pavelfeldman (approved):

packages/trace-viewer/src/ui/workbenchLoader.css
mxschmitt: no pointer events for cosmetic UX reasons, since then immediately the trace upload would get triggered.

packages/trace-viewer/src/ui/workbenchLoader.tsx
```diff
@@ -139,7 +139,9 @@ export const WorkbenchLoader: React.FunctionComponent<{
 
   return <div className='vbox workbench-loader' onDragOver={event => { event.preventDefault(); setDragOver(true); }}>
     <div className='hbox header'>
-      <div className='logo'>🎭</div>
+      <div className='logo'>
+        <img src="playwright-logo.svg" />
```
mxschmitt, line 143:
```suggestion
        <img src='playwright-logo.svg' />
```

PR #26569, opened 2023-08-21
docs(api): remove confusing async/await in generic assertions example

kevcenteno:
The use of async/await is confusing, as it implies that `toBe()` is an async assertion.

### DIFF
--- a/docs/src/api/class-genericassertions.md
+++ b/docs/src/api/class-genericassertions.md
@@ -9,7 +9,7 @@ import { test, expect } from '@playwright/test';
 
 test('assert a value', async ({ page }) => {
   const value = 1;
-  await expect(value).toBe(2);
+  expect(value).toBe(2);
 });
 ```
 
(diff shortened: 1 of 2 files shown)


kevcenteno:

docs/src/api/class-genericassertions.md
```diff
@@ -7,9 +7,9 @@ The [GenericAssertions] class provides assertion methods that can be used to mak
 ```js
 
-test('assert a value', async ({ page }) => {
+test('assert a value', ({ page }) => {
```
mxschmitt, line 10:
```suggestion
test('assert a value', async ({ page }) => {
```

I'd keep that, since our customers have always async tests.

mxschmitt (approved):

docs/src/api/class-genericassertions.md
kevcenteno: I think adding `async` to this exact example would cause a linting issue on customer setups.

pavelfeldman: The test should be async, I don't think lint should care about async functions not using await.

kevcenteno: Okay.  Force pushed

Issue #24603, opened 2023-08-04
[BUG] Typings of `Frame.page()` do not match the actual behaviour in some cases


Llois41:
### System info
- Playwright Version: [v1.35.1]
- Operating System: [All]
- Browser: [All]
- Other info:

### Source code

Our Playwright setup is big and complicated and I will try to explain the issue first as best as I can. I am willing to make a reproduction repository later on if you can't derive the issue from my description.

- [ ] I provided exact source code that allows reproducing the issue locally.

**Expected**

The typings match the actual behaviour of the code.

**Actual**

We have listeners on the network requests installed like this `this.browserContext.on('request', this.addRequest)`.
What we're now experiencing is that if we open a new tab (= Page in Playwright's terminology) by using the `Locator.click()`-method (`await element.click({ button: 'middle' })`, in this case `element` is the Locator of an anchor tag) the navigation request is reqistered by the event listener and passed to our `addRequest()` handler, but if we try to access the Frame's Page by calling `const page = request.frame().page()`, `page` will be `undefined`. Waiting a little (~50ms) and make the exact same call again (`request.frame().page()`) will now deliver a Page on the Frame.

I hope the description emphasizes that at least the typings for `Frame.page()` are wrong, since it does in fact return `undefined` in some cases or there is a bug in the method's implementation or in the creating process of a Request/Frame.
I took a look into the code and in `Frame.page()` it returns `this._page!` which seems dangerous to me.

Please tell me if the description is enough to understand the problem. or if it unclear and I should submit a reproduction example.

dgozman:
Investigation notes: make `request.frame()` throw in this case, explain to check `isNavigationRequest()` for that and `await request.response()`.

PR #26590, opened 2023-08-21
fix(route): make sure Route.fetch works for popup main request

dgozman:
References #24603.


packages/playwright-core/src/client/network.ts
```diff
@@ -277,6 +273,7 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
 
 export class Route extends ChannelOwner<channels.RouteChannel> implements api.Route {
   private _handlingPromise: ManualPromise<boolean> | null = null;
+  _context!: BrowserContext;
```
pavelfeldman, line 276: Let's do | undefined as it'll crash in line 322 otherwise?

pavelfeldman (approved):

PR #26619, opened 2023-08-22
feat(ui): show test trace events live

pavelfeldman:


packages/playwright-test/src/index.ts
```diff
@@ -644,18 +646,9 @@ class ArtifactsRecorder {
...
-              delete event.attachments;
-          }
-        }
-        const tracePath = path.join(this._artifactsDir, createGuid() + '.zip');
-        this._temporaryTraceFiles.push(tracePath);
-        await saveTraceFile(tracePath, events, this._traceOptions.sources);
-      }
+      const tracePath = path.join(this._artifactsDir, createGuid() + '.zip');
+      this._temporaryTraceFiles.push(tracePath);
+      await this._testInfo._tracing.stop(tracePath, this._traceOptions);
```
dgozman, line 651: Should we pass options when calling `start()`, like with regular tracing?

packages/playwright-test/src/worker/testTracing.ts
```diff
@@ -0,0 +1,171 @@
...
+      log: [],
+      attachments: serializeAttachments(attachments, initialAttachments),
+      error,
+    });
+  }
+
+  private _appendTraceEvent(event: trace.TraceEvent) {
+    this._traceEvents.push(event);
+    if (this._liveTraceFile)
+      fs.appendFileSync(this._liveTraceFile, JSON.stringify(event) + '\n');
```
dgozman, line 137: Isn't this called on every playwright api call, because each call is a step?

packages/playwright-test/src/matchers/toMatchSnapshot.ts
```diff
@@ -178,7 +178,7 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
     if (isWriteMissingMode) {
       writeFileSync(this.snapshotPath, actual);
       writeFileSync(this.actualPath, actual);
-      this.testInfo.attachments.push({ name: addSuffixToFilePath(this.snapshotName, '-actual'), contentType: this.mimeType, path: this.actualPath });
+      this.testInfo._attachmentsPush({ name: addSuffixToFilePath(this.snapshotName, '-actual'), contentType: this.mimeType, path: this.actualPath });
```
dgozman, line 181: Is this to avoid an "attach screenshot-actual.png" step? If so, why are there no test changes?

dgozman (approved):

packages/playwright-test/src/worker/testTracing.ts
pavelfeldman: It is!

packages/playwright-test/src/matchers/toMatchSnapshot.ts
pavelfeldman: This is to remove the dupes

PR #26626, opened 2023-08-22
chore: tolerate css imports

### LINKED ISSUE #24580: [Feature] Ability to import css files within @playwright/test and make them noop
Trying to write some visual tests using `@playwright/test`. We try to render our stories and take screenshoots. To do that, we import the stories to get the references.

The code might look something like this:

```jsx
import * as stories from './Button.stories';
import { getStoryId, loadStory } from '@/tests/story';

test('Button light', async ({ page }) => {
  await loadStory(page, getStoryId(stories, 'AllVariants'), 'light');
  await expect(page).toHaveScreenshot();
});

test('Button dark', async ({ page }) => {
  await loadStory(page, getStoryId(stories, 'AllVariants'), 'dark');
  await expect(page).toHaveScreenshot();
});
```

Importing stories also results in importing the components. If those components use css modules, Playwright will crash with an error failing to process those modules;

Something like this:
```sh
SyntaxError: Invalid or unexpected token

   at design-system/Loader/Loader.react.tsx:3

  1 | import React, { forwardRef } from 'react';
  2 | import { entries } from '../../lib/helpers/object';
> 3 | import loaderClassNames from './Loader.module.css';
```

Is there any workaround for this? For example, in `jest` its possible to make those imports a noop, with `moduleNameMapper`.

```js
moduleNameMapper: {
  "^.+\\.(css|less|scss)$": "identity-obj-proxy",
}
```

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/24580

### DIFF
--- a/tests/playwright-test/loader.spec.ts
+++ b/tests/playwright-test/loader.spec.ts
@@ -1007,3 +1007,18 @@ test('should allow test.extend.ts and test.ts files', async ({ runInlineTest }) 
   expect(result.exitCode).toBe(0);
   expect(result.passed).toBe(1);
 });
+
+test('should remove import css', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.test.ts': `
+      import './index.css';
+      import foo from './index.css';
+      import { bar } from './index.css';
+
+      import { test, expect } from '@playwright/test';
+      test('pass', async () => {});
+    `,
+  });
+  expect(result.exitCode).toBe(0);
+  expect(result.passed).toBe(1);
+});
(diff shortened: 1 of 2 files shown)


dgozman (approved):

tests/playwright-test/loader.spec.ts
```diff
@@ -1007,3 +1007,18 @@ test('should allow test.extend.ts and test.ts files', async ({ runInlineTest })
...
+
+test('should remove import css', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'a.test.ts': `
+      import './index.css';
+      import foo from './index.css';
+      import { bar } from './index.css';
+
+      import { test, expect } from '@playwright/test';
+      test('pass', async () => {});
```
mxschmitt, line 1019:
```suggestion
      test('pass', async () => {
        expect(foo).toBe('foo');
      });
```

PR #26645, opened 2023-08-23
test: unflake trace-viewer test

mxschmitt:
image


tests/library/trace-viewer.spec.ts
```diff
@@ -131,6 +131,7 @@ test('should contain action info', async ({ showTraceViewer }) => {
 test('should render network bars', async ({ page, runAndTrace, server }) => {
   const traceViewer = await runAndTrace(async () => {
     await page.goto(server.EMPTY_PAGE);
+    await page.evaluate(() => new Promise(requestAnimationFrame));
```
dgozman, line 134: I guess it's better to wait for `requestfinished` event instead?

pavelfeldman: That is weird, right? We wait until the end of goto

PR #26666, opened 2023-08-23
chore: render download progress in the host process

pavelfeldman:


packages/playwright-core/src/server/registry/oopDownloadBrowserMain.ts
```diff
@@ -15,37 +15,45 @@
...
+
+function browserDirectoryToMarkerFilePath(browserDirectory: string): string {
+  return path.join(browserDirectory, 'INSTALLATION_COMPLETE');
+}
+
+function downloadFile(options: DownloadParams): Promise<void> {
   log(`running download:`);
-  log(`-- from url: ${url}`);
-  log(`-- to location: ${destinationPath}`);
+  log(`-- from url: ${options.url}`);
```
dgozman, line 47: It looks like these three lines are already logged in the main process.

packages/playwright-core/src/server/registry/browserFetcher.ts
```diff
@@ -90,6 +95,20 @@ function downloadBrowserWithProgressBarOutOfProcess(title: string, browserDirect
...
+
+  debugLogger.log('install', `running download:`);
+  debugLogger.log('install', `-- from url: ${url}`);
+  debugLogger.log('install', `-- to location: ${zipPath}`);
+  const downloadParams: DownloadParams = {
+    title,
+    browserDirectory,
+    url,
+    zipPath,
+    executablePath,
```
dgozman, line 107: IIRC, passing `undefined` through IPC turns it into `null`? Let's make sure nothing breaks in this case.

packages/playwright-core/src/server/registry/oopDownloadBrowserMain.ts
```diff
@@ -86,85 +94,33 @@ function downloadFile(url: string, destinationPath: string, options: DownloadFil
...
-  const totalRows = 10;
-  const stepWidth = 8;
-  let lastRow = -1;
-  return (downloadedBytes: number, totalBytes: number) => {
-    const percentage = downloadedBytes / totalBytes;
-    const row = Math.floor(totalRows * percentage);
-    if (row > lastRow) {
-      lastRow = row;
-      const percentageString = String(percentage * 100 | 0).padStart(3);
+process.on('message', async message => {
```
dgozman, line 112: While we are here, let's self-destruct on `process.on('disconnect')`?

dgozman (approved):

packages/playwright-core/src/server/registry/browserFetcher.ts
pavelfeldman: I would think this happens only to the top-level value.

PR #26694, opened 2023-08-24
devops: create blob reports for all "tests 1"

yury-s:


packages/html-reporter/playwright.config.ts
```diff
@@ -15,13 +15,24 @@
  */
 
 import { devices, defineConfig } from '@playwright/experimental-ct-react';
+import type { ReporterDescription } from '@playwright/test';
+
+const reporters = () => {
+  const result: ReporterDescription[] = process.env.CI ? [
+    ['html'],
```
mxschmitt, line 22:
```suggestion
```

do we actually still need it?

mxschmitt (approved):

dgozman (approved):

packages/html-reporter/playwright.config.ts
yury-s: I guess no. Removed.

Issue #22186, opened 2023-04-04
[BUG] ignore_default_args doesn't work with --disable-features


bootstraponline:
**Context:**
- Playwright Version: v1.32.1
- Operating System: macOS 13.2.1
- Python version: Python 3.10.9
- Browser: Chrome Version 111.0.5563.146 

**Code Snippet**

Run in a REPL so you can check the process flags.

```python
$ python3
from playwright.sync_api import sync_playwright
playwright_cm = sync_playwright()
playwright = playwright_cm.__enter__()
chromium = playwright.chromium

browser = chromium.launch(
executable_path="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
headless=False,
ignore_default_args=["--disable-features"])
```

**Describe the bug**

The `--disable-features` flag is not ignored. I checked using `ps -ax | grep "Google Chrome"`

> 84522 ??         0:00.20 /Applications/Google Chrome.app/Contents/MacOS/Google Chrome --disable-field-trial-config --disable-background-networking --enable-features=NetworkService,NetworkServiceInProcess --disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-back-forward-cache --disable-breakpad --disable-client-side-phishing-detection --disable-component-extensions-with-background-pages --disable-component-update --no-default-browser-check --disable-default-apps --disable-dev-shm-usage --disable-extensions --disable-features=ImprovedCookieControls,LazyFrameLoading,GlobalMediaControls,DestroyProfileOnBrowserClose,MediaRouter,DialMediaRouteProvider,AcceptCHFrame,AutoExpandDetailsElement,CertificateTransparencyComponentUpdater,AvoidUnnecessaryBeforeUnloadCheckSync,Translate --allow-pre-commit-input --disable-hang-monitor --disable-ipc-flooding-protection --disable-popup-blocking --disable-prompt-on-repost --disable-renderer-backgrounding --disable-sync --force-color-profile=srgb --metrics-recording-only --no-first-run --enable-automation --password-store=basic --use-mock-keychain --no-service-autorun --export-tagged-pdf --enable-use-zoom-for-dsf=false --no-sandbox --user-data-dir=/var/folders/97/cjgcpwws2y97s9xnlpfzyyxc0000gn/T/playwright_chromiumdev_profile-fZKnxT --remote-debugging-pipe --no-startup-window

I also tried using `launch_persistent_context` and saw the same issue. If `ignore_default_args` is set to True then all the default flags are removed.
[...]

PR #26705, opened 2023-08-25
chore: Allow to filter defaultArgs by parameter name alone and not full parameter value

KuSh:
This avoid needing to specify full --enable-features=xxx or --disable-features=xxx command to be able to disable it

I was trying to be sure tu enable NetworkServiceInProcess2 which is the "new" name of NetworkServiceInProcess (I'll push a PR about that later) but struggle to do it until I found explanation in #22186

Specifying the complete parameter could stop working after an update if the defaultArg value has changed and that could be unfortunate


pavelfeldman:
Thank you for your PR. Please start with filing the issue explaining what your needs are!

KuSh:
> Thank you for your PR. Please start with filing the issue explaining what your needs are!

There's nothing in the contributing guide that suggest an issue is required and I took time to explain my needs in the PR :/

PR #26724, opened 2023-08-26
fix(android): use domain socket suffix chrome can discover

### LINKED ISSUE #26721: [BUG]: Running Android automation permanently prevents Chrome Remote Debugging from working
### System info
- Playwright Version: 1.37.1
- Operating System: Ubuntu 23.04
- Browser: Chrome for Android
- Other info: This occurs immediately, even if the automation fails with: _"Please install Android driver apk using `npx playwright install android`"_

### Source code

**Config file**

```js
// playwright.config.ts
const config = {
  testMatch: /.*\.playwright-test\.js/,
  fullyParallel: true,
  use: {
    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
    trace: 'on-first-retry',
  },
};
export default config;

```

```js
const { _android: android } = require('playwright');

(async () => {
  const [device] = await android.devices();
  // await device.shell('am force-stop com.android.chrome'); // This line does not prevent the problem
  const context = await device.launchBrowser(); // THIS IS THE LINE THAT BREAKS THE EMULATOR FOR GOOD
})();
```

**Steps**

**Warning: This process will effectively damage your emulator, rendering it unable to connect to Chrome Remote Debugging _ever again._**

1. Create test file (eg `android.js`) 
2. Confirm emulator is connected to Chrome's remote debugging via `chrome://inspect/#devices`
3. Run `node android.js`
4. Almost instantly you will see the remote debugging disappear. _Nothing can save this device once this occurs_

**Expected**
Remote debugging is not disrupted, or at least, not permanently prevented

**Actual**
Remote debugging for the emulator will _never_ work again, persisting through app data clearing, reboots, reinstalls, cold boots, most of the advice on StackOverflow, _everything_.

I confirmed this on two different emulators, one for API 31 and the other for API 33. Both suffer the same irreversible problem.

It presents as this device without anything under it, even though it had been connected to the device and showing its tabs immediately prior to running Playwright:
image

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/26721

### DIFF
--- a/utils/avd_start.sh
+++ b/utils/avd_start.sh
@@ -11,6 +11,7 @@ fi
 bash $PWD/utils/avd_stop.sh
 
 echo "Starting emulator"
+# nohup ${ANDROID_HOME}/emulator/emulator -avd android33 -gpu swiftshader &
 nohup ${ANDROID_HOME}/emulator/emulator -avd android33 -no-audio -no-window -no-boot-anim -no-snapshot &
 ${ANDROID_HOME}/platform-tools/adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed | tr -d '\r') ]]; do sleep 1; done; input keyevent 82'
 ${ANDROID_HOME}/platform-tools/adb devices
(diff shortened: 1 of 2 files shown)


utils/avd_start.sh
```diff
@@ -11,6 +11,7 @@ fi
 bash $PWD/utils/avd_stop.sh
 
 echo "Starting emulator"
+# nohup ${ANDROID_HOME}/emulator/emulator -avd android33 -gpu swiftshader &
```
yury-s, line 14: revert commented line?

yury-s (approved):

utils/avd_start.sh
pavelfeldman: that's as a reminder that on linux i should use software compositor, otherwise headed mode kills xserver

PR #26738, opened 2023-08-28
feat(webkit): roll WebKit to r1898

mxschmitt:
Fixes https://github.com/microsoft/playwright/pull/26678

### DIFF
--- a/packages/playwright-core/src/protocol/serializers.ts
+++ b/packages/playwright-core/src/protocol/serializers.ts
@@ -139,7 +139,7 @@ function innerSerializeValue(value: any, handleSerializer: (value: any) => Handl
     return { bi: value.toString() };
   if (isError(value)) {
     const error = value;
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/protocol/serializers.ts
```diff
@@ -139,9 +139,9 @@ function innerSerializeValue(value: any, handleSerializer: (value: any) => Handl
     return { bi: value.toString() };
   if (isError(value)) {
     const error = value;
-    if ('captureStackTrace' in globalThis.Error) {
+    if (error.stack?.startsWith(error.name + ': ' + error.message)) {
```
yury-s, line 142: If it's v8-specific, can we check for the browser name instead ?

pavelfeldman: remove this?

pavelfeldman (approved):

PR #26746, opened 2023-08-28
chore: flush har when recording tracing

pavelfeldman:


packages/playwright-core/src/server/har/harTracer.ts
```diff
@@ -517,13 +508,19 @@ export class HarTracer {
   }
 
   async flush() {
-    await Promise.all(this._barrierPromises);
+    const promises = [...this._barrierPromises];
+    this._barrierPromises.clear();
+    await Promise.race([
+      Promise.all(promises),
+      new Promise(f => setTimeout(f, 1000))
```
dgozman, line 515: I wonder whether 1 second is too low on CI to fetch response bodies?

dgozman (approved):

packages/playwright-core/src/server/har/harTracer.ts
pavelfeldman: It is on the server side and the content should be cached, so I would say should be plenty?

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -387,14 +390,29 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
...
 
   onEntryFinished(entry: har.Entry) {
+    this._pendingHarEntries.delete(entry);
     const event: trace.ResourceSnapshotTraceEvent = { type: 'resource-snapshot', snapshot: entry };
     const visited = visitTraceEvent(event, this._state!.networkSha1s);
     this._fs.appendFile(this._state!.networkFile, JSON.stringify(visited) + '\n', true /* flush */);
   }
 
+  flushHarEntries() {
+    this._harTracer.stop();
```
dgozman, line 404: This now calls `harTracer.stop` from the `stopChunk`, which we don't want.

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -387,14 +390,29 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
...
+    const harLines: string[] = [];
+    for (const entry of this._pendingHarEntries) {
+      const event: trace.ResourceSnapshotTraceEvent = { type: 'resource-snapshot', snapshot: entry };
+      const visited = visitTraceEvent(event, this._state!.networkSha1s);
+      harLines.push(JSON.stringify(visited));
+    }
+    this._pendingHarEntries.clear();
+    if (harLines.length)
+      this._fs.appendFile(this._state!.networkFile, harLines.join('\n'), true /* flush */);
```
dgozman, line 413: Let's have the last line with `\n` as well, otherwise the next write will concatenate two events.

PR #26766, opened 2023-08-29
feat(test-runner): allow list mode to print files only

unickq:
Ability to print test files only. It can be useful for CI, where you want to store a test list before running it.
For example, for https://circleci.com/docs/rerun-failed-tests/, where a test list is required, and PW grep is better than other  grep from circleci/linux

#### example 
```bash
npx playwright test --list --grep="example"                                                                                              

Listing tests:
  x.spec.ts:3:5 › example1
  x.spec.ts:6:5 › example2
Total: 2 tests in 1 file
```

```bash
PLAYWRIGHT_LIST_FILES_ONLY=true npx playwright test --list --grep="example"                                                               
 
 x.spec.ts:3:5
 x.spec.ts:6:5
```

#### x.spec.ts
```js

test("example1", async ({}) => {
  expect(1 + 1).toBe(2);
});
test("example2", async ({}) => {
  expect(1 + 1).toBe(2);
});

```


unickq:

pavelfeldman:
You can provide your own reporter to Playwright so that it would use the exact format you need. There is no need to change the default reporters for that.

unickq:
@pavelfeldman, but it's not a https://playwright.dev/docs/test-reporters#list-reporter. It's a CLI command to list tests without running them https://playwright.dev/docs/test-reporters#list-reporter

pavelfeldman:
> @pavelfeldman, but it's not a https://playwright.dev/docs/test-reporters#list-reporter. It's a CLI command to list tests without running them https://playwright.dev/docs/test-cli#reference

It still uses reporters to output list of tests and it will use the one you specified.

unickq:
> You can provide your own reporter to Playwright so that it would use the exact format you need. 

> It still uses reporters to output list of tests and it will use the one you specified.

Is there a way to use `--list` or any custom CLI command with non built-in reporter without running the tests?
cc @aslushnikov

pavelfeldman:
> Is there a way to use --list or any custom CLI command with non built-in reporter without running the tests?
cc @aslushnikov

I already replied to this above.

Issue #26529, opened 2023-08-17
In @playwright/experimental-ct-svelte, is it possible to pass a Context to a component?


kno-raziel:
## Brief intro

I have been playing with _Svelte_ lately, right now I'm exploring how to do testing, I have tried `Vitest` with `@testing-library/svelte` and was able to do some basic stuff but then had some issues when the components that I was testing includes svg and those svg are animated using the _Svelte_ transitions API, the reason seems to be that _JSDOM_ doesn’t support SVG since they don’t implement them. 

So looking at the[ _Playwright_ documentation](https://playwright.dev/docs/test-components) I found this experimental feature which seems promising compared to `Vitest` and `@testing-library/svelte` by the fact that the code runs directly in an browser engine.

## Issue

I did some test and seems promising, but then I found it lacks the support for passing Context to the component. Since in _Svelte_ we cant create components in same file like in _React_, I had to use the workaround of creating a dummy wrapper component were I have defined the Context. Example:

```svelte
<!-- DummyParentWithContext -->
<script lang="ts">
  import Example from '$lib/components/Example/Example.svelte';
  import { setContext } from 'svelte';

  setContext('contextName', { contextProp: 'contextPropValue' });
</script>

<Example />
```

And then in the test use this dummy wrapper instead of testing the component directly.

```ts
// Example.test.ts
import { test, expect } from '@playwright/experimental-ct-svelte';
import DummyParentWithContext from './DummyParentWithContext.svelte';

test.use({ viewport: { width: 500, height: 500 } });

test('Testing Context', async ({ mount }) => {
  const component = await mount(DummyParentWithContext);
  await expect(component).not.toBeEmpty();
});
```

I cant imaging doing this in a library or project that use context heavily 😖

## Requested Feature

My understanding is that the [Svelte API support passing context](https://svelte.dev/docs/client-side-component-api#creating-a-component) and is actually being used by [@testing-library/svelte](https://testing-library.com/docs/svelte-testing-library/api#render) and here is a code example of how is being use by them:

```ts
// Example.test.ts
import { render } from '@testing-library/svelte';
import { describe, expect, it } from 'vitest';
import Example from './Example.svelte';

describe('Example', async () => {
  it('should render', async () => {
    const { container } = render(Example, {
      // Example of passing context to component
      context: new Map(Object.entries({ contextName: { contextProp: 'contextPropValue' } })),
      // know issue with the types of testing library https://github.com/testing-library/svelte-testing-library/issues/212
[...]

Issue #24040, opened 2023-07-05
[Feature] Support passing Map/Set as props in component testing


segevfiner:
If you currently pass a Map or a Set in the component testing props, you will get blank object. It would be quite helpful if this just worked. Restricting component props to just plain JSON is limiting and not always the case.

pavelfeldman:
@sand4rt  would you be interested in giving it a try? We can go as far as support more builtins (Date, RegExp, Set, Map)

sand4rt:
Related to: https://github.com/microsoft/playwright/issues/22657, https://github.com/microsoft/playwright/issues/22194, https://github.com/microsoft/playwright/issues/19320, https://github.com/microsoft/playwright/issues/19151, https://github.com/microsoft/playwright/issues/18346

segevfiner:
BTW The same applies to `JSHandle` which will be quite helpful as that will allow manually serialize stuff into the browser that Playwright is not otherwise able to handle.

poulet42:
:wave: it seems the PR linked to this was reverted, are you considering reopening this issue ?

PR #26771, opened 2023-08-29
feat(ct): hooksConfig types less strict

sand4rt:
Svelte's `context` is a `Map`, would be great if i could pass that directly to the `hooksConfig`:

```typescript
test('render context', async ({ mount }) => {
  const component = await mount<HooksConfig>(Context, {
    hooksConfig: {
      context: new Map([
        ['context-key', 'context-value']
      ]),
    }
});
```

related to:
- https://github.com/microsoft/playwright/issues/24040
- https://github.com/microsoft/playwright/issues/26529

### DIFF
--- a/packages/playwright-ct-core/index.d.ts
+++ b/packages/playwright-ct-core/index.d.ts
@@ -23,7 +23,7 @@ import type {
   PlaywrightWorkerOptions,
   Locator,
 } from '@playwright/test';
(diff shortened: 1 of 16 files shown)


pavelfeldman:
Sounds like we are rewriting these types over and over. Do you have an idea on whether it is converting?

PR #26779, opened 2023-08-29
chore(ct): resolve internal type errors

sand4rt:


pavelfeldman:
Same here, sounds like types are still problematic. How do we stabilize it?

sand4rt:
A lot has happened in terms of types because it was a mess from the start. I tried to improve things incrementally, which resulted in quite a few PRs in terms of types. In this case the `type JsxComponent` was just wrong which caused a ripple effect.

I think there are still some mistakes in the internal types. I can leave them alone as this is probably getting annoying

pavelfeldman:
> I think there are still some mistakes in the internal types. I can leave them alone as this is probably getting annoying

My general observation was that we tightened them first and then started relaxing them back and this relaxing process started to get involved. So I was wondering if we should relax them back to `any` to stop testing users' patience. But at the same time, if we think we are almost there or there, we can keep going all in.

sand4rt:
> My general observation was that we tightened them first and then started relaxing them back and this relaxing process started to get involved. So I was wondering if we should relax them back to `any` to stop testing users' patience. But at the same time, if we think we are almost there or there, we can keep going all in.

I see what you mean. I closed the other PR for the time being.

There are still some minor mistakes in the internal types and it might be good to relax the `type hooksConfig` since Svelte context and Map/Set support. It would also be nice to have the slots/events typed into the mount function as well some time. Other than we should be good.

This PR should not have consequences for the end user. It's all about the internal types.

packages/playwright-ct-core/types/component.d.ts
```diff
@@ -19,16 +19,18 @@ type JsonValue = JsonPrimitive | JsonObject | JsonArray;
 type JsonArray = JsonValue[];
 export type JsonObject = { [Key in string]?: JsonValue };
 
+// JsxComponentChild can be anything, consider cases like: <>{1}</>, <>{null}</>
+export type JsxComponentChild = JsxComponent | string | number | boolean | null;
```
pavelfeldman, line 23: This type is very much public, so there are potential breakages still.

pavelfeldman (approved):

PR #26783, opened 2023-08-30
chore: fix har date types

pavelfeldman:


packages/trace-viewer/src/ui/networkTab.tsx
```diff
@@ -173,7 +182,10 @@ function sort(resources: Entry[], filter: Filter | undefined, negate: boolean) {
 }
 
 function comparator(filter: Filter | undefined) {
-  if (filter === 'time')
+  if (filter === 'start')
+    return (a: Entry, b: Entry) => a._startedDateTime - b._startedDateTime;
```
dgozman, line 186: - This needs some extra support for old traces.
- Why not sort by `Entry._monotonicTime` instead, and drop `_startedDateTime`?

dgozman (approved):

Issue #26810, opened 2023-08-31
[BUG] UI Mode feedback


mxschmitt:
- [ ] console.log(page) makes Console tab unusable.
  ```ts
    test('should allow me to add todo items', async ({ page }) => {
      console.log(page)
    });
  ```

- [ ] Make links in Console tab clickable
- [ ] Add a copy request/response body body
- [ ] Align the request/response headers in a table, then its easier to parse them with the eye (aka. like DevTools)
- [ ] When I'm in the "details view" of a specific network request, I don't want to loose information or have to go back.
  - [ ] Add the Status and Method
- [ ] On the left side there is a number, hard to understand what it is based on We should add least add a `title`.
- [ ] Would love to have a placeholder in the Inspector tab. `placeholder` seems not easily possible as per [here](https://codemirror.net/5/demo/placeholder.html).
- [ ] snapshotTitle seems wrong, for that execute the "should clear text input field when an item is added" test.
    ![Screenshot 2023-08-31 at 14 20 02](https://github.com/microsoft/playwright/assets/17984549/9dde84ee-6550-45fb-a892-789c510faa79)
 - [ ] There are performance issues when resizing the Window. It should not hang during resizing. Repro: Execute a test, open snapshot, have nothing selected in the timeline (not sure if it matters) and then resize the width. e.g. debounce.

pavelfeldman:
#### Why was this issue closed?

Thank you for your contribution to our project. This issue has been closed due to its limited upvotes and recent activity, and insufficient feedback for us to effectively act upon. Our priority is to focus on bugs that reflect higher user engagement and have actionable feedback, to ensure our bug database stays manageable.

Should you feel this closure was in error, please create a new issue and reference this one. We're open to revisiting it given increased support or additional clarity. Your understanding and cooperation are greatly appreciated.

PR #26817, opened 2023-08-31
chore(ui-mode): use WebGL renderer in xtermjs if available

mxschmitt:
Relates https://github.com/microsoft/playwright/issues/26810

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -6,7 +6,8 @@
   "dependencies": {
     "ansi-to-html": "^0.7.2",
     "codemirror": "^5.65.9",
-    "xterm": "^5.1.0",
-    "xterm-addon-fit": "^0.7.0"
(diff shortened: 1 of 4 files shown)


packages/web/package.json
```diff
@@ -6,7 +6,8 @@
   "dependencies": {
     "ansi-to-html": "^0.7.2",
     "codemirror": "^5.65.9",
-    "xterm": "^5.1.0",
-    "xterm-addon-fit": "^0.7.0"
+    "xterm": "^5.2.1",
```
pavelfeldman, line 9: I just realized that adding deps here does not add to our generated third party notice.

pavelfeldman:
Does not work for me on Linux...

pavelfeldman:
Closing as stale

PR #26879, opened 2023-09-05
docs: update node version

debs-obrien:
- update system requirement docs for node version from 16 to 18 as 16 is approaching end of life


mxschmitt:
We decided to keep supporting Node.js 16 for 1-2 releases.

Issue #22789, opened 2023-05-03
[BUG] 'React is not defined' on a shared package after updating to 1.33.0


gabriel-pq:
### System info
- Playwright Version: [v1.33.0]
- Operating System: [macOS 13.3.1]
- Browser: N/A
- Other info: React version 18.2.0; Node version 16.13.1

### Source code
The situation is quite simple:
I have a package with playwright tests `my-playwright-e2e-tests` and as part of the dependencies(package.json) I'm using `util-react`.

`util-react` is a local, shared react package, tested with jest and it's completely independent. I'm using this package just for importing some functions that are handling various strings based on some rules and nothing more!

I used Playwright v1.32.1 before the upgrade and everything was fine. After upgrading I get this error:
`ReferenceError: React is not defined` on a particular file within the `util-react` package. This is strange!

The file contains:
```
export type ErrorType = {
  code: number;
  ErrorText: JSX.Element;
  IncludeHomeLink: boolean;
  pageTitle: string;
};

export const errors: ErrorType[] = [
  {
    code: 500,
    pageTitle: 'Something went wrong',
    ErrorText: <p>Please try again.</p>,
    IncludeHomeLink: true,
  }
]
```

The solution was to add at the top of the file `import React from 'react';`. This is weird! One because the tests within the `util-react` passes without problems AND because in newer versions of React (since version 17), you no longer need to import React to use JSX syntax. This means that you can use JSX syntax and return JSX elements without importing React explicitly.

What's the catch?

**Expected**

Playwright test code should run without modifying unused code from a shared package that is working perfectly and the tests written for it passes.

yury-s:
@gabriel-pq can you share a sample project where this happens?

yury-s:
@sand4rt any clue why this may happen?

gabriel-pq:
@yury-s @sand4rt  - Please checkout [this sample project](https://github.com/gabriel-pq/my-bug-playwright) I just made.

It should work without the import because of the React version used.

Please let me know if you have any questions.

Without import:
![image](https://user-images.githubusercontent.com/85311517/235991649-4d4550f0-6c2f-4b54-82bb-1998e4b7b461.png)

With import:
![image](https://user-images.githubusercontent.com/85311517/235991805-228a9b58-d528-4ee5-bedb-17c96a7194c8.png)

sand4rt:
Thank you for providing the example, @gabriel-pq! I encounter the same error, unfortunately.

[...]

PR #26884, opened 2023-09-05
feat: respect jsx options in tsconfig

### LINKED ISSUE #26863: Error: Cannot find module '@playwright/test/jsx-runtime'
**Context:**
- Playwright Version: 1.37.1
- Operating System: Mac
- Node.js version: v16.20.0
- Visual Studio Code version: 1.81.1
- Playwright for VSCode extension version: v1.0.15
- Browser: All

**Code Snippet**

See example repo https://github.com/rubber-duck-software/playwright-typescript-monorepo-tsx

**Describe the bug**

When attempting to run a test in a pnpm monorepo, I get error 
```
Error: Cannot find module '@playwright/test/jsx-runtime'
```

This is because the playwright component testing is using its own import source for the JSX transform. See 
https://github.com/microsoft/playwright/blob/94c1b749b2ce8057c3f117ca2dbbcc713d4d2511/packages/playwright-test/bundles/babel/src/babelBundleImpl.ts#L65-L69

While changing the import source may be desired behavior for component testing, this has far reaching consequences in a monorepo. Imagine some backend package uses https://react.email or https://react-pdf.org . Changing the import source for all jsx (even indirectly) imported in a playwright test could cause a variety of issues.

### LINKED ISSUE #26936: Respect jsxImportSource from tsconfig instead of using mock jsx implementation
Currently playwright ignores whatever jsx configuration settings are used in your tsconfig and instead uses a mock implementation of jsx. This was introduced to resolve #22789. However, respecting the jsx configuration supplied in tsconfig is the better behavior. This would resolve the issue described in #26863. 

Since playwright loads the tsconfig anyways, this change would be minimally invasive.

rubber-duck-software:
resolves: #26863
resolves: #26936


tests/playwright-test/loader.spec.ts
```diff
@@ -518,24 +518,17 @@ test('should load a jsx/tsx files in ESM mode', async ({ runInlineTest }) => {
...
+      compilerOptions: {
+        jsx: 'react-jsx'
+      }
+    }),
     'a.spec.tsx': `
+      import { jsx } from 'react/jsx-runtime';
       import { test, expect } from '@playwright/test';
       const component = <div>Hello <span>world</span></div>;
       test('succeeds', () => {
-        expect(component).toEqual({
```
pavelfeldman, line 525: It looks like you are removing the test for https://github.com/microsoft/playwright/issues/22789 where empty tsconfig results in 'React is not defined'. I have a slight concern that if that is the case, you are prioritizing a rare TSC emit mode over Webpack/Babel/Vite/Rollup and other bundlers that are used to build web apps in the real world.

rubber-duck-software: Thanks for your response @pavelfeldman. 

You're right to point out the lack of a "no tsconfig" test case. I've updated this test to not include any tsconfig file so that it more closely matches its previous state. 

Now the only change in this test is that we expect the JSX to compile to react's JSX transform instead of the stub implementation supplied by playwright

pavelfeldman: > Now the only change in this test is that we expect the JSX to compile to react's JSX transform instead of the stub implementation supplied by playwright

What if I am a Vue user, I use Vite for bundling and I don't have a dependency to React at all?

rubber-duck-software: Typescript has first-class support for react's JSX transform (the jsx options are named `react`, `react-jsx`, `react-jsxdev`, `preserve`, and `react-native`). Other transforms (like Vue JSX), are supported via `importSource` (in this case `importSource: vue-jsx-runtime`). For older JSX transforms, `jsxFactory` and `jsxFragmentFactory` may be used.

So, the goal here is to match the behavior of tsc. 

There is one important divergence however. In typescript, if you omit the `jsx` option, it will throw an error if it encounters any JSX. In playwright right now, the default jsx transform is `react-jsx`. Practically this means that if you are a Vue user, in typescript you would get an error like 

```
Operator '<' cannot be applied to types 'boolean' and 'RegExp'.
```

and in playwright the error would look like 

```
JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
```

pavelfeldman: I am talking about the customers w/o tsconfig at all, they use Vite and Vue, they never rely upon tsc codegen (rare users do). Looks like you are breaking them.

rubber-duck-software: Let me check my understanding. Playwright users can roughly be split into three groups

Group A: I do not need playwright to do a JSX transform
Group B: I need playwright to do a JSX transform, but I do not care what playwright transforms my JSX into.
Group C: I need playwright to do a JSX transform and I care about what my JSX transforms into.

Obviously this whole discussion doesn't matter to users in Group A. The goal of this PR is to allow playwright to work for users in Group C. But if I'm hearing you correctly, we still need to care about users in Group B. They have  playwright importing JSX, but they are fine with `importSource: playwright`.

I have added another commit to this PR. Check out https://github.com/rubber-duck-software/playwright/commit/98912240899dd8fd15f4cd48d9dbddd2d9fed49d and see if you think that will resolve any issues for users in Group B.

phil294:
We'd love to see this shipped! (thanks @rubber-duck-software for looking into this)

In our setup, we are building parts of the HTML dynamically and inject it using `page.route(...)`. We can't purely rely on the statically built web files because parts of the index need are dynamic and are normally generated by our backend, and we want to avoid including the backend in the testing process. While this is an uncommon use case and perhaps not entirely how PW is meant to be used, it doesn't work solely because of one problem: React-server's `renderToString` is incompatible with the JSX compiler shipped by PW.

In other words, running e.g. `ReactDOMServer.renderToString(<div />)` with PW results in a cryptic `"Objects are not valid as a React child"` error while in normal Babel setups, this works fine.

This very PR solves it.

tests/playwright-test/loader.spec.ts
pavelfeldman: > Obviously this whole discussion doesn't matter to users in Group A. The goal of this PR is to allow playwright to work for users in Group C. But if I'm hearing you correctly, we still need to care about users in Group B. They have playwright importing JSX, but they are fine with importSource: playwright.

Exactly

packages/playwright/bundles/babel/src/babelBundleImpl.ts
```diff
@@ -26,7 +26,19 @@ export { parse } from '@babel/parser';
...
 export const traverse = traverseFunction;
 
-function babelTransformOptions(isTypeScript: boolean, isModule: boolean, pluginsPrologue: [string, any?][], pluginsEpilogue: [string, any?][]): TransformOptions {
+
+interface BabelTransformOptionsInput {
+  isTypeScript: boolean
+  isModule: boolean
+  pluginsPrologue: [string, any?][]
+  pluginsEpilogue: [string, any?][]
+  jsx: string | undefined
```
pavelfeldman, line 35: Love the input type (we would use type not interface for it). I'd hide jsx details into a nested struct

```js
jsx: { type, factory, fragmentFactory, importSource }
```

I'd also type the jsx type as enum if at all possible.

packages/playwright/src/transform/transform.ts
```diff
@@ -174,7 +183,7 @@ export function transformHook(originalCode: string, filename: string, moduleUrl?
   process.env.BROWSERSLIST_IGNORE_OLD_DATA = 'true';
 
   const { babelTransform }: { babelTransform: BabelTransformFunction } = require('./babelBundle');
-  const { code, map } = babelTransform(originalCode, filename, isTypeScript, !!moduleUrl, pluginsPrologue, pluginsEpilogue);
+  const { code, map } = babelTransform(originalCode, filename, { isTypeScript, isModule: !!moduleUrl, pluginsPrologue, pluginsEpilogue, jsx: tsconfig?.jsx, jsxFactory: tsconfig?.jsxFactory, jsxFragmentFactory: tsconfig?.jsxFragmentFactory, jsxImportSource: tsconfig?.jsxImportSource });
```
pavelfeldman, line 186: These now need to be a part of the `calculateHash` below since they affect the compilation result.

pavelfeldman (commented):
You code looks good now and accounts to the case I was worried about.

Two downsides of the change as I see it now:
- A lot of customers in group (B) have react-jsx as the default jsx value generated by create-react-app --typescript
-  The change further blurs the boundary between the test code (Node) and the component code (should be in the browser) in the case of component testing. Users will be accidentally importing chunks of the component code from their tests, which is breaking the execution context separation mental model. Let us consider pros and cons of landing it, apologies that I'm bringing it up late, but I was going over the pre-land considerations in my head and it bubbled up.

pavelfeldman:
Making a step back, it seems like you are using @playwright/test as a unit testing library rather than e2e framework. I wonder if you should consider using Jest for your use case?

phil294:
> Making a step back, it seems like you are using @playwright/test as a unit testing library rather than e2e framework. I wonder if you should consider using Jest for your use case?

Assuming this was meant as a reply to my comment - no, not really, unfortunately. As Jest is a Unit testing framework, but we very much make use of PW's headless browser features. I guess I could set up Jest with puppeteer and start a node browser process and migrate the code... but that's pretty much exactly where Playwright shines the most (including FF/Safari etc)

rubber-duck-software:
Hey @pavelfeldman! If you could take another look at this, that would be great :). People are using tools like react.email and react-pdf on the server and full stack frameworks like NextJS or Remix. Without this change, playwright cannot handle jsx on the server in full-stack frameworks correctly!

rubber-duck-software:
The playwright docs say
> You just write tests in TypeScript, and Playwright will read them, transform to JavaScript and run.

Playwright does my transformation to javascript from typescript. But any transformation from typescript to javascript is governed by typescript configuration. 

Email and PDF were only examples I was using- there are plenty of times when you might use JSX. JSX is just another way to call functions and JSX is fully supported by typescript. 

I think philosophically this change belongs in playwright. In order to do my transformation from typescript to javascript, playwright needs to support reading this tsconfig properties.

I have tests in this PR, but what can I do to increase confidence that this is not "potentially breaking the mainstream"

PR #26926, opened 2023-09-07
docs: writing tests python merge with node

debs-obrien:
- merge Python and node docs for this page as there is a lot of content and tables so it saves duplication
- only had to remove toHaveScreenshot as is only one that python doesn't have
- different headings added for when descriptions don't match
- this should make it easier to ensure docs don't go out of sync between langs


docs/src/writing-tests-js-python.md
```diff
@@ -52,11 +52,38 @@ test('get started link', async ({ page }) => {
 });
 ```
 
+```python title="test_example.py"
+import re
+from playwright.sync_api import Page, expect
+
+
+def test_homepage_has_Playwright_in_title_and_get_started_link_linking_to_the_intro_page(page: Page):
```
mxschmitt, line 60:
```suggestion
def test_homepage_has_playwright_in_title_and_get_started_link_linking_to_the_intro_page(page: Page):
```

docs/src/writing-tests-js-python.md
```diff
@@ -128,8 +168,20 @@ Playwright also includes async matchers that will wait until the expected condit
...
+Playwright includes [assertions](./test-assertions.md) that will wait until the expected condition is met. Using these assertions allows making the tests non-flaky and resilient. For example, this code will wait until the page gets the title containing "Playwright":
 
+```python
+import re
+from playwright.sync_api import expect
 
+expect(page).to_have_title(re.compile("Playwright"))
+```
+######
+* langs: js, python
```
mxschmitt, line 183:
```suggestion
```

I think you can remove this here right, since the whole docs is only for py/nodejs, so the check seems not relevant anymore.

docs/src/writing-tests-js-python.md
```diff
@@ -177,6 +242,29 @@ test.describe('navigation', () => {
...
+You can use various [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request) to execute code before or after your tests and to share objects between them. A `function` scoped fixture e.g. with autouse behaves like a beforeEach/afterEach. And a `module` scoped fixture with autouse behaves like a beforeAll/afterAll which runs before all and after all the tests.
+
+```python
+import pytest
+from playwright.sync_api import Page
+
+
+@pytest.fixture(scope="function", autouse=True)
+def before_each_after_each(page: Page):
+    print("beforeEach")
```
mxschmitt, line 257:
```suggestion
    print("before the test runs, aka. before each")
```

Not strong on this one, do you think it helps folks to have it a bit more verbose inside the print statement?

docs/src/writing-tests-js-python.md
```diff
@@ -177,6 +242,29 @@ test.describe('navigation', () => {
...
+    # Go to the starting url before each test.
+    page.goto("https://playwright.dev/")
+    yield
+    print("afterEach")
```
mxschmitt, line 261: ditto, if you think its better to have the afterEach more verbose.

mxschmitt (approved):

docs/src/writing-tests-js-python.md
debs-obrien: well its there for our benefit. we need the empty heading otherwise only prints in one language. this section is a shared section but is part of the assertions so it doesn't have a heading. if I leave the markdown and remove the langs part it looks too easy to just remove the empty heading and then it wont work across langs. so really this is our only way of splitting part of a section up just like when we add images, fake headings to trick the generator

docs/src/writing-tests-js-python.md
debs-obrien: yeh i like it. they can always change it as its just an example so why not especially at getting started stage

docs/src/writing-tests-js-python.md
debs-obrien: should we split this into 2 tests like we do in node so its same experience or is there a reason why in python we have it as only one???

mxschmitt: two tests sounds good.

docs/src/writing-tests-js-python.md
mxschmitt: sg

mxschmitt (approved):

Issue #26859, opened 2023-09-04
[BUG] No console output in Rust native module


samrg472:
### System info
- Playwright Version:1.37.1
- Operating System: Windows 11
- Browser: None
- Other info:
May have an effect on all operating systems.
The Rust toolchain is required.

### Source code

**Link to the GitHub repository with the repro**

https://github.com/samrg472/playwright-bug-26859

**Steps**

See the GitHub repository for reproducing.

**Expected**

I expect the the `no console output` test to have output. The provided `has console output` test is an example workaround by executing JS from Rust.

**Actual**

The `no console output` test has no console output.

pavelfeldman:
Does it appear if you run it from a simple Node JS program? Which stream does it print to?

samrg472:
> Does it appear if you run it from a simple Node JS program? Which stream does it print to?

Yes, and it prints to standard output. The issue still occurs when using the error output stream.

Issue #24591, opened 2023-08-03
[BUG] Forked process output won't get shown inside tests


mxschmitt:
```ts
test('should collect stdio', async ({ runInlineTest }) => {
  const { exitCode, report } = await runInlineTest({
    'stdio.spec.js': `
      import { test } from '@playwright/test';
      import { fork } from 'child_process';
      test('stdio', async () => {
        const child = fork('fork.js');
        await new Promise((resolve) => child.on('exit', (code) => resolve(code)));
      });
    `,
    'fork.js': `
      process.stdout.write('stdout text');
      process.stdout.write(Buffer.from('stdout buffer'));
      process.stderr.write('stderr text');
      process.stderr.write(Buffer.from('stderr buffer'));
    `
  });
  expect(exitCode).toBe(0);
  const testResult = report.suites[0].specs[0].tests[0].results[0];
  const { stdout, stderr } = testResult;
  expect(stdout).toEqual([{ text: 'stdout text' }, { buffer: Buffer.from('stdout buffer').toString('base64') }]);
  expect(stderr).toEqual([{ text: 'stderr text' }, { buffer: Buffer.from('stderr buffer').toString('base64') }]);
});
```

this won't work, since we only listen for `process.stdout.write = ...` and not listen to the actual FD. We should change it to listen on the parent process instead.

PR #26931, opened 2023-09-07
chore: collect stdio in the parent process rather than in the worker

mxschmitt:
https://github.com/microsoft/playwright/issues/26859
https://github.com/microsoft/playwright/issues/24591

### DIFF
--- a/packages/playwright/src/runner/uiMode.ts
+++ b/packages/playwright/src/runner/uiMode.ts
@@ -128,6 +128,10 @@ class UIMode {
       port: options.port,
       persistentContextOptions: {
         handleSIGINT: false,
+        ...(isUnderTest() ? {
+          viewport: { width: 1280, height: 800 },
+          colorScheme: 'light',
+        } : {})
       },
     };
     if (options.host !== undefined || options.port !== undefined) {
--- a/packages/playwright/src/worker/workerMain.ts
+++ b/packages/playwright/src/worker/workerMain.ts
@@ -15,9 +15,8 @@
  */
 
 import { colors, rimraf } from 'playwright-core/lib/utilsBundle';
-import util from 'util';
 import { debugTest, formatLocation, relativeFilePath, serializeError } from '../util';
-import type { TestBeginPayload, TestEndPayload, RunPayload, DonePayload, WorkerInitParams, TeardownErrorsPayload, TestOutputPayload } from '../common/ipc';
+import type { TestBeginPayload, TestEndPayload, RunPayload, DonePayload, WorkerInitParams, TeardownErrorsPayload } from '../common/ipc';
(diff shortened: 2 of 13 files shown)


packages/playwright/src/worker/workerMain.ts
```diff
@@ -75,25 +75,16 @@ export class WorkerMain extends ProcessRunner {
...
     process.on('uncaughtException', error => this.unhandledError(error));
-    process.stdout.write = (chunk: string | Buffer) => {
-      const outPayload: TestOutputPayload = {
-        ...chunkToParams(chunk)
-      };
-      this.dispatchEvent('stdOut', outPayload);
-      this._currentTest?._tracing.appendStdioToTrace('stdout', chunk);
-      return true;
-    };
+  }
```
pavelfeldman, line 78: I think it would be acceptable to intercept the write and send output of this process to tracing, while feeding the reporters off the worker process output as you did. It'll fix one half of the issue.

packages/playwright/src/runner/uiMode.ts
```diff
@@ -128,6 +128,10 @@ class UIMode {
       port: options.port,
       persistentContextOptions: {
         handleSIGINT: false,
+        ...(isUnderTest() ? {
```
mxschmitt, line 131: drive-by: the UI Mode tests were not passing on my dark-mode configured mac with a wide display attached (chrome window was too small and theme mismatch which made CSS tests failing)

packages/playwright/types/testReporter.d.ts
```diff
@@ -567,7 +567,7 @@ export interface JSONReportTestStep {
   steps?: JSONReportTestStep[];
 }
 
-export type JSONReportSTDIOEntry = { text: string } | { buffer: string };
```
mxschmitt, line 570: note: either revert or keep this. Since we collect the stdio from the parent process now with `setEncoding('utf8')` it will never be a buffer anymore.

tests/playwright-test/exit-code.spec.ts
```diff
@@ -36,8 +36,32 @@ test('should collect stdio', async ({ runInlineTest }) => {
   expect(exitCode).toBe(0);
   const testResult = report.suites[0].specs[0].tests[0].results[0];
   const { stdout, stderr } = testResult;
-  expect(stdout).toEqual([{ text: 'stdout text' }, { buffer: Buffer.from('stdout buffer').toString('base64') }]);
-  expect(stderr).toEqual([{ text: 'stderr text' }, { buffer: Buffer.from('stderr buffer').toString('base64') }]);
+  expect(stdout.map(e => e.text).join('')).toEqual('stdout textstdout buffer');
```
mxschmitt, line 39: note: platforms chunk/flush differently. We concatenate to have a good expectation which we can compare.

Issue #26327, opened 2023-08-07
[Feature] Add native tags annotiation


mxschmitt:
Customers don't want to have tags inside their outputDir, its noisy.
Tags should not influence the historyId.
We can use e.g. a decorator to annotate single tests.
Should it be possible to annotate describes?

Keep the native tag passing like we have in HTML and JSON reporter.

Related issues:

- https://github.com/microsoft/playwright/issues/23180
- https://github.com/microsoft/playwright/issues/24266
- https://github.com/microsoft/playwright/issues/17375

mxschmitt:
Folding into https://github.com/microsoft/playwright/issues/23180

PR #26970, opened 2023-09-08
feat: add native tags implementation

mxschmitt:
https://github.com/microsoft/playwright/issues/26327

### DIFF
--- a/tests/playwright-test/tags.spec.ts
+++ b/tests/playwright-test/tags.spec.ts
@@ -1,0 +1,331 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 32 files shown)


tests/playwright-test/tags.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    `
+  });
+  expect(exitCode).toBe(0);
+  expect(passed).toBe(6);
+});
+
+test('config.grep should work', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'playwright.config.ts': `
+      module.exports = { grep: /@tag1/ };
```
mxschmitt, line 55:
```
  tags: RegExp | string[] | string
  tagsInvert: <same-as-above>
```

p: (multiple strings are `and`, RegExp, is `or`)

tests/playwright-test/tags.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+    const result = await runInlineTest({
+      'playwright.config.ts': `
+      module.exports = {};
+    `,
+      'a.test.ts': `
+      import { test, expect } from '@playwright/test';
+      test.tag('tag1')('test1', async () => { console.log('\\n%% test1'); });
+      test('test2', async () => { console.log('\\n%% test2'); });
+    `,
+    }, { 'tag': 'tag1' });
```
dgozman, line 326:
```suggestion
    }, { 'tagInvert': 'tag1' });
```

Az8th:
Sounds really promising, thanks for this addition 🎉 
I have just a little concern about the HTML reporter : removing tags from the test title string may break the search if the regex only applies to it.
I think a filter by tags could be interesting to add, or even a dropdown menu with all detected ones, and checkboxes for each ruling their visibility

pavelfeldman:
Closing as stale

PR #26981, opened 2023-09-10
fix typo params.md

tolluset:
fix typo in docs

`{/projectName} -> /{projectName}`

### DIFF
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -1716,7 +1716,7 @@ Consider the following config:
 import { defineConfig } from '@playwright/test';
 
 export default defineConfig({
-  snapshotPathTemplate: '__screenshots__{/projectName}/{testFilePath}/{arg}{ext}',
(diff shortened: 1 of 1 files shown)


docs/src/api/params.md
```diff
@@ -1716,7 +1716,7 @@ Consider the following config:
 import { defineConfig } from '@playwright/test';
 
 export default defineConfig({
-  snapshotPathTemplate: '__screenshots__{/projectName}/{testFilePath}/{arg}{ext}',
+  snapshotPathTemplate: '__screenshots__/{projectName}/{testFilePath}/{arg}{ext}',
```
pavelfeldman, line 1719: That is not a typo, it means it is optional and is only used when there is a non-empty project name.

Tolluset: You're right. I didn't know that. I thought it's just a typo. Thank you for your comments!

PR #27010, opened 2023-09-12
feat(onEnd): allow overriding the exit code

### LINKED ISSUE #26858: [Feature]: Add ability to skip test after launch
Hello. We have a service that collects information about test runs in database and before each test run we check if it is disabled or not. This service can not only disable tests, but also mute temporarily unstable tests. Reasons why we use mutes:
- do not fail build in CI if flaky test fails. Just mute it (skip after run) and paint job to green
- sometimes tests fail for a while, but then they start to pass successfully (linked service/cdn/s3/et is broken). So there is no reason to skip them and waste developer time on fixing it. Moreover we mute only a certain number of tests from the total in order to not mute all tests
- testers checking the release of the service often need the results of running some muted tests (they can pass successfully on some retry)

So can we add ability to skip test after launch. Currently I get and error:
```
    Error: Test is skipped:

       6 |             await use();
       7 |
    >  8 |             test.skip();
         |                  ^
       9 |         },
      10 |         { scope: "test", auto: true }
      11 |     ],
```

When running code:
```
import { test as base, expect, TestInfo } from '@playwright/test';

const test = base.extend<{some: void}>({
    some: [
        async function({}, use) {
            await use();

            test.skip();
        },
        { scope: "test", auto: true }
    ],
});

test.only('test', async ({ page }) => {
});

```

pavelfeldman:
Fixes: https://github.com/microsoft/playwright/issues/26858

### DIFF
--- a/docs/src/test-reporter-api/class-reporter.md
+++ b/docs/src/test-reporter-api/class-reporter.md
@@ -106,8 +106,11 @@ The root suite that contains all projects, files and test cases.
 
 ## optional async method: Reporter.onEnd
 * since: v1.10
+- `result` ?<[Object]>
+  - `status` ?<[FullStatus]<"passed"|"failed"|"timedout"|"interrupted">>
 
 Called after all tests have been run, or testing has been interrupted. Note that this method may return a [Promise] and Playwright Test will await it.
+Reporter is allowed to override the status and hence affect the exit code of the test runner.
 
 ### param: Reporter.onEnd.result
 * since: v1.10
--- a/tests/playwright-test/reporter-onend.spec.ts
+++ b/tests/playwright-test/reporter-onend.spec.ts
@@ -1,0 +1,40 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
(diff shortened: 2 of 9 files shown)


docs/src/test-reporter-api/class-reporter.md
```diff
@@ -124,6 +124,8 @@ Result of the full test run.
 
 ## optional method: Reporter.onError
 * since: v1.10
+- `result` ?<[Object]>
```
dgozman, line 127: Should be in `onEnd`, not in `onError`.

tests/playwright-test/reporter-onend.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+const reporter = `
+class Reporter {
+  async onEnd() {
+    return { status: 'passed' };
+  }
+}
+module.exports = Reporter;
+`;
+
+test.only('should override exit code', async ({ runInlineTest }) => {
```
dgozman, line 28: No only! :)

dgozman (approved):

PR #27020, opened 2023-09-12
feat: make it possible to filter by column in HTML report

mxschmitt:
Our CLI can filter by column, users (at least me) copy the full fileName:line:column into their clipboard and expect that this will end up in a match.

### DIFF
--- a/packages/html-reporter/src/filter.ts
+++ b/packages/html-reporter/src/filter.ts
@@ -134,9 +134,7 @@ export class Filter {
         if (searchValues.text.includes(text))
           continue;
         const [fileName, line, column] = text.split(':');
-        if (line && typeof column === 'string' && searchValues.file.includes(fileName) && searchValues.line === line && searchValues.column === column)
-          continue;
-        else if (line && typeof column === 'undefined' && searchValues.file.includes(fileName) && searchValues.line === line)
+        if (searchValues.file.includes(fileName) && searchValues.line === line && (column === undefined || searchValues.column === column))
           continue;
         return false;
       }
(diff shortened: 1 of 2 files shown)


packages/html-reporter/src/filter.ts
```diff
@@ -132,8 +133,10 @@ export class Filter {
       for (const text of this.text) {
         if (searchValues.text.includes(text))
           continue;
-        const location = text.split(':');
-        if (location.length === 2 && searchValues.file.includes(location[0]) && searchValues.line.includes(location[1]))
+        const [fileName, line, column] = text.split(':');
+        if (line && typeof column === 'string' && searchValues.file.includes(fileName) && searchValues.line === line && searchValues.column === column)
```
dgozman, line 137:
```ts
if (searchValues.file.includes(fileName) && searchValues.line === line && (column === undefined || searchValues.column === column))
  continue;
```

tests/playwright-test/reporter-html.spec.ts
```diff
@@ -2064,6 +2064,32 @@ for (const useIntermediateMergeReport of [false, true] as const) {
...
+
+      expect(result.exitCode).toBe(1);
+      expect(result.passed).toBe(1);
+      expect(result.failed).toBe(1);
+
+      await showReport();
+
+      const searchInput = page.locator('.subnav-search-input');
+
+      await searchInput.fill('a.test.js:3:11');
```
dgozman, line 2084: Let's also add a test for `fileName:line` without a column.

dgozman (approved):

pavelfeldman (approved):

PR #27024, opened 2023-09-12
Render JUnit attachment paths relative to outputFile

dpaquette:
In the JUnit report XML, attachment paths are rendered relative to the configRoot. This makes sense when the report is output to stdio however it poses a problem when an output file is specified. When the report XML is saved to the specified outputFile, it becomes difficult/impossible to use the relative paths that are rendered for attachments. The attachment paths are relative to the configRoot but as a consumer of the JUnit report XML, we no longer have any knowledge of what the configRoot is. As a result, we have no way of finding those attachments.

This PR updates the JUnit reporter to attempt to make attachment paths relative to the report output file whenever possible. If no output file is specified, then the previous behavior will be maintained.

### DIFF
--- a/packages/playwright/src/reporters/junit.ts
+++ b/packages/playwright/src/reporters/junit.ts
@@ -31,6 +31,7 @@ class JUnitReporter extends EmptyReporter {
   private totalFailures = 0;
   private totalSkipped = 0;
   private outputFile: string | undefined;
+  private resolvedOutputFile: string | undefined;
   private stripANSIControlSequences = false;
 
   constructor(options: { outputFile?: string, stripANSIControlSequences?: boolean } = {}) {
@@ -51,6 +52,10 @@ class JUnitReporter extends EmptyReporter {
     this.suite = suite;
     this.timestamp = new Date();
     this.startTime = monotonicTime();
+    if (this.outputFile) {
+      assert(this.config.configFile || path.isAbsolute(this.outputFile), 'Expected fully resolved path if not using config file.');
+      this.resolvedOutputFile = this.config.configFile ? path.resolve(path.dirname(this.config.configFile), this.outputFile) : this.outputFile;
+    }
   }
 
   override async onEnd(result: FullResult) {
@@ -79,21 +84,11 @@ class JUnitReporter extends EmptyReporter {
 
     serializeXML(root, tokens, this.stripANSIControlSequences);
     const reportString = tokens.join('\n');
-    const outputFile = this.resolveOutputFile();
-    if (outputFile) {
-      await fs.promises.mkdir(path.dirname(outputFile), { recursive: true });
-      await fs.promises.writeFile(outputFile, reportString);
+    if (this.resolvedOutputFile) {
+      await fs.promises.mkdir(path.dirname(this.resolvedOutputFile), { recursive: true });
+      await fs.promises.writeFile(this.resolvedOutputFile, reportString);
     } else {
       console.log(reportString);
-    }
-  }
-
-  private resolveOutputFile(): string | undefined {
-    if (this.outputFile) {
-      assert(this.config.configFile || path.isAbsolute(this.outputFile), 'Expected fully resolved path if not using config file.');
-      return this.config.configFile ? path.resolve(path.dirname(this.config.configFile), this.outputFile) : this.outputFile;
-    } else {
-      return undefined;
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/playwright/src/reporters/junit.ts
```diff
@@ -79,16 +79,24 @@ class JUnitReporter extends EmptyReporter {
...
+    const outputFile = this.resolveOutputFile();
+    if (outputFile) {
       await fs.promises.mkdir(path.dirname(outputFile), { recursive: true });
       await fs.promises.writeFile(outputFile, reportString);
     } else {
       console.log(reportString);
     }
   }
 
+  private resolveOutputFile(): string | undefined {
```
dpaquette, line 91: I extracted this method to avoid duplication when resolving the outputFile when resolving attachment paths.

packages/playwright/src/reporters/junit.ts
```diff
@@ -190,7 +198,15 @@ class JUnitReporter extends EmptyReporter {
       for (const attachment of result.attachments) {
         if (!attachment.path)
           continue;
-        const attachmentPath = path.relative(this.config.rootDir, attachment.path);
+
+        let attachmentPath = path.relative(this.config.rootDir, attachment.path);
+        try {
+          const outputFile = this.resolveOutputFile();
```
dpaquette, line 204: Resolving the output path inside the for loop here seems wasteful. I wonder if we should change the reporter to resolve the output path once at the beginning of the process instead of resolving multiple times.

pavelfeldman:
I went down the history and found the spec that defined the current format:

https://docs.gitlab.com/ee/ci/testing/unit_test_reports.html#view-junit-screenshots-on-gitlab

So they assume that the root is specified externally at the point of the interpretation. Assuming we land your change:

1) When JUnit report is generated into the repo root, `./report.xml`.
    It points to artifacts under `./test-results`, relative to the report file
    In generates entry `<system-out>[[ATTACHMENT|test-results/foo]]</system-out>`

2) When JUnit report is generated into the repo subfolder, `./reports/report.xml`.
    It points to artifacts under `./test-results`, relative to the report file
    In generates entry `<system-out>[[ATTACHMENT|../test-results/foo]]</system-out>`

- breaks some existing users (acceptable, although the workaround for GitLab is not straightforward)
- makes resolution of resources self-contained (great)
- does not strike me as a better solution since you still need to interpret and carry the resources with you using some local script that would be instead aware of the config location and hence work without the modifications to the reporter.

dpaquette:
I don't quite follow this point:
- does not strike me as a better solution since you still need to interpret and carry the resources with you using some local script that could be aware of the config location and hence work without the modifications to the reporter.

With this change I can consume the report file and interpret the attachment paths as relative to the report I'm consuming. I don't need any additional knowledge of where those resources are (other than the report file).

I don't like the idea of breaking the GitLab but I think the GitLab approach likely already requires some moving around of the test results folder since the relative path from the Playwright configRoot might not be the same as the $CI_PROJECT_DIR that is assumed by GitLab.

Is there a way we can make this an option for the JUnit reporter? It would be nice to be able to opt-in to this functionality so that we don't break existing users.

pavelfeldman:
Could you share a bit more context? I'm trying to understand where you are processing junit result and how. Are you copying it first somewhere and then processing it or are you processing it right after the generation and then are copying it? Are you copying test-results folder as well? I assume it all starts with the `npx playwright test` call or alike in some known folder, but what happens next?

dpaquette:
Sure. here's the context.

I started by running `npx playwright init` to add playwright to my project, then added some tests to the tests folder. 

Next, I modified `playwright.config.ts` to generate a junit file by setting the reporter config to `reporter: [ ['line'],['junit', {outputFile: "results.xml" }], ]`. I also specified `use: { screenshot: 'on',    video: 'on',    trace: 'on', }` so I can capture details about each test run.

In an Azure Pipeline, I run the tests by running `npx playwright test` and then try to process the `results.xml` file inline. I'm not copying the results around but the same relative path issue would exist if I copied the report file and test-results folders.

The challenge I'm having is that I followed the documentation to add the junit reporter which puts the `results.xml` file in my project's root folder (relative to `package.json`). The results.xml file renders paths to the attachments as `..\test-results\*` which is relative to the `testDir` that is configured in playwright (in this case the `tests` folder).  The code that processes the junit report file has no context regarding the playwright config and there is no way to provide it a relative path to use when resolving attachments.

I was able to work around this by changing the outputFile to `test-results\results.xml` but that still feels strange because the attachment paths now are relative paths that indicate `..\test-results\` even though I'm already in the test-results folder. I think having the attachment paths relative to anything other than the location of the report file itself makes it harder for anyone to consume the report file.

pavelfeldman:
Ok, I did not realize it was relative to the test dir, I thought it was relative to the config dir, we should just change that and it'll be `[[ATTACHMENT|test-results/example-has-title-firefox/trace.zip]]`

pavelfeldman (approved):

packages/playwright/src/reporters/junit.ts
dpaquette: Done

PR #27038, opened 2023-09-13
chore: add --disable-search-engine-choice-screen flag to Chromium

mxschmitt:
As per `https://github.com/puppeteer/puppeteer/pull/10880`

### DIFF
--- a/packages/playwright-core/src/server/chromium/chromiumSwitches.ts
+++ b/packages/playwright-core/src/server/chromium/chromiumSwitches.ts
@@ -49,5 +49,7 @@ export const chromiumSwitches = [
   '--use-mock-keychain',
   // See https://chromium-review.googlesource.com/c/chromium/src/+/2436773
   '--no-service-autorun',
-  '--export-tagged-pdf'
+  '--export-tagged-pdf',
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/chromium/chromiumSwitches.ts
```diff
@@ -49,5 +49,7 @@ export const chromiumSwitches = [
   '--use-mock-keychain',
   // See https://chromium-review.googlesource.com/c/chromium/src/+/2436773
   '--no-service-autorun',
-  '--export-tagged-pdf'
+  '--export-tagged-pdf',
+  // https://chromium-review.googlesource.com/c/chromium/src/+/4853540
+  '--disable-search-engine-choice-screen',
```
yury-s, line 54: Does this actually affect us in Chromium?

yury-s (approved):

packages/playwright-core/src/server/chromium/chromiumSwitches.ts
mxschmitt: Might affect branded browsers.

PR #27047, opened 2023-09-13
chore: roll @zip.js/zip.js to 2.7.29

mxschmitt:


packages/trace-viewer/src/traceModelBackends.ts
```diff
@@ -16,23 +16,24 @@
 
 import type zip from '@zip.js/zip.js';
 // @ts-ignore
-import zipImport from '@zip.js/zip.js/dist/zip-no-worker-inflate.min.js';
```
yury-s, line 19: is it not minimized now?

packages/trace-viewer/src/progress.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-type Progress = (done: number, total: number) => void;
+type Progress = (done: number, total: number) => undefined;
```
yury-s, line 17: This looks very odd. Why do we need to change it to undefined? Event in zip.js it is [Promise<void>](https://github.com/gildas-lormeau/zip.js/blob/e140faab662147bb181fb5f4a2b8ac67de9fd416/index.d.ts#L1228C4-L1228C75) I believe.

yury-s (approved):

PR #27062, opened 2023-09-13
docs: fix typo

eventualbuddha:
"it's own" is equivalent to "it is own", when it should be "its own". See https://www.grammarly.com/blog/its-vs-its/.


debs-obrien:
Thank you

debs-obrien (approved):

PR #27074, opened 2023-09-14
fix: collect stdio from forked processes

### LINKED ISSUE #24591: [BUG] Forked process output won't get shown inside tests

### LINKED ISSUE #26859: [BUG] No console output in Rust native module

mxschmitt:
A summary about the attempts we tried in https://github.com/microsoft/playwright/pull/26931 for https://github.com/microsoft/playwright/issues/26859 and https://github.com/microsoft/playwright/issues/24591.

1. attempt: collect all the stdio from the worker host.
    - this made it hard to capture the live console logs for the live trace
2. attempt: collect all the stdio from the worker host, but keep proxying `process.{stdout,stderr}.write`.
    - this solved the live console logs problem, since we could still listen on them
    - we found out about a race that the stdout/stderr streams can still buffer data while the test already had finished
    - we created a best effort flush implementation but didn't like the complexity it introduced to the project
 3. attempt: keep things like they are but also listen to stdio from the worker host
    - this solves the race of console.log in the test (things stay like they are)
    - this keeps live trace working (things stay like they are)
    - a known pitfall is that it could result in logs which don't end up in the reporter onStdOut/onStdErr, but it is already a significant improvement over how it was before.
      - before: they were not getting displayed
      - after: they are getting displayed and there is a low likelihood that they don't end up in the reporter API if the write happens slightly before a test finished.

Closes #26931
Fixes #26859
Fixes #24591

### DIFF
--- a/packages/playwright/src/common/ipc.ts
+++ b/packages/playwright/src/common/ipc.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import util from 'util';
 import { serializeCompilationCache } from '../transform/compilationCache';
 import type { FullConfigInternal } from './config';
 import type { ReporterDescription, TestInfoError, TestStatus } from '../../types/test';
@@ -140,4 +141,12 @@ export function serializeConfig(config: FullConfigInternal): SerializedConfig {
     compilationCache: serializeCompilationCache(),
   };
   return result;
+}
+
+export function chunkToParams(chunk: Uint8Array | string): TestOutputPayload {
+  if (chunk instanceof Uint8Array)
+    return { buffer: Buffer.from(chunk).toString('base64') };
+  if (typeof chunk !== 'string')
+    return { text: util.inspect(chunk) };
+  return { text: chunk };
 }
--- a/packages/playwright/src/common/ipc.ts
+++ b/packages/playwright/src/common/ipc.ts
@@ -143,7 +143,7 @@ export function serializeConfig(config: FullConfigInternal): SerializedConfig {
   return result;
 }
 
-export function chunkToParams(chunk: Uint8Array | string): TestOutputPayload {
+export function stdioChunkToParams(chunk: Uint8Array | string): TestOutputPayload {
   if (chunk instanceof Uint8Array)
     return { buffer: Buffer.from(chunk).toString('base64') };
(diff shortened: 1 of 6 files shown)


packages/playwright/src/common/ipc.ts
```diff
@@ -141,3 +142,11 @@ export function serializeConfig(config: FullConfigInternal): SerializedConfig {
   };
   return result;
 }
+
+export function chunkToParams(chunk: Uint8Array | string): TestOutputPayload {
```
dgozman, line 146:
```suggestion
export function stdioChunkToParams(chunk: Uint8Array | string): TestOutputPayload {
```

dgozman (approved):

PR #27085, opened 2023-09-14
test: unflake 'filter should update stats' test

mxschmitt:
It's not possible to calculate the total test duration of the tests in the UI, since our `msToString` function is lossy. This patch unflakes the [test](https://github.com/microsoft/playwright/actions/runs/6183832112/job/16787806162?pr=27074#step:7:254) when it took longer than a second to start, by writing the test-ids to the DOM, getting them into the Node.js process, and calculating their test duration sum based on the JSON report.

Drive-by: I found a bug that we used the JSON report from the main test-runner process rather than from the merge process, so the test IDs were still old. (Without that change the `useIntermediateMergeReport` tests were not passing.

### DIFF
--- a/tests/playwright-test/reporter-html.spec.ts
+++ b/tests/playwright-test/reporter-html.spec.ts
@@ -20,6 +20,7 @@ import url from 'url';
 import { test as baseTest, expect as baseExpect, createImage } from './playwright-test-fixtures';
 import type { HttpServer } from '../../packages/playwright-core/src/utils';
 import { startHtmlReportServer } from '../../packages/playwright/lib/reporters/html';
+import { msToString } from '../../packages/web/src/uiUtils';
 const { spawnAsync } = require('../../packages/playwright-core/lib/utils');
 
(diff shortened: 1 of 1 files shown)


packages/html-reporter/src/testFileView.tsx
```diff
@@ -40,7 +40,7 @@ export const TestFileView: React.FC<React.PropsWithChildren<{
       {file.fileName}
     </span>}>
     {file.tests.filter(t => filter.matches(t)).map(test =>
-      <div key={`test-${test.testId}`} className={'test-file-test test-file-test-outcome-' + test.outcome}>
+      <div key={`test-${test.testId}`} data-test-id={test.testId} className={'test-file-test test-file-test-outcome-' + test.outcome}>
```
dgozman, line 43:
```suggestion
      <div key={`test-${test.testId}`} data-test-id={`test-${test.testId}`} className={'test-file-test test-file-test-outcome-' + test.outcome}>
```

dgozman (approved):

PR #27099, opened 2023-09-14
browser(firefox): support relative paths for firefox videos

### LINKED ISSUE #27086: [REGRESSION] recordVideo not working anymore after update
### System info
- Playwright Version: [v1.38.0]
- Operating System: macOS 13.5.1
- Browser: Firefox
- Other info:

### Source code

Minimal example following https://playwright.dev/docs/videos:

```js
import { firefox } from 'playwright-firefox';
import { datetime } from './util.js';
import { cfg } from './config.js';

const options = {
  headless: cfg.headless,
  viewport: { width: cfg.width, height: cfg.height },
  recordVideo: cfg.record ? { dir: 'data/record/', size: { width: cfg.width, height: cfg.height } } : undefined, // will record a .webm video for each page navigated; without size, video would be scaled down to fit 800x800
  recordHar: cfg.record ? { path: `data/record/eg-${datetime()}.har` } : undefined, // will record a HAR file with network requests and responses; can be imported in Chrome devtools
};
console.log(options);
const context = await firefox.launchPersistentContext(cfg.dir.browser, options);

const page = context.pages().length ? context.pages()[0] : await context.newPage(); // should always exist
await page.waitForTimeout(3000);
await context.close();
```

**Steps**

```console
SHOW=1 node recordVideo.js
{
  headless: false,
  viewport: { width: 1920, height: 1080 },
  recordVideo: { dir: 'data/record/', size: { width: 1920, height: 1080 } },
  recordHar: { path: 'data/record/eg-2023-09-14 17:23:32.955.har' }
}
```

**Expected**

Generate both a .har and .webm file in `data/record`.

**Actual**

Only a .har file is saved:
image

As you can see, this still worked on Aug 23rd.
I didn't change any of my code.

Playwright update ^1.31.0  →  ^1.37.1 seems to have broken it: https://github.com/vogler/free-games-claimer/commit/6a7cca31a4b87ecbe08387c1c5da966cc163056d

aslushnikov:
Firefox protocol requires absolute paths for video recording.

Fixes https://github.com/microsoft/playwright/issues/27086

### DIFF
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -20,6 +20,7 @@ import { Frame } from './frame';
 import * as network from './network';
 import type * as channels from '@protocol/channels';
 import fs from 'fs';
+import path from 'path';
 import { ChannelOwner } from './channelOwner';
 import { evaluationScript } from './clientHelper';
 import { Browser } from './browser';
@@ -469,6 +470,8 @@ export async function prepareBrowserContextParams(options: BrowserContextOptions
       size: options.videoSize
     };
   }
+  if (contextParams.recordVideo && contextParams.recordVideo.dir)
+    contextParams.recordVideo.dir = path.resolve(process.cwd(), contextParams.recordVideo.dir);
   return contextParams;
 }
 
--- a/tests/library/video.spec.ts
+++ b/tests/library/video.spec.ts
@@ -310,6 +310,23 @@ it.describe('screencast', () => {
     expect(fs.existsSync(path)).toBeTruthy();
   });
 
+  it('should work with relative path for recordVideo.dir', async ({ browser }, testInfo) => {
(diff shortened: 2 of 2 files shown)


packages/playwright-core/src/server/firefox/ffBrowser.ts
```diff
@@ -236,7 +237,7 @@ export class FFBrowserContext extends BrowserContext {
           // validateBrowserContextOptions ensures correct video size.
           options: {
             ...this._options.recordVideo!.size!,
-            dir: this._options.recordVideo!.dir,
+            dir: path.resolve(process.cwd(), this._options.recordVideo!.dir),
```
dgozman, line 240: Should we fix it on the client instead? Sending a relative path over protocol sounds strange.

dgozman (approved):

PR #27106, opened 2023-09-15
WIP: feat: add excludeArgs in launch options

### LINKED ISSUE #27104: [Feature] support `excludeArgs` in LaunchOptions to define patterns to exclude specific preset arguments
I find that playwright will add a bunch of preset arguments when starting a browser. For example: https://github.com/microsoft/playwright/blob/699e790f69b60adc5494195e9199914b42291666/packages/playwright-core/src/server/chromium/chromium.ts#L291-L349

But there are cases that users may want to gain more control of setting arguments. For example, `--disable-extensions` should not be set if user want to test websites with specific extensions. 

Now the LaunchOptions has an `args`  option that allow user to append arguments list. I think we need to have another options, `excludeArgs`, to exlucde options they don't want.  It should allow user to match the options they don't want with glob patterns. 

For example,

```typescript

const launchOpts = {
  channel: 'msedge',
  excludeArgs: [
    "--disable-*",
    "--enable-automation",
  ],
  args: [
   "--disable-background-timer-throttling",
  ],
}
```
which will exclude all the options that match the pattern `--disable-*` and `--enable-automation`.

Related issues:
* https://stackoverflow.com/questions/77104921/how-to-customize-msedges-command-line-options-with-playwright-python

link89:
close: https://github.com/microsoft/playwright/issues/27104


link89:
@microsoft-github-policy-service agree company="IKKEM"

mxschmitt:
There is [`ignoreDefaultArgs`](https://playwright.dev/docs/api/class-browsertype#browser-type-launch) why does that not work for you?

link89:
@mxschmitt Oops, I didn't find that. Maybe I can add glob pattern support for this options?

mxschmitt:
Closing as per https://github.com/microsoft/playwright/issues/27104#issuecomment-1721510262.

We recommend strict comparison there over glob.

Issue #26929, opened 2023-09-07
[Feature] Annotate function as an assertion to hide its implementation details.


DudaGod:
I want to write a new expect method which will wrap `toHaveScreenshot` with some custom logic before it. Currently I don't see a way how to do this. I can only add global matchers using `expect.extend`.

Moreover type with `toHaveScreenshot` options is not exported.

pavelfeldman:
I don't think expect allows reusing user-facing assertions when implementing your own assertions, would it make sense for your to wrap it as a JS call? We can try to annotate your user-land assertion so that the failure was shown on the call site rather than in this utility method. Would that work for you?

DudaGod:
> I don't think expect allows reusing user-facing assertions when implementing your own assertions

Hmm, I thought this was possible in expect. But I just wrote a piece of code and was convinced that no (it doesn't work). 

> would it make sense for your to wrap it as a JS call?

If there are no other options, then yes we will implement js method which will call the necessary methods from expect.

> We can try to annotate your user-land assertion so that the failure was shown on the call site rather than in this utility method. Would that work for you?

I think yes.

pavelfeldman:
Do you want to try implementing it?

DudaGod:
> Do you want to try implementing it?

I can do it if you can tell me which piece of code I need to pay attention to implement this feature.

pavelfeldman:
Folding into https://github.com/microsoft/playwright/issues/15951

luixo:
Hello,
Just stumbled upon this PR and I don't get how it's closed as completed, it seems like [connected PR](https://github.com/microsoft/playwright/pull/27262) adds chaining docs, but doesn't provide a way to extend page assertions.

PR #27133, opened 2023-09-16
chore: populate matcherResult in web assertions

pavelfeldman:
Ref https://github.com/microsoft/playwright/issues/26929

### DIFF
--- a/packages/playwright-core/src/client/locator.ts
+++ b/packages/playwright-core/src/client/locator.ts
@@ -18,7 +18,7 @@ import type * as structs from '../../types/structs';
 import type * as api from '../../types/types';
 import type * as channels from '@protocol/channels';
 import * as util from 'util';
-import { isString, monotonicTime } from '../utils';
+import { asLocator, isString, monotonicTime } from '../utils';
 import { ElementHandle } from './elementHandle';
 import type { Frame } from './frame';
 import type { FilePayload, FrameExpectOptions, Rect, SelectOption, SelectOptionOptions, TimeoutOptions } from './types';
@@ -355,7 +355,7 @@ export class Locator implements api.Locator {
   }
 
   toString() {
-    return `Locator@${this._selector}`;
+    return asLocator('javascript', this._selector, undefined, true);
   }
 }
 
--- a/packages/playwright/src/matchers/expect.ts
+++ b/packages/playwright/src/matchers/expect.ts
@@ -265,16 +265,6 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
         const message = jestError.message;
         if (customMessage) {
           const messageLines = message.split('\n');
-          // Jest adds something like the following error to all errors:
-          //    expect(received).toBe(expected); // Object.is equality
-          const uselessMatcherLineIndex = messageLines.findIndex((line: string) => /expect.*\(.*received.*\)/.test(line));
-          if (uselessMatcherLineIndex !== -1) {
-            // if there's a newline after the matcher text, then remove it as well.
-            if (uselessMatcherLineIndex + 1 < messageLines.length && messageLines[uselessMatcherLineIndex + 1].trim() === '')
-              messageLines.splice(uselessMatcherLineIndex, 2);
-            else
-              messageLines.splice(uselessMatcherLineIndex, 1);
-          }
           const newMessage = [
             customMessage,
             '',
--- a/tests/page/expect-matcher-result.spec.ts
+++ b/tests/page/expect-matcher-result.spec.ts
@@ -1,0 +1,247 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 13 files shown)


packages/playwright-core/src/client/locator.ts
```diff
@@ -355,7 +355,7 @@ export class Locator implements api.Locator {
   }
 
   toString() {
-    return `Locator@${this._selector}`;
+    return asLocator('javascript', this._selector);
```
dgozman, line 358:
```suggestion
    return asLocator('javascript', this._selector, false, true /* playSafe */);
```

tests/page/expect-timeout.spec.ts
```diff
@@ -20,25 +20,25 @@ import { test, expect } from './pageTest';
 test('should print timed out error message', async ({ page }) => {
   await page.setContent('<div id=node>Text content</div>');
   const error = await expect(page.locator('no-such-thing')).toHaveText('hey', { timeout: 1000 }).catch(e => e);
-  expect(stripAnsi(error.message)).toContain('Timed out 1000ms waiting for expect(received).toHaveText(expected)');
+  expect(stripAnsi(error.message)).toContain(`Timed out 1000ms waiting for expect(locator('no-such-thing')).toHaveText(hey)`);
```
dgozman, line 23: Should we quote "hey"?

packages/playwright/src/matchers/expect.ts
```diff
@@ -265,16 +265,6 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
         const message = jestError.message;
         if (customMessage) {
           const messageLines = message.split('\n');
-          // Jest adds something like the following error to all errors:
```
dgozman, line 268: Why remove this code?

tests/page/expect-matcher-result.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  const e2 = await expect(locator).not.toHaveText(/Text/, { timeout: 1 }).catch(e => e);
+  e2.matcherResult.message = stripAnsi(e2.matcherResult.message);
+  expect(e2.matcherResult).toEqual({
+    actual: 'Text content',
+    expected: /Text/,
+    message: expect.stringContaining(`Timed out 1ms waiting for expect(locator('#node')).not.toHaveText(/Text/)`),
+    name: 'toHaveText',
+    pass: true,
```
dgozman, line 41: How could `pass` be true? I find this API to be pretty strange for our users.

tests/page/expect-matcher-result.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+test('toMatchText-based assertions should have matcher result', async ({ page }) => {
+  await page.setContent('<div id=node>Text content</div>');
+  const locator = page.locator('#node');
+
+  const e1 = await expect(locator).toHaveText(/Text2/, { timeout: 1 }).catch(e => e);
+  e1.matcherResult.message = stripAnsi(e1.matcherResult.message);
+  expect(e1.matcherResult).toEqual({
+    actual: 'Text content',
+    expected: /Text2/,
+    message: expect.stringContaining(`Timed out 1ms waiting for expect(locator('#node')).toHaveText(/Text2/)`),
```
dgozman, line 29: `MatcherResult` type says it's a function.

tests/page/expect-matcher-result.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    message: expect.stringContaining(`Timed out 1ms waiting for expect(locator('#node2')).toBeVisible()`),
+    name: 'toBeVisible',
+    pass: false,
+  });
+
+  const e2 = await expect(page.locator('#node')).not.toBeVisible({ timeout: 1 }).catch(e => e);
+  e2.matcherResult.message = stripAnsi(e2.matcherResult.message);
+  expect(e2.matcherResult).toEqual({
+    actual: false,
+    expected: false,
```
dgozman, line 63: We also have things like `toBeChecked({ checked: false })`. I wonder whether we want `expected: true` or `expected: false` there.

tests/page/expect-timeout.spec.ts
pavelfeldman: Not sure, in fact stock expect leaves those as `received` and `expected` which I think it is weird. I wonder if I should be saying `locator` instead of the actual value. Let me see how it would look.

packages/playwright/src/matchers/expect.ts
pavelfeldman: This code was doing something weird in a weird place in a wrong way. Tell me what it tried to do and maybe we find a better place.

tests/page/expect-matcher-result.spec.ts
pavelfeldman: That is not for us to decide, failing not will have pass set to true in expect.

tests/page/expect-matcher-result.spec.ts
pavelfeldman: Exception.matcherResult is not of the type MatcherResult.

tests/page/expect-matcher-result.spec.ts
pavelfeldman: I would say we want expected false there, I can add explicit test for documentation purposes.

dgozman (approved):

PR #27162, opened 2023-09-18
fix(tracing): bump trace version to V5, migrate V4 traces to consoleMessage.args

### LINKED ISSUE #27144: [BUG]can not display trace
When I open trace, the page is blank
image

dgozman:
This moves the fix in #27095 from `modernize` to `appendEvent`. The reason is that `trace V4` is used both for older traces that do not have `consoleMessage.args` and the new ones with `args`. Since we do not call `modernize` for traces of the same version, the original fix does not help in this case.

Fixes #27144.


pavelfeldman (approved):

packages/trace-viewer/src/traceModel.ts
```diff
@@ -348,6 +343,48 @@ export class TraceModel {
...
+        return {
+          type: 'console',
+          time: event.time,
+          pageId: event.pageId,
+          messageType: consoleMessage.type,
+          text: consoleMessage.text,
+          args: consoleMessage.args,
+          location: consoleMessage.location,
+        };
+      }
```
yury-s, line 384: else return null

yury-s (approved):

PR #27220, opened 2023-09-20
test: speed up installation tests by not downloading from cdn multiple times

dgozman:


tests/installation/playwright-should-work.spec.ts
```diff
@@ -15,7 +15,7 @@
  */
 import { test, expect } from './npmTest';
 
-test(`playwright should work`, async ({ exec, nodeMajorVersion, installedSoftwareOnDisk }) => {
```
aslushnikov, line 18: shall we use realCDN here to have one full simple e2e test that executes main workflow?

tests/installation/npmTest.ts
```diff
@@ -77,25 +63,88 @@ _expect.extend({
...
+          response.writeHead(metaInfo.statusCode, metaInfo.statusMessage, metaInfo.headers);
+          const done = () => {
+            debug(`[cdn proxy] serving ${request.url!} finished`);
+            response.end();
+          };
+          fs.createReadStream(cachedPath).pipe(response).on('close', done).on('error', done);
+          debug(`[cdn proxy] serving ${request.url!} from cached ${cachedPath}`);
+        });
+
+        const PORT = 15689;
```
aslushnikov, line 142: shall we have PORT 0 here? Alternatively, can we start the CDN in global setup once for everybody?

tests/installation/npmTest.ts
```diff
@@ -77,25 +63,88 @@ _expect.extend({
...
+            };
+            const factory = options.protocol === 'https:' ? https : http;
+            let doneCallback = () => {};
+            const donePromise = new Promise<void>(f => doneCallback = () => {
+              debug(`[cdn proxy] downloading ${realUrl} finished`);
+              f();
+            });
+            const realRequest = factory.request(options, (realResponse: http.IncomingMessage) => {
+              const metaInfo = {
+                statusCode: realResponse.statusCode,
```
aslushnikov, line 119: I would appreciate if the `metaInfo` also has a version. This way, if there are changes to this storage format, then no developers would need to manually delete this cache.

aslushnikov (approved):

tests/installation/npmTest.ts
dgozman: The cache is actually removed in global setup, so we will not have to worry about this.

PR #27245, opened 2023-09-21
chore: export MakeMatchers type

KuznetsovRoman:
Export `MakeMatchers` type

I am developing npm package with custom `expect` matchers, and always extending `global` is not suitable, because user might not want to extend `expect` with all of them. I found out how to easily extend `expect` only with chosen matchers, but i need to extend `MakeMatchers` type to do this.


pavelfeldman:
MakeMatchers is an implementation detail, but I was planning to make expect.extend return the extended type.

Issue #21620, opened 2023-03-13
[BUG] `.toHaveScreenshot()` does not respect `backface-visibility: hidden;` CSS property


ngnijland:
### System info
- Playwright Version: v1.30.0
- Operating System: macOS, Docker with `mcr.microsoft.com/playwright:v1.30.0-focal`
- Browser: WebKit

### Source code

HTML and CSS is to sketch the situation. The screen recordings should give more context to the issue.

```html
<div class="card-wrapper">
  <div class="card card--rotate-to-front">
    <div class="card-face card-face--front">
      <div class="card-frame card-frame--front" style="transform: none;"></div>
    </div>
    <div class="card-face card-face--back">
      <div class="card-frame card-frame--back"></div>
    </div>
  </div>
</div>
```

```css
.card-wrapper {
     perspective: 800px;
}

.card {
    transform-style: preserve-3d;
}

.card--rotate-to-front {
    animation: rotateBounceToFront;
    animation-duration: 1s;
    animation-fill-mode: forwards;
}

.card--rotate-to-back {
    animation: rotateBounceToBack;
    animation-duration: 1s;
    animation-fill-mode: forwards;
}

.card-face {
    backface-visibility: hidden;
}
```

The above HTML and CSS (among other code) result in the following interface:
https://user-images.githubusercontent.com/4727742/224694220-047a0a1a-0ed4-4315-865e-e4f8480027b2.mov

**Config file**

```js
// playwright.config.ts
import { defineConfig, devices } from '@playwright/test';

export default defineConfig({
  projects: [
    {
      name: "Mobile Safari",
      use: { ...devices["iPhone 12"] },
    },
});
```

**Test file (self-contained)**

```js
test("gift cards", async ({ page, isMobile }) => {
    await page.goto("/gift-card");

    await page.getByText("Wondrous waterways").click();
    await expect.soft(page).toHaveScreenshot("card.png");
});

```

**Steps**
- Run test
- Let test take screenshot

**Expected**

Front of the card should be visible.

**Actual**

Back of the card is visible mirrored.

**NOTE:** The behavior of seeing the back of the card mirrored is the same as when you remove `backface-visibility: hidden;` css property from the `.card-face` class.

Video of the actual behavior:

https://user-images.githubusercontent.com/4727742/224695831-17ccd2f3-ee0e-4c77-85af-0fdc183595e3.mov

dgozman:
[...]

PR #27246, opened 2023-09-21
fix(webkit): screenshot in UI process on mac and linux

yury-s:
Reference #21620


packages/playwright-core/src/server/webkit/wkPage.ts
```diff
@@ -856,7 +857,20 @@ export class WKPage implements PageDelegate {
     const omitDeviceScaleFactor = scale === 'css';
     this.validateScreenshotDimension(rect.width, omitDeviceScaleFactor);
     this.validateScreenshotDimension(rect.height, omitDeviceScaleFactor);
-    const result = await this._session.send('Page.snapshotRect', { ...rect, coordinateSystem: documentRect ? 'Page' : 'Viewport', omitDeviceScaleFactor });
+
+    let result: any;
+    // Take screenshot in the UI process where possible. This way we capture CSS transformations.
+    const takeSnapshotInUIProcess = os.platform() === 'linux' ||
```
dgozman, line 863: `os.platform()` does not work for remote case, we should get the actual platform of the browser. We already do that in `crBrowser` somehow.

Alternatively, browser should figure our the best strategy and either `Playwright.takePageScreenshot` or forward to `Page.snapshotRect`.

pavelfeldman:
Closing as stale

PR #27248, opened 2023-09-22
feat(expect): allow chaining expects

pavelfeldman:


tests/playwright-test/expect.spec.ts
```diff
@@ -704,3 +709,155 @@ test('should not leak long expect message strings', async ({ runInlineTest }) =>
...
+              \`Expected: \${this.isNot ? 'not' : ''}\${this.utils.printExpected(expected)}\\n\` +
+              (matcherResult ? \`Received: \${this.utils.printReceived(matcherResult.actual)}\` : '') +
+              log
+          : () =>  this.utils.matcherHint('toBe', locator, expected, expectOptions) +
+              '\\n\\n' +
+              \`Expected: \${this.utils.printExpected(expected)}\n\` +
+              (matcherResult ? \`Received: \${this.utils.printReceived(matcherResult.actual)}\` : '') +
+              log;
+
+        return {
```
dgozman, line 837: If I don't return all these fields in my custom matcher, will things keep working?

utils/generate_types/overrides-test.d.ts
```diff
@@ -341,38 +341,92 @@ type AllMatchers<R, T> = PageAssertions & LocatorAssertions & APIResponseAsserti
...
+  <T = unknown>(actual: T, messageOrOptions?: string | { message?: string }): MakeMatchers<void, T, ExtendedMatchers>;
+  soft: <T = unknown>(actual: T, messageOrOptions?: string | { message?: string }) => MakeMatchers<void, T, ExtendedMatchers>;
   poll: <T = unknown>(actual: () => T | Promise<T>, messageOrOptions?: string | { message?: string, timeout?: number, intervals?: number[] }) => BaseMatchers<Promise<void>, T> & {
     /**
      * If you know how to test something, `.not` lets you test its opposite.
      */
      not: BaseMatchers<Promise<void>, T>;
   };
-  extend(matchers: any): void;
+  extend<MoreMatchers extends Record<string, (this: State, receiver: unknown, ...args: any[]) => MatcherReturnType | Promise<MatcherReturnType>>>(matchers: MoreMatchers): Expect<ExtendedMatchers & MoreMatchers>;
```
dgozman, line 424: Judging by our own `toHaveLoggedSoftwareDownload`, forcing the return type is a breaking change.

utils/generate_types/overrides-test.d.ts
```diff
@@ -341,38 +341,92 @@ type AllMatchers<R, T> = PageAssertions & LocatorAssertions & APIResponseAsserti
...
+  expectedColor?: MatcherHintColor;
+  isDirectExpectCall?: boolean;
+  isNot?: boolean;
+  promise?: string;
+  receivedColor?: MatcherHintColor;
+  secondArgument?: string;
+  secondArgumentColor?: MatcherHintColor;
+};
+
+export interface JestMatcherUtils {
```
dgozman, line 363: Are you sure we want to expose all of this?

utils/generate_types/overrides-test.d.ts
```diff
@@ -341,38 +341,92 @@ type AllMatchers<R, T> = PageAssertions & LocatorAssertions & APIResponseAsserti
 
 type IfAny<T, Y, N> = 0 extends (1 & T) ? Y : N;
 type Awaited<T> = T extends PromiseLike<infer U> ? U : T;
-type MakeMatchers<R, T> = {
+type ToUserMatcher<F> = F extends (a: any, ...args: infer Rest) => infer R ? (...args: Rest) => (R extends PromiseLike<infer U> ? Promise<void> : void) : never;
+type ToUserMatcherForArgType<T, ArgType> = T extends (arg: ArgType, ...rest: any[]) => any ? ToUserMatcher<T> : never;
```
dgozman, line 345: I'd inline `ToUserMatcher` here:

```suggestion
type ToUserMatcherForArgType<T, ArgType> = T extends (arg: ArgType, ...rest: infer Rest) => infer R ? (...args: Rest) => (R extends PromiseLike<infer U> ? Promise<void> : void) : never;
```

tests/playwright-test/expect.spec.ts
pavelfeldman: You must return `message` and `pass`. The rest is optional.

utils/generate_types/overrides-test.d.ts
pavelfeldman: I definitely don't, but I think this ship has sailed.

utils/generate_types/overrides-test.d.ts
pavelfeldman: I think toHaveLoggedSoftwareDownload is not right, i'll fix it.

dgozman (approved):

PR #27295, opened 2023-09-25
feat(ct): allow baseUrl and host config

### LINKED ISSUE #27283: [BUG] Vite preview host is ignored in component tests
### System info
- Playwright Version: v1.38.1
- Operating System: Mac 13.5.2
- Browser: Chromium
- Other info:

### Source code

**Config file**

```js
import { defineConfig } from "@playwright/experimental-ct-react";

const { PWTEST_CHROMIUM_WS_ENDPOINT } = process.env;

export default defineConfig({
  use: {
    ctPort: 3100,
    ctViteConfig: {
      preview: {
        host: "192.168.1.186",
      },
    },
  },
  projects: [
    {
      name: "chromium",
      use: { connectOptions: { wsEndpoint: PWTEST_CHROMIUM_WS_ENDPOINT } },
    },
  ],
});
```

**Steps**
- Launch chromium server inside docker and save the got `wsEndpoint` as `PWTEST_CHROMIUM_WS_ENDPOINT` process env
- Run any test file using component testing

**Expected**

- Test to pass

**Actual**

- Error: net::ERR_CONNECTION_REFUSED at http://localhost:3100/

### Explanation

I want to perform visual tests using Playwright. My CI runs on Github which uses linux machines. My personal computer is a Mac. 
Screenshots differ between operating systems so on my Mac I start Playwright chromium server inside docker ubuntu image and use my local Playwright to perform tests on it.

Browser run inside docker container needs to access Vite preview but in https://github.com/microsoft/playwright/blob/d6ec1ae3994f127e38b866a231a34efc6a4cac0d/packages/playwright-ct-core/src/vitePlugin.ts#L212C21-L212C45 the host is hardcoded to `localhost`. On Mac, docker doesn't support host network so in order to access local port inside docker I need to use my machine's ip.

### Workaround

I can make it work right now by overriding `process.env.PLAYWRIGHT_TEST_BASE_URL` because the same config file is also run inside launched server:

```js
import { defineConfig } from "@playwright/experimental-ct-react";

process.env.PLAYWRIGHT_TEST_BASE_URL = "http://192.168.1.186:3100/";

export default defineConfig({
```

### Potential fix

In https://github.com/microsoft/playwright/blob/d6ec1ae3994f127e38b866a231a34efc6a4cac0d/packages/playwright-ct-core/src/vitePlugin.ts#L212C21-L212C45 the `localhost` could be replaced by the actual Vite's preview host or by the value from `finalConfig`. To make things easier, new `ctHost` could be implemented which would fit the already existing `ctPort` config.

sand4rt:
closes: https://github.com/microsoft/playwright/issues/27283

### DIFF
--- a/packages/playwright-ct-core/src/vitePlugin.ts
+++ b/packages/playwright-ct-core/src/vitePlugin.ts
@@ -70,6 +70,7 @@ export function createPlugin(
 
       const use = config.projects[0].use as CtConfig;
       const port = use.ctPort || 3100;
+      const host = use.baseURL || 'localhost';
       const relativeTemplateDir = use.ctTemplateDir || 'playwright';
 
       // FIXME: use build plugin to determine html location to resolve this.
@@ -91,7 +92,8 @@ export function createPlugin(
           outDir: use.ctCacheDir ? path.resolve(configDir, use.ctCacheDir) : path.resolve(templateDir, '.cache')
         },
         preview: {
-          port
+          port,
+          host,
         },
         // Vite preview server will otherwise always return the index.html with 200.
         appType: 'custom',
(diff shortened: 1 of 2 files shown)


pavelfeldman:
It looks great, would be nice to have a test for this though. Something like the tests in tests/playwright-test/playwright.ct-react.spec.ts

sand4rt:
> It looks great, would be nice to have a test for this though. Something like the tests in tests/playwright-test/playwright.ct-react.spec.ts

i've added the tests

pavelfeldman (approved):

PR #27321, opened 2023-09-27
docs(trace-viewer.md): include trace options with object

parideis:
I'm limited to 1gb artifacts space on gitlab and I've got the useful advice to disable some trace options like screenshots and sources from a discord user.
This PR adds the missing information in the docs, that you can pass an object to `trace`.

### DIFF
--- a/docs/src/trace-viewer.md
+++ b/docs/src/trace-viewer.md
@@ -160,6 +160,13 @@ Available options to record a trace:
 
 You can also use `trace: 'retain-on-failure'` if you do not enable retries but still want traces for failed tests.
 
+To save disk space, you can use an object to configure the trace options and disable features like screenshots:
+- `'mode'` - Specify the trace mode, e.g. `'on-first-retry'`.
+- `'snapshots'` -  Set to `true` to capture snapshots during tracing.
+- `'screenshots'` - Set to `true` to capture screenshots during tracing.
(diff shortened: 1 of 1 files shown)


docs/src/trace-viewer.md
```diff
@@ -160,6 +160,13 @@ Available options to record a trace:
 
 You can also use `trace: 'retain-on-failure'` if you do not enable retries but still want traces for failed tests.
 
+To save disk space, you can use an object to configure the trace options and disable features like screenshots:
```
dgozman, line 163: Perhaps we should just link to https://playwright.dev/docs/api/class-testoptions#test-options-trace? That would be:

```
There are more granular options available, see [`property: TestOptions.trace`].
```

Let me know what you think.

dgozman (commented):
Thank you for the PR! Please see comments inline.

yury-s:
Closing per the response above, feel free to reopen when you get back to it!

docs/src/trace-viewer.md
parideis: Thanks I like your suggestion to link to the docs much more. So there is no need to update the docs in two places if the TestOptions.trace changes. 👍 
Because I took so long, I've opened a new PR: https://github.com/microsoft/playwright/pull/31182

PR #27332, opened 2023-09-27
chore: delete temporary artifacts after test

pavelfeldman:


packages/playwright/src/index.ts
```diff
@@ -586,12 +593,15 @@ class ArtifactsRecorder {
   }
 
   async didFinishTestFunction() {
-    if (this._testInfo._isFailure() && (this._screenshotMode === 'on' || this._screenshotMode === 'only-on-failure'))
+    const captureScreenshots = this._screenshotMode === 'on' || (this._screenshotMode === 'only-on-failure' && this._testInfo._isFailure());
```
dgozman, line 596: This is a semantic change: we start capturing screenshots at the test function end when `screenshot: 'on'`.

dgozman (approved):

packages/playwright/src/index.ts
pavelfeldman: Let's say this is intended.

PR #27408, opened 2023-10-03
feat: ability to specify maxFailures relative to total tests

### LINKED ISSUE #26861: [Feature] ability to set max failures relative to the total number of tests
Hello.

To my mind specify `maxFailures` as a percentage of the total number of tests is much more convenient than constantly maintaining a static value as the number of tests grows.

For example I have a small project with 100 tests and `maxFailures: 30` in the config. After some time the number of tests will increase to 300 and I will need to remember to change the value of `maxFailures` to `100` (1/3).

But if I able to set `maxFailures: 0.3` (percentage of the total number of tests) then it will always work correctly. It means I will not need to keep this value up to date as the number of tests grows.

DudaGod:
## What's done

Implement ability to specify `maxFailures` relative to total tests in order not to keep track of the relevance of the `maxFailures` value.

Fixes https://github.com/microsoft/playwright/issues/26861

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -215,10 +215,10 @@ export default defineConfig({
 The maximum number of test failures (test retry is also a failure) for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
 
 Can be specified as:
-* static value (positive integer number);
-* dynamic value (positive float number less than 1) relative to the total number of tests.
+* static value - positive integer number;
+* dynamic value - percentage relative to the total number of tests.
 
-For example, if 100 tests are run in the project and the `maxFailures` is set to 0.3 (dynamic value), then when 30 tests are failed, testing will stop. When using a static value you just need to specify 30.
+For example, if 100 tests are run in the project and the `maxFailures` is set to `'30%'` (dynamic value), then when 30 tests are failed, testing will stop. When using a static value you just need to specify `30``.
 
 Use dynamic value if you don't want to keep it up to date as the number of tests increases.
 
@@ -230,7 +230,7 @@ Also available in the [command line](../test-cli.md) with the `--max-failures` a
 import { defineConfig } from '@playwright/test';
 
 export default defineConfig({
-  maxFailures: process.env.CI ? 0.3 : 0,
+  maxFailures: process.env.CI ? '30%' : 0,
 });
 ```
 
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -212,7 +212,15 @@ export default defineConfig({
 * since: v1.10
 - type: ?<[int]>
 
-The maximum number of test failures for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
+The maximum number of test failures (test retry is also a failure) for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
+
+Can be specified as:
+* static value (positive integer number);
+* dynamic value (positive float number less than 1) relative to the total number of tests.
+
+For example, if 100 tests are run in the project and the `maxFailures` is set to 0.3 (dynamic value), then when 30 tests are failed, testing will stop. When using a static value you just need to specify 30.
+
+Use dynamic value if you don't want to keep it up to date as the number of tests increases.
 
 Also available in the [command line](../test-cli.md) with the `--max-failures` and `-x` options.
 
@@ -222,7 +230,7 @@ Also available in the [command line](../test-cli.md) with the `--max-failures` a
 import { defineConfig } from '@playwright/test';
 
 export default defineConfig({
-  maxFailures: process.env.CI ? 1 : 0,
+  maxFailures: process.env.CI ? 0.3 : 0,
 });
 ```
 
--- a/packages/playwright/src/common/config.ts
+++ b/packages/playwright/src/common/config.ts
@@ -206,8 +206,13 @@ function pathResolve(baseDir: string, relative: string | undefined): string | un
   return path.resolve(baseDir, relative);
 }
 
-function handleMaxFailures(maxFailures: number): number {
-  return !Number.isInteger(maxFailures) && maxFailures < 1 ? maxFailures : Math.trunc(maxFailures);
+function handleMaxFailures(maxFailures: number | string): number {
+  if (typeof maxFailures === 'string') {
+    const parsedMaxFailures = parseInt(maxFailures, 10);
+    return maxFailures.endsWith('%') ? Math.min(0.99, parsedMaxFailures / 100) : parsedMaxFailures;
+  }
+
+  return maxFailures;
 }
 
 function resolveReporters(reporters: Config['reporter'], rootDir: string): ReporterDescription[] | undefined {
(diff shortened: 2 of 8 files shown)


docs/src/test-api/class-testconfig.md
```diff
@@ -212,7 +212,15 @@ export default defineConfig({
 * since: v1.10
 - type: ?<[int]>
 
-The maximum number of test failures for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
+The maximum number of test failures (test retry is also a failure) for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
+
+Can be specified as:
+* static value (positive integer number);
+* dynamic value (positive float number less than 1) relative to the total number of tests.
```
DudaGod, line 219: Perhaps the description is not very clear and I should just use `integer number` and `float number` without `static` and `dynamic`.

packages/playwright/src/reporters/base.ts
```diff
@@ -254,11 +254,14 @@ export class BaseReporter implements ReporterV2 {
...
+    let maxFailures = this.config.maxFailures;
+    maxFailures = Number.isInteger(maxFailures) ? maxFailures : Math.ceil(maxFailures * this.totalTestCount);
+
+    if (!maxFailures)
       return;
-    if (this._failureCount < this.config.maxFailures)
+    if (this._failureCount < maxFailures)
       return;
-    console.log(colors.yellow(`Testing stopped early after ${this.config.maxFailures} maximum allowed failures.`));
+    console.log(colors.yellow(`Testing stopped early after ${maxFailures} maximum allowed failures.`));
```
DudaGod, line 264: Here in case of specifying the value as float number will be displayed only calculated value relative to total tests. Maybe in this case I should add also real specified `maxFailures` in config/cli.

docs/src/test-api/class-testconfig.md
pavelfeldman: For `--workers` we allow setting the value `50%`. This helps us stay away from floating point, lets do the same here.

packages/playwright/src/common/config.ts
```diff
@@ -83,7 +83,7 @@ export class FullConfigInternal {
       globalTimeout: takeFirst(configCLIOverrides.globalTimeout, config.globalTimeout, 0),
       grep: takeFirst(config.grep, defaultGrep),
       grepInvert: takeFirst(config.grepInvert, null),
-      maxFailures: takeFirst(configCLIOverrides.maxFailures, config.maxFailures, 0),
+      maxFailures: handleMaxFailures(takeFirst(configCLIOverrides.maxFailures, config.maxFailures, 0)),
```
pavelfeldman, line 86: This will align with the workers treatment as well.

packages/playwright/src/reporters/base.ts
```diff
@@ -254,11 +254,14 @@ export class BaseReporter implements ReporterV2 {
   }
 
   private _printMaxFailuresReached() {
-    if (!this.config.maxFailures)
+    let maxFailures = this.config.maxFailures;
+    maxFailures = Number.isInteger(maxFailures) ? maxFailures : Math.ceil(maxFailures * this.totalTestCount);
```
pavelfeldman, line 258: Having this logic here seems fragile - you should probably resolve failures count after parsing the test suite, or explicitly pass a boolean saying that max failures was reached.

packages/playwright/src/runner/failureTracker.ts
```diff
@@ -40,7 +42,9 @@ export class FailureTracker {
   }
 
   hasReachedMaxFailures() {
-    const maxFailures = this._config.config.maxFailures;
+    let maxFailures = this._config.config.maxFailures;
```
pavelfeldman, line 45: There is no point in recomputing it over and over, easier to resolve in constructor once.

docs/src/test-api/class-testconfig.md
DudaGod: > For --workers we allow setting the value 50%. This helps us stay away from floating point, lets do the same here.

Ok, I didn't notice such an implementation. Fixed

packages/playwright/src/common/config.ts
```diff
@@ -206,6 +206,15 @@ function pathResolve(baseDir: string, relative: string | undefined): string | un
   return path.resolve(baseDir, relative);
 }
 
+function handleMaxFailures(maxFailures: number | string): number {
+  if (typeof maxFailures === 'string') {
+    const parsedMaxFailures = parseInt(maxFailures, 10);
+    return maxFailures.endsWith('%') ? Math.min(0.99, parsedMaxFailures / 100) : parsedMaxFailures;
```
DudaGod, line 212: Here I had to set 0.99 because if I specify 1 I will not be able to correctly calculate the value relative to the total number of tests.  I mean that I will not be able to distinguish between the value `1` (static) and `100%` (dynamic, which will be transformed to `1`).

docs/src/test-api/class-testconfig.md
```diff
@@ -212,7 +212,15 @@ export default defineConfig({
 * since: v1.10
 - type: ?<[int]>
 
-The maximum number of test failures for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
+The maximum number of test failures (test retry is also a failure) for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
+
+Can be specified as:
+* static value - positive integer number;
+* dynamic value - percentage relative to the total number of tests.
```
pavelfeldman, line 219:
```suggestion
* relative value - percentage relative to the total number of tests.
```

docs/src/test-api/class-testconfig.md
```diff
@@ -212,7 +212,15 @@ export default defineConfig({
 * since: v1.10
 - type: ?<[int]>
 
-The maximum number of test failures for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
+The maximum number of test failures (test retry is also a failure) for the whole test suite run. After reaching this number, testing will stop and exit with an error. Setting to zero (default) disables this behavior.
+
+Can be specified as:
+* static value - positive integer number;
```
pavelfeldman, line 218:
```suggestion
* absolute value - positive integer number;
```

docs/src/test-api/class-testconfig.md
```diff
@@ -212,7 +212,15 @@ export default defineConfig({
...
+* dynamic value - percentage relative to the total number of tests.
+
+For example, if 100 tests are run in the project and the `maxFailures` is set to `'30%'` (dynamic value), then when 30 tests are failed, testing will stop. When using a static value you just need to specify `30``.
```
pavelfeldman, line 221:
```suggestion
For example, if 100 tests are run in the project and the `maxFailures` is set to `'30%'` (relative value), then when 30 tests are failed, testing will stop. To use an absolute value you can specify `30`.
```

docs/src/test-api/class-testconfig.md
```diff
@@ -212,7 +212,15 @@ export default defineConfig({
...
+
+Use dynamic value if you don't want to keep it up to date as the number of tests increases.
```
pavelfeldman, line 223:
```suggestion
Use relative value if you don't want to keep it up to date as the number of tests increases.
```

packages/playwright/src/common/config.ts
pavelfeldman: You should pass a number | string instead of this, or some up with a data structure. 0.99 is not great.

pavelfeldman:
Closing as stale

PR #27418, opened 2023-10-04
feat: toHaveAttribute without value

### LINKED ISSUE #27341: [Feature]: Allow a `toHaveAttribute` Overload Where Only the Attribute Is Provided
There are times where we need to test that an element (or a Custom Web Component) has (or lacks) an attribute, but we don't care about the value of that attribute. Typically, these are "boolean attributes" like `required` or `disabled`. These attributes trigger certain behaviors in the DOM as long as they are present; their values are irrelevant.

Playwright can try to handle these concerns with `toBe*`. For example, it has `toBeDisabled`. However, there _are_ matchers that Playwright will inevitably miss in the interim. (See #19194 and [JestDom.toBeRequired](https://github.com/testing-library/jest-dom#toberequired)). Even if Playwright does add these missing matchers, it _cannot_ account for [Web Components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components). Currently, if a Web Component exposes a _custom_ boolean attribute that drives its functionality, Playwright has no (proper) way to test this use case.

This makes it important to have a `toHaveAttribute` overload that allows the attribute name to be supplied on its own. The `toHaveAttribute` matcher is already supported in [Jest DOM](https://github.com/testing-library/jest-dom#tohaveattribute). [Cypress](https://docs.cypress.io/api/commands/should#Assert-the-anchor-element-has-href-attribute) also supports this use case.

yury-s:
This time not doing it in other languages due to unjustified generator complexity.

Fixes #27341

### DIFF
--- a/docs/src/api/class-locatorassertions.md
+++ b/docs/src/api/class-locatorassertions.md
@@ -1151,6 +1151,29 @@ Expected attribute value.
 ### option: LocatorAssertions.toHaveAttribute.timeout = %%-csharp-java-python-assertions-timeout-%%
 * since: v1.18
 
+## async method: LocatorAssertions.toHaveAttribute#2
+* since: v1.40
+* langs: js
+
+Ensures the [Locator] points to an element with given attribute. The method will assert attribute
+presence.
+
+```js
+const locator = page.locator('input');
+// Assert attribute existence.
+await expect(locator).toHaveAttribute('disabled');
+await expect(locator).not.toHaveAttribute('open');
+```
+
+### param: LocatorAssertions.toHaveAttribute#2.name
+* since: v1.40
+- `name` <[string]>
+
+Attribute name.
+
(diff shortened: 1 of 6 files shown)


docs/src/api/class-locatorassertions.md
```diff
@@ -1151,6 +1151,29 @@ Expected attribute value.
 ### option: LocatorAssertions.toHaveAttribute.timeout = %%-csharp-java-python-assertions-timeout-%%
 * since: v1.18
 
+## async method: LocatorAssertions.toHaveAttribute#2
```
dgozman, line 1154: Shouldn't we rename `toHaveAttribute` to `toHaveAttribute#1`? I thought that tooling would complain!

dgozman (approved):

mxschmitt (approved):

docs/src/api/class-locatorassertions.md
yury-s: Tried rolling the docs too - everything works fine.

PR #27462, opened 2023-10-05
fix(registry): don't download x64 binaries on other platforms

### LINKED ISSUE #27453: [BUG] Downloads binaries for the wrong architecture
### System info
- Playwright Version: v1.38.1
- Operating System: Raspbian GNU/Linux 12 (bookworm)
- Browser: All
- Other info: process.arch === 'arm'

### Source code

**Steps**

```
yarn create playwright
yarn playwright install
yarn playwright test
```

**Expected**

[Describe expected behavior]

Really, any of these would be reasonable:

- A browser that runs on my machine is downloaded
- An error says "Sorry, we only have binaries for "x64" and "arm64", not "arm"
- Instructions for adjusting launchOptions.executablePath are shown
- The website at https://playwright.dev/docs/intro#system-requirements gives the correct system requirements

**Actual**

[Describe actual behavior]

1. We get a message, "BEWARE: your OS is not officially supported by Playwright; downloading fallback build." As it happens, Raspbian's /etc/os-release contains `ID=raspbian`; changing this to `ID=debian` silences the warning, but the exact same build of chromium is downloaded anyway, so this is a red herring.
2. The downloader inspects os.arch(), sees that it is not arm64, and unpacks 956 MiB of browsers for x64 into ~/.cache/ms-playwright
3. The tests fail, loudly:
   ```
    [pid=15359][err] /home/josh/.cache/ms-playwright/chromium-1080/chrome-linux/chrome: 3: Syntax error: Unterminated quoted string
    [pid=15676][err] /home/josh/.cache/ms-playwright/firefox-1424/firefox/firefox: 4: ������
                                                                                            ��
                                                                                              ��
                                                                                                ��6�

    [pid=15676][err]

    [pid=15676][err]
                     @���DD��
                             ��
                               ��
                                 : not found
    ╔══════════════════════════════════════════════════════╗
    ║ Host system is missing dependencies to run browsers. ║
    ║ Missing libraries:                                   ║
    ║     libGLESv2.so.2                                   ║
    ╚══════════════════════════════════════════════════════╝
    ```
4. An empty file named `"\001\300\335@\230\005\f@8\f@\035\034\006\004@@@\240\002\240\002\b\003\004\340\002\340\002\340\002\034\034\001\001\004\330\301\330\301\020\001\005\320\320\320\375j" appears in my project. For all I know this recklessness is the fault of nodejs or libuv — child_process.execFile seems to ignore { shell: false } — but it's definitely garbage in, garbage out.

jleedev:
On Linux platforms, specifically check that process.arch is x64, rather than treating it as 'not arm64'.

Treat Raspbian's /etc/os-release file as Debian.

Document the supported platforms somewhat.

Fixes #27453


jleedev:

packages/playwright-core/src/utils/hostPlatform.ts
```diff
@@ -68,13 +68,13 @@ export const hostPlatform = ((): HostPlatform => {
         return ('ubuntu20.04' + archSuffix) as HostPlatform;
       return ('ubuntu22.04' + archSuffix) as HostPlatform;
     }
-    if (distroInfo?.id === 'debian' && distroInfo?.version === '11')
+    if ((distroInfo?.id === 'debian' || distroInfo?.id === 'raspbian') && distroInfo?.version === '11')
```
dgozman, line 71: I don't think we are comfortable making this change. We do not test on raspbian and so cannot say that it works. We should treat raspbian the same as any other distribution that is not officially supported.

packages/playwright-core/src/utils/hostPlatform.ts
```diff
@@ -55,7 +55,7 @@ export const hostPlatform = ((): HostPlatform => {
     return macVersion as HostPlatform;
   }
   if (platform === 'linux') {
-    const archSuffix = os.arch() === 'arm64' ? '-arm64' : '';
+    const archSuffix = '-' + os.arch();
```
dgozman, line 58: Looking at https://nodejs.org/api/os.html#osarch, there are quite a few more architectures. We should return `<unknown>` when its not `x64` or `arm64`.

dgozman (commented):
Thank you for the PR! Please see comments inline.

packages/playwright-core/src/utils/hostPlatform.ts
jleedev: For what it's worth, I was imitating the pop/neon/tuxedo above; I assume they have equally valid reasons for altering the ID line in os-release. I'd assume that the firefox build for debian would be preferable to the ubuntu one but have not tested it.

packages/playwright-core/src/utils/hostPlatform.ts
jleedev: This does make the error message worse:

```diff
- Error: ERROR: Playwright does not support chromium on debian12-arm
+ Error: ERROR: Playwright does not support chromium on <unknown>
```

It would be nice to give a clearer error message here, but it's also not great to pretend that "debian12-arm" is a real thing.

jleedev:
Thanks, I've merged that in and added the raspbian case.

It would still be nice to say what, specifically, is `<unknown>`, but the web page spells it out now and it's not worth complicating further.

I notice that [os-release (5)](https://www.man7.org/linux/man-pages/man5/os-release.5.html) documents ID_LIKE. If this is consistently implemented across various ubuntu flavors, it could be more maintainable than listing distributions which are unsupported anyway, e.g. "BEWARE: You are running raspbian on arm64, which claims to be like debian. Downloading fallback binaries for debian-arm64 but this is not officially supported by Playwright."

dgozman:
Thank you for the PR. Looks great, merging in!

> I notice that [os-release (5)](https://www.man7.org/linux/man-pages/man5/os-release.5.html) documents ID_LIKE. If this is consistently implemented across various ubuntu flavors, it could be more maintainable than listing distributions which are unsupported anyway, e.g. "BEWARE: You are running raspbian on arm64, which claims to be like debian. Downloading fallback binaries for debian-arm64 but this is not officially supported by Playwright."

I don't think it's worth it to investigate `ID_LIKE` on multiple distributions just for the nicer error message. Thank you for the pointer though!

dgozman (approved):

PR #27464, opened 2023-10-05
feat(trace): allow navigating from error to source

pavelfeldman:


packages/playwright/src/common/ipc.ts
```diff
@@ -33,7 +33,7 @@ export type ConfigCLIOverrides = {
   timeout?: number;
   ignoreSnapshots?: boolean;
   updateSnapshots?: 'all'|'none'|'missing';
-  workers?: number;
+  workers?: number | string;
```
dgozman, line 36: This looks unrelated!

dgozman (approved):

packages/playwright/src/common/ipc.ts
pavelfeldman: but it is correct!

PR #27496, opened 2023-10-06
chore: remove parsed stack trace

pavelfeldman:


packages/playwright-core/src/utils/stackTrace.ts
```diff
@@ -37,9 +37,7 @@ const internalStackPrefixes = [
 export const addInternalStackPrefix = (prefix: string) => internalStackPrefixes.push(prefix);
 
 export type ParsedStackTrace = {
```
dgozman, line 39: I thought this patch is about removing this type?

packages/playwright-core/src/client/connection.ts
```diff
@@ -98,8 +98,14 @@ export class Connection extends EventEmitter {
...
   }
 
-  pendingProtocolCalls(): ParsedStackTrace[] {
-    return Array.from(this._callbacks.values()).map(callback => callback.stackTrace).filter(Boolean) as ParsedStackTrace[];
+  pendingProtocolCalls(): String {
+    const lines: string[] = [];
+    for (const call of this._callbacks.values()) {
+      if (!call.apiName)
+        continue;
+      lines.push(`  - ${call.apiName}\n${stringifyStackFrames(call.frames)}\n`);
```
dgozman, line 106: Shouldn't this be `call.frames.slice(1)`? I don't think we had the full stack here before.

dgozman (approved):

PR #27511, opened 2023-10-09
docs: note that pytest CLI args only apply to default fixtures

### LINKED ISSUE #27413: Pytest CLI Flags should note their constraints
The list of CLI flags at https://github.com/microsoft/playwright.dev/blob/e8a87d6ac72b27366638d91ce8cc669d2e5868b5/python/docs/test-runners.mdx#L28-L38 does not list any of the preconditions for the described behavior, or the conditions under which the described behavior will not occur.

For example, several of the CLI flags such as `--video` only affect the behavior of the default `context` fixture and appear to be completely inoperative if one is manually creating a context.  This is contrary to the above-linked documentation which implies that `--video on` would, for example, record video in all cases without regard to how the tests are constructed.

dgozman:
Fixes #27413.

### DIFF
--- a/docs/src/test-runners-python.md
+++ b/docs/src/test-runners-python.md
@@ -26,6 +26,9 @@ addopts = --headed --browser firefox
 
 ## CLI arguments
 
+Note that CLI arguments are only applied to the default `browser`, `context` and `page` fixtures.
+If you create a browser, a context or a page with the API call like [`method: Browser.newContext`], the CLI arguments are not applied.
+
 - `--headed`: Run tests in headed mode (default: headless).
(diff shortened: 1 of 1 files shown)


yury-s (approved):

docs/src/test-runners-python.md
```diff
@@ -26,6 +26,9 @@ addopts = --headed --browser firefox
 
 ## CLI arguments
 
+Note that CLI arguments are only applied to the default `browser`, `context` and `page` fixtures.
```
mxschmitt, line 29:
```suggestion
Note that CLI arguments are only applied to the default `context` and `page` fixture.
```

Custom fixtures are not implemented afaik.

mxschmitt (approved):

Issue #18827, opened 2022-11-15
[BUG]


filipesmedeiros:
**Context:**
- Playwright Version: 1.27.1
- Operating System: MacOS 13.0
- Node.js version: 16.15.0
- Browser: All
- Extra: Also CI using MacOS 12 (Github Actions)

**Describe the bug**

Subsequent screenshots (on different test runs) of the same page are coming out with different heights. This is mainly due to a weird "compression" (like smashing) effect on one of the screenshots.

These two example come from the same code, same server (NextJS locally) running, just `yarn test` two times in a row:

![full-page-actual](https://user-images.githubusercontent.com/26208850/202003575-47a4f38d-31f9-4804-9f4e-69520d1e4fc8.png)
![full-page-expected](https://user-images.githubusercontent.com/26208850/202003582-edfa4175-5ba1-449e-84f4-86b26907ae60.png)

I'm sorry for the long image, but that may be part of the problem (although short pages sometimes have this problem too).

If you compare the images using something like https://www.diffchecker.com/image-diff/ you can see the weird effect.

Is this a known bug? The differences are even greater once we run the tests on CI (more screenshots are different), but it's flaky, so sometimes it happens and sometimes everything is good.

Thanks in advance!

pavelfeldman:
This looks like a bug to me. My only idea would be that the actual pixel data and the size of the page are captured in a racy manner and then the image is scaled based on the wrong size. Could you try waiting for layout to settle before you capture an image a little? Also, `await expect.toHaveScreenshot` should wait for the image to stabilize, are you using it?

filipesmedeiros:
@pavelfeldman I used `shouldMatchSnapshot` on all runs, including the first. Is that bad? I know the docs say to use `toHaveScreenshot` but since `shouldMatch` also writes if not present, I assumed it was the same.

I can test with that, though! Thanks for the reply!

filipesmedeiros:
Also, it's supposed to be `expect(await page.screenshot).toMatchScreenshot` right? We don't have to `await expect` like you said, I think (just checking)

pavelfeldman:
We do, that's the point of resilient screenshotting, should be `await expect(page).toHaveScreenshot()`.

filipesmedeiros:
@pavelfeldman  But expect doesn't return a Promise 😕

filipesmedeiros:
[...]

PR #27513, opened 2023-10-09
allow for a configurable amount of size difference between actual and expected images

paulsmithkc:
Proposed fix for #18827

Due to floating point error in both CSS and JS, it isn't practical/possible to ensure exact pixel sizing in the vast ecosystem of tools and frameworks. This changeset allows for a configurable tolerance in the size of the image.


paulsmithkc:
@microsoft-github-policy-service agree company="Byte"

docs/src/api/params.md
```diff
@@ -849,6 +849,12 @@ An acceptable amount of pixels that could be different. Default is configurable
 
 An acceptable ratio of pixels that are different to the total amount of pixels, between `0` and `1`. Default is configurable with `TestConfig.expect`. Unset by default.
 
+## assertions-max-diff-size
+* langs: js
+- `maxDiffSize` <[int]>
+
+An acceptable number of pixels to vary the size by, between `0` and `1`. Default is configurable with `TestConfig.expect`. 1 by default.
```
paulsmithkc, line 856: Are there any additional files that need to be updated to support the new config option?

packages/playwright-core/src/utils/comparators.ts
```diff
@@ -56,13 +56,16 @@ function compareImages(mimeType: string, actualBuffer: Buffer | string, expected
...
+  const size = { width: Math.min(expected.width, actual.width), height: Math.min(expected.height, actual.height) };
   let sizesMismatchError = '';
   if (expected.width !== actual.width || expected.height !== actual.height) {
-    sizesMismatchError = `Expected an image ${expected.width}px by ${expected.height}px, received ${actual.width}px by ${actual.height}px. `;
+    const maxDiffSize = 1 + (options.maxDiffSize ?? 1);
+    if (Math.abs(expected.width - actual.width) >= maxDiffSize || Math.abs(expected.height - actual.height) >= maxDiffSize)
+      sizesMismatchError = `Expected an image ${expected.width}px by ${expected.height}px, received ${actual.width}px by ${actual.height}px. `;
     actual = resizeImage(actual, size);
     expected = resizeImage(expected, size);
   }
```
paulsmithkc, line 67: Allowing for a tolerance level of < (maxDiffSize + 1) to account for floating point error in both CSS and JS.

paulsmithkc:
@maartenbreddels

> ❌ [playwright-test] › golden.spec.ts:93:5 › should compile with different option combinations

What needs to be updated to make the `maxDiffSize` and `maxDiffSizeRatio` be accepted?

tests/playwright-test/to-have-screenshot.spec.ts
```diff
@@ -405,6 +405,8 @@ test('should compile with different option combinations', async ({ runTSC }) =>
             threshold: 0.2,
             maxDiffPixels: 10,
             maxDiffPixelRatio: 0.2,
+            //maxDiffSize: 2,
```
yury-s, line 408: We'll need an actual test for this feature that demonstrates (at least on one platform that we support) where the different size is a legit use case.

docs/src/test-api/class-testconfig.md
```diff
@@ -44,13 +44,17 @@ export default defineConfig({
     - `threshold` ?<[float]> an acceptable perceived color difference between the same pixel in compared images, ranging from `0` (strict) and `1` (lax). `"pixelmatch"` comparator computes color difference in [YIQ color space](https://en.wikipedia.org/wiki/YIQ) and defaults `threshold` value to `0.2`.
     - `maxDiffPixels` ?<[int]> an acceptable amount of pixels that could be different, unset by default.
     - `maxDiffPixelRatio` ?<[float]> an acceptable ratio of pixels that are different to the total amount of pixels, between `0` and `1` , unset by default.
+    - `maxDiffSize` ?<[int]> an acceptable number of pixels to vary the width or height by, `1` by default.
```
yury-s, line 47: Let's make it work with existing `maxDiffPixels` and `maxDiffPixelRatio` options instead of introducing two new parameters. Let's consider those pixels which are missing on the images as different, as if you padded the images (from right and bottom) to make the sizes match.

Let's start with addressing exact rounding error problem. We can allow image dimensions to differ by up to 1px in each direction, if the difference is within specified tolerance thresholds, they will be equal. This should cover (hopefully) common cases where the screenshot differ due to rounding error. So far, we haven't seen pictures of legit cases where the difference would be more than 1px in each dimension.

yury-s (changes requested):

docs/src/test-api/class-testconfig.md
paulsmithkc: Since this change is significantly different, I have started a fresh PR here:
https://github.com/microsoft/playwright/pull/28453

paulsmithkc:
Since the requested change is significantly different, I have started a fresh PR here:
https://github.com/microsoft/playwright/pull/28453

PR #27522, opened 2023-10-10
feat(ct): expose mount result type

### LINKED ISSUE #27509: [Feature] `export` the `MountResult` interface so that it can be used in Page Object Models
Hello!
I am switching from Vitest to Playwright for component testing and I have an issue when writing a page object model because I cannot correctly type the return of `await mount(Page)` in my page object model.

Here is a excerpt from my test:
```typescript
const test = base.extend<{ component: LoginPage }>({
	component: async ({ mount }, use) => {
		const component = new LoginPage(await mount(Page))
		await use(component)
	}
})
```
I try to give the result of `mount` to my page object model, so that it can then abstract the logic for me in the following tests.
Here is part of the page object model:
```typescript
import type { Locator } from '@playwright/test'
import type Page from './+page.svelte'

export class LoginPage {
	public readonly username: Locator

	constructor(public readonly component: MountResult<Page>) {
		this.username = this.component.getByLabel('Username')
	}
}
```

Problem is that I cannot use type `MountResult` as it is not exported. I see two solutions:
1. Recreate the `MountResult` interface
2. Type the `component` as a `Locator` but doing so I would be unable to use the `unmount` and `update` and future added features.

I think that the `MountResult` interface should be exported, as it is part of the API that we manipulate (out of `mount`) and thus we should be able to type what we would store in a variable if needed.

Did I miss something? Is there another solution? If not, then can you please export this interface?

Thanks a lot!

sand4rt:
closes #27509 

CC @dgozman

### DIFF
--- a/packages/playwright-ct-react/index.d.ts
+++ b/packages/playwright-ct-react/index.d.ts
@@ -39,7 +39,7 @@ export interface MountOptions<HooksConfig extends JsonObject> {
   hooksConfig?: HooksConfig;
 }
 
-interface MountResult extends Locator {
+export interface MountResult extends Locator {
   unmount(): Promise<void>;
   update(component: JSX.Element): Promise<void>;
 }
--- a/packages/playwright-ct-react/register.d.ts
+++ b/packages/playwright-ct-react/register.d.ts
@@ -14,4 +14,4 @@
  * limitations under the License.
  */
 
-export default function pwRegister(components: Record<string, any>): void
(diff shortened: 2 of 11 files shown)


dgozman:
@sand4rt Thank you for the PR! We are not really sure whether we should commit to exposing this type, given how types in components experiments have been rapidly changing. Given the workaround in the issue, let's wait for types to settle before revisiting this.

sand4rt:
Oh alright, yeah makes sense

Issue #15874, opened 2022-07-22
[BUG] codegen stalls browser for some sites, even the close button is not working anymore


JannisBush:
**Context:**
- Playwright Version: 1.23.0
- Operating System: Linux, Mac
- Browser: Chromium, (Firefox)
- Extra: codegen

**Code Snippet**
Run `playwright codegen cloudflare.com`

**Describe the bug**

Wait for the site to load, then click outside of the cookie banner.
After that you cannot click anywhere anymore, only scrolling still works. 
Even the close button of the browser is not working anymore (minimize and maximize still work).
In Firefox, the close button still works but only with a double click. 
WebKit does not seem to be affected.

This is the only site where I have seen a complete stall of the browser, however on other sites I had additional issues with codegen that some elements could not be clicked or only a double click on them worked which might be related to this bug.

clehene:
@dgozman
Curious what caused the actual stall behavior and how to debug it. 
We're seeing this across all browser for a particular webpage (can't share) but could share some relavant details if I know how

PR #27532, opened 2023-10-10
fix: noscript sections visibility and text selection 

### LINKED ISSUE #27504: [BUG] 'noscript' sections always invisible in ui-mode
### System info
- Playwright Version: [v1.38.1]
- Operating System: [debian linux]
- Browser: [All]
- Other info:

### Source code

https://github.com/mash-graz/playwright-noscript-issue

**Steps**

run the example in the provided git repo in the ui-mode viewer

**Expected**

You will see, that the visual output of the `<noscript>...</noscript>` are not visible in the ui-mode viewer, although the tests look plausible and in VSCodium we even get correct visual results.

**Actual**

NoScript sections of html-code are never rendered in playwrights ui-mode.

As attachments you'll find screenshots of ui-mode rendering vs. vscodium usage.

Screenshot-ui-viewer
Screenshot-vscodium

mash-graz:
This PR fixes both issues reported in #27504

It makes sections within noscript tags visible in ui-mode and makes the contained text content selectable again.

To realize this 2nd goal, I had to revert #15887. Otherwise we can't use `getByText()` selectors to catch noscript wrapped text messages. I hope, this revert will not cause the troubles mentioned in #15874 again. If it's still the case, we have to look for other solutions to solve the issue there by more specific means. 

Fixes:  #27504


packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
```diff
@@ -337,8 +336,6 @@ export function frameSnapshotStreamer(snapshotStreamer: string) {
           if (rel === 'preload' || rel === 'prefetch')
             return;
         }
-        if (this._removeNoScript && nodeName === 'NOSCRIPT')
```
dgozman, line 340: I don't think this will help by itself. Trace viewer does not render the snapshot with javascript disabled, so the browser will not show the `noscript` tag contents.

packages/playwright-core/src/server/injected/selectorUtils.ts
```diff
@@ -53,7 +53,7 @@ export function matchesAttributePart(value: any, attr: AttributeSelectorPart) {
 
 export function shouldSkipForTextMatching(element: Element | ShadowRoot) {
   const document = element.ownerDocument;
-  return element.nodeName === 'SCRIPT' || element.nodeName === 'NOSCRIPT' || element.nodeName === 'STYLE' || document.head && document.head.contains(element);
+  return element.nodeName === 'SCRIPT' || element.nodeName === 'STYLE' || document.head && document.head.contains(element);
```
dgozman, line 56: This goes against the idea of text selector - it only accounts for text that is for the end user. That's why it ignores things like `<script>` or `<noscript>`. I also think changing this today will break existing tests, so we should not do that.

If anything, we can (maybe??) do this when javascript is disabled. However, that's some extra plumbing.

---

I'd like to step back and figure out why would you want selecting by text in the `<noscript>` element? If you'd like to test that your page has proper `<noscript>` support, just do something like:

```ts
await expect(page.locator('noscript')).toHaveText('....');
```

This way you rely on the browser to handle `noscript` as it would. There is no need to test that browser shows `noscript` content when it should - that's better to leave for the browser developers.

dgozman (commented):
Thank you for the PR! See inline comments.

packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
mash-graz: > Trace viewer does not render the snapshot with javascript disabled, so the browser will not show the noscript tag contents.

If you look at the actual test results, you'll immediately see, that it seems to work in practice! :)

well, the semantic of the `noscript` tag is rather complex, and I don't know, if this very simple modification catches all the possible corner cases, but at least for very basic usage scenarios it seems to work.

dgozman: Well, I tried it myself, and in the trace viewer I see (despite `<noscript>` tag present):
```txt
Here you should see a message, if JS was dissabled durring this Test:

""
This output should look identical:

"JS Disabled (2)"
```

while during test execution I see:

```txt
Here you should see a message, if JS was dissabled durring this Test:

"JS Disabled (1)"
This output should look identical:

"JS Disabled (2)"
```

So, I don't think it helps meaning that trace viewer will not show the snapshot as it was visible during the test itself.

packages/playwright-core/src/server/injected/selectorUtils.ts
mash-graz: > This goes against the idea of text selector - it only accounts for text that is for the end user. That's why it ignores things like <script> or <noscript>.

This very simple "Please enable JS in your browser" messages displayed by nearly any JS metaframeworks is very simple example for a readable text message targeting end users!

> I also think changing this today will break existing tests, ...

The present behavior of playwright doesn't represent the actual behavior of real world browsers.
If we fix this flaw, it will definitely affect some (but in fact **very few**) existing tests.

But if we don't fix this incorrect behavior, we are not able to reliable tests those cases at all -- all those cases, where the mentioned changes really make a significant difference worth testing.

But to argue a little bit more realistic:  
Do you see any negative side effects or unforeseen consequences affecting any of the rather wide spectra of practial applications represented by playwrights own integration tests?

> I'd like to step back and figure out why would you want selecting by text in the <noscript> element?

I just want't to test, that the output seen by end users looks correct.

Using a text selection is the simplest form of a test to archive this goal. It should simple work as expected, because it would highly irritate all playwright users, if it doesn't work for this kind of visible output in browsers just like in all other cases.

And it's also important to fix this kind of selection facility, because otherwise you have to modify the output (add test IDs etc.) to locate the relevant spans. It therefore enables minimal intrusive kind of observation, which do not need any additional preparation or modification of the analyzed object.

> If you'd like to test that your page has proper <noscript> support, just do something like: ...

I don't think, your suggested alternative will work.

The `<noscript>` tag has rather uncommon semantic specified by the html standard.
If JS is enabled, you will not even find it in the DOM hierachie, but otherwise (=JS is present) it's simple showing all the included text or html code fragment in an utterly transparent manner.

Please simply take a look at the output of the included test case and the actual change in behavior of playwright caused by this fix. That's hopefully more convincing than theoretical debates.

packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts
mash-graz: If my fix isn't enabled and JS disabled, you see the fist output, otherweise the second one.  
if JS is enabled, both strings are empty.

All browser handle this kind of test in a correct manner. They even send correct results to the tracer! It's really only the incorrect code in playright, which is responsible for the wrong results.

mash-graz: @dgozman 
I just looked at the test myself again to meditate about your claim.

what i saw:
If hover over the timeline, i get the correct result resp. the expected output after applying my patch and somthing that corresponds very well to the test code results.
**but**
the main view of the tracer still looks different! :(

Screenshot from 2023-10-13 01-51-06
Screenshot from 2023-10-13 01-50-29

dgozman: > If hover over the timeline, I get the correct result resp. the expected output after applying my patch and something that corresponds very well to the test code results.

Yep, that's what the browser renders.

> the main view of the tracer still looks different! :(

Exactly! This is rendered in the trace viewer from the snapshot that you are changing. That's what I mean by "this change will not help the trace viewer", because it will effectively ignore the `<noscript>` entirely.

mash-graz: I send you a working patch soon.

but I still have to find a nice way to transfer the javascriptEnabled context state into this method.

And I additionally had to mask Noscript tags in the trace viewer. Otherwise they will be interpreted by the display mechanism there again.

mash-graz:
Here is an improved version of the fix, which should fix the task viewers html display as well.

I really don't like the hack, how I set the `_removeNoScript` variable. :(
If you see a more elegant solution for this task, please change it!

packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -66,6 +66,9 @@ export class SnapshotRenderer {
         } else if (typeof n[0] === 'string') {
           // Element node.
           const builder: string[] = [];
+          // mask noscript tags in trace-viewers HTML display
+          if (n[0] === 'NOSCRIPT')
```
dgozman, line 70: Instead of post-processing, I think we should just [write a different `nodeName`](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/trace/recorder/snapshotterInjected.ts#L388) into tracing right away, because that's the only place where we know whether `noscript` should be rendered or not.

```ts
const result: NodeSnapshot = [nodeName === 'NOSCRIPT' ? 'MASKED-NOSCRIPT' : nodeName, attrs];
```

packages/playwright-core/src/server/trace/recorder/snapshotter.ts
```diff
@@ -105,8 +105,13 @@ export class Snapshotter {
   }
 
   async captureSnapshot(page: Page, callId: string, snapshotName: string, element?: ElementHandle): Promise<void> {
+
+    const noscript_remove = this._context._options.javaScriptEnabled
```
dgozman, line 109: We can add a second parameter to [`frameSnapshotStreamer` here](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/trace/recorder/snapshotter.ts#L89) to make it more straightforward.

```ts
    const removeNoScript = !!this._context._options.javaScriptEnabled;
    const initScript = `(${frameSnapshotStreamer})("${this._snapshotStreamer}", ${removeNoScript})`;
```

packages/playwright-core/src/server/trace/recorder/snapshotter.ts
```diff
@@ -105,8 +105,13 @@ export class Snapshotter {
...
+      ? `window["${this._snapshotStreamer}"]._removeNoScript = true;`
+      : `window["${this._snapshotStreamer}"]._removeNoScript = false;`;
+
     // Prepare expression synchronously.
-    const expression = `window["${this._snapshotStreamer}"].captureSnapshot(${JSON.stringify(snapshotName)})`;
+    const expression = `${noscript_remove} window["${this._snapshotStreamer}"].captureSnapshot(${JSON.stringify(snapshotName)})`;
```
dgozman, line 114: Let's add a test for this functionality in the trace. Take a look at this one for expiration: https://github.com/microsoft/playwright/blob/main/tests/library/trace-viewer.spec.ts#L656-L668. Basically, configure context, perform a `goto` action, and check that text from the `noscript` tag is visible in the snapshot.

You will need `test.use({ javaScriptEnabled: false })` similarly to the test in this PR. Let me know if you have any troubles with writing a test, or any other questions.

dgozman (commented):
@mash-graz Sorry for being late, I missed your update last time. I think this PR is great for fixing the `noscript` issue in the trace viewer. See inline comments.

Let's separate text selector changes in a separate PR. I have big concerns around backwards compatibility, so I'd recommend to address that separately. This way we'll be able to quickly fix the tracing issue.

mash-graz:
> @mash-graz Sorry for being late, I missed your update last time. I think this PR is great for fixing the `noscript` issue in the trace viewer. See inline comments.

That's o.k. -- There are indeed much more important issues to fix! ;)

> Let's separate text selector changes in a separate PR. I have big concerns around backwards compatibility, so I'd recommend to address that separately. This way we'll be able to quickly fix the tracing issue.

Yes – The visibility in the trace viewer and the access to the related text spans are in indeed two different issues.
But the confusion caused by these flaws manly stems from obvious **inconsistent** results. 
We should really fix them in a consequent manner:

*  the visual results in trace viewer should be the same as in any ordinary browser 
*  and the visible text should be accessible to capturing just as any other output on screen

That's why both of them should be cured together, because otherwise we just have other inconsistencies again.

Concerning the backward compatibility, I would really like to see at least one practical example, where this fix interferes with correct test results.

packages/trace-viewer/src/snapshotRenderer.ts
mash-graz: Hmmm -- I did it in the trace viewer, because it's a flaw caused by the used display type. if we wouldn't use HTML-widgets for the output of the report, everything would be fine and perfectly undistorted without this tricky workaround. So it's IMHO indeed mainly a viewer issue and should be handled there.

packages/playwright-core/src/server/trace/recorder/snapshotter.ts
mash-graz: Yes -- such a modification would be very useful!

But this change should be better written by someone much more familiar with this code base.
I really couldn't grasp all lifecycle concerns and context isolation efforts hidden behind some strange looking passages of the sources.

packages/playwright-core/src/server/trace/recorder/snapshotter.ts
mash-graz: I think, it should be already covered by the included tests.

But the correctness of the pure visual text representation on screen is hard to prove.
That's why we should look for a consistent behavior of display and text-selection fixes.

PR #27540, opened 2023-10-11
chore: unify target closed errors

pavelfeldman:


examples/todomvc/playwright.config.ts
```diff
@@ -10,15 +10,15 @@ export default defineConfig({
   testDir: './tests',
 
   /* Maximum time one test can run for. */
-  timeout: 30 * 1000,
+  timeout: 5_000,
```
dgozman, line 13: Seems pretty short!

packages/playwright-core/src/client/browserType.ts
```diff
@@ -189,7 +188,7 @@ export class BrowserType extends ChannelOwner<channels.BrowserTypeChannel> imple
         return result.result;
       } else {
         closePipe();
-        throw new Error(`Timeout ${params.timeout}ms exceeded`);
+        throw new Error(`browserType.connect: Timeout ${params.timeout}ms exceeded`);
```
dgozman, line 191: Shouldn't the `wrapApiCall` add this prefix automatically?

packages/playwright-core/src/client/connection.ts
```diff
@@ -110,8 +110,8 @@ export class Connection extends EventEmitter {
   }
 
   async sendMessageToServer(object: ChannelOwner, method: string, params: any, apiName: string | undefined, frames: channels.StackFrame[], wallTime: number | undefined): Promise<any> {
-    if (this._closedErrorMessage)
-      throw new Error(this._closedErrorMessage);
+    if (this._closedError)
+      throw this._closedError;
```
dgozman, line 114: This has an effect of giving the stack of the error rather than the stack of this call. Not sure this is intentional.

packages/playwright-core/src/server/dispatchers/dispatcher.ts
```diff
@@ -324,20 +323,15 @@ export class DispatcherConnection {
...
       callMetadata.result = validator(result, '', { tChannelImpl: this._tChannelImplToWire.bind(this), binary: this._isLocal ? 'buffer' : 'toBase64' });
     } catch (e) {
-      // Dispatching error
-      // We want original, unmodified error in metadata.
       callMetadata.error = serializeError(e);
-      if (callMetadata.log.length)
-        rewriteErrorMessage(e, e.message + formatLogRecording(callMetadata.log));
-      error = serializeError(e);
+      if (callMetadata.error.error)
+        callMetadata.error.error.log = callMetadata.log;
```
dgozman, line 334: Are we going to have two copies of the `log` in the trace?

packages/playwright-core/src/utils/index.ts
```diff
@@ -32,6 +32,7 @@ export * from './network';
 export * from './processLauncher';
 export * from './profiler';
 export * from './rtti';
+export * from '../protocol/serializers';
```
dgozman, line 35: I would think that check-deps complains about this.

packages/playwright-core/src/utils/manualPromise.ts
```diff
@@ -72,9 +72,10 @@ export class LongStandingScope {
   close(errorMessage: string) {
     this._isClosed = true;
     this._terminateErrorMessage = errorMessage;
-    for (const [p, e] of this._terminatePromises) {
-      rewriteErrorMessage(e, errorMessage);
-      p.resolve(e);
+    for (const [p, frames] of this._terminatePromises) {
+      const error = new Error(errorMessage);
+      error.stack = [errorMessage, ...frames].join('\n');
```
dgozman, line 77:
```suggestion
      error.stack = [error.name + ': ' + errorMessage, ...frames].join('\n');
```

tests/library/browsertype-connect.spec.ts
```diff
@@ -549,13 +550,11 @@ for (const kind of ['launchServer', 'run-server'] as const) {
       await disconnectedPromise;
       expect(browser.isConnected()).toBe(false);
 
-      const navMessage = (await navigationPromise).message;
-      expect(navMessage).toContain('Connection closed');
-      expect(navMessage).toContain('Closed by');
-      expect(navMessage).toContain(__filename);
```
dgozman, line 555: Why drop this line? Are missing the stack now?

dgozman (approved):

examples/todomvc/playwright.config.ts
pavelfeldman: I figured it is good for the example.

packages/playwright-core/src/client/browserType.ts
pavelfeldman: Hm, good question, let me investigate. I thought it was a pure client-side call site w/o wrap.

packages/playwright-core/src/client/connection.ts
pavelfeldman: As this call is always wrapped, stack will be rewritten.

tests/library/browsertype-connect.spec.ts
pavelfeldman: Resolved offline

dgozman (approved):

PR #27554, opened 2023-10-11
chore: Add flaky tests count and tags to junit reporter

Philip-Scott:
Adds `flaky` as a metric to be reported within the JUnit reporter. 

Motivation: Our team is working on tracking down intermittent test errors on our CI pipelines. We are currently using the JUnit reports to monitor the number of disabled tests (Both Jest & Playwright) within our repo, and we would like to use these files to track flaky tests as well. 

Playwright was already tracking these on the individual test outcome itself, this just exposes these values to the report.

I do want to note that this is not part of the official JUnit spec. But as this is an xml file, extra properties should just be ignored by parsers: https://github.com/testmoapp/junitxml

### DIFF
--- a/packages/playwright/src/reporters/junit.ts
+++ b/packages/playwright/src/reporters/junit.ts
@@ -29,6 +29,7 @@ class JUnitReporter extends EmptyReporter {
   private startTime!: number;
   private totalTests = 0;
   private totalFailures = 0;
+  private totalFlaky = 0;
   private totalSkipped = 0;
   private outputFile: string | undefined;
   private resolvedOutputFile: string | undefined;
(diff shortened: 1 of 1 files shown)


dgozman:
@Philip-Scott Could you please file an issue first? We'd love our PRs to be linked to an issue for easier search later.

Also, a test for this would be essential. Take a look at [this file](https://github.com/microsoft/playwright/blob/main/tests/playwright-test/reporter-junit.spec.ts) for inspiration. You can run these tests with `npm run ttest` (note double t).

If you could post an example of the junit.xml that's produced with this change, that would be ideal.

yury-s:
Closing this for now per the comment above. Let's start with an issue.

PR #27588, opened 2023-10-13
feat(webkit): roll to r1925

playwrightmachine:


yury-s:
`should reject if launched browser fails immediately
library\browsertype-launch.spec.ts:62` is an existing failure:

image

yury-s (approved):

Issue #27339, opened 2023-09-28
CORS trickery should at least be configurable.


fd:
I understand most issues about CORS in this repo are about people trying to ignore it. I also understand 
playwright tries to help by fixing up some common mistakes people make. But I really do want to test CORS and the following code 
makes that pretty much impossible. Can we have an options or flag somewhere to disable this?

https://github.com/microsoft/playwright/blob/28924678397cb01001096fa4c2df0d4b46b73dc3/packages/playwright-core/src/server/network.ts#L299-L314

yury-s:
Can you attach an example of the test code that you want to  work and which is broken by this logic? This will help us evaluate the request.

romainmenke:
We use `playwright` to verify that an API actually works as intended when called from a browser.

Part of these tests is a CORS check.
This test (`test CORS`) works fine.

But we also have a sanity check (`self test`) to make sure that the CORS check actually fails when the CORS related headers are omitted. This second test does the exact same thing as the first test but intercepts the response and erases the CORS related headers.

This gives us the guarantee that both tests actually did something that interacted with the API and that this required CORS related headers to be present.

However the second test does not throw an error.
When using `route.fulfill` CORS headers will always be inserted by `playwright`.
This makes it impossible to simulate an API/route that does not support CORS.

---------

Test code :

```js
test('test CORS', async ({ page }) => {
	await page.goto('http://127.0.0.1:4322/cors/test-cors.html');

	const err = await page.evaluate(async () => {
		try {
			await window.runTest();
		} catch (err) {
			return err;
		}
	});

	expect(err).toBeUndefined();
});

test('self test', async ({ page }) => {
	// Disable CORS.
	await page.route('**/graphql', async route => {
		// Fetch original response.
		const response = await route.fetch();
		const headers = response.headers();

		// Delete CORS headers.
		delete headers['access-control-request-headers'];
		delete headers['access-control-request-method'];
		delete headers['access-control-allow-origin'];

		route.fulfill({ response, headers });
	});

	await page.goto('http://127.0.0.1:4322/cors/test-cors.html');

	const err = await page.evaluate(async () => {
		try {
			await window.runTest();
		} catch (err) {
			return err;
		}
	});

	expect(err).toContain('Failed to fetch');
});
```

```js
// `window.runTest()` is roughly equivalent to :

window.runTest = async function runTest() {
	await fetch(`http://localhost:4321/graphql`, {
		"body": "{}",
		"method": "POST"
	});
}
```

[...]

Issue #2641, opened 2020-06-19
[BUG] Mocking CORS request does not work


flaurida:
**Context:**
- Playwright Version: 1.0.1
- Operating System: Ubuntu 20.04 under WSL2
- Node version: 14.4 (also tried with 12)
- Browser: Chromium, Firefox

**Code Snippet**

This does not work (POST):

```javascript
const playwright = require("playwright");
(async () => {
  for (const browserType of ["chromium", "webkit", "firefox"]) {
    const browser = await playwright[browserType].launch();
    const page = await browser.newPage();
    page.route("**/cars?**", async (route) => {
      route.fulfill({
        contentType: "application/json",
        headers: { "access-control-allow-origin": "*" },
        status: 200,
        body: JSON.stringify(["BMW", "Tesla"]),
      });
    });
    const resp = await page.evaluate(
      async () =>
        await (
          await fetch("https://example.com/cars?type=foobar", {
            method: "POST",
            headers: { "Content-Type": "application/json" },
            mode: "cors",
            body: JSON.stringify({ 'foo': 1 }) 
          })
        ).json()
    );
    console.log(resp);
    await browser.close();
  }
})();
```

This does work (GET):

```javascript
const playwright = require("playwright");
(async () => {
  for (const browserType of ["chromium", "webkit", "firefox"]) {
    const browser = await playwright[browserType].launch();
    const page = await browser.newPage();
    page.route("**/cars?**", async (route) => {
      route.fulfill({
        contentType: "application/json",
        headers: { "access-control-allow-origin": "*" },
        status: 200,
        body: JSON.stringify(["BMW", "Tesla"]),
      });
    });
    const resp = await page.evaluate(
      async () =>
        await (
          await fetch("https://example.com/cars?type=foobar", {
            method: "GET"
          })
        ).json()
    );
    console.log(resp);
    await browser.close();
  }
})();
```

**Describe the bug**

@pavelfeldman mentioned this in Slack a few days ago, following up here. Thank you 🙏

pavelfeldman:
I can only repro this on Chromium after I add a couple of missing awaits:

```js
const playwright = require("playwright");
(async () => {
  for (const browserType of ["chromium", "webkit", "firefox"]) {
    const browser = await playwright[browserType].launch();
    console.log(browserType.normalize());
    const page = await browser.newPage();
    await page.route("**/cars?**", async (route) => {
      await route.fulfill({
        contentType: "application/json",
[...]

PR #27591, opened 2023-10-13
[bugfix] Do not ignore CORS violations

pboos:
Remove CORS trickery that was added as part of [this bug ticket](https://github.com/microsoft/playwright/issues/2641).
In my opinion having such trickery is bad practice and we already had 2 incidents that should would have been detected if this trickery wasn't there (luckily Testcafe still detected them as we have not migrated all tests to playwright yet). In both incidents a header was added accidentally which caused CORS violations due to the backend not having these allowed in the CORS headers. This should be detected in e2e tests. Also in Testcafe CORS errors are not ignored if one adds request hooks.

Reasons in my opinion why this CORS trickery is bad:
- This hides real problems (as we learned the hard way)
- It is inconsistent over browsers, currently only chrome has this trickery. In Firefox (and maybe others) the CORS errors appear.
- If a CORS error appears in the test (as in the bug ticket) then this is a VALID error! It should not be disabled. If one wants to get rid of it there are other ways, but this should not be enforced for everyone using the e2e framework. The way those CORS errors can be ignored if they really want to:
    - They setup a proxy that routes the requests and there disables CORS
    - If the problem is that they run the tests against localhost instead of production -> run local server with self signed ssl certificate, add that certificate to be accepted on the host that runs the tests, adjust hosts to point the production url to 127.0.0.1

See issue: https://github.com/microsoft/playwright/issues/27339


pboos:
@microsoft-github-policy-service agree [company="GetYourGuide AG"]

pboos:
@microsoft-github-policy-service agree company="GetYourGuide AG"

pboos:
```
@microsoft-github-policy-service agree company="GetYourGuide AG"
```

pboos:
@microsoft-github-policy-service agree company="getyourguide"

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -255,30 +255,6 @@ export class CRNetworkManager {
       frame = this._page._frameManager.frameAttached(requestWillBeSentEvent.frameId, null);
     }
 
-    // CORS options preflight request is generated by the network stack. If interception is enabled,
-    // we accept all CORS options, assuming that this was intended when setting route.
-    //
-    // Note: it would be better to match the URL against interception patterns.
```
yury-s, line 261: Let's implement this url matching instead. That way you would fulfill all the analytics requests that you mentioned without going to the server while keeping the rest unaffected by the interception. And then all the tests (and the use cases they cover) that broke will keep working.

yury-s (changes requested):

yury-s:
Closing as part of PR cleanup. We are not ready to accept it in the current form as it will break too many existing clients. Let's continue discussion in the bug.

PR #27616, opened 2023-10-15
Add function to add items to localStorage when page initializes

sjarmero:
For applications that use `localStorage` heavily, it is useful to be able to initialize the state of `localStorage` in certain tests, before navigating to any page. This MR adds a method, `addInitLocalStorageItems`, that provides a way of setting `localStorage` items using `addInitScript`. It also supports an `overwrite` flag. Use this to override any `localStorage` item that was already set through the browser context.

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -430,6 +430,27 @@ Path to the JavaScript file. If `path` is a relative path, then it is resolved r
 
 Script to be evaluated in all pages in the browser context. Optional.
 
+## async method: BrowserContext.addInitLocalStorageItems
+* since: v1.8
+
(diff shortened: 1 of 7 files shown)


mxschmitt:
Hi Sergio, thank you for your contribution! Could you please start filing a feature request for it?

sjarmero:
> Hi Sergio, thank you for your contribution! Could you please start filing a feature request for it?

Done! Available here: https://github.com/microsoft/playwright/issues/27654

sjarmero:

yury-s:
Closing as there are already ways to initialize local storage by either loading storage state or creating init script as described in the bug. This API would be redundant and not justified for such a niche use case.

PR #27640, opened 2023-10-16
chore: plumb the target close reason when test fails

pavelfeldman:


packages/playwright-core/src/client/browser.ts
```diff
@@ -130,12 +130,12 @@ export class Browser extends ChannelOwner<channels.BrowserChannel> implements ap
     return buffer;
   }
 
-  async close(): Promise<void> {
+  async close(options: { reason?: string } = {}): Promise<void> {
     try {
       if (this._shouldCloseConnectionOnClose)
         this._connection.close();
```
dgozman, line 136: Might as well plumb it here too.

packages/playwright-core/src/server/fetch.ts
```diff
@@ -91,6 +91,7 @@ export abstract class APIRequestContext extends SdkObject {
   readonly fetchLog: Map<string, string[]> = new Map();
   protected static allInstances: Set<APIRequestContext> = new Set();
   readonly _activeProgressControllers = new Set<ProgressController>();
+  _closeReason: string | undefined;
```
dgozman, line 94: Plumb this one in a follow up?

packages/playwright/src/index.ts
```diff
@@ -374,7 +375,7 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
     const context = await (browser as any)._newContextForReuse(defaultContextOptions);
     (context as any)[kIsReusedContext] = true;
     await use(context);
-    await (browser as any)._stopPendingOperations('Test ended');
+    await (browser as any)._stopPendingOperations('Test ended.');
```
dgozman, line 378: This one could also be `Test timeout of ...`

dgozman (approved):

packages/playwright-core/src/client/browser.ts
pavelfeldman: I'll handle client side separately.

packages/playwright-core/src/server/fetch.ts
pavelfeldman: Can do.

PR #27669, opened 2023-10-17
chore: remove the usages of raw target closed message constant

pavelfeldman:


packages/playwright-core/src/common/errors.ts
```diff
@@ -25,16 +25,13 @@ class CustomError extends Error {
 
 export class TimeoutError extends CustomError {}
 
-export const kTargetClosedErrorMessage = 'Target page, context or browser has been closed';
-export const kTargetCrashedErrorMessage = 'Target crashed';
-
 export class TargetClosedError extends Error {
```
dgozman, line 28: Should this be `extends CustomError` similarly to `TimeoutError`?

packages/playwright-core/src/common/errors.ts
```diff
@@ -25,16 +25,13 @@ class CustomError extends Error {
...
-    super(kTargetClosedErrorMessage);
+  constructor(cause?: string, logs?: string) {
+    super((cause || 'Target page, context or browser has been closed') + (logs || ''));
     this.name = this.constructor.name;
   }
 }
 
 export function isTargetClosedError(error: Error) {
-  return error instanceof TargetClosedError || error.message.includes(kTargetClosedErrorMessage);
+  return error instanceof TargetClosedError || error.name === 'TargetClosedError';
```
dgozman, line 36: Given the change in `serializers.ts`, I wonder when does `error.name ===` code path trigger?

packages/playwright-core/src/utils/manualPromise.ts
```diff
@@ -99,8 +101,8 @@ export class LongStandingScope {
     const terminatePromise = new ManualPromise<Error>();
     if (this._terminateError)
       terminatePromise.resolve(this._terminateError);
-    if (this._terminateErrorMessage)
-      terminatePromise.resolve(new Error(this._terminateErrorMessage));
+    if (this._closeError)
+      terminatePromise.resolve(this._closeError);
```
dgozman, line 105: Ideally, this would be the same `errorClone` as above, to preserve `frames`.

dgozman (approved):

packages/playwright-core/src/common/errors.ts
pavelfeldman: I have no idea!

packages/playwright-core/src/common/errors.ts
pavelfeldman: That is for cloned errors.

PR #27671, opened 2023-10-17
fix: preserve lastModified timestamp in setInputFiles

### LINKED ISSUE #27452: [BUG] File.lastModified behaviour different between Chromium and WebKit
### System info
- Playwright Version: [v1.38.1]
- Operating System: [macOS 13.2]
- Browser: [Chromium, WebKit]
- Other info:

### Source code

This is a full test that fails on WebKit and succeeds on Chromium:
```typescript
import test, { expect } from '@playwright/test';

test('File.lastModified', async ({ page }) => {
    await page.setContent(`
        <input type="file" id="file-input" />
    `);

    const fileInput = await page.$('#file-input');

    // Get the file handle
    const [fileHandle] = await Promise.all([
        page.waitForEvent('filechooser'),
        fileInput.click(),
    ]);

    await fileHandle.setFiles(['playwright/assets/images/image_001.jpeg']);

    const [t0, t1, t2] = await fileInput.evaluate(
        async (el: HTMLInputElement) => {
            const selectedFile = el.files[0];

            return new Promise<[number, number, number]>(resolve => {
                const initial = selectedFile.lastModified;
                let after100: number;
                let after200: number;

                setTimeout(() => {
                    after100 = selectedFile.lastModified;
                }, 100);

                setTimeout(() => {
                    after200 = selectedFile.lastModified;
                    resolve([initial, after100, after200]);
                }, 200);
            });
        }
    );

    expect(t0).toBe(t1);
    expect(t1).toBe(t2);
});
```

Hi!

First of all, thank you for building this amazing piece of software!

In the last month, I worked on transitioning from Cypress to Playwright as e2e testing solution.

During a specific flow, I need to upload some images. The application will then generate an md5 hash from `${file.name}:${file.lastModified}:${file.size}`, at different times and in multiple parts of the codebase. The code rightly assumes that the file details won't change in between queries and that multiple calls to the hashing function with the same file will produce the same result.

I am using the following in Playwright to select the files to upload:
```
const [fileChooser] = await Promise.all([
  page.waitForEvent('filechooser'),
  this._selectPhotos.click(page),
]);

await fileChooser.setFiles(photoPaths);
```

Unfortunately, the `lastModified` value of the files is not stable and the behaviour varies between browsers:
- in a Chromium profile, `lastModified` will return the time the file was selected, and the value never changes from that
- in a WebKit profile, `lastModified` returns the current time, every time the property is accessed, meaning that the value is not stable and our hashing function becomes invalid

I also tested the application in Safari, thinking that it was a browser-specific issue, but it was not the case - the `lastModified` has a stable value. The only scenario in which the `lastModified` value changes is when running a WebKit profile from Playwright.

**Steps**
- [Select a couple of files]
- [Access `lastModified` a couple of times (eg. after 50ms, 100ms, 200ms)]
- [Note that the value differs every time it is accessed]

**Expected**

I am expecting the behaviour of accessing `File.lastModified` to be the same regardless of the browser - using the value of when the file was chosen seems sensible enough.

**Actual**

Currently, running a WebKit profile in Playwright makes `File.lastModified` return the equivalent of `Date.now()` on each call.

yury-s:
Fixes #27452

### DIFF
--- a/packages/playwright-core/src/client/elementHandle.ts
+++ b/packages/playwright-core/src/client/elementHandle.ts
@@ -152,9 +152,10 @@ export class ElementHandle<T extends Node = Node> extends JSHandle<T> implements
       throw new Error('Cannot set input files to detached element');
     const converted = await convertInputFiles(files, frame.page().context());
     if (converted.files) {
+      debugLogger.log('api', 'setting input buffers');
       await this._elementChannel.setInputFiles({ files: converted.files, ...options });
     } else {
-      debugLogger.log('api', 'switching to large files mode');
+      debugLogger.log('api', 'setting input file paths');
       await this._elementChannel.setInputFilePaths({ ...converted, ...options });
     }
   }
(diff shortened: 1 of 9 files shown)


packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -265,18 +266,14 @@ type InputFilesList = {
...
-
-  const stats = await Promise.all(items.filter(isString).map(item => fs.promises.stat(item as string)));
-  const totalFileSizeExceedsLimit = stats.reduce((acc, stat) => acc + stat.size, 0) > sizeLimit;
-  if (totalFileSizeExceedsLimit) {
+  if (items.some(item => typeof item === 'string')) {
+    if (!items.every(item => typeof item === 'string'))
+      throw new Error('File paths cannot be mixed with buffers');
     if (context._connection.isRemote()) {
       const streams: channels.WritableStreamChannel[] = await Promise.all(items.map(async item => {
         assert(isString(item));
```
dgozman, line 274: This assert is not needed anymore.

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -286,19 +283,17 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
...
-      };
-    } else {
-      return {
-        name: item.name,
-        mimeType: item.mimeType,
-        buffer: item.buffer,
-      };
-    }
+  const sizeLimit = 50 * 1024 * 1024;
+  const totalBufferSizeExceedsLimit = items.reduce((size, item) => size + ((typeof item === 'object' && item.buffer) ? item.buffer.byteLength : 0), 0) > sizeLimit;
```
dgozman, line 287: If you cast `items` to `FilePayload[]` before this line, there is no need for `typeof` and other checks.

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -286,19 +283,17 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
...
+  if (totalBufferSizeExceedsLimit)
+    throw new Error('Cannot set buffer larger than 50Mb, please write it to a file and pass its path instead.');
+
+  const filePayloads: SetInputFilesFiles = await Promise.all((items as FilePayload[]).map(async item => {
```
dgozman, line 291: No need for promises or any conversions here.

dgozman (approved):

iuliancmarcu:
Thanks for this work @yury-s 🙌🏼

PR #27672, opened 2023-10-17
chore: split error to client and server sides

pavelfeldman:


packages/playwright-core/src/server/errors.ts
```diff
@@ -1,12 +1,11 @@
 /**
- * Copyright 2018 Google Inc. All rights reserved.
- * Modifications copyright (c) Microsoft Corporation.
+ * Copyright (c) Microsoft Corporation.
```
dgozman, line 2: Bad merge?

dgozman (approved):

PR #27690, opened 2023-10-18
feat(test runner): introduce "ignored" test status

dgozman:
New "ignored" status differentiates from the "skipped":
- Test is ignored when the test runner never executed it. For example:
  - Testing was interrupted, and last few tests were ignored.
  - Testing stopped early because of maxFailures.
  - Failure in serial mode prevented following tests from running.
- Test is skipped when it is explicitly marked as skipped.

This resolves confusion around various retry cases:
- skipped + passed = passed
- skipped + failed = flaky
- ignored + passed = passed
- ignored + failed = failed

Changes in this PR:
- new status for `TestResult`, but no new status for `TestCase`;
- terminal reporters print ignored vs skipped separately;
- all other reporters visually treat ignored similar to skipped.

### DIFF
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -134,7 +134,7 @@ Errors thrown during test execution, if any.
 
 ## property: TestInfo.expectedStatus
 * since: v1.10
-- type: <[TestStatus]<"passed"|"failed"|"timedOut"|"skipped"|"interrupted">>
+- type: <[TestStatus]<"passed"|"failed"|"timedOut"|"skipped"|"interrupted"|"ignored">>
 
 Expected status for the currently running test. This is usually `'passed'`, except for a few cases:
 * `'skipped'` for skipped tests, e.g. with [`method: Test.skip#2`];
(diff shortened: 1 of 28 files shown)


docs/src/test-api/class-testinfo.md
```diff
@@ -134,7 +134,7 @@ Errors thrown during test execution, if any.
 
 ## property: TestInfo.expectedStatus
 * since: v1.10
-- type: <[TestStatus]<"passed"|"failed"|"timedOut"|"skipped"|"interrupted">>
+- type: <[TestStatus]<"passed"|"failed"|"timedOut"|"skipped"|"interrupted"|"ignored">>
```
pavelfeldman, line 137: How can the expected status be ignored?

docs/src/test-api/class-testinfo.md
```diff
@@ -401,7 +401,7 @@ Suffix used to differentiate snapshots between multiple test configurations. For
 
 ## property: TestInfo.status
 * since: v1.10
-- type: ?<[TestStatus]<"passed"|"failed"|"timedOut"|"skipped"|"interrupted">>
+- type: ?<[TestStatus]<"passed"|"failed"|"timedOut"|"skipped"|"interrupted"|"ignored">>
```
pavelfeldman, line 404: "notStarted" or "initial"?

packages/html-reporter/src/types.ts
```diff
@@ -96,7 +96,7 @@ export type TestResult = {
   steps: TestStep[];
   errors: string[];
   attachments: TestAttachment[];
-  status: 'passed' | 'failed' | 'timedOut' | 'skipped' | 'interrupted';
+  status: 'passed' | 'failed' | 'timedOut' | 'skipped' | 'interrupted' | 'ignored';
```
pavelfeldman, line 99: Why would there be a test result for a test that did not run?

PR #27692, opened 2023-10-18
fix(ct): react render array as child

sand4rt:


dgozman (commented):
The fix looks good, but can we add a test?

sand4rt:
> The fix looks good, but can we add a test?

@dgozman I've added the tests. Solid and Vue/tsx is probably broken as well. Can test that later on

tests/components/ct-react-vite/tests/children.spec.tsx
```diff
@@ -49,8 +49,9 @@ test('render string as child', async ({ mount }) => {
 });
 
 test('render array as child', async ({ mount }) => {
-  const component = await mount(<DefaultChildren>{[4,2]}</DefaultChildren>);
-  await expect(component).toContainText('42');
+  const component = await mount(<DefaultChildren>{[<h4>{[4]}</h4>,<p>2</p>]}</DefaultChildren>);
```
dgozman, line 52: Let's include array of arrays, since these also seem possible in React.

```suggestion
  const component = await mount(<DefaultChildren>{[<h4>{[4]}</h4>,[[<p>[2,3]</p>]]]}</DefaultChildren>);
```

pavelfeldman (approved):

pavelfeldman:
@sand4rt ^^ do you want to look at it?

sand4rt:
> @sand4rt ^^ do you want to look at it?

I've updated the PR. Not sure about the lint error, is that correct?

dgozman:
> I've updated the PR. Not sure about the lint error, is that correct?

That's from another PR, sorry for the noise.

> Could the issue remain open? I still wanted to test/fix the other frameworks.

Of course!

packages/playwright-ct-react/registerSource.mjs
```diff
@@ -76,23 +76,27 @@ async function __pwResolveComponent(component) {
 }
 
 /**
- * @param {JsxComponent | JsxComponentChild} component
+ * @param {JsxComponentChild} child
  */
-function __pwRender(component) {
-  if (!isComponent(component))
```
dgozman, line 82: I think removing this check breaks the following code: `await mount(23)` or `await component.update('some text')`. Not sure whether we actually supported this use case, wdyt?

sand4rt: This case is not supported because `23` doesn't conform to the expected `JSX.Element` type. Only JSX elements can be passed with the `mount` and `update` functions

dgozman: Thank you for clarification!

dgozman (approved):

PR #27700, opened 2023-10-19
feat: add pixel 7 with android 14

Th3S4mur41:
Adding a more recent pixel phone with current Android version.
This also matches the pixel devices available in Chrome/Edge Canary DevTools


Th3S4mur41:

packages/playwright-core/src/server/deviceDescriptorsSource.json
```diff
@@ -1388,6 +1388,36 @@
...
+  },
+    "Pixel 7": {
+    "userAgent": "Mozilla/5.0 (Linux; Android 14; Pixel 7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.6045.21 Mobile Safari/537.36",
+    "screen": {
+      "width": 412,
+      "height": 915
+    },
+    "viewport": {
+      "width": 412,
+      "height": 915
```
mxschmitt, line 1400:
```suggestion
      "height": 839
```

I created a virtual device in the AVD, navigated to playwright.dev, made sure to scroll down, so that the URL bar is not visible anymore and then I executed the following script:

```js
console.log({
    screen: {
        width: window.screen.width,
        height: window.screen.height,
    },
    viewport: {
        height: window.innerHeight,
        width: window.innerWidth,
    }
})
```

in DevTools `chrome://inspect/#devices`. This gave me 839 instead, I think it never becomes more than that, wdyt?

This then looks like that:

image

mxschmitt:
Also make sure to run `npm run lint` so the auto-generated sources will update.

packages/playwright-core/src/server/deviceDescriptorsSource.json
Th3S4mur41: You are right, I took the values from the DevTools
image


This also matches the screen resolution of 2400px divided by the deviceScaleFactor.
But it's not taking the bottom OS navigation into account.

Th3S4mur41:
@mxschmitt 
I also ran `npm run lint` as suggested, but this is still failing in CI
image

What am I missing?

mxschmitt (approved):

packages/playwright-core/src/server/deviceDescriptorsSource.json
```diff
@@ -1389,6 +1389,36 @@
...
+  },
+  "Pixel 7 landscape": {
+    "userAgent": "Mozilla/5.0 (Linux; Android 14; Pixel 7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.6045.33 Mobile Safari/537.36",
+    "screen": {
+      "width": 915,
+      "height": 412
+    },
+    "viewport": {
+      "width": 915,
+      "height": 336
```
mxschmitt, line 1415:
```suggestion
      "width": 863,
      "height": 360
```

PR #27706, opened 2023-10-19
fix(recorder): Locator picker had wrong initial language in language bindings

mxschmitt:
**Description**

When a language port was using Inspector with the "Locator Picker" feature, it only recognised JavaScript as a language by default. As a workaround the user was able to click record, interact with the page and then the language would be correctly used -> csharp e.g. would work in the "Locator Picker".

**Why?**

Our language bindings are setting `PW_LANG_NAME=<sdkLanguage>` env var -> good. Our recorder harness also uses this along its internal state here:

https://github.com/microsoft/playwright/blob/b9b289b6415515b1b8e1a2524ed6425c8992af5a/packages/playwright-core/src/server/recorder.ts#L369

and it gets used here (no parameter means: we use the first language aka. primary language):

https://github.com/microsoft/playwright/blob/b9b289b6415515b1b8e1a2524ed6425c8992af5a/packages/playwright-core/src/server/recorder.ts#L95

The only issue is that the Inspector frontend in the beginning does not know which language it should use and pass over to the server side, it then falls back to JavaScript.

**Proposed fix**

Instead of passing it over from the frontend to the server side, we just always use it from the server side, aka. "currentLanguage". When the user switches languages in the frontend, "currentLanguage" already gets updated properly via the "fileChanged" event.

https://github.com/microsoft/playwright-dotnet/issues/2718

### DIFF
--- a/tests/library/inspector/pause.spec.ts
+++ b/tests/library/inspector/pause.spec.ts
@@ -386,6 +386,31 @@ it.describe('pause', () => {
     await scriptPromise;
   });
 
+  it('should highlight on explore (csharp)', async ({ page, recorderPageGetter }) => {
+    await page.setContent('<button>Submit</button>');
+    process.env.TEST_INSPECTOR_LANGUAGE = 'csharp';
+    try {
+      const scriptPromise = (async () => {
+        await page.pause();
+      })();
+      const recorderPage = await recorderPageGetter();
+
+      const box1Promise = waitForTestLog<Box>(page, 'Highlight box for test: ');
+      await recorderPage.getByText('Locator', { exact: true }).click();
+      await recorderPage.locator('.tabbed-pane .CodeMirror').click();
+      await recorderPage.keyboard.type('GetByText("Submit")');
+      const box1 = await box1Promise;
+
+      const button = await page.$('text=Submit');
+      const box2 = await button.boundingBox();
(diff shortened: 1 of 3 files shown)


dgozman (approved):
Any chance for a test?

tests/library/inspector/pause.spec.ts
```diff
@@ -386,6 +386,31 @@ it.describe('pause', () => {
     await scriptPromise;
   });
 
+  it('should highlight on explore (csharp)', async ({ page, recorderPageGetter }) => {
+    await page.setContent('<button>Submit</button>');
+    process.env.TEST_INSPECTOR_LANGUAGE = 'csharp';
+    try {
```
mxschmitt, line 392:
```suggestion
    process.env.TEST_INSPECTOR_LANGUAGE = 'csharp';
    try {
      await page.setContent('<button>Submit</button>');
```

mxschmitt:
> Nice, that worked :) But still not 100% correctly, if I do Locator("<some_xpath_selector1>").Locator("<some_xpath_selector2>") then it doesn't work.
> 
> I had to put xpath= in front of it, so when having it like this, it works just fine, but shouldn't be required :) Locator("xpath=<some_xpath_selector1>").Locator("xpath=<some_xpath_selector2>")

This gets addressed in https://github.com/microsoft/playwright/pull/27742.

Issue #27707, opened 2023-10-19
[BUG] Inspector Locator Picker doesn't work with double quoted strings which contain single quotes


mxschmitt:
1. `npx playwright open "data:text/html,<div class=foo>bar</div>"`
1. Paste this in the Locator Picker field: `locator("//div[contains(@class, 'foo')]")`

Workaround: Use single quotes outside for now: `locator('//div[contains(@class, "foo")]')` which works. This won't work in C# tho, since a single quote string in C# is a char, no string.

Good: `locator('text="bar"')`
Bad: `locator("text='bar'")`
Good: `locator('//div[contains(@class, "foo")]')`
Bad: `locator("//div[contains(@class, 'foo')]")`

Code Pointer: https://github.com/microsoft/playwright/blob/d9eabda09dd9a08577e2d684c2a78c32a64573a6/packages/playwright-core/src/utils/isomorphic/locatorParser.ts#L213

Note: xpath does not work in general, since we compare the original selector with the generated selector which is `xpath=` prefixed.

Extracted from https://github.com/microsoft/playwright-dotnet/issues/2718.

PR #27742, opened 2023-10-22
fix(inspector): highlight xpath/css locators without engine prefix

### LINKED ISSUE #2718: Better Locator exceptions for better debugging of selectors using Playwright Inspector
Hi there,
I am using Playwright with POM, so I have Pages - Components - Elements structure...
Most important is the Components, because I use them for example to work with itemsLists... But in a website I am testing, there are often components inside another component (often another items component inside one item of an outer items component)...
Then elements inside these components have their own selector, so my locators are often very huge and looks like this:
Locator(...SOME_SELECTOR...).Locator(...ANOTHER_SELECTOR...).Locator(...ANOTHER...SELECTOR) and so on... 
(I don't manually write these selectors like this, they are automatically created for me by my framework, so for example I just write a selector for the whole itemsComponent, then I write another selector for locating items, then there is a class for oneItem of that list, and there are elements in it, for which I again write their own selectors, and I don't have to do anything, it is automatically joined together)

So my locators often could have like 3-5 chaining selectors, but when there is a mistake somewhere in one of those selectors, it is hard for me to recreate this into Playwright Inspector (which is super helpful for me to debug these kind of selectors), because Playwright inspector doesn't understand that .Locator(), that is c# not xpath or css, so whevenever I have an exception, that waiting for locator timeouted, I have to copy this from that exception and replace all .Locator( with playwright way >>

So, it would be awesome, if you make it easier :) So there are 2 possible solutions
1. Make that exception to be written with >> instead of .Locator, so I can just copy it and immediately debug 
2. Or you can make Playwright Inspector to be able to understand the notion .Locator()

I guess the first solution would be better. I already thought, that I will do it myself, I could catch all these exceptions and rethrow another, that would proccess the original exception message and recreate it with Playwright selectors using >>
But then it wouldn't be for everybody, only for me, so I suggest you to make this change, it will be very nice :)

PS: Because I always see this exception inside "Test Detail Summary" inside Test Explorer, I had an idea, that the very best way for this to be super easy would be to have actually some button in the "Test Detail Summary" that I could just click, and the whole selector would get copied into clipboard. 
Because I have sometimes hard time to select the whole selector when it is too long (whenever I try to select it with mouse, and my mouse go little under the row I am trying to select, it all unselects and the horizontal scrollbar of Test Detail Summary goes again all the way to the left). So I usually rather select it using just mouse click at the beginning holding shift and clicking at the end. But still, if there is some chance for some nice button appearing inside Test Detail Summary, that would just copy the whole selector and I can immediately paste it into Playwright Inspector, then it would be super awesome :)

PS2: And then inside Playwright Inspector, I often need to debug it step by step (first I add only the very first selector until the first >> and if it located what I wanted, I then copy there another part until another >> and so on...) and whenever I see that it is not locating what I wanted, I understand where the problem is. 
So there is another idea, what if Playwright Inspector doesn't have just one textbox for the selector, but it would rather understand, that selectors are made up of parts divided by >> and so after pasting the whole selector, it could automatically write each selector part into separate row and then those rows could have something like checkbox next to it, so I could just unselect all parts and then just by pressing these checkboxes one by one, I would turn on those parts.. This would be the super awesome experience and the fastest possible debugging :)

PS3: Playwright Inspector does not only misunderstand the .Locator(). but also something like First() or Last() and so on, so even these things need to be converted into Playwright way, like nth=1 or nth=-1 and so on...

If you consider to implement this debuging experience, I would appreciate it a lot and I would be super thankful :)
If I haven't explained it clearly, please ask any additional questions :) I might provide screenshots or video :)

mxschmitt:
Motivation: As of today when a user inspects a Locator which is a xpath, it won't work if the user has not prefixed it with `xpath=` because we internally compare the given with the generated locator.

Works: `locator('xpath=//div[contains(@class, "foo")]')`
Does not work: `locator('//div[contains(@class, "foo")]')`

Relates https://github.com/microsoft/playwright/issues/27707#issue-1952360264
Fixes https://github.com/microsoft/playwright-dotnet/issues/2718#issuecomment-1771073816

### DIFF
--- a/packages/playwright-core/src/utils/isomorphic/selectorParser.ts
+++ b/packages/playwright-core/src/utils/isomorphic/selectorParser.ts
@@ -123,11 +123,18 @@ function selectorPartsEqual(list1: ParsedSelectorPart[], list2: ParsedSelectorPa
   return stringifySelector({ parts: list1 }) === stringifySelector({ parts: list2 });
 }
 
-export function stringifySelector(selector: string | ParsedSelector): string {
+export function stringifySelector(selector: string | ParsedSelector, forceEngineName?: boolean): string {
   if (typeof selector === 'string')
     return selector;
   return selector.parts.map((p, i) => {
-    const prefix = p.name === 'css' ? '' : p.name + '=';
+    let hideEngine = false;
+    if (!forceEngineName) {
+      if (p.name === 'css')
+        hideEngine = true;
+      else if (p.name === 'xpath' && p.source.startsWith('//') || p.source.startsWith('..'))
+        hideEngine = true;
+    }
+    const prefix = hideEngine ? '' : p.name + '=';
     return `${i === selector.capture ? '*' : ''}${prefix}${p.source}`;
   }).join(' >> ');
 }
--- a/packages/playwright-core/src/utils/isomorphic/selectorParser.ts
+++ b/packages/playwright-core/src/utils/isomorphic/selectorParser.ts
@@ -127,14 +127,14 @@ export function stringifySelector(selector: string | ParsedSelector, forceEngine
   if (typeof selector === 'string')
     return selector;
   return selector.parts.map((p, i) => {
-    let hideEngine = false;
-    if (!forceEngineName) {
+    let includeEngine = true;
+    if (!forceEngineName && i !== selector.capture) {
       if (p.name === 'css')
-        hideEngine = true;
-      else if (p.name === 'xpath' && (p.source.startsWith('//') || p.source.startsWith('..')))
-        hideEngine = true;
+        includeEngine = false;
+      else if (p.name === 'xpath' && p.source.startsWith('//') || p.source.startsWith('..'))
+        includeEngine = false;
     }
-    const prefix = hideEngine ? '' : p.name + '=';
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/utils/isomorphic/selectorParser.ts
```diff
@@ -123,11 +123,18 @@ function selectorPartsEqual(list1: ParsedSelectorPart[], list2: ParsedSelectorPa
...
 }
 
-export function stringifySelector(selector: string | ParsedSelector): string {
+export function stringifySelector(selector: string | ParsedSelector, forceEngineName?: boolean): string {
   if (typeof selector === 'string')
     return selector;
   return selector.parts.map((p, i) => {
-    const prefix = p.name === 'css' ? '' : p.name + '=';
+    let hideEngine = false;
+    if (!forceEngineName) {
```
dgozman, line 131: While reading this, I realized that we cannot hide engine for the "capture" part.

```suggestion
    if (!forceEngineName && i !== selector.capture) {
```

dgozman (approved):

PR #27745, opened 2023-10-22
docs: .NET ParallelScope

### LINKED ISSUE #2732: [BUG/Question] Is [Parallelizable(ParallelScope.Children)] and or sharding supported in the dotnet version?
### System info
- Playwright Version: [v1.38]
- Operating System: [All, Windows 11, Ubuntu 20]
- Other info: I'm trying to figure out if parallel test case execution is supported in the .NET version of Playwright. The documentation mentions, [Running test in parallel using ParallelScope.All or ParallelScope.Fixtures is not supported.](https://playwright.dev/dotnet/docs/test-runners#running-nunit-tests-in-parallel) Does this mean parallel test execution isn't supported at all, or is it just specific tags on methods or classes that aren't supported?

```csharp
    /// <summary>
    /// Specifies the degree to which a test, and its descendants, 
    /// may be run in parallel.
    /// </summary>
    [Flags]
    public enum ParallelScope
    {
        /// <summary>
        /// The test may be run in parallel with others at the same level.
        /// Valid on classes and methods but has no effect on assemblies.
        /// </summary>
        Self = 1,
        /// <summary>
        /// Descendants of the test may be run in parallel with one another.
        /// Valid on assemblies and classes but not on non-parameterized methods.
        /// </summary>
        Children = 256,
        /// <summary>
        /// Descendants of the test down to the level of TestFixtures may be 
        /// run in parallel with one another. Valid on assemblies and classes
        /// but not on methods.
        /// </summary>
        Fixtures = 512,
        /// <summary>
        /// The test and its descendants may be run in parallel with others at
        /// the same level. Valid on classes and parameterized methods.
        /// For assemblies it is recommended to use <see cref="Children"/>
        /// instead, as <see cref="Self"/> has no effect on assemblies.
        /// </summary>
        All = Self + Children
    }
}
```

### Source code

- [X ] I provided exact source code that allows reproducing the issue locally.

**[Link to the GitHub repository with the repro](https://github.com/nullptrerror/DotNetE2EStarterKit)**

```csharp
using NUnit.Framework;
using NUnit_E2E.TestFixture.PlaywrightTest;

namespace NUnit_E2E.TestPlan.Httpbin.TestSuites;

[TestFixture]
[Parallelizable(ParallelScope.Children)]
public sealed class HttpbinTestSuite : BrowserTest
{
    [TestCase("get/200")]
    [TestCase("basic-auth/username/password")]
    [TestCase("headers")]
    [TestCase("ip")]
    [TestCase("user-agent")]
    [TestCase("cache")]
    [TestCase("cache/60")]
    [TestCase("etag/etag")]
    [TestCase("response-headers")]
    [TestCase("brotli")]
    [TestCase("defalte")]
    [TestCase("deny")]
    [TestCase("encoding/utf8")]
    [TestCase("gzip")]
    [TestCase("html")]
    [TestCase("json")]
    [TestCase("robots.txt")]
    [TestCase("xml")]
    public async Task GotoAsync(string endpoint)
    {
        string url = "https://httpbin.org/" + endpoint;
        await Page.GotoAsync(url);
        Assert.That(Page.Url, Does.StartWith(url));
    }
}
```

**Steps**
- [Run the test] 
- [Observe if tests are running in parallel or not]

**Expected**

Tests should run in parallel if parallel test case execution is supported.
Test should be shardable.

**Actual**

Tests run in parallel but I'm not sure if it's because my test cases are simple.
Test seem to be able to be shardable but I'm still working on it on my repo poc.

mxschmitt:
Fixes https://github.com/microsoft/playwright-dotnet/issues/2732

### DIFF
--- a/docs/src/test-runners-csharp.md
+++ b/docs/src/test-runners-csharp.md
@@ -110,7 +110,7 @@ dotnet test --filter "Name~Slogan"
 ### Running NUnit tests in Parallel
 
 By default NUnit will run all test files in parallel, while running tests inside each file sequentially (`ParallelScope.Self`). It will create as many processes as there are cores on the host system. You can adjust this behavior using the NUnit.NumberOfTestWorkers parameter.
-Only `ParallelScope.Self` is supported as of today.
+Only `ParallelScope.Self` is supported.
 
 For CPU-bound tests, we recommend using as many workers as there are cores on your system, divided by 2. For IO-bound tests you can use as many workers as you have cores.
 
--- a/docs/src/test-runners-csharp.md
+++ b/docs/src/test-runners-csharp.md
@@ -110,7 +110,7 @@ dotnet test --filter "Name~Slogan"
 ### Running NUnit tests in Parallel
 
 By default NUnit will run all test files in parallel, while running tests inside each file sequentially (`ParallelScope.Self`). It will create as many processes as there are cores on the host system. You can adjust this behavior using the NUnit.NumberOfTestWorkers parameter.
-Running test in parallel using `ParallelScope.All` or `ParallelScope.Fixtures` is not supported.
+Only `ParallelScope.Self` is supported as of today.
 
 For CPU-bound tests, we recommend using as many workers as there are cores on your system, divided by 2. For IO-bound tests you can use as many workers as you have cores.
 


docs/src/test-runners-csharp.md
```diff
@@ -110,7 +110,7 @@ dotnet test --filter "Name~Slogan"
 ### Running NUnit tests in Parallel
 
 By default NUnit will run all test files in parallel, while running tests inside each file sequentially (`ParallelScope.Self`). It will create as many processes as there are cores on the host system. You can adjust this behavior using the NUnit.NumberOfTestWorkers parameter.
-Running test in parallel using `ParallelScope.All` or `ParallelScope.Fixtures` is not supported.
+Only `ParallelScope.Self` is supported as of today.
```
Meir017, line 113:
```suggestion
Only `ParallelScope.Self` is supported as of v1.38
```

dgozman (approved):

docs/src/test-runners-csharp.md
mxschmitt, line 113:
```suggestion
Only `ParallelScope.Self` is supported.
```

PR #27766, opened 2023-10-24
chore: support await using for close() and dispose()

### LINKED ISSUE #27141: [Feature] Support `Symbol.asyncDispose` method (the `using` keyword) in addition to `close()`
This allows us to write this:

```ts
await using context = await browser.newContext()
// Code
```

which is more concise than:

```ts
try {
  // Code
} finally {
  await context.close();
}
```

pavelfeldman:
This change assumes that the user has Node 18 with Symbol.dispose available.

Fixes https://github.com/microsoft/playwright/issues/27141

### DIFF
--- a/packages/playwright-core/src/client/android.ts
+++ b/packages/playwright-core/src/client/android.ts
@@ -234,6 +234,10 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel> i
     return binary;
   }
 
+  async [Symbol.asyncDispose]() {
+    await this.close();
+  }
+
   async close() {
     try {
       if (this._shouldCloseConnectionOnClose)
@@ -306,6 +310,10 @@ export class AndroidSocket extends ChannelOwner<channels.AndroidSocketChannel> i
 
   async close(): Promise<void> {
     await this._channel.close();
+  }
+
+  async [Symbol.asyncDispose]() {
+    await this.close();
   }
 }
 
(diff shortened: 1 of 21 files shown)


packages/playwright-core/src/client/android.ts
```diff
@@ -234,6 +234,10 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel> i
     return binary;
   }
 
+  async [Symbol.asyncDispose]() {
```
dgozman, line 237: I _think_ we can support Node16+ with babel like this:

```ts
async [Symbol.asyncDispose ?? Symbol.for('Symbol.asyncDispose')]() {
  // ...
}
```

utils/generate_types/test/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "strict": true,
     "target": "ES2019",
+    "lib": ["esnext", "dom", "DOM.Iterable"],
```
dgozman, line 5: This suggests that this change will break compilation for our users that do not have `esnext` target enabled.

pavelfeldman: I don't think they'll compile them, and VS Code resolves it to @types/next:

Screenshot 2023-10-24 at 9 15 59 AM

jonkoops:
@pavelfeldman great feature, but it seems none of this behavior is documented. Are there any plans to document the functionality now that Node.js [has added explicit resource management](https://nodejs.org/en/blog/release/v20.4.0#support-to-the-explicit-resource-management-proposal)?

PR #27783, opened 2023-10-24
feat(ct): angular component testing

### LINKED ISSUE #14153: [Feature] Support for Component Tests in Angular apps
Please support Component Testing for Angular.

In v1.22.0, there is experimental support for react, vue, and svelte Component Tests
> https://github.com/microsoft/playwright/releases/tag/v1.22.0

F997E557-0366-4AF7-BB76-1BCDA37C0611

sand4rt:
closes: https://github.com/microsoft/playwright/issues/14153 and https://github.com/sand4rt/playwright-ct-angular

### TODO
- [x] ~~Enable vite-plugin-angular JIT mode: https://github.com/analogjs/analog/pull/374#issuecomment-1536991065~~
- [x] Fix `Sourcemap is likely to be incorrect: a plugin (@analogjs/vite-plugin-angular)` errors when transpiling: https://github.com/analogjs/analog/issues/410
- [x] Restructure NPM (peer)dependencies?: https://github.com/microsoft/playwright/pull/27783#discussion_r1372343294
- [x] Remove `typeof plugin.default` check: https://github.com/microsoft/playwright/pull/27783/files#diff-4592ac823d44ec894c518ba459cfab4bd544056a674739fda5fc145cdc596923R28
- [x] Update the documentation _(PR ready: https://github.com/sand4rt/playwright/pull/4)_
- [x] Update https://github.com/microsoft/create-playwright
-  [x] Do not handle routing and let users use router testing harness: https://github.com/microsoft/playwright/pull/27783#discussion_r1372305687
- [x] Update tests to use Angular 17 ? https://github.com/microsoft/playwright/pull/27783#issuecomment-1809269085
- [x] Remove `@analogjs/vite-plugin-angular` and related code and move it to `create-playwright` ? https://github.com/microsoft/playwright/pull/27783#discussion_r1372342351
- [x] Update old Angular logo: https://github.com/microsoft/playwright/pull/27783#pullrequestreview-1702918975


pavelfeldman:
What's out plan here, is it ready to land? Is it based on Younes's work or is this something different?

packages/playwright-ct-angular/index.js
```diff
@@ -0,0 +1,38 @@
...
+const path = require('path');
+
+process.env['NODE_ENV'] = 'test';
+
+function plugin() {
+  // Only fetch upon request to avoid resolution in workers.
+  const { createPlugin } = require('@playwright/experimental-ct-core/lib/vitePlugin');
+  return createPlugin(
+      path.join(__dirname, 'registerSource.mjs'),
+      () => import('@analogjs/vite-plugin-angular').then(plugin => {
```
yjaaidi, line 27: IMO, the most important conversation is about `@analogjs/vite-plugin-angular`

## 1. should `@analogjs/vite-plugin-angular` be optional?
Adding a vite plugin could be an overhead in some cases. I'm not here to advocate for Angular Single-File-Component (inline template / styles), but there could be issues (cf. https://github.com/analogjs/analog/issues/410)


## 2. Users should probably be able to customize `@analogjs/vite-plugin-angular` 
In most SFC cases, we don't need transforms and jit turns out to be faster so skipping the plugin is an interesting option.
I also noticed some issues when the plugin doesn't grab the right tsconfig etc... so we need to be able to pass the tsconfig to the plugin.

Shouldn't we allow users to pass options to the plugin?

## 3. `@analogjs/vite-plugin-angular` should be an opt-out
In most cases, we need the plugin so it should be enabled by default while we could still disable it if needed.

## 4. `@analogjs/vite-plugin-angular` should not be a dependency

Now the main problem with the current approach is that the current package is highly coupled to `@analogjs/vite-plugin-angular` which is coupled to Angular's major version and I guess that it could even break on Angular minors because Angular's eslint plugins are not public API.

Suppose `@analogjs/vite-plugin-angular` 0.3.0 is only compatible with Angular >=17, while v0.2.0 is only compatible with <=16. Bumping is going to be a hard decision, especially since playwright CT packages do not follow semver. We might not want to have different plugins for different Angular versions 🙈.

a. Maybe `@analogjs/vite-plugin-angular` should be a peer dependency.
b. Maybe https://github.com/microsoft/create-playwright should add the dependency and set up `playwright.config.ts` with `@analogjs/vite-plugin-angular` which is passed as a parameter to `defineConfig`.
c. Or my favorite option, always use the latest version of `@analogjs/vite-plugin-angular` as a dependency but let users control the plugin and not just the options so that they can provide an older version or a future unstable version if they live on t he bleeding edge or another plugin if they have one.

cc. @brandonroberts

packages/playwright-ct-angular/package.json
```diff
@@ -0,0 +1,63 @@
...
+    }
+  },
+  "dependencies": {
+    "@analogjs/vite-plugin-angular": "0.2.10",
+    "@angular-devkit/build-angular": "^16.1.0",
+    "@playwright/experimental-ct-core": "1.40.0-next",
+    "vite": "^4.4.9"
+  },
+  "devDependencies": {
+    "@angular/animations": "^16.1.7",
```
yjaaidi, line 38: shouldn't all `@angular/*` be peerDeps?

yjaaidi (commented):
Great job @sand4rt!
Here are a couple of things that would be nice to solve before merging.
Let me know if you want me to contribute to your branch directly with some PRs.

The main discussion here is about vite plugin replacement and configuration.

packages/playwright-ct-angular/package.json
yjaaidi: and we can remove:
`@angular/animations`
`@angular/common`
`@angular/compiler`
`@angular/compiler-cli`
`@angular/platform-browser`
as they are not used

yjaaidi:
> What's out plan here, is it ready to land? 
Hey @pavelfeldman! I just shared my feedback with @sand4rt.
Once we agree on that and fix what can be fixed, the PR should be ready.

> Is it based on Younes's work or is this something different?
It doesn't matter anymore. There was just a misunderstanding.
We were just a bit disappointed for not merging efforts but now we are working together to provide what the community is waiting for and that's what matters.

packages/playwright-ct-angular/registerSource.mjs
```diff
@@ -0,0 +1,237 @@
...
+  for (const [name, value] of Object.entries(props))
+    fixture.debugElement.children[0].context[name] = value;
+}
+
+/**
+ * @param {import('@angular/core/testing').ComponentFixture} fixture
+ */
+function __pwUpdateEvents(fixture, events = {}) {
+  for (const [name, value] of Object.entries(events)) {
+    fixture.debugElement.children[0].componentInstance[name] = {
```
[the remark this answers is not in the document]
yjaaidi: Cf. https://github.com/jscutlery/devkit/blob/6fa2f1af06249a154f59f0082108dc4eca10d9b5/packages/playwright-ct-angular/src/lib/register-source.mjs#L48

https://github.com/jscutlery/devkit/blob/6fa2f1af06249a154f59f0082108dc4eca10d9b5/packages/playwright-ct-angular/src/lib/register-source.mjs#L55

packages/playwright-ct-angular/index.js
brandonroberts: Thanks @yjaaidi. The thing is, it won't work without a plugin. You could use JIT mode, but unless you have the ability to use inline templates, inline styles, and inject everywhere including library code, things will break. 

Even if there was first-party support by Angular, it would still need to be done through a plugin.

The plugin is what connects Angular compilation to work within Vite, which is what Playwright uses for component testing. The plugin itself its built to support Angular >=16 and its intended to be kept that way and not be tied to specific versions of Angular.

However you want to add the plugin is fine, but its not optional in this implementation.

pavelfeldman: @yjaaidi @analogjs/vite-plugin-angular would be a tough dependency to add:

- it comes from a new publisher
- it does not have a ton of dependencies or downloads yet
- it does not seem to have a lot of code

Unless we anticipate a ton of maintenance work around it, I would consider implementing a plugin tailored around the Playwright needs instead. That would help with upstream fixes, versioning, etc. This is not a strict requirement while we are experimental, but that would become a GA blocker.

@brandonroberts I don't want it to come across as not appreciation of your hard work. I think your plugin is great, not seeing official support for Angular for Vite is quite unfortunate. It is just that the acceptance criteria for the npm modules Playwright can depend on is quite strict - we need to anticipate all kinds of risks and maintenance costs.

brandonroberts: @pavelfeldman no offense taken. I think your concerns are definitely valid and official support for Vite with Angular through a plugin would be ideal. I'll leave it up to you all on how you want to proceed from here.

sand4rt:
Hey guys, thanks for the comments! I'm AFK for a couple of days. Will hopefully look at the PR by the end of next week as well

packages/playwright-ct-angular/index.js
edbzn: Angular v17 comes with the vite dev-server by default, so it's quite stable. What do you think about directly reusing the [internal Vite dev-server setup function from the CLI](https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts#L349) for playwright? Note that Angular's CLI programmatic APIs are internal and subject to change without following SemVer, but this way, we don't add any new dependencies, and we stick to the framework's internals.

edbzn (commented):
The added logo is the Angular.js one, here's the link to [the correct Angular logo](https://angular.io/presskit).

packages/playwright-ct-angular/package.json
edbzn: What about making peer dependencies range version compatibility more strict by using something like `^15.0.0 || ^16.0.0` instead of `>=15.1.0 || >=16.0.0`? The problem with `>=` is that we allow any greater majors, but in reality, there are breaking changes so we cannot guarantee compatibility.

packages/playwright-ct-angular/index.js
yjaaidi: @edbzn I think that this would require lots of changes to Playwright and also the Angular vite server function as it would have to allow us customizing `entries` and probably a couple of other options.

yjaaidi: > The plugin itself its built to support Angular >=16 and its intended to be kept that way and not be tied to specific versions of Angular.

Oh! That's good news!

yjaaidi: @brandonroberts don't get me wrong. Currently `@analogjs/vite-plugin-angular` is undeniably the best option for integrating Angular with other tools.

## Concerning JIT mode
Concerning JIT mode, the only prerequisites are inline templates / inline styles / inject everywhere _(but AFAIK not in libraries. Libraries are built with partial mode which generates the `ɵfac` factory using `ɵɵinject`)_

Should this be the default setup? I don't think so. We can't make these prerequisites to Playwright Component Testing, but it should still be an option.

## `@analogjs/vite-plugin-angular` as a dep or not?

I think that maintaining an additional plugin is not an option. It's probably not much code but I know that it requires lots of research and maintenance to hook into `@angular-devkit/build-angular` internals. Cf. https://github.com/analogjs/analog/blob/main/packages/vite-plugin-angular/src/lib/utils/devkit.ts

The Angular team might have some future plans to provide a plugin but there is nothing official yet.
Let's hope that this happens before CT GA 😊.

As Brandon confirms that "The plugin itself its built to support Angular >=16 and its intended to be kept that way and not be tied to specific versions of Angular", then the versioning problem I mentioned before disappears. So the plugin could be added as a dependency.

## How to customize `@analogjs/vite-plugin-angular` options

Finally, in order to customize the plugin's options, I think that developers should pass the plugin as an option to `defineConfig` instead of passing the plugin's options.
Exposing the plugin's options in the `defineConfig` would highly couple this package to the plugin.

## TL;DR
My general conclusions:
1. `@analogjs/vite-plugin-angular` can be added as a dep meanwhile Angular exposes an official plugin
2. `defineConfig` should allow us to pass `plugins` (e.g. this way developers can use their custom plugins, or override `@analogjs/vite-plugin-angular`)

What do you think?

packages/playwright-ct-angular/package.json
yjaaidi: Good one @edbzn!

packages/playwright-ct-angular/registerSource.mjs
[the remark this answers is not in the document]
sand4rt: Nice, sounds good! Could you make a PR to this branch? I think you can just:

```bash
git clone https://github.com/sand4rt/playwright
git checkout hello-angular-ct
git checkout -b name-of-the-change
# push the changes and create a PR against https://github.com/sand4rt/playwright/tree/hello-ct-angular
```

@pavelfeldman Is it possible not to squash the commits for this PR while merging? This way, we ensure that the contributions made by @yjaaidi remain visible for the credits?

packages/playwright-ct-angular/registerSource.mjs
```diff
@@ -0,0 +1,237 @@
...
+  TestBed.configureTestingModule({
+    imports: [Component],
+    declarations: [WrapperComponent]
+  });
+
+  await TestBed.compileComponents();
+
+  __pwUpdateSlots(WrapperComponent, component.options?.slots, componentMetadata.selector);
+
+  // TODO: only inject when router is provided
```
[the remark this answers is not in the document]
sand4rt: Would you like to make the change? See related: https://github.com/microsoft/playwright/pull/27783/files#r1382610163

sand4rt:
> The added logo is the Angular.js one, here's the link to [the correct Angular logo](https://angular.io/presskit).

Hi @edbzn, thanks for the input! Would you like to make the change by creating a PR against this branch? See https://github.com/microsoft/playwright/pull/27783#discussion_r1382610163 for more details

packages/playwright-ct-angular/registerSource.mjs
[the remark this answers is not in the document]
yjaaidi: 🙌

packages/playwright-ct-angular/registerSource.mjs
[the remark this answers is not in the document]
sand4rt: I think to `Co-authored-by:` will be added when a PR under your name is merged to this branch. I'll double check that later on. At least it should be clear that this was a combined effort:)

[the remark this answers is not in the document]
yjaaidi: done! https://github.com/sand4rt/playwright/pull/3

yjaaidi:
I just noticed that the tests themselves are using Angular 15.
Should we create a different folder for Angular 16 & Angular 17 and that would require lots of duplication?
Or should we do something at the `test-all.spec.js` level but that would be a bit too magical?
Or should the "experimental" CT only handle the last version of the framework?
cc. @mxschmitt @pavelfeldman @sand4rt

packages/playwright-ct-angular/index.js
yjaaidi: I just noticed that Playwright CT is using Vite's `mergeConfig` to merge `ctViteConfig` with the config embedded in  `@playwright/experiment-ct-angular`.
This means that there is currently no way of choosing a different Vite plugin.

A. Should we move this to "user-land" and let `create-playwright` add `@analogjs/vite-plugin-angular` to `package.json` and set up `playwright.config.ts`'s `ctViteConfig` accordingly?
B. Should `defineConfig()` provide an additional opt-out like `ctDisableBuiltInPlugins`?
C. Should `defineConfig()` detect when `ctViteConfig.plugins` is set but that is not aligned with other plugins behavior and can be hard to maintain?

cc. @mxschmitt @pavelfeldman @sand4rt

tests/components/ct-angular/src/assets/logo.svg
mgechev, line 1: That's such a wonderful initiative! We'd love to get Playwright support for Angular.

Could you also update to the latest logo? https://angular.dev/press-kit

sand4rt: Hey @mgechev, thank you for the comment. If you have more feedback, please let us know!

The logo will be adjusted, see: https://github.com/microsoft/playwright/pull/27783#issuecomment-1793835321

yjaaidi:
Hey @sand4rt, I just created a `create-playwright` PR https://github.com/microsoft/create-playwright/pull/106 to move the Angular vite plugin config to "user-land".

- [ ] Do not handle routing and let users use router testing harness
- [ ] Update tests to use Angular 17
- [ ] Update the documentation
- [X] Update https://github.com/microsoft/create-playwright to generate playwright config with `@analogjs/vite-plugin-angular` and add it to project dependencies.

packages/playwright-ct-angular/registerSource.mjs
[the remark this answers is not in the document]
sand4rt: merged, cc @yjaaidi

packages/playwright-ct-angular/index.js
sand4rt: I believe that @yjaaidi has made the optimal choice _(to move `@analogjs/vite-plugin-angular` to `create-playwright`)_. In this scenario, @brandonroberts's Vite plugin is still used _(as I don't see much value without Vite plugin)_. Hopefully, it will be easier to gain approval from @pavelfeldman since the library is not in the main repository. See also the other pros and cons mentioned here: https://github.com/microsoft/playwright/pull/27783#issuecomment-1821277873.

It would be great if there were an official Vite plugin for Angular. @mgechev, do you have any information on whether this is gonna happen?

yjaaidi:
Hey @sand4rt are you available in the upcoming days or weeks so we can finish this 😊
Maybe if you could just update your branch I could send you a PR.
Thanks!

flobacher:
Great news @yjaaidi and @sand4rt! Thank you so much for your effort time and persistence!!!

sand4rt:
> Maybe if you could just update your branch I could send you a PR.

Hey @yjaaidi, https://github.com/microsoft/playwright/issues/29544 needs to be resolved before i can update. I could resolve the merge conflicts later on if needed, so don't let that hold you back.

For the people willing to contribute; beta testing would also help a lot to ensure everything operates as expected, your feedback is very welcome!

yjaaidi:
Thanks @sand4rt!
@edbzn and I had to update and fix #29544 before moving forward. 
I think that we fixed everything.

maartentibau:
Great great great work! Thank you!!
At this moment this will only support standalone components right?

zargham-leanix:
Great work @sand4rt @yjaaidi 
When this PR is expected to be merged ?

sand4rt:
@maartentibau correct, only standalone components are supported at the moment.

@zargham-leanix the last few open tasks of this PR are being resolved in https://github.com/sand4rt/playwright/pull/5. The progress could be followed over there.

sand4rt:
@yjaaidi i think it's just this tsc error that remains to be done: https://github.com/microsoft/playwright/actions/runs/8992193650/job/24701459373?pr=27783#step:7:32. Do you see other things left todo?

sand4rt:
Hey @pavelfeldman, the PR is finally set for review by the Playwright team! We have deliberately kept the adapter minimal for now and postponed the documentation until more people have tested it and shared their feedback. We would like to gradually make improvements here and there after the merge.

rainerhahnekamp:
Congratulations to all of you. That's a milestone in Angular testing history!

Just a small remark: From what I've seen, there is no possibility to provide services, right?

If that's the case, it would be great to add this as the next feature once this branch is merged.

As you know, the vast majority of Angular Components usually depend on Services. If we could mock these services, the number of people who could test on a large scale would increase "exponentially."

Either way, congratulations and a thousand thanks again. I'm looking forward to it

sand4rt:
> Congratulations to all of you. That's a milestone in Angular testing history!
> 
> Just a small remark: From what I've seen, there is no possibility to provide services, right?
> 
> If that's the case, it would be great to add this as the next feature once this branch is merged.
> 
> As you know, the vast majority of Angular Components usually depend on Services. If we could mock these services, the number of people who could test on a large scale would increase "exponentially."
> 
> Either way, congratulations and a thousand thanks again. I'm looking forward to it

Providers/imports are still open for discussion indeed. We've decided to merge the library first before continuing the discussion: https://github.com/sand4rt/playwright/pull/5#discussion_r1541178625 

I think it could either be supported like the other frameworks through the `hooksConfig`:

```ts
// playwright/index.ts
beforeMount(async ({ TestBed, hooksConfig }) => {
  TestBed.configureTestingModule({
    imports: hooksConfig?.imports,
    providers: hooksConfig?.providers
  });
});
```

```ts
test('hooks config', async ({ mount }) => {
  const component = await mount(AngularComponent, {
    hooksConfig: {
      imports: [Component],
      providers: [Service],
    }
  });
  await expect(component).toContainText('boop');
});
```
or with dedicated APIs:

```ts
test('imports and providers', async ({ mount }) => {
  const component = await mount(AngularComponent, {
    imports: [Component],
    providers: [Service],
  });
  await expect(component).toContainText('boop');
});
```

rainerhahnekamp:
Understood, then let's wait for the merge and then continue the discussion. Congrats again.

chronospatian:
@yjaaidi I am running into some issues while testing this library. 
https://github.com/sand4rt/playwright-ct-angular/issues/51
https://github.com/sand4rt/playwright-ct-angular/issues/52

yjaaidi:
> @yjaaidi I am running into some issues while testing this library. [sand4rt/playwright-ct-angular#50](https://github.com/sand4rt/playwright-ct-angular/issues/50) ~[sand4rt/playwright-ct-angular#51](https://github.com/sand4rt/playwright-ct-angular/issues/51)~ Resolved ~[sand4rt/playwright-ct-angular#52](https://github.com/sand4rt/playwright-ct-angular/issues/52)~ Resolved
> 
> I have pulled and linked the PR branch and still face these errors. I also cannot import from `@angular/material` without adding `"type": "module"` to my package.json, or changing all my files to `.mts`.
> 
> Other than that it's working great so far. Thanks so much for this.

Hi @chronospatian, what is the value of `ctViteConfig` in the playwright config file?
Did you try this:

```ts
    ctViteConfig: {
      plugins: [analog()], // or [swc.vite(swcAngularUnpluginOptions())]
      resolve: {
        /* @angular/material is using "style" as a Custom Conditional export to expose prebuilt styles etc... */
        conditions: ['style'],
      },
    },
```

chronospatian:
@yjaaidi I am using the same config as you have it here: https://github.com/sand4rt/playwright/blob/hello-angular-ct/tests/components/ct-angular/playwright.config.mts

Except I also added `globalThis.window = globalThis` to work around window being undefined when using the `beforeMount` hook.

chronospatian:
Actually it doesn't appear that the `beforeMount` hook is being called at all... Not sure what I'm doing wrong.

sand4rt:
@chronospatain you might want to check out https://github.com/microsoft/playwright/issues/23662 for the esm/cjs issues

yjaaidi:
> I was putting beforeMount in the wrong file, apologies.

In which file did you put it initially?
I'm curious about common mistakes/problems. 😉

chronospatian:
@yjaaidi I initially put it in the spec file, like beforeEach.

chronospatian:
Are there plans to add support for testing for Directives and Pipes?

packages/playwright-ct-angular/registerSource.mjs
```diff
@@ -0,0 +1,137 @@
...
+
+  for (const hook of window.__pw_hooks_before_mount || [])
+    await hook({ hooksConfig, TestBed });
+
+  const fixture = await __pwRenderComponent(component);
+
+  for (const hook of window.__pw_hooks_after_mount || [])
+    await hook({ hooksConfig });
+
+  __pwFixtureRegistry.set(rootElement.id, fixture);
```
dgozman, line 108: Why mapping from id, and not from the `rootElement` itself?

packages/playwright-ct-angular/registerSource.mjs
```diff
@@ -0,0 +1,137 @@
...
+import 'zone.js';
+import { reflectComponentType } from '@angular/core';
+import { getTestBed, TestBed } from '@angular/core/testing';
+import {
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+/** @type {WeakMap<import('@angular/core/testing').ComponentFixture, Record<string, import('rxjs').Subscription>>} */
+const __pwOutputSubscriptionRegistry = new WeakMap();
```
dgozman, line 31: Instead of a `WeakMap`, we usually use a symbol. However, in this case it seems like we can combine this map with `__pwFixtureRegistry`?

```
// rootElement -> fixture, subscriptions
Map<Element, {
  fixture: import('@angular/core/testing').ComponentFixture,
  subscriptions: Record<string,  import('rxjs').Subscription>
}>
```

packages/playwright-ct-angular/registerSource.mjs
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * @param {ObjectComponent} component
+ */
+async function __pwRenderComponent(component) {
+  const componentMetadata = reflectComponentType(component.type);
+  if (!componentMetadata?.isStandalone)
+    throw new Error('Only standalone components are supported');
+
+  TestBed.configureTestingModule({
+    imports: [component.type],
```
dgozman, line 50: If we aim to support components as parameters (for slots?), do we have to list all component types in this `imports` list?

chronospatian:
Locators don't seem to work properly.

```ts
import { Component } from '@angular/core';

@Component({
  standalone: true,
  template: `<div data-testid="test">test</div>`,
})
export class ComponentComponent {}
```

```ts
test('render a component without options', async ({ mount }) => {
  const component = await mount(ComponentComponent);
  await expect(component.getByTestId('test')).toContainText('test');
});

Fails with

```
 Error: Timed out 5000ms waiting for expect(locator).toContainText(expected)

    Locator: locator('#root').locator('internal:control=component').getByTestId('test')
    Expected string: "test"
    Received: <element(s) not found>
    Call log:
      - expect.toContainText with timeout 5000ms
      - waiting for locator('#root').locator('internal:control=component').getByTestId('test')
```

But this passes

```ts
test('render a component without options', async ({ mount }) => {
  const component = await mount(ComponentComponent);
  await expect(component.page().getByTestId('test')).toContainText('test');
});
```

chronospatian:
The test fixture appears to be anchored to the wrong element. This causes some getters, matchers and locators to fail or behave unexpectedly.

```ts
import { Component, ElementRef} from '@angular/core';

@Component({
  standalone: true,
  template: `<div>Test</div>`,
})
export class InjectComponent {
  constructor(element: ElementRef) {
    element.nativeElement.setAttribute('name', 'hello')
  }
}

```

```ts
import { InjectComponent } from '@/components/inject.component';
import { expect, test } from '@playwright/experimental-ct-angular';
import type { HooksConfig } from 'playwright';

test('inject a token', async ({ mount }) => {
  const component = await mount<HooksConfig>(InjectComponent, {
    hooksConfig: { injectToken: true },
  });
  // works
  await expect(component.page().locator('#root').getAttribute('name')).resolves.toBe('hello');
  // doesn't work
  await expect(component.getAttribute('name')).resolves.toBe('hello');

  // works
  await expect(component.page().locator('#root')).toHaveAttribute('name', 'hello');
  // doesn't work
  await expect(component).toHaveAttribute('name', 'hello');
});
```

packages/playwright-ct-angular/registerSource.mjs
```diff
@@ -0,0 +1,137 @@
...
+  const fixture = TestBed.createComponent(component.type);
+  fixture.nativeElement.id = 'root';
+
+  __pwUpdateProps(fixture, component.props);
+  __pwUpdateEvents(fixture, component.on);
+
+  fixture.autoDetectChanges();
+
+  return fixture;
+}
```
chronospatian, line 64: `TestBed.createComponent` discards the root element, which causes the locator to break. Suggested refactor below should resolve this issue:

```ts
/**
 * @param {ObjectComponent & MountOptions} component
 * @param {HTMLElement} rootElement
 */
async function __pwRenderComponent(component, rootElement) {
  await TestBed.compileComponents();

  const fixture = TestBed.createComponent(component.type)

  rootElement.replaceChildren(fixture.nativeElement)
  document.body.replaceChildren(rootElement)

  __pwUpdateProps(fixture, component.props);
  __pwUpdateEvents(fixture, component.on);

  fixture.autoDetectChanges();

  return fixture;
}
```

tests/components/ct-angular/playwright.config.mts
```diff
@@ -0,0 +1,53 @@
...
+export default defineConfig({
+  testDir: 'tests',
+  forbidOnly: !!process.env['CI'],
+  retries: process.env['CI'] ? 2 : 0,
+  reporter: process.env['CI'] ? 'html' : 'line',
+  use: {
+    trace: 'on-first-retry',
+    ctViteConfig: {
+      plugins: [angular({
+        tsconfig: resolve('./tsconfig.spec.json'),
```
chronospatian, line 30: For TestBed APIs like `TestBed.overrideComponent` to work, `jit` must be set to `true`.

tests/components/ct-angular/tests/events.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  component.update({
+    on: {
+      submit() {
+        called = true;
+      },
+    },
+  });
+
+  await component.click();
```
chronospatian, line 41: This line does not click the button.

Suggestion:
```ts
component.locator('css=button').click()
```

tests/components/ct-angular/tests/events.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  const messages: string[] = [];
+  const component = await mount(ButtonComponent, {
+    props: {
+      title: 'Submit',
+    },
+    on: {
+      submit: (data: string) => messages.push(data),
+    },
+  });
+  await component.click();
```
chronospatian, line 15: This line does not click the button.

Suggestion:
```ts
component.locator('css=button').click()
```

tests/components/ct-angular/tests/update.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+test('update event listeners without remounting', async ({ mount }) => {
+  const component = await mount(CounterComponent);
+
+  const messages: string[] = [];
+  await component.update({
+    on: {
+      submit: (data: string) => messages.push(data),
+    },
+  });
+  await component.click();
```
chronospatian, line 28: This line only incidentally clicks the underlying div because the position overlaps.

Suggestion:
```ts
component.locator('css=div').first().click()
```

chronospatian (commented):
With the locator fix these tests should be updated.

chronospatian:
@yjaaidi I've created a PR with fixes for the above and support for template strings https://github.com/sand4rt/playwright/pull/6

packages/playwright-ct-core/src/tsxTransform.ts
```diff
@@ -141,6 +143,24 @@ function collectJsxComponentUsages(node: T.Node): Set<string> {
   return names;
 }
 
+function collectClassMountUsages(node: T.Node): Set<string> {
+  const names = new Set<string>();
+  traverse(node, {
+    enter: p => {
+      // Treat calls to mount and all identifiers in arguments as component usages.
+      // e.g. mount(MyComponent, { imports: [OtherComponent], providers: [Token]})
```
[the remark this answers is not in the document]
sand4rt: Solely relying on the file extension might not be a good idea because of:

- Support for imports/providers in Angular is a must have _[(see this for more info)](https://github.com/microsoft/playwright/pull/27783#issuecomment-2105675103)_
- Native web components don't have a common file extension: _[(see this for more info)](https://github.com/sand4rt/playwright-ct-web/blob/158588c03f9520d75bc4d22bdc3a062626b0f9b6/ct-web-lit/tests/render.spec.ts#L1-L13)_
- Vue components are not always defined in a `.vue` file _[(see this for more info)](https://github.com/microsoft/playwright/pull/30269)_

tests/components/ct-angular/.gitignore
```diff
@@ -0,0 +1,45 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
+
```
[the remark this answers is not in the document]
sand4rt: Yeah, `tests/components/ct-angular` was generated by angular cli

tests/components/ct-angular/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "scripts": {
+    "ng": "ng",
+    "start": "ng serve",
+    "build": "ng build",
+    "watch": "ng build --watch --configuration development",
+    "test": "npx playwright test",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@analogjs/vite-plugin-angular": "1.3.0",
```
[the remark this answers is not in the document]
sand4rt: Angular does not have an official vite plugin unfortunately. This is created by @brandonroberts, he is well known within the Angular community and the creator of Analog _(a meta framework for Angular)_.

The idea was to not include the plugin directly in the adapter, but instead provide it by `create-playwright`: https://github.com/microsoft/create-playwright/pull/106/files#diff-45afeb7d29809ef5db32b613f6a441b2d4fac09f788ec0760174356d2bb5d857R34 until/if Angular has an official Vite plugin

tests/components/ct-angular/src/assets/logo.svg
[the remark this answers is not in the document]
sand4rt: The other generated ct test projects also contain the logo/favicon. This can be useful for distinguishing projects when working on multiple ct adapters. I would like to keep those if you don't mind?

tests/components/ct-angular/src/components/button.component.html
```diff
@@ -0,0 +1 @@
+<button (click)="submit.emit('hello')">{{title}}</button>
```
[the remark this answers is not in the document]
sand4rt: fixed

packages/playwright-ct-angular/index.d.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import type { Locator } from 'playwright/test';
+import type { JsonObject } from '@playwright/experimental-ct-core/types/component';
+import type { TestType } from '@playwright/experimental-ct-core';
+import type { Type } from '@angular/core';
+
+export type ComponentEvents = Record<string, Function>;
+
+export interface MountOptions<HooksConfig extends JsonObject, Component> {
+  props?: Partial<Component> | Record<string, unknown>, // TODO: filter props and handle signals
```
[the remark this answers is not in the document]
sand4rt: The types are quite loose now while they should be stricter. @yjaaidi you already made something for this right?

packages/playwright-ct-angular/registerSource.mjs
```diff
@@ -0,0 +1,137 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @ts-check
+// This file is injected into the registry as text, no dependencies are allowed.
+
+/** @typedef {import('../playwright-ct-core/types/component').ObjectComponent} ObjectComponent */
+
+import 'zone.js';
```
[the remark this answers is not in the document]
sand4rt: This is Angular's mechanism for change detection, the lib won't work without this.

packages/playwright-ct-core/src/tsxTransform.ts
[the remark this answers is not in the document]
dgozman: Hmm... Could you elaborate on the imports/providers bit? I followed the link, and it seems to import `AngularComponent`, presumably from the `AngularComponent.component` file?

I also think that vue and lit are a different story, so I'd rather not mix them up in this PR aiming to support angular. We can have a separate discussion with pros and cons about it in #30269.

tests/components/ct-angular/tests/events.spec.ts
sand4rt: The current test is correct _(and passing)_.

tests/components/ct-angular/playwright.config.mts
sand4rt: Would rather keep it off for now. This is only used for the internal code/tests

tests/components/ct-angular/tests/events.spec.ts
sand4rt: The current test is correct _(and passing)_

packages/playwright-ct-angular/registerSource.mjs
```diff
@@ -0,0 +1,137 @@
...
+
+    /* Store new subscription. */
+    outputSubscriptionRecord[name] = subscription;
+  }
+
+  /* Update output subscription registry. */
+  __pwOutputSubscriptionRegistry.set(fixture, outputSubscriptionRecord);
+}
+
+window.playwrightMount = async (component, rootElement, hooksConfig) => {
```
timdeschryver, line 96: I'm coming here from https://github.com/sand4rt/playwright/pull/5#discussion_r1602736402, which discussed the need to render a HTML template.

After giving it some more thought, I'd like to change my opinion.
With standalone components I don't see a real need support HTML templates.
Mainly because this will only(?) be used for Angular Component Tests, other TL libraries also don't support this AFAIK.

Before standalone components testing a directive contained more setup, but with the current state it's as simple as with other libraries. Yes, we'd have to create a container component manually, but this will also give more control to the user and less dependencies between Playwright and Angular. In my opinion this outweighs the benefits a user gets to create an inline template.

Without support:

```ts
test('renders a directive', async () => {
  // can also be declared outside of the scope to be reusable
  @Component({
    template: `<div appSpoiler data-testid="dir"></div>`,
    imports: [SpoilerDirective],
    standalone: true,
  })
  class WrapperComponent {}

  await mount(WrapperComponent );
});
``` 

With support:

```ts
test('renders a directive', async () => {
  await mount( `<div appSpoiler data-testid="dir"></div>`, {
     /* I suppose some configuration needs to come here to import the Directive */
  });
});
```

chronospatian: This does align with Playwright's suggestion for writing [test stories](https://playwright.dev/docs/test-components#test-stories). However the first example needs to be extracted to a separate file since it's not importable. It would be nice to have [something like storybook](https://gist.github.com/chronospatian/88e8433382e6e23d1fbb40da08afe055) to reduce the boilerplate. Maybe a future version could consume storybook CSF directly?

sand4rt: Thanks for the comments! @timdeschryver Yeah, that's what i'm leaning towards as well. I'll tag you/this when the discussion pops-up again after the merge.

@chronospatian The Storybook team is working on this:) https://github.com/microsoft/playwright/pull/26592#issuecomment-1890927499

sand4rt: Some more info on this: https://storybook.js.org/blog/portable-stories-for-playwright-ct 

related to: https://github.com/microsoft/playwright/issues/30981

chronospatian:
I get errors when I try to import other values from a component file.

```ts
// @/components.counter.component
import { Component, EventEmitter, Input, Output } from '@angular/core';

// adding this line
export const count = 10

@Component({
   template: `{{count}}`
})
export class CounterComponent {
   count = count
}


```

```ts
import { test, expect } from '@playwright/experimental-ct-angular';
import { count, CounterComponent } from '@/components/counter.component';

test('should mount', async ({ mount }) => {
  const component = await mount(CounterComponent);

  await expect(component).toHaveText(`${count}`)
});
```

When I try to run this test I get the following error

```ts
SyntaxError: Cannot use import statement outside a module

   at ../src/components/counter.component.ts:1

> 1 | import { Component, EventEmitter, Input, Output } from '@angular/core';
    | ^
  2 |
  3 | @Component({
  4 |   standalone: true,

    at Object.<anonymous> (/Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:1:1)
    at Object.<anonymous> (/Git/playwright/tests/components/ct-angular/tests/update.spec.ts:2:1)

Error: No tests found.
Make sure that arguments are regular expressions matching test files.
You may need to escape symbols like "$" or "*" and quote the arguments.
```

If I set "type: module" in package.json, then I get a different error.

```
Error: Standard Angular field decorators are not supported in JIT mode.

    at PropDecorator (/Git/playwright/tests/packages/core/src/util/decorators.ts:161:17)
    at applyDec (/Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:3:1813)
    at/Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:3:3514
    at _applyDecs (/Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:3:3685)
    at /Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:23:2
```

Lastly, if I change the test slightly:

```ts
export const count = { value: 10 }
```

 I get yet another error:

```ts
SyntaxError: Identifier 'CounterComponent' has already been declared
```

This is very confusing! I guess this is a limitation of the playwright component test, so the solution here would be to extract the variable to another file that has no Angular imports or component usages, or to not import those values into the test to begin with and use hard coded values in the test.

packages/playwright-ct-angular/package.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "devDependencies": {
+    "@angular/compiler": "^17.0.0",
+    "@angular/core": "^17.0.0",
+    "@angular/platform-browser-dynamic": "^17.0.0",
+    "rxjs": "~7.8.1",
+    "typescript": "~5.2.0",
+    "zone.js": "~0.14.0"
+  },
+  "peerDependencies": {
```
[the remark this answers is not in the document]
yjaaidi: Hi Dmitry, they are used in `registerSource.mjs` in order to create the testbed.

We have to depend on the same Angular major version as the user's. If we hard-depend on Angular 18, we can cause problems to users who didn't migrate to Angular 18 yet. Even worse, the tests could pass (as they are using Angular 18) and break when the app is built (using Angular 17 for example).

packages/playwright-ct-angular/index.d.ts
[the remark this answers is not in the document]
yjaaidi: Yes, exactly!
We implemented this here https://github.com/jscutlery/devkit/blob/353aeb322fecad64ba8b3667613deca537adb743/packages/playwright-ct-angular/src/index.ts#L35

IMHO, it's better if we add this later in a distinct PR to reduce the surface of this PR.

packages/playwright-ct-angular/registerSource.mjs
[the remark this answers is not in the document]
yjaaidi: This is a very good question Dmitry.
Short answer: we don't need anymore and we should let users add it in `playwright/index.ts` if they need it.

The main way of handling change detection in Angular is based on this library. Zone.js monkey patches the necessary browser APIs to know when to trigger change detection.
There has been way to make apps without zone.js but that was mainly community driven. Angular 18 added official experimental support for zoneless (no zone.js) apps.
This means that it's up to the users to add `zone.js` or not.

We can remove this import.
@sand4rt would you like to handle this or would you like to send a PR over your PR?

packages/playwright-ct-core/src/tsxTransform.ts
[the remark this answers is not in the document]
yjaaidi: I don't think that we can assume any suffix or file naming convention for these additional reasons:
- some users use different suffixes/extensions to express the responsibility of the component (e.g. `Page` & `.page.ts`)
- some users might use Analog SFCs (https://analogjs.org/docs/experimental/sfc#analog-sfcs)
- the `Component` suffix might not last (Angular team had already presented the possibility of using components in template without import `template: '<MyButton/>'` (not jsx though). If this happens, suffixes will probably be dropped.
- the `providers` option can receive classes that are not suffixed either and that should be forwarded to the browser (e.g. `mount(MyCmp, {providers: [MyService]})`) 

While the following approach works relatively well, it is limited. Nowadays, most providers are provided using functions that return the provider configuration [`provideAnimations()`](https://angular.dev/api/platform-browser/animations/provideAnimations).
There are three options:
1. keep the transformer as it is but users who want to use provider function will have to move the logic to hooks or story/test container components or create provider constants in other files: Cf. https://github.com/jscutlery/devkit/blob/353aeb322fecad64ba8b3667613deca537adb743/tests/playwright-ct-angular-demo/src/recipe-search.component.pw.ts#L34
2. forward function calls used in `providers`
3. introduce a "compiler function" like `forward(MyService)` that allows users to explicitly choose what should be transformed and "forwarded" to the browser.

packages/playwright-ct-angular/registerSource.mjs
yjaaidi: Nice catch! AFAIK, we could use `rootElement`. @sand4rt?

packages/playwright-ct-angular/registerSource.mjs
yjaaidi: Right now, the best option is to let the user create a parent story/test-container component:

```ts
// test-container.ts
@Component({
  imports: [A, B],
  template: `<a/><b/>`
})
class TestContainer {
}

// test
mount(TestContainer);
```

Otherwise, a more convenient option is to allow the option of providing a template to the `mount` function.
```ts
mount(`<a/><b/>`, {imports: [A, B]});
```

Cf. https://github.com/jscutlery/devkit/blob/353aeb322fecad64ba8b3667613deca537adb743/tests/playwright-ct-angular-wide/tests/template.pw.ts#L9

packages/playwright-ct-angular/registerSource.mjs
yjaaidi: There are different tradeoffs to consider.
For most cases, a test container / story component / portable storybook story is enough.

That said, there are cases where each test (or subgroup of tests) might need a totally different template construction. This can cause some clutter with a different component for each test.
Mounting template in the tests can make some tests a bit more readable with less clutter.
On the other hand, mounting templates will necessarily require JIT.
 
IMHO, this should be up to the testing strategy of the users. 
Also, this could be handled by a 3rd party library like `@jscutlery/playwright-ct-angular` which we are planning to maintain as an incubator for the official package ideas and also as a wrapper of the official support.

yjaaidi:
> I get errors when I try to import other values from a component file.
> 
> ```ts
> // @/components.counter.component
> import { Component, EventEmitter, Input, Output } from '@angular/core';
> 
> // adding this line
> export const count = 10
> 
> @Component({
>    template: `{{count}}`
> })
> export class CounterComponent {
>    count = count
> }
> ```
> 
> ```ts
> import { test, expect } from '@playwright/experimental-ct-angular';
> import { count, CounterComponent } from '@/components/counter.component';
> 
> test('should mount', async ({ mount }) => {
>   const component = await mount(CounterComponent);
> 
>   await expect(component).toHaveText(`${count}`)
> });
> ```
> 
> When I try to run this test I get the following error
> 
> ```ts
> SyntaxError: Cannot use import statement outside a module
> 
>    at ../src/components/counter.component.ts:1
> 
> > 1 | import { Component, EventEmitter, Input, Output } from '@angular/core';
>     | ^
>   2 |
>   3 | @Component({
>   4 |   standalone: true,
> 
>     at Object.<anonymous> (/Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:1:1)
>     at Object.<anonymous> (/Git/playwright/tests/components/ct-angular/tests/update.spec.ts:2:1)
> 
> Error: No tests found.
> Make sure that arguments are regular expressions matching test files.
> You may need to escape symbols like "$" or "*" and quote the arguments.
> ```
> 
> If I set "type: module" in package.json, then I get a different error.
> 
> ```
> Error: Standard Angular field decorators are not supported in JIT mode.
> 
>     at PropDecorator (/Git/playwright/tests/packages/core/src/util/decorators.ts:161:17)
>     at applyDec (/Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:3:1813)
>     at/Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:3:3514
>     at _applyDecs (/Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:3:3685)
>     at /Git/playwright/tests/components/ct-angular/src/components/counter.component.ts:23:2
> ```
> 
> Lastly, if I change the test slightly:
> 
> ```ts
> export const count = { value: 10 }
> ```
> 
> I get yet another error:
> 
> ```ts
> SyntaxError: Identifier 'CounterComponent' has already been declared
> ```
> 
> This is very confusing! I guess this is a limitation of the playwright component test, so the solution here would be to extract the variable to another file that has no Angular imports or component usages, or to not import those values into the test to begin with and use hard coded values in the test.

Hi @chronospatian, could you please provide a repro?
Indeed, `type: module` (or renaming the test to `.mts`) is the way to go.
Did you try enabling `experimentalDecorators` in the playwright's tsconfig?

packages/playwright-ct-angular/registerSource.mjs
yjaaidi: Nice catch @chronospatian!
This is related to the behavior of Playwright's `#root >> internal:control=component` which will use `#root` if it has multiple children or the child if there is a single child.

We should indeed insert the element inside the root element provided by playwright.

The best way of handling this should be by implementing the `TestComponentRenderer` service:

```ts

class PlaywrightTestComponentRenderer extends TestComponentRenderer {

  constructor(rootElement) {
    super();
    this._rootElement = rootElement;
  }

  insertRootElement(testRootElementId) {
    const testRootElement = document.createElement('div');
    testRootElement.id = testRootElementId;
    this._rootElement.appendChild(testRootElement);
    return testRootElement;
  }

  removeRootElement(testRootElement) {
    this._rootElement.removeChild(testRootElement);
  }

}
...
TestBed.configureTestingModule({
  imports: [component.type],
  providers: [
    {
      provide: TestComponentRenderer,
      useValue: new PlaywrightTestComponentRenderer(rootElement)
    }
  ]
});

await TestBed.compileComponents();

const fixture = TestBed.createComponent(component.type);
...
```

Instead of setting the `root` id on the fixture.

yjaaidi: Here is the PR: https://github.com/sand4rt/playwright/pull/7
cc. @sand4rt

packages/playwright-ct-angular/registerSource.mjs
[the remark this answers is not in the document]
yjaaidi: https://github.com/sand4rt/playwright/pull/8

chronospatian:
@yjaaidi Here's the reproduction: https://github.com/chronospatian/playwright/blob/debug-component-usages/tests/components/ct-angular/tests/update.spec.ts

I found a partial workaround. The test passes if I keep component usages and non-component usages in separate imports. I can even hold a reference to the component class itself if I import it a second time under an import alias. Neat!

It would be nice if the typescript transform could discriminate between component and non-component usages in the same import statement.

The test will always fail if the spec file tries to import any values from a file that contains references an Angular component with field decorators (such as `@Input()`. Haven't found a workaround for this yet.

packages/playwright-ct-angular/registerSource.mjs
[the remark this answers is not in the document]
sand4rt: fixed in https://github.com/sand4rt/playwright/pull/8

packages/playwright-ct-core/src/tsxTransform.ts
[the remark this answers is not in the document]
sand4rt: > Hmm... Could you elaborate on the imports/providers bit? I followed the link, and it seems to import AngularComponent, presumably from the AngularComponent.component file?

I meant to say that imports/providers _(especially providers)_ will receive classes that don't necessarily have a file name suffix. 

What concerns do you have regarding the current change?

packages/playwright-ct-angular/registerSource.mjs
sand4rt: This implementation would be more in line with the other frameworks:
https://github.com/sand4rt/playwright-ct-angular/blob/e8a74aa88b88fec329b01661cb0e0235f44d0f5a/ct-angular/tests/slots.spec.ts#L1-L47, but we thought it would be better to merge the library first before moving on to discussing the options

sand4rt:
@dgozman we are currently waiting for a response to resolve the following comments:

1. https://github.com/microsoft/playwright/pull/27783#discussion_r1600457607`
1. https://github.com/microsoft/playwright/pull/27783#discussion_r1600463382
1. https://github.com/microsoft/playwright/pull/27783#discussion_r1600464786
1. https://github.com/microsoft/playwright/pull/27783#discussion_r1600467728
1. https://github.com/microsoft/playwright/pull/27783#discussion_r1600620524
1. https://github.com/microsoft/playwright/pull/27783#discussion_r1600623382
1. https://github.com/microsoft/playwright/pull/27783#discussion_r1600626035

We also have to take care of the following:

1. Repair the tests that broke in https://github.com/microsoft/playwright/pull/27783/commits/cc7e9b80468db70e26c9c9facc448cf608846d87
1. https://github.com/microsoft/playwright/pull/27783#discussion_r1600997621

zargham-leanix:
@sand4rt  @yjaaidi  @dgozman Could I contribute somehow to move this forward ?

flobacher:
@sand4rt @dgozman do you maybe have an estimate on when this could land, or what is still missing? A short update would be greatly appreciated. Thank you very much in advance 🙏

pavelfeldman:
We are trying to make a responsible decision we would be able to stand by in the long run. Looking at the current component testing trends (https://npmtrends.com/@playwright/experimental-ct-react-vs-@playwright/experimental-ct-svelte-vs-@playwright/experimental-ct-vue), we already have reservations around the vue and svelte support viability. Hybrid frameworks are somewhat compromising the whole idea of component testing we well, which also does not make committing any easier. So please let us take time to figure out where we want to bring the component testing going forward. We would hate to add this support to Playwright only to remove it later.

pavelfeldman:
I'll close this PR as the decision is not in the realm of the code, so that we could go back to it when we are ready.

maartentibau:
With all the effort gone in by @sand4rt @yjaaidi and so many others, I really hope this functionality gets added to Playwright asap.

I do understand the concerns and it is very important that those are being addressed, but while this feature has been available in Cypress for a long time now, Playwright really can not stay behind if you ask me. This it an essential part of a testing strategy which cannot be ignored.

pfteter:
Hi @pavelfeldman, I don't think the stats are fine.

If you combine the stats for cypress and PW for react
https://www.npmjs.com/package/@cypress/react
@playwright/experimental-ct-react

you are at 200K weekly downloads. To that you can also add JEST/Testing Library mount component tests to have a good feeling how big the market is.

All Enterprise companies I know are now migrating from Cy to PW because of the recent Cy decisions.
Component Testing as part of Unit testing is a new Word for something that has been around for a while and still not adobted in the wild.

What is the alternative JEST/ Testing Library ? you cant write E2E tests for components you need the mount.

- Remove the experimental tag, add some documentation
- Do some marketing

Then make a decision.

I think PW / CT is by far the best technology and I used JEST/Tesing Lirbary / Cy Component Tests. I wrote my own PW CT for Web Components and it works amazingly well - the best thing there isn't a huge tradeoff compared with JEST but you get all the cool tools that make it a amazing developer experience.

Tallyb:
My 2c: (also answering @pfteter)
Testing components in a browser is a must! 
We went with Storybook for rendering the components, and running PW tests on top of it. For us, it is the best of both worlds: SB for rendering components in isolation (8 years of experience in this specific domain) and PW for testing and browser automation.

yjaaidi:
We, JSCutlery, will keep maintaining [`@jscutlery/playwright-ct-angular`](https://github.com/jscutlery/devkit/tree/main/packages/playwright-ct-angular). 

In case `@playwright/experimental-ct-core` diverges too much and we can't rely on the shared building blocks, we are ready to keep a similar API while maintaining our own.
A prerequisite to keep the same developer experience is to still be able to control babel plugins. @pavelfeldman is it too soon to tell if this will still be exposed?

mxschmitt:
@yjaaidi I invited you to the org, happy to have it there until Playwright's CT is ready.

sand4rt:
> @yjaaidi I invited you to the org, happy to have it there until Playwright's CT is ready.

@mxschmitt any change this PR could be moved to https://github.com/playwright-community?

JessicaSachs:
I wanted to take a moment to respond here, as this PR touches on an issue that’s been critical to the community: the future of browser-based component testing in Playwright.

After four years, the community is still waiting for a decision on whether Playwright will fully commit to component testing. I understand the need for careful, long-term decision-making, but the uncertainty is holding back progress. The concerns about supporting Vue, Svelte, and other frameworks would be valid if the metrics being used to evaluate them captured the full demand. @pavelfeldman You're solely citing experimental download counts for a product you do not teach or document. C'mon man.

As the former product and technical lead for Cypress Component Testing, I’ve had a front-row seat to the innovation and development of the browser-based testing space. Since leaving Cypress, I've been involved in the development of Vitest and its browser modes. Throughout that time I've consistently heard from developers across frameworks about the demand for browser-based component testing. The tools they use—Testing Library, Storybook, Vue Test Utils, and Enzyme—show just how important this feature is to the broader community.

Screenshot 2024-10-14 at 6 28 13 PM

Every time I teach a workshop, I hear the same question: “When will Playwright component testing go stable?” Recently, at VueConf US, a team lead from General Electric expressed their frustration at having to wait for a feature that they already had with Cypress before migrating to Playwright. The demand is clear—developers are ready for Playwright to catch up, but the lack of a stable component testing feature is frustrating.

This PR is especially significant because critical the Angular community and core team members spent a year working on and waiting for this effort to be completed. Angular has always been at the forefront of JavaScript testing, and their contribution here reflects how important this feature is. If Playwright isn’t going to move forward with first-party support for component testing, the community deserves to know, so we can focus on building the tools that will meet our needs.

There's need to see code in a browser and to be able to dispatch real events to the side effects (renders/events) created in the DOM. **Playwright is currently the BEST and MOST COMPOSABLE tool to create that browser environment and dispatch those actions.** I'm betting that Microsoft's priorities and staffing may not align with the community's needs. Which is fine, but leads me to ask for closure. We can read the room, but it would help everyone out a lot in their respective technical discussions and the way people choose to spend their time if the team explicitly signaled their intent for the feature.

Whether Playwright decides to fully commit to component testing or officially deprecate it, the important thing is giving developers the clarity they need to move forward. We’ve waited for four years, and developers need to know where to invest their time and resources.

I’m rooting for progress, whether that’s through Playwright or by empowering the community to build new tooling. Thank you for your time and consideration, and for all the work that’s gone into this so far.

sand4rt:
Hey @JessicaSachs, thank you for your response and your work on browser based component testing! I think we're on the same page here. For the past 2.5 years, i’ve been the primary force pushing Playwright component testing forward _(as part time volunteer)_ because i believe in it's potential. I completely understand the frustrations—believe me ;)

shilman:
A lot of folks here are disappointed that they cannot use Playwright CT on non-React renderers. Indeed, Playwright is the gold standard for testing pages and flows, so why wouldn't you want to test your isolated components using the same best-in-class tooling?

However, today you can already do CT with Playwright for Angular, Vue, Svelte, and much more. At Storybook, we've built our own [CT workflow based on Playwright](https://storybook.js.org/docs/writing-tests/test-runner) that you can use with any renderer that Storybook supports. It's also possible to test stories directly in Playwright as @Tallyb mentions above. 

In addition, we've partnered with the Vitest team on a next-generation runner, which also uses Playwright under the hood, but achieves mind-bending performance and built-in coverage support thanks to great work by the Vitest team. Please check out @yannbf ’s ViteConf talk for details:

https://viteconf.org/24/replay/storybook

If you're interested in figuring out CT for Angular with us, please hit me up on the [Storybook Discord](https://discord.com/invite/storybook) and we'd love to continue to build on the incredible foundation that Playwright provides.

yjaaidi:
Hey @JessicaSachs! It's been a while 😊

I agree with you that this is an egg and kitchen situation. I am also observing lots of people at the gate of Playwright CT waiting for the `experimental` prefix to drop before giving it a try.

I also understand the challenges that come with the current implementation of `experimental-ct-core` and the assumptions it makes that work well for some frameworks but less for others. While there is nothing insurmountable, there are also options that are framework agnostic, less intrusive to Playwright and thus low maintenance.
I would love to get the chance to discuss them with the team. 😅

JessicaSachs:
Thanks for the active responses. A lot of us have worked together directly in the past. I know almost everyone in this thread from one context or another and appreciate all of you.

> In addition, we've partnered with the Vitest team on a next-generation runner, which also uses Playwright under the hood, but achieves mind-bending performance and built-in coverage support thanks to great work by the Vitest team. Please check out @yannbf ’s ViteConf talk for details:

Yes! The Vitest Browser Mode effort and Storybooks what I was referring to when I said "the community carving out new tooling" and I've seen the great collaboration efforts happening between the two projects (this all happens within Discord - if others want to tune in). There's challenges in asking people to marry a particular stack and test runner, but it at least gets the job done.

The dismissiveness I saw above and incorrect stats with regard to community need is what triggered me to write up a long-form post _here_.

The tea-leaf readings I have to provide to normal application developers is what triggered me to ask for explicitness and commitment to a roadmap or vision general. Any public statement to wrap up the conversation addressing the shift in vision and scope of Playwright CT would be _very welcome_. **No rush**, just putting the ask out there. Even among folks maintaining and contributing to Testing Library and Vitest the plan for Playwright CT is unclear and you have to bring people "up to speed".

From what I can tell, it seems like collaborating with the Vitest and Storybook teams on Vitest Browser Mode and a Playwright-wrapped Storybook is the technical direction forward for this problem in the ecosystem.

That's all I have to communicate. I appreciate every single team and person who shows up to move the needle forward in the space. I look forward to seeing what the community will build, and I'll see you all within our various Discords 💟

Francesco-Borzi:
@pavelfeldman I hope you can reconsider this decision and merge this PR.

pfteter:
I'm concluding here, please correct me if i'm wrong. 

Since PW has deprecated CTs for vue2 and svelte in 1.49 and:
https://playwright.dev/docs/release-notes#version-149
https://github.com/microsoft/playwright/pull/27783#issuecomment-2400533724

I guess it's now a community effort ?
https://github.com/jscutlery/devkit/tree/main/packages/playwright-ct-angular

sand4rt:
> I'm concluding here, please correct me if i'm wrong.
> 
> Since PW has deprecated CTs for vue2 and svelte in 1.49 and: https://playwright.dev/docs/release-notes#version-149 [#27783 (comment)](https://github.com/microsoft/playwright/pull/27783#issuecomment-2400533724)

Vue2 and **Solid** will be no longer updated: https://playwright.dev/docs/release-notes#other-breaking-changes. Vue2 is end of life anyway and there was not much demand for Solid component testing. 

> I guess it's now a community effort ? https://github.com/jscutlery/devkit/tree/main/packages/playwright-ct-angular

mxschmitt provided a place for us to store the code under the community organization: https://github.com/playwright-community. However, we haven't had the time to migrate everything yet, as discussions are still ongoing regarding the best approach.

Krazybug:
So they will continue to support component testing in experimental for all the mainstream frameworks except Angular. What are their justifications as they just had to integrate the PR  ?

Martinspire:
So now that this discussion has been taking some time, isn't it about time for an update what Playwright is planning?

The way I see it, there are two paths forward:

* Either playwright provides this support themselves
* The community develops the support

For number one the solution seems simple: merge this PR and provide similar support for other frameworks to do component testing

For the second one, if this doesn't become part of playwright, then at least it should provide the necessary API's and logic for the community to build their own. Because it seems to me that enough projects want to use this that there's a use case to do at least the bare minimum to support it. If Playwright fails to add this, either people will fork it or start their own tools, and the result will both be that Playwright suffers from it. 

Sidenote: I also would like to see a solution to the known issues (about serialization) and I have a feeling this can also only be solved using the second solution. But that is a discussion for another time.

I can see why Playwright doesn't really like this situation. They are damned if they do, damned if they don't. But the way I see it, is that this is an area where a lot of development will happen in the next few years and that doing nothing is also going to achieve similar results. Personally, I think that the decision to hold off has been wrong, especially since they have had ample time to respond while this was being developed. On the other side, I think the Angular team needs to go above playwright to MS directly and have discussions there to solve it, since for them it is also an important step for the future of Angular. And if Playwright doesn't want to support it, I only see a single outcome: you guys gotta start your own E2E tool again. And I get why you guys didn't want to do that, but it seems that this is the only way the market will move forward.

Right now, this whole situation is taking too long imo

muhamedkarajic:
Still not giving off on expecting this feature.

Issue #27712, opened 2023-10-19
[BUG] Google Chrome leaves CrashPadMetrics files behind


mxschmitt:
<img width="2004" alt="image" src="https://github.com/microsoft/playwright/assets/17984549/d448ccad-97c1-451b-9a8a-1bade71ea54d">

mxschmitt:
- Its not reproducible all the time
- Its not reproducible with Puppeteer, most likely because we use rimraf and they use fs.promises.rm to remove the userDatadir
- If we change our logic to the following it works without leaving any directories behind or showing any error in the DEBUG logs:

![image](https://github.com/microsoft/playwright/assets/17984549/8e54eadc-a459-48c4-b439-02e4a8147b19)

Why does rimraf not delete the file? Because it does not retry `EPERM` errors while fs.promises.rm does. Confirmed while debugging as per here:

https://github.com/isaacs/rimraf/blob/81c631e8e27303ad7aa1f8de319463825b6ac7fc/src/retry-busy.ts#L26

PR #27790, opened 2023-10-25
chore: get rid of rimraf package

mxschmitt:
This seems more reliable nowadays as rimraf.

https://github.com/microsoft/playwright/issues/27712

### DIFF
--- a/packages/playwright-core/src/utils/fileUtils.ts
+++ b/packages/playwright-core/src/utils/fileUtils.ts
@@ -16,7 +16,6 @@
 
 import fs from 'fs';
 import path from 'path';
-import { rimraf } from '../utilsBundle';
 
 export const existsAsync = (path: string): Promise<boolean> => new Promise(resolve => fs.stat(path, err => resolve(!err)));
 
@@ -25,12 +24,10 @@ export async function mkdirIfNeeded(filePath: string) {
   await fs.promises.mkdir(path.dirname(filePath), { recursive: true }).catch(() => {});
 }
 
-export async function removeFolders(dirs: string[]): Promise<Array<Error|null|undefined>> {
-  return await Promise.all(dirs.map((dir: string) => {
-    return new Promise<Error|null|undefined>(fulfill => {
-      rimraf(dir, { maxRetries: 10 }).then(() => fulfill(undefined)).catch((e: Error) => fulfill(e));
-    });
-  }));
+export async function removeFolders(dirs: string[]): Promise<Error[]> {
+  return await Promise.all(dirs.map((dir: string) =>
+    fs.promises.rm(dir, { recursive: true, force: true, maxRetries: 10 })
+  )).catch(e => e);
 }
 
 export function canAccessFile(file: string) {
--- a/packages/playwright-core/src/utils/processLauncher.ts
+++ b/packages/playwright-core/src/utils/processLauncher.ts
@@ -15,11 +15,11 @@
  * limitations under the License.
  */
 
+import fs from 'fs';
 import * as childProcess from 'child_process';
 import * as readline from 'readline';
 import { isUnderTest } from './';
 import { removeFolders } from './fileUtils';
-import { rimraf } from '../utilsBundle';
 
 export type Env = {[key: string]: string | number | boolean | undefined};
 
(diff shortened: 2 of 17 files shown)


pavelfeldman:
Does it fix the bug?

mxschmitt:
> Does it fix the bug?

Yes see https://github.com/microsoft/playwright/issues/27712#issuecomment-1777024107 for investigation notes. rinraf was useful before Node.js had it built-in.

packages/playwright-core/src/utils/processLauncher.ts
```diff
@@ -245,7 +245,7 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
     options.log(`[pid=${spawnedProcess.pid || 'N/A'}] starting temporary directories cleanup`);
     for (const dir of options.tempDirectories) {
       try {
-        rimraf.sync(dir);
+        fs.rmSync(dir, { force: true, recursive: true, maxRetries: 10 });
```
dgozman, line 248: I wonder how does the sync version retry? Perhaps we should limit the number of retries here, say to 5, because max total time is quadratic?

dgozman (approved):

packages/playwright-core/src/utils/processLauncher.ts
mxschmitt, line 248:
```suggestion
        fs.rmSync(dir, { force: true, recursive: true, maxRetries: 5 });
```

packages/playwright-core/src/utils/processLauncher.ts
```diff
@@ -245,7 +245,7 @@ export async function launchProcess(options: LaunchProcessOptions): Promise<Laun
     options.log(`[pid=${spawnedProcess.pid || 'N/A'}] starting temporary directories cleanup`);
     for (const dir of options.tempDirectories) {
       try {
-        rimraf.sync(dir);
+        fs.rmSync(dir, { force: true, recursive: true, maxRetries: 5 });
```
mxschmitt, line 248:
```suggestion
        fs.rmSync(dir, { force: true, recursive: true, maxRetries: 10 });
```

packages/playwright-core/src/utils/processLauncher.ts
mxschmitt, line 248:
```suggestion
        fs.rmSync(dir, { force: true, recursive: true, maxRetries: 5 });
```

mxschmitt, line 248:
```suggestion
        fs.rmSync(dir, { force: true, recursive: true, maxRetries: 5 });
```

packages/playwright-core/src/utils/fileUtils.ts
```diff
@@ -25,12 +24,10 @@ export async function mkdirIfNeeded(filePath: string) {
...
-export async function removeFolders(dirs: string[]): Promise<Array<Error|null|undefined>> {
-  return await Promise.all(dirs.map((dir: string) => {
-    return new Promise<Error|null|undefined>(fulfill => {
-      rimraf(dir, { maxRetries: 10 }).then(() => fulfill(undefined)).catch((e: Error) => fulfill(e));
-    });
-  }));
+export async function removeFolders(dirs: string[]): Promise<Error[]> {
+  return await Promise.all(dirs.map((dir: string) =>
+    fs.promises.rm(dir, { recursive: true, force: true, maxRetries: 10 })
+  )).catch(e => e);
```
itchyny, line 30: This looks like a mistake of `.catch(e => e)));`, i.e. should catch each error of `fs.promises.rm`. Otherwise the return type should be `Promise<Error>`.

itchyny:
```
const fs = require('fs');
const dirs = ['/tmp/x', '/tmp/y'];

(async () => {
  await Promise.all(dirs.map((dir) => Promise.reject({ dir }))).catch(e => e).then(x => console.log(x));
  // { dir: '/tmp/x' }

  await Promise.all(dirs.map((dir) => Promise.reject({ dir }).catch(e => e))).then(x => console.log(x));
  // [ { dir: '/tmp/x' }, { dir: '/tmp/y' } ]
})();
```

itchyny: As a result, the following line of `packages/playwright/src/runner/tasks.ts` yields `TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))`.
```
await Promise.all(Array.from(outputDirs).map(outputDir => removeFolders([outputDir]).then(async ([error]) => {
```

mxschmitt: Good catch, are you interested in contributing a fix?

itchyny: Yes of course, I opened a pull request: https://github.com/microsoft/playwright/pull/28239.

PR #27807, opened 2023-10-26
chore: stream trace viewer logs

pavelfeldman:


packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -368,6 +368,13 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
     return this._captureSnapshot(event.inputSnapshot, sdkObject, metadata, element);
   }
 
+  onCallLog(sdkObject: SdkObject, metadata: CallMetadata, logName: string, message: string) {
+    if (logName !== 'api')
+      return;
+    const event = createActionLogTraceEvent(metadata, message);
+    this._appendTraceEvent(event);
```
dgozman, line 375: Let's not flush log events in non-live mode? See `_appendTraceEvent`.

packages/trace-viewer/src/traceModel.ts
```diff
@@ -17,8 +17,9 @@
 import type * as trace from '@trace/trace';
 import type * as traceV3 from './versions/traceV3';
 import type * as traceV4 from './versions/traceV4';
+import type * as traceV5 from './versions/traceV5';
```
dgozman, line 20: Missing the file I guess.

packages/trace-viewer/src/traceModel.ts
```diff
@@ -238,16 +249,16 @@ export class TraceModel {
...
+      return [event];
+    const lastVersion: trace.VERSION = 6;
     for (let version = this._version; version < lastVersion; ++version) {
       event = (this as any)[`_modernize_${version}_to_${version + 1}`].call(this, event);
       if (!event)
-        return null;
+        return [];
     }
-    return event;
+    return Array.isArray(event) ? event : [event];
```
dgozman, line 261: I guess we have to change all modernizers to return an array now, to avoid this hack.

packages/trace-viewer/src/traceModel.ts
```diff
@@ -184,11 +188,18 @@ export class TraceModel {
         existing!.point = event.point;
         break;
       }
+      case 'log': {
+        const existing = actionMap.get(event.callId);
+        existing!.log.push({
+          time: event.time,
```
dgozman, line 194:
```suggestion
          time: event.time === -1 ? existing.endTime : event.time,
```

packages/trace-viewer/src/traceModel.ts
```diff
@@ -384,6 +395,17 @@ export class TraceModel {
     }
     return event;
   }
+
+  _modernize_5_to_6(event: traceV5.TraceEvent): trace.TraceEvent | trace.TraceEvent[] | null {
+    if (event.type !== 'after' || !event.log.length)
+      return event;
+    return event.log.map(log => ({
```
dgozman, line 402: This seems to drop the actual `after` event on the floor?

packages/trace-viewer/src/traceModel.ts
pavelfeldman: I don't think we should do that.

dgozman (approved):

PR #27815, opened 2023-10-26
fix(html): rendered '0' in the header

mxschmitt:
This regressed in https://github.com/microsoft/playwright/commit/c8134bca5df3f6e460c5c7f30b0e421dd9c6cec6.

Note: 00 && <Some Element /> will render a 0 in React. Always check for length > 0.

Drive-by: Aligned it to `foo.length > 0` check while `!!foo.length` works too, just to keep it consistent.


dgozman (approved):

mxschmitt:
superseded by https://github.com/microsoft/playwright/commit/86c78c1e0fd97f0a32ccd94e438f13fe0f0e1f1e

Issue #6500, opened 2021-05-11
[BUG] Running Nightly as root in a regular user's session is not supported.


Th3S4mur41:
Running playwright automation in Github CI using docker image fails for Firefox

Github Actions Workflow looks as follow:
```
jobs:
  test_e2e:
    runs-on: self-hosted
    container:
      image: mcr.microsoft.com/playwright
      options: '--ipc=host --entrypoint /bin/bash'
    steps:
      - uses: actions/checkout@v2
      - run: npm ci --no-optional --no-audit --prefer-offline --progress=false
      - run: npm run ci:e2e
```

Output is: 

```
 npm run ci:e2e
  shell: sh -e {0}

> @clap/clap-frontend@0.1.0 ci:e2e /__w/clap-ui/clap-ui
> jest -c jest-e2e.config.js

Protocol error (Browser.enable): Browser closed.
==================== Browser output: ====================
<launching> /ms-playwright/firefox-1250/firefox/firefox -no-remote -headless -profile /tmp/playwright_firefoxdev_profile-8PNwPz -juggler-pipe -silent
<launched> pid=629
[pid=629][err] Running Nightly as root in a regular user's session is not supported.  ($HOME is /github/home which is owned by pwuser.)Error
    at /__w/clap-ui/clap-ui/node_modules/playwright/lib/server/firefox/ffConnection.js:55:63
    at new Promise (<anonymous>)
    at FFConnection.send (/__w/clap-ui/clap-ui/node_modules/playwright/lib/server/firefox/ffConnection.js:54:16)
    at Function.connect (/__w/clap-ui/clap-ui/node_modules/playwright/lib/server/firefox/ffBrowser.js:65:24)
    at Firefox._connectToTransport (/__w/clap-ui/clap-ui/node_modules/playwright/lib/server/firefox/firefox.js:65:38)
    at Firefox._innerLaunch (/__w/clap-ui/clap-ui/node_modules/playwright/lib/server/browserType.js:133:36)
    at async ProgressController.run (/__w/clap-ui/clap-ui/node_modules/playwright/lib/server/progress.js:76:28)
    at async Firefox.launch (/__w/clap-ui/clap-ui/node_modules/playwright/lib/server/browserType.js:78:25)
    at async BrowserServerLauncherImpl.launchServer (/__w/clap-ui/clap-ui/node_modules/playwright/lib/browserServerImpl.js:35:25)
[...]

PR #27832, opened 2023-10-27
Add HOME=/root to container samples in docs

j2ghz:
Reference: #6500 (https://github.com/microsoft/playwright/issues/6500#issuecomment-1664210041)

It seems that this is now required, so this PR applies that to the docs.


dgozman (approved):
Thank you for the PR, looks great!

Issue #27792, opened 2023-10-25
[internal] file upload Android tests are failing


mxschmitt:
Since: https://github.com/microsoft/playwright/pull/27671

https://github.com/microsoft/playwright-browsers/actions/workflows/tests-android.yml

PR #27835, opened 2023-10-27
fix: allow uploading large files on Android

mxschmitt:
https://github.com/microsoft/playwright/issues/27792

### DIFF
--- a/packages/playwright-core/src/server/android/android.ts
+++ b/packages/playwright-core/src/server/android/android.ts
@@ -21,7 +21,8 @@ import os from 'os';
 import path from 'path';
 import type * as stream from 'stream';
 import { wsReceiver, wsSender } from '../../utilsBundle';
-import { createGuid, makeWaitForNextTask, isUnderTest, getPackageManagerExecCommand } from '../../utils';
+import type { RegisteredListener } from '../../utils';
+import { createGuid, makeWaitForNextTask, isUnderTest, getPackageManagerExecCommand, eventsHelper } from '../../utils';
 import { removeFolders } from '../../utils/fileUtils';
 import type { BrowserOptions, BrowserProcess } from '../browser';
 import type { BrowserContext } from '../browserContext';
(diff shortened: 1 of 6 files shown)


packages/playwright-core/src/server/android/android.ts
```diff
@@ -343,6 +345,35 @@ export class AndroidDevice extends SdkObject {
     validateBrowserContextOptions(options, browserOptions);
 
     const browser = await CRBrowser.connect(this.attribution.playwright, androidBrowser, browserOptions);
+    {
+      const androidArtifacts = new Set<string>();
+      browser._interceptSetInputFilePaths = async (page, files) => {
+        const newLocalFiles = [];
+        for (const file of files) {
+          const localFileName = `/data/local/tmp/pw/${path.basename(file)}`;
```
yury-s, line 353: this may clash with other file names

packages/playwright-core/src/server/android/android.ts
```diff
@@ -343,6 +345,35 @@ export class AndroidDevice extends SdkObject {
     validateBrowserContextOptions(options, browserOptions);
 
     const browser = await CRBrowser.connect(this.attribution.playwright, androidBrowser, browserOptions);
+    {
+      const androidArtifacts = new Set<string>();
+      browser._interceptSetInputFilePaths = async (page, files) => {
```
yury-s, line 350: This is quite a hack. I'd rather get createTempFile to create files on the android device, wdyt?

packages/playwright-core/src/server/chromium/crPage.ts
```diff
@@ -333,6 +333,8 @@ export class CRPage implements PageDelegate {
   }
 
   async setInputFilePaths(progress: Progress, handle: dom.ElementHandle<HTMLInputElement>, files: string[]): Promise<void> {
+    if (this._browserContext._browser._interceptSetInputFilePaths)
```
yury-s, line 336: I'd rather revert the change that fixed lastModified, this becomes too hairy. Maybe we can move `convertInputFiles` logic to the server and let it properly handle the android case?

PR #27841, opened 2023-10-27
fix(css parser): support nested builtin functions

### LINKED ISSUE #27743: [BUG] locator.count: Error while parsing selector ":nth-last-child(1 of ...) ~ *"
Maybe it's not the bug but not yet supported selector by Playwright so sorry in advance 🙏 
I skimmed through the Stack Overflow and a couple of articles but couldn't find a similar issue.

### System info
- Playwright Version: [v1.XX] - default in playground
- Operating System: [All, Windows 11, Ubuntu 20, macOS 13.2, etc.] - MacOS 14, but ran in the playground
- Browser: [All, Chromium, Firefox, WebKit] - Chrome
- Other info:

### Source code

**Link to the test file**

https://try.playwright.tech/?l=javascript&s=khc2kb8

**Link to the JSFiddle with CSS selector usage**
https://jsfiddle.net/654t7skb/9/

**Steps**
- Open the JSFiddle, see that selector successfully highlights some elements
- Run the Playwright playground file, see that it throws an error after being unable to parse that selector

**Expected**

`nth-last-child( )` selector is supported

**Actual**

`nth-last-child( )` selector is NOT supported

dgozman:
Things like `:nth-child(1 of :has(span:nth-last-child(3)))`.

Fixes #27743.

### DIFF
--- a/packages/playwright-core/src/utils/isomorphic/cssParser.ts
+++ b/packages/playwright-core/src/utils/isomorphic/cssParser.ts
@@ -98,8 +98,16 @@ export function parseCSS(selector: string, customNames: Set<string>): { selector
     return tokens[p] instanceof css.CommaToken;
   }
 
+  function isOpenParen(p = pos) {
+    return tokens[p] instanceof css.OpenParenToken;
+  }
+
   function isCloseParen(p = pos) {
     return tokens[p] instanceof css.CloseParenToken;
+  }
+
+  function isFunction(p = pos) {
+    return tokens[p] instanceof css.FunctionToken;
   }
 
   function isStar(p = pos) {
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/utils/isomorphic/cssParser.ts
```diff
@@ -221,14 +225,22 @@ export function parseCSS(selector: string, customNames: Set<string>): { selector
 
   function consumeBuiltinFunctionArguments(): string {
     let s = '';
-    while (!isCloseParen() && !isEOF())
+    let balance = 1;  // First open paren is a part of a function token.
+    while (!isEOF()) {
+      if (isOpenParen() || (tokens[pos] instanceof css.FunctionToken))
```
yury-s, line 230: nit: introduce `isFunctionToken()` as there is at least one more usage of it

yury-s (approved):

PR #27843, opened 2023-10-27
chore: render full pathname in network panel

### LINKED ISSUE #27618: [BUG] UI Mode Network tab request value folded to unredable form
When I check Network tab for info about used resource I observe very limited (hard to read / find) info in Request column.
obraz
Compare same with Chrome (with Big request rows on)
obraz

I know about tooltips but still this look like a bug.

### System info
- Playwright Version: v1.39
- Operating System: All

### Source code

- [x ] I provided exact source code that allows reproducing the issue locally.

```js
import { expect, test } from "@playwright/test";
test("has exact title", async ({ page }) => {
  const expectedTitle = "6. Dodatkowy front-end do API";

  await page.goto("https://naukaapi.testoneo.com/6/");
  const observedTitle = page.locator("#title");

  await expect(observedTitle).toHaveText(expectedTitle);
});
```

**Steps**
- Open UI Mode 
- Run test
- Navigate to Network tab

**Expected**
I wish to have more info (instead just tooltip) regarding URL path in Request column, see Chrome example 
obraz

**Actual**
Just end of the path after last `/` what ends in many `/` without additional info
obraz

pavelfeldman:
Fixes #27618

### DIFF
--- a/packages/trace-viewer/src/ui/networkTab.tsx
+++ b/packages/trace-viewer/src/ui/networkTab.tsx
@@ -129,7 +129,8 @@ const NetworkResource: React.FunctionComponent<{
 }> = ({ resource, boundaries }) => {
   const { routeStatus, resourceName, contentType } = React.useMemo(() => {
     const routeStatus = formatRouteStatus(resource);
-    const resourceName = resource.request.url.substring(resource.request.url.lastIndexOf('/'));
+    const url = new URL(resource.request.url);
+    const resourceName = url.pathname;
     let contentType = resource.response.content.mimeType;
     const charset = contentType.match(/^(.*);\s*charset=.*$/);
     if (charset)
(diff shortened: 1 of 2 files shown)


packages/trace-viewer/src/ui/networkTab.tsx
```diff
@@ -129,7 +129,8 @@ const NetworkResource: React.FunctionComponent<{
 }> = ({ resource, boundaries }) => {
   const { routeStatus, resourceName, contentType } = React.useMemo(() => {
     const routeStatus = formatRouteStatus(resource);
-    const resourceName = resource.request.url.substring(resource.request.url.lastIndexOf('/'));
+    const url = new URL(resource.request.url);
```
dgozman, line 132: `try/catch` it just in case?

dgozman (approved):

Issue #23542, opened 2023-06-06
[BUG] VS Code extension loads incorrect config file and silently fails


OliverJAsh:
### System info
- Playwright Version: 1.34.4
- Operating System: macOS 13.4
- Browser: N/A
- Other info:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Link to the GitHub repository with the repro**

https://github.com/OliverJAsh/playwright-vs-code-extension-list-files-issue

This is a brand new Playwright project created using `yarn create playwright`. The only changes on top of this are:

- [Use `dotenv`](https://github.com/OliverJAsh/playwright-vs-code-extension-list-files-issue/commit/d884ab90be92199401cf173c278d098cd1a26711)
- [Add `tsconfig.json`](https://github.com/OliverJAsh/playwright-vs-code-extension-list-files-issue/commit/43d5c5ee6c7c5749d4565e9af4e7bfe030d0d8ff)

**Steps**
1. Clone the project
1. ```console
   $ yarn
   $ tsc
   ```
1. Open VS Code

**Expected**

The tests load, or if the tests fail to load, an error message should appear.

**Actual**

The tests never load, nor is there an error message to indicate an error occurred.

<img width="444" alt="image" src="https://github.com/microsoft/playwright/assets/921609/d8286032-32d3-402a-90fa-06c381e62694">

**Notes**

This issue started when I upgraded from 1.33.0 to 1.34.4.

I spent some time trying to debug this. In developer tools we can see this error message: `e.projects is not iterable`.

![image](https://github.com/microsoft/playwright/assets/921609/b1a6c814-b3ee-44ef-acfb-e45d5a4ded45)

I modified the extension to log the value of `e`:

<img width="788" alt="image" src="https://github.com/microsoft/playwright/assets/921609/5522ac2f-c1dc-451a-9b00-f35cc562b8eb">

It appears that when the config file is loaded via `listFiles` ([code](https://github.com/microsoft/playwright-vscode/blob/be9ad1e93fff6b7bba9ab44d42c648b91b85608f/src/playwrightTest.ts#L97)), the `FOO` environment variable does not exist. This suggests that `dotenv` is not working.

To debug this further I added a console log inside `listFiles` to see the value of the `configFolder` argument we pass to `runNode`:

<img width="786" alt="image" src="https://github.com/microsoft/playwright/assets/921609/e35f8900-8751-4f63-b0bd-bba48f5a4d3d">

[...]

PR #27846, opened 2023-10-28
Handle tsconfig paths without a baseUrl

BPScott:
Partial fix for #27933

Since TS 4.1, a tsconfig may specify paths without requiring a baseUrl. In that case, paths should be resolved relative to the tsconfig file the paths were specified in.

See https://www.typescriptlang.org/tsconfig#paths
See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#paths-without-baseurl

This PR makes playwright handle cases where paths are specified without a baseUrl, such as:

```json
{
  "compilerOptions": {
    "paths": {
      "~/*": ["./app/*"],
    }
  }
}
```

---

There is an edge case that this doesn't handle - if you specify paths without a baseUrl in an config that is extended from then the paths should be relative to the file that paths was specified in (i.e. the path looks right in the file that it was written in). In other words, adding the following test case to `tests/playwright-test/resolver.spec.ts` should pass, but it does not:

```ts
test('should use the location of the extended tsconfig as paths root when no baseUrl is specified', async ({ runInlineTest }) => {
    'tsconfig.json': `{
      "extends": ["./other/tsconfig.base.json"],
    }`,
    'other/tsconfig.base.json': `{
      "compilerOptions": {
        // paths without a baseUrl should resolve relative to the file they're in, which means
        // the base for these paths should be "other/". However playwright will currently
        // resolve them relative to the top-level tsconfig, not the extended from config
        // Note that adding baseurl:'.', fixes this within playwright but has the undesireable
        // effect of enabling the wildcard matching for additional top-level folders in the
        // "other" directory
        // "baseUrl": ".",
        "paths": {
          "util/*": ["../dir/foo/bar/util/*"],
        },
      },
    }`,
    'a.test.ts': `
      const { foo } = require('util/file');
      test('test', ({}, testInfo) => {
        expect(foo).toBe('foo');
      });
    `,
    'dir/foo/bar/util/file.ts': `
      module.exports = { foo: 'foo' };
    `,
  });

});
```

To fix this edge case would require greater refactoring than I'm comfortable doing. I'm ok with the idea that this PR is an incremental improvement - we fix paths without a baseUrl in base tsconfigs, while leaving paths without a baseUrl in extended configs as not working for the moment"

### DIFF
--- a/packages/playwright/src/transform/transform.ts
+++ b/packages/playwright/src/transform/transform.ts
@@ -58,11 +58,14 @@ export function transformConfig(): TransformConfig {
 }
 
 function validateTsConfig(tsconfig: TsConfigLoaderResult): ParsedTsConfigData | undefined {
-  if (!tsconfig.tsConfigPath || !tsconfig.baseUrl)
+  if (!tsconfig.tsConfigPath)
     return;
   // Make 'baseUrl' absolute, because it is relative to the tsconfig.json, not to cwd.
-  const absoluteBaseUrl = path.resolve(path.dirname(tsconfig.tsConfigPath), tsconfig.baseUrl);
-  const pathsFallback = [{ key: '*', values: ['*'] }];
+  // When no explicit baseUrl is set, resolve paths relative to the tsconfig file.
+  // See https://www.typescriptlang.org/tsconfig#paths
+  const absoluteBaseUrl = path.resolve(path.dirname(tsconfig.tsConfigPath), tsconfig.baseUrl ?? '.');
+  // Only add the catch-all mapping when baseUrl is specified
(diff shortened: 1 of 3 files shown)


BPScott:
Lint error seems to be the result of a change committed in #27844, unrelated to this change

BPScott:
ping @dgozma, as you've dealt with other tsconfig path/baseurl related issues over in #23542

pavelfeldman (approved):

pavelfeldman:
Awesome fix, thank you!

BPScott:
Thanks for the merge @pavelfeldman!

PR #27880, opened 2023-10-31
fix: Google Chrome Beta/Stable install on macOS

### LINKED ISSUE #27764: [BUG] Chrome Beta install fails on macOS
### System info
- Playwright Version: v1.39.0
- Operating System: macOS 14.0 Sonoma (23A344)
- Browser: Chromium
- Other info: https://bugs.chromium.org/p/chromium/issues/detail?id=1487920&q=sonoma%20timezone&can=2

### Source code

N/A

**Steps**

Playwright 1.39.0 installs Chromium 119.0.6045.9 (Developer Build) (arm64), which suffers from the above Chromium bug.

Launch a playwright test and in the Chromium developer console, on OSX Sonoma, run: 

```js
new Intl.DateTimeFormat().resolvedOptions().timeZone
```

**Expected**

This should not be `undefined`.

**Actual**

It is `undefined`.

From the Chromium issue above, it appears to have been fixed in the normal _Chrome_ browser as of version `118.0.5993.96`.

However, the version of Chromium pulled down by Playwright (119.0.6045.9) appears to be unfixed still. Is it possible for Playwright to bump your Chromium version? Or for us to specify a bleeding-edge version to get around this?

Thanks

mxschmitt:
This pull request addresses and resolves an issue where applications, such as Google Chrome Beta, became unusable after being copied via the Terminal. This issue was primarily due to the loss of extended attributes, specifically the `com.apple.quarantine` attribute, during the copying process.

#### Changes:

- Updated the copying command in our script to use `cp -pR`.

```bash
cp -pR "/Volumes/googlechromebeta.dmg/Google Chrome Beta.app" /Applications/
```

#### Background:

When files are downloaded from the internet, macOS tags them with the `com.apple.quarantine` attribute as a security measure. This attribute triggers Gatekeeper to verify the application's integrity, signature, and source before the application is allowed to run. 

We discovered that when copying applications via the Terminal using a plain `cp` command, essential extended attributes like `com.apple.quarantine` were not preserved, leading macOS to consider the application broken or damaged.

#### Benefits of `-pR` Options:

- **`-p` (Preserve):** Ensures that the extended attributes, including `com.apple.quarantine`, and Access Control Lists (ACLs) are preserved during the copy, maintaining the file’s integrity and original security attributes.
  
- **`-R` (Recursive):** Ensures that directories, or in the case of macOS, application bundles, are copied recursively, preserving the structure and content.

#### Conclusion:

By incorporating the `cp -pR` command, we maintain the necessary security attributes and metadata, ensuring that applications remain functional and recognized as secure by macOS after the copying process. This approach aligns the behavior of the copying process closer to that observed when using the Finder’s GUI for copying, thus ensuring consistent application behavior and integrity.

Drive-by: remove the `-f`, since we before removed the folder already.

Fixes https://github.com/microsoft/playwright/issues/27764

### DIFF
--- a/packages/playwright-core/bin/reinstall_chrome_beta_mac.sh
+++ b/packages/playwright-core/bin/reinstall_chrome_beta_mac.sh
@@ -6,7 +6,7 @@ rm -rf "/Applications/Google Chrome Beta.app"
 cd /tmp
 curl -o ./googlechromebeta.dmg -k https://dl.google.com/chrome/mac/universal/beta/googlechromebeta.dmg
 hdiutil attach -nobrowse -quiet -noautofsck -noautoopen -mountpoint /Volumes/googlechromebeta.dmg ./googlechromebeta.dmg
-cp -rf "/Volumes/googlechromebeta.dmg/Google Chrome Beta.app" /Applications
+cp -pR "/Volumes/googlechromebeta.dmg/Google Chrome Beta.app" /Applications
 hdiutil detach /Volumes/googlechromebeta.dmg
 rm -rf /tmp/googlechromebeta.dmg
 
--- a/packages/playwright-core/bin/reinstall_chrome_stable_mac.sh
+++ b/packages/playwright-core/bin/reinstall_chrome_stable_mac.sh
@@ -6,7 +6,7 @@ rm -rf "/Applications/Google Chrome.app"
 cd /tmp
 curl -o ./googlechrome.dmg -k https://dl.google.com/chrome/mac/universal/stable/GGRO/googlechrome.dmg
 hdiutil attach -nobrowse -quiet -noautofsck -noautoopen -mountpoint /Volumes/googlechrome.dmg ./googlechrome.dmg
-cp -rf "/Volumes/googlechrome.dmg/Google Chrome.app" /Applications
+cp -pR "/Volumes/googlechrome.dmg/Google Chrome.app" /Applications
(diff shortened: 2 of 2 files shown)


packages/playwright-core/bin/reinstall_chrome_beta_mac.sh
```diff
@@ -6,7 +6,7 @@ rm -rf "/Applications/Google Chrome Beta.app"
 cd /tmp
 curl -o ./googlechromebeta.dmg -k https://dl.google.com/chrome/mac/universal/beta/googlechromebeta.dmg
 hdiutil attach -nobrowse -quiet -noautofsck -noautoopen -mountpoint /Volumes/googlechromebeta.dmg ./googlechromebeta.dmg
-cp -rf "/Volumes/googlechromebeta.dmg/Google Chrome Beta.app" /Applications
+cp -pR "/Volumes/googlechromebeta.dmg/Google Chrome Beta.app" /Applications
```
dgozman, line 9: Should we use `ditto` instead? Did you manage to reproduce the issue?

yury-s (approved):

packages/playwright-core/bin/reinstall_chrome_beta_mac.sh
mxschmitt: Yes was able to reproduce. Both will work, I’d vote for cp, since its more gnu like.

PR #24486, opened 2023-07-28
devops: fix parameterised tests on flakiness dashboard

mxschmitt:
Fixes https://github.com/microsoft/devops.playwright.dev/issues/2

PR #27900, opened 2023-11-01
chore: fix parameterized tests on the flakiness dashboard

mxschmitt:
Abstract issue: I have parameterised tests via the titlePath (describe) and they are not shown on the flakiness dashboard. Before #24486 only title() was shown, after #24486 titlePath() will be displayed, but we still did have the testId based on the test.title() and not test.titlePath(). This ends up that they will still be treated as a single spec.

After this change they are not treated as a single spec anymore and treated as different ones:

image

Note: This is tested.

Follow-up on https://github.com/microsoft/playwright/pull/24486.

### DIFF
--- a/utils/flakiness-dashboard/package.json
+++ b/utils/flakiness-dashboard/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "",
+  "name": "flakiness-dashboard",
   "version": "",
   "description": "",
   "scripts": {
@@ -7,6 +7,6 @@
   },
   "author": "",
   "dependencies": {
-    "@azure/storage-blob": "^12.2.1"
+    "@azure/storage-blob": "^12.16.0"
   }
 }
--- a/utils/flakiness-dashboard/processing/dashboard_compressed_v1.js
+++ b/utils/flakiness-dashboard/processing/dashboard_compressed_v1.js
@@ -60,7 +60,7 @@ function compressReports(reports) {
         specs = new Map();
         files[spec.file] = specs;
       }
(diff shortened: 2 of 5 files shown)


utils/flakiness-dashboard/package.json
```diff
@@ -1,12 +1,12 @@
...
+  "name": "flakiness-dashboard",
   "version": "",
   "description": "",
   "scripts": {
     "test": "echo \"No tests yet...\""
   },
   "author": "",
   "dependencies": {
-    "@azure/storage-blob": "^12.2.1"
+    "@azure/storage-blob": "^12.16.0"
```
mxschmitt, line 10: I believe a healthy drive by.

yury-s (approved):

utils/flakiness-dashboard/processing/dashboard_compressed_v1.js
```diff
@@ -60,7 +60,7 @@ function compressReports(reports) {
         specs = new Map();
         files[spec.file] = specs;
       }
-      const specId = spec.file + '---' + spec.title + ' --- ' + spec.line;
+      const specId = spec.file + '---' + spec.titlePath.join('-') + ' --- ' + spec.line;
```
yury-s, line 63: Can we use same path delimiter when joining (it was `>` in https://github.com/microsoft/playwright/pull/24486) ?

yury-s: Make sure titlePath does not include project

PR #27920, opened 2023-11-02
fix: use non-relative paths in dynamic imports

### LINKED ISSUE #27919: [BUG] Dynamic imports with relative paths used in Playwright source code break compiled files
### System info
- Playwright Version: [v1.39] 
- Operating System: [Not Applicable]
- Browser: [Not Applicable]
- Other info:
  - This issue goes at least as far as [v1.30] ([PR](https://github.com/microsoft/playwright/commit/df05c00da36f483acec0bfe329e7fee095ad967c#diff-aeaef623ef2ef94fe741ea76569566ce7999d3626999115ed40bf71e1484a2c5R32))

### Source code

[https://github.com/duckception/playwright-relative-path-issue](https://github.com/duckception/playwright-relative-path-issue)

**Steps, Explanation, Expected & Actual Behaviour**
- See [README](https://github.com/duckception/playwright-relative-path-issue/blob/master/README.md) in the repro repo.

**Description**

I'm working on a tool built on top of Playwright, however, in the fixtures I rely on compiled versions of TypeScript files, which often utilize type imports from the `@playwright/test` and `playwright-core` packages. Currently, this is not working out of the box, and I have to patch the compiled files manually. This is because, in the source code of Playwright, you utilize dynamic imports with relative paths to `package.json` files. This is problematic because these dynamic imports with relative paths end up in the compiled versions of the files with the same format.

For example, if I have a file with `import type { Page } from '@playwright/test'` and I compile it as seen in the repro repo, the compiled file will contain a dynamic import: `__require2.resolve("../package.json");`. Now, if I put this compiled file in a directory that does not have access to the `../package.json` file with this path it errors with: `Error: Cannot find module '../package.json'`.

Editing compiled files through a script is one way to handle this, but this is only a short-term solution, and it should be handled in the Playwright source code itself.

duckception:
Fixes #27919

### DIFF
--- a/packages/playwright-core/src/cli/program.ts
+++ b/packages/playwright-core/src/cli/program.ts
@@ -31,12 +31,19 @@ import type { Page } from '../client/page';
 import type { BrowserType } from '../client/browserType';
 import type { BrowserContextOptions, LaunchOptions } from '../client/types';
 import { spawn } from 'child_process';
-import { wrapInASCIIBox, isLikelyNpxGlobal, assert, gracefullyProcessExitDoNotHang, getPackageManagerExecCommand } from '../utils';
+import {
+  wrapInASCIIBox,
+  isLikelyNpxGlobal,
+  assert,
+  gracefullyProcessExitDoNotHang,
+  getPackageManagerExecCommand,
+  getPackageJsonPath
+} from '../utils';
 import type { Executable } from '../server';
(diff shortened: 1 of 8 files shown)


duckception:

pavelfeldman:
I don't think we would want to do something like this.

PR #27939, opened 2023-11-02
fix(msedge): install with multiple artifacts

### LINKED ISSUE #22841: [BUG] Edge install failed (Cannot install msedge on win32)

mxschmitt:
The API response looked like this:

https://edgeupdates.microsoft.com/api/products

image

-> multiple entries for `Windows` and `X64`, but some without artifacts. -> We filter now for `artifacts.length > 0` which seems like a healthy change.

Fixes https://github.com/microsoft/playwright/issues/22841


mxschmitt:
This fixes also https://github.com/microsoft/playwright/actions/runs/6738064186/job/18316866173

yury-s (approved):
> multiple entries for Windows and X64, but some without artifacts.

This sounds like a bug on the server. Can you report that?

PR #27960, opened 2023-11-03
feat: Add support for dispatching device orientation events

### LINKED ISSUE #27887: [Feature] Add support for DeviceOrientationEvent in dispatchEvent
I'm building a web app that uses device orientation events. For testing, it would be handy to be able to dispatch this event type to see that the app reacts as it should.

It didn't look like such a big change so I tried to add the new event type to `dispatchEvent` and `eventType` in `packages/playwright-core/src/server/injected/injectedScript.ts` but when I try it out in my app's tests it gives this error:

```
locator.dispatchEvent: ReferenceError: DeviceOrientationEvent is not defined
    at InjectedScript.dispatchEvent (<anonymous>:5484:9)
    at eval (eval at eval (:4712:24), <anonymous>:2:22)
    at eval (eval at evaluate (:208:30), <anonymous>:18:20)
    at UtilityScript.evaluate (<anonymous>:210:17)
    at UtilityScript.<anonymous> (<anonymous>:1:44)
```

Using this line:

```
await this.page.locator('html').dispatchEvent('deviceorientation', {alpha: 10, beta: 20});
```

It does work fine in the unit tests I wrote, both Chromium and Firefox (in `tests/page/page-dispatchevent.spec.ts`), only not when I try to use the updated code in my own tests.

I'm running this only in Chromium right now, on Ubuntu Linux. I noticed that Firefox and Webkit has some extra code regarding events but I haven't found that for Chromium.

I would be willing to work on this if I could get some pointers on what I'm missing.

mattiasw:
Fixes #27887

### DIFF
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -67,6 +67,10 @@ export type HitTargetInterceptionResult = {
   stop: () => 'done' | { hitTargetDescription: string };
 };
 
+interface WebKitLegacyDeviceOrientationEvent extends DeviceOrientationEvent {
+  readonly initDeviceOrientationEvent: (type: string, bubbles: boolean, cancelable: boolean, alpha: number, beta: number, gamma: number, absolute: boolean) => void;
+}
+
 export class InjectedScript {
   private _engines: Map<string, SelectorEngine>;
   _evaluator: SelectorEvaluatorImpl;
@@ -1036,6 +1040,15 @@ export class InjectedScript {
       case 'focus': event = new FocusEvent(type, eventInit); break;
       case 'drag': event = new DragEvent(type, eventInit); break;
       case 'wheel': event = new WheelEvent(type, eventInit); break;
+      case 'deviceorientation':
+        try {
+          event = new DeviceOrientationEvent(type, eventInit);
+        } catch {
+          const { bubbles, cancelable, alpha, beta, gamma, absolute } = eventInit as {bubbles: boolean, cancelable: boolean, alpha: number, beta: number, gamma: number, absolute: boolean};
+          event = this.document.createEvent('DeviceOrientationEvent') as WebKitLegacyDeviceOrientationEvent;
+          event.initDeviceOrientationEvent(type, bubbles, cancelable, alpha, beta, gamma, absolute);
+        }
+        break;
       default: event = new Event(type, eventInit); break;
     }
     node.dispatchEvent(event);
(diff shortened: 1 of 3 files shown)


mattiasw:

mattiasw:
Looking into tests failing for Webkit...

mattiasw:
It's fixed for Webkit now. Apparently they use an old standard for creating the event. A couple of other checks seem a bit flaky though but they look unrelated to the changes.

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1036,6 +1040,14 @@ export class InjectedScript {
       case 'focus': event = new FocusEvent(type, eventInit); break;
       case 'drag': event = new DragEvent(type, eventInit); break;
       case 'wheel': event = new WheelEvent(type, eventInit); break;
+      case 'deviceorientation':
+        const { bubbles, cancelable, alpha, beta, gamma, absolute } = eventInit as {bubbles: boolean, cancelable: boolean, alpha: number, beta: number, gamma: number, absolute: boolean};
+        event = this.document.createEvent('DeviceOrientationEvent') as WebKitLegacyDeviceOrientationEvent;
+        if (typeof event.initDeviceOrientationEvent === 'function')
```
pavelfeldman, line 1046: style nit: no need to create and re-create:

```js
try {
  event = new DeviceOrientationEvent(type, { alpha, beta });
} catch () {
  event = this.document.createEvent('DeviceOrientationEvent');
  event.initDeviceOrientationEvent(type, ...);
}
```

pavelfeldman:
Thanks! If you are interesting in doing the same for DeviceMotionEvent (these often go hand in hand) in a separate PR, we would be happy to accept it!

mattiasw:
> Thanks! If you are interesting in doing the same for DeviceMotionEvent (these often go hand in hand) in a separate PR, we would be happy to accept it!

Sure!

mxschmitt:
Hey Mattias! This patch is failing on Android:

```
Error: locator.dispatchEvent: TypeError: event.initDeviceOrientationEvent is not a function
    at InjectedScript.dispatchEvent (<anonymous>:5496:17)
    at eval (eval at eval (:4719:24), <anonymous>:2:22)
    at eval (eval at evaluate (:226:30), <anonymous>:18:20)
    at UtilityScript.evaluate (<anonymous>:228:17)
    at UtilityScript.<anonymous> (<anonymous>:1:44)
Call log:
  - waiting for locator('html')
```

Maybe because this isn't available in the Chrome version we use on Android? Would appreciate if you could take a look, thanks!

mattiasw:
It looks like it doesn't have the `DeviceOrientationEvent` constructor (available since Chrome 59 on Android according to MDN) nor the `initDeviceOrientationEvent` function that WebKit uses (which seems reasonable). Is there a way for me to test this?

mxschmitt:
```
 npm run build
 ./utils/avd_install.sh
 ./utils/avd_recreate.sh
 ./utils/avd_start.sh
 npm run atest -- <your-test>
```

that should make it work under macOS.

mxschmitt:
I went ahead since it was a bit involved: https://github.com/microsoft/playwright/pull/28077

mattiasw:
Thanks, I suspected it would have to do with secure context and I did not know how to make that work. :sweat_smile:

Issue #19892, opened 2023-01-05
[BUG] Codegen browser does not record drag and drop events


hanshs:
**Context:**
- Playwright Version: 1.29.1
- Operating System: Windows
- Node.js version: 16.15.1
- Browser: Chromium

**Describe the bug**

Step to reproduce
1. Run ´npx playwright codegen´
2. In the automated browser window, go to https://www.w3schools.com/html/html5_draganddrop.asp 
3. Try to drag the logo in the first box to the second box

The inspector window only registers `click` event on the selector where the mouse was released. No `dragTo` event was registered
Is there any way to achieve this?

![chrome_JGgqMjXdUt](https://user-images.githubusercontent.com/19611757/210782223-3efce5d6-08bd-477e-bffa-efb8516ae1e1.gif)

mxschmitt:
Drag and drop is not yet implemented in the codegen, marking as a feature request.

chenshuichao:
any plan about this feature?thank you!

sushruth:
This feature is crucial for projects that use a drag-able canvas, or custom scrolling apparatus, or file drag and drop UIs - like Outlook. I hope this gets picked up soon.

Anirudhv19:
The drag and drop feature isn't working on Version 1.38.0

AnshPatel15:
are there any plans to implement this yet?

AnshPatel15:
@mxschmitt Could i please get a timeline or something for drag and drop?

JGJP:
@StuartAtmosi you can check [this reply](https://github.com/microsoft/playwright/pull/27985#issuecomment-1889725126)

It seems like it will not get implemented any time soon

AndiBMGitHub:
Would be very nice to have this implemented/fixed, any plans to do it soon?

marcusNumminen:
I would also like to see this

karunqa:
I would also like to see this feature implemented.

fasatrix:
+ 1 would be nice to have this implemented

autonomobil:
+1

3zzy:
+1

alepersichetti:
+1 Here, the feature is essential to test some feature of the product I'm working on.

naolmelesse:
+1 This feature would make codegen more handy

autonomobil:
any updates?

manjotsk:
Is there any active development we can contribute to?

ronaldohoch:
[...]

PR #27985, opened 2023-11-06
feat(codegen): basic drag'n drop support

mxschmitt:
https://github.com/microsoft/playwright/issues/19892

### DIFF
--- a/packages/playwright-core/src/server/injected/recorder.ts
+++ b/packages/playwright-core/src/server/injected/recorder.ts
@@ -39,6 +39,7 @@ interface RecorderTool {
   disable?(): void;
   onClick?(event: MouseEvent): void;
   onDragStart?(event: DragEvent): void;
+  onDrop?(event: DragEvent): void;
   onInput?(event: Event): void;
   onKeyDown?(event: KeyboardEvent): void;
   onKeyUp?(event: KeyboardEvent): void;
@@ -142,6 +143,7 @@ class RecordActionTool implements RecorderTool {
   private _hoveredElement: HTMLElement | null = null;
   private _activeModel: HighlightModel | null = null;
   private _expectProgrammaticKeyUp = false;
+  private _dragSourceElement: HTMLElement | null = null;
 
   constructor(private _recorder: Recorder) {
   }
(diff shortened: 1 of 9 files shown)


packages/playwright-core/src/server/injected/recorder.ts
```diff
@@ -349,6 +372,8 @@ class RecordActionTool implements RecorderTool {
       return true;
     if (nodeName === 'INPUT' && ['date'].includes((target as HTMLInputElement).type))
       return true;
+    if (target.draggable && nodeName !== 'A')
```
mxschmitt, line 375: this check requires further thinking. It made these tests fail:

```ts
[chromium] › library/inspector/cli-codegen-1.spec.ts:653:7 › cli codegen › should await popup ──
[chromium] › library/inspector/cli-codegen-2.spec.ts:200:7 › cli codegen › should download files 
```

images/links with a src/href are draggable by default.

mxschmitt:
Action items:

- test on real world drag'n drop scenarios
- Look into getting rid of the weird if

AnshPatel15:
Could I ask how long it would take for this functionality to be added? I have asked before but I really need this to work on my application.

chensce:
We also really need this to work.

chensce:
@mxschmitt ,we can reference the rrweb(https://www.rrweb.io/) about drag'n drop recording

packages/playwright-core/src/server/injected/recorder.ts
ruifigueira: I gave it a try and cherry-picked your code on top of my PR #28767, and those tests pass. You can take a look at:

https://github.com/ruifigueira/playwright/tree/feat/record-drag-drop

Keep in mind that I changed the way recorder works: it no longer consumes events and performs actions via playwright while recording. This makes the recording experience more natural and, at least until now, I didn't notice any strange behaviour because of those changes (all tests are passing, as you can check in my PR).

ruifigueira: Meanwhile I reverted most of the changes in the PR (it could be solved with a one-liner), but that branch still has them.

mxschmitt:
Closing for now, since this was an experiment and it has proven that it was not that trivial to implement. We might re-visit this feature request in the future and unfortunately can't commit on any eta.

autonomobil:
any updates on this?

PR #28038, opened 2023-11-08
fix: JSHandle preview text for non-ascii attributes/children

mxschmitt:
This surfaced in .NET that the string in the driver got incorrectly cut, then transferred to .NET as an invalid UTF8 character [`\ud835`](https://charbase.com/d835-unicode-invalid-character) which .NET wasn't able to parse and threw an error.

Drive-by: Move similar function from `packages/playwright-core/src/client/page.ts` into isomorphic `stringUtils`.

https://github.com/microsoft/playwright-dotnet/issues/2748

### DIFF
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -1059,9 +1059,7 @@ export class InjectedScript {
         attrs.push(` ${name}="${value}"`);
     }
     attrs.sort((a, b) => a.length - b.length);
-    let attrText = attrs.join('');
-    if (attrText.length > 50)
-      attrText = attrText.substring(0, 49) + '\u2026';
+    const attrText = trimStringWithEllipsis(attrs.join(''), 50);
     if (autoClosingTags.has(element.nodeName))
       return oneLine(`<${element.nodeName.toLowerCase()}${attrText}/>`);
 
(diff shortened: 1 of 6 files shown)


packages/playwright-core/src/utils/isomorphic/stringUtils.ts
```diff
@@ -103,3 +103,10 @@ export function escapeForAttributeSelector(value: string | RegExp, exact: boolea
   // so we escape them differently.
   return `"${value.replace(/\\/g, '\\\\').replace(/["]/g, '\\"')}"${exact ? 's' : 'i'}`;
 }
+
+export function trimStringWithEllipsis(input: string, cap: number): string {
+  const chars = [...input];
```
dgozman, line 108: To be more efficient, let's fast return:

```ts
if (input.length <= cap)
  return input;
```

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1072,10 +1070,8 @@ export class InjectedScript {
       for (let i = 0; i < children.length; i++)
         onlyText = onlyText && children[i].nodeType === Node.TEXT_NODE;
     }
-    let text = onlyText ? (element.textContent || '') : (children.length ? '\u2026' : '');
-    if (text.length > 50)
-      text = text.substring(0, 49) + '\u2026';
```
dgozman, line 1077: I wonder whether we have more `substring()` calls that produce bad UTF16? Can we somehow audit our usage?

mxschmitt: I did a quick audit by grepping for `.substring(0`, thats how I found the one in `page.ts`.

dgozman: I found this one as well: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/injected/selectorGenerator.ts#L279

dgozman (approved):

PR #28039, opened 2023-11-08
docs: move expect.extend docs from "test configuration" to "assertions"

dgozman:


mxschmitt (approved):

docs/src/test-assertions-js.md
```diff
@@ -134,7 +134,7 @@ expect(test.info().errors).toHaveLength(0);
 
 Note that soft assertions only work with Playwright test runner.
 
-## Custom Expect Message
+## Custom expect message
```
yury-s, line 137: Why are these changing? The style was consistent in this document and we use both across the docs.

docs/src/test-assertions-js.md
```diff
@@ -236,3 +236,86 @@ await expect(async () => {
   timeout: 60_000
 });
 ```
+
+## Add custom matchers using expect.extend
```
yury-s, line 240:
```suggestion
## Adding custom matchers using expect.extend
```

yury-s (approved):

PR #28040, opened 2023-11-08
docs(intro-js.md): add bun install

Morgy93:
Added the instructions for using bun to install playwright.

Reference: https://bun.sh/docs/cli/bun-create

### DIFF
--- a/docs/src/intro-js.md
+++ b/docs/src/intro-js.md
@@ -24,7 +24,8 @@ Get started by installing Playwright using npm or yarn. Alternatively you can al
   values={[
     {label: 'npm', value: 'npm'},
     {label: 'yarn', value: 'yarn'},
-    {label: 'pnpm', value: 'pnpm'}
+    {label: 'pnpm', value: 'pnpm'},
+    {label: 'bun', value: 'bun'}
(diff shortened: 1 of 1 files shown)


mxschmitt:
Hey! We appreciate the PR. While Bun tries to be Node.js compatible, they still lack essential features to make Playwright work as per: https://github.com/oven-sh/bun/issues/2492 and https://github.com/oven-sh/bun/issues/4751.

Once they have added these missing features, we're more than happy to re-consider recommending Bun to our users. (This requires also running our tests in Bun).

Jarred-Sumner:
@mxschmitt Bun's package manager is designed to be fully compatible with Node.js, you can use it as a package manager for Node.js similarly to `pnpm` or `yarn`.

Morgy93:
@mxschmitt You are now able to re-consider recommending Bun to your users:

>[Playwright v1.41.0](https://github.com/microsoft/playwright/releases/tag/v1.41.0) which includes the above change has now been tagged and works together with Bun v1.0.22+.
Source: https://github.com/oven-sh/bun/issues/2492#issuecomment-1894624557

PR #28043, opened 2023-11-09
chore: rework assert dialog

pavelfeldman:


packages/playwright-core/src/server/injected/recorder.ts
```diff
@@ -442,217 +442,181 @@ class RecordActionTool implements RecorderTool {
...
+    } else {
+      const { selector } = generateSelector(this._recorder.injectedScript, target, { testIdAttributeName: this._recorder.state.testIdAttributeName, forTextExpect: true });
       return {
         name: 'assertText',
-        selector: this._selectionHighlight.selector,
+        selector,
         signals: [],
-        text: this._selectionText.selectedText,
-        substring: this._selectionText.fullText !== this._selectionText.selectedText,
+        text: target.textContent!,
```
dgozman, line 522: To make sure we match our own definition of element text:

```suggestion
        text: elementText(new Map(), target).full,
```

packages/playwright-core/src/server/injected/recorder.ts
```diff
@@ -442,217 +442,181 @@ class RecordActionTool implements RecorderTool {
...
-    this._inputHighlight = null;
-    this._recorder.injectedScript.window.getSelection()?.empty();
-    this._recorder.updateHighlight(null, false);
+    this._dialogElement = this._recorder.document.createElement('x-pw-dialog');
+    this._keyboardListener = (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
+        if (this._dialogElement)
+          this._closeDialog();
+        else
+          this._recorder.delegate.setMode?.('recording');
```
dgozman, line 560: I think this is never reached.

dgozman (approved):

PR #28045, opened 2023-11-09
docs: add guide for unit testing

### LINKED ISSUE #27754: [Feature] Unit Testing
I have read about experimental component testing, but it would be great if we could use playwright also as a unit testing framework for singular functions, so no jest or vitest ist needed alongside. Is this currently already possible ?

boan-anbo:
This closes #27754

### DIFF
--- a/docs/src/test-unit-testing-js.md
+++ b/docs/src/test-unit-testing-js.md
@@ -1,0 +1,114 @@
+---
+id: test-unit-testing
+title: "Unit Testing"
+---
+
+## Example
+
+You can perform unit testing in Playwright as shown below:
+
+```js title="unit-testing-example.spec.ts"
+import { expect, test } from "@playwright/test";
+
+test("function should return answer", async () => {
+  const add = (a, b) => a + b;
+  expect(add(1, 2)).toBe(3);
+});
+```
+
+For unit testing, you can also use [parameterized tests](./test-parameterize-js.md) and [built-in](./test-fixtures-js.md#built-in-fixtures) or [custom](./test-fixtures-js.md#creating-a-fixture) fixtures, alongside your [E2E](./writing-tests-js.md) and [component tests](./test-components-js.md).
+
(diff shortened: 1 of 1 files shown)


MindaugasMateika:
In my opinion, a lot of unit testing usually involves mocking/stubbing of some sort. Playwright does not have this capability but I think it would be beneficial if not adding an example then just mentioning that mocking/stubbing should be performed with some 3rd  party libraries.

pavelfeldman:
Is this something you are actively working on? (docs / lint did not pass)

boan-anbo:
> Is this something you are actively working on? (docs / lint did not pass)

yes, I was waiting for early feedback. If the lint needs to pass first to be reviewed, just let me know!

boan-anbo:

docs/src/test-unit-testing-js.md
```diff
@@ -0,0 +1,114 @@
...
+});
+
+test(`testing with custom fixture`, async ({ defaultItem }) => {
+  expect(defaultItem).toBe("defaultItemName");
+});
+```
+
+## Required browser configuration
+
+Note that even for unit tests without browsers, Playwright requires at least one browser configuration in your [`config`](./browsers#configure-browsers) to run any tests at all. Without it, no tests will be picked up for execution, regardless of whether they use browser-related APIs or not.
```
yury-s, line 89: This is false. The project section could be omitted altogether, or one can have a project without `use` section.

yury-s (commented):
The idea we want to convey is that Playwright will only create fixtures that are used by the tests. If the tests don't use browser/context/page fixtures, no browsers are launched and none of the fixtures are created. In that case the user also doesn't need to install the browsers. Reading the document it's not clear to me. I'd rather we focused on this aspect and drop the parts where we explain differences between unit and e2e tests.

dgozman:
This one seems stale, closing.

PR #28071, opened 2023-11-09
chore: remove from client check if browser is co-located with server

yury-s:
Reference https://github.com/microsoft/playwright/issues/27792


packages/playwright-core/src/server/dispatchers/frameDispatcher.ts
```diff
@@ -218,19 +216,8 @@ export class FrameDispatcher extends Dispatcher<Frame, channels.FrameChannel, Br
...
-    let { localPaths } = params;
-    if (!localPaths) {
-      if (!params.streams)
-        throw new Error('Neither localPaths nor streams is specified');
-      localPaths = params.streams.map(c => (c as WritableStreamDispatcher).path());
-    }
-    for (const p of localPaths)
-      assert(path.isAbsolute(p) && path.resolve(p) === p, 'Paths provided to localPaths must be absolute and fully resolved.');
-    return await this._frame.setInputFiles(metadata, params.selector, { localPaths }, params);
+    const inputFileItems = await prepareFilesForUpload(this._frame, params);
```
dgozman, line 219: I feel like this belongs to the `server`, not `dispatchers`. WDYT?

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -322,7 +283,7 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
       error += ', please write it to a file and pass its path instead.';
     throw new Error(error);
   }
-  return { files: payloads };
+  return { payloads };
```
dgozman, line 286: Shouldn't you remove the check on the line 282 as well? Perhaps include the removal of this method and protocol field in the same PR?

packages/playwright-core/src/server/dispatchers/frameDispatcher.ts
yury-s: Moved.

packages/playwright-core/src/client/elementHandle.ts
yury-s: Done.

dgozman (approved):

PR #28094, opened 2023-11-12
docs: addInitScript workaround for Worker

### LINKED ISSUE #28029: [Feature] addInitScript for Worker
Hi. We are currently working on a project that needs to monkey patch calls to [WebAssembly.instantiate()](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/instantiate) and [WebAssembly.instantiateStreaming()](https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/instantiateStreaming). 

We have quite successfully done so using [addInitScript](https://playwright.dev/docs/api/class-page#page-add-init-script) api for main page, but I'm quite struggling to monkey patch calls inside worker.

Our main problem is that monkey patching worker is possible after initialization code of worker is evaluated, not before.

For example(example with Math.number for simplicity):

```ts
page.on('console', msg => console.log(msg));
  await page.addInitScript(monkeyPatch);
  // such api does not exist
  // await worker.addInitScript(monkeyPatch);
  
  // while initializing the worker, the monkey patch is not applied.
  await Promise.all([
    page.waitForEvent('worker'),
    page.goto('http://localhost:5173/')]);
  const worker = page.workers()[0];
  
  // this works but comes too late.
  await worker.evaluate(monkeyPatch);
  expect(await page.evaluate(() => self['workerFunction']())).toBe('42');
  expect(await worker.evaluate(() => self['workerFunction']())).toBe('42');
```

running above program, there's a initialization code of worker that utilizes the self['workerFunction']. This does not use monkey patched version that returns 42 as in last line, but uses the original one, returning some random number.

I made a repo that showcases the above problem to a [repo](https://github.com/doehyunbaek/monkeypatch-worker).

If these kinds of functionality is already supported, sorry for inconvenience and close the issue.

If not, I am planning to work on such features. It would be great if you let me know any technical reason why it is not appropriate or any pointers to get the implementation done.

doehyunbaek:
As per https://github.com/microsoft/playwright/issues/28029#issuecomment-1803977813 by @dgozman, I made docs change that documents current capabilities of Worker api. 

I ran the documentation linter, but unfortunately could not get the docusaurus working as it kept serving the old site.

If there is any deficiencies or inadequacies, please tell me.

fixes https://github.com/microsoft/playwright/issues/28029

### DIFF
--- a/docs/src/api/class-worker.md
+++ b/docs/src/api/class-worker.md
@@ -54,7 +54,7 @@ foreach(var pageWorker in page.Workers)
 
 **addInitScript for Worker**
 
-Currently, methods like [`method: Page.addInitScript`] is not supported for worker. See issue [#28029](https://github.com/microsoft/playwright/issues/28029).
+To implement something like [`method: Page.addInitScript`] for workers, use request routing:
 
 However, there is a workaround by routing the main worker script. For code example, see [this comment](https://github.com/microsoft/playwright/issues/28029#issuecomment-1802788239). For full repository demo, see [this repository](https://github.com/doehyunbaek/monkeypatch-worker).
 
--- a/docs/src/api/class-worker.md
+++ b/docs/src/api/class-worker.md
@@ -52,6 +52,12 @@ foreach(var pageWorker in page.Workers)
 }
 ```
 
+**addInitScript for Worker**
+
+Currently, methods like [`method: Page.addInitScript`] is not supported for worker. See issue [#28029](https://github.com/microsoft/playwright/issues/28029).
+
+However, there is a workaround by routing the main worker script. For code example, see [this comment](https://github.com/microsoft/playwright/issues/28029#issuecomment-1802788239). For full repository demo, see [this repository](https://github.com/doehyunbaek/monkeypatch-worker).
+
 ## event: Worker.close
 * since: v1.8
 - argument: <[Worker]>


docs/src/api/class-worker.md
```diff
@@ -52,6 +52,12 @@ foreach(var pageWorker in page.Workers)
 }
 ```
 
+**addInitScript for Worker**
+
+Currently, methods like [`method: Page.addInitScript`] is not supported for worker. See issue [#28029](https://github.com/microsoft/playwright/issues/28029).
```
dgozman, line 57:
```suggestion
To implement something like [`method: Page.addInitScript`] for workers, use request routing:
```

docs/src/api/class-worker.md
```diff
@@ -52,6 +52,12 @@ foreach(var pageWorker in page.Workers)
...
+
+However, there is a workaround by routing the main worker script. For code example, see [this comment](https://github.com/microsoft/playwright/issues/28029#issuecomment-1802788239). For full repository demo, see [this repository](https://github.com/doehyunbaek/monkeypatch-worker).
```
dgozman, line 59: Instead of linking to issues/repos, we prefer to make an inline code block with an example right here.

dgozman (commented):
Thank you for the PR! See a few inline comments.

Once you change the docs, run `npm run build` to regenerate the types and include those changes in the PR. This will make the "lint" bot green and will allow us to merge the PR.

docs/src/api/class-worker.md
doehyunbaek: I guess we should provide code examples for Node.js / Python / Java / .Net? As I have never used other apis other than Node.js, I think it is going to take a little bit more time to get the code blocks for other three languages.

doehyunbaek:
I am sorry but unfortunately I do not think I have time for writing code examples for other languages.

Closing this as I think I can not further this.

PR #28111, opened 2023-11-13
chore: allow asserting substring

pavelfeldman:


packages/playwright-core/src/server/injected/recorder.ts
```diff
@@ -622,36 +626,60 @@ class TextAssertionTool implements RecorderTool {
...
       labelElement.textContent = 'Value:';
       const checkboxElement = this._recorder.document.createElement('input');
       labelElement.appendChild(checkboxElement);
       checkboxElement.type = 'checkbox';
-      checkboxElement.checked = this._action.checked;
+      checkboxElement.checked = action.checked;
       checkboxElement.addEventListener('change', () => {
-        if (this._action?.name === 'assertChecked')
-          this._action.checked = checkboxElement.checked;
+        if (action.name === 'assertChecked')
```
dgozman, line 681: This is always `true` now.

dgozman (approved):

packages/playwright-core/src/server/injected/recorder.ts
pavelfeldman: will follow up

PR #28144, opened 2023-11-14
fix(html-reporter): Include specified host and port in the logged instructions to launch the HTML report

D4N14L:
Previously, options used to specify the host and port would be dropped from the logged-out command which starts the HTML report server. This is an issue for some scenarios (ex. Codespaces) which perform port-forwarding to specific addresses and ports to provide access to the service.

This change appends these options when provided using the `--host` and `--port` args.

Confirmed locally that the `--host` and `--port` args work as expected when passed into the `npx` command


mxschmitt (approved):
LGTM, thanks!

PR #28148, opened 2023-11-15
feat(webkit): roll to r1944

playwrightmachine:


packages/playwright-core/src/server/injected/domUtils.ts
```diff
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+let browserName = '';
```
dgozman, line 17: To emphasize we should not use it often:

```suggestion
let browserNameForWorkarounds = '';
```

dgozman (approved):

PR #28156, opened 2023-11-15
chore: mark outofprocess connection not as remote

mxschmitt:


packages/playwright-core/src/server/dispatchers/dispatcher.ts
```diff
@@ -181,21 +181,21 @@ export class DispatcherConnection {
   readonly _dispatchersByType = new Map<string, Set<string>>();
   onmessage = (message: object) => {};
   private _waitOperations = new Map<string, CallMetadata>();
-  private _isLocal: boolean;
+  private _bufferBinaryFormat: 'buffer' | 'base64';
 
-  constructor(isLocal?: boolean) {
-    this._isLocal = !!isLocal;
+  constructor(bufferBinaryFormat: 'buffer' | 'base64') {
```
yury-s, line 186: Why is anything changing in the server code? It should be pure client-side fix.

packages/playwright-core/src/outofprocess.ts
```diff
@@ -45,8 +45,7 @@ class PlaywrightClient {
     this._driverProcess.unref();
     this._driverProcess.stderr!.on('data', data => process.stderr.write(data));
 
-    const connection = new Connection(undefined, undefined);
-    connection.markAsRemote();
```
yury-s, line 49: Maybe add something like `markAsOutOfProcess` instead of passing encoding format in the constructor in all call sites?

mxschmitt: I can make it optional as well, just thought implicit is better. So its by default b64 and only for inprocess we need to specify 'buffer'?

pavelfeldman:
Closing as stale

yury-s (approved):

packages/playwright-core/src/client/connection.ts
```diff
@@ -90,6 +91,14 @@ export class Connection extends EventEmitter {
     return this._isRemote;
   }
 
+  useRawBuffers() {
+    this._rawBuffers = false;
```
mxschmitt, line 95:
```suggestion
    this._rawBuffers = true;
```

PR #28169, opened 2023-11-15
test: fix 'exposeFunction should not leak' in video mode

mxschmitt:
This is like how we do it with the other channel tests. In video mode we produce artifacts so we need to add them to our expectation.


yury-s (approved):
Can you add a description why it fixes the issue?

PR #28227, opened 2023-11-17
fix(merge): normalize path separators when merging across platforms

yury-s:
Previously it was not pretty:

image


packages/playwright/src/reporters/base.ts
```diff
@@ -398,7 +398,7 @@ export function formatResultFailure(test: TestCase, result: TestResult, initialI
 }
 
 export function relativeFilePath(config: FullConfig, file: string): string {
-  return path.relative(config.rootDir, file) || path.basename(file);
+  return toPosixPath(path.relative(config.rootDir, file)) || path.basename(file);
```
dgozman, line 401: I wonder whether this would regress line/list reporters on Windows?

packages/playwright/src/reporters/blob.ts
```diff
@@ -59,6 +60,7 @@ export class BlobReporter extends TeleReporterEmitter {
       userAgent: getUserAgent(),
       name: process.env.PWTEST_BLOB_REPORT_NAME,
       shard: config.shard ?? undefined,
+      pathSeparator: path.sep,
```
dgozman, line 63: I don't see this being used anywhere.

packages/playwright/src/reporters/merge.ts
```diff
@@ -74,6 +76,66 @@ export async function createMergedReport(config: FullConfigInternal, dir: string
...
 
+class PathSeparatorPatcher {
+  private _from: string;
+  private _to: string;
+  constructor() {
+    this._from = path.sep === '/' ? '\\' : '/';
+    this._to = path.sep;
+  }
+
+  updatePathSeparators(jsonEvent: JsonEvent) {
```
dgozman, line 87: Should we skip the work when `pathSeparator === path.sep`?

dgozman (approved):

packages/playwright/src/reporters/blob.ts
yury-s: Oops. Updated.

packages/playwright/src/reporters/merge.ts
```diff
@@ -377,3 +367,95 @@ class IdsPatcher {
...
+    if (jsonEvent.method === 'onProject') {
+      this._updateProject(jsonEvent.params.project as JsonProject);
+      return;
+    }
+    if (jsonEvent.method === 'onTestEnd') {
+      const testResult = jsonEvent.params.result as JsonTestResultEnd;
+      testResult.errors.forEach(error => this._updateLocation(error.location));
+      return;
+    }
+    if (jsonEvent.method === 'onTestEnd') {
```
dgozman, line 407: This will never execute, because of `return` at the line 405.

packages/playwright/src/reporters/merge.ts
```diff
@@ -186,8 +173,10 @@ async function mergeEvents(dir: string, shardReportFiles: string[], stringPool:
       salt = sha1 + '-' + i;
     saltSet.add(salt);
 
-    const idsPatcher = new IdsPatcher(stringPool, metadata.name, salt);
-    idsPatcher.patchEvents(parsedEvents);
+    const eventPatchers = new JsonEventPatchers();
+    eventPatchers.patchers.push(new IdsPatcher(stringPool, metadata.name, salt));
+    eventPatchers.patchers.push(new PathSeparatorPatcher(metadata.pathSeparator));
```
dgozman, line 178: I think we should only patch when running with `-c`, wdyt?

dgozman (approved):

PR #28229, opened 2023-11-18
feat(vrt): allow providing screenshot style

pavelfeldman:


packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -260,22 +259,19 @@ export class Screenshotter {
...
-
     await Promise.all((options.mask || []).map(async ({ frame, selector }) => {
       const pair = await frame.selectors.resolveFrameForSelector(selector);
       if (pair)
         framesToParsedSelectors.set(pair.frame, pair.info.parsed);
     }));
     progress.throwIfAborted(); // Avoid extra work.
 
-    await Promise.all([...framesToParsedSelectors.keys()].map(async frame => {
+    await Promise.all(this._page.frames().map(async frame => {
```
dgozman, line 274: I wonder whether this would never resolve because one ad iframe is perpetually navigating? Should we use `nonStallingEvaluateInExistingContext` for frames that are not explicitly specified by any `mask` locator?

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1130,10 +1130,45 @@ export class InjectedScript {
       this.hideHighlight();
     this._highlight = new Highlight(this);
     this._highlight.install();
+
     const elements = [];
     for (const selector of selectors)
       elements.push(this.querySelectorAll(selector, this.document.documentElement));
+    const elementsToMask = this.document.querySelectorAll('[data-test-screenshot="mask"]') as NodeListOf<HTMLElement>;
```
dgozman, line 1137: Let's use `this.querySelectorAll()` to work inside shadow dom.

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1130,10 +1130,45 @@ export class InjectedScript {
...
+    }
+  }
+
+  unmaskSelectors() {
+    this.hideHighlight();
+
+    const elementsToRemove = this.document.querySelectorAll('[data-test-screenshot="remove"]') as NodeListOf<HTMLElement>;
+    for (const element of elementsToRemove) {
+      const value = (element as any).__pw_display;
+      if (value)
```
dgozman, line 1159: We can instead do:

```ts
const value = (element as any).__pw_display || { value: '', priority: '' };
```

packages/playwright-core/src/server/screenshotter.ts
pavelfeldman: If that is an ad frame, we should not go into it at all. Masking in-frame locators sounds artificial to me altogether, it seems like letting the user "annotate" elements with the attributes beforehand is a better approach. With that in mind, having them as separate code paths make sense.

packages/playwright-core/src/server/injected/injectedScript.ts
pavelfeldman: Doh, that requires parsed selector...

packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -58,29 +58,75 @@ function inPagePrepareForScreenshots(hideCaret: boolean, disableAnimations: bool
...
-        root.append(styleTag);
-      styleTags.push(styleTag);
+  const hideCaretStyle = hideCaret ? `
+    *:not(#playwright-aaaaaaaaaa.playwright-bbbbbbbbbbb.playwright-cccccccccc.playwright-dddddddddd.playwright-eeeeeeeee) {
+      caret-color: transparent !important;
+    }` : '';
+
+  const styleText = `
+    ${hideCaretStyle}
+    [pw-mask-for-screenshot="true"] {
```
dgozman, line 71: Why separate attribute, and not `[data-test-screenshot="mask"]`?

packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -58,29 +58,75 @@ function inPagePrepareForScreenshots(hideCaret: boolean, disableAnimations: bool
...
+      for (const element of elementsToRemove) {
+        const value = (element as any).__pw_display || { value: '', priority: '' };
+        element.style.setProperty('display', value.value, value.priority);
+      }
+      for (const element of elementsToHide) {
+        const value = (element as any).__pw_visibility || { value: '', priority: '' };
+        element.style.setProperty('visibility', value.value, value.priority);
+      }
+      for (const element of elementsToMask)
+        element.removeAttribute('--pw-mask-for-screenshot');
```
dgozman, line 124:
```suggestion
        element.removeAttribute('pw-mask-for-screenshot');
```

packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -58,29 +58,75 @@ function inPagePrepareForScreenshots(hideCaret: boolean, disableAnimations: bool
...
+      document.documentElement.append(styleTag);
+    else
+      root.append(styleTag);
+
+    const elementsToRemove = root.querySelectorAll('[data-test-screenshot="remove"]') as NodeListOf<HTMLElement>;
+    for (const element of elementsToRemove) {
+      (element as any).__pw_display = { value: element.style.getPropertyValue('display'), priority: element.style.getPropertyPriority('display') };
+      element.style.setProperty('display', 'none', 'important');
+    }
+    const elementsToHide = root.querySelectorAll('[data-test-screenshot="hide"]') as NodeListOf<HTMLElement>;
```
dgozman, line 103: This needs some documentation.

dgozman (approved):
This

packages/playwright-core/src/server/screenshotter.ts
pavelfeldman: Because we only want it to be invisible during the screenshot.

packages/playwright-core/src/server/screenshotter.ts
pavelfeldman: You mean for the user or as a comment?

dgozman (approved):

PR #28235, opened 2023-11-19
[BUG] fix download with fulfill routing

ruifigueira:
Downloading an asset doesn't seem to trigger a `requestWillBeSentEvent`, and therefore it's always sending a `Fetch.continueRequest` which will skip matching fulfill routings.

This patch basically makes `requestWillBeSentEvent` optional and tries to fallback to `requestPausedEvent` when necessary.


dgozman (commented):
@ruifigueira Thank you for the PR!

Unfortunately, I am not sure we should go this route. Request without `requestWillBeSent` lacks a lot of properties that we need, and also makes the system much harder to reason about. The fact that this also does not work in WebKit is not helping the cause.

I'd recommend to start with filing an issue first. We should assess the usecases and priority of the problem before fixing it in non-trivial manner.

ruifigueira:
Done: https://github.com/microsoft/playwright/issues/28253

PR #28239, opened 2023-11-20
fix: collect all errors in removeFolders

itchyny:
This PR fixes https://github.com/microsoft/playwright/pull/27790#pullrequestreview-1738958803. Previously this function returns only the first error when some of the promises fail. But the type annotation suggests that the original intention was to collect all the errors. This commit fixes the error values, and unexpected `TypeError: object is not iterable`.

### DIFF
--- a/packages/playwright-core/src/utils/fileUtils.ts
+++ b/packages/playwright-core/src/utils/fileUtils.ts
@@ -28,8 +28,8 @@ export async function mkdirIfNeeded(filePath: string) {
 
 export async function removeFolders(dirs: string[]): Promise<Error[]> {
   return await Promise.all(dirs.map((dir: string) =>
-    fs.promises.rm(dir, { recursive: true, force: true, maxRetries: 10 })
-  )).catch(e => e);
+    fs.promises.rm(dir, { recursive: true, force: true, maxRetries: 10 }).catch(e => e)
+  ));
 }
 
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):

BPScott:
I also bumped into this `TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))` issue when attempting to update to 1.40.0. Applying this patch gets my tests working again.

Thanks @itchny!

itchyny:
@BPScott Thanks for checking the patch, I just fixed the suspicious code and was unconfident it really fixes the weird error.
@mxschmitt Would you make a new release, please? We can't update Playwright to 1.40.0 now.

mxschmitt:
We cherry-picked it to the 1.40 release branch and will release it as part of 1.40.1. We usually wait a few more days to collect other regressions. In the meantime I recommend canary: https://playwright.dev/docs/canary-releases

itchyny:
Okay, thanks for clarification.

philipearnshaw:
Do you have an expected release date for this bug? Thank you.

mxschmitt:
Most likely 1-2 business days. We recommend @next in the meantime: https://playwright.dev/docs/canary-releases

philipearnshaw:
@next works well but I can't use it for release. Thank you for the quick response, most appreciated. 👍

PR #28248, opened 2023-11-20
fix(matchers): improve types with explicitly readonly arrays

lazarljubenovic:
Shouldn't be a breaking change; the existing non-explicitly-readonly arrays will keep working.

### DIFF
--- a/packages/playwright/src/matchers/matchers.ts
+++ b/packages/playwright/src/matchers/matchers.ts
@@ -157,7 +157,7 @@ export function toBeInViewport(
 export function toContainText(
   this: ExpectMatcherContext,
   locator: LocatorEx,
-  expected: string | RegExp | (string | RegExp)[],
+  expected: string | RegExp | readonly (string | RegExp)[],
   options: { timeout?: number, useInnerText?: boolean, ignoreCase?: boolean } = {},
 ) {
   if (Array.isArray(expected)) {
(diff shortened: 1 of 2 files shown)


lazarljubenovic:

mxschmitt:
What is it fixing? Could you start filing an issue with expected/actual outcome and some code attached? This would help us to understand this change.

lazarljubenovic:
If you have a read-only array, you cannot pass it into the function now, and you should be able to. Functions which receive collections but don't write into them should type these parameters as read-only.

It's a common mistake; see e.g. [this section](https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes) on the Definitely Typed repo.

mxschmitt:
Thanks for the explanation. Minimal repro would be:

```ts
  const readOnlyArray: readonly string[] =  ['hi', 'hello', 'bar'];
  await expect(page.locator('')).toHaveText(readOnlyArray);
```

which does not work. Seems like a more general Playwright issue tho I feel like, so we should go over all the APIs instead and make them ReadOnlyArray. Will talk with the team about it next week.

mxschmitt:
We discussed it in the team and are happy to accept a contribution for it. This means all our generated API types need to use `ReadonlyArray` for it, this would be around [this](https://github.com/microsoft/playwright/blob/main/utils/generate_types/index.js) file. Are you interested there in contributing to it? Otherwise feel free to file an issue and we'll take care of it.

lazarljubenovic:
Yup, will take a look during the week(end). Thanks for the pointers.

PR #28337, opened 2023-11-25
docs: add note about serials and fixture env

MentalGear:
I stepped into this issue and found it rather tricky to debug, hence it would be good to add a note to check the fixture if the serial execution closes prematurely.

### DIFF
--- a/docs/src/test-parallel-js.md
+++ b/docs/src/test-parallel-js.md
@@ -97,6 +97,10 @@ You can annotate inter-dependent tests as serial. If one of the serial tests
 fails, all subsequent tests are skipped. All tests in a group are retried together.
 
 :::note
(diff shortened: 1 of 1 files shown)


docs/src/test-parallel-js.md
```diff
@@ -96,6 +96,10 @@ export default defineConfig({
 You can annotate inter-dependent tests as serial. If one of the serial tests
 fails, all subsequent tests are skipped. All tests in a group are retried together.
 
+:::note
```
pavelfeldman, line 99: Serial more is discouraged and hence is a fraction of the testing use cases. Extension testing in serial mode is even smaller fraction. Extension testing that uses fixtures in serial mode is even smaller population. I don't think we should bother serial mode users with this information.

pavelfeldman: We would be open to a PR that improves the error message though!

pavelfeldman:
Closing as per above.

MentalGear:
Yes, it's an edge case, but somewhat obnixiously hard to debug. Maybe we could add the hint somewhere else? Not sure if an error msg is the right place, since the error msg wont trigger for this case.

PR #28365, opened 2023-11-27
chore: do not add to the internal action logs

### LINKED ISSUE #28319: [REGRESSION]: Version 1.40.0 Produces corrupted traces
**Context:**
- GOOD Playwright Version: [1.37.1]
- BAD Playwright Version: [1.40.0]
- Operating System: [Windows/Mac]
- Node: [18.16.0/20.4.0]

Steps to reproduce:

1. Create new playwright project https://playwright.dev/docs/intro
2. Update to the latest:
`npm install -D @playwright/test@latest`
`npx playwright install`
`npm install`
4. make sure you have these
```
“devDependencies”: {
    “@playwright/test”: “^1.40.0”,
    “@types/node”: “^20.9.4”
  }
```
in  package.json
5. Enable trace: "on"  in `playwright.config.ts`  in the `export default defineConfig({use:})`  section
6.  Run the below test
```
import * as fs from "fs";

test("Test Trace", async ({ browser }) => {
    const auth = "storages/storageState.json";
    if (!fs.existsSync(auth)) {
        fs.writeFileSync(auth, "{}");
    }
    const context = await browser.newContext({ storageState: auth });
    await page.goto("https://demo.playwright.dev/todomvc");
    await page.context().storageState({ path: auth });
});
```
7. Run `npx playwright show-report` and check the attached trace in the report
8. Verify the traces are displayed and functional. Note: if the trace works on the first attempt - try to reload the page

**Expected result:** Traces should be visible and functional.
**Actual result:** Traces are corrupted. Error is displayed: "**Could not load trace from http://localhost:9323/data/6a329ada265b653d117eeb16a2ba7b231cc78086.zip. Make sure a valid Playwright Trace is accessible over this url.**"

image

[corruptedTrace.zip](https://github.com/microsoft/playwright/files/13453050/trace.zip)

Other details:
When comment out this line:
```
await page.context().storageState({ path: auth });
```
the traces is generated as expected.

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/28319

### DIFF
--- a/tests/playwright-test/ui-mode-trace.spec.ts
+++ b/tests/playwright-test/ui-mode-trace.spec.ts
@@ -19,7 +19,7 @@ import { test, expect, retries } from './ui-mode-fixtures';
 
 test.describe.configure({ mode: 'parallel', retries });
 
-test('should merge trace events', async ({ runUITest, server }) => {
+test('should merge trace events', async ({ runUITest }) => {
   const { page } = await runUITest({
     'a.test.ts': `
       import { test, expect } from '@playwright/test';
@@ -99,7 +99,7 @@ test('should merge screenshot assertions', async ({  runUITest }, testInfo) => {
   ]);
 });
 
-test('should locate sync assertions in source', async ({ runUITest, server }) => {
+test('should locate sync assertions in source', async ({ runUITest }) => {
   const { page } = await runUITest({
     'a.test.ts': `
       import { test, expect } from '@playwright/test';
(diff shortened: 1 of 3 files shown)


tests/playwright-test/ui-mode-trace.spec.ts
```diff
@@ -197,3 +197,32 @@ test('should show screenshot', async ({ runUITest, server }) => {
   await expect(page.getByText('Screenshots', { exact: true })).toBeVisible();
   await expect(page.locator('.attachment-item img')).toHaveCount(1);
 });
+
+test('should not fail on internal page logs', async ({ runUITest, server }) => {
```
dgozman, line 201: Ideally, this would be a library test.

dgozman (approved):

tests/playwright-test/ui-mode-trace.spec.ts
pavelfeldman: It is tricky to make it a library test because the code path only triggers via @playwright/test.

PR #28382, opened 2023-11-28
feat(ct): react component as props

### LINKED ISSUE #28367: [BUG] React: error 31 when passing array of ReactNode as prop to imported component
### System info
- Playwright Version: v1.40.0
- Operating System: Ubuntu 22
- Browser: All
- Other info: playwright/experimental-ct-react 1.40.0, React 18.2.0

### Source code

**Config file**

```ts
import { defineConfig, devices } from "@playwright/experimental-ct-react";
import react from "@vitejs/plugin-react";

export default defineConfig({
  testDir: "./src/app/_components",
  snapshotDir: "./__snapshots__",
  timeout: 10 * 1000,
  fullyParallel: true,
  forbidOnly: !!process.env.CI,
  workers: process.env.CI ? 1 : undefined,
  reporter: "html",
  use: {
    trace: "on-first-retry",
    ctPort: 3100,
    ctViteConfig: {
      plugins: [react()],
    },
  },
  projects: [
    {
      name: "chromium",
      use: { ...devices["Desktop Chrome"] },
    },
    {
      name: "firefox",
      use: { ...devices["Desktop Firefox"] },
    },
    {
      name: "webkit",
      use: { ...devices["Desktop Safari"] },
    },
  ],
});
```

**`things.tsx`**
```tsx
export function Things({ things }: { things: React.ReactNode[]; }) {
  return (
    <div>
      {things.map((thing, i) => <div key={i}>{thing}</div>)}
    </div>
  );
}
```

**`things.spec.tsx`**
```tsx
import { test, expect } from "@playwright/experimental-ct-react";
import { Things } from "./things";

test("currently failing", async ({ mount }) => {
  const jsx = (
    <Things things={[<p key="a">Content 1</p>, <p key="b">Content 2</p>]} />
  );
  console.log(JSON.stringify(jsx, null, 4));
  const component = await mount(jsx);
  await expect(component).toContainText("Content 1");
});
```

**Steps**
- Run the test

**Expected**
It should render two paragraphs, and the test should pass.

**Actual**
It times out looking for the expected text.
It has rendered nothing.

On its console it has given:

```
{
    "kind": "jsx",
    "type": "_long_string_with_path_and_component_name",
    "props": {
        "things": [
            {
                "kind": "jsx",
                "type": "p",
                "props": {
                    "key": "a"
                },
                "children": [
                    "Content 1"
                ]
            },
            {
                "kind": "jsx",
                "type": "p",
                "props": {
                    "key": "b"
                },
                "children": [
                    "Content 2"
                ]
            }
        ]
    },
    "children": []
}
Error
Error
Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7Bkind%2C%20type%2C%20props%2C%20children%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7Bkind%2C%20type%2C%20props%2C%20children%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at th (http://localhost:3100/assets/index-d4fbf8eb.js:27772:60)
    at J (http://localhost:3100/assets/index-d4fbf8eb.js:27782:377)
    at Yi (http://localhost:3100/assets/index-d4fbf8eb.js:27812:308)
    at Wk (http://localhost:3100/assets/index-d4fbf8eb.js:27914:9)
    at Vk (http://localhost:3100/assets/index-d4fbf8eb.js:27900:395)
    at Uk (http://localhost:3100/assets/index-d4fbf8eb.js:27900:324)
    at Jk (http://localhost:3100/assets/index-d4fbf8eb.js:27900:181)
    at Ok (http://localhost:3100/assets/index-d4fbf8eb.js:27891:88)
    at Hk (http://localhost:3100/assets/index-d4fbf8eb.js:27888:432)
    at J (http://localhost:3100/assets/index-d4fbf8eb.js:27596:205)
    at R (http://localhost:3100/assets/index-d4fbf8eb.js:27597:129)
    at EventHandlerNonNull* (http://localhost:3100/assets/index-d4fbf8eb.js:27597:310)
    at  (http://localhost:3100/assets/index-d4fbf8eb.js:27603:3)
```

This same `Things` component works just fine when called in the same way in Storybook and in my application.

I tried to reproduce the error in a single Playwright test file but couldn't get the same error to happen that I can as above where I import the component from another file.

The only other ticket I could find which seemed like it could be related is #27587, but that seems like a different circumstance.

sand4rt:
closes: https://github.com/microsoft/playwright/issues/28367#issuecomment-1830298864


packages/playwright-ct-react/registerSource.mjs
```diff
@@ -44,7 +44,7 @@ export function pwRegister(components) {
  * @returns {component is JsxComponent}
  */
 function isComponent(component) {
-  return !(typeof component !== 'object' || Array.isArray(component));
+  return component.kind === 'jsx';
```
pavelfeldman, line 47: This would result in false positives where a user object has a kind property with the jsx value. I was suggesting a property with a magic name in the issue comment. Btw, what do we do to the non-jsx notation in react? I think we still support it.

sand4rt: > I was suggesting a property with a magic name in the issue comment

I've added the magic property. I still have to test everything thoroughly. There are still some bugs in there.

> Btw, what do we do to the non-jsx notation in react? I think we still support it

I also think this should be supported. I'll create a separate ticket for it.

pavelfeldman (approved):

PR #28384, opened 2023-11-28
devops(docker): add base image ARG to Jammy Dockerfile

### LINKED ISSUE #28383: Add base image arg to Jammy Dockerfile
> Let us know what functionality you'd like to see in Playwright and what your use case is.

I'd like to add a `base_image` ARG to the Jammy Dockerfile. Our use case is that we need all containers running in our CI environment, which includes `playwright:jammy`, to used a hardened base image that we maintain. 

Currently, in order to build the Playwright Jammy Docker image on our custom hardened base image, we would need to maintain our own copy of the Dockerfile just to add a `base_image` ARG. Obviously, this is is not ideal, since over time our copy of the Dockerfile would drift from the upstream and miss out on important things like security fixes or other improvements.

> Do you think others might benefit from this as well?

For anyone else with a need to run Docker images using a hardened base image, yes, I could imagine this change being useful.

For those who don't need this change, the default of building on `ubuntu:jammy` would remain.

markdboyd:
This change adds a base_image ARG to the Dockerfile for Jammy to allow building the Docker image on custom base images

Fixes #28383

### DIFF
--- a/utils/docker/Dockerfile.jammy
+++ b/utils/docker/Dockerfile.jammy
@@ -1,4 +1,6 @@
-FROM ubuntu:jammy
+ARG base_image=ubuntu:jammy
+
+FROM ${base_image}
 
 ARG DEBIAN_FRONTEND=noninteractive
 ARG TZ=America/Los_Angeles


pavelfeldman:
Closing as per the issue conversation.

PR #28401, opened 2023-11-29
fix: replacing oauth tokens from headers

DudaGod:
## What's done

Implement replacing oauth token from stdout in order to avoid compromising them in CI tasks in case when used `process.env.DEBUG = "pw:browser,pw:api"`.


pavelfeldman:
I thought you were going to implement selenium support as a layer instead. We are not accepting patches into the selenium grid support atm.

PR #28405, opened 2023-11-29
devops: point GitHub action to default test results directory

pablopaul:
Point to default test results directory


pavelfeldman:
I believe what we are doing right now is correct.

Issue #28314, opened 2023-11-23
[BUG] setInputFiles changes UI trace


david-olink:
### System info
- Playwright Version: v1.40
- Operating System: Windows 11
- Browser: Chromium (UI-mode) maybe more?
- Other info:

### Source code

index.htm:
```html
<html>
    <body>
        <h1>setInputFiles changes UI trace</h1>

        <select id="aSelect">
            <option value="A">A option</option>
            <option value="B">B option</option>
            <option value="C">C option</option>
        </select>

        <input type="file" id="fileInput">
    </body>
</html>
```

test.spec.ts
```typescript
import { test, expect } from '@playwright/test';

test('setInputFiles changes UI trace', async ({ page }) => {
  await page.goto(' http://127.0.0.1:8080/demo/index.htm');

  {
    // Value is A
    const value = await page.locator('#aSelect').inputValue();
    console.log('Select value', value);
  }

  await page.locator('#aSelect').selectOption({ label: 'C option' });

  {
    // Value is C
    const value = await page.locator('#aSelect').inputValue();
    console.log('Select value', value);
  }

  const file = './tests/example.spec.ts';
  await test.step(`setInputFiles: ${file}`, async () => {
    // Problem 1: This action is not shown in the UI Actions log.
    await page.locator('#fileInput').setInputFiles(file);
    // Problem 2: file input still says "no file chosen" (minor problem)
  });

  {
    // Problem 3: (the biggest problem)
    // Value is C in console, but in the trace image/DOM in the UI mode 
    // the value has jumped back to A after setInputFiles has been called.
    const value = await page.locator('#aSelect').inputValue();
    console.log('Select value', value);

    // Note: setInputFiles has been called on #fileInput and not #aSelect
    // that has been affected in the UI trace.

    // This is really confusing when trying to figure out why the tests fail.

    // The same problem happens when using the fileChooser 
    // instead of directly setting files with setInputFiles
  }

  await expect(page.getByRole('heading')).toHaveText('setInputFiles changes UI trace');
});

```

- [x] I provided exact source code that allows reproducing the issue locally.

**Steps**
- Run the test.spec.ts against the index.htm file simple static server. (Adjust the page.goto in the test file)
- Run `npx playwright test --ui`
- Run the test.
- See details below and in the code comments above.

**Expected**

setInputFiles not to mess around with the trace DOM in the UI mode.

**Actual**

setInputFiles changes a select nearby. See the comments in the code for more details.

[...]

PR #28443, opened 2023-12-01
chore: use more explicit awaits when returning in client

mxschmitt:
This fixes "Problem 1" of https://github.com/microsoft/playwright/issues/28314.


dgozman (commented):
I am not sure this is the right rule for us. We want to make sure all async function calls are awaited, not that we have at least one await in a function.

package.json
```diff
@@ -85,7 +86,8 @@
     "electron-to-chromium": "^1.4.212",
     "enquirer": "^2.3.6",
     "esbuild": "^0.18.11",
-    "eslint": "^8.37.0",
+    "eslint": "^8.55.0",
+    "eslint-plugin-internal-playwright": "file:utils/eslint-plugin-internal-playwright",
```
mxschmitt, line 90: Note: According to my research there seems to be no other way of consuming an ESLint rule without treating it as a local package.

packages/playwright-core/src/client/network.ts
```diff
@@ -159,16 +159,16 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
     return this._provisionalHeaders.headers();
   }
 
-  _actualHeaders(): Promise<RawHeaders> {
+  async _actualHeaders(): Promise<RawHeaders> {
     if (this._fallbackOverrides.headers)
-      return Promise.resolve(RawHeaders._fromHeadersObjectLossy(this._fallbackOverrides.headers));
+      return await Promise.resolve(RawHeaders._fromHeadersObjectLossy(this._fallbackOverrides.headers));
```
dgozman, line 164:
```suggestion
      return RawHeaders._fromHeadersObjectLossy(this._fallbackOverrides.headers);
```

packages/playwright-core/src/client/network.ts
```diff
@@ -300,16 +300,16 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
...
+  private async _raceWithTargetClose(promise: Promise<any>): Promise<void> {
     // When page closes or crashes, we catch any potential rejects from this Route.
     // Note that page could be missing when routing popup's initial request that
     // does not have a Page initialized just yet.
-    return this.request()._targetClosedScope().safeRace(promise);
+    return await this.request()._targetClosedScope().safeRace(promise);
   }
 
-  _startHandling(): Promise<boolean> {
+  async _startHandling(): Promise<boolean> {
```
dgozman, line 310: This one seems weird. I see we only use it in a single place, let's inline?

utils/eslint-plugin-internal-playwright/rules/await-promise-in-class-returns.js
```diff
@@ -0,0 +1,54 @@
...
+      'ClassDeclaration MethodDefinition ReturnStatement': function (statement) {
+        if (statement.type === 'ReturnStatement' && statement.argument) {
+          if (tsutils.isThenableType(
+            parserServices.program.getTypeChecker(),
+            statement.argument,
+            parserServices.getTypeAtLocation(statement.argument)
+          )) {
+            context.report({
+              node: statement,
+              message: 'Return statement in a class should await a promise',
```
dgozman, line 42: Let's explain why to avoid confusion.

dgozman (approved):

packages/playwright-core/src/client/network.ts
mxschmitt: I think its a layer violation otherwise. So far the promise wasn't exposed.

PR #28444, opened 2023-12-01
fix(snapshot): broken snapshot after use of setInputFiles

mxschmitt:
This fixes _"Problem_ 3" of https://github.com/microsoft/playwright/issues/28314.

What happened was the following: 
image

Which ended up that snapshots after the setInputFiles action did not restore other values since it threw.

### DIFF
--- a/packages/trace-viewer/src/snapshotRenderer.ts
+++ b/packages/trace-viewer/src/snapshotRenderer.ts
@@ -223,7 +223,9 @@ function snapshotScript(...targetIds: (string | undefined)[]) {
         scrollLefts.push(e);
 
       for (const element of root.querySelectorAll(`[__playwright_value_]`)) {
-        (element as HTMLInputElement | HTMLTextAreaElement).value = element.getAttribute('__playwright_value_')!;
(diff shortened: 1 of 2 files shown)


packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -223,7 +223,16 @@ function snapshotScript(...targetIds: (string | undefined)[]) {
         scrollLefts.push(e);
 
       for (const element of root.querySelectorAll(`[__playwright_value_]`)) {
-        (element as HTMLInputElement | HTMLTextAreaElement).value = element.getAttribute('__playwright_value_')!;
+        const value = element.getAttribute('__playwright_value_')!;
+        const inputElement = element as HTMLInputElement;
+        if (inputElement.type === 'file') {
+          const file = new File([new Blob([''], { type: 'text/plain' })], value.split(/(\/|\\)/).pop()!, { type: 'text/plain' });
```
dgozman, line 229: Let's just ignore it. I don't think anyone needs a file there. It also probably does not work in WebKit.

dgozman (commented):
Let's add a test!

dgozman (approved):

Issue #28221, opened 2023-11-17
[BUG] Playright take extra 1px in screenshot in CI


balkarov:
### System info
- Playwright Version: v1.40.0
- Operating System: Windows 11
- Browser: Chromium
- Other info:

### Source code

- [ ] I provided exact source code that allows reproducing the issue locally.

**Config file**

```js
import type { PlaywrightTestConfig } from '@playwright/test';
import { devices } from '@playwright/test';
import * as dotenv from 'dotenv';

dotenv.config();

const config: PlaywrightTestConfig = {
    timeout: 450 * 1000,
    expect: {
        timeout: 5000,
        toHaveScreenshot: {
            threshold: 0.3,
            maxDiffPixels: 10,
            scale: 'device',
        },
    },
    globalSetup: require.resolve('./global/globalSetup'),
    fullyParallel: true,
    forbidOnly: !!process.env.CI,
    retries: process.env.CI ? 1 : 0,
    reporter: process.env.CI ? 'dot' : 'list',
    use: {
        baseURL: process.env.PLAYWRIGHT_TEST_BASE_URL,
        actionTimeout: 10000,
        trace: 'retain-on-failure',
        ignoreHTTPSErrors: true,
        storageState: 'global/storageState.json',
    },
    projects: [
        {
            name: 'chromium',
            use: {
                ...devices['Desktop Chrome'],
                viewport: {
                    width: 1920,
                    height: 1080,
                },
                launchOptions: {
                    ignoreDefaultArgs: ['--hide-scrollbars'],
                },
            },
        },
    ],
};
export default config;

```

**Test file (self-contained)**

```js
await test.step('Click background accordion', async () => {
    await page.locator('#visualization-tab-view_settings').click();
    await page.locator('#background-accordion-tab').click();
    await page.locator('#background-enabled__checkbox-accordion').click();

    await expect(page.locator('#background-enabled__checkbox-accordion')).toHaveScreenshot('result.png');
});
```

**Steps**
- Run the test

**Expected**

![expected](https://github.com/microsoft/playwright/assets/6030335/da10ce1b-68cb-43b0-9fa6-ee93b5530e76)

**Actual**

![actual](https://github.com/microsoft/playwright/assets/6030335/81686822-5bee-407c-9ebf-9628d9a5c6b0)

** Diff **
![diff](https://github.com/microsoft/playwright/assets/6030335/d8c81897-b04c-499d-97ec-8a9c502cf848)

[...]

PR #28453, opened 2023-12-01
add tolerances for the image size and floating point errors, when comparing images

paulsmithkc:
Proposed fix for https://github.com/microsoft/playwright/issues/18827 and https://github.com/microsoft/playwright/issues/28221

Due to floating point error in both CSS and JS, it isn't practical/possible to ensure exact pixel sizing in the vast ecosystem of tools and frameworks. This changeset allows for 1px tolerances in the size of the image.

Changes requested:

1. Allow images to differ in size by up to 1.5 pixels in size, to account for floating point rounding errors
2. Allow an additional 0.5 pixels to differ, for the sake of avoiding rounding errors
3. Use the max size of the two images when computing the difference ratio, instead of the size of the expected image

### DIFF
--- a/packages/playwright-core/src/utils/comparators.ts
+++ b/packages/playwright-core/src/utils/comparators.ts
@@ -59,10 +59,17 @@ function compareImages(mimeType: string, actualBuffer: Buffer | string, expected
   const size = { width: Math.max(expected.width, actual.width), height: Math.max(expected.height, actual.height) };
   let sizesMismatchError = '';
   if (expected.width !== actual.width || expected.height !== actual.height) {
-    sizesMismatchError = `Expected an image ${expected.width}px by ${expected.height}px, received ${actual.width}px by ${actual.height}px. `;
+    // only produce an error if the images differ by more than 1 pixel in size
+    // a threshold of 1.5 allows for floating point rounding errors
+    if (Math.abs(expected.width - actual.width) >= 1.5 || Math.abs(expected.height - actual.height) >= 1.5)
+      sizesMismatchError = `Expected an image ${expected.width}px by ${expected.height}px, received ${actual.width}px by ${actual.height}px. `;
+
+    // always resize both images, so that we can perform the pixel diff
     actual = resizeImage(actual, size);
     expected = resizeImage(expected, size);
   }
+
+  // count the number of pixels that are different
   const diff = new PNG({ width: size.width, height: size.height });
   let count;
   if (options._comparator === 'ssim-cie94') {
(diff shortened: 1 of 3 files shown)


paulsmithkc:
@dgozman @yury-s Can we get some eyes on this PR?

dgozman:
@paulsmithkc Sorry for the slow response, most of the team were enjoying the winter holidays 🎄  I have a few observations and questions about this PR.

- First of all, we believe that silently ignoring any pixel differences is a no-go. That's why screenshots that differ in size by a pixel are considered different. That said, we agree that it is a very strict limitation today, so it would be reasonable to count missing/extra pixels towards the allowance like `maxDiffPixels`. Looking at the PR, it seems to not do this, but just silently ignores the size difference.

- Taking #28221 as a reference, because it has a nice screenshot explaining the issue, it does not seem that an easy solution would help there anyway. Since the expected and actual screenshots have one extra pixel at the left, comparing these two images without shifting one of them by a pixel will produce a huge diff (something like 50 percent?), so it would not fit the `maxDiffPixels` or any other reasonable allowance. Do you have some specific examples where ignoring the size difference consistently helps with mismatched screenshots?

- As you explained, this PR starts to use the max size of the two images when computing the difference ratio. I am afraid this could be unexpected, depending on how we compare images of different sizes. For example, a same-size actual may fail a comparison because `maxDiffPixelRatio` results in a small allowance, while a much-larger actual may pass because `maxDiffPixelRatio` turns out to be large.

paulsmithkc:
@dgozman 

-  Sub-pixel size differences are not actually different and come down to minor rounding errors in the browser, css calculations, bounds calculations, and screenshooting code. Expecting there to be no sub-pixel errors in this whole code path is an impossible hurdle.

- If you look at and test the code you will find that `maxDiffPixels` errors are being detected an caught. In fact if you look at the tests you will see that the number of matching pixels has dropped, and the number of erroneous pixels has increased.

- Please notice that the number of pixels that deviate grows linear with image size, but the total number of pixels grows quadratically.
- It is critical to consider that in realworld images this 1px border/boundary will generally account for a tiny percentage of the overall images. Small test images are not a good test case.
- Determining which direction (left, right, up, down) the image is shifted is both computationally impractical, and unnecessary as the % deviation will be such a small number as to not matter.

- Computing the percentile difference from the max leads to sensible limits when a percentile difference is requested. Using the original image size leads to the percentile behaving unexpectedly and failing when the actual % difference is lower then the intended threshold.
- Notice that the 1px bound on size limits how different these can actually be, if we allowed for a larger size difference than that would be a valid concern, but I don't think it applies here.

JBill85:
I can probably provide more examples of this occurring.  Potentially, I can package the proposed change locally and provide an example of the error case before and after the patch.

pavelfeldman:
I'll archive this as I don't think we are comfortable landing it as is. @dgozman provided a great explanation on why. For example, PR suggests that it fixes https://github.com/microsoft/playwright/issues/28221, but my understanding is that since the extra pixel is on the left, this change is not addressing the problem. We acknowledge the problem and are open to the proposals, it is just that we don't consider this one to be appropriate for the reasons above.

PR #28463, opened 2023-12-02
docs: fix the path in snapshotPathTemplate

occo-lbl:
This PR is to correct a path description error in the `snapshotPathTemplate`.
https://playwright.dev/docs/api/class-testproject#test-project-snapshot-path-template

image

Also fixed `TestConfig` and `TestProject` Doc comments.


occo-lbl:

mxschmitt:
This is intended as per https://github.com/microsoft/playwright/pull/26981#discussion_r1322138029.

PR #28465, opened 2023-12-02
feat(api): introduce FrameLocator.waitForFunction()

### LINKED ISSUE #26829: Add new FrameLocator.waitForFunction() method
Today we have contentFrame.waitForFunction() but this does not work with Frame Locators.  What we would LIKE to do is this:
```
await locator.frameLocator('#my-frame').waitForFunction(() => { ... });
```
but what we have to do today is this:
```
const myFrame = await this.page.waitForSelector('#my-frame');
const myContentFrame = await myFrame.contentFrame();
await myContentFrame.waitForFunction(() => { ... });
```
Not the end of the world, but it's three async calls, and it takes away from a purely Locator-based approach that Playwright recommends.

Even something like this would get us halfway there:
```
const myContentFrame = await locator.frameLocator('#my-frame').contentFrame();
await myContentFrame.waitForFunction(() => { ... });
```
but there is no way to get the content frame from a frameLocator.

References:
- https://playwright.dev/docs/api/class-framelocator
- https://playwright.dev/docs/api/class-frame#frame-wait-for-function

joone:
Fixes #26829

### DIFF
--- a/docs/src/api/class-framelocator.md
+++ b/docs/src/api/class-framelocator.md
@@ -217,3 +217,144 @@ Returns locator to the n-th matching frame. It's zero based, `nth(0)` selects th
 ### param: FrameLocator.nth.index
 * since: v1.17
 - `index` <[int]>
+
+## async method: FrameLocator.waitForFunction
+* since: v1.41
+- returns: <[JSHandle]>
+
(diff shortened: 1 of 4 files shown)


joone:

joone:
@dgozman Could you explain why this pull request was closed without any stated reason?

PR #28472, opened 2023-12-04
docs: fix various syntax errors in .NET documentation

thatstatsguy:
As found [here](https://github.com/microsoft/playwright-dotnet/issues/2774) - current code is not syntactically correct

### DIFF
--- a/docs/src/trace-viewer.md
+++ b/docs/src/trace-viewer.md
@@ -241,9 +241,9 @@ await using var context = await browser.NewContextAsync();
 // Start tracing before creating / navigating a page.
 await context.Tracing.StartAsync(new()
 {
-    Screenshots = true,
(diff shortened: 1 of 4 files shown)


docs/src/trace-viewer.md
```diff
@@ -234,7 +234,8 @@ context.tracing().stop(new Tracing.StopOptions()
 ```
 
 ```csharp
-await using var browser = playwright.Chromium.LaunchAsync();
```
mxschmitt, line 237: we have this kind of bug in a few more places, do you mind fixing them too?

image

thatstatsguy: done :)

thatstatsguy:

mxschmitt (approved):

PR #28414, opened 2023-11-30
feat: wait for active route handlers on page/context close

yury-s:
Reference https://github.com/microsoft/playwright/issues/23781

Issue #23781, opened 2023-06-18
[BUG] 'Request context disposed' was reported when use route.fetch()


erma0:
**Context:**
- Playwright Version: 1.35.0
- Operating System: Windows 11
- Python version: 3.11.2
- Browser: Chromium-msedge
- Extra:  

**Code Snippet**

```python
from playwright.sync_api import Route, sync_playwright


def handle(route: Route):
    res = route.fetch()
    route.fulfill(response=res)


with sync_playwright() as playwright:
    browser = playwright.chromium.launch(channel='msedge', headless=False)
    context = browser.new_context()
    page = context.new_page()

    page.route("**/*", handle)
    page.goto("https://www.baidu.com/", wait_until='domcontentloaded')
    page.unroute("**/*") # not work

    page.close()
    context.close()
    browser.close()

```

**Describe the bug**
 'Request context disposed' or 'Response context disposed' 

I think it's because the page has been closed, but the route handle function is still running.

It appears to be related to microsoft/playwright-python#1402 microsoft/playwright-python#1433 microsoft/playwright-python#1458 and has been fixed, but this issue still persists.

How can I stop a running route handle function before the page is closed?

thanks.

dgozman:
Investigation notes:

- Playwright should not silently catch errors in user code to avoid masking problems.
- `fetch()` must throw when request cannot be made, because it returns a value.

Possible improvements, that could still lead to minor regression in existing test suites:
- `unroute()` to await all ongoing route handlers that it removes.
- `page.close()`/`context.close()` to automatically call `unrouteAll()` that would await ongoing handlers.
Both improvements do not help with pages that were closed by the browser, for example with `popupWindow.close()`.

daelmaak:
This is still a problem, and I didn't find a solution to this except for not doing `route.fetch()` at all. Any update on this?

nbaldzhiev:
I am getting `Error: apiRequestContext.fetch: Request context disposed.` errors, but in Webkit (playwright `1.38.0`).
A related discussion I found is https://github.com/microsoft/playwright/discussions/12810.
The suggestion there is to add a `page.waitForResponse` call, but it doesn't always help in our case as we still see this failure from time to time.

Our code (redacted):
```
  await page.route(url, async route => {
    const response = await page.request.fetch(route.request())
    const body = await response.json()
    do-some-modifications-to-body
    route.fulfill({
      response,
      body: JSON.stringify(body)
    })
  })

[...]

PR #28484, opened 2023-12-04
fix(webkit): delay request event until requestIntercepted is received

yury-s:
Previously we were wrongly firing `route` event for the request which are not in fact intercepted (e.g. requests from service worker).

Related https://github.com/microsoft/playwright/pull/28414

### DIFF
--- a/packages/playwright-core/src/server/webkit/wkPage.ts
+++ b/packages/playwright-core/src/server/webkit/wkPage.ts
@@ -1029,13 +1029,13 @@ export class WKPage implements PageDelegate {
     // TODO(einbinder) this will fail if we are an XHR document request
     const isNavigationRequest = event.type === 'Document';
     const documentId = isNavigationRequest ? event.loaderId : undefined;
-    let route = null;
+    const request = new WKInterceptableRequest(session, frame, event, redirectedFrom, documentId);
     // We do not support intercepting redirects.
     if (this._page.needsRequestInterception() && !redirectedFrom)
-      route = new WKRouteImpl(session, event.requestId);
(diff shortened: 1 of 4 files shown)


tests/page/page-event-request.spec.ts
```diff
@@ -98,6 +98,38 @@ it('should report requests and responses handled by service worker with routing'
   expect(await failedRequest.response()).toBe(null);
 });
 
+it('should not intercept requests handled by service worker with routing', async ({ page, server, isAndroid, isElectron, mode, platform, browserName }) => {
```
dgozman, line 101: Maybe we should merge this test into the previous one? It seems like just a few more expect calls are needed.

packages/playwright-core/src/server/webkit/wkPage.ts
```diff
@@ -1029,13 +1029,13 @@ export class WKPage implements PageDelegate {
     // TODO(einbinder) this will fail if we are an XHR document request
     const isNavigationRequest = event.type === 'Document';
     const documentId = isNavigationRequest ? event.loaderId : undefined;
-    let route = null;
+    const request = new WKInterceptableRequest(session, frame, event, redirectedFrom, documentId);
     // We do not support intercepting redirects.
     if (this._page.needsRequestInterception() && !redirectedFrom)
-      route = new WKRouteImpl(session, event.requestId);
-    const request = new WKInterceptableRequest(session, route, frame, event, redirectedFrom, documentId);
+      request._willBeIntercepted = true;
```
dgozman, line 1035: Let's have a map from request id to `Protocol.Network.requestWillBeSentPayload` and only create `WKInterceptableRequest` later, once the interception event arrives.

dgozman (approved):

PR #28487, opened 2023-12-05
chore: do not wait for route on close if route.continue() threw an error

yury-s:

### DIFF
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -319,25 +319,15 @@ export class Route extends ChannelOwner<channels.RouteChannel> implements api.Ro
   }
 
   async abort(errorCode?: string) {
-    this._checkNotHandled();
-    try {
+    await this._handleRoute(async () => {
       await this._raceWithTargetClose(this._channel.abort({ requestUrl: this.request()._initializer.url, errorCode }));
-      this._reportHandled(true);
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/client/network.ts
```diff
@@ -694,7 +711,7 @@ export class RouteHandler {
     if (this.noWaitOnUnrouteOrClose || noWait)
       handlerActivations.forEach(h => h.ignoreException = true);
     else
-      await Promise.all(handlerActivations.map(h => h.complete));
+      await Promise.all(handlerActivations.filter(h => !h.route._didTryToHandle).map(h => h.complete));
```
dgozman, line 714: - Should we set `ignoreException` flag for these? I'd prefer all invocations to be in one of the two possible states.
- What if my route does something after the `fulfill()`? I think we should only mark those that did try to handle and failed, not all of them.

yury-s: > * Should we set `ignoreException` flag for these? I'd prefer all invocations to be in one of the two possible states.

Route doesn't know anything about `handlerInvocation` structure. We can pass it an exception handler that will flip the flag and it's a property of the route rather than a particular call of a handler, so I don't quite see how we could reuse ignoreException flag for it.

dgozman (approved):

PR #28488, opened 2023-12-05
feat(pwt): add reporting of Error.cause

### LINKED ISSUE #26848: [BUG] Playwright should report Error#cause
### System info
- Playwright Version: v1.37.1
- Operating System: All
- Browser: All
- Other info:

### Source code

https://github.com/jan-molak/playwright-error-cause-serialization/blob/main/tests/example.spec.ts

```ts
test('error handling', async ({ page }) => {
  try {
    throw new Error('Nested error');
  }
  catch (error) {
    throw new Error('Top-level error', { cause: error })
  }
})
```

**Steps**
- Run tests
- Notice that the error stack trace DOES NOT contain information about the nested error cause

**Expected**

I'd expect Playwright to pass the information about the [Error#cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) when reporting errors that occurred in the test scenario. 

This might work like this, for example:

```
  1) [chromium] › example.spec.ts:20:5 › error handling ────────────────────────

    Error: Top-level error

      23 |   }
      24 |   catch (error) {
    > 25 |     throw new Error('Top-level error', { cause: error })
         |           ^
      26 |   }
      27 | })

        at /Users/jan/Projects/tmp/playwright-errors-repro/tests/example.spec.ts:25:11

    Cause: Error: Nested error
      21 |   try {
    > 22 |     throw new Error('Nested error');
         |           ^
      23 |   }
    
        at /Users/jan/Projects/tmp/playwright-errors-repro/tests/example.spec.ts:22:11        
```

**Actual**

At the moment, information about the `Error#cause` is lost when Playwright serializes the error. That's because:

- Error retrieved from the test has the cause information in the stack trace: https://github.com/microsoft/playwright/blob/v1.37.1/packages/playwright-test/src/worker/testInfo.ts#L236-L238
- but then `filteredStackTrace` removes this information as it doesn't look at `error.cause`
https://github.com/microsoft/playwright/blob/v1.37.1/packages/playwright-test/src/util.ts#L32-L49

This results in error reports that omit information about the underlying error:

```
  1) [chromium] › example.spec.ts:20:5 › error handling ────────────────────────

    Error: Top-level error

      23 |   }
      24 |   catch (error) {
    > 25 |     throw new Error('Top-level error', { cause: error })
         |           ^
      26 |   }
      27 | })

        at /Users/jan/Projects/tmp/playwright-errors-repro/tests/example.spec.ts:25:11
```

You can find the original report on this issue at serenity-js/serenity-js#1823, together with my notes on how I traced it to the root cause.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/26848

Thats us:
```
Error: Third level error

  30 |     }
  31 |   } catch (error) {
> 32 |     throw new Error('Third level error', { cause: error });
     |           ^
  33 |   }
  34 | });

    at fn (/Users/maxschmitt/Developer/playwright/tests/page/wheel.spec.ts:32:11)
[cause]: Error: Second level error
    at fn (/Users/maxschmitt/Developer/playwright/tests/page/wheel.spec.ts:29:13)
[cause]: Error: First level error
    at /Users/maxschmitt/Developer/playwright/tests/page/wheel.spec.ts:22:13
    at myFunc (/Users/maxschmitt/Developer/playwright/tests/page/wheel.spec.ts:23:6)
    at fn (/Users/maxschmitt/Developer/playwright/tests/page/wheel.spec.ts:27:7)
```

Thats by Node.js:

```
❯ node test.mjs 
file:///Users/maxschmitt/Developer/playwright/test.mjs:23
  throw new Error('Third level error', { cause: error });
        ^

Error: Third level error
    at file:///Users/maxschmitt/Developer/playwright/test.mjs:23:9
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25) {
  [cause]: Error: Second level error
      at file:///Users/maxschmitt/Developer/playwright/test.mjs:20:11
      at ModuleJob.run (node:internal/modules/esm/module_job:194:25) {
    [cause]: Error: First level error
        at file:///Users/maxschmitt/Developer/playwright/test.mjs:18:11
        at ModuleJob.run (node:internal/modules/esm/module_job:194:25)
  }
}

Node.js v18.18.2
```

### DIFF
--- a/packages/playwright/src/util.ts
+++ b/packages/playwright/src/util.ts
@@ -35,6 +35,12 @@ export function filterStackTrace(e: Error): { message: string, stack: string } {
     return { message: e.name + ': ' + e.message, stack: e.stack || '' };
 
   const stackLines = stringifyStackFrames(filteredStackTrace(e.stack?.split('\n') || []));
+  let cause: Error = e.cause as any;
+  while (cause) {
+    stackLines.push('[cause]: ' + cause.toString());
+    stackLines.push(...stringifyStackFrames(filteredStackTrace(cause.stack?.split('\n') || [])));
+    cause = cause.cause as any;
+  }
   return {
     message: e.name + ': ' + e.message,
     stack: `${e.name}: ${e.message}\n${stackLines.join('\n')}`
--- a/tests/playwright-test/reporter-base.spec.ts
+++ b/tests/playwright-test/reporter-base.spec.ts
@@ -89,6 +89,39 @@ for (const useIntermediateMergeReport of [false, true] as const) {
       expect(result.output).toContain(`> 4 |             const error = new Error('my-message');`);
     });
 
+    test('should report error.cause over multiple levels', async ({ runInlineTest }) => {
+      const result = await runInlineTest({
+        'a.spec.ts': `
+        import { test, expect } from '@playwright/test';
+        test('foo', () => {
+          const myFunc = () => {
+            (() => {
+              throw new Error('First level error');
+            })();
+          };
+          try {
+            try {
+              myFunc();
+            } catch (error) {
+              throw new Error('Second level error', { cause: error });
+            }
+          } catch (error) {
+            throw new Error('Third level error', { cause: error });
+          }
+        });
+        `
+      });
+      expect(result.exitCode).toBe(1);
+
+      expect(result.output).toContain('Third level error');
+      expect(result.output).toContain('[cause]: Error: Second level error');
+      expect(result.output).toContain('a.spec.ts:13:21');
+      expect(result.output).toContain('[cause]: Error: First level error');
+      expect(result.output).toContain('a.spec.ts:6:21');
+      expect(result.output).toContain('a.spec.ts:7:14');
+      expect(result.output).toContain('a.spec.ts:11:15');
+    });
+
     test('should filter out node_modules error in a codeframe', async ({ runInlineTest }) => {
       const result = await runInlineTest({
         'node_modules/utils/utils.js': `


tests/playwright-test/reporter-base.spec.ts
```diff
@@ -89,6 +89,39 @@ for (const useIntermediateMergeReport of [false, true] as const) {
...
+            }
+          } catch (error) {
+            throw new Error('Third level error', { cause: error });
+          }
+        });
+        `
+      });
+      expect(result.exitCode).toBe(1);
+
+      expect(result.output).toContain('Third level error');
```
mxschmitt, line 116: Should I be more strict here and asset that the full file paths + expected indention matches?

packages/playwright/src/util.ts
```diff
@@ -35,6 +35,12 @@ export function filterStackTrace(e: Error): { message: string, stack: string } {
     return { message: e.name + ': ' + e.message, stack: e.stack || '' };
 
   const stackLines = stringifyStackFrames(filteredStackTrace(e.stack?.split('\n') || []));
+  let cause: Error = e.cause as any;
+  while (cause) {
+    stackLines.push('[cause]: ' + cause.toString());
```
dgozman, line 40: I am somewhat worried that various tools expect stack trace lines to start with `     at`, and we will break them.

tests/playwright-test/reporter-base.spec.ts
dgozman: Not necessarily.

mxschmitt:
waiting for more customer requests before making changes to the API and risking breaking customers due to stack trace changes.

PR #28502, opened 2023-12-05
feat(trace): allow Trace Viewer to include credentials when fetching traces cross-origin

### LINKED ISSUE #28501: [Feature] Trace viewer: allow fetching traces from origins that require credentials
My team is working to allow our developers to load traces directly in the online [trace viewer](https://trace.playwright.dev/). Our ideal scenario is that a developer could open a test result document and jump directly to the trace in the viewer with a simple click of a link.

Our CI infrastructure uploads traces to a location protected by credentials — to fetch a trace, the request needs to contain valid authentication cookies. These cookies are sent if we open the trace URL in the browser directly, but we've found that the online trace viewer does the trace fetch _without_ cookies, even though the origin sends `Access-Control-Allow-Credentials: true`.[1]

We suspected this could be due to the trace viewer not applying the `credentials` option on the fetch request. We tested this theory by deploying the trace viewer locally and adding a `credentials: 'include'` option on [this line](https://github.com/microsoft/playwright/blob/8f056fbbcebe4dd1f1dcedebb43870b81cc475d1/packages/trace-viewer/src/traceModelBackends.ts#L35), and this allowed our local trace viewer to load the trace.

Our request is for this option to be added upstream. This would allow us to use the Playwright-hosted trace viewer rather than hosting it ourselves, and I suspect it would unblock other customers whose traces are credential-protected (assuming their origins also send the proper `Access-Control-Allow-Origin` and `Access-Control-Allow-Credentials` headers).

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials

pmattj:
Add the `'credentials': include` option on the trace fetch so the browser can include cookies when fetching from a different origin, assuming the origin returns the correct Access-Control-Allow-Origin and Access-Control-Allow-Credentials headers.

Fixes #28501

### DIFF
--- a/packages/trace-viewer/src/traceModelBackends.ts
+++ b/packages/trace-viewer/src/traceModelBackends.ts
@@ -32,7 +32,7 @@ export class ZipTraceModelBackend implements TraceModelBackend {
     this._traceURL = traceURL;
     zipjs.configure({ baseURL: self.location.href } as any);
     this._zipReader = new zipjs.ZipReader(
-        new zipjs.HttpReader(formatUrl(traceURL), { mode: 'cors', preventHeadRequest: true } as any),
+        new zipjs.HttpReader(formatUrl(traceURL), { mode: 'cors', credentials: 'include', preventHeadRequest: true } as any),
         { useWebWorkers: false });
     this._entriesPromise = this._zipReader.getEntries({ onprogress: progress }).then(entries => {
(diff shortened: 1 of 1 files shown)


dgozman (approved):
Thank you for the PR, looks great!

pmattj:
Thank you!

Issue #28524, opened 2023-12-06
[internal] failing tests in Node.js 21


mxschmitt:
```
    [playwright-test] › esm.spec.ts:138:5 › should use source maps ─────────────────────────────────
    [playwright-test] › esm.spec.ts:160:5 › should show the codeframe in errors ────────────────────
    [playwright-test] › esm.spec.ts:200:5 › should filter by line ──────────────────────────────────
    [playwright-test] › esm.spec.ts:481:5 › should resolve .js import to .tsx file in ESM mode for components 
    [playwright-test] › playwright.fetch.spec.ts:54:5 › should stop tracing on requestContext.dispose() 
    [playwright-test] › reporter-blob.spec.ts:593:1 › generate html with attachment urls ───────────
    [playwright-test] › reporter-html.spec.ts:224:5 › created › should include image diffs for same expectation 
    [playwright-test] › ui-mode-test-ct.spec.ts:113:5 › should run component tests after editing test and component 
    [playwright-test] › ui-mode-test-screencast.spec.ts:21:5 › should show screenshots ─────────────
    [playwright-test] › watch.spec.ts:60:5 › should print dependencies in ESM mode ─────────────────
    [playwright-test] › watch.spec.ts:377:5 › should run on changed deps in ESM ────────────────────
    [playwright-test] › watch.spec.ts:653:5 › should run CT on indirect deps change ESM mode ───────
```

and `library/browsercontext-fetch.spec.ts:788:3`

mxschmitt:
Caused by deprecated globalPreload look: https://nodejs.org/en/blog/announcements/v21-release-announce

jdalton:
:wave: @mxschmitt! I'm been digging into the playwright repo and how it runs its own unit tests. Are the loader hooks something that is used in public APIs or just internally? How does it fit into playwright?

mxschmitt:
Closing since this issue has been solved.

PR #28526, opened 2023-12-07
fix(esm): >= Node.js 21 globalPreload -> initialize

mxschmitt:
https://github.com/microsoft/playwright/issues/28524
https://github.com/microsoft/playwright/issues/28732

https://nodejs.org/en/blog/announcements/v21-release-announce#module-customization-hook-globalpreload-removed-use-register-and-initialize-instead

### DIFF
--- a/packages/playwright/src/cli.ts
+++ b/packages/playwright/src/cli.ts
@@ -21,7 +21,7 @@ import fs from 'fs';
 import path from 'path';
 import { Runner } from './runner/runner';
 import { stopProfiling, startProfiling, gracefullyProcessExitDoNotHang } from 'playwright-core/lib/utils';
-import { execArgvWithoutExperimentalLoaderOptions, execArgvWithExperimentalLoaderOptions, fileIsModule, serializeError } from './util';
+import { fileIsModule, serializeError } from './util';
 import { showHTMLReport } from './reporters/html';
 import { createMergedReport } from './reporters/merge';
 import { ConfigLoader, resolveConfigFile } from './common/configLoader';
@@ -33,6 +33,8 @@ import type { FullConfigInternal } from './common/config';
 import program from 'playwright-core/lib/cli/program';
 import type { ReporterDescription } from '../types/test';
 import { prepareErrorStack } from './reporters/base';
+import { registerESMLoader } from './common/esmLoaderHost';
+import { execArgvWithExperimentalLoaderOptions, execArgvWithoutExperimentalLoaderOptions, kSupportsModuleRegister } from './transform/esmUtils';
 
 function addTestCommand(program: Command) {
   const command = program.command('test [test-filter...]');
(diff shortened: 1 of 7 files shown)


packages/playwright/src/common/esmLoaderHost.ts
```diff
@@ -17,9 +17,22 @@
...
 import { PortTransport } from '../transform/portTransport';
+import url from 'url';
+import { hasExperimentalLoaderOptions } from '../util';
 
-const port = (globalThis as any).__esmLoaderPort;
-
+let port: MessagePort | undefined;
+if ((globalThis as any).__esmLoaderPort) {
+  port = (globalThis as any).__esmLoaderPort;
+  delete (globalThis as any).__esmLoaderPort;
```
yury-s, line 25: Why do we delete it?

packages/playwright/src/common/esmLoaderHost.ts
```diff
@@ -17,9 +17,21 @@
...
+} else if (process.env.PW_TS_ESM_ON) {
+  const { port1, port2 } = new MessageChannel();
+  require('node:module').register(require.resolve('../transform/esmLoader'), {
```
yury-s, line 28: In this case --experimental-loader flag is redundant, isn't it?

packages/playwright/src/common/esmLoaderHost.ts
```diff
@@ -17,9 +17,21 @@
...
+    parentURL: url.pathToFileURL(__filename),
+    data: { port: port2, from: 'playwright-test' },
+    transferList: [port2],
+  });
+  port = port1;
```
yury-s, line 33: As discussed, it seems that we don't need to restart any more but may need to wait for a response from our esm loader in case of Node 21+.

packages/playwright/src/transform/esmUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+import url from 'url';
+
+const kExperimentalLoaderOptions = [
+  '--no-warnings',
+  `--experimental-loader=${url.pathToFileURL(require.resolve('playwright/lib/transform/esmLoader')).toString()}`,
+];
+
+export const kSupportsModuleRegister = !!require('node:module').register;
```
yury-s, line 24: .register was `Added in: v20.6.0`, does the new logic work fine in v20?

packages/playwright/src/common/esmLoaderHost.ts
```diff
@@ -14,16 +14,37 @@
...
+// Node.js < 21
+if ((globalThis as any).__legacyEsmLoaderPort)
+  loaderChannel = createPortTransport((globalThis as any).__legacyEsmLoaderPort);
 
-const loaderChannel = port ? new PortTransport(port, async (method, params) => {
-  if (method === 'pushToCompilationCache')
-    addToCompilationCache(params.cache);
-}) : undefined;
+export function registerESMLoader() {
+  if (!kSupportsModuleRegister)
```
yury-s, line 29: This can never happen, right?

packages/playwright/src/common/esmLoaderHost.ts
```diff
@@ -14,16 +14,37 @@
...
+    return;
+  // Node.js >= 21
```
yury-s, line 31: Make it function comment?

packages/playwright/src/common/esmLoaderHost.ts
```diff
@@ -14,16 +14,37 @@
...
+  const { port1, port2 } = new MessageChannel();
+  // register will wait until the loader is initialized.
```
yury-s, line 33: Would be nice to put a link to the docs here, if there is one.

yury-s (approved):
Looks good. I'd recommend you have one more person have a look at this PR.

packages/playwright/src/transform/esmUtils.ts
mxschmitt: Yes it works in 20.X.

mxschmitt:
We'll hold off for now either until customer reports or until the Node.js 22 release.

packages/playwright/src/cli.ts
```diff
@@ -277,26 +279,33 @@ function restartWithExperimentalTsEsm(configFile: string | null): boolean {
     return false;
   if (process.env.PW_DISABLE_TS_ESM)
     return false;
-  if (process.env.PW_TS_ESM_ON) {
+  // Node.js < 20
+  if ((globalThis as any).__legacyEsmLoaderPort) {
```
pavelfeldman, line 283: Let's rename legacy to explicit version.

packages/playwright/src/cli.ts
```diff
@@ -277,26 +279,33 @@ function restartWithExperimentalTsEsm(configFile: string | null): boolean {
...
     return false;
-  const innerProcess = (require('child_process') as typeof import('child_process')).fork(require.resolve('./cli'), process.argv.slice(2), {
-    env: {
-      ...process.env,
-      PW_TS_ESM_ON: '1',
-    },
-    execArgv: execArgvWithExperimentalLoaderOptions(),
-  });
+    // Node.js < 20
+  if (!kSupportsModuleRegister) {
```
pavelfeldman, line 291: Same here, let's go by version.

Issue #28556, opened 2023-12-08
[BUG] locator `has` filter not working as expected


PsiKai:
### System info
- Playwright Version: 1.40.0
- Operating System: MacOS Sonoma 14.1.2
- Browser: All

### Source code

```js
import { test } from '@playwright/test';

test("should filter by 'has'", async ({ page }) => {
    await page.goto("https://nczdcr.csb.app/");
    const table = page.locator("table");
    const cell = table.getByRole("cell", { name: "Foo", exact: true });
    await cell.click();
    const row = table.locator("tr").filter({ has: cell });
    await row.click();
    const button = row.getByRole("button", { name: "Bar" });
    await button.click();
});
```

Testing against this [codesandbox](https://codesandbox.io/p/sandbox/table-example-for-playwright-nczdcr?file=%2Fsrc%2FApp.js%3A18%2C25)

**Steps**
- Copy paste the above test into your `test/` directory
- Execute test in any browser, headed or non.
- See failure to locate `row` element

**Expected**

Elements can be located by their child elements, as described in the [documentation](https://playwright.dev/docs/locators#filter-by-childdescendant).

**Actual**

```
locator.click: Timeout 10000ms exceeded.
=========================== logs ===========================
waiting for locator('table').locator('tr').filter({ has: locator('table').getByRole('cell', { name: 'Foo', exact: true }) })
```

refactoreric:
Hi, (disclaimer: I'm a community member, not a Playwright dev, and only half awake.)

I don't think this is a bug, but more a matter of confusion on a somewhat a confusing feature.
Right now, as you can see in the log, it's trying to find a tr in a table, which has a table inside, with a cell having name 'Foo'.
That's not the structure of the document.

The locators for the `has` filter should be kind of a partial/relative locator.
Right now, the cell locator is pretty absolute `page.locator("table").getByRole("cell", { name: "Foo", exact: true });`
If you would make it just `page.getByRole("cell", { name: "Foo", exact: true });` I think the `cell.click` by itself might sometimes find more than 1 match, but when used inside the `has` filter, it would work better than what you have now.

PsiKai:
> as you can see in the log, it's trying to find a tr in a table, which has a table inside, with a cell having name 'Foo'.
That's not the structure of the document.

If the inner locator would be relative to the locator that is being filtered, then the documentation is sorely lacking to mention that fact.  I do not interpret the log message that way.  There is no mention that it is searching relative to the `tr` element.

[...]

PR #28588, opened 2023-12-12
docs: emphasize that `has` is a relative locator

dgozman:
References #28556.

### DIFF
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -1023,8 +1023,10 @@ For example, `"Playwright"` matches `<article><div>Playwright</div></article>`.
 ## locator-option-has
 - `has` <[Locator]>
 
-Matches elements containing an element that matches an inner locator. Inner locator is queried against the outer one.
+Narrows down the results of the method to those which contain elements matching this relative locator.
 For example, `article` that has `text=Playwright` matches `<article><div>Playwright</div></article>`.
+
+Inner locator **must be relative** to the outer locator and is queried starting with the outer locator match, not the document root. For example, you can find `content` that has `div` in `<article><content><div>Playwright</div></content></article>`. However, looking for `content` that has `article div` will fail, because the inner locator must be relative and should not use any elements outside the `content`.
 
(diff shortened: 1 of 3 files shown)


docs/src/api/params.md
```diff
@@ -1023,9 +1023,11 @@ For example, `"Playwright"` matches `<article><div>Playwright</div></article>`.
 ## locator-option-has
 - `has` <[Locator]>
 
-Matches elements containing an element that matches an inner locator. Inner locator is queried against the outer one.
+Matches elements containing an element that matches an inner locator.
```
yury-s, line 1026:
```suggestion
Narrows down the results of the method to those which contain elements matching this relative locator.
```

docs/src/api/params.md
```diff
@@ -1023,9 +1023,11 @@ For example, `"Playwright"` matches `<article><div>Playwright</div></article>`.
...
 For example, `article` that has `text=Playwright` matches `<article><div>Playwright</div></article>`.
 
+Inner locator **must be relative** to the outer locator and is queried starting with the outer locator match, not the document root. For example, you can find `content` that has `div` in `<article><content><div>Playwright</div></content></article>`. However, looking for `content` that has `article >> div` will fail, because the inner locator must be relative and should not use any elements outside the `content`.
```
yury-s, line 1029: Let's use a css locator instead of `article >> div` ?

docs/src/locators.md
```diff
@@ -1014,6 +1016,55 @@ await Expect(Page
     .ToHaveCountAsync(1);
 ```
 
+The inner locator **must be relative** to the outer locator and is queried starting with the outer locator match, not the document root. Therefore, the following will not work, because the inner locator references the `<ul>` list element outside of the `<li>` list item:
```
yury-s, line 1019:
```suggestion
The inner locator **must be relative** to the outer locator and is queried starting with the outer locator match, not the document root. Therefore, the following will not work, because the inner locator references the `<ul>` list element inside of the `<li>` list item:
```
?

yury-s (approved):

PR #28604, opened 2023-12-12
fix: disable PaintHolding to be able to click in oopifs

### LINKED ISSUE #28023: [internal] 'should emit filechooser event for iframe' is flaky on Chromium
Investigation:

- image
- `Page.fileChooserOpened` CDP event does not fire.
- Adding a delay between `setContent` and the `click` makes it work 100/100 times.
  image

yury-s:
Fixes https://github.com/microsoft/playwright/issues/28023


mxschmitt (approved):

mxschmitt:
LGTM on the bots as well!

PR #28623, opened 2023-12-13
docs: Improve clarity on toHaveText and toContainText assertions

### LINKED ISSUE #28058: docs: toContainText check child elements?
So I did a search for this and surprisingly didn't find anything nor does the documentation state it. When we are asserting a locator has text.....does it also check descendant/child elements?

IE if I have a `<div>` and a `<span>` inside of it and my locator points to the `<div>` and I do `toContainText` on the `<div>` element. Will it still work? (Since it's technically "in" the `<div>` as well). (This is useful since I may have "cards" that are divs that contain text content and that's more obvious of a locator than adding a locator for a singlular `<span>` item or similar within the div card.

My assumption is this works but I figured it was worth asking since I surprisingly didn't see a question about this anywhere else.

geke-scottlogic:
This closes https://github.com/microsoft/playwright/issues/28058.

### DIFF
--- a/docs/src/api/class-locatorassertions.md
+++ b/docs/src/api/class-locatorassertions.md
@@ -933,7 +933,7 @@ await Expect(
 * langs:
   - alias-java: containsText
 
-Ensures the [Locator] points to an element that contains the given text. You can use regular expressions for the value as well.
+Ensures the [Locator] points to an element that contains the given text. All children of the element are evaluated for the specified text. You can use regular expressions for the value as well.
 
 **Details**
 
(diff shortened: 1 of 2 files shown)


geke-scottlogic:
@microsoft-github-policy-service agree company="Scott Logic"

docs/src/api/class-locatorassertions.md
```diff
@@ -933,7 +933,7 @@ await Expect(
 * langs:
   - alias-java: containsText
 
-Ensures the [Locator] points to an element that contains the given text. You can use regular expressions for the value as well.
+Ensures the [Locator] points to an element that contains the given text. All children of the element are evaluated for the specified text. You can use regular expressions for the value as well.
```
yury-s, line 936: This sounds to me as if Playwright would try each of the child elements and see if it contains the text. How about something like: "All nested elements will be considered when computing the text content of the element." ?

geke-scottlogic: Good point; thanks for the suggestion. I will implement the change.

yury-s (approved):

PR #28624, opened 2023-12-13
feat(ct): upgrade to vite-plugin-svelte 3

benmccann:
`@sveltejs/vite-plugin-svelte` 3 requires Svelte 4, so I added that as a peer dependency here.

The tests use an unmaintained `svelte-navigator` package, which doesn't support Svelte 4, so I had to add an `overrides` for that to ignore the peer dependency warning :confused:

### DIFF
--- a/packages/playwright-ct-svelte/package.json
+++ b/packages/playwright-ct-svelte/package.json
@@ -30,10 +30,19 @@
   },
   "dependencies": {
     "@playwright/experimental-ct-core": "1.41.0-next",
-    "@sveltejs/vite-plugin-svelte": "^2.1.1"
+    "@sveltejs/vite-plugin-svelte": "^3.0.1"
   },
   "devDependencies": {
-    "svelte": "^3.55.1"
+    "svelte": "^4.2.8"
+  },
+  "peerDependencies": {
+    "svelte": "^4.0.0",
+    "@sveltejs/vite-plugin-svelte": "^3.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@sveltejs/vite-plugin-svelte": {
+      "optional": true
+    }
   },
   "bin": {
     "playwright": "cli.js"
--- a/package-lock.json
+++ b/package-lock.json
@@ -102,22 +102,12 @@
       }
     },
     "node_modules/@ampproject/remapping": {
-      "version": "2.2.0",
-      "license": "Apache-2.0",
+      "version": "2.2.1",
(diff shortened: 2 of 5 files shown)


benmccann:
@sand4rt as a heads up, we released SvelteKit 2 yesterday, so I expect people to start hitting an issue with the versions used here pretty frequently. Hopefully we can get this out before too long

packages/playwright-ct-svelte/package.json
```diff
@@ -30,10 +30,19 @@
...
   "devDependencies": {
-    "svelte": "^3.55.1"
+    "svelte": "^4.2.8"
+  },
+  "peerDependencies": {
+    "svelte": "^4.0.0",
+    "@sveltejs/vite-plugin-svelte": "^3.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@sveltejs/vite-plugin-svelte": {
```
mxschmitt, line 43: why is this dependency optional?

pavelfeldman:
> The tests use an unmaintained svelte-navigator package, which doesn't support Svelte 4, so I had to add an overrides for that to ignore the peer dependency warning

Please feel free to modernize the tests under `tests/components/ct-svelte*` to make it match the latest best practices!

packages/playwright-ct-svelte/package.json
benmccann: It's already declared in `dependencies`, so the user doesn't need to install it - hence it's optional. But bad things will happen if the user installs v2 in their own project to use alongside this which is using v3, so that's why I made it a peerDependency as well. I'm not 100% sure how all the different package managers will handle this to be honest, but I'm relatively sure it shouldn't hurt at least. I'm hoping it would prevent people from using this alongside v2

Ultimately the best solution may be something like Storybook has with its support for various builders like Vite, webpack, etc. But in lieu of that major architectural change, this seemed like the best idea I could come up with

benmccann:
> Please feel free to modernize the tests under tests/components/ct-svelte* to make it match the latest best practices!

Thank you. I may do that in a follow up PR. The best practice really is to use SvelteKit. But one of the tests is trying to show that this works without Vite, which would necessitate not using SvelteKit and thus going off the beaten path and using something that I expect would be quite minimally maintained. Depending on the needs of the tests, one of the following alternatives may work:
- https://github.com/kriasoft/universal-router
- https://github.com/EmilTholin/svelte-routing
- https://github.com/ItalyPaleAle/svelte-spa-router

mxschmitt (approved):

PR #28636, opened 2023-12-14
chore: follow up to the stylePath review

pavelfeldman:


docs/src/test-snapshots-js.md
```diff
@@ -81,6 +81,42 @@ export default defineConfig({
 });
 ```
 
+### stylePath
```
dgozman, line 84: Can we link to this section from the params/expect options?

dgozman (approved):

PR #28658, opened 2023-12-14
chore: revert noWaitForFinish option, page.close does not wait for ro…

yury-s:
…ute handlers

### DIFF
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -647,15 +647,13 @@ export class RouteHandler {
   private readonly _times: number;
   readonly url: URLMatch;
   readonly handler: RouteHandlerCallback;
-  readonly noWaitOnUnrouteOrClose: boolean;
   private _activeInvocations: MultiMap<Page|null, { ignoreException: boolean, complete: Promise<void>, route: Route }> = new MultiMap();
 
-  constructor(baseURL: string | undefined, url: URLMatch, handler: RouteHandlerCallback, times: number = Number.MAX_SAFE_INTEGER, noWaitOnUnrouteOrClose: boolean = false) {
+  constructor(baseURL: string | undefined, url: URLMatch, handler: RouteHandlerCallback, times: number = Number.MAX_SAFE_INTEGER) {
     this._baseURL = baseURL;
     this._times = times;
     this.url = url;
     this.handler = handler;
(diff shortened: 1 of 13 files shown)


packages/playwright-core/src/client/network.ts
```diff
@@ -703,7 +701,7 @@ export class RouteHandler {
     // Note that context.route handler may be later invoked on a different page,
     // so we only swallow errors for the current page's routes.
     const handlerActivations = page ? this._activeInvocations.get(page) : [...this._activeInvocations.values()];
-    if (this.noWaitOnUnrouteOrClose || noWait) {
+    if (noWait) {
```
dgozman, line 704: nit: let's rename `noWait` into `behavior: 'wait' | 'ignoreErrors'` to be consistent?

packages/playwright-core/src/client/page.ts
```diff
@@ -560,17 +559,9 @@ export class Page extends ChannelOwner<channels.PageChannel> implements api.Page
     await this.close();
   }
 
-  private async _waitForActiveRouteHandlersToFinish() {
-    const promises = this._routes.map(routeHandler => routeHandler.stopAndWaitForRunningHandlers(this));
-    promises.push(...this._browserContext._routes.map(routeHandler => routeHandler.stopAndWaitForRunningHandlers(this)));
```
dgozman, line 565: Since we never call `stopAndWaitForRunningHandlers` on context's handlers and pass `page` anymore, perhaps we do not need the parameter and the map from page to activations?

packages/playwright-core/src/client/network.ts
```diff
@@ -703,7 +701,7 @@ export class RouteHandler {
...
       handlerActivations.forEach(h => h.ignoreException = true);
     } else {
       const promises = [];
```
dgozman, line 707: Do we still need the logic below about `route._didThrow`?

packages/playwright-core/src/client/page.ts
yury-s: Good point. Changed it to  `RouteHandler._ignoreException`.

packages/playwright-core/src/client/network.ts
yury-s: I believe we do as we should not be waiting for the handlers that threw an error already.

dgozman (approved):

PR #28662, opened 2023-12-15
fix imageDiffView.css

pussyck:
If the screenshots are not the same size, then the actual screenshot is stretched or compressed


pavelfeldman:
Thanks for your patch! I made a larger refactoring that fixes this issue among other things.

PR #28716, opened 2023-12-19
feat(expect): Allow expect timeout to be configured outside of the runner

### LINKED ISSUE #28702: [Feature] Make expect configurable when not using test runner
We are starting to use artillery with its playwright engine https://github.com/artilleryio/artillery/tree/main/packages/artillery-engine-playwright.

Artillery has its own runner so it does not interact with playwright within a test. This means we cannot use the playwright.config.ts or use features that only work inside of a test.

We have lots of existing page object models with lots of use of `expect` from `@playwright/test` we want to re-use as much as possible, things are working really well except for we cannot configure:

```ts
{
expect: {
    timeout: 15000,
}
}
```

I know we can call `.configure` and create our own modified expect but this does not seem to work for locators.
```ts
export const expect = baseExpect.configure({
  timeout: 15000,
});
```
i tried this and it seems to have no effect
image

I think this is happening here: https://github.com/microsoft/playwright/blob/e2c335a40cc49edbfc7a0b0596f71d4058f8dcfb/packages/playwright/src/matchers/expect.ts#L246 it seems there is no way to get the expect configuration into here.

I'm not sure if this is locator-specific or not.

Georgegriff:
Allow expect timeout to be configured outside of the runner

Fixes #28702

See the linked issue for use case

### DIFF
--- a/packages/playwright/src/matchers/expect.ts
+++ b/packages/playwright/src/matchers/expect.ts
@@ -243,6 +243,8 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
     }
     return (...args: any[]) => {
       const testInfo = currentTestInfo();
+      // We assume that the matcher will read the current expect timeout the first thing.
+      setCurrentExpectConfigureTimeout(this._info.timeout);
       if (!testInfo)
         return matcher.call(target, ...args);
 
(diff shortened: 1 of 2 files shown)


Georgegriff:
> @Georgegriff please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

pavelfeldman (approved):

PR #28723, opened 2023-12-19
feat(run-server): expose --host

mxschmitt:


packages/playwright-core/src/remote/playwrightServer.ts
```diff
@@ -73,13 +73,13 @@ export class PlaywrightServer {
...
     const wsEndpoint = await new Promise<string>((resolve, reject) => {
-      server.listen(port, () => {
+      server.listen(port, hostname, () => {
         const address = server.address();
         if (!address) {
           reject(new Error('Could not bind server socket'));
           return;
         }
-        const wsEndpoint = typeof address === 'string' ? `${address}${this._options.path}` : `ws://127.0.0.1:${address.port}${this._options.path}`;
+        const wsEndpoint = typeof address === 'string' ? `${address}${this._options.path}` : `ws://${hostname}:${address.port}${this._options.path}`;
```
mxschmitt, line 82: Note: We could use `address.address` there which would be more appropriate, but it results in `ws://::1:3000/` instead of what the user expects, `ws://localhost:3000/`. This is aligned with e.g. what we do in HTML report.

pavelfeldman (approved):

PR #28730, opened 2023-12-20
fix: input events triggered by playwright actions should have composed=true

### LINKED ISSUE #28726: [BUG] selectOption `input` event not `composed` when the actual browser event is
### System info
- Playwright Version: [v1.40]
- Operating System: macOS
- Browser: All
- Other info:

### Source code

Consider the following scenario:

```js
class CustomElement extends LitElement {
  handleInput(event) {
    console.log(event.composed); //logs `false` incorrectly
  }

  render() {
    <div class="internal-template">
        <select id="internal-select" @input={this.handleInput}>
          <option value="abc">1</option>
          <option value="def">2</option>
        </select>
    </div>
  }
}
customElements.define('some-el', CustomElement);
```
> This js using the Lit framework to create a custom element will create a custom element with a select in the shadow dom, and add an event listener for the `input` event and set the hander for that event to be the `handleInput` class method.

With document:
```html
<some-el></some-el>
```

and the locator:

```js
page.selectOption('def'); //selects the option with a `value` of `def`
```

When the event handler is called the `composed` property of the created `input` event is `false`. Where as in a non-headless browser installed directly on my mac, the same `console.log` logs `true`.

// test
image

// browser
image

**Expected**

The `input` event dispatched from selecting options is composed.

**Actual**

The `input` event dispatched from selecting options is not composed.

** Request**

Looking through the documentation, its not easy/possible to determine whether or not events emitted by Playwright commands are going to be `composed` or not. None of the event options like `bubbles` etc are listed. The documentation for `selectOption` only says that `change` and `input` events will be dispatched, but doesn't describe the config of those events in detail. For shadow DOM devs like myself, it would be good to get some deeper documentation around events

yury-s:
Fixes https://github.com/microsoft/playwright/issues/28726

### DIFF
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -732,7 +732,7 @@ export class InjectedScript {
     if (element.nodeName.toLowerCase() === 'input') {
       const input = element as HTMLInputElement;
       const type = input.type.toLowerCase();
-      const kInputTypesToSetValue = new Set(['color', 'date', 'time', 'datetime', 'datetime-local', 'month', 'range', 'week']);
+      const kInputTypesToSetValue = new Set(['color', 'date', 'time', 'datetime-local', 'month', 'range', 'week']);
       const kInputTypesToTypeInto = new Set(['', 'email', 'number', 'password', 'search', 'tel', 'text', 'url']);
       if (!kInputTypesToTypeInto.has(type) && !kInputTypesToSetValue.has(type)) {
         progress.log(`    input of type "${type}" cannot be filled`);
@@ -749,7 +749,7 @@ export class InjectedScript {
         input.value = value;
         if (input.value !== value)
           throw this.createStacklessError('Malformed value');
-        element.dispatchEvent(new Event('input', { 'bubbles': true }));
+        element.dispatchEvent(new Event('input', { 'bubbles': true, 'composed': true }));
         element.dispatchEvent(new Event('change', { 'bubbles': true }));
         return 'done';  // We have already changed the value, no need to input it.
       }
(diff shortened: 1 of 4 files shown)


packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -732,7 +732,7 @@ export class InjectedScript {
     if (element.nodeName.toLowerCase() === 'input') {
       const input = element as HTMLInputElement;
       const type = input.type.toLowerCase();
-      const kInputTypesToSetValue = new Set(['color', 'date', 'time', 'datetime', 'datetime-local', 'month', 'range', 'week']);
```
yury-s, line 735: There is no such input type as 'datetime'.

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -720,8 +720,8 @@ export class InjectedScript {
     select.value = undefined as any;
     selectedOptions.forEach(option => option.selected = true);
     progress.log('    selected specified option(s)');
-    select.dispatchEvent(new Event('input', { 'bubbles': true }));
-    select.dispatchEvent(new Event('change', { 'bubbles': true }));
+    select.dispatchEvent(new Event('input', { bubbles: true, composed: true }));
+    select.dispatchEvent(new Event('change', { bubbles: true }));
```
dgozman, line 724: I wonder whether `change` should also be composed?

dgozman (approved):

packages/playwright-core/src/server/injected/injectedScript.ts
yury-s: No, it's not.

PR #28749, opened 2023-12-21
chore: box soft steps

### LINKED ISSUE #28259: [BUG] Box step not works with expect.soft
### System info
- Playwright Version: [v1.39]
- Browser: [Chromium]
- Other info:

### Source code

So I added {box: true} in POM and it works perfect if I use expect alone, the error will directly point to my spec file. However, if I add expect.soft in the method then it will point out the error in the POM file not spec anymore.
w/o soft: 
image
with soft: 
image

**Steps**
- Used box step with method in POM
- Add expect.soft
- Run the test with invalid value to fail

**Expected**

The error should be shown in spec file

**Actual**

When add expect.soft to the POM method the error will be shown in POM file not spec even if option box = true

pavelfeldman:
Fixes https://github.com/microsoft/playwright/issues/28259

### DIFF
--- a/packages/playwright/src/worker/testInfo.ts
+++ b/packages/playwright/src/worker/testInfo.ts
@@ -27,6 +27,7 @@ import type { Location } from '../../types/testReporter';
 import { filteredStackTrace, getContainedPath, normalizeAndSaveAttachment, serializeError, trimLongString } from '../util';
 import { TestTracing } from './testTracing';
 import type { Attachment } from './testTracing';
+import type { StackFrame } from '@protocol/channels';
 
 export interface TestStepInternal {
   complete(result: { error?: Error | TestInfoError, attachments?: Attachment[] }): void;
@@ -35,6 +36,7 @@ export interface TestStepInternal {
   category: string;
   wallTime: number;
   location?: Location;
+  boxedStack?: StackFrame[];
   steps: TestStepInternal[];
   laxParent?: boolean;
   endWallTime?: number;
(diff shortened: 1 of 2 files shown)


packages/playwright/src/worker/testInfo.ts
```diff
@@ -281,13 +287,15 @@ export class TestInfoImpl implements TestInfo {
         let error: TestInfoError | undefined;
         if (result.error instanceof Error) {
           // Step function threw an error.
-          if (boxedStack) {
+          if (data.boxedStack) {
             const errorTitle = `${result.error.name}: ${result.error.message}`;
-            result.error.stack = `${errorTitle}\n${stringifyStackFrames(boxedStack).join('\n')}`;
+            result.error.stack = `${errorTitle}\n${stringifyStackFrames(data.boxedStack).join('\n')}`;
```
dgozman, line 292: So it looks like we actually do change the stack at runtime? Perhaps we should avoid modifying the error?

packages/playwright/src/worker/testInfo.ts
```diff
@@ -281,13 +287,15 @@ export class TestInfoImpl implements TestInfo {
...
           }
           error = serializeError(result.error);
         } else if (result.error) {
           // Internal API step reported an error.
+          if (data.boxedStack)
+            result.error.stack = `${result.error.message}\n${stringifyStackFrames(data.boxedStack).join('\n')}`;
```
dgozman, line 298: Should we also change `result.error.location` here?

packages/playwright/src/worker/testInfo.ts
pavelfeldman: Yes, I'll try to follow up with the fix.

packages/playwright/src/worker/testInfo.ts
pavelfeldman: There is no result.error.location

dgozman (approved):

PR #28762, opened 2023-12-22
chore: box step w/o modifying runtime errors

pavelfeldman:


packages/playwright/src/worker/testInfo.ts
```diff
@@ -45,6 +45,7 @@ export interface TestStepInternal {
   error?: TestInfoError;
   infectParentStepsWithError?: boolean;
   box?: boolean;
+  isSoft?: boolean;
```
dgozman, line 48: Is this the same flag as `infectParentStepsWithError`?

packages/playwright/src/worker/testInfo.ts
```diff
@@ -311,18 +305,20 @@ export class TestInfoImpl implements TestInfo {
...
-          error,
+          error: step.error,
         };
         this._onStepEnd(payload);
-        const errorForTrace = error ? { name: '', message: error.message || '', stack: error.stack } : undefined;
+        const errorForTrace = step.error ? { name: '', message: step.error.message || '', stack: step.error.stack } : undefined;
         this._tracing.appendAfterActionForStep(stepId, errorForTrace, result.attachments);
+
+        if (step.isSoft && result.error)
+          this._failWithError(result.error, false /* isHardError */);
```
dgozman, line 321: Instead of setting a symbol, perhaps we can just pass the step here directly? If that does not work, we need a comment that explains the situation where we have to pass the step through the symbol.

packages/playwright/src/matchers/toMatchSnapshot.ts
```diff
@@ -206,7 +206,7 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
       return this.createMatcherResult(message, true);
     }
     if (this.updateSnapshots === 'missing') {
-      this.testInfo._failWithError(serializeError(new Error(message)), false /* isHardError */);
+      this.testInfo._failWithError(new Error(message), false /* isHardError */);
```
dgozman, line 209: Is this the last caller outside of `TestInfo`? I wonder if we can do something here as well...

packages/playwright/src/matchers/expect.ts
```diff
@@ -248,45 +247,37 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
...
-          message: error.message,
-          stack: error.stack,
-        };
-        step?.complete({ error: serializedError });
-        if (this._info.isSoft)
-          testInfo._failWithError(serializedError, false /* isHardError */);
-        else
+        const error = new ExpectError(jestError, customMessage);
+        step.complete({ error });
+        if (!this._info.isSoft)
```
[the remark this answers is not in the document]
pavelfeldman: Only 2 cases and they require an extra conditional parameter.

packages/playwright/src/worker/testInfo.ts
pavelfeldman: Nope, that one changes dynamically.

packages/playwright/src/matchers/toMatchSnapshot.ts
pavelfeldman: I think it is the last one, but I am not sure there is something we can do.

dgozman (approved):

PR #28766, opened 2023-12-22
Update ci-intro.md

Jscripter-pk:


pavelfeldman:
We only accept fixes to the grammar as a part of the larger meaningful changes.

PR #28767, opened 2023-12-22
Add range input recording support

ruifigueira:
Playwright doesn't record range inputs because it consumes browser events to prevent the native action execution, and instead triggers a playwright action to replicate that same event. However, things like slider drag and drop on a range input are not properly replicated. That causes the slider not to move at all, because playwright just sends a click action into the center of the slider, while preventing me from manually sliding to the desired value:

chrome_ahWgoKXidX

To try it, just run:

```bash
npx playwright codegen 'data:text/html,<input type="range" min="0" max="10" value="5">' 
```

My approach was to discard the playwright action, and instead let all events do their thing, without consuming them. This simplifies the recording logic and the recorded page behaves much better, because it is handling all eventss in the exact same way as if the user was not recording at all. And with this approach, recording a range input automatically works. I just added a few more logic to skip the click action recording there, but that's basically it.

Here's a video of it working with the new changes:

chrome_i8oDvD74Un

I ran all tests in `tests/library/inspector`, and they all pass with just a minor change in one of the tests (playwright action was triggering more events than the actual browser events).

### DIFF
--- a/packages/playwright-core/src/server/injected/recorder.ts
+++ b/packages/playwright-core/src/server/injected/recorder.ts
@@ -188,7 +188,7 @@ class RecordActionTool implements RecorderTool {
       return;
     if (this._actionInProgress(event))
       return;
-    if (this._consumedDueToNoModel(event, this._hoveredModel))
+    if (this._consumedDueWrongTarget(event, this._hoveredModel))
       return;
 
     const checkbox = asCheckbox(this._recorder.deepEventTarget(event));
@@ -283,7 +283,7 @@ class RecordActionTool implements RecorderTool {
       }
 
       // Non-navigating actions are simply recorded by Playwright.
-      if (this._consumedDueWrongTarget(event))
+      if (this._consumedDueWrongTarget(event, this._activeModel))
         return;
       this._recorder.delegate.recordAction?.({
         name: 'fill',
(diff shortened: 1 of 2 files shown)


pavelfeldman:
You probably want to replace one line and add the range to the list here: https://github.com/microsoft/playwright/blob/5f14d427231e08bbb206127f4206c5db138c5aab/packages/playwright-core/src/server/injected/recorder.ts#L376

We like it when actions are performed by Playwright because it gives us more confidence that the recorded script will play.

ruifigueira:
> You probably want to replace one line and add the range to the list here:

True 😁 I already reverted all the other bits and left that change only.

 > We like it when actions are performed by Playwright because it gives us more confidence that the recorded script will play.

I thought it so, and I understand. However, I already hit some scenarios where the page behaved poorly with recording was enabled. For instance, when clicking some button on a scrollable area, playwright would start scrolling and click which was very strange (I know it now, it is playwright doing that magic, but nevertheless). And overall, interactions don't seem fluid in more complex pages, at least it's my feeling. But that is for another discussion.

ruifigueira:
It's failing with webkit, with the same problem I had with the other approach. For some reason, we get a click event with the body element as target and the input element as the hovered element.

A solution is to ensure the click event is handled only when the hovered element matches the event target

pavelfeldman (approved):

Issue #22791, opened 2023-05-03
[Feature] Har recording option for playwright test runner


Smrtnyk:
Similar to trace option, it would be cool to have har option specified.
Our use case is that when test fails on retry, we want to save har for it, since our product is about RUM, we heavily rely on network inspection, so we would need to save har per failed test.
Something I came up with is a custom fixture, that looks like this:
```typescript
recordHarFixture: [ async ({ recordHar, page }, use, testInfo) => {
        const shouldRecordHar = recordHar === "on-first-retry" && testInfo.retry === 1 || recordHar === "on";
        if (shouldRecordHar) {
            const harPath = testInfo.outputPath("network.har");
            await page.routeFromHAR(harPath, { update: true, updateContent: "embed" });
        }
        await use();
    }, { auto: true }],
```
This works, but maybe there is a better way to achieve this?
Another approach I can think of is overriding default context fixture, and calculating it and passing it manually to the browser options, but I like the one above more.

Please let me know if there is a better way to achieve this, via config?
If not is it possible to have it as a feature request?

yury-s:
Your approach looks good. Can you elaborate on why do you hesitate using it and what extra benefit a built-in option would give you?

Smrtnyk:
We already have piled up some fixtures, and this fixture also can be destructured from the test itself, which we would prefer that it would be private and not show up in autocompletion.
Doing something configurable like we can do for trace `trace: "on-first-retry",` looks more declarative and nicer, and there would be no extra code we would need to maintain.

cenfun:
Is it possible to provide some api similar to the [coverage api](https://playwright.dev/docs/api/class-coverage) that allows us to record har programmatically? for example:
```js
const page = await browser.newPage();
await page.harRecorder.start();
await page.goto('https://playwright.dev');
const harData = await page.harRecorder.stop();
fs.writeFileSync(myHarPath, JSON.stringify(harData));
```

yury-s:
@cenfun no such api at the moment. Feel free to file a separate feature request for this with a description of the use case.

yury-s:
@Smrtnyk another option would be to collect a trace with screenshots disabled. If you extract resulting trace.zip, you will find `trace.network` file which is essentially the har you are looking for

```js
    trace: {
      mode: 'on-first-retry',
      snapshots: true,
      screenshots: false,
      sources: false
    }
```
Will that work for you?

We might add har option to the config in the future, let's see if there is enough interest in the community.

[...]

PR #28796, opened 2023-12-27
feat(core): allow more fine granulary control over har files

chwoerz:
This PR solves two not so nice things about the har recorder. 

This partially solves the issue: https://github.com/microsoft/playwright/issues/22791

## Manual handling of updating HAR-files
We currently always have to change the `update` property when using `routeFromHar` manually, to enable and disable the update of our har-files.
This often gets overlooked and we recreate the har-files sometimes even when we don't want to or forget to update them. 
Currently we have to check if the files exist ourselfes.
With `ifNotExists` we allow the users to only update the har-files if they not exist.
```typescript
await page1.routeFromHAR(harPath, { update: 'ifNotExists', updateContent: 'embed', updateMode: 'full' });
```

## Handle saving of har-files 
Currently all har-files get recorded regardless if a testcase failes or not. 
By having an additional `saveHarFileOn` function we can check if we want to save the files. 
One example would be to check for the testCase result:

```typescript
await page1.routeFromHAR(harPath, {
    update: true,
    updateContent: 'embed',
    updateMode: 'full',
    saveHarFilesOn: _ => testInfo.status === 'passed'
  });
```

## Both new features combined

By combining these two features, we can now easily record new har files but only if the testcase was successful.
The users do no longer need to do this themselfes.

```typescript
await page1.routeFromHAR(harPath, {
    update: 'ifNotExists',
    updateContent: 'embed',
    updateMode: 'full',
    saveHarFilesOn: _ => testInfo.status === 'passed'
  });
``` 

## Breaking?
The `update` property has to be changed from `true` to `always` and `false` to `never`

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -1225,9 +1225,13 @@ Defaults to abort.
 
 ### option: BrowserContext.routeFromHAR.update
 * since: v1.23
-- `update` ?<[HarUpdateType]<boolean|"ifNotExists">>
+- `update` ?<[HarUpdate]<"always"|"never"|"ifNotExists">>
+* If set to 'always' the har files will be updated at every run.
+* If set to 'never' the har files will not be updated.
+* If set to 'ifNotExists' the har files will be created if they not exist.
 
-If specified as `true`, updates the given HAR with the actual network information instead of serving from file. If we specify it to `ifNotExists` it only updates the har-file if it not exists. The file is written to disk when [`method: BrowserContext.close`] is called.
+The file is written to disk when [`method: BrowserContext.close`] is called.
+
 
 ### option: BrowserContext.routeFromHAR.url
 * since: v1.23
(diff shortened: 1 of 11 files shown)


chwoerz:

docs/src/api/class-browsercontext.md
```diff
@@ -1225,9 +1225,13 @@ Defaults to abort.
 
 ### option: BrowserContext.routeFromHAR.update
 * since: v1.23
-- `update` ?<boolean>
+- `update` ?<[HarUpdate]<"always"|"never"|"ifNotExists">>
```
chwoerz, line 1228: I don't know how I can handle an update of a property. That's why I left it here at v1.23 but maybe its not really a good idea because the new options are only available from new versions.

pavelfeldman:
Please file an issue so that we could discuss the proposal there first! I'll close the PR, but we'll be able to re-open it once we agree on what gets implemented.

PR #28823, opened 2024-01-02
docs: remove the create method from extensibility

jason89521:
I saw the `create` method in the [Extensibility](https://playwright.dev/docs/extensibility) section of the document but didn't quite understand its actual purpose. Therefore, I attempted to add this method in the test examples, but it seems that regardless of what this method returns or if it throws an error, it has no effect. Upon further examination of the [selectors.ts](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/selectors.ts) and [types.ts](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/types.ts#L129) files, I noticed that this method is not defined. Consequently, I opened this pull request to remove the description of this method to prevent any potential misunderstandings by others.

If there are any misunderstandings, feel free to close this pull request. Thank you.

### DIFF
--- a/docs/src/extensibility.md
+++ b/docs/src/extensibility.md
@@ -7,8 +7,6 @@ title: "Extensibility"
 Playwright supports custom selector engines, registered with [`method: Selectors.register`].
 
 Selector engine should have the following properties:
-- `create` function to create a relative selector from `root` (root is either a `Document`, `ShadowRoot` or `Element`)
-  to a `target` element.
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):

jason89521:
@mxschmitt It seems that the workflows are not being executed. What should I do to kick-start them?

mxschmitt:
No action required from your side, we don't run certain workflows for docs only changes.

PR #28824, opened 2024-01-02
fix: change default color-scheme to no-preference

tmkx:
When connecting to an existing endpoint, Playwright will enforce the page to use a `light` color scheme.

https://github.com/microsoft/playwright/blob/76ace0fc09d92fedf2137cebd9a8a1427daa9911/packages/playwright-core/src/server/chromium/crPage.ts#L1080-L1093

because the default colorScheme is `light` rather than `no-preference`:
https://github.com/microsoft/playwright/blob/76ace0fc09d92fedf2137cebd9a8a1427daa9911/packages/playwright-core/src/server/page.ts#L444-L452

test code:
```ts
chromium.connectOverCDP('http://127.0.0.1:62374/'); // an existing dark page
```

### DIFF
--- a/packages/playwright-core/src/server/page.ts
+++ b/packages/playwright-core/src/server/page.ts
@@ -445,7 +445,7 @@ export class Page extends SdkObject {
     const contextOptions = this._browserContext._options;
     return {
       media: this._emulatedMedia.media || 'no-override',
-      colorScheme: this._emulatedMedia.colorScheme !== undefined ? this._emulatedMedia.colorScheme : contextOptions.colorScheme ?? 'light',
+      colorScheme: this._emulatedMedia.colorScheme !== undefined ? this._emulatedMedia.colorScheme : contextOptions.colorScheme ?? 'no-preference',
       reducedMotion: this._emulatedMedia.reducedMotion !== undefined ? this._emulatedMedia.reducedMotion : contextOptions.reducedMotion ?? 'no-preference',
       forcedColors: this._emulatedMedia.forcedColors !== undefined ? this._emulatedMedia.forcedColors : contextOptions.forcedColors ?? 'none',
     };


packages/playwright-core/src/server/page.ts
```diff
@@ -445,7 +445,7 @@ export class Page extends SdkObject {
     const contextOptions = this._browserContext._options;
     return {
       media: this._emulatedMedia.media || 'no-override',
-      colorScheme: this._emulatedMedia.colorScheme !== undefined ? this._emulatedMedia.colorScheme : contextOptions.colorScheme ?? 'light',
+      colorScheme: this._emulatedMedia.colorScheme !== undefined ? this._emulatedMedia.colorScheme : contextOptions.colorScheme ?? 'no-preference',
```
yury-s, line 448: Looks like it was changed to 'light' in https://github.com/microsoft/playwright/commit/ba29e99acec0462809a605ffb3d9e507b0fac083#diff-e2488cd41c99b3beafe59ebbf87b9ac81dcbdd1e6f6e363cfd016905eea3e1cf. @mxschmitt do you remember why?

I believe this entire return statement should be rewritten to something like:
```js
    return {
      media: this._emulatedMedia.media ?? 'no-override',
      colorScheme: this._emulatedMedia.colorScheme ?? contextOptions.colorScheme ?? 'no-override',
      reducedMotion: this._emulatedMedia.reducedMotion ?? contextOptions.reducedMotion ?? 'no-override',
      forcedColors: this._emulatedMedia.forcedColors ?? contextOptions.forcedColors ?? 'no-override',
    };
```

yury-s:
The reason it's set to 'light' is that we want to have consistent defaults across platforms and avoid test failures due to the host system changes. If you wan the color scheme not to be affected, you can explicitly set the option to `null`. If it doesn't work, can you provide a test?

tmkx:
I understand that it may be a by-design behavior, but it occurs in the first connect step and will turn all existing dark pages to light, additionally it cannot pass the `colorScheme` to the connect options:
```ts
const browser = await chromium.connectOverCDP('http://127.0.0.1:62374/', { /* connectOptions */ });
```

the workaround i'm using, but it will cause a screen flash:
```ts
await Promise.all(
  browser
    .contexts()
    .flatMap((context) => context.pages().map((page) => page.emulateMedia({ colorScheme: 'no-preference' })))
);
```

yury-s:
> I understand that it may be a by-design behavior, but it occurs in the first connect step and will turn all existing dark pages to light

The PR fixes a bug by breaking another (way more common) behavior which we cannot afford. Also, this fixes only one option of a few, there are others like `reducedMotion` and `forcedColors` in the same method that may also be changed on connect. If we were to fix color scheme, those options would need to be fixed as well. It seems that a proper fix would be to not enforce normal defaults when connecting over CDP and I'm not sure we want to introduce that complexity to the code base given that connecting to existing context is an edge case scenario compared to the common one where the client creates the context and controls all the options.

tmkx:
thanks for your detailed reply! ❤️

tmkx:
BTW, can we achieve it by passing `BrowserNewContextParams` to the `connect` function?

https://github.com/microsoft/playwright/blob/36c4c24f8a2e073737ae1a0ae4aa5cb42186b669/packages/playwright-core/src/server/chromium/chromium.ts#L100-L106

for example:
```ts
chromium.connectOverCDP('http://127.0.0.1:62374/', {
  persistent: { colorScheme: 'no-override', reducedMotion: 'no-override', /* ... */ }
});
```

this doesn't change the default behavior

PR #28849, opened 2024-01-03
docs: update merge example to upload-artifact@v4

### LINKED ISSUE #28800: [REGRESSION]: merge-reports example for sharding no longer works on @v4 GH actions
**Context:**
- GOOD Playwright Version:N/A
- BAD Playwright Version:N/A
- Operating System: GH actions ubuntu-latest
- Extra: N/A

**Code Snippet**

See [example](https://playwright.dev/docs/test-sharding#github-actions-example)

**Describe the bug**

When following the GH actions example for Playwright, it works great for `upload-artifact@v3` and `download-artifact@v3`, however the same script no longer works on the v4 versions of the scripts. I see the following error

image

It seems like the reports in each shard are somehow conflicting. Can an updated example for upload and download artifact v4 be provided?

yury-s:
Artifacts in upload-artifact@v4 are [immutable](https://github.com/actions/upload-artifact/tree/main?tab=readme-ov-file#breaking-changes), so we cannot simply copy reports into single artifact.

Fixes https://github.com/microsoft/playwright/issues/28800

### DIFF
--- a/docs/src/test-sharding-js.md
+++ b/docs/src/test-sharding-js.md
@@ -51,9 +51,9 @@ GitHub Actions supports [sharding tests between multiple jobs](https://docs.gith
 
 The following example shows you how to configure a job to run your tests on four machines in parallel and then merge the reports into a single report. Don't forget to add `reporter: process.env.CI ? 'blob' : 'html',` to your `playwright.config.ts` file as in the example above.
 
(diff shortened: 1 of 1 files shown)


mxschmitt (changes requested):
Line 54 needs a massage.

mxschmitt (approved):

Issue #26822, opened 2023-08-31
[Feature] Support adding custom loaders to the test runner


AlessioGr:
Ever since switching from CJS to ESM, we've needed a custom node loader to process .scss, .svg and png file imports.

For playwright, we had to create a [patch file](https://github.com/payloadcms/payload/blob/2.0/patches/%40playwright__test%401.37.1.patch) which patches playwright to point to our [custom loader](https://github.com/payloadcms/payload/blob/2.0/packages/payload/src/esm-loader-playwright.mjs).

Without this loader, we'd get the following error:
![Screenshot 2023-08-31 at 19 33 25@2x](https://github.com/microsoft/playwright/assets/70709113/80a4b58d-3235-4fc3-b690-4da053087e7a)

Instead of us having to patch playwright, could you either
- add a config option which allows us to add custom node loaders, or
- natively support scss/svg/png in [playwright's own loader](https://github.com/microsoft/playwright/blob/c92ad10f60f5f4f4cecbd671a03e9288ec2290c7/packages/playwright-test/src/transform/esmLoader.ts)?

hpierre74:
any update on this ?

jmikrut:
Also looking for an update on this one. We've patched this here:

https://github.com/payloadcms/payload/blob/alpha/patches/playwright%401.42.1.patch

But would love to sunset our reliance on patching Playwright. 

Thank you!

divmgl:
Currently running into this issue in our monorepo. It'd be nice to be able to use `tsx` to overcome the ESM/CJS module hell in Playwright.

avallonazevedo:
That'd be a fantastic feature. Any updates?

karlhorky:
@AlessioGr @jmikrut I noticed that you removed the Playwright patch - the `patches/` folder is no longer accessible:

- https://github.com/payloadcms/payload/tree/main/patches/

Did you come up with a different solution for this?

Would love to be able to load `.mdx` files using [`@mdx-js/node-loader`](https://www.npmjs.com/package/@mdx-js/node-loader) in test files!

karlhorky:
## Workaround: Node.js ESM Loaders via Customization Hooks

[...]

PR #28875, opened 2024-01-05
Add Bun support

### LINKED ISSUE #27139: [Feature] Compatibility with `bun`
https://github.com/microsoft/playwright/pull/26454#issuecomment-1685131076 states:

> Let's wait for the users to request interop with bun.

I'm a user, and I am requesting interop with `bun`! 🙋😄

I've just switched my largest project to `bun` (now that `bun` 1.0 [has been released](https://bun.sh/blog/bun-v1.0)), and everything works perfectly except running tests in Playwright.

A *lot* of packages run in `bun` without issue, but both `bun` and Playwright have difficult jobs, so it's not surprising that there would be challenges. Like `npx playwright`, Bun supports TypeScript and running tests directly — so I think it's fair to say that a developer who likes the ergonomics of either would be a good candidate for interest in the other.

https://github.com/oven-sh/bun/issues/2492 documents many folks trying to use Playwright from `bun`, although I don't understand the situation well enough to help fix it. But if there is anything Playwright can do to help be more compatible in `bun`, I'd be glad to be able to use that as soon as possible.

Jarred-Sumner:
Fixes https://github.com/microsoft/playwright/issues/27139

This adds a `package.json` `"exports"` condition for Bun which always chooses the CommonJS version of the module (for both ESM & CJS).

The built version of playwright uses `__esModule` to tell bundlers the code was ESM transplied into CommonJS. Bun supports this both at runtime and when using the bundler, but Node.js does not support this at runtime and that causes a compatibility issue when importing `@playwright/test`. It's unfortunately a compatibility issue whether or not we support `__esModule`, because choosing to not support it breaks libraries which otherwise only work when using a bundler.

Since Bun automatically converts CommonJS <> ESM and `module.exports` with dynamic properties are supported, we can choose to always use the CommonJS entry point for Playwright in Bun.

There were other changes necessary to get Playwright to work, which are mostly in https://github.com/oven-sh/bun/pull/7958. After this PR, the remaining blocker for using Playwright in Bun is a stability issue when multiple workers are in use, which we are working on fixing.

image

### DIFF
--- a/packages/playwright/package.json
+++ b/packages/playwright/package.json
@@ -14,6 +14,7 @@
   "exports": {
     ".": {
       "types": "./index.d.ts",
+      "bun": "./index.js",
       "import": "./index.mjs",
       "require": "./index.js",
       "default": "./index.js"
@@ -27,6 +28,7 @@
     "./lib/internalsForTest": "./lib/internalsForTest.js",
     "./lib/plugins": "./lib/plugins/index.js",
     "./jsx-runtime": {
+      "bun": "./jsx-runtime.js",
       "import": "./jsx-runtime.mjs",
       "require": "./jsx-runtime.js",
       "default": "./jsx-runtime.js"
@@ -41,6 +43,7 @@
     },
     "./test": {
       "types": "./test.d.ts",
+      "bun": "./test.js",
       "import": "./test.mjs",
       "require": "./test.js",
       "default": "./test.js"


nektro:
> the remaining blocker for using Playwright in Bun is a stability issue when multiple workers are in use

fixed! 

image

LevanKvirkvelia (commented):
LGTM

pavelfeldman:
Trying to understand the core of the issue. Wouldn't this be a problem for any compiled library that exports both CJS and ESM variants? I.e. shouldn't this be addressed on the Bun side for Node interoperability?

jdalton:
Related to https://github.com/microsoft/playwright/pull/7328/files.

Instead of tweaking the configs. What if we just remove:
https://github.com/JoelEinbinder/playwright-1/blob/41cd3802a13911287c804971246bd34b145f348b/packages/playwright-test/index.js#L23
> Object.defineProperty(combinedExports, '__esModule', { value: true });

It was added as part of https://github.com/microsoft/playwright/pull/7328/file to work with
https://www.typescriptlang.org/tsconfig#esModuleInterop but it may not be needed (could be an extra bit thrown in)

pavelfeldman:
@jdalton that's a nice finding. This line predates our ESM support and I don't see why we would need it today. I'll run all the bots w/o this line.

This breakage still suggests that there is an interoperability problem for Bun vs Node, not sure how important it is to the Bun folks though.

jdalton:
> @jdalton that's a nice finding. This line predates our ESM support and I don't see why we would need it today. I'll run all the bots w/o this line.
> 

Awesome! 🕺 

> This breakage still suggests that there is an interoperability problem for Bun vs Node, not sure how important it is to the Bun folks though.

This breakage is related to 
```js
Object.defineProperty(combinedExports, '__esModule', { value: true }); 
```
The issue is by design though the heuristic when to respect the `__esModule` annotation can totally be tweaked _(I'm looking into it - new 🫓  core here!)_. The issue _(simplified)_ is that Bun will try to respect the `__esModule` annotation. So when a CJS module has that it will treat it more like an ESM export having the `module.exports` properties treated as the specifiers. So when this CJS module was imported into another ESM module instead of the `module.exports` being treated as the `default` it was spread across as specifiers.

pavelfeldman:
One test failed for us without it:

```js
import t from '@playwright/test';
t('passed', () => {
  t.expect(1 + 1).toBe(2);
});
```

We compile Playwright using Babel. We also transpile TS tests on the fly using Babel. The Babel-to-babel interop works for us out of the box via __esModule:

```js
// https://babeljs.io/docs/babel-plugin-transform-modules-commonjs#strict
Object.defineProperty(combinedExports, '__esModule', { value: true });
```

and

```js
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
```

on the importing side.

In case of Bun, it takes over the transpilation of TS, so it seems like we need to align on the convention we are using. If Bun acts as Babel and trusts __esModule to already have `default` baked in, things should just work (tm).

---

Side note:
Playwright assumes it performs the TS(X) transpilation, which does not happen with Bun - bun takes care of it on its own. Assumptions about the transpilation between Bun and Playwright will differ, so some things won't work. At least component testing (TSX transform) and watch mode (compilation cache) won't work. Most of it is probably fixable though if we take it one step at a time.

pavelfeldman:
I landed https://github.com/microsoft/playwright/pull/28935, could you confirm it addresses your scenario?

mxschmitt:
This should work once https://github.com/microsoft/create-playwright/pull/108 has been landed and released: It was otherwise using npx/npm internally.

```bash
bun create playwright --next buntest
cd buntest
bun --bun x playwright install --with-deps
bun --bun x playwright test
```

Sometimes getting `Error: worker process exited unexpectedly (code=0, signal=null)` which seem the mentioned stability issues in https://github.com/microsoft/playwright/pull/28875#issue-2066851716.

Jarred-Sumner:
I suspect the worker issue still happens but much less frequently than before. Our `Worker` implementation isn't stable yet

pavelfeldman:
Closing this PR then. We can have a new tracking meta-issue for Bun in the Issues to discuss problems like this one!

karlhorky:
Is there already a new meta-issue?

Or should the old Bun issue be reopened?

- #27139

karlhorky:
With Bun support, I'm wondering whether that would also unlock custom loaders, as the community has asked for in a few places:

- https://github.com/microsoft/playwright/issues/26822
- https://github.com/microsoft/playwright/issues/31689

milosworks:
> Closing this PR then. We can have a new tracking meta-issue for Bun in the Issues to discuss problems like this one!

Wheres the meta-issue for Bun then?
Honestly i have not seen anything since 2024, Bun was also doing a lot to try and support playwright but i honestly, im wondering, whats the current status? There has been 0 updates on if playwright is going to ever work on bun.

Last time i tried playwright on bun, a few days ago, playwright couldn't even start the browser process

PR #28885, opened 2024-01-06
fix(chromium): get response status from extra info

### LINKED ISSUE #28779: [BUG] Different response code recorded in Playwright and Browser DevTools
While using `page.waitForResponse()` response status is logged as `200`, and in DevTools there is `304` (test run with VSC extension Browser and DevTools enabled)
in console
obraz
in browser
obraz

### System info
- Playwright Version: [v1.40.1]
- Browser: [Chromium]
- Other info: VSC with plugin Playwright Test for VSCode 1.0.20

### Source code

```typescript

test('test 304', async ({page}) => {
  const response200 = page.waitForResponse(
    (response) => {
      console.log( response.status(), response.url() );
      return response.url() ===  'https://automationintesting.online/'
    }
  );

  await page.goto('https://automationintesting.online/');
  expect((await response200).status()).toBe(200);
  
  const response304 = page.waitForResponse(
    (response) => {
      console.log( response.status(), response.url() );
      return response.url() ===  'https://automationintesting.online/'
    }
  );

  await page.reload();
  expect((await response304).status()).toBe(304);
  
});
```

**Steps**
- Enable option in VSC in Testing `Show browser`
- Open browser (run any test form VSC plugin)
- Open DevTools in Browser and go to Network tab
- Run the test `test 304`
- Compare console log response status with same request in DevTools Network for `https://automationintesting.online/`

**Expected**
Same response code

**Actual**
Different response codes:
200 logged by page.waitForResponse()
304 in DevTools

yury-s:
response/requestfinished/requestfailed events are deferred until `Network.responseReceivedExtraInfo` CDP event is received (if the response has `hasExtraInfo=true`).

In case of a redirect chain, it's safe to dispatch all pending events for a response once both of its Network.responseReceived and Network.responseReceivedExtraInfo events received, since by that time all previous (redirected) response events should have been received already. Otherwise it would mean that we received both Network.responseReceived and Network.responseReceivedExtraInfo for an earlier event before the current one.

Fixes https://github.com/microsoft/playwright/issues/28779

### DIFF
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -666,6 +666,9 @@ class ResponseExtraInfoTracker {
   requestServedFromCache(event: Protocol.Network.requestServedFromCachePayload) {
     const info = this._getOrCreateEntry(event.requestId);
     info.servedFromCache = true;
+    // TODO: should the info be deleted if the response is already there?
+    if (info.responses.length)
+      this._dispatchPendingResponseEvents(info.responses[info.responses.length - 1]);
   }
 
   servedFromCache(requestId: string): boolean {
@@ -725,11 +728,17 @@ class ResponseExtraInfoTracker {
       return;
     }
     const index = info.responses.indexOf(response);
-    if (info.responseReceivedExtraInfo[index]) {
+    if (info.responseReceivedExtraInfo[index] || info.servedFromCache) {
       callback();
       return;
     }
     this._pendingResponseEvents.set(response, callback);
+  }
+
+  private _dispatchPendingResponseEvents(response: network.Response) {
+    for (const callback of this._pendingResponseEvents.get(response))
+      callback();
+    this._pendingResponseEvents.deleteAll(response);
   }
 
   private _getOrCreateEntry(requestId: string): RequestInfo {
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -735,7 +770,8 @@ class ResponseExtraInfoTracker {
     if (response && responseExtraInfo) {
       response.setResponseHeadersSize(responseExtraInfo.headersText?.length || 0);
       response.setRawResponseHeaders(headersObjectToArray(responseExtraInfo.headers, '\n'));
-      info.responseReceivedExtraInfo[index] = undefined;
+      response.setRawStatus(responseExtraInfo.statusCode, getStatusText(responseExtraInfo.headersText));
```
mxschmitt, line 773: in case of HTTP/2 or QUIC headersText is empty

yury-s: Yes. What do you suggest?

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -311,7 +311,11 @@ export class CRNetworkManager {
       // right away, so that client can call it from the route handler.
       request.request.setRawRequestHeaders(headersObjectToArray(requestPausedEvent.request.headers, '\n'));
     }
-    (this._page?._frameManager || this._serviceWorker)!.requestStarted(request.request, route || undefined);
+    const dispatchRequestStarted = () => (this._page?._frameManager || this._serviceWorker)!.requestStarted(request.request, route || undefined);
+    if (redirectedFrom)
+      this._responseExtraInfoTracker.dispatchWhenResponseReady(redirectedFrom._requestId, redirectedFrom.request._existingResponse(), dispatchRequestStarted);
```
pavelfeldman, line 316: Instead of the callback indirection that elevates the complexity, I would prefer an explicit approach where you pass the event target explicitly in the tracker constructor and let it dispatch these events from its own code. Would be easier to follow.

yury-s:
Discussed in the team meeting and decided that the edge case feature doesn't justify added complexity.

PR #28901, opened 2024-01-08
fix(deps): check only .dll/.exe files on Windows

### LINKED ISSUE #28846: [BUG]_ Getting " PrintDeps.exe Bad image" pop-ups while running tests from terminal using ibm_db package with Playwright with JavaScript
System info
Playwright Version: v.1.40.1, JavaScript
Operating System: Windows 11 Enterprise 64, Version 10.0.19045 Build 19045
Browser: All (for example: Chromium version 
Node version: v.18.19.0
ibm_db v.3.2.1 (https://www.npmjs.com/package/ibm_db)

We are using Playwright with Javascript for our UI automation where we are also establishing a connection to IBM Data base by using following two :-

  var ibmdb = require('ibm_db')
  const open = util.promisify(ibmdb.open)

When I am executing the script I am getting 8 different PrintDeps.exe bad image pop-ops where I have to close all those 8 pop-ups manually and that's when the execution is starting.

We are using version :- "@playwright/test": "^1.40.1".

Note:- I did placed the " var ibmdb = require('ibm_db')" inside the function and I was able to execute the script without any issues. Actual issue comes when I execute my test scripts in Parallel or sequentially, first test script will get executed successfully and as soon second test script starts, we are seeing following 8 pop-ups.

Observation :- When we do npx playwright install on my machine it is downloading "chromium-1091" folder in "C:\Users\Userid\AppData\Local\ms-playwright" folder where I think it is causing the issue.

image

image

image

image

image

image

image

Any help is much appreciated.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/28846

### DIFF
--- a/packages/playwright-core/src/server/registry/dependencies.ts
+++ b/packages/playwright-core/src/server/registry/dependencies.ts
@@ -55,7 +55,11 @@ export function readDockerVersionSync(): null | { driverVersion: string, dockerI
   }
 }
 
-const checkExecutable = (filePath: string) => fs.promises.access(filePath, fs.constants.X_OK).then(() => true).catch(e => false);
+const checkExecutable = (filePath: string) => {
+  if (process.platform === 'win32')
+    return filePath.endsWith('.exe');
+  return fs.promises.access(filePath, fs.constants.X_OK).then(() => true).catch(() => false);
+};
 
 function isSupportedWindowsVersion(): boolean {
   if (os.platform() !== 'win32' || os.arch() !== 'x64')


packages/playwright-core/src/server/registry/dependencies.ts
```diff
@@ -290,7 +290,9 @@ async function executablesOrSharedLibraries(directoryPath: string): Promise<stri
     const basename = path.basename(filePath).toLowerCase();
     if (isSharedLib(basename))
       return filePath;
-    if (await checkExecutable(filePath))
+    if (process.platform !== 'win32' && await checkExecutable(filePath))
```
yury-s, line 293: Would it make sense to change checkExecutable to do the right thing on windows instead?

yury-s (approved):

54995610:
Hello Team, 

Can you please let me know if the issue related to printDeps.exe got resolved ? If so, request you kindly let me know in which version it got updated. 

As per the comments mentioned in the original issue (https://github.com/microsoft/playwright/pull/28901), I am trying to install ""@playwright/test": "^1.41.0",", but it says no matching version.

Any help is much appreciated, Thanks.

mxschmitt:
`npm i -D @playwright/test@1.41.0-alpha-jan-10-2024` should make it work.

54995610:
I tried this version but still the issue exists. 

The version what I have is "@playwright/test": "^1.41.0-alpha-jan-10-2024",

mxschmitt:
I tried it with `1.41.0-alpha-jan-10-2024` and `1.41.0-alpha-jan-11-2024` where both were working as expected. `1.41.0-alpha-jan-9-2024` doesn't work and ends up in this popup.

Did you do `npm install` after you modified the `package.json`?

Do you have multiple Playwright packages in your package.json? e.g. playwright, playwright-core? If so, you need to align all of them with this version.

54995610:
Thank you for responding.

Yes, I did npm install after adding playwright/test and still seeing the error pop-ups, and below is the screenshot of my package,json,
image


Also, as per above comment("node_modules/playwright-core/src/server/registry/dependencies.js")  I am not seeing src folder in the following screenshot.

image

I assume "@applitools/eyes-playwright" is only compatible with playwright@1.40.1.

Any help is much appreciated.

mxschmitt:
Have you tried if it reproduces with a new Playwright project? Might be caused by applitools eyes which sticks too an older Playwright version. I recommend trying on a new project and then seeing if it passes there or not (so we know if its caused by something in your existing larger project or not).

54995610:
Issue is resolved after performing following steps: -

1) Created a new project with the new playwright version (1.41.0-alpha-jan-11-2024)
2) Once node modules folder is created, copied the "dependencies.js" file from the location "node_modules/playwright-core/src/server/registry/dependencies.js") and pasted it my old original project, 
3) I have not updated any of my existing dependencies versions, just replacing this file, it is working like a champ. (It is working both parallel and also sequentially).

Thank you so much for creating a new version for the issue which helped us alot. Appreciate all the work you guys do for us.

PR #28921, opened 2024-01-09
chore(firefox): rewrite error message for $HOME directory ownership mismatch

### LINKED ISSUE #28896: [BUG]tests are failing for firefox browser in docker container
### System info
- Playwright Version: [v1.40.1]
- Operating System: [ Ubuntu 20]
- Browser: [Firefox]
- Other info:
- Docker image: mcr.microsoft.com/playwright:v1.40.0-jammy

### Source code

[We will only be able to work on the issues that we can reproduce.]
when we try to execute tests in parallel mode using github action workflow in docker container using the 1.40.0-jammy latest version the test are failing only in firefox browser where as passed in webkit and chrome browsers.

image

image

image

image

image

image

image

**Steps**
npx playwright test

**Expected**

All executed tests should work in all browsers in parallel mode with out any issue.

**Actual**

tests are failing in firefox browser

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/28896

Looks like this: 
image

### DIFF
--- a/packages/playwright-core/src/server/firefox/firefox.ts
+++ b/packages/playwright-core/src/server/firefox/firefox.ts
@@ -40,6 +40,9 @@ export class Firefox extends BrowserType {
   _doRewriteStartupLog(error: ProtocolError): ProtocolError {
     if (!error.logs)
       return error;
+    // https://github.com/microsoft/playwright/issues/6500
+    if (error.logs.includes(`as root in a regular user's session is not supported.`))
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/firefox/firefox.ts
```diff
@@ -40,6 +40,9 @@ export class Firefox extends BrowserType {
   _doRewriteStartupLog(error: ProtocolError): ProtocolError {
     if (!error.logs)
       return error;
+    // https://github.com/microsoft/playwright/issues/6500
```
yury-s, line 43: Let's check that we are running in Docker too?

yury-s (approved):

PR #28924, opened 2024-01-10
chore: network panel polish

pavelfeldman:


packages/web/src/components/splitView.tsx
```diff
@@ -25,12 +25,11 @@ export type SplitViewProps = {
   orientation?: 'vertical' | 'horizontal';
   minSidebarSize?: number;
   settingName?: string;
-  children: JSX.Element | JSX.Element[] | string;
 };
 
 const kMinSize = 50;
 
-export const SplitView: React.FC<SplitViewProps> = ({
+export const SplitView: React.FC<React.PropsWithChildren<SplitViewProps>> = ({
```
dgozman, line 32: Do you want to use ResizerView here as well?

packages/web/src/components/gridView.tsx
```diff
@@ -0,0 +1,107 @@
...
+  render: (item: T, column: keyof T, index: number) => React.ReactNode,
+  sorting?: Sorting<T>,
+  setSorting?: (sorting: Sorting<T> | undefined) => void,
+};
+
+export function GridView<T>(model: GridViewProps<T>) {
+  const initialOffsets: number[] = [];
+  for (let i = 0; i < model.columns.length - 1; ++i) {
+    const column = model.columns[i];
+    initialOffsets[i] = (initialOffsets[i - 1] || 0) + model.columnWidth(column);
```
dgozman, line 38: Do we want this computation to be done once instead of on every render, e.g. with `useMemo()`?

packages/web/src/components/gridView.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <span className='grid-view-header-cell-title'>{model.columnTitle(column)}</span>
+            <span className='codicon codicon-triangle-up' />
+            <span className='codicon codicon-triangle-down' />
+          </div>;
+        })}
+      </div>
+      <ListView
+        name={model.name}
+        items={model.items}
+        id={model.id ? ((item, index) => model.id!(item, index)) : undefined}
```
dgozman, line 74:
```suggestion
        id={model.id}
```

packages/web/src/components/gridView.tsx
```diff
@@ -0,0 +1,107 @@
...
+            {model.columns.map((column, i) => {
+              return <div
+                className='grid-view-cell'
+                style={{ width: offsets[i] - (offsets[i - 1] || 0) }}>
+                {model.render(item, column, index)}
+              </div>;
+            })}
+          </>;
+        }}
+        icon={model.icon ? ((item, index) => model.icon!(item, index)) : undefined}
```
dgozman, line 86: Why can't this and all others be `icon=model.icon` directly?

dgozman (approved):

packages/web/src/shared/resizeView.tsx
```diff
@@ -0,0 +1,99 @@
...
+      onPaneMouseUp={() => setResizing(null)}
+      onPaneMouseMove={event => {
+        if (!event.buttons) {
+          setResizing(null);
+        } else if (resizing) {
+          const delta = orientation === 'horizontal' ? event.clientX - resizing.clientX : event.clientY - resizing.clientY;
+          const newOffset = resizing.offset + delta;
+          const previous = resizing.index > 0 ? offsets[resizing.index - 1] : 0;
+          const next = orientation === 'horizontal' ? measure.width : measure.height;
+          const constrainedDelta = Math.min(Math.max(previous + minGap, newOffset), next - minGap) - offsets[resizing.index];
```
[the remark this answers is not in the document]
pavelfeldman: I actually intentionally allowed them to go beyond the fold

packages/web/src/components/splitView.tsx
pavelfeldman: I can try doing this in a follow-up

packages/web/src/components/gridView.tsx
pavelfeldman: I'm using it synchronously as a default value of the offsets state, so I guess I need to do it here.

packages/web/src/components/gridView.tsx
pavelfeldman: It can, previously it grid interface had a column parameter as well

PR #28929, opened 2024-01-10
fix(fetch): allow empty fileName in fetch API

### LINKED ISSUE #28811: [QUESTION] Playwright/Java: POST multipart with different content types
### System info
- Playwright Version: [v1.35]
- Other info: Java 17

### Question ###
Dear all, I have an endpoint which requires multipart/form-data with different content types. One of them is json payload, another one is an excel file.
How these types should be provided?
I've searched across documentation for java version, but there only an example without separate content-types is provided.
I've tried to write a test as in source code below but have different errors from server:

- when uncommenting `headers.put("Content-Type", "multipart/form-data"`, I receive "Internal server error: the request was rejected because no multipart boundary was found"
- if keeping it commented as advised in #20249, the server responds with "Internal server error: Content-Type 'application/octet-stream' is not supported"
- if replacing `data.set("file", Paths.get("src/test/resources/model_templates/ComponentsFromExcel.xlsx"))` with `data.set("file", new FilePayload("someFile.xlsx", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
            Files.readAllBytes(Paths.get("someFile.xlsx"))))` - still getting the same message with "Internal server error: Content-Type 'application/octet-stream' is not supported"

Looks like json part is accepted, but the problem is in the file part.
Please advise how to fix the issue and provide all required content-types correctly for such case.
Thanks in advance!

### Source code

```java
var playwright = Playwright.create();
final var headers = new HashMap<String, String>();
headers.put("accept", "application/json");
headers.put("authorization", "Basic dGVzdDpwYXNzd29yZA==");
// headers.put("Content-Type", "multipart/form-data");

var context = playwright.request().newContext(
    new APIRequest.NewContextOptions()
        .setBaseURL("http://some.url/")
        .setExtraHTTPHeaders(headers)
);

var requestOptions = RequestOptions.create();
var data = FormData.create()
    .set("textValue", "someString") //this one should be application/json
    .set("file", Paths.get("someFile.xlsx")); //this one should be excel file type

requestOptions.setMultipart(data); 
var response = context.post("endpointPath", requestOptions);
```

**Curl request/response example from this endpoint:**

```bash
curl -X 'POST' \
  'https://some/url' \
  -H 'accept: application/json' \
  -H 'Authorization: Basic ' \
  -H 'Content-Type: multipart/form-data' \
  -F 'textValue=someString;type=application/json' \
  -F 'file=@someFile.xlsx;type=application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/28811

### DIFF
--- a/docs/src/api/class-formdata.md
+++ b/docs/src/api/class-formdata.md
@@ -36,7 +36,7 @@ Field name.
 ### param: FormData.set.value
 * since: v1.18
 - `value` <[string]|[boolean]|[int]|[Path]|[Object]>
-  - `name` <[string]> File name
+  - `name` ?<[string]> File name
   - `mimeType` <[string]> File type
   - `buffer` <[Buffer]> File content
 
@@ -46,7 +46,7 @@ Field value.
 * since: v1.18
 * langs: csharp
 - `value` <[string]|[boolean]|[int]|[Object]>
-  - `name` <[string]> File name
+  - `name` ?<[string]> File name
   - `mimeType` <[string]> File type
   - `buffer` <[Buffer]> File content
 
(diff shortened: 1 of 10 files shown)


mxschmitt:
Closing for now, since in the browser this is not possible.

PR #28932, opened 2024-01-10
chore: validate deps during install

mxschmitt:
Motivation: On Windows we call around 50 times `PrintDeps.exe` which takes on a very fast machine 500+ms. On Linux we do it around 120 times (`ldd`) which takes around 150ms.

This change validates the dependencies once on browser install (`npx playwright install`). In case its failing, it will emit a warning, in case of a success, it will create a marker file that the binary has been validated. For future `launch()` calls, we'll read this file and if exists, we'll not validate again. Otherwise we'll validate again.

Note: If the marker file is older than 30 days, the browser will be validated again.


packages/playwright-core/src/cli/program.ts
```diff
@@ -159,6 +159,8 @@ program
         } else {
           const forceReinstall = hasNoArguments ? false : !!options.force;
           await registry.install(executables, forceReinstall);
+          for (const executable of executables)
+            await executable.validateHostRequirements(process.env.PW_LANG_NAME || 'javascript');
```
dgozman, line 163: - This now prints `Skipping host requirements validation logic because...` multiple times.
- `Cannot launch Firefox on Ubuntu 16.04! Minimum required Ubuntu version for Firefox browser is 20.04` error is now misplaced. We should leave it for the launch time.
- Which brings a question whether we'll get some false positives for unused WebKit installations.

mxschmitt: 1 addressed, 2 removed, since we don't do things like this for other browsers, and 16.04 is really outdated. For 3, I improved the error message like that we can clarify it with others, what they think:

```
root@d8d476073591:/work# npx playwright install
Failed to install browsers
Error: 
╔════════════════════════════════════════════════════════╗
║ Host system is missing dependencies to run browsers.   ║
║ Please install them with the following command:        ║
║                                                        ║
║     npx playwright install-deps                        ║
║                                                        ║
║ Alternatively, use apt:                                ║
║     apt-get install libnss3\                           ║
║         libnspr4\                                      ║
║         libdbus-1-3\                                   ║
║         libatk1.0-0\                                   ║
║         libatk-bridge2.0-0\                            ║
║         libcups2\                                      ║
║         libdrm2\                                       ║
║         libxkbcommon0\                                 ║
║         libatspi2.0-0\                                 ║
║         libxcomposite1\                                ║
║         libxdamage1\                                   ║
║         libxfixes3\                                    ║
║         libxrandr2\                                    ║
║         libgbm1\                                       ║
║         libasound2                                     ║
║                                                        ║
║ Or just install the browsers you actually want to run: ║
║                                                        ║
║     npx playwright install chromium                    ║
║                                                        ║
║ <3 Playwright Team                                     ║
╚════════════════════════════════════════════════════════╝
```

packages/playwright-core/src/server/registry/index.ts
```diff
@@ -859,6 +851,43 @@ export class Registry {
...
+
+  private async _validateHostRequirementsForExecutableIfNeeded(executable: Executable, sdkLanguage: string, updateMarkerFile: boolean) {
+    const kMaximumReValidationPeriod = 30 * 24 * 60 * 60 * 1000; // 30 days
+    // Executable does not require validation.
+    if (!executable.directory)
+      return;
+    const markerFile = path.join(executable.directory, 'DEPENDENCIES_VALIDATED');
+    const depsValidationFileExists: boolean = await fs.promises.access(markerFile).then(() => true).catch(() => false);
+    // Executable is already validated.
+    if (depsValidationFileExists && await fs.promises.stat(markerFile).then(stat => (Date.now() - stat.mtime.getTime()) < kMaximumReValidationPeriod))
```
dgozman, line 871: I think we can skip the `fs.access` call, and only do the `fs.stat` call.

packages/playwright-core/src/server/registry/index.ts
```diff
@@ -859,6 +851,43 @@ export class Registry {
...
+    } catch (error) {
+      debugLogger.log('install', `validation failed for ${executable.name}`);
+      throw error;
+    }
+
+    if (updateMarkerFile) {
+      if (depsValidationFileExists)
+        await fs.promises.utimes(markerFile, new Date(), new Date());
+      else
+        await fs.promises.writeFile(markerFile, '');
```
dgozman, line 887: Why not just `writeFile` unconditionally?

mxschmitt: Addressed.

packages/playwright-core/src/server/registry/dependencies.ts
```diff
@@ -257,6 +257,10 @@ export async function validateDependenciesLinux(sdkLanguage: string, linuxLddDir
       `Alternatively, use apt:`,
       `    ${maybeSudo}apt-get install ${[...missingPackages].join('\\\n        ')}`,
       ``,
+      `Or just install the browsers you actually want to run:`,
```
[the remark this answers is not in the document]
mxschmitt: Removed.

packages/playwright-core/src/server/registry/index.ts
mxschmitt: Good point

packages/playwright-core/src/server/browserType.ts
```diff
@@ -180,7 +180,7 @@ export abstract class BrowserType extends SdkObject {
       if (!registryExecutable || registryExecutable.browserName !== this._name)
         throw new Error(`Unsupported ${this._name} channel "${options.channel}"`);
       executable = registryExecutable.executablePathOrDie(this.attribution.playwright.options.sdkLanguage);
-      await registryExecutable.validateHostRequirements(this.attribution.playwright.options.sdkLanguage);
+      await registry.validateHostRequirementsForExecutablesIfNeeded([registryExecutable], this.attribution.playwright.options.sdkLanguage, /* updateValidationTimestamp */ false);
```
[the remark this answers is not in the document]
mxschmitt: I think thats how we discussed it in the meeting according to the notes:

> - add warning during npx playwright install
> - add verify label during validation, otherwise validate again in some flag file.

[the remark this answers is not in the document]
mxschmitt: Addressed.

dgozman (approved):

Issue #28758, opened 2023-12-21
[Feature] propagate headers to redirects in page.continue


dvoytenko:
The headers passed in `page.continue({headers: ...})` do not propagate to redirects. This is very clearly documented in https://playwright.dev/docs/api/class-route#route-fetch. However, often it's needed for navigation requests to propagate headers to redirects. In many cases it's an expected default. The ask here is to add a flag to the `page.continue()` to propagate headers to redirects. For instance:

```
page.continue({
  headers: {
    'Test-Metadata': testMetadata,
  },
  propagetHeadersToRedirects: true,
});
```

PR #28962, opened 2024-01-12
fix(chromium): preserve header overrides in redirects

yury-s:
Reference https://github.com/microsoft/playwright/issues/28758

### DIFF
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -285,10 +285,16 @@ export class CRNetworkManager {
     let route = null;
     if (requestPausedEvent) {
       // We do not support intercepting redirects.
-      if (redirectedFrom || (!this._userRequestInterceptionEnabled && this._protocolRequestInterceptionEnabled))
-        this._session._sendMayFail('Fetch.continueRequest', { requestId: requestPausedEvent.requestId });
-      else
+      if (redirectedFrom || (!this._userRequestInterceptionEnabled && this._protocolRequestInterceptionEnabled)) {
+        let headers = undefined;
+        const previousHeaderOverrides = redirectedFrom?._latestRoute?._headerOverrides;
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -508,6 +514,7 @@ class InterceptableRequest {
   readonly _timestamp: number;
   readonly _wallTime: number;
   readonly _route: RouteImpl | null;
+  readonly _latestRoute: RouteImpl | undefined;
```
dgozman, line 517: Can we walk the redirect chain on the call site instead? If not, let's call this `originalRequestRoute` since we know it only exists on the first (non-redirect) request.

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -549,6 +557,7 @@ class RouteImpl implements network.RouteDelegate {
   private readonly _session: CRSession;
   private _interceptionId: string;
   _alreadyContinuedParams: Protocol.Fetch.continueRequestParameters | undefined;
+  _headerOverrides: types.HeadersArray | undefined;
```
dgozman, line 560: Can we reuse `_alreadyContinuedParams`?

dgozman (approved):

packages/playwright-core/src/server/chromium/crNetworkManager.ts
yury-s: Done. We don't reset it now though.

packages/playwright-core/src/server/chromium/crNetworkManager.ts
yury-s: We do not store reference to the previous request in InterceptableRequest, neither do we store a reference from Request to InterceptableRequest.

dgozman (approved):

PR #28968, opened 2024-01-12
fix(driver): add external commands only for JS binding

mxschmitt:
Motivation: Before this change if a language binding invoked the CLI with `--help` or without any arguments (which will also show the help text) it was suggesting that test/merge-reports/show-report is something we support. After this change this does not get shown anymore.

### DIFF
--- a/packages/playwright-core/src/cli/cli.ts
+++ b/packages/playwright-core/src/cli/cli.ts
@@ -53,13 +53,18 @@ const kExternalPlaywrightTestCommands = [
   ['show-report', 'Show Playwright Test HTML report.'],
   ['merge-reports', 'Merge Playwright Test Blob reports'],
 ];
-for (const [command, description] of kExternalPlaywrightTestCommands) {
-  const playwrightTest = program.command(command).allowUnknownOption(true);
-  playwrightTest.description(`${description} Available in @playwright/test package.`);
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/cli/cli.ts
```diff
@@ -53,13 +53,18 @@ const kExternalPlaywrightTestCommands = [
...
+    const playwrightTest = program.command(command).allowUnknownOption(true);
+    playwrightTest.description(`${description} Available in @playwright/test package.`);
+    playwrightTest.action(async () => {
+      printPlaywrightTestError(command);
+      gracefullyProcessExitDoNotHang(1);
+    });
+  }
 }
 
+if (!process.env.PW_LANG_NAME)
```
yury-s, line 67: `|| === 'javascript'` - just in case we start running the driver out of process?

yury-s (approved):

packages/playwright-core/src/cli/cli.ts
mxschmitt: So far we use 'not set' as a good default.

yury-s: Yeah, whoever touches it will never know that this place need to be updated too.

PR #28970, opened 2024-01-12
Clarify how vscode plugin feature "Record at Cursor" works.

tomaskikutis:
After clicking "Record at Cursor" I expected that recording window would start, execute the test until the cursor and allow me to continue recording with my application being at the same state as if I was debugging the test and had a breakpoint on that line. I've clarified in the docs that it doesn't work like that and added instructions on how to achieve this using an alternative approach.

### DIFF
--- a/docs/src/codegen.md
+++ b/docs/src/codegen.md
@@ -44,7 +44,7 @@ Once you are done recording click the **cancel** button or close the browser win
 
 To record from a specific point in your test move your cursor to where you want to record more actions and then click the **Record at cursor** button from the Testing sidebar.
 
-**Important:** This feature does not start the test and execute the lines before the cursor. It only attaches to an existing VSCode web view and gets it into recording mode. You have to get VSCode web view into a correct state manually. The easiest way to do this is by adding a breakpoint where you want to continue recording, starting the test in debugging mode and exiting the debugger after it stops at your breakpoint. You have to have 'Show browser' checked in the sidebar in order for VSCode web view to remain open after exiting the debugger.
+**Important:** This feature does not start the test and execute the lines before the cursor. It only attaches to an existing VSCode-launched browser instance and gets it into recording mode. You have to get the browser into a correct state manually before starting to record. The easiest way to do this is by adding a breakpoint where you want to continue recording, starting the test in debugging mode and stopping the debugger after it pauses at you breakpoint. After stopping debugger you can start recording the test at cursor. You have to have 'Show browser' checked in the sidebar in order for VSCode-launched browser to remain open after exiting the debugger.
 
 ![record at cursor in vs code](https://github.com/microsoft/playwright/assets/13063165/77948ab8-92a2-435f-9833-0944da5ae664)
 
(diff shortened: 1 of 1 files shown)


docs/src/codegen.md
```diff
@@ -42,8 +42,9 @@ Once you are done recording click the **cancel** button or close the browser win
 
 ### Record at Cursor
 
-To record from a specific point in your test move your cursor to where you want to record more actions and then click the **Record at cursor** button from the Testing sidebar. If your browser window is not already open then first run the test with 'Show browser' checked and then click the **Record at cursor** button.
+To record from a specific point in your test move your cursor to where you want to record more actions and then click the **Record at cursor** button from the Testing sidebar.
 
+**Important:** This feature does not start the test and execute the lines before the cursor. It only attaches to an existing VSCode web view and gets it into recording mode. You have to get VSCode web view into a correct state manually. The easiest way to do this is by adding a breakpoint where you want to continue recording, starting the test in debugging mode and exiting the debugger after it stops at your breakpoint. You have to have 'Show browser' checked in the sidebar in order for VSCode web view to remain open after exiting the debugger.
```
yury-s, line 47: I guess by "VSCode web view" you mean the browser?

> The easiest way to do this is by adding a breakpoint where you want to continue recording, starting the test in debugging mode and exiting the debugger after it stops at your breakpoint.

This is misleading, we cannot record actions under the debugger and after resuming the execution playwright will continue to run the test till the end.

I don't think the new explanation is clearer then the existing one which describes two exact steps that one needs to perform.

yury-s (changes requested):

docs/src/codegen.md
tomaskikutis: > I guess by "VSCode web view" you mean the browser?

Kind of. I think VS code web view is a more accurate accurate term, because the browser that one launches manually wouldn't work. It has vs code icon in the OS taskbar so it seems it's a vs code window that has a browser embedded. I'm fine adjusting that. I could call it "vs code lauched browser" to distinguish from user launched instances.

> This is misleading, we cannot record actions under the debugger and after resuming the execution playwright will continue to run the test till the end.

I'll adjust instructions so it is clear that I'm not suggesting to debug and record the test at the same time. What I was trying to say is that **before** starting to record the test - you put the breakpoint, debug the test and when it breaks at that line, you stop the debugger - **not** resume execution. It does not continue runing the test after the debugger is stopped. When the debugger is stopped, you would start recording the test at cursor.

tomaskikutis:
> When I want to generate in the middle of the test I either comment out the tail or insert return; statement.

That's a good point. Nowadays I rarely do the debug procedure either. It was frustrating though as a new user to discover that the feature doesn't work the way I'd expect - that is run the test until the cursor and only then start recording. In retrospect I wouldn't even want it now since it would slow things down if I had to wait every time I wanted to add something. I think naming/docs are still confusing though.

pavelfeldman:
Closing as per above.

PR #28986, opened 2024-01-14
fix(ct): move import list into the compilation cache data

pavelfeldman:


packages/playwright/src/transform/compilationCache.ts
```diff
@@ -105,27 +113,30 @@ export function getFromCompilationCache(filename: string, hash: string, moduleUr
...
   return {
-    addToCache: (code: string, map: any) => {
+    addToCache: (code: string, map: any | undefined | null, data: Map<string, any>) => {
       if (isWorkerProcess())
         return;
       fs.mkdirSync(path.dirname(cachePath), { recursive: true });
       if (map)
         fs.writeFileSync(sourceMapPath, JSON.stringify(map), 'utf8');
+      if (data.size)
+        fs.writeFileSync(dataPath, JSON.stringify(Object.fromEntries(data.entries()), undefined, 2), 'utf8');
```
dgozman, line 132: I'm curios why pretty-print?

packages/playwright-ct-core/src/tsxTransform.ts
```diff
@@ -108,56 +126,30 @@ export default declare((api: BabelAPI) => {
...
-
       // Treat JSX-everything as component usages.
       if (t.isJSXElement(p.node)) {
         if (t.isJSXIdentifier(p.node.openingElement.name))
           names.add(p.node.openingElement.name.name);
         if (t.isJSXMemberExpression(p.node.openingElement.name) && t.isJSXIdentifier(p.node.openingElement.name.object) && t.isJSXIdentifier(p.node.openingElement.name.property))
           names.add(p.node.openingElement.name.object.name);
       }
-
-      // Treat mount(identifier, ...) as component usage if it is in the importedLocalNames list.
```
dgozman, line 138: What happened to this logic?

packages/playwright-ct-core/src/vitePlugin.ts
```diff
@@ -267,35 +257,13 @@ async function checkSources(buildInfo: BuildInfo): Promise<boolean> {
...
 }
 
-async function checkNewTests(suite: Suite, buildInfo: BuildInfo, componentRegistry: ComponentRegistry): Promise<boolean> {
-  const testFiles = new Set<string>();
-  for (const project of suite.suites) {
-    for (const file of project.suites)
-      testFiles.add(file.location!.file);
+async function populateComponentsFromTests(componentRegistry: ComponentRegistry, componentsByTestFile: Map<string, string[]>) {
+  const importInfos: Map<string, ImportInfo[]> = await getUserData('playwright-ct-core');
+  for (const [testFile, importList] of importInfos) {
```
dgozman, line 262: IIUC, this is not necessary the `testFile`, could be any imported dependency of the test file? Let's be more precise with the naming to avoid confusion in the future.

packages/playwright-ct-core/src/tsxTransform.ts
```diff
@@ -71,19 +73,35 @@ export default declare((api: BabelAPI) => {
...
+              continue;
+            const { localName, info } = importInfo(importNode, specifier, this.filename!);
+            importInfos.set(localName, info);
+          }
+          p.skip();
+          p.remove();
+          return;
+        }
+
+        // Convert JS imports into refs only if they are used by JSX.
```
dgozman, line 91: It took me awhile to understand this comment. I'd recommend `Convert JS imports that are used as components in JSX expressions into refs`.

packages/playwright-ct-core/src/vitePlugin.ts
```diff
@@ -198,20 +189,22 @@ export function createPlugin(
...
-        for (const componentName of testInfo.components) {
-          const component = componentRegistry.get(componentName);
-          component?.deps.forEach(d => deps.add(d));
+        for (const componentFile of componentFiles) {
+          const componentDeps = depsCollector.get(componentFile) || [];
+          for (const d of componentDeps)
+            deps.add(d);
         }
-        setExternalDependencies(filename, [...deps]);
+        setExternalDependencies(testFile, [...deps]);
```
dgozman, line 207: I guess I am not following something, but to me it seems that `testFile` here could be a dependency, and I'm not really sure whether `setExternalDependencies()` for a non-test file ever has any effect on the system.

pavelfeldman: Renaming, updating!

packages/playwright-ct-core/src/vitePlugin.ts
pavelfeldman: Same as above!

packages/playwright/src/transform/compilationCache.ts
pavelfeldman: For readability!

dgozman (approved):

PR #28434, opened 2023-11-30
docs(docker): add note on how to create your own

mxschmitt:
https://github.com/microsoft/playwright/issues/28383

PR #28994, opened 2024-01-15
docs(docker): various enhancements

### LINKED ISSUE #2241: [Feature] Integrations Docker Doc improve
When i reading Integrations Docker Doc,  I don't have a Smooth experience.
  https://playwright.dev/python/docs/docker
  beacuse after this change: https://github.com/microsoft/playwright-python/pull/1921.
  the docker image have no global installed Playwright library anymore.
  I must install  Playwright  by myself ,but Integrations Docker Doc does not reflect this.
  It took me a while to understand this.
  so,i hope to improve  Integrations Docker Doc as above.

thanks for your dedication.

mxschmitt:
Fixes https://github.com/microsoft/playwright-python/issues/2241

### DIFF
--- a/docs/src/docker.md
+++ b/docs/src/docker.md
@@ -5,7 +5,7 @@ title: "Docker"
 
 ## Introduction
 
-[Dockerfile.jammy] can be used to run Playwright scripts in Docker environment. These image includes all the dependencies needed to run browsers in a Docker container, and also include the browsers themselves.
+[Dockerfile.jammy] can be used to run Playwright scripts in Docker environment. These image includes the [Playwright browsers](./browsers.md#install-browsers) and [browser system dependencies](./browsers.md#install-system-dependencies). The Playwright package/dependency is not included in the image and should be installed separately.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


docs/src/docker.md
```diff
@@ -165,3 +164,18 @@ The image will be tagged as `playwright:localbuild-jammy` and could be run as:
...
+To run Playwright inside Docker, you need to have Python, [Playwright browsers](./browsers.md#install-browsers) and [browser system dependencies](./browsers.md#install-system-dependencies) installed. See the following Dockerfile:
+
+```Dockerfile
+FROM python:3.12-bookworm
+
+RUN pip install playwright==@%%VERSION%% && \
+    playwright install --with-deps
+```
+
+Note: official images published to [Microsoft Artifact Registry] are built using [`//utils/docker/build.sh`](https://github.com/microsoft/playwright-python/blob/main/utils/docker/build.sh) script.
```
yury-s, line 181: This is our implementation details, we shouldn't be advertising it. Those who curious can figure it out by looking at the sources already but I would not recommended to general clients.

docs/src/docker.md
```diff
@@ -165,3 +164,18 @@ The image will be tagged as `playwright:localbuild-jammy` and could be run as:
 docker run --rm -it playwright:localbuild /bin/bash
 ```
 
+## Development
+* langs: python
+
+### Build the image
```
yury-s, line 170: Why do we need this section? Our main recommendation is our image, other enthusiasts can figure it out themselves.

mxschmitt: we have a similar section, especially the:

```Dockerfile
FROM python:3.12-bookworm

RUN pip install playwright==@%%VERSION%% && \
    playwright install --with-deps
```

part for Node.js since https://github.com/microsoft/playwright/pull/28434 - for Python I thought it makes sense as well, since users often create their own / have their own image.

docs/src/docker.md
mxschmitt: Happy to remove this in JS + Python.

yury-s: Let's remove.

docs/src/docker.md
yury-s: My question is about the section in general regardless the language. This will just confuse people new to docker.

yury-s (approved):

PR #29034, opened 2024-01-18
fix docs typo for stylePath

romanimm:
apparently styleFile was renamed to stylePath before the release. 

btw. exactly what I need. 🎉


mxschmitt:
Thanks!

Issue #29039, opened 2024-01-18
[BUG] No quotes around show-trace path when test fails and path has spaces


guido-visser:
### System info
- Playwright Version: [v1.41.0]
- Operating System: [Windows 11]
- Browser: [Chromium]
- Other info:

**Steps**
- Make sure that you have a test in a folder that contains spaces
- Let the test fail

**Expected**

Being able to copy the message and paste it in a console. It should be: `npx playwright show-trace "test-results\page-01 - appointments-Create-chromium\trace.zip"`

**Actual**
 `npx playwright show-trace test-results\page-01 - appointments-Create-chromium\trace.zip`
The message doesn't have quotes around the path, causing it to give an error when you copy and paste the message

PR #29057, opened 2024-01-18
chore: add quotes to the path only when necessary

yury-s:
This is a follow-up to c76f5294cea1b98fc667b677d042d28d3e4b5bb7
Reference https://github.com/microsoft/playwright/issues/29039


packages/playwright/src/reporters/base.ts
```diff
@@ -373,6 +373,12 @@ export function formatFailure(config: FullConfig, test: TestCase, options: {inde
   };
 }
 
+function quoteIfContainsWhitespace(path: string): string {
+  if (/\s/.test(path))
```
mxschmitt, line 377: maybe test for the opposite? I feel like brackets would hurt us as well.

yury-s: Done. In that case we may also need to escape some characters. Let's wait if it bothers anyone.

dgozman (approved):

Issue #29067, opened 2024-01-19
Codegen/Recorder: not all clicks are being actioned nor recorded


Thomas-Nagel:
Steps to reproduce
1)    start “record new” in VS Code or codegen via cli npx playwright codegen
2)    navigate to google.com (it opens chromium on my machine)
3)    click on accept or reject cookies button
-> the click is not actioned - and not recorded

A click e.g. on “More options” is actioned and recorded.

The locators are correctly shown, adding a check for "visibility" and other options via the "locator toolbar" is working fine

Created code:

import { test, expect } from '@playwright/test';

test('test', async ({ page }) => {
  await page.goto('https://www.google.com/');
});

Setup:
Playwright typescript 1.41 (as far as I can recall, it was no issue with 1.40.1)
Chromium Version 121.0.6167.57 (Developer Build) (64-bit)
MS Windows 10 and 11

mxschmitt:
I can repro.

Bisected to https://github.com/microsoft/playwright/compare/214b3733b2dde6bdf51fa8994dd5c813d17d0fce...7bffff5790e28243a815c985135e908247b563db.

Thomas-Nagel:
After the upgrade to Version 1.41.1, the issue got sorted. 
Confirmation from our side that the sample used for reproducing the issue as well our real application is fixed.

ruifigueira:
This issue was caused by a contribution of mine, sorry about that 😔

The reason why it fails with google buttons is because they have a nested div, something like:

```html
<button>
  <div role="none">Accept all</div>
</button>
```

The recorder code ends up having the div as the hovered element, but its click event target is the button. In my change, recorder only handled clicks if hovered element and event target matched (this was my attempt to fix a problem with webkit and slider inputs, as described [here](https://github.com/microsoft/playwright/pull/28767#issuecomment-1868099687))

I'll add a regression test to ensure buttons like the ones at google page are recorded properly.

PR #29087, opened 2024-01-20
feat(codegen): slider support

ruifigueira:
This PR adds support for range slider inputs.

It handles the special case of webkit, which triggers a click event with the body element as target when sliding the input element. For that reason, I skip the click handling event if the hovered element is a range input, without consuming it.

In my first approach (#28767) I was naive and assumed that click events should have the same target element as the current hovered element, and that caused the regression reported in #29067. Sorry about that 😔

### DIFF
--- a/packages/playwright-core/src/server/injected/recorder/recorder.ts
+++ b/packages/playwright-core/src/server/injected/recorder/recorder.ts
@@ -225,6 +225,10 @@ class RecordActionTool implements RecorderTool {
   }
 
   onClick(event: MouseEvent) {
+    // in webkit, sliding a range element may trigger a click event with a different target if the mouse is released outside the element bounding box.
+    // So we check the hovered element instead, and if it is a range input, we skip click handling
+    if (isRangeInput(this._hoveredElement))
+      return;
     if (this._shouldIgnoreMouseEvent(event))
       return;
     if (this._actionInProgress(event))
@@ -313,6 +317,17 @@ class RecordActionTool implements RecorderTool {
         selector: this._activeModel!.selector,
         signals: [],
         files: [...((target as HTMLInputElement).files || [])].map(file => file.name),
+      });
+      return;
+    }
+
+    if (isRangeInput(target)) {
+      this._recorder.delegate.recordAction?.({
+        name: 'fill',
+        // must use hoveredModel instead of activeModel for it to work in webkit
+        selector: this._hoveredModel!.selector,
+        signals: [],
+        text: target.value,
       });
       return;
     }
@@ -414,7 +429,7 @@ class RecordActionTool implements RecorderTool {
     const nodeName = target.nodeName;
     if (nodeName === 'SELECT' || nodeName === 'OPTION')
       return true;
-    if (nodeName === 'INPUT' && ['date'].includes((target as HTMLInputElement).type))
+    if (nodeName === 'INPUT' && ['date', 'range'].includes((target as HTMLInputElement).type))
       return true;
     return false;
   }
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/injected/recorder/recorder.ts
```diff
@@ -271,6 +275,15 @@ class RecordActionTool implements RecorderTool {
...
     }
 
+    if (isRangeInput(target)) {
+      this._recorder.delegate.recordAction?.({
+        name: 'fill',
+        selector: this._hoveredModel!.selector,
+        signals: [],
+        text: target.value,
+      });
+    }
```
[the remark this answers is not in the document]
ruifigueira: It definitely needs a `return` there. It only records one instruction but it's just because it's not commiting it, but it is calling the `recordAction` function twice, yes.

I'll do that change and rebase it.

packages/playwright-core/src/server/injected/recorder/recorder.ts
```diff
@@ -225,6 +225,10 @@ class RecordActionTool implements RecorderTool {
   }
 
   onClick(event: MouseEvent) {
+    // in webkit, when clicking a range input, event target is the body element and not the input itself.
```
pavelfeldman, line 228: WebKit gives me <input> target, what do I do wrong?

Screenshot 2024-01-30 at 11 42 06 AM

ruifigueira: Apparently, it's my test `dragSlider` action that's causing the `body` click event.

Here is a test to check the target element of the click event:

```ts
  test('click range input', async ({ page }) => {
    await page.setContent(`<input type="range">`);
    await page.evaluate(() => {
      document.addEventListener('click', ({ target }) => console.log((target as HTMLElement).tagName), true);
    });
    const dragSlider = async () => {
      const { x, y, width, height } = await page.locator('input').boundingBox();
      await page.mouse.move(x + width / 2, y + height / 2);
      await page.mouse.down();
      await page.mouse.move(x + width, y + height / 2);
      await page.mouse.up();
    };

    const [msg] = await Promise.all([
      page.waitForEvent('console'),
      dragSlider(),
    ]);
    expect(msg.text()).toBe('INPUT');
  });
```

In chrome and firefox, the click event is triggered on the input element, in webkit it is on the body. I'm running those tests is windows, btw.

ruifigueira: The `click` event is triggered in the last hovered element, and because I'm moving to `x + width`, it's already in the border of the element and webkit already considers it the `body` element.

In my test above, it passes if mouse is released one pixel before, basically replacing:

```ts
await page.mouse.move(x + width, y + height / 2);
```

with:

```ts
await page.mouse.move(x + width - 1, y + height / 2);
```

But ir's normal that real users slide after the element width, so I think that the check is necessary.

I tried to record it manually without the ` isRangeInput(this._hoveredElement)` check, by running:

 ```
npx playwright codegen -b webkit 'data:text/html,<input type="range" min="0" max="10" value="5">' 
```

I get the following recorded lines:

```ts
  // if I restrain my mouse movements
  await page.getByRole('slider').fill('10');

  // if I move past the element (and the slider at the maximum value, that is, 10)
  await page.getByRole('slider').fill('5');
  await page.getByRole('slider').click();
```

pavelfeldman: I see what you are saying. We are lucky that webkit does not dispatch the mouse moves as you drag the slider though. If it did, it would reset your accounting as hovered element would become body (or some other element). Also your comment is not entirely accurate - if you are clicking the range element, you are getting a click on input. It is only if you drag to outside you are getting a body target. Let's document this aspect in the comment.

pavelfeldman (approved):

PR #29113, opened 2024-01-23
chore: speedup frequent element text normalization

dgozman:
We cache `ElementText` for frequent operations, but then call `normalizeWhitespace` on it every time which burns a lot of CPU.

### DIFF
--- a/packages/playwright-core/src/server/injected/selectorGenerator.ts
+++ b/packages/playwright-core/src/server/injected/selectorGenerator.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/injected/selectorGenerator.ts
```diff
@@ -237,7 +237,7 @@ function buildNoTextCandidates(injectedScript: InjectedScript, element: Element,
 
   const labels = getElementLabels(injectedScript._evaluator._cacheText, element);
   for (const label of labels) {
-    const labelText = label.full.trim();
```
yury-s, line 240: Is that expected that it is now normalized?

yury-s (approved):

packages/playwright-core/src/server/injected/selectorGenerator.ts
dgozman: Yep.

PR #29176, opened 2024-01-25
chore: automatically detect the dev server

pavelfeldman:


yury-s (approved):

CreativeTechGuy:
@Pharb you just beat me to it. 😛  This took a while for us to figure out why our tests were failing to start and it wasn't until I added logs for all requests on our server that I realized this stealthily changed from GET to HEAD.

It was definitely breaking for us. We use a plain NodeJS server in development which we added a API route which only responds to GET requests specifically to use in Playwright to check if it's online. It's an easy fix to let it respond to HEAD requests too, but would have been really nice to know this was needed.

Pharb:
I opened https://github.com/microsoft/playwright/pull/29176 to track it as a regression in v1.42.0

JakeHedman:
> I opened #29176 to track it as a regression in v1.42.0

#29732 *

PR #29177, opened 2024-01-25
fix: make TeleSuite.project work for all suites

### LINKED ISSUE #29173: [BUG] The `project()` function on the `Suite` object doesn't work the same with report merge workflow
### System info

- Playwright Version: v1.41.1
- Operating System: All
- Browser: All

### Source code

Source code is available at https://github.com/devpow112/playwright-bugs.

**Steps**

 - Check out code
 - `npm i`
 - `npm run normal-project-fn`
 - observe no issues and a bunch of logs of `'project' is defined`
 - `npm run issue-project-fn`
 - observe error is emitted due to `project` result being `undefined` after call to `project()`
   - `Error in reporter Error: 'project' is not defined`

**Expected**

This works without emitting any errors in both a normal execution and the `merge-reports` workflow.

**Actual**

Emits an error when using the `merge-reports` workflow due to the `project()` return `undefined` instead of the correct project.

### Notes

My use case here is that I have a custom reporter that wants to know this information for tracking purposes. This appears to be an issue related to how the [`project()` function is defined as part of the `TeleReporterReceiver` code](https://github.com/microsoft/playwright/blob/main/packages/playwright/src/isomorphic/teleReceiver.ts#L213). Seems to only be attaching the `project()` function that actually returns the project to root suites. When you call it nothing is returned unless you call it on the root suite. This can be worked around by getting the root suite always before trying to call `project()`. Ideally calling `project()` would work the same in both cases instead of requiring a workaround.

**Example**

```js
const getRootParent = (test) => {
	if (test.parent && test.parent.parent) {
		return getRootParent(test.parent);
	}

	return test;
};

export default class Reporter {
	onTestEnd(test) {
		const rootParent = getRootParent(test);
		const project = rootParent.project(); // will get the project in both cases

		if (project === undefined) {
			throw new Error('\'project\' is not defined');
		} else {
			console.log('\'project\' is defined');
		}
	}
}
```

The [documentation for `Suite`](https://playwright.dev/docs/api/class-suite#suite-project) implies `project()` can be called on any `Suite` object and will return the project configuration the suite belongs to. Either way the behaviour is not consistent.

yury-s:
Fixes https://github.com/microsoft/playwright/issues/29173

### DIFF
--- a/packages/playwright/src/isomorphic/teleReceiver.ts
+++ b/packages/playwright/src/isomorphic/teleReceiver.ts
@@ -206,10 +206,11 @@ export class TeleReporterReceiver {
     let projectSuite = this._rootSuite.suites.find(suite => suite.project()!.__projectId === project.id);
     if (!projectSuite) {
       projectSuite = new TeleSuite(project.name, 'project');
-      projectSuite._project = this._parseProject(project);
       this._rootSuite.suites.push(projectSuite);
       projectSuite.parent = this._rootSuite;
     }
+    // Always update project in watch mode.
+    projectSuite._project = this._parseProject(project);
     this._mergeSuitesInto(project.suites, projectSuite);
 
     // Remove deleted tests when listing. Empty suites will be auto-filtered
--- a/packages/playwright/src/isomorphic/teleReceiver.ts
+++ b/packages/playwright/src/isomorphic/teleReceiver.ts
@@ -206,11 +206,10 @@ export class TeleReporterReceiver {
     let projectSuite = this._rootSuite.suites.find(suite => suite.project()!.__projectId === project.id);
     if (!projectSuite) {
       projectSuite = new TeleSuite(project.name, 'project');
+      projectSuite._project = this._parseProject(project);
       this._rootSuite.suites.push(projectSuite);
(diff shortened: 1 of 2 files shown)


packages/playwright/src/isomorphic/teleReceiver.ts
```diff
@@ -206,11 +206,10 @@ export class TeleReporterReceiver {
     let projectSuite = this._rootSuite.suites.find(suite => suite.project()!.__projectId === project.id);
     if (!projectSuite) {
       projectSuite = new TeleSuite(project.name, 'project');
+      projectSuite._project = this._parseProject(project);
       this._rootSuite.suites.push(projectSuite);
       projectSuite.parent = this._rootSuite;
     }
-    const p = this._parseProject(project);
-    projectSuite.project = () => p;
```
dgozman, line 213: I think this line was updating `project` getter to return the new project for the same suite after reload (e.g. on watch).

yury-s: oh, ui mode! Fixed.

dgozman (approved):

PR #28539, opened 2023-12-07
fix: throw if fuliflled with unsupported status code

yury-s:
If request gets cancelled by the page before we fulfill it, we receive `loadingFailed` event. In that case we'll ignore interception error if any, otherwise the error will be propagated to the caller.

Fixes https://github.com/microsoft/playwright/issues/28490

PR #29180, opened 2024-01-25
fix: interception id not found error in route.continue

### LINKED ISSUE #29123: [BUG] route.continue: Protocol error (Fetch.continueRequest): Invalid InterceptionId.
### System info

- Playwright Version: v1.41.1
- Operating System: Ubuntu 20
- Browser: Chrome
- Other info: Using `mcr.microsoft.com/playwright:v1.41.1-focal`

### Source code

```js
await page.route('**/*', route => {
    const url = new URL(route.request().url());
    return domainsToBlock.has(url.hostname)
      ? route.abort()
      : route.continue()
//        .catch(e => {
//          console.error('Error while continuing route', url, e);
//        });
  });
```

**Steps**

Unfortunately, I don't have a great reproduction here -- we run this blocking interception all our tests. We're seeing enough failures as below though that it is pretty consistently resulting in our test suite failing.

**Expected**

Historically have never run into issues with this blocking, it aborts requests for domains we've blocked and otherwise just passes all other requests through without issue.

**Actual**

However, since updating to 1.41.1 (from 1.40.1) we're seeing rather random `route.continue: Protocol error (Fetch.continueRequest): Invalid InterceptionId.` being thrown for what appear to be legitimate requests that shouldn't be failing.

yury-s:
We stopped catching all exceptions in https://github.com/microsoft/playwright/pull/28539 in hope that we'll get loadingFailed even before Fetch.continue/fulfill command's error. Turns out this is racy and may fail if the test cancels the request while we are continuing it. The following test could in theory reproduce it if stars align and the timing is good:

```js
it('page.continue on canceled request', async ({ page }) => {
  let resolveRoute;
  const routePromise = new Promise<Route>(f => resolveRoute = f);
  await page.route('http://test.com/x', resolveRoute);

  const evalPromise = page.evaluate(async () => {
    const abortController = new AbortController();
    (window as any).abortController = abortController;
    return fetch('http://test.com/x', { signal: abortController.signal }).catch(e => 'cancelled');
  });
  const route = await routePromise;
  void page.evaluate(() => (window as any).abortController.abort());
  await new Promise(f => setTimeout(f, 10));
  await route.continue();
  const req = await evalPromise;
  expect(req).toBe('cancelled');
});
```

Fixes https://github.com/microsoft/playwright/issues/29123

### DIFF
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -28,6 +28,7 @@ import type * as types from '../types';
 import type { CRPage } from './crPage';
 import { assert, headersObjectToArray } from '../../utils';
 import type { CRServiceWorker } from './crServiceWorker';
+import { isProtocolError } from '../protocolError';
 
 type SessionInfo = {
   session: CRSession;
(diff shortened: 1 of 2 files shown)


dgozman (approved):

purepear:
Hi guys, thanks for taking care of this issue! 🙇 

Do you know if this will be released as a patch 1.41.2 or minor 1.42.0 release? (we started noticing some flakiness and we not sure if we should wait for 1.41.2 or downgrade until 1.42 is out)

yury-s:
> Hi guys, thanks for taking care of this issue! 🙇
> 
> Do you know if this will be released as a patch 1.41.2 or minor 1.42.0 release? (we started noticing some flakiness and we not sure if we should wait for 1.41.2 or downgrade until 1.42 is out)

We'll [cherry-pick](https://github.com/microsoft/playwright/pull/29222) the fix and push it in 1.41.2.

Issue #28798, opened 2023-12-27
[BUG] `determineFileExtension` / `toMatchSnapshot` does not detect JPEG in EXIF format


swissspidy:
### System info
- Playwright Version: [v1.XX]
- Operating System: [All, Windows 11, Ubuntu 20, macOS 13.2, etc.]
- Browser: [All, Chromium, Firefox, WebKit]
- Other info:

### Source code

Relevant code:

https://github.com/microsoft/playwright/blob/2d5488d8e31dcd3f04dbefc9e7d4f1c4b25fc8bc/packages/playwright/src/matchers/toMatchSnapshot.ts#L470-L471

`0xff, 0xd8, 0xff, 0xe0, 0x00, 0x10, 0x4a, 0x46, 0x49, 0x46, 0x00, 0x01, 0x01` is just one of the possible file signatures for JPEG.

See https://en.wikipedia.org/wiki/List_of_file_signatures for a list of more.

In my case, the sequence is `ff d8 ff e1 00 bc 45 78 69 66 00 00 49` instead.

Because `determineFileExtension` does not recognize this byte sequence, the jpeg is falsely saved with a `.dat` extension.

A workaround is to provide a custom snapshot name.

**Steps**

example image:

![demo-file](https://github.com/microsoft/playwright/assets/841956/ab5eb5b3-93a2-4129-84f9-42c7209fb47f)

Run `expect( /* get above jpeg as buffer here */ ).toMatchSnapshot()`

**Expected**

File should be saved as `.jpg`

**Actual**

File is saved as `.dat` because of unrecognized format

pavelfeldman:
Do you want to send a PR?

tanayv:
Hey @pavelfeldman,
I have a PR to fix this issue: https://github.com/microsoft/playwright/pull/29208, would be great if you could take a look, thanks!

mxschmitt:
Fixed in https://github.com/microsoft/playwright/pull/29208.

PR #29208, opened 2024-01-27
fix: Add support for exif jpeg file formats

tanayv:
- Hi team, I was running into the same issue reported here: https://github.com/microsoft/playwright/issues/28798, where playwright was not able to correctly match snapshots with EXIF jpegs. 
- This PR relaxes the hex signature check for the JPEG sequence to allow JPEGs with alternate byte sequences
- Added unit test to ensure this can correctly categorize different JPEG signatures


tanayv:

pavelfeldman (approved):

PR #29221, opened 2024-01-29
docs: add an example of how to use "+" key as part of a shortcut

bamanczak:
A collegue has spent more than an hour trying to figure out how to use a `+` sign as part of a custom keyboard shortcut in his project. He tried e.g. `await page.keyboard.press("Shift+Alt+Add")`, `await page.keyboard.press("Shift+Alt+"+"")` and other crazy options, before we figured out `++` works as well. 

Maybe adding it as an example would help others. I used `Control` + `+` as the most common shortcut using `+` that I could think of.


bamanczak:

pavelfeldman (approved):

PR #29248, opened 2024-01-31
feat(test runner): tags/annotations

### LINKED ISSUE #23180: [Feature] tags API
Currently [adding tags](https://playwright.dev/docs/test-annotations#tag-tests) is only possible by inserting them into test title, e.g.:
```ts
test('Test login page @fast @mobile', async ({ page }) => {
  // ...
});
```
It may have some downsides for reporting when you get very long titles, see https://github.com/microsoft/playwright/issues/17375

Suggestion is to introduce `test.tags()` method that will add tags to current test without altering the title:
```ts
test.tags('@fast', '@mobile');
test('Test login page', async ({ page }) => {
  // ...
});
```

### LINKED ISSUE #29229: [Feature] beforeEach/fixtures, but executed even for skipped tests
There is no way to execute code for every test, even those defined with `test.skip('', () => {})`.

This is required in order to customize reporting for skipped tests. For other tests it can be done using a fixture or with `beforeEach` hook, but skipped tests execute neither of those, and no customization, such as assigning labels in Allure dependent on test location, is possible.

dgozman:
API changes:
- `test(title, details, body)` where details contain `tag` and `annotation`.
- similar `details` property added to `test.skip`, `test.fail`, `test.fixme`, `test.only`, `test.describe` and other `test.describe.*` variations.
- `TestProject.tagFilter`/`TestConfig.tagFilter` that supports logical tag expressions with `(`, `)`, `and`, `or` and `not`.
- `--tag` CLI option to filter by tags.
- New annotations are available in `TestInfo.annotations` and `TestCase.annotations`.
- New tags are available in `TestCase.tags`.
    
Reporter changes:
- `json` reporter includes new tags in addition to old `@smoke`-style tags. **Breaking**: tags are now listed with the leading `@` symbol.
- `html` reporter filters by old and new tags with the same `@smoke` token.

Fixes #29229, fixes #23180.

### DIFF
--- a/docs/src/test-api/class-test.md
+++ b/docs/src/test-api/class-test.md
@@ -19,6 +19,9 @@ test('basic test', async ({ page }) => {
 
 Declares a test.
 
+* `test(title, body)`
+* `test(title, details, body)`
+
 **Usage**
 
 ```js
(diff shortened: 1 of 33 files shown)


pavelfeldman (approved):

docs/src/test-api/class-test.md
```diff
@@ -1552,6 +1552,73 @@ Step body.
 
 Whether to box the step in the report. Defaults to `false`. When the step is boxed, errors thrown from the step internals point to the step call site. See below for more details.
 
+## method: Test.tag
```
mxschmitt, line 1555: Should we add it on a describe as well?

docs/src/test-api/class-testproject.md
```diff
@@ -199,6 +199,34 @@ Use [`method: Test.describe.configure`] to change the number of retries for a sp
 
 Use [`property: TestConfig.retries`] to change this option for all projects.
 
+
+## property: TestProject.tagFilter
```
mxschmitt, line 203: Should we add TestInfo.tags as well?

Issue #29073, opened 2024-01-19
[BUG] screenshots hang on arm64 docker image


aristidesstaffieri:
### System info
- Playwright Version: v1.41.0
- Operating System: macOS 13.6.3
- Browser: Chromium
- Other info: Tests a browser extension using `chromium.launchPersistentContext`

### Source code
https://github.com/stellar/freighter/pull/1084
Tests are here - https://github.com/stellar/freighter/tree/master/extension/e2e-tests

**Steps**
`docker run --platform=linux/arm64 --ipc=host --rm --network host -v $(pwd):/work/ -w /work/ -it --cap-add=SYS_ADMIN --init mcr.microsoft.com/playwright:v1.41.0-jammy /bin/bash`
`yarn && yarn test:e2e`

**Expected**
Screenshots to be generated for linux.

**Actual**
Runner hangs indefinitely 

**Info**
I'm able to run these tests locally on my M1, and can include the `toHaveScreenshot` assertions and get the expected screenshots and desired behavior. When I try to run the same tests in the docker container, the runner hangs until it times out, unable to take the screenshot. I've tried increasing my timeout by a lot, and taking partial page screenshots, as well as disabling animations. Interestingly, the very first screenshot does get taken but any subsequent ones hang until a timeout happens.

<img width="904" alt="Screenshot 2024-01-19 at 1 20 16 PM" src="https://github.com/microsoft/playwright/assets/6886006/6f75520d-618b-4d6a-89b4-a5b9c70fc913">

yury-s:
Does it also hang without the extension?

yury-s:
> stellar/freighter#1084

Note that the linux screenshots are missing, so the tests are bound to fail. Can you generate the expected images and retry the tests?

Btw, on github actions the default hardware is x64, not arm and the tests are failing there too in your PR.

aristidesstaffieri:
> Does it also hang without the extension?

The whole project is a browser extension so I don't think I can test that with this repo, but if I write a very rudimentary test suite targeting a web app context it does not hang so I think it's fair to be suspicious of using `chromium.launchPersistentContext` and trying to take screenshots on that image.

aristidesstaffieri:
> > Does it also hang without the extension?
> 
> The whole project is a browser extension so I don't think I can test that with this repo, but if I write a very rudimentary test suite targeting a web app context it does not hang so I think it's fair to be suspicious of using `chromium.launchPersistentContext` and trying to take screenshots on that image.

[...]

PR #29272, opened 2024-01-31
fix: do not retry missing snapshot errors

yury-s:
When `updateSnapshots === 'missing'` we generate new expectations on the first attempt and don't retry the test afterwards instead of trying it retries-1 times and only writing new expectation on the last attempt. This logic infects all serial mode suites that contain the test with missing expectations, so they also will not be retried.

Reference https://github.com/microsoft/playwright/issues/29073

### DIFF
--- a/packages/playwright/src/worker/testInfo.ts
+++ b/packages/playwright/src/worker/testInfo.ts
@@ -349,6 +349,8 @@ export class TestInfoImpl implements TestInfo {
   }
 
   _failWithError(error: Error, isHardError: boolean, retriable: boolean) {
+    if (!retriable)
+      this._hasNonRetriableError = true;
     // Do not overwrite any previous hard errors.
     // Some (but not all) scenarios include:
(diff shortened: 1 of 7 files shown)


packages/playwright/src/worker/testInfo.ts
```diff
@@ -361,6 +363,8 @@ export class TestInfoImpl implements TestInfo {
     const step = (error as any)[stepSymbol] as TestStepInternal | undefined;
     if (step && step.boxedStack)
       serialized.stack = `${error.name}: ${error.message}\n${stringifyStackFrames(step.boxedStack).join('\n')}`;
+    if (!retriable)
```
dgozman, line 366: It seems like this should be placed before the early return from this function.

dgozman (approved):

packages/playwright/src/worker/testInfo.ts
yury-s: good point!

PR #29275, opened 2024-01-31
devops: run test-runner tests on Node.js 21

mxschmitt:
Motivation: Good to test the different code-paths of our ESM loader.


mxschmitt:
new ESM loader APIs are backported to v18, this means we are already testing them. Closing by that.

PR #29310, opened 2024-02-02
chore: do not set metadata.error for expect failure results

yury-s:
The metadata.error change was brought back in https://github.com/microsoft/playwright/pull/29271and it broke java port as we could have error and result set simulteniously. This PR moves the logic to the trace recorder instead and keeps the protocol contract clear that either error or result is present, but not both.

### DIFF
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -1372,6 +1372,14 @@ export class Frame extends SdkObject {
   }
 
   async expect(metadata: CallMetadata, selector: string, options: FrameExpectParams): Promise<{ matches: boolean, received?: any, log?: string[], timedOut?: boolean }> {
+    const result = await this._expectImpl(metadata, selector, options);
+    // Library mode special case for the expect errors which are return values, not exceptions.
+    if (metadata.result && metadata.result.matches === metadata.params?.isNot)
+      metadata.fakeError = { error: { name: 'Expect', message: 'Expect failed' }};
+    return result;
+  }
+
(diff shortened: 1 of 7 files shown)


packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -559,11 +559,17 @@ function createActionLogTraceEvent(metadata: CallMetadata, message: string): tra
 function createAfterActionTraceEvent(metadata: CallMetadata): trace.AfterActionTraceEvent | null {
   if (metadata.internal || metadata.method.startsWith('tracing'))
     return null;
+  let error = metadata.error?.error;
+  // Library mode special case for expect errors which are return values, not exceptions.
```
dgozman, line 563: Perhaps a separate field in CallMetadata, e.g. `fakeError`? I feel uneasy hardcoding expect here. I'd rather do the hack in `Frame.expect`, not even in `FrameDispatcher.expect`.

packages/playwright-core/src/server/frames.ts
```diff
@@ -1372,6 +1372,14 @@ export class Frame extends SdkObject {
   }
 
   async expect(metadata: CallMetadata, selector: string, options: FrameExpectParams): Promise<{ matches: boolean, received?: any, log?: string[], timedOut?: boolean }> {
+    const result = await this._expectImpl(metadata, selector, options);
+    // Library mode special case for the expect errors which are return values, not exceptions.
+    if (result.matches === options.isNot)
+      metadata.fakeError = { error: { name: 'Expect', message: 'Expect failed' }};
```
mxschmitt, line 1378: Lets use a symbol instead?
We still send fakeError to the client, would love to never send it to the client.

dgozman (approved):

PR #29321, opened 2024-02-02
fix(trace viewer): correctly trim locations on Windows

mxschmitt:


dgozman (approved):
Since this is not fixing the test, but rather the functionality, I'd say the right PR title should be `fix(trace viewer): correctly trim locations on Windows`

PR #29322, opened 2024-02-02
feat(electron): expose ElectronApplication console events

### LINKED ISSUE #5905: [Feature] console events for electron app

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/5905

### DIFF
--- a/docs/src/api/class-electronapplication.md
+++ b/docs/src/api/class-electronapplication.md
@@ -45,7 +45,7 @@ This event is issued when the application closes.
 * since: v1.42
 - argument: <[ConsoleMessage]>
 
-Emitted when JavaScript within the Electron main process calls one of console API methods, e.g. `console.log` or `console.dir`. Also emitted if the page throws an error or a warning.
+Emitted when JavaScript within the Electron main process calls one of console API methods, e.g. `console.log` or `console.dir`.
 
 The arguments passed into `console.log` are available on the [ConsoleMessage] event handler argument.
 
--- a/docs/src/api/class-electronapplication.md
+++ b/docs/src/api/class-electronapplication.md
@@ -41,6 +41,26 @@ const { _electron: electron } = require('playwright');
 
 This event is issued when the application closes.
 
+## event: ElectronApplication.console
+* since: v1.42
+- argument: <[ConsoleMessage]>
+
+Emitted when JavaScript within the Electron main process calls one of console API methods, e.g. `console.log` or `console.dir`. Also emitted if the page throws an error or a warning.
+
+The arguments passed into `console.log` are available on the [ConsoleMessage] event handler argument.
+
+**Usage**
+
+```js
+electronApp.on('console', async msg => {
+  const values = [];
+  for (const arg of msg.args())
+    values.push(await arg.jsonValue());
+  console.log(...values);
+});
+await electronApp.evaluate(() => console.log('hello', 5, { foo: 'bar' }));
+```
+
 ## event: ElectronApplication.window
 * since: v1.9
 - argument: <[Page]>
--- a/packages/playwright-core/src/server/console.ts
+++ b/packages/playwright-core/src/server/console.ts
@@ -14,28 +14,26 @@
  * limitations under the License.
  */
 
-import { SdkObject } from './instrumentation';
 import type * as js from './javascript';
 import type { ConsoleMessageLocation } from './types';
-import { Page } from './page';
+import type { Page } from './page';
 
-export class ConsoleMessage extends SdkObject {
+export class ConsoleMessage {
   private _type: string;
   private _text?: string;
   private _args: js.JSHandle[];
   private _location: ConsoleMessageLocation;
-  private _page: Page | undefined;
+  private _page: Page | null;
 
-  constructor(parent: SdkObject, type: string, text: string | undefined, args: js.JSHandle[], location?: ConsoleMessageLocation) {
-    super(parent, 'console-message');
-    this._page = parent instanceof Page ? parent : undefined;
+  constructor(page: Page | null, type: string, text: string | undefined, args: js.JSHandle[], location?: ConsoleMessageLocation) {
+    this._page = page;
     this._type = type;
     this._text = text;
     this._args = args;
     this._location = location || { url: '', lineNumber: 0, columnNumber: 0 };
   }
 
-  page(): Page | undefined {
+  page() {
     return this._page;
   }
 
--- a/packages/playwright-core/src/server/console.ts
+++ b/packages/playwright-core/src/server/console.ts
@@ -17,25 +17,25 @@
 import { SdkObject } from './instrumentation';
 import type * as js from './javascript';
 import type { ConsoleMessageLocation } from './types';
-import type { Page } from './page';
+import { Page } from './page';
 
 export class ConsoleMessage extends SdkObject {
   private _type: string;
   private _text?: string;
   private _args: js.JSHandle[];
   private _location: ConsoleMessageLocation;
-  private _page: Page;
+  private _page: Page | undefined;
 
-  constructor(page: Page, type: string, text: string | undefined, args: js.JSHandle[], location?: ConsoleMessageLocation) {
-    super(page, 'console-message');
-    this._page = page;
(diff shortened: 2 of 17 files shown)


packages/protocol/src/protocol.yml
```diff
@@ -970,6 +970,22 @@ Browser:
 
     close:
 
+ConsoleMessage:
+  type: mixin
+  properties:
+    page: Page?
```
dgozman, line 976: I don't like that we made `page` optional for context console. Let's move it out of the mixin and make unconditional?

packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -81,20 +86,54 @@ export class ElectronApplication extends SdkObject {
...
+      // - there's no execution context we can use to operate with message
+      //   arguments
+      // - these messages are reported before Playwright clients can subscribe
+      //   to the 'console'
+      //   page event.
+      //
+      // @see https://github.com/GoogleChrome/puppeteer/issues/3865
+      return;
+    }
+    const [crExecutionContext, nodeExecutionContext] = [this._nodeChromiumExecutionContext, this._nodeExecutionContext];
```
dgozman, line 116: I wonder why not use `this._` variables directly?

packages/playwright-core/src/server/dispatchers/electronDispatcher.ts
```diff
@@ -67,6 +81,15 @@ export class ElectronApplicationDispatcher extends Dispatcher<ElectronApplicatio
     return { handle: ElementHandleDispatcher.fromJSHandle(this, result) };
   }
 
+  async updateSubscription(params: channels.ElectronApplicationUpdateSubscriptionParams): Promise<void> {
+    if (params.enabled)
+      this._subscriptions.add(params.event);
+    else
+      this._subscriptions.delete(params.event);
+    if (params.event === 'console' && params.enabled)
+      this._electronApplication._emitBufferedConsoleMessages();
```
dgozman, line 90: Why buffer for electron? IIRC, we do not buffer for pages/contexts.

packages/playwright-core/src/server/dispatchers/electronDispatcher.ts
```diff
@@ -40,15 +41,28 @@ export class ElectronDispatcher extends Dispatcher<Electron, channels.ElectronCh
 export class ElectronApplicationDispatcher extends Dispatcher<ElectronApplication, channels.ElectronApplicationChannel, ElectronDispatcher> implements channels.ElectronApplicationChannel {
   _type_EventTarget = true;
   _type_ElectronApplication = true;
+  private readonly _subscriptions = new Set<channels.ElectronApplicationUpdateSubscriptionParams['event']>();
+  private readonly _electronApplication: ElectronApplication;
```
dgozman, line 45: I think you can use `this._object` instead.

packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -81,20 +86,54 @@ export class ElectronApplication extends SdkObject {
...
+    if (!crExecutionContext || !nodeExecutionContext) {
+      this._bufferedConsoleMessageEvents.push(event);
+      return;
+    }
+    const args = event.args.map(arg => crExecutionContext.createHandle(nodeExecutionContext, arg));
```
dgozman, line 121: I think you don't need `crExecutionContext` at all, just call `nodeExecutionContext.createHandle(arg)`.

packages/playwright-core/src/server/console.ts
```diff
@@ -17,25 +17,25 @@
...
   private _text?: string;
   private _args: js.JSHandle[];
   private _location: ConsoleMessageLocation;
-  private _page: Page;
+  private _page: Page | undefined;
 
-  constructor(page: Page, type: string, text: string | undefined, args: js.JSHandle[], location?: ConsoleMessageLocation) {
-    super(page, 'console-message');
-    this._page = page;
+  constructor(parent: SdkObject, type: string, text: string | undefined, args: js.JSHandle[], location?: ConsoleMessageLocation) {
```
dgozman, line 29: Since we do not have ConsoleMessageDispatcher anymore, we might as well drop `SdkObject` inheritance in this class. This should make your change easier.

docs/src/api/class-electronapplication.md
```diff
@@ -41,6 +41,26 @@ const { _electron: electron } = require('playwright');
 
 This event is issued when the application closes.
 
+## event: ElectronApplication.console
+* since: v1.42
+- argument: <[ConsoleMessage]>
+
+Emitted when JavaScript within the Electron main process calls one of console API methods, e.g. `console.log` or `console.dir`. Also emitted if the page throws an error or a warning.
```
dgozman, line 48: Why is this also emitted when the page throws an error or a warning? I guess I also missed the implementation of that.

packages/playwright-core/src/server/dispatchers/electronDispatcher.ts
mxschmitt: The Problem is that we get the `Runtime.consoleAPICalled` events before the `ElectronApplicationDispatcher` has been created in such a case. Moving the buffer emit logic into the constructor of ElectronApplicationDispatcher is also not working, since we don't know at that time if the client object has been created already.

I'm not a huge fan of dropping these ConsoleMessages on the floor, so buffering felt appropriate.

It's about this test: https://github.com/microsoft/playwright/blob/d3721631e3ff8f1feea4481bc4019da3b74ed7fc/tests/electron/electron-app.spec.ts#L77-L97

docs/src/api/class-electronapplication.md
dgozman: The test I see explicitly does `process.on('uncaughtException')`. I don't think we'll get a console message for an actual exception.

packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -73,28 +77,60 @@ export class ElectronApplication extends SdkObject {
...
         return;
       const crExecutionContext = new CRExecutionContext(this._nodeSession, event.context);
-      this._nodeExecutionContext = new js.ExecutionContext(this, crExecutionContext, 'electron');
       const { result: remoteObject } = await crExecutionContext._client.send('Runtime.evaluate', {
         expression: `require('electron')`,
         contextId: event.context.id,
         // Needed after Electron 28 to get access to require: https://github.com/microsoft/playwright/issues/28048
         includeCommandLineAPI: true,
       });
+      this._nodeExecutionContext = new js.ExecutionContext(this, crExecutionContext, 'electron');
```
dgozman, line 86: I think if you bring this line back before the `await`, it is guaranteed that execution context exists before any `Runtime.consoleAPICalled` events may arrive.

packages/playwright-core/src/server/dispatchers/electronDispatcher.ts
```diff
@@ -67,6 +79,15 @@ export class ElectronApplicationDispatcher extends Dispatcher<ElectronApplicatio
     return { handle: ElementHandleDispatcher.fromJSHandle(this, result) };
   }
 
+  async updateSubscription(params: channels.ElectronApplicationUpdateSubscriptionParams): Promise<void> {
+    if (params.enabled)
+      this._subscriptions.add(params.event);
+    else
+      this._subscriptions.delete(params.event);
+    if (params.event === 'console' && params.enabled)
+      this._object._emitBufferedConsoleMessages();
```
dgozman, line 88: This does not fit my understanding of event listeners. Usually, you do not get past events upon subscribing to an event listener. Are we trying to solve some particular usecase?

docs/src/api/class-electronapplication.md
mxschmitt: Why is this not an actual exception?

The reason why I'm doing `process.on('uncaughtException')` is to prevent this from getting shown / blocking the main thread:

image

mxschmitt: Removed, turns out the console.error was the reason why it was printed.

dgozman (approved):

PR #29333, opened 2024-02-03
Fix incorrect code for specifying Javascript should be disabled

thatstatsguy:
Minor tweak to documentation

### DIFF
--- a/docs/src/emulation.md
+++ b/docs/src/emulation.md
@@ -792,5 +792,5 @@ context = browser.new_context(
 ```
 
 ```csharp
-var context = await browser.NewContextAsync(new() { JavaScriptEnabled = true });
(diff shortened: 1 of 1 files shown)


docs/src/emulation.md
```diff
@@ -792,5 +792,5 @@ context = browser.new_context(
 ```
 
 ```csharp
-var context = await browser.NewContextAsync(new() { JavaScriptEnabled = true });
+var context = await browser.NewContextAsync(new() { JavaScriptEnabled = false });
```
mxschmitt, line 795: Let's fix the snippet in line 784 and 790 too? It should be `java_script_enabled`.

thatstatsguy: done :)

mxschmitt (approved):

PR #29373, opened 2024-02-05
fix(chromium): best effort 304 status on reload

yury-s:
Reference https://github.com/microsoft/playwright/issues/28779


yury-s:
The result with this approach is still flaky, like ~10 out of 1000 runs it is 200, the rest is 304.

dgozman (approved):

PR #29374, opened 2024-02-06
chore: simplify types handling in toHaveScreenshot

pavelfeldman:


packages/playwright/src/matchers/toMatchSnapshot.ts
```diff
@@ -116,15 +142,24 @@ class SnapshotHelper<T extends ImageComparatorOptions> {
       }
     }
 
-    options = {
-      ...configOptions,
-      ...options,
+    const filteredConfigOptions = configOptions;
+    for (const prop of NonConfigProperties)
+      delete (filteredConfigOptions as any)[prop];
```
dgozman, line 147: Is this deleting on the user's config options object? Perhaps make a clone?

dgozman (approved):

packages/playwright/src/matchers/toMatchSnapshot.ts
pavelfeldman: Sure, but I am deleting the properties that we don't want to see there, so maybe it is Ok.

PR #29406, opened 2024-02-07
docs(java): update JUnit examples with new fixtures

jfgreffier:
Documentation for JUnit integration aka JUnit fixtures https://github.com/microsoft/playwright-java/issues/1369

Work in progress
- [x] update examples
- [x] list available fixtures in `test-runners` page
- [x] Add examples and explanations of Options

### DIFF
--- a/docs/src/api-testing-java.md
+++ b/docs/src/api-testing-java.md
@@ -375,7 +375,7 @@ public class TestGitHubAPI {
 }
 ```
 
-See experimental [JUnit integration](./junit.md) to automatically initialize [Playwright] objects and more, thanks to fixtures.
+See experimental [JUnit integration](./junit.md) to automatically initialize Playwright objects and more.
 
 ## Prepare server state via API calls
 
--- a/docs/src/api-testing-java.md
+++ b/docs/src/api-testing-java.md
@@ -375,6 +375,8 @@ public class TestGitHubAPI {
 }
 ```
 
+See experimental [JUnit integration](./junit.md) to automatically initialize [Playwright] objects and more, thanks to fixtures.
+
 ## Prepare server state via API calls
 
 The following test creates a new issue via API and then navigates to the list of all issues in the
--- a/docs/src/junit-java.md
+++ b/docs/src/junit-java.md
@@ -77,7 +77,7 @@ Here is a list of the pre-defined fixtures:
 |page          |[Page]             |Isolated page for this test run.|
 |browserContext|[BrowserContext]   |Isolated context for this test run. The `page` fixture belongs to this context as well.|
 |browser       |[Browser]          |Browsers are shared across tests to optimize resources.|
-|playwright    |[Playwright]       |Playwright instance is shared between tests.|
+|playwright    |[Playwright]       |Playwright instance is shared between tests running on the same thread.|
 |request       |[APIRequestContext]|Isolated APIRequestContext for this test run. Learn how to do [API testing](./api-testing).|
 
 ## Customizing options
--- a/docs/src/junit-java.md
+++ b/docs/src/junit-java.md
@@ -124,11 +124,6 @@ to speed up execution (see [this page](https://junit.org/junit5/docs/snapshot/us
 Since it is not safe to use same Playwright objects from multiple threads without extra synchronization we recommend you create Playwright
 instance per thread and use it on that thread exclusively. Here is an example how to run multiple test classes in parallel.
 
-Use [`@TestInstance(TestInstance.Lifecycle.PER_CLASS)`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/TestInstance.html)
-annotation to make JUnit create one instance of a class for all test methods within that class (by default JUnit will create a new instance of the class
-for each test method). Store [Playwright] and [Browser] objects in instance fields. They will be shared between tests. Each instance of the class will use its
-own copy of Playwright. This is done automatically if you use the `@UsePlaywright` JUnit annotation.
-
 ```java
 @UsePlaywright
 class Test1 {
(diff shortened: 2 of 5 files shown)


jfgreffier:
@uchagani

My understanding is that `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` is not needed anymore for parallel tests when using `@UsePlaywright`. Correct me if I'm wrong

uchagani:
@jfgreffier correct it doesn't matter from a parallel testing perspective. The user can use it if they want but it's more controlling the lifecycle of the test class and not the parallelism.

The user only needs to set up parallelism according to Junit5 docs but playwright currently only supports class-level parallelism and not method-level.

yury-s (changes requested):
Since we are going to release JUnit integration as experimental in 1.42, let's keep the api/ docs as is. Ideally, we'd have a new page lilke `JUnit (experimental)` that would document the new way of writing tests with the fixtures (basically all the text below `### Fixtures`) . We can add links to it from other places such as api-testing and accessibility-testing pages as well as the main test runner page. Once we are sure the API the stable enough, we can apply the rest of the changes.

docs/src/api-testing-java.md
```diff
@@ -375,6 +375,8 @@ public class TestGitHubAPI {
 }
 ```
 
+See experimental [JUnit integration](./junit.md) to automatically initialize [Playwright] objects and more, thanks to fixtures.
```
yury-s, line 378: nit: let's drop ", thanks to fixtures" part here and below.

yury-s, line 378: nit: [Playwright] -> Playwright

docs/src/junit-java.md
```diff
@@ -0,0 +1,187 @@
...
+    page.locator("input[name=\"search\"]").fill("playwright");
+    page.locator("input[name=\"search\"]").press("Enter");
+    assertEquals("https://en.wikipedia.org/wiki/Playwright", page.url());
+  }
+}
+```
+
+## Fixtures
+
+You should add JUnit annotation `@UsePlaywright` to your test classes to use fixtures. Test fixtures are used to establish environment for each test, giving the test everything it needs and nothing else.
```
yury-s, line 56:
```suggestion
Simply add JUnit annotation `@UsePlaywright` to your test classes to enable Playwright fixtures. Test fixtures are used to establish environment for each test, giving the test everything it needs and nothing else.
```

docs/src/junit-java.md
```diff
@@ -0,0 +1,187 @@
...
+The `Page page` argument tells JUnit to setup the `page` fixture and provide it to your test method.
+
+Here is a list of the pre-defined fixtures:
+
+|Fixture       |Type               |Description                      |
+|:-------------|:------------------|:--------------------------------|
+|page          |[Page]             |Isolated page for this test run.|
+|browserContext|[BrowserContext]   |Isolated context for this test run. The `page` fixture belongs to this context as well.|
+|browser       |[Browser]          |Browsers are shared across tests to optimize resources.|
+|playwright    |[Playwright]       |Playwright instance is shared between tests.|
```
yury-s, line 80: nit: "Playwright instance is shared between tests running on the same thread."

docs/src/junit-java.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Running Tests in Parallel
+
+By default JUnit will run all tests sequentially on a single thread. Since JUnit 5.3 you can change this behavior to run tests in parallel
+to speed up execution (see [this page](https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-parallel-execution)).
+Since it is not safe to use same Playwright objects from multiple threads without extra synchronization we recommend you create Playwright
+instance per thread and use it on that thread exclusively. Here is an example how to run multiple test classes in parallel.
+
+Use [`@TestInstance(TestInstance.Lifecycle.PER_CLASS)`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/TestInstance.html)
+annotation to make JUnit create one instance of a class for all test methods within that class (by default each JUnit will create a new instance of the class
```
yury-s, line 128:
```suggestion
annotation to make JUnit create one instance of a class for all test methods within that class (by default JUnit will create a new instance of the class
```

yury-s (approved):
Looks very good, just a few nits.

docs/src/junit-java.md
```diff
@@ -0,0 +1,187 @@
...
+
+By default JUnit will run all tests sequentially on a single thread. Since JUnit 5.3 you can change this behavior to run tests in parallel
+to speed up execution (see [this page](https://junit.org/junit5/docs/snapshot/user-guide/index.html#writing-tests-parallel-execution)).
+Since it is not safe to use same Playwright objects from multiple threads without extra synchronization we recommend you create Playwright
+instance per thread and use it on that thread exclusively. Here is an example how to run multiple test classes in parallel.
+
+Use [`@TestInstance(TestInstance.Lifecycle.PER_CLASS)`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/TestInstance.html)
+annotation to make JUnit create one instance of a class for all test methods within that class (by default JUnit will create a new instance of the class
+for each test method). Store [Playwright] and [Browser] objects in instance fields. They will be shared between tests. Each instance of the class will use its
+own copy of Playwright. This is done automatically if you use the `@UsePlaywright` JUnit annotation.
```
uchagani, line 130: I don't think this necessary.  the extensions will automatically use the same Playwright instance per thread and 1 Browser instance per class without(or despite) this configuration.

jfgreffier: You're right, I think my intent was to highlight that @UsePlaywright automagically takes care of it. I just removed this section, it's confusing.

PR #29425, opened 2024-02-08
fix(tracing): do not serialise out-of-process Buffers

mxschmitt:


packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -496,10 +497,12 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
...
   if (Array.isArray(object))
     return object.map(o => visitTraceEvent(o, sha1s));
+  if (object instanceof Dispatcher)
+    return `<${(object as Dispatcher<any, any, any>)._type}>`;
   if (object instanceof Buffer)
-    return undefined;
+    return `<Buffer>`;
   if (object instanceof Date)
-    return object;
+    return `<Date:${object.toJSON()}>`;
```
pavelfeldman, line 505: Why did this change? Sounds like we had better traces with it as is?

tests/library/tracing.spec.ts
```diff
@@ -109,9 +109,7 @@ test('should not collect snapshots by default', async ({ context, page, server }
   expect(events.some(e => e.type === 'resource-snapshot')).toBeFalsy();
 });
 
-test('should not include buffers in the trace', async ({ context, page, server, mode }, testInfo) => {
-  test.skip(mode !== 'default', 'no buffers with remote connections');
```
mxschmitt, line 113: Note: Removing the skip covers the bug with a test.

packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -496,8 +497,10 @@ export class Tracing extends SdkObject implements InstrumentationListener, Snaps
 function visitTraceEvent(object: any, sha1s: Set<string>): any {
   if (Array.isArray(object))
     return object.map(o => visitTraceEvent(o, sha1s));
+  if (object instanceof Dispatcher)
```
pavelfeldman, line 500: when does this happen?

mxschmitt: When we call `_appendTraceEvent` (after) for `newPage`. This returns `{ result: PageDispatcher }`.

pavelfeldman (approved):

PR #29431, opened 2024-02-09
fix(electron): stalling on delayed process close

### LINKED ISSUE #27523: [REGRESSION]: electron tests sometimes freezes for 30 seconds
**Context:**
- GOOD Playwright Version: 1.36.2
- BAD Playwright Version: 1.38.1

**Code Snippet**

tested on main repository with command

```
npx playwright test electron-app.spec.ts
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.

Running 22 tests using 1 worker

  ok 1 tests\electron\electron-app.spec.ts:22:5 › should fire close event (1.2s)
  ok 2 tests\electron\electron-app.spec.ts:34:5 › should dispatch ready event (137ms)
  ok 3 tests\electron\electron-app.spec.ts:46:5 › should script application (145ms)
  ok 4 tests\electron\electron-app.spec.ts:51:5 › should preserve args (137ms)
  ok 5 tests\electron\electron-app.spec.ts:58:5 › should return windows (349ms)
  ok 6 tests\electron\electron-app.spec.ts:63:5 › should evaluate handle (147ms)
  ok 7 tests\electron\electron-app.spec.ts:68:5 › should route network (380ms)
  ok 8 tests\electron\electron-app.spec.ts:81:5 › should support init script (355ms)
  ok 9 tests\electron\electron-app.spec.ts:88:5 › should expose function (362ms)
  ok 10 tests\electron\electron-app.spec.ts:95:5 › should wait for first window (351ms)
  ok 11 tests\electron\electron-app.spec.ts:104:5 › should have a clipboard instance (151ms)
  ok 12 tests\electron\electron-app.spec.ts:111:5 › should test app that opens window fast (136ms)
  ok 13 tests\electron\electron-app.spec.ts:115:5 › should return browser window (378ms)
  ok 14 tests\electron\electron-app.spec.ts:122:5 › should bypass csp (453ms)
  ok 15 tests\electron\electron-app.spec.ts:137:5 › should create page for browser view (524ms)
  ok 16 tests\electron\electron-app.spec.ts:150:5 › should return same browser window for browser view pages (575ms)
  ok 17 tests\electron\electron-app.spec.ts:171:5 › should record video (1.5s)
  ok 18 tests\electron\electron-app.spec.ts:183:5 › should be able to get the first window when with a delayed navigation (2.3s)
  ok 19 tests\electron\electron-app.spec.ts:192:5 › should detach debugger on app-initiated exit (158ms)
  ok 20 tests\electron\electron-app.spec.ts:201:5 › should run pre-ready apis (140ms)
  ok 21 tests\electron\electron-app.spec.ts:205:5 › should resolve app path for folder apps (136ms)
  ok 22 tests\electron\electron-app.spec.ts:211:5 › should return app name / version from manifest (134ms)

  22 passed (41.1s)

```

**Describe the bug**

In the log you can see that the tests pass on average in 10 seconds, what happens in the remaining 30 seconds is unclear. Playwright freezes for 30 seconds after the last test is completed.

sometimes this command ends in 10 seconds, but rarely
previous version of playwright also ends this test in 10 seconds

mxschmitt:
**Investigation notes:**

When we call [`ElectronApplication.close()`](https://github.com/microsoft/playwright/blob/7a86bf4af62fe3e44a41f94c7a4ab7f645c205e1/packages/playwright-core/src/server/electron/electron.ts#L134-L140), we return before the actual process has been exited. This can/will lead under certain timing conditions to that [`gracefullyCloseSet`](https://github.com/microsoft/playwright/blob/7a86bf4af62fe3e44a41f94c7a4ab7f645c205e1/packages/playwright-core/src/utils/processLauncher.ts#L52) still has an entry.
When the worker then tries to exit, we call these entries, which will do [`options.attemptToGracefullyClose()`](https://github.com/microsoft/playwright/blob/7a86bf4af62fe3e44a41f94c7a4ab7f645c205e1/packages/playwright-core/src/server/electron/electron.ts#L217). A recursive call to `ElectronApplication.close()` which will never resolve, due to the fact that we call `BrowserContext.close` again (good) but the actual `close` event will only be fired once. See [here](https://github.com/microsoft/playwright/blob/7a86bf4af62fe3e44a41f94c7a4ab7f645c205e1/packages/playwright-core/src/server/electron/electron.ts#L136-L139). This is why it was stalling for 30s.

**Changes:**

Based on that, I see 2 changes which we should do:

1. Adjust ElectronApplication.close() so it can be called multiple times (like for a [normal browser](https://github.com/microsoft/playwright/blob/aff6cf3c83d5567ec2c662eebdc246090df39982/packages/playwright-core/src/server/browser.ts#L159))
2. Wait until the process has exited before returning from ElectronApplication.close() (like for a normal browser - [`Browser.close()`](https://github.com/microsoft/playwright/blob/aff6cf3c83d5567ec2c662eebdc246090df39982/packages/playwright-core/src/server/browser.ts#L159) -> [`gracefullyClose`](https://github.com/microsoft/playwright/blob/7a86bf4af62fe3e44a41f94c7a4ab7f645c205e1/packages/playwright-core/src/server/browserType.ts#L243)).

Fixes https://github.com/microsoft/playwright/issues/27523

### DIFF
--- a/packages/playwright-core/src/server/electron/electron.ts
+++ b/packages/playwright-core/src/server/electron/electron.ts
@@ -65,7 +65,7 @@ export class ElectronApplication extends SdkObject {
   private _startedClosing: boolean = false;
   private _appClosePromise: Promise<unknown>;
 
-  constructor(parent: SdkObject, browser: CRBrowser, nodeConnection: CRConnection, process: childProcess.ChildProcess, gracefullyClose: () => Promise<void>) {
+  constructor(parent: SdkObject, browser: CRBrowser, nodeConnection: CRConnection, process: childProcess.ChildProcess) {
     super(parent, 'electron-app');
     this._process = process;
     this._browserContext = browser._defaultContext as CRBrowserContext;
@@ -91,7 +91,6 @@ export class ElectronApplication extends SdkObject {
       const electronHandle = await this._nodeElectronHandlePromise;
       await electronHandle.evaluate(({ app }) => app.quit()).catch(() => {});
       this._nodeConnection.close();
-      await gracefullyClose();
     });
   }
 
@@ -274,7 +273,7 @@ export class Electron extends SdkObject {
       };
       validateBrowserContextOptions(contextOptions, browserOptions);
       const browser = await CRBrowser.connect(this.attribution.playwright, chromeTransport, browserOptions);
-      app = new ElectronApplication(this, browser, nodeConnection, launchedProcess, gracefullyClose);
+      app = new ElectronApplication(this, browser, nodeConnection, launchedProcess);
       await app.initialize();
       return app;
     }, TimeoutSettings.launchTimeout(options));
--- a/packages/playwright-core/src/server/electron/electron.ts
+++ b/packages/playwright-core/src/server/electron/electron.ts
@@ -61,10 +61,13 @@ export class ElectronApplication extends SdkObject {
   _nodeElectronHandlePromise: ManualPromise<js.JSHandle<typeof import('electron')>> = new ManualPromise();
   readonly _timeoutSettings = new TimeoutSettings();
   private _process: childProcess.ChildProcess;
+  private _gracefullyClose: () => Promise<void>;
+  private _startedClosing: boolean = false;
 
-  constructor(parent: SdkObject, browser: CRBrowser, nodeConnection: CRConnection, process: childProcess.ChildProcess) {
+  constructor(parent: SdkObject, browser: CRBrowser, nodeConnection: CRConnection, process: childProcess.ChildProcess, gracefullyClose: () => Promise<void>) {
     super(parent, 'electron-app');
     this._process = process;
+    this._gracefullyClose = gracefullyClose;
     this._browserContext = browser._defaultContext as CRBrowserContext;
     this._browserContext.on(BrowserContext.Events.Close, () => {
       // Emit application closed after context closed.
@@ -132,11 +135,12 @@ export class ElectronApplication extends SdkObject {
   }
 
   async close() {
-    const progressController = new ProgressController(serverSideCallMetadata(), this);
-    const closed = progressController.run(progress => helper.waitForEvent(progress, this, ElectronApplication.Events.Close).promise);
+    if (this._startedClosing)
+      return;
+    this._startedClosing = true;
     await this._browserContext.close({ reason: 'Application exited' });
     this._nodeConnection.close();
-    await closed;
+    await this._gracefullyClose();
   }
 
   async browserWindow(page: Page): Promise<js.JSHandle<BrowserWindow>> {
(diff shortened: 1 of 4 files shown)


tests/electron/electron-app.spec.ts
```diff
@@ -26,6 +26,7 @@ test('should fire close event', async ({ launchElectronApp }) => {
   electronApp.on('close', () => events.push('application'));
   electronApp.context().on('close', () => events.push('context'));
   await electronApp.close();
+  await electronApp.close();
```
dgozman, line 29:
```suggestion
  // Close one more time - this should be a noop.
  await electronApp.close();
```

packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -132,11 +132,13 @@ export class ElectronApplication extends SdkObject {
...
-    const progressController = new ProgressController(serverSideCallMetadata(), this);
-    const closed = progressController.run(progress => helper.waitForEvent(progress, this, ElectronApplication.Events.Close).promise);
-    await this._browserContext.close({ reason: 'Application exited' });
-    this._nodeConnection.close();
-    await closed;
+    if (!this._startedClosing) {
+      this._startedClosing = true;
+      // This will call BrowserContext.setCustomCloseHandler.
+      await this._browserContext.close({ reason: 'Application exited' });
+      this._nodeConnection.close();
```
dgozman, line 139: Shall we move this line to the custom close handler?

packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -132,11 +132,13 @@ export class ElectronApplication extends SdkObject {
...
+    }
+    await this._gracefullyClose();
```
dgozman, line 141: Why do we call this? Just in case that `app.quit()` did not work? Shall we just wait for the process exit instead? Shall we move this to the custom close handler?

mxschmitt: Yes, I don't trust `app.quit()`, since we catch it anyways, so gracefullyClose felt better. Moved it into the close handler.

packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -132,11 +133,12 @@ export class ElectronApplication extends SdkObject {
...
-    const closed = progressController.run(progress => helper.waitForEvent(progress, this, ElectronApplication.Events.Close).promise);
-    await this._browserContext.close({ reason: 'Application exited' });
-    this._nodeConnection.close();
-    await closed;
+    if (!this._startedClosing) {
+      this._startedClosing = true;
+      // This will call BrowserContext.setCustomCloseHandler.
+      await this._browserContext.close({ reason: 'Application exited' });
+    }
+    await this._appClosePromise;
```
dgozman, line 141: I think we should move this to line 93.5, otherwise `context.close()` does not actually wait for the process to exit. From what I see, our intent is that `app.close() === context.close() === gracefullyClose()`.

I also think we can remove `this._startedClosing` flag.

mxschmitt: Good catch.

dgozman (approved):

PR #29445, opened 2024-02-09
chore: extract --project-grep option for matching with regex

yury-s:
Reference https://github.com/microsoft/playwright/issues/15128

### DIFF
--- a/packages/playwright/src/program.ts
+++ b/packages/playwright/src/program.ts
@@ -64,7 +64,8 @@ function addListFilesCommand(program: Command) {
   const command = program.command('list-files [file-filter...]', { hidden: true });
   command.description('List files with Playwright Test tests');
   command.option('-c, --config <file>', `Configuration file, or a test directory with optional "playwright.config.{m,c}?{js,ts}"`);
-  command.option('--project <project-name...>', `Only run tests from the projects matching this regular expression (default: list all projects)`);
+  command.option('--project <project-name...>', `Only run tests from the specified list of projects (default: list all projects)`);
+  command.option('--project-grep <pattern>', `Only run tests from the projects matching this regular expression (default: list all projects)`);
   command.action(async (args, opts) => {
     try {
       await listTestFiles(opts);
@@ -166,12 +167,16 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
   if (!config)
     return;
 
+  if (opts.project && opts.projectGrep)
+    throw new Error('Only one of --project and --project-grep can be specified.');
+
   config.cliArgs = args;
   config.cliGrep = opts.grep as string | undefined;
   config.cliGrepInvert = opts.grepInvert as string | undefined;
   config.cliTagFilter = opts.tag;
   config.cliListOnly = !!opts.list;
   config.cliProjectFilter = opts.project || undefined;
+  config.cliProjectGrep = opts.projectGrep || undefined;
   config.cliPassWithNoTests = !!opts.passWithNoTests;
 
   const runner = new Runner(config);
@@ -210,7 +215,9 @@ export async function withRunnerAndMutedWrite(configFile: string | undefined, ca
 }
 
 async function listTestFiles(opts: { [key: string]: any }) {
-  await withRunnerAndMutedWrite(opts.config, async runner => runner.listTestFiles(opts.project));
+  if (opts.project && opts.projectGrep)
+    throw new Error('Only one of --project and --project-grep can be specified.');
+  await withRunnerAndMutedWrite(opts.config, async runner => runner.listTestFiles(opts.project, opts.projectGrep));
 }
 
 async function mergeReports(reportDir: string | undefined, opts: { [key: string]: any }) {
@@ -323,7 +330,8 @@ const testOptions: [string, string][] = [
   ['--no-deps', 'Do not run project dependencies'],
   ['--output <dir>', `Folder for output artifacts (default: "test-results")`],
   ['--pass-with-no-tests', `Makes test run succeed even if no tests were found`],
-  ['--project <project-name...>', `Only run tests from the projects matching this regular expression (default: run all projects)`],
+  ['--project <project-name...>', `Only run tests from the specified list of projects(default: run all projects)`],
+  ['--project-grep <pattern>', `Only run tests from the projects matching this regular expression (default: run all projects)`],
   ['--quiet', `Suppress stdio`],
   ['--repeat-each <N>', `Run each test N times (default: 1)`],
   ['--reporter <reporter>', `Reporter to use, comma-separated, can be ${builtInReporters.map(name => `"${name}"`).join(', ')} (default: "${defaultReporter}")`],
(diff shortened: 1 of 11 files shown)


packages/playwright/src/program.ts
```diff
@@ -65,7 +65,8 @@ function addListFilesCommand(program: Command) {
   const command = program.command('list-files [file-filter...]', { hidden: true });
   command.description('List files with Playwright Test tests');
   command.option('-c, --config <file>', `Configuration file, or a test directory with optional "playwright.config.{m,c}?{js,ts}"`);
-  command.option('--project <project-name...>', `Only run tests from the projects matching this regular expression (default: list all projects)`);
+  command.option('--project <project-name...>', `Only run tests from the specified list of projects (default: list all projects)`);
+  command.option('--project-grep <project-name...>', `Only run tests from the projects matching this regular expression (default: list all projects)`);
```
dgozman, line 69: Make this one a single-value only?

packages/playwright/src/program.ts
```diff
@@ -328,7 +330,8 @@ const testOptions: [string, string][] = [
   ['--no-deps', 'Do not run project dependencies'],
   ['--output <dir>', `Folder for output artifacts (default: "test-results")`],
   ['--pass-with-no-tests', `Makes test run succeed even if no tests were found`],
-  ['--project <project-name...>', `Only run tests from the projects matching this regular expression (default: run all projects)`],
+  ['--project <project-name...>', `Only run tests from the specified list of projects(default: run all projects)`],
+  ['--project-grep <pattern...>', `Only run tests from the projects matching this regular expression (default: run all projects)`],
```
dgozman, line 334: Same here, single value only?

packages/playwright/src/util.ts
```diff
@@ -91,6 +91,12 @@ export function createFileFiltersFromArguments(args: string[]): TestFileFilter[]
   });
 }
 
+export function createProjectMatcher(projectNames?: string[], projectGrep?: string[]): Matcher {
+  const names = projectNames || [];
+  const regexps = (projectGrep || []).map(grep => forceRegExp(grep));
+  return (project: string) => (!names.length && !regexps.length) || names.includes(project) || regexps.some(re => re.test(project));
```
dgozman, line 97: I'd think that both project name filter and project grep filter should match, not just one of them. Or maybe disallow that case in the CLI right away?

tests/playwright-test/config.spec.ts
```diff
@@ -259,33 +260,31 @@ test('should filter by project and parse as RegExp', async ({ runInlineTest }) =
...
-  expect(new Set(result.outputLines)).toEqual(new Set(['foobar']));
+    'a.test.ts': `
+      import { test, expect } from '@playwright/test';
+      test('pass', async ({}, testInfo) => {
+        console.log(testInfo.project.name);
+      });
+    `
+  }, { '--project-grep': ['aaa', 'bbb', 'SUite1'] });
+  expect(exitCode).toBe(1);
+  expect(output).toContain('Project(s) matching "aaa", "bbb" not found. Available named projects: "suite1", "suite2"');
```
dgozman, line 287: Why is this an error? Perhaps only show an error when no projects matched at all?

tests/playwright-test/config.spec.ts
```diff
@@ -304,7 +303,7 @@ test('should filter by project and allow passing RegExp start/end flags', async
       test('one', async ({}) => {
         console.log('%%' + test.info().project.name);
       });    `
-  }, { project: '/fooBar$/' });
+  }, { '--project-grep': '/fooBar$/' });
```
dgozman, line 306: I'd say with a separate option we can probably omit `//`.

packages/playwright/src/runner/projectUtils.ts
```diff
@@ -19,38 +19,56 @@ import path from 'path';
...
     if (!names.length)
       throw new Error(`No named projects are specified in the configuration file`);
-    const unknownProjectNames = Array.from(unmatchedProjectFilters.values()).map(n => `"${n}"`).join(', ');
-    throw new Error(`Project(s) ${unknownProjectNames} not found. Available named projects: ${names.map(name => `"${name}"`).join(', ')}`);
+    if (unmatchedProjectNames.size) {
+      const unknownProjectNames = Array.from(unmatchedProjectNames.values()).map(n => `"${n}"`).join(', ');
+      throw new Error(`Project(s) ${unknownProjectNames} not found. Available named projects: ${names.map(name => `"${name}"`).join(', ')}`);
+    }
+    const unknownProjectGreps = Array.from(unmatchedProjectGreps.values()).map(n => `"${n}"`).join(', ');
+    throw new Error(`Project(s) matching ${unknownProjectGreps} not found. Available named projects: ${names.map(name => `"${name}"`).join(', ')}`);
```
dgozman, line 71: I'd drop this logic, and only throw when `result.length === 0`. WDYT?

tests/playwright-test/config.spec.ts
yury-s: this matches our --grep behavior

tests/playwright-test/config.spec.ts
yury-s: The idea is to do the same as with --project, i.e. each specified regex should match something. But with single regex it should not be a problem.

packages/playwright/src/program.ts
```diff
@@ -152,6 +157,9 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
 }
 
 async function listTestFiles(opts: { [key: string]: any }) {
+  if (opts.project && opts.projectGrep)
```
dgozman, line 160: I guess this method should also respect `opts.projectGrep` when listing?

packages/playwright/src/runner/projectUtils.ts
```diff
@@ -19,39 +19,49 @@ import path from 'path';
...
     return [...projects];
-  const unmatchedProjectFilters = new Set<string>(projectNames);
+
+  const names = projects.map(p => p.project.name).filter(name => !!name);
+  if (!names.length)
+    throw new Error(`No named projects are specified in the configuration file`);
+
+  if (projectGrep) {
+    const regex = forceRegExp(projectGrep);
+    const result = projects.filter(project => regex.test(project.project.name));
```
dgozman, line 37: When testing on the same RegExp instance, you need to reset `lastIndex` property.

packages/playwright/src/util.ts
```diff
@@ -91,6 +91,12 @@ export function createFileFiltersFromArguments(args: string[]): TestFileFilter[]
   });
 }
 
+export function createProjectMatcher(projectNames?: string[], projectGrep?: string): Matcher {
+  const names = projectNames || [];
+  const regexp = projectGrep ? forceRegExp(projectGrep) : undefined;
+  return (project: string) => (!!names.length && names.includes(project)) || (!!regexp && regexp.test(project));
```
dgozman, line 97: The other place uses `toLowerCase` on the project names filter. I also wonder why do we have two separate filters in our codebase?

dgozman (approved):

packages/playwright/src/util.ts
yury-s: Good question. Change this place to use filterProjects too.

Issue #26777, opened 2023-08-29
[Feature] Pick the right context/launch options for VSCode codegen


mxschmitt:

scaleoutsean:
Will this also allow codegen with non-incognito browser sessions, or would that be a new enhancement request?
The reason I ask incognito won't allow installation of browser extensions, so anyone who needs to use extensions would want codegen to support non-incognito mode (unless there's a way to do it now).

mxschmitt:
This issue is not about that, it would be a new one. We recommend for now just calling page.pause() inside your persistent context, which should also give you codegen.

MoritzHenn:
Is there an update on this? I am currently looking into playwright and noticed the `testIdAttribute` not being respected when using the VS Code extension's "record" function. If I understood the various issues correctly this here would fix that as well? 🤔

Mickael-BL:
Hello,
The “testIdAttribute” property defined in the project configuration file is not taken into account when using the Playwright extension recorder in VS Code.
This is really annoying, as it forces us to manually call codegen.
I see that the integration of this problem has been postponed for several versions, can you make it happen quickly.
Original request #32681
Thanks

tkoscieln:
Hello,
we are facing the same issue with `testIdAttribute` being ignored. Is there a time estimate when this fix will be released?
Thank you

Skn0tt:
Working on it! No time estimate though.

PR #29454, opened 2024-02-12
feat(vscode): allow running codegen with selected browser/options

mxschmitt:
https://github.com/microsoft/playwright/issues/26777


packages/playwright/src/runner/runner.ts
```diff
@@ -70,7 +81,28 @@ export class Runner {
       report.projects.push({
         name: project.project.name,
         testDir: project.project.testDir,
-        use: { testIdAttribute: project.project.use.testIdAttribute },
+        use: {
```
pavelfeldman, line 84: I don't think we should send this information back and forth, it looks like this should rather be a part of the test server.

PR #29467, opened 2024-02-13
chore: pass the private data to the runner via config

pavelfeldman:


packages/playwright/src/common/config.ts
```diff
@@ -96,6 +97,11 @@ export class FullConfigInternal {
       workers: 0,
       webServer: null,
     };
+    for (const key in config) {
+      if (key.startsWith('@'))
+        (this.config as any)[key] = (config as any)[key];
```
dgozman, line 102: I wonder whether we `JSON.stringify()` the config somewhere, which would probably break?

packages/playwright/src/runner/runner.ts
```diff
@@ -48,10 +51,11 @@ export class Runner {
     this._config = config;
   }
 
-  async listTestFiles(projectNames: string[] | undefined, projectGrep: string | undefined): Promise<any> {
+  async listTestFiles(frameworkPackage: string | undefined, projectNames: string[] | undefined, projectGrep: string | undefined): Promise<any> {
```
dgozman, line 54:
```suggestion
  async listTestFiles(frameworkPackage: string | undefined, projectNames: string[] | undefined, projectGrep: string | undefined): Promise<ConfigListFilesReport> {
```

packages/playwright/src/runner/runner.ts
```diff
@@ -39,6 +40,8 @@ type ProjectConfigWithFiles = {
 
 type ConfigListFilesReport = {
   projects: ProjectConfigWithFiles[];
+  cliEntryPoint?: string;
+  error?: TestError;
```
dgozman, line 44: Where is this one set?

dgozman (approved):

packages/playwright/src/common/config.ts
pavelfeldman: I hope we never do that

packages/playwright/src/runner/runner.ts
pavelfeldman: withRunnerAndMutedWrite

Issue #29475, opened 2024-02-13
[Bug]: webServer not able to find self-signed certificate in certificate chain when using url option


JohnMarsden24:
### Version

1.41.1

### Steps to reproduce

1. Clone https://github.com/JohnMarsden24/playwright-bug
2. Run `npx playwright test`  -> tests will pass
3. In `playwright.config.ts` uncomment the `url` in the `webServer` options and comment out `port`
4. Run `DEBUG=pw:webserver npx playwright test` -> startup will fail with `Error while checking if https://127.0.0.1:3000/ is available: self-signed certificate in certificate chain`

### Expected behavior

I expect playwright to pick up the HTTPs certificate regardless if I use `url` or `port` setting in `webServer`

### Actual behavior

When using `url` it can not locate the HTTPs certificate and will timeout. Using `port` works just fine. This also has the knock on effect that we can't use the `PLAYWRIGHT_TEST_BASE_URL` environment variable because that gets overwritten [here](https://github.com/microsoft/playwright/blob/ff4191335d7d483673328ce93a4af4284f67c8ee/packages/playwright/src/plugins/webServerPlugin.ts#L191) as we're using the `port` option.

### Additional context

[I am using Nexts experimental HTTPs for local development](https://nextjs.org/docs/app/api-reference/next-cli#https-for-local-development)

### Environment

```shell
System:
    OS: macOS 14.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 79.95 MB / 16.00 GB
  Binaries:
    Node: 18.17.0 - ~/.nvm/versions/node/v18.17.0/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.0/bin/npm
    pnpm: 8.14.3 - ~/.nvm/versions/node/v18.17.0/bin/pnpm
  IDEs:
    VSCode: 1.85.2 - /usr/local/bin/code
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.41.2 => 1.41.2
```

mxschmitt:
Looks like your repository is private. We have an option to ignore HTTPS errors for the webServer option. It would look like this:

```ts
import { defineConfig } from '@playwright/test';

export default defineConfig({
  webServer: {
    command: 'npm run start',
    url: 'http://127.0.0.1:3000',
    reuseExistingServer: !process.env.CI,
[...]

PR #29490, opened 2024-02-14
docs(webServer): add missing options

mxschmitt:
The options were really hidden [here](https://playwright.dev/docs/api/class-testconfig#test-config-web-server).

This syncs them into the "guide doc".

Relates https://github.com/microsoft/playwright/issues/29475.

### DIFF
--- a/docs/src/test-webserver-js.md
+++ b/docs/src/test-webserver-js.md
@@ -32,8 +32,11 @@ export default defineConfig({
 | `command`| Shell command to start the local dev server of your app. |
 | `url`| URL of your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the server is ready to accept connections. |
 | `reuseExistingServer`| If `true`, it will re-use an existing server on the url when available. If no server is running on that url, it will run the command to start a new server. If `false`, it will throw if an existing process is listening on the url. To see the stdout, you can set the `DEBUG=pw:webserver` environment variable. |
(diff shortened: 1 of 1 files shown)


docs/src/test-webserver-js.md
```diff
@@ -32,8 +32,11 @@ export default defineConfig({
 | `command`| Shell command to start the local dev server of your app. |
 | `url`| URL of your http server that is expected to return a 2xx, 3xx, 400, 401, 402, or 403 status code when the server is ready to accept connections. |
 | `reuseExistingServer`| If `true`, it will re-use an existing server on the url when available. If no server is running on that url, it will run the command to start a new server. If `false`, it will throw if an existing process is listening on the url. To see the stdout, you can set the `DEBUG=pw:webserver` environment variable. |
+| `ignoreHTTPSErrors` | Whether to ignore HTTPS errors when fetching the `url`. Defaults to `false`. |
```
yury-s, line 35: Why do we have these options documented in 2 places, can we delete this one and refer to the API docs to avoid such bugs in the future?

yury-s (approved):

docs/src/test-webserver-js.md
mxschmitt: + that its only in a central place
- that its hard to link to the api docs, since they would have to scroll down.

I'd keep it for now as it is, less clicks required are for the users.

PR #29523, opened 2024-02-16
fix(fixtures): attribute teardown step to the right TestInfo instance

dgozman:


pavelfeldman (approved):

joeprivettsmart:
@dgozman I am seeing a lot of these messages "Internal error: fixture integrity at ..." in the terminal. I can see that you added this into the code, but what is this telling me about the fixture - is there a problem with the teardown process? How can I address this problem?

dgozman:
@joeprivettsmart Could you please file an issue with a repro? Internal error means that something went wrong in the test runner, we would love to fix it!

joeprivettsmart:
> @joeprivettsmart Could you please file an issue with a repro? Internal error means that something went wrong in the test runner, we would love to fix it!

I can't supply a repo or repro because this is part of a large application, but if you point me in the right direction, I can supply a snippet.

PR #29537, opened 2024-02-17
chore: remove --project-grep, add wildcard support

yury-s:
Reference #15128


packages/playwright-core/src/utils/index.ts
```diff
@@ -43,3 +43,4 @@ export * from './userAgent';
 export * from './zipFile';
 export * from './zones';
 export * from './isomorphic/locatorGenerators';
+export { escapeRegExp } from './isomorphic/stringUtils';
```
yury-s, line 46: Need to explicitly export just this function as otherwise isString clashes with the one exported from ./rtti.

dgozman (approved):

PR #29549, opened 2024-02-19
Expose playwright-core codegen 'open' internal so that playwright automation tests can be recorded inline

magebase-dev:


yury-s (changes requested):
Please start with filing a bug/feature request describing your use case and the motivation for this change. `program.ts` is an implementation detail and we don't commit to supporting it's typescript API compatibility going forward, so you might need a different solution for your problem.

pavelfeldman:
Closing as per message above.

Issue #29558, opened 2024-02-19
[Bug]: Call stack is not visible in the UI mode


spiderhands:
### Version

1.41.2

### Steps to reproduce

1. Install and initialize the starter Playwright project as described in the [docs](https://playwright.dev/docs/intro)
```
npm init playwright@latest
```

2. Run tests in UI mode with traces
```
playwright test --ui --trace on
```

3. In the opened UI click "Run all"
4. Select select any action in any test and open the **Source** tab
5. In separate terminal run
```
playwright show-trace ./test-results/example-get-started-link-chromium/trace.zip
```
6. In the trace-viewer select any action in any test and open the **Source** tab

### Expected behavior

I expect the Source tab in UI mode and trace viewer to show the **Call stack** sidebar
<img width="1392" alt="Screenshot 2024-02-19 at 4 08 34 PM" src="https://github.com/microsoft/playwright/assets/8925613/30bab14b-6f05-4420-9b78-e34100a61939">

### Actual behavior

The **Trace viewer** has the **Call stack** sidebar:
<img width="1392" alt="Screenshot 2024-02-19 at 4 08 34 PM" src="https://github.com/microsoft/playwright/assets/8925613/44a47395-755d-4ec2-9b0b-a9b08666345b">

The **UI mode** does not have a **Call stack** sidebar:
<img width="1392" alt="Screenshot 2024-02-19 at 4 11 01 PM" src="https://github.com/microsoft/playwright/assets/8925613/1f76b6ed-b7d4-41fa-bb0d-e4bd7e8bfedd">

### Additional context

I know that there is a way of **boxing** actions by using `test.step`, but that is not so helpful as being able to see everything in the side panel

### Environment

```shell
System:
  OS: macOS 14.2.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 111.27 MB / 16.00 GB
Binaries:
  Node: 20.11.0 - ~/.volta/tools/image/node/20.11.0/bin/node
  Yarn: 4.0.2 - ~/.volta/tools/image/yarn/4.0.2/bin/yarn
  npm: 10.2.4 - ~/.volta/tools/image/node/20.11.0/bin/npm
Languages:
  Bash: 3.2.57 - /bin/bash
```

pavelfeldman:
What's your use case? In the screenshots it looks like the stack would only take space. That's why it is hidden in the UI mode.

spiderhands:
[...]

Issue #29500, opened 2024-02-14
[Feature]: UI Mode - ability to go up and down the stack trace in the source tab (feature parity with trace viewer)


gal-checksum:
### 🚀 Feature Request

In the trace viewer, in the source tab, I can go up and down the stack trace to view the source of different functions.

![CleanShot 2024-02-14 at 15 52 34](https://github.com/microsoft/playwright/assets/114691635/e6b26838-5282-4fb9-b97e-f2447c6b2772)

However, this feature is not available in UI mode:
![CleanShot 2024-02-14 at 15 53 51](https://github.com/microsoft/playwright/assets/114691635/93e8736f-d246-42c3-b31b-4e0d09e533e3)

Can we implement it?

### Example

When a function fails, you can explore the source of the full stack. Maybe the error comes from a higher level function

### Motivation

Easier to debug tests + feature parity with trace viewer

mxschmitt:
I can reproduce, looks like a bug.

mxschmitt:
Folding into https://github.com/microsoft/playwright/issues/29558, looks intentional.

PR #29560, opened 2024-02-19
fix(ui): show stack frames in ui mode

spiderhands:
#29558, #29500

### DIFF
--- a/packages/trace-viewer/src/ui/sourceTab.tsx
+++ b/packages/trace-viewer/src/ui/sourceTab.tsx
@@ -27,10 +27,9 @@ import type { StackFrame } from '@protocol/channels';
 export const SourceTab: React.FunctionComponent<{
   stack: StackFrame[] | undefined,
   sources: Map<string, SourceModel>,
-  hideStackFrames?: boolean,
   rootDir?: string,
   fallbackLocation?: SourceLocation,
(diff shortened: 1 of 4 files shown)


spiderhands:

yury-s (dismissed):
This is undoing what was explicitly done in https://github.com/microsoft/playwright/pull/21590, let's make sure https://github.com/microsoft/playwright/issues/29558#issuecomment-1954810926 is addressed first and there is a good reason for reverting this mode.

pavelfeldman:
Looks good. A test caught your changes. You can update it to something like this:

```js
    test('should show trace source', async ({ runInlineTest, page, showReport }) => {
        'playwright.config.js': `
          module.exports = { use: { trace: 'on' } };
        `,
        'a.test.js': `
          async function evaluateWrapper(page, expression) {
            await page.evaluate(expression);
          }
          test('passes', async ({ page }) => {
            await evaluateWrapper(page, '2 + 2');
          });
        `,
      }, { reporter: 'dot,html' }, { PW_TEST_HTML_REPORT_OPEN: 'never' });
```

run this to check it:

```sh
npm run ttest reporter-html.spec.ts:438
```

spiderhands:
@pavelfeldman I've updated the PR and the tests are passing now.

pavelfeldman (approved):

Issue #29564, opened 2024-02-19
[Feature]: Make UI Mode compatible with subpath proxying


bockstaller:
### 🚀 Feature Request

UI-Mode supports test execution via docker and codespaces by starting it with the following command:

`npx playwright test --ui-port=8080 --ui-host=0.0.0.0`

The trace viewer is launched and a websocket connection to the test runner is established.

This breaks as soon as the port opened by playwright is proxied via a subpath instead of a subdomain because the websocket is only taking [the domain, port and guid into account.](https://github.com/microsoft/playwright/blob/31bd58f8f3cee1ece4838331f7614128a5581cab/packages/trace-viewer/src/ui/wsPort.ts#L23)

This mode isn't supported by codespaces, but other deployment methods for remote code instances support this https://coder.com/docs/code-server/latest/guide#using-a-subdomain [this way of proxying](https://coder.com/docs/code-server/latest/guide#using-a-subpath).

### Example

Following the [relative URI resolution](https://blog.cdivilly.com/2019/02/28/uri-trailing-slashes) should allow a different way of creating the Websocket connection URL.

Ap possible implementation replacing [this line](https://github.com/microsoft/playwright/blob/31bd58f8f3cee1ece4838331f7614128a5581cab/packages/trace-viewer/src/ui/wsPort.ts#L23) could be.

```
const wsURL = URL(`../${guid}`, document.location))
wsURL.protocol = (window.location.protocol === 'https:' ? 'wss:' : 'ws:')
const ws = new WebSocket(wsURL)
```

This should keep backward compatibility while allowing subpath proxying.

### Motivation

This change would make playwrights ui mode usable in situations where the devs influence on larger infrastructure is limited. 
Especially whenever it isn't possible to proxy via subdomains.

bockstaller:
@yury-s I created a PR making the suggested changes. Happy to iterate on it

bockstaller:
There is a small bug in the original implementation I am addressing in this PR:

https://github.com/microsoft/playwright/pull/29924

PR #29617, opened 2024-02-22
feat(uimode) uses relative paths to establish websocket connection

bockstaller:
Implements the changes discussed in https://github.com/microsoft/playwright/issues/29564 

Uses relative paths to establish the WebSocket connection.
This shouldn't change the behavior in cases where the UI mode is exposed via domain proxying - in cases where there is no preceding subpath.

**Testing**
- I didn't find any related tests that I could adapt
- I didn't find a guide on how to build and install Playwright to validate manually

Happy to do the work if you can point me in the right direction


yury-s (commented):
Let's add a test for this, you can look at

> I didn't find any related tests that I could adapt

You can look at tests/playwright-test/ui-mode-test-run.spec.ts, but we don't have a ready test that you could adopt.

> I didn't find a guide on how to build and install Playwright to validate manually

You can look at what our GHA workflows do: .github/workflows/tests_primary.yml

pavelfeldman:
I'd let it in w/o a test.

pavelfeldman (approved):

PR #29630, opened 2024-02-22
chore: use test server as a singleton

pavelfeldman:


packages/playwright/src/runner/testServer.ts
```diff
@@ -182,18 +198,42 @@ class Dispatcher {
...
     this._ws.send(JSON.stringify({ method, params }));
   }
+
+  private async _loadConfig(configFile: string, overrides?: ConfigCLIOverrides): Promise<FullConfigInternal> {
+    return loadConfig({ resolvedConfigFile: configFile, configDir: path.dirname(configFile) }, overrides);
+  }
+
+  private _syncEnv(env: NodeJS.ProcessEnv) {
+    for (const name in env)
+      process.env[name] = env[name];
```
dgozman, line 235: I still think we might get bugs because of not restoring `env` after commands. What do we even pass as `env` here?

dgozman (approved):

packages/playwright/src/runner/testServer.ts
```diff
@@ -145,27 +147,41 @@ class Dispatcher {
     await reporter.onExit();
   }
 
-  private async _runTests(reporterPath: string, locations: string[] | undefined, options: PlaywrightTestOptions) {
+  async test(params: {
```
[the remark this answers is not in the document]
pavelfeldman: The point is that I have a file that is shared between client and server.

packages/playwright/src/runner/testServer.ts
pavelfeldman: We are passing reporter parameters here, it is a temporary measure.

PR #29634, opened 2024-02-23
chore: add docker documentation back to Visual comparisons

rwu823:
Hello @pavelfeldman, We noticed that the update to the snapshot section using Docker was removed for reasons unknown. However, upon reviewing the original PR(https://github.com/microsoft/playwright/pull/28636) [modifications](https://github.com/microsoft/playwright/pull/28636/files#diff-e31b2be556e0eaa5f86f7d454d80796f0276712b4760b61e8207aeac49c48f14L45-L51), it seems unrelated to Docker. I propose reinstating the documentation, as having a unified golden standard for snapshots across different systems would be immensely beneficial.

@rayeedanwar @johnson-liang also proposed restoring the document [here](https://github.com/microsoft/playwright/commit/d242ff67ef3f3bd2bc3644b1c71068d20cb54d61#r137285360).

### DIFF
--- a/docs/src/test-snapshots-js.md
+++ b/docs/src/test-snapshots-js.md
@@ -55,6 +55,15 @@ npx playwright test --update-snapshots
 > Note that `snapshotName` also accepts an array of path segments to the snapshot file such as `expect().toHaveScreenshot(['relative', 'path', 'to', 'snapshot.png'])`.
 > However, this path must stay within the snapshots directory for each test file (i.e. `a.spec.js-snapshots`), otherwise it will throw.
 
+If you are not on the same operating system as your CI system, you can use Docker to generate/update the screenshots:
+
+```bash
+docker run --rm --network host -v $(pwd):/work/ -w /work/ -it mcr.microsoft.com/playwright:v%%VERSION%%-jammy /bin/bash
+npm install
+npx playwright test --update-snapshots
+```
+
+
 ## Options
 
 ### maxDiffPixels


pavelfeldman:
The paragraph was removed because we think that it is over-promises on the Docker solution. I.e. making the complete Docker story works is a large effort, while it was suggesting that this single-liner would solve it all.

rwu823:
> The paragraph was removed because we think that it is over-promises on the Docker solution. I.e. making the complete Docker story works is a large effort, while it was suggesting that this single-liner would solve it all.

@pavelfeldman Well, would it be considered adding a link to a detailed introduction page about Docker?(https://playwright.dev/docs/docker) Alternatively, we could add a paragraph to the Docker page explaining why Docker is used to update snapshots.

pavelfeldman:
> @pavelfeldman Well, would it be considered adding a link to a detailed introduction page about Docker?(https://playwright.dev/docs/docker) Alternatively, we could add a paragraph to the Docker page explaining why Docker is used to update snapshots.

My point is that you need to know a lot about Docker to be successful in using it for VRT. If user is not there, this hint in the docs will most likely end up wasting their time. If the user is there, telling them how to run a program in Docker is no huge help.

Also, Docker can be used in various ways:
- sources in docker, sources outside docker
- test runner in docker, test runner outside docker + playwright server in docker
- root vs pwuser for local and CI scenarios
- container life time, etc. etc.

Putting a one liner in the docs is like saying "... or you can build a rocket using this command".

In terms of things that could help, I'm hopeful comprehensive 3rd party tutorials will emerge that would demonstrate different models for Docker use with Playwright. Once those models mature, we'll be able to include them in the docs or link to the tutorials.

pavelfeldman:
Closing as per above.

Issue #29662, opened 2024-02-26
[Feature]: Remove cookie/cookies from context


PaulTriandafilov:
### 🚀 Feature Request

The feature should remove specific cookie/cookies from context

### Example

await context.removeCookies([cookieObject1, cookieObject2]);

### Motivation

Sometimes we set cookies via globalSetup to prepare context for most cases. For example, we set cookie to hide GDPR consent modal. However, in order to write tests to cover the DGPR consent modal cookie, we must remove the cookie from the context.
I believe there are other cases where this method would be useful.

PR #29698, opened 2024-02-27
feat(playwright-core): add remove cookies api

PaulTriandafilov:
Implements the changes suggested in https://github.com/microsoft/playwright/issues/29662

This patch contains removeCookies method to remove cookies from BrowserContext by it's names

### DIFF
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -269,11 +269,8 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
     await this._channel.clearCookies();
   }
 
-  async removeCookies(cookieNames: string | string[]): Promise<void> {
-    if (typeof cookieNames === 'string')
-      cookieNames = [cookieNames];
-
-    await this._channel.removeCookies({ cookieNames: cookieNames as string[] });
+  async removeCookies(cookies: network.RemoveNetworkCookieParam): Promise<void> {
+    await this._channel.removeCookies({ cookies });
   }
 
   async grantPermissions(permissions: string[], options?: { origin?: string }): Promise<void> {
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -269,6 +269,13 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
     await this._channel.clearCookies();
   }
 
+  async removeCookies(cookieNames: string | string[]): Promise<void> {
+    if (typeof cookieNames === 'string')
+      cookieNames = [cookieNames];
+
+    await this._channel.removeCookies({ cookieNames: cookieNames as string[] });
+  }
+
   async grantPermissions(permissions: string[], options?: { origin?: string }): Promise<void> {
     await this._channel.grantPermissions({ permissions, ...options });
   }
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -269,8 +269,8 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
     await this._channel.clearCookies();
   }
 
-  async removeCookies(cookies: network.RemoveNetworkCookieParam): Promise<void> {
-    await this._channel.removeCookies({ cookies });
+  async removeCookies(criteria: network.RemoveNetworkCookieParam): Promise<void> {
+    await this._channel.removeCookies({ criteria });
   }
 
   async grantPermissions(permissions: string[], options?: { origin?: string }): Promise<void> {
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -269,8 +269,8 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
     await this._channel.clearCookies();
   }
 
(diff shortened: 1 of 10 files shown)


packages/playwright-core/src/server/browserContext.ts
```diff
@@ -276,6 +276,15 @@ export abstract class BrowserContext extends SdkObject {
     return await this.doGetCookies(urls as string[]);
   }
 
+  async removeCookies(cookieNames: string | string[]): Promise<void> {
+    const setCookies = await this.cookies();
+
+    const newCookiesSet = setCookies.filter(c => !cookieNames.includes(c.name));
```
pavelfeldman, line 282: It seems like `context.removeCookies('long-cookie-name')` will remove cookies with names `long`, `cookie` and `name`.

packages/playwright-core/src/client/browserContext.ts
```diff
@@ -269,6 +269,13 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
     await this._channel.clearCookies();
   }
 
+  async removeCookies(cookieNames: string | string[]): Promise<void> {
```
pavelfeldman, line 272: The functionality you are implementing is more of a "remove cookies by name". I wonder if you'd like to be a bit more proactive and include optional `domain` and `path` parameters. 

```js
async removeCookies(cookies: { name?: string, domain?: string, path?: string });
```

that way you could be able to clear all cookies by name, all cookies for a given domain, all cookies for a given path and their combinations:

```js
await context.removeCookies({ name: 'session-id' });
await context.removeCookies({ domain: 'my-origin.com' });
await context.removeCookies({ path: '/api/v1' });
await context.removeCookies({ name: 'session-id', domain: 'my-origin.com' });
```

pavelfeldman (commented):
Looks great!

packages/playwright-core/src/client/browserContext.ts
PaulTriandafilov: Thank you for comments! Fixed. Please let me know if there is something else I could change or fix.

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -276,6 +276,22 @@ export abstract class BrowserContext extends SdkObject {
     return await this.doGetCookies(urls as string[]);
   }
 
+  async removeCookies(cookies: {name?: string, domain?: string, path?: string}): Promise<void> {
+    const setCookies = await this.cookies();
+
+    if (!cookies.name && !cookies.domain && !cookies.path)
```
Smrtnyk, line 282: the call above is a waste if you will early return here
maybe moving the `await  this.cookies()` under?

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -276,6 +276,22 @@ export abstract class BrowserContext extends SdkObject {
     return await this.doGetCookies(urls as string[]);
   }
 
+  async removeCookies(cookies: {name?: string, domain?: string, path?: string}): Promise<void> {
```
Smrtnyk, line 279: somehow I feel like `cookies` as name of the argument is misleading
I would expect array of cookies or something
maybe naming it `criteria` or something similar?

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -276,6 +276,22 @@ export abstract class BrowserContext extends SdkObject {
...
+    const setCookies = await this.cookies();
+
+    if (!cookies.name && !cookies.domain && !cookies.path)
+      return;
+
+    const filteredCookies = setCookies.filter(cookie => {
```
Smrtnyk, line 285: maybe naming this `cookiesToKeep` ?

packages/playwright-core/src/server/browserContext.ts
pavelfeldman: I second this, `filter` would probably be a better name, but we can fix it during the pre-release api review.

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -276,6 +276,22 @@ export abstract class BrowserContext extends SdkObject {
     return await this.doGetCookies(urls as string[]);
   }
 
+  async removeCookies(cookies: {name?: string, domain?: string, path?: string}): Promise<void> {
+    const setCookies = await this.cookies();
```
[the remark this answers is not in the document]
pavelfeldman: @Smrtnyk: We discourage stylistic review comments, especially from non-contributors.

packages/playwright-core/src/server/browserContext.ts
pavelfeldman: This should probably throw 'Either name, domain or path are required'. Same should be mentioned in the doc (md file).

packages/playwright-core/src/server/browserContext.ts
pavelfeldman: @Smrtnyk: We discourage stylistic review comments, especially from non-contributors.

pavelfeldman (commented):
Looks great, a couple of nits and it is ready to go!

packages/playwright-core/src/server/browserContext.ts
PaulTriandafilov: Got it! just pushed outdated fixes :(
let me quickly get things straightened out

PaulTriandafilov:
@pavelfeldman I guess it's done. Please, let me know if I missed something

pavelfeldman (approved):

PR #29721, opened 2024-02-28
chore: make CRNetworkManager handle multiple sessions

dgozman:
It was already handling worker sessions, but not OOPIFs. As a result, some functionality was properly implemented only for OOPIFs and not for workers.

This change removes OOPIFs fanout for network-related calls from CRPage and moves that to the CRNetworkManager, now also covering workers.

### DIFF
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -26,70 +26,88 @@ import type * as contexts from '../browserContext';
 import type * as frames from '../frames';
 import type * as types from '../types';
 import type { CRPage } from './crPage';
-import { assert, headersObjectToArray } from '../../utils';
+import { assert, headersArrayToObject, headersObjectToArray } from '../../utils';
 import type { CRServiceWorker } from './crServiceWorker';
-import { isProtocolError } from '../protocolError';
+import { isProtocolError, isSessionClosedError } from '../protocolError';
 
 type SessionInfo = {
   session: CRSession;
+  isMain?: boolean;
   workerFrame?: frames.Frame;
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -98,8 +116,19 @@ export class CRNetworkManager {
...
+    if (offline === this._offline)
+      return;
+    this._offline = offline;
+    await this._forEachSession(info => this._setOfflineForSession(info));
+  }
+
+  private async _setOfflineForSession(info: SessionInfo, initial?: boolean) {
+    if (initial && !this._offline)
+      return;
+    if (info.workerFrame)
```
yury-s, line 128: let's add a comment that there is no such method for workers

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -483,22 +528,11 @@ export class CRNetworkManager {
     (this._page?._frameManager || this._serviceWorker)!.requestFailed(request.request, !!event.canceled);
   }
 
-  private _maybeAdoptMainRequest(requestId: Protocol.Network.RequestId): InterceptableRequest | undefined {
+  private _maybeUpdateOOPIFMainRequest(sessionInfo: SessionInfo, request: InterceptableRequest) {
     // OOPIF has a main request that starts in the parent session but finishes in the child session.
-    if (!this._parentManager)
-      return;
-    const request = this._parentManager._requestIdToRequest.get(requestId);
     // Main requests have matching loaderId and requestId.
```
yury-s, line 533: Let's also check that this is a main request for non main session .

yury-s (approved):

PR #29722, opened 2024-02-28
docs: Improved JSDoc for expect.toPass

RenanGreca:
I added more details and a usage example to the `expect.toPass` function, making it easier to use without having to consult the website. The example is pulled from the [Assertions documentation](https://playwright.dev/docs/test-assertions#expecttopass).


RenanGreca:

pavelfeldman:
You are modifying the generated file. Consider modifying the source and running the build!

pavelfeldman (approved):

PR #29743, opened 2024-02-29
fix(playwright-core): async android close method in connect flow

kamal-kaur04:
Fixing asynchronous `AndroidDevice.close` method in connect flow (graceful device/browser close, i.e., async close() doesn't get resolved before worker exit).

The issue can be reproduced by extending fixture like below :

```
exports.test = base.test.extend({
  browser: async ({ playwright }, use, testInfo) => {
    if (testInfo.project.name.match(/cloud/)) {
      let vDevice;
      // define caps here
      vDevice = await playwright._android.connect(
        `wss://endpoint/playwright?caps=${encodeURIComponent(
          JSON.stringify(caps)
        )}`
      );
      await vDevice.shell("am force-stop com.android.chrome");
      vDevice.newContext = function() {
        return vDevice.launchBrowser.apply(this, arguments);
      };
    }
    await use(vDevice);
    await vDevice.close();
  },
});
```

Highly appreciate if we can consider accommodating this. Open for any suggestion here.

### DIFF
--- a/packages/playwright-core/src/client/android.ts
+++ b/packages/playwright-core/src/client/android.ts
@@ -118,6 +118,7 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel> i
   readonly _timeoutSettings: TimeoutSettings;
   private _webViews = new Map<string, AndroidWebView>();
   _shouldCloseConnectionOnClose = false;
+  private _closedPromise: Promise<void>;
 
   static from(androidDevice: channels.AndroidDeviceChannel): AndroidDevice {
     return (androidDevice as any)._object;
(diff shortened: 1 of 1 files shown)


kamal-kaur04:

packages/playwright-core/src/client/android.ts
```diff
@@ -244,6 +246,7 @@ export class AndroidDevice extends ChannelOwner<channels.AndroidDeviceChannel> i
         this._connection.close();
       else
         await this._channel.close();
+      await this._closedPromise;
```
pavelfeldman, line 249: Could you provide an example where this is necessary? The snippet in the description patches Playwright methods, so we can't use it as a justification.

pavelfeldman:
Closing as stale.

PR #29781, opened 2024-03-02
chore: inject string pool into the tele receiver

pavelfeldman:


packages/playwright/src/isomorphic/teleReceiver.ts
```diff
@@ -315,17 +314,16 @@ export class TeleReporterReceiver {
 
   private _onExit(): Promise<void> | void {
     // Free up the memory from the string pool.
```
dgozman, line 316: Does reusing the string pool leak memory, based on this comment? If not, please remove the comment.

dgozman (approved):

PR #29784, opened 2024-03-03
docs(class-coverage.md): Add example for V8 coverage report

cenfun:
Hi, Can I add a simple example to the docs, which is used to generate the V8 coverage report? Thanks,

### DIFF
--- a/docs/src/api/class-coverage.md
+++ b/docs/src/api/class-coverage.md
@@ -4,11 +4,11 @@
 
 Coverage gathers information about parts of JavaScript and CSS that were used by the page.
 
-An example of using JavaScript coverage to produce Istanbul report for page load:
-
 :::note
 Coverage APIs are only supported on Chromium-based browsers.
 :::
+
+An example of using JavaScript coverage to produce Istanbul report for page load:
 
 ```js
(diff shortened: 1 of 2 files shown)


dgozman:
@cenfun I am not sure Playwright is ready to recommend this package in an official code snippet, given that the package is 3 months old and does not have a lot of downloads on npm. This does not mean we don't like the functionality that the package provides! However, we would like to serve our users better by only recommending established packages. There must be a good reason to make an exception.

By the way, I already see Monocart reporter in the [third party reporters list](https://playwright.dev/docs/test-reporters#third-party-reporter-showcase), which looks like a good place for it.

cenfun:
@Smrtnyk thank you for tring.

@dgozman I understand, thanks for the explanation.
Yes, I have already built a Playwright custom reporter `monocart-reporter` in the third party reporters list, but this one `monocart-coverage-reports` is not a Playwright custom reporter, but a general coverage reporter. The reason for creating this coverage report is because this [example](https://playwright.dev/docs/api/class-coverage) actually doesn't work. It can't directly generate a coverage report, so just add this new example.

Issue #29710, opened 2024-02-28
[Bug]: playwright-core logs invalid instructions for missing install


remcohaszing:
### Version

1.42.0

### Steps to reproduce

In a clean environment, run the following JavaScript snippet:

```js
import { chromium } from 'playwright-core'
chromium.launch()
```

### Expected behavior

Playwright should instruct to run the following command:

```
npx playwright-core install
```

### Actual behavior

It instructs to run a command that doesn’t exist.

```
  ╔═════════════════════════════════════════════════════════════════════════╗
  ║ Looks like Playwright Test or Playwright was just installed or updated. ║
  ║ Please run the following command to download new browsers:              ║
  ║                                                                         ║
  ║     npx playwright install                                              ║
  ║                                                                         ║
  ║ <3 Playwright Team                                                      ║
  ╚═════════════════════════════════════════════════════════════════════════╝
```

### Additional context

The command needed to install playwright was changed somewhere in a minor version.

### Environment

```shell
System:
    OS: Linux 6.6 Pop!_OS 22.04 LTS
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 22.98 GB / 31.06 GB
    Container: Yes
  Binaries:
    Node: 20.11.1 - /usr/bin/node
    Yarn: 1.22.19 - ~/.local/bin/yarn
    npm: 10.2.4 - /usr/bin/npm
    pnpm: 8.11.0 - ~/.local/bin/pnpm
    bun: 1.0.29 - ~/.local/bin/bun
  IDEs:
    VSCode: 1.86.2 - /usr/bin/code
  Languages:
    Bash: 5.1.16 - /usr/bin/bash
  npmPackages:
    playwright-core: ^1.0.0 => 1.42.0
```

pavelfeldman:
#### Why was this issue closed?

Thank you for your contribution to our project. This issue has been closed due to its limited upvotes and recent activity, and insufficient feedback for us to effectively act upon. Our priority is to focus on bugs that reflect higher user engagement and have actionable feedback, to ensure our bug database stays manageable.

Should you feel this closure was in error, please create a new issue and reference this one. We're open to revisiting it given increased support or additional clarity. Your understanding and cooperation are greatly appreciated.

PR #29791, opened 2024-03-04
fix: use corresponding bin entry when reporting cli commands to the user

mxschmitt:
https://github.com/microsoft/playwright/issues/29710

### DIFF
--- a/packages/playwright-core/src/server/registry/index.ts
+++ b/packages/playwright-core/src/server/registry/index.ts
@@ -1040,7 +1040,8 @@ export function buildPlaywrightCLICommand(sdkLanguage: string, parameters: strin
       return `pwsh bin/Debug/netX/playwright.ps1 ${parameters}`;
     default: {
       const packageManagerCommand = getPackageManagerExecCommand();
-      return `${packageManagerCommand} playwright ${parameters}`;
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/registry/index.ts
```diff
@@ -1040,7 +1040,8 @@ export function buildPlaywrightCLICommand(sdkLanguage: string, parameters: strin
       return `pwsh bin/Debug/netX/playwright.ps1 ${parameters}`;
     default: {
       const packageManagerCommand = getPackageManagerExecCommand();
-      return `${packageManagerCommand} playwright ${parameters}`;
+      const binEntry = process.argv[1].endsWith(path.sep + 'playwright-core') ? 'playwright-core' : 'playwright';
```
dgozman, line 1043: - There are more places that call `getPackageManagerExecCommand()`, and some of them need updating as well.
- Also, there are two versions of this function: `buildBasePlaywrightCLICommand` and `buildPlaywrightCLICommand` - let's merge them?
- IIUC, this relies on `argv` being something like `npx playwright-core install`? How does this work in the issue example, where I manually import from `playwright-core`?

PR #29794, opened 2024-03-04
feat: exposes tags in testInfo

### LINKED ISSUE #29793: [Feature]: Expose test tags in test info.
### 🚀 Feature Request

I would like read-only access to the test tags in my fixtures.
Currently, only annotations are exposed.

### Example

_No response_

### Motivation

I am preparing the migration of a large e2e testing setup to playwright. 
The biggest pain points we are facing at the moment are:

- running only necessary tests in CI
- linking features and test implementation

I would like to achieve this via tags that could look like this: 

- `@feature:{feature_id}` -> this test tests feature with the given id
- `@view:{path}` -> this test touches this view

These tags need some enforcement to be useful. So I have implemented an automatic fixture ensuring that each tag type is present in each tests title path. 
Guaranteeing that all tests are tagged with both tag types.

This is not possible when tags are specified via the details object of a test. 
It isn't exposed in the testInfo object

bockstaller:
Fixes #29793.

### DIFF
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -214,7 +214,9 @@ Test function as passed to `test(title, testFunction)`.
 * since: v1.43
 - type: <[Array]<[string]>>
 
-Tags that apply to the tests.
+Tags that apply to the test. Learn more about [tags](../test-annotations.md#tag-tests).
+
+Note that any changes made to this list while the test is running will not be visible to test reporters.
 
 ## property: TestInfo.testId
 * since: v1.32
--- a/docs/src/test-api/class-testinfo.md
+++ b/docs/src/test-api/class-testinfo.md
@@ -210,6 +210,12 @@ Optional description that will be reflected in a test report.
 
 Test function as passed to `test(title, testFunction)`.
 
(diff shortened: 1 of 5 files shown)


bockstaller:

docs/src/test-api/class-testinfo.md
```diff
@@ -210,6 +210,12 @@ Optional description that will be reflected in a test report.
 
 Test function as passed to `test(title, testFunction)`.
 
+## property: TestInfo.tags
+* since: v1.43
+- type: <[Array]<[string]>>
+
+Tags that apply to the tests.
```
dgozman, line 217:
```suggestion
Tags that apply to the test. Learn more about [tags](../test-annotations.md#tag-tests).

Note that any changes made to this list while the test is running will not be visible to test reporters.
```

dgozman (commented):
@bockstaller Thank you for the great PR!

Unfortunately, there are lint issues due to us using two different version of playwright for our own tests. For now, you can fix them by adding `as any` where it complains, and we'll fix it later by rolling new test runner version.

dgozman (approved):
Excellent! Merging in.

PR #29828, opened 2024-03-06
feat: separate timeout for after hooks

dgozman:
Instead of sharing the timeout with the test, and then extending it when test times out, we give after hooks a separate timeout.


pavelfeldman (approved):

KidiIT (commented):
All reviewed and approved🫶❤️

rodrigojmartin:
Hello, how can I know in which version will this be released? Thanks!

mxschmitt:
It will be in the next stable version of Playwright, or help us by using Canary to beta-test Playwright: https://playwright.dev/docs/canary-releases

PR #29879, opened 2024-03-11
fix(playwright-ct-core): added missing exported test type

jan-molak:
This PR fixes the misalignment between `playwright-ct-core/index.js` and `playwright-ct-core/index.ts` and ensures the latter export `const test`.

https://github.com/microsoft/playwright/blob/7f6f17d1c343f88400c498fd1c3defbd7f420056/packages/playwright-ct-core/index.js#L42

https://github.com/microsoft/playwright/blob/7f6f17d1c343f88400c498fd1c3defbd7f420056/packages/playwright-ct-core/index.d.ts#L48

### DIFF
--- a/packages/playwright-ct-core/index.d.ts
+++ b/packages/playwright-ct-core/index.d.ts
@@ -38,6 +38,8 @@ export type TestType<ComponentFixtures> = BaseTestType<
   PlaywrightWorkerArgs & PlaywrightWorkerOptions
 >;
 
+export const test: TestType<ComponentFixtures>;
+
 export function defineConfig(config: PlaywrightTestConfig): PlaywrightTestConfig;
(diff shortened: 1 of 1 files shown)


mxschmitt:
Would it be possible to start with filing a bug? Users are not intended to import `'@playwright/experimental-ct-core'`. Are you creating a custom component testing integration? This would help us a lot to understanding the change.

jan-molak:
Hey @mxschmitt, we have a custom component testing integration for testing web components and noticed that the type defs in ct-core are not aligned with what's actually exported in index.js.

jan-molak:
@mxschmitt I raised #29903

pavelfeldman:
Closing as per discussion in the Issue.

PR #29882, opened 2024-03-11
feat(test runner): rework how tsconfig is applied

dgozman:
Previously, test runner would load tsconfig for each imported file based on its location, going up the directory tree.

Now, test runner mostly uses a single tsconfig for all imported files and respects `files`, `include` and `exclude` properties to determine whether to apply tsconfig to a particular file or not.

For backwards compatibility, root tsconfig is used to load `playwright.config.ts`, but when `testDir` has its own tsconfig, it used for loading all tests instead of the root tsconfig.

### DIFF
--- a/docs/src/test-typescript-js.md
+++ b/docs/src/test-typescript-js.md
@@ -5,9 +5,9 @@ title: "TypeScript"
 
 ## Introduction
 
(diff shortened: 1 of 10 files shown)


pavelfeldman (approved):

pavelfeldman:
Archving

PR #29889, opened 2024-03-12
fix: postDataJSON with application/x-www-form-urlencoded; charset=UTF-8

### LINKED ISSUE #29872: [Bug]: Request.post_data_json should not try to parse application/x-www-form-urlencoded; charset=UTF-8 as JSON data
### Version

main

### Steps to reproduce

1. Intercept a `Request` object sent to a server with a `Content-Type` HTTP header set to `application/x-www-form-urlencoded; charset=UTF-8` and POST data body of `param=value`.
2. Call `request.post_data_json`.

### Expected behavior

Dictionary `{"param": "value"}` is returned.

### Actual behavior

Error observed: `POST data is not a valid JSON object: {post_data}`.

### Additional context

Function `Request.post_data_json` has:
`if content_type == "application/x-www-form-urlencoded":`

This code path is not triggered if the `Content-Type` header specifies a UTF-8 character set, and the POST data body being `param=value` will be JSON parsed, failing with an error because this is not valid JSON data.

The check of the `Content-Type` header should also allow `application/x-www-form-urlencoded; charset=UTF-8`. Other non-UTF-8 character sets should then also either be supported by `Request.post_data_json` or this function should raise a separate error making it clear that an unsupported character set was found in the `Content-Type` header.

The RFC standard for the `Content-Type` header specifies at https://datatracker.ietf.org/doc/html/rfc9110#section-8.3.2 the possibility of including the `charset` in the header value.

### Environment

```Text
- Operating System: Fedora Rawhide
- CPU: x86_64
- Browser: Firefox
- Python Version: 3.12
- Other info: Playwright 1.41.2 with latest ubuntu20.04-x64 build of Firefox
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/29872

### DIFF
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -141,7 +141,7 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
       return null;
 
     const contentType = this.headers()['content-type'];
-    if (contentType === 'application/x-www-form-urlencoded') {
+    if (contentType.startsWith('application/x-www-form-urlencoded')) {
       const entries: Record<string, string> = {};
       const parsed = new URLSearchParams(postData);
       for (const [k, v] of parsed.entries())
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -141,7 +141,7 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
       return null;
 
     const contentType = this.headers()['content-type'];
-    if (contentType.startsWith('application/x-www-form-urlencoded')) {
+    if (contentType.includes('application/x-www-form-urlencoded')) {
       const entries: Record<string, string> = {};
       const parsed = new URLSearchParams(postData);
       for (const [k, v] of parsed.entries())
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/client/network.ts
```diff
@@ -141,7 +141,7 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
       return null;
 
     const contentType = this.headers()['content-type'];
-    if (contentType === 'application/x-www-form-urlencoded') {
```
yury-s, line 144: Should we also support more exotic `charset=UTF-8; application/x-www-form-urlencoded` ?

yury-s (approved):

packages/playwright-core/src/client/network.ts
```diff
@@ -141,7 +141,7 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
...
+    if (contentType.startsWith('application/x-www-form-urlencoded')) {
```
mxschmitt, line 144:
```suggestion
    if (contentType.includes('application/x-www-form-urlencoded')) {
```

PR #29890, opened 2024-03-12
fix(expect): throw better received error when no element was found

### LINKED ISSUE #29873: [Bug]:  toHaveText() fails with irrelevant error if applied to non-existing/non-visible elements
### Version

1.41.2

### Steps to reproduce

Create assertion expecting non-existing field to to have text. Ex:

```ts
test('field should be empty', async ({ page }) => {
            //case1
            await expect(page.locator('.i-do-not-exist'), 'field should be empty').toHaveText('');
            //case2
            await expect(page.locator('i-do-not-exist') , 'field should have 'abc' text').toHaveText('abc');
});     
```

### Expected behavior

the assertion should fail but signal the element do not exist instead of saying that element has '' text

### Actual behavior

case1: 

```
Locator: locator('.non-exising-element')
Expected string: ""
Received string: serializes to the same string
```

case2:

```
Error: field should have 'abc' text

    Timed out 15000ms waiting for expect(locator).toHaveText(expected)

    Locator: locator('.non-exising-element')
    Expected string: "abc"
    Received string: ""
```

### Additional context

_No response_

### Environment

```shell
System:
    OS: macOS 14.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 109.13 MB / 16.00 GB
  Binaries:
    Node: 18.18.2 - ~/.nvm/versions/node/v18.18.2/bin/node
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.41.2 => 1.41.2
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/29873

### DIFF
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -1382,6 +1382,8 @@ export class Frame extends SdkObject {
     // Library mode special case for the expect errors which are return values, not exceptions.
     if (result.matches === options.isNot)
       metadata.error = { error: { name: 'Expect', message: 'Expect failed' } };
+    if (result.log?.[result.log.length - 1].startsWith('waiting for '))
+      result.received = '<element(s) not found>';
     return result;
   }
 
--- a/packages/playwright-core/src/server/frames.ts
+++ b/packages/playwright-core/src/server/frames.ts
@@ -1382,8 +1382,6 @@ export class Frame extends SdkObject {
     // Library mode special case for the expect errors which are return values, not exceptions.
     if (result.matches === options.isNot)
       metadata.error = { error: { name: 'Expect', message: 'Expect failed' } };
-    if (result.log?.[result.log.length - 1].startsWith('waiting for '))
-      result.received = '<element(s) not found>';
     return result;
   }
 
@@ -1422,7 +1420,7 @@ export class Frame extends SdkObject {
         const injected = await context.injectedScript();
         progress.throwIfAborted();
 
-        const { log, matches, received, missingRecevied } = await injected.evaluate(async (injected, { info, options, callId }) => {
+        const { log, matches, received, missingReceived } = await injected.evaluate(async (injected, { info, options, callId }) => {
           const elements = info ? injected.querySelectorAll(info.parsed, document) : [];
           const isArray = options.expression === 'to.have.count' || options.expression.endsWith('.array');
           let log = '';
@@ -1434,16 +1432,16 @@ export class Frame extends SdkObject {
             log = `  locator resolved to ${injected.previewNode(elements[0])}`;
           if (callId)
             injected.markTargetElements(new Set(elements), callId);
-          return { log, ...(await injected.expect(elements[0], options, elements)) };
+          return { log, ...await injected.expect(elements[0], options, elements) };
         }, { info, options, callId: metadata.id });
 
         if (log)
           progress.log(log);
         // Note: missingReceived avoids `unexpected value "undefined"` when element was not found.
-        if (matches === options.isNot && !missingRecevied) {
-          lastIntermediateResult.received = received;
+        if (matches === options.isNot) {
+          lastIntermediateResult.received = missingReceived ? '<element(s) not found>' : received;
           lastIntermediateResult.isSet = true;
-          if (!Array.isArray(received))
+          if (!missingReceived && !Array.isArray(received))
             progress.log(`  unexpected value "${renderUnexpectedValue(options.expression, received)}"`);
         }
         if (!oneShot && matches === options.isNot) {
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -1166,7 +1166,7 @@ export class InjectedScript {
           throw this.createStacklessError('Element is not a checkbox');
         if (elementState === 'error:notconnected')
           throw this.createStacklessError('Element is not connected');
-        return { received: elementState, matches: elementState };
+        return { matches: elementState };
       }
     }
 
--- a/packages/playwright-core/src/server/injected/injectedScript.ts
+++ b/packages/playwright-core/src/server/injected/injectedScript.ts
@@ -1098,7 +1098,7 @@ export class InjectedScript {
     this.onGlobalListenersRemoved.add(addHitTargetInterceptorListeners);
   }
 
-  async expect(element: Element | undefined, options: FrameExpectParams, elements: Element[]): Promise<{ matches: boolean, received?: any, missingRecevied?: boolean }> {
+  async expect(element: Element | undefined, options: FrameExpectParams, elements: Element[]): Promise<{ matches: boolean, received?: any, missingReceived?: boolean }> {
     const isArray = options.expression === 'to.have.count' || options.expression.endsWith('.array');
     if (isArray)
       return this.expectArray(elements, options);
@@ -1119,7 +1119,7 @@ export class InjectedScript {
       if (options.isNot && options.expression === 'to.be.in.viewport')
         return { matches: false };
       // When none of the above applies, expect does not match.
-      return { matches: options.isNot, missingRecevied: true };
+      return { matches: options.isNot, missingReceived: true };
     }
     return await this.expectSingleElement(element, options);
   }
@@ -1166,7 +1166,7 @@ export class InjectedScript {
           throw this.createStacklessError('Element is not a checkbox');
         if (elementState === 'error:notconnected')
           throw this.createStacklessError('Element is not connected');
-        return { matches: elementState };
+        return { received: elementState, matches: elementState };
       }
     }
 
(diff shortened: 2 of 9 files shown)


packages/playwright-core/src/server/frames.ts
```diff
@@ -1444,16 +1445,18 @@ export class Frame extends SdkObject {
...
       // Q: Why not throw upon isSessionClosedError(e) as in other places?
       // A: We want user to receive a friendly message containing the last intermediate result.
       if (js.isJavaScriptErrorInEvaluate(e) || isInvalidSelectorError(e))
         throw e;
-      const result: { matches: boolean, received?: any, log?: string[], timedOut?: boolean } = { matches: options.isNot, log: metadata.log };
-      if (lastIntermediateResult.isSet)
+      const result: { matches: boolean, received?: any, log?: string[], timedOut?: boolean, resolvedtoAnElement: boolean } = { matches: options.isNot, log: metadata.log, resolvedtoAnElement: false };
+      if (lastIntermediateResult.isSet) {
         result.received = lastIntermediateResult.received;
+        result.resolvedtoAnElement = lastIntermediateResult.resolvedtoAnElement;
```
dgozman, line 1458: We can just update `received` here to `<element(s) not found>`.

packages/playwright/src/matchers/matcherHint.ts
```diff
@@ -20,6 +20,14 @@ import type { Locator } from 'playwright-core';
 import type { StackFrame } from '@protocol/channels';
 import { stringifyStackFrames } from 'playwright-core/lib/utils';
 
+const kNoElementsFoundMessage = '<element(s) not found>';
+
+export function rewriteIfNoElementFound(received: any, log: string[] | undefined): any {
+  if (log?.[log.length - 1].startsWith('waiting for '))
```
dgozman, line 26: I thought you'd still set `<element(s) not found>` on the server, and check the "received" here to avoid rewriting it into something on the client-side. This way, all language ports will also get it for free.

packages/playwright/src/matchers/matcherHint.ts
```diff
@@ -20,6 +20,7 @@ import type { Locator } from 'playwright-core';
 import type { StackFrame } from '@protocol/channels';
 import { stringifyStackFrames } from 'playwright-core/lib/utils';
 
+
```
mxschmitt, line 23:
```suggestion
```

packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1166,7 +1166,7 @@ export class InjectedScript {
           throw this.createStacklessError('Element is not a checkbox');
         if (elementState === 'error:notconnected')
           throw this.createStacklessError('Element is not connected');
-        return { received: elementState, matches: elementState };
+        return { matches: elementState };
```
mxschmitt, line 1169: seems like this was unused or at least not covered with a test?

packages/playwright/src/matchers/toMatchText.ts
```diff
@@ -69,11 +69,11 @@ export async function toMatchText(
       typeof expected === 'string'
         ? matcherHint(this, receiver, matcherName, 'locator', undefined, matcherOptions, timedOut ? timeout : undefined) +
         `Expected ${stringSubstring}: not ${this.utils.printExpected(expected)}\n` +
-        `Received string: ${printReceivedStringContainExpectedSubstring(
+        `Received string: ${receivedString.indexOf(expected) !== -1 ? printReceivedStringContainExpectedSubstring(
```
mxschmitt, line 72: `printReceivedStringContainExpectedSubstring` behaves weirdly if you pass:

```
 receivedString = something 
receivedString.indexOf(expected) = -1,
expected.length = something long
```

to it. Probably because it was intended only for the matching scenario.

dgozman: - Might as well do `Received:` instead of `Received string:`
- We should cover the regex case below as well.

packages/playwright-core/src/server/injected/injectedScript.ts
dgozman: This sounds strange. If it's not used in Node.js, it is probably used in other languages.

packages/playwright-core/src/server/frames.ts
```diff
@@ -1382,6 +1382,8 @@ export class Frame extends SdkObject {
     // Library mode special case for the expect errors which are return values, not exceptions.
     if (result.matches === options.isNot)
       metadata.error = { error: { name: 'Expect', message: 'Expect failed' } };
+    if (result.log?.[result.log.length - 1].startsWith('waiting for '))
```
dgozman, line 1385: Instead of relying on the logs, can we do this at the place where we actually know that element was not found?

packages/playwright-core/src/server/injected/injectedScript.ts
mxschmitt: will keep it for now.

tests/page/matchers.misc.spec.ts
```diff
@@ -23,3 +24,41 @@ it('should outlive frame navigation', async ({ page, server }) => {
   }, 1000);
   await expect(page.locator('.box').first()).toBeEmpty();
 });
+
+it('should print no-locator-resolved error when locator matcher did not resolve to any element', async ({ page, server }) => {
+  const myLocator = page.locator('.nonexisting');
+  const expectWithShortLivingTimeout = expect.configure({ timeout: 10 });
+  const locatorMatchers = [
```
dgozman, line 31: This test is not checking `.not.toHaveText()` that has changed extensively in the code. I'd recommend trimming this list to just a few, but add `.not` and any other interesting variants.

packages/playwright/src/matchers/toMatchText.ts
```diff
@@ -64,39 +64,30 @@ export async function toMatchText(
...
+  const notFound = received === kNoElementsFoundError;
+  const message = () => {
+    if (pass) {
+      if (typeof expected === 'string') {
+        if (notFound)
+          return messagePrefix + `Expected ${stringSubstring}: not ${this.utils.printExpected(expected)}\nReceived: ` + received + callLogText(log);
+        const printedReceived = printReceivedStringContainExpectedSubstring(receivedString, receivedString.indexOf(expected), expected.length);
+        return messagePrefix + `Expected ${stringSubstring}: not ${this.utils.printExpected(expected)}\nReceived string: ${printedReceived}` + callLogText(log);
+      } else {
+        if (notFound)
```
dgozman, line 77: It seems like you can move `notFound` case outside of both branches.

mxschmitt: The expected label seems different.

tests/page/matchers.misc.spec.ts
mxschmitt: Would you keep: one boolean matcher and one text matcher for each branch?

dgozman: Yeah, we should cover every branch, those are easy to mess up.

tests/page/matchers.misc.spec.ts
```diff
@@ -23,3 +24,23 @@ it('should outlive frame navigation', async ({ page, server }) => {
...
+    () => expectWithShortLivingTimeout(myLocator).toBeAttached(),                 // Boolean matcher
+    () => expectWithShortLivingTimeout(myLocator).toHaveJSProperty('abc', 'abc'), // Equal matcher
+    () => expectWithShortLivingTimeout(myLocator).not.toHaveText('abc'),          // Text matcher - pass / string
+    () => expectWithShortLivingTimeout(myLocator).not.toHaveText(/abc/),          // Text matcher - pass / RegExp
+    () => expectWithShortLivingTimeout(myLocator).toContainText('abc'),           // Text matcher - fail
```
dgozman, line 36: Perhaps test both fail/string and fail/RegExp?

tests/page/matchers.misc.spec.ts
```diff
@@ -23,3 +24,23 @@ it('should outlive frame navigation', async ({ page, server }) => {
...
+  ];
+  for (const matcher of locatorMatchers) {
+    await it.step(matcher.toString(), async () => {
+      const error = await matcher().catch(e => e);
+      expect(error).toBeInstanceOf(Error);
+      expect(error.message).toContain(`waiting for locator('.nonexisting')`);
+      expect(stripAnsi(error.message)).toMatch(/Received:  ?"?<element\(s\) not found>/);
```
dgozman, line 43: Now that we only have 6 cases, maybe inlining them with exact expected error message would be easier to understand?

dgozman (approved):

tests/page/matchers.misc.spec.ts
mxschmitt: (Same branch, but sure added it)

tests/page/matchers.misc.spec.ts
mxschmitt: Not a huge fan, since it would make the test much larger. This underlines our motivation that the `waiting for locator` and `Received` value should be displayed in all the scenarios.

PR #29902, opened 2024-03-12
fix(core): ElectronApplication.evaluate arg wrong type

james-tindal:
The first argument to the function passed to `evaluate` should be `Electron.RemoteMainInterface`, not `typeof Electron.CrossProcessExports`

If keeping `typeof import('electron')` is important it should be changed to:
```ts
type ElectronType = typeof import('electron');
type Remote = ElectronType.RemoteMainInterface;
```

### DIFF
--- a/packages/playwright-core/types/types.d.ts
+++ b/packages/playwright-core/types/types.d.ts
@@ -13855,7 +13855,7 @@ type AccessibilityNode = {
 export const devices: Devices;
 
 //@ts-ignore this will be any if electron is not installed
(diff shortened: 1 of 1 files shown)


james-tindal:
Thanks. I have done that.
Turns out `RemoteMainInterface` no longer exists in the latest version of electron and I don't think it was the right solution anyway.

mxschmitt:
Closing as per https://github.com/microsoft/playwright/issues/29912 since it seems right what we are doing.

PR #29904, opened 2024-03-12
Document more default config values

karlhorky:
Hi, first of all, thanks for the continued work in the Playwright ecosystem, amazing!

Quick PR to document a few more undocumented default config values, on these docs pages (hope I have the correct file, I didn't see a straightforward way to edit the docs from these pages):

- https://playwright.dev/docs/api/class-testconfig#test-config-forbid-only
- https://playwright.dev/docs/api/class-testconfig#test-config-fully-parallel

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -76,7 +76,8 @@ export default defineConfig({
 * since: v1.10
 - type: ?<[boolean]>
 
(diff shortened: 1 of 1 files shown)


karlhorky:
Ok, thanks for the answer.

Respectfully disagree here - in my experience, most well-documented projects in the JS/TS ecosystems talk about defaults with each of their options. The fact that defaults of other Playwright options are documented makes it even harder to figure out why some are not documented or to guess the defaults.

PR #29918, opened 2024-03-13
fix: postDataJSON without Content-Type header

mxschmitt:
Regressed after https://github.com/microsoft/playwright/commit/38fc74db7c24398095632fa2e65e913c80df99f4.

Test coverage: Some tests were failing on the flakiness dashboard

### DIFF
--- a/packages/playwright-core/src/client/network.ts
+++ b/packages/playwright-core/src/client/network.ts
@@ -141,7 +141,7 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
       return null;
 
     const contentType = this.headers()['content-type'];
-    if (contentType.includes('application/x-www-form-urlencoded')) {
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/client/network.ts
```diff
@@ -141,7 +141,7 @@ export class Request extends ChannelOwner<channels.RequestChannel> implements ap
       return null;
 
     const contentType = this.headers()['content-type'];
-    if (contentType.includes('application/x-www-form-urlencoded')) {
+    if (contentType?.includes('application/x-www-form-urlencoded')) {
```
yury-s, line 144: nit: to be even more accurate, I'd split the content type by `;` and then check if there is the mime type.

yury-s (approved):

PR #29921, opened 2024-03-13
fix(reuse): reset Origin Private File System API

### LINKED ISSUE #29901: [Bug]: Origin Private File System is not isolated between tests
### Version

1.42.1

### Steps to reproduce

1. Clone https://github.com/karyon/playwright-examples
2. git checkout opfs-should-be-cleared
3. In VSCode extension, enable "Show Browser"
4. Run `opfs` test in opfs-should-be-cleared.spec.ts
5. Without closing the browser, run it again

### Expected behavior

It succeeds twice

### Actual behavior

It fails the second time

### Additional context

https://playwright.dev/docs/browser-contexts uses very strong wording: "Every test runs independently from any other test. This means that each test has its own local storage, session storage, cookies etc. ... equivalent to incognito-like profiles ... completely isolated, even when running in a single browser."

This does not seem to apply to the [Origin Private File System ](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system). The test in the example repo checks whether a directory exists and if not, creates it. The second run sees the directory created by the first one. For comparison, I wrote a similar test with localStorage, where it works as expected.

### Environment

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
    Memory: 6.92 GB / 31.76 GB
  Binaries:
    Node: 18.17.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.6.7 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.15.4 - C:\Program Files\nodejs\pnpm.CMD
  IDEs:
    VSCode: 1.85.2 - C:\Program Files\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 5.2.26 - C:\Program Files\Git\usr\bin\bash.EXE
  npmPackages:
    @playwright/test: ^1.42.1 => 1.42.1
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/29901

This clears the Origin Private File System API when using VSCode extension.

### DIFF
--- a/tests/playwright-test/playwright.reuse.spec.ts
+++ b/tests/playwright-test/playwright.reuse.spec.ts
@@ -212,7 +212,7 @@ test('should clean storage', async ({ runInlineTest }) => {
       let lastContextGuid;
 
       test.beforeEach(async ({ page }) => {
-        await page.route('**/*', route => route.fulfill('<html></html>'));
+        await page.route('**/*', route => route.fulfill({ body: '<html></html>', contentType: 'text/html' }));
         await page.goto('http://example.com');
       });
 
@@ -273,7 +273,7 @@ test('should restore localStorage', async ({ runInlineTest }) => {
       });
 
       test.beforeEach(async ({ page }) => {
-        await page.route('**/*', route => route.fulfill('<html></html>'));
+        await page.route('**/*', route => route.fulfill({ body: '<html></html>', contentType: 'text/html' }));
         await page.goto('http://example.com');
       });
 
@@ -330,7 +330,7 @@ test('should clean db', async ({ runInlineTest }) => {
       let lastContextGuid;
 
       test.beforeEach(async ({ page }) => {
-        await page.route('**/*', route => route.fulfill('<html></html>'));
+        await page.route('**/*', route => route.fulfill({ body: '<html></html>', contentType: 'text/html' }));
         await page.goto('http://example.com');
       });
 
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/server/chromium/crPage.ts
```diff
@@ -369,6 +369,7 @@ export class CRPage implements PageDelegate {
   }
 
   async resetForReuse(): Promise<void> {
+    await this._mainFrameSession._client.send('Storage.clearDataForOrigin', { origin: '*', storageTypes: 'all' });
```
mxschmitt, line 372: Note: `file_systems` would work as well, but I feel like with that we are on the safer side.

pavelfeldman:
As agreed offline, let's do web api-based solution.

tests/playwright-test/playwright.reuse.spec.ts
```diff
@@ -212,7 +212,7 @@ test('should clean storage', async ({ runInlineTest }) => {
       let lastContextGuid;
 
       test.beforeEach(async ({ page }) => {
-        await page.route('**/*', route => route.fulfill('<html></html>'));
+        await page.route('**/*', route => route.fulfill({ body: '<html></html>', contentType: 'text/html' }));
```
mxschmitt, line 215: drive-by: it was never working before.

pavelfeldman (approved):

PR #29951, opened 2024-03-15
#29931 - Modified the docs to ble clearer on what is visible or not v…

thorleifjacobsen:


pavelfeldman (commented):
I like the existing one better.

thorleifjacobsen:
> I like the existing one better.

The "are not" and "are" are a bit hard to read for people with dyslexia and are almost "hidden" in a sense. Easy to misread it. Should clearly categorize them more. Espescially when you train the brain to read "are not" two times then add a "are" without the not.

pavelfeldman:
I don't think that we can efficiently improve our docs to be optimized for dyslexia in a non-coordinated manner. There is a lot of text above this list that is much harder to read and I find bold emphasis to address the 'are' vs 'are not' well enough.

PR #29965, opened 2024-03-15
chore: prepare to reuse test server from ui mode

pavelfeldman:


packages/playwright-core/src/utils/httpServer.ts
```diff
@@ -68,6 +77,27 @@ export class HttpServer {
...
 
+  createWebSocket(transport: Transport, guid?: string) {
+    assert(!this._wsGuid, 'can only create one main websocket transport per server');
+    this._wsGuid = guid || createGuid();
+    const wss = new wsServer({ server: this._server, path: '/' + this._wsGuid });
+    wss.on('connection', ws => {
+      transport.sendEvent = (method, params)  => ws.send(JSON.stringify({ method, params }));
+      transport.close = () => ws.close();
+      ws.on('message', async (message: string) => {
+        const { id, method, params } = JSON.parse(message);
```
dgozman, line 88: `try/catch` this whole function?

dgozman (approved):

packages/playwright-core/src/utils/httpServer.ts
pavelfeldman: I'll follow up.

PR #29966, opened 2024-03-15
chore: prepare to reuse test server from ui mode (2)

pavelfeldman:


packages/playwright/src/runner/uiMode.ts
```diff
@@ -142,20 +99,71 @@ class TestServer {
...
+      return runner.listTestFiles();
+    } catch (e) {
+      const error: TestError = serializeError(e);
+      error.location = prepareErrorStack(e.stack).location;
+      return { projects: [], error };
+    }
+  }
+
+  async listTests(params: { reporter?: string; fileNames: string[]; }) {
+    this._queue = this._queue.then(() => this._innerListTests(params));
```
dgozman, line 160: When `_innerListTests` throws, any subsequent operations in the queue will never run. Do we want this behavior? If not, perhaps introduce a `safeEnqueue` wrapper or catch right here? The same applies to all other `_queue` usage.

dgozman (approved):

PR #29971, opened 2024-03-16
fix(types): evaluate should not unpack return unions

mxschmitt:
Motivation: https://github.com/nuxt/nuxt/blob/977377777afcc38186c7a309a86e9e5613470652/test/basic.test.ts#L1340

### DIFF
--- a/utils/generate_types/test/test.ts
+++ b/utils/generate_types/test/test.ts
@@ -331,9 +331,9 @@ playwright.chromium.launch().then(async browser => {
   }
   // evaluteHandle with two different return types (ElementHandle)
   {
-    const handle = await page.evaluateHandle(() => '' as HTMLInputElement | HTMLTextAreaElement);
+    const handle = await page.evaluateHandle(() => '' as any as HTMLInputElement | HTMLTextAreaElement);
     await handle.evaluate(element => element.value);
-    const assertion: AssertType<ElementHandle<HTMLInputElement | HTMLTextAreaElement>, typeof handle> = true;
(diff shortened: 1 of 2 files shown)


utils/generate_types/test/test.ts
```diff
@@ -323,6 +323,13 @@ playwright.chromium.launch().then(async browser => {
   console.log(await resultHandle.jsonValue());
   await resultHandle.dispose();
 
+  // evaluteHandle with two different return types
+  {
+    const handle = await page.evaluateHandle(() => '' as string | number);
+    const result = await handle.evaluate(value => value);
```
mxschmitt, line 329: evaluate was not passing TS before since handle was 'JSHandle<boolean> |JSHandle<number>' instead of 'JSHandle<boolean | number>'.

utils/generate_types/test/test.ts
```diff
@@ -323,6 +323,13 @@ playwright.chromium.launch().then(async browser => {
   console.log(await resultHandle.jsonValue());
   await resultHandle.dispose();
 
+  // evaluteHandle with two different return types
+  {
+    const handle = await page.evaluateHandle(() => '' as string | number);
```
dgozman, line 328: Let's add a case where with `as HTMLInputElement | HTMLTextAreaElement` and check that result is an `ElementHandle<HTMLInputElement | HTMLTextAreaElement>`?

dgozman (approved):

basal-john:
Thanks for the fix :)

PR #30004, opened 2024-03-19
docs: note on locator handler trigger

### LINKED ISSUE #30003: [Docs] Improve documentation for addLocatorHandler
### Page(s)

https://playwright.dev/docs/api/class-page#page-add-locator-handler

### Description

So I hope I'll not be hurting your soul too much with that one, but I really need to trigger the locator handler at an arbitrary time without doing anything else (I need to be able to bypass the cookies authorization on arbitrary websites).

My workaround is to run do that: 

`await page.locator("body").click(button="right")`

Which does nothing, but triggers the locator handlers whenever I want them triggered.

The full commit looks like that:

https://github.com/Lookyloo/PlaywrightCapture/commit/26348f7bf5f28f7d6c6a6d48cb41fe77ed99f6f5

Maybe two things:

1. Add in the documentation that the locator handlers are only checked when there is an action, it wasn't clear to me before I found this bug https://github.com/microsoft/playwright/issues/29701
2. Add a way to forcefully trigger the locator handlers?

yury-s:
Fixes https://github.com/microsoft/playwright/issues/30003

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -3156,7 +3156,7 @@ This method lets you set up a special function, called a handler, that activates
 
 Things to keep in mind:
 * When an overlay is shown predictably, we recommend explicitly waiting for it in your test and dismissing it as a part of your normal test flow, instead of using [`method: Page.addLocatorHandler`].
-* Playwright checks for the overlay every time before executing or retrying an action that requires an [actionability check](../actionability.md), or before performing an auto-waiting assertion check. When overlay is visible, Playwright calls the handler first, and then proceeds with the action/assertion.
+* Playwright checks for the overlay every time before executing or retrying an action that requires an [actionability check](../actionability.md), or before performing an auto-waiting assertion check. When overlay is visible, Playwright calls the handler first, and then proceeds with the action/assertion. Note that the handler is called only when you perform an action and the locator is visible, if the overlay becomes visible but you don't call any action, the hadler will not be triggered.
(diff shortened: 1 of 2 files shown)


docs/src/api/class-page.md
```diff
@@ -3156,7 +3156,7 @@ This method lets you set up a special function, called a handler, that activates
 
 Things to keep in mind:
 * When an overlay is shown predictably, we recommend explicitly waiting for it in your test and dismissing it as a part of your normal test flow, instead of using [`method: Page.addLocatorHandler`].
-* Playwright checks for the overlay every time before executing or retrying an action that requires an [actionability check](../actionability.md), or before performing an auto-waiting assertion check. When overlay is visible, Playwright calls the handler first, and then proceeds with the action/assertion.
+* Playwright checks for the overlay every time before executing or retrying an action that requires an [actionability check](../actionability.md), or before performing an auto-waiting assertion check. When overlay is visible, Playwright calls the handler first, and then proceeds with the action/assertion. Note that the handler is called only when you perform an action and the locator is visible, if the overlay becomes visible but you don't call any action, the hadler will not be triggered.
```
dgozman, line 3159:
```suggestion
* Playwright checks for the overlay every time before executing or retrying an action that requires an [actionability check](../actionability.md), or before performing an auto-waiting assertion check. When overlay is visible, Playwright calls the handler first, and then proceeds with the action/assertion. Note that the handler is only called when you perform an action/assertion - if the overlay becomes visible but you don't perform any actions, the handler will not be triggered.
```

dgozman (approved):

PR #30013, opened 2024-03-20
feat: support importing client-only files (like scss) for ESM

### LINKED ISSUE #30153: [Feature]: Support importing client files like SCSS on ESM
### 🚀 Feature Request

Related issue: #26822 (doesn't close it, being able to provide your own loader would still be cool and needed)

Currently, if you import client extension files like scss within playwright and use ESM, it throws an `"Unknown file extension: .scss"` error. This PR fixes this issue and allows playwright to support importing those out-of-the-box.

This is unfortunate, because our own webpack loader is indeed able to import and bundle those scss files. Back when we used CJS, we were able to skip requiring client files using [this simple trick](https://github.com/payloadcms/payload/blob/1c0d43c61a0e29bc9de50a4e608cb7dff9d5cd48/packages/payload/src/config/load.ts#L21).

With ESM imports, this is no longer possible and there is no way for us to get playwright to support it. The "ESM-way" to do this is to use a custom loader which skips resolving those client files and loads them as raw JSON. That way, they won't error out and a bundler can then handle them accordingly. This is what this PR does.

Currently, we are patching playwright in order to get playwright to actually work in our project: https://github.com/payloadcms/payload/blob/alpha/patches/playwright%401.42.1.patch.

I'm aware the alternative to this is to either compile our project beforehand and run playwright on `dist`, or to run next.js in a separate process and make sure no client files are imported within playwright itself. This, however, is a very bad developer experience for us, especially since we cannot use a debugger that way.

### Example

_No response_

### Motivation

This will allow people to keep importing file which import SCSS within playwright when using ESM, without having to give up on debugging support

PR for this issue (needs more votes): https://github.com/microsoft/playwright/pull/30013#issuecomment-2023392184

AlessioGr:
Issue this fixes: https://github.com/microsoft/playwright/issues/30153

Related issue: #26822 (doesn't close it, being able to provide your own loader would still be cool and needed)

Currently, if you import client extension files like scss within playwright and use ESM, it throws an `"Unknown file extension: .scss"` error. This PR fixes this issue and allows playwright to support importing those out-of-the-box.

This is unfortunate, because our own webpack loader is indeed able to import and bundle those scss files. Back when we used CJS, we were able to skip requiring client files using [this simple trick](https://github.com/payloadcms/payload/blob/1c0d43c61a0e29bc9de50a4e608cb7dff9d5cd48/packages/payload/src/config/load.ts#L21).

With ESM imports, this is no longer possible and there is no way for us to get playwright to support it. The "ESM-way" to do this is to use a custom loader which skips resolving those client files and loads them as raw JSON. That way, they won't error out and a bundler can then handle them accordingly. This is what this PR does.

Currently, we are patching playwright in order to get playwright to actually work in our project: https://github.com/payloadcms/payload/blob/alpha/patches/playwright%401.42.1.patch. Getting client file imports to work is one out of three things we had to patch in order for playwright to support ESM. This PR will relieve some burden for us to patch it after each version, until hopefully in the future this is not needed at all anymore.

I'm aware the alternative to this is to either compile our project beforehand and run playwright on `dist`, or to run next.js in a separate process and make sure no client files are imported within playwright itself. This, however, is a very bad developer experience for us, especially since we cannot use a debugger that way.

### DIFF
--- a/packages/playwright/src/transform/clientExtensions.ts
+++ b/packages/playwright/src/transform/clientExtensions.ts
@@ -1,0 +1,11 @@
+export const CLIENT_EXTENSIONS = [
+  '.scss',
+  '.css',
+  '.svg',
+  '.png',
+  '.jpg',
+  '.eot',
+  '.ttf',
+  '.woff',
+  '.woff2',
+]
--- a/packages/playwright/src/transform/clientExtensions.ts
+++ b/packages/playwright/src/transform/clientExtensions.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 export const CLIENT_EXTENSIONS = [
   '.scss',
   '.css',
@@ -8,4 +24,4 @@ export const CLIENT_EXTENSIONS = [
   '.ttf',
   '.woff',
   '.woff2',
-]
+];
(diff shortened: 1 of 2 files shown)


AlessioGr:
@microsoft-github-policy-service agree [company="Payload CMS, Inc."]

AlessioGr:
@microsoft-github-policy-service agree company="Payload CMS, Inc."

pavelfeldman:
> I'm aware the alternative to this is to either compile our project beforehand and run playwright on dist, or to run next.js in a separate process and make sure no client files are imported within playwright itself. This, however, is a very bad developer experience for us, especially since we cannot use a debugger that way.

This would still be our recommendation. So far your issue did not collect the number of votes for us to consider the PR.

PR #30015, opened 2024-03-20
fix: add missing createElement stub for @babel/plugin-transform-react-jsx plugin

### LINKED ISSUE #26824: [BUG] The requested module '@playwright/test' does not provide an export named 'createElement'
### System info
- Playwright Version: 1.37.1 (patched with custom loader to support esm, see https://github.com/microsoft/playwright/issues/26822. I don't think the loader is causing this particular issue, as it only handles svg/png/scss files.)
- Operating System: macOS 14.0
- Browser: any
- Other info:

### Source code

https://github.com/payloadcms/payload/tree/2.0-esm/packages/payload/test/refresh-permissions

**Steps**
1. Clone the [payload repository](https://github.com/payloadcms/payload/tree/2.0-esm/packages/payload/test/refresh-permissions) (`2.0-esm` branch)
2. Run `pnpm install`
3. Run `pnpm test:e2e refresh-permissions`

**Expected**

The test passes

**Actual**

The test fails with the following error:
Screenshot 2023-08-31 at 19 45 56@2x

The issue is triggered by returning the `DefaultGlobalView` component in the [GlobalViewWithRefresh.tsx](https://github.com/payloadcms/payload/blob/2.0-esm/packages/payload/test/refresh-permissions/GlobalViewWithRefresh.tsx) component included in the test. If I return anything else, e.g. a paragraph, the issue does not happen.

My hypothesis is that playwright, for some reason, compiles the component down into nonsense where somehow `createElement` is imported from `@playwright/test` instead of from `React`. Running the test directory directly (`pnpm dev refresh-permissions`) which doesn't use playwright works fine.

This has been happening ever since we migrated payload from cjs+yarn to esm+pnpm+monorepo.

AlessioGr:
Fixes #26824 

## Issue explained

The default babel transform options include the `@babel/plugin-transform-react-jsx` plugin. With this plugin installed, we are encountering an issue in our environment which doesn't seem to appear in all environments. Look at [this specific code](https://github.com/babel/babel/blob/fb7cf6343704731802c2dac18999990d7d4ce484/packages/babel-plugin-transform-react-jsx/src/create-plugin.ts#L801) of `@babel/plugin-transform-react-jsx`

CleanShot 2024-03-19 at 22 42 57@2x

It makes it so, for example, jsx is imported from `source`/jsx-runtime. `source` being playwright. You can see that in the playwright compilation cache here:
CleanShot 2024-03-19 at 22 43 43

This (2nd cached import in screenshot) works, because playwright has a` /jsx-runtime` export [here](https://github.com/microsoft/playwright/blob/92ad1b385c7be5d952e4aa2a94a32d083ba5b9a3/packages/playwright/jsx-runtime.js).

But it breaks for createElement (3rd cached import in screenshot). For createElement, that babel plugin only returns `source` (see source code in screenshot above), which would be `/playwright`. And `/playwright` (the root index.js) does not contain a stub createElement. This causes the following issue:

```
SyntaxError: The requested module "/Users/alessio/Documents/GitHub/payload/node_modules/.pnpm/playwright@1.42.1_patch_hash=snj23idnz6qktfxfcravfxvpgu/node_modules/playwright' does not provide an export named 'createElement'
Error: No tests found.
Make sure that arguments are regular expressions matching test files.
You may need to escave svmbols like "s" or "*" and auote the arguments.#
```

This issue probably starts to appear with certain tsconfig options. I don't know what those are, but it happens in [payload](https://github.com/payloadcms/payload/tree/alpha).

## What this PR does
- moves the `jsx-runtime.js` and `jsx-runtime.mjs` files into its own dedicated `babel-react-transform` folder. This declutters the root and clears up confusion as for why jsx-runtime.js is even needed. As a newcomer to the playwright code-base (or even just looking at the error message mentioning this file) I was like "What the hell is that file even for?". Now, it's clear that it's needed for the babel react transform plugin.
- adds an index.js with a createElement stub export. This is what `@babel/plugin-transform-react-jsx` expects. If it cannot find it, it throws an error in certain configurations


pavelfeldman:
Closing as stale.

Issue #29899, opened 2024-03-12
[Bug]: When starting from VSCode extension, 'devtools: true' in launchOptions does nothing


karyon:
### Version

1.42.1

### Steps to reproduce

1. clone https://github.com/karyon/playwright-examples
2. git checkout repro-devtools-doing-nothing
3. npm install, npx playwright install
4. Using the VSCode extension, run any test with "show browser" enabled

### Expected behavior

Browser opens and has devtools open

### Actual behavior

Browser opens but does not have devtools open

### Additional context

[browserType.launch() docs](https://playwright.dev/docs/api/class-browsertype#browser-type-launch)
[launchOptions docs](https://playwright.dev/docs/test-use-options#more-browser-and-context-options) say any browserType.launch() option ([docs](https://playwright.dev/docs/api/class-browsertype#browser-type-launch)) can be used
In https://github.com/microsoft/playwright/issues/28529#issuecomment-1845009195 someone is reporting this is working for them
In playwright.config.ts, uncomment `args: ['--auto-open-devtools-for-tabs']`, this does work.

### Environment

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
    Memory: 6.92 GB / 31.76 GB
  Binaries:
    Node: 18.17.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.6.7 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.15.4 - C:\Program Files\nodejs\pnpm.CMD
  IDEs:
    VSCode: 1.85.2 - C:\Program Files\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 5.2.26 - C:\Program Files\Git\usr\bin\bash.EXE
  npmPackages:
    @playwright/test: ^1.42.1 => 1.42.1
```

mxschmitt:
Investigation: The `devtools: true` option does not get taken into account when using the VSCode extension.

When running the tests via the CLI, it works: `npx playwright test --project=chromium --headed api-mocking:5`.

karyon:
Ah I forgot! When using  `chromium.launch` in test code, it works too.

karyon:
[...]

PR #30025, opened 2024-03-20
feat: deprecate "devtools" launch option

dgozman:
References #29899.


mxschmitt (approved):

odinho:
Hm, why this change? I don't see how anything on that /debug page replaces this option.

Some people may use the VS Code integration, but that doesn't actually debug the running code on the page? It's just like the Playwright inspector which lets you step through the actual test file, but that's not why I like auto-opening the devtools.

Auto-opening the browser devtools is very useful to not have to click around to get it open when debugging an issue. :)

PR #30031, opened 2024-03-20
chore: remove ui mode update globals

pavelfeldman:


packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -83,62 +79,129 @@ export const UIModeView: React.FC<{}> = ({
...
-      const { hasBrowsers } = await connection.checkBrowsers();
-      setHasBrowsers(hasBrowsers);
-    })();
+    setTestServerConnection(new TestServerConnection(wsURL.toString()));
   }, []);
 
+  // Load tests on startup.
+  React.useEffect(() => {
+    inputRef.current?.focus();
+    setIsLoading(true);
```
dgozman, line 88: Shouldn't `reloadTests()` set loading to true itself, so that all other callers have it automatically? Or perhaps we already do this in "the main routine", so can just remove from here?

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -83,62 +79,129 @@ export const UIModeView: React.FC<{}> = ({
...
 
+
+  // This is the main routine, every time connection updates it starts the
+  // whole workflow.
   React.useEffect(() => {
-    inputRef.current?.focus();
+    if (!testServerConnection)
+      return;
+
+    let throttleTimer: NodeJS.Timeout | undefined;
```
dgozman, line 124: Clear this timer in `useEffect`'s dispose callback?

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -83,62 +79,129 @@ export const UIModeView: React.FC<{}> = ({
...
+    (async () => {
+      const status = await testServerConnection.runGlobalSetup();
+      if (status !== 'passed')
+        return;
+      const result = await testServerConnection.listTests({});
+      teleSuiteUpdater.processListReport(result.report);
+
+      testServerConnection.onListChanged(updateList);
+      testServerConnection.onReport(params => {
+        teleSuiteUpdater.processTestReport(params);
```
dgozman, line 162: nit: perhaps `processTestReportEvent` would be better, to differentiate with `processListReport`?

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -83,62 +79,129 @@ export const UIModeView: React.FC<{}> = ({
...
+  }, [projectFilters, testModel]);
 
-    setTestModel({ config, rootSuite, loadErrors });
-    setProjectFilters(new Map(projectFilters));
-    if (runningState && newProgress)
-      setProgress(newProgress);
-    else if (!newProgress)
+  // Update progress.
+  React.useEffect(() => {
+    if (runningState && testModel?.progress)
```
dgozman, line 195: I don't really understand this logic, but I guess it's the same as before...

dgozman (approved):

packages/trace-viewer/src/ui/uiModeView.tsx
pavelfeldman: This is to give it a more "immediate" feedback.

packages/trace-viewer/src/ui/uiModeView.tsx
pavelfeldman: No change here.

PR #30047, opened 2024-03-21
chore: bump TypeScript to 5.4.3

mxschmitt:


mxschmitt:
Waiting for 5.5, there is a ttest typing regression which is fixed there already.

PR #30061, opened 2024-03-22
devops(merge): validate zip before uploading

mxschmitt:
Follow-up on https://github.com/microsoft/playwright/pull/30060.

manpage of unzip for what `-t` does:

    test archive files.  This option extracts each specified file in memory and compares the CRC (cyclic redundancy check, an enhanced checksum) of the expanded file with the original file's stored CRC value.

### DIFF
--- a/.github/actions/upload-blob-report/action.yml
+++ b/.github/actions/upload-blob-report/action.yml
@@ -12,6 +12,12 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Integrity check
+      shell: bash
+      run: |
+        for file in ${{ inputs.report_dir }}/*.zip; do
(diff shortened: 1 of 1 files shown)


.github/actions/upload-blob-report/action.yml
```diff
@@ -12,6 +12,12 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Integrity check
+      shell: bash
+      run: |
+        for file in ${{ inputs.report_dir }}/*.zip; do
```
yury-s, line 18:
```suggestion
        for file in ${{ inputs.report_dir }}/**/*.zip; do
```

yury-s (approved):

PR #30071, opened 2024-03-23
fix(firefox): disable cache for request interception

### LINKED ISSUE #30000: [Bug]: page.route does not disable http cache in Firefox
### Version

1.41.0

### Steps to reproduce

https://try.playwright.tech/?l=javascript&s=lxhfadd

```js
// @ts-check

(async () => {
    const browser = await playwright.firefox.launch();

    // Route everything to disable the cache (this doesn't work)
    // "Enabling routing disables http cache." per https://playwright.dev/docs/api/class-page#page-route
    await page.route('**/*', route => route.continue());

    // Go to the homepage
    await page.goto('https://www.digitalocean.com');

    // Set a fast timeout, as we're expecting timeouts for this issue reproduction
    page.setDefaultTimeout(1500);

    for (let i = 0; i < 5; i++) {
        await Promise.all([
            page.waitForResponse(resp => {
                // Look for the API request we're about to make
                if (!resp.url().includes('/api/static-content/v1/tags/ubuntu')) return;

                // Log the response we got, including the status that Playwright sees
                console.log(i, 'playwright', resp.status());

                // We want to wait for a successful response
                return resp.ok();
            }).catch(e => {
                // Log a timeout but don't crash
                console.log(i, e.message);
            }),
            // Make the API request and log the status that the browser sees
            page.evaluate('fetch("/api/static-content/v1/tags/ubuntu").then(resp => resp.status)').then(status => console.log(i, 'browser', status)),
        ]);
    }

    // Allow some time for final logs to flush before closing
    await new Promise(res => setTimeout(res, 500));
    await browser.close();
})();
```

### Expected behavior

https://playwright.dev/docs/api/class-page#page-route states that when `page.route` is used, the HTTP cache should be disabled. As such, I'd expect `resp.status()` to return `200` every time, or the documentation to be updated to reflect that caching is not disabled in Firefox.

### Actual behavior

On the first (or first few depending on how Firefox feels), the `resp.status()` is observed to be `200`, however it then gets cached by Firefox and the `resp.status()` switches to being a `304` (while the status that the browser fetch call sees is still a `200`).

### Additional context

I believe this is somewhat related to https://github.com/microsoft/playwright/issues/27573

### Environment

```shell
Unsure how to get this info from https://try.playwright.tech
```

aslushnikov:
Fixes #30000

### DIFF
--- a/packages/playwright-core/src/server/firefox/ffNetworkManager.ts
+++ b/packages/playwright-core/src/server/firefox/ffNetworkManager.ts
@@ -50,7 +50,10 @@ export class FFNetworkManager {
   }
 
   async setRequestInterception(enabled: boolean) {
-    await this._session.send('Network.setRequestInterception', { enabled });
+    await Promise.all([
+      this._session.send('Page.setCacheDisabled', { cacheDisabled: enabled }),
+      this._session.send('Network.setRequestInterception', { enabled }),
+    ]);
   }
 
   _onRequestWillBeSent(event: Protocol.Network.requestWillBeSentPayload) {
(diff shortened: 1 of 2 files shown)


yury-s (approved):
lgtm provided it works in popups too. Perhaps we need a test for that.

aslushnikov:
ouch, doesn't work in popups! My bad

PR #30094, opened 2024-03-25
Html report improvements

amankagithub:
1. Copy annotation text by one click .
2. Search test case based on annotations .

### DIFF
--- a/packages/html-reporter/src/filter.ts
+++ b/packages/html-reporter/src/filter.ts
@@ -120,7 +120,7 @@ export class Filter {
         line: String(test.location.line),
         column: String(test.location.column),
         labels: test.tags.map(tag => tag.toLowerCase()),
-        annotations: test.annotations.map(a => a.type.toLowerCase() + ':' + a.description?.toLocaleLowerCase())
+        annotations: test.annotations.map(a => a.type.toLowerCase() + '=' + a.description?.toLocaleLowerCase())
       };
       (test as any).searchValues = searchValues;
     }
--- a/packages/html-reporter/src/filter.ts
+++ b/packages/html-reporter/src/filter.ts
@@ -15,12 +15,12 @@
 */
 
 import type { TestCaseSummary } from './types';
-
 export class Filter {
   project: string[] = [];
   status: string[] = [];
   text: string[] = [];
   labels: string[] = [];
+  annotations: string[] = [];
 
   empty(): boolean {
     return this.project.length + this.status.length + this.text.length === 0;
@@ -32,6 +32,7 @@ export class Filter {
     const status = new Set<string>();
     const text: string[] = [];
     const labels = new Set<string>();
+    const annotations = new Set<string>();
     for (const token of tokens) {
       if (token.startsWith('p:')) {
         project.add(token.slice(2));
@@ -45,6 +46,10 @@ export class Filter {
         labels.add(token);
         continue;
       }
+      if (token.startsWith('a:')) {
+        annotations.add(token.slice(2));
+        continue;
+      }
       text.push(token.toLowerCase());
     }
 
@@ -53,6 +58,7 @@ export class Filter {
     filter.project = [...project];
     filter.status = [...status];
     filter.labels = [...labels];
+    filter.annotations = [...annotations];
     return filter;
   }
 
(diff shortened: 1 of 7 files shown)


pavelfeldman:
Please start with filing a feature request, even if you plan to contribute this PR.

packages/html-reporter/src/filter.ts
```diff
@@ -114,6 +120,7 @@ export class Filter {
         line: String(test.location.line),
         column: String(test.location.column),
         labels: test.tags.map(tag => tag.toLowerCase()),
+        annotations: test.annotations.map(a => a.type.toLowerCase() + ':' + a.description?.toLocaleLowerCase())
```
pavelfeldman, line 123: To make sure you only use : once in the filter.

```suggestion
        annotations: test.annotations.map(a => a.type.toLowerCase() + '=' + a.description?.toLocaleLowerCase())
```

packages/html-reporter/src/filter.ts
```diff
@@ -144,7 +151,14 @@ export class Filter {
       if (!matches)
         return false;
     }
-
+    if (this.annotations.length) {
```
pavelfeldman, line 154: This will return all the tests w/o annotations as matching the search criteria.

packages/html-reporter/src/testCaseView.tsx
```diff
@@ -68,15 +69,19 @@ function renderAnnotationDescription(description: string) {
...
 }
 
 function TestCaseAnnotationView({ annotation: { type, description } }: { annotation: TestCaseAnnotation }) {
+  const [hover, setHover] = React.useState(false);
+  const onHover = () => setHover(true);
+  const onLeave = () => setHover(false);
   return (
-    <div className='test-case-annotation'>
-      <span style={{ fontWeight: 'bold' }}>{type}</span>
+    <div className='test-case-annotation' onMouseEnter={onHover} onMouseLeave={onLeave}>
```
pavelfeldman, line 81: It looks like you are re-rendering the component that you add listeners to on hover, which might results in unexpected results.

I'd suggest adding a new component next to CopyToClipboard called <CopyContainer><span>aaa</span></CopyContainer> that would allow copying its content. That way we can use it in more places. I would also recommend that the visibility is implemented using CSS.

pavelfeldman (commented):
Overall, looks good. There are some comments I'd like you to address and we are missing the tests for the new functionality.

packages/html-reporter/src/filter.ts
amankagithub: I tried and it did not . Only test case with annotations will be returned . In below example ,only two test cases which had some annotation is displayed . Am I missing something here?
Screenshot 2024-03-28 at 7 28 31 PM

packages/html-reporter/src/testCaseView.tsx
amankagithub: Not fully sure what you mean here . Made the required changes based on my what I could understand from your comment .

packages/html-reporter/src/filter.ts
pavelfeldman: Ah, you are right, I missed that `this` is a filter here. Still need a test for every new feature though.

amankagithub: I have added the test cases for copy feature . Please review .

packages/html-reporter/src/filter.ts
```diff
@@ -144,7 +151,14 @@ export class Filter {
...
+      const matches = this.annotations.every(annotation =>
+        searchValues.annotations.some(_annotation => (
```
yury-s, line 156: style nit: either drop `()` around `_annotation.includes(annotation)` or use `{}`

packages/html-reporter/src/filter.ts
```diff
@@ -144,7 +151,14 @@ export class Filter {
...
+          _annotation.includes(annotation)
+        )));
+      if (!matches)
+        return false;
```
yury-s, line 160: Please add a test to `tests/playwright-test/reporter-html.spec.ts` that checks that the annotation filter actually works.

amankagithub: Have added the test as you suggested . Please review .

dgozman:
@amankagithub Unfortunately, we found it hard to review and merge this PR. Therefore, it is going slowly, sorry for that!

To make this process better, let's follow our general guidelines:
- Start with filing a feature request following the issue template.
- Split this PR into multiple focused PRs, e.g. "copy to clipboard" vs "accessibility improvements" vs "filter by annotations".
- For any UI change, attach a screenshot that illustrates the change.

I'll close this PR, since it won't be merged due to the above.

Issue #30115, opened 2024-03-26
[Feature]: Emulate `navigator.cookieEnabled` false, to test When Cookies are disabled for the Website


nitzanashi:
### 🚀 Feature Request

To Test the scenario where Cookies are disabled for the Website, we would like to have the option to disable cookies for the `navigator` object of the `window`.

### Example

In Cypress it would work like the following: 
```js
cy.visit('/login', {
    onBeforeLoad(win) {
        cy.stub(win.navigator, 'cookieEnabled', false);
    },
});
```

### Motivation

We are migrating from Cypress to Playwright, and currently, we want to test the behaviour of our WebApp when Cookies are disabled.

nitzanashi:
@pavelfeldman Thanks a lot, that did the trick for us

pavelfeldman:
Closing as per above, please feel free to open a new issue if this does not cover your use case.

PR #30122, opened 2024-03-26
doc: document mocking read-only browser apis

pavelfeldman:
Documents https://github.com/microsoft/playwright/issues/30115

### DIFF
--- a/docs/src/mock-browser-js.md
+++ b/docs/src/mock-browser-js.md
@@ -60,6 +60,23 @@ test('show battery status', async ({ page }) => {
 
 ```
 
+## Mocking read-only APIs
+
+Some APIs are read-only so you won't be able to assign to a navigator property. For example,
+
+```js
+// Following line will have no effect.
+navigator.cookieEnabled = true;
+```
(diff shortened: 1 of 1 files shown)


dgozman (approved):

docs/src/mock-browser-js.md
```diff
@@ -60,6 +60,23 @@ test('show battery status', async ({ page }) => {
...
+```js
+// Following line will have no effect.
+navigator.cookieEnabled = true;
+```
+
+However, if the property is [configurable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#configurable), you can still override it using the plain JavaScript:
+
+```js
+await page.addInitScript(() => {
+  Object.defineProperty(navigator.__proto__, 'cookieEnabled', { value: false })
```
mxschmitt, line 76:
```suggestion
  Object.defineProperty(Object.getPrototypeOf(navigator), 'cookieEnabled', { value: false });
```

PR #30159, opened 2024-03-28
fix: preserve test declaration order in html and merged report

### LINKED ISSUE #29984: [Bug]: Test execution order is not same in HTML report as visible in terminal if we define nested describe block
### Version

1.41.2

### Steps to reproduce

I have declared three describe blocks and execution is working fine and visible in order in terminal but not in HTML report

image

image

### Expected behavior

The report should be show the same test execution order

### Actual behavior

If I have multiple nested describe block, insider describe block tests result are visible first then outer describe block tests (refer to above screenshot)

### Additional context

_No response_

### Environment

```shell
window 10
```

yury-s:
* Add `Suite.entries` that returns tests and suites in their declaration order
* Exposed `Suite.type` and `TestCase.type` for discriminating between different entry types.
* Blob report format is updated to store entries instead of separate lists for suites and tests.
* Bumped blob format version to 2, added modernizer.

Fixes https://github.com/microsoft/playwright/issues/29984


yury-s:
Not sure about the Suite subtypes, just keeping existing interface with a new field `type` for discriminating between suites and tests in the entries array maybe more clear. Also it's probably time to bump blob report version to 2 instead of supporting both new and old format for writing suites witing suites.

packages/playwright/src/isomorphic/teleReceiver.ts
```diff
@@ -57,8 +57,9 @@ export type JsonProject = {
 export type JsonSuite = {
   title: string;
   location?: JsonLocation;
-  suites: JsonSuite[];
-  tests: JsonTestCase[];
+  entries: (JsonSuite | JsonTestCase)[];
```
pavelfeldman, line 60: You would need to make entries optional if you want to support old reports in the merger.

packages/playwright/src/isomorphic/teleReceiver.ts
```diff
@@ -431,8 +449,71 @@ export class TeleSuite implements reporterTypes.Suite {
...
+    this.suites.push(suite);
+  }
+}
+
+export class TeleDescribeSuite extends TeleSuite implements reporterTypes.DescribeSuite {
+  type: 'describe' = 'describe';
+  override location!: reporterTypes.Location;
+  override parent!: TeleDescribeSuite | TeleFileSuite;
+  constructor(title: string) {
+    super(title, 'project');
```
pavelfeldman, line 472: Let's not introduce these wrappers and only use TeleSuite. Prone to errors like this one.

packages/playwright/src/reporters/merge.ts
```diff
@@ -387,14 +387,31 @@ class IdsPatcher {
...
   private _updateTestIds(suite: JsonSuite) {
-    suite.tests.forEach(test => {
-      test.testId = this._mapTestId(test.testId);
-      if (this._botName) {
-        test.tags = test.tags || [];
-        test.tags.unshift('@' + this._botName);
-      }
-    });
-    suite.suites.forEach(suite => this._updateTestIds(suite));
+    if (suite.entries) { // Starting 1.44
```
pavelfeldman, line 390: Why is this code changing?

packages/playwright/src/reporters/merge.ts
```diff
@@ -460,10 +477,19 @@ class PathSeparatorPatcher {
     this._updateLocation(suite.location);
     if (isFileSuite)
       suite.title = this._updatePath(suite.title);
-    for (const child of suite.suites)
-      this._updateSuite(child);
-    for (const test of suite.tests)
-      this._updateLocation(test.location);
+    if (suite.entries) { // starting 1.44
```
pavelfeldman, line 480: Why is this code changing?

packages/playwright/types/testReporter.d.ts
```diff
@@ -224,6 +237,29 @@ export interface TestCase {
   title: string;
 }
 
+export interface FileSuite extends Suite {
```
pavelfeldman, line 240: Let's not export these

docs/src/test-reporter-api/class-suite.md
```diff
@@ -26,6 +26,12 @@ Reporter is given a root suite in the [`method: Reporter.onBegin`] method.
 
 Returns the list of all test cases in this suite and its descendants, as opposite to [`property: Suite.tests`].
 
+## method: Suite.entries
+* since: v1.44
+- type: <[Array]<[TestCase]|[Suite]>>
+
+Test cases and suites defined directly in this suite. The elements returned in their declaration order. You can discriminate between different entry types using [`property: TestCase.type`] and [`property: Suite.type`].
```
pavelfeldman, line 33:
```suggestion
Test cases and suites defined directly in this suite. The elements are returned in their declaration order. You can discriminate between different entry types using [`property: TestCase.type`] and [`property: Suite.type`].
```

docs/src/test-reporter-api/class-suite.md
```diff
@@ -72,3 +78,10 @@ Suite title.
 - returns: <[Array]<[string]>>
 
 Returns a list of titles from the root down to this suite.
+
+## property: Suite.type
+* since: v1.44
+- returns: <[SuiteType]<'root' | 'project' | 'file' | 'describe'>>
+
+Returns type of the suite. The Suites form the following hierarchy:
```
pavelfeldman, line 86:
```suggestion
Returns the type of the suite. The Suites form the following hierarchy:
```

packages/playwright/src/isomorphic/teleReceiver.ts
```diff
@@ -336,31 +334,37 @@ export class TeleReporterReceiver {
...
+    targetSuite.location = this._absoluteLocation(jsonSuite.location);
+    if (jsonSuite.entries) {
+      jsonSuite.entries.forEach(e => {
+        if ('testId' in e)
+          this._mergeTestInto(e, targetSuite!);
+        else
+          this._mergeSuiteInto(e, targetSuite!);
+      });
+    } else {
+      // before 1.44
```
pavelfeldman, line 352: Do we still need this code?

packages/playwright/src/isomorphic/teleReceiver.ts
```diff
@@ -434,6 +453,18 @@ export class TeleSuite implements reporterTypes.Suite {
   project(): TeleFullProject | undefined {
     return this._project ?? this.parent?.project();
   }
+
+  _addTest(test: TeleTestCase) {
+    test.parent = this;
+    this._entries.push(test);
+    this.tests.push(test);
```
pavelfeldman, line 460: Are you pushing into the air here?

packages/playwright/src/isomorphic/teleReceiver.ts
```diff
@@ -434,6 +453,18 @@ export class TeleSuite implements reporterTypes.Suite {
...
+  }
+
+  _addSuite(suite: TeleSuite) {
+    suite.parent = this;
+    this._entries.push(suite);
+    this.suites.push(suite);
```
pavelfeldman, line 466: ditto

packages/playwright/src/reporters/merge.ts
```diff
@@ -508,3 +519,39 @@ class JsonEventPatchers {
...
+    return events;
+  }
+
+  _modernize_1_to_2(events: JsonEvent[]): JsonEvent[] {
+    return events.map(event => {
+      if (event.method === 'onProject') {
+        const modernizeSuite = (suite: blobV1.JsonSuite) => {
+          for (const child of suite.suites)
+            modernizeSuite(child);
+          (suite as any).entries = [...suite.suites, ...suite.tests];
```
pavelfeldman, line 544: Something like this can help getting rid of any. Otherwise there is not much point in strictness.

```js
const { suites, tests, ...remainder } = suite;
return { entries: [...suites, ...tests], ...remainder };
```

packages/playwright/src/isomorphic/teleReceiver.ts
yury-s: Nope. Deleted.

packages/playwright/src/isomorphic/teleReceiver.ts
yury-s: Apparently. Just in case 😄

packages/playwright/src/reporters/merge.ts
yury-s: Done.

pavelfeldman (approved):

PR #30192, opened 2024-03-31
feat(expect) - add ignoreCase option for toHaveURL

### LINKED ISSUE #30057: [Feature]: Add IgnoreCase parameter in PageAssertions.ToHaveURLAsync() assertion method / PageAssertionsToHaveURLOptions
### 🚀 Feature Request

Currently method PageAssertions.ToHaveURLAsync() is case sensitive i.e. assertion fails when expected url is 'https://www.facebook.com/microsoft/' but received 'https://www.facebook.com/Microsoft/'

I think in most cases this difference does not matter. My proposal is to add property **IgnoreCase** [bool]? (optional) to options PageAssertionsToHaveURLOptions? (optional)

https://playwright.dev/dotnet/docs/api/class-pageassertions#page-assertions-to-have-url

### Example

await Page.GotoAsync("https://www.facebook.com/microsoft"); // it will auto-change to Microsoft
await Expect(Page).ToHaveURLAsync("https://www.facebook.com/microsoft/"); // this will fail

### Motivation

- facilitate waiting for URL when case does not matter
- IgnoreCase parameter in ToHaveURLAsync would be consistent with ToHaveTextAsync() and ToContainTextAsync() methods that also compare text and support case insensitive comparison.

KozynchenkoVS:
feat(expect): add ignoreCase option for toHaveURL

Fixes #30057

### DIFF
--- a/docs/src/api/class-pageassertions.md
+++ b/docs/src/api/class-pageassertions.md
@@ -116,6 +116,13 @@ The opposite of [`method: PageAssertions.toHaveURL`].
 
 Expected URL string or RegExp.
 
+### option: PageAssertions.NotToHaveURL.ignoreCase
+* since: v1.44
+* langs: js
+- `ignoreCase` <[boolean]>
+
+Whether to perform case-insensitive match. [`option: ignoreCase`] option takes precedence over the corresponding regular expression flag if specified.
+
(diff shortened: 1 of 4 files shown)


KozynchenkoVS:

yury-s (approved):

PR #30223, opened 2024-04-03
exporting Devices type

AlfieJones:
This pr exports the auto-generated `Devices` type. 

Without exporting, you can only access the more generic version with `type Devices = typeof devices` which results in the type `[key: string]: DeviceDescriptor` which doesn't have the 'autocompleted' keys

### DIFF
--- a/utils/generate_types/index.js
+++ b/utils/generate_types/index.js
@@ -509,7 +509,7 @@ class TypesGenerator {
     let types = await generator.generateTypes(path.join(__dirname, 'overrides.d.ts'));
     const namedDevices = Object.keys(devices).map(name => `  ${JSON.stringify(name)}: DeviceDescriptor;`).join('\n');
     types += [
-      `type Devices = {`,
+      `export type Devices = {`,
       namedDevices,
       `  [key: string]: DeviceDescriptor;`,
       `}`,


dgozman:
@AlfieJones The following snippet correctly identifies that `instance.viewport` is missing `width`/`height`, so I'd think that `typeof devices` works nicely.

```ts
import { devices } from '@playwright/test';
type Devices = typeof devices;
type iPhone = Devices['iPhone X'];
const instance: iPhone = { viewport: {} };
```

AlfieJones:
The issue for me is replicating the autocomplete experience devices offers. When using the object directly, typescript is able to show be the available devices. However when using the derived type, I am not able to replicate the functionality.

```ts
import { devices } from 'playwright-core';

// lsp shows the available options
const iPhone = devices['iPhone X'];

type Devices = typeof devices;
const example = {} as Devices;

// lsp doesn't show available options
const iPhone = example ['iPhone X'];

```

AlfieJones:
In my project, https://github.com/pixeleye-io/pixeleye, I'm re-exporting this object of devices after adding additional properties to each of the elements. If possible, I'd rather avoid recreating a similar custom-type generator as you've implemented.

yury-s:
How exactly do you export modified devices? The following code works fine:

```js
// helper.ts
import { devices } from 'playwright-core';

type Devices = typeof devices;
export const example: Devices = {} as Devices;
```

image

dgozman:
Closing since the issue is not reproducible. Next time, could you please start with filing an issue instead of sending a PR right away?

PR #30233, opened 2024-04-03
feat(junit): includeProjectInTestName option

### LINKED ISSUE #30246: [Feature]:  Customizable Junit Output for title
### 🚀 Feature Request

Allow the customizing of the Junit Output, specifically in our case the title.

The current code removes the project and other parts : 

```
      attributes: {
        // Skip root, project, file
        name: test.titlePath().slice(3).join(' › '),
        // filename
        classname: suiteName,
        time: (test.results.reduce((acc, value) => acc + value.duration, 0)) / 1000

      },
```

### Example

In Azure Devops it's hard to differentiate between one testcase run against multiple projects:
the below is from 4 different projects.
image

### Motivation

My use case is in Azure devops, but there are many different CI/CD Systems, to it would be IDEAL if we could provide a template like jest-junit but anything to start off with would be good!

or a way for us to easily override parts of the junit-reporters, without having to hack into the imports of playwright

dgozman:
Fixes #30246.

### DIFF
--- a/packages/playwright/src/reporters/junit.ts
+++ b/packages/playwright/src/reporters/junit.ts
@@ -33,11 +33,13 @@ class JUnitReporter extends EmptyReporter {
   private outputFile: string | undefined;
   private resolvedOutputFile: string | undefined;
   private stripANSIControlSequences = false;
+  private includeProjectInTestName = false;
 
-  constructor(options: { outputFile?: string, stripANSIControlSequences?: boolean } = {}) {
+  constructor(options: { outputFile?: string, stripANSIControlSequences?: boolean, includeProjectInTestName?: boolean } = {}) {
     super();
     this.outputFile = options.outputFile || reportOutputNameFromEnv();
     this.stripANSIControlSequences = options.stripANSIControlSequences || false;
+    this.includeProjectInTestName = options.includeProjectInTestName || false;
   }
 
   override printsToStdio() {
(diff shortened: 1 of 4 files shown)


packages/playwright/src/reporters/junit.ts
```diff
@@ -33,11 +33,13 @@ class JUnitReporter extends EmptyReporter {
...
   private resolvedOutputFile: string | undefined;
   private stripANSIControlSequences = false;
+  private includeProjectInTestName = false;
 
-  constructor(options: { outputFile?: string, stripANSIControlSequences?: boolean } = {}) {
+  constructor(options: { outputFile?: string, stripANSIControlSequences?: boolean, includeProjectInTestName?: boolean } = {}) {
     super();
     this.outputFile = options.outputFile || reportOutputNameFromEnv();
     this.stripANSIControlSequences = options.stripANSIControlSequences || false;
+    this.includeProjectInTestName = options.includeProjectInTestName || false;
```
pavelfeldman, line 42: We have a new guidelines - all the reporter properties should be configurable via env. That way they can be merged w/o additional config.

pavelfeldman (approved):

packages/playwright/src/reporters/junit.ts
dgozman: Sounds good, I will follow up with env vars for all the options.

PR #30234, opened 2024-04-03
chore: move dev server to config-based framework extensibility

pavelfeldman:


dgozman (approved):

packages/playwright-ct-core/src/program.ts
```diff
@@ -14,19 +14,4 @@
  * limitations under the License.
  */
 
-import type { Command } from 'playwright-core/lib/utilsBundle';
-
-import { program } from 'playwright/lib/program';
-import { runDevServer } from './devServer';
 export { program } from 'playwright/lib/program';
```
dgozman, line 17: Is this file still used by someone?

dgozman (approved):

packages/playwright-ct-core/src/program.ts
pavelfeldman: Yes, it is used by all the CT modules.

PR #30257, opened 2024-04-04
test: unflake 'should not include trace resources from the previous chunks' test

### LINKED ISSUE #30247: [internal] look into flaky 'should not include trace resources from the previous chunks' test
Repro: `npm run ctest -- --workers=1  library/browsercontext-reuse.spec.ts:268:1 library/tracing.spec.ts:243:5`

mxschmitt:
The problem was the following:

1. [tests/library/browsercontext-reuse.spec.ts:270](https://github.com/microsoft/playwright/blob/32aecac55b67de801bbabf1b5ff4653b4de38139/tests/library/browsercontext-reuse.spec.ts#L270-L285) starts tracing but never stops it.
2. This means that [Connection._tracingCount](https://github.com/microsoft/playwright/blob/7ad255301f500c323030c2a67415a013e2cf0c35/packages/playwright-core/src/client/connection.ts#L76) is still 1 when starting to run other tests
3. In the affected test we start tracing again, and another chunk, so `tracingCount` is 2.
4. When stopping the last chunk, we add its record as an action to the trace, [if the count is > 0](https://github.com/microsoft/playwright/blob/7ad255301f500c323030c2a67415a013e2cf0c35/packages/playwright-core/src/client/connection.ts#L137) - in this case its 1, because there was a stale trace running. So `Tracing.stopChunk` gets added, which creates the resource.txt file and the test will fail.

This PR fixes it based on that in two places:

1. Actually close the re-used context when the reuse tests are done. This does not keep an unused context lying around.
2. Discard Tracing when a context gets closed. This will reset the connection trace counters and dispose all the used resources.

Fixes https://github.com/microsoft/playwright/issues/30247

### DIFF
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -432,7 +432,7 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
       this._browser._contexts.delete(this);
     this._browserType?._contexts?.delete(this);
     this._disposeHarRouters();
-    this.tracing._resetStackCounterIfNeeded();
+    this.tracing._resetStackCounter();
     this.emit(Events.BrowserContext.Close, this);
   }
 
--- a/packages/playwright-core/src/client/browserContext.ts
+++ b/packages/playwright-core/src/client/browserContext.ts
@@ -432,6 +432,7 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
       this._browser._contexts.delete(this);
     this._browserType?._contexts?.delete(this);
     this._disposeHarRouters();
+    this.tracing._resetStackCounterIfNeeded();
     this.emit(Events.BrowserContext.Close, this);
   }
 
--- a/tests/library/browsercontext-reuse.spec.ts
+++ b/tests/library/browsercontext-reuse.spec.ts
@@ -17,19 +17,17 @@
 import { browserTest, expect } from '../config/browserTest';
 import type { BrowserContext, BrowserContextOptions } from '@playwright/test';
 
-const test = browserTest.extend<{}, { reusedContext: (options?: BrowserContextOptions) => Promise<BrowserContext> }>({
-  reusedContext: [async ({ browserType, browser }, use) => {
-    let context;
+const test = browserTest.extend<{ reusedContext: (options?: BrowserContextOptions) => Promise<BrowserContext> }>({
+  reusedContext: async ({ browserType, browser }, use) => {
     await use(async (options: BrowserContextOptions = {}) => {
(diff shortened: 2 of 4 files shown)


packages/playwright-core/src/client/browserContext.ts
```diff
@@ -446,6 +446,7 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
     this._closeWasCalled = true;
     await this._wrapApiCall(async () => {
       await this._browserType?._willCloseContext(this);
+      await this.tracing.stop();
```
mxschmitt, line 449:
```suggestion
      // Discard all non-stopped tests
      await this.tracing.stop();
```

packages/playwright-core/src/client/browserContext.ts
```diff
@@ -432,6 +432,7 @@ export class BrowserContext extends ChannelOwner<channels.BrowserContextChannel>
       this._browser._contexts.delete(this);
     this._browserType?._contexts?.delete(this);
     this._disposeHarRouters();
+    this.tracing._resetStackCounterIfNeeded();
```
dgozman, line 435: Do the same for APIRequestContext?

tests/library/browsercontext-reuse.spec.ts
```diff
@@ -17,17 +17,19 @@
...
       const defaultContextOptions = (browserType as any)._defaultContextOptions;
-      const context = await (browser as any)._newContextForReuse({
+      context = await (browser as any)._newContextForReuse({
         ...defaultContextOptions,
         ...options,
       });
       return context;
     });
-  },
+    await context.close();
```
dgozman, line 31: We don't need this one anymore, do we?

packages/playwright-core/src/client/browserContext.ts
mxschmitt: `page.request._tracing === context.tracing` - its the same Tracing instance under the hood.

tests/library/browsercontext-reuse.spec.ts
mxschmitt: This would otherwise keep a context lying around after the last test of this file has been completed.

packages/playwright-core/src/client/browserContext.ts
dgozman: There is also a global `APIRequestContext` not attached to a `BrowserContext`.

tests/library/browsercontext-reuse.spec.ts
dgozman: Since this file is testing a reused browser context, perhaps we should keep it open to be reused? Anyway, this seems like an unrelated fix to me.

dgozman (approved):

PR #30258, opened 2024-04-05
test: check that ESM loader works without config with type:module

dgozman:


tests/playwright-test/esm.spec.ts
```diff
@@ -589,6 +589,28 @@ test('should load mts without config', async ({ runInlineTest, nodeVersion }) =>
   expect(result.passed).toBe(1);
 });
 
+test('should load type module without config', async ({ runInlineTest, nodeVersion }) => {
+  test.skip(nodeVersion.major < 18, 'ESM loader is enabled conditionally with older API');
```
mxschmitt, line 593: Note: As per [INFRASTRUCTURE.md](https://github.com/microsoft/playwright-internal/blob/main/INFRASTRUCTURE.md) we can in theory drop all the Node.js 16 bots.

mxschmitt (approved):

PR #30264, opened 2024-04-05
feat(ct): bump vite to v5.2.8

sand4rt:
Notable release notes of v5.1: https://vitejs.dev/blog/announcing-vite5-1.html

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -853,9 +853,9 @@
       }
     },
     "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.19.11",
-      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.19.11.tgz",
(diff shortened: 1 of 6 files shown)


mxschmitt (approved):
Looks like this is a test only change / will only affect our tests?

Since for users, who depend on any of the CT packages, we already had `"vite": "^5.0.11"` as a requirement, which means they could have consumed 5.2 already.

sand4rt:
Hi @mxschmitt, i believe this affects a small subset of users who rely on the deprecated Vite APIs because the components are transpiled using Playwright's Vite version

By the way, i filled out the form for the Playwright SWAG😎

mxschmitt:
> Hi @mxschmitt, i believe this affects a small subset of users who rely on the deprecated Vite APIs because the components are transpiled using Playwright's Vite version

Yes, but we depend on `"vite": "^5.0.11"` and dependencies which are shipped to the user are not locked from my understanding.

> By the way, i filled out the form for the Playwright SWAG some time ago😎 are they still in stock?

Please ping me on Discord, there we can find out whats going on!

Issue #23377, opened 2023-05-30
[BUG] some getByRole roles don't work


eatonjl:
### System info
- Playwright Version: 1.33
- Operating System: Mac OS
- Browser: Chromium
- Other info:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Config file**

```ts
import { defineConfig, devices } from '@playwright/test';

/**
 * Read environment variables from file.
 * https://github.com/motdotla/dotenv
 */
// require('dotenv').config();

/**
 * See https://playwright.dev/docs/test-configuration.
 */
export default defineConfig({
  testDir: './tests',
  /* Run tests in files in parallel */
  fullyParallel: true,
  /* Fail the build on CI if you accidentally left test.only in the source code. */
  forbidOnly: !!process.env.CI,
  /* Retry on CI only */
  retries: process.env.CI ? 2 : 0,
  /* Opt out of parallel tests on CI. */
  workers: process.env.CI ? 1 : undefined,
  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
  reporter: 'html',
  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
  use: {
    /* Base URL to use in actions like `await page.goto('/')`. */
    // baseURL: 'http://127.0.0.1:3000',

    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
    trace: 'on',
  },

  /* Configure projects for major browsers */
  projects: [
    {
      name: 'chromium',
      use: { ...devices['Desktop Chrome'] },
    },

    {
      name: 'firefox',
      use: { ...devices['Desktop Firefox'] },
    },

    {
      name: 'webkit',
      use: { ...devices['Desktop Safari'] },
    },

    /* Test against mobile viewports. */
    // {
    //   name: 'Mobile Chrome',
    //   use: { ...devices['Pixel 5'] },
    // },
    // {
    //   name: 'Mobile Safari',
    //   use: { ...devices['iPhone 12'] },
    // },

    /* Test against branded browsers. */
    // {
    //   name: 'Microsoft Edge',
    //   use: { ...devices['Desktop Edge'], channel: 'msedge' },
    // },
    // {
    //   name: 'Google Chrome',
    //   use: { ..devices['Desktop Chrome'], channel: 'chrome' },
    // },
  ],

  /* Run your local dev server before starting the tests */
  // webServer: {
  //   command: 'npm run start',
[...]

PR #30267, opened 2024-04-05
test: add `getByRole` tests for all roles

billfienberg:
## TLDR

This PR adds `getByRole` tests for every aria role, and annotates the the failing roles with the `fixme` [annotation](https://playwright.dev/docs/test-annotations). 

## Related Issues

- https://github.com/microsoft/playwright/issues/23377

### DIFF
--- a/tests/page/locator-role.spec.ts
+++ b/tests/page/locator-role.spec.ts
@@ -1,0 +1,150 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
+ * Modifications copyright (c) Microsoft Corporation.
+ *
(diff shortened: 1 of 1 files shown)


billfienberg:

dgozman:
@billfienberg Thank you for the PR! Unfortunately, most likely we would not merge it. We'd prefer PRs that identify the expected behavior based on the real-life examples rather than unit tests that just document the existing behavior. Ideally, there would be an improvement that fixes an issue and includes relevant test cases.

If you are willing to contribute in this space, I'd recommend to first collaborate on the issue to identify the usecases and possible improvements. Let me know what you think.

Issue #29928, opened 2024-03-13
[Bug]: 


pbrzosko:
### Version

1.42.1

### Steps to reproduce

1. Clony my repo https://github.com/pbrzosko/playwright-define-component
2. npm install
3. npm run test-ct
4. 2 out of 3 tests are failing

### Expected behavior

I would expect all tests to pass.

### Actual behavior

I am surprised that neither component imported from a separate file, but defined with defineComponent nor the one inline directly in the mount methods are not mounted and thus not visible in the test.

### Environment

```shell
System:
    OS: macOS 13.6.4
    CPU: (12) arm64 Apple M2 Pro
    Memory: 60.97 MB / 16.00 GB
  Binaries:
    Node: 18.16.1 - ~/.nvm/versions/node/v18.16.1/bin/node
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.1/bin/npm
    pnpm: 8.9.2 - ~/.nvm/versions/node/v18.16.1/bin/pnpm
  Languages:
    Bash: 5.2.15 - /opt/homebrew/bin/bash
  npmPackages:
    @playwright/experimental-ct-vue: ^1.42.1 => 1.42.1
```

mxschmitt:
defineComponent is currently not supported - we only support .vue files.

pbrzosko:
Understand, but this is a way of creating components in Vue, so I assume it should be supported.

mxschmitt:
Yup, marking it as a feature request!

per-toyra-stratsys:
Old issue, but we've are investigating a workaround for this by adding a `.vue` file that imports the component defined using `defineComponent`.

E.g. `test-wrapper.vue`:
```vue
<script src="@corp/my-component/define-component-file.ts"></script>
```

This seems to work just fine, but we haven't tested this very carefully yet. Could there be any unforeseen issues we have yet to discover by doing it this way?

ryaposov:
I am currently exploring ways to make component testing work for our Vue setup with Playwright and this becomes an issue. Do I understand it correctly, since it's not possible to import anything else but SFC .vue components, we're forced to wrap every tested component story in a separate file? As opposed to React support which naturally allows to import a number of components via one import as long as they are imported from outside of the file.

I'm currently migrating a big testing setup of our design system from Cypress and this then becomes a huge blocker since for some components we have 20 tests (stories) and more — I don't really see us creating a SFC for every single one of them.

[...]

PR #30269, opened 2024-04-05
feat(ct): class and object components

### LINKED ISSUE #29544: [Bug]: Angular/Web component tests fail with `.ts` or `.js` as file extension
### Version

1.41.1

### Steps to reproduce

1. clone the repo https://github.com/sand4rt/playwright-ct-web/pull/36
2. run `pnpm i`
3. run `pnpm test`

### Expected behavior

the test should pass

### Actual behavior

the tests fail

logs: https://github.com/sand4rt/playwright-ct-web/actions/runs/7943833528/job/21688654668?pr=36#step:8:68

### Additional context

I'm unable to upgrade the Angular and web/lit component testing libraries from version v1.41.0 to v1.41.1 because the transform process was improved recently. The issue lies in the components not being transformed into refs. This occurs because Angular/Lit components are used as `.ts` files, while the transformation process in playwright-ct-core only handles non-JavaScript imports, as indicated here: https://github.com/microsoft/playwright/blob/main/packages/playwright-ct-core/src/tsxTransform.ts#L78

Is it an option to verify if the import is used within the `mount()` instead of verifying the file extension?

Related to https://github.com/microsoft/playwright/pull/27783

sand4rt:
closes https://github.com/microsoft/playwright/issues/29544, https://github.com/microsoft/playwright/issues/29928, partially https://github.com/microsoft/playwright/issues/31217 and introduces an improved approach for creating multiple wrapper/story components within a single file _(considering it's only possible to define one component in a `.vue` file)_. @yjaaidi made the change in [`tsxTransform.ts`](https://github.com/sand4rt/playwright/pull/5/files/3449eb8e120b8685bf394cf024ef83161ff21013#diff-8dd3534dc5013c3779edeaded71324b0dd1c1807668f3c6d9e9a1aab1c20ae91)

### DIFF
--- a/packages/playwright-ct-core/src/tsxTransform.ts
+++ b/packages/playwright-ct-core/src/tsxTransform.ts
@@ -21,6 +21,7 @@ import { setTransformData } from 'playwright/lib/transform/transform';
 const t: typeof T = types;
 
 let jsxComponentNames: Set<string>;
+let classComponentNames: Set<string>;
 let importInfos: Map<string, ImportInfo>;
 
 export default declare((api: BabelAPI) => {
@@ -32,6 +33,7 @@ export default declare((api: BabelAPI) => {
       Program: {
         enter(path) {
           jsxComponentNames = collectJsxComponentUsages(path.node);
+          classComponentNames = collectClassMountUsages(path.node);
           importInfos = new Map();
         },
         exit(path) {
@@ -93,7 +95,7 @@ export default declare((api: BabelAPI) => {
           if (t.isImportNamespaceSpecifier(specifier))
             continue;
           const { localName, info } = importInfo(importNode, specifier, this.filename!);
-          if (jsxComponentNames.has(localName)) {
+          if (jsxComponentNames.has(localName) || classComponentNames.has(localName)) {
             importInfos.set(localName, info);
             ++importCount;
           }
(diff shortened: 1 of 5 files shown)


sand4rt:
@pavelfeldman any chance this one could passthrough?

pavelfeldman:
👋  @sand4rt 👋  taking care of our review debt. is this something you'd like me to look at?

sand4rt:
👋 @pavelfeldman, yes please. Let me know if more info is needed.

I think @dgozman also has an opinion about this BTW. He mentioned something about this during the Angular PR review

tests/components/ct-vue-vite/tests/render/render.spec.ts
```diff
@@ -2,6 +2,7 @@ import { test, expect } from '@playwright/experimental-ct-vue';
 import Button from '@/components/Button.vue';
 import EmptyTemplate from '@/components/EmptyTemplate.vue';
 import Component from '@/components/Component.vue';
+import { Story } from '@/components/Story';
```
pavelfeldman, line 5: How typical is this?

sand4rt: It depends on the app, but I don't it's that typical; its useful when a wrapper/story needs to be created to test components, where Playwright cannot serialise the data sent to the browser _(such as unserialisable props)_. It's possible to create multiple `.vue` files but that's quite cumbersome.

There are additional use cases where importing components from a `.ts` file would be beneficial tho, such as supporting Angular and web components _(without relying on a specific file extension)_.

packages/playwright-ct-core/src/tsxTransform.ts
```diff
@@ -141,6 +143,23 @@ function collectJsxComponentUsages(node: T.Node): Set<string> {
   return names;
 }
 
+function collectClassMountUsages(node: T.Node): Set<string> {
+  const names = new Set<string>();
+  traverse(node, {
+    enter: p => {
+      // Treat calls to mount and all identifiers in arguments as component usages e.g. mount(Component)
+      if (t.isCallExpression(p.node) && t.isIdentifier(p.node.callee) && p.node.callee.name === 'mount') {
```
pavelfeldman, line 151: We used to have this as you probably know. But the `isIdentifier(p.node.callee)` felt overly restrictive and limited us a couple of times.

sand4rt: I don't see how? What approach would you recommend to address this? In my opinion, relying solely on file extension is not a viable option.

tests/components/ct-vue-vite/tests/render/render.spec.ts
sand4rt: This use case is very typical for Vue and is not possible without this PR: https://github.com/microsoft/playwright/issues/31217

pavelfeldman:
We are slow on it, let me close it as we are figuring out what we do with CT. The code is not the problem, so there is no good reason for this PR to be here held back by the reviews.

PR #30271, opened 2024-04-05
fix(esmLoader): unref MessagePort in the main thread

dgozman:
This prevents the process from not exiting forever due to an open port.

### DIFF
--- a/packages/playwright/src/transform/portTransport.ts
+++ b/packages/playwright/src/transform/portTransport.ts
@@ -33,17 +33,33 @@ export class PortTransport {
       if (ackId) {
         const callback = this._callbacks.get(ackId);
         this._callbacks.delete(ackId);
+        this._resetRef();
         callback?.(result);
         return;
       }
     });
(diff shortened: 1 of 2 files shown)


packages/playwright/src/transform/portTransport.ts
```diff
@@ -33,17 +33,33 @@ export class PortTransport {
...
   async send(method: string, params: any) {
     return await new Promise<any>(f => {
       const id = ++this._lastId;
       this._callbacks.set(id, f);
+      this._refOrUnref();
       this._port.postMessage({ id, method, params });
     });
   }
+
+  private _refOrUnref() {
```
pavelfeldman, line 55: Could we be a little less Hamlet and just call unref after adding the listener?

pavelfeldman: At least call it resetRef?

pavelfeldman (approved):

Issue #29768, opened 2024-03-01
[Internal]: Deprecate config.rootDir, pass empty string into the merger.


pavelfeldman:
Deprecate config.rootDir as it has no semantics post-merge.

yury-s:
Discussion notes:
- document `FullConfig`
  - split FullConfig to TestFullConfig, FullConfig
- stop serializing top level config fields, move them to the project level instead
- switch local reporters code to use only project level fields
- introduce `configName` property (set globally, available on `FullProject` only)
- introduce explicit global `rootDir` which will be replaced by `configName` in the final reports

PR #30274, opened 2024-04-06
chore: read config file location from url params in ui mode

yury-s:
Reference https://github.com/microsoft/playwright/issues/29768


packages/trace-viewer/src/ui/uiModeFiltersView.tsx
```diff
@@ -80,9 +79,8 @@ export const FiltersView: React.FC<{
                 const copy = new Map(projectFilters);
                 copy.set(projectName, !copy.get(projectName));
                 setProjectFilters(copy);
-                const configFile = testModel?.config?.configFile;
```
pavelfeldman, line 83: It is safe to use testModel.config.configFile - it will always be there in the UI mode scenario.

pavelfeldman (commented):
I would land the refactoring, but revert the query param.

Issue #28321, opened 2023-11-24
[BUG] Ambiguous test outcome and status for serial mode


agoldis:
### System info
- Playwright Version: `v1.39.0`, `v1.40.0`
- Operating System: All
- Browser: n/a
- Other info:

### Source code

- [x] I provided exact source code that allows reproducing the issue locally.

**Link to the GitHub repository with the repro**

https://github.com/currents-dev/pw-issue-2023-11-23-001

**Steps**

- Checkout  the demo repo, follow README instructions
- Run the test for `1.39`
- Run the test for `1.40`

**Expected**

The following test suite:

```ts
import { expect, test } from "@playwright/test";

test.describe.configure({ mode: "serial", retries: 1 });

test("A", async ({ page }, { retry }) => {
  if (retry === 0) {
    throw new Error("oh!");
  }
  if (retry === 1) {
    expect(true).toBe(true);
  }

});

test("B", async ({ page }, { retry }) => {
  throw new Error("oh!");
});
```

Is supposed to produce the following outcome:
- Test A results status: failed, passed
- Test B results status: skipped, failed

- `onTestBegin` should be invoked for every attempt of every test in a serial group as in `1.39`
- all tests should exhaust all attempts in a serial group

**Actual**

`1.40` runs serial tests group differently and produces different results:
- Test A results status: failed, passed
- Test B results status: failed

- `1.40` doesn't invoke `onTestBegin` for test `B` when `A` fails
- `1.40` doesn't retry test `B` and "swallows" the last retry

mxschmitt:
Bisect NPM:
1.40.0-alpha-oct-25-2023 bad
1.40.0-alpha-oct-24-2023 good

RANGE: https://github.com/microsoft/playwright/compare/fe7847b126ce2059d02fbbbc1f632b04fdae3ef7...0bb9f7cdf7367185c89a24305ca21d0768e5d54f

Looks caused by https://github.com/microsoft/playwright/pull/27762

dgozman:
@agoldis This is an intended change. After A fails, B is not run at all, so we do not produce a "skipped" run. Would you like to still see a skipped run for B? What's the reason? Please provide as much details about your usecase as you can.

agoldis:
Would you like to still see a skipped run for B?

- Yes, if that's possible

What's the reason?

dgozman:
[...]

Issue #27455, opened 2023-10-05
[BUG] Serial + Retry : Failed test is marked as flaky even if it never passes


javamaniac:
### System info
- Playwright Version: [v1.38.0]
- Operating System: [All, Windows 11, Ubuntu 20, macOS 13.2, etc.]
- Browser: [All, Chromium, Firefox, WebKit]

### Source code
```js
import {expect, test} from '@playwright/test';

test.describe.serial("Tests", () => {
  // Flaky test
  test("First test", () => {
    expect(Math.random()).toBeGreaterThan(0.3); // some flaky step
  });

  test("Second test", () => {
    expect(true).toEqual(false);
  });
});
```

**Description**
The second test is marked as flaky if tests are in describe.serial block with "retries: 1" settled in tests config file.

**Steps**
Execute the test until:
- The second test is executed at least once and the first test failed on the last retry

**Expected**
```
  1 failed
    [chromium] › test-serial-retry.spec.ts:8:7 › Tests › Second test (Failed) ──────────
  1 passed (648ms)
```

**Actual**
If the second test is executed at least once and the first test failed on the last retry:
```
Running 2 tests using 1 worker

  ✓  1 [chromium] › test-serial-retry.spec.ts:4:7 › Tests › First test (21ms)
  ✘  2 [chromium] › test-serial-retry.spec.ts:8:7 › Tests › Second test (11ms)
  ✘  3 [chromium] › test-serial-retry.spec.ts:4:7 › Tests › First test (retry #1) (23ms)
  -  4 [chromium] › test-serial-retry.spec.ts:8:7 › Tests › Second test (retry #1)
...
  2 flaky
    [chromium] › test-serial-retry.spec.ts:4:7 › Tests › First test ────────────
    [chromium] › test-serial-retry.spec.ts:8:7 › Tests › Second test ──────────
```
If the second test is executed on the last retry:
```
Running 2 tests using 1 worker

  ✓  1 [chromium] › test-serial-retry.spec.ts:4:7 › Tests › First test (21ms)
  ✘  2 [chromium] › test-serial-retry.spec.ts:8:7 › Tests › Second test (11ms)
  ✓  3 [chromium] › test-serial-retry.spec.ts:4:7 › Tests › First test (retry #1) (21ms)
  ✘  4 [chromium] › test-serial-retry.spec.ts:8:7 › Tests › Second test (retry #1) (11ms)
...
  1 failed
[...]

Issue #17652, opened 2022-09-28
[Question] Skip on retry


JackSpa89:
Hello,

I have a question. Currently in our test suite we have some tests running, that are flaky in the `beforeEach` hook ( due to some frontend issue ). The test it self is supposed to be skipped under conditions ( that we only know after the `beforeEach` hook is executed. ). 
So the issue is, that we have the retry mechanism enabled and if the `beforeEach` hook fails for example in the first try, but succeeds in the 2nd try and then the test is supposed to skipped. 
But playwright still seems to mark the test as failed. 
I could put the `beforeEach` hook in the tests itself and this would "fix" it, but we have like 10 tests with the same preconditions and this does not seem like the best way.

So is there any way to mark the test still as skipped if first `beforeEach` hook fails? Or is this a bug from playwright side?
Any help would be appreciated :) 

Thanks and BR

aslushnikov:
To illustrate this:

```ts
// a.spec.ts

import { test } from '@playwright/test';
import fs from 'fs';
import path from 'path';

test.beforeEach(() => {
  const filename = path.join(__dirname, 'flag');
  if (fs.existsSync(filename))
    return;
  fs.writeFileSync(filename, 'foo', 'utf-8');
  throw new Error('failure');
});

test('should work', async ({ }) => {
  test.skip();
});
```

Running this with `npx playwright test` yields a failed test right now; should it yield skipped instead?

We'll discuss this tomorrow with the team.

JackSpa89:
Hi @aslushnikov ,

thanks for answering. That is also a question to be discussed sure, but my point really was if the `beforeEach` hook fails for example the first time, but passes the second time. So to illustrate:

```javascript
test.beforeEach(async () => {
    const random = Math.random();
    expect(random).toBeGreaterThan(0.5); // some flaky step
});

test('should work', async ({ }) => {
    test.skip();
});

```
In this case:

 1. retry `beforeEach` fails
 2. retry `beforeEach` passes -> label test as skipped

Then the test is marked as failed, and I would expect the status to be skipped as the test generally works ( just flaky )

In this case:

 1. retry `beforeEach` fails
 2. retry `beforeEach` fails
 3. retry `beforeEach` fails

Then the test should still be marked as failed imo, as then there is something wrong with the test. 

If you got my question right the first time, never mind this post :) 


BR

aslushnikov:
As per discussion, this is not too important for us to prioritize, but we might change this to report the test as `flaky` in future.

[...]

PR #30276, opened 2024-04-06
fix(playwright): test unexpected result if failed and skipped

### LINKED ISSUE #28322: [BUG] Test mistakenly identified as flaky
### System info
- Playwright Version: v1.39.0
- Operating System: All
- Browser: All
- Other info:

### Source code

```ts
import { expect, test } from "@playwright/test";

test.describe.configure({ mode: "serial", retries: 3 });

test("A", async ({ page }, { retry }) => {
  if (retry === 0 || retry === 2) {
    throw new Error("oh!");
  }
  if (retry === 1 || retry === 3) {
    expect(true).toBe(true);
  }
});

test("B", async ({ page }, { retry }) => {
  throw new Error("oh!");
});
```

**Steps**

- Run the test `npx playwright test`
- See both tests identified as flaky
- CLI exit status is 0

`1.39` produces the following outcome for attempts:

- Test A: `failed`, `passed`, `failed`, `passed`
- Test B: `skipped`, `failed`, `skipped`, `failed`

**Expected**

- Test `B` should not be `flaky`, it should be `failed`. All the non-skipped attempts do not match the expected status - i.e. the test never produces the expected `passed` status.
- CLI exit status should not be 0 to mark 

**Actual**

- Both tests are identified as flaky
- CLI exit status is 0 and the CI passes

BeeMargarida:
Fixes #28322

**Changes**:
A test result should be unexpected if it fails (and is not expected) and skips the rest of the runs.
This MR implements this logic, so that runs in serial mode in specific cases (see added unit test) don't cause false positives.

### DIFF
--- a/packages/playwright/src/common/test.ts
+++ b/packages/playwright/src/common/test.ts
@@ -290,7 +290,7 @@ export class TestCase extends Base implements reporterTypes.TestCase {
       return 'skipped';
 
     const failures = results.filter(result => result.status !== 'skipped' && result.status !== 'interrupted' && result.status !== this.expectedStatus);
-    const skipped = results.filter(result => result.status === 'skipped');
+    const skipped = results.filter(result => result.status === 'skipped' && result.status !== this.expectedStatus);
     const passed = results.filter(result => result.status === 'passed');
     if (!failures.length) // all passed
       return 'expected';
--- a/packages/playwright/src/common/test.ts
+++ b/packages/playwright/src/common/test.ts
@@ -296,7 +296,7 @@ export class TestCase extends Base implements reporterTypes.TestCase {
       return 'expected';
     if (failures.length === results.length) // all failed
       return 'unexpected';
-    if (failures.length && skipped.length && !passed.length) // some failed, none succedded and the rest where skipped
+    if (failures.length && skipped.length && !passed.length) // some failed, none succeeded and the rest were skipped
       return 'unexpected';
     return 'flaky'; // mixed bag
   }
--- a/packages/playwright/src/common/test.ts
+++ b/packages/playwright/src/common/test.ts
@@ -290,9 +290,13 @@ export class TestCase extends Base implements reporterTypes.TestCase {
       return 'skipped';
 
     const failures = results.filter(result => result.status !== 'skipped' && result.status !== 'interrupted' && result.status !== this.expectedStatus);
+    const skipped = results.filter(result => result.status === 'skipped');
+    const passed = results.filter(result => result.status === 'passed');
     if (!failures.length) // all passed
       return 'expected';
     if (failures.length === results.length) // all failed
+      return 'unexpected';
+    if (failures.length && skipped.length && !passed.length) // some failed, none succedded and the rest where skipped
       return 'unexpected';
     return 'flaky'; // mixed bag
   }
--- a/packages/playwright/src/common/test.ts
+++ b/packages/playwright/src/common/test.ts
@@ -290,8 +290,7 @@ export class TestCase extends Base implements reporterTypes.TestCase {
       return 'skipped';
 
     const failures = results.filter(result => result.status !== 'skipped' && result.status !== 'interrupted' && result.status !== this.expectedStatus);
-    const skipped = results.filter(result => result.status === 'skipped' && result.status !== this.expectedStatus);
-    const passed = results.filter(result => result.status === 'passed');
+    const passed = results.filter(result => result.status === this.expectedStatus);
     if (!failures.length) // all passed
       return 'expected';
     if (failures.length === results.length) // all failed
--- a/packages/playwright/src/common/test.ts
+++ b/packages/playwright/src/common/test.ts
@@ -296,7 +296,7 @@ export class TestCase extends Base implements reporterTypes.TestCase {
       return 'expected';
     if (failures.length === results.length) // all failed
       return 'unexpected';
-    if (failures.length && skipped.length && !passed.length) // some failed, none succeeded and the rest were skipped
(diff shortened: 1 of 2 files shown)


BeeMargarida:

tests/playwright-test/test-serial.spec.ts
```diff
@@ -140,6 +140,40 @@ test('test.describe.serial should work with retry', async ({ runInlineTest }) =>
...
+        });
+        test('test4', async ({}) => {
+          console.log('\\n%%test4');
+        });
+      });
+    `,
+  }, { retries: 1 });
+  expect(result.exitCode).toBe(1);
+  expect(result.passed).toBe(0);
+  expect(result.flaky).toBe(1);
```
yury-s, line 167: I believe in this scenario at least two tests `test1` and `test2` are flaky, aren't they?

yury-s (changes requested):

packages/playwright/src/common/test.ts
```diff
@@ -290,10 +290,14 @@ export class TestCase extends Base implements reporterTypes.TestCase {
       return 'skipped';
 
     const failures = results.filter(result => result.status !== 'skipped' && result.status !== 'interrupted' && result.status !== this.expectedStatus);
+    const skipped = results.filter(result => result.status === 'skipped' && result.status !== this.expectedStatus);
+    const passed = results.filter(result => result.status === 'passed');
```
yury-s, line 294:
```suggestion
    const passed = results.filter(result => result.status === this.expectedStatus);
```

yury-s: We probably need a test for this.

packages/playwright/src/common/test.ts
```diff
@@ -290,10 +290,14 @@ export class TestCase extends Base implements reporterTypes.TestCase {
...
     if (!failures.length) // all passed
       return 'expected';
     if (failures.length === results.length) // all failed
       return 'unexpected';
+    if (failures.length && skipped.length && !passed.length) // some failed, none succeeded and the rest were skipped
```
yury-s, line 299: I believe what we need here is instead of lines 297-300 something like:

```js
if (failures.length && !passed.length)
  return 'unexpected';
```
as we don't care if there were any "skips".

tests/playwright-test/test-serial.spec.ts
yury-s: My apologies, read the bug again, this is an intentional change. I'll discuss in the team meeting if this is change that we want.

yury-s: Discussed this in the meeting, we'll consider all the implications and other possibilities and @dgozman will come back to you.

packages/playwright/src/common/test.ts
BeeMargarida: Fixed :+1:

dgozman:
@BeeMargarida Thank you for the PR! I've researched our past changes in this area, and it turns out things are a bit more complicated:
- We have the same logic that calculates `outcome()` in two places.
- We should also align the "did not run" counter in reporters with this logic.
- There are more tests to be added/changed.
- There is a long history of changes here:
  - Related issues: #28322, #28321, #27455, #17652.
  - Prior changes: #27762, #26385, #28360, probably more.

Given the above, I have decided that it would not be worth it to iterate on this topic together, so I have created a PR myself: https://github.com/microsoft/playwright/pull/30529, and will most likely close this one.

Thank you for sending the PR though! If you are still willing to contribute to Playwright, I'd recommend to pick an [issue marked as "open-to-a-pull-request"](https://github.com/microsoft/playwright/issues?q=is%3Aopen+is%3Aissue+label%3Aopen-to-a-pull-request) and start there.

BeeMargarida:
> @BeeMargarida Thank you for the PR! I've researched our past changes in this area, and it turns out things are a bit more complicated:
> 
>     * We have the same logic that calculates `outcome()` in two places.
> 
>     * We should also align the "did not run" counter in reporters with this logic.
> 
>     * There are more tests to be added/changed.
> 
>     * There is a long history of changes here:
>       
>       * Related issues: [[BUG] Test mistakenly identified as flaky  #28322](https://github.com/microsoft/playwright/issues/28322), [[BUG] Ambiguous test outcome and status for serial mode #28321](https://github.com/microsoft/playwright/issues/28321), [[BUG] Serial + Retry : Failed test is marked as flaky even if it never passes #27455](https://github.com/microsoft/playwright/issues/27455), [[Question] Skip on retry #17652](https://github.com/microsoft/playwright/issues/17652).
>       * Prior changes: [chore(test runner): remove fake skipped test results #27762](https://github.com/microsoft/playwright/pull/27762), [fix(test runner): failed + skipped = flaky #26385](https://github.com/microsoft/playwright/pull/26385), [Revert "chore(test runner): remove fake skipped test results (#27762)" #28360](https://github.com/microsoft/playwright/pull/28360), probably more.
> 
> 
> Given the above, I have decided that it would not be worth it to iterate on this topic together, so I have created a PR myself: #30529, and will most likely close this one.
> 
> Thank you for sending the PR though! If you are still willing to contribute to Playwright, I'd recommend to pick an [issue marked as "open-to-a-pull-request"](https://github.com/microsoft/playwright/issues?q=is%3Aopen+is%3Aissue+label%3Aopen-to-a-pull-request) and start there.

Hey, thank you for looking into this! :raised_hands:  I completely understand, it seems like a big change in the result logic that could have side effects, so it's okay to improve on this in another PR. As long as the bug is solved, it's a happy ending for me :smile: 

I'll take a look at the issues, always happy to contribute!

PR #30279, opened 2024-04-07
feat(text matchers): add ignoreOrder option

### LINKED ISSUE #22275: [Feature] ignoreOrder option for toHaveText

I would like to be able to assert with a locator that returns a list and not care about the order
`await expect(locator).toHaveText([1, 3, 2], ignoreOrder: true)`

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/22275

### DIFF
--- a/docs/src/api/class-locatorassertions.md
+++ b/docs/src/api/class-locatorassertions.md
@@ -217,6 +217,12 @@ Expected substring or RegExp or a list of those.
 
 Whether to perform case-insensitive match. [`option: ignoreCase`] option takes precedence over the corresponding regular expression flag if specified.
 
+### option: LocatorAssertions.NotToContainText.ignoreOrder
+* since: v1.44
+- `ignoreOrder` <[boolean]>
+
+Weather to ignore the order of the elements when comparing the text. This option is only available when the expected value is an array.
+
 ### option: LocatorAssertions.NotToContainText.useInnerText
 * since: v1.18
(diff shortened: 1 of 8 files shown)


packages/playwright-core/src/server/injected/injectedScript.ts
```diff
@@ -1268,13 +1268,26 @@ export class InjectedScript {
 
       // Each matcher should get a "received" that matches it, in order.
       const matchers = options.expectedText.map(e => new ExpectedTextMatcher(e));
-      let mIndex = 0, rIndex = 0;
-      while (mIndex < matchers.length && rIndex < received.length) {
-        if (matchers[mIndex].matches(received[rIndex]))
-          ++mIndex;
-        ++rIndex;
+      if (options.ignoreOrder) {
```
dgozman, line 1271: I'd just `sort()` if the order does not matter, and use the same code as before.

docs/src/api/class-locatorassertions.md
```diff
@@ -217,6 +217,12 @@ Expected substring or RegExp or a list of those.
 
 Whether to perform case-insensitive match. [`option: ignoreCase`] option takes precedence over the corresponding regular expression flag if specified.
 
+### option: LocatorAssertions.NotToContainText.ignoreOrder
```
dgozman, line 220: Not too many upvotes on the feature request - I wonder why implement it now?

PR #30284, opened 2024-04-07
replace full ASCII box w/ only top and bottom to allow copy

### LINKED ISSUE #30126: [Feature]: do not wrap command output in ascii box
### 🚀 Feature Request

When attempting to run playwright before the appropriate browser and system packages are installed, a command to run is printed in an ascii box.

image

When copying the command, the ascii box is selected as well. 

I propose delimiting the command with top and bottom bars to ensure it's more visible but _not_ a full box. 

### Example

_No response_

### Motivation

Streamlines the process for new users.

jfgreffier:
Fixes #30126

Example with 0 padding :

```bash
═══════════════════════
Hello
World

echo "This is line 1" \
"This is line 2" \
"This is line 3"

multilines ?
═══════════════════════
```

### DIFF
--- a/packages/playwright-core/src/utils/ascii.ts
+++ b/packages/playwright-core/src/utils/ascii.ts
@@ -18,9 +18,9 @@ export function wrapInASCIIBox(text: string, padding = 0): string {
   const lines = text.split('\n');
   const maxLength = Math.max(...lines.map(line => line.length));
   return [
-    '╔' + '═'.repeat(maxLength + padding * 2) + '╗',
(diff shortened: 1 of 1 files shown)


mxschmitt:
We tried something like this in the past and customers were not happy with how it looks, see here: https://github.com/microsoft/playwright/commit/2e484b46a215803135a2f8ad96d685ed113d8a2f.

Let's keep it for now as it is or remove the whole long apt-get install deps... list all-together.

PR #30285, opened 2024-04-08
feat: support ffmpeg  download url using customized host

stormslowly:
as the title.


mxschmitt:
Could you please start filing an issue with what your use-case is for this and why the generic `PLAYWRIGHT_DOWNLOAD_HOST` does not work for you?

stormslowly:
oops , i had not found the PLAYWRIGHT_DOWNLOAD_HOST before;
it works, thanks a lot

PR #30288, opened 2024-04-08
fix(docker): set default shell encoding

### LINKED ISSUE #28642: [BUG] filechooser.setFiles will fail when filename includes Chinese character
### System info
- Playwright Version: 1.40.1
- Operating System: mcr.microsoft.com/playwright:v1.40.0-jammy
- Browser: Chromium

### Source code

[here is reproduce repository](https://github.com/xiefenga/playwright-filechooser-bug-reproduce)

**Steps**
- pull playwright docker image
- run the tests in docker

**Expected**
pass the test

**Actual**
the filechooser.setFiles operation of the file which filename includes Chinese character will fail, then cause expect fail

difference between `filechooser.setFiles('trace.zip')` and `filechooser.setFiles('压缩文件.zip')`

test---trace-1-chromium-linux

test---压缩文件-1-chromium-linux

### LINKED ISSUE #30245: [Docs]: Add a note to the Docker page concerning special characters
### Page(s)

https://playwright.dev/docs/docker
https://playwright.dev/docs/docker#run-the-image

### Description

When simply running the provided playwright image, the Linux `locale` will be set to `POSIX`.
This will cause issues with certain special characters (like §), Umlauts (ä,ö,ü) and accented characters (è,â,ç).
You won't be able to write them out in the command line and when copying files containing such characters in their filenames or contents, they will be escaped with number codes or blanks which is a problem if you need them within your tests..
My solution to this was to create a Dockerfile and adding `ENV LANG=C.UTF-8` so that these characters can be used and displayed properly.

The problem and a possible solution should be noted at the appropriate place in the documentation.

Reproduction:
Run the latest image (currently v1.42.1).
Then, try to type `touch Bei§piel.txt` - it won't work properly for the §-character
You can also try to create that file on your host system and copy it to the container.
When you perform ls in the container, it should look like this: `'Bei'$'\302\247''piel.txt'`

mxschmitt:
This seems to be common practise, as per [here](https://github.com/search?q=org%3Amicrosoft%20%2FENV%20LANG%3DC.UTF-8%2F&type=code).

Fixes https://github.com/microsoft/playwright/issues/30245
Fixes https://github.com/microsoft/playwright/issues/28642

### DIFF
--- a/utils/docker/Dockerfile.focal
+++ b/utils/docker/Dockerfile.focal
@@ -4,6 +4,9 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG TZ=America/Los_Angeles
 ARG DOCKER_IMAGE_NAME_TEMPLATE="mcr.microsoft.com/playwright:v%version%-focal"
 
+ENV LANG=C.UTF-8
+ENV LC_ALL=C.UTF-8
+
 # === INSTALL Node.js ===
 
 RUN apt-get update && \
--- a/utils/docker/Dockerfile.jammy
+++ b/utils/docker/Dockerfile.jammy
@@ -4,6 +4,9 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG TZ=America/Los_Angeles
 ARG DOCKER_IMAGE_NAME_TEMPLATE="mcr.microsoft.com/playwright:v%version%-jammy"
 
+ENV LANG=C.UTF-8
+ENV LC_ALL=C.UTF-8
+
 # === INSTALL Node.js ===
 
 RUN apt-get update && \


utils/docker/Dockerfile.focal
```diff
@@ -4,6 +4,8 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG TZ=America/Los_Angeles
 ARG DOCKER_IMAGE_NAME_TEMPLATE="mcr.microsoft.com/playwright:v%version%-focal"
 
+ENV LANG=C.UTF-8
```
dgozman, line 7: Other docker files also have `LC_ALL=C.UTF-8`. Not sure whether that's relevant.

utils/docker/Dockerfile.jammy
```diff
@@ -4,6 +4,8 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG TZ=America/Los_Angeles
 ARG DOCKER_IMAGE_NAME_TEMPLATE="mcr.microsoft.com/playwright:v%version%-jammy"
 
+ENV LANG=C.UTF-8
```
dgozman, line 7: Could this breaking in some situations? Let's make sure we have this in the miscellaneous list in the release notes.

dgozman (approved):

PR #30313, opened 2024-04-09
Update auth.md with small improvement to C# example

thatstatsguy:
- better indication on how to reference the newly created `playwright/.auth` file relative to where tests are executing

### DIFF
--- a/docs/src/auth.md
+++ b/docs/src/auth.md
@@ -305,7 +305,7 @@ await context.StorageStateAsync(new()
 // Create a new context with the saved storage state.
 var context = await browser.NewContextAsync(new()
 {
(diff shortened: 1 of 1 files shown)


mxschmitt:
Isn't it even better to have the `state.json` inside the bin folder? Then the user never need to worry to add it to the gitignore file.

thatstatsguy:
true - but what I'd do for this is remove the following part of the docs for c#

> create a `playwright/.auth` folder and add it to the git ignore 

since it's actually not needed (I assumed this had to be there for it to work). I assumed playwright would somehow automatically pick up that folder in the parent folder structure for the tests.

mxschmitt (approved):

PR #30347, opened 2024-04-11
Playwright Recorder webextension for Chrome / Edge and Firefox

ruifigueira:
I was able to set up a basic PoC for playwright recorder as a webextension, and currently runs in Chrome, Edge and Firefox.

Instead of trying to run the whole playwright stack as a webextension, it only runs the necessary bits from `recorder.ts`, which already has lots of isomorphic code. For instance, alll code generation logic is almost isomorphic right now, except the usage of [EventEmitter](https://www.npmjs.com/package/events), which is the only polyfill I had to import.

In terms of webextensions APIs, I'm only using chrome APIs that are compatible with firefox, namely:

* [chrome.runtime](https://developer.chrome.com/docs/extensions/reference/api/runtime) / [browser.runtime](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime)
* [chrome.tabs](https://developer.chrome.com/docs/extensions/reference/api/tabs) / [browser.tabs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs)
* [chrome.action](https://developer.chrome.com/docs/extensions/reference/api/action) / [browser.action](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_support_for_JavaScript_APIs#action)
* [chrome.webNavigation](https://developer.chrome.com/docs/extensions/reference/api/webNavigation) / [browser.webNavigation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation)

For the purpose of this PoC, I implemented lighter versions of server `page`, `frame` and `browserContext` with only the necessary bits, and copied the server `recorder.ts` to remove all the things that don't make sense for the webextension.

TLDR: most of the code needed for the recorder to work is already isomorphic or only needs some minor adjustments. Non-isomorphic code can be replaced with implementations that rely on webextension APIs.

### Firefox Demo:

https://github.com/microsoft/playwright/assets/1374559/0eb7da9d-0c60-4aa1-a699-36b9c5695b5c

### Chrome Demo:

https://github.com/microsoft/playwright/assets/1374559/d1c60c66-67c2-458e-a7cb-7769a2c4fcb8


pavelfeldman:
Archiving this

PR #30367, opened 2024-04-14
chore: align crServiceWorker with crPage

dgozman:
Simplify network-related methods because crNetworkManager already handles initial/non-initial calls.

### DIFF
--- a/packages/playwright-core/src/server/chromium/crServiceWorker.ts
+++ b/packages/playwright-core/src/server/chromium/crServiceWorker.ts
@@ -16,18 +16,15 @@
 import { Worker } from '../page';
 import type { CRBrowserContext } from './crBrowser';
 import type { CRSession } from './crConnection';
-import type * as types from '../types';
 import { CRExecutionContext } from './crExecutionContext';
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/chromium/crServiceWorker.ts
```diff
@@ -61,41 +58,28 @@ export class CRServiceWorker extends Worker {
...
   }
 
-  async updateExtraHTTPHeaders(initial: boolean): Promise<void> {
+  async updateExtraHTTPHeaders(): Promise<void> {
     if (!this._isNetworkInspectionEnabled())
       return;
-
-    const headers = network.mergeHeaders([
-      this._browserContext._options.extraHTTPHeaders,
-      this._extraHTTPHeaders,
```
yury-s, line 88: why did this go away?

yury-s (approved):

packages/playwright-core/src/server/chromium/crServiceWorker.ts
dgozman: This field was always `null`, so I removed it.

PR #30385, opened 2024-04-16
feat(chromium): roll to r1114

playwrightmachine:


dlockhart:
These Chromium updates aren't resulting in a NPM release -- would it be possible to release one of them? There were some regressions that Google has since fixed in `124.0.6367.49` that would be great to get into Playwright.

mxschmitt:
@dlockhart could you provide more details what they broke? Please file a new issue.

In the meantime, you can `channel: 'chrome'` or `channel: 'chrome-beta'`.

See here: https://playwright.dev/docs/browsers#google-chrome--microsoft-edge

dlockhart:
Thanks @mxschmitt. The defect [is this one](https://issues.chromium.org/issues/333414300), which was reverted and fixed in `124.0.6367.49`.

Issue #17969, opened 2022-10-10
[Feature] Split shards via test timing data


ffluk3:
Playwright has some great default behavior around sharding tests across multiple workers. It would be super helpful if the shard can take in a `--timing-data` file similar to CircleCI's [split tests logic](https://circleci.com/docs/parallelism-faster-jobs/#splitting-by-timing-data) so we can let Playwright internally split tests across the timing boundaries. This would allow Playwright to complete as fast as possible by splitting tests more optimally across multiple machines or workers.

michaelhays:
This is the one thing that has made me hesitate switching from Cypress to Playwright despite the many advantages.

Cypress Cloud solves this with their [Smart Orchestration](https://docs.cypress.io/guides/cloud/smart-orchestration) (specifically, the "Load Balancing" strategy).

kp-abhishek-agrawal:
This is one of the major limitation of playwright.

ofirpardo-artlist:
I'd also love to see this feature. I thought of creating a PR with the following additions:

- Add '--timing-file' flag, e.g: --timing-file="report.json"
- The file will have the duration and id for each test
- Based on the duration we can build sets of as close as possible duration, then assign the test groups based on the test id.

So far I tried to use the json reporter and extract this data from it which seems to work, but it's a bit messy because of the 'infinite' amount of suites that you need to go over. Not sure if I'd want to generate another smaller timing file from the json reporter, or just use the json report file and extract the data for it in the playwright runner.
Also thought of creating a completely new reporter which will only include the test name, test id and duration.

Based on my testing so far, we are able to create a pretty good balance for any amount of shards.
I believe more details are required, but please let me know whether this is something that could be implemented, otherwise I'd just create a custom external solution instead.

Thanks.

ofirpardo-artlist:
A small scale example, current implementation:
![image](https://github.com/microsoft/playwright/assets/110810683/d0015154-5210-4a72-bc3c-b761a945fba6)
![image](https://github.com/microsoft/playwright/assets/110810683/8281cc3b-3742-4021-849d-2c1ac613a828)
![image](https://github.com/microsoft/playwright/assets/110810683/d3c639ca-fb46-475e-94d6-035fbe7e54c8)

New balance based on json report:
![image](https://github.com/microsoft/playwright/assets/110810683/fec55b79-7c0e-4a51-a1b9-0970ad70b627)
[...]

PR #30388, opened 2024-04-16
feat(test runner): Use timing file to balance shards

ofirpardo-artlist:
This feature came from this issue: https://github.com/microsoft/playwright/issues/17969

The idea is to use the JSON reporter and extract the duration of each test, then map the test groups based on the duration so the shards are all as equal as possible, the amount of tests per shard is not relevant here, only the duration of the tests.
If there are tests that are not recorded in the JSON report they will be distributed like they currently do based on the amount.

This is still in a POC state to get some feedback, but the functionality works well.

Some open questions:
- Not sure about the behavior of non-parallel tests/files
- Optionally adding a config flag as well, currently it's only available from CI
- Do we want to use the JSON report, or create a custom 'TestDetails' report which only includes the necessary data

At the end of the day unless you come up with some cloud solution, the 'infra' for this feature would need to be done mostly manually, a static JSON report is optional but might get irrelevant very fast.
On my project I've managed it by averaging the durations of the last 10 successful runs, which is being updated to a CDN from CI.

Some more details on current POC can be seen here:
https://github.com/microsoft/playwright/issues/17969#issuecomment-2037376375
https://github.com/microsoft/playwright/issues/17969#issuecomment-2042791382

### DIFF
--- a/packages/playwright/src/common/config.ts
+++ b/packages/playwright/src/common/config.ts
@@ -57,6 +57,7 @@ export class FullConfigInternal {
   cliPassWithNoTests?: boolean;
   testIdMatcher?: Matcher;
   defineConfigWasUsed = false;
+  cliTimingFile?: string;
 
   constructor(location: ConfigLocation, userConfig: Config, configCLIOverrides: ConfigCLIOverrides) {
     if (configCLIOverrides.projects && userConfig.projects)
--- a/packages/playwright/src/program.ts
+++ b/packages/playwright/src/program.ts
@@ -189,6 +189,7 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
   config.cliListOnly = !!opts.list;
   config.cliProjectFilter = opts.project || undefined;
   config.cliPassWithNoTests = !!opts.passWithNoTests;
+  config.cliTimingFile = opts.timingFile || undefined;
 
   const runner = new Runner(config);
   let status: FullResult['status'];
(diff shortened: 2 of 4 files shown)


ofirpardo-artlist:

anujshah3:
Hi @ofirpardo-artlist 
Is there an expected release timeline for this PR? This is a much needed feature!

pavelfeldman:
Archiving this

ofirpardo-artlist:
> Archiving this

Was this closed due to inactivity? Or due to other reasons?
I was basically waiting for some feedback. Thanks.

ofirpardo-artlist:
> These usually start with filing an issue and figuring out the scope of the feature, what's feasible, edge cases like sharding, merging, CI, etc. This is a big feature and it is unlikely that the contribution will be in a form of a PR.

I understand that, but there is an open issue that's completely inactive and was created 2 years, so lets either make it active somehow by trying to answer the unknowns you mentioned, or find some other solution perhaps to see how we could follow the steps necessary.
As it is right now, nothing will happen, which I think is a shame especially when there is some working demo that we can use. Personally we already use it in production and it works nicely, and I'm sure it can be even better.

Hope to see it continues, thanks :)

pezholio:
For what it's worth, it seems pretty disrespectful of author's time and effort to just ignore the work and then close the PR a few years later with no context or explanation. I understand that there's probably procedures that need to be gone through to get a fairly big piece of work like this shipped (and equally, I know managing an open source project is not free or easy!), but it'd be good to set that expectation at the start of the process.

PR #30402, opened 2024-04-17
fix(role): explicitly hidden aria-labelledby should be recursively traversed

### LINKED ISSUE #29796: [Bug]: `getByRole` unable to find button with accessible name.
### Version

1.42.1

### Steps to reproduce

`getByRole('button', { name: "Name of my button" })` fails to find my button which has an accessible name.

Specifically, my button has an `aria-labelledby`, and receives its accessible name from a custom tooltip element.

1. Clone my repo at: https://github.com/khiga8/playwright-get-by-role-bug
2. Run `npm install` and `npx install playwright`
3. Run `npm run test` and see test failure.

### Expected behavior

I expect to see Playwright be able to locate my button because it has an `aria-labelledby` that points to an existing element.

### Actual behavior

Playwright is unable to locate my button and times out.

### Additional context

_No response_

### Environment

  System:
    OS: macOS 14.3
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 5.61 GB / 64.00 GB
  Binaries:
    Node: 20.2.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 9.6.6 - /usr/local/bin/npm
  IDEs:
    VSCode: 1.86.1 - /usr/local/bin/code
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.42.1 => 1.42.1 
    playwright: ^1.42.1 => 1.42.1

dgozman:
The accessible name computation spec has changed to explicitly mention this case:

Step 2A. Hidden Not Referenced. If the current node is hidden and is:
- Not part of an aria-labelledby or aria-describedby traversal, where the node directly referenced by that relation was hidden.
- Nor part of a native host language text alternative element (e.g. label in HTML) or attribute traversal, where the root of that traversal was hidden.

See https://w3c.github.io/accname/#computation-steps. Chromium, Firefox and Safari all agree with the spec.

Fixes #29796.

### DIFF
--- a/packages/playwright-core/src/server/injected/roleUtils.ts
+++ b/packages/playwright-core/src/server/injected/roleUtils.ts
@@ -389,9 +389,9 @@ export function getElementAccessibleName(element: Element, includeHidden: boolea
       accessibleName = normalizeAccessbileName(getElementAccessibleNameInternal(element, {
         includeHidden,
         visitedElements: new Set(),
-        embeddedInLabelledBy: 'none',
-        embeddedInLabel: 'none',
-        embeddedInTextAlternativeElement: false,
+        embeddedInLabelledBy: undefined,
+        embeddedInLabel: undefined,
+        embeddedInNativeTextAlternative: undefined,
         embeddedInTargetElement: 'self',
       }));
     }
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/server/injected/roleUtils.ts
```diff
@@ -445,7 +449,8 @@ function getElementAccessibleNameInternal(element: Element, options: AccessibleN
   const role = getAriaRole(element) || '';
 
   // step 2c.
-  if (options.embeddedInLabel !== 'none' || options.embeddedInLabelledBy !== 'none') {
+  // TODO: should we check embeddedInLabel here?
```
yury-s, line 452: is it still relevant?

yury-s (approved):

PR #30403, opened 2024-04-17
fix(role): embedded control inside the target element

### LINKED ISSUE #28848: [BUG] getByRole("heading") can't match name from input value contents
### System info
- Playwright Version: v1.40.1
- Operating System: All
- Browser: Chromium
- Other info:

### Source code

```js
test("heading should be located", async ({ page }) => {
  await page.setContent(`
<h1>
  <input type="text" value="Foo bar">
</h1>
  `);

  await expect(page.getByRole("heading", { name: "Foo bar" })).toBeVisible();
  // Fails to locate
});
```

**Steps**
- Run the test code

**Expected**

The `<h1>` should be located successfully. Chromium DevTools Accessibility inspector shows that the `<h1>`'s accName is `"Foo bar":

image

**Actual**

Fails to locate.

### Notes

`getByRole` *is* able to locate headings that get their accNames from content (ex: `<h1><a href="#">Foo bar</a></h1>` locates with `page.getByRole("heading", { name: "Foo bar" })` successfully.

Is Playwright actually consulting Chromium's accessibility tree for `getByRole`?

edit: I understand better now, no Playwright is not using [CDP's Accessibility Domain](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/). It's [roleSelectorEngine](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/injected/roleSelectorEngine.ts) is an injected script, working entirely in JS/DOM land attempting to mimic the browser's own accessibility tree computations.

I guess that makes sense, certainly more x-browser friendly. But sure seems like a tough job to match the browser 1:1, accessibility computations are so dang complicated (and often evolving). This approach also may provide a slight false sense of security as it could successfully locate an element that the browser actually botched the accessibility calc for (due to regression or quirk).

I'd be great if it was possible to [extend selectors](https://playwright.dev/docs/extensibility) with a **non** injected engine.

dgozman:
According to the spec, such controls should use the native value as long as they have "aria-label". The relevant spec section is 2D.

However, there is an open issue that claims this should always apply, and all browsers and wpt test actually do that: https://github.com/w3c/accname/issues/64.

Fixes #28848.

### DIFF
--- a/tests/assets/axe-core/accessible-text.js
+++ b/tests/assets/axe-core/accessible-text.js
@@ -186,7 +186,13 @@ module.exports = [
       '<input type="text" id="tb1"></div>' +
       '<label for="tb1">My form input</label>',
     target: '#target',
-    accessibleText: 'My form input',
+    // accessibleText: 'My form input',
+    // All browsers and the spec (kind of) agree that input inside the target element should
+    // use it's value as an "embedded control", rather than a label.
+    // From the spec:
+    //   If traversal of the current node is due to recursion and the current node
+    //   is an embedded control, ignore aria-label and skip to rule Embedded Control.
+    accessibleText: '',
   },
 
   {
--- a/packages/playwright-core/src/server/injected/roleUtils.ts
+++ b/packages/playwright-core/src/server/injected/roleUtils.ts
@@ -433,7 +433,10 @@ function getElementAccessibleNameInternal(element: Element, options: AccessibleN
 
   const labelledBy = getAriaLabelledByElements(element);
 
-  // step 2b.
+  // step 2b. LabelledBy:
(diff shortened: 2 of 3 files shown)


tests/assets/axe-core/accessible-text.js
```diff
@@ -186,7 +186,13 @@ module.exports = [
       '<input type="text" id="tb1"></div>' +
       '<label for="tb1">My form input</label>',
     target: '#target',
-    accessibleText: 'My form input',
+    // accessibleText: 'My form input',
```
yury-s, line 189: remove this line?

yury-s (approved):

PR #30426, opened 2024-04-19
fix: pass key attribute from jsx to component test

tanhauhau:
When using the `key` attribute in jsx inside the test modules, it is not serialised and passed to the browser in component test

### DIFF
--- a/tests/playwright-test/playwright.ct-react.spec.ts
+++ b/tests/playwright-test/playwright.ct-react.spec.ts
@@ -250,6 +250,57 @@ test('should work with JSX in variable', async ({ runInlineTest }) => {
   expect(result.passed).toBe(1);
 });
 
+test('should pass "key" attribute from JSX in variable', async ({ runInlineTest }) => {
+  const result = await runInlineTest({
+    'playwright.config.ts': playwrightCtConfigText,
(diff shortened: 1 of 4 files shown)


tests/playwright-test/playwright.ct-react.spec.ts
```diff
@@ -250,6 +250,57 @@ test('should work with JSX in variable', async ({ runInlineTest }) => {
...
+
+    'src/index.test.jsx': `
+      import { test, expect } from '@playwright/experimental-ct-react';
+      import { Button } from './button';
+      import { Container } from './container';
+
+      test('key should tear down and recreate component', async ({ mount }) => {
+        const component = await mount(
+          <Container>
+            <Button key='a' value={5} />
```
yury-s, line 288: Let's make the test more clear by using different values in this two buttons, like 1 and 10, so that the expectations below are more obvious.

yury-s (commented):
Please fix the style issues that lint found.

tests/playwright-test/playwright.ct-react.spec.ts
tanhauhau: updated

yury-s (approved):

Issue #18332, opened 2022-10-25
[Feature] locator for getting an element's description


kentcdodds:
I'm testing an error message.

Here's what I have to do now:

```ts
  const subjectDescriptionId = await page
    .getByRole('textbox', {name: /subject/i})
    .getAttribute('aria-describedby')
  await expect(page.locator(`id=${subjectDescriptionId}`)).toHaveText(/required/i)
```

Here's what I'd like to be able to do:

```ts
  const subjectDescription = await page
    .getByRole('textbox', {name: /subject/i})
    .getDescription()
  await expect(subjectDescription).toHaveText(/required/i)
```

Thanks!

mxschmitt:
Is there something similar in testing library? 

I like the feature request to get the aria describedby element!

kentcdodds:
Unfortunately we don't have anything like this in Testing Library either. Definitely would be a good one to have though.

pavelfeldman:
#### Why was this issue closed?

Thank you for your involvement. This issue was closed due to limited engagement (upvotes/activity), lack of recent activity, and insufficient actionability. To maintain a manageable database, we prioritize issues based on these factors.

If you disagree with this closure, please open a new issue and reference this one. More support or clarity on its necessity may prompt a review. Your understanding and cooperation are appreciated.

kentcdodds:
This is one of those things people don't know they need (hence the lower up-votes) but will use once they have.

WickyNilliams:
i think it would be useful to get the element's description, but also to locate by description. similar to `getByLabel` a `getByDescription` would be useful

lukasgebhard:
[@testing-library/jest-dom](https://github.com/testing-library/jest-dom) has three matchers specifically for accessibility testing:

1. https://github.com/testing-library/jest-dom#tohaveaccessibledescription
2. https://github.com/testing-library/jest-dom#tohaveaccessibleerrormessage
3. https://github.com/testing-library/jest-dom#tohaveaccessiblename

We've been using this API extensively, and would be happy to have this or something similar with Playwright, too!

kristojorg:
This would be very nice to have in playwright!

trieu-h:
Upvoting this since I'm running into a situation where having this feature would save me some time.

camchenry:
I would also find this quite useful. We are using Playwright for testing our UI pretty extensively, and this is very helpful for ensuring that the screen reader experience is tested automatically (to some extent). 

[...]

PR #30434, opened 2024-04-19
feat(role): roll and pass WPT accname tests, calculate description

dgozman:
There are new "non-manual" WPT accname tests that we now mostly pass, which required a few tweeks in calculating role and name.

Also implemented accessible description computation, which is just a small addition on top of accessible name, and passed respective wpt tests.

References #18332.


yury-s (approved):

tests/assets/wpt/README.md
```diff
@@ -1,10 +1,12 @@
-Web platform tests are copied from https://github.com/web-platform-tests/wpt.
+Web platform tests are copied from https://github.com/web-platform-tests/wpt at revision `5adfe4e8cd223729aa2942915e8a515c079ed0ef`.
```
mxschmitt, line 1: nit: a `roll.sh` script would be nice there (which does e.g. git clone, rm -rf .git, only keep the relevant folder). Since its only a subfolder?

mxschmitt (approved):

tests/assets/wpt/README.md
dgozman: So far I did it once, so I'll wait for the second time 😄

PR #30437, opened 2024-04-19
Try running audio/video tests on more browsers and platforms

stephenwade:
I'm trying to debug a Playwright issue with `AudioContext` on Windows runners (will file a separate issue if I can't figure it out), and I'm interested to see if Playwright's audio/video tests now work on all platforms and browsers.

This PR removes some `it.fixme` annotations from the audio/video tests in `capabilities.spec.ts`. All the `it.fixme` lines I removed were added more than a year ago, and all the linked issues look like they are resolved.

Related: https://github.com/microsoft/playwright/pull/30410

### DIFF
--- a/tests/library/capabilities.spec.ts
+++ b/tests/library/capabilities.spec.ts
@@ -64,13 +64,7 @@ it('should respect CSP @smoke', async ({ page, server }) => {
   expect(await page.evaluate(() => window['testStatus'])).toBe('SUCCESS');
 });
 
-it('should play video @smoke', async ({ page, asset, browserName, platform, mode }) => {
-  // TODO: the test passes on Windows locally but fails on GitHub Action bot,
-  // apparently due to a Media Pack issue in the Windows Server.
-  // Also the test is very flaky on Linux WebKit.
-  it.fixme(browserName === 'webkit' && platform !== 'darwin');
-  it.fixme(browserName === 'firefox', 'https://github.com/microsoft/playwright/issues/5721');
-  it.fixme(browserName === 'webkit' && platform === 'darwin' && parseInt(os.release(), 10) === 20, 'Does not work on BigSur');
+it('should play video @smoke', async ({ page, asset, browserName, mode }) => {
   it.skip(mode.startsWith('service'));
 
   // Safari only plays mp4 so we test WebKit with an .mp4 clip.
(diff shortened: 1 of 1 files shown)


stephenwade:

stephenwade:
cc @mxschmitt I think this needs a tag to run the affected tests.

mxschmitt:
Done!

stephenwade:
@mxschmitt Looks like "tests 2" didn't run when I pushed another commit, and the Installation Test needs to be re-run. Could you help?

stephenwade:
@mxschmitt Looks like the "should play audio" and "should play video" tests are flaky on `macos-12`. This lines up with what I found while testing audio playback on GitHub Actions - for some reason, they work on `macos-11`, `macos-13`, and `macos-14` but work only about half the time on `macos-12`. What do you think about skipping these tests on that macOS version?

stephenwade:
Hey @mxschmitt, just following up on this. I pushed a commit to skip the "should play audio" and "should play video" tests on Webkit on macos-12, since that's where I have observed flakiness. Please re-add the label to re-run the tests 😀

tests/library/capabilities.spec.ts
```diff
@@ -84,8 +83,7 @@ it('should play video @smoke', async ({ page, asset, browserName, platform, mode
 });
 
 it('should play webm video @smoke', async ({ page, asset, browserName, platform, mode }) => {
-  it.fixme(browserName === 'webkit' && platform === 'darwin' && parseInt(os.release(), 10) === 20, 'Does not work on BigSur');
-  it.fixme(browserName === 'webkit' && platform === 'win32');
+  it.fixme(browserName === 'webkit' && platform !== 'darwin');
```
mxschmitt, line 86: so this means it doesn't work on linux?

tests/library/capabilities.spec.ts
```diff
@@ -97,9 +95,9 @@ it('should play webm video @smoke', async ({ page, asset, browserName, platform,
 });
 
 it('should play audio @smoke', async ({ page, server, browserName, platform }) => {
-  it.fixme(browserName === 'firefox' && platform === 'win32', 'https://github.com/microsoft/playwright/issues/10887');
-  it.fixme(browserName === 'firefox' && platform === 'linux', 'https://github.com/microsoft/playwright/issues/10887');
-  it.fixme(browserName === 'webkit' && platform === 'win32', 'https://github.com/microsoft/playwright/issues/10892');
+  it.fixme(browserName === 'webkit' && platform !== 'darwin');
```
mxschmitt, line 98: it wasn't fixme on linux before?

lets try to stick with === rather than !== I think thats easier to read.

mxschmitt:
Re-triggered the bot. Also landed this https://github.com/microsoft/playwright/pull/30634 which should make things simpler.

pavelfeldman:
Archiving this

Issue #19933, opened 2023-01-07
[Feature] improve type safety hooksConfig


sand4rt:
The idea comes from Tanner: https://youtu.be/DXbfDKoMzWY?t=1467. 

```typescript
// ./playwright/index.tsx
import { beforeMount } from '@sand4rt/experimental-ct-web/hooks';

// Override HooksConfig so there is no need to specify it as a `mount<HooksConfig>` generic
declare module '@playwright/experimental-ct-solid/hooks' { 
  interface RegisterHooksConfig {
    route: string;
  }
}

beforeMount(async ({ hooksConfig }) => {
  hooksConfig.DOES_NOT_EXSISTS; // throws error
});
```

```typescript
// App.test.tsx
import { expect, test } from "@playwright/experimental-ct-solid";
import { App } from "./app";

test('type check', async ({ mount }) => {
  const component = await mount(<App />, { 
    hooksConfig: { DOES_NOT_EXSISTS: '' } // is typesafe without specifying the `mount<HooksConfig>` generic
  });

  await expect(component.getByText('label')).toBeVisible();
});
```

sand4rt:
@aslushnikov could the component tag be added to this ticket?

segevfiner:
I have hit another issue. `HooksConfig` is defined to `extends JsonObject` in the signature of `mount` which prevents using it with other more restricted types:
```ts
Type 'HooksConfig' does not satisfy the constraint 'JsonObject'.
  Index signature for type 'string' is missing in type 'HooksConfig'.ts(2344)
```

segevfiner:
OK this might be due to older Playwright version though. Couldn't upgrade yet due to https://github.com/microsoft/playwright/issues/23505

pavelfeldman:
#### Why was this issue closed?

Thank you for your involvement. This issue was closed due to limited engagement (upvotes/activity), lack of recent activity, and insufficient actionability. To maintain a manageable database, we prioritize issues based on these factors.

If you disagree with this closure, please open a new issue and reference this one. More support or clarity on its necessity may prompt a review. Your understanding and cooperation are appreciated.

PR #30455, opened 2024-04-21
feat(ct): resolve hooksConfig import refs

### LINKED ISSUE #30453: [Feature]: ability to pass components to `HooksConfig`
### 🚀 Feature Request

The ability to pass components to `HooksConfig`, probably by using the [`resolveImportRef`](https://github.com/microsoft/playwright/blob/main/packages/playwright-ct-core/src/injected/importRegistry.ts#L35) to resolve the [`hooksConfig`](https://github.com/microsoft/playwright/blob/main/packages/playwright-ct-core/src/mount.ts#L111C11-L111C33); 

### Example

```ts
beforeMount(async ({ app, hooksConfig }) => {
  for (const [name, component] of Object.entries(hooksConfig?.components || {}))
    app.component(name, component);
});
```

```ts
import { test, expect } from '@playwright/experimental-ct-vue';
import DefaultSlot from '@/components/DefaultSlot.vue';
import Button from '@/components/Button.vue';

test('render a component as slot', async ({ mount }) => {
  const component = await mount(DefaultSlot, {
    slots: {
      default: '<Button title="Submit" />',
    },
    hooksConfig: {
      components: { Button }
    }
  });
  await expect(component).toContainText('Submit');
});
```

### Motivation

It simplifies testing slots and other use cases such as providing plugins

sand4rt:
closes https://github.com/microsoft/playwright/issues/30453 and https://github.com/microsoft/playwright/issues/19933


packages/playwright-ct-core/types/component.d.ts
```diff
@@ -14,11 +14,6 @@
  * limitations under the License.
  */
 
-type JsonPrimitive = string | number | boolean | null;
-type JsonValue = JsonPrimitive | JsonObject | JsonArray;
-type JsonArray = JsonValue[];
-export type JsonObject = { [Key in string]?: JsonValue };
```
sand4rt, line 20: NOTE: `JsonObject` becomes obsolete because components can be passed to `hooksConfig`.

packages/playwright-ct-react/hooks.d.ts
```diff
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-import type { JsonObject } from '@playwright/experimental-ct-core/types/component';
+export interface RegisterHooksConfig {}
```
sand4rt, line 17: NOTE: `RegisterHooksConfig` is specified for [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html). See https://github.com/microsoft/playwright/issues/19933 for more info

sand4rt:
👋 @pavelfeldman, yes please. Let me know if more info is needed

pavelfeldman:
Tried reviewing it and failed - sounds like too many things are happening at a time. Any chance for it to be split in chunks? Also spotted a declare module there, we have general distaste to those, we like types to chain / infer explicity instead.

sand4rt:
> Tried reviewing it and failed - sounds like too many things are happening at a time. Any chance for it to be split in chunks? Also spotted a declare module there, we have general distaste to those, we like types to chain / infer explicitly instead.

Sure, no problem. I split the first part in https://github.com/microsoft/playwright/pull/31024. Will do the second part later on because they depend on eachother.

I also usually distaste `declare module` too, but i think this is a really good use case for them, and they are entirely optional. chain / infer explicitly is also still possible.

Issue #29854, opened 2024-03-07
[Bug]: Playwright CT 1.42.0 throws Vite/Rollup error when testing MUI TextField component


bethmaloney:
### Version

1.42.0

### Steps to reproduce

1. Install [MUI](https://mui.com/material-ui/getting-started/installation/)

2. Create the below test
```tsx
import { test } from "@playwright/experimental-ct-react";
import { TextField } from "@mui/material";

test("this test causes vite/rollup failure", async ({ mount }) => {
    await mount(<TextField />);
  });
```

3. Run `npm run test-ct`

### Expected behavior

Test builds and executes

### Actual behavior

Build fails with below output

```
> playwright-ct-example@0.0.0 test-ct
> playwright test -c playwright-ct.config.ts


Running 6 tests using 6 workers

vite v5.1.5 building for production...
✓ 3 modules transformed.
Error: [vite]: Rollup failed to resolve import "undefined" from "C:/Users/xxxx/source/repos/playwright-ct-example/playwright/index.tsx".                                                            
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`

    at viteWarn (C:\Users\xxxx\source\repos\playwright-ct-example\node_modules\vite\dist\node\chunks\dep-G-px366b.js:67108:27)
    at onwarn (C:\Users\xxxx\source\repos\playwright-ct-example\node_modules\@vitejs\plugin-react\dist\index.mjs:250:9)
    at onRollupWarning (C:\Users\xxxx\source\repos\playwright-ct-example\node_modules\vite\dist\node\chunks\dep-G-px366b.js:67133:9)
    at onwarn (C:\Users\xxxx\source\repos\playwright-ct-example\node_modules\vite\dist\node\chunks\dep-G-px366b.js:66840:13)
    at C:\Users\xxxx\source\repos\playwright-ct-example\node_modules\rollup\dist\es\shared\node-entry.js:18300:13
    at Object.logger (C:\Users\xxxx\source\repos\playwright-ct-example\node_modules\rollup\dist\es\shared\node-entry.js:19957:9)
    at ModuleLoader.handleInvalidResolvedId (C:\Users\xxxx\source\repos\playwright-ct-example\node_modules\rollup\dist\es\shared\node-entry.js:18901:26)
[...]

PR #30458, opened 2024-04-22
fix: support import components from node_modules

tanhauhau:


dgozman (commented):
Thank you for the PR! This has already been fixed by #30493 in the same way, so I'll have to close it. Next time, please attribute a PR to an issue it is fixing by adding `Fixes #29854` to the description.

Issue #28070, opened 2023-11-09
[Feature] APIRequestContext send multiple files in a single field with a single field name


dariusnemeiksis:
Hello, 

Use case: 
Back-end API service accepts multiple files in a single form field with a single field name.

Current behavior:
APIRequestContext in multipart supports sending multiple files with different field names. 
example: `curl -F "docpicture=@dog.gif" -F "catpicture=@cat.gif" $URL`
```
multipart: {
    file: stream1,
    oneFile: {
        name: "file1.jpg",
        mimeType: 'image/jpg',
        buffer: image1,
    }
}
```

Extended behavior:
APIRequestContext in multipart supports sending multiple files in a single field with a single field name.
example: `curl -F "pictures=@dog.gif,cat.gif" $URL`
```
multipart: {
    files: [stream1, stream2],
    multiFiles: [{
        name: "file1.jpg",
        mimeType: 'image/jpg',
        buffer: image1,
    },
    {
        name: "file2.jpg",
        mimeType: 'image/jpg',
        buffer: image2,
    }]
}
```

With this additional APIRequestContext extension, people can use Playwright directly without adding additional packages.

NikkTod:
Hello Playwright team,

we have an email API, that accepts multiple files, but they have to be send with same field name called 'attachment'.

We are not able to test the functionality as the APIRequestContext expects different field name for each attachment.

Would very much appreciate if you can add this feature to the APIRequestContext.  

cc: @dgozman, @pavelfeldman @mxschmitt

NikkTod:
@dgozman thanks for adding this one to the release, we are migrating our API tests to Playwright and need it desperately.

PR #30489, opened 2024-04-23
feat(fetch): support FormData as multipart parameter

yury-s:
Reference https://github.com/microsoft/playwright/issues/28070

### DIFF
--- a/docs/src/api/class-formdata.md
+++ b/docs/src/api/class-formdata.md
@@ -22,7 +22,7 @@ Appends a new value onto an existing key inside a FormData object, or adds the k
 does not already exist. File values can be passed either as `Path` or as `FilePayload`.
 Multiple fields with the same name can be added.
 
-The difference between set() and append() is that if the specified key already exists,
+The difference between [`method: FormData.set`] and [`method: FormData.append`] is that if the specified key already exists,
 set() will overwrite all existing values with the new one, whereas append() will append
 the new value onto the end of the existing set of values.
 
--- a/docs/src/api/class-formdata.md
+++ b/docs/src/api/class-formdata.md
@@ -14,6 +14,77 @@ FormData form = FormData.create()
 page.request().post("http://localhost/submit", RequestOptions.create().setForm(form));
 ```
 
+## method: FormData.append
+* since: v1.43
+- returns: <[FormData]>
+
+Appends a new value onto an existing key inside a FormData object, or adds the key if it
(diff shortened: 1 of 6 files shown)


docs/src/api/params.md
```diff
@@ -405,7 +405,7 @@ An instance of [FormData] can be created via [`method: APIRequestContext.createF
 
 ## js-python-fetch-option-multipart
 * langs: js, python
-- `multipart` <[Object]<[string], [string]|[float]|[boolean]|[ReadStream]|[Object]>>
+- `multipart` <[FormData]|[Object]<[string], [string]|[float]|[boolean]|[ReadStream]|[Object]>>
```
dgozman, line 408: - Is `FormData` available in python?
- We will need a custom link on `playwright.dev` leading to MDN.

docs/src/api/class-formdata.md
```diff
@@ -14,6 +14,77 @@ FormData form = FormData.create()
...
 
+## method: FormData.append
+* since: v1.43
+- returns: <[FormData]>
+
+Appends a new value onto an existing key inside a FormData object, or adds the key if it
+does not already exist. File values can be passed either as `Path` or as `FilePayload`.
+Multiple fields with the same name can be added.
+
+The difference between set() and append() is that if the specified key already exists,
```
dgozman, line 25:
```suggestion
The difference between [`method: FormData.set`] and [`method: FormData.append`] is that if the specified key already exists,
```

docs/src/api/class-formdata.md
```diff
@@ -14,6 +14,77 @@ FormData form = FormData.create()
...
+set() will overwrite all existing values with the new one, whereas append() will append
```
dgozman, line 26:
```suggestion
setting will overwrite all existing values with the new one, whereas appending will add
```

dgozman (approved):

docs/src/api/params.md
yury-s: > Is FormData available in python?

No, only csharp and java.

docs/src/api/class-formdata.md
yury-s: These links will not work in js and python.

docs/src/api/params.md
yury-s: https://github.com/microsoft/playwright.dev/pull/1360 adds FormData link

dgozman (approved):

Issue #30504, opened 2024-04-24
[Bug]: Worker-scoped fixture execution terminates before timeout


NoamGaash:
### Version

1.43.1

### Steps to reproduce

1. clone [my repo](https://github.com/NoamGaash/never-ending-fixture) - `git clone git@github.com:NoamGaash/never-ending-fixture.git`
2. npm install
3. npx playwright test -g "my test"

### Expected behavior

1. I expect to see both `"Fixture teardown starts"` and `"Fixture teardown ends"` log messages
2. I expect the string `"will never be thrown"` to be thrown
3. I expect the teardown to take at least 40 seconds

### Actual behavior

1. The "Fixture teardown starts" text appears, but "Fixture teardown ends" is missing
2. The test takes 30.2s
3. No exception is thrown

### Additional context

The documentation states that:

> for slow fixtures, especially [worker-scoped](https://playwright.dev/docs/test-fixtures#worker-scoped-fixtures) ones, it is convenient to have a separate timeout

That's why I believe timeouts are expected to work for worker-scoped fixtures.

### Environment

```shell
System:
    OS: Linux 6.5 Ubuntu 23.10 23.10 (Mantic Minotaur)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
    Memory: 14.82 GB / 31.01 GB
    Container: Yes
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.12.1/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.12.1/bin/npm
  IDEs:
    VSCode: 1.88.1 - /usr/bin/code
  Languages:
    Bash: 5.2.15 - /usr/bin/bash
  npmPackages:
    @playwright/test: ^1.43.1 => 1.43.1
```

pavelfeldman:
Investigation note:
- worker is stopped
- graceful worker termination is scheduled concurrently with the worker teardown

NoamGaash:
@dgozman
Thank you for addressing this issue.
Few words regarding our use case:
In every test, we are uploading test related resources to our servers. We don't want this long operation to block the worker from proceeding the run with other tests, that's why we use fixtures to wait for this operation to be completed during worker's teardown.
[...]

PR #30512, opened 2024-04-24
fix: gracefully close unused workers

NoamGaash:
This patch suggest a fixing issue #30504 by gracefully terminating the worker before exiting the process


dgozman (commented):
Thank you for the PR! Unfortunately, we cannot remove the `process.exit()` call after some timeout, because graceful termination may misbehave and leave zombie processes.

There should be a different fix for this, most likely the one that waits for graceful termination in the regular worker shutdown from the dispatcher. However, I'll have to experiment with it first to figure out the right fix. We'll look into this for the v1.45 release.

NoamGaash:
Investigation notes:

- the dispatcher calls `worker.stop()` because `_isWorkerRedundant(worker)` evaluates as `true`.
- the worker is considered to be redundant because `isWorkerRedundant` sees that the `_queuedOrRunningHashCount` is zero - there are no queued or running jobs, as the worker teardown doesn't count as a job

dgozman:
@NoamGaash I spent some time on the issue, and the fix is not that straightforward. Therefore, I went ahead and prepared a PR myself - https://github.com/microsoft/playwright/pull/30769. Thank you for the PR and investigation!

NoamGaash:
@dgozman Thank you so very much! It was a real blocker, and I'm so glad for your help over here. Also - it's a great learning opportunity. Do you mind if I'll ask a little to get a better understanding? I'm sure I'll have future opportunities to contribute, and it's inspiring to see the clean code and architecture.

As you said yourself, graceful termination may leave zombie processes, therefore I thought the right approach to solve this issue would be investigating why it's called in the first place. The root cause for calling the worker termination was the `_isWorkerRedundant` method that iterates all worker slots, and sees whether any of them is occupied with the task assigned to the current worker.

It seems like `slot.worker.didSendStop()` of the `_isWorkerRedundant` method evaluates to be true, so I thought the real problem lays somewhere inside the test runner architecture and there's some stop command being sent when the last test is executed. That's why I'm surprised to see your solution includes conditioning the force exit - is it a temporary solution, or that's the "right thing to do"?

Thanks again, both for responding and solving this issue so quickly and for Playwright as whole :smiley_cat:

dgozman:
@NoamGaash There is a difference between normal operation worker stop, and the case where something went wrong. So my change assumes that during normal worker stop, triggered by `_isWorkerRedundant` or `dispatcher.stop()`, worker teardown behaves. However, if something went wrong, upon Ctrl+C we'll disconnect and worker will force exit.

NoamGaash:
I see. Thanks for this clarification!

Issue #30091, opened 2024-03-25
[Regression]: Reports are not merged.


sleepmac:
### Last Good Version

1.38

### First Bad Version

1.42

### Steps to reproduce

Hi. After upgrading from version 1.38 to 1.42 after running pipeline the reports are not merged.
I have three projects desktop, iphone, lg. Each test is run for three projects, tests are run in parallel. 
Each project is divided into shards.

### Expected behavior

After passing the pipeline, the report shows only tests for one project lg. I expect three projects in the report

### Actual behavior

Report with one project.

### Additional context

gitlab-ci
```yaml
variables: 
  STAGE:
    value: "test"
  BASEURL: "https://${STAGE}.example.com/"
  PLAYWRIGHT_BLOB_REPORT: "blob-report"
  PLAYWRIGHT_TESTS_RESULTS: "test-results"
  PWTEST_BLOB_REPORT_NAME: "${PW_PROJECT}"
  PLAYWRIGHT_JUNIT_OUTPUT_NAME: "results.xml"
  PLAYWRIGHT_HTML_REPORT: "playwright-report"
  PROJECT:
    value: "all"
    options:
      - "all"
      - "desktop"
      - "mobile"
  GREP:
    value: ""

stages:
  - tests
  - reports

.playwright_job: &playwright_job
  image: mcr.microsoft.com/playwright:v1.42.0-jammy

.tests_job: &tests_job
  <<: *playwright_job
  stage: tests
  allow_failure: true
  script:
    - |
      options=()
      if [[ -n "$GREP" ]]; then
        options+=(--grep "$GREP")
      fi
      export PWTEST_BLOB_REPORT_NAME="${PW_PROJECT}"
      env
      PWTEST_BLOB_REPORT_NAME=$PW_PROJECT node -e "console.log(process.env.PWTEST_BLOB_REPORT_NAME)"
      node -e "console.log(process.env.PWTEST_BLOB_REPORT_NAME)"
      PWTEST_BLOB_REPORT_NAME=$PW_PROJECT npx playwright test \
        --retries=1 \
        --project=$PW_PROJECT \
        --shard=$SHARD \
        --reporter=blob,junit \
        "${options[@]}"
      apt update -y && apt install tree -y
      tree $PLAYWRIGHT_BLOB_REPORT
  artifacts:
    paths:
      - $PLAYWRIGHT_BLOB_REPORT
      - $PLAYWRIGHT_TESTS_RESULTS
[...]

PR #30528, opened 2024-04-24
fix(merge): include command hash into the report name

yury-s:
Reference https://github.com/microsoft/playwright/issues/30091

### DIFF
--- a/packages/playwright/src/program.ts
+++ b/packages/playwright/src/program.ts
@@ -20,7 +20,7 @@ import type { Command } from 'playwright-core/lib/utilsBundle';
 import fs from 'fs';
 import path from 'path';
 import { Runner } from './runner/runner';
-import { stopProfiling, startProfiling, gracefullyProcessExitDoNotHang } from 'playwright-core/lib/utils';
+import { stopProfiling, startProfiling, gracefullyProcessExitDoNotHang, calculateSha1 } from 'playwright-core/lib/utils';
 import { serializeError } from './util';
 import { showHTMLReport } from './reporters/html';
 import { createMergedReport } from './reporters/merge';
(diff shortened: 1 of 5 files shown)


packages/playwright/src/program.ts
```diff
@@ -201,6 +202,19 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
   gracefullyProcessExitDoNotHang(exitCode);
 }
 
+function computeCommandHash(args: string[], opts: { [key: string]: any }) {
+  const command = {
+    args,
+    opts: {} as any
+  };
+  // Omit shard from the hash, shard number is added explicitely at the end of the blob report file name.
+  const keys = Object.keys(opts).filter(k => k !== 'shard');
```
pavelfeldman, line 211: nit: i looked at the options and we don't seem to care about most of them. So maybe only use those definding the scope of the run? `[args], { grep, grepInvert, project, shard }`

pavelfeldman (commented):
How about supporting grep via env while we are there?

packages/playwright/src/runner/reporters.ts
```diff
@@ -90,9 +91,27 @@ function reporterOptions(config: FullConfigInternal, mode: 'list' | 'test' | 'me
...
 
+function computeCommandHash(config: FullConfigInternal) {
+  let parts = [];
+  // Include project names for readability.
+  if (config.cliProjectFilter)
+    parts.push(...config.cliProjectFilter);
+  const command = {} as any;
+  if (config.cliArgs.length)
+    command.cliArgs = config.cliArgs;
+  if (config.cliGrep)
```
pavelfeldman, line 106: This does not address the passed-via-env scenario.

pavelfeldman (approved):

1337doctor:
If your gh actions for saving or restoring shard from cache (artifacts) and your pipeline stopped working - maybe this post prevents some headache.

If you receive this error " Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved." while saving or restoring a cache / artifact and your used a glob with a specific filename - this commit might be the reason.

Just refer to mybasedir/blob-report instead of mybasedir/blob-report/report-1.zip (or something similiar).

Hope someone finds this helpful.

Issue #13517, opened 2022-04-13
[Feature] Aria methods and assertions


kevin940726:
Originally posted in https://github.com/microsoft/playwright/pull/12999#issuecomment-1092536668.

> Now that we have all these accessible attributes helpers, is it worth making them available under locator and assertions as well?
>
> For instance, we want to know whether an element is expanded or not:
>
> ```js
> const button = page.locator('role=button[name=Expand me]');
> const isExpanded = await button.expanded(); // `true` or `false`
> await expect(button).toBeExpanded(); // Error if the button is not expanded
> ```
>
> I think it's only natural to make them available both via querying and accessing.
>
> Currently, we only have `page.accessibility`, but that uses browsers' API and it's platform-specific, different from the JS implementation injected into the browsers.

It'll be great to add support for computing aria attributes on Locator and ElementHandle to make them more feature-complete.

Proposed methods on Locator (and ElementHandle):

- [ ] `locator.ariaRole(): string`: Get the aria role of the element.
- [ ] `locator.ariaName(): string`: Get the aria name of the element.
- [x] `locator.isChecked(): boolean`: Already available.
- [ ] `locator.isSelected(): boolean`: Returns whether the option is selected.
- [ ] `locator.isExpanded(): boolean`: Returns whether the element is expanded.
- [ ] `locator.isPressed(): boolean`: Returns whether the button is being pressed.
- [x] `locator.isDisabled(): boolean`: Already available.
- [ ] `locator.level(): number`: Get the level of the heading element.

And for assertions:

- [ ] `expect(locator).toHaveRole(role: string)`: Test whether the element has the aria role.
- [ ] `expect(locator).toHaveName(name: string | RegExp)`: Test whether the element has the aria name.
- [x] `expect(locator).toBeChecked()`: Already available.
- [ ] `expect(locator).toBeSelected()`: Expect the option to be selected.
- [ ] `expect(locator).toBeExpanded()`: Expect the element to be expanded.
- [ ] `expect(locator).toBePressed()`: Expect the button to be pressed.
- [x] `expect(locator).toBeDisabled()`: Already available.
- [ ] `expect(locator).toHaveLevel(level: number)`: Expect the heading element to have the level.

c.c. @dgozman

frontsideair:
It would be nice to have the accessible name selector as an option to filter, like `hasText`. It would look like this:
```ts
await page.locator('role=button', { hasName: /click me/i }).click();
// translates to
[...]

PR #30555, opened 2024-04-25
feat: expect(locator).toHaveRole(role)

### LINKED ISSUE #18332: [Feature] locator for getting an element's description

dgozman:
References #13517. Fixes #18332.


docs/src/api/class-locatorassertions.md
```diff
@@ -1629,6 +1646,53 @@ Property value.
...
 
+
+## async method: LocatorAssertions.toHaveRole
+* since: v1.44
+* langs:
+  - alias-java: hasRole
+
+Ensures the [Locator] points to an element with a given [ARIA role](https://www.w3.org/TR/wai-aria-1.2/#roles).
+
+Note that role is matched as a string, disregarding the ARIA role hierarchy. For example, asserting  a superclass role `"checkbox"` on an element with a subclass role `"switch"` will fail.
```
yury-s, line 1657:
```suggestion
Note that the role is matched as a string, disregarding the ARIA role hierarchy. For example, asserting  a superclass role `"checkbox"` on an element with a subclass role `"switch"` will fail.
```

yury-s (approved):

Issue #30577, opened 2024-04-27
[Bug]: Parallel test process execution causing race condition and empty babel transpiled file read from disk cache


sunilsurana:
### Version

1.42.1

### Steps to reproduce

It can be reproduced on Codespace or Azure VM's. Created this repo with steps for reproing issue. https://github.com/sunilsurana/reprofilesync

This repo mimics the behaviour of compilationcache.ts logic if it were to execute in parallel processes
Note that the issue may or may not repro on local laptop depending on disk and OS. It repros on codespace and azure vm

### Expected behaviour

The tests should work deterministically fine when multiple processes trigger tests at same time

### Actual behaviour

The test throws error because it could not find an import. It comes null because of empty file read when tests invoked thru multiple processes in parallel

### Additional context

We distribute our tests across multiple workers and processes. So on a single machine if multiple process start test at same time the disk cached code reading part reads empty file because another process just began writing to it.
This happens because the fs.writeFileSync create 3 OS level calls which we can see using strace

open("file.txt", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 9
pwrite(9, "content\n", 8, 0)            = 8
close(9)                                = 0

The first open sys call truncates file if it exist and creates a blank file. Now at same time if reader comes from another process the reader will get file exists and when reading file it get blank file.

We were able to verify this by putting logs in playwright where the read and write were happening at exact millisecond precision and causing test failures.

### Environment

```shell
System: Windows Azure vm 8 core 32GB RAM
OS image windows 2022 server
Node version 18LTS
```

sunilsurana:
Create a PR to fix it. We were able to patch playwright with this PR changes and verify it worked fine then.

https://github.com/microsoft/playwright/pull/30578

sunilsurana:
Bit more detailed explanation.

The current main code block look like this

```
try {
 read codepath
} catch {
  do nothing
}
codecontent = transpilefileusing babel
write codecontent to codepath
```
Not if 1st process makes call to write codeconent and 2nd process make call to read codecontent at same time the 2nd process gets and empty file becuase the writefilesync first creates empty file and them writes content and then closes filedescriptor

To fix this the code is changed to 
```
if marker file present which means the write by another process was fully completed
{
   try {
       read codepath
      } catch {
        do nothing
     }
}
codecontent = transpilefileusing babel
write codecontent to codepath using wx mode which is write exclusive
write marker file
if error.errorcode = file already exist do nothing
[...]

PR #30578, opened 2024-04-27
fix : Synchronize read and writes to avoid race condition reading babel transpiled files from local disk cache

sunilsurana:
1. Add a maker file to check if the previous write was fully completed.
2. In writefilesync section change the write mode to wx which uses exclusive lock 
3. If two writes happen same time and we get EEXIST error then ignore that 
FIX for #30577


dgozman:
Closing for now, see the linked issue.

Issue #6072, opened 2021-04-05
How to preform touch events?


ccpu:
Apparently  playwright only supports tab on touchscreen, but how other touch events such as touchmove, touchend and touchcancel can be simulated?

soulchainer:
@ccpu I think (not tried) you can do it with `elementHandle.dispatchEvent('eventType')`. Check the [`elementHandle.dispatchEvent` in docs](https://playwright.dev/docs/api/class-elementhandle#elementhandledispatcheventtype-eventinit).

ccpu:
@soulchainer indeed dispatchEvent is the solution, i have forgotten about that, thank you.

 [Here is the implementation](https://github.com/ccpu/storybook-addon-playwright/blob/master/src/page-extra/utils/dispatch-touch-event.ts) if someone needed.

PR #30587, opened 2024-04-29
feat: add the `page.touchscreen.swipe` method

### LINKED ISSUE #2903: [Feature] Support for touch events/gestures
**Note from maintainers**

At this time, we don’t plan to add native support for touch event emulation in Playwright. Clicks and other mouse actions already generate the corresponding `PointerEvents`, which is the modern and widely supported standard.

For projects that depend on legacy `TouchEvents`, we recommend using the approaches described in our [guide]( https://playwright.dev/docs/touch-events). These cover some scenarios where touch events are still required.

---

Web apps running on touch devices (e.g, mobile and Surface) rely on touch gestures like swiping and pinching to zoom. While `page.dispatchEvent` can manufacture synthetic touch points, having a simpler API would improve the dev experience.

Scenarios
1. Two-finger pinch to zoom in/out on a particular element (e.g., canvas in [Azure ML designer](https://docs.microsoft.com/en-us/azure/machine-learning/concept-designer))
2. Verify that two-finger left/right swipe is not leading to go back/forward

PupilTong:
https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchTouchEvent

Implement it for chromium only since it could  using dispatchTouchEvent to invoke scroll.

Tried firefox but it doesn't work.

 #2903


PupilTong:
> @PupilTong please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Bytedance"

PupilTong:
https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-synthesizeScrollGesture
This CDP method cannot scroll a container on xvfb headless chromium with `gestureSourceType:'touch'`. It might only works on a device with a screen. 
Also, this 'dispatchtouch' implementation is tried in Firefox. But it cannot scroll a container neither.

dgozman (commented):
Thank you for the PR! Unfortunately, we cannot make this a Chromium-only feature, it has to work in all three browser engines.

PupilTong:
> Thank you for the PR! Unfortunately, we cannot make this a Chromium-only feature, it has to work in all three browser engines.

How about expose the [Input.dispatchTouchEvent](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchTouchEvent) ?
Like add a `page.touchscreen.touch()` therefore we could create the `swipe` wrapper by ourselves.

The firefox also implemented the this CDP method.

Just like the `page.mouse.wheel()`, it is supported by chromium and firefox.

dgozman:
> How about expose the [Input.dispatchTouchEvent](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchTouchEvent) ? For example, add a `page.touchscreen.touch()` therefore we could create the `swipe` wrapper by ourselves.

I am not really sure what you mean by "exposing Input.dispatchTouchEvent". Could you please elaborate?

> Just like the `page.mouse.wheel()`, it is supported by chromium and firefox.

Generally, Playwright APIs should be supported by chromium, firefox and webkit.

PupilTong:
> > How about expose the [Input.dispatchTouchEvent](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchTouchEvent) ? For example, add a `page.touchscreen.touch()` therefore we could create the `swipe` wrapper by ourselves.
> 
> I am not really sure what you mean by "exposing Input.dispatchTouchEvent". Could you please elaborate?
> 
> > Just like the `page.mouse.wheel()`, it is supported by chromium and firefox.
> 
> Generally, Playwright APIs should be supported by chromium, firefox and webkit.

@dgozman 

* expose the CDP method `'Input.dispatchTouchEvent'`

Both the firefox and the chromium support this CDP method. I can submit another PR,  create a Playwright method like `page.touchscreen.touch()` as a wrapper of the CDP method. Therefore developers are able to create their own `swipe()` method based on it.

* About the compatibility

Currently Playwright already have a method ,`page.mouse.wheel()`, which cannot be called for webkit running env. I think this means if a new method only support chromium and firefox may be accepted. 
Considering the Safari team's attitude about the touch-related feature on desktop safari (They even do not provide the `TouchEvent` on the desktop safari.), I guess add a touchscreen method for playwright without safari support is reasonable.

dgozman:
@PupilTong Thank you for the answers!

If you believe that developers can implement `swipe()` based on exposed `dispatchTouchEvent` in Firefox, then you should be able to do the same. Therefore, I'd prefer to stick with `touchscreen.swipe()` API as in this PR.

As for compatibility, our goal is cross-browser support where possible. For example, we support `touchscreen.tap()` on every browser where `hasTouch` option was set. In contrast, `mouse.wheel()` is not supported in mobile webkit because there is no real mouse on devices with mobile webkit browsers, however `mouse.wheel()` is supported in desktop webkit.

Overall, the approach you take in this PR is a good one. Unfortunately, this feature most likely requires some work in the browser engines themselves, so it is not that straightforward to implement.

As a workaround for now, we recommend using `locator.dispatchEvent()` method and issue some specific events like `touchstart`/`touchmove`/`touchend`. This is not the real swipe, but it is something useful today. See also #6072.

PupilTong:
> @PupilTong Thank you for the answers!
> 
> If you believe that developers can implement `swipe()` based on exposed `dispatchTouchEvent` in Firefox, then you should be able to do the same. Therefore, I'd prefer to stick with `touchscreen.swipe()` API as in this PR.
> 
> As for compatibility, our goal is cross-browser support where possible. For example, we support `touchscreen.tap()` on every browser where `hasTouch` option was set. In contrast, `mouse.wheel()` is not supported in mobile webkit because there is no real mouse on devices with mobile webkit browsers, however `mouse.wheel()` is supported in desktop webkit.
> 
> Overall, the approach you take in this PR is a good one. Unfortunately, this feature most likely requires some work in the browser engines themselves, so it is not that straightforward to implement.
> 
> As a workaround for now, we recommend using `locator.dispatchEvent()` method and issue some specific events like `touchstart`/`touchmove`/`touchend`. This is not the real swipe, but it is something useful today. See also #6072.

Thanks for your answer!

Let me explain why the `dispatchEvent` is not enough to use.
The `locator.dispatchEvent()` cannot meet our requirements. It requires to "locate a dom". But for e2e tests, typically we have to send events based on the x&y position to test the gestures implementation.

I'll spend some time for Firefox about this.

For Safari, it seems they requires developers to use their emulator to test touch gestures. https://developer.apple.com/documentation/safari-developer-tools/inspecting-ios
I don't know if it could be implemented.

pavelfeldman:
Let me archive this for now, please feel free to re-post and link to this PR.

PR #30618, opened 2024-05-01
feat: implement flag to fail flaky tests

### LINKED ISSUE #30457: [Feature]: Add Configuration Option to Mark Test Suite as Failed When Containing Flaky Tests
### Feature Description
Currently, Playwright supports a "retries" configuration where tests that initially fail but subsequently pass are marked as "flaky". When such flaky tests exist, the overall test suite is considered successful. I propose adding a configuration option that allows marking the entire test suite as failed if it contains any flaky tests.

### Proposed Solution
Introduce a new configuration parameter, such as `failOnFlakyTests`, which when set to `true`, will cause the entire test suite to fail if any tests are marked as flaky. This will help teams differentiate between suites with consistent test failures and those with flaky tests, promoting more robust testing practices.

### Impact
This enhancement will allow developers and QA teams to enforce stricter quality controls by treating flaky tests as critical errors, ensuring that they receive immediate attention.

Thank you for considering this feature.

### Example

```json
{
  "failOnFlakyTests": true,
  "retries": 2
}
```

### Motivation

This feature would be beneficial for maintaining high code quality in continuous integration (CI) environments. By enabling this option, teams can ensure that flaky tests are highlighted and addressed promptly, rather than allowing them to degrade the reliability of the test suite over time.

Joe-Hendley:
Implements feature requested in https://github.com/microsoft/playwright/issues/30457

The test runner treats flaky tests as failures when the flag is enabled, but still reports flaky tests as flaky in the reporting interface. It feels like something worth discussing as this behaviour makes sense to me, but looked a bit odd to @BJSS-russell-pollock  when I ran this past him.

Closes #30457.


Joe-Hendley:
@microsoft-github-policy-service agree company="BJSS"

BJSS-russell-pollock:
Talking point of the failing on flaky behaviour.
To try this out, I set the the **playwright.config** to:
```
retries: process.env.CI ? 2 : 5,
failOnFlakyTests: true,
```
And then the test:
```
test('flaky test', async () => {

  const randomFail = Math.round(Math.random());
  expect(randomFail).toBe(0)
});
```

When running the test through the CLI - the retry would pass and then mark the test as flaky:
retry
result flaky

dgozman (commented):
Thank you for the PR!

I'd like to understand the feature a little better. Do we want to stop recognizing flaky tests at all? Are we after the test runner exit code? Or some changes in the terminal/html report?

Looking at the issue, my initial understanding is "introduce --fail-on-flaky cli option that will set exitCode=1 when at least one test was flaky". However, this PR introduces many more changes, so let's align on the intended behavior first.

Joe-Hendley:
Hi there,

The intended behaviour from this PR, and my understanding of the issue matches yours - any changes in the code extra to that reflect my unfamiliarity with the codebase.

I think Russell's comment / the discussion around behaviour reflects the user / manual tester perspective where the change in exit code isn't immediately visible.

dgozman:
> The intended behaviour from this PR, and my understanding of the issue matches yours - any changes in the code extra to that reflect my unfamiliarity with the codebase.

Sounds great! In this case, I think we should plumb the new CLI option similar to `--pass-with-no-tests`. Search for `passWithNoTests` across the codebase for the reference. And [this is a place](https://github.com/microsoft/playwright/blob/main/packages/playwright/src/runner/failureTracker.ts#L50) that determines whether everything was successful or something has failed, and it needs to be tweaked.

@BJSS-russell-pollock I think this should be clear since reporter shows the "Flaky" count as non-zero. I'd assume anyone using this option will know that flakiness means "failed test run", because the behavior was opt in. Let me know what you think.

BJSS-russell-pollock:
> @BJSS-russell-pollock I think this should be clear since reporter shows the "Flaky" count as non-zero. I'd assume anyone using this option will know that flakiness means "failed test run", because the behavior was opt in. Let me know what you think.

Absolutely makes sense.

Joe-Hendley:
> Sounds great! In this case, I think we should plumb the new CLI option similar to `--pass-with-no-tests`. Search for `passWithNoTests` across the codebase for the reference. And [this is a place](https://github.com/microsoft/playwright/blob/main/packages/playwright/src/runner/failureTracker.ts#L50) that determines whether everything was successful or something has failed, and it needs to be tweaked.

Excellent - I'll have a look and update it accordingly.

Joe-Hendley:
Okay, reworked as suggested - much less additional code!

Joe-Hendley:
@dgozman This should be ok for re-review when you're ready - the failed tests in previous pipeline runs look like noise to me, but happy to dig if you think it's deeper

dgozman (approved):
Looks great, merging in. Thank you for the PR!

PR #30619, opened 2024-05-01
chore: make NodeSnapshot type recursive and more

ruifigueira:
Also, deviceDescriptors are now imported with ESM import instead of require()

### DIFF
--- a/packages/playwright-core/src/server/deviceDescriptors.ts
+++ b/packages/playwright-core/src/server/deviceDescriptors.ts
@@ -15,7 +15,8 @@
  * limitations under the License.
  */
 
-/**
- * @type {import('./types').Devices}
- */
-module.exports = require("./deviceDescriptorsSource.json")
+import type { Devices } from './types';
+
+import deviceDescriptorsSource from './deviceDescriptorsSource.json';
+
+export const deviceDescriptors = deviceDescriptorsSource as Devices;
--- a/packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
+++ b/packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
@@ -26,7 +26,7 @@ import { Dispatcher } from './dispatcher';
 import { yazl, yauzl } from '../../zipBundle';
 import { ZipFile } from '../../utils/zipFile';
 import type * as har from '@trace/har';
-import type { HeadersArray, Devices } from '../types';
+import type { HeadersArray } from '../types';
 import { JsonPipeDispatcher } from '../dispatchers/jsonPipeDispatcher';
 import { WebSocketTransport } from '../transport';
 import { SocksInterceptor } from '../socksInterceptor';
@@ -40,6 +40,7 @@ import type http from 'http';
 import type { Playwright } from '../playwright';
 import { SdkObject } from '../../server/instrumentation';
 import { serializeClientSideCallMetadata } from '../../utils';
+import { deviceDescriptors as descriptors }  from '../deviceDescriptors';
 
 export class LocalUtilsDispatcher extends Dispatcher<{ guid: string }, channels.LocalUtilsChannel, RootDispatcher> implements channels.LocalUtilsChannel {
   _type_LocalUtils: boolean;
@@ -53,7 +54,6 @@ export class LocalUtilsDispatcher extends Dispatcher<{ guid: string }, channels.
 
   constructor(scope: RootDispatcher, playwright: Playwright) {
     const localUtils = new SdkObject(playwright, 'localUtils', 'localUtils');
-    const descriptors = require('../deviceDescriptors') as Devices;
     const deviceDescriptors = Object.entries(descriptors)
         .map(([name, descriptor]) => ({ name, descriptor }));
     super(scope, localUtils, 'LocalUtils', {
--- a/packages/trace-viewer/src/snapshotRenderer.ts
+++ b/packages/trace-viewer/src/snapshotRenderer.ts
@@ -14,7 +14,15 @@
  * limitations under the License.
  */
 
-import type { FrameSnapshot, NodeSnapshot, RenderedFrameSnapshot, ResourceSnapshot } from '@trace/snapshot';
+import type { FrameSnapshot, NodeNameAttributesChildNodesSnapshot, NodeSnapshot, RenderedFrameSnapshot, ResourceSnapshot, SubtreeReferenceSnapshot } from '@trace/snapshot';
+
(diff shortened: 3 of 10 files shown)


packages/playwright-core/src/server/deviceDescriptors.ts
```diff
@@ -15,7 +15,8 @@
...
 
-/**
- * @type {import('./types').Devices}
- */
-module.exports = require("./deviceDescriptorsSource.json")
+import type { Devices } from './types';
+
+import deviceDescriptorsSource from './deviceDescriptorsSource.json';
+
+export const deviceDescriptors = deviceDescriptorsSource as Devices;
```
dgozman, line 22: I am not against this change, but I wonder what's the reason?

packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
```diff
@@ -40,6 +40,7 @@ import type http from 'http';
 import type { Playwright } from '../playwright';
 import { SdkObject } from '../../server/instrumentation';
 import { serializeClientSideCallMetadata } from '../../utils';
+import { deviceDescriptors as descriptors }  from '../deviceDescriptors';
```
dgozman, line 43: I guess we can just import from the json file directly and remove `deviceDescriptors.ts` entirely?

packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -14,7 +14,15 @@
...
  */
 
-import type { FrameSnapshot, NodeSnapshot, RenderedFrameSnapshot, ResourceSnapshot } from '@trace/snapshot';
+import type { FrameSnapshot, NodeNameAttributesChildNodesSnapshot, NodeSnapshot, RenderedFrameSnapshot, ResourceSnapshot, SubtreeReferenceSnapshot } from '@trace/snapshot';
+
+function isNodeNameAttributesChildNodesSnapshot(n: unknown): n is NodeNameAttributesChildNodesSnapshot {
+  return Array.isArray(n) && typeof n[0] === 'string';
+}
+
+function isSubtreeReferenceSnapshot(n: unknown): n is SubtreeReferenceSnapshot {
```
dgozman, line 23:
```suggestion
function isSubtreeReferenceSnapshot(n: NodeSnapshot): n is SubtreeReferenceSnapshot {
```

packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -14,7 +14,15 @@
  * limitations under the License.
  */
 
-import type { FrameSnapshot, NodeSnapshot, RenderedFrameSnapshot, ResourceSnapshot } from '@trace/snapshot';
+import type { FrameSnapshot, NodeNameAttributesChildNodesSnapshot, NodeSnapshot, RenderedFrameSnapshot, ResourceSnapshot, SubtreeReferenceSnapshot } from '@trace/snapshot';
+
+function isNodeNameAttributesChildNodesSnapshot(n: unknown): n is NodeNameAttributesChildNodesSnapshot {
```
dgozman, line 19:
```suggestion
function isNodeNameAttributesChildNodesSnapshot(n: NodeSnapshot): n is NodeNameAttributesChildNodesSnapshot {
```

packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -14,7 +14,15 @@
...
+  return Array.isArray(n) && typeof n[0] === 'string';
```
dgozman, line 20: This check could also yield `NodeNameSnapshot` without attributes or children.

packages/trace/src/snapshot.ts
```diff
@@ -18,18 +18,21 @@ import type { Entry as HAREntry } from './har';
...
-  [ string, { [attr: string]: string }, ...any ];
+// Text node.
+export type TextNodeSnapshot = string;
+// Subtree reference, "x snapshots ago, node #y". Could point to a text node.
+// Only nodes that are not references are counted, starting from zero, using post-order traversal.
+export type SubtreeReferenceSnapshot = [ [number, number] ];
+// Just node name.
+export type NodeNameSnapshot = [ string ];
+// Node name, attributes, child nodes.
+export type NodeNameAttributesChildNodesSnapshot = [ string, Record<string, string>, ...NodeSnapshot[] ];
```
dgozman, line 29: IIRC, this was not compiling before, but I guess it is now. That's nice.

packages/trace/src/snapshot.ts
```diff
@@ -18,18 +18,21 @@ import type { Entry as HAREntry } from './har';
...
-  // Node name, attributes, child nodes.
-  // Unfortunately, we cannot make this type definition recursive, therefore "any".
-  [ string, { [attr: string]: string }, ...any ];
+// Text node.
+export type TextNodeSnapshot = string;
+// Subtree reference, "x snapshots ago, node #y". Could point to a text node.
+// Only nodes that are not references are counted, starting from zero, using post-order traversal.
+export type SubtreeReferenceSnapshot = [ [number, number] ];
+// Just node name.
+export type NodeNameSnapshot = [ string ];
```
dgozman, line 27: I'd merge this and the next type into a single exported alias, because there are no practical uses to differentiate between the two.

dgozman (commented):
Thank you for the PR! According to the guidelines https://github.com/microsoft/playwright/blob/main/CONTRIBUTING.md#commit-messages, you should name this `chore: make NodeSnapshot type recursive and more`.

See inline comments.

packages/playwright-core/src/server/deviceDescriptors.ts
ruifigueira: It makes imports more uniform, and besides it helps me in my https://github.com/ruifigueira/playwright-crx project, which uses vite to compile playwright to work on a chrome extension service worker, and vite doesn't like `requires` that much

packages/playwright-core/src/server/dispatchers/localUtilsDispatcher.ts
ruifigueira: In this specific file, it works, but in other files it complains because it cannot infer it's a `Devices` type. By having json content exposed in `deviceDescriptors.ts` as `Devices` we no longer have to cast it in other files.

ruifigueira: BTW, with `require('../deviceDescriptorsSource.json')`, inferred type was `any`:

image

dgozman (approved):
Looks good, merging in. Thank you for the PR!

PR #30641, opened 2024-05-03
fix: match client step and server action by explicit stepId

yury-s:
Matching bu `apiName@wallTime` fails when two actions start at the same time, e.g. two parallel api requests. Moreover, it results in trace actions that have parent set to themselves, which in turn causes infinite loop in the trace viewer. To avoid this problems we write stepId explicitly to the library trace and use those step ids to find corresponding test runner steps.

The stepId is passed via zone in case of expect, because the protocol step is quite deep in the call chain after or explicitly in case of API call steps created by the test runner instrumentation.

### DIFF
--- a/packages/playwright-core/src/client/channelOwner.ts
+++ b/packages/playwright-core/src/client/channelOwner.ts
@@ -143,11 +143,15 @@ export abstract class ChannelOwner<T extends channels.Channel = channels.Channel
           if (validator) {
             return async (params: any) => {
               return await this._wrapApiCall(async apiZone => {
-                const { apiName, frames, csi, callCookie, wallTime } = apiZone.reported ? { apiName: undefined, csi: undefined, callCookie: undefined, frames: [], wallTime: undefined } : apiZone;
+                const { apiName, frames, csi, callCookie, wallTime, stepId } = apiZone.reported ? { apiName: undefined, csi: undefined, callCookie: undefined, frames: [], wallTime: undefined, stepId: undefined } : apiZone;
                 apiZone.reported = true;
-                if (csi && apiName)
-                  csi.onApiCallBegin(apiName, params, frames, wallTime, callCookie);
-                return await this._connection.sendMessageToServer(this, prop, validator(params, '', { tChannelImpl: tChannelImplToWire, binary: this._connection.rawBuffers() ? 'buffer' : 'toBase64' }), apiName, frames, wallTime);
+                let currentStepId = stepId;
+                if (csi && apiName) {
+                  const out: { stepId?: string } = {};
+                  csi.onApiCallBegin(apiName, params, frames, wallTime, callCookie, out);
+                  currentStepId = out.stepId;
+                }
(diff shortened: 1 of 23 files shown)


packages/playwright-core/src/server/trace/recorder/tracing.ts
```diff
@@ -45,7 +45,7 @@ import type { ConsoleMessage } from '../../console';
 import { Dispatcher } from '../../dispatchers/dispatcher';
 import { serializeError } from '../../errors';
 
-const version: trace.VERSION = 6;
+const version: trace.VERSION = 7;
```
dgozman, line 48: There is no need to bump the version unless you have a meaningful `modernize_6_to_7` method. We can auto-detect by the presence of `stepId` in one of the actions.

packages/playwright/src/index.ts
```diff
@@ -258,6 +258,7 @@ const playwrightFixtures: Fixtures<TestFixtures, WorkerFixtures> = ({
           apiName,
           params,
           wallTime,
+          callId,
```
dgozman, line 261: I thought you would instead return `stepId` from here to embed into the action, to unify it with the expect story.

packages/trace-viewer/src/ui/modelUtil.ts
```diff
@@ -169,9 +169,20 @@ function mergeActionsAndUpdateTiming(contexts: ContextEntry[]) {
...
+  //   library context actions.
+  // - In the older versions the step id is not stored and the match is perfomed based on
+  //   action name and wallTime.
+  const matchByCallId = primaryContexts.some(c => c.actions.some(a => !!a.stepId));
+
   for (const context of primaryContexts) {
-    for (const action of context.actions)
-      map.set(`${action.apiName}@${action.wallTime}`, { ...action, context });
+    for (const action of context.actions) {
+      const key = matchByCallId ? (action.stepId || action.callId) : `${action.apiName}@${action.wallTime}`;
```
dgozman, line 183:
```suggestion
      const key = matchByCallId ? action.stepId : `${action.apiName}@${action.wallTime}`;
```

packages/playwright-core/src/client/channelOwner.ts
```diff
@@ -176,6 +180,7 @@ export abstract class ChannelOwner<T extends channels.Channel = channels.Channel
     // Enclosing zone could have provided the apiName and wallTime.
     const expectZone = zones.zoneData<ExpectZone>('expectZone');
     const wallTime = expectZone ? expectZone.wallTime : Date.now();
```
dgozman, line 182: I think we can get rid of storing `wallTime` in the step zone and using it here now.

yury-s: This will break clients with old trace viewer opening the new trace as the wallTime will be out of sync for some actions now.

pavelfeldman:
Is there a user report for this?

dgozman (approved):

PR #30644, opened 2024-05-03
devops: add Node.js 22 bots

mxschmitt:
Motivation: Node.js 22 is the new current and will be LTS in October. We expect users to already use Node.js 22.

---

Bisect results from the brotli breakage:

v21.0.0-nightly20230916 db8217b1bf /                 16-Sep-2023 08:30                   - good
v21.0.0-nightly20230917 56ecf29283 /                 17-Sep-2023 08:30                   - bad

https://github.com/nodejs/node/compare/db8217b1bf...56ecf29283

Downstream llhttp range: https://github.com/nodejs/llhttp/compare/v8.1.1...v9.1.2

### DIFF
--- a/.github/workflows/tests_secondary.yml
+++ b/.github/workflows/tests_secondary.yml
@@ -95,6 +95,8 @@ jobs:
           node_version: 16
         - os: ubuntu-latest
           node_version: 20
+        - os: ubuntu-latno est
+          node_version: 22
     timeout-minutes: 30
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/tests_secondary.yml
```diff
@@ -95,6 +95,8 @@ jobs:
           node_version: 16
         - os: ubuntu-latest
           node_version: 20
+        - os: ubuntu-latno est
```
mxschmitt, line 98:
```suggestion
        - os: ubuntu-latest
```

dgozman (approved):

PR #30658, opened 2024-05-04
docs(ct): multi framework typical example

sand4rt:


dgozman (commented):
It looks like the rest of this doc is react-specific, so I am not really sure this is going to help much. Perhaps we should merge "Example" into "How to get started", and then copy "How to get started" + "Test stories" + "Hooks" for other frameworks?

sand4rt:
@pavelfeldman i'll close this until i resolved the feedback of @dgozman

PR #30697, opened 2024-05-07
chore(trace): add context create event for test runner

yury-s:
Adding metadata event to the test.trace to simplify time computation logic.


pavelfeldman (approved):
I would store and use monotonicTime (not offset) for simplicity.

dgozman (approved):

PR #30703, opened 2024-05-08
feat(webkit): roll to r2006

playwrightmachine:


mxschmitt:
We can't merge this due to https://github.com/microsoft/playwright-browsers/issues/1028 - we should either file a bug upstream or update test expectations. Curious which exact cookies are causing this.

cc @dpino.

PR #30720, opened 2024-05-08
fix(trace-viewer): make call ids unique across trace files

yury-s:


packages/trace-viewer/src/ui/modelUtil.ts
```diff
@@ -160,6 +160,56 @@ function indexModel(context: ContextEntry) {
...
+    // Align action times across runner and library contexts within each trace file.
+    const map = mergeActionsAndUpdateTimingSameTrace(contexts);
+    result.push(...map.values());
+  }
+  result.sort((a1, a2) => {
+    if (a2.parentId === a1.callId)
+      return -1;
+    if (a1.parentId === a2.callId)
+      return 1;
+    return a1.wallTime - a2.wallTime || a1.startTime - a2.startTime;
```
dgozman, line 192: Should we sort by `startTime` only? Maybe in the next patch?

dgozman (approved):

packages/trace-viewer/src/ui/modelUtil.ts
yury-s: Let me tackle it in one of the next patches where I change how the times are aligned.

Issue #30141, opened 2024-03-27
[Feature]: HTML report improvements 


amankagithub:
### 🚀 Feature Request

1. I want to be able to copy the annotation text with one click rather than selecting the entire text and copy .
2. Also , why search on annotation text is not enabled . It can be useful when I want to get all the tests with certain annotation.

PR with required changes : https://github.com/microsoft/playwright/pull/30094

### Example

1.
<img width="349" alt="Screenshot 2024-03-27 at 9 52 28 AM" src="https://github.com/microsoft/playwright/assets/150931175/dea0cc13-c224-495c-a2b8-1f606f1e48e1">

2.
<img width="1051" alt="Screenshot 2024-03-27 at 9 53 44 AM" src="https://github.com/microsoft/playwright/assets/150931175/25c50276-3ce0-41fe-8745-9ca4fe22d223">

### Motivation

* Easy copy makes the enhances user experience.
* We already have searching mechanism on files , labels and status . Extending it annotation would enable the user to search on whatever they need .

AdamOakman:
The searching part of this feature request would be really helpful in the project I work for (the need mentioned in the rejected feature for allowing setting up tags dynamically but being able to search by annotations would also work fine in my case - https://github.com/microsoft/playwright/issues/30286 ).

siddharth2023:
+1

pavelfeldman:
#### Why was this issue closed?

Thank you for your involvement. This issue was closed due to limited engagement (upvotes/activity), lack of recent activity, and insufficient actionability. To maintain a manageable database, we prioritize issues based on these factors.

If you have additional information not present in this issue that you think will help prioritizing it, please open a new issue and reference this one. More support or clarity on its necessity may prompt a review. Your understanding and cooperation are appreciated.

PR #30749, opened 2024-05-11
Copy annotation text in HTML report 

amankagithub:
Feature Request : https://github.com/microsoft/playwright/issues/30141

In this PR I have added changes required to make annotation text be copied in one click . 

Annotation
Screenshot 2024-05-11 at 1 12 14 PM

On hover

Screenshot 2024-05-11 at 1 12 21 PM

On copy 
Screenshot 2024-05-11 at 1 12 34 PM

### DIFF
--- a/packages/html-reporter/src/testCaseView.css
+++ b/packages/html-reporter/src/testCaseView.css
@@ -74,3 +74,13 @@
   flex-direction: row;
   flex-wrap: wrap;
 }
+
+.annotation-copy-button{
+  padding-left: 3px;
+  display: none;
+}
+
+.test-case-annotation:hover .annotation-copy-button{
+  padding-left: 3px;
+  display: inline;
+}
--- a/packages/html-reporter/src/testCaseView.tsx
+++ b/packages/html-reporter/src/testCaseView.tsx
@@ -25,6 +25,7 @@ import './testCaseView.css';
 import { TestResultView } from './testResultView';
 import { hashStringToInt } from './labelUtils';
 import { msToString } from './uiUtils';
+import { CopyToClipboard } from './copyToClipboard';
 
 export const TestCaseView: React.FC<{
   projectNames: string[],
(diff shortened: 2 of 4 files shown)


packages/html-reporter/src/testCaseView.css
```diff
@@ -73,4 +73,14 @@
...
   flex-direction: row;
   flex-wrap: wrap;
+}
+
+.annotation-copy-button{
+  padding-left: 3px;
+  display: none;
+}
+
+.test-case-annotation:hover .annotation-copy-button{
```
pavelfeldman, line 83: Sounds like you would need this dance for every copy button application. Would it be possible to do something like this to encapsulate these styles:

```
<CopyToClipboardContainer description={description}>
  <span className='annotation-copy-button'/>
</CopyToClipboardContainer>
```
```

File: packages/html-reporter/src/testCaseView.tsx [hunk 2]
```diff
@@ -68,15 +69,16 @@ function renderAnnotationDescription(description: string) {
...
-  } catch {}
+  } catch { }
   return description;
 }
 
 function TestCaseAnnotationView({ annotation: { type, description } }: { annotation: TestCaseAnnotation }) {
   return (
     <div className='test-case-annotation'>
-      <span style={{ fontWeight: 'bold' }}>{type}</span>
+      <span style={{ fontWeight: 'bold', display: 'inline-block', marginBlock: '3px' }}>{type}</span>
```
pavelfeldman, line 79: display: 'inline-block' - what else would this span be?

packages/html-reporter/src/testCaseView.css
amankagithub: Not every copy button , just for this annotation copy button .  We want to display the copy button only when we hover over the text . Could you please help me with how your approach would help in this use case ?

dgozman:
Closing this as stale. If you still want to pursue this change, please reopen or create a new PR and link to this one.

PR #30751, opened 2024-05-11
Filter test cases by annotation text in HTML report

amankagithub:
Feature request : https://github.com/microsoft/playwright/issues/30141
At Atlassian , we add team name , component name to annotations . If we want to search test case by teams on the HTML report we don't have any way . 

In this PR , I had added the functionality to search by annotation . 

Out of all tests there are two test cases with annotations . 
Screenshot 2024-05-11 at 5 06 47 PM
Screenshot 2024-05-22 at 2 30 01 PM
Screenshot 2024-05-22 at 2 30 11 PM

If we just add to `author` in annotation search , both the test case appears . 
Screenshot 2024-05-22 at 2 27 46 PM

If we add `author=aman` only one test case appears . Same with `author=alex`
Screenshot 2024-05-22 at 2 29 47 PM
Screenshot 2024-05-22 at 2 29 55 PM

### DIFF
--- a/packages/html-reporter/src/filter.ts
+++ b/packages/html-reporter/src/filter.ts
@@ -15,12 +15,12 @@
 */
 
 import type { TestCaseSummary } from './types';
-
 export class Filter {
   project: string[] = [];
   status: string[] = [];
   text: string[] = [];
   labels: string[] = [];
+  annotations: string[] = [];
 
   empty(): boolean {
     return this.project.length + this.status.length + this.text.length === 0;
@@ -32,6 +32,7 @@ export class Filter {
     const status = new Set<string>();
     const text: string[] = [];
     const labels = new Set<string>();
+    const annotations = new Set<string>();
     for (const token of tokens) {
       if (token.startsWith('p:')) {
         project.add(token.slice(2));
@@ -45,6 +46,10 @@ export class Filter {
         labels.add(token);
         continue;
       }
+      if (token.startsWith('a:')) {
+        annotations.add(token.slice(2));
+        continue;
+      }
       text.push(token.toLowerCase());
     }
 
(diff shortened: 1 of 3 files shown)


packages/html-reporter/src/filter.ts
```diff
@@ -45,6 +46,10 @@ export class Filter {
         labels.add(token);
         continue;
       }
+      if (token.startsWith('a:')) {
```
pavelfeldman, line 49: a: would be ambiguous here (thing attachments, etc). How about "annot:"

pavelfeldman (commented):
a: comment, otherwise looks good!

packages/html-reporter/src/filter.ts
```diff
@@ -45,6 +46,10 @@ export class Filter {
         labels.add(token);
         continue;
       }
+      if (token.startsWith('annot:')) {
+        annotations.add(token.slice(6));
```
pavelfeldman, line 50:
```suggestion
        annotations.add(token.substring('annot:'.length));
```

packages/html-reporter/src/filter.ts
```diff
@@ -127,7 +133,12 @@ export class Filter {
       if (!matches)
         return false;
     }
-
+    if (this.annotations.length) {
+      const matches = this.annotations.every(annotation =>
+        searchValues.annotations.some(_annotation => _annotation.includes(annotation)));
```
pavelfeldman, line 138:
```suggestion
        searchValues.annotations.some(a => a.includes(annotation)));
```

pavelfeldman (approved):

amankagithub:

pavelfeldman:
Thanks! I'll merge it once the nits from the previous review are addressed.

pavelfeldman (approved):

itaditya:
woohoo 🎉

PR #30765, opened 2024-05-13
feat: APIRequestContext dispose reason

dgozman:
Similarly to page.close, we pass test-runner specific reason to facilitate better error messages.

```
  1) a.test.ts:10:11 › test

    Error: apiRequestContext.fetch: Fixture { request } from beforeAll cannot be reused in a test.
      - Recommended fix: use a separate { request } in the test.
      - Alternatively, manually create APIRequestContext in beforeAll and dispose it in afterAll.
    See https://playwright.dev/docs/api-testing#sending-api-requests-from-ui-tests for more details.

       9 |
      10 |       test('test', async () => {
    > 11 |         await context.fetch('http://example.com');
         |                       ^
      12 |       });
      13 |
```

Closes #29260.

### DIFF
--- a/tests/playwright-test/playwright.spec.ts
+++ b/tests/playwright-test/playwright.spec.ts
@@ -827,3 +827,24 @@ test('should save trace in two APIRequestContexts', async ({ runInlineTest, serv
   expect(result.exitCode).toBe(0);
   expect(result.passed).toBe(1);
 });
+
+test('should explain a failure when using a dispose APIRequestContext', async ({ runInlineTest, server }) => {
+  const result = await runInlineTest({
+    'a.test.ts': `
+      import { test } from '@playwright/test';
+
+      let context;
+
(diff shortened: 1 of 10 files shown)


tests/playwright-test/playwright.spec.ts
```diff
@@ -827,3 +827,24 @@ test('should save trace in two APIRequestContexts', async ({ runInlineTest, serv
...
+      });
+
+      test('test', async () => {
+        await context.fetch('http://example.com');
+      });
+    `,
+  }, { workers: 1 });
+  expect(result.exitCode).toBe(1);
+  expect(result.passed).toBe(0);
+  expect(result.output).toContain('foobar');
```
yury-s, line 849: Replace 'foobar' with real expectation?

yury-s (approved):

PR #30769, opened 2024-05-13
fix(test runner): regular worker termination finishes long fixtures

### LINKED ISSUE #30504: [Bug]: Worker-scoped fixture execution terminates before timeout

dgozman:
Previously, terminating worker always had a 30 seconds force exit.

Now, regular worker termination assumes that process will eventually finish tearing down all the fixtures and exits. However, the self-destruction routine keeps the 30 seconds timeout to avoid zombies.

Fixes #30504.

### DIFF
--- a/packages/playwright/src/common/process.ts
+++ b/packages/playwright/src/common/process.ts
@@ -44,11 +44,12 @@ export class ProcessRunner {
   }
 }
 
-let closed = false;
+let gracefullyCloseCalled = false;
+let forceExitInitiated = false;
 
 sendMessageToParent({ method: 'ready' });
 
-process.on('disconnect', gracefullyCloseAndExit);
+process.on('disconnect', () => gracefullyCloseAndExit(true));
 process.on('SIGINT', () => {});
 process.on('SIGTERM', () => {});
 
@@ -76,7 +77,7 @@ process.on('message', async (message: any) => {
     const keys = new Set([...Object.keys(process.env), ...Object.keys(startingEnv)]);
     const producedEnv: EnvProducedPayload = [...keys].filter(key => startingEnv[key] !== process.env[key]).map(key => [key, process.env[key] ?? null]);
     sendMessageToParent({ method: '__env_produced__', params: producedEnv });
-    await gracefullyCloseAndExit();
+    await gracefullyCloseAndExit(false);
     return;
   }
   if (message.method === '__dispatch__') {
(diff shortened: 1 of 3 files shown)


packages/playwright/src/common/process.ts
```diff
@@ -92,19 +93,22 @@ process.on('message', async (message: any) => {
...
+  if (forceExit && !forceExitInitiated) {
+    forceExitInitiated = true;
+    // Force exit after 30 seconds.
+    // eslint-disable-next-line no-restricted-properties
+    setTimeout(() => process.exit(0), 30000);
+  }
+  if (!gracefullyCloseCalled) {
+    gracefullyCloseCalled = true;
+    // Meanwhile, try to gracefully shutdown.
+    await processRunner?.gracefullyClose().catch(() => {});
```
yury-s, line 106: If a fixture hangs while shutting down, it will keep the process running now. Can we avoid that?

yury-s (approved):

yury-s (approved):

PR #30778, opened 2024-05-14
devops: migrate electron & video bots to re-usable run-test action

mxschmitt:


.github/workflows/tests_secondary.yml
```diff
@@ -229,7 +229,7 @@ jobs:
         PWTEST_CHANNEL: chrome
 
   chromium_tot:
-    name: Chromium TOT ${{ matrix.os }}
```
mxschmitt, line 232: drive-by: so we can just grep for tip-of-tree when rolling ToT. Happy to do it the other way around as well. Right now some bots have ToT in their name, some tip-of-tree.

yury-s (approved):

PR #30795, opened 2024-05-14
Await `onApiCallEnd`/`onApiCallBegin` hooks

Andarist:
`onApiCallBegin` doesn't completely serve its purpose today.

1. it gets called [here](https://github.com/microsoft/playwright/blob/f56f9654612bc3c52f18e4d3c4bc9e38ae30831a/packages/playwright-core/src/client/channelOwner.ts#L148-L150) *before* the "API call"
2. that call gets executed by [this async wrapper](https://github.com/microsoft/playwright/blob/f56f9654612bc3c52f18e4d3c4bc9e38ae30831a/packages/playwright-core/src/client/clientInstrumentation.ts#L58-L61) that executes all hooks **in sequeuence**
3. Playwright adds one `onApiCallBegin` hook [here](https://github.com/microsoft/playwright/blob/f56f9654612bc3c52f18e4d3c4bc9e38ae30831a/packages/playwright/src/index.ts#L250) so whatever gets installed after it gets executed in the next microtask (since this hook is synchronous). That means that the next hooks get called **after** the API call has already been dispatched since the engine exits from the non-awaited instrumentation wrapper call and proceeds to `this._connection.sendMessageToServer`


packages/playwright-core/src/client/channelOwner.ts
```diff
@@ -188,7 +189,7 @@ export abstract class ChannelOwner<T extends channels.Channel = channels.Channel
       logApiCall(logger, `=> ${apiName} started`, isInternal);
       const apiZone: ApiZone = { apiName, frames, isInternal, reported: false, csi, callCookie, wallTime };
       const result = await zones.run('apiZone', apiZone, async () => await func(apiZone));
-      csi?.onApiCallEnd(callCookie);
+      await csi?.onApiCallEnd(callCookie);
```
Andarist, line 192: similar issue happens here when this is not awaited - the test runs to completion, closes the browser and code in `onApiCallEnd` can get rejected with "Test ended." error because the browser was already closes by the time it gets executed

pavelfeldman:
How did you discover this? We only have one instrumentation instance, so things should run in order. Starting with an issue would always be great as well!

Andarist:
> How did you discover this? 

We are hooking into this not-so-stable API in our `@replayio/playwright` package. This is the only way to achieve what we want because we have to evaluate functions at precise~ moments (before steps actually happen in the browser). We add annotations to our runtime to aid our time-travel experience in Replay.

pavelfeldman:
Looks like the bots are unhappy with your approach.

mxschmitt:
Looks like this was addressed in https://github.com/microsoft/playwright/pull/31054?

Andarist:
It improves the situation by differentiating between sync and async handlers/listeners. It still doesn't allow decorating API calls in a way that would wait for completion. 

Is this something you'd consider? I'm not sure if it's even worth creating a feature request for this... since this whole API is just not quite public 😢 

There is a chance that this might work for us since we merely inject `page.evaluate` calls that call a synchronous function in the browser. So technically we don't have to await it... it should still get executed roughly at the correct time now (all commands go through a single channel so if this timing issue will no longer affect us then all subsequent calls should actually get dispatched after our `page.evaluate` calls). It's less explicit than I'd like though 😅

pavelfeldman:
> Is this something you'd consider?

I thought you tried it and the world broke: https://github.com/microsoft/playwright/pull/30795#issuecomment-2123464810. What would you like us to consider?

> So technically we don't have to await it... it should still get executed roughly at the correct time now

We routinely rely on the serialized nature of the posted tasks. So I would not hesitate, just make sure you have a good test coverage.

dgozman:
It seems like this PR is stale, and discussion dried out, so I'll close it. If you still want to pursue this change, please reopen or open a new PR.

PR #30810, opened 2024-05-14
feat(trace-viewer): show nework request source id

### LINKED ISSUE #28903: [Feature] Separate Network logs when working with two contexts

Currently, if you are working with two contexts then in the trace file, there is no way to identify which network request is part of which context. Wondering if there is a way to identify that.

Screenshot 2024-01-08 at 2 26 36 PM

yury-s:
image

Fixes https://github.com/microsoft/playwright/issues/28903

### DIFF
--- a/packages/trace-viewer/src/ui/networkTab.tsx
+++ b/packages/trace-viewer/src/ui/networkTab.tsx
@@ -21,7 +21,7 @@ import './networkTab.css';
 import { NetworkResourceDetails } from './networkResourceDetails';
 import { bytesToString, msToString } from '@web/uiUtils';
 import { PlaceholderPanel } from './placeholderPanel';
-import type { MultiTraceModel } from './modelUtil';
+import { context, type MultiTraceModel } from './modelUtil';
 import { GridView, type RenderedGridCell } from '@web/components/gridView';
 import { SplitView } from '@web/components/splitView';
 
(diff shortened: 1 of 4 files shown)


packages/trace-viewer/src/ui/networkTab.tsx
```diff
@@ -159,7 +182,67 @@ const renderCell = (entry: RenderedEntry, column: ColumnName): RenderedGridCell
...
+      this._pagerefToShortId.set(pageref, shortId);
+    }
+    return shortId;
+  }
+
+  private _apiRequestContextId(resource: Entry): string {
+    const contextEntry = context(resource);
+    if (!contextEntry)
+      return '';
+    let contextId = (contextEntry as any)[ContextIdGenerator._apiRequestContextIdSymbol];
```
dgozman, line 225: Why not use a map?

packages/trace-viewer/src/ui/workbench.tsx
```diff
@@ -172,7 +172,7 @@ export const Workbench: React.FunctionComponent<{
     id: 'network',
     title: 'Network',
     count: networkModel.resources.length,
-    render: () => <NetworkTab boundaries={boundaries} networkModel={networkModel} onEntryHovered={setHighlightedEntry}/>
+    render: () => <NetworkTab boundaries={boundaries} networkModel={networkModel} model={model} onEntryHovered={setHighlightedEntry}/>
```
dgozman, line 175: Can we avoid passing `model`? Perhaps calculate all the ids beforehand and put them into `NetworkModel` for better separation of concerns?

dgozman (approved):

Issue #30368, opened 2024-04-15
[Feature]: Support Ubuntu 24.04


Mister-Hope:
### 🚀 Feature Request

Ubuntu 24 will be released in this month, and the beta version is already release in 12th. (See https://wiki.ubuntu.com/Releases)

Hope playweight can add support once it release

### Motivation

Cuurently when running `pnpm exec playwright install --with-deps` on Ubuntu 24.04, errors are thrown

```
Hit:1 http://security.ubuntu.com/ubuntu noble-security InRelease
Get:2 http://archive.ubuntu.com/ubuntu noble InRelease [255 kB]
Get:3 http://archive.ubuntu.com/ubuntu noble-updates InRelease [89.7 kB]
Hit:4 http://archive.ubuntu.com/ubuntu noble-backports InRelease
Get:5 http://archive.ubuntu.com/ubuntu noble/main amd64 Packages [1404 kB]
Get:6 http://archive.ubuntu.com/ubuntu noble/universe amd64 Packages [15.2 MB]
Get:7 http://archive.ubuntu.com/ubuntu noble/universe Translation-en [6029 kB]
Get:8 http://archive.ubuntu.com/ubuntu noble/multiverse amd64 Packages [241 kB]                                                                                          
Fetched 23.2 MB in 7s (3262 kB/s)                                                                                                                                        
Reading package lists... Done
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'libatk-bridge2.0-0t64' instead of 'libatk-bridge2.0-0'
Note, selecting 'libatk1.0-0t64' instead of 'libatk1.0-0'
Note, selecting 'libatspi2.0-0t64' instead of 'libatspi2.0-0'
Note, selecting 'libcups2t64' instead of 'libcups2'
Note, selecting 'libgtk-3-0t64' instead of 'libgtk-3-0'
Note, selecting 'libpng16-16t64' instead of 'libpng16-16'
Note, selecting 'libevent-2.1-7t64' instead of 'libevent-2.1-7'
Package libasound2 is a virtual package provided by:
  liboss4-salsa-asound2 4.2-build2020-1ubuntu3
  libasound2t64 1.2.11-1build2 (= 1.2.11-1build2)
You should explicitly select one to install.

E: Package 'libasound2' has no installation candidate
E: Unable to locate package libicu70
E: Unable to locate package libffi7
E: Unable to locate package libx264-163
Failed to install browsers
[...]

PR #30825, opened 2024-05-15
feat: support Ubuntu 24.04

mxschmitt:
https://github.com/microsoft/playwright/issues/30368


mxschmitt:
Opening from the origin instead.

PR #30855, opened 2024-05-16
fix(electron): flaky startup if stderr comes in too fast

JoelEinbinder:
Chromium's `DevTools listening on` message sometimes arrives before Playwright is finished connecting to Node. Without this patch, it would miss the message and fail to connect.

### DIFF
--- a/packages/playwright-core/src/server/electron/electron.ts
+++ b/packages/playwright-core/src/server/electron/electron.ts
@@ -221,6 +221,8 @@ export class Electron extends SdkObject {
         onExit: () => app?.emit(ElectronApplication.Events.Close),
       });
 
+      // All waitForLines must be started immediately.
+      // Otherwise the lines might come before we are ready.
       const waitForXserverError = new Promise(async (resolve, reject) => {
         waitForLine(progress, launchedProcess, /Unable to open X display/).then(() => reject(new Error([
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/electron/electron.ts
```diff
@@ -232,17 +234,20 @@ export class Electron extends SdkObject {
           progress.metadata.log
         ].join('\n')))).catch(() => {});
       });
+      const nodeMatchPromise = waitForLine(progress, launchedProcess, /^Debugger listening on (ws:\/\/.*)$/);
+      const chromeMatchPromise = waitForLine(progress, launchedProcess, /^DevTools listening on (ws:\/\/.*)$/);
+      const debuggerDisconnectPromise = waitForLine(progress, launchedProcess, /Waiting for the debugger to disconnect\.\.\./);
 
-      const nodeMatch = await waitForLine(progress, launchedProcess, /^Debugger listening on (ws:\/\/.*)$/);
+      const nodeMatch = await nodeMatchPromise;
```
mxschmitt, line 241: nit: move nodeMatchPromise to the end and await immediately?

mxschmitt (approved):

pavelfeldman (approved):

PR #30860, opened 2024-05-16
Add a way to debug browser logs

Andarist:
This would be immensely useful for debugging purposes

### DIFF
--- a/packages/playwright-core/src/server/browserType.ts
+++ b/packages/playwright-core/src/server/browserType.ts
@@ -33,6 +33,7 @@ import type * as types from './types';
 import type * as channels from '@protocol/channels';
 import { DEFAULT_TIMEOUT, TimeoutSettings } from '../common/timeoutSettings';
 import { debugMode } from '../utils';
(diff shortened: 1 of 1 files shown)


pavelfeldman:
Could you explain which messages you are currently missing with the DEBUG=pw:browser today?

Andarist:
We are configuring Playwright to use our fork of Chromium at https://www.replay.io/ . At times, it's helpful for us to debug things happening there using its verbose stdout/stderr output. We can pass the required `RECORD_REPLAY_VERBOSE` env variable to it using `launchOptions` but we still miss the logs because the browser's output is not observable without this added `debug` call

pavelfeldman:
> Could you explain which messages you are currently missing with the DEBUG=pw:browser today?

^^

pavelfeldman:
I.e. if you run DEBUG=pw:browser npx playwright test, what kind of output does not end up printed?

pavelfeldman:
Let's follow the process and file an issue so that we could triage it!

PR #30885, opened 2024-05-18
feature(trace-viewer): embedded mode support PoC

ruifigueira:
Companion PR of https://github.com/microsoft/playwright-vscode/pull/483


pavelfeldman:
I'm supportive for those, let me know when they are no longer drafts!

ruifigueira:
@pavelfeldman I removed the draft. This is still a very basic implementation just to check if the integration was possible.

I found out some issues with this approach: when the trace viewer iframe gets the focus keyboard shortcuts stop working, and I ned to click outside it to open the command palette for instance.

pavelfeldman:
> I found out some issues with this approach: when the trace viewer iframe gets the focus keyboard shortcuts stop working, and I ned to click outside it to open the command palette for instance.

That's unfortunate! I was rendering webviews in vscode w/o such issues. Is iframe w/ snapshot breaking it? How is service worker doing, are snapshots rendered fine?

ruifigueira:
I tried to render the trace viewer html directly without iframe and [service worker doesn't work therer](https://github.com/microsoft/vscode/issues/194751). Nevertheless, we can just rely on the running server instead of the service worker. The only thing I still need to figure out is on how to make the websocket work, or if not possible, on how to emulate it.

pavelfeldman:
> we can just rely on the running server instead of the service worker.

Unfortunately, we can't. It is essential for SW to fulfill all the requests.

ruifigueira:
I am figuring that out the hard way :( I was doing some quick and dirty tests and deactivated the sw to rely on the webserver. I thought it was already handling trace requests as a fallback, but it was not, so I had to copy & paste the service worker logic there with necessary adjustments for it to work, but I stumbled on cross-domain snapshot resources that are handled by the sw and I don't see a way out of that...

I'll go back to the iframe solution and see if there's a workaround for it.

ruifigueira:
Problem explained here, with a workaround proposal:

https://github.com/microsoft/vscode/issues/65452#issuecomment-586036474

ruifigueira:
That did the trick, key bindings are triggered as expected when trace viewer is focused

packages/trace-viewer/src/index.tsx
```diff
@@ -39,5 +39,26 @@ import { WorkbenchLoader } from './ui/workbenchLoader';
     setInterval(function() { fetch('ping'); }, 10000);
   }
 
+  if (window.parent !== window && new URL(window.location.href).searchParams.has('embedded')) {
```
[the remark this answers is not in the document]
ruifigueira: Unfortunatelly no, because this code must run inside the iframe. But maybe create an `embedded.html` / `embedded.tsx`, similar to `index.tsx` and `uiMode.tsx`, we can just drop this code there as well as instanciating the parameterized `WorkbenchLoader`

ruifigueira:
The snapshot popout link is not working (`window.open` doesn't work in vscode). Using [vscode.env.openExternalUrl](https://code.visualstudio.com/api/references/vscode-api#env) does not work either because it won't share the sw.

We can either disable it (it will still work on non-embedded mode) or delegate it to the server (for instance, it can launch a playwright-controlled browser and inject the trace file for sw to load it before opening the `snapshot.html` URL).

ruifigueira:
I found out why I thought I needed to inject the trace file before opening snapshot.html. It's a bug: https://github.com/microsoft/playwright/pull/31033

packages/trace-viewer/src/index.tsx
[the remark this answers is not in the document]
pavelfeldman: > Unfortunatelly no, because this code must run inside the iframe.

I understand that the snapshot renderer should run in an iframe, but the app itself does not have to. Are you doing this for convenience?

packages/trace-viewer/src/ui/workbenchLoader.tsx
```diff
@@ -24,7 +24,8 @@ import { Workbench } from './workbench';
 import { TestServerConnection } from '@testIsomorphic/testServerConnection';
 
 export const WorkbenchLoader: React.FunctionComponent<{
-}> = () => {
+  embedded?: boolean
```
pavelfeldman, line 27:
```suggestion
  supportTraceDrop?: boolean
```

pavelfeldman (approved):

packages/trace-viewer/src/index.tsx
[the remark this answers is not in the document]
ruifigueira: The app depends on the SW, and for that reason there's no easy way to render it without the iframe because [webviews don't support SW ](https://github.com/microsoft/vscode/issues/194751)

packages/trace-viewer/src/ui/workbenchLoader.tsx
ruifigueira: `embedded` mode only hides the trace viewer header, so maybe call it `hideHeader` instead?

ruifigueira:
Quick status on this:

- the implementation with trace viewer rendered in iframe seems stable. and I added some tests on playwright-vscode side
  - until now, the only drawback I found was the key binding issue, which I "fixed" with [this workaround ](https://github.com/microsoft/playwright/pull/30885?notification_referrer_id=NT_kwDOABT5X7MxMDcxMjA1MjcyNzoxMzc0NTU5#issuecomment-2128294323)
- I added a setting to enable / disable embedded mode. Besides, embedded mode is only active if playwright version supports it
- embedded mode no longer requires that playwright server is launched with 0.0.0.0 when running remotely, because vscode performs local port forwarding
- embedded mode also works with devcontainers, remote SSH and [codespaces](https://github.com/microsoft/playwright-vscode/pull/483#issuecomment-2174560714)

I also tried once again to implement it without the iframe (the issues I stumbed when I first tried it were mostly caused by #31033. Even if not stable yet, I'm already confident that is possible to do it. It adds some complexity, though, because playwright server needs to serve trace resources just like service worker (we cannot rely on service worker when not rendering the trace viewer on an iframe, as explained before). You can find its current implementation here: https://github.com/ruifigueira/playwright/compare/feat/embedded-trace-viewer...ruifigueira:playwright:feat/embedded-trace-viewer-no-iframe

Let me know your thoughts if it's worth to go with this solution, or if the iframe solution is good enough.

chensce:
the iframe solution is good enough

chensce:
look forward for inspect（code gen） embedded mode

chensce:
record embedded mode

ruifigueira:
@chensce this is for trace viewer only, no recorder functionality is covered here, for that you already have the vscode recorder.

pavelfeldman:
> Let me know your thoughts if it's worth to go with this solution, or if the iframe solution is good enough.

Ok, let's try the iframe solution.

Looking at the code, I think we can improve the situation via untangling some of the existing complexities. Here is my thinking:
- Modern VS Code extension and UI Mode share the same testServer and communicate over the same wire, basically there is a lot of code reuse where either UI Mode or VS Code extension is the front-end working against the same server code.
- This server is actually inherited from trace viewer - `TestServer.start` calls `startTraceViewerServer`.
- When VS Code extension finds a Playwright config, it starts a server for it and uses it to list files, tests, etc. This means that you already have a trace viewer server running that can server static files and load traces.
- The live mode won't work because TraceViewer opened against test server won't have `onLoadTraceRequested` triggered - that is usually wired via the stdin in a standalone mode. But you can trigger it manually via sending a message to `embedded.tsx`.
- In the longer run, I don't think you should reuse `WorkbenchLoader` in `embedder.tsx` and rather use `Workbench` instead. There will be some code duplication, but it'll make it clear that embedder is using one way to load live traces and the standalone more would use another way. Think about it this way, we'll have:
- UI mode app
- Trace Viewer app
- Trace Viewer with StdIn app (external Trace viewer when used from VS Code extension)
- Embedder app (to use in a web view in VS Code extension)

ruifigueira:
> * The live mode won't work because TraceViewer opened against test server won't have onLoadTraceRequested triggered - that is usually wired via the stdin in a standalone mode. But you can trigger it manually via sending a message to embedded.tsx.
> * In the longer run, I don't think you should reuse WorkbenchLoader in embedder.tsx and rather use Workbench instead.

I was already addressing those two points in the [EmbeddedWorkbenchLoader](https://github.com/ruifigueira/playwright/commit/b8f0d427aca5830b7b53a5b2f214d8564a864476#diff-2d8293300b803ff227e5bce5e3ae6d60ef0df66777d80ad87616964a82bff772R36-R42) in my non-iframe alternative, I'll cherry-pick those changes.

pavelfeldman:
> I was already addressing those two points in the [EmbeddedWorkbenchLoader](https://github.com/ruifigueira/playwright/commit/b8f0d427aca5830b7b53a5b2f214d8564a864476#diff-2d8293300b803ff227e5bce5e3ae6d60ef0df66777d80ad87616964a82bff772R36-R42) in my non-iframe alternative, I'll cherry-pick those changes.

You can keep it as an iframe at first, to make changes more local.

ruifigueira:
I just updated this and vscode PR to accomodate your suggestions. Reusing the testServer was a great tip, it simplifies a lot and it's one less process to manage.

packages/trace-viewer/src/embedded.tsx
```diff
@@ -0,0 +1,69 @@
...
+      metaKey: e.metaKey,
+      repeat: e.repeat,
+    }, '*');
+  };
+  window.addEventListener('keydown', handleKeyEvent);
+  window.addEventListener('keyup', handleKeyEvent);
+
+  if (window.location.protocol !== 'file:') {
+    if (window.location.href.includes('isUnderTest=true'))
+      await new Promise(f => setTimeout(f, 1000));
```
pavelfeldman, line 49: I don't think we need it, I'll remove it from the other ones too.

packages/trace-viewer/src/ui/embeddedWorkbenchLoader.tsx
```diff
@@ -0,0 +1,91 @@
...
+      if (method === 'loadTraceRequested') {
+        setTraceURLs(params.traceUrl ? [params.traceUrl] : []);
+        setProcessingErrorMessage(null);
+      } else if (method === 'applyTheme') {
+        if (currentTheme() !== params.theme)
+          toggleTheme();
+      }
+    });
+    // notify vscode that it is now listening to its messages
+    window.parent?.postMessage({ type: 'loaded' }, '*');
```
pavelfeldman, line 41:
```suggestion
    window.parent!.postMessage({ type: 'loaded' }, '*');
```

packages/trace-viewer/src/ui/popout.ts
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+type PopoutFunction = (url: string, target?: string) => Window | any;
+
+let popoutFn: PopoutFunction = window.open;
+export function setPopoutFunction(fn: PopoutFunction) {
```
pavelfeldman, line 20: Let's plumb this via `<Workspace openPage={}>` instead, it is just two hops. Loaders will pass respective implementations.

pavelfeldman (approved):
A couple of nits and it is good to go.

pavelfeldman (approved):

PR #30904, opened 2024-05-20
fix(fetch): allow UTF-8 in Location header

### LINKED ISSUE #30903: [Bug]: redirect in route.fetch may jump to a wrong url
### Version

1.43.0

### Steps to reproduce

Example steps (replace with your own):
1. Clone my repo at [https://github.com/tqobqbq/playwright_bug_test.git](https://github.com/tqobqbq/playwright_bug_test.git)
2. python test.py
3. You should see the error come up

### Expected behavior

I expect 'https://www.suruga-ya.jp/search?search_word=%E3%83%9E' will jump to ' https://www.suruga-ya.jp/search?category=&search_word=%E3%83%9E'
And if I set a route function:
`
        def route_continue(route):
            return route.continue_()
`
or without setting any route function,I can get the proper behavior

### Actual behavior

however,if I set a route function:
`
        def route_fetch(route):
            return route.fulfill(response=route.fetch(max_redirects=0))
`
It will jump to a wrong url 'https://www.suruga-ya.jp/search?category=&search_word=%C3%A3%C2%83%C2%9E'

### Additional context

I think the error may stem from the url parse of response.headers['location'] in route.fetch

### Environment

```Text
- CPU: [arm64]
- Python Version: [3.9]
- Other info:
```

mxschmitt:
This brings up the discussion if we want to post-process all the headers as UTF-8 as well.

Fixes https://github.com/microsoft/playwright/issues/30903#issuecomment-2119959937

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -336,12 +336,15 @@ export abstract class APIRequestContext extends SdkObject {
             redirectOptions.rejectUnauthorized = false;
 
           // HTTP-redirect fetch step 4: If locationURL is null, then return response.
-          if (response.headers.location) {
+          // Best-effort UTF-8 decoding, per spec it's US-ASCII only, but browsers are more lenient.
+          // Node.js parses it as Latin1, so we convert it to UTF-8.
+          const locationHeaderValue = Buffer.from(response.headers.location ?? '', 'latin1').toString('utf8');
+          if (locationHeaderValue) {
             let locationURL;
             try {
-              locationURL = new URL(response.headers.location, url);
+              locationURL = new URL(locationHeaderValue, url);
             } catch (error) {
-              reject(new Error(`uri requested responds with an invalid redirect URL: ${response.headers.location}`));
+              reject(new Error(`uri requested responds with an invalid redirect URL: ${locationHeaderValue}`));
               request.destroy();
               return;
             }
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -337,7 +337,7 @@ export abstract class APIRequestContext extends SdkObject {
 
           // HTTP-redirect fetch step 4: If locationURL is null, then return response.
           // Best-effort UTF-8 decoding, per spec it's US-ASCII only, but browsers are more lenient.
-          // Node.js parses it as Latin1, so we convert it to UTF-8.
+          // Node.js parses it as Latin1 via std::v8::String, so we convert it to UTF-8.
           const locationHeaderValue = Buffer.from(response.headers.location ?? '', 'latin1').toString('utf8');
           if (locationHeaderValue) {
             let locationURL;
--- a/tests/library/browsercontext-fetch.spec.ts
+++ b/tests/library/browsercontext-fetch.spec.ts
@@ -198,6 +198,20 @@ it('should follow redirects', async ({ context, server }) => {
   expect(await response.json()).toEqual({ foo: 'bar' });
 });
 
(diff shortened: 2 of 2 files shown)


packages/playwright-core/src/server/fetch.ts
```diff
@@ -336,12 +336,14 @@ export abstract class APIRequestContext extends SdkObject {
             redirectOptions.rejectUnauthorized = false;
 
           // HTTP-redirect fetch step 4: If locationURL is null, then return response.
-          if (response.headers.location) {
+          // Best-effort UTF-8 decoding, per spec it's US-ASCII only, but browsers are more lenient.
+          const locationHeaderValue = Buffer.from(response.headers.location ?? '', 'binary').toString('utf8');
+          if (locationHeaderValue) {
```
yury-s, line 341: Should we also reject if location is present but cannot be parsed to utf-8?

packages/playwright-core/src/server/fetch.ts
```diff
@@ -336,12 +336,14 @@ export abstract class APIRequestContext extends SdkObject {
             redirectOptions.rejectUnauthorized = false;
 
           // HTTP-redirect fetch step 4: If locationURL is null, then return response.
-          if (response.headers.location) {
+          // Best-effort UTF-8 decoding, per spec it's US-ASCII only, but browsers are more lenient.
+          const locationHeaderValue = Buffer.from(response.headers.location ?? '', 'binary').toString('utf8');
```
yury-s, line 340: response.headers.location is already a string, which encoding is it in?

yury-s: Let's change it to `latin1` and add a comment that it is constructed as a one byte in v8 via [NewFromOneByte](https://source.chromium.org/chromium/chromium/src/+/main:v8/include/v8-primitive.h;bpv=1;bpt=1;l=432?q=v8::String::NewFromOneByte&ss=chromium%2Fchromium%2Fsrc&gsn=NewFromOneByte&gs=KYTHE%3A%2F%2FKrQCCpcBa3l0aGU6Ly9jaHJvbWl1bS5nb29nbGVzb3VyY2UuY29tL2NvZGVzZWFyY2gvY2hyb21pdW0vc3JjLy9tYWluP2xhbmc9YyUyQiUyQj9wYXRoPXY4L2luY2x1ZGUvdjgtcHJpbWl0aXZlLmgjM1ViNFRWNXdyQmRvYU1pUkRNVWc2LUFIdk9ILTN1THExUVBCcnF0T1BRMAqXAWt5dGhlOi8vY2hyb21pdW0uZ29vZ2xlc291cmNlLmNvbS9jb2Rlc2VhcmNoL2Nocm9taXVtL3NyYy8vbWFpbj9sYW5nPWMlMkIlMkI_cGF0aD12OC9pbmNsdWRlL3Y4LXByaW1pdGl2ZS5oI2V3SHBHX1lfVXRKOXJ4YzN0eVo0aXlxSXdmMFJ6d1ROOGVjYnczWVV3Mmc%3D)(

yury-s (approved):

packages/playwright-core/src/server/fetch.ts
```diff
@@ -336,12 +336,15 @@ export abstract class APIRequestContext extends SdkObject {
             redirectOptions.rejectUnauthorized = false;
 
           // HTTP-redirect fetch step 4: If locationURL is null, then return response.
-          if (response.headers.location) {
+          // Best-effort UTF-8 decoding, per spec it's US-ASCII only, but browsers are more lenient.
+          // Node.js parses it as Latin1, so we convert it to UTF-8.
```
mxschmitt, line 340:
```suggestion
          // Node.js parses it as Latin1 via NewFromOneByte, so we convert it to UTF-8.
```

PR #30912, opened 2024-05-20
feat(reporters): align and document environment variables

dgozman:
- Documents `PLAYWRIGHT_FORCE_TTY` and `FORCE_COLOR` across terminal reporters.
- New `PLAYWRIGHT_LIST_PRINT_STEPS`. Removes undocumented test-only `PW_TEST_DEBUG_REPORTERS_PRINT_STEPS`.
- Replaces `PLAYWRIGHT_HTML_REPORT` with `PLAYWRIGHT_HTML_OUTPUT_DIR` and `PW_TEST_HTML_REPORT_OPEN` with `PLAYWRIGHT_HTML_OPEN` for consistency, supports older versions for backwards compatibility.
- New `PLAYWRIGHT_HTML_HOST`, `PLAYWRIGHT_HTML_PORT` and `PLAYWRIGHT_HTML_ATTACHMENTS_BASE_URL`.
- New `PLAYWRIGHT_JUNIT_STRIP_ANSI` and `PLAYWRIGHT_JUNIT_INCLUDE_PROJECT_IN_TEST_NAME`.
- Removes `PW_HTML_REPORT` that was set for unknown reason.

### DIFF
--- a/packages/playwright-core/src/utils/env.ts
+++ b/packages/playwright-core/src/utils/env.ts
@@ -21,9 +21,13 @@ export function getFromENV(name: string): string | undefined {
   return value;
 }
 
-export function getAsBooleanFromENV(name: string): boolean {
+export function getAsBooleanFromENV(name: string, defaultValue?: boolean | undefined): boolean {
   const value = getFromENV(name);
-  return !!value && value !== 'false' && value !== '0';
+  if (value === 'false' || value === '0')
+    return false;
+  if (value)
+    return true;
+  return !!defaultValue;
 }
 
 export function getPackageManager() {
--- a/packages/playwright/src/reporters/html.ts
+++ b/packages/playwright/src/reporters/html.ts
@@ -62,14 +62,14 @@ class HtmlReporter extends EmptyReporter {
   private _outputFolder!: string;
   private _attachmentsBaseURL!: string;
   private _open: string | undefined;
+  private _port: number | undefined;
+  private _host: string | undefined;
   private _buildResult: { ok: boolean, singleTestId: string | undefined } | undefined;
   private _topLevelErrors: TestError[] = [];
 
   constructor(options: HtmlReporterOptions) {
     super();
     this._options = options;
-    if (options._mode === 'test')
-      process.env.PW_HTML_REPORT = '1';
   }
 
   override printsToStdio() {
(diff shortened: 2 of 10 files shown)


packages/playwright-core/src/utils/env.ts
```diff
@@ -21,9 +21,13 @@ export function getFromENV(name: string): string | undefined {
...
 
-export function getAsBooleanFromENV(name: string): boolean {
+export function getAsBooleanFromENV(name: string, defaultValue: boolean | undefined = false): boolean {
   const value = getFromENV(name);
-  return !!value && value !== 'false' && value !== '0';
+  if (value === 'false' || value === '0')
+    return false;
+  if (value)
+    return true;
+  return defaultValue;
```
yury-s, line 30: should either be !!defaultValue or the argument type should be boolean.

packages/playwright/src/reporters/html.ts
```diff
@@ -151,18 +155,20 @@ class HtmlReporter extends EmptyReporter {
 }
 
 function reportFolderFromEnv(): string | undefined {
-  if (process.env[`PLAYWRIGHT_HTML_REPORT`])
-    return path.resolve(process.cwd(), process.env[`PLAYWRIGHT_HTML_REPORT`]);
+  // Note: PLAYWRIGHT_HTML_REPORT is for backwards compatibility.
+  const envValue = process.env.PLAYWRIGHT_HTML_OUTPUT_DIR || process.env.PLAYWRIGHT_HTML_REPORT;
+  if (envValue)
+    return path.resolve(process.cwd(), envValue);
```
yury-s, line 161:
```suggestion
    return path.resolve(envValue);
```

packages/playwright/src/reporters/list.ts
```diff
@@ -33,9 +34,9 @@ class ListReporter extends BaseReporter {
   private _needNewLine = false;
   private _printSteps: boolean;
 
-  constructor(options: { omitFailures?: boolean, printSteps?: boolean } = {}) {
-    super(options);
```
yury-s, line 37: why is `options` not passed anymore, base.ts still reads omitFailures?

yury-s (approved):

packages/playwright-core/src/utils/env.ts
dgozman: Indeed!

packages/playwright/src/reporters/list.ts
dgozman: We do not have a public `omitFailures` option, and we do not use it privately either.

PR #30972, opened 2024-05-23
feat(config): respectGitIgnore option

### LINKED ISSUE #30553: [Feature]: Make "respectGitIgnore" a configurable parameter
### 🚀 Feature Request

Add the ability to override the default value for the currently internal "respectGitIgnore" project value in the config file. 

I'm willing to make the change if there is agreement on the feature.

### Example

```typescript

export default defineConfig({
  // Look for test files in the "tests" directory, relative to this configuration file.
  testDir: 'tests',

  // Run all tests in parallel.
  fullyParallel: true,

  // Should the .gitignore file be used to filter out tests
  respectGitIgnore: false
  
  
```

### Motivation

I just spent about 25 hours trying to debug why my tests weren't being run. I generate my test files dynamically, since I'm using Playwright to drive another test runner that is a web app, and with the original implementation in Puppeteer we found that breaking the tests into subgroups was a huge performance win. 

Since these test files are dynamically generated, they are added to the .gitignore so that any failure to clean them up (test runner crash) doesn't cause the file to get added to the repo by mistake.

In porting to Playwright, I didn't add an explicit "testDir", mostly because we didn't have one with Playwright. 

This combination led to Playwright not seeing the generated test files. And every time I changed the name of the generated files, I'd update the .gitignore, so the problem was bizarre to debug to say the least. It appeared like there was some kind of a cache issue.

yury-s:
Fixes https://github.com/microsoft/playwright/issues/30553

### DIFF
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -443,9 +443,7 @@ Test files that took more than `threshold` milliseconds are considered slow, and
 * since: v1.45
 - type: ?<[boolean]>
 
-Whether to skip entries from `.gitignore` when searching for test files. By default it is true, i.e. if
-neither [`property: TestConfig.retries`] nor [`property: TestProject.retries`] are explicitely specified,
-Playwright will ignore any test files matching `.gitignore` entries. This option allows to override that behavior.
+Whether to skip entries from `.gitignore` when searching for test files. By default, if neither [`property: TestConfig.testDir`] nor [`property: TestProject.testDir`] are explicitely specified, Playwright will ignore any test files matching `.gitignore` entries. This option allows to override that behavior.
 
 ## property: TestConfig.retries
 * since: v1.10
--- a/docs/src/test-api/class-testconfig.md
+++ b/docs/src/test-api/class-testconfig.md
@@ -439,6 +439,14 @@ export default defineConfig({
 
 Test files that took more than `threshold` milliseconds are considered slow, and the slowest ones are reported, no more than `max` number of them. Passing zero as `max` reports all test files that exceed the threshold.
 
+## property: TestConfig.respectGitIgnore
+* since: v1.45
+- type: ?<[boolean]>
+
+Whether to skip entries from `.gitignore` when searching for test files. By default it is true, i.e. if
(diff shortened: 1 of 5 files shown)


docs/src/test-api/class-testconfig.md
```diff
@@ -439,6 +439,14 @@ export default defineConfig({
 
 Test files that took more than `threshold` milliseconds are considered slow, and the slowest ones are reported, no more than `max` number of them. Passing zero as `max` reports all test files that exceed the threshold.
 
+## property: TestConfig.respectGitIgnore
+* since: v1.45
+- type: ?<[boolean]>
+
+Whether to skip entries from `.gitignore` when searching for test files. By default it is true, i.e. if
```
dgozman, line 446: How is `retries` related to `.gitignore`?

packages/playwright/src/common/config.ts
```diff
@@ -194,7 +194,7 @@ export class FullProjectInternal {
       const stylePaths = Array.isArray(this.expect.toHaveScreenshot.stylePath) ? this.expect.toHaveScreenshot.stylePath : [this.expect.toHaveScreenshot.stylePath];
       this.expect.toHaveScreenshot.stylePath = stylePaths.map(stylePath => path.resolve(configDir, stylePath));
     }
-    this.respectGitIgnore = !projectConfig.testDir && !config.testDir;
+    this.respectGitIgnore = projectConfig.respectGitIgnore ?? config.respectGitIgnore ?? (!projectConfig.testDir && !config.testDir);
```
dgozman, line 197: nit: use `takeFirst()`?

docs/src/test-api/class-testconfig.md
yury-s: It's `entries` not `retries`

dgozman, line 446: Let's not say that by default it is true? Instead, explain that default behavior is respecting `.gitignore` when neither `TestConfig.testDir` nor `TestProject.testDir` are specified. Because that's not equal to setting the property to `true` or `false`.

dgozman (approved):

PR #30999, opened 2024-05-24
feat(api): add host option in launchServer options

### LINKED ISSUE #30998: [Feature]: launchServer bind hostname or address
### 🚀 Feature Request

I really like the server method that playwright provides so that I can remotely control the test nodes. Currently, I can use docker to start a playwright server and expose ports to a control node playwright client for testing. In contrast, when I run playwright server directly without docker, I am only able to get the ws url of `localhost`, which prevents me from connecting to it directly through other control nodes within a connectable network.

### Example

```
const { chromium, webkit, firefox } = require('playwright');  // Or 'webkit' or 'firefox'.

(async () => {
  [chromium, webkit, firefox].forEach(async browserType => {
    const browserServer = await browserType.launchServer({
      // host: "0.0.0.0", // unsupported yet
      // port: 55432,
      // wsPath: "",
    });
    const wsEndpoint = browserServer.wsEndpoint();
    console.log(wsEndpoint, "for", browserType.name());
  })
  // Close browser instance.
  // await browserServer.close();
})();
```

### Motivation

Improve API usability and playability.

cavivie:
Closes [30998](https://github.com/microsoft/playwright/issues/30998).

### DIFF
--- a/docs/src/api/class-android.md
+++ b/docs/src/api/class-android.md
@@ -202,6 +202,12 @@ Prevents automatic playwright driver installation on attach. Assumes that the dr
 Optional device serial number to launch the browser on. If not specified, it will
 throw if multiple devices are connected.
 
+### option: Android.launchServer.host
+* since: v1.45
+- `host` <[string]>
+
+Host to use for the web socket. It is optional and if it is omitted, the server will accept connections on the unspecified IPv6 address (::) when IPv6 is available, or the unspecified IPv4 address (0.0.0.0) otherwise. Consider hardening it with picking a specific interface.
+
 ### option: Android.launchServer.port
 * since: v1.28
 - `port` <[int]>
--- a/docs/src/api/class-browsertype.md
+++ b/docs/src/api/class-browsertype.md
@@ -380,6 +380,12 @@ const { chromium } = require('playwright');  // Or 'webkit' or 'firefox'.
 ### option: BrowserType.launchServer.logger = %%-browser-option-logger-%%
 * since: v1.8
 
+### option: BrowserType.launchServer.host
+* since: v1.45
+- `host` <[string]>
+
+Host to use for the web socket. It is optional and if it is omitted, the server will accept connections on the unspecified IPv6 address (::) when IPv6 is available, or the unspecified IPv4 address (0.0.0.0) otherwise. Consider hardening it with picking a specific interface.
+
 ### option: BrowserType.launchServer.port
 * since: v1.8
 - `port` <[int]>
--- a/tests/library/browsertype-launch-server.spec.ts
+++ b/tests/library/browsertype-launch-server.spec.ts
@@ -26,6 +26,13 @@ it.describe('launch server', () => {
     await browserServer.close();
   });
 
+  it('should work with host', async ({ browserType }) => {
+    const host = '0.0.0.0';
+    const browserServer = await browserType.launchServer({ host });
+    expect(browserServer.wsEndpoint()).toContain(String(host));
+    await browserServer.close();
+  });
+
   it('should work with port', async ({ browserType }, testInfo) => {
     const port = 8800 + testInfo.workerIndex;
     const browserServer = await browserType.launchServer({ port });
--- a/docs/src/api/class-browserserver.md
+++ b/docs/src/api/class-browserserver.md
@@ -31,3 +31,5 @@ Browser websocket url.
 
 Browser websocket endpoint which can be used as an argument to [`method: BrowserType.connect`] to establish connection
 to the browser.
+
+Note that if the listen `host` option in `launchServer` options is not specified, localhost will be output anyway, even if the actual listening address is an unspecified address.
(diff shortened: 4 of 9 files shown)


docs/src/api/class-android.md
```diff
@@ -202,6 +202,12 @@ Prevents automatic playwright driver installation on attach. Assumes that the dr
 Optional device serial number to launch the browser on. If not specified, it will
 throw if multiple devices are connected.
 
+### option: Android.launchServer.host
+* since: v1.44.2
```
mxschmitt, line 206: I 
```suggestion
* since: v1.45
```

docs/src/api/class-android.md
```diff
@@ -202,6 +202,12 @@ Prevents automatic playwright driver installation on attach. Assumes that the dr
...
+- `host` <[string]>
+
+Host to use for the web socket. Defaults to localhost that accessible locally.
```
mxschmitt, line 209:
```suggestion
Host to use for the web socket. Defaults to `localhost`.
```

docs/src/api/class-browsertype.md
```diff
@@ -380,6 +380,12 @@ const { chromium } = require('playwright');  // Or 'webkit' or 'firefox'.
 ### option: BrowserType.launchServer.logger = %%-browser-option-logger-%%
 * since: v1.8
 
+### option: BrowserType.launchServer.host
+* since: v1.44.2
```
mxschmitt, line 384:
```suggestion
* since: v1.45
```

docs/src/api/class-browsertype.md
```diff
@@ -380,6 +380,12 @@ const { chromium } = require('playwright');  // Or 'webkit' or 'firefox'.
...
+- `host` <[string]>
+
+Host to use for the web socket. Defaults to localhost that accessible locally.
```
mxschmitt, line 387:
```suggestion
Host to use for the web socket. Defaults to localhost that accessible locally.
```
```suggestion
Host to use for the web socket. Defaults to `localhost`.
```

tests/library/browsertype-launch-server.spec.ts
```diff
@@ -26,6 +26,13 @@ it.describe('launch server', () => {
     await browserServer.close();
   });
 
+  it('should work with host', async ({ browserType }) => {
+    const host = '0.0.0.0';
```
mxschmitt, line 30: Add a test for Android.launchServer as well? Should be very similar.

mxschmitt (changes requested):
LGTM % comments

cavivie:
@mxschmitt Is this dirty error caused by me pushing twice quickly?

mxschmitt:
> @mxschmitt Is this dirty error caused by me pushing twice quickly?

You need to run npm run build, it will then update some generated files.

cavivie:
BTW, I found that although the ws server pointed out that the ws url was `localhost`, it actually listened at 0.0.0.0. This is because the http server listen uses net.Server listen. When the host is omitted in this method, the unspecified address will be used.

https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/wsServer.ts#L57-L83

https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback

To fix it, we need to do either of these things:
1. We may need to open a new bug issue for this and fix it: set hostname to localhost instead of undefined, because undefined means that the net server listens at the unspecified address ipv4(0.0.0.0)/ipv6(::). 
2. We only changed the document description to unspecified, and changed the hostname in the ws url to undefined value to 0.0.0.0 but not`localhost`, [line to change](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/wsServer.ts#L80).

packages/playwright-core/src/utils/wsServer.ts
```diff
@@ -77,7 +77,7 @@ export class WSServer {
           reject(new Error('Could not bind server socket'));
           return;
         }
-        const wsEndpoint = typeof address === 'string' ? `${address}${path}` : `ws://${hostname || 'localhost'}:${address.port}${path}`;
+        const wsEndpoint = typeof address === 'string' ? `${address}${path}` : `ws://${hostname || address.address}:${address.port}${path}`;
```
cavivie, line 80: For compatibility, I only changed the `host` option document to unspecified and specified the output ws hostname to be the address output by the server.

cavivie:
> 2\. We only change the document description to unspecified, ~and chang the hostname in the ws url to undefined value to 0.0.0.0 but not`localhost`, [line to change](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/wsServer.ts#L80)~.

Finally, I followed this to keep compatibility, what do you think?

docs/src/api/class-android.md
```diff
@@ -202,6 +202,12 @@ Prevents automatic playwright driver installation on attach. Assumes that the dr
 Optional device serial number to launch the browser on. If not specified, it will
 throw if multiple devices are connected.
 
+### option: Android.launchServer.host
+* since: v1.5
+- `host` <[string]>
+
+Host to use for the web socket. Defaults to `unspecified`.
```
pavelfeldman, line 209: Your change looks good. Please remove the mentions of 'unspecified', we imply that for the options, they are optional.

pavelfeldman (approved):

PR #31005, opened 2024-05-24
Fix docs incorrectly listing default for screenshot scaling

AlfieJones:
The docs currently say the screenshot scaling default is 'css' when it's 'device'


pavelfeldman:
> The docs currently say the screenshot scaling default is 'css' when it's 'device'

It is `css` in `toHaveScreenshot`

PR #31011, opened 2024-05-24
docs: link to input page instead of locators in testing-library migration guide

MatanBobi:
This PR changes the actions link from the [locators](https://playwright.dev/docs/api/class-locator) page to the [actions](https://playwright.dev/docs/input) page which seems better to me logically.

### DIFF
--- a/docs/src/testing-library-js.md
+++ b/docs/src/testing-library-js.md
@@ -18,7 +18,7 @@ If you use DOM Testing Library in the browser (for example, you bundle end-to-en
 | [screen](https://testing-library.com/docs/queries/about#screen) | [page](./api/class-page) and [component](./api/class-locator) |
 | [queries](https://testing-library.com/docs/queries/about) | [locators](./locators) |
 | [async helpers](https://testing-library.com/docs/dom-testing-library/api-async) | [assertions](./test-assertions) |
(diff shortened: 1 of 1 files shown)


docs/src/testing-library-js.md
```diff
@@ -18,7 +18,7 @@ If you use DOM Testing Library in the browser (for example, you bundle end-to-en
 | [screen](https://testing-library.com/docs/queries/about#screen) | [page](./api/class-page) and [component](./api/class-locator) |
 | [queries](https://testing-library.com/docs/queries/about) | [locators](./locators) |
 | [async helpers](https://testing-library.com/docs/dom-testing-library/api-async) | [assertions](./test-assertions) |
-| [user events](https://testing-library.com/docs/user-event/intro) | [actions](./api/class-locator) |
+| [user events](https://testing-library.com/docs/user-event/intro) | [actions](./input) |
```
pavelfeldman, line 21: Looks like class-locator is the right place to point to

MatanBobi: Since the locators page doesn't contain an explanation about actions, I was under the impression that the actions page is better:
This is the locators:
image

And this is the actions:
image

But, this is only my understanding. I'll close this one and feel free to re-open if you think my comment is worth it :)
Thanks!

Issue #30305, opened 2024-04-09
[Bug]: My site is not able to login on Webkit although Chromium & Firefox are working smoothly


danghung-30385:
### Version

Playwright 1.43

### Steps to reproduce

- Login function is run on 3 browsers (Chromium, Firefox, and Webkit)

### Expected behavior

 login function can works smoothly on Webkit

### Actual behavior

1. The login function is FAILED only on Webkit
2. The login function is PASSED on Chomium & Firefox

### Additional context

[trace (24).zip](https://github.com/microsoft/playwright/files/14916883/trace.24.zip)
[f10688cca915ec36251a6cea0258938d26be9f0f.webm](https://github.com/microsoft/playwright/assets/56102603/33c04c2f-e388-4453-883a-994b9066a3fd)

### Environment

1. Playwright Version: 1.43
2. Operating System: Ubuntu 22.04 LTS (Jammy Jellyfish).

mxschmitt:
When I'm running it in WebKit, I'm getting:

```
Info: The current domain is not authorized for OAuth operations. This will prevent signInWithPopup, signInWithRedirect, linkWithPopup and linkWithRedirect from working. Add your domain (development-webapp-common-flow-auth.eks.csp.scrum-dev.com) to the OAuth redirect domains list in the Firebase console -> Authentication -> Settings -> Authorized domains tab.
```

Could you confirm that your domain is listed there?

danghung-30385:
Hi mxschmitt,
Thanks for your reply quickly.

I would like add more info, This test case run well with Webkit on Windows and only got failed on CI & Ubuntu. I tried with many approaches such as: clear cookie, route... But the script always failed on CI & Ubuntu

ReactiveWarrior:
Webkit Mobile failing on Linux Docker container is well known problem:
1. https://github.com/microsoft/playwright/issues/10183
2. https://github.com/microsoft/playwright/issues/11101
3. https://github.com/microsoft/playwright/issues/23573

Recently, I stumbled upon this as well, my test was failing only on Mobile Safari and only on GitHub CI where Linux Docker container is used. I couldn't solve this issue as others also didn't. Unfortunately I had to disable the test for this browser.

mxschmitt:
Was able to find out the root-cause, its related to cookie parsing. Created a minimal reproduction here: https://github.com/microsoft/playwright/pull/30323

[...]

PR #31012, opened 2024-05-24
fix(webkit): make cookie SameSite=Lax by default

yury-s:
“Prevent cross-site tracking” [feature is on](https://support.apple.com/guide/safari/prevent-cross-site-tracking-sfri40732/mac) by default in Safari. It prevents SameSite=None cookie behavior, making such cookies Lax. With this PR the following behavior changes:

* The following cookies become SameSite=Lax when added via addCookies call:
  - `SameSite` is not specified
  - `SameSite=None` is specified

* Cookies added via `document.cookies='foo=bar'` are now reported as `Lax`.

Reference https://github.com/microsoft/playwright/issues/30305
Reference https://github.com/microsoft/playwright-browsers/issues/795

### DIFF
--- a/packages/playwright-core/src/server/webkit/wkBrowser.ts
+++ b/packages/playwright-core/src/server/webkit/wkBrowser.ts
@@ -256,7 +256,11 @@ export class WKBrowserContext extends BrowserContext {
   async doGetCookies(urls: string[]): Promise<channels.NetworkCookie[]> {
     const { cookies } = await this._browser._browserSession.send('Playwright.getAllCookies', { browserContextId: this._browserContextId });
     return network.filterCookies(cookies.map((c: channels.NetworkCookie) => {
-      const copy: any = { ... c };
+      const copy: any = { ...c };
+      // "Prevent cross-site tracking." is on by default in Safary, which essentially
+      // disables SameSite=None and such cookies become Lax.
+      if (copy.sameSite === 'None')
+        copy.sameSite = 'Lax';
       copy.expires = c.expires === -1 ? -1 : c.expires / 1000;
       delete copy.session;
       return copy as channels.NetworkCookie;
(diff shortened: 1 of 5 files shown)


yury-s:
This is based on https://github.com/mxschmitt/playwright/commit/0b6e317a503a2dda68f0a34bb1691fccb2855297

packages/playwright-core/src/server/webkit/wkBrowser.ts
```diff
@@ -256,7 +256,11 @@ export class WKBrowserContext extends BrowserContext {
   async doGetCookies(urls: string[]): Promise<channels.NetworkCookie[]> {
     const { cookies } = await this._browser._browserSession.send('Playwright.getAllCookies', { browserContextId: this._browserContextId });
     return network.filterCookies(cookies.map((c: channels.NetworkCookie) => {
-      const copy: any = { ... c };
+      const copy: any = { ...c };
+      // "Prevent cross-site tracking." is on by default in Safary, which essentially
+      // disables SameSite=None and such cookies become Lax.
+      if (copy.sameSite === 'None')
```
pavelfeldman, line 262: I think we should return the cookie as in Set-Cookie header and as in Network stack.

packages/playwright-core/src/server/webkit/wkBrowser.ts
```diff
@@ -267,7 +271,10 @@ export class WKBrowserContext extends BrowserContext {
     const cc = network.rewriteCookies(cookies).map(c => ({
       ...c,
       session: c.expires === -1 || c.expires === undefined,
-      expires: c.expires && c.expires !== -1 ? c.expires * 1000 : c.expires
+      expires: c.expires && c.expires !== -1 ? c.expires * 1000 : c.expires,
+      // With "Prevent cross-site tracking." on by default in Safary, SameSite=None cookies
+      // behave as Lax.
+      sameSite: (!c.sameSite || c.sameSite === 'None') ? 'Lax' : c.sameSite,
```
pavelfeldman, line 277: same here, cookie jar should contain raw value, should write None if it was specified here just fine.
Actual behavior of SameSite=None will be determined at runtime, depending on the tracking flag.

mxschmitt (approved):

PR #31026, opened 2024-05-27
fix(launchServer): return precise wsAddress

mxschmitt:
Follow-up to https://github.com/microsoft/playwright/pull/30999 following a similar pattern like in https://github.com/microsoft/playwright/pull/30853.

### DIFF
--- a/packages/playwright-core/src/utils/wsServer.ts
+++ b/packages/playwright-core/src/utils/wsServer.ts
@@ -77,8 +77,12 @@ export class WSServer {
           reject(new Error('Could not bind server socket'));
           return;
         }
-        const wsEndpoint = typeof address === 'string' ? `${address}${path}` : `ws://${hostname || 'localhost'}:${address.port}${path}`;
-        resolve(wsEndpoint);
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/utils/wsServer.ts
```diff
@@ -77,8 +77,12 @@ export class WSServer {
           reject(new Error('Could not bind server socket'));
           return;
         }
-        const wsEndpoint = typeof address === 'string' ? `${address}${path}` : `ws://${hostname || 'localhost'}:${address.port}${path}`;
-        resolve(wsEndpoint);
+        if (typeof address === 'string') {
+          resolve(`${address}${path}`);
+          return;
+        }
+        const resolvedHost = address.family === 'IPv4' ? address.address : `[${address.address}]`;
```
dgozman, line 84: - Shouldn't this be `hostname` when it is defined?
- I am afraid this could accidentally break someone expecting `localhost` in the address and getting `127.0.0.1` or `[::1]` instead. Should we wait for a bug report?

PR #31033, opened 2024-05-27
fix(trace-viewer): fix snapshot.html

ruifigueira:
Actual path to get trace contexts is /contexts, not /context


dgozman (approved):
Thank you for the PR, great catch! The fix looks good, but the bots were unhappy. I've restarted the failed ones, because it looked more like infra failure. If the tests pass, I'll merge this in.

PR #31053, opened 2024-05-28
Remove uneeded sentence in docs

devxan:
"To learn more about generating tests check out or detailed guide on Codegen." 

Should it be 'our'? And why does it link back to itself? Should be removed.


dgozman (approved):
Great catch, merging in! Thank you for the PR.

PR #31055, opened 2024-05-28
feat(ct): react optional declaration merging hooks config

sand4rt:
- Partial fix for: https://github.com/microsoft/playwright/issues/19933
- @pavelfeldman this is a split from: https://github.com/microsoft/playwright/pull/30455#issuecomment-2130464909
- Will follow up with the other frameworks if this one gets through

### DIFF
--- a/packages/playwright-ct-react17/hooks.d.ts
+++ b/packages/playwright-ct-react17/hooks.d.ts
@@ -14,11 +14,9 @@
  * limitations under the License.
  */
 
-export interface RegisterHooksConfig {}
-
-export declare function beforeMount<HooksConfig = RegisterHooksConfig>(
+export declare function beforeMount<HooksConfig>(
   callback: (params: { hooksConfig?: HooksConfig; App: () => JSX.Element }) => Promise<void | JSX.Element>
 ): void;
-export declare function afterMount<HooksConfig = RegisterHooksConfig>(
+export declare function afterMount<HooksConfig>(
   callback: (params: { hooksConfig?: HooksConfig }) => Promise<void>
 ): void;
--- a/packages/playwright-ct-react17/hooks.d.ts
+++ b/packages/playwright-ct-react17/hooks.d.ts
@@ -14,9 +14,11 @@
  * limitations under the License.
  */
 
-export declare function beforeMount<HooksConfig>(
+export interface RegisterHooksConfig {}
+
+export declare function beforeMount<HooksConfig = RegisterHooksConfig>(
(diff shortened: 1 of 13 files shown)


packages/playwright-ct-react17/hooks.d.ts
```diff
@@ -14,9 +14,11 @@
  * limitations under the License.
  */
 
-export declare function beforeMount<HooksConfig>(
+export interface RegisterHooksConfig {}
```
sand4rt, line 17: Note: I have not changed the `react-router.spec.tsx` for React17 to show that declaration merging is completely optional

tests/components/ct-react-vite/playwright/index.tsx
```diff
@@ -2,18 +2,19 @@ import { beforeMount, afterMount } from '@playwright/experimental-ct-react/hooks
 import { BrowserRouter } from 'react-router-dom';
 import '../src/assets/index.css';
 
-export type HooksConfig = {
-  route?: string;
-  routing?: boolean;
+declare module '@playwright/experimental-ct-react/hooks' {
+  interface RegisterHooksConfig {
```
dgozman, line 6: We prefer to not override types globally in Playwright. Can we help the user in some other way? Perhaps by casting `test` exported from the package like `const test = baseTest as SomeTestDeclaration<SpecificHooksConfig>`? Or by extending with a `mnt` fixture that is `mount<SpecificHooksConfig>`?

sand4rt: > We prefer to not override types globally in Playwright.

Normally i wouldn't use `declare module` in this way, but i think this is a good use case. Especially because the `mount()` is called so many times. The `beforeMount<T>`, `afterMount<T>()` and `mount<T>()` can also still be specified manually for the people who think it's too much magic.

> Can we help the user in some other way?

I don't think there's another way to make the `hooksConfig` in `beforeMount()`, `afterMount()` and `mount()` automatically type-safe with a single type definition right? type-casting `const test` would solve a part of the problem.

dgozman: > I don't think there's another way to make the `hooksConfig` in `beforeMount()`, `afterMount()` and `mount()` automatically type-safe with a single type definition right? type-casting `const test` would solve a part of the problem.

That's right. However, you only write `beforeMount()` and `afterMount()` once, so explicitly typing them is not a big deal. It looks like the main goal is to avoid `mount<T>` every time.

PR #31061, opened 2024-05-29
fix(merge-reports): only change test ids when needed

muhqu:
When merging blob reports test ids are patched to make sure there is no collision when merging reports that might have overlapping test ids. However, even if you were merging reports that had no overlapping ids, all test ids will be modified, which is an undesirable side effect.

This PR only modify test ids when the same test id has already been used in a previous blob report.

----

This change is also part of https://github.com/microsoft/playwright/pull/30962

### DIFF
--- a/packages/playwright/src/reporters/merge.ts
+++ b/packages/playwright/src/reporters/merge.ts
@@ -194,16 +194,25 @@ async function mergeEvents(dir: string, shardReportFiles: string[], stringPool: 
   printStatus(`merging events`);
 
   const reports: ReportData[] = [];
+  const testIdsUsedInOtherBlobs = new Set<string>();
 
   for (let i = 0; i < blobs.length; ++i) {
     // Generate unique salt for each blob.
     const { parsedEvents, metadata, localPath } = blobs[i];
     const eventPatchers = new JsonEventPatchers();
-    eventPatchers.patchers.push(new IdsPatcher(stringPool, metadata.name, String(i)));
+    const testIdsUsedInThisBlob = new Set<string>();
+    eventPatchers.patchers.push(new IdsPatcher(
+        stringPool,
+        metadata.name,
+        String(i),
+        testIdsUsedInThisBlob,
+        testIdsUsedInOtherBlobs,
+    ));
     // Only patch path separators if we are merging reports with explicit config.
     if (rootDirOverride)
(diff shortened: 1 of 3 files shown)


tests/playwright-test/reporter-blob.spec.ts
```diff
@@ -1813,6 +1813,78 @@ test('merge reports without --config preserves path separators', async ({ runInl
...
+      test('test 1', async ({}) => { });
+      test('test 2', async ({}) => { });
+      test('test 3', async ({}) => { });
+    `,
+  };
+  let testIdsFromSingleRun: string[];
+  let testIdsFromShard1: string[];
+  let testIdsFromShard2: string[];
+  let testIdsFromMergedReport: string[];
+  const parseTestIdsFromOutput = (output: string) => output.split('\n').filter(s => s.startsWith('test.id:')).map(s => s.split('test.id:')[1].trim()).sort();
```
dgozman, line 1862: If you change logging to `console.log('%%' + test.id)`, you'll get this for free with `result.outputLines`.

packages/playwright/src/reporters/merge.ts
```diff
@@ -406,7 +427,20 @@ class IdsPatcher {
   }
 
   private _mapTestId(testId: string): string {
-    return this._stringPool.internString(testId + this._salt);
+    const t1 = this._stringPool.internString(testId);
+    if (this._testIdsMap.has(t1))
+      // already mapped
+      return this._testIdsMap.get(t1)!;
+    if (this._testIdsUsedInOtherBlobs.has(t1)) {
```
dgozman, line 434: It seems like you can have a single `globalTestIdSet` that you populate every time you assign a new unique id across blobs? Should be a bit easier than two sets.

dgozman (commented):
Thank you for the PR! Looks good overall, see inline comments.

tests/playwright-test/reporter-blob.spec.ts
muhqu: I first wanted to use that, but `outputLines` was not yet available for `mergeReports()`… I can look into adding it.

muhqu: ok, done that.

dgozman (approved):

dgozman:

PR #31062, opened 2024-05-29
fix(runner): don't write last run info when listing tests

muhqu:
don't write last run info when listing tests.

----

This change is also part of https://github.com/microsoft/playwright/pull/30962


dgozman (approved):

PR #31093, opened 2024-05-30
chore(trace-viewer): preserve column widths after showing resource details

yury-s:
* Column widths are now stored on in the NetworkPanel context, this way they are not reset after selecting an empty range (and changing position of the NetworkGridView in the component tree).
* Column widths values are now preserved if column set changes (e.g. selecting entries from a single context and then from multiple contexts).

### DIFF
--- a/packages/trace-viewer/src/ui/networkTab.tsx
+++ b/packages/trace-viewer/src/ui/networkTab.tsx
@@ -76,6 +76,10 @@ export const NetworkTab: React.FunctionComponent<{
     return { renderedEntries };
   }, [networkModel.resources, networkModel.contextIdMap, sorting, boundaries]);
 
+  const [widths, setWidths] = React.useState<Map<ColumnName, number>>(() => {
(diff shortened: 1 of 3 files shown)


pavelfeldman:
> Column widths values are now preserved if column set changes (e.g. selecting entries from a single context and then from multiple contexts).

You can't adjust column set dynamically, it would be a bad UX.

packages/trace-viewer/src/ui/networkTab.tsx
```diff
@@ -87,7 +91,8 @@ export const NetworkTab: React.FunctionComponent<{
     onHighlighted={item => onEntryHovered(item?.resource)}
     columns={visibleColumns(!!selectedEntry, renderedEntries)}
     columnTitle={columnTitle}
-    columnWidth={columnWidth}
+    columnWidths={widths}
```
pavelfeldman, line 94: I would instead use key={} to split grids with different column sets. key can be a join of column names.

pavelfeldman (approved):

PR #31105, opened 2024-05-31
feat: increase length of printed html tags

### LINKED ISSUE #30977: [Feature]: print the entire html tag in debug when its intercepting click events
### 🚀 Feature Request

When performing the actionability of a click on an element and other elements intercepts pointer events, the latter should be printed with all his attributes.

### Example

```
  - retrying click action, attempt #26
  -   waiting 500ms
  -   waiting for element to be visible, enabled and stable
  -   element is visible, enabled and stable
  -   scrolling into view if needed
  -   done scrolling
  -   <div data-v-8e10e111="" data-v-b9bc3028="" class="qui…>…</div> from <div data-v-b9bc3028="" data-v-91a0e5df="">…</div> subtree intercepts pointer events
```
The feature request is about those `...`

### Motivation

I totally get that we can't have the whole html outputted (the second and the third `...`), but would it be possible to have the whole html tag ? In my example, it would mean having what is after `class="qui`.
As you can see in my example, the first attributes might not always lead us to the actual element in our code that intercept click, making debugging really difficult

Joe-Hendley:
resolves https://github.com/microsoft/playwright/issues/30977 by increasing the printed length of HTML tags from 50 -> 500 as suggested by @dgozman


dgozman (approved):
Thank you for the PR! Merging in.

Issue #30981, opened 2024-05-23
[Component Testing]: Service workers support


yannbf:
### 🚀 Feature Request

Seems that Playwright CT blocks service workers and is non-configurable: https://github.com/microsoft/playwright/blob/main/packages/playwright-ct-core/src/mount.ts#L43

If there was a way to allow service workers, or maybe specify which service workers could be allowed, it would be fantastic.

### Example

Check the logs to see the `Service worker registration blocked by Playwright` failure:

![image](https://github.com/microsoft/playwright/assets/1671563/b7108fbf-eeb4-423a-afab-0c52e819c70b)

And if I switch from `block` to `allow` in [this line](https://github.com/microsoft/playwright/blob/main/packages/playwright-ct-core/src/mount.ts#L43) in `/node_modules/@playwright/experimental-ct-core/lib/mount.js`, things work as expected:

![image](https://github.com/microsoft/playwright/assets/1671563/c347aa8c-82e5-4ab4-a8f5-efaf9236703a)

### Motivation

Using mocking tools such as [MSW](https://mswjs.io/) has become increasingly more common, and tools like MSW are used and shared across different environments e.g. Node with Vitest, Browser with Storybook, as well as Browser in the user's application. Ideally, users should be able to reuse their mocking setup in Playwright CT tests as well.

Storybook released [a new API](https://storybook.js.org/blog/portable-stories-for-playwright-ct/) that helps adopt Playwright CT while reusing their Storybook stories. However for set ups that use MSW, the components will render correctly in Storybook, but the tests will fail in Playwright with a message mentioning that service workers are blocked.

### Extra context

Seems like service workers were configurable from this PR:
https://github.com/microsoft/playwright/pull/14714

However later on this PR changed the code so that it's always blocked:
https://github.com/microsoft/playwright/pull/15477

Unfortunately there isn't much context in the latter PR, it would be amazing to know whether there are limitations that won't allow the usage of service workers or that maybe there is a chance to just make it configurable back again.

pavelfeldman:
Discussion:
- retrofit MSW apis on top of route
- allow MSW as a fixture for existing mocks reuse.
- do that as an experiment inside CT family

kettanaito:
@pavelfeldman, I'd love to hear more about the context of intentionally disabling service workers in the Component Testing feature of Playwright. 

Removing this single line sounds like a fix to the problem:

[...]

PR #31154, opened 2024-06-04
feat(ct): expose `msw` fixture that accepts msw-style request handlers

dgozman:
Implemented on top of `context.route()`.

### DIFF
--- a/docs/src/test-components-js.md
+++ b/docs/src/test-components-js.md
@@ -721,6 +721,45 @@ test('update', async ({ mount }) => {
 
 </Tabs>
 
+### Handling network requests
+
+Playwright provides [`method: Page.route`] method to route your network requests. See the [network mocking guide](./mock.md) for more details.
+
+### MSW support
+
+If you are using the [MSW library](https://mswjs.io/) to handle network requests during development or testing, try the **experimental fixture** `msw` and give us feedback. With this fixture, you can reuse request handlers between development and tests.
+
+```ts
+import { handlers } from '@src/mocks/handlers';
+
+test.beforeEach(async ({ msw }) => {
+  // install common handlers before each test
+  await msw.use(...handlers);
+})
+
+test('example test', async ({ mount }) => {
+  // test as usual, your handlers are active
+  // ...
+});
+```
+
+You can also introduce test-specific handlers.
+
+```ts
+import { http, HttpResponse } from 'msw';
+
+test('example test', async ({ mount, msw }) => {
+  await msw.use(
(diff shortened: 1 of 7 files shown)


docs/src/test-components-js.md
```diff
@@ -721,6 +721,45 @@ test('update', async ({ mount }) => {
 
 </Tabs>
 
+### Handling network requests
+
+Playwright provides [`method: Page.route`] method to route your network requests. See the [network mocking guide](./mock.md) for more details.
+
+### MSW support
+
+If you are using the [MSW library](https://mswjs.io/) to handle network requests during development or testing, try the `msw` fixture that Playwright provides. It supports the same API so that you can reuse your handlers in the test.
```
pavelfeldman, line 730: Let's say that we are experimenting with it?

pavelfeldman (approved):

packages/playwright-ct-core/src/msw.ts
```diff
@@ -0,0 +1,164 @@
...
+  const requestId = String(++lastRequestId);
+  const resolutionContext = { baseUrl };
+  for (const handler of handlers) {
+    const result = await handler.run({ request, requestId, resolutionContext });
+    if (result?.response)
+      return result.response;
+  }
+}
+
+async function globalFetch(...args: Parameters<typeof globalThis.fetch>) {
```
kettanaito, line 42: If you add `msw` as a dependency, it exposes the [`getResponse`](https://github.com/mswjs/msw/blob/bed402cdd9b79ef3084b3195cdcce0d83c7e2cfc/src/core/getResponse.ts#L12) function that you can reuse here. Less logic duplication.

packages/playwright-ct-core/src/msw.ts
```diff
@@ -0,0 +1,164 @@
...
+ * limitations under the License.
+ */
+
+import type * as playwright from 'playwright/test';
+
+interface RequestHandler {
+  run(args: { request: Request, resolutionContext: { baseUrl?: string }, requestId: string }): Promise<{ response: Response } | undefined>;
+}
+
+export interface MSW {
```
kettanaito, line 23: Would be nice to actually `import { RequestHandler } from 'msw'`.

kettanaito:
I have some concerns about this. Please see https://github.com/microsoft/playwright/issues/30981. 

TL;DR I'd rather people used the _same_ MSW experience than a compatible API. MSW works great in any browser, including browsers automated by Playwright. I see no reason to diverge the behavior, losing a significant chunk of MSW's benefits along the way.

Issue #6854, opened 2021-06-02
[Feature] Folder uploads


reidbuzby:
I am trying to upload a folder using playwright. I have been able to successfully use 
```javascript
page.on('filechooser', async fileChooser => {
  await fileChooser.setFiles('path/path_to_file.txt');
});
```
To upload _files_, however I want to upload a _folder_ instead. Something like:
```javascript
page.on('filechooser', async fileChooser => {
  await fileChooser.setFiles('path/path_to_folder/');
});
```
I have not found a way to do this after looking through the docs and searching Google, so I am guessing it is not currently supported. If I do try running that exact code above on a folder path I get the error:
```
fileChooser.setFiles: elementHandle.setInputFiles: EISDIR: illegal operation on a directory, read
```

I am hoping for a way to automate the process that would happen when clicking on `<input type=folder>`. Maybe the page could emit a different event 'folderchooser'.

mxschmitt:
If you want to upload all the files from a directory, you can do the following:

```js
const fs = require('fs');
const path = require('path');

const testFolder = path.join(__dirname);

console.log(fs.readdirSync(testFolder).map(file => !fs.statSync(file).isDirectory() ? path.join(testFolder, file) : null).filter(Boolean));
```

This will read all the files and folders in the directory, maps over them to check if its a file and if not filters it out via the Boolean filter.

pavelfeldman:
Let's collect user votes for this one!

KapustinMaxim:
This feature will be very useful!

maximkapusttin:
Add please

robertyoung1993:
its very useful

kentz007:
Let's collect user votes for this one!

eliw00d:
> Let's collect user votes for this one!

What number are you looking to reach for user votes before it's considered?

f-o-w-l:
This would be very helpful. Right now, when I pass multiple relative file paths to `setInputFiles`, their paths are not being passed to my onChange listener, and the `files` attribute returns `null`.

Example test:

Folder input component:
```javascript
  ...
  const onFolderChange = () => {
    const files = ref.current ? ref.current.files : null
    if (files && files.length > 0) {
      for (let i = 0; i < files.length; i += 1) {
        console.log(`name: "${files[i].name}"`)
        console.log(`path: "${files[i].path}"`)
        console.log(`webkitRelativePath: "${files[i].webkitRelativePath}"`)
      }
      // Doing stuff with file path...
    }
[...]

PR #31165, opened 2024-06-05
feat: allow folder uploads

mxschmitt:
Motivation: This adds support for `input[webkitdirectory]` folder uploads. For local mode, we just pass the directory to the browser, for "remote mode" we stream the files to the server and keeping the same parent directory name. This one is also used inside the browser in the `webkitRelativePath` field in the `File` type.

In Chromium headless, there is [a bug](https://issues.chromium.org/issues/345393164), that folder uploads won't consider nested files / is this a blocker for the feature to land?
~~Firefox + WebKit implementation have been landed and look like they are working. Requires rolls, hence the tests are still not passing.~~

Feature request: https://github.com/microsoft/playwright/issues/6854


docs/src/api/class-locator.md
```diff
@@ -2164,6 +2164,7 @@ When all steps combined have not finished during the specified [`option: timeout
 * since: v1.14
 
 Upload file or multiple files into `<input type=file>`.
+In order to upload a directory (`[webkitdirectory]`) pass the path to the directory.
```
dgozman, line 2167:
```suggestion
For inputs with a `[webkitdirectory]` attribute, only a single directory path is supported.
```

packages/playwright-core/src/server/dom.ts
```diff
@@ -633,7 +633,7 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
         return;
       if (element.tagName !== 'INPUT')
         throw injected.createStacklessError('Node is not an HTMLInputElement');
-      if (multiple && !(element as HTMLInputElement).multiple)
+      if (multiple && !(element as HTMLInputElement).multiple && !(element as HTMLInputElement).webkitdirectory)
```
dgozman, line 636: Let's throw a similar error when you pass multiple files into directory inputs.

packages/playwright-core/src/server/dom.ts
```diff
@@ -647,7 +647,12 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
         await Promise.all(localPaths.map(localPath => (
           fs.promises.access(localPath, fs.constants.F_OK)
         )));
+        const isDirectoryUpload = localPaths.length === 1 ? (await fs.promises.stat(localPaths[0])).isDirectory() : false;
+        const waitForChangeEvent = isDirectoryUpload ? this.evaluateInUtility(([_, node]) => new Promise<any>(fulfil => {
```
dgozman, line 651: typo: `fulfil` -> `fulfill`

packages/playwright-core/src/server/dom.ts
```diff
@@ -647,7 +647,12 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
         await Promise.all(localPaths.map(localPath => (
           fs.promises.access(localPath, fs.constants.F_OK)
         )));
+        const isDirectoryUpload = localPaths.length === 1 ? (await fs.promises.stat(localPaths[0])).isDirectory() : false;
```
dgozman, line 650: Perhaps only check this when `webkitdirectory` attribute is present?

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -262,16 +262,42 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
   if (items.some(item => typeof item === 'string')) {
     if (!items.every(item => typeof item === 'string'))
       throw new Error('File paths cannot be mixed with buffers');
+    const directoryCount = (await Promise.all(items.map(async item => (await fs.promises.stat(item as string)).isDirectory()))).filter(Boolean).length;
+    if (directoryCount > 1)
+      throw new Error('Only one directory can be uploaded at a time');
```
dgozman, line 267: Let's be more strict - either a single directory, or a list of files, but not files+directory.

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -262,16 +262,42 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
...
-        const lastModifiedMs = (await fs.promises.stat(item)).mtimeMs;
-        const { writableStream: stream } = await context._wrapApiCall(() => context._channel.createTempFile({ name: path.basename(item), lastModifiedMs }), true);
-        const writable = WritableStream.from(stream);
-        await pipelineAsync(fs.createReadStream(item), writable.stream());
-        return stream;
-      }));
-      return { streams };
+      let streams: channels.WritableStreamChannel[] | undefined;
+      let localPaths: string[] | undefined;
+      for (const item of items as string[]) {
```
dgozman, line 272: I think we should keep `Promise.all()` and extract a method here. It is getting complicated.

packages/protocol/src/protocol.yml
```diff
@@ -1191,6 +1191,22 @@ BrowserContext:
       returns:
         writableStream: WritableStream
 
+    createTempDirectory:
```
dgozman, line 1194: Let's merge this and previous methods and call it `createTempFiles`, giving an optional `rootDirName` parameter?

packages/protocol/src/protocol.yml
```diff
@@ -1191,6 +1191,22 @@ BrowserContext:
...
+        root: string
+        items:
+          type: array
+          items:
+            type: object
+            properties:
+              name: string
+              lastModifiedMs: number?
+      returns:
+        dir: string
```
dgozman, line 1205: Why does a client need to know a local path on the server?

packages/playwright-core/src/server/dom.ts
mxschmitt: I'm throwing on the client side instead, or would you prefer both for safety?

packages/playwright-core/src/server/dom.ts
mxschmitt: While file uploads to `input[webkitdirectory]` elements were possible before, this would mean we wait also for file uploads, while my change only waits for folder uploads. I prefer slightly only waiting for actual folder uploads, wdyt?

packages/protocol/src/protocol.yml
mxschmitt: it's a bit of a hack. We stream 3 files to the server (in the directory). The files itself could live under on the server:

- /tmp-random/my-dir/sub-dir/1.txt
- /tmp-random/my-dir/2.txt
- /tmp-random/my-dir/sub-dir/whatever/3.txt

So the server to this moment does not know what the root directory is. Instead we then send 3 files to the server over our protocol, get `/tmp-random/my-dir` back and pass it as a local-path to the server.

packages/playwright-core/src/client/elementHandle.ts
mxschmitt: I simplified it a little bit but now there are a bunch of `if isDirectory` checks. I guess no way around a method. What about moving the map callback into an own method accepting the item, streams and localPath?

packages/protocol/src/protocol.yml
dgozman: Yeah, we should not pass `localPaths` when talking to the server. Let's make server's temp files streams know what they need to.

packages/playwright-core/src/server/dom.ts
dgozman: Yeah, both sides sounds easier to understand.

packages/playwright-core/src/server/dom.ts
dgozman: I meant avoiding `fs.promises.stat` when `[webkitdirectory]` is not present. From what I see, we now always do this check unconditionally when passed a single file, which is the most common case.

mxschmitt: If you want to throw https://github.com/microsoft/playwright/pull/31165#discussion_r1630015449 from both sides, this means we'd need to call stat from the client and server. So there is no way around that anyways?

packages/playwright-core/src/client/elementHandle.ts
mxschmitt: Reverted to how it was before.

packages/playwright-core/src/server/dispatchers/browserContextDispatcher.ts
```diff
@@ -178,13 +178,19 @@ export class BrowserContextDispatcher extends Dispatcher<BrowserContext, channel
...
     const tmpDir = path.join(dir, 'upload-' + createGuid());
-    await fs.promises.mkdir(tmpDir);
+    const tempDirWithRootName = params.rootDirName ? path.join(tmpDir, path.basename(params.rootDirName)) : tmpDir;
+    await fs.promises.mkdir(tempDirWithRootName, { recursive: true });
     this._context._tempDirs.push(tmpDir);
-    const file = fs.createWriteStream(path.join(tmpDir, params.name));
-    return { writableStream: new WritableStreamDispatcher(this, file, params.lastModifiedMs) };
+    return {
+      writableStreams: await Promise.all(params.items.map(async item => {
+        await fs.promises.mkdir(path.dirname(path.join(tempDirWithRootName, item.name)), { recursive: true });
```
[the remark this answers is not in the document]
mxschmitt: `item.name` can be in a sub-directory, hence we need it there.

packages/playwright-core/src/server/dom.ts
```diff
@@ -625,29 +626,37 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
...
       progress.throwIfAborted();  // Avoid action that has side-effects.
-      if (localPaths) {
-        await Promise.all(localPaths.map(localPath => (
+      if (localPaths || localDirectory) {
+        const localPathsOrDirectory = localDirectory ? [localDirectory] : localPaths!;
+        await Promise.all((localPathsOrDirectory).map(localPath => (
           fs.promises.access(localPath, fs.constants.F_OK)
         )));
-        await this._page._delegate.setInputFilePaths(retargeted, localPaths);
+        const waitForInputEvent = localDirectory ? this.evaluate(node => new Promise<any>(fulfill => {
```
dgozman, line 655: Let's add a comment that we need this because browsers travers the directory asynchronously.

tests/electron/playwright.config.ts
```diff
@@ -54,6 +54,7 @@ config.projects.push({
   name: 'electron-api',
   use: {
     browserName: 'chromium',
+    headless: false,
```
dgozman, line 57: Seems unrelated.

packages/protocol/src/protocol.yml
```diff
@@ -2187,6 +2196,7 @@ Frame:
               name: string
               mimeType: string?
               buffer: binary
+        localDirectory: string?
```
dgozman, line 2199: It seems like this should also include `directoryStream: WritableStream?` for remote scenarios, to make things obvious.

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -262,18 +262,37 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
   if (items.some(item => typeof item === 'string')) {
     if (!items.every(item => typeof item === 'string'))
       throw new Error('File paths cannot be mixed with buffers');
+    const itemFileTypes = (await Promise.all((items as string[]).map(async item => (await fs.promises.stat(item)).isDirectory() ? 'directory' : 'file')));
```
dgozman, line 265: I think it would be easier to split `items` into `files` and `directory` variables here, and not use `items` in the code below. WDYT?

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -262,18 +262,37 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
...
-      const streams: channels.WritableStreamChannel[] = await Promise.all((items as string[]).map(async item => {
-        const lastModifiedMs = (await fs.promises.stat(item)).mtimeMs;
-        const { writableStream: stream } = await context._wrapApiCall(() => context._channel.createTempFile({ name: path.basename(item), lastModifiedMs }), true);
-        const writable = WritableStream.from(stream);
-        await pipelineAsync(fs.createReadStream(item), writable.stream());
-        return stream;
-      }));
+      const streams = (await Promise.all((items).map(async item => {
+        const isDirectory = (await fs.promises.stat(item as string)).isDirectory();
+        const files = isDirectory ? (await fs.promises.readdir(item as string, { withFileTypes: true, recursive: true })).filter(f => f.isFile()).map(f => path.join(f.path, f.name)) : [item];
```
dgozman, line 272: This line and this whole method beg for more spacing and extra variables, or for an extracted method. It is pretty hard to follow now, not to say it wasn't easy even before 😄

tests/electron/playwright.config.ts
mxschmitt: Before Electron / WebView2 tests were treated as headless, so the `headless` fixture was `true` before and my check for the fixed upstream bug was not working as expected.

packages/playwright-core/src/server/fileUploadUtils.ts
```diff
@@ -30,14 +30,17 @@ async function filesExceedUploadLimit(files: string[]) {
...
+  const { payloads, streams, directoryStream } = params;
+  let { localPaths, localDirectory } = params;
 
-  if ([payloads, localPaths, streams].filter(Boolean).length !== 1)
+  if ([payloads, localPaths, localDirectory, streams, directoryStream].filter(Boolean).length !== 1)
     throw new Error('Exactly one of payloads, localPaths and streams must be provided');
 
   if (streams)
-    localPaths = streams.map(c => (c as WritableStreamDispatcher).path());
+    localPaths = streams.map(c => (c as WritableStreamDispatcher)).map(s => s.path());
```
dgozman, line 40:
```suggestion
    localPaths = streams.map(c => (c as WritableStreamDispatcher).path());
```

packages/playwright-core/src/server/fileUploadUtils.ts
```diff
@@ -30,14 +30,17 @@ async function filesExceedUploadLimit(files: string[]) {
...
+  if (directoryStream)
+    localDirectory = (directoryStream as WritableStreamDispatcher | undefined)?.path();
```
dgozman, line 42:
```suggestion
    localDirectory = (directoryStream as WritableStreamDispatcher).path();
```

dgozman (approved):

PR #31198, opened 2024-06-07
chore(clock): split wall and monotonic time

pavelfeldman:


packages/playwright-core/src/server/injected/clock.ts
```diff
@@ -278,12 +241,32 @@ export class ClockController {
...
       this._timers.get(timer.id)!.callAt += timer.delay;
     else
       this._timers.delete(timer.id);
-    callFunction(timer.func, timer.args());
+
+    this._duringTick = true;
+    try {
+      if (typeof timer.func !== 'function') {
+        (() => { eval(timer.func); })();
+        return;
```
dgozman, line 256: Missing `postTask` call here?

dgozman (approved):

PR #31215, opened 2024-06-08
feat(trace-viewer): display text attachments in ui mode

### LINKED ISSUE #30976: [Feature]: UI mode - display text attachments as text, not links
### 🚀 Feature Request

I propose an enhancement to the UI mode of Playwright where text attachments (such as logs, plain text files, or text-based output) are displayed directly within the UI rather than as links that need to be clicked to view the content.

### Example

When running tests in UI mode, any text attachments generated by the tests would automatically be displayed in an expandable/collapsible section within the UI. This would allow users to quickly view relevant logs or output without the need to download and open separate files. For instance, if a test generates a log file as part of its output, the content of the log file would be displayed directly below the test result, allowing for quick review and debugging.

Current state
image

Possible solution
image

### Motivation

This feature would improve the usability and efficiency of the Playwright UI mode by reducing the number of steps needed to access important information. Currently, users must click on links to download text attachments, which can be cumbersome and time-consuming, especially when dealing with multiple files or large logs. By displaying text attachments directly within the UI, users can more quickly identify and diagnose issues, leading to faster development and debugging cycles. This improvement aligns with the goal of making Playwright a more user-friendly and powerful tool for end-to-end testing.

alvaromartmart:
Closes https://github.com/microsoft/playwright/issues/30976

Displays text attachment contents in UI mode

Screenshot 2024-06-09 at 00 45 45


packages/trace-viewer/src/ui/attachmentsTab.css
```diff
@@ -45,3 +45,9 @@
   max-width: 80%;
   box-shadow: 0 12px 28px 0 rgba(0,0,0,.2), 0 2px 4px 0 rgba(0,0,0,.1);
 }
+
+.attachment-item .attachment-item__text {
+  display: block;
+  width: 100%;
```
yury-s, line 51: Can you align the size of the text box with the content and make it not resizable?

alvaromartmart: I was actually wondering whether I should use a `textbox` at all to display the attachment since it doesn't make sense to make it editable. I only did it for the easy resizing capabilities, but since you're suggesting to disable it, maybe I'll just switch to a styled `div`, WDYT?

alvaromartmart: @yury-s any objections to changing from `textbox` to `div`? 🙂  otherwise I'll make the change

yury-s: Using `div` would be nice, but you'll need to preserve formatting and escape some characters, no objections otherwise.

alvaromartmart: done [3520c7a](https://github.com/microsoft/playwright/pull/31215/commits/3520c7a86b7ac66e9fdead42021052c80bc7ee70)

dgozman:
@alvaromartmart Thank you for the PR! We discussed this with the team, and came up with a few comments/suggestions:

- Could you please attach a screenshot that illustrates the UI implemented in the latest commit?
- We feel like the proposed UI is not very intuitive and we could bring more clarity and structure into it. Perhaps we can show `no preview available` for other attachments when there is at least a single preview? Otherwise the UI looks confusing to me.
- Looking at the code, multi-line attachments will be rendered as a single line.
- Perhaps we should make this more like network request/response body preview? [See here](https://github.com/microsoft/playwright/blob/main/packages/trace-viewer/src/ui/networkResourceDetails.tsx#L91) for the code pointer.
- In addition to `text/plain`, we can preview many more [textual mime types](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/mimeType.ts#L21-L23).

packages/trace-viewer/src/ui/attachmentsTab.tsx
```diff
@@ -20,9 +20,33 @@ import { ImageDiffView } from '@web/shared/imageDiffView';
 import type { MultiTraceModel } from './modelUtil';
 import { PlaceholderPanel } from './placeholderPanel';
 import type { AfterActionTraceEventAttachment } from '@trace/trace';
+import { isTextualMimeType } from 'playwright-core/lib/utils';
```
alvaromartmart, line 23: @yury-s I can't for the life of me figure out why this import doesn't work, the build fails.
I see the function is exported as part of utils index.ts, and VSCode correctly finds the path. However an error is thrown during the build:
```
"isTextualMimeType" is not exported by "../playwright-core/lib/utils/index.js", imported by "src/ui/attachmentsTab.tsx".
```
I'm not too familiar with the build system used in this repository so I assume I'll need to modify some additional config (tsconfig / vite related), but so far nothing I tried worked

dgozman: To fix this problem:
- move `packages/playwright-core/src/utils/mimeType.ts` to `packages/playwright-core/src/utils/isomorphic/mimeType.ts`
- update existing imports
- import it here as `import { isTextualMimeType } from '@isomorphic/mimeType'`

This is because trace viewer runs in the browser, so it cannot use Node.js-dependent code, but allowed to use isomorphic code.

alvaromartmart: thank you @dgozman, I managed to fix it following your instructions

alvaromartmart:
> @alvaromartmart Thank you for the PR! We discussed this with the team, and came up with a few comments/suggestions:
> 

Thank you for the feedback!

> * Could you please attach a screenshot that illustrates the UI implemented in the latest commit?

Here's a (WIP) screenshot with some annotations (more on them below)

Screenshot 2024-06-24 at 23 10 35

> * We feel like the proposed UI is not very intuitive and we could bring more clarity and structure into it. Perhaps we can show `no preview available` for other attachments when there is at least a single preview? Otherwise the UI looks confusing to me.

Agree on the "no preview available". Additionally, I was wondering whether we should
1. Add some `hr` dividers between attachment to clearly tell them apart
2. Since attachments could be potentially big and the user not always be interested in looking at them, should we 
2.a. render them by default
2.b. use an expansion panel, requiring user to expand them to see the contents
2.c. render by default files below some size threshold, otherwise use expansion panel

> * Looking at the code, multi-line attachments will be rendered as a single line.

They seem to render just fine, do you have a specific edge case in mind?

> * Perhaps we should make this more like network request/response body preview? [See here](https://github.com/microsoft/playwright/blob/main/packages/trace-viewer/src/ui/networkResourceDetails.tsx#L91) for the code pointer.
> * In addition to `text/plain`, we can preview many more [textual mime types](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/mimeType.ts#L21-L23).

The screenshot above is the result of some experimentation using the `CodeMirrorWrapper` as you suggested. However I have some questions if we go this way:

  3. should we infer when possible the `language` based on the mimeType to get the most accurate syntax highlighting?
  4. should we always render as plain text, leveraging the styling of the `CodeMirrorWrapper` but skipping language-specific stuff?
  5. attachments are displayed based on the `isTextualMimeType` utility now, but that utility includes `image/svg` - some users might expect the svg to be rendered instead of its source. Though this is likely an edge case that could be covered as a future iteration 😅

pavelfeldman:
All good questions - the lack of definitive answers to those is the exact reason we did not rush into implementing it. If you have a good idea on reasonable defaults, include those in the PR and we'll be happy to take a look.

alvaromartmart:
Ok so in [0eb57e7](https://github.com/microsoft/playwright/pull/31215/commits/0eb57e7774afc0ecad719b053c9e29dceab568c3) I placed the attachment in an expansion panel. You can see a demo here

https://github.com/microsoft/playwright/assets/18287793/8970dda7-7182-4fd0-b575-fc5dba51a430

This way:
- attachments can still be downloaded as previously by clicking on their name
- attachments are not rendered by default, user needs to expand the panel
  - we could add an opt-in setting to expand attachments by default, but I guess it would be useful to collect user feedback first? Also I don't know if there are already ui mode specific settings
  
As for the `CodeMirrorWrapper` questions I posted [here](https://github.com/microsoft/playwright/pull/31215#issuecomment-2187431541) earlier, and to avoid scope creep, I decided to keep things simple and just render the raw content without trying to infer language / display svg as images. I think that can be done if needed in a future PR, if users ask for it.

Final question: since it's now using an expansion panel, maybe I can get rid of the `<hr>` delimiters, WDYT?

alvaromartmart:
ugh, forgot to update the tests - tomorrow

packages/trace-viewer/src/ui/attachmentsTab.tsx
```diff
@@ -20,9 +20,52 @@ import { ImageDiffView } from '@web/shared/imageDiffView';
...
+type TextAttachmentProps = {
+  attachment: Attachment;
+  label?: string;
+};
+
+const TextAttachment: React.FunctionComponent<TextAttachmentProps> = ({ attachment, label }) => {
+  const [text, setText] = React.useState<string | null>(null);
+
+  React.useEffect(() => {
+    let isMounted = true;
```
pavelfeldman, line 39: Instead of this, I wonder if we should change the `<Expandable>` contract to be

`<Expandable expanded={} body={() => <TextAttachment/>}>`

We don't seem to have any usages of it yet, so easy to change.

pavelfeldman (commented):
Your change looks good. I have a code suggestion inline and a UX suggestion below:

- Expandable title should not click away, the entire title should be a click target for expand, that would make it a better UX
- You can have a download icon next to it to allow downloading, both in expanded and collapsed state

alvaromartmart:
> Your change looks good. I have a code suggestion inline and a UX suggestion below:
> 
> * Expandable title should not click away, the entire title should be a click target for expand, that would make it a better UX
> * You can have a download icon next to it to allow downloading, both in expanded and collapsed state

Done, thanks for the feedback: [700d061](https://github.com/microsoft/playwright/pull/31215/commits/700d06170d2a5d84a989bc5c0738fc539889932d)

Demo:

https://github.com/microsoft/playwright/assets/18287793/fc7a3048-51c6-4400-95c2-a825281e239e

packages/trace-viewer/src/ui/attachmentsTab.tsx
alvaromartmart: > Instead of this, I wonder if we should change the <Expandable> contract to be
>
> `<Expandable expanded={} body={() => <TextAttachment/>}>`

Not a React expert but I'm not sure I really like this change in contract vs. directly "projecting" the body into the `Expandable` component.

> We don't seem to have any usages of it yet, so easy to change.

The `Expandable` was already used [here](https://github.com/alvaromartmart/playwright/blob/700d06170d2a5d84a989bc5c0738fc539889932d/packages/trace-viewer/src/ui/uiModeFiltersView.tsx#L44-L55), though it doesn't rely on the body, it seems it's only used to toggle the state.

All things considered, I can apply the change if you want as it seems quite straightforward one. Just not sure it's worth it. WDYT?

pavelfeldman: It would look better than the manual isMounted management to me!

alvaromartmart:
In [677ecae](https://github.com/microsoft/playwright/pull/31215/commits/677ecae1585cfc05c5b4d37bb9709d2bdc548a85) I slightly simplified & improved the `ExpandableAttachment` and `TextAttachment` implementation so that the attachments are only fetched _once_ and only after the panel is expanded for the first time. This seems more efficient than fetching every time as the contents won't change.

packages/trace-viewer/src/ui/attachmentsTab.tsx
alvaromartmart: Got rid of the `isMounted` as part of https://github.com/microsoft/playwright/pull/31215#issuecomment-2198347800, but I'm not sure if your request here is still relevant.

On a different note, I wonder if at this point I should merge `ExpandableAttachment` and `TextAttachment` into a single component for simplicity. Please LMKWYT

packages/trace-viewer/src/ui/attachmentsTab.tsx
```diff
@@ -20,9 +20,58 @@ import { ImageDiffView } from '@web/shared/imageDiffView';
...
+};
+
+const ExpandableAttachment: React.FunctionComponent<{ attachment: Attachment }> = ({ attachment }) => {
+  const [expanded, setExpanded] = React.useState(false);
+  const [loaded, setLoaded] = React.useState(false);
+  const [attachmentText, setAttachmentText] = React.useState<string | null>(null);
+
+  React.useMemo(() => {
+    if (expanded && !loaded) {
+      fetch(attachmentURL(attachment)).then(response => response.text()).then(text => {
```
pavelfeldman, line 51: Sounds like you'll be fetching the attachment regardless of its isTextualMimeType value. Do you want to only render Expandable when there is a potential preview?

alvaromartmart: you're absolutely right, missed that while refactoring, thanks.

Should be fixed now: [342cace](https://github.com/microsoft/playwright/pull/31215/commits/342cace03d110f37acfb8aa85f9939d1d8fbfda0) - I also took the chance to simplify the component

pavelfeldman (approved):

PR #31224, opened 2024-06-10
feat(recorder): Make recorder inspect tool copy current selector into clipboard

Carlos-AMG:


pavelfeldman:
Please start with filing an issue. I am not sure I like my clipboard overwritten on click!

PR #31270, opened 2024-06-12
Fix select issue in Windows

### LINKED ISSUE #31248: [Bug]: Codegen can't record `selectOption` on Windows Edge
### Version

1.44.1

### Steps to reproduce

1. Run `npx playwright codegen --channel msedge https://html-preview.github.io/?url=https://gist.githubusercontent.com/mtsmfm/0e4d7c72231f3d6ab444d0aa6a878b18/raw/a4642d1b35555746419fffac64b877d76d699cdd/select.html` on PowerShell
2. Select an option

### Expected behavior

All `selectOption` operations should be recorded

### Actual behavior

Nothing is recorded.

https://github.com/microsoft/playwright/assets/1796864/36e40eb6-5ea8-4027-81fc-4b97b9a7500c

### Additional context

If we move mouse quickly to avoid focusing on any elements other than options, we can record the event it seems.
Especially we need to avoid putting the cursor on this space.

image

By using cursor keys it's also recorded.

With Chrome/Chromium, if we move mouse outside of the menu the operation isn't recorded.

### Environment

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
    Memory: 2.01 GB / 31.84 GB
  Binaries:
    Node: 20.14.0 - ~\AppData\Local\fnm_multishells\9052_1718098056641\node.EXE
    npm: 10.7.0 - ~\AppData\Local\fnm_multishells\9052_1718098056641\npm.CMD
  IDEs:
    VSCode: 1.90.0 - C:\Users\fumia\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 5.0.17 - C:\WINDOWS\system32\bash.EXE
```

mtsmfm:
Fix https://github.com/microsoft/playwright/issues/31248

On Windows, it seems `mousemove` event is emitted even while select menu is open.

I think we need to use `_activeModel` here.

I confirmed this fixes the issue for both Edge and Chrome.

Edge

https://github.com/microsoft/playwright/assets/1796864/33f49d7d-b898-429e-a3d5-fc1f356d6088

Chrome

https://github.com/microsoft/playwright/assets/1796864/a002ffc8-2cfb-4452-80df-1f2b0a644c17

### DIFF
--- a/packages/playwright-core/src/server/injected/recorder/recorder.ts
+++ b/packages/playwright-core/src/server/injected/recorder/recorder.ts
@@ -356,7 +356,7 @@ class RecordActionTool implements RecorderTool {
         return;
       this._performAction({
         name: 'select',
-        selector: this._hoveredModel!.selector,
+        selector: this._activeModel!.selector,
         options: [...selectElement.selectedOptions].map(option => option.value),
         signals: []
       });
--- a/tests/library/inspector/cli-codegen-1.spec.ts
+++ b/tests/library/inspector/cli-codegen-1.spec.ts
@@ -557,6 +557,7 @@ await page.Locator("#checkbox").UncheckAsync();`);
 
     const locator = await recorder.hoverOverElement('select');
     expect(locator).toBe(`locator('#age')`);
+    await page.locator('select').click();
 
     const [message, sources] = await Promise.all([
       page.waitForEvent('console', msg => msg.type() !== 'error'),


mtsmfm:
I confirmed it works even on Linux (WSL)

https://github.com/microsoft/playwright/assets/1796864/7c463b30-af1c-45db-8dd3-740d1ea08033

I think the problem here is that how `selectOption` emulates is not the same as users' operation

yury-s (approved):
Please sign the CLA so that we could merge this.

mtsmfm:

PR #31283, opened 2024-06-12
docs: drop macOS 12 from supported systems

yury-s:


mxschmitt (changes requested):
Language port files as well?

mxschmitt (approved):

Issue #28995, opened 2024-01-13
[BUG] Page.screenshot method hangs indefinitely 


leodoi3:
### System info
- Playwright Version: v1.40.0
- Operating System: All
- Browser: Chromium
- Other info: 32GB RAM

### Source code
```
from playwright.sync_api import sync_playwright

print("Starting...")

with sync_playwright() as p:
    try:
        browser = p.chromium.launch(headless=False)
        page = browser.new_page()
        for index in range(10000):
            print("Iteration:", index)
            page.goto('https://google.com')


    except Exception as error:
        if page:
            print("Error: ", {error})
            page.screenshot(path='error.png', timeout=0)
            print("Took a screenshot")

```

- [x] I provided exact source code that allows reproducing the issue locally.

**Link to the GitHub repository with the repro**

**Test file (self-contained)**

**Steps**
- Run the code
- Wait until the for loop crashes at about 2000-3000 iterations 

**Expected**

The for loop runs all the way through and not crash.
The `page.screenshot` runs normally. 

**Actual**
At the 2652 iteration, a Timeout exception gets caught. Chrome gets an Out of Memory error.
The code hangs at the `page.screenshot` line and never completes. 

## Screenshot
![image](https://github.com/microsoft/playwright-python/assets/46114199/4e3e83b1-fa6c-410a-bed7-52b5f1840848)

mxschmitt:
I can repro, hangs for me with:

```ts
await page.goto("chrome://crash").catch(e => console.log(e));
await page.screenshot({ path: `crash.png` })
```

at:

       pw:api waiting for fonts to load... +250ms

Rafiot:
I have the same issue, when I pass `PLAYWRIGHT_CHROMIUM_USE_HEADLESS_NEW=1` (which is very similar to `headless=False`, if I'm not mistaken).

And this issue has probably the same root cause: https://github.com/microsoft/playwright/issues/29968

If you have an idea how to somehow force a timeout on loading the fonts?

LucaLis:
Hi. I'm facing similar issue - sometimes my tests are getting stuck when i'm tryigng to take screenshot. I'm using Playwright 1.42.0. I've investigated logs using DEBUG=pw:* and i can see that right after calling page.screenshot playwright logs 

`pw:api waiting for fonts to load...`

which is followed by a bunch of 
`pw:protocol SEND {"id": XYZ (...) "document.fonts.ready" (...)`

usually these SEND requests recieve immediate response like:

[...]

PR #31295, opened 2024-06-13
fix(screenshotter): only wait for for document.fonts.ready on locator frame / page main frame

mxschmitt:
**Investigation**

~~We use `nonStallingEvaluateInExistingContext` as of today, which does `eval()` inside (from our utilityScript) which breaks for some sites. It causes a hang, since the returned `Promise` of `eval()` hangs. We don't know as of today why this happens. Without wrapping it ini `eval()` it does not hang.~~

~~Workaround: Do a plain Runtime.evaluate instead.~~

workaround: Only wait on main frame.

Relates https://github.com/microsoft/playwright/issues/28995 (keeping it open until they confirm that it helps)

### DIFF
--- a/packages/playwright-core/src/server/screenshotter.ts
+++ b/packages/playwright-core/src/server/screenshotter.ts
@@ -204,7 +204,7 @@ export class Screenshotter {
     return this._queue.postTask(async () => {
       progress.log('taking page screenshot');
       const { viewportSize } = await this._originalViewportSize(progress);
-      await this._preparePageForScreenshot(progress, options.style, options.caret !== 'initial', options.animations === 'disabled');
+      await this._preparePageForScreenshot(progress, this._page.mainFrame(), options.style, options.caret !== 'initial', options.animations === 'disabled');
       progress.throwIfAborted(); // Avoid restoring after failure - should be done by cleanup.
 
       if (options.fullPage) {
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/screenshotter.ts
```diff
@@ -267,8 +267,9 @@ export class Screenshotter {
     if (!process.env.PW_TEST_SCREENSHOT_NO_FONTS_READY) {
       progress.log('waiting for fonts to load...');
       await Promise.all(this._page.frames().map(async frame => {
-        await frame.nonStallingEvaluateInExistingContext('document.fonts.ready', false, 'utility').catch(() => {});
-      }));
+        const utilityContext = await frame._utilityContext();
```
pavelfeldman, line 270: This is opposite of what we agreed on though, wouldn't you want to go for the non-stalling evaluate? `nonStallingRawEvaluateInExistingMainContext` in this case.

dgozman (approved):

packages/playwright-core/src/server/screenshotter.ts
mxschmitt: Discussed offline to only wait for fonts on the frame of the locator/mainframe.

PR #31297, opened 2024-06-13
fix(clock): fix pauseAt to arrive at wall time

pavelfeldman:


packages/playwright-core/src/server/injected/clock.ts
```diff
@@ -99,40 +99,41 @@ export class ClockController {
...
-    this._innerSetTime(time);
+    this._innerSetTime(asWallTime(time));
   }
 
-  setSystemTime(time: number) {
+  setSystemTime(time: number): number {
     this._replayLogOnce();
-    this._innerSetTime(time);
+    this._innerSetTime(asWallTime(time));
+    return time - this._now.time;
```
dgozman, line 108: It looks like this value if always zero?

packages/playwright-core/src/server/injected/clock.ts
```diff
@@ -282,10 +285,12 @@ export class ClockController {
 
     this._advanceNow(timer.callAt);
 
-    if (timer.type === TimerType.Interval)
-      this._timers.get(timer.id)!.callAt += timer.delay;
-    else
+    if (timer.type === TimerType.Interval) {
+      const callAt = this._timers.get(timer.id)!.callAt;
```
dgozman, line 289: I wonder whether `this._timers.get(timer.id) === timer`?

dgozman (approved):

packages/playwright-core/src/server/injected/clock.ts
pavelfeldman: Yeah, I don't need this return value, it is leftovers, removed.

packages/playwright-core/src/server/injected/clock.ts
pavelfeldman: I'm being stupid - doing a local refactoring w/o looking at context.

PR #31317, opened 2024-06-14
docs(ct): format story example code

sand4rt:


docs/src/test-components-js.md
```diff
@@ -246,26 +248,27 @@ Create a story file for your component:
...
+}
 // Export more stories here.
 ```
 
 Then test the component via testing the story:
 
 ```js title="input-media.test.spec.tsx"
 test('changes the image', async ({ mount }) => {
-  let mediaSelected: string | null = null;
+  let mediaSelected: string;
```
yury-s, line 266: I believe, this will not compile unless the variable is initialized.

yury-s (approved):

Issue #6543, opened 2021-05-12
[Feature] Listen for print dialog


aaliakseyenka:
Hi, I would like to test printing of the page, particularly that print dialog is opened on user action. Although seems existing api in typescript doesn't allow that. `page.on('dialog')` seems doesn't work with `print()` call. Am I missing anything?

JoelEinbinder:
No you are not missing anything. Sorry we don't current support listening for the print dialog.

aaliakseyenka:
thanks for prompt answer. Wondering are there any plans to add this in near future?

mxschmitt:
It's marked as collecting feedback. When there are enough upvotes and interest, it gets considered for implementation.

qwpnqjucz:
also interested in this.

leochung042:
Hi All

For all whose use case like me, then you could check on https://github.com/microsoft/playwright-python/issues/739 .

For my use case is after click on print button, then the Browser Print preview shows up, then I would like to save Print preview as PDF file.

Thank you,
Leo

nbnds:
Some corporate web applications  have printing feature at its core, like bookkeeping software, where you need to print invoices, reports etc. 
It would be very useful if we could test that we can invoke printing function by clicking the designated button.

pavelfeldman:
#### Why was this issue closed?

We are prioritizing the features based on the upvotes, recency and activity in the issue thread. It looks like this issue only has a handful of upvotes, has not been touched recently and/or we lack sufficient feedback to act on it. We are closing issues like this one to keep our bug database maintainable. Please feel free to open a new issue and link this one to it if you think this is a mistake.

Digraj22:
Any workaround for this, this is a show stopper for me. Also I do not want to execute the script in headless mode. #16100 is also closed.

Is there any plan to provide the support in future?

uchagani:
Here is how I handle it:  Since all we need to do is ensure that the print was actually called when the print button is clicked we do: 

(in Java but easy to translate to other languages):

```java
page.onLoad(pa -> pa.evaluate("window.print = function() { console.log('Print was triggered'); };"));
```

then you can use [waitForConsoleMessage](https://playwright.dev/java/docs/api/class-browsercontext#browser-context-wait-for-console-message) to verify that `Print was triggered` was called.

pradeipp:
[...]

PR #31340, opened 2024-06-17
docs: add guide for print dialogs

mxschmitt:
https://github.com/microsoft/playwright-internal/issues/211
Relates https://github.com/microsoft/playwright/issues/6543

### DIFF
--- a/docs/src/dialogs.md
+++ b/docs/src/dialogs.md
@@ -5,7 +5,7 @@ title: "Dialogs"
 
 ## Introduction
 
-Playwright can interact with the web page dialogs such as [`alert`](https://developer.mozilla.org/en-US/docs/Web/API/Window/alert), [`confirm`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm), [`prompt`](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt) as well as [`beforeunload`](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) confirmation.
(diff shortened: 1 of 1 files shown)


docs/src/dialogs.md
```diff
@@ -126,3 +126,59 @@ Page.Dialog += async (_, dialog) =>
...
+page.expose_binding("print", lambda: wait_for_dialog_fut.set_result(None))
+
+await page.goto("<url>")
+await page.get_by_text("Print it!").click()
+
+await wait_for_dialog_fut
+```
+
+```python sync
+page.addInitScript("""window.waitForPrintDialog = new Promise(resolve => window.print = resolve);""")
```
mxschmitt, line 168:
```suggestion
page.evaluate("""window.waitForPrintDialog = new Promise(resolve => window.print = resolve);""")
```

docs/src/dialogs.md
```diff
@@ -126,3 +126,52 @@ Page.Dialog += async (_, dialog) =>
 };
 await Page.CloseAsync(new() { RunBeforeUnload = true });
 ```
+
+## Print dialogs
+
+In order to assert that a print dialog via [`window.print`](https://developer.mozilla.org/en-US/docs/Web/API/Window/print) was opened, you can use the following snippet:
+
+```js
+await page.evaluate('(() => {window.waitForPrintDialog = new Promise(f => window.print = f);})()');
```
mxschmitt, line 135: - evaluate: needs to be after navigation
- addInitScript: needs to be before navigation

Which one is better? evaluate and be explicit?

mxschmitt: Discussed offline: evaluate is simpler.

dgozman (approved):

PR #31341, opened 2024-06-17
fix(fs-watcher) ignore node_modules on windows

### LINKED ISSUE #31337: [Bug]: UI Mode - 1 minute very slow at start of ui mode due to watcher watching excessive files (esp. on windows)
### Version

1.44.1

### Steps to reproduce

we have a test directory that encompasses a few packages so each package has node_modules in it.

We start UI mode and there is 1 minute of slowness where the ui is responsive but the test server is not - you click to start a test and nothing happens.

### Expected behavior

UI Mode is fast

### Actual behavior

N/A See additional context

### Additional context

This code:
https://github.com/microsoft/playwright/blob/main/packages/playwright/src/fsWatcher.ts#L55

Does not work in windows, so all node_modules are watched.

When I debug the path seperator is `\\` but the path passed to the function is actually unix style.

Further even if it did work, it watches all the sub directories of node_modules since it forces a path seperator on the end.

Instead of the function, adding `'**/node_modules/**` works and I believe from the docs will work on all platforms.

Secondly... You do not pass the test ignore directories from options into the watcher. This would help us by ignoring some other packages. e.g. We have a backend directory with a nuget folder that is very large.

### Environment

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
    Memory: 38.38 GB / 63.78 GB
  Binaries:
    Node: 20.12.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 4.3.0 - c:\npmprefix\yarn.CMD
    npm: 10.5.0 - C:\Program Files\nodejs\npm.CMD
  IDEs:
    VSCode: 1.90.1 - C:\Program Files\Microsoft VS Code\bin\code.CMD
  npmPackages:
    playwright: 1.46.0-alpha-2024-06-17 => 1.46.0-alpha-2024-06-17
```

lukpsaxo:
Partially fixes https://github.com/microsoft/playwright/issues/31337 by supporting ignoring node_modules on windows.

When I debug the function it gets a unix style path filename on windows, so the function never ignores node_modules. The ignore path globs are expected to use the unix path seperator and I've tested this fix works on windows and I assume that since mac uses unix style, it also works there (this is a pretty standard glob construct (chokidar points at any match https://github.com/micromatch/anymatch and anymatch has this exact example in their readme.md)

### DIFF
--- a/packages/playwright/src/fsWatcher.ts
+++ b/packages/playwright/src/fsWatcher.ts
@@ -16,7 +16,6 @@
 
 import { chokidar } from './utilsBundle';
 import type { FSWatcher } from 'chokidar';
-import path from 'path';
 
 export type FSEvent = { event: 'add' | 'addDir' | 'change' | 'unlink' | 'unlinkDir', file: string };
 
@@ -52,7 +51,7 @@ export class Watcher {
     if (!this._watchedFiles.length)
       return;
 
-    const ignored = [...this._ignoredFolders, (name: string) => name.includes(path.sep + 'node_modules' + path.sep)];
+    const ignored = [...this._ignoredFolders, '**/node_modules/**'];
     this._fsWatcher = chokidar.watch(watchedFiles, { ignoreInitial: true, ignored }).on('all', async (event, file) => {
       if (this._throttleTimer)
         clearTimeout(this._throttleTimer);


dgozman (approved):
Thank you for the PR! This looks great, merging in.

Issue #31254, opened 2024-06-11
[Bug]: Codegen freezes 


Shaun-Regan:
### Version

1.44.1

### Steps to reproduce

Start codegen from the vs code extension or via a terminal window and browser freezes in under a min and unable to complete recording.  Tried on two different machine, one machine had a complete fresh install of NPM, VScode and playwright (Only has the playwright extension on this machine) and same issue as my normal work laptop codegen browser freezes.

When the codegen browser freezes the laptop fan is working overtime even though the cpu is under 50% and memory usage holding steady at around 60%.

### Expected behavior

Used codegen tool in the past no issue. really useful tool that now is unusable as unable to finish recording a test

### Actual behavior

Start codegen from the vs code extension or via a terminal window and browser freezes and unable to complete recording.  Tried on two different machine, one machine had a complete fresh install of NPM, VScode and playwright (Only has the playwright extension on this machine) and same issue as my normal work laptop codegen browser freezes.

When the codegen browser freezes the laptop fan is working overtime even though the cpu is under 50% and memory usage holding steady at around 60%.

### Additional context

Start codegen from the vs code extension or via a terminal window and browsers freezes and unable to complete recording.  Tried on two different machine, one machine had a complete fresh install of NPM, VScode and playwright (Only has the playwright extension on this machine) and same issue as my normal work laptop codegen browser freezes.

When the codegen browser freezes the laptop fan is working overtime even though the cpu is under 50% and memory usage holding steady at around 60%.

### Environment

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
    Memory: 5.79 GB / 15.61 GB
  Binaries:
    Node: 16.15.1 - C:\Program Files\nodejs\node.EXE
    npm: 8.11.0 - C:\Program Files\nodejs\npm.CMD
  IDEs:
    VSCode: 1.90.0 - C:\Users\shaun.regan\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  npmPackages:
    @playwright/test: ^1.44.1 => 1.44.1

PS C:\Dev\Access Applause\Automation> npx envinfo --preset playwright
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.
   
  System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
[...]

PR #31349, opened 2024-06-17
chore: cache normalized whitespaces in recorder

yury-s:
Reference: https://github.com/microsoft/playwright/issues/31254

On the web page from the bug it reduces time to compute selectors by 8x:

**Before:**
before

**After:**
after

### DIFF
--- a/packages/playwright-core/src/utils/isomorphic/stringUtils.ts
+++ b/packages/playwright-core/src/utils/isomorphic/stringUtils.ts
@@ -75,7 +75,7 @@ export function cacheNormalizedWhitespaces() {
 
 export function normalizeWhiteSpace(text: string): string {
   let result = normalizedWhitespaceCache?.get(text);
-  if (!result) {
+  if (result === undefined) {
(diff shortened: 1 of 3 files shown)


packages/playwright-core/src/utils/isomorphic/stringUtils.ts
```diff
@@ -67,8 +67,19 @@ function cssEscapeOne(s: string, i: number): string {
...
+let normalizedWhitespaceCache: Map<string, string> | undefined;
+
+export function cacheNormalizedWhitespaces() {
+  normalizedWhitespaceCache = new Map();
+}
+
 export function normalizeWhiteSpace(text: string): string {
-  return text.replace(/\u200b/g, '').trim().replace(/\s+/g, ' ');
+  let result = normalizedWhitespaceCache?.get(text);
+  if (!result) {
```
dgozman, line 78:
```suggestion
  if (result === undefined) {
```

dgozman (approved):

PR #31369, opened 2024-06-18
docs: use long for time in milliseconds

yury-s:
In Java and .NET int is not enough to store millis since epoch.

### DIFF
--- a/utils/doclint/documentation.js
+++ b/utils/doclint/documentation.js
@@ -866,6 +866,7 @@ function csharpOptionOverloadSuffix(option, type) {
     case 'Buffer': return 'Byte';
     case 'Serializable': return 'Object';
     case 'int': return 'Int';
+    case 'long': return 'Int64';
     case 'Date': return 'Date';
   }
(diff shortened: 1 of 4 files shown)


utils/doclint/documentation.js
```diff
@@ -866,6 +866,7 @@ function csharpOptionOverloadSuffix(option, type) {
     case 'Buffer': return 'Byte';
     case 'Serializable': return 'Object';
     case 'int': return 'Int';
+    case 'long': return 'Int64';
```
mxschmitt, line 869:
```suggestion
    case 'long': return 'long';
```

I think it should be long? https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.fromunixtimeseconds?view=net-8.0

yury-s: It is for nullable types AFAIU, corresponding boxed type is Int64 similar to Int above, no?

dgozman (approved):

mxschmitt (approved):

PR #31370, opened 2024-06-18
test: add "clock" to the metadata for dashboard traceability

dgozman:


mxschmitt (approved):
headless=new bot would be nice if we could have it as well

PR #31381, opened 2024-06-19
devops: add headless new to flakiness dashboard

mxschmitt:
As per https://github.com/microsoft/playwright/pull/31370#pullrequestreview-2126124140.

### DIFF
--- a/tests/library/playwright.config.ts
+++ b/tests/library/playwright.config.ts
@@ -128,7 +128,13 @@ for (const browserName of browserNames) {
       metadata: {
         platform: process.platform,
         docker: !!process.env.INSIDE_DOCKER,
-        headful: !!headed,
+        headless: (() => {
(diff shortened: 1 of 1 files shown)


tests/library/playwright.config.ts
```diff
@@ -128,7 +128,13 @@ for (const browserName of browserNames) {
       metadata: {
         platform: process.platform,
         docker: !!process.env.INSIDE_DOCKER,
-        headful: !!headed,
+        headless: (() => {
```
dgozman, line 131: Are you sure this will be fine on the dashboard? AFAIK, it is a boolean today, so maybe it won't work?

dgozman (approved):

tests/library/playwright.config.ts
mxschmitt: we'll see!

PR #31386, opened 2024-06-19
feat(fetch): maxRetries for fetch

### LINKED ISSUE #30978: [Feature]: retry-mechanism for API request incase failure due to ECONNRESET
### 🚀 Feature Request

ECONNRESET might happen due to connectivity issues. As multiple executions are happening through API calls, sometimes randomly playwright through ECONNRESET error.  I used my logic of retrying using try-catch apiContext.post/get, the error object is undefined on ECONNRESET occurs. It would be nice playwright threw the error.statuscode as  ECONNRESET / provided an option to retry the API

    let retryCount = 0;
    let maxRetries = 3;
    while (retryCount < maxRetries) {
      try {
        const Response = await apiContext.get(URL, {
          headers: {
            'Connection': 'keep-alive',
            'Authorization': SessionToken,
            'Content-Type': 'application/json',
          },
          timeout: 0,
          failOnStatusCode: true
        });
        await Response.json();
      }
     } catch (error) {
        console.log(error); 
        if (error.code.toString().includes("ECONNRESET")) {
          retryCount++;
          console.log(`Connection reset. Retrying... (${retryCount}/${maxRetries})`);
          continue;
        } else {
          throw error;
        }
      }

### Example

_No response_

### Motivation

This feature will help API Automation better,reliable & saves the time of retrigger of entire automation suite

yury-s:
Fixes https://github.com/microsoft/playwright/issues/30978

### DIFF
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -462,7 +462,7 @@ Defaults to `20`. Pass `0` to not follow redirects.
 * langs: js, python, csharp
 - `maxRetries` <[int]>
 
-Maximum number of times socket errors should be retried. Currently only `ECONNRESET` error are retried. An error will be thrown if the limit is exceeded. Defaults to `5`. Pass `0` to disable retries.
+Maximum number of times socket errors should be retried. Currently only `ECONNRESET` error is retried. An error will be thrown if the limit is exceeded. Defaults to `5`. Pass `0` to disable retries.
 
 ## evaluate-expression
 - `expression` <[string]>
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -458,6 +458,12 @@ Whether to ignore HTTPS errors when sending network requests. Defaults to `false
 Maximum number of request redirects that will be followed automatically. An error will be thrown if the number is exceeded.
 Defaults to `20`. Pass `0` to not follow redirects.
 
+## js-python-csharp-fetch-option-maxretries
+* langs: js, python, csharp
+- `maxRetries` <[int]>
+
+Maximum number of times socket errors should be retried. Currently only `ECONNRESET` error are retried. An error will be thrown if the limit is exceeded. Defaults to `5`. Pass `0` to disable retries.
+
 ## evaluate-expression
 - `expression` <[string]>
 
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -462,7 +462,7 @@ Defaults to `20`. Pass `0` to not follow redirects.
 * langs: js, python, csharp
 - `maxRetries` <[int]>
 
-Maximum number of times socket errors should be retried. Currently only `ECONNRESET` error is retried. An error will be thrown if the limit is exceeded. Defaults to `5`. Pass `0` to disable retries.
+Maximum number of times socket errors should be retried. Currently only `ECONNRESET` error is retried. An error will be thrown if the limit is exceeded. Defaults to `0` - no retries.
 
 ## evaluate-expression
 - `expression` <[string]>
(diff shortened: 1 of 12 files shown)


docs/src/api/params.md
```diff
@@ -458,6 +458,12 @@ Whether to ignore HTTPS errors when sending network requests. Defaults to `false
 Maximum number of request redirects that will be followed automatically. An error will be thrown if the number is exceeded.
 Defaults to `20`. Pass `0` to not follow redirects.
 
+## js-python-csharp-fetch-option-maxretries
+* langs: js, python, csharp
+- `maxRetries` <[int]>
+
+Maximum number of times socket errors should be retried. Currently only `ECONNRESET` error are retried. An error will be thrown if the limit is exceeded. Defaults to `5`. Pass `0` to disable retries.
```
dgozman, line 465:
```suggestion
Maximum number of times socket errors should be retried. Currently only `ECONNRESET` error is retried. An error will be thrown if the limit is exceeded. Defaults to `5`. Pass `0` to disable retries.
```

packages/playwright-core/src/server/fetch.ts
```diff
@@ -247,6 +247,26 @@ export abstract class APIRequestContext extends SdkObject {
...
   }
 
+  private async _sendRequestWithRetries(progress: Progress, url: URL, options: SendRequestOptions, postData?: Buffer): Promise<Omit<channels.APIResponse, 'fetchUid'> & { body: Buffer }>{
+    const maxRetries = 5;
+    let backoff = 250;
+    for (let i = 0; i <= maxRetries; i++) {
+      try {
+        return await this._sendRequest(progress, url, options, postData);
+      } catch (e) {
+        if (i === maxRetries - 1 || (options.deadline && monotonicTime() + backoff > options.deadline))
```
dgozman, line 257:
```suggestion
        if (i === maxRetries || (options.deadline && monotonicTime() + backoff > options.deadline))
```

packages/playwright-core/src/server/fetch.ts
```diff
@@ -247,6 +247,26 @@ export abstract class APIRequestContext extends SdkObject {
...
+          throw new Error(`Failed after ${i + 1} attempt(s): ${e}`);
+        // Retry on connection reset only.
+        if (e.code !== 'ECONNRESET')
+          throw e;
+        progress.log(`Received ECONNRESET, will retry after ${backoff}ms.`);
```
dgozman, line 262:
```suggestion
        progress.log(`  Received ECONNRESET, will retry after ${backoff}ms.`);
```

packages/playwright-core/src/server/fetch.ts
```diff
@@ -247,6 +247,26 @@ export abstract class APIRequestContext extends SdkObject {
     }
   }
 
+  private async _sendRequestWithRetries(progress: Progress, url: URL, options: SendRequestOptions, postData?: Buffer): Promise<Omit<channels.APIResponse, 'fetchUid'> & { body: Buffer }>{
+    const maxRetries = 5;
```
dgozman, line 251:
```suggestion
    const maxRetries = options.maxRetries ?? 5;
```

docs/src/api/params.md
```diff
@@ -458,6 +458,12 @@ Whether to ignore HTTPS errors when sending network requests. Defaults to `false
 Maximum number of request redirects that will be followed automatically. An error will be thrown if the number is exceeded.
 Defaults to `20`. Pass `0` to not follow redirects.
 
+## js-python-csharp-fetch-option-maxretries
```
dgozman, line 461: What about Java?

dgozman (approved):

docs/src/api/params.md
yury-s: The option is in RequestOptions

docs/src/api/params.md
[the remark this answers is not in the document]
yury-s: In that case, we should just make it `0' by default and let the users opt-in. This way they'll have to think about the implications. WDYT?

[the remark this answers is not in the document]
yury-s: Change to 0 by default, will discuss it during 1.46 API review.

PR #30334, opened 2024-04-11
test: update Electron to v30

mxschmitt:
Electron v29 is the last supported version [as of today](https://www.electronjs.org/docs/latest/tutorial/electron-timelines). It only receives security updates, so it doesn't get our backport of https://github.com/electron/typescript-definitions/pull/264.

Most of the fixme's which were removed were related to `'error: Browser context management is not supported.'` which is fixed since https://github.com/electron/electron/pull/41718.

Fixes https://github.com/microsoft/playwright/issues/12096

~~Blocked by https://github.com/electron/typescript-definitions/pull/264~~

PR #31390, opened 2024-06-20
fix(electron): allow downloads

### LINKED ISSUE #31389: [Bug]: Trying to silently download files in Electron app but facing issues
### Version

1.44.1

### Steps to reproduce

# Code

```
import { _electron as electron} from 'playwright';
import path from 'path';
import fs from 'fs/promises';

class ElectronAppController{
    electronApp;
    mainWindow;
    
    async launchApp() {
        console.log("Launching app.");
        this.electronApp = await electron.launch({
            executablePath: 'C:\\path\\to\\Electron\\App'
        });

        // Get the main window of the Electron app
        this.mainWindow = await this.electronApp.firstWindow();
    }
   async downloadFile(fileUrl, savePath) {
        await this.mainWindow.evaluate((url) => {
            const a = document.createElement('a');
            a.href = url;
            a.innerText = 'Download file';
            a.id = 'download-link';
            a.download = '';
            document.body.appendChild(a);
        }, fileUrl);


        const downloadPromise = this.mainWindow.waitForEvent('download', { timeout: 60000 });

        try {
            await this.mainWindow.click('#download-link');

            const download = await downloadPromise;

            const suggestedFilename = download.suggestedFilename();
            const downloadFilePath = path.join(savePath, suggestedFilename);
            await download.saveAs(downloadFilePath);

            console.log(`File downloaded to: ${downloadFilePath}`);
            return downloadFilePath;
        } catch (error) {
            console.error("Error during download:", error);

            throw error;
        }
    }
}

(async () => {
    await obj.launchApp();
    const fileUrl = 'https://files.testfile.org/Video%20MP4%2FMountain%20-%20testfile.org.mp4';
    const savePath = 'C:\\Users\\ahad.ghouri\\Downloads';
    const downloadedFilePath = await obj.downloadFile(fileUrl, savePath);
    console.log(`File downloaded to: ${downloadedFilePath}`);
})();
```

### Expected behavior

I expect the downloadFile method to automatically download the video on the specified file path.

### Actual behavior

It is working perfectly for my Web app, however, when I'm using the same function to test my Electron app, it is displaying the dialog box which confirms the directory in which our file is getting saved, and eventually, getting timed out. The picture of the dialog box is shown below:

Screenshot 2024-06-20 134622

After further debugging, I've found out that it is getting timed out on the following line of code:

```
const download = await downloadPromise;
```

### Additional context

I can achieve the same result using HTTP but I have to complete this task using playwright. Would appreciate someone looking into this!

### Environment

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (28) x64 Intel(R) Core(TM) i7-14700K
    Memory: 20.26 GB / 31.77 GB
  Binaries:
    Node: 20.14.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.7.0 - C:\Program Files\nodejs\npm.CMD
  IDEs:
    VSCode: 1.90.0 - C:\Users\ahad.ghouri\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  npmPackages:
    @playwright/test: ^1.44.1 => 1.44.1
```

mxschmitt:
Maybe we should move https://github.com/microsoft/playwright/blob/main/tests/library/download.spec.ts into "page tests"? Except this single test which does `browser.newPage({ acceptDownloads: false });`.

This was fixed by https://github.com/electron/electron/pull/41718 - hence we didn't do it before. We should probably wait with this one until August 20th, when Electron v29 [reaches EOL](https://www.electronjs.org/docs/latest/tutorial/electron-timelines), or make this call either "not throw" / add a version check to it, since it throws in v29.

Fixes https://github.com/microsoft/playwright/issues/31389
Depends on https://github.com/microsoft/playwright/pull/30334

### DIFF
--- a/packages/playwright-core/src/server/chromium/crBrowser.ts
+++ b/packages/playwright-core/src/server/chromium/crBrowser.ts
@@ -348,7 +348,7 @@ export class CRBrowserContext extends BrowserContext {
   override async _initialize() {
     assert(!Array.from(this._browser._crPages.values()).some(page => page._browserContext === this));
     const promises: Promise<any>[] = [super._initialize()];
-    if (this._browser.options.name !== 'electron' && this._browser.options.name !== 'clank' && this._options.acceptDownloads !== 'internal-browser-default') {
+    if (this._browser.options.name !== 'clank' && this._options.acceptDownloads !== 'internal-browser-default') {
       promises.push(this._browser._session.send('Browser.setDownloadBehavior', {
         behavior: this._options.acceptDownloads === 'accept' ? 'allowAndName' : 'deny',
         browserContextId: this._browserContextId,
--- a/packages/playwright-core/src/server/browserContext.ts
+++ b/packages/playwright-core/src/server/browserContext.ts
@@ -654,8 +654,13 @@ export function validateBrowserContextOptions(options: channels.BrowserNewContex
     throw new Error(`"deviceScaleFactor" option is not supported with null "viewport"`);
   if (options.noDefaultViewport && !!options.isMobile)
     throw new Error(`"isMobile" option is not supported with null "viewport"`);
-  if (options.acceptDownloads === undefined)
+  if (options.acceptDownloads === undefined && browserOptions.name !== 'electron')
     options.acceptDownloads = 'accept';
+  // Electron requires explicit acceptDownloads: true since we wait for
+  // https://github.com/electron/electron/pull/41718 to be widely shipped.
+  // In 6-12 months, we can remove this check.
+  else if (options.acceptDownloads === undefined && browserOptions.name === 'electron')
+    options.acceptDownloads = 'internal-browser-default';
   if (!options.viewport && !options.noDefaultViewport)
     options.viewport = { width: 1280, height: 720 };
   if (options.recordVideo) {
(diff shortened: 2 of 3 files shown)


packages/playwright-core/src/server/chromium/crBrowser.ts
```diff
@@ -359,6 +359,20 @@ export class CRBrowserContext extends BrowserContext {
     await Promise.all(promises);
   }
 
+  private _shouldSetDownloadBehavior(): boolean {
+    if (this._browser.options.name === 'electron') {
+      // Chromium v124 correlates to Electron v30 which is the first version that has https://github.com/electron/electron/pull/41718 backported.
```
mxschmitt, line 364: Note: This expect users to have the latest minor/patch release - probably better but still risky?

dgozman: Shall we wait a bit for electron 31 instead? It seems like we are not in a rush, given that existing users with old electron version won't get the support anyway.

dgozman:
> Maybe we should move https://github.com/microsoft/playwright/blob/main/tests/library/download.spec.ts into "page tests"?

Sounds good, but we'll need to skip the tests on Android.

dgozman (approved):

PR #31391, opened 2024-06-20
feat(trace-viewer) add copy to clipboard on the Source > Stacktrace tab

ryanrosello-og:
Simplify copying of the source file names to the clipboard.

It made sense to only copy the filename part instead of the full path since the full path maybe different when the tests are executed in CI.

https://github.com/microsoft/playwright/assets/50574915/e8f1fed8-97b4-4a26-a6eb-90ace230efd7


ryanrosello-og:
I've no idea why this is picking up my other previous commits :( 

seems to be related to this https://github.com/orgs/community/discussions/39031

PR #31394, opened 2024-06-20
feat(trace-viewer) add copy to clipboard on the Source > Stacktrace tab

ryanrosello-og:
Simplify copying of the source file names to the clipboard.

It made sense to only copy the filename part instead of the full path since the full path maybe different when the tests are executed in CI.

https://github.com/microsoft/playwright/assets/50574915/c54e8e47-cec5-4c46-b790-e5c357753832


dgozman:
@ryanrosello-og Thank you for the PR! The code looks good, but I have two comments about the UI:
- Clicking on the "Copy filename" button should not also reveal the location, like a regular click on the line would do.
- Filename and line number jumping to the left upon hover is not user friendly. I wonder whether we can make the "Copy filename" button just be on top of the text below, perhaps with some nice gradient fadeaway? Take a look at Chrome DevTools for inspiration, for example hover in the "Settings -> Locations" list.

ryanrosello-og:
@dgozman  this is the best I could do based on my hot css skillz:

https://github.com/microsoft/playwright/assets/50574915/e311a88d-3601-4387-908e-7f6885cdfce3

dgozman:
@ryanrosello-og Thank you for the changes. We discussed this UI with the team, and think that it would be better to have a copy button next to the fully visible filename above the source content. On the below screenshot, to the right of "2.spec.ts" seems like a good place. WDYT?

source tab

ryanrosello-og:
Sounds good, I can take a look

pavelfeldman:
sg, let us know when it is ready via commenting, and it'll pop up in our internal review dashboard. do not reply to this message unless you want us to take a look at the pr!

ryanrosello-og:
As requested, this is what it looks like now:

**Trace Viewer:**

https://github.com/microsoft/playwright/assets/50574915/c86c6e0c-9222-4f52-b49c-ee83195b7927

**UI Mode:**

https://github.com/microsoft/playwright/assets/50574915/04baebb8-f895-4f21-b2e2-640244f5c5f7





I also made an attempt at creating a unit test for this but spent way more time troubleshooting it than the implementing the change itself :( It seems there is something special setup for this project.

packages/trace-viewer/src/ui/sourceTab.css
```diff
@@ -31,3 +31,13 @@
   box-shadow: var(--vscode-scrollbar-shadow) 0 6px 6px -6px;
   z-index: 10;
 }
+
+.copy-icon.codicon {
```
pavelfeldman, line 35: This will affect the entire app, not just this local icon. Did you mean to scope it?

packages/trace-viewer/src/ui/sourceTab.tsx
```diff
@@ -80,9 +81,24 @@ export const SourceTab: React.FunctionComponent<{
 
   const showStackFrames = (stack?.length ?? 0) > 1;
 
+  const getFileName = (fullPath?: string, lineNum?: number) => {
```
pavelfeldman, line 84: You are creating a new function every time the component renders, consider moving this function out of the component.

packages/trace-viewer/src/ui/stackTrace.css
```diff
@@ -29,4 +29,4 @@
 
 .stack-trace-frame-line {
   flex: none;
-}
+}
```
pavelfeldman, line 32: Revert as a no meaningful change.

packages/trace-viewer/src/ui/stackTrace.tsx
```diff
@@ -46,4 +46,4 @@ export const StackTraceView: React.FunctionComponent<{
       </>;
     }}
     onSelected={frame => setSelectedFrame(frames.indexOf(frame))} />;
-};
+};
```
pavelfeldman, line 49: Revert as a no meaningful change.

pavelfeldman (commented):
Looks good, one CSS nit and it is good to go!

packages/trace-viewer/src/ui/sourceTab.css
ryanrosello-og: oops, my bad. I have fixed this now

packages/trace-viewer/src/ui/sourceTab.tsx
ryanrosello-og: ✅ fixed

packages/trace-viewer/src/ui/stackTrace.css
ryanrosello-og: ✅ fixed

packages/trace-viewer/src/ui/stackTrace.tsx
ryanrosello-og: ✅ fixed

packages/trace-viewer/src/ui/sourceTab.tsx
ryanrosello-og: ✅ fixed

ryanrosello-og:
This ready for another round of reviews 🙂

pavelfeldman (approved):

PR #31362, opened 2024-06-18
feat(chromium-tip-of-tree): roll to r1232

playwrightmachine:
Good:

```
#      CURRENT_VERSION: 128.0.6536.0
# BRANCH_BASE_POSITION: 1314318
BRANCH_COMMIT="11d56f88594ff95c8e34acabd4893e9a1d431cf1"
```

Bad:

```
#      CURRENT_VERSION: 128.0.6543.0
# BRANCH_BASE_POSITION: 1316206
BRANCH_COMMIT="6c15c08c7ff32fc9bf8899b07f7f56789fb8f3bc"
```

Failing PR: https://github.com/microsoft/playwright/pull/31362

Failing tests:

```
    [chromium-library] › library/browsercontext-credentials.spec.ts:27:1 › should fail without credentials 
    [chromium-library] › library/browsercontext-credentials.spec.ts:41:1 › should work with setHTTPCredentials 
    [chromium-library] › library/browsercontext-credentials.spec.ts:112:1 › should fail with correct credentials and mismatching scheme 
    [chromium-library] › library/browsercontext-credentials.spec.ts:124:1 › should fail with correct credentials and mismatching hostname 
    [chromium-library] › library/browsercontext-credentials.spec.ts:138:1 › should fail with correct credentials and mismatching port 
    [chromium-library] › library/download.spec.ts:641:3 › should be able to download a inline PDF file via navigation 
    [chromium-library] › library/emulation-focus.spec.ts:104:3 › should not affect screenshots ─────
    [chromium-library] › library/permissions.spec.ts:153:3 › should support clipboard read ─────────
    [chromium-library] › library/tracing.spec.ts:412:14 › should produce screencast frames fit ─────
    [chromium-library] › library/tracing.spec.ts:412:14 › should produce screencast frames crop ────
    [chromium-library] › library/tracing.spec.ts:412:14 › should produce screencast frames scale ───
    [chromium-library] › library/video.spec.ts:446:5 › screencast › should scale frames down to the requested size  
    [chromium-library] › library/video.spec.ts:695:5 › screencast › should capture full viewport ───
    [chromium-library] › library/video.spec.ts:730:5 › screencast › should capture full viewport on hidpi 
    [chromium-library] › library/video.spec.ts:767:5 › screencast › should work with video+trace ───
```

Bisect 1 (`library/browsercontext-credentials.spec.ts:27:1`):

[...]

PR #31395, opened 2024-06-20
feat(chromium-tip-of-tree): roll to r1233

playwrightmachine:


mxschmitt:
Skipping due to https://github.com/microsoft/playwright/pull/31362#issue-2359609731

PR #31399, opened 2024-06-20
docs: added more explanation to sharding to make it more understandable.

Romarionijim:
* added more explanation to the sharding section for it to be more understandable what it is and what it's purpose.
* because from the explanation in the docs. - it sound like that a "shard" is a machine that is provided by playwright in addition to the current setup , like a managed machine, but in reality it is simply a separate job in the CI, dividing tests to separate jobs is sharding, which sounds a lot easier when you think about it this way because it could be confusing at the beginning IMO.

### DIFF
--- a/docs/src/test-sharding-js.md
+++ b/docs/src/test-sharding-js.md
@@ -5,7 +5,11 @@ title: "Sharding"
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


debs-obrien (approved):
nice thanks :)

PR #31409, opened 2024-06-22
fix: correctly report overridden headers on redirected requests

### LINKED ISSUE #31351: [Bug]: Headers set in context.route are sent by the browsers but not available in the network tab in trace view on redirects
### Version

1.44.1

### Steps to reproduce

###Create a local Dev Server (this one is in node):
```
const http = require("http");

function onRequest (request, response){
    const urlObject = new URL(request.url, 'https://example.com');
    const next_url = urlObject.searchParams.get('next_url');
    const pathname = urlObject.pathname;

    switch (pathname){
        case "/redirect":
            response.writeHead(302, {'Location': next_url});
            response.end();
            break;
    }
}

http.createServer(onRequest).listen(8080);
```
- Run by `node server.js`

###Create a test which sets headers in context.route:
```
// @ts-check

test('redirect with custom headers', async ({ page, context }) => {
  await context.route('**/*', async (route, request) => {
    const headers = request.headers();
    const url = request.url();

    headers['x-custom-header'] = 'value';

    await route.fallback({ headers });
  });

  // redirects the page with 302 to next_url
  await page.goto('http://localhost:8080/redirect?next_url=https://manytools.org/http-html-text/http-request-headers');

  await expect(1).toBe(2); // Only to fail & see the trace
});
```
Playwright Config
```
const { defineConfig, devices } = require('@playwright/test');
module.exports = defineConfig({
  testDir: './tests',
  fullyParallel: true,
  forbidOnly: !!process.env.CI,
  workers: process.env.CI ? 1 : undefined,
  reporter: 'html',
  use: {
    trace: 'on-first-retry',
  },

  /* Configure projects for major browsers */
  projects: [
    {
      name: 'chromium',
    },
});

```

### Expected behavior

- Both `localhost:8080` & `https://manytools.org/http-html-text/http-request-headers/` but the trace only shows the header in `localhost:8080` should have `x-custom-header`

### Actual behavior

- Notice the `x-custom-header` is present in both `localhost:8080` & `https://manytools.org/http-html-text/http-request-headers/` but the trace only shows the header in `localhost:8080`
Screenshot 2024-06-18 at 12 31 14 PM

### Additional context

_No response_

### Environment

```shell
Binaries:
  Node: 20.11.1
npmPackages:
  @playwright/test: 1.44.1 => 1.44.1
```

yury-s:
Fixes https://github.com/microsoft/playwright/issues/31351

### DIFF
--- a/packages/playwright-core/src/server/chromium/crNetworkManager.ts
+++ b/packages/playwright-core/src/server/chromium/crNetworkManager.ts
@@ -332,12 +332,13 @@ export class CRNetworkManager {
     }
 
     let route = null;
+    let headersOverride: types.HeadersArray | undefined;
     if (requestPausedEvent) {
       // We do not support intercepting redirects.
       if (redirectedFrom || (!this._userRequestInterceptionEnabled && this._protocolRequestInterceptionEnabled)) {
         // Chromium does not preserve header overrides between redirects, so we have to do it ourselves.
-        const headers = redirectedFrom?._originalRequestRoute?._alreadyContinuedParams?.headers;
-        requestPausedSessionInfo!.session._sendMayFail('Fetch.continueRequest', { requestId: requestPausedEvent.requestId, headers });
+        headersOverride = redirectedFrom?._originalRequestRoute?._alreadyContinuedParams?.headers;
+        requestPausedSessionInfo!.session._sendMayFail('Fetch.continueRequest', { requestId: requestPausedEvent.requestId, headers: headersOverride });
       } else {
         route = new RouteImpl(requestPausedSessionInfo!.session, requestPausedEvent.requestId);
       }
@@ -353,7 +354,8 @@ export class CRNetworkManager {
       route,
       requestWillBeSentEvent,
       requestPausedEvent,
-      redirectedFrom
+      redirectedFrom,
+      headersOverride: headersOverride || null,
     });
     this._requestIdToRequest.set(requestWillBeSentEvent.requestId, request);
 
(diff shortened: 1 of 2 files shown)


tests/page/page-request-continue.spec.ts
```diff
@@ -412,6 +412,27 @@ it('continue should propagate headers to redirects', async ({ page, server, brow
   expect(serverRequest.headers['custom']).toBe('value');
 });
 
+it('redirected requests should report overridden headers', {
+  annotation: { type: 'issue', description: 'https://github.com/microsoft/playwright/issues/31351' }
+}, async ({ page, server, browserName }) => {
+  it.fixme(browserName === 'firefox');
```
dgozman, line 418: What's up with Firefox?

packages/playwright-core/src/server/chromium/crNetworkManager.ts
```diff
@@ -353,15 +354,16 @@ export class CRNetworkManager {
...
+      headersOverride: headersOverride || null,
     });
     this._requestIdToRequest.set(requestWillBeSentEvent.requestId, request);
 
     if (requestPausedEvent) {
       // We will not receive extra info when intercepting the request.
       // Use the headers from the Fetch.requestPausedPayload and release the allHeaders()
       // right away, so that client can call it from the route handler.
-      request.request.setRawRequestHeaders(headersObjectToArray(requestPausedEvent.request.headers, '\n'));
+      request.request.setRawRequestHeaders(headersOverride ?? headersObjectToArray(requestPausedEvent.request.headers, '\n'));
```
dgozman, line 366: I think this whole `if` block might not apply to overrides, because we continue them right away, and will received extra info from the network stack at some point. Should we follow up?

dgozman (approved):

tests/page/page-request-continue.spec.ts
yury-s: The headers are not propagated in firefox apparently (other similar tests also disabled). We can tackle that separately.

packages/playwright-core/src/server/chromium/crNetworkManager.ts
yury-s: We already handle overridden headers in that case in Request.headers/rawRequestHeaders

superhit0:
I see the changes have been made only on chromium server. Are other browsers not facing this issue?

PR #31414, opened 2024-06-24
fix(test-runner): don't add slow annotation twice

mxschmitt:
Discovered here: https://github.com/microsoft/devops.playwright.dev/pull/10#issue-2369497287

### DIFF
--- a/packages/playwright/src/worker/workerMain.ts
+++ b/packages/playwright/src/worker/workerMain.ts
@@ -261,7 +261,7 @@ export class WorkerMain extends ProcessRunner {
             testInfo.expectedStatus = 'failed';
           break;
         case 'slow':
-          testInfo.slow();
+          testInfo._timeoutManager.slow();
           break;
       }
     };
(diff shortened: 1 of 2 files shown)


packages/playwright/src/worker/workerMain.ts
```diff
@@ -250,7 +250,9 @@ export class WorkerMain extends ProcessRunner {
         attachment => this.dispatchEvent('attach', attachment));
 
     const processAnnotation = (annotation: Annotation) => {
-      testInfo.annotations.push(annotation);
+      // testInfo.slow() does the same thing, hence we skip it here.
+      if (annotation.type !== 'slow')
```
dgozman, line 254: I'd rather call `testInfo._timeoutManager.slow()` instead of `testInfo.slow()` below.

Ideally, we wouldn't have duplication here, for example we can call `testInfo._modifier()` but make it not throw.

dgozman (approved):

PR #31457, opened 2024-06-27
feat: introduce touchscreen.touch() for dispatching raw touch events

yury-s:


DiamondHunters:
I think this is necessary
Simulating these events with JavaScript cannot pass `isTrusted` verification
Implementing it using dispatche_event is extremely complex and difficult to apply on Python

Puppeter also has similar functions
https://pptr.dev/api/puppeteer.touchscreen.touchstart

docs/src/api/class-touchscreen.md
```diff
@@ -20,3 +20,23 @@ Dispatches a `touchstart` and `touchend` event with a single touch at the positi
...
+Type of the touch event.
+
+### param: Touchscreen.touch.touchPoints
+* since: v1.46
+- `touchPoints` <[Array]<[Object]>>
+  - `x` <[float]> x coordinate of the event in CSS pixels.
+  - `y` <[float]> y coordinate of the event in CSS pixels.
+  - `id` ?<[int]> Identifier used to track the touch point between events, must be unique within an event. Optional.
+
+List of touch points for this event. `id` is a unieue identifier of a touch point that helps indentify it between touch events for the duration of its movement around the surface.
```
pavelfeldman, line 42:
```suggestion
List of touch points for this event. `id` is a unique identifier of a touch point that helps identify it between touch events for the duration of its movement around the surface.
```

docs/src/api/class-touchscreen.md
```diff
@@ -20,3 +20,23 @@ Dispatches a `touchstart` and `touchend` event with a single touch at the positi
...
+
+Synthesizes a touch event.
+
+### param: Touchscreen.touch.type
+* since: v1.46
+- `type` <[TouchType]<"touchstart"|"touchend"|"touchmove"|"touchcancel">>
+
+Type of the touch event.
+
+### param: Touchscreen.touch.touchPoints
```
pavelfeldman, line 35: Since you advertise it as event generator, align with https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/touches.

```suggestion
### param: Touchscreen.touch.touches
```

packages/playwright-core/src/client/input.ts
```diff
@@ -89,4 +89,8 @@ export class Touchscreen implements api.Touchscreen {
   async tap(x: number, y: number) {
     await this._page._channel.touchscreenTap({ x, y });
   }
+
+  async touch(type: 'touchstart'|'touchmove'|'touchend'|'touchcancel', touchPoints: { x: number, y: number, id?: number }[]) {
```
pavelfeldman, line 93: Looking at it, I prefer separate page.touchscreen.touchStart(points), touchEnd, touchMove, touchCancel. Maybe we are brave and do `page.touchscreen.start`

pavelfeldman (approved):

packages/playwright-core/src/client/input.ts
yury-s: Let's postpone this to the API review, for now we have bigger fish to fry.

wangdatu:
Why did you restore touchscreen. touch()? Is there any other way to replace Touchstart now touchmove、touchend

Issue #31440, opened 2024-06-25
Run Playwright server for component tests in docker 


denselmann:
I would like to run Playwright component tests on a Playwright server in Docker while keeping the tests running on the host system.
Like explained in #26482 I've followed these steps:

### Run Playwright Server in Docker

```
docker run \
-p 9009:9009 \
--add-host=hostmachine:host-gateway \
--rm --init -it \
mcr.microsoft.com/playwright:v1.44.1-jammy \
/bin/bash -c "cd /home/pwuser && npx -y playwright@1.44.1 run-server --port 9009"
```

### Connect client to running server
```
PW_TEST_CONNECT_WS_ENDPOINT=ws://127.0.0.1:9009/ playwright test -c playwright-ct.config.ts
```

No I get the following error

```
Error: net::ERR_CONNECTION_REFUSED at http://localhost:3100/
Call log: - navigating to "http://localhost:3100/", waiting until "load" at Object.page (/Users/myuser/Development/project/node_modules/@playwright/experimental-ct-core/lib/mount.js:38:16)
```

I think, this is because of the fact, that Playwright uses Vite under the hood and Vite tries to start a new webserver on the hostmachine which doesn't know about the Server running in the docker container.

This is my `playwright-ct.config.ts`:

```
import { defineConfig } from '@playwright/experimental-ct-react';
import { ScreenshotMode, TraceMode, VideoMode } from 'playwright/types/test';

/**
 * See https://playwright.dev/docs/test-configuration.
 */
export default defineConfig({
  testDir: 'tests',
  /* The base directory, relative to the config file, for snapshot files created with toMatchSnapshot and toHaveScreenshot. */
  snapshotDir: './__snapshots__',
  /* Maximum time one test can run for. */
  timeout: 10 * 1000,
  /* Run tests in files in parallel */
  fullyParallel: true,
  /* Fail the build on CI if you accidentally left test.only in the source code. */
  forbidOnly: !!process.env.CI,
  /* Retry on CI only */
  retries: process.env.CI ? 1 : 0,
  /* Opt out of parallel tests on CI. */
  workers: 1,
  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
  reporter: 'list',
  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
  use: {
    /* Port to use for Playwright component endpoint. */
    ctPort: 3100,
    headless: process.env.PLAYWRIGHT_HEADED !== 'on',
[...]

PR #31468, opened 2024-06-27
fix(ct): allow custom baseURL in config

oskar-martensson:
References: #31440


pavelfeldman:
Let's keep discussion in the Issue, I am not sure PR is necessary.

oskar-martensson:
Closing since I was able to work around my issue by setting ctViteConfig.preview.host to my hosts local IP.

PR #31470, opened 2024-06-27
docs(reporters): add common test report format reporter

Ma11hewThomas:
Hello,

Added [Common Test Report Format](https://www.ctrf.io/) reporter to the third-party reporter showcase. 

Thanks!


pavelfeldman:
Respectfully, examining https://github.com/ctrf-io suggests that your reporter ecosystem did not yet reach the level of the popularity for us to refer to it. We try to be strict in terms of outgoing links in order to not be used as a SEO tool. I understand that your project is highly relevant, and this is not the case, but we'd like to stick to the common principle.

PR #31489, opened 2024-06-29
feat(html-reporter): hide annotations started with "_"

### LINKED ISSUE #30179: [Feature]: pass custom data to Reporter API
### 🚀 Feature Request

I have a custom reporter and I want to pass some data to this reporter from test runtime. The data is specific for this reporter so I want it to be invisible in any other reporters.

### Example

I think about adding `.metadata / .reporterData` property to `testInfo`:
```ts
testInfo.metadata = /* some data */
```
And then in the reporter API:
```ts
  onTestEnd(test: TestCase, result: TestResult) {
    console.log(result.metadata);
  }
```

### Motivation

This feature allows to build richer reporters without impacting other reporters.

I need it for `playwright-bdd` project. To build Cucumber reports I pass bdd steps metadata as a special attachment named `__bddData`. After reading this data I hide this attachment from the Cucumber report. But there is a downside: if user generates both Playwright HTML and Cucumber reports, the first one will still show this weird `__bddData` attachment.

vitalets:
Fixes: https://github.com/microsoft/playwright/issues/30179

### DIFF
--- a/docs/src/test-annotations-js.md
+++ b/docs/src/test-annotations-js.md
@@ -159,7 +159,7 @@ You can also filter tests in the configuration file via [`property: TestConfig.g
 
 ## Annotate tests
 
-If you would like to annotate your tests with something more substantial than a tag, you can do that when declaring a test. Annotations have a `type` and a `description` for more context. Annotations will be visible in the test report, except those where `type` starts with `_` symbol.
+If you would like to annotate your tests with something more substantial than a tag, you can do that when declaring a test. Annotations have a `type` and a `description` for more context and available in reporter API. Playwright's built-in HTML reporter shows all annotations, except those where `type` starts with `_` symbol.
 
 For example, to annotate a test with an issue url:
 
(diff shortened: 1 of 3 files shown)


vitalets:

docs/src/test-annotations-js.md
```diff
@@ -159,7 +159,7 @@ You can also filter tests in the configuration file via [`property: TestConfig.g
 
 ## Annotate tests
 
-If you would like to annotate your tests with something more substantial than a tag, you can do that when declaring a test. Annotations have a `type` and a `description` for more context, and will be visible in the test report.
+If you would like to annotate your tests with something more substantial than a tag, you can do that when declaring a test. Annotations have a `type` and a `description` for more context. Annotations will be visible in the test report, except those where `type` starts with `_` symbol.
```
pavelfeldman, line 162: Strictly speaking, everything is visible as a part of the test reporter API, so I would be more specific and mention that annotations that start with _ are not rendered by the built-in reporters.

vitalets: @pavelfeldman updated text. I've pointed that only built-in HTML reporter hides "_" annotations.

pavelfeldman (approved):

PR #31511, opened 2024-07-01
docs(clock): update time types in Python/.NET

mxschmitt:


docs/src/api/class-clock.md
```diff
@@ -107,6 +114,12 @@ await page.Clock.RunForAsync("30:00");
 
 Time may be the number of milliseconds to advance the clock by or a human-readable string. Valid string formats are "08" for eight seconds, "01:00" for one minute and "02:34:10" for two hours, 34 minutes and ten seconds.
 
+### param: Clock.runFor.ticks
+* langs: python
```
pavelfeldman, line 118: What about C#? Should it be long there? I would think that it could be int across the langs

utils/doclint/documentation.js
```diff
@@ -473,8 +473,10 @@ class Member {
       this.type.visit(visitor);
     for (const arg of this.argsArray)
       arg.visit(visitor);
+    for (const lang in this.langs.overrides || {})
```
mxschmitt, line 476: This is the actual change, which fixes the lang override comments in Python. The rest is drive-by ESLint auto fixes.

docs/src/api/class-clock.md
```diff
@@ -45,6 +45,13 @@ await page.Clock.FastForwardAsync("30:00");
 
 Time may be the number of milliseconds to advance the clock by or a human-readable string. Valid string formats are "08" for eight seconds, "01:00" for one minute and "02:34:10" for two hours, 34 minutes and ten seconds.
 
+### param: Clock.fastForward.ticks
+* langs: python
+* since: v1.45
+- `ticks` <[int]|[string]>
```
pavelfeldman, line 51: I would expect API `long` to map to Python's `int` by default.

docs/src/api/class-clock.md
pavelfeldman: Same here, no need to override

pavelfeldman (approved):

PR #31532, opened 2024-07-03
chore: bump TypeScript to v5.5

mxschmitt:
We want to make sure that our exposed @playwright/test types work with the latest TypeScript version and didn't introduce unwanted regressions. Also this release offers a lot of great enhancements which should make our developer experience better: https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/


tests/playwright-test/types.spec.ts
```diff
@@ -71,12 +71,6 @@ test('should check types of fixtures', async ({ runTSC }) => {
         // @ts-expect-error
         baz: true,
       });
-      const fail9 = test.extend<{ foo: string }>({
```
mxschmitt, line 74: This doesn't produce an error anymore - @dgozman do you know why we added this test case in the first place? Since `auto` fixtures for the `test` scope seem to work as well:

```ts

const test = base.extend<{ foo: string }>({
  foo: [async ({ }, use) => {
    console.log('hello from the fixture')
    await use('foo');
  }, { scope: 'test', auto: true }],
})

test('my test', async ({ }) => {
  console.log('hello from the test')
});
```

packages/playwright-core/src/client/elementHandle.ts
```diff
@@ -282,16 +282,16 @@ export async function convertInputFiles(files: string | FilePayload | string[] |
     if (!items.every(item => typeof item === 'string'))
       throw new Error('File paths cannot be mixed with buffers');
 
-    const [localPaths, localDirectory] = await resolvePathsAndDirectoryForInputFiles(items as string[]);
+    const [localPaths, localDirectory] = await resolvePathsAndDirectoryForInputFiles(items);
```
mxschmitt, line 285: This [is one of the highlights](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#inferred-type-predicates) in this TypeScript v5.5 release.

tests/playwright-test/types.spec.ts
dgozman: This is because we do not allow overriding a non-auto fixture with an auto-fixture. I wonder why does it work now?

dgozman (approved):
lgtm modulo the auto-override case

tests/playwright-test/types.spec.ts
dgozman: Proposed change [in the types](https://github.com/microsoft/playwright/blob/main/utils/generate_types/overrides-test.d.ts#L147-L149) to fix the problem:

```ts
  [K in Exclude<keyof W, keyof PW>]?: [WorkerFixtureValue<W[K], W & PW>, { scope: 'worker', auto?: boolean, option?: boolean, timeout?: number | undefined, title?: string, box?: boolean }];
} & {
  [K in Exclude<keyof T, keyof PT>]?: TestFixtureValue<T[K], T & W & PT & PW> | [TestFixtureValue<T[K], T & W & PT & PW>, { scope?: 'test', auto?: boolean, option?: boolean, timeout?: number | undefined, title?: string, box?: boolean }];
```

mxschmitt: Awesome!

PR #31534, opened 2024-07-03
feat(trace-viewer) add request urls for actions initiated via APIRequestContext 

ryanrosello-og:
Improve action list UX by adding request urls for `APIRequestContext` actions

Currently the action list does not provide any information about the request urls for APIRequestContext actions. This makes it difficult to understand what the action is doing and would require the user to double click on the action to see the request url in the action details.

It would be advantageous to include the request url in the action list so that the user can quickly understand what the action is doing without having to double click on the action.

Note that this information is already exposed in the HTML report, so this change would only be for the action list in the Trace Viewer.

Snag_149cc729

### Changes made

Add the request url to the action list for APIRequestContext actions (without the baseURL).  When user hovers over the url, they can see the full url.

### DIFF
--- a/packages/trace-viewer/src/ui/actionList.tsx
+++ b/packages/trace-viewer/src/ui/actionList.tsx
@@ -102,7 +102,7 @@ export const renderAction = (
       <span>{action.apiName}</span>
       {locator && <div className='action-selector' title={locator}>{locator}</div>}
       {action.method === 'goto' && action.params.url && <div className='action-url' title={action.params.url}>{action.params.url}</div>}
-      {action.class === 'APIRequestContext' && action.params.url && <div className='action-url' title={action.params.url}>{excludeBaseUrl(action.params.url)}</div>}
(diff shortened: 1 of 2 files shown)


packages/trace-viewer/src/ui/actionList.tsx
```diff
@@ -111,3 +112,8 @@ export const renderAction = (
     </div>}
   </>;
 };
+
+function excludeBaseUrl(url: string): string {
+  const match = url.match(/^(https?:\/\/[^/]+)(.*)$/);
+  return match ? match[2] : url;
+}
```
mxschmitt, line 119:
```suggestion
function excludeOrigin(url: string): string {
  try {
    const urlObject = new URL(url);
    return urlObject.pathname + urlObject.search + urlObject.hash;
  } catch (error) {
    // If the URL is invalid, return the original input
    return url;
  }
}
```

I think we want something like that which is simpler to understand.

mxschmitt (approved):

PR #31560, opened 2024-07-05
docs(clock): Fix C# version of clock documentation

RenderMichael:
Honestly I'm a little confused by the options on `InstallAsync`. There's a `Time` and a `TimeString` which do the exact same thing, and then a `TimeDate` as well?

If Playwright isn't too strict about API compatibility, even with a release that just came out a day ago, I suggest changing it to a single `DateTime? InitialTime`, and if people want a string they can `DateTime.Parse()` it.

Either way, the docs currently are out of line with the actual API exposed and should be fixed.


mxschmitt (approved):

mxschmitt:
Thank you for your feedback! We'll collect feedback for it.

PR #31568, opened 2024-07-05
doc(clock): fix code snippets

### LINKED ISSUE #31566: [Docs]: (Clock) Example tests are failing due to incorrect function
### Page(s)

https://playwright.dev/docs/clock

Also checked in the next version - https://playwright.dev/docs/next/clock

### Issue:

When we run the example tests provided in this page, tests were failing like below,

image

image

the reason for the failure is that we're using _**toLocaleTimeString**_ in the code but expecting both date and time details during text assertion.  

```javascript
> new Date().toLocaleTimeString()
> '10:00:00'
``` 

### Suggested fix:

To display both date and time in the html, we need to replace '**_toLocaleTimeString_**' with '**_toLocaleString_**' like below

```javascript
> new Date().toLocaleString()
> '2/2/2024, 10:00:00'
``` 
Now by running the tests again, all were passing like below,

image

> Note: Test inactivity monitoring case is skipped since it is independent of this issue

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/31566

### DIFF
--- a/docs/src/clock.md
+++ b/docs/src/clock.md
@@ -179,9 +179,9 @@ With the help of the clock, you can speed up time and test this feature quickly.
 ```html
 <div id="remaining-time" data-testid="remaining-time"></div>
 <script>
-  const endTime = new Date().getTime() + 5 * 60_000;
+  const endTime = Date.now() + 5 * 60_000;
   const renderTime = () => {
-    const diffInSeconds = Math.round((endTime - new Date().getTime()) / 1000);
+    const diffInSeconds = Math.round((endTime - Date.now()) / 1000);
     if (diffInSeconds <= 0) {
       document.getElementById('remaining-time').textContent =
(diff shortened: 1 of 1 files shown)


docs/src/clock.md
```diff
@@ -176,6 +176,26 @@ Inactivity monitoring is a common feature in web applications that logs out user
 Testing this feature can be tricky because you need to wait for a long time to see the effect.
 With the help of the clock, you can speed up time and test this feature quickly.
 
+```html
+<div id="remaining-time" data-testid="remaining-time"></div>
+<script>
+  const endTime = new Date().getTime() + 5 * 60_000;
```
dgozman, line 182: I'd think `Date.now()` would be more natural than `new Date().getTime()`.

dgozman (approved):

PR #31577, opened 2024-07-06
feat(html-reporter) Set test title as the Browser page title and set favicon based on test outcome

ryanrosello-og:
Currently the html report page title does not provide any useful context in terms of the test title nor the test status and it lacks a favicon. 

This change will:
* set the Page title based on the test case title
* set a favicon based on the outcome of the test

**Browser light-theme** (able to hover on the tab to show full test case title)

_ForPR_light

**Browser dark-theme**

_ForPR

**NOTES**

* Originally I placed all svg inside `packages/html-reporter/public` but I was seeing these warnings ...

* The colors are based off these elements:

* The file sizes of the svg files:

Let me know what changes you require.


ryanrosello-og:
Will recreate pr once I have fixed these two issues  :
- Previous test status retained when clicking on the filters
- Seems the build is not including my other svg files

PR #31588, opened 2024-07-08
fix(toHaveScreenshot): don't exit with 1 with updateSnapshot: missing

### LINKED ISSUE #31503: [Bug]: playwright.config.ts - updateSnapshots: "missing" will always make the test fail, when there is no existing snapshot
### Version

1.44.1

### Steps to reproduce

1.  have some test with toHaveScreenshot() method
2.  delete the current snapshot (if there is any)
3 . in playwright.config.ts file have updateSnapshots set to  "missing"
4. the test from point 1 will fail. the new snapshot will be created, though and next run will be successful
5.

### Expected behavior

new snapshot should be created and the test should pass

### Actual behavior

the test will always fail. 

### Additional context

pls note, that if we set updateSnapshots  to 'all', the new snapshot will be also created, but the test will pass.

### Environment

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
    Memory: 6.79 GB / 15.79 GB
  Binaries:
    Node: 22.2.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.7.0 - C:\Program Files\nodejs\npm.CMD
  IDEs:
    VSCode: 1.90.2 - C:\Users\borsky\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  npmPackages:
    @playwright/test: ^1.44.1 => 1.44.1
    playwright: ^1.39.0 => 1.44.1
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/31503

### DIFF
--- a/packages/playwright/src/matchers/toMatchSnapshot.ts
+++ b/packages/playwright/src/matchers/toMatchSnapshot.ts
@@ -226,7 +226,7 @@ class SnapshotHelper {
       this.testInfo.attachments.push({ name: addSuffixToFilePath(this.outputBaseName, '-actual'), contentType: this.mimeType, path: this.actualPath });
     }
     const message = `A snapshot doesn't exist at ${this.snapshotPath}${isWriteMissingMode ? ', writing actual.' : '.'}`;
-    if (this.updateSnapshots === 'all') {
+    if (['all', 'missing'].includes(this.updateSnapshots)) {
       /* eslint-disable no-console */
       console.log(message);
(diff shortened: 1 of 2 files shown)


mxschmitt:
Closing for now since this is working as intended. updateSnapshot: missing is the default behaviour, and we need to fail for it on CI.

PR #31603, opened 2024-07-09
feat(test runner): expose FullConfig.actualWorkers

dgozman:
This property is used by reporters, so stop plumbing it through metadata.


yury-s (approved):
IIRC, we wanted to store it per project only, so I'm not sure if we want to keep it on the full config.

PR #31618, opened 2024-07-09
chore: address deprecated Vite warnings

mxschmitt:


packages/html-reporter/tsconfig.node.json
```diff
@@ -4,5 +4,5 @@
     "module": "esnext",
     "moduleResolution": "node"
   },
-  "include": ["vite.config.ts"]
+  "include": ["vite.config.ts", "bundle.ts"]
```
pavelfeldman, line 7: My recollection is that it was making the linter very slow. If it no longer is - fine with me.

packages/html-reporter/bundle.ts
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-import fs, { existsSync } from 'fs';
-import path from 'path';
+import * as fs from 'fs';
```
pavelfeldman, line 17: Can we keep esModuleInterop on?

pavelfeldman (approved):

tsconfig.json
```diff
@@ -35,5 +35,6 @@
   "include": ["packages"],
   "exclude": [
     "packages/*/lib",
-  ],
+    "packages/html-reporter/"
```
mxschmitt, line 38:
```suggestion
    "packages/html-reporter/",
```

PR #31619, opened 2024-07-10
chore: add android keys for channels switching

matthewpango:
"channel +" and  "channel -" are the most used keys, IMHO.


matthewpango:

yury-s (approved):
Is there a real use case for this?

matthewpango:
> Is there a real use case for this?

My primary focus is on Android TV and Android set-top boxes (STB) rather than mobile devices, so I can confirm that this is a definite issue. I am currently using a workaround instead of those buttons. I am unable to share the code, but I can assure you that there are numerous real cases.

PR #31639, opened 2024-07-11
chore: use own socks5 server for tests

mxschmitt:


utils/generate_third_party_notice.js
```diff
@@ -49,8 +49,10 @@ This project incorporates components from the projects listed below. The origina
     const allPackages = {};
     const projectDir = path.join(__dirname, '..', 'packages', project);
     const bundlesDir = path.join(projectDir, 'bundles');
-    for (const bundle of fs.readdirSync(bundlesDir)) {
-      const dir = path.join(bundlesDir, bundle);
+    for (const bundle of fs.readdirSync(bundlesDir, { withFileTypes: true })) {
+      if (!bundle.isDirectory())
```
mxschmitt, line 53: drive-by: this skips it for files, in my case: `.DS_Store`.

tests/config/serverFixtures.ts
```diff
@@ -47,30 +49,9 @@ export const serverFixtures: Fixtures<ServerFixtures, ServerWorkerOptions> = {
...
-          'HTTP/1.1 200 OK',
-          'Connection: close',
-          'Content-Type: text/html',
-          'Content-Length: ' + Buffer.byteLength(body),
-          '',
-          body
-        ].join('\r\n'));
-      });
-      // Catch and ignore ECONNRESET errors.
-      socket.on('error', () => {});
```
dgozman, line 68: I wonder whether we need something to handle this case in the new code.

dgozman (approved):

tests/config/serverFixtures.ts
mxschmitt: It would emit `this._socksProxy.addListener(SocksProxy.Events.SocksError` which wouldn't end up in unhandled rejections with the new implementation since its not an event with the name `"error"`.

PR #31642, opened 2024-07-11
fix(ct): import ct* flavour types from ct-core and then from pwt

### LINKED ISSUE #31626: [Bug]: `@playwright/experimental-ct-react` does not declare dependency on `playwright`
### Version

1.44.0

### Steps to reproduce

This is hard to reproduce since it may only surface on complex dependency setup. We encountered this on a larger monorepo. The theoretical explanation for the bug is sound since dependencies must be declared explicitly (either via `dependencies` or `peerDependencies`) to ensure package managers can place them at a location where they're reachable.

### Expected behavior

`@playwright/experimental-ct-react` can reach its dependencies

### Actual behavior

`playwright` is not reachable since it's not a declared dependency

### Additional context

[`@playwright/experimental-ct-react` imports from `playwright`](https://github.com/microsoft/playwright/blob/v1.44.0/packages/playwright-ct-react/index.d.ts#L38) without declaring `playwright` as a dependency. This might result in `playwright` not being reachable from `@playwright/experimental-ct-react` depending on the package manager/node-module linker used (e.g. pnpm or Yarn Berry with pnpm node-module linker).

Right now, pnpm users have can work around this bug by adding
```json
  "pnpm": {
    "packageExtensions": {
      "@playwright/experimental-ct-react@1.44.1": {
        "dependencies": {
          "playwright": "1.44.1"
        }
      }
    }
  }
````

in their `package.json`.

In https://github.com/microsoft/playwright/pull/26946, the dependency on `@playwright/test` was replaced with `playwright`. However, common setups have `@playwright/test` and `@playwright/experimental-ct-react` installed. Since `playwright` is now tucked under `@playwright/test`, it's not reachable from `@playwright/experimental-ct-react`.

`@playwright/experimental-ct-react` should probably import from `@playwright/test` instead and declare it as a peer dependency to avoid a mismatch in `@playwright/test` versions.

### Environment

```shell
Can't use this since I'm getting `error Cannot set properties of null (setting 'peer')` in Node.js 20.14.0

We started seeing bugs when switching from pnpm 8.15.0 to 9.4.0 until we added the `packageExtensions` entry.
```

mxschmitt:
Motivation: ct* packages were importing from `playwright` while `playwright` was not declared as a dependency. This changes the imports to `ct-core` which internally re-exports the needed imports from `playwright`.

Fixes https://github.com/microsoft/playwright/issues/31626

### DIFF
--- a/packages/playwright-ct-core/index.d.ts
+++ b/packages/playwright-ct-core/index.d.ts
@@ -56,4 +56,4 @@ export function defineConfig(config: PlaywrightTestConfig, ...configs: Playwrigh
 export function defineConfig<T>(config: PlaywrightTestConfig<T>, ...configs: PlaywrightTestConfig[]): PlaywrightTestConfig<T>;
 export function defineConfig<T, W>(config: PlaywrightTestConfig<T, W>, ...configs: PlaywrightTestConfig[]): PlaywrightTestConfig<T, W>;
 
-export { expect, devices } from 'playwright/test';
+export { expect, devices, Locator } from 'playwright/test';
--- a/packages/playwright-ct-react/index.d.ts
+++ b/packages/playwright-ct-react/index.d.ts
@@ -14,8 +14,7 @@
  * limitations under the License.
  */
 
-import type { Locator } from 'playwright/test';
-import type { TestType } from '@playwright/experimental-ct-core';
+import type { TestType, Locator } from '@playwright/experimental-ct-core';
 
 export interface MountOptions<HooksConfig> {
   hooksConfig?: HooksConfig;
@@ -33,5 +32,4 @@ export const test: TestType<{
   ): Promise<MountResult>;
 }>;
 
-export { defineConfig, PlaywrightTestConfig } from '@playwright/experimental-ct-core';
-export { expect, devices } from 'playwright/test';
+export { defineConfig, PlaywrightTestConfig, expect, devices } from '@playwright/experimental-ct-core';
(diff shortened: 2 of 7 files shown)


packages/playwright-ct-core/index.d.ts
```diff
@@ -56,4 +56,4 @@ export function defineConfig(config: PlaywrightTestConfig, ...configs: Playwrigh
 export function defineConfig<T>(config: PlaywrightTestConfig<T>, ...configs: PlaywrightTestConfig[]): PlaywrightTestConfig<T>;
 export function defineConfig<T, W>(config: PlaywrightTestConfig<T, W>, ...configs: PlaywrightTestConfig[]): PlaywrightTestConfig<T, W>;
 
-export { expect, devices } from 'playwright/test';
+export * from 'playwright/test';
```
dgozman, line 59: - You should export this on the first line, so that any overrides take precedence, for example custom `TestType` and `defineConfig`.
- Should there be a non-`.d.ts` change as well? I don't think `pnpm` is worried about types.

packages/playwright-ct-react/index.d.ts
```diff
@@ -33,5 +32,4 @@ export const test: TestType<{
   ): Promise<MountResult>;
 }>;
 
-export { defineConfig, PlaywrightTestConfig } from '@playwright/experimental-ct-core';
-export { expect, devices } from 'playwright/test';
+export * from '@playwright/experimental-ct-core';
```
dgozman, line 35: Same comment here and everywhere.

dgozman (approved):

PR #31645, opened 2024-07-11
docs: add missing pytest-playwright installation command

devniel:
Hello, 

It seems the explicit `conda install pytest-playwright` is missing, otherwise an error about `page fixtures` is thrown.

btw: I would also add explicit indications to install `pytest` as it's also required.

Thanks

### DIFF
--- a/docs/src/intro-python.md
+++ b/docs/src/intro-python.md
@@ -45,7 +45,7 @@ Install the [Pytest plugin](https://anaconda.org/Microsoft/pytest-playwright):
 ```bash
 conda config --add channels conda-forge
 conda config --add channels microsoft
-conda install playwright
+conda install pytest-playwright
 ```
 
 </TabItem>


docs/src/intro-python.md
```diff
@@ -40,12 +40,13 @@ pip install pytest-playwright
 </TabItem>
 <TabItem value="anaconda">
 
-Install the [Pytest plugin](https://anaconda.org/Microsoft/pytest-playwright):
+Install Playwright and the [Pytest plugin](https://anaconda.org/Microsoft/pytest-playwright):
 
 ```bash
 conda config --add channels conda-forge
 conda config --add channels microsoft
 conda install playwright
```
mxschmitt, line 48: We can remove this now, since the pytest plugin depends on playwright

mxschmitt (changes requested):

docs/src/intro-python.md
devniel: Thanks, done.

mxschmitt (approved):

Issue #31284, opened 2024-06-12
[Feature]: Allow for links in the attachment section of the playwright html reports


williamlabrum:
### 🚀 Feature Request

Ability to supply `link` instead of body or path. This would allow for folks to attach links on a per test run basis. This is an alternative to the existing Annotation.push() method that contextualizes these links at the test level instead of the needed test run level.

### Example

Option 1: new contentType
```
this.testInfo.attach('Playwright Homepage', {
    contentType: 'text/html',
    body: `<a href="https://www.playwright.dev/" target="_blank">Playwright Homepage</a>`,
});
```

Option 2: link type
```
this.testInfo.attach('Playwright Homepage', {
    link: 'https://www.playwright.dev/',
    target?: '_blank', (_blank for default but could supply self if same window)
});
```

Option 3: extend path type (this seems less desirable as there is a lot of sanitization going on in this route)
```
this.testInfo.attach('Playwright Homepage', {
    path: 'https://www.playwright.dev/',
    target: '_blank',
});
```

### Motivation

My company is able to provide a unique id that can be used to trace requests in DataDog APM tracing. The ability to have this link just makes it easier for the end user to click on a link and get to a particular trace for a particular test run. Annotations are an existing way to create links, but those are at the test case level, not the test run level so the contextualization is a bit harder to tie to a test run. 

In summary: Having external links may be valuable at the run level (much like attachments) to facilitate additional observability.

yury-s:
Folding into https://github.com/microsoft/playwright/issues/30095

yury-s:
Sorry, I jumped the gun and folded it. Would  annotations work for your use case if we showed the annotations per retry (currently it's always from the last run)?

```js
test.info().annotations.push({ type: 'link - ' + it.info().retry, description: 'https://<data log url>' });
```

yury-s:
Capturing discussion notes. Here is how we want to consider annotations vs attachments:
- annotations - static, part of the test definition, don't depend on the retry
- attachments - anything that depends on the test run and may change between retries

williamlabrum:
[...]

PR #31673, opened 2024-07-13
chore: linkify urls in attachments body

yury-s:
Reference: https://github.com/microsoft/playwright/issues/31284


packages/html-reporter/src/links.tsx
```diff
@@ -77,9 +78,9 @@ export const AttachmentLink: React.FunctionComponent<{
   return <TreeItem title={<span>
     {attachment.contentType === kMissingContentType ? icons.warning() : icons.attachment()}
     {attachment.path && <a href={href || attachment.path} download={downloadFileNameForAttachment(attachment)}>{linkName || attachment.name}</a>}
-    {attachment.body && <span>{attachment.name}</span>}
+    {attachment.body && <span>{linkifyText(attachment.name)}</span>}
   </span>} loadChildren={attachment.body ? () => {
-    return [<div className='attachment-body'><CopyToClipboard value={attachment.body!}/>{attachment.body}</div>];
+    return [<div className='attachment-body'><CopyToClipboard value={attachment.body!}/>{linkifyText(attachment.body!)}</div>];
```
mxschmitt, line 83:
```suggestion
    return [<div className='attachment-body'><CopyToClipboard value={attachment.body!}/>{attachmentBody}</div>];
```

let's do `const attachmentBody = attachment.body ? linkifyText(attachment.body) : null` to be on the safe side?

mxschmitt (approved):

packages/html-reporter/src/links.tsx
yury-s: There is a check already at line 82, it's just typescript is not smart enough

PR #31691, opened 2024-07-15
chore: allow evaluating Error objects

dgozman:
Previously, Error objects were replaced with strings.
Now, Error objects are reconstructed back from the serialized value.


mxschmitt (approved):
Note: In the language bindings this would give us then a PlaywrightException in e.g. .NET or a Error class in Python filled with the appropriate values which seems reasonable.

dgozman:
> Note: In the language bindings this would give us then a PlaywrightException in e.g. .NET or a Error class in Python filled with the appropriate values which seems reasonable.

I think it should be just a plain `Exception` in .NET, because it's not an error thrown by Playwright.

PR #31703, opened 2024-07-16
fix(list reporter): print step ends in non-TTY mode

### LINKED ISSUE #31674: [Feature]: Make `list` reporter's `printSteps` option work on non-TTY
### 🚀 Feature Request

Currently, CI environments like GitHub Actions [don't have a TTY mode](https://github.com/actions/runner/issues/241). Playwright straight up disables the `printSteps` option of the `list` reporter in that case, except if `PLAYWRIGHT_FORCE_TTY` is passed, but then the output is quite broken.

It would be great if the `list` reporter's `printSteps` option could still work for non-TTY modes, even if it's with a "degraded" output/features compared to the TTY-enabled one.

### Example

The following config would work as expected and provide useful feedback for test steps in CI/non-TTY mode:

```ts

export default defineConfig({
  reporter: process.env.CI === 'true'
    ? [['list', { printSteps: true }], ['github']]
    : [['list', { printSteps: true }]],
});
```

### Motivation

This would be beneficial to monitor the ongoing status of tests in CI for instance, where output is quite limited for long-running tests. Test steps are a great way to provide feedback during each test's execution, but this doesn't work in most CI environments right now since they don't have a TTY mode.

We're currently running Playwright in CI with `DEBUG=pw:api` to provide feedback, but it's obviously extremely verbose and drowns useful information. Having test steps' logs display properly instead would be enough for us to know what's going on during long tests.

Skn0tt:
When used in a terminal, the `list` reporter prints out information about test steps to help debugging. In non-TTY environments like GitHub Actions, currently it doesn't.

This PR changes that, so that in non-TTY environments you'll see the "step end" messages appearing, but not the "step begin" messages. This is a good middleground, because it helps the user understand test progress, without being too verbose.

Closes https://github.com/microsoft/playwright/issues/31674

### DIFF
--- a/packages/playwright/src/reporters/list.ts
+++ b/packages/playwright/src/reporters/list.ts
@@ -212,7 +212,7 @@ class ListReporter extends BaseReporter {
   private _appendLine(text: string, prefix: string) {
     const line = prefix + this.fitToScreen(text, prefix);
     if (process.env.PW_TEST_DEBUG_REPORTERS) {
-      process.stdout.write(this._lastRow + ' : ' + line + '\n');
+      process.stdout.write('#' + this._lastRow + ' : ' + line + '\n');
     } else {
       process.stdout.write(line);
       process.stdout.write('\n');
@@ -223,7 +223,7 @@ class ListReporter extends BaseReporter {
   private _updateLine(row: number, text: string, prefix: string) {
     const line = prefix + this.fitToScreen(text, prefix);
     if (process.env.PW_TEST_DEBUG_REPORTERS)
-      process.stdout.write(row + ' : ' + line + '\n');
+      process.stdout.write('#' + row + ' : ' + line + '\n');
     else
       this._updateLineForTTY(row, line);
   }
--- a/packages/playwright/src/reporters/list.ts
+++ b/packages/playwright/src/reporters/list.ts
@@ -36,7 +36,7 @@ class ListReporter extends BaseReporter {
 
   constructor(options: { printSteps?: boolean } = {}) {
     super();
-    this._printSteps = isTTY && getAsBooleanFromENV('PLAYWRIGHT_LIST_PRINT_STEPS', options.printSteps);
+    this._printSteps = getAsBooleanFromENV('PLAYWRIGHT_LIST_PRINT_STEPS', options.printSteps);
   }
 
   override printsToStdio() {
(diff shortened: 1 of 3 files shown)


packages/playwright/src/reporters/list.ts
```diff
@@ -80,17 +80,17 @@ class ListReporter extends BaseReporter {
     if (step.category !== 'test.step')
       return;
     const testIndex = this._resultIndex.get(result) || '';
+    const ordinal = ((result as any)[lastStepOrdinalSymbol] || 0) + 1;
```
Skn0tt, line 83: these lines were just shifted up before the `return`, so that the `stepIndex` is filled.

packages/playwright/src/reporters/list.ts
```diff
@@ -109,7 +109,6 @@ class ListReporter extends BaseReporter {
     if (!this._printSteps) {
       if (isTTY)
         this._updateLine(this._testRows.get(test)!, colors.dim(formatTestTitle(this.config, test, step.parent)) + this._retrySuffix(result), this._testPrefix(testIndex, ''));
-      return;
```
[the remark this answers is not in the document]
Skn0tt: no longer true with the new changes!

Skn0tt:
The added test failed on Windows, because apparently that's using a different checkmark: Screenshot 2024-07-16 at 16 59 56

I'm assuming that's why all the other tests have the `lines.pop()` line. Re-adding that in https://github.com/microsoft/playwright/pull/31703/commits/2dacfec182d7cded382b34c4ac3d30044a8ee6d8.

packages/playwright/src/reporters/list.ts
```diff
@@ -109,11 +117,10 @@ class ListReporter extends BaseReporter {
...
```
dgozman, line 112: Shouldn't we leave this line now? It seems like when `!this._printSteps`, it will update the current line, and then *also* print the step information? I would think it should not print steps in this case.

I also wonder why all the tests pass. Do we miss coverage of non-printSteps mode?

dgozman (commented):
Looks pretty good!

packages/playwright/src/reporters/list.ts
Skn0tt: Alright, so I think that https://github.com/Skn0tt/playwright/blob/dc152ad614cc7f83bc92ec5bc8d8b97238442d5a/tests/playwright-test/reporter-list.spec.ts#L94-L127 perfectly covers non-printSteps. The reason we didn't see this regression in tests is that in https://github.com/Skn0tt/playwright/blob/dc152ad614cc7f83bc92ec5bc8d8b97238442d5a/packages/playwright/src/reporters/list.ts#L134, `this._stepRows.get(step)!` is undefined. That leads to the lines being prefixed with `undefined :`, and then https://github.com/Skn0tt/playwright/blob/dc152ad614cc7f83bc92ec5bc8d8b97238442d5a/tests/playwright-test/reporter-list.spec.ts#L110 filters that out. I'll see if we can work around that somehow.

Skn0tt: In https://github.com/microsoft/playwright/pull/31703/commits/91f8c7d5ec755ec176ae6750805f057938e7cdc5, i've prefixed the line numbers with `#`, and updated the regex us to use that for matching. This ensures that lines with an `undefined` row number aren't swallowed, see here:
Screenshot 2024-07-16 at 19 58 54

dgozman (approved):

PR #31714, opened 2024-07-16
feat(reporter): links in attachment names, attachments name only

### LINKED ISSUE #31284: [Feature]: Allow for links in the attachment section of the playwright html reports

yury-s:
* Allow calling `test.info().attach('My text');` without options (no path nor body).
* Highlight links in attachment names:

image

Fixes https://github.com/microsoft/playwright/issues/31284

### DIFF
--- a/packages/html-reporter/src/links.tsx
+++ b/packages/html-reporter/src/links.tsx
@@ -78,7 +78,7 @@ export const AttachmentLink: React.FunctionComponent<{
   return <TreeItem title={<span>
     {attachment.contentType === kMissingContentType ? icons.warning() : icons.attachment()}
     {attachment.path && <a href={href || attachment.path} download={downloadFileNameForAttachment(attachment)}>{linkName || attachment.name}</a>}
-    {attachment.body && <span>{linkifyText(attachment.name)}</span>}
+    {!attachment.path && <span>{linkifyText(attachment.name)}</span>}
   </span>} loadChildren={attachment.body ? () => {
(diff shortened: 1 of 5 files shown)


yury-s:
We should probably rename `name` parameter of `attach()` to something like `description`.

mxschmitt (approved):

PR #31724, opened 2024-07-17
chore: remove Chromium Windows proxy hacks

### LINKED ISSUE #17252: [Bug] Chromium proxy config throws on Windows
I need test on authentication proxy, so I wrote proxy setting in `playwright.config.ts` .
Proxy setting is successed on firefox, but chronium still have proxy error.
What should I to do?

playwright version: 1.25.1

playwright.config.ts
```
const config : PlaywrightTestConfig = {
...
use:{
   proxy: {
      server: 'xxxxx',
      username: 'xxxxx',
      password: 'xxxxx',
   },
  },
}
```

chronium error message
```
browser.newContext: Browser needs to be launched with the global proxy, If all contexts override the proxy, global proxy will be never used and can be any string, for example "launch({ proxy: {server:'http://per-context'}})"
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/17252

### DIFF
--- a/docs/src/network.md
+++ b/docs/src/network.md
@@ -147,7 +147,7 @@ const browser = await chromium.launch({
 Browser browser = chromium.launch(new BrowserType.LaunchOptions()
   .setProxy(new Proxy("http://myproxy.com:3128")
   .setUsername('usr')
-  .setPassword('pwd'));
+  .setPassword('pwd')));
 ```
 
 ```python async
@@ -179,16 +179,12 @@ await using var browser = await BrowserType.LaunchAsync(new()
 });
 ```
 
-When specifying proxy for each context individually, **Chromium on Windows** needs a hint that proxy will be set. This is done via passing a non-empty proxy server to the browser itself. Here is an example of a context-specific proxy:
+Its also possible to specify it per context:
 
 ```js tab=js-test title="playwright.config.ts"
 import { defineConfig } from '@playwright/test';
 export default defineConfig({
   use: {
-    launchOptions: {
-      // Browser proxy option is required for Chromium on Windows.
-      proxy: { server: 'per-context' }
-    },
     proxy: {
       server: 'http://myproxy.com:3128',
     }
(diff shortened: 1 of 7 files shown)


packages/playwright-core/src/server/chromium/chromium.ts
```diff
@@ -110,12 +110,6 @@ export class Chromium extends BrowserType {
       artifactsDir,
       downloadsPath: options.downloadsPath || artifactsDir,
       tracesDir: options.tracesDir || artifactsDir,
-      // On Windows context level proxies only work, if there isn't a global proxy
-      // set. This is currently a bug in the CR/Windows networking stack. By
-      // passing an arbitrary value we disable the check in PW land which warns
-      // users in normal (launch/launchServer) mode since otherwise connectOverCDP
-      // does not work at all with proxies on Windows.
-      proxy: { server: 'per-context' },
```
mxschmitt, line 118: Maybe we should keep this for a few more months since it would break connections to older Chromium and reference to `crrev.com/c/5688851` so we can remove it in 6-12 months?

docs/src/network.md
```diff
@@ -179,16 +179,12 @@ await using var browser = await BrowserType.LaunchAsync(new()
 });
 ```
 
-When specifying proxy for each context individually, **Chromium on Windows** needs a hint that proxy will be set. This is done via passing a non-empty proxy server to the browser itself. Here is an example of a context-specific proxy:
+Its also possible to specify it per context:
 
 ```js tab=js-test title="playwright.config.ts"
```
dgozman, line 184: This snippet should call `browser.newContext()` inside a test, because specifying it in the options makes it global for all contexts.

docs/src/network.md
```diff
@@ -197,42 +193,30 @@ export default defineConfig({
...
   proxy: { server: 'http://myproxy.com:3128' }
 });
 ```
 
 ```java
-Browser browser = chromium.launch(new BrowserType.LaunchOptions()
-  // Browser proxy option is required for Chromium on Windows.
-  .setProxy(new Proxy("per-context"));
+Browser browser = chromium.launch();
 BrowserContext context = chromium.launch(new Browser.NewContextOptions()
```
dgozman, line 204:
```suggestion
BrowserContext context = browser.newContext(new Browser.NewContextOptions()
```

dgozman (approved):

packages/playwright-core/src/server/chromium/chromium.ts
mxschmitt: nvm, this was only to silent a check which we are removing now.

PR #31745, opened 2024-07-18
fix(toHaveScreenshot): attach "expected" when writing a missing expectation

dgozman:
Previously, only the "actual" attachment was created, pointing to the file in `test-results`. Now, the "expected" attachment pointing to the file in `__screenshots__` is also created. This will help any reporters that would like to know the "expected" path, for example to do a manual accept/decline of the baseline.

Fixes #30693.


mxschmitt (approved):
lgtm

PR #31749, opened 2024-07-18
test: unflake contextmenu recorder test

mxschmitt:
Looking at the last 5 runs on the flakiness dashboard, it shows that Chromium based browsers on Windows are red on this test. This is because on windows 'auxclick' fires before 'contextmenu' (this was also reproduced on a minimal example) which causes an action to be in progress, so when the actual 'contextmenu' fires, it does not emit any actions.

In order to fix it, we ignore right click 'auxclick' events, which sound reasonable, since they are handled via 'contextmenu' event.

### DIFF
--- a/packages/playwright-core/src/server/injected/recorder/recorder.ts
+++ b/packages/playwright-core/src/server/injected/recorder/recorder.ts
@@ -230,6 +230,9 @@ class RecordActionTool implements RecorderTool {
     // So we check the hovered element instead, and if it is a range input, we skip click handling
     if (isRangeInput(this._hoveredElement))
       return;
+    // Right clicks are handled by 'contextmenu' event if its auxclick
+    if (event.button === 2 && event.type === 'auxclick')
+      return;
     if (this._shouldIgnoreMouseEvent(event))
       return;
     if (this._actionInProgress(event))
--- a/tests/library/inspector/cli-codegen-3.spec.ts
+++ b/tests/library/inspector/cli-codegen-3.spec.ts
@@ -559,7 +559,7 @@ await page.GetByLabel("Coun\\"try").ClickAsync();`);
     ]);
   });
 
-  test('should consume contextmenu events, despite a custom context menu', async ({ page, openRecorder }) => {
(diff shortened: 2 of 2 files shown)


packages/playwright-core/src/server/injected/recorder/recorder.ts
```diff
@@ -1004,7 +1004,12 @@ export class Recorder {
     removeEventListeners(this._listeners);
     this._listeners = [
       addEventListener(this.document, 'click', event => this._onClick(event as MouseEvent), true),
-      addEventListener(this.document, 'auxclick', event => this._onClick(event as MouseEvent), true),
+      addEventListener(this.document, 'auxclick', event => {
+        // Right clicks are handled by 'contextmenu' event.
+        if ((event as MouseEvent).button === 2)
```
dgozman, line 1009: I think this is not the best place for this check. We should bail out in the place that actually records, instead of here. This event is fanned out to the overlay and all kinds of tools, and the logic you put here is specific to `RecordActionTool`.

tests/library/inspector/cli-codegen-3.spec.ts
```diff
@@ -597,20 +597,36 @@ await page.GetByLabel("Coun\\"try").ClickAsync();`);
...
+        'button: mouseup',
+        'button: contextmenu',
+      ]);
+    } else {
+      expect(await page.evaluate('log')).toEqual([
+        // hover
+        'button: pointermove',
+        'button: mousemove',
+        // trusted right click
+        'button: pointerup',
```
dgozman, line 620: Do you understand why is there a `pointerup` at the start of the right-click?

jsoref: pointer events were a fun way to try to describe how the world works for non mouse based inputs... https://w3c.github.io/pointerevents/#the-pointerup-event

Making all the events compatible/with pages is such a fun mess :) (signed, ex-(mobile)-browser/standards-dev)

dgozman (approved):

PR #31750, opened 2024-07-18
test: fix client-certificates tests on Windows

mxschmitt:


tests/library/client-certificates.spec.ts
```diff
@@ -56,6 +56,15 @@ const test = base.extend<{ serverURL: string, serverURLRewrittenToLocalhost: str
   }
 });
 
+test.use({
```
mxschmitt, line 59: We'll remove that again once https://github.com/microsoft/playwright/issues/17252 landed after Chromium 129.

dgozman (approved):

PR #31761, opened 2024-07-18
fix(client-certificates): pass TLS servername for SNI

mxschmitt:
Quoting from the Node.js docs:

> servername: [<string>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type) Server name for the SNI (Server Name Indication) TLS extension. It is the name of the host being connected to, and must be a host name, and not an IP address. It can be used by a multi-homed server to choose the correct certificate to present to the client, see the SNICallback option to [tls.createServer()](https://nodejs.org/api/tls.html#tlscreateserveroptions-secureconnectionlistener).

Without that, it doesn't work at most of the sites in the internet.

### DIFF
--- a/packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
+++ b/packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import type net from 'net';
+import net from 'net';
 import path from 'path';
 import type https from 'https';
 import fs from 'fs';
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -101,6 +101,7 @@ class SocksProxyConnection {
     const tlsOptions: tls.ConnectionOptions = {
       socket: this.target,
       rejectUnauthorized: !this.socksProxy.ignoreHTTPSErrors,
+      servername: this.host,
```
yury-s, line 104: Can this.host be an IP address? Do we validate it elsewhere?

yury-s (approved):

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
mxschmitt: Its coming from socks, already split up / validated.

dgozman (approved):

PR #31765, opened 2024-07-19
fix(webkit): reenable CrossOriginOpenerPolicy

### LINKED ISSUE #14043: [BUG] SharedArrayBuffer should work in WebKit
This popped up in https://github.com/microsoft/playwright/pull/13976, SharedArrayBuffer was reenabled in Safari https://webkit.org/blog/12140/new-webkit-features-in-safari-15-2/ but we disabled COOP handling in https://github.com/microsoft/playwright/pull/9185 as it broke some other functionality in WebKit. We should reenable COOP to match Safari and make SharedArrayBuffer work.

yury-s:
Depends on https://github.com/microsoft/playwright-browsers/pull/1160
Fixes: https://github.com/microsoft/playwright/issues/14043

### DIFF
--- a/packages/playwright-core/src/server/webkit/wkPage.ts
+++ b/packages/playwright-core/src/server/webkit/wkPage.ts
@@ -250,11 +250,10 @@ export class WKPage implements PageDelegate {
   private _onTargetDestroyed(event: Protocol.Target.targetDestroyedPayload) {
     const { targetId, crashed } = event;
     if (this._provisionalPage && this._provisionalPage._session.sessionId === targetId) {
+      this._maybeCancelCoopNavigationRequest(this._provisionalPage);
       this._provisionalPage._session.dispose();
       this._provisionalPage.dispose();
-      const provisionalPage = this._provisionalPage;
       this._provisionalPage = null;
-      this._maybeCancelNavigationRequest(provisionalPage);
     } else if (this._session.sessionId === targetId) {
       this._session.dispose();
       eventsHelper.removeEventListeners(this._sessionListeners);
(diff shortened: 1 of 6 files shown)


packages/playwright-core/src/server/webkit/wkProvisionalPage.ts
```diff
@@ -43,16 +53,20 @@ export class WKProvisionalPage {
...
-      eventsHelper.addEventListener(session, 'Network.loadingFinished', overrideFrameId(e => wkPage._onLoadingFinished(e))),
-      eventsHelper.addEventListener(session, 'Network.loadingFailed', overrideFrameId(e => wkPage._onLoadingFailed(session, e))),
+      eventsHelper.addEventListener(session, 'Network.loadingFinished', overrideFrameId(e => this._onLoadingFinished(e))),
+      eventsHelper.addEventListener(session, 'Network.loadingFailed', overrideFrameId(e => this._onLoadingFailed(e))),
     ];
 
     this.initializationPromise = this._wkPage._initializeSession(session, true, ({ frameTree }) => this._handleFrameTree(frameTree));
   }
 
+  pendingNavigationRequest(): network.Request | undefined {
```
dgozman, line 66: I'd call these `coopNavigationRequest()` or similar.

packages/playwright-core/src/server/webkit/wkPage.ts
```diff
@@ -1157,6 +1190,12 @@ export class WKPage implements PageDelegate {
     if (!request)
       return;
 
+    // If a provisional page was created after Cross-Origin-Opener-Policy headers were received, it
+    // will take over the request (with a new id). We'll only fail it if the provisional navigation
+    // fails/get canceled.
+    if (this._provisionalPage?.pendingNavigationRequest() === request.request)
```
dgozman, line 1196: It would be ideal to get rid of the extra `loadingFailed` from the original page.

packages/playwright-core/src/server/webkit/wkPage.ts
```diff
@@ -1015,10 +1017,41 @@ export class WKPage implements PageDelegate {
...
+    }
+  }
+
+  _changeRequestId(navigationRequest: network.Request, newRequestId: string) {
+    for (const [requestId, request] of this._requestIdToRequest) {
+      if (request.request === navigationRequest) {
+        this._requestIdToRequest.delete(requestId);
+        request.changeRequestId(newRequestId);
+        this._requestIdToRequest.set(newRequestId, request);
+        // Simply ignore this event as it has already been dispatched from the original process
```
dgozman, line 1042: This comment is not needed here.

packages/playwright-core/src/server/webkit/wkProvisionalPage.ts
```diff
@@ -62,6 +78,29 @@ export class WKProvisionalPage {
     this._wkPage._onFrameAttached(this._mainFrameId, null);
   }
 
+  private _onRequestWillBeSent(event: Protocol.Network.requestWillBeSentPayload) {
+    if (this._coopNavigationRequest && this._coopNavigationRequest.url() === event.request.url) {
```
dgozman, line 82: IIUC, the first request must be the `coopNavigationRequest`? Perhaps we should assert that urls are equal instead of comparing?

dgozman (approved):

packages/playwright-core/src/server/webkit/wkProvisionalPage.ts
yury-s: That is correct. I'm not sure about turning this into an assert, in case something changes and the page starts cancelling the cross process provisional load a bit later or something like that it could start firing.

PR #31776, opened 2024-07-19
test: unflake client-certificate WebKit tests

mxschmitt:
Before it was failing with

    Playwright client-certificate error: Hostname/IP does not match certificate's altnames: Host: local.playwright. is not cert's CN: localho


Skn0tt (approved):
Smart!

dgozman (approved):

PR #31786, opened 2024-07-20
feat(client-certificates): add http2 support

mxschmitt:


packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -21,10 +21,49 @@ import fs from 'fs';
...
 import { createSocket } from '../utils/happy-eyeballs';
-import { globToRegex, isUnderTest } from '../utils';
+import { globToRegex, isUnderTest, ManualPromise } from '../utils';
 import type { SocksSocketClosedPayload, SocksSocketDataPayload, SocksSocketRequestedPayload } from '../common/socksProxy';
 import { SocksProxy } from '../common/socksProxy';
 import type * as channels from '@protocol/channels';
+import { debugLogger } from '../utils/debugLogger';
+
+class ALPNCache {
+  private _cache = new Map<string, ManualPromise<string>>();
```
mxschmitt, line 31: Q: Should we clear the cache at some point? Maybe make it bound to the `SocksProxyConnection` instance?

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -84,51 +123,65 @@ class SocksProxyConnection {
...
+
+        internalTLS.on('end', () => closeBothSockets());
+        targetTLS.on('end', () => closeBothSockets());
+
+        internalTLS.on('error', () => closeBothSockets());
+        targetTLS.on('error', error => {
+          debugLogger.log('socks', `error when connecting to target: ${error.message}`);
+          if (internalTLS?.alpnProtocol === 'h2') {
+          // https://github.com/nodejs/node/issues/46152
+          // TODO: http2.performServerHandshake does not work here for some reason.
```
mxschmitt, line 168: As per offline discussion with dgozman, we might follow-up and spending more time on it.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -84,51 +123,65 @@ class SocksProxyConnection {
...
+          internalTLS.end();
+          targetTLS.end();
+        };
+
+        internalTLS.on('end', () => closeBothSockets());
+        targetTLS.on('end', () => closeBothSockets());
+
+        internalTLS.on('error', () => closeBothSockets());
+        targetTLS.on('error', error => {
+          debugLogger.log('socks', `error when connecting to target: ${error.message}`);
```
mxschmitt, line 165: Q: General question: Maybe we need a new [logging namespace](https://github.com/microsoft/playwright/blob/383e4b3c73859636b750e6f73d6f3d0c8ec17d6b/packages/playwright-core/src/utils/debugLogger.ts#L20) for it?

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -208,3 +261,14 @@ export function clientCertificatesToTLSOptions(
   }
   return tlsOptions;
 }
+
+function rewriteToLocalhostIfNeeded(host: string): string {
+  return host === 'local.playwright' ? 'localhost' : host;
+}
+
+// For testing.
+if (require.main === module) {
```
dgozman, line 270: I'd prefer to have this code in a separate file under `tests` or `utils` instead.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
dgozman: Yeah, one per proxy instance sounds better.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -21,10 +21,49 @@ import fs from 'fs';
...
+import { globToRegex, isUnderTest, ManualPromise } from '../utils';
 import type { SocksSocketClosedPayload, SocksSocketDataPayload, SocksSocketRequestedPayload } from '../common/socksProxy';
 import { SocksProxy } from '../common/socksProxy';
 import type * as channels from '@protocol/channels';
+import { debugLogger } from '../utils/debugLogger';
+
+class ALPNCache {
+  private _cache = new Map<string, ManualPromise<string>>();
+
+  get(host: string, port: number, success: (protocol: string) => void, error: (error: Error) => void): void {
```
dgozman, line 33: Why take two callbacks instead of making this an async method? Also, instead of rejecting I'd return `http/1.1`, so that the regular error handling logic from the proxy took place instead of something special.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -84,51 +123,65 @@ class SocksProxyConnection {
...
-        const responseBody = 'Playwright client-certificate error: ' + error.message;
-        internalTLS.end([
-          'HTTP/1.1 503 Internal Server Error',
-          'Content-Type: text/html; charset=utf-8',
-          'Content-Length: ' + Buffer.byteLength(responseBody),
-          '\r\n',
-          responseBody,
-        ].join('\r\n'));
-        closeBothSockets();
+    alpnCache.get(rewriteToLocalhostIfNeeded(this.host), this.port, serverRequestedAlpnProtocol => {
```
dgozman, line 126:
```suggestion
    alpnCache.get(rewriteToLocalhostIfNeeded(this.host), this.port, alpnProtocolChosenByServer => {
```

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -84,51 +123,65 @@ class SocksProxyConnection {
...
+      debugLogger.log('socks', `Proxy->Target ${this.host}:${this.port} chooses ALPN ${serverRequestedAlpnProtocol}`);
+      const dummyServer = tls.createServer({
+        key: fs.readFileSync(path.join(__dirname, '../../bin/socks-certs/key.pem')),
+        cert: fs.readFileSync(path.join(__dirname, '../../bin/socks-certs/cert.pem')),
+        ALPNProtocols: serverRequestedAlpnProtocol === 'h2' ? ['h2', 'http/1.1'] : ['http/1.1'],
```
dgozman, line 131: I wonder whether we want to support `http/1.0`?

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -84,51 +123,65 @@ class SocksProxyConnection {
...
-          'HTTP/1.1 503 Internal Server Error',
-          'Content-Type: text/html; charset=utf-8',
-          'Content-Length: ' + Buffer.byteLength(responseBody),
-          '\r\n',
-          responseBody,
-        ].join('\r\n'));
-        closeBothSockets();
+    alpnCache.get(rewriteToLocalhostIfNeeded(this.host), this.port, serverRequestedAlpnProtocol => {
+      debugLogger.log('socks', `Proxy->Target ${this.host}:${this.port} chooses ALPN ${serverRequestedAlpnProtocol}`);
+      const dummyServer = tls.createServer({
```
dgozman, line 128: Do we need to `close` the dummy server?

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
dgozman: Sure, why not.

dgozman (approved):

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
mxschmitt: I think most of the web is 1.1 nowadays - I'd hold off for now.

PR #31790, opened 2024-07-22
feat(reporter): add copy button for annotations

anthonycr0:
Adds a copy-to-clipboard button for each annotation so that text can be copied easily.
This re-uses the existing `CopyToClipboard` component and adds a `small` variant that can be used inline. The icon size and colour have been chosen to avoid being overwhelming when used inline.

Related to #30141 
I opted not to introduce the hover behaviour from #30749 as it's less discoverable, but can understand why that might be favourable. Certainly open to suggestions 😄 

Screenshot 2024-07-22 at 3 23 53 PM


pavelfeldman:
- I don't think we need two sizes of the copy icon, either small or large would do, I don't think we need both
- I don't think we can afford copy icon next to every label in the report, so reserving the space, but only making it visible on hover would be my recommendation.

anthonycr0:
@microsoft-github-policy-service agree company="Atlassian"

anthonycr0:
@pavelfeldman Thanks for the suggestions, I've made the copy button only appear on hover of an annotation and removed the size variants
image

packages/html-reporter/src/testCaseView.spec.tsx
```diff
@@ -76,6 +76,13 @@ test('should render test case', async ({ mount }) => {
   await expect(component.getByText('My test')).toBeVisible();
 });
 
+test('should render copy buttons for annotations', async ({ mount }) => {
+  const component = await mount(<TestCaseView projectNames={['chromium', 'webkit']} test={testCase} run={0} anchor=''></TestCaseView>);
+  await expect(component.getByText('Annotation text', { exact: false }).first()).toBeVisible();
+  component.getByText('Annotation text', { exact: false }).first().hover();
+  await expect(component.getByRole('button').first()).toBeVisible();
```
pavelfeldman, line 83: This feels very permissive. If there is any button in component, we think that there is a button and it works.

packages/html-reporter/src/testCaseView.css
```diff
@@ -73,4 +73,5 @@
   display: flex;
   flex-direction: row;
   flex-wrap: wrap;
-}
\ No newline at end of file
+}
```
pavelfeldman, line 76: revert?

packages/html-reporter/src/testCaseView.spec.tsx
anthonycr0: Ah yeah I agree, I'll tighten this up to get the specific copy button and verify the copied text.

dgozman (approved):
Thank you for the PR, looks good!

PR #31727, opened 2024-07-17
feat(test runner): `--only-changed` option

Skn0tt:
Introduces an `--only-changed [base ref]` option.

`playwright test --only-changed` filters the test run to only run test suites that have uncommitted changes.
`playwright test --only-changed=foo` runs only tests that were changed since commit `foo`.

In pull request CI, this can be used to run changed tests first and fail fast: `--only-changed=$GITHUB_BASE_REF`.
During local development, it can be used to quickly filter down to the touched set of tests suites.
In some rare usecases, this can also help to cut down on CI usage for pull requests. Tread with caution though.

File dependencies are taken into account to ensure that if you touched a utility file, all relevant tests are still executed.

Closes https://github.com/microsoft/playwright/issues/15075

PR #31794, opened 2024-07-22
chore(ct): remove `suite` dependency by connecting dependency graphs at read time, not write time

Skn0tt:
Broken out of https://github.com/microsoft/playwright/pull/31727 as per @dgozman's [request](https://github.com/microsoft/playwright/pull/31727#discussion_r1685793229).

The PR goal is to remove the `suite` argument from the Component testing's Vite Plugin. `suite` is used to enrich Vite's dependency graph with information about dependencies between test suites and helper files. It essentially merges the Vite graph with the `compilationCache.ts > fileDependencies` graph, and then writes the result back into `compilationCache.ts > externalDependencies`.

By refactoring this to make the connection on the reading end in `collectAffectedTestFiles`, we can drop the `suite` parameter.

We didn't yet have a test that depended on the dependency graph being connected correctly between `fileDependencies` and `externalDepedencies`, so I've [extended an existing test](https://github.com/microsoft/playwright/commit/53a539938b8b90d4ffdce7783e340232dfb02a18) to capture that.

### DIFF
--- a/packages/playwright-ct-core/src/vitePlugin.ts
+++ b/packages/playwright-ct-core/src/vitePlugin.ts
@@ -20,7 +20,7 @@ import type { AddressInfo } from 'net';
 import path from 'path';
 import { assert, calculateSha1, getPlaywrightVersion, isURLAvailable } from 'playwright-core/lib/utils';
 import { debug } from 'playwright-core/lib/utilsBundle';
-import { internalDependenciesForTestFile, setExternalDependencies } from 'playwright/lib/transform/compilationCache';
+import { setExternalDependencies } from 'playwright/lib/transform/compilationCache';
 import { stoppable } from 'playwright/lib/utilsBundle';
 import type { FullConfig, Suite } from 'playwright/types/testReporter';
 import type { PluginContext } from 'rollup';
@@ -48,8 +48,8 @@ export function createPlugin(): TestRunnerPlugin {
       configDir = configDirectory;
     },
 
-    begin: async (suite: Suite) => {
-      const result = await buildBundle(config, configDir, suite);
+    begin: async () => {
+      const result = await buildBundle(config, configDir);
       if (!result)
         return;
 
@@ -87,7 +87,7 @@ type BuildInfo = {
   }
 };
 
-export async function buildBundle(config: FullConfig, configDir: string, suite: Suite): Promise<{ buildInfo: BuildInfo, viteConfig: Record<string, any> } | null> {
+export async function buildBundle(config: FullConfig, configDir: string): Promise<{ buildInfo: BuildInfo, viteConfig: Record<string, any> } | null> {
   const { registerSourceFile, frameworkPluginFactory } = frameworkConfig(config);
   {
     // Detect a running dev server and use it if available.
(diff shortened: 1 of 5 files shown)


packages/playwright-ct-core/src/vitePlugin.ts
```diff
@@ -48,8 +48,8 @@ export function createPlugin(): TestRunnerPlugin {
       configDir = configDirectory;
     },
 
-    begin: async (suite: Suite) => {
-      const result = await buildBundle(config, configDir, suite);
+    begin: async () => {
```
dgozman, line 51: Let's keep the unused `Suite` parameter here to mirror the `Plugin` interface.

packages/playwright/src/transform/compilationCache.ts
```diff
@@ -218,9 +218,15 @@ export function collectAffectedTestFiles(dependency: string, testFileCollector:
     if (deps.has(dependency))
       testFileCollector.add(testFile);
   }
-  for (const [testFile, deps] of externalDependencies) {
-    if (deps.has(dependency))
-      testFileCollector.add(testFile);
+  for (const [importingFile, deps] of externalDependencies) {
+    if (deps.has(dependency)) {
+      testFileCollector.add(importingFile);
```
dgozman, line 223: We should only add `importingFile` if it is a test file. That's what the line 215 checks for.

packages/playwright/src/transform/compilationCache.ts
```diff
@@ -218,9 +218,15 @@ export function collectAffectedTestFiles(dependency: string, testFileCollector:
...
+
+      for (const [testFile, deps] of fileDependencies) {
```
dgozman, line 225: nit: can we avoid shadowing the `deps` variable? I find it harder to read.

packages/playwright/src/transform/compilationCache.ts
```diff
@@ -218,9 +218,15 @@ export function collectAffectedTestFiles(dependency: string, testFileCollector:
     if (deps.has(dependency))
       testFileCollector.add(testFile);
   }
-  for (const [testFile, deps] of externalDependencies) {
-    if (deps.has(dependency))
-      testFileCollector.add(testFile);
+  for (const [importingFile, deps] of externalDependencies) {
```
dgozman, line 221: Let's also update `dependenciesForTestFile` function.

packages/playwright/src/transform/compilationCache.ts
Skn0tt: done!

packages/playwright-ct-core/src/vitePlugin.ts
Skn0tt: done

packages/playwright/src/transform/compilationCache.ts
Skn0tt: I've updated this in https://github.com/microsoft/playwright/pull/31794/commits/72c446d8ad202dc60517c99be6755231b2352885 and verified manually that it works correctly, but i'm unsure how to add a test for it. The problem is that the function is only ever used to update the file watcher, and it's hard to observe that. Where would you add a test for this?

dgozman (approved):

packages/playwright/src/transform/compilationCache.ts
dgozman: This is not a performance-critical place, so premature optimization is not needed. Overall, we are serious about performance on the critical path, and in general in the core.

packages/playwright/src/transform/compilationCache.ts
dgozman: I think we can live without a test for this for now.

PR #31779, opened 2024-07-19
docs: spelling

jsoref:
Fixes misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/playwright/actions/runs/10015023629#summary-27685777352

The action will report that the changes in this PR would make it happy: https://github.com/jsoref/playwright/actions/runs/10015023971#summary-27685778305

---

I understand that the commit messages will need to be reworded to match house style. For the time being, these are merely noting the changes they contain so that when I rebase or need to drop things, I can. -- I've already rebased once as someone fixed one of the items that my draft work was going to fix.

---

## Testing
* The tests _mostly_ passed when I managed to trigger them, but there were a handful of things that I didn't quite understand
* There are a large number of warnings relating to a bad interaction between any workflow that uses this local action https://github.com/microsoft/playwright/blob/b535139b3292885dccf9e27ba2753c6c8d337c16/.github/actions/run-test/action.yml#L74-L80 and the action it calls -- I've opened Azure/login#474 asking them to refactor their action so that it doesn't cause so much noise while running this repository's tests
* I'm vaguely curious as to why this repository has a `branch` constraint for its `pull_request` events in its workflows -- that constraint gave me a number of additional headaches while trying to prepare this branch for this PR.

PR #31795, opened 2024-07-22
chore(trace-viewer): spelling: performed

jsoref:
Follow-up to #31779


pavelfeldman:
We don't accept trivial typos in comments changes since they result in non-trivial CI costs.

PR #31687, opened 2024-07-15
fix: add 'window-management' to chromium browser

imtheish97:
this fixes #27198 with testing and added documentation

adds the 'windowManagement' permission to the list of permissions that can be granted on Chromium.

https://chromedevtools.github.io/devtools-protocol/tot/Browser/#type-PermissionType

Issue #27198, opened 2023-09-20
[feature] add new chromium permission "WindowManagement"


RajeshKumarMahto:
Hi Team,

In my organization, we use multimonitors for our application by giving "window management" permission manually. I want to automate the things by giving same permissions via playwright but as of now playwright does not have that option listed in "GrantPermission" method.
Its kind of blocker for us. Pls incorporate that permission.

aslushnikov:
@RajeshKumarMahto I've never heard of "window management" permission - is this some kind of chrome or web permission?

RajeshKumarMahto:
It's chrome permission just like other permissions.

RajeshKumarMahto:
![window management](https://github.com/microsoft/playwright/assets/46883900/d60b40b0-4c8f-4b55-936c-0ebf10b6800e)

RajeshKumarMahto:
@aslushnikov Hi, any update on above requested feature.

alvingeoatgit:
I'm also having the same issue with "window management". It is a showstopper for me as well.

tmallafre:
Try using this apparently unrelated permissions. It worked for me:

```
        const context = this.page.context(this.browserContextOptions);
        await context.grantPermissions(['clipboard-read', 'clipboard-write', 'geolocation']);
```

RajeshKumarMahto:
> Try using this apparently unrelated permissions. It worked for me:
> 
> ```
>         const context = this.page.context(this.browserContextOptions);
>         await context.grantPermissions(['clipboard-read', 'clipboard-write', 'geolocation']);
> ```

These permissions worked. I am more concerned about windows management permission.

RajeshKumarMahto:
Hi @aslushnikov,
any update on the requested feature. Or any workaround you can suggest?

cowboyd:
We need this permission in order to test our application as well. Why not have a safety valve to allow granting _any_ permission. That way, the mechanism does not become stale as browser vendors add new capabilities that require new permissions?

imtheish97:
@yury-s made this change to cover java https://github.com/microsoft/playwright-java/pull/1629, can you please review?

Sathisheu1105:
Hi, I am working on a project where I need to open my application in multiple monitors. We need this fix.

ferenc-a:
Is there at least a workaround, like a Chrome or Playwright CLI flag people could use?

4iAmAve:
I also really need this.

[...]

PR #31801, opened 2024-07-22
Revert "fix: add 'window-management' to chromium browser (#31687)"

yury-s:
This reverts commit 0aa2f06f68959e2515f62ea9bb570cf1815230b2.

Discussed the new permission in the API review and decided not to proceed with the feature as we are not ready to commit to supporting it yet:
* the API is Chromium specific
* the API is still experimental
* there is no clarity to what extend the screen manipulation APIs will work in old headless which is our main test environment

We'll keep an eye on the demand for the feature and may get back to implementing it in the future.

Reference: https://github.com/microsoft/playwright/issues/27198

### DIFF
--- a/docs/src/api/class-browsercontext.md
+++ b/docs/src/api/class-browsercontext.md
@@ -980,7 +980,6 @@ A permission or an array of permissions to grant. Permissions can be one of the 
 * `'notifications'`
 * `'payment-handler'`
 * `'storage-access'`
-* `'window-management'`
 
 ### option: BrowserContext.grantPermissions.origin
 * since: v1.8
--- a/packages/playwright-core/src/server/chromium/crBrowser.ts
+++ b/packages/playwright-core/src/server/chromium/crBrowser.ts
@@ -434,7 +434,6 @@ export class CRBrowserContext extends BrowserContext {
       // chrome-specific permissions we have.
       ['midi-sysex', 'midiSysex'],
       ['storage-access', 'storageAccess'],
-      ['window-management', 'windowManagement']
     ]);
     const filtered = permissions.map(permission => {
       const protocolPermission = webPermissionToProtocol.get(permission);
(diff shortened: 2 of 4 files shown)


mxschmitt (approved):

cowboyd:
I get that it stinks to have to maintain wrappers around these volatile, vendor-specific APIs, but the problem is that if the application you're trying to test and/or automate uses this permission, you're just dead in the water with respect to Playwright.

Is there some escape hatch we don't know about that lets us use playwright without monkey-patching the source code?

RajeshKumarMahto:
This is the main feature we wanted in playwright as all our orgs app runs on multi monitor. We thought this would be available in playwright and as a result we chose playwright as automation testing tool for our organization but seems the multi monitor automation is not possible because of lack of votes.

mxschmitt:
@RajeshKumarMahto @cowboyd could you try the following? it might work:

```ts
  const session = await page.context().newCDPSession(page);
  await session.send('Browser.grantPermissions', {
    permissions: ['windowManagement'],
  });
```

Sathisheu1105:
Hi Yury Semikhatsky,  
 This is a useful feature. I will be happy to know if this feature will put back in place. I am sure other people would be required this feature.

cowboyd:
@mxschmitt Hey, thanks for the reply. We ended up removing window management from our app entirely, so I can't test it anymore. If it becomes a requirement again, I'll give this a try. Thanks!

RajeshKumarMahto:
> @RajeshKumarMahto @cowboyd could you try the following? it might work:
> 
> ```ts
>   const session = await page.context().newCDPSession(page);
>   await session.send('Browser.grantPermissions', {
>     permissions: ['windowManagement'],
>   });
> ```

I have tried the above way but it is throwing me below error
image

PR #31802, opened 2024-07-22
feat(webkit): roll to r2049

playwrightmachine:


yury-s:
It will be rolled along with necessary client side changes in https://github.com/microsoft/playwright/pull/31765

PR #31804, opened 2024-07-22
chore: allow empty string in regex patterns

pavelfeldman:


yury-s (approved):
Is there a bug number?

PR #31826, opened 2024-07-23
chore: client certificates api review

mxschmitt:


docs/src/api/params.md
```diff
@@ -523,14 +523,17 @@ Does not enforce fixed viewport, allows resizing window in the headed mode.
...
 - `clientCertificates` <[Array]<[Object]>>
-  - `url` <[string]> Glob pattern to match the URLs that the certificate is valid for.
-  - `certs` <[Array]<[Object]>> List of client certificates to be used.
-    - `certPath` ?<[string]> Path to the file with the certificate in PEM format.
-    - `keyPath` ?<[string]> Path to the file with the private key in PEM format.
-    - `pfxPath` ?<[string]> Path to the PFX or PKCS12 encoded private key and certificate chain.
-    - `passphrase` ?<[string]> Passphrase for the private key (PEM or PFX).
-
-An array of client certificates to be used. Each certificate object must have both `certPath` and `keyPath` or a single `pfxPath` to load the client certificate. Optionally, `passphrase` property should be provided if the private key is encrypted. If the certificate is valid only for specific URLs, the `url` property should be provided with a glob pattern to match the URLs that the certificate is valid for.
+  - `origin` <[string]> Glob pattern to match against the request origin that the certificate is valid for.
```
dgozman, line 526:
```suggestion
  - `origin` <[string]> Glob pattern that matches origins requiring this certificate, for example `"https://*.example.com"`.
```

How do I match `example.com` and `*.example.com` with a glob? Is glob the best pattern we can use here, given that we don't match the path?

dgozman (approved):

docs/src/api/params.md
mxschmitt: TBD.

PR #31835, opened 2024-07-24
chore(ct): replace `find-related-test-files` hook with `populateDependencies`

Skn0tt:
As discussed in the team sync, now that we have the `populateDependencies` plugin step, we don't need this CLI hook anymore.

I'm not sure if this refactor is worth it though, because it means we'll start calling the Plugin steps out of order (we call `populateDependencies` without calling `setup` first). Should we also run `setup`? If we do that, does that mean we should also run `teardown()`? Should we use a proper `TaskRunner` instance for managing that? Or should we just stick with the hook we have for simplicity over architectural conformance?

Lots of questions that i'm sure @dgozman has great answers to :D

This is a follow-up to https://github.com/microsoft/playwright/pull/31727.

### DIFF
--- a/packages/playwright-ct-core/index.js
+++ b/packages/playwright-ct-core/index.js
@@ -16,7 +16,7 @@
 
 const { test: baseTest, expect, devices, defineConfig: originalDefineConfig } = require('playwright/test');
 const { fixtures } = require('./lib/mount');
-const { clearCacheCommand, runDevServerCommand, findRelatedTestFilesCommand } = require('./lib/cliOverrides');
+const { runDevServerCommand } = require('./lib/cliOverrides');
 const { createPlugin } = require('./lib/vitePlugin');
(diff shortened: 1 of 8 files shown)


dgozman:
I think we should create a task runner and call `plugin.setup/teardown` as usual. And we can also migrate `clear-cache` and `dev-server` to do the same.

Skn0tt:
The failing test `run-tests.spec.ts:1153:5 › should produce output twice` also fails on `main` and I don't see how it'd be related to this change, so I'm ignoring it.

Skn0tt:
I've implemented that, but feel a little unsure about the `runDevServer` hook. In order for it to work, we need to pass over `FullConfigInternal`. Is `Plugin` part of our public API? If it is, giving it an internal config sounds like a no-go.

Issue #31520, opened 2024-07-02
[Bug]: The live preview is blank sometimes


illright:
### Version

1.45.0

### Steps to reproduce

https://github.com/illright/simple-playwright

```
pnpm i
pnpm exec playwright install
pnpm exec playwright test --ui
```

Then enable Chromium and Setup projects and hit F5 to run the tests

### Expected behavior

I expect that the live preview shows where Playwright is stuck during the test execution

### Actual behavior

During the runtime of the test, the timeline and live preview are empty.

This only seems to happen the first time you run the test after Playwright was started. If you rerun, then the timeline is behaving as expected.

### Environment

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1
    Memory: 531.55 MB / 16.00 GB
  Binaries:
    Node: 22.3.0 - /opt/homebrew/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.8.1 - /opt/homebrew/bin/npm
    pnpm: 9.4.0 - /opt/homebrew/bin/pnpm
    bun: 1.1.17 - /opt/homebrew/bin/bun
  IDEs:
    VSCode: 1.90.2 - /usr/local/bin/code
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.45.0 => 1.45.0
```

mxschmitt:
Investigation Note: No test is selected, if you execute a single test, it will focus and show it's live preview. We'll put a placeholder suggesting to select a test.

illright:
> Since UI Mode runs tests in parallel, we can't automatically select the "current test".

This is actually an interesting thing to consider. 
1. What is the goal of allowing UI mode to run tests in parallel? 
2. Should this be the default way, or should it be opt-in?

I personally see the value in running tests in parallel to quickly identify failures, but I have no issues doing that through the CLI (provided that the CLI is sufficiently ergonomic), and then dive into individual test failures with UI mode. As such, I would rather have the default way of running tests to be sequential, and have an extra button to initiate parallel test runs for those who are into that kind of thing

PR #31841, opened 2024-07-24
feat(ui mode): introduce Testing Options and Settings

dgozman:
Testing Options control tests, while Settings are UI mode settings.

Screenshot 2024-07-25 at 10 54 22 AM

These sections are separately expandable, collapsed by default.

Screenshot 2024-07-24 at 2 06 25 PM

References #31520.

### DIFF
--- a/packages/trace-viewer/src/ui/uiModeView.tsx
+++ b/packages/trace-viewer/src/ui/uiModeView.tsx
@@ -29,7 +29,7 @@ import { ToolbarButton } from '@web/components/toolbarButton';
 import { Toolbar } from '@web/components/toolbar';
 import type { XtermDataSource } from '@web/components/xtermWrapper';
 import { XtermWrapper } from '@web/components/xtermWrapper';
-import { toggleTheme } from '@web/theme';
+import { useDarkModeSetting } from '@web/theme';
 import { settings, useSetting } from '@web/uiUtils';
 import { statusEx, TestTree } from '@testIsomorphic/testTree';
 import type { TreeItem  } from '@testIsomorphic/testTree';
@@ -39,6 +39,7 @@ import type { TestModel } from './uiModeModel';
 import { FiltersView } from './uiModeFiltersView';
 import { TestListView } from './uiModeTestListView';
 import { TraceView } from './uiModeTraceView';
+import { SettingsView } from './settingsView';
 
 let xtermSize = { cols: 80, rows: 24 };
 const xtermDataSource: XtermDataSource = {
@@ -93,6 +94,37 @@ export const UIModeView: React.FC<{}> = ({
   const [isDisconnected, setIsDisconnected] = React.useState(false);
   const [hasBrowsers, setHasBrowsers] = React.useState(true);
   const [testServerConnection, setTestServerConnection] = React.useState<TestServerConnection>();
+  const [settingsVisible, setSettingsVisible] = React.useState(false);
+  const [testingOptionsVisible, setTestingOptionsVisible] = React.useState(false);
+
+  const [runWorkers, setRunWorkers] = React.useState(queryParams.workers);
+  const singleWorkerSetting = React.useMemo(() => {
+    return [
+      runWorkers === '1',
+      (value: boolean) => {
+        // When started with `--workers=1`, the setting allows to undo that.
+        // Otherwise, fallback to the cli `--workers=X` argument.
+        setRunWorkers(value ? '1' : (queryParams.workers === '1' ? undefined : queryParams.workers));
+      },
+      'Single worker',
+    ] as const;
+  }, [runWorkers, setRunWorkers]);
+
+  const [runHeaded, setRunHeaded] = React.useState(queryParams.headed);
+  const showBrowserSetting = React.useMemo(() => [runHeaded, setRunHeaded, 'Show browser'] as const, [runHeaded, setRunHeaded]);
+
+  const [runUpdateSnapshots, setRunUpdateSnapshots] = React.useState(queryParams.updateSnapshots);
+  const updateSnapshotsSetting = React.useMemo(() => {
+    return [
+      runUpdateSnapshots === 'all',
+      (value: boolean) => setRunUpdateSnapshots(value ? 'all' : 'missing'),
+      'Update snapshots',
+    ] as const;
+  }, [runUpdateSnapshots, setRunUpdateSnapshots]);
+
+  const [, , showRouteActionsSetting] = useSetting('show-route-actions', true, 'Show route actions');
+
+  const darkModeSetting = useDarkModeSetting();
 
   const inputRef = React.useRef<HTMLInputElement>(null);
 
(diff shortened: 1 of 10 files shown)


packages/web/src/theme.ts
```diff
@@ -64,3 +65,13 @@ export function removeThemeListener(listener: (theme: Theme) => void) {
 export function currentTheme(): Theme {
   return document.body.classList.contains('dark-mode') ? 'dark-mode' : 'light-mode';
 }
+
+export function useDarkModeSetting() {
+  const [theme, setTheme] = React.useState(currentTheme() === 'dark-mode');
```
Skn0tt, line 70: On an unrelated note, I think it'd be really nice to have not just light and dark mode, but also "follow system preference". I'll open an issue about that.

packages/web/src/components/toolbar.tsx
```diff
@@ -20,12 +20,16 @@ import * as React from 'react';
...
 export const Toolbar: React.FC<React.PropsWithChildren<ToolbarProps>> = ({
   noShadow,
   children,
-  noMinHeight
+  noMinHeight,
+  className,
+  onClick,
 }) => {
-  return <div className={'toolbar' + (noShadow ? ' no-shadow' : '') + (noMinHeight ? ' no-min-height' : '')}>{children}</div>;
+  return <div className={'toolbar' + (noShadow ? ' no-shadow' : '') + (noMinHeight ? ' no-min-height' : '') + ' ' + (className || '')} onClick={onClick}>{children}</div>;
```
Skn0tt, line 34: compiling the `className` string like this screams for https://www.npmjs.com/package/clsx to make it more readable. It's just a couple lines of code, we could steal the implementation if we don't want to add a new dependency. What do you think?

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -93,6 +94,37 @@ export const UIModeView: React.FC<{}> = ({
...
+  const [testingOptionsVisible, setTestingOptionsVisible] = React.useState(false);
+
+  const [runWorkers, setRunWorkers] = React.useState(queryParams.workers);
+  const singleWorkerSetting = React.useMemo(() => {
+    return [
+      runWorkers === '1',
+      (value: boolean) => {
+        // When started with `--workers=1`, the setting allows to undo that.
+        // Otherwise, fallback to the cli `--workers=X` argument.
+        setRunWorkers(value ? '1' : (queryParams.workers === '1' ? undefined : queryParams.workers));
```
Skn0tt, line 107:
```suggestion
        setRunWorkers(value ? '1' : queryParams.workers);
```

Wouldn't this work as well? And then `runWorkers` would be the source of truth instead of always having also check `queryParams.workers`.

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -457,6 +488,31 @@ export const UIModeView: React.FC<{}> = ({
           requestedCollapseAllCount={collapseAllCount}
           setFilterText={setFilterText}
         />
+        <Toolbar noShadow={true} noMinHeight={true} className='settings-toolbar' onClick={() => setTestingOptionsVisible(!testingOptionsVisible)}>
```
Skn0tt, line 491: if we end up needing this more often, this should probably be abstracted into a `ToolbarSection` component.

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -457,6 +488,31 @@ export const UIModeView: React.FC<{}> = ({
...
+          <span
+            className={`codicon codicon-${testingOptionsVisible ? 'chevron-down' : 'chevron-right'}`}
+            style={{ marginLeft: 5 }}
+            title={testingOptionsVisible ? 'Hide Testing Options' : 'Show Testing Options'}
+          />
+          <div className='section-title'>Testing Options</div>
+        </Toolbar>
+        {testingOptionsVisible && <SettingsView settings={[
```
Skn0tt, line 499: If we had an abstracted `ToolbarSection`, this part would probably be passed as a child to it. That'd make the nesting a little easier to understand on this

packages/trace-viewer/src/ui/workbench.tsx
```diff
@@ -67,7 +68,11 @@ export const Workbench: React.FunctionComponent<{
   const activeAction = model ? highlightedAction || selectedAction : undefined;
   const [selectedTime, setSelectedTime] = React.useState<Boundaries | undefined>();
   const [sidebarLocation, setSidebarLocation] = useSetting<'bottom' | 'right'>('propertiesSidebarLocation', 'bottom');
-  const [showRouteActions, , showRouteActionsSetting] = useSetting('show-route-actions', true, 'Show route actions');
+  const [, , showRouteActionsSettingInternal] = useSetting(showRouteActionsSetting ? '' : 'show-route-actions', true, 'Show route actions');
```
Skn0tt, line 71:
```suggestion
  const [, , showRouteActionsSettingInternal] = useSetting(showRouteActionsSetting ? undefined : 'show-route-actions', true, 'Show route actions');
```

The type allows passing `undefined`, so let's use that.

packages/trace-viewer/src/ui/workbench.tsx
```diff
@@ -67,7 +68,11 @@ export const Workbench: React.FunctionComponent<{
...
+
+  const showSettings = !showRouteActionsSetting;
+  showRouteActionsSetting ||= showRouteActionsSettingInternal;
+  const showRouteActions = showRouteActionsSetting[0];
```
Skn0tt, line 75: This is all very hard to understand. Would it be easier, if you made the `showRouteActionsSetting` prop required, and always passed it down from every parent component? That'd mean that the `useSetting('show-route-actions', ...)` call is duplicated across multiple components, but it'd make the control flow easier to understand.

Skn0tt (commented):
Looks generally good - really like how concise those CSS changes are! Left some comments on areas I found hard to understand.

Also, in the PR description you referenced an issue that looks pretty unrelated at first glance. Was that a typo?

dgozman:
> Also, in the PR description you referenced an issue that looks pretty unrelated at first glance. Was that a typo?

Nope, that's the right one, but it is buried somewhere in the discussion.

packages/trace-viewer/src/ui/workbench.tsx
Skn0tt: I think two different props would be wildly easier to understand than this, honestly

Skn0tt (approved):

packages/web/src/components/toolbar.tsx
mxschmitt: +1 for that! clsx has 0 dependencies, we also used that in the past a lot, especially helpful with conditions etc.

dgozman: Filed https://github.com/microsoft/playwright/issues/31863 to track this and more.

packages/trace-viewer/src/ui/workbench.tsx
dgozman: As discussed offline, filed https://github.com/microsoft/playwright/issues/31863 to track the follow ups.

PR #31842, opened 2024-07-24
fix(ui): added test in watched file should be run

### LINKED ISSUE #22211: [BUG] Watch mode in UI Mode does not handle adding new test
### System info
- Playwright Version: [v1.32.1]
- Operating System: [macOS 13.2.1]
- Other info: UI Mode

**Steps**
- Open UI mode
- Turn on Watch mode for one of the files.
- Add a new test in that file.
- Save the file

**Expected**

Watch mode should run the newly added test as well.

**Actual**

Watch mode runs only previously existing tests. The new test just appears in the list.

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/22211

Currently, when the server notifies the UI about changed files, the UI determines what files to re-run based on an old test list. By listing tests before that, we make sure that the test list is up-to-date, and that added tests are included in the next run.

I've also removed the `listChanged` event as discussed in the team sync. The event isn't used anywhere and fires in exactly the same cases where `testFilesChanged` fired, so i've folded them into one another.


Skn0tt:
Dima and me had a look at this synchronously, and decided to try removing one of the callbacks so we spare a roundtrip.

Skn0tt:
Dima and me discussed that we should combine the two `Watcher` instances into one to prevent duplicate events, and remove the concept of a watcher mode at the same time. We don't remember why it exists.

Skn0tt:
I've successfully dropped the `listChanged` event in favor of `testFilesChanged`, and simplified the `Watcher` implementation.

There's two new tests in `test-server-connection.spec.ts` that assert on the interface we have to VS Code, so we can be a little more sure about changes in that area in the future. We can extend that with more test cases as we introduce new functionality to the test server.

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -336,9 +323,26 @@ export const UIModeView: React.FC<{}> = ({
 
   // Watch implementation.
```
dgozman, line 324: Let's update the comment, this is list update + watch now. Or just remove it.

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -336,9 +323,26 @@ export const UIModeView: React.FC<{}> = ({
...
   React.useEffect(() => {
-    if (!testServerConnection)
+    if (!testServerConnection || !teleSuiteUpdater)
       return;
-    const disposable = testServerConnection.onTestFilesChanged(params => {
+    const disposable = testServerConnection.onTestFilesChanged(async params => {
```
dgozman, line 328: Let's comment that we first update the list to get the up-to-date list of tests, and then run the watched ones.

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -336,9 +323,26 @@ export const UIModeView: React.FC<{}> = ({
...
+        } catch (e) {
+          // eslint-disable-next-line no-console
+          console.log(e);
+        } finally {
+          setIsLoading(false);
+        }
+      });
+      await commandQueue.current;
+
+      const testModel = teleSuiteUpdater.asModel();
```
dgozman, line 343: Perhaps return if `params.testFiles` is empty?

packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -336,9 +323,26 @@ export const UIModeView: React.FC<{}> = ({
...
+      const testTree = new TestTree('', testModel.rootSuite, testModel.loadErrors, projectFilters, pathSeparator);
```
dgozman, line 344: It is unfortunate that we have to create a new `TestTree` here, even though one would be created naturally through `useMemo()`. I wonder whether there is a better pattern to handle such a thing in React?

packages/playwright/src/runner/testServer.ts
```diff
@@ -86,8 +90,7 @@ class TestServerDispatcher implements TestServerInterface {
           gracefullyProcessExitDoNotHang(0);
       },
     };
-    this._globalWatcher = new Watcher('deep', () => this._dispatchEvent('listChanged', {}));
```
dgozman, line 89: We should double check that emitting `testFilesChanged` event with an empty list of test files is properly ignored in the VSCode extension.

tests/playwright-test/test-server-connection.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { test as baseTest, expect } from './ui-mode-fixtures';
+
+import { TestServerConnection } from '../../packages/playwright/src/isomorphic/testServerConnection';
```
dgozman, line 18: You probably want to import from `lib`, not `src`, otherwise path mapping goes crazy.

tests/playwright-test/test-server-connection.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  await writeFiles({
+    'utils.ts': `
+      export const expected = 123;
+      `,
+  });
+
+  await expect.poll(() => testServerConnection.events).toHaveLength(1);
+  expect(testServerConnection.events).toEqual([
+    ['testFilesChanged', { testFiles: [testInfo.outputPath('a.test.ts')] }]
```
dgozman, line 73: Are we trying to test that there is a single event here? Seems kinds of redundant, because we rely on chokidar to make it so.

tests/playwright-test/test-server-connection.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  ]);
+});
+
+test('test run events', async ({ testServerConnection, writeFiles }) => {
```
dgozman, line 77: Perhaps it would make more sense to test the end-user behavior? While I do think such a test makes sense for a public API that should preserve backwards compatibility, test server interface is not public yet. WDYT?

dgozman (commented):
This looks good, just a few comments.

packages/trace-viewer/src/ui/uiModeView.tsx
Skn0tt: I agree. The big problem with `useMemo` is that it only updates on the next render, which is too late for this. I feel like this component could benefit from separating some of the underlying state from the presentation. I'd like to do that in a separate PR, though.

Skn0tt: I agree. The big problem with useMemo is that it only updates on the next render, which is too late for this. I feel like this component could benefit from separating some of the underlying state from the presentation. I'd like to do that in a separate PR, though.

tests/playwright-test/test-server-connection.spec.ts
Skn0tt: We're trying to test that there are no other events. Prior to this PR, there would've also been a `listChanged` event.

Also, prior to this PR, we would've seen two `testFilesChanged` events because we had two chokidar instances emitting the same file changes. So there's some value in this test :D If it had existed before, we would've spotted earlier that the two `Watcher` instances are a redundant.

tests/playwright-test/test-server-connection.spec.ts
Skn0tt: This API might not be public, but we still need to preserve backwards compatibility so we don't break the VS Code extension. That's the main value in this test suite I think.

tests/playwright-test/test-server-connection.spec.ts
Skn0tt: good idea! that's probably what tripped up the linter.

packages/playwright/src/runner/testServer.ts
Skn0tt: will do!

Skn0tt (commented):
Thanks for the feedback! I've left comment and addressed the actionable feedback in c24be29a86187d513c2faa9835014b6f951e2ee9.

tests/playwright-test/test-server-connection.spec.ts
Skn0tt: `report` also is one of the events of `TestServerInterface`, but I agree that if its contents exclusively contain reporter output, then it doesn't make sense to test it.

In https://github.com/microsoft/playwright/pull/31842/commits/9ee6994979b3332bba881f265c70131e1e158491, i've removed the `report` event and rewrote this test to cover STDIO interception instead.

tests/playwright-test/test-server-connection.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        console.log("this goes to stdout");
+        console.error("this goes to stderr");
+        expect(true).toBe(true);
+      });
+      `,
+  });
+
+  const tests = await testServerConnection.runTests({ trace: 'on' });
+  expect(tests).toEqual({ status: 'passed' });
+  await expect.poll(() => testServerConnection.events).toHaveLength(8);
```
dgozman, line 91: nit: I'd recommend to not hard-code the length (for an easier-to-maintain test), and just expect some entries, something along these lines:

```ts
await expect.poll(() => testServerConnection.events).toEqual(expect.arrayContaining([
  ['stdio', { type: 'stderr', text: 'this goes to stderr\n' }],
  ['stdio', { type: 'stdout', text: 'this goes to stdout\n' }],
]));
```

dgozman (approved):
Looks good!

PR #31846, opened 2024-07-24
chore: remove glob from client-certificate matching

mxschmitt:


packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -224,21 +224,11 @@ export class ClientCertificatesProxy {
...
-  const matchingCerts = clientCertificates?.filter(c => {
-    let regex: RegExp | undefined = (c as any)[kClientCertificatesGlobRegex];
-    if (!regex) {
-      regex = globToRegex(c.origin);
-      (c as any)[kClientCertificatesGlobRegex] = regex;
-    }
-    regex.lastIndex = 0;
-    return regex.test(origin);
-  });
+  const matchingCerts = clientCertificates?.filter(c => c.origin === origin);
```
mxschmitt, line 231: Thought: maybe we should do:

     clientCertificates?.filter(c => new URL(c.origin).origin === origin);

in order to be more user-friendly.

docs/src/api/params.md
```diff
@@ -523,7 +523,7 @@ Does not enforce fixed viewport, allows resizing window in the headed mode.
 
 ## context-option-clientCertificates
 - `clientCertificates` <[Array]<[Object]>>
-  - `origin` <[string]> Glob pattern to match against the request origin that the certificate is valid for.
+  - `origin` <[string]> Exact origin that the certificate is valid for.
```
dgozman, line 526:
```suggestion
  - `origin` <[string]> Exact origin that the certificate is valid for. Origin includes `https` protocol, a hostname and optionally a port.
```

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
dgozman: Yeah, sounds great, but you'll need to try/catch to avoid an unexpected error.

dgozman (approved):

PR #31852, opened 2024-07-25
feat(trace viewer): show `baseURL` in Metadata pane

### LINKED ISSUE #31847: [Feature]: Trace Viewer: add baseURL to meta data
### 🚀 Feature Request

show baseURL in the metadata tab in the trace viewer

### Example

Screenshot 2024-07-24 at 18 13 00

### Motivation

the baseURL that is set in the config is very hard to find, you need to click on before hooks, fixture:context, browser.newContext and then open the call tab. clicking only on metadata to see the baseURL would be much simpler

Skn0tt:
Resolves https://github.com/microsoft/playwright/issues/31847 by adding playwright config's `baseURL` value to the `context-options` trace event, and showing that in the Trace Viewer.

Because the added property is optional, I didn't increment the trace format version.

I've also considered pulling the `baseURL` from the existing `browser.newContext` step to get around modifying the trace format, but that felt pretty hacky.

https://github.com/user-attachments/assets/ecaef747-727d-4937-9ca3-1605ca9907b9

### DIFF
--- a/packages/trace-viewer/src/ui/metadataView.tsx
+++ b/packages/trace-viewer/src/ui/metadataView.tsx
@@ -28,7 +28,7 @@ export const MetadataView: React.FunctionComponent<{
     {model.baseURL && (
       <>
         <div className='call-section' style={{ paddingTop: 2 }}>Config</div>
-        <div className='call-line'>baseURL:<a className='call-value string' href={model.baseURL} title={model.baseURL} target="_blank">{model.baseURL}</a></div>
+        <div className='call-line'>baseURL:<a className='call-value string' href={model.baseURL} title={model.baseURL} target='_blank' rel='noopener noreferrer'>{model.baseURL}</a></div>
       </>
     )}
     <div className='call-section' style={{ paddingTop: 2 }}>Time</div>
--- a/packages/trace-viewer/src/ui/metadataView.tsx
+++ b/packages/trace-viewer/src/ui/metadataView.tsx
@@ -25,6 +25,12 @@ export const MetadataView: React.FunctionComponent<{
   if (!model)
     return <></>;
   return <div className='metadata-view vbox'>
+    {model.baseURL && (
+      <>
+        <div className='call-section' style={{ paddingTop: 2 }}>Time</div>
+        <div className='call-line'>baseURL:<a className='call-value string' href={model.baseURL} title={model.baseURL} target="_blank">{model.baseURL}</a></div>
+      </>
+    )}
     <div className='call-section' style={{ paddingTop: 2 }}>Time</div>
     {!!model.wallTime && <div className='call-line'>start time:<span className='call-value datetime' title={new Date(model.wallTime).toLocaleString()}>{new Date(model.wallTime).toLocaleString()}</span></div>}
     <div className='call-line'>duration:<span className='call-value number' title={msToString(model.endTime - model.startTime)}>{msToString(model.endTime - model.startTime)}</span></div>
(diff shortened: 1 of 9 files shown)


packages/trace/src/trace.ts
```diff
@@ -43,6 +43,7 @@ export type ContextCreatedTraceEvent = {
   options: BrowserContextEventOptions,
   sdkLanguage?: Language,
   testIdAttributeName?: string,
+  baseURL?: string,
```
dgozman, line 46: This value is already written to the trace as `ContextCreatedTraceEvent.options.baseURL`, we just need to add it to the type.

packages/trace-viewer/src/ui/metadataView.tsx
```diff
@@ -25,6 +25,12 @@ export const MetadataView: React.FunctionComponent<{
   if (!model)
     return <></>;
   return <div className='metadata-view vbox'>
+    {model.baseURL && (
+      <>
+        <div className='call-section' style={{ paddingTop: 2 }}>Time</div>
+        <div className='call-line'>baseURL:<a className='call-value string' href={model.baseURL} title={model.baseURL} target="_blank">{model.baseURL}</a></div>
```
dgozman, line 31:
```suggestion
        <div className='call-line'>baseURL:<a className='call-value string' href={model.baseURL} title={model.baseURL} target='_blank' rel='noopener noreferrer'>{model.baseURL}</a></div>
```

I wonder why lint does not argue against double quotes...

Skn0tt: Because there's a specail `jsx-quote` rule that we don't have enabled yet. Went ahead and fixed that in https://github.com/microsoft/playwright/pull/31855, so we don't risk breaking the space-time continuum.

packages/trace/src/trace.ts
Skn0tt: Nice! Done in https://github.com/microsoft/playwright/pull/31852/commits/c8f2d6bed6212aa2afb2612d6e5b5e1852f4d85b.

Skn0tt (commented):
Thanks! Addressed the feedback. I'll also see if I can write a little e2e test on this ^^

tests/library/trace-viewer.spec.ts
```diff
@@ -30,6 +30,10 @@ test.slow();
 
 let traceFile: string;
 
+test.use({
+  baseURL: 'https://example.com',
```
dgozman, line 34: Instead of affecting all tests, let's pass `baseURL` option to the `newContext()` call on the line 38.

packages/trace-viewer/src/ui/metadataView.tsx
```diff
@@ -25,6 +25,12 @@ export const MetadataView: React.FunctionComponent<{
   if (!model)
     return <></>;
   return <div className='metadata-view vbox'>
+    {model.baseURL && (
```
dgozman, line 28: I'd move the `Config` section between `Browser` and `Viewport`. WDYT?

dgozman (approved):

tests/library/trace-viewer.spec.ts
Skn0tt: I somehow thought that wouldn't work, but it did. done in https://github.com/microsoft/playwright/pull/31852/commits/d0eb8056554323735b6129dacb3ea176431ee4b4

packages/trace-viewer/src/ui/metadataView.tsx
Skn0tt: yes! done in https://github.com/microsoft/playwright/pull/31852/commits/742ae72fed4939a00e0a01962aa3ce316b3f5956

PR #31859, opened 2024-07-25
chore: use happy eyeballs for client-certificates

mxschmitt:


packages/playwright-core/src/utils/happy-eyeballs.ts
```diff
@@ -61,12 +62,46 @@ export async function createSocket(host: string, port: number): Promise<net.Sock
           reject(err);
         if (socket)
           resolve(socket);
-      }, /* useTLS */ false).catch(err => reject(err));
+      }, false).catch(err => reject(err));
```
mxschmitt, line 65:
```suggestion
      }, /* useTLS */ false).catch(err => reject(err));
```

packages/playwright-core/src/utils/happy-eyeballs.ts
```diff
@@ -66,7 +67,41 @@ export async function createSocket(host: string, port: number): Promise<net.Sock
...
-async function createConnectionAsync(options: http.ClientRequestArgs, oncreate: ((err: Error | null, socket?: net.Socket) => void) | undefined, useTLS: boolean) {
+export async function createTLSSocket(options: tls.ConnectionOptions): Promise<tls.TLSSocket> {
+  return new Promise((resolve, reject) => {
+    assert(options.host, 'host is required');
+    if (net.isIP(options.host)) {
+      const socket = tls.connect(options)
+      socket.on('connect', () => resolve(socket));
+      socket.on('error', error => reject(error));
+    } else {
+      createConnectionAsync(options, (err, socket) => {
```
dgozman, line 78: I wonder if `createConnectionAsync` should allow IP addresses and deal with them? This would remove the ip-specific logic from all other callers of it. Maybe in a follow up if you like that.

dgozman (approved):

PR #31865, opened 2024-07-25
chore: reduce file reads in client-certificates internal TLS server

mxschmitt:


packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -27,6 +27,11 @@ import { SocksProxy } from '../common/socksProxy';
 import type * as channels from '@protocol/channels';
 import { debugLogger } from '../utils/debugLogger';
 
+const dummyServerCerts: tls.TlsOptions = {
```
dgozman, line 30: Maybe do it lazily? Otherwise we read certs unnecessary for 99.9% of our users.

mxschmitt: Fair point!

dgozman (approved):

PR #31867, opened 2024-07-25
fix(client-certificates): return target errors on response when using http2

mxschmitt:


packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -139,7 +146,6 @@ class SocksProxyConnection {
       dummyServer.emit('connection', this.internal);
       dummyServer.on('secureConnection', internalTLS => {
         debugLogger.log('client-certificates', `Browser->Proxy ${this.host}:${this.port} chooses ALPN ${internalTLS.alpnProtocol}`);
-        internalTLS.on('close', () => this.socksProxy._socksProxy.sendSocketEnd({ uid: this.uid }));
```
mxschmitt, line 142: My assumption is that the 'close' event of the `tls.TLSSocket` will bubble down onto the `internal` `net.Socket`.

tests/library/client-certificates.spec.ts
```diff
@@ -25,6 +25,7 @@ const { createHttpsServer, createHttp2Server } = require('../../packages/playwri
 type TestOptions = {
   startCCServer(options?: {
     http2?: boolean;
+    enableHTTP1FallbackWhenUsingHttp2?: boolean;
```
mxschmitt, line 28: drive-by: more strict, enable it only in the test when we need it, line 308. This also makes the new test safer.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -169,20 +177,36 @@ class SocksProxyConnection {
         internalTLS.on('error', () => closeBothSockets());
         targetTLS.on('error', error => {
           debugLogger.log('client-certificates', `error when connecting to target: ${error.message}`);
+          const responseBody = 'Playwright client-certificate error: ' + error.message;
           if (internalTLS?.alpnProtocol === 'h2') {
-          // https://github.com/nodejs/node/issues/46152
-          // TODO: http2.performServerHandshake does not work here for some reason.
+            // This method is available only in Node.js 18+
```
mxschmitt, line 182:
```suggestion
            // This method is available only in Node.js 20+
```

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -169,20 +177,36 @@ class SocksProxyConnection {
...
+          const responseBody = 'Playwright client-certificate error: ' + error.message;
           if (internalTLS?.alpnProtocol === 'h2') {
-          // https://github.com/nodejs/node/issues/46152
-          // TODO: http2.performServerHandshake does not work here for some reason.
+            // This method is available only in Node.js 20+
+            if ('performServerHandshake' in http2) {
+              this._interceptClose = true;
+              // @ts-expect-error
+              const session = http2.performServerHandshake(internalTLS);
+              session.on('stream', (stream: http2.ServerHttp2Stream) => {
```
dgozman, line 187: Do we need an error handler on session/stream?

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -169,20 +177,36 @@ class SocksProxyConnection {
         internalTLS.on('error', () => closeBothSockets());
         targetTLS.on('error', error => {
           debugLogger.log('client-certificates', `error when connecting to target: ${error.message}`);
+          const responseBody = 'Playwright client-certificate error: ' + error.message;
           if (internalTLS?.alpnProtocol === 'h2') {
-          // https://github.com/nodejs/node/issues/46152
-          // TODO: http2.performServerHandshake does not work here for some reason.
+            // This method is available only in Node.js 20+
+            if ('performServerHandshake' in http2) {
+              this._interceptClose = true;
```
dgozman, line 184: Can we instead remove the close listener, and move the respective comment here?

dgozman (approved):

PR #31877, opened 2024-07-26
docs: explain a bit more about fixture boxing and custom titles

dgozman:


mxschmitt (approved):

docs/src/test-fixtures-js.md
```diff
@@ -675,27 +675,30 @@ test('passes', async ({ database, page, a11y }) => {
...
 
 export const test = base.extend({
-  _helperFixture: [async ({}, use, testInfo) => {
+  helperFixture: [async ({}, use, testInfo) => {
+    // ...
   }, { box: true }],
 });
 ```
 
+This is useful for non-interesting helper fixtures. For example, an auto fixture that sets up some common data can be safely hidden from a test report.
```
Skn0tt, line 690:
```suggestion
This is useful for non-interesting helper fixtures. For example, an [automatic](./test-fixtures.md#automatic-fixtures) fixture that sets up some common data can be safely hidden from a test report.
```

When I googled for "auto fixture playwright", I didn't really find anything. The name in the docs is "automatic fixture", so let's also use that here. And a link so I don't have to google next time :D

Screenshot 2024-07-26 at 16 00 10

The google results for "automatic fixture playwright" aren't a lot better, which suggests that some work on SEO is due

docs/src/test-fixtures-js.md
```diff
@@ -675,27 +675,30 @@ test('passes', async ({ database, page, a11y }) => {
...
+
 ## Custom fixture title
 
-You can assign a custom title to a fixture to be used in error messages and in the
-reporters UI:
+You can assign a custom title to a fixture that will be shown instead of the usual fixture name in various test reports and error messages.
```
Skn0tt, line 694:
```suggestion
Instead of the usual fixture name, you can give fixtures a custom title that will be shown in test reports and error messages.
```

Moving the inner sentence to the start reads a little easier to me.

PR #31889, opened 2024-07-27
docs(ct): fix component.update example for vue and svelte

sand4rt:
Svelte does not support JSX syntax & most people use the TS syntax in Vue


mxschmitt:
Is this related to https://github.com/microsoft/playwright/issues/31927?

PR #31898, opened 2024-07-29
fix(client-certificates): don't use proxy when using BrowserContext.request

mxschmitt:
Motivation: When using `context.request` the `fetch()` call should still use the native TLS options rather than going the proxy route. Before this patch, it used the `defaultOptions` which contain the socksProxy in the `proxy` options and it re-used this. this emitted a warning when visiting `127.0.0.1`, probably due to an upstream bug:

    (node:13882) [DEP0123] DeprecationWarning: Setting the TLS ServerName to an IP address is not permitted by RFC 6066. This will be ignored in a future version.

Probably a regression of https://github.com/TooTallNate/proxy-agents/issues/308.

So based on this it seems fine to ignore the given proxy settings in fetch, if clientCertificates are given. (We would error before when creating the browser context that both is not supported.)

The test changes are needed in order to trigger this warning, which does not get emitted anymore.

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -168,7 +168,7 @@ export abstract class APIRequestContext extends SdkObject {
     const method = params.method?.toUpperCase() || 'GET';
     const proxy = defaults.proxy;
     let agent;
-    if (proxy && proxy.server !== 'per-context' && !shouldBypassProxy(requestUrl, proxy.bypass)) {
+    if (proxy && !defaults.clientCertificates?.length && proxy.server !== 'per-context' && !shouldBypassProxy(requestUrl, proxy.bypass)) {
       const proxyOpts = url.parse(proxy.server);
       if (proxyOpts.protocol?.startsWith('socks')) {
         agent = new SocksProxyAgent({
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/fetch.ts
```diff
@@ -168,7 +168,7 @@ export abstract class APIRequestContext extends SdkObject {
     const method = params.method?.toUpperCase() || 'GET';
     const proxy = defaults.proxy;
     let agent;
-    if (proxy && proxy.server !== 'per-context' && !shouldBypassProxy(requestUrl, proxy.bypass)) {
+    if (proxy && !defaults.clientCertificates?.length && proxy.server !== 'per-context' && !shouldBypassProxy(requestUrl, proxy.bypass)) {
```
dgozman, line 171:
```suggestion
    // When `clientCertificates` is present, we set the `proxy` property to our own socks proxy
    // for the browser to use. However, we don't need it here, because we already respect
    // `clientCertificates` when fetching from Node.js.
    if (proxy && !defaults.clientCertificates?.length && proxy.server !== 'per-context' && !shouldBypassProxy(requestUrl, proxy.bypass)) {
```

dgozman (approved):

Issue #31863, opened 2024-07-25
[internal] improve react code in UI mode, trace viewer and html report 


dgozman:
Some ideas:
- [x] introduce something like `clsx` for easier class names;
- [x] use react `context` to avoid passing many constant props around;
- [x] improve settings binding between `localStorage` and `useState()`;
- [ ] ...

mxschmitt:
Would be probably also great to migrate away from [ReactDOM.render](https://github.com/search?q=repo%3Amicrosoft%2Fplaywright%20%2FReactDOM.render%2F&type=code) to [`createRoot`](https://react.dev/reference/react-dom/render) - since the `render` method is deprecated.

Skn0tt:
We've refactored some of the React code, introduced `clsx`, and updated all our React properties to use concurrent rendering. Considering this done!

mxschmitt:
Great work!

PR #31900, opened 2024-07-29
refactor(ui): some react refactorings

Skn0tt:
Addresses https://github.com/microsoft/playwright/issues/31863. This PR is chonky, but the individual commits should be easy to review. If they're not, i'm happy to break them out into individual PRs.

There's two main things this does:

1. Remove some unused imports
2. Add a `clsx`-inspired helper function for classname templating

I wasn't able to replace `ReactDOM.render` with `ReactDOM.createRoot`. This is the new recommended way starting with React 18, and the existing one is going to be deprecated at some point. But it somehow breaks our tests, i'll have to investigate that separately.

### DIFF
--- a/packages/html-reporter/src/chip.tsx
+++ b/packages/html-reporter/src/chip.tsx
@@ -19,6 +19,7 @@ import './chip.css';
 import './colors.css';
 import './common.css';
 import * as icons from './icons';
+import { clsx } from '@web/uiUtils';
 
 export const Chip: React.FC<{
   header: JSX.Element | string,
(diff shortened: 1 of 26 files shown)


Skn0tt:
Interestingly, it looks like the `createRoot().render()` change broke lots of tests. I've reverted it in https://github.com/microsoft/playwright/pull/31900/commits/3e198ceffaf8be1534e884c6c190834fa6428e4b, let's see what CI says.

dgozman (approved):

mxschmitt (approved):

PR #31911, opened 2024-07-30
refactor(ui): synchronize settings via useSyncExternalStore instead of prop drilling

Skn0tt:
Broken out from https://github.com/microsoft/playwright/pull/31900, part of https://github.com/microsoft/playwright/issues/31863.

Synchronizes different `useSettings` calls via `useSyncExternalStore`. This saves us from having to drill down settings props everywhere, without the big refactoring that a `Context` would be.

### DIFF
--- a/packages/web/src/components/splitView.tsx
+++ b/packages/web/src/components/splitView.tsx
@@ -38,8 +38,14 @@ export const SplitView: React.FC<React.PropsWithChildren<SplitViewProps>> = ({
   settingName,
   children
 }) => {
-  const [hSize, setHSize] = useSetting<number>((settingName ?? 'dont-persist') + '.' + orientation + ':size', Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio, undefined, !settingName);
-  const [vSize, setVSize] = useSetting<number>((settingName ?? 'dont-persist') + '.' + orientation + ':size', Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio, undefined, !settingName);
+  const defaultSize = Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio;
+  const hSetting = useSetting<number>((settingName ?? 'unused') + '.' + orientation + ':size', defaultSize);
+  const vSetting = useSetting<number>((settingName ?? 'unused') + '.' + orientation + ':size', defaultSize);
+  const hState = React.useState(defaultSize);
+  const vState = React.useState(defaultSize);
+  const [hSize, setHSize] = settingName ? hSetting : hState;
+  const [vSize, setVSize] = settingName ? vSetting : vState;
+
   const [resizing, setResizing] = React.useState<{ offset: number, size: number } | null>(null);
   const [measure, ref] = useMeasure<HTMLDivElement>();
 
--- a/packages/web/src/components/splitView.tsx
+++ b/packages/web/src/components/splitView.tsx
@@ -38,8 +38,8 @@ export const SplitView: React.FC<React.PropsWithChildren<SplitViewProps>> = ({
   settingName,
   children
 }) => {
-  const [hSize, setHSize] = useSetting<number>(settingName ? settingName + '.' + orientation + ':size' : undefined, Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio);
-  const [vSize, setVSize] = useSetting<number>(settingName ? settingName + '.' + orientation + ':size' : undefined, Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio);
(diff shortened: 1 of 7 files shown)


packages/web/src/components/splitView.tsx
```diff
@@ -38,8 +38,8 @@ export const SplitView: React.FC<React.PropsWithChildren<SplitViewProps>> = ({
   settingName,
   children
 }) => {
-  const [hSize, setHSize] = useSetting<number>(settingName ? settingName + '.' + orientation + ':size' : undefined, Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio);
-  const [vSize, setVSize] = useSetting<number>(settingName ? settingName + '.' + orientation + ':size' : undefined, Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio);
+  const [hSize, setHSize] = useSetting<number>((settingName ?? 'dont-persist') + '.' + orientation + ':size', Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio, undefined, !settingName);
```
dgozman, line 41: I kind of liked the previous api with passing `undefined` more. Can we revert that part?

packages/web/src/uiUtils.ts
```diff
@@ -141,42 +141,64 @@ export function copy(text: string) {
...
-    defaultValue = settings.getObject(name, defaultValue);
-  const [value, setValue] = React.useState<S>(defaultValue);
-  const setValueWrapper = React.useCallback((value: React.SetStateAction<S>) => {
-    if (name)
-      settings.setObject(name, value);
-    setValue(value);
-  }, [name, setValue]);
+export function useSetting<S>(name: string, defaultValue: S, title?: string, dontPersist?: boolean): [S, (v: S) => void, Setting<S>] {
+  const value = React.useSyncExternalStore(
+      function subscribe(callback) {
```
dgozman, line 146: React recommends passing the same `subscribe` function instance every time. Perhaps `useCallback` for this?

packages/web/src/uiUtils.ts
```diff
@@ -141,42 +141,64 @@ export function copy(text: string) {
...
     localStorage[name] = value;
+    this.onChangeEmitter.dispatchEvent(new Event(name));
     if ((window as any).saveSettings)
       (window as any).saveSettings();
   }
 
   getObject<T>(name: string, defaultValue: T): T {
     if (!localStorage[name])
-      return defaultValue;
+      return this.sessionSettings[name] ?? defaultValue;
```
dgozman, line 181: I think we should support `undefined` and `null` as setting values, so better make `sessionSettings` a map, and use `has()` method.

dgozman (commented):
I like this approach!

packages/web/src/uiUtils.ts
Skn0tt: done in 8d92f44bd1024dfec4e3c728516ec015eede5f1b

packages/web/src/components/splitView.tsx
Skn0tt: Not really, because we need a way of attributing the `useSetting` call with a setting name to store it inside `Settings#sessionSettings`. If you don't want this saved, we could make it a `useState` call instead?

Skn0tt: Implemented in https://github.com/microsoft/playwright/pull/31911/commits/00a1574737802e74464e116582b272eaf6b9f265.

packages/web/src/components/splitView.tsx
```diff
@@ -38,8 +38,14 @@ export const SplitView: React.FC<React.PropsWithChildren<SplitViewProps>> = ({
...
   children
 }) => {
-  const [hSize, setHSize] = useSetting<number>(settingName ? settingName + '.' + orientation + ':size' : undefined, Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio);
-  const [vSize, setVSize] = useSetting<number>(settingName ? settingName + '.' + orientation + ':size' : undefined, Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio);
+  const defaultSize = Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio;
+  const hSetting = useSetting<number>((settingName ?? 'unused') + '.' + orientation + ':size', defaultSize);
+  const vSetting = useSetting<number>((settingName ?? 'unused') + '.' + orientation + ':size', defaultSize);
+  const hState = React.useState(defaultSize);
+  const vState = React.useState(defaultSize);
+  const [hSize, setHSize] = settingName ? hSetting : hState;
```
dgozman, line 46: This works for me, but we could as well move this choice back into `useSetting()` helper if we ever need the conditional again. WDYT?

dgozman (approved):

packages/web/src/components/splitView.tsx
Skn0tt: Yes we could. We could even move it into a separate Hook that uses `useSetting`, so that the `useSetting` implementation isn't cluttered.

PR #31925, opened 2024-07-31
refactor(ui): in splitview component, move sidebar and main from children into named properties

Skn0tt:
Pulled out from https://github.com/microsoft/playwright/pull/31900

I stumbled over `React.Children`, because it's the first time I saw that used. https://react.dev/reference/react/Children lists `React.Children` it as "Legacy" and mentions it's uncommon. Also, the fact that SplitView only displays its first two children, and all others are silently discarded, can be a surprise to some.

By separating things out into `sidebar` and `main`, not only do we give the two elements names (otherwise one needs to remember that sidebar is always the first child), but we also prevent any "third children" from being dropped.

### DIFF
--- a/packages/trace-viewer/src/ui/uiModeView.tsx
+++ b/packages/trace-viewer/src/ui/uiModeView.tsx
@@ -458,84 +458,83 @@ export const UIModeView: React.FC<{}> = ({
           />
         </div>
       </div>}
-      sidebar={
-        <div className='vbox ui-mode-sidebar'>
-          <Toolbar noShadow={true} noMinHeight={true}>
-            <img src='playwright-logo.svg' alt='Playwright logo' />
-            <div className='section-title'>Playwright</div>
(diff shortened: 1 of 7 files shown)


packages/trace-viewer/src/ui/uiModeView.tsx
```diff
@@ -452,84 +457,86 @@ export const UIModeView: React.FC<{}> = ({
...
-          updateSnapshotsSetting,
-        ]} />}
-        <Toolbar noShadow={true} noMinHeight={true} className='settings-toolbar' onClick={() => setSettingsVisible(!settingsVisible)}>
-          <span
-            className={`codicon codicon-${settingsVisible ? 'chevron-down' : 'chevron-right'}`}
-            style={{ marginLeft: 5 }}
-            title={settingsVisible ? 'Hide Settings' : 'Show Settings'}
+      </div>}
+      sidebar={
+        <div className='vbox ui-mode-sidebar'>
```
dgozman, line 462: nit: if you merge this line with the previous one, the diff would be much much smaller because of no indent changes 😄

dgozman (approved):

Issue #31912, opened 2024-07-30
[Bug]: download attachements doesn't work in ui mode/vscode show trace


debs-obrien:
### Version

@latest

### Steps to reproduce

test('download should work', async ({ page }, testInfo) => {
    await page.goto('https://playwright.dev/docs/intro');
    const screenshot = await page.screenshot();
    expect(screenshot).toMatchSnapshot('landing-page.png');
  });

### Expected behavior

download should work

### Actual behavior

in ui mode and vs code extension the download shows[::1] instead of localhost. when you click download it downloads a long name with no .png extension so you have to manually add that in order to be able to open the image. trace viewer from html report works as expected

### Additional context

**Beautiful filename**
<img width="366" alt="Screenshot 2024-07-30 at 11 53 02" src="https://github.com/user-attachments/assets/87a10f1b-7c1c-4411-8d87-2d6fb9a3a38f">

**hovering over shows [::1]**
<img width="368" alt="Screenshot 2024-07-30 at 11 53 25" src="https://github.com/user-attachments/assets/2c767c0c-91e5-4491-a95e-d508bd85f6b5">

**manually add .png to open the file**
<img width="222" alt="Screenshot 2024-07-30 at 11 56 43" src="https://github.com/user-attachments/assets/30384c2d-2c93-439e-af7f-a759ec16a2c0">

### Environment

```shell
System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.05 GB / 64.00 GB
  Binaries:
    Node: 20.11.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.4 - /usr/local/bin/npm
    pnpm: 8.5.0 - /usr/local/bin/pnpm
  IDEs:
    VSCode: 1.91.1 - /usr/local/bin/code
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.47.0-alpha-2024-07-29 => 1.47.0-alpha-2024-07-29
```

PR #31928, opened 2024-07-31
fix(trace): make sure the correct attachment name is used for downloads

dgozman:
When two attachments have the same content sha1, we used the first one's name for the downloaded file, no matter which one the user clicked to download. Now we pass the name explicitly.

References #31912.

### DIFF
--- a/packages/trace-viewer/src/ui/attachmentsTab.tsx
+++ b/packages/trace-viewer/src/ui/attachmentsTab.tsx
@@ -50,7 +50,7 @@ const ExpandableAttachment: React.FunctionComponent<ExpandableAttachmentProps> =
   }, [expanded, attachmentText, placeholder, attachment]);
 
   const title = <span style={{ marginLeft: 5 }}>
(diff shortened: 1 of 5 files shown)


packages/trace-viewer/src/ui/attachmentsTab.tsx
```diff
@@ -139,3 +139,10 @@ function attachmentURL(attachment: Attachment) {
     return 'sha1/' + attachment.sha1 + '?trace=' + encodeURIComponent(attachment.traceUrl);
   return 'file?path=' + encodeURIComponent(attachment.path!);
 }
+
+function downloadURL(attachment: Attachment) {
+  const url = attachmentURL(attachment) + '&dn=' + encodeURIComponent(attachment.name);
```
mxschmitt, line 144: I'd love to see `attachmentURL` using `URLSearchParams` inside and accept a `Record<string, string>` of additional options.

mxschmitt (approved):

PR #31932, opened 2024-07-31
feat(test runner): `--tsconfig` cli option

### LINKED ISSUE #12829: [Feature] Allow to provide tsconfig path
Now there is no option that would allow you to specify which tsconfig to use when running playwright tests, I would like to have this option since my project uses several tsconfig's

```
npx playwright test --tsconfig mega.tsconfig.json
```

dgozman:
Introduce `--tsconfig` to specify a single config to be used for all imported files, instead of looking up tsconfig for each file separately.

Fixes #12829.


packages/playwright/src/common/configLoader.ts
```diff
@@ -111,6 +114,15 @@ export async function loadConfig(location: ConfigLocation, overrides?: ConfigCLI
...
     }
   }
+
+  // 3. Load transform options from the playwright config.
+  const babelPlugins = (userConfig as any)['@playwright/test']?.babelPlugins || [];
+  const external = userConfig.build?.external || [];
+  setTransformConfig({ babelPlugins, external });
+
+  // 4. Sync new transform options to ESM loader.
+  await configureESMLoader();
```
yury-s, line 124: can this be called just once (above) ?

yury-s: Or just a separate call to `await loaderChannel.send('setTransformConfig', { config: transformConfig() });` wrapped in a function?

yury-s (approved):

PR #31937, opened 2024-07-31
fix: display text attachments as <pre>, linkify them

yury-s:
Screenshot 2024-07-31 at 2 05 34 PM


pavelfeldman (commented):
Please keep the codemirror in place, it covers more scenarios than the linkified url.

PR #31938, opened 2024-07-31
chore(trace-viewer): do not shrink metadata view

yury-s:
Avoids the following effect:
image


packages/trace-viewer/src/ui/callTab.css
```diff
@@ -105,3 +105,7 @@ a.call-value:hover {
   padding: 5px;
   line-height: 17px;
 }
+
+.metadata-view {
```
pavelfeldman, line 109: Let's inline the style and remove this class. Migrate to data-test='metadata-view' for tests

pavelfeldman (approved):

PR #31941, opened 2024-07-31
feat(events): allow waiting for removeAllListeners

pavelfeldman:


packages/playwright-core/src/client/eventEmitter.ts
```diff
@@ -258,8 +301,6 @@ export class EventEmitter implements EventEmitterType {
       for (let i = listeners.length - 1; i >= 0; i--)
```
yury-s, line 301: don't  we need to copy the listeners array here?

yury-s (approved):

packages/playwright-core/src/client/eventEmitter.ts
pavelfeldman: We are removing them all synchronously and you are not supposed to interact with event emitter during the service 'removeListener' events...

PR #31953, opened 2024-08-01
refactor(test runner): send back `closeReason` from server

Skn0tt:
Currently, when the server closes a BrowerContext/Page (e.g. because the page crashed), there's no way for the server to pass back a `_closeReason`. Because of that, server-originating closes will always show the fallback reason `'Target page, context or browser has been closed'`.

This PR makes the server send back the close reason as part of a new `reason` parameter on the `close` event, allowing both client- and server-originating closes to have a specific reason.


dgozman (commented):
- Let's do the same for `Browser._closeReason`.
- Let's also plumb the reason from `onPipeClosed` in `BrowserType._connect`?

Skn0tt:
Okay, so some of the `BrowserType.connect` tests are still flaky. I'll need to take a look at that.

PR #31970, opened 2024-08-01
fix(trace): do not place expect into unfinished api calls based on time

yury-s:
Fixes https://github.com/microsoft/playwright/issues/31959

### DIFF
--- a/packages/playwright/src/worker/testInfo.ts
+++ b/packages/playwright/src/worker/testInfo.ts
@@ -33,7 +33,7 @@ export interface TestStepInternal {
   complete(result: { error?: Error, attachments?: Attachment[] }): void;
   stepId: string;
   title: string;
-  category: 'hook' | 'fixture' | 'test.step' | 'expect' | 'pw:api' | string;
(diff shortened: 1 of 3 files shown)


packages/playwright/src/worker/testInfo.ts
```diff
@@ -33,7 +33,7 @@ export interface TestStepInternal {
   complete(result: { error?: Error, attachments?: Attachment[] }): void;
   stepId: string;
   title: string;
-  category: 'hook' | 'fixture' | 'test.step' | 'expect' | string;
+  category: 'hook' | 'fixture' | 'test.step' | 'expect' | 'pw:api' | string;
```
dgozman, line 36: I think it's a layering violation to list `pw:api` here in the test runner. Instead, we can explicitly blocklist all other categories in line 255. WDYT?

dgozman (approved):

packages/playwright/src/worker/testInfo.ts
yury-s: Replaced with `canNestByTime` as discussed off the thread.

dgozman (approved):

PR #31978, opened 2024-08-02
chore(fetch): include response text into failOnStatusCode errors

### LINKED ISSUE #31834: [Feature]: Displaying Response Text for 'failOnStatusCode: true'
### 🚀 Feature Request

Dear Team,

I hope this message finds you well. I am writing to propose a new feature that I believe would enhance our debugging process.

**Feature Description:** I would like to request the ability to print the ‘Response Text’ when a request fails due to the property ‘failOnStatusCode: true’. This feature would provide valuable insights into the actual response message when a request fails, aiding in quicker issue resolution.

Here is a snippet of the current code:
```
const response = await request.post(`${process.env.AUTH_BASE_URL}/oauth/token?grant_type=client_credentials`, {
    headers: { Authorization: 'Basic ' + Buffer.from(`${process.envCLIENT_ID}:${process.env.CLIENT_SECRET}`).toString('base64') },
    failOnStatusCode: true,
});
```

**Current Behavior:** Currently, when using ‘await expect(response).toBeOK();’, the ‘Response Text’ is displayed upon failure. However, this does not occur when ‘failOnStatusCode: true’ is set, making it challenging to understand the cause of the failure.

**Proposed Solution:** Enable the display of ‘Response Text’ when ‘failOnStatusCode: true’ is set and a request fails. This would align the behavior with that of ‘await expect(response).toBeOK();’ and provide a more consistent debugging experience.

### Example

_No response_

### Motivation

I believe this enhancement would greatly improve our ability to troubleshoot issues.

yury-s:
Fixes https://github.com/microsoft/playwright/issues/31834

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -220,7 +220,7 @@ export abstract class APIRequestContext extends SdkObject {
         let text = fetchResponse.body.toString('utf8');
         if (text.length > 1000)
           text = text.substring(0, 997) + '...';
-        responseText = `\nResponse text:\n${text}`
+        responseText = `\nResponse text:\n${text}`;
       }
       throw new Error(`${fetchResponse.status} ${fetchResponse.statusText}${responseText}`);
     }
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

harig-invenco:
Thanks for this feature, exactly what we need ! Could you please confirm when this commit will be included in an upcoming release ?

mxschmitt:
@harig-invenco this will be part of v1.47.0 or you can use Canary here: https://playwright.dev/docs/canary-releases

PR #31986, opened 2024-08-02
fix(clock): mock time in Event.prototype.timeStamp

### LINKED ISSUE #31924: [Bug]: Clock mocking impacting page click events
### Version

1.45.3

### Steps to reproduce

1. Clone the repository at https://github.com/EwenDC/Playwright-Clock-Bug
2. Execute the command `npm install`
3. Run the two included Playwright tests using the `npm test` command.

### Expected behavior

The test "works without mock clock" should pass, and the test "fails with mock clock" should fail, as the only difference between the two tests is the use of `page.clock` (which should have no effect on the functionality of this demo app)

### Actual behavior

Both tests pass, as when the mocked clock is installed and time is fast forwarded, the select list within the dialog stops receiving click events.

### Additional context

This is a minimal reproduction based off a legacy Vue 2 application maintained at our company. We recently decided to use `page.clock` to speed up our end-to-end tests by allowing us to skip minutes of idle time waiting for frontend polling events. However, once we added `page.clock` to our tests, some of our tests started failing for reasons unrelated to the polling.

Namely, we had a test involving verifying the values shown in a select list within a dialog. Before adding a `clock.fastForward` call, this test worked fine. However, after adding the `clock.fastForward` call, the select list became totally unresponsive to click events. We also sporadically observed this behaviour in other parts of the UI (even before `clock.fastForward` was called), namely in icon buttons with tooltips. These buttons would fail to register clicks if you clicked directly on the icon, but would still register clicks against the button whitespace. I have included both listed scenarios in the reproduction, but unfortunately I was not able to get the failed icon button click to reliably trigger.

We observed these scenarios both within test executions, and when manually interacting with the test browsers in debug mode. Interestingly, I observed that when using debug mode, it was possible to get the select list to work by calling `.click()` on the `<div>` with the class `el-input` that surrounds the `<input>` element. This leads me to believe the root cause of this issue is page events not properly propagating when `page.clock` is in use, though I have no idea why that would be the case.

### Environment

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1 Pro
    Memory: 177.72 MB / 32.00 GB
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    npm: 10.7.0 - ~/.nvm/versions/node/v20.14.0/bin/npm
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: 1.45.3 => 1.45.3
```

mxschmitt:
Ideally we generate the timestamp when the Event gets created. This patch adds a best-effort logic, since we can't override the constructor of natively created events, e.g. `MouseEvent`.

Fixes https://github.com/microsoft/playwright/issues/31924

### DIFF
--- a/docs/src/clock.md
+++ b/docs/src/clock.md
@@ -31,6 +31,7 @@ The recommended approach is to use `setFixedTime` to set the time to a specific 
   - `requestIdleCallback`
   - `cancelIdleCallback`
   - `performance`
+  - `Event.timeStamp`
 :::
 
 ## Test with predefined time
--- a/packages/playwright-core/src/server/injected/clock.ts
+++ b/packages/playwright-core/src/server/injected/clock.ts
@@ -697,6 +697,14 @@ export function install(globalObject: WindowOrWorkerGlobalScope, config: Install
       (globalObject as any).Intl = api[method]!;
     } else if (method === 'performance') {
       (globalObject as any).performance = api[method]!;
+      const kEventTimeStamp = Symbol('playwrightEventTimeStamp');
+      Object.defineProperty(Event.prototype, 'timeStamp', {
+        get() {
+          if (!this[kEventTimeStamp])
+            this[kEventTimeStamp] = api.performance?.now();
+          return this[kEventTimeStamp];
+        }
+      });
     } else {
       (globalObject as any)[method] = (...args: any[]) => {
         return (api[method] as any).apply(api, args);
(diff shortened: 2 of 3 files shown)


dgozman (approved):

mhemmings:
@mxschmitt Is it possible to have this released in a patch release? It's causing big issues for us currently. Thanks!

mxschmitt:
@mhemmings we usually fix only regressions in patch releases - since this wasn't a regression, we didn't cherry pick it. I recommend to use [Canary](https://playwright.dev/docs/canary-releases) for now.

mhemmings:
@mxschmitt Thanks, Matt. We'll try the Canary build

PR #31990, opened 2024-08-02
chore(trace): do not nest API actions based on time

yury-s:
They should be properly nested based on Node.js zones now.


pavelfeldman (approved):

dgozman:
I think this might affect the html report where we do not have `hide route actions` setting yet, and might get a long list of `route.something()` calls.

yury-s:
Given that we collapse similar route calls into one in the html report, I believe this should be fine:

image

PR #31998, opened 2024-08-04
test: have typed playwright-core/* imports

mxschmitt:
We want to keep importing from lib/ internally, so tsconfig paths are not an option.

This adds a way for typed imports for our most used pw-core imports.


tests/global.d.ts
```diff
@@ -0,0 +1,39 @@
+declare module '@playwright-core/utils' {
```
dgozman, line 1: Please add a thorough comment explaining why this file and what's put in here.

dgozman (approved):
This looks good. However, we use a different technique to export typed values from bundles, and also have an `internalsForTest` file that is along the same lines. It would be great to have a single way of doing things.

pavelfeldman:
What's happening here?

dgozman:
Closing as per offline discussion. This add extra complexity and does not solve enough problems. For example, "find all usages" does not really work unless you open the test.

Issue #1799, opened 2020-04-15
[Feature] allow client certificate selection and settings from Javascript


frzme:
Similarly to https://github.com/puppeteer/puppeteer/issues/540 

Currently when navigating to a page that requires client certificates and client certificates are available a popup is shown in Firefox and Chrome which asks to select which certificate to use. It would be beneficial to provide an API to select the correct certificate to use (or use none).

yyvess:
You can apply same workaround as for puppeteer.
First client certificate popup should only open if you have more than one client certificate installed. 
You can remove client certificate to avoid the popup be open.

Secondly you can inject dynamically an certificate as is =>


```
const fs = require('fs');
const request = require('request');
const browser = await chromium.launch();
const ctx = await browser.newContext();
await ctx.route('**/*', (route, req) => {
    const options = {
      uri: req.url(),
      method: req.method(),
      headers: req.headers(),
      body: req.postDataBuffer(),
      timeout: 10000,
      followRedirect: false,
      agentOptions: {
        ca: fs.readFileSync('./certs/ca.pem'),
        pfx: fs.readFileSync('./certs/user_cert.p12'),
        passphrase: fs.readFileSync('./certs/user_cert.p12.pwd'),
      },
    };
    let firstTry = true;
    const handler = function handler(err, resp, data) {
      if (err) {
        /* Strange random connection error on first request, do one re-try */
        if (firstTry) {
          firstTry = false;
          return request(options, handler);
        }
        console.error(`Unable to call ${options.uri}`, err.code, err);
        return route.abort();
      } else {
        return route.fulfill({
          status: resp.statusCode,
          headers: resp.headers,
          body: data,
        });
      }
    };
    return request(options, handler);
  });
```

gepd:
> You can apply same workaround as for puppeteer.
> First client certificate popup should only open if you have more than one client certificate installed.
> You can remove client certificate to avoid the popup be open.

Hey @yyvess do you have an example of how remove client certificate programmatically?

yyvess:
Hi @gepd, as I know you cannot remove it programmatically. But if you have only one certificate installed, the browser should not ask you to select one. Note that the popup is not present when you run your test in background.

mxschmitt:
Hey folks!

We are currently evaluating this feature and have a few questions so it will fit your needs:
[...]

PR #32008, opened 2024-08-05
chore(client-certificates): rewrite error for unsupported PFX errors

mxschmitt:
Depends on https://github.com/microsoft/playwright/pull/32007
Relevant commit to review: 7b4c116f205bd2f64fab347e039456bf3e82f459

This should make https://github.com/microsoft/playwright/issues/1799#issuecomment-2256697635 more clear.


packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -297,3 +298,17 @@ export function clientCertificatesToTLSOptions(
 function rewriteToLocalhostIfNeeded(host: string): string {
   return host === 'local.playwright' ? 'localhost' : host;
 }
+
+export function rewriteOpenSSLErrorIfNeeded(error: Error): Error {
+  if (error.message !== 'unsupported')
```
mxschmitt, line 303: note: Unfortunately they don't set the `.code` property. Maybe we should look at the stack to make sure its an OpenSSL error?

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -150,8 +150,11 @@ class SocksProxyConnection {
         };
 
         const handleError = (error: Error) => {
-          debugLogger.log('client-certificates', `error when connecting to target: ${error.message}`);
-          const responseBody = 'Playwright client-certificate error: ' + error.message;
+          error = rewriteOpenSSLErrorIfNeeded(error);
+          debugLogger.log('client-certificates', `error when connecting to target: ${error.message.replaceAll('\n', ' ')}`);
+          const responseBody = escapeHTML('Playwright client-certificate error: ' + error.message)
+              .replaceAll('\n', ' <br>')
+              .replaceAll(/(https?:\/\/[^\s]+)/g, '<a href="$1" target="_blank">$1</a>');
```
dgozman, line 157: I wouldn't do this after `escapeHTML` for security reasons. And doing it before is a challenge. All in all, just not worth it.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
dgozman: I think it's fine as it is.

dgozman (approved):

PR #32009, opened 2024-08-05
feat(tracing): record canvas screenshots

### LINKED ISSUE #23964: [BUG] Canvas is not visible in the trace
### System info
- Playwright Version: [v1.35.1]
- Operating System: [macOS 13.4]
- Browser: [Chromium]
- Other info:

The content of HTML5 canvas cannot be seen in the trace, which makes debugging the tests more difficult. I am unsure if this is expected behavior or not. I would appreciate any information that can be shared on the topic.

Screenshot 2023-06-29 at 20 46 50

[trace.zip](https://github.com/microsoft/playwright/files/11908319/trace.zip)

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/23964

Adds contents of `canvas` elements to trace snapshots using [`canvas.toDataURL()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL).

Some open questions:

- Is it OK to store the image content in the HTML string? Should we put it into a file in the ZIP instead?
- In https://github.com/microsoft/playwright/issues/23964#issuecomment-1615264528, Pavel talks about making this opt-in. I have no idea how high the performance impact is, but doubt it's big enough to warrant making this optional
- WebGL needs `preserveDrawingBuffer: true` in the application code for this to work reliably. If you forget setting it, your traces will flaky, depending on wether the browse has an image buffer available at the time of `toDataURL`. Where do we document this? Is this the kind of thing that Three.js etc could detect by default?


pavelfeldman:
FYI: we try to keep the snapshot collection time reasonable (think 2ms-10ms). I don't think we can afford capturing dynamic content like this time and memory-wise.

Skn0tt:
That sounds like the answer to my question :D Then opt-in is the way to go. Let's discuss how to do that tonight.

Skn0tt:
What's curious is that depending on the mimetype specified in `toDataURL`, some engines are able to render contents even with `preserveDrawingBuffer: false`. Maybe we can do some smart detection to make use of this, and show the contents even without `preserveDrawingBuffer`?

Skn0tt:
Tried using `toBlob` instead of `toDataURL`, but most browser don't seem to like it.

Skn0tt:
Closing this since `toDataURL` doesn't get us screen content reliably. I've tested it across different usecases of `canvas`. Go.JS worked fine, but on Google Maps we're getting weird artifacts, and on Three.js we're only seeing the backdrop of the WebGL canvas. Not workable, sadly! We brainstormed some different ideas for solving this issue yesterday, continuing that discussion in the issue.

Issue #32004, opened 2024-08-05
[BUG] clientCertificates troubleshooting


mxschmitt:
> Update: This issue should be fixed in the latest [@canary](https://playwright.dev/docs/canary-releases). See #31906 for further discussions.

Hi @mxschmitt I have tried this with the beta release and testOptions but it doesn't works as expected, I am still getting the pop-up asking to select the certificate.

![image](https://github.com/user-attachments/assets/755803de-e046-4f8e-8760-d9fefd11b4ef)

here's my config:
![image](https://github.com/user-attachments/assets/35cef9f4-6b22-4943-b37d-a811f8f3f8bf)

package.json:
![image](https://github.com/user-attachments/assets/1d8cb3be-3dee-49f1-b6ac-6afb917775fa)

test file:
![image](https://github.com/user-attachments/assets/77bc5ebc-30c0-47c9-ba41-70b203abf517)

_Originally posted by @amitAutoS in https://github.com/microsoft/playwright/issues/1799#issuecomment-2268097809_

mxschmitt:
@amitAutoS posting it here, so that we don't send notifications to the others. I see you are on Windows, are you using Google Chrome by any chance?

Do you mind providing us your full playwright.config.ts?

amitAutoS:
> @amitAutoS posting it here, so that we don't send notifications to the others. I see you are on Windows, are you using Google Chrome by any chance?
> 
> Do you mind providing us your full playwright.config.ts?

sure, here it is:

import { defineConfig, devices } from "@playwright/test";
import { resolve } from "path";
import { readFileSync } from "fs";

/**
 * See https://playwright.dev/docs/test-configuration.
 */
// const pfxPaths = resolve(__dirname, "./certs/DavidJonesTest_G5.pfx");
// console.log("pfxPaths: " + pfxPaths);
export default defineConfig({
  testDir: "./e2e",
  /* Run tests in files in parallel */
  fullyParallel: true,
  /* Fail the build on CI if you accidentally left test.only in the source code. */
  forbidOnly: !!process.env.CI,
  /* Retry on CI only */
  retries: process.env.CI ? 2 : 0,
  /* Opt out of parallel tests on CI. */
  workers: process.env.CI ? 1 : undefined,
  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
  reporter: [["list"], ["html", { open: "never" }]],
  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
[...]

PR #32022, opened 2024-08-06
chore: add more client-certificates debug logging

mxschmitt:
Might help us to figure out https://github.com/microsoft/playwright/issues/32004.


dgozman (commented):
Let's wait with this one until we have something to troubleshoot, so that we can land the logs we think would be useful.

PR #32029, opened 2024-08-06
chore(ui): abstract useSetting/state combo into own hook

Skn0tt:
As discussed in the sync yesterday, here's a PR to abstract the useSetting/useState combo in `SplitView` into its own custom hook, moving the complexity from the callsite into the lib side of things.

### DIFF
--- a/packages/web/src/components/splitView.tsx
+++ b/packages/web/src/components/splitView.tsx
@@ -43,8 +43,8 @@ export const SplitView: React.FC<SplitViewProps> = ({
   main,
 }) => {
   const defaultSize = Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio;
-  const [hSize, setHSize] = useSettingOrState(settingName + '.' + orientation + ':size', defaultSize, !!settingName);
-  const [vSize, setVSize] = useSettingOrState(settingName + '.' + orientation + ':size', defaultSize, !!settingName);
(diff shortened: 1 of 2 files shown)


packages/web/src/components/splitView.tsx
```diff
@@ -43,12 +43,8 @@ export const SplitView: React.FC<SplitViewProps> = ({
...
 }) => {
   const defaultSize = Math.max(minSidebarSize, sidebarSize) * window.devicePixelRatio;
-  const hSetting = useSetting<number>((settingName ?? 'unused') + '.' + orientation + ':size', defaultSize);
-  const vSetting = useSetting<number>((settingName ?? 'unused') + '.' + orientation + ':size', defaultSize);
-  const hState = React.useState(defaultSize);
-  const vState = React.useState(defaultSize);
-  const [hSize, setHSize] = settingName ? hSetting : hState;
-  const [vSize, setVSize] = settingName ? vSetting : vState;
+  const [hSize, setHSize] = useSettingOrState(settingName ? `${settingName}.${orientation}:size` : undefined, defaultSize);
+  const [vSize, setVSize] = useSettingOrState(settingName ? `${settingName}.${orientation}:size` : undefined, defaultSize);
```
Skn0tt, line 47: There's a bug hidden in here, where both `hSize` and `vSize` will always point to the same setting. I'll write a fix in a separate PR.

yury-s (commented):
I believe this is addressed in https://github.com/microsoft/playwright/pull/32018

Skn0tt:
It is!

PR #32035, opened 2024-08-06
docs: route.fallback() vs. route.continue()

### LINKED ISSUE #31983: [Feature]: Set response delay in har mocks
### 🚀 Feature Request

When playwright records mocks in har I see that the prop `wait` is always -1 like so:
```
"timings": { "send": -1, "wait": -1, "receive": 1.388 }
```
Here is how I record har:
```
await oxy.page.routeFromHAR(
        path.resolve(__dirname, "har/test1.har"),
        {
            url: "https://example.com/**/*",
            updateContent: "embed",
            update: true,
        },
    );
```
And when I run my tests using these mocks those mocked requests take much less time than actual requests. Even if I manually change `wait` prop to 1000ms - it does not affect my mocked requests duration.

I tried to use page.route to add delay for my requests, but it seems like in this case those requests are not being mocked with my har files. Here is how I used page.route:
```
await oxy.page.routeFromHAR(
        path.resolve(__dirname, "har/test1.har"),
        {
            url: "https://example.com/**/*",
            updateContent: "embed",
            update: false,
        },
    );

await oxy.page.route("https://example.com/**/*", async (route) => {
        await new Promise((resolve) => setTimeout(resolve, 1000)); // Delay
         // This will make an actual network request, even if the request
         // should have been served from the HAR file
        await route.continue()
    });
```

### Example

- Har files have proper values in `wait` prop when recorded
- requests mocked with har files simulate response delay according to the `wait` prop in corresponding har mock

### Motivation

The ability to set response wait time allows to make mocked requests to their real examples in order to recreate some bugs that don't happen 100% of time.
Also according to the [documentation](http://www.softwareishard.com/blog/har-12-spec/) "The send, wait and receive timings are not optional and must have non-negative values"

yury-s:
Fixes https://github.com/microsoft/playwright/issues/31983

### DIFF
--- a/docs/src/api/class-route.md
+++ b/docs/src/api/class-route.md
@@ -39,7 +39,7 @@ Optional error code. Defaults to `failed`, could be one of the following:
   - alias-java: resume
   - alias-python: continue_
 
-Continues route's request with optional overrides.
+Sends route's request to the network with optional overrides.
 
 **Usage**
 
@@ -104,7 +104,7 @@ await page.RouteAsync("**/*", async route =>
 
 Note that any overrides such as [`option: url`] or [`option: headers`] only apply to the request being routed. If this request results in a redirect, overrides will not be applied to the new redirected request. If you want to propagate a header through redirects, use the combination of [`method: Route.fetch`] and [`method: Route.fulfill`] instead.
 
-[`method: Route.continue`] will immediately send the request to the netwok, other matching handlers won't be invoked. Use [`method: Route.fallback`] If you want next matching handler in the chain to be invoked.
+[`method: Route.continue`] will immediately send the request to the network, other matching handlers won't be invoked. Use [`method: Route.fallback`] If you want next matching handler in the chain to be invoked.
 
 ### option: Route.continue.url
 * since: v1.8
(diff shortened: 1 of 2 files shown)


docs/src/api/class-route.md
```diff
@@ -104,6 +104,8 @@ await page.RouteAsync("**/*", async route =>
 
 Note that any overrides such as [`option: url`] or [`option: headers`] only apply to the request being routed. If this request results in a redirect, overrides will not be applied to the new redirected request. If you want to propagate a header through redirects, use the combination of [`method: Route.fetch`] and [`method: Route.fulfill`] instead.
 
+[`method: Route.continue`] will immediately send the request to the netwok, other matching handlers won't be invoked. Use [`method: Route.fallback`] If you want next matching handler in the chain to be invoked.
```
dgozman, line 107: Perhaps do a symmetric comment in `Route.fallback` that says to call `continue`/`fulfill` to avoid any more handlers from the chain to be invoked?

dgozman (approved):

docs/src/api/class-route.md
pavelfeldman: netwok!

PR #32066, opened 2024-08-08
fix(types): revert type changes made to support TS 5.5

### LINKED ISSUE #32056: [Bug]: 1.46.0 (TypeScript) - custom fixtures `extend` no longer chainable?
### Version

1.46.0

### Steps to reproduce

1. Clone my repo at https://github.com/halljus/playwright-bug-repro
2. Run `npm install`
3. Run `npm run typecheck`
4. Observe TypeScript errors that parameters passed to a `test.extend` that is a child of an `extend` without a generic type parameter are typed as `any`.

### Expected behavior

I expect parameters passed to a child `extend<MyFixtures>` with a `test.extend` parent to be fully typed as they were prior to v1.46.0.

image

### Actual behavior

As of v1.46.0, any `extend<MyFixtures>` with a generic type parameter that follow a parent `test.extend` without a generic type parameter will receive untyped arguments (they'll be of type `any`).

image

### Additional context

I can "fix" this by moving the `extend` without the generic type parameter AFTER the `extend<MyFixtures>`, but a) it worked before, and b) I'm not sure if that's equivalent behavior.

### Environment

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 13th Gen Intel(R) Core(TM) i5-1345U
    Memory: 39.83 GB / 63.64 GB
  Binaries:
    Node: 20.13.1 - ~\AppData\Local\nvs\default\node.EXE
    npm: 10.8.2 - ~\AppData\Local\nvs\default\npm.CMD
  IDEs:
    VSCode: 1.92.0 - C:\Users\user\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 5.1.16 - C:\WINDOWS\system32\bash.EXE
  npmPackages:
    @playwright/test: 1.46.0 => 1.46.0
```

dgozman:
Regressed in #31532. The TS5.5 changes broke chaining of `extend`s where the first `extend` did not specify any type arguments.

Fixes #32056.

### DIFF
--- a/tests/library/global-fetch.spec.ts
+++ b/tests/library/global-fetch.spec.ts
@@ -21,7 +21,7 @@ import { expect, playwrightTest as base } from '../config/browserTest';
 import { kTargetClosedErrorMessage } from 'tests/config/errors';
 
 const it = base.extend({
-  context: async () => {
+  context: async ({}, use) => {
     throw new Error('global fetch tests should not use context');
   }
 });
--- a/packages/playwright/types/test.d.ts
+++ b/packages/playwright/types/test.d.ts
@@ -4815,9 +4815,9 @@ export type Fixtures<T extends KeyValue = {}, W extends KeyValue = {}, PT extend
 } & {
   [K in keyof PT]?: TestFixtureValue<PT[K], T & W & PT & PW> | [TestFixtureValue<PT[K], T & W & PT & PW>, { scope: 'test', timeout?: number | undefined, title?: string, box?: boolean }];
 } & {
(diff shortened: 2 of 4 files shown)


tests/library/global-fetch.spec.ts
```diff
@@ -21,7 +21,7 @@ import { expect, playwrightTest as base } from '../config/browserTest';
 import { kTargetClosedErrorMessage } from 'tests/config/errors';
 
 const it = base.extend({
-  context: async () => {
+  context: async ({}, use) => {
```
mxschmitt, line 24: Stray change?

mxschmitt (approved):

tests/library/global-fetch.spec.ts
dgozman: Nope, it did not compile without this!

PR #32078, opened 2024-08-08
fix(test runner): align with typescript behaviour for resolving `index.js` and `package.json` through path mapping

### LINKED ISSUE #31811: [Regression]: ESM Directory Import is not working anymore with Playwright's ESM Loader activated
### Last Good Version

1.42

### First Bad Version

1.43

### Steps to reproduce

Repo + Reproduction steps are in the repository: https://github.com/schickling-test/2024-playwright-esm-repro

### Expected behavior

Works

### Actual behavior

Error: Directory import '/Users/maxschmitt/Developer/tmp/2024-playwright-esm-repro/packages/@acme/lib' is not supported resolving ES modules imported from /Users/maxschmitt/Developer/tmp/2024-playwright-esm-repro/packages/@acme/tests/src/hello.test.ts

### Additional context

We run into this if for the directory import: https://github.com/microsoft/playwright/blob/d23ea26947d6b9644ddc92702796d7682d206fc6/packages/playwright/src/transform/esmLoader.ts#L30

and then pass the directory as a `file://` URL to the defaultResolver which breaks.

I throws here: https://github.com/nodejs/node/blob/e192a32c27f223e7a3dcb43134fd891dd8ded760/lib/internal/modules/esm/resolve.js#L254

Workaround, add a check like `!require('fs').statSync(resolved).isDirectory()`.

### Environment

```shell
N/A
```

Skn0tt:
Supercedes https://github.com/microsoft/playwright/pull/31915, closes https://github.com/microsoft/playwright/issues/31811.

When TypeScript resolves a specifier via path mapping, it does not interpret `package.json`. If path mapping resolves to a directory, it only looks at the `index.js` file in that directory if it's in CommonJS mode.

We need to mirror this in our `esmLoader.ts`.

### DIFF
--- a/packages/playwright/src/util.ts
+++ b/packages/playwright/src/util.ts
@@ -304,7 +304,7 @@ const kExtLookups = new Map([
   ['.mjs', ['.mts']],
   ['', ['.js', '.ts', '.jsx', '.tsx', '.cjs', '.mjs', '.cts', '.mts']],
 ]);
-export function resolveImportSpecifierExtension(resolved: string): string | undefined {
+export function resolveImportSpecifierExtension(resolved: string, isPathMapping: boolean, isESM: boolean): string | undefined {
   if (fileExists(resolved))
     return resolved;
 
@@ -321,12 +321,16 @@ export function resolveImportSpecifierExtension(resolved: string): string | unde
 
   if (dirExists(resolved)) {
     // If we import a package, let Node.js figure out the correct import based on package.json.
-    if (fileExists(path.join(resolved, 'package.json')))
+    // TypeScript does not interpret package.json for path mappings: https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths-should-not-point-to-monorepo-packages-or-node_modules-packages
+    if (!isPathMapping && fileExists(path.join(resolved, 'package.json')))
       return resolved;
 
-    // Otherwise, try to find a corresponding index file.
-    const dirImport = path.join(resolved, 'index');
-    return resolveImportSpecifierExtension(dirImport);
+    // Following TypeScript's path mapping logic, index files are still resolved in CommonJS.
+    const shouldNotResolveIndex = isPathMapping && isESM;
+    if (!shouldNotResolveIndex) {
+      const dirImport = path.join(resolved, 'index');
+      return resolveImportSpecifierExtension(dirImport, isPathMapping, isESM);
+    }
   }
 }
 
--- a/packages/playwright/src/util.ts
+++ b/packages/playwright/src/util.ts
@@ -319,18 +319,17 @@ export function resolveImportSpecifierExtension(resolved: string, isPathMapping:
     break;  // Do not try '' when a more specific extension like '.jsx' matched.
   }
 
-  if (dirExists(resolved)) {
+  // Following TypeScript's path mapping logic, index files and package.json are not resolved in ESM.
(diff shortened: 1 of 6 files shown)


packages/playwright/src/util.ts
```diff
@@ -319,14 +319,17 @@ export function resolveImportSpecifierExtension(resolved: string): string | unde
     break;  // Do not try '' when a more specific extension like '.jsx' matched.
   }
 
-  if (dirExists(resolved)) {
+  // Following TypeScript's path mapping logic, index files and package.json are not resolved in ESM.
+  // TypeScript does not interpret package.json for path mappings: https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths-should-not-point-to-monorepo-packages-or-node_modules-packages
+  const shouldNotResolveDirectory = isPathMapping && isESM;
+
+  if (!shouldNotResolveDirectory && dirExists(resolved)) {
```
dgozman, line 326: It seems like TypeScript:
- does not honor `exports` field from `package.json` after the type mapping;
- does honor the default `index.js` and `main` field from `package.json` after the type mapping as explained here: https://www.typescriptlang.org/docs/handbook/modules/reference.html#directory-modules-index-file-resolution. Note that both `index.js` and `main` are ignored for ESM, as explained here: https://nodejs.org/dist/latest-v20.x/docs/api/modules.html#folders-as-modules.

The docs above match my manual testing as well. I think we should follow TS and implement the same logic.

Skn0tt: I've added a test to ensure we follow your first observation in https://github.com/microsoft/playwright/pull/32078/commits/9a80386c7d057027499609aa4d648549482474fa.

For the second one, isn't that exactly the behaviour we have right now? If there's a `package.json` file, then we let Node.js deal with resolving to either `index.js` or `main`, depending on the ESM mode.

dgozman: - For the `exports` field, it looks like the code will defer to Node.js (when not in ESM), and Node will honor the `exports`. Is that not the case?
- For the `main` field, TypeScript also checks various extensions (e.g. `ts`) after following it, and we currently do not, because we leave it to Node.js. That's a minor thing, but it would be ideal to mirror TypeScript there as well.
- Updating comments to explain these things would be great as well!

Skn0tt: 1. I believe the test https://github.com/Skn0tt/playwright/blob/9a80386c7d057027499609aa4d648549482474fa/tests/playwright-test/resolver.spec.ts#L609-L644 asserts just that behaviour. Maybe i'm looking at it wrong though.
2. The `.ts` extension is covered by https://github.com/Skn0tt/playwright/blob/9a80386c7d057027499609aa4d648549482474fa/tests/playwright-test/resolver.spec.ts#L572-L607. What other extensions should we cover?

Both of these are about the path mapping case. Are you referring to the case where there's no path mapping?

In https://github.com/microsoft/playwright/pull/32078/commits/62ff3fd736429eb9f3e4038694e4d192944489f8, i've added a test to ensure that `main` is ignored in ESM mode, and updated the comments.

dgozman: Yep, the tests look like they cover all the cases. I guess for the item 1, Node.js does not look at `exports` either, which works for us. Thank you for following through!

dgozman (approved):

PR #32079, opened 2024-08-08
chore(ui): update to react 18

Skn0tt:
Part of https://github.com/microsoft/playwright/issues/31863. Updates most of our React usage to React 18. `recorder` doesn't seem to like it yet. I suspect that some of our code isn't compatible with concurrent mode, i've investigated that in https://github.com/microsoft/playwright/pull/32101.

### DIFF
--- a/packages/recorder/src/index.tsx
+++ b/packages/recorder/src/index.tsx
@@ -22,6 +22,6 @@ import { Main } from './main';
 
 (async () => {
   applyTheme();
-  // we'd like to migrate this to React 18, but concurrent mode seems to break some of our tests.
+  // TODO: we'd like to migrate this to React 18, but concurrent mode seems to break some of our tests.
   ReactDOM.render(<Main/>, document.querySelector('#root'));
 })();
--- a/packages/recorder/src/index.tsx
+++ b/packages/recorder/src/index.tsx
@@ -17,11 +17,11 @@
 import '@web/common.css';
 import { applyTheme } from '@web/theme';
 import '@web/third_party/vscode/codicon.css';
-import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import { Main } from './main';
 
 (async () => {
   applyTheme();
+  // we'd like to migrate this to React 18, but concurrent mode seems to break some of our tests.
   ReactDOM.render(<Main/>, document.querySelector('#root'));
 })();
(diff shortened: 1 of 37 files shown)


packages/recorder/src/index.tsx
```diff
@@ -17,11 +17,11 @@
 import '@web/common.css';
 import { applyTheme } from '@web/theme';
 import '@web/third_party/vscode/codicon.css';
-import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import { Main } from './main';
 
 (async () => {
   applyTheme();
+  // we'd like to migrate this to React 18, but concurrent mode seems to break some of our tests.
```
mxschmitt, line 25:
```suggestion
  // TODO: we'd like to migrate this to React 18, but concurrent mode seems to break some of our tests.
```

then its easier to grep in the future.

Skn0tt:
I double-checked that `recorder` isn't imported in the UI or any other part that we update to React 18, so landing this PR before fixing the concurrent mode changes should be safe.

mxschmitt (approved):

Skn0tt:
Oh no, unanticipated failures! I think they come from a subtle change in when state changes are applied in Concurrent React.

Before, each `setState` led to its own individual rerender and execution of `useMemo`. Now, the `setState` calls are batched together differently. This means that these two state changes now result in a single state change, rather than two:

https://github.com/Skn0tt/playwright/blob/3711474682e7492fc85264922224aa75cf4355a1/packages/trace-viewer/src/ui/uiModeView.tsx#L322-L323

This has the unfortunate side effect that now, the render where we get a finished `testModel`, we also get an undefined `runningState`, so the selection isn't executed anymore: https://github.com/Skn0tt/playwright/blob/3711474682e7492fc85264922224aa75cf4355a1/packages/trace-viewer/src/ui/uiModeTestListView.tsx#L70-L71

Before the React 18 update, we got lucky that this worked. Apparently now they slightly changed their rendering so that it doesn't anymore. A hacky fix would be to put the second `setState` into a `setTimeout`, but i'll try to find a nicer solution.

Skn0tt:
Turns out that React 18 didn't introduce the bug, it just makes it happen more often! My fix made a test fail that incorrectly asserted the failed test *not* to be selected, which is not the intended behaviour. https://github.com/microsoft/playwright/pull/32079/commits/e4f4dfc4096452d8126d10fccf579ec7d832a691 updates that assertion.

mxschmitt (approved):

PR #32094, opened 2024-08-09
fix(test runner): run project dependencies of `--only-changed` test files

### LINKED ISSUE #32070: [Bug]: --only-changed flag and project dependencies
### Version

1.46.0

### Steps to reproduce

1. Create a repo with multiple projects
2. Add dependencies to the project
3. Modify some of the test files
4. Run test with --only-changed flag

### Expected behavior

The dependencies are run before the actual modified tests.

### Actual behavior

Only the modified test are run, which fail without the dependencies.

### Additional context

I have multiple projects with a login setup as dependency

### Environment

```shell
System:
    OS: macOS 14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 15.41 MB / 16.00 GB
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.40.1 => 1.46.0
```

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32070. We were applying `additionalFileMatcher` not just to `filteredProjectSuites`, but also to `projectSuites`.  `projectSuites` is where we take dependency projects from, though - so `--only-changed` led to empty dependency projects, resulting in the reported bug.

The fix is to only apply `additionalFileMatcher` on `filteredProjectSuites`.

### DIFF
--- a/packages/playwright/src/runner/loadUtils.ts
+++ b/packages/playwright/src/runner/loadUtils.ts
@@ -123,6 +123,7 @@ export async function createRootSuite(testRun: TestRun, errors: TestError[], sho
   const config = testRun.config;
   // Create root suite, where each child will be a project suite with cloned file suites inside it.
   const rootSuite = new Suite('', 'root');
+  const unfilteredProjectSuites = new Map<FullProjectInternal, Suite>();
   const projectSuites = new Map<FullProjectInternal, Suite>();
   const filteredProjectSuites = new Map<FullProjectInternal, Suite>();
 
@@ -136,6 +137,7 @@ export async function createRootSuite(testRun: TestRun, errors: TestError[], sho
 
     // Filter file suites for all projects.
     for (const [project, fileSuites] of testRun.projectSuites) {
+      unfilteredProjectSuites.set(project, createProjectSuite(project, fileSuites));
       const filteredFileSuites = additionalFileMatcher ? fileSuites.filter(fileSuite => additionalFileMatcher(fileSuite.location!.file)) : fileSuites;
       const projectSuite = createProjectSuite(project, filteredFileSuites);
       projectSuites.set(project, projectSuite);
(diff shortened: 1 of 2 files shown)


packages/playwright/src/runner/loadUtils.ts
```diff
@@ -200,8 +200,8 @@ export async function createRootSuite(testRun: TestRun, errors: TestError[], sho
     const projectClosure = new Map(buildProjectsClosure(rootSuite.suites.map(suite => suite._fullProject!)));
 
     // Clone file suites for dependency projects.
-    for (const project of projectClosure.keys()) {
-      if (projectClosure.get(project) === 'dependency')
+    for (const [project, level] of projectClosure.entries()) {
```
Skn0tt, line 203: just a driveby refactoring

packages/playwright/src/runner/loadUtils.ts
```diff
@@ -136,10 +136,10 @@ export async function createRootSuite(testRun: TestRun, errors: TestError[], sho
 
     // Filter file suites for all projects.
     for (const [project, fileSuites] of testRun.projectSuites) {
+      projectSuites.set(project, createProjectSuite(project, fileSuites));
```
dgozman, line 139: Can we avoid calling `createProjectSuite` twice? It does a deep clone, so I'd like to not call it more than necessary. We can probably move the `additionalFileMatcher` to be an option of `filterProjectSuite` to achieve that.

Skn0tt: sure thing! done uin https://github.com/microsoft/playwright/pull/32094/commits/1899d2745a6d856da518634aa1e8fe0a08cc5344

dgozman (approved):

PR #32101, opened 2024-08-09
chore(recorder): update to React 18

Skn0tt:
Part of https://github.com/microsoft/playwright/issues/31863. Updates `recorder` to use React 18.

### DIFF
--- a/packages/recorder/src/main.tsx
+++ b/packages/recorder/src/main.tsx
@@ -30,12 +30,14 @@ export const Main: React.FC = ({
   window.playwrightSetSources = setSources;
   window.playwrightSetPaused = setPaused;
   window.playwrightUpdateLogs = callLogs => {
-    const newLog = new Map<string, CallLog>(log);
-    for (const callLog of callLogs) {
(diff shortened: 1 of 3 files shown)


packages/recorder/src/main.tsx
```diff
@@ -30,12 +30,14 @@ export const Main: React.FC = ({
   window.playwrightSetSources = setSources;
   window.playwrightSetPaused = setPaused;
   window.playwrightUpdateLogs = callLogs => {
-    const newLog = new Map<string, CallLog>(log);
```
Skn0tt, line 33: this closes over `log` from the outer scope. If `setLog` is called later, this value won't change. Before Concurrent React, this apparently wasn't a problem because the `Main` component was rerendered often enough. It became a problem with Concurrent React though. The fix is to use an updater function so we always update based on the current version of the state.

yury-s (approved):

packages/recorder/src/main.tsx
yury-s: Thanks for the great description!

PR #32117, opened 2024-08-12
chore: less 'as any' in html-reporter

mxschmitt:


packages/web/src/uiUtils.ts
```diff
@@ -175,8 +181,8 @@ export class Settings {
   setString(name: string, value: string) {
     localStorage[name] = value;
     this.onChangeEmitter.dispatchEvent(new Event(name));
-    if ((window as any).saveSettings)
-      (window as any).saveSettings();
+    if (window.saveSettings)
+      window.saveSettings();
```
Skn0tt, line 185:
```suggestion
    window.saveSettings?.();
```

packages/web/src/uiUtils.ts
```diff
@@ -193,8 +199,8 @@ export class Settings {
     localStorage[name] = JSON.stringify(value);
     this.onChangeEmitter.dispatchEvent(new Event(name));
 
-    if ((window as any).saveSettings)
-      (window as any).saveSettings();
+    if (window.saveSettings)
+      window.saveSettings();
```
Skn0tt, line 203:
```suggestion
    window.saveSettings?.();
```

packages/web/src/uiUtils.ts
```diff
@@ -165,6 +165,12 @@ export function useSetting<S>(name: string | undefined, defaultValue: S, title?:
   return [value, setValueWrapper, setting];
 }
 
+declare global {
+  interface Window {
+    saveSettings(): Promise<void>;
```
Skn0tt, line 170:
```suggestion
    saveSettings?(): Promise<void>;
```

A couple lines down, you're checking if this is defined. That means that this should be optional, probably?

Same in `recorderApp`.

packages/playwright-core/src/server/recorder/recorderApp.ts
```diff
@@ -38,7 +38,7 @@ declare global {
     playwrightSetSelector: (selector: string, focus?: boolean) => void;
     playwrightUpdateLogs: (callLogs: CallLog[]) => void;
     dispatch(data: EventData): Promise<void>;
-    saveSettings(data: any): Promise<void>;
+    saveSettings(): Promise<void>;
```
Skn0tt, line 41:
```suggestion
    saveSettings?(): Promise<void>;
```

Skn0tt (approved):
Me likey!

PR #32129, opened 2024-08-13
chore: move output parsing to concrete browsers

yury-s:


packages/playwright-core/src/server/chromium/chromium.ts
```diff
@@ -349,6 +350,28 @@ export class Chromium extends BrowserType {
...
+class ChromiumReadyState implements BrowserReadyState {
+  private readonly _wsEndpoint = new ManualPromise<string>();
+
+  onBrowserOutput(message: string): void {
+    const match = message.match(/DevTools listening on (.*)/);
+    if (match)
+      this._wsEndpoint.resolve(match[1]);
+  }
+  onBrowserExit(): void {
+    this._wsEndpoint.reject(new Error('Browser exited'));
```
dgozman, line 370: I wonder why does this one reject, and Juggler's one resolves?

packages/playwright-core/src/server/browserType.ts
```diff
@@ -38,12 +38,18 @@ import { helper } from './helper';
...
 import { type ProtocolError, isProtocolError } from './protocolError';
 
 export const kNoXServerRunningError = 'Looks like you launched a headed browser without having a XServer running.\n' +
   'Set either \'headless: true\' or use \'xvfb-run <your-playwright-app>\' before running Playwright.\n\n<3 Playwright Team';
 
+
+export interface BrowserReadyState {
+  onBrowserOutput(message: string): void;
+  onBrowserExit(): void;
+  ready(): Promise<string|undefined>;
```
dgozman, line 50: I'd make this a bit more verbose:

```suggestion
  waitUntilReady(): Promise<{ wsEndpoint?: string }>;
```

dgozman (approved):

packages/playwright-core/src/server/chromium/chromium.ts
yury-s: Let's have both reject.

Issue #32076, opened 2024-08-08
Rewrite `PWTEST_WATCH` as consumer of TestServer


Skn0tt:
Lots of folks are asking for a CLI watch mode, see https://github.com/microsoft/playwright/issues/21960. We're interested in landing it, but for maintainability reasons first we need to rewrite it onto the existing `TestServer`.

PR #32132, opened 2024-08-13
chore(ui): separate transport from `TestServerConnection`

Skn0tt:
Preparation for https://github.com/microsoft/playwright/issues/32076. We'd like to use `TestServer` for the implementation of `PWTEST_WATCH`, where there's no need for a full-blown WebSocket. By separating the transport from `TestServerConnection`, we can use an in-memory transport instead.


Skn0tt:
i'll work on this a little more and then extract the refactoring once i'm happy

PR #32139, opened 2024-08-13
Export MakeMatchers type

holvi-sebastian:
MakeMatchers is a useful type for library authors who which to provide fluent APIs for interacting with Playwright's expect assertions. Without this type, library authors can't publish type definitions, since the inferred return type of 'expect' becomes too large to serialize into a .d.ts file.

I'm not certain if this is the correct place to expose this type, if there are upstream symbols that need to be exported, please let me know and I'll make the changes!

### DIFF
--- a/packages/playwright/types/test.d.ts
+++ b/packages/playwright/types/test.d.ts
@@ -6548,7 +6548,7 @@ export type MatcherReturnType = {
   log?: string[];
 };
 
-type MakeMatchers<R, T, ExtendedMatchers> = {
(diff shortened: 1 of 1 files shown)


pavelfeldman:
Please start with filing a feature request.

holvi-sebastian:
Hi! I've filed a feature request here: https://github.com/microsoft/playwright/issues/32284

Sorry about that, I didn't think that exporting types warranted a feature request...

Thanks for answering!

PR #32156, opened 2024-08-14
chore(test runner): rebase watch mode onto TestServerConnection

### LINKED ISSUE #32076: Rewrite `PWTEST_WATCH` as consumer of TestServer

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32076.

This PR rewrites `watchMode.ts` to use `TestServer` under the hood. It's essentially a complete rewrite, so don't pay too much attention on the old implementation. Note that there's no changes to tests, so all behaviour we have specced out there still works.

To make this work without a superfluous WebSocket connection, I had to refactor `TestServerConnection` a little. Originally, I pulled this into a [separate PR](https://github.com/microsoft/playwright/pull/32132), but then realised how small the refactoring is. So it's in this PR now. Let me know if you'd like to land it separately.


packages/playwright/src/runner/watchMode.ts
```diff
@@ -207,37 +202,32 @@ export async function runWatchModeLoop(config: FullConfigInternal): Promise<Full
...
-        await runTests(config, failedTestIdCollector, { title: 're-running tests' });
-        config.testIdMatcher = undefined;
+        await runTests(config, testServerConnection, { title: 're-running tests', testIds: [...lastRun.failedTestIds!] });
       }
       continue;
     }
 
     if (command === 'toggle-show-browser') {
-      await toggleShowBrowser(config, originalWorkers);
+      await toggleShowBrowser();
```
Skn0tt, line 230: this is entirely untested, and ideally i'd like to add a test for it - but i'm not sure how to do it best. I tried writing a demo page that reads from `localStorage` across the two test runs, but that gets cleared. how could we assert that a browser is reused?

Skn0tt:
Interesting, those tests all pass locally for me! Looks like i'll need to dig deeper.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -19,25 +19,77 @@ import { createGuid, getPackageManagerExecCommand, ManualPromise } from 'playwri
...
+  const originalStderrWrite = process.stderr.write;
+  testServerConnection.onStdio(chunk => {
+    const text = chunk.text ?? Buffer.from(chunk.buffer!, 'base64');
+    if (chunk.type === 'stderr')
+      originalStderrWrite.call(process.stderr, text);
+    else
+      originalStdoutWrite.call(process.stdout, text);
+  });
+
+  await testServerConnection.initialize({ interceptStdio: true });
```
dgozman, line 86: Why do we want to intercept stdio, when we immediately write it back?

packages/playwright/src/runner/watchMode.ts
```diff
@@ -19,25 +19,77 @@ import { createGuid, getPackageManagerExecCommand, ManualPromise } from 'playwri
...
+  constructor(public readonly send: (data: any) => void, public readonly close: () => void = () => {}) {
+    super();
+  }
+
+  addEventListener(event: string, listener: (e: any) => void) {
+    this.addListener(event, listener);
+  }
+}
+
+export async function runWatchModeLoop(config: FullConfigInternal): Promise<FullResult['status']> {
```
dgozman, line 42: I think we should not pass `FullConfigInternal` to this method. Instead, the `TestServer` should load the config, and we should operate on some kind of options, maybe even on `ConfigCLIOverrides`. See how `program.ts` calls `testServer.runUIMode` instead of loading the config.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -19,25 +19,77 @@ import { createGuid, getPackageManagerExecCommand, ManualPromise } from 'playwri
...
+      originalStderrWrite.call(process.stderr, text);
+    else
+      originalStdoutWrite.call(process.stdout, text);
+  });
+
+  await testServerConnection.initialize({ interceptStdio: true });
+  await testServerConnection.runGlobalSetup({});
+
+  await testServerConnection.listTests({});
+  await testServerConnection.watch({ fileNames: [] });
```
dgozman, line 90: It seems like watch mode and ui mode have different ideas about watch semantics. Let's write them down and figure out whether we want to unify them or maybe update test server to support both, or something else.

packages/playwright/src/runner/watchMode.ts
dgozman: I'd not worry about testing this piece for now.

packages/playwright/src/runner/watchMode.ts
Skn0tt: The big difference I see is that in `--watch`, there's no way of selecting individual parts of the test tree to be watched. I also don't think that'd be ergonomic in the terminal. Are there any other differences that we want to unify?

packages/playwright/src/runner/watchMode.ts
Skn0tt: done!

packages/playwright/src/runner/watchMode.ts
Skn0tt: good call, removed that.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -19,25 +19,77 @@ import { createGuid, getPackageManagerExecCommand, ManualPromise } from 'playwri
...
+import { EventEmitter } from 'stream';
+import { type TestServerSocket, TestServerConnection } from '../isomorphic/testServerConnection';
 
-class FSWatcher {
-  private _dirtyTestFiles = new Map<FullProjectInternal, Set<string>>();
-  private _notifyDirtyFiles: (() => void) | undefined;
-  private _watcher: CFSWatcher | undefined;
-  private _timer: NodeJS.Timeout | undefined;
+class InMemoryServerSocket extends EventEmitter implements TestServerSocket {
+  constructor(public readonly send: (data: any) => void, public readonly close: () => void = () => {}) {
```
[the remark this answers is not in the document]
Skn0tt: aye aye!

packages/playwright/src/runner/testServer.ts
```diff
@@ -392,7 +392,7 @@ class TestServerDispatcher implements TestServerInterface {
     await this._testRun?.run;
   }
 
-  async _setInterceptStdio(intercept: boolean) {
+  _setInterceptStdio(intercept: boolean) {
```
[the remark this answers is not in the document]
Skn0tt: 🫡

packages/playwright/src/isomorphic/testServerConnection.ts
```diff
@@ -200,8 +202,14 @@ export class TestServerConnection implements TestServerInterface, TestServerInte
...
   close() {
     try {
-      this._ws.close();
+      this._socket.close();
     } catch {
     }
   }
 }
+
+export class WSTestServerConnection extends TestServerConnection {
```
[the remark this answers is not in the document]
Skn0tt: done

packages/playwright/src/isomorphic/testServerConnection.ts
```diff
@@ -19,6 +19,8 @@ import * as events from './events';
 
 // -- Reuse boundary -- Everything below this line is reused in the vscode extension.
 
+export type TestServerSocket = Pick<WebSocket, 'addEventListener' | 'send' | 'close'>;
```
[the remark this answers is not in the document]
Skn0tt: good idea! made it explicit for now.

Skn0tt (commented):
thanks, that's good feedback! I've addressed it, and now that `config` isn't passed in anymore this starts to look more and more like the UI implementation.

packages/playwright/src/runner/watchMode.ts
dgozman: - UI mode watches a specific list of files, and passes all test files when "Watch All" mode is enabled, while this watch mode does not "watch" any files.
- UI mode does not run dependencies upon a watched test run, while this mode does.
- This mode applies filters like grep after being notified about changed files, while UI mode applies filters before passing the list of files to watch (IIUC).

packages/playwright/src/isomorphic/testServerInterface.ts
```diff
@@ -81,13 +81,15 @@ export interface TestServerInterface {
     locations?: string[];
     grep?: string;
     grepInvert?: string;
+    onlyChanged?: string;
```
dgozman, line 84: Let's make `onlyChanged`-related changes in a separate PR? IIUC, we do not support `onlyChanged` in UI mode as well?

packages/playwright/src/runner/watchMode.ts
```diff
@@ -15,130 +15,115 @@
...
-    this._dirtyTestFiles = new Map();
-    return result;
-  }
-}
+    onDirtyTestFiles.resolve?.();
+  });
+  testServerConnection.onReport(report => {
+    if (report.method === 'onProject') {
+      const { name } = report.params.project;
+      projectNames.push(name);
```
dgozman, line 98: We seems to push into this array over and over again, on each test run, don't we?

packages/playwright/src/isomorphic/testServerInterface.ts
```diff
@@ -81,13 +81,15 @@ export interface TestServerInterface {
...
   }): Promise<{
     report: ReportEntry[],
     status: reporterTypes.FullResult['status']
   }>;
 
   runTests(params: {
     locations?: string[];
+    files?: string[]; // todo: find a better name. this is like locations, without pattern matching and without line numbers.
```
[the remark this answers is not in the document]
dgozman: I think we should not need this if we align with UI mode.

packages/playwright/src/isomorphic/testServerInterface.ts
Skn0tt: Happy to move it into a different PR if you want. The reason I included it here is that we supported onlyChanged in the previous impl of watch mode, and had a test for it: https://github.com/Skn0tt/playwright/blob/e8ad5cac6f9071293297287053c4c5313c65b1d2/tests/playwright-test/only-changed.spec.ts#L169

packages/playwright/src/runner/watchMode.ts
Skn0tt: We do! We could either clear it on every new test run, or make it a `Set`. I've turned it into a `Set` in 05ea3955564ef4f46b9a2d3c6f100b465fc867e9.

packages/playwright/src/runner/watchMode.ts
Skn0tt: - In the current implementation, I don't see the UI calling `watch` at all. It calls `initialize` once with `watchTestDirs: true`, but apart from that it never changes the list of watched files. I made this implementation do the same in c6d72fe2a1e061acdfee64a59eea9c61d3b4208b.

- UI mode does run dependencies, see 184b2b03eb4c7e02d52824546efa39b2cdbf6784 for a test that proves that.

- UI mode applies the filters after being notified as well. It does that in https://github.com/Skn0tt/playwright/blob/184b2b03eb4c7e02d52824546efa39b2cdbf6784/packages/trace-viewer/src/ui/uiModeView.tsx#L352-L374.

packages/playwright/src/isomorphic/testServerInterface.ts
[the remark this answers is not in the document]
Skn0tt: If we align with UI mode's way of passing an exhaustive list of test IDs, then we wouldn't need it. In order to do so, it needs to walk the entire test tree though. That's complex. It's fine in the UI, because it allows watching only select tests. We don't need that functionality in CLI watch mode though.

We still need something to filter down the test run to dirty files, and this `files` property is an easy-to-understand way of doing that.

dgozman:
Summary of the offline discussion:
- remove `onlyChanged` support, emit the same warning as in UI mode;
- remove `files` option, intersect with `locations` on the caller side;
- consider switching to `TeleSuiteUpdater` and `testId` filter in a follow up;
- collect `projectNames` from the first `listTests()` report.

Skn0tt (commented):
I've implemented the discussed changes. `TeleSuiteUpdater` worked really well! Haven't done the switch to `testId` yet, but with the TeleSuiteUpdater already in place that should be pretty doable in a follow-up.

packages/playwright/src/program.ts
```diff
@@ -203,10 +203,13 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
 
   const runner = new Runner(config);
   let status: FullResult['status'];
-  if (process.env.PWTEST_WATCH)
+  if (process.env.PWTEST_WATCH) {
```
dgozman, line 206: We should move this further up, similar to the line 164, without loading the config at all, and without involving `Runner`.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -15,130 +15,124 @@
...
+  }
 
-      if (this._timer)
-        clearTimeout(this._timer);
-      this._timer = setTimeout(() => {
-        this._notifyDirtyFiles?.();
-      }, 250);
-    });
+  onerror(listener: () => void): void {
+    this.on('error', listener);
```
dgozman, line 49: Note: we do not ever emit `error` events, because failing to handle them makes Node exit. Rename the event to something else, e.g. `transportError`.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -15,130 +15,124 @@
...
-  const originalWorkers = config.config.workers;
-  const fsWatcher = new FSWatcher();
-  await fsWatcher.update(config);
-
-  let lastRun: { type: 'changed' | 'regular' | 'failed', failedTestIds?: Set<string>, dirtyTestFiles?: Map<FullProjectInternal, Set<string>> } = { type: 'regular' };
+interface WatchModeOptions {
+  files?: string[];
+  projects?: string[];
+  grep?: string;
+  onlyChanged?: string;
```
dgozman, line 69: This one is not supported now.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -15,130 +15,124 @@
...
+        }
+      }
+  );
+  testServerDispatcher.transport.sendEvent = (method, params) => {
+    transport.emit('message', JSON.stringify({ method, params }));
+  };
+  const testServerConnection = new TestServerConnection(transport);
+  transport.emit('open');
+
+  const telesuiteUpdater = new TeleSuiteUpdater({ pathSeparator: path.sep, onUpdate() { } });
```
dgozman, line 93: nit: `teleSuiteUpdater`?

packages/playwright/src/runner/watchMode.ts
```diff
@@ -250,71 +237,36 @@ export async function runWatchModeLoop(config: FullConfigInternal): Promise<Full
...
-    status = 'failed';
-  if (status === 'passed' && taskStatus !== 'passed')
-    status = taskStatus;
-  await taskRunner.reporter.onEnd({ status });
-  await taskRunner.reporter.onExit();
+async function runChangedTests(watchOptions: WatchModeOptions, testServerConnection: TestServerConnection, changedFiles: string[], title?: string) {
+  if (watchOptions.files?.length)
+    changedFiles = changedFiles.filter(createFileMatcherFromArguments(watchOptions.files));
+
+  await runTests(watchOptions, testServerConnection, { title: title || 'files changed', locations: changedFiles });
```
dgozman, line 250: I think it is now easier to pass `testId` filter instead, given that you have `teleSuiteUpdater.rootSuite.allTests()`.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -15,130 +15,124 @@
...
+  };
+  const testServerConnection = new TestServerConnection(transport);
+  transport.emit('open');
+
+  const telesuiteUpdater = new TeleSuiteUpdater({ pathSeparator: path.sep, onUpdate() { } });
+
+  const dirtyTestFiles = new Set<string>();
+  const onDirtyTestFiles: { resolve?(): void } = {};
+
+  testServerConnection.onTestFilesChanged(async ({ testFiles: changedFiles }) => {
```
dgozman, line 98: I think we want to queue these operations, because listing tests might take longer than subsequent `onTestFilesChanged` events.

dgozman (commented):
This looks very good now, almost there!

Skn0tt (commented):
Addressed the feedback!

packages/playwright/src/runner/watchMode.ts
```diff
@@ -15,130 +15,127 @@
...
+    queue = queue.then(async () => {
+      const { report } = await testServerConnection.listTests({ locations: options.files, projects: options.projects, grep: options.grep });
+      teleSuiteUpdater.processListReport(report);
+
+      for (const test of teleSuiteUpdater.rootSuite!.allTests()) {
+        if (changedFiles.includes(test.location.file))
+          dirtyTestIds.add(test.id);
+      }
+
+      // listTests can take longer than the debouncing interval of onTestFilesChanged,
```
dgozman, line 112: I'd instead add `changedFiles` from multiple `onTestFilesChanged` into a single set, while a single `listTests()` is in progress, and then run lines 107-110 against the combined set and trigger `onDirtyTests` synchronously. This will make it a single async hop that's easier to reason about.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -15,130 +15,127 @@
...
+  testServerDispatcher.transport.sendEvent = (method, params) => {
+    transport.emit('message', JSON.stringify({ method, params }));
+  };
+  const testServerConnection = new TestServerConnection(transport);
+  transport.emit('open');
+
+  const teleSuiteUpdater = new TeleSuiteUpdater({ pathSeparator: path.sep, onUpdate() { } });
+
+  const dirtyTestIds = new Set<string>();
+  const onDirtyTests: { resolve?(): void } = {};
```
dgozman, line 95: Let's use `ManualPromise` instead.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -147,32 +144,31 @@ export async function runWatchModeLoop(config: FullConfigInternal): Promise<Full
...
-      const { projectNames } = await enquirer.prompt<{ projectNames: string[] }>({
+      const { selectedProjects } = await enquirer.prompt<{ selectedProjects: string[] }>({
         type: 'multiselect',
-        name: 'projectNames',
+        name: 'selectedProjects',
         message: 'Select projects',
-        choices: config.projects.map(p => ({ name: p.project.name })),
-      }).catch(() => ({ projectNames: null }));
-      if (!projectNames)
+        choices: teleSuiteUpdater.rootSuite!.suites.map(s => s.title),
```
dgozman, line 166: Won't this be a truncated list after we run `listTests()` with a project filter once? I think `teleSuiteUpdater.config.projects` is a better source.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -204,40 +199,35 @@ export async function runWatchModeLoop(config: FullConfigInternal): Promise<Full
...
+        await runTests(options, testServerConnection, { title: 're-running tests' });
         continue;
       } else if (lastRun.type === 'changed') {
-        await runChangedTests(config, failedTestIdCollector, lastRun.dirtyTestFiles!, 're-running tests');
+        await runTests(options, testServerConnection, { title: 're-running tests', testIds: lastRun.dirtyTestIds });
       } else if (lastRun.type === 'failed') {
-        config.testIdMatcher = id => lastRun.failedTestIds!.has(id);
-        await runTests(config, failedTestIdCollector, { title: 're-running tests' });
-        config.testIdMatcher = undefined;
+        await runTests({}, testServerConnection, { title: 're-running tests', testIds: [...lastRun.failedTestIds!] });
```
dgozman, line 224: nit: I wonder why make a copy of `testIds`?

packages/playwright/src/runner/runner.ts
```diff
@@ -134,7 +134,10 @@ export class Runner {
   async watchAllTests(): Promise<FullResult['status']> {
```
dgozman, line 134: Let's get rid of this method entirely, inline it into `program.ts` and not create `Runner` for watch mode at all.

packages/playwright/src/runner/watchMode.ts
Skn0tt: I think this is a leftover from a previous iteration, where `failedTestIDs` was a Set that needed conversion to array. fixed.

packages/playwright/src/runner/watchMode.ts
Skn0tt: done

packages/playwright/src/runner/runner.ts
Skn0tt: done!

Skn0tt (commented):
addressed. gonna think about a solution for the project picker next

packages/playwright/src/runner/watchMode.ts
dgozman: > I'm not sure what the right solution is. We could use the projects from our initial `listTests` call, but then we'd need to remove the filters from that. Would probably be fine.

This sounds like the easiest solution for now.

We can probably update `TeleEmitter` to include projects, but this needs to be done carefully to make sure merger is ok, in a follow up PR.

dgozman (approved):
Looks good to me. Let's follow up with a projects list and add a test for that in a separate PR.

packages/playwright/src/runner/watchMode.ts
Skn0tt: oh no, I had an update to `listTests` typed but not committed before pushing 🤦 gonna work on that refactor next

PR #32158, opened 2024-08-14
fix(client-certificates): improve close handling from target and proxy

mxschmitt:
Motivation: TLS handshake errors were not reported correctly to the client. Before e.g. during ALPNCache.get neither catch/then was called which lead to stalling connections. This fixes it by adding the appropriate listeners and reporting it back to the client. 

https://github.com/microsoft/playwright/issues/32004

### DIFF
--- a/packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
+++ b/packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
@@ -79,14 +79,21 @@ class SocksProxyConnection {
   target!: net.Socket;
   // In case of http, we just pipe data to the target socket and they are |undefined|.
   internal: stream.Duplex | undefined;
+  internalTLS: tls.TLSSocket | undefined;
   private _targetCloseEventListener: () => void;
+  private _dummyServer: tls.Server | undefined;
 
   constructor(socksProxy: ClientCertificatesProxy, uid: string, host: string, port: number) {
     this.socksProxy = socksProxy;
     this.uid = uid;
     this.host = host;
     this.port = port;
-    this._targetCloseEventListener = () => this.socksProxy._socksProxy.sendSocketEnd({ uid: this.uid });
+    this._targetCloseEventListener = () => {
+      // Close the other end and cleanup TLS resources.
+      this.socksProxy._socksProxy.sendSocketEnd({ uid: this.uid });
+      this.internalTLS?.destroy();
+      this._dummyServer?.close();
+    };
   }
 
   async connect() {
@@ -101,8 +108,10 @@ class SocksProxyConnection {
   }
 
   public onClose() {
-    this.internal?.destroy();
+    // Close the other end and cleanup TLS resources.
     this.target.destroy();
+    this.internalTLS?.destroy();
+    this._dummyServer?.close();
   }
 
   public onData(data: Buffer) {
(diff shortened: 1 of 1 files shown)


packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -214,7 +212,6 @@ class SocksProxyConnection {
         });
 
         internalTLS.once('end', () => closeBothSockets());
-        targetTLS.once('end', () => closeBothSockets());
```
mxschmitt, line 217: The order is the following:

- `targetTLS` `'end'` event
- `targetTLS` `'error'` event (we are interested in that and want to write to the `internalTLS` there but the problem is that `internalTLS` gets closed in the previous `'end'` event.)

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -60,11 +60,9 @@ class ALPNCache {
       ALPNProtocols: ['h2', 'http/1.1'],
       rejectUnauthorized: false,
     }).then(socket => {
-      socket.once('secureConnect', () => {
```
mxschmitt, line 63: `createTLSSocket` is only used for client certificates. Its internally using `createConnectionAsync` which we don't want to modify.  Changing `createTLSSocket` seems reasonable.

tests/library/client-certificates.spec.ts
```diff
@@ -302,6 +303,47 @@ test.describe('browser', () => {
     await page.close();
   });
 
+  test('should not hang on tls errors during TLS 1.2 handshake', async ({ browser, asset, browserName }) => {
+    for (const tlsVersion of ['TLSv1.3', 'TLSv1.2'] as const) {
+      await test.step(`TLS version: ${tlsVersion}`, async () => {
+        const server = tls.createServer({
```
mxschmitt, line 309: I didn't want to pollute our existing server fixture with more specific options which are only relevant for this test, hence creating a pure TLS server here.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
dgozman: Sounds good.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -150,6 +148,9 @@ class SocksProxyConnection {
         };
 
         const handleError = (error: Error) => {
+          // In case of an 'error' event on the target connection this method takes care of closing the internalTLS socket.
+          targetTLS?.removeListener('close', closeBothSockets);
```
dgozman, line 152: - I think this line is no-op, because we add a different function as an even listener.
- The comment did not really help me understand why do we need to remove the listener.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
mxschmitt: It's save to not listen on `targetTLS.once('close` here since we already to it in line 94.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -194,26 +210,24 @@ class SocksProxyConnection {
           return;
         }
 
-        const tlsOptions: tls.ConnectionOptions = {
+        if (this._closed)
+          return;
```
dgozman, line 214:
```suggestion
        if (this._closed) {
          internalTLS.destroy();
          return;
        }
```

dgozman (approved):

PR #32162, opened 2024-08-14
chore: generate client-certificate on the fly

mxschmitt:
- maintained but they only fix critical bugs, last commit 2 years ago
- popular: 20 million downloads/week
- bundle size 500kb with esbuild
- dependencies 0
- takes on my M1 Pro machine 200ms to generate a TLS certificate


mxschmitt:
We'll try a different approach for now and use the client-certificates for the interception instead.

PR #32168, opened 2024-08-15
chore: validate client-certificates on context creation

mxschmitt:


packages/playwright-core/src/server/browserContext.ts
```diff
@@ -735,6 +736,25 @@ export function verifyClientCertificates(clientCertificates?: channels.BrowserNe
       throw new Error('key is specified without cert');
     if (cert.pfx && (cert.cert || cert.key))
       throw new Error('pfx is specified together with cert, key or passphrase');
+    try {
```
dgozman, line 739: If we move this into `ClientCertificatesProxy` constructor (and a similar place in fetch), perhaps we can save the resulting secure context and reuse it? Perhaps that gives us a bit more clarity (created a context ones and the using it) and a small performance win?

mxschmitt: I tried to apply this feedback to the SocksCertificateProxy, there it seems to work out well. For the APIRequestContext, `https.request` unfortunately doesn't accept a `tls.SecureContext` - so we keep things there as it is.

This seems to be aligned with the goal of the PR: clear and fast user-feedback which seems to be the case now for both, browser and APIRequestAPI (like before) usage.

dgozman (approved):

PR #32169, opened 2024-08-15
chore: do not open up all TLS sessions when using client-certificates

mxschmitt:
Relevant commit: 6de0d10d3eca74e5ec7bc5dca25500dcc6b09a6f


mxschmitt:
We can't do this, since this won't keep maintaining the users `ignoreHttpsErrors` setting.

PR #32181, opened 2024-08-15
remove redundant await

harsha509:


pavelfeldman:
We allow awaiting return values for readability.

PR #32192, opened 2024-08-16
chore: generate self-signed certificates for socks proxy

mxschmitt:


packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +25,179 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+
+export function generateSelfSignedCertificate(commonName: string) {
+  const { privateKey, publicKey } = crypto.generateKeyPairSync('rsa', { modulusLength: 2048 });
+  const publicKeyDer = publicKey.export({ type: 'pkcs1', format: 'der' });
+
+  const tbsCertificate = DER.encodeSequence([
+    DER.encodeForContext(0, DER.encodeInteger(1)), // version
+    DER.encodeInteger(1), // serialNumber
+    DER.encodeSequence([
+      DER.encodeObject('1.2.840.113549.1.1.11'),
```
dgozman, line 141: Let's add a human-readable name of each OID as a comment.

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +25,179 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+    for (let i = 2; i < parts.length; i++) {
+      output.push(encodeBase128(parts[i]));
+    }
+
+    return this._encode(0x06, Buffer.concat(output));
+  }
+  static encodeNull(): Buffer {
+    return Buffer.from([0x05, 0x00]);
+  }
+  static encodeSet(data: Buffer[]): Buffer {
```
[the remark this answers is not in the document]
mxschmitt: I added a comment, didn't find a comment about it in the spec tho. It only says:

> Use a set type to model a collection of variables whose number is known and modest and whose order is insignificant

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -32,10 +30,9 @@ let dummyServerTlsOptions: tls.TlsOptions | undefined = undefined;
 function loadDummyServerCertsIfNeeded() {
   if (dummyServerTlsOptions)
     return;
-  dummyServerTlsOptions = {
-    key: fs.readFileSync(path.join(__dirname, '../../bin/socks-certs/key.pem')),
-    cert: fs.readFileSync(path.join(__dirname, '../../bin/socks-certs/cert.pem')),
-  };
+  // TODO: do we want to have it unique per browser context, launch or global?
```
dgozman, line 33: I think it's good as it is, you can remove the todo 😄

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import assert from 'assert';
```
dgozman, line 17: For some reason, we use `assert` defined in `utils/debug.ts` instead.

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +26,164 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+    return this._encode(0x06, Buffer.concat(output));
+  }
+  static encodeNull(): Buffer {
+    return Buffer.from([0x05, 0x00]);
+  }
+  static encodeSet(data: Buffer[]): Buffer {
+    // We expect the data to be already sorted.
+    return this._encode(0x31, Buffer.concat(data));
+  }
+  static encodeImplicitContextDependent(tag: number, data: Buffer): Buffer {
```
dgozman, line 69: My bad, this is actually an explicit one.

```suggestion
  static encodeExplicitContextDependent(tag: number, data: Buffer): Buffer {
```

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +26,164 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+    ]), // signature
+    DER.encodeSequence([
+      DER.encodeSet([
+        DER.encodeSequence([
+          DER.encodeObjectIdentifier('2.5.4.3'), // commonName X.520 DN component
+          DER.encodePrintableString('localhost')
+        ]),
+        DER.encodeSequence([
+          DER.encodeObjectIdentifier('2.5.4.10'), // organizationName X.520 DN component
+          DER.encodePrintableString('Client Certificate Demo')
```
dgozman, line 137:
```suggestion
          DER.encodePrintableString('Playwright Client Certificate Support')
```

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +26,164 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+        ])
+      ])
+    ]), // issuer
+    DER.encodeSequence([
+      DER.encodeDate(new Date()), // notBefore
```
dgozman, line 142: Let's do `today minus one year` and `today plus one year` instead.

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +26,164 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+    ]), // validity
+    DER.encodeSequence([
+      DER.encodeSet([
+        DER.encodeSequence([
+          DER.encodeObjectIdentifier('2.5.4.3'), // commonName X.520 DN component
+          DER.encodePrintableString('localhost')
+        ]),
+        DER.encodeSequence([
+          DER.encodeObjectIdentifier('2.5.4.10'), // organizationName X.520 DN component
+          DER.encodePrintableString('Client Certificate Demo')
```
dgozman, line 155:
```suggestion
          DER.encodePrintableString('Playwright Client Certificate Support')
```

packages/playwright-core/src/utils/crypto.ts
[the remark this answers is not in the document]
dgozman: I see the following in `X.6901.doc` section 11.6:

```
The encodings of the component values of a set-of value shall appear in ascending order, the encodings being compared as octet strings with the shorter components being padded at their trailing end with 0-octets. 
NOTE – The padding octets are for comparison purposes only and do not appear in the encodings.
```

Shall we sort things? Or we can assert that `data.length === 1` for now to avoid sorting 😄

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +26,164 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+  // List of fields / structure: https://datatracker.ietf.org/doc/html/rfc2459#section-4.1
+  const tbsCertificate = DER.encodeSequence([
+    DER.encodeImplicitContextDependent(0, DER.encodeInteger(1)), // version
+    DER.encodeInteger(1), // serialNumber
+    DER.encodeSequence([
+      DER.encodeObjectIdentifier('1.2.840.113549.1.1.11'), // sha256WithRSAEncryption PKCS #1
+      DER.encodeNull()
+    ]), // signature
+    DER.encodeSequence([
+      DER.encodeSet([
```
dgozman, line 128: As discussed offline, this should be two sets of one-size sequences, not a single set of two-sized sequence.

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +26,164 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+          DER.encodePrintableString('Client Certificate Demo')
+        ])
+      ])
+    ]), // issuer
+    DER.encodeSequence([
+      DER.encodeDate(new Date()), // notBefore
+      DER.encodeDate(new Date()), // notAfter
+    ]), // validity
+    DER.encodeSequence([
+      DER.encodeSet([
```
dgozman, line 146: Same thing here about sets/sequences.

packages/playwright-core/src/utils/crypto.ts
```diff
@@ -25,3 +26,164 @@ export function calculateSha1(buffer: Buffer | string): string {
...
+  return Buffer.from(bytes.reverse());
+};
+
+// ASN1/DER Speficiation:   https://www.itu.int/rec/T-REC-X.680-X.693-202102-I/en
+class DER {
+  static encodeSequence(data: Buffer[]): Buffer {
+    return this._encode(0x30, Buffer.concat(data));
+  }
+  static encodeInteger(data: number): Buffer {
+    assert(data >= 0 && data <= 0xff);
```
dgozman, line 48:
```suggestion
    assert(data >= -128 && data <= 127);
```

dgozman (approved):

PR #32196, opened 2024-08-16
fix(only changed): make only-changed work together with list mode

### LINKED ISSUE #32161: [Feature]: Allow `--only-changed` to work with `--list`
### 🚀 Feature Request

It seems like trying to combine `--list` and `--only-changed` doesn't work. It will just list all the tests. Would be nice if these 2 options could be combined to list out only the tests that changed.

### Example

Would be nice if this output/called reporters with a filtered list of only tests that changed.

`npx playwright test --list --only-changed=main`

### Motivation

This can be handy for reporting reasons or for custom tooling. Right now we use a minimal reporter to output information about tests we care about into a json format. Being able to get a set of only changed tests through that workflow would be nice.

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32161

Turns out we were wrong in https://github.com/microsoft/playwright/pull/31727#discussion_r1685793870!

Adds support for `--only-changed` in combination with `--list` by removing our code to prevent that.


dgozman (approved):

Vitallu (commented):
My Brothers, thanks!

PR #32197, opened 2024-08-16
fix(only-changed): exit successfully if there were no changes

### LINKED ISSUE #32180: [Bug]: `--only-changed` with no tests found should have exit code 0
### Version

latest

### Steps to reproduce

Run with `--only-changed` if there are no tests. 

### Expected behavior

`No tests found` is printed, and it exits with exit code 0.

### Actual behavior

`No tests found` is printed, and it exits with exit code 1.

### Additional context

If folks want to use `--only-changed` in automation, then exit code 0 is a sane default.

### Environment

```shell
irrelevant
```

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32180

I was briefly wondering if we should output a log line a la "no tests found", but my understanding is that that's the reporters job - so I didn't change anything in that regard.


packages/playwright/src/program.ts
```diff
@@ -198,7 +198,7 @@ async function runTests(args: string[], opts: { [key: string]: any }) {
   config.cliGrepInvert = opts.grepInvert as string | undefined;
   config.cliListOnly = !!opts.list;
   config.cliProjectFilter = opts.project || undefined;
-  config.cliPassWithNoTests = !!opts.passWithNoTests;
+  config.cliPassWithNoTests = !!(opts.passWithNoTests ?? opts.onlyChanged);
```
dgozman, line 201: nit: I'd rather check `cliOnlyChanged` in the place where we check `passWithNoTests`. That feels cleaner to me, because you can always reason that `passWithNoTests` was set explicitly in the CLI.

dgozman (approved):
Looks good, but tests are not happy.

packages/playwright/src/program.ts
Skn0tt: addressed!

dgozman (approved):

PR #32210, opened 2024-08-18
feat(client-certificates): allow passing certificates from memory

### LINKED ISSUE #32200: [Feature]: Load client certificates from RAM
### 🚀 Feature Request

The API for client certificates should accept the actual certificate, not a path to a file containing the certificate.

### Example

The certificate could be coming in at runtime from a physical hardware token or from some other source that isn't a file.

### Motivation

The current implementation of the new client certificates feature from #1799 and #32004 has the API expecting a file path to receive the client certificates. However, the certificates might not be coming from a file: they could instead get generated at runtime and only exist in RAM. It would be preferable and also possibly more secure for the API to accept the actual certificate instead of a path to a file containing the certificate because a file might not strictly need to be written to do this operation. The certificates could come from a file or they could come from anything as long as the certificates are provided somehow.

I discussed this request with @mxschmitt on Discord.

mxschmitt:
Motivation: We received multiple feedback from users who want to add certificates from in-memory.

Fixes https://github.com/microsoft/playwright/issues/32200

### DIFF
--- a/docs/src/api/params.md
+++ b/docs/src/api/params.md
@@ -531,9 +531,9 @@ Does not enforce fixed viewport, allows resizing window in the headed mode.
 - `clientCertificates` <[Array]<[Object]>>
   - `origin` <[string]> Exact origin that the certificate is valid for. Origin includes `https` protocol, a hostname and optionally a port.
   - `certPath` ?<[path]> Path to the file with the certificate in PEM format.
-  - `cert` ?<[string]> Direct value of the certificate in PEM format.
+  - `cert` ?<[Buffer]> Direct value of the certificate in PEM format.
   - `keyPath` ?<[path]> Path to the file with the private key in PEM format.
-  - `key` ?<[string]> Direct value of the private key in PEM format.
+  - `key` ?<[Buffer]> Direct value of the private key in PEM format.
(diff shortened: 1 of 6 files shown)


docs/src/api/params.md
```diff
@@ -531,15 +531,18 @@ Does not enforce fixed viewport, allows resizing window in the headed mode.
 - `clientCertificates` <[Array]<[Object]>>
   - `origin` <[string]> Exact origin that the certificate is valid for. Origin includes `https` protocol, a hostname and optionally a port.
   - `certPath` ?<[path]> Path to the file with the certificate in PEM format.
+  - `cert` ?<[string]> Direct value of the certificate in PEM format.
```
mxschmitt, line 534: Unsure if we should do `[Buffer]` everywhere or `[Buffer]|[string]` for a better user experience. Doing `Buffer` only seems the simplest to avoid overloads in e.g. .NET.

dgozman: I think buffer everywhere is good.

dgozman (approved):

docs/src/api/params.md
mxschmitt: Done.

PR #32225, opened 2024-08-20
chore(ui): add test for font preview

Skn0tt:
Adds a test for the font preview feature.


tests/library/trace-viewer.spec.ts
```diff
@@ -289,6 +289,20 @@ test('should filter network requests by resource type', async ({ page, runAndTra
...
+    await page.goto(`${server.PREFIX}/network-tab/network.html`);
+  });
+  await traceViewer.selectAction('http://localhost');
+  await traceViewer.showNetworkTab();
+
+  await traceViewer.page.getByText('Font', { exact: true }).click();
+  await expect(traceViewer.networkRequests).toHaveCount(1);
+  await traceViewer.networkRequests.getByText('font.woff2').click();
+  await traceViewer.page.getByTestId('network-request-details').getByTitle('Body').click();
+  await expect(traceViewer.page.locator('.network-request-details-tab')).toHaveScreenshot({ maxDiffPixelRatio: 0.1 });
```
dgozman, line 303: The chances are this screenshot will not work across various operating systems and configurations. You can try adding a label `CQ1` to check that. I'd recommend checking that text "ABCDEF" is present in the preview instead.

Skn0tt: Today I learned about CQ1! Just for my understanding, do we run those fat tests on any label, or only on CQ1? looking at https://github.com/Skn0tt/playwright/blob/2feaecd34eeaecb81534d66d9317b6bbd8463319/.github/workflows/tests_secondary.yml#L12.

Skn0tt: updated the test in https://github.com/microsoft/playwright/pull/32225/commits/a86831dd974b51018536d6b23600f682e9a5f026

mxschmitt: The trigger is for any label but we never add labels for PRs, so we usually use just `CQ1`.

mxschmitt (approved):

Issue #32243, opened 2024-08-21
[Bug]: Missing chain operator when populating annotations in the html-reporter


siteriaitaliana:
### Version

latest

### Steps to reproduce

1. In my test framework I had an annotation set by mistake in this way:
`test.info.annotations.push(' //TODO ')`
nothing has caught this either the compilation considering we don't use typescript in this project
2. After spending hours debugging and debugging directly injecting my playwright base 64 report global variable into the html-reporter I have managed to find the test and the culprit code

We should fix this by ignoring this occurrences (the annotations wrongly set) using a property check [I have a forked PR for this which will attach to this...] so the report isn't blank as in the screenshot with the js console error not very descriptive, we could also mention the test name in the console logs when this happens, if needed let me know and will add in my fork PR

### Expected behaviour

A wrongly set annotation shouldn't corrupt thew whole report

### Actual behaviour

A blank report page is displayed in the browser

![empty_report](https://github.com/user-attachments/assets/d74575a5-44cb-4ae2-8564-e9a60b3e5c8f)

### Environment

```shell
System:
    OS: macOS 14.6.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 111.66 MB / 16.00 GB
  Binaries:
    Node: 18.19.0 - ~/.asdf/installs/nodejs/18.19.0/bin/node
    Yarn: 1.22.19 - ~/.asdf/shims/yarn
    npm: 10.2.5 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 8.6.5 - ~/.nvm/versions/node/v16.20.0/bin/pnpm
  IDEs:
    VSCode: 1.92.2 - /usr/local/bin/code
  Languages:
    Bash: 3.2.57 - /bin/bash
```

siteriaitaliana:
Forked PR: https://github.com/microsoft/playwright/pull/32244

mxschmitt:
Playwright can't prevent users from calling functions or pushing data into arrays in the wrong format. In your case you update one consumer but there are many more - so fixing it like that is not feasible. We strongly recommend using TypeScript and running `tsc --noEmit` on your CI. For me it then shows an error:

<img width="547" alt="image" src="https://github.com/user-attachments/assets/c718537a-3eaf-4fa1-8330-90538bba3cd6">

I'll update the docs to reflect that recommendation: https://github.com/microsoft/playwright/pull/32245

PR #32244, opened 2024-08-21
fix: added chain operator to avoid empty reports when an annotation i…

l-urbini:
…s wrongly set e.g. test.info.annotations.push(' //')


mxschmitt:
Closing as per https://github.com/microsoft/playwright/issues/32243#issuecomment-2301596439.

PR #32257, opened 2024-08-22
fix(test runner): expect.poll error reporting should handle non-expect errors

### LINKED ISSUE #32256: [Bug]: Test report says "undefined" when a promise is rejected inside expect.poll
### Version

1.46.1

### Steps to reproduce

1. Clone this sample repo https://github.com/bdcarr/playwright-poll-covers-error
2. `npm install && npm run test`

You can remove axios and just `return Promise.reject({ some: 'object', or: 'other' })` inside expect.poll, but I included the request & interceptor to explain our use case.

### Expected behavior

The test report prints the whole error object, same as happens when the error occurs outside of `expect.poll`.

### Actual behavior

The error printed in the report just says "undefined":
```
  1) test.spec.ts:19:5 › issue happens when request is inside expect.poll ──────────────────────────

    undefined

      20 |     await expect.poll(async () => {
      21 |         return request.get('https://www.fwoeifhwefwf.com/efhjio2wfhejilfuhwf');
    > 22 |     }).toMatchObject({ something: 'idk' });
         |        ^
      23 | });
      24 |
      25 | test('issue does not happen when request is outside expect.poll', async () => {

        at /Users/brycecarr/dev/src/my/playwright-poll-covers-error/test.spec.ts:22:8
```

### Additional context

In our actual repo, the report says "undefined: undefined". Not sure why it just says "undefined" once in the repro. Our real repo's on 1.40.1, maybe that's why.

### Environment

```shell
System:
    OS: macOS 14.6.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 78.86 MB / 32.00 GB
  Binaries:
    Node: 20.15.1 - ~/.local/state/fnm_multishells/77311_1724290084305/bin/node
    npm: 10.7.0 - ~/.local/state/fnm_multishells/77311_1724290084305/bin/npm
  IDEs:
    VSCode: 1.92.2 - /usr/local/bin/code
  Languages:
    Bash: 5.2.32 - /opt/homebrew/bin/bash
  npmPackages:
    @playwright/test: ^1.46.1 => 1.46.1
```

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32256

We were expecting all errors to be of type `ExpectError`, but apparently `expect` propagates rejections in the polling functions right through. So we also need to handle that case.

I wonder if we have more cases of this. Would it make sense to enable `useUnknownInCatchVariables` in TypeScript?

### DIFF
--- a/packages/playwright/src/matchers/expect.ts
+++ b/packages/playwright/src/matchers/expect.ts
@@ -60,7 +60,7 @@ import {
 } from '../common/expectBundle';
 import { zones } from 'playwright-core/lib/utils';
 import { TestInfoImpl } from '../worker/testInfo';
-import { ExpectError } from './matcherHint';
+import { ExpectError, isExpectError } from './matcherHint';
 
 // #region
 // Mirrored from https://github.com/facebook/jest/blob/f13abff8df9a0e1148baf3584bcde6d1b479edc7/packages/expect/src/print.ts
@@ -289,8 +289,9 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
 
       const step = testInfo._addStep(stepInfo);
 
-      const reportStepError = (jestError: ExpectError) => {
-        const error = new ExpectError(jestError, customMessage, stackFrames);
+      const reportStepError = (_error: unknown) => {
+        const jestError = _error instanceof Error ? _error : new Error(String(_error));
+        const error = isExpectError(jestError) ? new ExpectError(jestError, customMessage, stackFrames) : jestError;
         step.complete({ error });
         if (this._info.isSoft)
           testInfo._failWithError(error);
--- a/packages/playwright/src/matchers/expect.ts
+++ b/packages/playwright/src/matchers/expect.ts
@@ -289,8 +289,7 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
 
       const step = testInfo._addStep(stepInfo);
 
-      const reportStepError = (unknownError: unknown) => {
-        const jestError = unknownError instanceof Error ? unknownError : new Error(String(unknownError));
+      const reportStepError = (jestError: Error | unknown) => {
         const error = isExpectError(jestError) ? new ExpectError(jestError, customMessage, stackFrames) : jestError;
         step.complete({ error });
         if (this._info.isSoft)
--- a/packages/playwright/src/matchers/expect.ts
+++ b/packages/playwright/src/matchers/expect.ts
@@ -289,8 +289,8 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
 
       const step = testInfo._addStep(stepInfo);
 
-      const reportStepError = (_error: unknown) => {
-        const jestError = _error instanceof Error ? _error : new Error(String(_error));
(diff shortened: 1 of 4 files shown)


packages/playwright/src/matchers/expect.ts
```diff
@@ -289,8 +289,9 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
 
       const step = testInfo._addStep(stepInfo);
 
-      const reportStepError = (jestError: ExpectError) => {
-        const error = new ExpectError(jestError, customMessage, stackFrames);
+      const reportStepError = (_error: unknown) => {
+        const jestError = _error instanceof Error ? _error : new Error(String(_error));
```
dgozman, line 293: Why wrapping into `Error(...)`? I'd think the next line is sufficient.

packages/playwright/src/matchers/expect.ts
```diff
@@ -289,8 +289,9 @@ class ExpectMetaInfoProxyHandler implements ProxyHandler<any> {
 
       const step = testInfo._addStep(stepInfo);
 
-      const reportStepError = (jestError: ExpectError) => {
-        const error = new ExpectError(jestError, customMessage, stackFrames);
+      const reportStepError = (_error: unknown) => {
```
dgozman, line 292: style: we only use `_leadingUnderscore` for private fields, not for local variable.

Skn0tt: done in e4b1dc95dc5155dfbad915358cda5b09f89d4c7d

packages/playwright/src/matchers/expect.ts
Skn0tt: Because both `step.complete()` and `testInfo._failWithError` require the `Error` type and don't take `unknown`.

dgozman: Actually, both of these are ready for non-`Error` errors, but the types are incorrect. I think we should add a test for this just in case, similar to the one in this PR.

Skn0tt: Done in https://github.com/microsoft/playwright/pull/32257/commits/e4b1dc95dc5155dfbad915358cda5b09f89d4c7d. The test in this PR already has the non-`Error` case.

tests/playwright-test/expect-poll.spec.ts
```diff
@@ -232,3 +232,16 @@ test('should show intermediate result for poll that spills over test time', asyn
   expect(result.output).toContain('Expected: 2');
   expect(result.output).toContain('Received: 3');
 });
+
+test('should propagate promise rejections', { annotation: { type: 'issue', description: 'https://github.com/microsoft/playwright/issues/32256' } }, async ({ runInlineTest }) => {
```
dgozman, line 236: Let's add a second test that would have the predicate `async () => { throw 'String value'; }`.

dgozman (approved):

tests/playwright-test/expect-poll.spec.ts
Skn0tt: done in 6b07b2287b2bea98fbfdba1a0611952183393827

PR #32258, opened 2024-08-22
fix(client-certificates): errors during http2 TLS handshake

mxschmitt:
Investigation: Currently when a TLS handshake error happens on a http2 connection, the response gets corrupted. This is because:

1. The client (browser, internalTLS) sends a http2 handshake to our proxy
1. The target connection emits a `secureConnect` event
1. We attach listeners since we received secureConnect event and send the buffered "client hello" data over
1. The target connection emits an `error` event
1. We try to send a http2 response over but the client (browser) already received some parts of the handshake and is confused. This ends up in `Error: Received bad client magic byte string` and `net::ERR_HTTP2_PROTOCOL_ERROR`.

This brought up that using the [`session` event](https://nodejs.org/api/tls.html#event-session) seems to be the right one for our use-case.

Relates https://github.com/playwright-community/playwright-go/pull/480
Example error: https://github.com/playwright-community/playwright-go/actions/runs/10465991368/job/28982096640?pr=480#step:6:166

### DIFF
--- a/packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
+++ b/packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
@@ -169,7 +169,10 @@ class SocksProxyConnection {
               this.target.removeListener('close', this._targetCloseEventListener);
               // @ts-expect-error
               const session: http2.ServerHttp2Session = http2.performServerHandshake(internalTLS);
-              session.on('error', () => this._targetCloseEventListener());
+              session.on('error', () => {
+                this.target.destroy();
+                this._targetCloseEventListener();
+              });
               session.once('stream', (stream: http2.ServerHttp2Stream) => {
                 stream.respond({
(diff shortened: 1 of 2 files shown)


dgozman (approved):

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -169,6 +169,7 @@ class SocksProxyConnection {
               this.target.removeListener('close', this._targetCloseEventListener);
               // @ts-expect-error
               const session: http2.ServerHttp2Session = http2.performServerHandshake(internalTLS);
+              session.on('error', () => this._targetCloseEventListener());
```
dgozman, line 172: Looking at the line 180, shall we `this.target.destroy()` as well?

dgozman (approved):

PR #32268, opened 2024-08-22
test: rebase modernizer Linux tests

mxschmitt:
This fails since we unskipped some test in https://github.com/microsoft/playwright/pull/32147 which were skipped by accident on Linux and Windows. This PR rebases them.

Relates https://github.com/microsoft/playwright/issues/32288.


tests/library/modernizr.spec.ts
```diff
@@ -41,6 +41,9 @@ it('safari-14-1', async ({ browser, browserName, platform, server, headless, isM
 
   if (platform === 'linux') {
     expected.subpixelfont = false;
+    expected.applicationcache = false;
```
yury-s, line 44:
```
    expected.applicationcache = false;
    expected.inputsearchevent = false;
```
is already the case on mac as well. Does Windows differ or we can align across all OSes?

yury-s (approved):

dgozman (approved):

PR #32274, opened 2024-08-22
chore(ui): decouple TestServerConnection from websocket transport

Skn0tt:
Preparation for https://github.com/microsoft/playwright/issues/32076.


packages/playwright/src/isomorphic/testServerConnection.ts
```diff
@@ -19,6 +19,16 @@ import * as events from './events';
 
 // -- Reuse boundary -- Everything below this line is reused in the vscode extension.
 
+export interface TestServerSocket {
```
dgozman, line 22: - We usually call things like this one "Transport" for easier search across the codebase.
- This is a somewhat convoluted interface for a transport. Perhaps we should separate transport (onmessage, onclose, send, close) from a factory that creates a websocket transport (this one needs open/error/etc).

packages/trace-viewer/src/ui/workbenchLoader.tsx
```diff
@@ -102,7 +102,7 @@ export const WorkbenchLoader: React.FunctionComponent<{
       const guid = new URLSearchParams(window.location.search).get('ws');
       const wsURL = new URL(`../${guid}`, window.location.toString());
       wsURL.protocol = (window.location.protocol === 'https:' ? 'wss:' : 'ws:');
-      const testServerConnection = new TestServerConnection(wsURL.toString());
+      const testServerConnection = new TestServerConnection(new WebSocket(wsURL));
```
Meir017, line 105:
```js
const guid = new URLSearchParams(window.location.search).get('ws');
      const wsURL = new URL(`../${guid}`, window.location.toString());
      wsURL.protocol = (window.location.protocol === 'https:' ? 'wss:' : 'ws:');
      const testServerConnection = new TestServerConnection(new WebSocket(wsURL));
```

this code is also used in `packages/trace-viewer/src/ui/uiModeView.tsx` consider extracting to a utility function

packages/playwright/src/isomorphic/testServerConnection.ts
Skn0tt: Sounds good! How do you like https://github.com/microsoft/playwright/pull/32274/commits/0f37e55eeca8b81e5bc524088dafd213db98e8fc?

dgozman (approved):

PR #32299, opened 2024-08-23
fix(test runner): perform shallow clone check in config directory

Skn0tt:
Our CI operates on shallow clones. In vcs.ts, we perform a check for shallow clones in `process.cwd()` instead of the test directory. This makes the test in https://github.com/Skn0tt/playwright/blob/3c208aeeff255fd9ccf0528863e6b4b790d0f1b8/tests/playwright-test/only-changed.spec.ts#L201 failing in CI, but only for PRs. The fix is to perform the check on. the test directory.

### DIFF
--- a/packages/playwright/src/runner/vcs.ts
+++ b/packages/playwright/src/runner/vcs.ts
@@ -30,7 +30,7 @@ export async function detectChangedTestFiles(baseCommit: string, configDir: stri
 
       const unknownRevision = error.output.some(line => line?.includes('unknown revision'));
       if (unknownRevision) {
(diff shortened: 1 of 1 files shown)


mxschmitt:
While the fix looks reasonable, I didn't see any tests failing, do you mind sharing a link to the GHA which is failing?

Skn0tt:
Sure thing! Here's an example: https://github.com/microsoft/playwright/pull/32156#issuecomment-2307377656

pavelfeldman (approved):

PR #32305, opened 2024-08-23
chore: update removeAllListeners docs

### LINKED ISSUE #31474: [Docs]: Is there a way to cancel response.finished() promise?
### Page(s)

https://playwright.dev/docs/api/class-response#response-finished

### Description

I noticed that when I do   `await response.finished() // wait for response to finish` (inside the callback for `page.on('response", onResponse)`) that if the browser is closed, then this line throws an error "Target page, context or browser has been closed".
It throws even after the callback is unregistered with `page.off("response", onResponse)`, but that makes sense since any prior response is still in flight.

So is there a way though to cancel any inflight pending response.finished() promises so that they don't throw when the page is closed?

```
  page.on("response", onResponse)

  disposableStack.defer(() => {
    page.off("response", onResponse)
    page.off("close", onBrowserViewDestroyed)
  })

```

pavelfeldman:
Closes https://github.com/microsoft/playwright/issues/31474

### DIFF
--- a/docs/src/api/class-browser.md
+++ b/docs/src/api/class-browser.md
@@ -297,8 +297,10 @@ testing frameworks should explicitly create [`method: Browser.newContext`] follo
 
 ## async method: Browser.removeAllListeners
 * since: v1.47
+* langs: js
 
-Removes all the listeners of the given type if the type is given. Otherwise removes all the listeners.
+Removes all the listeners of the given type (or all registered listeners if no type given).
+Allows to wait for async listeners to complete or to ignore subsequent errors from these listeners.
 
 ### param: Browser.removeAllListeners.type
 * since: v1.47
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -3372,8 +3372,23 @@ By default, after calling the handler Playwright will wait until the overlay bec
 
 ## async method: Page.removeAllListeners
 * since: v1.47
+* langs: js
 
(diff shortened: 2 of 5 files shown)


docs/src/api/class-browser.md
```diff
@@ -297,8 +297,10 @@ testing frameworks should explicitly create [`method: Browser.newContext`] follo
 
 ## async method: Browser.removeAllListeners
 * since: v1.47
+* langs: js
 
-Removes all the listeners of the given type if the type is given. Otherwise removes all the listeners.
+Removes all the listeners of the given type (or all registered listeners if no type given).
```
yury-s, line 302:
```suggestion
Removes all the listeners of the given type (or all registered listeners if no type is given).
```

Here and below

docs/src/api/class-page.md
```diff
@@ -3372,8 +3372,23 @@ By default, after calling the handler Playwright will wait until the overlay bec
...
 
-Removes all the listeners of the given type if the type is given. Otherwise removes all the listeners.
+```js
+page.on('request', async request => {
+  const response = await request.response();
+  const body = await response.body();
+  console.log(body.byteLength);
+});
+await page.goto('https://playwright.dev', { waitUntil: 'domcontentloaded' });
+// Waits for all the reported 'request' events to resolve.
```
yury-s, line 3389: Not quite clear what `events to resolve` means, how about the following?
```suggestion
// Waits for all active 'request' event listeners to resolve.
```

yury-s (approved):

PR #32345, opened 2024-08-27
chore: split recorder into files

pavelfeldman:


packages/playwright-core/src/server/codegen/language.ts
```diff
@@ -32,6 +30,18 @@ export type LanguageGeneratorOptions = {
 export type LocatorType = 'default' | 'role' | 'text' | 'label' | 'placeholder' | 'alt' | 'title' | 'test-id' | 'nth' | 'first' | 'last' | 'has-text';
 export type LocatorBase = 'page' | 'locator' | 'frame-locator';
 
+export type FrameDescription = {
```
yury-s, line 33: Use those declared in types.ts instead?

yury-s (approved):

PR #32348, opened 2024-08-27
docs: global beforeEach/beforeAll hooks

### LINKED ISSUE #9468: [Feature] global beforeEach/beforeAll hooks
Instead of having the same test.beforeEach hook defined in every file, how do I only define the beforeEach hook in a single file (say hooks.js) at a global level, so then each file can use that hook?
In other words define my hooks in a single file once, but hooks are then used by all of my tests in all my spec files?
Thanks

yury-s:
Fixes https://github.com/microsoft/playwright/issues/9468

### DIFF
--- a/docs/src/test-fixtures-js.md
+++ b/docs/src/test-fixtures-js.md
@@ -725,17 +725,19 @@ export const test = base.extend({
 
 ## Adding global beforeEach/afterEach hooks
 
-[`method: Test.beforeEach`] and [`method: Test.afterEach`] hooks run before/after each test declared in the same file and same [`method: Test.describe`] block (if any). If you want to declare hooks that run before/after each test globally, you can declare them as auto fixtures with `scope: 'test'` like this:
+[`method: Test.beforeEach`] and [`method: Test.afterEach`] hooks run before/after each test declared in the same file and same [`method: Test.describe`] block (if any). If you want to declare hooks that run before/after each test globally, you can declare them as auto fixtures like this:
 
 ```ts title="fixtures.ts"
 import { test as base } from '@playwright/test';
 
 export const test = base.extend({
   forEachTest: [async ({ page, baseURL }, use) => {
+    // This code runs before every test.
     await page.goto('http://localhost:8000');
     await use();
+    // This code runs after every test.
     console.log('Last URL:', page.url());
-  }, { scope: 'test', auto: true } ],  // starts automatically for every test, we pass "auto" for that.
+  }, { auto: true }],  // starts automatically for every test, we pass "auto" for that.
 });
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/test-fixtures-js.md
```diff
@@ -722,3 +722,68 @@ export const test = base.extend({
...
+
+## Adding global beforeEach/afterEach hooks
+
+[`method: Test.beforeEach`] and [`method: Test.afterEach`] hooks run before/after each test declared in the same file and same [`method: Test.describe`] block (if any). If you want to declare hooks that run before/after each test globally, you can define them as auto fixtures with `scope: 'test'` like this:
+
+```ts title="fixtures.ts"
+import { test as base } from '@playwright/test';
+
+export const test = base.extend({
+  sharedAfterEach: [async ({ page, baseURL }, use) => {
```
pavelfeldman: This looks more like beforeEach

docs/src/test-fixtures-js.md
```diff
@@ -722,3 +722,68 @@ export const test = base.extend({
...
+    await page.goto(baseURL);
+    await use();
+  }, { scope: 'test', auto: true } ],  // starts automatically before every test, we pass "auto" for that.
+
+  sharedBeforeEach: [async ({ page }, use) => {
+    await use();
+    console.log('Test final URL:', page.url());
```
pavelfeldman: And this looks more like afterEach

docs/src/test-fixtures-js.md
```diff
@@ -722,3 +722,59 @@ export const test = base.extend({
...
+## Adding global beforeEach/afterEach hooks
+
+[`method: Test.beforeEach`] and [`method: Test.afterEach`] hooks run before/after each test declared in the same file and same [`method: Test.describe`] block (if any). If you want to declare hooks that run before/after each test globally, you can declare them as auto fixtures with `scope: 'test'` like this:
+
+```ts title="fixtures.ts"
+import { test as base } from '@playwright/test';
+
+export const test = base.extend({
+  forEachTest: [async ({ page, baseURL }, use) => {
+    await page.goto('http://localhost:8000');
```
pavelfeldman, line 735: Add a comment explaining that this code runs before the test

docs/src/test-fixtures-js.md
```diff
@@ -722,3 +722,59 @@ export const test = base.extend({
...
+    await use();
+    console.log('Last URL:', page.url());
```
pavelfeldman, line 737: Add a comment explaining that this code runs after the test

docs/src/test-fixtures-js.md
```diff
@@ -722,3 +722,59 @@ export const test = base.extend({
...
+  }, { scope: 'test', auto: true } ],  // starts automatically for every test, we pass "auto" for that.
```
pavelfeldman, line 738: I'd remove scope: test for simplicity

docs/src/test-fixtures-js.md
```diff
@@ -722,3 +722,59 @@ export const test = base.extend({
...
+
+[`method: Test.beforeAll`] and [`method: Test.afterAll`] hooks run before/after all tests declared in the same file and same [`method: Test.describe`] block (if any), once per worker process. If you want to declare hooks
+that run before/after all tests in every file, you can declare them as auto fixtures with `scope: 'worker'` as follows:
+
+```ts title="fixtures.ts"
+import { test as base } from '@playwright/test';
+
+export const test = base.extend({
+  forEachWorker: [async ({ browser }, use) => {
+    console.log('Before All', browser.version());
```
pavelfeldman, line 763: Users are often confused why Before All runs more than once, this only adds to the confusion. At the very least, say "`Before worker #${workerId}`"

pavelfeldman (approved):

docs/src/test-fixtures-js.md
yury-s: Removed.

yury-s:
Looks like the type inference is busted without explicit template parameter, changing the declaration to 
```js
export const test = base.extend<{ forEachTest: void }>({
...
```
should resolve the problem. I'll update the docs.

PR #32349, opened 2024-08-28
chore: pass explicit recorder app factory

pavelfeldman:


packages/playwright-core/src/cli/program.ts
```diff
@@ -571,7 +571,6 @@ async function codegen(options: Options & { target: string, output?: string, tes
     mode: 'recording',
     testIdAttributeName,
     outputFile: outputFile ? path.resolve(outputFile) : undefined,
-    handleSIGINT: false,
```
yury-s, line 574: is it related to this PR?

yury-s (approved):

packages/playwright-core/src/cli/program.ts
pavelfeldman: Yes, I don't think we need the handleSIGINT: true mode for recorder, at all, so removing.

PR #32351, opened 2024-08-28
fix(ui): respect --output param

### LINKED ISSUE #32331: [Bug]: UI doesn't call setup and passes incorrect outputDir
### Version

1.46.1

### Steps to reproduce

Note: the tests don't matter, it's purely a config & setup/teardown issue.

1. Set `globalSetup` to a file which contains `console.log("setup", config.projects[0].outputDir)`
2. Set `globalTeardown` to a file which contains `console.log("teardown", config.projects[0].outputDir)`
3. Run `playwright test --output blah` and observe that setup and teardown were called and logged appropriately
4. Run `playwright test --output blah --ui` and observe that setup was not called and teardown was called with the incorrect outputDir.

### Expected behavior

When running in UI mode, I expected setup to be called before the test started and for the correct config value to be passed. And I expected teardown to be called with the correct config value.

### Actual behavior

Setup was not called in UI mode and teardown was called with an incorrect (default) value.

### Additional context

The last version where this worked correctly was `1.42`. This has broken as of `1.43` and is still broken as of the latest at the time of writing `1.46.1`.

Related to https://github.com/microsoft/playwright/issues/30886

### Environment

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
    Memory: 10.74 GB / 15.95 GB
  Binaries:
    Node: 20.13.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.5.2 - C:\Program Files\nodejs\npm.CMD
  IDEs:
    VSCode: 1.92.2 - C:\Users\User\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 4.4.23 - C:\Program Files\Git\usr\bin\bash.EXE
  npmPackages:
    @playwright/test: ^1.43.0 => 1.43.0
```

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32331

We're already passing the `outputDir` param to the UI, but the UI isn't passing it back to the TestServer. This PR fixes that. I've added it to `listTests`, which is requires to that `TestServerDispatcher#_ignoredProjectOutputs` is populated with the correct output dir. And i've added it to `runGlobalSetup`, which is what the bug report was about.

### DIFF
--- a/packages/playwright/src/isomorphic/testServerInterface.ts
+++ b/packages/playwright/src/isomorphic/testServerInterface.ts
@@ -44,7 +44,7 @@ export interface TestServerInterface {
 
   installBrowsers(params: {}): Promise<void>;
 
-  runGlobalSetup(params: {}): Promise<{
+  runGlobalSetup(params: { outputDir?: string }): Promise<{
     report: ReportEntry[],
     status: reporterTypes.FullResult['status']
   }>;
@@ -81,6 +81,7 @@ export interface TestServerInterface {
     locations?: string[];
     grep?: string;
     grepInvert?: string;
+    outputDir?: string;
   }): Promise<{
     report: ReportEntry[],
     status: reporterTypes.FullResult['status']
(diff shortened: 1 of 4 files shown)


CreativeTechGuy:
Does this fix it for both setup and teardown?

Skn0tt:
I think so, but i can also add another test to prove that.

Skn0tt:
> Does this fix it for both setup and teardown?
It does: https://github.com/microsoft/playwright/pull/32351/commits/8f4b6209d331fc2ace727199e3ee719e0b8c09b7

pavelfeldman (approved):

PR #32366, opened 2024-08-29
feat(test runner): make `expect.extend` immutable

Skn0tt:
Changes `expect.extend` behaviour so that it doesn't mutate the global instance and behaves closer to what users expect. This is formally a breaking change, and I had to remove a test that asserts the breaking behaviour.

TODO:
- [x] decide wether this is a separate method or a flag for `expect.extend`
- [x] figure out if we need to change docs

### DIFF
--- a/packages/playwright/src/matchers/expect.ts
+++ b/packages/playwright/src/matchers/expect.ts
@@ -16,6 +16,7 @@
 
 import {
   captureRawStack,
+  createGuid,
   isString,
   pollAgainstDeadline } from 'playwright-core/lib/utils';
 import type { ExpectZone } from 'playwright-core/lib/utils';
@@ -61,7 +62,6 @@ import {
 import { zones } from 'playwright-core/lib/utils';
 import { TestInfoImpl } from '../worker/testInfo';
 import { ExpectError, isExpectError } from './matcherHint';
-import { randomUUID } from 'node:crypto';
 
 // #region
 // Mirrored from https://github.com/facebook/jest/blob/f13abff8df9a0e1148baf3584bcde6d1b479edc7/packages/expect/src/print.ts
@@ -131,7 +131,7 @@ function createExpect(info: ExpectMetaInfo, prefix: string[] = [], parentPrefixe
 
       if (property === 'extend') {
         return (matchers: any) => {
-          const qualifier = [...prefix, randomUUID()];
+          const qualifier = [...prefix, createGuid()];
 
           const wrappedMatchers: any = {};
           for (const [name, matcher] of Object.entries(matchers)) {
--- a/packages/playwright/src/matchers/expect.ts
+++ b/packages/playwright/src/matchers/expect.ts
@@ -189,7 +189,7 @@ function createExpect(info: ExpectMetaInfo, prefix: string[] = [], parentPrefixe
         newInfo.pollIntervals = configuration._poll.intervals;
       }
     }
-    return createExpect(newInfo);
+    return createExpect(newInfo, prefix, parentPrefixes);
   };
 
   return expectInstance;
--- a/packages/playwright/src/matchers/expect.ts
+++ b/packages/playwright/src/matchers/expect.ts
@@ -110,25 +110,6 @@ function createMatchers(actual: unknown, info: ExpectMetaInfo, prefix: string[])
 }
 
 function createExpect(info: ExpectMetaInfo, prefix: string[] = []) {
-
-  function extend(matchers: any, qualifier: string) {
-    const wrappedMatchers: any = {};
-    for (const [name, matcher] of Object.entries(matchers)) {
-      wrappedMatchers[qualifier + name] = function(...args: any[]) {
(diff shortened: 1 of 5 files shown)


packages/playwright/src/matchers/expect.ts
```diff
@@ -61,6 +61,7 @@ import {
 import { zones } from 'playwright-core/lib/utils';
 import { TestInfoImpl } from '../worker/testInfo';
 import { ExpectError, isExpectError } from './matcherHint';
+import { randomUUID } from 'node:crypto';
```
pavelfeldman, line 64: We are using createGuid() in our source base - Playwright was originally targeting Node 12, so consider using it for consistency, it has the same power.

packages/playwright/src/matchers/expect.ts
```diff
@@ -241,13 +252,28 @@ type ExpectMetaInfo = {
...
+    this._prefixes = prefixes;
   }
 
   get(target: Object, matcherName: string | symbol, receiver: any): any {
     let matcher = Reflect.get(target, matcherName, receiver);
+
+    if (typeof matcherName === 'string') {
+      for (const prefix of this._prefixes) {
+        for (let i = prefix.length; i > 0; i--) {
+          const qualifiedName = `${prefix.slice(0, i).join(':')}$${matcherName}`;
```
pavelfeldman, line 268: It would be nice to extract this key generation since it is used in two distant places.

packages/playwright/src/matchers/expect.ts
```diff
@@ -241,13 +252,28 @@ type ExpectMetaInfo = {
...
-  constructor(info: ExpectMetaInfo) {
+  constructor(info: ExpectMetaInfo, prefixes: string[][]) {
     this._info = { ...info };
+    this._prefixes = prefixes;
   }
 
   get(target: Object, matcherName: string | symbol, receiver: any): any {
     let matcher = Reflect.get(target, matcherName, receiver);
+
+    if (typeof matcherName === 'string') {
```
pavelfeldman, line 265: It looks like moving this block below eliminates the check.

packages/playwright/src/matchers/expect.ts
```diff
@@ -104,11 +105,13 @@ export const printReceivedStringContainExpectedResult = (
 
 type ExpectMessage = string | { message?: string };
 
-function createMatchers(actual: unknown, info: ExpectMetaInfo): any {
-  return new Proxy(expectLibrary(actual), new ExpectMetaInfoProxyHandler(info));
+function createMatchers(actual: unknown, info: ExpectMetaInfo, prefix: string[], parentPrefixes: string[][]): any {
```
pavelfeldman, line 108: It took me a while to understand why parentPrefixes are needed. I'd suggest a variation of the implementation where instead of parentPrefixes each expect contains a growing list of matchers it has been extended to so far. In that case merge could be implemented like so:

```js
export function mergeExpects(...expects: any[]) {
  let current = expect;
  for (const e of expects)
    current = current.extend(e[getMatchersSymbol]);
  return current;
}
```

Would that work?

packages/playwright/src/matchers/expect.ts
```diff
@@ -104,11 +105,13 @@ export const printReceivedStringContainExpectedResult = (
...
+  return new Proxy(expectLibrary(actual), new ExpectMetaInfoProxyHandler(info, [prefix, ...parentPrefixes]));
 }
 
-function createExpect(info: ExpectMetaInfo) {
+const getPrefixSymbol = Symbol('get prefix');
+
+function createExpect(info: ExpectMetaInfo, prefix: string[] = [], parentPrefixes: string[][] = []) {
```
pavelfeldman, line 114: In situations like this keep all the new params required, that would reduce the subsequent regressions where the param is omitted by accident.

Skn0tt (commented):
Good feedback, implemented all of it.

Skn0tt:
I took a look at our [docs for this](https://playwright.dev/docs/test-assertions#add-custom-matchers-using-expectextend). We're not showing the mutable case anywhere, so there's no changes needed.

Skn0tt:
@pavelfeldman could you take another look?

pavelfeldman (approved):

muhqu:
@Skn0tt this is indeed a breaking change. It causes custom asymmetric matchers to no longer get exposed on expect. With 1.47.1 the custom asymmetric matchers were accessible via `expect[matcherName]` even though playwright's type definition did not expose them. See https://github.com/microsoft/playwright/issues/32562#issuecomment-2363340070 

I've created 2 PRs, one branched from latest main (https://github.com/microsoft/playwright/pull/32740) and another one that branched right before this PR got merged (https://github.com/microsoft/playwright/pull/32739)

muhqu:
Ah… nevermind, got it working in https://github.com/microsoft/playwright/pull/32740 with https://github.com/microsoft/playwright/pull/32740/commits/978cd7d80a372808aa175ca92fa6735bbcba5074

PR #32375, opened 2024-08-29
chore(docs): fix APIResponse.headersArray() desc

PrzemyslawMalolepszy-TomTom:
Modify APIResponse.headersArray() description as it was mentioning request headers, while actually response headers are returned


PrzemyslawMalolepszy-TomTom:

mxschmitt (approved):

PR #32379, opened 2024-08-29
fix(trace viewer): fix memory leak

### LINKED ISSUE #32336: [Bug]: Playwright UI crashes with "Out of Memory"
### Version

1.46.1

### Steps to reproduce

* have a complex website to test 
* have a testcase with 100+ test.step(...) inside and 100+ of expect and actions inside of these - in my current case, the metadata tab says: Counts - pages: 1, actions: 387, events: 223
* run the test
* after completion click and/or hover through the actions tree - it will switch through the numberous snapshots of the website on the right side
* eventually the whole Plawright UI window crashes with an "Out of Memory" Error Message.

With any test of this size it is 100% reproducible within 10 seconds of browsing through the snapshots.
image
Anmerkung 2024-08-27 121156

### Expected behavior

no crash

### Actual behavior

crash!

### Additional context

Is there a way to reduce the memory footprint of playwright? We want to use it on non-dev notebooks with only 16 GB RAM, but that seems to be impossible with our scenario. 

### Environment

```shell
System:
    OS: Windows 10 10.0.19045
    Memory: 5.09 GB / 31.69 GB
  Binaries:
    Node: 20.11.0 - ~\scoop\apps\nodejs-lts\current\node.EXE
    Yarn: 4.3.1 - ~\scoop\apps\nodejs-lts\current\bin\yarn.CMD
    npm: 10.2.4 - ~\scoop\apps\nodejs-lts\current\npm.CMD
```

Skn0tt:
In the `visit` method, we currently cache the rendered HTML for every walked node. This re-use works well for traces that consist mostly of references to earlier snapshots.
But for traces that don't share much, this is a large memory overhead and leads to the memory crash documented in https://github.com/microsoft/playwright/issues/32336. For the algocracks amongst you, the current memory usage for an html tree $h$ is $\mathcal{O}(|h| * \text{height}(h))$.

This PR removes that cache from the nodes and replaces it with a snapshot-level cache, fixing the memory crash.
Traces *without* reference should not see a performance impact from this.
Traces *with* references will have slower initial rendering, but re-rendering maintains speed because of the snapshot-level cache.

Closes https://github.com/microsoft/playwright/issues/32336

### DIFF
--- a/packages/trace-viewer/src/snapshotRenderer.ts
+++ b/packages/trace-viewer/src/snapshotRenderer.ts
@@ -29,7 +29,7 @@ let cacheSize = 0;
 const cache = new Map<SnapshotRenderer, string>();
 const CACHE_SIZE = 300000000; // 300mb
 
-function cacheAndReturn(key: SnapshotRenderer, compute: () => string): string {
+function lruCache(key: SnapshotRenderer, compute: () => string): string {
   if (cache.has(key)) {
     const value = cache.get(key)!;
     // reinserting makes this the least recently used entry
@@ -42,9 +42,9 @@ function cacheAndReturn(key: SnapshotRenderer, compute: () => string): string {
   const result = compute();
 
   while (cache.size && cacheSize + result.length > CACHE_SIZE) {
-    const first = cache.keys().next().value;
-    cacheSize -= cache.get(first)!.length;
-    cache.delete(first);
+    const [firstKey, firstValue] = cache.entries().next().value;
+    cacheSize -= firstValue.length;
+    cache.delete(firstKey);
   }
 
   cache.set(key, result);
@@ -151,7 +151,7 @@ export class SnapshotRenderer {
     };
 
     const snapshot = this._snapshot;
-    const html = cacheAndReturn(this, () => {
+    const html = lruCache(this, () => {
       visit(snapshot.html, this._index, undefined, undefined);
 
       const html = result.join('');
--- a/packages/trace-viewer/src/snapshotRenderer.ts
+++ b/packages/trace-viewer/src/snapshotRenderer.ts
@@ -41,7 +41,7 @@ function cacheAndReturn(key: SnapshotRenderer, compute: () => string): string {
 
   const result = compute();
 
-  while (cacheSize + result.length > CACHE_SIZE) {
+  while (cache.size && cacheSize + result.length > CACHE_SIZE) {
     const first = cache.keys().next().value;
     cacheSize -= cache.get(first)!.length;
     cache.delete(first);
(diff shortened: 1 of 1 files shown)


packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -25,6 +25,34 @@ function isSubtreeReferenceSnapshot(n: NodeSnapshot): n is SubtreeReferenceSnaps
...
+    // reinserting makes this the least recently used entry
+    cache.delete(key);
+    cache.set(key, value);
+    return value;
+  }
+
+
+  const result = compute();
+
+  while (cacheSize + result.length > CACHE_SIZE) {
```
dgozman, line 44: To avoid npe when `result.length > CACHE_SIZE`.

```suggestion
  while (cache.size && cacheSize + result.length > CACHE_SIZE) {
```

packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -25,6 +25,34 @@ function isSubtreeReferenceSnapshot(n: NodeSnapshot): n is SubtreeReferenceSnaps
   return Array.isArray(n) && Array.isArray(n[0]);
 }
 
+let cacheSize = 0;
+const cache = new Map<SnapshotRenderer, string>();
+const CACHE_SIZE = 300000000; // 300mb
+
+function cacheAndReturn(key: SnapshotRenderer, compute: () => string): string {
```
dgozman, line 32: nit: now that it's LRU, I'd call this something like `lruCache(...)`.

packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -25,6 +25,34 @@ function isSubtreeReferenceSnapshot(n: NodeSnapshot): n is SubtreeReferenceSnaps
...
+    return value;
+  }
+
+
+  const result = compute();
+
+  while (cacheSize + result.length > CACHE_SIZE) {
+    const first = cache.keys().next().value;
+    cacheSize -= cache.get(first)!.length;
+    cache.delete(first);
```
dgozman, line 47:
```suggestion
    const [firstKey, firstValue] = cache.entries().next().value;
    cacheSize -= firstValue.length;
    cache.delete(firstKey);
```

dgozman (approved):

packages/trace-viewer/src/snapshotRenderer.ts
```diff
@@ -25,6 +25,34 @@ function isSubtreeReferenceSnapshot(n: NodeSnapshot): n is SubtreeReferenceSnaps
   return Array.isArray(n) && Array.isArray(n[0]);
 }
 
+let cacheSize = 0;
+const cache = new Map<SnapshotRenderer, string>();
+const CACHE_SIZE = 300000000; // 300mb
```
mxschmitt, line 30:
```suggestion
const CACHE_SIZE = 300_000_000; // 300mb
```
nit

PR #32383, opened 2024-08-29
feat(api): add method to force garbage collection

### LINKED ISSUE #32278: [Feature]: A way to trigger garbage collection when targeting WebKit
### 🚀 Feature Request

We would like to be able to trigger garbage collection when targeting WebKit.

Currently, it is possible to trigger garbage collection when targeting Chromium by sending the `HeapProfiler.collectGarbage` command directly to a `CDPSession`. There is no equivalent to `CDPSession` for other browsers.

Some ways to support this would be:

1. Add a function to `Page` that triggers a garbage collection.
2. Add a function to `BrowserContext` that provides direct access to the WebKit Remote Debugging Protocol session, analogous to `newCDPSession`. This could be used to trigger garbage collection by sending the `Heap.gc` command.

I'm not sure whether it would be better to do (1) or (2), or something completely different. Option (1) seems like it would be straightforward to implement for Chromium and WebKit since both debugging protocols already have a command that triggers garbage collection. It would be trickier for Firefox though as there is no equivalent command. I could be missing something though!

I'm submitting this request on behalf of Adobe. We are willing to do the implementation work for this feature. I Appreciate any feedback on how best to approach this!

### Example

We are working on an app that has strict memory requirements, particularly when running on mobile devices. We have a set of tests that we use to detect whether objects are leaked during certain workflows. At a high level, the test flow looks like this:

1. Set up the test.
2. Construct a set of `WeakRef`s to objects that we want to track for leaks.
3. Run the workflow.
4. Trigger a garbage collection.
5. Check whether the objects in the tracked set are `undefined`, indicating that they have been collected.

It is critical that a garbage collection be triggered before checking the tracked set to ensure that `WeakRef`s to objects that are no longer referenced are cleared.

The test flow looks something like this:

```js
// Track object for leak testing.
this.page.evaluate(() => {
    const object = getObjectThatMightLeak();
    window.trackedObject = new WeakRef(object);
});

// Run the workflow that might result in the tracked object being leaked
runWorkflow();

// Trigger garbage collection
this.page.collectGarbage();

// Alternative:
// const session = await this.page.context().newWKRDPSession(this.page);
// await session.send("Heap.gc");
// await session.detach();

// Verify that the tracked object was collected.
this.page.evaluate(() => {
    expect(window.trackedObject.deref()).toBe(undefined);
});
```

### Motivation

Currently, we are only able to run our leak tests against Chromium, because it's the only browser engine for which we are able to trigger garbage collection. This new functionality would make it possible for us (and others) to run leak tests like this targeting both Chromium and WebKit.

marmphco:
Add a method to Page that forces garbage collection. For now this method is supported only in Chromium and WebKit.

~I made several attempts to get this to work in Firefox, using both [Components.utils.schedulePreciseGC](https://udn.realityripple.com/docs/Mozilla/Tech/XPCOM/Language_Bindings/Components.utils.schedulePreciseGC) and [Components.utils.forceGC](https://udn.realityripple.com/docs/Mozilla/Tech/XPCOM/Language_Bindings/Components.utils.forceGC), but was unsuccessful. See https://github.com/marmphco/playwright/commit/42bd3ef906b43faacf329432600f3131203b8d1a for those changes. Either garbage collection isn't running or its effects aren't visible immediately to the test code like they are for WebKit and Chromium. I don't have a good enough understanding of Firefox's internals to figure this out so I left it as unsupported for now. If this is unacceptable I can try to implement via a different route. Suggestions for better approaches would be appreciated 🙂~

7c826ee includes support for Firefox.

Closes #32278

### DIFF
--- a/packages/playwright-core/src/server/firefox/ffPage.ts
+++ b/packages/playwright-core/src/server/firefox/ffPage.ts
@@ -401,7 +401,7 @@ export class FFPage implements PageDelegate {
   }
 
   async forceGarbageCollection(): Promise<void> {
-    // Not supported in Firefox.
+    await this._session.send('Heap.collectGarbage');
   }
 
   async addInitScript(initScript: InitScript, worldName?: string): Promise<void> {
--- a/packages/playwright-core/src/server/firefox/ffPage.ts
+++ b/packages/playwright-core/src/server/firefox/ffPage.ts
@@ -400,6 +400,10 @@ export class FFPage implements PageDelegate {
     return success;
   }
 
+  async forceGarbageCollection(): Promise<void> {
+    // Not supported in Firefox.
+  }
+
   async addInitScript(initScript: InitScript, worldName?: string): Promise<void> {
     this._initScripts.push({ initScript, worldName });
     await this._session.send('Page.setInitScripts', { scripts: this._initScripts.map(s => ({ script: s.initScript.source, worldName: s.worldName })) });
--- a/packages/protocol/src/protocol.yml
+++ b/packages/protocol/src/protocol.yml
@@ -1430,7 +1430,7 @@ Page:
         slowMo: true
         snapshot: true
 
-    forceGarbageCollection: {}
+    forceGarbageCollection:
 
     registerLocatorHandler:
       parameters:
--- a/packages/protocol/src/protocol.yml
+++ b/packages/protocol/src/protocol.yml
@@ -1430,6 +1430,8 @@ Page:
         slowMo: true
         snapshot: true
 
+    forceGarbageCollection: {}
+
     registerLocatorHandler:
       parameters:
         selector: string
--- a/browser_patches/firefox/juggler/protocol/PageHandler.js
+++ b/browser_patches/firefox/juggler/protocol/PageHandler.js
@@ -256,6 +256,14 @@ class PageHandler {
     return await this._contentPage.send('disposeObject', options);
   }
 
+  async ['Heap.collectGarbage']() {
+    await new Promise((resolve) => {
+      Cu.schedulePreciseGC(() => {
+        resolve();
+      });
+    });
+  }
+
   async ['Network.getResponseBody']({requestId}) {
     return this._pageNetwork.getResponseBody(requestId);
   }
--- a/browser_patches/firefox/juggler/protocol/PageHandler.js
+++ b/browser_patches/firefox/juggler/protocol/PageHandler.js
@@ -257,11 +257,10 @@ class PageHandler {
   }
 
   async ['Heap.collectGarbage']() {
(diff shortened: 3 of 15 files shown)


marmphco:
@microsoft-github-policy-service agree company="Adobe"

packages/playwright-core/src/server/firefox/ffPage.ts
```diff
@@ -399,6 +399,10 @@ export class FFPage implements PageDelegate {
     return success;
   }
 
+  async forceGarbageCollection(): Promise<void> {
+    // Not supported in Firefox.
```
pavelfeldman, line 403: We don't do that, should be supported to make its way into Playwright.

yury-s:
> Either garbage collection isn't running or its effects aren't visible immediately to the test code like they are for WebKit and Chromium.

You might need to run several gc cycles to free up objects reachable through DOM bindings (could be the case for WeakRefs too), at least that was the case in Chromium.

packages/playwright-core/src/server/firefox/ffPage.ts
marmphco: Understood, I'll look into it further.

marmphco:
Not much luck running multiple GC cycles, similarly to https://searchfox.org/mozilla-central/source/devtools/server/performance/memory.js#398. Still investigating other ways to get this to work.

marmphco:
Added Firefox support in d2add6900212288f6ac0d1ad1eaf59b9f18da5bf and 7c826ee427b8b14ec5b7c8f6476d6df5d019bc4f. The problem was that child content processes were not being notified to run garbage collection.

marmphco:
@pavelfeldman Can you take another look at this when you have some time? It's important for one of our products at Adobe. Much appreciated!

mxschmitt:
@marmphco we manage browser patches in a separate internal repository, where we have access to better infrastructure. I landed the changes there and they are compiling. Should take a few hours, then you can re-base and have a Firefox available with your Heap change in it. I'll ping you once thats finished.

mxschmitt:
New Firefox is in: https://github.com/microsoft/playwright/pull/32614

packages/protocol/src/protocol.yml
```diff
@@ -1430,6 +1430,8 @@ Page:
         slowMo: true
         snapshot: true
 
+    forceGarbageCollection: {}
```
mxschmitt, line 1433:
```suggestion
    forceGarbageCollection:
```

mxschmitt (approved):

marmphco:
@mxschmitt Ah sorry about the lint issue, forgot to re-run after rebasing. Thanks for fixing!

PR #32389, opened 2024-08-30
feat(html reporter): open html attachments in new tab

### LINKED ISSUE #32281: [Feature]: HTML reporter: preview html attachments
### 🚀 Feature Request

Right now the generated html reporter provides an inline preview of attachments for `text/*` as string.
It would be nice to have rendered html for `text/html` instead.

### Example

We use [axe](https://github.com/dequelabs/axe-core) for accessibility testing in playwright.
The resulting report is appended to the `TestInfo` as attachment.

Reports are deployed internally to be shared with non-technical personnel.
The plain text preview of the html report is barely human readable and the need of downloading and opening the file locally is inconvenient.

### Motivation

It would allow to customize and extend the generated report. Using the existing attachment 
features, the changes needed are minimal.

Images and videos are previewed already.

This feature could also be opt-in with a feature flag:

```ts
const body = `<h4>Report</h4>...`;
testInfo.attach('report.html', { body, contentType: 'text/html', preview: true })
```

or a separate property allowing different contents for preview and download:

```ts
const body = `<h1>Report</h1>...`;
const preview = `<h4>Report Preview</h4>...`;
testInfo.attach('report.html', { body, contentType: 'text/html', preview })
```

_If wanted, I'd volunteer to draft a PR. 😊_

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32281.

HTML attachments get a linkified name that opens the attachment in a new tab.

### DIFF
--- a/packages/html-reporter/src/testResultView.tsx
+++ b/packages/html-reporter/src/testResultView.tsx
@@ -67,15 +67,16 @@ export const TestResultView: React.FC<{
   anchor: 'video' | 'diff' | '',
 }> = ({ result, anchor }) => {
 
-  const { screenshots, videos, traces, otherAttachments, diffs } = React.useMemo(() => {
+  const { screenshots, videos, traces, otherAttachments, diffs, htmls } = React.useMemo(() => {
     const attachments = result?.attachments || [];
     const screenshots = new Set(attachments.filter(a => a.contentType.startsWith('image/')));
     const videos = attachments.filter(a => a.name === 'video');
     const traces = attachments.filter(a => a.name === 'trace');
+    const htmls = attachments.filter(a => a.contentType.startsWith('text/html'));
     const otherAttachments = new Set<TestAttachment>(attachments);
-    [...screenshots, ...videos, ...traces].forEach(a => otherAttachments.delete(a));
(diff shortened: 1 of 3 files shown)


packages/html-reporter/src/testResultView.tsx
```diff
@@ -135,7 +136,10 @@ export const TestResultView: React.FC<{
       </div>)}
     </AutoChip>}
 
-    {!!otherAttachments.size && <AutoChip header='Attachments'>
+    {!!(otherAttachments.size + htmls.length) && <AutoChip header='Attachments'>
+      {[...htmls].map((a, i) => (
```
pavelfeldman, line 140: Not sure if it bothers us, but attachments of different types are out of order now. I.e. if I have text, html, text, html sequence they end up grouped together.

pavelfeldman (approved):

packages/html-reporter/src/testResultView.tsx
Skn0tt: that was already the case for screenshots, videos and traces. I guess one could think of them differently because they're "special" attachments, but the same argument could be made for HTML attachments really.

🤷

PR #32407, opened 2024-09-02
fix(test runner): fix types to allow calling custom matchers on `expect.poll`

### LINKED ISSUE #32408: [Bug]: `expect.poll().toPassCustomMatcher()` should not be a type error
### Version

latest

### Steps to reproduce

```ts
import { test, expect as baseExpect } from '@playwright/test';

const expect = baseExpect.extend({
  toBeWithinRange(received, floor, ceiling) {
    return { passed: true, message: () => "passed" }
  },
});

test('should poll', async () => {
  let i = 0;
  await expect.poll(() => ++i).toBeWithinRange(3, Number.MAX_VALUE);
});
```

### Expected behavior

It should typecheck just fine.

### Actual behavior

Typescript complains that `toBeWithinRange` doesn't exist.

### Additional context

This is because in https://github.com/microsoft/playwright/blob/5c2e9962b440e72bff0196bf32b917c4dac632d4/utils/generate_types/overrides-test.d.ts#L411-L416, we use `BaseMatcher` so that none of our custom async matchers turn up. It should also include `ExtendedMatchers`.

### Environment

```shell
irrelevant
```

Skn0tt:
The `'should support custom matchers'` test asserts that the functionality works, but it was a type error. This PR updates the types so that it's allowed.

Closes https://github.com/microsoft/playwright/issues/32408

### DIFF
--- a/utils/generate_types/overrides-test.d.ts
+++ b/utils/generate_types/overrides-test.d.ts
@@ -405,15 +405,17 @@ type MakeMatchers<R, T, ExtendedMatchers> = {
   rejects: MakeMatchers<Promise<R>, any, ExtendedMatchers>;
 } & IfAny<T, AllMatchers<R, T>, SpecificMatchers<R, T> & ToUserMatcherObject<ExtendedMatchers, T>>;
 
+type PollMatchers<R, T, ExtendedMatchers> = {
+  /**
+   * If you know how to test something, `.not` lets you test its opposite.
+   */
+  not: PollMatchers<R, T, ExtendedMatchers>;
+} & BaseMatchers<R, T> & ToUserMatcherObject<ExtendedMatchers, T>;
+
 export type Expect<ExtendedMatchers = {}> = {
   <T = unknown>(actual: T, messageOrOptions?: string | { message?: string }): MakeMatchers<void, T, ExtendedMatchers>;
   soft: <T = unknown>(actual: T, messageOrOptions?: string | { message?: string }) => MakeMatchers<void, T, ExtendedMatchers>;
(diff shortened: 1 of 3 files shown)


utils/generate_types/overrides-test.d.ts
```diff
@@ -405,15 +405,17 @@ type MakeMatchers<R, T, ExtendedMatchers> = {
...
 export type Expect<ExtendedMatchers = {}> = {
   <T = unknown>(actual: T, messageOrOptions?: string | { message?: string }): MakeMatchers<void, T, ExtendedMatchers>;
   soft: <T = unknown>(actual: T, messageOrOptions?: string | { message?: string }) => MakeMatchers<void, T, ExtendedMatchers>;
-  poll: <T = unknown>(actual: () => T | Promise<T>, messageOrOptions?: string | { message?: string, timeout?: number, intervals?: number[] }) => BaseMatchers<Promise<void>, T> & {
-    /**
-     * If you know how to test something, `.not` lets you test its opposite.
-     */
-     not: BaseMatchers<Promise<void>, T>;
-  };
+  poll: <T = unknown>(actual: () => T | Promise<T>, messageOrOptions?: string | { message?: string, timeout?: number, intervals?: number[] }) => PollMatchers<Promise<void>,T, ExtendedMatchers>
```
dgozman, line 418:
```suggestion
  poll: <T = unknown>(actual: () => T | Promise<T>, messageOrOptions?: string | { message?: string, timeout?: number, intervals?: number[] }) => PollMatchers<Promise<void>, T, ExtendedMatchers>;
```

dgozman (approved):

PR #32425, opened 2024-09-03
chore(test runner): include projects in teleReceiver's config

Skn0tt:
Follow-up to https://github.com/microsoft/playwright/pull/32156#discussion_r1741628770. Transmits `config.projects` via the TeleEmitter, so it can be read from UI and watch mode.

Let me know where I should add more tests for this.

### DIFF
--- a/packages/playwright/src/reporters/merge.ts
+++ b/packages/playwright/src/reporters/merge.ts
@@ -304,7 +304,7 @@ function mergeConfigs(to: JsonConfig, from: JsonConfig): JsonConfig {
       actualWorkers: (to.metadata.actualWorkers || 0) + (from.metadata.actualWorkers || 0),
     },
     workers: to.workers + from.workers,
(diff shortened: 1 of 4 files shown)


packages/playwright/src/reporters/merge.ts
```diff
@@ -288,6 +289,11 @@ function mergeConfigureEvents(configureEvents: JsonEvent[], rootDirOverride: str
   };
 }
 
+function dedupByName<T extends { name: string }>(a: T[], b: T[]) {
```
dgozman, line 292: I don't really like deduping by name, because similarly named projects from different bots could be vastly different. I don't see an easy way to achieve what we want right now: from one side, we don't want 10 chromium projects, but from the other side we cannot just dedupe them.

I guess, we should just list tests without filters once in the watch mode and drop this PR.

PR #32456, opened 2024-09-05
fix: apply sort directly in injected code

vbrajon:
This change do not modify the behavior.
I would benefit from it on a project where "sort" is overwritten to not mutate the initial array.


pavelfeldman:
Don't go against JS, the world will fall apart.

PR #32458, opened 2024-09-05
chore: roll expect and move it to third party

pavelfeldman:


mxschmitt:
Do we have patches we want to put on top of expect? I'd try to land them upstream instead, seems a lot of effort to maintain it that way.

pavelfeldman:
> Do we have patches we want to put on top of expect? I'd try to land them upstream instead, seems a lot of effort to maintain it that way.

The idea is to improve the expect integration into Playwright. I.e. simplify the glue and the workarounds we have today.

PR #32474, opened 2024-09-05
chore: support bidi connection to chromium

yury-s:


packages/playwright-core/src/server/bidi/bidiBrowserType.ts
```diff
@@ -0,0 +1,205 @@
+/**
```
pavelfeldman, line 1: The rename diff looks broken, can't tell what changed here.

pavelfeldman (approved):

PR #32491, opened 2024-09-06
fix(test runner): allow directory imports with path mapping

### LINKED ISSUE #31811: [Regression]: ESM Directory Import is not working anymore with Playwright's ESM Loader activated

### LINKED ISSUE #32480: [Regression]: `tsconfig.json`'s `compilerOptions.paths` no longer working in 1.47
### Version

1.47.0

### Steps to reproduce

1. Add paths to `tsconfig.json`
    ```
    {
      "compilerOptions": {
        "paths": {
          "#/*": ["./src/*"]
        }
      }
    }
    ```
2. Import something using that path, i.e. `#/utils/my-util`.

### Expected behavior

This worked as expected in 1.46.

### Actual behavior

Error:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '#/utils/my-util' imported from <path>/tests/my-test.ts
```

### Additional context

I assume this is related to the `tsconfig` option in 1.47, but didn't dig into the issue yet sorry.

### Environment

```shell
System:
  OS: macOS 14.6.1
  CPU: (8) arm64 Apple M3
  Memory: 256.58 MB / 24.00 GB
Binaries:
  Node: 20.14.0 - ~/.local/state/fnm_multishells/35649_1725599073952/bin/node
  npm: 10.7.0 - ~/.local/state/fnm_multishells/35649_1725599073952/bin/npm
  pnpm: 9.9.0 - ~/.local/state/fnm_multishells/35649_1725599073952/bin/pnpm
IDEs:
  VSCode: 1.93.0 - /usr/local/bin/code
Languages:
  Bash: 3.2.57 - /bin/bash
```

dgozman:
We now hopefully align with `moduleResolution: bundler` tsconfig option, allowing directory imports in every scenario, and allowing proper module imports when not going through the type mapping.

This regressed in #32078. Fixes #32480, fixes #31811.

### DIFF
--- a/packages/playwright/src/util.ts
+++ b/packages/playwright/src/util.ts
@@ -307,8 +307,23 @@ function folderIsModule(folder: string): boolean {
   return require(packageJsonPath).type === 'module';
 }
 
-// This follows the --moduleResolution=bundler strategy from tsc.
-// https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#moduleresolution-bundler
+const packageJsonMainFieldCache = new Map<string, string | undefined>();
+
+function getMainFieldFromPackageJson(packageJsonPath: string) {
+  if (!packageJsonMainFieldCache.has(packageJsonPath)) {
+    let mainField: string | undefined;
+    try {
+      mainField = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8')).main;
+    } catch {
+    }
+    packageJsonMainFieldCache.set(packageJsonPath, mainField);
+  }
+  return packageJsonMainFieldCache.get(packageJsonPath);
+}
+
+// This method performs "file extension subsitution" to find the ts, js or similar source file
+// based on the import specifier, which might or might not have an extension. See TypeScript docs:
+// https://www.typescriptlang.org/docs/handbook/modules/reference.html#file-extension-substitution.
 const kExtLookups = new Map([
   ['.js', ['.jsx', '.ts', '.tsx']],
   ['.jsx', ['.tsx']],
@@ -316,7 +331,7 @@ const kExtLookups = new Map([
   ['.mjs', ['.mts']],
   ['', ['.js', '.ts', '.jsx', '.tsx', '.cjs', '.mjs', '.cts', '.mts']],
 ]);
-export function resolveImportSpecifierExtension(resolved: string): string | undefined {
+function resolveImportSpecifierExtension(resolved: string): string | undefined {
   if (fileExists(resolved))
     return resolved;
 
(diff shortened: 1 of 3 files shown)


tests/playwright-test/resolver.spec.ts
```diff
@@ -569,8 +569,8 @@ test('should resolve paths relative to the originating config when extending and
   expect(result.exitCode).toBe(0);
 });
 
-test('should import packages with non-index main script through path resolver', async ({ runInlineTest }) => {
-  const result = await runInlineTest({
+test('should import packages with non-index main script through path resolver', async ({ runInlineTest, runTSC }) => {
```
pavelfeldman, line 572:
```suggestion
test('should import folders with non-index main script through path resolver in CJS', async ({ runInlineTest, runTSC }) => {
```

tests/playwright-test/resolver.spec.ts
```diff
@@ -589,30 +589,38 @@ test('should import packages with non-index main script through path resolver',
...
   expect(result.output).not.toContain(`find module`);
-  expect(result.output).toContain(`foo=42`);
+
+  const tscResult = await runTSC(files);
+  expect(tscResult.exitCode).toBe(0);
 });
 
-test('should not honor `package.json#main` field in ESM mode', async ({ runInlineTest }) => {
-  const result = await runInlineTest({
+test('should honor `package.json#main` field in ESM mode', async ({ runInlineTest, runTSC }) => {
```
pavelfeldman, line 617: note to myself: same semantics, could be a cycle

tests/playwright-test/resolver.spec.ts
```diff
@@ -626,32 +634,43 @@ test('should not honor `package.json#main` field in ESM mode', async ({ runInlin
...
+  expect(result.exitCode).toBe(0);
+  expect(result.passed).toBe(1);
 
+  const tscResult = await runTSC(files);
+  expect(tscResult.exitCode).toBe(0);
+});
 
-test('does not honor `exports` field after type mapping', async ({ runInlineTest }) => {
-  const result = await runInlineTest({
+test('does not honor `exports` field after type mapping', async ({ runInlineTest, runTSC }) => {
```
pavelfeldman, line 661: ... in CJS

tests/playwright-test/resolver.spec.ts
```diff
@@ -767,64 +848,209 @@ test('should respect --tsconfig option', async ({ runInlineTest }) => {
   expect(result.output).not.toContain(`Could not`);
 });
 
-
-test('should resolve index.js in CJS after path mapping', async ({ runInlineTest }) => {
+test('should resolve index.js in CJS after path mapping', async ({ runInlineTest, runTSC }) => {
```
pavelfeldman, line 851:
```suggestion
test('should resolve index.js after path mapping in CJS', async ({ runInlineTest, runTSC }) => {
```

tests/playwright-test/playwright-test-fixtures.ts
```diff
@@ -367,8 +367,8 @@ export const test = base
 const TSCONFIG = {
   'compilerOptions': {
     'target': 'ESNext',
-    'moduleResolution': 'node',
-    'module': 'commonjs',
+    'moduleResolution': 'bundler',
```
pavelfeldman, line 370: why did this change?

packages/playwright/src/util.ts
```diff
@@ -332,13 +346,33 @@ export function resolveImportSpecifierExtension(resolved: string): string | unde
...
+      //  - `package.json#exports` is not respected, because it's not a proper module
+      //  - `package.json#main` is respected according to "folder as module" Node.js behavior
+      //  - `index.js` default is respected according to "folder as module" Node.js behavior
+      //
+      // More info:
+      //  - https://www.typescriptlang.org/docs/handbook/modules/reference.html#directory-modules-index-file-resolution
+      //  - https://nodejs.org/dist/latest-v20.x/docs/api/modules.html#folders-as-modules
+
+      const mainField = getMainFieldFromPackageJson(packageJsonPath);
+      const mainFieldResolved = mainField ? resolveImportSpecifierExtension(path.resolve(resolved, mainField), isPathMapping) : undefined;
```
pavelfeldman, line 363: Why do we resolve package's main via path mapping? I don't think we should respect aliases in package.json, so should we just defer to Node from here?

packages/playwright/src/util.ts
```diff
@@ -332,13 +346,33 @@ export function resolveImportSpecifierExtension(resolved: string): string | unde
...
+      return mainFieldResolved || resolveImportSpecifierExtension(path.join(resolved, 'index'), isPathMapping);
```
pavelfeldman, line 364: Ditto

pavelfeldman (approved):

tests/playwright-test/playwright-test-fixtures.ts
dgozman: This updates our default tsc config in tests to align with what we aim to support, particularly "bundler". Let me land it separately.

packages/playwright/src/util.ts
dgozman: We do not apply path mappings here. `resolveImportSpecifierExtension` handles extension substitution, similarly to  [typescript](https://www.typescriptlang.org/docs/handbook/modules/reference.html#file-extension-substitution).

Deferring to Node does not work, that's exactly the issue we are fixing. TypeScript allows ESM directory imports, while pure Node does not.

PR #32501, opened 2024-09-07
devops: run bidi tests on Firefox Stable

mxschmitt:


packages/playwright-core/src/server/bidi/bidiPage.ts
```diff
@@ -272,6 +272,8 @@ export class BidiPage implements PageDelegate {
   }
 
   private async _updateViewport(): Promise<void> {
+    if (this._browserContext._browser.options.name === 'firefox')
```
mxschmitt, line 275: I'd like to remove this change before landing this PR, without that, no test in Firefox is passing and ending up in:

```

    Error: browserContext.newPage: Protocol error (browsingContext.setViewport): unknown error
    Message: AbortError: Actor 'MessageHandlerFrame' destroyed before query 'MessageHandlerFrameParent:sendCommand' was resolved
```

```
npm run biditest -- --project=bidi-firefox-stable-page goto:22
```

packages/playwright-core/src/server/bidi/bidiPage.ts
```diff
@@ -272,6 +272,8 @@ export class BidiPage implements PageDelegate {
...
+      return;
```
mxschmitt, line 276:
```suggestion
```

tkarombang: Unsubscribe

Issue #22460, opened 2023-04-18
[BUG] serializedArgs.map is not a function when calling exposed function


eliassorensen:
I have noticed on several websites, that exposed functions are not working. Below is a working example of this issue on one URL. If you need more URLs where this occurs, just comment and I'll provide.

### System info
- Playwright Version: Version 1.32.3
- Operating System: Tested on Ubuntu 22.04
- Browser: Chromium
- Other info:

### Source code

- [ ] I provided exact source code that allows reproducing the issue locally.

```js
// repro.js
const { chromium } = require('playwright');

(async () => {
  function testInitScript(){
    if(window !== window.top) {
       return; // only run init script on root
    }

    setTimeout(function(){
        window.TestExposedFunction('This is a test');
    }, 2000);

  }
  const browser = await chromium.launch({ headless: false });
  const page = await browser.newPage();
  await page.exposeFunction("TestExposedFunction", (text) => console.log('TestExposedFunction', text));
  await page.addInitScript(testInitScript);

  await page.goto("https://essen.vienna.at/"); // ⬅️ This will cause an exception in the browser when calling the TestExposedFunction
  //await page.goto("https://www.google.com/"); // ⬅️ This will work just fine, and console.log the TestExposedFunction text

  await new Promise(resolve => setTimeout(resolve, 10000)); // Poor man's way of being able to wait a bit with the browser open
  
  await browser.close();
})();
```

**Steps**
- Run the repro.js (`node repro.js`) with each page.goto()

**Expected**

- Both URLs should give the same result (logging 'TestExposedFunction' with the text)

**Actual**

- essen.vienna.at does not console.log, but instead throw below exception:
![image](https://user-images.githubusercontent.com/1552175/232809385-36cfa090-9600-4293-8e96-a12cfd7196e1.png)

- google.com correctly outputs: `TestExposedFunction This is a test`

dgozman:
`https://essen.vienna.at/` deletes the `Array.map` function. I'll see what we can do to workaround that.

eliassorensen:
> `https://essen.vienna.at/` deletes the `Array.map` function. I'll see what we can do to workaround that.

Good catch, I didn't even think of that.

Looks like it's an old `prototype.js` version causing this, which makes it clear why I'm seeing this issue on multiple sites.

Is it fixable on your end to not rely directly on it (and being prone to websites overriding it)?

PR #32508, opened 2024-09-08
chore: improve error message when Array.toJSON misbehaves

### LINKED ISSUE #32507: [Bug]: Recorder doesn't work when `Array.prototype.toJSON` is declared
### Version

1.47.0

### Steps to reproduce

I was trying recorder on a site that defines `Array.prototype.toJSON` (it uses prototype.js 1.7), which messes up `JSON.stringify`.

To replicate it, run the following command (I ran it in powershell, so maybe it needs to be adjusted):

```bash
npx playwright codegen 'data:text/html,<script>Array.prototype.toJSON = () => `[]`;setTimeout(() => delete Array.prototype.toJSON, 5000)</script>'
```

Recorder only starts working when `Array.prototype.toJSON` gets deleted.

**Edit:** It also impacts `exposeFunction`:

```ts
test('should work with busted Array.prototype.toJSON', async ({ page }) => {
  await page.setContent(`<script>
    Array.prototype.toJSON = () => '';
  </script>`);
  await page.exposeFunction('add', (a, b) => a + b);
  expect(await page.evaluate('add(5, 6)')).toBe(11);
});
```

### Expected behavior

It should record the page.

### Actual behavior

It doesn't record the page.

### Additional context

This is a very similar to #22528. 

### Environment

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 13th Gen Intel(R) Core(TM) i5-1335U
    Memory: 10.19 GB / 31.66 GB
  Binaries:
    Node: 20.17.0 - ~\scoop\apps\nvm\current\nodejs\nodejs\node.EXE
    npm: 10.8.2 - ~\scoop\apps\nvm\current\nodejs\nodejs\npm.CMD
  IDEs:
    VSCode: 1.93.0 - C:\Users\rui.figueira\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 5.2.26 - C:\Users\rui.figueira\scoop\shims\bash.EXE
  npmPackages:
    @playwright/test: ^1.47.0 => 1.47.0
```

ruifigueira:
Fixes: #32507

### DIFF
--- a/packages/playwright-core/src/server/page.ts
+++ b/packages/playwright-core/src/server/page.ts
@@ -846,6 +846,8 @@ export class PageBinding {
         const handle = await context.evaluateHandle(takeHandle, { name, seq }).catch(e => null);
         result = await binding.playwrightFunction({ frame: context.frame, page, context: page._browserContext }, handle);
       } else {
+        if (!Array.isArray(serializedArgs))
+          throw new Error(`serializedArgs is not an array. This can happen when Array.prototype.toJSON is defined incorrectly`);
         const args = serializedArgs!.map(a => parseEvaluationResultValue(a));
         result = await binding.playwrightFunction({ frame: context.frame, page, context: page._browserContext }, ...args);
       }
--- a/tests/page/page-expose-function.spec.ts
+++ b/tests/page/page-expose-function.spec.ts
@@ -300,3 +300,12 @@ it('should work with busted Array.prototype.map/push', async ({ page, server }) 
   await page.exposeFunction('add', (a, b) => a + b);
   expect(await page.evaluate('add(5, 6)')).toBe(11);
 });
+
+it('should fail with busted Array.prototype.toJSON', async ({ page }) => {
+  await page.evaluateHandle(() => (Array.prototype as any).toJSON = () => '"[]"');
+
+  await page.exposeFunction('add', (a, b) => a + b);
+  await expect(() => page.evaluate(`add(5, 6)`)).rejects.toThrowError('serializedArgs is not an array. This can happen when Array.prototype.toJSON is defined incorrectly');
+
+  expect.soft(await page.evaluate(() => ([] as any).toJSON())).toBe('"[]"');
+});


packages/playwright-core/src/server/page.ts
```diff
@@ -903,7 +903,21 @@ function addPageBinding(playwrightBinding: string, bindingName: string, needsHan
       }
       payload = { name: bindingName, seq, serializedArgs };
     }
+
+    // if any of these types defines toJSON, it can mess up JSON.stringify
+    const toJSONs: [Object | Array<any> | Number | Boolean | String, any][] = [];
+    for (const ctor of [Object, Array, Number, Boolean, String]) {
+      if ('toJSON' in ctor.prototype) {
+        toJSONs.push([ctor, (ctor.prototype as any).toJSON]);
+        delete ctor.prototype.toJSON;
```
dgozman, line 912: I don't think we want to mess up with userland overrides. Can we have any other workaround instead?

ruifigueira: I restore them after `JSON.stringify` (and my test already checks it) but I suppose there are some corner cases that I'm not thinking of. One option would be to only rely on `JSON.stringify` if none of those types defines `toJSON`, otherwise implement it ourselves.

ruifigueira: There's probably a simple solution: ensure all properties of `SerializableValue` override `toJSON` as `null` if it's prototype defines it. I'll test it today.

dgozman: That's an interesting idea. Did you have a chance to try it out?

ruifigueira: I just force pushed a new patch. I implemented a custom "JSON.stringify" instead because it adding `toJSON` property to objects could lead to side effects.

It is working fine with `Array.prototype.toJSON`.

Primitive types serialization like number or string are not impacted with a custom `toJSON` implementation so I left them out.

I also left out `Object.prototype.toJSON`. If users mess up this function, it will make playwright fail, but I don't think there's much we can do, because it's a fallback mechanism for custom object serializattion:

https://github.com/microsoft/playwright/blob/8995ace8253ad25865a04452408967e929911205/packages/playwright-core/src/server/isomorphic/utilityScriptSerializers.ts#L214-L224

dgozman: I looked at the last commit, and I don't think it's worth it. Manually rolling `JSON.stringify` is non-trivial. For example, your current implementation ignores `toJSON()` for objects and does not handle infinite recursion. I am sure there are more edge cases that built-in implementation covers.

I think it's ok if `exposeBinding` does not work on some sites that mess up with builtins. To be testable, these will have to be updated. Alternatively, one can remove custom `toJSON()` override from `addInitScript()`, as long as the site keeps working.

ruifigueira: Just a note: my implementation is not intended to be generic, it doesn't handle recursion simply because the structure is not recursive (it's an object with `SerializableValue` objects). But I see your point

dgozman: For the record, we had a similar attempt (in spirit) to make codegen work on sites that mess up with builtins - see #6187. However, we have decided that's too much complexity for too little benefit, and it worked so far. This case seems similar.

ruifigueira: I agree, sometimes it's not worth it. I would propose a change however: this problem manifests itself as a `serializableArgs.map is not a function`, just like in #22460, from 

https://github.com/microsoft/playwright/blob/8995ace8253ad25865a04452408967e929911205/packages/playwright-core/src/server/page.ts#L849

If `serializedArgs` is not an array, it should throw a better error to the end user, so that they know Array.prototype.toJSOn is probably messed up.

dgozman: That sounds good to me, looking forward to the PR.

ruifigueira: I just pushed a new change, with a more explicit error message.

dgozman (approved):
Looks great, thank you for the PR!

PR #32512, opened 2024-09-09
fix(tracing): use page swap timestamp to find closest screenshot

Skn0tt:
Follow-up to https://github.com/microsoft/playwright/pull/32248. When we have it, we should use the page swap timestamp we get from Chromium to find the closest screenshot.

### DIFF
--- a/packages/trace-viewer/src/traceModernizer.ts
+++ b/packages/trace-viewer/src/traceModernizer.ts
@@ -395,10 +395,6 @@ export class TraceModernizer {
         result.push({ ...event, monotonicTime: 0, origin: 'library' });
         continue;
       }
-      if (event.type === 'frame-snapshot') {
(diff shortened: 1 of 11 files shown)


Skn0tt (commented):
addressed!

packages/trace-viewer/src/traceModernizer.ts
```diff
@@ -395,6 +395,10 @@ export class TraceModernizer {
         result.push({ ...event, monotonicTime: 0, origin: 'library' });
         continue;
       }
+      if (event.type === 'frame-snapshot') {
+        result.push({ ...event, snapshot: { ...event.snapshot, wallTime: 0 } });
```
dgozman, line 399: How about we make `wallTime` optional in the type definition, since there is no way we can have it for old traces anyway?

dgozman (approved):

PR #32514, opened 2024-09-09
feat(trace-viewer): show screenshot pointer

Skn0tt:
Follow-up to https://github.com/microsoft/playwright/pull/32248. Adds a glowing red circle that shows the click position. I made it glowing to show that its position is inaccurate.
Screenshot 2024-09-09 at 11 33 45

### DIFF
--- a/packages/trace-viewer/src/ui/clickPointer.tsx
+++ b/packages/trace-viewer/src/ui/clickPointer.tsx
@@ -1,0 +1,48 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import 'react';
+
+declare module 'react' {
+  namespace JSX {
+    interface IntrinsicElements {
+      'x-pw-pointer': JSX.IntrinsicElements['div'];
+    }
+  }
+}
+
+export function ClickPointer({ point }: { point: { x: number; y: number } }) {
+  return (
+    <x-pw-pointer
+      style={{
+        position: 'fixed',
+        backgroundColor: '#f44336',
+        boxShadow: '0px 0px 60px 20px #f44336',
+        width: '20px',
+        height: '20px',
+        borderRadius: '10px',
+        margin: '-10px 0 0 -10px',
+        zIndex: 2147483646,
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        left: `${point.x}px`,
+        top: `${point.y}px`,
+      }}
+      title='Click positions on screenshots are inaccurate'
+    />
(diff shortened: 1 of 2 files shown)


packages/trace-viewer/src/ui/clickPointer.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function ClickPointer({ point }: { point: { x: number; y: number } }) {
+  return (
+    <x-pw-pointer
+      style={{
+        position: 'fixed',
+        backgroundColor: '#f44336',
+        boxShadow: '0px 0px 60px 20px #f44336',
+        width: '20px',
```
dgozman, line 34: Do you want to scale the pointer with the screenshot? How does it look when screenshot is tiny?

packages/trace-viewer/src/ui/clickPointer.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+export function ClickPointer({ point }: { point: { x: number; y: number } }) {
+  return (
+    <x-pw-pointer
+      style={{
+        position: 'fixed',
+        backgroundColor: '#f44336',
+        boxShadow: '0px 0px 60px 20px #f44336',
```
dgozman, line 33: We use a "warning symbol" already to indicate that position is not 100% correct: https://github.com/microsoft/playwright/blob/main/packages/trace-viewer/src/snapshotRenderer.ts#L372-L382.

packages/trace-viewer/src/ui/clickPointer.tsx
```diff
@@ -0,0 +1,48 @@
...
+  namespace JSX {
+    interface IntrinsicElements {
+      'x-pw-pointer': JSX.IntrinsicElements['div'];
+    }
+  }
+}
+
+export function ClickPointer({ point }: { point: { x: number; y: number } }) {
+  return (
+    <x-pw-pointer
```
dgozman, line 29: Why not a regular div instead? This is a part of UI mode dom, so can be a regular element.

packages/trace-viewer/src/ui/clickPointer.tsx
Skn0tt: It already scales with the screenshot:
https://github.com/user-attachments/assets/8a42082f-816f-47db-aa42-9b8d46832d73

packages/trace-viewer/src/ui/clickPointer.tsx
Skn0tt: I saw that, but thought that a shadow communicates the "bluriness" of the pointer better. Now coming back to this after lunch, I think you're right and we should just reuse the pattern we already have :D Gonna change that.

packages/trace-viewer/src/ui/clickPointer.tsx
Skn0tt: Happy to change it to a `div`. Could you elaborate on why we use the custom element in the iframe? I imagined that we'd have special treatment for these components, but I cannot find that in our code.

dgozman: We use custom tags to avoid any user css applying to them. This is useful when these elements are embedded into the page itself, for example inside the snapshot iframe or in the codegen overlay.

packages/trace-viewer/src/ui/clickPointer.tsx
Skn0tt: done in https://github.com/microsoft/playwright/pull/32514/commits/f0d12bfaa53469ce429524f143a4acd3e021fbda, here's how it looks now: Screenshot 2024-09-09 at 13 29 14

dgozman (approved):

Issue #32488, opened 2024-09-06
Let's follow up with:


Skn0tt:
Let's follow up with:
- showing the click point;
- precise screencast timestamp from the browser;
- checking how this works with `snapshots: false` and `screenshots: false` traces.

_Originally posted by @dgozman in https://github.com/microsoft/playwright/pull/32248#pullrequestreview-2251054532_

PR #32515, opened 2024-09-09
feat(tracing): show screenshot behind iframe

Skn0tt:
@ruifigueira had a really nice idea in https://github.com/microsoft/playwright/issues/23964#issuecomment-2334662663: If we show the snapshot on top of the screenshot but with `opacity: 0`, then the locators will continue to work. This PR implements that, and I'm liking it! See the attached video for a demo.

https://github.com/user-attachments/assets/0e57df64-a1ea-43c9-a51d-c15bfd8489cc

Part of https://github.com/microsoft/playwright/issues/32488.

### DIFF
--- a/packages/trace-viewer/src/ui/inspectorTab.tsx
+++ b/packages/trace-viewer/src/ui/inspectorTab.tsx
@@ -17,7 +17,7 @@
 import { CodeMirrorWrapper } from '@web/components/codeMirrorWrapper';
 import type { Language } from '@web/components/codeMirrorWrapper';
 import { ToolbarButton } from '@web/components/toolbarButton';
-import { copy, useSetting } from '@web/uiUtils';
+import { copy } from '@web/uiUtils';
 import * as React from 'react';
 import './sourceTab.css';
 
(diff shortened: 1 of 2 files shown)


Skn0tt:
@dgozman i'm curious what you think about this approach. Do you think it makes sense?

dgozman:
I am not sure this would be useful and not confusing. Presumably, I would use "screenshot instead of snapshot" when there is no snapshot to begin with, or the snapshot renders incorrectly. In both of these cases, pick locator will end up picking the wrong element, because user assumes the element visible on the screen, while in the snapshot it is at a different place. This is also assuming the timing and scroll position of snapshot and screenshot perfectly align, which I'm not sure will be the case most of the time.

Skn0tt:
Yeah that makes sense. I think in a case where the invisible elements only make up part of the screen, e.g. a WebGL-based diagram, this implementation keeps locators working at least partially. You couldn't use locators for the diagram, but maybe for the inputs that control it.

The fact that timing + scroll positions don't necessarily align is the kicker though, agree. I'll go ahead and close!

PR #32518, opened 2024-09-09
chore: merge fetch params on server side

mxschmitt:
https://github.com/microsoft/playwright-python/pull/2546#discussion_r1750090592


packages/playwright-core/src/server/fetch.ts
```diff
@@ -159,6 +159,11 @@ export abstract class APIRequestContext extends SdkObject {
       for (const { name, value } of params.params)
         requestUrl.searchParams.append(name, value);
     }
+    if (params.encodedParams) {
```
dgozman, line 162: Perhaps just `requestUrl.search = params.encodedParams`?

dgozman (approved):

PR #32531, opened 2024-09-10
fix(ct): throw error if inline component is getting mounted

### LINKED ISSUE #32167: [Bug]: Playwright Components isn't working with inline components
### Version

1.46.0

### Steps to reproduce

Reproduction available here:
https://github.com/mrmckeb/tmp-pw-ct

1. Install (see README)
2. Run `pnpm run test-ct`

### Expected behavior

Should render a React component.

### Actual behavior

Fails to render React components unless declared within the `mount` function.

```tsx
import React from "react";
import { test, expect } from "@playwright/experimental-ct-react";

test.use({ viewport: { width: 500, height: 500 } });

test("does work", async ({ mount }) => {
  const component = await mount(<h1>Learn React</h1>);
  await expect(component).toContainText("Learn React");
});

function App() {
  return <h1>Learn React</h1>;
}

test("doesn't work", async ({ mount }) => {
  const component = await mount(<App />);
  await expect(component).toContainText("Learn React");
});
```

### Additional context

We've tested this on 1.45, 1.46 with React 18.3.1 and React 19 RC.

I've confirmed this in a real project as well, and on another system (with another engineer).

### Environment

```shell
System:
    OS: macOS 14.6.1
    CPU: (8) arm64 Apple M3
    Memory: 183.86 MB / 24.00 GB
  Binaries:
    Node: 18.14.0 - ~/.local/state/fnm_multishells/27565_1723703695253/bin/node
    npm: 9.3.1 - ~/.local/state/fnm_multishells/27565_1723703695253/bin/npm
    pnpm: 9.7.1 - ~/.local/state/fnm_multishells/27565_1723703695253/bin/pnpm
  IDEs:
    VSCode: 1.92.1 - /usr/local/bin/code
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/experimental-ct-react: 1.46.0 => 1.46.0
```

mxschmitt:
What was happening?
- When we use CT, we go over the test files, look at the imports using `tsxTransform.ts` and store them inside a map, these we feed into the import registry which we build using Vite and have access inside the browser
- In case of an inline component in the same file as where the test file is, this is not happening.
- jsx-runtime via babel kicks in, transforms every JSX component in something like that:

```
{
  __pw_type: 'jsx',
  type: [Function: MyInlineComponent],
  props: { value: 'Max' },
  key: undefined
}
```

this then gets passed into `wrapObject` which maps any function from the Node.js side into expose function calls so they work inside the browser. The assumption for `wrapObject` was to do it mostly for callbacks. So it does for `type` - which is actually our component. We then pass this to the React render function, which calls back the exposed function but we never return anything, so it mounts `undefined`.

---

While there have been experiments from certain vendors to get the 'client only' code inside a server side file, we should throw for now to not confuse users. We might revisit this in the future since Babel / TSX doesn't support it outside of the box.

Fixes https://github.com/microsoft/playwright/issues/32167

### DIFF
--- a/packages/playwright-ct-core/src/injected/serializers.ts
+++ b/packages/playwright-ct-core/src/injected/serializers.ts
@@ -66,6 +66,12 @@ export function transformObject(value: any, mapping: (v: any) => { result: any }
       result.push(transformObject(item, mapping));
     return result;
   }
+  if (value?.__pw_type === 'jsx' && typeof value.type === 'function') {
+    throw new Error([
+      'JSX component was not able to get resolved.',
+      'Make sure to define components outside of your test file.'
+    ].join('\n'));
+  }
   const result2: any = {};
   for (const [key, prop] of Object.entries(value))
     result2[key] = transformObject(prop, mapping);
--- a/packages/playwright-ct-core/src/injected/serializers.ts
+++ b/packages/playwright-ct-core/src/injected/serializers.ts
@@ -68,8 +68,9 @@ export function transformObject(value: any, mapping: (v: any) => { result: any }
   }
   if (value?.__pw_type === 'jsx' && typeof value.type === 'function') {
     throw new Error([
-      'JSX component was not able to get resolved.',
-      'Make sure to define components outside of your test file.'
+      `Component "${value.type.name}" cannot be mounted.`,
+      `Most likely, this component is defined in the test file. Create a test story instead.`,
+      `For more information, see https://playwright.dev/docs/test-components#test-stories.`,
     ].join('\n'));
   }
   const result2: any = {};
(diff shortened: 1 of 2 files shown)


packages/playwright-ct-core/src/injected/serializers.ts
```diff
@@ -66,6 +66,12 @@ export function transformObject(value: any, mapping: (v: any) => { result: any }
       result.push(transformObject(item, mapping));
     return result;
   }
+  if (value?.__pw_type === 'jsx' && typeof value.type === 'function') {
+    throw new Error([
```
dgozman, line 70:
```
Component "${value.type.name}" cannot be mounted.

Most likely, this component is defined in the test file. Create a test story instead.
For more information, see https://playwright.dev/docs/test-components#test-stories.
```

dgozman (approved):

PR #32533, opened 2024-09-10
fix(test-runner): do not consider retries for maxFailures

### LINKED ISSUE #26350: [Feature] Flag --max-failures respects --retries
This is a followup from https://github.com/microsoft/playwright/issues/24576.

In CI, I would like to stop after a single test has been retried until a full failure. Right now, this is not possible, as `--max-failures` counts tests that have been retried successfully.

This is important for velocity and cost-saving measures.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/26350

### DIFF
--- a/packages/playwright/src/runner/failureTracker.ts
+++ b/packages/playwright/src/runner/failureTracker.ts
@@ -31,8 +31,12 @@ export class FailureTracker {
   }
 
   onTestEnd(test: TestCase, result: TestResult) {
-    if (result.status !== 'skipped' && result.status !== test.expectedStatus)
+    if (result.status === 'skipped')
+      return;
+    if (result.status !== test.expectedStatus)
       ++this._failureCount;
+    else
+      this._failureCount -= result.retry;
   }
 
   onWorkerError() {
--- a/packages/playwright/src/runner/failureTracker.ts
+++ b/packages/playwright/src/runner/failureTracker.ts
@@ -31,12 +31,9 @@ export class FailureTracker {
   }
 
   onTestEnd(test: TestCase, result: TestResult) {
-    if (result.status === 'skipped')
-      return;
-    if (result.status !== test.expectedStatus)
+    // Test is considered failing after the last retry.
+    if (test.outcome() === 'unexpected' && test.results.length > test.retries)
       ++this._failureCount;
-    else
-      this._failureCount -= result.retry;
   }
 
   onWorkerError() {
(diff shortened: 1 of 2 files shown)


tests/playwright-test/max-failures.spec.ts
```diff
@@ -181,3 +181,40 @@ test('max-failures should work across phases', async ({ runInlineTest }) => {
...
+        maxFailures: 10,
+        retries: 10,
+      };
+    `,
+    'example.spec.ts': `
+      import fs from 'fs';
+      import { test, expect } from '@playwright/test';
+
+      test('I fail 9 times 1', () => {
+        let count = parseInt(fs.readFileSync('count1', 'utf8'), 10);
```
dgozman, line 198: You can always use `test.info().retry` instead of counting manually.

packages/playwright/src/runner/failureTracker.ts
```diff
@@ -31,8 +31,12 @@ export class FailureTracker {
   }
 
   onTestEnd(test: TestCase, result: TestResult) {
-    if (result.status !== 'skipped' && result.status !== test.expectedStatus)
+    if (result.status === 'skipped')
+      return;
+    if (result.status !== test.expectedStatus)
       ++this._failureCount;
+    else
+      this._failureCount -= result.retry;
```
dgozman, line 39: I find it hard to understand this logic. I'd think something like this, similar to `willRetry()` from `base.ts`:

```ts
// Test is considered failing after the last retry.
if (test.outcome() === 'unexpected' && test.results.length > test.retries)
  ++this._failureCount;
```

tests/playwright-test/max-failures.spec.ts
```diff
@@ -76,8 +76,8 @@ test('max-failures should work with retries', async ({ runInlineTest }) => {
     `,
   }, { 'max-failures': 2, 'retries': 4 });
   expect(result.exitCode).toBe(1);
-  expect(result.failed).toBe(1);
-  expect(result.output.split('\n').filter(l => l.includes('Received:')).length).toBe(2);
+  expect(result.failed).toBe(2);
+  expect(result.output.split('\n').filter(l => l.includes('Received:')).length).toBe(2 * (4 + 1)); // 2 tests * (4 retries + 1 original)
```
dgozman, line 80: I think this test will be 100% flaky now! I'd recommend `workers=1`, 3 tests in total, making the first test pass on the third retry, and the second test fail all retries.

dgozman (approved):

PR #32534, opened 2024-09-10
chore: remove unused api.ts

mxschmitt:
I think it was from the times when we had our custom coverage script.


packages/playwright-core/src/client/harRouter.ts
```diff
@@ -98,7 +98,7 @@ export class HarRouter {
   }
 
   async [Symbol.asyncDispose]() {
-    await this.dispose();
+    this.dispose();
```
mxschmitt, line 101: drive-by: awaiting a non-async method.

tests/library/client-certificates.spec.ts
```diff
@@ -76,15 +76,6 @@ const test = base.extend<TestOptions>({
   },
 });
 
-test.use({
```
mxschmitt, line 79: follow-up for https://github.com/microsoft/playwright/pull/31724.

PR #32551, opened 2024-09-11
feat(api): add method OPTIONS to request

erkde:
I have some API specs for validating CORS config - I couldn't find an `options` function on `request`, so wondering if that might be useful - just so we can do:

```
test('should allow * for origin', async ({ request }) => {
  const response = await request.options('https://api.github.com');
  expect(response.ok()).toBe(true);

  const headers = response.headers();
  expect(headers['access-control-allow-origin']).toEqual('*');
});
```

instead of:

```
const response = await request.fetch('https://api.github.com', { method: 'OPTIONS' });
```

### DIFF
--- a/docs/src/api/class-apirequestcontext.md
+++ b/docs/src/api/class-apirequestcontext.md
@@ -522,6 +522,65 @@ context cookies from the response. The method will automatically follow redirect
 ### option: APIRequestContext.head.maxRetries = %%-js-python-csharp-fetch-option-maxretries-%%
 * since: v1.46
 
+## async method: APIRequestContext.options
+* since: v1.48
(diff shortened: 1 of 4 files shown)


erkde:

erkde:
oh, 15 failed, 0 passed..  😂

dgozman:
@erkde Could you please file an issue by filling in a "Feature Request" template? We require an issue for any API changes.

erkde:
> @erkde Could you please file an issue by filling in a "Feature Request" template? We require an issue for any API changes.

Yeah, sorry I didn't know the process, I've just raised one here: https://github.com/microsoft/playwright/issues/32556

PR #32565, opened 2024-09-11
chore(tooling): install knip and remove unused files / dependencies

jeanpierrecarvalho:
chore(tooling): install knip and remove unused dependencies

**Changes Made:**

- Installed knip to analyze unused files and dependencies
- Removed unused files and dependencies identified by knip
- Add some folders to be ignored by knip for now, should be changed step by step

**Details of Changes:**

- Added knip to the project as a dev dependency to improve project cleanliness and remove dead code
- Ran an analysis using knip and removed files that were no longer in use or required by the project

**Additional Details:**

- These changes help streamline the project and reduce clutter, improving overall maintainability
- No functional changes were introduced, the focus was solely on project hygiene and code cleanup

**What is next?**

- remove folders and files in knip config file and solve the issues AND
- pnpm knip can run in the pre-commit hook OR / AND
- can be added to GHA on PR creation OR / AND
- can be added to release process and tag creation

### DIFF
--- a/knip.js
+++ b/knip.js
@@ -1,10 +1,42 @@
+const packagesToBeIgnored = [
+  "@types/codemirror",
+  "@types/formidable",
+  "@types/immutable",
+  "@types/ws",
+  "@types/xml2js",
+  "@typescript-eslint/utils",
+  "@vitejs/plugin-basic-ssl",
+  "ansi-styles",
+  "colors",
+  "dotenv",
+  "esbuild",
+  "eslint-plugin-internal-playwright",
+  "formidable",
+  "immutable",
+  "license-checker",
+  "mime",
+  "ssim.js",
+  "ws",
+  "xml2js",
+  "@playwright/test",
+];
+
+const binariesToBeIgnored = ["playwright"];
+
 const config = {
   workspaces: {
     ".": {
       //   entry: [""],
-      ignore: [],
-      ignoreDependencies: [],
-      ignoreBinaries: [],
+      ignore: [
+        "packages/**/*",
+        "utils/**/*",
+        ".github/**/*",
+        "tests/**/*",
+        "examples/**/*",
+        "browser_patches/**/*",
+      ],
+      ignoreDependencies: packagesToBeIgnored,
+      ignoreBinaries: binariesToBeIgnored,
     },
   },
 };
--- a/knip.js
+++ b/knip.js
@@ -1,0 +1,12 @@
+const config = {
+  workspaces: {
+    ".": {
+      //   entry: [""],
+      ignore: [],
+      ignoreDependencies: [],
+      ignoreBinaries: [],
+    },
+  },
+};
+
+export default config;
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,7 @@
         "eslint-plugin-react-hooks": "^4.6.2",
         "formidable": "^2.1.1",
         "immutable": "^4.3.7",
+        "knip": "^5.30.1",
         "license-checker": "^25.0.1",
         "mime": "^3.0.0",
         "node-stream-zip": "^1.15.0",
@@ -1708,6 +1709,24 @@
         "url": "https://github.com/sindresorhus/is?sponsor=1"
       }
     },
+    "node_modules/@snyk/github-codeowners": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@snyk/github-codeowners/-/github-codeowners-1.1.0.tgz",
+      "integrity": "sha512-lGFf08pbkEac0NYgVf4hdANpAgApRjNByLXB+WBip3qj1iendOIyAwP2GKkKbQMNVy2r1xxDf0ssfWscoiC+Vw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "commander": "^4.1.1",
+        "ignore": "^5.1.8",
+        "p-map": "^4.0.0"
+      },
+      "bin": {
+        "github-codeowners": "dist/cli.js"
+      },
+      "engines": {
+        "node": ">=8.10"
+      }
+    },
     "node_modules/@sveltejs/vite-plugin-svelte": {
       "version": "3.0.1",
       "resolved": "https://registry.npmjs.org/@sveltejs/vite-plugin-svelte/-/vite-plugin-svelte-3.0.1.tgz",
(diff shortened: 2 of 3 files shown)


dgozman (commented):
Thank you for the PR!

Unfortunately, I don't think we'd like to add another dependency, especially because the spirit of the change is to remove dependencies instead 😄 

We are happy to accept a PR that removes the dependencies we don't use. However, adding `knip` with its own config sounds like extra maintenance that we don't really need.

webpro:
Creator of [Knip](https://knip.dev) here. Btw what a coincidence, just yesterday I've added Playwright to Azure DevOps in my current gig. And this morning I'm seeing this. While I 100% appreciate the effort that went into this PR, this isn't how Knip should be configured for a monorepo and doesn't show much of its value. I'm sure we can change that, but wanted to know whether there's interest and I'll be happy to give it a good shot.

dgozman:
@webpro Thank you for chiming in! We are in general very careful about introducing new dependencies, so it's hard to tell whether adding `knip` would bring enough benefits for us. So far, we have not suffered much from unused files/dependencies, but perhaps we just don't know about them 😄 

I guess the above means that I cannot guarantee that we'll accept a PR that adds `knip`, but we would definitely take a look. It is up to you whether you'd like to spend time on this.

webpro:
Alright, let's go. This monorepo isn't the easiest one to analyze for Knip for two reasons:

- Entry paths diverge from the [default entry files](https://knip.dev/explanations/entry-files#default-entry-file-patterns). I've tried to find a few "strategic" entry points, but for someone less familiar with the project, this is a bit tedious.
- Using path aliases (`tsonfig.json#compilerOptions.paths`) to "cross workspace boundaries" doesn't play great with Knip's approach (https://knip.dev/reference/faq#why-cant-i-use-path-aliases-to-reference-other-workspaces).

However, I did explore a more "brute-force" approach. If we don't try to find all entry files, but instead add them all we won't be able to find unused files, but we can still get interesting reports regarding unused exports and unused + unlisted dependencies. Exports of entry files aren't reported by default, but there's `includeEntryExports: true`.

Here's the configuration I've used:

```json
{
  "$schema": "https://unpkg.com/knip@5/schema.json",
  "includeEntryExports": true,
  "workspaces": {
    ".": {
      "entry": ["**/*.{js,mjs,ts,tsx}"]
    },
    "packages/*": {
      "entry": ["**/*.{js,mjs,ts,tsx}"]
    },
    "tests/*/*": {
      "entry": ["**/*.{js,mjs,ts,tsx}"]
    },
    "packages/playwright/bundles/*": {
      "entry": ["**/*.{js,mjs,ts,tsx}"]
    },
    "packages/playwright-core/bundles/*": {
      "entry": ["**/*.{js,mjs,ts,tsx}"]
    }
  },
  "ignore": ["tests/expect/*.snapshots.js"]
}
```

Workspaces that aren't in `package.json#workspaces` were added manually.

So with all that in place, let's look at samples from the report:

```
$ knip --exports
Unused exports (266)
[..]
empty                                icons         unknown   packages/html-reporter/src/icons.tsx:106:14
externalLink                         icons         unknown   packages/html-reporter/src/icons.tsx:72:14
calendar                             icons         unknown   packages/html-reporter/src/icons.tsx:76:14
person                               icons         unknown   packages/html-reporter/src/icons.tsx:80:14
commit                               icons         unknown   packages/html-reporter/src/icons.tsx:84:14
default                                            unknown   packages/playwright-chromium/index.js:17:18
default                                            unknown   packages/playwright-chromium/index.mjs:18:8
colors                                             unknown   packages/playwright-core/bundles/utils/src/utilsBundleImpl.ts:18:14
debug                                              unknown   packages/playwright-core/bundles/utils/src/utilsBundleImpl.ts:21:14
```

This looks accurate to me, those exports don't seem to be referenced anywhere. Run Knip for the full list, pretty sure there's some dead code to be found here.

Let's look at the **unused** (dev) dependencies:

```
Unused dependencies (5)
https-proxy-agent                                   packages/playwright-core/bundles/utils/package.json
@babel/plugin-transform-private-property-in-object  packages/playwright/bundles/babel/package.json
core-js                                             tests/components/ct-vue-cli/package.json
core-js                                             tests/components/ct-vue2-cli/package.json
@playwright/test                                    tests/playwright-test/stable-test-runner/package.json
Unused devDependencies (8)
@babel/preset-react                package.json
@vitejs/plugin-basic-ssl           package.json
cross-env                          package.json
eslint-plugin-internal-playwright  package.json
mime                               package.json
@types/proper-lockfile             packages/playwright-core/bundles/utils/package.json
tslib                              tests/components/ct-svelte-vite/package.json
@vue/cli-plugin-router             tests/components/ct-vue-cli/package.json
```

There are a few false positives, but most of them seem indeed unused.

As for **unlisted** dependencies, here's a tiny sample of the report:

```
knip --dependencies
Unlisted dependencies (94)
[..]
@vue/compiler-dom                  packages/playwright-ct-vue/registerSource.mjs
vue                                packages/playwright-ct-vue/registerSource.mjs
playwright                         packages/playwright/bundles/babel/roll.js
@babel/template                    packages/playwright/bundles/babel/src/babelBundleImpl.ts
```

Results like this look valid to me as well. If a dependency is referenced, but not listed in either the workspace's or root `package.json`, it's reported as unlisted. Relying on transitive dependencies implicitly is generally not recommended.

Going to leave it at this as far as analysis and actually dealing with the reported issues goes, but feel free to ask away! Happy to assist and discuss. Note that Knip has a few options e.g. to slice and dice the report to focus on a specific issue type or a specific workspace or ignore certain binaries or dependencies, etc. etc. Docs are at https://knip.dev.

mxschmitt:
Thanks for looking into it - looks like the list contains mostly false positives (https://github.com/microsoft/playwright/pull/32650). Things like `import * as icons from './icons';` with `icons['commit']` are probably hard to catch.

I'll close it for now, if you have any concrete findings of dependencies or files which are unused, feel free to remove them individually. Thanks!

webpro:
> Thanks for looking into it - looks like the list contains mostly false positives (#32650). Things like `import * as icons from './icons';` with `icons['commit']` are probably hard to catch.

The PR you link to removes _exactly_ the "unused exports" as Knip reported. That's what Knip does: report unused clutter that can be removed. Not seeing a single false positive in this particular sample (there's a lot more). It was a quick config + run, there might be false positives, but they're not here.

> I'll close it for now, if you have any concrete findings of dependencies or files which are unused, feel free to remove them individually. Thanks!

Not sure how much more concrete this could be. Might be a misunderstanding, but also I'm not here to push Knip into your project. Feel free to use or ignore it as you please.

PR #32576, opened 2024-09-12
chore: perform double click while recording

pavelfeldman:


packages/playwright-core/src/server/injected/recorder/recorder.ts
```diff
@@ -315,6 +361,7 @@ class RecordActionTool implements RecorderTool {
   }
 
   onMouseMove(event: MouseEvent) {
+    this._commitPendingClickAction();
```
dgozman, line 364: I wonder whether we should allow movement while over the same click target? Users probably wiggle for 1-2 pixels.

dgozman (approved):

PR #32581, opened 2024-09-12
feat(tracing): add .pwtrace to trace file extension

### LINKED ISSUE #32226: [Feature]: Allow traces in test-results folder to have custom file extension e.g. `.pwtrace.zip`
### 🚀 Feature Request

Instead of using `.zip` use `.pwtrace.zip` as a file extension for trace files.

### Example

`trace.zip` -> `trace.pwtrace.zip`

or

`trace-1.zip` -> `trace-1.pwtrace.zip`

### Motivation

We like to identify trace files in our `test-results/` folder better. This is for a custom AzDO extension we are writing where we see a list files and would like to offer first-party Trace Viewer support.

HTML report doesn't work for us, since we use TRX file format in order to put the data back into AzDO. 

Note: This originates from the internal ODSP team.

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32226

I've updated every mention of `.trace.zip` except for the release notes.


packages/playwright/src/worker/testTracing.ts
```diff
@@ -214,7 +214,7 @@ export class TestTracing {
       });
     });
 
-    const tracePath = this._testInfo.outputPath('trace.zip');
+    const tracePath = this._testInfo.outputPath('trace.pwtrace.zip');
```
dgozman, line 217: There is also `generateNextTraceRecordingPath()`.

dgozman (approved):
I am surprised that not a lot of places in the source have been changed 😄  Let's run CQ1 tests just in case.

PR #32583, opened 2024-09-12
feat(test runner): don't run tests on --watch start

### LINKED ISSUE #32580: [Internal]: Watch mode should not run an upfront test run
Tests take long, there's a good chance that folks are only interested in changing test suites when they use watch mode. We shouldn't be doing a full run on start, but only a test list.

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32580.

### DIFF
--- a/packages/playwright/src/runner/testServer.ts
+++ b/packages/playwright/src/runner/testServer.ts
@@ -79,6 +79,7 @@ export class TestServerDispatcher implements TestServerInterface {
   private _serializer = require.resolve('./uiModeReporter');
   private _watchTestDirs = false;
   private _closeOnDisconnect = false;
+  private _populateDependenciesOnList = false;
 
   constructor(configLocation: ConfigLocation) {
     this._configLocation = configLocation;
@@ -113,6 +114,7 @@ export class TestServerDispatcher implements TestServerInterface {
     this._closeOnDisconnect = !!params.closeOnDisconnect;
     await this._setInterceptStdio(!!params.interceptStdio);
     this._watchTestDirs = !!params.watchTestDirs;
+    this._populateDependenciesOnList = !!params.populateDependenciesOnList;
   }
 
   async ping() {}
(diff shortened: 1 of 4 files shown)


packages/playwright/src/runner/testServer.ts
```diff
@@ -252,7 +252,7 @@ export class TestServerDispatcher implements TestServerInterface {
     config.cliListOnly = true;
 
     const status = await runTasks(new TestRun(config, reporter), [
-      createLoadTask('out-of-process', { failOnLoadErrors: false, filterOnly: false }),
+      createLoadTask('out-of-process', { failOnLoadErrors: false, filterOnly: false, populateDependencies: true }),
```
dgozman, line 255: We probably want an option for this, since most invocations of `listTests()` do not really need to populate dependencies, and it's costly to run vite build. I'd put an option into `initialize()`. Perhaps UI mode will opt into that at some point as well.

Skn0tt: We're only running the Vite build if there's a component testing plugin registered, in which case we need to run Vite to get a correct result. In the common case without component testing, `populateDependencies: true` is a no-op. Could you outline a scenario where this adds unnecessary cost?

Skn0tt: Wait, ignore me - listing tests does *not* need a vite build to return a correct result. Dependency population is what needs the vite build, but one could be interested in listing tests without caring about dependency population. Gonna implement your recommendation.

Skn0tt: done in https://github.com/microsoft/playwright/pull/32583/commits/7827ca77c547838385b0eca71cdf52e2dad514d1

dgozman (approved):

mxschmitt (commented):
To opt back into the old behaviour, you start watch mode and hit enter to run all of the tests?

Skn0tt:
> To opt back into the old behaviour, you start watch mode and hit enter to run all of the tests?

Correct! Hitting enter will always run all tests (minus the filtered ones).

PR #32584, opened 2024-09-12
feat(test runner): allow stopping testrun with escape

### LINKED ISSUE #32579: [Internal]: Watch mode should have a way of stopping tests
Tests can take long to execute. All of our interfaces have some way of stoping a test run, watch mode should too.

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32579


pavelfeldman:
What do you think about handling the first Ctrl+C (Cmd+C) to stop? It might be a bit more discoverable. Esc is designed to be used for escape sequences in the terminal environments, so there is an expectation that something follows Esc.

pavelfeldman:
> We're listening to both Ctrl+C, Escape and their Control sequence equivalents. Do you think we should be listening exclusively to Ctrl+C?

I misunderstood the intention behind the patch.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -251,12 +252,37 @@ export async function runWatchModeLoop(configLocation: ConfigLocation, initialOp
...
 
+function readKeyPress(handler: (text: string, key: any) => void): Disposable {
+  const rl = readline.createInterface({ input: process.stdin, escapeCodeTimeout: 50 });
+  readline.emitKeypressEvents(process.stdin, rl);
+  if (process.stdin.isTTY)
+    process.stdin.setRawMode(true);
+
+  process.stdin.on('keypress', handler);
+
+  return {
```
pavelfeldman, line 263: returning the callback effect-style would also work. We only use Disposable when we interoperate with vscode.

packages/playwright/src/runner/watchMode.ts
```diff
@@ -316,15 +338,9 @@ Change settings
...
-
-  process.stdin.on('keypress', handler);
-  void result.finally(() => {
-    process.stdin.off('keypress', handler);
-    rl.close();
-    if (process.stdin.isTTY)
-      process.stdin.setRawMode(false);
   });
+
+  void result.finally(reader.dispose);
```
pavelfeldman, line 343: I think readKeyPress should have this code, not readCommand

pavelfeldman (approved):

packages/playwright/src/runner/watchMode.ts
Skn0tt: i've moved `readKeyPress` to return a promise in https://github.com/microsoft/playwright/pull/32584/commits/f0de6ccd81ff07f9e947625839b3335e0da3ba4a. That changes quite a bit about the code, but makes it easier to reason about in general - and it also moves the disposal into `readKeyPress` 👍

packages/playwright/src/runner/watchMode.ts
Skn0tt: done in https://github.com/microsoft/playwright/pull/32584/commits/f0de6ccd81ff07f9e947625839b3335e0da3ba4a

Skn0tt:
Unsure why this is failing on Windows. Need to investigate

pavelfeldman (approved):

PR #32597, opened 2024-09-12
chore: move recorder trace to action collector

pavelfeldman:


packages/playwright-core/src/server/recorder/recorderCollection.ts
```diff
@@ -46,29 +53,55 @@ export class RecorderCollection extends EventEmitter {
...
+    if (error) {
+      // Do not clear current action on delayed error.
+      if (this._currentAction === actionInContext)
+        this._currentAction = null;
+    } else {
       this._currentAction = null;
+      this._actions.push(actionInContext);
+    }
+
+    this._lastAction = actionInContext;
```
dgozman, line 93: Do we want to set `_lastAction` in the error case?

dgozman (approved):

packages/playwright-core/src/server/recorder/recorderCollection.ts
pavelfeldman: Hm, I'll think about what this means. We definitely don't want to glue to the previous one, so maybe we want to reset it.

PR #32598, opened 2024-09-12
chore: recorder is trace viewer experiment

pavelfeldman:


packages/trace-viewer/src/ui/sourceTab.tsx
```diff
@@ -66,7 +66,9 @@ export const SourceTab: React.FunctionComponent<{
     highlight.push({ line: targetLine, type: 'running' });
 
     // After the source update, but before the test run, don't trust the cache.
-    if (source.content === undefined || shouldUseFallback) {
+    if (fallbackLocation?.source?.content !== undefined) {
+      source.content = fallbackLocation.source.content;
```
dgozman, line 70: I don't quite understand this one. Do we show the fallback source even when we have the real source content?

packages/playwright-core/src/server/recorder/recorderInTraceViewer.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+async function openApp(trace: string, options?: TraceViewerServerOptions & { headless?: boolean }) {
+  const server = await startTraceViewerServer(options);
+  await installRootRedirect(server, [trace], { ...options, webApp: 'recorder.html' });
+  const page = await openTraceViewerApp(server.urlPrefix('precise'), 'chromium', options);
+  page.on('close', () => gracefullyProcessExitDoNotHang(0));
+}
+
+class RecorderTransport implements Transport {
```
dgozman, line 82: Why do we need an empty transport?

dgozman (approved):

packages/trace-viewer/src/ui/sourceTab.tsx
pavelfeldman: Yes, fallbackLocation is a priority.

packages/playwright-core/src/server/recorder/recorderInTraceViewer.ts
pavelfeldman: That's our way of communicating to the front-end (trace viewer). Commands from recorder front-end will arrive over this transport.

PR #32605, opened 2024-09-13
chore: unhide merge-reports command

mxschmitt:
Context https://github.com/microsoft/playwright/pull/30962#issuecomment-2345789633
Reverts https://github.com/microsoft/playwright/pull/22509

cc @muhqu

### DIFF
--- a/packages/playwright/src/program.ts
+++ b/packages/playwright/src/program.ts
@@ -133,7 +133,7 @@ Examples:
 }
 
 function addMergeReportsCommand(program: Command) {
-  const command = program.command('merge-reports [dir]', { hidden: true });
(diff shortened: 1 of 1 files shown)


muhqu:
@mxschmitt I still wonder why it was explicitly marked hidden in the [first place](https://github.com/microsoft/playwright/pull/22509).

mxschmitt:
We released the feature in v1.37, which was released in Aug 11, 2023. We were hiding it in Apr 20, 2023, so probably it wasn't ready back then. We usually hide things if we don't consider them ready.

dgozman (approved):

PR #32606, opened 2024-09-13
chore: introduce option overrides on context/browser

mxschmitt:


packages/playwright-core/src/server/browserType.ts
```diff
@@ -92,23 +92,22 @@ export abstract class BrowserType extends SdkObject {
...
 
-  async launchPersistentContext(metadata: CallMetadata, userDataDir: string, options: channels.BrowserTypeLaunchPersistentContextOptions & { useWebSocket?: boolean }): Promise<BrowserContext> {
-    options = this._validateLaunchOptions(options);
+  async launchPersistentContext(metadata: CallMetadata, userDataDir: string, persistentContextOptions: channels.BrowserTypeLaunchPersistentContextOptions & { useWebSocket?: boolean }): Promise<BrowserContext> {
+    const launchOptions = this._validateLaunchOptions(persistentContextOptions);
     if (this._useBidi)
-      options.useWebSocket = true;
+      launchOptions.useWebSocket = true;
     const controller = new ProgressController(metadata, this);
-    const persistent: channels.BrowserNewContextParams = { ...options };
```
mxschmitt, line 100: this was introduced [here](https://github.com/microsoft/playwright/commit/9569cb5c1e5c01c7d4f3246875d778fe1421c84d#diff-6a35d60cc44656b79b6202dac37dda6541caed378a6a83358b9329e3ebf07491R80), probably fine to remove it now since we don't modify it anymore (client certificates), alternatively we can make a copy at the beginning of the function call.

packages/playwright-core/src/server/chromium/crPage.ts
```diff
@@ -543,7 +543,7 @@ class FrameSession {
       const options = this._crPage._browserContext._options;
       if (options.bypassCSP)
         promises.push(this._client.send('Page.setBypassCSP', { enabled: true }));
-      if (options.ignoreHTTPSErrors)
+      if (options.ignoreHTTPSErrors || options.ignoreHTTPSErrorsOverride)
```
dgozman, line 546: This line looks strange, because if `ignoreHTTPSErrorsOverride === false`, I'd assume we should not ignore the errors. I'd recommend to rename it to `internalIgnoreHTTPSErrors` to avoid this confusion.

tests/library/client-certificates.spec.ts
```diff
@@ -546,6 +546,7 @@ test.describe('browser', () => {
       keyPath: asset('client-certificates/client/trusted/key.pem'),
     };
     const page = await browser.newPage({
+      ignoreHTTPSErrors: true,
```
dgozman, line 549: Why this change? Seems unrelated.

packages/playwright-core/src/server/browserContext.ts
```diff
@@ -665,10 +665,7 @@ export async function createClientCertificatesProxyIfNeeded(options: channels.Br
   if ((options.proxy?.server && options.proxy?.server !== 'per-context') || (browserOptions?.proxy?.server && browserOptions?.proxy?.server !== 'http://per-context'))
     throw new Error('Cannot specify both proxy and clientCertificates');
   verifyClientCertificates(options.clientCertificates);
-  const clientCertificatesProxy = new ClientCertificatesProxy(options);
-  options.proxy = { server: await clientCertificatesProxy.listen() };
-  options.ignoreHTTPSErrors = true;
-  return clientCertificatesProxy;
+  return new ClientCertificatesProxy(options);
```
dgozman, line 668: - This method should probably take `types.BrowserContextOptions`. I also see 49 occurrences of `BrowserNewContext` in `src/server` that should probably all be replaced.
- Once we remove line 665, we can move the validation logic into `ClientCertificatesProxy` constructor and remove this method entirely.

tests/library/client-certificates.spec.ts
mxschmitt: Before we were overriding the `ignoreHTTPSErrors` field when creating the socks proxy. Later on we were reading it in the APIRequestContext when making a http request, there we were readying the updated value rather than the original value. This is e.g. what this PR is fixing, that things like this won't happen anymore.

packages/playwright-core/src/server/chromium/crPage.ts
mxschmitt: done.

dgozman (approved):

PR #32611, opened 2024-09-13
feat(client-certificates): add support for proxies

### LINKED ISSUE #32370: [Feature]: Allow TLS Client Certificates with proxy servers
### 🚀 Feature Request

In Playwright version 1.46, support for TLS Client Certificates was added. However, this feature does not work with Proxy Servers as per the documentation:

> Using Client Certificates in combination with Proxy Servers is not supported.
> [https://playwright.dev/java/docs/api/class-apirequest#api-request-new-context](url)

### Example

_No response_

### Motivation

We are using Playwright in our company. Our laptops are within the company network, which requires the use of proxies. We can't use the TLS Client Certificates locally, making this feature unusable for us. Our current workaround is to use RestAssured, since proxy servers are supported there.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/32370

### DIFF
--- a/packages/playwright-core/src/server/browserType.ts
+++ b/packages/playwright-core/src/server/browserType.ts
@@ -18,7 +18,7 @@ import fs from 'fs';
 import * as os from 'os';
 import path from 'path';
 import type { BrowserContext } from './browserContext';
-import { createClientCertificatesProxyIfNeeded, normalizeProxySettings, validateBrowserContextOptions } from './browserContext';
+import { normalizeProxySettings, validateBrowserContextOptions } from './browserContext';
 import type { BrowserName } from './registry';
 import { registry } from './registry';
 import type { ConnectionTransport } from './transport';
@@ -39,6 +39,7 @@ import { RecentLogsCollector } from '../utils/debugLogger';
 import type { CallMetadata } from './instrumentation';
 import { SdkObject } from './instrumentation';
 import { type ProtocolError, isProtocolError } from './protocolError';
+import { ClientCertificatesProxy } from './socksClientCertificatesInterceptor';
 
 export const kNoXServerRunningError = 'Looks like you launched a headed browser without having a XServer running.\n' +
   'Set either \'headless: true\' or use \'xvfb-run <your-playwright-app>\' before running Playwright.\n\n<3 Playwright Team';
@@ -92,7 +93,7 @@ export abstract class BrowserType extends SdkObject {
     return browser;
   }
 
-  async launchPersistentContext(metadata: CallMetadata, userDataDir: string, persistentContextOptions: channels.BrowserTypeLaunchPersistentContextOptions & { useWebSocket?: boolean }): Promise<BrowserContext> {
+  async launchPersistentContext(metadata: CallMetadata, userDataDir: string, persistentContextOptions: channels.BrowserTypeLaunchPersistentContextOptions & { useWebSocket?: boolean, internalIgnoreHTTPSErrors?: boolean }): Promise<BrowserContext> {
     const launchOptions = this._validateLaunchOptions(persistentContextOptions);
     if (this._useBidi)
       launchOptions.useWebSocket = true;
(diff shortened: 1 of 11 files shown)


packages/playwright-core/src/server/browserType.ts
```diff
@@ -92,17 +93,20 @@ export abstract class BrowserType extends SdkObject {
     return browser;
   }
 
-  async launchPersistentContext(metadata: CallMetadata, userDataDir: string, persistentContextOptions: channels.BrowserTypeLaunchPersistentContextOptions & { useWebSocket?: boolean }): Promise<BrowserContext> {
+  async launchPersistentContext(metadata: CallMetadata, userDataDir: string, persistentContextOptions: channels.BrowserTypeLaunchPersistentContextOptions & { useWebSocket?: boolean, internalIgnoreHTTPSErrors?: boolean }): Promise<BrowserContext> {
```
dgozman, line 96: I think the `persistentContextOptions` name is confusing here. It is both context and launch options, so I'd rather call it `options`.

packages/playwright-core/src/server/browserType.ts
```diff
@@ -92,17 +93,20 @@ export abstract class BrowserType extends SdkObject {
...
     controller.setLogName('browser');
     const browser = await controller.run(async progress => {
       // Note: Any initial TLS requests will fail since we rely on the Page/Frames initialize which sets ignoreHTTPSErrors.
-      const clientCertificatesProxy = await createClientCertificatesProxyIfNeeded(persistentContextOptions);
-      if (clientCertificatesProxy)
+      let clientCertificatesProxy: ClientCertificatesProxy | undefined;
+      if (persistentContextOptions.clientCertificates?.length) {
+        clientCertificatesProxy = new ClientCertificatesProxy(persistentContextOptions);
         launchOptions.proxyOverride = await clientCertificatesProxy?.listen();
+        persistentContextOptions.internalIgnoreHTTPSErrors = true;
```
dgozman, line 108: Since we do actually mutate options here, let's make a separate copy.

packages/playwright-core/src/server/fetch.ts
```diff
@@ -172,20 +172,9 @@ export abstract class APIRequestContext extends SdkObject {
     // When `clientCertificates` is present, we set the `proxy` property to our own socks proxy
```
dgozman, line 172: This comment does not apply anymore.

packages/playwright-core/src/server/fetch.ts
```diff
@@ -172,20 +172,9 @@ export abstract class APIRequestContext extends SdkObject {
...
-          port: proxyOpts.port || undefined,
-        });
-      } else {
-        if (proxy.username)
-          proxyOpts.auth = `${proxy.username}:${proxy.password || ''}`;
-        // TODO: We should use HttpProxyAgent conditional on proxyOpts.protocol instead of always using CONNECT method.
-        agent = new HttpsProxyAgent(proxyOpts);
-      }
-    }
+    if (proxy && proxy.server !== 'per-context' && !shouldBypassProxy(requestUrl, proxy.bypass))
```
dgozman, line 175: I'd rather comment that `per-context` is here for historic reasons and backwards compatibility.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -94,7 +97,12 @@ class SocksProxyConnection {
   }
 
   async connect() {
-    this.target = await createSocket(rewriteToLocalhostIfNeeded(this.host), this.port);
+    if (this.socksProxy.agent)
+      // @ts-expect-error
```
dgozman, line 101: Perhaps `as any` somewhere instead?

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -233,12 +241,15 @@ export class ClientCertificatesProxy {
   ignoreHTTPSErrors: boolean | undefined;
   secureContextMap: Map<string, tls.SecureContext> = new Map();
   alpnCache: ALPNCache;
+  agent: ReturnType<typeof createProxyAgent> | undefined;
```
dgozman, line 244:
```suggestion
  proxyAgentFromOptions: ReturnType<typeof createProxyAgent> | undefined;
```

dgozman (approved):

Issue #19621, opened 2022-12-21
[Question] Is there a way to return an api request response time?


BekaOkromchedlidze:
I am using the request fixture for API testing and I am just wondering is there a way to find what the api response time was?

pavelfeldman:
I don't think we have a feature like this. But request API is not done through the browser, so we don't have access to the detailed timing anyways.

klepik1990:
Did you see this info? Maybe it can helps you
https://playwright.dev/docs/api/class-request#request-timing

BekaOkromchedlidze:
@pavelfeldman I think this would be a very nice feature to have for api testing. It would be used to test performance sanity to make sure we get a response in an acceptable time.

@klepik1990 this is for the page fixture. We don't have this option for the request fixture

brunojbarros:
It's not completely accurate, but you can do something like:

```
const requestDate = new Date();

// your request: await request ...

const responseDate = new Date();

console.log(responseDate - requestDate);
```

I agree that this feature will be very useful 😄

srieas:
+1

fayasfb:
i agree this feature should be implemented in request fixture , as we need to do the performance of the api

brunojbarros:
Another option is to use testStep.duration: https://playwright.dev/docs/api/class-teststep#test-step-duration

shashwatahalder01:
Feature ought to be integrated within the request fixture.

Skn0tt:
Hey folks! Could you share a little more about your usecases? Specifically I'm curious about the granularity of timing you need. Do you care about the full roundtrip time, or also about time to first byte or time taken by the TLS handshake?

brunojbarros:
> Hey folks! Could you share a little more about your usecases? Specifically I'm curious about the granularity of timing you need. Do you care about the full roundtrip time, or also about time to first byte or time taken by a potential TLS handshake?

It would be ideal to have something similar to what k6 has: [https://grafana.com/docs/k6/latest/examples/get-timings-for-an-http-metric/](https://grafana.com/docs/k6/latest/examples/get-timings-for-an-http-metric/)

At least just the duration.

Skn0tt:
@brunojbarros thanks for getting back to me. I'm curious, are you performing load testing? Could you share a little bit more about your usecase?

brunojbarros:
> @brunojbarros thanks for getting back to me. I'm curious, are you performing load testing? Could you share a little bit more about your usecase?

Not really.

[...]

PR #32613, opened 2024-09-13
feat(fetch): record timings

Skn0tt:
Related to https://github.com/microsoft/playwright/issues/19621

Adds some instrumentation to collect timings for `APIRequestContext` requests and adds them to the HAR trace. Doesn't yet expose them via an API, but makes our `Duration` field in the trace viewer show a nice duration:

Screenshot 2024-09-14 at 11 46 04

I'm gonna add it to our API in a separate PR.

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -30,7 +30,7 @@ import { HttpsProxyAgent, SocksProxyAgent } from '../utilsBundle';
 import { BrowserContext, verifyClientCertificates } from './browserContext';
 import { CookieStore, domainMatches } from './cookieStore';
 import { MultipartFormData } from './formData';
-import { httpHappyEyeballsAgent, httpsHappyEyeballsAgent } from '../utils/happy-eyeballs';
+import { httpHappyEyeballsAgent, httpsHappyEyeballsAgent, timingForSocket } from '../utils/happy-eyeballs';
 import type { CallMetadata } from './instrumentation';
 import { SdkObject } from './instrumentation';
 import type { Playwright } from './playwright';
@@ -488,8 +488,9 @@ export abstract class APIRequestContext extends SdkObject {
 
       request.on('socket', socket => {
         // happy eyeballs don't emit lookup and connect events, so we use our custom ones
-        dnsLookupAt = (socket as any).dnsLookupAt;
-        tcpConnectionAt = (socket as any).tcpConnectionAt;
+        const happyEyeBallsTimings = timingForSocket(socket);
+        dnsLookupAt = happyEyeBallsTimings.dnsLookupAt;
+        tcpConnectionAt = happyEyeBallsTimings.tcpConnectionAt;
 
         // non-happy-eyeballs sockets
         socket.on('lookup', () => { dnsLookupAt = monotonicTime(); });
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -311,29 +311,26 @@ export abstract class APIRequestContext extends SdkObject {
       const requestOptions = { ...options, agent };
 
       const startAt = monotonicTime();
-      const timings: Record<'startAt' | 'requestFinishAt' | 'dnsLookupAt' | 'tcpConnectionAt' | 'tlsHandshakeAt' | 'firstByteAt' | 'endAt', number | undefined> = {
-        startAt,
-        dnsLookupAt: undefined,
-        tcpConnectionAt: undefined,
-        tlsHandshakeAt: undefined,
-        requestFinishAt: undefined,
-        firstByteAt: undefined,
-        endAt: undefined
-      };
+      let dnsLookupAt: number | undefined;
+      let tcpConnectionAt: number | undefined;
+      let tlsHandshakeAt: number | undefined;
+      let requestFinishAt: number | undefined;
+      let firstByteAt: number | undefined;
+      let endAt: number | undefined;
 
       const request = requestConstructor(url, requestOptions as any, async response => {
-        response.once('readable', () => { timings.firstByteAt = monotonicTime(); });
-        response.once('end', () => { timings.endAt = monotonicTime(); });
+        response.once('readable', () => { firstByteAt = monotonicTime(); });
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/utils/happy-eyeballs.ts
```diff
@@ -132,9 +134,13 @@ export async function createConnectionAsync(
         port: options.port as number,
         host: address });
 
+    (socket as any).dnsLookupAt = dnsLookupAt;
```
dgozman, line 137: I think we should make this a proper part of the API in some way.

packages/playwright-core/src/server/fetch.ts
```diff
@@ -307,8 +309,27 @@ export abstract class APIRequestContext extends SdkObject {
...
+      let tlsHandshakeAt: number | undefined;
+      let requestFinishAt: number | undefined;
+      let firstByteAt: number | undefined;
+      let endAt: number | undefined;
+
       const request = requestConstructor(url, requestOptions as any, async response => {
         const notifyRequestFinished = (body?: Buffer) => {
+          const timings: har.Timings = {
+            send: requestFinishAt! - startAt,
+            wait: firstByteAt! - requestFinishAt!,
```
dgozman, line 325: What if body was empty, and we did not set `firstByteAt`?

packages/playwright-core/src/server/fetch.ts
```diff
@@ -307,8 +309,27 @@ export abstract class APIRequestContext extends SdkObject {
...
+      let tcpConnectionAt: number | undefined;
+      let tlsHandshakeAt: number | undefined;
+      let requestFinishAt: number | undefined;
+      let firstByteAt: number | undefined;
+      let endAt: number | undefined;
+
       const request = requestConstructor(url, requestOptions as any, async response => {
         const notifyRequestFinished = (body?: Buffer) => {
+          const timings: har.Timings = {
+            send: requestFinishAt! - startAt,
```
dgozman, line 324: I see this method called multiple times, how are we sure that `requestFinishAt` has been set in all of them?

packages/playwright-core/src/server/fetch.ts
```diff
@@ -307,8 +309,27 @@ export abstract class APIRequestContext extends SdkObject {
...
+            wait: firstByteAt! - requestFinishAt!,
+            receive: endAt! - firstByteAt!,
+            dns: dnsLookupAt ? dnsLookupAt - startAt : -1,
+            connect: (tlsHandshakeAt ?? tcpConnectionAt!) - startAt,
```
dgozman, line 328: Why not just `tcpConnectionAt - startAt`? This warrants a comment at least.

packages/playwright-core/src/server/fetch.ts
Skn0tt: `notifyRequestFinished` is called exclusively inside the callback of `requestConstructor`, which is a listener for the `response` event. My reading of the Node.js docs is that the `finish` event (which sets `requestFinishAt`) is guaranteed to fire before the `response` event.

packages/playwright-core/src/server/fetch.ts
Skn0tt: fair! added comments in 947942349aa4c08cb7942821ad4bf09984d0b332

packages/playwright-core/src/server/fetch.ts
Skn0tt: Good catch. I've replaced `firstByteAt` with a measure of when the response headers are sent in 5d8ab8a33cde506ca1076fe349fd530870c17b69. I think that's more in-line with the HAR definition anyways.

packages/playwright-core/src/utils/happy-eyeballs.ts
Skn0tt: Agree it's a little hacky, but I can't find anything better. How would you do it?

packages/playwright-core/src/server/har/harTracer.ts
```diff
@@ -212,6 +212,10 @@ export class HarTracer {
     harEntry.response.statusText = event.statusMessage;
     harEntry.response.httpVersion = event.httpVersion;
     harEntry.response.redirectURL = event.headers.location || '';
+
+    harEntry.timings = event.timings;
```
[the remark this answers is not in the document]
Skn0tt: done in 2d0919c0f64b19eab53a2f0889e09e8a43f9b510

packages/playwright-core/src/server/fetch.ts
```diff
@@ -307,8 +309,28 @@ export abstract class APIRequestContext extends SdkObject {
...
+
       const request = requestConstructor(url, requestOptions as any, async response => {
+        const responseAt = monotonicTime();
         const notifyRequestFinished = (body?: Buffer) => {
+          const endAt = monotonicTime();
+          // spec: http://www.softwareishard.com/blog/har-12-spec/#timings
+          const timings: har.Timings = {
+            send: requestFinishAt! - startAt,
+            wait: responseAt - requestFinishAt!,
+            receive: endAt! - responseAt,
```
dgozman, line 327:
```suggestion
            receive: endAt - responseAt,
```

packages/playwright-core/src/utils/happy-eyeballs.ts
dgozman: Create a function here called `timingForSocket(socket)` that can read values that were set through symbols to avoid clashing with proper fields. Basically, encapsulate the internal symbols to this file.

dgozman (approved):

Skn0tt:
Should we be respecting `options.omitTiming`? I think it makes sense for timing, but I don't think we should respect `options.omitCookies` for API Requests for example.

dgozman:
> Should we be respecting `options.omitTiming`? I think it makes sense for timing, but I don't think we should respect `options.omitCookies` for API Requests for example.

Nice catch! I think we should respect all the options.

Skn0tt:
I've implemented `omitTiming` in https://github.com/microsoft/playwright/pull/32613/commits/f4a986e61afb67c9d9497c3132cc083a819d3ae6, and will do the rest in a separate PR.

dgozman (approved):

PR #32621, opened 2024-09-13
docs(docker): fix Docker container permissions

mxschmitt:
This reverts https://github.com/microsoft/playwright/commit/39a555513b01c4dc157f64c1fb0d52b4523c7877 and applies the recommended workaround instead. `1001` is the uid of the user on the host machine aka. the `1001(runner)` user. I tested that it results in the Firefox startup error without the `HOME` env and passing the `--user` fixes it. Thinking if we should add a comment and link to some upstream bug or say at least something about it.

Reference:

- https://github.com/cypress-io/cypress-docker-images#firefox-not-found
- https://github.com/cypress-io/cypress-documentation/commit/b9854ccdff69e6035defe2a5047034bb36875bcd
- https://github.com/microsoft/playwright/issues/6500

### DIFF
--- a/docs/src/ci.md
+++ b/docs/src/ci.md
@@ -209,6 +209,7 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: mcr.microsoft.com/playwright:v%%VERSION%%-jammy
+      options: --user 1001
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -218,8 +219,6 @@ jobs:
         run: npm ci
       - name: Run your tests
         run: npx playwright test
-        env:
-          HOME: /root
 ```
 
 ```yml python title=".github/workflows/playwright.yml"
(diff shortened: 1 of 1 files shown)


Skn0tt (approved):

mxschmitt:
cc @whimboo do you think we could relax [this check](https://github.com/mozilla/gecko-dev/blob/782abe7d859314a0d8a8b9832931731413ef6ea1/toolkit/xre/nsAppRunner.cpp#L3860) on Firefox side? I think that the testing ecosystem would benefit from it. Including WebDriver, Cypress and Playwright. It has caused a lot of confusion on our side and Cypress side too.

whimboo:
I think https://bugzilla.mozilla.org/show_bug.cgi?id=1323302#c5 explains it very well that we do not want an escape hatch that would put users / systems on risk.

mxschmitt:
okay, thanks for looking into it tho!

PR #32628, opened 2024-09-15
fix(recorder): address the react race condition

pavelfeldman:


packages/playwright-core/src/server/recorder.ts
```diff
@@ -135,7 +135,8 @@ export class Recorder implements InstrumentationListener, IRecorder {
     this._contextRecorder.on(ContextRecorder.Events.Change, (data: { sources: Source[], primaryFileName: string }) => {
       this._recorderSources = data.sources;
       this._pushAllSources();
-      this._recorderApp?.setFileIfNeeded(data.primaryFileName);
+      if (this._mode === 'none')
+        this._recorderApp?.setFile(data.primaryFileName);
```
dgozman, line 139: This seems to always switch to the `primaryFileName`, even after user explicitly changed from "javascript" to "python" in the language picker. Am I missing something?

pavelfeldman: When?

dgozman (approved):

PR #32629, opened 2024-09-15
[Feature] IDE mode trace viewer

ruifigueira:
Companion PR of https://github.com/microsoft/playwright-vscode/pull/531.

For VS code and trace viewer to communicate using events, this PR adds support to multiple websockets connections and event broadcasting.

It also extends testServer to provide trace viewer support, as well as converting embedded trace viewer page into IDE mode, which uses `testServerConnection` to dispatch events to VS code.

### DIFF
--- a/packages/playwright-core/src/utils/httpServer.ts
+++ b/packages/playwright-core/src/utils/httpServer.ts
@@ -17,7 +17,7 @@
 import type http from 'http';
 import fs from 'fs';
 import path from 'path';
-import { mime, wsServer } from '../utilsBundle';
+import { mime, wsServer, type WebSocket } from '../utilsBundle';
 import { assert } from './debug';
 import { createHttpServer } from './network';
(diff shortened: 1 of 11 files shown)


pavelfeldman:
@ruifigueira we are slow on reviewing these patches because it is unclear if this mode of trace viewer is making its way into the IDE atm. As I've shared previously, it does not seem like it is improving the experience significantly, while it introduces quite some maintenance burden in order to support multiple modes.

pavelfeldman:
I'm closing this as non-actionable as we have not made a decision on trace-in-IDE UX yet.

ruifigueira:
just a note: this was not the trace-in-IDE, it's similar to spawn traceviewer, the only differences is that it makes use of testServer to launch the traceviewer app, and establishes a vscode<->testserver<->traceviewer connection, so that vscode can listen to traceviewer events like selecting an action.

PR #32631, opened 2024-09-16
feat(watch mode): buffer mode

### LINKED ISSUE #32578: [Internal]: Give watch mode a buffer mode
Tests can take long to run, so running on every edit can be nerve-racking. It's the worst when somebody has Autosave enabled.

We should have a "buffer" mode that accrues a list of changed tests and executes them upon `Enter`.

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32578.

Adds a buffer mode that can be toggled by pressing <kbd>b</kbd>. When engaged, changed test files are collected and shown on screen. The test run is then kicked off by pressing <kbd>Enter</kbd>.

This changes the signal to start a test run from <kbd>Cmd+s</kbd> to a <kbd>Enter</kbd> press in the test terminal. It should help users with auto-save and make it easier to run on long-running tests. It feels very similar to running `npx playwright test`, but without having to write a filter.

https://github.com/user-attachments/assets/71e16139-9427-4e90-b523-8d218f09ed9d

### DIFF
--- a/packages/playwright/src/runner/watchMode.ts
+++ b/packages/playwright/src/runner/watchMode.ts
@@ -152,8 +152,8 @@ export async function runWatchModeLoop(configLocation: ConfigLocation, initialOp
     if (bufferMode && command === 'changed')
       continue;
 
-    const commandToTriggerTestRun = (bufferMode ? 'run' : 'changed');
-    if (command === commandToTriggerTestRun) {
+    const shouldRunChangedFiles = bufferMode ? command === 'run' : command === 'changed';
+    if (shouldRunChangedFiles) {
       if (dirtyTestIds.size === 0)
         continue;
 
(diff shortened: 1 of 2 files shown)


packages/playwright/src/runner/watchMode.ts
```diff
@@ -134,21 +139,27 @@ export async function runWatchModeLoop(configLocation: ConfigLocation, initialOp
...
+      readCommand(),
     ]);
-    if (!readCommandPromise.isDone())
-      readCommandPromise.resolve('changed');
 
-    const command = await readCommandPromise;
+    if (bufferMode && command === 'changed')
+      continue;
+
+    const commandToTriggerTestRun = (bufferMode ? 'run' : 'changed');
```
dgozman, line 155: I think something like `shouldRunChangedFiles: boolean` will be easier to understand. I had to wrap my head around this construct.

dgozman (approved):

PR #32633, opened 2024-09-16
docs: add reference to locator strictness if or resolves to multiple elements

### LINKED ISSUE #30069: [Bug]: Error: strict mode violation when using _or (with .locator())... resolved to 2 elements ?
### Version

1.42.0

### Steps to reproduce

(I have simplified the scenario below to try make it more explainable)

I have a page with an iframe. Inside the iframe I have two divs with id #a and #b which are attached to the DOM, but may be visible or hidden. I am trying to write code that waits till one of the two are visible, before proceeding.

```python
my_iframe = page.frame_locator('iframe[title*="foo"]')
a_locator = my_iframe.locator("#a")
b_locator = my_iframe.locator("#b")

await a_locator.is_visible(), await b_locator.is_visible() # <------ this returns False, True
```

But when I do this...

```python
await expect(a_locator.or_(b_locator)).to_be_visible(timeout=3000)
```

I get the following error:
```
Error: Error: strict mode violation: 
locator("#a).or_(frame_locator("iframe[title...").locator("#b")) resolved to 2 
elements:
    1) <div id="a"></div> aka locator("#a")
    2) <div id="b" role="alert">…</div> aka locator("#b")
```

But if I run
```
await a.count(), await b.count()  # <---  returns (1, 1) - which confirms there's just one element per locator
```

Is this working as intended?

### Expected behavior

if either a_locator or b_locator are is_visible() i'd expect the _or condition to succeed. It should work even if both are visible.

### Actual behavior

I get a strict mode violation: saying it resolved to 2 elements.

locator("#a).or_(frame_locator("iframe[title...").locator("#b")) resolved to 2 
elements:

### Additional context

_No response_

### Environment

```Text
- Operating System: Mac
- Browser: [Firefox]
- Python Version: [3.12]
```

### LINKED ISSUE #32616: [Docs]: clarify behaviour of or when both locators match
### Page(s)

https://playwright.dev/docs/api/class-locator#locator-or

### Description

I think the documentation on `or` could be improved by mentioning that it can resolve to _both_ locators and fail an assertion.

My expectation was that `or` would pick one locator _or_ the other. If both locators matched, I thought it would return either one. The docs do say: "Creates a locator that matches **either** of the two locators." It was unexpected to me that it actually matched **both**.

I found one other Github issue with someone who I think also tripped over this behaviour: https://github.com/microsoft/playwright/issues/30069

There's some helpful explanation in that issue:

> In your case a_locator.or_(b_locator) resolves to 2 divs which violates the condition.

I think that's confusing considering the docs say:

> Creates a locator that matches _either_ of the two locators.

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/32616
Fixes https://github.com/microsoft/playwright/issues/30069

### DIFF
--- a/docs/src/api/class-locator.md
+++ b/docs/src/api/class-locator.md
@@ -1654,7 +1654,9 @@ var banana = await page.GetByRole(AriaRole.Listitem).Nth(2);
   - alias-python: or_
 - returns: <[Locator]>
 
-Creates a locator that matches either of the two locators. [Locator strictness](./locators.md#strictness) still applies (if it results in multiple elements, it throws).
+Creates a locator matching all elements that match one or both of the two locators.
+
+Note that when both locators match something, the resulting locator will have multiple matches and violate [locator strictness](./locators.md#strictness) guidelines.
 
 **Usage**
 
--- a/docs/src/api/class-locator.md
+++ b/docs/src/api/class-locator.md
@@ -1656,7 +1656,7 @@ var banana = await page.GetByRole(AriaRole.Listitem).Nth(2);
 
 Creates a locator matching all elements that match one or both of the two locators.
 
-Note that when both locators match something, the resulting locator will have multiple matches and violate [locator strictness](./locators.md#strictness) guidelines.
+Note that when both locators match something, the resulting locator will have multiple matches and violate [locator strictness](../locators.md#strictness) guidelines.
 
 **Usage**
 
--- a/docs/src/api/class-locator.md
+++ b/docs/src/api/class-locator.md
@@ -1654,7 +1654,7 @@ var banana = await page.GetByRole(AriaRole.Listitem).Nth(2);
   - alias-python: or_
 - returns: <[Locator]>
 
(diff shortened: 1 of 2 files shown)


docs/src/api/class-locator.md
```diff
@@ -1654,7 +1654,7 @@ var banana = await page.GetByRole(AriaRole.Listitem).Nth(2);
   - alias-python: or_
 - returns: <[Locator]>
 
-Creates a locator that matches either of the two locators.
+Creates a locator that matches either of the two locators. [Locator strictness](./locators.md#strictness) still applies (if it results in multiple elements, it throws).
```
dgozman, line 1657: Perhaps explain a bit more?

```suggestion
Creates a locator matching all elements that match one or both of the two locators.

Note that when both locators match something, the resulting locator will have multiple matches and violate [locator strictness](./locators.md#strictness) guidelines.
```

dgozman (approved):

PR #32638, opened 2024-09-16
feat: make `npx playwright clear-cache` public

dgozman:


mxschmitt (commented):
Should we add docs for it?

yury-s (approved):
I agree with Max, let's add at least some documentation on why one may want to call the command?

PR #32647, opened 2024-09-17
feat(fetch): expose timings

### LINKED ISSUE #19621: [Question] Is there a way to return an api request response time?

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/19621. Adds the same `timings()` we have for browser responses to `APIResponse`. Please apply some extra care in reviewing the timing calculations.

In the protocol change, I was unsure wether to extend the existing `ResourceTiming` type or to add another property. I went with the added property to be in-line with how the events work for browser requests - let me know if we should do it differently instead.


docs/src/api/class-request.md
```diff
@@ -314,7 +314,7 @@ Returns resource size information for given request.
   - `startTime` <[float]> Request start time in milliseconds elapsed since January 1, 1970 00:00:00 UTC
   - `domainLookupStart` <[float]> Time immediately before the browser starts the domain name lookup for the
     resource. The value is given in milliseconds relative to `startTime`, -1 if not available.
-  - `domainLookupEnd` <[float]> Time immediately after the browser starts the domain name lookup for the resource.
+  - `domainLookupEnd` <[float]> Time immediately after the browser ends the domain name lookup for the resource.
```
Skn0tt, line 317: this is a drive-by fix - pretty sure it shouldn't say "start the domain name lookup"

packages/playwright-core/src/server/fetch.ts
```diff
@@ -422,14 +422,35 @@ export abstract class APIRequestContext extends SdkObject {
...
         const notifyBodyFinished = () => {
+          function relativeTime(time: number | undefined): number {
+            if (!time)
+              return -1;
+            return time - startAt;
+          }
+          const endAt = monotonicTime();
+          // spec: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming
+          const timing: channels.ResourceTiming = {
+            startTime: startAt,
```
dgozman, line 433: The docs say `startTime` is a wall time, not monotonic time.

packages/playwright-core/src/server/fetch.ts
```diff
@@ -422,14 +422,35 @@ export abstract class APIRequestContext extends SdkObject {
...
+            domainLookupStart: dnsLookupAt ? 0 : -1,
```
dgozman, line 434: Why zero and not `relativeTime()`?

packages/playwright-core/src/server/fetch.ts
```diff
@@ -422,14 +422,35 @@ export abstract class APIRequestContext extends SdkObject {
...
+            domainLookupEnd: dnsLookupAt ? dnsLookupAt! - startAt : -1,
```
dgozman, line 435: `relativeTime()`?

packages/playwright-core/src/server/fetch.ts
```diff
@@ -422,14 +422,35 @@ export abstract class APIRequestContext extends SdkObject {
...
           notifyRequestFinished(body);
           fulfill({
             url: response.url || url.toString(),
             status: response.statusCode || 0,
             statusText: response.statusMessage || '',
             headers: toHeadersArray(response.rawHeaders),
-            body
+            body,
+            timing,
+            responseEndTiming,
```
dgozman, line 453: Why do we have two sets of timings now?

packages/playwright-core/src/server/fetch.ts
Skn0tt: My understand is that the browser has some other steps like cache resolution before the DNS lookup, so there might be time between startTime and the DNS lookup start. On Node.js, I don't think there's anything between that - so it's zero, because we know the DNS lookup happens immediately after the request start.

packages/playwright-core/src/server/fetch.ts
Skn0tt: For browser requests, most of the timings are transmitted in the `Response` type, and then `responseEndTiming` arrives later in the `requestFinished` and `requestFailed` events. I opted to make this similar, so we have a big set of timings in `timing` and the final timing in `responseEndTiming`.

I thought about amending the `ResourceTiming` type instead, but then `requestFinished` would suddenly have the response end time both in `responseEndTiming` and in `response.timing.responseEnd` - that felt confusing.

packages/playwright-core/src/server/fetch.ts
Skn0tt: done in https://github.com/microsoft/playwright/pull/32647/commits/b1d523b009dc99d856a9dd870dfd639f4671fd00

packages/playwright-core/src/server/fetch.ts
Skn0tt: done in https://github.com/microsoft/playwright/pull/32647/commits/b717257013f459955f0172483e1b52b55ac7305d

Skn0tt:
We discussed this with the team and decided against exposing this via the API. Playwright isn't a network performance testing tool, and we don't want people using it like one. Rough timings can easily be measured in userland.

I'll open a separate PR to fix the bugs around HTTP / SOCKS Proxy we found in the existing measurements for HAR timings.

PR #32684, opened 2024-09-18
test: fix toolbar hydration test under frozen suite

mxschmitt:
Local repro: `PW_CLOCK=frozen npm run ctest -- --headed library/inspector/cli-codegen-3.spec.ts:743`

### DIFF
--- a/packages/playwright-core/src/server/injected/recorder/recorder.ts
+++ b/packages/playwright-core/src/server/injected/recorder/recorder.ts
@@ -1039,9 +1039,12 @@ export class Recorder {
 
     this.highlight.install();
     // some frameworks erase the DOM on hydration, this ensures it's reattached
-    const recreationInterval = setInterval(() => {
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/injected/recorder/recorder.ts
```diff
@@ -1039,7 +1039,7 @@ export class Recorder {
 
     this.highlight.install();
     // some frameworks erase the DOM on hydration, this ensures it's reattached
-    const recreationInterval = setInterval(() => {
+    const recreationInterval = this.injectedScript.builtinSetTimeout(() => {
```
dgozman, line 1042: This should be setInterval. Or call setTimeout recursively.

mxschmitt: good catch!

dgozman (approved):

PR #32687, opened 2024-09-18
chore: move Location type from testReporter.d.ts to test.d.ts

### LINKED ISSUE #32680: [Bug]: Playwright Test should export the Location interface
### Version

1.48.0-alpha-2024-09-18

### Steps to reproduce

PR #32504 introduces an overridable `location` parameter for the `test.step` API, to be included in Playwright 1.48.0.
 
When testing the new `test.step` API using build `1.48.0-alpha-2024-09-18`, I noticed that the `Location` type used by the `location` parameter is not exported from `@playwright/test`. This results in TypeScript throwing error `TS2353: Object literal may only specify known properties, and 'file' does not exist in type 'Location'.` during type checking, as TypeScript assumes that `Location` refers to `Location` type from `lib-dom`, rather than Playwright's [`Location`](https://github.com/microsoft/playwright/blob/9fc195bff5cd4a5dcb4337d8c9a4400683c9c62b/packages/playwright/types/testReporter.d.ts#L325-L340) interface.

To reproduce the error, write a Playwright Test script overriding the `location` parameter when calling the `test.step` API:

```ts

test.describe('Playwright Test', () => {

    test('should allow for the location to be overridden', async () => {

        await test.step(
            'my step',
            async () => { },
            { location: { file: 'my-file.ts', line: 42, column: 13 } }
        );
    });
});
```

Add a `tsconfig.json` file to your project, for example:

```json
{
  "compilerOptions": {
    "target": "es2016",
    "module": "commonjs",
    "esModuleInterop": true,
    "strict": true,
    "skipLibCheck": true
  }
}
```

Run the TypeScript type checker:

```
npx tsc --noEmit && npx playwright test
```

Observe the below error in the output:

```
spec/repro.spec.ts:25:27 - error TS2353: Object literal may only specify known properties, and 'file' does not exist in type 'Location'.

25             { location: { file: 'my-file.ts', line: 42, column: 13 } }
```

### Expected behavior

The `Location` type should be exported alongside other public interfaces of `@playwright/test`.
For example, it should be possible to do the following:

```ts
import { test, type TestInfo, type Location } from '@playwright/test'
```

I'd suggest moving the definition of Playwright's `Location` from `testReporter.d.ts` to `playwright/types/test.d.ts` since it's now part of the public API.

### Actual behavior

The [`Location`](https://github.com/microsoft/playwright/blob/9fc195bff5cd4a5dcb4337d8c9a4400683c9c62b/packages/playwright/types/testReporter.d.ts#L325-L340) is not exported

### Additional context

See #32504 and the corresponding feature request #30160 for more context

CC @dgozman, @osohyun0224, @vitalets, @WestonThayer

### Environment

```shell
System:
    OS: macOS 14.6.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 28.66 GB / 96.00 GB
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.17.0/bin/npm
  Languages:
    Bash: 3.2.57 - /bin/bash
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/32680

Problem: `Location` type is a global in `node_modules/typescript/lib/lib.dom.d.ts`.

Previously we used it only in our reporters (`testReporter.d.ts`) now we also use it inside `test.d.ts`. We don't want to import `testReporter` from `test.d.ts`, hence moving the type into `test.d.ts`.


dgozman (approved):
You will need to update sidebars at playwright.dev as well.

jan-molak:
Thanks for the fix, @mxschmitt and @dgozman - I'll check tomorrow when the next nightly build is available.

PR #32693, opened 2024-09-18
chore: iterate towards recording into trace (2)

pavelfeldman:


packages/recorder/src/main.tsx
```diff
@@ -27,7 +27,10 @@ export const Main: React.FC = ({
   const [mode, setMode] = React.useState<Mode>('none');
 
   window.playwrightSetMode = setMode;
-  window.playwrightSetSources = setSources;
+  window.playwrightSetSources = React.useCallback((sources: Source[]) => {
+    setSources(sources);
+    window.playwrightSourcesEchoForTest = sources;
+  }, []);
```
yury-s, line 33:
```suggestion
  }, [setSources]);
```
don't we need this dependency?

packages/playwright-core/src/server/recorder/recorderUtils.ts
```diff
@@ -213,62 +214,119 @@ export function callMetadataForAction(pageAliases: Map<Page, string>, actionInCo
...
       });
       continue;
     }
-    if (event.method === 'press') {
-      const tokens = event.params.key.split('+');
-      const modifiers = tokens.slice(0, tokens.length - 1);
+    if (method === 'press') {
+      const params = untypedParams as channels.FramePressParams;
+      const tokens = params.key.split('+');
+      const modifiers = tokens.slice(0, tokens.length - 1) as SmartKeyboardModifier[];
```
yury-s, line 323:
```suggestion
      const modifiers = tokens.slice(0, - 1) as SmartKeyboardModifier[];
```

yury-s (approved):

packages/recorder/src/main.tsx
pavelfeldman: state setters should not be passed into the watches

PR #32504, opened 2024-09-08
feat(test runner): allow to pass arbitrary location to test.step

osohyun0224:
Fixes https://github.com/microsoft/playwright/issues/30160

### Description:
This pull request introduces the ability to specify custom locations for test steps in Playwright. By enabling the provision of arbitrary locations to the test.step method, it resolves the limitation where helper methods obfuscate the original call site, providing more accurate and meaningful location data in test reports.

### Motivation:
To enhance the utility and clarity of test reports in Playwright. Specifically, it addresses the need to trace test steps back to their precise location in the code, which is especially important when steps are abstracted in helper functions. This feature is crucial for maintaining accurate documentation and facilitating debugging processes.

### Changes:
Added functionality to pass a custom location object to test.step.

### Expected Outcome:
This PR is expected to significantly improve the precision and usefulness of diagnostic data in test reports by allowing specific locations within helper functions to be accurately documented. It facilitates better tracking of test executions and simplifies the debugging process, making it easier for developers to understand and address issues within complex tests.

### References:
Closes https://github.com/microsoft/playwright/issues/30160 - "[Feature]: allow to pass arbitrary location to test.step"

**Code Check**
I conducted tests on this new feature by integrating it into some existing test codes, and it worked well. I will attach the code used for testing and a screenshot showing the successful outcome.

<details>
<summary>toggle dropdown</summary>
<div markdown="1">

```
import type { Location } from '../../../packages/playwright/types/testReporter'
...
test('should respect the back button', async ({ page }) => {
    await page.locator('.todo-list li .toggle').nth(1).check();
    await checkNumberOfCompletedTodosInLocalStorage(page, 1);
...
    await test.step('Showing active items', async () => {
      await page.getByRole('link', { name: 'Active' }).click();
    }, {location});
```

image
</div>
</details>

Issue #30160, opened 2024-03-28
[Feature]: allow to pass arbitrary location to test.step


vitalets:
### 🚀 Feature Request

Currently `test.step` method binds location in file to the place of it's call. 
When `test.step` is wrapped with helper methods, it produces the same location for different steps.
Suggestion is to allow to pass arbitrary location to `test.step`. See example below.

### Example

Imagine I have a `helper.ts` where I define step helper for my tests:
```ts
export async function clickButton(text: string) {
  return test.step(`User clicks button "${text}"`, async () => {
    // ...
  }); 
}
```
Now I run the following test from `./e2e/test.ts`:
```ts
test('test', async ({}) => {
  await clickButton('foo');
  await clickButton('bar');
  await clickButton('baz');
});
```
In the report I get all steps with the same location:
<img width="50%" src="https://github.com/microsoft/playwright/assets/1473072/06191a10-4d0b-4b85-9d90-74079dff88f1">

It would be great to see more meaningful locations in the report, e.g.:
```
User clicks button "foo"— e2e/test.ts:2
User clicks button "bar"— e2e/test.ts:3
User clicks button "baz"— e2e/test.ts:4
```

If I were able to provide a custom location to `test.step`, I would achieve it. Example:
```ts
import { Location } from '@playwright/test/reporter';
import errorStackParser from 'error-stack-parser';

export async function clickButton(text: string) {
  const location = getStepLocation();
  return test.step(`User clicks button "${text}"`, async () => {
    // ...
  }, { location }); 
}

function getStepLocation(): Location {
  const frame = errorStackParser.parse(new Error()).find((frame) => frame.fileName.endsWith('.test.ts'));
  return {
      file: frame.getFileName(),
      line: frame.getLineNumber(),
      column: frame.getColumnNumber(),
  };
}
```

### Motivation

In general this feature provides more opportunities for writing high order step helpers.

Personally, I need it for managing `playwright-bdd` project. I'm wrapping `test.step` with `Given / When / Then` helpers to have correct step titles in the report:
```ts
const Given = (text: string) => test.step(`Given ${text}`, () => { ... });
const When = (text: string) => test.step(`When ${text}`, () => { ... });
const Then = (text: string) => test.step(`Then ${text}`, () => { ... });
```
[...]

Issue #29323, opened 2024-02-02
[Feature]: Provide more reliable way to map attachments with steps


vitalets:
### 🚀 Feature Request

Provide more reliable way to map attachments with steps. It can be unique `attachmentId` field or some other approach.
Please, see example below.

### Motivation

This feature will give more opportunities to third-party reporters.
Personally I need this for [implementing Cucumber reports in playwright-bdd](https://github.com/vitalets/playwright-bdd/issues/9). As in Cucumber each attachment is mapped to a particular step although in Playwright attachments are mapped to a whole test result.

### Example

Imagine the following test:
```ts
test("my test", async () => {
  await test.step('step 1', async () => {
    await test.info().attach('my attachment', { body: 'foo' });
  });
  await test.step('step 2', async () => {
    await test.info().attach('my attachment', { body: 'bar' });
  });
});
```
When running this test I get in `onTestEnd` the following (Playwright version 1.41.2):
**result.attachments:**
```
[
  {
    name: 'my attachment',
    path: undefined,
    contentType: 'text/plain',
    body: <Buffer 66 6f 6f> // <- foo
  },
  {
    name: 'my attachment',
    path: undefined,
    contentType: 'text/plain',
    body: <Buffer 62 61 72> // <- bar
  }
]
```
**result.steps:**
```js
...
{
  title: 'step 1',
  titlePath: [Function: titlePath],
  parent: undefined,
  category: 'test.step',
  startTime: 2024-02-02T14:57:40.863Z,
  duration: 1,
  steps: [
    {
      title: 'attach "my attachment"',
      titlePath: [Function: titlePath],
      parent: [Circular *1],
      category: 'attach',
      startTime: 2024-02-02T14:57:40.863Z,
      duration: 1,
      steps: [],
      location: [Object],
      [Symbol(id)]: '609f19b4a231e384f62045156c98fafe'
    }
  ],
},
{
  title: 'step 2',
  titlePath: [Function: titlePath],
  parent: undefined,
  category: 'test.step',
  startTime: 2024-02-02T14:57:40.864Z,
  duration: 0,
  steps: [
    {
      title: 'attach "my attachment"',
      titlePath: [Function: titlePath],
      parent: [Circular *1],
      category: 'attach',
      startTime: 2024-02-02T14:57:40.864Z,
      duration: 0,
      steps: [],
      location: [Object],
      [Symbol(id)]: '97c374b0a6362757b8333da5f1263208'
    }
  ],
}
```
[...]

PR #32704, opened 2024-09-19
feat(playwright-test): test.step supports passing step params to trace view

jan-molak:
Extending the functionality requested in #30160 and #32680 and building on #32504 and #32687, this pull request enables the `test.step` API to accept an optional `params` object, to be included in the trace viewer.

This enables test automation frameworks that integrate with Playwright Test, such as [Serenity/JS](https://serenity-js.org), to add extra context to Playwright reports.

Example usage:

```ts
// helper.ts
import { test, TestType } from '@playwright/test';

export async function sayHello(name: string) {
  await test.step('says hello', () => {}, { params: { name } });
}
```
```ts
// playwright.config.ts
module.exports = {
  reporter: './reporter',
  use: {
    trace: 'on',
  }
};
```
```ts
// a.test.ts
import { sayHello } from './helper';

test('params', async () => {
  await sayHello('World');
});
```

Example result (using [Serenity/JS Playwright example](https://github.com/serenity-js/serenity-js/tree/main/examples/playwright-test-todomvc)):
Screenshot 2024-09-19 at 11 27 51

CC: @dgozman, @mxschmitt, @vitalets, @osohyun0224, @WestonThayer

### DIFF
--- a/docs/src/test-api/class-test.md
+++ b/docs/src/test-api/class-test.md
@@ -1715,6 +1715,11 @@ Whether to box the step in the report. Defaults to `false`. When the step is box
 - `location` <[Location]>
 Specifies a custom location for the step to be shown in test reports. By default, location of the [`method: Test.step`] call is shown.
 
+### option: Test.step.params
+* since: v1.48
+- `params` ?<[Object]>
+An optional key-value object where keys are step parameter names and values are their corresponding parameter values. This information is shown in the trace view.
+
 ## method: Test.use
 * since: v1.10
 
--- a/packages/playwright/src/common/testType.ts
+++ b/packages/playwright/src/common/testType.ts
@@ -259,11 +259,11 @@ export class TestTypeImpl {
     suite._use.push({ fixtures, location });
   }
 
-  async _step<T>(title: string, body: () => Promise<T>, options: {box?: boolean, location?: Location } = {}): Promise<T> {
+  async _step<T>(title: string, body: () => Promise<T>, options: { box?: boolean, location?: Location, params?: Record<string, any> } = {}): Promise<T> {
     const testInfo = currentTestInfo();
     if (!testInfo)
       throw new Error(`test.step() can only be called from a test`);
-    const step = testInfo._addStep({ category: 'test.step', title, location: options.location, box: options.box });
+    const step = testInfo._addStep({ category: 'test.step', title, location: options.location, params: options.params, box: options.box });
     return await zones.run('stepZone', step, async () => {
       try {
         const result = await body();
--- a/packages/playwright/types/test.d.ts
+++ b/packages/playwright/types/test.d.ts
@@ -4703,7 +4703,7 @@ export interface TestType<TestArgs extends KeyValue, WorkerArgs extends KeyValue
    * @param body Step body.
    * @param options
    */
-  step<T>(title: string, body: () => T | Promise<T>, options?: { box?: boolean, location?: Location }): Promise<T>;
+  step<T>(title: string, body: () => T | Promise<T>, options?: { box?: boolean, location?: Location, params?: Record<string, any> }): Promise<T>;
   /**
    * `expect` function can be used to create test assertions. Read more about [test assertions](https://playwright.dev/docs/test-assertions).
    *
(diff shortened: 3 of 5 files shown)


jan-molak:

pavelfeldman:
I'm not sure I like the change. It fits the steps infrastructure nicely with the params being passed directly into the trace, but at the same time it is unclear why it is bypassing the reporters API. Could you give me a better idea on a couple of typical use cases for this?

jan-molak:
Hi @pavelfeldman and thanks for reviewing the PR. 

> Could you give me a better idea on a couple of typical use cases for this?

Sure thing. Trace viewer is very useful in debugging complex test scenarios as it provides information about all the low-level Playwright API calls, network requests, and so on. However, I am not aware of a way to make it report the higher-level abstractions alongside the low-level calls.

For example, consider a test scenario that requires setting up some test data before making Playwright interact with the UI. Let's say we're testing a product search functionality and we need to inject product-related data into the system:

```ts
const products = [
  { name: 'apple', price: '£2.50', .... },
  { name: 'oranges', .. }, 
  // etc.
]
```

Let's also assume that setting up each product is itself quite involved and requires several REST API calls.

We could model a function to set up a test product using `test.step`, which allows us to abstract the lower-level API calls:

```ts
describe('Search', () => {

  it('find products based on name', ({ page, request }) => {

    for (const product of products) {
      await createTestProduct(product, request)
    }

    await page.goto('...')
    // enter search term, find the product, and so on
  
  })
}

async function createTestProduct(product: Product, request): Promise<void> {
  await test.step('create test product', async () => {
    await request... // create product
    await request... // update product details
    await request... // add product to search categories
    // ...
  });
}
```

The limitation here is that even though the report will now show several calls to `'create test product'`, it won't show us the exact details of the `product` object in a single location (yes, we could deduce the injected data from the details of lower-level requests, but that makes it more inconvenient than having it in one place).

While this example is pretty basic, I'm sure you could easily see the challenge if instead of apples and oranges we wanted to set up several test customers in a banking system, each customer with one or more accounts, each account with an initial balance, overdraft limits and so on. 

Being able to make Playwright aware of those higher-level parameters would make the report even more useful in such complex scenarios.

I hope this helps. Please let me know if you have any concerns or if there's a different way available in Playwright to support this use case.

pavelfeldman:
We could invest in improving the rendering of attachments:

```js

const products = [
  { name: 'apples', price: '£2.50' },
  { name: 'oranges', price: '$2' },
];

test.describe('Search', () => {
  test('find products based on name', async ({}) => {
    for (const product of products)
      await createTestProduct(product);
  });
});

async function createTestProduct(product): Promise<void> {
  await test.step('create test product', async () => {
    // ...
    await test.info().attach('product', { body: JSON.stringify(product, undefined, 2), contentType: 'text/plain' });
  });
}
```

that way information would get both into the report and the trace.

vitalets:
Linking a similar issue for attributing steps with attachments data: https://github.com/microsoft/playwright/issues/32748

Some points from previous discussions, that are important for attachments + steps:
1. Attachments can't be reliably mapped to parallel steps (see [#29323](https://github.com/microsoft/playwright/issues/29323#issuecomment-1924645277)): 
   > We lack expressiveness of the API to reliably map attachments to steps. If steps run concurrently, we don't know which step to insert the attachment into without changing the step API and passing the attachment sink in it.

2. Attachments [are not available](https://github.com/microsoft/playwright/issues/29323#issuecomment-1927311800) in `onStepEnd` in _merge-reports_ mode.

pavelfeldman:
These are good points, but to me they speak in favor of prioritizing the step attachment fixes! Closing this PR, please file an issue where we could track it. I'm supportive of fixing attachments for steps on the API level.

jan-molak:
@pavelfeldman @vitalets - would injecting the `testInfo` object into `test.step` address the issue with parallel steps? I think an interface like this could be quite neat:

```ts
async function createTestProduct(product): Promise<void> {
  await test.step('create test product', async ({ testInfo }) => {
    // ...
    await testInfo.attach('product', { body: JSON.stringify(product, undefined, 2), contentType: 'text/plain' });
  });
}
```

PR #32713, opened 2024-09-19
chore: enable @babel/plugin-syntax-import-attributes all the time

### LINKED ISSUE #32659: [Bug]: "Support for the experimental syntax 'importAttributes' isn't currently enabled"
### Version

v1.47.1

### Steps to reproduce

This script will reproduce the issue:

```bash
#!/usr/bin/env bash

docker run -i --rm --workdir /app mcr.microsoft.com/playwright:v1.47.1 bash <<EOD

npm init -y
npm i -D @playwright/test@1.47.1

cat <<EOF > playwright.config.mjs
import { defineConfig, devices } from "@playwright/test";
const config = defineConfig({ testDir: "./", projects: [{ name: "chromium", use: devices["Desktop Chrome"] }] });
export default config;
EOF

cat <<EOF > example.test.ts
import features from './data/features.json' with { type: 'json' };
EOF

npx playwright test
EOD
``` 

### Expected behavior

I expected to see Playwright say that there were no tests found to run.

### Actual behavior

```
SyntaxError: /app/example.test.ts: Support for the experimental syntax 'importAttributes' isn't currently enabled (1:76):

> 1 | import features from './data/features.json' with { type: 'json' };
    |                                                                            ^
  2 |

Add @babel/plugin-syntax-import-attributes (https://github.com/babel/babel/tree/main/packages/babel-plugin-syntax-import-attributes) to the 'plugins' section of your Babel config to enable parsing.

If you already added the plugin for this syntax to your config, it's possible that your config isn't being loaded.
You can re-run Babel with the BABEL_SHOW_CONFIG_FOR environment variable to show the loaded configuration:
        npx cross-env BABEL_SHOW_CONFIG_FOR=/app/foo.test.ts <your build command>
See https://babeljs.io/docs/configuration#print-effective-configs for more info.


   at example.test.ts:1

> 1 | import features from './data/features.json' with { type: 'json' };
    |                                                                           ^
  2 |
Error: No tests found
``` 

### Additional context

I found https://github.com/microsoft/playwright/pull/20009 and https://github.com/microsoft/playwright/issues/19928, which seem to imply that import assertions should be supported. I've checked and the relevant babel plugin _is_ still embedded in the Playwright bundle, so I'm not sure why it's not running properly.

### Environment

```shell
# docker run --rm -it mcr.microsoft.com/playwright:v1.47.1 bash

  System:
    OS: Linux 6.10 Ubuntu 24.04 LTS 24.04 LTS (Noble Numbat)
    CPU: (12) arm64 unknown
    Memory: 12.47 GB / 17.54 GB
    Container: Yes
  Binaries:
    Node: 20.17.0 - /usr/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 10.8.2 - /usr/bin/npm
  Languages:
    Bash: 5.2.21 - /usr/bin/bash
```

mxschmitt:
Fixes https://github.com/microsoft/playwright/issues/32659

Test is a copy of https://github.com/microsoft/playwright/blob/cc302fa5150734d712d6e1c02bd8c5a42a308633/tests/playwright-test/esm.spec.ts#L38 - maybe we should remove the test in esm.spec.ts and run it twice, once with type module and once without?

### DIFF
--- a/packages/playwright/bundles/babel/src/babelBundleImpl.ts
+++ b/packages/playwright/bundles/babel/src/babelBundleImpl.ts
@@ -45,6 +45,7 @@ function babelTransformOptions(isTypeScript: boolean, isModule: boolean, plugins
         [require('@babel/plugin-syntax-async-generators')],
         [require('@babel/plugin-syntax-object-rest-spread')],
         [require('@babel/plugin-transform-export-namespace-from')],
+        [require('@babel/plugin-syntax-import-attributes'), { deprecatedAssertSyntax: true }],
         [
           // From https://github.com/G-Rath/babel-plugin-replace-ts-export-assignment/blob/8dfdca32c8aa428574b0cae341444fc5822f2dc6/src/index.ts
           (
@@ -86,8 +87,6 @@ function babelTransformOptions(isTypeScript: boolean, isModule: boolean, plugins
         }
       })
     ]);
-  } else {
-    plugins.push([require('@babel/plugin-syntax-import-attributes'), { deprecatedAssertSyntax: true }]);
   }
 
   return {
--- a/tests/playwright-test/babel.spec.ts
+++ b/tests/playwright-test/babel.spec.ts
@@ -151,29 +151,31 @@ test('should not transform external', async ({ runInlineTest }) => {
   expect(result.output).toMatch(/(Cannot use import statement outside a module|require\(\) of ES Module .* not supported.)/);
 });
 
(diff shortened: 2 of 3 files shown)


dgozman:
> - maybe we should remove the test in esm.spec.ts and run it twice, once with type module and once without?

Or just move it to `babel.spec`.

dgozman (approved):

PR #32724, opened 2024-09-20
chore: throw on nested toPass calls

mxschmitt:


packages/playwright/src/matchers/matchers.ts
```diff
@@ -405,6 +407,8 @@ export async function toPass(
     timeout?: number,
   } = {},
 ) {
+  if (toPassContextStorage.getStore() !== undefined)
```
dgozman, line 410: Our policy is "no runtime behavior relies on zones".

PR #32728, opened 2024-09-20
docs: explain glob patterns

mxschmitt:
Playwright globs are not idiomatic or following a standard. We saw users often being confused about them. Since in all the language bindings we follow the same algorithm, I think its a great opportunity to document it, make it clearer for the users and for us.

### DIFF
--- a/docs/src/network.md
+++ b/docs/src/network.md
@@ -700,6 +700,28 @@ await Page.RouteAsync("**/title.html", async route =>
 });
 ```
 
+## Glob URL patterns
+
+Playwright uses glob patterns for URL matching in network interception methods like [`method: Page.route`] or [`method: Page.waitForResponse`]. These glob patterns are more specific than typical file globs and offer features similar to regular expressions:
+
+1. Asterisks:
+   - A single `*` matches any characters except `/`
+   - A double `**` matches any characters including `/`
+1. Question mark `?` matches any single character except `/`
+1. Character sets `[]` work like [RegExp character sets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Character_classes#types):
+   - `[abc]` matches any single character from the set (a, b, or c)
+   - `[a-z]` matches any single character in the range a to z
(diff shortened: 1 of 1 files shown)


docs/src/network.md
```diff
@@ -700,6 +700,28 @@ await Page.RouteAsync("**/title.html", async route =>
 });
 ```
 
+## Glob URL patterns
+
+Playwright uses glob patterns for URL matching in network interception methods like [`method: Page.route`] or [`method: Page.waitForResponse`]. These glob patterns are more specific than typical file globs and offer features similar to regular expressions:
```
yury-s, line 705:
```suggestion
Playwright uses glob patterns for URL matching in network interception methods like [`method: Page.route`] or [`method: Page.waitForResponse`]. These glob patterns are more limited than typical file globs:
```

docs/src/network.md
```diff
@@ -700,6 +700,28 @@ await Page.RouteAsync("**/title.html", async route =>
...
+
+1. Asterisks:
+   - A single `*` matches any characters except `/`
+   - A double `**` matches any characters including `/`
+1. Question mark `?` matches any single character except `/`
+1. Character sets `[]` work like [RegExp character sets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Character_classes#types):
+   - `[abc]` matches any single character from the set (a, b, or c)
+   - `[a-z]` matches any single character in the range a to z
+1. Alternatives `{}` work like a logical OR:
```
yury-s, line 714: These don't support nesting in our case. I'd not include this at all.

yury-s (commented):
Not sure if we want this documented, the implementation is incomplete and diverges from others in many places. Maybe document just the core syntax (*, **, ?) and recommend regex for anything more complicated?

yury-s (approved):
I feel like we'll get more questions about nested {} and ranges [].

mxschmitt:
let's see how it goes in the long term, happy to update once users are asking!

Issue #32703, opened 2024-09-19
[Bug]: Hover with trial should not trigger mouseover event


muhqu:
### Version

1.47.1

### Steps to reproduce

1. Clone my repo https://github.com/muhqu/playwright-issues-with-trial
2. npm install
3. npx playwright test
4. See it fail

### Expected behavior

* `locator.click({ trial: true })`
  * should not trigger `click` event.
* `locator.hover({ trial: true })`
  * should not trigger `mouseover` event.
* `locator.click({ trial: true, modifiers: ['Shift'] })` 
  * should not trigger `click` event
  * should not trigger `keydown` event

### Actual behavior

* `locator.click({ trial: true })`
  * 🟢 does not trigger `click` event. 
* `locator.hover({ trial: true })`
  * 🔴 **does trigger** `mouseover` event. 
* `locator.click({ trial: true, modifiers: ['Shift'] })` 
  * 🟢 does not trigger `click` event. 
  * 🔴 **does trigger** `keydown` event

### Environment

```shell
System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 138.27 MB / 64.00 GB
  Binaries:
    Node: 22.3.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.8.1 - /opt/homebrew/bin/npm
    pnpm: 7.28.0 - /opt/homebrew/bin/pnpm
  IDEs:
    VSCode: 1.93.0 - /usr/local/bin/code
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.47.1 => 1.47.1
```

1234abd:
try this
https://mega.co.nz/#!B31G2LaZ!uBBOCp9hLC7bq9kP8NC6s4DanQJJZuoFRr5FI_jkeic
I put the necessary dlls in the archive

mxschmitt:
What is your use-case for `hover` with `trial: true`?

The issue with `trial: true` in click for `keydown` events is definitely valid, but there can be scenarios that the target element e.g. changes / disappears / remounts on keydown events, hence having a semantically correct `trial`. In order to not have a potential regression there and break users we would hold back for now.

muhqu:
[...]

PR #32731, opened 2024-09-20
fix(hover): don't emit mouseover event on hover with trial: true

mxschmitt:
https://github.com/microsoft/playwright/issues/32703


yury-s (approved):

muhqu:
@mxschmitt @yury-s this fix is incomplete. 

When you use `hover({trial:true})` followed by `hover()` the second call will not trigger mousemove or mouseover events as the mouse is already in the right position due to the trial run. Here is an alternate fix that restores the mouse position on trial: https://github.com/microsoft/playwright/pull/32736

mxschmitt:
thanks, makes sense. Let's p3 for now and collect feedback about it before we start extending this logic.

PR #32733, opened 2024-09-20
test: unflake 'should properly synchronize local and remote time' test

mxschmitt:
We run the Trace Viewer in the tests using `launchApp` which uses [`noDefaultViewport: true`](https://github.com/microsoft/playwright/blob/33890eb6c5298b239b42763e4652b3c7e40bf68f/packages/playwright-core/src/server/launchApp.ts#L47). Our Trace Viewer does not always render `8.5s`, it also renders just `8.0s`, based on the viewport which was given. The window size on CI is probably just 1024x720 or something like that (assumption).

We could also check for `toBeLessThan`.


yury-s (approved):
toBeLessThan sounds better to me, but up to you. Thanks for fixing it!

PR #32736, opened 2024-09-20
fix(trial): restore mouse position and intercept mouseover event on hover trial

muhqu:
https://github.com/microsoft/playwright/issues/32703


pavelfeldman:
I don't think we are going with this one for hoever({trial}). Open to a PR that deprecates the trial in hover though.

Issue #32562, opened 2024-09-11
[Feature]: Custom AsymmetricMatchers


amberUCP:
### 🚀 Feature Request

The current custom matcher `extend` function does not seem to allow custom asymmetric matchers.

E.g. `fixtures.ts`
```
import { expect as baseExpect } from '@playwright/test';
export { test } from '@playwright/test';

export const expect =  baseExpect.extend({
  async toBeNumberOrNull(received) {
    const assertionName = `toBeNullOrType`;
    let pass: boolean;
    let matcherResult: any;
    const expected = 'Null or Number';
    let isNumber = false;
    let isNull = false;

    try {
      await baseExpect(received).toEqual(baseExpect.any(Number));
      
      isNumber = true;
    } catch (e: any) {
      matcherResult = e.matcherResult;
      isNumber = false;
    }

    try {
      await baseExpect(received).toEqual(null);
      isNull = true;
    } catch (e: any) {
      matcherResult = e.matcherResult;
      isNull = false;
    }
    
    pass = isNull || isNumber;

    const message = pass
      ? () => this.utils.matcherHint(assertionName, undefined, undefined, { isNot: this.isNot }) +
          '\n\n' +
          `Expected: ${this.isNot ? 'not' : ''}${this.utils.printExpected(expected)}\n` +
          (matcherResult ? `Received: ${this.utils.printReceived(matcherResult.actual)}` : '')
      : () =>  this.utils.matcherHint(assertionName, undefined, undefined, { isNot: this.isNot }) +
          '\n\n' +
          `Expected: ${this.utils.printExpected(expected)}\n` +
          (matcherResult ? `Received: ${this.utils.printReceived(matcherResult.actual)}` : '');

    return {
      message,
      pass,
      name: assertionName,
      expected,
      actual: matcherResult?.actual,
    };
  }
});
```

This works when writing an symmetric assertion such as:
```
  test('Test matcher', async () => {
   await expect(1).toBeNumberOrNull();
  });
```

This does not work when writing an asymmetric assertion such as:
```test.describe('Orders', () => {
  test('Gets order', async ({ request }) => {
    const response = await request.get('/order');

    expect(response.status()).toBe(200);

    const response = await response.json();

    expect(response.data).toMatchObject({
      id: expect.any(String),
      name: expect.any(String),
      item: {
        id: expect.any(Number),
        price: expect.any(Number),
        discount: expect.toBeNumberOrNull(), // this custom asymmetric matcher is not currently supported
      },
[...]

PR #32740, opened 2024-09-20
feat(extend): custom asymmetric matchers

muhqu:
#32562


dgozman:
@muhqu Thank you for the PR!

We have actually fixed the regression introduced by #32366 in #32795. I've added a test for asymmetric matches to make sure it won't regress in the future in #32829.

That said, we are not really sure this is the API we would love to support long-term, therefore we don't want to add types support for it right now. We'll think about the API as a part of #32562.

I think we can close this PR right now, unless I am missing something and there are changes we should land. Let me know if that's the case. Thank you for contributing!

muhqu:
@dgozman okay, sounds fair. I'm glad you fixed it for backward compatibility.

However, I think you should really consider retaining the type information of asymmetric matchers. It just feels broken, if there is something supported under the hood, but not exposed via TypeScript. Are you questioning extensibility of expect() in general? … we can also take the discussion to https://github.com/microsoft/playwright/issues/32562 if that makes sense.

I've updated this PR to only include the type definition fix.

dgozman:
@muhqu As I said, we are not sure we'd like to keep this API for asymmetric matchers going forward. It's better to keep discussion to the issue instead of the PR for future discoverability though.

pavelfeldman:
Closing this so that we could discuss it in the issue instead.

Issue #14584, opened 2022-06-02
[Feature] Allow test run to continue the case execution if a test.step() fails


gegoncalves:
It would be nice to have a flag or something that could be use to allow a test run go trough all the `test.step()`, even if any of them fail.

Something like: `test.describe.configure({ mode: 'continue' });` 😄 

The reason for this request is explained in this comment from the first line until the first bullet point: https://github.com/microsoft/playwright/issues/14508#issuecomment-1144617999

aslushnikov:
@gegoncalves You can use [`expect.soft`](https://playwright.dev/docs/test-assertions#soft-assertions) inside the steps, and they will continue running. Would it work for you?

gegoncalves:
@aslushnikov unfortunately no. If was only the assertion that would be nice. However, if there is any timeout before the other steps, it is going to stop the run.

nonsense code example: 4th line gonna timeout and stop the execution
````ts
import { test, expect } from '@playwright/test';

test('basic test', async ({ page }) => {
    await test.step('Test 1 Not fail', async () => {
        await page.goto('https://playwright.dev/');
        await page.locator('.navbar__inner .sdfdsfsdf').click();
        const title = page.locator('.navbar__inner .navbar__title');
        await expect(title).toHaveText('Playwright');
    });

    await test.step('Test 2 pass', async () => {
        await page.goto('https://playwright.dev/');
        const title = page.locator('.navbar__inner .navbar__title');
        await expect(title).toHaveText('Playwright');
    });

    await test.step('Test 3 should fail', async () => {
        await page.goto('https://playwright.dev/');
        const title = page.locator('.navbar__inner .navbar__title');
        await expect(title).toHaveText('blabla');
    });
});
````

aslushnikov:
> nonsense code example: 4th line gonna timeout and stop the execution

@gegoncalves ah, yes. In this case you'll have to fallback to using tests instead of steps. There are no plans to make test execution continue on step failure (aka `test.step.soft`)

gegoncalves:
Ah ok. That way I will be back to the problem of multiple instances of browsers being used from the issue https://github.com/microsoft/playwright/issues/14508#issuecomment-1144617999

I need to figure out how to [reuse single page between tests](https://playwright.dev/docs/test-retries#reuse-single-page-between-tests) using custom Fixtures. Unfortunately, is not possible to do using context or page.

Here is a stupid code question: @aslushnikov  Is it possible to do this code snippet on my basePage class?

[...]

Issue #14508, opened 2022-05-31
[Question] How to make test() use the same browser instance as the beforeAll. (Ignore test isolation)


gegoncalves:
~~UPDATE: Check my comment https://github.com/microsoft/playwright/issues/14508#issuecomment-1143274596~~
UPDATE 2: Check my comment https://github.com/microsoft/playwright/issues/14508#issuecomment-1144617999

I am trying to reuse the authentication example provided [at the documentation](https://playwright.dev/docs/test-auth#reuse-the-signed-in-page-in-multiple-tests) for multiple .spec files (I know that is not recommended but it is necessary for my case)

Is there any recommendation of POM with fixture that I can do to solve that issue?

I want to avoid rewrite this part multiple times:

```
test.beforeAll(async ({ browser }) => {
  // Create page once and sign in.
  page = await browser.newPage();
  await page.goto('https://github.com/login');
  await page.fill('input[name="user"]', 'user');
  await page.fill('input[name="password"]', 'password');
  await page.click('text=Sign in');
});
```

gegoncalves:
Unfortunately no. I am using a similar idea to [this one](https://github.com/akshayp7/playwright-typescript-playwright-test/blob/main/lib/BaseTest.ts)

When I go to the next test() it opens another browser instance instead of using the one created in beforeall.

I did what you said in a different file and extended it changing the line 14 and 15 for the { browser }



> Doesn't this work for you?
> 
> ```ts
> // my-test.ts
> import { Page, test as base } from '@playwright/test';
> 
> interface MyFixtures {
>   myPage: Page;
> }
> 
> export const myTest = base.extend<MyFixtures>({
>   myPage: async ({ browser }, use) => {
>     // Create page once and sign in.
>     const page = await browser.newPage();
>     await page.goto('https://github.com/login');
>     await page.fill('input[name="user"]', 'user');
>     await page.fill('input[name="password"]', 'password');
>     await page.click('text=Sign in');
>     use(page);
>   },
> });
> 
> // example.spec.ts
> myTest.describe('My suite', () => {
>   myTest('My test', async ({ myPage }) => {
>     await myPage.click('selector');
>   });
> });
> ```

kryaksy:
Yeah, got it now. This might work then. But I don't know if it behaves as you wish for multiple test suites.

```typescript
// my-test.ts
import { Page, test as base } from '@playwright/test';

interface MyWorkerScopedFixtures {
  myPage: Page;
}

[...]

PR #32753, opened 2024-09-22
added NON_ISOLATED_TESTS env for getting round test isolation if needed

udarrr:
I was inspired the requests and also i'd like to use that for my needs
https://github.com/microsoft/playwright/issues/14584
https://github.com/microsoft/playwright/issues/14508

In the PR was added NON_ISOLATED_TESTS env variable it's preventing creation new worker after failed test cases. All test cases within a suite will be executed in the same worker. That mean 

For test.describe/test file

- test.beforeAll, test.afterAll - will be executed only once even if any of test was failed
- workerFixture - will be executed only once even if any of test was failed
- error in test.beforeAll/workerFixture  will fail all tests
- error in test.afterAll/workerTeardown will fail the latest test case
- worker - will be the same for all tests in a test.describe/test file, but different in other test.describe/test files

For some reason especially for heavy conditions it's really helpful.

Hope playwright's team will support the feature as community is interested in the flexibility

### DIFF
--- a/packages/playwright/src/runner/dispatcher.ts
+++ b/packages/playwright/src/runner/dispatcher.ts
@@ -401,7 +401,7 @@ class JobDispatcher {
     // - there are no remaining
     // - we are here not because something failed
     // - no unrecoverable worker error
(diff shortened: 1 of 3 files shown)


pavelfeldman:
That probably won't work on too many levels. Please file your own issue with rationale!

PR #32764, opened 2024-09-23
fix(test runner): cleanup DEBUG_COLORS usage

### LINKED ISSUE #32543: [Feature]: allow disabling colors in debug reporter
### 🚀 Feature Request

I use jenkins CI with built in `list` and `junit` reporters.
To improve debugging test flakes, i set `DEBUG=pw:api,pw:browser`

It would be wonderful to have an explicit way to disable colors for the debug output.
I tried various hacks around this, even setting variables such as `DEBUG_FILE=/dev/stderr`, but this fails with ENXIO due to the way that node opens the file. I was hoping this would engage the color-filtering here: https://github.com/microsoft/playwright/blob/b5d968fa0e3acb052de4eab5952f6efb668d9111/packages/playwright-core/src/utils/debugLogger.ts#L40-L44

### Example

_No response_

### Motivation

CI logs are filled with ansi colors with no way to disable them.
I don't give playwright process a TTY (runs via docker) but the debugging logs are still colored.
Unfortunately, NO_COLOR environment variable is not respected: https://no-color.org/

It makes it extremely difficult to debug tests from:
* jenkins "Test Report" (/testReport). This displays the output from the `junit` reporter for a particular test (e.g. the one that failed) and is the most intuitive and fast way to triage a failing test. Unfortunately all the debug output shows as ansi escapes.
* full jenkins logfiles. We have a lot of tests and scrolling with the built-in console can be frustrating.

dgozman:
`DEBUG_COLORS` we default to `1`, but we should not do that when it is already defined to some value supplied by the user.

Closes #32543.

### DIFF
--- a/packages/playwright/src/runner/workerHost.ts
+++ b/packages/playwright/src/runner/workerHost.ts
@@ -37,7 +37,7 @@ export class WorkerHost extends ProcessHost {
     super(require.resolve('../worker/workerMain.js'), `worker-${workerIndex}`, {
       ...extraEnv,
       FORCE_COLOR: '1',
-      DEBUG_COLORS: '1',
+      DEBUG_COLORS: process.env.DEBUG_COLORS === undefined ? '1' : process.env.DEBUG_COLORS,
     });
     this.workerIndex = workerIndex;
     this.parallelIndex = parallelIndex;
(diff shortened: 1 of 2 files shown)


packages/playwright/src/runner/workerHost.ts
```diff
@@ -36,8 +36,7 @@ export class WorkerHost extends ProcessHost {
     const workerIndex = lastWorkerIndex++;
     super(require.resolve('../worker/workerMain.js'), `worker-${workerIndex}`, {
       ...extraEnv,
-      FORCE_COLOR: '1',
-      DEBUG_COLORS: '1',
+      DEBUG_COLORS: process.env.DEBUG_COLORS === undefined ? '1' : process.env.DEBUG_COLORS,
```
mxschmitt, line 39: I think the same as `process.env.DEBUG_COLORS ?? '1'` but more explicit.

mxschmitt (approved):

mxschmitt (approved):

PR #32774, opened 2024-09-23
test: add different expectation for macOS/headed 'should work for canvas'

mxschmitt:


yury-s (approved):
Can we just update the expectation or it varies between platforms?

dgozman (approved):

PR #32824, opened 2024-09-26
chore: address api review for page.forceGarbageCollection

dgozman:
- Renamed to `page.requestGC`.
- Added a useful snippet to the docs.

References #32278.

### DIFF
--- a/docs/src/api/class-page.md
+++ b/docs/src/api/class-page.md
@@ -2333,10 +2333,57 @@ last redirect. If cannot go forward, returns `null`.
 
 Navigate to the next page in history.
 
-## async method: Page.forceGarbageCollection
-* since: v1.47
+## async method: Page.requestGC
+* since: v1.48
 
-Force the browser to perform garbage collection.
+Request the page to perform garbage collection. Note that there is no guarantee that all unreachable objects will be collected.
+
+This is useful to help detect memory leaks. For example, if your page has a large object `'suspect'` that might be leaked, you can check that it does not leak by using a [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef).
+
+```js
+// 1. In your page, save a WeakRef for the "suspect".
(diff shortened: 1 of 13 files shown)


docs/src/api/class-page.md
```diff
@@ -2333,10 +2333,57 @@ last redirect. If cannot go forward, returns `null`.
...
+
+This is useful to help detect memory leaks. For example, if your page has a large object `'suspect'` that might be leaked, you can check that it does not leak by using a [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef).
+
+```js
+// 1. In your page, save a WeakRef for the "suspect".
+await page.evaluate(() => globalThis.suspectWeakRef = new WeakRef(suspect));
+// 2. Request garbage collection.
+await page.requestGC();
+// 3. Check that weak ref does not deref to the original object.
+expect(await page.evaluate(() => globalThis.suspectWeakRef.deref())).toBe(undefined);
```
mxschmitt, line 2349: `deref()` could return the object. The object might be a class which we only serialise in a best-effort manner. I recommend to either do `!!` inside.

mxschmitt (approved):

docs/src/api/class-page.md
```diff
@@ -2333,10 +2333,57 @@ last redirect. If cannot go forward, returns `null`.
...
+```
 
-Force the browser to perform garbage collection.
+```python async
+# 1. In your page, save a WeakRef for the "suspect".
+await page.evaluate("globalThis.suspectWeakRef = new WeakRef(suspect)")
+# 2. Request garbage collection.
+await page.request_gc()
+# 3. Check that weak ref does not deref to the original object.
+assert page.evaluate("!globalThis.suspectWeakRef.deref()")
```
mxschmitt, line 2367:
```suggestion
assert await page.evaluate("!globalThis.suspectWeakRef.deref()")
```

mxschmitt (approved):

PR #32830, opened 2024-09-26
docs: check that description has an empty line before it

dgozman:


mxschmitt (approved):
Can we adjust our linter to throw in such cases?

dgozman:
> Can we adjust our linter to throw in such cases?

Yes we can!

mxschmitt (approved):

PR #32855, opened 2024-09-27
fix(har timing): record `connect` timing for proxied connections

Skn0tt:
Fixes a bug discovered in https://github.com/microsoft/playwright/pull/32647. When using http proxy, the `connect` event isn't emitted so we don't populate `tcpConnectionAt`. The updated version of `https-proxy-agent` emits a `proxyConnect` as a replacement, so this PR updates and listens to that event.
For socks proxies, the `on("socket")` event is emitted once the SOCKS connection is established, which is the equivalent of having a TCP connection available.

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -508,6 +508,10 @@ export abstract class APIRequestContext extends SdkObject {
           }
         });
 
+        // when using socks proxy, having the socket means the connection got established
+        if (agent instanceof SocksProxyAgent)
+          tcpConnectionAt ??= monotonicTime();
+
         serverIPAddress = socket.remoteAddress;
         serverPort = socket.remotePort;
       });
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -697,7 +697,7 @@ export function createProxyAgent(proxy: types.ProxySettings) {
     proxyURL.username = proxy.username;
   if (proxy.password)
     proxyURL.password = proxy.password;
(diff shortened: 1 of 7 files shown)


packages/playwright-core/src/server/fetch.ts
```diff
@@ -680,17 +689,16 @@ export class GlobalAPIRequestContext extends APIRequestContext {
...
+    return new SocksProxyAgent(proxyURL);
+
   if (proxy.username)
-    proxyOpts.auth = `${proxy.username}:${proxy.password || ''}`;
+    proxyURL.username = proxy.username;
+  if (proxy.password)
+    proxyURL.password = proxy.password;
   // TODO: We should use HttpProxyAgent conditional on proxyOpts.protocol instead of always using CONNECT method.
-  return new HttpsProxyAgent(proxyOpts);
+  return new HttpsProxyAgent(proxy.server);
```
mxschmitt, line 701:
```suggestion
  return new HttpsProxyAgent(proxyURL);
```

shouldn't it be like that?

Skn0tt: yes! :facepalm:

Skn0tt: yes! :facepalm:

dgozman:
Do we have to upgrade the deps? I am always wary of introducing regressions.

Skn0tt:
For SOCKS proxy we don't, but for HTTP proxy yes. The `proxyConnect` event was only added in `6.1.0` and without it we have no way of knowing when the proxy started the connection.

If we don't want to update, another alternative would be leave things as for the HTTP proxy and set `connect: -1`.

Skn0tt:
Could you give this another look @dgozman?

dgozman (approved):

PR #32858, opened 2024-09-27
fix(ui): turn "copy as fetch" into text button

Skn0tt:
Turns the "copy as fetch" text with a copy button into a text button, as discussed in the meeting.

https://github.com/user-attachments/assets/01f72f0b-e3f2-440e-a75d-33385aabeec4

### DIFF
--- a/packages/trace-viewer/src/ui/copyToClipboard.css
+++ b/packages/trace-viewer/src/ui/copyToClipboard.css
@@ -18,6 +18,5 @@
   background-color: var(--vscode-editor-inactiveSelectionBackground);
   border: none;
   padding: 4px 12px;
-  margin: 8px 8px 8px 0px;
   cursor: pointer;
 }
--- a/packages/trace-viewer/src/ui/copyToClipboard.css
+++ b/packages/trace-viewer/src/ui/copyToClipboard.css
@@ -17,9 +17,7 @@
 .copy-to-clipboard-text-button {
   background-color: var(--vscode-editor-inactiveSelectionBackground);
   border: none;
+  padding: 4px 12px;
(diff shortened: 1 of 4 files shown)


pavelfeldman:
Let's make sure button does not change text as (once) I press it, it is really confusing and non-a11y friendly.

Skn0tt:
> Let's make sure button does not change text as (once) I press it

Done. I've also resized it a little, here's how it looks now:

Screenshot 2024-09-30 at 07 25 31

packages/trace-viewer/src/ui/copyToClipboard.css
```diff
@@ -0,0 +1,23 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+.copy-to-clipboard-text-button {
+  background-color: var(--vscode-editor-inactiveSelectionBackground);
+  border: none;
+  padding: 4px 12px;
+  margin: 8px 8px 8px 0px;
```
dgozman, line 21: I think it is strange to put margin into a component. I'd expect the parent component to manage margins and positioning of children.

dgozman (approved):

packages/trace-viewer/src/ui/copyToClipboard.css
Skn0tt: you're right. done in https://github.com/microsoft/playwright/pull/32858/commits/2761ad9d84444950b9579c9115469cc695f62f11

dgozman (approved):

Issue #32861, opened 2024-09-28
[Docs]: Getting started - VS Code


mattias-m-andersson:
### Page(s)

- [Getting started - VS Code](https://playwright.dev/docs/getting-started-vscode)

### The problem
On the page [getting started vscode](https://playwright.dev/docs/getting-started-vscode) the "getting started" is brought to an immediate halt on "Install Playwright" with the following error message in VS Code Terminal.

`PS C:\Dev\Playwright> npm init playwright@latest --yes -- --quiet --browser=chromium --browser=firefox --browser=webkit --gha
npm: The term 'npm' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.`

Nowhere is it mentioned the need for the package manager NPM, or how to install it.

### How I got into it...
After listening to [this](https://open.spotify.com/episode/0RJIFMnIsuV228NgGmQoKN?si=d644b32a4997425b) episode of podcast .NET Rocks, where it is mentioned that "... give it 7 minutes and try it out", I was so exited to get started.

### Description

Clearly state the dependencies and guide me through the process of getting there.
Also how to trouble shoot common errors would be nice.

vladkrasn:
Npm is a Node.js thing.
For .NET, see this https://playwright.dev/dotnet/docs/intro

Skn0tt:
Right, the VS Code Extension is only meant for usage with Node.js Playwright. I'll see what we can do to make this more clear.

mattias-m-andersson:
[...]

PR #32882, opened 2024-09-30
chore: call out that that the vscode extension is for Node.js

Skn0tt:
In https://github.com/microsoft/playwright/issues/32861, an interested .NET user wanted to try Playwright and found the VS Code Getting Started guide. It didn't work for them because the VS Code Extension is for usage with Node.js, and they don't have NPM installed. We can reduce confusion by mentioning that VS Code Getting started is for Node.js.

### DIFF
--- a/docs/src/getting-started-vscode-js.md
+++ b/docs/src/getting-started-vscode-js.md
@@ -18,7 +18,7 @@ Get started by installing Playwright and generating a test to see it in action. 
 
 ## Installation
 
-Install the [VS Code extension from the marketplace](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) or from the extensions tab in VS Code.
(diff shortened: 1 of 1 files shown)


docs/src/getting-started-vscode-js.md
```diff
@@ -18,7 +18,7 @@ Get started by installing Playwright and generating a test to see it in action.
 
 ## Installation
 
-Install the [VS Code extension from the marketplace](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) or from the extensions tab in VS Code.
+To use Playwright for Node.js from VS Code, install the [VS Code extension from the marketplace](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) or from the extensions tab in VS Code.
```
Skn0tt, line 21:
```suggestion
Playwright has a VS Code extension which is available when testing with Node.js. Install [it from the VS Code marketplace](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) or from the extensions tab in VS Code.
```

as per debbie's suggestion

Skn0tt:
Debbie's saying she approves.

PR #32884, opened 2024-09-30
Adds easier to use keyboard shortcuts for test runner

tommy-mode:


pavelfeldman:
These don't look easier to me. Please always start with filing the issue btw!

PR #32899, opened 2024-10-01
feat: wait for pending navigation to resolve before many actions

dgozman:
This includes all actions that perform locator handler check.
    
Note this makes it impossible to interact with the page while a main frame navigation is ongoing. This was already the case for Chromium, but now WebKit and Firefox align with it.

Setting `PLAYWRIGHT_SKIP_NAVIGATION_CHECK` environment variable disables this behavior.

### DIFF
--- a/packages/playwright-core/src/server/dom.ts
+++ b/packages/playwright-core/src/server/dom.ts
@@ -292,7 +292,7 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
     };
   }
 
-  async _retryAction(progress: Progress, actionName: string, action: (retry: number) => Promise<PerformActionResult>, options: { trial?: boolean, force?: boolean, skipLocatorHandlersCheckpoint?: boolean }): Promise<'error:notconnected' | 'done'> {
+  async _retryAction(progress: Progress, actionName: string, action: (retry: number) => Promise<PerformActionResult>, options: { trial?: boolean, force?: boolean, skipActionPreChecks?: boolean }): Promise<'error:notconnected' | 'done'> {
     let retry = 0;
     // We progressively wait longer between retries, up to 500ms.
     const waitTime = [0, 20, 100, 100, 500];
@@ -310,8 +310,8 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
       } else {
         progress.log(`attempting ${actionName} action${options.trial ? ' (trial run)' : ''}`);
       }
-      if (!options.skipLocatorHandlersCheckpoint && !options.force)
-        await this._frame._page.performLocatorHandlersCheckpoint(progress);
+      if (!options.skipActionPreChecks && !options.force)
+        await this._frame._page.performActionPreChecks(progress);
       const result = await action(retry);
       ++retry;
       if (result === 'error:notvisible') {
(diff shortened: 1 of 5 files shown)


packages/playwright-core/src/server/dom.ts
```diff
@@ -292,7 +292,7 @@ export class ElementHandle<T extends Node = Node> extends js.JSHandle<T> {
     };
   }
 
-  async _retryAction(progress: Progress, actionName: string, action: (retry: number) => Promise<PerformActionResult>, options: { trial?: boolean, force?: boolean, skipLocatorHandlersCheckpoint?: boolean }): Promise<'error:notconnected' | 'done'> {
+  async _retryAction(progress: Progress, actionName: string, action: (retry: number) => Promise<PerformActionResult>, options: { trial?: boolean, force?: boolean, skipActionPreChecls?: boolean }): Promise<'error:notconnected' | 'done'> {
```
pavelfeldman, line 295:
```suggestion
  async _retryAction(progress: Progress, actionName: string, action: (retry: number) => Promise<PerformActionResult>, options: { trial?: boolean, force?: boolean, skipActionPreChecks?: boolean }): Promise<'error:notconnected' | 'done'> {
```

packages/playwright-core/src/server/frames.ts
```diff
@@ -248,6 +248,11 @@ export class FrameManager {
     const frame = this._frames.get(frameId);
     if (!frame)
       return;
+    const pending = frame.pendingDocument();
```
pavelfeldman, line 251: Looks unrelated, land as a separate change?

packages/playwright-core/src/server/page.ts
```diff
@@ -455,7 +456,31 @@ export class Page extends SdkObject {
     this._locatorHandlers.delete(uid);
   }
 
-  async performLocatorHandlersCheckpoint(progress: Progress) {
+  async performActionPreChecks(progress: Progress) {
+    await this._performWaitForNavigationCheck(progress);
+    progress.throwIfAborted();
+    await this._performLocatorHandlersCheckpoint(progress);
```
pavelfeldman, line 462: What if this causes navigation?

pavelfeldman (approved):

packages/playwright-core/src/server/page.ts
dgozman: Good call, I'll wait for navigations after this one as well.

chrisbottin:
Hey @dgozman 

We have noticed that some of our tests stopped working after bumping our Playwright version from `1.48.2` to `1.49.0`.

I debugged the issue and I pinned it down to the changes in this pull request.

The problem we are having is when we click on a link which triggers the `beforeunload` dialog, after we dismiss the dialog and try to interact with the page again, we get timeout errors.

Here is a simple test replicating the error ...

```js
const {expect} = require('@playwright/test');

(async () => {

    await page.mainFrame().goto('https://www.npmjs.com/');

    expect(await page.mainFrame().title()).toStrictEqual('npm | Home');

    await page.evaluate(() => {
        window.onbeforeunload = () => false;
    });

    page.on('dialog', async dialog => {
        await dialog.dismiss();
    });

    await page.getByRole('menuitem', {name: 'Teams'}).click({noWaitAfter: true});

    await page.evaluate(() => {
        window.onbeforeunload = null;
    });

    // Timeout error will occur here:
    await page.getByRole('menuitem', {name: 'Teams'}).click({timeout: 1000});

    expect(await page.mainFrame().title()).toStrictEqual('npm | Teams');

    await browser.close();
})();
```

The test fails on the line `await page.getByRole('menuitem', {name: 'Teams'}).click({timeout: 1000});` with the error:

```
locator.click: Timeout 1000ms exceeded.
Call log:
  - waiting for getByRole('menuitem', { name: 'Teams' })
  - waiting for navigation to finish...
```

During debugging, I noticed the `pendingDocument()` is set to `{documentId: undefined, request: undefined}`

If this line https://github.com/microsoft/playwright/blob/b5bd543cc6935f62732da3b6183bd319eb2bdc2a/packages/playwright-core/src/server/page.ts#L472 is changed to the following, our problem disappears:

```js
if (!mainFrame || !mainFrame.pendingDocument() || !mainFrame.pendingDocument().documentId)
      return;
```

Is this a genuine issue that should be fixed?

dgozman:
@chrisbottin This is a genuine issue, I filed #33806 on your behalf. We'll take a look.

chrisbottin:
@dgozman Thanks, I also raised https://github.com/microsoft/playwright/issues/33803 after posting my comment here.
I've closed it as duplicate now.

PR #32900, opened 2024-10-01
fix(ct-react): support shorthand fragment notation

### LINKED ISSUE #32853: [Bug]: @playwright/experimental-ct-react fails to render if the root element is a React fragment
### Version

1.47.2

### Steps to reproduce

Code:
```tsx
import { test, expect } from '@playwright/experimental-ct-react';

test('should work', async ({ mount }) => {
  const component = await mount(<>Learn React</>); // fails to render
  await expect(component).toContainText('Learn React');
});
```
or:
1. Clone repo https://github.com/Kirdock/playwright-bugs
2. `npm install`
3. `npm run test:ct`
4. Execute the test that is shown
5. React error. Nothing rendered

### Expected behavior

The component renders even if a React fragment is on top.

### Actual behavior

Nothing is rendred at all. Empty page with react errors
image

### Additional context

Also worth to mention: `npm init playwright@latest -- --ct` did not work for me. The component test part was not included

### Environment

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (32) x64 13th Gen Intel(R) Core(TM) i9-13950HX
    Memory: 35.94 GB / 63.69 GB
  Binaries:
    Node: 20.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.8.3 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.15.4 - C:\Program Files\nodejs\pnpm.CMD
  IDEs:
    VSCode: 1.92.2 - C:\Users\klaus.striessnig\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 5.1.16 - C:\WINDOWS\system32\bash.EXE
  npmPackages:
    @playwright/experimental-ct-react: 1.47.2 => 1.47.2 
    @playwright/test: ^1.47.2 => 1.47.2
```

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32853

Vite turns the shorthand fragment notation `<></>` into `import { Fragment } from "react"; <Fragment></Fragment>`. On the Node.js side of things, this `react` import resolves to our mock version of React, which currently mocks `Fragment` as `{}`. Currently, we pass that straight to `React.createElement`, which throws an error.

The fix is to make our `Fragment` mock detectable with a tag, and when we render it replace it with the real `__pwReact.Fragment`.

### DIFF
--- a/packages/playwright-ct-react/registerSource.mjs
+++ b/packages/playwright-ct-react/registerSource.mjs
@@ -40,7 +40,7 @@ function __pwRender(value) {
     if (isJsxComponent(v)) {
       const component = v;
       let type = component.type;
-      if (type && type.__pw_type === 'fragment')
+      if (type && type.__pw_jsx_fragment) // detects the playwright mock Fragment (see jsx-runtime.js), used in <></> notation
         type = __pwReact.Fragment;
       const props = component.props ? __pwRender(component.props) : {};
       const key = component.key ? __pwRender(component.key) : undefined;
--- a/packages/playwright-ct-react/registerSource.mjs
+++ b/packages/playwright-ct-react/registerSource.mjs
@@ -34,7 +34,7 @@ function isJsxComponent(component) {
 
 /**
  * @param {any} type
- * @returns type is Playwright's mock JSX.Fragment
+ * @returns {boolean} type is Playwright's mock JSX.Fragment
  */
 function isJsxFragment(type) {
   return typeof type === 'object' && type?.__pw_jsx_fragment;
--- a/packages/playwright-ct-react/registerSource.mjs
+++ b/packages/playwright-ct-react/registerSource.mjs
@@ -40,7 +40,7 @@ function __pwRender(value) {
     if (isJsxComponent(v)) {
       const component = v;
       let type = component.type;
-      if ('__pw_type' in type && type.__pw_type === 'fragment')
+      if (type && type.__pw_type === 'fragment')
         type = __pwReact.Fragment;
       const props = component.props ? __pwRender(component.props) : {};
       const key = component.key ? __pwRender(component.key) : undefined;
--- a/packages/playwright-ct-react/registerSource.mjs
+++ b/packages/playwright-ct-react/registerSource.mjs
@@ -33,6 +33,7 @@ function isJsxComponent(component) {
 }
 
 /**
+ * Turns the Playwright representation of JSX (see jsx-runtime.js) into React.createElement calls.
  * @param {any} value
  */
 function __pwRender(value) {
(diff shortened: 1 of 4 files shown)


tests/playwright-test/playwright.ct-react.spec.ts
```diff
@@ -596,3 +596,22 @@ test('should allow import from shared file', async ({ runInlineTest }) => {
   expect(result.exitCode).toBe(0);
   expect(result.passed).toBe(1);
 });
+
+
+test('should render fragment at root', async ({ runInlineTest }) => {
```
mxschmitt, line 601: Maybe we can put the test [here](https://github.com/microsoft/playwright/blob/main/tests/components/ct-react-vite/tests/render.spec.tsx) instead? Since the tests in this file create a new pwt project while in the file I suggested they run in bulk.

packages/playwright-ct-react/registerSource.mjs
```diff
@@ -39,6 +39,9 @@ function __pwRender(value) {
   return window.__pwTransformObject(value, v => {
     if (isJsxComponent(v)) {
       const component = v;
+      let type = component.type;
+      if (type && type.__pw_type === 'fragment')
```
dgozman, line 43: I don't understand how is this supposed to work. `isJsxComponent(v)` checks for `__pw_type === 'jsx'` - how could it be equal to `fragment` here? What am I missing? Perhaps, having a comment with examples of various values we expect in this function would be beneficial.

Skn0tt: Changed the tag, updated the description and added a comment.

packages/playwright-ct-react/registerSource.mjs
```diff
@@ -33,12 +33,16 @@ function isJsxComponent(component) {
...
 /**
+ * Turns the Playwright representation of JSX (see jsx-runtime.js) into React.createElement calls.
  * @param {any} value
  */
 function __pwRender(value) {
   return window.__pwTransformObject(value, v => {
     if (isJsxComponent(v)) {
       const component = v;
+      let type = component.type;
+      if (type && type.__pw_jsx_fragment)
```
dgozman, line 44: - Could it be that `type` is not an object, and this code will throw?
- Let's make a function similar to `isJsxComponent` and a corresponding type declaration similar to `JsxComponent` that we have above. It could also check the `typeof`.

dgozman (commented):
Thank you for the comments, this is now much clearer!

packages/playwright-ct-react/registerSource.mjs
Skn0tt: done in https://github.com/microsoft/playwright/pull/32900/commits/bb1ae4c4ece2316a9e4f0c448021c57d654544e7

dgozman (approved):

Issue #32807, opened 2024-09-25
[Feature]: Pause function on Clock API


edmorrish:
### 🚀 Feature Request

A `pause` function on the [Clock API](https://playwright.dev/docs/api/class-clock) that freezes the `Date` and all timers without jumping forwards in time.

### Example

For example, when testing that a behaviour occurs exactly 5s after a key event.
  
```
await page.clock.install()
// Carry out some activity that relies on unpaused timers
await page.clock.pause()

await dispatchKeyEvent()

await page.clock.runFor(4999)

await expect(behaviourNotToHaveOccurred)

await page.clock.runFor(1)

await expect(behaviourToHaveOccurred)
```

### Motivation

Currently to pause the Clock API we need to use the `pauseAt` API. This can be cumbersome because we may not be concerned with the exact value of the Date, and want to pause instantly rather than at some point in the future.

yury-s:
Is there a reason you cannot do `await page.clock.pauseAt(new Date())` ?

edmorrish:
I'd thought about doing that but I'd assumed that that would be using the host time, and not the time inside the chrome environment. In some of our tests we've set the system time to a time different to the host time, then done some work without pausing timers, then would like to pause the timers when we've entered a certain part of our app, when we don't know the current system time that's being reported by the clock API `Date` mocks.

e.g. I would expect the following
```
await page.clock.setSystemTime(new Date('1970-01-01T00:00:00'))

// Do some initial work with running timers and elapsing time to get into the correct state of the app

await page.clock.pauseAt(new Date())
```
to jump me to today's date and then pause, which is not necessarily what I'd like, but I don't know how much time has elapsed since I set the system time.

Thanks very much for getting back to me, I'll give it a go and see if it works the way I expect or solved the problem.

yury-s:
> to jump me to today's date and then pause, which is not necessarily what I'd like, but I don't know how much time has elapsed since I set the system time.

You test is in full control of what the time is in the page after it called `setSystemTime`, so you can keep track of it in the test and then call `pauseAt` with the right value.

edmorrish:
Oh interesting, I think I'd misunderstood the difference between `setFixedTime` and `setSystemTime`. Does `setSystemTime` also pause the timers? Or is the suggestion here to do something like:

```
const initialRealTime = new Date()
const initialTestTime = new Date('1970-01-01T00:00:00')
await page.clock.setSystemTime(initialTestTime)

// Do initial work

[...]

PR #32901, opened 2024-10-01
fix(clock): amend setSystemTime docs

Skn0tt:
As discussed yesterday over https://github.com/microsoft/playwright/issues/32807. Adds some words to differentiate `setSystemTime` from `setFixedTime`.

### DIFF
--- a/docs/src/api/class-clock.md
+++ b/docs/src/api/class-clock.md
@@ -251,7 +251,7 @@ Time to be set.
 
 Sets system time, but does not trigger any timers. Use this to test how the web page reacts to a time shift, for example switching from summer to winter time, or changing time zones.
 
-Don't use together with [`method: Clock.install`]. Read docs on [clock emulation](../clock.md) to learn more.
+Use this method for simple scenarios where you only need to test with a predefined time. For more advanced scenarios, use [`method: Clock.install`] instead. Read docs on [clock emulation](../clock.md) to learn more.
 
 **Usage**
 
--- a/docs/src/api/class-clock.md
+++ b/docs/src/api/class-clock.md
@@ -249,7 +249,9 @@ Time to be set.
 ## async method: Clock.setSystemTime
 * since: v1.45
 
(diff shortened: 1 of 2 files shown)


docs/src/api/class-clock.md
```diff
@@ -249,7 +249,7 @@ Time to be set.
 ## async method: Clock.setSystemTime
 * since: v1.45
 
-Sets current system time but does not trigger any timers.
+Sets current system time without pausing, but does not trigger any timers. You most likely don't need this.
```
dgozman, line 252: - I don't think this is helpful. We better explain a usecase here, e.g. "testing how the web page reacts to a time shift, for example switching from summer to winter time, or changing a time zone". Also, it works when paused as well.
- Can we update `setFixedTime` docs to say you should either use it, or `install()`.
- Links to the clock guide would probably be helpful here as well.

Skn0tt: Thanks for all the feedback. I've pushed a second attempt, how do you like that?

docs/src/api/class-clock.md
```diff
@@ -249,7 +249,9 @@ Time to be set.
 ## async method: Clock.setSystemTime
 * since: v1.45
 
-Sets current system time but does not trigger any timers.
+Sets system time, but does not trigger any timers. Use this to test how the web page reacts to a time shift, for example switching from summer to winter time, or changing time zones.
+
+Don't use together with [`method: Clock.install`]. Read docs on [clock emulation](../clock.md) to learn more.
```
dgozman, line 254: - This should be a part of `setFixedTime`, not `setSystemTime`.

```suggestion
Use this method for simple scenarios where you only need to test with a predefined time. For more advanced scenarios, use [`method: Clock.install`] instead. Read docs on [clock emulation](../clock.md) to learn more.
```

Skn0tt: moved it in https://github.com/microsoft/playwright/pull/32901/commits/161070bc1b2b9c818c5f42999651dff1b54b682b

dgozman (approved):

PR #32907, opened 2024-10-01
fix(UI Mode): Update the Source section doc

meeroslaph:
Using UI Mode in the current v1.47.2 it's required to actually click the Action to show the correspondent line of code in the Source panel.

Hovering the Action would update content in the DOM snapshot, Call and Log tabs, but not the Source.

Thus, not sure if it's a typo in the documentation or a bug.

### DIFF
--- a/docs/src/test-ui-mode-js.md
+++ b/docs/src/test-ui-mode-js.md
@@ -60,7 +60,7 @@ Click on the pick locator button and hover over the DOM snapshot to see the loca
 
 ## Source
 
-When you click on an action in your test, the corresponding line of code is highlighted in the Source panel.
+When you click on an action in your test, the corresponding line of code is highlighted in the Source panel. If you are using VS Code, you can easily open your test by clicking the Open in VS Code icon on the right from the file name in the Source tab. This will open your test in VS Code right at the highlighted line of code.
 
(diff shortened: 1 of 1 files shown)


Skn0tt:
This sounds more like a bug. Could you open a bug report about this instead?

docs/src/test-ui-mode-js.md
```diff
@@ -108,7 +108,7 @@ Next to the Actions tab you will find the Metadata tab which will show you more
 
 ## Watch mode
 
-Next to the name of each test in the sidebar you will find an eye icon. Clicking on the icon will activate watch mode which will re-run the test when you make changes to it. You can watch a number of tests at the same time be clicking the eye icon next to each one or all tests by clicking the eye icon at the top of the sidebar. If you are using VS Code then you can easily open your test by clicking on the file icon next to the eye icon. This will open your test in VS Code right at the line of code that you clicked on.
```
meeroslaph, line 111: @Skn0tt, what do you think about this part?

The button has been moved to the Source tab, so I reflected the change [here](https://github.com/microsoft/playwright/pull/32907/files#diff-cd5a9688c278134c1b30e9e5db3afbbac143be7f95c531a14551ef04f7748673R63).

I can also open a separate MR for that if you believe it's not a bug.

meeroslaph:
> This sounds more like a bug. Could you open a bug report about this instead?

https://github.com/microsoft/playwright/issues/32915

PR #32908, opened 2024-10-01
chore(bidi): enable logs in Firefox on CI

yury-s:


mxschmitt (approved):

whimboo:
@yury-s thanks for the PR. I had a look at the log but it is huge (~400MB plain text). Given that amount of tests that you are running in CI I don't think it makes sense to keep the trace preference enabled. I would always have to download the log and cannot check directly in the GitHub Ci.

Would you mind backing it out (or parts for the pref usage) again? Thanks!

yury-s:
@whimboo thanks for taking a look, I'll disable the logging for now. If it helps, we can be enabling the logging for particular test files on CI via fixtures, so that the output is produced only for those tests or we can even move some tests into a separate GHA job, so that only those are run with certain settings and complete faster with less output.

whimboo:
> we can even move some tests into a separate GHA job, so that only those are run with certain settings and complete faster with less output.

@yury-s I think that would be great. At least for those tests which currently show a timeout error that I cannot explain. I'm happy to check those logs if they were available. Thanks.

PR #32911, opened 2024-10-01
fix(fetch): fallback to given URL if baseURL is invalid

mxschmitt:
Fetch baseURL handling should be identical to what we do with `page.goto`. I saw a bug report on Discord which worked with `page.goto` but didn't with `request.get`.


yury-s (approved):
Not sure if masking baseURL problem is a good idea, but we should be consistent at least

PR #32919, opened 2024-10-02
fix(trace viewer): reveal stack for highlighted action

### LINKED ISSUE #32915: [Bug]: The current line of code in the Source tab is not getting updated when hovering different actions
### Version

1.47.2

### Steps to reproduce

1. Run a test in UI Mode.
2. Open the Source tab.
3. Move through different actions by hovering them in the Actions tab.

### Expected behavior

The currently selected line of code gets updated in the Source tab as a user moves through the actions in the Actions tab (according to the [documentation](https://playwright.dev/docs/test-ui-mode#source)).

### Actual behavior

The currently selected line of code isn't updated in the Source tab as a user moves through the actions in the Actions tab. The user must actually click on the action to get the line of code in the Source tab updated. However, the content in the Call and the Log tabs is properly updated when hovering actions.

### Additional context

Initially I thought it might be the issue with documentation being outdated ([MR](https://github.com/microsoft/playwright/pull/32907)), but @Skn0tt [suggested](https://github.com/microsoft/playwright/pull/32907#issuecomment-2387839280) submitting a bug report instead.

### Environment

```shell
System:
    OS: macOS 15.0
    CPU: (10) arm64 Apple M1 Pro
    Memory: 50.83 MB / 16.00 GB
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.17.0/bin/npm
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.47.2 => 1.47.2
```

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32915.

In the `Call` and `Logs` tabs, we update the contents based on the hovered action. We document that this is also the case for the `Source` tab:

https://github.com/meeroslaph/playwright/blob/78054a7652377a316669e21163ede3d8dbc0d99c/docs/src/test-ui-mode-js.md?plain=1#L61-L65

But it isn't. Not sure if it's a regression or not, but this PR fixes it.

### DIFF
--- a/packages/trace-viewer/src/ui/workbench.tsx
+++ b/packages/trace-viewer/src/ui/workbench.tsx
@@ -114,15 +114,15 @@ export const Workbench: React.FunctionComponent<{
     }
   }, [model, selectedCallId]);
 
-  const revealedStack = React.useMemo(() => {
-    if (revealedError)
-      return revealedError.stack;
-    return selectedAction?.stack;
-  }, [selectedAction, revealedError]);
-
   const activeAction = React.useMemo(() => {
     return highlightedAction || selectedAction;
   }, [selectedAction, highlightedAction]);
+
+  const revealedStack = React.useMemo(() => {
+    if (revealedError)
+      return revealedError.stack;
+    return activeAction?.stack;
+  }, [activeAction, revealedError]);
 
   const onActionSelected = React.useCallback((action: modelUtil.ActionTraceEventInContext) => {
     setSelectedAction(action);
--- a/tests/config/traceViewerFixtures.ts
+++ b/tests/config/traceViewerFixtures.ts
@@ -77,6 +77,11 @@ class TraceViewerPage {
   }
 
   @step
+  async hoverAction(title: string, ordinal: number = 0) {
+    await this.page.locator(`.action-title:has-text("${title}")`).nth(ordinal).hover();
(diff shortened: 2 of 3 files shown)


dgozman (commented):
This could have been done on purpose, to avoid source tab flashing too much. Let's do some code archeology to figure this out!

Skn0tt:
Alright, i'll bisect around a little.

Skn0tt:
https://github.com/microsoft/playwright/pull/29268 broke it. Before, the Sources tab was passed the `activeAction` that respects highlights. After, it's passed the `revealedStack`, which isn't updated when the highlight changes. We didn't have any tests for it and I don't see any discussion about it in the PR, so i'm assuming it went unnoticed.

dgozman:
> #29268 broke it. Before, the Sources tab was passed the `activeAction` that respects highlights. After, it's passed the `revealedStack`, which doesn't respect highlights because it's not updated when the highlight changes. We didn't have any tests for it and I don't see any discussion about it in the PR, so i'm assuming it wasn't on purpose.

Excellent! Thank you for the analysis.

dgozman (approved):

PR #32930, opened 2024-10-02
feat(base-reporter): Add tags to test output 

mrkstwrt:
If you switch to the [new tag syntax](https://playwright.dev/docs/test-annotations#tag-tests) added in [v1.42.0](https://github.com/microsoft/playwright/releases/tag/v1.42.0) you can no longer see tags in test logs without duplicating the tags in the test name, this PR adds the `appendTags` option to the list reporter which enables adding the tags to the end of the test name in the log.

### DIFF
--- a/docs/src/test-reporters-js.md
+++ b/docs/src/test-reporters-js.md
@@ -102,6 +102,7 @@ List report supports the following configuration options and environment variabl
 | Environment Variable Name | Reporter Config Option| Description | Default
 |---|---|---|---|
 | `PLAYWRIGHT_LIST_PRINT_STEPS` | `printSteps` | Whether to print each step on its own line. | `false`
+| `PLAYWRIGHT_LIST_APPEND_TAGS` | `appendTags` | Whether to append tags to the end of the test name. | `false`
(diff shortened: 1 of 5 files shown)


mrkstwrt:

Skn0tt:
Hi! Could you open a feature request about this? It's always good for us to first understand & discuss the usecase before we add functionality.

mrkstwrt:
> Hi! Could you open a feature request about this? It's always good for us to first understand & discuss the usecase before we add functionality.

Sure, I've added https://github.com/microsoft/playwright/issues/32966

pavelfeldman:
Thanks for the patch and the issue. Your change makes sense to me. I would include tags unconditionally in base.ts so that all the reporters that use base.ts would benefit from the change. I would also not wrap it with [] for now so that it was of the form

[project] › file.spec.ts:1:1 › should work @smoke

mrkstwrt:
> Thanks for the patch and the issue. Your change makes sense to me. I would include tags unconditionally in base.ts so that all the reporters that use base.ts would benefit from the change. I would also not wrap it with [] for now so that it was of the form
> 
> [project] › file.spec.ts:1:1 › should work @smoke

Thanks for the feedback @pavelfeldman, I've made the change you suggested to always render the tags in base

pavelfeldman (approved):

Issue #32925, opened 2024-10-02
[Feature]: Dynamic tags in runtime


JanLechvar:
### 🚀 Feature Request

Implementation of Dynamic tags in runtime.
Allow users to push tags to the reporter when a condition described in the test is met.

### Example

  if (errorLogs.length > 0) {
   testInfo.annotations.push({
      type: "WARNING - Console Errors detected",
      description: "Check for console errors in the test",
    });
      testInfo.tags.push("@console-error");
  }

### Motivation

Imagine a test suite with over 100 tests. 
In each test, user wants to check if any errors appear in the browser console during the test execution without affecting the test results. If an error occurs in any test, user pushes this error, say in the form of an annotation, to the Playwright reporter.

The person reviewing the report after all tests have been executed has to expand each test result individually to see if a console error occurred during that specific test.

Dynamic tags in runtime would solve this problem, and the person reviewing the report wouldn’t have to open each test result because tests with console errors would be tagged.

This would save time and make the report more organized, as it would be immediately clear which tests to review for console errors thanks to the runtime tags.

Skn0tt:
Dynamic tags wouldn't work with `--grep "@tag"`, so I think that'd add undue confusion. As you explain, it's already possible to dynamically push an annotation. Reporters are free to display these annotations however they like. If you're using the HTML reporter, we could look into making annotations searchable. What reporter are you using?

JanLechvar:
Correct, such a tag wouldn't work with the --grep "@tag",
but I have couple of ideas, ordered according to my personal preferences:

1. Besides annotations and tags, introduce a third method, such as Note/Sticker/Badge which will behave similarly to tags in terms of display in the report but can be added dynamically to the report.

2. Annotation ‘Type’ can be displayed similarly to a tag (ideally, user can choose which type/annotation to display like a tag). For example:
{
    "annotation": {
        "type": "This-is-annotation-type",
        "description": "This is description",
        "displayTypeAsTag": true,
    }
}

![001 2024-10-02 15 24 47](https://github.com/user-attachments/assets/adbccc29-6e99-4f14-a4f7-310c2cc081e4)

3. Introduction of a filter for annotations, as you mentioned.

I am using the HTML reporter.

Skn0tt:
> Introduction of a filter for annotations, as you mentioned.

[...]

PR #32931, opened 2024-10-02
test(html reporter): filtering by type works

Skn0tt:
The folks who opened https://github.com/microsoft/playwright/issues/32925 would benefit from filtering by annotation existence. Turns out we already have it! This PR adds a test to ensure it stays that way.


dgozman (approved):
Note that I've updated the title to align with the common practices.

PR #32945, opened 2024-10-03
chore: lint java docs snippets

mxschmitt:
Motivation: Validate the syntax like we do in the other languages. As of today we didn't validate Java snippets.


docs/src/mock.md
```diff
@@ -195,15 +195,16 @@ await Expect(page.GetByTextAsync("Loquat", new () { Exact = true })).ToBeVisible
 page.route("*/**/api/v1/fruits", route -> {
   Response response = route.fetch();
   byte[] json = response.body();
-  parsed = new Gson().fromJson(json, JsonObject.class)
-  parsed.add(new JsonObject().add("name", "Loquat").add("id", 100));
+  JsonObject parsed = new Gson().fromJson(new String(json), JsonObject.class);
+  parsed.add("name", new JsonPrimitive("Loquat"));
```
dgozman, line 199: This is not equivalent, it should add a single object with two properties, not two properties to an array.

dgozman (approved):

PR #32955, opened 2024-10-04
feat(test runner): allow multiple global setups

### LINKED ISSUE #32369: [Feature]: Allow multiple global setup and teardown files in the playwright configuration
### 🚀 Feature Request

Allow globalSetup and globalTeardown properties to take in multiple files as input, thus customers can have multiple global setup and teardown files specified in their playwright configuration. 

### Example

```ts

export default defineConfig({
  globalSetup: [
    './database/global-setup',
    './base/global-setup'
  ],
  globalTeardown: [
   './database/global-teardown',
   './base/global-teardown'
  ]
});
```

### Motivation

1. A single function for setup and teardown makes it difficult for users to maintain setup/teardown files, where they need to communicate with a lot of services to setup the testing environment - adding mock data to database, signing up fake users for testing, bypassing auth, etc. Multiple files will keep this implementation clean.
2. In a large project, with multiple playwright test suites, we would like to have a single base playwright config (with global setup and teardown) that will be inherited by all the test suites. Each test suite, can also define their own global setup and teardowns. Currently, it gets overriden and each test suite needs to modify their own setup/teardown files to accomodate the base setup/teardown files.
Ideally, we would like to simply add the base setup/teardown files to the setup/teardown array in each configuration.

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32369

The idea behind ordering is that for pairs of setup/teardown, each pair is only torn down after all pairs executed after it are torn down. This allows "stacking" these setup steps onto another, where one setup depends on the previous setup.

For  `globalSetup: ['s1', 's2'], globalTeardown: ['t1', 't2']`, this results in order `s1, s2 ; t2, t1`.

There's a second API some people use where setup steps return their respective teardown steps. If that API is used exclusively, the same ordering applies. Mixing both APIs is discouraged, as that gives hard-to-reason-about ordering.


tests/playwright-test/global-setup.spec.ts
```diff
@@ -386,3 +386,67 @@ test('teardown after error', async ({ runInlineTest }) => {
...
+    'globalSetup2',
+    'globalSetup3',
+    'test a',
+    'test b',
+    'globalTeardown2',
+    'globalTeardown1',
+  ]);
+});
+
+test('debug logs show index of setup/teardown', async ({ runInlineTest }) => {
```
dgozman, line 431: I'd remove this test, it does not test an essential part of our test runner.

Instead, it would be great to test what exactly do we run in what order, when there are multiple global setups and teardowns, including teardown functions returned from global setups, and some of them fail.

tests/playwright-test/global-setup.spec.ts
```diff
@@ -386,3 +386,67 @@ test('teardown after error', async ({ runInlineTest }) => {
...
+
+    'a.test.js': `
+      import { test } from '@playwright/test';
+      test('a', () => console.log('test a'));
+      test('b', () => console.log('test b'));
+    `,
+  }, { reporter: 'line' });
+
+  expect(passed).toBe(2);
+  expectInOrder(output, [
```
dgozman, line 420: Look at how other tests use `result.outputLines`.

packages/playwright/src/runner/tasks.ts
```diff
@@ -161,15 +164,20 @@ function createPluginBeginTask(plugin: TestRunnerPluginRegistration): Task<TestR
   };
 }
 
-function createGlobalSetupTask(): Task<TestRun> {
+function createGlobalSetupTask(index: number, length: number): Task<TestRun> {
```
dgozman, line 167: I find the proposed behavior reasonable, so not really sure I understand the TODO in the description.

tests/playwright-test/global-setup.spec.ts
Skn0tt: I've removed the test in https://github.com/microsoft/playwright/pull/32955/commits/4316a0178dcf0f70afc700ec1fe8a4ea620f2e73 and and added lots of coverage in https://github.com/microsoft/playwright/pull/32955/commits/3ef8178cca9b70d4b9df39f4374e663750d8ae91. Also amended the PR description with my thoughts around ordering.

tests/playwright-test/global-setup.spec.ts
Skn0tt: Done in https://github.com/microsoft/playwright/pull/32955/commits/38b7dee4423f24645a1e43e88a4aac3f7217834e

Skn0tt:
Oof, my change swaps the ordering of existing `globalTeardown` files and setup-returned teardown functions. Looks like i'll need to revert that.

Skn0tt:
Discussed this with Dima in a call. Reverting the ordering change so we don't break anything. This makes the ordering harder to reason about when folks mix both APIs, but that's discouraged anyways.

packages/playwright/src/common/config.ts
```diff
@@ -70,13 +73,16 @@ export class FullConfigInternal {
...
+    this.globalTeardowns = (Array.isArray(userConfig.globalTeardown) ? userConfig.globalTeardown : [userConfig.globalTeardown]).map(s => resolveScript(s, configDir)).filter((script): script is string => !!script);
+
     this.config = {
       configFile: resolvedConfigFile,
       rootDir: pathResolve(configDir, userConfig.testDir) || configDir,
       forbidOnly: takeFirst(configCLIOverrides.forbidOnly, userConfig.forbidOnly, false),
       fullyParallel: takeFirst(configCLIOverrides.fullyParallel, userConfig.fullyParallel, false),
-      globalSetup: takeFirst(resolveScript(userConfig.globalSetup, configDir), null),
-      globalTeardown: takeFirst(resolveScript(userConfig.globalTeardown, configDir), null),
+      globalSetup: takeFirst(...this.globalSetups, null),
```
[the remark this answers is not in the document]
Skn0tt: Would you prefer this?
```suggestion
      globalSetup: this.globalSetups[0] ?? null,
```

docs/src/test-api/class-testconfig.md
```diff
@@ -110,9 +110,9 @@ export default defineConfig({
 
 ## property: TestConfig.globalSetup
 * since: v1.10
-- type: ?<[string]>
+- type: ?<[string]|[Array]<[string]>>
 
-Path to the global setup file. This file will be required and run before all the tests. It must export a single function that takes a [FullConfig] argument.
+Path to the global setup file. This file will be required and run before all the tests. It must export a single function that takes a [FullConfig] argument. Pass an array for multiple global setup files.
```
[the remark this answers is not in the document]
pavelfeldman: You are fine. It felt like a couple of words are missing. "Pass an array of file names to specify multiple global setup files"

packages/playwright/src/common/config.ts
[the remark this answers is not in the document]
pavelfeldman: nit means "while you are here if you also find it weird, replace it". Your suggestion works.

packages/playwright/src/runner/tasks.ts
pavelfeldman: I'd like to understand what the semantics here is. Let's say I have a database and an auth token. My database requires setup and teardown, my auth token only requires setup. Would there be a difference between the following configurations:

```js
globalSetup: [dbSetup.ts, authSetup.ts],
globalTeardown: dbTeardown.ts,
```
and

```
globalSetup: [authSetup.ts, dbSetup.ts],
globalTeardown: dbTeardown.ts,
```

This `index` parity confuses me and makes me think there is a difference.

Skn0tt: The difference is in execution ordering. First executes `dbSetup, authSetup ; dbTeardown`, second executes `authSetup, dbSetup ; dbTeardown`. In your usecase, this probably doesn't make a difference.

docs/src/test-api/class-testconfig.md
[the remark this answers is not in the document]
Skn0tt: good idea! added the words in https://github.com/microsoft/playwright/pull/32955/commits/c8c483cacbfe313bbe2b88283e65c95a065fcbb7.

tests/playwright-test/global-setup.spec.ts
```diff
@@ -386,3 +386,43 @@ test('teardown after error', async ({ runInlineTest }) => {
...
+        globalSetup: ['./globalSetup1.ts','./globalSetup2.ts','./globalSetup3.ts','./globalSetup4.ts'],
+        globalTeardown: ['./globalTeardown1.ts', './globalTeardown2.ts'],
+      };
+    `,
+    'globalSetup1.ts': `module.exports = () => { console.log('%%globalSetup1'); return () => { console.log('%%globalSetup1Function'); throw new Error('kaboom'); } };`,
+    'globalSetup2.ts': `module.exports = () => console.log('%%globalSetup2');`,
+    'globalSetup3.ts': `module.exports = () => { console.log('%%globalSetup3'); return () => console.log('%%globalSetup3Function'); }`,
+    'globalSetup4.ts': `module.exports = () => console.log('%%globalSetup4');`,
+    'globalTeardown1.ts': `module.exports = () => console.log('%%globalTeardown1')`,
+    'globalTeardown2.ts': `module.exports = () => console.log('%%globalTeardown2');`,
```
dgozman, line 403: I'd also throw here, and check that it did not prevent globalTeardown1 from running.

dgozman (approved):

tests/playwright-test/global-setup.spec.ts
Skn0tt: Did that in https://github.com/microsoft/playwright/pull/32955/commits/a72ec6e8245a1725995400429a4ed26481c7cc9b. This behaviour is surprising to me, was somehow expecting a failing `teardown` to stop further teardowns from being executed.

With that in mind, I don't think that a teardown function returned from `globalSetup` should behave any differently than a `globalTeardown` when it throws an error. I've made sure that a failing function doesn't block any `globalTeardown` script from being executed in https://github.com/microsoft/playwright/pull/32955/commits/9b785662f6e04f13f183985cf1f3b341e3480db3. Would be lovely if you could give that another look.

I think this could benefit from another refactoring where we pull `globalSetup` and `globalTeardown` scripts into separate tasks, but i'd like to get this PR merged first and then refactor in a follow-up.

Skn0tt: On a second note, this could be considered a breaking change. If folks are currently relying on the `globalSetup` result throwing an error to prevent `globalTeardown` from being executed, that's broken with https://github.com/microsoft/playwright/commit/9b785662f6e04f13f183985cf1f3b341e3480db3. So maybe we should just stick with the incongruence.

dgozman: I agree, let's not touch the relative behavior of `globalSetup` returned function vs `globalTeardown`. The aim here is to support multiple, and our general rule is that all teardowns run even if one of them fails. This applies to fixtures, afterEach and afterAll hooks, and now to global teardown as well.

PR #32988, opened 2024-10-07
fix(socks): downgrade socks-proxy-agent to fix two failing tests on main

Skn0tt:
`"should pass with matching certificates when a socks proxy is used"` was broken by https://github.com/microsoft/playwright/commit/de4a4d1ce15dd4c4c0ab3a43cc8775e10e3f514b. This PR fixes it by downgrading `socks-proxy-agent` to the version it was at before - I added that upgrade to the PR to keep it up-to-date, but we only needed to really update https-proxy-agent.

### DIFF
--- a/packages/playwright-core/bundles/utils/package.json
+++ b/packages/playwright-core/bundles/utils/package.json
@@ -24,7 +24,7 @@
     "proxy-from-env": "1.1.0",
     "retry": "0.12.0",
     "signal-exit": "3.0.7",
-    "socks-proxy-agent": "8.0.4",
+    "socks-proxy-agent": "6.1.1",
     "stack-utils": "2.0.5",
     "ws": "8.17.1"
   },
--- a/packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
+++ b/packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
@@ -28,6 +28,7 @@ import { debugLogger } from '../utils/debugLogger';
 import { createProxyAgent } from './fetch';
 import { EventEmitter } from 'events';
 import { verifyClientCertificates } from './browserContext';
+import { SocksProxyAgent } from '../utilsBundle';
 
 let dummyServerTlsOptions: tls.TlsOptions | undefined = undefined;
 function loadDummyServerCertsIfNeeded() {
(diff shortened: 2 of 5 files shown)


packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
```diff
@@ -97,8 +98,8 @@ class SocksProxyConnection {
   }
 
   async connect() {
-    if (this.socksProxy.proxyAgentFromOptions)
-      this.target = await this.socksProxy.proxyAgentFromOptions.connect(new EventEmitter() as any, { host: rewriteToLocalhostIfNeeded(this.host), port: this.port, secureEndpoint: false });
+    if (this.socksProxy.proxyAgentFromOptions instanceof SocksProxyAgent)
```
mxschmitt, line 101: Looks wrong, it can be also a HttpsProxyAgent.

Skn0tt: In a file called `socksClientCertificatesInterceptor`?

packages/playwright-core/bundles/utils/package.json
```diff
@@ -24,7 +24,7 @@
     "proxy-from-env": "1.1.0",
     "retry": "0.12.0",
     "signal-exit": "3.0.7",
-    "socks-proxy-agent": "8.0.4",
+    "socks-proxy-agent": "6.1.1",
```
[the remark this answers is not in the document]
Skn0tt: We need the up-to-date version for the fix from https://github.com/microsoft/playwright/commit/de4a4d1ce15dd4c4c0ab3a43cc8775e10e3f514b though. If we want to keep them in sync, we need to revert it until I figure out how we can update socks-proxy-agent.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
mxschmitt: Yes.

Skn0tt:
Alright, reverting in https://github.com/microsoft/playwright/pull/32989.

packages/playwright-core/src/server/socksClientCertificatesInterceptor.ts
mxschmitt: Background: Only the communication between the browser and Playwright is made via a Socks Proxy, hence its called like that. Internally when we make requests to the target (e.g. internal.corp.com it can be either direct or via an agent. An agent is a proxy agent which then can be https or socks.). (We don't intercept http traffic.)

PR #32989, opened 2024-10-07
Revert "fix(har timing): record `connect` timing for proxied connections"

Skn0tt:
Reverts microsoft/playwright#32855. This broke two tests on main, and we don't yet know how to fix it other than downgrading.


mxschmitt (approved):

mxschmitt:
There might be a trivial way of fixing the tests tho.  I'd try keep digging / we can try to reland it again alter on. If you need help we can also take a look together.

PR #32990, opened 2024-10-07
Add Devuan OS fallback to Debian

asherwoodcnet:
Hello! I'd like to be able to run `npx playwright install` on Devuan machines, so I've added a fallback to match an appropriate Debian version. I'm happy to open an issue, if that would be more appropriate.

Thank you!

### DIFF
--- a/packages/playwright-core/src/utils/hostPlatform.ts
+++ b/packages/playwright-core/src/utils/hostPlatform.ts
@@ -86,15 +86,20 @@ function calculatePlatform(): { hostPlatform: HostPlatform, isOfficiallySupporte
         return { hostPlatform: ('ubuntu22.04' + archSuffix) as HostPlatform, isOfficiallySupportedPlatform: false };
       return { hostPlatform: ('ubuntu24.04' + archSuffix) as HostPlatform, isOfficiallySupportedPlatform: false };
     }
-    if (distroInfo?.id === 'debian' || distroInfo?.id === 'raspbian') {
+    if (distroInfo?.id === 'debian' || distroInfo?.id === 'raspbian' || distroInfo?.id === 'devuan') {
       const isOfficiallySupportedPlatform = distroInfo?.id === 'debian';
-      if (distroInfo?.version === '11')
+      let debianVersion = distroInfo?.version;
+      if (distroInfo.id === 'devuan') {
+        // Devuan is debian-based but it's always 7 versions behind
+        debianVersion = String(parseInt(distroInfo.version, 10) + 7);
+      }
+      if (debianVersion === '11')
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):

yury-s (approved):

packages/playwright-core/src/utils/hostPlatform.ts
```diff
@@ -86,15 +86,20 @@ function calculatePlatform(): { hostPlatform: HostPlatform, isOfficiallySupporte
         return { hostPlatform: ('ubuntu22.04' + archSuffix) as HostPlatform, isOfficiallySupportedPlatform: false };
       return { hostPlatform: ('ubuntu24.04' + archSuffix) as HostPlatform, isOfficiallySupportedPlatform: false };
     }
-    if (distroInfo?.id === 'debian' || distroInfo?.id === 'raspbian') {
+    if (distroInfo?.id === 'debian' || distroInfo?.id === 'raspbian' || distroInfo?.id === 'devuan') {
       const isOfficiallySupportedPlatform = distroInfo?.id === 'debian';
-      if (distroInfo?.version === '11')
+      let debianVersion = distroInfo?.version;
+      if (distroInfo.id === 'devuan') {
+        // Devuan is debian-based but it's always 7 versions behind
```
mxschmitt, line 93: Reference: https://en.wikipedia.org/wiki/Devuan#:~:text=%5B22%5D-,Version%20history,-%5Bedit%5D

1ma:
I'm also interested in running Playwright on Devuan, but I'm still hitting errors on the latest release.

Maybe I'm dense, but it looks like there's been 3 releases since this PR was merged (1.48.0, 1.48.1 and 1.48.2) but they didn't pick the changes from this PR:

hostPlatform.ts on main branch contains the "devuan" string: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/hostPlatform.ts

However, each of the 3 releases since 8 Oct do not contain any "devuan" string in hostPlatform.ts:

https://github.com/microsoft/playwright/blob/v1.48.0/packages/playwright-core/src/utils/hostPlatform.ts
https://github.com/microsoft/playwright/blob/v1.48.1/packages/playwright-core/src/utils/hostPlatform.ts
https://github.com/microsoft/playwright/blob/v1.48.2/packages/playwright-core/src/utils/hostPlatform.ts

cc @mxschmitt @yury-s

asherwoodcnet:
> Maybe I'm dense, but it looks like there's been 3 releases since this PR was merged (1.48.0, 1.48.1 and 1.48.2) but neither of these included the changes from this PR.

does anybody have any advice as to why these changes weren't published, and why they were reverted? I must be missing something about how `hostPlatform.ts` is published

Issue #32341, opened 2024-08-27
[Bug]: attachments are wrong/confusing when `expectScreenshot` fails on time limit


urugator:
### Version

1.42.1

### Steps to reproduce

Create test, where `expectScreenshot` fails, because it's unable to capture **stable** snapshot within a time limit.
Attachments will contain 3 screenshots: **actual** (last screenshot taken), **expected** (golden snapshot) and **diff**.
Normally, if `expectScreenshot` fails on comparison, it creates **diff** between **actual** and **expected**.
However, if `expectScreenshot` fails because it's unable to capture same consecutive images (stable snapshot) within a time limit, it produces a **diff** between the **actual** and the **previous** (attempt):
https://github.com/microsoft/playwright/blob/3f085d568935b9a4422f689185f720c1206bac84/packages/playwright-core/src/server/page.ts#L644
The problem is that **previous** isn't provided as an attachment and therefore you see **diff** that is completely unrelated to the other two screenshots.
Here is the caller, which ignores `previous` and assumes the returned `diff` is always between `actual` and `expected`
https://github.com/microsoft/playwright/blob/3f085d568935b9a4422f689185f720c1206bac84/packages/playwright/src/matchers/toMatchSnapshot.ts#L426-L439

Sorry for not providing an actual test.

### Expected behavior

**expected** attachment should be the **previous** screenshot or attachments should include an extra **previous** screenshot.

### Actual behavior

**Diff** image is not a diff between **actual** and **expected** images when `expectScreenshot` is unable to capture stable snapshot within a time limit.

### Environment

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (4) x64 Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz
    Memory: 1.18 GB / 7.98 GB
  Binaries:
    Node: 20.13.1 - C:\Program Files\nodejs\node.EXE     
    npm: 10.5.2 - C:\Program Files\nodejs\npm.CMD        
    pnpm: 9.1.3 - C:\Program Files\nodejs\pnpm.CMD
  IDEs:
    VSCode: 1.92.2 - C:\Users\urug\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
  Languages:
    Bash: 5.0.17 - C:\WINDOWS\system32\bash.EXE
  npmPackages:
    @playwright/test: 1.42.1 => 1.42.1
```

GoldenCaterpie:
Download 
[...]

PR #32997, opened 2024-10-08
fix(screenshot): show image diff inline in errors list

yury-s:
The diff is now shown inline in the errors list.

There are 2 possible failures of toHaveScreenshot
* Previous and actual snapshot mismatch. In this case html report will show diff between Actual/Previous and have Expected as a separate screenshot.
* Actual/Previous are equal but they differ from the expected. In this case html report only contains Actual/Expected images and the diff.

Reference: https://github.com/microsoft/playwright/issues/32341

image

### DIFF
--- a/packages/html-reporter/src/testResultView.tsx
+++ b/packages/html-reporter/src/testResultView.tsx
@@ -98,9 +98,9 @@ export const TestResultView: React.FC<{
     {!!errors.length && <AutoChip header='Errors'>
       {errors.map((error, index) => {
         if (error.type === 'screenshot')
-          return <TestScreenshotErrorView errorPrefix={error.errorPrefix} diff={error.diff!} errorSuffix={error.errorSuffix}></TestScreenshotErrorView>;
-        return <TestErrorView key={'test-result-error-message-' + index} error={error.error!}></TestErrorView>
-        })}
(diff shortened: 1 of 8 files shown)


pavelfeldman (approved):

packages/html-reporter/src/testResultView.tsx
```diff
@@ -145,6 +150,44 @@ export const TestResultView: React.FC<{
   </div>;
 };
 
+function classifyErrors(testErrors: string[], diffs: ImageDiff[]) {
```
pavelfeldman, line 153: Would something like this read better?

```js
function classifyErrors(testErrors: string[], diffs: ImageDiff[]) {
  return testErrors.map(error => {
    if (error.includes('Screenshot comparison failed:')) {
      const matchingDiff = diffs.find(diff => {
        const attachmentName = diff.actual?.attachment.name;
        return attachmentName && error.includes(attachmentName);
      });

      if (matchingDiff) {
        const lines = error.split('\n');
        const index = lines.findIndex(line => /Expected:|Previous:|Received:/.test(line));
        const errorPrefix = index !== -1 ? lines.slice(0, index).join('\n') : lines[0];

        const diffIndex = lines.findIndex(line => / +Diff:/.test(line));
        const errorSuffix = diffIndex !== -1 ? lines.slice(diffIndex + 2).join('\n') : lines.slice(1).join('\n');

        return { type: 'screenshot', diff: matchingDiff, errorPrefix, errorSuffix };
      }
    }
    return { type: 'regular', error };
  });
}
```

pavelfeldman (approved):

PR #33003, opened 2024-10-08
Reapply "fix(har timing): record connect timing for proxied connections" (#32855)

Skn0tt:
This reapplies what we reverted in https://github.com/microsoft/playwright/pull/32989.

Max and me debugged this, and found that the test failures come from SOCKS proxy now preferring IPv6 over IPv4. We've updated the tests and made sure that this doesn't mask any breaking change.

I'm enabling CQ1 to make sure we don't oversee any other CI failures.

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -20,7 +20,6 @@ import http from 'http';
 import https from 'https';
 import type { Readable, TransformCallback } from 'stream';
 import { pipeline, Transform } from 'stream';
-import url from 'url';
 import zlib from 'zlib';
 import type { HTTPCredentials } from '../../types/types';
 import { TimeoutSettings } from '../common/timeoutSettings';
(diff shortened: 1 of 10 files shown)


packages/playwright-core/src/server/fetch.ts
```diff
@@ -515,11 +514,21 @@ export abstract class APIRequestContext extends SdkObject {
             }),
         );
 
+        // when using socks proxy, having the socket means the connection got established
+        if (agent instanceof SocksProxyAgent)
```
mxschmitt, line 518: this should be also the case for HttpsProxyAgent? So maybe we should check for just agent thruthyness there?

mxschmitt (approved):
LGTM % comment.

packages/playwright-core/src/server/fetch.ts
Skn0tt: HTTPSProxyAgent emits its own separate event that's more precise, so we use that: https://github.com/microsoft/playwright/pull/33003/files#diff-dc06386362e7da98cd82627a6b4f337ae01799a53cac4f208089b2b89c4dc832R527-R529

PR #33032, opened 2024-10-09
test(test-runner): test round-robin sharding with stable-test-runner

muhqu:
This PR contains the changes of https://github.com/microsoft/playwright/pull/30962 applied as patch to the stable-test-runner.

Here's a comparison of the run times for the tests in CI… 

image

The times above are from these test runs…
* before: https://github.com/microsoft/playwright/actions/runs/11259871632/job/31309848606?pr=33030
* after: https://github.com/microsoft/playwright/actions/runs/11260846783/job/31312968985?pr=33032

### DIFF
--- a/tests/playwright-test/playwright.config.ts
+++ b/tests/playwright-test/playwright.config.ts
@@ -36,6 +36,7 @@ export default defineConfig({
   forbidOnly: !!process.env.CI,
   workers: process.env.CI ? 2 : undefined,
   snapshotPathTemplate: '__screenshots__/{testFilePath}/{arg}{ext}',
+  shardingMode: 'round-robin',
   projects: [
     {
       name: 'playwright-test',
--- a/tests/playwright-test/stable-test-runner/package-lock.json
+++ b/tests/playwright-test/stable-test-runner/package-lock.json
@@ -4,8 +4,10 @@
   "requires": true,
   "packages": {
     "": {
+      "hasInstallScript": true,
       "dependencies": {
-        "@playwright/test": "1.48.0-beta-1728384960000"
+        "@playwright/test": "1.48.0-beta-1728384960000",
+        "patch-package": "8.0.0"
       }
     },
     "node_modules/@playwright/test": {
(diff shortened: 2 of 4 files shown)


pavelfeldman:
Do I understand it right that if the tests had beforeAll / worker fixture setups (as in real e2e tests), we would observe an opposite effect?

pavelfeldman:
I'm referring to this part:
```
+ *        [  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12]
+ * Shard 1:  ^---------^                                      : [  1, 2, 3 ]
+ * Shard 2:              ^---------^                          : [  4, 5, 6 ]
+ * Shard 3:                          ^---------^              : [  7, 8, 9 ]
+ * Shard 4:                                      ^---------^  : [ 10,11,12 ]
+ * ```
+ * Shards tests by round-robin.
+ *
+ * ```
+ *          [  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12]
+ * Shard 1:    ^               ^               ^              : [  1, 5, 9 ]
+ * Shard 2:        ^               ^               ^          : [  2, 6,10 ]
+ * Shard 3:            ^               ^               ^      : [  3, 7,11 ]
+ * Shard 4:                ^               ^               ^  : [  4, 8,12 ]
+ * ```
```

By default, we recommend using fullyParallel. This means that for the tests with non-trivial hooks we would do much more work in case of round-robin as we'll be running those hooks 3x more times. Making sure we are on the same page wrt the idea behind your change.

muhqu:
I think beforeAll kinda breaks fully-parallel as it causes tests in a spec file to be executed in the same test group, but in general you are right, round-robin can cause worker scoped hooks and fixtures to be executed more times than the current sharding algorithm. And setup / teardown times of worker scoped fixtures will probably not be included in test run times, so the duration-round-robin algorithm can not account for those yet… 

In our use-case, we are not making use of beforeAll, but instead have a heavy global setup. With global setup there is no difference between the sharding modes, as it's executed simply once per shard (not worker) regardless of the tests which will be run per shard.

There clearly is no silver bullet. However, I think that this solution gives the playwright user more flexibility to try-out and choose whatever works best for them.

pavelfeldman:
> There clearly is no silver bullet. However, I think that this solution gives the playwright user more flexibility to try-out and choose whatever works best for them.

Going back to my original sentiment, that's exactly why we are hesitant landing the PR. We are looking for the ways to cover more use cases (with hooks) with a smaller api surface / maintenance cost. For example, allow test lists for shards so that users could use third party solutions to tune their exact configuration. Or allow a callback that would take over scheduling tests. Developing those to be easy in maintenance requires consideration and time that we can't currently allocate to the problem. But we are very open to keeping this communication in case a nice proposal comes up.

muhqu:
Would you be open to an approach that allows users to implement their own test filter? …basically similar to grep, but instead of RegEx use a function that takes a list of test infos and returns a filtered list...

That you actually allow users to implement their own sharding logic.

muhqu:
Okay, sounds good. I just looked into filtering and implemented a first version: https://github.com/microsoft/playwright/pull/33049

Issue #32986, opened 2024-10-07
[Feature]: typesafe / generic matchers


sand4rt:
### 🚀 Feature Request

Is it possible for `toMatchObject` and similar functions to be type-safe and accept a generic like jest/vitest?

### Example

```ts
import type { UpdateDto } from '@my/generated-api-client';

expect(request).toMatchObject<UpdateDTO>({ id: '1337', status: 'DRAFT' })
```

### Motivation

I am generating DTO types from an OpenAPI/rest spec using [Orval](https://orval.dev/overview) to keep backend and frontend types in sync. When i regenerate the DTOs, I want my IDE to immediately highlight any failing tests/changed types, without needing to run them manually.

ruifigueira:
I opened PR #33035 with a proposal for generic `toMatchObject` and `toStrictEqual` matchers.

I left tests as `fixme` because they rely on `stable-test-runner`.

dgozman:
@sand4rt If you generate the following, would it work for you as well?

```ts
import type { UpdateDto } from '@my/generated-api-client';

expect(request).toMatchObject({ id: '1337', status: 'DRAFT' } satisfies UpdateDTO);
```

I *think* this could solve your problem more generally, not relying on any third-party library like Playwright to check your types. Let me know what you think.

sand4rt:
@dgozman Ah, i hadn't considered using `satisfies`, but that works too. I think i actually prefer it over the generic approach

ruifigueira:
That settles this then. I'll close my PR

PR #33035, opened 2024-10-10
chore(test): add optional generic parameters to matchers

ruifigueira:
PR with a proposed solution for #32986


tests/playwright-test/types.spec.ts
```diff
@@ -324,3 +324,77 @@ test('config should allow void/empty options', async ({ runTSC }) => {
   });
   expect(result.exitCode).toBe(0);
 });
+
+test.fixme('should check types of overloaded toMatchObject with generic parameter', async ({ runTSC }) => {
```
pavelfeldman, line 328: why is this fixme?

ruifigueira: Those tests rely on stable-test-runner which doesn't have those changes yet

dgozman: These tests should resolve `@playwright/test` from the head, so should work with your change. Otherwise, all our tests would be completely wrong and testing stable-test-runner!

ruifigueira: I just replicated similar type tests like these:

https://github.com/microsoft/playwright/blob/0a63427c778ba447e870fe1ade72fee35c483fe1/tests/playwright-test/expect.spec.ts#L302-L308

What is the best place to put those kind of tests?

I can create some normal tests that eventually `tsc` can validate. I noticed that `lint` npm script triggers `test-types` which  ends up validating test types.

packages/playwright/types/test.d.ts
```diff
@@ -6449,6 +6479,28 @@ interface GenericAssertions<R> {
...
+   * **Usage**
+   *
+   * ```js
+   * const value = { prop: 1 };
+   * expect(value).toStrictEqual({ prop: 1 });
+   * ```
+   *
+   * @param expected Expected value.
+   */
+  toStrictEqual<K>(expected: K): R;
```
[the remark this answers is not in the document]
ruifigueira: Sure. Not all those matchers exist in playwright, but these ones exist without the generic overload, so they're all good candidates:

- `toBe`
- `toBeInstanceOf`
- `toContain`
- `toContainEqual`
- `toEqual`
- `toHaveProperty`
- `toMatchObject`
- `toMatchSnapshot`
- `toStrictEqual`

I was also checking their generic types.

For instance, `toMatchObject`checks against the generic type while I was checking against the "deep partial" type, so I guess it's better to drop that change of mine and keep it like Jest.

tests/playwright-test/types.spec.ts
ruifigueira: You are right, I forgot to build playwright before running my tests. and then I got confused because I thought it was running those tests against this version:

https://github.com/microsoft/playwright/blob/main/tests/playwright-test/stable-test-runner/package.json#L4

docs/src/api/class-genericassertions.md
```diff
@@ -461,6 +473,8 @@ Compares contents of the value with contents of [`param: expected`], performing
 
 When comparing arrays, the number of items must match, and each item is checked recursively.
 
+Optionally, you can provide a type for the expected value via a generic. This is particularly useful for ensuring expected objects have the right structure.
```
dgozman, line 476: Please add an example to the usage.

docs/src/api/class-genericassertions.md
```diff
@@ -494,6 +508,8 @@ Differences from [`method: GenericAssertions.toEqual`]:
 * Array sparseness is checked. For example, `[, 1]` does not match `[undefined, 1]`.
 * Object types are checked to be equal. For example, a class instance with fields `a` and `b` will not equal a literal object with fields `a` and `b`.
 
+Optionally, you can provide a type for the expected value via a generic. This is particularly useful for ensuring expected objects have the right structure.
```
dgozman, line 511: Please add an example to the usage.

dgozman (commented):
I have reviewed the changes, and would like us to provide a real-life examples where the types are useful. The toy examples in the test are not really representative, so before adding this complexity to the code, docs and our users, it should be clear what the intended usage is.

ruifigueira:
I agree with [your comment](https://github.com/microsoft/playwright/issues/32986#issuecomment-2413896824) about using `satisfies` instead, and these changes were not making assertions more type-safe.

PR #33038, opened 2024-10-10
feat(html reporter): show attachments in attachment steps

### LINKED ISSUE #32748: [Feature]: Support attachment for particular test step
### 🚀 Feature Request

I heavily use playwright for API tests. I highly appreciate and extremely love that we can see all requests made by `APIRequestContext` in traces.

However sometimes test scenarios require a lot of API calls so having possibility to see request/response payloads right in the test step in html report would make debugging much easier.

### Example

I am not really good at designing but that's a simple option how it would look like

image

### Motivation

This feature would make debugging much easier not only for people who write automated tests but also for developers who  faced with failed tests  and want to reproduce the failure or just to investigate the root cause.

UPD: I provided single use case I want to have the feature to. But I know that some teams do not use `APIRequestContext` and use some external libraries as http clients. E.g. graphql-codegen for generating api methods on client. In this case it would also help them because they are not going to have network info in traces.
However I believe such feature could be helpful not only for logging API calls.

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32748.

Adds a `TestStep.attachments` array to the Reporter API that shows all attachments for the step. Removes the artificial `attach "foo"` step for all calls except `TestInfo.attach()`. Updates the HTML Report to show attachments in the step list.

Before:
Screenshot 2024-10-10 at 12 00 34

After:
Screenshot 2024-10-16 at 12 15 50

Note that the `Attachments` chip goes away in favour of showing *all* attachments in the steps view. This means there's less repetition in the UI, but it also makes it harder to find deeply nested attachments. I'll also work on an alternative UI to see what else we could show here.


Skn0tt:
Those test failures are all because attachments are now listed twice on the page, so Playwright doesn't know what element to resolve to. I'll update the tests once we've agreed on a UI for this.

packages/playwright/src/reporters/html.ts
```diff
@@ -498,6 +498,7 @@ class HtmlBuilder {
     const { step, duration, count } = dedupedStep;
     const result: TestStep = {
       title: step.title,
+      category: step.category,
```
Skn0tt, line 501: We could also just match on `title` and be fine. So if this has any negative side effects, let me know.

Skn0tt:
Turning into draft to implement what we discussed yesterday:

- TestStep.attachments return its own attachments
- remove attachment steps for everything but explicit TestInfo.attach()
- add UI for linking from steps to attachments to Trace Viewer and HTML report

Skn0tt:
Here's another draft for how the attachments UI could look like. It keeps the existing `Attachments` chip, and links to it from the step list:

https://github.com/user-attachments/assets/73cda2c6-1b7c-4065-a153-39f2a59bf686

It's not very polished, but I think I prefer the current approach where the attachments are inside the step list.

Skn0tt:
notes from call with Dima:

- "attach" step stays
- "attach" step gets attachment icon
- attachment icon somehow reveals attachment bodies (either instead of browser snapshot or in attachment tab)
- attachment tab has some sort of backlink to sources or actions (make it similar to errors)
- that way, trace structure doesn't need any changes
- open question: is attachment icon recursive? does it show the icon if child steps have attachments?
- for `toHaveScreenshot`: make sure that they don't have their own "attach" steps
- our own steps shouldn't rely on zones for parent detection, make it explicit
  - reuse step.complete

Unrelated:
- "expect.toHaveScreenshot" attachment shouldn't say "image diff", but have a title that references to toHaveScreenshot. Ideally another backlink

Skn0tt:
closing in favour of separate PRs

PR #33049, opened 2024-10-10
feat(test-runner): test filter

muhqu:
Adds a new way for playwright users to filter the list of test cases using a function. It's similar to `grep`, but instead of a `RegExp` you can use a function.

With this function it ~should~ is be possible to implement a custom sharding mechanism as a user of playwright. 

## New property: TestConfig.filter
- type: ?<[TestFilter]|[Array]<[TestFilter]>>

Filter test cases by function. `TestFilter` can either be predicate function or an object with `filterTests` or `filterTestGroups` methods.

**Usage**

```js title="playwright.config.ts"

export default defineConfig({
  filter: test => test.title === 'some test',
});
```
Or...
```js
export default defineConfig({
  filter: {
    filterTests: tests => tests.filter((test, index) => index % 2 === 0)
  },
});
```
Or...
```js
export default defineConfig({
  filter: {
    filterTestGroups: testgroups => {
      return testgroups.filter((testgroups, index) => index % 2 === 0)
    },
  },
});
```


packages/playwright/types/test.d.ts
```diff
@@ -1838,6 +1850,8 @@ export type TestDetails = {
   annotation?: TestDetailsAnnotation | TestDetailsAnnotation[];
 }
 
+export type TestFilter = (tests: TestCase[]) => TestCase[];
```
muhqu, line 1853: `(test: TestCase) => boolean` would probably be more straight forward here, but it also gives less context… e.g. number of test cases.

Maybe it's possible to support multiple signatures? e.g. `(test: TestCase) => boolean | (tests: TestCase[]) => TestCase[]` 
…but then it could be difficult to detect which one is used at runtime…

muhqu: Changed it in 5e212da2beb7bfe02af6f8077a1c039894ac1d8f to  ... https://github.com/microsoft/playwright/blob/5e212da2beb7bfe02af6f8077a1c039894ac1d8f/packages/playwright/types/test.d.ts#L1870-L1873

packages/playwright/types/test.d.ts
```diff
@@ -18,6 +18,8 @@
 import type { APIRequestContext, Browser, BrowserContext, BrowserContextOptions, Page, LaunchOptions, ViewportSize, Geolocation, HTTPCredentials, Locator, APIResponse, PageScreenshotOptions } from 'playwright-core';
 export * from 'playwright-core';
 
+import type { TestCase } from './testReporter';
```
muhqu, line 21: I'm not sure whether it is okay to simply cross reference `TestCase` here…

packages/playwright/types/test.d.ts
```diff
@@ -1838,6 +1885,11 @@ export type TestDetails = {
   annotation?: TestDetailsAnnotation | TestDetailsAnnotation[];
 }
 
+type TestFilterFunction = (test: TestCase) => boolean;
+type TestsFilter = { filterTests(tests: TestCase[]): TestCase[] }
+type TestGroupsFilter = { filterTestGroups(testGroups: { tests: TestCase[] }[]): { tests: TestCase[] }[] }
```
muhqu, line 1890: I was not sure whether it would be okay to expose `TestGroup` ? https://github.com/microsoft/playwright/blob/e215fb5c482ae9644de24ae532498f9968db0079/packages/playwright/src/runner/testGroups.ts#L19-L25

packages/playwright/types/test.d.ts
pavelfeldman: It is and it isn't, but let's solve other bits before resolving it.

pavelfeldman:
When is testgroups.filter useful?

muhqu:
Test group filtering is useful to make sure tests, which playwright wants to executed together, stay together (e.g. when they need to run sequential…). The logic which tests should be grouped together stays internal to playwright and all a user of playwright should know is which tests are grouped together, not why.

If we do not expose test groups, users of playwright need to make assumptions of which tests should belong together, which is suboptimal as it might break playwrights contracts (e.g. when tests need to run in sequence…)

dgozman:
@muhqu Thank you for the explanation. We've discussed this proposal, and here are some thoughts:
- It seems like operating on test groups is the only viable solution among the ones outlined above.
- We would really like a low level API to also be able to control the order of test execution. For example, one could imagine a strategy that would "run last failures first, and only if all of them pass - run all other tests". Ideally, we want strategies like this to be implementable on top of the proposed API.
- All of this is complicated by dependencies, serial mode, worker hashes, and probably more things that I didn't think about yet.

Given the above, I'd recommend to file an issue first, where we can outline the requirements and considerations before arriving to an API. We judge the interest of the community in the topic by engagement numbers on the issue, so that would be important for us.

Additionally, given an issue, we will also be able to experiment with some ideas from our side, instead of blocking you for a long time or asking to try things that would eventually lead to nowhere. For example, one of the ideas we discussed is allowing tests with different worker hashes in the same worker, which might simplify the custom scheduling API quite a bit.

muhqu:
@dgozman ok, this helps me understand what was meant by “low level api”. But given your list of requirements is clearly larger than mine, I think it makes more sense you come up with a feature request to make the test runner architecture more extensible… e.g. pluggable test runner strategies… that’s beyond what I was trying to solve.

lukpsaxo:
@dgozman I'd love to see this not die - there is an issue linked here - https://github.com/microsoft/playwright/issues/17969 with 41 up votes. It would be great to see the team engage in the issue and..

> outline the requirements and considerations before arriving to an API

we bet on muhqu's PR being merged, but we are at the point where there is a 20 minute difference between agents of our test runs and we would like to know if we should go back to our approach of splitting the tests ourselves and writing out a different config file per shard.

PR #32127, opened 2024-08-13
fix(codegen): fix unselect issue

mtsmfm:
Fixes https://github.com/microsoft/playwright/issues/31290

PR #33060, opened 2024-10-11
test: fix flaky unselect test

mxschmitt:
This is flaky since it was added in https://github.com/microsoft/playwright/pull/32127.

Follow-up to https://github.com/microsoft/playwright/pull/32127#discussion_r1714671469.

Added an isolated test for the WebKit/Linux  bug in https://github.com/microsoft/playwright/issues/33064.

### DIFF
--- a/tests/library/hit-target.spec.ts
+++ b/tests/library/hit-target.spec.ts
@@ -15,7 +15,7 @@
  */
 
 import { contextTest as it, expect } from '../config/browserTest';
-import type { ElementHandle } from 'playwright-core';
+import type { ElementHandle, Page } from 'playwright-core';
 
 declare const renderComponent;
 declare const e;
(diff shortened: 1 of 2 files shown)


tests/library/inspector/cli-codegen-1.spec.ts
```diff
@@ -682,41 +682,20 @@ await page.Locator(\"#age\").SelectOptionAsync(new[] { \"2\" });`);
...
   test('should select with multiple attribute', async ({ openRecorder }) => {
     const { page, recorder } = await openRecorder();
 
     await recorder.setContentAndWait(`<select id="age" multiple onchange="console.log('[' + [...age.selectedOptions].map(x => x.value).join(',') + ']')"><option value="1">1</option><option value="2">2</option></select>`);
 
     const locator = await recorder.hoverOverElement('select');
     expect(locator).toBe(`locator('#age')`);
-    await clickMultipleSelectOption(page.getByRole('option', { name: '1' }));
+    await page.locator('select').focus();
+    await page.selectOption('select', ['1']);
```
mxschmitt, line 693: note, if you don't focus, it throws [here](https://github.com/microsoft/playwright/blob/c105de4436cc47b54a243bca2b44d12aa3a82d77/packages/playwright-core/src/server/injected/recorder/recorder.ts#L430) that `activeModel` is null.

dgozman (approved):

tests/library/hit-target.spec.ts
```diff
@@ -464,3 +464,41 @@ it('should click in custom element', async ({ page }) => {
   await page.locator('input').click();
   expect(await page.evaluate('window.__clicked')).toBe(true);
 });
+
+for (const mode of ['same-context', 'different-context'] as const) {
+  it(`concurrent selectOption in ${mode} should work`, async ({ browser, browserName, isLinux }) => {
```
mxschmitt, line 469: Note #1: Was not sure in which file to put this test, since it needs to be a library test (not a page test).

Note #2: It passes in Firefox and Chromium.

PR #33074, opened 2024-10-12
feat(html-reporter): recognize video attachment by the contentType

### LINKED ISSUE #33073: [Feature]: It should be possible to see video preview for video attachment with custom name
### 🚀 Feature Request

Giving a custom name to the attachment should not prevent displaying a video preview on the report.

Currently one needs to name all video attachments as `video`, only then preview is available:
```await testInfo.attach('video', { path, contentType: 'video/webm' });```
Image

If one gives more meaningful names, preview is not available:
```await testInfo.attach('customName', { path, contentType: 'video/webm' });```
Image

I found that `html-reporter` in [testResultViewer.tsx](https://github.com/microsoft/playwright/blob/main/packages/html-reporter/src/testResultView.tsx) recognizes if a given attachment is a video attachment through name:
```
const screenshots = new Set(attachments.filter(a => a.contentType.startsWith('image/')));
const videos = attachments.filter(a => a.name === 'video'); // <---- here is the code I refer to
const traces = attachments.filter(a => a.name === 'trace');
const htmls = attachments.filter(a => a.contentType.startsWith('text/html'));
```

The suggestion is to recognize videos by `contentType`.

### Example

In case a test involves multiple pages, we could associate a meaningful name with each page, making report analysis easier.

### Motivation

In case of tests involving multiple pages/videos it will make analysis of the report easier.

### PR DESCRIPTION
**Pull request by RemigiuszDudek**
Closes https://github.com/microsoft/playwright/issues/33073.

### DIFF
--- a/packages/html-reporter/src/testResultView.tsx
+++ b/packages/html-reporter/src/testResultView.tsx
@@ -70,7 +70,7 @@ export const TestResultView: React.FC<{
   const { screenshots, videos, traces, otherAttachments, diffs, errors, htmls } = React.useMemo(() => {
     const attachments = result?.attachments || [];
     const screenshots = new Set(attachments.filter(a => a.contentType.startsWith('image/')));
-    const videos = attachments.filter(a => a.name === 'video');
+    const videos = attachments.filter(a => a.contentType === 'video/webm');
     const traces = attachments.filter(a => a.name === 'trace');
     const htmls = attachments.filter(a => a.contentType.startsWith('text/html'));
     const otherAttachments = new Set<TestAttachment>(attachments);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/html-reporter/src/testResultView.tsx [hunk 1]
```diff
@@ -70,7 +70,7 @@ export const TestResultView: React.FC<{
   const { screenshots, videos, traces, otherAttachments, diffs, errors, htmls } = React.useMemo(() => {
     const attachments = result?.attachments || [];
     const screenshots = new Set(attachments.filter(a => a.contentType.startsWith('image/')));
-    const videos = attachments.filter(a => a.name === 'video');
+    const videos = attachments.filter(a => a.contentType === 'video/webm');
```
dgozman, line 73: We should allow all video types:

```suggestion
    const videos = attachments.filter(a => a.contentType.startsWith('video/'));
```

dgozman (commented):
Thank you for the PR!

packages/html-reporter/src/testResultView.tsx
RemigiuszDudek: fair point, I wanted to be conservative in order not to introduce regression but if you say that all video types are supported I'm ok with your suggestion.

dgozman (approved):
Thank you, this looks great, merging in!

PR #33079, opened 2024-10-13
docs: update to "Matching one of the two alternative locators" section

Anand-M-Cherian:
The section for "Matching one of the two alternative locators" mentioned "matching all of the alternatives". However, since we are using or() operator, it should match "any one of the alternatives" to locate the element.

update_locators md_file

@playwrightmachine If my understanding is wrong, kindly correct me.

### DIFF
--- a/docs/src/locators.md
+++ b/docs/src/locators.md
@@ -1218,7 +1218,7 @@ var button = page.GetByRole(AriaRole.Button).And(page.GetByTitle("Subscribe"));
 
 ### Matching one of the two alternative locators
 
-If you'd like to target one of the two or more elements, and you don't know which one it will be, use [`method: Locator.or`] to create a locator that matches any one of the alternatives.
+If you'd like to target one of the two or more elements, and you don't know which one it will be, use [`method: Locator.or`] to create a locator that matches any one or both of the alternatives.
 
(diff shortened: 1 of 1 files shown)


docs/src/locators.md
```diff
@@ -1218,7 +1218,7 @@ var button = page.GetByRole(AriaRole.Button).And(page.GetByTitle("Subscribe"));
 
 ### Matching one of the two alternative locators
 
-If you'd like to target one of the two or more elements, and you don't know which one it will be, use [`method: Locator.or`] to create a locator that matches all of the alternatives.
+If you'd like to target one of the two or more elements, and you don't know which one it will be, use [`method: Locator.or`] to create a locator that matches any one of the alternatives.
```
dgozman, line 1221:
```suggestion
If you'd like to target one of the two or more elements, and you don't know which one it will be, use [`method: Locator.or`] to create a locator that matches any one or both of the alternatives.
```

dgozman (commented):
Thank you for the PR! Since the locator can match either the first, or the second, or both at the same time, I'd suggest to be as explicit as possible. WDYT?

docs/src/locators.md
Anand-M-Cherian: I think that is an even more comprehensive explanation of the or() operator leaving no room for doubt for an absolute beginner. Let's go ahead with this.

dgozman (approved):
Looks great, merging in!

Anand-M-Cherian:
image
After the changes were deployed....
Thank you @dgozman

PR #33083, opened 2024-10-14
fix(test runner): when sharding with beforeAll, use shards total instead of workers

### LINKED ISSUE #33077: [Bug]: Playwright not sharding tests evenly
### Version

^1.48.0(latest)

### Steps to reproduce

Example steps to reproduce.

1. Copy this code into a file called `tests/shard-test.spec.js`
```js

const arr = Array.from({ length: 2000 }, (_, i) => i + 1);

test.beforeAll(async () => {
    console.log('Before tests');
});

arr.forEach((num) => {
    test(`@smoke should check Math.pow functionality for ${num}`, async ({ page }) => {
        expect(num ** 2).toBe(Math.pow(num, 2));
    });
})
```
2. Also have fullyParallel set to true in your playwright config file
3. Then run the test using this command `npx playwright test tests/shard-test.spec.js --shard=2/8`
4. Look at the logs and you will notice that no tests run.
5. However on shard 1 it gets 500 tests allocated to it(when in fact each shard should get 2000/8 = 250 tests)

### Expected behavior

- Each shard should get same amount of tests allocated to it.
- So in this example each shard should get 2000/8 = 250 tests each.
- And finally some shards should not exit immediately `doing no work`

### Actual behavior

- Bug in playwright where tests are not sharded evenly
- Noticed that this happens if I added test.beforeAll in this file. If I comment that out, then it works fine.

### Additional context

_No response_

## System:
 - OS: Windows 10 10.0.19045
 - CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
 - Memory: 6.76 GB / 15.88 GB
## Binaries:
 - Node: 20.16.0 - C:\Program Files\nodejs\node.EXE
 - npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
## IDEs:
 - VSCode: 1.94.0 - C:\Users\Owner\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
## Languages:
 - Bash: 5.1.16 - C:\WINDOWS\system32\bash.EXE
## npmPackages:
 - @playwright/test: ^1.48.0 => 1.48.0

dgozman:
Otherwise, we might split the `beforeAll`-grouped test group into `workers` parts instead of `shard.total` parts as the user would expect.

Fixes #33077.

### DIFF
--- a/packages/playwright/src/runner/loadUtils.ts
+++ b/packages/playwright/src/runner/loadUtils.ts
@@ -176,8 +176,11 @@ export async function createRootSuite(testRun: TestRun, errors: TestError[], sho
   if (config.config.shard) {
     // Create test groups for top-level projects.
     const testGroups: TestGroup[] = [];
-    for (const projectSuite of rootSuite.suites)
-      testGroups.push(...createTestGroups(projectSuite, config.config.workers));
+    for (const projectSuite of rootSuite.suites) {
+      // Split beforeAll-grouped tests into "config.shard.total" groups when needed.
+      // Later on, we'll re-split them between workers by using "config.workers" instead.
+      testGroups.push(...createTestGroups(projectSuite, config.config.shard.total));
+    }
 
     // Shard test groups.
     const testGroupsInThisShard = filterForShard(config.config.shard, testGroups);
--- a/packages/playwright/src/runner/testGroups.ts
+++ b/packages/playwright/src/runner/testGroups.ts
@@ -24,7 +24,7 @@ export type TestGroup = {
   tests: TestCase[];
 };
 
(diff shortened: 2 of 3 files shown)


pavelfeldman (approved):

rohitkrishna094:
@dgozman Thanks for the quick fix. I see that this has 1.49 release label. Do you know when this will be released? Does playwright have a monthly release?

dgozman:
@rohitkrishna094 Yes, Playwright has releases approximately once a month. That said, 1.48 has just been released, so it will be a while until 1.49. Meanwhile, you can try a [canary release](https://playwright.dev/docs/canary-releases).

PR #33087, opened 2024-10-14
fix(har): account for reused sockets

### LINKED ISSUE #32960: [internal] Investigate fetch timings when Socket is reused
I wonder what does this mean for the timings? If the socket is reused, will it get zero `connect` timings? What if sending does not start right away - should that count as `blocked` or `queued`?

_Originally posted by @dgozman in https://github.com/microsoft/playwright/issues/32956#issuecomment-2393704472_

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/32960

If the socket is reused, the connect and DNS timings are set to -1, because that timing doesn't apply to the current request. The time between request start and the socket being free is counted as `blocked`.

### DIFF
--- a/packages/playwright-core/src/server/fetch.ts
+++ b/packages/playwright-core/src/server/fetch.ts
@@ -329,6 +329,11 @@ export abstract class APIRequestContext extends SdkObject {
             blocked: -1,
           };
 
+          if (request.reusedSocket) {
+            timings.connect = -1;
+            timings.dns = -1;
+          }
+
           const requestFinishedEvent: APIRequestFinishedEvent = {
             requestEvent,
             httpVersion: response.httpVersion,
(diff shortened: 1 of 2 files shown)


packages/playwright-core/src/server/fetch.ts
```diff
@@ -491,8 +498,14 @@ export abstract class APIRequestContext extends SdkObject {
       request.on('socket', socket => {
         // happy eyeballs don't emit lookup and connect events, so we use our custom ones
         const happyEyeBallsTimings = timingForSocket(socket);
-        dnsLookupAt = happyEyeBallsTimings.dnsLookupAt;
-        tcpConnectionAt ??= happyEyeBallsTimings.tcpConnectionAt;
+        if (request.reusedSocket) {
+          reusedSocketAt = monotonicTime();
+          dnsLookupAt = startAt;
```
dgozman, line 503: Perhaps set both `dnsLookupAt` and `tcpConnectionAt` to `undefined`? Would that make the code cleaner?

dgozman (approved):

packages/playwright-core/src/server/fetch.ts
Skn0tt: Good idea. I've refactored it a little more, how do you like that?

dgozman (approved):

PR #33095, opened 2024-10-14
fix(routeWebSocket): make sure ws url without trailing slash is supported

### LINKED ISSUE #33085: [Bug]: WebSocket route does not handle full URLs in Playwright
### Version

1.48.0

### Steps to reproduce

Pre-req: Ensure you have playwright repo already setup

1. Use page.routeWebSocket() to intercept a WebSocket URL
2. Add the below test and run it
```
test('intercepted websocket', async ({ page }) => {
        await page.routeWebSocket('wss://echo.websocket.org/', ws => {
            ws.onMessage(message => {
                if (message === 'Hello chatbot') {
                    ws.send('Intercepted Hello chatbot');  // Intercept and modify the response.
                } else {
                    ws.send(message);  // Echo the received message otherwise.
                }
            });
        });

        // Navigate to the WebSocket page
        await page.goto('https://echo.websocket.org/.ws');

        // Interact with the page's UI to send a WebSocket message
        await page.getByRole('button', { name: 'Pause Messaging' }).click();  // Clicking a button.
        await page.locator('#content').fill('Hello chatbot');  // Filling input with a message.
        await page.getByRole('button', { name: 'Send Message' }).click();  // Sending the message.

        // Assert that the console has the expected text
        await expect(page.locator('#console div')).toHaveText(['attempting to connect']);
    });
```

### Expected behavior

I was expecting to see "Intercepted Hello chatbot" shown

I was also expecting the following to work but it didn't

await page.routeWebSocket('/echo.websocket.org/'...

### Actual behavior

Observe that nothing happened. Looks like no intercept occured. This was posted in discord and the following was spotted

mxschmitt — 
do you mind filing a bug on GitHub?
Looks like its parsed as glob and the wss:// confuses the matching logic.
to make it work you can do:
  await page.routeWebSocket(/echo\.websocket\.org/, ws => {

This worked after making the above change

### Additional context

_No response_

### Environment

  System:
    OS: macOS 14.6.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 56.78 MB / 18.00 GB
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    npm: 10.7.0 - ~/.nvm/versions/node/v20.14.0/bin/npm
  IDEs:
    VSCode: 1.93.1 - /usr/local/bin/code
  Languages:
    Bash: 3.2.57 - /bin/bash
  npmPackages:
    @playwright/test: ^1.48.0 => 1.48.0

dgozman:
Fixes #33085.

### DIFF
--- a/tests/library/route-web-socket.spec.ts
+++ b/tests/library/route-web-socket.spec.ts
@@ -508,3 +508,27 @@ test('should throw when connecting twice', async ({ page, server }) => {
   const error = await promise;
   expect(error.message).toContain('Already connected to the server');
 });
+
+test('should work with no trailing slash', async ({ page, server }) => {
+  const log: string[] = [];
+  // No trailing slash!
+  await page.routeWebSocket('ws://localhost:' + server.PORT, ws => {
+    ws.onMessage(message => {
+      log.push(message as string);
+      ws.send('response');
+    });
+  });
+
+  await page.goto('about:blank');
+  await page.evaluate(({ port }) => {
+    window.log = [];
+    // No trailing slash!
+    window.ws = new WebSocket('ws://localhost:' + port);
+    window.ws.addEventListener('message', event => window.log.push(event.data));
+  }, { port: server.PORT });
+
+  await expect.poll(() => page.evaluate(() => window.ws.readyState)).toBe(1);
+  await page.evaluate(() => window.ws.send('query'));
+  await expect.poll(() => log).toEqual(['query']);
+  expect(await page.evaluate(() => window.log)).toEqual(['response']);
+});
--- a/packages/playwright-core/src/server/injected/webSocketMock.ts
+++ b/packages/playwright-core/src/server/injected/webSocketMock.ts
@@ -143,6 +143,7 @@ export function inject(globalThis: GlobalThis) {
 
       this.url = typeof url === 'string' ? url : url.href;
       try {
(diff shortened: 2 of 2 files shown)


tests/library/route-web-socket.spec.ts
```diff
@@ -508,3 +508,27 @@ test('should throw when connecting twice', async ({ page, server }) => {
...
+      log.push(message as string);
+      ws.send('response');
+    });
+  });
+
+  await page.goto('about:blank');
+  await page.evaluate(({ port }) => {
+    window.log = [];
+    // No trailing slash!
+    window.ws = new WebSocket('ws://localhost:' + port);
```
mxschmitt, line 526: I'd try to fix it the other way around, so that we would match `/` to `new WebSocket("ws://localhost:1234")` because on the actual networking layer the WebSocket has a trailing slash, it gets added by the browser. This should also require less hacks for the clients. Since having a WebSocket on `/` seems such a rare scenario, I'd prefer this over a lax matching logic. wdyt?

Looking at https://github.com/microsoft/playwright/blob/a395fb22c4cd414e58a291f2b1d87aa3a3714043/packages/playwright-core/src/server/injected/webSocketMock.ts#L153.

mxschmitt (approved):

PR #33104, opened 2024-10-15
fix deprecation messages for `timeout` arguments that get ignored so that they show as deprecated in the typescript language server

### LINKED ISSUE #33017: [Feature]: remove the deprecated `timeout` argment to `Locator.isVisible`, or make the deprecation more obvious to the user
### 🚀 Feature Request

the argument has been deprecated for a few years now, and the current behavior where it just silently ignores the argument is even more confusing than the previous behavior.

### Example

there is no visible deprecation warning on code that uses it, which makes it very easy to miss for someone who doesn't already know it's deprecated:

image

this could be addressed by turning the method into an overload:

```ts
// playwright-core/types/types.d.ts
export interface Locator {
  /**
   * @deprecated the `timeout` option is ignored.
   * @param options
   */
  isVisible(options: {
    timeout: number;
  }): Promise<boolean>;
  isVisible(): Promise<boolean>;
}
```

which will make vscode display a deprecation hint:

image

this would also need to be done for all the languages (eg. python using the `@deprecated` decorator), which i assume would require updates to the code generation scripts. it's probably not worth the effort and imo it's probably time to just delete the argument entirely since it's been several years since it was deprecated.

in python, the deprecation notice is even easier to miss as it's in the middle of a large docstring:

```py
    def is_visible(self, *, timeout: typing.Optional[float] = None) -> bool:
        """Locator.is_visible

        Returns whether the element is [visible](https://playwright.dev/python/docs/actionability#visible).

        **NOTE** If you need to assert that element is visible, prefer `locator_assertions.to_be_visible()` to avoid
        flakiness. See [assertions guide](https://playwright.dev/python/docs/test-assertions) for more details.

        **Usage**

        ```py
        visible = page.get_by_role(\"button\").is_visible()
        ```

        Parameters
        ----------
        timeout : Union[float, None]
            Deprecated: This option is ignored. `locator.is_visible()` does not wait for the element to become visible and returns immediately.

        Returns
        -------
        bool
        """

        return mapping.from_maybe_impl(
            self._sync(self._impl_obj.is_visible(timeout=timeout))
        )
```

especially considering that so many functions take a `timeout` argument, it's very unlikely that the user is going to look there for a description of what that argument does


### Motivation

i just reviewed someone's pull request where they used hardcoded `page.wait_for_timeout` calls all over the place because they were relying on `locator.is_visible` with a timeout, didn't realize it was silently being ignored and assumed that adding an additional wait beforehand was necessary.

i think silently ignoring an argument is a bad way to deprecate something, because it renders code that uses it completely incorrect with no error or warning to the user. if you ask me, this is even worse than just deleting the argument out of the blue with no deprecation phase at all.

IMO when it comes to deprecations, the thing being deprecated should at least still function until its eventual removal, or at the very least provide an IDE or type checker warning to the user.

DetachHead:
fixes #33017

### DIFF
--- a/packages/playwright-core/types/types.d.ts
+++ b/packages/playwright-core/types/types.d.ts
@@ -3501,13 +3501,31 @@ export interface Page {
      * element, the call throws an exception.
      */
     strict?: boolean;
+  }): Promise<boolean>;
+  /**
+   * **NOTE** Use locator-based [locator.isHidden([options])](https://playwright.dev/docs/api/class-locator#locator-is-hidden)
+   * instead. Read more about [locators](https://playwright.dev/docs/locators).
+   *
+   * Returns whether the element is hidden, the opposite of [visible](https://playwright.dev/docs/actionability#visible).
+   * [`selector`](https://playwright.dev/docs/api/class-page#page-is-hidden-option-selector) that does not match any
+   * elements is considered hidden.
+   * @deprecated The `timeout` option is ignored.
+   * @param selector A selector to search for an element. If there are multiple elements satisfying the selector, the first will be
+   * used.
+   * @param options
+   */
+  isHidden(selector: string, options: {
+    /**
+     * When true, the call requires selector to resolve to a single element. If given selector resolves to more than one
+     * element, the call throws an exception.
+     */
+    strict?: boolean;
 
     /**
-     * @deprecated This option is ignored.
      * [page.isHidden(selector[, options])](https://playwright.dev/docs/api/class-page#page-is-hidden) does not wait for
      * the element to become hidden and returns immediately.
      */
-    timeout?: number;
+    timeout: number;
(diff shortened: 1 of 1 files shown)


DetachHead:
oh, that file is generated

mxschmitt:
hey! are you planning to work on it? We try to keep the number of open PRs low - so if you don't, I'd close it for now so that it doesn't show up in our internal dashboards as untriaged.

DetachHead:
sorry i assumed leaving it as draft would be fine until i get around to looking at this again. will close it for now

xmedeko:
Yep, it's very confusing this is not in doc, see my SO question https://stackoverflow.com/questions/79416204/what-is-flaky-in-playwright-locator-dom-state-getters/79416267

PR #33105, opened 2024-10-15
chore: generate match snapshot

pavelfeldman:


packages/playwright-core/src/server/injected/recorder/recorder.ts
```diff
@@ -704,6 +704,18 @@ class TextAssertionTool implements RecorderTool {
           value: (target as (HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement)).value,
         };
       }
+    } if (this._kind === 'snapshot') {
```
yury-s, line 707:
```suggestion
    }
    if (this._kind === 'snapshot') {
```

yury-s (approved):

packages/playwright-core/src/server/injected/recorder/recorder.ts
pavelfeldman: I'll follow up!

PR #33114, opened 2024-10-15
fix(trace viewer): Screenshot click target placement

### LINKED ISSUE #32847: [internal]: screenshot instead of snapshot polishing
Before we ship "screenshot instead of snapshot", let's smooth out the rough edges.

- [ ] attempt to select text causes image drag-n-drop - https://github.com/microsoft/playwright/pull/33089
- [ ] click target is misplaced
- [ ] discoverability: add a hint inside canvas in DOM snapshot - https://github.com/microsoft/playwright/pull/33092
- [x] #33115

Skn0tt:
This closes the last element of https://github.com/microsoft/playwright/issues/32847, so it reverts the changes where we hide the toggle.

Misplacement before:

https://github.com/user-attachments/assets/5805ab42-000a-4691-836e-6ecf3b657fa2

Correct placement after:

https://github.com/user-attachments/assets/73a443d1-265f-4b67-be33-6eed334c933f

The bug was two-fold.

1. Because it's rendered as `position: fixed`, we need to take the `<BrowserFrame />`'s height into account.
2. Our centering with `margin: -10px 0 0 -10px` was relying on a 20px width that wasn't perfect. `transform` works better, learned that through https://css-tricks.com/quick-css-trick-how-to-center-an-object-exactly-in-the-center/.

Closes https://github.com/microsoft/playwright/issues/32847


dgozman (approved):

Skn0tt:
closing in favour of https://github.com/microsoft/playwright/pull/33119

PR #33119, opened 2024-10-15
feat(tracing): clip `canvas` contents from screenshots

### LINKED ISSUE #33115: [internal] try to clip canvas from the image and stick on top of the canvas in DOM

Skn0tt:
Closes https://github.com/microsoft/playwright/issues/33115.

If we're happy with it, this can supercedes the "show screenshot instead of snapshot" implementation. Will remove that in a separate PR.

This PR adds logic to snapshot rendering that fetches the closest screenshot for the rendered snapshot to clip the canvas contents from it. This is best-effort, and the time alignment won't always be perfect. This is fine if there's no scrolling and the canvas has static contents. We show a warning icon to let the user know about that. Because it's based on viewport screenshots, we don't have data outside of it. Parts of the canvas that aren't available will be shown with a checkered background. Here's a demo:

https://github.com/user-attachments/assets/e23cb827-a6e5-4817-84e2-fa332f0669af


tests/library/trace-viewer.spec.ts
```diff
@@ -1439,6 +1439,15 @@ test.skip('should allow showing screenshots instead of snapshots', async ({ runA
   await expect(screenshot).toBeVisible();
 });
 
+test('canvas clipping', async ({ runAndTrace, page, server }) => {
+  const traceViewer = await runAndTrace(async () => {
+    await page.goto(server.PREFIX + '/screenshots/canvas.html');
+    await page.waitForTimeout(1000); // ensure we could take a screenshot
+  });
+
+  await traceViewer.page.pause();
```
Skn0tt, line 1448: this test isn't gonna stay, obviously. but you can use it to play around with the feature: `npm run ctest -- --grep 'clipping' --headed`

@dgozman what's the best way of testing this - should we trace a page with a `canvas` and then take a screenshot of the trace, to ensure the screenshot isn't blank? Or somehow compute if it's blank without storing a golden image?

packages/trace-viewer/src/sw/snapshotServer.ts
```diff
@@ -18,24 +18,46 @@ import type { URLSearchParams } from 'url';
...
     const snapshot = this._snapshot(pathname.substring('/snapshot'.length), searchParams);
     if (!snapshot)
       return new Response(null, { status: 404 });
-    const renderedSnapshot = snapshot.render();
+
+    let screenshotUrl: URL | undefined;
+    const { wallTime, timestamp, pageId } = snapshot.snapshot();
+    const page = this._pages.get(pageId);
+    if (page) {
+      const closestFrame = (wallTime && page.screencastFrames[0]?.frameSwapWallTime) ? findClosest(page.screencastFrames, frame => frame.frameSwapWallTime!, wallTime) : findClosest(page.screencastFrames, frame => frame.timestamp, timestamp);
```
dgozman, line 55: I'd prefer this code to live in `SnapshotRenderer` instead.

packages/trace-viewer/src/sw/snapshotServer.ts
```diff
@@ -18,24 +18,46 @@ import type { URLSearchParams } from 'url';
...
+      if (closestFrame)
+        screenshotUrl = new URL(`./sha1/${closestFrame.sha1}`, swScope);
```
dgozman, line 57: Injecting `swScope` here is unfortunate. Perhaps we should have a relative url for the screenshot, e.g. `/screenshot/<pageOrFrameId>`, and route it through sw main?

packages/trace-viewer/src/sw/snapshotRenderer.ts
```diff
@@ -299,6 +299,25 @@ function snapshotScript(...targetIds: (string | undefined)[]) {
...
+          for (const canvas of canvases) {
+            const context = canvas.getContext('2d')!;
+
+            const boundingRect = canvas.getBoundingClientRect();
+            const xStart = boundingRect.left / window.innerWidth;
+            const yStart = boundingRect.top / window.innerHeight;
+            const xEnd = boundingRect.right / window.innerWidth;
+            const yEnd = boundingRect.bottom / window.innerHeight;
+
+            context.drawImage(img, xStart * img.width, yStart * img.height, (xEnd - xStart) * img.width, (yEnd - yStart) * img.height, 0, 0, canvas.width, canvas.height);
```
dgozman, line 315: - Is this going to work nicely for the out-of-viewport canvas? Perhaps we should say "not captured" there instead?
- This code must run after `scrollLeft/scrollTop` are restored in `onLoad()`.
- We should still have a warning somewhere that would explain that image might be wrong.

tests/library/trace-viewer.spec.ts
dgozman: I think that would be a flaky test, unfortunately. Perhaps just check some test-specific logging, like we do in recorder tests:

```js
if (isUnderTest) console.error(`drawn 15;23-184;566 screenshot rect on the canvas`);
```

packages/trace-viewer/src/sw/snapshotRenderer.ts
```diff
@@ -299,6 +299,25 @@ function snapshotScript(...targetIds: (string | undefined)[]) {
         }
       }
 
+      const canvases = root.querySelectorAll('canvas');
+      if (canvases.length > 0 && screenshotURL) {
+        const img = new Image();
+        img.onload = () => {
```
dgozman, line 305: Please handle `img.onerror` as well.

packages/trace-viewer/src/sw/snapshotServer.ts
Skn0tt: Would `SnapshotRenderer` get access to `screencastFrames`? Would we pass them into `.render()`, or do we pass them into the constructor?

packages/trace-viewer/src/sw/snapshotServer.ts
Skn0tt: I had that in a first iteration, but that requires making sw main listen to not just `swScope`, but also to requests from the `iframe` origin. Would that be fine? Alternatively, we could also use a fake origin like `playwright.sw`.

Skn0tt:
feedback from today:

- make the pattern a grey & white checkerboard
- put warning item in the corner

packages/trace-viewer/src/sw/snapshotRenderer.ts
Skn0tt: done

packages/trace-viewer/src/sw/snapshotRenderer.ts
Skn0tt: implemented all of it

packages/trace-viewer/src/sw/snapshotServer.ts
Skn0tt: I've since moved the code a little, but it's still inside the server. Let me know how to proceed.

packages/trace-viewer/src/sw/snapshotServer.ts
Skn0tt: made it work with a new route in the service worker 👍

tests/library/trace-viewer.spec.ts
Skn0tt: thanks, good idea! implemented

Skn0tt (commented):
feedback addressed :)

Skn0tt:
I've seen that this doesn't perfectly work on Firefox - gonna investigate more.

packages/trace-viewer/src/sw/snapshotServer.ts
Skn0tt: we discussed this offline and ended up with https://github.com/microsoft/playwright/pull/33119/commits/45d1b54226e31615e26bbdb6e47830c4cc665e48.

Skn0tt:
> I've seen that this doesn't work perfectly on Firefox - gonna investigate more.

Seems to be a known issue with screencast frames in Firefox in headed moe.

Skn0tt:
Alright, this is ready for one last review!

tests/library/trace-viewer.spec.ts
```diff
@@ -1439,6 +1439,32 @@ test.skip('should allow showing screenshots instead of snapshots', async ({ runA
   await expect(screenshot).toBeVisible();
 });
 
+test('canvas clipping', async ({ runAndTrace, page, server }) => {
+  const traceViewer = await runAndTrace(async () => {
+    await page.goto(server.PREFIX + '/screenshots/canvas.html#canvas-on-edge');
+    await page.waitForTimeout(1000); // ensure we could take a screenshot
```
dgozman, line 1445: I predict this will be flaky 😄 We usually do a few rafs in the hopes it will redraw, but even that does not always help.

packages/trace-viewer/src/sw/snapshotRenderer.ts
```diff
@@ -393,6 +418,73 @@ function snapshotScript(...targetIds: (string | undefined)[]) {
...
+            context.fillRect(pattern.width / 2, pattern.height / 2, pattern.width, pattern.height);
+            return context.createPattern(pattern, 'repeat')!;
+          }
+
+          context.fillStyle = createCheckerboardPattern();
+          context.fillRect(0, 0, canvas.width, canvas.height);
+        }
+
+
+        if (window.parent.parent !== window.parent) {
```
dgozman, line 442: There is `isTopFrame` somewhere around already, let's reuse it.

packages/trace-viewer/src/sw/snapshotRenderer.ts
```diff
@@ -393,6 +418,73 @@ function snapshotScript(...targetIds: (string | undefined)[]) {
           }
         }
       }
+
+      if (canvasElements.length > 0) {
+        function drawWarningBackground(context: CanvasRenderingContext2D, canvas: HTMLCanvasElement) {
```
dgozman, line 423: It's not really a warning anymore, just a checkerboard.

packages/trace-viewer/src/sw/snapshotRenderer.ts
```diff
@@ -393,6 +418,73 @@ function snapshotScript(...targetIds: (string | undefined)[]) {
...
+            drawWarningBackground(context, canvas);
+
+            context.drawImage(img, xStart * img.width, yStart * img.height, (xEnd - xStart) * img.width, (yEnd - yStart) * img.height, 0, 0, canvas.width, canvas.height);
+            if (isUnderTest)
+              // eslint-disable-next-line no-console
+              console.log(`canvas drawn:`, JSON.stringify([xStart, yStart, xEnd, yEnd].map(v => Math.floor(v * 100))));
+
+            if (xEnd > 1 || yEnd > 1) {
+              if (xStart > 1 || yStart > 1)
+                canvas.title = `Playwright couldn't capture canvas contents because it's located outside the viewport.`;
```
dgozman, line 471: Let's skip both `drawWarningBackground` and `drawImage` calls in this case.

dgozman (approved):

tests/library/trace-viewer.spec.ts
Skn0tt: What's a raf?

packages/trace-viewer/src/sw/snapshotRenderer.ts
Skn0tt: I've made it skip `drawImage`, but I don't think we should skip `drawCheckerboard` - having that around is still valuable to show that there's something missing.

Skn0tt: I've also made it skip `drawCheckerboard` to align with what we discussed in the meeting. Gonna potentially change that in a follow-up. https://github.com/microsoft/playwright/pull/33119/commits/ea4116e64546dca6b41160c50a42c69e86cc3fcd

tests/library/trace-viewer.spec.ts
Skn0tt: It's [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame)! See https://github.com/Skn0tt/playwright/blob/ea4116e64546dca6b41160c50a42c69e86cc3fcd/tests/page/pageTest.ts#L39.

PR #33122, opened 2024-10-15
chore: fix ff test for codegen

pavelfeldman:


yury-s (approved):
Any insight on why it was slow?

mxschmitt:
Investigation: `recorderPage.getByRole('button', { name: 'Record' }).click();` is retrying an infinite amount of time because `page.pause()` disables the test timeout. This stalls everything and `globalTimeout` kicks in.

PR #33125, opened 2024-10-15
chore: implement locator.ariaSnapshot

pavelfeldman:


docs/src/api/class-locator.md
```diff
@@ -150,6 +150,63 @@ var button = page.GetByRole(AriaRole.Button).And(page.GetByTitle("Subscribe"));
...
+page.get_by_role("link").aria_snapshot()
+```
+
+```csharp
+await page.GetByRole(AriaRole.Link).AriaSnapshotAsync();
+```
+
+**Details**
+
+This method captures the aria snapshot of the given element. The snapshot is a string that represents the state of the element and its children.
```
yury-s, line 183: Consider adding a note that it is not necessarily full aria subtree and may lack some of the nodes.

yury-s (approved):

docs/src/api/class-locator.md
pavelfeldman: I'll need to have a dedicated page for it...

PR #33127, opened 2024-10-16
bug(artifacts) Fix generic screenshot paths causing collisions in multi-sharded DevOps builds

mastrzyz:
Append the sanitized test name to the automatic screenshot mechanism of Playwright.

Why?

Currently the names are very generic and cause conflicts in CI/CD Systems like Azure Devops when we have multiple jobs failing with the same generic name:

image

It's also not very clear which number responds to which testcase when we are able to upload:

image

Fix!

Added the sanitized file name, a good example of how this looks like are the fixed unit tests : 

image

### DIFF
--- a/packages/playwright/src/index.ts
+++ b/packages/playwright/src/index.ts
@@ -627,9 +627,10 @@ class ArtifactsRecorder {
 
   private _createScreenshotAttachmentPath() {
     const testFailed = this._testInfo._isFailure();
+    const sanitizedTestName = this._testInfo._fsSanitizedTestName();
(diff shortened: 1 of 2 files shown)


dgozman:
@mastrzyz Could you please file an issue first? I've got a bunch of questions, and we prefer to discuss things in an issue rather than in a PR, for better discoverability in the future. Thank you!

mxschmitt:
Closing as per above - lets continue the discussion in https://github.com/microsoft/playwright/issues/33143 for now.

Issue #3146, opened 2020-07-24
[Question] Deno Integration?


DemGiran:
Hello, and congratulations on your project. I see that you are heavily invested in trying to integrate it with platforms other than Node. Thank you for maintaining an official Python repo. Have you thought about Deno? https://github.com/denoland/deno

How about making the npm package play well with skypack.dev? This might make  it possible to import Playwright directly from npm? I think there is a potential opportunity there with minimal (if any) additional technical cost. Thoughts?

Thanks for the library! <3

aslushnikov:
> Have you thought about Deno? https://github.com/denoland/deno

@DemGiran we're keeping an eye on deno, but so far there are no plans / user requests to get a better story there.

> How about making the npm package play well with skypack.dev? 

Out of curiosity, how hard would it be?

DemGiran:
https://github.com/rollup/rollup/wiki/pkg.module

mxschmitt:
Since we rely heavily on `child_process` e.g. or other standard Node.js modules we would need to wait until they are available: https://github.com/denoland/deno/blob/master/std/node/README.md
Not sure about other modules tho.

aslushnikov:
Closing this since there are no plans for this for now.

shirshak55:
Is the situation still same in 2021? I think deno has improved a lot. And deno security + single binary feature are too good for many people like me to ditch deno. 

I hope the decision will be reconsidered in future.

Thanks

mquandalle:
I’m also interested in Deno support.
Note that there is a fork of puppeteer with Deno support : https://deno.land/x/puppeteer

josephrocca:
> Since we rely heavily on child_process e.g. or other standard Node.js modules we would need to wait until they are available

@lucacasonato Wondering whether you could add a general comment on this, since you wrote the puppeteer port for Deno. Seems like there's [support for child_process.spawn](https://github.com/denoland/deno_std/pull/785). Were there many hoops to jump through while porting Puppeteer?

@aslushnikov Wondering whether this is worth re-opening, or still something to postpone for now?

ruifortes:
Yeah... let's not stiffen development Microsoft ;-)

andremarcondesteixeira:
I would love to be able to use Deno with Playwright. Please make it happen!

trusktr:
[...]

PR #33128, opened 2024-10-16
fix: improve compatibility of using ESM/TS with Deno

nathanwhit:
### 
Deno does not support custom ESM loaders, which playwright uses to provide TS/ESM support. This currently means that deno users cannot use playwright with ESM. However, deno supports running typescript and ESM natively, so disabling the loader allows this to work.

A second small issue is that deno doesn't require a `package.json` file, and defaults to ESM (as opposed to node, which defaults to CommonJS). This means that currently if you don't have a `package.json` file, when running in deno playwright would incorrectly assume the project is commonjs.

Two small changes here:
- Disable the ESM loader when running in deno (as detected by looking for `process.versions.deno`)
- Consider files to be ESM, even without a package.json present, when running in deno

---

Overall this greatly improves the experience when using playwright with deno, and I hope the changes are minimal enough to be acceptable. Thanks for your time!

### DIFF
--- a/packages/playwright/src/common/configLoader.ts
+++ b/packages/playwright/src/common/configLoader.ts
@@ -347,8 +347,8 @@ export async function loadEmptyConfigForMergeReports() {
 }
 
 export function restartWithExperimentalTsEsm(configFile: string | undefined, force: boolean = false): boolean {
-  // Opt-out switch.
-  if (process.env.PW_DISABLE_TS_ESM)
+  // Opt-out switch. Default to disabled if running in Deno, which supports TS and ESM natively.
+  if (process.env.PW_DISABLE_TS_ESM || typeof process.versions.deno === 'string')
     return false;
 
   // There are two esm loader APIs:
--- a/packages/playwright/src/util.ts
+++ b/packages/playwright/src/util.ts
@@ -293,6 +293,7 @@ function folderIsModule(folder: string): boolean {
   const isDeno = typeof process.versions.deno === 'string';
   if (!packageJsonPath)
     return isDeno;
+  // Rely on `require` internal caching logic.
   const type = require(packageJsonPath).type;
   if (type === 'module')
     return true;
--- a/packages/playwright/src/util.ts
+++ b/packages/playwright/src/util.ts
@@ -289,10 +289,16 @@ export function fileIsModule(file: string): boolean {
 
 function folderIsModule(folder: string): boolean {
   const packageJsonPath = getPackageJsonPath(folder);
+  // deno is ESM unless package.json specifies otherwise
+  const isDeno = typeof process.versions.deno === 'string';
   if (!packageJsonPath)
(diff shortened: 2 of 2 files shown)


nathanwhit:
@microsoft-github-policy-service agree company="Deno Land Inc"

dgozman:
@nathanwhit Thank you for the PR! We expect [custom ESM loaders](https://nodejs.org/api/module.html#customization-hooks) to become stable Node.js API soon, so that's probably some feedback to Deno to implement them and stay compatible with Node.js ecosystem.

For now, we'd prefer not to introduce Deno-specific code, since we haven't seen much interest from the community. I'd recommend to file an issue first, so that we can measure the community interest in supporting Deno.

I'll close this PR for now, since there are no immediate plans to land it. However, we can reopen it later if the Deno support becomes a widely requested feature.

nathanwhit:
I respect the decision, but I would push back a bit on there not being interest – [this issue](https://github.com/microsoft/playwright/issues/3146) requesting deno support from a few years back had dozens of upvotes and people commenting in support of the request.

Would you be open to landing just the one commit? https://github.com/microsoft/playwright/pull/33128/commits/458e6c81fe772d810813849f011a7332ebce6f5c
It's a one line change, and makes a big difference in usability for our users.

marvinhagemeister:
FWIW we keep getting a couple of requests each week in the Deno discord asking for playwright support. We'd love to use playwright ourselves as well to test projects here at Deno. Merging this PR would unblock Deno users on being able to use playwright.

dgozman:
> Would you be open to landing just the one commit? https://github.com/microsoft/playwright/commit/458e6c81fe772d810813849f011a7332ebce6f5c

As a workaround, you can just pass `PW_DISABLE_TS_ESM` environment variable today.

> FWIW we keep getting a couple of requests each week in the Deno discord asking for playwright support.

Could you please file an issue? Unfortunately, it is hard to track requests for Playwright somewhere in Deno discord 😄 However, an issue would be a clear indicator of the interest and a great place for discussion. I assume we are talking exclusively about the recent Deno 2.0 release, and I don't think we've seen it explicitly mentioned on our issue tracker just yet.

> Merging this PR would unblock Deno users on being able to use playwright.

Have you tried running with `PW_DISABLE_TS_ESM`? Is there any feedback? Does everything work? We are hesitant at landing partial support, because it's unclear whether there are more blocks down the road. For example, disabling the custom ESM loader breaks dependencies detection for watch mode, tsx preprocessor, our tsconfig path resolution, etc.

It would be interesting to run Playwright's own tests with deno and see what's passing and what's not. That could help us make an informed decision.

eavidy:
@dgozman I noticed that a PR #38599 related to Bun has been merged recently. Will this PR be reconsidered now?

dgozman:
@eavidy Please file an issue and we'll consider it.

PR #33001, opened 2024-10-08
feat(testType): add support for test.fail.only method

pengooseDev:
### Description

Fixes [issue #30662](https://github.com/microsoft/playwright/issues/30662)

This PR adds support for `test.fail.only` in Playwright Test, allowing developers to focus on a specific test that is expected to fail.

>- Implemented the `test.fail.only` method.
>- Added regression tests for `test.fail.only` to ensure correct behavior.
>- Updated documentation and TypeScript definitions (`type.d.ts`) to include `test.fail.only`.

As this is my first contribution, I would appreciate feedback on the regression tests and documentation updates. If there are any changes needed, I'm happy to make them. :)

PR #33142, opened 2024-10-16
fix: correct types for things like `test.describe.only`

dgozman:
For more esoteric chained method we had the basic description pulled from `test.describe` instead of a specialized description we have in the docs for `test.describe.only`.

Noticed while playing with #33001.

### DIFF
--- a/utils/generate_types/overrides-test.d.ts
+++ b/utils/generate_types/overrides-test.d.ts
@@ -75,52 +75,83 @@ export type TestDetails = {
   annotation?: TestDetailsAnnotation | TestDetailsAnnotation[];
 }
 
-interface SuiteFunction {
-  (title: string, callback: () => void): void;
-  (callback: () => void): void;
-  (title: string, details: TestDetails, callback: () => void): void;
-}
+type TestBody<TestArgs> = (args: TestArgs, testInfo: TestInfo) => Promise<void> | void;
+type ConditionBody<TestArgs> = (args: TestArgs) => boolean;
 
-interface TestFunction<TestArgs> {
-  (title: string, body: (args: TestArgs, testInfo: TestInfo) => Promise<void> | void): void;
-  (title: string, details: TestDetails, body: (args: TestArgs, testInfo: TestInfo) => Promise<void> | void): void;
-}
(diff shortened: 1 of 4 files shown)


utils/generate_types/overrides-test.d.ts
```diff
@@ -75,52 +75,83 @@ export type TestDetails = {
...
+export interface TestType<TestArgs extends KeyValue, WorkerArgs extends KeyValue> {
+  (title: string, body: TestBody<TestArgs & WorkerArgs>): void;
+  (title: string, details: TestDetails, body: TestBody<TestArgs & WorkerArgs>): void;
+
+  only(title: string, body: TestBody<TestArgs & WorkerArgs>): void;
+  only(title: string, details: TestDetails, body: TestBody<TestArgs & WorkerArgs>): void;
+
+  describe: {
+    (title: string, callback: () => void): void;
+    (callback: () => void): void;
```
Skn0tt, line 90: today I learned about anonymous groups!

utils/generate_types/overrides-test.d.ts
```diff
@@ -75,52 +75,83 @@ export type TestDetails = {
...
-  only: TestFunction<TestArgs & WorkerArgs>;
-  describe: SuiteFunction & {
-    only: SuiteFunction;
-    skip: SuiteFunction;
-    fixme: SuiteFunction;
-    serial: SuiteFunction & {
-      only: SuiteFunction;
+    skip(title: string, callback: () => void): void;
+    skip(callback: () => void): void;
+    skip(title: string, details: TestDetails, callback: () => void): void;
```
Skn0tt, line 99: not related to this PR, but I find it surprising that you can't pass a condition to `test.describe.skip`

Skn0tt (commented):
Looks good to me. Left two curiosity questions.

Skn0tt (approved):

utils/generate_types/overrides-test.d.ts
dgozman: If you want to skip the describe conditionally, you put `test.skip(condition)` inside. For historical reasons 😄

PR #33157, opened 2024-10-17
chore: compute aria text consistently with the role accumulated text

pavelfeldman:


tests/page/page-aria-snapshot.spec.ts
```diff
@@ -15,26 +15,373 @@
...
+      Line 2
+      Line 3
+    </p>
+  `);
+
+  await checkAndMatchSnapshot(page.locator('body'), `
+    - paragraph: Line 1 Line 2 Line 3
+  `);
+  await expect(page.locator('body')).toMatchAriaSnapshot(`
+    - paragraph: |
```
yury-s, line 170: Supporting multiline here is weird.

yury-s (approved):

dgozman (approved):

tests/page/page-aria-snapshot.spec.ts
pavelfeldman: It is a convenience line wrapping practice in YAML...

PR #33161, opened 2024-10-17
chore(docs): fix documentation url

leotm:
fix link to home page, consistent with

https://github.com/microsoft/playwright/blob/2d150eec25958f28e0b0ba7ae214572f1df7546b/README.md?plain=1#L5

instead of linking to: Docs > Getting Started > Installation


dgozman (approved):
Thank you for the PR, looks good.

